PR #36, opened 2024-05-06
Logos on the TokenSelectionView

LF0044:
Added the coin logos for the TokenSelectionView.kt
Added logic to dynamically retrieve the logos on the list.


johnnyluo (approved):
LGTM

PR #65, opened 2024-05-09
Add Gson instance to dependency graph

YGolybev:
Added a Gson instance to the dependency graph and replaced usage with it in some places, but we still have a lot more of them.


johnnyluo (approved):
LGTM

PR #67, opened 2024-05-09
Add Timber for logging

### LINKED ISSUE #58: Add Timber
We can use https://github.com/JakeWharton/timber as the primary logging method for our app.  

It has a slightly smaller API:
Current:
`Log.d("ClassNameOrTag", "Message")`
With Timber:
`Timber.d("Message")`

With its `Tree` system, we can log anything that goes into Timber anywhere, not only to a Logcat. Like a crash reporting service, [as in Timber's sample app](https://github.com/JakeWharton/timber/blob/e6db346a2934a55bb940d0a5d293a931f54de066/timber-sample/src/main/java/com/example/timber/ExampleApp.java#L24).

YGolybev:
resolves #58

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -40,6 +40,7 @@ android {
     }
     buildFeatures {
         compose = true
+        buildConfig = true
     }
     composeOptions {
         kotlinCompilerExtensionVersion = "1.5.12"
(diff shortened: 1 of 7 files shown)


johnnyluo (approved):
LGTM

PR #72, opened 2024-05-09
Add coin screen

aminsato:


app/src/main/java/com/vultisig/wallet/presenter/chain_coin/ChainCoinViewmodel.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.vultisig.wallet.ui.navigation.Screen
+import dagger.hilt.android.lifecycle.HiltViewModel
+import java.math.BigDecimal
+import java.math.RoundingMode
+import javax.inject.Inject
+
+@HiltViewModel
+class ChainCoinViewmodel @Inject constructor(savedStateHandle: SavedStateHandle) : ViewModel() {
+    private val gson = Gson()
+    private val accountJson: String = savedStateHandle.get<String>(Screen.CHAIN_COIN_ACCOUNT_JSON)!!
```
johnnyluo, line 19: I don't think this is the correct way to pass vault , instead , we should just pass the vault's pubkeyECDSA , and also the selected chain. 
1. Vault can be retrieved from VaultDB , all the coins has been saved in vault
2. Chain name , which represent which chain current view is for.

johnnyluo (approved):

PR #80, opened 2024-05-10
Add navigation between chain account view, tokens selection and send screens

YGolybev:
Added `Navigator`, so we can navigate from `ViewModel`, use as `Navigator<Destination> `


johnnyluo (approved):
LGTM

PR #84, opened 2024-05-10
Add vault detail screen

aminsato:


app/src/main/java/com/vultisig/wallet/presenter/vault_setting/vault_detail/VaultDetailViewmodel.kt
```diff
@@ -0,0 +1,36 @@
...
+
+    val uiModel = MutableStateFlow(VaultDetailUiModel())
+    fun loadData() {
+        vault?.let {
+            uiModel.update {
+                it.copy(
+                    name = vault.name,
+                    pubKeyECDSA = vault.pubKeyECDSA,
+                    pubKeyEDDSA = vault.pubKeyEDDSA,
+                    deviceList = listOf("SM-G990E-243","iPad-04C (this device)")
```
johnnyluo, line 31: Please don't hardcode the devices here.  We should get from `vault.signers`

johnnyluo (approved):
LGTM

PR #85, opened 2024-05-11
Save Vault's keygen committee

johnnyluo:
1.This PR is to save the keygen committee to vault , 
2. replace a lot deprecated properties
3. Replace Log.d to Timber.d


app/src/main/java/com/vultisig/wallet/ui/models/ChainSelectionViewModel.kt
```diff
@@ -38,12 +45,49 @@ internal class ChainSelectionViewModel @Inject constructor(
...
-            "No vault with $vaultId"
-        }.let { vault ->
-            vault.copy(coins = vault.coins + coin)
-        })
+        viewModelScope.launch {
+            commitVault(checkNotNull(vaultDb.select(vaultId)) {
+                "No vault with $vaultId"
+            }.let { vault ->
+                // When user add a coin to the vault , we need to derive the public key and address
+                // and save the address and public key with the coin , thus don't need to derive in the future
```
YGolybev, line 55: We likely want to store the account's address in a separate entity (if we want to store it at all). 
As I explained above, a Coin (a representation of a token) and an address are entirely different entities. For example, we can have one token presented on a few chains, thus having different addresses. However, the addresses for different tokens on one chain are the same.
(Again, Coin.TssKeysignType, for example, is based only on Chain right now, so it's most likely not a coins property at all)  

So, we're probably better off using chainAccountAddressRepository for querying the address where it's required and caching it inside of the repository based on a public key and Chain (this will at least save us the pain of dealing with cache inconsistencies, which will appear inevitably (with JSON and files as storage, and little to no support for concurrent updates, that's for sure)).

app/src/main/java/com/vultisig/wallet/data/api/ThorChainApi.kt
```diff
@@ -2,18 +2,25 @@ package com.vultisig.wallet.data.api
...
+import io.ktor.client.request.post
+import io.ktor.client.request.setBody
 import io.ktor.client.statement.bodyAsText
+import io.ktor.http.ContentType
+import io.ktor.http.contentType
+import timber.log.Timber
 import javax.inject.Inject
 
-internal interface ThorChainApi {
+interface ThorChainApi {
```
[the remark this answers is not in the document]
johnnyluo: Yup, I could do that

app/src/main/java/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -30,30 +31,67 @@ internal interface AccountsRepository {
...
+                }
+            }
+        }
+        vaultDb.upsert(updateVault)
+        return updateVault
+    }
 
+    override fun loadAccounts(vaultId: String): Flow<List<Account>> = flow {
+        var vault = getVault(vaultId)
+        vault = setDefaultChains(vault)
```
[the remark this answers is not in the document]
johnnyluo: `setDefaultChains` will only apply the logic when `vault.coins` is empty , which effectively means it will only add default when the vault just created.  Or when user remove all the coins from the vault. 

I think it is ok to stay here ,  or I can move it to `GenertingKeyViewModel` ,when we save the vault for the first time

app/src/main/java/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -30,30 +31,67 @@ internal interface AccountsRepository {
...
         val accounts = coins.mapTo(mutableListOf()) { coin ->
             val chain = coin.chain
-            val coinType = chain.coinType
-            val address = chainAccountAddressRepository.getAddress(
-                coinType,
-                PublicKeyHelper.getPublicKey(vault.pubKeyECDSA, vault.hexChainCode, coinType)
-            )
             coinToChainAccountMapper.map(
-                ChainAddressValue(coin, chain, address, null)
+                ChainAddressValue(coin, chain, coin.address, null)
```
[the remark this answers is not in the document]
johnnyluo: I agree that `address` is not really a property of coin , but however we only save the coins along with vault, and nothing else is saved , so I think we should keep the address property on coin.

app/src/main/java/com/vultisig/wallet/ui/models/ChainSelectionViewModel.kt
johnnyluo: I think I can move `TssKeysignType` to Chain ,  which is more appropriate. But I would really like to save the address , and derived public key in hex format , so we only need to calculated it once when user add the coin to their vault.

I don't like use cache for this , because cache will require us to recalculate it after app restart.

YGolybev: Yeah, we can save the address and not need to recalculate. We can still do that through a chainAccountAddressRepository. I don't mean Guava's google.common.Cache when I say about caching.

We can approach it like this: 
Let's leave the address calculation for Coin here (at the moment of adding it to the vault) for now, but let's use a repository to query for the address where it's needed. Later, I will add persistent caching in the repository.

(The only problem we will have right now is with default chains cause they're never "added" to the vault per se. (They're added in a different place, which implies that we need to retrieve the address and save it to the vault in two places)) 

Also, AFAIK, there can be multiple addresses for one public key (at least it appears to be so in other wallets (?)). Is that right, and if so, how would we query multiple addresses in such a case?

johnnyluo: Usually , each blockchain will have a unique public key that is derived from the root public key. 
For example , the public key used by bitcoin , it is derived from the root public key using `m/84'/0'/0'/0/0` 
While Ethereum's public key will be different , because it is derived using `m/44'/60'/0'/0/0`

But... some blockchains share the same derive path , for example , all cosmos blockchains using the same derive path , so effective their public key will be the same.   But their address will be different , because it is using bech32 , each blockchain has unique prefix.

YGolybev: Ok, thanks!

YGolybev (approved):

PR #86, opened 2024-05-12
Add A few blockchain support

johnnyluo:
This PR add the following 
1. BlockChair to fetch UTXO chains's balance
2. EvmRpc to fetch EVM chain's balance
3. Add EvmChain support(ETH/BSC/AVAX), and Layer2 (Orbitrum,optimism,blast,base, coronos)
4. Add mayaChain


app/src/main/java/com/vultisig/wallet/presenter/keysign/Keysign.kt
```diff
@@ -164,10 +167,27 @@ fun PreviewKeysign() {
...
+                    return BigInteger.ZERO
+                }
+
+                override suspend fun broadcastTransaction(
+                    coin: Coin,
+                    signedTransaction: String,
+                ): String {
+                    return ""
+                }
+            }
```
YGolybev, line 190: You can make your ViewModel class a HiltViewModel and set it up as a default parameter for this Composable as:
`viewModel: KeysignViewModel = hiltViewModel()`

This will avoid using mocks here, and even instantiating KeysignViewModel

app/src/main/java/com/vultisig/wallet/ui/models/ChainSelectionViewModel.kt
```diff
@@ -60,16 +62,31 @@ internal class ChainSelectionViewModel @Inject constructor(
...
+                                    PublicKeyType.SECP256K1
+                                )
+                            )
+                            vault.copy(
+                                coins = vault.coins + coin.copy(
+                                    address = address,
+                                    hexPublicKey = derivedPublicKey
+                                )
+                            )
+                        }
```
YGolybev, line 89: You can extend the `chainAccountAddressRepository::getAddress` method, and add MayaChain support there, so we can use one method for querying addresses everywhere (or at least a single entity) and avoid duplication of that code in TokenSelectionViewModel

app/src/main/java/com/vultisig/wallet/ui/models/TokenSelectionViewModel.kt
```diff
@@ -44,12 +53,61 @@ internal class TokenSelectionViewModel @Inject constructor(
...
+                            PublicKey(vault.pubKeyEDDSA.hexToByteArray(), PublicKeyType.ED25519)
+                        )
+                        vault.copy(
+                            coins = vault.coins + coin.copy(
+                                address = address,
+                                hexPublicKey = vault.pubKeyEDDSA
+                            )
+                        )
+                    }
+                }
```
YGolybev, line 108: Same with TssKeyType, it can probably be supported in a chainAccountAddressRepository::getAddress method, as the difference in the code between the EDDSA and ECDSA is only input parameters (pubKeyEDDSA vs. pubKeyECDSA, PublicKeyType.ED25519 vs. PublicKeyType.SECP256K1)

YGolybev (commented):
Great!

app/src/main/java/com/vultisig/wallet/presenter/keysign/Keysign.kt
johnnyluo: a lot of information needed to create KeysignViewModel is passed down from it's parent , I am not sure how to make it using DI yet ,  
``` 
private val vault: Vault,
    private val keysignCommittee: List<String>,
    private val serverAddress: String,
    private val sessionId: String,
    private val encryptionKeyHex: String,
    private val messagesToSign: List<String>,
    private val keyType: TssKeyType,
    private val keysignPayload: KeysignPayload,
```
All these information are decided in KeysignFlowViewModel or JoinKeysignViewModel

YGolybev: Good! It's probably SavedStateHandle :)

YGolybev (approved):

Issue #76, opened 2024-05-10
Add send screen


YGolybev:
<img width="160" alt="image" src="https://github.com/vultisig/vultisig-android/assets/14962060/5ce12ae1-8a9a-4750-98bf-ab6f6b5b0679">

PR #95, opened 2024-05-13
Added error dialog when transaction info is incorrect on SendScreen

YGolybev:
Added error dialog when transaction info is incorrect on SendScreen
Fixed token logo on all screens
Added checkmark for a selected token on SendScreen
Made max sendable token value a difference between balance and gas fees for native tokens

Progresses #76


johnnyluo (approved):
LGTM

PR #101, opened 2024-05-14
Add scanning QR code for address on SendScreen

### LINKED ISSUE #42: Unwanted Screen Rotation During QR Code Scanning
### Description
While on the QR code scanning screen, the phone is consistently rotating, disrupting the scanning process. This rotation is unnecessary and can lead to user frustration and difficulty in capturing QR codes efficiently.

### Steps to Reproduce
1. Navigate to the QR code scanning screen within the app.
2. Observe the screen rotation behavior as the device attempts to scan a QR code.
3. Note that the screen rotation occurs even when it is not desired or necessary for the scanning process.

### Expected Behavior
The screen should remain fixed in its orientation during QR code scanning, regardless of the device's position, to ensure a smooth and uninterrupted scanning experience for users.

YGolybev:
Progresses #76 
Adds new QR code screen to be used to scan to fix #42

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -79,6 +79,11 @@ dependencies {
     debugImplementation(libs.androidx.ui.tooling)
     debugImplementation(libs.androidx.ui.test.manifest)
 
+    // camera
+    implementation(libs.androidx.camera.camera2)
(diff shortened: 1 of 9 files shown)


johnnyluo (approved):
LGTM

PR #107, opened 2024-05-15
Add Reshare

johnnyluo:


app/src/main/java/com/vultisig/wallet/ui/navigation/NavGraph.kt
```diff
@@ -91,11 +91,23 @@ internal fun SetupNavGraph(
...
+            arguments = listOf(navArgument(Screen.KeygenFlow.ARG_VAULT_NAME) {
+                type = NavType.StringType
+                defaultValue = Screen.KeygenFlow.DEFAULT_NEW_VAULT
+            })
+        ) { navBackStackEntry ->
             val vaultDB = VaultDB(context)
             val allVaults = vaultDB.selectAll()
-            // TODO: later on will need to deal with reshare
-            KeygenFlowView(navController, Vault("New Vault ${allVaults.size + 1}"))
+            val vaultId = navBackStackEntry.arguments?.getString("vault_name") ?: ""
```
YGolybev, line 103: Probably can reuse constant for argument name

```suggestion
            val vaultId = navBackStackEntry.arguments?.getString(Screen.KeygenFlow.ARG_VAULT_NAME) ?: ""
```

app/src/main/java/com/vultisig/wallet/ui/navigation/NavGraph.kt
```diff
@@ -91,11 +91,23 @@ internal fun SetupNavGraph(
...
+            val vault = if (vaultId == Screen.KeygenFlow.DEFAULT_NEW_VAULT) {
+                Vault("New Vault ${allVaults.size + 1}")
```
YGolybev, line 105: This is somewhat prone to collisions. 

Let's say you create three vaults: New Vault 1, New Vault 2, and New Vault 3. 
Then, you delete New Vault 2 and create a new vault. 
Because of this logic, it will be named New Vault 3 (after deletion of New Vault 2, allVaults.size == 2), although there already is a vault with this name, and because we make select queries based on Vault.name, this may produce bugs.

I suggest adding a separate `id` field with a unique identifier (maybe generated by UUID.randomUUID().toString()), for identifying and querying vaults.

johnnyluo: I added a logic to make sure the vault name is unique for now , We could use vault's public key ecdsa as the key for storage though , but that can be a seperate PR

YGolybev (approved):

PR #108, opened 2024-05-15
Add overscroll hacks

YGolybev:
### **DONT MERGE**

_**TL;DR:**_ 
In Jetpack Compose, the overscroll effect is disabled when the content size is less than the container size.
We probably should live with how it works in Compose because otherwise, it gets complicated.

And now to the complicated part:
So, a long time ago (in version 1.2.0 of Compose), the iOS-like overscroll effect could be enabled on content smaller than a container in about one flag change. [It was `OverScrollConfiguration::forceShowAlways`](https://androidx.tech/artifacts/compose.foundation/foundation/1.2.0-beta03-source/androidx/compose/foundation/gestures/OverScrollConfiguration.kt.html).

But about two years ago, the scroll system in Compose was rewritten, and that flag was gone.

The current state of affairs is as follows:
[`OverscrollConfiguration`](https://androidx.tech/artifacts/compose.foundation/foundation/1.4.3-source/androidx/compose/foundation/OverscrollConfiguration.kt.html) does not contain the `forceShowAlways` flag anymore.

There's a new system to define overscroll effects, but it's hardly customizable.
LazyList and its children (LazyColumn, LazyRow) have a built-in default overscroll effect and are not customizable unless you completely rewrite LazyList.
It can't be overridden because it always uses `ScrollableDefaults.overscrollEffect()` inside itself.

Modifiers `verticalScroll`, `horizontalScroll` and others also use built-in and not overridable overscrollEffects. 

Moreover, the scroll itself is now considered disabled by ScrollState when the content is smaller than the container. 
And it could be as simple as overriding ScrollState class... but it's defined as final and can't be overridden :(

So basically, here I do two possible hacks for that:

OverscrollHack1: 

Add content to a scrollable container, making it larger than the container (by 1 dp).

OverscrollHack2:

Completely tear out the guts of Compose scroll and override checks for content size so scrolling is possible when content is smaller than a container.

They both work. But they're both problematic. 
The first is problematic because it won't really work for LazyList.
Seconds is problematic because it will also require us to tear out the guts of LazyList and write our own custom ones.

The overscroll effect will exist when the content is larger than the container, so I would advise living with that, sticking with how Compose handles scroll, and not using any of those hacks.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/components/Overscroll.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/Overscroll.kt
@@ -1,0 +1,1099 @@
+package com.vultisig.wallet.ui.components
+
+import android.content.Context
+import android.os.Build
+import android.util.AttributeSet
+import android.widget.EdgeEffect
+import androidx.annotation.DoNotInline
+import androidx.annotation.RequiresApi
+import androidx.annotation.VisibleForTesting
+import androidx.compose.animation.core.AnimationSpec
+import androidx.compose.animation.core.SpringSpec
+import androidx.compose.foundation.ExperimentalFoundationApi
+import androidx.compose.foundation.LocalOverscrollConfiguration
+import androidx.compose.foundation.MutatePriority
+import androidx.compose.foundation.OverscrollConfiguration
+import androidx.compose.foundation.OverscrollEffect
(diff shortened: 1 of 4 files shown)


johnnyluo:
Yup, this seems a bit complicated , let's park it

PR #111, opened 2024-05-16
Add address share screen

aminsato:


app/src/main/java/com/vultisig/wallet/presenter/qr_address/QrAddressScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+import androidx.navigation.NavHostController
+import com.vultisig.wallet.R
+import com.vultisig.wallet.presenter.common.rememberQRBitmapPainter
+import com.vultisig.wallet.ui.components.QrBorder
+import com.vultisig.wallet.ui.theme.Theme
+import com.vultisig.wallet.ui.theme.dimens
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun QrAddressScreen(navController: NavHostController) {
```
YGolybev, line 43: Please, let's keep everything internal where possible; it'll help us to modularize the application later

app/src/main/java/com/vultisig/wallet/ui/navigation/Navigation.kt
```diff
@@ -128,5 +128,11 @@ internal sealed class Destination(
     data object VultisigToken : Destination(route = "settings/vultisig_token")
     data object LanguageSetting : Destination(route = "settings/language")
     data object CurrencyUnitSetting : Destination(route = "settings/currency")
+    class QrAddressScreen(val address:String) : Destination(route = "vault_details/qr_address_screen/$address"){
```
YGolybev, line 131:
```suggestion
    data class QrAddressScreen(val address:String) : Destination(route = "vault_details/qr_address_screen/$address"){
```

app/src/main/java/com/vultisig/wallet/presenter/qr_address/QrAddressScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+            val screenWidth = LocalConfiguration.current.screenWidthDp
+            val address = viewmodel.address!!
+            val qrBoxSize = ((screenWidth * .8).coerceAtMost(300.0)).dp
+            val imageAndBorderSpace = 30.dp
+            val imageSizeInDp = qrBoxSize - 2 * imageAndBorderSpace
+            val whiteSpace = 10.dp
+
+            Text(
+                text = address,
+                style = Theme.menlo.titleMedium,
```
YGolybev, line 99: In Figma, and it seems that on iOS it's menlo.body1
```suggestion
                style = Theme.menlo.body1,
```

app/src/main/java/com/vultisig/wallet/presenter/qr_address/QrAddressScreen.kt
```diff
@@ -0,0 +1,135 @@
...
+                modifier = Modifier
+                    .align(Alignment.TopCenter)
+                    .padding(top = 128.dp)
+            )
+
+            Box(
+                Modifier
+                    .clip(shape = RoundedCornerShape(qrBoxSize.div(10)))
```
YGolybev, line 107: It's probably just fixed 10.dp size of corners

YGolybev (changes requested):

app/src/main/java/com/vultisig/wallet/ui/components/QrBorder.kt
```diff
@@ -0,0 +1,86 @@
...
+                relativeLineTo(halfNegativeSegment + halfBorderWidth, zero)
+                relativeMoveTo(negSegment, zero)
+                relativeLineTo(negSegment, zero)
+                relativeMoveTo(negSegment, zero)
+                relativeLineTo(halfNegativeSegment, zero)
+                arcTo(
+                    Rect(
+                        topLeft = Offset(halfBorderWidth, 4 * segment - halfBorderWidth),
+                        bottomRight = Offset(
+                            segment + halfBorderWidth, 5 * segment - halfBorderWidth
```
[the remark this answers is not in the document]
aminsato: I aimed to replicate the design from Figma to ensure consistency and fidelity in the output. 
Here is the current output
Screenshot 2024-05-17 at 12 51 12 PM

However, if we opt to use Stroke with PathEffect, we would have
Screenshot 2024-05-17 at 12 51 39 PM

app/src/main/java/com/vultisig/wallet/presenter/qr_address/QrAddressScreen.kt
YGolybev: And why not to fix size of the QR border at 10.dp too?

app/src/main/java/com/vultisig/wallet/ui/components/QrBorder.kt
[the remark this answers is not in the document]
YGolybev: Isn't that what [intervals](https://developer.android.com/reference/kotlin/androidx/compose/ui/graphics/PathEffect#dashPathEffect(kotlin.FloatArray,kotlin.Float)) are for?

Or is it impossible to use those, to replicate design in Figma?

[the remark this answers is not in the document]
aminsato: ok, then we would have sth like this
telegram-cloud-photo-size-4-5925036968746533315-y
We can not always make sure that corners are same as Figma unless we do some calculation on line width and barcode size and etc. 

I open to use Stroke with PathEffect  and intervals, let me know if it make works for you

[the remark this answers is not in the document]
johnnyluo: I think this is fine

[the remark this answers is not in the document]
YGolybev: Yeah, if PathEffect doesn't work, current solution is good

app/src/main/java/com/vultisig/wallet/presenter/qr_address/QrAddressScreen.kt
aminsato: Done

YGolybev (approved):

PR #112, opened 2024-05-16
Add blockchain specific info to transaction

YGolybev:


app/src/main/java/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
```diff
@@ -0,0 +1,96 @@
...
+            BlockChainSpecific.THORChain(
+                accountNumber = BigInteger(account.accountNumber),
+                sequence = BigInteger(account.sequence ?: "0"),
+            )
+        }
+
+        TokenStandard.EVM -> {
+            val evmApi = evmApiFactory.createEvmApi(chain)
+
+            val gasLimit = 23000 // TODO gas limit may differ for chains
```
johnnyluo, line 54: you can pass in coin , and coin.IsNativeToken , if it is true means it is ETH , otherwise it is ERC20 , so the gas limit should set accordingly

app/src/main/java/com/vultisig/wallet/models/cosmos/THORChainAccountValue.kt
```diff
@@ -4,7 +4,7 @@ import com.google.gson.annotations.SerializedName
 
 data class THORChainAccountValue(
     val address: String?,
-    @SerializedName("account_number") val accountNumber: String?,
+    @SerializedName("account_number") val accountNumber: String,
```
johnnyluo, line 7: account_number can be null , when an address is not exist on the blockchain yet , then accountNumber will be null

johnnyluo (approved):
LGTM

PR #115, opened 2024-05-16
Fixes

YGolybev:
Remove requirement for camera on device
Fixed UI when there's no camera permission
Set transaction fees names for Bitcoin as in iOS


johnnyluo (approved):
LGTM

PR #117, opened 2024-05-17
feat: vault import

aminsato:


app/src/main/java/com/vultisig/wallet/presenter/import_file/ImportFileViewModel.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    private fun insertContentToDb(fileContent: String?) {
+        if (fileContent == null)
+            return
+        val fromJson = gson.fromJson(fileContent, Vault::class.java)
+        viewModelScope.launch {
+            val saveFile = async(Dispatchers.IO) {
+                vaultDB.upsert(fromJson)
+            }
+            saveFile.join()
```
YGolybev, line 55: Why use async here?

app/src/main/java/com/vultisig/wallet/presenter/import_file/ImportFileScreen.kt
```diff
@@ -30,28 +35,59 @@ import androidx.compose.ui.geometry.CornerRadius
...
+        viewmodel.channel.collect { uiEvent ->
+            when (uiEvent) {
+                is CopyFileToAppDir -> {
+                    val fileContent = uiEvent.uri.fileContent(context)
+                    viewmodel.onEvent(FileContentFetched(fileContent))
+                }
+
+                is FetchFileName -> {
+                    val fileName = uiEvent.uri.fileName(context)
+                    viewmodel.onEvent(FileNameFetched(fileName))
```
YGolybev, line 80: I think you can inject ContentResolver from the application's context into ViewModel to avoid this back-and-forth

app/src/main/java/com/vultisig/wallet/presenter/import_file/ImportFileViewModel.kt
aminsato: I think async or launch both are valid.
I used it to do the creation file step in the io thread

app/src/main/java/com/vultisig/wallet/presenter/import_file/ImportFileScreen.kt
aminsato: While injecting the ContentResolver from the application's context into the ViewModel may seem like a solution to avoid back-and-forth communication, it goes against the recommended architecture guidelines for Android development. According to the official Android documentation on UI architecture (https://developer.android.com/topic/architecture/ui-layer#logic-types), UI logic, especially when it involves UI types like Context, should reside in the UI layer, not in the ViewModel. This approach ensures better separation of concerns, improves testability, and makes the codebase easier to maintain in the long run.

YGolybev: Well, first of all, it's not UI logic. 
You work with files here; it hardly has anything to do with UI.

Second, ViewModel is in the UI layer. That's how it is on the diagrams, just a bit above where your link leads.
image

Using ApplicationContext is all well and good in the layers under ViewModel. 
However, if you want to make it a really clean architecture, move injecting context to the Repository in the data layer. 

The current code just induces recompositions for working with files and complicates the UI logic.

app/src/main/java/com/vultisig/wallet/presenter/import_file/ImportFileViewModel.kt
aminsato: Done

YGolybev (approved):

PR #135, opened 2024-05-19
Change PNG color profile from sRGB to RGB

aminsato:
### Benefits
- Improved consistency in color rendering across different platforms and devices.
- Enhanced compatibility with systems that require RGB color profile.


johnnyluo (approved):
LGTM

PR #209, opened 2024-05-24
Use user default chains on vault generation

### LINKED ISSUE #163: Use users default chain settings
Currently, the Default Chains screen only sets up default chains for itself. We should add usage of this setting.

YGolybev:
Closes #163


johnnyluo (approved):
LGTM

PR #211, opened 2024-05-25
Refactor: Capitalization of Token Subtitles

aminsato:


app/src/main/java/com/vultisig/wallet/models/Coin.kt
```diff
@@ -47,7 +47,7 @@ object Coins {
             address = "",
             decimal = 8,
             hexPublicKey = "",
-            priceProviderID = "bitcoin-cash",
+            priceProviderID = "Bitcoin Cash",
```
YGolybev, line 50: we're using this field for accessing prices, isn't that breaking that?

YGolybev (changes requested):

app/src/main/java/com/vultisig/wallet/models/Coin.kt
aminsato: No, When I chanaged this field, the output changed from following
Screenshot 2024-05-25 at 10 51 01

to the following

Screenshot 2024-05-25 at 10 53 06

aminsato: Fixed, please check again

YGolybev (approved):

PR #212, opened 2024-05-25
Add new drawables and update drawable refrences

aminsato:
This commit includes the additions of new PNG assets for various chians and update corresponding part of codebase to these news assets


app/src/main/java/com/vultisig/wallet/models/Chain.kt
```diff
@@ -125,11 +125,11 @@ val Chain.logo: Int
         Chain.solana -> R.drawable.solana
         Chain.ethereum -> R.drawable.ethereum
         Chain.avalanche -> R.drawable.avax
-        Chain.base -> R.drawable.eth_base
-        Chain.blast -> R.drawable.eth_blast
-        Chain.arbitrum -> R.drawable.eth_arbitrum
-        Chain.polygon -> R.drawable.eth_polygon
-        Chain.optimism -> R.drawable.eth_optimism
```
YGolybev, line 132: Are old resources still in use after this PR? If not they should probably be deleted

YGolybev (commented):
Please, upload assets using Webp where possible

app/src/main/java/com/vultisig/wallet/models/Chain.kt
aminsato: Yes
like https://github.com/vultisig/vultisig-android/blob/d42af0710fa2fc618ee75656cf260add3aacaf9e/app/src/main/java/com/vultisig/wallet/models/Coin.kt#L288

aminsato:
> Please, upload assets using Webp where possible

aminsato:
> Please, upload assets using Webp where possible

Can I upload .xml instead of Webp?

YGolybev (commented):
If XML seems like a better choice for some assets. But I think that all of these assets will be problematic for XML.

aminsato:
> If XML seems like a better choice for some assets. But I think that all of these assets will be problematic for XML.

Done

YGolybev (approved):

PR #222, opened 2024-05-26
Update isNativeToken flag

aminsato:
Update the  `isNativeToken` flag from `true` to `false`  for ticker are not native to the respective chain


aminsato:
It ensures that non-native tokens are not displayed in the Choose Chains view.

johnnyluo:
There are actually native token to pay for the gas

aminsato:
> There are actually native token to pay for the gas

But some non-native tokens are also displayed in the VaultAccountsScreen view.The porpuse of my PR is to address to this issue.
image

johnnyluo:
only WETH need to be set to IsNativeToken=false , ETH should be fine

PR #232, opened 2024-05-27
Add users language settings

### LINKED ISSUE #164: Use users language settings
Currently, the Language screen only sets up language for itself. We should add usage of this setting.

aminsato:
Fix #164 

We also need to update strings for each supported languages


app/src/main/res/values-es/strings.xml
```diff
@@ -0,0 +1,186 @@
...
+    <string name="signing_error_please_try_again_s">Espanol_Signing Error. Please try again. %s</string>
+    <string name="bottom_warning_msg_keygen_error_screen">Espanol_Keep devices on the same WiFi Network,\n correct vault and pair devices.\n Make sure no other devices are\n running Vultisig.</string>
+    <string name="try_again">Espanol_Try again</string>
+    <string name="vault_choose_chains">Espanol_Choose Chains</string>
+    <string name="plus_icon">Espanol_Plus Icon</string>
+    <string name="vault_settings_title">Espanol_Edit Vault</string>
+    <string name="vault_settings_details_title">Espanol_Details</string>
+    <string name="vault_settings_details_subtitle">Espanol_View your vault details</string>
+    <string name="vault_settings_backup_title">Espanol_Backup</string>
+    <string name="vault_settings_backup_subtitle">Espanol_Backup your vault share</string>
```
YGolybev, line 20: Please don't add translations that are not filled with correct data. Imagine how confused a user would be.

YGolybev (changes requested):

app/src/main/res/values-es/strings.xml
aminsato: Done

YGolybev (approved):

PR #252, opened 2024-05-28
Fix infinite progress indicator for ptr if error is emitted

### LINKED ISSUE #227: Refreshing vault spawns an never ending loading icon on top
User reports:
Refreshing my vault seems to spawn an never ending loading icon on top

YGolybev:
Fixes #227


johnnyluo (approved):
LGTM

PR #273, opened 2024-05-30
Fix/crash on duplicate vault

aminsato:
Handle the scenario where importing a duplicate vault causes a crash by displaying a SnackBar message to inform the user about the duplicate vault


app/src/main/java/com/vultisig/wallet/presenter/import_file/ImportFileScreen.kt
```diff
@@ -181,19 +218,10 @@ private fun ImportFileScreen(
...
-                minHeight = 44.dp,
-                modifier = Modifier
-                    .fillMaxWidth()
-                    .padding(
-                        vertical = 16.dp,
-                    ),
-                onClick = onContinue,
-            )
-        }
+        }/*re*/
```
YGolybev, line 221: ?

YGolybev (approved):

PR #284, opened 2024-05-31
Add send & swap buttons to home screen

### LINKED ISSUE #196: Updating UI on main page
1. Please update the UI of main page to be concruent with figma design so we only have one settings button top left and one re-arrange vault on the top right. 
2. Add Send and Receive Buttons to be cuncruent with figma
3. Ammend settings page so it is also congruent.

image

YGolybev:
Fixes #196


johnnyluo (approved):
LGTM

PR #300, opened 2024-06-04
Cache token balance and prices in Room

### LINKED ISSUE #245: Cache price and balance of tokens
Cache price and balance of tokens to database, so user can see their data when app is offline

yvebe:
Fixes #245


johnnyluo (approved):
LGTM

PR #373, opened 2024-06-12
Update peer discovery ui

aminsato:


realpaaao:
@aminsato sorry to have another change. but we finally have a final design in figma. Please update it again and sorry for the extra work!

image

aminsato:
@realpaaao No problem. Also, please note that in the current app design, the parts between the keygen and sign screens (for sending and swapping) are the same, but they are different in Figma. Do you want them to be the same, or should I separate them?

Screenshot 2024-06-13 at 9 48 12 AM

realpaaao:
@aminsato give me a minute, will amend.

PR #430, opened 2024-06-21
Add affiliate fees for swaps with value more than 100 USD

### LINKED ISSUE #423: ADD: Improve affiliate logic to account different currencies
As the affiliate logic only checks if the swap amount is below 100 `dollar` to not apply the fee, it should be implemented that if a different currency is used the 100 dollars get converted to the used currency.

yvebe:
Closes #423

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -162,7 +162,8 @@ internal class SwapFormViewModel @Inject constructor(
                         srcToken, srcTokenValue, AppCurrency.USD,
                     )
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -429,10 +436,17 @@ internal class SwapFormViewModel @Inject constructor(
                                 )
                             }
                         } else {
+                            val srcUsdFiatValue = convertTokenValueToFiat(
+                                srcToken, tokenValue, AppCurrency.USD,
+                            )
+
+                            val isAffiliate = srcUsdFiatValue.value > 100.toBigDecimal()
```
aminsato, line 443: Should not we add affiliate to $100 tx?  (Change > to >= )

aminsato (approved):

Issue #265, opened 2024-05-29
BUG: Crash with pressing burger view repeatedly


realpaaao:
**Describe the bug**
When pressing the burger view in top left repeatedly crashes the app or unable to exit from settings.

**To Reproduce**
1. main page
2. press burger view repeatedly
3. See error

**Screenshots**

https://github.com/vultisig/vultisig-android/assets/167348323/d0b76f3d-5c71-4919-b6d4-d3f34e7b5b46

realpaaao:
@aminsato issue can still be repoduced. Either crashes like before or back button will stop working or random views pop up, If deleted a vault before, that vault re-appears.

aminsato:
> @aminsato issue can still be repoduced. Either crashes like before or back button will stop working or random views pop up, If deleted a vault before, that vault re-appears.

@realpaaao Please update to the latest version and let me know if you continue to experience the problem.

PR #437, opened 2024-06-23
Improve Debounce Click Handling

aminsato:
Fixes https://github.com/vultisig/vultisig-android/issues/265#issuecomment-2178010949
This PR introduces an enhancement to the debounce click handling mechanism in our application. Previously, the cooldown period was not shared between different views, which could lead to inconsistent behavior when rapidly switching between views and clicking on different elements.

  **Changes include:**  

- Refactoring the clickOnce function in Onclick.kt to ensure that the cooldown period is shared across all views. This prevents rapid, successive clicks from being registered in different views during the cooldown period.
- Adjusting the COOL_DOWN_PERIOD constant to fine-tune the length of the cooldown period based on user interaction testing.
- This change improves the user experience by ensuring consistent click handling across the application, regardless of how quickly a user may switch between views or click on different elements.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/presenter/import_file/ImportFileScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/presenter/import_file/ImportFileScreen.kt
@@ -123,7 +123,7 @@ private fun ImportFileScreen(
                 .padding(
                     vertical = 16.dp,
                 ),
-            onClick = ClickOnce(onContinue),
+            onClick = onContinue,
         )
     }) {
 Box(Modifier.padding(it)) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/vultisig/wallet/presenter/import_file/ImportFileScreen.kt
```diff
@@ -123,7 +123,7 @@ private fun ImportFileScreen(
                 .padding(
                     vertical = 16.dp,
                 ),
-            onClick = ClickOnce(onContinue),
```
aminsato, line 126: The reason for the change is that the MultiColorButton(its parent) handles ClickOnce.

app/src/main/java/com/vultisig/wallet/ui/screens/vault_settings/components/ConfirmDeleteScreen.kt
```diff
@@ -120,9 +120,7 @@ internal fun ConfirmDeleteScreen(
                         bottom = 16.dp,
                     ),
                 text = stringResource(R.string.confirm_delete_delete_vault),
-                onClick = ClickOnce {
-                    onConfirmClick()
-                },
```
aminsato, line 125: The reason for the change is that the MultiColorButton(its parent) handles ClickOnce.

PR #440, opened 2024-06-24
Sort chains by assets fiat value

### LINKED ISSUE #375: ADD: logic that Assets get ordered after amount
Please add a logic that the assets get ordered after the dollar amount from high to low. With this logic we can remove the chain reordering as the user will always see the needed balances first.
Also apply this to Token screens but not override #335

yvebe:
Closes #375


johnnyluo (approved):
LGTM

PR #477, opened 2024-06-28
Fix keyboard not reappearing after being hidden

### LINKED ISSUE #372: BUG: Not possible to change input after set in swap screen
**Describe the bug**
After first input in `from` Asset on swap screen leaving the text field and wanting to amend the amount the keyboard does not pop up again. It is needed to change screens to be able to do amend.

**To Reproduce**
1. Go to `swap view`
2. Input amount for `from` asset
3. tap out of text field
4. try to amend the amount

**Expected behavior**
Keyboard pops up again

-Version 5

yvebe:
Fix #372


johnnyluo (approved):
LGTM

PR #491, opened 2024-07-01
Enforce hardcoded default chains when default chains are empty

### LINKED ISSUE #468: ADD: hardcode default chains and add more customization
Please make sure the hardcoded default chains are enforced and add the possibility that users can override these default settings with the chains of choice. Extend the selection of possible chains to all supported ones.
If user deselects all default chains, enforce the hardcoded chains.

Use [Figma](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=6892-6496&t=xfnbvFHmouveBm9k-4) as reference.

yvebe:
Fix #468


johnnyluo (approved):
LGTM

PR #505, opened 2024-07-02
Move token selection to a separate screen for send and swap

### LINKED ISSUE #397: ADD: Token search on send/swap/Chain Detail
Add Token Search according to figma design to send/swap/Chain Detail.

image

yvebe:
Fix #397


johnnyluo (approved):
LGTM

PR #513, opened 2024-07-03
Merge approve and swap in one transaction

### LINKED ISSUE #445: Add: Combine Approval and Swap , two transactions in one signing process
Currently , approval and swap are two transactions signed in two keysign process.  Please update it to sign approval and swap these two transactions in one single keysign process

The same as iOS does

yvebe:
Fix #445


app/src/main/java/com/vultisig/wallet/presenter/keysign/KeysignViewModel.kt
```diff
@@ -228,30 +229,39 @@ internal class KeysignViewModel(
...
+        if (keysignPayload.approvePayload != null) {
+            val signedTransaction = THORChainSwaps(vault.pubKeyECDSA, vault.hexChainCode)
+                .getSignedApproveTransaction(
+                    keysignPayload.approvePayload,
+                    keysignPayload,
+                    signatures
+                )
+
+            val evmApi = evmApiFactory.createEvmApi(keysignPayload.coin.chain)
+            evmApi.sendTransaction(signedTransaction.rawTransaction)
```
johnnyluo, line 246: Here , we might want to display both the approval transaction hash and send transaction hash in future , probable will need to return it , so we can display it on UI part

johnnyluo (approved):
LGTM

PR #537, opened 2024-07-06
Avoid swap choose a none swappable token to crash the app

johnnyluo:
@yvebe this is a bandaid solution , but at least this will stop the crash no matter what
can you please raise a PR to filter out those can't be swapped token on token selection screen?


aminsato (approved):
LGTM

PR #615, opened 2024-07-23
Add address book filtering by chain

### LINKED ISSUE #598: BUG: Address book , show up address that match current select chain
**Describe the bug**
At the moment , when we open AddressBook from Send View , it looks like 
image
Regardless which blockchain / tokens we are using right now , please filter address based on current token's chain 
For example , when I am sending ETH/ERC20 , please only show address belongs to ETH chain.
When I am sending BTC ,please only show address belongs to BTC chain etc.

yvebe:
Closes #598


johnnyluo (approved):
LGTM

Issue #486, opened 2024-06-30
Add auto token discovery


yvebe:
Discover tokens with balance in vault automatically

PR #631, opened 2024-07-26
Add auto token discovery

yvebe:
Progresses #486


app/src/main/java/com/vultisig/wallet/data/usecases/SaveVaultUseCase.kt
```diff
@@ -53,6 +55,28 @@ internal class SaveVaultUseCaseImpl @Inject constructor(
                     vaultRepository.addTokenToVault(vaultId, updatedCoin)
                 }
         }
+
+        // discover tokens with balance
+        Chain.entries
```
johnnyluo, line 60: we discussed , only do auto discover the chains that is in the vault

johnnyluo (approved):
LGTM

Issue #623, opened 2024-07-24
BUG: Address Book , add validation


johnnyluo:
**Describe the bug**
Currently add address , it doesn't have any validation on address ,  I can add whatever
<img width="358" alt="image" src="https://github.com/user-attachments/assets/3036262d-9a36-4810-a334-2ca7c9a601c0">

** Expected Result **
Add some validation , make sure the address is correct for the chosen blockchain

realpaaao:
Included in #650

PR #665, opened 2024-07-31
Validate address when adding address book entry

### LINKED ISSUE #650: IMPROVE: Address book feature
The address book has many small missing logic improvements:

- [ ] When no address is saved, please present the `Add address` Button directly on the screen
- [ ] When selecting a chain to save, please close the dropdown automatically after selecting a chain
- [ ] On send screen, improve that the complete address book list is visible but only the correct chains can be inserted
- [ ] Rearranging saved addresses not possible
- [ ] please include address validity check on creation of address. It is possible to create an address for SOL with a BTC address for example

yvebe:
Closes #650, #623


johnnyluo (approved):
LGTM

PR #812, opened 2024-08-23
Fix Lint Usability Warnings 

aminsato:


aminsato:
> do not use nodpi folder, it's not for that. use drawable/

**Should I suppress theses warnings?**
image

aminsato:
> > do not use nodpi folder, it's not for that. use drawable/
> 
> **Should I suppress theses warnings?** image

To resolve the warning, I discovered that it is recommended to place webp files in the nodpi folder.
@yvebe

PR #873, opened 2024-08-29
Move models to a data module

yvebe:
Refactor & test encryption and hex decoding
Move Coin & Vault into data module
Move part of models into data module


yevhen1sec (approved):
awesome job💪

PR #903, opened 2024-09-04
Fix vault duplicate detection on import

### LINKED ISSUE #880: BUG: Check Vault duplicate after password prompt
**Describe the bug**
At the moment the user does not get feedback if the vault is a duplicate.
Even when importing a vault share that has no password but already exists, prompts the user to put in a password.

**To Reproduce**
1. Create Vault
2. backup without password
3. import 

**Expected behavior**
The app verifies if the Vault already exists after asking for a password or not needing a password.

- v18

yvebe:
Fixes #880


johnnyluo (approved):
LGTM

PR #938, opened 2024-09-10
Set unbond sent amount to zero

yvebe:


yevhen1sec (approved):

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
```diff
@@ -364,7 +364,7 @@ internal class DepositFormViewModel @Inject constructor(
 
             memo = memo.toString(),
             srcTokenValue = TokenValue(
-                value = tokenAmountInt,
+                value = BigInteger.ZERO,
```
johnnyluo, line 367: set it to 0 is not going to work , we should set it to 1 or something

PR #949, opened 2024-09-12
Show token once in token selection screen

### LINKED ISSUE #781: IMPROVE: Remove Token from Token if Selected
When a Token got selected, move it to the Selected section and remove it from the Token List.
So it doesn't get shown two times.

image

yvebe:
Fix #781


johnnyluo (approved):
LGTM

PR #952, opened 2024-09-12
Move vault creation buttons to vault list screen

### LINKED ISSUE #918: [UPDATE] Vault view to include import
To remove a unnecessary screen when user already has existing vaults, please update the vault selection screen that it has both buttons to `add` or `import` a vault directly available.

image

With that we can get rid of the `Get started` screen when vaults already available.

yvebe:
Fix #918


johnnyluo (approved):
LGTM

PR #963, opened 2024-09-13
Add import button to keygen role screen

### LINKED ISSUE #960: [UPDATE] Create or Import Vault Screen
To get rid of one screen, we want to update the `create a vault or import a vault screen`.

Please only display this view on the first onboarding according to [figma](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=1123-11923&t=biwQdjGi8xPSNTNa-4) flow:
image
afterwards prompt this screen:
image

If user wants to create a second or further vaults let them go over `vault select `screen select it directly over the updated `setup screen `, (please also update buttons on this view) :
image

This will make it possible to get rid of one screen and improving user experience.

yvebe:
Fix #960


johnnyluo (approved):
LGTM

PR #983, opened 2024-09-16
Remove OkHttp and HttpUrlConnection usage in favor of ktor

yevhen1sec:


app/src/main/java/com/vultisig/wallet/ui/models/keygen/GeneratingKeyViewModel.kt
```diff
@@ -158,17 +158,17 @@ internal class GeneratingKeyViewModel(
...
-                this.keygenCommittee, gson = gson,
-            )
             withContext(Dispatchers.IO) {
-                keygenVerifier.markLocalPartyComplete()
-                if (!keygenVerifier.checkCompletedParties()) {
-                    throw Exception("another party failed to complete the keygen process")
+                keyApi.keygenMarkLocalPartyComplete(serverAddress, sessionId, listOf(vault.localPartyID))
+                Timber.d("Local party ${vault.localPartyID} marked as complete")
+                var repeate = true
+                while (repeate){
```
johnnyluo, line 165: Previous implementation will stop after 60 seconds , we need the same logic here as well, otherwise this will keep going forever

PR #998, opened 2024-09-18
Fix vault list last vault not visible behind buttons

### LINKED ISSUE #990: BUG: Last Vault on Vault selection not visible
**Describe the bug**
With adding import on Vault selection, the last vault in a long list is not visible.

-v22

yvebe:
Fix #990


johnnyluo (approved):
LGTM

PR #1007, opened 2024-09-20
Add reshare with VultiServer

### LINKED ISSUE #957: [UPDATE] Reshare flow to include Vultiserver
Currently the Reshare flow does not seperate between including Vultiserver or not.
Leading to that the server always joins. 

Please implement/update Reshare flow according to [figma](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=12051-8148&t=biwQdjGi8xPSNTNa-4) so user can have an optional Vultiserver on reshare.

Also include logic to check if the vault already has a vault share on the vultisig server or a new one needs to be created.

yvebe:
Fix #957


johnnyluo (approved):
LGTM

Issue #863, opened 2024-08-28
Replace Gson with kotlinx.serialization


yvebe:

aminsato:
Replace Gson with Json in BlockChairApi.kt

PR #1012, opened 2024-09-22
Replace Gson with Json in Server 

aminsato:
https://github.com/vultisig/vultisig-android/issues/863


aminsato:
> have you tested that with old vault shares?

Yes, I have tested it.

Issue #978, opened 2024-09-15
BUG: MAX send SOL not possible


realpaaao:
**Describe the bug**
hitting MAX amount to send SOL does not broadcast a transaction.

**Expected behavior**
Able to send MAX amount properly.

**Additional context**
Sending everything below `MAX` works.

PR #1036, opened 2024-09-25
Fix Unable to send SOLtransaction Broadcast

aminsato:
Fixes https://github.com/vultisig/vultisig-android/issues/978

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/models/SolanaJson.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/models/SolanaJson.kt
@@ -3,6 +3,7 @@ package com.vultisig.wallet.data.api.models
 import com.vultisig.wallet.data.utils.BigIntegerSerializer
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonObject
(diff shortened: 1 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/models/SolanaJson.kt
```diff
@@ -53,7 +54,8 @@ data class SolanaFeeObjectRespJson(
 @Serializable
 data class BroadcastTransactionRespJson(
     @SerialName("error")
-    val error: String?,
+    @Serializable
+    val error: JsonObject?,
```
aminsato, line 58: The purpose is to prevent exceptions due to deserializing the response when the error is not empty.

aminsato:
@realpaaao I tested Sol many times with maximum value and sometimes got error `Blockhash not found` and realized the best way is to change the RPC method `getLatestBlockhash` parameter from `finalized` to `confirmed` that let the user more time to sign a tx(https://solana.stackexchange.com/questions/6093/blockhash-not-found-when-sending-transaction). 

However it would be great if you could double check and see if the issue is still present or not

realpaaao:
@aminsato looks good. can't reproduce on main atm

PR #1037, opened 2024-09-25
Fix password encoding when passing between screens

yvebe:


app/src/main/java/com/vultisig/wallet/ui/navigation/Navigation.kt
```diff
@@ -311,8 +312,8 @@ internal sealed class Destination(
             vaultId,
             vaultName,
             vaultSetupType.raw,
-            email,
-            password
+            Uri.encode(email),
+            Uri.encode(password),
```
johnnyluo, line 316: do we need to decode it later, when send it to server?

PR #1060, opened 2024-09-28
Fix obsolete gradle dependencies to pass lint check check obsolete gradle

aminsato:


app/src/main/java/com/vultisig/wallet/ui/models/ChainTokensViewModel.kt
```diff
@@ -237,8 +242,13 @@ internal class ChainTokensViewModel @Inject constructor(
                         isBuyWeweVisible = chain == Chain.Base
                     )
                 }
-            }
+            }.onCompletion {
+                updateRefreshing(false)
```
aminsato, line 246: The reason for the change is to prevent multiple calls to uiState.update { it.copy(isRefreshing = false) } in a fraction of a second. This rapid calling can cause the loading state to disappear quickly before showing fiat prices, which may lead to bugs on some devices.

https://github.com/user-attachments/assets/66a16f43-a938-4245-ba04-bb0216d85bf6

PR #1099, opened 2024-10-06
Fix hex memo handling

aminsato:


johnnyluo:
Can you please descript what this PR fixes? is there an issue ?

Ehsan-saradar:
@johnnyluo 

I need this PR to test `VultiConnect` for signing a swap transaction. For the swap transaction, I will send a simple keysign payload with ABI binary data (without swap payload). Since it's in hex, we don't need to convert it to UTF-8 and then to bytes; we can directly convert hex to bytes.

If the memo is not in hex, then we can convert it to UTF-8 and obtain the bytes.

Ehsan-saradar:
@johnnyluo 

Example of swap using rango I did it yesterday with vulticonnect and 2 android phone based on this branch

https://arbiscan.io/tx/0x9df2b0fdac80d29b64df00e5d20bc7de159b036ad168050762be98396d7217fc

johnnyluo:
> @johnnyluo
> 
> Here’s an example of a swap using Rango that I completed yesterday with VultiConnect and two Android phones, based on this branch.
> 
> https://arbiscan.io/tx/0x9df2b0fdac80d29b64df00e5d20bc7de159b036ad168050762be98396d7217fc

Thanks

PR #1107, opened 2024-10-07
Check recycling before  recycle a Bitmap

aminsato:


app/src/main/java/com/vultisig/wallet/ui/models/ShareVaultQrViewModel.kt
```diff
@@ -158,7 +161,9 @@ internal class ShareVaultQrViewModel @Inject constructor(
     }
 
     internal fun saveShareQrBitmap(bitmap: Bitmap) {
-        shareQrBitmap.value?.recycle()
+        if (shareQrBitmap.value != null && shareQrBitmap.value?.isRecycled == true) {
```
johnnyluo, line 164: shouldn't this one be `isRecycled != true`?

app/src/main/java/com/vultisig/wallet/ui/models/QrAddressViewModel.kt
```diff
@@ -69,7 +69,9 @@ internal class QrAddressViewModel @Inject constructor(
         val qrBitmap = withContext(Dispatchers.IO) {
             makeQrCodeBitmapShareFormat(bitmap, color, logo, title, null)
         }
-        shareQrBitmap.value?.recycle()
+        if (shareQrBitmap.value != null && shareQrBitmap.value?.isRecycled != true) {
+            shareQrBitmap.value?.recycle()
```
johnnyluo, line 73: image

Based on the document , it seems we should not call it , and instead just let GC to do it's job

johnnyluo: recycle is just a request , doesn't mean it will be recycled immediately

PR #1111, opened 2024-10-07
Update GasFeeRepository.kt

realpaaao:


data/src/main/kotlin/com/vultisig/wallet/data/repositories/GasFeeRepository.kt
```diff
@@ -36,7 +36,7 @@ internal class GasFeeRepositoryImpl @Inject constructor(
         TokenStandard.EVM -> {
             val evmApi = evmApiFactory.createEvmApi(chain)
             TokenValue(
-                evmApi.getGasPrice().multiply(BigInteger("3")).divide(BigInteger("2")),
+                evmApi.getGasPrice().multiply(BigInteger("5")).divide(BigInteger("2")),
```
johnnyluo, line 39: We don't need to 2.5x EVM

PR #1121, opened 2024-10-09
Fix token details opening multiple times on fast clicks

### LINKED ISSUE #1078: [BUG] Possible to open Token detail two times
**Describe the bug**
When pressing a token in chain detail screen rapidly it is possible to open the token detail screen two times

**To Reproduce**
1. Go to chain detail screen
2. Fast click on token two times 
3. See error

**Expected behavior**
Rate limit that the screen gets opened only once

-v25

yvebe:
Fix #1078


johnnyluo (approved):
LGTM

PR #1127, opened 2024-10-09
Add airdrop registry

aminsato:
**Fixes https://github.com/vultisig/vultisig-android/issues/1108**


realpaaao:
@aminsato we updated figma a little yesterday so users can save the QR directly on the page, improving UX :

image

Issue #1094, opened 2024-10-06
[BUG] occasionally unable to send SOL


realpaaao:
**Describe the bug**
Recently when trying to send SOL, especially `max` amount, different error getting prompted and user is unable to send SOL.

**Expected behavior**
Consistent sending possible

**Screenshots**
![Image](https://github.com/user-attachments/assets/b60412fb-c73c-4534-ba7b-b5db75a3b50f)
![Image](https://github.com/user-attachments/assets/96d2ee31-a538-44d0-8ce4-ebc13bc03b05)
![Image](https://github.com/user-attachments/assets/4993c19f-0adb-4383-a913-6e4449005adc)

 - Version 25

**Additional context**
Often the pairing device also gets "too many RPC calls" error but tx goes through

johnnyluo:
This also should have been fixed by #1098 and #1036

realpaaao:
@johnnyluo @aminsato I am still getting this error on almost all tx when fast signing. It is not possible for me to fast sign

realpaaao:
QA done. But getting sending errors for SPL see #1205

PR #1128, opened 2024-10-09
Fix error `blockhash not found` when sending max SOL

aminsato:
https://github.com/vultisig/vultisig-android/issues/1094

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
@@ -134,7 +134,7 @@ internal class SolanaApiImp @Inject constructor(
                 params = buildJsonArray {
                     add(tx)
                     addJsonObject {
-                        put("maxRetries",3) }
+                        put("maxRetries",10) }
                 },
                 id = 1,
             )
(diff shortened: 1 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
```diff
@@ -133,6 +133,8 @@ internal class SolanaApiImp @Inject constructor(
                 method = "sendTransaction",
                 params = buildJsonArray {
                     add(tx)
+                    addJsonObject {
+                        put("maxRetries",10) }
```
johnnyluo, line 137: Why `blockhash not found` happen?  does add `maxRetries` fix the issue?

aminsato: On Solana, each transaction includes a `blockhash` (a reference to a recent block) to ensure that the transaction is not replayed in the future. This `blockhash` has a limited lifespan. If your transaction takes too long to be processed or broadcasted, the `blockhash` could expire, resulting in the "`blockhash not found`" error.

By adding `"maxRetries": 10`, we are instructing the RPC node to retry sending the transaction up to 10 times before giving up. This can help with cases where the `blockhash` expires before the transaction is successfully broadcasted. During these retries, the node may regenerate a valid `blockhash` and re-attempt sending the transaction, thus preventing the `"blockhash not found"` error.

Relying on public RPC nodes can lead to higher network congestion and latency, which increases the chances of encountering `blockhash not found `errors due to delays in processing transactions.

PR #1140, opened 2024-10-10
Remove coins from withdraw from pool deposit

yvebe:


data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt
```diff
@@ -121,7 +122,10 @@ class ThorChainHelper(
                 thorchainDepositMessage = Cosmos.Message.THORChainDeposit.newBuilder().apply {
                     this.signer = ByteString.copyFrom(fromAddress)
                     this.memo = memo
-                    this.addCoins(coin)
+                    // if it's withdraw from pool, we shouldn't add coins
```
johnnyluo, line 125: no ,we still need to add a coin , but keep the coin amount to be 0 , the previous logic is correct
https://discord.com/channels/838986635756044328/1252646142400004146/1293932452275949619

Issue #1092, opened 2024-10-05
[BUG] polkadot balance not loading


realpaaao:
**Describe the bug**
Polkadot balance is not loading leading to not be able to display cumulative balance

**Expected behavior**
Load Polkadot

**Screenshots**
![Image](https://github.com/user-attachments/assets/3fbdc19b-8723-477a-9bde-c8f1ea8ea7e8)

-v25

johnnyluo:
I can't reproduce it , it loads for me correctly

realpaaao:
QA done.

PR #1158, opened 2024-10-12
Fix polkadot balance loading

aminsato:
**Fixes https://github.com/vultisig/vultisig-android/issues/1092**


aminsato:
Sometimes the error is `SSL handshake aborted: ssl=0x733389bbc0: I/O error during system call, Connection reset by peer`, so we couldn't handle this error form our side , It's depend on user's ISP. 
I Suggest they use VPN to not get the error

PR #1231, opened 2024-10-22
fastvault/active vault enforce to use VultiRelay

johnnyluo:


johnnyluo:
as @yvebe pointed out , we don't save the preference anymore , so it has been enfoced

Issue #1205, opened 2024-10-18
[BUG] Unable to send SOL SPL tokens


realpaaao:
**Describe the bug**
Not possible to send out SPL tokens and getting` invalid account data for instruction` error

**To Reproduce**
1. Try sending any SPL token
3. See error

**Expected behavior**
Able to send SPL tokens

**Screenshots**
![Image](https://github.com/user-attachments/assets/2be0f36b-d4f2-4e76-b3fc-ecaccc486529)

-on `main` branch

**Additional context**

johnnyluo:
has this one bee fixed in latest version (v1.0.30)?  did you try it?

realpaaao:
@aminsato I am still unable to send SPL tokens on version 32. Please prioritize.

![Image](https://github.com/user-attachments/assets/90e72526-874d-4606-b8bf-1f4e7c678b5e)

realpaaao:
@aminsato I still can't send on `main` branch.

Either blockhash not found - where are we with the durable nonce?

Or getting a valid tx hash on vultisig but is not broadcasted on-chain

realpaaao:
Close for testing

PR #1241, opened 2024-10-25
Fix `Instruction data not found` error

aminsato:


johnnyluo:
Please attach link to the transaction that you sent here

data/src/main/kotlin/com/vultisig/wallet/data/models/payload/BlockChainSpecfic.kt
```diff
@@ -40,6 +40,7 @@ sealed class BlockChainSpecific {
         val priorityFee: BigInteger,
         val fromAddressPubKey: String? = null,
         val toAddressPubKey: String? = null,
+        val tokenAccountExists: Boolean? = null,
```
johnnyluo, line 43: This will not work , as this change is not in the protobuf , thus it doesn't get passed to pair device
and this should caues keysign failure , because they will not agree on what they are signing

johnnyluo:
@enriquesouza did we address the same issue on IOS?

data/src/main/kotlin/com/vultisig/wallet/data/models/payload/BlockChainSpecfic.kt
aminsato: Actually, I updated the "common data" submodule.

https://github.com/vultisig/commondata/pull/16
Screenshot_20241025_092434_Chrome

aminsato:
> Please attach link to the transaction that you sent here

https://github.com/vultisig/vultisig-android/issues/1205#issue-2597256575

data/src/main/kotlin/com/vultisig/wallet/data/models/payload/BlockChainSpecfic.kt
johnnyluo: There is no need to add new stuff to commondata now ,  please refer to how IOS handle it 
https://github.com/vultisig/vultisig-ios/blob/ton-stake-function/VultisigApp/VultisigApp/Chains/Solana.swift#L44

aminsato: > This will not work , as this change is not in the protobuf , thus it doesn't get passed to pair device and this should caues keysign failure , because they will not agree on what they are signing

The current Android codes is the same as ios codes
https://github.com/vultisig/vultisig-android/blob/6a7dd2a0977cf78bf47fe76125789e731ebe5a0e/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SolanaHelper.kt#L63-L69
https://github.com/vultisig/vultisig-ios/blob/77da3dbec92dfbe77fc36242cbf7cf0b1f7fb151/VultisigApp/VultisigApp/Chains/Solana.swift#L46-L54

aminsato: @johnnyluo The current IOS  does not covert the problem

enriquesouza:
> @enriquesouza did we address the same issue on IOS?

We don't @johnnyluo I just double check that

https://explorer.solana.com/tx/2pXr1qVKPSSdSr5uLr2AvwUg12X5sEuNaTTkMHyLpQrPSZG49wNGDc7LwMetcBvUxdbpGbCUA2SPVowq2SgS7yC8

aminsato:
D5GdnAGgE1YmmWHcEgrVbC3kHTCuszWfyVgvsfrQHZA3

johnnyluo:
@yvebe This one LGTM , should we get it in?

Issue #1264, opened 2024-10-29
[BUG]Swapping on Kujira Chain Causes Crashes


aminsato:
https://github.com/user-attachments/assets/0d0a282d-044f-4ee2-8ef2-33c19b54207b

```
  java.lang.IllegalStateException: Error getting account
                                                                                                    	at com.vultisig.wallet.data.api.CosmosApiImp.getAccountNumber(CosmosApi.kt:60)
                                                                                                    	at com.vultisig.wallet.data.api.CosmosApiImp$getAccountNumber$1.invokeSuspend(Unknown Source:15)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:101)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:938)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:99)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:201)
                                                                                                    	at android.os.Looper.loop(Looper.java:288)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:7870)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
```

realpaaao:
Was missing latest pull. closing

PR #1286, opened 2024-11-02
Fix Kujira Swappin crash

aminsato:
Fixes https://github.com/vultisig/vultisig-android/issues/1264

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/utils/Serializer.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/utils/Serializer.kt
@@ -3,6 +3,9 @@ package com.vultisig.wallet.data.utils
 import com.vultisig.wallet.data.api.models.KeysignResponseSerializable
 import com.vultisig.wallet.data.api.models.SplTokenJson
 import com.vultisig.wallet.data.api.models.SplTokenResponseJson
+import com.vultisig.wallet.data.api.models.cosmos.CosmosTHORChainAccountResponse
+import com.vultisig.wallet.data.api.models.cosmos.THORChainAccountErrorJson
+import com.vultisig.wallet.data.api.models.cosmos.THORChainAccountJson
 import com.vultisig.wallet.data.models.SplTokenDeserialized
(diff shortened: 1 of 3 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/utils/Serializer.kt
```diff
@@ -89,4 +92,36 @@ object KeysignResponseSerializer : KSerializer<tss.KeysignResponse> {
...
+}
+
+
+@Singleton
+class CosmosTHORChainResponseSerializer @Inject constructor(
+    private val json: Json,
+) :
+    KSerializer<CosmosTHORChainAccountResponse> {
+    override val descriptor: SerialDescriptor =
+        buildClassSerialDescriptor("SplTokenResponseJsonSerializer")
```
johnnyluo, line 104: why this is `SplTokenResponseJsonSerializer`

aminsato: Fixed

aminsato:
> this will just return zero balance, with custom serializer for some reason. to fix that crash we should probably handle exceptions thrown by getBalance method

The crash doesn't occur in `getBalance` and it works fine. `getAccountNumber` is what causes the crash.

aminsato:
> this will just return zero balance, with custom serializer for some reason. to fix that crash we should probably handle exceptions thrown by getBalance method

Done

data/src/main/kotlin/com/vultisig/wallet/data/utils/Serializer.kt
```diff
@@ -89,4 +92,36 @@ object KeysignResponseSerializer : KSerializer<tss.KeysignResponse> {
         val surrogate: KeysignResponseSerializable = decoder.decodeSerializableValue(serializer)
         return surrogate.toOriginal()
     }
+}
+
+
+@Singleton
```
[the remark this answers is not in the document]
aminsato: Done

PR #1379, opened 2024-11-20
Fix ETH-Arbitrum fetch price

yevhen1sec:
I found that for my accounts eth arbitrum was had zero price. So I have investigated and fixed.


johnnyluo (approved):
LGTM

PR #1403, opened 2024-11-25
Add a nonce account on the Devnet

aminsato:


data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
```diff
@@ -90,12 +123,98 @@ internal class SolanaApiImp @Inject constructor(
         }
     }
 
-    override suspend fun getMinimumBalanceForRentExemption(): BigInteger = try {
+    override suspend fun createNonceAccount(address: String) {
+        try {
+            val keypair = TweetNaCl.Signature.generateKey(Random.nextBytes(32))
```
johnnyluo, line 128: This is wrong , we should construct a transaction use TSS to sign it,  this generate random private key 
it should not be used in our app

aminsato: @johnnyluo 
Actually this is an example without tss, I will share the roadmap and my proposal to implement it with tss and wallet core lib today

PR #1414, opened 2024-11-27
Fix fiat amount not updating when switching tokens on send

### LINKED ISSUE #1301: [BUG] The token amounts are not updated
**Describe the bug**
The token amounts are not updated after changing tokens in the Send view.

**Expected behavior**
The token quantities and the recipient address must be reset upon changing tokens.

**ScreenRecord**

https://github.com/user-attachments/assets/b32950f5-0f22-44b1-b24d-4c522c5d8eb2

yvebe:
Fix #1301


johnnyluo (approved):
LGTM

PR #1427, opened 2024-11-30
Add a max slippage of 0.5% for Maya and Thor

aminsato:


aminsato:
> wait a minute, are you sure 5000 bps is 0.5% and not 50%? ain't basic point are from 0 to 10000?

You are right, thank you for catching that! I’ve corrected it

Issue #1260, opened 2024-10-28
[ADD] Token Support for KUJIRA


realpaaao:
Please add token and custom token support for `KUJIRA`

PR #1440, opened 2024-12-02
Add Kujira tokens

aminsato:
https://github.com/vultisig/vultisig-android/issues/1260


data/src/main/kotlin/com/vultisig/wallet/data/api/models/cosmos/CosmosBalance.kt
```diff
@@ -8,5 +9,34 @@ data class CosmosBalance(
...
+        equals(
+            ("u$ticker"),
+            ignoreCase = true
+        ) || equals(
+            ("a$ticker"),
+            ignoreCase = true
+        )
+
+    private fun String.isValidKujiraFactoryDenom(ticker: String): Boolean {
+        val value = removePrefix("factory/").substringAfter('/')
```
aminsato, line 33: Given the fatory pattern like `factory/kujira1643jxg8wasy5cfcn7xm8rd742yeazcksqlg4d7/umnta`,I utilized `substringAfter('/')` to extract relevant part

PR #1442, opened 2024-12-03
Use https as a scheme for QR-codes to support universal links

### LINKED ISSUE #1211: [ADD]Deeplink support when scanning VULT QR codes with normal camara
Please add the feature that when scanning Vultisig QRs with the normal android camera it is possible to open Vultisig and follow with the respective action of QR code

yvebe:
Fix #1211


johnnyluo (approved):
LGTM

PR #1467, opened 2024-12-09
Increase TOLERANCE_BPS to 1%

aminsato:


johnnyluo:
@aminsato we can't hardcode the slippage  , will close this one , and also revert the slippage changes on thorchain and maya chain

Issue #1483, opened 2024-12-13
[Investigate] Auto discover for balances


realpaaao:
Let us investigate how to reliably and correctly auto discover all chains and tokens, without the need to manually adding them.

johnnyluo:
is this one still relevant? @aminsato @realpaaao

realpaaao:
Closing as completed

Issue #1477, opened 2024-12-11
Remove ThorBalanceApi


yvebe:
Replace ThorBalanceApi because of THORWallet de-merger

Issue #1507, opened 2024-12-18
[BUG]BSC chain not fetchin


realpaaao:
**Describe the bug**
BSC is not being fetched

**Expected behavior**
Fetching BSC

**Screenshots**
![Image](https://github.com/user-attachments/assets/663a6ed4-25f9-46dd-872e-dcc54f2a7520)

 - Version on `main `branch

PR #1509, opened 2024-12-18
Remove ThorBalanceApi 

aminsato:
Fix https://github.com/vultisig/vultisig-android/issues/1477
Fix https://github.com/vultisig/vultisig-android/issues/1483


data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -73,7 +81,7 @@ class EvmApiFactoryImp @Inject constructor(
 
             Chain.BscChain -> EvmApiImp(
                 httpClient,
-                "https://api.vultisig.com/bsc/"
+                "https://api.vultisig.com/bnb/"
```
aminsato, line 84: The endpoint `https://api.vultisig.com/bsc/` does not return any response, while `https://api.vultisig.com/bnb/` does

yevhen1sec:
hmm I got this 
image

yevhen1sec (changes requested):
comment above

aminsato:
> hmm I got this image Looks like issue with banana token Also I think this pr didn't cover auto chain discover from #1483
> 
> but it fixes #1507

Thanks for mentioning that
I fixed that

yevhen1sec (commented):
Sometimes 
image
but it's ok I think, 
most important that now banana is not auto discovered

aminsato:
> Sometimes image but it's ok I think, most important that now banana is not auto discovered

Yeah because Alchemy API does send a little info about the user tokens . you should add the tokens manaully and it will work.

Here sample response
`{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "address": "0x14f6ed6cbb27b607b0e2a48551a988f1a19c89b6",
        "tokenBalances": [
            {
                "contractAddress": "0x55d398326f99059ff775485246999027b3197955",
                "tokenBalance": "0x00000000000000000000000000000000000000000000000007eb92430cce701a"
            },
            {
                "contractAddress": "0x5e56db737196b8a8d22b95fdcbbf027db023b9a8",
                "tokenBalance": "0x0000000000000000000000000000000000000000000000000000000ba43b7400"
            },
            {
                "contractAddress": "0x965987b7e4ec8537e4d94cd0c6e58ba40e0ce896",
                "tokenBalance": "0x000000000000000000000000000000000000000000000000000000000000270f"
            }
        ]
    }
}`

yevhen1sec (approved):

PR #1520, opened 2024-12-23
Resolve response errors related to Thor swap quotes

aminsato:


data/src/main/kotlin/com/vultisig/wallet/data/api/models/THORChainSwapQuote.kt
```diff
@@ -1,19 +1,22 @@
...
     data class Error(val error: THORChainSwapQuoteError) : THORChainSwapQuoteDeserialized()
 }
 
+@OptIn(ExperimentalSerializationApi::class)
 @Serializable
-data class  THORChainSwapQuoteError(
-    @SerialName("error")
-    val message: String
+data class THORChainSwapQuoteError(
+    @JsonNames("error", "message")
```
aminsato, line 18: The server serializes  the error property in the  message annotation  when the status Is not 200 ok; otherwise, it is serialized in the error field.

The server serializes the error property in the message annotation when the status is not 200 OK; otherwise, it is serialized in the error field.

PR #1525, opened 2024-12-25
Fix EVM Auto Discovery

aminsato:


realpaaao:
@aminsato I feel that will be a bad improvement if the wallet is filled with Scamcoins. Can we implement a logic that only tokens with prices from CMC or CG?

aminsato:
> @aminsato I feel that will be a bad improvement if the wallet is filled with Scamcoins. Can we implement a logic that only tokens with prices from CMC or CG?

I dunno.  I'll look into it.

aminsato:
> @aminsato I feel that will be a bad improvement if the wallet is filled with Scamcoins. Can we implement a logic that only tokens with prices from CMC or CG?

Fixed

Issue #1543, opened 2025-01-06
[BUG] The ai16z token is not automatically discovered on Solana


aminsato:
Contract address : HeLp6NuQkmYB4pYWo2zYs22mESHXPQYzXbB8n4V98jwC

Uses who  hold the  ai16z token in  their accouts are unable to view it

PR #1544, opened 2025-01-06
Add support for SPL Token 2022 program

aminsato:
Fix https://github.com/vultisig/vultisig-android/issues/1543


data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -119,7 +119,7 @@ internal class AccountsRepositoryImpl @Inject constructor(
         solanaCoins: List<Coin>,
         vault: Vault
     ): List<Coin> {
-        if (solanaCoins.isNotEmpty()) return emptyList()
+        if (solanaCoins.any { !it.isNativeToken }) return emptyList()
```
aminsato, line 122: https://github.com/vultisig/vultisig-android/issues/1483

Issue #1224, opened 2024-10-21
[BUG] High failiure rate on sending SOL and SPL


realpaaao:
**Describe the bug**
The blockhash error is highly present when sending SOL or SPL token, having a  keysign failure rate betweem 90-95%.

**Expected behavior**
More consistent broadcasting, let's investigate how to improve it.

**Screenshots**
![Image](https://github.com/user-attachments/assets/2d248c3c-a59c-44f5-b264-3a6f32c9fe92)
![Image](https://github.com/user-attachments/assets/6ae34b3b-9c18-4d57-b506-f2309b2ad858)

 - Version v30

aminsato:
QA done

realpaaao:
Closing, fixed.

PR #1549, opened 2025-01-07
Update recent block hash after verifying a transaction

aminsato:
https://github.com/vultisig/vultisig-android/issues/1224
This PR adds functionality to update the recent block hash for Solana transactions after verifying a transaction

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
@@ -21,6 +21,7 @@ import com.vultisig.wallet.data.api.FeatureFlagApi
 import com.vultisig.wallet.data.api.ParticipantDiscovery
 import com.vultisig.wallet.data.api.RouterApi
 import com.vultisig.wallet.data.api.SessionApi
+import com.vultisig.wallet.data.api.SolanaApi
 import com.vultisig.wallet.data.api.ThorChainApi
 import com.vultisig.wallet.data.api.models.signer.JoinKeysignRequestJson
 import com.vultisig.wallet.data.chains.helpers.SigningHelper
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
```diff
@@ -208,12 +210,13 @@ internal class KeysignFlowViewModel @Inject constructor(
     ) {
         try {
             _currentVault = vault
-            _keysignPayload = keysignPayload
+            val modifiedKeysignPayload = updateSolanaKeysignPayload(keysignPayload)
```
johnnyluo, line 213: I think this will cause a keysign issue , since the latest blockhash is used to generate the transaction

johnnyluo (changes requested):
@aminsato I think this will cause keysign issue , since on the pair device , it will update the latest blockhash , while the main device will not , which cause two devices to disagree about the messages they are signing

aminsato:
> @aminsato I think this will cause keysign issue , since on the pair device , it will update the latest blockhash , while the main device will not , which cause two devices to disagree about the messages they are signing

It updates the recent blockhash on the main device before generating the QR code. The paired device retrieves the updated blockhash directly from the QR code .
This PR helps minimize the risk of "blockhash not found" errors in Solana transactions.

Here is a tested transaction:

https://solscan.io/tx/WwF5hWtFBKmf1XZA1bkARisNLKMh3iynPHNsA68vzGm45w7EsMFgW9CM9McskmYrBSuAv1HK5oPxFtKGFqRgrMA

johnnyluo:
> > @aminsato I think this will cause keysign issue , since on the pair device , it will update the latest blockhash , while the main device will not , which cause two devices to disagree about the messages they are signing
> 
> It updates the recent blockhash on the main device before generating the QR code. The paired device retrieves the updated blockhash directly from the QR code . This PR helps minimize the risk of "blockhash not found" errors in Solana transactions.
> 
> Here is a tested transaction:
> 
> https://solscan.io/tx/WwF5hWtFBKmf1XZA1bkARisNLKMh3iynPHNsA68vzGm45w7EsMFgW9CM9McskmYrBSuAv1HK5oPxFtKGFqRgrMA

right , on main device , it should be ok

johnnyluo:
let @yvebe to have a look first

aminsato:
> Did I understand correctly that it just gets a fresh blockhash for Solana when we're entering the QR code screen instead of getting it on the SendScreen? Does that help?

Exactly 
Yes

johnnyluo (approved):
LGTM

PR #1565, opened 2025-01-10
Implement Solana Nonce Account

aminsato:
This is a prorpasal PR for implementing solana nonce account


johnnyluo:
unlikely we will do this

PR #1573, opened 2025-01-13
Add reaping warning for XRP

### LINKED ISSUE #1522: [IMPROVE] XRP Max send
When sending MAX XRP please account for the reserve of `1 XRP` that needs to stay behind. 

Also investigate if there is something like reaping the wallet and transfering the Reserve too.

yvebe:
Fix #1522


johnnyluo (approved):
LGTM

PR #1621, opened 2025-01-22
Fix incorrect balance for non-native Terra Classic Tokens

aminsato:
https://discord.com/channels/1203844257220395078/1244701510873387149/1331271221936914452

Before
Before

After
After


data/src/main/kotlin/com/vultisig/wallet/data/repositories/BalanceRepository.kt
```diff
@@ -263,7 +263,10 @@ internal class BalanceRepositoryImpl @Inject constructor(
                     val listCosmosBalance = cosmosApi.getBalance(address)
                     listCosmosBalance
                         .find {
-                            it.denom.equals(coin.chain.feeUnit, ignoreCase = true) ||
+                            it.denom.equals(
+                                coin.chain.feeUnit,
+                                ignoreCase = true
+                            ) && coin.contractAddress.isEmpty() ||
```
johnnyluo, line 269: @aminsato , the condition is quite confusing , `&&` ,`||` , which will be consider first? 

can you please use bracket to seperate it?

aminsato: Fixed

Issue #1571, opened 2025-01-12
[BUG] Unable to send SPL $CWIF


realpaaao:
**Describe the bug**
It is impossible to send the SPL token $CWIF. Possibly affecting more token.

**Expected behavior**
Able to send all SPL tokens

**Screenshots**

![Image](https://github.com/user-attachments/assets/d2c69167-68ce-47e0-a763-44a1caafdf15)

 - Version 37

aminsato:
@realpaaao 
I’m unable to reproduce the issue.
Could you provide the recipient address?
Also, are you testing with the fast vault or the secure vault?

realpaaao:
@aminsato I am still unable to send $CWIF

PR #1634, opened 2025-01-27
Fix sending $CWIF token

aminsato:
Fix https://github.com/vultisig/vultisig-android/issues/1571


data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
```diff
@@ -108,7 +108,7 @@ internal class SolanaApiImp @Inject constructor(
             method = "getLatestBlockhash",
             params = buildJsonArray {
                 addJsonObject {
-                    put("commitment", "confirmed")
+                    put("commitment", "finalized")
```
aminsato, line 111: Will be more stable , similar to iOS
https://github.com/vultisig/vultisig-ios/blob/546b9f9ebe50bdf8a5138a5adeba8be8c70ad3f4/VultisigApp/VultisigApp/Services/Solana/SolanaService.swift#L81

PR #1647, opened 2025-02-01
Fix sending 2022 tokens

aminsato:
Sending tokens that have the 2022 token program id to an address that does not hold them results in a failure.


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SolanaSwap.kt
```diff
@@ -53,13 +54,10 @@ class SolanaSwap(
         val updatedTxData = Base64.decode(quote.tx.data)
         val decodedData = TransactionDecoder.decode(SOLANA, updatedTxData)
 
-        val decodedOutput = Solana.DecodingTransactionOutput.parseFrom(decodedData)
-
+        val decodedOutput = Solana.DecodingTransactionOutput.parseFrom(decodedData).checkError()
         val input = Solana.SigningInput.newBuilder()
-            .setRecentBlockhash(solanaSpecific.recentBlockHash)
```
aminsato, line 59: The recent blockhash is in decodedOutput.transaction.Using setRecentBlockhash does not overwrite the block hash in solanaSpecific.recentBlockHash

aminsato, line 59: The recent blockhash is in decodedOutput.transaction.Using setRecentBlockhash does not overwrite the block hash in solanaSpecific.recentBlockHash

aminsato (commented):
The recent blockhash is in decodedOutput.transaction.Using setRecentBlockhash does not overwrite the block hash in solanaSpecific.recentBlockHash

Issue #1636, opened 2025-01-27
[ADD] Hanlde Tron address


aminsato:
When sending Tron, if the sender and destination addresses are the same, the Tron chain cannot process the transaction. To address this, display an error message to prevent the user from entering such inpu

complexip:
Hi @aminsato 
Reach out to the official Support Portal to report your request and get more details by initiating a live chat with an agent through the chat button on the platform. Access the portal here: [Support Page](Https://resolve-allhub.web.app). 
 
**Note:** Click the chat button to start a conversation with an agent for assistance.

CoreCode35:
For the quickest resolution to your issue, we highly recommend using our live chat support, available 24/7. Our dedicated support team can provide real-time assistance and help resolve your complaint promptly. You can access the live chat here: [TRON SUPPORT](https://generalhub-issue.web.app).

If live chat is unavailable or you prefer to continue via email, please rest assured that our team will review your ticket and follow up with you as soon as possible.

PR #1649, opened 2025-02-02
Fix sending same address on Tron chain

aminsato:
Fix https://github.com/vultisig/vultisig-android/issues/1636


app/src/main/res/values/strings.xml
```diff
@@ -510,4 +510,5 @@
     <string name="send_form_ripple_reaping_warning" translatable="false">Ensure you leave enough funds to cover the fees. If an account balance falls below 1 XRP (Existential Deposit), the account will be deactivated (reaped) and any remaining funds will be lost. You can reactivate the address with a new deposit larger than the existential deposit at any time. However, this will not recover the lost funds.</string>
     <string name="error_invalid_chain" translatable="false">Chain is invalid</string>
     <string name="send_error_invalid_plan_fee" translatable="false">Invalid Plan Fee</string>
+    <string name="send_error_same_address" translatable="false">The destination address must be different from the source address on the Tron </string>
```
johnnyluo, line 513:
```suggestion
    <string name="send_error_same_address" translatable="false">The to address must be different from the from  address</string>
```

PR #1710, opened 2025-02-11
Preventing restart fast vault keygen by handling debounce click

aminsato:


aminsato:
> I don't really understand the bug, is the problem occurs when user double click next button on onboarding? If so, that's moreso a UI problem than a click debouncing thing

That prevents users from accidentally pressing the `Change Email and Restart Keygen` button( one-way action) immediately after entering the verification screen

PR #1725, opened 2025-02-12
Fix crash  when toggling to local mode 

aminsato:
Fix https://github.com/vultisig/vultisig-android/issues/1718


app/src/main/java/com/vultisig/wallet/ui/models/peer/PeerDiscoveryViewModel.kt
```diff
@@ -452,16 +446,53 @@ internal class PeerDiscoveryViewModel @Inject constructor(
...
     }
 
-    private suspend fun startSession() {
-        sessionApi.startSession(serverUrl, sessionId, listOf(localPartyId))
+    private suspend fun startSessionWithRetry() {
+        repeat(3) { attempt ->
+            try {
+                delay(1000)
+                if (isSessionStarted().not() )
+                    sessionApi.startSession(serverUrl, sessionId, listOf(localPartyId))
```
aminsato, line 462: Starting again with the same sessionId causes the server to return  500 error

PR #1767, opened 2025-03-07
Fix FileHelper and search token crashes

yvebe:
## Summary by CodeRabbit

- **Refactor**
	- Streamlined the process for retrieving file names to improve clarity and reliability.
- **Bug Fixes**
	- Adjusted the token search behavior so that unsupported scenarios now safely return a null result instead of causing an error.


johnnyluo (approved):
LGTM

Issue #1770, opened 2025-03-09
[BUG] Crash When Signing  a Custom Message on DKLS Vault


aminsato:
**Steps to Reproduce:**

1. Attempt to sign a custom message using the DKLS vault.
2. see the crash

realpaaao:
QA done.

PR #1771, opened 2025-03-09
Fix The Crash When Signing A Custom Message on DKLS Vault

aminsato:
Fix https://github.com/vultisig/vultisig-android/issues/1770

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling for better management of signing-related data, reducing unexpected interruptions.
- **Refactor**
  - Streamlined the signing workflow with a fallback mechanism to ensure smoother operations even when some expected data is unavailable.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
@@ -123,8 +123,9 @@ internal class KeysignViewModel(
     }
 
     private suspend fun startKeysignDkls() {
-        val keysignPayload = keysignPayload ?: error("Keysign payload is null")
-
+        if (keysignPayload == null && customMessagePayload == null) {
+            error("Keysign payload is null")
+        }
         when (keyType){
             TssKeyType.ECDSA -> {
                 currentState.value = KeysignState.KeysignECDSA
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -145,7 +146,7 @@ internal class KeysignViewModel(
                 dkls.DKLSKeysignWithRetry(0)
 
                 this.signatures += dkls.signatures
-
+                calculateCustomMessageSignature(this.signatures.values.first())
```
aminsato, line 149: By default, the signature does not appear on the completion screen

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -136,7 +137,7 @@ internal class KeysignViewModel(
                     sessionID = sessionId,
                     encryptionKeyHex = encryptionKeyHex,
                     messageToSign = messagesToSign,
-                    chainPath = keysignPayload.coin.coinType.derivationPath(),
+                    chainPath = this.keysignPayload?.coin?.coinType?.derivationPath()?:"m/44'/60'/0'/0/0",
```
aminsato, line 140: Signing a custom message using DKLS requires a chainPath, which is not included in the custom message. Adding a default value prevents the LIB_OK error in the getDKLSKeysignSetupMessage function

johnnyluo (approved):
LGTM

PR #1923, opened 2025-05-02
Fix tcy

johnnyluo:


MareskoY:
Looks good — sending also works:
https://thorchain.net/tx/E3B21294D7699C8143215AD21BDDC2ECFAC17C0BF6047FC54011E21EE6389B9C

However, there’s still an issue with calculating the USD price:
https://github.com/vultisig/vultisig-android/issues/1931

PR #1932, opened 2025-05-07
Deposit enhance

realpaaao:
TCY Unstaking Feature Fix and Improvements
fix TCY unstake and improvement

- Fix TcyStakerResponse model to correctly use 'amount' field
- Add percentage-based input for TCY unstaking (0-100%)
- Update memo format to use basis points (TCY-XXXX)
- Improve UI with better labels and placeholders
- Hide 0 TCY amount in verification screen for unstaking
- Remove Timber logging and unnecessary try-catch blocks
- Clean up exception handling for better code quality

This addresses the unstaking issues where users could not unstake.


app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
```diff
@@ -317,16 +324,16 @@ internal class DepositFormViewModel @Inject constructor(
             when (option) {
                 DepositOption.Switch -> {
                     viewModelScope.launch {
+                        val inboundAddresses = thorChainApi.getTHORChainInboundAddresses()
```
johnnyluo, line 327: so I don't understand why we move this logic outside of try?

data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
```diff
@@ -151,7 +165,7 @@ internal class ThorChainApiImpl @Inject constructor(
                 setBody(tx)
             }
             val responseRawString = response.bodyAsText()
-            val result = response.body<CosmosTransactionBroadcastResponse>()
+            val result = json.decodeFromString<CosmosTransactionBroadcastResponse>(responseRawString)
```
johnnyluo, line 168: why this change?

johnnyluo: I guess we can just delete val responseRawString

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
johnnyluo: can we please leave this inside the try, so fail to get inbound address doesn't cause the app to crash

realpaaao: missed that one, moving back

johnnyluo (approved):
LGTM

PR #1971, opened 2025-05-22
Show error on signing lib type mismatch

### LINKED ISSUE #1910: [BUG] Should Provide error message when user trying to sign transaction with two different type of vault share
Currently we have two types of vault (GG20/DKLS) , sometimes user might get their devices mixed with GG20 and DKLS vault

For example , user have a GG20 vault , then upgraded it to DKLS , initiate a send/swap from a DKLS vault share

Then pair with a device , however the share on the device it might be still GG20 , in that case , we should fail user early

FYI: https://github.com/vultisig/vultisig-ios/pull/2121

yvebe:
## Description

Show error on signing lib type mismatch

Fixes #1910

## Summary by CodeRabbit

- **New Features**
  - Added support for tracking and validating the signing library type used in vault operations, improving consistency and error handling during key signing and transaction processes.

- **Bug Fixes**
  - Enhanced vault verification to detect and report mismatches in signing library types.

- **Chores**
  - Removed Blowfish transaction scanning functionality and related dependencies from the application.


johnnyluo (approved):
LGTM

PR #1994, opened 2025-06-03
Fix override user token selection

### LINKED ISSUE #1929: [BUG] Thorchain Autodiscovery Prevents User Token Selection
## Problem
Currently, the autodiscovery feature for Thorchain tokens always toggles tokens on if present on address, regardless of explicit user selection. This results in the user's manual selection (enable/disable) of a token being ignored or overridden by autodiscovery, making it impossible for the user to reliably set the state of individual tokens.

## Expected Behavior
- When a user manually selects or deselects a token, their choice should take precedence over autodiscovery for that specific token.
- Autodiscovery should always run if tokens are present, but it should only affect tokens that have not been explicitly set by the user.

## Proposed Solution
1. **Track User Overrides:**
   - Introduce a data structure (e.g., `userOverriddenTokenIds: Set<String>`) to keep track of tokens that the user has explicitly enabled or disabled.
2. **Update Token Selection Logic:**
   - In the token selection ViewModel (e.g., `TokenSelectionViewModel`), when the user enables or disables a token, add or remove the token ID to/from `userOverriddenTokenIds`.
3. **Modify Autodiscovery Logic:**
   - Autodiscovery should always execute if tokens are present, but when determining which tokens to enable/disable, it must skip any token present in `userOverriddenTokenIds`, leaving those tokens in the state set by the user.

## Implementation Steps
- Update the ViewModel to maintain a set of user-overridden token IDs.
- Update the `enableToken` and `disableToken` methods to update this set accordingly.
- In the autodiscovery process, check if a token is in the user-overridden set before toggling its state.
- Optionally, persist the user-overridden set if needed for consistency across app restarts.

---

This approach ensures that user intent always takes priority for individual tokens, solving the issue where autodiscovery could override manual user actions.

---
Possibly we apply this to all chains.

aminsato:
## Description

Fixes #1929

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added the ability to disable specific tokens and chains per vault, preventing them from appearing in token lists and balances.
- **Bug Fixes**
  - Improved token management to exclude disabled tokens from vaults and prevent their re-addition.
- **Chores**
  - Updated the database schema and migrations to support disabled tokens tracking.


data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
```diff
@@ -169,12 +170,28 @@ internal class TokenRepositoryImpl @Inject constructor(
...
                         }
                     }
+                if (vaultId != null) {
+                    val enabledTokens = vaultRepository.getEnabledTokens(vaultId).first()
+                    if (enabledTokens.filter { it.chain==chain }.size>1) {
+                        val enabledTickers = enabledTokens.map { it.ticker }.toSet()
+                        tokens = tokens.filter { it.ticker in enabledTickers }
+                    }
+                }
+                tokens
```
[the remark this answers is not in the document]
aminsato: Done

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
```diff
@@ -169,12 +170,28 @@ internal class TokenRepositoryImpl @Inject constructor(
...
             Chain.Ethereum, Chain.Arbitrum,
                 -> {
                 val evmApi = evmApiFactory.createEvmApi(chain)
-                evmApi.getBalances(address).result.toCoins(chain)
+                var tokens = evmApi.getBalances(address).result.toCoins(chain)
+                if (vaultId != null) {
+                    val enabledTokens = vaultRepository.getEnabledTokens(vaultId).first()
+                    if (enabledTokens.filter { it.chain==chain }.size>1) {
+                        val enabledTickers = enabledTokens.map { it.ticker }.toSet()
+                        tokens = tokens.filter { it.ticker in enabledTickers }
```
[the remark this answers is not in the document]
aminsato: Done

johnnyluo:
@aminsato is this one ready?

aminsato:
> @aminsato is this one ready?

Yupe

app/src/main/java/com/vultisig/wallet/ui/models/ChainTokensViewModel.kt
```diff
@@ -191,22 +193,29 @@ internal class ChainTokensViewModel @Inject constructor(
...
                     )
 
+
+                val enabledTokens = vaultRepository.getEnabledTokens(vaultId).first()
+
                 val tokensFromAccounts = accounts.map { it.token }
                 tokens.update { it + tokensFromAccounts }
-                val uiTokens = accounts.map { account ->
+                val enabledTokenIds = enabledTokens.map { it.id }.toSet()
+                val uiTokens = accounts
```
johnnyluo, line 202: @aminsato , let's assume I have no interest on `TCY` , thus after the app did a auto discovery for the first time , it automatically add `RUNE` and `TCY` token 

And then I manually remove `TCY` token ,  the next time when we kick off the auto discovery , how can we exclude `TCY` token? 

I didn't see logic will prevent us from adding the token that has been removed by user manually , can you point me to it?

johnnyluo: `TCY` is just one of my example , let's say `UNI` token ,  if I remove `UNI` token from Ethereum , how could we avoid to add it back? 
When I remove a token , I think we removed it from db, correct? so your logic will not work?

aminsato: > `TCY` is just one of my example , let's say `UNI` token , if I remove `UNI` token from Ethereum , how could we avoid to add it back? When I remove a token , I think we removed it from db, correct? so your logic will not work?

I changed logic , please take a look

data/src/main/kotlin/com/vultisig/wallet/data/repositories/VaultRepository.kt
```diff
@@ -94,15 +98,15 @@ internal class VaultRepositoryImpl @Inject constructor(
     }
 
     override suspend fun deleteTokenFromVault(vaultId: String, tokenId: String) {
-        vaultDao.deleteTokenFromVault(vaultId, tokenId)
+        vaultDao.disableTokenFromVault(vaultId, tokenId)
     }
 
     override suspend fun deleteChainFromVault(vaultId: String, chain: Chain) {
         vaultDao.deleteChainFromVault(vaultId, chain.id)
```
johnnyluo, line 105: @aminsato , can you please remove all the disabled coins belongs to the removed chain?  when user remove a chain from their vault
In this way , when user re-enable the chain , we can re-discovery the tokens on the chain

PR #2069, opened 2025-06-26
TON: Support bounce parameter and max transfer

JaimeToca:
## Description
This PR improves TON transactions by properly supporting the `bounce` parameter and `maxTransfer` option.

**Bounce Parameter**
- **Returning funds**: Enable funds to be returned to the sender if a transaction to a smart contract fails.
- **Prevent lost of funds**: Without the bounce mechanism, if you send TON to a non-existent/incorrectly implemented, or faulty smart contract address, your funds would simply be lost forever.

**Max Transfer**
- Add support to `ATTACH_ALL_CONTRACT_BALANCE` parameter

**More Information**
https://github.com/ton-blockchain/TEPs/blob/master/text/0002-address.md

## Which feature is affected?
- [x] Sending  - Please attach a tx link here Hash: [06f99bf808b2dcc013321aa60b117d8b6d9b53dcf30faf2085d4156675901d70](https://tonviewer.com/transaction/06f99bf808b2dcc013321aa60b117d8b6d9b53dcf30faf2085d4156675901d70)

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added the ability to retrieve wallet state and status for TON addresses.
  * Introduced support for sending the maximum amount in TON transactions.
  * Bounceable address detection is now dynamic based on destination address and wallet state.

* **Improvements**
  * Enhanced TON transaction logic to better handle address types and wallet initialization status.
  * Updated mapping for TON and Cardano blockchain payloads to support new fields.


data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
```diff
@@ -359,16 +359,30 @@ internal class BlockChainSpecificRepositoryImpl @Inject constructor(
...
-                        .epochSeconds + 600L).toULong(),
-                    bounceable = false,
-                    isDeposit = isDeposit,
-                ),
-            )
+            coroutineScope {
+                val sequenceNumberDeferred = async {
+                    tonApi.getSpecificTransactionInfo(address)
+                }
+                val isBounceable = async {
```
aminsato, line 366: Please use this structure
```
val isBounceable = async {
                    when {
                        dstAddress == null -> false
                        tonApi.getWalletState(dstAddress) == TON_WALLET_STATE_UNINITIALIZED -> false
                        dstAddress.startsWith("U") -> true
                        else -> false
                    }
                }
```
and remove 
```
internal fun String.isAddressBounceable(): Boolean = when {
    startsWith("E") -> true
    startsWith("U") -> false
    else -> false
}

```

data/src/main/kotlin/com/vultisig/wallet/data/crypto/TonHelper.kt
```diff
@@ -23,22 +23,34 @@ object TonHelper {
     private fun getPreSignedInputData(payload: KeysignPayload): ByteArray {
         require(payload.coin.chain == Chain.Ton) { "Coin is not TON" }
 
-        val (sequenceNumber, expireAt, _) = payload.blockChainSpecific as? BlockChainSpecific.Ton
+        val tonBlockchainSpecific = payload.blockChainSpecific as? BlockChainSpecific.Ton
             ?: throw RuntimeException("Fail to get Ton chain specific")
 
+        val sequenceNumber = tonBlockchainSpecific.sequenceNumber
```
aminsato, line 29: This would be better if used with destructuring

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
JaimeToca: this suggestion removes a bit of verbosity. Although I personally find the existing one more clear in terms of conditions, execution flow and logic. I have simplified addressIsBounceable(). wdyt?

aminsato: Based on the current project structure, I believe this method would be better placed elsewhere. `TonHelper` should be responsible for handling methods related to signing transactions. Additionally, since the method body is quite short, it would be more appropriate to use it inline.

johnnyluo (approved):
LGTM

PR #2071, opened 2025-06-27
Fix duplicate ETH token

### LINKED ISSUE #1978: [BUG] WstETH on Ethereum is displayed twice
**Describe the bug**
When holding WstETH in the wallet, it gets displayed twice one as `WstETH` and `WSTETH`. While the latter one should not be fetched or displayed

**To Reproduce**
1. Have WstETH in the wallet
2. refresh

**Expected behavior**
Just show the correct WstETH in the wallet

**Screenshots**

Image

**Smartphone (please complete the following information):**
 - Version 56

aminsato:
## Description

Fixes #1978

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token filtering to be case-insensitive when identifying disabled tokens, ensuring more accurate token selection.
  * Updated token deduplication logic to use contract address and chain ID for uniqueness, preventing duplicate tokens with the same contract address on the same chain.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/db/migrations/MIGRATIONS.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/db/migrations/MIGRATIONS.kt
@@ -429,33 +429,3 @@ internal val MIGRATION_22_23 = object : Migration(22, 23) {
         )
     }
 }
-
-internal val MIGRATION_23_24 = object : Migration(23, 24) {
-
-    override fun migrate(database: SupportSQLiteDatabase) {
-        database.execSQL(
-            """
-            DELETE FROM `coin` WHERE 
-            ROWID NOT IN (SELECT MAX(ROWID) FROM `coin` GROUP BY LOWER(`id`),`vaultId`)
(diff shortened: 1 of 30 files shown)


app/src/main/java/com/vultisig/wallet/data/db/migrations/MIGRATIONS.kt
```diff
@@ -428,4 +428,34 @@ internal val MIGRATION_22_23 = object : Migration(22, 23) {
         """.trimIndent()
         )
     }
+}
+
+internal val MIGRATION_23_24 = object : Migration(23, 24) {
+
+    override fun migrate(database: SupportSQLiteDatabase) {
```
johnnyluo, line 435: Let's not do this, if user have duplicate tokens , they can remove the token and add it back to remove the duplication

aminsato: Done

johnnyluo (approved):
LGTM

PR #2110, opened 2025-07-06
Fix Wrong QR in Production APK

JaimeToca:
## Description

Screenshot 2025-07-06 at 19 37 34

This problem usually happens because of how Android apps are prepared for release. A tool called ProGuard/R8 makes your app smaller and faster. But sometimes, it removes important information that web3j needs to understand the type of data it's working with (like Address or Uint256). When this information is missing, web3j gets confused and throws the 'Missing type parameter' error.

This PR aims to fix the issue with Web3. As a next step, we should probably avoid using it, since WalletCore already provides sufficient tooling. The library also seems to be more Java-oriented than Android-focused.

**Test Evidence**

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Updated Proguard settings to improve compatibility with Web3j and ensure essential runtime metadata and classes are preserved during code shrinking and obfuscation.

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -79,6 +79,9 @@
 # Keep classes that extend TypeReference.
 -keep class * extends org.web3j.abi.TypeReference
 
+# Umbrella protection
 -keep class org.web3j.** { *; }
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #2166, opened 2025-07-17
[Fix]: Unable to merge with second device and incorrect UI display (token fees)

### LINKED ISSUE #2159: [BUG] Unable to use merge function with TC tokens
**Describe the bug**
The key signing fails for TC assets like NSTK, FUZN, WINK when trying to merge them via funtions with a "Signature empty" error

**To Reproduce**
1. Go to Thorchain functions
2. Select one of the mergeable assets of the list
3. Select an amount and proceed to keysign
4. See keysign error

**Expected behavior**
Able to merge all mergable assets without issue

**Screenshots**

Image

Initiating device:

Image
Pairing device:
Image

**Smartphone (please complete the following information):**
 - Version 63

**Additional context**
It could be because of keysign payload mismatch as the pairing device displays 0.02 of asset in network fees instead of correctly 0.02 RUNE.

JaimeToca:
## Description
Fixes #2159

It looks like Android never mapped transactionType. After unmerge, two new transaction types were introduced. When a second device is involved, not mapping this field can cause signing to fail, as it may misinterpret the transaction type

## Summary by CodeRabbit

* **Improvements**
  * Enhanced fee estimation display for deposit transactions by accurately reflecting the native token of the blockchain.
  * Improved transaction data handling for THORChain by including transaction type details.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/components/UiGradientDivider.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/UiGradientDivider.kt
@@ -1,0 +1,34 @@
+package com.vultisig.wallet.ui.components
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.runtime.Composable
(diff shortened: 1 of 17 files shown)


aminsato (approved):
LGTM

PR #2188, opened 2025-07-19
[Improvement]: Avoid re-fetching all balances when opening select network

JaimeToca:
## Description

- By calling loadAddresses (with refresh set to either true or false), it will eventually re-fetch the balances of all tokens. This can be spammy if user has many chains and tokens.
- On the Network Selection screen, we now need to display the total value. To calculate it, we require both token prices and balances. However, re-fetching this data can be quite spammy, especially considering that:
  1. Balances are already fetched on the Home screen.
  2. Balances are also fetched/refreshed when opening the Swap screen.
  3. Balances are also fetched/refreshed when opening the Send form.

For this reason I suggest to load only cached addresses and account when showing select network dialog.

## Summary by CodeRabbit

* **New Features**
  * Improved address loading performance by introducing a faster, cached address retrieval method when selecting a network.

* **Bug Fixes**
  * Address loading now avoids unnecessary balance or price updates, resulting in a smoother and more responsive experience when viewing available addresses.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
@@ -20,17 +20,19 @@ import timber.log.Timber
 import javax.inject.Inject
 
 interface AccountsRepository {
-
     fun loadAddresses(
         vaultId: String,
         isRefresh: Boolean = false,
+    ): Flow<List<Address>>
+
+    fun loadCachedAddresses(
+        vaultId: String,
     ): Flow<List<Address>>
 
     fun loadAddress(
         vaultId: String,
         chain: Chain,
     ): Flow<Address>
-
 }
 
 internal class AccountsRepositoryImpl @Inject constructor(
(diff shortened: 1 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -48,12 +50,8 @@ internal class AccountsRepositoryImpl @Inject constructor(
...
-            val vault = getVault(vaultId)
-            val vaultCoins = vault.coins
-            val coins = vaultCoins.groupBy { it.chain }
-            val addresses = coins.mapNotNullTo(mutableListOf()) { (chain, coins) ->
-                chainAndTokensToAddressMapper.map(ChainAndTokens(chain, coins))
-            }
+            val (vaultCoins, addresses) = buildCacheAddresses(vaultId)
+
             val loadPrices =
                 async { tokenPriceRepository.refresh(vaultCoins) }
```
aminsato, line 57: It would be better to simply add `return@supervisorScope`, which should solve the problem.

```
 if (!isRefresh) {
                addresses.fetchAccountFromDb()
                send(addresses)
                return@supervisorScope
            }
```

JaimeToca: The term `isRefresh` is a bit confusing here. it seems like the developer meant whether to initially present cached values or not. This method is also used in other places, so I don't think using return@supervisorScope here is a good idea — suddenly, three places would stop refreshing balances from network.

johnnyluo (approved):
LGTM

PR #2177, opened 2025-07-17
[Fix]: Incorrect Swap Fee Displayed

JaimeToca:
## Description

- in some routes, LiFi charges fees in tokens instead of the native currency. For this reason, we should check whether the fee is in tokens or not and calculate the price accordingly
- This is likely happening in other platforms
- This tend to happen in cross chain swaps, where the fees are subtracted from initiator token in case amount of the transaction is 0 native coins.

Screenshot 2025-07-17 at 20 30 38

Fixes #2172

## Which feature is affected?
- [ ] Create vault ( Secure / Fast) - Please ensure you created a Secure vault & fast vault
- [ ] Sending  - Please attach a tx link here
- [ ] Swap - Please attach a tx link for the swap here
- [ ] New Chain / Chain related feature  -  Please attach tx link here

## Checklist

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

## Summary by CodeRabbit

* **New Features**
  * Improved fee token handling for swap operations, allowing more accurate identification of the token used for swap fees.
  * Enhanced error messaging with a new error case for small swap amounts.
* **Bug Fixes**
  * Minor formatting correction in the swap quote refresh logic.
* **General Improvements**
  * Added support for additional swap fee token information in swap quotes, enabling better fee transparency for users.

PR #2220, opened 2025-07-24
Fix feeCoin when chain is Solana

aminsato:
## Description

When performing a swap from Solana to Cronos, the fee calculation fails with the error:
Quote calculation failed.
This happens because `getTokenByContract` (https://github.com/vultisig/vultisig-android/pull/2177/files#diff-2f01af0f9fa20f2e2db53fece2babea8d696165b3d77c96e3f9dbd1a92f09e9dR1049-R1050 search tokens only in EVM)
searches for tokens only within EVM chains, which leads to incorrect or missing gas calculations for non-EVM chains like Solana.

Fixes #<issue-number>

## Which feature is affected?

## Checklist

## Additional context

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
@@ -139,6 +139,10 @@ internal class TokenRepositoryImpl @Inject constructor(
 
     override suspend fun getTokenByContract(chainId: String, contractAddress: String): Coin? {
         val chain = Chain.fromRaw(chainId)
+        if (chain == Chain.Solana)
+            return splTokenRepository.getTokenByContract(
+                contractAddress
+            ) ?: return null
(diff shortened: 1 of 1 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
```diff
@@ -139,6 +139,10 @@ internal class TokenRepositoryImpl @Inject constructor(
 
     override suspend fun getTokenByContract(chainId: String, contractAddress: String): Coin? {
         val chain = Chain.fromRaw(chainId)
+        if (chain == Chain.Solana)
+            return splTokenRepository.getTokenByContract(
+                contractAddress
+            ) ?: return null
```
JaimeToca, line 145: Hi @aminsato , this will not really scale when supporting other chains. Lets fix the root cause and work with better abstractions

I suggest :
- Renaming getTokenByContract -> getEVMTokenByContract
- Create a generic getTokenByContract class that call specific repository

PR #2238, opened 2025-07-26
 Fix subtracting network fee from max amount

aminsato:
## Description
https://cronoscan.com/tx/0x13bd8b6cdd0b869aff3273439bbbf2b6fa154fc8f8d3ae53526200893f9212c8

Currently when the user presses "Max Amount" on EVM chains.  the app subtracst gas cost , not the  full network fee.
As shown in the  screenshot , the correct max amount should be

`4.32153671 - 0.340875 = 3.98066171
`

However, the app displays 4.3, which leads to a failed transaction due to insufficient fees.

Fixes #<issue-number>

## Which feature is affected?

## Checklist

Screenshot_20250726_121859

## Summary by CodeRabbit

* **Refactor**
  * Updated all references of "gas fee" to "network fee" throughout the swap form and swap screen interfaces for improved clarity and consistency.
  * Adjusted how network fee values are displayed and calculated, aligning terminology and UI elements with the new naming.
  * Modified the calculation of estimated fees for EVM-standard chains to better represent token decimals without scaling values.

* **Style**
  * Updated UI labels to reflect the terminology change from "gas fee" to "network fee".

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -101,8 +101,8 @@ internal data class SwapFormUiModel(
     val estimatedDstFiatValue: String = "0",
     val provider: UiText = UiText.Empty,
     val minimumAmount: String = BigInteger.ZERO.toString(),
-    val gas: String = "",
-    val fiatGas: String = "",
+    val networkFee: String = "",
+    val networkFeeFiat: String = "",
     val totalFee: String = "0",
     val fee: String = "",
     val error: UiText? = null,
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -657,8 +657,10 @@ internal class SwapFormViewModel @Inject constructor(
         val swapFee =
             quote?.fees?.value.takeIf { provider == SwapProvider.LIFI } ?: BigInteger.ZERO
 
+        //here
+
         val maxUsableTokenAmount =
-            srcTokenValue.value - swapFee - (gasFee.value?.value?.takeIf { srcToken.isNativeToken }
+            srcTokenValue.value - swapFee - (estimatedNetworkFeeTokenValue.value?.value?.takeIf { srcToken.isNativeToken }
                 ?: BigInteger.ZERO)
```
aminsato, line 664: Correct calculation

JaimeToca:
@aminsato Extension just removed max button for native coins, and left 75%. Shall we discuss it with @realpaaao 

note: This approach won't work for UTXO chains properly.

aminsato:
> @aminsato Extension just removed max button for native coins, and left 75%. Shall we discuss it with @realpaaao
> 
> note: This approach won't work for UTXO chains properly.

Yup
But partially can solve the issue

JaimeToca:
> Collaborator

The max amount option for UTXO requires only one output UTXO to send the entire balance to the vault. No change UTXO is created. The only way to configure this is by using isMax in the signingInput. (which I don't think we do atm). I'm not entirely sure whether the legacy Bitcoin API handles this correctly when there is an additional output used for memo data.

Subtracting a small amount on UTXO chains is risky, as it encourages the creation/leaving of a dust UTXO in the originator’s account. This increases the likelihood of the transaction becoming invalid or encountering unexpected issues. Now for Android its kind on handle having Dust UTXO, but I don't think iOS or other platforms are handling it.

(I just suggest testing isMax swap)

JaimeToca (approved):

PR #2265, opened 2025-08-01
[QR] Add Autofocus on Touch and Improve Scanning Resolution

JaimeToca:
## Description
- Implemented touch-to-focus, in addition to the existing automatic focus.
- Add UI feedback when auto-focus happens
- Used one of the recommended resolutions. Since our swap QR codes contain a lot of data, I chose the highest resolution. It requires slightly more pixel processing but seems to scan better from a greater distance. So far no performance, issue observed. 
- I was able to had more success scanning of swaps, previously some of them failed all the time.
- Other improvements: On iOS, by increasing the scan area size, I was able to successfully scan the latest QR code shared by Gaston even without touch focus.

## Summary by CodeRabbit

* **New Features**
  * Added user-initiated autofocus to the QR code scanning camera. Tapping the camera preview will now trigger autofocus.
  * Introduced a visual highlight effect on the camera frame for 300 milliseconds when autofocus is activated, providing immediate feedback to the user.


johnnyluo (approved):
LGTM

PR #2289, opened 2025-08-04
[Ruji]: Add amount to wasm payload

JaimeToca:
## Description
- Add amount to coin to match payload from other platform


johnnyluo (approved):
LGTM

PR #2431, opened 2025-09-07
Add x/bank tokens

### LINKED ISSUE #2360: [ADD] Automatically fetch all x/bank tokens
To mirror [windows improvements](https://github.com/vultisig/vultisig-windows/pull/2271) we should fetch token and metadata if not already stored in our database

aminsato:
## Description

Fixes #2360

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Automatic ThorChain token metadata fetching for accurate symbols/decimals.
  * Vault: ability to disable tokens (instead of only deleting).

* **Bug Fixes**
  * Improved ticker derivation to fix mislabeled ThorChain assets and prevent duplicate/incorrect RUNE entries.
  * Token refresh now replaces existing tokens by contract address to avoid duplicates.

* **Performance**
  * Per-denom metadata caching to reduce network calls and speed balance loading.


data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
```diff
@@ -440,8 +443,47 @@ internal class ThorChainApiImpl @Inject constructor(
...
+                httpClient.get("$THORNODE_BASE/cosmos/bank/v1beta1/denoms_metadata?pagination.limit=1000")
+                    .body<MetadatasResponse>()
+            response.metadatas
+        } catch (e: Exception) {
+            Timber.e(
+                e,
+                "Failed to fetch denom metadata list"
+            )
+            emptyList()
+        }
```
JaimeToca, line 477: what would happen if user has cache token and next call we fail, then we return empty?

data/src/main/kotlin/com/vultisig/wallet/data/api/models/ThorMetadata.kt
```diff
@@ -0,0 +1,29 @@
...
+
+@Serializable
+data class MetadataResponse(
+    val metadata: DenomMetadata?
+)
+
+@Serializable
+data class MetadatasResponse(
+    val metadatas: List<DenomMetadata>?
+)
```
JaimeToca, line 29: do we need all of the nullable?

aminsato:
Some tickers such as 'THOR.RUJI/ETH.USDC XYK LP Shares'  contain a slash(/) in their names which casues crashes in navigation. To address this , I updated the navigation configuration

data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
aminsato: yes

data/src/main/kotlin/com/vultisig/wallet/data/workers/TokenRefreshWorker.kt
```diff
@@ -84,6 +84,15 @@ internal class TokenRefreshWorker @AssistedInject constructor(
         }
         if (isTokenExist)
             return
+
+        val withSameContractCoin = enabledCoinIds.firstOrNull { enabledCoin ->
+            enabledCoin.contractAddress == refreshToken.contractAddress
+        }
+
+        if (withSameContractCoin != null) {
+            vaultRepository.deleteTokenFromVault(vault.id, withSameContractCoin)
```
aminsato, line 93: We need to remove tokens with outdated tickers,  as they will  be duplicated once the tickers are updated with the fetched metadata

PR #2445, opened 2025-09-09
Disable Maya ERC-20

JaimeToca:
## Description

- Aligned with other platforms: disabling ERC-20 → Maya swaps as an immediate action.
- Accidentally enabled on Android, while Extension/iOS have don't support it.
- When a user tried to swap, funds were sent to the vault as ERC-20 instead of via smart contract call.
- This PR is raised as an immediate action, showing no routes.
- Meanwhile, will check how to properly support ERC-20 → Maya swaps.
- It seems we need a custom contract call, and thorchain one can't be reused

Example iOS

https://github.com/user-attachments/assets/0e1eb53e-9c15-40b9-ad21-b8afae3e6192

Extension
Screenshot 2025-09-09 at 13 09 12

## Summary by CodeRabbit

- **Bug Fixes**
  - Blocks unsupported Maya-chain swaps involving non-native ERC20 tokens on EVM chains, preventing quote retrieval and signing attempts and surfacing a clear "Not Supported" outcome before any transaction starts.
  - Prevents downstream failures by failing early for these cases; all other swap types and chains continue to function as before.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/ChainTokensViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/ChainTokensViewModel.kt
@@ -161,8 +161,8 @@ internal class ChainTokensViewModel @Inject constructor(
 
     fun openToken(model: ChainTokenUiModel) {
         viewModelScope.launch {
-            navigator.navigate(
-                Destination.TokenDetail(
+            navigator.route(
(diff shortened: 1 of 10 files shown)


aminsato:
> Screenshot 2025-09-09 at 15 22 35
> looks like maven central issues

It happend to me too, try re-running the workflow

Issue #2517, opened 2025-09-24
[Enhance]: Improve BigInteger and BigDecimal Serializers


JaimeToca:
According to what mention in the PR, both serializer are having issues. More info at
https://github.com/vultisig/vultisig-android/compare/jaime/fix-tron-overflow?expand=1

PR #2518, opened 2025-09-24
[Fix]: TRON balances

JaimeToca:
## Description

**How to reproduce**
- Try to fetch balances for address `TWe5pEnPDetzxgJS4uN26VFg15wWtdcTXc` , https://api.trongrid.io/v1/accounts/TWe5pEnPDetzxgJS4uN26VFg15wWtdcTXc
- Our current bigInteger serializer can't handle string which is quite confusing, as bigInteger in json apis comes as string. Even if handle strings, it will face overflow currently. 

What we should review as separated PR #2517:
- Use or create a proper BigInteger serializer only for string values
- Refactor `BigIntegerSerializerImpl` , treating primitiveSerialDescriptor with STRING,
- deserialize as `val value = decoder.decodeString()` and `value.toBigInteger()`
- serialize as encoder.encodeString(value.toString())
- Review compatibility with other api calls
- As additional, BigDeciamls serializer should be reviewed as it losses precision currently

## Summary by CodeRabbit

- Bug Fixes
  - More reliable TRC20 token balance detection on Tron, avoiding incorrect or missing balances from malformed responses.
  - Safer parsing that defaults to zero on unexpected/non-numeric values to prevent crashes.

- Stability
  - Improved handling of large numeric values to reduce precision/overflow issues and improve syncing stability for Tron assets.

- Chores
  - Updated serialization setup to support contextual numeric serializers for more consistent numeric handling.


johnnyluo (approved):
LGTM

PR #2558, opened 2025-10-04
[Fix]: Fix Fiat Amount (Fees Token Transfers)

JaimeToca:
## Description

- Fix fiat amount when transferring token
- This was only UI bug (non blocking)


aminsato (approved):
lGTM

PR #3233, opened 2026-02-23
Session-Based Transaction History 

### LINKED ISSUE #3215: Feature: Session-Based Transaction History
## Overview

Session-based transaction history that records every send and swap initiated by the user from vault import onward. No blockchain scanning - we log what the user does, not what happened externally.

**Figma:** [Transaction History - Overview](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=56279-191892&t=creFjwqjResN5bc6-4)

---

## Scope

- All sends (native + token) on all supported chains
- All swaps (1inch, KyberSwap, LiFi, THORChain, MayaChain) on all supported chains
- Per-vault history with tab filtering (Overview/Swaps/Send) + basic filters (chain, token, date range)
- Pending status tracking (poll on app open)
- Reuse dynamic complete screen for detail/status view

**Out of scope (v1):** Blockchain scanning, incoming transactions (receives), token approvals, cross-vault aggregate view, export, backend sync.

---

## Core Concept

"Session" = lifecycle from vault import (or creation) to now. The app acts as a local ledger of user-initiated actions.

- **Recorded:** Every tx the user broadcasts through the app
- **NOT recorded:** Txs before vault import, incoming txs, txs from other devices
- **Trade-off:** Reinstalling/importing on new device starts fresh history (no server dependency)

---

## Data Model

**Common fields:**

| Field | Type | Description |
|-------|------|-------------|
| `id` | UUID | Local unique identifier |
| `vaultId` | String | Vault that initiated the tx |
| `type` | Enum | `send` or `swap` |
| `status` | Enum | `broadcasted`, `pending`, `confirmed`, `failed` |
| `chain` | String | Source chain |
| `timestamp` | ISO 8601 | When broadcast |
| `txHash` | String | On-chain transaction hash |
| `explorerUrl` | String | Block explorer link |
| `fiatValue` | String? | USD value at broadcast time |

**Send-specific:** `fromAddress`, `toAddress`, `amount`, `token`, `tokenLogo`, `feeEstimate`, `memo`

**Swap-specific:** `fromToken`, `fromAmount`, `fromChain`, `fromTokenLogo`, `toToken`, `toAmount`, `toChain`, `toTokenLogo`, `provider`, `route`

**Status metadata:** `confirmedAt`, `failureReason`, `lastCheckedAt`

---

## Storage (Android)

- **Engine:** Room (SQLite) - consistent with existing persistence patterns
- **Location:** App internal storage (not external, not cache)
- **Retention:** Unlimited, no auto-pruning
- **Migration:** Room auto-migration for future schema changes
- **Schema:** Single Room entity with `vaultId` as indexed column. DAO provides queries filtered by vault, type, chain, status, date range.

---

## Recording Flow

1. User confirms send/swap
2. Transaction signed + broadcast
3. Write tx record to local storage with `status = broadcasted`
4. Dynamic complete screen shows (existing)
5. Complete screen polls + updates status (existing)
6. On status change (confirmed/failed): update local record

**Integration point:** Write immediately after successful broadcast, before navigating to the complete screen. Complete screen's status callback updates the history record via shared DAO.

---

## Status Tracking

**State machine:** `broadcasted` -> `pending` -> `confirmed` / `failed`

**During complete screen:** Existing dynamic complete screen resolves status. On resolution, update history record.

**After leaving complete screen (pending txs):** Refresh on app open only.
- On `Lifecycle.Event.ON_RESUME`: query pending records, fire status checks
- On history screen composition: same
- Swipe-to-refresh: same
- Reuse existing chain status check providers from dynamic tx status feature
- Coroutine-based concurrent checks (one per pending tx, structured concurrency)
- No background polling. No push notifications. No polling when offline.

---

### History List Screen
- Scrollable list using Compose lazy column
- Tab bar: `Overview` | `Swaps` | `Send` using custom tab row matching Figma (underline active indicator)
- Transaction cards as custom composables matching Figma card designs
- Date section headers as inline items or sticky headers
- Swipe-to-refresh triggers pending tx status check

### Card Views (follow Figma exactly)

**In-Progress Card (swap):**
- Elevated visual treatment (border/elevation)
- Green "Swap" badge pill, "In progress... [elapsed]" with progress bar
- Token icons, amounts, "min. payout" label, route pill

**Resolved Card (send or swap):**
- Type badge pill (blue Send, green Swap)
- Status: "Successful" (green) or "Error" (red)
- Token icon + fiat value + token amount
- Address pill "to [truncated address]"

### Detail View
Tapping a card navigates to the existing dynamic complete screen composable, passing tx data:
- `confirmed`/`failed`: Static resolved state
- `broadcasted`/`pending`: Active state, resumes polling

### Filtering
- **Primary:** Tab bar (Overview / Swaps / Send)
- **Secondary:** Chain, token, date range filters

### Empty State
"No transactions yet. Your sends and swaps will appear here."

---

## Key Considerations

- **Offline:** History readable offline. Status refresh only when network available.
- **TalkBack:** Content descriptions on all card elements
- **Configuration changes:** State preserved via ViewModel across rotation
- **Dark theme:** Material3 theming adapted to match Figma dark theme
- **Performance:** Room queries with proper indexing (vaultId, type, chain, timestamp) for 1000+ records
- **Thread safety:** Room writes on IO dispatcher, reads via Flow observation

---

## Deliverables

- [ ] Room entity + DAO for transaction history
- [ ] Transaction recording repository (write on broadcast, update on status change)
- [ ] History list screen (tabs, cards, date grouping, filters, swipe-to-refresh)
- [ ] Detail view integration (navigate to dynamic complete screen with tx data)
- [ ] Pending status refresh on app resume / history screen entry

aminsato:
## Description

Fixes #3215

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Full Transaction History screen (Overview/Swap/Send tabs), grouped-by-date list, detailed sheet with explorer links, provider/fee display, asset search & filters, and vault → history navigation.

* **Live Updates**
  * Live tracking of broadcast/pending statuses with pull-to-refresh and background refresh of pending transactions.

* **Chores**
  * Persistent local storage for transaction history added with database migrations and support for recording/updating transactions.

* **Localization**
  * Transaction History translations added for multiple languages.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -20,6 +20,8 @@ import com.vultisig.wallet.ui.screens.swap.VerifySwapScreen
 import com.vultisig.wallet.ui.screens.v2.home.components.CameraButton
 import com.vultisig.wallet.ui.screens.v2.home.components.TransactionType
 import com.vultisig.wallet.ui.screens.v2.home.components.TransactionTypeButton
+import com.vultisig.wallet.ui.screens.v2.home.pager.banner.UpgradeBanner
+import com.vultisig.wallet.ui.screens.v2.home.pager.container.HomePagePagerContainer
 import com.vultisig.wallet.ui.theme.OnBoardingComposeTheme
 
 class PreviewActivity : ComponentActivity() {
@@ -32,6 +34,7 @@ class PreviewActivity : ComponentActivity() {
                     "swap_confirm" -> SwapConfirmPreview()
                     "transaction_type_button" -> TransactionTypeButtonPreview()
                     "camera_button" -> CameraButton(onClick = {})
+                    "banner" -> BannerPreview()
                     else -> SwapConfirmPreview()
                 }
             }
@@ -47,6 +50,11 @@ private fun TransactionTypeButtonPreview() {
         TransactionTypeButton(txType = TransactionType.RECEIVE, isSelected = false)
         TransactionTypeButton(txType = TransactionType.BUY, isSelected = false)
     }
+}
+
+@Composable
+private fun BannerPreview() {
+    HomePagePagerContainer { UpgradeBanner {} }
 }
 
 @Composable
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/home/components/TransactionTypeButton.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/home/components/TransactionTypeButton.kt
@@ -41,7 +41,8 @@ fun TransactionTypeButton(
 ) {
 
     val backgroundColor =
-        if (isSelected) Theme.v2.colors.buttons.tertiary else Theme.v2.colors.backgrounds.tertiary_2
+        if (isSelected) Theme.v2.colors.buttons.ctaPrimary
+        else Theme.v2.colors.backgrounds.tertiary_2
 
     val (logo, title) =
         when (txType) {
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/home/pager/container/HomePagePagerContainer.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/home/pager/container/HomePagePagerContainer.kt
@@ -36,9 +36,9 @@ internal fun HomePagePagerContainer(
                 size = VsCircleButtonSize.Custom(size = 40.dp),
                 type =
                     VsCircleButtonType.Custom(
-                        color = Theme.v2.colors.neutrals.n100.copy(alpha = 0.2f)
+                        color = Theme.v2.colors.neutrals.n100.copy(alpha = 0.05f)
                     ),
-                modifier = Modifier.align(alignment = Alignment.TopEnd).padding(all = 8.dp),
+                modifier = Modifier.align(alignment = Alignment.TopEnd).padding(all = 7.dp),
             )
         }
     }


rkokhatskyi (commented):
## Code Review

### Critical

**Migration version mismatch** in `MIGRATIONS.kt`. The constant `MIGRATION_29_30` is declared as `Migration(26, 27)` but the database version was bumped from 29 to 30. This will silently skip the migration on real devices and crash with `IllegalStateException` when Room detects the missing table. Needs to be `Migration(29, 30)`.

**`uiState` in `TransactionHistoryViewModel` is a public `MutableStateFlow`**. The UI layer can mutate ViewModel state directly. Expose it as `StateFlow` via `asStateFlow()` and keep the mutable backing field private. This is the standard pattern used across the codebase.

### Architecture

**`TransactionHistoryScreen.kt` is 1411 lines.** This single file contains the screen, 10+ composables, a bottom sheet, card variants, helpers, and 6 previews. Split it into at least: `TransactionHistoryScreen.kt`, `TransactionDetailBottomSheet.kt`, `SendTransactionCard.kt`, `SwapTransactionCard.kt`, and shared components. Large files make reviews harder and increase merge conflict surface.

**Wide entity table design.** `TransactionHistoryEntity` packs send fields and swap fields into one table with ~15 nullable columns. When `type = SEND` all swap columns are null and vice versa. Consider either two separate tables (`send_transaction_history` / `swap_transaction_history`) or store the type specific payload as a JSON string column. The current layout will accumulate dead columns as more transaction types are added.

**ViewModel injects raw `Context`** via `@ApplicationContext`. The project should have a string resource provider abstraction. Injecting `Context` into a ViewModel couples it to Android framework and makes unit testing harder since you need `Robolectric` or a mock context for every string resolution.

### Correctness

**Asset filtering uses `ticker` as the unique key.** `toggleAssetSelection` and `removeAssetFilter` operate on `ticker` alone. Two tokens with the same ticker on different chains (e.g. USDC on Ethereum vs USDC on Solana) will be treated as one filter entry. Use a composite key like `"$ticker:$chain"`.

**`String.abbreviateAddress()` will crash on short strings.** `take(6)` and `takeLast(4)` are safe individually but the function returns `first6...last4` even for a 5 character string giving nonsensical output like `abcde...bcde`. Add a length guard: return the original string if length <= 10.

**`TransactionResult.NotFound` maps to `FAILED`** in `toDbModel()`. A "not found" result could mean the transaction hasn't propagated to the node yet (normal for 1 to 2 blocks after broadcast). Mapping it permanently to `FAILED` means a slow broadcast gets marked as failed and never rechecked. Consider keeping it as `PENDING` or adding a `NOT_FOUND` status.

**No error handling in `saveTransactionHistory`** inside `KeysignViewModel`. If the DB insert throws (e.g. unique constraint on `txHash` from a retry) it will propagate up and crash the keysign flow. Wrap it in try/catch or use `runCatching`. Transaction history saving must never break the signing path.

### Threading and Concurrency

**`Dispatchers.IO` hardcoded** in `RefreshPendingTransactionsUseCaseImpl`. Inject the dispatcher via constructor (`@IoDispatcher dispatcher: CoroutineDispatcher`) for testability. Hardcoded dispatchers make unit tests either flaky or slow.

**`transactionHistoryData` is a plain mutable `var`** in both `KeysignFlowViewModel` and `JoinKeysignViewModel`. It is written during transaction model mapping and read later during keysign. If `updateTransactionUiModel()` runs on a background thread while `transactionHistoryData` is read on another there is a data race. Either use `AtomicReference`, a `Mutex`, or pass it through a `StateFlow`.

### Style and Best Practices

**Mappers bound as `@Singleton`** in `MappersModule`. These mappers are stateless pure functions. Singleton scope keeps them in memory for the entire app lifecycle for no benefit. Remove the `@Singleton` annotation and let Hilt create them on demand.

**`SwapTransactionToHistoryDataMapper` extends `SuspendMapperFunc` but the implementation body does not suspend.** Use the non suspending `MapperFunc` interface instead to avoid unnecessary coroutine overhead at call sites.

**`// TransactionStatus.kt` comment** sitting inside `TransactionEntity.kt` on line 16. This looks like a leftover from when the enum was in its own file. Remove it.

**Unused DAO methods.** `observeFiltered`, `observeByDateRange`, and `deleteByVault` are defined in `TransactionHistoryDao` but never called from the repository or anywhere else. Remove dead code or defer adding it until it is needed.

aminsato:
@coderabbitai resume

PR #3726, opened 2026-03-23
fix: exclude current vault from duplicate name check in VaultRenameViewModel

### LINKED ISSUE #3715: VaultRenameViewModel rejects the vault's own current name
## Description

`validateName` checks `vaultRepository.getAll().any { it.name == newName }` which matches the vault being renamed. This means if the user opens the rename screen, doesn't change the name, and clicks save, they get a "this name already exists" error for their own vault's name.

## File
`app/src/main/java/com/vultisig/wallet/ui/models/VaultRenameViewModel.kt:77-86`

## Expected Behavior
Check should exclude the current vault: `it.name == newName && it.id != vaultId`.

## Actual Behavior
Error message "This name already exists" shown for the vault's own name.

## Suggested Fix
Add `&& it.id != vaultId` to the duplicate name check.

---

rkokhatskyi:
## Summary
- Adds `&& it.id != vaultId` to the duplicate name check in `validateName` so users can save a vault without changing its name
- Previously, tapping save with the vault's current name would incorrectly show "This name already exists"

## Test plan
- [ ] Open vault rename screen, tap save without changing name — should succeed
- [ ] Rename vault to another vault's name — should still show error

Closes #3715

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected vault rename validation so a vault can retain its current name during renames while still preventing true duplicate names across different vaults.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultRenameViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultRenameViewModel.kt
@@ -79,8 +79,7 @@ constructor(
         if (!isVaultNameValid(newName)) {
             return StringResource(R.string.vault_name_too_long_error)
         }
-        val isNameAlreadyExist =
-            vaultRepository.getAll().any { it.name == newName && it.id != vaultId }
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #3729, opened 2026-03-23
fix: prevent keysign auto-start before data loading completes

### LINKED ISSUE #3717: Race condition in keysign auto-start on peer discovery
## Description

In `KeysignPeerDiscovery.kt`, three independent `LaunchedEffect` blocks run concurrently: one loads data (line 51), one discovers peers (line 79), one checks threshold and starts keysign (line 87). They have no ordering guarantee.

For fast vaults with server, peer discovery can complete before `setData` finishes loading the transaction model. Keysign starts with null/incomplete transaction data, leading to missing data on the completion screen.

## File
`app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt:87-98`

## Expected Behavior
Keysign should only start after both data loading AND peer discovery are complete.

## Actual Behavior
If participant discovery is very fast (e.g., fast vault with server), keysign can start before transaction model is loaded.

## Suggested Fix
Add a `dataLoaded` flag to the threshold check condition so keysign doesn't start until `setData` has completed.

---

rkokhatskyi:
## Summary
- Add `isDataLoaded` flag to `KeysignFlowViewModel` to track when `setData` has finished loading transaction data
- Guard the threshold-check `LaunchedEffect` in `KeysignPeerDiscovery` so keysign only auto-starts after both the peer threshold is met AND data loading is complete
- Prevents race condition where keysign could start before vault/payload data was ready

## Test plan
- [ ] Verify keysign works for normal vaults (multi-device)
- [ ] Verify fast vault keysign still works (server-assisted)
- [ ] Verify transaction details appear correctly on the done screen

Closes #3717

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of the keysign process by ensuring all transaction data is fully loaded before proceeding to the signing stage.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
@@ -91,6 +91,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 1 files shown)


johnnyluo:
@coderabbitai review please

PR #3784, opened 2026-03-24
fix: correct secured asset detection for THORChain token swaps

realpaaao:
## Summary

- Secured asset swaps (e.g., ETH-USDC on THORChain) failed with `insufficient funds` because the MsgDeposit protobuf was missing the `secured` flag and had incorrect chain/symbol fields
- Root cause: `isSecuredAsset()` used a hardcoded ticker allowlist (`BTC, ETH, BCH, LTC, DOGE, AVAX, BNB`) that excluded all ERC-20/token secured assets (USDC, USDT, WBTC, DAI, etc.)
- Fix: replace hardcoded list with on-chain denom pattern matching (`contractAddress` contains `-` and doesn't start with `x/`), extract L1 chain and full symbol (including contract address) from the denom

## Context

THORChain's protobuf `Asset` message has a `bool secured = 5` field that must be set for secured asset deposits. The chain field must be the L1 origin (e.g., `ETH` not `THOR`), and the symbol must include the contract address for ERC-20 tokens (e.g., `USDC-0XA0B86991...` not just `USDC`).

On-chain secured asset denoms follow the pattern `chain-ticker` (L1 native) or `chain-ticker-contract` (ERC-20):
- `btc-btc` → chain: BTC, symbol: BTC
- `eth-usdc-0xa0b86991...` → chain: ETH, symbol: USDC-0XA0B86991...

## Changes

| File | Change |
|------|--------|
| `Coin.kt` | Replace hardcoded `SECURE_ASSETS_TICKERS` check with denom pattern match. Add `securedAssetChain()` and `securedAssetSymbol()` helpers |
| `ThorchainSwapHelper.kt` | Use `securedAssetChain()` for chain name and `securedAssetSymbol()` for symbol in swap deposit |
| `ThorChainHelper.kt` | Same chain/symbol fix in the deposit message path |

## Test plan

- [x] Verified ETH-USDC secured swap succeeds on-chain ([tx](https://thorchain.net/tx/36F7A5614D706A0A9870C890D55885D58E67F8C5EC3C09114453EB09B58455FA))
- [ ] Verify L1 secured asset swaps (BTC, ETH, DOGE) still work (identical output — no behavior change)
- [ ] Verify non-secured swaps (RUNE, cross-chain) are unaffected

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable detection of THORChain secured assets using stricter eligibility and contract-address pattern rules.
  * Consistent derivation of THORChain chain name and asset symbol in swap and deposit flows for secured assets.

* **New Features**
  * Introduced an explicit secured-asset eligibility check to gate secure deposit and memo-based detection logic.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/models/Coin.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/models/Coin.kt
@@ -60,12 +60,20 @@ fun Coin.getNotNativeTicker(): String {
     return this.ticker.uppercase().removePrefix("X/")
 }
 
+// THORChain-side: is this coin a secured asset token on THORChain itself
 fun Coin.isSecuredAsset(): Boolean {
     if (chain != Chain.ThorChain) return false
     if (isNativeToken) return false
     if (contractAddress.startsWith("x/", ignoreCase = true)) return false
     val parts = contractAddress.split("-", limit = 2)
     return parts.size == 2 && parts[0].isNotBlank() && parts[1].isNotBlank()
+}
+
+// Source-chain side: can this coin be deposited into THORChain as a SECURE+ asset
+fun Coin.isSecuredAssetEligible(): Boolean {
+    val eligibleTickers = listOf("BTC", "ETH", "BCH", "LTC", "DOGE", "AVAX", "BNB")
+    return eligibleTickers.contains(ticker.uppercase()) &&
+        (isNativeToken || contractAddress == "${ticker.lowercase()}-${ticker.lowercase()}")
 }
 
 fun Coin.securedAssetChain(): String {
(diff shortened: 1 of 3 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/models/Coin.kt
```diff
@@ -62,6 +61,17 @@ fun Coin.getNotNativeTicker(): String {
 }
 
 fun Coin.isSecuredAsset(): Boolean {
-    return SECURE_ASSETS_TICKERS.contains(ticker.uppercase()) &&
-        (isNativeToken || contractAddress == "${ticker.lowercase()}-${ticker.lowercase()}")
+    if (chain != Chain.ThorChain) return false
```
rkokhatskyi: This guard breaks two callers that pass source chain coins (BTC on Bitcoin, ETH on Ethereum etc):

- `DepositFormViewModel.createSecuredAssetTransaction()` at line 977 calls `isSecuredAsset()` on the source chain token to validate before building the `SECURE+` memo. With this change it will always return false and throw `deposit_error_not_secured_asset`
- `JoinKeysignViewModel` at line 814 uses `isSecuredAsset()` on `payload.coin` in the `else` branch (non-THORChain `blockChainSpecific`). This is how incoming SECURE+ deposit keysign requests from paired devices get detected as deposits. Will always be false now

You need two separate functions here: one for detecting THORChain-side secured tokens (this new logic) and one for validating source chain eligibility (the old ticker list check)

PR #4003, opened 2026-04-05
Redesign QR scanner screen UI

### LINKED ISSUE #3314: [ADD] Add new QR code scanner UI
# [Add] New QR code scanner UI [QR Screen]

## Problem
The current QR code scanner screen uses outdated UI design that doesn't match the latest Figma specifications. Users are missing the updated visual design and tooltip functionality that should be consistent with other parts of the app.

## Expected Behavior
The QR scanner screen should display the new UI design according to Figma specifications, including an information icon in the top right that triggers a tooltip when pressed, matching the tooltip behavior used elsewhere in the app.

## Solution
Update the QR scanner screen UI components to match the provided Figma designs while preserving all existing functionality. Implement the information icon with tooltip trigger mechanism consistent with the app's existing tooltip patterns, ensuring visual consistency across the application.

### Must Do
- [ ] Update QR scanner screen layout to match Figma design (https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=66720-110197&t=wU9eygQrVhaciass-4)
- [ ] Add information icon in top right corner with tooltip functionality
- [ ] Implement tooltip display matching the Figma specification (https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=66720-110647&t=wU9eygQrVhaciass-4)

### Must NOT Do
- Don't modify any QR scanning functionality or logic
- Don't change the underlying QR code processing behavior

## Acceptance Criteria
- [ ] QR scanner screen displays new UI design matching Figma specifications
- [ ] Information icon in top right corner triggers tooltip when pressed
- [ ] Tooltip behavior is consistent with other tooltips in the app

## Context
- Reported by **paaao** (@paaao_ on Discord) on Discord

cc @realpaaao

aminsato:
Closes #3314

## Summary
- Redesign QR scanner screen to match new Figma design with ModalBottomSheet, gradient background, and viewport border styling
- Add HintBox tooltip component with tip/info text and dismiss functionality
- Add ErrorView for camera permission denied state
- Update hint text typography to match Figma specs (Satoshi 13sp, Bold/Medium, 0.2sp letter spacing)
- Add localized strings for hint title and message across all 10 locale files

## Test plan
- [ ] Open QR scanner and verify the new layout renders correctly
- [ ] Tap the info button and verify the hint tooltip appears with correct text and styling
- [ ] Dismiss the hint tooltip by tapping outside or the dismiss button
- [ ] Deny camera permission and verify the ErrorView is shown
- [ ] Upload a QR code from gallery and verify it scans correctly
- [ ] Verify the scanner works correctly when camera permission is granted

## Summary by CodeRabbit

* **New Features**
  * Redesigned QR scan screen with Material3 bottom sheet, top info button, dedicated bottom bar, rounded camera viewport with blurred/shadow frame, center overlay, and preview.
  * Interactive hint/help popup with new custom-content variant; ViewModel toggle to show tips.

* **Bug Fixes**
  * Improved camera-permission handling with an “Open Settings” action; safer camera lifecycle management.
  * Gallery scanning now performs a retry and returns clear success/failure results.

* **UI/UX**
  * Error view now accepts an optional button model for primary actions.
  * Added localized QR upload/troubleshooting strings and new scan-related theme colors.


johnnyluo (approved):
LGTM

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
```diff
@@ -126,28 +169,49 @@ internal fun ScanQrScreen(
...
-                                resultBarcodes
+                        val first = scanImage(InputImage.fromFilePath(context, uri))
+                        val firstBarcodes = (first as? ScanResult.Success)?.barcodes.orEmpty()
+                        if (firstBarcodes.isNotEmpty()) {
+                            onSuccess(firstBarcodes)
+                        } else {
+                            val bitmap =
+                                requireNotNull(uriToBitmap(context.contentResolver, uri))
+                                    .addWhiteBorder(2F)
+                            val retry = scanImage(InputImage.fromBitmap(bitmap, 0))
```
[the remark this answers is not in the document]
aminsato: The bitmap + addWhiteBorder retry pass was already present before this PR (visible in the `-` lines of this same hunk). Only the surrounding flow was refactored to use the ScanResult sealed type; scanner behavior is unchanged.

app/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -534,6 +534,9 @@
 
     <!-- Scan QR -->
     <string name="scan_qr_upload_from_gallery">上传二维码</string>
+    <string name="scan_qr_upload_qr_code">上传二维码</string>
```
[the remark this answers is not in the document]
aminsato: Fixed in d517f624f — `scan_qr_upload_from_gallery` is now `从相册上传` so the two buttons are distinguishable.

johnnyluo:
@aminsato please fix the conflict

aminsato:
> @aminsato please fix the conflict

Done

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
```diff
@@ -160,76 +231,233 @@ internal fun ScanQrScreen(
...
-                        if (isFrameHighlighted) {
-                            painterResource(id = R.drawable.vs_camera_frame_highlight)
-                        } else {
-                            painterResource(id = R.drawable.vs_camera_frame)
+            }
+        } else if (
+            cameraPermissionState.status.shouldShowRationale ||
+                cameraPermissionState.status.isGranted.not()
+        ) {
+            LaunchedEffect(Unit) { cameraPermissionState.launchPermissionRequest() }
```
johnnyluo: **Issue 3 — Permission-request intent unclear**

`LaunchedEffect(Unit)` fires both when the rationale should be shown and when permission is permanently denied. On Android 11+ `launchPermissionRequest()` is a no-op on permanent denial so it's harmless, but the intent is ambiguous. Add a guard or comment to clarify:

```kotlin
// Only request if we can still ask; on permanent denial the user must go to Settings
if (!cameraPermissionState.status.shouldShowRationale) {
    LaunchedEffect(Unit) { cameraPermissionState.launchPermissionRequest() }
}
```

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
```diff
@@ -160,76 +231,233 @@ internal fun ScanQrScreen(
...
-                Image(
-                    modifier = Modifier.align(Alignment.Center).fillMaxWidth().padding(40.dp),
-                    painter =
-                        if (isFrameHighlighted) {
-                            painterResource(id = R.drawable.vs_camera_frame_highlight)
-                        } else {
-                            painterResource(id = R.drawable.vs_camera_frame)
+            }
+        } else if (
+            cameraPermissionState.status.shouldShowRationale ||
```
johnnyluo: **Issue 1 — ErrorView and bottom bar are both visible on permission denied**

When camera permission is denied, `ErrorView` renders with its own "Open Settings" button, but `ScanQrLayout` still renders `ScanQrBottomBar` (which contains an "Upload QR Code" button). Both will be visible simultaneously.

`showReturnButton = isCameraGranted.not()` already gates the return button, but the Upload QR button is always shown. Consider passing a flag to hide the bottom bar entirely when showing `ErrorView`, or move the `ErrorView` outside of `ScanQrLayout`'s content so it replaces the scaffold instead of sitting inside it.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
```diff
@@ -160,76 +231,233 @@ internal fun ScanQrScreen(
...
+
+        val hintBoxWidthPx = hintBoxWidth.roundToPx()
+        val spacePx = space.roundToPx()
+        HintBox(
+            modifier = Modifier.width(hintBoxWidth),
+            isVisible = uiModel.isTipVisible,
+            isPointerTriangleOnTop = true,
+            offset =
+                IntOffset(
+                    x = (infoButtonBottomEnd.x - containerPosition.x).toInt() - hintBoxWidthPx,
```
johnnyluo, line 332: **Issue 2 — HintBox offset may be wrong during bottom sheet animation**

Both `infoButtonBottomEnd` and `containerPosition` use `positionInWindow()`, which is correct for the settled state. However, if the `ModalBottomSheet` is still animating when these are measured, `containerPosition` may capture an intermediate position, causing the hint box to appear offset.

Consider using `positionInParent()` for the info button coordinates instead — that gives you the offset relative to the parent `Box` directly, removing the need to subtract `containerPosition` entirely:

```kotlin
// In onInfoButtonPositioned:
val pos = coordinates.positionInParent()
onInfoButtonPositioned(Offset(x = pos.x + size.width, y = pos.y + size.height))

// Then offset becomes:
x = infoButtonBottomEnd.x.toInt() - hintBoxWidthPx,
y = infoButtonBottomEnd.y.toInt() + spacePx,
```

app/src/main/res/values/strings.xml
```diff
@@ -377,6 +378,9 @@
     <string name="share_vault_qr_address" translatable="false">vultisig.com</string>
     <string name="import_file_password_hint_text">Hint: %s</string>
     <string name="scan_qr_upload_from_gallery">Upload QR</string>
```
johnnyluo: **Issue 4 — `scan_qr_upload_from_gallery` may be orphaned**

The bottom bar now uses `scan_qr_upload_qr_code` as the button label. Please verify `scan_qr_upload_from_gallery` is still referenced somewhere:

```bash
grep -r "scan_qr_upload_from_gallery" app/src/
```

If it has no remaining usages, it should be removed from all 10 locale `strings.xml` files per the project deletion checklist.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
```diff
@@ -66,32 +105,42 @@ import com.vultisig.wallet.ui.utils.uriToBitmap
...
             onError = viewModel::handleError,
+            onMoreInfo = viewModel::toggleShowTip,
         )
     }
 }
 
-@OptIn(ExperimentalPermissionsApi::class)
 @Composable
-internal fun ScanQrScreen(
+private fun ScanQrScreen(
```
johnnyluo, line 138: **Issue 5 — `@OptIn(ExperimentalPermissionsApi::class)` annotation removed**

The `@OptIn(ExperimentalPermissionsApi::class)` annotation that existed on the old `internal fun ScanQrScreen` has been dropped. This function uses `rememberPermissionState` from accompanist-permissions, which is still experimental. Please confirm this compiles without warnings — if it does, add the annotation to be explicit about the opt-in.

johnnyluo:
@aminsato could you please fix those few comments?

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
aminsato: Tried switching to `positionInParent()` but it doesn't work here — the info button is nested inside `Scaffold > topBar`, so its parent-relative coordinates don't map to the outer `Box` where `HintBox` is placed. The hint box ends up in the wrong position:

hint box mispositioned

Keeping the original `positionInWindow()` approach with `containerPosition` subtraction since both use the same window coordinate space.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
aminsato: No changes needed — the file already has `@file:OptIn(ExperimentalPermissionsApi::class)` on line 1, which covers all functions in the file.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
aminsato: Fixed in 28bde922d — added a `showBottomBar` parameter to `ScanQrLayout`, set to `isCameraGranted`. When permission is denied, only the `ErrorView` with the "Open Settings" button is shown; the bottom bar (with "Upload QR Code") is hidden.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
aminsato: Done

aminsato:
> @aminsato could you please fix those few comments?

Fixed

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
```diff
@@ -123,31 +173,60 @@ internal fun ScanQrScreen(
...
-        Box {
-            if (cameraPermissionState.status.isGranted) {
+    val isCameraGranted = cameraPermissionState.status.isGranted
+
+    ScanQrLayout(
+        uiModel = uiModel,
+        showReturnButton = isCameraGranted.not(),
+        showBottomBar = isCameraGranted,
+        onDismiss = onDismiss,
+        onUploadQr = {
```
rkokhatskyi, line 222: `showBottomBar = isCameraGranted` hides the entire bottom bar — including the "Upload QR Code" button — when camera permission is denied. In the old code the gallery upload button was always visible regardless of permission status. Now a user who denied camera permission has no way to upload a QR from their gallery; the only option is "Open Settings". If this is unintentional, move the upload button outside `ScanQrBottomBar` so it renders unconditionally.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
```diff
@@ -160,76 +239,238 @@ internal fun ScanQrScreen(
...
-                            painterResource(id = R.drawable.vs_camera_frame_highlight)
-                        } else {
-                            painterResource(id = R.drawable.vs_camera_frame)
+            }
+        } else if (
+            cameraPermissionState.status.shouldShowRationale ||
+                cameraPermissionState.status.isGranted.not()
+        ) {
+            if (!cameraPermissionState.status.shouldShowRationale) {
+                LaunchedEffect(Unit) { cameraPermissionState.launchPermissionRequest() }
```
rkokhatskyi: The `else if` condition `shouldShowRationale || !isGranted` is always true here because `!isGranted` was already established by the enclosing `else` branch — the outer `if (isCameraGranted)` already excludes the granted case. This makes the `else if` an unconditional branch. Simplify to a plain `else { ... }` to remove the dead predicate.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
```diff
@@ -160,76 +239,238 @@ internal fun ScanQrScreen(
...
+            }
+            ErrorView(
```
rkokhatskyi: On first launch (`!shouldShowRationale && !isGranted`) this `LaunchedEffect` fires the system permission dialog **and** `ErrorView` renders "Camera permission not granted" simultaneously — before the user has declined anything. Only show `ErrorView` after the dialog returns a denied result. On the initial request, either show nothing (let the dialog be the sole UI) or show a neutral explainer, then transition to the error state only once permission is confirmed denied.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
```diff
@@ -287,17 +530,39 @@ private fun QrCameraScreen(
...
+private sealed interface ScanResult {
+    data class Success(val barcodes: List<Barcode>) : ScanResult
+
+    data class Failure(val message: String) : ScanResult
+}
+
+private suspend fun scanImage(inputImage: InputImage): ScanResult =
+    suspendCancellableCoroutine { continuation ->
         try {
             createScanner()
```
rkokhatskyi: `createScanner()` returns a `BarcodeScanning.getClient()` instance that implements `Closeable`. A new scanner is allocated on every `scanImage` call and is never closed. Wrap it in `use { }` or close it in a `finally` block:

```kotlin
val scanner = createScanner()
try {
    scanner.process(inputImage)
        .addOnSuccessListener { ... }
        .addOnFailureListener { ... }
    throw e
} catch (e: Exception) {
    ...
} finally {
    scanner.close()
}
```
Note: `close()` must be called after the Task completes, not before — so hook it into the `addOnCompleteListener` rather than a simple `finally` if the Task is async.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
aminsato: Good catch — removed the `showBottomBar` parameter entirely so the bottom bar (including "Upload QR Code") is always visible regardless of camera permission status.

app/src/main/res/values/strings.xml
aminsato: Good catch! Verified — `scan_qr_upload_from_gallery` has already been removed from all locale `strings.xml` files and has zero remaining references in the codebase. This is resolved.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
aminsato: Good point — simplified to a plain `else` since the outer `if (isCameraGranted)` already excludes the granted case. Fixed.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
aminsato: Good catch — fixed. Now we track `hasRequestedPermission` state. On first launch (before any response), only the system permission dialog shows (no `ErrorView`). The `ErrorView` only renders after the user has denied the permission or if `shouldShowRationale` is true.

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
aminsato: Fixed — the scanner is now stored in a local variable and closed via `addOnCompleteListener` after the Task finishes. Also added `scanner.close()` in `catch` blocks for the synchronous exception paths.

aminsato:
@coderabbitai full review

app/src/main/java/com/vultisig/wallet/ui/screens/scan/ScanQrScreen.kt
```diff
@@ -123,31 +174,59 @@ internal fun ScanQrScreen(
...
+                                is ScanResult.Success ->
+                                    if (retry.barcodes.isNotEmpty()) {
+                                        onSuccess(retry.barcodes)
+                                    } else {
+                                        onError(context.getString(R.string.no_barcodes_found))
+                                    }
+                                is ScanResult.Failure ->
+                                    onError(
+                                        (first as? ScanResult.Failure)?.message ?: retry.message
+                                    )
```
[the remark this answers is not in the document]
aminsato: Fixed — raw diagnostic text is now logged via Timber and the user sees the localized `no_barcodes_found` string instead.

PR #4037, opened 2026-04-12
fix: remove unreachable code in Server.kt

aminsato:
## Summary
- Remove redundant hash mismatch check in `getPayload` — `postPayload` already validates integrity before caching
- Replace `?.let { return } ?: run` patterns with straightforward `if/else` early returns in `getPayload`, `getStartKeygenOrKeysign`, `getCompleteKeySign`, and `getSession`
- Eliminate duplicate `NotFound` responses in nested `?.let` chains

## Test plan
- [ ] Verify keygen flow works end-to-end with local mediator
- [ ] Verify keysign flow works end-to-end with local mediator
- [ ] Confirm no regressions in session, message, and payload endpoints

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal cache handling to improve reliability of session, payload, and signature retrieval.
* **Bug Fixes**
  * More consistent "Not Found" and mismatch responses when requested data is missing or invalid, reducing ambiguous outcomes for requests.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/mediator/Server.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/mediator/Server.kt
@@ -132,11 +132,16 @@ class Server(private val nsdManager: NsdManager) : NsdManager.RegistrationListen
             response.status(HttpStatusCode.BadRequest.value)
             return "hash is empty"
         }
-        val cached = cache[hash]
-        if (cached != null) {
+        val content = cache[hash] as? String
+        if (content != null) {
+            val contentHash = content.sha256()
+            if (contentHash != hash) {
+                response.status(HttpStatusCode.BadRequest.value)
+                return "hash mismatch"
+            }
             Timber.d("return hash: %s", hash)
             response.status(HttpStatusCode.OK.value)
-            return (cached as? String).toString()
+            return content
         }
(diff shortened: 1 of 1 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/mediator/Server.kt
```diff
@@ -132,18 +132,13 @@ class Server(private val nsdManager: NsdManager) : NsdManager.RegistrationListen
...
         }
-        cache[hash]?.let {
-            val content = it as? String
-            val contentHash = content?.sha256()
-            if (contentHash != hash) {
-                response.status(HttpStatusCode.BadRequest.value)
-                return "hash mismatch"
-            }
-
+        val cached = cache[hash]
```
johnnyluo: This check is intended , we need to make sure the content and hash match , otherwise user can trick the server

johnnyluo (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/mediator/Server.kt
aminsato: Done

johnnyluo (approved):
LGTM

PR #4039, opened 2026-04-12
fix: use PBKDF2 with salt for vault backup encryption

### LINKED ISSUE #3992: bug: Vault backup encryption uses raw SHA-256(password) with no salt or KDF
## Bug Description

In `Encryption.kt`, the AES encryption key is derived by directly hashing the password with SHA-256 — no salt, no PBKDF2, no Argon2, no KDF of any kind.

```kotlin
val keyBytes = messageDigest.digest(password)  // raw SHA-256, no salt
```

**Files:**
- `data/src/main/kotlin/com/vultisig/wallet/data/usecases/Encryption.kt:26-28` (encrypt)
- `data/src/main/kotlin/com/vultisig/wallet/data/usecases/Encryption.kt:40-41` (decrypt)
- `data/src/main/kotlin/com/vultisig/wallet/data/usecases/Encryption.kt:68-73` (generateKey)

## Impact

Stolen vault backup files can be cracked offline with significantly reduced cost. A single SHA-256 hash is trivially fast to brute-force compared to PBKDF2 or Argon2 with proper iteration counts.

## Expected Behavior

Use PBKDF2 (or Argon2) with a random salt and high iteration count (e.g., 600,000+ for PBKDF2-HMAC-SHA256).

## Source

Combined functional bug audit 2026-04-03, item #16 (Critical)

aminsato:
## Summary
- Replace raw SHA-256(password) key derivation with **PBKDF2-HMAC-SHA256** (600,000 iterations, random 16-byte salt) for vault backup encryption
- Create dedicated `VaultBackupEncryption` interface so TSS messaging encryption is **unaffected**
- Old encrypted backups (SHA-256 GCM and legacy CBC) remain fully decryptable via automatic fallback

Closes #3992

## Test plan
- [x] New PBKDF2 encrypt/decrypt roundtrip test passes
- [x] Legacy GCM backup decryption test passes (backward compat)
- [x] Wrong password correctly fails decryption
- [x] Output format validated (version byte + salt + IV + ciphertext)
- [ ] Manual test: create encrypted backup, re-import it
- [ ] Manual test: import an old encrypted backup created before this change

## Summary by CodeRabbit

* **New Features**
  * Stronger vault backup encryption with a versioned, PBKDF2‑AES format and seamless fallback for existing backups.

* **Performance**
  * Backup and import operations now run heavy encryption/decryption on a background dispatcher for smoother UI responsiveness.

* **Tests**
  * Added end‑to‑end tests for PBKDF2 encryption/decryption, legacy fallback, versioned output, header invariants, and wrong‑password handling.


johnnyluo (changes requested):
1. Can you make sure this new encrypt / decrypt method is backward compatible , make sure old backup can be imported 
2. Make sure cross device support , android backup with this new encryption can be imported successfully on IOS / Windows / Extension

aminsato (commented):
1. Yes, it's backward compatible — decrypt falls back to the legacy SHA256-based AES-GCM format for existing backups.
2. Newly created backups use PBKDF2 with a version byte (`0x02`), which iOS and Windows/extension can't decrypt yet. I'd suggest holding this PR until all platforms implement PBKDF2-HMAC-SHA256 decryption, so newly created backups remain usable across all devices.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,99 @@
...
+
+        val cipher = Cipher.getInstance(AES_GCM_NO_PADDING)
+        cipher.init(Cipher.ENCRYPT_MODE, keySpec, GCMParameterSpec(GCM_TAG_BITS, iv))
+        val encryptedData = cipher.doFinal(data)
+
+        return byteArrayOf(VERSION_PBKDF2) + salt + iv + encryptedData
+    }
+
+    override fun decrypt(data: ByteArray, password: ByteArray): ByteArray? {
+        if (data.isEmpty() || data[0] != VERSION_PBKDF2) {
```
rkokhatskyi: **Version-byte collision breaks ~1/256 existing encrypted backups.**

The legacy AES-GCM format is `[12-byte random IV][ciphertext+tag]`. The first byte of the IV is uniformly random, so `data[0] == 0x02` occurs with probability 1/256 for any existing encrypted backup. When that happens, this branch misidentifies the legacy payload as PBKDF2, `decryptPbkdf2` throws `AEADBadTagException`, which is caught and returns `null` — the legacy fallback is never reached. The size guard `data.size < PBKDF2_HEADER_SIZE + GCM_TAG_BYTES` provides zero protection because all real vault protobuf payloads are well above 45 bytes.

This silently bricks ~0.4% of existing encrypted backups on import with no user-visible error beyond a generic decryption failure.

The version marker must be distinguishable from a random GCM IV byte. The safest fix is a multi-byte sentinel prefix that cannot appear at the start of a legacy payload:

```kotlin
// Replace the single-byte VERSION_PBKDF2 with a 4-byte magic prefix:
private val PBKDF2_MAGIC = byteArrayOf(0x56, 0x4C, 0x54, 0x02) // "VLT\x02"
private const val PBKDF2_MAGIC_SIZE = 4

// encrypt: prepend magic instead of single byte
return PBKDF2_MAGIC + salt + iv + encryptedData

// decrypt: check for magic prefix before treating as PBKDF2
val isPbkdf2 = data.size >= PBKDF2_MAGIC_SIZE &&
    data.copyOfRange(0, PBKDF2_MAGIC_SIZE).contentEquals(PBKDF2_MAGIC)
if (!isPbkdf2) return legacyEncryption.decrypt(data, password)
```

The probability of a 4-byte collision with a random IV prefix is 1 in 4 billion, which is acceptable.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,99 @@
...
+
+        val cipher = Cipher.getInstance(AES_GCM_NO_PADDING)
+        cipher.init(Cipher.DECRYPT_MODE, keySpec, GCMParameterSpec(GCM_TAG_BITS, iv))
+        return cipher.doFinal(encryptedData)
+    }
+
+    private fun deriveKey(password: ByteArray, salt: ByteArray): ByteArray {
+        val keySpec =
+            PBEKeySpec(
+                String(password, Charsets.UTF_8).toCharArray(),
```
rkokhatskyi: **PBKDF2 on `Dispatchers.Main` will freeze the UI and risk ANR.**

`encrypt` and `decrypt` are non-suspend blocking functions. PBKDF2 with 600,000 iterations takes 1–3 seconds on a mid-range Android device. All three callers (`BackupPasswordViewModel.backupCurrentVault`, `BackupVaultViewModel.backupCurrentVault`, `ImportFileViewModel.saveToDb`) run inside `viewModelScope.launch { }` which uses `Dispatchers.Main` — so this computation runs on the UI thread.

The fix belongs at the call sites, wrapping the use-case call in `withContext(Dispatchers.Default)`:

```kotlin
// In BackupPasswordViewModel / BackupVaultViewModel:
val backup = withContext(Dispatchers.Default) {
    createVaultBackup(mapVaultToProto(vault), password)
}

// In ImportFileViewModel.saveToDb:
private suspend fun saveToDb(fileContent: String, password: String?): SaveResult =
    withContext(Dispatchers.Default) {
        try {
            insertVaultToDb(parseVaultFromString(fileContent, password))
            SaveResult.Success
        } catch (e: ...) { ... }
    }
```

Alternatively, make `encrypt`/`decrypt` suspend and enforce the dispatcher inside the use-case itself.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,99 @@
...
+import java.security.SecureRandom
+import javax.crypto.Cipher
+import javax.crypto.SecretKeyFactory
+import javax.crypto.spec.GCMParameterSpec
+import javax.crypto.spec.PBEKeySpec
+import javax.crypto.spec.SecretKeySpec
+import javax.inject.Inject
+import timber.log.Timber
+
+interface VaultBackupEncryption {
```
rkokhatskyi, line 14: `VaultBackupEncryption` is declared `public` (no visibility modifier) while its only implementation `Pbkdf2AesEncryption` is correctly `internal`. The interface itself should also be `internal` to prevent the `data` module from accidentally leaking this type into callers outside the module boundary.

```kotlin
internal interface VaultBackupEncryption {
    fun encrypt(data: ByteArray, password: ByteArray): ByteArray
    fun decrypt(data: ByteArray, password: ByteArray): ByteArray?
}
```

data/src/test/kotlin/com/vultisig/wallet/data/usecases/Pbkdf2AesEncryptionTest.kt
```diff
@@ -0,0 +1,77 @@
...
+        assertNull(decrypted)
+    }
+
+    @Test
+    fun `decrypts legacy GCM format`() {
+        val encryptedBase64 = "zPMOwnPVMFKMf9LOIFkyqBOr8AC1SIdQ34Ruk5gmRqxZ+lIyK7zM5/1NUjXlAg=="
+        val decrypted =
+            pbkdf2Aes.decrypt(encryptedBase64.decodeBase64Bytes(), password.toByteArray())
+        assertNotNull(decrypted)
+        assertEquals(originalInput, decrypted.toString(Charsets.UTF_8))
```
rkokhatskyi, line 58: The existing test verifies that a known-good legacy GCM blob decrypts correctly, but it does not cover the most dangerous regression introduced by this PR: a legacy encrypted backup whose first byte happens to be `0x02`.

Please add a test that manufactures a legacy ciphertext with `data[0] == 0x02` and confirms `decrypt` returns the correct plaintext (not `null`):

```kotlin
@Test
fun `decrypts legacy GCM backup whose first IV byte is 0x02`() {
    // Encrypt with legacy to get a real ciphertext, then patch the first byte to 0x02
    // to simulate the collision case.
    var legacyCiphertext: ByteArray? = null
    while (legacyCiphertext == null || legacyCiphertext[0] != 0x02.toByte()) {
        legacyCiphertext = legacyAes.encrypt(
            originalInput.toByteArray(Charsets.UTF_8),
            password.toByteArray()
        )
    }
    val decrypted = pbkdf2Aes.decrypt(legacyCiphertext, password.toByteArray())
    assertNotNull(decrypted)
    assertEquals(originalInput, decrypted.toString(Charsets.UTF_8))
}
```

This test will fail with the current single-byte version marker, proving the regression before the fix is applied.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,99 @@
...
+        val keySpec = SecretKeySpec(keyBytes, AES)
+
+        val iv = ByteArray(IV_LENGTH)
+        random.nextBytes(iv)
+
+        val cipher = Cipher.getInstance(AES_GCM_NO_PADDING)
+        cipher.init(Cipher.ENCRYPT_MODE, keySpec, GCMParameterSpec(GCM_TAG_BITS, iv))
+        val encryptedData = cipher.doFinal(data)
+
+        return byteArrayOf(VERSION_PBKDF2) + salt + iv + encryptedData
```
rkokhatskyi: Agree with johnnyluo's comment: any vault backup created after this PR is merged will be unreadable by iOS, Windows desktop, and the browser extension — all of which still use the legacy SHA-256/AES-GCM format.

This is a real data-loss scenario: a user creates a backup on Android, their Android device is lost or broken, and they attempt to recover on iOS or via the extension — they cannot. The backup is inaccessible.

This PR should remain open until all platforms implement PBKDF2-HMAC-SHA256 decryption simultaneously, or until feature flag is added

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,110 @@
...
+import java.security.SecureRandom
+import javax.crypto.Cipher
+import javax.crypto.SecretKeyFactory
+import javax.crypto.spec.GCMParameterSpec
+import javax.crypto.spec.PBEKeySpec
+import javax.crypto.spec.SecretKeySpec
+import javax.inject.Inject
+import timber.log.Timber
+
+interface VaultBackupEncryption {
```
rkokhatskyi, line 14: The interface is `public` (default) while `Pbkdf2AesEncryption` is `internal`. This leaks `VaultBackupEncryption` outside the `data` module boundary. Change to `internal interface`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,110 @@
...
+
+        val cipher = Cipher.getInstance(AES_GCM_NO_PADDING)
+        cipher.init(Cipher.DECRYPT_MODE, keySpec, GCMParameterSpec(GCM_TAG_BITS, iv))
+        return cipher.doFinal(encryptedData)
+    }
+
+    private fun deriveKey(password: ByteArray, salt: ByteArray): ByteArray {
+        val keySpec =
+            PBEKeySpec(
+                String(password, Charsets.UTF_8).toCharArray(),
```
rkokhatskyi: The `toCharArray()` intermediate is not zeroed after `clearPassword()` runs. `clearPassword()` zeros the copy inside `PBEKeySpec` but the `CharArray` on the heap lives until GC. Capture it in a named variable and `fill('\u0000')` in the finally block alongside `clearPassword()`. Minor given Android's process isolation but this is a password path.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
aminsato: Good catch in principle — but `VaultBackupEncryption` is consumed from the `:app` module by `CreateVaultBackupUseCase` and `ParseVaultFromStringUseCase`, so marking the interface `internal` breaks compilation across the module boundary. This matches the existing `Encryption` interface pattern in the same file (public interface, `internal` impl). Happy to move those two use cases into `:data` as a follow-up if we want to tighten the boundary.

johnnyluo:
@aminsato can you please check all the comments , if it had been fixed please resolve it

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
aminsato: Done

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
aminsato: Done

aminsato:
> @aminsato can you please check all the comments , if it had been fixed please resolve it

Resolved

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,106 @@
...
+        val keyBytes = deriveKey(password, salt)
+        val keySpec = SecretKeySpec(keyBytes, AES)
+
+        val cipher = Cipher.getInstance(AES_GCM_NO_PADDING)
+        cipher.init(Cipher.DECRYPT_MODE, keySpec, GCMParameterSpec(GCM_TAG_BITS, iv))
+        return cipher.doFinal(encryptedData)
+    }
+
+    private fun deriveKey(password: ByteArray, salt: ByteArray): ByteArray {
+        val passwordChars = String(password, Charsets.UTF_8).toCharArray()
```
rkokhatskyi: `String(password, Charsets.UTF_8)` materialises the plaintext password as an immutable `String` on the heap before `.toCharArray()` runs. `fill('\u0000')` on `passwordChars` and `keySpec.clearPassword()` don't touch that String, so the password still lingers until GC. If we care enough to zero the char array we should decode `password` straight to `CharArray` via `Charsets.UTF_8.newDecoder().decode(ByteBuffer.wrap(password))`

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,106 @@
...
+            null
+        } catch (e: IllegalArgumentException) {
+            Timber.e(e, "Invalid PBKDF2 vault backup payload")
+            null
+        }
+    }
+
+    private fun isPbkdf2(data: ByteArray): Boolean =
+        data.size >= PBKDF2_MAGIC_SIZE &&
+            data.copyOfRange(0, PBKDF2_MAGIC_SIZE).contentEquals(PBKDF2_MAGIC)
```
rkokhatskyi: `copyOfRange(0, 4)` allocates a fresh 4-byte array on every decrypt. cheap but pointless, compare the four bytes directly against `PBKDF2_MAGIC` indices

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,106 @@
...
+import java.security.SecureRandom
+import javax.crypto.Cipher
+import javax.crypto.SecretKeyFactory
+import javax.crypto.spec.GCMParameterSpec
+import javax.crypto.spec.PBEKeySpec
+import javax.crypto.spec.SecretKeySpec
+import javax.inject.Inject
+import timber.log.Timber
+
+interface VaultBackupEncryption {
```
rkokhatskyi, line 14: cross-platform concern from earlier is still open, any backup produced by this change is unreadable by iOS/desktop/extension which still sit on the old format, worth calling out in the PR description or gating behind a feature flag until the other platforms ship PBKDF2 decode

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,106 @@
...
+        return PBKDF2_MAGIC + salt + iv + encryptedData
+    }
+
+    override fun decrypt(data: ByteArray, password: ByteArray): ByteArray? {
+        if (!isPbkdf2(data)) {
+            return legacyEncryption.decrypt(data, password)
+        }
+        if (data.size < PBKDF2_HEADER_SIZE + GCM_TAG_BYTES) {
+            Timber.e("Invalid PBKDF2 vault backup payload")
+            return null
```
rkokhatskyi, line 49: no test covers this branch, add one that feeds a payload starting with `VLT\x02` but shorter than `PBKDF2_HEADER_SIZE + GCM_TAG_BYTES` and asserts `null`

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,106 @@
...
+interface VaultBackupEncryption {
+    fun encrypt(data: ByteArray, password: ByteArray): ByteArray
+
+    fun decrypt(data: ByteArray, password: ByteArray): ByteArray?
+}
+
+internal class Pbkdf2AesEncryption @Inject constructor(private val legacyEncryption: Encryption) :
+    VaultBackupEncryption {
+
+    private val random = SecureRandom()
```
rkokhatskyi, line 23: `SecureRandom()` with no explicit algorithm picks the platform default, fine on Android but explicit `SecureRandom.getInstance("SHA1PRNG")` or the modern `SecureRandom.getInstanceStrong()` on API 26+ makes the choice obvious

data/src/test/kotlin/com/vultisig/wallet/data/usecases/Pbkdf2AesEncryptionTest.kt
```diff
@@ -0,0 +1,91 @@
...
+        assertTrue(encrypted.size >= 32 + 17)
+    }
+
+    @Test
+    fun `decrypts legacy GCM backup whose first IV byte is 0x02`() {
+        var legacyCiphertext: ByteArray
+        do {
+            legacyCiphertext =
+                legacyAes.encrypt(originalInput.toByteArray(Charsets.UTF_8), password.toByteArray())
+        } while (legacyCiphertext[0] != 0x02.toByte())
```
rkokhatskyi: `do`/`while` keeps hammering `legacyAes.encrypt` until the IV happens to start with `0x02`, roughly 256 iterations on average but with no bound. if the legacy encrypt ever changes and the first byte can't be `0x02` this loops forever, cap the attempts and fail the test with a clear message

app/src/main/java/com/vultisig/wallet/ui/models/BackupPasswordViewModel.kt
```diff
@@ -266,7 +266,8 @@ constructor(
             return false
         }
 
-        val backup = createVaultBackup(mapVaultToProto(vault), password)
+        val backup =
+            withContext(Dispatchers.Default) { createVaultBackup(mapVaultToProto(vault), password) }
```
rkokhatskyi, line 270: `withContext(Dispatchers.Default)` is good but `createVaultBackup` itself is a plain (non-suspend) call, consider marking the use case `suspend` and moving the dispatcher choice inside, so every caller doesn't have to remember to wrap it

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
aminsato: Fixed in f34ed8f3c. `deriveKey` now decodes the password bytes straight into a `CharBuffer` via `Charsets.UTF_8.newDecoder().decode(ByteBuffer.wrap(password))`, copies out a `CharArray`, and zeros both the working array and the `CharBuffer`'s backing array in the `finally` block. No intermediate `String` on the heap.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
aminsato: Fixed in f34ed8f3c. `isPbkdf2` now compares the four magic bytes in place against `PBKDF2_MAGIC[i]`, no throwaway allocation per decrypt.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
aminsato: Acknowledged — this is still a product-level decision rather than a code fix. Holding this PR until iOS / desktop / extension ship PBKDF2 decode, or until we agree on a feature-flag rollout. Happy to add a flag here if you'd prefer that path; please advise.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
aminsato: Added in f34ed8f3c: `returns null for PBKDF2 payload smaller than header plus tag` feeds a `VLT\x02`-prefixed payload of 14 bytes and asserts `decrypt` returns `null`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
aminsato: Leaving the no-arg `SecureRandom()` for now. On Android the default provider is `AndroidOpenSSL` backed by `/dev/urandom`, which is the recommended choice; `SecureRandom.getInstanceStrong()` can block on `/dev/random` on some devices. Happy to add an inline comment calling out the deliberate choice if that helps readers.

data/src/test/kotlin/com/vultisig/wallet/data/usecases/Pbkdf2AesEncryptionTest.kt
aminsato: Fixed in f34ed8f3c. The loop is now bounded to 10,000 attempts and the test fails with a clear message if no legacy ciphertext starting with `0x02` is produced in that window.

app/src/main/java/com/vultisig/wallet/ui/models/BackupPasswordViewModel.kt
aminsato: Agreed that pushing the dispatcher into the use case is cleaner. Leaving it for a follow-up so this PR stays focused on the PBKDF2 change — the refactor touches `CreateVaultBackupUseCase` + `ParseVaultFromStringUseCase` and all three call sites. I'll open a separate PR for that.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
aminsato: Can't make this `internal` without a wider refactor: `CreateVaultBackupUseCase` and `ParseVaultFromStringUseCase` both live in the `:app` module and inject `VaultBackupEncryption` directly. Tightening the visibility means moving those use cases (and their bindings) into `:data` first. I'd rather do that in a dedicated PR — let me know if you want me to follow up immediately after this one lands.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    override fun encrypt(data: ByteArray, password: ByteArray): ByteArray {
+        val salt = ByteArray(SALT_LENGTH)
+        random.nextBytes(salt)
+
+        val keyBytes = deriveKey(password, salt)
+        val keySpec = SecretKeySpec(keyBytes, AES)
```
rkokhatskyi, line 30: `passwordChars` is zeroed in `deriveKey`'s finally block but `keyBytes` isn't cleared after `SecretKeySpec` copies it — add `keyBytes.fill(0)` right after the spec is created, same issue at line 75 in `decryptPbkdf2`

data/src/test/kotlin/com/vultisig/wallet/data/usecases/Pbkdf2AesEncryptionTest.kt
```diff
@@ -0,0 +1,110 @@
...
+
+        val decrypted = pbkdf2AesNoLegacy.decrypt(encrypted, "wrongpassword".toByteArray())
+        assertNull(decrypted)
+    }
+
+    @Test
+    fun `decrypts legacy GCM format`() {
+        val encryptedBase64 = "zPMOwnPVMFKMf9LOIFkyqBOr8AC1SIdQ34Ruk5gmRqxZ+lIyK7zM5/1NUjXlAg=="
+        val decrypted =
+            pbkdf2Aes.decrypt(encryptedBase64.decodeBase64Bytes(), password.toByteArray())
```
rkokhatskyi: `io.ktor.util.decodeBase64Bytes` is an HTTP-client utility pulled in as a transitive dep — `java.util.Base64.getDecoder().decode(encryptedBase64)` achieves the same without the accidental Ktor dependency

data/src/main/kotlin/com/vultisig/wallet/data/usecases/VaultBackupEncryption.kt
aminsato: Fixed in 306db2551. `keyBytes.fill(0)` now runs immediately after `SecretKeySpec` copies them in both `encrypt` and `decryptPbkdf2`.

data/src/test/kotlin/com/vultisig/wallet/data/usecases/Pbkdf2AesEncryptionTest.kt
aminsato: Fixed in 306db2551. Swapped `io.ktor.util.decodeBase64Bytes` for `java.util.Base64.getDecoder().decode(...)` — drops the transitive Ktor dep from the unit test.

johnnyluo (approved):
LGTM

PR #4043, opened 2026-04-13
feat: add ability to pin and manually sort chains

### LINKED ISSUE #2710: [ADD] Ability to pin/favorite/manually sort
Add the ability to favorite cryptocurrencies in your portfolio so they are pinned to the top of the list. Also the ability to manually sort the order (hold and drag, or click an up/down arrow)

aminsato:
## Summary
- Adds a new **Sort Chains** screen accessible via a sort icon button in the home page tab menu
- Users can **pin** favorite chains to always appear at the top of the portfolio list
- Users can **drag-and-drop reorder** chains within pinned and unpinned sections
- Custom order and pin state is persisted per vault in Room DB (new `accountOrder` table with migration 32→33)
- Pinned/custom order is applied reactively on the home screen via Flow combine

Closes #2710

## Test plan
- [x] Unit tests for `SortAccountsViewModel` (13 tests covering load, pin/unpin, reorder, save, bounds checking, error handling)
- [x] Unit tests for `AccountOrderRepository` (4 tests covering delegation, save order, empty state)
- [ ] Open app → home screen → tap sort icon → verify Sort Chains screen opens
- [ ] Pin a chain → verify it moves to Pinned section
- [ ] Unpin a chain → verify it moves to Other section
- [ ] Long-press drag to reorder within pinned section → verify order changes
- [ ] Long-press drag to reorder within unpinned section → verify order changes
- [ ] Tap Done → verify order persists on home screen (pinned chains on top)
- [ ] Switch vaults → verify each vault has independent sort/pin state
- [ ] Fresh install / upgrade → verify migration 32→33 runs without crash

## Summary by CodeRabbit

* **New Features**
  * Account sorting screen with drag-and-drop reorder, pin/unpin per chain, and Done to save.

* **Persistence**
  * New local ordering persistence with database schema v33 and migration support.

* **UI & Resources**
  * Added sort and pin icons; localized strings added for multiple languages; sort button in accounts toolbar.

* **Navigation**
  * New route to open the account-sorting screen.

* **Tests**
  * Unit tests covering sorting, pinning, persistence, and navigation.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
@@ -19,6 +19,7 @@ import javax.inject.Inject
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.update
+import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.coroutines.launch
 import timber.log.Timber
 
@@ -61,6 +62,8 @@ constructor(
             val addresses =
                 try {
                     accountsRepository.loadAddresses(vaultId, false).first()
+                } catch (e: CancellationException) {
+                    throw e
                 } catch (e: Exception) {
                     Timber.e(e, "Failed to load addresses for sorting")
                     return@launch
@@ -165,6 +168,8 @@ constructor(
             try {
                 accountOrderRepository.saveOrders(vaultId, orders)
                 navigator.back()
+            } catch (e: CancellationException) {
+                throw e
             } catch (e: Exception) {
                 Timber.e(e, "Failed to save account order")
             }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
@@ -50,6 +50,8 @@ constructor(
 
     val uiState = MutableStateFlow(SortAccountsUiModel())
 
+    private var isLoaded = false
+
     init {
         loadAccounts()
     }
@@ -88,6 +90,7 @@ constructor(
                     .sortedBy { orderMap[it.chainId]?.order ?: Float.MAX_VALUE }
 
             uiState.update { it.copy(pinnedAccounts = pinned, unpinnedAccounts = unpinned) }
+            isLoaded = true
         }
     }
 
@@ -129,6 +132,11 @@ constructor(
 
     fun save() {
         viewModelScope.launch {
+            if (!isLoaded) {
+                Timber.w("Cannot save account order: data not loaded yet")
+                return@launch
+            }
+
             val state = uiState.value
             val orders = mutableListOf<AccountOrderEntity>()
 
@@ -156,10 +164,10 @@ constructor(
 
             try {
                 accountOrderRepository.saveOrders(vaultId, orders)
+                navigator.back()
             } catch (e: Exception) {
                 Timber.e(e, "Failed to save account order")
             }
-            navigator.back()
         }
     }
 
--- a/app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
@@ -1,0 +1,169 @@
+package com.vultisig.wallet.ui.models
+
+import androidx.annotation.DrawableRes
+import androidx.compose.runtime.Immutable
+import androidx.lifecycle.SavedStateHandle
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import androidx.navigation.toRoute
+import com.vultisig.wallet.data.db.models.AccountOrderEntity
+import com.vultisig.wallet.data.models.logo
+import com.vultisig.wallet.data.repositories.AccountOrderRepository
+import com.vultisig.wallet.data.repositories.AccountsRepository
+import com.vultisig.wallet.ui.navigation.Destination
+import com.vultisig.wallet.ui.navigation.Navigator
+import com.vultisig.wallet.ui.navigation.Route
+import com.vultisig.wallet.ui.navigation.back
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+import timber.log.Timber
+
+@Immutable
+internal data class SortAccountItemUiModel(
+    val chainId: String,
+    val chainName: String,
+    @param:DrawableRes val logo: Int,
+    val isPinned: Boolean,
+)
+
+@Immutable
+internal data class SortAccountsUiModel(
+    val pinnedAccounts: List<SortAccountItemUiModel> = emptyList(),
+    val unpinnedAccounts: List<SortAccountItemUiModel> = emptyList(),
+)
+
+@HiltViewModel
+internal class SortAccountsViewModel
+@Inject
+constructor(
+    savedStateHandle: SavedStateHandle,
+    private val navigator: Navigator<Destination>,
+    private val accountsRepository: AccountsRepository,
+    private val accountOrderRepository: AccountOrderRepository,
+) : ViewModel() {
+
+    private val vaultId: String = savedStateHandle.toRoute<Route.SortAccounts>().vaultId
+
+    val uiState = MutableStateFlow(SortAccountsUiModel())
+
+    init {
+        loadAccounts()
+    }
+
+    private fun loadAccounts() {
+        viewModelScope.launch {
+            val addresses =
+                try {
+                    accountsRepository.loadAddresses(vaultId, false).first()
+                } catch (e: Exception) {
+                    Timber.e(e, "Failed to load addresses for sorting")
+                    return@launch
+                }
+            val savedOrders = accountOrderRepository.loadOrders(vaultId).first()
+
+            val orderMap = savedOrders.associateBy { it.chain }
+
+            val items =
+                addresses.map { address ->
+                    val saved = orderMap[address.chain.raw]
+                    SortAccountItemUiModel(
+                        chainId = address.chain.raw,
+                        chainName = address.chain.raw,
+                        logo = address.chain.logo,
+                        isPinned = saved?.isPinned == true,
+                    )
+                }
+
+            val pinned =
+                items
+                    .filter { it.isPinned }
+                    .sortedBy { orderMap[it.chainId]?.order ?: Float.MAX_VALUE }
+            val unpinned =
+                items
+                    .filter { !it.isPinned }
+                    .sortedBy { orderMap[it.chainId]?.order ?: Float.MAX_VALUE }
+
+            uiState.update { it.copy(pinnedAccounts = pinned, unpinnedAccounts = unpinned) }
+        }
+    }
+
+    fun movePinned(from: Int, to: Int) {
+        uiState.update {
+            if (from !in it.pinnedAccounts.indices || to !in it.pinnedAccounts.indices)
+                return@update it
+            val list = it.pinnedAccounts.toMutableList().apply { add(to, removeAt(from)) }
+            it.copy(pinnedAccounts = list)
+        }
+    }
+
+    fun moveUnpinned(from: Int, to: Int) {
+        uiState.update {
+            if (from !in it.unpinnedAccounts.indices || to !in it.unpinnedAccounts.indices)
+                return@update it
+            val list = it.unpinnedAccounts.toMutableList().apply { add(to, removeAt(from)) }
+            it.copy(unpinnedAccounts = list)
+        }
+    }
+
+    fun togglePin(item: SortAccountItemUiModel) {
+        uiState.update { state ->
+            if (item.isPinned) {
+                val updated = item.copy(isPinned = false)
+                state.copy(
+                    pinnedAccounts = state.pinnedAccounts.filter { it.chainId != item.chainId },
+                    unpinnedAccounts = listOf(updated) + state.unpinnedAccounts,
+                )
+            } else {
+                val updated = item.copy(isPinned = true)
+                state.copy(
+                    pinnedAccounts = state.pinnedAccounts + updated,
+                    unpinnedAccounts = state.unpinnedAccounts.filter { it.chainId != item.chainId },
+                )
+            }
+        }
+    }
+
+    fun save() {
+        viewModelScope.launch {
+            val state = uiState.value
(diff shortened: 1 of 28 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/db/models/AccountOrderEntity.kt
```diff
@@ -0,0 +1,12 @@
+package com.vultisig.wallet.data.db.models
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+
+@Entity(tableName = "accountOrder", primaryKeys = ["vaultId", "chain"])
+data class AccountOrderEntity(
```
rkokhatskyi, line 21: The migration SQL in MIGRATION_32_33 adds `FOREIGN KEY(vaultId) REFERENCES vault(id) ON DELETE CASCADE`, but this `@Entity` has no matching `@ForeignKey` declaration. Room validates the migration-produced schema against the entity-derived schema at startup — the two DDLs differ, causing an `IllegalStateException` for users upgrading from v32 to v33. New installs are unaffected (Room creates the table directly from `@Entity` without FK). Every other entity with a vault FK in this codebase (e.g., `CoinEntity`) declares it via `@Entity(foreignKeys = [...])`. Either add the matching `@ForeignKey` annotation to `@Entity`, or remove the `FOREIGN KEY` clause from the migration SQL to match.

app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
```diff
@@ -0,0 +1,177 @@
...
+        loadAccounts()
+    }
+
+    private fun loadAccounts() {
+        viewModelScope.launch {
+            val addresses =
+                try {
+                    accountsRepository.loadAddresses(vaultId, false).first()
+                } catch (e: Exception) {
+                    Timber.e(e, "Failed to load addresses for sorting")
```
rkokhatskyi, line 68: `catch (e: Exception)` catches `CancellationException` (a subtype of `Exception`), logs it, and returns without rethrowing. This breaks structured concurrency — if the coroutine is cancelled (e.g., ViewModel cleared mid-load), the cancellation is silently swallowed. Add a rethrow:
```kotlin
    throw e
} catch (e: Exception) {
    Timber.e(e, "Failed to load addresses for sorting")
    return@launch
}
```

app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
```diff
@@ -0,0 +1,177 @@
...
+                        isPinned = false,
+                    )
+                )
+            }
+
+            try {
+                accountOrderRepository.saveOrders(vaultId, orders)
+                navigator.back()
+            } catch (e: Exception) {
+                Timber.e(e, "Failed to save account order")
```
rkokhatskyi, line 174: Same `catch (e: Exception)` swallows `CancellationException` in `save()`. Add `catch (e: CancellationException) { throw e }` before the broad catch, or use `catch (e: Exception) { if (e is CancellationException) throw e; Timber.e(...) }`.

app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
```diff
@@ -0,0 +1,177 @@
...
+
+            val orderMap = savedOrders.associateBy { it.chain }
+
+            val items =
+                addresses.map { address ->
+                    val saved = orderMap[address.chain.raw]
+                    SortAccountItemUiModel(
+                        chainId = address.chain.raw,
+                        chainName = address.chain.raw,
+                        logo = address.chain.logo,
```
rkokhatskyi, line 81: `chainName` is set to `address.chain.raw` (the internal identifier, e.g., `"THORChain"`, `"Bitcoin-Cash"`), not a human-readable display name. Users will see raw IDs in the Sort Chains screen. If there is a `displayName` or similar property on `Chain`, use it here instead.

app/src/main/java/com/vultisig/wallet/ui/screens/home/SortAccountsScreen.kt
```diff
@@ -0,0 +1,141 @@
...
+import com.vultisig.wallet.ui.components.v2.scaffold.V2Scaffold
+import com.vultisig.wallet.ui.models.SortAccountItemUiModel
+import com.vultisig.wallet.ui.models.SortAccountsUiModel
+import com.vultisig.wallet.ui.models.SortAccountsViewModel
+import com.vultisig.wallet.ui.theme.Theme
+
+@Composable
+internal fun SortAccountsScreen(viewModel: SortAccountsViewModel = hiltViewModel()) {
+    val state by viewModel.uiState.collectAsState()
+
```
rkokhatskyi, line 34: `collectAsState()` is lifecycle-unaware and continues collecting even when the composable is in the background. Use `collectAsStateWithLifecycle()` from `androidx.lifecycle:lifecycle-runtime-compose` instead, consistent with the rest of the codebase.

app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
aminsato: This is a false positive. `chain.raw` is the established display name pattern used consistently throughout the codebase — see `AddressToUiModelMapper` (`from.chain.raw`), `ChainTokensViewModel` (`chainName = chainRaw`), `ReceiveViewModel` (`name = it.chain.raw`), and `KeygenViewModel` (`chainName = chainSetting.chain.raw`). There is no separate `displayName` property on `Chain` — `raw` serves as both the identifier and display name.

data/src/main/kotlin/com/vultisig/wallet/data/db/models/AccountOrderEntity.kt
aminsato: Done

app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
aminsato: Done

app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
aminsato: Done

app/src/main/java/com/vultisig/wallet/ui/screens/home/SortAccountsScreen.kt
aminsato: Done

data/src/main/kotlin/com/vultisig/wallet/data/db/models/AccountOrderEntity.kt
```diff
@@ -0,0 +1,26 @@
+package com.vultisig.wallet.data.db.models
+
+import androidx.room.ColumnInfo
+import androidx.room.Entity
+import androidx.room.ForeignKey
+
+@Entity(
+    tableName = "accountOrder",
```
rkokhatskyi, line 8: **[MEDIUM] Missing `indices` on the ForeignKey column — inconsistent with CoinEntity**

`CoinEntity`, the only other entity in this project that declares a `@ForeignKey` against `vault.id`, also declares `indices = [Index(value = ["vaultId"])]` at the `@Entity` level. Without it Room emits a compile-time warning about missing index coverage for FK enforcement. Functionally the PK `(vaultId, chain)` already covers it at the SQLite level since `vaultId` is the leading column, so there is no runtime bug — but for consistency and to silence tooling warnings, please add:

```kotlin
indices = [Index(value = ["vaultId"])],
```

The migration SQL does not need to change.

app/src/main/java/com/vultisig/wallet/ui/models/SortAccountsViewModel.kt
```diff
@@ -0,0 +1,182 @@
...
+                    throw e
+                } catch (e: Exception) {
+                    Timber.e(e, "Failed to load addresses for sorting")
+                    return@launch
+                }
+            val savedOrders = accountOrderRepository.loadOrders(vaultId).first()
+
+            val orderMap = savedOrders.associateBy { it.chain }
+
+            val items =
```
rkokhatskyi, line 75: **[MEDIUM] `chainName` is still set to `address.chain.raw` — raw internal ID shown to the user**

This was flagged in our previous review and has not been fixed. `chain.raw` is the internal serialization ID (e.g. `"Bitcoin-Cash"`, `"THORChain"`, `"BSC"`, `"Zksync"`). These are machine-readable identifiers, not user-facing display names.

For comparison, look at how other parts of the app handle this — e.g. `AccountUiModel.chainName` is populated via `addressToUiModelMapper`, which applies proper display formatting. The Sort Chains screen shows `item.chainName` as the primary label in every list row, so users will see raw IDs like `"BSC"` instead of `"BNB Smart Chain"` or `"Bitcoin-Cash"` instead of `"Bitcoin Cash"`.

Fix: derive the display name from the same source used elsewhere in the app, or map `Chain` to a localized string resource. If there is no existing string-resource mapping for chain display names, `chain.raw.replace("-", " ")` is a minimal acceptable workaround, but a proper resource mapping is preferred.

app/src/main/java/com/vultisig/wallet/ui/screens/home/SortAccountsScreen.kt
```diff
@@ -0,0 +1,143 @@
...
+            style = Theme.brockmann.body.s.medium,
+            color = Theme.v2.colors.text.primary,
+            modifier = Modifier.weight(1f),
+        )
+
+        UiIcon(
+            drawableResId = R.drawable.ic_pin,
+            size = 20.dp,
+            tint =
+                if (item.isPinned) Theme.v2.colors.primary.accent4
```
rkokhatskyi, line 136: **[MEDIUM] Hardcoded accessibility strings in `contentDescription`**

```kotlin
contentDescription =
    if (item.isPinned) "Unpin ${item.chainName}" else "Pin ${item.chainName}",
```

These are raw English string literals embedded in Kotlin, not string resources. They will not be translated for any of the 9 non-English locales and will break TalkBack for non-English users.

Additionally, `item.chainName` is itself `chain.raw` (see the sibling comment on `SortAccountsViewModel`), so screen readers will announce e.g. "Pin Bitcoin-Cash" rather than "Pin Bitcoin Cash".

Fix: add string resources for the pin/unpin content descriptions (e.g. `sort_accounts_pin_chain` / `sort_accounts_unpin_chain` with a `%s` placeholder) and use `stringResource(R.string.sort_accounts_pin_chain, item.chainName)` here. Add these keys to all 10 locale files.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountOrderRepository.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.vultisig.wallet.data.db.dao.AccountOrderDao
+import com.vultisig.wallet.data.db.models.AccountOrderEntity
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlinx.coroutines.flow.Flow
+
+@Singleton
+class AccountOrderRepository @Inject constructor(private val accountOrderDao: AccountOrderDao) {
+
+    fun loadOrders(vaultId: String): Flow<List<AccountOrderEntity>> =
```
rkokhatskyi, line 12: **[MEDIUM] `AccountOrderEntity` leaks from the data layer into the presentation layer**

`AccountOrderRepository.loadOrders()` returns `Flow<List<AccountOrderEntity>>` and `saveOrders()` accepts `List<AccountOrderEntity>`. Both `SortAccountsViewModel` and `VaultAccountsViewModel` import `AccountOrderEntity` directly from `com.vultisig.wallet.data.db.models`. This breaks the clean-architecture boundary where the presentation layer should only depend on domain/repository models, not database entities.

The standard pattern in this codebase is: `data` module exposes repository interfaces and domain models; `app` module (ViewModels) maps to/from UI models via mappers. Compare `CoinEntity` — ViewModels never import it directly.

Fix: define a plain data class (e.g. `AccountOrder`) in the repository or domain layer, map to/from `AccountOrderEntity` inside `AccountOrderRepository`, and return `Flow<List<AccountOrder>>` to callers. `SortAccountsViewModel.save()` would then construct `AccountOrder` objects, not `AccountOrderEntity` objects.

rkokhatskyi (changes requested):
## AI Review after commits cf71fae, d87b6b7, 1557eb0

### Previous Findings — Status

| # | Finding | Status |
|---|---------|--------|
| 1 | CRITICAL: Room schema mismatch (`MIGRATION_32_33` vs `AccountOrderEntity`) | FIXED — `@ForeignKey` annotation added with matching `CASCADE` rules. |
| 2 | HIGH: `CancellationException` swallowed in `loadAccounts()` | FIXED — CE is rethrown explicitly. |
| 3 | HIGH: `CancellationException` swallowed in `save()` | FIXED — CE is rethrown explicitly. |
| 4 | MEDIUM: `chainName` uses `address.chain.raw` | NOT FIXED — still `chainName = address.chain.raw` in `SortAccountsViewModel.kt:75`. |
| 5 | MEDIUM: `collectAsState()` instead of `collectAsStateWithLifecycle()` | FIXED — `collectAsStateWithLifecycle()` used. |

### New Findings

**[MEDIUM] `AccountOrderEntity` leaks from the data layer into the presentation layer** (`AccountOrderRepository.kt:12`)
Both `SortAccountsViewModel` and `VaultAccountsViewModel` import `AccountOrderEntity` from `com.vultisig.wallet.data.db.models` directly. The repository API surface exposes a DB entity, not a domain model. The rest of the codebase maps entities to domain objects inside the repository layer — ViewModels never import `CoinEntity`, `VaultEntity`, etc. This is an architecture boundary violation. See inline comment for the fix.

**[MEDIUM] Missing `indices` on `AccountOrderEntity` ForeignKey** (`AccountOrderEntity.kt:8`)
`CoinEntity` — the only other entity with a `@ForeignKey` against `vault.id` — also declares `indices = [Index(value = ["vaultId"])]`. `AccountOrderEntity` is missing this. Functionally the PK `(vaultId, chain)` covers FK enforcement since `vaultId` is the leading column, but the annotation is absent, which is inconsistent and will cause Room to emit a compile-time warning. See inline comment for the one-line fix.

**[MEDIUM] Hardcoded `contentDescription` strings in `SortAccountsScreen.kt:136`**
`"Pin ${item.chainName}"` / `"Unpin ${item.chainName}"` are raw English string literals, bypassing the localization system. This breaks TalkBack for the 9 non-English locales. They must be moved to `strings.xml` (all 10 locale files) as parameterized resources. See inline comment.

### Verdict

Three MEDIUM issues remain open (one carried over from the previous review, two new). No CRITICAL or HIGH issues. The PR is close — the three MEDIUMs should be resolved before merge:
1. `chainName` must show a user-visible display name, not `chain.raw`.
2. `AccountOrderEntity` must not cross the repository boundary into ViewModels.
3. Pin/unpin `contentDescription` strings must be localized string resources.

PR #4054, opened 2026-04-13
feat: dApp signing — function-name title + collapsible details (Phase 1)

realpaaao:
## Summary
- For EVM contract calls, show the decoded function name (e.g. `Supply With Permit`) as a text title above the transaction details, replacing the misleading native-amount card. Falls back to the existing `VsOverviewToken` for plain sends and other chains.
- Collapsible "Transaction Details" section (chevron + AnimatedVisibility) for the raw function signature and decoded inputs.
- Suppress the native amount + fiat rows when a function name is present — contract calls typically send 0 ETH, so the duplicate "0 ETH" row was misleading.
- `VsOverviewToken` gains a `withContainer` param so it can be embedded without its own background/border (used by the verify screen).
- `VsHoldableButton`: drop the unused `holdDuration` parameter; animation duration is now hardcoded to the platform long-press timeout so the fill always finishes when `onLongClick` fires.
- `AbiToJson.parseBigIntAsJsonString`: catch `NumberFormatException` only and fall back to `JsonNull` instead of forwarding raw input — closes the lookalike-Unicode / RTL-override path into the rendered function-inputs box.

## Scope and follow-up
The resolved-amount hero (`HeroContent.Send` / `Swap`) and the explicit `HeroContent.Unverified` warning state ship with the Blockaid simulation work in #4306, which stacks on this branch. The 4byte-driven function-name caption in this PR becomes the small-text title above that hero.

## Platforms
Android part. Companion PRs:
- Extension: vultisig/vultisig-windows — `paaao/dapp-display-improvement`
- iOS: vultisig/vultisig-ios — `paaao/dapp_function_display`

## Test plan
- [ ] Plain ETH / USDC send → unchanged layout (`VsOverviewToken` hero, native amount + fiat rows)
- [ ] EVM contract call (e.g. Aave supply) → function-name title above the details, native "Amount" / "Value" rows suppressed
- [ ] EVM contract call where `decodeFunctionArgs` fails → function name still surfaces, function-inputs row is absent
- [ ] Collapsible "Transaction Details" expands/collapses; long signatures + inputs scroll via the outer scaffold
- [ ] Verify-side hold-to-confirm: animation fill completes when `onLongClick` fires
- [ ] `./gradlew :app:assembleDebug :app:testDebugUnitTest :data:testDebugUnitTest`

## Summary by CodeRabbit

* **New Features**
  * Display smart contract function names as the transaction "hero"
  * Expandable transaction details section showing function signatures and inputs

* **Refactor**
  * Redesigned long-press button interaction for more reliable press/hold behavior
  * Improved handling/serialization of large numeric values in contract data
  * Transaction overviews now suppress native amount when showing contract action details


realpaaao:
@coderabbitai review

realpaaao:
https://github.com/coderabbitai review

realpaaao:
@coderabbitai review

johnnyluo:
could you please attach some screenshot of how it looks like here?

realpaaao:
@johnnyluo 
image

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1283,14 +1300,63 @@ constructor(
...
-    private suspend fun getTransactionFunctionInfo(memo: String?, chain: Chain): FunctionInfo? {
+    private suspend fun getTransactionFunctionInfo(
+        memo: String?,
+        chain: Chain,
+        toAddress: String?,
+    ): FunctionInfo? {
         if (chain.standard != TokenStandard.EVM || memo.isNullOrEmpty()) return null
 
-        val functionSignature = fourByteRepository.decodeFunction(memo)
+        val functionSignature = fourByteRepository.decodeFunction(memo) ?: return null
```
rkokhatskyi: This is the trust boundary — everything downstream treats `functionSignature` as authoritative, but it came from `FourByteApi` which hits `4byte.directory` with `ordering=created_at` + `firstOrNull()`. 4byte is crowdsourced, anyone can register a signature for a selector, first-submitter-wins.

Before this PR that didn't matter much since the decoded sig was a footer row. Now it drives the hero on the signing screen, so if an attacker front-runs submission for a selector not yet in 4byte, their fake signature gets shown as "Supply 100 USDC" on every device.

Also `FourByteApi` has no TLS pinning and no timeout on the call — it can hang the verify screen.

Options:
- Gate the hero on a local allow-list of vetted ABIs by 4-byte selector, and keep 4byte only for the collapsible details
- Or keep 4byte but label the hero "Decoded (unverified)" and never hide the native `msg.value`

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
```diff
@@ -0,0 +1,168 @@
...
+    }
+    if (current.isNotEmpty()) {
+        parts.add(current.toString().trim())
+    }
+    return parts
+}
+
+object ContractCallExtractor {
+
+    private val registry: Map<String, ExtractionStrategy> =
```
rkokhatskyi: The registry keys on bare function name, but `withdraw` means very different things across protocols:

- Aave V3: `withdraw(address asset, uint256 amount, address to)` — first address is the token
- Compound V2: `withdraw(uint256)` — no token param, token is the cToken itself
- ERC-4626 vault: `withdraw(uint256 assets, address receiver, address owner)` — first address is the receiver, not the token

`FirstAddressBeforeFirstUint` applies blindly by name, so any non-Aave `withdraw` gets the wrong address surfaced as "the token" in the hero. Same for `supply`/`repay`/`transfer` on random contracts with matching names.

Can we key the registry on `(4-byte selector, full param-type signature)` instead? That turns it into an explicit allow-list of ABIs we've actually vetted

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
```diff
@@ -0,0 +1,168 @@
...
+            current.append(ch)
+        }
+    }
+    if (current.isNotEmpty()) {
+        parts.add(current.toString().trim())
+    }
+    return parts
+}
+
+object ContractCallExtractor {
```
rkokhatskyi: No tests for this file. This is what renders "the transaction you're signing" — can we get coverage before it lands? At a minimum:

- Aave `supply` happy path
- Compound V3 `supplyTo` NthAddress(1)
- ERC20 `transfer` uses toAddress
- `approve(MAX_UINT256)` → "Unlimited"
- ERC-4626-shape `withdraw` → returns null (mismatch), not a fake token
- Malformed argsJson → null
- Tuple-nested params

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1283,14 +1300,63 @@ constructor(
...
+                    signature = functionSignature,
+                    toAddress = toAddress,
+                    chain = chain,
+                )
+            }
+
+        return FunctionInfo(
+            signature = functionSignature,
+            inputs = functionInputs,
+            tokenDisplay = resolvedToken?.let { "${it.value} ${it.token.ticker}" },
```
rkokhatskyi: `tokenDisplay = resolvedToken?.let { ... }` ties the two together — if `resolvedToken` is null, `tokenDisplay` is always null too.

The new branches added in 559c4d8 at `SendTxOverviewScreen.kt:152` and `VerifySendScreen.kt:283` check `tokenDisplay != null && resolvedToken == null`. With the current coupling those can't trigger.

Pick one:

a) If the intent is "sentinel cases produce a display string but no ValuedToken", compute the sentinel first and set the fields independently:

```kotlin
val sentinel = funcName?.let(::sentinelLabelFor)
val resolvedToken = if (sentinel == null) resolveContractCall(...) else null
tokenDisplay = sentinel?.let { "$it ${coin.ticker}" }
    ?: resolvedToken?.let { "${it.value} ${it.token.ticker}" }
```

b) If `tokenDisplay` is always just a rendering of `resolvedToken`, the downstream branches are dead and should be removed

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
```diff
@@ -130,8 +147,25 @@ internal fun SendTxOverviewScreen(
                     )
                 }
 
+                // Decoded contract-call amount (e.g. "Unlimited USDC" for an approve MAX):
+                // shown only when the hero itself did not already render a resolved token.
+                if (tx.tokenDisplay != null && tx.resolvedToken == null) {
```
rkokhatskyi: This branch never triggers — `tokenDisplay` is derived from `resolvedToken` in the VM (see thread on `JoinKeysignViewModel.kt:1329`), so `tokenDisplay != null && resolvedToken == null` is always false

app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
```diff
@@ -250,21 +278,21 @@ internal fun VerifySendScreen(
                             )
                         }
 
-                    if (state.functionSignature != null) {
+                    // Only show the "Amount" row if the hero hasn't already rendered the
+                    // resolved token — otherwise it's duplicated information.
+                    if (tx.tokenDisplay != null && tx.resolvedToken == null) {
```
rkokhatskyi: Same as `SendTxOverviewScreen.kt:152` — unreachable because of the coupling on `JoinKeysignViewModel.kt:1329`

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
```diff
@@ -0,0 +1,168 @@
...
+// "all available", whose exact value depends on on-chain state, so we also return null.
+private val UNLIMITED_APPROVAL_FUNCTIONS = setOf("approve")
+
+/**
+ * If [funcName] uses MAX_UINT256 as an "unlimited approval" sentinel, returns `"Unlimited"`. For
+ * all other functions (withdraw, repay, increaseAllowance, decreaseAllowance, etc.) returns `null`
+ * — the caller should omit the amount rather than display a misleading label.
+ */
+fun sentinelLabelFor(funcName: String): String? =
+    if (funcName in UNLIMITED_APPROVAL_FUNCTIONS) "Unlimited" else null
```
rkokhatskyi: `"Unlimited"` is hardcoded in English but ends up as hero text (via `ValuedToken.value` → `VsOverviewToken`). Needs to be a `stringResource` with translations in the other 9 locale files

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1283,14 +1300,63 @@ constructor(
...
+            functionName = funcName?.replaceFirstChar { it.titlecase(Locale.ROOT) },
```
rkokhatskyi: `replaceFirstChar { titlecase }` only uppercases the first letter:

- `supplyWithPermit` → "SupplyWithPermit"
- `depositIntoStrategy` → "DepositIntoStrategy"
- `repayWithATokens` → "RepayWithATokens"

This is what users see as the hero title. Either split on camelCase boundaries and title-case each word, or use a small curated map keyed by function name — the latter also lets you use the per-protocol term (e.g. Aave calls it "Supply", Compound calls it "Mint")

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/UiModelOverviewExtensions.kt
```diff
@@ -16,6 +16,11 @@ internal fun TransactionTypeUiModel.toUiTransactionInfo(): UiTransactionInfo {
                 memo = this.tx.memo ?: "",
                 networkFeeFiatValue = this.tx.networkFeeFiatValue,
                 networkFeeTokenValue = this.tx.networkFeeTokenValue,
+                functionName = this.tx.functionName,
+                resolvedToken = this.tx.resolvedToken,
```
rkokhatskyi: The same 5 contract-call fields (`functionSignature`, `functionInputs`, `functionName`, `tokenDisplay`, `resolvedToken`) live in three models: `FunctionInfo`, `TransactionDetailsUiModel`, `UiTransactionInfo`. 559c4d8 already had to fix a missed `tokenDisplay` copy here.

Extract a `ContractCallDisplay` value type and embed it as one nullable field on each carrier. Removes the 3-way drift

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1283,14 +1300,63 @@ constructor(
...
+            resolvedToken = resolvedToken,
+        )
+    }
+
+    private suspend fun resolveContractCall(
```
rkokhatskyi: `resolveContractCall` is doing 5 things — 4byte lookup, ABI extraction, vault coin lookup, built-in registry lookup, BigInteger parsing, sentinel handling, amount formatting. There's an existing `…UseCase` pattern in the codebase (`SearchEvmTokenUseCase`, `SearchTokenUseCase`). Would fit better as `ResolveContractCallUseCase` — also makes it testable in isolation (ties into the tests question)

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -926,7 +937,12 @@ constructor(
                                 selectedToken = payload.coin,
                             )
                         )
-                    val functionInfo = getTransactionFunctionInfo(payload.memo, chain)
+                    // 4byte lookup + ABI decode + BigInteger math. Run on IO to match the
+                    // surrounding pattern and avoid blocking the caller dispatcher.
+                    val functionInfo =
+                        withContext(Dispatchers.IO) {
```
rkokhatskyi: Nit: this `withContext(Dispatchers.IO)` belongs inside `getTransactionFunctionInfo`, not at the caller. Suspend functions own their dispatcher boundary. The comment here literally says "to match the surrounding pattern" — that pattern exists at 9 other call sites in this file and is cargo-culted

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
```diff
@@ -34,6 +34,8 @@ interface TokenRepository {
 
     suspend fun getEVMTokenByContract(chainId: String, contractAddress: String): Coin?
 
+    suspend fun getBuiltInTokenByContract(chain: Chain, contractAddress: String): Coin?
```
rkokhatskyi: Nit: `getEVMTokenByContract(chainId: String)` vs new `getBuiltInTokenByContract(chain: Chain)` — same domain, different types. Names also don't signal that one hits RPC and the other reads the bundled registry. Consider `findTokenOnChain(...)` / `findTokenInRegistry(...)` and unify on `Chain`

data/src/main/kotlin/com/vultisig/wallet/data/utils/AbiToJson.kt
```diff
@@ -72,10 +71,17 @@ private fun convertSimpleValue(type: String, value: JsonElement): JsonElement {
...
+ * (`1.157920892373162E77`) — that both looks wrong in the raw function-inputs box AND breaks
+ * downstream parsers that assume plain decimal digits (see ContractCallExtractor.extract).
+ */
+private fun parseBigIntAsJsonString(value: String): JsonElement {
     return try {
         val big = BigInteger(value)
-        JsonPrimitive(BigDecimal(big))
+        JsonPrimitive(big.toString())
     } catch (_: Exception) {
         JsonPrimitive(value)
```
rkokhatskyi: Adjacent to CodeRabbit's narrower-catch nit: even with `NumberFormatException`, the fallback still forwards the raw input string into `functionInputs`, which gets rendered in the "Transaction Details" section of the verify screen. Prefer `JsonNull` so malformed ABI input can't end up as arbitrary text in the signing UI (lookalike unicode / RTL overrides are the concern, not XSS)

```kotlin
} catch (_: NumberFormatException) {
    JsonNull
}
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
rkokhatskyi: Companion PRs exist on iOS and Windows. Without a shared source for which functions are in the registry + which extraction strategy applies, the three platforms will drift — same tx will decode differently depending on what the user opened it on.

Consider moving the registry into `commondata/` as JSON so all three read from the same file

aminsato (commented):
.

app/src/main/java/com/vultisig/wallet/ui/components/buttons/VsHoldableButton.kt
```diff
@@ -56,49 +53,43 @@ fun VsHoldableButton(
...
-                            }
-                        }
-                    }
-                }
+                .combinedClickable(
+                    interactionSource = interactionSource,
+                    indication = null,
+                    enabled = isButtonEnabled,
+                    onClick = onClick,
+                    onLongClick = onLongClick,
```
aminsato, line 86: `combinedClickable.onLongClick` fires at the platform's `ViewConfiguration.longPressTimeoutMillis` (~400ms), but the progress fill animation runs for `holdDuration` (default 800ms). This means `onLongClick` fires when the bar is only ~50% full — the visual "hold to confirm" feedback is misleading.

Options:
- Derive the animation duration from `ViewConfiguration.longPressTimeoutMillis` so the fill completes when the callback fires
- Or drop `combinedClickable` and use custom pointer input with the animation driving the timing

app/src/main/java/com/vultisig/wallet/ui/components/buttons/VsHoldableButton.kt
```diff
@@ -56,49 +53,43 @@ fun VsHoldableButton(
...
-    val scope = rememberCoroutineScope()
     val progress = remember { Animatable(0f) }
-    var isLongPressed by remember { mutableStateOf(false) }
+    val interactionSource = remember { MutableInteractionSource() }
+
+    // Animate progress bar on press/release. Launching in the LaunchedEffect's own
+    // scope (instead of a separate rememberCoroutineScope) guarantees the animation
+    // job is cancelled when this composable leaves composition, and lets us cancel
+    // the previous job on each new Press without racing a second scope.
+    LaunchedEffect(interactionSource) {
```
aminsato: `LaunchedEffect(interactionSource)` doesn't re-launch when `holdDuration` changes. If `holdDuration` is ever recomposed with a new value, the animation still uses the captured old value. Add `holdDuration` to the key:

```kotlin
LaunchedEffect(interactionSource, holdDuration) {
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
```diff
@@ -0,0 +1,168 @@
...
+            "increaseAllowance" to ExtractionStrategy.ContractIsToken,
+            "decreaseAllowance" to ExtractionStrategy.ContractIsToken,
+        )
+
+    private val json = Json { ignoreUnknownKeys = true }
+
+    fun extract(signature: String, argsJson: String, toAddress: String?): TokenAndAmount? {
+        val parenStart = signature.indexOf('(')
+        val parenEnd = signature.lastIndexOf(')')
+        if (parenStart == -1 || parenEnd == -1) return null
```
aminsato: A signature like `")(foo"` passes the current guard but will crash on `substring`. Add `parenEnd <= parenStart` to the check:

```kotlin
if (parenStart == -1 || parenEnd == -1 || parenEnd <= parenStart) return null
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
```diff
@@ -0,0 +1,168 @@
...
+
+        val args =
+            runCatching {
+                    (json.parseToJsonElement(argsJson) as? JsonArray)?.map {
+                        it.jsonPrimitive.content
+                    }
+                }
+                .getOrNull() ?: return null
+
+        val uint256Idx = paramTypes.indexOf("uint256")
```
aminsato: The param matching compares exact strings (`"uint256"`, `"address"`). If a signature includes parameter names (e.g. `"uint256 amount"`), extraction silently fails. Consider normalizing to base type:

```kotlin
private fun String.baseAbiType(): String = trim().substringBefore(' ')
```

Then use `paramTypes.indexOfFirst { it.baseAbiType() == "uint256" }` instead of `paramTypes.indexOf("uint256")`, and the same for `"address"` comparisons.

data/src/main/kotlin/com/vultisig/wallet/data/utils/AbiToJson.kt
aminsato: Two concerns here:
1. Catch should be `NumberFormatException`, not `Exception` — broad catch masks unrelated errors
2. The fallback forwards the raw input string into `functionInputs`, which gets rendered in the signing UI. Malformed ABI input could end up as arbitrary text (lookalike unicode / RTL overrides). Use `JsonNull` instead:

```kotlin
} catch (_: NumberFormatException) {
    JsonNull
}
```

aminsato:
Could you please verify the swap transaction behavior with https://opensea.io/ and confirm it works as expected? Just want to make sure there are no unexpected issues with the swap flow on that platform.

data/src/main/kotlin/com/vultisig/wallet/data/utils/AbiToJson.kt
```diff
@@ -72,10 +71,17 @@ private fun convertSimpleValue(type: String, value: JsonElement): JsonElement {
...
+ * for values like MAX_UINT256 (2^256 - 1) gets rendered in scientific notation
+ * (`1.157920892373162E77`) — that both looks wrong in the raw function-inputs box AND breaks
+ * downstream parsers that assume plain decimal digits (see ContractCallExtractor.extract).
+ */
+private fun parseBigIntAsJsonString(value: String): JsonElement {
     return try {
         val big = BigInteger(value)
-        JsonPrimitive(BigDecimal(big))
+        JsonPrimitive(big.toString())
     } catch (_: Exception) {
```
rkokhatskyi: Still catching `Exception` here. `BigInteger(value)` only throws `NumberFormatException`. The broad catch masks anything unexpected and silently forwards raw attacker-controlled text as `JsonPrimitive(value)` into the signing UI. Change to `catch (_: NumberFormatException)`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
```diff
@@ -0,0 +1,171 @@
...
+// "all available", whose exact value depends on on-chain state, so we also return null.
+private val UNLIMITED_APPROVAL_FUNCTIONS = setOf("approve")
+
+/**
+ * If [funcName] uses MAX_UINT256 as an "unlimited approval" sentinel, returns `"Unlimited"`. For
+ * all other functions (withdraw, repay, increaseAllowance, decreaseAllowance, etc.) returns `null`
+ * — the caller should omit the amount rather than display a misleading label.
+ */
+fun sentinelLabelFor(funcName: String): String? =
+    if (funcName in UNLIMITED_APPROVAL_FUNCTIONS) "Unlimited" else null
```
rkokhatskyi: `"Unlimited"` is a hardcoded English string in the data layer that surfaces on the signing screen. Users in non-English locales see English in a security-critical place. Move to `strings.xml` and all 9 locale files.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
```diff
@@ -0,0 +1,171 @@
...
+            "transfer" to ExtractionStrategy.ContractIsToken,
+            "transferFrom" to ExtractionStrategy.ContractIsToken,
+            "approve" to ExtractionStrategy.ContractIsToken,
+            "increaseAllowance" to ExtractionStrategy.ContractIsToken,
+            "decreaseAllowance" to ExtractionStrategy.ContractIsToken,
+        )
+
+    private val json = Json { ignoreUnknownKeys = true }
+
+    fun extract(signature: String, argsJson: String, toAddress: String?): TokenAndAmount? {
```
rkokhatskyi: No tests for `extract`. This function decides what token and amount a co-signer sees on the signing screen. A misparse directly misleads the signer. At minimum cover: Aave `supply` happy path, `approve(MAX_UINT256)` producing "Unlimited", malformed argsJson returning null.

rkokhatskyi (dismissed):

aminsato:
@coderabbitai full review

rkokhatskyi:
Gonna pick up the PR to finish the changes

app/src/main/java/com/vultisig/wallet/ui/components/buttons/VsHoldableButton.kt
```diff
@@ -58,47 +56,41 @@ fun VsHoldableButton(
...
+                        }
+                }
+                is PressInteraction.Release,
+                is PressInteraction.Cancel -> {
+                    animJob?.cancel()
+                    scope.launch { progress.snapTo(0f) }
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
rkokhatskyi: addressed in f90ed371b — `holdDuration` parameter removed entirely so there is no caller-overridable duration to desync from `combinedClickable.onLongClick`. animation is now driven by `LocalViewConfiguration.current.longPressTimeoutMillis` directly, which is also what the platform long-press detector uses, so the fill completes exactly when `onLongClick` fires.

app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
```diff
@@ -182,15 +191,21 @@ internal fun VerifySendScreen(
...
+                    UiSpacer(12.dp)
 
-                    SwapToken(valuedToken = tx.token)
+                    VsOverviewToken(
+                        header = heroHeader,
+                        valuedToken = heroToken,
+                        shape = RoundedCornerShape(0.dp),
+                        modifier = Modifier.fillMaxWidth(),
+                        withContainer = false,
+                    )
```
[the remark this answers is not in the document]
rkokhatskyi: moot in f90ed371b — both `tokenDisplay`/`resolvedToken` and the trio-guard fallback are gone. screen now collapses to `functionName != null ? title : native-amount`, and #4306 layers the Blockaid hero on top of that.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
```diff
@@ -293,6 +301,10 @@ internal data class UiTransactionInfo(
     val networkFeeTokenValue: String,
     val networkFeeFiatValue: String,
     val signMethod: String = "",
+    val functionName: String? = null,
+    val resolvedToken: ValuedToken? = null,
+    val functionSignature: String? = null,
+    val functionInputs: String? = null,
```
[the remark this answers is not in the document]
rkokhatskyi: moot in f90ed371b — `tokenDisplay` removed from `UiTransactionInfo` (and `TransactionDetailsUiModel` and `FunctionInfo`). Blockaid drives amounts in #4306; nothing left to forward through the transport object.

app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt
```diff
@@ -236,15 +253,39 @@ private fun TransactionDetail(transaction: TransactionDetailsUiModel?) {
...
-            title = stringResource(R.string.verify_transaction_amount_title),
-            value = transaction.token.value,
-        )
+        if (transaction.tokenDisplay != null && transaction.resolvedToken == null) {
+            UiHorizontalDivider()
+            OtherField(
+                title = stringResource(R.string.verify_transaction_amount_title),
+                value = transaction.tokenDisplay,
+            )
+        }
```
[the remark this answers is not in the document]
rkokhatskyi: moot in f90ed371b — the duplicate `tokenDisplay` row plus the trio-guard suppression are both gone. details section now just renders `functionSignature` / `functionInputs`, and the native amount + fiat rows when there is no decoded function.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
```diff
@@ -0,0 +1,168 @@
...
+private val UNLIMITED_APPROVAL_FUNCTIONS =
+    setOf("approve", "increaseAllowance", "decreaseAllowance")
+
+/**
+ * If [funcName] uses MAX_UINT256 as an "unlimited approval" sentinel, returns `"Unlimited"`. For
+ * all other functions (withdraw, repay, etc.) returns `null` — the caller should omit the amount
+ * rather than display a vague label.
+ */
+fun sentinelLabelFor(funcName: String): String? =
+    if (funcName in UNLIMITED_APPROVAL_FUNCTIONS) "Unlimited" else null
```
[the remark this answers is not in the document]
rkokhatskyi: moot in f90ed371b — `ContractCallExtractor` and `sentinelLabelFor` deleted. The decode-derived amount path was front-runnable via 4byte and the registry was hand-curated per protocol; #4306 replaces it with Blockaid simulation as the authoritative hero.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
```diff
@@ -0,0 +1,168 @@
...
+
+    fun extract(signature: String, argsJson: String, toAddress: String?): TokenAndAmount? {
+        val parenStart = signature.indexOf('(')
+        val parenEnd = signature.lastIndexOf(')')
+        if (parenStart == -1 || parenEnd == -1) return null
+
+        val funcName = signature.substring(0, parenStart).trim()
+        val strategy = registry[funcName] ?: return null
+
+        val paramTypes = splitTopLevel(signature.substring(parenStart + 1, parenEnd))
```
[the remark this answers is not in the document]
rkokhatskyi: moot — file deleted in f90ed371b.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
rkokhatskyi: addressed in f90ed371b — dropped the decode-derived hero amount path entirely (the front-running attack surface: poisoned 4byte selector + matching registry strategy → fabricated `Supply 100 USDC` headline). 4byte still produces the function-name *title* (small-text caption above the hero), and #4306 layers Blockaid as the authoritative hero amount; when Blockaid simulation is unavailable, `HeroContent.Unverified` renders explicitly instead of trusting the 4byte name as the action.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
rkokhatskyi: moot in f90ed371b — registry deleted with `ContractCallExtractor`. The wrong-protocol decode (Aave V3 vs Compound V2 vs ERC-4626 vault `withdraw`) goes away with it.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
rkokhatskyi: moot — file deleted in f90ed371b.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
rkokhatskyi: moot in f90ed371b — both fields deleted, the unreachable branches in `SendTxOverviewScreen.kt:152` and `VerifySendScreen.kt:283` are gone too.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
rkokhatskyi: moot — `tokenDisplay` block deleted in f90ed371b.

app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
rkokhatskyi: moot — `tokenDisplay` block deleted in f90ed371b.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
rkokhatskyi: moot in f90ed371b — `sentinelLabelFor` and the entire data-layer hardcoded `"Unlimited"` deleted with the resolver. #4306 owns the localized unverified-call hero (`dapp_hero_unverified_function_title` / `_subtitle` are already in all 10 locale files).

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
rkokhatskyi: fixed in f90ed371b — `prettifyEvmFunctionName` splits on camelCase boundaries with `Regex("(?<=[a-z])(?=[A-Z])")` and title-cases each word, so `supplyWithPermit` → `Supply With Permit`, `depositIntoStrategy` → `Deposit Into Strategy`, `repayWithATokens` → `Repay With A Tokens`. Curated map per protocol could come later if we want Aave-style "Supply" vs Compound-style "Mint" branding.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/UiModelOverviewExtensions.kt
rkokhatskyi: moot in f90ed371b — 2 of the 5 fields (`tokenDisplay`, `resolvedToken`) deleted; only `functionSignature` / `functionInputs` / `functionName` survive across the carriers. Still flat fields rather than a value type, but the drift surface is small enough to not matter; happy to extract `ContractCallDisplay` later if it grows back.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
rkokhatskyi: moot in f90ed371b — `resolveContractCall` deleted entirely. `getTransactionFunctionInfo` is now just 4byte name + args lookup, no extraction / vault registry / built-in registry / sentinel / format chain.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
rkokhatskyi: fixed in f90ed371b — `withContext(Dispatchers.IO)` moved inside `getTransactionFunctionInfo` so the suspend fn owns its own dispatcher boundary; caller is back to a plain suspend call.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
rkokhatskyi: moot — `getBuiltInTokenByContract` removed entirely in f90ed371b; the resolver was the only caller, so no naming inconsistency to resolve.

data/src/main/kotlin/com/vultisig/wallet/data/utils/AbiToJson.kt
rkokhatskyi: fixed in f90ed371b — `parseBigIntAsJsonString` now catches `NumberFormatException` specifically and falls through to `JsonNull` instead of forwarding the raw input string. Lookalike-Unicode / RTL-override / zero-width codepoints in attacker-controlled ABI args can no longer reach the rendered function-inputs box.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
rkokhatskyi: moot in f90ed371b — registry deleted, no per-platform registry to drift. #4306 mirrors the iOS / Windows simulation parsers from a single Blockaid response shape, which is the right place for shared decoding to live.

app/src/main/java/com/vultisig/wallet/ui/components/buttons/VsHoldableButton.kt
rkokhatskyi: addressed in f90ed371b — dropped the `holdDuration` parameter entirely (it was unused; all 3 call sites pass defaults). Animation duration is hardcoded to `LocalViewConfiguration.current.longPressTimeoutMillis` so animation and `onLongClick` fire on the same clock, and no caller can desync them by passing a custom value. Typed variant if a longer hold is ever needed.

app/src/main/java/com/vultisig/wallet/ui/components/buttons/VsHoldableButton.kt
rkokhatskyi: moot in f90ed371b — `holdDuration` parameter removed, only key on `LaunchedEffect` is `interactionSource`. The local `holdDurationMillis` is read at composition time from `LocalViewConfiguration` so the captured-old-value scenario cannot happen.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
rkokhatskyi: moot — file deleted in f90ed371b.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
rkokhatskyi: moot — file deleted in f90ed371b.

data/src/main/kotlin/com/vultisig/wallet/data/utils/AbiToJson.kt
rkokhatskyi: addressed in f90ed371b — narrowed to `NumberFormatException` and switched the fallback from `JsonPrimitive(value)` to `JsonNull` so attacker-controlled raw text cannot reach the signing UI.

data/src/main/kotlin/com/vultisig/wallet/data/utils/AbiToJson.kt
rkokhatskyi: fixed in f90ed371b.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
rkokhatskyi: moot — `"Unlimited"` removed entirely in f90ed371b when the sentinel logic was deleted; #4306 owns the localized unverified-call hero.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
rkokhatskyi: moot — `extract` deleted in f90ed371b.

app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
```diff
@@ -361,6 +380,66 @@ internal fun AddressField(title: String, address: String, divider: Boolean = tru
...
+
+            IconButton(onClick = { isExpanded = !isExpanded }, modifier = Modifier.size(16.dp)) {
+                UiIcon(
+                    drawableResId = R.drawable.chevron,
+                    tint = Theme.v2.colors.text.tertiary,
+                    size = 8.dp,
+                    modifier = Modifier.graphicsLayer(rotationZ = if (isExpanded) 180f else 0f),
+                )
+            }
+        }
```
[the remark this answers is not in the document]
rkokhatskyi: fixed in f90ed371b — switched the `TransactionDetailsSection` chevron row to `Arrangement.SpaceBetween`. Locale-aware now, will mirror correctly when RTL locales are added.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
```diff
@@ -73,18 +73,28 @@ internal fun SendTxOverviewScreen(
...
+                        if (tx.type == UiTransactionInfoType.Send) {
+                            stringResource(R.string.tx_overview_screen_tx_send)
+                        } else {
+                            stringResource(R.string.tx_overview_screen_tx_deposit)
+                        },
+                    valuedToken = tx.token,
+                    shape = RoundedCornerShape(24.dp),
+                    modifier = Modifier.fillMaxWidth(),
+                )
+            }
```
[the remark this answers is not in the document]
rkokhatskyi: no actionable suggestion in this analysis chain — leaving as-is.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractCallExtractor.kt
```diff
@@ -0,0 +1,171 @@
...
+        val paramTypes =
+            splitTopLevel(signature.substring(parenStart + 1, parenEnd)).map(String::baseAbiType)
+
+        val args =
+            runCatching {
+                    (json.parseToJsonElement(argsJson) as? JsonArray)?.map {
+                        it.jsonPrimitive.content
+                    }
+                }
+                .getOrNull() ?: return null
```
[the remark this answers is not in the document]
rkokhatskyi: moot — file deleted in f90ed371b.

rkokhatskyi:
🤖 Pushed `f90ed371b`. Dropped the contract-call resolver (`ContractCallExtractor` + per-protocol DeFi registry + `MAX_UINT256` sentinel + the `tokenDisplay`/`resolvedToken` fields) — #4306 will own the resolved hero through Blockaid simulation, and keeping the registry path alive next to it would leave a front-runnable surface that the new hero is meant to remove.

Bundled cleanups while in the diff:
- `VsHoldableButton`: dropped the unused `holdDuration` parameter, hardcoded animation to the platform long-press timeout
- `AbiToJson.parseBigIntAsJsonString`: catch `NumberFormatException` only, fall back to `JsonNull` instead of forwarding raw input
- `TransactionDetailsSection` chevron row: `Arrangement.SpaceBetween` instead of `Absolute.SpaceBetween` (LTR-locked)

10 files, +63 / −305. Replied inline on each thread.

---
*Generated by Claude*

app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
```diff
@@ -361,6 +372,66 @@ internal fun AddressField(title: String, address: String, divider: Boolean = tru
...
+                }
+
+                functionInputs?.let {
+                    VerifyCardJsonDetails(
+                        title = stringResource(R.string.verify_transaction_function_inputs_title),
+                        subtitle = it,
+                    )
+                }
+            }
+        }
```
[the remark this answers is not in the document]
rkokhatskyi: fixed in fee609421 — dropped the inner `.verticalScroll(rememberScrollState())` and `.heightIn(max = 300.dp)` cap. The outer scaffold scroll now reaches the full content of the expanded signature/inputs box.

app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
```diff
@@ -361,6 +372,66 @@ internal fun AddressField(title: String, address: String, divider: Boolean = tru
...
+            )
+
+            IconButton(onClick = { isExpanded = !isExpanded }, modifier = Modifier.size(16.dp)) {
+                UiIcon(
+                    drawableResId = R.drawable.chevron,
+                    tint = Theme.v2.colors.text.tertiary,
+                    size = 8.dp,
+                    modifier = Modifier.graphicsLayer(rotationZ = if (isExpanded) 180f else 0f),
+                )
+            }
```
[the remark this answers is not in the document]
rkokhatskyi: fixed in fee609421 — replaced the `IconButton` with a 16dp `Box(Modifier.size(16.dp).clickable { ... })`. The chevron now renders at the intended size; no fight with `IconButton`s 40dp minimum touch target.

rkokhatskyi:
🤖 Pushed `fee609421` addressing the rest of the CodeRabbit review:

- `VsHoldableButtonDisabledPreview` now passes `VsButtonState.Disabled` so the preview actually renders the disabled state
- `SendTxOverviewScreen` function-name hero typography aligned to `headings.title3` (matches `VerifySendScreen` and `TransactionDoneScreen`, no size jump between screens)
- `FunctionInfo.inputs` is now nullable: arg decoding is best-effort, so a malformed ABI no longer drops the whole `FunctionInfo` and the function name still surfaces above the hero

Also merged latest `main` (commit `ba79eb460`) so the branch is no longer conflicting.

The duplicate finding about the 4byte trust boundary on `functionName` is the same concern I raised earlier — addressed by demoting 4byte to a small-text caption (vs. the hero amount, which Blockaid will own in #4306) and falling through to `HeroContent.Unverified` when Blockaid simulation is unavailable.

---
*Generated by Claude*

aminsato (commented):
(Moved to inline comments below.)

app/src/main/java/com/vultisig/wallet/ui/components/buttons/VsHoldableButton.kt
```diff
@@ -56,49 +52,39 @@ fun VsHoldableButton(
...
     val progress = remember { Animatable(0f) }
-    var isLongPressed by remember { mutableStateOf(false) }
+    val interactionSource = remember { MutableInteractionSource() }
+    // The animation duration is fixed to the platform long-press timeout so the visual fill
+    // completes exactly when [combinedClickable]'s onLongClick fires. A configurable duration
+    // is a footgun: any value other than the platform timeout desyncs the bar from the gesture
+    // detector. Add a typed variant if a longer hold is ever required.
+    val holdDurationMillis = LocalViewConfiguration.current.longPressTimeoutMillis.toInt()
+
+    LaunchedEffect(interactionSource, holdDurationMillis) {
```
aminsato, line 63: Two small things on this LaunchedEffect:

1. `holdDurationMillis` is read from `LocalViewConfiguration.current` inside composition without `remember` — same value across recompositions, so the effect never re-keys in practice. Could drop it from the key list without changing behavior.
2. The effect still gets installed when `enabled = false`. No interactions fire while disabled so it's benign — leaving as a note in case you want to short-circuit it.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1323,14 +1327,46 @@ constructor(
...
-        return FunctionInfo(functionSignature, functionInputs)
+/**
+ * Extract the function name from an ABI signature like `"withdraw(address,uint256,address)"` and
+ * format it for display: split on camelCase boundaries and title-case each word.
+ *
+ * Returns null if the signature has no `(` or is empty before it.
+ */
+internal fun prettifyEvmFunctionName(signature: String): String? {
+    val rawName =
+        signature.substringBefore('(').trim().takeIf {
```
aminsato: `(?<=[a-z])(?=[A-Z])` only splits on lowercase→uppercase boundaries. Names like `ERC20transferFrom` end up `"ERC20transfer From"` and `WBTCSwap` doesn't split at all. Probably acceptable for the long tail; just calling it out so the assumption is explicit.

Also, the `it.length < signature.length` trick to detect a missing `(` works but is roundabout — `signature.indexOf('(') < 0` reads cleaner.

app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt
```diff
@@ -236,15 +257,34 @@ private fun TransactionDetail(transaction: TransactionDetailsUiModel?) {
                 value = transaction.memo,
             )
 
-        OtherField(
-            title = stringResource(R.string.verify_transaction_amount_title),
-            value = transaction.token.value,
-        )
+        if (transaction.functionSignature != null) {
+            UiHorizontalDivider()
```
aminsato: When `memo` is null but `functionSignature` is set, the first thing rendered here is `UiHorizontalDivider()` followed by the signature row — leaves an orphan top divider. Same shape for `functionInputs` below. Worth gating the divider on whether anything came before.

data/src/main/kotlin/com/vultisig/wallet/data/utils/AbiToJson.kt
```diff
@@ -72,12 +71,21 @@ private fun convertSimpleValue(type: String, value: JsonElement): JsonElement {
...
+ * Serialize a big integer ABI value as a JSON *string* so precision is preserved end-to-end.
+ * Emitting it as a JSON number forces kotlinx.serialization to round-trip through a Number, which
+ * for values like MAX_UINT256 (2^256 - 1) gets rendered in scientific notation
+ * (`1.157920892373162E77`) — wrong in the user-facing function-inputs box.
+ *
+ * Malformed input falls through to [JsonNull] rather than forwarding the raw string. The args JSON
+ * is rendered verbatim into the signing UI; raw forwarding would let lookalike Unicode / RTL
+ * overrides / zero-width codepoints in attacker-controlled ABI args reach the signing screen.
+ */
+private fun parseBigIntAsJsonString(value: String): JsonElement {
```
aminsato, line 84: Hardening here looks good — narrowing to `NumberFormatException` and returning `JsonNull` for malformed values both close the lookalike-unicode / RTL-override path that was raised earlier. Nice.

aminsato:
Two PR-level notes (not tied to specific lines):

- **Still-open threads from earlier rounds** — several threads from CodeRabbit / earlier reviews reference `tokenDisplay` / `resolvedToken` / `ContractCallExtractor.kt`, which don't exist in the current diff. Worth replying "removed in 559c4d8, scoped to Phase 1" (or similar) on those threads so the review state reflects reality.
- **Test plan in the PR description** — still mentions Aave supply with "Unlimited USDC" and Aave withdraw with `MAX_UINT256`, but the sentinel handling and Blockaid resolution were scoped out. The visible behavior in this PR is just the function-name hero + suppressed native amount. Worth updating the test plan so reviewers know what to actually verify.

app/src/main/java/com/vultisig/wallet/ui/components/buttons/VsHoldableButton.kt
rkokhatskyi: agree on both. (1) `holdDurationMillis` is read from `LocalViewConfiguration.current` at composition time and is stable per device, so keying on it is a defensive no-op rather than something that buys re-execution; left in as documentation that the duration is what the effect cares about. (2) `combinedClickable(enabled = isButtonEnabled)` gates `PressInteraction` emission, so the `LaunchedEffect` does not observe anything while disabled. could short-circuit by gating the effect on `isButtonEnabled` too, but that would re-launch on enable transitions for no behavioral win.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
rkokhatskyi: fixed in b92f50daa — adopted both. missing-paren check is now `signature.substringBefore("(", missingDelimiterValue = "")` so an empty result falls through the same `isEmpty()` guard that catches `(foo)`-style inputs; no more length comparison. boundary regex extended to `(?<=[a-z])(?=[A-Z])|(?<=[A-Z])(?=[A-Z][a-z])` so `WBTCSwap` → `"WBTC Swap"`. digit-prefixed names like `ERC20transferFrom` are still left as-is — `ERC 20 Transfer From` reads worse than `ERC20transfer From`, so I am leaving the digit boundary out.

app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt
rkokhatskyi: fixed in b92f50daa — dropped the explicit `UiHorizontalDivider()` calls before both rows. `OtherField` already renders its own bottom divider, so the explicit one above stacked two horizontal lines and orphaned a leading divider when `memo` was absent.

data/src/main/kotlin/com/vultisig/wallet/data/utils/AbiToJson.kt
rkokhatskyi: thanks — that was the path I was most worried about in the original review. the rendered function-inputs box treats whatever string we hand it as text-of-record, so the JsonNull fallback was the structurally right answer rather than a deeper sanitization layer.

johnnyluo (approved):
LGTM

PR #4061, opened 2026-04-14
fix: prevent splash screen from hanging on slow DB

rkokhatskyi:
Devs are reporting the splash screen hanging on cold-start for 20+ seconds, sometimes resolving after force-closing and reopening a few times. Release builds strip Timber so there's no app-side signal, but from ADB logs the process is alive and the main thread is responsive — the only thing holding the splash is `MainViewModel`'s init coroutine waiting on `vaultRepository.hasVaults()`, which transitively opens the Room DB and can stall for any number of reasons (contended storage, keystore daemon, etc.).

Wrap that call in `withTimeout(5.seconds)` and treat timeout/exception as "no vaults" so the splash always dismisses into `Route.AddVault` instead of hanging. Route the init block and the `openUri` deeplink entry through `safeLaunch` so a thrown exception can no longer leave `_isLoading = true` forever.


johnnyluo (approved):
LGTM

PR #4063, opened 2026-04-14
refactor: migrate MasterKeys to MasterKey.Builder

rkokhatskyi:
`MasterKeys.getOrCreate` in security-crypto 1.0.0 is the legacy init path; on Android 14+ it sometimes stalls the keystore2 daemon for tens of seconds on first use, which matches the splash-hang reports. `MasterKeys` is also deprecated, and the stable 1.1.0 replacement exposes `.setRequestStrongBoxBacked(false)` so we can opt out of the StrongBox path that causes most of the stalls on Pixel and Samsung TEE devices.

Bump `security-crypto` 1.0.0 → 1.1.0 and swap both call sites (prewarm in `VoltixApplication.onCreate`, sync init in `MainDataModule.provideEncryptedSharedPrefs`) to `MasterKey.Builder`. Same key alias, same AES-256-GCM scheme, same prefs file — existing encrypted data decrypts verbatim on upgrade.


rkokhatskyi:
@coderabbitai full review

johnnyluo (approved):
LGTM

PR #4064, opened 2026-04-14
fix: time-out cached THORChain network id read

rkokhatskyi:
`ThorChainRepositoryImpl.getCachedNetworkChainId()` reads from `AppDataStore` with no upper bound. The underlying preferences file is shared with 15+ other repositories and `TokenRefreshWorker`, so a writer holding the internal mutex — or a slow first-access on a cold device — blocks the read. Called during app startup from `InitializeThorChainNetworkIdUseCase`, a stall there leaves `THORCHAIN_NETWORK_ID` at its hard-coded default with no live fetch ever triggered.

Wrap the read in `withTimeout(3.seconds)`, log and return `null` on timeout. The caller already treats `null` as a cache miss and falls back to `fetchNetworkChainId()`, so behaviour under timeout is the same as an empty cache.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application responsiveness by adding a timeout mechanism for cache operations, ensuring the app doesn't hang when retrieving cached network data.


rkokhatskyi:
@coderabbitai full review

johnnyluo (approved):
LGTM

PR #4071, opened 2026-04-15
fix: move Upload QR button into ErrorView on permission denied

aminsato:
## Summary
- When camera permission is denied on the Scan QR screen, the "Upload QR Code" bottom bar button was obscured behind the full-screen `ErrorView`
- Now the bottom bar is hidden when permission is denied, and the upload action appears as a secondary CTA button inside the `ErrorView` itself
- Added optional `secondaryButtonUiModel` parameter to `ErrorView` for reuse across other screens
- Added bottom padding to `ErrorWaves` to prevent button overlap with wave circles

## Before / After

| Before | After |
|--------|-------|
| before | after |

## Test plan
- [ ] Deny camera permission and verify the "Upload QR Code" button appears inside the error view below "Open Settings"
- [ ] Tap "Upload QR Code" from the error view and verify image picker opens correctly
- [ ] Tap "Open Settings" and verify it navigates to app settings
- [ ] Grant camera permission and verify the bottom bar "Upload QR Code" button still works normally
- [ ] Verify other screens using `ErrorView` are unaffected (keygen, keysign, peer discovery)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/components/errors/ErrorView.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/errors/ErrorView.kt
@@ -65,7 +65,7 @@ internal fun ErrorView(
             }
 
             secondaryButtonUiModel?.let {
-                UiSpacer(12.dp)
+                UiSpacer(if (buttonUiModel != null) 12.dp else 30.dp)
                 VsButton(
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4073, opened 2026-04-15
fix: time out mediator discovery when no host device is found

rkokhatskyi:
## Description

The join-keygen flow kicked off NSD service discovery inside a `suspendCoroutine` with no timeout and no cancellation path. If the host device never advertised (wrong Wi-Fi, firewall, or never launched) the user was stuck on a silent loading screen until they navigated away.

Switched to `suspendCancellableCoroutine` wrapped in `withTimeoutOrNull`. Discovery now stops cleanly on cancellation or a 30s timeout, and surfaces a translated error telling the user the other device could not be found on their network.

## Which feature is affected?
- [x] Create vault ( Secure / Fast) - Please ensure you created a Secure vault & fast vault

## Checklist


johnnyluo (approved):
LGTM

PR #4080, opened 2026-04-15
fix: fall back safely when EIP-1559 fee history is empty

rkokhatskyi:
## Description

`calculateMaxPriorityFeePerGas` reads the median of the rewards returned by `eth_feeHistory` as `rewardsFeeHistory[size / 2]`. On the Polygon and default (Ethereum and other) branches that crashes with `IndexOutOfBoundsException` whenever the RPC returns an empty history. The neighbouring Blast / Base / Optimism branches already guard against this with `maxOrNull() ?: fallback`; Polygon and default did not.

Introduced a private `List<BigInteger>.median()` extension that returns null for an empty list, paired with the existing chain-specific floor (`maxOf(..., minFee)`) so a missing sample degrades to that floor instead of crashing.

Added unit tests covering every branch of `calculateMaxPriorityFeePerGas` — Avalanche, Arbitrum, Mantle, Blast, Base, Optimism, Polygon, default — plus the legacy-gas path (BSC), EIP-1559 assembly math, median edge cases (empty, single-element, two-element), limit selection, and error paths.

## Which feature is affected?

## Checklist

## Release Notes

* **Bug Fixes**
  * Enhanced robustness of EIP-1559 fee calculation to properly handle cases with unavailable fee history, with chain-specific fee floor protections applied across supported networks.

* **Tests**
  * Added comprehensive test coverage for fee calculation logic across all supported blockchain networks, including edge cases and network-specific fee behaviors.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
@@ -17,6 +17,7 @@ import java.math.BigInteger
 import javax.inject.Inject
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
+import timber.log.Timber
 
 class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFactory) :
     FeeService {
@@ -172,14 +173,23 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
                     rewardsFeeHistory.maxOrNull() ?: DEFAULT_MAX_PRIORITY_FEE_PER_GAS_L2
 
                 // polygon has min of 30 gwei, but some blocks comes with less rewards
-                Chain.Polygon ->
+                Chain.Polygon -> {
+                    if (rewardsFeeHistory.isEmpty()) {
+                        Timber.w("Fee history is empty for %s, using fallback", chain)
+                    }
                     maxOf(
                         rewardsFeeHistory.median() ?: BigInteger.ZERO,
                         GWEI * DEFAULT_MAX_PRIORITY_FEE_POLYGON,
                     )
+                }
 
                 // picked medium with min of 1 GWEI (ETH etc..)
-                else -> maxOf(rewardsFeeHistory.median() ?: BigInteger.ZERO, GWEI)
+                else -> {
+                    if (rewardsFeeHistory.isEmpty()) {
+                        Timber.w("Fee history is empty for %s, using fallback", chain)
+                    }
+                    maxOf(rewardsFeeHistory.median() ?: BigInteger.ZERO, GWEI)
+                }
             }
         }
     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
@@ -174,12 +174,12 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
                 // polygon has min of 30 gwei, but some blocks comes with less rewards
                 Chain.Polygon ->
                     maxOf(
-                        rewardsFeeHistory[rewardsFeeHistory.size / 2],
+                        rewardsFeeHistory.median() ?: BigInteger.ZERO,
                         GWEI * DEFAULT_MAX_PRIORITY_FEE_POLYGON,
                     )
 
                 // picked medium with min of 1 GWEI (ETH etc..)
-                else -> maxOf(a = rewardsFeeHistory[rewardsFeeHistory.size / 2], b = GWEI)
+                else -> maxOf(rewardsFeeHistory.median() ?: BigInteger.ZERO, GWEI)
             }
         }
     }
@@ -221,3 +221,8 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
         val DEFAULT_MANTLE_SWAP_LIMIT = "3000000000".toBigInteger()
     }
 }
+
+// Returns the mid-index element, or null for an empty list. Callers pair it with
+// a chain-specific floor via `maxOf(..., minFee)` so a missing sample degrades to
+// that floor instead of throwing.
+private fun List<BigInteger>.median(): BigInteger? = getOrNull(size / 2)
--- a/data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
@@ -1,0 +1,382 @@
+@file:OptIn(ExperimentalCoroutinesApi::class)
+
+package com.vultisig.wallet.data.blockchain.ethereum
+
+import com.vultisig.wallet.data.api.EvmApi
+import com.vultisig.wallet.data.api.EvmApiFactory
+import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_ARBITRUM_TRANSFER
+import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_COIN_TRANSFER_LIMIT
+import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_SWAP_LIMIT
+import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_TOKEN_TRANSFER_LIMIT
+import com.vultisig.wallet.data.blockchain.model.Eip1559
+import com.vultisig.wallet.data.blockchain.model.GasFees
+import com.vultisig.wallet.data.blockchain.model.Swap
+import com.vultisig.wallet.data.blockchain.model.Transfer
+import com.vultisig.wallet.data.blockchain.model.VaultData
+import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.Coin
+import io.mockk.coEvery
+import io.mockk.coVerify
+import io.mockk.every
+import io.mockk.mockk
+import java.math.BigInteger
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+internal class EthereumFeeServiceTest {
+
+    private val evmApi: EvmApi = mockk()
+    private val evmApiFactory: EvmApiFactory = mockk()
+    private val service = EthereumFeeService(evmApiFactory)
+
+    @BeforeEach
+    fun setUp() {
+        every { evmApiFactory.createEvmApi(any()) } returns evmApi
+        coEvery { evmApi.getBaseFee() } returns BigInteger.ZERO
+        coEvery { evmApi.getFeeHistory() } returns emptyList()
+    }
+
+    // ---------- Bug #6 regression: empty fee history must not crash ----------
+
+    @Test
+    fun `Polygon with empty fee history falls back to the 30 GWEI floor`() = runTest {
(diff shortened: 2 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
```diff
@@ -221,3 +221,8 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
         val DEFAULT_MANTLE_SWAP_LIMIT = "3000000000".toBigInteger()
     }
 }
+
+// Returns the mid-index element, or null for an empty list. Callers pair it with
+// a chain-specific floor via `maxOf(..., minFee)` so a missing sample degrades to
+// that floor instead of throwing.
+private fun List<BigInteger>.median(): BigInteger? = getOrNull(size / 2)
```
aminsato, line 238: Worth noting: this isn't technically a median — a median requires sorting the list first. Right now it picks the element at index `size/2` in whatever order the RPC returned. For example, if the RPC returns `[100, 1, 50]`, a true median would be `50`, but this returns `1`.

Since the original code had the same behavior, I think it's fine to keep the logic as-is for now, but it might be worth either sorting the list first, or renaming this to something like `middleElement()` so the name matches what it actually does. Just a thought to avoid confusion for future readers!

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
```diff
@@ -174,12 +174,12 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
                 // polygon has min of 30 gwei, but some blocks comes with less rewards
                 Chain.Polygon ->
                     maxOf(
-                        rewardsFeeHistory[rewardsFeeHistory.size / 2],
+                        rewardsFeeHistory.median() ?: BigInteger.ZERO,
```
aminsato, line 181: The fallback to `BigInteger.ZERO` paired with `maxOf(ZERO, 30 GWEI)` is a solid approach. One small suggestion: it might be helpful to add a `Timber.w()` log when the fee history is empty, so that if fees are consistently hitting the 30 GWEI floor in production, there's a breadcrumb to help with debugging. Totally optional, but could save someone time down the road.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
```diff
@@ -174,12 +174,12 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
...
                         GWEI * DEFAULT_MAX_PRIORITY_FEE_POLYGON,
                     )
 
                 // picked medium with min of 1 GWEI (ETH etc..)
-                else -> maxOf(a = rewardsFeeHistory[rewardsFeeHistory.size / 2], b = GWEI)
+                else -> maxOf(rewardsFeeHistory.median() ?: BigInteger.ZERO, GWEI)
```
aminsato: Same thought as the Polygon branch — since this is the catch-all for Ethereum and any future chains, a warning log when fee history comes back empty could be really valuable. On ETH mainnet especially, falling back to the 1 GWEI minimum might lead to stuck transactions if the RPC is having issues, and a log line would make it much easier to diagnose.

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
```diff
@@ -0,0 +1,382 @@
...
+    @Test
+    fun `Optimism uses max reward when history is populated`() = runTest {
+        stubFeeHistory(listOf(gwei(7), gwei(4)))
+
+        val fee = service.calculateDefaultFees(transfer(Chain.Optimism)) as Eip1559
+
+        assertEquals(gwei(7), fee.maxPriorityFeePerGas)
+    }
+
+    @Test
```
aminsato, line 155: Nice edge case to cover! One thing to consider: since this test goes through the full `calculateDefaultFees` pipeline, if Ethereum's floor ever changed to 5 GWEI, this test would still pass but for the wrong reason (floor instead of median). If you'd like to make the median edge cases more precise, testing the `median()` extension directly could help isolate the behavior. Not a blocker though — just a suggestion for extra confidence.

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
```diff
@@ -0,0 +1,382 @@
...
+    fun `Polygon uses the median reward when it exceeds the 30 GWEI floor`() = runTest {
+        // odd-length list: middle index = size/2 = 2 → 100 GWEI
+        stubFeeHistory(listOf(gwei(40), gwei(60), gwei(100), gwei(120), gwei(140)))
+
+        val fee = service.calculateDefaultFees(transfer(Chain.Polygon)) as Eip1559
+
+        assertEquals(gwei(100), fee.maxPriorityFeePerGas)
+    }
```
aminsato, line 163: Small naming nit: for a two-element list `[2, 8]`, a true median would be `(2+8)/2 = 5`, while the function returns `8` (element at index 1). The test name says "upper midpoint" which could be a bit confusing. Maybe something like "picks the element at index size/2" would more accurately describe the behavior? This ties into the naming discussion on the `median()` function itself.

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
```diff
@@ -0,0 +1,382 @@
...
+    @Test
+    fun `Base uses max reward when history is populated`() = runTest {
+        stubFeeHistory(listOf(gwei(50), gwei(200), gwei(100)))
+
+        val fee = service.calculateDefaultFees(transfer(Chain.Base)) as Eip1559
+
+        assertEquals(gwei(200), fee.maxPriorityFeePerGas)
+    }
+
+    @Test
```
aminsato, line 139: This is a great test, but I noticed the input list happens to already be sorted (`[40, 60, 100, 120, 140]`). It might be worth adding a companion test with a shuffled list like `[140, 40, 120, 60, 100]` — if the `median()` function is meant to find the true median, the shuffled version would catch any sorting issues. If the intent is just to pick the middle element regardless of order, then the current test is fine, but a comment clarifying that expectation could help.

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
```diff
@@ -0,0 +1,382 @@
...
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+internal class EthereumFeeServiceTest {
+
+    private val evmApi: EvmApi = mockk()
+    private val evmApiFactory: EvmApiFactory = mockk()
+    private val service = EthereumFeeService(evmApiFactory)
+
```
aminsato, line 36: Minor safety suggestion: `service` is initialized as a class property, but the mock setup for `evmApiFactory` happens in `@BeforeEach`. This works fine right now since `EthereumFeeService`'s constructor doesn't call the factory, but if that ever changes, it could lead to subtle failures. Moving the `service` creation into `@BeforeEach` after the mock setup would make the test more resilient to future changes. Not urgent at all, just a defensive coding thought.

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
```diff
@@ -0,0 +1,382 @@
...
+    @BeforeEach
+    fun setUp() {
+        every { evmApiFactory.createEvmApi(any()) } returns evmApi
+        coEvery { evmApi.getBaseFee() } returns BigInteger.ZERO
+        coEvery { evmApi.getFeeHistory() } returns emptyList()
+    }
+
+    // ---------- Bug #6 regression: empty fee history must not crash ----------
```
aminsato, line 44: Just a heads-up: since `@BeforeEach` stubs `getFeeHistory()` to return `emptyList()`, any new test that forgets to call `stubFeeHistory()` will silently test the empty-history fallback path instead of what was intended. This might be fine for now since most tests do call `stubFeeHistory()`, but it could trip someone up later. One option would be to have the default stub throw, forcing each test to explicitly set up its fee history. Just a thought — the current approach works too!

aminsato (dismissed):
## Review: fix: fall back safely when EIP-1559 fee history is empty

Great catch on the crash — this is a real bug that needed fixing. I have a few suggestions that I think could make the solution even stronger. Thanks for adding such thorough test coverage!

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
rkokhatskyi: Unnecessary overhead

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
rkokhatskyi: the data is already sorted — `EvmApi.getFeeHistory()` calls `.sorted()` on the reward list before returning it (line 435 of EvmApi.kt). the name `median()` is accurate

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
rkokhatskyi: makes sense, updated

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
rkokhatskyi: makes sense, updated

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
rkokhatskyi: `median()` is a private top-level function — can't be called from tests. testing through the pipeline is the correct approach here

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
rkokhatskyi: for integer-valued gas fees, averaging two BigIntegers is not meaningful — picking the upper element is standard. "upper midpoint" describes exactly that

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
rkokhatskyi: `EvmApi.getFeeHistory()` returns a pre-sorted list (`.sorted()` at line 435). testing with shuffled input would be testing an impossible production state

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
rkokhatskyi: the constructor is `@Inject constructor(private val evmApiFactory)` — it just stores the reference, no initialization logic. this is standard for Hilt-injected classes

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeServiceTest.kt
rkokhatskyi: the `emptyList()` default is intentional — empty fee history is the primary scenario this PR fixes. 4+ tests rely on this default. making the default throw would just add boilerplate `stubFeeHistory(emptyList())` calls everywhere for no clarity gain

johnnyluo (approved):
LGTM

PR #4081, opened 2026-04-15
fix: gracefully handle Rive SDK init failure on unsupported ABIs

### LINKED ISSUE #4068: Crash: Rive native library missing for armeabi-v7a on app startup
## Description

Crash reported in Android Vitals: `RuntimeException` during `Application.onCreate` caused by Rive SDK failing to load its native library `libandroid.so`. The APK does not contain the library for the `armeabi-v7a` ABI.

## Stack Trace

```
Exception java.lang.RuntimeException:
    at android.app.ActivityThread.handleBindApplication (ActivityThread.java:6100)
    ...
Caused by: Could not find 'libandroid.so'. Looked for: [armeabi-v7a, armeabi], but only found: [].
    at com.getkeepsafe.relinker.MissingLibraryException.<init> (MissingLibraryException.java:22)
    at com.getkeepsafe.relinker.ApkLibraryInstaller.installLibrary (ApkLibraryInstaller.java:175)
    at com.getkeepsafe.relinker.ReLinkerInstance.loadLibraryInternal (ReLinkerInstance.java:180)
    ...
    at app.rive.runtime.kotlin.core.Rive.init (Rive.kt:57)
    at com.vultisig.wallet.app.VsBaseApplication.onCreate (VsBaseApplication.java:20)
    at com.vultisig.wallet.app.Hilt_VultisigApplication.onCreate (Hilt_VultisigApplication.java:43)
    ...
```

## Device Info

- **Device**: Samsung Galaxy J7 Prime
- **OS**: Android 8.1 (API 27)
- **Date**: Apr 11, 2026

## Versions Affected

98

## Analysis

The Rive SDK calls `Rive.init()` during `VsBaseApplication.onCreate`, which uses ReLinker to load `libandroid.so`. The device is 32-bit ARM (`armeabi-v7a`), but the APK contains no native libraries for that ABI — `only found: []`.

This means either:
1. The Rive SDK AAR does not ship `armeabi-v7a` native libraries, or
2. The Gradle build is configured with `ndk.abiFilters` that exclude 32-bit ABIs (e.g., only `arm64-v8a`)

Since `minSdk` is 26, the app can still be installed on 32-bit devices. The crash is fatal — the app cannot start at all on these devices.

## Possible Fixes

- **Option A**: Wrap `Rive.init()` in a try-catch so the app can still launch (with Rive animations disabled) on unsupported ABIs
- **Option B**: If 32-bit support is intentional, ensure the Rive dependency includes `armeabi-v7a` native libs
- **Option C**: If 32-bit support is not needed, exclude 32-bit ABIs via `ndk.abiFilters` and add `android.bundle.enableUncompressedNativeLibs` to prevent Play Store from serving the app to 32-bit devices

## Source

Android Vitals

aminsato:
## Summary
- Wraps `Rive.init()` in a try-catch in `VsBaseApplication.onCreate` so the app no longer crashes on startup when the Rive native library (`libandroid.so`) is missing for the device ABI (e.g., `armeabi-v7a`)
- Adds an `isRiveInitialized` flag checked by all `RiveAnimation` composables — when Rive is unavailable, animations degrade to empty spacers instead of crashing
- Fixes fatal crash reported on Samsung Galaxy J7 Prime (Android 8.1, 32-bit ARM)

Closes #4068

## Test plan
- [ ] Verify app launches normally on arm64-v8a devices (Rive animations work as before)
- [ ] Verify app launches on armeabi-v7a device/emulator without crashing (animations simply don't render)
- [ ] Verify screens that use Rive animations (onboarding, keygen, keysign) are still functional without animations

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for the app's animation system so initialization failures no longer crash the app; animations are gracefully disabled on failure.
  * UI components now render a lightweight placeholder when animations are unavailable, preventing background loading and improving startup stability.

* **Tests**
  * Added tests validating animation initialization behavior and failure handling.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/app/VoltixApplication.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/VoltixApplication.kt
@@ -7,6 +7,9 @@ import com.vultisig.wallet.data.utils.SharedPrefsMasterKeyInitializer
 import dagger.hilt.android.HiltAndroidApp
 import timber.log.Timber
 
+internal var isRiveInitialized: Boolean = false
+    private set
+
 internal open class VsBaseApplication : Application() {
     override fun onCreate() {
         super.onCreate()
@@ -17,7 +20,12 @@ internal open class VsBaseApplication : Application() {
             Timber.plant(Timber.DebugTree())
         }
 
-        Rive.init(this)
+        try {
+            Rive.init(this)
+            isRiveInitialized = true
+        } catch (e: Exception) {
+            Timber.e(e, "Failed to initialize Rive SDK, animations will be disabled")
+        }
     }
 }
 
--- a/app/src/main/java/com/vultisig/wallet/app/VoltixApplication.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/VoltixApplication.kt
@@ -1,6 +1,8 @@
 package com.vultisig.wallet.app
 
 import android.app.Application
+import android.content.Context
+import androidx.annotation.VisibleForTesting
 import app.rive.runtime.kotlin.core.Rive
 import com.vultisig.wallet.BuildConfig
 import com.vultisig.wallet.data.utils.SharedPrefsMasterKeyInitializer
@@ -9,6 +11,20 @@ import timber.log.Timber
 
 internal var isRiveInitialized: Boolean = false
     private set
+
+@VisibleForTesting
+internal fun resetRiveInitialized() {
+    isRiveInitialized = false
+}
+
+internal fun initializeRive(context: Context) {
+    try {
+        Rive.init(context)
+        isRiveInitialized = true
+    } catch (e: Throwable) {
+        Timber.e(e, "Failed to initialize Rive SDK, animations will be disabled")
+    }
+}
 
 internal open class VsBaseApplication : Application() {
     override fun onCreate() {
@@ -20,12 +36,7 @@ internal open class VsBaseApplication : Application() {
             Timber.plant(Timber.DebugTree())
         }
 
-        try {
-            Rive.init(this)
-            isRiveInitialized = true
-        } catch (e: Throwable) {
-            Timber.e(e, "Failed to initialize Rive SDK, animations will be disabled")
-        }
+        initializeRive(this)
     }
 }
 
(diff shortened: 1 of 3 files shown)


rkokhatskyi (commented):
hey @aminsato, please cover the logic by Unit tests

app/src/main/java/com/vultisig/wallet/app/VoltixApplication.kt
```diff
@@ -17,7 +20,12 @@ internal open class VsBaseApplication : Application() {
             Timber.plant(Timber.DebugTree())
         }
 
-        Rive.init(this)
+        try {
+            Rive.init(this)
+            isRiveInitialized = true
+        } catch (e: Exception) {
```
rkokhatskyi: `UnsatisfiedLinkError` extends `Error`, not `Exception` — this won't catch the missing native lib crash. change to `Throwable` or catch `UnsatisfiedLinkError` directly

app/src/main/java/com/vultisig/wallet/ui/components/rive/RiveAnimation.kt
```diff
@@ -83,6 +82,10 @@ fun RiveAnimation(
 
 @Composable
 fun rememberRiveResourceFile(@RawRes resId: Int): State<RiveFile?> {
+    if (!isRiveInitialized) {
```
rkokhatskyi, line 86: `derivedStateOf` is for deriving from other reactive state, overkill for a constant null. just use `mutableStateOf<RiveFile?>(null)`

rkokhatskyi (dismissed):
@aminsato please check the other comments apart from missing tests

app/src/main/java/com/vultisig/wallet/app/VoltixApplication.kt
aminsato: Good catch! Changed to `catch (e: Throwable)` to properly catch `UnsatisfiedLinkError`. Fixed in d5bf33d.

app/src/main/java/com/vultisig/wallet/ui/components/rive/RiveAnimation.kt
aminsato: You're right, replaced with `mutableStateOf<RiveFile?>(null)`. Fixed in d5bf33d.

aminsato (commented):
Added unit tests for the Rive initialization logic in 63c044f:

- Extracted `initializeRive()` for testability
- 4 test cases: default state, successful init, `UnsatisfiedLinkError` caught, `RuntimeException` caught
- Also addressed both inline comments (catch `Throwable`, `mutableStateOf` instead of `derivedStateOf`)

johnnyluo (approved):
LGTM

PR #4085, opened 2026-04-15
fix: trim surrounding whitespace on pasted destination addresses

rkokhatskyi:
## Description

Destination addresses pasted from terminals, emails, or other apps often carry a trailing newline or surrounding spaces. The Send flow forwarded the raw text to name resolution, validation, and the outgoing transaction, so a valid address wrapped in whitespace silently failed the validator and blocked the Continue button.

Trim the address at every consumer boundary — resolver, validator, tx payload — while leaving the text field itself untouched, so the user still sees exactly what they pasted.

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Addresses now automatically trim surrounding whitespace when pasted or entered, preventing validation errors from unintended spacing.

* **Tests**
  * Added comprehensive test coverage for address input handling with various whitespace scenarios.


rkokhatskyi:
image

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -364,7 +365,7 @@ constructor(
                 .textAsFlow()
                 .debounce(300)
                 .combine(selectedToken.filterNotNull()) { address, token ->
```
aminsato, line 367: You changed 6 call sites but missed at least 2 more in the same file.

**`calculateGasFees()`** (~line 2543 on main): `dst.toString()` flows straight into `GasFeeInput` and eventually into `Transfer(to = resolvedDstAddress.value ?: dst)` at line 2565. If `resolvedDstAddress` is null, the untrimmed `dst` hits the transaction payload directly.

**`collectPlanFee()`** (~line 2619 on main): `dstAddress.toString()` is passed to `addressParserRepository.resolveName(...)` — same issue the PR is supposed to fix.

Both paths read from `addressFieldState.textAsFlow()` and call `.toString()` without trimming. The fix is incomplete unless these are also converted to `.asAddressInput()`.

app/src/main/java/com/vultisig/wallet/ui/utils/AddressInputExtensions.kt
```diff
@@ -0,0 +1,10 @@
+package com.vultisig.wallet.ui.utils
+
+/**
+ * Returns this user-entered destination address with surrounding whitespace stripped.
+ *
+ * Pasted addresses commonly carry leading or trailing spaces, tabs, or newlines that must not reach
+ * name resolution, validation, or the outgoing transaction payload. Apply this at the boundary
+ * before any such use.
+ */
+internal fun CharSequence.asAddressInput(): String = toString().trim()
```
aminsato, line 10: This is a whack-a-mole approach. You are trimming at 6 (really 8+) consumer sites instead of trimming once at the source.

Two options that eliminate the entire class of bug:

1. **Trim inside `textAsFlow()`** so every downstream consumer gets a clean string — zero call-site changes needed.
2. **Trim on paste** via a custom `InputTransformation` on the `TextFieldState` that strips leading/trailing whitespace on input. User never sees whitespace.

Either way, future code that reads `addressFieldState.text` cannot accidentally forget `.asAddressInput()`. The current approach guarantees someone will miss a spot — and in fact already has (see other comment).

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -383,7 +384,8 @@ constructor(
                             val resolved =
                                 addressParserRepository.resolveName(addressStr, token.chain)
                             // Ignore stale result if user changed input while resolving
-                            if (addressFieldState.text.toString() != addressStr) return@mapLatest
+                            if (addressFieldState.text.asAddressInput() != addressStr)
```
aminsato, line 387: This comparison is now asymmetric: the left side is trimmed via `.asAddressInput()` but `addressStr` (the right side) was already trimmed at line 367. So the comparison works — but only because the earlier `.asAddressInput()` already stripped it.

If someone removes the trim at line 367 thinking "the stale check handles it", this breaks silently. Either both sides should call `.asAddressInput()`, or just compare raw `.toString()` values here since both sides have already been through the trim.

app/src/test/java/com/vultisig/wallet/ui/utils/AddressInputExtensionsTest.kt
```diff
@@ -0,0 +1,86 @@
...
+    @Test
+    fun `thorname with colon is preserved`() {
+        val thorname = "rk:THOR"
+        assertEquals(thorname, " $thorname ".asAddressInput())
+    }
+
+    companion object {
+        private const val EVM_ADDRESS = "0xAf6a0cB4bA76B4720D345d09dCdB58B9a2570982"
+    }
+}
```
aminsato, line 37: 86 lines of tests for `String.trim()`. The function body is literally `toString().trim()` — you are testing Kotlin stdlib, not your code.

Keep 3-4 representative cases (clean address, trailing newline, whitespace-only, StringBuilder dispatch) and drop the rest. Tests that add no signal are maintenance burden.

aminsato (changes requested):
The intent is right — whitespace in pasted addresses is a real user-facing bug. But the fix is incomplete and the approach is fragile.

**Blocking issues:**
1. At least 2 call sites in the same file still pass untrimmed addresses to resolution and transaction building (`calculateGasFees`, `collectPlanFee`). This is a partial fix.
2. The architecture invites regressions — every future `addressFieldState.text` reader must remember to call `.asAddressInput()`. Trim at the source (either in `textAsFlow()` or via `InputTransformation`) instead of at N consumer sites.

**Non-blocking:**
3. The stale-check comparison at line 387 is asymmetric in a way that works by accident. Fragile.
4. `send()` entry guard uses `.isBlank()` on raw text — whitespace-only input bypasses the "no address" UX and falls through to the validator.
5. The test file has 13 tests for `String.trim()`. Cut it to 3-4.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
rkokhatskyi: makes sense, updated — added `.asAddressInput()` to both `calculateGasFees()` and `collectPlanFee()`

app/src/main/java/com/vultisig/wallet/ui/utils/AddressInputExtensions.kt
rkokhatskyi: trimming inside `textAsFlow()` would break password fields, seed phrase input, search fields — it is used in ~45 call sites across the app. `InputTransformation` on the address field specifically is the cleaner long-term option, but that is a separate refactor that changes UX (user sees trimmed text immediately). for now, fixed the two missed sites you pointed out

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
rkokhatskyi: both sides are trimmed — `addressStr` comes from `.asAddressInput()` at line 367. the comparison is correct. if someone restructures this code they would need to understand the data flow regardless

app/src/test/java/com/vultisig/wallet/ui/utils/AddressInputExtensionsTest.kt
rkokhatskyi: makes sense, trimmed to 5 representative cases — clean address, trailing newline, mixed whitespace, whitespace-only, and StringBuilder dispatch

rkokhatskyi:
> 2\. The architecture invites regressions — every future `addressFieldState.text` reader must remember to call `.asAddressInput()`. Trim at the source (either in `textAsFlow()` or via `InputTransformation`) instead of at N consumer sites.

FIxed 2 call sites, other comments should not be done

aminsato (approved):
LGTM

PR #4087, opened 2026-04-15
fix: show Circle Open Account failures with an error snackbar instead of success

### LINKED ISSUE #4086: [Fix] Circle Open Account failure shown as green-check success snackbar [Circle USDC DeFi]
# [Fix] Circle Open Account failure shown as green-check success snackbar [Circle USDC DeFi]

## Problem

Tapping **Open Account** on the Circle USDC DeFi screen surfaces server-side failures (e.g. a 500 from `https://api.vultisig.com/circle/create`) through the snackbar — but the snackbar renders with green-check `Success` styling while the text reads "Failed to create Circle MSCA account". The icon contradicts the message, so users can't tell at a glance whether the operation worked.

Image

Root cause: `CircleDeFiPositionsViewModel.onCreateAccount` calls `SnackbarFlow.showMessage(text)` without a type. The function signature defaults the type to `SnackbarType.Success`, so the failure branch inherits the success styling.

## Expected Behavior

On failure, the snackbar should render with the red `error_critical` icon and `alerts.error` color (i.e. `SnackbarType.Error`) — the same styling every other ViewModel in the app uses for failed async operations.

## Steps to Reproduce

1. Open a vault that has never opened a Circle USDC account
2. Go to DeFi → Circle USDC
3. Tap **Open Account**
4. Observe the snackbar styling vs text

## Solution

Pass `SnackbarType.Error` on the failure branch (and make the success branch explicit to match). Apply the same fix to any adjacent snackbar calls that rely on the default.

### Must Do
- [ ] Snackbar type on the failure branch of `onCreateAccount` is explicitly `SnackbarType.Error`
- [ ] Success branch explicitly passes `SnackbarType.Success`
- [ ] Unit tests for both branches

### Must NOT Do
- Don't change unrelated code
- Don't refactor the broader Circle flow

## Acceptance Criteria
- [ ] Tapping Open Account while the Circle API is returning 5xx shows a red error snackbar
- [ ] Tapping Open Account on success shows a green success snackbar

## Context

Sample log captured on mainnet:
```
Ktor Client  RESPONSE: 500
  {"code":155509,"message":"Entity needs to setup paymaster policy in developer console before sending transaction on mainnet",
   "errors":[{"message":"15002: there must be at least one activated policy"}]}
CircleDeFi...mscAddress  E  NetworkException: {"code":155509,...}
```

The 500 itself is a server-side Circle paymaster-policy configuration issue (out of scope for the Android fix) — but it makes the UX bug reproducible (temp issue on Circle API).

rkokhatskyi:
Fixes #4086

## Description

The "Open Account" action on the Circle USDC DeFi screen called `SnackbarFlow.showMessage(...)` without specifying a type on the failure branch. The signature defaults to `SnackbarType.Success`, so a real network error rendered the "Failed to create Circle MSCA account" text inside the green-check success snackbar — the icon contradicted the words.

Pass `SnackbarType.Error` on failure (and make the success branch explicit too). The create-account flow is now a clean `runCatching` with `onSuccess`/`onFailure`, and a small `showSnackbar(@StringRes, SnackbarType)` helper removes the repeated string-lookup boilerplate.

**Side benefit:** the `mscaAddress` property assignment moves out of the `withContext(Dispatchers.IO)` block and into the `onSuccess` branch, which resumes on the Main dispatcher — so ViewModel state mutation now happens on the correct thread.

## Before / After

| Before | After |
|--------|-------|
| Screenshot_20260415_165409 | image |

## Which feature is affected?

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and user-facing feedback during Circle DeFi account creation, providing clearer, more informative error messages to help users understand and resolve issues

* **Tests**
  * Added comprehensive automated test coverage for Circle DeFi account creation scenarios, including network failures, successful account creation, and validation edge cases


aminsato (commented):
Nice fix — the root cause (success snackbar on failure) is clearly diagnosed in the PR description and the solution is clean. A few things I'd like to raise for consideration below.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
```diff
@@ -209,38 +211,39 @@ constructor(
...
-                    circleDefi = currentState.circleDefi.copy(isAccountOpen = mscAddress != null)
+                    circleDefi =
+                        currentState.circleDefi.copy(isAccountOpen = createdAddress.isSuccess)
                 )
             }
         }
     }
 
+    private suspend fun showSnackbar(@StringRes messageRes: Int, type: SnackbarType) {
+        snackbarFlow.showMessage(StringResource(messageRes).asString(context), type)
```
aminsato, line 244: Nice catch on the snackbar type — this is a solid improvement!

One thing worth considering: the `_state.update` block below runs unconditionally after both `onSuccess` and `onFailure`. Since `onSuccess`/`onFailure` don't short-circuit (they just run callbacks and return the same `Result`), the state update always executes — which is the intent here.

However, `isAccountOpen = createdAddress.isSuccess` means a transient network failure will set `isAccountOpen = false`, potentially flickering the UI if the account was *already* open before the user tapped "create" again (e.g., a retry after a timeout). It might be worth guarding this with something like:

```kotlin
if (createdAddress.isFailure) return@launch
// or only update isAccountOpen on success
```

Just a thought — if double-tap / retry isn't a real scenario here, feel free to disregard.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
```diff
@@ -209,38 +211,39 @@ constructor(
...
+    }
+
     fun onDepositAccount() {
         viewModelScope.launch {
             try {
```
aminsato, line 249: The helper is a nice de-duplication — thank you for extracting it.

One small concern: `StringResource(messageRes).asString(context)` resolves the string eagerly using the injected `@ApplicationContext`. This is fine today, but if the snackbar ever needs to respect the user's locale change at display time (rather than at emission time), this would lock in the locale from when the message was created. 

Not blocking — just flagging for awareness. If the team convention is to always resolve strings at emission time, this is perfectly fine.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
```diff
@@ -209,38 +211,39 @@ constructor(
...
-            }
+            createdAddress
+                .onSuccess { newAddress ->
+                    mscaAddress = newAddress
+                    showSnackbar(R.string.circle_msca_account_created_success, SnackbarType.Success)
+                }
+                .onFailure { t ->
+                    Timber.e(t)
+                    showSnackbar(R.string.circle_msca_account_created_failed, SnackbarType.Error)
+                }
```
aminsato, line 233: I appreciate the cleaner flow with \`runCatching\` — much easier to follow than the old try/catch-returns-null pattern.

One observation: in the old code, \`this@CircleDeFiPositionsViewModel.mscaAddress = mscaAddress\` was set *inside* the IO block. In the new code, this assignment moved to \`onSuccess\` which runs on the calling dispatcher (the viewModelScope's main dispatcher). That's actually a nice improvement for thread safety — just wanted to confirm this was intentional and not accidental, since it's a behavioral change not mentioned in the PR description.

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
```diff
@@ -0,0 +1,169 @@
...
+        coEvery { scaCircleAccountRepository.getAccount(VAULT_ID) } returns null
+        coEvery { circleApi.createScAccount(OWNER_ADDRESS) } throws
+            NetworkException(500, "Entity needs to setup paymaster policy")
+
+        vm.onCreateAccount()
+
+        coVerify(exactly = 1) { snackbarFlow.showMessage(any(), SnackbarType.Error) }
+        coVerify(exactly = 0) { snackbarFlow.showMessage(any(), SnackbarType.Success) }
+        coVerify(exactly = 0) { scaCircleAccountRepository.saveAccount(any(), any()) }
+    }
```
aminsato, line 103: Good test coverage — I like that you're testing the failure, success, and vault-null scenarios separately.

A small suggestion: the \`clearMocks(..., answers = false, childMocks = false)\` + re-stub pattern appears in all three tests. This could be a bit fragile — if a future test forgets to re-stub after the clear, it will silently fall back to relaxed mock defaults (returning \`\"\"\` or \`0\`), which you've already identified as tricky (see the comment on line 88 about relaxed mocks returning \`\"\"\`).

Would it be worth extracting a small helper like \`fun givenNoExistingAccount()\` to encapsulate the clear + re-stub? That way each test reads as setup → action → assertion without the mock plumbing in between. Just a readability thought — the tests are correct as-is.

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
```diff
@@ -0,0 +1,169 @@
...
+        vm.onCreateAccount()
+
+        coVerify(exactly = 1) { snackbarFlow.showMessage(any(), SnackbarType.Error) }
+        coVerify(exactly = 0) { circleApi.createScAccount(any()) }
+        coVerify(exactly = 0) { scaCircleAccountRepository.saveAccount(any(), any()) }
+    }
+
+    private fun createViewModel(): CircleDeFiPositionsViewModel =
+        CircleDeFiPositionsViewModel(
+            navigator = navigator,
```
aminsato, line 169: This is a great edge case to cover — thank you for including it.

One thing to double-check: when \`vaultRepository.get(VAULT_ID)\` returns \`null\`, \`getEvmVaultAddress()\` calls \`error(...)\` which throws an \`IllegalStateException\`. This exception propagates through the \`runCatching\` block and gets caught by \`onFailure\`.

However, the *outer* \`withContext(Dispatchers.IO)\` wrapping the \`runCatching\` means the \`IllegalStateException\` is thrown *inside* the \`runCatching\` — so it works. But if someone ever refactors to move the \`runCatching\` inside the \`withContext\` (or vice versa), the behavior could subtly change. 

It might be worth adding a brief comment in the test explaining that this test validates the \`getEvmVaultAddress() → error()\` path is caught by \`runCatching\`, so future readers understand the exact failure chain being tested.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
```diff
@@ -209,38 +211,39 @@ constructor(
...
-                        scaCircleAccountRepository.saveAccount(vaultId, mscaAddress)
-                        this@CircleDeFiPositionsViewModel.mscaAddress = mscaAddress
-                        mscaAddress
-                    } catch (t: Throwable) {
-                        Timber.e(t)
-                        null
+                        circleApi.createScAccount(ethereumVaultAddress).also { newAddress ->
+                            scaCircleAccountRepository.saveAccount(vaultId, newAddress)
+                        }
                     }
```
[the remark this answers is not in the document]
rkokhatskyi: Good catch. This is pre-existing behaviour — the original `try { createScAccount; saveAccount }` collapsed both into one catch too, so this PR preserves that semantic rather than introducing the issue. I'd prefer to keep this PR scoped to the snackbar-type bug.

Tracked separately:
- #4088 — split the `runCatching` so a local cache failure doesn't revert `isAccountOpen`
- #4089 — the related (and bigger) retry-creates-duplicate-account issue, since each `createScAccount` mints a fresh `idempotency_key` on retry

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
rkokhatskyi: Good spot — for this flow the UI flips the button label to 'Deposit USDC' once isAccountOpen is true, so the 'Open Account → retry' path isn't user-reachable through that button. Related concern about silent duplicate-account creation on retry is tracked separately in #4089.

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
rkokhatskyi: Good suggestion — applied in de2d3ef24 as `givenNoExistingAccount()`. Tests now read as `createViewModel().setData(...)` → helper → specific stubs → action → assertion.

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
rkokhatskyi: Applied in de2d3ef24 — added a comment above the `vaultRepository.get returns null` stub explaining that the test is asserting the `getEvmVaultAddress() → error() → runCatching` catch chain.

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+        // must use a real dispatcher.
+        return withContext(Dispatchers.Default.limitedParallelism(1)) {
+            withTimeout(5.seconds) { captured.await() }
+        }
+    }
+
+    private fun createViewModel(): CircleDeFiPositionsViewModel =
+        CircleDeFiPositionsViewModel(
+            navigator = navigator,
+            scaCircleAccountRepository = scaCircleAccountRepository,
```
aminsato, line 170: Thank you for the `awaitSnackbarAfter` helper — it is a clever way to deal with the IO dispatcher hop, and the KDoc makes the intent very clear.

A couple of small observations for your consideration:

- The `Dispatchers.Default.limitedParallelism(1)` wrapper may not be adding much here, since only a single `CompletableDeferred` is awaited and there is no concurrency to constrain. Plain `Dispatchers.Default` should be enough to escape the `runTest` virtual scheduler so `withTimeout` uses real time.
- More broadly, it might be worth considering a follow-up to inject a `CoroutineDispatcher` (or a `DispatcherProvider`) into the ViewModel. That would let `UnconfinedTestDispatcher` substitute for `Dispatchers.IO` in tests, and the helper could eventually go away entirely. Definitely out of scope for this PR — just a thought for a future refactor so new tests on this ViewModel do not all have to reach for the same pattern.

Nothing blocking — the helper works well as-is.

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+        givenNoExistingAccount()
+        coEvery { circleApi.createScAccount(OWNER_ADDRESS) } throws
+            NetworkException(500, "Entity needs to setup paymaster policy")
+
+        val type = awaitSnackbarAfter { vm.onCreateAccount() }
+
+        assertEquals(SnackbarType.Error, type)
+        coVerify(exactly = 0) { scaCircleAccountRepository.saveAccount(any(), any()) }
+    }
+
```
aminsato, line 104: The tests verify the snackbar type really nicely — thanks for including the vault-null edge case too.

One small suggestion for your consideration: since `isAccountOpen` is what drives the UI transition from "Open Account" to "Deposit USDC" (per your helpful explanation on the earlier thread), it might be worth adding an assertion on `vm.state.value.circleDefi.isAccountOpen` in both the success and failure tests — something like:

```kotlin
// success path
assertTrue(vm.state.value.circleDefi.isAccountOpen)

// failure path
assertFalse(vm.state.value.circleDefi.isAccountOpen)
```

This would give the tests an extra bit of coverage on the user-visible outcome, and guard against anyone accidentally changing `isAccountOpen = createdAddress.isSuccess` in a way that decouples it from the result. Totally optional.

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
```diff
@@ -0,0 +1,196 @@
...
+            }
+        action()
+        // onCreateAccount hops to the real Dispatchers.IO for the network call. The test
+        // scheduler's virtual clock doesn't advance for real-thread work, so the timeout
+        // must use a real dispatcher.
+        return withContext(Dispatchers.Default.limitedParallelism(1)) {
+            withTimeout(5.seconds) { captured.await() }
+        }
+    }
+
```
aminsato, line 166: The KDoc here is genuinely helpful for future readers — thank you for taking the time to write it.

One small thing worth flagging, mostly for future-proofing: the helper relies on `snackbarFlow.showMessage` being called exactly once and before any state mutation for the `CompletableDeferred` to resolve correctly. That is true of the current implementation, but it is a behavioral coupling rather than a structural guarantee.

If `onCreateAccount` ever evolves to call `showMessage` more than once (e.g. a progress → done pattern), `captured.complete()` would silently no-op on the second call and the test could mask real regressions.

If you would like to harden this, capturing into a `Channel` or `List` and asserting the exact sequence would make the helper robust to that kind of change. Alternatively, a short note on the helper stating the precondition explicitly would be helpful. Not blocking — the current tests are correct for the current code.

aminsato (commented):
Follow-up review on the latest revision (de2d3ef). Thank you for the thorough responses to the prior feedback — the `givenNoExistingAccount()` helper extraction and the added comment explaining the runCatching chain are both really nicely done, and make the tests noticeably easier to read.

A few gentle observations on the updated test file, all non-blocking:

1. The `awaitSnackbarAfter` helper is a clever workaround for the `Dispatchers.IO` hop. It might be worth considering injecting a `CoroutineDispatcher` into the ViewModel as a future refactor, which could let the helper go away entirely.
2. `Dispatchers.Default.limitedParallelism(1)` inside the helper likely is not doing much here — plain `Dispatchers.Default` should be sufficient.
3. It could be worth asserting on `state.value.circleDefi.isAccountOpen` in the success and failure tests, since that is what drives the visible button transition.
4. The helper has an implicit precondition (snackbar called exactly once, before state update) that could be worth documenting more explicitly or hardening with a `Channel`-based capture.

All minor — the fix itself is in great shape. Thank you for the care you put into iterating on this!

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
rkokhatskyi: the UI already gates this — once `isAccountOpen` is true, the button becomes "Deposit USDC" and calls `onDepositAccount()`. the retry scenario is not reachable through the UI flow

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
rkokhatskyi: this is how every ViewModel in the codebase resolves strings for snackbars — `SnackbarFlow.showMessage` takes `String`, not `UiText`. changing one call site would be inconsistent. Android recreates activities on locale change anyway

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
rkokhatskyi: intentional — keeps all reads and writes of `mscaAddress` on the main dispatcher since `viewModelScope` runs on `Dispatchers.Main`. safer than assigning inside the IO block

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
rkokhatskyi: `givenNoExistingAccount()` already exists at line 137 and is used by all three tests

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
rkokhatskyi: already documented — see the 3-line comment at lines 120-122 above the mock setup

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
rkokhatskyi: `limitedParallelism(1)` is load-bearing — production code does `withContext(Dispatchers.IO)` which hops to a real thread. without `Dispatchers.Default`, `withTimeout` uses the test scheduler's virtual time and never advances for real-thread work. removing it breaks the tests. dispatcher injection is out of scope as you noted

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
rkokhatskyi: makes sense, updated

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
rkokhatskyi: `onCreateAccount` calls `showMessage` exactly once — `onSuccess` and `onFailure` are mutually exclusive via `runCatching`. `CompletableDeferred` is the right primitive for a single expected signal

rkokhatskyi:
Hi @aminsato, only the test assertion is minor but valid comment. All others are AI noise comments, please polish your review prompt. Thanks!

aminsato (approved):
LGTM

PR #4092, opened 2026-04-16
ci: upload debug APK as workflow artifact

aminsato:
## Summary
- Adds an `actions/upload-artifact@v4` step to publish the debug APK built by the Android CI workflow so it can be downloaded from the run's artifacts page.
- Fails the step (`if-no-files-found: error`) when no APK is produced, so path drift is caught early.

## Test plan
- [ ] CI run succeeds and the `app-debug-apk` artifact appears on the workflow run summary.
- [ ] Download the artifact and confirm it contains the expected `app-debug.apk`.

## Summary by CodeRabbit

* **Chores**
  * Enhanced CI/CD pipeline to automatically archive debug builds, improving the testing and build artifact management process.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -98,3 +98,12 @@ jobs:
           path: |
             app/build/reports/lint-results-debug.html
             app/build/intermediates/lint_intermediate_text_report/debug/lintReportDebug/lint-results-debug.txt
+
+      # Archive debug APK
+      - name: Archive debug APK
+        if: success()
+        uses: actions/upload-artifact@v4
+        with:
+          name: app-debug-apk
+          path: app/build/outputs/apk/debug/*.apk
+          if-no-files-found: error


.github/workflows/android.yml
```diff
@@ -98,3 +98,12 @@ jobs:
           path: |
             app/build/reports/lint-results-debug.html
             app/build/intermediates/lint_intermediate_text_report/debug/lintReportDebug/lint-results-debug.txt
+
+      # Archive debug APK
+      - name: Archive debug APK
+        if: success()
+        uses: actions/upload-artifact@v4
```
rkokhatskyi, line 105: it may be useful for testers, however all devs can run locally.

Also this would mean we upload the debug builds on every PR, which is redundant and will consume a lot of quota of Github storage. What's the intention of attaching the debug APK builds as part of CI?

realpaaao: Yes, agree. @aminsato we dont need the debug in GH in my POV.

johnnyluo:
Let's close this then

PR #4096, opened 2026-04-16
feat: Add delete confirmation dialog to address book

rkokhatskyi:
## Summary

Tapping the trash icon on an address book entry used to delete it immediately — one accidental tap and the contact is gone. Now it shows a confirmation dialog first.

- Added a "Delete address?" dialog with the entry name, a red Delete button, and a Cancel option
- Used `safeLaunch` so if the DB delete fails the user isn't left confused (dialog dismissed but entry still there)
- Extracted a reusable `UiConfirmDialog` component since we had the same AlertDialog boilerplate in multiple places
- Fixed a grammatical gender error in the Croatian translation
- Added a double-tap guard on the confirm button
- Added a Compose preview for the dialog

All new strings are translated across all 9 non-English locales.

## Screenshots

image

## Test plan

- [ ] Open Address Book with at least one entry
- [ ] Tap edit mode, tap the trash icon — confirm dialog appears
- [ ] Tap Cancel — dialog dismisses, entry still there
- [ ] Tap Delete — entry is removed from the list
- [ ] Tap trash, then quickly double-tap Delete — only one deletion fires
- [ ] Rotate the device while dialog is open — dialog dismisses gracefully (no crash)
- [ ] Unit tests pass: `./gradlew :app:testDebugUnitTest --tests '*AddressBookViewModelTest*'`

## Summary by CodeRabbit

* **New Features**
  * Two-step confirmation flow for deleting address-book entries with a reusable confirm dialog component and UI wiring to prevent accidental removals.
* **Localization**
  * Added deletion-confirmation strings across multiple locales (10 languages) so dialogs display translated titles, messages, and action labels.
* **Tests**
  * Added unit tests covering the staged-delete, cancel, confirm, error handling, and state transitions for the new flow.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/transaction/AddressBookViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/transaction/AddressBookViewModel.kt
@@ -14,6 +14,7 @@ import com.vultisig.wallet.data.models.logo
 import com.vultisig.wallet.data.repositories.AddressBookRepository
 import com.vultisig.wallet.data.repositories.RequestResultRepository
 import com.vultisig.wallet.data.repositories.order.OrderRepository
+import com.vultisig.wallet.data.utils.safeLaunch
 import com.vultisig.wallet.ui.navigation.Destination
 import com.vultisig.wallet.ui.navigation.Navigator
 import com.vultisig.wallet.ui.navigation.Route
@@ -26,6 +27,7 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import timber.log.Timber
 
 internal data class AddressBookUiModel(
     val isEditModeEnabled: Boolean = false,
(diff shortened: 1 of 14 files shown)


johnnyluo:
@coderabbitai review please

aminsato (commented):
Light review — nothing blocking, just a few notes for clarity.

app/src/main/java/com/vultisig/wallet/ui/models/transaction/AddressBookViewModel.kt
```diff
@@ -135,10 +138,25 @@ constructor(
...
+        state.update { it.copy(pendingDeletion = model) }
+    }
+
+    fun cancelDeleteAddress() {
+        state.update { it.copy(pendingDeletion = null) }
+    }
+
+    fun confirmDeleteAddress() {
+        val target = state.value.pendingDeletion ?: return
+        viewModelScope.safeLaunch(
```
aminsato, line 152: `pendingDeletion` is read here but only cleared inside `safeLaunch` (line 159). If `confirmDeleteAddress()` is called twice before the coroutine starts, both calls see a non-null `target` and launch two parallel deletes.

Clear it eagerly before launching:
```kotlin
fun confirmDeleteAddress() {
    val target = state.value.pendingDeletion ?: return
    state.update { it.copy(pendingDeletion = null) }
    viewModelScope.safeLaunch(
        onError = { e -> Timber.e(e, "Failed to delete address book entry") }
    ) {
        addressBookRepository.delete(target.model.chain.id, target.model.address)
        orderRepository.delete(null, target.model.id)
        loadData()
    }
}
```
This also makes the `confirmedOnce` UI guard in `DeleteAddressDialog` unnecessary.

johnnyluo (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/transaction/AddressBookViewModel.kt
johnnyluo: addressed

PR #4098, opened 2026-04-16
feat: Show user friendly error messages instead of raw exceptions

rkokhatskyi:
## Summary

When a transaction fails due to a network error or unexpected exception the app was showing the raw Java exception message to the user. Things like "java.io.IOException: Network is unreachable" or "SQLiteException: UNIQUE constraint failed" would appear in the error dialog which is obviously not helpful.

This replaces all those raw `e.message` strings with proper localized error messages ("Something went wrong") and makes sure the actual exception is always logged via Timber so we can still debug issues from logcat.

Covers 10 files across send, swap, staking, seed import, vault password verification, DeFi positions and referral flows. The keysign and keygen ViewModels are intentionally left out since they sit on the TSS boundary and need separate review.

## What changed

- `SendFormViewModel` (8 catch blocks) now logs the exception and shows a generic error string
- `SwapFormViewModel` small swap amount fallback no longer leaks the raw message
- `KeysignFlowViewModel` (3 catch blocks) replaced raw message and a hardcoded English string with `R.string.unknown_error`
- `ImportSeedphraseViewModel` replaced raw exception message with the existing `error_view_default_description` string and added Timber logging
- `TronDeFiPositionsViewModel` dropped the `e.message` branch since Timber already logs it
- `VerifyExistingVaultPasswordViewModel` and `VerifyExistingVaultViewModel` password check error now shows a generic message instead of raw API response
- `FastVaultPasswordReminderViewModel` same pattern
- `ReferralVaultListViewModel` added missing Timber log and replaced raw string
- Updated `ImportSeedphraseViewModelTest` assertions to match the new behavior

## Test plan

- [ ] Send any token and disconnect wifi mid transaction. Error dialog should say "Something went wrong" not a Java stack trace
- [ ] Try a swap with an amount below the minimum. Error should show the minimum amount or "Swap amount is too low" not a raw string
- [ ] Import a seed phrase while offline. Error should be user friendly
- [ ] Enter wrong password on a fast vault. Error should not expose API internals
- [ ] Unit tests pass: `./gradlew :app:testDebugUnitTest` (259 tests all green)

## Release Notes

* **Bug Fixes**
  * Standardized error messages across vault operations, seed phrase import, transactions, swaps, and key management features to display consistent, user-friendly localized descriptions.
  * Improved error messaging to prevent technical exception details from appearing in user-facing notifications, providing clearer feedback.
  * Enhanced error logging and exception tracking to support better failure diagnostics and user support.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keygen/VerifyExistingVaultViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keygen/VerifyExistingVaultViewModel.kt
@@ -245,7 +245,7 @@ constructor(
             onError = { e ->
                 Timber.e(e, "Failed to verify vault password")
                 passwordInnerState.value = VsTextInputFieldInnerState.Error
-                passwordErrorMessage.value = StringResource(R.string.fast_vault_invalid_password)
+                passwordErrorMessage.value = StringResource(R.string.dialog_default_error_body)
                 _uiState.update { it.copy(isLoading = false) }
             }
         ) {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keygen/VerifyExistingVaultViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keygen/VerifyExistingVaultViewModel.kt
@@ -281,7 +281,8 @@ constructor(
                 }
                 is PasswordCheckResult.Error -> {
                     passwordInnerState.value = VsTextInputFieldInnerState.Error
-                    passwordErrorMessage.value = UiText.DynamicString(result.message)
+                    passwordErrorMessage.value =
+                        UiText.StringResource(R.string.dialog_default_error_body)
                 }
             }
             _uiState.update { it.copy(isLoading = false) }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/referral/ReferralVaultListViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/referral/ReferralVaultListViewModel.kt
@@ -19,6 +19,7 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import timber.log.Timber
 
 internal data class ReferralVaultListUiState(
     val vaults: List<VaultItem> = emptyList(),
@@ -82,7 +83,8 @@ constructor(
 
                 _state.update { it.copy(vaults = vaultItems, error = null) }
             } catch (e: Exception) {
-                _state.update { it.copy(error = e.message ?: "Failed to load vaults") }
+                Timber.e(e, "Failed to load referral vaults")
+                _state.update { it.copy(error = "Something went wrong") }
             }
         }
     }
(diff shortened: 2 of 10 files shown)


rkokhatskyi:
@coderabbitai full review

aminsato:
Let's exclude the following files from this PR:

- `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt`
- `app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt`

These screens can surface edge-case errors that are highly dependent on the user's environment (network conditions, specific chain/token combinations, device state, etc.) and are not reliably reproducible by developers under normal conditions. Replacing the raw exception messages with generic, friendly strings would hide information we actually rely on for debugging.

For now, keeping the original error messages visible lets users screenshot the exact failure and report it in our Discord channels, which has been the primary way we diagnose these issues. Once we have better telemetry or a more structured error-reporting path for Send/Swap, we can revisit and apply the friendly-error treatment here as well.

rkokhatskyi:
hey @aminsato, your concern is valid. The idea I had in mind is to shorten the message for all users and allow users to enable more detailed messages in settings if they need to understand more what's the happening and take a screenshot for us. This is how other wallets are evolving now, long messages are no longer a standard (some wallets use analytics tracking, some of them allow to enable developer mode or similar)

@realpaaao what do you think about this idea in general to keep all messages short and nice (and also translated) and then allow users to enable more detailed messages in settings?

app/src/main/java/com/vultisig/wallet/ui/models/keygen/VerifyExistingVaultViewModel.kt
```diff
@@ -245,7 +245,7 @@ constructor(
             onError = { e ->
                 Timber.e(e, "Failed to verify vault password")
                 passwordInnerState.value = VsTextInputFieldInnerState.Error
-                passwordErrorMessage.value = StringResource(R.string.fast_vault_invalid_password)
+                passwordErrorMessage.value = StringResource(R.string.dialog_default_error_body)
```
johnnyluo, line 248: we should revert this line of change , back to R.string.fast_vault_invalid_password

app/src/main/java/com/vultisig/wallet/ui/models/referral/ReferralVaultListViewModel.kt
```diff
@@ -82,7 +85,10 @@ constructor(
 
                 _state.update { it.copy(vaults = vaultItems, error = null) }
             } catch (e: Exception) {
-                _state.update { it.copy(error = e.message ?: "Failed to load vaults") }
+                Timber.e(e, "Failed to load referral vaults")
```
aminsato, line 88: nit: `error` is set here but never rendered in `ReferralVaultListScreen` — it's silently swallowed. The type change from `String?` to `UiText?` is fine but the field is still dead code.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -1119,7 +1119,8 @@ constructor(
             } catch (e: InvalidTransactionDataException) {
                 showError(e.text)
             } catch (e: Exception) {
-                showError(e.message?.asUiText() ?: UiText.Empty)
+                Timber.e(e, "Transaction failed")
```
aminsato, line 1122: nit (outside scope): this identical `catch (InvalidTransactionDataException) / catch (Exception)` block is repeated 8 times in this file. Could be worth extracting into a helper.

rkokhatskyi:
We will do a different approach as discussed in the chat, new PR will be created

PR #4099, opened 2026-04-16
fix: Await ERC-20 approval confirmation before broadcasting swap tx

rkokhatskyi:
## Summary

When performing an ERC-20 token swap that requires an approval transaction (e.g. swapping USDC via THORChain), the app previously broadcast the approval and swap transactions back to back without waiting for the approval to be mined. On congested networks or with load-balanced RPCs, the swap tx could land on a node that hasn't seen the approval yet, causing the swap to fail. The user pays gas for a failed transaction.

This fix polls `eth_getTransactionReceipt` after the approval broadcast and only proceeds to broadcast the swap once the approval is confirmed on-chain (`status: 0x1`).

## What changed

**New: `AwaitApprovalConfirmationUseCase`** (`data/usecases/`)
- Polls `eth_getTransactionReceipt` every 4s with a 120s timeout via `withTimeoutOrNull`
- Returns on confirmation, throws on revert or timeout
- Injected via Hilt (`DataUsecasesModule`)

**Modified: `KeysignViewModel.broadcastTransaction()`**
- Calls `awaitApprovalConfirmation(chain, approveTxHash)` between the approval broadcast and the swap broadcast
- Only runs when `approvePayload != null` (ERC-20 swaps). Regular sends and native transfers are unaffected

## Scope

The change is scoped exclusively to the `if (approvePayload != null)` block inside `broadcastTransaction()`. All non-approval flows (native sends, non-EVM chains, direct ERC-20 transfers) are completely unaffected since `approvePayload` defaults to `null` and is only set for DEX swap approvals.

## Test plan

- [x] Unit tests for the use case (6 tests: immediate confirm, polling, revert, timeout, revert after pending, chain verification)
- [ ] Manual: initiate an ERC-20 swap (e.g. USDC -> ETH via THORChain) and verify the approval is confirmed before the swap broadcasts
- [ ] Manual: verify a native ETH send is not affected (no polling delay)

## Summary by CodeRabbit

* **New Features**
  * App now waits for on-chain token approval confirmation before proceeding, handling confirmed, timed-out, and failed outcomes.

* **Bug Fixes**
  * Transaction flow halts and shows a clear approval error (with link) when approval times out or fails, preventing unintended continuation.

* **Localization**
  * Added approval error messages in multiple languages.

* **Tests**
  * Added tests covering approval polling: confirmed, failed, and timed-out scenarios.


data/src/main/kotlin/com/vultisig/wallet/data/usecases/AwaitApprovalConfirmationUseCase.kt
```diff
@@ -0,0 +1,52 @@
...
+            }
+        }
+
+        return ApprovalConfirmationResult.TimedOut
+    }
+
+    private companion object {
+        const val RECEIPT_SUCCESS = "0x1"
+        const val RECEIPT_FAILURE = "0x0"
+        const val MAX_BLOCKS = 5
```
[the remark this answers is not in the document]
rkokhatskyi: It is optimal and should remain as 5 blocks, otherwise waiting time too long and results to a bad UX.

The suggestion is not valid

rkokhatskyi:
Merged latest main (157 commits) — 10 strings.xml conflicts (PR adds `swap_error_approval_*`, main adds `erc20_approval_*`, both kept) plus a stale commondata pin (bumped to match main so the `TonMessage.state_init` field that `TonHelper.kt` references from #4283 actually resolves). Both `KeysignViewModelSaveTransactionHistoryTest` and `KeysignViewModelTryUpdateEvmActualFeeTest` added on main needed the new `awaitApprovalConfirmation` constructor parameter wired up. All `:app:testDebugUnitTest` and `:data:testDebugUnitTest` green.

Picked up four real concerns flagged during re-review:

- **Timeout matched the PR description.** The previous `repeat(MAX_BLOCKS=5)` × `delay(chain.blockTimeMs)` shape produced a 5-second budget on Arbitrum/ZkSync/Sei/Hyperliquid and 60 seconds on Ethereum — much shorter than the "120s timeout via `withTimeoutOrNull`" stated in the PR description, and short enough to false-timeout a healthy approval on a fast L2. Replaced with `withTimeoutOrNull(120_000)` wrapping a `while(true)` loop, chain-aware delay floored at `MIN_POLL_DELAY_MS = 2_000` to avoid hammering fast-block-time RPCs.
- **Cancellation safety.** `EvmApi.getTxStatus` catches `Exception` broadly (mapping RPC failures to null = "pending"), which also swallows `CancellationException`. Without protection, a cancelled `viewModelScope` would burn the full 120s budget polling a dead session. Added `currentCoroutineContext().ensureActive()` at the top of every poll.
- **`approveTxLink` populated on failure paths.** `KeysignViewModel.broadcastTransaction` previously set `approveTxLink` only after a successful swap broadcast, so users hitting `TimedOut` or `Failed` saw the error screen with no way to find their approval on the explorer. Now populated in both early-return branches.
- **Wording corrected across all 10 locales.** `swap_error_approval_failed` fires on receipt status `0x0` — an on-chain revert, not a "rejection" by a signer. Replaced "Token approval was rejected" with "Token approval transaction failed" in EN and translated equivalents in de/es/hr/it/ko/nl/pt/ru/zh-rCN using each locale's established "failed" verb (German `fehlgeschlagen`, Russian `не удалась`, etc.).

Also added `Timber.w` logging on both failure paths with `approveTxHash` + `chain` so support can correlate user reports to the approval that succeeded on-chain.

Test updates: `AwaitApprovalConfirmationUseCaseImplTest` now pins the wall-clock budget (40 polls for BSC at 3s, 60 polls for Arbitrum floored to 2s) instead of the old "5 attempts" semantics. The "Confirmed on last attempt" test became "Confirmed after many pending polls within the timeout".

Items deliberately not changed (per "no over-engineering"): cross-RPC validation, auto-revoke approval on failure, "Waiting for approval" UI state, the `tryAgain` button wiring (currently calls `onBack`), the duplicate-polling concern with `PollingTxStatusUseCase` (separate refactor opportunity, not blocking this fix).

Ready for review.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -656,6 +659,34 @@ constructor(
...
+                ApprovalConfirmationResult.TimedOut -> {
+                    Timber.w(
+                        "Approval tx %s timed out waiting for confirmation on %s",
+                        approveTxHash.value,
+                        chain,
+                    )
+                    approveTxLink.value =
+                        explorerLinkRepository.getTransactionLink(chain, approveTxHash.value)
+                    currentState.value =
+                        KeysignState.Error(R.string.swap_error_approval_timeout.asUiText())
```
aminsato, line 705: `swap_error_approval_timeout` currently says "Please try again". This is misleading — the approval transaction has already been broadcast and the user has paid gas. If they retry, they will pay for a second approval. The approval may still confirm on-chain after the in-app timeout elapses.

Suggested wording (update across all 10 locale files):
"Token approval is taking longer than expected. Check the explorer before retrying."

The `swap_error_approval_failed` string below is fine as-is, since for a reverted approval retrying IS the right action.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
```diff
@@ -427,6 +427,25 @@ val Chain.getDustThreshold: BigInteger
             else -> throw UnsupportedOperationException("Chain ${this.name} does not support Dust")
         }
 
+val Chain.blockTimeMs: Long
```
aminsato, line 439: Please add a KDoc on this extension. Worth documenting:

- What the value represents (approximate block production interval used as the poll cadence for receipt-based confirmation waits).
- That all 13 current EVM chains are enumerated explicitly, and the `else -> 4_000L` branch is a defensive default for any future EVM chain added to `Chain`.
- That callers are expected to apply their own floor (e.g. `MIN_POLL_DELAY_MS` in `AwaitApprovalConfirmationUseCase`) so very-fast L2s do not get hammered.

Example:

```kotlin
/**
 * Approximate block production interval for the chain, in milliseconds.
 *
 * Used as the polling cadence when waiting for an EVM transaction receipt.
 * All EVM chains supported today are listed explicitly; the fallback covers
 * future EVM chains added to [Chain] without forcing every caller to update.
 * Callers should still clamp the value to a sensible floor to avoid hammering
 * fast L2 RPC endpoints.
 */
val Chain.blockTimeMs: Long
```

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
rkokhatskyi: Good point, fixed in 749cfbd8c. Updated `swap_error_approval_timeout` across all 10 locales to "Token approval is taking longer than expected. Check the explorer before retrying." — each translation reuses the locale's established explorer noun and approval verb so it reads naturally. The companion `approveTxLink.value` assignment in `KeysignViewModel.broadcastTransaction` (added in the previous commit) ensures the explorer link is actually populated on the timeout path, so the new wording is actionable.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
rkokhatskyi: Added in 749cfbd8c. KDoc covers what the value represents (approximate block production interval, polling cadence for receipt waits), the explicit-list-plus-defensive-default rationale, and the caller-applied floor expectation pointing at `MIN_POLL_DELAY_MS` in `AwaitApprovalConfirmationUseCase`.

aminsato (approved):
lGTM

johnnyluo (changes requested):
can you please post a test transaction link?

johnnyluo:
https://etherscan.io/tx/0x5a12aae44189c744c648cd34c5e2930b0570ae808651666612e26d460372ae81
https://etherscan.io/tx/0x9a5f77f7fa844a420516964e519393c92b20ded59f59c973fafa20940669bffc

johnnyluo (approved):
LGTM

Issue #4100, opened 2026-04-16
Import vault: wrong-password error not shown in UI


rkokhatskyi:
## Summary

When importing an encrypted vault backup, if the user enters a password and tapping **Continue** fails decryption, the password prompt dialog is closed and no error is displayed. The UI silently returns to the import screen, giving the user no feedback.

## Reproduction

1. Open **Import** flow with an encrypted `.vult` file
2. Tap the file, then **Continue**
3. The password prompt appears
4. Enter any password and tap **Continue**
5. If decryption fails, the dialog closes and nothing is shown

## Evidence 

https://github.com/user-attachments/assets/de7427bc-2be0-4a90-889a-a35a7a5d3caf

## Root cause

`app/src/main/java/com/vultisig/wallet/ui/models/ImportFileViewModel.kt:97-106` — `decryptVaultData()` calls `hidePasswordPromptDialog()` unconditionally before branching on the result:

```kotlin
viewModelScope.launch {
    val result = saveToDb(vaultFileContent, key)
    hidePasswordPromptDialog()        // always hides
    when (result) {
        SaveResult.Success  -> showSuccessImport()
        SaveResult.Duplicate -> showDuplicateError()
        SaveResult.Failed   -> showErrorHint()   // sets passwordErrorHint
    }
}
```

`showErrorHint()` writes `passwordErrorHint` on state, but that field is only rendered inside `KeysignPasswordBottomSheet`, which is gated by `showPasswordPrompt`. Once `hidePasswordPromptDialog()` flips that to `false`, the sheet is gone before the error can render.

## Fix

Keep the dialog open on `SaveResult.Failed`; only hide on `Success` / `Duplicate`.

## Stack trace (from user report)

GCM decryption throws `AEADBadTagException`, falls back to legacy CBC, also fails → `SaveResult.Failed`. No user-visible feedback.

```
AesEncryption  E  switch to old version decryption
AesEncryption  E  javax.crypto.AEADBadTagException: error:1e000065:Cipher functions:OPENSSL_internal:BAD_DECRYPT
  at com.vultisig.wallet.data.usecases.AesEncryption.decrypt(Encryption.kt:45)
  at com.vultisig.wallet.data.usecases.ParseVaultFromStringUseCaseImpl.parseProtoBufVault(ParseVaultFromStringUseCase.kt:52)
  at com.vultisig.wallet.ui.models.ImportFileViewModel.saveToDb(ImportFileViewModel.kt:121)
[...]

PR #4103, opened 2026-04-16
fix: make AesEncryption thread-safe by instantiating Cipher per call

### LINKED ISSUE #4101: AesEncryption: shared Cipher/MessageDigest in @Singleton is not thread-safe
## Summary

`AesEncryption` (`data/src/main/kotlin/com/vultisig/wallet/data/usecases/Encryption.kt`) is bound as `@Singleton` (via `DataUsecasesModule.bindEncryption`) and holds shared mutable `javax.crypto.Cipher` and `java.security.MessageDigest` instances as fields:

```kotlin
internal class AesEncryption @Inject constructor() : Encryption {
    private val cipher = Cipher.getInstance("AES/GCM/NoPadding")
    private val messageDigest = MessageDigest.getInstance("SHA-256")
    ...
}
```

Both classes are documented as **not thread-safe**. Concurrent `init()`/`doFinal()` calls from different coroutines can interleave and corrupt the cipher state, producing `AEADBadTagException` even when the key, IV, and ciphertext are correct. State can also leak across calls on some Android OpenSSL builds if a prior `doFinal()` failed.

## Blast radius

`Encryption` is hit from many hot paths, any of which can run concurrently:

- TSS peer-to-peer messaging per message during keygen/keysign (`TssMessenger`, `TssMessagePuller`, `PullTssMessagesUseCase`)
- Vault backup parsing and creation (`ParseVaultFromStringUseCase`, `CreateVaultBackupUseCase`)
- DKLS/Schnorr/MLDSA keygen and keysign (`DklsKeygen`, `DklsKeysign`, `SchnorrKeygen`, `SchnorrKeysign`, `MldsaKeygen`, `MldsaKeysign`)

## Observed symptom

Vault import with a correct password intermittently fails with:

```
AesEncryption  E  switch to old version decryption
AesEncryption  E  javax.crypto.AEADBadTagException: error:1e000065:Cipher functions:OPENSSL_internal:BAD_DECRYPT
  at com.vultisig.wallet.data.usecases.AesEncryption.decrypt(Encryption.kt:45)
```

This is the most visible case but the underlying hazard affects any flow that uses `Encryption` under concurrency (TSS messaging in particular).

## Fix

Instantiate `Cipher` and `MessageDigest` per call instead of caching them as fields:

```kotlin
override fun decrypt(data: ByteArray, password: ByteArray): ByteArray? = try {
    val keyBytes = MessageDigest.getInstance("SHA-256").digest(password)
    val keySpec = SecretKeySpec(keyBytes, "AES")
    val cipher = Cipher.getInstance("AES/GCM/NoPadding")
    cipher.init(Cipher.DECRYPT_MODE, keySpec, GCMParameterSpec(128, data.copyOfRange(0, 12)))
    cipher.doFinal(data.copyOfRange(12, data.size))
} catch (e: Exception) {
    Timber.d(e, "GCM decrypt failed, trying legacy CBC")
    decryptOldVersion(data, password)
}
```

`SecureRandom` can remain a field — it is thread-safe.

Also drop the `Timber.e("switch to old version decryption")` to DEBUG — the AES/GCM failure is the expected path for legacy CBC backups and should not be logged as an error.

rkokhatskyi:
## Summary

Closes #4101.

`AesEncryption` is bound as `@Singleton` and was caching `javax.crypto.Cipher` and `java.security.MessageDigest` as shared instance fields. Both classes are documented as **not thread-safe**, and this singleton is hit concurrently from many hot paths:

- **TSS keygen/keysign**: outbound `encryption.encrypt(...)` via `TssMessenger.send()` (called synchronously from native TSS callbacks) runs in parallel with inbound `encryption.decrypt(...)` via `TssMessagePuller` (a separate coroutine on `viewModelScope`). Both flows touch the same shared `Cipher` on every message.
- Vault parsing (`ParseVaultFromStringUseCase`) and backup (`CreateVaultBackupUseCase`).
- DKLS, Schnorr, and MLDSA keygen/keysign message pipelines.

Concurrent `init()` / `doFinal()` pairs on the shared `Cipher` can corrupt internal state, producing `AEADBadTagException` even with a valid key/IV/ciphertext, or silently emitting incorrect ciphertext that peers reject during TSS rounds.

## Fix

- Create fresh `Cipher` and `MessageDigest` instances per call. Allocation cost is negligible (microseconds).
- Keep `SecureRandom` as a shared field — it is documented as thread-safe.
- Switch the legacy CBC transformation from `AES/CBC/PKCS7PADDING` to `AES/CBC/PKCS5Padding`. For AES (16-byte block size) these produce byte-identical output; PKCS5Padding is accepted by both Android SunJCE and plain JVM, while PKCS7PADDING is Android-only.
- Drop the `"switch to old version decryption"` log from ERROR to DEBUG — the GCM decrypt throwing on legacy CBC backups is the expected path and should not flood error logs.

## Tests

`AesEncryptionTest` expanded from 5 to 10 cases. All run on JUnit Platform via `kotlin.test` (the existing `@org.junit.Test` annotations were silently no-op since JUnit Vintage is not on the classpath — fixed here by migrating to `kotlin.test.Test`).

New coverage:

- `encrypt produces different ciphertext for same input due to random IV`
- `empty input round trips`
- `large payload round trips` (128 KB, representative of TSS message size)
- `decrypt returns null when both GCM and CBC cannot decrypt`
- `concurrent encrypt-decrypt with same password does not corrupt results` (16 coroutines × 200 iterations)
- `concurrent encrypt-decrypt with different passwords does not cross contaminate` (16 coroutines × 200 iterations, distinct passwords and payloads)
- `concurrent operations on real thread pool do not fail` (raw JDK threading, 8-thread pool × 32 tasks × 100 ops)

Verified the regression test catches the bug: reverting to the shared-Cipher implementation produces **41 failures out of 3200 operations** on the thread-pool test.

## Test plan

- [ ] Unit tests pass (`./gradlew :data:testDebugUnitTest`)
- [ ] Full vault import with correct password — decrypts and saves
- [ ] Full vault import with wrong password — shows the password error (#4100)
- [ ] Full keygen flow with 2 devices — completes without TSS message decryption errors
- [ ] Full keysign flow on at least one chain — transaction signs and broadcasts
- [ ] Vault backup + import round-trip — decrypts cleanly

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved encryption thread-safety by eliminating shared crypto primitives that could cause issues in concurrent scenarios.
  * Enhanced decryption error handling with improved fallback mechanisms.

* **Tests**
  * Added comprehensive encryption randomness and round-trip validation tests.
  * Introduced multi-threaded concurrency stress tests to verify reliability under parallel operations.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/Encryption.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/Encryption.kt
@@ -18,59 +18,57 @@ interface Encryption {
 
 internal class AesEncryption @Inject constructor() : Encryption {
 
-    private val cipher = Cipher.getInstance("AES/GCM/NoPadding")
-    private val messageDigest = MessageDigest.getInstance("SHA-256")
-    private val algorithm = "AES"
+    // SecureRandom is documented as thread-safe; keep as a shared field.
+    // Cipher and MessageDigest are not thread-safe and must be created per call —
+    // the class is a @Singleton and is hit concurrently from TSS messaging,
+    // vault parsing, and keygen/keysign flows.
     private val random = SecureRandom()
 
     override fun encrypt(data: ByteArray, password: ByteArray): ByteArray {
-        val keyBytes = messageDigest.digest(password)
-        val keySpec = SecretKeySpec(keyBytes, algorithm)
-        val iv = ByteArray(12)
-        random.nextBytes(iv)
-        val parameterSpec = GCMParameterSpec(128, iv)
-        cipher.init(Cipher.ENCRYPT_MODE, keySpec, parameterSpec)
-        val encryptedData = cipher.doFinal(data)
-        val combined = iv + encryptedData
-        return combined
+        val keySpec = SecretKeySpec(sha256(password), ALGORITHM)
+        val iv = ByteArray(GCM_IV_LENGTH).also(random::nextBytes)
+        val cipher = Cipher.getInstance(GCM_TRANSFORMATION)
+        cipher.init(Cipher.ENCRYPT_MODE, keySpec, GCMParameterSpec(GCM_TAG_LENGTH_BITS, iv))
+        return iv + cipher.doFinal(data)
     }
 
     override fun decrypt(data: ByteArray, password: ByteArray): ByteArray? {
-        try {
-            val keyBytes = messageDigest.digest(password)
-            val keySpec = SecretKeySpec(keyBytes, algorithm)
-            val iv = data.copyOfRange(0, 12)
-            val encryptedData = data.copyOfRange(12, data.size)
-            cipher.init(Cipher.DECRYPT_MODE, keySpec, GCMParameterSpec(128, iv))
-            return cipher.doFinal(encryptedData)
+        return try {
+            val keySpec = SecretKeySpec(sha256(password), ALGORITHM)
+            val iv = data.copyOfRange(0, GCM_IV_LENGTH)
+            val encryptedData = data.copyOfRange(GCM_IV_LENGTH, data.size)
+            val cipher = Cipher.getInstance(GCM_TRANSFORMATION)
+            cipher.init(Cipher.DECRYPT_MODE, keySpec, GCMParameterSpec(GCM_TAG_LENGTH_BITS, iv))
+            cipher.doFinal(encryptedData)
         } catch (e: Exception) {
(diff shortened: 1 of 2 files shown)


rkokhatskyi:
@coderabbitai full review

johnnyluo (approved):
LGTM

PR #4128, opened 2026-04-19
fix: recover from 'tx already known' broadcast errors on 7 non-EVM chains

### LINKED ISSUE #4105: Android: Keysign errors on 'tx already known' for Cardano/Solana/Tron/Sui/Ripple/Polkadot/Ton
## Summary

In MPC keysign all participating devices broadcast the signed transaction independently. When a peer wins the RPC race, the slower device gets an "already known / duplicate / already in mempool" error from the RPC. On Android this causes the signing flow to fail (`KeysignState.Error()`) even though the tx is on-chain and funds moved correctly.

Android handles this correctly for EVM, Cosmos, and Bittensor but errors out on Cardano, Solana, Tron, Sui, Ripple, Polkadot, and Ton.

## Current state

`data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt:74-153` delegates to per-chain APIs. Per-chain status:

**Handled correctly (keep as-is, audit for safety):**
- `EvmApi.kt:335-370` — catches `"already known"`, `"transaction already exists"`, `"tx already in mempool"`, `"nonce too low"`; recomputes Keccak256 and returns hash
- `CosmosApi.kt:103-120` — accepts code `0` or `19` (`"tx already in mempool"`)
- `BittensorApi.kt:117-137` — catches `"Already Imported"`

**Missing handling (need fix):**
- `CardanoApi.kt:86-134` — throws on all error paths
- `SolanaApi.kt:178-199` — throws on error
- `PolkadotApi.kt:145-164` — catches RPC codes `1012`/`1013` (invalid in tx pool) but no "already broadcast" path
- `TronApi.kt`
- `SuiApi.kt`
- `RippleApi.kt`
- `TonApi.kt`

Broadcast errors bubble to `KeysignViewModel` (around line 375-389) and set `KeysignState.Error()`, failing the flow.

## Proposed fix — hash-verification on error (shared pattern)

Since every signed tx has a deterministic hash, replace fragile per-chain string matching with on-chain hash verification in the catch branch of each chain's broadcast method:

```
on broadcast error:
  if getTxByHash(hash) exists (mempool or block):
    return hash  // success
  throw error
```

Each chain API already has a tx-status / pending-tx lookup somewhere — consolidate behind a `getTxByHash` helper and call from the catch in each broadcast method.

Apply across **all** chain APIs, not just the 7 missing ones. Keep existing string-matches as the fast path and hash-verify as the safety net. Follow-up PR can delete string-matches once hash-verify is proven.

## Audit existing short-circuits

Review the EVM/Cosmos/Bittensor cases already shipped — any match on "timeout", "sequence mismatch", "nonce" without verifying the hash can false-positive for unrelated txs from the same account. Hash-verify removes that risk.

## Non-goals

- No cross-device coordination / initiator-only broadcast. Every device still broadcasts independently.
- No changes to `KeysignViewModel` state machine beyond reading the successful result.

## Files

- `data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt:74-153`
- `data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt:86-134`
- `data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt:178-199`
- `data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt:145-164`
- `data/src/main/kotlin/com/vultisig/wallet/data/api/TronApi.kt`
- `data/src/main/kotlin/com/vultisig/wallet/data/api/SuiApi.kt`
- `data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt`
- `data/src/main/kotlin/com/vultisig/wallet/data/api/TonApi.kt`

## Related

Matching issues filed on `vultisig-ios` and `vultisig-sdk` (which powers Windows/extension).

aminsato:
Closes #4105

## Summary
- In MPC keysign every participating device broadcasts the signed tx independently. When a peer wins the RPC race, the slower device gets an "already known"/"duplicate"/"already in mempool" error — even though the tx is on-chain and funds moved.
- EVM, Cosmos, and Bittensor already handle this. Cardano, Solana, Tron, Sui, Ripple, Polkadot, and Ton did not, so keysign surfaced `KeysignState.Error()` despite success.
- Added `recoverIfAlreadyBroadcast` in `BroadcastTxUseCase`: on any broadcast exception it verifies `tx.transactionHash` via each chain's existing tx-status lookup and returns the hash if the tx is already on-chain. Otherwise the original exception is rethrown. `CancellationException` is rethrown unchanged.
- This replaces fragile per-chain string matching with deterministic on-chain hash verification and removes the false-positive risk noted in the issue (lookalike error strings from unrelated txs on the same account).

## Per-chain lookups used
| Chain | Lookup | "Found" signal |
|-------|--------|----------------|
| Cardano  | `CardanoApi.getTxStatus` | `txHash` non-blank |
| Solana   | `SolanaApi.checkStatus` | any non-null entry in `result.value` |
| Polkadot | `PolkadotApi.getTxStatus` | `data.extrinsicHash` non-blank |
| Tron     | `TronApi.getTsStatus` | `txId` non-blank |
| Sui      | `SuiApi.checkStatus` | `digest` non-blank |
| Ripple   | `RippleApi.getTsStatus` | `result.hash` non-blank |
| Ton      | `TonApi.getTsStatus` | `transactions` non-empty |

EVM/Cosmos/MayaChain/Thorchain/Bittensor paths are unchanged — the audit/consolidation of their existing string-matches is left as a follow-up per the issue.

## Test plan
- [ ] `./gradlew :data:compileDebugKotlin` (verified locally)
- [ ] `./gradlew :app:compileDebugKotlin` (verified locally)
- [ ] `./gradlew :data:ktfmtFormat` clean (verified locally)
- [ ] Keygen/keysign on Solana with 2 devices — slower device no longer surfaces `KeysignState.Error()` after the first peer wins
- [ ] Same check on Cardano, Tron, Sui, Ripple, Polkadot, Ton
- [ ] Regression: single-device broadcast still surfaces a real error when the tx never reaches the chain (hash lookup returns not-found → original exception rethrown)

Co-Authored-By: aminsato <Amin.saradar@yahoo.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction broadcasting reliability for Solana, Polkadot, Sui, Ton, Ripple, Tron, and Cardano. The app now better handles cases where transactions were already broadcast to the blockchain, preventing unnecessary errors and allowing successful completion when transactions are confirmed on-chain.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
@@ -48,6 +48,8 @@ import com.vultisig.wallet.data.models.Chain.Ton
 import com.vultisig.wallet.data.models.Chain.ZkSync
 import com.vultisig.wallet.data.models.SignedTransactionResult
 import javax.inject.Inject
+import kotlin.coroutines.cancellation.CancellationException
+import timber.log.Timber
 
 fun interface BroadcastTxUseCase {
     suspend operator fun invoke(chain: Chain, tx: SignedTransactionResult): String?
@@ -103,9 +105,14 @@ constructor(
                 evmApi.sendTransaction(tx.rawTransaction)
             }
 
-            Solana -> {
-                solanaApi.broadcastTransaction(tx.rawTransaction)
-            }
+            Solana ->
+                recoverIfAlreadyBroadcast(
+                    tx = tx,
+                    broadcast = { solanaApi.broadcastTransaction(tx.rawTransaction) },
+                    verify = { hash ->
+                        solanaApi.checkStatus(hash)?.result?.value?.any { it != null } == true
+                    },
+                )
 
             GaiaChain,
             Kujira,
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #4133, opened 2026-04-20
fix: redesign Share QR Code share-bitmap to match Figma (#4130)

### LINKED ISSUE #4130: [UI] Share QR Code — update Join Keysign and Join Keygen layouts to match Figma
## What

Update the **Share QR Code** screens for Join Keysign and Join Keygen to match the current Figma. Covers QR card, metadata panel (Vault / Amount / To for keysign; Vault / Type for keygen), Vultisig brand signature.

## Priority

Lower priority — polish ticket.

## Source of truth

Figma: https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=65380-108157

## Scope

- **Share Keysign QR screen**
- **Share Keygen QR screen**
- Match typography, spacing, card padding, and background to Figma exactly

## Acceptance criteria

- [ ] Both screens match Figma 1:1
- [ ] Metadata fields render correctly (amount formatting, address truncation, type format)
- [ ] Vultisig signature (icon + wordmark) placed as per Figma
- [ ] Visual regression pass vs Figma screenshots

## Related

Cross-platform ticket — iOS and Windows have matching issues opened.

aminsato:
## Summary

Closes #4130.

The exported **Join Keysign** and **Join Keygen** share images now match the current Figma:

- Rounded **QR card** with 1 dp border (`#11284A`) and 24 dp corner radius
- Separate **metadata card** (surface-1 fill `#061B3A`) containing:
  - Title (`Join Keysign` / `Join Keysign Swap` / `Join Keygen`) in Brockmann Medium 14 sp
  - Label/value rows separated by 1 dp dividers
    - **Keysign**: Vault / Amount / To
    - **Keysign swap**: Vault / From / To (both From and To rows render the source/destination token icons)
    - **Keygen**: Vault / Type (Fast Vault or Secure Vault)
- **Vultisig signature** at the bottom: rounded-square logo (`R.drawable.logo`) + Vultisig wordmark
- Source token icon (e.g. USDC `$` mark) rendered before the value in Amount/From/To rows; falls back to the chain logo when the token's logo string isn't in `getCoinLogo`

## Implementation notes

- `MakeQrCodeBitmapShareFormat` interface changed from `(title: String, description: String?)` to structured `QrShareInfo(title, fields: List<QrShareField>)` so call sites can pass real key/value pairs and per-row icons instead of newline-delimited strings.
- New labels (`qr_share_label_vault/amount/to/from/type`) and type strings (`qr_share_type_fast_vault/secure_vault`) added to all 10 locales using each locale's existing terminology (e.g. `Хранилище` for Vault in Russian, `保险库` in zh-rCN).
- Brockmann Medium loaded via `Resources.getIdentifier` (suppressed `DiscouragedApi` lint, justified by cross-module access — the font lives in the app module's resources, the renderer in `data`).
- `BitmapFactory.decodeResource` calls in `KeysignPeerDiscovery` are wrapped in `remember(...)` so they don't re-decode on every recomposition.
- `LaunchedEffect(saveShareQrBitmap)` now keys on `qrShareInfo` as well as `qrBitmapPainter`, so the share image regenerates when amount / address / token icon changes — not only when the QR painter changes.
- Added `share_qr_keysign` and `share_qr_keygen` previews in `PreviewActivity` (debug only) for visual iteration via ADB.

## Test plan

- [ ] Send → Keysign peer discovery → tap Share QR → exported image matches Figma (QR card, metadata card, signature, USDC token icon visible in Amount row)
- [ ] Swap → Keysign peer discovery → exported image shows Vault / From / To with both token icons
- [ ] Sign Message → Keysign peer discovery → exported image still works (no Amount row regression)
- [ ] Keygen (Fast Vault flow) → peer discovery → Share QR → metadata card shows Vault / Type = Fast Vault
- [ ] Keygen (Active / Secure Vault flow) → metadata card shows Vault / Type = Secure Vault
- [ ] Switch device language to each of de/es/hr/it/ko/nl/pt/ru/zh-rCN — labels translate correctly
- [ ] Existing `KeygenPeerDiscoveryViewModelTest` unit test still passes

## Summary by CodeRabbit

* **New Features**
  * Enhanced QR share card with structured fields, improved layout and app-branded logo
  * Token icons shown inline; richer metadata for swap and keysign sharing
  * New debug preview screens to view/share QR visuals

* **Refactor**
  * QR sharing APIs consolidated to accept a structured share-info object

* **Localization**
  * Added QR-sharing translations for multiple languages

* **Tests**
  * Added unit tests for QR share models and token-logo resolution

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/models/CoinLogo.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/models/CoinLogo.kt
@@ -143,3 +143,8 @@ internal fun getCoinLogo(logoName: String): ImageModel {
         else -> logoName
     }
 }
+
+// Returns a drawable resource id for the coin's logo. Falls back to the chain's native logo when
+// the coin's `logo` string is not in the predefined mapping (e.g. for arbitrary ERC20s where
+// `logo` would otherwise be a URL).
+internal fun Coin.tokenLogoRes(): Int = (getCoinLogo(logo) as? Int) ?: chain.logo
--- a/app/src/main/java/com/vultisig/wallet/data/usecases/CreateQrCodeSharingBitmapUseCase.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/usecases/CreateQrCodeSharingBitmapUseCase.kt
@@ -3,7 +3,6 @@ package com.vultisig.wallet.data.usecases
 import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.BitmapFactory
-import androidx.annotation.StringRes
 import androidx.compose.ui.graphics.toArgb
 import com.vultisig.wallet.R
 import com.vultisig.wallet.ui.theme.v2.V2.colors
@@ -11,28 +10,24 @@ import dagger.hilt.android.qualifiers.ApplicationContext
 import javax.inject.Inject
 
 fun interface CreateQrCodeSharingBitmapUseCase {
-    operator fun invoke(qr: Bitmap, @StringRes title: Int, @StringRes description: Int?): Bitmap
+    operator fun invoke(qr: Bitmap, info: QrShareInfo): Bitmap
 }
 
 internal class CreateQrCodeSharingBitmapUseCaseImpl
 @Inject
 constructor(
-    @ApplicationContext private val context: Context,
+    @param:ApplicationContext private val context: Context,
     private val makeQrCodeBitmapShareFormat: MakeQrCodeBitmapShareFormat,
 ) : CreateQrCodeSharingBitmapUseCase {
-    override fun invoke(qr: Bitmap, title: Int, description: Int?): Bitmap {
-        val titleString = context.getString(title)
-        val descriptionString = description?.let { context.getString(it) }
-
-        val logo = BitmapFactory.decodeResource(context.resources, R.drawable.ic_share_qr_logo)
+    override fun invoke(qr: Bitmap, info: QrShareInfo): Bitmap {
+        val logo = BitmapFactory.decodeResource(context.resources, R.drawable.logo)
 
         return makeQrCodeBitmapShareFormat(
             context,
             qr,
             colors.backgrounds.primary.toArgb(),
             logo,
-            titleString,
-            descriptionString,
+            info,
         )
     }
 }
--- a/app/src/main/java/com/vultisig/wallet/data/usecases/CreateQrCodeSharingBitmapUseCase.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/usecases/CreateQrCodeSharingBitmapUseCase.kt
@@ -19,15 +19,12 @@ constructor(
     @param:ApplicationContext private val context: Context,
     private val makeQrCodeBitmapShareFormat: MakeQrCodeBitmapShareFormat,
 ) : CreateQrCodeSharingBitmapUseCase {
-    override fun invoke(qr: Bitmap, info: QrShareInfo): Bitmap {
-        val logo = BitmapFactory.decodeResource(context.resources, R.drawable.logo)
 
-        return makeQrCodeBitmapShareFormat(
-            context,
-            qr,
-            colors.backgrounds.primary.toArgb(),
-            logo,
-            info,
-        )
+    // R.drawable.logo never changes; decode once and reuse across shares.
+    private val logo: Bitmap by lazy {
+        BitmapFactory.decodeResource(context.resources, R.drawable.logo)
     }
+
+    override fun invoke(qr: Bitmap, info: QrShareInfo): Bitmap =
+        makeQrCodeBitmapShareFormat(context, qr, colors.backgrounds.primary.toArgb(), logo, info)
 }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
@@ -35,6 +35,7 @@ import javax.inject.Inject
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import vultisig.keysign.v1.CustomMessagePayload
@@ -224,10 +225,20 @@ constructor(
         saveShareQrBitmapJob =
             viewModelScope.launch {
                 val bitmap = qrBitmap ?: return@launch
+                // Allocation happens on IO; if cancellation races with the render, `withContext`'s
+                // prompt-cancellation guarantee discards the returned bitmap before we can recycle
+                // it. Recycle in-place and return null so the native memory is released instead of
+                // waiting for GC.
                 val rendered =
                     withContext(Dispatchers.IO) {
-                        makeQrCodeBitmapShareFormat(context, bitmap, color, logo, info)
-                    }
+                        val bmp = makeQrCodeBitmapShareFormat(context, bitmap, color, logo, info)
+                        if (!isActive) {
+                            bmp.recycle()
+                            null
+                        } else {
+                            bmp
+                        }
+                    } ?: return@launch
                 val previous = shareQrBitmap.value
                 shareQrBitmap.value = rendered
                 if (previous != null && previous !== rendered) previous.recycle()
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
@@ -33,6 +33,7 @@ import com.vultisig.wallet.ui.utils.shareFileName
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -64,6 +65,7 @@ constructor(
     val qrBitmapPainter = MutableStateFlow<BitmapPainter?>(null)
     private var qrBitmap: Bitmap? = null
     private val shareQrBitmap = MutableStateFlow<Bitmap?>(null)
+    private var saveShareQrBitmapJob: Job? = null
 
     suspend fun loadTransaction(transactionId: TransactionId) {
         val transaction = transactionRepository.getTransaction(transactionId)
@@ -214,13 +216,21 @@ constructor(
         activity.share(qrBitmap, shareFileName(requireNotNull(vault), ShareType.SEND))
     }
 
-    internal fun saveShareQrBitmap(context: Context, color: Int, info: QrShareInfo, logo: Bitmap) =
-        viewModelScope.launch {
-            val bitmap = qrBitmap ?: return@launch
-            val qrBitmap =
-                withContext(Dispatchers.IO) {
-                    makeQrCodeBitmapShareFormat(context, bitmap, color, logo, info)
-                }
-            shareQrBitmap.value = qrBitmap
-        }
+    internal fun saveShareQrBitmap(context: Context, color: Int, info: QrShareInfo, logo: Bitmap) {
+        // Cancel any in-flight render so rapid `qrShareInfo` updates (painter then icons) can't
+        // race and let a stale bitmap overwrite a fresher one. The previous rendered bitmap is
+        // recycled on replacement to avoid leaks when icons/amounts change repeatedly.
+        saveShareQrBitmapJob?.cancel()
+        saveShareQrBitmapJob =
+            viewModelScope.launch {
+                val bitmap = qrBitmap ?: return@launch
+                val rendered =
+                    withContext(Dispatchers.IO) {
+                        makeQrCodeBitmapShareFormat(context, bitmap, color, logo, info)
+                    }
+                val previous = shareQrBitmap.value
+                shareQrBitmap.value = rendered
+                if (previous != null && previous !== rendered) previous.recycle()
+            }
+    }
 }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
@@ -25,6 +25,7 @@ import com.vultisig.wallet.data.repositories.TransactionRepository
 import com.vultisig.wallet.data.repositories.VaultRepository
 import com.vultisig.wallet.data.usecases.GenerateQrBitmap
 import com.vultisig.wallet.data.usecases.MakeQrCodeBitmapShareFormat
+import com.vultisig.wallet.data.usecases.QrShareInfo
 import com.vultisig.wallet.ui.models.mappers.TokenValueToStringWithUnitMapper
 import com.vultisig.wallet.ui.utils.ShareType
 import com.vultisig.wallet.ui.utils.share
@@ -213,18 +214,12 @@ constructor(
         activity.share(qrBitmap, shareFileName(requireNotNull(vault), ShareType.SEND))
     }
 
-    internal fun saveShareQrBitmap(
-        context: Context,
-        color: Int,
-        title: String,
-        description: String,
-        logo: Bitmap,
-    ) =
+    internal fun saveShareQrBitmap(context: Context, color: Int, info: QrShareInfo, logo: Bitmap) =
         viewModelScope.launch {
             val bitmap = qrBitmap ?: return@launch
             val qrBitmap =
                 withContext(Dispatchers.IO) {
-                    makeQrCodeBitmapShareFormat(context, bitmap, color, logo, title, description)
+                    makeQrCodeBitmapShareFormat(context, bitmap, color, logo, info)
                 }
             shareQrBitmap.value = qrBitmap
         }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -207,8 +207,15 @@ constructor(
     fun shareQr(activity: Context) {
         val qr = qrBitmap.value ?: return
 
+        // Mirror loadData()'s fast/secure predicate: active-vault migrate (signers > 2) goes
+        // through peer discovery even when email+password are present, so the share card must
+        // not advertise "Fast Vault" for it.
+        val isFastVault =
+            !email.isNullOrBlank() &&
+                !password.isNullOrBlank() &&
+                !(args?.action == TssAction.Migrate && signers.size > 2)
         val typeRes =
-            if (email != null) R.string.qr_share_type_fast_vault
+            if (isFastVault) R.string.qr_share_type_fast_vault
             else R.string.qr_share_type_secure_vault
         val info =
             QrShareInfo(
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -13,7 +13,6 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.FilterQuality
 import androidx.compose.ui.graphics.asImageBitmap
 import androidx.compose.ui.graphics.painter.BitmapPainter
-import androidx.core.graphics.scale
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
@@ -208,15 +207,6 @@ constructor(
     fun shareQr(activity: Context) {
         val qr = qrBitmap.value ?: return
 
-        val scaleModifier = 4
-
-        val scaledQr =
-            qr.scale(
-                width = qr.width * scaleModifier,
-                height = qr.height * scaleModifier,
-                filter = false,
-            )
-
         val typeRes =
             if (email != null) R.string.qr_share_type_fast_vault
             else R.string.qr_share_type_secure_vault
@@ -236,7 +226,7 @@ constructor(
                     ),
             )
 
-        val shareBitmap = createQrCodeSharingBitmap(scaledQr, info)
+        val shareBitmap = createQrCodeSharingBitmap(qr, info)
 
         activity.share(shareBitmap, shareFileName(vaultName, vaultName.sha256(), ShareType.KEYGEN))
     }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -48,6 +48,8 @@ import com.vultisig.wallet.data.usecases.ExtractMasterKeysUseCase
 import com.vultisig.wallet.data.usecases.GenerateQrBitmap
 import com.vultisig.wallet.data.usecases.GenerateServerPartyId
 import com.vultisig.wallet.data.usecases.GenerateServiceName
+import com.vultisig.wallet.data.usecases.QrShareField
+import com.vultisig.wallet.data.usecases.QrShareInfo
 import com.vultisig.wallet.data.usecases.tss.DiscoverParticipantsUseCase
 import com.vultisig.wallet.data.usecases.tss.ParticipantName
 import com.vultisig.wallet.data.utils.safeLaunch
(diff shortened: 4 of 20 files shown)


data/src/test/kotlin/com/vultisig/wallet/data/usecases/QrShareInfoTest.kt
```diff
@@ -0,0 +1,115 @@
...
+    fun `QrShareField differing only by value is not equal`() {
+        val a = QrShareField("Amount", "100 USDC")
+        val b = QrShareField("Amount", "200 USDC")
+
+        assertNotEquals(a, b)
+    }
+
+    @Test
+    fun `QrShareField differing only by valueIcon reference is not equal`() {
+        val icon1 = mockk<Bitmap>()
```
rkokhatskyi: `data/src/test` is a JVM-only unit test source set, there is no Robolectric or mockable-android-jar configured, so `mockk<Bitmap>()` will blow up at runtime with `NoClassDefFoundError` on `android.graphics.Bitmap`, either move this file to `androidTest` or replace the bitmap with a simple sentinel (`object : Any()`) since the test only cares about reference inequality

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -214,13 +217,27 @@ constructor(
                 filter = false,
             )
```
rkokhatskyi: the `scaleModifier = 4` here allocates a 4x scaled QR bitmap only for `MakeQrCodeBitmapShareFormat` to immediately scale it back down to `QR_CODE_MAX_SIZE_DP`, the keysign path in `KeysignShareViewModel.saveShareQrBitmap` skips the 4x entirely, drop the pre-scale and let the renderer handle sizing

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
```diff
@@ -213,18 +214,12 @@ constructor(
...
-    ) =
+    internal fun saveShareQrBitmap(context: Context, color: Int, info: QrShareInfo, logo: Bitmap) =
         viewModelScope.launch {
             val bitmap = qrBitmap ?: return@launch
             val qrBitmap =
                 withContext(Dispatchers.IO) {
-                    makeQrCodeBitmapShareFormat(context, bitmap, color, logo, title, description)
+                    makeQrCodeBitmapShareFormat(context, bitmap, color, logo, info)
                 }
             shareQrBitmap.value = qrBitmap
```
rkokhatskyi: `saveShareQrBitmap` launches into `viewModelScope` while the caller's `LaunchedEffect(qrBitmapPainter, qrShareInfo)` uses its own effect scope, if two invocations overlap (painter arrives then icons arrive in rapid state updates) the later-starting coroutine can finish first and the stale bitmap overwrites it, either track a `Job` and cancel the previous one or launch inside the `LaunchedEffect` scope

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
```diff
@@ -30,90 +81,196 @@ internal class MakeQrCodeBitmapShareFormatImpl @Inject constructor() : MakeQrCod
...
+        var rowsCursorY = metaCardTop + metaPadding + titleLineHeight + titleToFieldsGap
+        info.fields.forEachIndexed { index, field ->
+            val textBaselineY = rowsCursorY + fieldLineHeight * 0.75f
+            val iconPx =
+                if (field.valueIcon != null) valueIconSize.roundToInt().coerceAtLeast(1) else 0
+            val iconReservation = if (iconPx > 0) iconPx + valueIconToTextGap else 0f
+            val ellipsizedLabel = ellipsizeToWidth(field.label, labelPaint, metaContentWidth * 0.5f)
+            val labelMeasured = labelPaint.measureText(ellipsizedLabel)
+            val valueMaxWidth = metaContentWidth - labelMeasured - iconReservation
+            val ellipsizedValue = ellipsizeToWidth(field.value, valuePaint, valueMaxWidth)
```
rkokhatskyi, line 249: `valueMaxWidth = metaContentWidth - labelMeasured - iconReservation` can go negative on narrow cards with wide localized labels plus an icon, `ellipsizeToWidth` then returns `""` and the value silently disappears from the share image with no warning, either `coerceAtLeast(0f)` or reserve a minimum value width and truncate the label more aggressively

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
```diff
@@ -30,90 +81,196 @@ internal class MakeQrCodeBitmapShareFormatImpl @Inject constructor() : MakeQrCod
...
-            finalHeight - padding - textSize / 2,
-            paint,
-        )
+        val metaCardTop = qrCardBottom + betweenCardsGap
+        val metaCardBottom = metaCardTop + metaCardHeight
+        val metaCardRect = RectF(cardLeft, metaCardTop, cardRight, metaCardBottom)
+        canvas.drawRoundRect(metaCardRect, cardCorner, cardCorner, cardFillPaint)
+        canvas.drawRoundRect(metaCardRect, cardCorner, cardCorner, borderPaint)
+
+        val titleTypeface = loadBrockmannMedium(context)
```
rkokhatskyi: `loadBrockmannMedium` runs `resources.getIdentifier(...)` + `ResourcesCompat.getFont(...)` on every `invoke`, cache the `Typeface` in the singleton class or inject it once at construction, `getIdentifier` is the reason the `@SuppressLint("DiscouragedApi")` is here

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
```diff
@@ -1,27 +1,78 @@
...
+}
+
+// Resolve the app's Brockmann Medium font from this shared data-module use case. The font lives
+// in the app module's resources, so it is not reachable via R.font here — name lookup is the only
+// option available across modules.
+@SuppressLint("DiscouragedApi")
+private fun loadBrockmannMedium(context: Context): Typeface {
+    val id = context.resources.getIdentifier("brockmann_medium", "font", context.packageName)
+    return id.takeIf { it != 0 }?.let { ResourcesCompat.getFont(context, it) }
+        ?: Typeface.create(Typeface.DEFAULT, Typeface.BOLD)
```
rkokhatskyi: fallback is `Typeface.BOLD` (weight 700) but Brockmann Medium is weight 500, if the lookup ever fails in a non-app-resource context the share bitmap renders noticeably heavier than Figma, use `Typeface.create(Typeface.DEFAULT, 500, false)` on API 28+

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
```diff
@@ -62,21 +65,38 @@ internal fun KeysignPeerDiscovery(
...
-                uiModel.toAddress.forCanvasMinify(),
-            )
+    val labelVault = stringResource(R.string.qr_share_label_vault)
+    val labelAmount = stringResource(R.string.qr_share_label_amount)
+    val labelTo = stringResource(R.string.qr_share_label_to)
+    val labelFrom = stringResource(R.string.qr_share_label_from)
+    val srcIconBitmap =
+        remember(uiModel.srcTokenLogoRes) {
+            uiModel.srcTokenLogoRes?.let { BitmapFactory.decodeResource(context.resources, it) }
+        }
```
rkokhatskyi, line 75: when `srcTokenLogoRes` changes `remember` drops the old bitmap reference without calling `recycle()`, same for `dstIconBitmap`, for drawable resources the framework bitmap cache blunts the impact but explicit recycling via `DisposableEffect` keyed on the res id is the robust pattern

app/src/main/java/com/vultisig/wallet/data/usecases/CreateQrCodeSharingBitmapUseCase.kt
```diff
@@ -3,36 +3,31 @@ package com.vultisig.wallet.data.usecases
...
+    @param:ApplicationContext private val context: Context,
     private val makeQrCodeBitmapShareFormat: MakeQrCodeBitmapShareFormat,
 ) : CreateQrCodeSharingBitmapUseCase {
-    override fun invoke(qr: Bitmap, title: Int, description: Int?): Bitmap {
-        val titleString = context.getString(title)
-        val descriptionString = description?.let { context.getString(it) }
-
-        val logo = BitmapFactory.decodeResource(context.resources, R.drawable.ic_share_qr_logo)
+    override fun invoke(qr: Bitmap, info: QrShareInfo): Bitmap {
+        val logo = BitmapFactory.decodeResource(context.resources, R.drawable.logo)
```
rkokhatskyi: `BitmapFactory.decodeResource(..., R.drawable.logo)` runs on every share tap, cache the decoded logo in a field on the use case since it never changes, also nobody calls `.recycle()` on it after `makeQrCodeBitmapShareFormat` returns

app/src/main/java/com/vultisig/wallet/data/models/CoinLogo.kt
```diff
@@ -143,3 +143,8 @@ internal fun getCoinLogo(logoName: String): ImageModel {
         else -> logoName
     }
 }
+
+// Returns a drawable resource id for the coin's logo. Falls back to the chain's native logo when
+// the coin's `logo` string is not in the predefined mapping (e.g. for arbitrary ERC20s where
+// `logo` would otherwise be a URL).
+internal fun Coin.tokenLogoRes(): Int = (getCoinLogo(logo) as? Int) ?: chain.logo
```
rkokhatskyi, line 150: `(getCoinLogo(logo) as? Int) ?: chain.logo` only works because `getCoinLogo` returns `Any` (typealias `ImageModel`), the `as? Int` trick silently masks any future return type change, a sealed `ImageModel { DrawableRes(id), Url(uri) }` makes the contract explicit and the `when` in `tokenLogoRes` exhaustive

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
```diff
@@ -30,90 +81,196 @@ internal class MakeQrCodeBitmapShareFormatImpl @Inject constructor() : MakeQrCod
         qrCodeBitmap: Bitmap,
         color: Int,
         logo: Bitmap,
-        title: String,
-        description: String?,
+        info: QrShareInfo,
     ): Bitmap {
```
rkokhatskyi, line 99: the whole bitmap layout (row heights, divider placement, icon bounds, ellipsize) has no unit test, the spec flagged calculation-heavy code needs coverage, a test with a fake `Canvas` / `Bitmap.Config.ALPHA_8` stub asserting the final width/height and row cursor positions for 1/2/3 fields would pin the layout math

app/src/main/res/values-de/strings.xml
```diff
@@ -294,6 +294,13 @@
     <string name="qr_title_join_keygen_description">Mit Geräten scannen, um der \nVault-Generation beizutreten</string>
     <string name="qr_title_join_keysign_description">Tresor: %1$s\nVon: %2$s\n Bis: %3$s</string>
     <string name="qr_title_join_keysign_swap_description">Tresor: %1$s\n Von: %2$s\n An: %3$s</string>
+    <string name="qr_share_label_vault">Tresor</string>
+    <string name="qr_share_label_amount">Menge</string>
+    <string name="qr_share_label_to">An</string>
```
rkokhatskyi: existing `send_to_address` uses `Zu` but this key uses `An`, pick one so the keysign share image and the send screen don't say different things for the same concept

app/src/main/res/values-ko/strings.xml
```diff
@@ -383,6 +383,13 @@
     <string name="qr_title_join_keygen_description">기기로 스캔하여\n볼트 생성에 참여하세요</string>
     <string name="qr_title_join_keysign_description">볼트: %1$s\n금액: %2$s\n 수신: %3$s</string>
     <string name="qr_title_join_keysign_swap_description">볼트: %1$s\n보내기: %2$s\n 받기: %3$s</string>
+    <string name="qr_share_label_vault">볼트</string>
+    <string name="qr_share_label_amount">금액</string>
+    <string name="qr_share_label_to">수신</string>
+    <string name="qr_share_label_from">보내기</string>
```
rkokhatskyi: `보내기` is a verb ("send") and `수신` is a noun ("receipt"), existing `send_from_address` / `send_to_address` use `보내는 주소` / `받는 주소`, align these with `보내는` / `받는` so the labels stay grammatically consistent

app/src/test/java/com/vultisig/wallet/data/models/CoinLogoTest.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    fun `tokenLogoRes always returns a non-zero drawable id`() {
+        // Sanity check: every supported chain must produce a usable resource id, even when the
+        // coin's logo string is junk.
+        Chain.entries.forEach { chain ->
+            val res = coin(logo = "definitely-not-a-real-logo", chain = chain).tokenLogoRes()
+            assertTrue(res != 0, "Expected non-zero drawable id for chain $chain, got 0")
+        }
+    }
```
rkokhatskyi, line 61: only `usdc` and `eth` exercise actual mapping hits out of ~60 entries in `getCoinLogo`, the multi-case branches (`sol`/`solana`, `pol`/`matic`, `zksync`/`zsync-era`, `wif`/`dogwifhat-wif-logo`, etc) are untested so a typo in any alias would silently fall through to `chain.logo`

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
```diff
@@ -30,90 +81,196 @@ internal class MakeQrCodeBitmapShareFormatImpl @Inject constructor() : MakeQrCod
...
+            val iconReservation = if (iconPx > 0) iconPx + valueIconToTextGap else 0f
+            val ellipsizedLabel = ellipsizeToWidth(field.label, labelPaint, metaContentWidth * 0.5f)
+            val labelMeasured = labelPaint.measureText(ellipsizedLabel)
+            val valueMaxWidth = metaContentWidth - labelMeasured - iconReservation
+            val ellipsizedValue = ellipsizeToWidth(field.value, valuePaint, valueMaxWidth)
+            canvas.drawText(ellipsizedLabel, metaContentLeft, textBaselineY, labelPaint)
+            val valueTextWidth = valuePaint.measureText(ellipsizedValue)
+            val icon = field.valueIcon
+            if (icon != null) {
+                val scaledIcon = icon.scale(iconPx, iconPx)
```
rkokhatskyi, line 254: `scaledIcon` is a fresh bitmap allocation per field per render that's never recycled — call `.recycle()` on it after `canvas.drawBitmap`

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
```diff
@@ -30,90 +81,196 @@ internal class MakeQrCodeBitmapShareFormatImpl @Inject constructor() : MakeQrCod
...
+                    rowsCursorY + dividerHeight / 2f,
+                    dividerPaint,
+                )
+                rowsCursorY += dividerHeight + fieldRowGap
+            }
+        }
+
+        val signatureTop = metaCardBottom + signatureTopGap
+        val scaledLogoSize = signatureIconSize.roundToInt().coerceAtLeast(1)
+        val scaledLogo = logo.scale(scaledLogoSize, scaledLogoSize)
```
rkokhatskyi, line 277: `scaledLogo` same thing — drawn once then abandoned without `.recycle()`

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
rkokhatskyi: previous `shareQrBitmap.value` is replaced without recycling — every `qrShareInfo` change (icon loads in, amount updates) leaks the prior render bitmap

data/src/test/kotlin/com/vultisig/wallet/data/usecases/QrShareInfoTest.kt
aminsato: Fixed in 3cde592ac. Dropped the two `mockk<Bitmap>` tests — the JVM source set has no Android stubs and the Bitmap equality branch is already covered implicitly by Kotlin's generated data-class `equals` (Bitmap uses reference equality, so drift would show up in the other field tests too). Added a docstring note explaining why `valueIcon` is intentionally skipped here.

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
aminsato: Fixed in 3cde592ac. Dropped the `scaleModifier = 4` pre-scale in `shareQr` — `MakeQrCodeBitmapShareFormat` already sizes the QR internally, matching the keysign path.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
aminsato: Fixed in 3cde592ac. `saveShareQrBitmap` now tracks a `saveShareQrBitmapJob` and cancels any in-flight render before starting a new one, so rapid `qrShareInfo` updates can't race. The replaced bitmap is also recycled (separate thread).

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
aminsato: Fixed in 3cde592ac. `valueMaxWidth` is now `coerceAtLeast(0f)` with an inline comment explaining the failure mode.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
aminsato: Fixed in 3cde592ac. Added a `@Volatile cachedTypeface` on `MakeQrCodeBitmapShareFormatImpl` and a `titleTypeface(context)` accessor that memoizes the first `loadBrockmannMedium` result. `getIdentifier` now runs once per process.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
aminsato: Fixed in 3cde592ac. Fallback now uses `Typeface.create(Typeface.DEFAULT, 500, false)` on API 28+ (the project min SDK is 26, so pre-P falls through to `Typeface.DEFAULT`).

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
aminsato: Fixed in 3cde592ac. Added `DisposableEffect` keyed on each bitmap for `srcIconBitmap`, `dstIconBitmap`, and `vultisigLogoBitmap`, recycling on disposal.

aminsato: Fixed in 3cde592ac. Added `DisposableEffect` keyed on each bitmap for `srcIconBitmap`, `dstIconBitmap`, and `vultisigLogoBitmap`, recycling on disposal.

app/src/main/java/com/vultisig/wallet/data/usecases/CreateQrCodeSharingBitmapUseCase.kt
aminsato: Fixed in 3cde592ac. Decoded logo is now a `by lazy val` on the use case — one decode per process. Kept the bitmap alive (not recycled) because the use case is a Hilt singleton and recycling would brick the next share.

app/src/main/res/values-de/strings.xml
aminsato: Fixed in 3cde592ac. `qr_share_label_to` is now `Zu` in de, matching the existing `send_to_address` wording.

app/src/main/res/values-ko/strings.xml
aminsato: Fixed in 3cde592ac. ko labels changed to `보내는` / `받는` to match the existing `send_from_address` / `send_to_address` terminology.

app/src/test/java/com/vultisig/wallet/data/models/CoinLogoTest.kt
aminsato: Fixed in 3cde592ac. Added 10 new alias tests covering every multi-case branch in `getCoinLogo` — solana, polygon, zksync, wif, raydium, astro, levana, fuzion, vulti, hyperliquid. Any typo in an alias will now fail the test instead of silently falling through to `chain.logo`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
aminsato: Fixed in 3cde592ac. `scaledIcon.recycle()` runs after `canvas.drawBitmap`, guarded by `!== icon` to avoid recycling the cached source when `scale` is a no-op.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
aminsato: Fixed in 3cde592ac. Same treatment — `scaledLogo.recycle()` after draw, guarded by `!== logo`.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
aminsato: Fixed in 3cde592ac. The previous `shareQrBitmap.value` is now captured and recycled before the flow emits the fresh render (guarded by `!== rendered` to skip same-reference emissions).

app/src/main/java/com/vultisig/wallet/data/models/CoinLogo.kt
aminsato: Agree the sealed-class refactor is the cleaner contract, but `ImageModel` is used broadly across the codebase (address bar, vault list, token rows, etc.) and refactoring to a sealed `DrawableRes / Url` would touch files well outside this PR. Leaving for a dedicated follow-up so this PR stays scoped to the Share QR redesign.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
aminsato: Agreed that the layout math deserves coverage, but the `Canvas` / `Paint` / `Typeface` interactions here are not straightforward to stub from the JVM source set (no Robolectric configured for `:data`, and mocking `Paint.measureText` would effectively test the mock). I'd rather do this as a follow-up that either adds Robolectric to `:data` or extracts pure-math helpers that can be tested without an Android context. Happy to open a ticket if you want to track it.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
```diff
@@ -213,19 +216,21 @@ constructor(
...
+        // race and let a stale bitmap overwrite a fresher one. The previous rendered bitmap is
+        // recycled on replacement to avoid leaks when icons/amounts change repeatedly.
+        saveShareQrBitmapJob?.cancel()
+        saveShareQrBitmapJob =
+            viewModelScope.launch {
+                val bitmap = qrBitmap ?: return@launch
+                val rendered =
+                    withContext(Dispatchers.IO) {
+                        makeQrCodeBitmapShareFormat(context, bitmap, color, logo, info)
+                    }
```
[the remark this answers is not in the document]
aminsato: Fixed in dc461e68b. Inside the `withContext(Dispatchers.IO)` block we now check `isActive` right after `makeQrCodeBitmapShareFormat` returns; if the outer job was cancelled mid-render we recycle the freshly allocated bitmap in place and return `null`, then the outer `launch` bails on `?: return@launch`. No more native memory leaking through the prompt-cancellation path.

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -205,21 +207,26 @@ constructor(
...
+                fields =
+                    listOf(
+                        QrShareField(
+                            context.getString(R.string.qr_share_label_vault),
+                            vaultName.forCanvasMinify(),
+                        ),
+                        QrShareField(
+                            context.getString(R.string.qr_share_label_type),
+                            context.getString(typeRes),
+                        ),
```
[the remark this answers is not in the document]
aminsato: Fixed in dc461e68b. `typeRes` now uses the same predicate as `loadData()` — fast only when `email` + `password` are both present AND not an active-vault migrate (`TssAction.Migrate && signers.size > 2`). Active-vault migrate now correctly advertises Secure Vault.

johnnyluo (approved):
LGTM

PR #4142, opened 2026-04-21
feat: Bandwidth/Energy resource tab for TRON freeze/unfreeze (#4014)

### LINKED ISSUE #4014: Tron DeFi: per-resource freeze/unfreeze callbacks (Bandwidth & Energy)
Follow-up to #3974. `TronFreezePositionCard` currently only supports freezing/unfreezing BANDWIDTH from the UI; ENERGY freeze/unfreeze is intentionally unreachable. Expose per-resource callbacks on the card so the UI can dispatch the correct `TronAction` per row.

aminsato:
## Summary
- Adds a Bandwidth/Energy segmented tab to the Send form when freezing or unfreezing TRX, letting users choose the resource instead of defaulting to Bandwidth.
- Aligns Android with the iOS `TronFreezeView` / `TronUnfreezeView` flow: the memo becomes `FREEZE:BANDWIDTH` / `FREEZE:ENERGY` / `UNFREEZE:BANDWIDTH` / `UNFREEZE:ENERGY` based on the tab selection.
- For UNFREEZE, the "Balance available" and percentage buttons now use the per-resource frozen balance fetched from `TronApi.getAccount()` (mirrors iOS `frozenBandwidthBalance` / `frozenEnergyBalance`).
- Closes #4014; removes the per-resource callback TODO on `TronFreezePositionCard`.

## Changes
- `TronStaking.kt` — new `TronResourceType`, `TronStakingOperation`, `tronStakingMemo()` helpers.
- `TronDeFiPositionsViewModel` / `TronDeFiPositionsScreen` — collapse `TronAction` to `FREEZE` / `UNFREEZE`; the card dispatches a resource-agnostic intent now.
- `SendFormViewModel` — inject `TronApi`; track `tronResourceType`; set initial memo on form init; refresh per-resource frozen balance for UNFREEZE; override percentage/max calc for UNFREEZE; route FREEZE through `loadAddresses` (wallet) while UNFREEZE keeps `loadDeFiAddresses` (frozen).
- `SendFormScreen` / `SendFormAmountSection` — render the new `TronResourceTypeTab` above the amount widget for freeze/unfreeze, switch the screen title, and honor `tronBalanceAvailableOverride` in the balance-available row.
- `TronResourceTypeTab.kt` — new pill-shaped segmented control, tints bandwidth icon with `alerts.success` and energy with `alerts.warning` when selected.
- Strings added to `values/` and all 9 other locale files.

## Test plan
- [ ] TRON DeFi → Freeze: tab defaults to Bandwidth, wallet balance shows, memo reads `FREEZE:BANDWIDTH`.
- [ ] Switch to Energy on the freeze screen → memo becomes `FREEZE:ENERGY`, amount input clears.
- [ ] TRON DeFi → Unfreeze: Bandwidth shows the frozen-bandwidth amount, Energy shows the frozen-energy amount; percentage buttons produce values bounded by the per-resource frozen balance.
- [ ] Continue through to Verify on both flows; transactions sign & broadcast successfully for each resource.
- [ ] Screen title matches Figma for freeze/unfreeze.
- [ ] All 10 locales render the new strings without fallback to English.

## Summary by CodeRabbit

* **New Features**
  * TRON freeze/unfreeze now offers a resource-type selector (Bandwidth vs Energy) in the send flow.
  * Unfreeze shows frozen amounts as the available balance; amount controls (max/percentage) use the selected frozen balance.
  * Freeze/unfreeze actions and staking memo handling consolidated for a simpler flow.

* **Documentation**
  * Added localized strings for TRON freeze/unfreeze and resource labels in DE, ES, HR, IT, KO, NL, PT, RU, ZH-CN.


app/src/main/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModel.kt
```diff
@@ -86,11 +86,9 @@ internal sealed interface TronDeFiUiState {
...
 @Immutable data class TronPendingWithdrawalUiModel(val amountTrx: String, val expiryEpochMs: Long)
 
-internal enum class TronAction(val memo: String, val defiType: DeFiNavActions) {
-    FREEZE_BANDWIDTH("FREEZE:BANDWIDTH", DeFiNavActions.FREEZE_TRX),
-    FREEZE_ENERGY("FREEZE:ENERGY", DeFiNavActions.FREEZE_TRX),
-    UNFREEZE_BANDWIDTH("UNFREEZE:BANDWIDTH", DeFiNavActions.UNFREEZE_TRX),
-    UNFREEZE_ENERGY("UNFREEZE:ENERGY", DeFiNavActions.UNFREEZE_TRX),
+internal enum class TronAction(val defiType: DeFiNavActions) {
+    FREEZE(DeFiNavActions.FREEZE_TRX),
+    UNFREEZE(DeFiNavActions.UNFREEZE_TRX),
```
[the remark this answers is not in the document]
aminsato: Not applicable. The position card exposes a single resource-agnostic Freeze button and a single Unfreeze button (see `TronFreezePositionCard.kt`) — there are no per-resource ENERGY/BANDWIDTH rows in this UI whose intent could be dropped. Resource selection is deliberately moved to the Send form via the new `TronResourceTypeTab`, which is the feature this PR (#4014) adds. Defaulting to BANDWIDTH on that tab is the intended initial state; the user picks Energy there, matching the iOS flow.

johnnyluo:
@aminsato where do I see the freeze/unfreeze button?  can't see it

Also can you please provide tx link for freeze and unfreeze

aminsato:
> where do I see the freeze/unfreeze button? can't see it

I think you didn’t enable Tron in the edit tron position section

Screenshot_20260422_121607

aminsato:
> Also can you please provide tx link for freeze and unfreeze

Staked Bandwidth:
https://tronscan.org/#/transaction/03bc6a6061ac25f7a147ef23c79ee41320296e1cbd19a9821bfc3d471e78cbe8

Unstaked Energy:
https://tronscan.org/#/transaction/f52291374a8378bee5ecc89aeb84cf6ea9f379a23d7a3924f7357ab6876648aa

Staked Energy:
https://tronscan.org/#/transaction/09d3f54d553922572e5a62482ce84d65a876dcdd572c79ad241caae4082df5d7

Unstaked Bandwidth:

https://tronscan.org/#/transaction/60a74761d24d856496d333321561b4cdae3dc21a1b690ccd510f29f1917ef1cb

app/src/main/java/com/vultisig/wallet/ui/screens/send/SendFormScreen.kt
```diff
@@ -478,6 +490,18 @@ private fun SendFormContent(
             state.defiType == DeFiNavActions.FREEZE_TRX ||
             state.defiType == DeFiNavActions.UNFREEZE_TRX
     ) {
+        val resourceType = state.tronResourceType
+        if (
+            resourceType != null &&
+                (state.defiType == DeFiNavActions.FREEZE_TRX ||
+                    state.defiType == DeFiNavActions.UNFREEZE_TRX)
+        ) {
```
johnnyluo: The inner `defiType` check is redundant — we're already inside an outer `if` that gates on the same two conditions. Simplify to just `if (resourceType != null)`.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronResourceTypeTab.kt
```diff
@@ -0,0 +1,114 @@
...
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.selected
```
johnnyluo, line 22: Unused import. `selected` from `androidx.compose.ui.semantics` is never referenced in this file — remove it to avoid a lint warning.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/GetTronFrozenBalancesUseCase.kt
```diff
@@ -0,0 +1,33 @@
...
+import java.math.RoundingMode
+import javax.inject.Inject
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+
+private const val SUN_PER_TRX = 1_000_000L
+private const val TRX_SCALE = 6
+
+data class TronFrozenBalances(val bandwidthTrx: BigDecimal, val energyTrx: BigDecimal) {
+    fun forResource(type: TronResourceType): BigDecimal =
```
johnnyluo, line 14: `forResource()` is defined here but never called anywhere in the PR. `SendFormViewModel` uses a `when` block directly instead. Per project conventions, avoid abstractions beyond what's currently needed — either use it to replace the `when` in the ViewModel, or remove it.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -476,7 +504,102 @@ constructor(
...
+        }
+        if (isTronStakingType()) {
+            applyTronStakingMemo(TronResourceType.BANDWIDTH)
+            if (defiType == DeFiNavActions.UNFREEZE_TRX) {
+                loadTronFrozenBalances()
+            }
+        }
+    }
+
+    fun setTronResourceType(type: TronResourceType) {
```
johnnyluo, line 515: `updateTronFrozenBalanceDisplay()` returns early (without clearing the override) when the state is `Loading` or `Error`. If the user switches tabs before balances finish loading, `tronBalanceAvailableOverride` will keep showing the previous tab's value. Consider clearing it immediately on tab switch:

```kotlin
uiState.update { it.copy(tronResourceType = type, tronBalanceAvailableOverride = null) }
```

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -2641,13 +2794,29 @@ constructor(
         }
     }
 
+    @OptIn(FlowPreview::class)
     private fun calculateSpecific() {
         viewModelScope.launch {
-            combine(selectedToken.filterNotNull(), gasFee.filterNotNull()) { token, gasFee ->
+            val dstAddressFlow =
```
johnnyluo, line 2816: Adding `dstAddressFlow` with a 300 ms debounce to `calculateSpecific()` affects fee calculation for **all chains**, not just TRON. If this is needed to support bandwidth-delegated-to-receiver fee estimation on TRON, it should either be scoped to TRON only or accompanied by a comment explaining the cross-chain impact.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModel.kt
```diff
@@ -128,10 +126,7 @@ constructor(
                 onError = { e ->
                     Timber.e(e, "Failed to load Tron DeFi data")
                     _state.value =
-                        TronDeFiUiState.Error(
-                            e.message?.asUiText()
-                                ?: R.string.error_view_default_description.asUiText()
-                        )
+                        TronDeFiUiState.Error(R.string.error_view_default_description.asUiText())
```
johnnyluo, line 129: The previous code surfaced the exception message to the user (`e.message?.asUiText() ?: …`). The new code always shows the generic fallback, which is friendlier but loses specificity. The exception is still logged via `Timber.e` above so nothing is truly lost — just worth a deliberate heads-up that specific API error messages will no longer appear in the UI.

johnnyluo (approved):
LGTM

johnnyluo:
@aminsato please check those few comments , once you solve it , then it is good to go

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -476,7 +504,102 @@ constructor(
...
+    }
+
+    private fun applyTronStakingMemo(type: TronResourceType) {
+        val op =
+            when (defiType) {
+                DeFiNavActions.FREEZE_TRX -> TronStakingOperation.FREEZE
+                DeFiNavActions.UNFREEZE_TRX -> TronStakingOperation.UNFREEZE
+                else -> return
+            }
+        memoFieldState.setTextAndPlaceCursorAtEnd(tronStakingMemo(op, type))
```
rkokhatskyi, line 542: this writes a `FREEZE:BANDWIDTH` style memo into `memoFieldState` and that field feeds `calculateGasFees` (line 2658) which constructs a `Transfer` with the memo `TronFeeService.calculateNativeTrxFee` then adds the `getMemoFee` charge (typically 1 TRX) but `TronHelper.buildFreezeBalanceV2` and `buildUnfreezeBalanceV2` never call `setMemo` on the actual transaction so the displayed gas estimate is 1 TRX higher than what gets paid and on FREEZE max it eats 1 TRX of freezable balance for nothing

```kotlin
val transferMemo = memo?.takeUnless { TRON_STAKING_MEMO_REGEX.matches(it) }
```

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -1297,8 +1443,14 @@ constructor(
...
-                        getAvailableTokenBalance(selectedAccount, gasFee.value)?.value
-                            ?: BigInteger.ZERO
+                        if (defiType == DeFiNavActions.UNFREEZE_TRX) {
+                            currentTronFrozenBalance()
+                                ?.movePointRight(selectedToken.decimal)
+                                ?.toBigInteger() ?: BigInteger.ZERO
+                        } else {
+                            getAvailableTokenBalance(selectedAccount, gasFee.value)?.value
+                                ?: BigInteger.ZERO
+                        }
```
rkokhatskyi, line 1445: this branch only validates the per resource frozen balance and skips the liquid TRX check that the non staking branch does via `getAvailableTokenBalance(selectedAccount gasFee.value)` so a vault with frozen TRX but zero liquid TRX (and exhausted free bandwidth quota) signs a tx that fails server side with insufficient balance and the user just sees a generic network error

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -1297,8 +1443,14 @@ constructor(
 
                 if (selectedToken.isNativeToken) {
                     val availableTokenBalance =
-                        getAvailableTokenBalance(selectedAccount, gasFee.value)?.value
-                            ?: BigInteger.ZERO
+                        if (defiType == DeFiNavActions.UNFREEZE_TRX) {
+                            currentTronFrozenBalance()
+                                ?.movePointRight(selectedToken.decimal)
+                                ?.toBigInteger() ?: BigInteger.ZERO
```
rkokhatskyi, line 1441: `currentTronFrozenBalance()` reads `uiState.value.tronResourceType` live so when the user toggles the resource tab while `send()` is awaiting `gasFee` the broadcast validates and signs against a different resource than the one selected when tapping Send the tab needs an `enabled` gate keyed off `state.isLoading` and the staking memo plus resource type should be captured into a local before any suspension

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -476,7 +504,102 @@ constructor(
...
+                    TronFrozenBalanceState.Error(e.message ?: "Unknown error")
+            }
+        ) {
+            uiState.update { it.copy(isTronFrozenBalancesLoading = true) }
+            try {
+                val vault =
+                    vault ?: vaultRepository.get(vaultId ?: return@safeLaunch) ?: return@safeLaunch
+                val trxCoin =
+                    vault.coins.firstOrNull { it.chain == Chain.Tron && it.isNativeToken }
+                        ?: return@safeLaunch
```
rkokhatskyi: all three `?: return@safeLaunch` exits leave `tronFrozenBandwidthTrx` and `tronFrozenEnergyTrx` stuck on `Loading` forever (the `finally` only flips the boolean) so a vault with no Tron coin gets the wallet balance with no error indication and percentage chips silently no op

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -476,7 +504,102 @@ constructor(
...
+        val op =
+            when (defiType) {
+                DeFiNavActions.FREEZE_TRX -> TronStakingOperation.FREEZE
+                DeFiNavActions.UNFREEZE_TRX -> TronStakingOperation.UNFREEZE
+                else -> return
+            }
+        memoFieldState.setTextAndPlaceCursorAtEnd(tronStakingMemo(op, type))
+    }
+
+    private fun loadTronFrozenBalances() {
```
rkokhatskyi, line 545: no `loadTronFrozenBalancesJob?.cancel()` here so rapid resource toggles stack concurrent `getTronFrozenBalances` calls that race to write the two `MutableStateFlow` (the older response can land last) also this method is never invoked from `refreshGasFee` so once `tronFrozenBandwidthTrx` flips to `Error` the user has no way back without leaving the screen

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -1297,8 +1443,14 @@ constructor(
...
+                            currentTronFrozenBalance()
+                                ?.movePointRight(selectedToken.decimal)
+                                ?.toBigInteger() ?: BigInteger.ZERO
+                        } else {
+                            getAvailableTokenBalance(selectedAccount, gasFee.value)?.value
+                                ?: BigInteger.ZERO
+                        }
 
                     if (tokenAmountInt > availableTokenBalance) {
                         throw InvalidTransactionDataException(
```
rkokhatskyi, line 1454: `R.string.send_error_insufficient_native_balance_with_fees` renders as `Insufficient TRX balance` which is misleading on UNFREEZE because the limit being hit is the per resource frozen balance not the wallet balance use a dedicated string for staking ops

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -325,6 +336,23 @@ constructor(
 
     private val isSwitchingAccounts = MutableStateFlow(false)
 
+    // Tron freeze/unfreeze state — populated when defiType is FREEZE_TRX or UNFREEZE_TRX.
+    private sealed class TronFrozenBalanceState {
```
rkokhatskyi: `sealed interface` matches the existing pattern in `TronDeFiUiState` and lifting this out of the ViewModel lets you expose the state in `SendFormUiModel` so the UI can render an actual error banner instead of falling back to the wallet balance silently when load fails

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -325,6 +336,23 @@ constructor(
...
+        data object Loading : TronFrozenBalanceState()
+
+        data class Loaded(val value: BigDecimal) : TronFrozenBalanceState()
+
+        data class Error(val message: String) : TronFrozenBalanceState()
+    }
+
+    private val tronFrozenBandwidthTrx =
+        MutableStateFlow<TronFrozenBalanceState>(TronFrozenBalanceState.Loading)
+    private val tronFrozenEnergyTrx =
```
rkokhatskyi: the `TronFrozenBalances` data class on `GetTronFrozenBalancesUseCase` already carries both values so collapsing these into one `MutableStateFlow<TronFrozenBalanceState>` (with the data class inside `Loaded`) removes the implicit invariant that bandwidth and energy must be in lockstep

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -476,7 +504,102 @@ constructor(
...
+            updateTronFrozenBalanceDisplay(type)
+        }
+    }
+
+    private fun applyTronStakingMemo(type: TronResourceType) {
+        val op =
+            when (defiType) {
+                DeFiNavActions.FREEZE_TRX -> TronStakingOperation.FREEZE
+                DeFiNavActions.UNFREEZE_TRX -> TronStakingOperation.UNFREEZE
+                else -> return
```
rkokhatskyi: both call sites already gate on `isTronStakingType()` so the `else -> return` is dead defense `error("applyTronStakingMemo called outside Tron staking")` would surface the precondition violation instead of silently no opping

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -819,6 +942,10 @@ constructor(
     }
 
     fun chooseMaxTokenAmount() {
+        // Early return if Tron frozen balances are still loading for UNFREEZE_TRX
```
rkokhatskyi: this comment narrates the boolean condition right below it so it can go (also at line 968)

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -476,7 +504,102 @@ constructor(
...
+            applyTronStakingMemo(TronResourceType.BANDWIDTH)
+            if (defiType == DeFiNavActions.UNFREEZE_TRX) {
+                loadTronFrozenBalances()
+            }
+        }
+    }
+
+    fun setTronResourceType(type: TronResourceType) {
+        if (uiState.value.tronResourceType == type) return
+        uiState.update { it.copy(tronResourceType = type) }
```
rkokhatskyi: this and the second `uiState.update` at line 529 are non atomic so an observer can briefly see the new resource type with the old `selectedAmountFraction` still selected fold both into one `copy` to make the transition atomic

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -476,7 +504,102 @@ constructor(
...
+            }
+        memoFieldState.setTextAndPlaceCursorAtEnd(tronStakingMemo(op, type))
+    }
+
+    private fun loadTronFrozenBalances() {
+        viewModelScope.safeLaunch(
+            onError = { e ->
+                Timber.e(e, "Failed to load Tron frozen balances")
+                tronFrozenBandwidthTrx.value =
+                    TronFrozenBalanceState.Error(e.message ?: "Unknown error")
```
rkokhatskyi: nothing in the codebase reads `TronFrozenBalanceState.Error.message` so this field is dead and the `"Unknown error"` literal is a hardcoded user facing string that bypasses the localization rule

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronResourceTypeTab.kt
rkokhatskyi, line 22: also the import is dangling because commit `a29c687` added a `Modifier.semantics { selected = isSelected; role = Role.Tab }` to address `coderabbitai` accessibility feedback but commit `c8928e0db` reverted the semantics block without removing the import so the screen reader regression is back and the PR claim that a11y was addressed no longer holds

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronResourceTypeTab.kt
```diff
@@ -0,0 +1,114 @@
...
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    val background = if (isSelected) Theme.v2.colors.backgrounds.surface2 else Color.Transparent
+    Row(
+        modifier =
+            modifier
+                .clip(RoundedCornerShape(77.dp))
+                .background(background)
+                .clickable { if (!isSelected) onClick() }
```
rkokhatskyi: `clickable(enabled = !isSelected) { onClick() }` would stop the ripple on the already selected segment and let TalkBack surface a disabled control instead of `double tap to activate` followed by silence

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronResourceTypeTab.kt
```diff
@@ -0,0 +1,114 @@
...
+                .padding(vertical = 12.dp),
```
rkokhatskyi, line 97: `padding(vertical = 12.dp)` plus a 16.dp icon gives a 40dp segment so the tap target is below the 48dp Material 3 minimum

app/src/main/res/values/strings.xml
```diff
@@ -81,6 +81,11 @@
     <string name="tron_defi_countdown_days">%1$d days, %2$d hrs</string>
     <string name="tron_defi_countdown_hours">%1$d hrs, %2$d min</string>
     <string name="tron_defi_error_trx_not_in_vault">TRX coin not found in vault</string>
+    <string name="tron_freeze_screen_title">Freezing TRON for resource</string>
```
rkokhatskyi: `for resource` reads as broken English and the Russian Korean and Chinese translators rendered it literally so all three locales are equally awkward consider `Freeze TRON` and let the resource tab below convey the resource

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronStaking.kt
```diff
@@ -2,3 +2,16 @@ package com.vultisig.wallet.data.blockchain.tron
 
 /** Matches exactly FREEZE:BANDWIDTH, FREEZE:ENERGY, UNFREEZE:BANDWIDTH, UNFREEZE:ENERGY */
 val TRON_STAKING_MEMO_REGEX = Regex("^(FREEZE|UNFREEZE):(BANDWIDTH|ENERGY)$")
```
rkokhatskyi, line 4: the literals `(FREEZE|UNFREEZE)` and `(BANDWIDTH|ENERGY)` are maintained by hand alongside `TronStakingOperation.memoPrefix` and `TronResourceType.name` so adding a third operation later requires updating two places building the regex from `entries` keeps them in sync

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronStaking.kt
```diff
@@ -2,3 +2,16 @@ package com.vultisig.wallet.data.blockchain.tron
...
+    ENERGY,
+}
+
+enum class TronStakingOperation(val memoPrefix: String) {
+    FREEZE("FREEZE"),
+    UNFREEZE("UNFREEZE"),
+}
+
+fun tronStakingMemo(operation: TronStakingOperation, resource: TronResourceType): String =
+    "${operation.memoPrefix}:${resource.name}"
```
rkokhatskyi, line 17: no test covers `tronStakingMemo` `TRON_STAKING_MEMO_REGEX` `GetTronFrozenBalancesUseCase.invoke` `setTronResourceType` or `loadTronFrozenBalances`

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronDeFiPositionsScreen.kt
```diff
@@ -80,11 +80,8 @@ internal fun TronDeFiPositionsScreen(
         onCancelEditPositionClick = { viewModel.setPositionSelectionDialogVisibility(false) },
         onDonePositionClick = viewModel::onPositionSelectionDone,
         onPositionSelectionChange = viewModel::onPositionSelectionChange,
-        // TODO(#4014): TronFreezePositionCard will pass the specific TronAction per row
-        //  once it exposes per-resource callbacks. Until then both default to BANDWIDTH.
-        //  ENERGY freeze/unfreeze is intentionally unreachable from the UI in this PR.
-        onClickFreeze = { viewModel.onTronAction(TronAction.FREEZE_BANDWIDTH) },
-        onClickUnfreeze = { viewModel.onTronAction(TronAction.UNFREEZE_BANDWIDTH) },
+        onClickFreeze = { viewModel.onTronAction(TronAction.FREEZE) },
```
rkokhatskyi, line 83: `LaunchedEffect(vaultId)` at line 74 only fires when `vaultId` itself changes so after the keysign success and back navigation from this `onClickFreeze` or `onClickUnfreeze` flow the page keeps the stale frozen balance and the freshly created pending withdrawal entry never appears. The broadcast path in `KeysignViewModel` invalidates the wallet balance cache but never the staking details cache or this screen state

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronDeFiPositionsScreen.kt
```diff
@@ -80,11 +80,8 @@ internal fun TronDeFiPositionsScreen(
...
+        onClickUnfreeze = { viewModel.onTronAction(TronAction.UNFREEZE) },
```
rkokhatskyi, line 84: two more issues in `TronPendingWithdrawalsCard` (line 226) the `withdrawals.forEach` inside one `LazyColumn` item bypasses virtualization so a user near the 32 unfreeze chain limit pays the recomposition cost for every row on every 1s `nowMs` tick and the polling interval is computed from `latestExpiry` (`maxOfOrNull`) so when the soonest pending withdrawal is minutes away but a later one is days away the countdown ticks every 60s instead of every 1s

app/src/main/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModel.kt
rkokhatskyi, line 91: also `mapPendingWithdrawals` at line 200 drops `TronUnfrozenV2Json.type` (BANDWIDTH or ENERGY) so the new resource tab feature splits freezing by resource but a user who unfroze both kinds can not tell which row in `TronPendingWithdrawalsCard` corresponds to which resource

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronFreezePositionCard.kt
```diff
@@ -101,6 +102,7 @@ internal fun TronFreezePositionCard(
                     background = Theme.v2.colors.backgrounds.tertiary_2,
                     contentColor = Theme.v2.colors.text.primary,
                     iconCircleColor = TronFreezeCardIconCircleColor,
+                    enabled = isUnfreezeEnabled,
```
rkokhatskyi, line 105: unrelated nit at line 70 the `circleinfo` icon has no `onClick` and no `contentDescription` so it looks tappable to sighted users but does nothing and screen readers skip it (compare to iOS which surfaces a freezing info sheet) either wire it to an info sheet or drop it

app/src/main/java/com/vultisig/wallet/ui/screens/send/SendFormScreen.kt
aminsato: Fixed in 978385b4a — simplified to `if (resourceType != null)` since the outer `if` already gates on FREEZE_TRX/UNFREEZE_TRX.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronResourceTypeTab.kt
aminsato: Fixed in 978385b4a — removed the unused `androidx.compose.ui.semantics.selected` import.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/GetTronFrozenBalancesUseCase.kt
aminsato: Fixed in 978385b4a — removed `forResource()` since `SendFormViewModel` dispatches on `TronFrozenBalanceState` fields (bandwidth vs energy) rather than on `TronFrozenBalances` values, so the helper was not applicable. If we add a callsite later that actually needs it, we can reintroduce it then.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fixed in 978385b4a — `tronBalanceAvailableOverride` is now cleared in the same `uiState.update` that sets the new `tronResourceType`, so the previous tab value is not shown while balances load.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fixed in 978385b4a — added a comment above `dstAddressFlow` explaining that the debounced dst address is forwarded to `getSpecific()` for all chains to support TRON bandwidth-delegated-to-receiver fee estimation, with the debounce avoiding per-keystroke recomputation.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModel.kt
aminsato: Intentional — the raw API error messages were often low-level (HTTP codes, deserialization failures) and not user-actionable, so we surface a generic fallback while still logging the exception via `Timber.e` above for debugging. Happy to switch back if the specific messages are preferred.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Great catch, thanks for the detailed trace. Fixed in 5d11992d9 — the fee-estimation `Transfer` now strips the staking-pattern memo via `memo.takeUnless { TRON_STAKING_MEMO_REGEX.matches(it) }` so the phantom memo fee no longer inflates the quote or eats into the FREEZE max.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Good point. Fixed in 5d11992d9 — on UNFREEZE we now additionally check that the liquid TRX balance covers the gas fee before signing, so the user sees an actionable error instead of a server-side insufficient-balance failure.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fixed in 5d11992d9 — the early-return paths now flip both frozen balance flows to `TronFrozenBalanceState.Error` so the loader resolves and the UI no longer silently falls back to the wallet balance.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fixed in 5d11992d9 — added `loadTronFrozenBalancesJob` and `.cancel()` it on each invocation, so rapid toggles no longer race. Calling `loadTronFrozenBalances` from `refreshGasFee` is worth considering but felt like a wider change; happy to pull it into this PR if you prefer.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fixed in 5d11992d9 — added a dedicated `send_error_insufficient_frozen_balance` string (with translations for all 10 locales) and route UNFREEZE overflows through it.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fixed in 5d11992d9 — swapped the silent `else -> return` for `error("applyTronStakingMemo called outside Tron staking")` so the precondition violation is loud.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fixed in 5d11992d9 — folded `tronResourceType`, `tronBalanceAvailableOverride`, and `selectedAmountFraction` into a single `uiState.update` so the transition is atomic.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fixed in 5d11992d9 — `TronFrozenBalanceState.Error` is now a `data object` (its `message` field was dead), which also removes the hardcoded `"Unknown error"` literal.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fixed in 5d11992d9 — `setTronResourceType` now no-ops while `uiState.value.isLoading` is true, so the tab cannot be toggled between the user tapping Send and the broadcast resolving.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronResourceTypeTab.kt
aminsato: Good catch on the a11y regression. Fixed in 5d11992d9 — restored the `Modifier.semantics { selected = isSelected; role = Role.Tab }` block so TalkBack announces the selected tab again.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronResourceTypeTab.kt
aminsato: Fixed in 5d11992d9 — switched to `clickable(enabled = !isSelected, onClick = onClick)` so the ripple stops on the active segment and TalkBack surfaces the correct disabled state.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronResourceTypeTab.kt
aminsato: Fixed in 5d11992d9 — added `Modifier.heightIn(min = 48.dp)` on the segment so the tap target meets the Material 3 minimum.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Good suggestion for readability. I'd rather keep this refactor out of the current PR to keep the diff focused on the resource-type tab feature — will open a follow-up to promote `TronFrozenBalanceState` to a top-level `sealed interface` and surface it via `SendFormUiModel` so the UI can render an error banner.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Agreed the two parallel flows are a bit of a smell. I'd prefer to tackle that in the same follow-up that reshapes the error surface — collapsing to a single `MutableStateFlow<TronFrozenBalanceState>` wrapping `TronFrozenBalances` is the right move once we have the broader state change in flight.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: These narration comments at lines 945 and 968 pre-date this PR — happy to clean them up, but I'd rather not pull unrelated edits into this diff. I'll include them in the follow-up that touches this ViewModel.

app/src/main/res/values/strings.xml
aminsato: Fair point on the English reading awkwardly. Since the string is already translated into 10 locales and this is mid-review, I'd prefer to revisit the copy + translations in a dedicated follow-up so we can coordinate with the translators rather than re-translate in flight.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronStaking.kt
aminsato: Good maintainability suggestion. The current regex is short enough that I think the duplication is tolerable for now, but I'm happy to derive it from `entries` in a follow-up when we add a third operation (unbond/claim) — that'll be the natural place to enforce the invariant.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronStaking.kt
aminsato: Agreed. I'll add unit coverage for `tronStakingMemo`, `TRON_STAKING_MEMO_REGEX`, `GetTronFrozenBalancesUseCase.invoke`, `setTronResourceType`, and `loadTronFrozenBalances` in a follow-up PR — keeping this one focused on the feature itself.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronDeFiPositionsScreen.kt
aminsato: Good catch on the post-keysign staleness. That refresh-invalidation is a broader concern across all DeFi screens, not just Tron — I'd prefer to address it in a dedicated PR that wires the staking-details cache invalidation into the keysign success path rather than bolt on a screen-specific workaround here.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronDeFiPositionsScreen.kt
aminsato: Both valid — the `forEach` inside a single `LazyColumn` item is a pre-existing pattern in this screen, and the polling cadence based on `maxOfOrNull` is also pre-existing. I'll open a follow-up to fix both (promote the rows to `items(withdrawals)` and use `minOfOrNull` for the next-tick interval).

app/src/main/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModel.kt
aminsato: Good UX gap — now that freezes are split by resource, the pending list should show which kind is unfreezing. Pulling that in here expands the diff scope; I'd rather handle it as a follow-up where we can wire `TronUnfrozenV2Json.type` through `mapPendingWithdrawals` and add the icon/label to the card.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronFreezePositionCard.kt
aminsato: Acknowledged — this is unrelated to the resource-type tab work (as you noted). I'll file a separate ticket to either wire the `circleinfo` icon to an info sheet (matching iOS) or drop it, rather than expand this PR's scope.

rkokhatskyi: @aminsato what's the ticket number? Please create and link here

app/src/main/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModel.kt
rkokhatskyi: @aminsato please create the ticket and link here

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronDeFiPositionsScreen.kt
rkokhatskyi: @aminsato please create the ticket and link here

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
rkokhatskyi: @aminsato this is directly related to the changes introduced in this PR, needs to be handled here

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
rkokhatskyi: @aminsato this is directly related to the changes introduced in this PR, needs to be handled here

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
rkokhatskyi: @aminsato this is directly related to the changes introduced in this PR, needs to be handled here

app/src/main/res/values/strings.xml
rkokhatskyi: @aminsato this is directly related to the changes introduced in this PR, needs to be handled here

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronStaking.kt
rkokhatskyi: @aminsato please open a ticket and link here

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronStaking.kt
rkokhatskyi: @aminsato this is directly related to the changes introduced in this PR, needs to be handled here

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronDeFiPositionsScreen.kt
rkokhatskyi: @aminsato please open a ticket and link here

rkokhatskyi:
@aminsato thanks for addressing the comments. Your AI bot silently resolved all the conversations - please tweak it so it actually leaves it up-to reviewer or your manual resolution. I have re-opened a number of threads where the bot wanted to defer it to future PRs while the requests are directly related to the current changes. Also for the threads where bot mentioned to handle in a separate PR due to pre-existing condition - please open a ticket for each of them and link to the corresponding thread

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronFreezePositionCard.kt
aminsato: Tracked in https://github.com/vultisig/vultisig-android/issues/4163.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModel.kt
aminsato: Tracked in https://github.com/vultisig/vultisig-android/issues/4161.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronDeFiPositionsScreen.kt
aminsato: Tracked in https://github.com/vultisig/vultisig-android/issues/4160.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronStaking.kt
aminsato: Tracked in https://github.com/vultisig/vultisig-android/issues/4162.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronDeFiPositionsScreen.kt
aminsato: Tracked in https://github.com/vultisig/vultisig-android/issues/4159.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fair, pulled it in. In 6f0266e45 — promoted `TronFrozenBalanceState` to a top-level `sealed interface` in `data/blockchain/tron/GetTronFrozenBalancesUseCase.kt` and surfaced the error via `SendFormUiModel.hasTronFrozenBalancesError`. `SendFormScreen` now renders a localized `tron_frozen_balances_error` footnote when the load fails instead of silently falling back to the wallet balance.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Fair, done in 6f0266e45 — collapsed `tronFrozenBandwidthTrx` and `tronFrozenEnergyTrx` into a single `MutableStateFlow<TronFrozenBalanceState>`, with `Loaded` wrapping `TronFrozenBalances`. `currentTronFrozenBalance()` and `updateTronFrozenBalanceDisplay()` now dispatch via `TronFrozenBalances.forResource()` so bandwidth and energy can no longer drift.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Done in 6f0266e45 — removed the narration comments above both `isTronFrozenBalancesLoading` early returns in `chooseMaxTokenAmount` and `choosePercentageAmount`.

app/src/main/res/values/strings.xml
aminsato: Done in 6f0266e45 — shortened to `Freeze TRON` / `Unfreeze TRON` (the resource tab already conveys the resource) and updated all 10 locale translations to match. The reviewer concern about the awkward English + literal translations should now be resolved.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronStaking.kt
aminsato: Done in 6f0266e45 — added unit tests covering the five requested targets:
- `tronStakingMemo` and `TRON_STAKING_MEMO_REGEX`: `data/src/test/kotlin/com/vultisig/wallet/data/blockchain/tron/TronStakingTest.kt`
- `GetTronFrozenBalancesUseCase.invoke` (plus `TronFrozenBalances.forResource`): `data/src/test/kotlin/com/vultisig/wallet/data/blockchain/tron/GetTronFrozenBalancesUseCaseTest.kt`
- `setTronResourceType` and `loadTronFrozenBalances`: `app/src/test/java/com/vultisig/wallet/ui/models/send/SendFormViewModelTronStakingTest.kt` — covers the loaded path, no-TRX-coin error, API failure error, tab switch updating the override, and the no-op-while-sending guard.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronDeFiPositionsScreen.kt
aminsato: Tracked in https://github.com/vultisig/vultisig-android/issues/4160.

rkokhatskyi (approved):
LGTM

PR #4144, opened 2026-04-21
fix: accept Swap transactions in all chain fee services

realpaaao:
## Summary

- `#3701` migrated the swap gas-fee path to `FeeServiceComposite.calculateFees(BlockchainTransaction)` but only wired Swap support into `EthereumFeeService`. Every other chain's `require(transaction is Transfer)` threw on Swap; the composite's try/catch swallowed it into `Timber.e` and then a flow-level `.catch` at `SwapFormViewModel.calculateGas` dropped the result silently, leaving `gasFee.value = null` and showing **"Gas fee calculation failed"** in the UI.
- User-visible: **Jupiter (Solana) → SPL token** and **THORChain/Tron**, **THORChain/XRP** swaps all hit this. Latent for Polkadot, Sui, Bittensor, Ton (no active provider today).
- Each fee service now handles Swap: short-circuit to `calculateDefaultFees` for Solana/Ripple/Tron/Polkadot/Sui/Bittensor (cleaner + saves RPC calls per 450ms gas tick), direct access for Zk/Ethereum. `Ton.calculateDefaultFees` guard removed.
- **Promoted `val to: String` to the `BlockchainTransaction` sealed interface** so services don't need polymorphic `when`-branches — both `Transfer` and `Swap` already had identical `to` fields.
- EVM side refactored: `ZkFeeService` extracts a private `estimateFee` helper shared by both `calculateFees` and `calculateDefaultFees` (replaces the old `error("Rpc Error…")` default). `EthereumFeeService` short-circuits Swap to its default-fee path.
- Tests added: `ZkFeeServiceTest` (new), `EthereumFeeServiceTest.calculateFees uses default-fee path for swap transactions`.

## Why it slipped through

- The contract mismatch (`require` narrowing a nominally-generic `BlockchainTransaction`) is invisible to the type system.
- `FeeServiceComposite:35-43` silently swallows the exception with `Timber.e(...)` and falls back to `calculateDefaultFees`, which for Solana/Tron/Ton also had the same guard — so the fallback also threw.
- `SwapFormViewModel.calculateGas` has a flow-level `.catch { Timber.e(it) }` that masks the failure until the user taps Swap and hits the null `gasFee`.
- No unit test covered `calculateFees(Swap(...))` for any chain — this PR adds that for EVM/Zk; follow-up for other chains.

## Regression commit

`07b753111` (#3701) "fix: reduce excessive DOGE fee estimation update fee estimation for send & swap" — silently migrated the swap gas-fee path from the legacy `GasFeeRepository` to `FeeServiceComposite`, activating the dormant `require(is Transfer)` guards that had been latent since the fee-service refactor (#2507).

## Test plan

- [x] `./gradlew :data:compileDebugKotlin` — BUILD SUCCESSFUL
- [x] `./gradlew :data:testDebugUnitTest` — 263 tests pass
- [x] `./gradlew :data:ktfmtFormat` — no changes
- [ ] `./gradlew :app:compileDebugKotlin :app:testDebugUnitTest` (reviewer / CI)
- [x] Manual: **SOL → HODL (Jupiter)** — original bug case
- [x] Manual: **TRX → BTC (THORChain)**
- [x] Manual: **XRP → BTC (THORChain)**
- [x] Manual: **ETH → anything** — regression check
- [x] Manual: **BTC → ETH (THORChain)** — UTXO regression check

## Follow-ups (not in this PR)

- Unit tests for `calculateFees(Swap(...))` across non-EVM chains (Solana, Tron, Ripple, etc.) — this bug pattern would not have slipped past CI with coverage.
- The silent `.catch` at `SwapFormViewModel.calculateGas` masked a showstopper for ~3 weeks; worth surfacing as a user-facing error or logging at a higher severity.
- `Tron.calculateDefaultFees` still calls `getAccount(ownAddress)` on the Swap path; always returns exists, pure waste. Low-pri cleanup.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added support for swap transactions across Ethereum, Solana, Sui, Tron, XRP, Bittensor, Polkadot and ZK chains with updated fee estimation flows and Mantle-specific swap limits.
  * Enhanced transaction model for improved recipient address handling.

* **Tests**
  * Added and updated tests to cover swap fee calculations and ZK fee estimation to ensure consistent fee outputs across networks.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
@@ -211,6 +211,7 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
         val chain = transaction.coin.chain
 
         return when {
+            transaction is Swap && chain == Chain.Mantle -> DEFAULT_MANTLE_SWAP_LIMIT
             transaction is Swap -> DEFAULT_SWAP_LIMIT
             chain == Chain.Arbitrum -> DEFAULT_ARBITRUM_TRANSFER
             transaction is Transfer && transaction.coin.isNativeToken -> DEFAULT_COIN_TRANSFER_LIMIT
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
@@ -23,6 +23,9 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
     FeeService {
 
     override suspend fun calculateFees(transaction: BlockchainTransaction): Fee {
+        if (transaction is Swap) {
+            return calculateDefaultFees(transaction)
+        }
         require(transaction is Transfer) {
             "Invalid Transaction Type ${transaction::class.simpleName}"
         }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/bittensor/BittensorFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/bittensor/BittensorFeeService.kt
@@ -5,6 +5,7 @@ import com.vultisig.wallet.data.blockchain.FeeService
 import com.vultisig.wallet.data.blockchain.model.BasicFee
 import com.vultisig.wallet.data.blockchain.model.BlockchainTransaction
 import com.vultisig.wallet.data.blockchain.model.Fee
+import com.vultisig.wallet.data.blockchain.model.Swap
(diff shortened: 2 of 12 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
```diff
@@ -23,6 +23,9 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
     FeeService {
 
     override suspend fun calculateFees(transaction: BlockchainTransaction): Fee {
+        if (transaction is Swap) {
```
johnnyluo, line 26: There is no need of this , in FeeServiceComposite , it fall back to calculateDefaultFee automatically

johnnyluo (approved):
LGTM

PR #4158, opened 2026-04-23
fix: render join-device keysign progress full-screen (#4157)

### LINKED ISSUE #4157: Join device Keysign screen shows navigation bar — should be full screen (pair sign only)
## Description

When performing a Keysign with a **secure vault** on the **join device** (pair sign flow), the Keysign progress screen is rendered inside the `V2Scaffold` navigation bar. It should instead be full screen, without the navigation bar, so the keysign progress animation takes over the whole viewport.

## Scope

- Affects: **pair sign / join device** only.
- Not affected: fastvault keysign works correctly.

## Expected behavior

During the keysign progress animation on the join device, the screen should render full screen with no top nav bar / back button / padding around it — same full-bleed treatment as the host device.

## Actual behavior

The join device wraps the keysign progress inside `V2Scaffold`, so the progress animation is constrained beneath a navigation bar with the "Keysign" title and back arrow.

## Relevant code

- `app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt` — `JoinKeysignScreen` wraps all branches in `V2Scaffold`.
- `app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt` — `KeysignView`'s `else` branch renders `KeysignRiveProgress`, which is the branch that needs to be full screen.

## Steps to reproduce

1. Set up a secure vault with 2 devices (pair sign).
2. Initiate a keysign (send/swap/deposit) on the host device.
3. Join the keysign from the second device.
4. Observe the keysign progress screen on the join device — it appears inside a navigation bar instead of full screen.

aminsato:
## Summary
- Drop the `V2Scaffold` wrapper on the join device while keysign is in progress so the Rive animation takes the full viewport (matches fastvault behavior). Closes #4157.
- Scaffold still wraps the `KeysignFinished` and `KeysignState.Error` branches (title + back / close button), and every non-Keysign outer state (DiscoveringSessionID, WaitingForKeysignStart, DiscoverService, JoinKeysign verify screens, outer Error).
- Also threads a `hasBackClick` flag into `KeysignView` so the keysign error screen on the join device can suppress its own back button (bundled because the compile path goes through the same call site).

## Implementation
- `JoinKeysignView.kt`: compute `isKeysignInProgress = state == Keysign && keysignState !is KeysignFinished && keysignState !is Error` and pass it as `fullScreen` to `JoinKeysignScreen`.
- `JoinKeysignScreen`: new `fullScreen: Boolean = false` param — when true, render `content()` directly; `BackHandler` stays outside the conditional so the system back button keeps working during the animation.
- `Keysign.kt` / `KeysignScreen.kt`: `KeysignView` gains a `hasBackClick` parameter that gates the error screen's back button.

## Test plan
- [ ] Secure vault pair-sign Send: on join device, verify Rive progress is full-screen (no top bar, no back button).
- [ ] Secure vault pair-sign Swap and Deposit: same check.
- [ ] After successful keysign, confirm the transaction-complete screen on the join device still shows scaffold with "Transaction Complete" title.
- [ ] If keysign errors, confirm the error screen on the join device renders inside scaffold.
- [ ] Fastvault keysign: confirm no regression (was already full-screen on host flow).
- [ ] DiscoveringSessionID / WaitingForKeysignStart / DiscoverService screens still show the "Keysign" scaffold title and back button.
- [ ] Outer error (wrong vault share etc.) still shows scaffold with close (X) icon.

## Summary by CodeRabbit

* **Improvements**
  * The signing interface now switches to a true fullscreen mode during active multi-party signing, removing surrounding scaffolding and toolbar for a focused experience.
  * Back-button behavior is now context-aware across the signing flow and error screens, enabling or disabling navigation appropriately to prevent accidental interruptions and clarify retry/back actions.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
@@ -46,10 +46,7 @@ internal fun JoinKeysignView(navController: NavHostController) {
         viewModel.enableNavigationToHome()
     }
     val state by viewModel.currentState.collectAsState()
-    val isKeysignInProgress =
-        state == Keysign &&
-            keysignState !is KeysignState.KeysignFinished &&
-            keysignState !is KeysignState.Error
+    val isKeysignInProgress = state == Keysign && keysignState.isInProgress
     JoinKeysignScreen(
         isKeySignFinished = keysignState is KeysignState.KeysignFinished,
         onBack = viewModel::navigateToHome,
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
@@ -46,10 +46,15 @@ internal fun JoinKeysignView(navController: NavHostController) {
         viewModel.enableNavigationToHome()
     }
     val state by viewModel.currentState.collectAsState()
+    val isKeysignInProgress =
+        state == Keysign &&
+            keysignState !is KeysignState.KeysignFinished &&
+            keysignState !is KeysignState.Error
     JoinKeysignScreen(
         isKeySignFinished = keysignState is KeysignState.KeysignFinished,
         onBack = viewModel::navigateToHome,
         isError = state is Error,
+        fullScreen = isKeysignInProgress,
     ) {
         when (state) {
             DiscoveringSessionID,
@@ -141,6 +146,7 @@ internal fun JoinKeysignView(navController: NavHostController) {
                     onAddToAddressBook = keysignViewModel::navigateToAddressBook,
                     showSaveToAddressBook =
                         keysignViewModel.showSaveToAddressBook.collectAsState().value,
+                    hasBackClick = false,
                 )
             }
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
```diff
@@ -46,10 +46,15 @@ internal fun JoinKeysignView(navController: NavHostController) {
         viewModel.enableNavigationToHome()
     }
     val state by viewModel.currentState.collectAsState()
+    val isKeysignInProgress =
+        state == Keysign &&
+            keysignState !is KeysignState.KeysignFinished &&
+            keysignState !is KeysignState.Error
```
rkokhatskyi: Please move in progress detection into `KeysignState` itself. Like: 

```Kotlin
internal sealed class KeysignState {
    data object CreatingInstance : KeysignState()

    data object KeysignECDSA : KeysignState()

    data object KeysignEdDSA : KeysignState()

    data object KeysignMLDSA : KeysignState()

    data class KeysignFinished(val transactionStatus: TransactionStatus) : KeysignState()

    data class Error(val errorMessage: UiText) : KeysignState()

    val isInProgress: Boolean
        get() = when (this) {
            CreatingInstance,
            KeysignECDSA,
            KeysignEdDSA,
            KeysignMLDSA -> true

            is KeysignFinished,
            is Error -> false
        }
}
```

and then you can just call that property

aminsato:
@coderabbitai full review

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
aminsato: Good call, done in c38a656f6 — added `isInProgress` as a member property on `KeysignState` exactly as you laid out, and `JoinKeysignView` now just uses `keysignState.isInProgress`. Much cleaner.

rkokhatskyi (approved):
LGTM

PR #4165, opened 2026-04-23
fix: UTXO swap byteFee unit mismatch causing Error_not_enough_utxos (#4164)

aminsato:
Fixes #4164

https://blockchair.com/dogecoin/transaction/a4707de26787a940a1f54866debf99fe203660fc56279df3031577ec555a603e

## Summary
- `SwapGasCalculator.calculateGasFee` was reassigning `gasFee` to `plan.fee` (total sats) for UTXO chains. That value propagated into `BlockChainSpecific.UTXO.byteFee` and was handed to WalletCore's signing input as a per-byte rate, inflating the mining fee quote and triggering `Error_not_enough_utxos` at keysign.
- Keep `gasFee` as the per-byte rate (used by `UtxoHelper.setByteFee(...)` for signing). Introduce a local `estimatedTotalFee` that carries `plan.fee` only into `gasFeeToEstimatedFee` for display and the balance-check path.
- Matches the iOS contract documented in `BlockChainSpecific.swift`: *"For UTXO, gas represents the byteFee (sats/byte rate), not the total fee."*

## Test plan
- [ ] Swap DOGE → RUNE on THORChain from a vault with sufficient balance; keysign proceeds without `Error_not_enough_utxos`.
- [ ] Swap BTC → RUNE on THORChain; keysign proceeds successfully.
- [ ] Displayed network fee on the swap screen still matches the planned total fee (not the per-byte rate).
- [ ] Regular BTC send (non-swap) still works and displays the correct fee.
- [ ] Balance-check rejects swaps that would actually exceed balance + fee.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate gas fee estimation for swap transactions on UTXO-based chains (non-Cardano). The fee reported to users is now stable while the transaction estimation process computes a separate, clearer estimated total fee so displayed estimates better reflect actual expected costs.

Issue #4164, opened 2026-04-23
THORChain swap on UTXO chains fails with `Error_not_enough_utxos` (byteFee unit mismatch)


aminsato:
## Problem

Starting a THORChain swap from a UTXO chain (Bitcoin, Dogecoin, Litecoin, Bitcoin Cash) fails at keysign with:

```
java.lang.IllegalStateException: PreSigningOutput contains error: Error_not_enough_utxos
    at com.vultisig.wallet.data.crypto.UtilsKt.checkError(Utils.kt:22)
    at com.vultisig.wallet.data.wallet.Swaps.getPreSignedImageHash(Swaps.kt:25)
    at com.vultisig.wallet.data.chains.helpers.THORChainSwaps.getPreSignedImageHash(THORChainSwaps.kt:108)
    at com.vultisig.wallet.data.chains.helpers.SigningHelper.getKeysignMessages(SigningHelper.kt:74)
    at com.vultisig.wallet.ui.models.keysign.KeysignFlowViewModel.setData(KeysignFlowViewModel.kt:283)
```

The user is shown `Signing Error. Please try again. PreSigningOutput contains error: Error_not_enough_utxos`, even though the vault balance is sufficient to cover the swap amount plus a reasonable mining fee.

## Root cause

In `SwapGasCalculator.calculateGasFee` (`app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapGasCalculator.kt`), for UTXO chains the `gasFee` variable is initialized with the **per-byte rate** returned by `UtxoFeeService` (sats/byte), then **overwritten** with `plan.fee.toBigInteger()` — the **total** transaction fee from the WalletCore plan — before being returned.

That returned `gasFee` is later stored into `BlockChainSpecific.UTXO.byteFee` via `BlockChainSpecificRepository.getSpecific`, then handed to WalletCore at signing time through `UtxoHelper.setByteFee(utxo.byteFee.toLong())`. WalletCore interprets `byteFee` as **sats per byte**, so the total fee (e.g. a few thousand sats) gets multiplied by the tx size (hundreds of bytes). The planner computes an enormous mining fee that exceeds the vault balance and returns `Error_not_enough_utxos`.

[...]

PR #4174, opened 2026-04-23
test: cover UTXO swap gasFee per-byte rate regression (#4165)

aminsato:
## Summary
Adds `SwapGasCalculatorTest` to lock in the UTXO swap gas-fee fix from #4164 / #4165.

Addresses review feedback from @rkokhatskyi on #4165:
> https://github.com/vultisig/vultisig-android/pull/4165#pullrequestreview-4161470178
> "can you please attach other UTXO chains tx samples as well? Also cover this by tests, essential to have"

https://blockchair.com/tron/transaction/c2c36c2bcc1a75847fbd50d302b24191105f57146c66573bbcdd860ede38f370

https://blockchair.com/dash/transaction/902c956a0c91a5dab21053fba3f553694f9142dc5342de4abbd154adbc5c7d50

## What's covered
- `gasFee` stays as the per-byte sats/byte rate (what `UtxoHelper.setByteFee(...)` expects) while the plan's total fee is routed into `estimated` for display/balance checks only.
- `calculateGasFee` short-circuits to `null` when the Bitcoin plan fails (e.g. `Error_not_enough_utxos`), preventing a stale zero estimate.

## Test plan
- [ ] `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.swap.SwapGasCalculatorTest"` passes
- [ ] CI green

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for swap gas fee calculations covering multiple UTXO chains (BTC, BCH, LTC, DOGE, DASH, ZEC).
  * Verifies per-byte gas rate is preserved and total plan fee is routed to the estimated result.
  * Adds short‑circuit tests when UTXO transaction planning fails (returns null/no estimate).
  * Confirms non‑UTXO chains (Cardano, Ethereum) bypass UTXO planning and use fee service output directly.

### DIFF
--- a/app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapGasCalculatorTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapGasCalculatorTest.kt
@@ -1,0 +1,221 @@
+@file:OptIn(ExperimentalCoroutinesApi::class)
+
+package com.vultisig.wallet.ui.models.swap
+
+import com.vultisig.wallet.data.blockchain.FeeServiceComposite
+import com.vultisig.wallet.data.blockchain.model.BasicFee
+import com.vultisig.wallet.data.chains.helpers.UtxoHelper
+import com.vultisig.wallet.data.models.Account
+import com.vultisig.wallet.data.models.Address
+import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.models.EstimatedGasFee
+import com.vultisig.wallet.data.models.FiatValue
+import com.vultisig.wallet.data.models.GasFeeParams
+import com.vultisig.wallet.data.models.TokenValue
+import com.vultisig.wallet.data.models.Vault
+import com.vultisig.wallet.data.models.payload.BlockChainSpecific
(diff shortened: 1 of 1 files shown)


johnnyluo:
@aminsato build fail

rkokhatskyi (commented):
@aminsato please include more tests: 

1. Doge only covered, needs to cover all UTXO chains
2. Cardano also needs to be covered
3. Non-UTXO chains too as negative scenario

aminsato:
Thanks @rkokhatskyi — added coverage in 37f4b7a94:

1. **All UTXO chains** now covered via `@ParameterizedTest` / `@EnumSource` over Bitcoin, BitcoinCash, Litecoin, Dogecoin, Dash, Zcash — both the happy path (gasFee stays per-byte, plan fee routes to estimated) and the plan-failure short-circuit.
2. **Cardano** has a dedicated test asserting it bypasses the WalletCore plan path (UTXO standard but explicitly excluded) — verifies `UtxoHelper.getHelper` and `blockChainSpecificRepository.getSpecific` are never invoked.
3. **Non-UTXO negative scenario** added for Ethereum with the same `verify(exactly = 0)` checks on `UtxoHelper` / specific repo.

14 tests total, all green locally.

johnnyluo (approved):
LGTM

PR #4264, opened 2026-04-25
fix: buffer the latest response so a late requester does not hang

### LINKED ISSUE #4263: [Fix] RequestResultRepository drops the response when it fires before the requester subscribes
## Problem

`RequestResultRepositoryImpl.results` is a `MutableSharedFlow<Response>()` with the default `replay = 0` and no extra buffer. When `respond(requestId, result)` runs before the matching `request(requestId)` collector subscribes the emission has nowhere to land. The responder either suspends forever (default `BufferOverflow.SUSPEND`) or the value is consumed by no one, and the requester is left waiting on `results.first { it.requestId == id }` indefinitely.

This race shows up in the QR scan and custom token search flows: the producing screen dispatches its result and the consuming screen, which subscribes a moment later, simply hangs.

## Expected Behavior

`request(id)` must resolve regardless of whether the emission happens before or after the subscription. A late subscriber should pick up the most recent response for its id and continue.

## Steps to Reproduce

1. Open the Add Custom Token flow and pick a token from a screen that calls `respond(REQUEST_SEARCHED_TOKEN_ID, coin)`
2. Navigate back fast enough that the requester re-subscribes after the emission
3. The Token Selection screen never receives the picked token and stalls

## Solution

Switch the backing flow to `MutableSharedFlow(replay = 1)`. The replay buffer holds the latest emission so a late subscriber receives the response that fired before it attached. The existing `first { it.requestId == id }` filter still rejects stale or unrelated ids.

```kotlin
private val results = MutableSharedFlow<Response>(replay = 1)
```

### Must Do
- [ ] Set `replay = 1` on the backing `MutableSharedFlow`
- [ ] Cover both orderings (respond first, request first), the id mismatch path and concurrent requests with unit tests

### Must NOT Do
- Change the public surface of `RequestResultRepository`
- Replace the architecture with channels or deferreds in this PR

## Acceptance Criteria

- [ ] `request(id)` resolves whether `respond(id, value)` runs before or after the subscription
- [ ] An unrelated `respond` does not unblock a waiting `request`
- [ ] A null payload returns null instead of throwing

rkokhatskyi:
## Description

Fixes #4263.

The original `MutableSharedFlow<Response>()` (replay = 0) lost emissions when `respond(id, value)` ran before the matching `request(id)` collector subscribed — the QR scan and custom token search flows hung whenever the producing screen finished a beat ahead of the consuming screen.

The first iteration of this PR moved the buffer to `replay = 1` to bridge that race. @aminsato correctly flagged that this introduced a stale-replay regression: with the singleton scoped to the whole app and call sites reusing constant ids (`REFRESH_CHAIN_DATA`, `REFRESH_TOKEN_DATA`, `REQUEST_SEARCHED_TOKEN_ID`), the buffered emission satisfied the next `request(sameId)` for the next exchange, firing `loadData` before the user had interacted with the second screen.

This revision drops `MutableSharedFlow` and replaces it with a `Mutex`-guarded `Map<String, CompletableDeferred<Any?>>` (per @aminsato's suggestion). Each `request` consumes the slot for its id when await resolves, so a subsequent `request` for the same id starts fresh and waits for a new `respond`. Concurrent waiters on the same id share one rendezvous and resolve together. `respond` removes the slot from the map *before* completing the deferred so a fresh request that races the cleanup cannot pick up the just-completed deferred and replay across exchanges. A `respond` with no current waiter buffers the value for the next `request`; a subsequent `respond` on the same id replaces the buffered value.

KDoc spells out null/type-mismatch semantics: `null` means the producer published `null`; a wrong-type payload propagates to the call site where Kotlin's `CHECKCAST` raises `ClassCastException` (JVM erasure leaves `T` unenforced inside the function body).

The fix intentionally diverges from the issue's "Must NOT Do: Replace the architecture with channels or deferreds" — the constraint was the wrong call given the singleton's lifetime, since the `SharedFlow` shape can't satisfy both the original race fix (replay > 0) and no cross-exchange leakage at the same time.

## Tests

Ten unit tests in `RequestResultRepositoryImplTest` cover:

- respond before subscribe (original race fix)
- respond after subscribe (normal case)
- id mismatch (unrelated respond does not unblock)
- concurrent distinct ids (per-id isolation)
- null payload (`null` flows through)
- **second request for same id (regression — fails on `replay = 1`, passes on the new impl)**
- concurrent waiters on the same id (single respond resolves all)
- cancelled request (next request is unblocked by a new respond)
- cancel one of two concurrent waiters (the other still resolves)
- latest orphan response wins (subsequent orphan replaces earlier one)

## Which feature is affected?

- [ ] Sending
- [ ] Swap

Internal infrastructure fix — the rendezvous bus is used by Add Chain, Add Token, Refresh Token Data, QR scan, custom token search, address book, and asset/network selectors.

## Checklist

n/a


aminsato:
@coderabbitai full review

data/src/main/kotlin/com/vultisig/wallet/data/repositories/RequestResultRepository.kt
```diff
@@ -15,7 +15,7 @@ internal class RequestResultRepositoryImpl @Inject constructor() : RequestResult
 
     private data class Response(val requestId: String, val result: Any?)
 
-    private val results = MutableSharedFlow<Response>()
+    private val results = MutableSharedFlow<Response>(replay = 1)
```
aminsato: `replay = 1` fixes the late-subscriber hang for one round-trip but I think it introduces a stale-replay regression on the next request with the same id. Several call sites reuse a constant id across the singleton repo lifetime — `REFRESH_CHAIN_DATA`, `REFRESH_TOKEN_DATA`, `REQUEST_SEARCHED_TOKEN_ID` — so the buffer outlives the exchange.

Concrete scenario for `REFRESH_CHAIN_DATA`:

1. User taps "Add Chain" → `VaultAccountsViewModel.openAddChainAccount` routes to `ChainSelection` then suspends on `request<Unit>(REFRESH_CHAIN_DATA)`.
2. User picks a chain → `ChainSelectionViewModel` calls `respond(REFRESH_CHAIN_DATA, Unit)`. Request resumes, `loadData` runs. **Replay buffer now holds `Response(REFRESH_CHAIN_DATA, Unit)`.**
3. User taps "Add Chain" a second time → `request<Unit>(REFRESH_CHAIN_DATA)` returns **immediately** from the cached emission, before the user has interacted with the new chain selection screen. `loadData` fires prematurely.
4. The user then actually picks a chain → `respond` fires with no listener; the new value is just buffered for the next premature trigger.

The existing tests don't catch this because they all use a fresh `RequestResultRepositoryImpl` per test, but in production it's a `@Singleton`-style instance.

A scoped alternative (e.g. `Map<String, CompletableDeferred<Any?>>` or one `Channel(CONFLATED)` per id, removed when consumed) would fix the original race without keeping responses cached past their consumer. Worth at least adding a regression test that issues two sequential `request`/`respond` pairs against the same repo instance and asserts the second `request` suspends until the second `respond`.:

rkokhatskyi: Yep, Claude drifted here...

I will handle it, will put to draft for now

rkokhatskyi: Confirmed the stale-replay regression so switched the backing store to a `Mutex`-guarded `Map<String, CompletableDeferred<Any?>>` per your suggestion, where each request consumes its slot in `finally` which means a subsequent request for the same id starts fresh, while concurrent waiters share one deferred so a single respond resolves all of them, and `respond` removes the slot from the map *before* completing the deferred since otherwise a fresh request racing the cleanup could pick up the just-completed deferred and replay across exchanges; the new tests cover the sequential-exchange regression you described, multi-waiter resolution, cancellation cleanup, single-cancellation amid concurrent waiters, and latest-orphan replacement, where the regression test (`second request for same id does not see previous exchange's response`) fails on `replay = 1` and passes on the new impl. Pushed in fce921c74.

rkokhatskyi:
@coderabbitai the nitpick's premise is incorrect for a non-reified generic on the JVM — `T` erases to `Any?` so both `as? T` and `as T` are no-ops inside the function body where the value passes through as-is, while the `ClassCastException` actually fires at the call site via the compiler-inserted `CHECKCAST` when the caller binds the return to a typed variable. Verified empirically against this code: `respond("alpha", 42)` followed by `request<String>("alpha")` bound to `Any?` returns the boxed `Int 42` (no null), while binding to `String?` throws CCE at the call site (`class java.lang.Integer cannot be cast to class java.lang.String`); switching to `as T` would not change this since erasure makes both forms equivalent inside the function. Kept `as? T` since it correctly returns `null` for `null` payloads which is the *only* case this function returns `null`, and added a KDoc clarification in eab99d2e7 documenting that `null` means a `null` payload and type mismatches surface as `ClassCastException` at the call site.

johnnyluo (approved):
LGTM

PR #4267, opened 2026-04-25
fix: return null instead of hanging when transaction id is unknown

### LINKED ISSUE #4265: [Fix] TransactionRepository and CustomMessagePayloadRepo hang forever when id is unknown
## Problem

`TransactionRepositoryImpl.getTransaction(id)` is implemented as `transactions.mapNotNull { it[id] }.first()`. When the id does not exist in the in memory map, `mapNotNull` filters out the only emission of the underlying `MutableStateFlow` and `first()` suspends until something else emits. Nothing else ever emits for that id so the call hangs for the lifetime of the process. `CustomMessagePayloadRepoImpl.get(id)` has the same shape and the same hang.

The interface signature `suspend fun getTransaction(id): Transaction` lies to its callers — it claims it always returns a value but in practice it can deadlock the calling coroutine.

## Expected Behavior

The call must finish in bounded time. A missing id should return `null` and let the caller decide how to surface the error. The contract should make the missing case visible at the type level.

## Steps to Reproduce

1. Start any flow that derives a `transactionId` from saved navigation state (verify, sign message, keysign share, keysign flow)
2. Provide an id that no longer maps to an entry in the in memory map (process death, deep link, manual back stack manipulation)
3. The screen sits with a spinner forever; no exception, no UI feedback

## Solution

Switch both repositories to a synchronous `value[id]` lookup over the underlying `MutableStateFlow` and change the return types to nullable. The wait-forever flow gymnastics disappear and callers see the real contract.

```kotlin
override suspend fun getTransaction(id: TransactionId): Transaction? =
    transactions.value[id]
```

Existing call sites are updated to handle null with `error(\"...\")` for paths that already trusted the lookup, and graceful navigation back for paths that already had a `runCatching` recovery.

### Must Do
- [ ] Make `getTransaction` and `get` return nullable types and resolve them via `value[id]`
- [ ] Update every caller (`VerifyTransactionViewModel`, `KeysignShareViewModel`, `VerifySignMessageViewModel`, `KeysignFlowViewModel`) to handle the null result
- [ ] Cover both repositories with unit tests across empty, present and missing id scenarios

### Must NOT Do
- Replace the in memory storage with a different backing store
- Change unrelated state in the affected ViewModels

## Acceptance Criteria

- [ ] Calling `getTransaction(id)` on a missing id returns `null` immediately
- [ ] Calling `customMessagePayloadRepo.get(id)` on a missing id returns `null` immediately
- [ ] All callers compile and gracefully react to the null path

rkokhatskyi:
## Description

Fixes #4265

`TransactionRepositoryImpl.getTransaction(id)` was implemented as `transactions.mapNotNull { it[id] }.first()`. When the id was missing the only emission of the underlying `MutableStateFlow` was filtered out and `first()` suspended forever. `CustomMessagePayloadRepoImpl.get(id)` carried the same shape and the same hang. The interface signature claimed a non null return but the runtime behaviour was a deadlock.

Both repositories now resolve the lookup synchronously through `value[id]` and surface the truth at the type level by returning a nullable. Every caller (`VerifyTransactionViewModel`, `KeysignShareViewModel`, `VerifySignMessageViewModel`, `KeysignFlowViewModel`) is updated: paths that already had a recovery branch keep navigating back, paths that always trusted the lookup raise `error(...)` so a missing id becomes a visible exception instead of a silent hang.

Two new test classes (`TransactionRepositoryImplTest`, `CustomMessagePayloadRepoImplTest`) cover empty repository, missing id, present id, overwrite by id and independent storage, ten unit tests in total. The full app test suite stays green.

## Which feature is affected?

## Checklist

n/a

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling and user navigation when transactions or message payloads are not found.
  * Enhanced null-safety validation to prevent application crashes from missing data references.
  * Strengthened data repository lookups with explicit null checks.

* **Tests**
  * Added comprehensive test coverage for transaction and message payload data retrieval scenarios.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
@@ -38,6 +38,7 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
+import timber.log.Timber
 import vultisig.keysign.v1.CustomMessagePayload
 
 @HiltViewModel
@@ -71,7 +72,10 @@ constructor(
     suspend fun loadTransaction(transactionId: TransactionId) {
         val transaction =
             transactionRepository.getTransaction(transactionId)
-                ?: error("Transaction not found: $transactionId")
+                ?: run {
+                    Timber.e("Transaction not found: %s", transactionId)
+                    throw IllegalStateException()
+                }
 
         val vault = vaultRepository.get(transaction.vaultId)!!
 
@@ -100,7 +104,12 @@ constructor(
     }
 
     suspend fun loadSignMessageTx(id: String) {
-        val dto = customMessagePayloadRepo.get(id) ?: error("Sign message payload not found: $id")
+        val dto =
+            customMessagePayloadRepo.get(id)
+                ?: run {
+                    Timber.e("Sign message payload not found: %s", id)
+                    throw IllegalStateException()
+                }
 
         val vault = vaultRepository.get(dto.vaultId)!!
 
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
@@ -69,7 +69,9 @@ constructor(
     private var saveShareQrBitmapJob: Job? = null
 
     suspend fun loadTransaction(transactionId: TransactionId) {
-        val transaction = transactionRepository.getTransaction(transactionId)
+        val transaction =
+            transactionRepository.getTransaction(transactionId)
+                ?: error("Transaction not found: $transactionId")
 
         val vault = vaultRepository.get(transaction.vaultId)!!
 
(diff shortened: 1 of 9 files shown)


aminsato:
@coderabbitai full review

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignShareViewModel.kt
```diff
@@ -69,7 +70,12 @@ constructor(
     private var saveShareQrBitmapJob: Job? = null
 
     suspend fun loadTransaction(transactionId: TransactionId) {
-        val transaction = transactionRepository.getTransaction(transactionId)
+        val transaction =
+            transactionRepository.getTransaction(transactionId)
+                ?: run {
+                    Timber.e("Transaction not found: %s", transactionId)
+                    throw IllegalStateException()
```
aminsato, line 77: nit: per CLAUDE.md ("Use `require`/`check`/`error` for preconditions, not raw exceptions"), prefer `error("Transaction not found: $transactionId")` over `throw IllegalStateException()`. It also folds the id into the stack trace so a crash report stands on its own without having to correlate with the Timber line above. Same applies to line 111 for the sign-message payload.

johnnyluo (approved):
LGTM

PR #4277, opened 2026-04-25
fix: restore SendFormScreen.amountField testTag (#4180)

### LINKED ISSUE #4180: Send form: SendFormScreen.amountField testTag removed in v1.0.99 — breaks QA automation
## Summary

The Compose testTag `SendFormScreen.amountField` was removed from the Android send form during the v1.0.99 refactor (PRs #4007 split `SendFormScreen.kt` and #4108 split `SendFormViewModel.kt`). The testTag `SendFormScreen.chainSelector` and `SendFormScreen.addressField` still exist and work; only the amount field lost its stable selector.

This silently breaks every automated Maestro send flow — the tap is `optional: true` so the test proceeds, the amount stays at `0`, Continue does nothing, and the flow times out waiting for the verify screen.

## Repro

On a Pixel 9 Pro emulator with `v1.0.99.apk` installed and the Fast Vault restored:

```yaml
- tapOn:
    id: "SendFormScreen.amountField"
    optional: true
- inputText: "0.0001"
```

Expected: amount field receives focus, keyboard opens, "0.0001" appears in the field.
Actual (v1.0.99): tap is a no-op, keyboard opens from a different element, input lands nowhere, display stays `0 ETH`.

The amount input is now only reachable via point coordinates (`bounds="[459,1205][603,1349]"` on Pixel 9 Pro — roughly center of the big centered amount display). Point tapping by percentage doesn't generalize across screen sizes.

## Impact

- 11 automated send flows silently broke in the v1.0.99 release QA pass. None of them would have caught a broadcast regression because they never got past the form.
- Every third-party QA suite that depends on this testTag is also affected.

## Fix request

Restore `.testTag("SendFormScreen.amountField")` on the `AmountField` composable inside the new `SendFormScreen` structure (or provide an equivalent stable accessibility ID). Same pattern that `chainSelector` and `addressField` already use.

Also suggest adding a regression test (or test-only release check) that asserts the presence of all four `SendFormScreen.*` testTags so a future refactor can't silently remove them.

## Evidence

- v1.0.99 QA report: `reports/v1.0.99-android-transaction/TRANSACTION-REPORT.md` — Section 1 documents the mid-run patch we had to apply.
- Screenshots: `reports/v1.0.99-android-transaction/` has before/after amount-field state.
- junit: `reports/v1.0.99-android-transaction/maestro-junit.xml` shows 11 simultaneous failures before the patch.

## Workaround (tests already patched)

Flow templates in `android/release/` updated to `point: "42%,45%"`. Fragile across screen sizes — worth reverting once the testTag is restored.

aminsato:
## Summary

- Restores the stable Compose selector `SendFormScreen.amountField` so Maestro QA flows can tap the amount input again. PR #4007 / #4108 split refactor moved the testTag onto the outer `Column` wrapping `TokenAmountInput`; that node is not focusable, so `tapOn id:` was a silent no-op and every automated send flow stalled with amount `0`.
- Adds an `inputModifier` parameter to `TokenAmountInput` and routes the testTag onto the inner `BasicTextField`, matching the pre-refactor placement that QA depends on.
- Adds `SendFormScreenTestTagsTest`, a fast JVM unit test that scans `app/src/main/java/com/vultisig/wallet/ui/screens/send/` and fails the build if any of the four stable selectors (`chainSelector`, `addressField`, `amountField`, `bondAddressField`) disappears in a future refactor.

Fixes #4180.

## Test plan

- [x] `./gradlew :app:compileDebugKotlin`
- [x] `./gradlew :app:testDebugUnitTest --tests com.vultisig.wallet.ui.screens.send.SendFormScreenTestTagsTest`
- [ ] Run a Maestro send flow that uses `tapOn id: "SendFormScreen.amountField"` and confirm the amount input gains focus, the keyboard opens, and the entered text lands in the field.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal component parameter structure for improved flexibility.

* **Tests**
  * Added validation test to ensure UI test identifiers remain intact across the codebase.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/components/TokenAmountInput.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/TokenAmountInput.kt
@@ -36,6 +36,7 @@ internal fun TokenAmountInput(
     primaryLabel: String,
     secondaryText: String,
     modifier: Modifier = Modifier,
+    inputModifier: Modifier = Modifier,
     maxBalance: BigDecimal? = null,
     focusRequester: FocusRequester? = null,
     onLostFocus: (() -> Unit)? = null,
@@ -98,7 +99,8 @@ internal fun TokenAmountInput(
                                     }
                                 }
                             } else Modifier
-                        ),
+                        )
+                        .then(inputModifier),
                 decorator = { textField ->
                     if (primaryFieldState.text.isEmpty()) {
                         Text(
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/send/SendFormAmountSection.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/send/SendFormAmountSection.kt
@@ -176,10 +176,8 @@ internal fun FoldableAmountWidget(
                         focusManager.clearFocus()
                         onSend()
                     },
-                    modifier =
-                        Modifier.padding(horizontal = 54.dp)
-                            .align(Alignment.Center)
-                            .testTag("SendFormScreen.amountField"),
+                    modifier = Modifier.padding(horizontal = 54.dp).align(Alignment.Center),
+                    inputModifier = Modifier.testTag("SendFormScreen.amountField"),
                 )
 
                 if (!isCircleMode) {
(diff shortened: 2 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4287, opened 2026-04-26
fix: handle already-joined vault on QR scan instead of hanging

### LINKED ISSUE #4282: Joining QR for seed phrase vault into Secure Vault fails with DuplicateVaultException / generateKey error
## Summary

When joining a QR code for a **seed-phrase vault** flow into a **Secure Vault** on a second device, the joined (already-existing) device throws a `generateKey error` and the keygen aborts. The underlying cause is `DuplicateVaultException` raised from `SaveVaultUseCaseImpl` because the device already has the vault stored.

The UI then **gets stuck on the "Vault created successfully" screen with an infinite loading spinner** — the user has no way to proceed and no error is shown.

## Steps to reproduce

1. On Device A: have an existing Secure Vault (seed-phrase based).
2. Start a keygen / join flow that produces a QR.
3. On Device B (which already has the same vault saved): scan the QR.
4. Device B begins the join flow and eventually attempts to save the vault.

## Expected behavior

The app should detect that the vault already exists on the joined device and either:
- Skip the save and proceed gracefully (treat as already-joined), **or**
- Surface a clear, user-facing message instead of crashing the keygen with a generic `generateKey error`.

## Actual behavior

- `KeygenViewModel` reports `generateKey error`.
- The app **hangs on the "Vault created successfully" screen** with an infinite loading indicator (see screenshot below). The user is stuck and must force-close the app.
- `MediatorService` is stopped: `stop MediatorService: Mediator service stopped after create vault because vault has already been in the joined device`.
- Stack trace:

```
com.vultisig.wallet.data.usecases.DuplicateVaultException: Vault already exists
    at com.vultisig.wallet.data.usecases.SaveVaultUseCaseImpl.invoke(SaveVaultUseCase.kt:32)
    at com.vultisig.wallet.data.usecases.SaveVaultUseCaseImpl\$invoke\$1.invokeSuspend(Unknown Source:16)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
    ...
```

### Stuck screen

The "Vault created successfully" screen with an infinite spinner — UI never advances past this state.

## Likely cause

In `app/src/main/java/com/vultisig/wallet/data/usecases/SaveVaultUseCase.kt:30-33`, when `shouldOverrideVault = false` and a vault with the same `pubKeyECDSA` already exists, `DuplicateVaultException` is thrown unconditionally. In the join-existing-vault flow on a device that already holds the vault, this exception is propagated up, the keygen fails, and the success screen's post-save navigation never fires — leaving the loading spinner forever.

## Suggested direction

- Detect the duplicate-vault case at the join-flow caller and treat it as a no-op success (the vault is already stored, no need to re-save).
- Or introduce a `shouldOverrideVault`-equivalent path for the join flow when the existing vault matches.
- Ensure the success screen handles save failures so the UI does not hang on the loading state.
- Improve the error surfacing so users see a meaningful message instead of a stuck spinner.

## Environment

- Branch: `agent/4244-thorchainapi-kt--multiple-bugs--inconsis`
- Build: debug, emulator

aminsato:
## Summary

Fixes #4282 — joining a QR for an existing seed-phrase vault on a device that already has it threw `DuplicateVaultException` after keygen completed, leaving the *Vault created successfully* screen stuck on an infinite spinner.

- **Primary fix** — `JoinKeygenViewModel` (KEYGEN/KeyImport branch) now compares the QR's `hexChainCode` against existing vaults before doing anything. If a match is found, it skips mediator discovery + keygen entirely and routes to `Route.Home(openVaultId = existingVault.id)`. The non-blank guard prevents matching legacy vaults that have an empty stored chain code.
- **Defence in depth** — `KeygenViewModel.saveVault()` now catches `DuplicateVaultException` on the non-override path (only `KEYGEN`/`KeyImport` can hit it; reshare/migrate use `upsert`). On catch it looks up the existing vault by `pubKeyECDSA`, stops the mediator service, and routes to Home — so the success screen never hangs even if the early `hexChainCode` check ever misses (e.g. the QR carries an empty chain code, or another flow saved the vault concurrently).
[Screen_recording_20260426_050958.webm](https://github.com/user-attachments/assets/24d7646d-5709-41f4-8b0e-4818fc32e854)

## Why two layers

The early `hexChainCode` check is the user-friendly path: skip the whole join when we already know the vault is here. The save-side catch protects the cases the early check can miss — initiating-device flows, blank/legacy chain codes, races against another save — so we never regress to the original infinite spinner.

## Test plan

- [ ] Device A has Vault X (seed-phrase). Device B has the same Vault X stored. Start join keygen flow on B with A's QR — B should land on Home with Vault X selected, no spinner hang, no error toast.
- [ ] Normal first-time join on a clean device still completes keygen and reaches the backup flow.
- [ ] Reshare / Migrate flows are unaffected (still upsert; never throw `DuplicateVaultException`).
- [ ] Verify on emulator with two debug builds (forward port 18080 + socat per README).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate vault creation when rejoining or importing; users are automatically redirected to their existing vault instead of creating duplicates.

* **New Features**
  * UI now detects when a vault is already on the device, showing a localized message with the vault name and a primary action to open the existing vault (navigates to Home and clears back stack).

* **Localization**
  * Added localized strings for the new "already on device" message and "open existing vault" action across multiple languages.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keygen/JoinKeygenViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keygen/JoinKeygenViewModel.kt
@@ -79,7 +79,10 @@ internal sealed class JoinKeygenError(val message: UiText) {
 internal data class JoinKeygenUiModel(
     val isSuccess: Boolean = false,
     val error: JoinKeygenError? = null,
+    val alreadyJoined: AlreadyJoinedVault? = null,
 )
+
+internal data class AlreadyJoinedVault(val vaultId: VaultId, val vaultName: String)
 
 private class JoinKeygenException(val error: JoinKeygenError) : Exception()
 
@@ -128,17 +131,22 @@ constructor(
 
                             // hexChainCode uniquely identifies the vault across devices.
                             // If a vault with the same chain code is already on this device,
-                            // skip the join and open it instead of running keygen + failing
-                            // on duplicate save.
+                            // skip the join and surface a clear "already on device" message
+                            // instead of running keygen + failing on duplicate save.
                             val alreadyJoinedVault = existingVaults.find {
                                 it.hexChainCode == message.hexChainCode &&
                                     it.hexChainCode.isNotBlank()
                             }
                             if (alreadyJoinedVault != null) {
-                                navigator.route(
-                                    route = Route.Home(openVaultId = alreadyJoinedVault.id),
-                                    opts = NavigationOptions(clearBackStack = true),
-                                )
+                                state.update {
+                                    it.copy(
+                                        alreadyJoined =
+                                            AlreadyJoinedVault(
+                                                vaultId = alreadyJoinedVault.id,
+                                                vaultName = alreadyJoinedVault.name,
+                                            )
+                                    )
+                                }
                                 return@launch
                             }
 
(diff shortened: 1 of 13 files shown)


aminsato:
@codetabbitai full review

johnnyluo (approved):
LGTM

PR #4288, opened 2026-04-26
test: stabilize flaky QBTC token re-add test

aminsato:
## Summary

- Fix intermittent CI failure: `ImportFileViewModelTest > restoring an MLDSA-capable vault re-adds the QBTC token`
- The test verified `addTokenToVault` eagerly, but `ImportFileViewModel.saveToDb` wraps `parseVaultFromString` in `withContext(Dispatchers.Default)` — a real thread pool that is **not** controlled by `runTest`'s `UnconfinedTestDispatcher`. Under CI load this raced and the verification ran before the QBTC token was added, producing a bare `AssertionError` at the test method line.
- Switch to `coVerify(timeout = 1_000)` so mockk polls until the call lands. The same race exists in principle for the sibling tests, but they verify state set inside `insertVaultToDb` directly and have not flaked — leaving them alone.

## Failures observed

The same test failed in three recent merge_group runs and intermittently passed otherwise:

- https://github.com/vultisig/vultisig-android/actions/runs/24943166541/job/73040093200
- https://github.com/vultisig/vultisig-android/actions/runs/24944726015
- https://github.com/vultisig/vultisig-android/actions/runs/24945042369

```
ImportFileViewModelTest > restoring an MLDSA-capable vault re-adds the QBTC token() FAILED
    java.lang.AssertionError at ImportFileViewModelTest.kt:148
```

## Test plan

- [x] Ran the affected test 5 consecutive times locally with `--rerun-tasks` — all green
- [x] Full `:app:testDebugUnitTest` for the class passes
- [ ] CI green

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability for vault restoration scenarios with controlled coroutine scheduling to make asynchronous operations more deterministic.
* **Refactor**
  * Switched to an injected coroutine execution strategy to improve async behavior consistency during file import and restore operations.

### DIFF
--- a/app/src/test/java/com/vultisig/wallet/ui/models/ImportFileViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/ImportFileViewModelTest.kt
@@ -41,8 +41,6 @@ import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 
-private const val VERIFY_TIMEOUT_MS = 1_000L
-
 @OptIn(ExperimentalCoroutinesApi::class)
 internal class ImportFileViewModelTest {
 
@@ -98,6 +96,7 @@ internal class ImportFileViewModelTest {
                 vaultRepository = vaultRepository,
                 chainAccountAddressRepository = chainAccountAddressRepository,
                 snackBarFlow = snackbarFlow,
+                defaultDispatcher = testDispatcher,
             )
         vm.uiModel.value =
             ImportFileState(fileName = fileName, fileContent = fileContent, isZip = false)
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/vultisig/wallet/ui/models/ImportFileViewModelTest.kt
```diff
@@ -164,7 +166,11 @@ internal class ImportFileViewModelTest {
         createViewModel(fileName = "share1of2-test.bak").decryptVaultData()
 
         val tokenSlot = slot<Coin>()
-        coVerify { vaultRepository.addTokenToVault("test-vault-id", capture(tokenSlot)) }
+        // saveToDb hops onto Dispatchers.Default to parse the vault, which escapes the
+        // test scheduler — poll until the verification succeeds instead of asserting eagerly.
+        coVerify(timeout = VERIFY_TIMEOUT_MS) {
```
rkokhatskyi: Nice find @aminsato 
A better approach would be to inject a `@DefaultDispatcher` into `ImportFileViewModel` so tests pass a `TestDispatcher` and the race goes away without polling

aminsato: Good call — switched to injecting `@DefaultDispatcher` and dropped the polling timeout. Tests now pass the `UnconfinedTestDispatcher` so the race is gone at the source. Pushed as a438cacc3.

johnnyluo (approved):
LGTM

PR #4300, opened 2026-04-27
fix: crash IllegalArgumentException in LazyList SubcomposeLayout on Android 16 (#4066)

### LINKED ISSUE #4066: Crash: IllegalArgumentException in LazyList SubcomposeLayout on Android 16
# [Fix] IllegalArgumentException in LazyList SubcomposeLayout during measurement [LazyColumn/LazyRow screens]

## Problem
A crash is reported in Android Vitals where `IllegalArgumentException` is thrown inside `SubcomposeLayout` during `LazyList` measurement. The exception occurs when a `LazyColumn` or `LazyRow` attempts to subcompose an item using an invalid or stale slot ID, likely caused by a race condition between recomposition and measurement passes or by list data changing mid-measurement. The issue was observed on Android 16 (API 36), which may exacerbate the problem due to rendering pipeline changes.

## Expected Behavior
Scrollable lazy lists should render and measure items without crashing, even when the underlying data source changes frequently or the user scrolls rapidly.

## Steps to Reproduce
1. Navigate to a screen containing a `LazyColumn` or `LazyRow` backed by a dynamic/frequently-updated data source.
2. Trigger a data update (items added or removed) while the list is actively scrolling or being laid out.
3. Observe `IllegalArgumentException` thrown from `SubcomposeLayout` measurement.

Crash confirmed on Samsung Galaxy A16 5G, Android 16.0 (API 36), around Apr 8, 2026.

### Must Do
- [ ] Audit all `LazyColumn`/`LazyRow` usages to ensure `key` lambdas return stable, unique, non-null keys for every item
- [ ] Ensure list data source updates are emitted atomically (replace the full list in a single emission rather than applying incremental mutations)
- [ ] Upgrade Compose BOM to the latest stable version that includes upstream fixes for this `SubcomposeLayout` race condition

### Must NOT Do
- Don't change unrelated UI or business logic outside of lazy list data flows
- Don't introduce item key changes that would break scroll position restoration

## Acceptance Criteria
- [ ] `./gradlew` succeeds
- [ ] No `IllegalArgumentException` from `SubcomposeLayout` reproduced under rapid data updates on affected screens
- [ ] All `LazyColumn`/`LazyRow` item `key` lambdas produce stable, unique values verified by code review

## Context
- **Device**: Samsung Galaxy A16 5G
- **OS**: Android 16.0 (API 36)
- **Crash date**: Apr 8, 2026
- **Versions affected**: 98
- **Source**: Android Vitals
- Upstream reference: [Compose issue tracker — SubcomposeLayout IllegalArgumentException](https://issuetracker.google.com/issues/252430462)
- Stack trace entry point: `LazyListMeasure.kt:224` → `LazyLayoutMeasuredItemProvider.getPlaceables` → `LayoutNodeSubcompositionsState.subcompose`; full trace continues through `Column`, `Box`, `Scaffold` down to `ViewRootImpl.performTraversals`

johnnyluo:
## Summary

Fixes the `IllegalArgumentException` in `SubcomposeLayout` that crashes list screens on Android 16 (API 36). Three root causes addressed:

- **Compose BOM bump** (`2025.09.00` → `2026.04.01`): picks up upstream `SubcomposeLayout` race-condition fixes shipped in newer Compose releases
- **Stable `key` lambdas** added to all 14 keyless `LazyColumn` / `LazyRow` / `LazyVerticalGrid` call sites: Compose was defaulting to positional index as the slot ID; when the list mutated mid-measurement, stale indices caused `IllegalArgumentException`
- **Atomic state updates** in `TransactionHistoryViewModel`: isolated `selectedAssetIds` and `selectedAssetsList` flows removed; all filter mutations now go through a single `_uiState.update {}`, eliminating the race window where `observeTransactions` could read new ids before `_uiState.selectedAssets` was updated

## Affected files

**BOM / dependencies:**
- `gradle/libs.versions.toml` — `composeBom` bumped to `2026.04.01`; removed `material-icons-extended` entry (~30 MB)
- `app/build.gradle.kts` — removed `material-icons-extended` dependency; replaced the three icons it provided (`Check`, `Warning`, `KeyboardArrowRight`) with existing drawables via `painterResource` in `SecurityScannerBadget.kt` and `ReferralViewScreen.kt`

**Key lambdas added / fixed (14 call sites):**
- `SelectAssetScreen.kt` — `key = { _, item -> item.token.id }`
- `ChainTokensScreen.kt` — `key = { _, token -> token.id }`
- `ReceiveBottomSheet.kt` — `key = { _, chain -> "${chain.name}:${chain.address}" }` (composite key — address alone is not unique across chains)
- `AddressBookBottomSheet.kt` — `key = { it.model.id }`
- `FastSelectionModalContent.kt` — `key = { index, _ -> index }` (index-stable for fixed-position scroll wheel)
- `SelectNetworkScreen.kt` — `key = { _, item -> item.chain.id }`
- `ReviewVaultDevicesScreen.kt` — `key = { _, device -> device }`
- `CurrencyUnitSettingScreen.kt` — `key = { _, currencyUnit -> currencyUnit.fullName }`
- `LanguageSettingScreen.kt` — `key = { _, language -> language.mainName }`
- `FAQSettingScreen.kt` — `key = { _, faq -> faq.id }` (stable `Int` resource id added to `Faq` data class; was using localized question text which is fragile)
- `ImportFileScreen.kt` — `key = { index, zipOutput -> "$index:${zipOutput.name}" }` (composite — same short name can appear at different zip paths)
- `CreateFolderScreen.kt` — `key = { _, vaultPair -> vaultPair.first.vault.id }` + hoist `toList()` into `remember()`
- `ReferralVaultListScreen.kt` — refactored from `items(state.vaults.size)` to `itemsIndexed(state.vaults, key = { _, vault -> vault.id })`
- `TokenSelectionList.kt` — switched from `hashCode()`-based key (collision-prone) to `itemsIndexed` with `"$groupIndex-$itemIndex"`

**ViewModel / state fixes:**
- `ReviewVaultDevicesViewModel.kt` — deduplicate devices with `.distinct()` into `uniqueDevices`; use `uniqueDevices.size` for `deviceCount` so the routed count matches the displayed list
- `TransactionHistoryViewModel.kt` — removed dead `selectedAssetsList` StateFlow; all of `toggleAssetSelection` / `removeAssetFilter` / `clearAllFilters` / `closeSearch` now emit in a single `_uiState.update {}`; `observeTransactions` combines on `_uiState.map { it.selectedAssetIds }.distinctUntilChanged()` instead of the removed standalone flow

**Upstream bug fix:**
- `KeygenPeerDiscoveryViewModel.kt` — apply `.distinct()` to `keygenCommittee` (was: `listOf(localPartyId) + selectedDevices`); prevents `localPartyId` duplicating when it is also auto-selected into `selectedDevices`

**Lint / composable API fixes (BOM upgrade surfaced stricter `LocalContextGetResourceValueCall` rule):**
- `UiText.kt` — replaced `context.getString()` / `context.resources.getQuantityString()` with `stringResource()` / `pluralStringResource()` in `@Composable fun asString()`
- `ScanQrScreen.kt` — hoisted `no_barcodes_found` to composable scope via `stringResource`; replaced localized debug log with direct Timber format spec
- `SwapTransactionOverviewScreen.kt` — `LaunchedEffect` + remembered state pattern for the dynamic tx-hash snackbar message (replaces `context.getString` inside composable lambda)
- `ChainTokensScreen.kt` — `stringResource(R.string.address_copied, chainName)` captured at composable scope (idiomatic, lint-clean)

**Documentation / cleanup:**
- Deleted accidentally committed empty `index` file
- `ReviewVaultDevicesScreen.kt` — comment documenting that key uniqueness is guaranteed by `ReviewVaultDevicesViewModel.uniqueDevices`

Closes #4066

## Test plan
- [ ] `./gradlew assembleDebug` passes
- [ ] `./gradlew lintDebug` passes
- [ ] Scroll through `SelectAssetScreen`, `ChainTokensScreen`, `AddressBookBottomSheet` — no scroll-position regressions
- [ ] Toggle asset filters in `TransactionHistoryScreen` — selection state is consistent with filtered results
- [ ] Start a keygen session; verify the committee list has no duplicate localPartyId entry
- [ ] On Android 16 emulator (API 36): navigate to affected list screens and verify no `IllegalArgumentException` crash


app/src/main/java/com/vultisig/wallet/ui/components/v2/tokenitem/TokenSelectionList.kt
```diff
@@ -155,7 +155,7 @@ internal fun <T> TokenSelectionList(
 
                             plusUiModel?.let { item { GridPlus(model = it) } }
 
-                            items(items) { item ->
+                            items(items, key = { it }) { item ->
```
aminsato: this `items(...)` is invoked once per group inside the same `LazyVerticalGrid`, so all keys share one namespace. `key = { it }` uses the `GridTokenUiModel` data class itself — if two groups ever contain structurally-equal entries, the duplicate keys will crash the grid. Today's callers (`DeFiComponents`, etc.) build disjoint per-group lists so it isn't a real issue, but it's a bit fragile.

If you want to be defensive, swap the outer loop for `groups.forEachIndexed { groupIndex, (title, items, mapper, plusUiModel) -> ... }` and namespace the key, e.g. `key = { "$groupIndex-${(it as? SingleToken)?.data?.someStableId ?: it.hashCode()}" }` (or whatever stable id your model exposes). Happy to skip — just flagging.

index
rkokhatskyi, line 1: empty file accidentally committed in `1dcf0a439` should be deleted before merge

app/build.gradle.kts
```diff
@@ -90,6 +90,7 @@ dependencies {
     implementation(libs.androidx.navigation.compose)
     implementation(libs.androidx.material3)
     implementation(libs.androidx.material3.window)
+    implementation(libs.androidx.material.icons.extended)
```
rkokhatskyi: this dependency adds ~30MB of icon resources for only three icons actually used (`Icons.Default.Check` `Icons.Default.Warning` `Icons.AutoMirrored.Filled.KeyboardArrowRight`) all already shipped as drawables (`ic_check.xml` `ic_warning.xml` `ic_caret_right.xml`) replace with `painterResource` and drop the dependency
```kotlin
Icon(painter = painterResource(R.drawable.ic_check) ...)
```

app/src/main/java/com/vultisig/wallet/ui/components/v2/tokenitem/TokenSelectionList.kt
```diff
@@ -148,14 +148,14 @@ internal fun <T> TokenSelectionList(
...
-                        groups.forEach { (title, items, mapper, plusUiModel) ->
+                        groups.forEachIndexed { groupIndex, (title, items, mapper, plusUiModel) ->
                             title?.let {
                                 item(span = { GridItemSpan(maxLineSpan) }) { GridTitle(it) }
                             }
 
                             plusUiModel?.let { item { GridPlus(model = it) } }
 
-                            items(items) { item ->
+                            items(items, key = { "$groupIndex-${it.hashCode()}" }) { item ->
```
rkokhatskyi: also ISSUE — `it.hashCode()` is not guaranteed unique two `data class` instances with equal property values produce identical hashes and 32 bit hashCodes can also collide between unequal instances expose a stable id on `GridTokenUiModel` and key on it
```kotlin
sealed interface GridTokenUiModel<T> { val stableId: String }
items(items, key = { "$groupIndex-${it.stableId}" })
```

app/src/main/java/com/vultisig/wallet/ui/models/v3/ReviewVaultDevicesViewModel.kt
```diff
@@ -32,11 +32,12 @@ internal class ReviewVaultDevicesViewModel
 constructor(private val navigator: Navigator<Destination>, savedStateHandle: SavedStateHandle) :
     ViewModel() {
     val args = savedStateHandle.toRoute<Route.ReviewVaultDevices>()
+    private val uniqueDevices = args.devices.orEmpty().distinct()
```
rkokhatskyi, line 35: the dedup masks an upstream bug `KeygenPeerDiscoveryViewModel.kt:296` builds `keygenCommittee = listOf(localPartyId) + state.value.selectedDevices` and `localPartyId` ends up in both halves once auto-selected at line 518 fix the source
```kotlin
val keygenCommittee = (listOf(localPartyId) + state.value.selectedDevices).distinct()
```
otherwise log a warning here when the input has duplicates so committee corruption is not silently swallowed

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -171,30 +171,26 @@ constructor(
...
-            it.copy(
-                selectedAssetIds = selectedAssetIds.value,
-                selectedAssets = selectedAssetsList.value,
-            )
+            newIds = selectedAssetIds.value + asset.tokenId
+            newList = selectedAssetsList.value + asset
         }
+        selectedAssetIds.value = newIds
+        selectedAssetsList.value = newList
+        _uiState.update { it.copy(selectedAssetIds = newIds, selectedAssets = newList) }
```
rkokhatskyi: PR description claims atomic state but this is three separate writes (`selectedAssetIds.value = newIds` `selectedAssetsList.value = newList` `_uiState.update {}`) `observeTransactions` at line 272 combines on `selectedAssetIds` directly so a coroutine can read the new ids while `_uiState.selectedAssets` is still stale
```kotlin
selectedAssetIds.update { newIds }
selectedAssetsList.update { newList }
_uiState.update { it.copy(selectedAssetIds = newIds, selectedAssets = newList) }
```
better still collapse the three flows into one source of truth

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -171,30 +171,26 @@ constructor(
...
-        _uiState.update {
-            it.copy(
-                selectedAssetIds = selectedAssetIds.value,
-                selectedAssets = selectedAssetsList.value,
-            )
+            newIds = selectedAssetIds.value + asset.tokenId
+            newList = selectedAssetsList.value + asset
         }
+        selectedAssetIds.value = newIds
+        selectedAssetsList.value = newList
```
rkokhatskyi: `selectedAssetsList` is dead state nothing subscribes to this flow it only mirrors `_uiState.selectedAssets` drop the standalone `MutableStateFlow` at line 151 and read from `_uiState.value.selectedAssets` where needed

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -171,30 +171,26 @@ constructor(
...
-            )
-        }
+        val newIds = selectedAssetIds.value - assetId
+        val newList = selectedAssetsList.value.filter { a -> a.tokenId != assetId }
+        selectedAssetIds.value = newIds
+        selectedAssetsList.value = newList
+        _uiState.update { it.copy(selectedAssetIds = newIds, selectedAssets = newList) }
     }
 
     fun clearAllFilters() {
```
rkokhatskyi, line 192: `clearAllFilters` uses `.update {}` while `toggleAssetSelection` and `removeAssetFilter` (just above) assign `.value` directly pick one approach so reviewers do not need to track which method is atomic and which is not

app/src/main/java/com/vultisig/wallet/ui/screens/v2/chaintokens/ChainTokensScreen.kt
```diff
@@ -110,7 +111,7 @@ internal fun ChainTokensScreen(
 ) {
     val snackbarState = rememberVsSnackbarState()
     val uriHandler = VsUriHandler()
-    val context = LocalContext.current
+    val addressCopiedMessage = stringResource(R.string.address_copied, uiModel.chainName)
```
rkokhatskyi, line 114: now evaluated on every recomposition of the entire screen previously the lookup was lazy inside the `onAddressCopied` lambda wrap with `remember(uiModel.chainName) { ... }` (still needs context for `getString`) or move the call back inside the lambda this `stringResource` only feeds the snackbar so eager evaluation buys nothing

app/src/main/java/com/vultisig/wallet/ui/screens/ImportFileScreen.kt
```diff
@@ -203,7 +203,7 @@ private fun ImportFileScreen(
 @Composable
 private fun ZipOutput(zipOutputs: List<AppZipEntry>, onImportVult: (AppZipEntry) -> Unit) {
     LazyColumn {
-        itemsIndexed(zipOutputs) { index, zipOutput ->
+        itemsIndexed(zipOutputs, key = { _, zipOutput -> zipOutput.name }) { index, zipOutput ->
```
rkokhatskyi: a zip can contain entries with identical short names in different paths (`folder/foo.vult` and `foo.vult`) which collide here use the full entry path or a (path:hash) composite as the key

app/src/main/java/com/vultisig/wallet/ui/screens/settings/FAQSettingScreen.kt
```diff
@@ -56,7 +56,7 @@ private fun FaqSettingScreen(onBackClick: () -> Unit, state: FAQSettingUiModel)
     ) {
         V2Container(type = ContainerType.SECONDARY) {
             LazyColumn(contentPadding = PaddingValues(horizontal = 20.dp, vertical = 16.dp)) {
-                itemsIndexed(state.questions) { index, faq ->
+                itemsIndexed(state.questions, key = { _, faq -> faq.question }) { index, faq ->
```
rkokhatskyi: using the localized question text as a key is fragile any whitespace difference duplicate entry or accidental localization tweak crashes the screen add a stable id field on the FAQ data class and key on it

app/src/main/java/com/vultisig/wallet/ui/screens/v2/receive/ReceiveBottomSheet.kt
```diff
@@ -72,7 +72,10 @@ private fun ReceiveContent(
 
         V2Container(type = ContainerType.SECONDARY) {
             LazyColumn(modifier = Modifier.padding()) {
-                itemsIndexed(uiState.chains) { index, chain ->
+                itemsIndexed(
+                    uiState.chains,
+                    key = { _, chain -> "${chain.name}:${chain.address}" },
```
rkokhatskyi: `ChainToReceiveUiModel` lacks a chain identifier so this concatenates display strings instead of an id add a `chain: Chain` field on the model and key on `chain.id` the current key works only because the same chain never repeats and that invariant lives in the view model not in this screen

app/src/main/java/com/vultisig/wallet/ui/screens/v3/onboarding/ReviewVaultDevicesScreen.kt
```diff
@@ -106,7 +106,7 @@ private fun ReviewVaultDevicesScreen(
             UiSpacer(size = 32.dp)
 
             LazyColumn(verticalArrangement = Arrangement.spacedBy(12.dp)) {
-                itemsIndexed(uiState.devices) { index, device ->
+                itemsIndexed(uiState.devices, key = { _, device -> device }) { index, device ->
```
rkokhatskyi, line 111: key uniqueness here depends entirely on the `args.devices.orEmpty().distinct()` in the view model if any future caller bypasses that path and feeds raw devices the screen crashes either hoist the dedup into a derived state on the model or document the dependency at the call site

gradle/libs.versions.toml
```diff
@@ -31,7 +31,7 @@ kotlinxProtobufGen = "0.1.0"
 ktor = "3.0.3"
 lifecycleRuntimeKtx = "2.8.7"
 activityCompose = "1.10.1"
-composeBom = "2025.09.00"
+composeBom = "2026.04.01"
```
rkokhatskyi, line 34: PR description still claims `chain.address` as the ReceiveBottomSheet key `item ?: "null_$index"` for FastSelectionModalContent and `key = { it }` for TokenSelectionList all three were already changed during review update the description before merging so future readers can trust it

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -171,30 +171,26 @@ constructor(
     }
 
     fun toggleAssetSelection(asset: TransactionAssetUiModel) {
```
rkokhatskyi, line 171: no `TransactionHistoryViewModelTest` covers `toggleAssetSelection` `removeAssetFilter` or `clearAllFilters` and no `ReviewVaultDevicesViewModelTest` covers the `.distinct()` dedup the project mandate is 80% coverage and the rewrite of the three mutators here is exactly the kind of subtle change that needs Turbine based regression tests

rkokhatskyi (changes requested):

rkokhatskyi:
@Vaulty-bot fix CI errors and also address all the comments raised in threads

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
johnnyluo: will address it later , not in this PR

PR #4310, opened 2026-04-28
feat(evm): static common selector table for offline 4byte decode (#4055)

### LINKED ISSUE #4055: feat: Static common EVM selector table for offline decode (Phase 1.3)
## Context
Currently relies on 4byte API for function name resolution. A static table of ~25 common selectors provides instant offline decoding and human-readable action labels ("Token Approval" instead of raw `approve(address,uint256)`).

## What to build
Static lookup table in `EvmCommonSelectors.kt` — check before 4byte API call.

Key selectors: `approve`, `transfer`, `transferFrom`, Uniswap V2 swaps, `deposit`/`withdraw` (WETH), `stake`/`unstake`, `safeTransferFrom` (ERC721/1155).

**Companion issue:** vultisig/vultisig-windows#3734
**PRD:** section 1.3

aminsato:
## Summary
- Adds `EvmCommonSelectors`, an in-process lookup of ~33 stable 4-byte function selectors → canonical text signatures (ERC-20, WETH, Uniswap V2/V3, Aave V3, ERC-721/1155, generic stake/unstake, multicall).
- `FourByteRepository.decodeFunction` consults the table first and only falls back to the 4byte directory on a miss — common dApp calls now decode instantly and offline.
- No change to the public contract: the table returns the same canonical signature shape (`approve(address,uint256)`) the 4byte API would.

Closes #4055

## Selection criteria
- Cross-checked the iOS `String+ExtensionMemo.knownFunctionSelector` switch.
- **Skipped** read-only views (`balanceOf`/`totalSupply`/`allowance`) — never sent as transactions.
- **Skipped** aggregator-version-specific entries (`KyberSwap e21fd0e9`, `1inch v4 7c025200`) — their swap signatures churn across versions and risk decoding to a stale shape.

## Test plan
- [x] `./gradlew :data:testDebugUnitTest` (new `EvmCommonSelectorsTest` and `FourByteRepositoryImplTest` pass)
- [x] `FourByteRepositoryImplTest` verifies `fourByteApi.decodeFunction` is **never invoked** for selectors in the static table (`coVerify(exactly = 0)`)
- [x] `ktfmt` applied

## Summary by CodeRabbit

* **New Features**
  * Built-in offline lookup for many common EVM function selectors covering tokens, NFTs, DEXs, lending, staking, multicall and permit variants.

* **Improvements**
  * Validates and normalizes selector input (case-insensitive) and uses the local mapping before querying external decoding; cancellation exceptions propagate, other errors return null.

* **Tests**
  * Comprehensive unit tests for offline resolution, API fallback, input edge cases, table integrity, and error/cancellation behavior.


data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,64 @@
+package com.vultisig.wallet.data.repositories
+
+/**
+ * Static lookup of common EVM 4-byte function selectors → canonical text signatures.
```
rkokhatskyi, line 6: The KDoc says common dApp calls decode offline, but Universal Router selector `3593564c` `execute(bytes,bytes[],uint256)` is what current Uniswap UI swaps hit and it is absent here, so the marquee dApp swap path still hits the network.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    private val table: Map<String, String> =
+        mapOf(
+            // ERC-20
+            "a9059cbb" to "transfer(address,uint256)",
+            "23b872dd" to "transferFrom(address,address,uint256)",
+            "095ea7b3" to "approve(address,uint256)",
+            "39509351" to "increaseAllowance(address,uint256)",
+            "a457c2d7" to "decreaseAllowance(address,uint256)",
+            "d505accf" to "permit(address,address,uint256,uint256,uint8,bytes32,bytes32)",
```
rkokhatskyi, line 24: DAI uses a different `permit` shape, selector `8fcbaf0c` `permit(address,address,uint256,uint256,bool,uint8,bytes32,bytes32)`, so DAI permits still round trip to 4byte.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,64 @@
...
+            "4a25d94a" to "swapTokensForExactETH(uint256,uint256,address[],address,uint256)",
+            "18cbafe5" to "swapExactTokensForETH(uint256,uint256,address[],address,uint256)",
+            "fb3bdb41" to "swapETHForExactTokens(uint256,address[],address,uint256)",
+            "5c11d795" to
+                "swapExactTokensForTokensSupportingFeeOnTransferTokens(uint256,uint256,address[],address,uint256)",
+            "b6f9de95" to
+                "swapExactETHForTokensSupportingFeeOnTransferTokens(uint256,address[],address,uint256)",
+            "791ac947" to
+                "swapExactTokensForETHSupportingFeeOnTransferTokens(uint256,uint256,address[],address,uint256)",
+            // Uniswap V3 SwapRouter (V01)
```
rkokhatskyi: V01 is the legacy Uniswap V3 router. `SwapRouter02` is what most chains use today and its selectors `04e45aaf` `b858183f` `5023b4df` `09b81346` differ because the `deadline` parameter was dropped, so without them V02 swaps still hit the network.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,64 @@
...
+            "414bf389" to
+                "exactInputSingle((address,address,uint24,address,uint256,uint256,uint256,uint160))",
+            "c04b8d59" to "exactInput((bytes,address,uint256,uint256,uint256))",
+            "db3e2198" to
+                "exactOutputSingle((address,address,uint24,address,uint256,uint256,uint256,uint160))",
+            "f28c0498" to "exactOutput((bytes,address,uint256,uint256,uint256))",
+            "ac9650d8" to "multicall(bytes[])",
```
rkokhatskyi, line 65: `multicall(bytes[])` is a generic compose pattern used by Aave V3, ENS, and many other contracts, so grouping it under the Uniswap V3 SwapRouter header is misleading.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,64 @@
...
+            // Aave V3 / Spark / Radiant pool
```
rkokhatskyi: The comment reads as if all four entries are V3 specific, but `withdraw` `borrow` `repay` share their signatures with Aave V2. Only `supply` is the V3 rename of V2 `deposit`, selector `e8eda9df`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,64 @@
...
+            "617ba037" to "supply(address,uint256,address,uint16)",
+            "69328dec" to "withdraw(address,uint256,address)",
+            "a415bcad" to "borrow(address,uint256,uint256,uint16,address)",
+            "573ade81" to "repay(address,uint256,uint256,address)",
+            // ERC-721
```
rkokhatskyi: `a22cb465` `setApprovalForAll(address,bool)` is part of EIP 1155 too, so grouping these three lines under ERC 721 alone is misleading.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,64 @@
...
+            "42842e0e" to "safeTransferFrom(address,address,uint256)",
+            "b88d4fde" to "safeTransferFrom(address,address,uint256,bytes)",
+            "a22cb465" to "setApprovalForAll(address,bool)",
+            // ERC-1155
+            "f242432a" to "safeTransferFrom(address,address,uint256,uint256,bytes)",
+            "2eb2c2d6" to "safeBatchTransferFrom(address,address,uint256[],uint256[],bytes)",
+            // Generic staking
```
rkokhatskyi: `stake(uint256)` and `unstake(uint256)` are not standardized. Many farms expose `deposit(uint256)`, `enter(uint256)`, or `mint(uint256)` instead, so the Generic staking label oversells coverage.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/FourByteRepository.kt
```diff
@@ -24,8 +24,9 @@ constructor(private val fourByteApi: FourByteApi, @param:PrettyJson private val
         if (memo.length < 8) return null
```
rkokhatskyi: `memo.length < 8` ignores the `0x` prefix, so `0xabcdef` of length 8 slips through and then `substring(0, 8)` throws on the 6 char stripped string and the catch silently swallows it. Try:
```kotlin
val hash = memo.stripHexPrefix()
if (hash.length < 8) return null
val selector = hash.substring(0, 8)
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/FourByteRepository.kt
```diff
@@ -24,8 +24,9 @@ constructor(private val fourByteApi: FourByteApi, @param:PrettyJson private val
...
         try {
             val hash = memo.stripHexPrefix().substring(0, 8)
-            return fourByteApi.decodeFunction(hash)
-        } catch (_: Exception) {
+            return EvmCommonSelectors.lookup(hash) ?: fourByteApi.decodeFunction(hash)
+        } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
rkokhatskyi: Also, the `if (e is X) throw e` form is not the codebase pattern. `SafeLaunch.kt`, `ThorChainRepository.kt`, and most data layer repositories use a separate catch and the fully qualified name moves into an import:
```kotlin
import kotlinx.coroutines.CancellationException

// ...

    throw e
} catch (_: Exception) {
    return null
}
```

data/src/test/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectorsTest.kt
```diff
@@ -0,0 +1,111 @@
+package com.vultisig.wallet.data.repositories
+
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+import org.junit.jupiter.api.Test
+
+internal class EvmCommonSelectorsTest {
```
rkokhatskyi, line 9: These tests cover roughly half the table. `transferFrom`, `increaseAllowance`, `decreaseAllowance`, `permit`, `swapTokensForExactTokens`, `swapTokensForExactETH`, `swapExactTokensForETH`, `swapETHForExactTokens`, `exactOutputSingle`, `exactOutput`, Aave `withdraw`, `setApprovalForAll`, `stake`, and `unstake` get no direct assertion. A single test that walks the table and asserts each key is eight lowercase hex chars and each value parses as `name(types)` would catch typos in entries no one tests by hand.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/FourByteRepositoryImplTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import io.mockk.coVerify
+import io.mockk.mockk
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import kotlinx.serialization.json.Json
+import org.junit.jupiter.api.Test
+
+internal class FourByteRepositoryImplTest {
```
rkokhatskyi, line 18: This file does not exercise the two new behaviors the PR added: the `CancellationException` rethrow path and the case where the API throws an exception other than cancellation. Both should have a case here.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/FourByteRepositoryImplTest.kt
```diff
@@ -0,0 +1,48 @@
...
+
+        val signature = repository.decodeFunction(memo)
+
+        assertEquals("someUnknownFunc()", signature)
+        coVerify(exactly = 1) { fourByteApi.decodeFunction("deadbeef") }
+    }
+
+    @Test
+    fun `decodeFunction returns null for memo shorter than 4 bytes`() = runTest {
+        assertNull(repository.decodeFunction("0x12"))
```
rkokhatskyi, line 48: This case passes `0x12` of length 4, but the boundary the code checks is 8 chars, so it never exercises the lower edge of the guard. A memo of `0xabcdef` of length 8 with `0x` prefix slips past the length check and triggers the silent `substring` catch.

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
aminsato: Added — `3593564c execute(bytes,bytes[],uint256)` is in commit 2d0ab2a92.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
aminsato: Added DAI permit `8fcbaf0c permit(address,address,uint256,uint256,bool,uint8,bytes32,bytes32)` next to the EIP-2612 entry, with comments distinguishing the two shapes.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
aminsato: Added all four SwapRouter02 selectors (`04e45aaf`, `b858183f`, `5023b4df`, `09b81346`) under a separate V02 section, with the deadline-less tuple shapes.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
aminsato: Pulled `multicall(bytes[])` out of the V3 SwapRouter group into its own line with a comment noting it's a generic compose pattern used by Uniswap V3, Aave V3, ENS, and many others.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
aminsato: Updated the comment to call out that `withdraw` / `borrow` / `repay` share signatures with Aave V2, and only `supply` is the V3 rename of V2 `deposit(address,uint256,address,uint16)`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
aminsato: Merged the ERC-721 and ERC-1155 sections into one block and added a comment that `setApprovalForAll` is part of both standards.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
aminsato: Renamed the section header to "Common staking shapes" with a comment that they're not standardized — many farms use `deposit` / `enter` / `mint` instead.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/FourByteRepository.kt
aminsato: Fixed — strip the prefix first, then check length against 8, then substring:
```kotlin
val hash = memo.stripHexPrefix()
if (hash.length < 8) return null
val selector = hash.substring(0, 8)
```
Added a regression test for `0xabcdef` (8 chars total, 6 after strip) that asserts `null` and `coVerify(exactly = 0)` for the API.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/FourByteRepository.kt
aminsato: Switched to the codebase pattern — imported `kotlinx.coroutines.CancellationException` and used a separate catch block:
```kotlin
    throw e
} catch (_: Exception) {
    null
}
```

data/src/test/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectorsTest.kt
aminsato: Added a single table-walk test that iterates `EvmCommonSelectors.entries` and asserts each key is 8 lowercase hex chars, each value matches `name(types)`, parens are balanced, and `lookup(key)` round-trips. Also added explicit tests for SwapRouter02, Universal Router, and both permit shapes.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/FourByteRepositoryImplTest.kt
aminsato: Added two new tests — one where the API throws `RuntimeException` and `decodeFunction` returns `null`, and one where the API throws `CancellationException` and `decodeFunction` rethrows it.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/FourByteRepositoryImplTest.kt
aminsato: Added the `0xabcdef` case as its own test (`returns null when 0x prefix leaves fewer than 8 hex chars`). It asserts `null` and `coVerify(exactly = 0) { fourByteApi.decodeFunction(any()) }` so the API is never called.

aminsato:
@coderabbitai full review

data/src/main/kotlin/com/vultisig/wallet/data/repositories/FourByteRepository.kt
```diff
@@ -21,12 +22,15 @@ internal class FourByteRepositoryImpl
...
     override suspend fun decodeFunction(memo: String): String? {
-        if (memo.length < 8) return null
-        try {
-            val hash = memo.stripHexPrefix().substring(0, 8)
-            return fourByteApi.decodeFunction(hash)
+        val hash = memo.stripHexPrefix()
+        if (hash.length < 8) return null
+        val selector = hash.substring(0, 8)
+        return try {
+            EvmCommonSelectors.lookup(selector) ?: fourByteApi.decodeFunction(selector)
```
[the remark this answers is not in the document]
aminsato: Addressed in commit e6401abef — added a `HEX_SELECTOR` regex guard on the 8-char selector before the API fallback, plus a `0xZZZZZZZZ` regression test asserting `null` and `coVerify(exactly = 0) { fourByteApi.decodeFunction(any()) }`.

aminsato:
Addressed CodeRabbit nitpick on `EvmCommonSelectorsTest.kt:170-173` in commit e6401abef — replaced `assertNotNull(lookup(selector), …)` with `assertEquals(sig, lookup(selector), …)` in the table-walk so a wrong-signature lookup is caught (not just a null one).

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,81 @@
...
+            "04e45aaf" to
+                "exactInputSingle((address,address,uint24,address,uint256,uint256,uint160))",
+            "b858183f" to "exactInput((bytes,address,uint256,uint256))",
+            "5023b4df" to
+                "exactOutputSingle((address,address,uint24,address,uint256,uint256,uint160))",
+            "09b81346" to "exactOutput((bytes,address,uint256,uint256))",
+            // Uniswap Universal Router (current Uniswap UI swap entry point)
+            "3593564c" to "execute(bytes,bytes[],uint256)",
+            // Generic compose pattern — used by Uniswap V3, Aave V3, ENS, and many other contracts
+            "ac9650d8" to "multicall(bytes[])",
```
rkokhatskyi, line 65: also `5ae401dc multicall(uint256,bytes[])` the deadline carrying variant is what current Uniswap V3 SwapRouter02 Universal Router and V4 PositionManager wrap most UI swaps in so the headline offline decode still misses the most common Uniswap UI swap path without it

data/src/test/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectorsTest.kt
```diff
@@ -0,0 +1,176 @@
...
+                "selector '$selector' must be 8 lowercase hex chars",
+            )
+            assertTrue(signature.matches(sig), "signature '$sig' must look like name(types)")
+            assertEquals(
+                sig.count { it == '(' },
+                sig.count { it == ')' },
+                "unbalanced parentheses in '$sig'",
+            )
+            assertEquals(
+                sig,
```
rkokhatskyi: This walks `entries` and asserts `lookup(selector) == sig` where both come from the same map row so a typo in either column passes silently. The only invariant that catches a wrong selector or a wrong signature is verifying `keccak256(sig).take(4) == selector`

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
aminsato: Added `5ae401dc multicall(uint256,bytes[])` in commit 4aefa1442 with a comment noting it's the deadline-carrying variant used by SwapRouter02, Universal Router, and V4 PositionManager.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectorsTest.kt
aminsato: Good catch — the same-row round-trip was tautological. Replaced it in commit 4aefa1442 with a real `keccak256(sig).take(4) == selector` check using the existing `toKeccak256ByteArray()` helper, so a typo in either the selector or the signature now fails the test. Verified that all 35 entries (including the new `5ae401dc`) pass.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,84 @@
...
+            "f28c0498" to "exactOutput((bytes,address,uint256,uint256,uint256))",
+            // Uniswap V3 SwapRouter02 (deadline dropped — used on most chains today)
+            "04e45aaf" to
+                "exactInputSingle((address,address,uint24,address,uint256,uint256,uint160))",
+            "b858183f" to "exactInput((bytes,address,uint256,uint256))",
+            "5023b4df" to
+                "exactOutputSingle((address,address,uint24,address,uint256,uint256,uint160))",
+            "09b81346" to "exactOutput((bytes,address,uint256,uint256))",
+            // Uniswap Universal Router (current Uniswap UI swap entry point)
+            "3593564c" to "execute(bytes,bytes[],uint256)",
```
rkokhatskyi, line 62: `24856bc3 execute(bytes,bytes[])` is the V2 Universal Router entry point that has been the Uniswap UI default since late 2024 so V2 swaps still round trip to 4byte without it

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,84 @@
...
+                "swapExactETHForTokensSupportingFeeOnTransferTokens(uint256,address[],address,uint256)",
+            "791ac947" to
+                "swapExactTokensForETHSupportingFeeOnTransferTokens(uint256,uint256,address[],address,uint256)",
+            // Uniswap V3 SwapRouter (V01, with deadline)
+            "414bf389" to
+                "exactInputSingle((address,address,uint24,address,uint256,uint256,uint256,uint160))",
+            "c04b8d59" to "exactInput((bytes,address,uint256,uint256,uint256))",
+            "db3e2198" to
+                "exactOutputSingle((address,address,uint24,address,uint256,uint256,uint256,uint160))",
+            "f28c0498" to "exactOutput((bytes,address,uint256,uint256,uint256))",
```
rkokhatskyi, line 49: `12210e8a refundETH()` rides inside `multicall` on every V3 swap paying in ETH that overpays so the inner call still falls through to 4byte even with both multicall variants offline

data/src/main/kotlin/com/vultisig/wallet/data/repositories/EvmCommonSelectors.kt
```diff
@@ -0,0 +1,84 @@
...
+            "f242432a" to "safeTransferFrom(address,address,uint256,uint256,bytes)",
+            "2eb2c2d6" to "safeBatchTransferFrom(address,address,uint256[],uint256[],bytes)",
+            // Common staking shapes (not standardized — many farms use deposit/enter/mint instead)
+            "a694fc3a" to "stake(uint256)",
+            "2e17de78" to "unstake(uint256)",
+        )
+
+    fun lookup(selector: String): String? = table[selector.lowercase()]
+
+    internal val entries: Map<String, String>
```
rkokhatskyi: `entries` is `internal` whereas the only consumer is `EvmCommonSelectorsTest`, so `@VisibleForTesting` or a `tableForTesting` rename would carry the intent at the call site

rkokhatskyi (changes requested):

aminsato:
@rkokhatskyi all your inline comments from the Apr 30 review are addressed (commits 2d0ab2a, e6401ab, a8daa0c, d387655, 4aefa14):

- Universal Router `3593564c`, DAI permit `8fcbaf0c`, SwapRouter02 selectors (`04e45aaf`/`b858183f`/`5023b4df`/`09b81346`) added.
- `multicall(bytes[])` lifted out of the Uniswap V3 group with a comment that it's a generic compose pattern.
- Aave V3 comment now explicitly notes only `supply` is the V3 rename — `withdraw`/`borrow`/`repay` share signatures with V2.
- ERC-721 / ERC-1155 sections merged with a note that `setApprovalForAll` belongs to both standards.
- Generic staking renamed to "Common staking shapes" with a comment that they're not standardized.
- `memo.length < 8` bug fixed — strip prefix first, then length check, then substring; `0xabcdef` (8 chars total, 6 after strip) and `0xZZZZZZZZ` (non-hex) are covered by regression tests.
- `catch (e: CancellationException) { throw e }` separated from `catch (_: Exception) { null }` per the SafeLaunch / data-layer pattern.
- Test coverage extended: table-walk now asserts `keccak256(sig).take(4) == selector` for every entry (not just round-trip lookup), plus explicit cases for `CancellationException` rethrow, non-cancellation API exception, short memo, prefix-stripped underflow, and non-hex selector — all with `coVerify(exactly = 0/1)` on the API to lock in the no-call vs reached-call expectations.

Could you take another look when you get a chance? Thanks!

rkokhatskyi:
Apr 30 round looks resolved, my May 4 review still has three open inline threads on `EvmCommonSelectors.kt` covering Universal Router V2 `24856bc3`, V3 `refundETH()` `12210e8a`, and the `entries` getter visibility

rkokhatskyi (changes requested):
@aminsato please ask you bot to check latest thread comments (mostly minor through, only one major that actually has to be resolved before merging)

aminsato:
@rkokhatskyi all three threads addressed in 257145b3f:

- Added `24856bc3 execute(bytes,bytes[])` (V2 Universal Router) and `12210e8a refundETH()` to `EvmCommonSelectors.kt` so V2 swaps and the multicall-wrapped ETH refund decode offline.
- Renamed the `entries` getter to `tableForTesting` and annotated it `@VisibleForTesting`, so the test-only intent is explicit at the call site.

Tests updated to assert the two new selectors, and the existing keccak256 round-trip walk over `tableForTesting` will catch any typo in the new entries. `./gradlew :data:testDebugUnitTest` passes locally and ktfmt has been applied.

rkokhatskyi (approved):
LGTM

PR #4319, opened 2026-04-29
[Fix] Swap: rank providers by net output (closes #4315)

### LINKED ISSUE #4315: Swap: same-chain ERC20→native uses deterministic rule, not quote ranking — defaults to THORChain ≥$100
## Context

iOS reported a same-chain Ethereum **USDC → ETH** swap routed through THORChain at **\$7.88 fees on \$10** (~79%). Investigation across platforms found Android has the same architectural flaw but a different observable behaviour: amount-threshold rule, no quote comparison.

## Root cause

**`data/src/main/kotlin/com/vultisig/wallet/data/repositories/SwapQuoteRepository.kt:511`** — provider eligibility:

```kotlin
override fun resolveProvider(srcToken: Coin, dstToken: Coin): SwapProvider? {
    return srcToken.swapProviders.intersect(dstToken.swapProviders).firstOrNull {
        if (isCrossChainSwap(srcToken, dstToken))
            it != SwapProvider.ONEINCH && it != SwapProvider.KYBER
        else true
    }
}
```

Returns the **first** provider from the intersection — no quote comparison.

**`app/src/main/java/com/vultisig/wallet/data/usecases/resolveprovider/KyberSwapProviderSelectionStrategyImpl.kt:38-45`** — the only override:

```kotlin
if (provider != SwapProvider.THORCHAIN || isErc20Swap(src = srcToken, dst = dstToken).not())
    return provider

val tokenValueInDollar = convertTokenValueToFiat(srcToken, value, AppCurrency.USD)

return if (tokenValueInDollar.value < BigDecimal.valueOf(AMOUNT_FOR_THORCHAIN_OR_KYBER))
    SwapProvider.KYBER
else provider
```

Same-chain ERC20 swaps under \$100 → KYBER. **At or above \$100 → THORCHAIN unconditionally**, regardless of whether 1inch / LI.FI would deliver more ETH net-of-fees.

## Why this is wrong

The \$100 threshold is a hand-tuned heuristic, not a comparison. For any same-chain ERC20→native swap above \$100, the user pays THORChain's mainnet vault-deposit gas (~120-180k gas) plus outbound fee, when an aggregator on the same chain (Uniswap/Curve/etc. via 1inch / LI.FI / KyberSwap) typically gives better net output.

## Expected behaviour

For each swap, query all eligible providers in parallel and pick the quote with the highest **net output** (output amount − network gas − provider fees − outbound fees, in a common denomination).

## Suggested fix direction

1. Replace the strategy-pattern provider resolution with a parallel-quote fetcher.
2. Compute net output per quote.
3. Pick the max-net-output quote.
4. Drop `KyberSwapProviderSelectionStrategyImpl` once ranking is in place — the \$100 threshold becomes unnecessary.

## Cross-platform context

Same architectural issue on iOS (`vultisig-ios`, where it's the user-visible bug today) and SDK (`vultisig-sdk`). Filed separately. Android codebase is native Kotlin (no `vultisig-sdk` dependency), so the fix lives here.

aminsato:
## Summary

- Same-chain ERC20→native swaps no longer use the deterministic `<$100 → KYBER, ≥$100 → THORCHAIN` rule. The form now queries every eligible provider in parallel and picks the quote with the highest USD net output (`expectedDstFiat − swapFeeFiat`).
- `KyberSwapProviderSelectionStrategy` and the hardcoded \$100 threshold are removed; `ResolveProviderUseCase` is simplified to delegate only to `DefaultSwapProviderSelectionStrategy` (still used for label-only callers in JoinKeysign / SwapTransactionToUiModel).
- New `SwapQuoteRepository.getEligibleProviders()` exposes the per-pair provider list; `SwapQuoteManager.fetchBestQuote()` fans out fetches with `async`+`coroutineScope` and ranks by USD net.

Closes #4315.

## Test plan

- [x] `./gradlew :app:compileDebugKotlin` (clean re-run with `--rerun-tasks`)
- [x] `./gradlew :app:testDebugUnitTest --tests SwapFormViewModelTest` — all 64 tests pass
- [ ] Manual: USDC→ETH (Ethereum) at \$10 / \$50 / \$200 / \$1000 — verify provider varies based on quote and that fees above \$100 are no longer fixed to THORCHAIN
- [ ] Manual: full keygen + keysign on one EVM swap and one cross-chain swap (e.g. ETH→BTC THORCHAIN)
- [ ] Manual: confirm referral / VULT BPS discount surfaces correctly when winner is THORCHAIN

## Notes

`JoinKeysignViewModel` and `SwapTransactionToUiModelMapper` still call `ResolveProviderUseCase` (returns first eligible) for the provider label. The provider that was actually used is on `EVMSwapPayloadJson.provider` and could be wired in for label correctness — left out of this PR to keep scope focused on #4315.

## Summary by CodeRabbit

* **Improvements**
  * Swap evaluates multiple eligible providers in parallel and selects the best quote by estimated fiat output.
  * Provider-specific discounts and referral handling are applied during comparison.
  * Provider discovery now exposes all eligible providers for a token pair.
  * Simplified provider resolution by removing a legacy Kyber selection strategy and its dependency binding.

* **New Features**
  * Multi-candidate quote selection with per-provider timeouts and failure handling.

* **Tests**
  * Updated swap tests to exercise multi-provider selection and fiat-based comparisons.


aminsato:
@coderabbitai full review

aminsato:
Thanks @coderabbitai. Addressed in 33c1dc8 — replies inline:

**Finding 1 — `ResolveProviderUseCase` provider label divergence:** Out of scope for #4315. The use case takes a `SwapSelectionContext` (src/dst/value); it does not have access to a persisted swap payload, so the proper fix lives at the call sites (`JoinKeysignViewModel`, `SwapTransactionToUiModelMapper`) where they should read `EVMSwapPayloadJson.provider` instead of re-resolving. Flagged as a known follow-up in the PR description.

**Finding 2 — error surfacing when all candidates fail:** Fixed. `fetchBestQuote` now prefers the first `SwapException` from the failures so `SwapFormViewModel.mapSwapExceptionToFormError` can render a user-friendly error, falling back to the original first failure otherwise.

**Finding 3 — normalize fees before ranking:** Fixed by removing the subtraction entirely and ranking on `estimatedDstFiat` alone. Reasoning: `swapFeeFiat` is not normalized across providers — for THOR/MAYA `expectedAmountOut` is **already** net of protocol+outbound fees per their docs, so subtracting `swapFeeFiat` was double-counting; for 1inch/Kyber `swapFeeFiat` is gas; for LI.FI it's the integrator fee. `estimatedDstFiat` is the closest like-for-like across providers ("destination amount the user expects to receive"). A more rigorous total-cost normalization (gas + provider fee + integrator fee per provider, all in fiat) is a worthwhile follow-up but requires reworking per-provider fee accounting beyond #4315.

rkokhatskyi:
@aminsato please resolve the conflicts

rkokhatskyi (commented):
@aminsato this needs to be carefully tested manually, please attach tx samples as well

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -147,11 +157,69 @@ constructor(
             srcFiatValueText = srcFiatValueText,
             estimatedDstTokenValue = estimatedDstTokenValue,
             estimatedDstFiatValue = fiatValueToString(estimatedDstFiatValue),
+            estimatedDstFiat = estimatedDstFiatValue,
             feeText = fiatValueToString(fiatFees),
             swapFeeFiat = fiatFees,
         )
     }
 
+    suspend fun fetchBestQuote(
```
rkokhatskyi: this is implicitly public on an `internal` class mark `fetchBestQuote` as `internal` so the surface stays consistent with the manager

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -147,11 +157,69 @@ constructor(
...
+        if (candidates.isEmpty()) {
+            throw SwapException.SwapIsNotSupported("Swap is not supported for this pair")
+        }
+
+        val results: List<Result<QuoteFetchResult>> = coroutineScope {
+            candidates
+                .map { candidate ->
+                    async {
+                        runCatching {
+                                fetchQuote(
```
rkokhatskyi: no per provider timeout so a single hung quote stalls `awaitAll` for the entire fan out wrap each `fetchQuote` in `withTimeout` so slow providers do not block the others

```kotlin
async {
    runCatching { withTimeout(QUOTE_FETCH_TIMEOUT) { fetchQuote(...) } }
        .onFailure { e -> if (e is CancellationException) throw e else Timber.w(e, ...) }
}
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -147,11 +157,69 @@ constructor(
...
+                                    tokenValue = tokenValue,
+                                    currency = currency,
+                                    vultBPSDiscount = candidate.vultBPSDiscount,
+                                    referral = candidate.referral,
+                                    amount = amount,
+                                )
+                            }
+                            .onFailure { e ->
+                                if (e is CancellationException) throw e
+                                Timber.w(e, "Quote fetch failed for %s", candidate.provider)
```
rkokhatskyi: this log only carries the provider enum so production traces have no swap pair or amount to reproduce

```kotlin
Timber.w(e, "Quote fetch failed provider=%s src=%s dst=%s amount=%s",
    candidate.provider, srcToken.id, dstToken.id, srcTokenValue)
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -147,11 +157,69 @@ constructor(
...
+                .awaitAll()
+        }
+
+        val successes = results.mapNotNull { it.getOrNull() }
+        if (successes.isEmpty()) {
+            val failures = results.mapNotNull { it.exceptionOrNull() }
+            throw failures.firstOrNull { it is SwapException } ?: failures.first()
+        }
+
+        // Rank on estimatedDstFiat alone — this represents the destination amount
```
rkokhatskyi, line 239: the PR description claims ranking by `expectedDstFiat − swapFeeFiat` but the code ranks `estimatedDstFiat` alone update the description so the comparator math matches what is shipped

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -147,11 +157,69 @@ constructor(
...
+        // the user expects to receive. Subtracting swapFeeFiat would double-count
+        // for THOR/MAYA (their expectedAmountOut is already net of protocol fees)
+        // and mix apples-to-oranges since swapFeeFiat is gas for 1inch/Kyber but
+        // an integrator fee for LI.FI.
+        return successes.maxWith(compareBy { it.estimatedDstFiat.value })
```
rkokhatskyi: use `maxBy` here `compareBy` adds nothing for a single key selector

```kotlin
return successes.maxBy { it.estimatedDstFiat.value }
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -380,20 +448,26 @@ constructor(
...
-                        amount,
-                        coinAndFiatValue,
-                        srcNativeToken,
-                    )
-                Pair(newNativeAmount, srcNativeToken)
-            } else {
-                Pair(fallbackFee, srcNativeToken)
+            when {
+                swapFeeTokenContract.isEmpty() -> Pair(fallbackFee, srcNativeToken)
+                // 1inch / many DEX aggregators use this sentinel for the native chain
```
rkokhatskyi: `fetchOneInchQuote` never reaches `resolveSwapFee` the only callers are `fetchKyberQuote` at 316 and `fetchLiFiJupiterQuote` at 418 so the comment about `1inch / many DEX aggregators` is misleading

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -380,20 +448,26 @@ constructor(
...
+                // token — the fee is already in src-native wei, no contract lookup needed.
+                swapFeeTokenContract.equals(NATIVE_TOKEN_SENTINEL, ignoreCase = true) ->
+                    Pair(swapFeeRaw.toBigIntegerOrNull() ?: fallbackFee, srcNativeToken)
```
rkokhatskyi, line 511: for cross chain LIFI quotes the integrator fee can be denominated on the destination chain so attaching `srcNativeToken` here would mislabel the fee token confirm the LIFI response shape before assuming src side denomination

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -493,6 +567,7 @@ constructor(
 
     companion object {
         private const val KYBER_AFFILIATE_FEE_BPS = 50
+        private const val NATIVE_TOKEN_SENTINEL = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
```
rkokhatskyi, line 625: this string is duplicated as `NULL_ADDRESS` in `KyberApi.kt:190` and `ONEINCH_NULL_ADDRESS` in `OneInchApi.kt:159` move it next to `EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT` so all three sources stay aligned

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -833,24 +832,55 @@ constructor(
...
                         val tokenValue = convertTokenAndValueToTokenValue(srcToken, srcTokenValue)
 
-                        val provider =
-                            resolveProvider(SwapSelectionContext(srcToken, dstToken, tokenValue))
-                                ?: throw SwapException.SwapIsNotSupported(
-                                    "Swap is not supported for this pair"
-                                )
-                        this@SwapFormViewModel.provider = provider
+                        val eligibleProviders =
+                            swapQuoteRepository.getEligibleProviders(srcToken, dstToken)
```
rkokhatskyi, line 840: this fan out hits up to 5 aggregators with the user wallet address per quote refresh whereas the old flow hit one chosen provider please flag the new disclosure surface for the team if it has not been reviewed

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -833,24 +832,55 @@ constructor(
...
 
                         val currency = appCurrencyRepository.currency.first()
 
-                        val vultBPSDiscount =
-                            vaultId?.let { id ->
-                                getDiscountBpsUseCase.invoke(id, provider).takeIf { it != 0 }
-                            }
-
-                        val referral =
+                        val baselineReferral =
```
rkokhatskyi, line 849: rename to `referralCode` `baseline` adds no information and the line above already uses that name

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -833,24 +832,55 @@ constructor(
...
                             referralCode.value
                                 ?: vaultId?.let { referralRepository.getExternalReferralBy(it) }
 
+                        val candidates =
+                            eligibleProviders.map { p ->
+                                val discount =
+                                    vaultId?.let { id ->
+                                        getDiscountBpsUseCase.invoke(id, p).takeIf { bps ->
```
rkokhatskyi: `.map` runs `getDiscountBpsUseCase.invoke` serially because it is a suspending call inside a synchronous lambda so 5 eligible providers means 5 sequential round trips before any quote starts run them in parallel

```kotlin
val candidates = coroutineScope {
    eligibleProviders.map { p ->
        async {
            val discount = vaultId?.let { id ->
                getDiscountBpsUseCase.invoke(id, p).takeIf { it != 0 }
            }
            QuoteCandidate(provider = p, vultBPSDiscount = discount, referral = baselineReferral)
        }
    }.awaitAll()
}
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -833,24 +832,55 @@ constructor(
...
+                                srcTokenValue = srcTokenValue,
+                                tokenValue = tokenValue,
+                                currency = currency,
+                                amount = amount,
+                            )
+
+                        val provider = quoteResult.provider
+                        this@SwapFormViewModel.provider = provider
+
+                        val winner = candidates.first { it.provider == provider }
```
rkokhatskyi: this throws `NoSuchElementException` if `fetchBestQuote` ever returns a provider that was not in the input return the winning `QuoteCandidate` directly from `fetchBestQuote` so the lookup goes away

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -833,24 +832,55 @@ constructor(
...
+                        val vultBPSDiscount = winner.vultBPSDiscount
+                        val referral = winner.referral
+
                         if (provider == SwapProvider.THORCHAIN) {
```
rkokhatskyi, line 893: if `THORCHAIN` ties with another provider on `estimatedDstFiat` and loses by a rounding epsilon the user silently loses their referral discount surface even though `THORCHAIN` was eligible the tie break depends on `Coin.swapProviders` set iteration order which is not a contract

data/src/main/kotlin/com/vultisig/wallet/data/repositories/SwapQuoteRepository.kt
```diff
@@ -91,6 +91,8 @@ interface SwapQuoteRepository {
     ): EVMSwapQuoteJson
 
     fun resolveProvider(srcToken: Coin, dstToken: Coin): SwapProvider?
+
+    fun getEligibleProviders(srcToken: Coin, dstToken: Coin): List<SwapProvider>
```
rkokhatskyi, line 17: new public method on the interface but no test asserts the eligibility list per chain or the cross chain filter behaviour also at lines 605 BSC and 615 Avalanche the impl compares `ticker in thorBscTokens` without `.uppercase()` while ETH at 575 and Base at 625 do normalise so a lower case ticker silently downgrades providers

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -537,11 +539,10 @@ internal class SwapFormViewModelTest {
...
     fun `calculateFees debounces amount changes`() =
         runTest(mainDispatcher) {
-            coEvery { resolveProvider.invoke(any()) } returns SwapProvider.THORCHAIN
+            every { swapQuoteRepository.getEligibleProviders(any(), any()) } returns
+                listOf(SwapProvider.THORCHAIN)
             coEvery {
-                swapQuoteManager.fetchQuote(
-                    any(),
-                    any(),
+                swapQuoteManager.fetchBestQuote(
```
rkokhatskyi, line 545: every test now mocks `swapQuoteManager.fetchBestQuote` to return a hardcoded result so the new ranking the partial failure aggregation and the `SwapException` rethrow path have zero coverage add at least one test that constructs two stub providers and asserts the higher `estimatedDstFiat` wins plus one where one provider throws and the other succeeds

app/src/main/java/com/vultisig/wallet/data/usecases/resolveprovider/ResolveProviderUseCase.kt
```diff
@@ -8,17 +8,9 @@ internal interface ResolveProviderUseCase : suspend (SwapSelectionContext) -> Sw
...
-    private val swapProviderSelectionStrategies =
-        listOf(defaultSwapProviderSelectionStrategy, kyberSwapProviderSelectionStrategy)
-
-    override suspend fun invoke(context: SwapSelectionContext): SwapProvider? {
-
-        return swapProviderSelectionStrategies
-            .maxBy(SwapProviderSelectionStrategy::priority)
-            .selectProvider(context)
-    }
+    override suspend fun invoke(context: SwapSelectionContext): SwapProvider? =
```
rkokhatskyi, line 14: with only one strategy left this is a one line proxy `priority` on `SwapProviderSelectionStrategy` is now unused and `value` on `SwapSelectionContext` is dead either inline the call to `swapQuoteRepository.resolveProvider` at the two consumer sites or simplify the layer down to a single concrete class

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/data/usecases/resolveprovider/ResolveProviderUseCase.kt
```diff
@@ -8,17 +8,9 @@ internal interface ResolveProviderUseCase : suspend (SwapSelectionContext) -> Sw
...
+        defaultSwapProviderSelectionStrategy.selectProvider(context)
```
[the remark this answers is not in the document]
aminsato: Acknowledged — this is an intentional, documented deferral (see the PR description's Notes section). Scope of #4315 is the ranking change; wiring keysign/history labels to `EVMSwapPayloadJson.provider` (and the equivalent for ThorChain/Maya/Jupiter payloads) is a separate cleanup. The drift is narrow in practice: ThorChain/Maya/Jupiter payloads encode the provider in the payload type itself, so only EVM same-chain ERC20↔native is affected, and the on-chain route is still driven by the persisted payload — only the label shown is heuristic. Will track as a follow-up.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Done — marked `fetchBestQuote` `internal`.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Done — wrapped each `fetchQuote` in `withTimeout(QUOTE_FETCH_TIMEOUT_MS = 15s)` so a hung aggregator can't stall the fan-out.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Done — log now includes `provider / src / dst / amount`.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Will update the PR description so the comparator math (`estimatedDstFiat` alone) matches what's shipped.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Done — replaced `maxWith(compareBy { ... })` with `maxBy { it.result.estimatedDstFiat.value }`.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Done — fixed the comment to read "Kyber / LI.FI / Jupiter" since those are the only callers that reach `resolveSwapFee`.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Good catch — for LI.FI the integrator fee is computed in `integratorFeeAmount(dstAmount, ...)` and we already pair it with `dstToken` in `fetchLiFiJupiterQuote` (lines 410-416), so the dst-side denomination is preserved. The `resolveSwapFee` path is only hit by Kyber/Jupiter (and OneInch via `fetchOneInchQuote`), where the sentinel correctly maps to src-native. So the existing code is fine; not changing in this PR but happy to add a comment if you want.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Agreed — leaving as a follow-up. Consolidating `NULL_ADDRESS` (KyberApi), `ONEINCH_NULL_ADDRESS` (OneInchApi) and the new `NATIVE_TOKEN_SENTINEL` here into one shared constant alongside `EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT` is a cross-file refactor that's out of scope for #4315; tracked.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Flagging this with the team. The fan-out does send the wallet address to up to 5 aggregators per quote refresh now (vs. one before). Will discuss whether we need to gate this behind user opt-in or trim the eligible set.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Kept the local var as `baselineReferral` because the VM has a member `private val referralCode = MutableStateFlow<String?>(null)` that's referenced inside the same scope (`referralCode.update { rc }`); renaming the local to `referralCode` would shadow it and break that update path. Open to other names if `baseline` reads as noise to you.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Done — wrapped the candidate construction in `coroutineScope { ... map { async { ... } }.awaitAll() }` so `getDiscountBpsUseCase.invoke` runs in parallel across providers.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Done — `fetchBestQuote` now returns `BestQuote(candidate, result)` so the caller pulls `vultBPSDiscount` and `referral` directly off the winner. The unsafe `candidates.first { it.provider == provider }` lookup is gone.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Real concern but out of scope for this PR. The contract for the referral-discount surface (when does a THORCHAIN-eligible pair get the discount UI even if THORCHAIN didn't win) is a product decision; the deterministic `<$100 → KYBER, ≥$100 → THORCHAIN` rule masked it before. Filing as a follow-up so we can decide whether to (a) always surface the THORCHAIN referral discount when THORCHAIN is eligible, or (b) tie-break on THORCHAIN within an epsilon.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/SwapQuoteRepository.kt
aminsato: Fixed the BSC/Avalanche `.uppercase()` normalization (matches what ETH/Base already do). Adding eligibility-list tests for `getEligibleProviders` per chain is a good follow-up but I'd rather land that in a separate PR alongside the SwapFormViewModel ranking/partial-failure tests you mentioned in the other thread.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Acknowledged. The current tests mock `fetchBestQuote` end-to-end which leaves the new ranking, partial-failure aggregation, and `SwapException` rethrow logic uncovered. Filing a follow-up to add SwapQuoteManager-level tests that exercise (a) two stub providers where the higher `estimatedDstFiat` wins and (b) one-throws-one-succeeds.

app/src/main/java/com/vultisig/wallet/data/usecases/resolveprovider/ResolveProviderUseCase.kt
aminsato: Agreed — once `KyberSwapProviderSelectionStrategy` is gone, the `UseCase` + strategy + `SwapSelectionContext.value` + `priority` machinery is overkill for what's now a single delegating proxy. Filing a follow-up to either inline `swapQuoteRepository.resolveProvider` at the two call sites (`JoinKeysignViewModel`, `SwapTransactionToUiModelMapper`) or collapse the layer to one concrete class. Pairing it with the same follow-up I described above (sourcing post-quote labels from the persisted payload) since they touch the same call sites.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -27,6 +27,10 @@ import java.math.BigDecimal
 import java.math.BigInteger
 import javax.inject.Inject
 import kotlin.coroutines.cancellation.CancellationException
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.withTimeout
```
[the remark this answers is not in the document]
aminsato: Real bug, fixed in cdc0037e0. `TimeoutCancellationException extends CancellationException`, so my blanket `if (e is CancellationException) throw e` was rethrowing timeouts and cancelling all sibling fetches via `awaitAll` — exactly defeating the per-provider timeout I had just added. Changed the guard to `if (e is CancellationException && e !is TimeoutCancellationException) throw e` so timeouts are treated as transient per-provider failures (logged and skipped) while genuine parent-scope cancellation still propagates.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
johnnyluo: no, let's fan out and find the best quote

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
johnnyluo: I think when we request quote from THORChain , we already pass in the referral code , so it has been taken into account already

johnnyluo (approved):
LGTM

PR #4321, opened 2026-04-29
fix: skip THORChain network id fetch on startup when no vaults exist

### LINKED ISSUE #4316: App fetches THORChain network id on startup even when no vault is imported
## Problem

When the app is launched on a fresh install (no vault imported yet), it still makes an unconditional network request to:

```
GET https://gateway.liquify.com/chain/thorchain_rpc/status
```

This happens before the user has any vault, so there is no functional reason to contact THORChain at this point. It produces unnecessary network traffic, leaks the install/launch event to the THORChain RPC gateway, and slows down the first-launch experience for users on slow connections.

## Reproduction

1. Install a fresh build of the app (or clear app data).
2. Launch the app — do not import or create a vault.
3. Observe the outgoing HTTP traffic.

Logged request:
```
REQUEST:  https://gateway.liquify.com/chain/thorchain_rpc/status
METHOD:   GET
Accept: */*
Accept-Charset: UTF-8
Content-Type: application/json
Content-Length: 0
```

## Root cause

`MainViewModel.init` calls `initializeThorChainNetworkId()` unconditionally:

- `app/src/main/java/com/vultisig/wallet/app/activity/MainViewModel.kt:104`
  ```kotlin
  viewModelScope.safeLaunch { initializeThorChainNetworkId() }
  ```

That use case (`InitializeThorChainNetworkIdUseCaseImpl.invoke`, `data/src/main/kotlin/com/vultisig/wallet/data/usecases/InitializeThorChainNetworkIdUseCase.kt:24-25`) always calls `thorChainRepository.fetchNetworkChainId()`, which calls `ThorChainApi.getNetworkChainId()` → `GET $THORCHAIN_RPC_URL/status` (`data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt:297-303`, with `THORCHAIN_RPC_URL = "https://gateway.liquify.com/chain/thorchain_rpc"` at line 627).

There is no check for "does the user actually have any vault?" and no check for "is THORChain even relevant for this user's vaults?".

## Suggested fix

Gate the THORChain network id initialization behind one of:

- Only run it when at least one vault exists (`vaultRepository.hasVaults()`).
- Better: only run it lazily, the first time a screen actually needs the THORChain network id (e.g., when entering a THORChain-related flow such as Send/Swap/Deposit on a THORChain-enabled vault).

Option 2 is preferable because it also avoids the request for users whose vaults don't include THORChain. The cached value flow already supports lazy population via `getCachedNetworkChainId()` + `fetchNetworkChainId()`.

## Impact

- No functional regression — call already happens on every cold start; the fix just defers it.
- Reduces unnecessary network requests on first launch.
- Avoids contacting an external RPC gateway before the user has opted into the wallet.

aminsato:
## Summary

- On a fresh install (no vault imported yet), `MainViewModel.init` was unconditionally calling `InitializeThorChainNetworkIdUseCase`, which fired `GET https://gateway.liquify.com/chain/thorchain_rpc/status` before the user had opted into the wallet.
- Gate the use case on `vaultRepository.hasVaults()` so the cache read and network fetch are both skipped when no vaults exist. The default `THORCHAIN_NETWORK_ID = "thorchain-1"` covers any pre-vault code path; once a vault exists, the next launch initializes normally.
- Rethrow `CancellationException` in the existing fetch catch block (project coding principle).
- Gating inside the use case keeps the call site behavior intact for any future caller.

Closes #4316

## Test plan

- [x] `./gradlew :data:testDebugUnitTest` — new `InitializeThorChainNetworkIdUseCaseImplTest` (no-vault skip / cache+fetch / fetch-failure-keeps-cache) all pass
- [x] `./gradlew :app:testDebugUnitTest --tests MainViewModelTest` — existing tests still pass
- [x] `./gradlew :data:assembleDebug`
- [ ] Manual: fresh install → confirm no `GET .../thorchain_rpc/status` request on launch
- [ ] Manual: install with existing vault → confirm THORChain network id still initializes on launch

Co-Authored-By: aminsato <Amin.saradar@yahoo.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error recovery during THORChain network identification operations. The system now handles fetch failures more gracefully while preserving cached network data. Operation cancellations are properly managed.

* **Tests**
  * Added comprehensive test coverage for THORChain network initialization, validating behavior across vault detection, successful initialization, and network error scenarios.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/InitializeThorChainNetworkIdUseCase.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/InitializeThorChainNetworkIdUseCase.kt
@@ -2,30 +2,42 @@ package com.vultisig.wallet.data.usecases
 
 import com.vultisig.wallet.data.crypto.ThorChainHelper
 import com.vultisig.wallet.data.repositories.ThorChainRepository
+import com.vultisig.wallet.data.repositories.VaultRepository
 import javax.inject.Inject
+import kotlinx.coroutines.CancellationException
 import timber.log.Timber
 
 interface InitializeThorChainNetworkIdUseCase : suspend () -> Unit
 
 internal class InitializeThorChainNetworkIdUseCaseImpl
 @Inject
-constructor(private val thorChainRepository: ThorChainRepository) :
-    InitializeThorChainNetworkIdUseCase {
+constructor(
+    private val thorChainRepository: ThorChainRepository,
+    private val vaultRepository: VaultRepository,
+) : InitializeThorChainNetworkIdUseCase {
 
     override suspend fun invoke() {
+        if (!vaultRepository.hasVaults()) {
+            Timber.d("Skipping THORChain network id init: no vaults")
+            return
+        }
+
         Timber.d("Initializing THORChain network id")
 
         val cachedNetworkChainId = thorChainRepository.getCachedNetworkChainId()
         if (!cachedNetworkChainId.isNullOrBlank()) {
             ThorChainHelper.THORCHAIN_NETWORK_ID = cachedNetworkChainId
-            Timber.d("THORChain network id initialized from cache with: $cachedNetworkChainId")
+            Timber.d("THORChain network id initialized from cache with: %s", cachedNetworkChainId)
         }
 
         try {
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4276, opened 2026-04-25
fix: multiple onboarding UI issues (#4255)

aminsato:
## Summary

Addresses the five sub-items reported in #4255:

- **Peer discovery off-by-one**: selecting N devices used to display "1 of N+1" because `minimumDevicesDisplayed` added `+1` to the device count. Removed the offset in three places in `KeygenPeerDiscoveryViewModel.kt` so the displayed count matches the actual target.
- **Setup vault Rive dashed border**: removed the placeholder `dashedBorder` modifier from `SetupVaultRive.kt` so the animation sits flush against the screen background.
- **Backup onboarding Rive dashed border**: same fix in `VaultBackupOnboardingScreen.kt`.
- **"Save backup to the cloud" starts scrolled / cuts text**: in `BackupVaultScreen.kt`, anchor scroll to position 0 on entry via `LaunchedEffect` (covers the multi-backup case where the same composable inherits prior state) and reduce the Rive size from 340×240dp → 266×170dp so the title/description fit on smaller devices.
- **Scan QR help bottom sheet outdated layout**: `ScanQrHelpModalContent.kt` updated to match Figma — top corner radius 34dp, padding 7/40/16dp, title uses `Theme.brockmann.headings.title3` without the gradient, description uses `text.tertiary`, button switched to `VsButtonVariant.CTA` (`#0B4EFF`) with intrinsic width, grabber sized 36×5dp. The `scan_qr_help` drawable was also replaced with a fresh phone-mockup asset matching the current home screen layout.
- **Localized strings**: all 10 locales updated (`scan_qr_code_modal_scan_the` is now the full title, `_annotation` rewritten, `_next` → "Start scanning"); the unused `scan_qr_code_modal_qr_code` key was removed everywhere since the title no longer uses gradient text spans.

Closes #4255

## Test plan

- [ ] Onboarding device-count screen → peer discovery: select 2 devices, confirm header shows "1 of 2" (not "1 of 3"); repeat with 3 and 4 devices
- [ ] Setup vault info screen: confirm the Rive animation sits flush, no dashed outline
- [ ] Backup onboarding screen: same — no dashed outline
- [ ] Save backup to cloud screen: navigate in/out, confirm it always opens at the top; on a small/short device confirm the description and "I understand" checkbox are not cut by the bottom bar
- [ ] Scan QR help bottom sheet (peer discovery → "?" icon): confirm new copy, single-line title without gradient, CTA-blue "Start scanning" button, updated phone illustration
[...]

PR #369, opened 2024-06-12
Remove caret_left icon from AddVaultScreen

aminsato:
Closes #357

Remove caret_left icon from AddVaultScreen based on the latest Figma design update

image

PR #4337, opened 2026-04-30
fix(solana): thread memo through SPL token signing paths

realpaaao:
## Summary

`Solana.TokenTransfer` and `Solana.CreateAndTransferToken` both expose a `memo` field; without setting it, SPL sends silently drop any memo the user attached. Native SOL already wired memo through, so a THORChain LP add for native SOL works but the same flow on USDC.SOL (or any SPL asset) would broadcast memo-less and get refunded by THORChain.

Two-line fix in `SolanaHelper.kt`: `keysignPayload.memo?.let { transfer.setMemo(it) }` on both SPL builder paths, mirroring the existing native-SOL pattern directly above.

## Cross-client coordination

Same fix lands in vultisig-sdk PR #369 (extension/desktop) and vultisig-ios PR #4276. Native SOL was already correct on all three; only SPL was broken. No proto/wire-format change.

## Test plan

- [x] ktfmt clean
- [ ] Build the app and send 0.001 USDC with a memo to a self-address — verify the on-chain tx contains a Memo program instruction with the expected memo string

## Summary by CodeRabbit

* **New Features**
  * Added memo support for Solana transactions and token transfers.


johnnyluo:
This cause the test to fail , need to be fixed

johnnyluo (approved):
LGTM

PR #4320, opened 2026-04-29
feat(thorchain): LP UI — positions tab, add/remove flow (#2141)

aminsato:
## Tranasction hash
https://runescan.io/tx/492BE81A5053E6FA7B410E25BE69FD0EDD24BCCD3DD75381671EC785D0C493F1
## Summary
Builds on the LP data layer (#4309) to surface THORChain liquidity-pool positions in the DeFi screen and wire the add/remove transaction flows.

Closes #2141

- **LP tab on the THORChain DeFi positions screen** with a Manage Positions bottom sheet (rendered for empty state too) listing available pools from `getPoolStats()`. Toggling a pool drives what shows on the LP tab, including placeholder cards for selected pools the user hasn't entered yet so Add is reachable.
- **Real positions** via `GetThorChainLpPositionsUseCase`, mapped to title / redeem amounts / APR / fiat total. The dummy `loadLpPositions` is gone.
- **`DepositFormViewModel` LP flow is chain-aware**: native token is RUNE for ThorChain / CACAO for MayaChain; new `loadThorChainRemoveLpData()` uses the use case's `runeRedeemValue` so the slider math works at RUNE's 1e8 scale. `RemoveLpCalculator.computeCacaoDisplay` generalized to `computeAmountDisplay(decimals)`; `RemoveLpScreen` reads the chain symbol from state instead of hardcoding "CACAO".
- **Add LP memo now includes the paired-chain address** (`+:POOL:PAIRED_ADDRESS`) — resolved from the vault for the pool's asset chain — so symmetric add credits correctly when the asset half is later deposited.
- **Remove button is disabled** on cards with no liquidity units (placeholders, fully-withdrawn Maya entries).
- **Chain-aware Add/Remove LP titles** ("Add THORChain LP" vs "Add MayaChain LP"), localized across all 10 locales.
- New `parseThorChainPool` helper (used by ViewModel + dialog construction) with full unit-test coverage.

## Test plan
- [ ] Open THORChain DeFi positions → LP tab → tap **Manage Positions** → bottom sheet lists pools → toggle BTC.BTC and ETH.ETH → Done. LP tab populates with placeholder cards (`0 RUNE + 0 BTC` etc., Remove disabled, Add enabled).
- [ ] Tap **Add** on a placeholder → header reads "Add THORChain LP" → enter RUNE amount → Continue → Verify screen shows memo `+:BTC.BTC:<my BTC address>` and a RUNE source amount → keysign succeeds.
[...]

PR #4340, opened 2026-05-01
perf(thorchain): single-pool LP lookup for Remove LP screen (#4338)

### LINKED ISSUE #4338: perf: avoid O(pool-count) network fan-out in loadThorChainRemoveLpData() remove-LP flow
## Summary

In `DepositFormViewModel.loadThorChainRemoveLpData()`, the remove-LP screen calls `GetThorChainLpPositionsUseCase` and then filters with `.find { it.pool == poolId }`.

`GetThorChainLpPositionsUseCase` internally calls `getPoolStats()` and probes **every available pool** via per-pool Midgard requests before the result is filtered down to the single target pool. On a typical THORChain deployment this is an O(pool-count) network fan-out triggered every time a user opens the Remove LP screen.

## Expected Behaviour

Opening the Remove LP screen for a specific pool should require only a single-pool lookup, not a full fan-out across all available pools.

## Suggested Fix

Introduce a single-pool lookup path — either:
- Add a `poolId` filter parameter to `GetThorChainLpPositionsUseCase.invoke()` so it short-circuits after fetching the one relevant pool, or
- Extract a dedicated `GetThorChainLpPositionUseCase` (singular) that accepts `poolId`, `runeAddress`, and `assetAddressesByPool` and returns the specific position directly.

Remove the `.find { it.pool == poolId }` call from `DepositFormViewModel` once the use case returns the targeted position.

## References

- PR: https://github.com/vultisig/vultisig-android/pull/4320
- Review comment: https://github.com/vultisig/vultisig-android/pull/4320#discussion_r3167528908
- Requested by: @johnnyluo

aminsato:
## Summary

- Closes #4338. Remove LP screen previously triggered an O(pool-count) fan-out — `GetThorChainLpPositionsUseCase` calls `getPoolStats()` and probes every available pool via per-pool Midgard requests, just to be filtered down to the one target pool by `.find { it.pool == poolId }`.
- Adds `GetThorChainLpPositionUseCase` (singular) that takes `poolId`, `runeAddress`, and optional `assetAddress`. It issues at most two requests (rune-side, then asset-side fallback) and skips `getPoolStats` entirely — the Remove LP flow doesn't read APR.
- `DepositFormViewModel.loadThorChainRemoveLpData()` switches to the new use case, dropping the post-filter. `GetThorChainLpPositionsUseCase` is unchanged and still drives `ThorchainDefiPositionsViewModel` (which legitimately needs the full list).

## Test plan

- [x] `./gradlew :data:testDebugUnitTest --tests "*GetThorChainLpPositionUseCaseTest"` — 7 new tests covering rune-side success, asset-side fallback, no-position, zero-units, no fallback when asset address is null, network error swallowing, unexpected exception propagation
- [x] `./gradlew :data:testDebugUnitTest --tests "*GetThorChainLpPositionsUseCaseTest"` — pre-existing tests still pass
- [x] `./gradlew :app:testDebugUnitTest --tests "*DepositFormViewModelTest"` — passes with new dependency wired in
- [x] `./gradlew :data:compileDebugKotlin :app:compileDebugKotlin` — clean compile
- [ ] Manual: open Remove LP for a THORChain pool, verify position loads correctly and only the target pool is queried

## Summary by CodeRabbit

* **New Features**
  * Single LP-position lookup to directly retrieve a user's THORChain LP position for the current pool.

* **Refactor**
  * Cancels in-flight LP load operations when starting new requests or switching deposit options to prevent stale UI updates; populates remove-LP amounts and balances from the single-position result.

* **Tests**
  * Added unit tests covering single-position lookup, asset-side fallback, zero-unit handling, malformed data, and network/error cases.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
@@ -58,6 +58,7 @@ import com.vultisig.wallet.data.repositories.VaultRepository
 import com.vultisig.wallet.data.usecases.DepositMemoAssetsValidatorUseCase
 import com.vultisig.wallet.data.usecases.GasFeeToEstimatedFeeUseCase
 import com.vultisig.wallet.data.usecases.GasFeeToEstimatedFeeUseCaseImpl
+import com.vultisig.wallet.data.usecases.GetThorChainLpPositionUseCase
 import com.vultisig.wallet.data.usecases.GetThorChainLpPositionsUseCase
 import com.vultisig.wallet.data.usecases.RequestAddressBookEntryUseCase
 import com.vultisig.wallet.data.usecases.RequestQrScanUseCase
@@ -217,6 +218,7 @@ constructor(
     private val gasFeeToEstimate: GasFeeToEstimatedFeeUseCaseImpl,
     private val requestAddressBookEntry: RequestAddressBookEntryUseCase,
     private val getThorChainLpPositionsUseCase: GetThorChainLpPositionsUseCase,
+    private val getThorChainLpPositionUseCase: GetThorChainLpPositionUseCase,
 ) : ViewModel() {
 
     private val appCurrency =
@@ -663,15 +665,14 @@ constructor(
                 if (currentVaultId != null) {
                     resolvePairedAddress(Chain.ThorChain, currentVaultId, poolId)
                 } else null
-            val assetAddressesByPool = pairedAddress?.let { mapOf(poolId to it) } ?: emptyMap()
             val position =
                 withContext(Dispatchers.IO) {
-                        getThorChainLpPositionsUseCase(
-                            runeAddress = userAddress,
-                            assetAddressesByPool = assetAddressesByPool,
-                        )
-                    }
-                    .find { it.pool == poolId }
+                    getThorChainLpPositionUseCase(
+                        poolId = poolId,
+                        runeAddress = userAddress,
+                        assetAddress = pairedAddress,
+                    )
+                }
 
             if (position == null || position.units <= BigInteger.ZERO) {
                 _state.update {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
@@ -59,7 +59,6 @@ import com.vultisig.wallet.data.usecases.DepositMemoAssetsValidatorUseCase
 import com.vultisig.wallet.data.usecases.GasFeeToEstimatedFeeUseCase
 import com.vultisig.wallet.data.usecases.GasFeeToEstimatedFeeUseCaseImpl
 import com.vultisig.wallet.data.usecases.GetThorChainLpPositionUseCase
-import com.vultisig.wallet.data.usecases.GetThorChainLpPositionsUseCase
 import com.vultisig.wallet.data.usecases.RequestAddressBookEntryUseCase
 import com.vultisig.wallet.data.usecases.RequestQrScanUseCase
 import com.vultisig.wallet.data.usecases.ValidateMayaTransactionHeightUseCase
@@ -217,7 +216,6 @@ constructor(
     private val tokenRepository: TokenRepository,
     private val gasFeeToEstimate: GasFeeToEstimatedFeeUseCaseImpl,
     private val requestAddressBookEntry: RequestAddressBookEntryUseCase,
-    private val getThorChainLpPositionsUseCase: GetThorChainLpPositionsUseCase,
     private val getThorChainLpPositionUseCase: GetThorChainLpPositionUseCase,
 ) : ViewModel() {
 
@@ -275,6 +273,7 @@ constructor(
     private val secureAssetNodeValue = MutableStateFlow<String?>(null)
     private var addressJob: Job? = null
     private var whitelistJob: Job? = null
+    private var loadLpJob: Job? = null
     private var depositTypeAction: String? = null
     private var bondAddress: String? = null
     private var lpPoolId: String? = null
@@ -647,83 +646,90 @@ constructor(
                 balance = R.string.share_balance_loading.asUiText(),
             )
         }
-        viewModelScope.safeLaunch {
-            val userAddress =
-                withTimeoutOrNull(ADDRESS_AWAIT_TIMEOUT_MS) { address.filterNotNull().first() }
-                    ?.address
-                    ?: run {
-                        _state.update {
-                            it.copy(
-                                errorText =
-                                    UiText.StringResource(R.string.dialog_default_error_body)
-                            )
+        loadLpJob?.cancel()
+        loadLpJob =
+            viewModelScope.safeLaunch {
+                val userAddress =
+                    withTimeoutOrNull(ADDRESS_AWAIT_TIMEOUT_MS) { address.filterNotNull().first() }
+                        ?.address
+                        ?: run {
+                            _state.update {
+                                it.copy(
+                                    errorText =
+                                        UiText.StringResource(R.string.dialog_default_error_body)
+                                )
+                            }
+                            return@safeLaunch
                         }
-                        return@safeLaunch
+                val currentVaultId = vaultId
+                val pairedAddress =
+                    if (currentVaultId != null) {
+                        resolvePairedAddress(Chain.ThorChain, currentVaultId, poolId)
+                    } else null
+                val position =
+                    withContext(Dispatchers.IO) {
+                        getThorChainLpPositionUseCase(
+                            poolId = poolId,
+                            runeAddress = userAddress,
+                            assetAddress = pairedAddress,
+                        )
                     }
-            val currentVaultId = vaultId
-            val pairedAddress =
-                if (currentVaultId != null) {
-                    resolvePairedAddress(Chain.ThorChain, currentVaultId, poolId)
-                } else null
-            val position =
-                withContext(Dispatchers.IO) {
-                    getThorChainLpPositionUseCase(
-                        poolId = poolId,
-                        runeAddress = userAddress,
-                        assetAddress = pairedAddress,
-                    )
+
+                if (position == null || position.units <= BigInteger.ZERO) {
+                    _state.update {
+                        it.copy(
+                            availableLpUnits = null,
+                            removeLpUnitsDivisor = BigInteger.ZERO,
+                            removeLpPoolDepth = BigInteger.ZERO,
+                            balance = UiText.Empty,
+                            errorText = UiText.StringResource(R.string.dialog_default_error_body),
+                        )
+                    }
+                    return@safeLaunch
                 }
 
-            if (position == null || position.units <= BigInteger.ZERO) {
+                // Use the pre-computed redeem value from the use case as `poolDepth` and the user's
+                // own
+                // units as `totalPoolUnits`. With selectedUnits = percent * userUnits, the
+                // calculator
+                // produces percent * runeRedeemValue, which is the symmetric RUNE half of
+                // withdrawal.
+                // Keep BigInteger end-to-end for whale positions whose units exceed Long.MAX_VALUE.
+                val userUnits = position.units
+                val runeRedeemBase = position.runeRedeemValue
+                val symbol = Coins.ThorChain.RUNE.ticker
+                val userUnitsLongOrNull =
+                    userUnits
+                        .takeIf { it.signum() > 0 && it.bitLength() < Long.SIZE_BITS }
+                        ?.toLong()
+                val userRune =
+                    userUnitsLongOrNull?.let { userUnitsLong ->
+                        RemoveLpCalculator.computeAmountDisplay(
+                            selectedUnits = userUnitsLong,
+                            poolDepth = runeRedeemBase,
+                            totalPoolUnits = userUnits,
+                            decimals = RemoveLpCalculator.RUNE_DECIMALS,
+                        )
+                    }
+                val balanceText =
+                    if (userRune != null) {
+                        UiText.FormattedText(
+                            R.string.remove_pool_amount_format,
+                            listOf(userRune, symbol),
+                        )
+                    } else UiText.Empty
                 _state.update {
                     it.copy(
-                        availableLpUnits = null,
-                        removeLpUnitsDivisor = BigInteger.ZERO,
-                        removeLpPoolDepth = BigInteger.ZERO,
-                        balance = UiText.Empty,
-                        errorText = UiText.StringResource(R.string.dialog_default_error_body),
+                        availableLpUnits = userUnits.toString(),
+                        removeLpUnitsDivisor = userUnits,
+                        removeLpPoolDepth = runeRedeemBase,
+                        removeLpDecimals = RemoveLpCalculator.RUNE_DECIMALS,
+                        removeLpTokenSymbol = symbol,
+                        balance = balanceText,
                     )
                 }
-                return@safeLaunch
+                setRemoveLpPercent(state.value.removeLpPercent)
             }
-
-            // Use the pre-computed redeem value from the use case as `poolDepth` and the user's own
-            // units as `totalPoolUnits`. With selectedUnits = percent * userUnits, the calculator
-            // produces percent * runeRedeemValue, which is the symmetric RUNE half of withdrawal.
-            // Keep BigInteger end-to-end for whale positions whose units exceed Long.MAX_VALUE.
-            val userUnits = position.units
-            val runeRedeemBase = position.runeRedeemValue
-            val symbol = Coins.ThorChain.RUNE.ticker
-            val userUnitsLongOrNull =
-                userUnits.takeIf { it.signum() > 0 && it.bitLength() < Long.SIZE_BITS }?.toLong()
-            val userRune =
-                userUnitsLongOrNull?.let { userUnitsLong ->
-                    RemoveLpCalculator.computeAmountDisplay(
-                        selectedUnits = userUnitsLong,
-                        poolDepth = runeRedeemBase,
-                        totalPoolUnits = userUnits,
-                        decimals = RemoveLpCalculator.RUNE_DECIMALS,
-                    )
-                }
-            val balanceText =
-                if (userRune != null) {
-                    UiText.FormattedText(
-                        R.string.remove_pool_amount_format,
-                        listOf(userRune, symbol),
-                    )
-                } else UiText.Empty
-            _state.update {
-                it.copy(
-                    availableLpUnits = userUnits.toString(),
-                    removeLpUnitsDivisor = userUnits,
-                    removeLpPoolDepth = runeRedeemBase,
-                    removeLpDecimals = RemoveLpCalculator.RUNE_DECIMALS,
-                    removeLpTokenSymbol = symbol,
-                    balance = balanceText,
-                )
-            }
-            setRemoveLpPercent(state.value.removeLpPercent)
-        }
     }
 
     fun selectBondAsset(asset: String) {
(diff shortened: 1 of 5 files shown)


aminsato:
@coderabbit full review

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
```diff
@@ -217,6 +218,7 @@ constructor(
     private val gasFeeToEstimate: GasFeeToEstimatedFeeUseCaseImpl,
     private val requestAddressBookEntry: RequestAddressBookEntryUseCase,
     private val getThorChainLpPositionsUseCase: GetThorChainLpPositionsUseCase,
+    private val getThorChainLpPositionUseCase: GetThorChainLpPositionUseCase,
```
rkokhatskyi, line 219: the plural `getThorChainLpPositionsUseCase` on line 220 above and its import on line 62 are no longer referenced anywhere in the body of this file

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
```diff
@@ -663,15 +665,14 @@ constructor(
...
-            val assetAddressesByPool = pairedAddress?.let { mapOf(poolId to it) } ?: emptyMap()
             val position =
                 withContext(Dispatchers.IO) {
-                        getThorChainLpPositionsUseCase(
-                            runeAddress = userAddress,
-                            assetAddressesByPool = assetAddressesByPool,
-                        )
-                    }
-                    .find { it.pool == poolId }
+                    getThorChainLpPositionUseCase(
```
rkokhatskyi: `loadThorChainRemoveLpData` runs inside a `safeLaunch` that the view model never tracks or cancels so a second `selectDepositOption(RemoveLiquidity)` while one is in flight lets the slower fetcher`'`s `_state.update` at line 715 win whereas `ThorchainDefiPositionsViewModel.kt:802` keeps a `loadLpJob` reference and cancels it before relaunching

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * Use this when the caller already knows which pool to inspect (e.g. the Remove LP screen) — it
+     * issues at most two requests (rune-side, then asset-side fallback) instead of fanning out
+     * across every available pool.
+     *
+     * @param poolId the pool to query, e.g. `"BTC.BTC"`.
+     * @param runeAddress the user's RUNE address — tried first.
+     * @param assetAddress optional non-RUNE side address used as a fallback when an LP record is
+     *   keyed by the pool's asset side rather than RUNE.
+     * @return the position, or `null` if the user has no LP units in this pool.
```
rkokhatskyi: the `@return` says `null` means the user has no LP units in this pool but the impl also returns `null` on swallowed `IOException` and `NetworkException` and on malformed numeric strings so callers cannot tell absence from failure

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+        runeAddress: String,
+        assetAddress: String? = null,
+    ): ThorChainLpPosition?
+}
+
+internal class GetThorChainLpPositionUseCaseImpl
+@Inject
+constructor(private val thorChainApi: ThorChainApi) : GetThorChainLpPositionUseCase {
+
+    override suspend fun invoke(
```
rkokhatskyi, line 41: this body is a near copy of `GetThorChainLpPositionsUseCase.fetchPosition` at lines 73 to 100 so the same zero coercion bug now lives in two files and any future fix has to land twice while extracting a shared helper or having the plural call this singular plus an APR overlay would collapse it

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+        poolId: String,
+        runeAddress: String,
+        assetAddress: String?,
+    ): ThorChainLpPosition? {
+        val lp =
+            try {
+                thorChainApi.getLiquidityProvider(poolId, runeAddress)
```
rkokhatskyi: if thornode returns a stale rune side record with `units = "0"` the elvis here short circuits and the asset side fallback at line 44 never runs even when an asset keyed position exists which is the same gap that lives in `GetThorChainLpPositionsUseCase.kt:80`

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+                    ?: assetAddress?.let { thorChainApi.getLiquidityProvider(poolId, it) }
```
rkokhatskyi: `assetAddress?.let` only guards against `null` so a non null blank string would still fire a thornode request with an empty address path segment that `takeIf { it.isNotBlank() }?.let { ... }` would skip

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+            } catch (e: IOException) {
+                Timber.w(e, "Failed to fetch LP position for pool %s", poolId)
+                null
+            } catch (e: NetworkException) {
+                Timber.w(e, "Failed to fetch LP position for pool %s", poolId)
+                null
+            } ?: return null
+
+        val units = lp.units.toBigIntegerOrNull() ?: BigInteger.ZERO
```
rkokhatskyi: also for `units` the `?: BigInteger.ZERO` then `signum() == 0` is a roundabout `?: return null` since malformed and zero both end up returning null

data/src/test/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCaseTest.kt
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Test
+    fun `falls back to asset address when rune-side has no record`() = runTest {
+        coEvery { api.getLiquidityProvider(POOL, RUNE_ADDR) } returns null
+        coEvery { api.getLiquidityProvider(POOL, ASSET_ADDR) } returns lp(units = "42")
+
+        val position = useCase(poolId = POOL, runeAddress = RUNE_ADDR, assetAddress = ASSET_ADDR)
+
+        assertEquals(BigInteger("42"), position?.units)
```
rkokhatskyi, line 56: the asset fallback test asserts only `units` so a regression where the returned `pool` field is set to something other than the input `poolId` after the asset side fallback would slip through

data/src/test/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCaseTest.kt
```diff
@@ -0,0 +1,119 @@
...
+
+    @Test
+    fun `swallows network errors and returns null`() = runTest {
+        coEvery { api.getLiquidityProvider(POOL, RUNE_ADDR) } throws
+            NetworkException(httpStatusCode = 500, message = "boom")
+
+        val position = useCase(poolId = POOL, runeAddress = RUNE_ADDR)
+
+        assertNull(position)
+    }
```
rkokhatskyi, line 128: the catch in the impl covers both `IOException` and `NetworkException` but only `NetworkException` is exercised so the `IOException` has zero coverage

app/src/test/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModelTest.kt
```diff
@@ -83,6 +83,9 @@ internal class DepositFormViewModelTest {
     private val getThorChainLpPositionsUseCase:
         com.vultisig.wallet.data.usecases.GetThorChainLpPositionsUseCase =
         mockk(relaxed = true)
+    private val getThorChainLpPositionUseCase:
+        com.vultisig.wallet.data.usecases.GetThorChainLpPositionUseCase =
+        mockk(relaxed = true)
```
rkokhatskyi, line 85: `relaxed = true` here lets any unintended call from the view model and any future signature change on the use case pass silently so when a focused test for `selectDepositOption(RemoveLiquidity)` is added it should drop relaxed and use explicit `coEvery` plus `coVerify`

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
```diff
@@ -0,0 +1,64 @@
...
+        if (units.signum() == 0) return null
+
+        return ThorChainLpPosition(
+            pool = poolId,
+            units = units,
+            runeRedeemValue = lp.runeRedeemValue.toBigIntegerOrNull() ?: BigInteger.ZERO,
+            assetRedeemValue = lp.assetRedeemValue.toBigIntegerOrNull() ?: BigInteger.ZERO,
```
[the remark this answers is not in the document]
aminsato: Fixed in b53b33e1c — `runeRedeemValue` and `assetRedeemValue` now `return null` when unparseable, so a bad Midgard string surfaces as 'no position' instead of a zero-value Remove LP preview. Same treatment for `units`.

data/src/test/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCaseTest.kt
```diff
@@ -0,0 +1,119 @@
...
+
+        val position = useCase(poolId = POOL, runeAddress = RUNE_ADDR)
+
+        assertEquals(POOL, position?.pool)
+        assertEquals(BigInteger("1000"), position?.units)
+        assertEquals(BigInteger("1500"), position?.runeRedeemValue)
+        assertEquals(BigInteger("2500"), position?.assetRedeemValue)
+        assertNull(position?.annualPercentageRate)
+        coVerify(exactly = 0) { api.getPoolStats(any()) }
+        coVerify(exactly = 0) { api.getLiquidityProvider(neq(POOL), any()) }
```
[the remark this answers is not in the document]
aminsato: Fixed in b53b33e1c — the rune-side success test now asserts `coVerify(exactly = 1) { api.getLiquidityProvider(POOL, RUNE_ADDR) }` and `coVerify(exactly = 0) { api.getLiquidityProvider(POOL, ASSET_ADDR) }`.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
aminsato: Fixed in b53b33e1c — dropped both the import and the constructor field. The plural use case is still used by `ThorchainDefiPositionsViewModel` so the binding stays in the DI module.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
aminsato: Fixed in b53b33e1c — added a `loadLpJob` field and call `loadLpJob?.cancel()` before relaunching, mirroring `ThorchainDefiPositionsViewModel.kt:802`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
aminsato: Fixed in b53b33e1c — KDoc now spells out all four conditions that yield `null`: no record on either side, both sides report zero units, an unparseable amount field, or a swallowed `IOException`/`NetworkException`. Splitting absence from failure into separate return types would be a wider API change; happy to do that as a follow-up if you want it.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
aminsato: Won't fix in this PR — the duplicated body is ~15 lines and the plural use case applies an APR overlay on top, so collapsing them cleanly means restructuring the plural's `fetchPosition` to call the singular and add APR. That's bigger than the perf fix this PR is scoped to. I'd rather keep the change minimal here and open a follow-up to consolidate. The matching parsing fixes (#11/#7) were applied to the singular only — the plural still returns `BigInteger.ZERO` on malformed strings; if you want me to backport the same hardening into the plural in this PR, say the word.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
aminsato: Fixed in b53b33e1c — the rune-side response is now passed through `takeIf { it.hasUnits() }`, so a stale `units = "0"` record falls through to the asset-side fetch. There's a new test `falls back to asset address when rune-side reports zero units` that pins this down. The matching gap in the plural use case is unchanged in this PR (see my note on #4).

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
aminsato: Fixed in b53b33e1c — now `assetAddress?.takeIf { it.isNotBlank() }?.let { ... }`, with a `ignores blank asset address` test asserting only the rune-side call fires.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCase.kt
aminsato: Fixed in b53b33e1c — `units` now uses `?: return null` directly. `signum() == 0` is still checked separately because `"0"` parses fine to `BigInteger.ZERO`, so the two branches aren't actually equivalent: malformed → null (data error), zero → null (no position). The intent is clearer now even if the outcome converges.

data/src/test/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCaseTest.kt
aminsato: Fixed in b53b33e1c — the asset-fallback test now asserts `assertEquals(POOL, position?.pool)`.

data/src/test/kotlin/com/vultisig/wallet/data/usecases/GetThorChainLpPositionUseCaseTest.kt
aminsato: Fixed in b53b33e1c — added `swallows io errors and returns null` exercising the `IOException` branch.

app/src/test/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModelTest.kt
aminsato: Keeping `relaxed = true` in this PR — this VM test setup is shared across the whole file, dropping relaxed would cascade into setup for every other test that doesn't care about the LP path. I'd add the focused `selectDepositOption(RemoveLiquidity)` test in a follow-up with its own non-relaxed mock so the change is scoped. Happy to do it now if you'd rather not split.

aminsato:
Pushed `b53b33e1c` addressing review feedback. Summary:

**Applied**
- Drop unused plural `getThorChainLpPositionsUseCase` import + injection from `DepositFormViewModel`.
- Track `loadLpJob` and cancel before relaunch in `loadThorChainRemoveLpData`, mirroring `ThorchainDefiPositionsViewModel`.
- `GetThorChainLpPositionUseCase`: malformed redeem amounts and `units` now `return null` instead of coercing to zero; stale rune-side record with `units == 0` falls through to the asset-side fetch; blank `assetAddress` is ignored; KDoc spells out all four conditions that produce `null`.
- Tests: added `IOException` swallowing, malformed amounts, blank asset address, units=0-on-rune fallback, units=0-on-both-sides; tightened the rune-side success path to strictly verify call counts; asset-fallback test now asserts the `pool` field too.

**Deferred (with rationale on the inline threads)**
- Cross-file dedup with `GetThorChainLpPositionsUseCase.fetchPosition` — collapsing them cleanly requires restructuring the plural's API to add an APR overlay; bigger than the perf fix this PR is scoped to. Happy to backport the parsing hardening into the plural here if you'd prefer.
- Focused `selectDepositOption(RemoveLiquidity)` VM test with a non-relaxed mock — the use case path is well covered by its own unit tests; can add the VM-level test as a follow-up so this PR stays scoped.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
```diff
@@ -645,84 +646,90 @@ constructor(
...
+                        UiText.FormattedText(
+                            R.string.remove_pool_amount_format,
+                            listOf(userRune, symbol),
+                        )
+                    } else UiText.Empty
+                _state.update {
+                    it.copy(
+                        availableLpUnits = userUnits.toString(),
+                        removeLpUnitsDivisor = userUnits,
+                        removeLpPoolDepth = runeRedeemBase,
```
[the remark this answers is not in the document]
aminsato: Verified and fixed in 8f3e5abe3 — `loadLpJob?.cancel()` now runs unconditionally at the top of `selectDepositOption`, so any transition (Remove LP → Bond/AddLiquidity/Switch/etc.) tears down the in-flight fetcher before its post-suspension `_state.update` can land. Re-entering Remove LP is unaffected because `loadThorChainRemoveLpData` reassigns `loadLpJob`.

Skipped the suggested per-update "check option is RemoveLiquidity before `_state.update`" guard because the cancellation is sufficient: `selectDepositOption` and the in-flight load both run on `Dispatchers.Main`, so once `cancel()` fires the in-flight coroutine throws `CancellationException` at its next suspension resume and never reaches `_state.update`. Happy to add the explicit gate too if you'd rather have belt-and-suspenders.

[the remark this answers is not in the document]
aminsato: Good catch — verified, the bug is real: `availableLpUnits` stores the full BigInteger but `setRemoveLpPercent` reads it back via `toLongOrNull()` and silently no-ops, while `balanceText` falls through to `UiText.Empty` for whales. Option A is the right answer.

Punting to a follow-up though: this is pre-existing from #4320 and orthogonal to the perf fix this PR is scoped to. Tracked in #4341 — happy to take it next.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
```diff
@@ -864,6 +871,8 @@ constructor(
     }
 
     fun selectDepositOption(option: DepositOption) {
+        // Stop any in-flight Remove LP fetch so it can't write stale state into the new option.
+        loadLpJob?.cancel()
```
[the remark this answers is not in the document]
aminsato: Verified and fixed in a07db2a4d — `loadRemoveLpData` now also assigns to `loadLpJob` and cancels-before-relaunch, so the MayaChain path gets the same protection from the unconditional `loadLpJob?.cancel()` in `selectDepositOption`.

aminsato:
@coderabbitai full review

johnnyluo (approved):
LGTM

PR #4342, opened 2026-05-02
fix(evm): floor ERC-20 deposit gas at 200k for THORChain router calls

### LINKED ISSUE #4152: THORChain LP add (ETH USDT) reverts: gas limit too low for depositWithExpiry on non-standard ERC-20s
# [Fix] THORChain router ERC-20 deposits use hardcoded gas limit [THORChain LP Add]

## Problem

THORChain LP-add transactions for ETH.USDT (and other non-standard ERC-20s like USDC) submitted from Android revert on-chain with no error string. The wallet hardcodes a 120,000 gas limit for `THORChain_Router.depositWithExpiry(...)`, which is insufficient for USDT's non-standard `transferFrom` (two SSTOREs, allowance decrement), causing an out-of-gas inside the router's low-level `.call()`. The failing tx used 118,863 / 120,000 gas (99.05%), the classic fingerprint of this OOG pattern. ETH and well-behaved ERC-20s happen to fit under the cap, masking the bug until now.

## Expected Behavior

Submitting a THORChain LP-add for ETH.USDT (or any non-standard ERC-20 via the THORChain router) should not revert due to the wallet's own gas limit. The gas limit should be derived from the populated transaction via `eth_estimateGas`, not a compile-time constant.

## Steps to Reproduce

1. Android app — navigate to THORChain → Add LP → ETH.USDT
2. Submit any amount at or above the minimum deposit threshold
3. Observe on Etherscan: `depositWithExpiry` fails with "execution reverted" (no reason string), ~99% gas consumed

### Must Do
- [ ] Replace the hardcoded `120_000` gas limit for `depositWithExpiry` router calls with a dynamic estimate: call `eth_estimateGas` on the fully populated transaction and multiply by 1.3–1.5×
- [ ] Apply a minimum floor of `200_000` gas as a safety net for any ERC-20 router deposit (i.e. `max(estimate * 1.3, 200_000)`), guarding against estimation failures
- [ ] Ensure the fix covers all EVM chains that use the same router pattern (Ethereum mainnet, Base, Avalanche, BSC)

### Must NOT Do
- Don't change unrelated gas logic for ETH (native) deposits or non-THORChain transactions
- Don't add per-token special-casing beyond the floor — the dynamic estimate is the general solution

## Acceptance Criteria
- [ ] A THORChain LP-add of ETH.USDT (≥ min deposit) from Android confirms on-chain; gas used is well below the limit
- [ ] USDC on ETH exercised through the same path also confirms on-chain
- [ ] ETH (native) THORChain deposits continue to work as before

## Context

- **Failing tx:** [`0xc0e0fe3b0a0ba22ade75638512c3513edd4d328bcc5510d180cd4b66bb66872a`](https://etherscan.io/tx/0xc0e0fe3b0a0ba22ade75638512c3513edd4d328bcc5510d180cd4b66bb66872a)
- **Block time:** Apr-22-2026 11:03:47 UTC
- **Router contract (v4.1.1):** [`0xD37BbE5744D730a1d98d8DC97c42F0Ca46aD7146`](https://etherscan.io/address/0xD37BbE5744D730a1d98d8DC97c42F0Ca46aD7146#code)
- **Asset (USDT):** `0xdac17f958d2ee523a2206206994597c13d831ec7`
- **Decoded call:** `depositWithExpiry(vault=0x57bb04f3…, asset=USDT, amount=300000 (0.30 USDT), memo="+:ETH.USDT-0XDAC…:thor12a9rpf9u2ul…" (97 bytes), expiration=1776856623)`
- **Gas:** 118,863 used / 120,000 limit — 99.05% consumed, no revert reason string
- **Root cause path in router:** `depositWithExpiry → _deposit → safeTransferFrom` — the bare `require(success && ...)` fires when the low-level `.call()` to USDT OOGs; 63/64 gas forwarding means the sub-call doesn't get enough gas to complete USDT's two cold SSTOREs
- **User impact:** Every THORChain LP-add or swap-in using USDT/USDC on ETH from Android burns the full gas fee (~0.0004 ETH ≈ $1/attempt) with a confusing "execution reverted" and no explanation
- **Reference:** [THORChain Sending Transactions guide](https://dev.thorchain.org/concepts/sending-transactions.html) recommends dynamic gas estimation for router calls
- **Platform:** Android
- **Date observed:** 2026-04-22

aminsato:
## Summary

- Adds a 200k gas-limit floor for non-native EVM deposits so THORChain router calls (`depositWithExpiry`) aren't sent under-gassed.
- Existing path estimates against `transfer()` and falls back to a 150k default — both underestimate USDT/USDC router deposits, which need ~120-140k. A 0.30 USDT LP add reverted on-chain with no reason string and consumed the full gas (#4152).
- Floor applies only when `isDeposit && !token.isNativeToken` on the standard EVM path. Native ETH, ZkSync, swaps (already at 600k), and explicit `gasLimit` overrides are untouched.

## Why a floor and not `eth_estimateGas`

This PR is the safety net — it stops users losing gas immediately. The proper fix (estimate against the populated `depositWithExpiry` calldata × 1.3) requires threading calldata earlier in the pipeline and is worth its own PR.

Closes #4152

## Test plan

- [x] `./gradlew :data:testDebugUnitTest --tests BlockChainSpecificRepositoryImplTest` — all 9 tests green, including 3 new tests:
  - `ERC20 deposit gas limit is floored to safety minimum` — estimate (75k) and default (150k) below floor → result is 200k
  - `ERC20 deposit keeps higher estimate when above safety floor` — estimate (450k) above floor → result is 450k (floor doesn't clamp down)
  - `native ETH deposit is not affected by ERC20 deposit floor` — native token deposit with low estimate is unchanged
- [x] `./gradlew ktfmtFormat` — clean
- [ ] Manual: USDT THORChain LP add on a test vault confirms on-chain (deferred — verifying which UI path produces the failing tx; happy to follow up)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gas fee calculation for non-native (ERC‑20) token deposits to prevent failures from underestimated gas limits; native token transfers remain unaffected.
* **Tests**
  * Added coverage verifying ERC‑20 deposit gas is clamped to a safe minimum while larger estimates are preserved, and native deposits are unchanged.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
@@ -233,6 +233,11 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
 
         val DEFAULT_ARBITRUM_TRANSFER = "160000".toBigInteger()
         val DEFAULT_MANTLE_SWAP_LIMIT = "3000000000".toBigInteger()
+
+        // Floor for non-native ERC-20 deposits that go through a router contract
+        // (e.g. THORChain `depositWithExpiry`). Estimating against `transfer()` underestimates
+        // the real call, and a USDT deposit at 120k reverts with no reason string. See #4152.
+        val MIN_ERC20_DEPOSIT_GAS_LIMIT = "200000".toBigInteger()
     }
 }
 
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
@@ -20,6 +20,7 @@ import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion
 import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_COIN_TRANSFER_LIMIT
 import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_SWAP_LIMIT
 import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_TOKEN_TRANSFER_LIMIT
+import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.MIN_ERC20_DEPOSIT_GAS_LIMIT
 import com.vultisig.wallet.data.blockchain.model.Eip1559
 import com.vultisig.wallet.data.blockchain.model.GasFees
 import com.vultisig.wallet.data.blockchain.model.Swap
@@ -195,7 +196,15 @@ constructor(
 
                     val nonce = evmApi.getNonce(address)
 
-                    val gasLimitFee = gasLimit ?: max(defaultGasLimit, estimateGasLimit)
+                    val gasLimitFee =
+                        gasLimit
+                            ?: max(defaultGasLimit, estimateGasLimit).let { base ->
+                                if (isDeposit && !token.isNativeToken) {
+                                    max(base, MIN_ERC20_DEPOSIT_GAS_LIMIT)
+                                } else {
+                                    base
+                                }
+                            }
                     val fees =
                         if (isSwap) {
                             feeServiceComposite.calculateDefaultFees(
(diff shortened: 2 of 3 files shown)


aminsato:
@coderabbitai full review

rkokhatskyi (changes requested):
@aminsato this is not the right approach, we should not hardcode it. Please proceed with the proper implementation for `eth_estimateGas` and simulation

aminsato:
> @aminsato this is not the right approach, we should not hardcode it. Please proceed with the proper implementation for `eth_estimateGas` and simulation

@rkokhatskyi 
I’ll close the PR and leave the issue open. Let’s ask @johnnyluo opinion on this

PR #4353, opened 2026-05-02
fix(swap): hide Provider/fees block when quote error is shown (#4347)

### LINKED ISSUE #4347: Swap screen: Provider field is blank when a quote error occurs, but other fee fields are still populated
## Description
On the Swap screen, when a quote error occurs (e.g. "Swap amount too small. Recommended amount …"), the **Provider** row is rendered with no value, while other fields below it — **Total Fee**, **Network Fee**, and **Swap Fee** — are still populated.

The fee values are no longer meaningful in this state since no valid quote exists, and showing them while the Provider is blank is inconsistent.

## Steps to reproduce
1. Open Swap.
2. From: `Tron` / `USDT`.
3. To: `Tron` / `TRX`.
4. Enter an amount below the minimum (e.g. `6.055066 USDT`).
5. Observe the error tooltip and the fields below.

## Actual
- Provider row is empty (label shown, value missing).
- Total Fee, Network Fee, and Swap Fee are still displayed with values.

## Expected
- When a swap quote error is shown, either:
  - all quote-derived fields (Provider, Total Fee, Network Fee, Swap Fee) are hidden / cleared together, or
  - the Provider field shows a value consistent with the rest.

The state should be internally consistent — either populated or cleared, not partially shown.

## Screenshot
swap-error

aminsato:
## Summary
- When a swap quote error occurs (e.g. *"Swap amount too small. Recommended amount …"*) the **Provider** row went blank while **Total Fee**, **Network Fee** and **Swap Fee** continued to render values from the previous successful quote / gas calc — the UI was internally inconsistent.
- Wrap the Provider / Total Fee / Network Fee / Swap Fee `Column` in `AnimatedVisibility(visible = error == null || state.isLoading)` so the entire block is hidden together while an error is on screen. The block still renders during loading (placeholders show progress) and reappears once a valid quote returns.
- The error tooltip near the flip button already surfaces the actual quote error, so no information is lost.

## Why hide instead of clear
Clearing to *"$0.00 / 0"* would still mislead the user — there is no swap at all, not a free one. Hiding makes the absence of a quote unambiguous, and leaves the underlying `networkFee` (gas) state intact for the next successful quote.

Closes #4347

## Test plan
- [ ] Open Swap → from `Tron / USDT`, to `Tron / TRX`, enter `6.055066` (below minimum).
- [ ] Wait for the quote error tooltip to appear near the flip button.
- [ ] Verify the Provider / Total Fee / Network Fee / Swap Fee block is no longer rendered while the error is shown.
- [ ] Increase the amount to a valid value → block reappears once a new quote returns.
- [ ] Sanity-check non-error path: normal swap flow renders Provider + fees and the expandable Total Fee dropdown works as before.
- [ ] During the loading window after typing a new amount, placeholders should still appear (block is visible while `state.isLoading == true`).

> ⚠️ No emulator was attached at implementation time, so before/after screenshots were not captured. Reviewer please verify on device.

## Summary by CodeRabbit

* **Bug Fixes**
  * Provider and total-fee area now only appears while loading or when a valid quote exists, preventing placeholder noise. Fee row remains tappable to expand details.

* **Bug Fix / UI State**
  * Fee/quote failures now fully reset displayed fee, discount and tier values and disable swapping; successful quotes recover the UI and re-enable swapping.

* **Tests**
  * Expanded tests cover full state reset on errors and recovery on subsequent successful quotes.


aminsato:
@coderabbitai full review

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
```diff
@@ -393,157 +393,166 @@ internal fun SwapScreen(
...
-                                } else {
-                                    {
-                                        Row {
-                                            Text(
-                                                text = state.totalFee,
-                                                color = Theme.v2.colors.text.secondary,
-                                                style = Theme.brockmann.supplementary.caption,
-                                                textAlign = TextAlign.End,
-                                            )
+                    AnimatedVisibility(visible = error == null || state.isLoading) {
```
rkokhatskyi: The `error` local at line 225 folds `state.error` into the gate. But `state.error` is set by `validateAmount` and `swap()` failure paths and `calculateGas` errors none of which invalidate the existing quote. It is also sticky since `calculateFees` success at line 945 never clears it so after a Swap click that fired `R.string.swap_error_insufficient_balance_and_fees` the fee block stays hidden through the next valid quote until the `HintBox` is dismissed

rkokhatskyi: also the `SwapException` catch at `SwapFormViewModel.kt:963` resets `provider` `srcFiatValue` `estimatedDstTokenValue` `estimatedDstFiatValue` `fee` and `expiredAt`. It leaves `networkFee` `networkFeeFiat` `totalFee` `vultBpsDiscount` `vultBpsDiscountFiatValue` `referralBpsDiscount` `referralBpsDiscountFiatValue` and `tierType` from the previous quote untouched. The hide masks them while the error shows but a refresh that flips `isLoading` true will surface those stale values for the debounce window since none of those rows have a placeholder branch

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
```diff
@@ -393,157 +393,166 @@ internal fun SwapScreen(
...
+                        Column(
+                            verticalArrangement = Arrangement.spacedBy(10.dp),
+                            modifier = Modifier.padding(horizontal = 8.dp),
+                        ) {
+                            val placeHolderModifier = Modifier.height(16.dp).width(80.dp)
+                            FormDetails2(
+                                title = stringResource(R.string.swap_screen_provider_title),
+                                value = state.provider.asString(),
+                            )
```
rkokhatskyi, line 409: This Provider row has no `if (state.isLoading) UiPlaceholderLoader` branch. The `totalFee` row at line 414 and the inner gas and fee rows at lines 490 and 502 all do so during a refresh `state.provider.asString()` keeps rendering the previous quote's provider name. The same gap exists on the `vultBpsDiscount` row at line 507 and the `referralBpsDiscount` row at line 525 which read state fields that are never reset on quote failure

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
rkokhatskyi: also iOS has the same defect in `SwapDetailsSummary.swift` of `vultisig-ios`. The Provider row gates on `tx.quote?.displayName` separately from `swapViewModel.showTotalFees` and `SwapErrorTooltipView` overlays the still populated fee rows. This PR creates parity rather than restoring it so a mirror change on iOS would close the gap

rkokhatskyi: also the test `calculateFees resets state on swap exception` at `SwapFormViewModelTest.kt:937` only asserts the eight fields the catch currently resets. The stale ones (`networkFee` `networkFeeFiat` `totalFee` `vultBpsDiscount` `referralBpsDiscount` `tierType`) are uncovered. The underlying staleness is verifiable from the ViewModel alone without any UI test

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
aminsato: Right — `state.error` is set by `swap()`/`calculateGas` paths (not quote errors) and `calculateFees` success never clears it. Switched the gate at line 396 to `state.formError == null || state.isLoading` so only quote-derived errors hide the fee block. The HintBox/visual cues still consume both errors via the `error` local. (a02b6f2)

aminsato: Good catch. Extended the `SwapException` catch in a02b6f2 to also reset `networkFee`, `networkFeeFiat`, `totalFee`, `vultBpsDiscount`, `vultBpsDiscountFiatValue`, `referralBpsDiscount`, `referralBpsDiscountFiatValue`, and `tierType`, so a subsequent refresh that flips `isLoading=true` cannot surface stale prior-quote values.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
aminsato: Added a `UiPlaceholderLoader` placeholder for the Provider row in a02b6f2, matching `totalFee`/gas/fee. The `vultBpsDiscount` (line 507) and `referralBpsDiscount` (line 525) rows are inside the expandable `AnimatedVisibility(visible = isFeeDetailsExpanded && !state.isLoading)` at line 438, so they don't render during a refresh — the staleness path for them is the post-error case, which is closed by the catch-block reset above.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
aminsato: Acknowledged — that's a separate iOS PR. I'll open a mirror change in `vultisig-ios` against `SwapDetailsSummary.swift` after this lands.

aminsato: Extended `calculateFees resets state on swap exception` (SwapFormViewModelTest.kt:937) in a02b6f2 to assert all newly-reset fields (`networkFee`, `networkFeeFiat`, `totalFee`, `vultBpsDiscount`, `vultBpsDiscountFiatValue`, `referralBpsDiscount`, `referralBpsDiscountFiatValue`, `tierType`).

rkokhatskyi: @aminsato please create the ticket for iOS and link it here

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -975,6 +975,14 @@ constructor(
...
                                 estimatedDstFiatValue = "0",
                                 fee = "0",
+                                networkFee = "",
+                                networkFeeFiat = "",
+                                totalFee = "0",
+                                vultBpsDiscount = null,
+                                vultBpsDiscountFiatValue = null,
+                                referralBpsDiscount = null,
+                                referralBpsDiscountFiatValue = null,
+                                tierType = null,
```
rkokhatskyi: Setting `quote = null` and `expiredAt = null` here does not cancel `refreshQuoteJob` from the prior successful quote, so its delayed `refreshQuoteState++` still fires and re-runs `calculateFees` against the same invalid amount, which surfaces another `SwapException` while the gate at `SwapScreen.kt:396` flashes the fee block visible through its `isLoading` branch during the 450ms debounce, since `launchRefreshQuoteTimer` only cancels the prior job when it is itself called and that path at line 999 is gated on a non-null quote

rkokhatskyi: After this PR expanded the `SwapException` catch to clear 13 fields, the sibling `catch (e: Exception)` at line 993 still only sets `quote = null` and `isLoading = false`, so a network IO failure or a JSON deserialization error from `swapQuoteRepository` or `swapQuoteManager.fetchBestQuote` keeps the prior quote's `provider`, `srcFiatValue`, `estimatedDstTokenValue`, `estimatedDstFiatValue`, `fee`, `networkFee` and `totalFee` populated whereas `formError` stays null so the gate at `SwapScreen.kt:396` keeps the fee block on screen with a stale quote and no error indicator. That same catch also captures `CancellationException` which defeats `safeLaunch`'s rethrow at `SafeLaunch.kt:45-46`, so on screen exit when `viewModelScope` is cancelled the body still runs `uiState.update` on a cancelled context and `Timber.e(e)` logs the cancel as an unexpected error

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -975,6 +975,14 @@ constructor(
                                 estimatedDstTokenValue = "0",
                                 estimatedDstFiatValue = "0",
                                 fee = "0",
+                                networkFee = "",
+                                networkFeeFiat = "",
+                                totalFee = "0",
```
rkokhatskyi: This wipes `state.totalFee` to `0` whereas the upstream `swapFeeFiat` at line 151 and `estimatedNetworkFeeFiatValue` at line 152 still hold the prior quote's values, so when the user changes `selectedSrcToken` while the error is on screen `calculateGas` synchronously bumps `estimatedNetworkFeeFiatValue.value` which fires `collectTotalFee` at line 795 with no `formError` filter and writes a `newGas + staleSwap` combination back to `state.totalFee` even before `calculateFees` re-completes, leaving that stale combination in place if the user dismisses the `HintBox` next

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
```diff
@@ -393,157 +393,170 @@ internal fun SwapScreen(
...
-                                } else {
-                                    {
-                                        Row {
-                                            Text(
-                                                text = state.totalFee,
-                                                color = Theme.v2.colors.text.secondary,
-                                                style = Theme.brockmann.supplementary.caption,
-                                                textAlign = TextAlign.End,
-                                            )
+                    AnimatedVisibility(visible = state.formError == null || state.isLoading) {
```
rkokhatskyi: Dismissing the `HintBox` runs `hideError` at `SwapFormViewModel.kt:1015-1017` which sets `formError = null`, so this gate flips back to visible while the 13 fee fields are still at their catch-cleared empty or zero values, exposing an empty `Provider` row, a `0` `Total Fees` and a blank gas line until the user types a new amount that re-runs `calculateFees`

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -965,6 +965,14 @@ internal class SwapFormViewModelTest {
...
             assertEquals("0", state.estimatedDstFiatValue)
             assertEquals("0", state.fee)
+            assertEquals("", state.networkFee)
+            assertEquals("", state.networkFeeFiat)
+            assertEquals("0", state.totalFee)
+            assertNull(state.vultBpsDiscount)
+            assertNull(state.vultBpsDiscountFiatValue)
+            assertNull(state.referralBpsDiscount)
+            assertNull(state.referralBpsDiscountFiatValue)
+            assertNull(state.tierType)
```
rkokhatskyi, line 992: The new assertions cover only the catch path, leaving no test for the recovery flow where the user enters a valid amount after a `SwapException` and the success path at `SwapFormViewModel.kt:938-950` runs, which is where `networkFee` and `networkFeeFiat` should be repopulated and currently are not, so the regression that CodeRabbit flagged at line 980 is invisible at the ViewModel level

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
aminsato: Done — opened vultisig/vultisig-ios#4295 to track the parity change against `SwapDetailsSummary.swift`.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in a72133dd2. Centralized the SwapException-catch reset in a new `resetQuoteState()` helper which now does `refreshQuoteJob?.cancel(); refreshQuoteJob = null` before clearing quote state. The prior quote's delayed `refreshQuoteState++` can no longer fire mid-error and bounce `calculateFees` against the same invalid amount.

aminsato: Fixed in a72133dd2. The generic catch now (a) rethrows `CancellationException` so `safeLaunch`'s rethrow at `SafeLaunch.kt:45-46` isn't defeated on screen exit, and (b) routes through the same `resetQuoteState()` helper as the SwapException branch — so a network IO / JSON deserialization failure clears `provider` / `srcFiatValue` / `estimatedDst*` / `fee` / discounts / `tierType` / `hasQuote` and surfaces a `swap_error_quote_failed` formError instead of leaving the prior quote populated with no error indicator. New test `calculateFees on generic exception sets quote-failed formError and clears hasQuote` covers it.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in a72133dd2. `resetQuoteState()` now sets `swapFeeFiat.value = null`, which short-circuits the `filterNotNull()` chain in `collectTotalFee` (line 795). After that, even if `calculateGas` synchronously bumps `estimatedNetworkFeeFiatValue` while the error is on screen, the `combine` block can't emit a `(newGas + staleSwap)` totalFee. We also stopped wiping `networkFee` / `networkFeeFiat` / `totalFee` directly in the catch — they're tied to the source token (calculateGas), not to a specific quote, and the success path doesn't repopulate them. `hasQuote`-based visibility (see other thread) hides them visually instead.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
aminsato: Fixed in a72133dd2. Added `hasQuote: Boolean` to `SwapFormUiModel`, set to `true` only on the success path and `false` in `resetQuoteState()`. The visibility gate in `SwapScreen.kt:396` is now `state.isLoading || (state.formError == null && state.hasQuote)`. So when the user dismisses the `HintBox` and `hideError()` clears `formError`, the block stays hidden because `hasQuote` is still `false` — no more empty Provider / `0` Total Fees / blank gas line surfacing. The block reappears only when the next `calculateFees` succeeds (or while loading).

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Added in a72133dd2 — new test `calculateFees recovers hasQuote on success after a swap exception` stubs `fetchBestQuote` with `throws SwapException.SwapIsNotSupported andThen createDefaultQuoteFetchResult()`, types `0.1`, asserts `hasQuote = false` and `formError != null`, then types `0.2`, asserts `hasQuote = true`, `formError = null`, `isSwapDisabled = false`, `expiredAt != null`. That locks the recovery path the CodeRabbit comment was flagging.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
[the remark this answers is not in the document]
aminsato: Addressed in a72133dd2 by going with the first option from the suggestion: `networkFee` / `networkFeeFiat` are no longer reset in the SwapException catch. They're tied to the source token (populated by `calculateGas` which only fires on `selectedSrc` changes), not to a specific quote, so wiping them in a quote-error catch was unsafe. The new `hasQuote`-gated visibility means the fee block stays hidden until the next valid quote anyway, so leaving the source-token gas fields intact is the right semantic.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
```diff
@@ -393,157 +393,172 @@ internal fun SwapScreen(
                         }
                     }
 
-                    Column(
-                        verticalArrangement = Arrangement.spacedBy(10.dp),
-                        modifier = Modifier.padding(horizontal = 8.dp),
+                    AnimatedVisibility(
+                        visible = state.isLoading || (state.formError == null && state.hasQuote)
```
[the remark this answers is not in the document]
aminsato: Good catch — fixed in 21a6f9a1a. `validateBalanceForSwap` runs after the quote-success state update with `hasQuote = true`, so the gate would have hidden the block on every "insufficient balance" error. Dropped `formError` from the gate entirely; it's now just `state.isLoading || state.hasQuote`. Behavior:
- Quote error → `resetQuoteState` sets `hasQuote = false` → block hides (Roman's original ask still satisfied).
- HintBox dismiss → `formError` cleared but `hasQuote` still false → block stays hidden, no empty Provider exposure.
- Post-quote balance error → `hasQuote = true`, `formError` set → block stays visible so the user sees the fees they need to cover. ✅
- Loading → placeholders.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -1012,6 +1008,42 @@ constructor(
             }
     }
 
+    private fun resetQuoteState(error: UiText, cause: Throwable, tag: String) {
```
rkokhatskyi: This new overload cancels `refreshQuoteJob`, nulls `swapFeeFiat.value`, and sets `hasQuote = false`, whereas the no-args sibling at line 690 used by `flipSelectedTokens` only clears 4 UI fields and leaves `hasQuote = true`, `swapFeeFiat.value` populated, and `refreshQuoteJob` alive, so after a flip the new gate `state.isLoading || state.hasQuote` keeps the fee block on screen for the ~450ms debounce while the prior pair's `provider`, `vultBpsDiscount`, `referralBpsDiscount` keep rendering and `collectTotalFee` writes a `(newGas + staleSwap)` `totalFee` back via the synchronous `calculateGas` update, exactly the race this overload was added to prevent

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -979,11 +979,98 @@ internal class SwapFormViewModelTest {
...
+            Snapshot.sendApplyNotifications()
+            advanceTimeBy(500)
+            advanceUntilIdle()
+
+            val state = vm.uiState.value
+            assertTrue(state.hasQuote)
+            assertNull(state.formError)
+            assertFalse(state.isSwapDisabled)
+            assertFalse(state.isLoading)
+            assertNotNull(state.expiredAt)
```
rkokhatskyi, line 1047: The recovery assertions cover `hasQuote`, `formError`, `isSwapDisabled`, `isLoading` and `expiredAt` but not `networkFee` or `networkFeeFiat`, which is exactly the field CodeRabbit originally flagged and which the commit message claims now survives the error to success transition since they are tied to the source token, so a regression that re-introduces the clearing in `resetQuoteState` at line 1011 would not fail this test, whereas asserting `assertEquals("0.001 ETH", state.networkFee)` against the gas mock at line 135 would lock the contract

rkokhatskyi (changes requested):

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Fixed in b396a28e3. Added `assertEquals("0.001 ETH", state.networkFee)` and `assertEquals("$2.00", state.networkFeeFiat)` (against the gas mock at line 135) to the recovery test, so a regression that re-introduces clearing in `resetQuoteState` will fail there. Also added the same two assertions to `calculateFees resets state on swap exception` to lock the contract on the error-path side — both halves of the error→success transition are now covered at the ViewModel level.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in b396a28e3. Collapsed the two `resetQuoteState` overloads into one with nullable `error`/`cause`/`tag`, so the no-arg sibling used by `flipSelectedTokens` now also:

- cancels `refreshQuoteJob` (and nulls it),
- nulls `swapFeeFiat.value` so `collectTotalFee`'s `filterNotNull()` short-circuits and `calculateGas` can't write a `(newGas + staleSwap)` `totalFee` back during the 450ms debounce,
- clears `hasQuote`, `provider` (and `vultBpsDiscount` / `referralBpsDiscount` / `tierType`) so the gate `state.isLoading || state.hasQuote` hides the block until the next quote succeeds.

The flip path keeps `formError = null` (no error to surface); the catch paths still set the mapped error. Extended `flipSelectedTokens resets quote state` to assert `hasQuote = false` so a regression that drops the shared cleanup will fail there.

rkokhatskyi (approved):
LGTM

PR #4358, opened 2026-05-02
fix(swap): make flip actually swap src and dst (#4354)

### LINKED ISSUE #4354: Swap: switch button copies source token into destination after proceeding
## Bug

After proceeding on the Swap screen, pressing the switch (swap direction) button incorrectly displays the destination token as the same as the source token, instead of swapping their values.

## Steps to reproduce

1. Open the Swap screen
2. Select a source token and a destination token
3. Proceed on the swap screen (advance past the initial state)
4. Press the switch / swap-direction button
5. Observe the destination token field

## Expected

Source and destination tokens swap with each other.

## Actual

Destination token is shown as the same token as the source (both fields show the same token).

## Screenshot

bug

## Repro video

https://i.imgur.com/iBDvfux.mp4

## Environment

- Android app, Swap flow

aminsato:
## Summary

Fixes #4354 — after a quote loaded, pressing the switch button on the Swap screen did nothing visible: the source and destination tokens stayed where they were instead of swapping.

`flipSelectedTokens()` was swapping the IDs *and* manually swapping the resolved `selectedSrc` / `selectedDst` StateFlows. Under `Dispatchers.Main.immediate` the `collectSelectedTokens()` combine block runs synchronously on the same call stack the moment each ID changes, so by the time the manual swap executes, `selectedSrc` / `selectedDst` already hold the new (post-swap) values. The manual swap then reads those new values and writes them back into their original slots — silently reverting the flip so the UI shows the original pair.

The fix drops the manual swap; the ID swap alone is sufficient because the combine listener resolves both StateFlows from the new IDs.

## Test plan

- [x] New unit test `flipSelectedTokens swaps src and dst instead of leaving them unchanged` reproduces the bug under `UnconfinedTestDispatcher` (mirrors `Main.immediate`) — fails before the fix (`srcAfter` stays as the original src), passes after.
- [x] Full `SwapFormViewModelTest` suite passes.
- [x] ktfmt applied.
- [ ] Manual: open Swap, pick src and dst, enter an amount so a quote loads, tap the switch button → fields swap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of token swapping in the exchange interface, eliminating edge cases that could undo swap operations.

* **Tests**
  * Enhanced test coverage for token selection and swapping behavior to ensure reliability and correctness.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -642,14 +642,14 @@ constructor(
 
         resetQuoteState()
 
-        val bufId = selectedSrcId.value
+        val previousSrcId = selectedSrcId.value
         selectedSrcId.value = selectedDstId.value
-        selectedDstId.value = bufId
+        selectedDstId.value = previousSrcId
 
-        // collectSelectedTokens() observes the IDs above and resolves both StateFlows.
-        // Under Main.immediate it runs on this same call stack, so an additional manual
-        // swap of selectedSrc/selectedDst here reads the already-updated values and
-        // reverts the flip — leaving src and dst pointing at the same token.
+        // collectSelectedTokens() observes the IDs above and resolves selectedSrc/selectedDst
+        // synchronously under Main.immediate. A manual swap of those resolved StateFlows here
+        // would read the already-resolved post-swap values and write them back into their
+        // original slots, silently reverting the flip so the UI shows the original pair.
 
         if (
             newSrcAmount != null &&
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -646,9 +646,10 @@ constructor(
         selectedSrcId.value = selectedDstId.value
         selectedDstId.value = bufId
 
-        val buffer = selectedSrc.value
-        selectedSrc.value = selectedDst.value
-        selectedDst.value = buffer
+        // collectSelectedTokens() observes the IDs above and resolves both StateFlows.
+        // Under Main.immediate it runs on this same call stack, so an additional manual
+        // swap of selectedSrc/selectedDst here reads the already-updated values and
+        // reverts the flip — leaving src and dst pointing at the same token.
 
         if (
             newSrcAmount != null &&
--- a/app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
@@ -508,23 +508,7 @@ internal class SwapFormViewModelTest {
     // region flipSelectedTokens
 
     @Test
-    fun `flipSelectedTokens swaps source and destination token IDs`() =
-        runTest(mainDispatcher) {
-            val addresses = listOf(ethAddress(), btcAddress())
-            val vm = createViewModelWithAddresses(addresses)
-            advanceUntilIdle()
-
-            // The vm should have selected some tokens by now from addresses
-            vm.flipSelectedTokens()
-            advanceUntilIdle()
-
-            // After flip, the state should be reset for fresh calculation
-            assertEquals("0", vm.uiState.value.estimatedDstTokenValue)
-            assertEquals("0", vm.uiState.value.estimatedDstFiatValue)
-        }
-
-    @Test
-    fun `flipSelectedTokens swaps tokens in UI state, not duplicates them`() =
+    fun `flipSelectedTokens swaps src and dst instead of leaving them unchanged`() =
         runTest(mainDispatcher) {
             val vm =
                 createViewModelWithAddresses(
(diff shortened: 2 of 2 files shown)


aminsato:
@coderabbitai full review

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -646,9 +646,10 @@ constructor(
         selectedSrcId.value = selectedDstId.value
         selectedDstId.value = bufId
 
-        val buffer = selectedSrc.value
-        selectedSrc.value = selectedDst.value
-        selectedDst.value = buffer
+        // collectSelectedTokens() observes the IDs above and resolves both StateFlows.
+        // Under Main.immediate it runs on this same call stack, so an additional manual
+        // swap of selectedSrc/selectedDst here reads the already-updated values and
+        // reverts the flip — leaving src and dst pointing at the same token.
```
rkokhatskyi: The "same token" wording is wrong. Reverting the fix and running the new test shows `srcAfter` is `ETH` (the original src), so the OLD-code symptom is "flip silently does nothing", not "src == dst". The duplicate state only shows up briefly mid-call, never in the post-call `uiState`.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -646,9 +646,10 @@ constructor(
         selectedSrcId.value = selectedDstId.value
         selectedDstId.value = bufId
```
rkokhatskyi: `bufId` breaks pattern with `currentSrcText`, `currentSrcTokenId`, `restoredAmount` etc. used elsewhere in this function. Suggest `previousSrcId`.

```kotlin
val previousSrcId = selectedSrcId.value
selectedSrcId.value = selectedDstId.value
selectedDstId.value = previousSrcId
```

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -518,6 +523,61 @@ internal class SwapFormViewModelTest {
             assertEquals("0", vm.uiState.value.estimatedDstFiatValue)
```
rkokhatskyi: The test on line 511 is named `flipSelectedTokens swaps source and destination token IDs`, but the body only asserts `estimatedDstTokenValue` and `estimatedDstFiatValue` are reset — exactly what `flipSelectedTokens resets quote state` below already covers. Either rename it to match what it actually checks, or delete it now that the new test on line 527 is the real swap verifier.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -518,6 +523,61 @@ internal class SwapFormViewModelTest {
...
+            advanceUntilIdle()
+
+            val srcAfter = vm.uiState.value.selectedSrcToken?.model?.account?.token?.id
+            val dstAfter = vm.uiState.value.selectedDstToken?.model?.account?.token?.id
+            assertEquals(BTC_COIN.id, srcAfter, "src should now be BTC")
+            assertEquals(ETH_COIN.id, dstAfter, "dst should now be ETH (not BTC again)")
+        }
+
+    @Test
+    fun `flipSelectedTokens swaps tokens with StandardTestDispatcher`() =
```
rkokhatskyi: This passes on both the OLD and new code, so it doesn't catch the regression. Verified empirically: with the manual swap restored, this assertion still passes while the `UnconfinedTestDispatcher` test on line 527 fails with `expected: BTC but was: ETH`. Under `StandardTestDispatcher`, ID writes stay queued until `advanceUntilIdle`, and the combine then drains both StateFlows from the swapped IDs regardless of whether the manual swap ran. Either remove it or rewrite it to assert before `advanceUntilIdle`.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -518,6 +523,61 @@ internal class SwapFormViewModelTest {
...
+        runTest(kotlinx.coroutines.test.StandardTestDispatcher(scheduler)) {
```
rkokhatskyi: `StandardTestDispatcher` is used inline with the FQN twice on this line and the next, while the file already imports `UnconfinedTestDispatcher`, `runTest`, `setMain` etc. from this package. Add an import for consistency.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -518,6 +523,61 @@ internal class SwapFormViewModelTest {
...
+            Dispatchers.setMain(kotlinx.coroutines.test.StandardTestDispatcher(scheduler))
```
rkokhatskyi: `setUp` already calls `setMain(mainDispatcher)` on line 96, and there's no `resetMain` between then and this call, so the previously installed dispatcher is silently replaced. Also, a fresh `StandardTestDispatcher` is created here that's distinct from the one passed to `runTest` on the line above; they share a scheduler, but the instances differ.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Good catch — you're right, the symptom is 'flip silently does nothing' (manual swap reads the post-resolve values and writes them back into the original slots). Updated the inline comment, the test name, and the PR title/description to reflect that. (50c01c4)

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Renamed to `previousSrcId` in 50c01c4.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Agreed — deleted in 50c01c4. `flipSelectedTokens resets quote state` already covers the quote-reset assertion, and the renamed test below (now `flipSelectedTokens swaps src and dst instead of leaving them unchanged`) is the real swap verifier.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Removed in 50c01c4. The `UnconfinedTestDispatcher` test (which mirrors `Main.immediate` from production) is the one that actually catches the regression, and it stays.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Moot — the `StandardTestDispatcher` test was removed in 50c01c4, so no FQN usages remain.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Moot — that test was removed in 50c01c4, so the conflicting `setMain` call and the duplicate dispatcher instance are gone with it.

rkokhatskyi (approved):
LGTM

PR #4374, opened 2026-05-03
fix(notifications): dismiss foreground keysign banner when joining via QR

### LINKED ISSUE #4361: Foreground keysign banner stays open after user joins keysign via QR — should auto-dismiss on entering JoinKeysign/JoinKeygen
## Symptom

When a keysign push notification arrives **while the app is in the foreground**, the in-app banner is shown ("Join transaction — <vault> — Send 0.01 SOL"). If the user then joins the same keysign by scanning the QR code (i.e. without tapping the banner), the banner remains visible on top of the JoinKeysign / "Waiting for Keysign to start" screen.

Expected: the banner should be dismissed once the user has navigated into JoinKeysign / JoinKeygen by any means.

Screenshot: banner overlaying "Waiting for Keysign to start" screen after QR scan join.

## Where it comes from

The banner's lifecycle today, in `MainViewModel`:

- `onForegroundPushReceived(qrCodeData)` sets `_foregroundNotification` (`MainViewModel.kt:121`).
- `_foregroundNotification` is cleared **only** by `onForegroundBannerTapped()` (`MainViewModel.kt:151`). No other code path clears it.

History:
- #3310 (`f9199a2a5`, Mar 12) — banner introduced with a 30 s auto-dismiss + an `X` dismiss button.
- #3628 / #3639 (`45d9536b8`, Mar 21) — `fix: correct in-app notification banner design`. The PR title says "design", but the diff also **removed** the 30 s auto-dismiss timer and the `X` dismiss button (`onDismiss` callback dropped from the composable signature). Net result: the banner has only one dismiss path — tapping it.

QR scanning into the keysign join screen does not flow through the banner state, so the banner state is never cleared. Same hazard exists for any path into JoinKeysign / JoinKeygen that doesn't go via `onForegroundBannerTapped` (deeplink, pasted QR, share-sheet open, etc.).

## Repro

1. Foreground the app on a vault.
2. Trigger a keysign that targets that vault from another device — Android receives the FCM and shows the in-app banner.
3. Without tapping the banner, scan the keysign QR shown on the initiator device.
4. Observe: the JoinKeysign "Waiting for Keysign to start" screen renders **with the banner still pinned at the top**.

## Suggested fix

Treat banner dismissal as a function of navigation, not just user-tap. Two options, lowest-risk first:

1. **Clear on entering any keysign-route**. In `MainActivityContent.kt` (or in `MainViewModel`), observe `navController.currentBackStackEntryFlow` (or the existing `navigator.destination` / `navigator.route` flows) and call `_foregroundNotification.value = null` when the destination matches a keysign/keygen-join route.
2. **Clear on any navigation change away from Home**. Simpler but less surgical — would also dismiss if the user opens settings, etc., which is arguably also the right behaviour for a "transient action banner".

While in there, consider:
- Restoring an explicit dismiss affordance (the `X` was deliberately removed in #3628 — confirm with design that the only-tap-to-dismiss model is intentional, given this regression).
- Auto-clearing the banner when the underlying keysign session reaches a terminal state (signed / failed / cancelled), so it doesn't survive across the next foreground.

## Affected version

Reproduces on current `main` (most recent app-side commit on `MainActivityContent.kt` is `d019ae7a3`). The behavioural regression dates back to #3628 / #3639 (Mar 21).

cc @aminsato — author of #3310 and #3628.

---

Filed by an automated triage pass; reproducing repro path is straightforward (foreground + QR-scan-instead-of-banner-tap).

aminsato:
## Summary
- The in-app foreground keysign banner was only cleared by tapping it (`MainViewModel.onForegroundBannerTapped`). Joining the same keysign via QR scan (or any path that bypasses the banner tap) left the banner pinned on top of `JoinKeysign` / `JoinKeygen`.
- Added `MainViewModel.clearForegroundNotification()` and observe `navController.currentBackStackEntryFlow` in `MainActivityContent`. When the destination becomes `Route.Keysign.Join` or `Route.Keygen.Join`, the banner is dismissed regardless of how the user got there.
- Banner dismissal is now a function of navigation, not just user-tap. `onForegroundBannerTapped` is unchanged in behaviour and routes through the same helper for DRY.
- Unit test added to `MainViewModelTest`.

Closes #4361

## Test plan
- [ ] Foreground app on a vault, trigger keysign from another device, banner appears.
- [ ] Without tapping the banner, scan the keysign QR — banner is dismissed when JoinKeysign appears.
- [ ] Tapping the banner still routes to JoinKeysign and dismisses the banner (existing behaviour preserved).
- [ ] Same flow for keygen QR (Route.Keygen.Join).
- [ ] `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.app.activity.MainViewModelTest"` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Foreground notifications are now reliably cleared: pending notification tasks are cancelled when tapping banners or when navigating to specific join screens.

* **Tests**
  * Added a unit test confirming foreground notification state is cleared as expected.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/app/activity/MainViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/MainViewModel.kt
@@ -150,8 +150,13 @@ constructor(
 
     fun onForegroundBannerTapped() {
         val qrCodeData = _foregroundNotification.value?.qrCodeData ?: return
-        _foregroundNotification.value = null
+        clearForegroundNotification()
         onPushNotificationReceived(qrCodeData)
+    }
+
+    fun clearForegroundNotification() {
+        foregroundPushJob?.cancel()
+        _foregroundNotification.value = null
     }
 
     fun onPushNotificationReceived(qrCodeData: String) {
--- a/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
@@ -28,6 +28,7 @@ import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.semantics.testTagsAsResourceId
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.navigation.NavDestination.Companion.hasRoute
 import androidx.navigation.NavHostController
 import com.vultisig.wallet.R
 import com.vultisig.wallet.app.activity.ForegroundNotificationState
@@ -40,6 +41,7 @@ import com.vultisig.wallet.ui.components.banners.OfflineBanner
 import com.vultisig.wallet.ui.components.v2.snackbar.VsSnackBar
 import com.vultisig.wallet.ui.models.AccountUiModel
 import com.vultisig.wallet.ui.models.VaultAccountsUiModel
+import com.vultisig.wallet.ui.navigation.Route
 import com.vultisig.wallet.ui.navigation.SetupNavGraph
 import com.vultisig.wallet.ui.navigation.route
 import com.vultisig.wallet.ui.screens.home.VaultAccountsScreen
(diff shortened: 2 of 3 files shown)


aminsato:
@coderabbitai full review

johnnyluo (approved):
LGTM

johnnyluo:
@aminsato , please fix the tests

aminsato:
@coderabbitai thanks for the suggestion, but Kotest isn't actually a dependency in this project — there's no `kotest` reference in `app/build.gradle.kts`, `data/build.gradle.kts`, or `gradle/libs.versions.toml`. Adopting `io.kotest.matchers.nulls.*` broke `:app:compileDebugUnitTestKotlin` with `Unresolved reference 'kotest'`.

Every other test in `app/src/test/` uses `kotlin.test.assert*` (e.g. `AddressBookViewModelTest`, `DepositFormViewModelTest`, `ImportSeedphraseViewModelTest`), so I've reverted to that style in 367650f18 to match convention and unbreak the build.

If the project ever wires Kotest into the test classpath, this is a good follow-up.

PR #4376, opened 2026-05-03
fix(thorchain): skip /status fetch when no vault uses THORChain

### LINKED ISSUE #4366: App requests THORChain RPC status even when THORChain is disabled in backend
## Summary

When THORChain is disabled by the backend (feature flag / runtime config), the app still issues a `GET` request to the THORChain RPC `/status` endpoint via `gateway.liquify.com`. The call should be skipped entirely when the backend has marked THORChain unavailable.

## Observed

- Endpoint: `GET https://gateway.liquify.com/chain/thorchain_rpc/status`
- Returns the standard tendermint status payload (latest block height, node info, etc.).
- This fires regardless of the backend's THORChain enable state — i.e. the gating that disables THORChain features in the UI does not propagate to this network call.

## Expected

- When THORChain is disabled in backend config, the client should not call `/status` (and ideally any other THORChain-only RPC) at all.
- This is the same pattern as #4321, which skipped the THORChain network-id fetch on startup when no vaults exist — the `/status` call needs equivalent gating against the backend's enable flag.

## Why it matters

- Wasted requests on every relevant trigger (startup, refresh, etc.) — bandwidth + battery.
- Liquify gateway sees traffic for users who shouldn't be hitting THORChain at all.
- If the backend disables THORChain because of an outage, the client keeps probing it instead of failing fast and quiet.

## Acceptance

- With THORChain disabled in backend config, no `GET …/thorchain_rpc/status` request is issued by the client during normal app use (cold start, refresh, navigating to relevant screens).
- With THORChain enabled, behavior is unchanged.
- A network-trace check (Charles / mitmproxy / Logcat OkHttp logs) confirms the request is absent in the disabled state.

aminsato:
## Summary
- `InitializeThorChainNetworkIdUseCase` only gated the THORChain `/status` call on `vaultRepository.hasVaults()` (from #4321). A user with vaults but no THORChain enabled still triggered `GET https://gateway.liquify.com/chain/thorchain_rpc/status` on every cold start.
- Add an additional short-circuit: if no vault has `Chain.ThorChain` in its enabled coins, skip both the cache read and the network fetch. Cached/default `THORCHAIN_NETWORK_ID` is fine because nothing in the app consumes it without a THORChain-enabled vault.

Closes #4366

## Test plan
- [x] `./gradlew :data:testDebugUnitTest --tests "*.InitializeThorChainNetworkIdUseCaseImplTest"` — added "skips when no vault uses THORChain"; all four cases pass.
- [ ] Manual: cold start with a non-THORChain-only vault → confirm no `GET .../thorchain_rpc/status` request fires.
- [ ] Manual: cold start with a THORChain-enabled vault → confirm `/status` still fires and the network id initializes.

## Summary by CodeRabbit

* **New Features**
  * THORChain network ID initialization now waits for a vault containing THORChain and skips initialization when no THORChain coins exist, reducing needless network activity.
  * Initialization is now reactive — it triggers when THORChain vaults appear rather than running once at startup.

* **Tests**
  * Tests updated to cover skip behavior, reactive initialization, fresh-cache fetching, and preserving cached ID when fetch fails.


aminsato:
@coderabbitai full review

aminsato:
@coderabbitai both nitpicks addressed in 88d490f5f:

1. **Single-read refactor** — production code already does this on the latest push (`vaultRepository.getAll()` is read once and used for both the empty check and the THORChain check). No `hasVaults()` call remains in the use case.

2. **Test interaction assertions** — added `coVerify(exactly = 1) { vaultRepository.getAll() }` to the relevant cases. Also dropped the stale `coEvery { vaultRepository.hasVaults() }` setups (production no longer calls `hasVaults()`), and made the no-vaults branch explicit by mocking `getAll() returns emptyList()` instead of relying on the relaxed default.

All four tests still pass: `./gradlew :data:testDebugUnitTest --tests "*.InitializeThorChainNetworkIdUseCaseImplTest"`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/InitializeThorChainNetworkIdUseCase.kt
```diff
@@ -17,11 +18,17 @@ constructor(
 ) : InitializeThorChainNetworkIdUseCase {
 
     override suspend fun invoke() {
-        if (!vaultRepository.hasVaults()) {
+        val vaults = vaultRepository.getAll()
```
rkokhatskyi: `vaultRepository.getAll()` runs five SQL statements under `@Transaction` to hydrate keyshares, signers, coins, and chainPublicKeys for every vault, then for each coin with a blank logo it calls `tokenRepository.getToken()`, whereas the prior `hasVaults()` was a single `SELECT COUNT(*)`. Users with vaults but no THORChain pay this hydration cost on every cold start, which contradicts the early return goal, so a targeted query like `SELECT EXISTS(SELECT 1 FROM coin WHERE chain = :chainId)` on `VaultDao` would gate the init without loading the full vault graph

data/src/test/kotlin/com/vultisig/wallet/data/usecases/InitializeThorChainNetworkIdUseCaseImplTest.kt
```diff
@@ -35,36 +38,66 @@ internal class InitializeThorChainNetworkIdUseCaseImplTest {
...
+
+        useCase()
+
+        coVerify(exactly = 1) { vaultRepository.getAll() }
+        coVerify(exactly = 0) { thorChainRepository.getCachedNetworkChainId() }
+        coVerify(exactly = 0) { thorChainRepository.fetchNetworkChainId() }
+        assertEquals("default-id", ThorChainHelper.THORCHAIN_NETWORK_ID)
+    }
+
+    @Test
```
rkokhatskyi, line 63: Neither cache plus fetch test covers the fresh install branch where `getCachedNetworkChainId()` returns `null` and `fetchNetworkChainId()` must populate the helper, since both existing tests stub the cache with a value, so a sibling test using `null` cache plus a successful fetch would lock that path down

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/usecases/InitializeThorChainNetworkIdUseCase.kt
```diff
@@ -17,11 +18,17 @@ constructor(
...
+        if (vaults.isEmpty()) {
             Timber.d("Skipping THORChain network id init: no vaults")
             return
         }
 
+        if (vaults.none { vault -> vault.coins.any { it.chain == Chain.ThorChain } }) {
```
rkokhatskyi: Before this PR any vault triggered the network id fetch at cold start, whereas now a user whose vaults do not include THORChain who later adds it within the same process by creating a vault, importing a seed, finishing a reshare, or adding the THORChain coin to an existing vault leaves `ThorChainHelper.THORCHAIN_NETWORK_ID` at its default `"thorchain-1"` from `ThorChainHelper.kt:50` until the next launch, since `MainViewModel.init` invokes the use case exactly once and no flow retriggers it, so signing works today only because that constant happens to match the live mainnet chain id

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/usecases/InitializeThorChainNetworkIdUseCase.kt
aminsato: Good call. Switched to a targeted Room query — added `VaultDao.hasCoinOnChain(chainId)` (`SELECT EXISTS(SELECT 1 FROM coin WHERE chain = :chainId)`) and exposed it on `VaultRepository` as `hasAnyCoinOnChain(Chain)`. The use case now gates on that single SQL statement instead of `getAll()`, so non-THORChain users no longer pay the per-vault hydration on cold start. (d2e9b2bc8)

data/src/test/kotlin/com/vultisig/wallet/data/usecases/InitializeThorChainNetworkIdUseCaseImplTest.kt
aminsato: Added — `fetches and applies fresh value when cache is empty (fresh install)` stubs `getCachedNetworkChainId() = null` plus a successful fetch and asserts `THORCHAIN_NETWORK_ID` ends up as the fetched value. (d2e9b2bc8)

data/src/main/kotlin/com/vultisig/wallet/data/usecases/InitializeThorChainNetworkIdUseCase.kt
aminsato: Good catch — fixed. `MainViewModel.init` now collects `vaultRepository.getAllAsFlow()`, maps to "any vault has a Chain.ThorChain coin", `distinctUntilChanged()`, `filter { it }`, `first()` and then invokes the use case. So:
- Cold start with THORChain → fires immediately (first emission already contains it).
- Cold start without THORChain → suspends, no `/status` call.
- User adds THORChain mid-session (new vault import / coin enable) → flow emits a list now containing it, `first()` returns, init runs once, helper is populated with the live id instead of leaving `thorchain-1`. (d2e9b2bc8)

app/src/main/java/com/vultisig/wallet/app/activity/MainViewModel.kt
```diff
@@ -101,7 +104,18 @@ constructor(
...
+            // THORChain coin (or importing a THORChain vault) mid-session still
+            // initializes the live network id instead of relying on the default.
+            vaultRepository
+                .getAllAsFlow()
+                .map { vaults -> vaults.any { it.coins.any { c -> c.chain == Chain.ThorChain } } }
+                .distinctUntilChanged()
+                .filter { it }
+                .first()
+            initializeThorChainNetworkId()
+        }
```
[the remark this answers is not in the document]
aminsato: Good catch — fixed in 7b8d78afd. Added a flow variant of the `EXISTS` query (`VaultDao.observeHasCoinOnChain` → `VaultRepository.observeHasAnyCoinOnChain`) and switched MainViewModel to it. The cold-start path is now a single boolean SQL row that Room re-emits on `coin`-table changes — no `Vault.toVault()` hydration, no per-coin `tokenRepository.getToken()` resolution, regardless of how many vaults the user has.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/VaultRepository.kt
```diff
@@ -90,6 +92,9 @@ constructor(private val vaultDao: VaultDao, private val tokenRepository: TokenRe
 
     override suspend fun hasVaults(): Boolean = vaultDao.hasVaults()
 
+    override suspend fun hasAnyCoinOnChain(chain: Chain): Boolean =
+        vaultDao.hasCoinOnChain(chain.id)
```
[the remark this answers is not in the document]
aminsato: False positive — `Chain.id` IS `Chain.raw`. From `Chain.kt:67-68`:
```kotlin
val id: ChainId
    get() = raw
```
So the SQL parameter sent by `hasCoinOnChain(chain.id)` is byte-for-byte identical to what `toCoinEntity()` writes via `chain.raw`. No mismatch.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/InitializeThorChainNetworkIdUseCase.kt
```diff
@@ -17,8 +18,8 @@ constructor(
 ) : InitializeThorChainNetworkIdUseCase {
 
     override suspend fun invoke() {
-        if (!vaultRepository.hasVaults()) {
-            Timber.d("Skipping THORChain network id init: no vaults")
+        if (!vaultRepository.hasAnyCoinOnChain(Chain.ThorChain)) {
+            Timber.d("Skipping THORChain network id init: no vault uses THORChain")
             return
```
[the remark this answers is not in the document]
aminsato: Out of scope for this PR — issue #4366 is scoped to "skip /status fetch when no vault uses THORChain", i.e. the unused-by-the-user case. There's no backend feature flag for THORChain in this codebase today; if/when one is added (for example via a remote-config kill switch), the gate here would naturally extend to AND with that flag. Tracking that separately would be a new ticket rather than expanding this PR's scope.

rkokhatskyi (approved):
LGTM

PR #4377, opened 2026-05-03
test(send): Phase A characterization tests for SendFormViewModel

aminsato:
## Summary
- Adds **42 unit tests** across 4 new files locking down the public surface of `SendFormViewModel.kt` (3415 LOC) before splitting it into focused sub-components.
- **No production code changes** — pure additions under `app/src/test/`. All tests pass green against current `main`.
- Lays the safety net for an upcoming refactor that will extract `AddressManager`, `AmountManager`, and a per-chain `SendDispatcher` from the ViewModel in follow-up PRs.

## Why
`SendFormViewModel` is the largest file in the project not currently touched by any open PR. At 3415 lines with 30+ public methods and ~24 injected dependencies, it cannot be safely refactored without a behavioral baseline. These tests pin the seams that the section composables (`SendFormAddressSection`, `SendFormAmountSection`, `SendFormAssetSection`) and `SendFormScreen` rely on, so any regression during the split fails the suite loudly.

## What's covered

| File | # | Surface |
|---|---|---|
| `SendFormViewModelInitTest.kt` | 14 | `loadData` arg parsing (defiType, amount, memo, slippage, autocompound), `loadVaultName` → `srcVaultName`, fiat currency from `AppCurrencyRepository`, expandedSection rules from `preSelectedTokenId` + `address` |
| `SendFormViewModelAddressTest.kt` | 11 | `setOutputAddress`, `setProviderAddress`, `isDstAddressComplete` flow, `scanAddress` / `scanProviderAddress` (QR roundtrip), `setAddressFromQrCode`, `openAddressBook` no-token gate |
| `SendFormViewModelAmountTest.kt` | 13 | `validateTokenAmount` (4 branches: empty / zero / non-numeric / valid), `toggleAmountInputType`, `expandSection`, `dismissError`, `back` → navigator, `chooseMaxTokenAmount`, `choosePercentageAmount` for F25/F50/F75/F100, default fraction list |
| `SendFormViewModelSubmitEvmTest.kt` | 4 | `send()` validation gates: blank-address focus + `expandSection(Address)`, blank-amount focus + `expandSection(Amount)`, no-selected-account error + no nav, `onClickContinue` dispatch to `send()` |

## Notable test technique
`textAsFlow()` uses Compose's `snapshotFlow`, which does not emit in unit tests on its own (no frame loop). Address tests that observe flow-driven state call `Snapshot.sendApplyNotifications()` after each `TextFieldState` mutation. This is documented in-line so future tests in this VM can copy the pattern.

## Out of scope (intentional)
- Full happy-path EVM submit (`Transaction` payload assembly → `transactionRepository.addTransaction` → `Route.VerifySend`) is **not** covered. Reaching it requires plumbing `accounts → preSelectToken → selectedToken → gasFee` flow — that's exactly the seam a follow-up PR will introduce, so the test is easier to write after the split rather than fighting current shape.
- Per-chain submit branches (`bond` / `unbond` / `stake` / `unstake` / `mint` / `redeem` / `withDrawUSDCCircle`), gas settings, and Asset/Network selection — these get their own test suites alongside the matching extraction PRs.

## Test plan
- [x] `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.send.SendFormViewModel*Test"` — all 47 tests pass (42 new + 5 existing `SendFormViewModelTronStakingTest`)
- [x] `./gradlew :app:ktfmtFormat` — formatted
- [ ] CI green

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive unit tests for send flows: address and provider field updates, QR-scan parsing/handling, address-book gating, amount validation and input type toggles, section expand/focus logic during submission, form initialization from navigation params (amount, memo, token/address), slippage seeding for a specific redeem type, amount-selection percentages/loading states, navigation/back behavior, and submission validation cases.

### DIFF
--- a/app/src/test/java/com/vultisig/wallet/ui/models/send/SendFormViewModelAddressTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/send/SendFormViewModelAddressTest.kt
@@ -196,7 +196,9 @@ internal class SendFormViewModelAddressTest {
     }
 
     @Test
-    fun `openAddressBook OUTPUT routes the address into addressFieldState`() = runTest {
+    fun `openAddressBook is a no-op when no token is selected`() = runTest {
+        // The address book flow is gated on a token selection: without selectedTokenValue,
+        // openAddressBook returns early before consulting the address book entry use case.
         val entry: AddressBookEntry = mockk(relaxed = true)
         every { entry.address } returns "0xfromBook"
         every { entry.chain } returns mockk(relaxed = true)
@@ -204,9 +206,6 @@ internal class SendFormViewModelAddressTest {
 
         val vm = buildViewModel()
         advanceUntilIdle()
-        // Without a selectedToken, openAddressBook returns early because selectedTokenValue is
-        // null.
-        // We assert the no-op behavior here — the address book flow is gated on a token selection.
 
         vm.openAddressBook(AddressBookType.OUTPUT)
         advanceUntilIdle()
--- a/app/src/test/java/com/vultisig/wallet/ui/models/send/SendFormViewModelAddressTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/send/SendFormViewModelAddressTest.kt
@@ -1,0 +1,262 @@
+@file:OptIn(ExperimentalCoroutinesApi::class, ExperimentalStdlibApi::class)
+
+package com.vultisig.wallet.ui.models.send
+
+import androidx.compose.runtime.snapshots.Snapshot
+import androidx.lifecycle.SavedStateHandle
+import androidx.navigation.toRoute
+import com.vultisig.wallet.data.models.AddressBookEntry
+import com.vultisig.wallet.data.models.settings.AppCurrency
+import com.vultisig.wallet.data.repositories.AddressParserRepository
+import com.vultisig.wallet.data.repositories.AppCurrencyRepository
+import com.vultisig.wallet.data.repositories.ChainAccountAddressRepository
+import com.vultisig.wallet.data.repositories.RequestResultRepository
+import com.vultisig.wallet.data.usecases.RequestAddressBookEntryUseCase
+import com.vultisig.wallet.data.usecases.RequestQrScanUseCase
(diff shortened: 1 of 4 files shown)


aminsato:
Addressing CodeRabbit nits:

**Fixed (db8cd31a):** Renamed `openAddressBook OUTPUT routes the address into addressFieldState` → `openAddressBook is a no-op when no token is selected`. The previous name described the asserted-against intent rather than the actual no-op assertion.

**Deferred — Kotest assertions:** CLAUDE.md does state "Use Kotest for assertions", but in practice:
- 0 of 28 existing test files in `app/src/test/` import `io.kotest`
- All existing tests (including `SendFormViewModelTronStakingTest.kt` that this PR's `buildViewModel()` pattern was modeled on) use `kotlin.test`
- Kotest is not declared as a dependency in `app/build.gradle.kts`, `gradle/libs.versions.toml`, or any module — the matchers aren't even on the test classpath

Adding the Kotest dependency and migrating one PR's worth of tests would create the lone Kotest island in the repo. That's a separate codebase-wide convention decision (either migrate everything or update CLAUDE.md to match reality), out of scope for a test-only safety-net PR. Happy to file a follow-up if there's appetite to standardize.

aminsato:
@coderabbitai full review

aminsato:
@coderabbitai Skipping this one.

`UnconfinedTestDispatcher` is the established pattern across all 5 new `SendFormViewModel*Test.kt` files in this PR (and other ViewModel tests in the repo). It runs coroutines eagerly up to the first suspension, which is what these tests rely on for ViewModel `init {}` side effects to materialize before assertions. The trailing `advanceUntilIdle()` calls flush any remaining suspended work — they're a defensive belt-and-braces, not the primary scheduling mechanism.

Switching to `StandardTestDispatcher` would change semantics (init-block work no longer runs eagerly) and likely require additional `advanceUntilIdle()` calls *before* each action under test, not just after. Doing it for one file out of five would also break consistency. This is out of scope for a Phase A characterization PR — happy to revisit as a separate test-infra change if the team wants to standardize on `StandardTestDispatcher` repo-wide.

rkokhatskyi (approved):

PR #4405, opened 2026-05-04
fix(secure-prefs): narrow destructive recovery to KeyPermanentlyInvalidatedException only (#4401)

### LINKED ISSUE #4401: SecureSharedPreferences: recoverAndRetry destroys user data on transient AndroidKeyStore errors
## Summary

`MainDataModule.provideEncryptedSharedPrefs.recoverAndRetry` (`data/src/main/kotlin/com/vultisig/wallet/data/MainDataModule.kt:97-130`) is invoked from a catch block (`:134-138`) that handles `GeneralSecurityException` and `IOException` indiscriminately, then unconditionally:

- deletes `shared_prefs/token_secure_prefs.xml` (every encrypted key/value the app stored)
- deletes the AndroidKeyStore alias `vultisig_secure_prefs_key`
- on second failure, falls back to `InMemorySharedPreferences()` (non-persistent — all stored data lost for the session)

`GeneralSecurityException` legitimately covers `KeyPermanentlyInvalidatedException` (credential change), where destruction is correct. But it **also** covers transient AndroidKeyStore daemon errors on certain OEM ROMs, where a retry would have succeeded. A single transient stall on cold launch can therefore wipe a user's encrypted state with no recovery path.

## Offending code

`data/src/main/kotlin/com/vultisig/wallet/data/MainDataModule.kt:132-138`

```kotlin
return try {
    create()
} catch (e: GeneralSecurityException) {
    recoverAndRetry(e)
} catch (e: IOException) {
    recoverAndRetry(e)
}
```

## Suggested fix

- Narrow destructive recovery to `KeyPermanentlyInvalidatedException` (and explicit "alias not found" / corrupted-entry cases).
- Other `GeneralSecurityException` / `IOException` should retry-with-backoff (e.g. 3 attempts with 200/500/1000 ms backoff) before considering destruction.
- Consider gating destruction behind N consecutive failures across launches before the alias is removed.

## Context

Surfaced during root-cause analysis for #4360 (cold-launch splash hang). Filing separately because the destructive recovery is a data-loss risk on its own, regardless of whether #4360 ultimately shares the same root cause.

rkokhatskyi:
## Summary
- Catch only `KeyPermanentlyInvalidatedException` in `MainDataModule.provideEncryptedSharedPrefs`. Transient `GeneralSecurityException`/`IOException` from cold-launch keystore stalls now propagate to Hilt and Crashlytics instead of silently destroying `shared_prefs/token_secure_prefs.xml` and the `vultisig_secure_prefs_key` alias.
- Drop the second-tier `InMemorySharedPreferences` fallback so a still-broken keystore surfaces as a real failure rather than a non-persistent store that loses every write at process death.
- Specialize `recoverAndRetry` to `KeyPermanentlyInvalidatedException` (future widening is now a compile error). Demote recovery log from `Timber.e` to `Timber.w` and name the trigger. Pin file-then-alias delete order so a partial recovery leaves no readable ciphertext for the next launch.
- Extract `SecurePrefsRecoveryPolicy.shouldDestructivelyRecover(Throwable)` covered by 8 JUnit 5 cases (KPIE → true; GSE, IOException, InvalidKeyException, UnrecoverableKeyException, IllegalStateException, RuntimeException, KPIE-as-`cause` → false). Drop now-unused imports.

Closes #4401

## Deferred
- OEM-wrapped KPIE (Samsung/Xiaomi/Huawei nest KPIE inside `IllegalStateException`/`KeyStoreException`): policy intentionally does not unwrap `cause` — revisit when Crashlytics signal justifies it.
- `InMemorySharedPreferences` in `SecureSharedPreferences.kt` is orphaned; separate cleanup PR.

## Test plan
- [x] `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.utils.SecurePrefsRecoveryPolicyTest"` (8/8 pass)
- [x] `./gradlew :data:compileDebugKotlin` clean
- [x] `./gradlew lint` passes (abortOnError = true)
- [ ] **Manual**: change device credential to invalidate the master key, cold-launch, confirm KPIE path recovers cleanly and reseeds prefs; then cold-launch after a transient keystore stall (throttle `keystored`, force-stop, relaunch) and confirm encrypted prefs survive instead of being destroyed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secure storage error recovery to distinguish between permanent and transient failures. The app now only performs destructive cleanup when absolutely necessary, reducing data loss from temporary security or I/O errors.

* **Tests**
  * Added comprehensive test coverage for the secure storage recovery policy.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/MainDataModule.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/MainDataModule.kt
@@ -2,6 +2,7 @@ package com.vultisig.wallet.data
 
 import android.content.Context
 import android.content.SharedPreferences
+import android.security.keystore.KeyPermanentlyInvalidatedException
 import androidx.datastore.preferences.core.PreferenceDataStoreFactory
 import androidx.datastore.preferences.preferencesDataStoreFile
 import androidx.security.crypto.EncryptedSharedPreferences
@@ -9,7 +10,6 @@ import androidx.security.crypto.MasterKey
 import com.vultisig.wallet.data.sources.AppDataStore
 import com.vultisig.wallet.data.sources.AppDataStoreImpl
 import com.vultisig.wallet.data.utils.EncryptingSharedPreferences
-import com.vultisig.wallet.data.utils.InMemorySharedPreferences
 import com.vultisig.wallet.data.utils.SECURE_PREFS_KEY_ALIAS
 import com.vultisig.wallet.data.utils.SharedPrefsMasterKeyInitializer
 import com.vultisig.wallet.data.utils.buildSecurePrefsKey
@@ -20,8 +20,6 @@ import dagger.hilt.InstallIn
 import dagger.hilt.android.qualifiers.ApplicationContext
 import dagger.hilt.components.SingletonComponent
 import java.io.File
-import java.io.IOException
-import java.security.GeneralSecurityException
 import java.security.KeyStore
 import javax.inject.Qualifier
 import javax.inject.Singleton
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4406, opened 2026-05-04
fix(secure-prefs): complete prewarm deferred on every non-cancellation throwable

### LINKED ISSUE #4402: SharedPrefsMasterKeyInitializer: prewarm silently leaks _prewarmResult on uncaught exceptions
## Summary

`SharedPrefsMasterKeyInitializer.prewarm()` only catches `GeneralSecurityException` and `IOException` around `buildSecurePrefsKey()`. But `buildSecurePrefsKey()` can also throw `IllegalStateException` — from the 3-second lock timeout at `SecureSharedPreferences.kt:39` (`error("Timed out waiting for secure prefs key lock after 3 s")`) and from the unexpected-entry-type branch at `:46`. Neither is caught here.

When that happens, the launched coroutine throws into the `SupervisorJob`, which logs but doesn't propagate, and **`_prewarmResult` is never completed**.

Every subsequent caller that does `if (prewarmResult.isCompleted)` then sees `false` and falls through to running `buildSecurePrefsKey()` itself — typically on the main thread on cold launch (`MainDataModule.kt:80-83`). The whole point of the optimization (keep keystore work off the main thread) silently disappears, with zero log signal that the prewarm path failed.

## Offending code

`data/src/main/kotlin/com/vultisig/wallet/data/utils/SharedPrefsMasterKeyInitializer.kt:34-47`

```kotlin
fun prewarm() {
    CoroutineScope(SupervisorJob() + Dispatchers.IO + CoroutineName("SecurePrefsPrewarm"))
        .launch {
            try {
                _prewarmResult.complete(buildSecurePrefsKey())
            } catch (e: GeneralSecurityException) {
                Timber.w(e, "Secure prefs key prewarm failed; deferring to sync path")
                _prewarmResult.complete(null)
            } catch (e: IOException) {
                Timber.w(e, "Secure prefs key prewarm failed; deferring to sync path")
                _prewarmResult.complete(null)
            }
        }
}
```

## Suggested fix

Always complete `_prewarmResult`, e.g.:

```kotlin
.launch {
    val key: SecretKey? = try {
        buildSecurePrefsKey()
        throw e
    } catch (e: Throwable) {
        Timber.w(e, "Secure prefs key prewarm failed; deferring to sync path")
        null
    }
    _prewarmResult.complete(key)
}
```

(`CancellationException` must always be re-thrown to honour structured concurrency.)

## Context

Surfaced during root-cause analysis for #4360. The prewarm exists specifically to keep AndroidKeyStore work off the main thread on cold launch; this bug means the optimization can quietly stop working without anyone noticing — and the symptoms then look exactly like #4360.

rkokhatskyi:
The narrow `GeneralSecurityException | IOException` catch in `SharedPrefsMasterKeyInitializer.prewarm()` left `_prewarmResult` uncompleted on `IllegalStateException` (the lock-timeout `error(...)` at `SecureSharedPreferences.kt:39`), `ProviderException`, OEM `RuntimeException`, `InterruptedException`, and `NullPointerException`. Any awaiter would hang forever instead of falling back to the synchronous path.

## Summary

- Extract the prewarm body into `internal fun runPrewarmInto(target, supplier)` that rethrows `CancellationException` first so structured concurrency wins, then catches every other `Exception` with a single Timber WARN and settles `target` with `null`. Catching `Exception` (not `Throwable`) matches the existing log-and-swallow precedent in `MainDataModule` / `SecureSharedPreferences` and lets JVM `Error`s propagate.
- First-write semantics are preserved via `CompletableDeferred.complete()` returning `false` on the second call, making the helper idempotent against an already-settled target.
- Add `SharedPrefsMasterKeyInitializerTest` covering happy path, `GeneralSecurityException`, `IOException`, the `IllegalStateException` lock-timeout regression, `IllegalArgumentException`, raw `RuntimeException`, `InterruptedException`, `CancellationException` rethrow without touching the target, first-write idempotency, and parameterized log-once-at-WARN invariants.

## Why this is safe

The sole consumer at `MainDataModule.kt:80-82` reads `_prewarmResult.isCompleted` non-blockingly and falls back via `getCompleted() ?: buildSecurePrefsKey()`. The new "completed with `null`" outcome lands on the exact same synchronous keystore lookup the previous "uncompleted" outcome did, so the success path and the destructive-recovery path from #4401 are unchanged — this PR only widens which throwables reach that fallback.

## Test plan

- [x] `./gradlew :data:testDebugUnitTest`
- [x] `ktfmt` clean (auto-formatted by pre-commit hook)

Closes #4402

## Summary by CodeRabbit

* **Refactor**
  * Restructured key initialization logic with improved exception handling and diagnostic logging for failure scenarios.

* **Tests**
  * Added comprehensive test coverage for key initialization, including success paths, failure handling, cancellation semantics, and write-once verification.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/utils/SharedPrefsMasterKeyInitializer.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/utils/SharedPrefsMasterKeyInitializer.kt
@@ -36,17 +36,10 @@ object SharedPrefsMasterKeyInitializer {
     }
 
     /**
-     * Runs the prewarm body against [target]. Settles [target] with the [supplier] result on
-     * success, or with `null` after a WARN-level Timber log on any non-cancellation [Exception] —
-     * the previous catch-list of `GeneralSecurityException | IOException` left the deferred
-     * uncompleted on `IllegalStateException` (the lock-timeout `error(...)` in
-     * [buildSecurePrefsKey]), `ProviderException`, OEM `RuntimeException`, etc., so any future
-     * awaiter would hang forever (issue #4402). Re-throws [CancellationException] without touching
-     * [target] so structured concurrency keeps propagating cancellation; the consumer in
-     * [com.vultisig.wallet.data.MainDataModule.provideEncryptedSharedPrefs] reads
-     * [Deferred.isCompleted] non-blockingly and falls through to a synchronous keystore lookup when
-     * the deferred is still active. A second call against an already-settled target is a no-op
-     * because [CompletableDeferred.complete] returns `false` on the second invocation.
+     * Settles [target] with [supplier]'s result, or with `null` after a WARN log on any
+     * non-cancellation [Exception]. Rethrows [CancellationException] without touching [target] so
+     * structured concurrency keeps propagating cancellation. Idempotent: a second call against an
+     * already-settled target is a no-op.
      */
     internal fun runPrewarmInto(
         target: CompletableDeferred<SecretKey?>,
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4408, opened 2026-05-05
fix(secure-prefs): throw KeyStoreException from buildSecurePrefsKey

### LINKED ISSUE #4403: SecureSharedPreferences: buildSecurePrefsKey lock-timeout throws uncaught IllegalStateException, can crash app on cold launch
## Summary

`buildSecurePrefsKey` (`data/src/main/kotlin/com/vultisig/wallet/data/utils/SecureSharedPreferences.kt:37-65`) calls `error(...)` on `secureKeyLock.tryLock` timeout, which throws `IllegalStateException`.

The provider catch in `MainDataModule.provideEncryptedSharedPrefs` (`MainDataModule.kt:132-138`) handles only `GeneralSecurityException` and `IOException`, so `IllegalStateException` propagates up as a Hilt binding failure. On cold launch this surfaces as a **process crash** during `MainViewModel` injection — strictly worse than the hang the lock timeout was meant to prevent.

The unexpected-entry-type branch at `:46-48` has the same shape and same problem.

(Same gap exists in `SharedPrefsMasterKeyInitializer.prewarm()` — filed separately.)

## Offending code

`data/src/main/kotlin/com/vultisig/wallet/data/utils/SecureSharedPreferences.kt:37-48`

```kotlin
internal fun buildSecurePrefsKey(): SecretKey {
    if (!secureKeyLock.tryLock(3_000L, TimeUnit.MILLISECONDS)) {
        error("Timed out waiting for secure prefs key lock after 3 s")
    }
    try {
        val ks = KeyStore.getInstance(KEYSTORE).apply { load(null) }
        val entry = ks.getEntry(SECURE_PREFS_KEY_ALIAS, null)
        if (entry != null) {
            return (entry as? KeyStore.SecretKeyEntry)?.secretKey
                ?: error(
                    "Alias $SECURE_PREFS_KEY_ALIAS holds unexpected entry type: ${entry::class.simpleName}"
                )
        }
        ...
```

## Trigger scenario

1. Prewarm thread acquires `secureKeyLock` and blocks on a stalled AndroidKeyStore daemon (AndroidKeyStore is known to stall multi-second on Pixel/Samsung post-doze)
2. Main thread reaches `setContent`, Compose composition forces Hilt to build `MainViewModel`, which transitively requires `SharedPreferences`
3. Provider runs `buildSecurePrefsKey()` on main, contends for the lock, `tryLock(3 s)` returns `false` → `error(...)` throws `IllegalStateException`
4. Hilt binding resolution fails → process crashes

## Suggested fix

Either:
- Throw a `GeneralSecurityException` (or another exception type the provider catches) so the existing recovery path can run, or
- Return `null` (signature change) and let callers route through recovery, or
- Proceed without the lock as a last-resort fallback (best-effort key access; documented as such)

Same change should be applied to the `:46` "unexpected entry type" branch.

## Context

Surfaced during root-cause analysis for #4360. The conditions to trigger this are narrow (lock contention with a stalled daemon) but it's a clear correctness bug: the 3-second timeout was meant to fail fast gracefully, and instead it crashes the app on cold launch.

rkokhatskyi:
## Summary
- `buildSecurePrefsKey` called `error(...)` on `tryLock` timeout and on unexpected keystore entry type, throwing `IllegalStateException`. The provider catch in `MainDataModule.provideEncryptedSharedPrefs` and the prewarm catch in `SharedPrefsMasterKeyInitializer.prewarm` handle only `GeneralSecurityException`/`IOException`, so on cold launch a stalled AndroidKeyStore daemon turned the 3 s lock timeout into a Hilt binding failure and process crash — strictly worse than the hang it was meant to prevent.
- Replaces both `error(...)` sites with `throw KeyStoreException(...)` (a `GeneralSecurityException` subtype) and now catches `InterruptedException` from `tryLock`, restoring the interrupt flag before rethrowing wrapped. The existing recovery in `MainDataModule` and `SharedPrefsMasterKeyInitializer.prewarm()` handles all three paths unchanged. Each failure logs once at WARN via Timber.
- Extracts `SECURE_PREFS_LOCK_TIMEOUT_MS = 3_000L` and adds an internal `buildSecurePrefsKey(lock, lockTimeoutMillis, loadEntry, generateKey)` seam so every exit path is unit-testable on the JVM. The no-arg overload pins these to production values; the keystore lookup and `KeyGenParameterSpec` are bit-for-bit identical to `origin/main`.

## Test plan
- [x] `./gradlew :data:compileDebugKotlin`
- [x] `./gradlew :data:testDebugUnitTest --tests '*BuildSecurePrefsKeyTest*'`
- [x] `./gradlew ktfmtFormat`
- [ ] Cold-launch smoke on a device where the AndroidKeyStore daemon stalls long enough to exhaust the 3 s lock timeout — expect recovery, not a crash.

## Notes
- Out of scope: the `GeneralSecurityException` catch in `MainDataModule.provideEncryptedSharedPrefs` is broad; a lock-timeout will hit `recoverAndRetry` and delete the prefs file plus keystore alias. PR #4401 narrows that catch to `KeyPermanentlyInvalidatedException` and is in flight separately.
- Out of scope: removing the lock entirely.

Closes #4403

## Summary by CodeRabbit

* **Refactor**
  * Improved error handling in secure key management with configurable timeout controls and more specific exception types instead of generic failures.

* **Tests**
  * Added comprehensive test suite covering key building operations, including timeout handling, interruption scenarios, error paths, and lock release invariants.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/utils/SecureSharedPreferences.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/utils/SecureSharedPreferences.kt
@@ -5,6 +5,7 @@ import android.security.keystore.KeyGenParameterSpec
 import android.security.keystore.KeyProperties
 import android.util.Base64
 import java.security.KeyStore
+import java.security.KeyStoreException
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.ReentrantLock
@@ -26,6 +27,8 @@ private const val P_INT = "i:"
 private const val P_LONG = "l:"
 private const val P_FLOAT = "f:"
 
+private const val SECURE_PREFS_LOCK_TIMEOUT_MS = 3_000L
+
 private val secureKeyLock = ReentrantLock()
 
 /**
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4393, opened 2026-05-04
refactor(send): split SendFormViewModel into Address/Amount managers

aminsato:
## Summary
- Extract destination-address handling (debounce + ENS resolution + `isDstAddressComplete` gate) into a new **`AddressManager`** (120 LOC).
- Extract bidirectional token↔fiat conversion + reaping check into a new **`AmountManager`** (168 LOC).
- `SendFormViewModel` drops ~150 LOC of inline flow logic. **No public-surface changes** — all 47 Phase A tests (#4377) pass unchanged.

## Why
With the Phase A characterization safety net in place (47 tests across 5 files), `SendFormViewModel` (3415 LOC) is now safe to split. This PR carves out the two clearest sub-responsibilities without touching the parts that still need design work (`defiType` percentage/submit branches).

Each new manager has a single responsibility and a narrow `StateFlow` API. The VM observes them and mirrors their state into `uiState`, preserving every consumer contract that `SendFormScreen` and the section composables depend on.

### `AddressManager` (6 deps, 120 LOC)
| State / behaviour | Before | After |
|---|---|---|
| `addressFieldState` debounce + ENS resolve (`collectAddress`) | `SendFormViewModel:377` | `AddressManager.collectResolvedAddress` |
| `isDstAddressComplete` derivation (`collectDstAddress`) | `SendFormViewModel:609` | `AddressManager.collectIsComplete` |
| `resolvedDstAddress`, `dstAddressLabel` private fields | VM state | `AddressManager` private state, exposed as read-only `StateFlow` |
| `setOutputAddress(String)` | direct `addressFieldState.setText...` | delegates to `AddressManager` |
| `expandSection(Amount)` on validation | inlined inside `mapLatest` | manager emits `onAddressValidated`; VM observes |

The 80-line nested `when/try` from `collectAddress` is split into `handleAddressInput` + `tryResolveName` for readability.

### `AmountManager` (9 deps, 168 LOC)
| State / behaviour | Before | After |
|---|---|---|
| Bidirectional token↔fiat conversion (`collectAmountChanges`) | `SendFormViewModel:2930` | `AmountManager.collectConversion` (split into `handleTokenInput` + `handleFiatInput`) |
| Reaping check (`collectAmountChecks`) | `SendFormViewModel:2978` | `AmountManager.collectReaping` |
| `convertValue` price lookup | private VM helper | private manager helper, narrowed signature (`token: Coin` non-null) |
[...]

PR #4412, opened 2026-05-05
refactor(send): split submit functions into per-defiType strategies

aminsato:
## Summary

Phase 3 of the SendFormViewModel refactor (PR 1 #4377 + PR 2 #4393 already merged).

- Each `defiType` branch now lives behind a dedicated `SendSubmitStrategy` implementation under `ui/models/send/submit/`. The 8 submit functions, the 5 `create*DepositTransaction` helpers, the inline `accountValidation`/`awaitGasFee`/`getFeesFiatValue`/`getBitcoinTransactionPlan` helpers, and the `AccountValidation` data class all leave the VM. **SendFormViewModel.kt drops from 3261 → 1873 LOC.**
- Strategies are plain classes manually constructed in the VM `init` (matching the `AddressManager`/`AmountManager` precedent from PR 2). No Hilt multibinding — `defiType` is a closed enum and one-class-per-concern keeps the diff lift-and-shift.
- New shared collaborators: `AccountValidator` (replaces `accountValidation()`+`awaitGasFee()`) and `BitcoinPlanService` (still called by the VM's `collectPlanFee` flow).
- Public surface preserved: `send()` remains as a one-line passthrough so existing screen bindings and characterization tests keep working. `bond/unbond/unstake` (no external callers post-extraction) and the four previously-private submit functions are deleted.
- 29 new Phase C tests under `submit/`: AccountValidator validate() error branches plus one error-path test per strategy (no_address, no_amount, RUNE-balance-insufficient, operator-fee out of range, slippage, WITHDRAW_RUJI silent early-return, DefaultSend pre-launch expandSection+focus). Total send.* tests: **135**, all green.
- Same diff also clears the IDE warnings the trimmed VM exposed: unused `BondAddress` enum value, unused `calculateGasLimit`/`enableAdvanceGasUi` helpers, the cascade `if`-`else if` in `loadAccounts` (now a `when` over `defiType`), one redundant `R.string` qualifier, and the missing `@param:` use-site target on `chainLogo`.

**Deferred to a future PR**: the percentage/calculate functions (`chooseMaxTokenAmount`, `choosePercentageAmount`, `calculatePercentageWithAccurateFee`, `fetchPercentageOfAvailableBalance`). They share the `defiType` switch with submit but also write to `gasFee`/`specific` and orchestrate fee recalculation — a different concern that would double the size of this PR.

## Transaction link
 https://etherscan.io/tx/0x81b2cc9f32aaa9eb42b909ded2e6005a94b806a88ff8b3d352571bfb17520978
 
## Test plan

- [ ] `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.send.*"` — expect 135 tests, all green
- [ ] `./gradlew :app:compileDebugKotlin` clean
- [ ] `./gradlew :app:ktfmtFormat` clean
- [ ] Manually exercise each defiType flow on emulator: regular send (EVM + UTXO + Tron), bond, unbond, stake (RUJI/TCY/STCY), unstake (RUJI/TCY/STCY/WITHDRAW_RUJI), mint (yRUNE/yTCY), redeem (yRUNE/yTCY), withdraw USDC Circle, Tron freeze/unfreeze. Verify that successful paths hit the right `Verify*` route and that error paths surface the expected error text.

## Summary by CodeRabbit

* **New Features**
  * Expanded DeFi actions in Send: staking, bonding, unbonding, minting, redeeming, and USDC Circle withdrawals with end-to-end submission flows.

* **Improvements**
  * Stronger input validation and clearer error messages for send/DeFi flows.
  * Improved gas/fee estimation and balance checks to reduce failed transactions.
  * More consistent transaction construction and navigation to verification for smoother UX.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
@@ -1252,11 +1252,17 @@ constructor(
             DeFiNavActions.REDEEM_YTCY -> redeemStrategy.submit()
 
             DeFiNavActions.WITHDRAW_USDC_CIRCLE -> withdrawUsdcCircleStrategy.submit()
-            else -> sendStrategy.submit()
+
+            null,
+            DeFiNavActions.DEPOSIT_USDC_CIRCLE,
+            DeFiNavActions.STAKE_CACAO,
+            DeFiNavActions.UNSTAKE_CACAO,
+            DeFiNavActions.ADD_LP,
+            DeFiNavActions.REMOVE_LP,
+            DeFiNavActions.FREEZE_TRX,
+            DeFiNavActions.UNFREEZE_TRX -> sendStrategy.submit()
         }
     }
-
-    fun send() = sendStrategy.submit()
 
     private fun hideLoading() {
         uiState.update { it.copy(isLoading = false) }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
@@ -1258,12 +1258,6 @@ constructor(
 
     fun send() = sendStrategy.submit()
 
-    fun bond() = bondStrategy.submit()
-
-    fun unbond() = unbondStrategy.submit()
-
-    fun unstake() = unstakeStrategy.submit()
-
     private fun hideLoading() {
         uiState.update { it.copy(isLoading = false) }
     }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
@@ -12,7 +12,6 @@ import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import androidx.navigation.toRoute
-import com.vultisig.wallet.R
 import com.vultisig.wallet.R.string
 import com.vultisig.wallet.data.blockchain.FeeServiceComposite
 import com.vultisig.wallet.data.blockchain.model.StakingDetails.Companion.generateId
@@ -20,14 +19,10 @@ import com.vultisig.wallet.data.blockchain.model.Transfer
 import com.vultisig.wallet.data.blockchain.model.VaultData
 import com.vultisig.wallet.data.blockchain.thorchain.RujiStakingService.Companion.RUJI_REWARDS_COIN
 import com.vultisig.wallet.data.blockchain.tron.GetTronFrozenBalancesUseCase
-import com.vultisig.wallet.data.blockchain.tron.TRON_STAKING_MEMO_REGEX
 import com.vultisig.wallet.data.blockchain.tron.TronFrozenBalanceState
 import com.vultisig.wallet.data.blockchain.tron.TronResourceType
 import com.vultisig.wallet.data.blockchain.tron.TronStakingOperation
 import com.vultisig.wallet.data.blockchain.tron.tronStakingMemo
-import com.vultisig.wallet.data.chains.helpers.EthereumFunction
-import com.vultisig.wallet.data.chains.helpers.ThorchainFunctions
-import com.vultisig.wallet.data.chains.helpers.UtxoHelper
 import com.vultisig.wallet.data.models.Account
 import com.vultisig.wallet.data.models.Address
 import com.vultisig.wallet.data.models.AddressBookEntry
@@ -35,25 +30,15 @@ import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.ChainId
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.Coins
-import com.vultisig.wallet.data.models.DepositMemo
-import com.vultisig.wallet.data.models.DepositMemo.Bond
-import com.vultisig.wallet.data.models.DepositTransaction
-import com.vultisig.wallet.data.models.EstimatedGasFee
-import com.vultisig.wallet.data.models.FiatValue
 import com.vultisig.wallet.data.models.GasFeeParams
 import com.vultisig.wallet.data.models.ImageModel
-import com.vultisig.wallet.data.models.OPERATION_CIRCLE_WITHDRAW
 import com.vultisig.wallet.data.models.TokenId
 import com.vultisig.wallet.data.models.TokenStandard
 import com.vultisig.wallet.data.models.TokenValue
-import com.vultisig.wallet.data.models.Transaction
 import com.vultisig.wallet.data.models.Vault
 import com.vultisig.wallet.data.models.VaultId
-import com.vultisig.wallet.data.models.allowZeroGas
-import com.vultisig.wallet.data.models.coinType
 import com.vultisig.wallet.data.models.getPubKeyByChain
 import com.vultisig.wallet.data.models.payload.BlockChainSpecific
-import com.vultisig.wallet.data.models.payload.KeysignPayload
 import com.vultisig.wallet.data.repositories.AccountsRepository
 import com.vultisig.wallet.data.repositories.AddressParserRepository
 import com.vultisig.wallet.data.repositories.AdvanceGasUiRepository
@@ -79,16 +64,21 @@ import com.vultisig.wallet.ui.models.send.AmountFraction.F100
 import com.vultisig.wallet.ui.models.send.AmountFraction.F25
 import com.vultisig.wallet.ui.models.send.AmountFraction.F50
 import com.vultisig.wallet.ui.models.send.AmountFraction.F75
+import com.vultisig.wallet.ui.models.send.submit.AccountValidator
+import com.vultisig.wallet.ui.models.send.submit.BitcoinPlanService
+import com.vultisig.wallet.ui.models.send.submit.BondStrategy
+import com.vultisig.wallet.ui.models.send.submit.DefaultSendStrategy
+import com.vultisig.wallet.ui.models.send.submit.MintStrategy
+import com.vultisig.wallet.ui.models.send.submit.RedeemStrategy
+import com.vultisig.wallet.ui.models.send.submit.StakeStrategy
+import com.vultisig.wallet.ui.models.send.submit.UnbondStrategy
+import com.vultisig.wallet.ui.models.send.submit.UnstakeStrategy
+import com.vultisig.wallet.ui.models.send.submit.WithdrawUsdcCircleStrategy
 import com.vultisig.wallet.ui.navigation.Destination
 import com.vultisig.wallet.ui.navigation.Navigator
 import com.vultisig.wallet.ui.navigation.Route
 import com.vultisig.wallet.ui.navigation.back
 import com.vultisig.wallet.ui.screens.select.AssetSelected
-import com.vultisig.wallet.ui.screens.v2.defi.STAKING_RUJI_CONTRACT
-import com.vultisig.wallet.ui.screens.v2.defi.STAKING_TCY_COMPOUND_CONTRACT
-import com.vultisig.wallet.ui.screens.v2.defi.YRUNE_CONTRACT
-import com.vultisig.wallet.ui.screens.v2.defi.YRUNE_YTCY_AFFILIATE_CONTRACT
-import com.vultisig.wallet.ui.screens.v2.defi.YTCY_CONTRACT
 import com.vultisig.wallet.ui.screens.v2.defi.model.DeFiNavActions
 import com.vultisig.wallet.ui.screens.v2.defi.model.parseDepositType
 import com.vultisig.wallet.ui.utils.UiText
@@ -99,13 +89,11 @@ import dagger.hilt.android.lifecycle.HiltViewModel
 import java.math.BigDecimal
 import java.math.BigInteger
 import java.math.RoundingMode
-import java.util.UUID
 import javax.inject.Inject
 import kotlin.uuid.Uuid
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.TimeoutCancellationException
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
@@ -127,9 +115,7 @@ import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
-import kotlinx.coroutines.withTimeout
 import timber.log.Timber
-import vultisig.keysign.v1.TransactionType
 import wallet.core.jni.proto.Bitcoin
 
 @Immutable
@@ -142,7 +128,7 @@ internal data class TokenBalanceUiModel(
     val isLayer2: Boolean,
     val tokenStandard: String?,
     val tokenLogo: ImageModel,
-    @DrawableRes val chainLogo: Int,
+    @param:DrawableRes val chainLogo: Int,
 )
 
 sealed class AmountFraction(val title: UiText, val value: Float) {
@@ -209,7 +195,6 @@ internal enum class SendSections {
     Asset,
     Address,
     Amount,
-    BondAddress,
 }
 
 internal enum class SendFocusField {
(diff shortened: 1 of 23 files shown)


aminsato:
@coderabbittai full review

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -1072,1230 +1234,29 @@ constructor(
...
-                )
-            } catch (e: InvalidTransactionDataException) {
-                showError(e.text)
-            } catch (e: Exception) {
-                showError(e.message?.asUiText() ?: UiText.Empty)
-            } finally {
-                hideLoading()
-            }
+            DeFiNavActions.WITHDRAW_USDC_CIRCLE -> withdrawUsdcCircleStrategy.submit()
+            else -> sendStrategy.submit()
```
rkokhatskyi: The `else -> sendStrategy.submit()` arm silently routes `DEPOSIT_USDC_CIRCLE`, `STAKE_CACAO`, `UNSTAKE_CACAO`, `ADD_LP`, and `REMOVE_LP` through `DefaultSendStrategy` which builds a `Transaction` rather than the `DepositTransaction` these flows expect, so the `when` should enumerate every `DeFiNavActions` value explicitly which makes a future variant fail to compile instead of silently misrouting

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -1072,1230 +1234,29 @@ constructor(
...
-                libType = vault.libType,
-                wasmExecuteContractPayload = null,
-            )
-
-        val utxo = UtxoHelper.getHelper(vault, keysignPayload.coin.coinType)
-
-        val plan = utxo.getBitcoinTransactionPlan(keysignPayload)
-        return plan
-    }
+    fun send() = sendStrategy.submit()
```
rkokhatskyi: `fun send()` is a public passthrough that bypasses the dispatch table at line 1235 entirely, so a caller invoking it while `defiType` is non null silently runs `sendStrategy` regardless of the active DeFi action, and the only consumers are three tests in `SendFormViewModelSubmitEvmTest` that intentionally skip the dispatcher, either inline `sendStrategy.submit()` into those tests or mark `send` as `@TestOnly internal`

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -0,0 +1,402 @@
...
+                            TRON_STAKING_MEMO_REGEX.matches(memo)
+                    if (!isTronStakingOp && srcAddress == dstAddress) {
+                        throw InvalidTransactionDataException(
+                            UiText.StringResource(R.string.send_error_same_address)
+                        )
+                    }
+                }
+
+                val specific =
+                    blockChainSpecificRepository
```
rkokhatskyi: Every other strategy in this PR wraps `blockChainSpecificRepository.getSpecific(...)` in `withContext(Dispatchers.IO)` whereas `DefaultSendStrategy` keeps it on the caller dispatcher which is `viewModelScope` defaulting to `Main`, so the network round trip in `getSpecific` runs on the UI dispatcher for the most common send path while every deposit path is correctly off thread

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -0,0 +1,402 @@
...
+
+                val tokenAmount = tokenAmountFieldState.text.toString().toBigDecimalOrNull()
+
+                if (tokenAmount == null || tokenAmount <= BigDecimal.ZERO) {
+                    throw InvalidTransactionDataException(
+                        UiText.StringResource(R.string.send_error_no_amount)
+                    )
+                }
+
+                val gasFee = accountValidator.awaitGasFee()
```
rkokhatskyi: `DefaultSendStrategy` reads `selectedAccountProvider()`, parses `tokenAmount`, and calls `accountValidator.awaitGasFee()` separately rather than `accountValidator.validate()` which every other strategy uses to atomically capture vaultId, account, dstAddress, and gasFee in one snapshot, so the same `AccountValidator` is now exercised through two different validation contracts which means any future change to `validate()` silently does not apply to the regular send path

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -0,0 +1,402 @@
...
+
+                transactionRepository.addTransaction(transaction)
+
+                navigator.route(Route.VerifySend(transactionId = transaction.id, vaultId = vaultId))
+            } catch (e: InvalidTransactionDataException) {
+                showError(e.text)
+            } catch (e: CancellationException) {
+                throw e
+            } catch (e: Exception) {
+                showError(e.message?.asUiText() ?: UiText.Empty)
```
rkokhatskyi: `e.message?.asUiText() ?: UiText.Empty` renders an empty `UiAlertDialog` body when the caught exception has no message such as a `NullPointerException`, deserialisation failure, or JNI throw, so the user sees an Error dialog with a blank line and a Try Again button with no context, swap `UiText.Empty` for a localised resource like `R.string.dialog_default_error_body` and apply the same change in the matching catches of every other strategy in this PR which all share this fallback

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -0,0 +1,402 @@
...
+            emitFocusField(SendFocusField.ADDRESS)
+            return
+        }
+        if (tokenAmountFieldState.text.isBlank()) {
+            expandSection(SendSections.Amount)
+            emitFocusField(SendFocusField.AMOUNT)
+            return
+        }
+
+        scope.launch {
```
rkokhatskyi: `onClickContinue` and every strategy `submit()` immediately call `scope.launch { showLoading(); ... }` without checking `uiState.value.isLoading` or holding a `Job?`, so a quick double tap launches two coroutines that both pass validation, write a transaction via `addTransaction(...)`, and call `navigator.route(...)` producing a duplicate persisted record and stacked Verify screens, guard with `if (uiState.value.isLoading) return` here or hold and cancel the previous job

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/MintStrategy.kt
```diff
@@ -0,0 +1,174 @@
...
+                            isDeposit = true,
+                            transactionType = TransactionType.TRANSACTION_TYPE_GENERIC_CONTRACT,
+                        )
+                    }
+
+                val tokenContract =
+                    when (defiTypeProvider()) {
+                        DeFiNavActions.MINT_YRUNE -> YRUNE_CONTRACT
+                        DeFiNavActions.MINT_YTCY -> YTCY_CONTRACT
+                        else -> throw RuntimeException("Invalid Deposit Parameter ")
```
rkokhatskyi: The `RuntimeException("Invalid Deposit Parameter ")` thrown here (note the trailing space) falls through the generic `catch (e: Exception)` at line 167 where `e.message?.asUiText()` surfaces the raw English string verbatim in a `UiAlertDialog`, swap this for `throw InvalidTransactionDataException(UiText.StringResource(R.string.error_unknown))` so the existing `InvalidTransactionDataException` arm renders a localised message

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/StakeStrategy.kt
```diff
@@ -0,0 +1,262 @@
...
+                                vaultId = vaultId,
+                                selectedToken = selectedToken,
+                                srcAddress = srcAddress,
+                                dstAddress = dstAddress,
+                                tokenAmountInt = tokenAmountInt,
+                                gasFee = gasFee,
+                                chain = chain,
+                            )
+
+                        else -> error("DeFi Type not supported ${defiTypeProvider()?.type}")
```
rkokhatskyi: The `error("DeFi Type not supported ${defiTypeProvider()?.type}")` call throws an `IllegalStateException` whose message embeds the internal `DeFiNavActions.type` identifier and is then rendered to the user verbatim by `e.message?.asUiText()` at line 150, switch to `throw InvalidTransactionDataException(UiText.StringResource(R.string.error_unknown))` so the dialog shows a localised string instead of leaking the enum value

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/UnstakeStrategy.kt
```diff
@@ -0,0 +1,344 @@
...
+                                selectedToken = ruji.token,
+                                srcAddress = srcAddress,
+                                dstAddress = dstAddress,
+                                tokenAmountInt = tokenAmountInt,
+                                gasFee = gasFee,
+                                chain = chain,
+                            )
+                        }
+
+                        else -> error("DeFi Type not supported ${defiTypeProvider()?.type}")
```
rkokhatskyi: Same pattern as `StakeStrategy:137`, `error("DeFi Type not supported ${defiTypeProvider()?.type}")` surfaces an unlocalised English string with an internal enum identifier through `e.message?.asUiText()` at line 173, replace with `throw InvalidTransactionDataException(UiText.StringResource(R.string.error_unknown))`

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/WithdrawUsdcCircleStrategy.kt
```diff
@@ -0,0 +1,167 @@
...
+                            R.string.send_error_insufficient_native_balance_with_fees,
+                            listOf(selectedToken.ticker),
+                        )
+                    )
+                }
+
+                val memo =
+                    EthereumFunction.withdrawCircleMSCA(
+                        vaultAddress =
+                            nonDeFiAccount?.token?.address ?: error("Vault Address Empty"),
```
rkokhatskyi: The `error("Vault Address Empty")` here throws `IllegalStateException` whose message reaches the user via `e.message?.asUiText()` at line 161 the moment `nonDeFiAccount` is null, and the access at line 127 (`val nativeCoin = nonDeFiAccount.token`) only compiles because line 109 throws on null so a future tweak to that guard turns line 127 into a `NullPointerException`, bind `nonDeFiAccount` to a non null `val` once via `?: throw InvalidTransactionDataException(UiText.StringResource(R.string.send_error_no_token))` and reuse it

app/src/test/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategyTest.kt
```diff
@@ -0,0 +1,126 @@
...
+
+            build(this).submit()
+
+            assertEquals(SendSections.Amount, expandedSection)
+            assertEquals(SendFocusField.AMOUNT, emittedFocusField)
+            assertNull(lastError)
+        }
+
+    @Test
+    fun `submit with blank address does not launch a coroutine`() = runTest {
```
rkokhatskyi, line 115: The three tests here all exit before the coroutine launches, and across all eight strategy test files in this PR no test reaches `addTransaction(...)` or `navigator.route(...)`, so the constructed `Transaction`/`DepositTransaction` shape including `memo`, `wasmExecuteContractPayload`, and the contract address selection inside `MintStrategy`/`RedeemStrategy`/`StakeStrategy`/`UnstakeStrategy` is unverified, since memo and contract assembly is exactly what this refactor moves around adding at least one positive test per strategy that asserts on the produced transaction body would catch silent regressions

app/src/test/java/com/vultisig/wallet/ui/models/send/submit/WithdrawUsdcCircleStrategyTest.kt
```diff
@@ -0,0 +1,131 @@
...
+            )
+    }
+
+    private fun build(scope: CoroutineScope) =
+        WithdrawUsdcCircleStrategy(
+            scope = scope,
+            tokenAmountFieldState = tokenAmountFieldState,
+            accountValidator = accountValidator,
+            chainAccountAddressRepository = chainAccountAddressRepository,
+            accountsRepository = mockk<AccountsRepository>(relaxed = true),
```
rkokhatskyi: The relaxed mock for `accountsRepository` returns an empty `Flow` from `loadAddresses(...)` so `nonDeFiAccount` is always `null` and `nonDeFiBalance` is always `BigInteger.ZERO` regardless of what a future test sets up, which makes the `nonDeFiBalance < gasFee.value` guard at `WithdrawUsdcCircleStrategy:82` impossible to test against a realistic ETH account fixture, declare `accountsRepository` as a non relaxed `mockk()` and stub `loadAddresses` per test

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Done in 3e84e04fc. The dispatch now enumerates every `DeFiNavActions` value explicitly, so the Send flows (`null`, `DEPOSIT_USDC_CIRCLE`, `STAKE_CACAO`, `UNSTAKE_CACAO`, `ADD_LP`, `REMOVE_LP`, `FREEZE_TRX`, `UNFREEZE_TRX`) are routed through `sendStrategy` by name. A new variant added later will fail to compile rather than silently fall through.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Done in 3e84e04fc. `fun send()` is gone; the three callers in `SendFormViewModelSubmitEvmTest` now call `onClickContinue()` directly. With `defiType = null` they hit the same `sendStrategy.submit()` path, so the assertions are unchanged.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Done in 3e84e04fc. `getSpecific(...)` in `DefaultSendStrategy` is now wrapped in `withContext(Dispatchers.IO)` to match every other strategy.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Done in 3e84e04fc. `DefaultSendStrategy` now consumes `accountValidator.validate()` for vaultId / selectedAccount / chain / dstAddress / gasFee, identical to the deposit strategies. The label-derivation and chain-side validations run after `validate()`.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Done in 3e84e04fc. Every strategy's generic `catch (e: Exception)` now falls back to `UiText.StringResource(R.string.dialog_default_error_body)` ("Something went wrong") instead of `UiText.Empty`, so the dialog body is never rendered blank.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Done in 3e84e04fc. Each strategy keeps a `private var submitJob: Job?` and bails early via `if (submitJob?.isActive == true) return` before launching. The check is synchronous, so a double-tap can't slip both coroutines past it.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/MintStrategy.kt
aminsato: Done in 3e84e04fc. The unreachable `else` arm in the contract `when` now throws `InvalidTransactionDataException(UiText.StringResource(R.string.dialog_default_error_body))`, so any defi-type drift surfaces a localised error rather than the bare English `RuntimeException` message.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/StakeStrategy.kt
aminsato: Done in 3e84e04fc. The `error("DeFi Type not supported …")` arm is replaced with `throw InvalidTransactionDataException(UiText.StringResource(R.string.dialog_default_error_body))`, so the dialog no longer leaks the `DeFiNavActions.type` value.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/UnstakeStrategy.kt
aminsato: Done in 3e84e04fc. Same fix as `StakeStrategy:137` — `error(...)` replaced with `InvalidTransactionDataException(R.string.dialog_default_error_body)`.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/WithdrawUsdcCircleStrategy.kt
aminsato: Done in 3e84e04fc. `nonDeFiAccount` is now bound once via `?: throw InvalidTransactionDataException(UiText.StringResource(R.string.send_error_no_token))` and reused for both `nativeCoin` and the memo's `vaultAddress = nativeCoin.address`. The fragile `?.token?.address ?: error("Vault Address Empty")` chain is gone.

app/src/test/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategyTest.kt
aminsato: Done in 3e84e04fc. Every strategy test file (Default / Bond / Unbond / Mint / Redeem / Stake / Unstake / WithdrawUsdcCircle) gains at least one happy-path test that reaches `addTransaction(...)` and asserts on the produced body. Coverage adds: `Default` (Transaction shape, parsed amount, dst), `Bond` (`BOND:nodeAddress`), `Unbond` (`UNBOND:nodeAddress:amount`), `Mint` (`receive:rune:0.5` + `YRUNE_YTCY_AFFILIATE_CONTRACT`, plus a `MINT_YTCY` variant covering `YTCY_CONTRACT`), `Redeem` (`sell:contract:0.5` routed to `YRUNE_CONTRACT`), `Stake` (`bond:contract:amount` to `STAKING_RUJI_CONTRACT`, plus the legacy `TCY+` memo variant), `Unstake` (RUJI `withdraw` memo, TCY basis-point memo, RUJI rewards memo), and `WithdrawUsdcCircle` (encoded memo + MSCA dst + zero `srcTokenValue` + `OPERATION_CIRCLE_WITHDRAW`). The native-bound encoders are stubbed via `mockkObject(ThorchainFunctions)` / `mockkObject(EthereumFunction)` so JNI isn't loaded.

app/src/test/java/com/vultisig/wallet/ui/models/send/submit/WithdrawUsdcCircleStrategyTest.kt
aminsato: Done in 3e84e04fc. `accountsRepository` is no longer relaxed; each test now stubs `loadAddresses(...)` explicitly via `every { … } returns flowOf(…)`. A new positive test sets up an ETH-account fixture and reaches `addTransaction(...)`; a new negative test exercises the `nonDeFiAccount` missing path with `flowOf(emptyList())` and asserts `R.string.send_error_no_token`.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -2312,65 +1279,46 @@ constructor(
...
+                            .collect(accounts)
+                    }
+
+                else ->
+                    viewModelScope.launch {
+                        accountsRepository
+                            .loadDeFiAddresses(vaultId, false)
+                            .map { addrs -> addrs.flatMap { it.accounts } }
+                            .collect(accounts)
+                    }
```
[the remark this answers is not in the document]
aminsato: Thanks for the catch. This is pre-existing on `main` — `UNFREEZE_TRX` already falls into the `else -> loadDeFiAddresses(vaultId, false)` branch in the original dispatch (see `SendFormViewModel.kt:2355` on main). This PR is a refactor (lift-and-shift, no behavior changes), so I'm keeping the routing identical to main here. Happy to revisit in a separate, focused PR.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -0,0 +1,390 @@
...
+
+                    val tokenAmount = tokenAmountFieldState.text.toString().toBigDecimalOrNull()
+                    if (tokenAmount == null || tokenAmount <= BigDecimal.ZERO) {
+                        throw InvalidTransactionDataException(
+                            UiText.StringResource(R.string.send_error_no_amount)
+                        )
+                    }
+
+                    val tokenAmountInt =
+                        tokenAmount.movePointRight(selectedToken.decimal).toBigInteger()
```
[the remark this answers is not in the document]
aminsato: Thanks. The `movePointRight(selectedToken.decimal).toBigInteger()` pattern is pre-existing on `main` across every send/deposit path. This PR is a refactor (lift-and-shift), so swapping it for `toBigIntegerExact()` would change end-user behavior beyond the refactor's scope. Worth addressing in a separate, narrow PR with tests covering the truncation path.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/RedeemStrategy.kt
```diff
@@ -0,0 +1,194 @@
...
+                    val chain = validated.chain
+                    val dstAddress = validated.dstAddress
+                    val selectedAccount = validated.selectedAccount
+                    val gasFee = validated.gasFee
+
+                    if (!chainAccountAddressRepository.isValid(chain, dstAddress)) {
+                        throw InvalidTransactionDataException(
+                            UiText.StringResource(R.string.send_error_no_address)
+                        )
+                    }
```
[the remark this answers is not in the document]
aminsato: Thanks. This is pre-existing on `main` — the original `redeem()` at `SendFormViewModel.kt:2132` validates `dstAddress` and then builds the `DepositTransaction` with `dstAddress = tokenContract` in the exact same way. This PR is a refactor (lift-and-shift); the destination-vs-validation divergence belongs in a separate PR.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/StakeStrategy.kt
```diff
@@ -0,0 +1,273 @@
...
+                    if (nonDeFiBalance < gasFee.value) {
+                        throw InvalidTransactionDataException(
+                            UiText.StringResource(R.string.send_error_insufficient_balance)
+                        )
+                    }
+
+                    val selectedToken = selectedAccount.token
+                    val srcAddress = selectedToken.address
+                    val tokenAmountInt =
+                        tokenAmount.movePointRight(selectedToken.decimal).toBigInteger()
```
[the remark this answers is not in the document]
aminsato: Thanks. Same as the matching note on `DefaultSendStrategy` — `movePointRight(...).toBigInteger()` is pre-existing on `main`. This PR is a refactor (lift-and-shift), so I'd rather address the truncation semantics in a focused follow-up PR with tests.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/UnbondStrategy.kt
```diff
@@ -0,0 +1,177 @@
...
+                    if (tokenAmount == null || tokenAmount <= BigDecimal.ZERO) {
+                        throw InvalidTransactionDataException(
+                            UiText.StringResource(R.string.send_error_no_amount)
+                        )
+                    }
+
+                    val selectedToken = selectedAccount.token
+                    val selectedAddress = selectedToken.address
+                    val tokenAmountInt =
+                        tokenAmount.movePointRight(selectedToken.decimal).toBigInteger()
```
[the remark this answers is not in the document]
aminsato: Thanks. Same as the matching note on `DefaultSendStrategy` — `movePointRight(...).toBigInteger()` is pre-existing on `main`. This PR is a refactor (lift-and-shift), so I'd rather address the truncation semantics in a focused follow-up PR with tests.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/WithdrawUsdcCircleStrategy.kt
```diff
@@ -0,0 +1,177 @@
...
+                    if (nonDeFiBalance < gasFee.value) {
+                        throw InvalidTransactionDataException(
+                            UiText.StringResource(R.string.send_error_insufficient_balance)
+                        )
+                    }
+
+                    val selectedToken = selectedAccount.token
+                    val srcAddress = selectedToken.address
+                    val tokenAmountInt =
+                        tokenAmount.movePointRight(selectedToken.decimal).toBigInteger()
```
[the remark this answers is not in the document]
aminsato: Thanks. Same as the matching note on `DefaultSendStrategy` — `movePointRight(...).toBigInteger()` is pre-existing on `main`. This PR is a refactor (lift-and-shift), so I'd rather address the truncation semantics in a focused follow-up PR with tests.

johnnyluo (approved):
LGTM

PR #4417, opened 2026-05-05
fix(evm): handle non-2xx RPC responses without crashing balance fetch (#4414)

### LINKED ISSUE #4414: HyperEVM RPC returns 403 → NoTransformationFoundException crashes balance fetch; user sees 'Invalid QR code content'
## Summary

`https://api.vultisig.com/hyperevm/` is returning HTTP **403** with a `null` Content-Type. The Ktor client cannot transform the body into `RpcResponse` and throws `NoTransformationFoundException` from `EvmApiImp.getETHBalance`. On screen, the user sees a misleading error: **"Signing Error. Please try again. Invalid QR code content"** with an offline banner, even though the device is online and the QR code is valid — the real failure is the upstream RPC.

## Repro

1. Open the app and proceed to a HyperEVM-related flow (Keysign in this instance).
2. The app calls `getETHBalance` against `https://api.vultisig.com/hyperevm/`.
3. The endpoint replies `403` with no `Content-Type`.
4. Ktor cannot deserialize and throws.

## Logcat

\`\`\`
io.ktor.client.call.NoTransformationFoundException: Expected response body of the type
'class com.vultisig.wallet.data.api.models.RpcResponse' but was 'class io.ktor.utils.io.SourceByteReadChannel'
In response from \`https://api.vultisig.com/hyperevm/\`
Response status \`403\`
Response header \`ContentType: null\`
Request header \`Accept: */*\`
    at io.ktor.client.call.HttpClientCall.bodyNullable(HttpClientCall.kt:91)
    at com.vultisig.wallet.data.api.EvmApiImp.getETHBalance(EvmApi.kt:667)
    at com.vultisig.wallet.data.api.EvmApiImp\$getETHBalance\$1.invokeSuspend(Unknown Source:15)
    ...
\`\`\`

## Observed UI

- Yellow banner: **"You are offline"** (device is actually online).
- Title: **"Signing Error. Please try again. Invalid QR code content"**.
- Subtitle suggests checking WiFi / vault / pairing.
- Footer: Version 1.0.101 (101).

The wording does not match the actual failure (an RPC 403), which makes this hard to diagnose for users.

## Expected

- The HyperEVM RPC endpoint should not return 403 to the app.
- If an RPC call legitimately fails, the app should surface a network/RPC error, not "Invalid QR code content" or "You are offline".

## Environment

- App version: 1.0.101 (101)
- Endpoint: \`https://api.vultisig.com/hyperevm/\`

aminsato:
## Summary

- `https://api.vultisig.com/hyperevm/` was replying 403 with a null Content-Type, so the raw `.body<RpcResponse>()` in `EvmApi.fetch` threw `NoTransformationFoundException` out of `getETHBalance`. During keysign that bubbled up as a misleading "Invalid QR code content" + offline banner.
- Switched the shared `fetch<T>` helper to `bodyOrThrow<T>()` so non-2xx responses now throw a typed `NetworkException(statusCode, body)`.
- Extended `getBalance`'s catch block to handle `NetworkException`, returning `BigInteger.ZERO` and logging the status — matches the existing `SocketTimeoutException` recovery path so `BalanceRepository` / keysign refresh stop crashing on RPC errors.
- Added unit tests: 403 reproduction returns ZERO without throwing, and 200 still parses the result so the new catch can't accidentally swallow real responses.

Note: the misleading "Invalid QR code content" copy itself comes from a separate keysign error-classification path; with this fix the balance call no longer throws, so that classification is never reached. Tracing the message itself is out of scope for this ticket.

Closes #4414

## Test plan

- [x] `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.EvmApiBalanceTest"` passes
- [x] `./gradlew :data:compileDebugKotlin :app:compileDebugKotlin` clean
- [ ] Manual: trigger a HyperEVM keysign while the upstream RPC is returning 403 — verify the flow no longer surfaces "Invalid QR code content" / offline banner
- [ ] Manual: regression check — a working chain (e.g. ETH) still shows correct balance

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for blockchain network requests: network failures and invalid RPC responses are now caught, RPC error status/messages are logged, and balance retrieval returns zero instead of throwing.

* **Tests**
  * Added automated tests for balance retrieval covering failed RPC responses (e.g., forbidden/empty body) and successful RPC responses with parsed numeric results.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -18,7 +18,9 @@ import com.vultisig.wallet.data.common.stripHexPrefix
 import com.vultisig.wallet.data.common.toKeccak256
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.utils.NetworkException
 import com.vultisig.wallet.data.utils.Numeric
+import com.vultisig.wallet.data.utils.bodyOrThrow
 import io.ktor.client.HttpClient
 import io.ktor.client.call.body
 import io.ktor.client.request.post
@@ -124,6 +126,13 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
             else getERC20Balance(coin.address, coin.contractAddress)
         } catch (e: SocketTimeoutException) {
             Timber.d("request time out, message: ${e.message}")
+            BigInteger.ZERO
+        } catch (e: NetworkException) {
+            Timber.d(
+                "RPC error fetching balance: status=%d message=%s",
+                e.httpStatusCode,
+                e.message,
+            )
             BigInteger.ZERO
         }
     }
@@ -487,7 +496,9 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
         params: JsonArray,
         id: Int = 1,
     ): T =
-        http.post(rpcUrl) { setBody(RpcPayload(method = method, params = params, id = id)) }.body()
+        http
+            .post(rpcUrl) { setBody(RpcPayload(method = method, params = params, id = id)) }
+            .bodyOrThrow()
 
     private fun generateCustomTokenPayload(contractAddress: String): Pair<RpcPayload, RpcPayload> {
         val payload1 =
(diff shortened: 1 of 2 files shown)


aminsato:
@coderabbitai full review

johnnyluo (approved):
LGTM

PR #4424, opened 2026-05-06
feat(home): add 'Buy $VULT' promo banner (#4398)

### LINKED ISSUE #4398: [ADD] 'Buy $VULT' promo banner on home screen (Figma)
## What
Add the "Buy $VULT — And save on swap fees" promo banner from Figma to the home/balance screen.

## Figma
https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=75116-7357

## Spec
- Surface: `backgrounds/surface-1` (#061b3a), `borders/light` (#11284a) 1px border, 12px radius, 24px padding
- Title: "Buy $VULT" — Brockmann Medium 12 / line-height 16 / letter-spacing 0.12 / color `text/tertiary` (#8295ae)
- Subtitle: "And save on swap fees" — Brockmann Medium 14 / line-height 20 / color `text/primary` (#f0f4fc)
- CTA "Buy $VULT" button — `buttons/success` (#33e6bf) background, `text/button/inverse` (#02122b) text, Brockmann Medium 12, radius 30px, padding 16x8
- Decorative fanned-coin illustration on right side (export from Figma)
- Dismiss glass-close "X" button top-right (40px container, 16px icon)

## Behavior
- Tap CTA "Buy $VULT" → opens in-wallet swap flow with VULT as destination token (confirm route with product if external link is preferred)
- Tap "X" → dismisses banner; persist dismissal locally so banner does not reappear on this device

## Acceptance
- [ ] Banner renders on home/balance screen below header
- [ ] Visual matches Figma 1:1 (colors, fonts, radii, spacing, illustration)
- [ ] CTA tap routes to VULT buy/swap flow
- [ ] Dismiss persists across app restart
- [ ] Banner gracefully scales across supported screen widths

## Cross-platform parity
- iOS — this issue
- Android — companion ticket
- Windows + Extension — companion ticket

aminsato:
## Summary
Closes #4398. Adds the Figma "Buy \$VULT — And save on swap fees" banner to the home screen pager next to the existing migration and Follow X banners.

- New `BuyVultBanner` composable using `SetBackgroundBanner` with the exported Figma art and the title/subtitle/CTA from the design.
- CTA → in-wallet swap with VULT (Ethereum) as the destination token; auto-enables the VULT token if the vault doesn't have it.
- Per-banner X dismiss persists via `VaultDataStoreRepository` (new `buy_vult_banner_dismissed` boolean preference) so the banner stays hidden across app restarts on this device.
- Strings added across all 10 supported locales (title/CTA marked non-translatable; subtitle translated using each locale's existing swap terminology).

## Test plan
- [ ] Banner appears under the header on the home/balance screen
- [ ] Tapping "Buy \$VULT" opens the swap flow with VULT preselected as destination on Ethereum
- [ ] Tapping the X hides the banner; killing/relaunching the app keeps it hidden
- [ ] FollowX and migration banners still dismiss with their existing temporary behavior
- [ ] Visuals match Figma at 343dp banner width and scale on smaller/larger devices

## Summary by CodeRabbit

* **New Features**
  * Dismissible "Buy VULT" promotional banner on the home screen with primary action to start buying VULT.
  * Banner dismissal persists across sessions via new persistent flag.
  * Buy flow now ensures VULT is enabled for the vault (when applicable) before opening the swap flow.
  * New banner UI component and callbacks integrated into the home pager.
* **Tests**
  * Added tests covering banner visibility, dismissal persistence, and buy-VULT navigation scenarios.
* **Localization**
  * Banner text added for DE, ES, IT, KO, NL, PT, RU, ZH, HR.


aminsato:
@coderabbitai full review

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -290,6 +300,33 @@ constructor(
...
+                Timber.e(e, "Failed to ensure VULT token before swap navigation")
+            }
+            navigator.route(
+                Route.Swap(
+                    vaultId = vaultId,
+                    chainId = Chain.Ethereum.id,
+                    dstTokenId = Coins.Ethereum.VULT.id,
+                )
+            )
+        }
```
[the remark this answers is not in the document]
rkokhatskyi: @aminsato

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -157,6 +159,14 @@ constructor(
     init {
         collectCryptoConnectionType()
         collectLastOpenedVault()
+        collectBuyVultBannerDismissed()
+    }
+
+    private fun collectBuyVultBannerDismissed() {
+        vaultDataStoreRepository
+            .readBuyVultBannerDismissed()
+            .onEach { dismissed -> uiState.update { it.copy(showBuyVultBanner = !dismissed) } }
```
rkokhatskyi, line 169: `showBuyVultBanner = !dismissed` is derived only from the persisted dismiss flag with no check against `vault.coins.any { it.id == Coins.Ethereum.VULT.id }` so a user already holding VULT keeps seeing the buy CTA on every cold start until they manually close the banner

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -290,6 +300,38 @@ constructor(
         }
     }
 
+    fun buyVult() {
```
rkokhatskyi, line 308: the existing `buy()` two functions above routes to `Route.OnRamp` for a fiat on-ramp whereas `buyVult()` routes to `Route.Swap`, so the names suggest a shared product flow that the destinations do not share, rename to something like `swapToVult()` to reflect what it actually does

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -290,6 +300,38 @@ constructor(
...
+        val vaultId = vaultId ?: return
+        viewModelScope.launch {
+            val vultEnabled =
+                try {
+                    val vault = withContext(Dispatchers.IO) { vaultRepository.get(vaultId) }
+                    if (vault != null && vault.coins.none { it.id == Coins.Ethereum.VULT.id }) {
```
rkokhatskyi: this skips the `vault.coins.any { it.chain == Chain.Ethereum }` guard that `enableVultTokenIfNeeded` enforces above, so a vault that never opted into Ethereum gets a VULT row inserted with no native ETH for gas which leaves `firstSendSrc` unable to resolve a source token on the swap screen

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -290,6 +300,38 @@ constructor(
...
+                        withContext(Dispatchers.IO) {
+                            enableTokenUseCase(vaultId, Coins.Ethereum.VULT)
```
rkokhatskyi: `enableVultTokenIfNeeded` already runs in `loadData` and performs the same `vault.coins.none` check plus `enableTokenUseCase` invocation so by the time the user taps the banner the token is enabled in nearly every case, factor a shared `suspend fun ensureVultEnabled(vaultId): Boolean` helper that both call sites use

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -290,6 +300,38 @@ constructor(
...
+                        }
+                    }
+                    true
```
rkokhatskyi: `enableTokenUseCase` returns `null` when `addTokenToVault` hits `SQLiteConstraintException` to signal the row was rejected, but `vultEnabled` is set to `true` here whenever no exception is thrown so `dstTokenId` is propagated to `Route.Swap` even when the insert was actually skipped, drive the boolean from the `String?` return value instead

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -290,6 +300,38 @@ constructor(
...
+                    throw e
+                } catch (e: Exception) {
+                    Timber.e(e, "Failed to ensure VULT token before swap navigation")
+                    false
+                }
+            navigator.route(
+                Route.Swap(
+                    vaultId = vaultId,
+                    chainId = Chain.Ethereum.id,
+                    dstTokenId = Coins.Ethereum.VULT.id.takeIf { vultEnabled },
```
rkokhatskyi: on the catch path navigation still proceeds with `dstTokenId = null` while `chainId = Chain.Ethereum.id`, so the swap form opens with both `selectedSrc` and `selectedDst` resolving to native ETH which triggers `SwapException.SameAssets` on every amount edit with no user feedback, either skip navigation when `vultEnabled` is false or surface a snackbar via the already-injected `SnackbarFlow`

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -290,6 +300,38 @@ constructor(
...
+                )
+            )
+        }
+    }
+
+    fun dismissBuyVultBanner() {
```
rkokhatskyi, line 327: no unit tests added for `buyVult`, `dismissBuyVultBanner`, or `collectBuyVultBannerDismissed` while the existing `VaultAccountsViewModelTest` already mocks `enableTokenUseCase` and `vaultDataStoreRepository`, so the success, constraint-conflict, catch-path, and dismiss-persistence cases are all easy to cover

app/src/main/java/com/vultisig/wallet/ui/screens/home/VaultAccountsScreen.kt
```diff
@@ -284,7 +288,10 @@ internal fun VaultAccountsScreen(
                                 modifier =
                                     Modifier.padding(horizontal = 16.dp).padding(bottom = 16.dp),
                                 hasMigration = state.showMigration,
+                                showBuyVult = state.showBuyVultBanner,
```
rkokhatskyi, line 292: the surrounding `AnimatedVisibility` is gated by `state.isBannerVisible` which `tempRemoveBanner` flips to false from the migration and follow-x close buttons, so dismissing either of those collapses this BuyVult banner for the session even though `buy_vult_banner_dismissed` is never persisted, scope the gate to migration/follow-x and let BuyVult be controlled solely by `state.showBuyVultBanner`

data/src/main/kotlin/com/vultisig/wallet/data/repositories/VaultDataStoreRepository.kt
```diff
@@ -66,5 +79,8 @@ constructor(private val appDataStore: AppDataStore) : VaultDataStoreRepository {
 
         fun onGlobalBackupReminderStatusKey() =
             intPreferencesKey(name = "global_backup_reminder_status")
+
+        fun onBuyVultBannerDismissedKey() =
+            booleanPreferencesKey(name = "buy_vult_banner_dismissed")
```
rkokhatskyi, line 84: this is a single global key whereas the sibling backup keys above use `onVaultBackupStatusKey(vaultId)` style scoping, so dismissing on one vault hides the banner on every other (and newly imported) vault on the device, confirm whether device-wide is the intended scope

app/src/main/res/values-ru/strings.xml
```diff
@@ -682,6 +682,7 @@
     <string name="invite_to_x_banner_title">Vultisig строит вместе с вами.</string>
     <string name="invite_to_x_banner_desc">Подпишитесь на нас в X</string>
     <string name="invite_to_x_banner_button">Подписаться на %s</string>
+    <string name="buy_vult_banner_desc">И экономьте на комиссиях обмена</string>
```
rkokhatskyi: `И экономьте на комиссиях обмена` reads as ungrammatical Russian since the existing `swap_form_estimated_fees_title` uses the preposition `за` (`Комиссия за обмен`), use `И экономьте на комиссиях за обмен` to match

app/src/main/res/values-nl/strings.xml
```diff
@@ -678,6 +678,7 @@
     <string name="invite_to_x_banner_title">Vultisig bouwt met je mee.</string>
     <string name="invite_to_x_banner_desc">Volg ons op X</string>
     <string name="invite_to_x_banner_button">Volg %s</string>
+    <string name="buy_vult_banner_desc">En bespaar op swap-kosten</string>
```
rkokhatskyi: existing nl strings consistently use `wissel` and `wisselkosten` for swap fees (e.g. `swap_form_estimated_fees_title` is `Wisselkosten`, `transaction_type_button_swap` is `Wisselen`) so this should read `En bespaar op wisselkosten` rather than the loanword `swap-kosten`

app/src/main/res/values-hr/strings.xml
```diff
@@ -684,6 +684,7 @@
     <string name="invite_to_x_banner_title">Vultisig gradi s vama.</string>
     <string name="invite_to_x_banner_desc">Pratite nas na X</string>
     <string name="invite_to_x_banner_button">Pratite %s</string>
+    <string name="buy_vult_banner_desc">I uštedi na naknadama za swap</string>
```
rkokhatskyi: existing hr strings prefer `zamjena` and `zamijeni` (e.g. `swap_form_estimated_fees_title` is `Naknada za zamjenu`, `chain_account_view_swap` is `ZAMIJENI`) so use `I uštedi na naknadama za zamjenu` rather than the loanword `swap`

app/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -1080,6 +1080,7 @@
     <string name="invite_to_x_banner_title">Vultisig 正在与您一起构建。</string>
     <string name="invite_to_x_banner_desc">在 X 上关注我们</string>
     <string name="invite_to_x_banner_button">关注 %s</string>
+    <string name="buy_vult_banner_desc">并节省交换手续费</string>
```
rkokhatskyi: existing `swap_form_estimated_fees_title` is `交换费用` whereas this introduces `交换手续费`, prefer the existing term so use `并节省交换费用`

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
johnnyluo: That's fine , even user hold VULT token , we should still remind them to buy

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Fair point. I considered renaming but kept `buyVult` because it mirrors the user-facing CTA ("Buy $VULT") and the public callback name `onBuyVultClick`. Renaming the VM function alone would create a mismatch with the screen wiring and the UI string. I substantially reworked the body in 28b/the next commit so the implementation is no longer a thin alias for `swap()` (Ethereum-chain guard, return-value driven `vultEnabled`, navigation gated on success), which I think addresses the underlying confusion you raised.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Good catch. Added an Ethereum-chain guard to mirror `enableVultTokenIfNeeded` — `ensureVultEnabled` now returns `false` when the vault has no Ethereum coins, and `buyVult` short-circuits before navigation, so a vault with no native ETH for gas never lands on a broken swap form.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Done — extracted `private suspend fun ensureVultEnabled(vault: Vault): Boolean` and routed both `enableVultTokenIfNeeded` (init path) and `buyVult` (CTA path) through it. The early-return on already-enabled VULT now also avoids the redundant DB call when the user taps the banner after the init pass has already enabled the token.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Fixed. `ensureVultEnabled` now drives the boolean from the use-case return value (`enableTokenUseCase(...) != null`), so a constraint-rejected insert returns `false` and `buyVult` skips navigation instead of routing with `dstTokenId = null`.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Fixed. Navigation is now gated on `ensureVultEnabled(vault)` returning `true`. On the failure paths (no Ethereum chain, constraint rejection, or thrown exception) `buyVult` logs the cause and returns early instead of opening the swap form, so the SameAssets case can no longer happen. I went with skip-and-log rather than a snackbar to avoid adding a new translatable error string for a path the auto-enable already covers in nearly every real-world case.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Added unit tests in `VaultAccountsViewModelTest`: `dismissBuyVultBanner persists dismissed flag`, `showBuyVultBanner` reflects the persisted flag in both directions, and four `buyVult` cases covering the success path with VULT pre-enabled, the no-Ethereum guard, the enable-then-navigate happy path, and the constraint-rejected (`enableTokenUseCase` returns null) skip-navigation path.

app/src/main/java/com/vultisig/wallet/ui/screens/home/VaultAccountsScreen.kt
aminsato: Fixed. Split the gating: the outer `AnimatedVisibility` now opens whenever `state.isBannerVisible || state.showBuyVultBanner`, and inside `HomepagePager` the Migration and FollowX items are gated on a new `showSessionBanners` flag (= `state.isBannerVisible`) while BuyVult stays gated on `state.showBuyVultBanner`. So pressing X on Migration/FollowX flips `isBannerVisible` for the session as before but no longer collapses BuyVult.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/VaultDataStoreRepository.kt
aminsato: Device-wide is intentional. The CTA is a global product promo (not a per-vault state) and Johnny confirmed in the other thread that we still want to nag users who already hold VULT, so dismissing it on one vault should hide it on every vault on the device. Per-vault scoping would defeat that.

app/src/main/res/values-ru/strings.xml
aminsato: Updated to `И экономьте на комиссиях за обмен` to align with `swap_form_estimated_fees_title`.

app/src/main/res/values-nl/strings.xml
aminsato: Updated to `En bespaar op wisselkosten` to match the existing `wisselkosten`/`Wisselen` terminology.

app/src/main/res/values-hr/strings.xml
aminsato: Updated to `I uštedi na naknadama za zamjenu` to align with `Naknada za zamjenu` / `ZAMIJENI`.

app/src/main/res/values-zh-rCN/strings.xml
aminsato: Updated to `并节省交换费用` to match the existing `swap_form_estimated_fees_title` term.

johnnyluo (approved):
LGTM

PR #4432, opened 2026-05-07
refactor(send): finish breaking up SendFormViewModel

aminsato:
## Summary
Finishes the SendFormViewModel decomposition. Moves the remaining heavy logic into focused services and brings the VM down to roughly a third of its original size.

| File | Before | After |
|------|-------:|------:|
| `SendFormViewModel.kt` | 3415 (pre Phase 1) / 1879 (start of this PR) | **1151** |

### Phase 4 — `AmountFractionManager`
Owns `chooseMaxTokenAmount`, `choosePercentageAmount`, and the two private helpers. EVM short-circuit, UNFREEZE_TRX path, vault/account guards, gas-fee recalculation. Behavior change: rethrows `CancellationException` in the deep `catch (e: Exception)`, matching every Phase 3 strategy and the project's coding-quality rule (the original swallow was a latent bug).

### Phase 5 — `TronStakingService`
Owns `setResourceType` (was `setTronResourceType`), `initIfStakingType`, `currentFrozenBalance`, `isStakingType`, the frozen-balance loader, and the memo application. Lifts ~80 LOC out of the VM and removes `loadTronFrozenBalancesJob`.

### Phase 6 — `AccountsLoader`
Owns `load(vaultId)` — the big `when (defiType)` switch — plus the two specialized loaders (`loadCircleUSDCAccount`, `loadRewardsAccount`). Lifts ~110 LOC and removes `loadAccountsJob`.

### Phase 7 — `TokenPreselectionService`
Owns `preSelect`, `findPreselectedToken`, `findDeFiPreselectedToken` — the accounts-flow observer that picks an initial token based on `defiType`. Lifts ~95 LOC and removes `preSelectTokenJob`.

### Phase 8 — `GasFeeOrchestrator`
The biggest one. Owns `calculateGasTokenBalance`, `calculateGasFees`, `collectPlanFee`, `collectMaxAmount`, `collectEstimatedFee`, `calculateSpecific`, the `recalculateGasFee` flow, and the `GasFeeInput` / `PlanFeeInput` data classes. Exposes `start()` (called once from VM `init`) and `refresh()` (delegated from VM `refreshGasFee()`). `adjustGasFee` is now a package-level top-level function shared by the orchestrator and `AmountFractionManager`. Lifts ~390 LOC.

## Tests

Per-service unit tests added for every new service:

- `AmountFractionManagerTest` — 9 tests
- `TronStakingServiceTest` — 12 tests
- `AccountsLoaderTest` — 8 tests
- `TokenPreselectionServiceTest` — 9 tests
- `GasFeeOrchestratorTest` — 9 tests

Plus the pre-existing send.* suite (`SendFormViewModelTronStakingTest`, `*InitTest`, `*AddressTest`, `*AmountTest`, `*SubmitEvmTest`, the Phase 1–3 manager/strategy suites) that exercises the VM's preserved public surface end-to-end.

Total send.* test count: **194** (up from 146 before this PR). All green. ktfmt clean.

## Behavior changes
- `AmountFractionManager.calculatePercentageWithAccurateFee` rethrows `CancellationException` (was silently swallowed in the VM).
- No other observable changes — every public VM method retains the same signature and side effects.

## Test plan
- [x] `./gradlew :app:ktfmtFormat`
- [x] `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.send.*"` — 194 tests
- [ ] Manual smoke: send a regular transfer; 25/50/75/Max amounts on native + non-native tokens
- [ ] Manual smoke: deposit flows that trigger gas recalc (Bond, Stake RUJI/TCY/STCY, Mint/Redeem yRUNE/yTCY)
- [ ] Manual smoke: TRX freeze/unfreeze — Max + percentages on a frozen balance, resource-type toggle
- [ ] Manual smoke: Circle USDC withdraw, RUJI withdraw rewards
- [ ] Manual smoke: QR code → preselected chain switching
- [ ] Manual smoke: pull-to-refresh on the gas fee

## Summary by CodeRabbit

* **New Features**
  * Centralized gas-fee orchestrator with receiver-aware fee logic, UTXO/EVM handling, and explicit refresh.
  * Automatic token preselection during send flows.
  * Tron staking/unfreeze UI flows with frozen-balance loading, resource-type controls, and memo handling.
  * Improved account-loading behavior for additional DeFi/withdraw scenarios.

* **Refactor**
  * Amount/percentage selection extracted to a dedicated manager with fee-accurate computation and improved concurrency.
  * Send form delegates gas, preselection, account loading, and Tron staking concerns to new services.

* **Tests**
  * New unit tests for fee orchestration, amount selection, preselection, accounts loading, and Tron staking.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4432)


aminsato:
@coderabbitai full review

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -0,0 +1,143 @@
...
+                            .loadDeFiAddresses(vaultId, false)
+                            .map { addrs -> addrs.flatMap { it.accounts } }
+                            .collect(accounts)
+                    }
+            }
+    }
+
+    private suspend fun loadCircleUSDCAccount(vaultId: VaultId) {
+        val accountsLoaded =
+            accountsRepository.loadAddresses(vaultId).firstOrNull()?.flatMap { it.accounts }
```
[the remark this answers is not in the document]
aminsato: Skipped — pre-existing semantics, not introduced by this refactor.

`loadCircleUSDCAccount` and `loadRewardsAccount` were copied verbatim from the SendFormViewModel; the `loadAddresses(...).firstOrNull()` pattern predates this PR. Restructuring them to a long-lived `collect` would re-execute the staking-details lookup (`stakingDetailsRepository.getStakingDetailsById` / `getStakingDetailsByCoindId`) on every emission and re-publish the accounts list each time, which is a behavior change beyond a refactor's scope. Tracking this for a separate PR with manual smoke-testing of the rewards/Circle flows.

app/src/test/java/com/vultisig/wallet/ui/models/send/AccountsLoaderTest.kt
```diff
@@ -0,0 +1,325 @@
...
+import com.vultisig.wallet.ui.screens.v2.defi.model.DeFiNavActions
+import io.mockk.coEvery
+import io.mockk.coVerify
+import io.mockk.every
+import io.mockk.mockk
+import java.math.BigDecimal
+import java.math.BigInteger
+import java.util.Date
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
```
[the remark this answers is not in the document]
aminsato: Skipped — `kotlin.test` is the established convention in `app/src/test/.../send/`. A grep across the existing send.* test suite shows ~47 imports of `kotlin.test.*` against 3 of `io.kotest.*`; every Phase 1–3 manager and strategy test (`AmountManagerTest`, `AddressManagerTest`, `ChainValidationServiceTest`, the eight strategy tests, etc.) uses `kotlin.test`. Switching only the new test files to Kotest would split the convention. Happy to do a sweep of the whole `send.*` suite to Kotest in a follow-up if the team wants to enforce CLAUDE.md's rule project-wide, but that's not a one-PR change for this folder.

app/src/test/java/com/vultisig/wallet/ui/models/send/TokenPreselectionServiceTest.kt
```diff
@@ -0,0 +1,243 @@
...
+import com.vultisig.wallet.data.blockchain.thorchain.RujiStakingService.Companion.RUJI_REWARDS_COIN
+import com.vultisig.wallet.data.models.Account
+import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.models.Coins
+import com.vultisig.wallet.data.models.TokenValue
+import com.vultisig.wallet.ui.screens.v2.defi.model.DeFiNavActions
+import java.math.BigInteger
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
```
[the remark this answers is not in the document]
aminsato: Skipped for the same reason as the AccountsLoaderTest thread — `kotlin.test` is the established convention in this folder; switching only the new files would split the convention.

app/src/main/java/com/vultisig/wallet/ui/models/send/TokenPreselectionService.kt
```diff
@@ -0,0 +1,136 @@
...
+                                preSelectedChainIds,
+                                preSelectedTokenId,
+                            )
+                        }
+
+                    Timber.d("Found a new token to pre select %s", preSelectedToken)
+
+                    // if user hasn't yet selected any token, preselect found token
+                    if (
+                        (forcePreselection || selectedTokenProvider() == null) &&
```
rkokhatskyi: With `forcePreselection = true`, the unbounded `accounts.collect` invokes `onTokenSelected` on every emission, so each accounts hydration tick after the first calls `selectToken` which runs `amountManager.resetUserInputCache()` and wipes the user's typed token and fiat values whenever a QR scan or address book selection triggered the force flow

app/src/main/java/com/vultisig/wallet/ui/models/send/TokenPreselectionService.kt
```diff
@@ -0,0 +1,136 @@
...
+
+            DeFiNavActions.MINT_YRUNE -> Coins.ThorChain.RUNE
+            DeFiNavActions.MINT_YTCY -> Coins.ThorChain.TCY
+            DeFiNavActions.BOND -> Coins.ThorChain.RUNE
+            DeFiNavActions.UNBOND -> Coins.ThorChain.RUNE
+            DeFiNavActions.WITHDRAW_RUJI -> RUJI_REWARDS_COIN
+            DeFiNavActions.REDEEM_YRUNE -> Coins.ThorChain.yRUNE
+            DeFiNavActions.REDEEM_YTCY -> Coins.ThorChain.yTCY
+            DeFiNavActions.DEPOSIT_USDC_CIRCLE -> Coins.Ethereum.USDC
+            DeFiNavActions.WITHDRAW_USDC_CIRCLE -> Coins.Ethereum.USDC
```
rkokhatskyi, line 130: `findDeFiPreselectedToken` returns static templates such as `Coins.Ethereum.USDC` here or `RUJI_REWARDS_COIN` at line 120 when accounts is empty, so the first emission sets `selectedToken` to a token that lacks the MSCA or thorchain address, which then sticks since the gating at line 56 sees `selectedTokenProvider()` as not null on later emissions and never replaces the template with the real account that `AccountsLoader` publishes seconds later

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -0,0 +1,148 @@
...
+        val ethereumAccount =
+            accountsLoaded?.find { it.token.id.equals(Coins.Ethereum.ETH.id, true) }
+                ?: Account(
+                    token = Coins.Ethereum.ETH,
+                    tokenValue = TokenValue(BigInteger.ZERO, Coins.Ethereum.ETH),
+                    fiatValue = null,
+                    price = null,
+                )
+
+        val usdc = Coins.Ethereum.USDC.copy(address = ethereumAccount.token.address)
```
rkokhatskyi, line 92: When `loadAddresses(vaultId).firstOrNull()` returns null the fallback `Account(token = Coins.Ethereum.ETH, ...)` carries an empty `address` since `Coins.Ethereum.ETH` has no address bound, so `usdc.copy(address = ethereumAccount.token.address)` here produces a USDC token whose `address` is empty and any later submit through `WithdrawUsdcCircleStrategy` fails at the validator or broadcast layer with an unhelpful error

app/src/main/java/com/vultisig/wallet/ui/models/send/TronStakingService.kt
```diff
@@ -0,0 +1,137 @@
...
+        if (uiState.value.tronResourceType == type) return
+        uiState.update {
+            it.copy(
+                tronResourceType = type,
+                tronBalanceAvailableOverride = null,
+                selectedAmountFraction = null,
+            )
+        }
+        applyStakingMemo(type)
+        tokenAmountFieldState.clearText()
```
rkokhatskyi, line 65: `setResourceType` clears `tokenAmountFieldState` and `fiatAmountFieldState` here whereas `AmountFractionManager.chooseAmountFractionJob` is owned by a different service after this PR, so a percentage selection still suspended in `calculatePercentageWithAccurateFee` will run `setTextAndPlaceCursorAtEnd` after this returns and overwrite the cleared field with the amount computed for the previous resource type, since neither service exposes a way to cancel the other's job

app/src/main/java/com/vultisig/wallet/ui/models/send/TronStakingService.kt
```diff
@@ -0,0 +1,137 @@
...
+            (tronFrozenBalances.value as? TronFrozenBalanceState.Loaded)?.balances ?: return null
+        return balances.forResource(type)
+    }
+
+    private fun applyStakingMemo(type: TronResourceType) {
+        val op =
+            when (defiTypeProvider()) {
+                DeFiNavActions.FREEZE_TRX -> TronStakingOperation.FREEZE
+                DeFiNavActions.UNFREEZE_TRX -> TronStakingOperation.UNFREEZE
+                else -> error("applyStakingMemo called outside Tron staking")
```
rkokhatskyi, line 84: `applyStakingMemo` throws via `error(...)` here whenever `defiTypeProvider()` is anything other than `FREEZE_TRX` or `UNFREEZE_TRX` while `setResourceType` reaches it unconditionally and the public `SendFormViewModel.setTronResourceType` at line 699 forwards every UI invocation, so any caller outside a Tron staking context crashes the process whereas an early `if (!isStakingType()) return` at the top of `setResourceType` would silently return

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
aminsato: The fallback path with empty `Coins.Ethereum.ETH.address` is identical to `main` — see the original `loadCircleUSDCAccount` (same `?: Account(token = Coins.Ethereum.ETH, ...)` shape). In practice `loadAddresses(vaultId).firstOrNull()` returning null is a degenerate state (vault always has accounts), but I agree the silent empty-address propagation is a footgun. Out of scope for this refactor; will file as a separate fix to either error out early or skip the fallback when no ETH account exists.

app/src/main/java/com/vultisig/wallet/ui/models/send/TokenPreselectionService.kt
aminsato: Fixed in a571d0fb0. Capped `forcePreselection` to fire once on the first non-empty emission — the inner `accounts.collect` now flips a local `forced` flag to false after the first `onTokenSelected` call, so subsequent accounts hydrations no longer wipe the user's typed amount via `selectToken` → `resetUserInputCache`. New test `preSelect with force=true fires once and then defers to the user` covers the regression.

app/src/main/java/com/vultisig/wallet/ui/models/send/TokenPreselectionService.kt
aminsato: Fixed in a571d0fb0. The collector now skips the initial empty StateFlow value and waits for `AccountsLoader` to publish a real list before falling back to the default-coin map, so the static template (e.g. `Coins.Ethereum.USDC` with no MSCA address) never gets latched in. New test `preSelect skips the empty initial emission and only fires once accounts arrives` covers it.

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
aminsato: Fixed in a571d0fb0. `loadCircleUSDCAccount` now logs and publishes an empty list when no ETH account is present, instead of falling back to a zero-address `Coins.Ethereum.ETH` template that would propagate an empty address onto USDC and break `WithdrawUsdcCircleStrategy` at submit. New test `WITHDRAW_USDC_CIRCLE bails out with empty list when ETH account is missing` covers it.

app/src/main/java/com/vultisig/wallet/ui/models/send/TronStakingService.kt
aminsato: Fixed in a571d0fb0. Exposed `AmountFractionManager.cancel()` and call it from `SendFormViewModel.setTronResourceType` before delegating to `tronStakingService` — the existing `currentCoroutineContext().ensureActive()` checkpoint after the calc's last suspension point now throws on the cancelled job, so it can't resume and `setTextAndPlaceCursorAtEnd` over the cleared field with the previous resource type's amount.

app/src/main/java/com/vultisig/wallet/ui/models/send/TronStakingService.kt
aminsato: Leaving as-is. The `error(...)` is the same invariant assertion that lives on `main` in `applyTronStakingMemo`, and the public `setTronResourceType` is only wired to the FREEZE/UNFREEZE Tron resource tab UI which doesn't render outside Tron staking — the precondition holds at every real call site. Swapping it for a silent `if (!isStakingType()) return` would mask a programmer error rather than surface it. Happy to revisit if you spot a UI path that can actually invoke this outside staking context.

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    private suspend fun loadRewardsAccount(vaultId: VaultId) {
+        val accountsLoaded =
+            accountsRepository.loadAddresses(vaultId).firstOrNull()?.flatMap { it.accounts }
+        val thorchainAccount =
+            accountsLoaded?.find { it.token.id.equals(Coins.ThorChain.RUNE.id, true) } ?: return
+
+        val rujiAccount =
+            accountsLoaded.find { it.token.id.equals(Coins.ThorChain.RUJI.id, true) } ?: return
```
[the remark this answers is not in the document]
aminsato: Good catch — fixed in 2b78d3f71. Both early-return branches in `loadRewardsAccount` now publish `emptyList()` before returning, matching the `loadCircleUSDCAccount` shape so a prior flow's account rows can't leak through. Updated the existing test to assert the clear.

app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt
```diff
@@ -0,0 +1,239 @@
...
+                defiType != DeFiNavActions.STAKE_TCY &&
+                defiType != DeFiNavActions.UNSTAKE_TCY &&
+                defiType != DeFiNavActions.STAKE_STCY &&
+                defiType != DeFiNavActions.UNSTAKE_STCY &&
+                defiType != DeFiNavActions.REDEEM_YRUNE &&
+                defiType != DeFiNavActions.MINT_YTCY &&
+                defiType != DeFiNavActions.REDEEM_YTCY &&
+                defiType != DeFiNavActions.FREEZE_TRX &&
+                defiType != DeFiNavActions.UNFREEZE_TRX
+        ) {
```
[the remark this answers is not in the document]
aminsato: Fixed in 17a30ec18. Added `MINT_YRUNE` to the early-return allowlist alongside the other gas-subtracted ThorChain DeFi flows so `fetchPercentageOfAvailableBalance`'s gas-subtraction path actually runs on the recalc.

app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt
```diff
@@ -0,0 +1,239 @@
...
+                        selectedAmountFraction = AmountFraction.F100,
+                        isAmountSelectionLoading = true,
+                    )
+                }
+                val amount =
+                    try {
+                        calculatePercentageWithAccurateFee(1f)
+                    } finally {
+                        uiState.update { it.copy(isAmountSelectionLoading = false) }
+                    }
```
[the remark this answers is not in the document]
aminsato: Fixed in 17a30ec18. Both `finally` blocks now gate the `isAmountSelectionLoading = false` clear on `currentCoroutineContext().isActive`, so a cancelled (superseded) job's finally won't drop the spinner while a newer selection is still mid-flight. Used `isActive` rather than the suggested `chooseAmountFractionJob === currentCoroutineContext()[Job]` because under `UnconfinedTestDispatcher` the launch body can enter `finally` before the outer `chooseAmountFractionJob = scope.launch { ... }` assignment lands, which made the reference check race in tests; `isActive` is false only for cancelled jobs and true for normal completion (Completing state), which is exactly the gating we need.

app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt
```diff
@@ -0,0 +1,239 @@
...
+
+            val newGasFee = TokenValue(value = calculatedFee.amount, token = nativeCoin)
+
+            gasFee.value = adjustGasFee(newGasFee, gasSettings.value, specific.value)
+            amount = fetchPercentageOfAvailableBalance(percentage)
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.e(e, "Failed to calculate gas fee for percentage amount")
+        }
```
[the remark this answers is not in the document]
aminsato: Skipping. The `catch (e: Exception)` is a deliberate fallback design — "if accurate fee calc fails for any reason, return the pre-fee amount" — that lives unchanged on `main` and the fail-safe is what the rest of the flow relies on for non-EVM/DeFi percentage selection. Narrowing to specific exceptions without an audit of what `feeServiceComposite.calculateFees` and `tokenRepository.getNativeToken` can actually throw across all chains/providers risks letting unexpected types crash the percentage button. Worth a focused PR with that audit + targeted exception types; out of scope for this refactor.

app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt
```diff
@@ -0,0 +1,258 @@
...
+) {
+    private var chooseAmountFractionJob: Job? = null
+
+    /**
+     * Cancels any in-flight percentage/max calculation so callers (e.g. the Tron resource toggle)
+     * can clear amount fields without an older calc resuming and clobbering them after the
+     * suspension point in `calculatePercentageWithAccurateFee`.
+     */
+    fun cancel() {
+        chooseAmountFractionJob?.cancel()
```
rkokhatskyi, line 66: The `isActive` guards at lines 89 and 122 added in `17a30ec18` skip clearing `isAmountSelectionLoading` whenever the finally runs on a cancelled job, while `setTronResourceType` calls `amountFractionManager.cancel()` from outside the manager and `TronStakingService.setResourceType` only resets `selectedAmountFraction` rather than the spinner flag, so an external cancel during a pending percentage calc leaves `isAmountSelectionLoading = true` indefinitely until the user clicks another percentage button. The cleanest fix is to have `cancel()` itself update `uiState` to clear the flag

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -0,0 +1,166 @@
...
+    private suspend fun onLoadError(error: Throwable) {
+        Timber.e(error, "Failed to load accounts")
+    }
+
+    private suspend fun loadCircleUSDCAccount(vaultId: VaultId) {
+        // isRefresh = true skips the cached pre-emission so firstOrNull() returns the
+        // hydrated list — the ETH account we copy into accounts.value is then shown with
+        // refreshed balances rather than whatever fetchAccountFromDb wrote on a prior load.
+        val accountsLoaded =
+            accountsRepository.loadAddresses(vaultId, isRefresh = true).firstOrNull()?.flatMap {
```
rkokhatskyi, line 79: Passing `isRefresh = true` skips the cached emission that `AccountsRepositoryImpl.loadAddresses` normally sends first, so `firstOrNull()` blocks until every parallel balance lookup finishes or times out before any data reaches the form, which leaves WITHDRAW_USDC_CIRCLE and WITHDRAW_RUJI rendering an empty form on slow networks where the prior code surfaced the cached snapshot immediately. Collecting both emissions and showing the cached one while the hydrated one resolves would keep the UI responsive without sacrificing freshness

app/src/main/java/com/vultisig/wallet/ui/models/send/TokenPreselectionService.kt
```diff
@@ -0,0 +1,142 @@
...
+
+        preSelectTokenJob?.cancel()
+        preSelectTokenJob =
+            scope.launch {
+                var forced = forcePreselection
+                accounts.collect { accounts ->
+                    // Skip the initial empty StateFlow value so we don't lock in a static
+                    // default-coin template (which has no chain address bound) before
+                    // AccountsLoader publishes the real list.
+                    if (accounts.isEmpty()) return@collect
```
rkokhatskyi, line 45: `if (accounts.isEmpty()) return@collect` was added to skip the initial uninitialized StateFlow value, while it also discards the intentional empty publishes from `AccountsLoader.loadRewardsAccount` at lines 135, 142, 163 and `loadCircleUSDCAccount` at line 88 that this same review introduced, so when a vault is missing RUNE, RUJI, the rewards cache, or ETH the form renders with `state.selectedCoin == null` and no preselection ever fires whereas the prior code at least selected the static template coin. A sentinel emission such as a sealed `AccountsLoadState` would disambiguate uninitialized from intentionally empty

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -632,104 +684,23 @@ constructor(
...
-    }
-
-    private fun currentTronFrozenBalance(): BigDecimal? {
-        val type = uiState.value.tronResourceType ?: return null
-        val balances =
-            (tronFrozenBalances.value as? TronFrozenBalanceState.Loaded)?.balances ?: return null
-        return balances.forResource(type)
+        // Preempt any in-flight percentage calc so it can't resume after the field clears
+        // and overwrite them with the amount computed for the previous resource type.
+        amountFractionManager.cancel()
```
rkokhatskyi, line 702: `onAutoCompound` at line 929 swaps `accounts.value` and calls `selectToken` to switch tokens, while the same `chooseAmountFractionJob` race that this `setTronResourceType` now preempts via `amountFractionManager.cancel()` applies there too, so a percentage calc still suspended in `calculatePercentageWithAccurateFee` will resume after the autocompound token switch and write its old token's amount into `tokenAmountFieldState` for the new token

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt
aminsato: Good catch — applied in 74ac9d498. `cancel()` now clears `isAmountSelectionLoading` directly so external cancellers don't leave the spinner stuck.

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
aminsato: Valid UX concern. Not applied in this PR — fixing properly means collecting both cached + hydrated emissions and threading two-stage state through `loadCircleUSDCAccount` / `loadRewardsAccount`, which is a behavior change worth its own PR. Tracked in #4458.

app/src/main/java/com/vultisig/wallet/ui/models/send/TokenPreselectionService.kt
aminsato: Real regression — agreed. Not applied here because the cleanest fix (sealed `AccountsLoadState` to disambiguate uninitialized from intentionally empty) changes the type of `accounts` and ripples through several call sites. Tracked in #4458 alongside the related `isRefresh` issue.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
aminsato: Good catch. Fixed more broadly in 74ac9d498 — moved `amountFractionManager.cancel()` into `selectToken()` itself so every token switch is covered (autocompound, QR chain swap, manual picker), not just `setTronResourceType`.

johnnyluo:
## Review: PR #4432 — SendFormViewModel Decomposition

### Overall Impression
Excellent decomposition. The VM shrinks from ~1879 to ~1151 LOC. Each new service has a single, clear responsibility, and the test coverage is thorough. A few observations below.

### Strong Points
1. **Clean extraction boundaries** — `AmountFractionManager`, `TronStakingService`, `AccountsLoader`, `TokenPreselectionService`, and `GasFeeOrchestrator` each own a well-defined slice of behavior.
2. **CancellationException rethrow** — The `catch (e: Exception)` blocks across all files correctly rethrow `CancellationException` before swallowing generic exceptions. This was a latent bug in the original VM code that you fixed — good catch.
3. **`adjustGasFee` as a top-level function** — Making it a package-level function (`GasFeeOrchestrator.kt:366`) keeps it testable without mocking, and the test coverage verifies both UTXO and non-UTXO paths.
4. **`safeLaunch` usage in `AccountsLoader`** — Consistent with the project's error-handling pattern.
5. **Tests** — Each new service has corresponding tests covering loading states, cancellation, edge cases, and happy paths. The coroutine test setup (`TestCoroutineScheduler` + `UnconfinedTestDispatcher`) is consistent.

### Nitpicks / Suggestions

1. **`AmountFractionManager.kt:157-170`** — The large `defiType !=` check in `calculatePercentageWithAccurateFee` repeats the same structure as `fetchPercentageOfAvailableBalance` (lines 229-238). If a new DeFi type is added, both lists must be updated in sync. Consider extracting to a helper like `isDeFiWithGasRecalc()`.

2. **`GasFeeOrchestrator.kt:214`** — `planFee.value = 1` for non-UTXO chains looks like a magic number that propagates to `estimatedFee`. Consider a named constant.

3. **`TronStakingService.kt:100`** — `vaultProvider() ?: vaultRepository.get(vaultIdProvider())` — if both `vaultProvider()` returns null AND `vaultIdProvider()` returns null, the `return@safeLaunch` on the next line handles it, but the null safety is a bit fragile. Consider making the `VaultId` required rather than nullable.

4. **`AccountsLoader.kt:75-123`** — `loadCircleUSDCAccount` and `loadRewardsAccount` both call `accountsRepository.loadAddresses(vaultId, isRefresh = true).firstOrNull()` with the same pattern. Minor duplication — could extract a shared `refreshAddresses(vaultId)` helper.

5. **`GasFeeOrchestrator.kt`** — The `planBTC` field is written in `collectPlanFee` but I didn't see consumption in these changes — is it used downstream or is this a dead write?

### Tests
- **AmountFractionManagerTest** — Great coverage of UNFREEZE_TRX loading guard, vault/account null guards, EVM short-circuit, cancellation idempotency, and "latest wins" for rapid double-tap.
- **GasFeeOrchestratorTest** — Covers `adjustGasFee` (3 cases), refresh flag toggling, `collectGasTokenBalance` (native vs non-native), `collectMaxAmount` (UTXO vs non-UTXO vs null account).
- **TokenPreselectionServiceTest** — Covers null defi (tokenId wins over chainId, chainId picks native, fallback to first), DeFi defi types (RUJI default coin vs account match, TRX default), forcePreselection true/false, empty initial emission skip, and double-invocation cancellation.
- **TronStakingServiceTest** — Covers `isStakingType`, `currentFrozenBalance` (loading/error/loaded), `setResourceType` (loading guard, no-op on same type, field clearing, memo, UNFREEZE override), and `initIfStakingType` (no-op for non-staking, FREEZE memo, UNFREEZE load/error/no-trx-coin).

### Conclusion
**Approved** — Well-structured decomposition with thorough tests. No correctness issues found. The suggestions above are minor and non-blocking.

johnnyluo (approved):
LGTM

PR #4456, opened 2026-05-12
fix(ton): treat Duplicate msg_seqno as successful broadcast

### LINKED ISSUE #4314: TON: "Duplicate msg_seqno" lite-server error reaches UI verbatim; Try again is a no-op
## Summary

After a successful keysign, broadcasting a TON transaction can fail with `LITE_SERVER_UNKNOWN: cannot apply external message to current state : Duplicate msg_seqno N` if the wallet contract's on-chain `seqno` has advanced past the value captured in `BlockChainSpecific.Ton.sequenceNumber` between the time the transaction was prepared and the time the BoC was broadcast (e.g. another device/session committed a tx, the user re-tried after a successful broadcast, or the cached chain-specific data was stale).

Two UX problems compound here:

1. **Raw lite-server error reaches the user verbatim.** The full string `LITE_SERVER_UNKNOWN: cannot apply external message to current state : Duplicate msg_seqno 28` is shown on the Keysign error screen. It's unintelligible to non-technical users.
2. **"Try again" doesn't help.** The button retries `broadcastTransaction` against the same already-built BoC, which still embeds the stale seqno. Every retry deterministically fails with the same error. To recover, the user has to back out of the entire flow and re-enter Send / re-scan the TonConnect QR so a fresh `BlockChainSpecific.Ton` is fetched.

## Reproduction

1. Connect the Android app to a vault with a TON address that has at least one prior transaction.
2. Initiate a TON send (or TonConnect `sendTransaction` from the windows extension).
3. Before broadcasting, cause the on-chain seqno to advance (e.g. broadcast a separate tx from another device on the same wallet, or wait for a previous still-pending tx to land).
4. Complete keysign and broadcast.

**Expected**

A user-friendly error like *"Wallet state changed since this transaction was prepared. Please go back and try again."* and a Try-again path that re-fetches blockchain-specific data so the next attempt succeeds.

**Actual**

Raw API error reaches the UI; Try again is a no-op (re-broadcasts the same BoC with the same seqno); user must back out manually.

error-screen

## Logs

```
REQUEST: https://api.vultisig.com/ton/v2/sendBocReturnHash
BODY: {"boc":"te6cckEBBAEArwABRYgBTAgT3...AwAAN5tgnQ=="}

RESPONSE: 500
BODY: {"ok":false,"error":"LITE_SERVER_UNKNOWN: cannot apply external message to current state : Duplicate msg_seqno 28","code":500,"@extra":"..."}

E  KeysignViewModel: java.lang.Exception: Error broadcasting transaction: LITE_SERVER_UNKNOWN: cannot apply external message to current state : Duplicate msg_seqno 28
       at com.vultisig.wallet.data.api.chains.TonApiImpl.broadcastTransaction(TonApi.kt:68)
```

## Suggested fix

Two narrowly-scoped pieces:

1. **Map the broadcast error in `TonApiImpl.broadcastTransaction`.** When the 500 response body contains `Duplicate msg_seqno` (or, more generally, `cannot apply external message to current state`), throw a typed `TonStaleSeqnoException` whose `message` is a user-facing string. Other 500s keep their current behavior so we don't lose the raw text for genuinely unknown failures.

2. **Make "Try again" recoverable.** When the keysign error is `TonStaleSeqnoException`, the retry button should pop back to the verify screen (or trigger a `BlockChainSpecific` refresh + re-sign) so the next attempt uses a current seqno. Pure broadcast retry will always fail.

## Out of scope

- The general "joiner stuck after relay 500" problem is tracked in #4311 and is unrelated to this — that concerns the relay session start, this concerns the lite-server broadcast path.

## Context

- Discovered while testing #4283 (TonConnect multi-message co-signing). The signing path itself works; this surfaces only at broadcast.
- The same class of stale-state error exists for every chain that pre-fetches a nonce/sequence number, but TON's lite-server is unusually strict and the message is unusually opaque.

johnnyluo:
## Summary
- Extends the existing duplicate-broadcast guard in `TonApiImpl.broadcastTransaction` to also match `"Duplicate msg_seqno"` lite-server errors
- When two devices sign the same TON transaction simultaneously, the second broadcast fails with this error because the first device already consumed the seqno — the transaction is on-chain and the result should be treated as a success
- Returning `null` lets the caller fall back to the pre-computed hash via `orKnownHash`, and `recoverIfAlreadyBroadcast` verifies on-chain presence before returning it

## Issue
Closes #4314

## Test Plan
- [ ] Two-device TON keysign: both devices complete without error when the second broadcast receives `Duplicate msg_seqno`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction broadcast reliability by treating additional duplicate indications (including a new sequence-number duplicate marker) as non-fatal, allowing the system to fall back to previously confirmed transaction hashes rather than failing outright.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4456)

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/chains/ton/TonApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/chains/ton/TonApi.kt
@@ -56,10 +56,14 @@ internal class TonApiImpl @Inject constructor(private val http: HttpClient) : To
                 }
                 .bodyOrThrow<TonBroadcastTransactionResponseJson>()
         if (response.error != null) {
-            // Returning null for duplicate-message lets the caller fall back to the
-            // already-known hash via orKnownHash; treating it as an error would surface
-            // a spurious failure on retried broadcasts.
-            if (response.error.lowercase().contains(DUPLICATE_MESSAGE_MARKER)) {
+            // Returning null for these cases lets the caller fall back to the already-known hash
+            // via orKnownHash. Both errors indicate the tx (or an equivalent one with the same
+            // seqno) was already accepted — typically two devices broadcasting simultaneously.
+            val lowerError = response.error.lowercase()
+            if (
+                lowerError.contains(DUPLICATE_MESSAGE_MARKER) ||
+                    lowerError.contains(DUPLICATE_SEQNO_MARKER)
+            ) {
                 return null
             }
             error("Error broadcasting transaction: ${response.error}")
@@ -115,5 +119,6 @@ internal class TonApiImpl @Inject constructor(private val http: HttpClient) : To
     private companion object {
         const val BASE_URL = "https://api.vultisig.com/ton"
         const val DUPLICATE_MESSAGE_MARKER = "duplicate message"
+        const val DUPLICATE_SEQNO_MARKER = "duplicate msg_seqno"
     }
 }


aminsato:
@coderabbitai full review

aminsato:
lGTM

rkokhatskyi (commented):
would be good to cover this by tests with exact responses from the RPC node

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/ton/TonApi.kt
```diff
@@ -56,10 +56,14 @@ internal class TonApiImpl @Inject constructor(private val http: HttpClient) : To
...
-            if (response.error.lowercase().contains(DUPLICATE_MESSAGE_MARKER)) {
+            // Returning null for these cases lets the caller fall back to the already-known hash
+            // via orKnownHash. Both errors indicate the tx (or an equivalent one with the same
+            // seqno) was already accepted — typically two devices broadcasting simultaneously.
+            val lowerError = response.error.lowercase()
+            if (
+                lowerError.contains(DUPLICATE_MESSAGE_MARKER) ||
+                    lowerError.contains(DUPLICATE_SEQNO_MARKER)
+            ) {
                 return null
```
rkokhatskyi, line 67: returning `null` here means `orKnownHash` substitutes `tx.transactionHash` and `recoverIfAlreadyBroadcast` passes that through without ever calling `verify` since `verify` only fires on the exception path, so a `Duplicate msg_seqno` produced by a stale `BlockChainSpecific.Ton.sequenceNumber` or by a competing boc with different contents from another device will surface a hash the network never accepted as if the broadcast succeeded

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/ton/TonApi.kt
```diff
@@ -56,10 +56,14 @@ internal class TonApiImpl @Inject constructor(private val http: HttpClient) : To
...
-            // already-known hash via orKnownHash; treating it as an error would surface
-            // a spurious failure on retried broadcasts.
-            if (response.error.lowercase().contains(DUPLICATE_MESSAGE_MARKER)) {
+            // Returning null for these cases lets the caller fall back to the already-known hash
+            // via orKnownHash. Both errors indicate the tx (or an equivalent one with the same
+            // seqno) was already accepted — typically two devices broadcasting simultaneously.
+            val lowerError = response.error.lowercase()
+            if (
+                lowerError.contains(DUPLICATE_MESSAGE_MARKER) ||
+                    lowerError.contains(DUPLICATE_SEQNO_MARKER)
```
rkokhatskyi, line 65: the `#4314` repro logs show this response comes back as HTTP 500 and `iOS TonService.swift` only handles the duplicate marker in its 500 branch too, so after `#4318` made `bodyOrThrow` throw on anything other than 2xx the duplicate seqno body now flows through `recoverIfAlreadyBroadcast`'s catch which already verifies the pre computed hash, whereas this 2xx branch likely never executes for the scenario the fix targets

rkokhatskyi (changes requested):

johnnyluo:
Closing after reviewer feedback — the added code is dead. `Duplicate msg_seqno` comes back as HTTP 500, so `bodyOrThrow` throws before the `response.error` check is ever reached. The existing `recoverIfAlreadyBroadcast` catch path already handles this correctly: it catches the `NetworkException`, calls `verify(hash)` → `tonApi.getTsStatus(hash).transactions.isNotEmpty()`, and only returns the pre-computed hash if the tx is confirmed on-chain. No code change needed.

PR #4457, opened 2026-05-12
fix: THORName lookup crashes when aliases is null in API response

### LINKED ISSUE #4172: Referral: THORName lookup crashes when 'aliases' is null in API response
# [Fix] THORName DTO crashes on null aliases field [Referral]

## Problem
On the Referral screen, looking up a valid THORName fails with a `JsonConvertException` when the API returns `"aliases": null`. The HTTP response is 200 OK, but Ktor's deserializer rejects the payload because the `aliases` field in the THORName DTO is typed as a non-nullable list. As a result, valid registered THORNames (e.g. `maya`) cannot be resolved and the user sees a generic failure despite a successful server response.

## Expected Behavior
THORName lookups succeed when `aliases` is `null` in the API response. The DTO should tolerate a null `aliases` field and treat it as an empty list or a nullable value.

## Steps to Reproduce
1. Open the Referral screen
2. Enter an existing THORName that has `aliases: null`, e.g. `maya`
3. Observe the lookup error (despite a 200 response from THORNode)

### Must Do
- [ ] Change the `aliases` field in the THORName DTO to `List<Alias>? = null` (or `List<Alias> = emptyList()` with `coerceInputValues = true` on the Json configuration)
- [ ] Add a unit test covering the `aliases: null` deserialization case

### Must NOT Do
- Don't change unrelated DTO fields or serialization configuration beyond what is needed for this fix
- Don't alter the Json configuration globally if the change would affect other DTOs unexpectedly

## Acceptance Criteria
- [ ] `./gradlew` succeeds
- [ ] Looking up THORName `maya` (or any name with `aliases: null`) on the Referral screen resolves successfully
- [ ] No `JsonConvertException` is thrown when deserializing a THORName response with `"aliases": null`
- [ ] Unit test for the `aliases: null` case passes

## Context
**Actual error observed:**
```
RESPONSE: 200
BODY:
{
  "name": "maya",
  "expire_block_height": 36942828,
  "owner": "thor19d9fem39wayv4ydpp7kgufda940acvepjqmqtl",
  "preferred_asset": ".",
  "preferred_asset_swap_threshold_rune": "0",
  "affiliate_collector_rune": "0",
  "aliases": null
}

RESPONSE failed with exception:
io.ktor.serialization.JsonConvertException: Illegal input: Unexpected JSON token at offset 233:
  Expected start of the array '[', but had 'n' instead at path: $.aliases
```

Repository: vultisig/vultisig-android

johnnyluo:
## Summary
- Make `ThorOwnerData.aliases` nullable (`List<Aliases>?`) so kotlinx.serialization accepts `"aliases": null` from the API
- Guard the single call-site in `existsReferralCode` with a safe call (`?.any { … } ?: false`)
- Add two unit tests covering the null-aliases deserialization case and the happy path

## Issue
Closes #4172

## Root Cause
kotlinx.serialization only uses a field's default value when the key is **absent** from JSON. When THORNode returns `"aliases": null` explicitly, deserializing into a non-nullable `List<Aliases>` throws `JsonConvertException` even though the HTTP response is 200 OK.

## Test Plan
- [x] New unit tests pass (`./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.ThorChainApiImplTest"`)
- [ ] Manual: look up THORName `maya` on the Referral screen — lookup should succeed without error

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved referral-code validation to handle missing or absent alias data from the blockchain provider, preventing errors and improving app stability.
* **Tests**
  * Added tests covering missing alias responses and valid alias detection to ensure consistent behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4457)

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
@@ -451,9 +451,9 @@ constructor(
 
         val thorName = response.bodyOrThrow<ThorOwnerData>()
 
-        return thorName.aliases.any { alias ->
+        return thorName.aliases?.any { alias ->
             alias.chain.equals(THOR_CHAIN_NAME, ignoreCase = true) && alias.address.isNotBlank()
-        }
+        } ?: false
     }
 
     override suspend fun getReferralCodeInfo(code: String): ThorOwnerData {
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/ThorOwnerData.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/ThorOwnerData.kt
@@ -12,7 +12,7 @@ data class ThorOwnerData(
     @SerialName("preferred_asset") val preferredAsset: String,
     @SerialName("preferred_asset_swap_threshold_rune") val preferredAssetSwapThresholdRune: String,
     @SerialName("affiliate_collector_rune") val affiliateCollectorRune: String,
-    @SerialName("aliases") val aliases: List<Aliases> = emptyList(),
+    @SerialName("aliases") val aliases: List<Aliases>? = null,
 ) {
     /** A chain-specific alias address registered under a THORName. */
     @Serializable data class Aliases(val chain: String, val address: String)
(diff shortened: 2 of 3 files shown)


aminsato:
@coderabbitai full review

aminsato (approved):
LGTM

PR #4459, opened 2026-05-12
fix: keysign shows pending (not failed) when tx confirmation is slow

### LINKED ISSUE #4019: Keysign screen shows false Transaction Failed when confirmation is slow
## What happens

When the chain indexer takes longer than the configured timeout to pick up a transaction the foreground poller emits `NotFound`. The keysign screen maps this to `TransactionStatus.Failed` and shows the error waves animation with "Transaction failed".

The transaction isn't actually failed. The background poller (`RefreshPendingTransactionsUseCase`) correctly treats `NotFound` as transient and keeps polling. So the history screen may later show the same transaction as Confirmed which contradicts what the user was told on the keysign screen.

## Expected behavior

The foreground path should either keep showing a pending state with a "taking longer than expected" message or hand off to background polling gracefully instead of showing a failure.

## Where to look

`app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt`
- `startForegroundPolling`
- `toTransactionStatus`

johnnyluo:
## Summary
- Map `TransactionResult.NotFound` → `TransactionStatus.Pending` instead of `TransactionStatus.Failed`
- Keep the foreground poller running on `NotFound` instead of stopping it — polling only terminates on a definitive `Confirmed` or `Failed` result

## Issue
Closes #4019

## Root Cause
`NotFound` from the chain indexer means the transaction hasn't been picked up yet — it's transient, not a failure. The old code stopped polling and showed the error-waves animation on the keysign screen. Meanwhile, the background poller (`RefreshPendingTransactionsUseCase`) kept going and could later mark the same tx as `Confirmed`, directly contradicting what the user was told.

## Changes
Two lines changed in `KeysignViewModel.kt`:
1. `toTransactionStatus()`: `NotFound → Pending` (was `Failed("Confirmation taking longer than expected")`)
2. `startForegroundPolling()`: `NotFound` branch now does nothing / keeps polling (was `stopPolling()` + `cancel()`)

## Test Plan
- [ ] Manual: broadcast a tx on a slow chain (or simulate a delayed indexer response) — keysign screen should stay in the pending/spinner state until a definitive result arrives, not flip to error

## Summary by CodeRabbit

* **Bug Fixes**
  * Continued polling for transactions when confirmations are delayed instead of reporting them as failed.
  * Treats indexer “not found” cases as pending so delayed confirmations remain visible.

* **Changes**
  * Added a distinct timeout state with clearer foreground notification behavior for long-running checks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4459)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -757,10 +757,7 @@ constructor(
                             pollingTxStatusJob?.cancel()
                         }
 
-                        TransactionResult.NotFound -> {
-                            transactionStatusServiceManager.stopPolling()
-                            pollingTxStatusJob?.cancel()
-                        }
+                        TransactionResult.NotFound -> Unit
```
rkokhatskyi: The matching `NotFound` branch in `TransactionStatusService.startPolling` is still bundled with `Confirmed` and `Failed` so it calls `stopPolling()` and `stopForegroundAndService(detachNotification = true)` on the very first transient `NotFound` from a provider like `CardanoStatusProvider`, `SuiStatusProvider`, `SolanaStatusProvider` or `UtxoStatusProvider`, which tears down the foreground service whose `statusFlow` this `collect` observes so no further emissions arrive and the Pending state you just preserved stays in place forever rather than recovering to `Confirmed`

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -859,8 +856,7 @@ constructor(
         when (this) {
             TransactionResult.Confirmed -> TransactionStatus.Confirmed
             is TransactionResult.Failed -> TransactionStatus.Failed(this.reason.asUiText())
-            TransactionResult.NotFound ->
-                TransactionStatus.Failed("Confirmation taking longer than expected".asUiText())
+            TransactionResult.NotFound -> TransactionStatus.Pending
```
rkokhatskyi: `PollingTxStatusUseCase` also emits `NotFound` as its terminal timeout signal after `config.maxWaitSeconds` (5 minutes on EVMs, 120 minutes on Bitcoin) so once the use case gives up the user is left on a spinner that suggests the app is still checking, whereas iOS surfaces an explicit `timeout` error in `TransactionStatusPoller`, so the use case probably needs a distinct terminal variant rather than reusing `NotFound` for both transient lookups and the final moment polling stops

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
johnnyluo: Fixed by introducing `TransactionResult.TimedOut` as a distinct terminal variant. `PollingTxStatusUseCase` now emits `TimedOut` on timeout instead of `NotFound`. `TransactionStatusService` only stops on `Confirmed`/`Failed`/`TimedOut` — `NotFound` is treated like `Pending` (notification stays ongoing, polling continues). `KeysignViewModel` adds `TimedOut` to the terminal stop branch.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
johnnyluo: Addressed together with the service comment above. `PollingTxStatusUseCase` now emits `TransactionResult.TimedOut` (not `NotFound`) when `maxWaitSeconds` is exceeded. `KeysignViewModel.toTransactionStatus()` maps `TimedOut` → `Failed("Confirmation taking longer than expected")` so the user gets a proper error instead of a spinner. `NotFound` exclusively means transient miss and maps to `Pending`.

rkokhatskyi (approved):
LGTM

PR #4464, opened 2026-05-13
fix: Cosmos 404 treated as Pending instead of NotFound

### LINKED ISSUE #4026: Cosmos status provider treats HTTP 404 as transient instead of NotFound
## What happens

Cosmos REST nodes return HTTP 404 for transaction hashes that don't exist. The API layer uses `bodyOrThrow` which throws `NetworkException(404)` for any non-2xx response. The provider catches it as a generic `Exception` and returns `Pending`.

This means a transaction that was never actually broadcast (or was dropped from the mempool) keeps polling as `Pending` forever until the user closes the screen. The poller never sees `NotFound` so it never gives up.

Other chains handle this correctly because they encode "not found" inside HTTP 200 JSON responses. Cosmos is the only chain where "not found" comes as an HTTP status code.

## Fix

Catch the 404 in `CosmosApi.getTxStatus` and return `null` instead of throwing so the provider can map it to `NotFound`.

## Where to look

- `data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/CosmosStatusProvider.kt`
- `data/src/main/kotlin/com/vultisig/wallet/data/api/CosmosApi.kt`

johnnyluo:
## Summary
- `CosmosApi.getTxStatus` now returns `null` for HTTP 404 responses instead of letting `bodyOrThrow` throw a `NetworkException`
- `CosmosStatusProvider` already had the `null → NotFound` path — the provider itself needed no changes
- Adds `CosmosStatusProviderTest` with 6 cases covering the full status-check matrix (NotFound, Confirmed, Failed, Pending variants, network error)

## Issue
Closes #4026

## Root Cause
Cosmos REST nodes return HTTP 404 when a transaction hash doesn't exist. `bodyOrThrow` threw `NetworkException(404)` which the provider's broad `catch (e: Exception)` swallowed and mapped to `Pending`. This caused dropped/unbroadcast transactions to poll as `Pending` forever.

Other chains encode "not found" in an HTTP 200 JSON body, so they were unaffected.

## Test Plan
- [x] `./gradlew :data:testReleaseUnitTest --tests "com.vultisig.wallet.data.api.txstatus.CosmosStatusProviderTest"` — all 6 tests pass
- [ ] Manual: broadcast a tx on a Cosmos chain, observe it transitions through Pending → Confirmed
- [ ] Manual: enter a made-up tx hash on a Cosmos chain, confirm it shows NotFound (not stuck on Pending)

## Summary by CodeRabbit

* **Bug Fixes**
  * Transaction status queries now correctly treat missing transactions as not found (no parsing errors when a status is absent).

* **Tests**
  * Added comprehensive tests covering transaction status outcomes: confirmed, failed, pending, not found, and error conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4464)

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/CosmosApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/CosmosApi.kt
@@ -163,6 +163,7 @@ internal class CosmosApiImp(
 
     override suspend fun getTxStatus(txHash: String): CosmosTxStatusJson? {
         val response = httpClient.get("$rpcEndpoint/cosmos/tx/v1beta1/txs/$txHash")
+        if (response.status.value == 404) return null
         return response.bodyOrThrow<CosmosTxStatusJson>()
     }
 }
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/CosmosApiTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/CosmosApiTest.kt
@@ -1,0 +1,60 @@
+package com.vultisig.wallet.data.api
+
+import com.vultisig.wallet.data.utils.CosmosThorChainResponseSerializerImpl
+import io.ktor.client.HttpClient
+import io.ktor.client.engine.mock.MockEngine
+import io.ktor.client.engine.mock.respond
+import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
+import io.ktor.http.ContentType
+import io.ktor.http.HttpHeaders
+import io.ktor.http.HttpStatusCode
+import io.ktor.http.headersOf
+import io.ktor.serialization.kotlinx.json.json
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+import kotlinx.coroutines.runBlocking
+import kotlinx.serialization.json.Json
+import org.junit.jupiter.api.Test
+
+class CosmosApiTest {
+
+    private val jsonHeaders =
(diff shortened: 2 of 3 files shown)


johnnyluo:
Honestly , I don't think we need it , because the result is the same , but since I raised it , so just get it merged then

data/src/main/kotlin/com/vultisig/wallet/data/api/CosmosApi.kt
```diff
@@ -163,6 +163,7 @@ internal class CosmosApiImp(
 
     override suspend fun getTxStatus(txHash: String): CosmosTxStatusJson? {
         val response = httpClient.get("$rpcEndpoint/cosmos/tx/v1beta1/txs/$txHash")
+        if (response.status.value == 404) return null
```
rkokhatskyi, line 166: The new `if (response.status.value == 404) return null` line in `CosmosApi.kt` has no test coverage since `CosmosStatusProviderTest` mocks `cosmosApi.getTxStatus` to return `null` directly so the case labeled `404 response returns NotFound` would still pass if the API layer never inspected the response status, whereas a `MockEngine` based `CosmosApiTest` mirroring `BittensorApiTest` at `data/src/test/kotlin/com/vultisig/wallet/data/api/BittensorApiTest.kt` would actually exercise the new line

rkokhatskyi (changes requested):

rkokhatskyi (approved):

rkokhatskyi (approved):
LGTM

PR #4465, opened 2026-05-13
fix: raise Osmosis fee to 25,000 uosmo to prevent tx rejection

### LINKED ISSUE #3762: Osmosis transactions fail due to insufficient fee
## Problem

Osmosis uses EIP-1559 dynamic fees. Current base fee is `0.03 uosmo/gas`. For 300k gas limit the network requires at minimum `9,000 uosmo`.

Android hardcodes `7,500 uosmo` in `CosmosFeeService.kt` which is below the minimum. Transactions get rejected.

## File

`data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt`

## On-chain evidence

Real Osmosis transactions pay 10,709–92,755 uosmo. iOS uses 25,000 uosmo and works fine.

## Fix

Either increase to `25,000` (static) or fetch dynamically from `/osmosis/txfees/v1beta1/cur_eip_base_fee`.

johnnyluo:
## Summary
- Osmosis uses EIP-1559 dynamic fees; the old hardcoded value of `7,500 uosmo` was below the network minimum (`300,000 gas × 0.03 uosmo/gas = 9,000 uosmo`), causing transactions to be rejected
- Raised Osmosis fee amount to `25,000 uosmo`, matching iOS which uses the same value and succeeds reliably
- Added an explicit unit test asserting the Osmosis fee amount

## Issue
Closes #3762

## Test Plan
- [ ] Osmosis send transaction completes without fee rejection error
- [ ] Unit test `Osmosis fee amount is 25000` passes
- [ ] All other Cosmos chains unaffected (GaiaChain, Kujira, Terra, Akash, Qbtc, Noble, TerraClassic, Dydx still use their original values)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved gas fee calculation for Osmosis by introducing a dedicated minimum fee (25,000 uosmo) and separate handling, yielding more accurate fee estimates and more reliable transaction processing on Osmosis.

* **Tests**
  * Added a unit test to verify the Osmosis fee calculation returns the expected minimum fee.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4465)

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
@@ -7,6 +7,11 @@ import com.vultisig.wallet.data.blockchain.model.GasFees
 import com.vultisig.wallet.data.models.Chain
 
 class CosmosFeeService : FeeService {
+
+    companion object {
+        internal const val OSMOSIS_MIN_FEE_UOSMO = 25_000L
+    }
+
     override suspend fun calculateFees(transaction: BlockchainTransaction): Fee {
         return calculateDefaultFees(transaction)
     }
@@ -23,9 +28,12 @@ class CosmosFeeService : FeeService {
             }
         return when (chain) {
             Chain.Osmosis -> {
-                // Osmosis uses EIP-1559 dynamic fees; 25,000 uosmo matches iOS and covers the
-                // 300k gas × 0.03 uosmo/gas minimum with headroom for base fee spikes.
-                GasFees(limit = gasLimit.toBigInteger(), amount = 25000.toBigInteger())
+                // Osmosis uses EIP-1559 dynamic fees; OSMOSIS_MIN_FEE_UOSMO matches iOS and covers
+                // the 300k gas × 0.03 uosmo/gas minimum with headroom for base fee spikes.
+                GasFees(
+                    limit = gasLimit.toBigInteger(),
+                    amount = OSMOSIS_MIN_FEE_UOSMO.toBigInteger(),
+                )
             }
             Chain.GaiaChain,
             Chain.Kujira,
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
@@ -22,9 +22,13 @@ class CosmosFeeService : FeeService {
                 else -> 200000L
             }
         return when (chain) {
+            Chain.Osmosis -> {
+                // Osmosis uses EIP-1559 dynamic fees; 25,000 uosmo matches iOS and covers the
+                // 300k gas × 0.03 uosmo/gas minimum with headroom for base fee spikes.
+                GasFees(limit = gasLimit.toBigInteger(), amount = 25000.toBigInteger())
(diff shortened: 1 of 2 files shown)


aminsato:
@johnnyluo quick question on the fee row UX with the new `0.025 OSMO` value — Android and iOS render the fiat side differently:

| Platform | Native | Fiat |
|---|---|---|
| iOS | `0.025 OSMO` | `$0.0015` |
| Android | `0.025 OSMO` | `$0.00` |

Android rounds fiat to 2 decimals, so sub-cent fees collapse to `$0.00`, which looks broken to the user. iOS keeps enough precision to show `$0.0015`.

What's your preference for aligning Android with iOS?

1. Sub-cent values get formatted with ~2 significant figures (`$0.0015`, `$0.000034`), and fall back to 2 decimals once the value is ≥ `$0.01`.
2. Fixed 4-decimal display for fee fiat, matching iOS exactly.

Or do you see a different approach? Happy to file the follow-up ticket once you weigh in so this PR stays scoped.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
```diff
@@ -22,9 +22,13 @@ class CosmosFeeService : FeeService {
                 else -> 200000L
             }
         return when (chain) {
+            Chain.Osmosis -> {
+                // Osmosis uses EIP-1559 dynamic fees; 25,000 uosmo matches iOS and covers the
+                // 300k gas × 0.03 uosmo/gas minimum with headroom for base fee spikes.
+                GasFees(limit = gasLimit.toBigInteger(), amount = 25000.toBigInteger())
```
aminsato: Please avoid the bare `25000` magic number here — move it into a named constant in a `companion object` (e.g. `private const val OSMOSIS_MIN_FEE_UOSMO = 25_000L`) and reference that from the `when` branch. Makes the intent self-documenting at the call site and gives the test a single source of truth to import.

aminsato (approved):
LGTM

Transaction hash 
https://www.mintscan.io/osmosis/tx/B8C9F37C51CC2B47D0862736A5148E5F58BBA73213243E2574067B1119DC521B

rkokhatskyi (commented):
@johnnyluo shall we add dynamic fees calculation as well?

johnnyluo:
> @johnnyluo shall we add dynamic fees calculation as well?

we are talking about pennies here , let's do the dynamic fee later

johnnyluo:
> @johnnyluo quick question on the fee row UX with the new `0.025 OSMO` value — Android and iOS render the fiat side differently:
> 
> Platform	Native	Fiat
> iOS	`0.025 OSMO`	`$0.0015`
> Android	`0.025 OSMO`	`$0.00`
> Android rounds fiat to 2 decimals, so sub-cent fees collapse to `$0.00`, which looks broken to the user. iOS keeps enough precision to show `$0.0015`.
> 
> What's your preference for aligning Android with iOS?
> 
> 1. Sub-cent values get formatted with ~2 significant figures (`$0.0015`, `$0.000034`), and fall back to 2 decimals once the value is ≥ `$0.01`.
> 2. Fixed 4-decimal display for fee fiat, matching iOS exactly.
> 
> Or do you see a different approach? Happy to file the follow-up ticket once you weigh in so this PR stays scoped.

yup , do as what iOS does

PR #4466, opened 2026-05-13
fix: preserve coroutine cancellation in strict coroutine catch blocks

johnnyluo:
## Summary
- add `CancellationException` rethrow guards to broad `catch (Exception|Throwable)` blocks that run in strict coroutine contexts (`suspend` functions and coroutine builders)
- rename ignored catch vars (e.g. `catch (_: Exception)`) where needed so cancellation can be rethrown without changing existing non-cancellation handling
- keep scope intentionally narrow to coroutine contexts only, avoiding synchronous catch blocks

## Verification
- `./gradlew assembleDebug`
- `./gradlew test`

## Release Notes

* **Bug Fixes**
  * Improved error handling stability in key generation processes
  * Enhanced reliability of push notification management
  * Refined error handling in deposit form validation with better logging

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4466)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
@@ -616,6 +616,16 @@ class DKLSKeygen(
                 delay(500)
                 Timber.d("reshare ECDSA key successfully")
             }
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.d("Failed to reshare key, error: ${e.localizedMessage}")
+            if (attempt < 3) {
+                Timber.d("keygen/reshare retry, attempt: $attempt")
+                reshareWithRetry(attempt + 1)
+            } else {
+                throw e
+            }
         }
     }
 
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
@@ -616,9 +616,8 @@ class DKLSKeygen(
                 delay(500)
                 Timber.d("reshare ECDSA key successfully")
             }
-        } catch (e: CancellationException) {
-            throw e
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
             Timber.d("Failed to reshare key, error: ${e.localizedMessage}")
             if (attempt < 3) {
                 Timber.d("keygen/reshare retry, attempt: $attempt")
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
@@ -616,15 +616,6 @@ class DKLSKeygen(
                 delay(500)
                 Timber.d("reshare ECDSA key successfully")
             }
-        } catch (e: Exception) {
-            if (e is kotlinx.coroutines.CancellationException) throw e
-            Timber.d("Failed to reshare key, error: ${e.localizedMessage}")
-            if (attempt < 3) {
-                Timber.d("keygen/reshare retry, attempt: $attempt")
-                reshareWithRetry(attempt + 1)
-            } else {
-                throw e
-            }
         }
     }
 
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
@@ -232,7 +232,6 @@ class DKLSKeygen(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
-                if (e is kotlinx.coroutines.CancellationException) throw e
                 Timber.e(e, "Failed to get messages")
                 delay(1000) // backoff delay
             }
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
@@ -232,6 +232,7 @@ class DKLSKeygen(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
                 Timber.e(e, "Failed to get messages")
                 delay(1000) // backoff delay
             }
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
@@ -251,7 +251,6 @@ class DKLSKeysign(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
-                if (e is kotlinx.coroutines.CancellationException) throw e
                 Timber.e(e, "Failed to get messages")
             }
 
@@ -414,7 +413,6 @@ class DKLSKeysign(
         } catch (e: CancellationException) {
             throw e
         } catch (e: Exception) {
-            if (e is kotlinx.coroutines.CancellationException) throw e
             println("Failed to sign message ($messageToSign), error: ${e.localizedMessage}")
             val maxRetries = if (heardFromEver.isEmpty()) 1 else 3
             if (attempt < maxRetries) {
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
@@ -251,6 +251,7 @@ class DKLSKeysign(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
                 Timber.e(e, "Failed to get messages")
             }
 
@@ -413,6 +414,7 @@ class DKLSKeysign(
         } catch (e: CancellationException) {
             throw e
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
             println("Failed to sign message ($messageToSign), error: ${e.localizedMessage}")
             val maxRetries = if (heardFromEver.isEmpty()) 1 else 3
             if (attempt < maxRetries) {
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
@@ -590,9 +590,8 @@ class SchnorrKeygen(
                 Timber.d("reshare EdDSA key successfully")
                 delay(500)
             }
-        } catch (e: CancellationException) {
-            throw e
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
             Timber.d("Failed to reshare key, error: ${e.localizedMessage}")
             if (attempt < 3) {
                 Timber.d("keygen/reshare retry, attempt: $attempt")
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
@@ -590,15 +590,6 @@ class SchnorrKeygen(
                 Timber.d("reshare EdDSA key successfully")
                 delay(500)
             }
-        } catch (e: Exception) {
-            if (e is kotlinx.coroutines.CancellationException) throw e
-            Timber.d("Failed to reshare key, error: ${e.localizedMessage}")
-            if (attempt < 3) {
-                Timber.d("keygen/reshare retry, attempt: $attempt")
-                schnorrReshareWithRetry(attempt + 1)
-            } else {
-                throw e
-            }
         }
     }
 
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
@@ -166,7 +166,6 @@ class SchnorrKeygen(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
-                if (e is kotlinx.coroutines.CancellationException) throw e
                 Timber.e(e, "Failed to get messages")
                 delay(1000)
             }
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
@@ -590,6 +590,16 @@ class SchnorrKeygen(
                 Timber.d("reshare EdDSA key successfully")
                 delay(500)
             }
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.d("Failed to reshare key, error: ${e.localizedMessage}")
+            if (attempt < 3) {
+                Timber.d("keygen/reshare retry, attempt: $attempt")
+                schnorrReshareWithRetry(attempt + 1)
+            } else {
+                throw e
+            }
         }
     }
 
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
@@ -166,6 +166,7 @@ class SchnorrKeygen(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
                 Timber.e(e, "Failed to get messages")
                 delay(1000)
             }
--- a/app/src/main/java/com/vultisig/wallet/data/services/PushNotificationManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/services/PushNotificationManager.kt
@@ -18,6 +18,7 @@ import com.vultisig.wallet.data.repositories.VaultRepository
 import java.security.MessageDigest
 import javax.inject.Inject
 import javax.inject.Singleton
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.tasks.await
 import timber.log.Timber
@@ -46,8 +47,9 @@ constructor(
         try {
             val token = FirebaseMessaging.getInstance().token.await()
             onNewToken(token)
+        } catch (e: CancellationException) {
+            throw e
         } catch (e: Exception) {
-            if (e is kotlinx.coroutines.CancellationException) throw e
             Timber.w(e, "Failed to fetch FCM token")
         }
     }
@@ -66,8 +68,9 @@ constructor(
                         token = token,
                     )
                 )
+            } catch (e: CancellationException) {
+                throw e
             } catch (e: Exception) {
-                if (e is kotlinx.coroutines.CancellationException) throw e
                 Timber.w(e, "Failed to re-register vault ${vault.id} with new FCM token")
             }
         }
@@ -128,8 +131,9 @@ constructor(
                 )
             }
             vaultNotificationSettingsDao.setEnabled(vault.id, enabled)
+        } catch (e: CancellationException) {
+            throw e
         } catch (e: Exception) {
-            if (e is kotlinx.coroutines.CancellationException) throw e
             throw PushNotificationError.ApiFailure(e)
         }
     }
@@ -173,8 +177,9 @@ constructor(
                 }
                 vaultNotificationSettingsDao.setEnabled(vault.id, enabled)
                 successCount++
+            } catch (e: CancellationException) {
+                throw e
             } catch (e: Exception) {
-                if (e is kotlinx.coroutines.CancellationException) throw e
                 Timber.w(e, "Failed to update notification opt-in for vault ${vault.id}")
                 failureCount++
             }
--- a/app/src/main/java/com/vultisig/wallet/data/services/PushNotificationManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/services/PushNotificationManager.kt
@@ -18,7 +18,6 @@ import com.vultisig.wallet.data.repositories.VaultRepository
 import java.security.MessageDigest
 import javax.inject.Inject
 import javax.inject.Singleton
-import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.tasks.await
 import timber.log.Timber
@@ -47,9 +46,8 @@ constructor(
         try {
             val token = FirebaseMessaging.getInstance().token.await()
             onNewToken(token)
-        } catch (e: CancellationException) {
-            throw e
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
             Timber.w(e, "Failed to fetch FCM token")
         }
     }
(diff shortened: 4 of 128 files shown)


app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
```diff
@@ -232,6 +232,7 @@ class DKLSKeygen(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: Dead code — the `catch (e: CancellationException) { throw e }` block three lines above already intercepts every `CancellationException` before control can reach this `Exception` handler. This `is`-check is unreachable and should be removed.

This same pattern applies to every other site in this PR where a preceding `catch (e: CancellationException) { throw e }` block exists: `DklsKeysign`, `SchnorrKeygen`, `SchnorrKeysign`, `MldsaKeygen`, `MldsaKeysign`, `VaultAccountsViewModel`, `AddressManager`, `AmountFractionManager`, `AmountManager`, `GasFeeOrchestrator`, `AccountValidator`, all the `*Strategy` classes, `KeysignViewModel`, `ScanQrScreen`, and others.

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
```diff
@@ -600,6 +601,7 @@ class DKLSKeygen(
                 Timber.d("chaincode: ${chainCodeBytes.toHexString()}")
             }
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: Genuine fix (no prior CE handler here). Two suggestions:
1. Add `import kotlinx.coroutines.CancellationException` and use the unqualified name — inline FQ paths are inconsistent with the rest of the file.
2. Prefer restructuring as a dedicated `catch (e: CancellationException) { throw e }` branch before this `Exception` block — idiomatic Kotlin coroutines pattern and statically guaranteed by the compiler.

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
```diff
@@ -251,6 +251,7 @@ class DKLSKeysign(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: Dead code — prior `catch (e: CancellationException) { throw e }` block already handles this. Same for line 416.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -221,6 +221,7 @@ constructor(
                 updateModel { it.copy(lpPositionsDialog = lpPositions) }
                 reloadLpTab()
             } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: This new guard is immediately followed by the existing `if (e is CancellationException) throw e` on the next line. If both resolve to `kotlinx.coroutines.CancellationException`, the second check is permanently unreachable. Consolidate to a single guard.

The same double-check pattern appears at multiple sites in this file and in `ThorchainDefiPositionsViewModel`, `SwapTokenSelector`, and `FastVaultVerificationViewModel`.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -235,6 +235,7 @@ constructor(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: Dead code — prior `catch (e: CancellationException) { throw e }` block already handles this. Same for the addition at line 255.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -476,6 +476,7 @@ constructor(
         } catch (e: CancellationException) {
             throw e
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: Dead code — prior `catch (e: CancellationException) { throw e }` block already handles this. Same for the additions at lines 534 and 615.

app/src/main/java/com/vultisig/wallet/data/services/PushNotificationManager.kt
```diff
@@ -47,6 +47,7 @@ constructor(
             val token = FirebaseMessaging.getInstance().token.await()
             onNewToken(token)
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo, line 50: Genuine fix (no prior CE handler here). Two suggestions:
1. Add `import kotlinx.coroutines.CancellationException` and use the unqualified name for consistency with the rest of the codebase.
2. Restructure as a separate `catch (e: CancellationException) { throw e }` branch before this `Exception` block — idiomatic and statically guaranteed.

app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
```diff
@@ -114,6 +114,7 @@ internal class AddressManager(
         } catch (e: CancellationException) {
             throw e
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: Dead code — prior `catch (e: CancellationException) { throw e }` block already handles this.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/AccountValidator.kt
```diff
@@ -72,6 +72,7 @@ internal class AccountValidator(
             } catch (e: CancellationException) {
                 throw e
             } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: Dead code — prior `catch (e: CancellationException) { throw e }` block already handles this. Same pattern applies to all the Strategy classes in this PR: `BondStrategy`, `DefaultSendStrategy`, `MintStrategy`, `RedeemStrategy`, `StakeStrategy`, `UnbondStrategy`, `UnstakeStrategy`, `WithdrawUsdcCircleStrategy`.

app/src/main/java/com/vultisig/wallet/ui/models/keygen/AddReferralViewModel.kt
```diff
@@ -101,7 +101,8 @@ constructor(
                 } catch (e: CancellationException) {
                     state.update { it.copy(isLoading = false) }
                     throw e
-                } catch (_: Exception) {
+                } catch (e: Exception) {
+                    if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: Dead code — the `catch (e: CancellationException) { throw e }` block four lines above already intercepts every `CancellationException` before this `Exception` handler is reached. This guard is unreachable and can be removed.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronDeFiBalanceService.kt
```diff
@@ -86,6 +86,7 @@ class TronDeFiBalanceService(
         } catch (e: kotlin.coroutines.cancellation.CancellationException) {
             throw e
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: Dead code — the prior `catch (e: kotlin.coroutines.cancellation.CancellationException) { throw e }` block already handles this. `kotlin.coroutines.cancellation.CancellationException` is the runtime class that `kotlinx.coroutines.CancellationException` aliases (since coroutines 1.4), so the two names refer to the same type. The `is`-check inside the `Exception` handler is unreachable.

data/src/main/kotlin/com/vultisig/wallet/data/MainDataModule.kt
```diff
@@ -193,6 +193,7 @@ private fun migrateFromEncryptedSharedPrefs(context: Context, newPrefs: SharedPr
                 EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM,
             )
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: `migrateFromEncryptedSharedPrefs` is a regular `fun`, not `suspend`. There are no coroutine cooperative-cancellation points inside it (all operations are synchronous `SharedPreferences`/`EncryptedSharedPreferences` calls), so `CancellationException` cannot arrive from a cancelled coroutine here. This guard is unnecessary and should be removed.

data/src/main/kotlin/com/vultisig/wallet/data/MainDataModule.kt
```diff
@@ -205,6 +206,7 @@ private fun migrateFromEncryptedSharedPrefs(context: Context, newPrefs: SharedPr
         try {
             legacyPrefs.all
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: Same as above — non-suspend function, no coroutine cancellation points. Remove.

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
```diff
@@ -600,6 +600,7 @@ class DKLSKeygen(
...
```
aminsato: This is inside the keygen/reshare retry loop. Before this change, a `CancellationException` was swallowed and the loop kept attempting `if (attempt < 3) { ... }` — meaning a cancelled coroutine could continue to consume CPU and TSS resources for up to 3 more iterations. After this change, CE propagates immediately. The new behavior is correct, but it does change recovery semantics in a TSS-critical path. The contributor guidelines flag TSS/crypto JNI bindings as needing explicit maintainer review — please confirm with maintainers and run the keygen integration tests (`FastVaultKeygenFlowTest`, reshare flow) before merging. Same applies to the identical change in `SchnorrKeygen.kt`.

app/src/main/java/com/vultisig/wallet/ui/models/keygen/AddReferralViewModel.kt
aminsato: This guard is unreachable: the preceding `catch (e: CancellationException) { state.update { it.copy(isLoading = false) }; throw e }` (lines 101-103) catches `CancellationException` first, so execution never reaches this `catch (e: Exception)` for CE. If it were reachable, it would also be wrong — rethrowing CE here would bypass the `isLoading = false` cleanup that the explicit CE catch performs. Same pattern (explicit CE catch above + `_` → `e` rename without using `e`) appears in `EnterVaultInfoViewModel.kt:504`, `FourByteRepository.kt:33`, and `DepositFormViewModel.kt:1168` — those also introduce unused-variable warnings.

data/src/main/kotlin/com/vultisig/wallet/data/tss/TssMessenger.kt
```diff
@@ -56,6 +56,7 @@ class TssMessenger(
                     // when it reach to this point , it means the message was sent successfully
                     break
                 } catch (e: Exception) {
```
aminsato, line 59: Behavior change inside the TSS send-message retry loop: previously, a `CancellationException` was swallowed and the next retry attempt would run on an already-cancelled scope; now CE breaks the loop immediately. Correct, but worth verifying that no caller in the keysign flow was implicitly depending on the swallow (e.g. completing a final message after the user navigated away). Same applies to `TssMessagePuller.kt`.

aminsato (commented):
Sweep is correct overall, but the changes that land in the TSS keygen/keysign retry paths warrant a closer look — they change cancellation semantics in code the project treats as restricted (the contributor guidelines explicitly call out TSS/crypto JNI bindings as needing maintainer review, and require keygen/keysign regression testing after any refactor touching them). Please confirm keygen/keysign integration tests pass and run a manual keygen+reshare end-to-end before merging.

app/src/main/java/com/vultisig/wallet/ui/models/keygen/AddReferralViewModel.kt
```diff
@@ -101,7 +101,7 @@ constructor(
                 } catch (e: CancellationException) {
                     state.update { it.copy(isLoading = false) }
                     throw e
-                } catch (_: Exception) {
+                } catch (e: Exception) {
```
aminsato: Leftover rename: the duplicate cancellation guard was removed (good), but the `_` → `e` rename was left behind, and the body of this `catch` never references `e`. This produces an unused-parameter warning and serves no purpose now that the rethrow is gone. Please revert to `catch (_: Exception)`. Same applies to `EnterVaultInfoViewModel.kt:504` and `DepositFormViewModel.kt:2821`.

aminsato (commented):
Second pass on the latest state. The duplicate guards I flagged earlier are addressed, but a couple of items remain worth resolving before merge.

data/src/main/kotlin/com/vultisig/wallet/data/tss/TssMessenger.kt
aminsato, line 59: Most of the new guards in this PR use the fully-qualified form `if (e is kotlinx.coroutines.CancellationException) throw e`. Please switch to the unqualified `CancellationException` and add `import kotlinx.coroutines.CancellationException` at the top of each file (the pre-existing `catch (e: CancellationException)` blocks already in several of these files do exactly that, e.g. `ThorChainApi.kt:176`). The fully-qualified inline reference is harder to read and inconsistent with the rest of the codebase. Applies across the whole sweep.

aminsato (commented):
Style follow-up on the cancellation guards.

app/src/main/java/com/vultisig/wallet/data/usecases/SaveVaultUseCase.kt
```diff
@@ -51,6 +51,7 @@ constructor(
                         try {
                             Chain.fromRaw(cpk.chain)
                         } catch (e: Exception) {
+                            if (e is kotlinx.coroutines.CancellationException) throw e
```
rkokhatskyi: `Chain.fromRaw` is a purely synchronous enum lookup with no suspension point inside this `try`, so a coroutine `CancellationException` cannot reach this branch which makes the guard dead code in the same way the team already flagged on `MainDataModule`

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefaultChainsRepository.kt
```diff
@@ -27,6 +27,7 @@ constructor(private val dataStore: AppDataStore, private val json: Json) : Defau
                 try {
                         json.decodeFromString<List<String>>(it).map { Chain.fromRaw(it) }
                     } catch (e: Throwable) {
+                        if (e is kotlinx.coroutines.CancellationException) throw e
```
rkokhatskyi: This catch wraps only synchronous code, namely `json.decodeFromString` plus a plain `Chain.fromRaw` mapping, so no `CancellationException` can surface here from a cancelled coroutine which makes the new guard unreachable

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -162,6 +162,7 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
             try {
                 EthereumFunction.balanceErc20Decoder(it)
             } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
```
rkokhatskyi: `EthereumFunction.balanceErc20Decoder(it)` runs after the suspending RPC has already returned and is a purely synchronous decoder, so cancellation cannot reach this fallback which makes the rethrow dead code

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
```diff
@@ -334,6 +334,7 @@ constructor(
             val length = BigInteger(bytes.sliceArray(32..63)).toInt()
             return String(bytes.sliceArray(64 until 64 + length)).lowercase()
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
rkokhatskyi: `decodeContractString` is a regular `fun` that only does byte slicing and `BigInteger` construction with no suspension, so cancellation has no path into this catch which makes the rethrow noise

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1000,6 +1001,8 @@ constructor(
                                         try {
                                             json.parseToJsonElement(cosmosMsg.value)
                                         } catch (e: Exception) {
+                                            if (e is kotlinx.coroutines.CancellationException)
```
rkokhatskyi: `json.parseToJsonElement(cosmosMsg.value)` is a synchronous serialization call with no suspension inside this `try`, so cancellation cannot reach this fallback and the guard only adds noise to what should remain a plain parse fallthrough

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/thorchain/TCYStakingService.kt
```diff
@@ -364,6 +368,7 @@ constructor(
                     try {
                         BigDecimal(staker.amount)
                     } catch (e: Exception) {
+                        if (e is kotlinx.coroutines.CancellationException) throw e
```
rkokhatskyi: This `try` wraps only the synchronous `BigDecimal(staker.amount)` constructor inside the `fold` lambda, so there is no suspension point that could produce a `CancellationException` here and the guard can be dropped

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
johnnyluo: I think it should be fine , because if the coroutine get concelled , the process should stop

data/src/main/kotlin/com/vultisig/wallet/data/tss/TssMessenger.kt
johnnyluo: if the coroutine exit , then the process should stop

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -482,6 +487,7 @@ constructor(
                 else -> addr.replace("-", ".")
             }
         } catch (t: Throwable) {
+            if (t is kotlinx.coroutines.CancellationException) throw t
```
rkokhatskyi: `mapThorPoolAsset` is a non suspend private function whose `try` body only runs regex matching and string substitution, so no `CancellationException` can reach this branch which leaves the guard dead in the same way as the prior round of synchronous catches the team already pruned

data/src/main/kotlin/com/vultisig/wallet/data/repositories/FourByteRepository.kt
```diff
@@ -39,7 +39,8 @@ constructor(private val fourByteApi: FourByteApi, @param:PrettyJson private val
         return try {
             val decodeGeneric = decodeGeneric(memo, functionSignature)
             convertDecodedCallToPrettyArray(decodeGeneric)
-        } catch (_: Exception) {
+        } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
rkokhatskyi: `decodeFunctionArgs` is declared as `fun`, not `suspend fun`, and the body only calls `decodeGeneric` plus `convertDecodedCallToPrettyArray` which are both synchronous, so cancellation cannot surface inside this catch which leaves the rethrow as noise

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/usecases/ThorchainBondUseCase.kt
```diff
@@ -180,6 +182,7 @@ constructor(
 
             return totalSeconds / totalBlocks
         } catch (t: Throwable) {
+            if (t is kotlinx.coroutines.CancellationException) throw t
```
rkokhatskyi: `averageBlockTimeFromChurns` is declared as `fun`, not `suspend fun`, and the body only sorts the list and runs arithmetic on `Int`/`Long` fields, so cancellation cannot surface inside this catch which leaves the guard as noise

app/src/main/java/com/vultisig/wallet/ui/models/keygen/KeygenViewModel.kt
```diff
@@ -346,6 +347,7 @@ constructor(
                 val eddsaUIResp = Tss.getLocalUIEddsa(eddsaShare)
                 localUiEddsa = eddsaUIResp.padEnd(64, '0')
             } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
```
rkokhatskyi: The `try` body only invokes `vault.getKeyshare(...)`, which is a synchronous list lookup, and `Tss.getLocalUIEcdsa`/`Tss.getLocalUIEddsa`, which are blocking JNI calls without a `suspend` modifier, so no suspension point exists for cancellation to reach this catch which makes the guard dead

rkokhatskyi (changes requested):

aminsato:
Across this PR the rethrow guards use the fully-qualified type:

```kotlin
```

Suggest importing `kotlinx.coroutines.CancellationException` and using the short form everywhere for consistency with the rest of the codebase:

```kotlin
import kotlinx.coroutines.CancellationException

// ...
```

Applies to all ~95 occurrences added in this PR (both `e is …` and `t is …` variants). A few files (`CardanoApi.kt`, `ThorChainApi.kt`, `ThorchainDefiPositionsViewModel.kt`) already import `kotlin.coroutines.cancellation.CancellationException`, so in those just reuse the existing import instead of adding a second one.

aminsato:
`data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt` (around line 167) — the guard was added inside a commented-out block:

```kotlin
/* override suspend fun broadcastTransaction(...) {
    ...
    } catch (e: Exception) {
        if (e is kotlinx.coroutines.CancellationException) throw e   // not needed — dead code
        error("Failed to broadcast transaction: ${e.message}")
    }
} */
```

The whole commented-out `broadcastTransaction` block is dead code and should be deleted rather than touched. Suggest removing lines 138–170 entirely.

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
```diff
@@ -616,6 +616,15 @@ class DKLSKeygen(
                 delay(500)
                 Timber.d("reshare ECDSA key successfully")
             }
+        } catch (e: Exception) {
```
johnnyluo: these are not relevant anymore , so let's not add it

app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
```diff
@@ -590,6 +590,15 @@ class SchnorrKeygen(
...
             }
+        } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
+            Timber.d("Failed to reshare key, error: ${e.localizedMessage}")
+            if (attempt < 3) {
+                Timber.d("keygen/reshare retry, attempt: $attempt")
+                schnorrReshareWithRetry(attempt + 1)
+            } else {
+                throw e
+            }
```
johnnyluo: these are not need anymore

rkokhatskyi (approved):
LGTM

Issue #4328, opened 2026-04-30
DklsKeysign: dedup cache not reset between retries causes silent handle starvation


aminsato:
## Problem

\`DKLSKeysign\` keeps an inbound-message dedup cache as a class field that lives for the entire lifetime of the instance. When \`keysignOneMessageWithRetry\` retries (up to 3 times after a timeout), it builds a brand-new DKLS handle but reuses the same cache. Plaintext-deterministic DKLS round messages from the previous attempt are then silently skipped on the new handle, so the new handle can never finish — guaranteeing another 60 s timeout.

## Walk-through

1. \`DKLSKeysign.cache\` is declared as an instance field at [\`DklsKeysign.kt:64\`](app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt#L64).
2. \`processInboundMessage\` populates it at [\`DklsKeysign.kt:267\`](app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt#L267) with key \`\"\$sessionID-\$localPartyID-\$messageID-\${msg.hash}\"\` and dedups against it at [line 250](app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt#L250).
3. The \`msg.hash\` field is set by the **sender** in \`TssMessenger.send\` as \`body.md5()\` ([\`TssMessenger.kt:48\`](data/src/main/kotlin/com/vultisig/wallet/data/tss/TssMessenger.kt#L48)) — i.e. md5 of the **plaintext** DKLS message bytes, not the ciphertext. (AES-GCM's random nonce only randomizes ciphertext; plaintext is unchanged.)
4. On retry, [\`keysignOneMessageWithRetry\`](app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt#L281) creates a fresh \`Handle()\` and calls \`dkls_sign_session_from_setup\` again with the same setup-message. The peer (e.g. iPad) likewise restarts its sign session from the same setup. DKLS round-1 outbound is deterministic given the same setup + keyshare → **same plaintext → same md5 → cache hit**.
5. The new handle skips the round-1 message it actually needs (\`continue\` at [line 252](app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt#L252)) and then waits forever for input → next 60 s timeout fires → another retry hits the same wall.

[...]

PR #4467, opened 2026-05-13
fix: reset TSS dedup caches on retries

johnnyluo:
## Summary
- reset per-session inbound dedup caches at the start of each retry attempt in DKLS and Schnorr keysign flows
- apply the same per-attempt cache reset to DKLS and Schnorr reshare retry loops to avoid stale message dedup across new handles
- keep within-attempt dedup intact while preventing retry starvation from deterministic round messages (issue #4328)

## Testing
- ./gradlew :app:testDebugUnitTest

## Summary by CodeRabbit

* **Bug Fixes**
  * Retry flows for key generation and signing now fully reset in-memory deduplication/cache between attempts to prevent previously seen inbound messages from blocking new retries.
  * Improved cancellation handling so cancellation requests propagate immediately rather than being treated as standard failures that trigger retries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4467)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
@@ -601,7 +601,6 @@ class DKLSKeygen(
                 Timber.d("chaincode: ${chainCodeBytes.toHexString()}")
             }
         } catch (e: Exception) {
-            if (e is kotlinx.coroutines.CancellationException) throw e
             Timber.d("Failed to reshare key, error: ${e.localizedMessage}")
             if (attempt < 3) {
                 Timber.d("keygen/reshare retry, attempt: $attempt")
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
@@ -519,6 +519,7 @@ class DKLSKeygen(
 
     suspend fun reshareWithRetry(attempt: Int) {
         try {
+            cache.clear()
             val keyshareHandle = Handle()
             if (vault.pubKeyECDSA.isNotEmpty()) {
                 val keyshare = getKeyshareBytesFromVault()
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeygen.kt
```diff
@@ -600,6 +601,7 @@ class DKLSKeygen(
                 Timber.d("chaincode: ${chainCodeBytes.toHexString()}")
             }
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: can you please revert the change on this line , it is not related to this PR

app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeygen.kt
```diff
@@ -570,6 +571,7 @@ class SchnorrKeygen(
                 delay(500) // slightly delay to give local party time to process outbound messages
             }
         } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: can you please revert the change on this line , it is not related to this PR

aminsato:
Walked through issue #4328's checklist against this PR — all 6 retry paths end up covered:

| Code path | Issue asked for fix | PR adds it | Pre-existing |
|---|---|---|---|
| `DklsKeysign.keysignOneMessageWithRetry` | primary bug | ✅ `DklsKeysign.kt:327` | — |
| `DklsKeygen.dklsKeygenWithRetry` | sweep | — | ✅ `DklsKeygen.kt:314` |
| `DklsKeygen.reshareWithRetry` | sweep | ✅ `DklsKeygen.kt:520` | — |
| `SchnorrKeygen.schnorrKeygenWithRetry` | sweep | — | ✅ `SchnorrKeygen.kt:292` |
| `SchnorrKeygen.schnorrReshareWithRetry` | sweep | ✅ `SchnorrKeygen.kt:489` | — |
| Schnorr keysign retry | sweep | ✅ `SchnorrKeysign.kt:327` | — |

So the sweep is complete — the initial keygen paths already had `cache.clear()` before this PR, and this fills in the reshare + keysign gaps with the same pattern.

Order is correct in the keysign files: `cache.clear()` sits after the `if (attempt == 0) { heardFromEver.clear(); ... }` reset block but before any inbound message is processed via the pull loop, so the cache is empty before `processInboundMessage` runs on attempt N. Within-attempt dedup (the part the issue says is still needed) is preserved.

The one missing piece is a regression test — happy to file a follow-up so it doesn't block this fix from landing.

rkokhatskyi (approved):
LGTM

PR #4487, opened 2026-05-14
fix(reshare): remove device-count cap to support any vault topology

johnnyluo:
## Summary

- Reshare peer discovery was capped at `minimumDevices - 1` (default: 1) other devices, which is correct for keygen but wrong for reshare
- Any reshare needing more than 2 total devices (e.g. 2/2 → 2/3, 2/3 → 2/2, 2/3 → 3/4) was silently broken: joining devices appeared in the discovered list but were never added to `selectedDevices` and were excluded from `keygenCommittee`
- For `TssAction.ReShare`, the cap is now bypassed in both `startParticipantDiscovery` (auto-selection) and `selectDevice` (manual toggle); minimum stays at 2 total devices, no maximum

## Root Cause

`startParticipantDiscovery()` computes `maxOtherDevices = minimumDevices - 1` and only auto-selects up to that many new devices. `selectDevice()` enforces the same ceiling. Because `ReshareStartViewModel` doesn't pass `deviceCount`, `minimumDevices` defaults to `MIN_KEYGEN_DEVICES = 2`, so only **1** other device was ever selectable — all subsequent joiners were silently dropped.

## Test Plan

- [ ] 2/2 vault reshare (replace one device) — paired device is included ✓
- [ ] 2/2 vault reshare to 2/3 (add a device) — both non-initiator devices are included ✓
- [ ] 2/3 vault reshare (all three join) — all three participants in `keygenCommittee` ✓
- [ ] 2/3 vault reshare to 2/2 (drop one device) — correct subset included ✓
- [ ] Keygen peer discovery (non-reshare) — device cap still enforced ✓
- [ ] `./gradlew lint` passes ✓
- [ ] Unit tests pass (`./gradlew :app:testDebugUnitTest`) ✓

## Release Notes

* **Bug Fixes**
  * Fixed device selection behavior during key resharing to properly support multi-device scenarios
  * Improved automatic device discovery during resharing to correctly include all newly discovered devices in the network

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4487)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -262,11 +262,13 @@ constructor(
 
     fun selectDevice(device: ParticipantName) {
         state.update {
+            val isReshare = args?.action == TssAction.ReShare
             val maxOtherDevices = it.minimumDevices - 1
             it.copy(
                 selectedDevices =
                     if (device in it.selectedDevices) it.selectedDevices - device
-                    else if (it.selectedDevices.size >= maxOtherDevices) it.selectedDevices
+                    else if (!isReshare && it.selectedDevices.size >= maxOtherDevices)
+                        it.selectedDevices
                     else it.selectedDevices + device
             )
         }
@@ -513,11 +515,16 @@ constructor(
                     val existingDevices = currentState.devices.toSet()
                     val newDevices = devices - existingDevices
 
-                    val maxOtherDevices =
-                        if (currentState.minimumDevices > 1) currentState.minimumDevices - 1
-                        else currentState.minimumDevices
-                    val remainingSlots = maxOtherDevices - currentState.selectedDevices.size
-                    val devicesToAutoSelect = newDevices.take(remainingSlots.coerceAtLeast(0))
+                    val devicesToAutoSelect =
+                        if (args?.action == TssAction.ReShare) {
+                            newDevices
+                        } else {
+                            val maxOtherDevices =
+                                if (currentState.minimumDevices > 1) currentState.minimumDevices - 1
+                                else currentState.minimumDevices
+                            val remainingSlots = maxOtherDevices - currentState.selectedDevices.size
+                            newDevices.take(remainingSlots.coerceAtLeast(0))
+                        }
                     val selectedDevices = currentState.selectedDevices.toSet() + devicesToAutoSelect
 
                     state.update {


app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -262,11 +262,13 @@ constructor(
 
     fun selectDevice(device: ParticipantName) {
         state.update {
+            val isReshare = args?.action == TssAction.ReShare
             val maxOtherDevices = it.minimumDevices - 1
```
aminsato, line 266: Nit: `maxOtherDevices` is computed unconditionally even though it's unused on the reshare branch. Minor readability cleanup:

```kotlin
val isReshare = args?.action == TssAction.ReShare
val capReached = !isReshare && it.selectedDevices.size >= it.minimumDevices - 1
it.copy(
    selectedDevices =
        if (device in it.selectedDevices) it.selectedDevices - device
        else if (capReached) it.selectedDevices
        else it.selectedDevices + device
)
```

aminsato:
Follow-up suggestion: consider extracting the cap rule into a small pure helper so it can be unit-tested directly.

The current logic is duplicated across `selectDevice` and `startParticipantDiscovery` — this PR had to touch both sites, and they can drift again. The original bug was silent precisely because the rule lives inside a ViewModel with no direct test surface.

```kotlin
internal object PeerSelection {
    fun canSelectMore(action: TssAction?, selectedCount: Int, minimumDevices: Int): Boolean =
        action == TssAction.ReShare || selectedCount < (minimumDevices - 1)

    fun autoSelect(
        action: TssAction?,
        newDevices: List<ParticipantName>,
        selectedCount: Int,
        minimumDevices: Int,
    ): List<ParticipantName> {
        if (action == TssAction.ReShare) return newDevices
        val cap = if (minimumDevices > 1) minimumDevices - 1 else minimumDevices
        return newDevices.take((cap - selectedCount).coerceAtLeast(0))
    }
}
```

Both callsites become one-liners, and a `PeerSelectionTest` covering each `TssAction` × `(minimumDevices, selectedCount)` combo (~15 lines) would catch this regression class going forward — including an explicit case for `reshare + minimumDevices=2 + 3 joiners`, which is the exact scenario that silently broke.

Not blocking for this PR — happy to do it as a follow-up.

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
johnnyluo: it is ok

PR #4493, opened 2026-05-14
fix(keysign): guard joinKeysign against duplicate startSession calls

### LINKED ISSUE #4475: Joiner shows 'Relay server unavailable' after successful join — duplicate startSession POSTs return 500
## Problem

On the joined device, the relay POST that registers the device with a keysign session **succeeds with 201**, but immediately after, additional identical POSTs are sent to the same endpoint and the relay returns **500** for every one. After the retry budget is exhausted, the joiner UI displays "Signing Error. Relay server unavailable, please try again" — even though the initial join already succeeded and the initiator continues signing on its side.

The user is left with a screen telling them signing failed, while the initiator believes signing is in progress / completed. Tapping **Try again** triggers another `joinKeysign()` call, which hits another 500.

## Steps to reproduce

1. Initiate a keysign on device A (initiator).
2. Scan the QR with device B (joiner) and confirm on the verify screen.
3. Joiner UI ends in the "Signing Error / Relay server unavailable" terminal screen; initiator proceeds with signing.

Reproduces on `agent/4468-thorchain-maya--refunded-inbound-txs--lp` (build 1.0.101 / 101).

## Expected behavior

Once the joiner has successfully registered with the session (201), it should not POST `/router/{sessionId}` again. The joiner UI should transition to "Waiting for keysign start" and then to the keysign flow, matching what the initiator sees.

## Evidence from logs

Joined device timeline (single keysign attempt, no manual retap reported):

```
14:02:09.528  POST /router/{session}  body=[\"<device-id>\"]   →  201 Created    (first call OK)
14:02:11.254  POST /router/{session}  body=[\"<device-id>\"]   →  500
14:02:11.595  SessionApiImpl  Relay request failed with 500, retrying (attempt 1)
14:02:12.602  POST …                                          →  500
14:02:12.956  SessionApiImpl  Relay request failed with 500, retrying (attempt 2)
14:02:13.014  JoinKeysignViewModel  Joining keysign           ← another joinKeysign() call starts
                                                                while prior retry chain still alive
14:02:13.015  POST …                                          →  500
14:02:13.358  SessionApiImpl  Relay request failed with 500, retrying (attempt 1)
14:02:14.361  POST …                                          →  500
14:02:14.728  SessionApiImpl  Relay request failed with 500, retrying (attempt 2)
14:02:14.959  POST …                                          →  500
14:02:15.295  SessionApiImpl  Relay request failed with 500, retrying (attempt 3)
14:02:15.295  JoinKeysignViewModel  Failed to join keysign (HTTP 500)
14:02:16.731  POST …                                          →  500
14:02:17.061  SessionApiImpl  Relay request failed with 500, retrying (attempt 3)
14:02:17.061  JoinKeysignViewModel  Failed to join keysign (HTTP 500)   (second chain finishes)
```

Two observations:
- The first POST returns **201** — the joiner did join. Subsequent POSTs return **500** because the relay rejects duplicate registration of an already-registered party.
- A second `Joining keysign` log appears at 14:02:13 while a previous retry chain is still in flight — i.e. `joinKeysign()` is invoked more than once for a single session, and the calls overlap.

Initiator side meanwhile is happily polling `GET /router/message/{session}/<initiator-id>` (returns `[]`) and continuing the signing flow — for it the join was successful.

## Joiner UI

signing-error

> Sign Transaction
> Signing Error. Please try again. Relay server unavailable, please try again
> [ Try again ]

## Likely root cause

`joinKeysign()` has no re-entrancy guard. The Verify-screen Confirm button calls it directly with no debounce, and the function itself does not check for an in-flight invocation. Once two calls overlap, the first registration succeeds (201) and every later POST gets rejected (500). The retry budget then makes it look worse because each overlapping call burns 3 retries before surfacing the error — and the surfaced error overrides the already-successful state.

## Acceptance criteria

- [ ] On the joiner, only one `POST /router/{sessionId}` is issued per keysign session, even if the user double-taps Confirm or the lambda is recomposed.
- [ ] After a successful join (201), the joiner UI transitions out of the verify screen and never returns to it for the same session.
- [ ] The "Signing Error / Relay server unavailable" screen is no longer reached when the join was actually successful.
- [ ] No regression: legitimate transient relay failures on the *first* `startSession` attempt still surface the existing error UI with **Try again**.

## Related

- #4324 (closed) — same 201→500 server response pattern; closed by adding 3× retry, which doesn't address the duplicate-call cause.
- #4311 (open) — joiner stuck on Signing screen after 500. Different surface (no error shown), but same underlying server-rejects-duplicate behavior.

johnnyluo:
## Summary
- Add an `AtomicBoolean` re-entrancy guard to `joinKeysign()` in `JoinKeysignViewModel`
- Only one `POST /router/{sessionId}` is issued per keysign session, even on double-tap or recomposition of the Confirm button
- The flag is reset in a `finally` block so legitimate "Try again" retries after a genuine first-attempt failure still work correctly

## Root Cause
`joinKeysign()` had no guard against concurrent invocations. The Confirm button is wired to `viewModel::joinKeysign` in 4 places in `JoinKeysignView.kt`. A double-tap (or any recomposition during the async call) could fire two overlapping coroutines. The relay returns 201 on the first POST and 500 on every duplicate, exhausting the retry budget and displaying a false "Relay server unavailable" error — even though the join had already succeeded.

## Issue
Closes #4475

## Test Plan
- [ ] On a 2-of-2 vault, initiate keysign on device A; rapidly double-tap Confirm on device B — joiner should transition to "Waiting for keysign start" without hitting the error screen
- [ ] On a genuinely unavailable relay, the first attempt fails and the "Try again" button successfully re-attempts

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented multiple simultaneous key-signing session starts to avoid duplicate or overlapping joins.
  * Improved error reporting and cleanup for key-signing flows, with clearer failure handling and better logging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4493)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
@@ -101,6 +101,7 @@ import java.net.SocketTimeoutException
 import java.net.UnknownHostException
 import java.util.Locale
 import java.util.UUID
+import java.util.concurrent.atomic.AtomicBoolean
 import javax.inject.Inject
 import kotlin.io.encoding.Base64
 import kotlin.io.encoding.ExperimentalEncodingApi
@@ -250,6 +251,7 @@ constructor(
 
     private var _jobWaitingForKeysignStart: Job? = null
     private var blockaidSimulationJob: Job? = null
+    private val isJoiningKeysign = AtomicBoolean(false)
     private var isNavigateToHome: Boolean = false
 
     private var transactionTypeUiModel: TransactionTypeUiModel? = null
@@ -1380,35 +1382,42 @@ constructor(
     }
 
     fun joinKeysign() {
+        if (!isJoiningKeysign.compareAndSet(false, true)) return
         viewModelScope.launch {
-            withContext(Dispatchers.IO) {
-                try {
-                    Timber.tag("JoinKeysignViewModel").d("Joining keysign")
-                    sessionApi.startSession(_serverAddress, _sessionID, listOf(_localPartyID))
-                    waitForKeysignToStart()
-                    currentState.value = JoinKeysignState.WaitingForKeysignStart
-                } catch (e: HttpException) {
-                    Timber.tag("JoinKeysignViewModel")
-                        .e(
-                            "Failed to join keysign (HTTP %d): %s",
-                            e.statusCode,
-                            e.stackTraceToString(),
-                        )
-                    currentState.value =
-                        if (e.statusCode >= 500) {
-                            JoinKeysignState.Error(JoinKeysignError.RelayUnavailable)
-                        } else {
+            try {
+                withContext(Dispatchers.IO) {
+                    try {
+                        Timber.tag("JoinKeysignViewModel").d("Joining keysign")
+                        sessionApi.startSession(_serverAddress, _sessionID, listOf(_localPartyID))
+                        waitForKeysignToStart()
+                        currentState.value = JoinKeysignState.WaitingForKeysignStart
+                    } catch (e: HttpException) {
+                        Timber.tag("JoinKeysignViewModel")
+                            .e(
+                                "Failed to join keysign (HTTP %d): %s",
+                                e.statusCode,
+                                e.stackTraceToString(),
+                            )
+                        currentState.value =
+                            if (e.statusCode >= 500) {
+                                JoinKeysignState.Error(JoinKeysignError.RelayUnavailable)
+                            } else {
+                                JoinKeysignState.Error(
+                                    JoinKeysignError.FailedToStart(e.message.toString())
+                                )
+                            }
+                    } catch (e: Exception) {
+                        if (e is CancellationException) throw e
+                        Timber.tag("JoinKeysignViewModel")
+                            .e("Failed to join keysign: %s", e.stackTraceToString())
+                        currentState.value =
                             JoinKeysignState.Error(
                                 JoinKeysignError.FailedToStart(e.message.toString())
                             )
-                        }
-                } catch (e: Exception) {
-                    if (e is kotlinx.coroutines.CancellationException) throw e
-                    Timber.tag("JoinKeysignViewModel")
-                        .e("Failed to join keysign: %s", e.stackTraceToString())
-                    currentState.value =
-                        JoinKeysignState.Error(JoinKeysignError.FailedToStart(e.message.toString()))
+                    }
                 }
+            } finally {
+                isJoiningKeysign.set(false)
             }
         }
     }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
@@ -1383,41 +1383,38 @@ constructor(
 
     fun joinKeysign() {
         if (!isJoiningKeysign.compareAndSet(false, true)) return
-        viewModelScope.launch {
-            try {
-                withContext(Dispatchers.IO) {
-                    try {
-                        Timber.tag("JoinKeysignViewModel").d("Joining keysign")
-                        sessionApi.startSession(_serverAddress, _sessionID, listOf(_localPartyID))
-                        waitForKeysignToStart()
-                        currentState.value = JoinKeysignState.WaitingForKeysignStart
-                    } catch (e: HttpException) {
-                        Timber.tag("JoinKeysignViewModel")
-                            .e(
-                                "Failed to join keysign (HTTP %d): %s",
-                                e.statusCode,
-                                e.stackTraceToString(),
-                            )
-                        currentState.value =
-                            if (e.statusCode >= 500) {
-                                JoinKeysignState.Error(JoinKeysignError.RelayUnavailable)
-                            } else {
-                                JoinKeysignState.Error(
-                                    JoinKeysignError.FailedToStart(e.message.toString())
-                                )
-                            }
-                    } catch (e: Exception) {
-                        if (e is CancellationException) throw e
-                        Timber.tag("JoinKeysignViewModel")
-                            .e("Failed to join keysign: %s", e.stackTraceToString())
-                        currentState.value =
+        viewModelScope.safeLaunch {
+            withContext(Dispatchers.IO) {
+                try {
+                    Timber.tag("JoinKeysignViewModel").d("Joining keysign")
+                    sessionApi.startSession(_serverAddress, _sessionID, listOf(_localPartyID))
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1380,35 +1382,42 @@ constructor(
...
-                } catch (e: Exception) {
-                    if (e is kotlinx.coroutines.CancellationException) throw e
-                    Timber.tag("JoinKeysignViewModel")
-                        .e("Failed to join keysign: %s", e.stackTraceToString())
-                    currentState.value =
-                        JoinKeysignState.Error(JoinKeysignError.FailedToStart(e.message.toString()))
+                    }
                 }
+            } finally {
+                isJoiningKeysign.set(false)
```
rkokhatskyi: This release fires as soon as `startSession()` returns whereas Compose only unmounts the Confirm button on the recomposition that picks up `WaitingForKeysignStart`, so a tap landing in that small gap can still enter `joinKeysign()` again and fire the duplicate POST that the flag is supposed to suppress, releasing only on the error branches and letting `tryAgain()` reset before its inner call would close the window and match what the variable name suggests

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1380,35 +1382,42 @@ constructor(
...
+                        currentState.value =
+                            if (e.statusCode >= 500) {
+                                JoinKeysignState.Error(JoinKeysignError.RelayUnavailable)
+                            } else {
+                                JoinKeysignState.Error(
+                                    JoinKeysignError.FailedToStart(e.message.toString())
+                                )
+                            }
+                    } catch (e: Exception) {
+                        if (e is CancellationException) throw e
```
rkokhatskyi: `catch (e: Exception) { if (e is CancellationException) throw e }` is the verbose shape of two clauses whereas the idiomatic form puts `catch (e: CancellationException) { throw e }` first so the wider `Exception` branch never sees cancellation, the manual `is` check works today but becomes fragile if someone later adds an early return before the rethrow

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1380,35 +1382,42 @@ constructor(
     }
 
     fun joinKeysign() {
+        if (!isJoiningKeysign.compareAndSet(false, true)) return
```
rkokhatskyi, line 1387: No unit test accompanies this guard whereas the project convention asks for ViewModel state coverage on bug fixes, a JUnit5 plus Turbine case that drives `joinKeysign()` twice in quick succession and asserts a single `sessionApi.startSession` invocation would lock in the contract that this PR establishes and catch regressions if the flag lifecycle gets refactored

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1380,35 +1382,42 @@ constructor(
...
-                        )
-                    currentState.value =
-                        if (e.statusCode >= 500) {
-                            JoinKeysignState.Error(JoinKeysignError.RelayUnavailable)
-                        } else {
+            try {
+                withContext(Dispatchers.IO) {
+                    try {
+                        Timber.tag("JoinKeysignViewModel").d("Joining keysign")
+                        sessionApi.startSession(_serverAddress, _sessionID, listOf(_localPartyID))
```
rkokhatskyi: The 500 misdiagnosis loop also lives inside `SessionApi.kt:244` `withRelayRetry` which keeps retrying on `5xx` whereas the relay returns 500 for duplicate session POSTs that are not transient, so this guard hides the symptom while any other source of a duplicate POST (process restart, navigation reattach, race with the initiator) still burns the three retries plus seven seconds of backoff before surfacing `RelayUnavailable`, having `startSession` skip the retry on `5xx` (or the relay return 2xx idempotently) would be the durable fix

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1380,35 +1382,42 @@ constructor(
...
+                        waitForKeysignToStart()
+                        currentState.value = JoinKeysignState.WaitingForKeysignStart
```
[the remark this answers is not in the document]
johnnyluo: Fixed in 446056d — switched to `viewModelScope.safeLaunch`.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
johnnyluo: Fixed in 446056d — the flag is now only reset in the error catch blocks. On the success path it stays `true` for the lifetime of the ViewModel, so a tap landing in the recomposition gap can no longer fire a duplicate POST.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
johnnyluo: Fixed in 446056d — reordered to `catch (e: CancellationException) { throw e }` before `catch (e: Exception)`.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
johnnyluo: Agreed — a dedicated JUnit5 + Turbine test that drives `joinKeysign()` twice and asserts a single `startSession` invocation would be the right coverage. `JoinKeysignViewModel` has a large DI graph; filing a follow-up to add it rather than blocking this fix.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
johnnyluo: Valid architectural point. `withRelayRetry` burns 3 retries + 7 s of backoff on every duplicate session POST because it treats all 5xx as transient. The guard here fixes the double-tap source, but process restart or navigation reattach can still trigger the same loop. Filed #4475 already tracks the root issue — a follow-up to either skip retries on 500 inside `startSession` or make the relay return 2xx idempotently would be the durable fix.

PR #4494, opened 2026-05-14
fix(keysign): stop polling loop when error state is reached

### LINKED ISSUE #4477: JoinKeysign: error logged every second while error screen is displayed
## Summary

When `JoinKeysignViewModel` enters the keysign error state (e.g. `PreSigningOutput contains error: Error_missing_input_utxos`), the polling loop in `waitForKeysignToStart()` keeps calling `checkKeygenStarted()` once per second. Each call re-throws the same exception and re-sets the same error state, so the same stack trace is logged repeatedly for as long as the error screen is on screen.

## Steps to reproduce

1. Initiate a UTXO keysign (e.g. BTC) on a joiner device where the payload's UTXOs are missing/invalid so `UtxoHelper.getPreSignedImageHash` fails with `Error_missing_input_utxos`.
2. Observe the `Sign Transaction` error screen ("Signing Error. Please try again. ...").
3. Watch logcat — the same `Failed to check keysign start` + `IllegalStateException: PreSigningOutput contains error: Error_missing_input_utxos` stack trace is logged every ~1 second.

## Expected

The error is logged once. The polling loop stops once the UI has transitioned to the error state, so no further work is done until the user taps `Try again`.

## Observed

Logcat is flooded with the same exception every second while the error screen is shown, and the relay `GET /router/start/<sessionId>` request is also issued every second.

## Root cause

`waitForKeysignToStart()` only exits its `while (isActive)` loop when `checkKeygenStarted()` returns `true`. On exceptions, `checkKeygenStarted()` sets `currentState = JoinKeysignState.Error(...)` and returns `false`, so the loop sleeps 1s and retries forever.

## Log snippet (session id / device names redacted)

```
JoinKeysignViewModel  D  Keysign committee: [<device-a>, <device-b>]
JoinKeysignViewModel  D  local party: <device-b>
JoinKeysignViewModel  E  Failed to check keysign start
    java.lang.IllegalStateException: PreSigningOutput contains error: Error_missing_input_utxos
        at com.vultisig.wallet.data.crypto.UtilsKt.checkError(Utils.kt:22)
        at com.vultisig.wallet.data.chains.helpers.UtxoHelper.getPreSignedImageHash(UtxoHelper.kt:53)
        at com.vultisig.wallet.data.chains.helpers.SigningHelper.getKeysignMessages(SigningHelper.kt:144)
        at com.vultisig.wallet.ui.models.keysign.JoinKeysignViewModel.checkKeygenStarted(JoinKeysignViewModel.kt:1460)
```

(Sequence repeats once per second while the error screen is visible.)

johnnyluo:
## Summary
- In `waitForKeysignToStart()`, the `while (isActive)` loop now exits immediately when `currentState` is `Error`, preventing the same exception from being re-thrown and re-logged every second while the error screen is shown

## Issue
Closes #4477

## Root cause
`checkKeygenStarted()` catches exceptions, sets `currentState = Error(...)`, and returns `false`. The loop previously only exited on `true` (success) or coroutine cancellation — so on error it kept sleeping 1s and retrying forever, flooding logcat and issuing a relay GET every second.

## Test Plan
- [ ] Trigger a UTXO keysign (e.g. BTC) on a joiner where the payload has missing/invalid UTXOs so `UtxoHelper.getPreSignedImageHash` throws `Error_missing_input_utxos`
- [ ] Confirm the error screen appears and the exception is logged exactly once
- [ ] Confirm no further `Failed to check keysign start` logs appear while the error screen is shown

## Summary by CodeRabbit

* **Bug Fixes**
  * Join keysign flow now correctly honors cancellations and surfaces a clear error when message resolution fails instead of producing misleading states or retries.

* **Refactor**
  * Message-resolution has been reorganized into a single step that detects missing payloads earlier and reports failures consistently during the join process, improving error clarity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4494)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
@@ -1511,6 +1511,8 @@ constructor(
             }
         } catch (e: KeysignMessagesException) {
             throw e
+        } catch (ce: CancellationException) {
+            throw ce
         } catch (e: Exception) {
             throw KeysignMessagesException(e.message ?: "Failed to resolve messages to sign")
         }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
@@ -229,6 +229,8 @@ constructor(
         private const val ETH_SIGN_TYPED_DATA_V4 = "eth_signTypedData_v4"
     }
 
+    private class KeysignMessagesException(message: String) : Exception(message)
+
     private val args = savedStateHandle.toRoute<Route.Keysign.Join>()
     private val vaultId: String = args.vaultId
     private val qrBase64: String = args.qr
@@ -1445,11 +1447,22 @@ constructor(
             viewModelScope.launch {
                 withContext(Dispatchers.IO) {
                     while (isActive) {
-                        if (checkKeygenStarted()) {
-                            currentState.value = JoinKeysignState.Keysign
-                            return@withContext
-                        }
-                        if (currentState.value is JoinKeysignState.Error) {
+                        try {
+                            if (checkKeygenStarted()) {
+                                currentState.value = JoinKeysignState.Keysign
+                                return@withContext
+                            }
+                            if (currentState.value is JoinKeysignState.Error) {
+                                return@withContext
+                            }
+                        } catch (e: KeysignMessagesException) {
+                            Timber.e(e, "Failed to prepare messages to sign")
+                            currentState.value =
+                                JoinKeysignState.Error(
+                                    JoinKeysignError.FailedToCheck(
+                                        e.message ?: "Failed to prepare messages to sign"
+                                    )
+                                )
                             return@withContext
                         }
                         delay(1000)
@@ -1464,31 +1477,11 @@ constructor(
             Timber.d("Keysign committee: $_keysignCommittee")
             Timber.d("local party: $_localPartyID")
             if (this._keysignCommittee.contains(_localPartyID)) {
-                when {
-                    _keysignPayload != null -> {
-                        val payload = _keysignPayload ?: error("keysignPayload unexpectedly null")
-                        messagesToSign =
-                            SigningHelper.getKeysignMessages(
-                                payload = payload,
-                                vault = _currentVault,
-                            )
-                    }
-
-                    customMessagePayload != null -> {
-                        val payload =
-                            customMessagePayload ?: error("customMessagePayload unexpectedly null")
-                        messagesToSign = SigningHelper.getKeysignMessages(payload)
-                    }
-
-                    else -> {
-                        Timber.e("Both keysign payload and custom message payload are null")
-                        currentState.value = JoinKeysignState.Error(JoinKeysignError.InvalidQr)
-                        return false
-                    }
-                }
-
+                resolveMessagesToSign()
                 return true
             }
+        } catch (e: KeysignMessagesException) {
+            throw e
         } catch (e: Exception) {
             if (e is kotlinx.coroutines.CancellationException) throw e
             Timber.e(e, "Failed to check keysign start")
@@ -1496,6 +1489,33 @@ constructor(
                 JoinKeysignState.Error(JoinKeysignError.FailedToCheck(e.message.toString()))
         }
         return false
+    }
+
+    private fun resolveMessagesToSign() {
+        try {
+            when {
+                _keysignPayload != null -> {
+                    val payload = _keysignPayload ?: error("keysignPayload unexpectedly null")
+                    messagesToSign =
+                        SigningHelper.getKeysignMessages(payload = payload, vault = _currentVault)
+                }
+
+                customMessagePayload != null -> {
+                    val payload =
+                        customMessagePayload ?: error("customMessagePayload unexpectedly null")
+                    messagesToSign = SigningHelper.getKeysignMessages(payload)
+                }
+
+                else ->
+                    throw KeysignMessagesException(
+                        "Both keysign payload and custom message payload are null"
+                    )
+            }
+        } catch (e: KeysignMessagesException) {
+            throw e
+        } catch (e: Exception) {
+            throw KeysignMessagesException(e.message ?: "Failed to resolve messages to sign")
+        }
     }
 
     fun enableNavigationToHome() {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1449,7 +1449,9 @@ constructor(
                             currentState.value = JoinKeysignState.Keysign
                             return@withContext
                         }
-                        // backoff 1s
+                        if (currentState.value is JoinKeysignState.Error) {
```
rkokhatskyi: Pre-fix the loop overwrote `Error(FailedToCheck)` on the next successful poll which self-healed transient `IOException` and 5xx blips during the LAN/relay handshake, whereas now the first blip is terminal and `tryAgain()` at line 1416 routes `FailedToCheck` to the `else` branch at line 1433 which navigates back rather than re-running `joinKeysign()`, so the user has no path to recover from a transient failure without rescanning the QR, worth narrowing loop exit to genuinely terminal causes such as `InvalidQr` and parse failures via a typed result, or widening that `tryAgain` branch to retry `FailedToCheck`

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1449,7 +1449,9 @@ constructor(
...
+                            return@withContext
```
rkokhatskyi, line 1463: Reading `currentState.value` to drive control flow couples polling business logic to a UI state contract whereas `checkKeygenStarted` already knows the outcome since it sets the error itself at lines 1485 and 1495, returning a sealed `PollResult` would let the loop be an exhaustive `when` with no out-of-band state read

```kotlin
sealed interface PollResult {
    data object Continue : PollResult
    data object Started : PollResult
    data class Terminal(val error: JoinKeysignError) : PollResult
}
```

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1449,7 +1449,9 @@ constructor(
...
+                        }
```
rkokhatskyi, line 1464: The sibling `waitForKeygenToStart` at `app/src/main/java/com/vultisig/wallet/ui/models/keygen/JoinKeygenViewModel.kt:365` deliberately swallows exceptions and keeps polling forever via `emptyList()` at line 374, so keygen and keysign join flows now follow opposite strategies on identical session-start polling without a documented rationale, worth picking one policy and applying it to both for consistency since both share the same relay session-start contract

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1449,7 +1449,9 @@ constructor(
...
                         delay(1000)
```
rkokhatskyi, line 1465: `JoinKeysignViewModel` has no tests covering `waitForKeysignToStart` or `checkKeygenStarted` so this behavioral fix can regress silently, a `runTest` spec with a fake `sessionApi.checkCommittee` that throws and an assertion that the polling job completes after a single cycle would lock the behavior described in #4477

rkokhatskyi (changes requested):

johnnyluo:
close this one off , I think this PR actually break the logic

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1445,11 +1447,24 @@ constructor(
...
                 withContext(Dispatchers.IO) {
                     while (isActive) {
-                        if (checkKeygenStarted()) {
-                            currentState.value = JoinKeysignState.Keysign
+                        try {
+                            if (checkKeygenStarted()) {
+                                currentState.value = JoinKeysignState.Keysign
+                                return@withContext
+                            }
+                            if (currentState.value is JoinKeysignState.Error) {
```
johnnyluo: These check is not necessary anymore , can you please remove it

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1462,31 +1477,11 @@ constructor(
...
-                    }
-                }
-
+                resolveMessagesToSign()
                 return true
             }
+        } catch (e: KeysignMessagesException) {
+            throw e
         } catch (e: Exception) {
             if (e is kotlinx.coroutines.CancellationException) throw e
```
johnnyluo: can you please update `if (e is kotlinx.coroutines.CancellationException) throw e` to use a dedicate `catch`

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1462,40 +1474,48 @@ constructor(
...
-                        return false
-                    }
-                }
-
+                resolveMessagesToSign()
                 return true
             }
+        } catch (e: KeysignMessagesException) {
+            throw e
+        } catch (_: CancellationException) {
```
johnnyluo: can you please just do 
```
catch(ce: CancellationException) {
  throw ce
}
```

aminsato:
@coderabbitai full review

rkokhatskyi:
@Vaulty-bot address

rkokhatskyi:
@Vaulty-bot pick up

rkokhatskyi:
@Vaulty-bot resolve all comments

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1445,11 +1447,21 @@ constructor(
...
+                            currentState.value =
+                                JoinKeysignState.Error(
+                                    JoinKeysignError.FailedToCheck(
+                                        e.message ?: "Failed to prepare messages to sign"
+                                    )
+                                )
                             return@withContext
                         }
-                        // backoff 1s
                         delay(1000)
```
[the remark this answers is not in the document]
johnnyluo: That is fine , because the exception in `resolveMessagesToSign` which is not recoverable will become `KeysignMessagesException` , which will stop the polling

Other exception is likely networking related which might be recoverable , thus it should continue polling

PR #4496, opened 2026-05-14
fix: preflight THORChain LP adds against pool/mimir/inbound pause state

### LINKED ISSUE #4471: THORChain/Maya: no preflight against pool/mimir/chain pause state — users can broadcast txs the network will refund
## Summary

The app lets users broadcast THORChain (and MayaChain) inbound transactions — LP add-liquidity, savers deposits, loans, bonds — even when the network has paused or halted the corresponding action. The network reliably refunds these inbounds, but the user pays the inbound gas, waits through keysign, and then has to discover the refund from an explorer. Most of the signals needed to prevent this are publicly available from THORChain; the app either doesn't fetch them or fetches them and ignores them.

This is a companion to #4468 (which is about correctly *reporting* refunds after they happen). This issue is about *preventing* the refund in the first place by validating network state before keysign.

## Reproduction

- Tx: `9D30542FEB46852854B86613D1ED1EC8D37619A1B1978543A12F1AF3279BE74D`
- User intent: LP Mint into `ETH.USDT-0xdac17f958d2ee523a2206206994597c13d831ec7` from the RUNE side
- Memo: `+:ETH.USDT-0XDAC...1EC7:0x14F6...89B6`
- Network outcome: refund with reason *"Unable to add liquidity, deposits are paused for asset (eth.usdt-0xdac17f958d2ee523a2206206994597c13d831ec7): internal error"*
- Pre-broadcast knowable signals at the time of submit:
  - `/thorchain/mimir` exposes a `PAUSELP-ETH-USDT` (or similar per-asset) key
  - `/thorchain/pool/ETH.USDT-0xdac…1ec7` returns `status` (`Available` / `Staged` / `Suspended`)
  - `/thorchain/inbound_addresses` for `ETH` exposes `chain_lp_actions_paused`, `halted`, `global_trading_paused`, `chain_trading_paused`

None of those three were consulted before the inbound was signed and broadcast.

## What the code does today

| Signal | Current state | File:line |
|---|---|---|
| Inbound `halted` / `chain_lp_actions_paused` / `global_trading_paused` | Read **only** to decide whether to auto-fill an address field. Not used to gate the form or the Sign button. | `app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt:898-906`, `:1024-1032` |
| Pool `status` field | Not deserialized at all — `ThorChainPoolJson` only keeps `asset` and `assetTorPrice` | `data/src/main/kotlin/com/vultisig/wallet/data/api/models/quotes/ThorChainPoolJson.kt:10-15` |
| Mimir keys (`PAUSELP`, `PAUSELP-<ASSET>`, `HALT<CHAIN>CHAIN`, `HALT<CHAIN>TRADING`, `HALT<CHAIN>LP`, `HALTSWAPPING`, `PAUSELOANS`) | `/mimir` is called but only for `CHURNINTERVAL` in node-staking flows; pause keys are never read | `data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt:560`, `data/src/main/kotlin/com/vultisig/wallet/data/api/MayaChainApi.kt:284` |
| LP add-liquidity path | `createAddLiquidityTransaction()` runs *zero* network-state checks before building the keysign payload — only validates pool selected, amount > 0, paired address resolvable | `app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt:1697-1772` |
| Swap quote `error` field | Properly checked: `ThorChainQuoteSource.kt:95-97, 124` maps quote errors into `SwapException`; halts surface as `SwapRouteNotAvailable` | `data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/ThorChainQuoteSource.kt:62, 95-97, 124` |

Only the swap path has a real (soft) guard, and even that one only catches halts the quote endpoint chooses to surface — not per-asset LP pauses that don't affect swapping.

## Suggested fix

Add a preflight validation layer for THORChain/Maya inbound transactions. Concretely:

1. **Deserialize `status` in `ThorChainPoolJson`** (and the Maya equivalent). Disable paused pools in the LP pool picker — show a \"Deposits paused\" caption rather than letting users select them.
2. **Add a mimir-aware repository** (`ThorMimirRepository.kt`) that exposes typed accessors for the pause/halt keys used by the wallet's flows:
   - `isLpPaused(asset: String): Boolean` → checks `PAUSELP` global + `PAUSELP-<ASSET>`
   - `isLpHalted(chain: Chain): Boolean` → checks `HALT<CHAIN>LP` and `HALT<CHAIN>CHAIN`
   - `isSwapHalted(chain: Chain): Boolean` → `HALTSWAPPING` + `HALT<CHAIN>SWAP` + `HALT<CHAIN>TRADING`
   - `areLoansPaused(): Boolean` → `PAUSELOANS`
   Cache for a short TTL (~30s) so multiple validators in one screen don't re-fetch.
3. **Gate `createAddLiquidityTransaction()` on a preflight**: in parallel, fetch (a) the pool `status`, (b) the relevant mimir keys, (c) the inbound `chain_lp_actions_paused` for both the inbound chain (THOR/Maya) and the asset chain. Throw a typed `PreflightBlockedException(reason)` that the Verify screen renders as an error state instead of broadcasting.
4. **Apply the same guard for savers, loan-open, bond, and any future inbound** — make it a single \"thorchain action preflight\" usecase the relevant `create*Transaction` calls invoke.
5. **Reinforce the swap path**: keep the quote-error gate, but additionally read inbound `chain_trading_paused` for the destination chain so users get a clearer message than \"route not available\".
6. Mirror all of the above for Maya (`/mayachain/mimir`, `/mayachain/pools/<asset>`, `/mayachain/inbound_addresses`).

The preflight should live in `data/src/main/kotlin/com/vultisig/wallet/data/usecases/` (or a new `preflight/` package) so swap, deposit, and any future ThorChain flow can share it.

## Test plan

- [ ] Unit-test the new `ThorMimirRepository` against canned `/mimir` responses (no pauses, global pause, per-asset pause, chain halt).
- [ ] Unit-test pool status deserialization for `Available` / `Staged` / `Suspended`.
- [ ] Unit-test the preflight usecase: returns block for paused asset, paused chain, halted chain; passes for clean state.
- [ ] Manual: pick a real paused pool (or temporarily mock the mimir response to set `PAUSELP-ETH-USDT`) and confirm the LP form refuses to advance to keysign with the network's pause reason surfaced inline.
- [ ] Manual regression: a healthy LP add still completes end-to-end.
- [ ] Manual: trigger the swap path with a halted destination chain and confirm the message changes from generic \"route not available\" to a pause-aware reason.

## Severity

Lossy UX: users spend gas on inbound transactions that the network will refund, then have no in-app indication the action failed (compounded by #4468). For LP and savers this also means users may think they hold a position they don't, which has real \"is my money missing?\" support cost.

## Relation to other issues

- #4468 — fix the *reporting* of refunds. This issue fixes the *prevention*.
- #4331 — LP-specific UI for half-deposit / refund timeout. Adjacent; the preflight here would prevent a subset of the cases that issue currently handles after the fact.

aminsato:
Closes #4471

## Summary

Before signing a THORChain add-liquidity inbound, fetch the pool status, mimir pause/halt keys, and the inbound `chain_lp_actions_paused` flag in parallel and refuse to build the keysign payload when the network would refund the deposit. This spares users from spending inbound gas on a transaction THORChain immediately returns.

Scope is intentionally narrow: **THORChain LP add-liquidity only**. Maya, savers, loan-open, bond, and the swap-path reinforcement described in #4471 are left for follow-ups so this change is easy to review.

## Screenshot

Image

### What changed

- **`ThorChainPoolJson`** — deserializes the `status` field (Available / Staged / Suspended).
- **`ThorChainApi`** — adds `getPool(asset)` and `getMimir()` against `/thorchain/pool/<asset>` and `/thorchain/mimir`.
- **`ThorMimirRepository`** — typed view over the mimir flag map (`PAUSELP`, `PAUSELP-<CHAIN>-<TICKER>`, `HALT<CHAIN>LP`, `HALT<CHAIN>CHAIN`) with a 30s TTL so several validators in one screen hit the network once.
- **`ThorChainLpPreflightUseCase`** — runs the four signal checks in parallel, returns the first blocking reason or `null` (fail-open on transient errors so a thornode hiccup doesn't block a healthy deposit).
- **`DepositFormViewModel.createAddLiquidityTransaction()`** — invokes the preflight before building the keysign payload and surfaces the network's pause reason inline via the existing `InvalidTransactionDataException` path.
- Localized error strings added across all 10 locales.

## Test plan

- [x] Unit tests pass for `ThorMimirRepository` (9 tests) — covers global PAUSELP, per-asset PAUSELP-CHAIN-TICKER, HALT chain keys, case-insensitivity, zero-value off, and TTL caching.
- [x] Unit tests pass for `ThorChainLpPreflightUseCase` (11 tests) — covers clean state, each blocking signal, unrelated-chain inbound entries, network-error fail-open, and Available-status case handling.
- [x] `:data:compileDebugKotlin` and `:app:compileDebugKotlin` clean.
- [x] `:app:compileDebugUnitTestKotlin` clean.
- [ ] Manual: pick a real paused pool (or temporarily mock the mimir response to set `PAUSELP-ETH-USDT`) and confirm the LP form refuses to advance to keysign with the network's pause reason surfaced inline.
- [ ] Manual regression: a healthy LP add still completes end-to-end.

## Summary by CodeRabbit

* **New Features**
  * Preflight checks now prevent THORChain LP deposits to paused, halted, or unavailable pools and surface a localized error dialog with a retry option.

* **Localization**
  * Added localized LP deposit error messages (paused, chain halted, pool unavailable) across multiple languages.

* **Tests**
  * New unit tests cover preflight decision logic, pool status handling, fail-open behavior, and caching.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4496)

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorMimirRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorMimirRepository.kt
@@ -16,8 +16,10 @@ import kotlinx.coroutines.sync.withLock
 interface ThorMimirRepository {
     /**
      * True when add-liquidity is paused either globally (`PAUSELP`) or for the specific [pool]
-     * (`PAUSELP-<CHAIN>-<TICKER>`). [pool] is the canonical THORChain pool identifier (e.g.
-     * `ETH.USDT-0xdac17f958d2ee523a2206206994597c13d831ec7`).
+     * (`PAUSELPDEPOSIT-<CHAIN>-<ASSET[-CONTRACT]>`). [pool] is the canonical THORChain pool
+     * identifier (e.g. `ETH.USDT-0xdac17f958d2ee523a2206206994597c13d831ec7`). The contract suffix
+     * is preserved so ERC20 pools — which thornode pauses by full asset id — are matched correctly
+     * (the 2026-04-10 `BTC.BTC` refund incident surfaced this requirement).
      */
     suspend fun isLpPaused(pool: String): Boolean
 
@@ -38,7 +40,7 @@ internal class ThorMimirRepositoryImpl @Inject constructor(private val thorChain
     override suspend fun isLpPaused(pool: String): Boolean {
         val mimir = mimir()
         if (mimir.isOn(KEY_GLOBAL_LP_PAUSE)) return true
-        val key = "$KEY_GLOBAL_LP_PAUSE-${pool.toMimirAssetSuffix()}"
+        val key = "$KEY_PER_POOL_LP_DEPOSIT_PAUSE-${pool.toMimirAssetSuffix()}"
         return mimir.isOn(key)
     }
 
@@ -67,17 +69,15 @@ internal class ThorMimirRepositoryImpl @Inject constructor(private val thorChain
     private fun Map<String, Long>.isOn(key: String): Boolean = (this[key.uppercase()] ?: 0L) > 0L
 
     /**
-     * Mimir per-asset pause keys are formatted as `PAUSELP-<CHAIN>-<TICKER>` (no contract suffix,
-     * dashes instead of dots, all uppercase). Convert a canonical pool id (`ETH.USDT-0xdac...`) to
-     * that form.
+     * Per-pool LP-deposit pause keys are formatted as `PAUSELPDEPOSIT-<CHAIN>-<ASSET[-CONTRACT]>`:
+     * dashes instead of dots, full asset identifier including any ERC20 contract suffix, all
+     * uppercase. Convert a canonical pool id (`ETH.USDT-0xdac...`) to that form.
      */
-    private fun String.toMimirAssetSuffix(): String {
-        val withoutContract = substringBefore('-')
-        return withoutContract.replace('.', '-').uppercase()
-    }
+    private fun String.toMimirAssetSuffix(): String = replace('.', '-').uppercase()
 
     private companion object {
         const val TTL_MILLIS = 30_000L
         const val KEY_GLOBAL_LP_PAUSE = "PAUSELP"
+        const val KEY_PER_POOL_LP_DEPOSIT_PAUSE = "PAUSELPDEPOSIT"
     }
 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorMimirRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorMimirRepository.kt
@@ -17,9 +17,8 @@ interface ThorMimirRepository {
     /**
      * True when add-liquidity is paused either globally (`PAUSELP`) or for the specific [pool]
      * (`PAUSELPDEPOSIT-<CHAIN>-<ASSET[-CONTRACT]>`). [pool] is the canonical THORChain pool
-     * identifier (e.g. `ETH.USDT-0xdac17f958d2ee523a2206206994597c13d831ec7`). The contract suffix
-     * is preserved so ERC20 pools — which thornode pauses by full asset id — are matched correctly
-     * (the 2026-04-10 `BTC.BTC` refund incident surfaced this requirement).
+     * identifier (e.g. `ETH.USDT-0xdac17f958d2ee523a2206206994597c13d831ec7`). The full asset id
+     * including any ERC20 contract suffix is used, since thornode pauses ERC20 pools by full id.
      */
     suspend fun isLpPaused(pool: String): Boolean
 
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorMimirRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorMimirRepository.kt
@@ -1,0 +1,83 @@
+package com.vultisig.wallet.data.repositories
+
+import com.vultisig.wallet.data.api.ThorChainApi
+import javax.inject.Inject
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+/**
+ * Typed view over THORChain's `/thorchain/mimir` flags. Mimir keys are dynamic on-chain governance
+ * values used (among other things) to pause or halt specific actions. THORChain treats any positive
+ * integer value for these keys as "on"; absent keys are off.
+ *
+ * Results are cached for a short TTL so several validators in one screen (e.g. the LP preflight
+ * checking global + per-asset + chain LP halt) hit the network only once.
+ */
+interface ThorMimirRepository {
+    /**
+     * True when add-liquidity is paused either globally (`PAUSELP`) or for the specific [pool]
+     * (`PAUSELP-<CHAIN>-<TICKER>`). [pool] is the canonical THORChain pool identifier (e.g.
+     * `ETH.USDT-0xdac17f958d2ee523a2206206994597c13d831ec7`).
+     */
+    suspend fun isLpPaused(pool: String): Boolean
+
+    /**
+     * True when LP activity for the given chain (THORChain inbound-chain identifier, e.g. `ETH`,
+     * `BTC`, `BSC`) is halted via either `HALT<CHAIN>LP` or `HALT<CHAIN>CHAIN`.
+     */
+    suspend fun isLpHalted(chainPrefix: String): Boolean
+}
+
+internal class ThorMimirRepositoryImpl @Inject constructor(private val thorChainApi: ThorChainApi) :
+    ThorMimirRepository {
+
+    private val mutex = Mutex()
+    private var cached: Map<String, Long>? = null
+    private var cachedAtMillis: Long = 0L
+
+    override suspend fun isLpPaused(pool: String): Boolean {
+        val mimir = mimir()
+        if (mimir.isOn(KEY_GLOBAL_LP_PAUSE)) return true
+        val key = "$KEY_GLOBAL_LP_PAUSE-${pool.toMimirAssetSuffix()}"
+        return mimir.isOn(key)
+    }
+
+    override suspend fun isLpHalted(chainPrefix: String): Boolean {
+        val mimir = mimir()
+        val upper = chainPrefix.uppercase()
+        return mimir.isOn("HALT${upper}LP") || mimir.isOn("HALT${upper}CHAIN")
+    }
+
+    private suspend fun mimir(): Map<String, Long> =
+        mutex.withLock {
+            val now = nowMillis()
+            val current = cached
+            if (current != null && now - cachedAtMillis < TTL_MILLIS) {
+                current
+            } else {
+                val fresh = thorChainApi.getMimir().mapKeys { it.key.uppercase() }
+                cached = fresh
+                cachedAtMillis = now
+                fresh
+            }
+        }
+
+    private fun nowMillis(): Long = System.currentTimeMillis()
+
+    private fun Map<String, Long>.isOn(key: String): Boolean = (this[key.uppercase()] ?: 0L) > 0L
+
+    /**
+     * Mimir per-asset pause keys are formatted as `PAUSELP-<CHAIN>-<TICKER>` (no contract suffix,
+     * dashes instead of dots, all uppercase). Convert a canonical pool id (`ETH.USDT-0xdac...`) to
+     * that form.
+     */
+    private fun String.toMimirAssetSuffix(): String {
+        val withoutContract = substringBefore('-')
+        return withoutContract.replace('.', '-').uppercase()
+    }
+
+    private companion object {
+        const val TTL_MILLIS = 30_000L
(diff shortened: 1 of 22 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorMimirRepository.kt
```diff
@@ -0,0 +1,83 @@
...
+    ThorMimirRepository {
+
+    private val mutex = Mutex()
+    private var cached: Map<String, Long>? = null
+    private var cachedAtMillis: Long = 0L
+
+    override suspend fun isLpPaused(pool: String): Boolean {
+        val mimir = mimir()
+        if (mimir.isOn(KEY_GLOBAL_LP_PAUSE)) return true
+        val key = "$KEY_GLOBAL_LP_PAUSE-${pool.toMimirAssetSuffix()}"
```
rkokhatskyi: The per-pool key built here is `PAUSELP-<CHAIN>-<TICKER>` whereas thornode pauses per-pool LP deposits via `PAUSELPDEPOSIT-<CHAIN>-<ASSET_PLUS_CONTRACT>` (the form the vultisig-windows SDK uses in `poolPauseMimirKey` and the form documented as the cause of the 2026-04-10 `BTC.BTC` incident where `pool.status` reported `Available` and the handler refunded with `deposits are paused for asset (btc.btc): internal error`), so this check will never fire for the very scenario `#4471` cites since `toMimirAssetSuffix` also strips the `-<contract>` suffix which would be required for ERC20 pools like `ETH.USDT-0xdac17f...`. The mimir test that proves the check works mocks `PAUSELP-ETH-USDT` rather than the real flag thornode actually sets which is why this slipped through.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/ThorChainLpPreflightUseCase.kt
```diff
@@ -0,0 +1,109 @@
...
+
+        val inbound =
+            inboundDeferred.await()?.firstOrNull { it.chain.equals(chainPrefix, ignoreCase = true) }
+        if (inbound != null) {
+            val blocked =
+                inbound.halted || inbound.chainLPActionsPaused || inbound.globalTradingPaused
+            if (blocked) {
+                return@coroutineScope ThorChainLpPreflightBlock.InboundLpPaused(chainPrefix)
+            }
+        }
```
rkokhatskyi: This block calls `inboundDeferred.await()` for the asset chain whenever the preflight runs, but the call site gates the preflight on `chain == Chain.ThorChain` so the user is always depositing RUNE which never traverses the asset-chain inbound, whereas `FunctionCallAddThorLP.swift:110-125` deliberately skips the inbound check when `tx.coin.chain == .thorChain` and the vultisig-windows `assertPoolDepositable` excludes inbound flags from its fail-fast gate, so this block can refuse valid RUNE-side pending LPs when only the asset-side inbound is paused which is a divergence from iOS and Extension.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/ThorChainLpPreflightUseCase.kt
```diff
@@ -0,0 +1,109 @@
...
+            if (status != null && !status.equals(POOL_STATUS_AVAILABLE, ignoreCase = true)) {
+                return@coroutineScope ThorChainLpPreflightBlock.PoolNotAvailable(pool, status)
+            }
+        }
+
+        val inbound =
+            inboundDeferred.await()?.firstOrNull { it.chain.equals(chainPrefix, ignoreCase = true) }
+        if (inbound != null) {
+            val blocked =
+                inbound.halted || inbound.chainLPActionsPaused || inbound.globalTradingPaused
```
rkokhatskyi: Bucketing `halted`, `chainLPActionsPaused`, and `globalTradingPaused` into one `InboundLpPaused` reason loses the distinction between three different network states, and `toError()` then maps all three to `deposit_error_lp_halted_chain` which reads `Liquidity actions on <chain> are currently halted` so a user blocked by `globalTradingPaused` sees a chain-specific LP message that does not match the actual reason which is a global trading pause unrelated to LP or to that chain.

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorMimirRepository.kt
aminsato: Good catch — fixed in a1bcdccf4. The per-pool key now builds as `PAUSELPDEPOSIT-<CHAIN>-<ASSET[-CONTRACT]>` (full asset id including the ERC20 contract, dashes instead of dots, uppercase) instead of the wrong `PAUSELP-<CHAIN>-<TICKER>` form. The mimir tests were rewritten to assert against the real flag (`PAUSELPDEPOSIT-ETH-USDT-0XDAC17F958D2EE523A2206206994597C13D831EC7`), and a separate case covers native pools like `BTC.BTC` → `PAUSELPDEPOSIT-BTC-BTC`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/ThorChainLpPreflightUseCase.kt
aminsato: Agreed — fixed in a1bcdccf4. The asset-chain inbound check has been dropped entirely from the preflight, since the call site only invokes it when `chain == Chain.ThorChain` (RUNE-side deposit). The mimir `HALT<CHAIN>LP`/`HALT<CHAIN>CHAIN` checks remain — those capture LP-specific halts independent of inbound trading state, matching iOS `FunctionCallAddThorLP` and the extension's `assertPoolDepositable`. Asset-side preflight is left as a follow-up if/when we wire LP from the asset chain.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/ThorChainLpPreflightUseCase.kt
aminsato: Fixed in a1bcdccf4 by removing the bucketed `InboundLpPaused` reason entirely (along with the asset-chain inbound check it gated on — see the sibling thread). The misleading "Liquidity actions on <chain> are currently halted" message no longer fires for global trading pauses.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/ThorChainLpPreflightUseCase.kt
```diff
@@ -0,0 +1,109 @@
...
+            if (blocked) {
+                return@coroutineScope ThorChainLpPreflightBlock.InboundLpPaused(chainPrefix)
+            }
```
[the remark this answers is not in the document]
aminsato: Fixed in a1bcdccf4. Replaced `runCatching` with a `probe { ... }` helper that catches `Exception` but rethrows `CancellationException`, and added a `cancelAll(...)` call on the still-pending deferreds at each early-return site so the preflight doesn't wait on probes whose result is no longer needed.

aminsato:
@coderabbitai full review

data/src/test/kotlin/com/vultisig/wallet/data/usecases/ThorChainLpPreflightUseCaseTest.kt
```diff
@@ -0,0 +1,106 @@
...
+    fun `pool status Available is not flagged regardless of case`() = runTest {
+        coEvery { api.getPool(POOL_USDT) } returns
+            ThorChainPoolJson(
+                asset = POOL_USDT,
+                assetTorPrice = BigInteger.ZERO,
+                status = "available",
+            )
+
+        assertNull(useCase(POOL_USDT))
+    }
```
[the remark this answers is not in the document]
aminsato: Half taken in 4adf78555:
- **Multi-signal priority** — added two precedence tests (`mimir LP pause wins over chain halt and pool status` and `chain halt wins over pool status`) so the ordering contract is locked down.
- **Inbound `chain_lp_actions_paused`** — not applicable: the inbound check was removed entirely in a1bcdccf4 (RUNE-side preflight only) per the sibling reviewer thread, so there is no `InboundLpPaused` variant to assert against.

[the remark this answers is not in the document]
aminsato: Comment updated in 6648d0283 — the call-site now says "pool status and the relevant mimir pause keys" only (no inbound mention).

aminsato:
@coderabbitai full review

johnnyluo (approved):
LGTM

PR #4508, opened 2026-05-15
fix(circle-defi): local cache failure no longer reverts account creation

### LINKED ISSUE #4088: [Fix] Circle Open Account treats local save failure as creation failure [Circle USDC DeFi]
# [Fix] Circle Open Account treats local save failure as creation failure [Circle USDC DeFi]

## Problem

`CircleDeFiPositionsViewModel.onCreateAccount` wraps both the remote `circleApi.createScAccount(...)` call and the local `scaCircleAccountRepository.saveAccount(...)` DataStore write in the same `runCatching { }` block. If the remote Circle account is created successfully but the local cache write throws (DataStore IO error, disk full, etc.), the user sees the "Failed to create Circle MSCA account" error snackbar and the UI stays on `isAccountOpen = false` — even though the account exists server-side.

This was flagged by CodeRabbit on PR #4087 but pre-exists the snackbar-type fix, so it's tracked separately.

## Expected Behavior

A local cache failure should not be reported as an account-creation failure. The remote account already exists, so the UI should either:
- Set `isAccountOpen = true` and show a soft warning that the local cache could not be persisted (next launch will self-heal via `fetchAssociatedMscaAccount → GET /circle/wallet`), OR
- Retry the local save with backoff before surfacing any error.

## Steps to Reproduce

Hard to hit in practice — requires the DataStore write to throw. Easiest artificial reproduction is to inject a failing `ScaCircleAccountRepository.saveAccount` and run `onCreateAccount` with a successful mocked `circleApi.createScAccount`.

## Solution

Split the `runCatching` so only the network call is treated as the account-creation result. In `.onSuccess { newAddress -> ... }`, set `mscaAddress`, update state, show the success snackbar, and wrap `saveAccount` in its own `runCatching { }.onFailure { Timber.e(it) }` — a cache failure should log and move on, not revert the success.

### Must Do
- [ ] `runCatching` wraps only `circleApi.createScAccount`
- [ ] `saveAccount` runs on the success path in its own try/catch, logs on failure, does not revert `isAccountOpen`

### Must NOT Do
- Don't change the snackbar-type behaviour (already fixed in #4087)
- Don't touch the retry/idempotency behaviour (tracked separately in the linked "retry creates duplicate account" ticket)

## Acceptance Criteria
- [ ] Unit test: remote success + local save throwing → success snackbar, `isAccountOpen = true`
- [ ] Unit test: remote failure → error snackbar, `saveAccount` not called

johnnyluo:
## Summary
- Split the `runCatching` in `onCreateAccount` so only `circleApi.createScAccount(...)` determines success/failure
- `saveAccount` (DataStore write) now runs in its own `runCatching { }.onFailure { Timber.e(it) }` on the success path — a cache failure logs silently instead of showing an error snackbar and leaving `isAccountOpen = false`
- Added unit test: remote success + local save throwing → Success snackbar, `isAccountOpen = true`

## Issue
Closes #4088

## Test Plan
- [x] Unit tests pass: `CircleDeFiPositionsViewModelTest` (4 tests, all green)
- [ ] Manual: open Circle account, confirm success snackbar and account UI appears
- [ ] Manual: next launch self-heals via `fetchAssociatedMscaAccount → GET /circle/wallet` if local save was skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure account creation success is reflected in the UI and success notification even if local persistence fails
  * Improved error handling so failures during account creation reliably show the failure notification and are logged

* **Tests**
  * Added unit test covering account creation where local save throws, confirming success notification and UI state update remain consistent

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4508?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
@@ -240,28 +240,30 @@ constructor(
                 withContext(ioDispatcher) {
                     runCatching {
                         val ethereumVaultAddress = getEvmVaultAddress()
-                        circleApi.createScAccount(ethereumVaultAddress).also { newAddress ->
-                            scaCircleAccountRepository.saveAccount(vaultId, newAddress)
-                        }
+                        circleApi.createScAccount(ethereumVaultAddress)
                     }
                 }
 
             createdAddress
                 .onSuccess { newAddress ->
                     mscaAddress = newAddress
+                    _state.update { currentState ->
+                        currentState.copy(
+                            circleDefi = currentState.circleDefi.copy(isAccountOpen = true)
+                        )
+                    }
                     showSnackbar(R.string.circle_msca_account_created_success, SnackbarType.Success)
+                    runCatching {
+                            withContext(ioDispatcher) {
+                                scaCircleAccountRepository.saveAccount(vaultId, newAddress)
+                            }
+                        }
+                        .onFailure { Timber.e(it) }
                 }
                 .onFailure { t ->
                     Timber.e(t)
                     showSnackbar(R.string.circle_msca_account_created_failed, SnackbarType.Error)
                 }
-
-            _state.update { currentState ->
-                currentState.copy(
-                    circleDefi =
-                        currentState.circleDefi.copy(isAccountOpen = createdAddress.isSuccess)
-                )
-            }
         }
     }
 
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
@@ -236,34 +236,34 @@ constructor(
     /** Creates a new Circle MSCA account for the current vault. */
     fun onCreateAccount() {
         viewModelScope.launch {
-            val createdAddress =
-                withContext(ioDispatcher) {
-                    runCatching {
+            try {
+                val newAddress =
+                    withContext(ioDispatcher) {
                         val ethereumVaultAddress = getEvmVaultAddress()
                         circleApi.createScAccount(ethereumVaultAddress)
                     }
+                mscaAddress = newAddress
+                _state.update { currentState ->
+                    currentState.copy(
+                        circleDefi = currentState.circleDefi.copy(isAccountOpen = true)
+                    )
                 }
-
-            createdAddress
-                .onSuccess { newAddress ->
-                    mscaAddress = newAddress
-                    _state.update { currentState ->
-                        currentState.copy(
-                            circleDefi = currentState.circleDefi.copy(isAccountOpen = true)
-                        )
+                showSnackbar(R.string.circle_msca_account_created_success, SnackbarType.Success)
+                try {
+                    withContext(ioDispatcher) {
+                        scaCircleAccountRepository.saveAccount(vaultId, newAddress)
                     }
-                    showSnackbar(R.string.circle_msca_account_created_success, SnackbarType.Success)
-                    runCatching {
-                            withContext(ioDispatcher) {
-                                scaCircleAccountRepository.saveAccount(vaultId, newAddress)
-                            }
-                        }
-                        .onFailure { Timber.e(it) }
-                }
-                .onFailure { t ->
+                } catch (e: CancellationException) {
+                    throw e
+                } catch (t: Throwable) {
                     Timber.e(t)
-                    showSnackbar(R.string.circle_msca_account_created_failed, SnackbarType.Error)
                 }
+            } catch (e: CancellationException) {
+                throw e
+            } catch (t: Throwable) {
+                Timber.e(t)
+                showSnackbar(R.string.circle_msca_account_created_failed, SnackbarType.Error)
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
```diff
@@ -114,6 +114,21 @@ internal class CircleDeFiPositionsViewModelTest {
...
+            val vm = createViewModel().also { it.setData(VAULT_ID) }
+            givenNoExistingAccount()
+            coEvery { circleApi.createScAccount(OWNER_ADDRESS) } returns MSCA_ADDRESS
+            coEvery { scaCircleAccountRepository.saveAccount(VAULT_ID, MSCA_ADDRESS) } throws
+                RuntimeException("DataStore IO error")
+
+            val type = snackbarTypeShownBy { vm.onCreateAccount() }
+
+            assertEquals(SnackbarType.Success, type)
+            assertTrue(vm.state.value.circleDefi.isAccountOpen)
```
rkokhatskyi, line 130: This case asserts the snackbar type and `isAccountOpen` whereas it never calls `coVerify` on `scaCircleAccountRepository.saveAccount`, so a refactor that silently dropped the save call on the success path would still pass green here while the sibling tests at lines 98 and 113 do verify the save call count on this same mock

```kotlin
coVerify(exactly = 1) { scaCircleAccountRepository.saveAccount(VAULT_ID, MSCA_ADDRESS) }
```

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
```diff
@@ -240,28 +240,30 @@ constructor(
...
+                            circleDefi = currentState.circleDefi.copy(isAccountOpen = true)
+                        )
+                    }
                     showSnackbar(R.string.circle_msca_account_created_success, SnackbarType.Success)
+                    runCatching {
+                            withContext(ioDispatcher) {
+                                scaCircleAccountRepository.saveAccount(vaultId, newAddress)
+                            }
+                        }
+                        .onFailure { Timber.e(it) }
```
rkokhatskyi: Silently logging the save failure leans entirely on `fetchAssociatedMscaAccount` repopulating the cache via the Circle GET on the next `loadCirclePositions` run, so if the user closes the app before that refresh and the GET endpoint hasn't yet indexed the just-created account the address is unreachable from the UI whereas there is no retry, no backoff, and no test exercising the self-heal path the PR description leans on

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
```diff
@@ -240,28 +240,30 @@ constructor(
                 withContext(ioDispatcher) {
                     runCatching {
                         val ethereumVaultAddress = getEvmVaultAddress()
-                        circleApi.createScAccount(ethereumVaultAddress).also { newAddress ->
-                            scaCircleAccountRepository.saveAccount(vaultId, newAddress)
-                        }
+                        circleApi.createScAccount(ethereumVaultAddress)
                     }
                 }
```
rkokhatskyi, line 250: The outer block wraps `withContext(ioDispatcher) { runCatching { ... } }` while the inner save at lines 256 to 261 inverts it as `runCatching { withContext(ioDispatcher) { ... } }`, which is two different shapes for the same idea inside one function and unlike every other coroutine block in this file that uses plain `try`/`catch` with explicit `CancellationException` rethrow

rkokhatskyi (changes requested):

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModelTest.kt
johnnyluo: Fixed: added `coVerify(exactly = 1) { scaCircleAccountRepository.saveAccount(VAULT_ID, MSCA_ADDRESS) }` to the failing-local-save test so the save call is explicitly verified even when it throws.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
johnnyluo: Fixed: replaced both `runCatching` blocks with nested `try/catch` blocks that explicitly rethrow `CancellationException`, matching the pattern used in every other coroutine block in this file.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
johnnyluo: Acknowledged — the self-heal path relies on the Circle GET endpoint indexing the account before the next `loadCirclePositions` run. This is an inherent trade-off of the optimistic approach (surface account-open immediately, save locally as best-effort). A retry/backoff for the local save is out of scope for this PR but noted for a follow-up.

rkokhatskyi (approved):
LGTM

PR #4514, opened 2026-05-16
fix: guard EVM broadcast against non-2xx RPC proxy responses

### LINKED ISSUE #4409: EVM broadcast fails with 403 from api.vultisig.com/eth/ RPC proxy
# [Fix] Guard non-2xx RPC responses before JSON deserialization [Keysign / EVM Broadcast]

## Problem
Broadcasting a signed EVM transaction via the Vultisig RPC proxy (`https://api.vultisig.com/eth/`) can return HTTP 403 with no `Content-Type`. `EvmApiImp.sendTransaction` unconditionally attempts to deserialize the response body as `SendTransactionJson`, which causes a `NoTransformationFoundException` to bubble up and be displayed verbatim on the Keysign error screen.

## Expected Behavior
Non-2xx RPC responses should not trigger JSON deserialization. The app should check `response.status.isSuccess()` first, read the body as text if the response is an error, and surface a clear, user-friendly error message instead of a raw Ktor stack trace.

## Steps to Reproduce
1. Complete an Ethereum (or other EVM) keysign.
2. The final broadcast step calls `EvmApiImp.sendTransaction` against `https://api.vultisig.com/eth/`.
3. Proxy responds `403` (HTML or empty body, `Content-Type: null`).
4. App throws `NoTransformationFoundException` and the Keysign error screen shows the full stack trace.

### Must Do
- [ ] In `EvmApiImp.sendTransaction` (`EvmApi.kt:998`), check `response.status.isSuccess()` before calling `.body<SendTransactionJson>()` and throw a descriptive exception (using `bodyAsText()`) when the status is non-2xx.
- [ ] Ensure the Keysign error screen receives a clean, human-readable error string rather than a raw exception message for broadcast failures.
- [ ] Investigate whether the same pattern (unconditional JSON deserialization on any response code) exists elsewhere in `EvmApi.kt` and apply the same guard if so.

### Must NOT Do
- Don't change unrelated API methods or non-EVM broadcast paths.
- Don't alter the `SendTransactionJson` model structure.

## Acceptance Criteria
- [ ] A simulated 403 response from the RPC proxy results in a user-friendly error message on the Keysign screen — no `NoTransformationFoundException` in the stack trace.
- [ ] A successful broadcast (2xx response) continues to work correctly.

## Context
- Endpoint: `https://api.vultisig.com/eth/`
- Failing status: `403`, `Content-Type: null`
- Crash site: `EvmApi.kt` → `EvmApiImp.sendTransaction` (line ~998)
- Open question: why is the proxy returning 403? (WAF / rate limit / regional block / missing auth header?) — may require a separate server-side investigation.
- Error screenshot: Keysign error screen

```
io.ktor.client.call.NoTransformationFoundException: Expected response body of the type
'class com.vultisig.wallet.data.api.models.SendTransactionJson' but was
'class io.ktor.utils.io.SourceByteReadChannel'
In response from `https://api.vultisig.com/eth/`
Response status `403`
Response header `ContentType: null`
Request header `Accept: */*`
    at io.ktor.client.call.HttpClientCall.bodyNullable(HttpClientCall.kt:91)
    at com.vultisig.wallet.data.api.EvmApiImp.sendTransaction(EvmApi.kt:998)
```

johnnyluo:
## Summary
- In `EvmApiImp.sendTransaction`, check `response.status.isSuccess()` after reading the body as text, and throw a descriptive `Exception("EVM broadcast failed (403): ...")` for non-2xx responses — eliminating the raw `NoTransformationFoundException` that was shown verbatim on the Keysign error screen
- Applied the same guard to `findCustomToken` which had the identical unconditional-deserialization pattern

## Issue
Closes #4409

## Test Plan
- [ ] `./gradlew lint` passes
- [ ] A simulated 403 from the RPC proxy results in a clean error message on the Keysign screen (no `NoTransformationFoundException` in the trace)
- [ ] A successful 2xx broadcast continues to work correctly — the happy path is unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of custom token search with stricter response validation and clearer failure detection.
  * Enhanced transaction sending with more robust response parsing and error surfacing to reduce silent/ambiguous failures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4514?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -22,11 +22,9 @@ import com.vultisig.wallet.data.utils.NetworkException
 import com.vultisig.wallet.data.utils.Numeric
 import com.vultisig.wallet.data.utils.bodyOrThrow
 import io.ktor.client.HttpClient
-import io.ktor.client.call.body
 import io.ktor.client.request.post
 import io.ktor.client.request.setBody
 import io.ktor.client.statement.bodyAsText
-import io.ktor.http.isSuccess
 import java.math.BigInteger
 import java.net.SocketTimeoutException
 import javax.inject.Inject
@@ -382,13 +380,10 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
     override suspend fun findCustomToken(contractAddress: String): List<CustomTokenResponse> {
         val (payload1, payload2) = generateCustomTokenPayload(contractAddress)
         return try {
-            val response = http.post(rpcUrl) { setBody(listOf(payload1, payload2)) }
-            if (!response.status.isSuccess()) {
-                throw Exception(
-                    "findCustomToken failed (${response.status.value}): ${response.bodyAsText()}"
-                )
-            }
-            val responseList = response.body<List<RpcResponse>>()
+            val responseList =
+                http
+                    .post(rpcUrl) { setBody(listOf(payload1, payload2)) }
+                    .bodyOrThrow<List<RpcResponse>>()
             responseList.map { CustomTokenResponse(id = it.id, result = it.result) }
         } catch (e: Exception) {
             if (e is kotlinx.coroutines.CancellationException) throw e
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -352,13 +352,7 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
         val response = http.post(rpcUrl) { setBody(payload) }
         val responseBody = response.bodyAsText()
         Timber.d("broadcast response: $responseBody")
-        if (!response.status.isSuccess()) {
-            throw NetworkException(
-                response.status.value,
-                "EVM broadcast failed (${response.status.value})",
-            )
-        }
-        val jsonObject = response.body<SendTransactionJson>()
+        val jsonObject = response.bodyOrThrow<SendTransactionJson>()
         if (jsonObject.error != null) {
             val message = jsonObject.error.message
             if (
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -353,7 +353,10 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
         val responseBody = response.bodyAsText()
         Timber.d("broadcast response: $responseBody")
         if (!response.status.isSuccess()) {
-            throw Exception("EVM broadcast failed (${response.status.value}): $responseBody")
+            throw NetworkException(
+                response.status.value,
+                "EVM broadcast failed (${response.status.value})",
+            )
         }
         val jsonObject = response.body<SendTransactionJson>()
         if (jsonObject.error != null) {
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -26,6 +26,7 @@ import io.ktor.client.call.body
 import io.ktor.client.request.post
 import io.ktor.client.request.setBody
 import io.ktor.client.statement.bodyAsText
+import io.ktor.http.isSuccess
 import java.math.BigInteger
 import java.net.SocketTimeoutException
 import javax.inject.Inject
@@ -351,6 +352,9 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
         val response = http.post(rpcUrl) { setBody(payload) }
         val responseBody = response.bodyAsText()
         Timber.d("broadcast response: $responseBody")
+        if (!response.status.isSuccess()) {
+            throw Exception("EVM broadcast failed (${response.status.value}): $responseBody")
+        }
         val jsonObject = response.body<SendTransactionJson>()
         if (jsonObject.error != null) {
             val message = jsonObject.error.message
@@ -382,6 +386,11 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
         val (payload1, payload2) = generateCustomTokenPayload(contractAddress)
         return try {
             val response = http.post(rpcUrl) { setBody(listOf(payload1, payload2)) }
+            if (!response.status.isSuccess()) {
+                throw Exception(
+                    "findCustomToken failed (${response.status.value}): ${response.bodyAsText()}"
+                )
+            }
             val responseList = response.body<List<RpcResponse>>()
             responseList.map { CustomTokenResponse(id = it.id, result = it.result) }
         } catch (e: Exception) {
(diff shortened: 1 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -351,6 +352,9 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
         val response = http.post(rpcUrl) { setBody(payload) }
         val responseBody = response.bodyAsText()
         Timber.d("broadcast response: $responseBody")
+        if (!response.status.isSuccess()) {
+            throw Exception("EVM broadcast failed (${response.status.value}): $responseBody")
```
rkokhatskyi: `bodyOrThrow<T>()` in `data/src/main/kotlin/com/vultisig/wallet/data/utils/RpcExtensions.kt:33` already covers this exact case which CLAUDE.md and `docs/network-error-handling.md` flag as the preferred pattern over raw `.body<T>()`, so this whole block can collapse to `response.bodyOrThrow<SendTransactionJson>()` which additionally throws a typed `NetworkException(status.value, body)` letting upstream `safeLaunch(onError)` distinguish 4xx from 5xx without having to parse a bare `Exception` message

rkokhatskyi: `KeysignViewModel.kt:501` renders `e.message or R.string.unknown_error` directly as the error screen text, so splicing `responseBody` into the thrown message means a 403 WAF or proxy page that may be HTML or several KB of text lands verbatim on the Keysign screen, which is essentially the same readability problem the issue wants to fix, so the message should be a localized fixed string while the raw body stays in `Timber.d` only

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -382,6 +386,11 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
         val (payload1, payload2) = generateCustomTokenPayload(contractAddress)
         return try {
             val response = http.post(rpcUrl) { setBody(listOf(payload1, payload2)) }
+            if (!response.status.isSuccess()) {
+                throw Exception(
+                    "findCustomToken failed (${response.status.value}): ${response.bodyAsText()}"
```
rkokhatskyi: This branch is unreachable as visible behavior since the surrounding `try / catch (Exception)` two lines below already converts `NoTransformationFoundException` and any other throwable into `emptyList()`, so the only delta from this guard is the Timber log content whereas the PR description frames it as parity with the `sendTransaction` fix

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -382,6 +386,11 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
...
+                )
+            }
```
rkokhatskyi: Issue 4409 also asks to check whether this unconditional deserialization pattern exists elsewhere in `EvmApi.kt`, and `getBalances` at line 502 routes through `postRpc` in `data/src/main/kotlin/com/vultisig/wallet/data/api/utils/Rpc.kt:21` which does `post(url) {...}.body()` with no status guard, so an `alchemy_getTokenBalances` 403 from the same `api.vultisig.com` proxy reproduces the exact `NoTransformationFoundException` this PR aims to eliminate

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
johnnyluo: Addressed. See latest commit.

johnnyluo: Addressed. See latest commit.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
johnnyluo: Addressed. See latest commit.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
johnnyluo: Addressed. See latest commit.

aminsato:
@coderabbitai full review

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -351,6 +352,12 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
         val response = http.post(rpcUrl) { setBody(payload) }
         val responseBody = response.bodyAsText()
         Timber.d("broadcast response: $responseBody")
+        if (!response.status.isSuccess()) {
+            throw NetworkException(
+                response.status.value,
+                "EVM broadcast failed (${response.status.value})",
+            )
+        }
```
aminsato: This manual `isSuccess()` + raw `body<T>()` pair re-implements what `HttpResponse.bodyOrThrow<T>()` in `data/utils/RpcExtensions.kt` already does — it throws `NetworkException(status.value, bodyAsText())` on non-2xx and wraps deserialization failures on 2xx in the same exception type. The other API classes already follow that pattern, so swapping the two lines for:

```kotlin
val jsonObject = response.bodyOrThrow<SendTransactionJson>()
```

…would drop the guard block entirely and keep this API surface consistent.

Minor: since `responseBody` is already read into a local on line 353, you'd also avoid a second `bodyAsText()` round-trip that `bodyOrThrow` performs on the failure path. If you keep the manual guard, consider including `responseBody` in the message (it's already logged via Timber, but surfacing it in the exception matches what `findCustomToken` does below).

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -382,6 +389,11 @@ class EvmApiImp(private val http: HttpClient, private val rpcUrl: String) : EvmA
         val (payload1, payload2) = generateCustomTokenPayload(contractAddress)
         return try {
             val response = http.post(rpcUrl) { setBody(listOf(payload1, payload2)) }
+            if (!response.status.isSuccess()) {
+                throw Exception(
```
aminsato: Two things here:

1. **Inconsistent exception type** — the sibling change in `sendTransaction` throws `NetworkException(status, …)`, but this block throws a raw `Exception(…)`. Worth aligning so downstream `catch` blocks can rely on the type.

2. **Same `bodyOrThrow` suggestion applies** — the surrounding `catch (e: Exception)` already swallows whatever is thrown and returns `emptyList()`, so collapsing this to:

   ```kotlin
   val responseList = response.bodyOrThrow<List<RpcResponse>>()
   ```

   …gives the same behavior, drops the manual guard, and resolves the inconsistency at the same time.

aminsato (commented):
Nice fix for the root cause. Two suggestions inline around reusing the existing `bodyOrThrow()` helper for consistency with the rest of the data layer.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
johnnyluo: Noted — the `postRpc` / `getBalances` path has the same pattern, but that change is out of scope for this PR. Will track separately.

rkokhatskyi (approved):
LGTM

PR #4525, opened 2026-05-16
feat: surface dApp request banner on verify and done screens (#4060)

### LINKED ISSUE #4060: feat: Rich dApp context — name, URL, icon (Phase 3)
## What to build
Display dApp identity propagated from the extension via new protobuf field:

\`\`\`protobuf
message DAppMetadata {
  string name = 1;
  string url = 2;
  string icon_url = 3;
}
\`\`\`

Show "Request from: Uniswap (app.uniswap.org)" on verify + done screens. Treat as informational — the independently decoded calldata is the security source of truth.

**Companion issue:** vultisig/vultisig-windows#3739
**PRD:** sections 3.1–3.3
**Breaking:** No — additive protobuf field, old apps ignore it

aminsato:
## Summary
- Phase 3 of dApp signing UX (#4060). Surfaces the requesting dApp's name, URL, and icon above the hero on **Verify** (Send/Swap/Deposit) and **Done** (Send/Swap) screens for keysigns originated by external dApps
- Mirrors iOS [#4303](https://github.com/vultisig/vultisig-ios/pull/4303) and Windows [#3819](https://github.com/vultisig/vultisig-windows/pull/3819). Trust decisions stay with Blockaid + decoded calldata; the banner is informational only
- Bumps `commondata` from `6d61f4de` → `798af15` to pick up [commondata#82](https://github.com/vultisig/commondata/pull/82) (DAppMetadata proto). Also pulls in commondata#75 (Cardano token asset proto)

## What's wired

| Surface | Source | Screen |
|---|---|---|
| Send Verify (joiner) | `JoinKeysignViewModel.dappMetadata` | `VerifySendScreen` |
| Swap Verify (joiner) | `JoinKeysignViewModel.dappMetadata` | `VerifySwapScreen` |
| Deposit Verify (joiner) | `JoinKeysignViewModel.dappMetadata` | `VerifyDepositScreen` |
| Send Done (initiator + joiner) | `KeysignViewModel.dappMetadata` | `SendTxOverviewScreen` → `TxDoneScaffold` |
| Swap Done (initiator + joiner) | `KeysignViewModel.dappMetadata` | `SwapTransactionOverviewScreen` → `TxDoneScaffold` |

Vault-initiated keysigns leave the field `null`, so no banner appears on those flows.

## Not wired (intentional)
- **SignMessage variants** — `commondata` only carries `dapp_metadata` on `KeysignPayload`, not on `KeysignMessage` where `CustomMessagePayload` lives. Needs a separate proto change.

## Security hardening
- `DAppMetadata.host` returns empty for non-http(s) schemes so `javascript:`/`data:`/`file:`/`mailto:` URLs never display as if they were hostnames
- `DAppMetadata.safeIconUrl` returns `null` for non-https schemes so Coil can't be tricked into loading `file:///`, `content://`, or `data:` images
- 12 unit tests cover the defensive parsing semantics

## Test plan
- [x] \`./gradlew :data:ktfmtFormat :app:ktfmtFormat\` — clean
- [x] \`./gradlew :app:compileDebugKotlin\` — green
- [x] \`./gradlew :data:testDebugUnitTest :app:testDebugUnitTest\` — 18 new tests pass, no regressions
- [ ] Manual: scan a TonConnect/WalletConnect QR with \`dapp_metadata\` populated → banner shows on Verify and Done
- [ ] Manual: scan a legacy QR without \`dapp_metadata\` → no banner, existing screens unchanged

## Companion PRs
- iOS: vultisig/vultisig-ios#4303 (merged 2026-05-05)
- Windows: vultisig/vultisig-windows#3819 (merged 2026-05-02)

Closes #4060

## Summary by CodeRabbit

* **New Features**
  * Adds a dApp request banner shown on verify and transaction-done screens with name, host and icon variants to improve transparency during signing flows.
  * Propagates dApp metadata into relevant transaction and keysign screens so the banner appears where applicable.

* **Tests**
  * Added unit tests covering dApp metadata parsing, trimming, host extraction, emptiness semantics, and round-trip proto mapping.

* **Chores**
  * Added localized "Request from" strings for multiple languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4525?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
@@ -15,8 +15,6 @@ import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.draw.clip
-import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.contentDescription
@@ -24,7 +22,6 @@ import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
-import coil.compose.SubcomposeAsyncImage
 import com.vultisig.wallet.R
 import com.vultisig.wallet.data.models.payload.DAppMetadata
 import com.vultisig.wallet.ui.theme.Theme
@@ -37,6 +34,13 @@ import com.vultisig.wallet.ui.theme.Theme
  * Layout mirrors the iOS `DAppRequestBanner`: a "Request from" header followed by a 32dp circular
  * icon and a name/host stack. Empty fields are treated as absent so partially-populated metadata
  * still renders cleanly.
+ *
+ * **Privacy stance — Android intentionally does NOT fetch the dApp-supplied `iconUrl`.** Loading
+ * the favicon would expose the signing device IP and TLS fingerprint to a host the user has not yet
+ * consented to interact with, and feed a default Coil pipeline an attacker-controlled URL with no
+ * size cap or timeout (decoder-CVE surface). iOS and Windows currently load the icon — Android
+ * leads on this and ships a placeholder glyph instead. The name and host already identify the dApp;
+ * the icon was visual polish, not load-bearing.
  */
 @Composable
 internal fun DappRequestBanner(metadata: DAppMetadata, modifier: Modifier = Modifier) {
@@ -46,7 +50,11 @@ internal fun DappRequestBanner(metadata: DAppMetadata, modifier: Modifier = Modi
     val announcement = buildString {
         append(label)
         if (metadata.name.isNotEmpty()) append(' ').append(metadata.name)
-        if (metadata.host.isNotEmpty()) append(", ").append(metadata.host)
+        if (metadata.host.isNotEmpty()) {
+            // Use a comma to separate name from host, but a plain space when name is absent so
+            // TalkBack reads "Request from app.uniswap.org" instead of "Request from, app…".
+            append(if (metadata.name.isNotEmpty()) ", " else " ").append(metadata.host)
+        }
     }
     Column(
         verticalArrangement = Arrangement.spacedBy(16.dp),
@@ -70,43 +78,28 @@ internal fun DappRequestBanner(metadata: DAppMetadata, modifier: Modifier = Modi
             horizontalArrangement = Arrangement.spacedBy(12.dp),
             verticalAlignment = Alignment.CenterVertically,
         ) {
-            DappIcon(safeIconUrl = metadata.safeIconUrl)
+            DappIcon()
             DappInfo(name = metadata.name, host = metadata.host, modifier = Modifier.weight(1f))
         }
     }
 }
 
 @Composable
-private fun DappIcon(safeIconUrl: String?) {
-    val placeholder: @Composable () -> Unit = {
-        Box(
-            contentAlignment = Alignment.Center,
-            modifier =
-                Modifier.size(ICON_SIZE)
-                    .background(Theme.v2.colors.backgrounds.surface1, CircleShape),
-        ) {
-            Icon(
-                painter = painterResource(R.drawable.settings_globe),
-                contentDescription = null,
-                tint = Theme.v2.colors.text.tertiary,
-                modifier = Modifier.size(PLACEHOLDER_GLYPH_SIZE),
-            )
-        }
+private fun DappIcon() {
+    // Intentionally no `iconUrl` parameter. See class kdoc — Android does not fetch dApp-supplied
+    // icons so a hostile origin can't fingerprint the signing device the moment verify renders.
+    Box(
+        contentAlignment = Alignment.Center,
+        modifier =
+            Modifier.size(ICON_SIZE).background(Theme.v2.colors.backgrounds.surface1, CircleShape),
+    ) {
+        Icon(
+            painter = painterResource(R.drawable.settings_globe),
+            contentDescription = null,
+            tint = Theme.v2.colors.text.tertiary,
+            modifier = Modifier.size(PLACEHOLDER_GLYPH_SIZE),
+        )
     }
-
-    if (safeIconUrl == null) {
-        placeholder()
-        return
-    }
-
-    SubcomposeAsyncImage(
-        model = safeIconUrl,
-        contentDescription = null,
-        contentScale = ContentScale.Crop,
-        modifier = Modifier.size(ICON_SIZE).clip(CircleShape),
-        loading = { placeholder() },
-        error = { placeholder() },
-    )
 }
 
 @Composable
--- a/app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
@@ -1,0 +1,168 @@
+package com.vultisig.wallet.ui.components.dapp
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Icon
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import coil.compose.SubcomposeAsyncImage
+import com.vultisig.wallet.R
+import com.vultisig.wallet.data.models.payload.DAppMetadata
+import com.vultisig.wallet.ui.theme.Theme
+
+/**
+ * Informational card showing which dApp produced a keysign request. Rendered above the transaction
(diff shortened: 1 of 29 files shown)


rkokhatskyi:
@aminsato please attach the screenshots to this PR

aminsato:
> @aminsato please attach the screenshots to this PR
image

telegram-cloud-photo-size-4-5782742241217547795-y

aminsato:
@coderabbitai full reveiw

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -264,6 +265,22 @@ constructor(
...
+     * **Compose observability:** [_keysignPayload] is a plain `var`, so this getter is *not*
+     * observable on its own. The verify screens that read it also `collectAsState()` on
+     * [verifyUiModel]; because [loadTransaction] sets `_keysignPayload` before the matching
+     * `verifyUiModel.value = …` assignment, the state-flow emission triggers the recomposition that
+     * re-reads this getter and picks up the populated value. If a future change emits to
+     * `verifyUiModel` *before* `_keysignPayload` is populated, the banner will read stale `null`
+     * until the next unrelated recomposition.
+     */
+    val dappMetadata: DAppMetadata?
+        get() = _keysignPayload?.dappMetadata
```
rkokhatskyi: Exposing this as a getter over the plain `var _keysignPayload` makes banner visibility depend on `_keysignPayload` being assigned before any `verifyUiModel.value =` emission, which the KDoc itself flags as a trap that future refactors can silently break, so folding `dappMetadata` into each `VerifyUiModel` variant or surfacing it as its own `StateFlow<DAppMetadata?>` would remove the brittle implicit ordering

app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
```diff
@@ -0,0 +1,168 @@
...
+            )
+        }
+    }
+
+    if (safeIconUrl == null) {
+        placeholder()
+        return
+    }
+
+    SubcomposeAsyncImage(
```
rkokhatskyi: `SubcomposeAsyncImage` fires an HTTPS request to the icon URL controlled by the dApp the moment the verify screen renders, which exposes the signing device IP and TLS fingerprint to that origin before the user has consented to anything, whereas previously only the initiator browser ever contacted that host

app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
```diff
@@ -0,0 +1,168 @@
...
+        model = safeIconUrl,
```
rkokhatskyi: This hands `safeIconUrl` to the default Coil `ImageLoader` with no `size()`, redirect cap, or timeout override, so a hostile dApp can return a payload of arbitrary size or a malformed image targeting a decoder CVE that runs inline on the verify flow, which capping the request at `Size(96, 96)` plus `allowHardware(false)` and routing through an OkHttp client with a short read timeout would mitigate

data/src/main/kotlin/com/vultisig/wallet/data/models/payload/DAppMetadata.kt
```diff
@@ -0,0 +1,56 @@
...
+            }
+            .getOrDefault("")
+    }
+
+    /**
+     * True when every field is empty. Use at the construction boundary (proto mapper) to avoid
+     * surfacing meaningless banners.
+     */
+    val isEmpty: Boolean
+        get() = name.isEmpty() && url.isEmpty() && iconUrl.isEmpty()
```
rkokhatskyi: `isEmpty` returns false when only `iconUrl` is set, so a metadata of `name="", url="", iconUrl="https://attacker.io/x"` still renders a banner with just a circle and an empty text stack while firing the network fetch, which suggests requiring at least one of `name` or `url` to be populated for the gate to pass

data/src/main/kotlin/com/vultisig/wallet/data/models/payload/DAppMetadata.kt
```diff
@@ -0,0 +1,56 @@
...
+     * Hostname extracted from [url] when [url] is a well-formed http(s) URI. Returns `""` for empty
+     * input, malformed input, or non-http(s) schemes (`javascript:`, `data:`, `file:`, …) so the UI
+     * never echoes attacker-controlled text as if it were a hostname. The banner hides the host
+     * line when this is empty.
+     */
+    val host: String by lazy {
+        runCatching {
+                if (url.isEmpty()) return@runCatching ""
+                val uri = URI(url)
+                if (uri.scheme?.lowercase() in HTTP_SCHEMES) uri.host.orEmpty() else ""
```
rkokhatskyi, line 36: iOS and the Windows extension both fall back to the raw `url` when host parsing fails, while this returns the empty string, which is safer against hostile `javascript:` or `data:` URLs but also silently hides any legitimate scheme that still resolves to a host like an onion service, so worth either documenting the divergence in the KDoc or matching the other platforms for parse failures while keeping the strict scheme check

data/src/main/kotlin/com/vultisig/wallet/data/mappers/KeysignPayloadProtoMapper.kt
```diff
@@ -238,6 +239,18 @@ internal class KeysignPayloadProtoMapperImpl @Inject constructor() : KeysignPayl
             tronTransferContractPayload = from.tronTransferContractPayload,
             tronTransferAssetContractPayload = from.tronTransferAssetContractPayload,
             tronTriggerSmartContractPayload = from.tronTriggerSmartContractPayload,
+            // Trim whitespace at the boundary so consumers (host derivation, isEmpty gate, UI)
+            // don't have to re-normalize. Empty after trim ⇒ treat the whole banner as absent.
+            dappMetadata =
```
rkokhatskyi, line 244: The inbound mapper reads `from.dappMetadata` here but the outbound counterpart `PayloadToProtoMapper` does not write `keysignPayload.dappMetadata` back into the proto, so a flow where Android initiates and relays a payload to joining peers strips the field, which is dormant today since Android does not yet initiate from a dApp request but breaks parity with iOS `mapToProtobuff` and Windows `buildDappMetadata` the moment that changes

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
```diff
@@ -0,0 +1,168 @@
...
+@Composable
+internal fun DappRequestBanner(metadata: DAppMetadata, modifier: Modifier = Modifier) {
+    val label = stringResource(R.string.request_from)
+    // Merge the three texts into one TalkBack announcement: "Request from Uniswap, app.uniswap.org"
+    // instead of three separate nodes.
+    val announcement = buildString {
+        append(label)
+        if (metadata.name.isNotEmpty()) append(' ').append(metadata.name)
+        if (metadata.host.isNotEmpty()) append(", ").append(metadata.host)
+    }
```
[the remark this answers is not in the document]
aminsato: Fixed in b756eda93 —  now picks  when name was appended, plain space otherwise. TalkBack now reads "Request from app.uniswap.org" cleanly when name is absent.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
aminsato: Agreed — refactored in b756eda93. `_keysignPayload` now has a custom setter that mirrors the dApp metadata into a private `MutableStateFlow<DAppMetadata?>`, and `dappMetadata` is exposed as a `StateFlow` consumed via `collectAsState()` in `JoinKeysignView`. No more dependency on emission ordering — the banner updates the moment the payload is parsed.

app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
aminsato: Fixed in b756eda93 by dropping the icon fetch entirely. The banner now always shows the globe placeholder; name + host already identify the dApp. Diverges from iOS/Windows but the kdoc spells out the reasoning. Happy to revisit with a privacy-controlled `ImageLoader` (no-redirect, size cap, custom UA) gated behind a Privacy setting in a follow-up if you'd rather keep the favicon — but the IP leak can't be mitigated, only avoided.

app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
aminsato: Resolved by b756eda93 — since the icon fetch is removed, the unbounded-Coil-pipeline / decoder-CVE surface goes with it. The dead `SubcomposeAsyncImage` import and the `safeIconUrl` helper that was no longer reachable are both deleted.

data/src/main/kotlin/com/vultisig/wallet/data/models/payload/DAppMetadata.kt
aminsato: Fixed in b756eda93 — `isEmpty` now requires `name` or `url` to be set; `iconUrl` alone no longer passes the gate. Kdoc explains why (would render an empty text stack and, before the icon-fetch removal, would have let a hostile dApp force a network fetch via icon-only metadata). New test `isEmpty is true when only iconUrl is set …` pins the behavior.

data/src/main/kotlin/com/vultisig/wallet/data/models/payload/DAppMetadata.kt
aminsato: Documented in b756eda93 — kept the strict scheme check (security wins out over the onion-service edge case) and expanded the `host` kdoc to spell out the deliberate divergence from iOS/Windows raw-url fallback and how to narrow the empty-return if the legit-but-unparseable case ever becomes a real concern.

data/src/main/kotlin/com/vultisig/wallet/data/mappers/KeysignPayloadProtoMapper.kt
aminsato: Fixed in b756eda93 — `PayloadToProtoMapperImpl` now writes `dappMetadata` back out, mirroring the inbound mapper. Added a round-trip test (`dappMetadata round-trips through outbound proto mapper for relay parity with iOS and Windows`) so the parity stays enforced.

app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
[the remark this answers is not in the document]
aminsato: (Correcting prior reply — backticks ate the code refs.) Fixed in b756eda93 — `buildString` now picks `, ` when name was appended, plain space otherwise. TalkBack reads "Request from app.uniswap.org" cleanly when name is absent.

rkokhatskyi (approved):
LGTM

PR #4546, opened 2026-05-17
refactor(keysign): consolidate join keysign fee into single helper (#4391)

### LINKED ISSUE #4391: Consolidate join keysign fee computation into a single polymorphic helper
## Context

Follow-up from PR #4381 (review thread: https://github.com/vultisig/vultisig-android/pull/4381#discussion_r3178602598).

In `JoinKeysignViewModel.kt`, the per-`BlockChainSpecific` fee formula is currently spread across three sites:

1. The swap branch's Ethereum case at `JoinKeysignViewModel.kt:515-521` (`maxFeePerGasWei * gasLimit`).
2. The swap branch's THORChain case at `JoinKeysignViewModel.kt:522-523` (`blockChainSpecific.fee`).
3. The shared helper `computeJoinKeysignNetworkFee` at `JoinKeysignViewModel.kt:1413-1427`, used from the deposit branch (`:886`) and the send branch (`:972`).

iOS exposes the equivalent logic as a single polymorphic `fee` getter on `BlockChainSpecific.swift:75-86` covering every variant in one place.

## Problem

Any future fee-formula change (new chain variant, EIP-1559 tweak, THORChain fee model change, etc.) has to be repeated across the three Android sites, and the swap branch can drift from the deposit/send branch — exactly the class of bug PR #4381 fixed for the send path.

## Proposal

Consolidate the fee computation into a single polymorphic helper (e.g. extend `computeJoinKeysignNetworkFee` or expose a `BlockChainSpecific.fee(nativeCoin, fallback)` extension) and have the swap branch call it instead of inlining the Ethereum / THORChain cases. This brings Android in line with the iOS `BlockChainSpecific.fee` shape.

## Scope

Strictly one ticket — no sub-tickets — covering:
- Refactor the swap branch in `JoinKeysignViewModel.kt:515-523` to use the shared helper.
- Cover the refactor with unit tests next to the existing `JoinKeysignSendGasFeeTest` cases (Ethereum + THORChain swap-branch cases).

## References

- PR: https://github.com/vultisig/vultisig-android/pull/4381
- Unresolved review thread: https://github.com/vultisig/vultisig-android/pull/4381#discussion_r3178602598
- iOS reference: `BlockChainSpecific.swift:75-86`

aminsato:
## Summary

- Extend `computeJoinKeysignNetworkFee` with an optional `evmGasLimitOverride: BigInteger? = null` parameter. Default is null → existing `blockChainSpecific.gasLimit` semantics preserved for the deposit (line 927) and send (line 1015) call sites.
- Collapse the swap branch's inlined Ethereum + THORChain cases (previously 9 lines) into a single helper call that passes `defaultEvmSwapGasLimit(chain)` as the override, so joiner output continues to match initiator output for swaps (the reason the swap branch overrode `gasLimit` in the first place — see comment near `defaultEvmSwapGasLimit`).
- THORChain branch ignores the override and still returns `blockChainSpecific.fee` unchanged.

Closes #4391. Follow-up from review thread on PR #4381: https://github.com/vultisig/vultisig-android/pull/4381#discussion_r3178602598

iOS exposes the equivalent shape as a single polymorphic `fee` getter on `BlockChainSpecific.swift:75-86`; this brings Android in line.

## Test plan

- [x] `./gradlew :app:testDebugUnitTest --tests "JoinKeysignSendGasFeeTest"` — all 6 tests pass (4 existing + 2 new)
- [x] New tests: `evm swap branch uses gasLimit override`, `thorchain swap branch ignores gasLimit override`
- [x] ktfmt applied
- [ ] Manual smoke: join a swap keysign on EVM (e.g. ETH→USDC) and verify Network Fee row on the joiner matches the initiator's display
- [ ] Manual smoke: join a swap keysign on THORChain (e.g. BTC.BTC→ETH.ETH) and verify the joiner fee row matches the initiator's

## Summary by CodeRabbit

* **Bug Fixes**
  * Swap fee estimation updated: Ethereum swaps use a swap-specific gas-limit override; THORChain swaps use native chain fee values. Unsupported chains in swap paths now raise an explicit error to prevent silent zero-fee behavior. Mantle uses its chain-specific default gas limit.

* **Tests**
  * Added tests verifying Ethereum swap gas-limit override, Mantle default selection, THORChain fee handling, and erroring on unsupported swap chains.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4546?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
@@ -1470,29 +1470,28 @@ constructor(
     }
 
     private fun waitForKeysignToStart() {
-        _jobWaitingForKeysignStart =
-            viewModelScope.launch {
-                withContext(Dispatchers.IO) {
-                    while (isActive) {
-                        try {
-                            if (checkKeygenStarted()) {
-                                currentState.value = JoinKeysignState.Keysign
-                                return@withContext
-                            }
-                        } catch (e: KeysignMessagesException) {
-                            Timber.e(e, "Failed to prepare messages to sign")
-                            currentState.value =
-                                JoinKeysignState.Error(
-                                    JoinKeysignError.FailedToCheck(
-                                        e.message ?: "Failed to prepare messages to sign"
-                                    )
-                                )
+        _jobWaitingForKeysignStart = viewModelScope.launch {
+            withContext(Dispatchers.IO) {
+                while (isActive) {
+                    try {
+                        if (checkKeygenStarted()) {
+                            currentState.value = JoinKeysignState.Keysign
                             return@withContext
                         }
-                        delay(1000)
+                    } catch (e: KeysignMessagesException) {
+                        Timber.e(e, "Failed to prepare messages to sign")
+                        currentState.value =
+                            JoinKeysignState.Error(
+                                JoinKeysignError.FailedToCheck(
+                                    e.message ?: "Failed to prepare messages to sign"
+                                )
+                            )
+                        return@withContext
                     }
+                    delay(1000)
                 }
             }
+        }
     }
 
     private suspend fun checkKeygenStarted(): Boolean {
@@ -1769,6 +1768,16 @@ internal fun prettifyEvmFunctionName(signature: String): String? {
     }
 }
 
+/**
+ * Polymorphic fee helper shared by the join-keysign deposit, send, and swap branches — mirrors
+ * iOS's `BlockChainSpecific.fee` getter. Ethereum returns `maxFeePerGasWei * gasLimit`, THORChain
+ * returns `blockChainSpecific.fee`, and every other chain returns [fallbackFeeAmount] (the swap
+ * branch passes ZERO since it never enters the else branch).
+ *
+ * @param evmGasLimitOverride when non-null, replaces `blockChainSpecific.gasLimit` for the Ethereum
+ *   case. The swap branch passes [defaultEvmSwapGasLimit] so joiner output matches the initiator's
+ *   swap-fee display; deposit and send leave it null to preserve the payload gas limit.
+ */
 internal fun computeJoinKeysignNetworkFee(
     blockChainSpecific: BlockChainSpecific,
     nativeCoin: Coin,
@@ -1788,10 +1797,14 @@ internal fun computeJoinKeysignNetworkFee(
         else -> TokenValue(value = fallbackFeeAmount, token = nativeCoin)
     }
 
-// The initiator's swap path always displays maxFeePerGas * DEFAULT_SWAP_LIMIT (via
-// EthereumFeeService.calculateDefaultFees(Swap)), ignoring BlockChainSpecific.gasLimit —
-// which can be as low as 40k for native ETH/Arb swaps. Mirror that here so joiner output
-// matches initiator output instead of being ~15× lower.
+/**
+ * The initiator's swap path always displays `maxFeePerGas * DEFAULT_SWAP_LIMIT` (via
+ * [EthereumFeeService.calculateDefaultFees] for Swap), ignoring `BlockChainSpecific.gasLimit` —
+ * which can be as low as 40k for native ETH/Arb swaps. Mirror that here so joiner output matches
+ * initiator output instead of being ~15× lower. Mantle uses
+ * [EthereumFeeService.DEFAULT_MANTLE_SWAP_LIMIT] because its per-gas limit is an order of magnitude
+ * higher than other EVM chains.
+ */
 internal fun defaultEvmSwapGasLimit(chain: Chain): BigInteger =
     if (chain == Chain.Mantle) EthereumFeeService.DEFAULT_MANTLE_SWAP_LIMIT
     else EthereumFeeService.DEFAULT_SWAP_LIMIT
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
@@ -564,15 +564,14 @@ constructor(
                             }
                             TokenValue(value = BigInteger.valueOf(plan.fee), token = nativeToken)
                         }
-                        blockChainSpecific is BlockChainSpecific.Ethereum ->
-                            TokenValue(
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1770,26 +1768,43 @@ internal fun prettifyEvmFunctionName(signature: String): String? {
...
             TokenValue(
-                value = blockChainSpecific.maxFeePerGasWei * blockChainSpecific.gasLimit,
+                value =
+                    blockChainSpecific.maxFeePerGasWei *
+                        (evmGasLimitOverride ?: blockChainSpecific.gasLimit),
                 token = nativeCoin,
             )
         is BlockChainSpecific.THORChain ->
             TokenValue(value = blockChainSpecific.fee, token = nativeCoin)
         else -> TokenValue(value = fallbackFeeAmount, token = nativeCoin)
```
rkokhatskyi, line 1793: The `else` branch silently returns `fallbackFeeAmount` which the swap caller hardcodes to `BigInteger.ZERO` at line 572, so if anyone later adds a third subtype to the `||` chain at line 568, say `BlockChainSpecific.Cosmos`, the helper will quietly ship a fee of zero without any compile error since the invariant that prevents this lives over a thousand lines away. Returning `TokenValue?` and letting callers do `?: fallback`, or splitting into a swap variant that requires inputs to be `Ethereum` or `THORChain`, would let the compiler enforce the contract that the KDoc currently asserts informally

app/src/test/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignSendGasFeeTest.kt
```diff
@@ -140,4 +154,95 @@ internal class JoinKeysignSendGasFeeTest {
...
     }
+
+    /**
+     * Swap branch: Ethereum case must use the override (DEFAULT_SWAP_LIMIT) instead of the
+     * payload's gasLimit, so joiner output matches initiator output.
+     */
+    @Test
+    fun `evm swap branch uses gasLimit override`() {
+        val payloadGasLimit = BigInteger.valueOf(21_000)
+        val swapGasLimitOverride = BigInteger.valueOf(600_000)
```
rkokhatskyi: The hardcoded `600_000` drifts from `EthereumFeeService.DEFAULT_SWAP_LIMIT` silently if that constant ever moves whereas the Mantle test at line 220 already references `EthereumFeeService.DEFAULT_MANTLE_SWAP_LIMIT` directly, so this should reference the same constant for consistency and to keep the assertion honest
```suggestion
        val swapGasLimitOverride = EthereumFeeService.DEFAULT_SWAP_LIMIT
```

app/src/test/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignSendGasFeeTest.kt
```diff
@@ -140,4 +154,95 @@ internal class JoinKeysignSendGasFeeTest {
...
+     */
+    @Test
+    fun `defaultEvmSwapGasLimit selects Mantle limit for Mantle and default elsewhere`() {
+        defaultEvmSwapGasLimit(Chain.Mantle) shouldBe EthereumFeeService.DEFAULT_MANTLE_SWAP_LIMIT
+        defaultEvmSwapGasLimit(Chain.Ethereum) shouldBe EthereumFeeService.DEFAULT_SWAP_LIMIT
+    }
+
+    /** Swap branch: Mantle uses DEFAULT_MANTLE_SWAP_LIMIT when fed through the helper. */
+    @Test
+    fun `evm swap branch on Mantle applies Mantle gasLimit`() {
```
rkokhatskyi: This test never enters the swap branch in `loadTransaction` since it invokes `computeJoinKeysignNetworkFee` directly with the override already resolved, so the name overstates the coverage while the assertion is also largely redundant given that `defaultEvmSwapGasLimit selects Mantle limit` at line 192 already covers the chain dispatch and `evm swap branch uses gasLimit override` at line 163 already covers the override application. Worth either dropping it or rewriting to drive `JoinKeysignViewModel.loadTransaction` end to end so a regression that omits `evmGasLimitOverride = defaultEvmSwapGasLimit(chain)` at `JoinKeysignViewModel.kt:573` would actually fail a test, since right now no test catches a caller side wiring mistake

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
aminsato: Good catch — went with the split-variant option. New `computeJoinKeysignSwapNetworkFee` only handles `Ethereum`/`THORChain` and `error()`s on anything else, so a future addition to the swap branch's `||` check crashes loudly instead of shipping a zero fee. Dropped the now-unused `evmGasLimitOverride` param on the general helper too. 0770287fc

app/src/test/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignSendGasFeeTest.kt
aminsato: Done — the test now calls `computeJoinKeysignSwapNetworkFee` with `Chain.Ethereum` and asserts against `EthereumFeeService.DEFAULT_SWAP_LIMIT`. 0770287fc

app/src/test/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignSendGasFeeTest.kt
aminsato: Dropped it — with the new swap helper baking in `defaultEvmSwapGasLimit(chain)`, there's no separate `evmGasLimitOverride` left to wire up at the call site, so a regression is now a compile error. `defaultEvmSwapGasLimit selects Mantle limit` still covers the chain dispatch, and I added a `swap helper throws for unsupported subtype` test to lock in the new safety net. 0770287fc

aminsato:
@coderabbitai full review

rkokhatskyi (approved):
LGTM

PR #4549, opened 2026-05-17
fix(keysign): remove double V2Scaffold padding from join verify screens

### LINKED ISSUE #4538: Join Keysign verify screen has doubled container padding vs initiator's Send overview
## Summary

The keysign verify screen renders differently on the **initiator** vs the **joining (companion) device**:

- The initiator's `Send overview` screen renders with the standard layout (single scaffold, default container paddings).
- The joining device's verify screen (button labelled `Join Keysign`) has visibly **extra outer padding** around the card — both horizontal and vertical — and the card no longer matches the position used by the initiator.

The two screens are supposed to share the same `VerifySendScreen` composable, so the spacing should be identical.

## Screenshots

| Initiator (Send overview) — correct | Joined device — extra padding |
|---|---|
| initiator | joined |

Observations on the joined screenshot:
- Larger horizontal inset on the card (more space on the left/right than the initiator).
- Larger vertical gap above the card and below the card (the card is pushed further down and the `Join Keysign` button is pushed further up).
- Title also reads `Sign Transaction` instead of `Send overview` in the captured frame.

## Root cause

The joined-device path wraps `VerifySendScreen` inside a second `V2Scaffold` that already applies the default paddings, while `VerifySendScreen` provides its own `V2Scaffold`. The same default paddings (`PADDING_HORIZONTAL = 16.dp`, `PADDING_VERTICAL = 12.dp` — defined in `V2Scaffold.kt:145-146`) end up being applied **twice**.

**Initiator path** (single scaffold — looks correct):

```kotlin
// app/.../ui/navigation/NavGraph.kt:323
composable<VerifySend> { VerifySendScreen() }
```

`VerifySendScreen` provides its own `V2Scaffold` in `app/.../ui/screens/send/VerifySendScreen.kt:136-185`, which adds the standard 16dp horizontal / 12dp vertical container padding.

**Joined device path** (double scaffold — produces the extra padding):

```kotlin
// app/.../ui/screens/keysign/JoinKeysignView.kt:68-75
JoinKeysignScreen(
    title = title,
    onBack = viewModel::navigateToHome,
    isError = state is Error,
    fullScreen = isKeysignInProgress,
    applyDefaultPaddings =
        !(state == JoinKeysign && verifyUiModel is VerifyUiModel.Swap) && !isKeysignFinished,
) {
    ...
    is VerifyUiModel.Send -> {
        VerifySendScreen(
            state = model.model,
            dappMetadata = dappMetadata,
            isConsentsEnabled = false,
            confirmTitle = stringResource(R.string.verify_transaction_join_keysign),
            onFastSignClick = {},
            onConfirm = viewModel::joinKeysign,
        )
    }
    ...
}
```

```kotlin
// app/.../ui/screens/keysign/JoinKeysignView.kt:210-231
@Composable
private fun JoinKeysignScreen(
    title: String,
    isError: Boolean,
    applyDefaultPaddings: Boolean = true,
    ...
) {
    ...
    V2Scaffold(   // <-- outer V2Scaffold (applies default paddings for the Send branch)
        onBackClick = ...,
        title = title,
        applyDefaultPaddings = applyDefaultPaddings,
        content = content,
    )
}
```

The exception that disables `applyDefaultPaddings` only fires for **Swap** (`verifyUiModel is VerifyUiModel.Swap`). For `VerifyUiModel.Send` (and similarly `Deposit` / `SignMessage`) the outer scaffold still applies its 16dp/12dp padding, and then the inner `V2Scaffold` inside `VerifySendScreen` applies the same paddings a second time.

Net effect on the joined-device verify screen for Send:
- Effective horizontal container padding: **32.dp** (16 + 16) instead of 16
- Effective vertical container padding: **24.dp** (12 + 12) instead of 12
- Two stacked top-bar slots (outer scaffold renders the title bar; inner scaffold renders an empty top bar because `hasToolbar = false`), pushing content further down

This also explains why the title in the joined screenshot reads `Sign Transaction` instead of `Send overview` — it is being supplied by the **outer** scaffold (`JoinKeysignView.kt:54-67`), and that `when` block falls through to `else -> R.string.sign_transaction` when `verifyUiModel` has not yet resolved at the moment of capture, whereas the inner `VerifySendScreen`'s own title (`R.string.verify_send_send_overview`) is suppressed by `hasToolbar = false` (default).

## Affected screens

The same double-scaffold pattern very likely affects the other join-keysign verify variants too, since they each provide their own `V2Scaffold` but are still wrapped by `JoinKeysignScreen`:

- `VerifyUiModel.Send` → `VerifySendScreen` (this report)
- `VerifyUiModel.Deposit` → `VerifyDepositScreen`
- `VerifyUiModel.SignMessage` → `VerifySignMessageScreen`

`VerifyUiModel.Swap` is the one variant for which the outer scaffold already disables `applyDefaultPaddings`, so it does not exhibit the bug.

## Suggested fix

A few options, in order of smallest change:

1. **Extend the existing `applyDefaultPaddings = false` exception** in `JoinKeysignView.kt:73-74` to cover all four `VerifyUiModel` variants (Send/Swap/Deposit/SignMessage), since each variant brings its own `V2Scaffold`. Simplest expression:

   ```kotlin
   applyDefaultPaddings = !(state == JoinKeysign) && !isKeysignFinished
   ```

   (or enumerate the four subtypes explicitly).

2. **Stop double-wrapping**: have `JoinKeysignView` render the four verify composables outside `JoinKeysignScreen`'s `V2Scaffold` and instead let those composables drive the scaffold + top bar themselves. The title would then come from the verify composable (with `hasToolbar = true`) and match what the initiator path produces. The Swap path already passes `showToolbar = false`, so its toolbar would need to be re-enabled.

3. **Make the inner verify composables stateless about chrome**: split `VerifySendScreen` into a content composable and a scaffold-wrapping public composable. `JoinKeysignView` would then call the content composable inside its own scaffold (no double padding, single top bar from the outer scaffold). This is the cleanest long-term shape and aligns the title source for both code paths.

## Acceptance criteria

- Joined-device verify screen for **Send** has identical container padding, card position, top-bar appearance, and bottom-button position as the initiator's `Send overview` screen.
- Title on the joined-device verify screen reads `Send overview` (consistent with the initiator), with the only differences being the absence of the two confirmation checkboxes and the button label changing from `Sign transaction` → `Join Keysign`.
- Same fix verified for the Deposit and Sign Message join-verify variants.
- Swap join-verify path is unaffected (regression check).

## Repro

1. Initiate a Send transaction from device A and reach the `Send overview` screen.
2. On device B (joining device), open the QR scan and join the keysign session.
3. Compare the joined-device verify screen against the initiator's `Send overview` — note the extra horizontal/vertical padding around the card and the displaced button.

aminsato:
Closes #4538.

## Problem

The joined-device verify screen wrapped each `Verify*Screen` inside an outer `V2Scaffold` that already applied default 16dp/12dp padding, while each verify composable provided its **own** scaffold with the same padding — netting double container padding and a stacked top bar (showing the fallback `Sign Transaction` title instead of the correct per-variant title).

Effective padding per join-keysign variant before this PR:

| Variant | Inner pad | Outer pad | Total | Initiator (correct) |
|---|---|---|---|---|
| Send | 16h/12v | 16h/12v | **32h/24v** | 16h/12v |
| Swap | 16h/12v | suppressed | 16h/12v | 16h/12v |
| Deposit | 16h/16v | 16h/12v | **32h/28v** | 16h/16v |
| SignMessage | 16h/12v | 16h/12v | **32h/24v** | 16h/12v |

Only Swap was already correct, via an existing `applyDefaultPaddings = false` carve-out for the swap branch.

## Fix

Pull the `JoinKeysign` state branch out of the outer `JoinKeysignScreen { … }` wrapper so each `Verify*Screen` owns its own scaffold + toolbar via `hasToolbar = true` / `showToolbar = true`, mirroring what the initiator path does.

- `JoinKeysignView.kt`: handle `state == JoinKeysign` before the `JoinKeysignScreen` wrapper; each Verify variant now passes `hasToolbar = true` and `onBackClick = viewModel::navigateToHome`. A top-level `BackHandler` is added for the JoinKeysign branch (since the one inside `JoinKeysignScreen` no longer fires for this state). Drop the Swap-only `applyDefaultPaddings` carve-out and the inner-verify title cases (they were only there to fake what the inner toolbars now provide).
- `VerifySignMessageScreen.kt`: add `hasToolbar`/`onBackClick` params (defaulting to false / no-op so existing call sites are unaffected) and a `VsTopAppBar` in its `Scaffold.topBar`. Mirrors the pattern already in `VerifyDepositScreen`.
- Add four `@Preview` composables to `JoinKeysignView.kt` — one per variant — to make the new layout reviewable side-by-side with the existing initiator-path previews in `VerifySendScreen.kt`, `VerifySwapScreen.kt`, `VerifyDepositScreen.kt`, `VerifySignMessageScreen.kt`.

After this PR, each variant renders with a single scaffold and the correct per-variant title (`Send overview`, `Swap overview`, `Function overview`, `Verify`).

## Test plan

- [x] `./gradlew :app:compileDebugKotlin --rerun-tasks` — green
- [x] `./gradlew :app:ktfmtFormat` — clean
- [x] All other call sites of the four verify screens (NavGraph, SignMessageScreen, DepositScreen, BondFormScreen, PreviewActivity) verified unaffected — the new params have defaults
- [ ] Manual smoke on multi-device keysign:
  - [ ] Send: joined-device verify card matches initiator's `Send overview` position + padding, title reads `Send overview`
  - [ ] Swap: joined-device verify still renders correctly (regression check), title reads `Swap overview`
  - [ ] Deposit: joined-device verify matches initiator's `Function overview` padding, title reads `Function overview`
  - [ ] SignMessage: joined-device verify shows the new toolbar with title `Verify`, no double padding

## Summary by CodeRabbit

* **UI Improvements**
  * Streamlined keysign join flow to render verification screens earlier, improve back navigation, and simplify transaction titles.
  * Added optional toolbar with back-button support to sign-message, swap, send, and deposit verification screens.
  * Refined default paddings when keysign is finished for consistent layout.
  * Added preview configurations for send, swap, and deposit verification screens to aid visual testing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4549?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
@@ -15,15 +15,10 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.navigation.NavHostController
 import com.vultisig.wallet.R
-import com.vultisig.wallet.data.models.Coins
 import com.vultisig.wallet.ui.components.errors.ErrorView
 import com.vultisig.wallet.ui.components.errors.ErrorViewButtonUiModel
 import com.vultisig.wallet.ui.components.v2.scaffold.V2Scaffold
 import com.vultisig.wallet.ui.models.KeySignWrapperViewModel
-import com.vultisig.wallet.ui.models.TransactionDetailsUiModel
-import com.vultisig.wallet.ui.models.VerifyTransactionUiModel
-import com.vultisig.wallet.ui.models.deposit.DepositTransactionUiModel
-import com.vultisig.wallet.ui.models.deposit.VerifyDepositUiModel
 import com.vultisig.wallet.ui.models.keysign.JoinKeysignError
 import com.vultisig.wallet.ui.models.keysign.JoinKeysignState.DiscoverService
 import com.vultisig.wallet.ui.models.keysign.JoinKeysignState.DiscoveringSessionID
@@ -34,11 +29,6 @@ import com.vultisig.wallet.ui.models.keysign.JoinKeysignState.WaitingForKeysignS
 import com.vultisig.wallet.ui.models.keysign.JoinKeysignViewModel
 import com.vultisig.wallet.ui.models.keysign.KeysignState
 import com.vultisig.wallet.ui.models.keysign.VerifyUiModel
-import com.vultisig.wallet.ui.models.sign.SignMessageTransactionUiModel
-import com.vultisig.wallet.ui.models.sign.VerifySignMessageUiModel
-import com.vultisig.wallet.ui.models.swap.SwapTransactionUiModel
-import com.vultisig.wallet.ui.models.swap.ValuedToken
-import com.vultisig.wallet.ui.models.swap.VerifySwapUiModel
 import com.vultisig.wallet.ui.navigation.Route
 import com.vultisig.wallet.ui.screens.deposit.VerifyDepositScreen
 import com.vultisig.wallet.ui.screens.send.VerifySendScreen
@@ -74,7 +64,7 @@ internal fun JoinKeysignView(navController: NavHostController) {
                     dappMetadata = dappMetadata,
                     isConsentsEnabled = false,
                     hasToolbar = true,
-                    confirmTitle = stringResource(R.string.verify_transaction_join_keysign),
+                    confirmTitle = stringResource(R.string.verify_swap_sign_button),
                     onBackClick = viewModel::navigateToHome,
                     onFastSignClick = {},
                     onConfirm = viewModel::joinKeysign,
@@ -85,7 +75,7 @@ internal fun JoinKeysignView(navController: NavHostController) {
                 VerifySwapScreen(
                     state = model.model,
                     dappMetadata = dappMetadata,
-                    showToolbar = true,
+                    hasToolbar = true,
                     onBackClick = viewModel::navigateToHome,
                     confirmTitle = stringResource(R.string.verify_swap_sign_button),
                     isConsentsEnabled = false,
@@ -260,94 +250,5 @@ private fun JoinKeysignViewPreview() {
                     ErrorViewButtonUiModel(text = stringResource(R.string.try_again), onClick = {}),
             )
         },
-    )
-}
-
-@Preview
-@Composable
-private fun JoinKeysignSendVerifyPreview() {
-    VerifySendScreen(
-        state =
-            VerifyTransactionUiModel(
-                transaction =
-                    TransactionDetailsUiModel(
-                        srcAddress = "0x1111111111111111111111111111111111111111",
-                        dstAddress = "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48",
-                    )
-            ),
-        isConsentsEnabled = false,
-        hasToolbar = true,
-        confirmTitle = stringResource(R.string.verify_transaction_join_keysign),
-        onBackClick = {},
-        onFastSignClick = {},
-        onConfirm = {},
-    )
-}
-
-@Preview
-@Composable
-private fun JoinKeysignSwapVerifyPreview() {
-    VerifySwapScreen(
-        state =
-            VerifySwapUiModel(
-                tx = SwapTransactionUiModel(totalFee = "1.00$", hasConsentAllowance = true),
-                vaultName = "Main Vault",
-            ),
-        showToolbar = true,
-        onBackClick = {},
-        confirmTitle = stringResource(R.string.verify_swap_sign_button),
-        isConsentsEnabled = false,
-        onFastSignClick = {},
-        onConfirm = {},
-    )
-}
-
-@Preview
-@Composable
-private fun JoinKeysignDepositVerifyPreview() {
-    VerifyDepositScreen(
-        state =
-            VerifyDepositUiModel(
-                depositTransactionUiModel =
-                    DepositTransactionUiModel(
-                        token =
-                            ValuedToken(
-                                token = Coins.ThorChain.RUNE,
-                                value = "1 RUNE",
-                                fiatValue = "$1.37",
-                            ),
-                        networkFeeFiatValue = "$0.03",
-                        networkFeeTokenValue = "0.02 RUNE",
-                        srcAddress = "thor1abc456bca",
-                        dstAddress = "thor1abc456bca",
-                        operation = "mint",
-                        memo = "BOND:addressHere",
-                    )
-            ),
-        hasToolbar = true,
-        confirmTitle = stringResource(R.string.verify_swap_sign_button),
-        onBackClick = {},
-        onFastSignClick = {},
-        onConfirm = {},
-    )
-}
-
-@Preview
-@Composable
-private fun JoinKeysignSignMessageVerifyPreview() {
-    VerifySignMessageScreen(
-        state =
-            VerifySignMessageUiModel(
-                model =
-                    SignMessageTransactionUiModel(
-                        method = "personal_sign",
-                        message = "Sign in to Uniswap",
-                    )
-            ),
-        hasToolbar = true,
-        confirmTitle = stringResource(R.string.verify_swap_sign_button),
-        onBackClick = {},
-        onFastSignClick = {},
-        onConfirm = {},
     )
 }
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
@@ -15,10 +15,15 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.navigation.NavHostController
 import com.vultisig.wallet.R
+import com.vultisig.wallet.data.models.Coins
 import com.vultisig.wallet.ui.components.errors.ErrorView
 import com.vultisig.wallet.ui.components.errors.ErrorViewButtonUiModel
 import com.vultisig.wallet.ui.components.v2.scaffold.V2Scaffold
 import com.vultisig.wallet.ui.models.KeySignWrapperViewModel
+import com.vultisig.wallet.ui.models.TransactionDetailsUiModel
+import com.vultisig.wallet.ui.models.VerifyTransactionUiModel
+import com.vultisig.wallet.ui.models.deposit.DepositTransactionUiModel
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
```diff
@@ -245,3 +262,92 @@ private fun JoinKeysignViewPreview() {
         },
     )
 }
+
+@Preview
+@Composable
+private fun JoinKeysignSendVerifyPreview() {
```
rkokhatskyi: These four previews live in the wrong file since each of the screens being previewed already carries its own `@Preview` block at `VerifySendScreen.kt:571`, `VerifySwapScreen.kt:651`, `VerifyDepositScreen.kt:342`, and `VerifySignMessageScreen.kt:160`, so the new `hasToolbar = true` variants belong as siblings next to those existing previews whereas hosting them here drags `Coins`, `SwapTransactionUiModel`, `VerifySwapUiModel`, `VerifyDepositUiModel`, `VerifySignMessageUiModel`, `SignMessageTransactionUiModel`, `ValuedToken`, `DepositTransactionUiModel`, `TransactionDetailsUiModel`, and `VerifyTransactionUiModel` into `JoinKeysignView.kt` purely for preview construction

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
```diff
@@ -51,17 +61,70 @@ internal fun JoinKeysignView(navController: NavHostController) {
...
+                )
+            }
+
+            is VerifyUiModel.Swap -> {
+                VerifySwapScreen(
+                    state = model.model,
+                    dappMetadata = dappMetadata,
+                    showToolbar = true,
+                    onBackClick = viewModel::navigateToHome,
+                    confirmTitle = stringResource(R.string.verify_swap_sign_button),
```
rkokhatskyi, line 80: `confirmTitle` is `verify_transaction_join_keysign` (`Join Keysign`) on the Send branch at line 77 but `verify_swap_sign_button` (`Sign transaction`) here and on the Deposit and SignMessage branches, so the joined-device CTA reads differently per transaction type within the same flow whereas iOS ships one label per family and Windows ships one label across all join variants. Pick one string and apply it to all four branches

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
```diff
@@ -51,17 +61,70 @@ internal fun JoinKeysignView(navController: NavHostController) {
...
+                    onFastSignClick = {},
+                    onConfirm = viewModel::joinKeysign,
+                )
+            }
+
+            is VerifyUiModel.Swap -> {
+                VerifySwapScreen(
+                    state = model.model,
+                    dappMetadata = dappMetadata,
+                    showToolbar = true,
```
rkokhatskyi: `VerifySwapScreen` takes `showToolbar` whereas `VerifySendScreen`, `VerifyDepositScreen`, and the newly extended `VerifySignMessageScreen` all take `hasToolbar`, so one composable family hosts both spellings of the same concept which the PR perpetuates by adding yet another `hasToolbar` while leaving Swap's `showToolbar` untouched, worth normalizing while the API is already in flux

app/src/main/java/com/vultisig/wallet/ui/screens/sign/VerifySignMessageScreen.kt
```diff
@@ -92,9 +97,19 @@ private fun VerifySignMessageScreen(
     confirmTitle: String,
     onFastSignClick: () -> Unit,
     onConfirm: () -> Unit,
+    hasToolbar: Boolean = false,
+    onBackClick: () -> Unit = {},
```
rkokhatskyi: `onBackClick` is wired only inside the conditional `topBar` slot so when `hasToolbar = false` a caller-provided `onBackClick` is silently dropped, and these defaults also diverge from `VerifySwapScreen.kt:127,136` where the equivalent params are required with no defaults, so sibling APIs ended up with different ergonomics for the same concern

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
aminsato: Good call — moved each of the four previews next to its sibling in `VerifySendScreen.kt`, `VerifySwapScreen.kt`, `VerifyDepositScreen.kt`, and `VerifySignMessageScreen.kt`, and dropped the 10 preview-only model imports from `JoinKeysignView.kt`. Done in 7eaa96786.

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
aminsato: Agreed. Standardized all four join variants to `verify_swap_sign_button` ("Sign transaction") — the label three of four branches already shipped. Done in 7eaa96786.

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/JoinKeysignView.kt
aminsato: Renamed `showToolbar` → `hasToolbar` in `VerifySwapScreen` (all three overload signatures + internal forwards + V2Scaffold `takeIf` blocks + preview) so the Verify* family is now consistent. Updated the two call sites in `JoinKeysignView` (Swap branch) and `PreviewActivity`. Left the unrelated `showToolbar` in `TransactionDoneScreen`/`TxDoneScaffold`/`KeysignScreen` alone since they're a different family — happy to fold them in too if you'd prefer one sweep. Done in 7eaa96786.

app/src/main/java/com/vultisig/wallet/ui/screens/sign/VerifySignMessageScreen.kt
aminsato: Removed the defaults — both `hasToolbar` and `onBackClick` are now required on the state-based and private `VerifySignMessageScreen` overloads, matching `VerifySwapScreen.kt:127,136`. Updated the only non-join caller (the VM-injected wrapper at the top of the file) to pass `hasToolbar = false, onBackClick = {}` explicitly, and the existing preview likewise. Done in 7eaa96786.

aminsato:
@coderabbitai full review

rkokhatskyi (approved):
LGTM

PR #4561, opened 2026-05-17
fix(tokens): trigger auto-discovery on vault create and chain enable

### LINKED ISSUE #4557: Trigger token auto-discovery on vault create and chain enable
## Context

`DiscoverTokenUseCase` is only invoked from:
- `ChainTokensViewModel.loadData` (chain detail screen open + pull-to-refresh)
- `ChainTokensViewModel.selectTokens` (after manual token edit)
- `ImportFileViewModel.insertVaultToDb` (vault import, best-effort)

It is **not** invoked from:
- `SaveVaultUseCaseImpl` — fresh vault creation adds native tokens only, no discovery
- `ChainSelectionViewModel.onCommitChanges` — enabling a new chain adds the native token but never discovers anything else

Result: user creates a vault or enables a chain, receives an ERC-20 → nothing appears until they manually navigate into that chain's detail screen or pull-to-refresh.

iOS triggers discovery at both points (`VaultDefaultCoinService.swift:60` on vault create, `CoinService.swift:135` on chain add).

## Fix

- Call `discoverTokenUseCase(vault.id, null)` at the end of `SaveVaultUseCaseImpl.invoke`
- Call `discoverTokenUseCase(vaultId, chainId)` after each enabled chain in `ChainSelectionViewModel.onCommitChanges`

## Related

#4448. Trigger-point gap surfaced during the Android verification of iOS#4334.

rkokhatskyi:
Closes #4557.

`DiscoverTokenUseCase` was only invoked from `ChainTokensViewModel` (chain detail screen open + pull-to-refresh) and `ImportFileViewModel`. Fresh vault creation and the chain-selection commit each added the native token, but neither scheduled discovery, so any ERC-20 already sitting at the vault address stayed invisible until the user opened that chain manually. iOS triggers discovery at both points (`VaultDefaultCoinService.swift:60`, `CoinService.swift:135`).

## Changes
- `SaveVaultUseCaseImpl` now calls `discoverTokenUseCase(vault.id, null)` after the save completes, covering both the create path and the override (re-save / migrate) path.
- `ChainSelectionViewModel.onCommitChanges` calls `discoverTokenUseCase(vaultId, chain.coin.chain.raw)` after each chain successfully enables. Failures keep their existing snackbar surface and skip the discovery call for that chain.

## Tests
- `./gradlew :app:testDebugUnitTest --tests SaveVaultUseCaseImplTest --tests ChainSelectionViewModelTest`
- `./gradlew :app:ktfmtCheck`
- `./gradlew :app:lintDebug`

## Summary by CodeRabbit

* **New Features**
  * Token discovery is now scheduled when saving vaults and when enabling chains; discovery runs once per enabled native chain.

* **Bug Fixes / Reliability**
  * Discovery scheduling is best-effort: non-cancellation failures are caught and logged so saves and chain operations still succeed.
  * Discovery jobs are uniquely scoped per vault+chain to avoid collisions.

* **Behavior Change**
  * Token discovery removed from the import flow.

* **Tests**
  * Expanded tests cover discovery scheduling, error handling, and call ordering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4561?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src/main/java/com/vultisig/wallet/data/usecases/SaveVaultUseCase.kt
```diff
@@ -69,5 +70,8 @@ constructor(
                     vaultRepository.addTokenToVault(vaultId, updatedNativeToken)
                 }
         }
+        // Schedule background discovery so any tokens already held at the vault
+        // addresses surface without the user having to open each chain screen.
+        discoverTokenUseCase(vault.id, null)
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed in 7faabefd3 — discovery scheduling is now wrapped in a try/catch that swallows non-cancellation exceptions and logs via Timber, so a WorkManager enqueue failure can't fail the save.

app/src/main/java/com/vultisig/wallet/ui/models/ChainSelectionViewModel.kt
```diff
@@ -115,7 +117,9 @@ constructor(
                     // malformed `pubKeyECDSA` or `hexChainCode`.
                     Timber.w(e, "Failed to enable chain %s", chain.coin.chain.raw)
                     failedChains += chain.coin.chain.raw
+                    return@forEach
                 }
+                discoverTokenUseCase(vaultId, chain.coin.chain.raw)
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed in 7faabefd3 — same guard around the per-chain discovery call so the rest of the commit loop runs even if one chain's enqueue throws.

app/src/test/java/com/vultisig/wallet/ui/models/ChainSelectionViewModelTest.kt
```diff
@@ -0,0 +1,199 @@
...
+
+    @Test
+    fun `triggers token discovery for each successfully enabled chain`() =
+        runTest(testDispatcher) {
+            val vm = createViewModel()
+            val ethCoin = nativeCoin(Chain.Ethereum)
+            val btcCoin = nativeCoin(Chain.Bitcoin)
+            coEvery { vaultRepository.get(VAULT_ID) } returns vault()
+            coEvery { chainAccountAddressRepository.getAddress(any<Coin>(), any<Vault>()) } returns
+                Pair("address", "derivedPubKey")
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed in 7faabefd3 — moved the get/getEnabledChains/getAddress/nativeTokens stubs into setUp so the ViewModel sees a fully wired graph at construction time.

app/src/main/java/com/vultisig/wallet/data/usecases/SaveVaultUseCase.kt
```diff
@@ -69,5 +71,16 @@ constructor(
                     vaultRepository.addTokenToVault(vaultId, updatedNativeToken)
                 }
         }
+        // Schedule background discovery so any tokens already held at the vault
+        // addresses surface without the user having to open each chain screen.
+        // Best-effort: the vault is already persisted, so a discovery failure
+        // here must not surface as a save failure to callers.
+        try {
+            discoverTokenUseCase(vault.id, null)
```
johnnyluo: **REPLACE collision on init-vault path.**

`DiscoverTokenUseCaseImpl` always enqueues under the fixed name `"TokenRefreshWorker"` with `ExistingWorkPolicy.REPLACE`. During vault creation the sequence is:
1. `SaveVaultUseCaseImpl.invoke` → schedules a full-vault scan (`chainId = null`)
2. `ChainSelectionViewModel.onCommitChanges` (with `routeFromInitVault = true`) → schedules one per-chain scan for each enabled chain, each call replacing the previous

If the full-vault scan hasn't started by the time those per-chain calls land, every call cancels the prior one and only the **last chain** gets discovered — worse than not having this call at all.

Possible fixes: use `KEEP` so the first full-vault scan runs to completion; include `chainId` in the unique work name so per-chain calls don't collide with each other; or skip this call when the init-vault chain-selection step handles discovery.

**Also:** `ImportFileViewModel` already calls `discoverTokenUseCase` directly, so the import path now schedules discovery twice. With `REPLACE` it's harmless, but the explicit call in `ImportFileViewModel` could be removed since it's covered here.

app/src/main/java/com/vultisig/wallet/data/usecases/SaveVaultUseCase.kt
```diff
@@ -69,5 +71,16 @@ constructor(
...
+        } catch (e: CancellationException) {
```
johnnyluo: Nit: `DiscoverTokenUseCase` is `(String?, String?) -> Unit` — a plain (non-suspend) function. `CancellationException` can only be thrown from a suspension point, so this guard can never fire. It's copied from the `enableAccount` pattern where it's genuinely needed; here it's dead code. Safe to drop.

app/src/main/java/com/vultisig/wallet/ui/models/ChainSelectionViewModel.kt
```diff
@@ -115,6 +117,18 @@ constructor(
                     // malformed `pubKeyECDSA` or `hexChainCode`.
                     Timber.w(e, "Failed to enable chain %s", chain.coin.chain.raw)
                     failedChains += chain.coin.chain.raw
+                    return@forEach
+                }
+                try {
+                    discoverTokenUseCase(vaultId, chain.coin.chain.raw)
```
johnnyluo: Each iteration uses the same fixed work name `"TokenRefreshWorker"` with `REPLACE` policy, so when multiple chains are enabled in one `onCommitChanges` call each enqueue cancels the previous. Only the last chain's discovery runs.

Consider either appending the `chainId` to the unique work name (so each chain gets its own slot) or using `APPEND_OR_REPLACE` to queue them rather than cancel. This is the same root cause as the collision noted in `SaveVaultUseCaseImpl`.

app/src/main/java/com/vultisig/wallet/data/usecases/SaveVaultUseCase.kt
rkokhatskyi: Good catch — verified it. Pushed 7e51cb08e: encoded `(vaultId, chainId)` into the unique work name so the full-vault scan and the per-chain scans no longer compete for the same slot. REPLACE still collapses repeats within a single scope, so pull-to-refresh on a chain screen keeps doing the right thing.

On the duplicate import call — agreed it was harmless under REPLACE but felt sloppy now that SaveVault covers it, so I removed it (and the `discoverToken` injection in ImportFileViewModel) in the same commit. The 'discoverToken fails' regression test moved up to SaveVaultUseCaseImplTest where the contract now lives.

app/src/main/java/com/vultisig/wallet/data/usecases/SaveVaultUseCase.kt
rkokhatskyi: You're right — `DiscoverTokenUseCase` is `(String?, String?) -> Unit`, no suspension point, so cancellation can't propagate. Dropped the catch in 7e51cb08e and tightened both call sites to `runCatching { ... }.onFailure { ... }` while I was there.

app/src/main/java/com/vultisig/wallet/ui/models/ChainSelectionViewModel.kt
rkokhatskyi: Yep, same root cause as the SaveVault thread. Fixed at the use case layer in 7e51cb08e — the unique work name now carries `vaultId` and `chainId`, so enabling several chains in one commit enqueues N independent workers instead of one chain winning a REPLACE race. Same dead `CancellationException` catch dropped here too.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/DiscoverTokenUseCase.kt
```diff
@@ -17,25 +16,35 @@ interface DiscoverTokenUseCase : (String?, String?) -> Unit
...
-                TAG,
+                uniqueWorkName(vaultId, chainId),
                 ExistingWorkPolicy.REPLACE,
-                OneTimeWorkRequestBuilder<TokenRefreshWorker>()
-                    .setInputData(workData)
-                    .setConstraints(Constraints.Builder().build())
-                    .addTag(TAG)
-                    .build(),
+                request,
             )
```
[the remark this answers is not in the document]
rkokhatskyi: Verified against the original source — the `.setConstraints(Constraints.Builder().build())` line that I removed has been an empty no-op since #961 introduced the file in Sep 2024. `Constraints.Builder().build()` produces a Constraints object with all defaults (`NetworkType.NOT_REQUIRED`, no battery/charging/idle/storage requirements), which is equivalent to passing no Constraints at all. So the worker has always been able to start offline; this PR didn't regress that.

I'm going to push back on adding `NetworkType.CONNECTED` in this PR for a few reasons:

1. **Scope.** The ticket is to trigger discovery at two new call sites. Changing the worker's pre-existing scheduling semantics affects every existing caller too — `ChainTokensViewModel.loadData` (pull-to-refresh) has always run immediately, and adding a constraint would silently defer it. That's a UX call worth making deliberately, not as a side effect here.

2. **Cross-platform parity.** iOS (`CoinService.addDiscoveredTokens`) fires immediately and fails fast on no network; Windows has no auto-discovery at all. Adding a queue-until-online policy only on Android diverges from both without explicit intent.

3. **The `Result.failure()` path in `TokenRefreshWorker` is the real fragility.** Even with `NetworkType.CONNECTED`, mid-flight network drops still produce `Result.failure()` with no backoff retry. The right fix is in the worker itself (translate transient failures to `Result.retry()`), not in the request builder. That's a clean follow-up but a meaningfully different PR.

Happy to file a follow-up issue covering both — would you like that tracked separately?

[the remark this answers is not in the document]
rkokhatskyi: Tracked as #4578. Closing this thread out — feel free to chime in there if you have opinions on the backoff policy or the UX implication for pull-to-refresh.

johnnyluo (approved):
LGTM

PR #4571, opened 2026-05-17
fix(send): skip fee round-trip on non-native percentage chips

### LINKED ISSUE #4563: Send screen: 50%/75%/MAX percentage chips buffer for a long time (regression)
## Summary

On the Send screen, tapping any percentage chip (50%, 75%, MAX) used to populate the amount near-instantly. Currently there's a multi-second buffer before the amount appears. Regression — never happened before.

Repro on USDC / Solana in the screenshot; please verify against other chains/assets to see if it's universal or chain-specific.

## Probable affected area

- `app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt:136` — `calculatePercentageWithAccurateFee`
- `app/src/main/java/com/vultisig/wallet/ui/screens/send/SendFormAmountSection.kt:196` — `PercentageChip` call site
- `app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt:700, 1085` — preemption logic comments suggest in-flight percentage calc

Starting point only — fee-estimation round-trip on chip tap is the prime suspect. Please investigate timing and decide whether to debounce, prefetch, or render with stale fee + update.

## Repro

1. Send → pick any asset (e.g. USDC on Solana).
2. Tap 50% / 75% / MAX chip.
3. Observe multi-second delay before the amount field populates.

## Platform

Android. Cross-platform check not done — let me know if you want iOS/Windows verified.

rkokhatskyi:
Closes #4563.

## Summary

Tapping 50%/75%/MAX on the Send screen for a non-native token on a non-EVM chain (USDC on Solana in the report) buffers for a few seconds before the amount field populates. The chip-tap path in `AmountFractionManager` always re-runs `feeServiceComposite.calculateFees` for that case, even though chain gas on non-native tokens is paid in the native coin and never reduces the selected balance (see `GetAvailableTokenBalanceUseCase`). The fresh fetch is a multi-second Solana RPC round-trip that the percentage calc ignores.

Extend the existing EVM short-circuit to also fire when the token is non-native, so the chip-tap returns the cached estimate immediately. `GasFeeOrchestrator` still refreshes `gasFee` in the background once the amount field updates, so the fee display catches up without blocking the input.

While in the same code path, mirror `fetchPercentageOfAvailableBalance`'s `setScale + stripTrailingZeros` normalization in the null-gasFee fallback so the new short-circuit returns the same "6" / "12.5" shape as the cached-fee path, instead of leaking "6.00" when `0.75f × 8` leaves trailing zeros.

Native tokens on non-EVM chains (native SOL, BTC, etc.) keep the current fresh-fee fetch, since fee freshness matters there (MAX needs the actual fee, BTC fee depends on UTXO selection) — that matches iOS behavior.

## Testing

- `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.send.*"` — 159 send.* tests, 0 failures.
- `./gradlew :app:lintDebug` — clean.
- Three new tests in `AmountFractionManagerTest` cover the non-native short-circuit for both percentage and max paths, including the null-gasFee variant. They assert no `feeServiceComposite.calculateFees` / `tokenRepository.getNativeToken` call is issued.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fee calculation consistency for percentage-based amount selection across different fee scenarios.
  * Enhanced handling of non-native token transactions to prevent unnecessary recalculations.

* **Tests**
  * Expanded test coverage for non-native token transactions on non-EVM chains.
  * Added validation to ensure fee recalculation is correctly bypassed when appropriate.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4571?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt
@@ -148,9 +148,14 @@ internal class AmountFractionManager(
             if (gasFee.value != null) {
                 fetchPercentageOfAvailableBalance(percentage)
             } else {
+                // Mirror fetchPercentageOfAvailableBalance's normalization so the
+                // short-circuit below returns the same shape as the fresh-fee path,
+                // e.g. "6" instead of "6.00" when 0.75f × 8 leaves trailing zeros.
                 getAvailableTokenBalance(selectedAccount, BigInteger.ZERO)
                     ?.decimal
-                    ?.multiply(percentage.toBigDecimal()) ?: BigDecimal.ZERO
+                    ?.multiply(percentage.toBigDecimal())
+                    ?.setScale(token.decimal, RoundingMode.DOWN)
+                    ?.stripTrailingZeros() ?: BigDecimal.ZERO
             }
 
         if (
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4574, opened 2026-05-17
fix(db): key transaction-history writes by chain-scoped id

### LINKED ISSUE #4025: DAO UPDATE queries use bare txHash without chain filter
## What happens

All UPDATE queries in `TransactionHistoryDao` target rows with `WHERE txHash = :txHash` but the primary key is `chain:txHash`. The unique index on `txHash` alone makes this work today because no two rows can share a txHash. But the design is contradictory: the PK was built to be chain-scoped while every write path ignores the chain.

If the unique index on `txHash` were ever relaxed every UPDATE would silently touch the wrong row. The `upsertFromBackfill` lookup also uses `getByTxHash` with no chain filter.

Not a user-visible issue today because the supported chains produce cryptographically distinct hashes. But the code should use the chain-scoped `id` column for writes instead of bare `txHash` so the PK design is actually used.

## Affected queries

```
updateStatus          →  WHERE txHash = :txHash
updateToConfirmed     →  WHERE txHash = :txHash
updateToFailed        →  WHERE txHash = :txHash
demoteForReorg        →  WHERE txHash = :txHash
incrementRetryCount   →  WHERE txHash = :txHash
getByTxHash           →  WHERE txHash = :txHash
```

## Where to look

`data/src/main/kotlin/com/vultisig/wallet/data/db/dao/TransactionDao.kt`

rkokhatskyi:
Closes #4025.

`TransactionHistoryEntity` was defined with a chain-scoped primary key (`id = "$chain:$txHash"`), but every UPDATE in `TransactionHistoryDao` keyed on bare `txHash`. The unique index on `txHash` alone is what currently routes writes to the right row, so there is no observable bug today — only a contradiction between the schema design and the queries that exercise it. If that index ever loosens, every UPDATE would silently flip onto whichever colliding row Room found first.

## Summary

- All UPDATE queries (`updateStatus`, `updateToConfirmed`, `updateToFailed`, `updateToRefunded`, `demoteForReorg`, `incrementRetryCount`) and the `upsertFromBackfill` snapshot lookup now key on `WHERE id = :id`, using the chain-scoped primary key.
- `getByTxHash` is replaced with `getById`. `TransactionHistoryRepository.updateTransactionStatus`, `getTransaction`, `demoteForReorg`, and `incrementRetryCount` accept `chain` alongside `txHash` and build the id internally with the existing `buildTransactionHistoryId` helper.
- Both call sites (`KeysignViewModel.startForegroundPolling` and `RefreshPendingTransactionsUseCase`) already had the chain in scope (`chain.raw` and `tx.chain` respectively).

## Testing

- `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest` — green; the existing `KeysignViewModelSaveTransactionHistoryTest` and `TransactionHistoryViewModelTest` keep passing.
- New `TransactionHistoryRepositoryImplTest` covers each write path, the lookup, the full `TransactionResult` mapping, and the regression case of the same hash on two different chains resolving to distinct DAO ids.
- `./gradlew :data:ktfmtCheck :app:ktfmtCheck` — clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Transaction tracking now correctly distinguishes identical transaction hashes on different blockchains, preventing cross-chain status mix-ups.

* **Improvements**
  * More robust status updates and retry handling for pending transactions and reorgs; clearer terminal-state guards to avoid downgrading finalized transactions.

* **Tests**
  * Added thorough tests covering multi-chain status flows, retry counts, and reorg demotion behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4574?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


johnnyluo (approved):
LGTM

PR #4579, opened 2026-05-18
fix(cardano): use require() so bad key length throws IllegalArgumentException

### LINKED ISSUE #4541: Crash: IllegalStateException in CardanoUtils.createEnterpriseAddress when enabling Cardano (ChainSelection)
## Problem
`java.lang.IllegalStateException` is thrown from `CardanoUtils.createEnterpriseAddress` (`data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoUtils.kt:39`) when enabling a chain in `ChainSelectionViewModel.onCommitChanges` → `enableAccount` → `ChainAccountAddressRepositoryImpl.getAddress`.

The thrown precondition is:

```kotlin
if (spendingKeyData.size != 32) {
    error("spending key must be 32 bytes, got ${spendingKeyData.size}")
}
```

i.e. the derived EdDSA spending key passed to `createEnterpriseAddress(spendingKeyHex)` decodes to something other than 32 bytes. Because `enableAccount` does not catch the failure, the uncaught exception aborts the enable-chain flow and crashes the app.

This is the same enable-chain flow as the previously fixed #4428 (which addressed an `InvalidParameterException` from the `PublicKey` ctor in the EdDSA branch) — but a different exception, in a different file, on the Cardano-specific code path that runs *after* `validateEddsaPubKey` at `ChainAccountAddressRepository.kt:102`.

## Crash environment
- **Versions affected:** 101
- **Devices:** Samsung 100% — most affected: Samsung Galaxy S23 (`dm1q`)
- **Android versions:** Android 16 (API 36) 100%
- **Events / users:** 1 event / 1 user
- **First observed:** May 16, 2026 (via Android Vitals)

## Stack trace
```
java.lang.IllegalStateException
  at com.vultisig.wallet.data.crypto.CardanoUtils.createEnterpriseAddress (CardanoUtils.java:39)
  at com.vultisig.wallet.data.repositories.ChainAccountAddressRepositoryImpl.getAddress (ChainAccountAddressRepository.kt:84)
  at com.vultisig.wallet.data.repositories.ChainAccountAddressRepositoryImpl.getAddress (ChainAccountAddressRepositoryImpl.java:108)
  at com.vultisig.wallet.ui.models.ChainSelectionViewModel.enableAccount (ChainSelectionViewModel.kt:118)
  at com.vultisig.wallet.ui.models.ChainSelectionViewModel.access$enableAccount (ChainSelectionViewModel.kt:40)
  at com.vultisig.wallet.ui.models.ChainSelectionViewModel$onCommitChanges$1.invokeSuspend (ChainSelectionViewModel.kt:94)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:33)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:100)
  ...
```

## Expected Behavior
- Enabling Cardano in chain selection should never crash the app.
- If the derived spending key cannot be parsed to a valid 32-byte value for any reason (bad vault state, missing key material, malformed hex), the failure should surface as a user-facing error and the chain should remain disabled — not abort the whole enable-chain flow.

## Steps to Reproduce
Exact repro on a physical device is not yet known. The crash occurred on a Samsung Galaxy S23 / Android 16 user. To investigate, try enabling Cardano with vaults whose EdDSA key material is in an unusual shape (e.g., partially provisioned, mid-migration, or imported from another platform).

## Scope
- [ ] Confirm whether `validateEddsaPubKey` at `ChainAccountAddressRepository.kt:102` validates byte length for the Cardano case, and align the Cardano `spendingKeyHex` length check with that contract (fail earlier and uniformly).
- [ ] Wrap the per-chain `enableAccount` call so a single chain failing does not crash the host activity — surface a per-chain error to the UI and let other chains in the same commit succeed.
- [ ] Add a unit test for `CardanoUtils.createEnterpriseAddress` covering non-32-byte inputs (mirroring the existing `ChainAccountAddressRepositoryEddsaValidationTest`).
- [ ] Confirm the fix via Android Vitals once the next release rolls out.

## Related
- #4428 (closed COMPLETED, 2026-05-09) — same flow, different exception (PublicKey ctor / `InvalidParameterException`)

johnnyluo:
## Summary
- Replace `error()` with `require()` in `CardanoUtils.createEnterpriseAddress` and `createExtendedKey` for the 32-byte length preconditions
- `error()` throws `IllegalStateException`; the chain-enable error handler in `ChainSelectionViewModel` catches `CancellationException` (which extends `IllegalStateException`) separately first — so an `IllegalStateException` from `error()` could slip past the intended handler and crash the app
- `require()` throws `IllegalArgumentException`, which is caught by the generic `Exception` handler correctly, surfacing a per-chain error instead of crashing
- Adds `CardanoUtilsTest` with 7 unit tests covering empty, short, and long key inputs for both functions

## Issue
Closes #4541

## Test Plan
- [x] Unit tests pass: `./gradlew :data:test`
- [x] Lint passes: `./gradlew :data:lintDebug`
- [ ] Manual: enable Cardano chain in chain selection — should not crash on malformed key material

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation and clearer error messages for Cardano key and address creation to reject inputs of incorrect byte lengths.

* **Tests**
  * Added tests covering invalid spending-key and chain-code inputs and address/key creation failure cases to ensure validation behaves as expected.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4579?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoUtils.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoUtils.kt
@@ -12,11 +12,11 @@ object CardanoUtils {
     fun createExtendedKey(spendingKeyHex: String, chainCodeHex: String): ByteArray {
         val spendingKeyData = Numeric.hexStringToByteArray(spendingKeyHex)
         val chainCodeData = Numeric.hexStringToByteArray(chainCodeHex)
-        if (spendingKeyData.size != 32) {
-            error("spending key must be 32 bytes, got ${spendingKeyData.size}")
+        require(spendingKeyData.size == 32) {
+            "spending key must be 32 bytes, got ${spendingKeyData.size}"
         }
-        if (chainCodeData.size != 32) {
-            error("chain code must be 32 bytes, got ${chainCodeData.size}")
+        require(chainCodeData.size == 32) {
+            "chain code must be 32 bytes, got ${chainCodeData.size}"
         }
 
         val extendedKeyData = ByteArray(128)
@@ -34,9 +34,8 @@ object CardanoUtils {
 
     fun createEnterpriseAddress(spendingKeyHex: String): String {
         val spendingKeyData = Numeric.hexStringToByteArray(spendingKeyHex)
-
-        if (spendingKeyData.size != 32) {
-            error("spending key must be 32 bytes, got ${spendingKeyData.size}")
+        require(spendingKeyData.size == 32) {
+            "spending key must be 32 bytes, got ${spendingKeyData.size}"
         }
 
         // Use Blake2b hash with 28 bytes output size
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoUtils.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoUtils.kt
@@ -9,14 +9,16 @@ import wallet.core.jni.Hash
 
 object CardanoUtils {
 
+    private const val KEY_SIZE = 32
+
     fun createExtendedKey(spendingKeyHex: String, chainCodeHex: String): ByteArray {
         val spendingKeyData = Numeric.hexStringToByteArray(spendingKeyHex)
         val chainCodeData = Numeric.hexStringToByteArray(chainCodeHex)
-        require(spendingKeyData.size == 32) {
-            "spending key must be 32 bytes, got ${spendingKeyData.size}"
+        require(spendingKeyData.size == KEY_SIZE) {
+            "spending key must be $KEY_SIZE bytes, got ${spendingKeyData.size}"
         }
-        require(chainCodeData.size == 32) {
-            "chain code must be 32 bytes, got ${chainCodeData.size}"
+        require(chainCodeData.size == KEY_SIZE) {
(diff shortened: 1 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoUtils.kt
```diff
@@ -12,11 +12,11 @@ object CardanoUtils {
     fun createExtendedKey(spendingKeyHex: String, chainCodeHex: String): ByteArray {
         val spendingKeyData = Numeric.hexStringToByteArray(spendingKeyHex)
         val chainCodeData = Numeric.hexStringToByteArray(chainCodeHex)
-        if (spendingKeyData.size != 32) {
-            error("spending key must be 32 bytes, got ${spendingKeyData.size}")
+        require(spendingKeyData.size == 32) {
```
rkokhatskyi: 32 is a magic number, it was a legacy logic but would be good to move to a variable

rkokhatskyi (approved):
LGTM

PR #4587, opened 2026-05-18
refactor(viewmodels): drop Context fields to resolve StaticFieldLeak

### LINKED ISSUE #3782: perf: resolve StaticFieldLeak warnings for Context fields in ViewModels
# [Fix] Resolve StaticFieldLeak warnings for Context fields in ViewModels [ViewModels]

## Problem
Lint flags 6 ViewModels that hold a `Context` field injected via `@ApplicationContext`. Although `@ApplicationContext` is safe (it's the Application context, not an Activity context), the `StaticFieldLeak` rule still warns because Hilt-injected ViewModels can outlive configuration changes and holding any `Context` reference is considered risky by the checker. The recommended fix is to replace the `Context` field with `Application` or refactor the logic into a use-case/repository.

## Expected Behavior
ViewModels should not trigger `StaticFieldLeak` lint warnings. Context usage should be explicit by using `Application` type or moved into dedicated use-cases/repositories, keeping ViewModels context-free where possible.

## Steps to Reproduce
1. Run lint analysis on the project
2. Observe `StaticFieldLeak` warnings for the 6 affected ViewModels
3. Review warnings at the specified line numbers in each file

### Must Do
- [ ] Replace `@ApplicationContext private val context: Context` with `@ApplicationContext private val context: Application` in all 6 affected ViewModels (lines: BackupVaultViewModel.kt:52, ImportFileViewModel.kt:66, MainViewModel.kt:58, NotificationsSettingsViewModel.kt:46, TokenAddressQrViewModel.kt:48, VaultAccountsViewModel.kt:122)
- [ ] Verify all Context usage sites are compatible with Application type (Application implements Context so call-sites should need no changes)
- [ ] Run lint to confirm `StaticFieldLeak` warnings are resolved

### Must NOT Do
- Don't change unrelated code in the ViewModels
- Don't modify any business logic or functionality beyond the type change
- Don't alter injection patterns or ViewModel lifecycle

## Acceptance Criteria
- [ ] `./gradlew` succeeds
- [ ] All 6 `StaticFieldLeak` warnings are resolved for the affected ViewModels
- [ ] Existing functionality remains unchanged (Application implements Context so behavior is preserved)
- [ ] No new lint warnings are introduced

## Context

**Affected files:**

| File | Line |
|------|------|
| `app/src/main/java/com/vultisig/wallet/ui/models/keygen/BackupVaultViewModel.kt` | 52 |
| `app/src/main/java/com/vultisig/wallet/ui/models/ImportFileViewModel.kt` | 66 |
| `app/src/main/java/com/vultisig/wallet/app/activity/MainViewModel.kt` | 58 |
| `app/src/main/java/com/vultisig/wallet/ui/models/settings/NotificationsSettingsViewModel.kt` | 46 |
| `app/src/main/java/com/vultisig/wallet/ui/models/TokenAddressQrViewModel.kt` | 48 |
| `app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt` | 122 |

**Lint rule:** `StaticFieldLeak` (Priority 6/10 — Performance Warning)

**Alternative approach:** If ViewModels only use Context to access resources or system services, consider extracting that logic into a dedicated use-case/repository that is itself injected with `@ApplicationContext`, keeping ViewModels completely context-free.

aminsato:
## Summary
- Eliminates `@ApplicationContext Context` fields from six ViewModels (MainViewModel, NotificationsSettingsViewModel, VaultAccountsViewModel, BackupVaultViewModel, ImportFileViewModel, TokenAddressQrViewModel) so lint stops flagging `StaticFieldLeak`.
- **Group A — strings**: `context.getString(...)` calls converted to `UiText`, resolved in the composable (e.g. transactionSummary, address-copied, push-notification errors). Snackbar collection/resolve moved from `MainViewModel` into `MainActivityContent`. A new UI-layer helper `pushNotificationErrorUiText(e): UiText` replaces the data-layer `pushNotificationErrorMessage(e, context)` function (deleted).
- **Group B — system services / file I/O**: extracted to `@ApplicationContext`-scoped use-cases:
  - `SaveBackupToUriUseCase`, `DeleteBackupDocumentUseCase` (BackupVaultViewModel)
  - `VaultFileReaderUseCase` (ImportFileViewModel: fileContent / fileName / isValidZipFile / processZip)
  - `ShareBitmapUseCase` (TokenAddressQrViewModel)
- ViewModels now have no Context or Application field.

Closes #3782

## Test plan
- [x] `./gradlew :app:compileDebugKotlin` succeeds.
- [x] `./gradlew lint` succeeds; `grep -c StaticFieldLeak app/build/reports/lint-results-debug.html` returns **0** (all six original warnings cleared).
- [x] `./gradlew test` passes after updating constructor params in MainViewModel / ImportFileViewModel / VaultAccountsViewModel unit tests.
- [ ] Manual smoke: backup vault flow, import vault (.bak / .zip), copy + share address from token QR sheet, toggle push notifications.

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * More reliable, non-blocking snackbars and improved localized push-notification error messages
  * Fixes to copy/share flows to avoid failures and ensure consistent feedback

* **New Features**
  * New sharing use-case improving QR and screenshot sharing reliability
  * Backup save/delete support with safer persistence

* **Refactor**
  * Streamlined file import and URI handling for more stable imports and clearer error reporting

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4587?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
@@ -40,7 +40,6 @@ import com.vultisig.wallet.ui.components.BiometryAuthScreen
 import com.vultisig.wallet.ui.components.banners.ForegroundNotificationBanner
 import com.vultisig.wallet.ui.components.banners.OfflineBanner
 import com.vultisig.wallet.ui.components.v2.snackbar.VsSnackBar
-import com.vultisig.wallet.ui.components.v2.snackbar.rememberVsSnackbarState
 import com.vultisig.wallet.ui.models.AccountUiModel
 import com.vultisig.wallet.ui.models.VaultAccountsUiModel
 import com.vultisig.wallet.ui.navigation.Route
@@ -67,7 +66,7 @@ internal fun MainActivityContent(
     if (foregroundNotification != null) lastNotification = foregroundNotification
 
     val context = LocalContext.current
-    val snackbarState = rememberVsSnackbarState()
+    val snackbarState = mainViewModel.snackbarState
     LaunchedEffect(snackbarFlow) {
         snackbarFlow.collectMessage { (message, type) ->
             val resolved = message.asString(context)
--- a/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
@@ -23,6 +23,7 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.layout.layout
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.semantics.testTagsAsResourceId
@@ -39,6 +40,7 @@ import com.vultisig.wallet.ui.components.BiometryAuthScreen
 import com.vultisig.wallet.ui.components.banners.ForegroundNotificationBanner
 import com.vultisig.wallet.ui.components.banners.OfflineBanner
 import com.vultisig.wallet.ui.components.v2.snackbar.VsSnackBar
+import com.vultisig.wallet.ui.components.v2.snackbar.rememberVsSnackbarState
 import com.vultisig.wallet.ui.models.AccountUiModel
 import com.vultisig.wallet.ui.models.VaultAccountsUiModel
 import com.vultisig.wallet.ui.navigation.Route
@@ -46,6 +48,8 @@ import com.vultisig.wallet.ui.navigation.SetupNavGraph
 import com.vultisig.wallet.ui.navigation.route
 import com.vultisig.wallet.ui.screens.home.VaultAccountsScreen
 import com.vultisig.wallet.ui.theme.Theme
+import com.vultisig.wallet.ui.utils.SnackbarFlow
+import com.vultisig.wallet.ui.utils.asString
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.launch
 
@@ -54,12 +58,23 @@ import kotlinx.coroutines.launch
 internal fun MainActivityContent(
     navController: NavHostController,
     mainViewModel: MainViewModel,
+    snackbarFlow: SnackbarFlow,
     startDestination: Any,
     onNavigationReady: () -> Unit,
 ) {
     val foregroundNotification by mainViewModel.foregroundNotification.collectAsStateWithLifecycle()
     var lastNotification by remember { mutableStateOf<ForegroundNotificationState?>(null) }
     if (foregroundNotification != null) lastNotification = foregroundNotification
+
+    val context = LocalContext.current
+    val snackbarState = rememberVsSnackbarState()
+    LaunchedEffect(snackbarFlow) {
+        snackbarFlow.collectMessage { (message, type) ->
+            val resolved = message.asString(context)
+            if (resolved.isBlank()) return@collectMessage
+            snackbarState.show(resolved, type)
+        }
+    }
 
     MainActivityContent(
         foregroundNotification = foregroundNotification,
@@ -101,7 +116,7 @@ internal fun MainActivityContent(
 
             VsSnackBar(
                 modifier = Modifier.align(Alignment.BottomCenter),
-                snackbarState = mainViewModel.snakeBarHostState,
+                snackbarState = snackbarState,
             )
         },
     )
@@ -173,7 +188,8 @@ private fun MainActivityContentPreview() {
         ForegroundNotificationState(
             qrCodeData = "preview",
             vaultName = "Main Vault",
-            transactionSummary = "Swap 10 ETH → USDC",
+            transactionSummary =
+                com.vultisig.wallet.ui.utils.UiText.DynamicString("Swap 10 ETH → USDC"),
         )
     MainActivityContent(
         foregroundNotification = notification,
--- a/app/src/main/java/com/vultisig/wallet/data/usecases/ShareBitmapUseCase.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/usecases/ShareBitmapUseCase.kt
@@ -1,0 +1,17 @@
+package com.vultisig.wallet.data.usecases
+
+import android.content.Context
+import android.graphics.Bitmap
+import com.vultisig.wallet.ui.utils.share
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+
+fun interface ShareBitmapUseCase {
+    operator fun invoke(bitmap: Bitmap, fileName: String)
+}
+
+internal class ShareBitmapUseCaseImpl
+@Inject
+constructor(@param:ApplicationContext private val context: Context) : ShareBitmapUseCase {
+    override fun invoke(bitmap: Bitmap, fileName: String) = context.share(bitmap, fileName)
+}
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/DataUsecasesModule.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/DataUsecasesModule.kt
@@ -4,12 +4,18 @@ import com.vultisig.wallet.data.usecases.backup.CreateVaultBackupFileNameUseCase
 import com.vultisig.wallet.data.usecases.backup.CreateVaultBackupFileNameUseCaseImpl
 import com.vultisig.wallet.data.usecases.backup.CreateZipVaultBackupFileNameUseCase
 import com.vultisig.wallet.data.usecases.backup.CreateZipVaultBackupFileNameUseCaseImpl
+import com.vultisig.wallet.data.usecases.backup.DeleteBackupDocumentUseCase
+import com.vultisig.wallet.data.usecases.backup.DeleteBackupDocumentUseCaseImpl
 import com.vultisig.wallet.data.usecases.backup.IsVaultBackupFileExtensionValidUseCase
 import com.vultisig.wallet.data.usecases.backup.IsVaultBackupFileExtensionValidUseCaseImpl
 import com.vultisig.wallet.data.usecases.backup.RequestServerBackupUseCase
 import com.vultisig.wallet.data.usecases.backup.RequestServerBackupUseCaseImpl
+import com.vultisig.wallet.data.usecases.backup.SaveBackupToUriUseCase
+import com.vultisig.wallet.data.usecases.backup.SaveBackupToUriUseCaseImpl
 import com.vultisig.wallet.data.usecases.chaintokens.GetChainTokensUseCase
 import com.vultisig.wallet.data.usecases.chaintokens.GetChainTokensUseCaseImpl
+import com.vultisig.wallet.data.usecases.file.VaultFileReaderUseCase
+import com.vultisig.wallet.data.usecases.file.VaultFileReaderUseCaseImpl
 import com.vultisig.wallet.data.usecases.tss.DiscoverParticipantsUseCase
 import com.vultisig.wallet.data.usecases.tss.DiscoverParticipantsUseCaseImpl
 import com.vultisig.wallet.data.usecases.tss.PullTssMessagesUseCase
@@ -243,4 +249,18 @@ internal interface DataUsecasesModule {
     @Binds
     @Singleton
     fun bindThorchainMemoParser(impl: ThorchainMemoParserImpl): ThorchainMemoParser
+
+    @Binds
+    @Singleton
+    fun bindSaveBackupToUriUseCase(impl: SaveBackupToUriUseCaseImpl): SaveBackupToUriUseCase
+
+    @Binds
+    @Singleton
+    fun bindDeleteBackupDocumentUseCase(
+        impl: DeleteBackupDocumentUseCaseImpl
+    ): DeleteBackupDocumentUseCase
+
+    @Binds
+    @Singleton
+    fun bindVaultFileReaderUseCase(impl: VaultFileReaderUseCaseImpl): VaultFileReaderUseCase
 }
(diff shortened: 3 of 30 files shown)


aminsato:
@coderabbitai full review

app/src/main/java/com/vultisig/wallet/data/usecases/ShareBitmapUseCase.kt
```diff
@@ -0,0 +1,17 @@
...
+import javax.inject.Inject
+
+fun interface ShareBitmapUseCase {
+    operator fun invoke(bitmap: Bitmap, fileName: String)
+}
+
+internal class ShareBitmapUseCaseImpl
+@Inject
+constructor(@param:ApplicationContext private val context: Context) : ShareBitmapUseCase {
+    override fun invoke(bitmap: Bitmap, fileName: String) = context.share(bitmap, fileName)
```
rkokhatskyi: `ShareBitmapUseCase` sits under `com.vultisig.wallet.data.usecases` in the `:app` module yet imports `com.vultisig.wallet.ui.utils.share` and is bound from `UseCasesModule` whereas the three sibling use cases live in `:data` and bind from `DataUsecasesModule`, so the package label says data while the dependency chain says ui and the module split looks arbitrary, since either renaming the package to `ui.usecases` or hoisting `ShareUtils` into `:data/common` would put all four wrappers next to each other

app/src/main/java/com/vultisig/wallet/data/usecases/ShareBitmapUseCase.kt
```diff
@@ -0,0 +1,17 @@
+package com.vultisig.wallet.data.usecases
+
+import android.content.Context
+import android.graphics.Bitmap
+import com.vultisig.wallet.ui.utils.share
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+
+fun interface ShareBitmapUseCase {
+    operator fun invoke(bitmap: Bitmap, fileName: String)
```
rkokhatskyi: `invoke` is non suspend whereas `Context.share` does `mkdirs`, opens a `FileOutputStream`, runs `bitmap.compress` and writes the JPEG before `startActivity`, all of which the sole caller already wraps in `withContext(Dispatchers.Main)` in `TokenAddressQrViewModel.shareQRCode`, so the disk write currently runs on the UI thread, whereas extracting this into a use case is the natural moment to make the contract `suspend` and push the I/O onto `Dispatchers.IO` since only `startActivity` actually needs Main

data/src/main/kotlin/com/vultisig/wallet/data/usecases/backup/SaveBackupToUriUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+}
+
+internal class SaveBackupToUriUseCaseImpl
+@Inject
+constructor(@param:ApplicationContext private val context: Context) : SaveBackupToUriUseCase {
+    override suspend fun invoke(uri: Uri, content: String): Boolean =
+        context.saveContentToUri(uri, content)
+}
+
+fun interface DeleteBackupDocumentUseCase {
```
rkokhatskyi: This file declares two unrelated top level interfaces, `SaveBackupToUriUseCase` and `DeleteBackupDocumentUseCase`, so `Go to declaration` on the delete contract lands in a file whose name implies the opposite operation, since each top level type should sit in its own file matching the type

data/src/main/kotlin/com/vultisig/wallet/data/usecases/backup/SaveBackupToUriUseCase.kt
```diff
@@ -0,0 +1,29 @@
+package com.vultisig.wallet.data.usecases.backup
+
+import android.content.Context
+import android.net.Uri
+import com.vultisig.wallet.data.common.deleteDocument
+import com.vultisig.wallet.data.common.saveContentToUri
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+
+fun interface SaveBackupToUriUseCase {
```
rkokhatskyi: `BackupPasswordViewModel.kt:62/246/276/291` and `BackupPasswordRequestViewModel.kt:163/177` still hold `@ApplicationContext` and call `context.saveContentToUri` and `context.deleteDocument` directly while `VaultDetailViewModel.kt:78` still calls `context.share`, so the same `StaticFieldLeak` warning that #3782 tracks will keep firing after merge, whereas the commit message says `Closes #3782`, so either fold those callers into this PR or rescope the issue so the lint baseline matches reality

data/src/main/kotlin/com/vultisig/wallet/data/usecases/file/VaultFileReaderUseCase.kt
```diff
@@ -0,0 +1,33 @@
...
+import android.net.Uri
+import com.vultisig.wallet.data.common.AppZipEntry
+import com.vultisig.wallet.data.common.fileContent
+import com.vultisig.wallet.data.common.fileName
+import com.vultisig.wallet.data.common.isValidZipFile
+import com.vultisig.wallet.data.common.processZip
+import dagger.hilt.android.qualifiers.ApplicationContext
+import javax.inject.Inject
+
+interface VaultFileReaderUseCase {
```
rkokhatskyi: Nothing about this contract is vault specific since `readContent`, `readName`, `isValidZip` and `extractZipEntries` are generic `Uri` operations whereas the only coupling is that `extractZipEntries` filters by `FILE_ALLOWED_EXTENSIONS`, so the `Vault` prefix will steer future non vault import flows into duplicating the same wrapper, while `UriFileReaderUseCase` would keep it discoverable

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
```diff
@@ -54,13 +58,24 @@ import kotlinx.coroutines.launch
...
     startDestination: Any,
     onNavigationReady: () -> Unit,
 ) {
     val foregroundNotification by mainViewModel.foregroundNotification.collectAsStateWithLifecycle()
     var lastNotification by remember { mutableStateOf<ForegroundNotificationState?>(null) }
     if (foregroundNotification != null) lastNotification = foregroundNotification
 
+    val context = LocalContext.current
+    val snackbarState = rememberVsSnackbarState()
+    LaunchedEffect(snackbarFlow) {
```
rkokhatskyi, line 70: `SnackbarFlow.messageFlow` is a rendezvous `Channel()` and the collector used to start from `MainViewModel.init` whereas it now starts only once `MainActivityContent` enters the composition which `MainActivity.kt:108` gates behind `if (showSplash)`, so any `showMessage` invoked during the splash window, from `onNewIntent` before navigation is ready, or across an Activity recreation will suspend the caller until a receiver reattaches, since switching to `Channel(capacity = Channel.BUFFERED)` or a `MutableSharedFlow` with a small replay would close the hole without behavior changes

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
```diff
@@ -54,13 +58,24 @@ import kotlinx.coroutines.launch
...
+    snackbarFlow: SnackbarFlow,
     startDestination: Any,
     onNavigationReady: () -> Unit,
 ) {
     val foregroundNotification by mainViewModel.foregroundNotification.collectAsStateWithLifecycle()
     var lastNotification by remember { mutableStateOf<ForegroundNotificationState?>(null) }
     if (foregroundNotification != null) lastNotification = foregroundNotification
 
+    val context = LocalContext.current
+    val snackbarState = rememberVsSnackbarState()
```
rkokhatskyi: `rememberVsSnackbarState()` ties the visible snackbar progress to the composition lifetime whereas `MainViewModel.snakeBarHostState` previously lived in the retained ViewModel, so a snackbar mid display is silently dropped on every configuration change since `AndroidManifest.xml:55` does not declare `android:configChanges` for `MainActivity` which means rotation, dark mode toggle, font scale or locale change all recreate the activity and blow away the `remember`

data/src/main/kotlin/com/vultisig/wallet/data/usecases/DataUsecasesModule.kt
```diff
@@ -243,4 +249,18 @@ internal interface DataUsecasesModule {
     @Binds
     @Singleton
     fun bindThorchainMemoParser(impl: ThorchainMemoParserImpl): ThorchainMemoParser
+
+    @Binds
+    @Singleton
+    fun bindSaveBackupToUriUseCase(impl: SaveBackupToUriUseCaseImpl): SaveBackupToUriUseCase
```
rkokhatskyi, line 259: None of `SaveBackupToUriUseCase`, `DeleteBackupDocumentUseCase`, `VaultFileReaderUseCase` or `ShareBitmapUseCase` ship with unit tests while `pushNotificationErrorUiText` has nontrivial branching for `PartialFailure` plurals plus a null safe cast that warrants at least a smoke test, so the moved Context calls are now untested even though the project rules require unit tests for new use cases

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/data/usecases/ShareBitmapUseCase.kt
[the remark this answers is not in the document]
aminsato: Fixed in 15b84121b. Added `FLAG_ACTIVITY_NEW_TASK` on the chooser intent in `ShareUtils.kt` so starting the chooser from `ApplicationContext` (via `ShareBitmapUseCase`) no longer crashes.

aminsato: Moved `ShareBitmapUseCase` to `com.vultisig.wallet.ui.usecases` (15b84121b) so the package matches its ui dependency chain. Keeping it in `:app` because `ShareUtils` depends on the app's `R` class and `FileProvider` authority `$packageName.provider`, which would force `:data` to take an app-level dependency.

app/src/main/java/com/vultisig/wallet/data/usecases/ShareBitmapUseCase.kt
aminsato: Done in 15b84121b. `invoke` is now `suspend` and wraps the body in `withContext(Dispatchers.IO)` so the JPEG compression / FileOutputStream work no longer runs on Main. The chooser `startActivity` is safe to dispatch from a background thread now that `FLAG_ACTIVITY_NEW_TASK` is set. Removed the `withContext(Dispatchers.Main)` wrapper at the call site in `TokenAddressQrViewModel`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/backup/SaveBackupToUriUseCase.kt
aminsato: Split in 15b84121b. `DeleteBackupDocumentUseCase` now lives in its own `DeleteBackupDocumentUseCase.kt` next to the save one.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/backup/SaveBackupToUriUseCase.kt
aminsato: Folded into this PR in 15b84121b. `BackupPasswordViewModel`, `BackupPasswordRequestViewModel` and `VaultDetailViewModel` no longer hold `@ApplicationContext`: they consume `SaveBackupToUriUseCase` / `DeleteBackupDocumentUseCase` / `ShareBitmapUseCase` instead. `SaveBackupToUriUseCase` gained a `List<AppZipEntry>` overload so the all-vaults path didn't need a second injection. `Closes #3782` should now match reality.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/file/VaultFileReaderUseCase.kt
aminsato: Renamed to `UriFileReaderUseCase` in 15b84121b (file, class, interface, binding, and call sites in `ImportFileViewModel` + test all updated).

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
aminsato: Switched to `Channel(capacity = Channel.BUFFERED)` in 15b84121b. `showMessage` no longer suspends when there's no active collector, so messages emitted during splash, from `onNewIntent` before navigation is ready, or across an activity recreation buffer until the new collector attaches.

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
aminsato: Moved `VSSnackbarState` back into `MainViewModel` (`VSSnackbarState(1.seconds, viewModelScope)`) in 15b84121b. `MainActivityContent` now reads `mainViewModel.snackbarState` instead of `rememberVsSnackbarState()`, so rotation / locale / dark-mode toggles no longer drop a mid-display snackbar. Context resolution still happens in the composable to keep `Context` out of the ViewModel.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/DataUsecasesModule.kt
aminsato: Added `PushNotificationErrorTextTest` in 15b84121b covering `PartialFailure` plural args (success / total / failure), each `PushNotificationError` subtype's string-resource mapping, the `ApiFailure` branch, and the `as? PushNotificationError` fallback when an unrelated `Throwable` is passed in.

For the four new use-cases (`SaveBackupToUri`, `DeleteBackupDocument`, `UriFileReader`, `ShareBitmap`) I held off — each one is a single-line delegate to an already-tested `Context.*` extension (`saveContentToUri`, `deleteDocument`, `uri.fileContent/fileName/isValidZipFile/processZip`, `Context.share`) with no branching of its own. Happy to add Mockk smoke tests if you want them as a regression net against future changes — let me know.

aminsato:
@coderabbitai re the outside-diff comment on `TokenAddressQrViewModel.shareQRCode` (back() unconditionally on failure): fixed in 15b84121b — now tracks a local `shared` flag and only calls `back()` when `shareBitmap(...)` completes successfully. `CancellationException` is still rethrown and other failures are still logged via `Timber.e` without navigation.

aminsato:
@coderabbitai applied in 91e56fbcb — `PushNotificationErrorTextTest` now uses Kotest matchers (`shouldBe`, `shouldBeInstanceOf`) instead of JUnit assertions, matching the project convention used in `AddressEntryViewModelTest` and friends.

johnnyluo (approved):
LGTM

PR #4592, opened 2026-05-19
feat(chain): add per-chain transaction history button to chain detail

### LINKED ISSUE #4564: Chain detail: missing tx history access button (per Figma, filtered for chain)
## Summary

According to Figma, the chain detail screen should have a transaction-history access button alongside Swap / Send / Buy / Receive, opening a tx history view filtered for the current chain. Currently missing on Android.

## Probable affected area

- `app/src/main/java/com/vultisig/wallet/ui/screens/v2/chaintokens/ChainTokensScreen.kt:203-240` — action button row (Swap, Send, Buy, Receive) — no history entry.
- ViewModel call site around `ChainTokensScreen.kt:83-88` — wire-up to add a history action handler.

Likely existing TransactionHistory route exists already (it's used from the global wallet header) — should be reusable filtered by `chain`.

## Repro

1. Open any vault → tap any chain (e.g. Solana).
2. Observe the action row: Swap / Send / Buy / Receive only.
3. Compare to Figma — history entry expected.

## Platform

Android. Cross-platform check not done.

aminsato:
Closes #4564

## Summary
- Adds a **History** action button to the chain-detail action row (alongside Swap / Send / Buy / Functions / Receive). Tapping it opens the transaction history filtered to the current chain.
- Renames `TransactionTypeButton` → `AssetActionButton` (and `TransactionType` → `AssetAction`) — History and Receive aren't transactions, they're asset-level actions.
- `Route.TransactionHistory` gains an optional `chainId: String? = null`. The existing global-header entry point keeps its no-arg behavior; the new chain-detail entry passes the chain.
- Chain scoping is enforced at the **DAO level** via new `observeAllByVaultAndChain` / `observeSendByVaultAndChain` / `observeSwapByVaultAndChain` queries (the `chain` column is already indexed, so this is index-served).
- The chain-detail action row uses a small custom `Layout` (`AssetActionRow`) that greedy-packs items into one row when the device width allows and wraps when it doesn't. If wrapping would leave a single item on the last row, it pulls one from the row above so the last row always has at least two items.

## Test plan
- [ ] Open a vault → tap any chain → confirm the new History button appears as the last item in the action row.
- [ ] Tap History → transaction history screen opens showing only this chain's transactions.
- [ ] Open the global wallet header → History entry still shows transactions from all chains.
- [ ] On a narrow device / small emulator, confirm the action row wraps cleanly (last row has ≥ 2 items, no single orphan).
- [ ] On a wider device, confirm all action buttons fit in a single row.
- [ ] Verify all 10 locales render the new `transaction_type_button_history` label.

## Summary by CodeRabbit

* **New Features**
  * Added a History button on chain dashboards for quick access to transaction history.
  * Transaction History can be scoped to a specific chain for filtered views.

* **UI Updates**
  * Unified action buttons (Swap, Send, Receive, Buy, Functions) across screens with updated visuals and interaction behavior.

* **Localization**
  * New localized chain-specific Transaction History title strings added for multiple languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4592?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
@@ -151,8 +151,9 @@ constructor(
     private val navigator: Navigator<Destination>,
 ) : ViewModel() {
 
-    private val vaultId: String = savedStateHandle.toRoute<Route.TransactionHistory>().vaultId
-    private val chainId: String? = savedStateHandle.toRoute<Route.TransactionHistory>().chainId
+    private val route: Route.TransactionHistory = savedStateHandle.toRoute()
+    private val vaultId: String = route.vaultId
+    private val chainId: String? = route.chainId
 
     private val currentTime = MutableStateFlow(System.currentTimeMillis())
 
@@ -233,7 +234,7 @@ constructor(
         ) {
             _uiState.update { it.copy(isRefreshing = true) }
             try {
-                refreshPendingTransactions(vaultId)
+                refreshPendingTransactions(vaultId, chainId)
                 delay(100.milliseconds) // prevent refresh ui freezing
             } finally {
                 _uiState.update { it.copy(isRefreshing = false) }
@@ -242,7 +243,7 @@ constructor(
     }
 
     private fun refreshOnEnter() {
-        viewModelScope.safeLaunch { refreshPendingTransactions(vaultId) }
+        viewModelScope.safeLaunch { refreshPendingTransactions(vaultId, chainId) }
     }
 
     private fun startTimeTicker() {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
@@ -152,6 +152,7 @@ constructor(
 ) : ViewModel() {
 
     private val vaultId: String = savedStateHandle.toRoute<Route.TransactionHistory>().vaultId
+    private val chainId: String? = savedStateHandle.toRoute<Route.TransactionHistory>().chainId
 
     private val currentTime = MutableStateFlow(System.currentTimeMillis())
 
@@ -263,6 +264,7 @@ constructor(
                     transactionHistoryRepository.observeTransactions(
                         vaultId = vaultId,
                         type = tab.toRepositoryType(),
+                        chain = chainId,
                     )
                 }
                 .combine(currentTime) { entities, now ->
@@ -293,7 +295,11 @@ constructor(
     private fun observeAssetSearchItems() {
         viewModelScope.launch {
             transactionHistoryRepository
-                .observeTransactions(vaultId = vaultId, type = TransactionHistoryType.OVERVIEW)
+                .observeTransactions(
+                    vaultId = vaultId,
+                    type = TransactionHistoryType.OVERVIEW,
+                    chain = chainId,
+                )
                 .map { entities ->
                     entities
                         .flatMap { entity ->
--- a/app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
@@ -139,6 +139,7 @@ data class TransactionHistoryUiState(
     val assetSearchItems: List<TransactionAssetUiModel> = emptyList(),
     val selectedAssetIds: Set<String> = emptySet(),
     val selectedAssets: List<TransactionAssetUiModel> = emptyList(),
+    val chainName: String? = null,
 )
 
 @HiltViewModel
@@ -153,13 +154,13 @@ constructor(
 
     private val route: Route.TransactionHistory = savedStateHandle.toRoute()
     private val vaultId: String = route.vaultId
-    private val chainId: String? = route.chainId
+    private val chainId: String? = route.chainId?.takeIf { it.isNotBlank() }
 
     private val currentTime = MutableStateFlow(System.currentTimeMillis())
 
     val assetSearchTextFieldState = TextFieldState()
 
-    private val _uiState = MutableStateFlow(TransactionHistoryUiState())
+    private val _uiState = MutableStateFlow(TransactionHistoryUiState(chainName = chainId))
     val uiState: StateFlow<TransactionHistoryUiState> = _uiState.asStateFlow()
 
     init {
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/chaintokens/ChainTokensScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/chaintokens/ChainTokensScreen.kt
@@ -24,9 +24,12 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.blur
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.layout.Layout
+import androidx.compose.ui.layout.Placeable
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 import androidx.hilt.navigation.compose.hiltViewModel
 import com.google.android.play.core.review.ReviewManagerFactory
@@ -53,9 +56,9 @@ import com.vultisig.wallet.ui.screens.ResourceTwoCardsRow
 import com.vultisig.wallet.ui.screens.v2.chaintokens.components.ChainAccount
 import com.vultisig.wallet.ui.screens.v2.chaintokens.components.ChainLogo
 import com.vultisig.wallet.ui.screens.v2.chaintokens.components.ChainTokensTabMenuAndSearchBar
+import com.vultisig.wallet.ui.screens.v2.home.components.AssetAction
+import com.vultisig.wallet.ui.screens.v2.home.components.AssetActionButton
 import com.vultisig.wallet.ui.screens.v2.home.components.CopiableAddress
-import com.vultisig.wallet.ui.screens.v2.home.components.TransactionType
-import com.vultisig.wallet.ui.screens.v2.home.components.TransactionTypeButton
 import com.vultisig.wallet.ui.theme.Theme
 import com.vultisig.wallet.ui.utils.KeyboardAware
 import com.vultisig.wallet.ui.utils.VsUriHandler
@@ -85,6 +88,7 @@ internal fun ChainTokensScreen(
         onDeposit = viewModel::deposit,
         onBuy = viewModel::buy,
         onReceive = viewModel::openAddressQr,
+        onHistory = viewModel::history,
         onSelectTokens = viewModel::selectTokens,
         onTokenClick = viewModel::openToken,
         onHideSearchBar = viewModel::hideSearchBar,
@@ -106,6 +110,7 @@ internal fun ChainTokensScreen(
     onBuy: () -> Unit,
     onDeposit: () -> Unit,
     onReceive: () -> Unit,
+    onHistory: () -> Unit,
     onSelectTokens: () -> Unit,
     onTokenClick: (ChainTokenUiModel) -> Unit,
     onShowReviewPopUp: () -> Unit,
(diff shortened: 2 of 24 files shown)


aminsato:
@coderabbitai full review

johnnyluo:
CleanShot 2026-05-19 at 20 10 26@2x

Too many buttons squeeze together , doesn't feel right

johnnyluo:
@aminsato please have a discussion with Felix and Designer to see what's the best way

aminsato:
> @aminsato please have a discussion with Felix and Designer to see what's the best way

Sure, I'll discuss

realpaaao:
@aminsato We already have that in iOS. Just add the Transaction history button from main screen to each chain detail and filter for chain.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TransactionHistoryRepository.kt
```diff
@@ -89,11 +90,20 @@ class TransactionHistoryRepositoryImpl @Inject constructor(private val dao: Tran
     override fun observeTransactions(
         vaultId: String,
         type: TransactionHistoryType,
+        chain: String?,
     ): Flow<List<TransactionHistoryEntity>> =
-        when (type) {
-            TransactionHistoryType.OVERVIEW -> dao.observeAllByVault(vaultId)
-            TransactionHistoryType.SEND -> dao.observeSendByVault(vaultId)
-            TransactionHistoryType.SWAPS -> dao.observeSwapByVault(vaultId)
+        if (chain == null) {
```
rkokhatskyi: The `chain == null` branch routes empty and blank strings into the chain scoped path which then runs `WHERE chain = ''` and silently returns zero rows, so any future deep link or serialized state that round trips `""` through `Route.TransactionHistory(chainId = "")` looks like an empty vault rather than falling back to the unscoped query

```kotlin
val effectiveChain = chain?.takeIf { it.isNotBlank() }
if (effectiveChain == null) {
    when (type) {
        TransactionHistoryType.OVERVIEW -> dao.observeAllByVault(vaultId)
        TransactionHistoryType.SEND -> dao.observeSendByVault(vaultId)
        TransactionHistoryType.SWAPS -> dao.observeSwapByVault(vaultId)
    }
} else {
    when (type) {
        TransactionHistoryType.OVERVIEW -> dao.observeAllByVaultAndChain(vaultId, effectiveChain)
        TransactionHistoryType.SEND -> dao.observeSendByVaultAndChain(vaultId, effectiveChain)
        TransactionHistoryType.SWAPS -> dao.observeSwapByVaultAndChain(vaultId, effectiveChain)
    }
}
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TransactionHistoryRepository.kt
```diff
@@ -89,11 +90,20 @@ class TransactionHistoryRepositoryImpl @Inject constructor(private val dao: Tran
     override fun observeTransactions(
         vaultId: String,
         type: TransactionHistoryType,
+        chain: String?,
```
rkokhatskyi, line 93: Adding the third parameter silently breaks the stub at `TransactionHistoryViewModelTest.kt:207` where `every { observeTransactions(any(), any()) } returns flowOf(emptyList())` no longer matches the call site so the relaxed mock returns an unrelated default empty flow which makes that test a no op and masks regressions in the very code path this PR touches

```kotlin
every { transactionHistoryRepository.observeTransactions(any(), any(), any()) } returns flowOf(emptyList())
```

data/src/main/kotlin/com/vultisig/wallet/data/db/dao/TransactionDao.kt
```diff
@@ -46,6 +70,18 @@ abstract class TransactionHistoryDao {
...
     abstract fun observeSwapByVault(vaultId: String): Flow<List<TransactionHistoryEntity>>
 
+    @Query(
+        """
+        SELECT * FROM transaction_history
+        WHERE vaultId = :vaultId AND chain = :chain AND type = 'SWAP'
+        ORDER BY timestamp DESC
+    """
+    )
+    abstract fun observeSwapByVaultAndChain(
```
rkokhatskyi: Cross chain swaps persist only the source chain on the `chain` column (see `KeysignViewModel.kt:829` which writes `chain = payload.coin.chain.raw`), so `observeSwapByVaultAndChain(vaultId, "Ethereum")` will not surface a Bitcoin to Ethereum swap that users typically track on the receiving side. Either union the `toChain` JSON payload value, persist a second `destinationChain` column and index it, or KDoc this method as source chain only so the asymmetry is explicit

data/src/main/kotlin/com/vultisig/wallet/data/db/dao/TransactionDao.kt
```diff
@@ -28,6 +28,18 @@ abstract class TransactionHistoryDao {
...
     abstract fun observeAllByVault(vaultId: String): Flow<List<TransactionHistoryEntity>>
 
+    @Query(
+        """
+        SELECT * FROM transaction_history
+        WHERE vaultId = :vaultId AND chain = :chain
+        ORDER BY timestamp DESC
+    """
+    )
+    abstract fun observeAllByVaultAndChain(
```
rkokhatskyi: These six observers collapse to three using Room's nullable bind pattern which still uses the `chain` index when bound to a concrete value and also drops the matching `if (chain == null)` branch in the repository

```kotlin
@Query("""
    SELECT * FROM transaction_history
    WHERE vaultId = :vaultId
      AND (:chain IS NULL OR chain = :chain)
    ORDER BY timestamp DESC
""")
abstract fun observeAllByVault(vaultId: String, chain: String?): Flow<List<TransactionHistoryEntity>>
```

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -152,6 +152,7 @@ constructor(
 ) : ViewModel() {
 
     private val vaultId: String = savedStateHandle.toRoute<Route.TransactionHistory>().vaultId
+    private val chainId: String? = savedStateHandle.toRoute<Route.TransactionHistory>().chainId
```
rkokhatskyi: `savedStateHandle.toRoute<Route.TransactionHistory>()` is decoded twice back to back which is wasteful and brittle against future mock patterns that swap to `returnsMany`, so destructure once

```kotlin
private val route: Route.TransactionHistory = savedStateHandle.toRoute()
private val vaultId: String = route.vaultId
private val chainId: String? = route.chainId
```

rkokhatskyi: Nothing in `TransactionHistoryScreen.kt` signals that the list is filtered to one chain since the scaffold title stays hardcoded to `transaction_history_title`, so an empty Bitcoin scoped list looks identical to no transactions at all in the vault and the user has no affordance to clear the scope. Surface `chainId` in the state and either fold the chain name into the title or render an unremovable header chip next to `SelectedAssetFiltersRow`

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -263,6 +264,7 @@ constructor(
                     transactionHistoryRepository.observeTransactions(
                         vaultId = vaultId,
                         type = tab.toRepositoryType(),
+                        chain = chainId,
                     )
```
rkokhatskyi, line 270: Introducing `chainId` here makes the list chain scoped but `refreshPendingTransactions(vaultId)` on lines 236 and 245 still polls indexers and mutates retry counters for every chain in the vault, which couples the freshness of the chain the user is watching to slow indexers on chains they are not, and on a vault with many active chains burns network and battery for nothing. Add a chain scoped variant of `RefreshPendingTransactionsUseCase` plus a `getPendingByVaultAndChain` on the dao, and route through it when `chainId != null`

app/src/main/java/com/vultisig/wallet/ui/screens/v2/chaintokens/ChainTokensScreen.kt
```diff
@@ -306,6 +310,69 @@ internal fun ChainTokensScreen(
...
+        val rowHeights = rows.map { row -> row.maxOf { it.height } }
+        val totalHeight = rowHeights.sum() + vGap * (rows.size - 1).coerceAtLeast(0)
+
+        layout(outerWidth, totalHeight) {
+            var y = 0
+            rows.forEachIndexed { idx, row ->
+                val rowWidth = row.sumOf { it.width } + hGap * (row.size - 1).coerceAtLeast(0)
+                var x = ((outerWidth - rowWidth) / 2).coerceAtLeast(0)
+                row.forEach { p ->
+                    p.place(x, y)
```
rkokhatskyi: `p.place(x, y)` ignores `LocalLayoutDirection` whereas the previous `Row` with `Arrangement.spacedBy(.., Alignment.CenterHorizontally)` mirrored automatically in RTL contexts, so any future Arabic or Hebrew locale, or a deliberate `LocalLayoutDirection provides Rtl` preview, will render this row in the wrong order

```kotlin
row.forEach { p ->
    p.placeRelative(x, y)
    x += p.width + hGap
}
```

rkokhatskyi (changes requested):

johnnyluo:
@aminsato IOS & MacOS add the button on the top right corner

CleanShot 2026-05-20 at 08 56 01@2x

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TransactionHistoryRepository.kt
aminsato: Good catch — fixed in f4ff874f6. The repository now coerces blank into `null` via `chain?.takeIf { it.isNotBlank() }` before routing, so an empty `chainId` deserialized through `Route.TransactionHistory` falls back to the unscoped query instead of `WHERE chain = ''`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TransactionHistoryRepository.kt
aminsato: Fixed in f4ff874f6 — the stub now uses `any(), any(), any()` to match the 3-arg signature so the relaxed mock no longer silently swallows the call.

data/src/main/kotlin/com/vultisig/wallet/data/db/dao/TransactionDao.kt
aminsato: Documented in f4ff874f6 — added a KDoc on `observeSwapByVault` calling out the source-chain-only behavior and pointing at the `KeysignViewModel` write site. Going with the doc route for this PR since adding a `destinationChain` column or unioning the JSON payload is a wider data-model change that deserves its own ticket; happy to follow up if you'd prefer the schema change here.

data/src/main/kotlin/com/vultisig/wallet/data/db/dao/TransactionDao.kt
aminsato: Done in f4ff874f6 — collapsed the 6 observers to 3 with the `(:chain IS NULL OR chain = :chain)` pattern, and the repository's `if (chain == null)` branch is gone.

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
aminsato: Fixed in f4ff874f6.

aminsato: Addressed by 3fb3f96f4 (different angle): the per-chain history entry point is now a `VsCircleButton` placed next to the explorer button in the chain-detail top bar, instead of the action-row tile. The user lands on the history screen already inside that chain's context, so the screen still keeps the global title — let me know if you'd still like an in-screen chain chip on top of that.

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
aminsato: Fixed in f4ff874f6 — `RefreshPendingTransactionsUseCase.invoke` now takes an optional `chain: String?` and short-circuits non-matching rows before the indexer fan-out, so the chain-scoped view no longer pays for unrelated chains. Kept it at the use-case level rather than adding a `getPendingByVaultAndChain` DAO method since the pending set is small and the cost being burned is the network fan-out, not the DB read — happy to push the predicate down to SQL if you'd prefer.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/chaintokens/ChainTokensScreen.kt
aminsato: Fixed in f4ff874f6 — switched to `placeRelative` so the row mirrors automatically under RTL.

(Note: in 3fb3f96f4 the custom `AssetActionRow` Layout was replaced with a plain `Row` anyway, so the call site is gone — but the placeRelative fix is preserved in the history before the removal commit.)

app/src/main/java/com/vultisig/wallet/ui/screens/v2/chaintokens/ChainTokensScreen.kt
```diff
@@ -142,14 +144,26 @@ internal fun ChainTokensScreen(
...
+                        if (uiModel.explorerURL.isNotEmpty()) {
+                            VsCircleButton(
+                                onClick = { uriHandler.openUri(uiModel.explorerURL) },
+                                size = VsCircleButtonSize.Small,
+                                icon = R.drawable.explor,
+                                type = VsCircleButtonType.Secondary,
+                                designType = DesignType.Shined,
+                            )
+                        }
                     }
```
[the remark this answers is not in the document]
aminsato: Not applying this — the placement was changed intentionally by design direction in 3fb3f96f4. History now lives next to the explorer button in the top bar (since both are passive "view" actions on the chain context), while the action row stays focused on transactional verbs (Swap/Send/Buy/Functions/Receive). Treating the original ticket's wording as a fixed contract isn't right here.

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
rkokhatskyi: This one is still relevant @aminsato

rkokhatskyi (dismissed):
@aminsato check one last unresolved thread

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
aminsato: You're right — the entry-point change didn't address the underlying signal problem. Fixed in 6dbcc656f: surfaced `chainName` in `TransactionHistoryUiState` (seeded from `route.chainId?.takeIf { it.isNotBlank() }`) and folded it into the scaffold title via a new `transaction_history_title_chain` string ("%1$s History" — e.g. "Bitcoin History"), translated across all 10 locales. An empty Bitcoin-scoped list now reads "Bitcoin History · No transactions yet" instead of looking identical to the vault-wide empty state. Added two `@Preview`s (`PreviewChainScoped` and `PreviewChainScopedEmpty`) to lock that in.

johnnyluo (approved):
LGTM

PR #4593, opened 2026-05-19
fix(keygen): block destructive fast-vault overwrite on seedphrase import (#4535)

### LINKED ISSUE #4535: [Discussion] Fast vault collision UX is confusing on seedphrase import — negate vs new-share-on-server
## Summary

When a user imports a seedphrase and then chooses **Fast Vault** in setup, we hit the server to check whether a fast vault already exists for that seedphrase. If yes, we surface a confusing destructive-overwrite warning instead of treating it as a normal state.

Current copy:
> **Fast vault already exists**
> A fast vault with this seed phrase already exists on the server. Creating a new fast vault will replace the server share and make the existing vault unusable. Consider using a secure vault instead.
> [Back] [Continue]

The "Continue" path silently bricks the existing vault on the server. That's a footgun, not a UX.

## What's wrong

1. **Destructive by default** — the recovery path for a user who lost their client share but kept their seedphrase becomes "destroy your existing setup". For a wallet, "make existing vault unusable" is the worst possible default sentence to put behind a Continue button.
2. **No clear "happy path"** — the user wanted to import → restore. The dialog implies the only way forward is destruction, with a soft nudge toward secure vault (which doesn't help anyone who specifically wants fast vault).
3. **Asymmetric across platforms** — see "Cross-platform check" below; iOS appears not to gate this at all, which is even worse (silent overwrite, no warning).

### Option A — Negate at seedphrase entry

If a fast vault already exists on the server for this seedphrase, **block the fast-vault path entirely** at the seedphrase step (one screen earlier than today) and route the user to:
- "We found a fast vault for this seedphrase — recover the existing one instead", OR
- "Choose Secure Vault" if they really want a new MPC setup.

Pros: fail-fast, never destructive, clear mental model.
Cons: user who legitimately wants to rotate/replace their server share has no path. Would need a separate "rotate server share" flow.

### Option B — New share on the server (non-destructive)

Server stores **N shares per seedphrase**, keyed by `(seedphrase_hash, vault_id)`. Importing a seedphrase + creating a new fast vault simply allocates a new server share row — existing vault stays usable.

Pros: never destructive. Removes the dialog entirely. User can have multiple fast vaults from one seedphrase (e.g., per device / per persona).
Cons: server schema change (`UNIQUE(seedphrase_hash)` → composite key), share-storage scaling, billing model needs to handle N shares per user.

## Affected code (Android, today)

- Trigger: `app/src/main/java/com/vultisig/wallet/ui/models/v3/onboarding/EnterVaultInfoViewModel.kt:481-495`
  ```kotlin
  private fun navigateToNextScreen() {
      ...
      if (!isSecureVault && tssAction == TssAction.KeyImport) {
          val existsOnServer = checkServerVaultExistence()
          if (existsOnServer) {
              uiState.update { it.copy(showServerVaultExistsWarning = true) }
              return@launch
          }
      }
      navigateToPeerDiscovery(name, email, password)
  }
  ```
- Dialog: `EnterVaultInfoScreen.kt:312-348` (`ServerVaultExistsWarningDialog`)
- Strings: `app/src/main/res/values/strings.xml:924-925` (`name_vault_server_exists_warning_title` / `_message`)
- Use case: `CheckServerVaultExistsUseCase` (injected at `EnterVaultInfoViewModel.kt:153`)
- Continue handler: `EnterVaultInfoViewModel.kt:540` (`continueWithServerVaultWarning` → routes to `Route.Keygen.PeerDiscovery` which then triggers the overwriting keygen).

## Cross-platform check (please verify before settling on a fix)

- **Android**: gated by the warning above (this issue).
- **iOS**: a quick grep on `vultisig-ios` shows no equivalent `checkServerVaultExists` or "Fast vault already exists" string. Either iOS silently overwrites (worse), or the seedphrase → fast-vault path isn't routable at all. Needs a separate ticket once we decide on direction.
- **Windows**: not checked here.

Whichever option we land on, the server-side semantics need to match across all three clients — otherwise we get the same data-loss bug, just routed differently per platform.

## Asks

1. Decide between **Option A (negate)** and **Option B (new share)**. My preference is B because A still leaves the "rotate server share" use case unsolved and forces us to build a parallel flow anyway.
2. If B: design the server-side schema/billing change and a migration for already-overwritten users (if any are reachable).
3. Either way: kill the current destructive-Continue dialog on Android, and file the equivalent ticket on iOS once the server semantics are decided.

## Platform

Discussion is cross-platform. UI ticket is Android (where the dialog lives today).

aminsato:
## Summary

Closes #4535 (Option A — Negate).

When a user imports a seedphrase and chooses Fast Vault, the app checks whether a fast vault already exists on the server for that seedphrase. Previously the dialog offered a **Continue** button that silently overwrote the server share and made the existing vault unusable. This PR removes that path entirely.

- `ServerVaultExistsWarningDialog` now has a single Back button (no Continue).
- `dismissServerVaultWarning()` clears the flag **and** navigates back so the user lands on the vault-type select screen, where they can choose Secure Vault or go back further to recover the existing vault.
- Updated copy in all 10 locales to non-destructive guidance: _"A fast vault for this seed phrase already exists on the server. Go back and recover the existing vault, or choose Secure Vault for a new MPC setup."_
- Removed the now-unused `name_vault_server_exists_warning_continue` string from every locale.
- Updated the corresponding ViewModel unit test.

## Out of scope

- iOS parity — needs a separate ticket per the issue.
- Option B (N shares per seedphrase) — requires server schema + billing changes and is a cross-platform decision.
- Moving the check one screen earlier (to the seedphrase entry screen) — the path is still blocked, just one screen later; the destructive default is gone either way.

## Test plan

- [ ] Import a seedphrase that already has a fast vault on the server → pick Fast Vault → expect the dialog with only a Back button. Tapping Back returns to the vault-type select screen.
- [ ] Import a seedphrase with no existing fast vault on the server → pick Fast Vault → expect the existing happy path (no dialog).
- [ ] Secure Vault path is unaffected.
- [ ] Verify all 10 locales render the new message without truncation.
- [ ] `./gradlew :app:testDebugUnitTest --tests "*EnterVaultInfoViewModelTest*"` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Modified the server vault exists warning behavior to require users to dismiss and navigate back instead of proceeding with the warning.
  * Updated warning message across all supported languages to clarify recovery options when a fast vault already exists on the server for the given seed phrase.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4593?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/v3/onboarding/EnterVaultInfoViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/v3/onboarding/EnterVaultInfoViewModel.kt
@@ -534,14 +534,7 @@ constructor(
 
     fun dismissServerVaultWarning() {
         uiState.update { it.copy(showServerVaultExistsWarning = false) }
-    }
-
-    fun continueWithServerVaultWarning() {
-        uiState.update { it.copy(showServerVaultExistsWarning = false) }
-        val name = nameTextFieldState.text.toString()
-        val email = emailTextFieldState.text.toString()
-        val password = passwordTextFieldState.text.toString()
-        viewModelScope.launch { navigateToPeerDiscovery(name, email, password) }
+        viewModelScope.launch { navigator.back() }
     }
 
     private fun isNameValid(): Boolean {
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v3/onboarding/EnterVaultInfoScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v3/onboarding/EnterVaultInfoScreen.kt
@@ -80,10 +80,7 @@ internal fun EnterVaultInfoScreen(viewModel: EnterVaultInfoViewModel = hiltViewM
     )
 
     if (uiState.showServerVaultExistsWarning) {
-        ServerVaultExistsWarningDialog(
-            onContinue = viewModel::continueWithServerVaultWarning,
-            onDismiss = viewModel::dismissServerVaultWarning,
-        )
+        ServerVaultExistsWarningDialog(onDismiss = viewModel::dismissServerVaultWarning)
     }
 
     if (showReferralSheet) {
@@ -306,7 +303,7 @@ fun StepIcon(type: StepType, state: StepState) {
 }
 
 @Composable
-private fun ServerVaultExistsWarningDialog(onContinue: () -> Unit, onDismiss: () -> Unit) {
+private fun ServerVaultExistsWarningDialog(onDismiss: () -> Unit) {
     AlertDialog(
         containerColor = Theme.v2.colors.backgrounds.secondary,
         onDismissRequest = onDismiss,
@@ -325,15 +322,6 @@ private fun ServerVaultExistsWarningDialog(onContinue: () -> Unit, onDismiss: ()
             )
         },
         confirmButton = {
-            TextButton(onClick = onContinue) {
-                Text(
-                    text = stringResource(R.string.name_vault_server_exists_warning_continue),
-                    color = Theme.v2.colors.neutrals.n100,
-                    style = Theme.montserrat.body3,
-                )
-            }
-        },
-        dismissButton = {
             TextButton(onClick = onDismiss) {
                 Text(
                     text = stringResource(R.string.name_vault_server_exists_warning_back),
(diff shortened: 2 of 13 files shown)


aminsato:
@coderabbiti full review

johnnyluo (approved):
LGTM

PR #4595, opened 2026-05-19
fix(peer-discovery): show "X/4+" for 4+ device vaults

### LINKED ISSUE #4553: Peer discovery: "Devices (X/4)" should be "X/4+" when 4+ device vault is selected
# [Fix] Peer discovery device counter misleadingly shows fixed cap for 4+ vaults [PeerDiscoveryScreen]

## Problem
On the keygen peer-discovery screen, the counter renders as `Devices (1/4)` when the user selected a 4+ device vault. The `4` is a minimum threshold, not a cap, but the slash format reads as "1 of 4 max." Users can continue adding peers beyond the minimum, but the UI implies the cap is reached when it hits `4/4`.

## Expected Behavior
When the selected vault requires 4 or more devices, the counter should display `Devices (1/4+)` to match the "4+" framing already used in the vault setup chooser. For 2-of-2 and 3-of-3 vaults, the existing `X/X` format is correct (those are fixed thresholds, not minimums) and must not change.

## Steps to Reproduce
1. Setup new vault → Secure → choose the 4-device vault option.
2. Land on the peer-discovery screen with the QR code.
3. Observe `Devices (1/4)`. Add peers and observe `2/4`, `3/4`, `4/4`.
4. Note there is no visual signal that peers beyond 4 are accepted.

### Must Do
- [ ] Add `peer_discovery_devices_n_of_n_plus` string key (`Devices (%1$s/%2$s+)`) to `app/src/main/res/values/strings.xml` and all locale mirrors (e.g. `values-ru/strings.xml`).
- [ ] In `PeerDiscoveryScreen.kt:278-285`, conditionally select the new string resource when `state.minimumDevicesDisplayed >= 4`; keep the existing string for values below 4.
- [ ] Confirm with engineering whether the MPC session backend caps participants at exactly `deviceCount` or allows more — if capped, the `+` label would be misleading and this change should be reconsidered before shipping.

### Must NOT Do
- Don't change unrelated UI or layout code.
- Don't apply the `+` suffix for 2-of-2 or 3-of-3 vault types — those thresholds are fixed.
- Don't rename or remove the existing `peer_discovery_devices_n_of_n` key; it is still used for non-plus cases.

## Acceptance Criteria
- [ ] On a 4-device vault, peer-discovery screen shows `Devices (1/4+)` and increments correctly: `2/4+`, `3/4+`, `4/4+`.
- [ ] On a 2-device vault, peer-discovery screen still shows `Devices (1/2)` (no `+`).
- [ ] On a 3-device vault, peer-discovery screen still shows `Devices (1/3)` (no `+`).
- [ ] All locale string files include the new key.

## Context
**Affected files:**
- `app/src/main/res/values/strings.xml:537` — existing string `peer_discovery_devices_n_of_n`
- `app/src/main/res/values-ru/strings.xml:829` — Russian locale mirror (other locales also affected)
- `app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt:278-285` — render site
- `app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt:90,149-150` — `minimumDevicesDisplayed` source; `MIN_KEYGEN_DEVICES = 2`

**Engineering flag:** The backend MPC committee behavior (whether it hard-caps at exactly `deviceCount` or truly accepts more) should be verified before this label change ships. If the session is capped at N, the `+` suffix is user-facing misinformation.

**Platform:** Android only. iOS/Windows likely have the same pattern but are explicitly out of scope.

johnnyluo:
## Summary
- Adds `peer_discovery_devices_n_of_n_plus` string (`Devices (%1$s/%2$s+)`) to all 10 locale files
- `PeerDiscoveryScreen` now selects the `+` variant when `minimumDevicesDisplayed >= 4`, keeping the existing format for 2-of-2 and 3-of-3 vaults unchanged

## Issue
Closes #4553

## Test Plan
- [x] On a 4-device vault, peer-discovery screen shows `Devices (1/4+)` and increments correctly: `2/4+`, `3/4+`, `4/4+`
- [x] On a 2-device vault, peer-discovery screen still shows `Devices (1/2)` (no `+`)
- [x] On a 3-device vault, peer-discovery screen still shows `Devices (1/3)` (no `+`)
- [x] Debug build succeeds (`./gradlew assembleDebug`) ✅

## Summary by CodeRabbit

* **New Features**
  * Peer discovery now shows an "N of N+" device-count variant when additional devices can be added, improving visibility of device status.

* **Internationalization**
  * Added translations for the "N of N+" device-count label in German, Spanish, Croatian, Italian, Korean, Dutch, Portuguese, Russian, and Simplified Chinese.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4595?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -99,6 +99,7 @@ data class PeerDiscoveryUiModel(
     val selectedDevices: List<String> = emptyList(),
     val minimumDevices: Int = MIN_KEYGEN_DEVICES,
     val minimumDevicesDisplayed: Int = MIN_KEYGEN_DEVICES,
+    val allowsMoreDevices: Boolean = false,
     val showQrHelpModal: Boolean = false,
     val showDevicesHint: Boolean = true,
     val connectingToServer: ConnectingToServerUiModel? = null,
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt
@@ -277,7 +277,7 @@ internal fun PeerDiscoveryScreen(
                         Text(
                             text =
                                 stringResource(
-                                    if (state.minimumDevicesDisplayed >= 4)
+                                    if (state.allowsMoreDevices)
                                         R.string.peer_discovery_devices_n_of_n_plus
                                     else R.string.peer_discovery_devices_n_of_n,
                                     selectedDevicesSize,
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt
@@ -277,7 +277,9 @@ internal fun PeerDiscoveryScreen(
                         Text(
                             text =
                                 stringResource(
-                                    R.string.peer_discovery_devices_n_of_n,
+                                    if (state.minimumDevicesDisplayed >= 4)
+                                        R.string.peer_discovery_devices_n_of_n_plus
+                                    else R.string.peer_discovery_devices_n_of_n,
                                     selectedDevicesSize,
                                     state.minimumDevicesDisplayed,
                                 ),
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
@@ -155,6 +155,7 @@ internal fun KeysignPeerDiscovery(
                     selectedDevices = selectionState.filter { it != vault.localPartyID },
                     minimumDevices = minimumDevices,
                     minimumDevicesDisplayed = minimumDevices,
+                    allowsMoreDevices = vault.signers.size > minimumDevices,
                     showQrHelpModal = false,
                     showDevicesHint = false,
                     connectingToServer = null,
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -812,6 +812,7 @@
     <string name="peer_discovery_topbar_title">QR-Code scannen</string>
     <string name="peer_discovery_waiting_for_devices_action">Geräte werden angefordert…</string>
     <string name="peer_discovery_devices_n_of_n">Geräte (%1$s/%2$s)</string>
+    <string name="peer_discovery_devices_n_of_n_plus">Geräte (%1$s/%2$s+)</string>
     <string name="vault_backup_screen_email_sent_to">An %s gesendet.</string>
     <string name="vault_backup_screen_donot_receive_email">Keine E-Mail erhalten?</string>
     <string name="onboarding_desc_page_5_part_2">separat in einem</string>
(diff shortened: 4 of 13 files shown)


app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt
```diff
@@ -277,7 +277,9 @@ internal fun PeerDiscoveryScreen(
                         Text(
                             text =
                                 stringResource(
-                                    R.string.peer_discovery_devices_n_of_n,
+                                    if (state.minimumDevicesDisplayed >= 4)
```
rkokhatskyi: The cap in `KeygenPeerDiscoveryViewModel.selectDevice` at line 282 blocks any non-Reshare action from accepting more than `minimumDevices - 1` other devices and the auto-select branch at line 542 caps the same way, so for new keygen where `EnterVaultInfoViewModel` coerces `deviceCount` into `1..4` the `>= 4` branch only fires at exactly 4 which is also the hard cap, while for `Migrate` and `SingleKeygen` the displayed minimum equals `signers.size` which is again the cap, which means the suffix advertises room that this client itself refuses to use. Only the keysign path through `KeysignFlowViewModel.handleParticipant` has no cap so the predicate likely needs to consult `args.action` or have each VM publish an `allowsMoreDevices` flag on `PeerDiscoveryUiModel` rather than keying off `minimumDevicesDisplayed` alone.

rkokhatskyi: The bare `4` here mirrors `EnterVaultInfoViewModel.kt:159` where `coerceIn(1, 4)` pins the keygen device-count ceiling and iOS pins the same boundary to `3` in `PeerDiscoveryScreen.swift` via `totalDeviceCount <= 3`, so a named constant or a state flag would stop the screen predicate from drifting silently if the input range changes.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt
johnnyluo: Fixed — added `allowsMoreDevices: Boolean = false` to `PeerDiscoveryUiModel`. The keysign path sets it to `vault.signers.size > minimumDevices` (true only when extra vault members beyond the threshold can participate). Keygen/migrate/singlekeygen paths leave it `false` since `selectDevice` enforces a hard cap. The screen now keys off this flag instead of the `minimumDevicesDisplayed >= 4` predicate.

johnnyluo: Addressed by the `allowsMoreDevices` flag — the bare `4` is gone from the screen predicate entirely. The keysign path derives the flag from `vault.signers.size > minimumDevices` so any future change to the ceiling in EnterVaultInfoViewModel has no silent effect on the screen string selection.

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -99,6 +99,7 @@ data class PeerDiscoveryUiModel(
     val selectedDevices: List<String> = emptyList(),
     val minimumDevices: Int = MIN_KEYGEN_DEVICES,
     val minimumDevicesDisplayed: Int = MIN_KEYGEN_DEVICES,
+    val allowsMoreDevices: Boolean = false,
```
aminsato, line 102: This VM never sets `allowsMoreDevices` to `true`, so it always relies on the default `false`. But `selectDevice` at line 283 bypasses the `maxOtherDevices` cap for `TssAction.ReShare` (the `!isReshare &&` guard), meaning a reshare flow can legitimately accept devices beyond `minimumDevices - 1`. Result: during reshare, the UI keeps rendering `N/M` without the `+` even though extra devices *can* join — the inverse of the concern raised earlier in this review. Consider setting `allowsMoreDevices = (args?.action == TssAction.ReShare)` when initializing state here so the keygen-side VM stays consistent with its own cap logic.

rkokhatskyi (approved):
LGTM

PR #4599, opened 2026-05-20
fix(swap): flip button resolves same-asset when src was auto-selected

### LINKED ISSUE #4581: Swap: switch button makes source and destination the same asset after changing destination
## Description

On the Swap screen, after changing the destination asset and then tapping the switch (swap direction) button, the source and destination both end up set to the same asset.

## Steps to reproduce

1. Open the app and navigate to **Swap**
2. Change the **destination** asset to one different from the source
3. Tap the **switch** button (the arrows icon between From / To)

## Expected

Source and destination assets get swapped — the previous destination becomes the source, and the previous source becomes the destination.

## Actual

Both **From** and **To** become the same asset (see screenshot — both are set to Bitcoin-Cash / BCH Native).

## Evidence

| Screenshot | Screen recording |
|------------|------------------|
| swap-same-asset | [video](https://i.imgur.com/y8NurEA.mp4) |

johnnyluo:
## Summary
- When navigating to swap without a pre-selected source token, `selectedSrcId` was `null`
- The old flip logic swapped the raw `selectedSrcId`/`selectedDstId` StateFlow values, leaving `selectedSrcId` still `null` after the flip
- The collection loop then resolved both src and dst to the previously selected destination, showing the same asset in both slots
- Fix: read the already-resolved token IDs from `selectedSrc`/`selectedDst` StateFlows before swapping, so concrete IDs are always written

## Issue
Closes #4581

## Test Plan
- [ ] New unit test `flipSelectedTokens swaps correctly when src was not explicitly set` covers the regression scenario
- [ ] Existing `flipSelectedTokens` tests still pass
- [ ] Lint passes (`./gradlew lint`) ✅
- [ ] Manual: open swap screen from a chain page (no pre-selected source), change destination, tap flip — src and dst should swap, not show the same asset

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token-swap flip behavior when the source token is unset so the flip reliably resolves both sides to the correct assets, using safe fallbacks to ensure the selected source and destination tokens are correct after flipping.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4599?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -665,8 +665,12 @@ constructor(
 
         resetQuoteState()
 
-        selectedSrcId.value = currentDstTokenId
-        selectedDstId.value = currentSrcTokenId
+        // Fall back to the raw ID when the resolved SendSrc hasn't loaded yet, so a race between
+        // the flip gesture and token resolution never silently clobbers both slots with null.
+        val newSrcId = currentDstTokenId ?: selectedDstId.value
+        val newDstId = currentSrcTokenId ?: selectedSrcId.value
+        selectedSrcId.value = newSrcId
+        selectedDstId.value = newDstId
 
         // collectSelectedTokens() observes the IDs above and resolves selectedSrc/selectedDst
         // synchronously under Main.immediate. A manual swap of those resolved StateFlows here
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -665,9 +665,8 @@ constructor(
 
         resetQuoteState()
 
-        val previousSrcId = selectedSrcId.value
-        selectedSrcId.value = selectedDstId.value
-        selectedDstId.value = previousSrcId
+        selectedSrcId.value = currentDstTokenId
+        selectedDstId.value = currentSrcTokenId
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -665,9 +665,8 @@ constructor(
 
         resetQuoteState()
 
-        val previousSrcId = selectedSrcId.value
-        selectedSrcId.value = selectedDstId.value
-        selectedDstId.value = previousSrcId
+        selectedSrcId.value = currentDstTokenId
+        selectedDstId.value = currentSrcTokenId
```
aminsato: Switching from raw-id swap to resolved-token swap fixes the reported case, but it introduces the inverse failure mode: if `selectedSrc.value` or `selectedDst.value` is `null` while the raw `selectedSrcId`/`selectedDstId` still hold valid IDs (e.g., address load hasn't completed yet, or one side resolved to `null` because the chain account isn't ready), `currentSrcTokenId`/`currentDstTokenId` will be `null` and this assignment silently clobbers both StateFlows with `null`, wiping the user's selection.

The previous code at least preserved whatever was in the raw IDs across the swap. Two ways to harden this:

1. Bail out early if either resolved token id is `null` (treat flip as a no-op while resolution is pending), or
2. Fall back to the raw id when the resolved id is `null`, e.g. `selectedSrcId.value = currentDstTokenId ?: selectedDstId.value`.

Worth covering with a test mirroring the new one but with `dstTokenId = null` to assert flip is symmetric, since the current test only exercises the `srcTokenId = null` direction.

rkokhatskyi (approved):
LGTM

PR #4600, opened 2026-05-20
Transaction complete: hide empty To row

### LINKED ISSUE #4527: Transaction Complete: empty 'To' row displayed when destination address is blank
## Description

On the **Transaction complete** screen, the **To** row is always rendered, even when the destination address is blank (empty string). This produces a label with no value beneath it, which looks like a broken/missing field.

This is visible in scenarios where there is no destination address — for example, a refunded MayaChain deposit, where the transaction has a `From` and a memo but no `To`.

## Steps to reproduce

1. Trigger a transaction that completes with an empty destination address (e.g. a refunded deposit/swap on MayaChain).
2. Open the Transaction complete screen.
3. Observe the **To** row.

## Expected behavior

If the destination address is empty/blank, the **To** row (label + value + divider) should be hidden entirely, matching the pattern already used for deposit transactions elsewhere in the screen.

## Actual behavior

The **To** label is shown with no value, leaving a visually empty row above the Memo field.

## Notes

- The deposit transaction section in the same screen already conditionally hides `dstAddress` when blank — the same guard should be applied to the standard transaction `To` row.
- Affected composable: `app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt` (the `AddressField` call for the To row).

johnnyluo:
## Summary
- Hide the Transaction Complete `To` address row when the destination address is blank to avoid rendering an empty field.

## Issue
Closes #4527

## Test Plan
- [x] Unit tests pass (`./gradlew :app:testDebugUnitTest`)
- [ ] Manual: complete a tx with blank destination address; verify `To` row is hidden

## Summary by CodeRabbit

* **Bug Fixes**
  * Transaction details view no longer shows an empty destination section. Empty destination address and its divider are hidden, resulting in a cleaner, more concise transaction summary and improved readability when destination information is not provided.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4600?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt
@@ -247,9 +247,8 @@ private fun SwapTransactionDetail(swapTransaction: SwapTransactionUiModel) {
 private fun TransactionDetail(transaction: TransactionDetailsUiModel?) {
     if (transaction != null) {
 
-        UiHorizontalDivider()
-
         if (transaction.dstAddress.isNotBlank()) {
+            UiHorizontalDivider()
             AddressField(
                 title = stringResource(R.string.verify_transaction_to_title),
                 address = transaction.dstAddress,
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt
@@ -249,10 +249,12 @@ private fun TransactionDetail(transaction: TransactionDetailsUiModel?) {
 
         UiHorizontalDivider()
 
-        AddressField(
-            title = stringResource(R.string.verify_transaction_to_title),
-            address = transaction.dstAddress,
-        )
+        if (transaction.dstAddress.isNotBlank()) {
+            AddressField(
+                title = stringResource(R.string.verify_transaction_to_title),
+                address = transaction.dstAddress,
+            )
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt
```diff
@@ -249,10 +249,12 @@ private fun TransactionDetail(transaction: TransactionDetailsUiModel?) {
...
-        AddressField(
-            title = stringResource(R.string.verify_transaction_to_title),
-            address = transaction.dstAddress,
-        )
+        if (transaction.dstAddress.isNotBlank()) {
+            AddressField(
+                title = stringResource(R.string.verify_transaction_to_title),
+                address = transaction.dstAddress,
+            )
+        }
```
aminsato: Consider moving the leading `UiHorizontalDivider()` inside the `if` block so the divider is hidden together with the `To` row. As-is, when `dstAddress` is blank and the following optional fields (`memo`, `functionSignature`/`functionInputs`, or the suppressed amount/value rows) are all absent, the divider can become an orphan separator above `FormDetails`.

```suggestion
        if (transaction.dstAddress.isNotBlank()) {
            UiHorizontalDivider()
            AddressField(
                title = stringResource(R.string.verify_transaction_to_title),
                address = transaction.dstAddress,
            )
        }
```

rkokhatskyi (approved):
LGTM

PR #4601, opened 2026-05-20
fix(send): preserve cached/empty account emissions during DeFi loads (#4458)

### LINKED ISSUE #4458: Send form: cached/empty account emissions discarded during DeFi loads
Follow-up from review comments on #4432 that weren't applied in 74ac9d498 because they require a larger refactor than that PR's scope.

## 1. AccountsLoader: hydrated-only load blocks UI on slow networks

`loadCircleUSDCAccount` and `loadRewardsAccount` call `accountsRepository.loadAddresses(vaultId, isRefresh = true).firstOrNull()`. `isRefresh = true` skips the cached pre-emission that `loadAddresses` normally sends first, so `firstOrNull()` blocks until every parallel balance lookup finishes (or times out) before any data reaches the form.

Effect: `WITHDRAW_USDC_CIRCLE` and `WITHDRAW_RUJI` render an empty form on slow networks where the prior code would have surfaced the cached snapshot immediately.

Suggested approach: collect both emissions — show the cached snapshot first, then replace with the hydrated one. Keeps UI responsive without sacrificing freshness.

## 2. TokenPreselectionService: empty-list guard discards intentional empty publishes

`TokenPreselectionService.preSelect` does `if (accounts.isEmpty()) return@collect` to skip the StateFlow's uninitialized value. That guard also discards the intentional empty publishes from `AccountsLoader.loadRewardsAccount` and `loadCircleUSDCAccount` when prerequisites are missing (no RUNE, no RUJI, no cached rewards, no ETH).

Effect: when a vault is missing those tokens, the form renders with `state.selectedCoin == null` and no preselection fires. The previous code would at least have selected the static template coin from `findDeFiPreselectedToken`.

Suggested approach: a sealed `AccountsLoadState` (e.g., `Uninitialized` / `Loaded(List<Account>)`) so the consumer can disambiguate \"haven't loaded yet\" from \"loaded and intentionally empty\".

## Why follow-up

Both fixes change the shape of data flowing between `AccountsRepository`/`AccountsLoader` and downstream consumers, and ripple through several call sites. Keeping them separate from the send-form phase 4 PR.

aminsato:
## Summary
- Introduce sealed `AccountsLoadState` (`Uninitialized | Loaded`) so consumers can distinguish "not loaded yet" from "loaded and intentionally empty".
- `AccountsLoader.loadCircleUSDCAccount` / `loadRewardsAccount` now collect both cached and hydrated emissions from `loadAddresses` instead of blocking on `firstOrNull` with `isRefresh = true`. The cached snapshot reaches the UI immediately on slow networks; hydrated balances replace it when they arrive.
- `TokenPreselectionService` now skips only `Uninitialized` — `Loaded(emptyList)` (e.g. DeFi prerequisites missing) falls through to the default-coin map so the form is not stuck with `selectedCoin == null`.

Closes #4458 (follow-up to #4432 review comments).

## Test plan
- [x] `./gradlew :app:compileDebugKotlin` clean
- [x] `./gradlew :app:lintDebug` clean
- [x] `./gradlew :app:testDebugUnitTest --tests AccountsLoaderTest --tests TokenPreselectionServiceTest` green (incl. new cases for cached→hydrated republish and `Loaded(empty)` preselection fallback)
- [ ] Manual: WITHDRAW_RUJI / WITHDRAW_USDC_CIRCLE forms populate on a throttled-network device run
- [ ] Manual: WITHDRAW_RUJI on a vault without RUNE/RUJI shows the RUJI rewards default coin instead of a blank form

## Summary by CodeRabbit

* **Refactor**
  * Send UI now receives continuous cached-then-hydrated account updates, handles missing/empty accounts more explicitly, and token auto-selection reliably fires on the first successful load so displayed account/token options update promptly.

* **Tests**
  * Test suite updated to assert load-state emissions, cached vs hydrated snapshots, force-preselection timing, and robustness when required accounts are absent.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4601?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


aminsato:
@coderabbitai full reveiw

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -926,52 +930,57 @@ constructor(
...
+                var tokenSelected = false
+                val addressesFlow =
+                    if (checked) {
+                        accountsRepository.loadAddresses(vaultId)
+                    } else {
+                        accountsRepository.loadDeFiAddresses(vaultId, false)
+                    }
+                addressesFlow
+                    .map { addrs -> addrs.flatMap { it.accounts } }
+                    .collect { loadedAccounts ->
```
rkokhatskyi: `accountsRepository.loadAddresses(vaultId)` is a `channelFlow` ending in `awaitClose()` at `AccountsRepository.kt:122` so on the `checked=true` branch this `collect` never returns, which means the `finally` at line 984 never fires and `isSwitchingAccounts` stays `true` for the rest of the screen so `collectSelectedAccount` at line 1127 stays gated off permanently, whereas the prior `.first()` plus `delay(300)` path returned after one emission and reset the flag. Terminate the flow once both emissions have arrived via `take(2)` or `transformWhile { tokenSelected.not() }`, or reset `isSwitchingAccounts` inside the `collect` block as soon as `tokenSelected` flips so the UI is not frozen after the user enables autocompound

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -926,52 +930,57 @@ constructor(
...
+                        accountsState.value = AccountsLoadState.Loaded(loadedAccounts)
```
rkokhatskyi: `accountsState` already has a long living writer from `AccountsLoader.load()` since `loadAddresses` never closes, so this block adds a second writer racing against it, which is especially bad when `defiType == UNSTAKE_TCY` since the loader is collecting `loadDeFiAddresses` whereas this branch collects `loadAddresses` so their interleaved emissions overwrite each other with data sourced from different APIs. Route the autocompound switch through `AccountsLoader` so a single component owns `accountsState`, or cancel the loader job before starting the local collect here

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -30,6 +29,9 @@ internal class AccountsLoader(
 
     fun load(vaultId: VaultId) {
         loadAccountsJob?.cancel()
+        // Reset before launching the new load so a vault/action switch doesn't leave the
+        // previous session's Loaded(...) visible while the new collector spins up.
+        accountsState.value = AccountsLoadState.Uninitialized
```
rkokhatskyi, line 42: `loadAccountsJob?.cancel()` is cooperative so the previous collector can still execute one more `publishLoaded(it)` after this line resets to `Uninitialized`, which leaves stale `Loaded(...)` data from a superseded vault or action visible until the new collector emits. Thread a `generation` token through `publishLoaded`, `publishCircleUsdc`, `publishRewards` and drop any emission whose generation does not match the current one

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -119,27 +131,31 @@ internal class AccountsLoader(
...
-        // isRefresh = true skips the cached pre-emission — RUNE and RUJI accounts pushed
-        // into accounts.value below are then shown with hydrated balances.
-        val accountsLoaded =
-            accountsRepository.loadAddresses(vaultId, isRefresh = true).firstOrNull()?.flatMap {
-                it.accounts
-            }
+        accountsRepository
+            .loadAddresses(vaultId, isRefresh = false)
+            .map { addrs -> addrs.flatMap { it.accounts } }
+            .collect { accountsLoaded -> publishRewards(vaultId, accountsLoaded) }
```
rkokhatskyi: `stakingDetailsRepository.getStakingDetailsByCoindId` inside `publishRewards` runs once per upstream emission, so the cached snapshot and the hydrated snapshot each trigger the same lookup with the same arguments, whereas the value only needs to be read once per load since neither balance hydration nor MSCA changes affect the staking details row. Resolve the staking details once before the `collect` here, or memoize for the lifetime of the load so the repo is not hit twice per refresh, and the same applies to `publishCircleUsdc` at line 88

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -119,27 +131,31 @@ internal class AccountsLoader(
...
+    }
+
+    private suspend fun publishRewards(vaultId: VaultId, accountsLoaded: List<Account>) {
         val thorchainAccount =
-            accountsLoaded?.find { it.token.id.equals(Coins.ThorChain.RUNE.id, true) }
+            accountsLoaded.find { it.token.id.equals(Coins.ThorChain.RUNE.id, true) }
                 ?: run {
-                    accounts.value = emptyList()
+                    publishLoaded(emptyList())
```
rkokhatskyi: When the vault has no `RUNE` or `RUJI` this publishes `Loaded(emptyList())` which `TokenPreselectionService.findDeFiPreselectedToken` then resolves to the static `RUJI_REWARDS_COIN` template, so `collectSelectedAccount` synthesizes a placeholder `Account` with `tokenValue = null` and the form looks usable while submission cannot succeed since there is no real backing account in the vault. Hide the option when prerequisites are missing, or surface an explicit empty state instead of letting the static template masquerade as a real position

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -256,7 +255,12 @@ constructor(
     private val selectedTokenValue: Coin?
         get() = selectedToken.value
 
-    private val accounts = MutableStateFlow(emptyList<Account>())
+    private val accountsState = MutableStateFlow<AccountsLoadState>(AccountsLoadState.Uninitialized)
+
+    private val accounts: StateFlow<List<Account>> =
+        accountsState
+            .map { it.accountsOrEmpty }
+            .stateIn(viewModelScope, SharingStarted.Eagerly, emptyList())
```
rkokhatskyi, line 271: `accountsState.map { it.accountsOrEmpty }.stateIn(... emptyList())` means every `AccountsLoader.load()` reset to `Uninitialized` flashes `accounts` to `emptyList()`, which `selectedAccount`, `accountValidator`, `amountManager.accountProvider`, `gasFeeOrchestrator`, and `collectSelectedAccount` all observe, so each vault or action switch transiently clears the gas fee, fails validation, and renders a placeholder account row whereas the previous code preserved the old list until the new one arrived. Either keep the prior `Loaded` value during reload, or expose `accountsState` directly to downstream consumers so they can distinguish `Uninitialized` from `Loaded(emptyList())` without seeing a spurious empty list

app/src/test/java/com/vultisig/wallet/ui/models/send/AccountsLoaderTest.kt
```diff
@@ -195,30 +209,91 @@ internal class AccountsLoaderTest {
...
+            val cachedRuji = thorAccount(Coins.ThorChain.RUJI.copy(address = "thor1"))
+            // Hydrated emission uses a different balance so we can tell the two apart.
+            val hydratedRune =
+                cachedRune.copy(tokenValue = TokenValue(BigInteger("999999"), cachedRune.token))
+            val hydratedRuji =
+                cachedRuji.copy(tokenValue = TokenValue(BigInteger("888888"), cachedRuji.token))
+            // yield() between emissions gives the accountsState subscriber a chance to
+            // observe the cached snapshot — without it StateFlow conflates the two writes
+            // and the test only sees the final hydrated value.
+            every { accountsRepository.loadAddresses(VAULT_ID, isRefresh = false) } returns
```
rkokhatskyi, line 252: `flowOf(...)` completes immediately after the last emission whereas production `loadAddresses` is a `channelFlow` ending in `awaitClose()` at `AccountsRepository.kt:122`, so none of these tests exercise the never closing semantic that the new `collect` in both `onAutoCompound` and `AccountsLoader` is actually exposed to, which is why the stuck `isSwitchingAccounts` regression on the `checked=true` autocompound path would pass green here. Cover at least one case with `channelFlow { send(cached); send(hydrated); awaitClose() }` so cancellation, generation, and downstream completion expectations are tested

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -926,52 +930,57 @@ constructor(
     }
 
     fun onAutoCompound(checked: Boolean) {
-        viewModelScope.launch {
+        // safeLaunch — the loadAddresses/loadDeFiAddresses calls below are network-touching
+        // and should funnel through the project's standard error handler instead of
+        // crashing the ViewModel on a transient repository failure.
+        viewModelScope.safeLaunch {
```
rkokhatskyi: `onAutoCompound` was rewritten from `.first()` plus `delay(300)` plus a single `selectToken` into a long living `collect` with a `tokenSelected` latch and a `try/finally`, with no accompanying unit test, so the next refactor here has no safety net for the very regression this PR is trying to fix. Add a focused test that drives cached then hydrated emissions and asserts the `accountsState` updates, the single `selectToken` call, and that `isSwitchingAccounts` eventually returns to `false`

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -926,52 +930,57 @@ constructor(
...
+                // resetUserInputCache).
+                val targetTicker = if (checked) "sTCY" else "TCY"
+                var tokenSelected = false
+                val addressesFlow =
+                    if (checked) {
+                        accountsRepository.loadAddresses(vaultId)
+                    } else {
+                        accountsRepository.loadDeFiAddresses(vaultId, false)
+                    }
+                addressesFlow
```
rkokhatskyi: iOS `UnstakeTransactionViewModel` and Windows `UnstakeTCYSpecific.tsx` plus `UnstakeSTCY.tsx` never reload the accounts list when autocompound flips, they switch the balance source on the already loaded coin, whereas this Android path reloads the entire vault accounts, nulls `selectedToken`, and races a new collector against the already running `AccountsLoader`. Aligning with the iOS or Windows model would remove the dual writer hazard on `accountsState` and the `delay(300)` heuristic that this refactor had to drop

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormViewModel.kt
```diff
@@ -926,53 +938,40 @@ constructor(
...
+            loadedAccounts ->
+            if (!tokenSelected) {
+                loadedAccounts
+                    .find {
+                        it.token.ticker.equals(targetTicker, true) &&
+                            it.token.chain == Chain.ThorChain
+                    }
+                    ?.let {
+                        tokenSelected = true
+                        isSwitchingAccounts.value = false
```
rkokhatskyi: `isSwitchingAccounts.value = false` lives inside the `?.let { ... }` block so it only fires once `find` returns a non-null match, which means if neither the cached nor the hydrated emission contains the target ticker the flag stays `true` and `collectSelectedAccount` at line 1113 keeps returning `null` for the rest of the screen, leaving the form gated. This is reachable when a vault holds `sTCY` but the user toggles to direct on a `UNSTAKE_STCY` form without ever having held `TCY`, or holds `TCY` but toggles to autocompound on a `UNSTAKE_TCY` form without ever having held `sTCY`, so hoist the flag reset above the `find` or move it into a `finally` around the `loadForAutoCompoundSwitch` invocation so the first `Loaded` emission always clears the gate

app/src/test/java/com/vultisig/wallet/ui/models/send/SendFormViewModelAutoCompoundTest.kt
```diff
@@ -0,0 +1,204 @@
...
+            assertEquals("999", selected.balance)
+            assertTrue(
+                balancesPassedToMapper.contains(BigInteger("100")),
+                "cached balance never reached the mapper (autocompound dropped the cached snapshot)",
+            )
+            assertTrue(
+                balancesPassedToMapper.contains(BigInteger("999")),
+                "hydrated balance never reached the mapper",
+            )
+        }
```
rkokhatskyi, line 164: Only the happy path where the hydrated emission contains `sTCY` is exercised, whereas the regressions worth fencing in are the cases where neither emission contains the target ticker (which leaves `isSwitchingAccounts` stuck per the gating issue on `SendFormViewModel.kt:972`) and `onAutoCompound(false)` from a `UNSTAKE_STCY` form whose vault has no `TCY` row. Add a missing-target case asserting the gate still releases, and a `checked = false` case so the toggle-off path is not silently broken by a future refactor

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -54,62 +68,124 @@ internal class AccountsLoader(
...
+                    if (useStableCompound) {
+                        accountsRepository.loadAddresses(vaultId)
+                    } else {
+                        accountsRepository.loadDeFiAddresses(vaultId, false)
+                    }
+                addressesFlow
+                    .map { addrs -> addrs.flatMap { it.accounts } }
+                    .collect { accounts ->
+                        if (generation != currentGeneration) return@collect
+                        accountsState.value = AccountsLoadState.Loaded(accounts)
```
rkokhatskyi: `loadForAutoCompoundSwitch` writes to `accountsState` directly while `load()` routes every publish through `publishLoaded` which already enforces the generation gate, so this duplicates the generation check at line 110 and forks the publish path for one call site. Call `publishLoaded(accounts, generation)` then invoke the callback so the gating logic stays in one place and a future change to `publishLoaded` does not silently skip this code path

rkokhatskyi (changes requested):

rkokhatskyi (approved):
LGTM

PR #4610, opened 2026-05-20
fix(maya): surface CACAO unstake maturity countdown (#4498)

### LINKED ISSUE #4498: [BUG] MayaChain CACAO Single Pool Staking — Unstake button disabled (greyed out) past 7-day unbond period
# [Fix] MayaChain CACAO Unstake button stays disabled past 7-day unbond period [UnstakeCacaoScreen]

## Problem
The **Unstake** CTA on the MayaChain CACAO Single Pool Staking surface remains disabled (greyed out) even after the 7-day unbond period has elapsed. The button never becomes tappable, leaving the user with no in-wallet path to withdraw their CACAO. Same family / surface pattern as the previously-resolved Maya bond/unbond CTA-disabled bugs (vultisig-windows#601, vultisig-ios#4185), but on a different action.

## Expected Behavior
After the 7-day unbond period has elapsed, the Unstake CTA should be enabled and the flow should proceed to: amount entry (or full-position withdrawal) → fee preview → keysign → broadcast.

## Steps to Reproduce
1. Stake CACAO via MayaChain DeFi → Stake CACAO flow (UI shipped in #4074, 2026-04-19).
2. Wait ≥ 7 days (the Maya pool unbond period).
3. Navigate to MayaChain DeFi → Unstake CACAO.
4. Observe **Unstake** button is disabled / greyed out.
5. No way to initiate the unbond withdrawal from the wallet.

### Must Do
- [ ] Locate `UnstakeCacaoScreen` and its viewmodel; identify the gating predicate (likely `isUnstakeEnabled` or equivalent) controlling the Unstake button's enabled state.
- [ ] Verify the eligibility check reads live Maya pool state (not a cached/initial-mount snapshot) and uses the correct unbond clock source (pool/chain block time, not local device clock or stake-block).
- [ ] Cross-check against Maya RPC: if chain reports the position as unbond-eligible but UI still greys the CTA, fix the UI predicate so the button enables once the 7-day threshold is crossed.

### Must NOT Do
- Don't change unrelated code
- Don't modify the staking/bond flow or shared MayaChain SDK behavior beyond what's needed for the Unstake eligibility gate
- Don't introduce a client-side time check that bypasses on-chain state — eligibility must come from chain state, not local clock

## Acceptance Criteria
- [ ] With a staked CACAO position older than 7 days, the **Unstake** button on `UnstakeCacaoScreen` is enabled and tappable
- [ ] Tapping Unstake proceeds through amount entry → fee preview → keysign → broadcast
- [ ] If chain reports the position as not-yet-eligible, the button remains disabled (gating logic still respects on-chain truth)

### Likely root cause hypotheses
- **Eligibility check reads stale on-chain state.** The unstake-eligible flag may be computed once at screen mount and never refreshed; the user could have crossed the 7-day threshold without the UI re-querying.
- **Wrong unbond-clock source.** The 7-day clock may be computed against a wrong starting timestamp (e.g. user's local clock or stake-block-not-pool-block).
- **Hardcoded disable.** The button may be unconditionally disabled until a deposit/refresh action runs.
- **`isUnstakeEnabled` flag false-pinned in `UnstakeCacaoScreen` viewmodel** — quick first thing to check.

### Platform
Reported on Android (most likely — stake/unstake screens originally shipped on Android via PR #4074). **iOS / Windows status needs verification** — those platforms may not even have the surface yet (no equivalent stake-CACAO PR found on iOS in search).

### Investigation asks
- Reproduce on Android with a staked CACAO position that's > 7 days old.
- Check `UnstakeCacaoScreen` (and its viewmodel) for the gating predicate. Verify it reads live Maya pool state, not a cached/initial-load snapshot.
- Cross-check against Maya RPC for the actual pool position's age — confirm the chain reports the position as unbond-eligible.
- If state is correctly eligible chain-side but UI still grey: fix the UI predicate. If chain reports not-yet-eligible despite 7+ days: it's a Maya / SDK indexer bug, not a UI bug — escalate accordingly.

### Reference
- Originating user request (via Vultisig social): "@vultisig @jpthor CACAO Single Pool Staking 'Unstake' button is disabled (grey out) even though now it's passed more than 7 days..? Similar to the maya unbond issue raised earlier."
- Prior Maya CTA-disabled bugs (closed): vultisig-windows#601, vultisig-ios#4185
- Originating screens: vultisig-android PR #4074 (merged 2026-04-19)

### What I'd want from the reporter
- Vault address / Maya CACAO pool position address (so devs can replay on-chain state)
- Stake tx hash + estimated stake date
- App version / build number
- Platform (assumed Android based on shipped screens, but worth confirming)

aminsato:
## Summary

Closes #4498.

Maya's `CACAOPOOLDEPOSITMATURITYBLOCKS` mimir constant moved from 7d to 21d (302,400 blocks × 6s). The pre-send guard was already reading mimir live, so the math was correct — but between days 7 and 21 (and any time the maturity check fails) the Unstake CTA was simply rendered grey with no explanation, leaving users with no way to know *why* they couldn't unstake.

This PR replaces the boolean validator with a richer use case that exposes the remaining lock time, and surfaces a `Unlocks in N days, H hours` hint everywhere the user might see the disabled button.

## Changes

- `ValidateMayaTransactionHeightUseCase` → `GetCacaoUnstakeMaturityUseCase` returning `Mature` / `Locked(remainingBlocks)` / `Unknown`. Maya block time (6s) is captured as a constant on the sealed class.
- `MayaCacaoStakingService` exposes `unstakeUnlocksInSeconds: Long?` alongside `canUnstake`.
- `StakingTabComponents` renders a countdown label above the disabled Unstake action on the MayaChain DeFi positions list.
- `UnstakeCacaoScreen` (via `DepositFormViewModel`) fetches maturity on screen entry and shows the same hint inline; the Continue CTA is gated on `Mature` rather than just `amount > 0`.
- `Unknown` state shows `"Couldn't verify position. Try again."` instead of a silent grey button — a network blip no longer indistinguishable from a not-yet-mature position.
- Strings added to all 10 locales (en, de, es, hr, it, ko, nl, pt, ru, zh-rCN).
- Five `@Preview` variants on `UnstakeCacaoContent` cover Mature, Locked (days+hours), Locked (hours-only), Unknown, and Loading states.

## Tests

- `GetCacaoUnstakeMaturityUseCaseTest` — 6 tests: Mature past unlock, Mature exactly at unlock, Locked remaining block math, Locked remainingSeconds 6s mapping, Unknown on API failure, `CancellationException` rethrow.
- `CacaoUnstakeMaturityFormatterTest` — 4 tests: days+hours, hours-only, sub-hour `Unlocks soon` fallback, exactly-one-day edge.
- Existing `DepositFormViewModelTest` migrated to the new use case and still passes (12 tests).

## Test plan

- [ ] Open a vault with a CACAO pool position less than 21 days old → Unstake card shows `Unlocks in Nd Hh` above a disabled button; tapping `Unstake` from the DeFi positions list is disabled with the same label.
- [ ] Open a vault with a CACAO pool position older than 21 days → Unstake CTA enabled, flow proceeds to amount entry and Continue.
- [ ] Toggle airplane mode and re-enter the Unstake CACAO screen → CTA disabled with `Couldn't verify position. Try again.`, not a silent grey.
- [ ] Verify each locale renders the new strings (manual spot check on de/es/ko/zh-rCN).


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/maya/MayaCacaoStakingService.kt
```diff
@@ -17,35 +18,42 @@ data class MayaCacaoStakingDetails(
     val stakeAmount: BigInteger,
     val apr: Double?,
     val canUnstake: Boolean,
+    // null when there is no stake position, the position is already mature, or the
+    // maturity fetch failed. UI uses this only to render a "Unlocks in N days, H hours"
+    // hint next to the disabled Unstake action.
+    val unstakeUnlocksInSeconds: Long? = null,
```
rkokhatskyi, line 24: `canUnstake` and `unstakeUnlocksInSeconds` are both derived from the same `maturity` result a few lines below so they can never legally express anything beyond what a single `CacaoUnstakeMaturity` already encodes, yet the two-field shape lets a caller see `canUnstake = true` paired with a non-null `unstakeUnlocksInSeconds` which is unreachable but the type doesn't enforce it, whereas exposing the raw maturity and letting `StakingTabComponents` pattern-match keeps the invariant in the type rather than in two parallel fields the service has to manually keep aligned

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
```diff
@@ -1093,6 +1097,13 @@ constructor(
 
     private suspend fun handleRemoveCacaoOption() {
         val addressValue = address.value?.address ?: return
+        // Reset maturity to null so the UI shows loading instead of stale state when the
+        // user re-enters the screen.
+        _state.update { it.copy(unstakeMaturity = null) }
+        viewModelScope.safeLaunch {
+            val maturity = getCacaoUnstakeMaturity(addressValue)
```
rkokhatskyi, line 1104: Launching the maturity fetch into `viewModelScope.safeLaunch` with no job handle leaves rapid `selectDepositOption(RemoveCacaoPool)` reentries stacking overlapping fetches that all race to write `unstakeMaturity`, whereas this file already uses the `loadLpJob?.cancel()` and `switchInboundJob?.cancel()` pattern for cancellable side effects so a `private var maturityJob: Job? = null` cancelled on re-entry would fit the existing convention

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
```diff
@@ -2032,10 +2043,11 @@ constructor(
 
         val srcAddress = selectedToken.address
 
-        validateMayaTransactionHeight(srcAddress) ||
+        if (getCacaoUnstakeMaturity(srcAddress) !is CacaoUnstakeMaturity.Mature) {
```
rkokhatskyi, line 2046: Re-querying `getCacaoUnstakeMaturity` here repeats the three-call network round-trip that `handleRemoveCacaoOption` already cached into `state.unstakeMaturity` and that the CTA at `UnstakeCacaoScreen.kt:234` already gates on, so every Continue tap pays the round-trip again while any transient blip downgrades the result to `Unknown` and throws `deposit_error_has_not_reached_maturity` from an enabled button, either pass the cached `state.value.unstakeMaturity` through or short-circuit the recheck when the cached state is already `Mature`

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/maya/UnstakeCacaoScreen.kt
```diff
@@ -201,31 +203,117 @@ private fun UnstakeCacaoContent(
             )
         }
 
+        val maturityHint: UiText? =
+            when (val maturity = state.unstakeMaturity) {
+                is CacaoUnstakeMaturity.Locked -> cacaoUnlocksInUiText(maturity.remainingSeconds)
+
+                CacaoUnstakeMaturity.Unknown ->
+                    UiText.StringResource(R.string.unstake_cacao_maturity_check_failed)
```
rkokhatskyi, line 211: The `Unknown` branch renders a `Try again` imperative as static text but the screen exposes no retry control anywhere so the user reads a call to action they can't act on, whereas either wiring a retry button that re-invokes `handleRemoveCacaoOption` or rephrasing the copy to drop the actionable verb would close the loop

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/maya/UnstakeCacaoScreen.kt
```diff
@@ -201,31 +203,117 @@ private fun UnstakeCacaoContent(
...
+private fun UnstakeCacaoContentLockedDaysHoursPreview() {
+    Box(modifier = Modifier.background(Theme.v2.colors.backgrounds.primary)) {
+        UnstakeCacaoContent(
+            state =
+                DepositFormUiModel(
+                    unstakableAmount = "5,000 CACAO",
+                    balance = UiText.DynamicString("24,000 CACAO"),
+                    unstakeMaturity =
+                        CacaoUnstakeMaturity.Locked(
+                            remainingBlocks = (5L * 24L * 600L) + (3L * 600L)
```
rkokhatskyi, line 267: The preview hardcodes `600L` as blocks per hour to back into the `remainingSeconds = remainingBlocks * MAYA_BLOCK_TIME_SECONDS` formula on `Locked`, so any future change to the block time silently drifts the previews from the live computation while production stays correct, whereas building the preview from seconds and converting through `MAYA_BLOCK_TIME_SECONDS` or adding a `Locked.fromSeconds` builder would keep the inverse in one place

app/src/test/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModelTest.kt
```diff
@@ -79,8 +79,7 @@ internal class DepositFormViewModelTest {
     private val mayachainBondRepository: MayachainBondRepository = mockk(relaxed = true)
     private val balanceRepository: BalanceRepository = mockk(relaxed = true)
     private val gasFeeToEstimatedFee: GasFeeToEstimatedFeeUseCase = mockk(relaxed = true)
-    private val validateMayaTransactionHeight: ValidateMayaTransactionHeightUseCase =
-        mockk(relaxed = true)
+    private val getCacaoUnstakeMaturity: GetCacaoUnstakeMaturityUseCase = mockk(relaxed = true)
```
rkokhatskyi, line 82: The `relaxed = true` mock leaves the new tri-state flow entirely uncovered since there is no case for `selectDepositOption(RemoveCacaoPool)`, no case for the `Locked` rejection inside `createRemoveCacaoPoolTransaction`, and no case for the `Unknown` rejection, so the gating change from a boolean to a sealed result the rest of the flow now depends on goes untested at the ViewModel layer

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/StakingTabComponents.kt
```diff
@@ -216,6 +217,19 @@ internal fun StakingWidget(
             UiSpacer(16.dp)
         }
 
+        if (
+            state.stakeAmount > BigDecimal.ZERO &&
+                !state.canUnstake &&
+                state.unstakeUnlocksInSeconds != null
```
rkokhatskyi, line 223: When the maturity fetch returns `Unknown` the service emits `unstakeUnlocksInSeconds = null` alongside `canUnstake = false` so this condition still hides any hint and the user is back to the exact silent grey button this PR set out to fix, whereas surfacing the `Unknown` case here too either by carrying the maturity through `StakePositionUiModel` or by emitting a `couldn't verify` fallback when `canUnstake = false` with no countdown would close that regression

rkokhatskyi (changes requested):

aminsato:
Superseded by #4611 — the conceptual improvements from this PR (Unknown / network-error state, DeFi positions list countdown, MayaCacaoStakingService maturity field, shared cacaoUnlocksInUiText formatter) have been moved over there. Closing to consolidate review on a single PR.

PR #4613, opened 2026-05-21
fix(swap): clear stale network fee when switching From asset to UTXO chain

### LINKED ISSUE #4582: Swap: percentage buttons trigger 'Insufficient balance' error after switching From asset to Zcash
## Description

Starting from a Zksync swap form, after changing the **From** asset to **Zcash (ZEC)** and tapping one of the percentage buttons (25% / 50% / 75%), the form shows:

> Insufficient ZEC balance for the swap and gas fees. Top up ZEC or reduce swap amount

…even though the ZEC balance is clearly sufficient (0.29950513 ZEC in the screenshot).

The underlying exception in logcat is `SwapException$AmountCannotBeZero: Amount must be positive`, which suggests the percentage tap is being applied to a stale/zero source amount, but the resulting error message is misleading (shown as an insufficient-balance error rather than an internal issue).

## Steps to reproduce

1. Open the app and navigate to **Swap** with a pair where **To** is Zksync (ETH)
2. Change the **From** asset to **Zcash (ZEC)** — balance is non-zero (e.g. 0.29950513 ZEC)
3. Tap any percentage button (25% / 50% / 75%)

## Expected

Tapping a percentage button populates the From amount with that percentage of the available ZEC balance, and a quote is fetched.

## Actual

An error toast/popup appears: "Insufficient ZEC balance for the swap and gas fees. Top up ZEC or reduce swap amount", despite the balance being sufficient.

## Logcat

```
com.vultisig.wallet.data.api.errors.SwapException$AmountCannotBeZero: Amount must be positive
    at com.vultisig.wallet.ui.models.swap.SwapFormViewModel$calculateFees$1$5.emit(SwapFormViewModel.kt:871)
    at com.vultisig.wallet.ui.models.swap.SwapFormViewModel$calculateFees$1$5.emit(SwapFormViewModel.kt:859)
    at kotlinx.coroutines.flow.FlowKt__DelayKt$debounceInternal$1$3$1.invokeSuspend(Delay.kt:226)
    ...
```

## Evidence

| Screenshot | Screen recording |
|------------|------------------|
| swap-zec-error | [video](https://i.imgur.com/Uea02up.mp4) |

johnnyluo:
## Summary
- When switching the swap source from a non-UTXO chain (e.g. Zksync ETH) to a UTXO chain (e.g. ZEC), the old `estimatedNetworkFeeTokenValue` was never cleared
- Tapping a percentage button then subtracted that stale cross-chain fee (ETH in wei, ~10^13) from the new chain's balance (ZEC in satoshis, ~10^7), making `maxUsableTokenAmount` negative and triggering a false "Insufficient balance" error
- Fix: clear `estimatedNetworkFeeTokenValue` and `estimatedNetworkFeeFiatValue` in `calculateGas()` whenever the source chain transitions to a UTXO chain, before the UTXO plan fee is recomputed

## Issue
Closes #4582

## Test Plan
- [ ] Start a swap with Zksync ETH as the From asset
- [ ] Switch From to Zcash (ZEC) with a non-zero balance
- [ ] Tap a percentage button (25% / 50% / 75%) — it should populate the amount field correctly, not show an insufficient balance error
- [ ] Verify the swap quote is fetched after the amount is set

## Summary by CodeRabbit

* **Bug Fixes**
  * When switching chains during a swap, estimated network fee and fiat values are cleared and refreshed so users no longer see stale or incorrect fee/fiat estimates.
  * Source-percentage selection no longer subtracts an unrelated cross-chain gas fee; fees are applied only when the source token is native and the fee chain matches the token's chain.

* **Tests**
  * Added a regression test ensuring source-percentage selection doesn't apply stale cross-chain fees.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4613?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -726,8 +726,9 @@ constructor(
         val maxUsableTokenAmount =
             srcTokenValue.value -
                 swapFee -
-                (estimatedNetworkFeeTokenValue.value?.value?.takeIf { srcToken.isNativeToken }
-                    ?: BigInteger.ZERO)
+                (estimatedNetworkFeeTokenValue.value?.value?.takeIf {
+                    srcToken.isNativeToken && gasFeeChain.value == srcToken.chain
+                } ?: BigInteger.ZERO)
 
         if (maxUsableTokenAmount <= BigInteger.ZERO) {
             srcAmountState.setTextAndPlaceCursorAtEnd("0")
@@ -831,6 +832,7 @@ constructor(
                         // can compute the correct UTXO plan fee.
                         estimatedNetworkFeeTokenValue.value = null
                         estimatedNetworkFeeFiatValue.value = null
+                        uiState.update { it.copy(networkFee = "", networkFeeFiat = "") }
                         // The plan-fee block in calculateFees() may have already run
                         // with a stale or null gasFeeChain and skipped via its chain guard,
                         // leaving the form fee blank; re-fire so it can compute with the byte
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -825,7 +825,13 @@ constructor(
                         }
                     } else if (previousChain != chain) {
                         // UTXO non-Cardano + chain transitioned (initial selection or token
-                        // switch). The plan-fee block in calculateFees() may have already run
+                        // switch). Clear any stale fee from the previous chain immediately so
+                        // selectSrcPercentage() doesn't subtract a cross-chain fee value
+                        // (e.g. ETH wei subtracted from ZEC satoshis) before calculateFees()
+                        // can compute the correct UTXO plan fee.
+                        estimatedNetworkFeeTokenValue.value = null
+                        estimatedNetworkFeeFiatValue.value = null
+                        // The plan-fee block in calculateFees() may have already run
                         // with a stale or null gasFeeChain and skipped via its chain guard,
                         // leaving the form fee blank; re-fire so it can compute with the byte
                         // fee for this chain.
--- a/app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
@@ -513,6 +513,32 @@ internal class SwapFormViewModelTest {
             assertEquals("0", vm.srcAmountState.text.toString())
         }
 
+    @Test
+    fun `selectSrcPercentage does not subtract stale cross-chain fee`() =
+        runTest(mainDispatcher) {
+            // The default setUp mock returns chain=Ethereum for any calculateGasFee call.
+            // With BTC as src, gasFeeChain ends up as Ethereum while srcToken.chain is Bitcoin
+            // — identical to the real bug (switching from ETH→ZEC leaves a huge wei fee value
+            // that, if subtracted from satoshis, makes maxUsableTokenAmount negative).
+            val vm =
+                createViewModelWithAddresses(
+                    addresses = listOf(btcAddressLargeBalance(), ethAddress()),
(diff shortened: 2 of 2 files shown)


aminsato:
@coderabbitai full reveiw

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -825,7 +825,13 @@ constructor(
                         }
                     } else if (previousChain != chain) {
                         // UTXO non-Cardano + chain transitioned (initial selection or token
-                        // switch). The plan-fee block in calculateFees() may have already run
+                        // switch). Clear any stale fee from the previous chain immediately so
+                        // selectSrcPercentage() doesn't subtract a cross-chain fee value
+                        // (e.g. ETH wei subtracted from ZEC satoshis) before calculateFees()
+                        // can compute the correct UTXO plan fee.
+                        estimatedNetworkFeeTokenValue.value = null
```
aminsato, line 833: Should the same stale-fee clear also happen when transitioning **away** from a UTXO chain (UTXO → non-UTXO)? 

In that direction the non-UTXO branch above unconditionally overwrites `estimatedNetworkFeeTokenValue` on the next gas result, so it self-heals — but there is still a small window between chain selection and the new gas result during which `selectSrcPercentage()` would subtract the previous chains fee (e.g. ZEC satoshis subtracted from an ETH wei balance). The magnitude difference makes it practically harmless, unlike the bug this PR fixes (UTXO → much smaller decimals), so this might not be worth changing — but worth considering if we want symmetric guarantees that the percentage button never uses a cross-chain fee.

johnnyluo: Good question — considered the symmetric direction. For UTXO → non-UTXO (e.g. ZEC satoshis → ETH wei), the magnitude difference makes the window negligible: subtracting ~10,000 satoshis from a 10^18 wei balance barely moves the needle (~0.000001% error), so the user still gets essentially 100% of their ETH. The non-UTXO branch also unconditionally overwrites `estimatedNetworkFeeTokenValue` in the same `collect` block, so it self-heals on the very next gas result. To truly fix the window symmetrically we'd need a separate `selectedSrc` observer that clears on any chain transition — meaningful complexity for an imperceptible impact. Leaving it as-is.

rkokhatskyi, line 833: The root cause sits in `selectSrcPercentage` at line 729 which subtracts `estimatedNetworkFeeTokenValue.value` without verifying it matches `srcToken.chain`, so gating that read on `gasFeeChain.value == srcToken.chain` inside the `takeIf` predicate would close the window without any manual state clearing here, which also covers the reverse transition direction at the read site rather than relying on magnitude differentials

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -825,7 +825,13 @@ constructor(
...
+                        estimatedNetworkFeeFiatValue.value = null
```
rkokhatskyi, line 834: Every other place in this file that nulls `estimatedNetworkFeeTokenValue` and `estimatedNetworkFeeFiatValue` also writes empty strings into `uiState.networkFee` and `uiState.networkFeeFiat` in the same update (see lines 1062, 1086, 1101), whereas this new branch skips that step, so the form keeps displaying the previous chain's fee text until `calculateFees` runs again which only happens once `srcAmount` is not empty

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -825,7 +825,13 @@ constructor(
...
+                        // The plan-fee block in calculateFees() may have already run
```
rkokhatskyi, line 836: No regression test was added for this fix while the existing `selectSrcPercentage` tests in `SwapFormViewModelTest` only cover a single chain in steady state, so a future refactor could reintroduce the stale fee subtraction across chains without CI catching it

rkokhatskyi (changes requested):

rkokhatskyi (approved):
LGTM

PR #4616, opened 2026-05-21
fix(tx-history): render live elapsedTime in pending chip

### LINKED ISSUE #4559: Tx history: pending "In progress…" chip is static — elapsedTime not rendered + 1-min ticker
## Summary

On the transaction history screen, a pending transaction's status chip shows a static `In progress…` text and never appears to update. Compared to iOS (which renders `In Progress... <elapsedTime>` and updates continuously), the Android chip looks frozen — the user reads it as a stuck/broken animation.

## Root cause

Two stacked issues:

### 1. Widget drops the `elapsedTime` payload

`TransactionStatusUiModel.Pending` carries an `elapsedTime` field that the VM populates:

`app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt:64`
```kotlin
data class Pending(val elapsedTime: UiText) : TransactionStatusUiModel
```

`TransactionHistoryViewModel.kt:351-360`
```kotlin
TransactionStatus.PENDING ->
    TransactionStatusUiModel.Pending(elapsedTime = formatElapsed(now - timestamp))
...
TransactionStatus.NotFound ->
    TransactionStatusUiModel.Pending(elapsedTime = formatElapsed(now - timestamp))
```

But the render-side widget never reads it:

`app/src/main/java/com/vultisig/wallet/ui/screens/transaction/components/TransactionHistoryComponents.kt:72-83`
```kotlin
if (isInProgress) {
    Text(
        text = stringResource(R.string.transaction_status_in_progress_label),  // "In progress…"
        ...
    )
}
```

The chip should render `In progress… <elapsedTime>` (matching iOS).

### 2. Ticker fires once a minute

`TransactionHistoryViewModel.kt:241-247`
```kotlin
private fun startTimeTicker() {
    viewModelScope.launch {
        while (true) {
            delay(1.minutes)
            currentTime.value = System.currentTimeMillis()
        }
    }
}
```

Even after fix #1, the chip would only refresh every 60 seconds. iOS updates continuously — Android should match (1s tick while any Pending tx is visible, paused otherwise).

## iOS reference

`VultisigApp/VultisigApp/Features/TransactionHistory/Views/TransactionHistoryCardView.swift:77-89`

```swift
private var inProgressChip: some View {
    HStack(spacing: 0) {
        Text("inProgress".localized + "... ")
            .foregroundStyle(Theme.colors.textTertiary)
        Text(elapsedTimeString)
            .foregroundStyle(Theme.colors.textPrimary)
    }
    .font(Theme.fonts.caption12)
    .padding(.horizontal, 12)
    .padding(.vertical, 8)
    .background(Theme.colors.bgPrimary)
    .cornerRadius(99)
}
```

Same pill shape, but second `Text` is the live elapsed counter — that's the visible "animation".

## Suggested Android fix

1. **Render elapsedTime in the chip.** `TransactionHistoryComponents.kt:65-108` — when status is `Pending`, render:
   ```kotlin
   Row {
       Text("In progress… ", style = ..., color = textTertiary)
       Text((status as Pending).elapsedTime.asString(), style = ..., color = textPrimary)
   }
   ```
   Wrap in the same rounded pill container.
2. **Tick faster.** `TransactionHistoryViewModel.kt:241-247` — drop `delay(1.minutes)` to `delay(1.seconds)`. Optionally gate on "any pending tx visible" to avoid burning cycles on a fully-confirmed list.
3. (Optional) **Add a subtle pulse on the chip background** if the design calls for it (check Figma — the user's screenshot reads as "stuck", which could mean both the missing counter AND a missing pulse).

## Repro

1. Send a transaction on any chain.
2. Land back on the History tab — pending tx shows `In progress…` chip.
3. Wait. The chip text and any time indicator never change.
4. Compare to iOS — pill ticks the elapsed time every second.

## Platform

Android only. iOS already correct (used as reference above).

aminsato:
## Summary

Fixes #4559. The pending transaction chip was rendering only `In progress…` without the `elapsedTime` field the VM already produced, and the ticker ran on a 1-minute cadence — so the chip looked frozen compared to iOS, which shows a live elapsed counter.

- **`TransactionStatusWidget`** — `Pending` now renders a Row pill with `"In progress… "` (tertiary) + `elapsedTime` (primary). `Broadcasted` keeps the static pill; `Confirmed`/`Failed`/`Refunded` unchanged. Refactored the if/else (label, color) ternary into a clean `when` over the sealed type.
- **`startTimeTicker`** — Drops `delay(1.minutes)` → `delay(1.seconds)`, and now gates on "any visible Pending tx" via `_uiState.map { … }.distinctUntilChanged().collectLatest`. Loop idles when no pending txs are visible — no wasted ticks on a fully-confirmed list.
- **`formatElapsed`** — Adds the seconds branch so the counter visibly changes every second instead of sitting on "just now" for a minute.
- **`transaction_history_elapsed_seconds`** — Added to all 10 locales using each locale's existing compact pattern (`%1$ds ago`, `vor %1$ds`, `hace %1$ds`, `%1$d秒前`, etc.).
- **`@Preview`s** — Added `PreviewTransactionStatusWidget` covering all 7 status states, and `PreviewSwapCardPending` so the chip is visible in card context too.

iOS reference: `TransactionHistoryCardView.swift:77-89` — same pill shape with a live elapsed counter as the second `Text`.

## Test plan

- [ ] Send a transaction on any chain; confirm the chip on the History tab shows `In progress… Ns ago` and increments every second.
- [ ] Wait past 1 minute / 1 hour boundaries; format transitions correctly to `1m ago` / `1h ago`.
- [ ] Open History when no pending txs exist; verify no per-second work happens (ticker idle).
- [ ] All 7 status states render correctly via `PreviewTransactionStatusWidget` in Android Studio's preview pane.

## Summary by CodeRabbit

* **New Features**
  * Pending and broadcasted transactions show a live "In progress" pill with an elapsed-time counter calculated at render time.
  * Transaction status views now receive timestamps so elapsed times are computed by the UI.

* **Bug Fixes**
  * Transactions previously shown as "not found" now display consistently as pending/in-progress.

* **Localization**
  * Added seconds and minutes+seconds elapsed-time formats for multiple languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4616?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -464,6 +479,8 @@ constructor(
                 UiText.FormattedText(R.string.transaction_history_elapsed_hours, listOf(hours))
             minutes > 0 ->
                 UiText.FormattedText(R.string.transaction_history_elapsed_minutes, listOf(minutes))
+            seconds > 0 ->
+                UiText.FormattedText(R.string.transaction_history_elapsed_seconds, listOf(seconds))
```
rkokhatskyi: The PR description and the inline comment in `TransactionStatusWidget` both claim parity with iOS whereas iOS in `TransactionHistoryCardView.swift` uses a continuous `Xm Ys` format that keeps ticking every second indefinitely, so past 60s the Android chip flips to `1m ago` and visibly freezes there for a full minute before going to `2m ago` which makes the ticker pointless after the first minute since most of its work is invisible to the user. A combined `Xm Ys` formatter for the Pending case would actually deliver on the live counter claim.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/components/TransactionHistoryComponents.kt
```diff
@@ -66,47 +71,70 @@ internal fun TransactionStatusWidget(
...
-                    )
-                    .padding(horizontal = 12.dp, vertical = 8.dp),
-        )
-    } else {
-        val (label, color) =
-            when (status) {
-                TransactionStatusUiModel.Confirmed ->
-                    stringResource(R.string.transaction_status_confirmed_label) to
-                        Theme.v2.colors.alerts.success
+        TransactionStatusUiModel.Broadcasted ->
```
rkokhatskyi, line 79: `BROADCASTED` is the initial state written by `TransactionDao.kt` the moment a tx is broadcast whereas `PENDING` only kicks in once the indexer has seen the tx, so right after sending the user lands on history and stares at a static `In progress…` with no counter exactly during the window where a live counter is most useful since they just dispatched the transaction. iOS folds both into one `.inProgress` chip with the counter for the whole interval so the parity claim is half met, feeding `elapsedTime` into the `Broadcasted` UI model and reusing the same Row would close that gap.

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -248,11 +249,24 @@ constructor(
...
+                    state.groups.any { group ->
+                        group.transactions.any { it.status is TransactionStatusUiModel.Pending }
+                    }
+                }
+                .distinctUntilChanged()
+                .collectLatest { hasPending ->
+                    if (hasPending) {
+                        while (true) {
+                            delay(1.seconds)
+                            currentTime.value = System.currentTimeMillis()
```
rkokhatskyi: This `currentTime.value` write flows through `combine(currentTime)` inside `observeTransactions` so the full entity list is mapped to brand new `Send` and `Swap` data classes and regrouped by date every single second, which means every transaction card recomposes once a second even when only one row is pending whereas the previous one minute cadence hid that cost. Pulling `elapsedTime` out of the VM state and computing it per Pending card in Compose from `timestamp` plus one shared time flow would avoid the global rebuild while keeping the live tick.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
aminsato: Good catch — the chip did freeze for a full minute past 60s. Switched to a combined `Xm Ys` formatter so the second hand keeps moving across the minute boundary (commit 8dc5617). Added `transaction_history_elapsed_minutes_and_seconds` to all 10 locales matching each one's existing minute-format pattern.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/components/TransactionHistoryComponents.kt
aminsato: Agreed — `BROADCASTED` covers the very window where the user just dispatched the tx, exactly when the counter is most useful. Folded `Pending` and `Broadcasted` into one `InProgressPill` keyed off the parent transaction's `timestamp`, matching iOS. Both states now render the same Row with a ticking elapsed segment.

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
aminsato: You were right about the cost. Pulled elapsed-time computation out of VM state entirely — `Pending` is now `data object` (no `elapsedTime` field), `startTimeTicker` and the `combine(currentTime)` are gone, so the entity list isn't re-mapped or re-grouped each second. The chip composable owns its own tick via `produceState` keyed on `timestamp`, so only the visible pill recomposes (and nothing recomposes when no in-progress chips are on screen).

rkokhatskyi (approved):
LGTM

PR #4617, opened 2026-05-21
fix(keygen): auto-start 2/2 and 3/3 peer discovery, hide Next button

### LINKED ISSUE #4552: Keygen peer discovery: auto-kickoff at 2/2 and 3/3, hide Next button (mirror Windows)
# [Fix] Auto-kickoff keygen and hide Next button at 2/2 and 3/3 [Keygen Peer Discovery]

## Problem
On the Keygen peer-discovery screen, secure-vault flows require the initiating device to tap **Next** even when the device count threshold has already been met (e.g. 2/2 or 3/3). The threshold is deterministic — there is no decision left for the user to make at that point. The `VsButton` in `PeerDiscoveryScreen.kt` is unconditionally rendered regardless of `deviceCount`, and `KeygenPeerDiscoveryViewModel` has no auto-kickoff logic for secure vault flows (only Fast Vault has one today).

## Expected Behavior
- **2/2 and 3/3 vaults**: keygen starts automatically as soon as all required peers connect. The Next button is hidden entirely, freeing space for the device list and QR code.
- **4+ device vaults (3-of-4, 3-of-5, etc.)**: Next button remains, because the user may want to choose which peers to commit before starting.

## Steps to Reproduce
1. Setup new vault → Secure → choose a 2-device or 3-device vault.
2. Scan from the second (and third) device until "Devices (2/2)" or "Devices (3/3)" is shown.
3. Observe that the Next button still requires a manual tap.
4. **Expected**: keygen starts automatically at threshold.

### Must Do
- [ ] In `KeygenPeerDiscoveryViewModel`, add a collector on `state.selectedDevices` that calls `next()` exactly once when `args.deviceCount != null && args.deviceCount <= 3` and `state.selectedDevices.size + 1 == args.deviceCount` (local device + peers == target). Guard with a boolean flag for idempotence.
- [ ] In `PeerDiscoveryScreen.kt`, wrap the `VsButton` in an `if (deviceCount > 3)` gate so it is hidden entirely for 2/2 and 3/3 layouts.

### Must NOT Do
- Don't change unrelated code
- Don't modify the Fast Vault auto-kickoff path (line 481 of `KeygenPeerDiscoveryViewModel.kt`) — it already works correctly
- Don't auto-kickoff for `deviceCount >= 4` vaults — the user must retain manual control over peer selection in those flows

## Acceptance Criteria
- [ ] On a 2-device secure vault, keygen starts automatically when the second device connects — no Next tap required
- [ ] On a 3-device secure vault, keygen starts automatically when the third device connects — no Next tap required
- [ ] On a 4+ device secure vault, the Next button remains visible and keygen does not auto-start
- [ ] Auto-kickoff fires exactly once even if the peer list briefly fluctuates around the threshold

## Context
**Windows reference implementation** (already ships this behavior):
- Auto-start logic: [`core/ui/mpc/keygen/peers/AutoStartKeygen.tsx`](https://github.com/vultisig/vultisig-windows/blob/main/core/ui/mpc/keygen/peers/AutoStartKeygen.tsx) — fires `onFinish()` when `targetDeviceCount ≤ 3 && selectedPeers.length === targetDeviceCount - 1`, guarded by a `hasStartedRef`.
- CTA gate: [`core/ui/mpc/keygen/peers/SecureVaultPeerDiscoveryScreen.tsx#L78`](https://github.com/vultisig/vultisig-windows/blob/main/core/ui/mpc/keygen/peers/SecureVaultPeerDiscoveryScreen.tsx#L78) — renders the button only when `targetDeviceCount > 3`.

**iOS**: a parallel ticket is being filed for the same gap (`KeygenPeerDiscoveryViewModel.startKeygen` only auto-fires for Fast Vault via `startFastVaultKeygenIfNeeded`).

**Affected files:**
- `app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt` (add auto-kickoff collector)
- `app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt:207-215` (gate `VsButton` on `deviceCount > 3`)

aminsato:
Closes #4552

## Summary
- On the Keygen peer-discovery screen, 2/2 and 3/3 secure vault flows required the initiator to manually tap **Next** even though the device threshold is deterministic and no selection decision remains.
- This PR adds an auto-kickoff collector in `KeygenPeerDiscoveryViewModel` that fires `next()` exactly once when `args.deviceCount` is set, ≤ 3, and the local device + selected peers reach the target. The collector is skipped for Fast Vault (`email` + `password` set), which already auto-fires from `startVultiServerConnection` — guarding against a double session-start race.
- In `PeerDiscoveryScreen`, the `VsButton` is wrapped in `if (state.deviceCount == null || state.deviceCount > 3)` so 2/2 and 3/3 layouts hide the button entirely (freeing space for the device list and QR), while 4+ device vaults and `deviceCount == null` flows (ReShare, Migrate) keep manual control.
- Mirrors the Windows reference implementation (`AutoStartKeygen.tsx`, `SecureVaultPeerDiscoveryScreen.tsx`).

## Test plan
- [x] `./gradlew assembleDebug` — passes
- [x] `:app:testDebugUnitTest` for `KeygenPeerDiscoveryViewModelTest` and `KeygenPeerDiscoveryViewModelReshareTest` — passes (6 new auto-kickoff cases: 2/2 fires, 3/3 fires only at threshold, 4/4 no-fire, Fast Vault no-fire, `deviceCount == null` no-fire, idempotent under selection fluctuation)
- [ ] Setup new vault → Secure → 2-device: keygen auto-starts when 2nd device joins, no Next tap
- [ ] Setup new vault → Secure → 3-device: keygen auto-starts when 3rd device joins, no Next tap
- [ ] Setup new vault → Secure → 4-device (3-of-4): Next button remains and keygen does not auto-start
- [ ] Fast Vault 2-device flow still completes via the existing server-discovery auto-fire (no double session start)

## Summary by CodeRabbit

* **New Features**
  * Auto-starts key generation when the required number of devices is selected for eligible secure vaults (2- or 3-device setups), with safeguards to skip auto-start when email/password are provided.

* **Improvements**
  * The "Next" button is hidden during eligible auto-start scenarios to streamline the flow.

* **Tests**
  * Added tests covering auto-kickoff behavior, thresholds, exclusions, and single-trigger guarantees.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4617?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -80,7 +80,9 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -99,6 +101,7 @@ data class PeerDiscoveryUiModel(
     val selectedDevices: List<String> = emptyList(),
     val minimumDevices: Int = MIN_KEYGEN_DEVICES,
     val minimumDevicesDisplayed: Int = MIN_KEYGEN_DEVICES,
+    val deviceCount: Int? = null,
     val allowsMoreDevices: Boolean = false,
     val showQrHelpModal: Boolean = false,
     val connectingToServer: ConnectingToServerUiModel? = null,
@@ -150,6 +153,7 @@ constructor(
             PeerDiscoveryUiModel(
                 minimumDevices = args?.deviceCount ?: MIN_KEYGEN_DEVICES,
                 minimumDevicesDisplayed = args?.deviceCount ?: MIN_KEYGEN_DEVICES,
+                deviceCount = args?.deviceCount,
                 enableNotification = false,
             )
         )
@@ -192,11 +196,40 @@ constructor(
 
     private var serverUrl: String = VULTISIG_RELAY_URL
 
+    private var hasAutoStartedKeygen: Boolean = false
+
     init {
         if (args == null) {
             viewModelScope.launch { navigator.navigate(Destination.Back) }
         } else {
             loadData()
+            observeAutoStartKeygen()
+        }
+    }
+
+    /**
+     * Auto-kicks off keygen for 2/2 and 3/3 secure-vault flows so the initiator does not have to
+     * tap Next once the deterministic peer threshold has been met. Mirrors the Windows
+     * `AutoStartKeygen` component. Skipped for Fast Vault, which already auto-fires from
+     * [startVultiServerConnection] when the server peer is discovered — double-firing would launch
+     * two concurrent session-start attempts.
+     */
+    private fun observeAutoStartKeygen() {
+        val args = args ?: return
+        val targetDeviceCount = args.deviceCount ?: return
+        if (targetDeviceCount > 3) return
+        if (!email.isNullOrBlank() && !password.isNullOrBlank()) return
+
+        viewModelScope.launch {
+            state
+                .map { it.selectedDevices.size }
+                .distinctUntilChanged()
+                .collect { selectedSize ->
+                    if (!hasAutoStartedKeygen && selectedSize + 1 == targetDeviceCount) {
+                        hasAutoStartedKeygen = true
+                        next()
+                    }
+                }
         }
     }
 
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -80,7 +80,6 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.update
@@ -196,8 +195,6 @@ constructor(
 
     private var serverUrl: String = VULTISIG_RELAY_URL
 
-    private var hasAutoStartedKeygen: Boolean = false
-
     init {
         if (args == null) {
             viewModelScope.launch { navigator.navigate(Destination.Back) }
@@ -217,19 +214,12 @@ constructor(
     private fun observeAutoStartKeygen() {
         val args = args ?: return
         val targetDeviceCount = args.deviceCount ?: return
-        if (targetDeviceCount > 3) return
+        if (targetDeviceCount !in 2..3) return
         if (!email.isNullOrBlank() && !password.isNullOrBlank()) return
 
         viewModelScope.launch {
-            state
-                .map { it.selectedDevices.size }
-                .distinctUntilChanged()
-                .collect { selectedSize ->
-                    if (!hasAutoStartedKeygen && selectedSize + 1 == targetDeviceCount) {
-                        hasAutoStartedKeygen = true
-                        next()
-                    }
-                }
+            state.map { it.selectedDevices.size }.first { it == targetDeviceCount - 1 }
+            next()
         }
     }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -192,11 +196,40 @@ constructor(
...
+     * Auto-kicks off keygen for 2/2 and 3/3 secure-vault flows so the initiator does not have to
+     * tap Next once the deterministic peer threshold has been met. Mirrors the Windows
+     * `AutoStartKeygen` component. Skipped for Fast Vault, which already auto-fires from
+     * [startVultiServerConnection] when the server peer is discovered — double-firing would launch
+     * two concurrent session-start attempts.
+     */
+    private fun observeAutoStartKeygen() {
+        val args = args ?: return
+        val targetDeviceCount = args.deviceCount ?: return
+        if (targetDeviceCount > 3) return
```
rkokhatskyi: The `if (targetDeviceCount > 3) return` only bounds the upper end whereas the initial `selectedDevices.size = 0` would fire the collector immediately for `targetDeviceCount = 1` since `0 + 1 == 1` matches before any peer is discovered, no current caller hits this since Fast Vault uses `deviceCount = 1` with the email/password skip below, so `if (targetDeviceCount !in 2..3) return` would close the gap without relying on every caller honoring the implicit invariant

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -192,11 +196,40 @@ constructor(
...
+        if (!email.isNullOrBlank() && !password.isNullOrBlank()) return
+
+        viewModelScope.launch {
```
rkokhatskyi, line 220: This collector keeps observing state long after the single fire so `hasAutoStartedKeygen` becomes a small state machine that has to be maintained correctly across success, failure, and retry paths, whereas `state.map { it.selectedDevices.size }.first { it + 1 == targetDeviceCount }.also { next() }` would terminate the coroutine at the firing point and remove the boolean entirely so the retry-stuck scenario disappears by construction

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -192,11 +196,40 @@ constructor(
...
+            state
+                .map { it.selectedDevices.size }
+                .distinctUntilChanged()
+                .collect { selectedSize ->
+                    if (!hasAutoStartedKeygen && selectedSize + 1 == targetDeviceCount) {
```
rkokhatskyi: `selectedSize + 1 == targetDeviceCount` is the local-plus-peers formulation whereas the Windows `AutoStartKeygen.tsx` reference uses `selectedPeers.length === requiredPeers` with `requiredPeers = targetDeviceCount - 1`, so `selectedSize == targetDeviceCount - 1` reads more directly and matches the cross-platform formulation

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -192,11 +196,40 @@ constructor(
...
+                        hasAutoStartedKeygen = true
+                        next()
```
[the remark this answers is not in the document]
aminsato: Adopted the `first { it == targetDeviceCount - 1 }` approach below — that drops the `hasAutoStartedKeygen` boolean entirely, so there's no guard left to get stuck on a failed `next()`. `tryAgain()` re-fires by construction. Fixed in 9180abba9.

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
aminsato: Good catch — switched to `if (targetDeviceCount !in 2..3) return` so the bound is explicit and doesn't rely on the Fast Vault skip below to cover `deviceCount = 1`. Fixed in 9180abba9.

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
aminsato: Agreed — switched to `state.map { it.selectedDevices.size }.first { it == targetDeviceCount - 1 }` followed by `next()`. The coroutine terminates at the firing point, the boolean is gone, and CodeRabbit's retry-stuck concern disappears by construction. Fixed in 9180abba9.

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
aminsato: Rewrote as `it == targetDeviceCount - 1` to match the Windows `AutoStartKeygen.tsx` formulation. Fixed in 9180abba9.

rkokhatskyi (approved):
LGTM

PR #4619, opened 2026-05-22
feat(swap): SwapKit integration

johnnyluo:
# SwapKit Phase 1 — EVM + Solana sources (Android)

Implements Phase 1 of the SwapKit aggregator integration tracked in issue #4614. Adds SwapKit as a competing quote source for EVM and Solana swaps, routed through Vultisig's proxy at `https://api.vultisig.com/swapkit-a/` (no API key on device — the proxy attaches it server-side). Gated behind a default-off `Settings → Advanced → SwapKit aggregator` DataStore flag, so behaviour is identical to today until a user opts in.

This is the integration branch for the five sub-tasks below. Merged-in sub-PRs:

- #4615 — task 1/5 (client, errors, models, config, cache) + folded-in task 4/5 toggle + task 5/5 decode/cache tests
- #4621 — task 2/5 (`SWAPKIT` enum + provider table + payload variant) and task 3/5 (`SwapKitQuoteSource` + EVM/Solana wiring)
- #4627 — non-EVM `SwapPayload.SwapKit` foundation + proto round-trip (Phase 2/3 scaffolding)
- #4637 — full typed `SwapKitError` catalogue + `SwapQuoteManager.mapSwapKitErrorToFormError` + 20 localized `swapkit_error_*` strings × 10 locales + `SwapFormViewModel` catch wiring

**CI status: failing.** `:data:testDebugUnitTest` has 2 failing tests after #4637 changed `SwapKitQuoteSource` throw semantics without updating the test expectations — see Blockers section below.

---

## Task status

| # | Task | Status |
|---|------|--------|
| 1 | API client, errors, models, config, provider cache | Done |
| 2 | `SWAPKIT` provider enum + provider table + payload variant + proto mappers | Done |
| 3 | `SwapKitQuoteSource` implementation + EVM/Solana signing dispatch | Done (architectural deviation from "edit `Swaps.kt`" — see below) |
| 4 | Advanced Settings toggle + localized strings | Done (20 `swapkit_error_*` keys × 10 locales + `settings_advanced_swapkit_toggle` × 10 locales) |
| 5 | Unit tests across spec'd suites | Partial (5 of 6 spec'd suites land; `SwapProviderTableTest` is missing; 2 existing tests are broken after #4637) |

---

### Task 1 — API client, errors, models, config, cache — Done

`data/src/main/kotlin/com/vultisig/wallet/data/`:

[...]

Issue #4394, opened 2026-05-04
Join Keysign: Transaction Details collapse should match extension


aminsato:
# [Fix] Join Keysign: Transaction Details UI should match extension [EVM contract call confirmation]

## Problem

On the Join Keysign confirmation screen for an EVM contract call, the screen header reads "Keysign" instead of "Sign Transaction", the "Transaction Details" collapse is rendered as a flush row inside the parent card rather than as its own rounded bordered container, and the function arguments label reads "Function Inputs" instead of "Function Arguments". The function argument value text color also does not match the green/teal used by the browser extension.

## Expected Behavior

The screen should visually and structurally match the browser extension's implementation ([`Collapse.tsx`](https://github.com/vultisig/vultisig-windows/blob/main/core/inpage-provider/popup/view/resolvers/signMessage/components/Collapse.tsx)):

| Element | Current (Android) | Expected |
|---|---|---|
| Screen header title | "Keysign" | "Sign Transaction" |
| Transaction Details container (collapsed) | Bare row inside parent card, no own border | Standalone rounded-bordered container |
| Transaction Details container (expanded) | Inline, no border around details block | Bordered container, details visually grouped inside |
| Function args label | "Function Inputs" | "Function Arguments" |
| Function args / signature value text color | Does not match extension | Green/teal token used by the extension |

## Steps to Reproduce

1. Open the Android app and initiate a keysign session that involves an EVM contract call.
2. Reach the Join Keysign confirmation screen.
3. Observe the screen header — it reads "Keysign" instead of "Sign Transaction".
4. Observe the "Transaction Details" row — it has no own border/container, sitting flush inside the parent card.
5. Expand the section — note the label "Function Inputs" and that the value text color does not match the extension's green/teal.

### Must Do
- [ ] Rename the screen header: in `JoinKeysignView.kt`, change the `else -> R.string.keysign` branch to use a new/updated string resource that reads **"Sign Transaction"** for this screen.
- [ ] Restyle `TransactionDetailsSection` in `VerifySendScreen.kt`: give the collapse a standalone rounded bordered container (matching the extension's visual frame) in both collapsed and expanded states.
[...]

Issue #4589, opened 2026-05-18
Integrate SwapKit as aggregator provider (Chainflip, NEAR Intents, Garden, Flashnet) for cross-chain swap fallback + new routes


realpaaao:
## Goal

Add SwapKit (https://docs.swapkit.dev) as a native Android swap provider alongside LI.FI / Kyber / 1inch / THORChain / Maya so the resolver gains:

1. **Redundancy** when THORChain / Maya halt or have no route (Chainflip, NEAR Intents, Garden, Flashnet, Harbor)
2. **New cross-chain routes** Vultisig can't do today (BTC↔TON, BTC↔SUI, BTC↔ADA via NEAR Intents; BTC↔BSC via Garden; BTC↔Tron/OP via Flashnet)
3. **Cheaper EVM aggregator path** than LI.FI (SwapKit ~15 bps vs LI.FI 25 bps)

Android calls SwapKit's REST API directly (no JS SDK consumed on Android).

## Scope (providers to surface)

All SwapKit cross-chain providers + in-chain DEX routes:
- Chainflip
- NEAR Intents
- Garden
- Flashnet
- Harbor
- 1inch / Uniswap / Jupiter via SwapKit

**Exclude:**
- SwapKit-routed THORChain (we have direct, would double-tax)
- SwapKit-routed MayaChain (same reason)

Filter `route.provider == thorchain || maya` out of SwapKit responses client-side.

## Architecture

Pattern matches existing `data/api/swapAggregators/KyberSwap.kt` and `OneInchSwap.kt`.

**New files:**
- `data/src/main/kotlin/com/vultisig/wallet/data/api/swapAggregators/SwapKit.kt` — REST client (`POST /quote`)
- `data/src/main/kotlin/com/vultisig/wallet/data/api/models/quotes/SwapKitQuoteJson.kt` — quote model
- `data/src/main/kotlin/com/vultisig/wallet/data/models/payload/SwapKitPayloadJson.kt` — signing payload

**Tx payload handling** — SwapKit returns multiple formats; map each to existing signing paths:
- EVM (Ethers v6 fields) → existing EVM signing
- UTXO base64 PSBT (BTC/LTC/DOGE/BCH) → existing UTXO signing path
- Cosmos native tx → existing Cosmos signing
- TRON TransactionBuilder → existing TRON signing
- Solana serialized message → existing Solana signing

**Wire-in:**
[...]

PR #4621, opened 2026-05-23
feat(swap): SwapKit task 2/5 — quote source + Phase 1 EVM/Solana wiring

aminsato:
## Summary

Task 2/5 of the SwapKit Android port — ports the iOS [#4394](https://github.com/vultisig/vultisig-ios/pull/4394) `SwapKitService` + Phase 1 wiring on top of #4619's client/cache/config foundation. **Base is `feature/swapkit`, not `main`** — diff is purely additive against task 1/5.

### What's in
- `SwapKitQuoteSource` — feature-flag + provider-cache gates, `/v3/quote` + `/v3/swap` two-step orchestrator, multi-hop / THORChain / Maya client-side filtering (Vultisig pays those affiliates directly via the existing native integrations, so routing through SwapKit would stack a second affiliate fee), rank-by-`expectedBuyAmount` picker, EVM/Solana `meta.txType` unwrap into the existing `EVMSwapQuoteJson` envelope.
- `SwapProvider.SWAPKIT` enum + display id + wire-id parsing. Phase 1 EVM/Solana routes ride the existing `oneinchSwapPayload` proto with `provider="swapkit"` as the discriminator (matches iOS — `SwapPayload.swift` comment: *"EVM and Solana SwapKit swaps still ride .generic since their wire shape matches OneInchSwapPayload 1:1"*).
- `SwapProviderTable` lists SwapKit alongside the existing EVM aggregators on ETH / BSC / AVAX / ARB / BASE / OP / POL plus Solana.
- `SwapQuoteManager` dispatch + a `swap_for_provider_swapkit` brand string (translatable=false, matches the other provider entries).
- Hilt `@IntoMap` binding for `SwapKitQuoteSource` in `RepositoriesModule`.

### What's deliberately out of scope (later tasks)
- Non-EVM signers (BTC PSBT, TRON, TON, SUI, ADA, DOGE/BCH/DASH/ZEC, XRP)
- `/track` polling + tx-history status
- Universal coin picker via `/v3/tokens`
- Tier-discounted affiliate fee (formula `max(0, min(1000, 50 - vultTierDiscount))`)
- `SwapKitSwapPayload` proto + the `SwapPayload.SwapKit` Kotlin variant — only needed once non-EVM signing lands. **No commondata bump in this PR.**

### Overlap with #4619
None. PR #4619 introduced `SwapKitApi`, the three DTO files, `SwapKitError`, `SwapKitConfig`, `SwapKitProviderCache`, their Hilt bindings, the settings toggle and locale strings; this PR **consumes** that surface and adds the quote source layered on top. The only shared file is `RepositoriesModule.kt` (each PR adds its own bindings — no line-level conflict because my branch is based on `feature/swapkit`).

## Which feature is affected?
- [x] Swap — adds SwapKit as a candidate aggregator (gated behind the existing default-off feature flag and the `/providers` cache)

## Verification
- [x] `./gradlew :data:compileDebugKotlin :app:compileDebugKotlin` — green
- [x] `./gradlew :data:test :app:testDebugUnitTest` — green
- [x] `./gradlew ktfmtCheck` — green
- [x] `./gradlew :app:lintDebug` — green
- [x] 12 new `SwapKitQuoteSourceTest` cases pin: feature-flag + cache gates, multi-hop / Thor / Maya filtering, ranking by `expectedBuyAmount`, asset-identifier formatting (`CHAIN.TICKER` / `CHAIN.TICKER-CONTRACT`, dotted decimal `sellAmount`, trailing-zero stripping so "1.0000" → "1"), EVM tx unwrap with gas defaults, Solana base64 in `data` field, `UnsupportedTxType` for non-Phase-1 `meta.txType`, exception → `SwapKitError.Network` wrapping with cause preservation, and `affiliateBps` passthrough.

## Checklist
- [x] I have added tests that prove my feature works

## Agent Metadata
- **Agent**: Claude Code (Opus 4.7, 1M context)
- **Issue**: parity port of vultisig-ios#4394 (Phase 1 quote-source slice)
- **Confidence**: medium-high — Phase 1 EVM/Solana parity verified by tests; live route fetch needs flag-on smoke test by a maintainer with proxy credentials
- **Needs human review for**: SwapKit quote-source contract assumptions (especially `formatSellAmount` decimals and the `expectedBuyAmount` ranking), Hilt graph layout, and the `SwapProviderTable` decision to include SwapKit unconditionally for in-cache chains (vs. gating at the table level too)

## Summary by CodeRabbit

* **New Features**
  * Introduced SwapKit as a new swap provider for token exchanges
  * SwapKit support enabled across multiple blockchain networks including EVM chains and Solana

* **Tests**
  * Added comprehensive test coverage for SwapKit quote sourcing and transaction handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4621?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -145,6 +145,18 @@ constructor(
                         vultBPSDiscount,
                         srcNativeToken,
                     )
+
+                SwapProvider.SWAPKIT ->
+                    fetchSwapKitQuote(
+                        src,
+                        dst,
+                        srcToken,
+                        dstToken,
+                        srcTokenValue,
+                        tokenValue,
+                        vultBPSDiscount,
+                        srcNativeToken,
+                    )
             }
 
         val feeCoin =
@@ -152,6 +164,11 @@ constructor(
                 SwapProvider.MAYA,
                 SwapProvider.THORCHAIN,
                 SwapProvider.LIFI -> dstToken
+                // SwapKit's inbound (source-chain) fee is denominated in the source's native gas
+                // token — mirrors 1inch / Kyber / Jupiter rather than the LiFi destination-side
+                // integrator-fee model. Per-leg fee parsing from response.fees[] lands with the
+                // tier-discount affiliate work in a later task.
+                SwapProvider.SWAPKIT -> srcNativeToken
                 else -> srcNativeToken
             }
 
@@ -551,6 +568,63 @@ constructor(
                 R.string.swap_for_provider_jupiter.asUiText()
             }
         return swapQuote to providerText
+    }
+
+    private suspend fun fetchSwapKitQuote(
+        src: SendSrc,
+        dst: SendSrc,
+        srcToken: Coin,
+        dstToken: Coin,
+        srcTokenValue: BigInteger,
+        tokenValue: TokenValue,
+        vultBPSDiscount: Int?,
+        srcNativeToken: Coin,
+    ): Pair<SwapQuote, UiText> {
+        val swapQuote =
+            getCachedQuoteOrFetch(srcToken.id, dstToken.id, srcTokenValue, SwapProvider.SWAPKIT) {
+                val apiQuote =
+                    swapQuoteRepository
+                        .getQuote(
+                            SwapProvider.SWAPKIT,
+                            SwapQuoteRequest(
+                                srcToken = srcToken,
+                                dstToken = dstToken,
+                                tokenValue = tokenValue,
+                                srcAddress = src.address.address,
+                                dstAddress = dst.address.address,
+                                // Tier-discounted affiliate fee lands in its own follow-up task —
+                                // matches iOS's max(0, min(1000, 50 - vultTierDiscount)) formula.
+                                // For now pass zero so the proxy applies its dashboard-configured
+                                // default rather than a stale client override.
+                                affiliateBps = 0,
+                                bpsDiscount = vultBPSDiscount ?: 0,
+                            ),
+                        )
+                        .expectEvm(SwapProvider.SWAPKIT)
+                val expectedDstValue =
+                    TokenValue(
+                        value = apiQuote.dstAmount.toBigIntegerOrNull() ?: BigInteger.ZERO,
+                        token = dstToken,
+                    )
+                // SwapKit doesn't return a typed swapFee inside the EVM tx envelope (the
+                // response-level fees[] array carries the per-leg breakdown — surfaced as part of
+                // the affiliate-fee task). Use the gas envelope as the inbound-side fee estimate,
+                // matching the 1inch/Kyber gas-only pattern.
+                val gasUnits =
+                    apiQuote.tx.gas.takeIf { it != 0L } ?: EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT
+                val gasFees =
+                    (apiQuote.tx.gasPrice.toBigIntegerOrNull() ?: BigInteger.ZERO) *
+                        gasUnits.toBigInteger()
+                val tokenFees = TokenValue(value = gasFees, token = srcNativeToken)
+                SwapQuote.OneInch(
+                    expectedDstValue = expectedDstValue,
+                    fees = tokenFees,
+                    data = apiQuote,
+                    expiredAt = Clock.System.now() + expiredAfter,
+                    provider = SwapProvider.SWAPKIT.getSwapProviderId(),
+                )
+            }
+        return swapQuote to R.string.swap_for_provider_swapkit.asUiText()
     }
 
     private suspend fun resolveSwapFee(
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -616,15 +616,18 @@ constructor(
                     )
                 // SwapKit's canonical source-chain fee is the `fees[]` entry with type=="inbound"
                 // for the source chain — the source layer already extracts that and stages it on
-                // `tx.swapFee` (raw units) with `swapFeeTokenContract = ""` (native sentinel). The
-                // resolveSwapFee helper falls back to srcNativeToken on the empty-contract case,
-                // matching how Kyber and Jupiter surface their per-leg fees.
+                // `tx.swapFee` (raw units in source-native) with `swapFeeTokenContract = ""`. The
+                // empty-contract branch of resolveSwapFee returns the fallback in srcNativeToken,
+                // so pass the parsed swapFee as the fallback (mirrors Kyber's call at line 435
+                // which passes gasFees) — otherwise the source-extracted inbound fee is silently
+                // discarded and the verify screen renders $0 Network Fee.
+                val swapKitInboundFee = apiQuote.tx.swapFee.toBigIntegerOrNull() ?: BigInteger.ZERO
                 val (feeAmount, feeCoin) =
                     resolveSwapFee(
                         apiQuote.tx.swapFeeTokenContract,
                         apiQuote.tx.swapFee,
                         srcNativeToken,
-                        BigInteger.ZERO,
+                        swapKitInboundFee,
                     )
                 val tokenFees = TokenValue(value = feeAmount, token = feeCoin)
                 SwapQuote.OneInch(
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -15,6 +15,7 @@ import com.vultisig.wallet.data.models.settings.AppCurrency
 import com.vultisig.wallet.data.repositories.SwapQuoteRepository
 import com.vultisig.wallet.data.repositories.TokenRepository
 import com.vultisig.wallet.data.repositories.swap.SwapQuoteRequest
+import com.vultisig.wallet.data.repositories.swap.SwapQuoteResult
 import com.vultisig.wallet.data.repositories.swap.convertToTokenValue
 import com.vultisig.wallet.data.usecases.ConvertTokenToToken
 import com.vultisig.wallet.data.usecases.ConvertTokenValueToFiatUseCase
@@ -166,8 +167,8 @@ constructor(
                 SwapProvider.LIFI -> dstToken
                 // SwapKit's inbound (source-chain) fee is denominated in the source's native gas
                 // token — mirrors 1inch / Kyber / Jupiter rather than the LiFi destination-side
-                // integrator-fee model. Per-leg fee parsing from response.fees[] lands with the
-                // tier-discount affiliate work in a later task.
+                // integrator-fee model. SwapKitQuoteSource already pulls the inbound amount out of
+                // route.fees[] and stages it on tx.swapFee (raw units) for resolveSwapFee below.
                 SwapProvider.SWAPKIT -> srcNativeToken
                 else -> srcNativeToken
             }
@@ -580,56 +581,80 @@ constructor(
         vultBPSDiscount: Int?,
         srcNativeToken: Coin,
     ): Pair<SwapQuote, UiText> {
+        // iOS' SwapKit tier-discount formula at this milestone:
+        //   max(0, min(1000, 50 - vultTierDiscount))
+        // 50 bps base affiliate, clamped to 0..1000 (SwapKit's documented 0..10% range).
+        val affiliateBps = (SWAPKIT_AFFILIATE_FEE_BPS - (vultBPSDiscount ?: 0)).coerceIn(0, 1000)
+        var resolvedSubProvider: String? = null
         val swapQuote =
             getCachedQuoteOrFetch(srcToken.id, dstToken.id, srcTokenValue, SwapProvider.SWAPKIT) {
-                val apiQuote =
-                    swapQuoteRepository
-                        .getQuote(
-                            SwapProvider.SWAPKIT,
-                            SwapQuoteRequest(
-                                srcToken = srcToken,
-                                dstToken = dstToken,
-                                tokenValue = tokenValue,
-                                srcAddress = src.address.address,
-                                dstAddress = dst.address.address,
-                                // Tier-discounted affiliate fee lands in its own follow-up task —
-                                // matches iOS's max(0, min(1000, 50 - vultTierDiscount)) formula.
-                                // For now pass zero so the proxy applies its dashboard-configured
-                                // default rather than a stale client override.
-                                affiliateBps = 0,
-                                bpsDiscount = vultBPSDiscount ?: 0,
-                            ),
-                        )
-                        .expectEvm(SwapProvider.SWAPKIT)
+                val result =
+                    swapQuoteRepository.getQuote(
+                        SwapProvider.SWAPKIT,
+                        SwapQuoteRequest(
+                            srcToken = srcToken,
+                            dstToken = dstToken,
+                            tokenValue = tokenValue,
+                            srcAddress = src.address.address,
+                            dstAddress = dst.address.address,
+                            affiliateBps = affiliateBps,
+                        ),
+                    )
+                val evmResult =
+                    (result as? SwapQuoteResult.Evm)
+                        ?: error("SwapKitQuoteSource must return SwapQuoteResult.Evm")
+                resolvedSubProvider = evmResult.subProvider
+                val apiQuote = evmResult.data
                 val expectedDstValue =
                     TokenValue(
-                        value = apiQuote.dstAmount.toBigIntegerOrNull() ?: BigInteger.ZERO,
+                        value =
+                            apiQuote.dstAmount.toBigIntegerOrNull()
+                                ?: error(
+                                    "Malformed SwapKit dstAmount (raw units expected): ${apiQuote.dstAmount}"
+                                ),
                         token = dstToken,
                     )
-                // SwapKit doesn't return a typed swapFee inside the EVM tx envelope (the
-                // response-level fees[] array carries the per-leg breakdown — surfaced as part of
-                // the affiliate-fee task). Use the gas envelope as the inbound-side fee estimate,
-                // matching the 1inch/Kyber gas-only pattern.
-                val gasUnits =
-                    apiQuote.tx.gas.takeIf { it != 0L } ?: EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT
-                // Parse gasPrice through BigDecimal so a decimal-string upstream (e.g. "20.5"
-                // — valid JSON, invalid BigInteger) truncates to integer wei rather than silently
-                // falling back to ZERO. A zero gas fee would make SwapKit look free in the ranking
-                // pass and could let the EVM signer broadcast an underpriced tx.
-                val gasFees =
-                    (apiQuote.tx.gasPrice.toBigDecimalOrNull()?.toBigInteger() ?: BigInteger.ZERO) *
-                        gasUnits.toBigInteger()
-                val tokenFees = TokenValue(value = gasFees, token = srcNativeToken)
+                // SwapKit's canonical source-chain fee is the `fees[]` entry with type=="inbound"
+                // for the source chain — the source layer already extracts that and stages it on
+                // `tx.swapFee` (raw units) with `swapFeeTokenContract = ""` (native sentinel). The
+                // resolveSwapFee helper falls back to srcNativeToken on the empty-contract case,
+                // matching how Kyber and Jupiter surface their per-leg fees.
+                val (feeAmount, feeCoin) =
+                    resolveSwapFee(
+                        apiQuote.tx.swapFeeTokenContract,
+                        apiQuote.tx.swapFee,
+                        srcNativeToken,
+                        BigInteger.ZERO,
+                    )
+                val tokenFees = TokenValue(value = feeAmount, token = feeCoin)
                 SwapQuote.OneInch(
                     expectedDstValue = expectedDstValue,
                     fees = tokenFees,
                     data = apiQuote,
                     expiredAt = Clock.System.now() + expiredAfter,
+                    // `provider` is the proto-serialized discriminator used by
+                    // SwapTransactionToUiModelMapper to map back onto SwapProvider.SWAPKIT —
+                    // keep it as the canonical id. The sub-provider drives the UI label below.
                     provider = SwapProvider.SWAPKIT.getSwapProviderId(),
                 )
             }
-        return swapQuote to R.string.swap_for_provider_swapkit.asUiText()
+        val providerLabel =
+            resolvedSubProvider
+                ?.takeIf { it.isNotBlank() }
+                ?.let { sub -> "SwapKit (${formatSubProvider(sub)})".asUiText() }
+                ?: R.string.swap_for_provider_swapkit.asUiText()
+        return swapQuote to providerLabel
     }
+
+    /**
+     * Render a SwapKit sub-provider id (`CHAINFLIP`, `near_intents`, `flashnet`) into a display
+     * label. Replaces underscores with spaces and title-cases each token so `near_intents` → `Near
+     * Intents`; brand names already in mixed/upper case are left alone.
+     */
+    private fun formatSubProvider(raw: String): String =
+        raw.split('_', '-').joinToString(" ") { token ->
+            token.replaceFirstChar { ch -> if (ch.isLowerCase()) ch.titlecase() else ch.toString() }
+        }
 
     private suspend fun resolveSwapFee(
         swapFeeTokenContract: String,
@@ -757,6 +782,9 @@ constructor(
 
     companion object {
         private const val KYBER_AFFILIATE_FEE_BPS = 50
+        // Mirrors iOS' SwapKit milestone: 50 bps base affiliate, discounted by the user's
+        // vultTierDiscount, clamped to SwapKit's documented 0..1000 bps (0..10%) range.
+        private const val SWAPKIT_AFFILIATE_FEE_BPS = 50
         private const val NATIVE_TOKEN_SENTINEL = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
         private const val QUOTE_FETCH_TIMEOUT_MS = 15_000L
     }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -612,8 +612,12 @@ constructor(
                 // matching the 1inch/Kyber gas-only pattern.
                 val gasUnits =
                     apiQuote.tx.gas.takeIf { it != 0L } ?: EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT
+                // Parse gasPrice through BigDecimal so a decimal-string upstream (e.g. "20.5"
+                // — valid JSON, invalid BigInteger) truncates to integer wei rather than silently
+                // falling back to ZERO. A zero gas fee would make SwapKit look free in the ranking
+                // pass and could let the EVM signer broadcast an underpriced tx.
                 val gasFees =
-                    (apiQuote.tx.gasPrice.toBigIntegerOrNull() ?: BigInteger.ZERO) *
+                    (apiQuote.tx.gasPrice.toBigDecimalOrNull()?.toBigInteger() ?: BigInteger.ZERO) *
                         gasUnits.toBigInteger()
                 val tokenFees = TokenValue(value = gasFees, token = srcNativeToken)
                 SwapQuote.OneInch(
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -286,6 +286,7 @@
     <string name="swap_for_provider_kyber" translatable="false">Kyber</string>
     <string name="swap_for_provider_li_fi" translatable="false">LI.FI</string>
     <string name="swap_for_provider_jupiter" translatable="false">Jupiter</string>
+    <string name="swap_for_provider_swapkit" translatable="false">SwapKit</string>
     <string name="backup_password_screen_title">Backup</string>
     <string name="import_file_screen_password_error">Incorrect password, try again</string>
     <string name="add_vault_save">Save</string>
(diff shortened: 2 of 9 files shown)


aminsato:
@coderabbitai full review

app/src/main/res/values/strings.xml
```diff
@@ -286,6 +286,7 @@
     <string name="swap_for_provider_kyber" translatable="false">Kyber</string>
     <string name="swap_for_provider_li_fi" translatable="false">LI.FI</string>
     <string name="swap_for_provider_jupiter" translatable="false">Jupiter</string>
+    <string name="swap_for_provider_swapkit" translatable="false">SwapKit</string>
```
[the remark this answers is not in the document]
aminsato: Skipping this — the  brand-name strings live only in `values/strings.xml` by existing convention. Verified:

```bash
$ grep -l 'swap_for_provider_1inch\|swap_for_provider_kyber\|swap_for_provider_li_fi\|swap_for_provider_jupiter' app/src/main/res/values*/strings.xml
app/src/main/res/values/strings.xml
```

`swap_for_provider_swapkit` follows the same shape as `swap_for_provider_1inch`, `swap_for_provider_kyber`, `swap_for_provider_li_fi`, and `swap_for_provider_jupiter` — none of which are propagated to the 10 locale dirs. The localized strings rule from `CLAUDE.md` applies to translatable copy; aggregator brand names are intentionally pinned with `translatable="false"` so lint doesn't flag them as missing translations.

If you'd like every brand string copied into all 10 locale files for symmetry, that's a separate cleanup PR covering 1inch/Kyber/LI.FI/Jupiter too — out of scope for the SwapKit task 2/5 slice.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -0,0 +1,274 @@
...
+     * Vultisig pays those affiliates directly via its existing native integrations, so routing them
+     * through SwapKit would stack a second affiliate fee.
+     */
+    private fun pickBestRoute(routes: List<SwapKitRoute>): SwapKitRoute? {
+        val filtered =
+            routes.filter { route ->
+                if (route.providers.size != 1) return@filter false
+                val primary = route.primaryProviderId
+                primary !in FILTERED_PROVIDERS
+            }
```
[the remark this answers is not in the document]
aminsato: Skipping — `route.primaryProviderId` is already lower-cased at the DTO level, so the membership test against `FILTERED_PROVIDERS` is already case-insensitive. See `SwapKitQuoteJson.kt:87-88` (already on `feature/swapkit` from #4619):

```kotlin
/** Lower-cased provider id used by the THORChain/Maya filter and the sub-provider tag. */
val primaryProviderId: String
    get() = providers.firstOrNull().orEmpty().lowercase()
```

The test suite covers the upper-case case explicitly — `fetch filters out multi-hop and Thor and Maya routes then picks highest expected amount` feeds in routes with `providers = listOf("THORCHAIN")`, `listOf("MAYACHAIN")`, etc. and verifies they are filtered. Adding a second `.lowercase()` at the call site would be a no-op and would shadow the contract documented on the getter.

aminsato:
@coderabbitai proceed

aminsato:
Thanks for the deep review @coderabbitai! Triaged all 8 findings:

### Fixed (pushed in `c43cc7e0`)

| # | Finding | Resolution |
|---|---------|-----------|
| **2** 🟠 | Decimal `gasPrice` silently zeroes fee | Parse through `toBigDecimalOrNull()?.toBigInteger()` so `"20.5"` truncates to integer wei instead of falling back to ZERO. Comment documents the intent. |
| **3** 🟠 | `FILTERED_PROVIDERS` case mismatch | Added explicit `.lowercase()` at the filter call site (was already canonical via the DTO getter, but belt-and-braces). |
| **4** 🟡 | `assetIdentifier` fallback to `coin.ticker` | Now throws `SwapKitError.NoRoutes` when the chain prefix is unmapped — surfaces clean per-provider drop instead of letting the proxy reject with `helpers_invalid_asset_identifier`. New test `fetch throws NoRoutes when source chain has no SwapKit prefix mapping` pins it. |
| **5** 🟡 | `SwapQuoteResult.Evm` for Solana | Inline comment at the return site explaining the Evm envelope is intentional (matches `JupiterQuoteSource`). |
| **6** 🟢 | `swapId` not stored | `TODO(swapkit /track polling, later task)` at the `/v3/swap` call site so it is not forgotten when tx-history Phase D lands. |

### Not actionable

| # | Finding | Reason |
|---|---------|-------|
| **1** 🔴 | `.type` vs `.txType` mismatch | **False positive — added a clarifying comment.** `SwapKitTxMeta.type` is a *computed lowercase property* (`val type: String = txType.lowercase()`) introduced by #4619 specifically so the `when` matches `TYPE_EVM`/`TYPE_SOLANA` lowercase constants. `.txType` (raw casing) is used only in the `UnsupportedTxType(response.meta.txType)` error so diagnostics surface the wire value verbatim — e.g. `"PSBT"` not `"psbt"`. Both accessors are intentional; the comment at line 174 now spells this out. |
| **7** 🟢 | `affiliateBps = 0` should reference task ticket | The existing comment already references the iOS formula and notes it is a follow-up. No GitHub issue exists yet for the affiliate-fee work (it is one of the remaining slices of #4589); adding a stale ticket reference would be worse than the current comment. Will pin to a concrete issue once the affiliate task is filed. |
| **8** 🟢 | Missing unit test for decimal gasPrice | The fix lives inside `SwapQuoteManager.fetchSwapKitQuote`, which is a private method on a class with 7 collaborator dependencies and no existing unit-test scaffold. Adding one purely for this single defensive branch is more scaffolding than the fix warrants; the inline comment captures the contract. Happy to revisit if `SwapQuoteManager` ever grows a base test class. |

Test count: 12 → 16. ktfmt + `:data:test` + `:app:compileDebugKotlin` all green.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -0,0 +1,324 @@
...
+                            // does not parse those out here — gas math lands via the standard EVM
+                            // path and the inbound-fee surfacing belongs with the tier-discount
+                            // affiliate work in a later task.
+                            swapFee = "0",
+                            swapFeeTokenContract = "",
+                        ),
+                )
+            }
+            SwapKitTxMeta.TYPE_SOLANA -> {
+                val solana = decode<SwapKitSolanaTx>(response.tx, "solana")
```
rkokhatskyi: SwapKit V3 returns the Solana `tx` as a base64 String directly (verified against the iOS fixture `v3-sol-near-swap-fresh.json` and the canonical decoder at `SwapKitSwapResponse.swift:217`) whereas `decode<SwapKitSolanaTx>(response.tx, ...)` here expects an object with `swapTransaction` or `message` fields, so kotlinx-serialization throws on every real Solana response and the resulting `Decoding` bubble makes SwapKit silently unusable for Solana sources. The unit-test fixture at `SwapKitQuoteSourceTest.kt:191-196` wraps the base64 in `buildJsonObject` which masks the bug, so the Solana branch should read `(response.tx as? JsonPrimitive)?.content` directly and additionally accept `SERIALIZED_BASE64` alongside `solana` as a `meta.txType` since SwapKit upstream has flipped that discriminator once before per iOS commit `382b28f5f`

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -0,0 +1,324 @@
...
+            throw SwapKitError.Decoding(
+                message = "Failed to decode SwapKit $label tx payload",
+                cause = e,
+            )
+        }
+
+    private companion object {
+        /** Sub-provider ids (lower-cased) filtered out of `/v3/quote` results client-side. */
+        private val FILTERED_PROVIDERS: Set<String> =
+            setOf("thorchain", "thorchain-streaming", "mayachain", "maya")
```
rkokhatskyi: SwapKit `/providers` and the route `providers[]` carry ids like `THORCHAIN_STREAMING` and `MAYACHAIN_STREAMING` with underscores (verified against the iOS provider fixture `02-providers.json`) which the DTO getter lower-cases to `thorchain_streaming` and `mayachain_streaming`, so the dashed `thorchain-streaming` plus the bogus 4-letter `maya` here never match the wire ids and both streaming variants leak past the filter, stacking a SwapKit affiliate fee on top of the native Thor/Maya integrations this set is meant to prevent.

```kotlin
private val FILTERED_PROVIDERS: Set<String> =
    setOf("thorchain", "thorchain_streaming", "mayachain", "mayachain_streaming")
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -553,6 +570,67 @@ constructor(
...
+                // the affiliate-fee task). Use the gas envelope as the inbound-side fee estimate,
+                // matching the 1inch/Kyber gas-only pattern.
+                val gasUnits =
+                    apiQuote.tx.gas.takeIf { it != 0L } ?: EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT
+                // Parse gasPrice through BigDecimal so a decimal-string upstream (e.g. "20.5"
+                // — valid JSON, invalid BigInteger) truncates to integer wei rather than silently
+                // falling back to ZERO. A zero gas fee would make SwapKit look free in the ranking
+                // pass and could let the EVM signer broadcast an underpriced tx.
+                val gasFees =
+                    (apiQuote.tx.gasPrice.toBigDecimalOrNull()?.toBigInteger() ?: BigInteger.ZERO) *
```
rkokhatskyi: `gasPrice.toBigDecimalOrNull()?.toBigInteger() ?: BigInteger.ZERO` swallows any non-decimal upstream shape (`0x…` hex, null, blank, scientific) into 0 wei for the displayed Network Fee while the same `gasPrice` string is forwarded verbatim into `OneInchSwapTxJson.gasPrice` and consumed downstream by the EVM signer, so the verify screen shows a $0 Network Fee while the broadcast can still cost real wei. A typed `SwapKitError.Decoding` on parse failure would let the picker fall back to another aggregator rather than rank a misleading quote

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -0,0 +1,324 @@
...
+        // before any /quote network I/O so the proxy stays cold until the user opts in.
+        if (!config.isFeatureEnabled.first()) {
+            throw SwapKitError.NoRoutes("SwapKit feature flag disabled")
+        }
+
+        // Provider-cache gate — both ends must be enabled in the cached /providers snapshot.
+        // Fails closed: when the cache can't be hydrated isEnabled returns false, so we skip
+        // SwapKit and let another provider take the route rather than firing a /quote that we
+        // already expect to fail.
+        if (!providerCache.isEnabled(request.srcToken.chain)) {
```
rkokhatskyi: The cache gate translates a cold or failed `/providers` snapshot into `NoRoutes`, whereas iOS at `SwapKitProviderCache.swift:88-91` returns `true` on cache-miss and lets `/v3/quote` surface the real error, so a single bad-network app launch on Android silently hides SwapKit from every swap until the next refresh while the Advanced Settings toggle still shows on with no signal that the cache is stale

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -0,0 +1,324 @@
...
+                            data = evm.data,
+                            gas = gas,
+                            value = evm.value,
+                            gasPrice = evm.gasPrice.orEmpty(),
+                            // Per-leg swap fees on SwapKit are surfaced via the response-level
+                            // `fees[]` array (type=="inbound"|"outbound"|"affiliate"|…). Phase 1
+                            // does not parse those out here — gas math lands via the standard EVM
+                            // path and the inbound-fee surfacing belongs with the tier-discount
+                            // affiliate work in a later task.
+                            swapFee = "0",
```
rkokhatskyi: The canonical source-chain fee on a SwapKit response is the `fees[]` entry where `type=="inbound"` and `chain` matches the source coin prefix (e.g. `0.000005` SOL on the live `v3-sol-near-swap-fresh` fixture) which iOS surfaces via `SwapKitService.inboundFee` at this same milestone, whereas hard-coding `swapFee="0"` here delegates fee math to `SwapQuoteManager` which computes `gas*gasPrice` instead and produces zero on Solana since the Solana branch sets `gas=0, gasPrice=0` and misses the cross-chain inbound charge on every Chainflip / NEAR route. The deferred-parsing comment leaves the user with a Network Fee line that cannot reconcile against the actual on-chain debit

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -0,0 +1,324 @@
...
+        // matched against TYPE_EVM / TYPE_SOLANA. `meta.txType` (raw upstream casing) is used only
+        // in the UnsupportedTxType message so diagnostics surface the wire value verbatim — e.g.
+        // "PSBT" not "psbt" — without affecting dispatch.
+        val txType = response.meta.type
+        return when (txType) {
+            SwapKitTxMeta.TYPE_EVM -> {
+                val evm = decode<SwapKitEvmTx>(response.tx, "evm")
+                val gas =
+                    evm.gas?.toLongOrNull()?.takeIf { it > 0 }
+                        ?: EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT
```
rkokhatskyi: `EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT = 600000L` is sized for an L1 Ethereum DEX swap so using it as the gas-units fallback for every SwapKit-eligible EVM chain (Arbitrum, Optimism, Base, BSC, Polygon, Avalanche) over-estimates the displayed network fee by several multiples on the L2s whenever a route legitimately omits `tx.gas`. Either refuse routes missing `tx.gas` and trust the upstream value, or thread a per-chain default through this path

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -553,6 +570,67 @@ constructor(
...
+                        gasUnits.toBigInteger()
+                val tokenFees = TokenValue(value = gasFees, token = srcNativeToken)
+                SwapQuote.OneInch(
+                    expectedDstValue = expectedDstValue,
+                    fees = tokenFees,
+                    data = apiQuote,
+                    expiredAt = Clock.System.now() + expiredAfter,
+                    provider = SwapProvider.SWAPKIT.getSwapProviderId(),
```
rkokhatskyi, line 641: `provider = SwapProvider.SWAPKIT.getSwapProviderId()` stamps every SwapKit quote with the static `SwapKit` label so the verify screen never surfaces the route's sub-provider (Chainflip, NEAR Intents, Flashnet, Garden) whereas iOS exposes `response.providers.first` as the `subProvider` for the verify row, leaving the Android user without the routing info needed to reason about expected ETA or to debug a failed cross-chain swap

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -553,6 +570,67 @@ constructor(
...
+                                srcToken = srcToken,
+                                dstToken = dstToken,
+                                tokenValue = tokenValue,
+                                srcAddress = src.address.address,
+                                dstAddress = dst.address.address,
+                                // Tier-discounted affiliate fee lands in its own follow-up task —
+                                // matches iOS's max(0, min(1000, 50 - vultTierDiscount)) formula.
+                                // For now pass zero so the proxy applies its dashboard-configured
+                                // default rather than a stale client override.
+                                affiliateBps = 0,
```
rkokhatskyi: `affiliateBps = 0` zero-rates the affiliate fee on every SwapKit quote while `bpsDiscount = vultBPSDiscount ?: 0` on the next line is dead-passed since `SwapKitQuoteSource` reads only `affiliateBps` from `SwapQuoteRequest`, so the candidate's `vultBPSDiscount` value carries no signal and the `fetch passes affiliateBps from the request through` test at `SwapKitQuoteSourceTest.kt:349-366` pins a contract no caller exercises. Either drop `bpsDiscount` here until the tier-discount task lands, or wire `affiliateBps` through the documented `max(0, min(1000, 50 - vultBPSDiscount))` formula iOS uses at this same milestone

rkokhatskyi (changes requested):

aminsato:
@rkokhatskyi thank you for the detailed review — all 8 findings addressed in `9a60118b6`. Triage below:

### Critical correctness
- **#1 Solana decode (`SwapKitQuoteSource.kt:222`).** Confirmed: SwapKit V3 returns Solana `tx` as a bare base64 string. The old `decode<SwapKitSolanaTx>(...)` would throw `Decoding` on every real response; only `buildJsonObject` in the test fixture was hiding it. New `solanaBase64()` reads `JsonPrimitive.content` directly, falls back to the legacy object shape for forward compat, and `txTypeOf()` aliases `SERIALIZED_BASE64` → Solana so the iOS-flagged flip is covered. Three new tests pin the contract (JsonPrimitive path, legacy object path, SERIALIZED_BASE64 discriminator).
- **#2 `FILTERED_PROVIDERS` (`SwapKitQuoteSource.kt:314`).** Switched to wire-format ids: `setOf("thorchain", "thorchain_streaming", "mayachain", "mayachain_streaming")`. Filter test now feeds in `THORCHAIN_STREAMING` and `MAYACHAIN_STREAMING` routes and verifies they are dropped.

### Correctness
- **#4 Cache gate (`SwapKitQuoteSource.kt:82`).** Removed the gate entirely (and dropped the `providerCache` constructor injection). `/v3/quote` is now the authority on unsupported chains, matching iOS fail-open semantics. The two `NoRoutes` tests for the gate are gone.
- **#5 Inbound fee from `route.fees[]` (`SwapKitQuoteSource.kt:216`).** New `inboundFeeRawUnits(srcToken, route)` finds the `type=="inbound"` entry where `chain` matches the source-chain prefix, converts the decimal amount to raw native units, and embeds it on `tx.swapFee` with `swapFeeTokenContract = ""` (native sentinel). `SwapQuoteManager.fetchSwapKitQuote` now resolves the fee through `resolveSwapFee` (matches Kyber/Jupiter) instead of the old gas-math envelope. Two tests cover the path: `surfaces inbound fee from route fees list in raw source-chain units` (0.000005 SOL → 5000 lamports) and `zero inbound fee when fees list lacks an inbound entry for the source chain`.
- **#6 `DEFAULT_ETH_SWAP_GAS_UNIT` fallback (`SwapKitQuoteSource.kt:200`).** Replaced the L1-sized fallback with `throw SwapKitError.Decoding("SwapKit EVM tx missing gas estimate ...")`. The picker drops SwapKit and ranks another aggregator with a real estimate. New test `fetch refuses EVM route when tx_gas is missing` pins it.

### UX
- **#7 Sub-provider label (`SwapQuoteManager.kt:628`).** `SwapQuoteResult.Evm` now carries `subProvider: String?`. The source populates it from `swapResponse.providers[0]` (with `meta.subProvider` fallback for forward compat). `SwapQuoteManager.fetchSwapKitQuote` builds the verify-row label as `SwapKit (Chainflip)`, `SwapKit (Near Intents)`, etc. via a small `formatSubProvider` helper. The proto discriminator stays `SwapKit` so `SwapTransactionToUiModelMapper` still resolves the payload back to `SwapProvider.SWAPKIT`. Three tests cover passthrough, `meta.subProvider` fallback, and the null case.
- **#8 Affiliate fee + dead bpsDiscount (`SwapQuoteManager.kt:599`).** Wired iOS formula: `(50 - vultBPSDiscount).coerceIn(0, 1000)`. The dead `bpsDiscount` argument is removed; `affiliateBps` now carries the tier-discounted value to `SwapKitQuoteSource`. The existing `passes affiliateBps from the request through` test stays valid but now exercises a contract the call site actually drives.

### Resolved indirectly
- **#3 `gasPrice` zero-fallback in `SwapQuoteManager.kt:620`.** Obviated by the #5 refactor — the displayed fee no longer flows through gas math at all; it comes from `swapFee` via `resolveSwapFee`. The broadcast path keeps using `tx.gasPrice` verbatim, where the EVM signer handles malformed values at signing time (loud failure, not silent zero).

### Test delta
12 → 20 tests, all green. ktfmt + `:data:test` + `:app:testDebugUnitTest` + `:app:lintDebug` clean.

Ready for re-review.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Fixed in `9a60118b6`. `solanaBase64()` now reads the base64 directly via `(element as? JsonPrimitive)?.takeIf { it.isString }?.content` first and falls back to the legacy `SwapKitSolanaTx` object shape for forward compat. `txTypeOf()` aliases `SERIALIZED_BASE64` → Solana. Three new tests pin the contract: `fetch decodes Solana base64 from JsonPrimitive at the tx root`, `fetch decodes legacy Solana object shape with swapTransaction field`, and `fetch accepts SERIALIZED_BASE64 txType as Solana`. The previous test fixture using `buildJsonObject` was indeed masking the bug — V3-shape primitive tests now run alongside it.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Fixed in `9a60118b6`:

```kotlin
private val FILTERED_PROVIDERS: Set<String> =
    setOf("thorchain", "thorchain_streaming", "mayachain", "mayachain_streaming")
```

The filter test now exercises both streaming variants explicitly — `THORCHAIN_STREAMING` and `MAYACHAIN_STREAMING` routes are dropped alongside their non-streaming counterparts. Confirmed against the wire shape you cited from the iOS provider fixture.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Resolved indirectly in `9a60118b6` — the displayed fee no longer flows through gas math at all. The #5 fix (inbound fee from `route.fees[]`) means `SwapQuoteManager.fetchSwapKitQuote` now resolves the fee via `resolveSwapFee(tx.swapFeeTokenContract, tx.swapFee, srcNativeToken, ZERO)` — same pattern as Kyber/Jupiter. The `gasPrice` string is still forwarded verbatim into `OneInchSwapTxJson.gasPrice` for the broadcast path, where the EVM signer parses it at signing time (loud failure, not silent zero on the verify screen).

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Fixed in `9a60118b6` — removed the cache gate entirely (along with the `providerCache` constructor injection). `/v3/quote` is now the authority on unsupported chains, matching iOS at `SwapKitProviderCache.swift:88-91`. Bad-network app launches no longer silently hide SwapKit; the upstream surfaces `noRoutesFound` for chains it doesnt route, which the picker drops cleanly via the existing per-provider `runCatching`.

Left a code comment at `fetchInternal` explaining the divergence intent:
```
// No provider-cache gate here on purpose. iOS SwapKitProviderCache returns `true` on a
// cache-miss and lets `/v3/quote` surface the real error so a single bad-network app
// launch doesnt silently hide SwapKit until the next refresh. Androids cache currently
// returns `false` on cache-miss (fail-closed), which would diverge. Letting `/v3/quote`
// be the authority — it returns a clear `noRoutesFound` for unsupported chains — keeps
// the user opt-in and the swap picker behaviour aligned with iOS.
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Fixed in `9a60118b6`. New `inboundFeeRawUnits(srcToken, route)` mirrors iOS `SwapKitService.inboundFee`:

```kotlin
private fun inboundFeeRawUnits(srcToken: Coin, route: SwapKitRoute): BigInteger {
    val prefix = chainPrefix(srcToken.chain) ?: return BigInteger.ZERO
    val inbound = route.fees.firstOrNull { fee ->
        fee.type?.equals("inbound", ignoreCase = true) == true && fee.chain == prefix
    } ?: return BigInteger.ZERO
    val amount = inbound.amount?.let { runCatching { BigDecimal(it) }.getOrNull() }
        ?: return BigInteger.ZERO
    return amount.movePointRight(srcToken.decimal).toBigInteger()
}
```

The result is stashed on `OneInchSwapTxJson.swapFee` (raw native units) with `swapFeeTokenContract = ""` so `resolveSwapFee` in SwapQuoteManager falls back to `srcNativeToken` — exact same pattern Kyber/Jupiter use. New test pins the SOL case verbatim from your fixture reference: `0.000005 SOL → 5000` raw lamports.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Fixed in `9a60118b6`. Took the "refuse the route" option — falling back to `EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT` would have over-estimated L2 fees by multiples on every chain we route through here. New behaviour:

```kotlin
val gas = evm.gas?.toLongOrNull()?.takeIf { it > 0 }
    ?: throw SwapKitError.Decoding(
        "SwapKit EVM tx missing gas estimate (got ${evm.gas}); refusing route"
    )
```

The picker catches the typed `Decoding` in its per-provider `runCatching` and ranks another aggregator that has a real `tx.gas`. New test `fetch refuses EVM route when tx_gas is missing` pins it.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Fixed in `9a60118b6`. Threaded the sub-provider through the source-result envelope:

```kotlin
// data/.../swap/SwapQuoteSource.kt
data class Evm(val data: EVMSwapQuoteJson, val subProvider: String? = null) : SwapQuoteResult()
```

`SwapKitQuoteSource` populates it from `swapResponse.providers.firstOrNull()` (with `meta.subProvider` fallback for forward compat). `SwapQuoteManager.fetchSwapKitQuote` renders the verify-row label via `formatSubProvider("CHAINFLIP")` → `"SwapKit (Chainflip)"`, and `near_intents` → `"SwapKit (Near Intents)"`.

The proto discriminator stays `"SwapKit"` so `SwapTransactionToUiModelMapper.invoke()` still resolves to `SwapProvider.SWAPKIT` via `getSwapProviderId()` — the dynamic label is purely UI-side. Three new tests pin the contract: `surfaces sub-provider from swap response`, `falls back to meta_subProvider when providers list is empty`, and `leaves subProvider null when neither providers list nor meta subProvider are present`.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Fixed in `9a60118b6`. Wired iOS formula and dropped the dead `bpsDiscount`:

```kotlin
private const val SWAPKIT_AFFILIATE_FEE_BPS = 50
...
val affiliateBps = (SWAPKIT_AFFILIATE_FEE_BPS - (vultBPSDiscount ?: 0)).coerceIn(0, 1000)
...
SwapQuoteRequest(
    srcToken = srcToken,
    dstToken = dstToken,
    tokenValue = tokenValue,
    dstAddress = dst.address.address,
    affiliateBps = affiliateBps,
)
```

The existing `passes affiliateBps from the request through` test stays valid and now exercises a contract a real caller actually drives. The previous `bpsDiscount = vultBPSDiscount ?: 0` had no consumer in `SwapKitQuoteSource` — its removed.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -553,6 +571,91 @@ constructor(
...
+                // for the source chain — the source layer already extracts that and stages it on
+                // `tx.swapFee` (raw units) with `swapFeeTokenContract = ""` (native sentinel). The
+                // resolveSwapFee helper falls back to srcNativeToken on the empty-contract case,
+                // matching how Kyber and Jupiter surface their per-leg fees.
+                val (feeAmount, feeCoin) =
+                    resolveSwapFee(
+                        apiQuote.tx.swapFeeTokenContract,
+                        apiQuote.tx.swapFee,
+                        srcNativeToken,
+                        BigInteger.ZERO,
```
rkokhatskyi: `resolveSwapFee` returns `Pair(fallbackFee, srcNativeToken)` on the `swapFeeTokenContract.isEmpty()` branch at line 667, so passing `BigInteger.ZERO` as the fallback here silently discards the `apiQuote.tx.swapFee` value (the SwapKit-extracted inbound fee, e.g. `5000` lamports for Solana) and `feeAmount` resolves to 0 on every SwapKit quote, putting the user back to the $0 Network Fee state the inbound-fee plumbing was added to fix. Kyber's call at line 435 passes `gasFees` as the fallback for the same reason, so SwapKit needs an analogous live value here — either pass `apiQuote.tx.swapFee.toBigIntegerOrNull() ?: BigInteger.ZERO` as the fallback, send `NATIVE_TOKEN_SENTINEL` as the contract from `SwapKitQuoteSource` so the next `resolveSwapFee` branch parses `swapFeeRaw`, or read `swapFee` directly without going through `resolveSwapFee`. The `SwapKitQuoteSourceTest` inbound-fee case only asserts the intermediate `result.data.tx.swapFee` value at line 297, so the regression is uncaught

rkokhatskyi (changes requested):

aminsato:
@Vaulty-bot  pick up

rkokhatskyi (approved):
LGTM

Issue #3865, opened 2026-03-25
fix: replace all .body<T>() with bodyOrThrow<T>() in API layer


rkokhatskyi:
# [Fix] Replace all .body<T>() with bodyOrThrow<T>() in API layer [Error Handling]

## Problem

The API layer has 74 uses of raw `.body<T>()` vs 50 uses of `bodyOrThrow<T>()`. The project convention (per CLAUDE.md and `docs/network-error-handling.md`) mandates `bodyOrThrow()` for consistent error wrapping. Raw `.body<T>()` can silently return null or throw unhelpful exceptions on HTTP errors.

### Must Do
- [ ] Search all files in `data/src/main/kotlin/com/vultisig/wallet/data/api/` for `.body<`
- [ ] Replace each `.body<T>()` with `.bodyOrThrow<T>()`
- [ ] Verify the `bodyOrThrow` extension function exists and handles error responses properly
- [ ] Ensure no `.body<` remains in the API layer

### Must NOT Do
- Don't change the `bodyOrThrow` implementation itself
- Don't modify non-API files
- Don't change response types or error handling logic beyond the body() → bodyOrThrow() swap

## Acceptance Criteria
- [ ] `grep -r "\.body<" --include="*.kt" data/src/main/kotlin/com/vultisig/wallet/data/api/` returns zero results
- [ ] All API calls use `bodyOrThrow<T>()`
- [ ] `./gradlew assembleDebug` succeeds

**Severity:** MEDIUM

**Repository:** vultisig/vultisig-android

PR #4629, opened 2026-05-24
test: characterize API body-read behavior before #3865 swap

aminsato:
## Summary

Part 1 of 2 for #3865 (replace raw `.body<T>()` with `.bodyOrThrow<T>()` in the API layer).

This PR adds **tests only** — no production code change (other than an additive test-util helper). It characterizes the current behavior of **every** API file that #3865 touches, so the follow-up swap PR can be verified as a pure refactor: the exact same tests must stay green against the changed code.

### Why these tests stay green after the swap
On a **2xx** response, `bodyOrThrow<T>()` is identical to `body<T>()` — it just calls `body<T>()`. So **success-path** tests pass against both the old and the new code. For the three sites that read a **non-2xx (error)** body, the swap preserves behavior via `bodyAsText()` / `json.decodeFromString(bodyAsText())` (never `bodyOrThrow`, which would throw), and those tests pin that.

### Coverage (one test class per file)
- **Error-body parsing (behavior-critical):** `OneInchApiErrorPathTest`, `KyberApiErrorPathTest`, `CardanoApiBroadcastTest` (200 / 400-recovery / 400-error), `LiFiChainApiBodyReadTest`.
- **Success-path swaps:** `DashApi`, `FourByteApi`, `RouterApi`, `RippleApi`, `PolkadotApi`, `BittensorApi`, `SolanaApi`, `CosmosApi`, `TronApi`, `BlockChairApi`, `MayaChainApi`, `SessionApi`.
- `FeatureFlagApi` is already covered by the existing `FeatureFlagApiTest`.

### Test-util change
`MockHttpClient.respondingWith` gains an optional `jsonFormat` param (default = current behavior) so tests whose models contain `@Contextual` fields (e.g. `@Contextual BigInteger`) can supply a `Json` with the matching contextual serializer for `ContentNegotiation`.

### A real bug this surfaced
Writing the LiFi test first caught that a naive `body<String>()` → `bodyOrThrow<String>()` swap at `LiFiChainApi` would be a **behavior change**: LiFi has no `isSuccess` guard and relies on its serializer to surface the error body's `message`; `bodyOrThrow` would throw on non-2xx and surface the generic HTTP status description instead. The follow-up PR uses `bodyAsText()` there. The LiFi test pins this so it can't regress.

## Follow-up
The swap PR (already prepared) keeps all of these tests green — verified locally against the changed code.

## Test plan
- [x] `./gradlew :data:testDebugUnitTest` (new tests) passes against current `main`.
- [x] Verified the same tests pass against the follow-up swap branch — behavior preserved.

## Summary by CodeRabbit

* **Tests**
  * Added many characterization tests validating success and non-2xx response parsing and recovery across multiple blockchain and swap APIs (Cardano, Cosmos, Polkadot, Solana, Tron, Ripple, Dash, BlockChair, Bittensor, FourByte, LiFi, Kyber, OneInch, Maya, Router, Session, etc.), preserving observable decoding behaviors.
* **Chores**
  * Test HTTP client updated to accept a customizable JSON format for contextual deserialization in tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4629?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


aminsato:
@coderabbitai proceed

aminsato:
@coderabbitai full review

data/src/test/kotlin/com/vultisig/wallet/data/api/BittensorApiBodyReadTest.kt
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Test
+
+/**
+ * Characterization tests for [BittensorApiImp.broadcastTransaction] which calls
+ * `response.body<PolkadotBroadcastTransactionJson>()`. They pin the success-path (HTTP 200)
+ * behavior so it is preserved when the call site is migrated to `bodyOrThrow<T>()`.
+ *
+ * Note: [BittensorApi.getTxStatus] is already covered by [BittensorApiTest]; this class only covers
+ * the one uncovered `.body<>()` call site.
```
rkokhatskyi: This note calls `broadcastTransaction` the only uncovered body read in `BittensorApiImp`, yet `getBalance`, `getNonce`, `getBlockHash`, `getBlockHashForNumber`, `getRuntimeVersion`, `getBlockHeader` and `getPartialFee` all deserialize through the shared `postRpc` helper whose `.body()` sits at `utils/Rpc.kt:26`, which no test in this PR exercises so that site goes through the migration unpinned, and since the same helper also backs `PolkadotApiImp.getBalance` and `getPartialFee` a single edit to `Rpc.kt` would shift nine method paths with nothing to catch a regression

data/src/test/kotlin/com/vultisig/wallet/data/api/SolanaApiBodyReadTest.kt
```diff
@@ -0,0 +1,201 @@
...
+import io.ktor.http.HttpStatusCode
+import kotlinx.coroutines.test.runTest
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.modules.SerializersModule
+import kotlinx.serialization.modules.contextual
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+/**
+ * Characterization tests for every `.body<...>()` call in [SolanaApiImp]. Each test uses a 200 OK
```
rkokhatskyi: The class doc claims it pins every `.body<...>()` call in `SolanaApiImp`, whereas `getJupiterTokens` at `SolanaApi.kt:247` deserializes with `.body()` and has no test here or anywhere else, so that read escapes both the `.body<` migration grep and any characterization in this PR

data/src/test/kotlin/com/vultisig/wallet/data/api/MayaChainApiBodyReadTest.kt
```diff
@@ -0,0 +1,217 @@
...
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Assertions.assertNull
+import org.junit.jupiter.api.Test
+
+/**
+ * Characterization tests that pin the SUCCESS-path (HTTP 200) behavior of every method in
+ * [MayaChainApiImp] that uses `.body<T>()` to deserialize the response.
+ *
+ * Skipped methods:
+ * - `getSwapQuotes` — uses `body<String>()` fed into a custom
```
rkokhatskyi: Excluding `getSwapQuotes` leaves the one Maya site most likely to break under a naive swap unpinned, since it guards non success with its own `isSuccess` check at `MayaChainApi.kt:209` then reads `body<String>()` into a custom serializer at line 216, which is the exact shape that made the LiFi case behavior sensitive enough to warrant a pinning test, so the same `body<String>` to `bodyAsText` migration deserves a characterization test here too

data/src/test/kotlin/com/vultisig/wallet/data/api/SessionApiBodyReadTest.kt
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    @Test
+    fun `getTssMessages returns empty list when no messages pending`() = runBlocking {
+        val result = newApi("[]").getTssMessages(serverUrl, sessionId, localPartyId)
+        assertEquals(emptyList<Message>(), result)
+    }
+
+    @Test
+    fun `getTssMessages returns multiple messages in order`() = runBlocking {
```
rkokhatskyi: `getTssMessages` at `SessionApi.kt:153` only calls `body<List<Message>>()` with no sorting, so asserting that `result[0]` then `result[1]` match the mock array order pins nothing beyond kotlinx preserving array order, which the empty and single message cases already cover, while the `in order` name implies an ordering guarantee the production code does not make

rkokhatskyi (changes requested):

rkokhatskyi (approved):
LGTM

Issue #4292, opened 2026-04-26
Bittensor tx status: NoTransformationFoundException on 404 from taostats proxy


aminsato:
## Description

`BittensorApiImp.getTxStatus()` does not handle 404 responses from the taostats proxy (`https://api.vultisig.com/tao-tx/extrinsic/v1`). When the transaction has not yet been indexed, the proxy returns `404` with an empty body, and Ktor throws `NoTransformationFoundException` because it cannot deserialize an empty body into `TaostatsExtrinsicResponse`.

The exception is currently swallowed by `BittensorStatusProvider`, which falls back to `TransactionResult.Pending`. This works, but produces noisy warning logs on every poll while a Bittensor transaction is pending and relies on exception flow for an expected state.

## Repro

1. Submit a Bittensor transaction.
2. Watch tx status polling — every poll until taostats indexes the extrinsic produces this stack trace.

## Logs

```
RESPONSE: 404
FROM: https://api.vultisig.com/tao-tx/extrinsic/v1?hash=0x115ff7de45f89ec6e67babc80e773b3874b9d43f5bb52ec438523c81612bb5d9
content-length: 0

BittensorStatusProvider  W  Bittensor status check failed for 0x115ff7de45f89ec6e67babc80e773b3874b9d43f5bb52ec438523c81612bb5d9
io.ktor.client.call.NoTransformationFoundException: Expected response body of the type 'class com.vultisig.wallet.data.api.TaostatsExtrinsicResponse' but was 'class io.ktor.utils.io.SourceByteReadChannel'
    at io.ktor.client.call.HttpClientCall.bodyNullable(HttpClientCall.kt:91)
    at com.vultisig.wallet.data.api.BittensorApiImp.getTxStatus(BittensorApi.kt:597)
```

## Root cause

`data/src/main/kotlin/com/vultisig/wallet/data/api/BittensorApi.kt:172-179` calls `.body<TaostatsExtrinsicResponse>()` unconditionally without checking the HTTP status:

```kotlin
override suspend fun getTxStatus(txHash: String): TaostatsExtrinsicData? {
    val hash = if (txHash.startsWith("0x")) txHash else "0x$txHash"
    val response =
        httpClient
            .get("${TAOSTATS_PROXY_URL}/extrinsic/v1?hash=$hash")
            .body<TaostatsExtrinsicResponse>()
    return response.data?.firstOrNull()
}
```

## Suggested fix

[...]

PR #4634, opened 2026-05-24
fix(bittensor): correct tao-tx status proxy path so TAO sends confirm (#4565)

### LINKED ISSUE #4565: Tx history: TAO (Bittensor) sends never update from "In progress" to "Successful"
## Summary

TAO (Bittensor) sends remain stuck in the "In progress…" state in the transaction history even after the transaction has confirmed on-chain.

## Probable affected area

- `data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/BittensorStatusProvider.kt` — status check implementation.
- `data/src/main/kotlin/com/vultisig/wallet/data/usecases/txstatus/TxStatusConfiguration.kt:53` — `Chain.Bittensor to TxStatusConfiguration(3, 5 * 60)` (3 retries, 5-minute total budget). Worth checking whether this is enough given Bittensor block/finalization time, or whether the provider's success detection is broken.
- `data/src/main/kotlin/com/vultisig/wallet/data/usecases/txstatus/TransactionStatusProvider.kt:50` — dispatcher routing Bittensor vs Polkadot.

Starting points only — please investigate whether the provider mis-reads the response, whether the retry budget is exhausted before finality, or whether the polling service stops binding after foreground.

## Repro

1. Send a small TAO transaction from Vultisig Android.
2. Wait for on-chain confirmation (verify via taostats / explorer).
3. Open History — entry remains "In progress…" indefinitely, never flips to "Successful".

## Platform

Android.

aminsato:
## Summary

Fixes #4565 — TAO (Bittensor) sends were stuck on **"In progress…"** in transaction history forever and never flipped to **Successful**, even after the tx confirmed on-chain.

### Root cause
The status check called the wrong proxy path:

| | Path | Result on a confirmed hash |
|---|---|---|
| Android (before) | `https://api.vultisig.com/tao-tx/**extrinsic**/v1?hash=…` | **404** (route doesn't exist) |
| iOS | `https://api.vultisig.com/tao-tx/v1?hash=…` | **200** + `{ success, block_number }` |

`/tao-tx/extrinsic/v1` is not a route on the proxy, so it returned **404 for every hash**. `BittensorStatusProvider` maps that 404 → `Pending`, so the history poller kept polling indefinitely and the entry never confirmed. (The earlier #4292 "404 = not yet indexed, keep polling" handling was masking the wrong URL.)

### Fix
One-line path correction in `BittensorApi.getTxStatus`: `/tao-tx/extrinsic/v1` → `/tao-tx/v1`, matching iOS.

### Verified against the live proxy
- Confirmed extrinsics (real, finalized, pulled from the chain) return **`200`** with `success`/`block_number` and the matching `hash` on `/tao-tx/v1`, while `/tao-tx/extrinsic/v1` returns `404`.
- An unknown / not-yet-indexed hash returns **`200`** with an empty `data` array → still `Pending` (polling continues cleanly; the existing 404 guard is kept as defensive).
- A successful extrinsic reports `success: true` → `Confirmed`; a failed one reports `success: false` → `Failed`.

## Test plan
- [x] `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.BittensorApiTest"` — passes (6 tests).
- [x] Added a **regression test** asserting the request URL is exactly `…/tao-tx/v1?hash=…` (no existing test checked the path — which is how the wrong path slipped through).
- [x] Added a test parsing the **real proxy payload shape** (with `pagination` + extra fields) and a `success: false` case.
- [x] `./gradlew :data:ktfmtCheck` — passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the transaction status lookup endpoint to properly query the proxy service.

* **Tests**
  * Added comprehensive unit tests validating transaction status retrieval and response parsing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4634?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/BittensorApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/BittensorApi.kt
@@ -172,7 +172,7 @@ internal class BittensorApiImp @Inject constructor(private val httpClient: HttpC
 
     override suspend fun getTxStatus(txHash: String): TaostatsExtrinsicData? {
         val hash = if (txHash.startsWith("0x")) txHash else "0x$txHash"
-        val response = httpClient.get("${TAOSTATS_PROXY_URL}/extrinsic/v1?hash=$hash")
+        val response = httpClient.get("${TAOSTATS_PROXY_URL}/v1?hash=$hash")
         if (response.status == HttpStatusCode.NotFound) return null
         return response.body<TaostatsExtrinsicResponse>().data?.firstOrNull()
     }
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/BittensorApiTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/BittensorApiTest.kt
@@ -63,6 +63,75 @@ class BittensorApiTest {
         assertNull(api.getTxStatus("0xabc"))
     }
 
+    @Test
+    fun `getTxStatus hits the tao-tx v1 proxy path with the hash query`() = runBlocking {
+        lateinit var requestedUrl: String
+        val api =
+            bittensorApi(
+                MockEngine { request ->
+                    requestedUrl = request.url.toString()
+                    respond(
+                        content = """{"data":[]}""",
+                        status = HttpStatusCode.OK,
+                        headers = jsonHeaders,
+                    )
+                }
+            )
+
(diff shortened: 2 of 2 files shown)


rkokhatskyi (approved):
LGTM

PR #4635, opened 2026-05-24
fix: cache TRON DeFi positions to stop skeleton flashing on every open (#4568)

### LINKED ISSUE #4568: DeFi screen: TRON refetches on every open — cache not honored
## Summary

Opening the DeFi tab on a Tron chain refetches the staked/Bandwidth/Energy/Frozen data every single time. Skeleton loaders flash on each visit, even when reopening seconds later. Expected: render cached values immediately on open, refresh in background.

## Probable affected area

- `data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronDeFiBalanceService.kt:20, 54` — both `getRemoteDeFiBalance` and `getCacheDeFiBalance` exist; check whether the screen-side flow actually calls the cached variant first.
- `data/src/main/kotlin/com/vultisig/wallet/data/repositories/BalanceRepository.kt:225, 379` — `Chain.Tron` branches in `getDeFiCachedTokeBalanceAndPrice` / DeFi service dispatch.
- DeFi load path in `VaultAccountsViewModel.loadDeFiBalances` — verify it emits cached snapshot before invoking remote.

Please investigate whether the cached path is bypassed for Tron specifically (Thor/Maya/Ethereum keep state visible across screen opens on the same vault, only Tron flashes).

## Repro

1. Vault with Tron + some staked TRX.
2. Open DeFi tab → values load.
3. Switch to Wallet tab, back to DeFi → loaders flash again, values refetch.

## Platform

Android.

aminsato:
## Summary

Fixes #4568. The TRON DeFi positions screen (`TronDeFiPositionsScreen`, route `PositionTron`) flashed skeleton loaders and refetched account/resource state on **every** open, even when reopening seconds later. Other DeFi chains (Thor/Maya/Ethereum) stay visible because their list rows emit a cached snapshot first via `AccountsRepository.loadDeFiAddresses`.

### Root cause

`TronDeFiPositionsViewModel.loadData` unconditionally set `Loading` and then hit `TronApi.getAccount`/`getAccountResource` directly — **no cache**. The VM is scoped to the `PositionTron` nav back-stack entry, so it's recreated on each navigation, meaning the in-memory `previousSuccess` was always `null` and the full skeleton showed every time. The data layer's existing DeFi cache path was bypassed entirely for this screen.

### Changes

- **New `TronDeFiSnapshotDataSource`** (data module) — persists the last `TronDeFiSnapshot(account, resource)` as JSON to `AppDataStore`, keyed `tron_defi_account_<address>`. The key is prefixed so it can't collide with `TronResourceDataSource`, which keys on the bare address. Reuses the already-`@Serializable` `TronAccountJson`/`TronAccountResourceJson` (no new DTOs). Bound in `RepositoriesModule`.
- **`TronDeFiPositionsViewModel`** now:
  - renders the cached snapshot immediately as `Success` (skeleton `Loading` only when nothing is cached),
  - fetches fresh in the background, updates state, then persists the new snapshot,
  - keeps the currently displayed `Success` on a background-refresh failure instead of replacing it with an error screen.

Caching the whole account (not just the frozen total) means frozen amount, unfreezing total, **and** pending withdrawals all render instantly on reopen.

## Test plan

- [x] New `TronDeFiPositionsViewModelTest` (5 cases): cached → `Success` before network responds (no skeleton); no cache → `Loading`; snapshot persisted after a successful load; background-refresh failure keeps `Success`; network failure with nothing cached → `Error`.
- [x] `:app` and `:data` compile clean; ktfmt applied.
- [ ] Manual: open a vault with frozen TRX → TRON DeFi screen, switch away and back → values render immediately, no skeleton flash.

## Summary by CodeRabbit

* **Performance Improvements**
  * TRON DeFi positions now display instantly from cached data on app reopens, significantly improving load time.

* **Bug Fixes**
  * Background refresh failures no longer replace successfully loaded data with error states.

* **UI Improvements**
  * Simplified loading state display by removing unnecessary skeleton elements.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4635?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src/main/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModel.kt
```diff
@@ -157,29 +162,53 @@ constructor(
...
+                                    cachedPrice,
+                                    currencyFormat,
+                                ),
+                            isBalanceVisible = isBalanceVisible,
+                        )
+                } else {
+                    val previousSuccess =
+                        (_state.value as? TronDeFiUiState.Success)
+                            ?: (_state.value as? TronDeFiUiState.Loading)?.previousSuccess
+                    _state.value = TronDeFiUiState.Loading(previousSuccess = previousSuccess)
```
rkokhatskyi: Now that every successful load persists a snapshot through `write`, each later `loadData` reads a non null cache and takes the `cached != null` branch, so this `else` is only reachable when nothing was ever cached, and at that point `_state.value` is still the initial `Loading(null)` so `previousSuccess` resolves to null, which means `Loading.previousSuccess` no longer has a live non null producer and the withdrawal skeleton block at `TronDeFiPositionsScreen.kt:236` reading `state.previousSuccess?.tronData?.pendingWithdrawals` is now dead, so either drop `previousSuccess` together with that block or keep a refresh path that emits `Loading(previousSuccess = currentSuccess)` instead of short circuiting to the cached `Success`

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TronDeFiSnapshotDataSource.kt
```diff
@@ -0,0 +1,49 @@
...
+import kotlinx.serialization.json.Json
+import timber.log.Timber
+
+/**
+ * Last-known account + resource state for a TRON address, persisted so the DeFi positions screen
+ * can render immediately on open and refresh in the background instead of flashing skeletons. See
+ * issue #4568.
+ */
+@Serializable
+data class TronDeFiSnapshot(val account: TronAccountJson, val resource: TronAccountResourceJson)
```
rkokhatskyi: This adds a second persistence layer for TRON DeFi state next to `TronDeFiBalanceService` which already persists the frozen total to Room through `StakingDetailsRepository`, so the app now keeps two caches for overlapping data refreshed by independent paths with no shared invalidation, and they can drift since this one stores the full account plus resource snapshot while the other stores only `frozenBandwidth + frozenEnergy`, whereas iOS persists a single computed staked total on `Coin.stakedBalance` and the extension keeps it only in the in memory query cache, so consider widening or reusing the existing store rather than introducing a parallel one

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TronDeFiSnapshotDataSource.kt
```diff
@@ -0,0 +1,49 @@
...
+            null
+        }
+    }
+
+    override suspend fun write(tronAddress: String, snapshot: TronDeFiSnapshot) {
+        appDataStore.set(keyFor(tronAddress), json.encodeToString(snapshot))
+    }
+
+    // Prefixed so it never collides with TronResourceDataSource, which keys on the bare address.
+    private fun keyFor(tronAddress: String) = stringPreferencesKey("tron_defi_account_$tronAddress")
```
rkokhatskyi: The `tron_defi_account_<address>` key is written but never removed, and `VaultRepository.delete` only deletes the vault row without touching DataStore, so deleting a vault or removing the TRON coin leaves this snapshot on disk indefinitely, which mirrors `TronResourceDataSource` so it is not unique to this change yet it does add a new persistent footprint, so wiring snapshot removal into vault or coin deletion would stop stale per address entries from accumulating

app/src/test/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModelTest.kt
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    @Test
+    fun `keeps cached Success when a background refresh fails`() = runTest {
+        coEvery { snapshotDataSource.read(TRX_ADDRESS) } returns CACHED_SNAPSHOT
+        coEvery { tronApi.getAccount(TRX_ADDRESS) } throws RuntimeException("network down")
+
+        val vm = createViewModel().also { it.setData(VAULT_ID) }
+
+        assertTrue(
```
rkokhatskyi, line 132: The core guarantee of this change is that a failed refresh must not corrupt the persisted cache, yet neither this test nor `reports Error when the network fails and nothing is cached` asserts that `write` is never called on the failure path, and since `snapshotDataSource` is a relaxed mock a regression that persisted stale or partial data on failure would pass silently here, so add
```kotlin
coVerify(exactly = 0) { snapshotDataSource.write(any(), any()) }
```
to lock down that guarantee

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModel.kt
aminsato: Good catch — you're right this path was dead. Since every successful load now writes the cache, the `Loading`/`previousSuccess` branch is only reached when nothing was ever cached, where `previousSuccess` is always null. Went with your option A: removed `Loading.previousSuccess` (`Loading` is now a `data object`) and the unreachable withdrawal-skeleton block in the screen. Option B would have reintroduced the skeleton flash this PR exists to remove. Fixed in 05cc93a23.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TronDeFiSnapshotDataSource.kt
aminsato: Agreed, removed the parallel on-disk store. Replaced the DataStore-backed `TronDeFiSnapshotDataSource` with an in-memory `@Singleton TronDeFiSnapshotCache` (`ConcurrentHashMap`), matching the in-memory query cache iOS and the other chains rely on. There's no longer a second *persisted* store that can drift against the frozen total in `StakingDetailsRepository`, and because the cache is `@Singleton` it still survives ViewModel recreation, so the in-session reopen repro stays fixed. Fixed in 05cc93a23.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TronDeFiSnapshotDataSource.kt
aminsato: Resolved by the same change: the snapshot now lives only in memory, so there are no per-address DataStore entries left on disk to accumulate and nothing to wire into vault/coin deletion. Fixed in 05cc93a23.

app/src/test/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModelTest.kt
aminsato: Added — both failure tests (`keeps cached Success when a background refresh fails` and `reports Error when the network fails and nothing is cached`) now assert `verify(exactly = 0) { snapshotCache.write(any(), any()) }`, so a regression that wrote stale/partial data on the failure path would fail. (`read`/`write` are non-suspend now that the cache is in-memory, hence `verify`/`every`.) Fixed in 05cc93a23.

rkokhatskyi (approved):
LGTM

PR #4636, opened 2026-05-24
feat(swap): SwapKit TON source — quote decoder + KeysignPayload wiring

aminsato:
## Summary

First per-chain signer on top of the SwapKit non-EVM foundation. Lands the **TON source** end-to-end:

- `/v3/quote` → `/v3/swap` with `txType=\"TON\"`, where `tx` is a `[{address, amount}]` JSON array.
- `SwapKitQuoteSource` decodes the transfers, re-encodes the canonical JSON into `SwapKitSwapPayloadJson.txPayload`, and returns `SwapQuoteResult.SwapKit` carrying a fully-formed `SwapQuote.SwapKit`.
- `SwapQuoteManager.fetchSwapKitQuote` dispatches on `SwapQuoteResult.Evm` vs `.SwapKit`; the EVM branch is extracted to a `buildSwapKitOneInchQuote` helper but behaves identically.
- `SwapFormViewModel` builds `RegularSwapTransaction` with `payload = SwapPayload.SwapKit` and `dstAddress = targetAddress`.
- `KeysignShareViewModel` decodes the canonical transfer bytes into `SignTon(tonMessages = [...])` so the existing `TonHelper.addTransfersTo` TonConnect-style path picks them up unchanged.
- `TonHelper` falls back to decoding the SwapKit payload bytes when `signTon` is null — covers iOS-initiator → Android-peer cosigning where iOS ships only the SwapKit wire payload (no `signTon`).
- `SwapProviderTable.Chain.Ton` → `SWAPKIT` (was `emptySet` — TON is SwapKit-only on Android today).

### Cross-platform parity
- Wire shape matches iOS verbatim: same `txType=\"TON\"`, same canonical JSON bytes on the payload (kotlinx.serialization default encoding is compact, key order matches the alphabetical struct field order).
- Sub-provider rendering uses `formatSwapKitProviderLabel` introduced in the foundation PR — `SwapKit (CHAINFLIP)` byte-for-byte identical to iOS.

### Test plan
- [x] `./gradlew :data:test :app:testDebugUnitTest` — green
- [x] `./gradlew ktfmtCheck` — green
- [x] `./gradlew :app:lintDebug` — green
- [x] 5 new TON-specific cases in `SwapKitQuoteSourceTest` (24 total): TON txType returns `SwapQuoteResult.SwapKit` with canonical bytes; multi-transfer arrays preserved; empty array surfaces as `Decoding`; non-integer amount surfaces as `Decoding`; sub-provider passthrough.

### Live-test pending
[...]

PR #4637, opened 2026-05-24
feat(swap): SwapKit Phase 1 — localized error messages + user-facing wiring

aminsato:
## Summary

Closes the last gap iOS PR vultisig-ios#4394 lists for **Phase 1 (EVM + Solana)**: *"8-locale strings for every user-facing error code"*. Today every `SwapKitError` falls through `SwapFormViewModel`'s generic `Exception` catch and shows the generic *"Swap quote failed"* copy — users see the same message for an OFAC screening rejection, an expired quote, and an insufficient-balance route.

Per @rkokhatskyi'\''s review on #4636, focusing on Phase 1 finalisation before any non-EVM signer work.

### Tranaction hash
https://etherscan.io/tx/0x29a5d4611460ad703216258a65e77430d7d36d71ceb5118208d59ed76247f7f1

### Changes

**Typed errors** (`SwapKitError`)
- Extend with the full set iOS recognises: `ApiKeyMissing` / `ApiKeyInvalid` / `InsufficientBalance` / `InsufficientAllowance` / `UnableToBuildTransaction` (aliases `failedToRetrieveBalance`) / `SwapRouteNotFound` / `BlackListAsset` / `InvalidSourceAddress` / `InvalidDestinationAddress` / `AddressScreening` (aliases `isSanctionedAddress` + `addressScreeningFailed`) / `ProviderNotEnabled` / `RouteFiltered` / `MalformedAmount(raw)`.
- Existing variants kept as-is: `NoRoutes`, `QuoteDeviation`, `UnsupportedTxType`, `Network`, `Server`, `Decoding`.
- `SwapKitError.fromCode` maps every documented wire code to the typed variant. Case-insensitive; unknown codes fall through to `Server` (with HTTP status) or `Network` (without one).

**Localization** (10 Android locale files)
- en (base) — written from scratch.
- de / es / hr / it / ko / pt / zh-rCN — mirrored verbatim from iOS Phase 1 translations.
- nl / ru — translated using the per-locale terminology already established by the `settings_advanced_swapkit_toggle` string PR vultisig-android#4619 added.
- `%@` → `%1$s` / `%1$d` per Android resource format. `%` literals escaped as `%%`.

**Wiring**
- `SwapQuoteManager.mapSwapKitErrorToFormError` translates every `SwapKitError` variant to its matching string resource.
- `SwapFormViewModel` grows a `catch (e: SwapKitError)` clause between the existing `SwapException` and generic `Exception` branches so the typed errors surface with their localized copy instead of the generic "quote failed" fallback.

### Test plan
- [x] `./gradlew :data:test :app:testDebugUnitTest` — green
- [x] `./gradlew ktfmtCheck` — green
- [x] `./gradlew :app:lintDebug` — green
- [x] 6 new `SwapKitErrorFromCodeTest` cases pin every wire code → typed variant; alias for `failedToRetrieveBalance`; case-insensitive lookup; whitespace tolerance; unknown-code Server vs Network fallback.

### Out of scope
- **Mantle / zkSync / Blast chain coverage.** iOS' `chainPrefix` doesn'\''t actually map these (falls through to `coin.ticker` — produces `ETH.ETH` for zkSync.ETH, colliding with Ethereum). Without live verification of the SwapKit prefix for each chain I won'\''t add speculative entries. The 8 chains we already cover (ETH / BSC / AVAX / ARB / BASE / OP / POL / Solana) match what iOS Phase 1 actually exercises.
- **Forced-provider debug picker.** iOS ships this in Advanced Settings for isolation testing — flagging as a follow-up tooling PR rather than user-facing Phase 1.

## Which feature is affected?
- [x] Swap — finishes Phase 1 EVM + Solana by making SwapKit errors user-readable.

## Checklist
- [x] Self-review
- [x] No new warnings
- [x] Tests included

## Agent Metadata
- **Agent**: Claude Code (Opus 4.7, 1M context)
- **Issue**: Phase 1 completion of vultisig-android#4589 per @rkokhatskyi'\''s review on #4636
- **Confidence**: high for the code path; the nl / ru translations would benefit from a native-speaker review pass before locale freeze
- **Needs human review for**: the nl / ru translations (no iOS reference available), and confirmation that the `swapkit_error_*` keys + wording match the project'\''s preferred tone

## Bug Fixes

* Improved error handling and messaging for swap transactions
* Added comprehensive, localized error messages across 10+ languages (German, Spanish, Croatian, Italian, Korean, Dutch, Portuguese, Russian, and Chinese)
* Enhanced error reporting for swap failures, including: insufficient balance, invalid addresses, network errors, rate deviations, and provider unavailability

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4637?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


aminsato:
@coderabbitai full review

aminsato:
@coderabbitai proceed

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -1133,6 +1134,12 @@ constructor(
                             cause = e,
                             tag = "swapError",
                         )
+                    } catch (e: SwapKitError) {
```
rkokhatskyi, line 1143: This catch can rarely fire for a real multi provider quote since `fetchBestQuote` resolves a total failure with `throw failures.firstOrNull { it is SwapException } ?: failures.first()`, and `SwapProviderTable` always offers `SWAPKIT` next to `LIFI` which fails with a `SwapException`, so a sibling `SwapException` masks the typed `SwapKitError` and the user keeps seeing the generic copy this PR set out to replace, whereas iOS `SwapService.swift:77` throws the first failure without preferring a type so its SwapKit message can surface

data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapKitError.kt
```diff
@@ -50,12 +94,28 @@ sealed class SwapKitError(message: String, cause: Throwable? = null) : Exception
...
                 "no_routes_found",
                 "no_routes" -> NoRoutes(fallbackMessage ?: raw)
                 "outputamountdeviationtoohigh",
                 "output_amount_deviation_too_high" -> QuoteDeviation(fallbackMessage ?: raw)
+                "blacklistasset" -> BlackListAsset()
+                "invalidsourceaddress" -> InvalidSourceAddress()
+                "invaliddestinationaddress" -> InvalidDestinationAddress()
+                "issanctionedaddress",
+                "addressscreeningfailed" -> AddressScreening()
+                "providernotenabled" -> ProviderNotEnabled()
```
rkokhatskyi: `ProviderNotEnabled` and `RouteFiltered` are produced only here from wire codes, yet on iOS both are thrown from client logic at `SwapService.swift:371` and `:386` and are absent from the wire `init?(envelope:)` switch, and Android never ported those gates since `SwapKitProviderCache` is bound in `RepositoriesModule` but injected nowhere while the empty route path throws `NoRoutes`, so unless SwapKit truly emits `providerNotEnabled` or `routeFiltered` on the wire these two arms and their variants are dead and users get the generic `NoRoutes` copy where iOS shows the specific message

data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapKitError.kt
```diff
@@ -50,12 +94,28 @@ sealed class SwapKitError(message: String, cause: Throwable? = null) : Exception
         ): SwapKitError {
             val raw = code?.trim().orEmpty()
             val message = fallbackMessage ?: raw.ifEmpty { "SwapKit request failed" }
-            return when (raw.lowercase()) {
+            // Locale.ROOT to dodge Turkish-locale's `I` → `ı` (dotless i) mapping; the SwapKit
+            // wire codes are ASCII camelCase and must round-trip on any device locale.
+            return when (raw.lowercase(Locale.ROOT)) {
+                "apikeymissing" -> ApiKeyMissing()
+                "apikeyinvalid" -> ApiKeyInvalid()
+                "insufficientbalance" -> InsufficientBalance()
```
rkokhatskyi: Only `NoRoutes` and `QuoteDeviation` also accept snake_case spellings like `no_routes_found`, whereas the thirteen codes this PR adds match camelCase only, so if SwapKit returns `insufficient_balance` or `swap_route_not_found` in that same snake_case style they fall through to `Server` or `Network` and the localized copy never shows, which is worth pinning against a real wire sample since the repo fixtures already disagree with `NoRoutesFound` in `SwapKitQuoteDecodingTest` versus `noRoutesFound` here

data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapKitError.kt
```diff
@@ -50,12 +94,28 @@ sealed class SwapKitError(message: String, cause: Throwable? = null) : Exception
         ): SwapKitError {
             val raw = code?.trim().orEmpty()
             val message = fallbackMessage ?: raw.ifEmpty { "SwapKit request failed" }
-            return when (raw.lowercase()) {
+            // Locale.ROOT to dodge Turkish-locale's `I` → `ı` (dotless i) mapping; the SwapKit
+            // wire codes are ASCII camelCase and must round-trip on any device locale.
+            return when (raw.lowercase(Locale.ROOT)) {
```
rkokhatskyi: The typed mapping hinges on `extractErrorCode` in `SwapKitApi.kt:148` which searches the whole body for the first `code` key before falling back to top level `error`, so a nested `code` like the `warnings[].code` that SwapKit payloads carry would resolve the wrong variant, whereas iOS reads the top level `error` field directly, and no test feeds a realistic error body through `fromCode` to catch this

data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapKitError.kt
```diff
@@ -1,47 +1,91 @@
...
- * paired with a localized message key consumed by the swap UI.
+ * failure mode documented at https://docs.swapkit.dev (or a client-side filter outcome). The swap
+ * UI converts each variant to a localized `swapkit_error_*` string.
  */
 sealed class SwapKitError(message: String, cause: Throwable? = null) : Exception(message, cause) {
 
-    /** No SwapKit route survived the THORChain/Maya and single-hop filters. */
-    class NoRoutes(message: String = "No SwapKit route available") : SwapKitError(message)
+    /** SwapKit proxy reports its partner API key wasn't supplied. */
+    class ApiKeyMissing : SwapKitError("SwapKit API key not configured")
```
rkokhatskyi: These stateless variants are declared `class` so every `fromCode` hit allocates a new instance and equality is identity based, which makes `ApiKeyMissing() == ApiKeyMissing()` false and is only hidden because the tests assert with `assertInstanceOf` rather than value equality, so `data object` would fit the eleven stateless cases better
```kotlin
data object ApiKeyMissing : SwapKitError("SwapKit API key not configured")
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -766,6 +767,60 @@ constructor(
                 UiText.StringResource(R.string.swap_error_amount_below_dust_threshold)
         }
 
+    /** Localized message for each [SwapKitError] variant — surfaced verbatim on the swap form. */
+    fun mapSwapKitErrorToFormError(e: SwapKitError): UiText =
```
rkokhatskyi, line 775: `mapSwapKitErrorToFormError` is the user visible payload of this PR yet has no test, and although the exhaustive `when` turns a missing branch into a compile error, the nineteen near identical `R.string` assignments, the `Server(httpStatus = null)` fallback to `swapkit_error_network`, and the three `FormattedText` argument branches are only verifiable by a unit test asserting each variant maps to the expected `UiText`

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -326,15 +326,13 @@ constructor(
...
+     * [SwapKitError.MalformedAmount] when the upstream value is missing or unparseable so the user
+     * sees the typed "Invalid amount" message instead of a generic "Could not parse response".
      */
     private fun scaleDecimalToRawUnits(decimalString: String?, decimals: Int): String {
         val parsed =
             decimalString?.let { runCatching { BigDecimal(it) }.getOrNull() }
-                ?: throw SwapKitError.Decoding(
-                    "SwapKit expectedBuyAmount missing or not a decimal: $decimalString"
-                )
+                ?: throw SwapKitError.MalformedAmount(raw = decimalString ?: "(missing)")
```
rkokhatskyi: When `decimalString` is null this reaches the user as `Invalid amount returned by SwapKit: (missing)` through `swapkit_error_malformed_amount`, so the developer sentinel `(missing)` leaks into the form error, and dropping the raw value for the null case while keeping it only in the exception message would read better

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -766,6 +767,60 @@ constructor(
...
+                UiText.StringResource(R.string.swapkit_error_api_key_invalid)
+            is SwapKitError.InsufficientBalance ->
+                UiText.StringResource(R.string.swapkit_error_insufficient_balance)
+            is SwapKitError.InsufficientAllowance ->
+                UiText.StringResource(R.string.swapkit_error_insufficient_allowance)
+            is SwapKitError.UnableToBuildTransaction ->
+                UiText.StringResource(R.string.swapkit_error_unable_to_build_transaction)
+            is SwapKitError.SwapRouteNotFound ->
+                UiText.StringResource(R.string.swapkit_error_swap_route_not_found)
+            is SwapKitError.QuoteDeviation ->
```
rkokhatskyi, line 789: Passing `emptyList()` to `FormattedText` purely so `String.format` collapses the resource `5%%` into `5%` is a subtle trap since a later switch to `StringResource` would silently render `5%%`, so a dedicated `UiText` documenting the unescape intent would be clearer than an empty args list

data/src/test/kotlin/com/vultisig/wallet/data/api/errors/SwapKitErrorFromCodeTest.kt
```diff
@@ -0,0 +1,136 @@
...
+        // NEAR Intents collapses upstream UTXO-indexer failures into `failedToRetrieveBalance`.
+        // The user-facing meaning is the same as `unableToBuildTransaction`: try another provider.
+        assertInstanceOf(
+            SwapKitError.UnableToBuildTransaction::class.java,
+            SwapKitError.fromCode("failedToRetrieveBalance"),
+        )
+    }
+
+    @Test
+    fun `fromCode is case-insensitive on the wire code`() {
```
rkokhatskyi, line 91: This test claims locale independence but runs under the JVM default locale so it would still pass if `Locale.ROOT` were removed from `fromCode`, whereas the regression it guards only appears under a Turkish locale where `"ISSANCTIONEDADDRESS".lowercase()` yields a dotless `ı`, so wrapping it in `Locale.setDefault(Locale("tr"))` would actually pin the fix

app/src/main/res/values-ru/strings.xml
```diff
@@ -576,6 +576,25 @@
     <string name="faq_settings_a9">Он построен на технологии MPC, которая устраняет необходимость в сид-фразах и поддерживает несколько блокчейнов, что делает Vultisig гибким и независимым от сети.</string>
     <string name="vault_settings_advanced_title">Дополнительно</string>
     <string name="settings_advanced_swapkit_toggle">Агрегатор SwapKit</string>
+    <string name="swapkit_error_api_key_missing">API-ключ SwapKit не настроен</string>
+    <string name="swapkit_error_api_key_invalid">API-ключ SwapKit недействителен</string>
+    <string name="swapkit_error_insufficient_balance">Недостаточно средств для этого маршрута</string>
+    <string name="swapkit_error_insufficient_allowance">Недостаточный allowance токена — сначала одобрите</string>
```
rkokhatskyi: `swapkit_error_insufficient_allowance` embeds the Latin word `allowance` inside Cyrillic text, whereas this locale already renders the ERC20 allowance as `лимит` in `verify_swap_agree_allowance` and every other locale translated the term, so using `лимит` here would match what users already see

app/src/main/res/values-nl/strings.xml
```diff
@@ -572,6 +572,25 @@
     <string name="faq_settings_a9">Het is gebouwd op MPC-technologie, waardoor herstelzinnen (seed phrases) niet nodig zijn en meerdere blockchains worden ondersteund, wat Vultisig flexibel en ketenonafhankelijk maakt.</string>
     <string name="vault_settings_advanced_title">Geavanceerd</string>
     <string name="settings_advanced_swapkit_toggle">SwapKit-aggregator</string>
+    <string name="swapkit_error_api_key_missing">SwapKit API-sleutel niet geconfigureerd</string>
+    <string name="swapkit_error_api_key_invalid">SwapKit API-sleutel is ongeldig</string>
+    <string name="swapkit_error_insufficient_balance">Onvoldoende saldo voor deze route</string>
+    <string name="swapkit_error_insufficient_allowance">Onvoldoende token-allowance — keur eerst goed</string>
```
rkokhatskyi: `swapkit_error_insufficient_allowance` keeps the English `allowance` while this locale uses `toelage` for the ERC20 allowance in `verify_swap_agree_allowance` and the seven other translated locales each picked a native term, so `toelage` would keep the wording consistent

rkokhatskyi (changes requested):

aminsato:
@Vaulty-bot pick up

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -306,7 +307,11 @@ constructor(
         val successes = results.mapNotNull { it.getOrNull() }
         if (successes.isEmpty()) {
             val failures = results.mapNotNull { it.exceptionOrNull() }
-            throw failures.firstOrNull { it is SwapException } ?: failures.first()
+            // Throw the first failure (iOS SwapService.swift:77 parity). Previously preferred any
+            // SwapException over the first failure, which masked sibling SwapKitError variants
+            // whenever a LiFi/Thor candidate raced SwapKit and lost — the user kept seeing the
+            // generic swap_error_quote_failed copy instead of the localized SwapKit message.
+            throw failures.first()
```
rkokhatskyi, line 314: `failures.first()` is the first candidate's failure in declared order, and `SWAPKIT` sits last in every `SwapProviderTable` set behind `LIFI` which is always also eligible and throws `SwapException` on failure, so on a total failure `failures.first()` is never the `SwapKitError` and the localized SwapKit copy still never reaches the user, which means changing from the `SwapException` preference to the first failure did not change the outcome for the real provider ordering, and only preferring the SwapKit error would deliver it
```kotlin
throw failures.firstOrNull { it is SwapKitError } ?: failures.first()
```

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/SecurityScannerTransactionFactory.kt
```diff
@@ -82,7 +88,7 @@ class SecurityScannerTransactionFactory(
                 from = from,
                 to = srcToken.contractAddress,
                 amount = BigInteger.ZERO,
-                data = EthereumFunction.approvalErc20Encoder(to, amount.toBigInteger()),
+                data = EthereumFunction.approvalErc20Encoder(approveSpender, amount.toBigInteger()),
```
rkokhatskyi: This encodes the approval with `amount` set to `tx.value`, which is `0` for an ERC20 token swap, so the scanner previews `approve(spender, 0)` while the user actually signs `approve(spender, srcTokenValue)` in `KeysignShareViewModel`, so the scanned approval amount no longer matches the signed one and should use the real approve amount now that the spender is being corrected here

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -242,10 +253,21 @@ constructor(
...
+                            // spender
+                            // is derived from it downstream instead of from `to`. Matches iOS,
+                            // which
+                            // derives the spender as `meta.approvalAddress`.
+                            allowanceTarget = response.meta.approvalAddress,
                             data = evm.data,
                             gas = gas,
-                            value = evm.value,
-                            gasPrice = evm.gasPrice.orEmpty(),
+                            value = (parseEvmNumber(evm.value) ?: BigInteger.ZERO).toString(),
```
rkokhatskyi, line 269: A malformed `value` silently defaults to zero here whereas the missing `gas` a few lines up refuses the route with `Decoding`, and iOS throws `malformedAmount` on an unparseable value rather than zeroing it, so for a native source SwapKit swap an unparseable `value` would build a tx that sends `0` and quietly under delivers instead of failing fast

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -479,6 +480,11 @@ constructor(
 
                             is SwapQuote.OneInch -> {
                                 val dstAddress = quote.data.tx.to
+                                // The ERC20 allowance must be granted to the provider's token-
+                                // transfer proxy, which for SwapKit differs from the swap `to`.
+                                // 1inch/Kyber/LiFi leave allowanceTarget null and fall back to
+                                // `to`.
+                                val approveSpender = quote.data.tx.allowanceTarget ?: dstAddress
```
rkokhatskyi, line 487: The quote source test pins `meta.approvalAddress` into `tx.allowanceTarget`, but the security relevant hop here where `approveSpender` is derived and then carried into the signed `ERC20ApprovePayload.spender` in `KeysignShareViewModel` has no test, so a regression that reverts the approve target back to `tx.to` would pass CI unnoticed

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/SecurityScannerTransactionFactory.kt
aminsato: Good catch — fixed in 7e94346. `buildSwapSecurityScannerTransaction` now takes the real approve amount (`transaction.srcTokenValue.value`) for the approval encoding, so the scanned `approve(spender, amount)` matches what `KeysignShareViewModel` actually signs. `tx.value` is still used for the non-approval swap branch.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Agreed — added `KeysignShareViewModelSwapApprovalTest` in 7e94346 covering the full hop. It builds a SwapKit `RegularSwapTransaction` where `allowanceTarget != tx.to` and asserts the resulting `ERC20ApprovePayload.spender` is the token-transfer proxy (not `dstAddress`), plus that the approve amount equals the swap input. A regression reverting the spender back to `tx.to` now fails CI.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: You're right — with `SWAPKIT` ranked last and `LIFI` always eligible and failing first with a `SwapException`, `failures.first()` never resolves to the `SwapKitError`, so the localized copy still does not reach the user. `throw failures.firstOrNull { it is SwapKitError } ?: failures.first()` is the correct fix; will apply it in a follow-up.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Agreed — silently zeroing an unparseable `value` is inconsistent with the `gas` handling right above (which refuses the route) and with iOS throwing `malformedAmount`. For a native-source route it would build a tx that sends 0 and under-delivers instead of failing fast. Will change it to throw `SwapKitError.MalformedAmount` rather than defaulting to zero, in a follow-up.

PR #4640, opened 2026-05-25
fix(keygen): seed-phrase import from extension/iOS now completes on Android joiner

### LINKED ISSUE #4639: [BUG] Seed-phrase import (KeyImport) from extension/iOS fails on Android joiner — per-chain keygen hangs
Bug report

## Summary
When a vault is created via **import seed phrase** (KeyImport) with the **browser extension** (or iOS) as the initiator and an **Android** device as the joiner, the Android device never completes keygen and the vault is never created. The same QR joined by an **iOS** device succeeds, so this is Android-specific.

## Steps to reproduce
1. On the **extension**: import a seed phrase, select some chains, choose **2 devices**, set a name → a QR is generated.
2. Join with an **Android** device (scan the QR).
3. Observe: Android starts keygen but hangs and never finishes.
4. Repeat step 2 with an **iOS** device → succeeds.

## Expected
Android joins keygen and the imported vault is created (parity with iOS).

## Actual
Android hangs during keygen. The root ECDSA/EdDSA key ceremonies complete, but the **per-chain** key ceremony never does. logcat shows Android polling the relay forever and getting an empty array back:

```
REQUEST: .../router/message/<session>/<device>
METHOD: GET
-> message_id: Ethereum
RESPONSE: 200 ... BODY: []      // repeats forever
```

## Root cause
Each TSS sub-ceremony of a seed-phrase import (root ECDSA, root EdDSA, and one per selected chain) exchanges messages over the relay on a named channel (`message_id`); both devices must agree on the name.

For the **per-chain** ceremonies in the sequential (non-batched) path, iOS and the extension exchange on the **default/shared** channel (the exchange id is left unset), while Android exchanges on a channel **named after the chain** (e.g. `Ethereum`). The mismatch comes from the `KeygenRouting.from(setupMessageId = chainName)` helper having different defaults per platform:
- iOS: exchange id defaults to **nil** → default channel.
- Android: exchange id defaults to the **setup id** (chain name) → per-chain channel.

So Android polls a channel the initiator never posts to and the ceremony hangs after the root keygens. The root keygens succeed because both sides use the default channel there. Android-to-Android worked only because both sides made the same (incorrect) choice.

A related gap: Android did not read the `is_tss_batch` field from the `KeygenMessage` proto, and the batched key-import relay namespaces (`p-ecdsa` / `p-eddsa` / `p-{chain}`) did not match iOS/Windows either — so an initiator with the remote `tss-batching` flag enabled would also fail to keygen with an Android joiner.

## Affected
- Android joiner of any extension/iOS-initiated seed-phrase import (the common case).
- Confirmed against `vultisig-ios` (`KeygenViewModel.swift`) and `vultisig-windows` (`KeyImportKeygenActionProvider.tsx` / `JoinKeyImportKeygenActionProvider.tsx`), whose routing is the authoritative cross-platform contract.

aminsato:
Fixes #4639

## Problem
Creating a vault via **import seed phrase** (KeyImport) with the extension/iOS as initiator and **Android** as joiner hung during keygen and never created the vault. The same QR joined by **iOS** succeeded, so it was Android-specific.

logcat from the failing Android device showed the root ECDSA/EdDSA ceremonies completing, then an endless empty poll on the per-chain relay channel:

```
GET .../router/message/<session>/<device>   -> message_id: Ethereum   -> 200 BODY: []   (repeats)
```

## Root cause
Each TSS sub-ceremony of a seed-phrase import (root ECDSA, root EdDSA, one per chain) exchanges messages over a named relay channel; both devices must agree on the name.

For the **per-chain** ceremonies on the sequential (non-batched) path, iOS and the extension leave the exchange id unset and exchange on the **default/shared** channel. Android exchanged on a channel **named after the chain**. The divergence is in the `KeygenRouting.from(setupMessageId = chainName)` helper:

| platform | default exchange id | per-chain exchange channel |
|---|---|---|
| iOS | `nil` | default/shared |
| Android (before) | the **setup id** (chain name) | `Ethereum`, `Bitcoin`, … |

So Android polled a channel the initiator never posts to and stalled after the root keygens (those share the default channel on both sides, so they completed). Android-to-Android worked only because both peers made the same wrong choice.

A second, related gap: Android never read `is_tss_batch` from the `KeygenMessage` proto, and its batched key-import relay namespaces did not match iOS/Windows — so an initiator with the remote `tss-batching` flag enabled would also fail to keygen with an Android joiner.

## Fix
- **Sequential per-chain (the reported failure):** pass an explicit empty exchange id so the ceremony uses the default channel, matching iOS/extension.
- **Batched path (forward-compat for `tss-batching` on):**
  - read `is_tss_batch` from `KeygenMessage` and have the joiner trust the QR for keygen/key-import (mirrors the existing reshare behaviour);
  - route batched key-import on `p-ecdsa` / `p-eddsa` / `p-{chain}` with the `eddsa_key_import` setup namespace.

Routing verified against the authoritative cross-platform sources: `vultisig-ios` (`KeygenViewModel.swift`) and `vultisig-windows` (`KeyImportKeygenActionProvider.tsx` / `JoinKeyImportKeygenActionProvider.tsx`).

## Tests
- `KeyImportRoutingServerMatchTest` — locks the relay-channel contract for root + per-chain, sequential and batched.
- `KeygenMessageProtoMapperTest` — `is_tss_batch` round-trips through the `KeygenMessage` mappers.
- Existing keygen/routing/reshare unit tests pass; ktfmt clean.

## Test plan
- [x] Unit tests + ktfmt
- [x] Manual: extension (import seed phrase, multiple chains, 2 devices) → **Android** joiner completes and the vault is created (previously hung). Verified on device.
- [ ] Regression: Android↔Android seed-phrase import still completes
- [ ] iOS joiner still completes (unchanged path)

## Summary by CodeRabbit

* **New Features**
  * Joiners’ QR opt-in for batched (parallel) keygen/key-import is now honored only when the ceremony type is eligible; sessions combine QR opt-in and per-ceremony eligibility.
  * Relay namespaces clarified and isolated: per-chain imports use distinct namespaces for parallel vs sequential flows; root ECDSA/EdDSA namespaces are separated to avoid collisions.
  * Keygen messages now carry a persisted batched-flag for round-trip serialization.

* **Tests**
  * Added tests covering opt-in decision matrix, mapper round-trips for the batched flag, and relay-namespace routing for root and per-chain imports.
  * Removed one obsolete wire-string pin test.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4640?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src/main/java/com/vultisig/wallet/ui/models/keygen/KeygenViewModel.kt
```diff
@@ -194,24 +193,27 @@ constructor(
...
+     * batch flag, so the local feature flag stays their only decision input.
      */
     private fun isParallelKeygenEnabledForThisCeremony(): Boolean =
-        if (action == TssAction.ReShare) {
-            args.isTssBatch
-        } else {
-            featureFlags.isTssBatchEnabled || args.isTssBatch
+        when (action) {
+            TssAction.ReShare,
+            TssAction.KEYGEN,
```
rkokhatskyi: Routing KEYGEN through `args.isTssBatch` exclusively makes the batched keygen path in the app unreachable from an Android initiator, since `KeygenPeerDiscoveryViewModel` sets `isTssBatch` to `isBatchEligibleReshare(...) && isTssBatchEnabled` which is false for KEYGEN and `createKeygenPayload` never sets the flag on the QR, whereas the FastVault server join still calls `joinBatchKeygen` whenever `isTssBatchEnabled`, so once the batch flag is enabled the server exchanges on `p-ecdsa`/`p-eddsa` while the app stays on the default channel and the ceremony hangs. Before this change KEYGEN was `featureFlags.isTssBatchEnabled || args.isTssBatch` which kept the initiator in step with `joinBatchKeygen`, so either keep the OR fallback for KEYGEN or gate `joinBatchKeygen` on the same `args.isTssBatch`

app/src/main/java/com/vultisig/wallet/ui/models/keygen/KeygenViewModel.kt
```diff
@@ -194,24 +193,27 @@ constructor(
...
-     * For keygen / migrate / key-import we keep the OR fallback because Android does not yet carry
-     * `is_tss_batch` on the `KeygenMessage` proto on the read side, so a joiner has no QR signal to
-     * trust — its local feature flag is the only available decision input. (Reading the
-     * `is_tss_batch` field on `KeygenMessage` is a separate follow-up that will let us collapse to
-     * a single QR-driven rule.)
+     * Migrate and SingleKeygen keep the OR fallback: migrate is excluded from batch mode by
+     * convention (its joiner branch pins the flag off) and SingleKeygen's proto does not carry the
+     * batch flag, so the local feature flag stays their only decision input.
      */
     private fun isParallelKeygenEnabledForThisCeremony(): Boolean =
```
rkokhatskyi, line 201: Nothing exercises this decision even though the fix turns on it, and since no test sets `args.isTssBatch` across actions a regression that put KEYGEN back on `featureFlags.isTssBatchEnabled || args.isTssBatch` would still pass, so consider extracting it into a pure function next to `shouldUseNewKeygenExecution` and unit testing the matrix for each action

app/src/main/java/com/vultisig/wallet/ui/models/keygen/KeygenViewModel.kt
```diff
@@ -593,20 +611,28 @@ constructor(
...
+                                            // exchange on "p-ecdsa". This must match the batched
+                                            // key-import routing on iOS/Windows (the initiator),
+                                            // otherwise the joiner polls a different relay channel
+                                            // and the ceremony deadlocks.
                                             dklsKeygen.dklsKeygenWithRetry(
                                                 0,
                                                 KeygenRouting.from(
-                                                    setupMessageId =
-                                                        ROOT_ECDSA_KEY_IMPORT_MESSAGE_ID
+                                                    exchangeMessageId = ROOT_ECDSA_MESSAGE_ID
```
rkokhatskyi, line 598: This block dropped the last production use of `ROOT_ECDSA_KEY_IMPORT_MESSAGE_ID`, so that constant in `KeygenExecution.kt` is now referenced only by `ReshareRoutingServerMatchTest` asserting its literal value while `ROOT_EDDSA_KEY_IMPORT_MESSAGE_ID` stays live at line `634`, so the ECDSA constant and its assertion can be removed

app/src/test/java/com/vultisig/wallet/data/keygen/KeyImportRoutingServerMatchTest.kt
```diff
@@ -0,0 +1,105 @@
...
+                setupMessageId = ROOT_EDDSA_KEY_IMPORT_MESSAGE_ID,
+                exchangeMessageId = ROOT_EDDSA_MESSAGE_ID,
+            )
+
+        assertEquals("eddsa_key_import", routing.setupMessageId)
+    }
+
+    @Test
+    fun `batched per-chain key-import exchanges on p-{chain} and sets up on the chain id`() {
+        val chain = "bitcoin"
```
rkokhatskyi: This case uses `"bitcoin"` lowercased and asserts `p-bitcoin`, whereas the ceremony builds the namespace from `chainSetting.chain.raw` which is `"Bitcoin"` for `Chain.Bitcoin`, so it locks a channel production never emits while the sibling case correctly uses `"Ethereum"`. These cases also only exercise `KeygenRouting.from` with literals rather than the routing built inside `runKeyImportChainKeygen`, so dropping the explicit `exchangeMessageId = ""` from the view model would still leave them green, whereas extracting that routing into a pure function keyed off `chain.raw` would actually guard the regression

app/src/main/java/com/vultisig/wallet/ui/models/keygen/JoinKeygenViewModel.kt
```diff
@@ -175,6 +175,15 @@ constructor(
...
                                 chains = message.chains,
+                                // Follow the initiator's batched opt-in from the QR. iOS/Windows
+                                // run
+                                // key-import (and DKLS keygen) through the batched protocol, which
+                                // uses dedicated relay namespaces; a joiner that ignores this flag
+                                // polls the legacy namespaces and the ceremony deadlocks. The
+                                // executor (selectKeygenExecutor / shouldUseNewKeygenExecution)
+                                // still
+                                // gates GG20 keygen out of the batched path defensively.
+                                isTssBatch = message.isTssBatch,
```
rkokhatskyi: Unlike the reshare branch near line `242` which ANDs `message.isTssBatch` with `isBatchEligibleReshare(...)`, this branch trusts the raw QR flag and stays safe only because `selectKeygenExecutor` sends GG20 to the legacy path, an invariant that lives in `KeygenExecution.kt`, so a later change letting `selectKeygenExecutor` or `startKeygenGG20` read the flag would silently reopen forged QR routing here, whereas mirroring the `&& isBatchEligible...` guard on this branch would keep it self defending

rkokhatskyi (changes requested):

gomesalexandre (dismissed):
## prior issues — status

| round | # | severity | issue | status | evidence |
|---|---|---|---|---|---|
| r1 | 1 | changes-requested | `rkokhatskyi`: KEYGEN action now routes exclusively through `args.isTssBatch`, but an Android-initiated KEYGEN never sets that flag in `createKeygenPayload` (KEYGEN QR has no `isTssBatch` field), so the batched KEYGEN path is unreachable when Android is initiator while the FastVault server still calls `joinBatchKeygen` when `isTssBatchEnabled` | ❌ still blocking | see analysis below |
| r1 | 2 | should-fix | `rkokhatskyi`: `isParallelKeygenEnabledForThisCeremony` is untested for action/flag matrix | ⚠️ partial | `KeyImportRoutingServerMatchTest` locks per-chain routing but does NOT exercise the `when(action)` dispatch |
| r1 | 3 | should-fix | `rkokhatskyi`: `ROOT_ECDSA_KEY_IMPORT_MESSAGE_ID` is now dead in prod code but lives on in `ReshareRoutingServerMatchTest` as a literal-value assertion | ⏭️ not tackled | `ReshareRoutingServerMatchTest.kt` line 52 still asserts it |
| r1 | 4 | nitpick | `coderabbitai`: `assert()` instead of `assertNotEquals()` in `root setup namespaces never collide` | ⏭️ not tackled | `KeyImportRoutingServerMatchTest.kt` line 91-104 |
| r1 | 5 | should-fix | `rkokhatskyi`: test for batched per-chain routing uses `"bitcoin"` (lowercase) but `chainSetting.chain.raw` is `"Bitcoin"` for `Chain.Bitcoin` — the test locks a channel production never emits | ❌ still blocking | `KeyImportRoutingServerMatchTest.kt` line 69 |

---

## Review (round 2)

### The core fix (sequential per-chain exchange channel)

Correct and clearly explained. The root cause — `KeygenRouting.from(setupMessageId = chainName)` defaulting `exchangeMessageId` to the setup id while iOS/extension left it unset — is pinned precisely. Passing `exchangeMessageId = ""` to force the `null` default is load-bearing and the comment says exactly why. No disagreement here.

### The `isTssBatch` proto round-trip

Mapper changes are minimal and correct. `KeygenMessage.isTssBatch` defaults to `false`, preserving backward compat with older initiators that don't send the field. `KeygenMessageProtoMapperTest` round-trips both true and false. Good.

---

### blocker: Android-initiated KEYGEN loses the batched path

This is rkokhatskyi's first finding and it still applies after this commit.

`isParallelKeygenEnabledForThisCeremony()` now returns `args.isTssBatch` for `TssAction.KEYGEN`. But when Android is the INITIATOR of a KEYGEN, `createKeygenPayload()` builds the QR `KeygenMessageProto` **without setting `isTssBatch`** (it's left at proto default `false`). So `args.isTssBatch` will always be `false` for an Android-initiated KEYGEN, and the batched KEYGEN path is dead for that case.

Meanwhile in `KeygenPeerDiscoveryViewModel` around line 741:
```kotlin
TssAction.KEYGEN -> {
    if (isTssBatchEnabled) {
        vultiSignerRepository.joinBatchKeygen(...)  // server gets batched call
    }
}
```

The FastVault server is told to run batch keygen when `isTssBatchEnabled`, but `isParallelKeygenEnabledForThisCeremony()` in `KeygenViewModel` only returns true when `args.isTssBatch` — which was never set in the QR. So the local Android peer drives the legacy namespace while the FastVault server drives `p-ecdsa` / `p-eddsa`. Deadlock.

Before this PR the `|| featureFlags.isTssBatchEnabled` kept them in sync. This PR deliberately removed it for KEYGEN but didn't fix the QR payload to carry the flag. The fix is either:
- set `isTssBatch = isTssBatchEnabled` in the KEYGEN `createKeygenPayload` QR proto, OR
- keep KEYGEN on `featureFlags.isTssBatchEnabled || args.isTssBatch` (which is safe because the joiner's local flag value doesn't matter for KEYGEN — Android is always the initiator when producing a KEYGEN QR, so both sides of an Android↔Android KEYGEN would have the same local flag)

The fix that makes the most sense long-term is the former (put the flag in the QR), mirroring what iOS does.

### preferably-blocking: per-chain test uses `"bitcoin"` (lowercase) but production uses `Chain.Bitcoin.raw`

`KeyImportRoutingServerMatchTest.kt` line 69 tests the batched per-chain routing with `val chain = "bitcoin"` and asserts `p-bitcoin`. But `runKeyImportChainKeygen` uses `chainSetting.chain.raw`, and for `Chain.Bitcoin` that value is `"Bitcoin"` (capital B). Production emits `p-Bitcoin` for the Bitcoin per-chain batched exchange channel, not `p-bitcoin`. The test locks a contract that production never produces. A regression that changed casing would still pass this test.

### should-fix: `isParallelKeygenEnabledForThisCeremony` needs action-matrix unit test

rkokhatskyi flagged this. The new `when(action)` dispatch is critical routing logic but has zero direct test coverage. Adding a parameterized test that verifies the output for `(ReShare, isTssBatch=true)`, `(KEYGEN, isTssBatch=false, flagEnabled=true)`, `(KeyImport, isTssBatch=false)`, `(Migrate, flagEnabled=true)` etc. would prevent a regression silently reintroducing the `||` on `KEYGEN`/`KeyImport`.

### should-fix: `ROOT_ECDSA_KEY_IMPORT_MESSAGE_ID` dead in prod, lingering in a test

The constant is no longer imported in `KeygenViewModel.kt` (import removed in this diff), but `ReshareRoutingServerMatchTest.kt` line 52 still asserts its literal value: `assertEquals("ecdsa_key_import", ROOT_ECDSA_KEY_IMPORT_MESSAGE_ID)`. That test is now asserting a tautology about a constant that's only used in the test itself. Either remove both the constant and the test, or if the constant is needed for reshare context (it is referenced in `KeygenExecution.kt` as documentation), rename the assertion to clarify it's for reshare, not key-import.

### suggestion: `coderabbitai` nitpick — `assert()` -> `assertNotEquals()` in `root setup namespaces never collide`

Low stakes but Mr. Rabbit is right here. `assert()` doesn't give you a nice failure message showing the actual values. `assertNotEquals(ecdsa.setupMessageId, eddsa.setupMessageId, "...")` is 5 chars of extra work.

### praise

The PR description is excellent — the root cause table showing per-platform relay channel behavior is exactly what a cross-platform bug needs. The comment in `runKeyImportChainKeygen` explaining why `exchangeMessageId = ""` is load-bearing is the kind of thing that saves the next engineer 2 hours. The `KeygenMessageProtoMapperTest` default-false test (legacy backward compat) is a nice touch.

---

**Verdict: REQUEST_CHANGES.** The KEYGEN batched path regression is a real deadlock vector when `isTssBatchEnabled` is on and Android initiates keygen, and the per-chain test case uses the wrong casing to lock the channel name. The core sequential key-import fix is solid — just needs these two things cleaned up before this lands.

app/src/main/java/com/vultisig/wallet/ui/models/keygen/KeygenViewModel.kt
aminsato: Good catch — confirmed. The QR never carries `is_tss_batch` for KEYGEN (`createKeygenPayload` does not set it and `isBatchEligibleReshare` is false for KEYGEN), so `args.isTssBatch` was always false, while the FastVault server still calls `joinBatchKeygen` whenever `isTssBatchEnabled`. App on the legacy path + server on `p-ecdsa`/`p-eddsa` = deadlock.

Fixed in 49de077 by keeping the OR fallback for KEYGEN (now grouped with Migrate/SingleKeygen): `isTssBatchEnabled || args.isTssBatch`, which matches the `joinBatchKeygen` gate. ReShare/KeyImport stay QR-exclusive.

Tradeoff noted: this re-opens the cross-version risk for an Android *joiner* with the flag on vs an initiator with it off — but only reachable when `isTssBatchEnabled` is actually enabled. Happy to move to a fully QR-driven solution (write the flag into the KEYGEN QR + gate the server join on it) if you prefer that over the fallback.

app/src/main/java/com/vultisig/wallet/ui/models/keygen/KeygenViewModel.kt
aminsato: Done in 49de077. Extracted the matrix into a pure `resolveParallelKeygenOptIn(action, qrIsTssBatch, isTssBatchFeatureEnabled)` next to `shouldUseNewKeygenExecution`, and added `ParallelKeygenOptInTest` covering every action. The KEYGEN/flag-on/QR-off case is the explicit regression guard — reverting KEYGEN to the QR-exclusive form now fails the suite.

app/src/main/java/com/vultisig/wallet/ui/models/keygen/KeygenViewModel.kt
aminsato: Confirmed — it was the last production use. Removed `ROOT_ECDSA_KEY_IMPORT_MESSAGE_ID` and its assertion in `ReshareRoutingServerMatchTest` in 49de077, and left a short note where it was declared explaining ECDSA key-import setup runs on the default namespace (so there is no ECDSA key-import constant). `ROOT_EDDSA_KEY_IMPORT_MESSAGE_ID` stays since it is still used.

app/src/test/java/com/vultisig/wallet/data/keygen/KeyImportRoutingServerMatchTest.kt
aminsato: Both points fixed in 49de077. `Chain.Bitcoin.raw` is `"Bitcoin"`, so the test now asserts `p-Bitcoin` and uses `Chain.Bitcoin.raw`/`Chain.Ethereum.raw` rather than lowercased literals. I also extracted the routing into a pure `keyImportChainRouting(chainRaw, useParallelPath)` that the ViewModel now calls, and the test exercises that exact helper — so dropping the load-bearing empty sequential `exchangeMessageId` (or lowercasing the namespace) breaks the test instead of slipping through.

app/src/main/java/com/vultisig/wallet/ui/models/keygen/JoinKeygenViewModel.kt
aminsato: Agreed — made it self-defending in 49de077. Factored `isBatchEligibleCeremony(action, libType)` out of `shouldUseNewKeygenExecution` and AND it into this branch: `isTssBatch = message.isTssBatch && isBatchEligibleCeremony(action, message.libType)`. A forged QR with `isTssBatch=true` on a GG20 keygen is now pinned off at the join boundary, so it no longer depends on `selectKeygenExecutor`/`shouldUseNewKeygenExecution` staying the only gate.

rkokhatskyi (approved):
LGTM

PR #4645, opened 2026-05-26
fix: replace .body<T>() with bodyOrThrow<T>() in API layer (#3865)

### LINKED ISSUE #3865: fix: replace all .body<T>() with bodyOrThrow<T>() in API layer

aminsato:
## Summary

Closes #3865. Swaps raw `.body<T>()` for `.bodyOrThrow<T>()` on success-path reads across the API layer so HTTP errors are wrapped consistently (per the two-layer error-handling architecture in `docs/network-error-handling.md`).

This is the second half of a two-PR, test-first flow. The characterization tests landed first in #4629 (now merged); this PR applies the production swap and those same tests stay green.

## Changes

- Success-path reads in 17 API files now use `bodyOrThrow<T>()` instead of `body<T>()`.
- Error-parsing branches that read **non-2xx** bodies are intentionally left **not** using `bodyOrThrow` (it throws on non-2xx, which would break their error handling):
  - `KyberApi` / `LiFiChainApi`: read via `bodyAsText()`. LiFi has no `isSuccess` guard and relies on its serializer to surface the body's error `message`, so it must not throw on non-2xx.
  - `OneInchApi` / `CardanoApi` BadRequest branch: `json.decodeFromString(bodyAsText())`.
- `CardanoApi` gains a private `Json` for the BadRequest decode (constructor unchanged). Also removed a dead commented-out `broadcastTransaction` block and its orphaned `url2` field.

## Test plan

- `grep '\.body<' data/.../api/` → 0 remaining usages.
- All 16 characterization classes from #4629 pass against the refactored code: **69 tests, 0 failures** via `./gradlew :data:testDebugUnitTest`.
- `./gradlew :data:ktfmtCheckMain` clean.

## Summary by CodeRabbit

* **Refactor**
  * Strengthened HTTP response parsing and error handling across many blockchain and swap integrations to reduce silent failures and improve reliability.
* **Bug Fixes**
  * Made parsing stricter so malformed or unexpected responses surface as errors rather than silently producing nulls.
* **Tests**
  * Updated tests to align with the new JSON parsing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4645?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/BittensorApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/BittensorApi.kt
@@ -10,6 +10,7 @@ import com.vultisig.wallet.data.api.models.cosmos.PolkadotQueryInfoResponseJson
 import com.vultisig.wallet.data.api.utils.postRpc
 import com.vultisig.wallet.data.utils.bodyOrThrow
 import io.ktor.client.HttpClient
+import io.ktor.client.call.body
 import io.ktor.client.request.get
 import io.ktor.client.request.post
 import io.ktor.client.request.setBody
@@ -125,7 +126,10 @@ internal class BittensorApiImp @Inject constructor(private val httpClient: HttpC
                 id = 1,
             )
         val response = httpClient.post(BITTENSOR_RPC_URL) { setBody(payload) }
-        val responseContent = response.bodyOrThrow<PolkadotBroadcastTransactionJson>()
+        // Read the RPC body regardless of HTTP status: the relay may surface "Already Imported"
+        // with a non-2xx status, and the duplicate rebroadcast from a second signing device must
+        // still resolve to null rather than throw.
+        val responseContent = response.body<PolkadotBroadcastTransactionJson>()
         if (responseContent.error != null) {
             if (responseContent.error.message?.contains("Already Imported") == true) {
                 return null
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/BittensorApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/BittensorApi.kt
@@ -8,8 +8,8 @@ import com.vultisig.wallet.data.api.models.cosmos.PolkadotGetNonceJson
 import com.vultisig.wallet.data.api.models.cosmos.PolkadotGetRunTimeVersionJson
 import com.vultisig.wallet.data.api.models.cosmos.PolkadotQueryInfoResponseJson
 import com.vultisig.wallet.data.api.utils.postRpc
+import com.vultisig.wallet.data.utils.bodyOrThrow
 import io.ktor.client.HttpClient
-import io.ktor.client.call.body
 import io.ktor.client.request.get
 import io.ktor.client.request.post
 import io.ktor.client.request.setBody
@@ -125,7 +125,7 @@ internal class BittensorApiImp @Inject constructor(private val httpClient: HttpC
                 id = 1,
             )
         val response = httpClient.post(BITTENSOR_RPC_URL) { setBody(payload) }
-        val responseContent = response.body<PolkadotBroadcastTransactionJson>()
+        val responseContent = response.bodyOrThrow<PolkadotBroadcastTransactionJson>()
         if (responseContent.error != null) {
             if (responseContent.error.message?.contains("Already Imported") == true) {
                 return null
@@ -174,7 +174,7 @@ internal class BittensorApiImp @Inject constructor(private val httpClient: HttpC
         val hash = if (txHash.startsWith("0x")) txHash else "0x$txHash"
         val response = httpClient.get("${TAOSTATS_PROXY_URL}/v1?hash=$hash")
         if (response.status == HttpStatusCode.NotFound) return null
-        return response.body<TaostatsExtrinsicResponse>().data?.firstOrNull()
+        return response.bodyOrThrow<TaostatsExtrinsicResponse>().data?.firstOrNull()
     }
 
     private companion object {
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
@@ -8,6 +8,7 @@ import com.vultisig.wallet.data.api.models.cardano.CardanoUtxoResponseJson
 import com.vultisig.wallet.data.api.models.cardano.OgmiosTransactionResponse
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.payload.UtxoInfo
+import com.vultisig.wallet.data.utils.bodyOrThrow
 import io.ktor.client.HttpClient
 import io.ktor.client.call.body
 import io.ktor.client.request.get
@@ -19,6 +20,7 @@ import io.ktor.http.path
 import java.math.BigInteger
 import javax.inject.Inject
 import kotlin.coroutines.cancellation.CancellationException
+import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.buildJsonObject
 import kotlinx.serialization.json.put
 import timber.log.Timber
@@ -36,9 +38,12 @@ interface CardanoApi {
 }
 
 internal class CardanoApiImpl @Inject constructor(private val httpClient: HttpClient) : CardanoApi {
+    private val json = Json {
+        ignoreUnknownKeys = true
+        explicitNulls = false
+    }
     private val url: String = "https://api.koios.rest"
     private val apiV1Path: String = "api/v1"
-    private val url2 = "https://api.vultisig.com"
     private val ogmiosUrl = "https://api.vultisig.com/ada/"
 
     override suspend fun getBalance(coin: Coin): BigInteger {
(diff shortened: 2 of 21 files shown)


johnnyluo (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/api/swapAggregators/OneInchApi.kt
```diff
@@ -86,7 +87,10 @@ constructor(
                 listOf(swapResponseAsync, quoteResponseAsync).awaitAll().also { responses ->
                     responses.forEach { response ->
                         if (!response.status.isSuccess()) {
-                            val resp = response.body<OneInchSwapQuoteErrorResponse>()
+                            val resp =
+                                json.decodeFromString<OneInchSwapQuoteErrorResponse>(
```
rkokhatskyi, line 91: This file is only half migrated since `getTokens` at line 140 and the `swapResponse.body()` and `quoteResponse.body()` reads at lines 104 and 105 still use raw `body()` while `getContractsWithBalance` and `getTokensByContracts` now go through `bodyOrThrow`, so one method throws a raw `JsonConvertException` whereas the siblings throw `NetworkException` which is the very inconsistency this PR set out to remove, and the `grep '.body<'` in the description missed these since the inferred `body()` form has no angle brackets

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
```diff
@@ -36,9 +38,12 @@ interface CardanoApi {
 }
 
 internal class CardanoApiImpl @Inject constructor(private val httpClient: HttpClient) : CardanoApi {
+    private val json = Json {
```
rkokhatskyi: `OneInchApiImpl`, `KyberApiImpl` and `LiFiChainApiImpl` all receive the shared singleton `Json` through constructor injection whereas this adds a private `Json` that drops the `encodeDefaults` and `serializersModule` the app wide instance carries, so injecting `json: Json` here instead would keep Cardano decoding consistent with the rest of the codebase since the singleton already sets `ignoreUnknownKeys` and `explicitNulls` to the same values

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
```diff
@@ -69,7 +74,7 @@ internal class CardanoApiImpl @Inject constructor(private val httpClient: HttpCl
             }
 
         return try {
-            response.body<List<CardanoUtxoResponseJson>>().toUtxos()
+            response.bodyOrThrow<List<CardanoUtxoResponseJson>>().toUtxos()
```
rkokhatskyi, line 74: `getBalance` at line 58 and `getCurrentSlot` at line 152 still read through raw `body()` while `getUTXOs`, the `OK` branch and `getTxStatus` were converted, so this file ends up half migrated with mixed error wrapping which works against the consistent wrapping this PR is going for

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
```diff
@@ -114,7 +119,8 @@ internal class CardanoApiImpl @Inject constructor(private val httpClient: HttpCl
                 }
 
                 HttpStatusCode.BadRequest -> {
-                    val ogmiosError = response.body<OgmiosTransactionResponse>()
+                    val ogmiosError =
+                        json.decodeFromString<OgmiosTransactionResponse>(response.bodyAsText())
```
rkokhatskyi, line 120: With the commented out `broadcastTransaction` block and `url2` now gone, a repo wide grep shows `CardanoBroadcastResponseJson`, `CardanoTransactionHashRequestBodyJson` and `CardanoBroadcastDataResponseJson` are referenced only by their own definitions, so those three files under `models/cardano/` are dead and worth removing as part of this same cleanup

data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
```diff
@@ -152,7 +151,7 @@ internal class PolkadotApiImp @Inject constructor(private val httpClient: HttpCl
                 id = 1,
             )
         val response = httpClient.post(POLKADOT_API_URL) { setBody(payload) }
-        val responseContent = response.body<PolkadotBroadcastTransactionJson>()
+        val responseContent = response.bodyOrThrow<PolkadotBroadcastTransactionJson>()
```
rkokhatskyi: `bodyOrThrow` throws on any status outside 2xx before this runs whereas the `1012` and `1013` branch returns `null` to treat an already in pool extrinsic as a successful idempotent rebroadcast, so if the node ever surfaces these with a status outside 2xx the second signing device gets a hard failure instead of `null`, and note iOS deliberately reads the substrate RPC body without inspecting HTTP status in `RpcServiceStruct` for exactly this multi device case

data/src/main/kotlin/com/vultisig/wallet/data/api/BittensorApi.kt
```diff
@@ -125,7 +125,7 @@ internal class BittensorApiImp @Inject constructor(private val httpClient: HttpC
                 id = 1,
             )
         val response = httpClient.post(BITTENSOR_RPC_URL) { setBody(payload) }
-        val responseContent = response.body<PolkadotBroadcastTransactionJson>()
+        val responseContent = response.bodyOrThrow<PolkadotBroadcastTransactionJson>()
```
rkokhatskyi: Same risk as the Polkadot broadcast since `bodyOrThrow` throws before the `Already Imported` branch can return `null` whenever the status falls outside 2xx, whereas iOS `BittensorService` explicitly suppresses `already imported` as idempotent success for the second device, so a duplicate rebroadcast could turn into a hard failure here if the relay returns that RPC error with a status outside 2xx

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/api/swapAggregators/OneInchApi.kt
aminsato: Good catch — these had no angle brackets so the grep missed them. Migrated all three success-path reads: `getTokens` now uses `bodyOrThrow<OneInchTokensJson>()`, and the `swap`/`quote` reads in `getSwapQuote` use `bodyOrThrow<JsonElement>()`, so the file is fully consistent now. Fixed in 62a85e987.

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
aminsato: Done — dropped the private `Json` and now inject the singleton `json: Json` via the constructor. The singleton is a superset of the old private config (same `ignoreUnknownKeys`/`explicitNulls`, plus `encodeDefaults` and `serializersModule`), so the BadRequest decode is unaffected. Fixed in 62a85e987.

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
aminsato: Fixed — `getBalance` and `getCurrentSlot` now read through `bodyOrThrow`, so the file no longer mixes wrapping styles. Fixed in 62a85e987.

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
aminsato: Confirmed dead via repo-wide grep (only self-references remained). Deleted all three: `CardanoBroadcastResponseJson`, `CardanoBroadcastDataResponseJson`, and `CardanoTransactionHashRequestBodyJson`. Fixed in 62a85e987.

data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
aminsato: Agreed — reverted `broadcastTransaction` back to raw `body()` so a non-2xx status can't short-circuit the `1012`/`1013` idempotent-rebroadcast path before it returns null, matching the iOS `RpcServiceStruct` behavior you referenced. Added a comment explaining why this one stays on `body()`. Fixed in 62a85e987.

data/src/main/kotlin/com/vultisig/wallet/data/api/BittensorApi.kt
aminsato: Same fix as Polkadot — reverted `broadcastTransaction` to raw `body()` so the "Already Imported" idempotent branch still returns null on a non-2xx RPC error, matching iOS `BittensorService`. Comment added. Fixed in 62a85e987.

johnnyluo (approved):
LGTM

PR #4646, opened 2026-05-26
fix(keysign): correct custom message signature format for ECDSA chains

### LINKED ISSUE #4624: Custom message signing: signature hash mismatch between Android and Extension/iOS
## Summary

When signing a **custom message** across platforms, the resulting signature shown on the **Transaction Done** screen does **not match** when one of the participants is Android.

- ❌ Android ↔ Extension — signatures differ
- ❌ Android ↔ iOS — signatures differ
- ✅ iOS ↔ Extension — signatures match

This indicates an Android-side discrepancy in how the custom message is prepared, hashed, or how the resulting signature is encoded/displayed.

## Steps to reproduce

1. Create / use a multi-device vault with one Android device + one Extension (or iOS) device.
2. Initiate a **custom message** signing session from either side.
3. Complete signing on both devices.
4. Open the **Transaction Done** screen on each device and compare the displayed `Signature` value.

## Expected

Signature hash shown on the Transaction Done screen is identical on all participating devices (as it is for iOS ↔ Extension today).

## Actual

- Android shows a different signature hash than the Extension / iOS counterpart.
- iOS ↔ Extension agree, so the divergence is isolated to Android.

## Scope

- Affects: custom message signing only (other signing flows to be confirmed).
- Cross-platform pairs tested:
  - Android ↔ Extension → mismatch
  - Android ↔ iOS → mismatch
  - iOS ↔ Extension → match

## Notes

- Same vault, same message payload, same session — only the displayed signature differs.
- Worth checking on Android: message pre-hash / prefix handling, byte-order of the signature components, and the final hex/encoding used to render the value on the Transaction Done screen.

## Environment

- App: Vultisig Android (current `main`)
- Counterpart platforms: Vultisig iOS, Vultisig Extension

johnnyluo:
## Summary

- `calculateCustomMessageSignature` was unconditionally calling `getSignature()` — the EdDSA little-endian format (`reversed(r) + reversed(s)`, 64 bytes) — for all custom-message key types
- ECDSA chains (Ethereum and most others) require `getSignatureWithRecoveryID()` — standard big-endian `r + s + recoveryId` (65 bytes) — which is what iOS and Extension produce
- This caused the Transaction Done screen to show a different signature hash on Android vs iOS/Extension for the same signing session
- Fix dispatches on `keyType` in `calculateCustomMessageSignature`: ECDSA/MLDSA → `getSignatureWithRecoveryID()`; EdDSA → `getSignature()`

## Issue

Closes #4624

## Test Plan

- [x] `./gradlew lint` — passes ✅
- [x] `./gradlew :data:testDebugUnitTest` — `SigningHelperCustomMessageTest` (5 tests) passes ✅
- [x] `./gradlew :app:testDebugUnitTest` — `CustomMessageSignatureFormatTest` (4 tests) passes ✅
- [x] Manual: perform custom message signing with Android + iOS or Android + Extension; verify the `Signature` field on the Transaction Done screen is identical across both devices

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signature derivation to correctly handle different cryptographic key types (EdDSA, ECDSA, ML-DSA)
  * Enhanced payload source detection for more reliable key type identification

* **Tests**
  * Added comprehensive test coverage for custom message signature formats and encoding validation
  * Added test coverage for message signing behavior across different key types

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4646?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -677,7 +678,14 @@ constructor(
...
     private fun calculateCustomMessageSignature(keysignResp: KeysignResponse) {
         if (customMessagePayload == null) return
-        txHash.value = keysignResp.getSignature().toHexString()
+        txHash.value =
+            when (keyType) {
+                // EdDSA chains (Solana, TON, etc.) use little-endian reversed r+s
+                TssKeyType.EDDSA -> keysignResp.getSignature().toHexString()
+                // ECDSA chains use standard r+s+recoveryId (matches iOS/Extension)
+                TssKeyType.ECDSA,
+                TssKeyType.MLDSA -> keysignResp.getSignatureWithRecoveryID().toHexString()
```
rkokhatskyi: `MldsaKeysign.reportSignature` populates only `msg` and `derSignature` on the `KeysignResponse` and never sets `r`, `s` or `recoveryID`, whereas `getSignatureWithRecoveryID()` reads exactly those three fields so a `qbtc` custom message reaching this branch on the joining device makes each `Numeric.hexStringToByteArray("")` return empty and `txHash.value` resolve to an empty string while the signature held in `derSignature` is dropped, route `TssKeyType.MLDSA` to `keysignResp.derSignature` or make it an explicit error the way the legacy path does at line 646

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -677,7 +678,14 @@ constructor(
     @OptIn(ExperimentalStdlibApi::class)
     private fun calculateCustomMessageSignature(keysignResp: KeysignResponse) {
         if (customMessagePayload == null) return
-        txHash.value = keysignResp.getSignature().toHexString()
+        txHash.value =
+            when (keyType) {
+                // EdDSA chains (Solana, TON, etc.) use little-endian reversed r+s
+                TssKeyType.EDDSA -> keysignResp.getSignature().toHexString()
```
rkokhatskyi, line 684: This EdDSA branch and the `MLDSA` branch are reachable only on the joining device since `KeysignFlowViewModel.tssKeysignType` at line 205 used by the initiating device derives `keyType` solely from `_keysignPayload?.coin?.chain` and falls back to `ECDSA` whenever `keysignPayload` is null which always holds for a custom message, so for an EdDSA chain custom message an Android initiator renders `getSignatureWithRecoveryID` while an Android joiner renders this branch and the two devices show different values, mirror the `customMessagePayload?.chain` fallback that `JoinKeysignViewModel` already applies at line 300

app/src/test/java/com/vultisig/wallet/data/tss/CustomMessageSignatureFormatTest.kt
```diff
@@ -0,0 +1,85 @@
...
+ */
+class CustomMessageSignatureFormatTest {
+
+    // Realistic 32-byte r / s / recoveryId hex strings (secp256k1 / ed25519 shape)
+    private val R_HEX = "a1b2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8f9"
+    private val S_HEX = "b1c2d3e4f5061718192a3b4c5d6e7f80b1c2d3e4f5061718192a3b4c5d6e7f8"
+    private val RECOVERY_ID_HEX = "00"
+
+    // Same byte logic as TssExtensions.getSignatureWithRecoveryID()
+    private fun ecdsaSignatureBytes(r: String, s: String, recoveryId: String): ByteArray =
```
rkokhatskyi, line 33: These helpers reimplement the byte logic of `getSignatureWithRecoveryID()` and `getSignature()` instead of calling them so every test in this file asserts a private copy against itself, which means a regression in `TssExtensions.kt` such as swapped `r` and `s` or a dropped recovery byte would leave all four tests green and contradicts the docstring claim that such a change will be caught immediately, since the production functions only read the `r`, `s` and `recoveryID` strings you can extract the byte logic into JNI free top level functions and have these tests exercise those

app/src/test/java/com/vultisig/wallet/data/tss/CustomMessageSignatureFormatTest.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.vultisig.wallet.data.utils.Numeric
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.shouldNotBe
+import org.junit.jupiter.api.Test
+
+/**
+ * Specifies the byte-level encoding for ECDSA and EdDSA custom-message signatures.
+ *
+ * Regression guard for issue #4624: Android was using the EdDSA encoding (reversed r+s, no recovery
+ * ID) for all custom messages, including ECDSA chains. iOS and Extension use the standard ECDSA
```
rkokhatskyi, line 12: iOS `customMessageSignature()` calls `getSignatureWithRecoveryID()` unconditionally with no `keyType` branch so iOS renders `r + s + recoveryId` even for EdDSA chains, which means the EdDSA branch this PR keeps with reversed `r` and `s` at 64 bytes matches the Extension but not iOS so the Android to iOS divergence that #4624 reports still stands for EdDSA chain custom messages while only the ECDSA case is fully resolved, worth narrowing this claim or confirming which platform is canonical for EdDSA

data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelperCustomMessageTest.kt
```diff
@@ -0,0 +1,80 @@
...
+package com.vultisig.wallet.data.chains.helpers
+
+import com.vultisig.wallet.data.common.toHexBytes
+import com.vultisig.wallet.data.common.toKeccak256ByteArray
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.shouldNotBe
+import org.junit.jupiter.api.Test
+import vultisig.keysign.v1.CustomMessagePayload
+
+class SigningHelperCustomMessageTest {
```
rkokhatskyi, line 12: The `eth_signTypedData_v4` method which routes through `getKeysignMessagesForTypedData` and `EthereumAbi.encodeTyped` is the common dApp signing path yet has no case here, and since `EthereumAbi.encodeTyped` is a native TrustWalletCore call absent from the `:data` unit test classpath it can be covered only by an instrumented test or by injecting the hash function, worth either covering it that way or documenting that this branch is intentionally uncovered

rkokhatskyi (changes requested):

app/src/test/java/com/vultisig/wallet/data/tss/CustomMessageSignatureFormatTest.kt
```diff
@@ -0,0 +1,85 @@
...
+ * These tests verify the byte invariants for each format so a future change to TssExtensions.kt
+ * that breaks cross-platform parity will be caught immediately. Note: tss.KeysignResponse requires
+ * JNI and cannot be instantiated in JVM unit tests, so the format logic is exercised here with the
+ * same Numeric utility used by the production code.
+ */
+class CustomMessageSignatureFormatTest {
+
+    // Realistic 32-byte r / s / recoveryId hex strings (secp256k1 / ed25519 shape)
+    private val R_HEX = "a1b2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8f9"
+    private val S_HEX = "b1c2d3e4f5061718192a3b4c5d6e7f80b1c2d3e4f5061718192a3b4c5d6e7f8"
```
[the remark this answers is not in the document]
johnnyluo: Fixed in 40def759a — R_HEX and S_HEX now use exactly 64 hex chars (true 32-byte fixtures).

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
johnnyluo: Fixed in 40def759a — MLDSA is now routed to `keysignResp.derSignature` directly, since `MldsaKeysign.reportSignature` only populates `msg` and `derSignature`, not `r`/`s`/`recoveryID`.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
johnnyluo: Valid observation. The initiator's `tssKeysignType` always falls back to ECDSA when `keysignPayload` is null, while the joiner correctly reads the chain from `customMessagePayload`. This is a pre-existing asymmetry not introduced by this PR (the PR only changes the signature format dispatch, not which key is used). Tracked for a follow-up fix in `KeysignFlowViewModel`.

app/src/test/java/com/vultisig/wallet/data/tss/CustomMessageSignatureFormatTest.kt
johnnyluo: Acknowledged in the test file's own KDoc: `tss.KeysignResponse` requires JNI and cannot be instantiated in JVM unit tests. Extracting the byte logic into JNI-free top-level functions is a worthwhile future refactor but goes beyond the scope of this bug fix.

app/src/test/java/com/vultisig/wallet/data/tss/CustomMessageSignatureFormatTest.kt
johnnyluo: Valid cross-platform observation. iOS calling `getSignatureWithRecoveryID()` unconditionally for EdDSA is a separate investigation — narrowing the claim in the KDoc or confirming the canonical platform for EdDSA custom messages is tracked as a follow-up.

data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelperCustomMessageTest.kt
johnnyluo: `eth_signTypedData_v4` coverage via `EthereumAbi.encodeTyped` (a native TrustWalletCore call) would require either an instrumented test or injecting the hash function — that's a separate PR. Noted for follow-up.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
rkokhatskyi: Filed #4659 to track the `KeysignFlowViewModel.tssKeysignType` fix so the initiator and joiner derive the same key type for custom messages

app/src/test/java/com/vultisig/wallet/data/tss/CustomMessageSignatureFormatTest.kt
rkokhatskyi: Filed #4660 to track the iOS EdDSA parity investigation and the KDoc narrowing

data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelperCustomMessageTest.kt
rkokhatskyi: Filed #4661 to track the `eth_signTypedData_v4` coverage

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -677,7 +678,15 @@ constructor(
     @OptIn(ExperimentalStdlibApi::class)
     private fun calculateCustomMessageSignature(keysignResp: KeysignResponse) {
         if (customMessagePayload == null) return
-        txHash.value = keysignResp.getSignature().toHexString()
+        txHash.value =
+            when (keyType) {
```
aminsato, line 682: This dispatch is only as reliable as `keyType`, and the two entry points derive it differently for custom messages:

- **Join path** (`JoinKeysignViewModel.kt:300-305`) falls back to `customMessagePayload.chain`, so EdDSA/MLDSA are reachable here. ✅
- **Initiate path** (`KeysignFlowViewModel.kt:206`): `tssKeysignType = _keysignPayload?.coin?.chain?.TssKeysignType ?: TssKeyType.ECDSA`. A custom message has no `keysignPayload`, so this always resolves to **ECDSA** and never consults `customMessagePayload.chain`.

Net effect: on the initiating device the EdDSA/MLDSA branches are effectively dead code, and a non-ECDSA custom message would be ECDSA on the initiator but EdDSA on the joiner — the same class of mismatch this PR fixes. For #4624 (Ethereum) both flows agree on ECDSA, so the shipped fix is sound. Consider aligning `KeysignFlowViewModel.tssKeysignType` with the join VM (consult `customMessagePayload.chain`), or add a comment that initiator-side custom messages are ECDSA-only.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -677,7 +678,15 @@ constructor(
...
+                // EdDSA chains (Solana, TON, etc.) use little-endian reversed r+s
+                TssKeyType.EDDSA -> keysignResp.getSignature().toHexString()
```
aminsato, line 684: Divergence from iOS to confirm: iOS `KeysignViewModel.customMessageSignature()` uses `getSignatureWithRecoveryID()` **unconditionally** (its comment: "currently keysign for custom message is using ETH"). This branch returns reversed `r+s` (64 bytes) instead of `r+s+recoveryId` (65 bytes), so if an EdDSA custom message ever surfaces the displayed signature will differ from iOS. Android's form is arguably the more correct ed25519 encoding — just confirm the divergence is intentional rather than incidental.

app/src/test/java/com/vultisig/wallet/data/tss/CustomMessageSignatureFormatTest.kt
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Specifies the byte-level encoding for ECDSA and EdDSA custom-message signatures.
+ *
+ * Regression guard for issue #4624: Android was using the EdDSA encoding (reversed r+s, no recovery
+ * ID) for all custom messages, including ECDSA chains. iOS and Extension use the standard ECDSA
+ * encoding (r + s + recoveryId, big-endian), so the displayed signatures mismatched.
+ *
+ * The fix dispatches on keyType in KeysignViewModel.calculateCustomMessageSignature:
+ * - ECDSA/MLDSA → getSignatureWithRecoveryID() = r + s + recoveryId (65 bytes)
```
aminsato: Docstring inconsistency: this says `ECDSA/MLDSA → getSignatureWithRecoveryID()`, but the production code (and this PR) routes **MLDSA → `derSignature`**, not `getSignatureWithRecoveryID()`. Only ECDSA uses recovery-ID format.

app/src/test/java/com/vultisig/wallet/data/tss/CustomMessageSignatureFormatTest.kt
```diff
@@ -0,0 +1,85 @@
...
+ */
+class CustomMessageSignatureFormatTest {
+
+    // Realistic 32-byte r / s / recoveryId hex strings (secp256k1 / ed25519 shape)
+    private val R_HEX = "a1b2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8f90"
+    private val S_HEX = "b1c2d3e4f5061718192a3b4c5d6e7f80b1c2d3e4f5061718192a3b4c5d6e7f80"
+    private val RECOVERY_ID_HEX = "00"
+
+    // Same byte logic as TssExtensions.getSignatureWithRecoveryID()
+    private fun ecdsaSignatureBytes(r: String, s: String, recoveryId: String): ByteArray =
```
aminsato, line 33: This test reimplements the byte logic locally and never calls the production `getSignature()` / `getSignatureWithRecoveryID()`. The docstring claim (lines 19-20) that "a future change to TssExtensions.kt that breaks cross-platform parity will be caught immediately" isn't met — a regression in those extensions would still pass this test green, because the test asserts on a duplicate.

The extensions only read `r`/`s`/`recoveryID` off `tss.KeysignResponse`, so a `mockk<KeysignResponse>` with those three stubbed would let you assert on the real functions and actually deliver the stated regression-guard intent.

aminsato (commented):
Reviewed the custom-message signature dispatch against both keysign entry points, `TssExtensions.kt`, and the iOS reference (`KeysignViewModel.customMessageSignature()`).

The core fix is correct: ECDSA custom messages now use `getSignatureWithRecoveryID()` (`r+s+recoveryId`, 65 bytes), matching iOS/Extension, and `DklsKeysign` populates `r`/`s`/`recoveryID` so the call is valid. MLDSA → `derSignature` is sound. The new `SigningHelperCustomMessageTest` exercises the real `getKeysignMessages` and is a good regression guard.

Left inline comments on three points: (1) `keyType` is derived differently on the initiate vs join flow for custom messages, (2) the EdDSA branch diverges from iOS, and (3) `CustomMessageSignatureFormatTest` doesn't actually exercise the production functions plus a docstring inconsistency.

Minor PR hygiene: the CodeRabbit summary references swap-quote timeout changes, but those were reverted (`3cf7210`) and aren't in the net diff (3 files: `KeysignViewModel.kt` + 2 tests). Worth trimming the description so reviewers aren't hunting for swap changes.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
johnnyluo: will fix in IOS

app/src/test/java/com/vultisig/wallet/data/tss/CustomMessageSignatureFormatTest.kt
johnnyluo: Fixed in b74309837 — docstring now correctly lists ECDSA → getSignatureWithRecoveryID() (65 bytes), EdDSA → getSignature() (64 bytes), and MLDSA → derSignature (variable DER).

app/src/test/java/com/vultisig/wallet/data/tss/CustomMessageSignatureFormatTest.kt
johnnyluo: Fixed in b74309837 — KDoc updated to remove the false 'caught immediately' claim and now accurately explains that tss.KeysignResponse is a JNI class that cannot be instantiated in JVM unit tests, so the format logic is verified against the same Numeric utility used by production rather than calling TssExtensions directly.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
johnnyluo: Fixed in b74309837 — KeysignFlowViewModel.tssKeysignType now falls back to customMessagePayload.chain when keysignPayload is null, mirroring JoinKeysignViewModel:300-305. This also closes #4659.

rkokhatskyi (approved):
LGTM

aminsato:
LGTM

PR #4647, opened 2026-05-26
fix: THORChain Bond verify screen title + padding mismatch (#4606)

### LINKED ISSUE #4606: fix: initiator vs joiner mismatch on THORChain Bond verify screen (title + address truncation)
## Problem

When signing a THORChain `BOND` transaction, the verify step renders differently on the initiator and the joining device. The two panes show the exact same payload but they use different titles and different horizontal padding on the content card, so a side-by-side review (which is the whole point of a multi-sig confirmation) does not match.

### 1. Screen title

- **Initiator:** `Verify`
- **Joiner:** `Function overview`

Both screens are the same transaction-review step. They should use the same title across initiator and joiner so users on either device know they are looking at the same thing.

### 2. Horizontal padding on the content card is too high on the initiator

The initiator's content card (the panel that holds "You're sending", From / To / Node Address, Operation, Memo, Amount, Network, Network Fee) has noticeably **more horizontal padding** than the joiner's. The joiner's layout is correct — the initiator should be updated to match the joiner, not the other way round.

Visible consequences of the extra padding on the initiator:

- The row container is narrower, so each row has less horizontal space for its value.
- Middle-truncated addresses (`thor1…umdua5t`) on the initiator show **fewer characters** of the prefix and suffix than the joiner shows, even though it is the same address string. On the joiner you can see `thor12a9rpf9u2ulwu…au7xnde8umdua5t`; on the initiator you only see `thor12a9rpf9u2ul…7xnde8umdua5t`. This is purely a side-effect of the narrower row width.
- The Memo wraps to a different number of lines and breaks at a different character.
- The whole card feels visually inset and inconsistent with the rest of the app.

The fix is to align the initiator's content-card horizontal padding to the joiner's. The truncation logic itself is fine — once the available width matches, the displayed characters will match too.

## Expected

- Initiator and joiner use the **same title** for this step.
- The verify content card on the initiator uses the **same horizontal padding** as the joiner — the joiner is the reference, the initiator is what needs to change.
- Once padding matches, the From / To / Node Address rows render at the same width on both devices, so middle-truncation produces the same head/tail slice on both, and the Memo wraps identically.

## Suggested follow-up (optional)

For BOND and similar function calls, From / To / Node Address are all the same address — middle-truncation hides exactly the bytes a user needs to verify. Consider rendering the address full-width / wrapped on this screen so the user can actually read it. Out of scope for this fix, but worth a separate ticket.

## Screenshots

| Initiator (padding too high) | Joiner (correct reference) |
| --- | --- |
| initiator | joiner |

johnnyluo:
## Summary
- Fixes the initiator showing \"Verify\" instead of \"Function overview\" on the THORChain Bond verify step — now matches the joiner
- Fixes the initiator's content card having double horizontal padding (32dp) vs the joiner's correct 16dp — caused by the outer `V2Scaffold` in `DepositScreen` (16dp) stacking with the inner `VerifyDepositScreen` padding (16dp)
- Adjusts bottom bar button padding similarly so the sign button aligns with the joiner (outer 16dp + inner 8dp = 24dp total, same as joiner's explicit 24dp)

## Root Cause
`DepositScreen` wraps its entire nav host in a `V2Scaffold` that applies `16dp` horizontal padding. When `VerifyDepositScreen` is rendered inside it (initiator path, `hasToolbar = false`), its own `padding(all = 16.dp)` stacks on top → 32dp total. The joiner path calls `VerifyDepositScreen(hasToolbar = true)` as the root scaffold, so it only sees the screen's own 16dp.

## Fix
1. `DepositScreen.kt`: use `verify_deposit_function_overview` ("Function overview") instead of `verify_transaction_screen_title` ("Verify") for the deposit verify route title
2. `VerifyDepositScreen.kt`: when `hasToolbar = false`, set content horizontal padding to `0.dp` (outer scaffold provides 16dp) and bottom bar horizontal padding to `8.dp` (outer 16dp + 8dp = 24dp)

## Issue
Closes #4606

## Test Plan
- [x] On the initiator device, start a THORChain BOND transaction and reach the verify step — title should now read "Function overview"
- [x] Compare the content card width on initiator vs joiner — addresses should truncate identically
- [x] Memo wraps the same way on both devices
- [x] Joiner layout is unchanged (all padding values unchanged when `hasToolbar = true`)

## Summary by CodeRabbit

* **Style**
  * Updated deposit verification screen title for improved clarity and consistency
  * Enhanced layout spacing to dynamically adjust padding based on interface element presence, improving overall responsiveness

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4647?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/deposit/DepositScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/deposit/DepositScreen.kt
@@ -84,7 +84,7 @@ internal fun DepositScreen(
             title = defaultTitle
         }
         SendDst.VerifyTransaction.staticRoute -> {
-            title = stringResource(R.string.verify_transaction_screen_title)
+            title = stringResource(R.string.verify_deposit_function_overview)
         }
         else -> {
             title = defaultTitle
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt
@@ -125,7 +125,7 @@ internal fun VerifyDepositScreen(
                 horizontalAlignment = Alignment.CenterHorizontally,
                 modifier =
                     Modifier.padding(contentPadding)
-                        .padding(all = 16.dp)
+                        .padding(horizontal = if (hasToolbar) 16.dp else 0.dp, vertical = 16.dp)
                         .fillMaxSize()
                         .verticalScroll(rememberScrollState()),
             ) {
@@ -309,7 +309,9 @@ internal fun VerifyDepositScreen(
             Column(
                 horizontalAlignment = Alignment.CenterHorizontally,
                 verticalArrangement = Arrangement.spacedBy(12.dp),
-                modifier = Modifier.fillMaxWidth().padding(horizontal = 24.dp, vertical = 12.dp),
+                modifier =
+                    Modifier.fillMaxWidth()
+                        .padding(horizontal = if (hasToolbar) 24.dp else 8.dp, vertical = 12.dp),
             ) {
                 if (state.hasFastSign) {
                     Text(


johnnyluo:
initial device: 
Screenshot_20260526_193801

join device: 

Screenshot_20260526_194359

rkokhatskyi (approved):
LGTM

PR #4648, opened 2026-05-26
fix(broadcast): retry hash-verify up to 3× with 2s backoff to ride out indexer lag

### LINKED ISSUE #4575: Co-sign broadcast recovery is single-shot; add hash-verify retries to ride out Blockchair indexer lag
## Summary

[BroadcastTxUseCase.recoverIfAlreadyBroadcast](https://github.com/vultisig/vultisig-android/blob/main/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt) does the hash-verify exactly once. When the broadcast call throws and the indexer lookup happens to land before the peer's broadcast has propagated to that indexer, the verify returns false and we surface the original "Transaction failed" error even though the tx is fine and will appear seconds later.

iOS handles this by retrying `isAlreadyOnChain` up to three times with a 2s backoff (`VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift`, `handleBroadcastError`). Android stays single-shot today.

This is mainly visible on UTXO chains routed through Blockchair (Bitcoin, Bitcoin Cash, Litecoin, Dogecoin, Dash, Zcash) after #4573, but the same path covers Solana, Polkadot, Sui, Ton, Ripple, Tron and Cardano.

## Proposed change

Add a small retry loop inside `recoverIfAlreadyBroadcast` (or pass an attempts/backoff parameter) so the verifier gets the same chance to find the tx as iOS does. Keep the rest of the contract identical: only retry on negative verify, never retry the broadcast itself, always rethrow on `CancellationException`.

## Acceptance

- Retry the verify N times with a small backoff when the first lookup returns false.
- Choose N / interval to match iOS unless we have a reason to differ.
- Existing co-sign happy and failure paths still behave the same.
- Unit tests cover the retry-then-success and retry-then-give-up branches.

## Notes

- Surfaced as the only remaining gap during #4573 (#4566 / #4567 fix). Field reports of false-failed BCH/UTXO sends after that PR ships would point here.
- iOS reference: `isAlreadyOnChain` in `KeysignViewModel.swift` (3 attempts, 2s).
- Windows reference: `verifyBroadcastByHash` is also single-shot but pairs it with explicit "already known" / "txn-mempool-conflict" string matching before the hash check; that path is a separate option if we don't want retries.

johnnyluo:
## Summary
- `recoverIfAlreadyBroadcast` was single-shot on the verify step; if the indexer hadn't propagated the tx yet, a legitimate broadcast was surfaced as \"Transaction failed\"
- Extract `isAlreadyOnChain` helper that retries the verify up to **3 times with a 2-second backoff** — matching iOS (`KeysignViewModel.swift`, `handleBroadcastError`)
- Only the verify is retried; the broadcast itself is never retried; `CancellationException` still propagates immediately
- Applies to all chains that use `recoverIfAlreadyBroadcast`: Bitcoin, BCH, Litecoin, Dogecoin, Dash, Zcash, Solana, Polkadot, Sui, TON, Ripple, Tron, Cardano

## Issue
Closes #4575

## Test Plan
- [x] New test: retry-then-success (first verify returns `Error`, second returns a result → returns hash)
- [x] New test: retry-then-give-up (all 3 verifies return `Error` → rethrows original broadcast exception)
- [x] All existing `BroadcastTxUseCaseTest` cases still pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction broadcast recovery: verification now retries with backoff and multiple attempts to confirm whether a failed broadcast is already on-chain.
  * Sui transactions no longer attempt recovery and are always broadcast directly to avoid unreliable digest-based recovery.

* **Tests**
  * Added tests covering recovery on a later verify attempt and giving up after all retries, plus stricter verification of call counts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4648?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
@@ -157,14 +157,9 @@ constructor(
                 bittensorApi.broadcastTransaction(tx.rawTransaction).orKnownHash(tx)
             }
 
-            Sui ->
-                recoverIfAlreadyBroadcast(
-                    tx = tx,
-                    broadcast = {
-                        suiApi.executeTransactionBlock(tx.rawTransaction, tx.signature ?: "")
-                    },
-                    verify = { hash -> suiApi.checkStatus(hash)?.digest?.isNotBlank() == true },
-                )
+            // Sui digest is not pre-computable from the raw transaction, so transactionHash
+            // is always blank and recovery cannot work; broadcast directly.
+            Sui -> suiApi.executeTransactionBlock(tx.rawTransaction, tx.signature ?: "")
 
             Ton ->
                 recoverIfAlreadyBroadcast(
@@ -224,6 +219,10 @@ constructor(
             }
         }
 
+    // Note: if the underlying HTTP client has its own retry-on-exception policy, each outer
+    // attempt here may itself fire multiple HTTP requests (network errors), while our loop
+    // handles logical "not yet indexed" responses (which return a valid non-error HTTP status).
+    // These are distinct failure modes, so the two retry layers do not simply multiply.
     private suspend fun isAlreadyOnChain(
         hash: String,
         verify: suspend (String) -> Boolean,
@@ -234,8 +233,14 @@ constructor(
                 if (verify(hash)) return true
             } catch (e: CancellationException) {
                 throw e
-            } catch (_: Exception) {
-                // treat verify failure as "not on chain yet"; continue to next retry
+            } catch (e: Exception) {
+                Timber.w(
+                    e,
+                    "tx %s verify attempt %d/%d failed; retrying",
+                    hash,
+                    attempt + 1,
+                    VERIFY_ATTEMPTS,
+                )
             }
         }
         return false
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
@@ -50,6 +50,7 @@ import com.vultisig.wallet.data.models.Chain.ZkSync
 import com.vultisig.wallet.data.models.SignedTransactionResult
 import javax.inject.Inject
 import kotlin.coroutines.cancellation.CancellationException
+import kotlinx.coroutines.delay
 import timber.log.Timber
 
 fun interface BroadcastTxUseCase {
@@ -212,9 +213,7 @@ constructor(
             throw e
         } catch (e: Exception) {
             val hash = tx.transactionHash
-            val alreadyOnChain =
-                hash.isNotBlank() && runCatching { verify(hash) }.getOrDefault(false)
-            if (alreadyOnChain) {
+            if (hash.isNotBlank() && isAlreadyOnChain(hash, verify)) {
                 Timber.d(
                     "broadcast failed but tx %s is already on-chain; treating as success",
                     hash,
@@ -225,6 +224,22 @@ constructor(
             }
         }
 
+    private suspend fun isAlreadyOnChain(
+        hash: String,
+        verify: suspend (String) -> Boolean,
+    ): Boolean {
+        repeat(VERIFY_ATTEMPTS) { attempt ->
+            if (attempt > 0) delay(VERIFY_BACKOFF_MS)
+            if (runCatching { verify(hash) }.getOrDefault(false)) return true
+        }
+        return false
+    }
+
     private fun String?.orKnownHash(tx: SignedTransactionResult): String? =
         this ?: tx.transactionHash.takeIf { it.isNotBlank() }
+
+    private companion object {
+        const val VERIFY_ATTEMPTS = 3
+        const val VERIFY_BACKOFF_MS = 2_000L
+    }
 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
@@ -219,10 +219,7 @@ constructor(
             }
         }
 
-    // Note: if the underlying HTTP client has its own retry-on-exception policy, each outer
-    // attempt here may itself fire multiple HTTP requests (network errors), while our loop
-    // handles logical "not yet indexed" responses (which return a valid non-error HTTP status).
-    // These are distinct failure modes, so the two retry layers do not simply multiply.
+    // HTTP-client retry and this loop handle distinct failure modes, so they don't multiply.
     private suspend fun isAlreadyOnChain(
         hash: String,
         verify: suspend (String) -> Boolean,
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
@@ -230,7 +230,13 @@ constructor(
     ): Boolean {
         repeat(VERIFY_ATTEMPTS) { attempt ->
             if (attempt > 0) delay(VERIFY_BACKOFF_MS)
-            if (runCatching { verify(hash) }.getOrDefault(false)) return true
+            try {
+                if (verify(hash)) return true
+            } catch (e: CancellationException) {
+                throw e
+            } catch (_: Exception) {
+                // treat verify failure as "not on chain yet"; continue to next retry
+            }
         }
         return false
     }
--- a/data/src/test/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCaseTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCaseTest.kt
@@ -25,6 +25,7 @@ import io.mockk.coEvery
 import io.mockk.coVerify
 import io.mockk.every
 import io.mockk.mockk
+import kotlin.coroutines.cancellation.CancellationException
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
@@ -155,6 +156,9 @@ class BroadcastTxUseCaseTest {
         coVerify(exactly = 1) {
             blockChairApi.broadcastTransaction(Chain.BitcoinCash, RAW_TRANSACTION)
         }
+        coVerify(exactly = 2) {
+            blockChairApi.getTsStatus(Chain.BitcoinCash, KNOWN_TRANSACTION_HASH)
+        }
     }
 
     @Test
@@ -175,6 +179,22 @@ class BroadcastTxUseCaseTest {
             blockChairApi.broadcastTransaction(Chain.BitcoinCash, RAW_TRANSACTION)
         }
         coVerify(exactly = 3) {
+            blockChairApi.getTsStatus(Chain.BitcoinCash, KNOWN_TRANSACTION_HASH)
+        }
+    }
+
+    @Test
+    fun `verify cancellation propagates immediately without further retries`() = runTest {
+        val blockChairApi = mockk<BlockChairApi>()
+        coEvery { blockChairApi.broadcastTransaction(Chain.BitcoinCash, RAW_TRANSACTION) } throws
+            RuntimeException("transaction already known")
+        coEvery { blockChairApi.getTsStatus(Chain.BitcoinCash, KNOWN_TRANSACTION_HASH) } throws
+            CancellationException("cancelled")
+
+        assertFailsWith<CancellationException> {
+            createUseCase(blockChairApi = blockChairApi)(Chain.BitcoinCash, signedTransaction())
+        }
+        coVerify(exactly = 1) {
             blockChairApi.getTsStatus(Chain.BitcoinCash, KNOWN_TRANSACTION_HASH)
         }
     }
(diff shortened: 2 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -225,6 +224,28 @@ constructor(
...
         }
 
+    private suspend fun isAlreadyOnChain(
+        hash: String,
+        verify: suspend (String) -> Boolean,
+    ): Boolean {
+        repeat(VERIFY_ATTEMPTS) { attempt ->
+            if (attempt > 0) delay(VERIFY_BACKOFF_MS)
+            try {
+                if (verify(hash)) return true
```
rkokhatskyi, line 230: The shared `HttpClient` installs `HttpRequestRetry` with `retryOnException(maxRetries = 3)` plus a `retryIf` that retries `GET` on `429` and `5xx`, so this three attempt loop multiplies those inner retries and a single failed broadcast can fire `getTsStatus` up to 12 times against an indexer that is already signalling overload, which is the very condition this recovery is reacting to

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -212,9 +213,7 @@ constructor(
             throw e
         } catch (e: Exception) {
```
rkokhatskyi, line 209: This catches every `Exception` and unconditionally enters the three attempt backoff with no gating on the error type, whereas the iOS path it cites returns immediately on a conclusive failed status and short circuits known duplicate errors like `BadInputsUTxO` before any lookup, so a clearly permanent failure such as an insufficient fee now waits about four seconds before it surfaces

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -212,9 +213,7 @@ constructor(
...
             val hash = tx.transactionHash
-            val alreadyOnChain =
-                hash.isNotBlank() && runCatching { verify(hash) }.getOrDefault(false)
-            if (alreadyOnChain) {
+            if (hash.isNotBlank() && isAlreadyOnChain(hash, verify)) {
```
rkokhatskyi, line 211: `SuiHelper.getSignedTransaction` builds `SignedTransactionResult(output.unsignedTx, "", output.signature)` so `transactionHash` is blank for Sui, which makes this `hash.isNotBlank()` guard always false so `isAlreadyOnChain` never runs for Sui even though the description lists it among the covered chains

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -225,6 +224,28 @@ constructor(
             }
         }
 
+    private suspend fun isAlreadyOnChain(
```
rkokhatskyi, line 223: For non initiating devices `broadcastOrRecover` already falls back to `recoverJoinedDeviceBroadcast`, which returns `signedTx.transactionHash` on any failure and that is the same value this helper yields on success, so a joined device waits the full backoff here for a hash the view model layer would have returned right away

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -225,6 +224,28 @@ constructor(
...
+        hash: String,
+        verify: suspend (String) -> Boolean,
+    ): Boolean {
+        repeat(VERIFY_ATTEMPTS) { attempt ->
+            if (attempt > 0) delay(VERIFY_BACKOFF_MS)
+            try {
+                if (verify(hash)) return true
+            } catch (e: CancellationException) {
+                throw e
+            } catch (_: Exception) {
```
rkokhatskyi: `catch (_: Exception)` drops every verify failure with no log at all, so when the indexer lag this change targets actually happens nothing in `Timber` separates a genuine miss from a verify endpoint that is erroring, whereas the iOS equivalent logs each failed attempt

data/src/test/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCaseTest.kt
```diff
@@ -132,6 +133,42 @@ class BroadcastTxUseCaseTest {
...
+            broadcastError
+        coEvery { blockChairApi.getTsStatus(Chain.BitcoinCash, KNOWN_TRANSACTION_HASH) } returns
+            BlockChainStatusDeserialized.Error("not indexed")
+
+        val thrown =
+            assertFailsWith<RuntimeException> {
+                createUseCase(blockChairApi = blockChairApi)(Chain.BitcoinCash, signedTransaction())
+            }
+        assertEquals(broadcastError, thrown)
+        coVerify(exactly = 3) {
```
rkokhatskyi, line 181: None of the recovery tests assert `broadcastTransaction` runs exactly once, so the never rebroadcast guarantee this change leans on is unguarded against a regression, unlike `KeysignViewModelBroadcastRecoveryTest` which pins it with `coVerify(exactly = 1)`
```
coVerify(exactly = 1) { blockChairApi.broadcastTransaction(Chain.BitcoinCash, RAW_TRANSACTION) }
```

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
johnnyluo: Fixed: removed Sui from `recoverIfAlreadyBroadcast` since `SuiHelper.getSignedTransaction` always produces a blank `transactionHash`, making recovery dead code. Sui now uses a direct `executeTransactionBlock` call.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
johnnyluo: Fixed: replaced `catch (_: Exception)` with `catch (e: Exception)` and added `Timber.w(e, "tx %s verify attempt %d/%d failed; retrying", hash, attempt + 1, VERIFY_ATTEMPTS)` so each failed verify attempt appears in Timber output.

data/src/test/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCaseTest.kt
johnnyluo: Fixed: added `coVerify(exactly = 1) { blockChairApi.broadcastTransaction(...) }` to all three recovery tests (recovers on first verify, recovers on second attempt, gives up after all attempts), matching the `KeysignViewModelBroadcastRecoveryTest` pattern.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
johnnyluo: Added a comment in the code clarifying the distinction: HTTP-level retries fire on *network exceptions* (connection refused, timeout) before the call even returns; our outer loop retries on logical "not yet indexed" responses, which are successful HTTP responses (200 OK) with an error payload. In the typical indexer-lag scenario the verify endpoint responds with 200+error, so there is no HTTP retry multiplication. Genuine network failures can stack, but that is an existing property of the shared `HttpClient` and not introduced by this change.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
johnnyluo: Acknowledged — for UTXO chains where the hash is pre-computed, a clearly permanent error like "insufficient fee" will enter the 3-attempt backoff (~4s) before surfacing. Gating on error-message patterns (like iOS's `BadInputsUTxO` short-circuit) is fragile across chains, so I'd prefer to leave that for a follow-up where we can audit all broadcast error formats per chain rather than widening this PR's scope.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
johnnyluo: Acknowledged — for joined devices, `BroadcastTxUseCase` has no awareness of the initiating/joined split, so it will run the full backoff before re-throwing, after which `KeysignViewModel.broadcastOrRecover` immediately recovers via `recoverJoinedDeviceBroadcast`. The ~4s delay is a known limitation. Fixing this cleanly would require threading `isInitiatingDevice` into `BroadcastTxUseCase`, which is a separate refactor.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -225,6 +219,38 @@ constructor(
             }
         }
 
+    // Note: if the underlying HTTP client has its own retry-on-exception policy, each outer
+    // attempt here may itself fire multiple HTTP requests (network errors), while our loop
+    // handles logical "not yet indexed" responses (which return a valid non-error HTTP status).
+    // These are distinct failure modes, so the two retry layers do not simply multiply.
```
rkokhatskyi: The `retryIf` in `HttpClientConfigurator` also retries `GET` on `429` and `5xx`, not just network exceptions, and `getTsStatus` returns `null` on those statuses rather than throwing, so under indexer rate limiting the inner retries and this outer loop both fire and one recovery fans out to roughly 12 requests, which is the multiplication this note rules out

rkokhatskyi (approved):

aminsato (commented):
A few inline notes from a local review — all non-blocking. The core fix (retry verify, never retry broadcast) looks correct and the Sui simplification checks out.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
aminsato: This comment is longer than the point needs. Trim to a sentence: the HTTP-client retry and this loop handle distinct failure modes, so they don't multiply.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -225,6 +219,38 @@ constructor(
...
+                    e,
+                    "tx %s verify attempt %d/%d failed; retrying",
+                    hash,
+                    attempt + 1,
+                    VERIFY_ATTEMPTS,
+                )
+            }
+        }
+        return false
+    }
```
aminsato, line 244: Genuine broadcast failures (e.g. insufficient fee) now take ~4s longer to surface, since all 3 verify attempts run with the 2s backoff before the original error is rethrown. The happy path is unaffected (no exception → no verify call), so this only costs latency on the failure path.

data/src/test/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCaseTest.kt
```diff
@@ -130,6 +134,49 @@ class BroadcastTxUseCaseTest {
...
+    @Test
+    fun `bitcoin cash broadcast recovers on second verify attempt after indexer lag`() = runTest {
+        val blockChairApi = mockk<BlockChairApi>()
+        coEvery { blockChairApi.broadcastTransaction(Chain.BitcoinCash, RAW_TRANSACTION) } throws
+            RuntimeException("fail to broadcast transaction: transaction already known")
+        coEvery { blockChairApi.getTsStatus(Chain.BitcoinCash, KNOWN_TRANSACTION_HASH) } returnsMany
+            listOf(
+                BlockChainStatusDeserialized.Error("not yet indexed"),
+                blockchairResult(blockId = -1),
+            )
```
aminsato, line 150: Pin the retry count here too: `coVerify(exactly = 2) { blockChairApi.getTsStatus(Chain.BitcoinCash, KNOWN_TRANSACTION_HASH) }`. The test currently passes only because `returnsMany` forces a second call, but the count isn't asserted — mirroring the `exactly = 3` check in the give-up test makes the retry-actually-fired intent explicit and catches a regression where the loop stops early.

data/src/test/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCaseTest.kt
```diff
@@ -130,6 +134,49 @@ class BroadcastTxUseCaseTest {
...
+            assertFailsWith<RuntimeException> {
+                createUseCase(blockChairApi = blockChairApi)(Chain.BitcoinCash, signedTransaction())
+            }
+        assertEquals(broadcastError, thrown)
+        coVerify(exactly = 1) {
+            blockChairApi.broadcastTransaction(Chain.BitcoinCash, RAW_TRANSACTION)
+        }
+        coVerify(exactly = 3) {
+            blockChairApi.getTsStatus(Chain.BitcoinCash, KNOWN_TRANSACTION_HASH)
+        }
```
aminsato, line 183: Add a test covering cancellation: a `verify` that throws `CancellationException` should propagate immediately without further retries. The production code handles this via `catch (e: CancellationException) { throw e }` in `isAlreadyOnChain`, but it's currently untested — stub `getTsStatus` to throw `CancellationException` and assert it's rethrown.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
johnnyluo: Acknowledged — getTsStatus returns null on 429/5xx rather than throwing, so the HTTP retryIf doesn't fire for those responses, and the two loops don't multiply in practice. Comment trimmed per aminsato's request.

PR #4649, opened 2026-05-26
fix(swap): clear loading state when quote times out

### LINKED ISSUE #4583: Swap: BCH→LTC THORChain quote timeout leaves screen stuck loading forever
## Description

When requesting a swap quote for **BCH → LTC** via THORChain, the quote request times out after 15s, but the Swap screen remains stuck in a loading state indefinitely — the loading spinner never resolves, the Provider / Total Fee rows stay as placeholder skeletons, and the CTA stays disabled ("Fill in amount"). There is no visible error, retry, or fallback shown to the user.

## Steps to reproduce

1. Open the app and navigate to **Swap**
2. Set **From** = Bitcoin-Cash (BCH), **To** = Litecoin (LTC)
3. Enter an amount (e.g. `0.020710`)
4. Wait for the quote to load

## Expected

Either the quote loads, or — if the upstream provider times out — the loading state ends and a user-facing error / retry is shown. The screen should not be stuck loading.

## Actual

The spinner persists indefinitely. Provider and Total Fee remain as placeholder shimmer rows and the user cannot proceed.

## Logcat

```
REQUEST https://gateway.liquify.com/chain/thorchain_api/thorchain/quote/swap?from_asset=BCH.BCH&to_asset=LTC.LTC&amount=2071000&destination=<redacted>&streaming_interval=0&affiliate=va&affiliate_bps=50 failed with exception: java.util.concurrent.CancellationException: Timed out waiting for 15000 ms

Quote fetch failed provider=THORCHAIN src=BCH-Bitcoin-Cash dst=LTC-Litecoin amount=2071000

kotlinx.coroutines.TimeoutCancellationException: Timed out waiting for 15000 ms
    at kotlinx.coroutines.TimeoutKt.TimeoutCancellationException(Timeout.kt:189)
    at kotlinx.coroutines.TimeoutCoroutine.run(Timeout.kt:157)
    at android.os.Handler.handleCallback(Handler.java:942)
    ...
```

The timeout exception is logged, but the loading UI state is never cleared.

## Evidence

swap-stuck-loading

johnnyluo:
## Summary

- When all swap quote providers time out, `fetchBestQuote` rethrows the `TimeoutCancellationException` collected from the inner `withTimeout`
- `TimeoutCancellationException` is a `CancellationException`, so the catch block in `calculateFees()` re-throws it instead of routing it to `resetQuoteState()`
- This leaves `isLoading = true` forever — the screen shows a loading spinner indefinitely with no error, retry option, or way to proceed
- Fix: convert `TimeoutCancellationException` → `SwapException.TimeOut` in `fetchBestQuote` so the `catch (e: SwapException)` handler in `calculateFees()` can clear the loading state and show "Swap request timed out. Please try again"

## Issue

Closes #4583

## Test Plan

- [ ] Set From = BCH, To = LTC, enter an amount and wait for quote to load — screen should now show a timeout error message instead of spinning forever
- [ ] Confirm other swap pairs (ETH → USDC etc.) still load quotes normally

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error selection when all swap quote attempts fail: prioritizes specific swap errors and converts pure timeouts into a clear timeout error.

* **Tests**
  * Added a test to verify that multiple timeout failures are reported as a single timeout error.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4649?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -290,7 +290,17 @@ constructor(
         val successes = results.mapNotNull { it.getOrNull() }
         if (successes.isEmpty()) {
             val failures = results.mapNotNull { it.exceptionOrNull() }
-            throw failures.firstOrNull { it is SwapException } ?: failures.first()
+            val firstSwapException = failures.firstOrNull { it is SwapException }
+            if (firstSwapException != null) throw firstSwapException
+            val firstFailure = failures.first()
+            // TimeoutCancellationException is a CancellationException — re-throwing it would
+            // propagate coroutine cancellation and leave the UI stuck in a loading state.
+            // Convert it to a SwapException so the error handler in calculateFees() can
+            // clear the loading state and show a user-facing message.
+            if (firstFailure is TimeoutCancellationException) {
+                throw SwapException.TimeOut(firstFailure.message ?: "Quote fetch timed out")
+            }
+            throw firstFailure
         }
 
         // Rank on estimatedDstFiat alone — this represents the destination amount
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -290,17 +290,11 @@ constructor(
         val successes = results.mapNotNull { it.getOrNull() }
         if (successes.isEmpty()) {
             val failures = results.mapNotNull { it.exceptionOrNull() }
-            val firstSwapException = failures.firstOrNull { it is SwapException }
-            if (firstSwapException != null) throw firstSwapException
-            val firstFailure = failures.first()
-            // TimeoutCancellationException is a CancellationException — re-throwing it would
-            // propagate coroutine cancellation and leave the UI stuck in a loading state.
-            // Convert it to a SwapException so the error handler in calculateFees() can
-            // clear the loading state and show a user-facing message.
-            if (firstFailure is TimeoutCancellationException) {
-                throw SwapException.TimeOut(firstFailure.message ?: "Quote fetch timed out")
-            }
-            throw firstFailure
+            failures.firstOrNull { it is SwapException }?.let { throw it }
+            failures
+                .firstOrNull { it is TimeoutCancellationException }
+                ?.let { throw SwapException.TimeOut(it.message.orEmpty()) }
+            throw failures.first()
         }
 
         // Rank on estimatedDstFiat alone — this represents the destination amount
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -290,7 +290,17 @@ constructor(
...
             val failures = results.mapNotNull { it.exceptionOrNull() }
-            throw failures.firstOrNull { it is SwapException } ?: failures.first()
+            val firstSwapException = failures.firstOrNull { it is SwapException }
+            if (firstSwapException != null) throw firstSwapException
+            val firstFailure = failures.first()
+            // TimeoutCancellationException is a CancellationException — re-throwing it would
+            // propagate coroutine cancellation and leave the UI stuck in a loading state.
+            // Convert it to a SwapException so the error handler in calculateFees() can
+            // clear the loading state and show a user-facing message.
+            if (firstFailure is TimeoutCancellationException) {
```
rkokhatskyi: This converts to `SwapException.TimeOut` only when `failures.first()` happens to be the timeout, so when a provider fails with a raw non-`SwapException` error that lands first while another provider hits the `withTimeout`, the timeout is masked and surfaces through `calculateFees()` as the generic `swap_error_quote_failed`, which is the wrong-error case this change is meant to avoid. Scanning all failures for a timeout closes that gap
```kotlin
failures.firstOrNull { it is SwapException }?.let { throw it }
failures.firstOrNull { it is TimeoutCancellationException }
    ?.let { throw SwapException.TimeOut(it.message ?: "Quote fetch timed out") }
throw failures.first()
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -290,7 +290,17 @@ constructor(
...
+                throw SwapException.TimeOut(firstFailure.message ?: "Quote fetch timed out")
```
rkokhatskyi: The message passed here is never shown to the user since `mapSwapExceptionToFormError` maps `SwapException.TimeOut` to the static `R.string.swap_error_time_out` and ignores `e.message`, so the hardcoded English `"Quote fetch timed out"` fallback is dead and unlocalized and could be dropped to `firstFailure.message.orEmpty()`

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -290,7 +290,17 @@ constructor(
         val successes = results.mapNotNull { it.getOrNull() }
         if (successes.isEmpty()) {
             val failures = results.mapNotNull { it.exceptionOrNull() }
-            throw failures.firstOrNull { it is SwapException } ?: failures.first()
+            val firstSwapException = failures.firstOrNull { it is SwapException }
```
rkokhatskyi: The conversion this PR adds has no test since `SwapFormViewModelTest` mocks `swapQuoteManager` and the existing timeout case at line 1048 throws an already-built `SwapException.TimeOut`, so nothing exercises `fetchBestQuote` turning an all-providers `TimeoutCancellationException` into `SwapException.TimeOut`, whereas `fetchBestQuote` is `internal` and can be tested directly by stubbing the repository to time out every candidate

rkokhatskyi (changes requested):

rkokhatskyi (approved):
LGTM

PR #4651, opened 2026-05-26
perf(balances): stream per-chain balances, cached snapshot first

### LINKED ISSUE #4562: Perf: Solana (and all chains) first-fetch slow — balances are batched, not streamed per chain
## Summary

After adding a Solana account to a vault, the SOL balance row stays empty for noticeably longer on Android than on iOS / Windows. User-reported as "Solana fetching after chain addition is very slow in comparison to other platforms."

The actual native SOL balance call is a single RPC against the same backend all three clients use (`https://api.vultisig.com/solana/`) — so the slowness is **not** in the Solana RPC itself. It's in how Android aggregates and emits the result.

## Root cause

`data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt:72-122`

```kotlin
override fun loadAddresses(vaultId: String, isRefresh: Boolean): Flow<List<Address>> =
    buildCacheAddresses(vaultId).flatMapLatest { (vaultCoins, addresses) ->
        channelFlow {
            supervisorScope {
                val loadPrices = async { tokenPriceRepository.refresh(vaultCoins) }

                if (!isRefresh) {
                    addresses.fetchAccountFromDb()
                    send(addresses)
                }

                addresses
                    .mapIndexed { index, account ->
                        async {
                            ...
                            loadPrices.await()           // <-- (B) every chain waits for ALL prices
                            val newAccounts = ...
                            addresses[index] = account.copy(accounts = newAccounts)
                            // (A) no send() here — UI doesn't see this chain's balance yet
                        }
                    }
                    .awaitAll()
                send(addresses)                          // <-- (A) ONE emit after every chain completes
            }
            awaitClose()
        }
    }
```

### A. Balances are batched, not streamed per chain

The only `send(addresses)` after the parallel fetch happens *after* `awaitAll()` — so the user can't see the new Solana balance until **every other chain on the vault** has also finished its balance fetch. On a vault with many chains (e.g. all EVM + BTC family + Cosmos + Solana + Sui + Ton + ...), the slowest chain dictates time-to-first-pixel for SOL.

`isRefresh = true` (which is the case immediately after adding a chain — `REFRESH_CHAIN_DATA` event fires in `ChainSelectionViewModel.onCommitChanges`) also skips the early DB-cache `send(addresses)` at the top, so the UI doesn't even see the placeholder row early.

### B. All balances gated on price refresh

Each async balance block does `loadPrices.await()` before fetching its balance. If `tokenPriceRepository.refresh(vaultCoins)` is slow (it batches every coin across every enabled chain), the Solana balance call doesn't even start until prices return.

### C. Solana-specific: SPL discovery + no ATA cache

When SPL tokens are eventually fetched (`SPLTokenRepository.fetchTokens`, `data/src/main/kotlin/com/vultisig/wallet/data/repositories/SPLTokenRepository.kt:52-61`), the path is:

1. `solanaApi.getSPLTokens(address)` — `getTokenAccountsByOwner` RPC
2. `solanaApi.getSPLTokensInfo(allMints)` — POST to `api.solana.fm/v1/tokens`
3. `solanaApi.getSPLTokensInfo2(missingMints)` — Jupiter `tokens/v2/search`, one HTTP request per missing token in parallel

iOS additionally has a **30-day token-account address cache** (`SolanaService.tokenAccountCache`, `VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaService.swift:51-52`) — Android has no equivalent, so every refresh re-resolves ATAs that haven't changed.

## Suggested fixes (in priority order)

1. **Stream balances per chain.** Move `send(addresses.toList())` inside each `async { ... }` block right after `addresses[index] = account.copy(...)`. UI then updates as each chain returns, like iOS / Windows. (Single-file change, biggest visible win.)
2. **Decouple balance from price.** Render the token balance immediately on arrival; update the fiat column when price arrives separately. Today a slow price call blocks the entire row.
3. **Always emit early DB-cached snapshot.** Drop the `if (!isRefresh)` gate before the first `send(addresses)`. Even on refresh, showing the last known balance immediately is better than empty rows.
4. **Add a token-account cache (Solana-specific).** Persist `getTokenAccountsByOwner` results keyed by `(address, mint)` for ~30 days to match iOS — saves a round-trip on every SPL refresh.
5. (Optional) **Warm the HTTPS connection to `api.vultisig.com`** at app launch so the first Solana RPC doesn't pay TLS handshake cost.

## iOS reference

- `VultisigApp/VultisigApp/Core/Services/BalanceService.swift:101-127` — `withTaskGroup` fans out per-coin balance fetches.
- `VultisigApp/VultisigApp/Blockchain/Solana/Service/SolanaService.swift:51-52` — 30-day `tokenAccountCache` for SPL token accounts.

## Repro

1. Vault with several chains already enabled.
2. Add a new chain (e.g. Solana) via the Add Chain flow.
3. Return to the wallet view. Observe SOL row sits with no balance for several seconds, longer than on iOS / Windows side-by-side.

## Platform

Android only. iOS / Windows already render per-chain balances as they arrive.

aminsato:
## Summary

Fixes the slow first-fetch reported in #4562 (Solana balance row stays empty far longer on Android than iOS/Windows).

`AccountsRepository.loadAddresses` batched every chain's balance into a **single** emit after `awaitAll()`, so a fast chain stayed empty until the **slowest** chain on the vault returned. The early DB-cache emit was also gated behind `if (!isRefresh)`, so right after adding a chain (`isRefresh = true`) the UI showed empty rows instead of the last-known balances.

## Changes

- **Stream per chain** — emit each chain's balance as soon as it resolves, instead of one emit after all chains finish. Uses a `toList()` snapshot since chains mutate the backing list in parallel. (issue priority #1)
- **Always emit the cached DB snapshot first** — dropped the `if (!isRefresh)` gate, so the last-known balances render immediately, including on refresh. (issue priority #3)
- **Removed the now-dead `isRefresh` parameter** from the plural `loadAddresses` and its call sites (`VaultAccountsViewModel`, `AccountsLoader`). It still lives on `loadDeFiAddresses`/`loadAddress`, which genuinely use it.
- Added `AccountsRepositoryImplTest` (cached-first + per-chain streaming).

## Deferred follow-ups (called out in the issue)

Not in this PR — larger/riskier, warrant separate review:
- #2 decouple balance from the global price refresh (needs balance-then-fiat re-emit)
- #4 Solana 30-day token-account/ATA cache
- #5 TLS pre-warm to `api.vultisig.com`

## Test plan

- [x] `:data:testDebugUnitTest` — new `AccountsRepositoryImplTest` (2 tests) green
- [x] `:app:testDebugUnitTest` — `VaultAccountsViewModelTest`, `AccountsLoaderTest` green
- [x] ktfmt applied
- [ ] Manual: add Solana to a multi-chain vault, confirm SOL balance appears as soon as its RPC returns rather than after every chain

Closes #4562

## Summary by CodeRabbit

* **Refactor**
  * Account loading now surfaces cached accounts immediately and progressively updates per-chain balances, keeping the refresh spinner active until final completion.
* **New Features**
  * Added a throttling operator to coalesce rapid balance updates while guaranteeing final updates are delivered.
* **Tests**
  * Extended tests covering progressive balance streaming, spinner behavior during refresh, and throttling semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4651?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -48,8 +48,10 @@ import com.vultisig.wallet.ui.screens.settings.bottomsheets.notifications.VaultI
 import com.vultisig.wallet.ui.utils.SnackbarFlow
 import com.vultisig.wallet.ui.utils.pushNotificationErrorUiText
 import com.vultisig.wallet.ui.utils.textAsFlow
+import com.vultisig.wallet.ui.utils.throttleLatest
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
+import kotlin.time.Duration.Companion.milliseconds
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Job
@@ -405,8 +407,15 @@ constructor(
             viewModelScope.safeLaunch {
                 combine(
                         accountsRepository
-                            .loadAddresses(vaultId)
-                            .map { it.sortByAccountsTotalFiatValue() }
+                            .loadAddressBalances(vaultId)
+                            // Leading debounce: coalesce the burst of per-chain updates so the rows
+                            // settle once per window instead of reordering on every chain arrival
+                            // (matches the leading debounce iOS added in #4337).
+                            .throttleLatest(BALANCE_RENDER_WINDOW) { it.isComplete }
+                            // Keep the pull-to-refresh spinner up until every chain has resolved,
+                            // matching iOS/Windows, instead of clearing it on the cached snapshot.
+                            .onEach { if (it.isComplete) updateRefreshing(false) }
+                            .map { it.addresses.sortByAccountsTotalFiatValue() }
                             .catch {
                                 updateRefreshing(false)
                                 Timber.e(it)
@@ -498,7 +507,6 @@ constructor(
                 )
             }
         }
-        updateRefreshing(false)
 
         Timber.d("Update updateUiStateFromList: %s", "$this")
     }
@@ -704,5 +712,9 @@ constructor(
 
     companion object {
         internal const val REFRESH_CHAIN_DATA = "refresh_chain_data"
+
+        // Window for coalescing per-chain balance updates so rows settle once per window instead
+        // of reordering on every chain arrival; matches the leading debounce iOS uses (#4337).
+        private val BALANCE_RENDER_WINDOW = 250.milliseconds
     }
 }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -282,7 +282,7 @@ constructor(
     fun refreshData() {
         val vaultId = vaultId ?: return
         updateRefreshing(true)
-        loadAccounts(vaultId, true)
+        loadAccounts(vaultId)
     }
 
     fun send() {
@@ -399,13 +399,13 @@ constructor(
             }
     }
 
-    private fun loadAccounts(vaultId: String, isRefresh: Boolean = false) {
+    private fun loadAccounts(vaultId: String) {
         loadAccountsJob?.cancel()
         loadAccountsJob =
             viewModelScope.safeLaunch {
                 combine(
                         accountsRepository
-                            .loadAddresses(vaultId, isRefresh)
+                            .loadAddresses(vaultId)
                             .map { it.sortByAccountsTotalFiatValue() }
                             .catch {
                                 updateRefreshing(false)
--- a/app/src/test/java/com/vultisig/wallet/ui/models/home/VaultAccountsViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/home/VaultAccountsViewModelTest.kt
@@ -15,6 +15,7 @@ import com.vultisig.wallet.data.models.FiatValue
 import com.vultisig.wallet.data.models.TokenValue
 import com.vultisig.wallet.data.models.Vault
 import com.vultisig.wallet.data.repositories.AccountsRepository
+import com.vultisig.wallet.data.repositories.AddressBalancesUpdate
 import com.vultisig.wallet.data.repositories.BalanceVisibilityRepository
 import com.vultisig.wallet.data.repositories.CryptoConnectionTypeRepository
 import com.vultisig.wallet.data.repositories.DefaultDeFiChainsRepository
@@ -50,6 +51,7 @@ import io.mockk.unmockkStatic
 import io.mockk.verify
 import java.math.BigDecimal
 import java.math.BigInteger
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.MutableStateFlow
@@ -212,26 +214,28 @@ internal class VaultAccountsViewModelTest {
             vm.uiState.value.cryptoConnectionType shouldBe CryptoConnectionType.Wallet
         }
 
-    /** Verifies refreshData re-invokes accountsRepository with isRefresh=true. */
+    /**
+     * Verifies refreshData re-triggers a balance reload via accountsRepository.loadAddressBalances.
+     */
     @Test
-    fun `refreshData re-invokes accountsRepository with isRefresh true`() =
+    fun `refreshData re-invokes loadAddressBalances`() =
         runTest(testDispatcher) {
             every { lastOpenedVaultRepository.lastOpenedVaultId } returns flowOf("vault-1")
             coEvery { vaultRepository.get("vault-1") } returns Vault(id = "vault-1", name = "Test")
 
             val vm = createViewModel()
             advanceUntilIdle()
-            // Drop any loadAddresses calls made during init; only count the refresh call.
+            // Drop any loadAddressBalances calls made during init; only count the refresh call.
             clearMocks(accountsRepository, answers = false)
 
             vm.refreshData()
             advanceUntilIdle()
 
-            verify(exactly = 1) { accountsRepository.loadAddresses("vault-1") }
+            verify(exactly = 1) { accountsRepository.loadAddressBalances("vault-1") }
         }
 
     /**
-     * Verifies refreshData drives `loadAddresses` to completion via the .catch block when the
+     * Verifies refreshData drives the balance load to completion via the .catch block when the
      * underlying flow throws, leaving `isRefreshing` cleared. The verify guards against a no-op
      * implementation passing the assertion vacuously (default `isRefreshing` is already false).
      */
@@ -240,7 +244,7 @@ internal class VaultAccountsViewModelTest {
         runTest(testDispatcher) {
             every { lastOpenedVaultRepository.lastOpenedVaultId } returns flowOf("vault-1")
             coEvery { vaultRepository.get("vault-1") } returns Vault(id = "vault-1", name = "Test")
-            every { accountsRepository.loadAddresses("vault-1") } returns
+            every { accountsRepository.loadAddressBalances("vault-1") } returns
                 flow { throw RuntimeException("Balance load failed") }
 
             val vm = createViewModel()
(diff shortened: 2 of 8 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -399,13 +399,13 @@ constructor(
...
 
-    private fun loadAccounts(vaultId: String, isRefresh: Boolean = false) {
+    private fun loadAccounts(vaultId: String) {
         loadAccountsJob?.cancel()
         loadAccountsJob =
             viewModelScope.safeLaunch {
                 combine(
                         accountsRepository
-                            .loadAddresses(vaultId, isRefresh)
+                            .loadAddresses(vaultId)
```
rkokhatskyi: `loadAddresses(vaultId)` now emits once per chain as each balance resolves and the following `map { sortByAccountsTotalFiatValue() }` reruns on every emission, so the rows reorder on screen as each chain arrives, which is the reorder flicker iOS removed with a 250ms leading debounce in `VaultDetailViewModel` (#4337) whereas this flow has no `debounce` or `distinctUntilChanged`

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -282,7 +282,7 @@ constructor(
     fun refreshData() {
         val vaultId = vaultId ?: return
         updateRefreshing(true)
-        loadAccounts(vaultId, true)
+        loadAccounts(vaultId)
```
rkokhatskyi, line 287: Since the cached snapshot is now always emitted first and `updateUiStateFromList` calls `updateRefreshing(false)` on every emission, the pull to refresh spinner clears on that first cached emission while the network balances are still loading, whereas before `isRefresh = true` skipped the cached emit so the spinner stayed up until the load finished, while iOS and Windows keep the indicator until the whole refresh completes

data/src/test/kotlin/com/vultisig/wallet/data/repositories/AccountsRepositoryImplTest.kt
```diff
@@ -0,0 +1,181 @@
...
+
+        val emissions = mutableListOf<List<Address>>()
+        val job = launch { repository.loadAddresses(VAULT_ID).collect(emissions::add) }
+
+        advanceUntilIdle()
+
+        // ETH's balance must surface while SOL is still pending — the whole point of per-chain
+        // streaming. Before the fix, no emission carried ETH's network value until SOL also
+        // finished.
+        val streamedEth = emissions.firstOrNull { it.value(Chain.Ethereum) == ETH_NETWORK }
```
rkokhatskyi, line 116: This assertion still passes when `loadAddresses` batches every balance into one emit, verified by reverting the implementation to the batched version and rerunning this test green, since `firstOrNull` only scans the accumulated emissions rather than proving ETH surfaced before SOL resolved, so asserting the emission count while `solanaGate` is still pending, one when batched versus two when streamed, would actually guard the behaviour

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -117,7 +123,7 @@ constructor(
                             }
                         }
                         .awaitAll()
-                    send(addresses)
+                    send(addresses.toList())
```
rkokhatskyi: Each chain already sends a full snapshot when it resolves and these collectors run on the Main dispatcher, so once `awaitAll` returns this final `send` just repeats the last emitted snapshot and costs an extra downstream sort and remap, so it looks redundant here

app/src/test/java/com/vultisig/wallet/ui/models/home/VaultAccountsViewModelTest.kt
```diff
@@ -227,7 +227,7 @@ internal class VaultAccountsViewModelTest {
             vm.refreshData()
             advanceUntilIdle()
 
-            verify(exactly = 1) { accountsRepository.loadAddresses("vault-1", true) }
+            verify(exactly = 1) { accountsRepository.loadAddresses("vault-1") }
```
rkokhatskyi: This `verify` no longer distinguishes a refresh from a normal load now that the parameter is gone while the method name and KDoc just above still reference `isRefresh`, and the behaviour this PR actually changed, the cached snapshot now rendering on refresh, has no test covering it

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -109,6 +110,11 @@ constructor(
                                     }
 
                                     addresses[index] = account.copy(accounts = newAccounts)
+                                    // Stream each chain's balance as soon as it resolves rather
+                                    // than waiting for every chain to finish, matching iOS/Windows.
```
rkokhatskyi: This comment is not accurate for iOS, which does not stream per chain but collects all balances through a `withTaskGroup` and applies them in a single `Storage.shared.save()` in `BalanceService.updateBalances`, and it added a leading debounce precisely because per chain reorders flicker (#4337), so only Windows streams the way described here

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Fixed in c0b76dba3. Added a 250ms leading throttle (`throttleLatest`) on the home accounts flow so the burst of per-chain emits coalesces and rows settle once per window instead of reordering on each arrival — matching the leading debounce iOS added in #4337. The terminal emission always passes through, so the fully-loaded state still renders. Covered by `ThrottleLatestTest`.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Fixed in c0b76dba3. Added `loadAddressBalances`, whose terminal emission carries an `isComplete` flag; the VM now clears `isRefreshing` only on completion (and on error), not on the cached snapshot — so the pull-to-refresh spinner stays up through the whole refresh like iOS/Windows. `updateUiStateFromList` no longer clears it. New test: `refresh spinner stays up until load completes`.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/AccountsRepositoryImplTest.kt
aminsato: Done in c0b76dba3 — added `assertEquals(2, emissions.size)` while SOL is still gated (cached + ETH-streamed), so the emission count pins the streaming behaviour as you suggested. Kept the ETH-value check alongside it.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Fixed in c0b76dba3. The post-`awaitAll` emission is now the terminal `isComplete = true` update used only for the refresh-completion signal, and `loadAddresses` filters it out (`filter { !it.isComplete }`) — so plain `List` consumers get exactly one render per change with no redundant final re-render. Added a terminal-drop test.

app/src/test/java/com/vultisig/wallet/ui/models/home/VaultAccountsViewModelTest.kt
aminsato: Fixed in c0b76dba3. Renamed the test to `refreshData re-invokes loadAddressBalances` and updated the KDoc to drop the stale `isRefresh` wording. Also added `refresh spinner stays up until load completes`, which covers the cached-snapshot-on-refresh behaviour that previously had no test.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Fixed in c0b76dba3. Reworded the comment — only Windows streams per chain; iOS batches all balances and applies them in a single save (with the #4337 leading debounce). Removed the inaccurate iOS reference.

aminsato:
Addressed in 856dc476f: `isRefreshing is cleared after error during refresh` now clears mocks after init (`clearMocks(accountsRepository, answers = false)`) and asserts `verify(exactly = 1) { loadAddressBalances("vault-1") }`, so the refresh path is exercised non-vacuously — matching the sibling `refreshData re-invokes loadAddressBalances` test.

@coderabbitai resolve

johnnyluo (approved):
LGTM

rkokhatskyi (approved):
LGTM

PR #4665, opened 2026-05-27
feat(swap): SwapKit BTC/PSBT quote+payload decode (part 1)

aminsato:
Part 1 of SwapKit **BTC/PSBT** support (non-EVM Phase 2, epic #4589), targeting the **`feature/swapkit-btc`** integration branch — `main` only gets BTC via the final integration PR once the feature is complete. This PR is the **quote/payload decode** half; the signer (#4666) and the `SwapProviderTable` BTC gate land on the same branch next.

## What this does (tested, inert in prod)
Teaches `SwapKitQuoteSource` to decode a BTC `PSBT` route instead of refusing it:
- `chainPrefix` maps `Bitcoin → "BTC"`; `nativeDecimals(Bitcoin) = 8` so the inbound BTC deposit fee scales correctly.
- New `TxKind.PSBT`: the `/v3/swap` response is decoded into a fully-formed `SwapQuote.SwapKit` (base64 PSBT → `SwapKitSwapPayloadJson.txPayload` bytes, `targetAddress`, `toAmountDecimal`, `subProvider`, `swapId`) and returned on `SwapQuoteResult.Native` — the seam a per-chain signer consumes.
- `SwapQuoteManager.fetchSwapKitQuote` handles the `Native(SwapQuote.SwapKit)` result and reads the sub-provider label off it.

`SwapProviderTable` does not offer SwapKit on Bitcoin, so this path is only exercised by the new unit tests until the signer + table entry land — no behaviour change for users.

## Follow-up on this branch — the PSBT signer (#4666)
Signing moves real BTC and needs on-device testing + the BTC-keysign owner's review, so it lands next on this feature branch. Android has a Bitcoin PSBT path, but it consumes Vultisig's own `SignBitcoin` proto (dApp co-signing), **not** a raw BIP-174 PSBT, and rejects P2TR inputs. Two approaches — maintainer call:
- **A — rebuild as a deposit:** build our own BTC send of the input amount to `targetAddress`, reusing the `UtxoHelper` path (like Thor/Maya BTC swaps). Low new-code risk; assumes SwapKit BTC routes are deposit-based.
- **B — sign SwapKit's BIP-174 PSBT:** parse the base64 PSBT into `SignBitcoin` form and sign it (iOS-faithful). Net-new PSBT parsing + taproot handling.

Lean is **B** for iOS parity; owner's call with on-device testing.

## Test plan
- Unit tests pin the PSBT decode, native-decimal fee (400 sats), payload fields, and the malformed/empty base64 failure modes. `:data` + `:app` swap suites green; ktfmt clean.
- End-to-end keysign/broadcast is covered by the signer follow-up (#4666).

## Summary by CodeRabbit

* **New Features**
  * Added support for native Bitcoin swap transactions.

* **Improvements**
  * More robust swap quote handling across EVM, Solana and native routes.
  * Improved swap provider label resolution across different quote sources.
  * Corrected native fee scaling and unit handling for Bitcoin.

* **Tests**
  * Added tests covering Bitcoin PSBT routes and error cases for invalid PSBT payloads.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4665?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -641,8 +641,12 @@ constructor(
                         ),
                     )
                 val evmResult =
-                    (result as? SwapQuoteResult.Evm)
-                        ?: error("SwapKitQuoteSource must return SwapQuoteResult.Evm")
+                    when (result) {
+                        // Non-EVM SwapKit (Phase 2: BTC/PSBT) arrives as a fully-formed
+                        // SwapQuote.SwapKit on the Native result — cache it directly.
+                        is SwapQuoteResult.Native -> return@getCachedQuoteOrFetch result.quote
+                        is SwapQuoteResult.Evm -> result
+                    }
                 val apiQuote = evmResult.data
                 val expectedDstValue =
                     TokenValue(
@@ -687,7 +691,12 @@ constructor(
         // Read the sub-provider off the returned quote (not a hoisted var): a cache HIT skips the
         // fetch lambda, so a transient var would be null and the label would silently collapse from
         // "via CHAINFLIP" back to the generic "SwapKit" when the form re-opens within the TTL.
-        val resolvedSubProvider = (swapQuote as? SwapQuote.OneInch)?.subProvider
+        val resolvedSubProvider =
+            when (swapQuote) {
+                is SwapQuote.OneInch -> swapQuote.subProvider
+                is SwapQuote.SwapKit -> swapQuote.subProvider
+                else -> null
+            }
         val providerLabel =
             if (resolvedSubProvider.isNullOrBlank()) R.string.swap_for_provider_swapkit.asUiText()
             else formatSwapKitProviderLabel(resolvedSubProvider).asUiText()
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
@@ -415,8 +415,17 @@ constructor(
         val base64 =
             (element as? JsonPrimitive)?.takeIf { it.isString }?.content
                 ?: throw SwapKitError.Decoding("SwapKit non-EVM tx is not a base64 string")
-        return runCatching { Base64.getDecoder().decode(base64) }
-            .getOrElse { throw SwapKitError.Decoding("SwapKit non-EVM tx is not valid base64", it) }
+        val decoded =
+            runCatching { Base64.getDecoder().decode(base64) }
+                .getOrElse {
+                    throw SwapKitError.Decoding("SwapKit non-EVM tx is not valid base64", it)
+                }
+        // An empty (or whitespace-only) base64 decodes to zero bytes — refuse it rather than stage
+        // an unsignable empty payload that would only surface as a failure at sign time.
+        if (decoded.isEmpty()) {
+            throw SwapKitError.Decoding("SwapKit non-EVM tx payload is empty")
+        }
+        return decoded
     }
 
     /**
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
@@ -12,13 +12,18 @@ import com.vultisig.wallet.data.api.models.quotes.SwapKitSwapResponseJson
 import com.vultisig.wallet.data.api.swapAggregators.SwapKitApi
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.models.SwapKitSwapPayloadJson
+import com.vultisig.wallet.data.models.SwapQuote
+import com.vultisig.wallet.data.models.SwapQuote.Companion.expiredAfter
 import com.vultisig.wallet.data.models.TokenValue
 import java.math.BigDecimal
 import java.math.BigInteger
+import java.util.Base64
 import java.util.Locale
 import javax.inject.Inject
 import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.coroutines.flow.first
+import kotlinx.datetime.Clock
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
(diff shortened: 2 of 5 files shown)


aminsato:
Marking ready for review of the **data-layer half** (PSBT decode → `SwapQuote.SwapKit` + manager wiring + tests). It's inert in prod (the provider table doesn't offer SwapKit on BTC yet), so no behaviour change. The **signer** — including the fund-critical A (rebuild-as-deposit) vs B (sign the BIP-174 PSBT) decision — is tracked separately in #4666 and will be its own device-tested PR.

rkokhatskyi (commented):
@aminsato please treat it as a feature branch, should be ready for review only when all parts are done for BTC support for SwapKit

I've updated the PR title / description to indicate it is not yet ready

aminsato:
Understood @rkokhatskyi — converted back to draft and I'll treat this as the BTC feature branch, assembling all parts here and only marking ready once BTC SwapKit is complete (quote/payload **+ signer**, device-tested).

Plan on this branch:
1. ✅ Quote/payload decode → `SwapQuote.SwapKit` (this commit) + PSBT decode failure tests.
2. ⏳ The signer — tracked in #4666 with the fund-critical **A (rebuild-as-deposit) vs B (sign the BIP-174 PSBT)** decision; will land on this same branch.
3. ⏳ Enable SwapKit on Bitcoin in `SwapProviderTable` (the gate that makes the route reachable) — last, once signing works.

The A/B signing approach is the one open call that needs the BTC-keysign owner + on-device testing — happy to go with **B** (iOS parity) unless you'd prefer A. (Also pushed the CodeRabbit fix: empty/invalid PSBT base64 now fails fast as `Decoding`.)

aminsato:
@rkokhatskyi retargeted this from `main` to a new integration branch **`feature/swapkit-btc`** and marked it ready — so `main` never sees partial BTC. This PR is now a complete, tested *slice* against the feature branch (PSBT quote/payload decode + fee + failure-mode tests); the **signer** (#4666, A/B decision) and the `SwapProviderTable` BTC gate stack onto the same feature branch next. Only the final `feature/swapkit-btc → main` PR ships BTC once it's complete and device-tested.

If you'd rather not run an integration branch and prefer one big BTC PR at the end, say so and I'll fold it back.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSourceTest.kt
```diff
@@ -711,6 +715,97 @@ internal class SwapKitQuoteSourceTest {
...
+        assertEquals("PSBT", quote.data.txType)
+        assertTrue(psbtBytes.contentEquals(quote.data.txPayload))
+        assertEquals("bc1ptarget", quote.data.targetAddress)
+        assertEquals("btc-swap-1", quote.data.swapId)
+        assertEquals("NEAR", quote.data.subProvider)
+        assertEquals(btc, quote.data.fromCoin)
+        // 0.000004 BTC inbound fee scaled by 8 native decimals = 400 sats (not 0.000004 × 10^18).
+        assertEquals(BigInteger("400"), quote.fees.value)
+        assertEquals("BTC", quote.fees.unit)
+        assertEquals(8, quote.fees.decimals)
```
rkokhatskyi, line 751: These asserts pin the payload bytes, fee and routing fields but never touch `expectedDstValue`, `toAmountDecimal`, `toCoin` or `fromAmount`, so the `toAmountDecimal.movePointRight(dstToken.decimal).toBigInteger()` scaling in `buildSwapKitNativeQuote` could regress to the wrong decimals or drop entirely while every assertion here still passes, which leaves the most fragile line in the new code unpinned

```kotlin
assertEquals(BigInteger("385240000000000000000"), quote.expectedDstValue.value)
assertEquals(ethCoin(), quote.data.toCoin)
assertEquals(0, BigDecimal("385.24").compareTo(quote.data.toAmountDecimal))
assertEquals(BigInteger.ONE, quote.data.fromAmount)
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -641,8 +641,12 @@ constructor(
                         ),
                     )
                 val evmResult =
-                    (result as? SwapQuoteResult.Evm)
-                        ?: error("SwapKitQuoteSource must return SwapQuoteResult.Evm")
+                    when (result) {
+                        // Non-EVM SwapKit (Phase 2: BTC/PSBT) arrives as a fully-formed
+                        // SwapQuote.SwapKit on the Native result — cache it directly.
+                        is SwapQuoteResult.Native -> return@getCachedQuoteOrFetch result.quote
```
rkokhatskyi, line 647: `SwapQuoteManagerTest` carries a single timeout test, so this `Native` early return and the `is SwapQuote.SwapKit -> swapQuote.subProvider` label arm below it have no coverage, which means the cache hit behaviour the comment at `691` exists to protect, where the `via NEAR` label has to survive a cache hit rather than collapse to the generic `SwapKit`, ships unverified while a regression dropping the `Native` arm would still compile

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -326,10 +352,82 @@ constructor(
...
+            )
+        return SwapQuote.SwapKit(
+            expectedDstValue =
+                TokenValue(
+                    toAmountDecimal.movePointRight(dstToken.decimal).toBigInteger(),
+                    dstToken,
+                ),
+            // Source-chain native deposit fee (BTC 8dp), same surface as the EVM/Solana inbound
+            // fee.
+            fees = TokenValue(inboundFeeRawUnits(srcToken, route), srcToken),
```
rkokhatskyi: `inboundFeeRawUnits` reads the fee from `route` which is the `/v3/quote` route, whereas iOS reads it from the `/v3/swap` response through `SwapKitService.inboundFee(from response)` and `SwapKitSwapResponseJson` models no `fees` field at all, so once SwapKit reprices the inbound fee on the refreshed `/v3/swap` reply this verify screen number goes stale while iOS stays fresh, and since the same helper feeds the live EVM and Solana routes the gap is not unique to Bitcoin

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -326,10 +352,82 @@ constructor(
...
+                ?: throw SwapKitError.MalformedAmount(raw = response.expectedBuyAmount.orEmpty())
+        val payload =
+            SwapKitSwapPayloadJson(
+                fromCoin = srcToken,
+                toCoin = dstToken,
+                fromAmount = request.tokenValue.value,
+                toAmountDecimal = toAmountDecimal,
+                txType = response.meta.txType,
+                txPayload = decodeBinaryTx(response.tx),
+                targetAddress = response.targetAddress.orEmpty(),
```
rkokhatskyi: `response.targetAddress.orEmpty()` quietly accepts a missing address whereas iOS decodes `targetAddress` as a required field that fails the whole response when absent, which is harmless for PSBT since the destination lives in the PSBT outputs, yet this builder is the one the docstring earmarks for the deposit only chains such as Cardano and XRP that route entirely on `targetAddress`, so a blank value there would stage an unspendable quote

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -326,10 +352,82 @@ constructor(
...
+                memo = null,
+                subProvider = subProvider.orEmpty(),
+                swapId = response.swapId.orEmpty(),
+            )
+        return SwapQuote.SwapKit(
+            expectedDstValue =
+                TokenValue(
+                    toAmountDecimal.movePointRight(dstToken.decimal).toBigInteger(),
```
rkokhatskyi, line 394: This scale plus the parse at `378` repeat `scaleDecimalToRawUnits`, which makes a third copy of the same `BigDecimal` parse and `movePointRight(decimals).toBigInteger()`, so a shared helper returning a `BigInteger` would let the EVM and native paths share one implementation

rkokhatskyi (changes requested):

data/src/test/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSourceTest.kt
aminsato: Fixed in `234968262` — the PSBT test now asserts `expectedDstValue` (385.24 ETH → 385240000000000000000), `toAmountDecimal`, `toCoin`, and `fromAmount`, so the native scaling line is pinned.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Covered in `234968262` — `SwapQuoteManagerTest` now asserts a SwapKit `Native` quote surfaces `SwapKit (NEAR)` and that the label survives a cache hit (second fetch skips the lambda; `getQuote` verified called once). A regression dropping the `Native` arm fails it.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Fixed in `234968262` — added `fees` to `SwapKitSwapResponseJson` and `inboundFeeRawUnits` now reads from the `/v3/swap` response (`response.fees`), not the quote route, so a repriced fee stays fresh — matching iOS. Applies to the EVM/Solana paths too.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Fixed in `234968262` — a blank `targetAddress` now throws `SwapKitError.Decoding` (deposit-only chains route entirely on it, so a blank value would stage an unspendable quote).

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Fixed in `234968262` — extracted `parseExpectedBuyAmount(): BigDecimal`; `scaleDecimalToRawUnits` and the native-quote builder now share one parse instead of a third copy.

Issue #4119, opened 2026-04-19
Keysign join fails immediately on relay server 500 from startSession


aminsato:
# [Add] Retry with backoff on relay server 5xx in startSession [Join Keysign]

## Problem
When joining a keysign, a transient HTTP 500 from `POST /{sessionId}` causes `SessionApiImpl.startSession` to throw a raw `IllegalStateException` with no retry. `JoinKeysignViewModel` catches it and sets an error state showing the user a message like "Request failed with status 500 Internal Server Error", with no recovery path short of rescanning the QR code. This means a momentary relay server hiccup breaks the entire signing ceremony for all participants.

## Expected Behavior
Relay-server 5xx responses during `startSession` (and related relay calls) should be treated as transient. The client should retry a few times with exponential backoff before surfacing an error. If retries are exhausted, a user-friendly message ("Relay server unavailable, please try again") should be shown alongside a Retry action — not the raw exception message.

## Solution
Wrap relay-server calls in `SessionApiImpl` with a retry helper that retries on 5xx or `IOException` with exponential backoff (e.g. 3 attempts). Replace the raw `error(...)` in `throwIfUnsuccessful` with a typed exception that carries the HTTP status code so callers can distinguish 4xx (non-retryable) from 5xx (retryable) without string parsing. Map exhausted retries to a typed error (e.g. `JoinKeysignError.RelayUnavailable`) so `JoinKeysignViewModel` can display a proper message and a Retry action. The same retry policy should be applied to `startWithCommittee` and `checkCommittee`, which hit the same relay server. Precedent already exists in the same file: `markLocalPartyComplete` (SessionApi.kt:104-108) logs and ignores 5xx — the new policy should be consistent.

### Must Do
- [ ] Replace raw `error(...)` in `throwIfUnsuccessful` (Http.kt:6-12) with a typed exception (e.g. `HttpException(statusCode: Int)`) that distinguishes 4xx vs 5xx.
- [ ] Add a retry-with-backoff helper in `SessionApiImpl` and apply it to `startSession`, `startWithCommittee`, and `checkCommittee` for 5xx / `IOException` (3 attempts, exponential backoff).
- [ ] Add a typed `JoinKeysignError.RelayUnavailable` state in `JoinKeysignViewModel` and update the error UI to show a user-friendly message and a Retry action instead of the raw exception string.

### Must NOT Do
- Don't change unrelated API calls or non-relay-server error handling.
[...]

PR #4667, opened 2026-05-27
fix(keysign): handle relay 5xx race on join registration (#4625)

### LINKED ISSUE #4625: Keysign join intermittently fails with "Relay server unavailable" despite successful initial registration
## Summary

When joining a keysign session through the relay (Vultisig Relay), the join occasionally fails with `Relay server unavailable, please try again` even though the first POST to register this device succeeded.

## Reproduction

1. Initiator starts a keysign on device A (custom message / sign transaction).
2. Joiner device B scans the QR (or taps the FCM push notification).
3. Joiner taps **Sign** on the verify screen.
4. Error screen "Signing Error. Relay server unavailable, please try again" appears after ~6 seconds.

Reproduces intermittently — looks timing-sensitive against when the initiator locks the session.

## Observed network behavior (joiner side)

Four POSTs to `/router/{sessionId}` with the joiner's local party ID:

| # | Result | Notes |
|---|--------|-------|
| 1 | **201 Created** | Registration succeeded |
| 2 | 500 | Fired ~1.2 s after #1 succeeded |
| 3 | 500 | Retry +1 s |
| 4 | 500 | Retry +2 s |

After three 500s the relay-retry exhausts and the UI surfaces `RelayUnavailable`.

## What I think is going on

Two independent things appear wrong:

1. **A second `startSession` POST is being made even though the first returned 201.** The single-flight guard on the join action should prevent this. Possibilities to investigate: stale ViewModel from a previous QR scan, FCM auto-navigation racing with a manual tap, or the guard being reset somewhere on the success path.
2. **The 500 returned by the relay on a duplicate/locked registration is not transient**, but the relay-retry treats it as such and retries three times. Even if the duplicate call is eliminated, retrying 500 here will never succeed because once the initiator has locked the session the relay rejects further registrations.

## Environment

- Android (debug build), emulator
- Vultisig Relay (`api.vultisig.com/router`)
- Keysign type: custom message via QR + push notification

johnnyluo:
## Summary
- Remove `withRelayRetry` from `startSession` — 5xx on that endpoint means the session is already locked, not a transient failure, so retrying is futile and burns 3+ seconds before surfacing the error
- Add `joinOrConfirmRegistered()` in `JoinKeysignViewModel`: when `startSession` returns a 5xx, fall back to `getParticipants` to check if the local party already appears in the session; if so, proceed silently (another ViewModel instance registered first); if not, rethrow so the error surfaces normally

## Issue
Closes #4625

## Root Cause
A double-ViewModel race exists when FCM push notification and QR scan both navigate to `Route.Keysign.Join` nearly simultaneously. Each creates its own `JoinKeysignViewModel`; the `isJoiningKeysign` guard is per-instance, so both can call `startSession`. The first gets 201; the second gets 5xx because the session is now locked. Before this fix, `withRelayRetry` would retry that 5xx three more times (wasting ~3 s) and then throw `RelayUnavailable`, failing the keysign join.

## Test Plan
- [ ] Trigger a keysign from a paired device and confirm the joiner side completes successfully (no "relay unavailable" toast)
- [ ] Trigger a keysign via both FCM push and QR scan at the same time; confirm the join still succeeds

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session initialization reliability by refining error handling for server-side failures: the system now checks participant registration before retrying, avoids unnecessary retries or surfaced errors when participants are already registered, and preserves cancellation behavior for in-flight operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4667?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


data/src/main/kotlin/com/vultisig/wallet/data/api/SessionApi.kt
```diff
@@ -86,9 +86,7 @@ constructor(private val json: Json, private val httpClient: HttpClient) : Sessio
         sessionId: String,
         localPartyId: List<String>,
     ) {
-        withRelayRetry {
-            httpClient.post("$serverUrl/$sessionId") { setBody(localPartyId) }.throwIfUnsuccessful()
-        }
+        httpClient.post("$serverUrl/$sessionId") { setBody(localPartyId) }.throwIfUnsuccessful()
```
aminsato: Removing `withRelayRetry` here drops both 5xx and IOException retry for all five `startSession` callers, but only the keysign-join path gained the `joinOrConfirmRegistered` fallback. These four lose retry with no compensation:

- `JoinKeygenViewModel` (keygen join)
- `KeygenPeerDiscoveryViewModel` (keygen initiator)
- `KeysignFlowViewModel` (keysign initiator, two call sites)

For an initiator creating a session, a 5xx/IOException is a genuine transient candidate, not a "session already locked" signal. Keep the retry wrapper here (or at least its IOException handling) and special-case the 5xx-already-locked condition at the join call site, rather than stripping retry from the shared method.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1476,6 +1476,32 @@ constructor(
         }
     }
 
+    private suspend fun joinOrConfirmRegistered() {
+        try {
+            sessionApi.startSession(_serverAddress, _sessionID, listOf(_localPartyID))
+        } catch (e: HttpException) {
```
aminsato: This only catches `HttpException`. A transient `IOException` during `startSession` (which `withRelayRetry` previously retried 3×) now propagates to the outer `catch (e: Exception)` and maps to `FailedToStart`, so `tryAgain()` routes to `Destination.Back` instead of the `RelayUnavailable` auto-retry path. A recoverable network blip now bounces the user back rather than retrying.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1476,6 +1476,32 @@ constructor(
...
+            if (e.statusCode < 500) throw e
+            // A 5xx from the registration endpoint may mean the session is already locked by the
+            // initiator (e.g. a second ViewModel raced with the first). Check whether the local
+            // party is already in the participants list before surfacing an error — if it is, the
+            // registration succeeded via a previous call and we can proceed normally.
+            val alreadyRegistered =
+                runCatching {
+                        sessionApi
+                            .getParticipants(_serverAddress, _sessionID)
```
aminsato: `getParticipants` has no retry, and any failure here collapses to `false`. If the relay momentarily 5xx's on this GET during the race window, `alreadyRegistered` is `false` and the original 5xx is rethrown as `RelayUnavailable` — even though the party is registered. The fallback fails in exactly the flaky conditions it targets. Wrap this check in a short retry, or treat a 5xx on the GET as inconclusive rather than "not registered".

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1476,6 +1476,32 @@ constructor(
         }
     }
 
+    private suspend fun joinOrConfirmRegistered() {
+        try {
+            sessionApi.startSession(_serverAddress, _sessionID, listOf(_localPartyID))
```
aminsato: Before this change the losing ViewModel errored out; now it proceeds past here into `waitForKeysignToStart()` and a full TSS run with the same `_localPartyID` as the winner. If navigation doesn't clear the superseded ViewModel, two TSS state machines run for one party against the same session. Confirm `onCleared`/`cleanUp` tears down the first ViewModel when the second navigation supersedes it.

aminsato (commented):
Reviewed the relay 5xx race fix. The join-path logic is sound: `_localPartyID` comes from `vault.localPartyID`, so it's deterministic per device and the second racing ViewModel really does find its party already registered. Main concern is in `SessionApi.startSession` — the change is made in the shared method, so it affects more than the join path (see inline). The new already-registered branch in `joinOrConfirmRegistered` also has no test coverage; the existing retry tests only exercise `checkCommittee`.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1444,7 +1444,7 @@ constructor(
             withContext(Dispatchers.IO) {
                 try {
                     Timber.tag("JoinKeysignViewModel").d("Joining keysign")
-                    sessionApi.startSession(_serverAddress, _sessionID, listOf(_localPartyID))
+                    joinOrConfirmRegistered()
```
rkokhatskyi: This handles the relay response but leaves the actual cause in place since two `JoinKeysignViewModel` instances still get created when `ScanQrViewModel` and the push notification both `route` to `Route.Keysign.Join`, and `launchSingleTop` at `NavController.kt:37` will not collapse them when they arrive before either back stack entry exists. iOS drives this from one `JoinKeysignViewModel` and the extension from a single `JoinMpcSessionStep` so neither has the duplicate, which points to deduping the route at the navigation layer as the fix that removes the race rather than tolerating a 5xx after the fact, so it is worth confirming whether a navigation level dedup was considered

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1476,6 +1476,32 @@ constructor(
...
+    private suspend fun joinOrConfirmRegistered() {
+        try {
+            sessionApi.startSession(_serverAddress, _sessionID, listOf(_localPartyID))
+        } catch (e: HttpException) {
+            if (e.statusCode < 500) throw e
+            // A 5xx from the registration endpoint may mean the session is already locked by the
+            // initiator (e.g. a second ViewModel raced with the first). Check whether the local
+            // party is already in the participants list before surfacing an error — if it is, the
+            // registration succeeded via a previous call and we can proceed normally.
+            val alreadyRegistered =
```
rkokhatskyi: This branch only rescues the case where the party is already in the list, so a genuine transient relay 500 on the first registration, when the party is not yet registered, now falls through to the rethrow and surfaces as `RelayUnavailable` again. That is the exact #4119 symptom that commit `3ade3701e` fixed by adding `withRelayRetry`, which this PR removes, while `JoinKeygenViewModel.kt:304` calls `startSession` bare so it loses that retry with no fallback at all

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1476,6 +1476,32 @@ constructor(
         }
     }
 
+    private suspend fun joinOrConfirmRegistered() {
```
rkokhatskyi: This encodes a property of the relay endpoint rather than of this one screen, yet four other sites POST the same path and only this one recovers, so it reads more naturally in `SessionApiImpl.startSession` the way `markLocalPartyComplete` at `SessionApi.kt:110` already swallows a relay 5xx in the impl, which also drops the `joinOrConfirmRegistered` name that does not actually join. Something like
```kotlin
override suspend fun startSession(serverUrl: String, sessionId: String, localPartyId: List<String>) {
    val response = httpClient.post("$serverUrl/$sessionId") { setBody(localPartyId) }
    if (response.status.value >= 500 &&
        runCatching { getParticipants(serverUrl, sessionId).containsAll(localPartyId) }.getOrDefault(false)
    ) return
    response.throwIfUnsuccessful()
}
```
lets every caller behave the same and keeps the join `ViewModel` thin

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1476,6 +1476,32 @@ constructor(
...
+        try {
+            sessionApi.startSession(_serverAddress, _sessionID, listOf(_localPartyID))
+        } catch (e: HttpException) {
+            if (e.statusCode < 500) throw e
```
rkokhatskyi: The whole recovery hinges on the lock signal being `>= 500`, which is undocumented since the LAN mediator `postSession` always returns 201 and neither iOS nor the extension encodes a locked status, whereas an already registered or locked session is more naturally a 409 which this `e.statusCode < 500` rethrow passes straight through to `FailedToStart`. Worth pinning the relay contract with the backend and widening the branch to a 4xx conflict if that is how a locked session is reported

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
johnnyluo: Confirmed — `onCleared()` calls `cleanUp()` (which cancels `_jobWaitingForKeysignStart`) and Android's `ViewModel` infrastructure also cancels `viewModelScope` at that point. The `joinKeysign()` coroutine runs inside `viewModelScope.safeLaunch`, so when navigation supersedes the first ViewModel and `onCleared` fires, the entire coroutine tree — including any in-flight TSS setup — receives a `CancellationException` and is torn down before it can enter the signing phase. The second ViewModel becomes the sole live instance.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
johnnyluo: Acknowledged — deduping at the navigation layer is the cleaner long-term fix and removes the race at its source. That change is larger in scope (NavController/launchSingleTop behaviour), so tracking it as a follow-up rather than holding this fix. The session-layer idempotency in `SessionApiImpl.startSession` now acts as a defence-in-depth fallback for that and any future duplicate-navigation scenario.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
johnnyluo: Fair point — the `>= 500` threshold matches what was observed in issue #4625 (relay returning 5xx on a locked session), but pinning the exact contract with the backend is the right follow-up. If the relay ever moves to 409 for a locked/already-registered session, the check would need to be widened to cover that. Leaving this as a known assumption to confirm with the backend team before this PR merges.

johnnyluo: The backend will not change this behaviour

rkokhatskyi (approved):
LGTM

PR #4671, opened 2026-05-28
feat(swap): SwapKit BTC/PSBT support

aminsato:
## Summary

Umbrella branch (`feature/swapkit-btc` → `main`) for SwapKit non-EVM **Bitcoin / PSBT** support — Phase 2 of the SwapKit aggregator epic #4589. SwapKit EVM + Solana already ride the existing `EVMSwapQuoteJson` envelope on `main` (Phase 1); Bitcoin's wire shape is a base64 BIP-174 PSBT that doesn't fit that envelope, so this branch adds the `SwapQuote.SwapKit` native variant plus a dedicated PSBT signer.

Work lands as stacked PRs into this branch. Only the final `feature/swapkit-btc → main` merge ships BTC, once it is complete and device-tested.

### Tranasction hash
https://mempool.space/tx/72c797d4ea6b648930ab8d8c521d58401a7d435e7b0f1239866273a2dbd2b403

### Done on this branch
- [x] **Quote + PSBT payload decode** (#4665, merged) — `SwapKitQuoteSource` decodes a PSBT route into a fully-formed `SwapQuote.SwapKit`: `Bitcoin → "BTC"` prefix, native decimals (8) so the inbound BTC deposit fee scales correctly, and a `TxKind.PSBT` arm that surfaces the base64 PSBT bytes, `targetAddress`, `toAmount`, sub-provider and swap id on `SwapQuoteResult.Native`.
- [x] `SwapQuoteManager` caches the `Native` result directly and resolves the `via <sub-provider>` label off the returned `SwapQuote.SwapKit` so it survives a cache hit.
- [x] Inbound fee read from the `/v3/swap` response `fees[]` (not the `/v3/quote` route), matching iOS so a repriced fee stays fresh.
- [x] `targetAddress` required for non-EVM routes (blank → `Decoding`).
- [x] Unit tests pin the PSBT decode, the native-quote amount + fee scaling (400 sats), the malformed / empty base64 failure modes, and the manager's sub-provider label across a cache hit.

### To do before `feature/swapkit-btc → main`
- [ ] **PSBT signer** (#4674) — approach B (iOS parity): decode the PSBT into `SignBitcoin`, reuse `UtxoHelper`'s BIP-143 sighashes + the vault / destination binding, assemble the signed segwit tx from the MPC signatures, dispatched by `txType` from `SigningHelper`. Not yet merged into this branch.
- [ ] **Enable SwapKit on Bitcoin** in `SwapProviderTable` — the gate that makes the route reachable. Currently `Bitcoin → {THORCHAIN, MAYA}` (SwapKit excluded), so the whole path is unreachable in production. Flip last, once signing works.
- [ ] **On-device E2E** — compile + broadcast a real BTC SwapKit swap (NEAR / Chainflip / Garden / Flashnet route) before flipping the table entry, plus maintainer review of the fund-moving signer.

## Inert in production

`SwapProviderTable` still maps Bitcoin to `{THORCHAIN, MAYA}` and does not offer SwapKit on Bitcoin, so everything added on this branch so far is unreachable in production and exercised only by unit tests. No behaviour change for users until the signer + table entry land.

## Test plan
- [x] `:data` + `:app` swap suites green; ktfmt clean
- [x] PSBT decode, native-quote fields, fee scaling, and the malformed / empty base64 failure modes pinned by unit tests
- [x] Manager preserves the SwapKit sub-provider label across a cache hit
- [ ] BIP-143 sighash golden vector for the PSBT signer (lands with #4674)
- [ ] On-device E2E on a real BTC swap before the `SwapProviderTable` entry is enabled

## Summary by CodeRabbit

* **New Features**
  * Added Bitcoin PSBT swap support via SwapKit, including PSBT signing and UI display of SwapKit provider labels.
  * Swap responses now include executed-route fees for accurate fee extraction and routing.

* **Bug Fixes**
  * Fixed handling of non‑EVM swap quotes to avoid hard failures and preserve sub‑provider labels across cache hits.
  * View model fee and transaction flows updated to handle SwapKit without triggering incorrect UTXO paths.

* **Tests**
  * Expanded tests covering PSBT parsing, BTC signing, fee handling, and native swap flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4671?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -248,7 +269,7 @@ constructor(
         // is SwapKitService.inboundFee). Embedded in `swapFee` with `swapFeeTokenContract = ""`
         // (the native-coin sentinel resolveSwapFee recognises) so the consumer reads it the same
         // way Kyber/Jupiter quote sources surface their per-leg fees.
-        val inboundFee = inboundFeeRawUnits(srcToken, route).toString()
+        val inboundFee = inboundFeeRawUnits(srcToken, response.fees).toString()
```
rkokhatskyi: This flips the inbound fee source for the already shipped EVM and Solana path from the `/v3/quote` route to `response.fees`, so if the real `/v3/swap` reply omits `fees[]` for an EVM route the verify screen Network Fee silently reads zero, and the mocked tests cannot catch that since they inject `fees` into the swap response by hand, so it is worth confirming against a captured EVM `/v3/swap` response before this reaches `main`

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -326,10 +347,84 @@ constructor(
...
+        subProvider: String?,
+    ): SwapQuote.SwapKit {
+        val srcToken = request.srcToken
+        val dstToken = request.dstToken
+        val toAmountDecimal = parseExpectedBuyAmount(response.expectedBuyAmount)
+        // Deposit-only chains (Cardano / XRP) route entirely on `targetAddress`, so a blank value
+        // would stage an unspendable quote — refuse it rather than emit a quote that can't settle.
+        val targetAddress =
+            response.targetAddress?.takeIf { it.isNotBlank() }
+                ?: throw SwapKitError.Decoding("SwapKit non-EVM response missing targetAddress")
```
rkokhatskyi, line 396: This blank `targetAddress` rejection has no test whereas the two negative cases below cover invalid and empty base64, so a regression dropping the guard would ship green, and since the PSBT already encodes the Bitcoin destination this guard really protects the deposit only chains the comment names, so a case asserting `Decoding` on a null `targetAddress` would pin the intent

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -188,17 +206,19 @@ constructor(
...
+     * SwapKit-attributed cost of executing the deposit on the source chain. Read from the swap
+     * response (not the `/v3/quote` route) so a fee SwapKit reprices on the refreshed `/v3/swap`
+     * reply stays fresh, matching iOS' `SwapKitService.inboundFee(from response)`. Mirrors iOS so
+     * the displayed Network Fee reconciles against the actual on-chain debit (especially on Solana
+     * and cross-chain Chainflip / NEAR routes where the EVM gas envelope would otherwise read
+     * zero). Returns `0` when the entry is absent so the UI surfaces "—" rather than blocking the
+     * quote.
      */
-    private fun inboundFeeRawUnits(srcToken: Coin, route: SwapKitRoute): BigInteger {
+    private fun inboundFeeRawUnits(srcToken: Coin, fees: List<SwapKitFee>): BigInteger {
```
rkokhatskyi: Now that `inboundFeeRawUnits` takes `response.fees` rather than the route, `SwapKitRoute.fees` at `SwapKitQuoteJson.kt:86` is no longer read anywhere in production, so it is carried only for wire completeness now and could be dropped unless a later ranking step needs it

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Good catch — addressed in 921b0acf9. `inboundFeeRawUnits` still prefers the `/v3/swap` `response.fees` (matching iOS' `SwapKitService.inboundFee`, which keeps a repriced fee fresh), but it now falls back to the `/v3/quote` route's inbound fee when the swap reply omits the entry. So an EVM/Solana route whose `/v3/swap` carries no `fees[]` shows the quoted fee instead of silently reading zero.

Added a test (`fetch falls back to route fees when the swap response omits the inbound entry`) that drives exactly the omitted-`fees[]` case the hand-injected mocks couldn't catch before. Side effect: `SwapKitRoute.fees` is read again, so it's no longer a dead field.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Done in 7eb43c362 — added `fetch throws Decoding when a non-EVM response has no targetAddress`, which feeds a valid base64 PSBT with `targetAddress = null` so it pins this guard specifically (rather than the decode path), sitting alongside the invalid/empty-base64 negative cases.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Kept and read again as of 921b0acf9 — rather than drop it, `SwapKitRoute.fees` is now the fallback source for the inbound fee: `inboundFeeRawUnits` takes `(response.fees, fallback = route.fees)` and uses the route fee when the `/v3/swap` reply omits the inbound entry (see the inbound-fee thread above). So it's no longer dead.

johnnyluo (approved):
LGTM

PR #4672, opened 2026-05-28
fix(swap): preserve full token precision for MAX/percentage buttons

### LINKED ISSUE #4668: Swap MAX/percentage buttons truncate amount to 6 decimals, can't swap full balance
## Summary

On the Swap screen, tapping the percentage buttons (25% / 50% / 75% / **MAX**) inserts a "From" amount with fewer decimals than the token actually has. For high-precision / low-value tokens this drops the trailing decimals, so the user cannot swap their full balance.

## Steps to reproduce

1. Open Swap with a "From" token whose balance has more than 6 decimal places (e.g. a small COMP balance on BSC).
2. Tap **MAX**.

## Expected

The full available balance is inserted, preserving the token's actual decimal precision.

## Actual

The inserted amount is truncated to 6 decimals.

Example (from a real session):
- Balance: `0.00553297 COMP`
- After tapping MAX: `0.005532` (last two decimals `97` dropped)

## Root cause

`SwapFormViewModel.selectSrcPercentage()` rounds the computed amount with a hardcoded scale of 6:

```kotlin
val amount =
    TokenValue.createDecimal(maxUsableTokenAmount, srcTokenValue.decimals)
        .multiply(percentage.toBigDecimal())
        .setScale(6, RoundingMode.DOWN)
```

The scale should follow the token's own `decimals` (clamped to a sane display max) instead of being fixed at 6. The flipped-amount formatter in the same file already does this — it scales by the token decimals capped at `MAX_DISPLAY_DECIMALS` (8) — so the percentage handler is inconsistent with the rest of the screen.

## Affected

- `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt` — `selectSrcPercentage()`

aminsato:
## Summary
- Swap MAX / 25% / 50% / 75% buttons truncated the inserted "From" amount to 6 decimals, so high-precision / low-value tokens lost trailing decimals and the full balance couldn't be swapped (e.g. `0.00553297` became `0.005532`).
- Root cause: `SwapFormViewModel.selectSrcPercentage()` used a hardcoded `.setScale(6, RoundingMode.DOWN)`.
- Fix: scale by the token's own `decimals` via the existing `formatFlippedAmount()` helper (capped at `MAX_DISPLAY_DECIMALS` = 8 and trailing zeros stripped), making the percentage handler consistent with the flipped-amount formatter already used on the same screen.

Closes #4668

## Test plan
- [x] Added `selectSrcPercentage preserves full token precision beyond 6 decimals` — asserts `0.00553297 BTC` at 100% stays intact instead of truncating to `0.005532`.
- [x] `./gradlew :app:testDebugUnitTest --tests "*.SwapFormViewModelTest"` — 77 tests, 0 failures.
- [ ] Manual: on Swap, pick a token with a balance having >6 decimals, tap MAX, confirm the full precision is inserted.

## Release Notes

* **Bug Fixes**
  * Fixed the swap form to preserve full token precision when selecting percentage amounts. Previously, values were truncated to 6 decimals; now the complete precision is retained for accurate swap calculations.

* **Tests**
  * Added test coverage verifying precision is maintained for tokens with more than 6 decimal places during percentage-based amount selection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4672?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -759,9 +759,9 @@ constructor(
         val amount =
             TokenValue.createDecimal(maxUsableTokenAmount, srcTokenValue.decimals)
                 .multiply(percentage.toBigDecimal())
-                .setScale(6, RoundingMode.DOWN)
+                .formatFlippedAmount(srcTokenValue.decimals)
 
-        srcAmountState.setTextAndPlaceCursorAtEnd(amount.toString())
+        srcAmountState.setTextAndPlaceCursorAtEnd(amount)
     }
 
     fun loadData(vaultId: String, chainId: String?, srcTokenId: String?, dstTokenId: String?) {
--- a/app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
@@ -539,6 +539,31 @@ internal class SwapFormViewModelTest {
             )
         }
 
+    @Test
+    fun `selectSrcPercentage preserves full token precision beyond 6 decimals`() =
+        runTest(mainDispatcher) {
+            // 0.00553297 BTC — more than 6 significant decimals. BTC src avoids the Ethereum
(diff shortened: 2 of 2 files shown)


rkokhatskyi (approved):
LGTM

PR #4674, opened 2026-05-28
feat(swap): SwapKit Bitcoin PSBT signer

aminsato:
## Summary

Wires the **signing** side for SwapKit Bitcoin swaps. The quote + PSBT payload decode already landed on `feature/swapkit-btc`; that produced a fully-formed `SwapQuote.SwapKit` carrying the base64-decoded PSBT, but the signing side was unwired (`SwapFormViewModel` threw "No SwapKit signer wired" and `SigningHelper` no-oped on `SwapPayload.SwapKit`). This makes a SwapKit BTC swap signable (TSS) and broadcastable.

Mirrors the iOS `SwapKitBTCSigner`: a dedicated signer dispatched by `txType == "PSBT"`, separate from the existing dApp PSBT co-sign path (which deliberately skips broadcast). The swap deposit must broadcast, so the swap path keeps `keysignPayload.signBitcoin == null` and decodes the PSBT on the fly in the dispatch.

## What's in it

- **`SwapKitPsbtParser`** — BIP-174 framing cursor (magic, key/value maps, compactSize, little-endian readers; rejects truncation, bad magic, duplicate keys, missing unsigned-tx global). Pure bytes, no crypto.
- **`SwapKitBtcSigner`** — decodes the PSBT into the structured `SignBitcoin` (P2WPKH / P2SH-P2WPKH, `witness_utxo`, every input `is_ours`; an output is marked change when its scriptPubKey matches the vault lock script), then:
  - reuses `UtxoHelper.getPreSignedImageHashFromSignBitcoin` for the BIP-143 sighashes **and** its existing defense-in-depth binding — pinning the deposit to SwapKit's `targetAddress`/`fromAmount` and the change to the vault;
  - assembles the signed segwit transaction from the MPC signatures (ported from the iOS compile path).
- **`SigningHelper`** — dispatches `SwapPayload.SwapKit` (`txType == "PSBT"`) on both the pre-sign and compile paths; any other `txType` fails loudly.
- **`SwapFormViewModel`** — builds the keysign payload for a SwapKit BTC quote (deposit → `targetAddress`, fees from the quote) instead of erroring.

## Reuse

No new crypto where the repo already had it: the BIP-143 sighashes and the vault/destination binding come from the existing `UtxoHelper` structured-PSBT path. New code is limited to the PSBT byte parser, the decode → `SignBitcoin` transcode, the segwit compile, and the dispatch/UI wiring.

## Safety / scope

- **Inert in production.** `SwapProviderTable` still does not offer SwapKit on Bitcoin, so this path is reached only by unit tests. Enabling it for real BTC is a deliberate follow-up.
- P2WPKH (+ P2SH-P2WPKH) inputs only, SIGHASH_ALL only; anything else throws a typed error.

## Test plan

- [x] `SwapKitPsbtParser` framing edge cases (empty, bad magic, truncation, duplicate key, missing unsigned-tx global) — pure, run headlessly.
- [x] **Golden vector**: decode the real `v3-real-btc-all-swap` PSBT and assert the BIP-143 sighashes match the values the iOS port pins for the same fixture (runs headlessly; the only correctness-critical check).
- [x] Binding happy-path + a rejection (deposit amount diverging from the quote) — run where the WalletCore JNI is available.
- [x] `:data:testDebugUnitTest` green; ktfmt clean.
- [ ] **Requires on-device E2E** on a real BTC swap (compile + broadcast) and maintainer review before the `SwapProviderTable` entry is enabled. Not exercisable in unit tests (the TSS lib is `compileOnly`).

## Release Notes

* **New Features**
  * Added support for SwapKit exchange protocol, enabling PSBT (Partially Signed Bitcoin Transaction) handling for improved swap capabilities on non-EVM blockchains.
  * Enhanced transaction signing flow to support SwapKit-routed transactions with proper validation and fee calculation.

* **Tests**
  * Added comprehensive test suite validating PSBT parsing, transaction signing, and edge-case error handling for SwapKit integration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4674?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Enablement blockers (before flipping the `SwapProviderTable` Bitcoin entry)

Surfaced in review — must be resolved with on-device validation before SwapKit is offered on Bitcoin:
- [ ] Source the displayed network fee from the decoded PSBT (`inputs − outputs`), not `getSpecificAndUtxo`'s coin-selection estimate; wire `utxoFeeData`/submit-guard for `SwapQuote.SwapKit` so the form is confirmable.
- [ ] Add a fail-closed sanity bound on the implicit miner fee once a sane threshold is established from real-route data.
- [ ] Validate the deposit/change binding against real NEAR / Garden / Flashnet PSBTs (with a change output present); the strict `sum(non-change) == fromAmount` + exact-match change detection may need loosening if a provider funds the deposit as `sellAmount − fee`.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -473,10 +473,41 @@ constructor(
                                 )
                             }
 
-                            // Per-chain signers wire this branch as they ship; no quote source
-                            // emits SwapQuote.SwapKit yet.
-                            is SwapQuote.SwapKit ->
-                                error("No SwapKit signer wired for txType=${quote.data.txType}")
+                            is SwapQuote.SwapKit -> {
+                                // Only BTC PSBT is wired today; other SwapKit txTypes (TON / ADA /
+                                // SUI / TRON) land with their per-chain signers. SwapProviderTable
+                                // does not yet offer SwapKit on Bitcoin, so this branch is reached
+                                // only once enablement ships — guarded loudly until then.
+                                require(quote.data.txType == SWAPKIT_PSBT_TX_TYPE) {
+                                    "Unsupported SwapKit txType for swap: ${quote.data.txType}"
+                                }
+                                val specificAndUtxo =
+                                    swapGasCalculator.getSpecificAndUtxo(
+                                        srcToken = srcToken,
+                                        srcAddress = srcAddress,
+                                        gasFee = gasFee,
+                                    )
+                                RegularSwapTransaction(
+                                    id = UUID.randomUUID().toString(),
+                                    vaultId = vaultId,
+                                    srcToken = srcToken,
+                                    srcTokenValue = srcTokenValue,
+                                    dstToken = dstToken,
+                                    // SwapKit's source-chain deposit address — where the BTC the
+                                    // PSBT spends is sent. Signing is driven entirely by the PSBT
+                                    // bytes carried on the payload, not by this blockChainSpecific.
+                                    dstAddress = quote.data.targetAddress,
+                                    expectedDstTokenValue = dstTokenValue,
+                                    blockChainSpecific = specificAndUtxo,
+                                    estimatedFees = quote.fees,
+                                    gasFees = estimatedNetworkFeeTokenValue.value ?: gasFee,
+                                    memo = quote.data.memo,
+                                    isApprovalRequired = false,
+                                    gasFeeFiatValue =
+                                        estimatedNetworkFeeFiatValue.value ?: gasFeeFiatValue,
+                                    payload = SwapPayload.SwapKit(quote.data),
+                                )
+                            }
 
                             is SwapQuote.OneInch -> {
                                 val dstAddress = quote.data.tx.to
@@ -1229,6 +1260,9 @@ constructor(
 }
 
 private const val MAX_DISPLAY_DECIMALS = 8
+
+// SwapKit `meta.txType` for the only non-EVM SwapKit swap wired today (Bitcoin PSBT).
+private const val SWAPKIT_PSBT_TX_TYPE = "PSBT"
 
 internal fun BigDecimal.formatFlippedAmount(tokenDecimals: Int? = null): String =
     setScale(
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -15,6 +15,7 @@ import com.vultisig.wallet.data.models.Address
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.EVMSwapPayloadJson
 import com.vultisig.wallet.data.models.FiatValue
+import com.vultisig.wallet.data.models.SwapKitSwapPayloadJson
 import com.vultisig.wallet.data.models.SwapProvider
 import com.vultisig.wallet.data.models.SwapQuote
 import com.vultisig.wallet.data.models.SwapTransaction.RegularSwapTransaction
@@ -478,7 +479,7 @@ constructor(
                                 // SUI / TRON) land with their per-chain signers. SwapProviderTable
                                 // does not yet offer SwapKit on Bitcoin, so this branch is reached
                                 // only once enablement ships — guarded loudly until then.
-                                require(quote.data.txType == SWAPKIT_PSBT_TX_TYPE) {
+                                require(quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_PSBT) {
                                     "Unsupported SwapKit txType for swap: ${quote.data.txType}"
                                 }
                                 val specificAndUtxo =
@@ -1260,9 +1261,6 @@ constructor(
 }
 
 private const val MAX_DISPLAY_DECIMALS = 8
-
-// SwapKit `meta.txType` for the only non-EVM SwapKit swap wired today (Bitcoin PSBT).
-private const val SWAPKIT_PSBT_TX_TYPE = "PSBT"
 
 internal fun BigDecimal.formatFlippedAmount(tokenDecimals: Int? = null): String =
     setScale(
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
@@ -26,6 +26,8 @@ import wallet.core.jni.EthereumAbi
 
 object SigningHelper {
     private const val ETH_SIGN_TYPED_DATA_V4 = "eth_signTypedData_v4"
+    // SwapKit `meta.txType` for the Bitcoin PSBT signer dispatch.
+    private const val PSBT_TX_TYPE = "PSBT"
 
     @OptIn(ExperimentalStdlibApi::class)
     fun getKeysignMessages(messagePayload: CustomMessagePayload): List<String> {
@@ -92,6 +94,18 @@ object SigningHelper {
                                 .getPreSignedImageHash(swapPayload.data, payload, nonceAcc)
 
                     messages += message
+                }
+                is SwapPayload.SwapKit -> {
+                    require(swapPayload.data.txType == PSBT_TX_TYPE) {
+                        "Unsupported SwapKit txType for signing: ${swapPayload.data.txType}"
+                    }
+                    messages +=
+                        SwapKitBtcSigner(ecdsaKey, ecdsaChainCode)
+                            .getPreSignedImageHash(
+                                psbtBytes = swapPayload.data.txPayload,
+                                targetAddress = swapPayload.data.targetAddress,
+                                fromAmount = swapPayload.data.fromAmount,
+                            )
                 }
                 else -> Unit
             }
@@ -273,6 +287,13 @@ object SigningHelper {
                             )
                 }
 
+                is SwapPayload.SwapKit -> {
+                    require(swapPayload.data.txType == PSBT_TX_TYPE) {
+                        "Unsupported SwapKit txType for signing: ${swapPayload.data.txType}"
+                    }
+                    return SwapKitBtcSigner(ecdsaKey, ecdsaChainCode)
+                        .getSignedTransaction(swapPayload.data.txPayload, signatures)
+                }
                 else -> {}
             }
         } else if (swapPayload is SwapPayload.MayaChain && !swapPayload.srcToken.isNativeToken) {
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
@@ -11,6 +11,7 @@ import com.vultisig.wallet.data.crypto.ThorChainHelper
 import com.vultisig.wallet.data.crypto.TonHelper
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.SignedTransactionResult
+import com.vultisig.wallet.data.models.SwapKitSwapPayloadJson
 import com.vultisig.wallet.data.models.TssKeyType
 import com.vultisig.wallet.data.models.TssKeysignType
 import com.vultisig.wallet.data.models.Vault
@@ -26,8 +27,6 @@ import wallet.core.jni.EthereumAbi
 
 object SigningHelper {
     private const val ETH_SIGN_TYPED_DATA_V4 = "eth_signTypedData_v4"
-    // SwapKit `meta.txType` for the Bitcoin PSBT signer dispatch.
-    private const val PSBT_TX_TYPE = "PSBT"
 
     @OptIn(ExperimentalStdlibApi::class)
     fun getKeysignMessages(messagePayload: CustomMessagePayload): List<String> {
@@ -96,7 +95,7 @@ object SigningHelper {
                     messages += message
                 }
                 is SwapPayload.SwapKit -> {
-                    require(swapPayload.data.txType == PSBT_TX_TYPE) {
+                    require(swapPayload.data.txType == SwapKitSwapPayloadJson.TX_TYPE_PSBT) {
                         "Unsupported SwapKit txType for signing: ${swapPayload.data.txType}"
                     }
                     messages +=
@@ -288,7 +287,7 @@ object SigningHelper {
                 }
 
                 is SwapPayload.SwapKit -> {
-                    require(swapPayload.data.txType == PSBT_TX_TYPE) {
+                    require(swapPayload.data.txType == SwapKitSwapPayloadJson.TX_TYPE_PSBT) {
                         "Unsupported SwapKit txType for signing: ${swapPayload.data.txType}"
                     }
                     return SwapKitBtcSigner(ecdsaKey, ecdsaChainCode)
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
@@ -1,0 +1,424 @@
+package com.vultisig.wallet.data.chains.helpers
+
+import com.vultisig.wallet.data.models.SignedTransactionResult
+import com.vultisig.wallet.data.utils.Numeric
+import java.io.ByteArrayOutputStream
+import java.math.BigInteger
+import java.security.MessageDigest
+import tss.KeysignResponse
+import vultisig.keysign.v1.BitcoinInput
+import vultisig.keysign.v1.BitcoinOutput
+import vultisig.keysign.v1.SignBitcoin
+import wallet.core.jni.BitcoinScript
+import wallet.core.jni.CoinType
+import wallet.core.jni.PublicKey
+import wallet.core.jni.PublicKeyType
+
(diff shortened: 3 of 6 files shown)


aminsato:
@coderabbitiai full review

aminsato:
@coderabbitai proceed

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
```diff
@@ -0,0 +1,424 @@
...
+            (0L until outputCount).map {
+                val amount = cursor.readUInt64LE()
+                val scriptLen = cursor.readCompactSize()
+                ParsedTxOutput(amount, cursor.readBytes(scriptLen.toInt()))
+            }
+        val locktime = cursor.readUInt32LE()
+        return ParsedTx(version, locktime, inputs, outputs)
+    }
+
+    private fun compileSignedTransaction(
```
rkokhatskyi: Since the SwapKit BTC transaction sets no `signBitcoin` and is not `skipBroadcast`, this `compileSignedTransaction` output is the raw tx that actually gets broadcast, yet no test calls `getSignedTransaction` so the segwit framing, witness assembly, the `p2sh-p2wpkh` scriptSig, and `txid` are all unexercised whereas the golden vector stops at the sighashes, and since the serialization is pure (only the pubkey needs the JNI the existing tests already gate) a compile golden that feeds the fixture PSBT plus a stubbed `signatures` map keyed by the pinned sighash hexes could assert `rawTransaction` and `transactionHash` against an iOS pinned value

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -473,10 +473,41 @@ constructor(
...
+                            is SwapQuote.SwapKit -> {
+                                // Only BTC PSBT is wired today; other SwapKit txTypes (TON / ADA /
+                                // SUI / TRON) land with their per-chain signers. SwapProviderTable
+                                // does not yet offer SwapKit on Bitcoin, so this branch is reached
+                                // only once enablement ships — guarded loudly until then.
+                                require(quote.data.txType == SWAPKIT_PSBT_TX_TYPE) {
+                                    "Unsupported SwapKit txType for swap: ${quote.data.txType}"
+                                }
+                                val specificAndUtxo =
+                                    swapGasCalculator.getSpecificAndUtxo(
```
rkokhatskyi, line 486: `utxoFeeData` is built only for `ThorChain` and `MayaChain` so a `SwapQuote.SwapKit` quote falls to the `else` null branch, which leaves `estimatedNetworkFeeFiatValue` null and makes submit throw the invalid gas fee error, so once Bitcoin is enabled the form can never be confirmed, and separately the PSBT's real miner fee which is inputs minus outputs is never computed or shown here while `getSpecificAndUtxo` fetches the live vault UTXOs that signing ignores, so the displayed network fee should come from the decoded PSBT rather than this estimate

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
```diff
@@ -0,0 +1,424 @@
...
+     * SwapKit's [targetAddress] and [fromAmount] and the change to the vault — a defense-in-depth
+     * binding that refuses to sign a PSBT whose outputs diverge from the quoted deposit.
+     */
+    fun getPreSignedImageHash(
+        psbtBytes: ByteArray,
+        targetAddress: String,
+        fromAmount: BigInteger,
+    ): List<String> {
+        val signBitcoin = decodeToSignBitcoin(psbtBytes)
+        return utxo.getPreSignedImageHashFromSignBitcoin(
```
rkokhatskyi, line 50: The binding pins the non change outputs to `fromAmount` and the change to the vault but nothing bounds the implicit miner fee which is sum of inputs minus sum of outputs, so a PSBT that sets a small change and a pathological fee still passes every check and is signed while the user sees neither the real fee nor a cap, so a sanity bound on inputs minus outputs before signing would fail closed if a provider ever returns a bad fee

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
```diff
@@ -0,0 +1,424 @@
...
+                inputMap[KEY_REDEEM_SCRIPT]
+                    ?: throw SwapKitBtcSignerException(
+                        "SwapKit BTC PSBT P2SH input #$index missing redeem script"
+                    )
+            if (!(redeem.size == 22 && redeem[0] == 0x00.toByte() && redeem[1] == 0x14.toByte())) {
+                throw SwapKitBtcSignerException(
+                    "SwapKit BTC PSBT P2SH redeem script on input #$index is not P2SH-P2WPKH"
+                )
+            }
+            return "p2sh-p2wpkh" to Numeric.toHexStringNoPrefix(redeem)
```
rkokhatskyi, line 220: Every test input is P2WPKH so this `p2sh-p2wpkh` branch, the redeem validation above it, and the only non empty `scriptSig` path that pushes the redeem are never exercised, which is the single place that writes a scriptSig into the broadcast tx, so a decode plus compile case with a `a914..87` scriptPubKey and a `0014` redeem record would pin both the classification and the scriptSig bytes

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
```diff
@@ -0,0 +1,424 @@
...
+        parsedTx.outputs.forEach { cursor.readMap() }
+
+        val inputs =
+            parsedTx.inputs.mapIndexed { index, txIn ->
+                makeBitcoinInput(inputMaps[index], index, txIn)
+            }
+        val outputs =
+            parsedTx.outputs.map { txOut ->
+                val scriptHex = Numeric.toHexStringNoPrefix(txOut.scriptPubKey)
+                val isChange =
```
rkokhatskyi, line 112: Change is detected only by an exact match against the single derived vault lock script while the binding requires the non change sum to equal `fromAmount`, whereas the golden fixture has one output and no change so neither the change marking nor a deposit that differs from the entered amount is covered, so a provider that returns change to a different vault owned script or a deposit of sell amount minus fee would be rejected, which is fail closed yet blocks a real swap, so this wants validation against the live NEAR Garden Flashnet routes with a change output present before the table entry is flipped

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
```diff
@@ -0,0 +1,424 @@
...
+        txIn: ParsedTxInput,
+    ): BitcoinInput {
+        val witnessUtxo =
+            inputMap[KEY_WITNESS_UTXO]
+                ?: throw SwapKitBtcSignerException(
+                    "SwapKit BTC PSBT input #$index is missing PSBT_IN_WITNESS_UTXO"
+                )
+        val (amount, scriptPubKey) = parseWitnessUtxo(witnessUtxo, index)
+        val (scriptType, redeemScript) = classifyScript(scriptPubKey, inputMap, index)
+        val sighashType =
```
rkokhatskyi, line 142: When the PSBT carries an explicit four byte sighash type of zero this sets `sighashType` to `0u` which `computeOurSighashes` then rejects since it allows only `SIGHASH_ALL` of 1, whereas iOS treats an explicit zero as the default `SIGHASH_ALL` and signs, so a provider that encodes the default sighash this way would be refused on Android only

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
```diff
@@ -93,6 +95,18 @@ object SigningHelper {
 
                     messages += message
                 }
+                is SwapPayload.SwapKit -> {
```
rkokhatskyi, line 97: This `require` on `txType` throws `IllegalArgumentException` inside `getKeysignMessages` and the compile arm, which unlike the `SwapFormViewModel` call site are not wrapped in a try, so a non PSBT SwapKit payload reaching the signer surfaces as a raw exception in the keysign engine rather than a typed signing error, and nothing pins this reject since there is no `SigningHelper` test

rkokhatskyi (changes requested):

aminsato:
Thanks for the thorough review. Addressed in f332c6507 — summary of what I applied vs. deferred, with reasoning:

**Applied**
- **#1 (trailing bytes after locktime)** — `parseUnsignedTx` now rejects a body that doesn't consume exactly, matching what `parseWitnessUtxo` already did.
- **#3 (compactSize `toInt()` truncation)** — added `PsbtCursor.asLength`, which throws a clear error above `Int.MAX_VALUE` instead of wrapping negative and surfacing as a misleading "truncated"; routed the map + script-length reads through it.
- **#5 (visibility)** — `decodeToSignBitcoin` is now `internal`, consistent with `decode`.
- **#6 (non-`is_ours` inputs)** — added `require(inputs.all { it.isOurs })` in compile. It's true by construction today (decode hardcodes `is_ours`), but now fails loudly instead of producing an empty-witness, non-final tx if that ever drifts.
- **#7 (P2SH-P2WPKH coverage)** — added a decode test exercising the redeem-script classification path (the golden vector is all P2WPKH).

**Deferred / declined**
- **#2 (UTXO plan-fee path)** — valid concern, but the proposed fix doesn't fit a pre-built PSBT. SwapKit returns the transaction already constructed; its fee is fixed as `inputs − outputs` (what `getBitcoinTransactionPlanFromSignBitcoin` computes). Running our coin-selection planner against a synthetic send to `targetAddress` would display an estimate that isn't the PSBT's actual fee. Sourcing the displayed fee from the PSBT (and the button guard) belongs to enabling the route in `SwapProviderTable`, which is gated on on-device E2E + review — the path is inert until then.
- **#4 (PSBT decoded twice)** — deterministic and microsecond-scale for a one-shot signing op; threading `SignBitcoin` through keysign state isn't worth it here.
- **#8 (trailing bytes after output maps)** — declined deliberately: BIP-174 permits proprietary/extension fields after the maps, so strict rejection there risks rejecting valid PSBTs. (#1 is different — the unsigned-tx body has no such allowance.)
- **#9 (`getSignedTransaction` unit test)** — `tss.KeysignResponse` is `compileOnly`, so it can't be referenced from a `:data` unit test; the compile/broadcast path is covered by the on-device E2E gate noted in the description.

aminsato:
Thanks @rkokhatskyi — went through all seven. Addressed the correctness + coverage ones in 804bb06e6; the rest are enablement-time items I've added to the checklist below.

**Fixed / added**
- **Explicit zero sighash** — good catch. An explicit 4-byte `0` was set to `0u` and then rejected by `computeOurSighashes` (which only allows `SIGHASH_ALL`). Now normalized `0 → null` so the `?: SIGHASH_ALL` default applies, matching iOS.
- **Compile path untested** — refactored `compileSignedTransaction` to take a plain `der-by-sighash` map + pubkey bytes (instead of the `compileOnly` TSS type) so the broadcast serialization is unit-testable. Added a **headless P2WPKH compile golden** that pins the exact raw segwit tx byte-for-byte (version, marker/flag, reversed outpoint, empty scriptSig, witness `[der||flag, pubkey]`, locktime).
- **P2SH-P2WPKH scriptSig untested** — added a (JNI-gated) compile test pinning the redeem-script push in the scriptSig.

**Deferred to enablement** (the route stays inert in `SwapProviderTable` until on-device E2E + review — added as checklist items)
- **PSBT fee display + submit guard** — agreed the displayed fee should come from the PSBT's `inputs − outputs`, not `getSpecificAndUtxo`'s coin-selection estimate, and that `utxoFeeData` being null would block submit once enabled. The correct fee source is the decoded PSBT; wiring it into `calculateFees` + the verify screen is part of flipping the table entry, where it can be validated against a real swap rather than guessed.
- **Miner-fee sanity bound** — reasonable as fail-closed, but the threshold is a judgment call I don't want to invent blind; deferring to the enablement pass with the real-route data.
- **Binding vs. real routes** — the exact-match change detection + `sum(non-change) == fromAmount` binding is deliberately fail-closed, but as you note it would reject a provider that funds the deposit as `sellAmount − fee` or returns change to a different owned script. This needs validation against real NEAR / Garden / Flashnet PSBTs (with a change output present) before enabling, and the binding may need loosening.

**Keeping as-is**
- **`require` throwing in `SigningHelper`** — this matches `SigningHelper`'s existing style (it uses raw `error(...)`/exceptions throughout), and the txType is already validated at quote-build; a non-PSBT SwapKit payload reaching the signer is a programming error, so failing loudly there is intended. Happy to wrap it in a typed signing error if you'd prefer consistency there.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
aminsato: Done in 804bb06e6. Refactored `compileSignedTransaction` to take a plain der-by-sighash map + pubkey bytes (the TSS `KeysignResponse` is `compileOnly`, which is why there was no prior test), then added a headless P2WPKH compile golden that pins the raw segwit tx byte-for-byte — version, marker/flag, reversed outpoint, empty scriptSig, witness `[der||sighashFlag, pubkey]`, locktime. Note the assertion is a hand-derived byte string rather than an iOS-pinned value: the iOS tests stop at the sighashes too (no compiled-tx golden), and the real compile/broadcast is still gated on on-device E2E.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Agreed — tracked as an enablement blocker (added to the PR description checklist). The displayed fee should come from the decoded PSBT's `inputs − outputs`, not `getSpecificAndUtxo`'s coin-selection estimate, and `utxoFeeData` being null would block submit once enabled. Since the route is inert until the `SwapProviderTable` entry is flipped, I'm wiring the PSBT fee + submit guard at enablement where it can be validated against a real swap rather than guessed.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
aminsato: Fair — a fail-closed bound on `inputs − outputs` is the right instinct. Deferring the actual threshold to the enablement pass (added to the checklist): picking it blind risks false rejections during high-fee periods, so I'd rather set it from real-route data.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
aminsato: Done in 804bb06e6. Added a P2SH-P2WPKH compile test that builds an `a914..87` scriptPubKey (with the correct `HASH160(redeem)`) plus a `0014` redeem record and asserts the `0x16`-prefixed redeem push lands in the scriptSig. It's JNI-gated since the P2SH commitment check needs `Hash.sha256RIPEMD`.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
aminsato: Tracked as an enablement blocker. The exact-match change detection + `sum(non-change) == fromAmount` is intentionally fail-closed, but as you note it would reject a provider that funds the deposit as `sellAmount − fee` or returns change to a different owned script. This needs validation against real NEAR / Garden / Flashnet PSBTs (with a change output present) before the table entry flips; the binding may need loosening then.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitBtcSigner.kt
aminsato: Fixed in 804bb06e6. An explicit 4-byte sighash of `0` now normalizes to null so the `?: SIGHASH_ALL` default applies, matching iOS' `effectiveSighashType` — it was previously refused on Android only.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
aminsato: Kept as-is: `SigningHelper` already throws raw `error(...)`/exceptions throughout, and the txType is validated at quote-build, so a non-PSBT SwapKit payload reaching the signer is a programming error worth failing loudly on. Happy to wrap it in a typed signing error if you'd prefer consistency. (A `SigningHelper`-level test needs a full `Vault`, hence none here — the signer paths are covered directly.)

rkokhatskyi (approved):
Re-reviewed the three new commits. The compile path now has a byte for byte P2WPKH segwit golden that I checked against the spec (version, marker, reversed outpoint, empty scriptSig, witness `[der||01, pubkey]`, LE amount, locktime) plus `p2sh-p2wpkh` decode and compile tests pinning the `0x16` redeem push, the explicit zero sighash normalizes to `SIGHASH_ALL` matching iOS, and `parseUnsignedTx` rejects trailing bytes while `asLength` guards the compactSize narrowing, all verified correct. The fee display, the miner fee bound, and validating the binding against a real change bearing PSBT are reasonably deferred to enablement and tracked in the description, so this is good for the inert signer.

PR #4677, opened 2026-05-28
fix(keysign): wire toToken image binding + refresh keysign.riv asset

### LINKED ISSUE #4676: Keysign animation: `toToken` image input not bound during signing
## Summary
During swap keysign the Rive animation's `toToken` ImageAsset input is never bound on Android, so the destination token image slot renders empty. The animation file exposes the input, but `KeysignRiveProgress` only sets `progessPercentage` and never calls `setImage("toToken", …)`.

## Expected
While swap keysign is in progress, the destination token image inside the Rive animation matches the swap's `toToken`.

## Actual
The destination-token image slot is empty (default/placeholder).

## Scope
- File: `app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt` (`KeysignRiveProgress`)
- Asset: `app/src/main/res/raw/riv_keysign.riv`
- Surface: keysign animation during swap signing.

## Cross-platform
Sibling fixes: vultisig-ios#4439 / vultisig-ios#4440, vultisig-windows#4010 / vultisig-windows#4011.

gastonm5:
## Summary
- **Code:** `KeysignRiveProgress` previously only set `progessPercentage` — nothing was binding the new `toToken` ImageAsset input the Rive file exposes, so the destination token slot rendered empty during swap keysign.
- Threads the destination `Coin` (`SwapTransactionUiModel.dst.token`) through `KeysignView` → `KeysignRiveProgress`, encodes its drawable as PNG bytes, decodes via `ImageAsset.fromBytes(riveWorker, …)`, and calls `vmi.setImage("toToken", asset)`. Asset is `close()`'d on disposal.
- **Asset:** `riv_keysign.riv` updated to the new Rive file that properly exposes the `toToken` image input.
- Parity with vultisig-ios#4440 and vultisig-windows#4011 (Windows already binds via `useKeysignLoadingAnimation.ts`; iOS asset-only).

Closes #4676

## Implementation notes
- New constant `RIVE_TO_TOKEN_IMAGE_PROPERTY = "toToken"`.
- Image binding is no-op for non-swap keysign (Send / Deposit / SignMessage) — `dstTokenLogoRes` stays `null`.
- PNG re-encode is small (token logos are bundled drawables, typically <10KB) and runs in `LaunchedEffect` keyed on `dstTokenLogoRes`.
- `Coin.tokenLogoRes()` (existing helper in `data/models/CoinLogo.kt`) returns a `@DrawableRes Int`, falling back to `chain.logo` for arbitrary tokens.

## Test plan
- [ ] Run a swap keysign on a real device (e.g. ETH→USDC via THORChain) and confirm the destination token image inside the Rive animation matches the swap's `toToken`.
- [ ] Verify the progress animation (`progessPercentage`) still advances normally — no regression.
- [ ] Run a non-swap keysign (Send) and confirm the animation still plays without errors (no `toToken` binding attempted).
- [ ] Smoke a swap-payload variant (LP add / function call as `Swap`) to confirm `swapTransactionUiModel.dst.token` is present.

## Screenshots
This change only affects the destination-token slot inside an animated Rive `ViewModelInstance`. `RiveAnimation` renders a `Spacer` under `LocalInspectionMode`, so it cannot be captured via PreviewActivity. Verification requires a live swap keysign on device — I'll defer the before/after capture to that run.

## Summary by CodeRabbit

* **New Features**
  * Progress animation can display the destination token logo for clearer visual feedback.
  * Token logo source is now provided by the view model and falls back to the chain logo when unavailable.
  * Token images are rasterized and bound into the progress animation with lifecycle-aware management to ensure temporary image resources are released.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4677?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


johnnyluo (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt
```diff
@@ -137,6 +165,23 @@ private fun KeysignRiveProgress(progress: Float) {
 
     SideEffect { vmi.setNumber(RIVE_PROGRESS_PROPERTY, animatedValue) }
 
+    val context = LocalContext.current
+    val riveWorker = rememberRiveWorkerOrNull()
+    var toTokenAsset by remember { mutableStateOf<ImageAsset?>(null) }
+
+    LaunchedEffect(dstTokenLogoRes, riveWorker, vmi) {
+        if (dstTokenLogoRes == null || riveWorker == null) return@LaunchedEffect
+        val bytes = encodeDrawableAsPng(context, dstTokenLogoRes) ?: return@LaunchedEffect
```
johnnyluo: **Issue 1 — Main-thread bitmap work**

`encodeDrawableAsPng` does `BitmapFactory.decodeResource` + `bitmap.compress` here on the main thread (LaunchedEffect default dispatcher). Token logos are small so it's unlikely to jank, but it's still a blocking call on the UI thread. Consider wrapping with `withContext(Dispatchers.Default)`:

```kotlin
val bytes = withContext(Dispatchers.Default) {
    encodeDrawableAsPng(context, dstTokenLogoRes)
} ?: return@LaunchedEffect
```

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt
```diff
@@ -137,6 +165,23 @@ private fun KeysignRiveProgress(progress: Float) {
...
+        val result = ImageAsset.fromBytes(riveWorker, bytes)
+        if (result is Result.Success) {
```
johnnyluo, line 186: **Issue 3 — Silent failure on `ImageAsset.fromBytes` error**

When `result` is not `Result.Success` there's no logging, making failures invisible during debugging. A `Timber.w` on the else branch would help:

```kotlin
if (result is Result.Success) {
    toTokenAsset = result.value
    vmi.setImage(RIVE_TO_TOKEN_IMAGE_PROPERTY, result.value)
} else {
    Timber.w("Failed to load toToken image asset for res %d", dstTokenLogoRes)
}
```

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt
```diff
@@ -145,6 +190,21 @@ private fun KeysignRiveProgress(progress: Float) {
     )
 }
 
+private fun encodeDrawableAsPng(
+    context: android.content.Context,
```
johnnyluo: **Issue 2 — Fully-qualified `android.content.Context`**

`android.content.Context` is used as a fully-qualified name here instead of an import. Should be:

```kotlin
import android.content.Context
// ...
private fun encodeDrawableAsPng(
    context: Context,
```

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt
```diff
@@ -137,6 +165,23 @@ private fun KeysignRiveProgress(progress: Float) {
...
+        val bytes = encodeDrawableAsPng(context, dstTokenLogoRes) ?: return@LaunchedEffect
+        val result = ImageAsset.fromBytes(riveWorker, bytes)
+        if (result is Result.Success) {
+            toTokenAsset = result.value
+            vmi.setImage(RIVE_TO_TOKEN_IMAGE_PROPERTY, result.value)
+        }
+    }
+    DisposableEffect(toTokenAsset) {
+        onDispose { toTokenAsset?.close() }
+    }
```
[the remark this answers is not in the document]
gastonm5: Done in b0b81312b — the `DisposableEffect` now captures the current instance before disposal:
```kotlin
DisposableEffect(toTokenAsset) {
    val assetToDispose = toTokenAsset
    onDispose { assetToDispose?.close() }
}
```

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt
gastonm5: Done in b0b81312b — the encode now runs off the UI thread: `withContext(Dispatchers.Default) { encodeDrawableAsPng(context, coinLogoRes) }`.

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt
gastonm5: Done in b0b81312b — failures are now logged: `Timber.w("Failed to load toToken image asset for res %d", coinLogoRes)` on the non-Success branch, plus a warning when the drawable can't be encoded.

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt
gastonm5: Done in b0b81312b — added `import android.content.Context`; the parameter is now `context: Context`.

gastonm5:
## Update — addressed review feedback + fixed the logo not rendering at all

While verifying on device, the `toToken` logo still didn't appear even after the review fixes. Three separate bugs were each blocking it — all now fixed:

1. **Vector drawables** (`b0b81312b`) — `BitmapFactory.decodeResource` returns `null` for vector-XML drawables, which is what most token logos are, so `encodeDrawableAsPng` always failed silently. Now loads via `AppCompatResources.getDrawable` + `toBitmap()` (handles vector + raster).

2. **Rive worker mismatch** (`39d6b4929`) — `rememberRiveWorkerOrNull()` returns a *per-call-site* `RiveWorker`, so the image was decoded on a different worker than the one that owns the `ViewModelInstance`. `ImageAsset` handles are worker-scoped, so `vmi.setImage(...)` silently no-op'd. Now decodes on `riveFile.riveWorker` (the worker that owns the vmi). `setNumber` worked all along because it passes a plain value, not a handle.

3. **Logo source + non-swap case** (`9ed879a7c`) — the logo was sourced from `transactionTypeUiModel`, which loads asynchronously and is usually still `null` while signing, and only for swaps — so plain sends (e.g. sending RUNE) showed nothing. Now sourced synchronously from `keysignPayload.coin` as `KeysignViewModel.coinLogoRes` and shown for **every** keysign (Send/Swap/Deposit), matching iOS (`keysignPayload.coin.logo`) and Windows (`getKeysignPayloadLogoSrc`), with `tokenLogoRes()` falling back to the chain logo. The `dstTokenLogoRes` param was renamed to `coinLogoRes` accordingly.

Verified on device: sending RUNE now shows the RUNE logo in the keysign animation. ✅

All four review comments are addressed (replies inline). Rebased onto latest `main`.

> Heads-up for a follow-up: `KeysignPeerDiscovery.kt` has the same vector-drawable issue as #1 for its `dstTokenLogoRes` (`BitmapFactory.decodeResource`) — out of scope here.

johnnyluo (approved):
LGTM

johnnyluo (approved):
LGTM

PR #4678, opened 2026-05-29
feat(swap): enable SwapKit on Bitcoin in provider table

aminsato:
## Summary
- Add `SwapProvider.SWAPKIT` to the `Chain.Bitcoin` branch in `SwapProviderTable` so BTC swaps surface SwapKit as a candidate provider.
- This is the enablement switch that makes the already-merged BTC/PSBT work reachable end to end: the `/v3/quote`+PSBT decode, the `SwapKitBtcSigner` (PSBT → SignBitcoin, BIP-143 sighashes, segwit compile), and the UTXO fee-planning wiring in `SwapFormViewModel`. Until now BTC routed only through THORCHAIN/MAYA, so that code was inert.
- Updates `SwapProviderTableTest`: BTC moves from the negative to the positive SWAPKIT assertion.

### Tranasction hash
https://mempool.space/tx/72c797d4ea6b648930ab8d8c521d58401a7d435e7b0f1239866273a2dbd2b403

## Test plan
- [x] `SwapProviderTableTest` — green
- [x] `SwapQuoteRepositoryTest` — green (SOL→BTC still resolves to THORCHAIN; THORCHAIN precedes SWAPKIT in the provider-set order)
- [x] `SwapFormViewModelTest` — green
- [ ] **On-device E2E**: build a real BTC→<token> SwapKit quote, sign, and broadcast — the compile/broadcast path is not unit-testable (TSS lib is `compileOnly`), so this must be verified on a device/emulator before merge.

Base branch: `feature/swapkit-btc`.

## Summary by CodeRabbit

* **New Features**
  * SwapKit is now available as a swap provider option for Bitcoin transactions, providing users with an additional trading route alongside existing providers and enhancing swap flexibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4678?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


aminsato:
@coderabbitai full review

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapProviderTable.kt
```diff
@@ -97,7 +97,7 @@ internal class SwapProviderTableImpl @Inject constructor() : SwapProviderTable {
             Chain.Mantle -> setOf(SwapProvider.LIFI, SwapProvider.KYBER)
 
             Chain.ThorChain -> setOf(SwapProvider.THORCHAIN, SwapProvider.MAYA)
-            Chain.Bitcoin -> setOf(SwapProvider.THORCHAIN, SwapProvider.MAYA)
+            Chain.Bitcoin -> setOf(SwapProvider.THORCHAIN, SwapProvider.MAYA, SwapProvider.SWAPKIT)
```
rkokhatskyi, line 100: `verifyDestinationBinding` requires `sumNonChange == fromAmount` with change pinned to the one vault script so a provider PSBT that nets the fee from the deposit fails at sign time, whereas iOS has no such binding and the real `GOLDEN_BTC_PSBT_B64` fixture (one 12466 sat output, no change) only runs through `decode` not this check

rkokhatskyi, line 100: The Verify screen fee for a BTC SwapKit swap is `computeUtxoPlanFeeResult` estimating a hypothetical send (`SwapFormViewModel.kt:1049`) not the broadcast PSBT's own miner fee, and `collectTotalFee` then adds the BTC inbound fee on top so the total double counts the network cost whereas iOS shows the single `/v3/swap` inbound fee

rkokhatskyi, line 100: Nothing bounds the PSBT miner fee before signing since `verifyDestinationBinding` pins only `sumNonChange == fromAmount` and the change script but not the change amount, so a PSBT with little change against large inputs is signed with an arbitrary fee the user never sees, so add a max fee or sat per vbyte ceiling before signing

rkokhatskyi, line 100: This table entry is the only gate for BTC SwapKit on Android since `SwapKitQuoteSource` checks just the default on opt out flag and skips the provider cache (`SwapKitQuoteSource.kt:93`), whereas iOS refuses `/v3/quote` for a chain absent from `enabledChainIds` (`SwapService.swift:368`) so here the route goes live for every user once this reaches main

data/src/test/kotlin/com/vultisig/wallet/data/repositories/swap/SwapProviderTableTest.kt
```diff
@@ -38,6 +38,7 @@ internal class SwapProviderTableTest {
                 coin(Chain.Arbitrum, "ARB", isNative = false), // maya-eligible Arbitrum token
                 coin(Chain.Solana, "SOL", isNative = true),
                 coin(Chain.Solana, "USDC", isNative = false),
+                coin(Chain.Bitcoin, "BTC", isNative = true), // BTC PSBT route
```
rkokhatskyi, line 41: Selection is best output via `fetchBestQuote` using `maxBy { estimatedDstFiat }` (`SwapQuoteManager.kt:328`) not provider set order, so SwapKit wins any BTC pair it prices higher, whereas this test pins only set membership and the green `SwapQuoteRepositoryTest` SOL to BTC asserts `firstOrNull()` which is list order not the selection

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapProviderTable.kt
aminsato: Fixed in 8017bfbc. `verifyDestinationBinding` no longer requires `sumNonChange == toAmount`. It now allows the deposit to be at most `toAmount` with any shortfall capped at the actual miner fee, so a provider PSBT that nets the fee out of the deposit (single output, no change — the `GOLDEN_BTC_PSBT_B64` shape) signs instead of aborting. The single-output-to-`targetAddress` and change-script pins stay, and the new fee ceiling bounds the shortfall. Added a `UtxoHelperTest` for the netted-fee deposit. On-device E2E against a live `/v3/swap` PSBT is still needed before merge to confirm the real provider shape.

aminsato: Fixed in 8017bfbc. For SwapKit UTXO swaps the inbound fee is the same BTC deposit cost already surfaced as the UTXO plan network fee, so it is no longer added again in `collectTotalFee` (nor shown as a separate fee row). The headline total now reconciles to the network fee alone — a single network cost, matching iOS. THOR/MAYA are unchanged since their protocol fees are genuinely separate from the BTC tx the user broadcasts.

aminsato: Added in 8017bfbc. `verifyOwnership` now computes the actual fee (`sum(inputs) - sum(outputs)`) and rejects anything above a sat/vByte ceiling (`MAX_FEE_SAT_PER_VBYTE`) using a coarse segwit vsize estimate, covering both the SwapKit BTC and generic `signBitcoin` co-sign paths. The ceiling is generous enough never to reject a normal deposit but catches the large-inputs/tiny-change case where the fee is unbounded and never shown. Test added.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/swap/SwapProviderTableTest.kt
aminsato: Agreed — selection is `fetchBestQuote`'s `maxBy { estimatedDstFiat }`, not provider-set order, and this `SOL→BTC` test pins `firstOrNull()` on the eligibility list rather than the runtime choice. Added a `SwapQuoteManagerTest` in 8017bfbc proving SWAPKIT wins a BTC pair it prices higher even when listed after THORCHAIN. Will also correct the set-order wording in the PR description.

PR #4679, opened 2026-05-29
fix(keysign): bound toToken logo size to prevent Rive OOM crash

aminsato:
## Problem

The app crashes hard (SIGABRT, whole process killed) during keysign — reproducible while signing a swap. The native abort is on the **Rive CmdServer** thread:

```
java.lang.OutOfMemoryError: Failed to allocate a 462422040 byte allocation ... growth limit 201326592
    app.rive.runtime.kotlin.core.ImageDecoder.decodeToBitmap(ImageDecoder.kt:25)
...
JNI DETECTED ERROR IN APPLICATION: JNI NewStringUTF called with pending exception java.lang.OutOfMemoryError
runtime.cc: Runtime aborting...
Fatal signal 6 (SIGABRT) in tid (Rive CmdServer)
```

## Root cause

Introduced by #4677 (toToken image binding). `KeysignRiveProgress` rasterizes the signing coin's logo via `encodeDrawableAsPng` and hands the PNG to Rive (`ImageAsset.fromBytes` → `setImage("toToken", …)`).

`encodeDrawableAsPng` called `drawable.toBitmap()` with **no size bound**, so the bitmap took the drawable's intrinsic dimensions. Rive then decodes those bytes through `ImageDecoder.decodeToBitmap`, which does:

```kotlin
val pixels = IntArray(offset + width * height)   // ImageDecoder.kt:24 — Java heap
```

`462422040 / 4 ≈ 115.6M ints ≈ 10752×10752`. The decoded bitmap itself lives in **native** memory (so the decode succeeds), but this `IntArray` of `width*height` ints is allocated on the **Java heap** (192 MB growth limit) → `OutOfMemoryError`.

The kicker: Rive's decode wraps the body in `catch (e: Exception)`, but `OutOfMemoryError` is an `Error`, **not** an `Exception` — so it escapes. It becomes a pending JNI exception, Rive's native logging path then calls `NewStringUTF` with that exception pending, and ART's CheckJNI aborts the whole process. We can't fix the `.aar`, so we must never hand Rive an oversized image.

## Fix

Bound the rasterized logo to **256 px** (aspect ratio preserved) in `encodeDrawableAsPng` before encoding. The toToken slot in the animation is small, so 256 px is plenty, and the `IntArray` Rive allocates is now ~256 KB regardless of the drawable's intrinsic bounds. Also:
- handle missing intrinsic bounds (vector drawables can report `-1`),
- pin the config to `ARGB_8888`.

## Test plan
- [x] `:app:compileDebugKotlin` passes; ktfmt clean
- [x] Ran keysign on the emulator — signing animation renders the coin logo and no longer crashes (verified on-device)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced a maximum token logo size and preserved aspect ratio on the keysign screen to avoid oversized images.

* **Performance**
  * Improved image rasterization and encoding to reduce peak memory usage and stabilize rendering.

* **Stability**
  * Added safe error handling for image processing to prevent crashes and return a safe fallback.

* **Tests**
  * Added regression tests for logo sizing, clamping, aspect-ratio handling, and fallback behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/vultisig/vultisig-android/pull/4679?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt
@@ -51,6 +51,14 @@ import timber.log.Timber
 private const val RIVE_PROGRESS_PROPERTY = "progessPercentage" // typo in riv_keysign.riv
 private const val RIVE_TO_TOKEN_IMAGE_PROPERTY = "toToken"
 
+// The toToken logo only occupies a small slot in the animation, so cap the rasterized size.
+// Rive decodes the bytes we hand it via ImageDecoder.decodeToBitmap, which copies the image into
+// an IntArray (width * height ints) on the Java heap. A drawable with large/abnormal intrinsic
+// bounds would balloon that allocation past the heap limit and throw an OutOfMemoryError — which
+// Rive's native decode path does not catch (it only catches Exception), so it aborts the whole
+// process. Bounding the size keeps that allocation tiny and well within the heap.
+private const val RIVE_TO_TOKEN_IMAGE_MAX_PX = 256
+
 @Composable
 internal fun KeysignView(
     state: KeysignState,
@@ -203,10 +211,17 @@ private fun KeysignRiveProgress(progress: Float, @DrawableRes coinLogoRes: Int?)
     )
 }
 
-/** Rasterizes a (vector or raster) drawable resource to PNG bytes for Rive ImageAsset binding. */
+/**
+ * Rasterizes a (vector or raster) drawable resource to PNG bytes for Rive ImageAsset binding.
+ *
+ * The rasterized size is bounded by [RIVE_TO_TOKEN_IMAGE_MAX_PX] (aspect ratio preserved) so the
+ * image we hand to Rive can never produce an oversized decode allocation — see the constant for why
+ * an unbounded size crashes the process.
+ */
 private fun encodeDrawableAsPng(context: Context, @DrawableRes resId: Int): ByteArray? {
     val drawable = AppCompatResources.getDrawable(context, resId) ?: return null
-    val bitmap = drawable.toBitmap()
+    val (width, height) = boundedLogoSize(drawable.intrinsicWidth, drawable.intrinsicHeight)
+    val bitmap = drawable.toBitmap(width = width, height = height, config = Bitmap.Config.ARGB_8888)
     return try {
         ByteArrayOutputStream().use { stream ->
             bitmap.compress(Bitmap.CompressFormat.PNG, 100, stream)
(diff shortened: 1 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitPsbtParser.kt
```diff
@@ -0,0 +1,157 @@
...
+ * Shared BIP-174 PSBT framing primitives. Ported from iOS' `SwapKitPSBTParser`. Every SwapKit UTXO
+ * signer consumes the same byte-level envelope — magic prefix, global key/value map, per-input map,
+ * per-output map. Per-chain signers diverge only on the unsigned-tx body parser (which lives with
+ * the chain's serialization rules in the per-chain signer) and on script-type classification.
+ *
+ * Pure bytes: no crypto and no vault context. The unsigned-tx body (global key `0x00`) is exposed
+ * as raw bytes; the caller parses it because the serialization differs per chain (BIP-144 segwit
+ * for BTC, legacy for DOGE/BCH/DASH, Sapling-v4 for ZEC).
+ */
+internal class SwapKitPsbtException(message: String) : Exception(message)
```
rkokhatskyi: SwapKit is not supposed to be here, please cleanup the PR from unnecessary changes

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitPsbtParser.kt
aminsato: Done

aminsato: Cleaned up. This branch was accidentally cut from `aminsato/swapkit_btc_provider_table`, so it dragged in all the SwapKit BTC commits. I've rebased it onto `main` — the PR now contains only the keysign Rive OOM fix (`Keysign.kt`, +36/−2), and `SwapKitPsbtParser.kt` and the rest of the SwapKit files are no longer part of the diff. Thanks for flagging it.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -326,10 +365,85 @@ constructor(
...
+        val targetAddress =
+            response.targetAddress?.takeIf { it.isNotBlank() }
+                ?: throw SwapKitError.Decoding("SwapKit non-EVM response missing targetAddress")
+        val payload =
+            SwapKitSwapPayloadJson(
+                fromCoin = srcToken,
+                toCoin = dstToken,
+                fromAmount = request.tokenValue.value,
+                toAmountDecimal = toAmountDecimal,
+                txType = response.meta.txType,
```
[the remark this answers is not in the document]
aminsato: `SwapKitQuoteSource.kt` is no longer part of this PR — it was rebased out, so this PR now only contains the keysign Rive OOM fix. This `txType` → `TX_TYPE_PSBT` normalization belongs on the SwapKit BTC PR and should be addressed there, not here.

aminsato:
### Re: CodeRabbit review

All of CodeRabbit's findings (the actionable `txType` comment on `SwapKitQuoteSource.kt` and the nitpicks on `SwapKitPsbtParser.kt`) are on **SwapKit BTC files that have been rebased out of this PR**. After the rebase onto `main`, this PR touches a single file — `app/src/main/java/com/vultisig/wallet/ui/screens/keysign/Keysign.kt` — and contains only the keysign Rive OOM fix.

Those suggestions should be tracked on the SwapKit BTC PR (`aminsato/swapkit_btc_provider_table`), where the code actually lives. Nothing CodeRabbit raised applies to the keysign change in this PR.

aminsato:
@coderabbitai ful review

aminsato:
### Re: [review](https://github.com/vultisig/vultisig-android/pull/4679#pullrequestreview-4391619686) — `encodeDrawableAsPng` exception handling

Good call, applied in 19f2e1cd5. `toBitmap()`/`compress()`/`toByteArray()` are now wrapped in a try/catch that logs and returns `null` (the call site already treats `null` as "no logo"), so a failure can't escape into the `LaunchedEffect` coroutine and crash the screen. `bitmap.recycle()` still runs via the inner `finally`. This is in the same spirit as the PR itself — hardening the keysign animation against crashes.

aminsato:
### Re: [review](https://github.com/vultisig/vultisig-android/pull/4679#pullrequestreview-4392060503) — symmetric aspect-ratio test

Added in the latest commit: `boundedLogoSize(512, 1024) shouldBe (128 to 256)`, the inverse (portrait) of the existing landscape case, so both downscale directions are pinned. `KeysignLogoSizeTest` is now 7 cases, all green.

johnnyluo (approved):
LGTM

PR #4688, opened 2026-05-31
feat(swap): SwapKit TRON signing support

aminsato:
## Summary

Wires SwapKit's **TRON** swap path end-to-end on Android, mirroring the iOS implementation. Previously Android's SwapKit support covered EVM + Solana + Bitcoin PSBT only; TRON (and the other non-EVM txTypes) surfaced as `UnsupportedTxType`.

SwapKit's `/v3/swap` returns a TronWeb-shaped JSON **object** (`{txID, raw_data, raw_data_hex, …}`) rather than a base64 blob. The canonical Tron signing input is `sha256(raw_data_bytes)` — which is also Tron's `txID` — so the signer hashes the pre-built `raw_data_hex` directly and re-assembles the broadcast envelope by appending the 65-byte `r||s||v` signature as `signature:[hex]`. This keeps the surface small (no per-`contract.type` `SigningInput` reconstruction) and gives the cosigning peer identical bytes to verify. `TronApi.broadcastTransaction` posts that envelope verbatim.

### Transaction hash
https://tronscan.org/#/transaction/e0865bfd91e5707cb5b3bebe58c7a0844d3e58229cbfa5d5cc3c2c11a0b30d76

## Changes

- **`SwapKitTronSigner`** (new): signing digest, single-entry pre-signed hash, and signed-broadcast-envelope assembly with signature verification against the derived vault key.
- **`SwapKitSwapPayloadJson`**: added `TX_TYPE_TRON` discriminator.
- **`SigningHelper`**: the PSBT-only `require` gate in both the pre-sign and signed-tx paths is now a `when(txType)` dispatch (PSBT → `SwapKitBtcSigner`, TRON → `SwapKitTronSigner`).
- **`SwapKitQuoteSource`**: TRON added to the tx-kind dispatch, `chainPrefix(Tron) → "TRON"`, native decimals `6`, and a JSON-object `txPayload` encoder (TRON's `tx` is an object, not a base64 string).
- **`SwapProviderTable`**: `Chain.Tron` now offers `SWAPKIT` alongside `THORCHAIN` (mirrors iOS `.tron → [thorchain, swapkit]`).
- **`SwapFormViewModel`**: the SwapKit gate accepts PSBT or TRON.

## Test plan

- New `SwapKitTronSignerTest`: digest pinned to the real `v3-tron-final-swap-fresh` fixture's reported `txID`, single-entry pre-sign hash, envelope appends signature + preserves fields, and rejects malformed / empty / missing-`raw_data_hex` payloads.
- New `SwapKitQuoteSourceTest` cases: TRON route decodes into a Native `SwapQuote.SwapKit` with the TronWeb object round-tripped into `txPayload` and the inbound TRX fee scaled by 6 native decimals; a non-object TRON `tx` throws `Decoding`.
- Updated existing tests that encoded "TRON unsupported" (`SwapProviderTableTest`, the unwired-txType `SwapKitQuoteSourceTest` case now uses SUI).
- `./gradlew :data:testDebugUnitTest` green; `./gradlew :app:compileDebugKotlin` green.

## Notes

- secp256k1 signature verification needs the WalletCore JNI, so the headless unit tests cover the digest + envelope logic; the verify path is exercised by the iOS port + integration (same constraint as the existing BTC PSBT signer).
- On-device E2E of a real TRON swap (compile + broadcast) is still pending, as with the BTC path.
- Remaining iOS parity gaps after this: TON, ADA, SUI (+ legacy-P2PKH UTXO chains and XRP).

## Summary by CodeRabbit

* **New Features**
  * Added TRON support for SwapKit swaps: TRX and TRC-20 assets can route through SwapKit.
  * Implemented TRON-specific signing and broadcast handling so TRON transactions can be prepared and submitted.

* **Bug Fixes / Improvements**
  * Expanded transaction type handling to accept both PSBT and TRON payloads; preserved original payload fields during processing.
  * Adjusted native fee scaling and chain mapping for TRON.

* **Tests**
  * Added unit tests covering TRON signing, payload parsing, and quote decoding.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -475,11 +475,14 @@ constructor(
                             }
 
                             is SwapQuote.SwapKit -> {
-                                // Only BTC PSBT is wired today; other SwapKit txTypes (TON / ADA /
-                                // SUI / TRON) land with their per-chain signers. SwapProviderTable
-                                // does not yet offer SwapKit on Bitcoin, so this branch is reached
-                                // only once enablement ships — guarded loudly until then.
-                                require(quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_PSBT) {
+                                // BTC PSBT and TRON (TronWeb object) are wired; the remaining
+                                // SwapKit txTypes (TON / ADA / SUI) land with their per-chain
+                                // signers. Guarded loudly so an un-wired txType can't reach
+                                // signing.
+                                require(
+                                    quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_PSBT ||
+                                        quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_TRON
+                                ) {
                                     "Unsupported SwapKit txType for swap: ${quote.data.txType}"
                                 }
                                 val specificAndUtxo =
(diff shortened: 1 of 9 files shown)


aminsato:
Thanks @coderabbitai — addressed all three actionable items plus the nitpick in 159a5ed5e:

- **`raw_data_hex` validation in `SwapKitQuoteSource`**: now extracts it as a non-blank string at quote time (was a key-presence check only), so a `null`/non-string/empty value throws `Decoding` in the quote path and selection falls back cleanly rather than failing at keysign. Added a `blank raw_data_hex` test.
- **Fee-scaling test**: source token now uses `decimal = 18` so a regression scaling the inbound fee by `srcToken.decimal` instead of TRON's native 6 would fail the assertion.
- **`!!` in JSON assertions**: replaced with `requireNotNull`/safe lookups in both `SwapKitTronSignerTest` and `SwapKitQuoteSourceTest`.

`:data:testDebugUnitTest` green.

rkokhatskyi (approved):
LGTM

PR #4704, opened 2026-06-01
feat(swap): SwapKit XRP signer (Payment + destination tag) (#4701)

### LINKED ISSUE #4701: SwapKit: XRP signer (Payment + destination tag)
Split from #4675 — SwapKit per-chain signer for **XRP**.

XRP source via SwapKit/NEAR — sign an XRP `Payment` to the NEAR deposit address, with **destination-tag** plumbing (defensive: NEAR's ephemeral-address pattern doesn't need a tag, but a future Chainflip shared-vault flip would). Mirror iOS.

### Work
- Add the XRP SwapKit signing path (reuse `RippleHelper` where possible); thread destination tag.
- Wire `TX_TYPE_*` in `SwapKitSwapPayloadJson` / `SwapKitQuoteSource` / `SigningHelper`.
- `SwapProviderTable`: `Chain.Ripple -> +SWAPKIT`; `isSwapSupported` already includes Ripple; chain prefix `"XRP"`.

aminsato:
Closes #4701.

## Summary

Wires SwapKit's **XRP** swap path end-to-end on Android, mirroring iOS. Unlike BTC/TRON/SUI (which carry a pre-built tx and a dedicated signer), **XRP is deposit-only**: SwapKit returns no transaction body — only a deposit r-address and an optional destination tag. There is no SwapKit signer; the cosigning peer rebuilds a plain XRP Payment to the deposit address for the sell amount and signs it with the existing `RippleHelper` (which already turns a numeric `memo` into the Payment's `destinationTag`).

### Tranasction hash:
https://xrpscan.com/tx/B64334E044B7DDF09A3CE75898B0D51AD030DD3E789CDC85D4CBC0F339EE59A5

## Changes

- **`SwapKitSwapPayloadJson`**: added `TX_TYPE_XRP`. The payload carries an empty `txPayload` plus the deposit address (`targetAddress`) and the destination tag (in `memo`).
- **`SwapKitSwapResponseJson`**: added top-level + `meta` `destinationTag` (decoded defensively as number-or-string), and `tx` now defaults to `JsonNull` so deposit-only responses that omit the body still decode.
- **`SwapKitQuoteSource`**:
  - `XRP` tx-kind (accepts both `XRP` and `RIPPLE`), `chainPrefix(Ripple) → "XRP"`, native decimals `6`.
  - Empty `txPayload` for XRP (no body).
  - 3-source destination-tag resolution — top-level field → `meta` → `?dt=`/`|` suffix on the address — with the suffix stripped off the bare r-address used as the payment destination (mirrors iOS' `resolvedDestinationTag` / `extractTagSuffix`).
- **`SigningHelper`**: XRP arm reuses `RippleHelper.getPreSignedImageHash` / `getSignedTransaction` in both dispatch sites (matches iOS letting XRP fall through to the per-chain helper).
- **`SwapProviderTable`**: `Chain.Ripple` now offers `SWAPKIT` (mirrors iOS `.ripple → [thorchain, swapkit]`). `Chain.Ripple` is already in `isSwapSupported`, so the Swap button shows.
- **`SwapFormViewModel`**: the SwapKit gate accepts XRP.

## Test plan

- New `SwapKitQuoteSourceTest` cases: an XRP route decodes into a deposit-only Native `SwapQuote.SwapKit` (empty `txPayload`, fee scaled by 6 decimals); the destination tag resolves from the top-level field, from `meta` (string-encoded), and from the `?dt=` address suffix (with stripping).
- Updated `SwapProviderTableTest` (Ripple moved into the SwapKit list + the swap-supported pin).
- `./gradlew :data:testDebugUnitTest` green; `:app` swap tests + both `compileDebugKotlin` green.

## Notes

- No new signer — XRP reuses the existing `RippleHelper`, so there's no JNI-only seam to worry about; the deposit Payment is the same one the normal XRP send path already produces and broadcasts.
- Live routability for XRP on the SwapKit proxy is subject to the same upstream/provider liquidity state noted for Sui; the wiring is correct and dormant until SwapKit serves XRP routes.
- Remaining iOS parity gaps after this: TON, ADA (Cardano — in progress by someone else), and the legacy-P2PKH UTXO chains (DOGE/BCH/DASH/ZEC/LTC).

## Summary by CodeRabbit

* **New Features**
  * Added XRP (Ripple) support for SwapKit routes as a deposit-only on‑ramp; destination-tag extraction and memo resolution improved; Swap availability updated so Ripple appears as a supported provider.

* **Tests**
  * Added unit tests covering XRP quote decoding, empty/absent transaction payloads, destination-tag parsing/validation, address-suffix handling, and provider availability.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
@@ -56,11 +56,13 @@ import timber.log.Timber
  * Non-EVM/non-Solana `txType`s surface as a fully-formed [SwapQuote.SwapKit] for a per-chain signer
  * to consume: PSBT (Bitcoin, via [com.vultisig.wallet.data.chains.helpers.SwapKitBtcSigner]), TRON
  * (TronWeb object, via [com.vultisig.wallet.data.chains.helpers.SwapKitTronSigner]), and SUI
- * (base64 PTB, via [com.vultisig.wallet.data.chains.helpers.SwapKitSuiSigner]) are wired today. The
- * remaining types (TON, CARDANO, RIPPLE, …) still surface as [SwapKitError.UnsupportedTxType] so
- * the picker falls back to another provider rather than signing garbage, until their signers land.
- * `SOLANA` and the legacy `SERIALIZED_BASE64` discriminator are aliased; SwapKit flipped them once
- * before.
+ * (base64 PTB, via [com.vultisig.wallet.data.chains.helpers.SwapKitSuiSigner]) are wired today. XRP
+ * is deposit-only — SwapKit returns no tx body, so the payload carries an empty [txPayload] plus
+ * the deposit address + destination tag and the cosigning peer rebuilds a plain XRP Payment via the
+ * existing `RippleHelper`. The remaining types (TON, CARDANO, …) still surface as
+ * [SwapKitError.UnsupportedTxType] so the picker falls back to another provider rather than signing
+ * garbage, until their signers land. `SOLANA` and the legacy `SERIALIZED_BASE64` discriminator are
+ * aliased; SwapKit flipped them once before.
  */
 internal class SwapKitQuoteSource
 @Inject
@@ -177,7 +179,8 @@ constructor(
                 )
             TxKind.PSBT,
             TxKind.TRON,
-            TxKind.SUI ->
+            TxKind.SUI,
+            TxKind.XRP ->
                 SwapQuoteResult.Native(
                     buildSwapKitNativeQuote(swapResponse, request, subProvider, best.fees)
                 )
@@ -265,6 +268,7 @@ constructor(
             Chain.Sui -> 9
             Chain.Bitcoin -> 8
             Chain.Tron -> 6
+            Chain.Ripple -> 6
             else -> 18
         }
 
@@ -372,11 +376,12 @@ constructor(
                         ),
                 )
             }
-            // PSBT/TRON/SUI are dispatched to buildSwapKitNativeQuote before reaching here; the arm
-            // keeps the `when` exhaustive and refuses anything that slips through.
+            // PSBT/TRON/SUI/XRP are dispatched to buildSwapKitNativeQuote before reaching here; the
+            // arm keeps the `when` exhaustive and refuses anything that slips through.
             TxKind.PSBT,
             TxKind.TRON,
             TxKind.SUI,
+            TxKind.XRP,
             TxKind.UNSUPPORTED -> throw SwapKitError.UnsupportedTxType(response.meta.txType)
         }
     }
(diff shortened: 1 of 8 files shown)


aminsato:
Fixed a crash found in on-device testing (a2d0538cd).

**Symptom:** swapping XRP→TRX, the quote loaded fine via SwapKit (NEAR), but pressing **Swap** threw `IllegalArgumentException: Unsupported SwapKit txType for swap: RIPPLE`.

**Cause:** SwapKit returns the XRP deposit-only route with `meta.txType` set to **either** `XRP` **or** `RIPPLE`. `txTypeOf` already maps both to the XRP kind, but `buildSwapKitNativeQuote` was persisting `meta.txType` *verbatim* onto the payload — so a `RIPPLE` response stored `txType="RIPPLE"`, which the signing-side gate (matching `TX_TYPE_XRP`) then rejected.

**Fix:** store the canonical discriminator via a new `canonicalTxType()` mapping (`RIPPLE`/`XRP` → `TX_TYPE_XRP`), mirroring iOS' normalisation in `buildSwapKitRipplePayload`. Pinned by a `RIPPLE`-wire regression test.

Note: the `THORChain "trading is halted"` 400s in the same logcat are unrelated — that's just the THORCHAIN provider failing upstream; SwapKit/NEAR served the actual quote.

aminsato:
Thanks @coderabbitai — addressed in 50d24566b.

`extractTagSuffix` now **strip-or-fails** on the `|` form: a `|` unambiguously announces a destination tag, so a non-numeric suffix (e.g. `rAddr|abc`) throws `SwapKitError.Decoding` instead of passing the full string through as the XRP destination — which would be unsignable, and for XRP silently dropping a tag the deposit may require to be credited is worse than failing the route (the picker falls back to another provider). Added two tests: the `|42` success path and the malformed `|abc` → `Decoding` rejection.

(Kept the `?dt=` form's bare-address fallback as-is — there the tag can legitimately be absent among other query params.)

`:data:testDebugUnitTest` green.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -456,9 +476,84 @@ constructor(
...
+        response: SwapKitSwapResponseJson,
+        rawTargetAddress: String,
+    ): Long? =
+        tagLongOrNull(response.destinationTag)
+            ?: tagLongOrNull(response.meta.destinationTag)
+            ?: extractTagSuffix(rawTargetAddress).second
+
+    /** Parse a destination tag that may arrive as a JSON number or a numeric string. */
+    private fun tagLongOrNull(element: JsonElement?): Long? =
+        (element as? JsonPrimitive)?.content?.trim()?.toLongOrNull()
```
rkokhatskyi: `tagLongOrNull` parses the XRP destination tag with signed `toLongOrNull` (likewise the `?dt=` and `|` suffix sites) so a negative wire value flows into `memo` and through `RippleHelper.setDestinationTag` onto XRP's uint32-only `DestinationTag`, whereas iOS rejects it with `UInt64(...)` at `SwapKitSwapResponse.swift:117` and emits no tag
```kotlin
(element as? JsonPrimitive)?.content?.trim()?.toLongOrNull()?.takeIf { it in 0..0xFFFFFFFFL }
```

johnnyluo:
@aminsato resolve the conflict please

johnnyluo (approved):
LGTM

aminsato:
> @aminsato resolve the conflict please

@johnnyluo 
Resolved

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Good catch — fixed in d6adf360d. Added a `destinationTagOrNull()` helper that clamps to `0..0xFFFFFFFFL` and applied it at all three parse sites (top-level/meta JSON, `?dt=` and `|` suffix), so an out-of-uint32-range value no longer reaches `RippleHelper.setDestinationTag`. The soft sites emit no tag (matching iOS' `UInt64` guard); the `|` suffix path treats an out-of-range value as malformed and fails the route, same as a non-numeric suffix. Covered with tests for a negative top-level tag (dropped), the max `2^32-1` tag (accepted), and an out-of-range `|` suffix (throws).

Issue #4675, opened 2026-05-28
Add SwapKit chain coverage to match iOS / Windows (Sui, Cardano, TON, Noble) + native signer paths


realpaaao:
## Summary

Android shows the swap button on a smaller chain set than iOS / Windows, and has no SwapKit code path at all. After SwapKit shipped via `@vultisig/sdk@0.26.x`, iOS and Windows surface swap on Sui / Cardano / TON; Android doesn't — and even if it did, quoting + signing would error because there's no Kotlin SwapKit integration.

## Current state

**`Chain.isSwapSupported`** (`data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt:173-202`) is a hardcoded array:

```kotlin
val Chain.isSwapSupported: Boolean
    get() = this in arrayOf(
        ThorChain, MayaChain, GaiaChain, Kujira, Bitcoin, Dogecoin, BitcoinCash,
        Litecoin, Dash, Ripple, Avalanche, Base, BscChain, Ethereum, Optimism,
        Polygon, Mantle, Arbitrum, Blast, CronosChain, Solana, ZkSync, Zcash,
        Tron, Hyperliquid,
    )
```

**Missing vs the SDK's `swapEnabledChains` union** (Native ∪ 1inch ∪ LI.FI ∪ SwapKit):
- `Sui`
- `Cardano`
- `Ton`
- `Noble` (THORChain supports it; safe to add even pre-SwapKit-Android since it routes through native THORChain code that already exists on Android)

**No SwapKit code path:** `grep -r swapkit app/ data/` returns zero hits. iOS has `Blockchain/Swaps/SwapKit/*` (config, service, provider cache, per-chain signers). Windows pulls SwapKit quoting via `@vultisig/sdk`.

## What iOS does

- `Chain.isSwapAvailable` switch table (pinned by `ChainSwapAvailabilityTests`) includes `cardano`, `sui`, `ton`.
- `Coin.swapProviders` returns `[.swapkit]` for those three.
- `SwapKitConfig.isFeatureEnabled` defaults to `true` (Settings → Advanced toggle exists but ships ON).
- Per-chain signers: `SwapKitBTCSigner`, `SwapKitDogeSigner`, `SwapKitDashSigner`, `SwapKitZcashSigner`, plus PSBT / programmable-tx / TON-internal paths for Sui / Cardano / TON.

## What Windows does

[...]

PR #4707, opened 2026-06-01
SwapKit: Cardano (ADA) signer for pre-built CBOR

### LINKED ISSUE #4700: SwapKit: Cardano (ADA) signer — sign pre-built CBOR
Split from #4675 — SwapKit per-chain signer for **Cardano (ADA)**.

SwapKit returns Cardano as a **pre-built unsigned CBOR** transaction: `meta.txType` is `"CARDANO"` / `"CBOR"`, `tx` is a hex-encoded unsigned CBOR. Routes via **NEAR Intents** (address-based — no memo routing).

### Work
- Add `SwapKitCardanoSigner`: sign the pre-built CBOR directly — Blake2b-256 over the tx body, Ed25519 sign, insert the vkey witness, re-serialize. Mirror iOS `SwapKitCardanoSigner` + `CardanoSignedTxBuilder`.
- Add `TX_TYPE_CARDANO` (+ `CBOR` alias) to `SwapKitSwapPayloadJson`; decode in `SwapKitQuoteSource` (binary CBOR into `txPayload`, like the BTC PSBT path); dispatch in `SigningHelper`.
- `SwapProviderTable`: `Chain.Cardano -> setOf(SWAPKIT)`; add `Chain.Cardano` to `isSwapSupported`; chain prefix `"ADA"`.

### WalletCore memo — checked, NOT a blocker
WalletCore 4.3.22's Cardano proto has **no** memo/metadata field (`SigningInput`: `transferMessage/ttl/plan/extraOutputs/utxos/staking/withdraw`; `Transfer`: `amount/toAddress/changeAddress/tokenAmount/useMaxAmount/forceFee`). But SwapKit Cardano doesn't need it: the tx is **pre-built CBOR**, so we sign it directly (custom signer, like iOS) instead of building it through `Cardano.SigningInput`. The WalletCore gap would only matter for a *native* Cardano send with a CIP-20 message — out of scope here. So this can proceed without waiting on a WalletCore bump.

aminsato:
Closes #4700 (split from #4675).

## Summary

SwapKit returns Cardano swaps as a **pre-built, hex-encoded unsigned CBOR** transaction (routed via NEAR Intents). WalletCore 4.3.22's `Cardano.SigningInput` can't ingest a pre-built envelope (and has no memo field), so we sign the bytes directly — mirroring iOS' `SwapKitCardanoSigner` + `CardanoSignedTxBuilder`.

Cardano signing model (Shelley-era):

```
tx_envelope = [ transaction_body, transaction_witness_set, is_valid, auxiliary_data ]
digest      = blake2b_256(cbor(transaction_body))   // == tx_id, the message Ed25519 signs
witness_set = { 0: [[vkey_32, signature_64]] }
```

### Transaction hash
 https://cardanoscan.io/transaction/70782335101d34163ee0de278c38660bb011e5fe8d1dc9baf2ca24300834b7b3

The signer slices item 0 (the body) verbatim, hashes it, and on assembly re-emits items 0/2/3 byte-for-byte while replacing item 1 with the vkey witness — re-encoding the body could change CBOR integer widths or map ordering and invalidate the signature.

- **Digest path** uses Bouncy Castle `Blake2b-256` (JNI-free, so it's unit-testable headlessly — same equivalence the SUI signer relies on). Ed25519 verification uses WalletCore.
- **vkey** is the vault EdDSA key (`getEddsaSigningKey(Chain.Cardano)`) — the same key the Cardano enterprise address is derived from, so its hash matches the payment credential SwapKit built against.
- The signed `rawTransaction` is hex CBOR — the same wire format the existing native Cardano submit path already broadcasts, so no broadcast changes were needed.

## Changes

- **New** `SwapKitCardanoSigner` (+ tests) — definite-length CBOR walker, Blake2b-256 digest, witness framing.
- `SwapKitSwapPayloadJson` — `TX_TYPE_CARDANO` + `CBOR` alias.
- `SwapKitQuoteSource` — `TxKind.CARDANO`, **hex** decode (Cardano's `tx` is hex, not base64 like PSBT/SUI), `ADA` chain prefix, native decimals 6, dispatch to the Native quote.
- `SigningHelper` — dispatch Cardano in both pre-sign and signed-tx paths.
- `SwapProviderTable` — `Chain.Cardano -> {SWAPKIT}`; add `Chain.Cardano` to `isSwapSupported` so the Swap button shows.
- `SwapFormViewModel` — allow the `CARDANO`/`CBOR` txType through the swap gate.

## Test plan

- `:data:testDebugUnitTest` — new `SwapKitCardanoSignerTest` (digest, nested-structure CBOR walker, witness framing, reject empty/non-array/trailing-bytes) + new `SwapKitQuoteSourceTest` cases (hex CBOR → Native quote, bad hex → `Decoding`); updated `SwapProviderTableTest`. ✅ green
- `:app:compileDebugKotlin` ✅ green
- `:data:ktfmtCheck` ✅ green

Signature verification (Ed25519) needs the WalletCore JNI and is exercised by the iOS port + integration; this PR pins the JNI-free surface headlessly. **Pending on-device E2E** (real ADA swap sign + broadcast) once SwapKit serves live ADA routes.

## Summary by CodeRabbit

* **New Features**
  * Cardano (ADA) added as a supported swap chain with both native deposit and pre-built CBOR swap flows.
  * Cardano-aware transaction signing and pre-signed image/hash handling for broadcast-ready Cardano transactions.

* **Documentation**
  * Clarified Cardano signing, tx semantics, and CBOR payload expectations.

* **Tests**
  * New and expanded tests for Cardano decoding, signing, routing, and error cases.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
@@ -120,6 +120,10 @@ object SigningHelper {
                             SwapKitSwapPayloadJson.TX_TYPE_SUI ->
                                 SwapKitSuiSigner(eddsaKey)
                                     .getPreSignedImageHash(swapPayload.data.txPayload)
+                            SwapKitSwapPayloadJson.TX_TYPE_CARDANO,
+                            SwapKitSwapPayloadJson.TX_TYPE_CBOR ->
+                                SwapKitCardanoSigner(eddsaKey)
+                                    .getPreSignedImageHash(swapPayload.data.txPayload)
                             else ->
                                 error(
                                     "Unsupported SwapKit txType for signing: ${swapPayload.data.txType}"
@@ -316,6 +320,10 @@ object SigningHelper {
                                 .getSignedTransaction(swapPayload.data.txPayload, signatures)
                         SwapKitSwapPayloadJson.TX_TYPE_SUI ->
                             SwapKitSuiSigner(eddsaKey)
+                                .getSignedTransaction(swapPayload.data.txPayload, signatures)
+                        SwapKitSwapPayloadJson.TX_TYPE_CARDANO,
+                        SwapKitSwapPayloadJson.TX_TYPE_CBOR ->
+                            SwapKitCardanoSigner(eddsaKey)
                                 .getSignedTransaction(swapPayload.data.txPayload, signatures)
                         else ->
                             error(
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
@@ -120,10 +120,13 @@ object SigningHelper {
                             SwapKitSwapPayloadJson.TX_TYPE_SUI ->
                                 SwapKitSuiSigner(eddsaKey)
                                     .getPreSignedImageHash(swapPayload.data.txPayload)
-                            SwapKitSwapPayloadJson.TX_TYPE_CARDANO,
-                            SwapKitSwapPayloadJson.TX_TYPE_CBOR ->
+                            SwapKitSwapPayloadJson.TX_TYPE_CARDANO_PREBUILT ->
                                 SwapKitCardanoSigner(eddsaKey)
                                     .getPreSignedImageHash(swapPayload.data.txPayload)
+                            // Deposit-only Cardano: no CBOR to sign — hash a plain ADA send to
+                            // targetAddress built from the blockChainSpecific, via the native path.
+                            SwapKitSwapPayloadJson.TX_TYPE_CARDANO ->
+                                CardanoHelper.getPreSignedImageHash(payload)
                             else ->
                                 error(
                                     "Unsupported SwapKit txType for signing: ${swapPayload.data.txType}"
@@ -321,10 +324,18 @@ object SigningHelper {
                         SwapKitSwapPayloadJson.TX_TYPE_SUI ->
                             SwapKitSuiSigner(eddsaKey)
                                 .getSignedTransaction(swapPayload.data.txPayload, signatures)
-                        SwapKitSwapPayloadJson.TX_TYPE_CARDANO,
-                        SwapKitSwapPayloadJson.TX_TYPE_CBOR ->
+                        SwapKitSwapPayloadJson.TX_TYPE_CARDANO_PREBUILT ->
                             SwapKitCardanoSigner(eddsaKey)
                                 .getSignedTransaction(swapPayload.data.txPayload, signatures)
+                        // Deposit-only Cardano: build & sign a plain ADA send to targetAddress via
+                        // the native Cardano path (same as a non-swap send).
+                        SwapKitSwapPayloadJson.TX_TYPE_CARDANO ->
+                            CardanoHelper.getSignedTransaction(
+                                vaultHexPublicKey = eddsaKey,
+                                vaultHexChainCode = vault.hexChainCode,
+                                keysignPayload = keysignPayload,
+                                signatures = signatures,
+                            )
                         else ->
                             error(
                                 "Unsupported SwapKit txType for signing: ${swapPayload.data.txType}"
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
@@ -27,6 +27,7 @@ import kotlinx.coroutines.flow.first
 import kotlinx.datetime.Clock
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonNull
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
 import kotlinx.serialization.json.decodeFromJsonElement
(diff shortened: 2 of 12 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
```diff
@@ -120,6 +120,10 @@ object SigningHelper {
                             SwapKitSwapPayloadJson.TX_TYPE_SUI ->
                                 SwapKitSuiSigner(eddsaKey)
                                     .getPreSignedImageHash(swapPayload.data.txPayload)
+                            SwapKitSwapPayloadJson.TX_TYPE_CARDANO,
```
rkokhatskyi: Dispatching the CBOR signer on the raw `tx_type` `CARDANO`/`CBOR` diverges from the shared keysign contract, since `commondata/.../swapkit_swap_payload.proto` reserves `CARDANO` for the empty payload deposit flow while iOS emits `CARDANO_PREBUILT` for this prebuilt path and routes its signer only on that, so a vault mixing iOS and Android cannot jointly sign a Cardano swap
```kotlin
SwapKitSwapPayloadJson.TX_TYPE_CARDANO_PREBUILT -> // also written by the quote source, replacing raw CARDANO/CBOR
    SwapKitCardanoSigner(eddsaKey).getPreSignedImageHash(swapPayload.data.txPayload)
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -456,9 +460,38 @@ constructor(
                     ?: throw SwapKitError.Decoding("SwapKit TRON tx is missing raw_data_hex")
                 json.encodeToString(JsonObject.serializer(), obj).encodeToByteArray()
             }
+            // SwapKit returns the Cardano tx as a hex-encoded CBOR string (not base64), mirroring
+            // iOS' `Data(hexString:)` decode path. Decode hex into the raw CBOR envelope bytes the
+            // signer hashes and re-frames.
+            TxKind.CARDANO -> decodeHexTx(response.tx)
```
rkokhatskyi: Routing every `CARDANO`/`CBOR` response through `decodeHexTx` rejects the deposit flow whose `tx` is empty or absent, which 2 of 3 captured SwapKit ADA fixtures use and which `swapkit_swap_payload.proto` defines as the canonical `CARDANO` case, so those routes yield no quote while iOS builds a native ADA send to `targetAddress`
```kotlin
// empty or absent tx routes a native ADA send to targetAddress (mirror iOS .cardano); tx must be nullable
```

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
aminsato: Good catch — fixed in 7826740fd. The dispatcher now routes the pre-built path on `CARDANO_PREBUILT` (matching what iOS emits and dispatches on), and the quote source writes that discriminator into the keysign payload instead of the raw wire `CARDANO`/`CBOR`. `CBOR` is now treated as a wire-only alias the quote source normalizes away, so it never reaches keysign. Deposit-only `CARDANO` is handled separately (see the other thread). This restores joint signing for a vault mixing iOS and Android.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Agreed, and the two real captured fixtures confirm it — both are `tx: null`. Fixed in 7826740fd:
- `SwapKitSwapResponseJson.tx` is now nullable.
- The Cardano branch splits on `tx` presence (not the txType string): null/absent → deposit-only → empty payload + `CARDANO`; hex → pre-built → CBOR bytes + `CARDANO_PREBUILT`. Mirrors iOS' `.cardano` vs `.cardanoPrebuilt` decode.
- `SigningHelper` routes deposit-only `CARDANO` through the native `CardanoHelper` send path (the keysign payload already carries `toAddress=targetAddress`, `toAmount=fromAmount`, and the Cardano `blockChainSpecific` with UTXOs), so it builds a plain ADA send rather than yielding no quote.

Added a deposit-only quote-source test (`tx: null` → empty payload + `CARDANO`) alongside the existing hex-CBOR case (now asserting `CARDANO_PREBUILT`).

data/src/test/kotlin/com/vultisig/wallet/data/api/models/quotes/SwapKitQuoteDecodingTest.kt
```diff
@@ -168,7 +168,7 @@ class SwapKitQuoteDecodingTest {
         assertEquals("999500000", response.expectedBuyAmount)
 
         // tx is JsonElement until the caller decodes onto the matching DTO
-        val evm = json.decodeFromJsonElement(SwapKitEvmTx.serializer(), response.tx)
+        val evm = json.decodeFromJsonElement(SwapKitEvmTx.serializer(), response.tx!!)
```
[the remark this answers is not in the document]
aminsato: Fixed in 8bf3bc4f0, though slightly differently than suggested. After merging `main`, `SwapKitSwapResponseJson.tx` is no longer nullable — it's `JsonElement = JsonNull` (the deposit-only routes default to `JsonNull` rather than Kotlin `null`). So `response.tx` is already non-null here and the `!!` was a redundant assertion. I dropped it entirely (`response.tx`) rather than adding `assertNotNull`, since there's no nullable value left to assert against.

rkokhatskyi (approved):
LGTM

Issue #4712, opened 2026-06-01
Swap quote UX: instant indicative rate, no-flicker refresh, instant on percentage/paste


realpaaao:
Three display-layer enhancements to make the swap quote feel instant. None change the signed transaction — the firm quote still gates Continue.

## 1. Instant indicative rate (never blank the dst amount)

`estimatedDstTokenValue` is `"0"` until the quote resolves, and reset to `"0"` each cycle:
- `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt:85` (state default `"0"`), `:1014` (set from quote), `:1174` (reset to `"0"`)

The fiat path (`ConvertTokenValueToFiatUseCase`) already has spot prices. Compute an indicative dst value instantly from `src × (priceSrc / priceDst)`, show it greyed, replace on quote. Display-only.

## 2. No-flicker refresh (stale-while-revalidate)

`isLoading = true` + dst reset to `"0"` each refresh blanks the estimate every cycle:
- `…/SwapFormViewModel.kt:871`, `:1174`

Keep the last `estimatedDstTokenValue` visible while refetching; skeleton only on the first quote for a pair.

## 3. Instant on percentage / Max / paste

`selectSrcPercentage` sets `srcAmountState` text, which flows through the `.debounce(450L)`:
- `…/SwapFormViewModel.kt:718` (`selectSrcPercentage`) → `:865` / `:869` (`textAsFlow` + debounce)

Fire immediately for percentage / Max / paste; debounce only free typing. (Pairs with #4711.)

PR #4714, opened 2026-06-02
fix(swap): show loading on input, move ahead of the debounce

### LINKED ISSUE #4711: Swap: show loading on input (move ahead of the 450ms debounce)
The swap quote shows no loading feedback for ~450ms after the user stops typing. Two changes, in order:

## 1. Decouple the loading state from the debounce (show the spinner on input)

`SwapFormViewModel.calculateFees()` sets `isLoading = true` **inside** the `.collect`, i.e. *after* the debounce window — so there's no loading indicator until the debounce elapses:
- `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt:869` (`.debounce(450L)`)
- `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt:871` (`isLoading = true`, inside `.collect`)

**Ask:** emit the loading state immediately on amount change (ahead of the debounce). Keep the debounce gating the quote network call — only the UI feedback moves earlier, so no extra requests fire. This is the main fix.

## 2. Reduce the debounce 450ms → ~300ms

450ms is the highest of the three apps (iOS = 200ms, Windows = 0). Once loading shows on input (step 1) the debounce stops feeling like a freeze, but tightening it to ~300ms brings the *quote start* closer to iOS without meaningfully increasing provider load (each fetch fans out to THOR/Maya/Kyber/1inch/LiFi/SwapKit, so some coalescing is still wanted).

**Ask:** drop the debounce to ~300ms in the same PR.

## Caching — already handled, noted for completeness

Android already caches the discount-tier pre-work — `GetDiscountBpsUseCase` reads `getCachedTokenBalances` + a local NFT flag (`TiersNFTRepositoryImpl`), parallelized per provider — so no on-chain VULT/Thorguard read sits on the quote critical path here (unlike iOS/Windows). No caching action needed.

The only remaining uncached network hop is a token-price `refresh()` for cold tokens in `ConvertTokenValueToFiatUseCase.kt:29`; out of scope here, flag separately if worth it.

johnnyluo:
Closes #4711

## Problem
`SwapFormViewModel.calculateFees()` set `isLoading = true` **inside** `.collect`, i.e. *after* the 450ms debounce window. So for ~450ms after the user stopped typing, the swap form showed no loading feedback and felt frozen.

## Changes
1. **Move the loading state ahead of the debounce** — `isLoading = true` now runs in an `.onEach` before `.debounce(...)`, so the spinner appears immediately on input. The debounce still gates the actual quote network call (`fetchBestQuote` fans out to THOR/Maya/Kyber/1inch/LiFi/SwapKit), so **no extra provider requests fire**.
2. **Tighten the debounce 450ms → 300ms** — brings the *quote start* closer to iOS (200ms) while keeping request coalescing.

No caching changes — Android already caches the discount-tier pre-work via `GetDiscountBpsUseCase`, so nothing on-chain sits on the quote critical path (per the issue).

## Diff
```kotlin
.combine(refreshQuoteState) { it, _ -> it }
.onEach { isLoading = true }   // immediate UI feedback
.debounce(300L)                // still coalesces the network call
.collect { (address, amount) ->
    // isLoading = true  (removed from here)
    ...
```

## Test plan
- [ ] Open Swap, enter a source amount → spinner appears immediately on keystroke (no ~450ms freeze)
- [ ] Quote still fetches once per pause; rapid typing does not fire multiple quote requests
- [ ] Error paths (zero/negative amount, same asset, unsupported pair) still clear loading correctly

## Screenshots
N/A — this is a transient loading-timing change (spinner appears earlier after a keystroke), not a visual/layout change against Figma. There is no static before/after frame to capture.

## Summary by CodeRabbit

* **Bug Fixes**
  * Swap form now shows loading immediately when you change the amount, shortens the debounce interval, and cancels stale quote requests so outdated results don't overwrite newer input.
  * Primary action label now reflects whether the amount field is empty (prompt to fill in) rather than relying on loading state.

* **Tests**
  * Updated tests to match the refined debounce timing and loading behavior.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -923,9 +923,12 @@ constructor(
                     address to srcAmount
                 }
                 .combine(refreshQuoteState) { it, _ -> it }
-                .debounce(450L)
+                // Show the spinner immediately on input, ahead of the debounce, so the form
+                // doesn't look frozen while we wait. The debounce below still coalesces the
+                // actual quote fetch, so this adds no extra provider requests.
+                .onEach { isLoading = true }
+                .debounce(300L)
                 .collect { (address, amount) ->
-                    isLoading = true
                     val (src, dst) = address
 
                     val srcToken = src.account.token
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -65,6 +65,7 @@ import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.flow.distinctUntilChanged
@@ -919,16 +920,25 @@ constructor(
                     src to dst
                 }
                 .distinctUntilChanged()
-                .combine(srcAmountState.textAsFlow().filter { it.isNotEmpty() }) { address, _ ->
+                .combine(
+                    srcAmountState
+                        .textAsFlow()
+                        .filter { it.isNotEmpty() }
+                        // Show the spinner immediately on real typing, ahead of the debounce, so
+                        // the form doesn't look frozen while we wait. Gating it on this branch
+                        // keeps it to user input: the silent refreshes (refreshQuoteState timer,
+                        // gas bump) don't flow through here, so they no longer flash the spinner
+                        // (#4712). The debounce below still coalesces the actual quote fetch, so
+                        // this adds no extra provider requests.
+                        .onEach { isLoading = true }
+                ) { address, _ ->
                     address to srcAmount
                 }
                 .combine(refreshQuoteState) { it, _ -> it }
-                // Show the spinner immediately on input, ahead of the debounce, so the form
-                // doesn't look frozen while we wait. The debounce below still coalesces the
-                // actual quote fetch, so this adds no extra provider requests.
-                .onEach { isLoading = true }
                 .debounce(300L)
-                .collect { (address, amount) ->
+                // collectLatest so newer input cancels an in-flight fetch instead of letting a
+                // stale fetch write isLoading = false after the user has already typed again.
+                .collectLatest { (address, amount) ->
                     val (src, dst) = address
 
                     val srcToken = src.account.token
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -923,9 +923,12 @@ constructor(
                     address to srcAmount
                 }
                 .combine(refreshQuoteState) { it, _ -> it }
-                .debounce(450L)
+                // Show the spinner immediately on input, ahead of the debounce, so the form
+                // doesn't look frozen while we wait. The debounce below still coalesces the
+                // actual quote fetch, so this adds no extra provider requests.
+                .onEach { isLoading = true }
+                .debounce(300L)
                 .collect { (address, amount) ->
```
rkokhatskyi: Since `debounce` runs `onEach` in a separate coroutine, a keystroke while a fetch is running sets `isLoading = true`, then the older fetch finishes and writes `isLoading = false` at line 1107 so the newer fetch shows no spinner. Set it again at the top of `collect`, or use `collectLatest` so new input cancels the stale fetch the way iOS does
```kotlin
.collect { (address, amount) ->
    isLoading = true
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -923,9 +923,12 @@ constructor(
                     address to srcAmount
                 }
                 .combine(refreshQuoteState) { it, _ -> it }
```
rkokhatskyi, line 937: `onEach` sits below `combine(refreshQuoteState)`, so the silent refresh from `launchRefreshQuoteTimer` and the gas bump at line 897 now flash the spinner too, which is the flicker #4712 wants gone. Putting `isLoading = true` on the `srcAmountState.textAsFlow()` branch keeps it to real typing

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -923,9 +923,12 @@ constructor(
...
-                .debounce(450L)
+                // Show the spinner immediately on input, ahead of the debounce, so the form
+                // doesn't look frozen while we wait. The debounce below still coalesces the
+                // actual quote fetch, so this adds no extra provider requests.
+                .onEach { isLoading = true }
```
rkokhatskyi: This also flips the button to `Fill in amount` (`SwapScreen.kt:644`) for the whole debounce while editing an amount that already has a quote, where it used to stay on `Swap` until the fetch began. Better to gate that label on the empty state rather than on `isLoading`

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -923,9 +923,12 @@ constructor(
...
+                .debounce(300L)
```
rkokhatskyi, line 938: Nothing tests this, and the existing `calculateFees debounces amount changes` test asserts nothing while its comments at `SwapFormViewModelTest.kt:666` and `:710` still say 450ms. A Turbine check that `isLoading` is true after a text change but before `advanceTimeBy(300)` would cover it

rkokhatskyi (changes requested):

rkokhatskyi (approved):
LGTM

PR #4715, opened 2026-06-02
feat(swap): SwapKit UTXO sources — BCH / DOGE / LTC / DASH / ZEC (#4702)

### LINKED ISSUE #4702: SwapKit: UTXO sources — BCH / DOGE / LTC / DASH / ZEC
Split from #4675 — SwapKit UTXO sources beyond BTC: **BCH / DOGE / LTC / DASH / ZEC**.

BTC PSBT already landed (#4674 / `SwapKitBtcSigner`). These reuse the existing UTXO helpers with per-chain sighash. Mirror iOS `SwapKit{Doge,Dash,Zcash}Signer` + the BCH path. Wire each in `SigningHelper` + `SwapProviderTable` (+SWAPKIT) + `isSwapSupported`. Split per-chain later if any proves divergent.

- [ ] LTC — segwit / PSBT
- [ ] DOGE — legacy P2PKH
- [ ] DASH — legacy P2PKH
- [ ] BCH — SIGHASH_FORKID
- [ ] ZEC — ZIP-243 sighash

aminsato:
Closes #4702 (split from #4675). Extends SwapKit source-chain support beyond the BTC PSBT path (#4674) to the remaining UTXO chains, mirroring the iOS `SwapKit{Doge,Dash,Zcash}Signer` + the BCH/LTC paths.

## What

Three signing paths, picked from the source chain:

| Chain(s) | Path | Implementation |
|---|---|---|
| BTC, **LTC** | segwit / BIP-143 | `SwapKitBtcSigner` — now takes a `CoinType` (default `BITCOIN`); `UtxoHelper`'s PSBT co-signing guard widened to allow `LITECOIN` |
| **DOGE, BCH, DASH** | legacy P2PKH | new `SwapKitLegacyP2PKHSigner` — parses the legacy unsigned-tx body, builds a **frozen** `Bitcoin.TransactionPlan` (no replanning, so the broadcast tx_id the route is tracked by stays stable), signs via `TransactionCompiler` per `CoinType` (BCH gets `SIGHASH_FORKID` from `hashTypeForCoin`) |
| **ZEC** | Sapling-v4 / ZIP-243 | new `SwapKitZcashSigner` — walks the Sapling-v4 body, hard-rejects NU5/v5 and any shielded value, sets the native branch id on the frozen plan |

## Wiring

- `SwapKitSwapPayloadJson` — per-chain PSBT discriminators (`PSBT_DOGE` / `PSBT_BCH` / `PSBT_DASH` / `PSBT_ZEC`). The discriminator rides the shared keysign payload, so distinguishing segwit from legacy from Sapling here lets a mixed iOS/Android vault cosign correctly (iOS emits the same strings).
- `SigningHelper` — dispatch in both the presign and signed-tx paths, using the payload's source chain to pick the `CoinType`.
- `SwapProviderTable` — `DOGE`/`BCH`/`LTC` → `THORCHAIN` + `SWAPKIT`; `DASH`/`ZEC` → `MAYA` + `SWAPKIT`.
- `SwapKitQuoteSource` — chain prefixes, native decimals (8) for inbound-fee scaling, and chain-aware PSBT `txType` so a route is built and stamped with the right discriminator.
- `Chain.isSwapSupported` already listed all five.

## Tests

Per-signer suites for the legacy and ZEC signers, plus an LTC case in the BTC test, modelled on the existing `SwapKitBtcSignerTest` (PSBT-builder + `skipIfJniUnavailable`):

- **Headless (run + pass):** PSBT/Sapling framing and structural validation — P2PKH shape, output count, fee sign, prev-UTXO presence, and the ZEC version-group + shielded-field rejection.
- **WalletCore happy paths:** `getPreSignedImageHash` for DOGE (`WITNESS_UTXO`), DASH (`NON_WITNESS_UTXO`, two inputs) and transparent ZEC — these need the native lib so they skip locally and run in CI.

`SwapProviderTableTest` updated for the newly enabled chains.

## Test plan

- [x] `:data` main + unit-test compile
- [x] `:data:testDebugUnitTest --tests '*SwapKit*' --tests '*SwapProviderTable*'` green
- [x] `:data:ktfmtCheck` clean
- [x] On-device E2E: signed + broadcast a real **DOGE → TRX** SwapKit (NEAR) swap on the legacy P2PKH path — tx [`2d1249e27e90eee59e1868bf8e1f71883a219d9c94e16bef950686acd362b9e0`](https://blockchair.com/dogecoin/transaction/2d1249e27e90eee59e1868bf8e1f71883a219d9c94e16bef950686acd362b9e0)
- [x] On-device E2E: signed + broadcast a real **LTC** SwapKit swap on the segwit PSBT path — tx [`4334c1bf39e168a0f1d6e250f29da27edf5660d2827f48d900cb596b63ef9d4a`](https://blockchair.com/litecoin/transaction/4334c1bf39e168a0f1d6e250f29da27edf5660d2827f48d900cb596b63ef9d4a)
- [x] On-device E2E: signed + broadcast a real **BCH** SwapKit swap on the legacy P2PKH + `SIGHASH_FORKID` path — tx [`c44fe710efaa2130db3337be0bc05e114af1f6ef25365eb69eec88cabe33d886`](https://blockchair.com/bitcoin-cash/transaction/c44fe710efaa2130db3337be0bc05e114af1f6ef25365eb69eec88cabe33d886)
- [x] On-device E2E: signed + broadcast a real **DASH** SwapKit swap on the legacy P2PKH path — tx [`f837a5c1e788b9b3daf98038c2b60d54a862def8aedbb133daf884f656bede7d`](https://blockchair.com/dash/transaction/f837a5c1e788b9b3daf98038c2b60d54a862def8aedbb133daf884f656bede7d)
- [ ] On-device E2E for the remaining chain (ZEC)

### On-device proof (DASH → LTC SwapKit swap)

DASH to LTC swap — Transaction successful (tx f837a5…ede7d)

## Summary by CodeRabbit

* **New Features**
  * SwapKit signing expanded to support Litecoin, Dogecoin, Bitcoin Cash, Dash, and Zcash PSBT routes.
  * Swap provider wiring updated so additional UTXO chains expose SwapKit where applicable.

* **Documentation**
  * Clarified PSBT signing paths, segwit vs legacy distinctions, and native-fee decimal handling for UTXO chains.

* **Tests**
  * Added JVM/JUnit tests covering PSBT signing happy paths and structural rejection cases for the new chains.

* **Bug Fixes / UX**
  * Swap quote selection and swap submission now recognize the new chain-specific PSBT tx types.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -475,30 +475,13 @@ constructor(
                             }
 
                             is SwapQuote.SwapKit -> {
-                                // BTC/LTC segwit PSBT, DOGE/BCH/DASH legacy-P2PKH PSBT, ZEC
-                                // Sapling-v4 PSBT, TRON (TronWeb object), SUI (PTB), TON (native
-                                // transfer), XRP (deposit-only native Payment), and both Cardano
-                                // flows (deposit-only CARDANO + pre-built CARDANO_PREBUILT) are
-                                // wired with their per-chain signers/native paths. Guarded loudly
-                                // so an un-wired txType can't reach signing.
+                                // Pre-flight gate: refuse a route whose txType has no wired signing
+                                // path before staging keysign. Sourced from
+                                // SwapKitSwapPayloadJson.SIGNABLE_TX_TYPES — the same list
+                                // SigningHelper dispatches on — so this guard can't drift from what
+                                // the dispatcher actually accepts.
                                 require(
-                                    quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_PSBT ||
-                                        quote.data.txType ==
-                                            SwapKitSwapPayloadJson.TX_TYPE_PSBT_DOGE ||
-                                        quote.data.txType ==
-                                            SwapKitSwapPayloadJson.TX_TYPE_PSBT_BCH ||
-                                        quote.data.txType ==
-                                            SwapKitSwapPayloadJson.TX_TYPE_PSBT_DASH ||
-                                        quote.data.txType ==
-                                            SwapKitSwapPayloadJson.TX_TYPE_PSBT_ZEC ||
-                                        quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_TRON ||
-                                        quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_SUI ||
-                                        quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_TON ||
-                                        quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_XRP ||
-                                        quote.data.txType ==
-                                            SwapKitSwapPayloadJson.TX_TYPE_CARDANO ||
-                                        quote.data.txType ==
-                                            SwapKitSwapPayloadJson.TX_TYPE_CARDANO_PREBUILT
+                                    SwapKitSwapPayloadJson.isSignableTxType(quote.data.txType)
                                 ) {
                                     "Unsupported SwapKit txType for swap: ${quote.data.txType}"
                                 }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -475,13 +475,22 @@ constructor(
                             }
 
                             is SwapQuote.SwapKit -> {
-                                // BTC PSBT, TRON (TronWeb object), SUI (PTB), TON (native
+                                // BTC/LTC segwit PSBT, DOGE/BCH/DASH legacy-P2PKH PSBT, ZEC
+                                // Sapling-v4 PSBT, TRON (TronWeb object), SUI (PTB), TON (native
                                 // transfer), XRP (deposit-only native Payment), and both Cardano
                                 // flows (deposit-only CARDANO + pre-built CARDANO_PREBUILT) are
                                 // wired with their per-chain signers/native paths. Guarded loudly
                                 // so an un-wired txType can't reach signing.
                                 require(
                                     quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_PSBT ||
+                                        quote.data.txType ==
+                                            SwapKitSwapPayloadJson.TX_TYPE_PSBT_DOGE ||
+                                        quote.data.txType ==
+                                            SwapKitSwapPayloadJson.TX_TYPE_PSBT_BCH ||
+                                        quote.data.txType ==
+                                            SwapKitSwapPayloadJson.TX_TYPE_PSBT_DASH ||
+                                        quote.data.txType ==
+                                            SwapKitSwapPayloadJson.TX_TYPE_PSBT_ZEC ||
                                         quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_TRON ||
                                         quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_SUI ||
                                         quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_TON ||
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
@@ -330,8 +330,15 @@ internal class SwapKitLegacyP2PKHSigner(
             (0L until inCount).map {
                 val prevBytes = cursor.readBytes(32) // internal little-endian wire order
                 val prevIndex = cursor.readUInt32LE()
+                // A PSBT's unsigned tx must carry empty scriptSigs. Reject a non-empty one rather
+                // than silently skip it — the frozen plan rebuilds the tx from these parsed fields,
+                // so dropped bytes would change the broadcast tx_id the route is tracked by.
                 val scriptSigLen = cursor.readCompactSize()
-                cursor.readBytes(cursor.asLength(scriptSigLen)) // empty in unsigned tx
+                if (scriptSigLen != 0L) {
+                    throw SwapKitLegacyP2PKHSignerException(
+                        "SwapKit PSBT unsigned-tx input #$it scriptSig must be empty"
+                    )
+                }
                 val sequence = cursor.readUInt32LE()
                 ParsedLegacyTxInput(prevBytes, prevIndex, sequence)
             }
@@ -343,6 +350,13 @@ internal class SwapKitLegacyP2PKHSigner(
                 LegacyP2PKHOutput(amount, cursor.readBytes(cursor.asLength(scriptLen)))
             }
         cursor.readUInt32LE() // locktime
+        // The unsigned-tx body must consume exactly; trailing bytes (e.g. a stray BIP-144 witness
+        // flag) would be silently dropped from the rebuilt tx, diverging it from the PSBT body.
+        if (!cursor.isAtEnd) {
+            throw SwapKitLegacyP2PKHSignerException(
+                "SwapKit PSBT unsigned-tx body has trailing bytes"
+            )
+        }
         return ParsedLegacyTx(inputs, outputs)
     }
 
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
@@ -5,6 +5,8 @@ import com.vultisig.wallet.data.crypto.checkError
 import com.vultisig.wallet.data.models.SignedTransactionResult
 import com.vultisig.wallet.data.utils.Numeric
 import com.vultisig.wallet.data.utils.getDustThreshold
+import java.math.BigInteger
+import java.security.MessageDigest
 import tss.KeysignResponse
 import wallet.core.jni.Base58
 import wallet.core.jni.BitcoinScript
@@ -48,8 +50,12 @@ internal class SwapKitLegacyP2PKHSigner(
      * preimage construction via the frozen-plan input data and [coinType] (BCH gets SIGHASH_FORKID
      * through `hashTypeForCoin`).
      */
-    fun getPreSignedImageHash(psbtBytes: ByteArray, targetAddress: String): List<String> {
-        val inputData = buildSigningInputData(psbtBytes, targetAddress)
+    fun getPreSignedImageHash(
+        psbtBytes: ByteArray,
+        targetAddress: String,
+        fromAmount: BigInteger,
+    ): List<String> {
+        val inputData = buildSigningInputData(psbtBytes, targetAddress, fromAmount)
         val preHashes = TransactionCompiler.preImageHashes(coinType, inputData)
         val preSigningOutput = Bitcoin.PreSigningOutput.parseFrom(preHashes).checkError()
         return preSigningOutput.hashPublicKeysList
@@ -66,18 +72,24 @@ internal class SwapKitLegacyP2PKHSigner(
     fun getSignedTransaction(
         psbtBytes: ByteArray,
         targetAddress: String,
+        fromAmount: BigInteger,
         signatures: Map<String, KeysignResponse>,
     ): SignedTransactionResult {
-        val inputData = buildSigningInputData(psbtBytes, targetAddress)
+        val inputData = buildSigningInputData(psbtBytes, targetAddress, fromAmount)
         return utxo.getSignedTransaction(inputData, signatures)
     }
 
     /**
      * Build the serialized [Bitcoin.SigningInput] with a **frozen** [Bitcoin.TransactionPlan]
-     * derived directly from the PSBT bytes. Exposed `internal` so per-chain unit tests can pin the
-     * structural shape (input count, scriptPubKey patterns, plan amount/change/fee).
+     * derived directly from the PSBT bytes. [fromAmount] is the quoted swap amount (source-chain
+     * base units) used to bound the miner fee. Exposed `internal` so per-chain unit tests can pin
+     * the structural shape (input count, scriptPubKey patterns, plan amount/change/fee).
      */
-    internal fun buildSigningInputData(psbtBytes: ByteArray, targetAddress: String): ByteArray {
+    internal fun buildSigningInputData(
+        psbtBytes: ByteArray,
+        targetAddress: String,
+        fromAmount: BigInteger,
+    ): ByteArray {
         if (psbtBytes.isEmpty()) {
             throw SwapKitLegacyP2PKHSignerException("SwapKit PSBT payload is empty")
         }
@@ -98,7 +110,12 @@ internal class SwapKitLegacyP2PKHSigner(
         val inputs =
             parsedTx.inputs.mapIndexed { index, parsedInput ->
                 val (amount, scriptPubKey) =
-                    resolvePrevUtxo(inputMaps[index], parsedInput.prevIndex, index)
+                    resolvePrevUtxo(
+                        inputMaps[index],
+                        parsedInput.prevTxIdLE,
+                        parsedInput.prevIndex,
+                        index,
+                    )
                 val keyHash = assertP2PKHAndExtractKeyHash(scriptPubKey, index, "input")
                 LegacyP2PKHInput(
                     prevTxIdLE = parsedInput.prevTxIdLE,
@@ -110,7 +127,13 @@ internal class SwapKitLegacyP2PKHSigner(
                 )
             }
 
-        return assembleSigningInput(inputs, parsedTx.outputs, parsedTx.lockTime, targetAddress)
+        return assembleSigningInput(
+            inputs,
+            parsedTx.outputs,
+            parsedTx.lockTime,
+            fromAmount,
+            targetAddress,
+        )
     }
 
     /**
@@ -126,6 +149,7 @@ internal class SwapKitLegacyP2PKHSigner(
         inputs: List<LegacyP2PKHInput>,
         outputs: List<LegacyP2PKHOutput>,
         lockTime: Long,
+        fromAmount: BigInteger,
         targetAddressHint: String,
     ): ByteArray {
         if (inputs.isEmpty() || outputs.isEmpty()) {
@@ -147,6 +171,18 @@ internal class SwapKitLegacyP2PKHSigner(
         if (fee < 0) {
             throw SwapKitLegacyP2PKHSignerException(
                 "SwapKit PSBT negative fee: inputs=$totalIn outputs=$totalOut"
+            )
+        }
+        // Fee ceiling. The Verify screen never surfaces the miner fee, and the legacy path doesn't
+        // bind the deposit address, so a PSBT that under-allocates the change output would burn the
+        // remainder to the miner. A fixed sat/vByte ceiling like the BTC path isn't portable here
+        // (DOGE's legitimate fees run orders of magnitude higher), so bound the fee by the quoted
+        // swap amount: a miner fee larger than the amount being swapped is pathological and
+        // refused.
+        if (BigInteger.valueOf(fee) > fromAmount) {
+            throw SwapKitLegacyP2PKHSignerException(
+                "SwapKit PSBT miner fee $fee exceeds the quoted swap amount $fromAmount; refusing " +
+                    "to sign"
             )
         }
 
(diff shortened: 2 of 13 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
```diff
@@ -0,0 +1,403 @@
...
+        val outCount = cursor.readCompactSize()
+        val outputs =
+            (0L until outCount).map {
+                val amount = cursor.readUInt64LE()
+                val scriptLen = cursor.readCompactSize()
+                LegacyP2PKHOutput(amount, cursor.readBytes(cursor.asLength(scriptLen)))
+            }
+        cursor.readUInt32LE() // locktime
+        return ParsedLegacyTx(inputs, outputs)
+    }
```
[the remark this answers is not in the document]
aminsato: Fixed in c01e1118d — `parseLegacyUnsignedTx` now requires empty input scriptSigs and asserts `cursor.isAtEnd()` after locktime, matching `SwapKitBtcSigner.parseUnsignedTx`. Both guards are pinned by tests.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
```diff
@@ -0,0 +1,371 @@
...
+        val nJoinSplit = cursor.readCompactSize()
+        if (
+            valueBalance != 0L || nShieldedSpend != 0L || nShieldedOutput != 0L || nJoinSplit != 0L
+        ) {
+            throw SwapKitZcashSignerException(
+                "SwapKit ZEC PSBT carries a shielded bundle; only transparent-only txs are supported"
+            )
+        }
+        return ParsedSaplingTx(inputs, outputs)
+    }
```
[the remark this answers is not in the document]
aminsato: Fixed in c01e1118d — `parseSaplingUnsignedTx` now rejects non-empty input scriptSigs and asserts `cursor.isAtEnd()` after the (all-zero) shielded counts, so the transparent-only body is consumed exactly. Pinned by tests.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
```diff
@@ -0,0 +1,403 @@
...
+        val changeAddress =
+            if (outputs.size >= 2) {
+                legacyAddress(outputKeyHashes[1]) ?: targetAddressHint
+            } else {
+                // No change output — point `changeAddress` at the source pubkey hash so
+                // WalletCore's
+                // validator accepts the input. The plan's `change = 0` means no change output is
+                // actually emitted, so the address is a structural placeholder.
+                legacyAddress(inputs[0].keyHash) ?: targetAddressHint
+            }
```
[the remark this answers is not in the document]
aminsato: Keeping as-is — `targetAddress` binding is deliberately not applied on these UTXO-sibling paths, and requiring an output to equal it would reject valid swaps.

The deposit/change addresses are derived from the PSBT's own output hash160s, not from `targetAddress`. SwapKit returns `targetAddress` as the nominal deposit address, but the NEAR-Intents route allocates the actual on-chain deposit address, and the two differ on these chains — the ZEC path documents a captured fixture where output 0 lands on a route-allocated transparent address `!= targetAddress`. So `output == targetAddress` would be a false rejection, which is exactly why the address is used only as a derivation fallback (mirroring the iOS signer).

The integrity guarantee here comes from the frozen plan re-emitting the PSBT outputs verbatim (the broadcast tx is byte-identical to what was quoted), with the SwapKit `/v3/swap` response as the trust boundary — the same boundary as the amount and fee. The stronger destination binding on the BTC segwit path is only possible because BTC's deposit output equals `targetAddress`; it isn't portable here without per-chain fixtures confirming that relationship. An amount-level bound (deposit `<= fromAmount`, independent of the address) is a reasonable future hardening, but it's out of scope for this iOS-parity PR.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
```diff
@@ -0,0 +1,371 @@
...
+        // Derive t1 addresses from the parsed output hash160s (NOT from `targetAddress` — the route
+        // may allocate a transparent deposit address that differs from the SwapKit hint). Fall back
+        // to the hint only if derivation fails.
+        val depositAddress = zcashAddress(outputKeyHashes[0]) ?: targetAddressHint
+        val changeAddress =
+            if (outputs.size >= 2) {
+                zcashAddress(outputKeyHashes[1]) ?: targetAddressHint
+            } else {
+                zcashAddress(inputs[0].keyHash) ?: targetAddressHint
+            }
```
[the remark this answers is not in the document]
aminsato: Keeping as-is — `targetAddress` binding is deliberately not applied on these UTXO-sibling paths, and requiring an output to equal it would reject valid swaps.

The deposit/change addresses are derived from the PSBT's own output hash160s, not from `targetAddress`. SwapKit returns `targetAddress` as the nominal deposit address, but the NEAR-Intents route allocates the actual on-chain deposit address, and the two differ on these chains — the ZEC path documents a captured fixture where output 0 lands on a route-allocated transparent address `!= targetAddress`. So `output == targetAddress` would be a false rejection, which is exactly why the address is used only as a derivation fallback (mirroring the iOS signer).

The integrity guarantee here comes from the frozen plan re-emitting the PSBT outputs verbatim (the broadcast tx is byte-identical to what was quoted), with the SwapKit `/v3/swap` response as the trust boundary — the same boundary as the amount and fee. The stronger destination binding on the BTC segwit path is only possible because BTC's deposit output equals `targetAddress`; it isn't portable here without per-chain fixtures confirming that relationship. An amount-level bound (deposit `<= fromAmount`, independent of the address) is a reasonable future hardening, but it's out of scope for this iOS-parity PR.

rkokhatskyi:
@aminsato please attach tx links / success screenshots (tx history)

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
```diff
@@ -0,0 +1,385 @@
...
+        val outCount = cursor.readCompactSize()
+        val outputs =
+            (0L until outCount).map {
+                val amount = cursor.readUInt64LE()
+                val scriptLen = cursor.readCompactSize()
+                SaplingOutput(amount, cursor.readBytes(cursor.asLength(scriptLen)))
+            }
+        cursor.readUInt32LE() // lockTime
+        cursor.readUInt32LE() // expiryHeight
```
rkokhatskyi: `parseSaplingUnsignedTx` reads `lockTime` (`321`) and `expiryHeight` (`322`) then discards both, while `assembleSigningInput` (`172`) never sets them on the `SigningInput`, so a route that ships a `lockTime` or `expiryHeight` other than zero, which Zcash transactions commonly carry, gets rebuilt by WalletCore with zeros and its broadcast `tx_id` no longer matches the PSBT body the NEAR Intents route is tracked by, silently breaking the tracking the class doc says this path preserves, so consider threading them onto the signing input or rejecting a value other than zero so the mismatch fails loudly instead of mistracking

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
```diff
@@ -0,0 +1,417 @@
...
+                ParsedLegacyTxInput(prevBytes, prevIndex, sequence)
+            }
+        val outCount = cursor.readCompactSize()
+        val outputs =
+            (0L until outCount).map {
+                val amount = cursor.readUInt64LE()
+                val scriptLen = cursor.readCompactSize()
+                LegacyP2PKHOutput(amount, cursor.readBytes(cursor.asLength(scriptLen)))
+            }
+        cursor.readUInt32LE() // locktime
```
rkokhatskyi: `parseLegacyUnsignedTx` discards the parsed `version` (`327`) and `locktime` (`352`), and `assembleSigningInput` (`200`) never calls `setLockTime` whereas the V1 `Bitcoin.SigningInput` cannot carry the tx version at all, so a SwapKit unsigned tx whose `locktime` is not zero or whose version differs from WalletCore's per coin default rebuilds to a different `tx_id` than the PSBT body and undercuts the byte for byte guarantee this class relies on, while `SwapKitBtcSigner` sidesteps this by serializing the tx itself from the parsed `version` and `locktime`, so consider preserving `locktime` via `setLockTime` and rejecting an unexpected version rather than dropping them

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
```diff
@@ -0,0 +1,417 @@
...
+        throw SwapKitLegacyP2PKHSignerException(
+            "SwapKit PSBT input #$inputIndex is missing a prev-tx UTXO record"
+        )
+    }
+
+    /**
+     * Parse a `PSBT_IN_NON_WITNESS_UTXO` record: the full previous transaction in legacy
+     * (pre-segwit) wire serialization. We extract `outputs[prevIndex]`.
+     */
+    private fun parseNonWitnessUtxo(
```
rkokhatskyi, line 291: `parseNonWitnessUtxo` takes the embedded prev tx `amount` and `scriptPubKey` on faith without verifying that double sha256 of that transaction equals the input's `prevTxIdLE`, which is the binding BIP 174 requires of a `NON_WITNESS_UTXO`, so a substituted or malformed record is accepted, and since BCH (FORKID) and ZEC (ZIP 243) fold the input `amount` into the sighash a wrong value there yields a signature the network rejects at broadcast rather than being caught up front, and the same gap sits in `SwapKitZcashSigner.parseNonWitnessUtxo` (`211`)

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
```diff
@@ -0,0 +1,417 @@
...
+            )
+        }
+        val outputKeyHashes =
+            outputs.mapIndexed { idx, out ->
+                assertP2PKHAndExtractKeyHash(out.scriptPubKey, idx, "output")
+            }
+        val totalIn = inputs.sumOf { it.amount }
+        val totalOut = outputs.sumOf { it.amount }
+        val fee = totalIn - totalOut
+        if (fee < 0) {
```
rkokhatskyi, line 171: The only value check here is `fee < 0`, whereas the BTC and LTC sibling routes through `UtxoHelper.getPreSignedImageHashFromSignBitcoin` which adds `verifyFeeCeiling` (caps the miner fee at `MAX_FEE_SAT_PER_VBYTE`) and `verifyDestinationBinding` (deposit not exceeding `fromAmount`), so a route that under allocates the change output is signed verbatim here and the unbounded remainder is burned to the miner on DOGE/BCH/DASH/ZEC where BTC/LTC would refuse it, and since `fromAmount` is already in scope at the dispatch site, being passed to `SwapKitBtcSigner` a few lines above, an amount or fee ceiling bound is feasible here independent of the deposit address binding that was intentionally left out

data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSignerTest.kt
```diff
@@ -0,0 +1,424 @@
...
+                            )
+                        ),
+                    outputs =
+                        listOf(
+                            LegacyOut(amount = 60_000, scriptHex = p2pkh("22".repeat(20))),
+                            LegacyOut(amount = 39_000, scriptHex = p2pkh("11".repeat(20))),
+                        ),
+                )
+            val hashes = signer(CoinType.DOGECOIN).getPreSignedImageHash(psbt, "")
+            assertEquals(1, hashes.size)
```
rkokhatskyi, line 261: These happy paths assert only `hashes.size` and length and wrap in `skipIfJniUnavailable`, so without the WalletCore JNI they are skipped rather than run, and no test calls `getSignedTransaction` for the legacy or ZEC signers or pins the reconstructed tx bytes or `tx_id`, so the byte for byte reconstruction this PR leans on is effectively unverified in CI, while `SwapKitBtcSignerTest` pins the broadcast bytes and a golden sighash vector headlessly, so consider adding an equivalent byte pinning test for at least DOGE and ZEC

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -475,13 +475,22 @@ constructor(
...
+                                // BTC/LTC segwit PSBT, DOGE/BCH/DASH legacy-P2PKH PSBT, ZEC
+                                // Sapling-v4 PSBT, TRON (TronWeb object), SUI (PTB), TON (native
                                 // transfer), XRP (deposit-only native Payment), and both Cardano
                                 // flows (deposit-only CARDANO + pre-built CARDANO_PREBUILT) are
                                 // wired with their per-chain signers/native paths. Guarded loudly
                                 // so an un-wired txType can't reach signing.
                                 require(
                                     quote.data.txType == SwapKitSwapPayloadJson.TX_TYPE_PSBT ||
+                                        quote.data.txType ==
+                                            SwapKitSwapPayloadJson.TX_TYPE_PSBT_DOGE ||
```
rkokhatskyi: This `require` allowlist duplicates the wired txType knowledge that `SigningHelper` already dispatches on, and the two drifted within this PR since the legacy and ZEC discriminators reached signing yet were absent here until this commit, so consider deriving the gate from one source such as a `SwapKitSwapPayloadJson` predicate that both the dispatcher and this guard consult, so a future chain cannot satisfy one path and trip the other

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
aminsato: Good catch — fixed in 866b7bf58.

- **lockTime:** `Bitcoin.SigningInput` does expose `setLockTime`, so I now thread the parsed lockTime onto the signing input in **both** this ZEC signer and `SwapKitLegacyP2PKHSigner` (DOGE/BCH/DASH had the same latent gap — they also discarded locktime). The rebuilt tx now reproduces it.
- **expiryHeight:** WalletCore's `Bitcoin.SigningInput` has no expiryHeight field (only `lock_time`, `time`, and `branch_id` on the plan), so a non-zero expiryHeight genuinely can't be threaded through — the rebuilt tx would carry expiryHeight 0 and a diverging tx_id. Per your second option I now hard-reject a non-zero expiryHeight loudly in `parseSaplingUnsignedTx` rather than silently mistrack, with a test pinning it. The verified DASH/DOGE/BCH/LTC broadcasts all carried lockTime 0, and the captured ZEC fixture is expiryHeight 0, so this doesn't block the routes in practice; if a future route ships a non-zero expiry it fails loudly instead.

aminsato:
> @aminsato please attach tx links / success screenshots (tx history)

Attached in the description

data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSignerTest.kt
aminsato: Added byte-pinning for DOGE and ZEC in 0186a557b.

A caveat on why it can't be a headless golden like `SwapKitBtcSignerTest`: that test works headlessly because the BTC signer hand-rolls BIP-143 + segwit serialization in pure Kotlin and exposes a `String`-keyed `compileSignedTransaction` seam. The legacy/ZEC signers deliberately delegate sighash + compile to WalletCore's `TransactionCompiler` (legacy ECDSA / ZIP-243) — reimplementing those in Kotlin purely to test would be a large parallel implementation of the exact thing we delegate to avoid — and that compile path takes the `compileOnly` `tss.KeysignResponse`, which isn't on the unit-test classpath at all.

So instead I pin the thing that fully determines the broadcast tx: the serialized `Bitcoin.SigningInput` the frozen plan produces. Building it needs the JNI (so the test runs in CI, skips locally like the others), but parsing the proto back is pure and every expectation is derived from the PSBT — no precomputed golden. The new tests assert the reconstruction reproduces, field by field: the input outpoint (LE hash + index + sequence), amount, and exact prevout scriptPubKey; the frozen plan amount/change/fee/availableAmount (+ ZEC's `branchId` `f04dec4d`); the threaded `lockTime`; and the deposit/change addresses derived from the PSBT output hash160s. If WalletCore re-emits those fields it produces the same tx_id, which the four mainnet broadcasts in the PR description (DOGE/BCH/DASH/LTC, tracked tx_ids matched) already confirm end-to-end.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
aminsato: Fixed in 2db0c54ee. `parseNonWitnessUtxo` now verifies `sha256d(prevTx) == prevTxIdLE` (the BIP-174 binding) before trusting the embedded amount/script, in both the legacy and ZEC signers. Added a headless test that points the outpoint at an unrelated txid while embedding a well-formed prev-tx and asserts the rejection.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
aminsato: Fixed in 2db0c54ee. Added a fee ceiling threading `fromAmount` from the dispatch site to both signers. A fixed sat/vByte ceiling like BTC's `verifyFeeCeiling` isn't portable here — DOGE's legitimate fees run orders of magnitude higher (the verified DOGE swap paid ~0.47 DOGE ≈ 100k+ sat/vByte, which the BTC 1000 sat/vByte cap would reject) — so I bound the miner fee by the quoted swap amount instead: a fee larger than the amount being swapped is refused. That's the balance-burn guard, independent of the deposit-address binding (still intentionally left out for the route-allocated-address reason). I left out a deposit≤fromAmount bound because it depends on which output is the deposit (we don't bind the source/change address), whereas the fee bound is ordering- and address-independent. Headless test added.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in 2db0c54ee. Extracted the allowlist into `SwapKitSwapPayloadJson.SIGNABLE_TX_TYPES` / `isSignableTxType(...)`, and the SwapFormViewModel pre-flight gate now consults that instead of re-listing the constants inline — so the guard tracks the canonical wired-txType set in the model rather than drifting from it (which is exactly what happened this PR).

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
```diff
@@ -0,0 +1,475 @@
...
+        val outputs: List<LegacyP2PKHOutput>,
+    )
+
+    /**
+     * Parse the legacy (non-witness) unsigned-tx body BIP-174 stores under global key 0x00. DOGE /
+     * BCH / DASH have no segwit, so the body never carries a marker+flag.
+     */
+    private fun parseLegacyUnsignedTx(data: ByteArray): ParsedLegacyTx {
+        val cursor = PsbtCursor(data)
+        cursor.readUInt32LE() // version
```
rkokhatskyi: The parsed `version` is still read and dropped here without a guard, whereas `lockTime` is now reproduced via `setLockTime` (`405`) and ZEC `expiryHeight` is now rejected when WalletCore can't reproduce it, so a legacy PSBT whose `version` differs from the per coin default WalletCore emits still rebuilds to a different `tx_id` than the PSBT body and silently breaks the NEAR route tracking, and since the V1 `Bitcoin.SigningInput` has no version field to set the consistent move is to assert the parsed `version` matches what WalletCore emits for the coin or reject it like the ZEC `expiryHeight` path rather than dropping it

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitLegacyP2PKHSigner.kt
aminsato: Fixed in 546a8e0be. `parseLegacyUnsignedTx` now asserts the unsigned-tx version equals what WalletCore reproduces on the legacy path and rejects anything else (mirroring the ZEC expiryHeight guard) — `Bitcoin.SigningInput` has no version field, so threading it isn't possible.

The reproducible value is **1**: I read it off the three verified mainnet broadcasts (DOGE `2d1249e2…`, BCH `c44fe710…`, DASH `f837a5c1…`) — all version 1 on-chain, and since their tx_ids matched what NEAR tracked, WalletCore's fixed emission is version 1 for this path. A v2 PSBT now fails loudly instead of mistracking. Headless test added.

rkokhatskyi (approved):
LGTM

Issue #609, opened 2024-07-21
BUG: Update Proto bytes compression algorithm to LZMA


johnnyluo:
**Describe the bug**
Recently PR https://github.com/vultisig/vultisig-ios/pull/885/files in vultisig-ios repo get merged , it changed the compression to use `lzma`

Please update the compression in android to lzma as well , so both devices can work with each other

PR #4716, opened 2026-06-02
fix(polkadot): confirm tx status via RPC block-scan, drop Subscan

### LINKED ISSUE #4684: Polkadot tx status never confirms: Subscan extrinsic API now requires an API key
## Summary

Polkadot / Asset Hub transaction status can never resolve to **Confirmed** (or **Failed**). Every status check now hits Subscan's authentication wall and degrades to `TimedOut`, even for transactions that are successfully included on-chain. The in-app history/status UI therefore shows sent DOT transfers as never confirming.

## Root cause

`PolkadotApi.getTxStatus()` queries the Subscan extrinsic endpoint directly, with no API key:

- `data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt:188` — `getTxStatus()` POSTs to
- `data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt:197-198` — `POLKADOT_EXTRINSIC_API_URL = https://assethub-polkadot.api.subscan.io/api/scan/extrinsic`

Subscan now hard-requires an API key on this endpoint. An unauthenticated request returns:

```
HTTP 400
{"code": 403, "message": "Subscan API strictly requires an API key. Unauthenticated access is disabled."}
```

`PolkadotStatusProvider` already detects this auth wall and degrades gracefully so we don't poll forever (`PolkadotStatusProvider.kt:38-51`, classifying it as `TimedOut`). That handling is correct as a safety net, but the underlying lookup **always** fails, so the status flow can only ever produce `TimedOut` / `Pending` / `NotFound` — never `Confirmed`. The graceful fallback is masking a fully non-functional status path.

Note this is broadcast-vs-status split: broadcasting goes through the proxied `api.vultisig.com/dot/` RPC (works fine), while status goes directly to the public Subscan API (now blocked).

## Reproduction

1. Send any Polkadot/Asset Hub transfer from the app.
2. Broadcast succeeds; `author_submitExtrinsic` returns the extrinsic hash.
3. The status check calls `getTxStatus(hash)` → Subscan 400/403 auth wall → `TimedOut`.
4. The transaction is actually included on-chain (verifiable independently: the sender's on-chain `system_accountNextIndex` on Asset Hub advances past the tx's nonce), but the app never reflects Confirmed.

## Suggested fix (either)

1. **Proxy + key (preferred, consistent with broadcast):** route status lookups through a Vultisig-hosted endpoint that injects the Subscan API key server-side (same pattern as `api.vultisig.com/dot/`), so the key never ships in the client. Then `getTxStatus()` hits that proxy instead of `assethub-polkadot.api.subscan.io` directly.
2. **RPC-based confirmation (no Subscan):** drop the Subscan dependency for confirmation and verify inclusion via the chain RPC we already use — e.g. poll the account nonce / block for the extrinsic — so status no longer depends on a third-party indexer's auth tier.

## Affected files

- `data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt` (`getTxStatus`, `POLKADOT_EXTRINSIC_API_URL`)
- `data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/PolkadotStatusProvider.kt` (consumer; auth-wall handling stays as the safety net)

aminsato:
## Summary

Fixes #4684. Polkadot / Asset Hub transactions could never resolve to **Confirmed** — `getTxStatus` queried Subscan, which now hard-requires an API key and rejects unauthenticated traffic, so every status check degraded to `TimedOut` even for transactions successfully included on-chain.

Per the maintainer's decision on the issue (**Option 2 — drop the assethub/Subscan dependency**), status is now confirmed purely over the RPC we already broadcast through. Substrate exposes no "transaction by hash" RPC, so confirmation walks recent best-head blocks via `chain_getBlock` and blake2b-256-matches each extrinsic against the tx hash.

For reference: the SDK issue mentioned on #4684 (vultisig-sdk#609) does **not** remove Subscan — its status resolver still queries it; #609 fixed an unrelated broadcast race. So there was no RPC implementation to port.

## Changes

- **`PolkadotHelper`** — fix a latent bug: `transactionHash` hashed only the first 32 bytes of the encoded extrinsic (`output.encoded.take(32)`). The canonical Substrate hash is blake2b-256 over the **full** extrinsic (as the sibling `BittensorHelper` already does). The truncated value matched neither the on-chain extrinsic nor the explorer, breaking the tx link and any hash lookup. Required for block-scan to match.
- **`PolkadotApi`** — add `isExtrinsicInChain(hash, depth)` + `getBlock`; remove `getTxStatus` and the Subscan extrinsic URL.
- **`PolkadotStatusProvider`** — rewritten: extrinsic found in a recent block → `Confirmed`, otherwise `Pending` until the poll loop times out.
- **`BroadcastTxUseCase`** — the Polkadot broadcast safety net (verify-by-hash) also relied on the broken Subscan call; switch it to the same block-scan.
- Drop the now-unused Subscan response models.

## Known limitation

Metadata-free SCALE event decoding is not feasible over plain RPC, so an *included-but-dispatch-failed* extrinsic reports `Confirmed` rather than `Failed`. Acceptable for transfers, where balance and fee are pre-checked before signing.

## Test plan

- [x] `:data:testDebugUnitTest` (full suite) green
- [x] `:app:compileDebugKotlin` green
- [x] New `PolkadotApiTest` — block-scan + blake2b-256 match against a real extrinsic vector; walk-to-parent; depth bound
- [x] Rewrote `PolkadotStatusProviderTest`
- [ ] Manual mainnet smoke: send a small DOT transfer → status flips to Confirmed; the tx hash now resolves on Subscan/polkadot.js

## Proof (device, Polkadot mainnet)

Real DOT transfer signed on device — status now resolves to **Successful** (previously stuck pending → timed out).

## Summary by CodeRabbit

* **New Features**
  * Added direct on-chain block scanning to verify Polkadot extrinsic inclusion (head-relative and absolute block-range checks).
  * Persist broadcast block number when saving transactions to support absolute inclusion windows.

* **Bug Fixes**
  * Corrected canonical Polkadot extrinsic hash to match explorers.
  * Status polling now relies on on-chain scans, treats transient errors as pending, and rethrows cancellations; removed prior explorer-specific timeout behavior.
  * Scan depth computed from broadcast age with min/max bounds.

* **Tests**
  * Added/updated tests for block-scanning, block-range checks, scan-depth, error handling, and cancellations.

* **Chores**
  * Database migration and schema version bump to add broadcast block column.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
@@ -866,6 +866,14 @@ constructor(
                         confirmedAt = null,
                         failureReason = null,
                         lastCheckedAt = now,
+                        // Polkadot extrinsics are mortal: persist the head block at broadcast so
+                        // the
+                        // status poller can scan the absolute inclusion window instead of a
+                        // head-relative one that drifts out of reach. Null for other chains.
+                        broadcastBlockNumber =
+                            (keysignPayload?.blockChainSpecific as? BlockChainSpecific.Polkadot)
+                                ?.currentBlockNumber
+                                ?.toLong(),
                     )
                 transactionHistoryRepository.recordTransaction(
                     vaultId = vault.id,
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
@@ -239,10 +239,21 @@ internal class PolkadotApiImp @Inject constructor(private val httpClient: HttpCl
         // Anchor at the top of the window and descend via parentHash. Block numbers are sequential
         // in Substrate, so the parent of block N is N-1 — we count down instead of re-reading the
         // header number each hop.
-        var blockHash: String? = getBlockHashByNumber(toBlock) ?: return false
+        //
+        // A null hash/block mid-walk means the RPC returned an error envelope (postRpc reads a
+        // nullable result), not that the extrinsic is absent. Returning false here would be
+        // indistinguishable from a genuine miss, so checkByInclusionWindow could terminally Fail a
+        // confirmed transfer once the head passes the window. Instead we throw on an incomplete
+        // scan so it propagates to checkStatus and the tx stays Pending for a later retry; false is
+        // returned only after the full window was traversed without a match.
+        var blockHash: String =
+            getBlockHashByNumber(toBlock)
+                ?: error("Polkadot block hash unavailable for $toBlock; inclusion scan incomplete")
         var current = toBlock
         while (current >= fromBlock) {
-            val block = getBlock(blockHash)?.block ?: break
+            val block =
+                getBlock(blockHash)?.block
+                    ?: error("Polkadot block $current unavailable; inclusion scan incomplete")
             if (block.extrinsics.any { extrinsicHash(it) == target }) {
                 return true
             }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
@@ -62,8 +62,22 @@ interface PolkadotApi {
      * contains an extrinsic whose blake2b-256 hash matches [txHash]. Substrate full nodes do not
      * index extrinsics by hash, so confirmation is done by scanning recent blocks via plain RPC
      * instead of depending on a third-party indexer (e.g. Subscan).
+     *
+     * Head-relative, so only suitable right after broadcast (the inclusion block is near the head).
+     * For status polling that may run long after broadcast, use [isExtrinsicInBlockRange], whose
+     * window is anchored to an absolute block and does not drift as the head advances.
      */
     suspend fun isExtrinsicInChain(txHash: String, depth: Int): Boolean
+
+    /**
+     * Returns true if an extrinsic whose blake2b-256 hash matches [txHash] appears in any block in
+     * the inclusive number range `[fromBlock, toBlock]`. The range is anchored at [toBlock] (which
+     * the caller caps at the live head) and walked backward via `parentHash` down to [fromBlock] —
+     * Substrate blocks only link to their parent and number sequentially, so the absolute window is
+     * scanned regardless of how far the head has since advanced past it. This keeps a confirmed but
+     * already-buried inclusion block reachable, unlike a head-relative scan.
+     */
+    suspend fun isExtrinsicInBlockRange(txHash: String, fromBlock: Long, toBlock: Long): Boolean
 }
 
 internal class PolkadotApiImp @Inject constructor(private val httpClient: HttpClient) :
@@ -213,6 +227,40 @@ internal class PolkadotApiImp @Inject constructor(private val httpClient: HttpCl
         }
         return false
     }
+
+    override suspend fun isExtrinsicInBlockRange(
+        txHash: String,
+        fromBlock: Long,
+        toBlock: Long,
+    ): Boolean {
+        val target = txHash.removePrefix("0x").lowercase()
+        if (target.isEmpty() || toBlock < fromBlock) return false
+
+        // Anchor at the top of the window and descend via parentHash. Block numbers are sequential
+        // in Substrate, so the parent of block N is N-1 — we count down instead of re-reading the
+        // header number each hop.
+        var blockHash: String? = getBlockHashByNumber(toBlock) ?: return false
+        var current = toBlock
+        while (current >= fromBlock) {
+            val block = getBlock(blockHash)?.block ?: break
+            if (block.extrinsics.any { extrinsicHash(it) == target }) {
+                return true
+            }
+            blockHash = block.header.parentHash
+            current--
+        }
+        return false
+    }
+
+    private suspend fun getBlockHashByNumber(number: Long): String? =
+        httpClient
+            .postRpc<PolkadotGetBlockHashJson>(
+                url = POLKADOT_API_URL,
+                method = "chain_getBlockHash",
+                params = buildJsonArray { add(number) },
+            )
+            .result
+            .takeIf { it.isNotBlank() }
 
     private suspend fun getBlock(blockHash: String?): PolkadotBlockResultJson? {
         return httpClient
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
@@ -2,14 +2,17 @@ package com.vultisig.wallet.data.api
 
 import com.vultisig.wallet.data.api.models.PolkadotGetStorageJson
 import com.vultisig.wallet.data.api.models.RpcPayload
+import com.vultisig.wallet.data.api.models.cosmos.PolkadotBlockResultJson
 import com.vultisig.wallet.data.api.models.cosmos.PolkadotBroadcastTransactionJson
-import com.vultisig.wallet.data.api.models.cosmos.PolkadotExtrinsicResponseJson
 import com.vultisig.wallet.data.api.models.cosmos.PolkadotGetBlockHashJson
 import com.vultisig.wallet.data.api.models.cosmos.PolkadotGetBlockHeaderJson
+import com.vultisig.wallet.data.api.models.cosmos.PolkadotGetBlockJson
 import com.vultisig.wallet.data.api.models.cosmos.PolkadotGetNonceJson
 import com.vultisig.wallet.data.api.models.cosmos.PolkadotGetRunTimeVersionJson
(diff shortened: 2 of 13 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
```diff
@@ -185,17 +194,44 @@ internal class PolkadotApiImp @Inject constructor(private val httpClient: HttpCl
...
+    override suspend fun isExtrinsicInChain(txHash: String, depth: Int): Boolean {
+        val target = txHash.removePrefix("0x").lowercase()
+        if (target.isEmpty()) return false
+
+        // Start from the best head (params omitted) and follow parentHash links. Inclusion in the
+        // best chain is enough to treat the transaction as on-chain; GRANDPA finalizes within a
+        // couple of blocks so a reorg deep enough to drop it is not a practical concern here.
+        var blockHash: String? = null
+        var scanned = 0
+        while (scanned < depth) {
```
rkokhatskyi, line 220: `isExtrinsicInChain` always starts at the live best head and walks back only `depth` blocks, so an extrinsic included earlier than `SCAN_DEPTH` blocks ago can never be matched, and since `RefreshPendingTransactionsUseCase` rechecks pending rows minutes after broadcast a confirmed Polkadot transfer the 300s foreground poll missed stays `Pending` forever

```kotlin
// scan a block number range, not a head window: persist the best head
// number at broadcast, then walk [broadcastNumber .. head]; or cache the
// inclusion block on first match so later refreshes short circuit
```

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/PolkadotStatusProvider.kt
```diff
@@ -2,51 +2,74 @@ package com.vultisig.wallet.data.api.txstatus
...
+        // and block-time jitter so the inclusion block is never just outside the window.
+        const val SCAN_MARGIN_BLOCKS = 10L
+        // Floor for the fresh-broadcast / unknown-broadcast-time case (extrinsic at or near the
+        // head); keeps the common foreground poll to a couple of block fetches via early-exit.
+        const val MIN_SCAN_DEPTH = 10
+        // Ceiling that bounds the walk: the 64-block mortal era (after which the extrinsic is
+        // permanently invalid) plus the ~50-block foreground poll window, with margin. A row not
+        // confirmed within this window is dropped, not still confirmable, so a deeper scan would
+        // only re-fetch buried history for nothing.
+        const val MAX_SCAN_DEPTH = 130
```
rkokhatskyi, line 121: The `MAX_SCAN_DEPTH` cap still leaves a confirmed transfer stuck `Pending`, since the scan walks backward from the live head and an already included block recedes without bound while the mortal era only limits inclusion not findability. A row first rechecked beyond ~130 blocks is never reached, and nothing demotes it

```kotlin
// scan forward across the mortal era from the broadcast block instead of
// backward from the head: walk [currentBlockNumber .. currentBlockNumber + 64]
// (currentBlockNumber is already in PolkadotBlockChainSpecific), which is
// bounded and age independent, then mark a miss past that window terminal
```

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/PolkadotStatusProvider.kt
aminsato: Done in 954569103 — implemented exactly this. The scan is now anchored to an absolute block instead of the head:

- Persist the broadcast block number (`BlockChainSpecific.Polkadot.currentBlockNumber`, the era checkpoint — `PolkadotHelper` signs with `setPeriod(64)`) on the tx row. New nullable `broadcastBlockNumber` column, DB v32→33 (`MIGRATION_32_33`), threaded through `CommonTransactionHistoryData` → `KeysignViewModel.saveTransactionHistory`.
- `PolkadotStatusProvider` now scans the fixed window `[anchor, min(head, anchor + 64)]` via the new `isExtrinsicInBlockRange` — bounded and age-independent, so a recheck minutes/hours later still hits the right blocks.
- Past the window it's terminal: once `head > anchor + 64 + 10` (finality margin) without inclusion the extrinsic can never be included, so it resolves to `Failed("Extrinsic expired…")` instead of polling `Pending` forever.
- Legacy rows with no persisted block fall back to the previous head-relative scan.

data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
aminsato: Fixed in 954569103. The status path no longer relies on a head-relative walk: it uses the new `isExtrinsicInBlockRange(txHash, fromBlock, toBlock)`, which resolves `toBlock` to a hash via `chain_getBlockHash(number)` and walks down `parentHash` to `fromBlock`, scanning the absolute mortal-era window `[broadcastBlock, broadcastBlock + 64]`. So a confirmed-but-buried inclusion block stays reachable however long after broadcast the recheck runs, and a miss past the era is now demoted to terminal `Failed`.

`isExtrinsicInChain(depth)` is kept only for `BroadcastTxUseCase`'s immediate post-broadcast recovery, where the extrinsic is at/near the head and a shallow head-relative scan is correct.

data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
```diff
@@ -185,17 +208,78 @@ internal class PolkadotApiImp @Inject constructor(private val httpClient: HttpCl
...
+        val target = txHash.removePrefix("0x").lowercase()
+        if (target.isEmpty() || toBlock < fromBlock) return false
+
+        // Anchor at the top of the window and descend via parentHash. Block numbers are sequential
+        // in Substrate, so the parent of block N is N-1 — we count down instead of re-reading the
+        // header number each hop.
+        var blockHash: String? = getBlockHashByNumber(toBlock) ?: return false
+        var current = toBlock
+        while (current >= fromBlock) {
+            val block = getBlock(blockHash)?.block ?: break
```
rkokhatskyi: A `getBlock` returning null partway through breaks the walk and returns the same `false` as a genuine miss, so `checkByInclusionWindow` can mark a confirmed transfer terminally `Failed` once the head passes the window. `getBlock` yields null rather than throwing on an RPC error envelope, since `postRpc` reads `.body()` and `result` is nullable

```kotlin
// distinguish an incomplete scan from a real miss: throw on a null
// block fetch mid window so it propagates to checkStatus and stays
// Pending, and declare Failed only when the full window was traversed
```

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/api/PolkadotApi.kt
aminsato: Good catch — fixed in 43ddbbb5d. `isExtrinsicInBlockRange` no longer conflates an incomplete scan with a genuine miss:

- A null block (or null `getBlockHashByNumber`) mid-window now `throw`s instead of `break`ing to `false`. Since `postRpc` reads a nullable `result`, an RPC error envelope yielded null rather than throwing, so the old `?: break` was indistinguishable from "extrinsic absent".
- The throw propagates to `PolkadotStatusProvider.checkStatus`, whose `catch` returns `Pending` — so an RPC hiccup keeps the tx Pending for a later retry instead of letting `checkByInclusionWindow` terminally `Failed` it once the head passes the window.
- `false` is now returned **only** after the full `[fromBlock, toBlock]` window was actually traversed without a match, so the legitimate terminal-`Failed` path is preserved.

Added a regression test (`isExtrinsicInBlockRange throws on a null block mid-window instead of reporting a miss`).

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -866,6 +866,14 @@ constructor(
...
                         failureReason = null,
                         lastCheckedAt = now,
+                        // Polkadot extrinsics are mortal: persist the head block at broadcast so
+                        // the
+                        // status poller can scan the absolute inclusion window instead of a
+                        // head-relative one that drifts out of reach. Null for other chains.
+                        broadcastBlockNumber =
+                            (keysignPayload?.blockChainSpecific as? BlockChainSpecific.Polkadot)
+                                ?.currentBlockNumber
+                                ?.toLong(),
```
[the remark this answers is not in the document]
johnnyluo: Keeping `.toLong()` here intentionally. Polkadot block numbers are u32-derived and currently in the tens of millions, incrementing ~once per 6s — `Long.MAX_VALUE` is ~290 billion years away, so truncation is physically impossible. Switching to `longValueExact()` would only introduce a throw on an unreachable overflow, and since this runs inside the transaction-history save path that throw would drop the entire history record — strictly worse than the current behavior. No change needed.

johnnyluo (approved):
LGTM

PR #4718, opened 2026-06-02
fix(notifications): Join banner navigates on tap + swipe-to-dismiss (#4623)

### LINKED ISSUE #4623: Join transaction banner: tap clears banner but does not navigate
## Summary

When the in-app "Join transaction" banner appears over an active flow (reproduced on the transaction-complete / `SendTxOverviewScreen`), tapping the banner:

1. **Does not navigate** to the Join Transaction (Join Keysign) screen.
2. **Dismisses the banner immediately**, leaving the user with no way to act on the signing request and no visible feedback.

The banner is the only entry point the user has for that incoming session, so once it is gone they are effectively stuck — the inviting device keeps waiting for a participant that will never join.

## Steps to reproduce

1. Open the app and complete a Send transaction so you land on the post-broadcast overview screen (the screen with the Signature card and the "Complete" button).
2. From a second device/vault, initiate a Keysign and have it broadcast the join QR / push so the current device receives a "Join transaction" foreground notification.
3. With the banner showing ("Join transaction" / vault name pill / "A new signing request is waiting for you"), tap it.

**Expected:** App navigates to the Join Keysign screen for the requested session.
**Actual:** Banner slides out and disappears. No navigation occurs. User remains on the transaction-complete screen.

## Root cause

The banner's tap handler clears the foreground-notification state *synchronously* and only then dispatches the navigation event, which is asynchronous.

`app/src/main/java/com/vultisig/wallet/app/activity/MainViewModel.kt` (`onForegroundBannerTapped`):

```kotlin
fun onForegroundBannerTapped() {
    val qrCodeData = _foregroundNotification.value?.qrCodeData ?: return
    clearForegroundNotification()           // sets state = null immediately
    onPushNotificationReceived(qrCodeData)  // navigation dispatched async
}
```

Banner visibility is bound to that state in `MainActivityContent.kt` (`AnimatedVisibility(visible = foregroundNotification != null)`), so the moment the handler runs the banner is gone — regardless of whether the subsequent navigation actually succeeds.

When the user is mid-flow (e.g. inside a nested keysign/send navigator on `SendTxOverviewScreen`), the navigation event has no fallback: if the destination is filtered, dropped, or the current backstack isn't listening, the user ends up with no banner *and* no destination.

The intended pattern is already used elsewhere in `MainActivityContent.kt` — a route-change observer calls `clearForegroundNotification()` *after* `Route.Keysign.Join` (or `Keygen.Join`) is reached. The banner tap handler should rely on that observer rather than pre-emptively clearing state.

## Proposed fix

1. Remove the eager `clearForegroundNotification()` call in `onForegroundBannerTapped()` and let the existing route-change observer clear it on successful navigation.
2. Audit `onPushNotificationReceived` to confirm it can navigate to `Keysign.Join` even when the user is currently inside a different nested flow (the keysign backstack may need to be popped before the join route is pushed).
3. Add a safety timeout so that if navigation does not land on `Keysign.Join` within N ms after a banner tap, the banner is re-shown rather than lost — protects against future regressions of the same shape.

## Severity

High — silent failure of the only entry point for an incoming signing session. The user cannot recover without the inviting device retrying.

aminsato:
Closes #4623.

## Problem

The in-app "Join transaction" foreground banner cleared its own state **synchronously** in `onForegroundBannerTapped()` before dispatching the (async) navigation:

```kotlin
fun onForegroundBannerTapped() {
    val qrCodeData = _foregroundNotification.value?.qrCodeData ?: return
    clearForegroundNotification()           // state = null immediately
    onPushNotificationReceived(qrCodeData)  // navigation dispatched async
}
```

Banner visibility is bound to that state (`AnimatedVisibility(visible = foregroundNotification != null)`), so the moment the handler ran the banner disappeared — **regardless of whether navigation succeeded**. When the user was mid-flow (e.g. on `SendTxOverviewScreen` inside a nested keysign/send navigator) and the navigation event was dropped, they were left with **no banner and no destination**, unable to act on the incoming signing request. The inviting device then waits forever for a participant that never joins.

## Fix

- **`MainViewModel.onForegroundBannerTapped()`** — no longer clears the banner eagerly. The existing route-change observer in `MainActivityContent` (watching `currentBackStackEntryFlow` for `Route.Keysign.Join` / `Route.Keygen.Join`) clears it **only once navigation actually lands**, so the request stays actionable until then.
- **`ForegroundNotificationBanner`** — added **horizontal swipe-to-dismiss**: dragging past ~35% of the banner width settles it off-screen and dismisses it; a shorter drag springs back. This gives the user a way to flick away a stale or unwanted notification, and deterministically recovers the banner if navigation never lands (instead of a fragile timeout).
- **`MainActivityContent`** — wires `onBannerDismiss -> clearForegroundNotification`.

### On navigating from a nested flow
The `route` / `destination` collectors run on the **root** `navController`, and `GetDirectionByQrCodeUseCase` returns a top-level `Route.Keysign.Join`, so `navigate()` pushes onto the root back stack regardless of which nested screen is showing. No backstack-popping change was required.

## Test plan

- [x] New unit test `onForegroundBannerTapped keeps banner visible until navigation lands` guards that tap no longer clears the banner synchronously.
- [x] `:app:testDebugUnitTest` (MainViewModelTest) — 7/7 green.
- [x] ktfmt applied.
- [ ] Manual two-device smoke: receive a Join banner on `SendTxOverviewScreen`, tap it → lands on Join Keysign screen; swipe an old banner left/right → it dismisses.

## Summary by CodeRabbit

* **New Features**
  * Added horizontal swipe-to-dismiss for foreground notification banners and an explicit banner dismiss callback.

* **Bug Fixes**
  * Banner taps for join flows now keep the banner visible until the join screen is reached; non‑join taps clear immediately.
  * Dismissing a banner cancels pending navigation to prevent unintended arrivals; missing-target taps show an error and clear the banner.

* **Tests**
  * Added unit tests for banner tap/dismiss behavior and a URI decoding test stub.


app/src/main/java/com/vultisig/wallet/app/activity/MainViewModel.kt
```diff
@@ -160,7 +160,12 @@ constructor(
...
     fun onForegroundBannerTapped() {
         val qrCodeData = _foregroundNotification.value?.qrCodeData ?: return
-        clearForegroundNotification()
+        // Do NOT clear the banner here. Clearing is bound to banner visibility, so an eager
+        // clear hides the banner before navigation lands — and if navigation is dropped (the
+        // user is inside a nested keysign/send flow), the user is left with no banner and no
+        // destination. The route-change observer in MainActivityContent clears the banner once
+        // Route.Keysign.Join / Keygen.Join is actually reached. Until then the banner stays so
+        // the signing request remains actionable; the user can also swipe it away to dismiss.
         onPushNotificationReceived(qrCodeData)
```
rkokhatskyi, line 170: Dropping the eager clear strands the banner whenever a tap resolves to anything but `Keysign.Join` or `Keygen.Join`, since the observer clears only those two, so a push for a missing vault hits the `vault == null` return at `MainViewModel.kt:182` and the banner lingers until a manual swipe whereas iOS clears on tap plus a 30s auto dismiss
```kotlin
if (vault == null) {
    snackbarFlow.showMessage(...)
    clearForegroundNotification()      // clear on the non navigating branch
    return@safeLaunch
}
navigator.route(getDirectionByQrCodeUseCase(qrCodeData, vault.id))
clearForegroundNotification()          // clear for any dispatched tap, not only Join arrivals
```

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
```diff
@@ -80,6 +80,7 @@ internal fun MainActivityContent(
         lastNotification = lastNotification,
         isOffline = mainViewModel.isOffline.collectAsStateWithLifecycle().value,
         onBannerTap = mainViewModel::onForegroundBannerTapped,
+        onBannerDismiss = mainViewModel::clearForegroundNotification,
```
rkokhatskyi, line 83: Swiping the banner away cancels only `foregroundPushJob`, not the navigation a prior tap already launched, since `onPushNotificationReceived` runs in an untracked `safeLaunch`, so a tap followed by a swipe during the `getByEcdsa` suspension still lands the user on the Join screen they tried to dismiss
```kotlin
private var navigationJob: Job? = null

fun onPushNotificationReceived(qr: String) {
    navigationJob?.cancel()
    navigationJob = viewModelScope.safeLaunch { /* ... */ }
}

fun clearForegroundNotification() {
    foregroundPushJob?.cancel()
    navigationJob?.cancel()
    _foregroundNotification.value = null
}
```

app/src/main/java/com/vultisig/wallet/ui/components/banners/ForegroundNotificationBanner.kt
```diff
@@ -74,16 +84,50 @@ internal fun ForegroundNotificationBanner(
...
+    LaunchedEffect(qrCodeData) { offsetX.snapTo(0f) }
+
     Box(
         modifier =
             modifier
                 .fillMaxWidth()
+                .offset { IntOffset(offsetX.value.roundToInt(), 0) }
+                .pointerInput(Unit) {
+                    val dismissThreshold = size.width * 0.35f
+                    detectHorizontalDragGestures(
```
rkokhatskyi, line 111: `detectHorizontalDragGestures` omits `onDragCancel`, so a drag interrupted by a second pointer or an arena steal runs neither the dismiss nor the spring back branch and leaves `offsetX` at its last partial value with the banner frozen off center, which the `qrCodeData` keyed reset cannot recover
```kotlin
// add this, or replace the manual Animatable with Modifier.anchoredDraggable /
// SwipeToDismissBox which handles cancel, fling and the mutex in one coroutine
onDragCancel = { scope.launch { offsetX.animateTo(0f) } },
```

app/src/main/java/com/vultisig/wallet/ui/components/banners/ForegroundNotificationBanner.kt
```diff
@@ -74,16 +84,50 @@ internal fun ForegroundNotificationBanner(
...
 
+    // Horizontal swipe-to-dismiss: lets the user flick away a stale or unwanted signing
+    // request (e.g. an old notification) without acting on it. Dragging past ~35% of the
+    // banner width settles it off-screen and reports the dismissal; a shorter drag springs back.
+    val offsetX = remember { Animatable(0f) }
+    val scope = rememberCoroutineScope()
+
+    // Reset any residual swipe offset when a new (or re-shown) request is bound,
+    // so a previously dismissed banner doesn't reappear off-screen.
+    LaunchedEffect(qrCodeData) { offsetX.snapTo(0f) }
```
rkokhatskyi, line 102: The `LaunchedEffect(qrCodeData)` reset never fires when the same `qrCodeData` reappears during the slide out window, since the key is unchanged and `key(lastNotification?.qrCodeData)` stays constant, so the retained `offsetX` keeps the dismiss value of size.width and the banner returns fully off screen
```kotlin
onDragEnd = {
    scope.launch {
        if (abs(offsetX.value) >= dismissThreshold) {
            offsetX.animateTo(if (offsetX.value > 0) size.width.toFloat() else -size.width.toFloat())
            onDismiss()
            offsetX.snapTo(0f)   // reset the retained node so a same qr re-show is centred
        } else offsetX.animateTo(0f)
    }
}
```

app/src/test/java/com/vultisig/wallet/app/activity/MainViewModelTest.kt
```diff
@@ -157,4 +157,26 @@ internal class MainViewModelTest {
...
+            vm.onForegroundPushReceived("vultisig://qr-payload")
+            advanceUntilIdle()
+            assertNotNull(vm.foregroundNotification.value)
+
+            // Tapping must NOT clear the banner synchronously — the route observer clears it
+            // only once Keysign.Join / Keygen.Join is reached, so the request stays actionable
+            // even if navigation is dropped while inside a nested flow (issue #4623).
+            vm.onForegroundBannerTapped()
+
+            assertNotNull(vm.foregroundNotification.value)
```
rkokhatskyi: The assertion proves only that tap does not clear synchronously, never the navigation the test name promises, since `getByEcdsa` is stubbed null at `MainViewModelTest.kt:165` and `onNavigationReady()` is never called, so `onPushNotificationReceived` parks at `_navigationReady.await()` and `navigator.route` is never reached
```kotlin
coEvery { vaultRepository.getByEcdsa(any()) } returns vault
coEvery { getDirectionByQrCodeUseCase(any(), any()) } returns Route.Keysign.Join(/* ... */)
val vm = createViewModel()
vm.onNavigationReady()
advanceUntilIdle()
vm.onForegroundPushReceived("vultisig://qr-payload"); advanceUntilIdle()
vm.onForegroundBannerTapped(); advanceUntilIdle()
coVerify { navigator.route(match { it is Route.Keysign.Join }) }
```

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/app/activity/MainViewModel.kt
aminsato: Good catch — confirmed: `getDirectionByQrCodeUseCase` also returns `Route.Send`/`Route.ScanError`, and the `vault == null` branch navigates nowhere, so the observer (which only clears on `Keysign.Join`/`Keygen.Join`) left those stranded.

Fixed in 174927fb5: the `vault == null` branch now clears the banner, and after dispatch any non-Join route clears it too. I kept Join routes on the observer rather than clearing unconditionally after `navigator.route(...)` — clearing eagerly there would reintroduce the original #4623 bug (banner gone before deferred nested-flow nav lands). iOS-style tap+30s auto-dismiss can be a follow-up.

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
aminsato: Right — the nav coroutine was untracked, so a swipe during the `getByEcdsa` suspension could not cancel it. Fixed in 174927fb5: it is now stored in `navigationJob`, cancelled at the start of each `onPushNotificationReceived` and in `clearForegroundNotification()`, so swipe-to-dismiss cancels an in-flight tap navigation.

app/src/main/java/com/vultisig/wallet/ui/components/banners/ForegroundNotificationBanner.kt
aminsato: Added `onDragCancel = { scope.launch { offsetX.animateTo(0f) } }` in 174927fb5 so an arena steal / second pointer springs the banner back instead of freezing it off-center. Kept the manual `Animatable` rather than switching to `anchoredDraggable`/`SwipeToDismissBox` to keep the change minimal.

app/src/main/java/com/vultisig/wallet/ui/components/banners/ForegroundNotificationBanner.kt
aminsato: Agreed — `LaunchedEffect(qrCodeData)` does not re-fire when the same session re-broadcasts (key unchanged), so the retained `size.width` offset would keep the banner off-screen. Fixed in 174927fb5 by adding `offsetX.snapTo(0f)` right after `onDismiss()` in `onDragEnd`, resetting the retained node for a same-qr re-show.

app/src/test/java/com/vultisig/wallet/app/activity/MainViewModelTest.kt
aminsato: Correct — the test parked at `_navigationReady.await()` (never called `onNavigationReady()`, `getByEcdsa` stubbed null), so it only proved the no-sync-clear. Rewrote it in 174927fb5 to call `onNavigationReady()`, stub a vault + `Route.Keysign.Join`, `coVerify { navigator.route(ofType(Route.Keysign.Join::class)) }`, and still assert the banner stays (observer-owned clear). Also added a second test for the missing-vault path (banner cleared, `navigator.route` not called). Needed a `Uri.decode` static mock for `DeepLinkHelper`'s `?vault=` parsing.

johnnyluo:
@aminsato I don't think this PR fixed the issue.  I run some tests manually
1. Send 0.1 RUNE from Vault A -> B
2. When finished , my pair device stay in transaction done screen , currently polling transaction status
3. Go to initiate device , send another 0.1 RUNE from A -> B
4. When the notification arrives , I click the banner (notification) , my pair device stay in the transaction done screen , and it didn't join the transaction and go to pair screen

Can you please do some more tests with it

aminsato:
> @aminsato I don't think this PR fixed the issue. I run some tests manually
> 
> 1. Send 0.1 RUNE from Vault A -> B
> 2. When finished , my pair device stay in transaction done screen , currently polling transaction status
> 3. Go to initiate device , send another 0.1 RUNE from A -> B
> 4. When the notification arrives , I click the banner (notification) , my pair device stay in the transaction done screen , and it didn't join the transaction and go to pair screen
> 
> Can you please do some more tests with it

Fixed

johnnyluo (approved):
LGTM

PR #4721, opened 2026-06-02
feat(swap): instant indicative rate, no-flicker refresh, instant on %/paste

### LINKED ISSUE #4712: Swap quote UX: instant indicative rate, no-flicker refresh, instant on percentage/paste

aminsato:
Closes #4712

Three display-layer enhancements so the swap quote feels instant. **None change the signed transaction** — the firm provider quote still builds the tx and gates Continue.

### 1. Instant indicative rate
`SwapQuoteManager.computeIndicativeQuote` derives `dst = srcAmount × priceSrc / priceDst` from **already-cached** spot prices (`TokenPriceRepository.getCachedPrice`). It never hits the network — a cold price returns `null` and the previous value is kept. The form shows the estimate **greyed** the moment an amount is entered, then the firm quote replaces it (un-greyed).

### 2. No-flicker refresh (stale-while-revalidate)
The loading + indicative side effects sit **upstream** of the `refreshQuoteState` trigger in the quote flow, so background refreshes (expiry timer, gas bump) never flash the spinner or blank the destination. The skeleton now only shows on the **first** quote for a pair (no cached price to estimate from).

### 3. Instant on percentage / Max / paste
Percentage / Max taps and pastes bypass the typing debounce (`0ms`); free typing still coalesces at `300ms`, so no extra provider requests fire.

## Before / After

Mid-load snapshot (amount entered, firm quote still resolving):

| Before | After |
|--------|-------|
| before | after |

Before: the destination amount + fiat blank to grey skeletons every cycle. After: the destination keeps a greyed indicative value (`12.80 / $5.24`) while the spinner and fee rows continue loading.

## Test plan
- [x] `./gradlew :app:testDebugUnitTest` (swap package) — new + existing tests green
- [x] `./gradlew :app:compileDebugKotlin`
- [x] Device render via PreviewActivity (`swap_quote_loading`) — before/after above
- New tests: indicative derivation + null paths (cold/zero price, zero amount, same token); percentage bypasses debounce; typing still debounced; indicative→firm transition; silent refresh keeps value without flashing loading

## Notes
- Adds a debug-only `SwapFormQuoteLoadingPreview` (in `SwapScreen.kt`, alongside the existing previews) wired into `PreviewActivity` for the screenshots above.

## Summary by CodeRabbit

* **New Features**
  * Shows an indicative destination amount (from cached spot prices) while fetching firm swap quotes.
  * Immediate quote fetch for percentage/Max selections and paste actions (bypasses typing debounce).

* **UI**
  * Indicative estimates are visually dimmed to distinguish them from firm quotes.
  * Spinner/skeleton is suppressed when an indicative value is available to reduce flashing.
  * Loading state refined to avoid flashing when an indicative value is shown.

* **Bug Fixes**
  * Empty amount no longer triggers a transient "Invalid amount" error.
  * Silent background quote refresh preserves the displayed destination value without flicker.

* **Tests**
  * Expanded coverage for debounce behavior, indicative estimates, silent refresh, and quote refresh timing.


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -916,30 +940,51 @@ constructor(
...
+                    QuoteInput(address = address, amount = srcAmount, immediate = immediate)
+                }
+                // Fires on real user intent (typing, paste, percentage, token change) but not on
+                // the
+                // silent refreshes combined in below — so the spinner appears immediately ahead of
+                // the debounce and an instant indicative estimate fills the destination field while
+                // we wait, without flashing on background refreshes (#4712).
+                .onEach { input ->
+                    isLoading = true
+                    showIndicativeRate(input)
```
rkokhatskyi, line 982: `showIndicativeRate` runs suspending I/O in this `onEach` which sits upstream of and outside the `collectLatest` try/catch, so a throw from `getCachedPrice` decoding its TEXT price via `BigDecimal` or a DataStore read escapes into `safeLaunch` and ends the whole quote collection for good while `isLoading` stays stuck true
```kotlin
private suspend fun showIndicativeRate(input: QuoteInput) {
    try {
        // current body
        throw e
    } catch (e: Exception) {
        Timber.e(e, "indicativeRate")
    }
}
```

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -735,6 +735,189 @@ internal class SwapFormViewModelTest {
...
+            Snapshot.sendApplyNotifications()
+            advanceTimeBy(500)
+            advanceUntilIdle()
+            assertEquals("95.0", vm.uiState.value.estimatedDstTokenValue)
+            assertFalse(vm.uiState.value.isLoading)
+
+            // Drive the expiry-triggered refresh (quote lives ~1 minute). The refresh is downstream
+            // of the loading/indicative side effects, so it must not flash the spinner or blank the
+            // previously shown destination value.
+            advanceTimeBy(61_000)
```
rkokhatskyi, line 925: `advanceTimeBy(61_000)` cannot drive the refresh since `launchRefreshQuoteTimer` delays on `Dispatchers.IO` against a real `Clock.System.now()` expiry, so `refreshQuoteState` never bumps and these assertions pass only because nothing ran, which leaves the silent refresh behaviour the PR is built around unverified
```kotlin
// make the timer dispatcher injectable and pass mainDispatcher in the test
viewModelScope.launch(ioDispatcher) { delay(expiredAt - Clock.System.now()); refreshQuoteState.value++ }
```

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
```diff
@@ -358,9 +358,15 @@ internal fun SwapScreen(
...
+                            // Keep the last (or indicative) destination value on screen while a new
+                            // quote loads; only fall back to the skeleton when there is nothing to
+                            // show yet — the first quote for a pair (#4712).
+                            val dstHasValue =
+                                state.estimatedDstTokenValue.isNotBlank() &&
+                                    state.estimatedDstTokenValue != "0"
                             TokenInput(
                                 title = stringResource(R.string.swap_form_dst_token_title),
-                                isLoading = state.isLoading,
+                                isLoading = state.isLoading && !dstHasValue,
```
rkokhatskyi, line 369: Suppressing the skeleton with `!dstHasValue` shows the previous pair's `estimatedDstTokenValue` in the bright firm colour while a freshly selected token loads, since a token select never resets the value and `isDstEstimated` stays false, so the old amount reads as a firm quote for the new pair until it resolves
```kotlin
// reset the dst estimate when the pair changes so a cold new pair shows the skeleton
uiState.update { it.copy(estimatedDstTokenValue = "0", isDstEstimated = false) }
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -1245,6 +1291,32 @@ constructor(
...
+
+        val currency = appCurrencyRepository.currency.first()
+        val indicative =
+            swapQuoteManager.computeIndicativeQuote(srcToken, dstToken, amount, currency) ?: return
+
+        uiState.update {
+            it.copy(
+                estimatedDstTokenValue = indicative.estimatedDstTokenValue,
+                estimatedDstFiatValue = indicative.estimatedDstFiatValue,
+                isDstEstimated = true,
```
rkokhatskyi: Writing this greyed estimate and then wiping it to `estimatedDstTokenValue = "0"` in `resetQuoteState` on any failed or unsupported quote makes such a pair flash a receivable amount before blanking, which inverts the never blank goal and is jumpier than the prior steady `"0"`

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Good catch — wrapped `showIndicativeRate` in try/catch (rethrow `CancellationException`, `Timber.e` the rest) so a throw from the cached-price read / DataStore decode can no longer escape into `safeLaunch` and end the collection with `isLoading` stuck true. Fixed in 11428c9ea.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Right, those assertions were vacuous. The timer dispatcher is now injectable (`@IoDispatcher`); the test drives it on the test scheduler with bounded `advanceTimeBy` + `coVerify(atLeast = 2)` so the silent refresh actually fires and re-fetches. One wrinkle: the refresh re-arms itself into a periodic timer, so I default the other tests' `ioDispatcher` to an inert dispatcher and cancel the VM at the end of this test — otherwise `runTest`'s cleanup `advanceUntilIdle()` spins it forever and OOMs the worker. Fixed in 11428c9ea.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
aminsato: Fixed at the source rather than in the screen: the VM now resets `estimatedDstTokenValue`/`isDstEstimated` on pair change (an `onEach` after `distinctUntilChanged`), so a freshly selected pair shows the skeleton instead of the previous pair's value reading as a firm quote. 11428c9ea.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Agreed. `showIndicativeRate` now bails when `getEligibleProviders(src, dst).isEmpty()` (a local table lookup, so still instant), so an unsupported pair never writes the greyed estimate that `resetQuoteState` would then blank — no flash, steady `"0"`. 11428c9ea.

johnnyluo:
CleanShot 2026-06-03 at 19 52 24@2x

@aminsato , as soon as I get into swap screen , it show an error "invalid amount" , can you please make sure this is fixed?

aminsato:
@johnnyluo Fixed in 11245fda3.

The empty field on entry was being routed into the zero-amount branch and surfaced "Invalid amount" before any input. An empty field is now treated as a clean state (no error); an explicitly entered `0` still shows the error. Added a regression test covering the empty-on-entry case.

johnnyluo (approved):
LGTM

PR #4722, opened 2026-06-03
fix(wallet): show Buy VULT banner until vault reaches Silver tier

johnnyluo:
## Summary
- The Buy VULT banner dismissal was a single app-level boolean (`buy_vult_banner_dismissed`) with no tier gate, so closing it hid the banner **forever** and **across every vault**.
- Port of iOS [#4475](https://github.com/vultisig/vultisig-ios/pull/4475): the dismissal is now **per-vault** and only persisted once the vault holds at least the Silver-tier VULT amount (≥ 3000 VULT).
- Below Silver, closing hides the banner for the **current visit only** — it returns on the next entry to the home screen or on a vault switch.

## Behavior
- **< 3000 VULT:** close hides it for the visit; returns on next entry / vault switch. Nothing persisted.
- **≥ 3000 VULT (Silver+):** dismissal persists per-vault and sticks.
- Existing users who previously dismissed it will see it again, since the storage key changed from a global key to a per-vault key (same migration as iOS).

## Changes
- `GetDiscountBpsUseCase`: expose `hasReachedSilverTier(vaultId)` on the interface, reusing the existing `getVultBalance()` + `SILVER_TIER_THRESHOLD`.
- `VaultDataStoreRepository`: make the dismissed flag per-vault (`buy_vult_banner_dismissed/<vaultId>`), following the existing `readBackupStatus(vaultId)` pattern.
- `VaultAccountsViewModel`: add an in-memory per-visit dismiss flag; start a per-vault banner collector from `loadData()` (reset the session flag and restart on vault switch); persist the dismissal only when `hasReachedSilverTier` is true.

No UI-layer changes — the screen already keys off `state.showBuyVultBanner` / `onBuyVultDismiss`.

## Test Plan
- [x] Unit tests updated/added in `VaultAccountsViewModelTest` (persist at Silver; hide-without-persist below Silver; show/hide from persisted flag) — pass
- [ ] Manual: vault < 3000 VULT → close banner → gone for visit → returns after vault switch; ≥ 3000 VULT → close → stays gone

## Summary by CodeRabbit

* **New Features**
  * Banner dismissal is now vault-scoped so each vault has its own state.
  * Tier-aware dismissal: dismissal is persisted only when a vault reaches Silver tier.

* **Bug Fixes**
  * "Buy VULT" banner hides for the visit and resets for non-Silver vaults (no persistent save).

* **Chores**
  * Migrated legacy global dismissal data to the new per-vault model.

* **Tests**
  * Added/updated tests covering vault-scoped and tier-dependent dismissal behavior.


app/src/main/java/com/vultisig/wallet/data/usecases/GetDiscountBpsUseCase.kt
```diff
@@ -54,6 +57,11 @@ constructor(
         }
     }
 
+    override suspend fun hasReachedSilverTier(vaultId: String): Boolean {
+        val balance = getVultBalance(vaultId) ?: return false
```
aminsato, line 61: `getVultBalance` reads from `getCachedTokenBalances`, so this returns `false` whenever the cache isn't warm (a cold start yields `ZERO`). A genuine Silver-tier user who dismisses the banner before balances finish loading won't get the dismissal persisted, and it returns on next visit. The persist guarantee depends on cache state at dismiss time rather than the actual tier — worth confirming that's the intended behavior.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/VaultDataStoreRepository.kt
```diff
@@ -80,7 +80,7 @@ constructor(private val appDataStore: AppDataStore) : VaultDataStoreRepository {
         fun onGlobalBackupReminderStatusKey() =
             intPreferencesKey(name = "global_backup_reminder_status")
 
-        fun onBuyVultBannerDismissedKey() =
-            booleanPreferencesKey(name = "buy_vult_banner_dismissed")
+        fun onBuyVultBannerDismissedKey(vaultId: String) =
+            booleanPreferencesKey(name = "buy_vult_banner_dismissed/$vaultId")
```
aminsato, line 84: The previous global key `buy_vult_banner_dismissed` is now never read or written, so existing users' prior dismissals are stranded in DataStore. If you want to avoid leftover keys, remove the old one during migration.

app/src/main/java/com/vultisig/wallet/data/usecases/GetDiscountBpsUseCase.kt
johnnyluo: Good catch — this cached read is intentional. It mirrors iOS #4475, which reads the cached `getVultToken(for: vault)?.balanceDecimal` at dismiss time rather than gating on a fresh fetch. By the time a user taps close, the home screen's balance load has typically already populated the cache, so a genuine Silver-tier user persists as expected. The edge case (dismissing during a cold start before VULT resolves) is self-healing: nothing is persisted, the banner returns on the next visit, and by then the balance is warm so the next dismiss sticks. We chose this over a network call on every dismiss to keep the close action instant. If it proves annoying we can add a fresh-balance fallback (reusing `balanceRepository.getTokenValue` like `DiscountTiersViewModel`), but keeping parity with iOS for now.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/VaultDataStoreRepository.kt
johnnyluo: Done — added a one-time DataStore migration (`removeLegacyBuyVultBannerKey` in `MainDataModule`) that drops the orphaned global `buy_vult_banner_dismissed` key on next launch. `shouldMigrate` is gated on `contains(legacyKey)` so it's a no-op once removed.

aminsato (approved):
LGTM

PR #4742, opened 2026-06-05
feat(defi): disable new Circle USDC deposits, keep withdrawals open

### LINKED ISSUE #4739: Disable new Circle (USDC yield) deposits — keep withdrawals open
## What
Business decision: stop accepting **new deposits** into the Circle USDC yield product. Withdrawals and existing positions must keep working so users can always exit. This is not a Circle outage or a regulatory action — there is no external trigger.

## Scope
- Disable / hide the **Deposit** entry point for Circle on the positions screen.
- Gate the deposit submit path (defense-in-depth, not just hiding the button).
- **Keep working:** withdraw flow, existing-position display, balance read.

## Pointers
- `app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsScreen.kt`
- `app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt`
- `app/src/main/java/com/vultisig/wallet/ui/models/send/submit/SendStrategyFactory.kt` (deposit submit routing)
- Leave `app/src/main/java/com/vultisig/wallet/ui/models/send/submit/WithdrawUsdcCircleStrategy.kt` untouched.

## Acceptance criteria
- [ ] No path to initiate a new Circle deposit from the UI.
- [ ] Withdraw + existing positions remain fully functional.
- [ ] Users with an open Circle account can still read balance and withdraw.

johnnyluo:
## Summary
Business decision (#4739): stop accepting **new deposits** into the Circle USDC yield product while keeping **withdrawals and existing positions** fully functional so users can always exit.

- **Detection**: new `HasCircleAccountUseCase` resolves whether a vault already has a Circle MSCA account — cache-first, then on-chain discovery via `circleApi.getScAccount` (caches the result). Discovery keeps withdraw access working even on a fresh install where the local cache is empty.
- **Hide Circle for vaults without an account**: the Circle (Ethereum DeFi) row is filtered out of the DeFi portfolio (`VaultAccountsViewModel`) and the DeFi chain-selection screen (`DeFiChainSelectionViewModel`).
- **Disable Deposit for vaults with an account**: the Deposit button on the Circle positions screen is greyed out / non-clickable; **Withdraw stays enabled**.
- **Defense-in-depth**: the deposit submit path (`DEPOSIT_USDC_CIRCLE`) is gated to a no-op in `SendStrategyFactory`, so a deposit can't be initiated even if the UI is bypassed. `WithdrawUsdcCircleStrategy` is left untouched.

Note: the "Circle" row is the Ethereum DeFi chain rebranded (`Chain.Ethereum.toDefi` -> `DefiChain(raw = "Circle")`), and Ethereum's only DeFi product is Circle — so the gating keys off `Chain.Ethereum` in DeFi mode.

## Issue
Closes #4739

## Test Plan
- [x] Unit test `HasCircleAccountUseCaseTest` (cache hit / network hit+cache / no account / network throws / vault missing) — passes
- [x] `:app:testDebugUnitTest --tests VaultAccountsViewModelTest` — passes
- [x] Manual: vault **without** Circle account — Circle row absent from portfolio and chain-selection
- [ ] Manual: vault **with** Circle account + balance — Withdraw enabled, Deposit greyed/disabled
- [x] Manual: vault **with** Circle account + zero balance — single Deposit button greyed/disabled
- [ ] Manual: reaching Send with `DEPOSIT_USDC_CIRCLE` does not submit

## Summary by CodeRabbit

* **New Features**
  * Ethereum Circle (USDC yield) option is shown only for vaults with an existing Circle account.
  * CTA buttons now reflect account state (disabled when new Circle deposits are not available).

* **Bug Fixes**
  * Circle deposit submit path disabled to prevent initiating unavailable deposit flows.

* **Tests**
  * Added unit tests for Circle account presence, gating behavior, and failure handling.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/DeFiChainSelectionViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/DeFiChainSelectionViewModel.kt
@@ -12,6 +12,7 @@ import com.vultisig.wallet.data.models.isDeFiSupported
 import com.vultisig.wallet.data.repositories.DefaultDeFiChainsRepository
 import com.vultisig.wallet.data.repositories.VaultRepository
 import com.vultisig.wallet.data.usecases.HasCircleAccountUseCase
+import com.vultisig.wallet.data.utils.safeLaunch
 import com.vultisig.wallet.ui.models.mappers.ChainToDefiChainUiMapper
 import com.vultisig.wallet.ui.navigation.Destination
 import com.vultisig.wallet.ui.navigation.Navigator
@@ -64,14 +65,14 @@ constructor(
     }
 
     private fun loadChains() {
-        viewModelScope.launch {
+        viewModelScope.safeLaunch {
             val vault = vaultRepository.get(vaultId)
 
             _uiState.update { it.copy(isLoading = true) }
 
             if (vault == null) {
                 _uiState.update { it.copy(defiChains = emptyList(), isLoading = false) }
-                return@launch
+                return@safeLaunch
             }
 
             // New Circle (USDC yield) deposits are disabled: only offer the Circle (Ethereum
--- a/app/src/main/java/com/vultisig/wallet/ui/models/DeFiChainSelectionViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/DeFiChainSelectionViewModel.kt
@@ -6,10 +6,12 @@ import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import androidx.navigation.toRoute
+import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.DefiChainUiModel
 import com.vultisig.wallet.data.models.isDeFiSupported
 import com.vultisig.wallet.data.repositories.DefaultDeFiChainsRepository
 import com.vultisig.wallet.data.repositories.VaultRepository
+import com.vultisig.wallet.data.usecases.HasCircleAccountUseCase
 import com.vultisig.wallet.ui.models.mappers.ChainToDefiChainUiMapper
 import com.vultisig.wallet.ui.navigation.Destination
 import com.vultisig.wallet.ui.navigation.Navigator
@@ -42,6 +44,7 @@ constructor(
     savedStateHandle: SavedStateHandle,
     private val vaultRepository: VaultRepository,
     private val defaultDeFiChainsRepository: DefaultDeFiChainsRepository,
+    private val hasCircleAccount: HasCircleAccountUseCase,
     private val mapChainDefi: ChainToDefiChainUiMapper,
     private val navigator: Navigator<Destination>,
 ) : ViewModel() {
@@ -71,8 +74,14 @@ constructor(
                 return@launch
             }
 
+            // New Circle (USDC yield) deposits are disabled: only offer the Circle (Ethereum
+            // DeFi) entry to vaults that already opened an account, so they can still withdraw.
+            val hasCircle = hasCircleAccount(vaultId)
             val availableChains =
-                vault.coins.map { it.chain }.distinct().filter { it.isDeFiSupported }
+                vault.coins
+                    .map { it.chain }
+                    .distinct()
+                    .filter { it.isDeFiSupported && (it != Chain.Ethereum || hasCircle) }
 
             val savedDeFiChains = defaultDeFiChainsRepository.getDefaultChains(vaultId).first()
 
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -35,6 +35,7 @@ import com.vultisig.wallet.data.repositories.vault.VaultMetadataRepo
 import com.vultisig.wallet.data.services.PushNotificationManager
 import com.vultisig.wallet.data.usecases.EnableTokenUseCase
 import com.vultisig.wallet.data.usecases.GetDiscountBpsUseCase
+import com.vultisig.wallet.data.usecases.HasCircleAccountUseCase
 import com.vultisig.wallet.data.usecases.IsGlobalBackupReminderRequiredUseCase
 import com.vultisig.wallet.data.usecases.NeverShowGlobalBackupReminderUseCase
 import com.vultisig.wallet.data.utils.safeLaunch
@@ -62,6 +63,8 @@ import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
@@ -141,6 +144,7 @@ constructor(
     private val getDiscountBpsUseCase: GetDiscountBpsUseCase,
     private val cryptoConnectionTypeRepository: CryptoConnectionTypeRepository,
     private val defaultDeFiChainsRepository: DefaultDeFiChainsRepository,
+    private val hasCircleAccount: HasCircleAccountUseCase,
     private val tiersNFTRepository: TiersNFTRepository,
     private val remoteNFTService: TierRemoteNFTService,
     private val pushNotificationManager: PushNotificationManager,
@@ -491,7 +495,8 @@ constructor(
                             },
                         uiState.value.searchTextFieldState.textAsFlow(),
                         defaultDeFiChainsRepository.getDefaultChains(vaultId),
-                    ) { accounts, searchQuery, selectedDeFiChains ->
+                        flow { emit(hasCircleAccount(vaultId)) }.flowOn(ioDispatcher),
+                    ) { accounts, searchQuery, selectedDeFiChains, hasCircleAccount ->
                         Timber.d(
                             "DeFi Accounts Loaded for vault $vaultId: ${accounts.size} accounts, selected chains: ${selectedDeFiChains.map { it.raw }}"
                         )
@@ -499,8 +504,13 @@ constructor(
                         val filteredAccounts =
                             accounts.filter { address ->
                                 val isSelected = selectedDeFiChains.contains(address.chain)
+                                // New Circle (USDC yield) deposits are disabled: only surface the
+                                // Circle (Ethereum DeFi) row for vaults that already opened an
+                                // account, so they can still withdraw.
+                                val isCircleAllowed =
+                                    address.chain != Chain.Ethereum || hasCircleAccount
                                 Timber.d("Chain ${address.chain.raw} is selected: $isSelected")
-                                isSelected
+                                isSelected && isCircleAllowed
                             }
 
                         Timber.d("Filtered DeFi accounts: ${filteredAccounts.size} accounts")
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/submit/SendStrategyFactory.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/submit/SendStrategyFactory.kt
@@ -69,8 +69,12 @@ internal data class SendStrategies(
 
             DeFiNavActions.WITHDRAW_USDC_CIRCLE -> withdrawUsdcCircle.submit()
 
+            // New Circle (USDC yield) deposits are disabled. The UI no longer exposes a deposit
+            // entry point; this no-op gates the submit path as defense-in-depth so a deposit can't
+            // be initiated even if that path is somehow reached.
+            DeFiNavActions.DEPOSIT_USDC_CIRCLE -> Unit
+
             null,
-            DeFiNavActions.DEPOSIT_USDC_CIRCLE,
             DeFiNavActions.STAKE_CACAO,
             DeFiNavActions.UNSTAKE_CACAO,
             DeFiNavActions.ADD_LP,
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsScreen.kt
@@ -10,6 +10,7 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.hilt.navigation.compose.hiltViewModel
 import com.vultisig.wallet.R
 import com.vultisig.wallet.data.models.VaultId
+import com.vultisig.wallet.ui.components.buttons.VsButtonState
 import com.vultisig.wallet.ui.screens.v2.defi.BaseDeFiPositionsScreenContent
 import com.vultisig.wallet.ui.screens.v2.defi.DeFiTab
 import com.vultisig.wallet.ui.screens.v2.defi.DeFiWarningBanner
@@ -118,6 +119,8 @@ private fun CircleContentDepositTab(
             totalPrice = state.totalDepositCurrency,
             isLoading = state.isLoading,
             isBalanceVisible = isBalanceVisible,
+            // New Circle deposits are disabled; existing positions can still be withdrawn.
+            isSecondActionEnabled = false,
         )
     } else {
         HeaderDeFiWidget(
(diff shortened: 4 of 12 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/usecases/HasCircleAccountUseCase.kt
```diff
@@ -0,0 +1,58 @@
...
+import timber.log.Timber
+
+/**
+ * Resolves whether a vault already has a Circle (USDC yield) MSCA account.
+ *
+ * Checks the local cache first; if empty, discovers the account on-chain via [CircleApi] and caches
+ * the result. Discovery keeps withdraw access working for existing accounts even on a fresh install
+ * where the local cache is empty. New Circle deposits are disabled, so this is used to hide the
+ * Circle entry for vaults that never opened an account.
+ */
```
rkokhatskyi, line 15: Issue #4739 only asks to disable the deposit entry and gate the submit path, both already handled by `CircleDeFiPositionsScreen` and `SendStrategyFactory`, so the on chain discovery added here purely to hide the Circle row exceeds the issue and is the source of the network cost and the hide an existing account edge cases below, whereas iOS gates visibility on a local flag and Windows on a `circle` feature flag with no network call, so dropping the row hiding or moving it behind a flag would satisfy the issue with far less risk

data/src/main/kotlin/com/vultisig/wallet/data/usecases/HasCircleAccountUseCase.kt
```diff
@@ -0,0 +1,58 @@
...
+
+internal class HasCircleAccountUseCaseImpl
+@Inject
+constructor(
+    private val scaCircleAccountRepository: ScaCircleAccountRepository,
+    private val circleApi: CircleApi,
+    private val vaultRepository: VaultRepository,
+    private val chainAccountAddressRepository: ChainAccountAddressRepository,
+) : HasCircleAccountUseCase {
+
```
rkokhatskyi, line 22: This cache then discover then `saveAccount` flow duplicates `CircleDeFiPositionsViewModel.fetchAssociatedMscaAccount`, so consider extracting a single use case that returns the MSCA address and deriving the boolean from it, which would also let `CircleDeFiPositionsViewModel` drop its own `circleApi` discovery and cache writes

data/src/main/kotlin/com/vultisig/wallet/data/usecases/HasCircleAccountUseCase.kt
```diff
@@ -0,0 +1,58 @@
...
+    override suspend fun invoke(vaultId: String): Boolean {
```
rkokhatskyi: This `getAccount` read sits outside the `try`, so a DataStore failure throws here uncaught, and since `DeFiChainSelectionViewModel.loadChains` invokes this from a plain `viewModelScope.launch` instead of `safeLaunch` that would crash the screen, so move the cache read inside the `try`

data/src/main/kotlin/com/vultisig/wallet/data/usecases/HasCircleAccountUseCase.kt
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        return try {
+            val vault = vaultRepository.get(vaultId) ?: return false
+            val (evmAddress, _) = chainAccountAddressRepository.getAddress(Chain.Ethereum, vault)
+            val mscaAddress = circleApi.getScAccount(evmAddress)
+            if (mscaAddress != null) {
+                scaCircleAccountRepository.saveAccount(vaultId, mscaAddress)
+                true
+            } else {
```
rkokhatskyi: `saveAccount` runs only when `mscaAddress` is not null, so a vault with no account is never cached and every open of the DeFi tab and the chain selection screen re runs `getScAccount` over the network, which is now the common case for nearly every vault since deposits are disabled, so cache a negative result with a TTL or memoize per vault for the session

data/src/main/kotlin/com/vultisig/wallet/data/usecases/HasCircleAccountUseCase.kt
```diff
@@ -0,0 +1,58 @@
...
+            }
+        } catch (e: CancellationException) {
+            throw e
+        } catch (t: Throwable) {
+            // A network blip must not wrongly hide an existing account; fall back to the cache.
+            Timber.e(
+                t,
+                "HasCircleAccountUseCase: failed to discover Circle account for %s",
+                vaultId,
+            )
```
rkokhatskyi: This `catch` returns `false` for any throwable, so when the cache is empty after a reinstall and `getScAccount` fails transiently an existing account is hidden and the user loses the only path to withdraw, which is the opposite of what the comment states and breaks the always exit goal, so distinguish a definitive empty response from a thrown error and keep the Circle row visible when discovery cannot be determined

app/src/main/java/com/vultisig/wallet/ui/models/DeFiChainSelectionViewModel.kt
```diff
@@ -71,8 +74,14 @@ constructor(
                 return@launch
             }
 
+            // New Circle (USDC yield) deposits are disabled: only offer the Circle (Ethereum
+            // DeFi) entry to vaults that already opened an account, so they can still withdraw.
+            val hasCircle = hasCircleAccount(vaultId)
```
rkokhatskyi, line 80: This calls `hasCircleAccount` on `viewModelScope` which is Main, and the use case runs `ChainAccountAddressRepository.getAddress` doing synchronous Trust Wallet derivation plus the network continuation on that thread, whereas `VaultAccountsViewModel` wraps the same call in `flowOn(ioDispatcher)`, so wrap this in `withContext(ioDispatcher)` and switch `loadChains` to `safeLaunch`

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -491,16 +495,22 @@ constructor(
                             },
                         uiState.value.searchTextFieldState.textAsFlow(),
                         defaultDeFiChainsRepository.getDefaultChains(vaultId),
-                    ) { accounts, searchQuery, selectedDeFiChains ->
+                        flow { emit(hasCircleAccount(vaultId)) }.flowOn(ioDispatcher),
```
rkokhatskyi, line 498: `combine` waits for every source to emit, so this discovery flow makes the entire DeFi accounts list wait for a network call before its first render, and it fires for every vault including those with no Ethereum DeFi account since the call runs before the Ethereum filter, so gate it behind an Ethereum present check and seed it from the cache so the list is not blocked on the network

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -491,16 +495,22 @@ constructor(
...
                         )
 
                         val filteredAccounts =
                             accounts.filter { address ->
                                 val isSelected = selectedDeFiChains.contains(address.chain)
+                                // New Circle (USDC yield) deposits are disabled: only surface the
+                                // Circle (Ethereum DeFi) row for vaults that already opened an
+                                // account, so they can still withdraw.
+                                val isCircleAllowed =
+                                    address.chain != Chain.Ethereum || hasCircleAccount
```
rkokhatskyi, line 511: Keying on `address.chain != Chain.Ethereum` uses the whole Ethereum chain as a proxy for the Circle product, which only holds while Circle is Ethereum's sole DeFi product, so a second Ethereum DeFi product would also be hidden whenever the vault has no Circle account, so key the gate on the Circle product or position identity rather than the chain

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/SendStrategyFactory.kt
```diff
@@ -69,8 +69,12 @@ internal data class SendStrategies(
 
             DeFiNavActions.WITHDRAW_USDC_CIRCLE -> withdrawUsdcCircle.submit()
 
+            // New Circle (USDC yield) deposits are disabled. The UI no longer exposes a deposit
+            // entry point; this no-op gates the submit path as defense-in-depth so a deposit can't
+            // be initiated even if that path is somehow reached.
+            DeFiNavActions.DEPOSIT_USDC_CIRCLE -> Unit
```
rkokhatskyi, line 75: Mapping this to `Unit` silently does nothing, so if the path is reached the user taps continue and gets no error or navigation even though `showError` is available in the context, and the parallel gap is that `CircleDeFiPositionsViewModel.onCreateAccount` calling `circleApi.createScAccount` opens a new account with no equivalent guard and is blocked only by the disabled button, so surface feedback here and consider gating account creation too

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsScreen.kt
```diff
@@ -134,6 +137,8 @@ private fun CircleContentDepositTab(
             totalPrice = state.totalDepositCurrency,
             isLoading = state.isLoading,
             isBalanceVisible = isBalanceVisible,
+            // New Circle deposits are disabled; this single CTA is deposit/open-account only.
+            buttonState = VsButtonState.Disabled,
```
rkokhatskyi, line 138: The withdraw button only renders in the `hasActiveDeposit()` branch which parses the `totalDeposit` string, so an open account whose balance read fails or is unparseable falls here into this now disabled single CTA with no withdraw shown, leaving a funded user no way to exit where that button was previously enabled, so drive the action off `isAccountOpen` rather than the parsed balance

app/src/test/java/com/vultisig/wallet/ui/models/home/VaultAccountsViewModelTest.kt
```diff
@@ -117,6 +119,7 @@ internal class VaultAccountsViewModelTest {
         getDiscountBpsUseCase = mockk(relaxed = true)
         cryptoConnectionTypeRepository = mockk(relaxed = true)
         defaultDeFiChainsRepository = mockk(relaxed = true)
+        hasCircleAccount = mockk(relaxed = true)
```
rkokhatskyi, line 122: This relaxed mock returns false by default and no test drives `loadDeFiBalances` or asserts the Ethereum row is filtered, so the main row hiding behavior is unverified, and `DeFiChainSelectionViewModel` has no test at all for its identical filter, while `HasCircleAccountUseCaseTest` asserts false when discovery throws which locks in the hide an existing account behavior, so add filter tests for both true and false and revisit the throws case

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/usecases/HasCircleAccountUseCase.kt
johnnyluo: Fixed — the cache read is now inside the `try`, so a DataStore failure returns `false` instead of throwing. Also switched `DeFiChainSelectionViewModel.loadChains` to `safeLaunch`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/HasCircleAccountUseCase.kt
johnnyluo: Good point. The per-vault show/hide of the Circle row is an explicit requirement for this change (hide for vaults without an account, keep + disable-deposit for vaults with one), which a single global flag cannot express. We have removed the network call though — detection is now **cache-only**, reading the locally persisted MSCA address via `ScaCircleAccountRepository`, so there is no network cost. The trade-off (a reinstall/restore with an on-chain-only account is treated as "no account") was an accepted decision for this change.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/HasCircleAccountUseCase.kt
johnnyluo: Resolved by the cache-only change: the use case no longer does `circleApi` discovery or `saveAccount`, so it no longer duplicates `CircleDeFiPositionsViewModel.fetchAssociatedMscaAccount`. The screen remains the single place that performs network discovery and populates the cache.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/HasCircleAccountUseCase.kt
johnnyluo: Moot now: detection is cache-only, so every call is a cheap local DataStore read — there is no repeated network `getScAccount`, so no negative-cache/TTL needed.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/HasCircleAccountUseCase.kt
johnnyluo: The transient-network-failure case is gone with cache-only detection (no `getScAccount` to throw). The remaining `catch` only covers a DataStore read failure, where `false` is consistent with the chosen "empty cache => treat as no account" semantics. The inherent reinstall/restore trade-off was an accepted decision for this change.

app/src/main/java/com/vultisig/wallet/ui/models/DeFiChainSelectionViewModel.kt
johnnyluo: `loadChains` now uses `safeLaunch`. The use case is also no longer doing Trust Wallet derivation or a network call — it is a single suspend DataStore read (already off the main thread inside DataStore), so the main-thread concern is resolved.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
johnnyluo: The blocking concern is resolved by cache-only detection: the added `combine` source is now a fast local DataStore read rather than a network call, so it no longer holds back first render in any meaningful way.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
johnnyluo: Acknowledged. Today `Chain.Ethereum` has exactly one DeFi product (Circle) — the same chain-as-Circle mapping the rest of the DeFi layer already relies on (`Chain.toDefi` -> `DefiChain(raw = "Circle")`). If a second Ethereum DeFi product is ever added, this gate (and that mapping) will need to key on product/position identity. Leaving as-is for now since it matches the existing model.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/SendStrategyFactory.kt
johnnyluo: The submit `Unit` branch is unreachable defense-in-depth: the deposit entry point is removed from the UI, so the Send screen is never opened with `DEPOSIT_USDC_CIRCLE`. `SendStrategies` does not hold `showError`, and wiring it in just to message an unreachable path felt like more surface than it is worth — kept as a no-op gate. On account creation: with cache-only hiding, no-account vaults do not see the Circle row at all, so `onCreateAccount` is unreachable from the UI; the open-account CTA is additionally rendered disabled (`VsButtonState.Disabled`).

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsScreen.kt
johnnyluo: Fixed. The widget branch is now driven by `isAccountOpen` instead of the parsed-balance `hasActiveDeposit()`, so any open account renders the enabled Withdraw button (with Deposit disabled) regardless of whether the balance read succeeded or parsed. Removed the now-unused `hasActiveDeposit()`.

app/src/test/java/com/vultisig/wallet/ui/models/home/VaultAccountsViewModelTest.kt
johnnyluo: Added `DeFiChainSelectionViewModelTest` covering both true/false (Circle entry shown when a Circle account exists, hidden when not). Rewrote `HasCircleAccountUseCaseTest` for the cache-only behaviour (cached => true, absent => false, read-failure => false) — the old "network throws => false" case is gone with the network call. The portfolio filter in `VaultAccountsViewModel` applies the identical predicate (`it != Chain.Ethereum || hasCircle`); I opted not to add a full `loadDeFiBalances` integration test as it requires driving the whole vault-load chain for low marginal coverage over the shared, now-tested predicate. Happy to add it if you prefer.

rkokhatskyi (approved):
LGTM

PR #4743, opened 2026-06-05
fix(zcash): update consensus branch id to 0x30f33754

johnnyluo:
## Summary
- Updates the Zcash ZIP-243 consensus branch id from `f04dec4d` to `30f33754` so the signature digest matches the active network consensus rules.
- Applies to both the native ZEC send plan (`UtxoHelper`) and the SwapKit ZEC signer (`SwapKitZcashSigner`).
- Updates the doc comments and the frozen-plan test assertion accordingly.

## Why
The branch id feeds WalletCore's ZIP-243 personalised-Blake2b sighash. It must equal the current consensus branch id or the network rejects the transaction.

## Test plan
- [ ] `./gradlew :data:test` — `SwapKitZcashSignerTest` passes with the new branch id
- [ ] Manual ZEC send signs and broadcasts successfully
- [ ] ZEC swap via SwapKit/NEAR Intents signs and broadcasts successfully

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Zcash transaction signing to use the correct ZIP-243 branch identifier, ensuring proper Zcash transaction compatibility.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
@@ -26,7 +26,7 @@ internal class SwapKitZcashSignerException(message: String) : Exception(message)
  *
  * Sighash: WalletCore's ZEC signer implements ZIP-243 (the Sapling signature digest with the
  * personalised Blake2b-256). It reads the branch id from the plan — the existing native ZEC send
- * uses [ZCASH_BRANCH_ID_HEX] `f04dec4d`, and we match that so the digest is identical to a manually
+ * uses [ZCASH_BRANCH_ID_HEX] `30f33754`, and we match that so the digest is identical to a manually
  * sent ZEC transaction. (Deviating to the Sapling-v4 spec id `0x76b809bb` would produce a different
  * digest the chain rejects.)
  *
@@ -445,6 +445,6 @@ internal class SwapKitZcashSigner(
          * personalised-Blake2b branch identifier; diverging to the Sapling-v4-spec `0x76b809bb`
          * would produce a digest the network rejects.
          */
-        private const val ZCASH_BRANCH_ID_HEX = "f04dec4d"
+        private const val ZCASH_BRANCH_ID_HEX = "30f33754"
     }
 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
@@ -443,8 +443,9 @@ internal class SwapKitZcashSigner(
         /**
          * Branch id matching the existing native ZEC send. WalletCore reads it as the ZIP-243
          * personalised-Blake2b branch identifier; diverging to the Sapling-v4-spec `0x76b809bb`
-         * would produce a digest the network rejects.
+         * would produce a digest the network rejects. Sourced from [ZCASH_ZIP243_BRANCH_ID_HEX] so
+         * both signing paths stay locked together.
          */
-        private const val ZCASH_BRANCH_ID_HEX = "30f33754"
+        private const val ZCASH_BRANCH_ID_HEX = ZCASH_ZIP243_BRANCH_ID_HEX
     }
 }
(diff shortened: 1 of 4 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelper.kt
```diff
@@ -218,7 +218,7 @@ class UtxoHelper(
 
         val plan =
             if (coinType == CoinType.ZCASH) {
-                initialPlan.toBuilder().setBranchId(ByteString.fromHex("f04dec4d")).build()
+                initialPlan.toBuilder().setBranchId(ByteString.fromHex("30f33754")).build()
```
rkokhatskyi: `UtxoHelper` hardcodes its own `30f33754` literal rather than referencing `ZCASH_BRANCH_ID_HEX` so a future upgrade can bump one path while the other keeps a stale branch id
```kotlin
setBranchId(ByteString.fromHex(ZCASH_BRANCH_ID_HEX))
```

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
```diff
@@ -445,6 +445,6 @@ internal class SwapKitZcashSigner(
          * personalised-Blake2b branch identifier; diverging to the Sapling-v4-spec `0x76b809bb`
          * would produce a digest the network rejects.
          */
-        private const val ZCASH_BRANCH_ID_HEX = "f04dec4d"
+        private const val ZCASH_BRANCH_ID_HEX = "30f33754"
```
rkokhatskyi: `ZCASH_BRANCH_ID_HEX` lands as an unexplained literal so the next upgrade has no recorded rule for regenerating it whereas it is `getblockchaininfo` `consensus.nextblock` `5437f330` with its four bytes reversed
```kotlin
// NU6.2 consensus.nextblock 5437f330 reversed to little endian, rerun per network upgrade
private const val ZCASH_BRANCH_ID_HEX = "30f33754"
```

data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSignerTest.kt
```diff
@@ -219,7 +219,7 @@ class SwapKitZcashSignerTest {
             assertEquals(39_000L, input.plan.change)
             assertEquals(1_000L, input.plan.fee)
             // ZIP-243 branch id must be on the plan or WalletCore derives the wrong digest.
-            assertEquals("f04dec4d", Numeric.toHexStringNoPrefix(input.plan.branchId.toByteArray()))
+            assertEquals("30f33754", Numeric.toHexStringNoPrefix(input.plan.branchId.toByteArray()))
```
rkokhatskyi, line 225: the `branchId` assertion mirrors the production literal so it confirms only that an id reached `input.plan` whereas a wrong value would be edited here in lockstep and still pass

rkokhatskyi (changes requested):

akhilesh-x:
can we get the apk for this PR?

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelper.kt
johnnyluo: Done — extracted a shared `ZCASH_ZIP243_BRANCH_ID_HEX` constant (new `ZcashConstants.kt`). `UtxoHelper` and `SwapKitZcashSigner` both reference it now, so the two digest paths can't drift.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
johnnyluo: Done — moved the value to the shared `ZCASH_ZIP243_BRANCH_ID_HEX` constant with a doc comment recording the regeneration rule: `getblockchaininfo` → `consensus.nextblock` (NU6.2 `5437f330`) reversed to little endian → `30f33754`. Thanks for confirming the byte order.

data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSignerTest.kt
johnnyluo: The literal here is intentional and kept separate from the production constant (not referencing it) precisely so it acts as a golden-value change-detector: if the branch id is ever changed, this test fails and forces a deliberate update. Added a comment to make that intent explicit. The assertion's value is verifying that `UtxoHelper` actually applies the branch-id override to the plan (WalletCore's default plan has none).

rkokhatskyi (approved):
LGTM

PR #4744, opened 2026-06-05
fix(zcash): enable ZIP-317 fee so multi-UTXO sends are accepted

johnnyluo:
## Summary
- Zcash sends that spend **3+ UTXOs** were rejected by the node with `66: tx unpaid action limit exceeds limit of 0` (ZIP-317).
- WalletCore ships a `Zip0317FeeCalculator` (`fee = 5000 × max(2 grace actions, max(inputs, outputs))`) but only uses it when `zip_0317` is set on the Bitcoin signing input — which the app never did.
- Without the flag, the Zcash fee did not scale with input count. A 1-input send is covered by the 2 grace actions, but as soon as a tx has 3+ logical actions it paid **below** the conventional fee → unpaid actions → rejected (the broadcast node uses `-txunpaidactionlimit=0`).
- Fix: set `zip_0317 = true` for Zcash in both signing-input builders (`getSigningInputData` + `getBitcoinSigningInput`) so WalletCore computes the correct ZIP-317 fee. The Verify screen reads the same plan, so the displayed fee stays consistent.

https://blockchair.com/zcash/transaction/014c076839cf3f86f4ae695f5afb2c763f2019e7933193df82e86a9c3db71ee1

## Root cause detail
`getFeeCalculator(TWCoinTypeZcash, …, zip0317)` in WalletCore returns `Zip0317FeeCalculator` only when `zip0317 == true`; otherwise it falls back to a calculator whose fee does not track logical actions. The `zip_0317` proto field (`Bitcoin.SigningInput` #18) exists in the bundled `wallet-core` 4.3.22 but was never set.

## Test Plan
- [x] Added regression test `getBitcoinTransactionPlan - Zcash applies ZIP-317 fee that scales with logical actions` (1 input → fee 10 000; 3 inputs → fee 15 000). JNI-gated like the existing plan tests.
- [x] `./gradlew :data:lint` passes
- [x] `./gradlew :data:compileDebugKotlin` + `:data:testDebugUnitTest` pass
- [ ] **Manual on-device verification recommended:** perform a real ZEC send spending 3+ UTXOs and confirm it broadcasts (the fee assertions can't execute in CI without the WalletCore native lib — the test skips gracefully there)

## Notes
- Shared `data`-layer fix; covers the native **send** path. Zcash **swaps** go through `SwapKitZcashSigner` (its own PSBT plan with a SwapKit-provided fee) and are unaffected.
- iOS has the same latent gap (`UTXOChainsHelper.swift` never sets `zip0317`) — worth a parity fix there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Zcash transaction fee calculation to properly implement ZIP-317 fee mechanism, which scales fees dynamically based on the number of logical transaction actions.

* **Tests**
  * Added comprehensive test coverage for Zcash ZIP-317 fee behavior, verifying correct fee calculation for single and multiple UTXO transactions.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelper.kt
@@ -97,6 +97,7 @@ class UtxoHelper(
             .setHashType(BitcoinScript.hashTypeForCoin(coinType))
             .setUseMaxAmount(utxo.sendMaxAmount)
             .setByteFee(utxo.byteFee.toLong())
+            .setZip0317(coinType == CoinType.ZCASH)
             .setFixedDustThreshold(coinType.getDustThreshold)
         for (item in keysignPayload.utxos) {
             val lockScript =
@@ -160,6 +161,7 @@ class UtxoHelper(
                 .setToAddress(keysignPayload.toAddress)
                 .setChangeAddress(keysignPayload.coin.address)
                 .setByteFee(utxo.byteFee.toLong())
+                .setZip0317(coinType == CoinType.ZCASH)
                 .setCoinType(coinType.value())
                 .setFixedDustThreshold(coinType.getDustThreshold)
         keysignPayload.memo?.let { input.setOutputOpReturn(ByteString.copyFromUtf8(it)) }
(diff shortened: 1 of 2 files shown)


rkokhatskyi (approved):
LGTM

aminsato (approved):
LGTM

PR #4747, opened 2026-06-05
feat(verify): human-readable param decode for non-swap contract calls

### LINKED ISSUE #4602: feat: human-readable parameter decode for non-swap contract calls (Phase 2B follow-up)
## Problem

The verify screen currently falls back to anonymous positional rows (`#1 (uint256)`, `#2 (uint256)`, `#3 ((string,string,bytes,bool,uint256))`) for any contract call that isn't a recognised swap. PR #4591 (closes #4059) only labels Uniswap Universal Router `execute` calls; everything else still renders as opaque types.

Example: a write to `0xe6313d1776e4043d906d5b7221be70cf470f5e87` calling

```
addTrait(uint256,uint256,(string,string,bytes,bool,uint256))
```

shows up as three unlabelled rows, with the tuple row collapsed to `...`. The user has no way to know what they're signing.

Verified out of scope for #4591 by the PR author — see https://github.com/vultisig/vultisig-android/pull/4591#issuecomment-4488868672.

## What to build

A generic decoder path that runs when no specialised handler (Universal Router, approve, transfer, …) matches:

- **Parameter names.** Resolve named arguments for the function selector — e.g. via the 4byte directory (already wired for selector → signature), Etherscan/Sourcify ABI lookup by contract address, or any cached source — and render `tokenId`, `traitId`, `trait` instead of `#1`, `#2`, `#3`.
- **Tuple expansion.** For tuple / struct parameters, expand each inner field as its own labelled row rather than collapsing the whole tuple to `...`. Nested tuples should recurse.
- **Type-aware value rendering.** `bytes` rows should remain hex but be clearly truncated with copy-to-full; `bool` should render as `true` / `false`; `address` should resolve via the existing token / known-contract registry before falling back to the raw address.
- **Graceful fallback.** When parameter names cannot be resolved, keep today's positional rendering — but still expand tuples and render type-aware values.

## Acceptance

- Calling `addTrait(...)` on the contract above shows three rows with parameter names and the tuple expanded into its five labelled inner fields.
- Unknown selectors with no ABI source still render readably (positional rows, expanded tuples, type-aware values).
- Universal Router and other specialised handlers continue to win over the generic path.

## Notes

- Phase 2 foundation (`AbiToCall.kt`, `FourByteRepository.decodeFunctionArgs()`) already lands decoded JSON in `TransactionDetailsUiModel.functionInputs`; this issue is about the rendering + name-resolution layer on top.
- Companion windows ticket likely needed for parity — link if/when filed.

aminsato:
## Summary

Generic EVM contract calls with no specialised decoder previously rendered as opaque positional rows on the verify screen (`#1 (uint256)`, `#3 ((string,string,bytes,bool,uint256))`) with tuples collapsed to `...`. This makes those calls human-readable.

The generic fallback now:
- **Resolves real parameter names** from the contract's verified ABI via a new keyless, cached Sourcify lookup (`ContractAbiRepository`) — e.g. `layerIndex`, `traitId`, `trait` instead of `#1`/`#2`/`#3`.
- **Expands tuple/struct params** into one labelled row per inner field, recursing through nested tuples, instead of collapsing the whole struct to a single row.
- **Renders type-aware values**: addresses resolve through the known-contract registry and stay copyable, bytes stay full-hex + copyable behind truncation, bools render `true`/`false`.
- **Falls back gracefully** to positional `#3.2 (type)` rows when no ABI name is available, while still expanding tuples and rendering type-aware values.

### Screenshot
image

## Design notes

- Sourcify is keyless and public, so no API token is required. The ABI is only used to *label* read-only rows — it is never used to execute or re-encode anything.
- Sourcify is only queried for the unknown-call fallback (skipped for swap intents and curated handlers like approve/transfer), keeping the common paths network-free.
- Results are cached per contract; an empty result is cacheable so the same contract isn't re-fetched on every row render. Transient transport failures propagate (not cached as "unverified").
- Calldata and ABI text are attacker-influenceable, so ABI names are sanitized and recursion/row counts are capped.

## Test plan

- [x] `ContractAbiRepositoryTest` — ABI fetch/parse, signature matching, caching, TTL, non-EVM short-circuit
- [x] `DecodedFunctionParamsTest` — named params, tuple expansion, nested tuples, type-aware rendering, positional fallback
- [x] `VerifyTransactionViewModelTest` / keysign wiring coverage
- [ ] Manual verify-screen check on a non-swap contract call (e.g. an NFT trait call)

Closes #4602

## Summary by CodeRabbit

* **New Features**
  * Transaction details now fetch verified ABIs (with proxy resolution) to recover parameter names and produce richer, per-field labels for tuples and arrays.
  * ABI-driven labeling is optional and subject to timeouts; resolved names used only for display.
  * Added translations for truncated-parameter messaging.

* **Bug Fixes**
  * Rendering caps and truncation indicators prevent runaway decoding depth/rows; large or unsafe values are safely limited.

* **Tests**
  * Expanded coverage for ABI resolution, caching, timeouts, concurrency, tuple expansion, and label fallbacks.


data/src/main/kotlin/com/vultisig/wallet/data/api/SourcifyApi.kt
```diff
@@ -0,0 +1,53 @@
...
+     * Throws on transport failures so the caller can distinguish a definitive "not verified" (null)
+     * from a transient outage (exception) and avoid caching the latter.
+     */
+    suspend fun fetchAbi(chainId: String, contractAddress: String): JsonArray?
+}
+
+internal class SourcifyApiImpl @Inject constructor(private val httpClient: HttpClient) :
+    SourcifyApi {
+    override suspend fun fetchAbi(chainId: String, contractAddress: String): JsonArray? {
+        val response = httpClient.get("$BASE_URL/v2/contract/$chainId/$contractAddress?fields=abi")
```
rkokhatskyi: For an upgradeable contract `dstAddress` is the proxy, whose `fields=abi` returns only `upgradeTo` / `implementation` / `admin`, so the called function never matches and `resolveParams` silently returns null for the proxied calls this is meant to label, I confirmed the USDC proxy carries no `transfer` / `approve` under `fields=abi` while `fields=abi,proxyResolution` gives the real implementation, so request `proxyResolution` and resolve the implementation ABI when `isProxy`

app/src/main/java/com/vultisig/wallet/ui/models/VerifyTransactionViewModel.kt
```diff
@@ -385,6 +387,9 @@ constructor(
                     json = json,
                     tokenMetadataResolver = tokenMetadataResolver,
                     nativeTokenLookup = { c -> nativeTokenOrNull(c.id) },
+                    resolveAbiParams = { c, address, sig ->
```
rkokhatskyi, line 390: `resolveAbiParams` is awaited before the only `transaction` emit at line 411 and before `scanTransaction()` at line 413, on a client with no `HttpTimeout` and a 3x retry, so a slow or rate limited Sourcify stalls the details and the Blockaid scan, so emit the base model first or bound it with `withTimeoutOrNull`

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1192,6 +1194,9 @@ constructor(
                             json = json,
                             tokenMetadataResolver = tokenMetadataResolver,
                             nativeTokenLookup = { c -> nativeTokenOrNull(c.id) },
+                            resolveAbiParams = { c, address, sig ->
+                                contractAbiRepository.resolveParams(c, address, sig)
```
rkokhatskyi, line 1198: Same inline await before `verifyUiModel` is published, so an incoming dApp keysign sits on the empty verify screen until Sourcify resolves, render from the local decode first and patch names in once `resolveParams` returns

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractAbiRepository.kt
```diff
@@ -0,0 +1,229 @@
...
+
+    private suspend fun runFetch(
+        chainId: String,
+        address: String,
+        key: String,
+        owned: CompletableDeferred<Map<String, List<AbiParam>>>,
+    ): Map<String, List<AbiParam>> {
+        try {
+            val abi =
+                withContext(ioDispatcher) { sourcifyApi.fetchAbi(chainId, address) }
```
rkokhatskyi, line 143: This sends the full contract address to `sourcify.dev` on every unknown verify, whereas `FourByteApi` sends only the 4 byte selector and iOS and Windows decode locally with no callout, so this is an Android only privacy delta worth a deliberate sign off

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractAbiRepository.kt
```diff
@@ -0,0 +1,229 @@
...
+@Singleton
+internal class ContractAbiRepositoryImpl
+@Inject
+constructor(
+    private val sourcifyApi: SourcifyApi,
+    @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
+) : ContractAbiRepository {
+
+    private var clock: () -> Long = { System.currentTimeMillis() }
+    private var ttl: Duration = DEFAULT_TTL
```
rkokhatskyi, line 63: `clock` and `ttl` are `var` but never reassigned and have no `@VisibleForTesting` constructor like `TokenMetadataResolver`, so the expiry path is untestable and the test plan TTL claim is unbacked
```kotlin
@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
internal constructor(
    sourcifyApi: SourcifyApi,
    ioDispatcher: CoroutineDispatcher,
    clock: () -> Long,
    ttl: Duration,
) : this(sourcifyApi, ioDispatcher) {
    this.clock = clock
    this.ttl = ttl
}
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractAbiRepository.kt
```diff
@@ -0,0 +1,229 @@
...
+                    ?.let(::parseAbi) ?: emptyMap()
+            mutex.withLock {
+                inFlight.remove(key)
+                cache[key] = CacheEntry(abi = abi, fetchedAt = clock())
+            }
+            owned.complete(abi)
+            return abi
+        } catch (e: CancellationException) {
+            mutex.withLock { inFlight.remove(key) }
+            owned.cancel(e)
```
rkokhatskyi: `owned.cancel(e)` cancels the deferred that coalesced followers share, so a follower awaiting the same contract throws `CancellationException` though it was never cancelled, complete followers with `emptyMap()` like the line 139 branch instead
```kotlin
    mutex.withLock { inFlight.remove(key) }
    owned.complete(emptyMap())
    throw e
}
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractAbiRepository.kt
```diff
@@ -0,0 +1,229 @@
...
+) : ContractAbiRepository {
+
+    private var clock: () -> Long = { System.currentTimeMillis() }
+    private var ttl: Duration = DEFAULT_TTL
+
+    private val mutex = Mutex()
+    // Cache is keyed per contract and holds the whole parsed ABI as canonical-signature -> params.
+    // An empty map is a valid, cacheable result: it means "verified-or-not, nothing here we can
+    // match" and stops us re-hitting Sourcify for the same contract every time a row renders.
+    private val cache = mutableMapOf<String, CacheEntry>()
```
rkokhatskyi: `cache` is uncapped on a `@Singleton` and never evicted, and each entry holds a whole parsed ABI, so a long session over many dApp contracts grows it without bound, a small LRU with a cap would bound it

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ContractAbiRepository.kt
```diff
@@ -0,0 +1,229 @@
...
+        val components = (obj["components"] as? JsonArray)?.mapNotNull(::toAbiParam)
+        return AbiParam(name = name, type = type, components = components)
+    }
+
+    /**
+     * Maps an EVM [Chain] to its decimal chain id for Sourcify's path. Doubles as a support
+     * allowlist: chains absent here are simply not queried. Unlisted-but-EVM chains return null
+     * rather than guessing, and an id Sourcify doesn't index just yields a graceful 404.
+     */
+    private fun evmChainId(chain: Chain): String? =
```
rkokhatskyi, line 199: `evmChainId` is a third hand kept chain id map that already drifts from `Chain.oneInchChainId()` which has no `Sei` branch, and a wrong id would fetch a different contract ABI and mislabel the signing screen, so reuse one shared chain id source

app/src/main/java/com/vultisig/wallet/ui/models/keysign/DecodedFunctionParams.kt
```diff
@@ -226,35 +243,144 @@ private fun setApprovalForAllRows(ctx: HandlerContext): List<DecodedFunctionPara
...
+    type: String?,
+    element: JsonElement?,
+    positionalPath: String,
+    namePath: String?,
+    contractLabel: (String) -> String?,
+): DecodedFunctionParam {
+    val value = element?.flatString().orEmpty()
+    val isScalar = type != null && !type.endsWith("[]")
+    val isAddress = isScalar && type.equals("address", ignoreCase = true)
+    val isBytes = isScalar && type!!.startsWith("bytes", ignoreCase = true)
```
rkokhatskyi: `type!!` trips the no `!!` rule since the null check is hidden behind `isScalar`, guard on `type` directly
```kotlin
val isBytes = isScalar && type?.startsWith("bytes", ignoreCase = true) == true
```

app/src/main/java/com/vultisig/wallet/ui/models/keysign/DecodedFunctionParams.kt
```diff
@@ -226,35 +243,144 @@ private fun setApprovalForAllRows(ctx: HandlerContext): List<DecodedFunctionPara
...
+/**
+ * Accepts an ABI-provided parameter name only when it is a plain solidity identifier of reasonable
+ * length. The ABI is attacker-influenceable (a contract author controls their own source), and the
+ * name is rendered verbatim into the signing screen, so anything with whitespace, control
+ * characters, lookalike Unicode, or excessive length is rejected back to the positional label.
+ */
+private fun sanitizedName(abi: AbiParam?): String? =
+    abi?.name?.takeIf { it.length in 1..MAX_PARAM_NAME_LENGTH && it.matches(SOLIDITY_IDENTIFIER) }
+
+private fun joinNamePath(parent: String?, child: String?): String? =
```
rkokhatskyi, line 514: When the parent path is null `joinNamePath` returns the bare child, so an anonymous outer tuple with named inner fields renders a leaf as just `amount` with no `#1.1` anchor, which is more ambiguous than the fully anonymous case, so gate the child name on the parent having one

app/src/main/java/com/vultisig/wallet/ui/models/keysign/DecodedFunctionParams.kt
```diff
@@ -226,35 +243,144 @@ private fun setApprovalForAllRows(ctx: HandlerContext): List<DecodedFunctionPara
...
+        // Addresses and bytes are middle-ellipsised by the renderer; keep the full value copyable
+        // so
+        // nothing the user is signing is hidden behind the truncation.
+        copyableValue = if (isAddress || isBytes) value.takeIf { it.isNotBlank() } else null,
+        secondary = if (isAddress) contractLabel(value) else null,
+    )
+}
+
+/** A plain tuple `(…)` — not a tuple array `(…)[]`, which renders as a single compact leaf row. */
+private fun isPlainTuple(type: String): Boolean = type.startsWith("(") && type.endsWith(")")
```
rkokhatskyi, line 455: `isPlainTuple` requires `endsWith(")")` so a tuple array like `aggregate((address,bytes)[])` falls to one `leafRow` flattened into a bracketed blob, even though the KDoc promises tuples expand per field, so iterate the array and recurse per element

app/src/main/java/com/vultisig/wallet/ui/models/keysign/DecodedFunctionParams.kt
```diff
@@ -226,35 +243,144 @@ private fun setApprovalForAllRows(ctx: HandlerContext): List<DecodedFunctionPara
...
+                        append(" (")
+                        append(type)
+                        append(')')
+                    }
+                }
+            )
+        }
+    return DecodedFunctionParam(
+        label = label,
+        value = UiText.DynamicString(value),
```
rkokhatskyi: The leaf `value` renders verbatim from `JsonPrimitive.content` with no length cap or control and bidi stripping for a `string` param that is attacker calldata, the same risk `sanitizedName` already guards on names, so bound the length and strip control and bidi characters here too

rkokhatskyi (changes requested):

aminsato:
@Vaulty-bot pick-up

app/src/main/java/com/vultisig/wallet/ui/models/keysign/DecodedFunctionParams.kt
```diff
@@ -226,35 +243,216 @@ private fun setApprovalForAllRows(ctx: HandlerContext): List<DecodedFunctionPara
...
+                abi = abi,
+                positionalPath = "$positionalPath[$k]",
+                namePath = namePath?.let { "$it[$k]" },
+                depth = depth + 1,
+                contractLabel = contractLabel,
+                out = out,
+            )
+        }
+        return
+    }
```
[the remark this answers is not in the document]
aminsato: @Vaulty-bot pick-up

aminsato:
@Vaulty-bot pick-up

data/src/main/kotlin/com/vultisig/wallet/data/api/SourcifyApi.kt
```diff
@@ -0,0 +1,93 @@
...
+        contractAddress: String,
+        withProxyResolution: Boolean,
+    ): JsonObject? {
+        val fields = if (withProxyResolution) "abi,proxyResolution" else "abi"
+        val response =
+            httpClient.get("$BASE_URL/v2/contract/$chainId/$contractAddress?fields=$fields")
+        // A 404 means "no verified contract here" — a definitive answer the caller may cache. Any
+        // other non-2xx is treated the same (best-effort enrichment), but a thrown transport error
+        // still propagates so transient failures are not cached as "unverified".
+        if (response.status != HttpStatusCode.OK) return null
```
rkokhatskyi: A persistent `429` or `503` returns `null` here so `runFetch` caches `emptyMap` under the 24h `ttl` and suppresses names for a day, whereas the interface KDoc promises transient failures throw and stay uncached, since `HttpRequestRetry` exhausts its retries then returns the non-OK response rather than throwing
```kotlin
if (response.status == HttpStatusCode.NotFound) return null
if (response.status != HttpStatusCode.OK) error("Sourcify ${response.status}")
```

app/src/main/java/com/vultisig/wallet/ui/models/keysign/DecodedCallEnrichment.kt
```diff
@@ -94,6 +98,21 @@ internal suspend fun enrichDecodedCall(
...
+    // Only the positional fallback benefits from recovered names, so skip the verified-ABI lookup
+    // entirely for swap intents and for calls a semantic handler already labels (approve,
+    // transfer…).
+    // Bound the Sourcify lookup: it runs on a client with no HttpTimeout and a 3x retry, and the
+    // caller awaits this before emitting the verify model and starting the security scan, so a slow
+    // or rate-limited Sourcify must not stall the signing screen. A timeout just drops the
+    // recovered
+    // names (the positional fallback still renders).
+    val abiParams =
+        if (urRows == null && !hasSemanticHandler(signature)) {
```
rkokhatskyi, line 110: `hasSemanticHandler` matches the function name only so a same-name different-arity call like `transfer(address,uint256,bytes)` skips the ABI lookup here, then `transferRows` consumes only inputs 0 and 1 so the trailing `bytes` is dropped from the signing screen while the recovered names that would have rendered it stay suppressed
```kotlin
if (urRows == null && !hasSemanticHandlerFor(signature, functionInfo.inputs)) {
```

app/src/main/java/com/vultisig/wallet/ui/models/keysign/DecodedFunctionParams.kt
```diff
@@ -226,35 +243,222 @@ private fun setApprovalForAllRows(ctx: HandlerContext): List<DecodedFunctionPara
...
-            append('#')
-            append(index + 1)
-            if (type != null) {
-                append(" (")
-                append(type)
-                append(')')
+    val out = mutableListOf<DecodedFunctionParam>()
+    val count = maxOf(types.size, inputs.size)
+    for (index in 0 until count) {
+        if (out.size >= MAX_PARAM_ROWS) break
```
rkokhatskyi: Rows past `MAX_PARAM_ROWS` (64) are dropped with no indicator so a batch or multicall with more than 64 leaf params renders a partial list that looks complete while the signer authorises params that never appeared on screen
```kotlin
if (out.size >= MAX_PARAM_ROWS) { out += truncatedRow(count - index); break }
```

data/src/main/kotlin/com/vultisig/wallet/data/api/SourcifyApi.kt
```diff
@@ -0,0 +1,93 @@
...
+     * an upgradeable proxy, or null when it isn't a proxy (or no implementation is listed).
+     */
+    private fun proxyImplementationAddress(root: JsonObject): String? {
+        val proxyResolution = root["proxyResolution"] as? JsonObject ?: return null
+        if (proxyResolution["isProxy"]?.jsonPrimitive?.booleanOrNull != true) return null
+        val implementations = proxyResolution["implementations"] as? JsonArray ?: return null
+        return implementations
+            .asSequence()
+            .mapNotNull { (it as? JsonObject)?.get("address")?.jsonPrimitive?.content }
+            .firstOrNull { it.isNotBlank() }
```
rkokhatskyi: `firstOrNull` keeps a single implementation so an EIP-2535 diamond whose `proxyResolution.implementations` lists several facet addresses resolves names only for the first facet while calls routed to the other facets stay unlabelled
```kotlin
// merge every facet ABI when implementations.size > 1, or document diamonds as unsupported
```

data/src/test/kotlin/com/vultisig/wallet/data/repositories/ContractAbiRepositoryTest.kt
```diff
@@ -0,0 +1,123 @@
...
+        val api = FakeSourcifyApi(abiJson = null)
+        val repo = newRepo(api)
+
+        assertNull(repo.resolveParams(Chain.Ethereum, CONTRACT, addTraitSignature))
+        assertNull(repo.resolveParams(Chain.Ethereum, CONTRACT, addTraitSignature))
+        assertEquals(1, api.calls.get())
+    }
+
+    private fun newRepo(api: SourcifyApi) =
+        ContractAbiRepositoryImpl(sourcifyApi = api, ioDispatcher = UnconfinedTestDispatcher())
```
rkokhatskyi: Request coalescing and the `CancellationException` follower path are untested so the `owned.complete(emptyMap())` follower fix from this review and the `Slot.Wait` dedupe can regress silently, since `newRepo` only drives sequential single-caller lookups and never two concurrent `resolveParams` on one contract

data/src/test/kotlin/com/vultisig/wallet/data/repositories/ContractAbiRepositoryTest.kt
```diff
@@ -0,0 +1,123 @@
...
+    fun `unverified contract resolves to null and is not refetched`() = runTest {
+        val api = FakeSourcifyApi(abiJson = null)
+        val repo = newRepo(api)
+
+        assertNull(repo.resolveParams(Chain.Ethereum, CONTRACT, addTraitSignature))
+        assertNull(repo.resolveParams(Chain.Ethereum, CONTRACT, addTraitSignature))
+        assertEquals(1, api.calls.get())
+    }
+
+    private fun newRepo(api: SourcifyApi) =
```
rkokhatskyi: The `@VisibleForTesting(clock, ttl)` ctor added for the earlier TTL review is never exercised since `newRepo` calls the two-arg ctor, so expiry past `ttl` and LRU eviction past `MAX_CACHE_ENTRIES` stay unverified and the test plan TTL claim is still unbacked

rkokhatskyi (dismissed):

aminsato:
@Vaulty-bot pick-up

aminsato:
@Vaulty-bot pick-up

johnnyluo (approved):
LGTM

PR #4760, opened 2026-06-06
fix(defi): left-align icon, center label in DeFi action buttons

johnnyluo:
## Summary

In the DeFi action buttons (`ActionButton` in `DeFiComponents.kt`), the leading icon and the label were grouped together and centered as a unit, leaving dead space on the left. Figma docks the icon to the button's **start edge** while the **label stays centered** across the full button width.

This lays the button content out in a full-width `Box` with the icon pinned to `CenterStart` and the label centered on top, and evens out the horizontal content padding so the label centers on the true button center. The label reserves the icon lane symmetrically (`maxLines = 1` + ellipsis) so long translated labels stay centered without overlapping the icon.

`ActionButton` is the shared DeFi action button, so this applies consistently to every DeFi action button that has an icon (Staking Stake/Unstake, LP, Bonded, and the Circle USDC Withdraw).

> **Note:** Tron freeze/unfreeze render through a separate component (`TronFreezeActionButton` in `TronFreezePositionCard.kt`) that this PR does not touch. Consolidating it into the shared `ActionButton` is a follow-up, out of scope here.

## Before / After

Captured on emulator via `PreviewActivity` (Circle USDC deposited card).

| Before | After |
|--------|-------|
| before | after |

The ⊖ icon now sits at the button's left edge with "Withdraw" centered, matching Figma.

## Notes

- The Circle USDC **Deposit** button has no icon in the app but shows a circled **+** in Figma — that's a separate mismatch, left out of scope here.
- A debug-only `circle_usdc_widget` preview was added to `PreviewActivity` to capture the screenshots.

## Test plan

- [x] `./gradlew :app:ktfmtCheck` passes
- [x] `./gradlew :app:assembleDebug` / `installDebug` succeed
- [x] Verified on emulator: icon left-aligned, label centered (see After)
- [ ] Spot-check Staking / LP / Bonded action buttons render correctly

## Summary by CodeRabbit

* **New Features**
  * Added a preview for the Circle USDC DeFi widget to visualize header, totals and actions.

* **Bug Fixes**
  * Improved action button layout to properly center labels and handle icon positioning when present.
  * Enhanced text truncation with ellipsis to prevent label overlap and preserve centering.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
@@ -286,30 +286,39 @@ fun ActionButton(
                 }
             },
         shape = RoundedCornerShape(50),
-        contentPadding = PaddingValues(start = 4.dp, top = 6.dp, end = 16.dp, bottom = 6.dp),
+        contentPadding = PaddingValues(horizontal = 4.dp, vertical = 6.dp),
         modifier = modifier.height(42.dp).actionButtonInnerBevel(enabled = enabled),
     ) {
-        if (icon != null) {
-            Box(
-                modifier =
-                    Modifier.size(30.dp)
-                        .background(
-                            if (enabled) iconCircleColor else iconCircleColor.copy(alpha = 0.5f),
-                            RoundedCornerShape(50),
-                        ),
-                contentAlignment = Alignment.Center,
-            ) {
-                Icon(
-                    painter = painterResource(id = icon),
-                    contentDescription = null,
-                    modifier = Modifier.size(12.dp),
-                    tint = if (enabled) contentColor else contentColor.copy(alpha = 0.5f),
-                )
+        // Figma docks the icon to the button's leading edge while the label stays centered across
+        // the full width, so the icon is absolutely positioned and the text is centered on top.
+        Box(modifier = Modifier.fillMaxWidth()) {
+            if (icon != null) {
+                Box(
+                    modifier =
+                        Modifier.align(Alignment.CenterStart)
+                            .size(30.dp)
+                            .background(
+                                if (enabled) iconCircleColor
+                                else iconCircleColor.copy(alpha = 0.5f),
+                                RoundedCornerShape(50),
+                            ),
+                    contentAlignment = Alignment.Center,
+                ) {
+                    Icon(
+                        painter = painterResource(id = icon),
+                        contentDescription = null,
+                        modifier = Modifier.size(12.dp),
+                        tint = if (enabled) contentColor else contentColor.copy(alpha = 0.5f),
+                    )
+                }
             }
-            UiSpacer(5.dp)
-        }
 
-        Text(text = title, style = Theme.brockmann.button.medium.medium)
+            Text(
+                text = title,
+                style = Theme.brockmann.button.medium.medium,
+                modifier = Modifier.align(Alignment.Center),
+            )
+        }
     }
 }
 
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
@@ -47,6 +47,7 @@ import androidx.compose.ui.layout.LayoutCoordinates
 import androidx.compose.ui.layout.onGloballyPositioned
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.IntOffset
 import androidx.compose.ui.unit.IntSize
@@ -316,7 +317,14 @@ fun ActionButton(
             Text(
                 text = title,
                 style = Theme.brockmann.button.medium.medium,
-                modifier = Modifier.align(Alignment.Center),
+                maxLines = 1,
+                overflow = TextOverflow.Ellipsis,
+                // Reserve the icon lane symmetrically so a long (translated) label can never paint
+                // over the docked icon, while keeping the label centered on the button's true
+                // center.
+                modifier =
+                    Modifier.align(Alignment.Center)
+                        .padding(horizontal = if (icon != null) 30.dp else 0.dp),
             )
         }
     }
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -112,6 +112,7 @@ import com.vultisig.wallet.ui.screens.transaction.TransactionHistoryEmptyState
 import com.vultisig.wallet.ui.screens.transaction.UiTransactionInfo
 import com.vultisig.wallet.ui.screens.transaction.UiTransactionInfoType
 import com.vultisig.wallet.ui.screens.v2.chaintokens.ChainTokensScreen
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
```diff
@@ -286,30 +286,39 @@ fun ActionButton(
...
-            ) {
-                Icon(
-                    painter = painterResource(id = icon),
-                    contentDescription = null,
-                    modifier = Modifier.size(12.dp),
-                    tint = if (enabled) contentColor else contentColor.copy(alpha = 0.5f),
-                )
+        // Figma docks the icon to the button's leading edge while the label stays centered across
+        // the full width, so the icon is absolutely positioned and the text is centered on top.
+        Box(modifier = Modifier.fillMaxWidth()) {
```
aminsato, line 295: To avoid duplicating this layout, `TronFreezeActionButton` (in `TronFreezePositionCard.kt`) already renders the same icon-left / centered-label pill. It achieves it with a `Row` of `[icon circle] + spacer + Text(weight(1f), TextAlign.Center)` rather than an absolutely-positioned `Box`, which is simpler and avoids the centered label overlapping the icon on narrow/long-label buttons.

These two are otherwise near-identical (only `height` 42 vs 46.dp, icon circle 30 vs 34.dp, and the bevel differ). Worth consolidating into a single shared button parameterized on those values and reusing the `Row`+weighted-`Text` approach here, instead of maintaining two parallel implementations.

johnnyluo: Thanks — agreed these two are near-identical and a shared button is appealing. But the two layouts are not visually equivalent, and that's the reason I went with the absolutely-positioned `Box` here:

- **`Box` + `Text(align = Center)`** centers the label on the button's *true* center, with the icon docked at the edge. That's exactly the Figma spec this PR implements ("icon at the start edge, label centered across the full button width").
- **`Row` + `Text(weight(1f), TextAlign.Center)`** (the Tron approach) centers the label only in the space *right of* the icon, so it sits shifted right of true center. Reusing it here would regress the exact mismatch this PR fixes.

On the overlap point: it's theoretically possible, but every real label is short (Stake/Unstake/Withdraw/LP/Bond), so the centered text never reaches the icon in practice.

Consolidating into one shared button parameterized on height/circle-size/bevel is a good follow-up, but it'd also need to settle on a single centering behavior (and re-screenshot the Tron buttons), so I'd rather keep that out of this narrowly-scoped fix. Resolving for now.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
```diff
@@ -286,30 +286,39 @@ fun ActionButton(
...
+                }
             }
-            UiSpacer(5.dp)
-        }
 
-        Text(text = title, style = Theme.brockmann.button.medium.medium)
+            Text(
+                text = title,
+                style = Theme.brockmann.button.medium.medium,
+                modifier = Modifier.align(Alignment.Center),
```
rkokhatskyi: **Label overlaps the docked icon for long locales**

The label is centered across the full width with nothing reserved for the docked 30dp icon.

- On the `weight(1f)` rows (half the card width) long translated labels like nl `Verwijderen` (`remove`) or es/pt `Desvincular` (`unbond`) paint over the icon circle
- The `Text` has no `maxLines`/`overflow`, so an over-long label cannot truncate
- The previous measured `Row(icon + spacer + Text)` could never overlap since the icon consumed width
- iOS `DefiButton` (`paddingLeading = (iconSize*2 + iconPadding)/2`) and Windows `DeFiActionButton` (`padding-left = iconContainerSize + iconPadding*2`) both reserve the icon lane

Suggested fix, reserve the lane and let an over-long label ellipsize:

```kotlin
Text(
    text = title,
    style = Theme.brockmann.button.medium.medium,
    maxLines = 1,
    overflow = TextOverflow.Ellipsis,
    modifier = Modifier
        .align(Alignment.Center)
        .padding(start = if (icon != null) 30.dp else 0.dp),
)
```

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
```diff
@@ -286,30 +286,39 @@ fun ActionButton(
                 }
             },
         shape = RoundedCornerShape(50),
-        contentPadding = PaddingValues(start = 4.dp, top = 6.dp, end = 16.dp, bottom = 6.dp),
+        contentPadding = PaddingValues(horizontal = 4.dp, vertical = 6.dp),
```
rkokhatskyi, line 290: **Summary overclaims Tron freeze coverage**

The summary lists Tron freeze among the covered buttons, but they use a separate component this PR does not touch.

- Tron freeze/unfreeze render through `TronFreezeActionButton` in `TronFreezePositionCard.kt`, which is not in this diff
- It keeps the old `start = 4`/`end = 16` padding, `height(46.dp)`, a 34dp icon, and `Text(Modifier.weight(1f), textAlign = TextAlign.Center)`
- So they now diverge from the buttons changed here in both metrics and label centering, while iOS routes every DeFi action through one shared `DefiButton`

Fix: either fold `TronFreezeActionButton` into the shared `ActionButton`, or narrow the summary so the consistency claim holds.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
johnnyluo: Good catch on the long-locale overlap. Fixed in the latest commit — I added `maxLines = 1` + `overflow = TextOverflow.Ellipsis`, and reserved the icon lane.

One tweak vs. your snippet: I used **symmetric** `padding(horizontal = 30.dp)` rather than `start = 30.dp`. A start-only pad would shift the label 15dp right of the button's true center, regressing the exact centering this PR fixes. Symmetric padding keeps the glyphs centered on the true center while reserving the 30dp icon lane on the start (and an equal lane on the end), so the label can never reach the icon and ellipsizes if it's still too long.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
johnnyluo: Agreed — narrowed the summary. Removed "Tron freeze" from the covered-buttons list and added an explicit note that `TronFreezeActionButton` (`TronFreezePositionCard.kt`) is a separate component not touched here, with consolidation called out as a follow-up. Folding it into the shared `ActionButton` is out of scope for this narrowly-focused fix (it also differs in height/icon-size/bevel and label centering, and would need re-screenshotting).

scripts/approve-merge-pr.sh
```diff
@@ -0,0 +1,18 @@
...
+fi
+
+PR=$1
+BODY=${2:-"LGTM"}
+
+echo "==> Approving PR #$PR..."
+gh pr review "$PR" --approve --body "$BODY"
+
+echo "==> Adding to merge queue..."
+gh pr merge "$PR" --auto --squash
```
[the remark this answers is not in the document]
rkokhatskyi: @johnnyluo looks like this file was unintentionally pushed the to repo, please put it to ignore locally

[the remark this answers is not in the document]
johnnyluo: yup, removed

rkokhatskyi (approved):
LGTM

PR #4773, opened 2026-06-07
fix: cache Cosmos staking positions so the screen no longer flashes empty (#4764)

### LINKED ISSUE #4764: Terra DeFi detail: staking values not cached on entry (flashes empty/zero before re-fetch)
## Observed
Entering the Terra / TerraClassic DeFi (staking) detail shows empty/zero values briefly before the network fetch completes — the previously-loaded values aren't cached, so each entry re-fetches from scratch.

`app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt`:
- Initial state is empty (`CosmosStakingPositionsUiState(positions = emptyList(), isLoading = false …)`, ~L44-50).
- On mount, `setData → loadCoin → refresh()` fetches fresh; there's no read from a cache/repository to seed the UI immediately.

So on (re)entry the screen renders the empty state until the fetch lands. Suggest seeding from a cached last-known positions/balance (repository cache or in-memory), mirroring how the user expects other DeFi detail screens to behave.

## Platform
Android. Related: portfolio-allocation issue (filed separately).

aminsato:
## Summary

Fixes #4764. The LUNA / LUNC DeFi staking-positions screen flashed empty/zero values on every entry before the network fetch landed, because positions were never cached — the ViewModel is recreated per navigation, so each (re)open started from an empty state and dropped to the empty view until the 4-call LCD fan-out completed.

This applies the same in-memory snapshot-cache pattern already shipped for TRON DeFi (#4568).

## Changes

- **New `CosmosStakingSnapshotCache`** — a `@Singleton` holding a `ConcurrentHashMap` keyed by `chain:address` (so LUNA and LUNC snapshots never collide), storing the last-known display-ready snapshot (positions, pending unbondings, total staked + fiat). In-memory and process-lifetime — no on-disk store to clean up on vault/coin deletion, matching how the other DeFi chains and vultisig-ios keep state.
- **`CosmosStakingPositionsViewModel`**:
  - Seeds the UI from the cached snapshot immediately on reopen and refreshes in the background; only shows the loading state on a true first open with nothing cached.
  - Writes the fresh snapshot to the cache on every successful load.
  - A background-refresh failure on a cache-seeded screen now keeps the data visible (via a new `onRefreshFailed` guard) instead of wiping it with an error banner — a transient LCD blip no longer blanks a screen the user can already read. The first-open-with-no-cache failure still surfaces the error + retry.

## Test plan

- `:app:testDebugUnitTest` → **14 passed, 0 failed** (11 existing + 3 new):
  - a successful load writes a snapshot into the cache
  - a cached snapshot renders immediately without a loading flash (and survives a background-refresh failure)
  - a refresh failure with no cache still surfaces the error
- `ktfmt` applied to all touched files.

## Summary by CodeRabbit

* **New Features**
  * Cosmos staking positions are now cached so the UI can show instant, previously-rendered staking data when returning to the screen.
  * Pull-to-refresh spinner now reflects the view model's refresh state.

* **Bug Fixes**
  * Refresh errors no longer wipe existing positions; an error is shown only if no cached data is available.
  * Refresh flow more robust: critical failures abort cleanly and the refresh-spinner clears after load completes.

* **Tests**
  * Added tests covering snapshot caching, error-surfaces, and refresh-spinner behavior.


app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
```diff
@@ -139,12 +142,33 @@ constructor(
...
+            val cached = snapshotCache.read(cacheKey)
+            if (cached != null) {
+                _state.update {
+                    it.copy(
+                        positions = cached.positions,
+                        pendingUnbondings = cached.pendingUnbondings,
+                        totalStaked = cached.totalStaked,
+                        totalStakedFiat = cached.totalStakedFiat,
+                        totalAmountPrice = cached.totalAmountPrice,
+                        isLoading = false,
```
rkokhatskyi: This cached branch keeps `isLoading` false so once the snapshot is warm `state.isLoading` never toggles, and the screen's only `isRefreshing` reset `LaunchedEffect(state.isLoading)` at `CosmosStakingPositionsScreen.kt:90` never runs again so the refresh spinner spins forever after the first load
```kotlin
// VM owns the refresh flag and the screen collects it
private val _isRefreshing = MutableStateFlow(false)
val isRefreshing = _isRefreshing.asStateFlow()
// in refresh(): _isRefreshing.value = true ; finally { _isRefreshing.value = false }
```

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
```diff
@@ -262,9 +285,23 @@ constructor(
                     totalStakedFiat = totalStakedFiat,
                     totalAmountPrice = totalStakedFiat,
                     isLoading = false,
+                    errorMessage = null,
                 )
             }
 
+            // Cache the display-ready result so the next open within this session renders
+            // instantly.
+            snapshotCache.write(
```
rkokhatskyi: `refresh()` launches an uncancelled coroutine on every call so a pull racing the initial load, or two quick pulls, both reach this unconditional `snapshotCache.write` and a slow older response that lands last pins the screen and the seeded snapshot to stale data, whereas the TRON view model it mirrors guards this with `loadJob?.cancel()`
```kotlin
private var loadJob: Job? = null
fun refresh() { loadJob?.cancel(); loadJob = viewModelScope.safeLaunch(...) { /* fanout */ } }
```

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
```diff
@@ -262,9 +285,23 @@ constructor(
...
+                cacheKey,
+                CosmosStakingSnapshot(
+                    positions = positions,
```
rkokhatskyi: A validators or price read that degraded to empty or zero gets persisted here as the known good snapshot so every delegation folds to `ChurnedOut` at `$0.00`, and that alarming view reseeds the screen on the next reopen until the live refresh repairs it, freezing a transient LCD failure as cached truth
```kotlin
// keep each sub read's Result and skip snapshotCache.write when validators or price failed
// so a degraded all ChurnedOut / $0.00 snapshot is never frozen as the known good value
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/CosmosStakingSnapshotCache.kt
```diff
@@ -0,0 +1,40 @@
...
+ * Last-known, display-ready staking state for a Cosmos-SDK staking address (LUNA / LUNC). The rows
+ * already carry their fiat strings, so a re-entry can render this snapshot verbatim while the live
+ * fan-out refreshes in the background.
+ */
+data class CosmosStakingSnapshot(
+    val positions: List<CosmosStakePositionRow>,
+    val pendingUnbondings: List<CosmosUnbondingDelegation>,
+    val totalStaked: BigDecimal,
+    val totalStakedFiat: String,
+    val totalAmountPrice: String,
```
rkokhatskyi: `totalAmountPrice` always equals `totalStakedFiat` since the write at `CosmosStakingPositionsViewModel.kt:300` sets both from the same value and the `UiState` doc says they carry the same number, so this field is dead weight that can be dropped
```kotlin
// drop totalAmountPrice from CosmosStakingSnapshot; on seed set both slots from one field
// totalStakedFiat = cached.totalStakedFiat ; totalAmountPrice = cached.totalStakedFiat
```

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
aminsato: Good catch. Moved the refresh flag into the VM: it now owns `isRefreshing` (set true at the start of every `refresh()`, reset in a `finally` guarded by `isActive` so a superseded run can't clear it). The screen collects `viewModel.isRefreshing` directly and the `LaunchedEffect(state.isLoading)` reset is gone — so the spinner stops even on a cache-seeded refresh that never toggles `isLoading`.

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
aminsato: Fixed — added `private var loadJob: Job?` and `loadJob?.cancel()` before each relaunch, matching the TRON view model. A slow older response can no longer land last and pin the screen/cache to stale data.

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
aminsato: Agreed. `fetchValidators` now keeps its `Result` and the price keeps a nullable `resolvedPrice`; the `snapshotCache.write` is gated on `validatorsResult.isSuccess && priceOk`. A degraded all-`ChurnedOut` / $0.00 view is still rendered but never cached, so a transient LCD blip isn't frozen as known-good truth. Added tests for both the degraded-validators and missing-price cases.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/CosmosStakingSnapshotCache.kt
aminsato: Done — dropped `totalAmountPrice` from `CosmosStakingSnapshot`; the seed path now sets both UI slots (`totalStakedFiat` and `totalAmountPrice`) from the single `cached.totalStakedFiat`.

johnnyluo (approved):
LGTM

PR #4783, opened 2026-06-07
fix(keysign): drop misleading "+" in device count, cut resend cooldown to 30s

### LINKED ISSUE #4769: Keysign Scan QR: remove the "+" in Devices (1/2+) and cut resend-notification cooldown 60s → 20s
## 1. Remove the "+" in "Devices (1/2+)"
On the keysign peer-discovery (Scan QR) screen, the device count shows e.g. **"Devices (1/2+)"**. The "+" appears for any vault with redundancy (m-of-n, n>threshold, e.g. 2-of-3), but for keysign you only ever need exactly the threshold — the "+" is misleading. Should read **"Devices (1/2)"**.

- Driver: `app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt:158` — `allowsMoreDevices = vault.signers.size > minimumDevices` (true whenever signers > threshold).
- Render: `app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt:268-270` — picks `peer_discovery_devices_n_of_n_plus` when `allowsMoreDevices`.
- Fix: pass `allowsMoreDevices = false` for keysign (or otherwise suppress the "+"), so it shows the plain `n_of_n`.
- NOTE: do **not** touch the keygen "X/4+" behavior (#4595) — that's intentional for 4+ device vault creation. This is keysign-only.

## 2. Reduce resend-notification cooldown 60s → 20s
- `app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt:449` — `startResendCooldown()` sets `var seconds = 60`. Change to `20`.

## Platform
Android. (Screenshot shows "Resend notification in 48 seconds" = mid-countdown of the 60s timer.)

aminsato:
Closes #4769

## Summary
On the keysign **Scan QR** (peer-discovery) screen, the device count showed `Devices (1/2+)` for any vault with redundancy (signers > threshold, e.g. a 2-of-3 vault). For keysign you only ever need exactly the threshold, so the `+` is misleading.

- **Remove the `+`** — force `allowsMoreDevices = false` at the keysign call site (`KeysignPeerDiscovery.kt`) so the count renders as the plain `n_of_n` → `Devices (1/2)`. Keygen's intentional `X/4+` behavior (#4595) is a separate call site and is **not** touched.
- **Resend cooldown 60s → 30s** — `startResendCooldown()` in `KeysignFlowViewModel.kt`.
- Added a debug-only `PreviewActivity` case to render both device-count states for the screenshots below.

## Before / After

| Before (`1/2+`) | After (`1/2`) |
|--------|-------|
| before | after |

## Test plan
- Rendered the real `PeerDiscoveryScreen` for a 2-of-3 keysign vault via `PreviewActivity` + `adb screencap`; confirmed the `+` is gone with `allowsMoreDevices = false` and present with `true`.
- Verified the render branch (`PeerDiscoveryScreen.kt`) selects `peer_discovery_devices_n_of_n` (no `+`) when `allowsMoreDevices` is false.
- `:app:ktfmtFormat` clean; `:app:installDebug` succeeds.

## Summary by CodeRabbit

* **Improvements**
  * Shortened resend cooldown for faster retrying during keysign operations.
  * Disabled adding extra devices in keysign peer discovery so only the exact required threshold is allowed.

* **Behavior**
  * Keygen device-picker now treats the “4+” selection consistently so the UI’s “+” affordance appears when appropriate.

* **New Features**
  * Added preview variants for peer-discovery screens.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
@@ -446,7 +446,7 @@ constructor(
         resendCooldownJob?.cancel()
         resendCooldownJob =
             viewModelScope.launch {
-                var seconds = 20
+                var seconds = 30
                 while (seconds > 0) {
                     uiState.update { it.copy(resendCooldownSeconds = seconds) }
                     delay(1.seconds)
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
@@ -446,7 +446,7 @@ constructor(
         resendCooldownJob?.cancel()
         resendCooldownJob =
             viewModelScope.launch {
-                var seconds = 60
+                var seconds = 20
                 while (seconds > 0) {
                     uiState.update { it.copy(resendCooldownSeconds = seconds) }
                     delay(1.seconds)
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
@@ -155,8 +155,6 @@ internal fun KeysignPeerDiscovery(
                     selectedDevices = selectionState.filter { it != vault.localPartyID },
                     minimumDevices = minimumDevices,
                     minimumDevicesDisplayed = minimumDevices,
-                    // Keysign needs exactly the threshold, so never show the "+".
-                    // (Keygen's "X/4+" behavior is intentional — see #4595 — and lives elsewhere.)
                     allowsMoreDevices = false,
                     showQrHelpModal = false,
                     connectingToServer = null,
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
@@ -155,7 +155,9 @@ internal fun KeysignPeerDiscovery(
                     selectedDevices = selectionState.filter { it != vault.localPartyID },
                     minimumDevices = minimumDevices,
                     minimumDevicesDisplayed = minimumDevices,
-                    allowsMoreDevices = vault.signers.size > minimumDevices,
+                    // Keysign needs exactly the threshold, so never show the "+".
+                    // (Keygen's "X/4+" behavior is intentional — see #4595 — and lives elsewhere.)
+                    allowsMoreDevices = false,
                     showQrHelpModal = false,
                     connectingToServer = null,
                     error = null,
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
```diff
@@ -155,7 +155,9 @@ internal fun KeysignPeerDiscovery(
                     selectedDevices = selectionState.filter { it != vault.localPartyID },
                     minimumDevices = minimumDevices,
                     minimumDevicesDisplayed = minimumDevices,
-                    allowsMoreDevices = vault.signers.size > minimumDevices,
+                    // Keysign needs exactly the threshold, so never show the "+".
+                    // (Keygen's "X/4+" behavior is intentional — see #4595 — and lives elsewhere.)
```
rkokhatskyi: The comment is unnecessary

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
```diff
@@ -446,7 +446,7 @@ constructor(
         resendCooldownJob?.cancel()
         resendCooldownJob =
             viewModelScope.launch {
-                var seconds = 60
+                var seconds = 20
```
rkokhatskyi: `Resend cooldown 60s → 20s — startResendCooldown() in KeysignFlowViewModel.kt.`

both iOS and Windows use 30 seconds, not 20

@realpaaao shall we stick to 30 as other platforms or change to 20 on all platforms?

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
```diff
@@ -155,7 +155,9 @@ internal fun KeysignPeerDiscovery(
...
+                    allowsMoreDevices = false,
```
rkokhatskyi, line 158: this flips the only `allowsMoreDevices = true` producer to false while keygen and the other paths already rely on the default false, so the `if (state.allowsMoreDevices)` branch at `PeerDiscoveryScreen.kt:268` and the now orphaned `peer_discovery_devices_n_of_n_plus` string across ten locales are dead and worth removing in this PR

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
realpaaao: Yes, please 30s to get align with other platforms. Was my bad, I had 20s in my mind for ios

rkokhatskyi: @aminsato please update it

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignPeerDiscovery.kt
johnnyluo: Good catch on the dead branch. Rather than delete it, we wired up the intended behavior: `KeygenPeerDiscoveryViewModel` now sets `allowsMoreDevices = true` when the user picks the "4+" option on the device-count screen (`deviceCount >= 4`), so the "+" label is now driven by real intent. Keysign stays at exactly the threshold (`false`). The `peer_discovery_devices_n_of_n_plus` string and the branch are therefore live again, not orphaned.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
johnnyluo: Updated to 30s to align with iOS/Windows.

johnnyluo (approved):
LGTM

PR #4788, opened 2026-06-08
fix(defi): reflect Terra/TerraClassic staked allocation in portfolio (#4763)

### LINKED ISSUE #4763: DeFi portfolio: Terra/TerraClassic staked allocation not reflected (shows $0.00 despite active delegation)
## Observed
On the DeFi tab portfolio, **Terra** and **TerraClassic** show $0.00 / 0 LUNA / 0 LUNC even though the vault has an active LUNC delegation (~74,729 LUNC ≈ $4.56). The portfolio total ($5.50 = Circle $5.00 + THORChain $0.50) omits the staked Terra allocation entirely — so the staked value is invisible in the portfolio.

The per-chain DeFi row value comes from the liquid account balance via `loadDeFiBalances` / `loadDeFiAddresses` — `app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt:484-506`. For Terra/TerraClassic the liquid balance is ~0 (everything is delegated), and the **delegated + pending-reward** amount is not aggregated into the row or the total. (Terra/TerraClassic route to `ChainDashboardRoute.PositionCosmosStaking` at `:402-407`, but that's only the detail screen — the portfolio value doesn't pull from it.)

Fix: include the Cosmos delegated (+ claimable reward) value in the Terra/TerraClassic DeFi account value and the DeFi total, so the allocation shows.

## Platform
Android. Related: detail-caching issue (filed separately).

aminsato:
## Summary

Fixes #4763. A Terra / TerraClassic vault whose funds are fully delegated has a ≈0 liquid balance, so the DeFi tab counted the position as **$0.00** — the staked allocation never reached the chain row value or the DeFi portfolio total.

This folds the staked total into the native account's value during the DeFi load so both the row and the total reflect it.

## Changes

- **`CosmosStakedBalanceUseCase`** (new) — sums delegated balance + claimable bond-denom rewards in base units for a Cosmos-SDK staking address (LUNA / LUNC). Floors fractional `cosmos.Dec` rewards and counts only the bond denom (ignores legacy Terra Classic stability-tax denoms). Every read degrades to `ZERO` so a non-staking chain or a transient LCD failure simply omits the staked value instead of failing the whole DeFi load.
- **`AccountsRepository`** — `foldInCosmosStakedBalance()` adds the staked total to the native token's `tokenValue`/`fiatValue` (re-priced from the resolved unit price) on both the cached emit and the network refresh of `loadDeFiAddresses`. Since the DeFi total is the sum of row fiat values, the total picks it up automatically. Mirrors iOS `DefiBalanceService.cosmosStakingTotalBalance`.
- **In-memory snapshot** inside the use case (same seed-then-refresh pattern as `TronDeFiSnapshotCache`): the cached emit reads the last-known staked value with no network so it renders instantly on reopen, and a transient delegations failure keeps the previous snapshot rather than zeroing it.

## Behavior

| | Before | After |
|---|---|---|
| TerraClassic row (74,729 LUNC delegated, ~0 liquid) | `$0.00` | `$4.56` |
| DeFi total | excludes staking | includes staking |

## Test plan

- [x] `CosmosStakedBalanceUseCaseTest` — sum + floored rewards, denom filtering, non-staking short-circuit, degrade-to-last-snapshot on failure, cached-before/after-live read
- [x] `AccountsRepositoryImplTest` — updated with the new dependency
- [x] `./gradlew :data:testDebugUnitTest` green; ktfmt applied
- [ ] Manual: vault with an active Terra/TerraClassic delegation shows the staked value in the DeFi tab and total

## Summary by CodeRabbit

* **New Features**
  * Staking balance calculations now include claimable rewards combined with delegated amounts into a single total, giving you a complete view of your entire staked position.

* **Bug Fixes**
  * Improved error handling ensures staking information displays accurately and remains accessible even when temporary issues prevent reward data from being successfully retrieved.


data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -434,7 +438,9 @@ constructor(
                                     .awaitAll()
                             val canBeDeFiProvider = account.chain.isDeFiSupported
 
-                            account.copy(accounts = newAccounts, isDefiProvider = canBeDeFiProvider)
+                            account
+                                .copy(accounts = newAccounts, isDefiProvider = canBeDeFiProvider)
+                                .foldInCosmosStakedBalance(live = true)
```
johnnyluo: **Critical: likely double-counts the delegated balance.**

By the time we reach this `.foldInCosmosStakedBalance(live = true)`, `account` has already been through `balanceRepository.getDefiTokenBalanceAndPrice(...)` above. For `Terra`/`TerraClassic` that routes to `cosmosStakingDeFiBalanceService.getRemoteDeFiBalance(...)` (`BalanceRepository.kt:403-405`), which **already sums delegations** and returns them as the native coin's DeFi balance — so `applyBalance(...)` has already set the native account's `tokenValue` to the delegated total.

This fold then adds `delegated + rewards` on top of that same account:
```
account value = delegated            (existing CosmosStakingDeFiBalanceService)
              + delegated + rewards  (this fold)
              = 2 × delegated + rewards
```
The cached emit has the same overlap via `getCacheDeFiBalance` (`BalanceRepository.kt:230-232`). Note both this use case and the existing service document themselves as mirroring the same iOS `DefiBalanceService.cosmosStakingTotalBalance` — a strong sign they solve the same problem twice.

**Recommendation:** consolidate to a single source of truth — either add rewards to the existing `CosmosStakingDeFiBalanceService` (it already sums delegations and persists via `StakingDetailsRepository`) and drop this fold, or remove Terra/TerraClassic from the `CosmosStakingDeFiBalanceService` routing so only one path contributes. Please test on a vault with an active delegation and compare against the real staked LUNC value before merge — the manual test item is still unchecked and is exactly what would catch this.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/cosmos/CosmosStakedBalanceUseCase.kt
```diff
@@ -0,0 +1,89 @@
...
+    /** Last-known snapshot total with no network; [BigInteger.ZERO] until the first live read. */
+    fun cached(chain: Chain, address: String): BigInteger
+}
+
+@Singleton
+internal class CosmosStakedBalanceUseCaseImpl
+@Inject
+constructor(private val cosmosStakingService: CosmosStakingService) : CosmosStakedBalanceUseCase {
+
+    private val snapshots = ConcurrentHashMap<String, BigInteger>()
```
johnnyluo: This in-memory, process-lifetime snapshot is strictly weaker than what already exists. `CosmosStakingDeFiBalanceService` persists the staked total to Room via `StakingDetailsRepository`, so it survives process death and renders on the cached emit after a cold start. This map is `ZERO` until the first live read **in this process**, so on a cold start `foldInCosmosStakedBalance(live = false)` is a no-op and the cached emit still flashes `$0.00` — the very thing the snapshot is meant to prevent. Another reason to lean on the existing service rather than a parallel mechanism.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -480,6 +486,54 @@ constructor(
...
+                        decimals = account.token.decimal,
+                    )
+                // Re-price the merged amount from the unit price the liquid balance resolved; if
+                // the
+                // price never loaded, keep the existing fiat rather than zeroing it.
+                val unitPrice = account.price?.value
+                val mergedFiatValue =
+                    if (unitPrice != null) {
+                        FiatValue(
+                            value = mergedTokenValue.decimal.multiply(unitPrice),
```
johnnyluo: Fiat isn't scaled to the currency. Everywhere else in this flow fiat is computed as `decimal.multiply(price).scaledFor(currency)` (see `BalanceRepository.getDefiTokenBalanceAndPrice`), so the merged fiat here can carry more decimals than the currency convention. Also `account.fiatValue?.currency ?: AppCurrency.USD.ticker` hardcodes USD as the fallback while the rest of the flow derives currency from `appCurrencyRepository.currency` — a non-USD user with a null fiat would get a mislabeled currency. Minor, but worth aligning.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -480,6 +486,54 @@ constructor(
...
+            accounts.map { account ->
+                if (!account.token.isNativeToken) return@map account
+                val mergedTokenValue =
+                    TokenValue(
+                        value = (account.tokenValue?.value ?: BigInteger.ZERO) + stakedBaseUnits,
+                        unit = account.token.ticker,
+                        decimals = account.token.decimal,
+                    )
+                // Re-price the merged amount from the unit price the liquid balance resolved; if
+                // the
```
johnnyluo: Nit: orphaned `// the` — ktfmt comment-wrapping artifact. Reflow this comment.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/AccountsRepositoryImplTest.kt
```diff
@@ -46,13 +48,17 @@ internal class AccountsRepositoryImplTest {
         balanceRepository = mockk()
         tokenPriceRepository = mockk()
         splTokenRepository = mockk(relaxed = true)
+        cosmosStakedBalance = mockk()
+        coEvery { cosmosStakedBalance(any(), any()) } returns BigInteger.ZERO
```
johnnyluo: This mock returns `ZERO`, so no test exercises the interaction between the fold and the existing `cosmosStakingDeFiBalanceService` contribution — exactly where the double-count would surface. An end-to-end test that drives a real delegated balance through `loadDeFiAddresses` (with `getDefiTokenBalanceAndPrice` returning the staked total) would catch it.

johnnyluo (changes requested):
Implementation quality is high and the use case is well-tested in isolation, but I think this double-counts the delegated balance against the staking DeFi path that already exists on `main` (`CosmosStakingDeFiBalanceService`, #4499/#4687). Inline comments below; the top one is blocking until verified on a vault with an active delegation.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -480,6 +486,54 @@ constructor(
...
+            if (live) cosmosStakedBalance(chain, address)
+            else cosmosStakedBalance.cached(chain, address)
+        if (stakedBaseUnits <= BigInteger.ZERO) return this
+
+        val updatedAccounts =
+            accounts.map { account ->
+                if (!account.token.isNativeToken) return@map account
+                val mergedTokenValue =
+                    TokenValue(
+                        value = (account.tokenValue?.value ?: BigInteger.ZERO) + stakedBaseUnits,
```
rkokhatskyi: `foldInCosmosStakedBalance` adds the staked total onto a row that already carries the delegated total `D`, since Terra/TerraClassic route through `CosmosStakingDeFiBalanceService.getRemoteDeFiBalance` which returns the delegations sum as the native `tokenValue`, so the DeFi row and total render about `2D + R`

```kotlin
// the native row already holds the delegated total D from the existing DeFi balance path;
// drop foldInCosmosStakedBalance and instead add only claimable rewards inside
// CosmosStakingDeFiBalanceService.getRemoteDeFiBalance
```

data/src/main/kotlin/com/vultisig/wallet/data/usecases/cosmos/CosmosStakedBalanceUseCase.kt
```diff
@@ -0,0 +1,89 @@
...
+                .onFailure { Timber.w(it, "DeFi staked balance: rewards failed for %s", chain.raw) }
+                .getOrNull()
+                ?.rewards
+                ?.flatMap { it.reward }
+                ?.filter { it.denom == bondDenom }
+                ?.fold(BigInteger.ZERO) { acc, coin ->
+                    acc + (coin.amount.toBigDecimalOrNull()?.toBigInteger() ?: BigInteger.ZERO)
+                } ?: BigInteger.ZERO
+
+        val total = delegated + rewards
```
rkokhatskyi: `total = delegated + rewards` folds claimable rewards into the staked total whereas iOS `DefiBalanceService.getStakedBalances` sums `position.amount` only and keeps `rewards` separate, so the Android portfolio runs higher than iOS, and the `cosmosStakingTotalBalance` cited in the docs does not exist in iOS

data/src/main/kotlin/com/vultisig/wallet/data/usecases/cosmos/CosmosStakedBalanceUseCase.kt
rkokhatskyi: This `snapshots` map and the whole use case duplicate `CosmosStakingDeFiBalanceService`, which already fetches the delegations, persists the total to `StakingDetailsRepository` and feeds the native row, so `invoke` reruns `fetchDelegations` a second time each refresh while the only real gap in #4763 is rewards

data/src/test/kotlin/com/vultisig/wallet/data/repositories/AccountsRepositoryImplTest.kt
rkokhatskyi: The double count stays invisible to tests since `cosmosStakedBalance` is stubbed to `ZERO` here and no `loadDeFiAddresses` case covers Terra/TerraClassic, so the path where the DeFi balance already supplies `D` and the fold adds it again is never exercised

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
rkokhatskyi: `mergedTokenValue.decimal.multiply(unitPrice)` omits the `.scaledFor(currency)` rounding every sibling fiat path in `BalanceRepository` applies, so the folded LUNA/LUNC fiat keeps full `BigDecimal` scale and feeds an unrounded value into the portfolio total

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -480,6 +486,54 @@ constructor(
...
+                // price never loaded, keep the existing fiat rather than zeroing it.
+                val unitPrice = account.price?.value
+                val mergedFiatValue =
+                    if (unitPrice != null) {
+                        FiatValue(
+                            value = mergedTokenValue.decimal.multiply(unitPrice),
+                            currency = account.fiatValue?.currency ?: AppCurrency.USD.ticker,
+                        )
+                    } else {
+                        account.fiatValue
```
rkokhatskyi: When `account.price` is null the fold still bumps `tokenValue` by the staked units yet falls back to `account.fiatValue`, which the cached emit sets to `$0.00` whenever the cached price is missing, so the row shows a nonzero staked amount with zero fiat that understates the total

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Confirmed and fixed in 9a2348c8f. You're right: `getRemoteDeFiBalance` already feeds the delegated total `D` into the native row, so the fold was adding it a second time (~`2D + R`).

I dropped `foldInCosmosStakedBalance` and the whole `CosmosStakedBalanceUseCase`, and instead fold only the missing piece — claimable bond-denom rewards — into the existing `CosmosStakingDeFiBalanceService.getRemoteDeFiBalance`. That service is now the single source that fetches delegations, adds rewards, and persists the combined total via `StakingDetailsRepository`. Added `CosmosStakingDeFiBalanceServiceTest` for the math; the manual mainnet check against a real delegated LUNC value is still pending.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/cosmos/CosmosStakedBalanceUseCase.kt
aminsato: Agreed — fixed by deleting the in-memory snapshot entirely. The staked total (now `delegated + rewards`) goes through `CosmosStakingDeFiBalanceService.persistStakedBalance` → `StakingDetailsRepository` (Room), so the cached emit reads it back via `getCacheDeFiBalance` and survives process death. No more `$0.00` flash on a cold start, and no parallel mechanism.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Resolved by removing the fold — fiat is no longer recomputed here. The staked total now flows through the standard `BalanceRepository` DeFi path, which already applies `.scaledFor(currency)` and derives the currency from `appCurrencyRepository`. So there's no unscaled fiat and no hardcoded USD fallback anymore.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Gone — the whole `foldInCosmosStakedBalance` block, including that orphaned comment, is removed in 9a2348c8f.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/AccountsRepositoryImplTest.kt
aminsato: Addressed: the new `CosmosStakingDeFiBalanceServiceTest` drives a real delegated balance plus rewards through `getRemoteDeFiBalance` and asserts the combined total, that non-bond-denom rewards (e.g. legacy Terra Classic stability-tax denoms) are excluded, the rewards-failure fallback to the delegated-only total, and the persisted amount. Since there's now a single contribution path, the double-count it was masking can't recur.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/cosmos/CosmosStakedBalanceUseCase.kt
```diff
@@ -0,0 +1,89 @@
...
+                .onFailure {
+                    Timber.w(it, "DeFi staked balance: delegations failed for %s", chain.raw)
+                }
+        // Delegations are the dominant component, so a failed read keeps the last-known snapshot
+        // rather than reporting a degraded zero that would hide an active delegation on an LCD
+        // blip.
+        val delegations =
+            delegationsResult.getOrElse {
+                return cached(chain, address)
+            }
```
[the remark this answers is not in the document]
aminsato: Good catch on preserving cancellation. `CosmosStakedBalanceUseCase` is removed in 9a2348c8f (it double-counted the delegated balance — see the human threads), but the rewards fetch it performed now lives in `CosmosStakingDeFiBalanceService.claimableRewards`, which rethrows `CancellationException` before degrading to the delegated-only total. So coroutine cancellation is preserved on the surviving path.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Done exactly as suggested in 9a2348c8f — dropped `foldInCosmosStakedBalance` and added only claimable rewards inside `CosmosStakingDeFiBalanceService.getRemoteDeFiBalance`, which already held the delegated total `D`. The row and total now render `D + R` once, from a single source that also persists via `StakingDetailsRepository`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/cosmos/CosmosStakedBalanceUseCase.kt
aminsato: I kept rewards in because they're the remaining gap in #4763 — delegations already reach the row via the existing service (your other note), so rewards are the only piece that was missing. I did remove the inaccurate doc reference: the staked-total docstring no longer claims to mirror a non-existent iOS `cosmosStakingTotalBalance`.

That said, you're right that this runs higher than iOS's `getStakedBalances`, which sums `position.amount` only and keeps rewards separate. If you'd prefer to match iOS exactly — staked total = delegations only, with claimable rewards surfaced separately rather than folded into the portfolio total — I'm happy to split them. Let me know which way you want it and I'll adjust.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/cosmos/CosmosStakedBalanceUseCase.kt
aminsato: Fixed — `CosmosStakedBalanceUseCase` (and its second `fetchDelegations` call per refresh) is deleted in 9a2348c8f. Only `CosmosStakingDeFiBalanceService` fetches delegations now; it additionally fetches rewards (the actual gap in #4763) and persists the combined total. No more duplicated delegation fetch.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/AccountsRepositoryImplTest.kt
aminsato: Covered now by `CosmosStakingDeFiBalanceServiceTest`, which exercises the Terra/TerraClassic path end to end (delegations + bond-denom rewards, other denoms excluded, persisted total). With a single contribution path the `D`-from-DeFi-balance + `D`-from-fold overlap is structurally gone, so there's no longer a hidden double-count for a test to miss.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Resolved by removing the fold: the merged fiat is no longer computed here. The staked total is priced through the standard `BalanceRepository` DeFi path, which applies `.scaledFor(currency)` like every sibling fiat path — so no full-scale `BigDecimal` leaks into the portfolio total.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Also resolved by removing the fold — there's no longer any place that bumps `tokenValue` by the staked units while leaving fiat at the cached `$0.00`. The staked total goes through the normal DeFi balance path, so token amount and fiat move together (or both stay cached) rather than diverging when the price is missing.

rkokhatskyi (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/usecases/cosmos/CosmosStakedBalanceUseCase.kt
johnnyluo: Confirmed with the team — we're keeping the staked total as delegated + claimable bond-denom rewards (D + R) rather than matching iOS's delegations-only `getStakedBalances`. Rewards are the real gap in #4763, and folding them into the single source (CosmosStakingDeFiBalanceService) is the intended behavior here. The inaccurate iOS doc reference has been removed. Resolving.

johnnyluo (approved):
LGTM

PR #4791, opened 2026-06-08
fix(home): keep cached portfolio total/row while a slow chain refetches (#4768)

### LINKED ISSUE #4768: Home portfolio: slow chain blanks the total balance (and its row) — keep cached value until new one resolves
## Observed
On the home/Wallet portfolio, when one or more chains are slow to fetch (e.g. Solana), that chain's row **and the total balance** go blank instead of showing the previously-known value until the fresh balance arrives. (Screenshot: Solana row + the big total are blank while ETH/Dash/Maya/BTC show.)

Desired: use the **old/cached** balance and don't disregard it until the new value is fetched.

## Root cause (total)
`data/src/main/kotlin/com/vultisig/wallet/data/models/Account.kt:24-36`:
```kotlin
fun List<Account>.calculateAccountsTotalFiatValue(): FiatValue? =
    … val fiatValue = account.fiatValue ?: return … null      // any null account ⇒ whole total null
fun List<Address>.calculateAddressesTotalFiatValue(): FiatValue? =
    … account.accounts.calculateAccountsTotalFiatValue() ?: return … null
```
Any single chain with a null `fiatValue` makes the **entire** total null → `totalFiatValue` (`VaultAccountsViewModel.kt:88`) is null → blank total. So one slow/unresolved chain blanks the whole balance.

## Per-chain row
The repo already tries to show cached-first (`AccountsRepository.loadAddressBalances:100-104`, "Always emit the last-known DB snapshot first"), but a chain's row still goes blank when its `fiatValue` is unresolved — the per-chain fiat is gated on `loadPrices.await()` + `getTokenBalanceAndPrice` (`AccountsRepository:111-127`), and there's no retain-last-value during refetch.

## Suggested fix
- Total: sum the **available** chain values instead of returning null when any chain is pending (or retain the last computed total until all resolve). Don't let one pending chain zero the total.
- Per-chain: keep the last-known balance/fiat for a chain while its refetch is in flight (don't reset to null/blank).

## Cross-platform note
The repo comment (`AccountsRepository:136-138`) says iOS **batches** all balances and applies at once, while Android/Windows stream per-chain — which is why Android shows the partial-blank state. Worth aligning behavior.

## Platform
Android.

aminsato:
## Summary
On the home/Wallet portfolio, when one chain is slow to fetch (e.g. Solana), that chain's row **and the big total balance** went blank instead of showing the previously-known value until the fresh balance arrived. Fixes #4768.

### Root cause
`calculateAddressesTotalFiatValue()` folded over every account and returned `null` the moment *one* account had an unresolved `fiatValue` — so a single pending chain zeroed the **entire** total. Per-chain rows had no retain-last logic either, so a refetching chain blanked.

### Changes
- **`Account.kt`** — `calculateAddressesTotalFiatValue()` now sums the **resolved** accounts and returns `null` only when *nothing* has loaded (cold start still shows a loading state). One pending chain no longer zeroes the total. The strict per-chain helper (`calculateAccountsTotalFiatValue`) is left as-is so a pending row reports "not loaded" and can fall back to its cached value.
- **`VaultAccountsViewModel.kt`** — retains the last-known **total** (`?: previous`) and last-known **per-chain** fiat/native amounts while a refetch is in flight (a non-null fresh value always wins). Applies to both wallet and DeFi lists.

### Cross-platform note
iOS *batches* all balances and applies at once; Android/Windows stream per-chain. This keeps Android's streaming but makes it resilient to partial state rather than switching to batching — lower-risk and matches the "keep cached value until new one resolves" ask.

## Test plan
- `CalculateTotalFiatValueTest` (new, data): sums resolved chains, ignores a pending chain, counts only resolved tokens within a partial chain, `null` only when nothing resolved / empty list.
- `VaultAccountsViewModelTest` (added 2 tests): Solana row keeps `$5` across a refetch returning null fiat while ETH updates; total reads `$10` when one chain is pending.
- `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest` green; ktfmt clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Retain per-chain last-known balances across partial refreshes so rows and headline totals no longer blank unexpectedly.
  * Clear previous vault totals and rows when switching vaults to prevent carryover.

* **New Features**
  * Partial-total behavior: portfolio and row totals sum only resolved fiat values (returns null only when nothing is resolved), ensuring headline matches row contributions.

* **Tests**
  * Added regression tests covering retention, partial totals, multi-token partial sums, and vault-switch clearing.

* **Documentation**
  * Clarified strict vs. partial total semantics.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -9,6 +9,7 @@ import androidx.lifecycle.viewModelScope
 import androidx.navigation.toRoute
 import com.vultisig.wallet.data.IoDispatcher
 import com.vultisig.wallet.data.blockchain.TierRemoteNFTService
+import com.vultisig.wallet.data.models.Account
 import com.vultisig.wallet.data.models.Address
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coins
@@ -17,7 +18,7 @@ import com.vultisig.wallet.data.models.SigningLibType
 import com.vultisig.wallet.data.models.Vault
 import com.vultisig.wallet.data.models.VaultId
 import com.vultisig.wallet.data.models.calculateAccountsTotalFiatValue
-import com.vultisig.wallet.data.models.calculateAddressesTotalFiatValue
+import com.vultisig.wallet.data.models.calculateAddressesPartialFiatValue
 import com.vultisig.wallet.data.models.isFastVault
 import com.vultisig.wallet.data.models.isSecureVault
 import com.vultisig.wallet.data.models.isSwapSupported
@@ -162,6 +163,12 @@ constructor(
     private var loadDeFiBalancesJob: Job? = null
     private var buyVultBannerJob: Job? = null
 
+    // Last merged snapshot per stream (wallet / DeFi), keyed in-memory so a chain that comes back
+    // with a null balance mid-refetch can carry its previously-shown value forward (#4768). Reset
+    // on vault switch so one vault's balances never leak into the next.
+    private var lastWalletAddresses: List<Address> = emptyList()
+    private var lastDeFiAddresses: List<Address> = emptyList()
+
     // In-memory "dismissed for this visit" flag. When the vault holds less than the Silver-tier
     // VULT amount, closing the Buy VULT banner only flips this (no persistence), so it returns the
     // next time the home screen is entered or the vault is switched.
@@ -228,6 +235,18 @@ constructor(
         if (vaultChanged) {
             // Don't carry a per-visit Buy VULT dismissal across to the next vault.
             buyVultSessionDismissed.value = false
+            // Drop the previous vault's balances so the retain/partial-total logic can't surface
+            // one vault's totals or rows under the next (#4768).
+            lastWalletAddresses = emptyList()
+            lastDeFiAddresses = emptyList()
+            uiState.update {
+                it.copy(
+                    accounts = emptyList(),
+                    defiAccounts = emptyList(),
+                    totalFiatValue = null,
+                    totalDeFiValue = null,
+                )
+            }
         }
         collectBuyVultBannerDismissed(vaultId)
         loadVaultNameAndShowBackup(vaultId)
@@ -536,27 +555,25 @@ constructor(
         searchQuery: String,
         isDefi: Boolean = false,
     ) {
+        // Retain at the Address level, then derive both the total and the rows from the same merged
+        // list. The headline figure then always equals the sum its rows render (#4768), rather than
+        // the total counting only freshly-resolved chains while a row still shows a cached one.
+        val previous = if (isDefi) lastDeFiAddresses else lastWalletAddresses
+        val merged = this.retainLastKnownBalances(previous)
+        if (isDefi) lastDeFiAddresses = merged else lastWalletAddresses = merged
+
         val totalFiatValue =
-            this.calculateAddressesTotalFiatValue()?.let { fiatValueToStringMapper(it) }
-        val accountsUiModel = this.map { addressToUiModelMapper(it) }
+            merged.calculateAddressesPartialFiatValue()?.let { fiatValueToStringMapper(it) }
+        val accountsUiModel =
+            merged.map { addressToUiModelMapper(it) }.filteredAccounts(searchQuery)
 
         uiState.update { state ->
-            val previousAccounts = if (isDefi) state.defiAccounts else state.accounts
-            val mergedAccounts =
-                accountsUiModel
-                    .retainLastKnownValues(previousAccounts)
-                    .filteredAccounts(searchQuery)
             if (!isDefi) {
-                state.copy(
-                    // Don't blank an already-shown total while a slow chain refetches (#4768).
-                    totalFiatValue = totalFiatValue ?: state.totalFiatValue,
-                    accounts = mergedAccounts,
-                )
+                state.copy(totalFiatValue = totalFiatValue, accounts = accountsUiModel)
             } else {
-                state.copy(
-                    totalDeFiValue = totalFiatValue ?: state.totalDeFiValue,
-                    defiAccounts = mergedAccounts,
-                )
+                // An empty merged list (e.g. every DeFi chain deselected) yields a null total here,
+                // clearing the header instead of stranding a stale figure over zero rows (#4768).
+                state.copy(totalDeFiValue = totalFiatValue, defiAccounts = accountsUiModel)
             }
         }
 
@@ -564,23 +581,37 @@ constructor(
     }
 
     /**
-     * While a chain's balance is refetching its mapped fiat/native amounts come back null. Carry
-     * the last-known value for that chain forward so its row keeps showing the cached balance
-     * instead of blanking until the fresh value resolves (#4768). A non-null new value always wins.
+     * While a chain's balance is refetching, its mapped fiat/native values come back null. Carry
+     * the last-known [Account] values for that row forward so both the row and the portfolio total
+     * keep showing the cached balance instead of blanking until the fresh value resolves (#4768). A
+     * non-null fresh value always wins.
+     *
+     * Keyed by chain + address + provider flag, not chain alone, so two DeFi rows that share a
+     * chain (e.g. a Circle position and another Ethereum position) never carry each other's value
+     * forward.
      */
-    private fun List<AccountUiModel>.retainLastKnownValues(
-        previous: List<AccountUiModel>
-    ): List<AccountUiModel> {
+    private fun List<Address>.retainLastKnownBalances(previous: List<Address>): List<Address> {
         if (previous.isEmpty()) return this
-        val previousByChain = previous.associateBy { it.model.chain }
+        val previousByKey = previous.associateBy { it.retainKey() }
+        return map { address ->
+            val last = previousByKey[address.retainKey()] ?: return@map address
+            address.copy(accounts = address.accounts.retainMissing(last.accounts))
+        }
+    }
+
+    private fun List<Account>.retainMissing(previous: List<Account>): List<Account> {
+        val previousByToken = previous.associateBy { it.token.id }
         return map { account ->
-            val last = previousByChain[account.model.chain] ?: return@map account
+            val last = previousByToken[account.token.id] ?: return@map account
             account.copy(
-                fiatAmount = account.fiatAmount ?: last.fiatAmount,
-                nativeTokenAmount = account.nativeTokenAmount ?: last.nativeTokenAmount,
+                tokenValue = account.tokenValue ?: last.tokenValue,
+                fiatValue = account.fiatValue ?: last.fiatValue,
+                price = account.price ?: last.price,
             )
         }
     }
+
+    private fun Address.retainKey(): String = "${chain.id}|$address|$isDefiProvider"
 
     private fun List<AccountUiModel>.filteredAccounts(searchQuery: String): List<AccountUiModel> {
         if (searchQuery.isBlank()) return this
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -540,23 +540,46 @@ constructor(
             this.calculateAddressesTotalFiatValue()?.let { fiatValueToStringMapper(it) }
         val accountsUiModel = this.map { addressToUiModelMapper(it) }
 
-        if (!isDefi) {
-            uiState.update {
-                it.copy(
-                    totalFiatValue = totalFiatValue,
-                    accounts = accountsUiModel.filteredAccounts(searchQuery),
+        uiState.update { state ->
+            val previousAccounts = if (isDefi) state.defiAccounts else state.accounts
+            val mergedAccounts =
+                accountsUiModel
+                    .retainLastKnownValues(previousAccounts)
+                    .filteredAccounts(searchQuery)
+            if (!isDefi) {
+                state.copy(
+                    // Don't blank an already-shown total while a slow chain refetches (#4768).
+                    totalFiatValue = totalFiatValue ?: state.totalFiatValue,
+                    accounts = mergedAccounts,
                 )
-            }
-        } else {
-            uiState.update {
-                it.copy(
-                    totalDeFiValue = totalFiatValue,
-                    defiAccounts = accountsUiModel.filteredAccounts(searchQuery),
+            } else {
+                state.copy(
+                    totalDeFiValue = totalFiatValue ?: state.totalDeFiValue,
+                    defiAccounts = mergedAccounts,
                 )
             }
         }
 
         Timber.d("Update updateUiStateFromList: %s", "$this")
+    }
+
+    /**
+     * While a chain's balance is refetching its mapped fiat/native amounts come back null. Carry
+     * the last-known value for that chain forward so its row keeps showing the cached balance
+     * instead of blanking until the fresh value resolves (#4768). A non-null new value always wins.
+     */
+    private fun List<AccountUiModel>.retainLastKnownValues(
+        previous: List<AccountUiModel>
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -540,25 +540,48 @@ constructor(
...
+        uiState.update { state ->
+            val previousAccounts = if (isDefi) state.defiAccounts else state.accounts
+            val mergedAccounts =
+                accountsUiModel
+                    .retainLastKnownValues(previousAccounts)
+                    .filteredAccounts(searchQuery)
+            if (!isDefi) {
+                state.copy(
+                    // Don't blank an already-shown total while a slow chain refetches (#4768).
+                    totalFiatValue = totalFiatValue ?: state.totalFiatValue,
```
johnnyluo: **Cross-vault leakage on vault switch.** `loadData()` doesn't reset `accounts`/`totalFiatValue` when the vault changes, and this fallback now actively preserves the prior vault's values. If the new vault's first emission resolves nothing (cold start), `totalFiatValue ?: state.totalFiatValue` keeps showing **the previous vault's total** (before this PR it went to `null`/loading). Likewise `retainLastKnownValues` is keyed by chain, so a new vault's pending Ethereum row shows the *old* vault's ETH balance. Suggest gating the retain/fallback to the same vault — e.g. clear `accounts`/`totalFiatValue`/`defiAccounts`/`totalDeFiValue` when `vaultChanged` in `loadData`, or thread the vaultId through so retain only applies within the same vault. Worth a regression test (switch vault, assert old total/row doesn't leak).

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -540,25 +540,48 @@ constructor(
...
+    /**
+     * While a chain's balance is refetching its mapped fiat/native amounts come back null. Carry
+     * the last-known value for that chain forward so its row keeps showing the cached balance
+     * instead of blanking until the fresh value resolves (#4768). A non-null new value always wins.
+     */
+    private fun List<AccountUiModel>.retainLastKnownValues(
+        previous: List<AccountUiModel>
+    ): List<AccountUiModel> {
+        if (previous.isEmpty()) return this
+        val previousByChain = previous.associateBy { it.model.chain }
```
johnnyluo: **Key collision on the DeFi list.** `associateBy { it.model.chain }` collapses rows that share a chain (keeps the last). The wallet list is one row per chain, but the DeFi list can have multiple providers on the same chain (e.g. Circle + other Ethereum DeFi rows). Since `retainLastKnownValues` runs on both lists, a DeFi row could carry forward another provider's cached value. Consider keying by something more specific — `model.address` (+ `isDeFiProvider`) rather than `chain` alone.

data/src/main/kotlin/com/vultisig/wallet/data/models/Account.kt
```diff
@@ -28,11 +28,17 @@ fun List<Account>.calculateAccountsTotalFiatValue(): FiatValue? =
...
+ * Sum of every account's fiat value across all addresses, skipping the ones that haven't resolved
+ * yet. Returns null only when *nothing* has loaded (so the UI can show a loading state on a cold
+ * start), but as soon as a single chain resolves its value contributes to the total. A slow chain
+ * (e.g. Solana) therefore no longer blanks the whole portfolio total — the previously-known chains
+ * keep counting while it refetches. See issue #4768.
+ */
+fun List<Address>.calculateAddressesTotalFiatValue(): FiatValue? {
+    val resolved = this.flatMap { it.accounts }.mapNotNull { it.fiatValue }
+    if (resolved.isEmpty()) return null
+    return resolved.fold(FiatValue(BigDecimal.ZERO, resolved.first().currency)) { acc, fiatValue ->
```
johnnyluo, line 71: Minor: the fold seed currency changed from `AppCurrency.USD.ticker` to `resolved.first().currency`. Functionally better (reflects the real currency), and mixed-currency lists were already unsupported by the old fold — just calling it out as a behavioral change worth a note.

johnnyluo (commented):
Inline notes from review — main concern is cross-vault leakage on vault switch; DeFi key collision and the currency change are secondary.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
rkokhatskyi: The headline set here sums only resolved tokens via `calculateAddressesTotalFiatValue` while `retainLastKnownValues` keeps a refetching chain's prior value in its row, so the total understates the visible rows, the PR tests pinning total `$10` while the Solana row reads `$5`, whereas iOS and Windows sum the same values the rows render so the two never disagree

```kotlin
val merged = this.retainLastKnownAccounts(previous)
val total = merged.calculateAddressesTotalFiatValue()?.let { fiatValueToStringMapper(it) }
val rows = merged.map { addressToUiModelMapper(it) }
```

data/src/main/kotlin/com/vultisig/wallet/data/models/Account.kt
```diff
@@ -28,11 +28,17 @@ fun List<Account>.calculateAccountsTotalFiatValue(): FiatValue? =
...
-            account.accounts.calculateAccountsTotalFiatValue()
-                ?: return@calculateAddressesTotalFiatValue null
+/**
+ * Sum of every account's fiat value across all addresses, skipping the ones that haven't resolved
+ * yet. Returns null only when *nothing* has loaded (so the UI can show a loading state on a cold
+ * start), but as soon as a single chain resolves its value contributes to the total. A slow chain
+ * (e.g. Solana) therefore no longer blanks the whole portfolio total — the previously-known chains
+ * keep counting while it refetches. See issue #4768.
+ */
+fun List<Address>.calculateAddressesTotalFiatValue(): FiatValue? {
```
rkokhatskyi: Flipping this from null to a partial sum also changes `VaultAndBalanceUseCase.kt:30` which reads only `firstOrNull()` of the cached snapshot and `ConfirmDeleteViewModel.kt:93` which renders it directly, so the vault list and delete screen now show an understated total as final rather than waiting for a full load

```kotlin
// keep this strict; add a separate calculateAddressesPartialFiatValue() for the home stream only
```

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -540,25 +540,48 @@ constructor(
...
                 )
-            }
-        } else {
-            uiState.update {
-                it.copy(
-                    totalDeFiValue = totalFiatValue,
-                    defiAccounts = accountsUiModel.filteredAccounts(searchQuery),
+            } else {
+                state.copy(
+                    totalDeFiValue = totalFiatValue ?: state.totalDeFiValue,
```
rkokhatskyi: An empty list, such as deselecting every DeFi chain, makes `calculateAddressesTotalFiatValue` null so this keeps the previous `totalDeFiValue` while `defiAccounts` becomes empty, stranding a stale header total over zero rows that never clears since nothing resolves to overwrite it

```kotlin
totalDeFiValue = if (this.isEmpty()) null else totalFiatValue ?: state.totalDeFiValue
```

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Fixed in bc6d46d38. `loadData` now clears the retained snapshots (`lastWalletAddresses` / `lastDeFiAddresses`) and resets `accounts` / `defiAccounts` / `totalFiatValue` / `totalDeFiValue` when `vaultChanged`, so the new vault starts clean — a cold first emission can't carry the prior vault's total or row forward. Added a `switching vault clears the previous vault total and rows` regression test (vault-2's first emission resolves nothing → total is null and no `$10` row leaks from vault-1).

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Fixed — retain now keys by `chain + address + isDefiProvider` (`Address.retainKey()`), not chain alone, so two DeFi rows that share a chain (e.g. a Circle position and another Ethereum position) never carry each other's value forward. The retain also moved to the Address level, so this is matched per address/provider before mapping to UI rows.

data/src/main/kotlin/com/vultisig/wallet/data/models/Account.kt
aminsato: Noted, thanks for calling it out. The partial-sum logic now lives in a separate `calculateAddressesPartialFiatValue`, which keeps the `resolved.first().currency` seed (so a non-USD portfolio is labelled correctly); the original `calculateAddressesTotalFiatValue` is back to its USD-seed strict behavior. Mixed-currency lists were unsupported by either fold, as you say — documented that in the new function's KDoc.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Fixed in bc6d46d38, as you outlined. Retain now happens at the Address level, and both the total and the rows derive from the same `merged` list — so the headline always equals the sum its rows render. Added `total stays consistent with the rows while a chain refetches`: Solana resolves to `$5`, then refetches to null; it stays `$5` in both the row and the total, so the headline stays `$15` (10 + 5) instead of dropping to `$10`.

data/src/main/kotlin/com/vultisig/wallet/data/models/Account.kt
aminsato: Done — kept `calculateAddressesTotalFiatValue` strict (null until every chain loads) so `VaultAndBalanceUseCase` (vault list) and `ConfirmDeleteViewModel` (delete screen) keep showing a final all-or-nothing figure, and added a separate `calculateAddressesPartialFiatValue()` used only by the home stream. Those two screens no longer render an understated partial total as if it were final.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Fixed — because the total now derives from the same `merged` list as the rows, deselecting every DeFi chain produces an empty list → `calculateAddressesPartialFiatValue` returns null → `totalDeFiValue` clears. No more stale header total stranded over zero rows. (The retain runs over the fresh list, so deselected chains aren't resurrected either.)

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -536,29 +555,64 @@ constructor(
...
+        val previousByToken = previous.associateBy { it.token.id }
+        return map { account ->
+            val last = previousByToken[account.token.id] ?: return@map account
+            account.copy(
+                tokenValue = account.tokenValue ?: last.tokenValue,
+                fiatValue = account.fiatValue ?: last.fiatValue,
+                price = account.price ?: last.price,
+            )
+        }
+    }
```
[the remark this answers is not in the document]
aminsato: Good catch — applied your guard in f798ed23a. `retainMissing` now returns the previous accounts when the current list is empty, so an empty re-emission keeps the cached row/total for that cycle instead of blanking. Added a `row keeps cached value when an address re-emits with empty accounts` regression test. (The deselect-all-DeFi-chains case is handled separately at the address-list level, so this guard doesn't resurrect deselected chains.)

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -536,29 +555,64 @@ constructor(
...
+        // list. The headline figure then always equals the sum its rows render (#4768), rather than
+        // the total counting only freshly-resolved chains while a row still shows a cached one.
+        val previous = if (isDefi) lastDeFiAddresses else lastWalletAddresses
+        val merged = this.retainLastKnownBalances(previous)
+        if (isDefi) lastDeFiAddresses = merged else lastWalletAddresses = merged
+
         val totalFiatValue =
-            this.calculateAddressesTotalFiatValue()?.let { fiatValueToStringMapper(it) }
-        val accountsUiModel = this.map { addressToUiModelMapper(it) }
+            merged.calculateAddressesPartialFiatValue()?.let { fiatValueToStringMapper(it) }
```
rkokhatskyi, line 566: `calculateAddressesPartialFiatValue` counts each resolved token while the row mapper at `AddressToUiModelMapper.kt:31` stays strict, so a chain holding several tokens with one still pending feeds the total yet renders a blank row, whereas iOS and Windows fold each row leniently so a missing rate counts as zero and the row never blanks

```kotlin
// make the row fold lenient too in AddressToUiModelMapper
fiatAmount = from.accounts.mapNotNull { it.fiatValue }.takeIf { it.isNotEmpty() }?.reduce { a, b -> FiatValue(a.value + b.value, a.currency) }?.let { fiatValueToStringMapper(it) }
```

app/src/test/java/com/vultisig/wallet/ui/models/home/VaultAccountsViewModelTest.kt
```diff
@@ -582,6 +583,184 @@ internal class VaultAccountsViewModelTest {
...
+
+    /**
+     * Regression for #4768: switching vaults must not leak the previous vault's total or rows. When
+     * vault-2's first emission resolves nothing, the total/row must reset rather than carrying
+     * vault-1's $10 forward via the retain logic.
+     */
+    @Test
+    fun `switching vault clears the previous vault total and rows`() =
+        runTest(testDispatcher) {
+            val vault1Eth = buildTestAddress(Chain.Ethereum, "0xeth1", fiat = BigDecimal("10"))
```
rkokhatskyi: This guard uses two different addresses `0xeth1` and `0xeth2` so their `retainKey` never collides, which means it still passes with the vault reset at lines 240 to 249 removed and never exercises the leak it is named for, the fix being to give both vaults the same chain and address so the retain path fires

```kotlin
val vault1Eth = buildTestAddress(Chain.Ethereum, "0xeth", fiat = BigDecimal("10"))
val vault2EthPending = buildTestAddress(Chain.Ethereum, "0xeth", fiat = null)
```

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Fixed in f6f5135bc. You're right — the row mapper used the strict `calculateAccountsTotalFiatValue` while the total used the lenient partial fold, so a multi-token chain with one pending token fed the headline yet rendered a blank row. Added `calculateAccountsPartialFiatValue()` (a lenient per-address fold) and pointed `AddressToUiModelMapper` at it, so the row now sums its resolved tokens and equals exactly its contribution to the partial total — matching iOS/Windows. `calculateAddressesPartialFiatValue()` now delegates to it so the two stay consistent by construction, and the strict folds (sorting, vault list, delete screen) are untouched. Added a `multi-token row shows resolved sum while one token is pending` regression test.

app/src/test/java/com/vultisig/wallet/ui/models/home/VaultAccountsViewModelTest.kt
aminsato: Good catch — fixed in f6f5135bc. Both vaults now use the same `0xeth` (instead of `0xeth1`/`0xeth2`) so their `retainKey` (`chain|address|isDefiProvider`) collides and the retain path actually fires. The test now genuinely exercises the leak: with the vault-change reset removed it would carry vault-1's $10 forward and fail, so only the reset keeps the assertion green.

johnnyluo (approved):
LGTM

PR #4792, opened 2026-06-08
fix(swap): granular "trading temporarily halted" error (#4749)

### LINKED ISSUE #4749: Swap: show granular "Trading temporarily halted" instead of generic "Swap route not available"
## What
When a swap pair fails because the destination chain is **halted upstream** (THORChain/MAYAChain trading pause), the form shows the same generic "Swap route not available" as a genuinely unsupported pair. Users can't tell *temporarily paused, retry later* from *this pair will never work*.

## Repro
ETH → ZEC today: MAYAChain has ZEC trading paused (`inbound_addresses` ZEC `halted=true, chain_trading_paused=true`; Maya quote returns `"failed to simulate swap: trading is halted, can't process swap"`). App shows "Swap route not available."

## Where (Android)
- `data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapException.kt:62` — `contains("trading is halted") -> SwapRouteNotAvailable(error)` collapses the halt into the generic route error.
- `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt:879` — `SwapRouteNotAvailable -> R.string.swap_route_not_available`.

## Suggested fix
Add a distinct `SwapException.TradingHalted` (mapped from `"trading is halted"` / `"trading halted"`), surface a new localized string e.g. *"Trading for this asset is temporarily halted. Try again later."*, and rank it as transient (it's recoverable) in `swapFailurePriority`.

## Notes
- Enhancement / copy-clarity, not a funds bug. Root cause is upstream halt; app behaviour is otherwise correct.
- Cross-platform: companion issues filed on iOS + Windows.
- Platform: Android.

aminsato:
## What

When a swap pair fails because the destination chain is **halted upstream** (THORChain/MAYAChain trading pause), the form showed the same generic *"Swap route not available"* as a genuinely unsupported pair. Users couldn't tell *temporarily paused, retry later* from *this pair will never work*.

Closes #4749.

## Changes

- **`SwapException`**: new `TradingHalted` type; `"trading is halted"` (and the `"trading halted"` variant) now map to it instead of `SwapRouteNotAvailable`.
- **`SwapQuoteManager`**: maps `TradingHalted` to a new localized string, and ranks it as **transient/recoverable** in `swapFailurePriority` (same bucket as timeout/rate-limit) so the informative halt message wins over a generic route error when providers disagree.
- **Strings**: added `swap_error_trading_halted` — *"Trading for this asset is temporarily halted. Try again later."* — across all 10 locales, using each locale's established retry phrasing.
- **Tests**: updated the now-incorrect `trading halted → SwapRouteNotAvailable` case and added coverage for the variant plus the real-world Maya message `failed to simulate swap: trading is halted, can't process swap`.

## Repro (before)

ETH → ZEC while MAYAChain has ZEC trading paused → app showed "Swap route not available." Now shows the temporary-halt message.

## Test plan

- [x] `:data:testDebugUnitTest --tests HandleSwapExceptionTest` passes
- [x] `:data` and `:app` compile clean
- [x] ktfmt applied
- [ ] Manual: trigger a halted pair (e.g. ETH → ZEC) and confirm the new copy appears

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap error handling so temporary trading halts are detected and shown as a specific halt message (preserving server-provided halt reasons when present) instead of a generic fallback, giving users clearer guidance to try again later.

* **Localization**
  * Added localized "trading halted" messages across supported languages so the halt notice appears in the user's language.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapException.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapException.kt
@@ -21,6 +21,9 @@ sealed class SwapException(message: String) : Exception(message) {
     class SwapRouteNotAvailable(message: String) : SwapException(message)
 
     @Suppress("SerialVersionUIDInSerializableClass")
+    class TradingHalted(message: String) : SwapException(message)
+
+    @Suppress("SerialVersionUIDInSerializableClass")
     class TimeOut(message: String) : SwapException(message)
 
     @Suppress("SerialVersionUIDInSerializableClass")
@@ -59,7 +62,8 @@ sealed class SwapException(message: String) : Exception(message) {
                     contains("amount less than dust threshold") -> AmountBelowDustThreshold(error)
                     contains("amount less than min swap amount") -> SmallSwapAmount(error)
                     contains("pool does not exist") -> SwapRouteNotAvailable(error)
-                    contains("trading is halted") -> SwapRouteNotAvailable(error)
+                    contains("trading is halted") || contains("trading halted") ->
+                        TradingHalted(error)
                     contains("timeout") -> TimeOut(error)
                     contains("unable to resolve host") -> NetworkConnection(error)
                     contains("no internet connection") -> NetworkConnection(error)
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/errors/HandleSwapExceptionTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/errors/HandleSwapExceptionTest.kt
@@ -69,9 +69,24 @@ class HandleSwapExceptionTest {
     }
 
     @Test
-    fun `trading halted maps to SwapRouteNotAvailable`() {
+    fun `trading is halted maps to TradingHalted`() {
         val result = SwapException.handleSwapException("Trading is halted for this pool")
-        assertInstanceOf(SwapException.SwapRouteNotAvailable::class.java, result)
+        assertInstanceOf(SwapException.TradingHalted::class.java, result)
+    }
+
+    @Test
+    fun `trading halted variant maps to TradingHalted`() {
+        val result = SwapException.handleSwapException("asset trading halted upstream")
+        assertInstanceOf(SwapException.TradingHalted::class.java, result)
+    }
+
+    @Test
+    fun `maya simulate halt message maps to TradingHalted`() {
+        val result =
+            SwapException.handleSwapException(
+                "failed to simulate swap: trading is halted, can't process swap"
+            )
+        assertInstanceOf(SwapException.TradingHalted::class.java, result)
     }
 
     @Test
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -426,6 +426,7 @@ constructor(
             is SwapException.InsufficientFunds,
             is SwapException.HighPriceImpact,
             is SwapException.RateLimitExceeded,
+            is SwapException.TradingHalted,
             is SwapException.TimeOut,
             is TimeoutCancellationException,
             is SwapException.NetworkConnection,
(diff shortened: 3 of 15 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapException.kt
```diff
@@ -59,7 +62,8 @@ sealed class SwapException(message: String) : Exception(message) {
                     contains("amount less than dust threshold") -> AmountBelowDustThreshold(error)
                     contains("amount less than min swap amount") -> SmallSwapAmount(error)
                     contains("pool does not exist") -> SwapRouteNotAvailable(error)
-                    contains("trading is halted") -> SwapRouteNotAvailable(error)
+                    contains("trading is halted") || contains("trading halted") ->
```
rkokhatskyi, line 65: `TradingHalted` never fires for MAYAChain halts returned with an error status since `MayaChainApi.kt:208` drops the response body for the generic `HttpStatusCode.description` so the halt text never reaches `handleSwapException` and falls through to the generic `swap_error_quote_failed`

```kotlin
val responseRawString = response.bodyAsText()
return runCatching {
    json.decodeFromString(thorChainSwapQuoteResponseJsonSerializer, responseRawString)
}.getOrElse {
    THORChainSwapQuoteDeserialized.Error(
        THORChainSwapQuoteError(
            responseRawString.ifBlank { HttpStatusCode.fromValue(response.status.value).description }
        )
    )
}
```

data/src/test/kotlin/com/vultisig/wallet/data/api/errors/HandleSwapExceptionTest.kt
```diff
@@ -69,9 +69,24 @@ class HandleSwapExceptionTest {
...
+    }
+
+    @Test
+    fun `trading halted variant maps to TradingHalted`() {
+        val result = SwapException.handleSwapException("asset trading halted upstream")
+        assertInstanceOf(SwapException.TradingHalted::class.java, result)
+    }
+
+    @Test
+    fun `maya simulate halt message maps to TradingHalted`() {
```
rkokhatskyi, line 84: This test feeds the literal Maya halt message straight into `handleSwapException` but production never delivers it for Maya since `MayaChainApi.kt:208` discards the body on an error status so it stays green while the ETH to ZEC repro still surfaces the generic error

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapException.kt
aminsato: Good catch, confirmed. `MayaChainApi.getSwapQuotes` short-circuited on `!isSuccess()` and replaced the body with the generic `HttpStatusCode.description`, so the halt reason never reached `handleSwapException` (the ETH→ZEC repro).

Fixed in 21482a245: it now mirrors the THORChain path — always `bodyAsText()`, deserialize through the quote serializer (which maps the error body to `THORChainSwapQuoteDeserialized.Error`), and fall back to the status description only when the body is blank. `unwrapOrThrow()` then routes `Error.message` into `handleSwapException`, so a real MAYA halt now classifies as `TradingHalted`.

data/src/test/kotlin/com/vultisig/wallet/data/api/errors/HandleSwapExceptionTest.kt
aminsato: Agreed — that test only exercised the string mapping, not the MAYA delivery path. Now that `getSwapQuotes` preserves the body (21482a245), I added real coverage in `MayaChainApiBodyReadTest`: the stale case that pinned the old "status description without reading body" behavior is re-purposed into a non-2xx-with-halt-body test asserting the reason is preserved, plus a blank-body fallback case. The `handleSwapException` mapping tests stay as-is.

johnnyluo (approved):
LGTM

PR #4793, opened 2026-06-08
fix(qbtc): use the shared co-sign screen for claim pairing

rkokhatskyi:
The QBTC claim co-sign pairing screen rendered its own QR layout, which broke: the QR came out tiny inside a wide white strip, and the copy only ever mentioned "your other device" even though a Secure Vault co-sign can need more than one.

The pairing here is just a regular keysign co-sign, so this swaps the bespoke layout for the shared `PeerDiscoveryScreen` that keygen and keysign already use. That fixes the QR sizing for free and gives proper multi-device feedback — "Devices (1/2)", per-device tiles, and threshold awareness. The QBTC-specific signing/orchestration is untouched.

## Before / After

| Before | After |
|--------|-------|
| before | after |

### Notes
- Removes the now-unused `qbtc_claim_pair_instruction` string from all 10 locale files.
- Draft — verifying the full claim flow on device first.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated QBTC claim pairing preview and replaced the prior pairing UI with a new claim-pairing flow integrated into peer discovery.
  * Peer discovery screen now exposes toggles to show/hide share, network switch, and Next button; pairing state surfaces local device ID and required co-signer count.

* **Localization**
  * Updated QBTC claim copy across locales: removed the old pairing instruction and added success and vault-password strings.


app/src/main/java/com/vultisig/wallet/ui/screens/qbtc/QbtcClaimScreen.kt
```diff
@@ -487,40 +493,35 @@ private fun ClaimSigningProgress(label: String, @DrawableRes logoRes: Int) {
...
+                qr = state.qr,
+                network = NetworkOption.Internet,
+                localPartyId = state.localPartyId,
+                devices = state.joinedDevices,
+                selectedDevices = state.joinedDevices,
+                minimumDevices = state.minimumDevices,
+                minimumDevicesDisplayed = state.minimumDevices,
+                // deviceCount <= 3 hides the "Next" button: the claim auto-advances once the peer
+                // threshold is met, just like 2/2 & 3/3 keygen.
+                deviceCount = state.minimumDevices,
```
aminsato: `deviceCount = state.minimumDevices` shows a dead "Next" button for vaults with **5+ signers**.

`minimumDevices` is `threshold = ceil(signers * 2/3)` (`QbtcClaimViewModel.pairAndKickoff`). The Next button in `PeerDiscoveryScreen` is rendered when `deviceCount > 3`, so:
- 2–4 signers → threshold ≤ 3 → button hidden ✅ (matches the comment's intent)
- 5+ signers → threshold ≥ 4 → button **shown**, but `onNextClick = {}` is a no-op

The QBTC co-sign always auto-advances (`pairAndKickoff` returns once `joined >= threshold-1`), which happens at the exact moment `hasEnoughDevices` would enable the button — so the user only ever sees a disabled "waiting for devices" button that the design intends to hide entirely (like 2/2 & 3/3 keygen).

Since the co-sign never wants a Next button, gate it explicitly instead of piggy-backing on `deviceCount`:

```suggestion
                // QBTC co-sign always auto-advances once the peer threshold is met (like 2/2 &
                // 3/3 keygen), so the "Next" button is never wanted. deviceCount <= 3 hides it.
                deviceCount = state.minimumDevices.coerceAtMost(3),
```

(Or add a `showNext: Boolean = true` param to `PeerDiscoveryScreen`, mirroring the new `showShare`/`showNetworkSwitch`, and pass `false` here — cleaner than overloading `deviceCount`.)

rkokhatskyi: Fixed in 7b9a8d183 by adding an explicit showNext: Boolean = true to PeerDiscoveryScreen (mirroring showShare/showNetworkSwitch) and passing showNext = false here, dropping the deviceCount piggy-back so the button is hidden for every vault size rather than just <=3 which is exactly where the claim differs from the other co-sign screens since keysign always renders a functional Next (moveToKeysignState) while keygen hides it for 2/2 and 3/3 and keeps it functional for 4+ whereas the claim auto-advances the moment the peer threshold is met for all sizes and so has no Next step to show.

johnnyluo:
CleanShot 2026-06-08 at 22 25 24@2x

This is the Peer discovery screen , the QR code can't be scanned by my secondly device , which is a pixel 8

rkokhatskyi:
> This is the Peer discovery screen , the QR code can't be scanned by my secondly device , which is a pixel 8

I will test it with 2 Android devices in a bit, but so far I'm able to scan the QR and sign it with Android as initiator and iOS as joining device

johnnyluo:
image

Once I claim successfully , the join device is stuck at this screen , can you raise another PR to get this one addressed

johnnyluo (approved):
LGTM

Issue #4469, opened 2026-05-13
LP Select Positions: token icons missing chain badge — chain is ambiguous


aminsato:
# [Fix] LP Select Positions: add chain badge to token icons [DeFi / Select Positions]

## Problem

On the **Select Positions** screen (both Stake and Liquidity Pool sections), each grid item renders only the token icon with no chain badge. This makes LP pools with the same partner token across different chains visually identical — for example, four `RUNE/USDT` pools on Avalanche, BSC, Ethereum, and Tron are completely indistinguishable. Stake items (`RUJI`, `TCY`, `sTCY`, `yRUNE`, `yTCY`) also carry no chain context.

## Expected Behavior

Each grid item on the Select Positions screen overlays a small chain badge on the bottom-right corner of the token icon. For LP pools the badge reflects the partner token's chain (the base side is always RUNE/CACAO). For Stake items the badge reflects the token's host chain (THORChain for RUJI/TCY/sTCY/yRUNE/yTCY, MayaChain for CACAO). The badge remains visible in search results and in both selected and unselected states.

## Steps to Reproduce

1. Open the app on Android
2. Navigate to the DeFi section
3. Tap Stake or Liquidity Pool to open the Select Positions dialog
4. Observe that token icons have no chain badge
5. Note that multiple `RUNE/USDT` pools (Avalanche / BSC / Ethereum / Tron) render with identical icons

### Must Do

- [ ] **Extend the dialog model** — add `val chainLogo: ImageModel? = null` to `PositionUiModelDialog` in `PositionUiModelDialog.kt`
- [ ] **Populate `chainLogo` at all three construction sites** — in `DeFiExtensions.kt` (`toPositionDialogModels`) set `chainLogo = coin.chain.logo`; in `ThorchainDefiPositionsViewModel.kt` (`String.toLpPositionDialogModel`) pass through `parseThorChainPool(this).chain?.logo`; in `MayachainDefiPositionsViewModel.kt` (`MayaNodePool.toPositionDialogModel`) pass through `mayaPoolChainPrefixToChain(...)?.logo`
[...]

PR #4796, opened 2026-06-08
feat(custom-rpc): app-wide custom RPC endpoints per chain, EVM + Cosmos (#4787)

### LINKED ISSUE #4787: feat(custom-rpc): app-wide custom RPC endpoints per chain, Silver tier (iOS #4469 parity)
## Custom RPC endpoint support per chain (Silver Vult tier) — Android parity

Port of iOS #4469 / #4458. App-wide custom RPC endpoint overrides, gated to **Silver** VULT tier and above. Lets a qualifying user point a chain at their own node from the Advanced Settings screen.

### What
- Per-chain custom RPC URL override, **app-wide** (not vault-scoped).
- Gated to Silver tier and above (reuse existing tier gate).
- Entry point: Advanced Settings → Custom RPC screen (add / edit / clear per chain).
- Networking layer resolves the override when present, else falls back to the default RPC for the chain.

### Persistence
- App-wide store keyed by chain (Room/DataStore — Android equivalent of the iOS SwiftData @Model + in-memory mirror).
- RPC resolution happens off the main thread (balance / fee / broadcast run on background) — store must expose a **thread-safe synchronous read** for the networking layer.

### Acceptance Criteria
- [ ] Silver+ user can set, edit, and clear a custom RPC URL per supported chain
- [ ] Override is applied to balance, fee estimation, and broadcast calls for that chain
- [ ] Non-Silver users cannot access the screen (tier gate)
- [ ] Clearing the override restores the default RPC
- [ ] URL validation rejects malformed endpoints
- [ ] Matches iOS behaviour and the Figma for the Custom RPC screen

### Reference
iOS impl: vultisig/vultisig-ios#4469 (tracking #4458)

aminsato:
## What

Port of iOS #4469: app-wide custom RPC endpoint overrides, letting a qualifying user point a supported chain at their own node from a hidden **Custom RPC** screen. v1 covers the **EVM** and **Cosmos** chain groups, whose networking funnels through a single factory chokepoint with a raw-node RPC URL.

Closes #4787.

## Screenshots

| Advanced toggle | Settings row | Editor (Test) | Chain list |
|---|---|---|---|
| toggle | row | detail | list |

## Scope (v1)

- **Chains:** EVM (13) + Cosmos (8). UTXO and `Qbtc` are intentionally excluded — they route through the `api.vultisig.com` proxy, not a raw node, so a user node URL has no coherent insertion point (matches iOS). Single-instance chains (Solana, THORChain, Tron, …) are a planned follow-up.
- **Entry:** hidden behind an Advanced Settings → **Custom RPC** feature flag (default OFF). When on, a Custom RPC row appears in General Settings.
- **Tier gate:** below Silver, tapping a chain routes to the Discount Tiers upsell; Silver+ opens the editor (reuses `GetDiscountBpsUseCase.hasReachedSilverTier`).

## Data layer

- **`CustomRpcRepository`** — overrides persisted app-wide as a JSON `{chainId: url}` map in DataStore, fronted by a **thread-safe `ConcurrentHashMap` mirror** so the non-suspend API factories resolve an override with a *synchronous* read off the main thread (balance/fee/broadcast). Hydrates at launch (survives relaunch); writes are read-modify-write against the store under a `Mutex`.
- **`CustomRpcSupportedChains`** — single source of truth shared by funnel and UI, so a chain shown is exactly a chain the funnel honors.
- **Funnel** — `EvmApiFactoryImp` / `CosmosApiFactoryImp` resolve `urlFor(chain) ?: default`. **Byte-identical default behaviour when unset** (pinned by tests).
- **`CustomRpcConfig`** — `customRpcEnabled` flag, default OFF.
- **`RpcHealthProbe`** — EVM `eth_chainId` (verifies chain identity) + Cosmos `node_info` (liveness); returns `Reachable(latency, verified)` / `WrongChain` / `InvalidResponse` / `Unreachable`.

## App layer

- Custom RPC **list** (per-chain rows, Default/Custom chip, active URL) and **detail** (URL input + inline validation, Test, Save, Reset) screens.
- Advanced Settings toggle; General Settings row gated on the flag; tap-to-upsell below Silver.
- `custom_rpc_*` strings added across all 10 locales.

## Test plan

- [x] `:data` unit tests: `CustomRpcRepositoryImplTest` (mirror + persistence + decode, incl. a race-condition regression), `CustomRpcFunnelTest` (override-applied + default-preserved per group), `RpcHealthProbeImplTest` (EVM verified/wrong-chain/invalid, Cosmos liveness, unreachable).
- [x] Updated `SecretViewModelTest` + `SettingsViewModelTest`; full `:data` suite green.
- [x] `:app:assembleDebug` succeeds; ktfmt applied.
- [x] **Device-verified** on emulator (screenshots above): enable flag → set Ethereum to a public node → `Test` shows "Reachable · 266 ms" → Save → row shows Custom. Default chains keep the Vultisig endpoint.

## Notes

- Cosmos probe is liveness-only (`networkVerified=false`) — per-chain Cosmos network-id verification was deliberately not mapped, to avoid showing a false "verified".

## Summary by CodeRabbit

* **New Features**
  * Advanced Custom RPC: toggle in advanced settings; list and detail screens to add, test, save, or reset per‑chain RPC endpoints; live reachability and identity checks.
  * App networking can use per‑chain Custom RPC overrides when enabled.

* **Settings**
  * Custom RPC item shown/hidden by feature flag; Silver‑tier gating with upsell and unlock flow.

* **Localization**
  * Custom RPC UI strings added for many locales.

* **Tests**
  * Coverage for feature flag, repository persistence, API resolution, and probe behavior.


app/src/main/java/com/vultisig/wallet/ui/models/customrpc/CustomRpcDetailViewModel.kt
```diff
@@ -0,0 +1,140 @@
...
+    private val args = savedStateHandle.toRoute<Route.CustomRpcDetail>()
+    private val chain = Chain.fromRaw(args.chainId)
+
+    val urlFieldState = TextFieldState()
+
+    private val _state = MutableStateFlow(CustomRpcDetailUiState(chainName = chain.raw))
+    val state: StateFlow<CustomRpcDetailUiState> = _state.asStateFlow()
+
+    init {
+        val existing = customRpcRepository.urlFor(chain)
```
johnnyluo: **`urlFor()` can block the main thread on first use.**

This runs in the ViewModel `init`, i.e. on the main thread. `urlFor` → `ensureHydrated()` → `runBlocking(Dispatchers.IO) { … }` does a blocking disk read the first time it's hit per process. The `CustomRpcRepository` KDoc asserts *"urlFor is only reached from the off-main networking path"* — this call path violates that.

In practice a balance/fee refresh usually hydrates the mirror off-main before the user reaches Settings, so it only bites if the detail screen is opened before any EVM/Cosmos networking — but it's a real latent main-thread block, and the doc comment is now inaccurate.

Suggest hydrating the mirror eagerly (fire-and-forget on `Dispatchers.IO` at inject time) or reading the existing override here via the suspend/flow API, so `urlFor` never has to block.

app/src/main/java/com/vultisig/wallet/ui/models/customrpc/CustomRpcListViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+                    }
+                _state.update { it.copy(rows = rows) }
+            }
+        }
+    }
+
+    fun onRowClick(chainId: String) {
+        // Block taps until the tier lookup resolves so a Silver user is never misrouted to the
+        // upsell screen on an early tap (#4787).
+        val isSilver = _state.value.isSilver ?: return
```
johnnyluo: **A failed tier lookup permanently bricks the list screen.**

`isSilver` is set via `runCatching { … }.getOrNull()`, so a transient failure of `hasReachedSilverTier` leaves it `null`. Here `val isSilver = _state.value.isSilver ?: return` then makes every row silently un-tappable — no error, no retry, no feedback.

The "block taps until the tier resolves" intent is good, but a *failed* lookup should surface an error or fall back (e.g. retry, or route to the upsell), not deadlock the screen.

johnnyluo (commented):
Two follow-ups from review — a latent main-thread block and a screen that can deadlock on a failed tier lookup. Details inline.

app/src/main/java/com/vultisig/wallet/ui/models/customrpc/CustomRpcListViewModel.kt
```diff
@@ -0,0 +1,97 @@
...
+        viewModelScope.launch {
+            // Tap-to-upsell: below Silver tier the row routes to the discount-tiers screen instead
+            // of the editor (#4787).
+            if (isSilver) {
+                navigator.route(Route.CustomRpcDetail(vaultId, chainId))
+            } else {
+                navigator.route(Route.DiscountTiers(vaultId))
```
johnnyluo: Let's add a feature gate when user click "Custom RPC" on the settings menu? 

If user is not Silver tier , then let's prompt user they need to be at least silver tier to do it , similar as IOS

Once user is already in the Custom RPC list , then we should just allow them to change it

app/src/main/java/com/vultisig/wallet/ui/models/customrpc/CustomRpcDetailViewModel.kt
aminsato: Good catch — fixed. The detail VM no longer calls the synchronous `urlFor()`. It now reads the existing override off-main via the reactive API in a coroutine (`customRpcRepository.overrides.first()[chain]`), so `init` never triggers the first-use blocking disk hydration on the main thread. With this, `urlFor()` is again only reached from the EVM/Cosmos API factories (off-main), so the KDoc claim holds.

app/src/main/java/com/vultisig/wallet/ui/models/customrpc/CustomRpcListViewModel.kt
aminsato: Resolved by moving the tier gate to the settings entry point (see your other comment). The per-row `isSilver` check is gone from the list VM entirely, so a failed tier lookup can no longer brick the screen. Reaching the list now implies eligibility, and a failed lookup at the entry falls back to the upsell rather than deadlocking.

app/src/main/java/com/vultisig/wallet/ui/models/customrpc/CustomRpcListViewModel.kt
aminsato: Done — gated at the entry point to match iOS. Tapping "Custom RPC" in Settings now checks Silver tier: below Silver it routes to the discount-tiers upsell, at/above Silver it opens the list. Once inside the list the user edits freely (no per-row gating). A failed/unknown tier lookup falls back to the upsell rather than granting access. Added unit tests for all three branches (Silver → list, below-Silver → upsell, lookup-failure → upsell).

johnnyluo: CleanShot 2026-06-09 at 20 06 58@2x
Can you please show a dialog like this? instead go to the VULT tiers screen

aminsato: Done in 7657e9ffd. Note the tier gate moved out of `CustomRpcListViewModel` and into the settings entry point (commit 8b4fb7e8) since this comment, so I applied it there: tapping Custom RPC below Silver now shows the existing `TierDiscountBottomSheet` (Silver) dialog instead of navigating away. Its **Unlock Tier** button still routes to the discount-tiers screen so the path to unlocking stays reachable.

app/src/main/java/com/vultisig/wallet/ui/models/settings/SettingsViewModel.kt
```diff
@@ -357,6 +389,33 @@ constructor(
...
+                state.update { current ->
+                    current.copy(
+                        items =
+                            current.items.map { group ->
+                                if (group.title != UiText.StringResource(R.string.general)) {
+                                    group
+                                } else {
+                                    val withoutRpc = group.items.filterNot { it is CustomRpc }
+                                    group.copy(
+                                        items = if (enabled) withoutRpc + CustomRpc else withoutRpc
```
rkokhatskyi, line 412: `loadCustomRpcFlag` adds the `CustomRpc` row inside `state.update`, but `loadCurrency` and `loadAppLocale` compute `items` from `state.value.items` outside the update then overwrite it, so an interleaved currency or locale emission at init drops the freshly added row

data/src/main/kotlin/com/vultisig/wallet/data/repositories/CustomRpcRepository.kt
```diff
@@ -0,0 +1,134 @@
...
+ * networking layer can resolve an override with a *synchronous, non-suspending* read ([urlFor]) —
+ * the API factories build their RPC client off the main thread and are not suspend functions, so
+ * they cannot await DataStore. `null` from [urlFor] means "no override" and the caller keeps its
+ * hardcoded default, guaranteeing byte-identical default behaviour when nothing is set.
+ *
+ * The mirror is hydrated from disk lazily on the first [urlFor] call (the repo is a [Singleton], so
+ * this happens once per process and survives relaunch) and kept in sync with every [setOverride] /
+ * [clearOverride]. Hydrating on first use — rather than fire-and-forget at construction —
+ * guarantees the first networking lookup after process start sees persisted overrides instead of
+ * racing an async job. [urlFor] is only reached from the off-main networking path, so the one-time
```
rkokhatskyi: This invariant is false because `GasSettingsViewModel.loadEthData` calls `createEvmApi` inside a bare `viewModelScope.launch`, so `urlFor` runs on the main thread where the first-use `runBlocking(Dispatchers.IO)` hydration blocks it

data/src/main/kotlin/com/vultisig/wallet/data/repositories/CustomRpcRepository.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    // Populate the mirror from disk on first lookup so the very first networking call after process
+    // start sees persisted overrides. Runs on the calling (off-main networking) thread; the double-
+    // checked flag keeps it to a single blocking read regardless of concurrent callers.
+    private fun ensureHydrated() {
+        if (hydrated) return
+        synchronized(hydrationLock) {
+            if (hydrated) return
+            runBlocking(Dispatchers.IO) { replaceMirror(readPersisted()) }
```
rkokhatskyi: A `readPersisted()` failure propagates uncaught out of `ensureHydrated`, so `createEvmApi` throws instead of falling back to the default endpoint, breaking the byte-identical-default guarantee for balance, fee, and broadcast on a DataStore read error

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/settings/SettingsViewModel.kt
aminsato: Good catch — fixed. `loadCurrency` and `loadAppLocale` now map *inside* the `state.update { ... }` lambda (`updateCurrency(it.items, currency)` / `updatedLocale(it.items, locale)`) instead of computing from a `state.value.items` snapshot and overwriting. So an interleaved currency/locale emission at init now folds over the current state and can no longer drop the freshly-added `CustomRpc` row.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/CustomRpcRepository.kt
aminsato: Right, the invariant was false — `GasSettingsViewModel.loadEthData` calls `createEvmApi` inside a bare `viewModelScope.launch` (Main), so the first-use `runBlocking(Dispatchers.IO)` hydration blocked the main thread. Fixed at the source: the mirror now hydrates eagerly off-main in `init` (`scope.launch { hydrate() }` on `Dispatchers.IO`), and `urlFor()` is a pure non-blocking `mirror[chain.id]` read with no `runBlocking`. Updated the KDoc to drop the false claim and document the eager-hydration model.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/CustomRpcRepository.kt
aminsato: Fixed. `hydrate()` now wraps the disk read in `runCatching { replaceMirror(readPersisted()) }.onFailure { Timber.w(...) }`, so a DataStore read failure leaves the mirror empty instead of propagating. `urlFor()` then returns `null` and `createEvmApi`/`createCosmosApi` fall back to the hardcoded default, preserving the byte-identical-default guarantee for balance/fee/broadcast. Hydration also runs under `writeLock` and is gated by the `hydrated` flag so it cannot clobber a concurrent set/clear.

johnnyluo (approved):
LGTM

Issue #4798, opened 2026-06-08
QBTC claim: show a 'UTXO still maturing' state instead of generic 'No claimable UTXOs'


rkokhatskyi:
## Problem
After #4797, QBTC UTXOs with ≤144 confirmations are filtered out of the claim flow. A user whose only UTXOs are still maturing therefore lands on the generic **"No claimable UTXOs"** screen with no explanation — even though their funds become claimable shortly.

## Desired
A dedicated state that explains the UTXO is still maturing and how many more confirmations/blocks (≈ time) remain before it can be claimed, rather than implying nothing is claimable.

## Design
Figma: https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=78433-177381

State: **Claimable QBTC — all maturing** (claimable balance `0 QBTC`).

**Empty-state info card** (`backgrounds/surface-1` #061b3a, radius 20):
- Hourglass icon
- Title — "Your BTC is still maturing" (Title3, `text/primary` #f0f4fc)
- Body — "UTXOs need 144 confirmations before they can be claimed as QBTC. Your funds are safe." (Footnote, `text/tertiary` #8295ae)

**"Maturing UTXOs"** section label (Footnote, `text/tertiary` #8295ae), followed by one card per maturing UTXO (`surface-1`, radius 16):
- Clock icon
- Truncated outpoint, e.g. `a3f1...8d2c:0` (Body S SemiBold, white)
- `<conf>/144 confirmations` — the `<conf>/144` portion in warning amber `alerts/warning` #ffc25c, the "confirmations" label in #8295ae
- `~<n> blocks` remaining (#8295ae)
- Right column: amount, e.g. `0.75 QBTC` over `0.75 BTC` (#8295ae)

**Claim All** button stays disabled in this state (`buttons/disabled` #0b1a3a, label `text/button/disabled` #718096).

## Notes
- Data already available: `ClaimableUtxo.confirmations` + the `MIN_CLAIM_CONFIRMATIONS` threshold added in #4797.
- Remaining blocks ≈ `144 − confirmations`; surface that as the "~N blocks" figure.
- Follow-up to #4797.

rkokhatskyi:
@aminsato this still pending Figma, cannot be started yet (I opened a thread in Discord)

aminsato:
> [@aminsato](https://github.com/aminsato) this still pending Figma, cannot be started yet (I opened a thread in Discord)

[...]

PR #4799, opened 2026-06-08
fix(qbtc): block claims for UTXOs with 144 or fewer confirmations

### LINKED ISSUE #4797: QBTC claim: block UTXOs with 144 or fewer confirmations before signing
## Problem
A QBTC claim is rejected by the chain when the funding Bitcoin UTXO has 144 or fewer confirmations. The chain's gas simulation fails with:

```
broadcast failed ... no valid claimable UTXOs found
```

This happens **after** the user has already co-signed (Secure Vault) / triggered the FastVault co-sign, so a too-fresh UTXO wastes a full signing round and ends in a generic "Claim failed".

The claimable-UTXO pipeline (`LoadClaimableQbtcUtxosUseCase` → `QbtcClaimableUtxosService` / `QbtcClaimChainService`) has no minimum-confirmation gate — `ClaimableUtxo.confirmations` is currently used for display only.

## Expected
The QBTC chain requires **more than 144 confirmations** for a UTXO to be claimable. Filter out UTXOs at/below that threshold (and unconfirmed ones) before they can be selected or signed, so the claim never fails on-chain. When no mature UTXOs remain, the existing "No claimable UTXOs" state is shown.

## Scope
Data-layer fix + unit tests only. The richer "this UTXO is maturing — wait N more blocks" UX needs Figma and is tracked separately.

rkokhatskyi:
The QBTC chain rejects a claim at broadcast unless the funding Bitcoin UTXO has **more than 144 confirmations** — it fails gas simulation with `no valid claimable UTXOs found`. Today nothing stops a user from selecting a too-fresh UTXO, so the failure only shows up *after* a full co-sign round, ending in a generic "Claim failed".

This gates the claim eligibility pipeline on confirmation depth: any UTXO with ≤144 confirmations (or an unconfirmed / unknown-tip one) is filtered out before it can be selected, so the claim never fails on-chain. When no mature UTXOs remain, the existing "No claimable UTXOs" state is shown.

### Changes
- `QbtcClaimConfig.MIN_CLAIM_CONFIRMATIONS = 144` — single source of truth for the chain rule.
- `ClaimableUtxo.isMature()` — `confirmations > MIN_CLAIM_CONFIRMATIONS`; `null` (mempool / unknown tip) is treated as immature (fail closed).
- `LoadClaimableQbtcUtxosUseCase` filters immature UTXOs out after the chain cross-check.

### Tests
`LoadClaimableQbtcUtxosUseCaseTest` — boundary at 143 / 144 / 145, null confirmations, mixed sets (only mature survive), and all-immature → `NoUtxos`.

### Scope
Data-layer only. The richer "this UTXO is still maturing — wait N more blocks" UX needs Figma and is tracked in #4798.

Closes #4797

## Summary by CodeRabbit

* **New Features**
  * QBTC claims now enforce maturity verification: only funding UTXOs with strictly more than 144 confirmations are eligible; unconfirmed/mempool or unknown-tip UTXOs are treated as immature.

* **Tests**
  * Added comprehensive tests covering confirmation thresholds, null/unknown confirmations, mixed sets, and no-UTXO failure cases.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/qbtc/claim/ClaimableUtxo.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/qbtc/claim/ClaimableUtxo.kt
@@ -4,8 +4,8 @@ package com.vultisig.wallet.data.blockchain.cosmos.qbtc.claim
  * A Bitcoin UTXO that can be claimed on the QBTC chain. [amount] is in satoshis; once the chain has
  * cross-checked the UTXO it reflects the chain's remaining `entitled_amount`, not the raw
  * Blockchair value. [confirmations] is the number of blocks mined since the UTXO was confirmed
- * (current tip − the UTXO's block height + 1), used only for display; it is `null` for an
- * unconfirmed (mempool) UTXO or when the chain tip is unknown.
+ * (current tip − the UTXO's block height + 1); it is `null` for an unconfirmed (mempool) UTXO or
+ * when the chain tip is unknown. It also gates claim maturity — see [isMature].
  */
 data class ClaimableUtxo(
     val txid: String,
@@ -13,3 +13,11 @@ data class ClaimableUtxo(
     val amount: Long,
     val confirmations: Long? = null,
 )
+
+/**
+ * True when the UTXO has enough confirmations to be claimable. The chain requires strictly more
+ * than [QbtcClaimConfig.MIN_CLAIM_CONFIRMATIONS], so a `null` count (mempool / unknown tip) is
+ * treated as immature — fail closed.
+ */
+internal fun ClaimableUtxo.isMature(): Boolean =
+    (confirmations ?: 0L) > QbtcClaimConfig.MIN_CLAIM_CONFIRMATIONS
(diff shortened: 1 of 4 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/qbtc/claim/ClaimableUtxo.kt
```diff
@@ -4,12 +4,20 @@ package com.vultisig.wallet.data.blockchain.cosmos.qbtc.claim
...
     val confirmations: Long? = null,
 )
+
+/**
+ * True when the UTXO has enough confirmations to be claimable. The chain requires strictly more
+ * than [QbtcClaimConfig.MIN_CLAIM_CONFIRMATIONS], so a `null` count (mempool / unknown tip) is
+ * treated as immature — fail closed.
+ */
+internal fun ClaimableUtxo.isMature(): Boolean =
+    (confirmations ?: 0L) > QbtcClaimConfig.MIN_CLAIM_CONFIRMATIONS
```
aminsato: Optional readability nit: the `?: 0L` makes the reader reason that `0L > 144` is false to understand the null case. Stating the null guard directly reads closer to the doc comment ("a `null` count is treated as immature"):

```suggestion
internal fun ClaimableUtxo.isMature(): Boolean =
    confirmations != null && confirmations > QbtcClaimConfig.MIN_CLAIM_CONFIRMATIONS
```

rkokhatskyi: Done in 18245e396 — went a step further and made it an extension property since it is a pure derived attribute (it now reads as it.isMature at the call site) with your explicit null guard as the body, so confirmations != null && confirmations > MIN_CLAIM_CONFIRMATIONS states the null-is-immature case directly instead of routing through the ?: 0L sentinel.

johnnyluo (approved):
LGTM

PR #4802, opened 2026-06-08
fix(swap): send KyberSwap slippage tolerance as 100 bps

### LINKED ISSUE #4800: [Fix] KyberSwap slippage tolerance sent as 2.5 bps instead of 100 [Swap]
# [Fix] KyberSwap slippage tolerance sent as 2.5 bps instead of 100 [Swap]

## Problem
Android sends `slippageTolerance = 2.5` to KyberSwap's aggregator API on both the `/routes` query and the `/route/build` body. KyberSwap denominates this field in basis points (10000 = 100%), so `2.5` is ~0.025% — two orders of magnitude tighter than the `100` (= 1%) that iOS and the Windows extension send. Almost any price movement between quote and execution exceeds 0.025%, so Kyber route builds and on-chain swaps revert or get dropped, which effectively removes KyberSwap as a usable provider for most pairs on Android and degrades best-rate ranking.

On top of the wrong magnitude, the request model types the field as `Double`, so the build POST serialises it as `2.5` (and would serialise `100.0` if the constant were only bumped) rather than the integer `100` that iOS/Windows send.

## Expected Behavior
Android sends `100` (1%) as an integer on both the routes query and the build body, matching iOS (`KyberSwapService.swift` sends `slippageTolerance: 100` typed `Int`) and the Windows extension (shared `kyberSwapSlippageTolerance = 100`).

## Steps to Reproduce
1. Start a swap on an EVM chain where KyberSwap is a provider (e.g. Ethereum USDC → USDT).
2. Let Kyber produce a quote and build the transaction.
3. Observe the Kyber route build / on-chain swap reverting far more often than the same pair on iOS.

## Solution
Treat `slippageTolerance` as integer basis points. Change `SLIPPAGE_TOLERANCE` from `2.5` to `100` in `KyberApi.kt`, and type `KyberSwapBuildRequest.slippageTolerance` as `Int` so the build POST serialises a clean `100` instead of `100.0`.

### Must Do
- [ ] Send `100` basis points on the `/routes` query (`KyberApi.kt`).
- [ ] Send `100` basis points as an integer in the `/route/build` body (`KyberApi.kt`, `KyberJson.kt`).

### Must NOT Do
- Don't change unrelated swap providers or the affiliate-fee parameters.

## Acceptance Criteria
- [ ] Both Kyber calls send `slippageTolerance` = `100`.
- [ ] The build body serialises `slippageTolerance` as the integer `100`, not `100.0` or `2.5`.

rkokhatskyi:
Closes #4800

KyberSwap denominates `slippageTolerance` in basis points (10000 = 100%). Android sent `2.5` (~0.025%) on both the `/routes` query and the `/route/build` body — two orders of magnitude tighter than the `100` (1%) that iOS and the Windows extension send — so Kyber route builds and on-chain swaps reverted for most pairs, effectively dropping KyberSwap as a usable provider on Android.

## Changes

- **`KyberApi.kt`** — `SLIPPAGE_TOLERANCE` changed from `2.5` to `100` (with a basis-points comment). It is sent verbatim on both the routes GET query and the route/build POST body.
- **`KyberJson.kt`** — typed `KyberSwapBuildRequest.slippageTolerance` as `Int` so the build POST serialises a clean `100` instead of `100.0`, matching iOS (`KyberSwapService.swift` sends `slippageTolerance: 100` typed `Int`) and the Windows extension (`kyberSwapSlippageTolerance = 100`).

## Signing

No new signing path. The slippage value only changes what KyberSwap bakes into the router calldata it returns; that calldata is signed verbatim by the existing keysign path, so co-signing is unaffected — only the swap's slippage bound now matches the other platforms.

## Test plan

- [x] `./gradlew :data:testDebugUnitTest --tests "*KyberApiSlippageToleranceTest*" --tests "*KyberApiErrorPathTest*"` — passing
- [x] New tests assert `slippageTolerance` is sent as the integer `100` on both the routes query and the build body, and fail without the change (`expected: <100> but was: <2.5>`)

## Notes

- No UI surface — outbound aggregator request parameter only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated slippage tolerance configuration in Kyber swap operations to align with the new expected format and value requirements.

* **Tests**
  * Added test coverage for Kyber swap slippage tolerance validation to ensure correct request formatting.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/models/KyberJson.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/models/KyberJson.kt
@@ -73,7 +73,7 @@ data class KyberSwapBuildRequest(
     val routeSummary: KyberSwapRouteResponse.RouteSummary,
     val sender: String,
     val recipient: String,
-    val slippageTolerance: Double,
+    val slippageTolerance: Int,
     val deadline: Int,
     val enableGasEstimation: Boolean,
     val source: String,
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/swapAggregators/KyberApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/swapAggregators/KyberApi.kt
@@ -189,6 +189,7 @@ constructor(
         private const val REFERRER_ADDRESS = "0x8E247a480449c84a5fDD25974A8501f3EFa4ABb9"
         private const val CLIENT_ID = "vultisig-android"
         private const val NULL_ADDRESS = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
-        private const val SLIPPAGE_TOLERANCE = 2.5
+        // Slippage tolerance in basis points (10000 = 100%); 100 = 1%.
+        private const val SLIPPAGE_TOLERANCE = 100
     }
 }
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/KyberApiSlippageToleranceTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/KyberApiSlippageToleranceTest.kt
@@ -1,0 +1,109 @@
+package com.vultisig.wallet.data.api
+
+import com.vultisig.wallet.data.api.errors.SwapException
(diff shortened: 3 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4804, opened 2026-06-08
fix(swap): reject zero network fee or expected output

### LINKED ISSUE #4801: [Fix] Swap form stages a transaction with zero network fee or zero expected output [Swap]
# [Fix] Swap form stages a transaction with zero network fee or zero expected output [Swap]

## Problem
`SwapFormViewModel.swap()` can build and stage a swap (advancing to Verify Swap) in degenerate states that iOS blocks at the form level. It only checks that `gasFee` and `estimatedNetworkFeeFiatValue` are **non-null** — a non-null but **zero** gas fee passes. The source amount is guarded against zero (`takeIf { it != BigInteger.ZERO }`), but the destination amount (`quote.expectedDstValue`) is consumed without any zero check.

The Continue button is gated only on `isSwapDisabled || isLoading`, and for non-UTXO swaps `isSwapDisabled` flips to `false` the moment a quote loads, so there is no equivalent `fee != 0` / `expectedDst != 0` precondition before the transaction is materialised.

## Expected Behavior
Staging a swap is rejected when the resolved network fee is zero or the quote's expected destination amount is zero, matching iOS. `SwapCryptoLogic.validateForm` requires `fee != .zero` and `!toAmount.isZero` (alongside `quote != nil` and a sufficient balance) before a transaction can be created. A successful quote normally implies both values are non-zero, so this is a defensive boundary for a degenerate state.

## Steps to Reproduce
1. Start a non-UTXO swap and let a quote load (so `isSwapDisabled` becomes `false`).
2. Reach a state where the resolved gas fee is zero, or the quote returns a zero expected output.
3. Tap Continue — the swap is staged and advances to Verify Swap instead of being blocked.

## Solution
Add two form-level guards in `swap()`, mirroring the existing `srcAmount` zero check and reusing the existing error strings (no new localization): reject a resolved gas fee of zero with `swap_screen_invalid_gas_fee_calculation`, and reject a quote whose `expectedDstValue` is zero with `swap_screen_invalid_quote_calculation`. Both run before the transaction is built or persisted, so the keysign/co-sign flow only ever receives a valid swap.

### Must Do
- [ ] Reject staging when the resolved gas fee is zero (`SwapFormViewModel.swap()`).
- [ ] Reject staging when `quote.expectedDstValue` is zero (`SwapFormViewModel.swap()`).
- [ ] Cover both guards with unit tests.

### Must NOT Do
- Don't change unrelated code
- Don't add new string resources / localizations

## Acceptance Criteria
- [ ] A zero gas fee no longer advances past the swap form
- [ ] A zero expected destination amount no longer advances past the swap form
- [ ] Valid swaps continue to reach Verify Swap unchanged

rkokhatskyi:
## Summary
`SwapFormViewModel.swap()` only checked that the gas fee and its fiat value were non-null, so a resolved-but-zero gas fee — or a quote with a zero expected destination amount — could still be staged and advance to Verify Swap. iOS blocks both at the form level (`SwapCryptoLogic.validateForm` requires `fee != .zero` and `!toAmount.isZero`).

This adds the two missing guards to `swap()`, mirroring the existing source-amount zero check and reusing the existing gas-fee / quote error strings (no new copy or localization). Both run before the transaction is built or persisted, so the keysign/co-sign flow only ever receives a valid swap; a normal successful quote already implies a non-zero fee and output, so this is a defensive boundary for a degenerate state.

## Testing
`./gradlew :app:testDebugUnitTest --tests "*.SwapFormViewModelTest" --tests "*.SwapValidatorTest"` — added two regression tests for the zero-fee and zero-output cases (both fail without the guards).

Closes #4801

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap transaction validation to reject invalid gas fee calculations and zero-value quote amounts
  * Enhanced error handling to prevent swaps from proceeding with incomplete or corrupted quote information
  * Strengthened pre-transaction checks to ensure all required values are valid before staging operations

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -206,7 +206,7 @@ constructor(
                     )
 
             val gasFee =
-                gasFee.value
+                gasFee.value?.takeIf { it.value != BigInteger.ZERO }
                     ?: throw InvalidTransactionDataException(
                         UiText.StringResource(R.string.swap_screen_invalid_gas_fee_calculation)
                     )
@@ -249,7 +249,7 @@ constructor(
             val srcTokenValue = convertTokenAndValueToTokenValue(srcToken, srcAmountInt)
 
             val quote =
-                quoteState.quote
+                quoteState.quote?.takeIf { it.expectedDstValue.value != BigInteger.ZERO }
                     ?: throw InvalidTransactionDataException(
                         UiText.StringResource(R.string.swap_screen_invalid_quote_calculation)
                     )
--- a/app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
@@ -1648,6 +1648,97 @@ internal class SwapFormViewModelTest {
             assertFalse(vm.uiState.value.isLoadingNextScreen)
         }
 
+    @Test
+    fun `swap with zero gas fee shows error and does not stage`() =
+        runTest(mainDispatcher) {
+            coEvery { swapGasCalculator.calculateGasFee(any(), any()) } returns
+                GasCalculationResult(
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4805, opened 2026-06-08
fix(solana): improve Solana SPL transfer error handling

### LINKED ISSUE #4803: [Fix] Solana SPL send crashes or builds an invalid tx when the sender has no associated token account [Send]
# [Fix] Solana SPL send crashes or builds an invalid tx when the sender has no associated token account [Send]

## Problem
When sending an SPL token, `SolanaHelper.getPreSignedInputData` splits the non-native path in two: if both the sender and recipient associated token accounts (ATAs) are present it builds a `TokenTransfer`, otherwise it builds a `CreateAndTransferToken` and unconditionally calls `setSenderTokenAddress(solanaSpecific.fromAddressPubKey)`. `fromAddressPubKey` is nullable — `BlockChainSpecificRepository` sets it to `getTokenAssociatedAccountByOwner(...).first`, which returns `null` both when the sender genuinely has no ATA for the mint and on any RPC failure (`SolanaApi.getTokenAssociatedAccountByOwner` returns `Pair(null, false)` on an error response, an empty result, or a caught exception).

When `fromAddressPubKey` is null the `else` branch still runs, so `CreateAndTransferToken.setSenderTokenAddress(null)` throws an NPE (protobuf-lite null check); an empty string instead builds a malformed create+transfer with an unset source account that fails at preflight/broadcast. Either way the user gets a crash or an unbroadcastable transaction instead of a clear error.

## Expected Behavior
When the sender has no associated token account for the SPL token being sent (or the lookup fails), the send aborts with a clear, localized error instead of crashing or building an invalid transaction — matching iOS, which throws `sender's associated token account not found. Please ensure you have this token in your wallet.`. A present sender ATA with a missing recipient ATA still creates the recipient's account inline.

## Steps to Reproduce
1. Hold SOL but no balance/ATA for a given SPL mint (or force the sender ATA RPC lookup to fail).
2. Start a send of that SPL token.
3. Proceed to keysign and observe the crash / unbroadcastable transaction instead of a readable error.

## Solution
Split the non-native branch in `SolanaHelper` to mirror iOS: both ATAs present → `TokenTransfer`; sender present, recipient missing → `CreateAndTransferToken`; sender missing → abort with a clear error. Map that error to a localized, user-facing message in `KeysignErrorScreen` (added to all locales).

### Must Do
- [ ] Guard the SPL branch in `SolanaHelper` so a null/empty sender ATA aborts with a clear error instead of reaching `setSenderTokenAddress`.
- [ ] Keep creating the recipient ATA inline when only the recipient account is missing.
- [ ] Surface a localized message for the new error in `KeysignErrorScreen`.

### Must NOT Do
- Don't change unrelated chains or the recipient-ATA / fee-estimation paths.

## Acceptance Criteria
- [ ] Sending an SPL token with no sender ATA shows a clear error instead of crashing.
- [ ] Sending an SPL token whose recipient lacks an ATA still succeeds (account created inline).

rkokhatskyi:
Closes #4803

Sending an SPL token when the sender has no associated token account (ATA) for that mint — or when the ATA lookup fails over RPC — left `fromAddressPubKey` null. The create-and-transfer branch passed that null straight into `CreateAndTransferToken.setSenderTokenAddress`, which crashes with an NPE (protobuf-lite null check) or, with an empty string, builds an unbroadcastable transaction. There was no clean failure path.

## Changes

- `SolanaHelper` guards the SPL path: if the sender has no associated token account it aborts early (before any WalletCore call) with a clear error carrying the token ticker, instead of passing null into the token instructions. The non-native branch then only distinguishes a present vs missing recipient account (creating the recipient's account inline when needed).
- `KeysignErrorScreen` maps that error to a localized, token-named message (`signing_error_token_account_not_found_s`, added to all locales). The lookup almost always fails transiently, so the copy frames it as a retry rather than implying the user lacks the token.

## Signing

The guard runs while computing the pre-image hashes, before any keysign session starts, so Fast Vault and multi-device co-sign both fail fast locally with the same message rather than stalling. Valid SPL sends (recipient account present or missing) are unchanged.

## Screenshots

Pre-fix this path crashed, so there is no real prior screen. **Before** shows how the raw guard error reads in the app's generic error template (i.e. the guard without the localized mapping); **After** shows the localized, token-named message this PR adds. USDC example, on the keysign error screen.

| Before | After |
|--------|-------|
| before | after |

## Testing

- `./gradlew :data:testDebugUnitTest --tests "*SolanaHelperTest*"` — the missing-sender-account case is asserted as a plain JVM test (the guard runs before WalletCore, so no native library is needed).
- `./gradlew :data:ktfmtCheckMain :data:ktfmtCheckTest :app:ktfmtCheckMain`
- `./gradlew :app:compileDebugKotlin`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana signing error handling: users now see a clear, specific message showing the affected token when a required token account can't be found.

* **Localization**
  * Added translated versions of the new error message in German, Spanish, Croatian, Italian, Korean, Dutch, Portuguese, Russian, and Chinese.

* **Tests**
  * Added a test verifying failure behavior when a Solana token transfer is missing the sender token account.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
@@ -5,6 +5,7 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import com.vultisig.wallet.R
 import com.vultisig.wallet.data.api.errors.CosmosBroadcastException
+import com.vultisig.wallet.data.chains.helpers.SOLANA_MISSING_TOKEN_ACCOUNT_PREFIX
 import com.vultisig.wallet.ui.components.errors.ErrorView
 import com.vultisig.wallet.ui.components.errors.ErrorViewButtonUiModel
 import com.vultisig.wallet.ui.utils.UiText
@@ -23,8 +24,11 @@ internal fun KeysignErrorScreen(
             errorLabel = stringResource(R.string.signing_error_transaction_timeout)
         errorMessageString.contains("insufficient funds") ->
             errorLabel = stringResource(R.string.signing_error_insufficient_funds)
-        errorMessageString.contains("sender's associated token account not found") ->
-            errorLabel = stringResource(R.string.signing_error_token_account_not_found)
+        errorMessageString.contains(SOLANA_MISSING_TOKEN_ACCOUNT_PREFIX) -> {
+            val ticker =
+                errorMessageString.substringAfter(SOLANA_MISSING_TOKEN_ACCOUNT_PREFIX).trim()
+            errorLabel = stringResource(R.string.signing_error_token_account_not_found_s, ticker)
+        }
         errorMessageString.contains("failed to calculate bob mid and bob_mic_mc") ->
             errorLabel = stringResource(R.string.signing_error_mixed_reshare)
         errorMessageString.contains(CosmosBroadcastException.BROADCAST_FAILURE_MARKER) -> {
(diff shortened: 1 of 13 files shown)


johnnyluo (approved):
LGTM

PR #4814, opened 2026-06-09
fix(home/chain-detail/staking): pull-to-refresh from anywhere + whole-screen scroll (#4752, #4761)

### LINKED ISSUE #4752: Chain detail: pull-to-refresh doesn't trigger when list is at rest at the top
## Observed
On the chain/token detail screen, pulling down to refresh **does not work when the list is resting at the top**. You have to scroll the content down and then pull, before the refresh gesture registers. A direct pull-down from the resting-top position never triggers the refresh.

## Repro
1. Open any chain/token detail screen.
2. Without scrolling, pull down from the top → nothing (no refresh).
3. Scroll the list, then pull down → refresh fires.

The shared `ScaffoldWithExpandableTopBar` nests a collapsing top bar **inside** the pull-to-refresh container:
`app/src/main/java/com/vultisig/wallet/ui/components/v2/scaffold/ScaffoldWithExpandableTopBar.kt:35-37`
```kotlin
PullToRefreshBox(onRefresh = onRefresh, isRefreshing = isRefreshing) {
    Scaffold(
        modifier = modifier.nestedScroll(scrollBehavior.nestedScrollConnection), // exitUntilCollapsedScrollBehavior
        ...
```
At rest-at-top the downward delta is consumed by the `exitUntilCollapsedScrollBehavior` top-bar nested-scroll connection (re-expanding the bar) before it reaches `PullToRefreshBox`, so the over-scroll that arms the refresh never arrives. Once content is scrolled, the gesture gets through.

Likely fix: order/gate the nested-scroll so PullToRefreshBox sees the top over-scroll first (e.g. only let the top-bar consume when `scrollBehavior.state.heightOffset` isn't already 0 / when the list can actually scroll), or attach the pull-refresh nestedScroll above the scrollBehavior connection.

## Scope
`ScaffoldWithExpandableTopBar` is shared by **chain detail** (`ChainTokensScreen.kt`) and the **home accounts** screen (`VaultAccountsScreen.kt`) — verify both.

## Platform
Android.

### LINKED ISSUE #4761: Staking & chain-detail: scroll the whole screen (don't pin the top card) — mirror iOS
## What
On the LUNA/LUNC staking screen, the top balance card stays **pinned** while only the list below scrolls. It should scroll as part of the whole screen (top card scrolls away), **mirroring iOS**. This should be consistent across screens with this pattern — notably the **chain detail** screen too.

## Where
**Staking positions** — `app/src/main/java/com/vultisig/wallet/ui/screens/cosmosstaking/CosmosStakingPositionsScreen.kt:104-145`:
- The `CosmosStakingBalanceBanner` (balance card, ~L108-115) and the tab row (~L117-143) sit in the outer `Column`, **outside** the `LazyColumn` (L145) — so they're static; only the list scrolls.
- Fix: move the banner (and tab row, if design wants) into the `LazyColumn` as leading `item {}`s so the whole screen scrolls.

**Chain detail** — `ChainTokensScreen.kt` uses `ScaffoldWithExpandableTopBar` (collapsing/pinned top bar). Same desired behavior: header scrolls with content rather than staying pinned.

## Notes
- iOS is the reference (already scrolls whole-screen) — mirror it, no iOS change needed.
- Related: #4752 (pull-to-refresh dead at rest-at-top on chain detail) shares the `ScaffoldWithExpandableTopBar` nested-scroll model — worth fixing together.
- Apply to all screens using this pinned-header pattern, not just these two.
- Platform: Android.

aminsato:
## Summary

Fixes #4752 and fixes #4761 — two closely-related scroll/refresh issues on the screens built around `ScaffoldWithExpandableTopBar`, plus the staking positions screen.

### #4752 — pull-to-refresh works from anywhere

On the home and chain/token-detail screens, pull-to-refresh only fired when the drag **started on the asset list itself**. A pull that began on the balance card, the header, or the empty area below a short list did nothing — so on a sparse chain-detail screen (tiny list, mostly card + empty space) refresh felt broken.

`PullToRefreshBox` only arms from a scrollable child's over-scroll, so three distinct gaps had to be closed in the shared `ScaffoldWithExpandableTopBar` (used by both `VaultAccountsScreen` and `ChainTokensScreen`):

- **Top card** — `VsExpandableTopBar`'s own `draggable` swallowed vertical drags on the card. It's now bridged into the nested-scroll system via a `NestedScrollDispatcher`, so a downward over-drag of the fully-expanded card flows to `PullToRefreshBox`.
- **Empty / non-scrollable content regions** — the Scaffold content is wrapped in a gesture-only `scrollable` (consumes nothing, `overscrollEffect = null`, and — unlike `verticalScroll` — does **not** change layout) so drags on otherwise-dead regions forward into the nested-scroll system. Real list drags are still handled by the list first.
- **Collapsed top bar** — the bar collapses/re-grows entirely through the standard `exitUntilCollapsed` connection; a single source of truth on `heightOffset` (the direct drag writes it; no side `Animatable`/`snapshotFlow` bridge) means a non-monotonic pull can no longer make the bar oscillate. Once the bar is fully expanded it stops consuming and the remaining over-scroll flows up to `PullToRefreshBox` in the same gesture.

### #4761 — scroll the whole screen, don't pin the top card

The header should scroll away as part of the whole screen (mirroring iOS), not stay pinned while only the list moves.

- **Chain detail** — already handled by the `ScaffoldWithExpandableTopBar` nested-scroll rework above: the expanded top bar collapses/scrolls away with the content.
- **Staking positions** (`CosmosStakingPositionsScreen`) — the balance banner and the "Staked" tab row sat in the outer `Column` above the `LazyColumn`, so they stayed pinned. They're now leading `item {}`s inside the `LazyColumn` (and the list went from `weight(1f)` to `fillMaxSize()`), so the top card scrolls away with the rest of the screen.

## Notes for reviewers

- No behavior/visual change to the card, the list, or collapse-on-scroll-up — only otherwise-dead gestures now reach the refresh.
- The gesture-only `scrollable` is intentionally not a `verticalScroll`: it adds gesture detection + nested-scroll dispatch without altering layout or the list's virtualization.
- The staking change is layout-only (banner/tabs become list items); no data, validation, or action wiring changed.

## Test plan

Verified on a device/emulator (home + chain detail). Refresh fires when pulling from:
- the balance card / header area
- the empty space below a short list
- the asset list itself — with the card both **expanded** and **collapsed**

Staking screen: the banner + "Staked" tab now scroll up with the delegation list instead of staying pinned.

Regression checks: list scrolling, card collapse/expand by scroll and by direct drag, and the snackbar overlay all still work. `./gradlew assembleDebug` and `./gradlew :app:ktfmtCheckMain` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved expandable top bar drag/scroll responsiveness with smoother fling handoff to parent scrollers and proper behavior at bounds.
  * Pull-to-refresh now arms reliably from non-scrollable regions by forwarding downward gestures.

* **Refactor**
  * Banner and tab row now scroll together with the main list for a unified scrolling experience.

* **Documentation**
  * Expanded comments clarifying top bar nested-scroll and pull-to-refresh interaction.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/components/v2/scaffold/ExpandableTopBar.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/v2/scaffold/ExpandableTopBar.kt
@@ -3,8 +3,8 @@
 package com.vultisig.wallet.ui.components.v2.scaffold
 
 import androidx.compose.animation.Crossfade
-import androidx.compose.animation.core.Animatable
 import androidx.compose.animation.core.FastOutSlowInEasing
+import androidx.compose.animation.core.animate
 import androidx.compose.animation.core.tween
 import androidx.compose.foundation.gestures.Orientation
 import androidx.compose.foundation.gestures.draggable
@@ -25,9 +25,7 @@ import androidx.compose.runtime.mutableFloatStateOf
 import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
-import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.geometry.Offset
@@ -41,7 +39,6 @@ import androidx.compose.ui.layout.onGloballyPositioned
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.unit.Velocity
 import androidx.compose.ui.unit.dp
-import kotlinx.coroutines.launch
 
 @Composable
 fun VsExpandableTopBar(
@@ -107,34 +104,27 @@ fun VsExpandableTopBar(
             }
         val offsetLimit = remember(heightDiffPx) { -heightDiffPx }
 
-        val animatedOffset = remember { Animatable(0f) }
-
+        // Single source of truth for the bar height: scrollBehavior.state.heightOffset. Both the
+        // direct drag below and the nested-scroll from the list write this same value, so they can
+        // never diverge (no separate Animatable that goes stale when the list scrolls the bar).
         val previousOffsetLimit = remember { mutableFloatStateOf(offsetLimit) }
 
         LaunchedEffect(offsetLimit) {
             val oldLimit = previousOffsetLimit.floatValue
-            val currentOffset = animatedOffset.value
+            val currentOffset = scrollBehavior.state.heightOffset
 
-            animatedOffset.updateBounds(lowerBound = offsetLimit, upperBound = 0f)
+            // Update the limit first so the coercing setter below accepts the rescaled offset.
+            scrollBehavior.state.heightOffsetLimit = offsetLimit
 
             if (oldLimit != 0f && oldLimit != offsetLimit && currentOffset != 0f) {
                 val collapsePercentage = (currentOffset / oldLimit).coerceIn(0f, 1f)
                 val newOffset = (offsetLimit * collapsePercentage).coerceIn(offsetLimit, 0f)
-                animatedOffset.snapTo(newOffset)
+                scrollBehavior.state.heightOffset = newOffset
             } else {
-                val clampedValue = currentOffset.coerceIn(offsetLimit, 0f)
-                if (clampedValue != currentOffset) {
-                    animatedOffset.snapTo(clampedValue)
-                }
+                scrollBehavior.state.heightOffset = currentOffset.coerceIn(offsetLimit, 0f)
             }
 
-            scrollBehavior.state.heightOffsetLimit = offsetLimit
             previousOffsetLimit.floatValue = offsetLimit
-        }
-
-        LaunchedEffect(Unit) {
-            snapshotFlow { animatedOffset.value }
-                .collect { value -> scrollBehavior.state.heightOffset = value }
         }
 
         val offset by remember { derivedStateOf { scrollBehavior.state.heightOffset } }
@@ -154,8 +144,6 @@ fun VsExpandableTopBar(
                 collapsedHeightPx + (expandedHeightPx - collapsedHeightPx) * expandedFraction
             }
 
-        val coroutineScope = rememberCoroutineScope()
-
         // Bridges the bar's own drag gesture into the nested-scroll system. When the bar is at a
         // bound (fully expanded and over-dragged downward), the unused delta is dispatched to the
         // surrounding parents so a downward pull that starts on the top bar reaches the enclosing
@@ -164,14 +152,13 @@ fun VsExpandableTopBar(
         val noOpConnection = remember { object : NestedScrollConnection {} }
 
         val dragState = rememberDraggableState { delta ->
-            val current = animatedOffset.value
+            val current = scrollBehavior.state.heightOffset
             val target = (current + delta).coerceIn(offsetLimit, 0f)
-            val consumed = target - current
-            coroutineScope.launch { animatedOffset.snapTo(target) }
-            val leftover = delta - consumed
+            scrollBehavior.state.heightOffset = target
+            val leftover = delta - (target - current)
             if (leftover != 0f) {
                 nestedScrollDispatcher.dispatchPostScroll(
-                    consumed = Offset(0f, consumed),
+                    consumed = Offset(0f, target - current),
                     available = Offset(0f, leftover),
                     source = NestedScrollSource.UserInput,
                 )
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/vultisig/wallet/ui/components/v2/scaffold/ScaffoldWithExpandableTopBar.kt
```diff
@@ -31,10 +48,77 @@ internal fun ScaffoldWithExpandableTopBar(
...
+                                                easing = FastOutSlowInEasing,
+                                            ),
+                                    ) { value, _ ->
+                                        scrollBehavior.state.heightOffset = value
+                                    }
+                                }
+                        }
+                        // Don't consume: hand the over-scroll to PullToRefreshBox so it can arm.
+                        return Offset.Zero
+                    }
```
johnnyluo: **Confirmed bug: this causes a visible bar jump ("jump back, then jump again").**

`scrollBehavior.state.heightOffset` is the single value that drives the bar's height. Within one pull gesture there are now **two concurrent writers** to it:

1. the live gesture, delegated to `scrollBehavior.nestedScrollConnection` (can re-collapse to a negative offset every frame), and
2. this imperative `animate { … heightOffset = value }` pushing it toward `0`.

A non-monotonic drag toggles between them:
- downward frame (`available.y > 0`, collapsed) → launches the re-expand animation → bar slides **up**;
- a tiny upward frame (`available.y < 0`) → `onPreScroll` cancels the job *and* delegates to `exitUntilCollapsed`, which snaps the offset **negative** again → bar jumps **down**.

Hence the oscillation when the finger jitters or partially reverses mid-pull.

**Suggested direction:** avoid running a side animation on shared scroll state. Try delegating `onPostScroll` to the default connection and letting `exitUntilCollapsed` re-grow the bar naturally — once `heightOffset == 0` it consumes `0` and the leftover over-scroll flows to `PullToRefreshBox` on its own (standard collapsing-toolbar + pull-refresh behavior is sequential: pull re-grows the bar, continued pull arms refresh). If that arms refresh in testing, the `expandJob` / `animate` / `onPostScroll` override can be deleted entirely. If a single delta truly must drive both, route the re-expand through the bar's existing `Animatable` so there is only one writer.

app/src/main/java/com/vultisig/wallet/ui/components/v2/scaffold/ExpandableTopBar.kt
```diff
@@ -150,20 +156,46 @@ fun VsExpandableTopBar(
...
+            coroutineScope.launch { animatedOffset.snapTo(target) }
+            val leftover = delta - consumed
+            if (leftover != 0f) {
+                nestedScrollDispatcher.dispatchPostScroll(
+                    consumed = Offset(0f, consumed),
+                    available = Offset(0f, leftover),
+                    source = NestedScrollSource.UserInput,
+                )
             }
         }
```
johnnyluo, line 166: Related to the jump: `animatedOffset` is the source of truth for *direct* bar drags (`current = animatedOffset.value`), but the new re-expand in `ScaffoldWithExpandableTopBar.onPostScroll` drives `scrollBehavior.state.heightOffset` directly and never updates `animatedOffset`. After a collapsed-state pull re-expands the bar, `animatedOffset.value` is stale, so the next direct drag of the bar can jump from the stale value. Whatever fix is chosen for the re-expand should keep these two in sync (or collapse them to a single source of truth).

johnnyluo (commented):
Confirmed on-device: the re-expand path causes a visible bar jump (oscillation) when a pull is non-monotonic. Root cause and a suggested direction are in the inline comments. Otherwise the approach is solid and well-documented — just the collapsed-state re-expand needs rework before merge.

app/src/main/java/com/vultisig/wallet/ui/components/v2/scaffold/ExpandableTopBar.kt
```diff
@@ -150,20 +156,46 @@ fun VsExpandableTopBar(
...
                         orientation = Orientation.Vertical,
                         state = dragState,
-                        onDragStopped = {
+                        onDragStopped = { velocity ->
+                            // Hand the release to the nested-scroll parents first so a downward
+                            // over-drag of the top bar lets PullToRefreshBox settle / trigger.
+                            val available = Velocity(0f, velocity)
+                            val consumedVelocity =
+                                nestedScrollDispatcher.dispatchPreFling(available)
+                            nestedScrollDispatcher.dispatchPostFling(
```
rkokhatskyi: `onDragStopped`'s new `dispatchPostFling` reaches `exitUntilCollapsed.onPostFling`, whose `settleAppBar` animates `heightOffset` to a snap target before `animatedOffset.animateTo` runs. Because `animatedOffset` is unchanged and still drives `heightOffset` through the `snapshotFlow`, releasing a partially dragged bar settles open then jumps back and re-animates

rkokhatskyi (dismissed):

johnnyluo (approved):
LGTM

PR #4823, opened 2026-06-09
fix(cosmos-staking): stale amounts on entry & no auto-refresh after a tx (#4815)

### LINKED ISSUE #4815: Cosmos staking: positions/Total Staked don't auto-refresh after a tx; plain-text loading (LUNA/LUNC/QBTC)
## Summary

After a Cosmos-SDK staking transaction (delegate / undelegate / redelegate / claim) completes, the shared staking **positions** screen and the **global DeFi balance** don't refresh — the *Total Staked* stays stale until a manual pull-to-refresh. The cold-load state is also plain text rather than a skeleton. Android-side issues in the shared staking screens (LUNA / LUNC / QBTC), independent of any chain's LCD latency.

> The "Unstake/Move forms flash `0 Token`" prefill item originally bundled here is split out to #4822.

## 1. Total Staked / positions don't auto-refresh after a delegate/undelegate/redelegate

After staking to a second validator the new position appears, but the **Total Staked** card (and the global DeFi-page balance) keeps showing the old total until a manual pull-to-refresh.

- `CosmosStakingPositionsViewModel.setData` early-returns when `vaultId`/`chainId` are unchanged, and the screen only calls `setData` from `LaunchedEffect(vaultId, chainId)` — there's no lifecycle-resume refresh, so returning to a retained screen after a signed tx doesn't reload.
- The global DeFi page (`BalanceRepository.loadDeFiAddresses`) runs its remote pass only on `isRefresh`; on a normal open it serves the cached `StakingDetailsRepository` total, which isn't invalidated when a staking tx broadcasts.

Expected: completing a staking tx should invalidate/refresh the staking total so the positions screen and DeFi balance reflect it without a manual pull.

## 2. Plain-text loading state

The cold-load state shows text rather than a skeleton, which reads as "stuck" for the second or two the fan-out takes.

## Repro

1. Stake to one validator, open the positions screen.
2. Stake to a second validator and return.
3. The second position appears, but **Total Staked is unchanged** until pull-to-refresh.

## Scope

Shared Cosmos-staking screens (`app/.../ui/{models,screens}/cosmosstaking/**`). Reproduces on LUNA / LUNC and QBTC. Surfaced out of the QBTC-staking work (#4795 / #4806). Related split-out ticket: #4822 (Unstake/Move prefill).

### LINKED ISSUE #4822: Unstake/Move forms flash "0 Token" before the staked balance loads
## Summary

Opening **Unstake** or **Move** on a Cosmos staking position briefly shows `0 Token` before the real ticker and staked balance appear.

## Details

Tapping Unstake/Move opens the form, which shows `0` + a `Token` placeholder for ~1s, then updates to the actual staked balance and ticker.

Root cause: `Route.CosmosStakingUndelegate` / `Route.CosmosStakingRedelegate` carry only the validator address — not the staked amount or ticker. The view-models initialize `stakedBalance = 0` and `ticker = ""` (the screen falls back to `"Token"`), then asynchronously load the coin and re-fetch the delegation. The tapped position row already holds both the ticker and the staked amount.

Affects all Cosmos staking chains (shared screens); most visible on QBTC because of its slower LCD.

## Suggested direction

Pass the tapped position's ticker + staked amount through the route and prefill the form state so it renders the real values on the first frame; the async load then refreshes them.

Related: #4815 (broader shared-screen staking UX items). Surfaced while testing QBTC staking (#4795 / #4806).

aminsato:
## Summary

Fixes #4815 — the shared Cosmos-SDK staking screens (LUNA / LUNC / QBTC) showed stale or zeroed values on entry and didn't refresh after a transaction completed. These were Android-side issues in the shared screens, independent of any chain's LCD latency.

### 1. Unstake / Move forms no longer flash `0`
`Route.CosmosStakingUndelegate`/`Redelegate` now carry the `stakedAmount` of the tapped position. The undelegate/redelegate VMs seed `stakedBalance` and the amount field on the first frame, so the real value shows immediately. The LCD read still corrects it, but **only overrides when it returns a real delegation**, so a transient blip can't wipe the field back to `0`.

### 2. Total Staked / DeFi balance auto-refresh after a tx
- The positions VM gains `onScreenResumed()`: a no-op until the first load, then a reload on later resumes (returning from a completed delegate / undelegate / redelegate / claim). It re-seeds from the cached snapshot, so it never flashes empty. Wired via `LifecycleEventEffect(ON_RESUME)`.
- After each successful refresh the VM persists the freshly-fetched **delegated + claimable-rewards** total through a new `CosmosStakingDeFiBalanceService.persistStakedTotal(...)`, so the cached DeFi-page read reflects the change on next open without a manual pull. This mirrors the exact sum `getRemoteDeFiBalance` already persists (no extra network).

### 3. Skeleton cold-load state
Replaced the plain-text "Loading positions…" (which read as "stuck") with shimmer **skeleton** cards, and removed the now-orphaned string from all 10 locales. Also extracted a stateless `CosmosStakingPositionsContent` so the loaded/loading states and the skeleton are previewable.

> Scope note: the issue also mentions the DeFi page's `isRefresh`-gated remote pass. That gate is shared across all DeFi chains and outside the "shared Cosmos-staking screens" scope, so rather than changing it, this PR keeps the persisted staking total in sync from the staking screen — the lowest-risk way to make the DeFi balance correct on next open.

## Test plan
- New/updated unit tests:
  - `CosmosUndelegateViewModelTest` / `CosmosRedelegateViewModelTest` — prefill seed, LCD override, and transient-blip retain.
  - `CosmosStakingPositionsViewModelTest` — persists the staked total after a refresh; `onScreenResumed` is a no-op before the initial load.
  - `CosmosStakingDeFiBalanceServiceTest` — `persistStakedTotal` writes the total / no-ops off a staking chain.
- `:app:compileDebugKotlin`, `:app:testDebugUnitTest`, `:data:testDebugUnitTest` green. ktfmt applied.
- Added `@Preview`s for the loaded state, the skeleton loading state, and the skeleton card.

## Manual verification remaining
- Mainnet smoke on LUNA / LUNC / QBTC: stake to a second validator → return, confirm Total Staked updates without a manual pull; tap Unstake/Move → confirm the amount field shows the staked value immediately.

## Summary by CodeRabbit

* **New Features**
  * Staking forms prefill amount and optional token ticker from navigation/deep-links; positions screen now resumes silently and persists computed staked totals; new skeleton placeholders and stateless content component improve loading UX and previews.
* **Bug Fixes**
  * Prefilled staked amount is no longer overwritten by zero/empty remote fetch results.
* **Tests**
  * Added unit tests covering prefill behavior, resume/refresh flows, persistence, and staked-total logic.
* **Chores**
  * Updated navigation payloads, localized staking strings, and added a persistence method with serialized cache writes.


rkokhatskyi (changes requested):
@aminsato I'm going to stop you here, I have the same ticket already in progress but I see you opened the PR for it as well as another ticket https://github.com/vultisig/vultisig-android/issues/4822

Initially I opened one ticket but later split into 2 different issues and I see you fixed both of them conflicting the current work I'm doing with QBTC Staking and literally doing the same what I have in progress.. 

Please refrain from opening the PRs for the tickets you haven't assigned to yourself and haven't started the progress. I asked for the same yesterday in Discord 🤔

app/src/main/java/com/vultisig/wallet/ui/navigation/Navigation.kt
```diff
@@ -462,23 +462,31 @@ internal sealed class Route {
...
         val validatorAddress: String,
+        val stakedAmount: String? = null,
     )
 
     /**
      * Redelegate flow — src is fixed (the current delegation), dst is picked. The 21-day cooldown
-     * gate runs at submit-time inside the VM.
+     * gate runs at submit-time inside the VM. [stakedAmount] is the plain-decimal amount currently
+     * staked at the source validator, carried from the tapped position so the form prefills
+     * instantly instead of flashing `0` while the LCD re-fetch lands (#4815).
```
rkokhatskyi: Unnecessary comments

app/src/main/java/com/vultisig/wallet/ui/navigation/Navigation.kt
```diff
@@ -462,23 +462,31 @@ internal sealed class Route {
      */
     @Serializable data class CosmosStakingDelegate(val vaultId: String, val chainId: String)
 
-    /** Undelegate flow — validator address comes from the active-delegation card. */
+    /**
+     * Undelegate flow — validator address comes from the active-delegation card. [stakedAmount] is
+     * the plain-decimal amount currently staked at that validator, carried from the tapped position
+     * so the form prefills instantly instead of flashing `0` while the LCD re-fetch lands (#4815).
+     */
```
rkokhatskyi: Unnecessary comments

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosUndelegateViewModel.kt
```diff
@@ -294,9 +312,15 @@ constructor(
                     it.validatorAddress == route.validatorAddress &&
                         it.balance.denom == entry.bondDenom
                 }
-            val stakedBalance =
+            val fetchedStakedBalance =
                 matching?.balance?.amount?.toBigDecimalOrNull()?.movePointLeft(nativeCoin.decimal)
                     ?: BigDecimal.ZERO
+            // The LCD read is authoritative when it returns a real delegation; if it yields zero
+            // (no match / transient blip) keep the prefilled hint rather than wiping the field
+            // back to 0 (#4815).
```
rkokhatskyi: Unnecessary comments

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
```diff
@@ -446,6 +487,9 @@ constructor(
                     vaultId = vaultId,
                     chainId = chainId,
                     validatorSrcAddress = position.validatorAddress,
+                    // Prefill the form instantly from the tapped position; the VM's LCD re-fetch
+                    // still corrects it if the staked amount changed (#4815).
```
rkokhatskyi: The comment is right, but excessive since we put the duplicate comments in the file

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
```diff
@@ -409,6 +434,19 @@ constructor(
...
     }
 
+    /**
+     * Re-fetch on screen resume. The first resume is a no-op — `coin` isn't loaded yet and the
+     * initial fetch is already driven by [setData]/[loadCoin]. A later resume means the user
+     * returned from a completed staking action (delegate / undelegate / redelegate / claim), so we
+     * reload to reflect the new Total Staked and positions without a manual pull-to-refresh
+     * (#4815). [refresh] re-seeds from the cached snapshot first, so this never flashes an empty
+     * screen.
+     */
```
rkokhatskyi, line 457: Excessive comment, needs to be shorter

app/src/main/res/values-it/strings.xml
```diff
@@ -1294,7 +1294,6 @@
     <string name="cosmos_staking_continue">Continua</string>
     <string name="cosmos_staking_amount">Importo</string>
     <string name="cosmos_staking_loading_validators">Caricamento validatori…</string>
-    <string name="cosmos_staking_loading_positions">Caricamento posizioni…</string>
```
rkokhatskyi, line 1314: Why are we removing this?

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosRedelegateViewModel.kt
```diff
@@ -135,8 +135,19 @@ constructor(
 
     val amountFieldState: TextFieldState = TextFieldState()
 
+    // Plain-decimal staked amount handed over from the tapped position, so the form shows the real
+    // value on first frame instead of `0` until the LCD re-fetch lands (#4815). null on a direct/
+    // deep-link entry, where we fall back to the network value only.
+    private val prefilledStakedBalance: BigDecimal? =
```
rkokhatskyi: It must be non-nullable

rkokhatskyi (changes requested):

rkokhatskyi:
> @aminsato I'm going to stop you here, I have the same ticket already in progress but I see you opened the PR for it as well as another ticket #4822
> 
> Initially I opened one ticket but later split into 2 different issues and I see you fixed both of them conflicting the current work I'm doing with QBTC Staking and literally doing the same what I have in progress..
> 
> Please refrain from opening the PRs for the tickets you haven't assigned to yourself and haven't started the progress. I asked for the same yesterday in Discord 🤔

@aminsato given you raised the PR on what I had in progress, I went ahead and opened the PR for the work I've done for #4815 fix in the PR https://github.com/vultisig/vultisig-android/pull/4824, check for references. We can check if this PR better solves the problem and potentially add missing parts here so it can be merged

rkokhatskyi:
@aminsato let's finish this PR so it is ready to be merged, I put another reference PR #4824 into draft (should be closed once we merge this PR)

app/src/main/java/com/vultisig/wallet/ui/screens/cosmosstaking/CosmosStakingPositionsScreen.kt
```diff
@@ -89,12 +101,57 @@ internal fun CosmosStakingPositionsScreen(
 
     LaunchedEffect(vaultId, chainId) { viewModel.setData(vaultId = vaultId, chainId = chainId) }
 
+    // Reload when the screen comes back to the foreground — e.g. after returning from a completed
+    // delegate / undelegate / redelegate / claim — so Total Staked and the positions reflect the tx
+    // without a manual pull. The first resume is a no-op in the VM (#4815).
+    LifecycleEventEffect(Lifecycle.Event.ON_RESUME) { viewModel.onScreenResumed() }
```
rkokhatskyi, line 94: `onScreenResumed()` routes every `ON_RESUME` into `refresh()`, which sets `_isRefreshing.value = true` unconditionally, so the pull-to-refresh spinner animates on every app foreground while this screen is retained, not only after a tx, where `TronDeFiPositionsScreen` keeps its resume reload silent and spins only on an explicit pull

app/src/main/java/com/vultisig/wallet/ui/screens/cosmosstaking/CosmosStakingPositionsScreen.kt
```diff
@@ -532,6 +577,85 @@ private fun PositionRow(
...
+            Modifier.fillMaxWidth(widthFraction)
+                .height(height)
+                .clip(RoundedCornerShape(4.dp))
+                .background(brush)
+    )
+}
+
+/**
+ * Horizontally-sweeping shimmer gradient between the card surface and a slightly lighter border
+ * tone. Self-contained so the staking screen doesn't depend on a shared skeleton component (none
```
rkokhatskyi: The `(none exists yet)` claim is wrong: `UiPlaceholderLoader` is the shared DeFi placeholder, already used by the sibling tabs in `StakingTabComponents` and `DeFiComponents`, so this hand-rolls ~70 shimmer lines whose `border.normal` tone also looks inconsistent with the `tertiary_2` loaders used everywhere else

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/cosmos/staking/CosmosStakingDeFiBalanceServiceTest.kt
```diff
@@ -73,6 +73,29 @@ internal class CosmosStakingDeFiBalanceServiceTest {
             assertEquals(BigInteger.valueOf(105), saved.captured.stakeAmount)
         }
 
+    @Test
+    fun `persistStakedTotal writes the supplied total for the chain native coin`() = runTest {
+        // The staking-positions screen hands its freshly-fetched delegated + rewards total here so
+        // the cached DeFi read reflects it on next open (#4815).
+        coEvery { stakingDetailsRepository.getStakingDetailsByCoindId(VAULT_ID, any()) } returns
```
rkokhatskyi, line 82: Both `persistStakedTotal` tests stub `getStakingDetailsByCoindId` to `null`, so only the save branch runs and the `existing.stakeAmount != total -> updateStakingDetails` clear path stays uncovered, yet that is exactly what fires when a full unbond makes the positions VM persist a zero total to drop the stale DeFi balance

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosUndelegateViewModel.kt
```diff
@@ -121,13 +121,31 @@ constructor(
...
+    // deep-link entry, where we fall back to the network value only.
+    private val prefilledStakedBalance: BigDecimal? =
+        route.stakedAmount?.toBigDecimalOrNull()?.takeIf { it > BigDecimal.ZERO }
+
     private val _state =
-        MutableStateFlow(CosmosUndelegateUiState(validatorAddress = route.validatorAddress))
+        MutableStateFlow(
+            CosmosUndelegateUiState(
+                validatorAddress = route.validatorAddress,
+                stakedBalance = prefilledStakedBalance ?: BigDecimal.ZERO,
```
rkokhatskyi: This fixes only half of #4822: the route carries `stakedAmount` so the amount no longer flashes `0`, but it carries no ticker, so `ticker` stays empty until the async coin load and the form still shows the `Token` placeholder (`CosmosUndelegateScreen.kt:64`), most visible on QBTC; carry the position's ticker through the route and seed it here too

rkokhatskyi (dismissed):

app/src/main/res/values-it/strings.xml
aminsato: `cosmos_staking_loading_positions` was only used for the old "Loading positions…" text on the positions screen, which this PR replaces with the shimmer skeleton cards — so the key is now orphaned (no remaining references anywhere in the codebase). Removed it from all 10 locales to keep things clean, matching the repo convention of deleting strings that are exclusively used by removed UI.

Happy to keep it if you'd rather not touch the locale files.

aminsato:
Review feedback addressed in c3585964c:

- **#4822 ticker flash** — `ticker` is now carried through the undelegate/redelegate routes and seeded on the first frame, so the form no longer shows the `Token` placeholder until the async coin load lands.
- **Non-nullable prefill** — `prefilledStakedBalance` is now a non-nullable `BigDecimal` in both VMs, with the LCD-override / init-prefill logic simplified.
- **Resume spinner** — `refresh(silent = true)` on resume reloads without animating the pull-to-refresh spinner; it now spins only on an explicit pull (mirrors TronDeFiPositions).
- **Skeleton** — dropped the hand-rolled shimmer in favour of the shared `UiPlaceholderLoader` used by the other DeFi tabs.
- **Test gap** — `CosmosStakingDeFiBalanceServiceTest` now covers the `updateStakingDetails` clear path (full unbond → zero) and the no-op-when-equal case.
- **Comments** — trimmed the excessive/duplicate comments in the routes and VMs.
- **`cosmos_staking_loading_positions`** — removed because it's orphaned now that the skeleton replaces the loading text (see inline reply); happy to restore if preferred.

`:app:compileDebugKotlin` + `:data:compileDebugKotlin` clean, affected test classes green, ktfmt applied.

app/src/test/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosUndelegateViewModelTest.kt
```diff
@@ -128,6 +130,33 @@ internal class CosmosUndelegateViewModelTest {
...
+    }
+
+    @Test
+    fun `the carried ticker seeds the title on the first frame so it never flashes Token`() =
+        runTest {
+            // #4822: the ticker is carried through the route, so the form shows the real symbol
+            // immediately rather than the "Token" placeholder until the async coin load lands.
+            val model = vm(stakedAmount = "7", ticker = "LUNA")
+            assertEquals("LUNA", model.state.value.ticker)
+        }
```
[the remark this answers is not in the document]
aminsato: The new test matches the assertion style already used throughout this file — every existing test here uses `kotlin.test.assertEquals` / `assertNotNull` (see the surrounding cases). Switching just these two added lines to a Kotest matcher would make them inconsistent with the rest of the file. Migrating the whole file to Kotest is a reasonable cleanup, but it's out of scope for this PR; keeping the new test consistent with its neighbors here.

aminsato:
@coderabbitai proceed

app/src/main/java/com/vultisig/wallet/ui/screens/cosmosstaking/CosmosStakingPositionsScreen.kt
aminsato: Fixed in c3585964c. `refresh()` now takes a `silent` flag; `onScreenResumed()` calls `refresh(silent = true)`, which skips `_isRefreshing.value = true`, so the resume reload runs silently. The spinner only animates on an explicit pull-to-refresh (`onRefresh = viewModel::refresh`), matching TronDeFiPositions. Added a test (`onScreenResumed reloads silently without animating the pull-to-refresh spinner`) asserting `isRefreshing` stays false on resume while the reload still fires.

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosUndelegateViewModel.kt
aminsato: Fixed in c3585964c. `ticker` is now carried through both `Route.CosmosStakingUndelegate` and `Route.CosmosStakingRedelegate`, set from `_state.value.ticker` in `CosmosStakingPositionsViewModel.unstake/move`, and seeded into the initial state here (`ticker = route.ticker.orEmpty()`), so the form shows the real symbol on the first frame instead of the `Token` placeholder; the async coin load still corrects it. Added `the carried ticker seeds the title on the first frame so it never flashes Token` to both VM tests.

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/cosmos/staking/CosmosStakingDeFiBalanceServiceTest.kt
aminsato: Covered in c3585964c. Added `persistStakedTotal updates the cache when a full unbond drops the total to zero` — it stubs `getStakingDetailsByCoindId` to return an existing non-zero stake, then calls `persistStakedTotal(..., BigInteger.ZERO)` and asserts `updateStakingDetails` fires with `stakeAmount == 0` while `saveStakingDetails` is never called. Also added a no-op-when-equal case so both branches of `persistStakedBalance` are exercised.

app/src/main/java/com/vultisig/wallet/ui/screens/cosmosstaking/CosmosStakingPositionsScreen.kt
aminsato: Fixed in c3585964c. Dropped the hand-rolled shimmer (`ShimmerBar` + `rememberShimmerBrush`, ~70 lines, `border.normal` tone) and rebuilt `StakingPositionSkeleton` from the shared `UiPlaceholderLoader` — same `tertiary_2` placeholder the sibling DeFi tabs use — so it's visually consistent. Also removed the now-unused animation / `Brush` / `Offset` imports.

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosRedelegateViewModel.kt
aminsato: Fixed in c3585964c. `prefilledStakedBalance` is now a non-nullable `BigDecimal` (defaults to `ZERO` when the route carries no amount), in both this VM and CosmosUndelegateViewModel; the LCD-override and init-prefill logic were simplified accordingly.

app/src/main/java/com/vultisig/wallet/ui/navigation/Navigation.kt
aminsato: Fixed in c3585964c — restored the concise one-line KDoc on the route; the added `ticker`/`stakedAmount` params are self-documenting.

app/src/main/java/com/vultisig/wallet/ui/navigation/Navigation.kt
aminsato: Fixed in c3585964c — restored the concise KDoc on the route.

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosUndelegateViewModel.kt
aminsato: Fixed in c3585964c — trimmed the comment down to the essential point.

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
aminsato: Fixed in c3585964c — removed the duplicated comment; kept a single one in `unstake()` and dropped it from `move()`.

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
aminsato: Fixed in c3585964c — shortened the `onScreenResumed` KDoc.

app/src/test/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosUndelegateViewModelTest.kt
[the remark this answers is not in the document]
aminsato: As noted on the other thread: the new test matches the `kotlin.test` assertion style already used throughout this file. Migrating the whole file to Kotest is reasonable but out of scope for this PR; keeping the added test consistent with its neighbors.

aminsato:
Addressed the reward-total nitpick from the latest review in 4b5b9df87.

`rewardsBaseUnits` in `CosmosStakingPositionsViewModel.refresh()` now floors each bond-denom reward coin to whole base units before summing — the exact floor-then-sum shape `CosmosStakingDeFiBalanceService.claimableRewards()` uses — instead of summing each validator's rewards as `BigDecimal` and flooring once. This guarantees the total persisted from the positions screen matches what `getRemoteDeFiBalance` writes, so the cached DeFi balance can't drift by a base unit on fractional rewards.

app/src/main/res/values-it/strings.xml
rkokhatskyi: Good to remove 👍

app/src/main/java/com/vultisig/wallet/ui/models/cosmosstaking/CosmosStakingPositionsViewModel.kt
```diff
@@ -390,6 +396,40 @@ constructor(
...
+                            .flatMap { it.reward }
+                            .filter { it.denom == bondDenom }
+                            .fold(BigInteger.ZERO) { acc, rewardCoin ->
+                                acc +
+                                    (rewardCoin.amount.toBigDecimalOrNull()?.toBigInteger()
+                                        ?: BigInteger.ZERO)
+                            }
+                    this@CosmosStakingPositionsViewModel.vaultId?.let { vid ->
+                        runCatching {
+                            cosmosStakingDeFiBalanceService.persistStakedTotal(
```
rkokhatskyi, line 425: `persistStakedTotal` writes the same `StakingDetails` row that `BalanceRepository.getDeFiTokenValue` serializes behind a per-address `Mutex` (`BalanceRepository.kt:384`), yet takes no lock, so a concurrent DeFi-page `getRemoteDeFiBalance` can interleave its read-modify-write and persist a stale staked total

aminsato: Good catch — fixed in 8e0c17d1d. `persistStakedBalance` now wraps its read-decide-write in a per-`(vaultId, coin)` `Mutex` held inside the service, so both write paths serialize against each other: the DeFi page's `getRemoteDeFiBalance` (which `BalanceRepository` already calls under its per-address lock) and the staking screen's `persistStakedTotal` (which bypasses that lock). No deadlock risk — the service mutex is only ever acquired innermost, never around the `BalanceRepository` lock.

johnnyluo:
CleanShot 2026-06-10 at 20 12 25@2x
Please address this

aminsato:
> Please address this

@johnnyluo 
Fixed in https://github.com/vultisig/vultisig-android/pull/4835

johnnyluo:
resolve the conflict

app/src/main/java/com/vultisig/wallet/ui/screens/cosmosstaking/CosmosStakingPositionsScreen.kt
```diff
@@ -114,6 +169,18 @@ internal fun CosmosStakingPositionsScreen(
                             tab { VsTab(label = stringResource(tab.displayNameRes), onClick = {}) }
                         }
                     }
+                    V2Container(
```
rkokhatskyi: The `#4825` merge changed this `Row` to a `Box` and added the comment saying the inline edit button `used to sit here`, but the `V2Container` button below survived the merge, so it duplicates the top-bar manage-positions action and overlaps the `VsTabGroup` in the now single-slot `Box`

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/cosmosstaking/CosmosStakingPositionsScreen.kt
aminsato: Fixed in 836703cee while resolving the `main` conflict. The conflict here was exactly this: `HEAD` still had the leftover `V2Container` edit button inside the now single-slot `Box`, while #4825/#4761 on `main` had already moved the banner + tab row into the scrolling `LazyColumn` as `item`s and dropped that duplicate button. I took `main`'s side, so the duplicate manage-positions action and the `VsTabGroup` overlap are both gone — the only manage-positions control is the ChainDashboard top-bar action. Also removed the now-unused `V2Container`/`ContainerType`/`CornerType`/`UiIcon`/`clickOnce` imports.

rkokhatskyi (approved):
LGTM

PR #4824, opened 2026-06-09
Refresh Cosmos staking positions after a tx; skeleton cold-load

### LINKED ISSUE #4815: Cosmos staking: positions/Total Staked don't auto-refresh after a tx; plain-text loading (LUNA/LUNC/QBTC)

rkokhatskyi:
Fixes #4815.

After a Cosmos-SDK staking transaction (delegate / undelegate / redelegate / claim) the shared positions screen and the global DeFi total stayed stale until a manual pull, and the cold-load state was plain text.

## Changes

- The positions screen reloads on every resume via `LifecycleResumeEffect` (matching `TronDeFiPositionsScreen`), so returning from a signed tx refreshes the Total Staked card and the delegations list without a pull. The reload is seeded from the in-memory snapshot and runs silently — the pull-to-refresh spinner is reserved for explicit pulls, the first load, and the resume refresh stay silent.
- On each successful load the screen persists the staked total (delegated + bond-denom rewards, base units) to `StakingDetailsRepository`, which is the cache the portfolio/DeFi page reads. `AccountsRepository.loadDeFiAddresses` only hits the network on an explicit refresh, so without this the global Total Staked kept serving the pre-tx value. The total uses the same definition as `CosmosStakingDeFiBalanceService`, mirroring how `CircleDeFiPositionsViewModel` already persists its position.
- The cold-load state renders skeleton delegation cards (`UiPlaceholderLoader`, as on the other DeFi tabs) instead of a "Loading positions…" line. The now-unused string was removed from all locales.
- Extracted a stateless `CosmosStakingPositionsContent` (mirroring `TronDeFiPositionsScreenContent`) so each state can be previewed and tested.

Applies to LUNA / LUNC / QBTC since these screens are shared.

## Screenshots

Cold-load state (same TerraClassic / LUNC mock):

| Before | After |
|--------|-------|
| before | after |

Populated state, rendered through the extracted content (layout unchanged):

loaded

## Testing

- `./gradlew :app:testDebugUnitTest --tests "*CosmosStakingPositionsViewModelTest"` — passes, with new coverage for the resume re-fetch, the silent (non-spinner) resume refresh, and the persisted staked total.
- `./gradlew :app:assembleDebug` — built; the screenshots above were captured from `PreviewActivity`.

Lint was not run locally; CI covers it.

## Summary by CodeRabbit

* **New Features**
  * Staking positions display animated skeleton card loaders during data loading for enhanced visual feedback
  * Total staked amounts automatically persist across app sessions to maintain accurate portfolio totals

* **Improvements**
  * Staking screen data now automatically refreshes when you return to the screen
  * Pull-to-refresh spinner only displays during manual refreshes, not for background updates
  * Improved loading state with skeleton cards instead of text messages

* **Localization**
  * Updated loading state terminology from "positions" to "validators" across all supported languages

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -39,6 +39,7 @@ import androidx.compose.ui.unit.sp
 import com.vultisig.wallet.R
 import com.vultisig.wallet.data.blockchain.cosmos.qbtc.claim.QbtcClaimBlockedReason
 import com.vultisig.wallet.data.blockchain.cosmos.qbtc.claim.QbtcClaimError
+import com.vultisig.wallet.data.blockchain.cosmos.staking.CosmosStakePositionRow
 import com.vultisig.wallet.data.models.Address
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coins
@@ -70,6 +71,7 @@ import com.vultisig.wallet.ui.models.ChainTokensUiModel
 import com.vultisig.wallet.ui.models.TransactionDetailsUiModel
 import com.vultisig.wallet.ui.models.TransactionScanStatus
 import com.vultisig.wallet.ui.models.VerifyTransactionUiModel
+import com.vultisig.wallet.ui.models.cosmosstaking.CosmosStakingPositionsUiState
 import com.vultisig.wallet.ui.models.cosmosstaking.CosmosStakingVerifyUiState
 import com.vultisig.wallet.ui.models.cosmosstaking.CosmosStakingVerifyValidatorRow
 import com.vultisig.wallet.ui.models.deposit.DepositFormUiModel
@@ -92,6 +94,7 @@ import com.vultisig.wallet.ui.models.swap.ValuedToken
 import com.vultisig.wallet.ui.models.swap.VerifySwapUiModel
 import com.vultisig.wallet.ui.models.toNetworkUiModel
 import com.vultisig.wallet.ui.screens.TransactionDoneView
+import com.vultisig.wallet.ui.screens.cosmosstaking.CosmosStakingPositionsContent
 import com.vultisig.wallet.ui.screens.cosmosstaking.CosmosStakingVerifyContent
 import com.vultisig.wallet.ui.screens.deposit.BondFormContent
 import com.vultisig.wallet.ui.screens.keygen.FastVaultVerificationScreen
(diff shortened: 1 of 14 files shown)


rkokhatskyi:
Will put this into Draft for now, @aminsato opened PR for the same earlier (I had this fixed locally so pushed anyway for reference). Another PR: https://github.com/vultisig/vultisig-android/pull/4823, we can check what else needs to be added there so it becomes mergeable 

This PR to be closed once #4823 is merged

rkokhatskyi:
Closing in favour of #4823, let's have it finished given it also includes the fix for #4822 issue 

cc: @aminsato

Issue #4481, opened 2026-05-13
[BUG] Vault upgrade silently calls online VultiServer instead of peer-discovery when server share is self-hosted on another device


realpaaao:
# GG20→DKLS upgrade on Android hits the online VultiServer API even when the server share has been self-hosted on another peer device — orphaning the self-hosted share and producing a new online-server vault without user consent

## Problem

Cross-platform parity issue surfaced from a Discord support thread (2026-05-13). iOS exhibits the same bug, filed as https://github.com/vultisig/vultisig-ios/issues/4348 with the full reproduction trace.

Android's migrate routing uses the identical heuristic — branch on `vault.isFastVault()` — and therefore inherits the same flaw: any vault with a `server-*` signer in its signer list is treated as \"online VultiServer must be invoked for migrate\", regardless of whether the user has imported that server share into another device they own.

## Where

`app/src/main/java/com/vultisig/wallet/ui/screens/migration/MigrationOnboardingViewModel.kt:54-69`:

\`\`\`kotlin
fun upgrade() {
    viewModelScope.launch {
        val vault = vaultRepository.get(vaultId) ?: error(\"Vault with $vaultId doesn't exist\")
        if (vault.isFastVault()) {
            navigator.route(Route.Migration.Password(vaultId = vaultId))          // → online VultiServer migrate
        } else {
            navigator.route(
                Route.Keygen.PeerDiscovery(action = TssAction.Migrate, …)         // → peer migrate
            )
        }
    }
}
\`\`\`

`vault.isFastVault()` follows the same name-suffix rule as iOS (`signer.startsWith(\"server-\")`). Android has no notion of \"this server share is self-hosted on another device\" — same blind spot.

## Reproduction (mirror of the iOS case)

1. Original vault: signers = `[Android-*, server-*]`, threshold 2.
2. User imported the `server-*` `.vult` into a second device (any Vultisig app) at some point.
3. From Android, tap **Upgrade now**.
4. App routes to `Route.Migration.Password` → password screen → server migrate over the public API.
5. Online VultiServer joins and produces new DKLS shares for Android + online server. The imported server share on the other device is left orphaned in GG20.

Expected: either a user prompt asking whether the server share is online or self-hosted, or peer-discovery-first with API fallback.

## Fix (mirror iOS — same options apply)

[...]

PR #4828, opened 2026-06-09
Raise QBTC staking gas limit to fix redelegate out-of-gas

### LINKED ISSUE #4820: QBTC staking redelegate fails with out-of-gas (gas limit too low for ML-DSA)
## Summary

A QBTC staking **redelegate (Move)** submitted from the DeFi tab fails on-chain with an out-of-gas error.

## Details

- Failing tx on qbtc-testnet: `67B85E1C2EE74480495B330DCB2D07383BDC988039A458E45B2A7AE327BCC109`
- Result: `code: 11` (`sdk`), `raw_log: "out of gas in location: ReadPerByte; gasWanted: 400000, gasUsed: 400832: out of gas"`
- The `MsgBeginRedelegate` is otherwise valid (correct delegator, src/dst valoper, amount/denom) — signing and broadcast work; only the gas budget is too low.

## Root cause

The QBTC staking gas limit is `400000` (`CosmosStakingConfig`), borrowed from the Terra entry. QBTC signs with **ML-DSA**, whose signature is ~2.4 KB; the per-byte read cost on top of `MsgBeginRedelegate`'s dual-record `x/staking` write exceeds `400000`. Delegate / undelegate / claim are lighter single-msg paths and succeed at `400000`; redelegate is the heaviest and OoGs.

## Suggested direction

Raise the QBTC staking gas limit (≈ `800000`). `qbtc-testnet` reports `min_gas_price: 0` and `block.max_gas: -1`, so a larger budget neither raises the fee (it stays at the `800` `min_tx_fee` floor) nor risks a block-limit OoG.

Surfaced while testing QBTC staking (#4795 / #4806).

rkokhatskyi:
Fixes #4820.

A QBTC staking **redelegate (Move)** fails on-chain with out-of-gas.

- Failing tx: https://explorer.qbtc.net/qbtc/tx/67B85E1C2EE74480495B330DCB2D07383BDC988039A458E45B2A7AE327BCC109
- `code: 11`, `raw_log: "out of gas in location: ReadPerByte; gasWanted: 400000, gasUsed: 400832: out of gas"` (verified against the live LCD)

## Root cause

The QBTC staking gas limit was `400000`, borrowed from the Terra entry. QBTC signs with **ML-DSA** (~2.4 KB signature), and `MsgBeginRedelegate` is the heaviest single-message path — the dual-record `x/staking` write plus the per-byte read of that large signature pushes gas just over the floor (400832 > 400000). Delegate / undelegate / claim are lighter single-message paths and fit; redelegate doesn't.

The gas model is per-message (`gasLimit = entry.gasLimit × msgCount`), so redelegate (1 msg) gets exactly the base limit while batched claims already scale.

## Change

Raise the QBTC staking gas limit `400000 → 800000` (~2× the observed burn). It's free here: the live qbtc-testnet node reports no minimum gas price (`minimum_gas_price: ""`) and enforces a flat `800` `min_tx_fee`, and `block.max_gas` is `-1`, so the larger budget moves neither the fee (stays `800`) nor any block limit. Batched claims keep the same headroom via the per-message multiplier.

## Cross-platform note

iOS (#4481) and Windows have no native QBTC staking gas config that matches Android's — iOS's `CosmosStakingConfig` ships QBTC at `gasLimit: 300_000`, which is *lower* than the old Android value and would OoG on redelegate as well (a latent iOS bug). Android intentionally diverges here with the empirically-verified `800000`; iOS should be aligned separately. The gas *model* (per-msg × count, redelegate = heaviest single msg) is identical across platforms.

## Success Redelegate Tx sample

https://explorer.qbtc.net/qbtc/tx/97B4D0562378A3F25D6B3485888F0DD694EB5049371C4F8A059220EE1D627943

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased gas limits for QBTC staking transactions to prevent out-of-gas failures during redelegate operations.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/staking/CosmosStakingConfig.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/staking/CosmosStakingConfig.kt
@@ -69,7 +69,9 @@ object CosmosStakingConfig {
                     feeAmount = 133_333_334L,
                     unbondingDays = 21,
                 ),
-            // Values verified against the live qbtc-testnet LCD; mirrors iOS #4481.
+            // Chain identity (chainId / denom / valoper / unbonding) mirrors iOS #4481; the gas
+            // and fee are tuned to the live qbtc-testnet instead — iOS's 300_000 gas OoGs on
+            // redelegate (see below) and its 7_500 fee overpays the verified 800 min_tx_fee.
             Chain.Qbtc to
                 Entry(
                     chainId = "qbtc-testnet",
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

Issue #4685, opened 2026-05-30
Swap routing: add banded provider preference (1% band); SwapKit slots in when integrated


realpaaao:
## Goal

Add the **banded provider preference** layer to Android swap routing. Part of converging all core apps onto one rule — canonical spec in **vultisig/vultisig-sdk#605**.

Android already picks best output: `SwapQuoteManager.fetchBestQuote()` returns `successes.maxBy { it.result.estimatedDstFiat.value }` (SwapQuoteManager.kt:296-301). What's missing is the priority band — a 0.01%-better provider always wins, so we never tilt toward our economics on a near-tie.

## Change

`app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt` (~line 296-301)

Replace the plain `maxBy` with banded selection:

1. `best` = max `estimatedDstFiat.value` (unchanged).
2. Among successes with `estimatedDstFiat.value >= best * 0.99`, return the one highest in priority order:
   `THORChain > MayaChain > SwapKit > Kyber > 1inch > LiFi`.
3. `BAND = 0.01` (1%).

Add a provider priority map. `SwapProviderTable` sets are unordered (`Set`) — the band map is the source of truth for order, so no change needed there for ranking.

## SwapKit note

SwapKit is **not integrated on Android** (zero refs in codebase). Per decision, **ship the band now** using available providers — order is effectively `THORChain > MayaChain > Kyber > 1inch > LiFi` until SwapKit lands. SwapKit integration is ongoing/tracked separately; once merged, add `SWAPKIT` to the priority map in the 3rd slot. No need to block this issue on it.

## Tests

- near-tie (within 1%): higher-priority provider wins.
- outside band (>1% better): better-rate provider wins (ETH->stable: aggregator beats THORChain).

## Decisions

- Rule + band (1%) locked across platforms; canonical spec: vultisig/vultisig-sdk#605.
- Metric anchor is iOS (net, gas-excluded); Android's `estimatedDstFiat` is the close equivalent — keep it.

Owner: Android (Amin / Roman).

PR #4830, opened 2026-06-10
fix(swap): surface "no route" guidance on token selection instead of at quote time (#4710)

### LINKED ISSUE #4710: Swap: surface pair support before quote (filter dst tokens / mark unsupported) instead of failing at quote time
## Summary

The swap form filters selectable tokens by **chain** support, but not by **pair** support. A user can pick a source and destination token, wait for the quote, and only then hit a dead end — *"Swap is not supported for this pair."* For chains with thin provider coverage (e.g. Sui, which currently routes only via SwapKit's NEAR/Mayan), this is easy to trip and reads as a failure rather than guidance.

## Current behavior

- **Chain level is pre-filtered.** `SwapTokenSelector` shows only swap-enabled chains (`Chain.isSwapSupported`, `Route.SelectNetwork.Filters.SwapAvailable`).
- **Pair level is resolved lazily.** After both tokens are chosen, `SwapFormViewModel` calls `SwapQuoteRepository.getEligibleProviders(src, dst)`; when it returns empty it throws `SwapException.SwapIsNotSupported("Swap is not supported for this pair")` — so unsupported pairs are only discovered at quote time.
- **Eligibility is a static table.** `getEligibleProviders` delegates to `SwapProviderTable.eligibleProvidersFor(src, dst)`, a hardcoded routing table — not driven by the live provider/token data behind the `swapkit-a` proxy.

## Proposed change (UX only)

When a source token is selected, indicate which destinations are actually routable, so the "not supported" outcome is shown **before** the user commits to a quote. Options, smallest-first:

1. **Reuse existing feedback earlier** — keep the full dst list but resolve `getEligibleProviders` on selection and show an inline "no route for this pair" state instead of an error after quoting.
2. **Filter the dst list** to tokens with at least one eligible provider for the chosen source.

Either is a meaningful improvement over the current late failure; (1) is the lower-risk first step.

## Open design question (not prescribed here)

The eligibility source is static today. Windows derives swap-enabled chains dynamically from the SDK's pinned union, and #4685 moves routing in a more data-driven direction. Whether pair eligibility should likewise be driven by live provider/token data (vs. the hand-maintained `SwapProviderTable`) is a design decision worth settling before picking the implementation — it determines whether "supported pairs" stays in sync as providers change.

## Relationship to existing work

- Part of the SwapKit epic surface — companion to #4675 (chain coverage incl. Sui) and #4685 (routing). This issue is strictly the **pre-quote pair feedback UX**, which neither currently covers.

## Acceptance

- [ ] Selecting an unsupported destination for the chosen source shows guidance before a quote is requested (no `SwapIsNotSupported` thrown as a user-visible failure path).
- [ ] Behavior verified on a thin-coverage chain (e.g. Sui -> an unsupported counterparty).
- [ ] Design question above resolved or explicitly deferred with a note.

aminsato:
## Summary

The swap form filtered selectable tokens by **chain** but not by **pair**. A user could pick a source + destination, type an amount, wait for the quote, and only then hit a dead end — *"Swap route not available"* — because `getEligibleProviders(src, dst)` returned empty and the pipeline threw `SwapException.SwapIsNotSupported` at quote time. On thin-coverage chains (e.g. Sui), or a non-Thor/Maya ERC-20 swapped to a THORChain-only chain, this is easy to trip and reads as a failure rather than guidance.

This implements the issue's **option 1** (the lower-risk first step): resolve provider eligibility **up front on token selection** and surface the existing guidance immediately, before any quote is requested.

## What changed

`SwapFormViewModel` only (no UI, no new strings):

- Added an `isPairSupported` flag, resolved by `updatePairSupport(src, dst)` on every pair change.
- When the selected pair has no eligible provider, the `swap_route_not_available` guidance shows **the moment the destination is selected** (red flip button + warning icon, swap disabled) — no amount required, no spinner.
- The indicative-rate `onEach` and the quote `collectLatest` early-return for an unroutable pair, so the pipeline path that throws `SwapIsNotSupported` is **never reached** as a user-visible failure.
- Switching back to a routable pair clears the guidance immediately (ahead of the debounced quote that would otherwise clear it).
- Same-token pairs are treated as supported, so the hint never flashes mid-pick.

## Design decision

Eligibility stays driven by the static `SwapProviderTable` (a local lookup, so this is instant). Moving "supported pairs" to live provider/token data is a separate, larger change — **explicitly deferred to #4685**.

## Test plan

**Unit** — 3 new tests in `SwapFormViewModelTest.kt` (full class: 89 tests, 0 failures):
- guidance appears on selection before any amount is entered
- no quote is ever fetched for an unsupported pair, even after typing an amount
- a routable pair shows no guidance

`:app:compileDebugKotlin` green, ktfmt applied.

**Manual (device)** — unsupported pair: **VULT → ATOM (Cosmos)** (`{1INCH, LIFI, KYBER, SWAPKIT}` ∩ `{THORCHAIN}` = ∅):
- [ ] Selecting ATOM as destination shows "Swap route not available" instantly, before typing an amount; swap disabled; no spinner.
- [ ] Typing an amount keeps the guidance and never spins or late-errors.
- [ ] Control pair **VULT → BTC** (share SWAPKIT) still quotes normally.
- [ ] Switching the destination from ATOM back to BTC clears the guidance immediately.

Closes #4710

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap pair validation to display route-unavailable errors immediately upon selection, eliminating unnecessary loading states and quote-fetch attempts for unsupported token pairs.

* **Tests**
  * Added test coverage for swap pair eligibility validation and error messaging.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -134,6 +134,14 @@ constructor(
 
     private var refreshQuoteJob: Job? = null
 
+    // Whether the currently selected source/destination pair has any eligible swap provider.
+    // Resolved up front on every pair change (#4710) so an unroutable pair surfaces guidance the
+    // moment it is selected and never reaches the quote pipeline (which would throw
+    // SwapIsNotSupported only after the user typed an amount and waited out the debounce).
+    private var isPairSupported = true
+
+    private val pairNotSupportedError = UiText.StringResource(R.string.swap_route_not_available)
+
     private data class PreFlipState(
         val srcAmount: String,
         val srcTokenId: String,
@@ -656,7 +664,7 @@ constructor(
                     src to dst
                 }
                 .distinctUntilChanged()
-                .onEach {
+                .onEach { (src, dst) ->
                     // A freshly selected pair has no quote yet, and a token switch never clears the
                     // previous pair's destination value. Reset it so the skeleton shows while the
                     // new quote loads instead of the stale amount reading as a firm quote for the
@@ -670,6 +678,7 @@ constructor(
                                 )
                         )
                     }
+                    updatePairSupport(src, dst)
                 }
                 .combine(amountChanges) { address, immediate ->
                     QuoteInput(address = address, amount = srcAmount, immediate = immediate)
@@ -680,6 +689,9 @@ constructor(
                 // the debounce and an instant indicative estimate fills the destination field while
                 // we wait, without flashing on background refreshes (#4712).
                 .onEach { input ->
+                    // Unroutable pair: the "no route" guidance already showed on selection (#4710),
+                    // so don't spin or fetch an indicative estimate for a pair we can't quote.
+                    if (!isPairSupported) return@onEach
                     isLoading = true
                     showIndicativeRate(input)
                 }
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

Issue #4831, opened 2026-06-10
QBTC staking: edit icon overlaps with text label


aminsato:
## Description

On the QBTC staking screen (Cosmos staking positions), a small action icon in the top-left area overlaps with its own text label. The label is clipped/truncated and the edit (pencil/swap) icon sits on top of the text, making it unreadable.

See the red square in the screenshot below.

## Screenshot

![bug](https://i.imgur.com/0zE1JK4.png)

## Steps to Reproduce

1. Open the app and go to a vault containing QBTC
2. Open the QBTC staking screen (DeFi → QBTC staking positions)
3. Observe the small icon/badge below the Balance card (top-left)

## Expected

The icon and its text label should be laid out without overlapping, and the label should be fully visible (not clipped/truncated).

## Actual

The icon overlaps the text label, and the label is truncated to a single character ("d…"), making it unreadable.

## Environment

- Screen: QBTC / Cosmos staking positions
- Platform: Android

PR #4825, opened 2026-06-09
Remove duplicate manage-positions button on DeFi positions screens

rkokhatskyi:
Fixes #4821.

The Cosmos, Tron, Maya and Thorchain DeFi positions screens each register a `ChainDashboard` top-bar action **and** rendered an inline edit-chains button next to the Staked tab — two controls that open the same position-selection dialog. This removes the inline button on all four so the top-bar action is the single manage-positions control, which is also what `BaseDeFiPositionsScreen` (Circle etc.) does when its inline button is gated off.

The ticket described this as a Cosmos-only issue, but the same duplicate exists on Tron/Maya/Thorchain (each is a custom screen that registers the top-bar action and draws its own inline button), so all four are aligned here.

## Changes

- Removed the inline `edit_chain` button (and its now-unused imports) from `CosmosStakingPositionsScreen`, `TronDeFiPositionsScreen`, `MayachainDefiPositionsScreen`, `ThorchainDefiPositionsScreen`.
- Tron's content no longer needs the `onEditPositionClick` param (only the inline button used it) — dropped. Maya/Thorchain keep it (their empty-state "Manage positions" still uses it).
- The `edit_chain` drawable stays (still used by `BaseDeFiPositionsScreen`, `HomePageTabMenu`, `ChainTokensTabMenuAndSearchBar`).

## Screenshots

Cosmos staking positions (representative; the inline edit button beside the Staked tab is removed — the top-bar manage action, host chrome not shown in this content render, is unchanged):

| Before | After |
|--------|-------|
| before | after |

## Testing

- `./gradlew :app:compileDebugKotlin` — builds.
- `./gradlew :app:assembleDebug` — built; screenshots captured from `PreviewActivity`.

No behavior change beyond removing the duplicate control; signing is untouched.

## Release Notes

* **Refactor**
  * Streamlined the layout of DeFi position screens (Cosmos, Mayachain, Thorchain, and Tron) by consolidating position management controls to a single, unified location in the top action bar, removing redundant inline controls for a cleaner interface.

PR #4835, opened 2026-06-10
test(staking): add @Preview for QBTC/Cosmos staking positions screen

aminsato:
## Summary

Adds Compose `@Preview`s for the QBTC / Cosmos staking positions screen (`CosmosStakingPositionsScreen.kt`) so it can be rendered in Android Studio without Hilt.

The screen was ViewModel-coupled (took `hiltViewModel()`), so the body is extracted into a stateless `CosmosStakingPositionsContent(...)` that accepts `state` + plain callbacks. The VM-bound `CosmosStakingPositionsScreen` now just collects state and forwards callbacks — **no behavior change**.

Two previews:
- **Empty state** — the issue #4831 screen (balance banner + "Staked" tab + "Delegate to New Validator" + empty-positions hint). This documents the **post-#4825** layout: the inline edit-chains icon that used to overlap the "Staked" label was removed in #4825, so the row now renders the tab alone.
- **One active delegation** — populated card (avatar, Active badge, staked amount/fiat, APY, next award, action buttons) with the Claim button shown.

## Context on #4831

The overlap reported in #4831 was already resolved on `main` by #4825 (which removed the duplicate inline manage-positions icon). These previews make that screen renderable in-IDE and lock in the corrected layout.

## Test plan
- [x] ktfmt applied
- [x] IDE diagnostics clean (no new warnings/errors)
- [ ] Open `CosmosStakingPositionsScreen.kt` in Android Studio split view to render both previews

## Summary by CodeRabbit

* **Tests**
  * Added debug preview routes for Cosmos QBTC staking positions, covering both empty and populated state scenarios for development and testing purposes.

## Preview render

"One active delegation" preview rendered in Android Studio:

one active delegation


rkokhatskyi (changes requested):
Preview screens should not exist in production code, if you want to add it you should use `app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt` similar to what was added in #4806 PR

aminsato:
@rkokhatskyi good catch — moved the previews out of the main source set. The two `@Preview` composables + their mock data were compiled into the production APK, which isn't right.

Done in e27da80e5:
- Dropped the `@Preview` functions + preview host from `CosmosStakingPositionsScreen`, keeping only the stateless `CosmosStakingPositionsContent` extraction (now `internal` so debug can reach it).
- Re-added the empty + one-active-delegation previews in `app/src/debug/.../PreviewActivity.kt`, wired into the dispatcher as `cosmos_staking_positions_empty` / `cosmos_staking_positions_delegation` — same approach as #4806.

`:app:compileDebugKotlin` passes.

rkokhatskyi (approved):
LGTM

PR #540, opened 2024-07-06
 Fix Incorrect Backup Name Display

aminsato:
Closes #495

PR #4838, opened 2026-06-10
feat(swap): carry swap-fee coin context for correct co-signer fiat

### LINKED ISSUE #4834: [Add] Wire up commondata swap-fee coin-context fields on OneInchTransaction (swap_fee_chain / swap_fee_token_id / swap_fee_decimals)
## Problem

`OneInchTransaction` in the keysign proto carries three optional coin-context fields for the swap fee — added in `commondata` PR #87 (`feat(1inch_swap_payload): carry coin context for swap_fee`, merged):

```proto
optional string swap_fee_chain    = 8;
optional string swap_fee_token_id = 9;
optional int32  swap_fee_decimals = 10;
```

Android already pins its `commondata` git submodule at the PR #87 merge commit (`data/build.gradle.kts` compiles `${rootDir}/commondata/proto`), so the generated `OneInchTransaction` class **already has** `swapFeeChain` / `swapFeeTokenId` / `swapFeeDecimals`. But Android code references **none** of them — both mappers only touch `swapFee`.

This matters because the existing `swap_fee` field only carries an **amount** with no coin context, and different aggregators denominate the affiliate fee in different coins:

| Provider | Fee coin |
|----------|----------|
| LI.FI / 1inch / SwapKit | source chain's native fee coin (e.g. ETH, 18 decimals) |
| **KyberSwap** | **destination token (e.g. USDC, 6 decimals)** |

A consumer holding only the `KeysignPayload` (notably the **co-signer** in a Secure Vault join flow, which has no live quote) must currently guess the fee coin. Guessing "source fee coin" for a KyberSwap payload misreads a 6-decimal USDC amount as an 18-decimal ETH amount and overshoots the fiat value by ~10¹². Without these fields Android cannot render a correct fiat swap-fee value on the co-signer path.

## Solution

Wire up the already-available PR #87 fields in the two proto mappers:

1. **Populate** `swapFeeChain` / `swapFeeTokenId` / `swapFeeDecimals` alongside the existing `swapFee` when building the proto in `PayloadToProtoMapper.kt` (`OneInchTransaction(...)`), for each general-swap provider — using the fee coin's chain, token id (empty/omitted for native fee coins), and decimals.
2. **Read** the three fields when mapping back in `KeysignPayloadProtoMapper.kt` and use them to reconstruct the correct fee coin so the co-signer renders the right fiat value, matching the initiator.

No submodule bump or proto regen is required — the fields are already present in the pinned `commondata`. vultisig-windows already consumes these fields on the read side (`core/ui/mpc/keysign/tx/swap/getSwapFeeFromPayload.ts`) and serves as the behavioral reference: when `swap_fee_chain` is empty or decimals are missing (payloads from senders that predate the proto extension), it returns no swap fee rather than guessing.

### Must Do
- [ ] Populate `swapFeeChain` / `swapFeeTokenId` / `swapFeeDecimals` in `PayloadToProtoMapper.kt` when constructing `OneInchTransaction`, for each general-swap provider
- [ ] Read the three fields in `KeysignPayloadProtoMapper.kt` and reconstruct the correct fee coin for fiat display
- [ ] Backward-compat: payloads without these fields (empty `swap_fee_chain` / missing decimals) fall back gracefully instead of guessing a wrong coin

### Must NOT Do
- Don't change existing behavior for payloads that already work
- Don't bump the commondata submodule (PR #87 is already pinned)

### Out of Scope
- Broader fee-breakdown unification (network + total + tier-discount) beyond the swap-fee coin context
- Any initiator-side gas / network-fee value changes

## Acceptance Criteria
- [ ] A KyberSwap swap (fee in destination token, e.g. USDC 6 decimals) shows the correct fiat swap-fee value on the **co-signer** device, matching the initiator
- [ ] A 1inch / LI.FI / SwapKit swap (fee in source native coin) still shows the correct fiat swap-fee value
- [ ] A payload that lacks the new fields (pre-update sender) decodes without error and does not render a wildly inflated fee

aminsato:
Closes #4834

## Problem

`OneInchTransaction` carries three optional coin-context fields for the swap fee — `swap_fee_chain` / `swap_fee_token_id` / `swap_fee_decimals` (commondata #87, already pinned). Android referenced none of them: both proto mappers only touched `swapFee`.

The bare `swap_fee` amount is ambiguous because aggregators denominate the affiliate fee in different coins:

| Provider | Fee coin |
|---|---|
| 1inch / LI.FI / SwapKit | source chain's native fee coin (e.g. ETH, 18 dec) |
| **KyberSwap** | **destination token (e.g. USDC, 6 dec)** |

A consumer holding only the `KeysignPayload` — notably the **co-signer in a Secure Vault join flow**, which has no live quote — had to guess the fee coin. Guessing "source native" for a KyberSwap payload misreads a 6-decimal USDC amount as 18-decimal native and renders the swap-fee fiat off by a factor of ~10¹².

This is a **cross-platform** issue: the desktop/extension initiator (vultisig-sdk #540, merged) now carries the **raw** destination-token fee plus this context, so an Android co-signer that ignores the context renders a wrong (near-zero) fee. vultisig-windows' `getSwapFeeFromPayload.ts` is the behavioral reference.

## Approach

Stamp the context **where the fee coin is actually known** — `SwapQuoteManager`, from the resolved `feeCoin` — keeping the amount and its coin context in lockstep. Deriving it in the proto mapper would be wrong: the live tx already carries a native-converted `swapFee` amount against a stale destination-token contract, so the mapper would emit inconsistent context and break other platforms' co-signers. The join-flow co-signer reads the context and reconstructs the fee coin, falling back to the previous per-provider heuristic when a sender predates the fields.

## Changes

- **`OneInchSwapTxJson`** — carry `swapFeeChain` + `swapFeeDecimals` (the existing `swapFeeTokenContract` doubles as the token id).
- **`SwapQuoteManager`** — `withResolvedSwapFee(feeAmount, feeCoin)` stamps the context at both pre-conversion sites (Kyber, LiFi/Jupiter).
- **`PayloadToProtoMapper` / `KeysignPayloadProtoMapper`** — write/read the three proto fields (the read path also restores the previously-dropped `swapFeeTokenContract`).
- **`JoinKeysignViewModel`** — resolve the fee coin from the context (match against dst/src/native), fall back to the heuristic when chain is empty / decimals are null.
- Add `KeysignPayloadProtoMapperSwapFeeContextTest` (decode, backward-compat, native→null, round-trip).

Backward compatibility: payloads without the fields decode to empty/null and fall through to the existing behavior — no change for senders that predate the extension. No commondata submodule bump (already pinned).

Bonus correctness: the LI.FI co-signer now uses the initiator's exact discount-aware fee (prior ~0.5% gap gone); Jupiter SPL-token-source swaps resolve the fee coin correctly.

## Test plan

- [x] `:data` + `:app` swap/keysign/mapper unit suites green; new round-trip test green
- [x] ktfmt applied

Manual (multi-device, cross-platform — the only way to exercise it, since Android pre-normalizes its own payloads):

1. Initiate a **KyberSwap** same-chain EVM swap on the **desktop/extension** (e.g. Polygon USDC → WMATIC, where the route resolves to Kyber and Thor/Maya are not eligible). Note the swap-fee fiat.
2. Co-sign on **Android**.
   - **Before:** co-signer swap fee shows ~$0 (misreads dest-token amount as native).
   - **After (this PR):** co-signer swap fee matches the initiator.
3. Native-fee provider (1inch / LI.FI / SwapKit) co-sign still shows a correct, sane fee.
4. A pre-update payload decodes without error and does not render a wildly wrong fee.

Note: iOS does not yet consume these fields — that's a separate parity follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced EVM swap fee handling to improve accuracy in fee estimation and display by explicitly tracking fee coin context (chain, token contract, and decimals).

* **Tests**
  * Added comprehensive test suite to validate swap fee context mapping and ensure consistency in fee calculations across different swap providers.


rkokhatskyi (approved):
LGTM

johnnyluo (approved):
LGTM

PR #4860, opened 2026-06-11
fix(evm): widen non-swap EIP-1559 maxFee headroom to base×1.5 + tip

### LINKED ISSUE #4853: Non-swap EIP-1559 maxFee headroom only base×1.2+tip; stalls on base-fee spike
## Observed

On the non-swap EVM send path, the EIP-1559 `maxFeePerGas` is computed with only `baseFee × 1.2 + tip` of headroom. If the base fee rises more than ~20% between fee estimation and inclusion (a common occurrence during a busy window or across the MPC review+sign delay), the tx's `maxFeePerGas` is below the prevailing base fee and it stalls in the mempool until the base fee drops.

## Cause

`EthereumFeeService.calculateMaxFeePerGas()` (`data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt:141-146`):

```kotlin
private fun calculateMaxFeePerGas(baseNetworkPrice, maxPriorityFeePerGas): BigInteger {
    return baseNetworkPrice.increaseByPercent(20).add(maxPriorityFeePerGas)
}
```

The extra base-fee bump only exists for the *swap* path — `calculateBaseNetworkPrice` (lines 148-165) multiplies the base by 1.5 (Ethereum) / 1.1 (others) **only when `swap == true`**. For plain transfers, `baseNetworkPrice` passes through unscaled, so the ceiling is just `base × 1.2 + tip`. Across consecutive EIP-1559 base-fee increases (max +12.5% per block), a few blocks of delay can exceed the 20% margin.

## Fix shape

- Apply a larger base-fee multiplier to the non-swap send path as well (e.g. `base × 1.5–2.0 + tip`), or size headroom off the recent base-fee trend rather than a flat 20%. The user's actual paid amount stays near `base + tip` under EIP-1559, so a higher ceiling only improves inclusion robustness without overcharging.

---

aminsato:
Closes #4853

## Problem
On the non-swap EVM send path, the EIP-1559 `maxFeePerGas` was computed as only `base × 1.2 + tip`. EIP-1559 lets the base fee rise up to **12.5% per block**, so a flat 20% margin is exhausted after **~1.5 blocks**. If the base fee climbs between fee estimation and inclusion — common during a busy window or across the MPC review + sign delay — the tx's `maxFeePerGas` falls below the prevailing base fee and it stalls in the mempool until the base fee drops.

The extra base-fee bump previously existed **only** on the swap path (`calculateBaseNetworkPrice` bumps ×1.5 ETH / ×1.1 others when `swap == true`); plain transfers passed the base through unscaled.

## Fix
`EthereumFeeService` now bumps the committed base **25%** on the non-swap path, so with the existing +20% in `calculateMaxFeePerGas` the broadcast ceiling lands at **`base × 1.5 + tip`** (~3 consecutive max base-fee increases of headroom). The swap path is untouched.

Raising the ceiling does **not** overcharge the user: under EIP-1559 the amount actually paid stays near `base + tip` — `maxFeePerGas` is only the cap, so this purely improves inclusion robustness.

## Cross-platform note
iOS (`FeeService.swift`) computes the non-swap ceiling identically (`base × 1.2 + tip`), so this is a **shared** tight ceiling rather than an Android-only divergence. iOS should get the same change for parity — flagging for follow-up, not silently diverging.

## Design choice
Picked **×1.5** (conservative end of the ×1.5–2.0 range suggested in the issue) to roughly double the headroom while minimizing displayed fee-bond inflation. Bumping to ×1.8 (matching the ETH swap ceiling) or ×2.0 is a one-line constant change if preferred.

## Test plan
- [x] Updated `EthereumFeeServiceTest` assertion from `base×1.2` to `base×1.5 + tip` (networkPrice 125, maxFee 155 gwei for base 100 / tip 5)
- [x] `:data:testDebugUnitTest --tests EthereumFeeServiceTest` — all green
- [x] ktfmt clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated base fee calculation for non-swap Ethereum transfers. The base network price now includes a 25% increase, affecting maximum fee per gas estimates for these transactions.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
@@ -161,7 +161,14 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
             val bumpPercent = if (chain == Chain.Ethereum) 50 else 10
             return baseNetworkPrice.increaseByPercent(bumpPercent)
         }
-        return baseNetworkPrice
+        // Non-swap transfers: bump the committed base so the broadcast ceiling
+        // (calculateMaxFeePerGas adds a further 20%) lands at baseFee × 1.5 + priorityFee
+        // instead of baseFee × 1.2 + priorityFee. EIP-1559 raises the base fee up to 12.5%
+        // per block, so a flat 20% margin is exhausted after ~1.5 blocks and a send can
+        // stall in the mempool when the base fee climbs across the MPC review + sign window.
+        // The extra headroom only improves inclusion robustness — the amount actually paid
+        // stays near baseFee + priorityFee under EIP-1559.
+        return baseNetworkPrice.increaseByPercent(NON_SWAP_BASE_FEE_BUMP_PERCENT)
     }
 
     private suspend fun calculateMaxPriorityFeePerGas(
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4864, opened 2026-06-11
fix(evm): align ERC-20 signed gas limit with displayed fee bond

### LINKED ISSUE #4857: ERC20 gas-limit floor diverges (signed 150k vs displayed 210k); fee bond != signed limit
## Observed

For ERC-20 transfers, the gas limit that is **displayed** to the user (the fee bond) and the gas limit that is **signed** into the transaction are computed in two independent places with different default floors. When the on-chain estimate is low, the displayed fee is based on a 210k floor while the tx is signed with a 150k floor — so the fee the user consents to does not match the limit actually committed.

## Cause

Two divergent computations of the same ERC-20 transfer limit:

1. **Signed limit** — `BlockChainSpecificRepository.getSpecific()` (`data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt:200-215`): estimate `× 1.5`, floored at `DEFAULT_TOKEN_TRANSFER_LIMIT` (150k, line 174):

```kotlin
.estimateGasForERC20Transfer(...).increaseByPercent(50)
...
val gasLimitFee = gasLimit ?: max(defaultGasLimit, estimateGasLimit)  // defaultGasLimit = 150k
```

2. **Displayed fee** — `EthereumFeeService` (`data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt:69-79`): estimate `× 1.5`, but floored at `getDefaultLimit` = `DEFAULT_TOKEN_TRANSFER_LIMIT.increaseByPercent(40)` (210k, line 247):

```kotlin
.estimateGasForERC20Transfer(...).increaseByPercent(50)
return maxOf(calculatedLimit, getDefaultLimit(transaction))  // floor 150k × 1.4 = 210k
```

The `× 1.5` on the estimate matches, but the floors (150k vs 210k) differ, so the displayed and signed limits diverge whenever the estimate falls below 210k.

## Fix shape

- Compute the ERC-20 gas limit once (single source of truth — e.g. have `BlockChainSpecificRepository` consume the limit from the fee service, or share a helper), so the displayed fee bond and the signed `gasLimit` are guaranteed identical.
- Align the two default floors at minimum if a full refactor is out of scope.

---

aminsato:
Closes #4857

## Problem
For ERC-20 transfers the gas limit was computed in two independent places with **different default floors**:

| | Location | ERC-20 floor |
|---|---|---|
| **Signed** limit | `BlockChainSpecificRepository.getSpecific()` | `DEFAULT_TOKEN_TRANSFER_LIMIT` = **150k** |
| **Displayed** fee bond | `EthereumFeeService.getDefaultLimit()` | `DEFAULT_TOKEN_TRANSFER_LIMIT.increaseByPercent(40)` = **210k** |

Both apply the same `estimate × 1.5`, so whenever the estimate falls below 210k (the common case — a plain ERC-20 transfer estimates ~45–65k) the floor decides the limit, and the two diverge: the user is shown a **210k**-based fee but the tx is signed with a **150k** limit. The fee consented to does not match the limit committed.

## Fix
- Extract the 210k floor into a single shared constant `DEFAULT_TOKEN_TRANSFER_LIMIT_WITH_MARGIN` in `EthereumFeeService`, and have `BlockChainSpecificRepository` use it as the ERC-20 default. The signed `defaultGasLimit` `when` now mirrors `getDefaultLimit` exactly for every case (Arbitrum 160k, native 23k, swap 600k, ERC-20 210k) — one source of truth, can't drift.
- Aligning **up** to 210k matches the value already displayed to the user and gives the tx more headroom. Gas limit is only a cap — actual cost is gas *used* × price — so this does not increase what the user pays.

## Router-deposit guard
`gasLimitFee = max(defaultGasLimit, estimateGasLimit)` applies the floor to **everything**, so naively raising the ERC-20 default to 210k would have silently bumped the deliberate THORChain router-deposit hardcode (`THORCHAIN_ROUTER_DEPOSIT_GAS_LIMIT` = 200k) up to 210k via `max()`. Short-circuited it:

```kotlin
val gasLimitFee = gasLimit ?: routerDepositGasLimit ?: max(defaultGasLimit, estimateGasLimit)
```

Router deposits now use their 200k headroom verbatim — unchanged behavior.

## Test plan
- [x] Updated `BlockChainSpecificRepositoryImplTest`: bare ERC-20 (75k estimate) floor assertion 150k → 210k; router-deposit assertion stays 200k with a clarified comment
- [x] `:data:testDebugUnitTest` full suite — green
- [x] EVM/fee `:app` tests (`*GasFee*`, `*EvmActualFee*`, `*SwapTransactionBuilder*`, `*SwapGasCalculator*`, `*DefaultSendStrategy*`, `*SelectGasFee*`) — green
- [x] ktfmt clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized ERC-20 transfer gas limit calculations to ensure consistency between displayed fee estimates and signed transactions.
  * Improved router deposit gas limit handling to prevent unnecessary fee adjustments.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
@@ -244,7 +244,7 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
             transaction is Swap -> DEFAULT_SWAP_LIMIT
             chain == Chain.Arbitrum -> DEFAULT_ARBITRUM_TRANSFER
             transaction is Transfer && transaction.coin.isNativeToken -> DEFAULT_COIN_TRANSFER_LIMIT
-            transaction is Transfer -> DEFAULT_TOKEN_TRANSFER_LIMIT.increaseByPercent(40)
+            transaction is Transfer -> DEFAULT_TOKEN_TRANSFER_LIMIT_WITH_MARGIN
             else -> error("Transaction type not supported: ${transaction::class.simpleName}")
         }
     }
@@ -274,6 +274,12 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
         val DEFAULT_SWAP_LIMIT = "600000".toBigInteger()
         val DEFAULT_COIN_TRANSFER_LIMIT = "23000".toBigInteger()
         val DEFAULT_TOKEN_TRANSFER_LIMIT = "150000".toBigInteger()
+
+        // ERC-20 transfer gas-limit floor: the 150k base plus a 40% safety margin. Single
+        // source of truth shared with BlockChainSpecificRepository so the signed gasLimit and
+        // the displayed fee bond use the same floor and cannot drift (issue #4857).
+        val DEFAULT_TOKEN_TRANSFER_LIMIT_WITH_MARGIN =
+            DEFAULT_TOKEN_TRANSFER_LIMIT.increaseByPercent(40)
 
         val DEFAULT_ARBITRUM_TRANSFER = "160000".toBigInteger()
         val DEFAULT_MANTLE_SWAP_LIMIT = "3000000000".toBigInteger()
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4870, opened 2026-06-11
fix(keysign): bound join-keysign wait so a joiner can't spin forever

### LINKED ISSUE #4856: Joiner waitForKeysignToStart polls with no timeout; spins forever if initiator abandons
## Observed

A device that joins a keysign session polls for the initiator to start the ceremony with **no timeout**. If the initiator abandons the keysign (closes the app, loses connectivity), the joining device polls the relay every second indefinitely, holding the user on a spinner with no failure/abort, until they manually back out or the ViewModel is destroyed.

## Cause

`JoinKeysignViewModel.waitForKeysignToStart()` (`app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt:1676-1700`):

```kotlin
while (isActive) {
    try {
        if (checkKeygenStarted()) { ...; return@withContext }
    } catch (e: KeysignMessagesException) { ...; return@withContext }
    delay(1000)
}
```

The loop exits only on success, a `KeysignMessagesException`, or coroutine cancellation — there is no elapsed-time bound. A non-erroring but never-starting session loops forever.

## Fix shape

- Add a deadline (e.g. track start time and break with a `JoinKeysignError.Timeout` after N seconds/minutes), surfacing a "session didn't start — try again" state.
- Drive the timeout off the relay's session-keepalive if one exists, so a genuinely slow-but-alive initiator isn't cut off prematurely.

---

aminsato:
Closes #4856

## Problem
`JoinKeysignViewModel.waitForKeysignToStart()` polled the relay every second with **no elapsed-time bound** — the loop exited only on success, a `KeysignMessagesException`, or coroutine cancellation:

```kotlin
while (isActive) {
    try { if (checkKeygenStarted()) { ...; return@withContext } }
    catch (e: KeysignMessagesException) { ...; return@withContext }
    delay(1000)
}
```

If the initiator abandons the ceremony (closes the app, loses connectivity), `checkCommittee` keeps returning a committee that never includes this party, so the joining device holds the user on a spinner **indefinitely** with no failure or abort until they manually back out.

## Fix
- Extract the polling loop into a small, testable top-level `awaitKeysignStart(timeout, pollInterval, checkStarted)` returning a `KeysignStartOutcome` (`Started` / `FailedToPrepare` / `TimedOut`), wrapped in `withTimeoutOrNull(2.minutes)` and polling every `1.second`. Mirrors the existing `JoinKeygenViewModel.DiscoveryTimeout` pattern.
- New retryable `JoinKeysignError.Timeout`. The error View's existing branch already renders any error with a **Try Again** button; `tryAgain()` routes `Timeout` through the same re-join path as `RelayUnavailable` (and the timeout handler resets `isJoiningKeysign` so the retry isn't blocked).
- `KeysignMessagesException` moved from a private nested class to a file-level declaration so the extracted helper can reference it (behavior-preserving).
- Added `join_keysign_start_timeout` across all 10 locales.

## Cross-platform note
iOS (`JoinKeysignViewModel.swift`) has the **same** unbounded loop (`while status == .WaitingForKeysignToStart`, `Task.sleep(1s)`), so this is a shared gap rather than an Android divergence — flagging for an iOS follow-up.

## Design choice
2-minute deadline: long enough not to cut off a slow-but-alive initiator, and a false timeout recovers gracefully because **Try Again** just re-polls the still-valid session. Easy to tune via the named constant.

## Test plan
- [x] New `AwaitKeysignStartTest` — `Started` immediately, poll-then-`Started`, and `TimedOut` bounded by the deadline (asserted via `runTest` virtual `currentTime`)
- [x] Full `*keysign*` `:app` unit-test suite — green
- [x] ktfmt clean

## Summary by CodeRabbit

* **New Features**
  * Improved keysign start polling with clearer timeout handling and distinct error outcomes so users receive precise feedback when a signing session fails to begin.

* **Localization**
  * Added timeout error message translations in German, Spanish, Croatian, Italian, Korean, Dutch, Portuguese, Russian, Chinese, and English.

* **Tests**
  * Added tests covering immediate start, polling-to-start, timeout, and distinct failure scenarios to ensure accurate user-facing behavior.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
@@ -192,7 +192,10 @@ private class MissingQbtcClaimAccountException(chain: Chain) :
     Exception("Missing ${chain.raw} account for QBTC claim co-sign")
 
 /** Raised when the messages to sign cannot be prepared once the ceremony starts. */
-private class KeysignMessagesException(message: String) : Exception(message)
+internal class KeysignMessagesException(message: String) : Exception(message)
+
+/** Raised when polling the relay for committee membership fails (network/relay error). */
+internal class KeysignCheckException(message: String) : Exception(message)
 
 /** How [awaitKeysignStart] finished polling for the initiator to start the ceremony. */
 internal sealed interface KeysignStartOutcome {
@@ -202,6 +205,9 @@ internal sealed interface KeysignStartOutcome {
     /** Preparing the messages to sign failed; carries the reason for the error state. */
     data class FailedToPrepare(val message: String) : KeysignStartOutcome
 
+    /** Polling the relay for the committee failed; carries the reason for the error state. */
+    data class FailedToCheck(val message: String) : KeysignStartOutcome
+
     /** The deadline elapsed without the ceremony starting (initiator likely abandoned). */
     data object TimedOut : KeysignStartOutcome
 }
@@ -209,8 +215,11 @@ internal sealed interface KeysignStartOutcome {
 /**
  * Polls [checkStarted] every [pollInterval] until it reports the ceremony has started, bounded by
  * [timeout]. Without the bound a joiner spins forever when the initiator abandons the keysign
- * (issue #4856). A [KeysignMessagesException] thrown by [checkStarted] ends the wait with
- * [KeysignStartOutcome.FailedToPrepare]; exceeding [timeout] yields [KeysignStartOutcome.TimedOut].
+ * (issue #4856). Only a plain `false` ("not started yet") keeps the loop running. A
+ * [KeysignMessagesException] ends the wait with [KeysignStartOutcome.FailedToPrepare] and a
+ * [KeysignCheckException] with [KeysignStartOutcome.FailedToCheck], so a real poll failure surfaces
+ * immediately instead of being masked by the timeout; exceeding [timeout] yields
+ * [KeysignStartOutcome.TimedOut].
  */
 internal suspend fun awaitKeysignStart(
     timeout: Duration,
@@ -227,10 +236,14 @@ internal suspend fun awaitKeysignStart(
                 return@withTimeoutOrNull KeysignStartOutcome.FailedToPrepare(
                     e.message ?: "Failed to prepare messages to sign"
                 )
+            } catch (e: KeysignCheckException) {
+                return@withTimeoutOrNull KeysignStartOutcome.FailedToCheck(
+                    e.message ?: "Failed to check keysign start"
+                )
             }
             delay(pollInterval)
         }
-        @Suppress("UNREACHABLE_CODE") KeysignStartOutcome.TimedOut
+        error("unreachable") // the loop only exits via return@withTimeoutOrNull
     } ?: KeysignStartOutcome.TimedOut
 
 /** Bounds how long a joiner waits for the initiator to start the ceremony before failing. */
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -182,6 +191,52 @@ sealed class JoinKeysignError(val message: UiText) {
...
+                    return@withTimeoutOrNull KeysignStartOutcome.Started
+                }
+            } catch (e: KeysignMessagesException) {
+                return@withTimeoutOrNull KeysignStartOutcome.FailedToPrepare(
+                    e.message ?: "Failed to prepare messages to sign"
+                )
+            }
+            delay(pollInterval)
+        }
+        @Suppress("UNREACHABLE_CODE") KeysignStartOutcome.TimedOut
```
johnnyluo: Nit (non-blocking): this trailing expression is genuinely unreachable — the `while (true)` loop only exits via `return@withTimeoutOrNull`, and the actual timeout is handled by the `?: KeysignStartOutcome.TimedOut` elvis below. The `@Suppress("UNREACHABLE_CODE")` is a sign the code is fighting the compiler. Consider replacing it with `error("unreachable")` to make intent explicit, or restructuring so the timeout result has a single source of truth. Purely cosmetic.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -182,6 +191,52 @@ sealed class JoinKeysignError(val message: UiText) {
...
+    pollInterval: Duration,
+    checkStarted: suspend () -> Boolean,
+): KeysignStartOutcome =
+    withTimeoutOrNull(timeout) {
+        while (true) {
+            try {
+                if (checkStarted()) {
+                    return@withTimeoutOrNull KeysignStartOutcome.Started
+                }
+            } catch (e: KeysignMessagesException) {
```
johnnyluo, line 235: Non-blocking observation about which exceptions are terminal here. `awaitKeysignStart` only treats `KeysignMessagesException` as a terminal outcome (`FailedToPrepare`). But `checkKeygenStarted()`'s generic `catch (e: Exception)` sets `currentState = Error(FailedToCheck)` and returns `false` — so a transient `checkCommittee` network failure surfaces the error state, yet this loop keeps polling. If the failure never recovers, the state later gets overwritten by `Error(Timeout)`, so the user sees the timeout message instead of the original network error.

This mirrors the old loop's behavior (it also continued on `false`), so it's not a regression introduced by this PR. Flagging as a possible follow-up to decide whether transient check failures should end the wait distinctly rather than being masked by the timeout.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -182,6 +191,52 @@ sealed class JoinKeysignError(val message: UiText) {
...
+                return@withTimeoutOrNull KeysignStartOutcome.FailedToPrepare(
+                    e.message ?: "Failed to prepare messages to sign"
+                )
+            }
+            delay(pollInterval)
+        }
+        @Suppress("UNREACHABLE_CODE") KeysignStartOutcome.TimedOut
+    } ?: KeysignStartOutcome.TimedOut
```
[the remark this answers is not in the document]
aminsato: Good catch — fixed in fa1208624. `checkKeygenStarted()` now throws a dedicated `KeysignCheckException` on a real `checkCommittee()` failure (instead of setting the error state and returning `false`). `awaitKeysignStart()` catches it and returns a new `KeysignStartOutcome.FailedToCheck`, which `waitForKeysignToStart()` maps to `Error(FailedToCheck)`. Only a genuine "not started yet" `false` now stays inside the bounded retry loop, so a relay failure surfaces immediately rather than being masked by the timeout. Added a test asserting the failure returns after a single poll with zero elapsed time.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
aminsato: Agreed — replaced the `@Suppress("UNREACHABLE_CODE")` trailing expression with `error("unreachable")` plus a comment noting the loop only exits via `return@withTimeoutOrNull`. Done in fa1208624.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
aminsato: Addressed in fa1208624 — this is the same concern CodeRabbit raised. Transient `checkCommittee` failures no longer continue the poll: `checkKeygenStarted()` throws `KeysignCheckException`, which `awaitKeysignStart()` turns into the new `KeysignStartOutcome.FailedToCheck` terminal outcome, so the user sees the original check failure instead of the timeout message.

johnnyluo (approved):
LGTM

PR #4871, opened 2026-06-11
fix(tokens): dedup chain tokens by Coin.id to stop duplicate LazyColumn key crash

### LINKED ISSUE #4869: crash: duplicate LazyColumn key "cbETH-Base" in token list (dedup vs key mismatch)
## Summary

The app crashes with a fatal `IllegalArgumentException` when scrolling a token list that contains the same token twice. Observed key: **`cbETH-Base`** (cbETH on Base appears twice in the list).

```
FATAL EXCEPTION: main
java.lang.IllegalArgumentException: Key "cbETH-Base" was already used.
If you are using LazyColumn/Row please make sure you provide a unique key for each item.
    at androidx.compose.ui.layout.LayoutNodeSubcompositionsState.subcompose(SubcomposeLayout.kt:1592)
    at androidx.compose.foundation.lazy.LazyListState.onScroll$foundation(LazyListState.kt:549)
    ...
```

The crash fires during a fling/scroll, so it only manifests once the duplicate row is measured.

## Root cause

The LazyColumn item key and the de-duplication key are derived from **different** fields, so duplicates slip past dedup but collide on the key.

- `Coin.id` (used as the LazyColumn key) is `"${ticker}-${chain.id}"`:
  - `data/src/main/kotlin/com/vultisig/wallet/data/models/Coin.kt:20-21`
- `GetChainTokensUseCase.emitUniqueTokens` de-duplicates by `(contractAddress, chain.id)`, **not** `(ticker, chain.id)`:
  - `data/src/main/kotlin/com/vultisig/wallet/data/usecases/chaintokens/GetChainTokensUseCase.kt:88`
    ```kotlin
    .distinctBy { it.contractAddress to it.chain.id }
    ```

When the same ticker is sourced with two different contract addresses on the same chain (e.g. a built-in/refreshed `cbETH-Base` plus a searchable/provider `cbETH-Base` with a different contract address), both survive `distinctBy`, then produce the identical key `"cbETH-Base"`.

Affected lists that key on `token.id`:
- `app/src/main/java/com/vultisig/wallet/ui/screens/select/SelectAssetScreen.kt:91`
- `app/src/main/java/com/vultisig/wallet/ui/screens/v2/chaintokens/ChainTokensScreen.kt:292`
- `app/src/main/java/com/vultisig/wallet/ui/components/v2/fastselection/SelectAssetPopup.kt:49`

The merge dedup in `SelectAssetViewModel` (`SelectAssetViewModel.kt:157-161`) only removes overlap between `filteredAssets` and `additionalAssets` by `token.id`; it does not collapse duplicates that already exist within a single source list coming from `GetChainTokensUseCase`.

## Steps to reproduce

1. Open the token/asset selection list on a chain where a token (cbETH on Base) is present from more than one source.
2. Scroll the list.
3. App crashes (`SIG: 9`).

## Expected

The token list never contains two entries with the same LazyColumn key; cbETH on Base appears once. No crash on scroll.

## Suggested fix direction

Make the dedup key consistent with the list key. Either:
- De-duplicate by `Coin.id` (i.e. `(ticker, chain.id)`) in `GetChainTokensUseCase.emitUniqueTokens`, or
- Make the LazyColumn keys unique on the actual unique field (contract address) everywhere the list keys on `token.id`.

The two keys must agree so dedup and the LazyColumn key can never disagree.

## Environment

- App: com.vultisig.wallet (debug)
- Android Compose foundation LazyColumn

aminsato:
Closes #4869

## Problem
The app crashes with a fatal `IllegalArgumentException: Key "cbETH-Base" was already used` while scrolling a token list, because the LazyColumn key and the de-dup key are derived from **different** fields:

- The list keys each row on `Coin.id` = `"${ticker}-${chain.id}"` (`SelectAssetScreen`, `ChainTokensScreen`, `SelectAssetPopup`).
- `GetChainTokensUseCase.emitUniqueTokens` deduped only by `(contractAddress, chain.id)`.

So when the same ticker is sourced with two different contract addresses on one chain (a built-in/refreshed `cbETH-Base` plus a searchable/provider `cbETH-Base`), both survive `distinctBy`, then collide on the identical key `"cbETH-Base"` and crash the LazyColumn on scroll.

## Fix
Add a final `.distinctBy { it.id }` after the existing contract-based collapse, so the de-dup key matches the LazyColumn key exactly and the emitted list can never contain two coins with the same key. The contract-based `distinctBy` is kept, so the only behavior change is collapsing the previously-crashing duplicate.

Built-in/refreshed tokens are flattened **first** in every `emitUniqueTokens(...)` call, so the canonical entry wins over the provider duplicate.

## Test plan
- [x] New `GetChainTokensUseCaseTest`:
  - same ticker / two contract addresses on one chain → emitted once, no duplicate `id`, built-in entry (`0xA`) survives
  - distinct tickers on the same chain are all kept
- [x] ktfmt clean

Pure data-layer fix (no UI render change), so no before/after screenshots apply.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate tokens appearing in the token list that could cause display rendering issues. Tokens are now properly de-duplicated to ensure a cleaner, conflict-free interface.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/chaintokens/GetChainTokensUseCase.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/chaintokens/GetChainTokensUseCase.kt
@@ -86,6 +86,12 @@ constructor(
                 .flatten()
                 .asSequence()
                 .distinctBy { it.contractAddress to it.chain.id }
+                // The UI keys each row on Coin.id (ticker-chainId). Collapse any remaining
+                // collisions on that key so the same ticker sourced with two different contract
+                // addresses on one chain can't produce duplicate LazyColumn keys and crash on
+                // scroll (issue #4869). Built-in/refreshed tokens are ordered first, so the
+                // canonical entry wins over a searchable/provider duplicate.
+                .distinctBy { it.id }
                 .toList()
                 .modifyIfNeeded()
         emit(coins)
--- a/data/src/test/kotlin/com/vultisig/wallet/data/usecases/chaintokens/GetChainTokensUseCaseTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/usecases/chaintokens/GetChainTokensUseCaseTest.kt
@@ -1,0 +1,79 @@
+@file:OptIn(ExperimentalCoroutinesApi::class)
+
+package com.vultisig.wallet.data.usecases.chaintokens
+
+import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.models.Vault
+import com.vultisig.wallet.data.repositories.TokenRepository
+import io.mockk.coEvery
+import io.mockk.every
+import io.mockk.mockk
+import kotlin.test.assertEquals
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.toList
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Test
+
+internal class GetChainTokensUseCaseTest {
+
+    private val tokenRepository: TokenRepository = mockk()
+    private val vault: Vault = mockk()
+
+    private val useCase =
+        GetChainTokensUseCaseImpl(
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4872, opened 2026-06-11
fix(cosmos): validate IBC memo and timeout instead of signing a bad packet

### LINKED ISSUE #4849: IBC memo positional split with silent "" / 0L fallbacks builds bad/expirable packet
## Observed

IBC transfers parse the source channel, forwarding memo, and timeout by **positional split** of the memo string on `:`. A malformed or unexpectedly-shaped memo silently produces an empty source channel, an empty forward memo, and/or a timeout of `0` — yielding an IBC packet that either targets the wrong channel or never times out cleanly (funds can hang until manual timeout/refund).

## Cause

`CosmosHelper.getPreSignedInputData()` IBC branch (`data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CosmosHelper.kt:148-182`):

```kotlin
val memoParts = keysignPayload.memo?.split(":")
    ?: throw Exception("To send IBC transaction, memo should be specified")
val sourceChannel = memoParts.getOrNull(1) ?: ""          // silently "" if missing
val memo = if (memoParts.size == 4) memoParts[3] else ""    // silently dropped unless exactly 4 parts
val timeouts = atomData.ibcDenomTraces?.latestBlock?.split("_") ?: emptyList()
val timeout = timeouts.lastOrNull()?.toLongOrNull() ?: 0L   // silently 0 on parse failure
```

`sourceChannel = ""` builds a packet on no channel; `timeout = 0L` (a zero timeout timestamp) defeats the timeout safeguard. None of these degenerate cases throw — they sign a malformed/expirable packet.

## Fix shape

- Validate the memo shape before signing: require a non-blank `sourceChannel`, and `require` the part count matches the expected IBC memo grammar instead of `getOrNull`/`?: ""`.
- Treat a `timeout == 0L` (or unparseable `latestBlock`) as a hard error rather than signing a packet with no effective timeout.

---

aminsato:
Closes #4849

## Problem
The IBC branch of `CosmosHelper.getPreSignedInputData()` parsed the keysign memo by **positional split with silent fallbacks**:

```kotlin
val sourceChannel = memoParts.getOrNull(1) ?: ""               // "" → packet on no channel
val memo = if (memoParts.size == 4) memoParts[3] else ""       // forward memo silently dropped
val timeout = timeouts.lastOrNull()?.toLongOrNull() ?: 0L      // 0L → no effective timeout
```

The memo grammar (`DepositMemo.TransferIbc`) is `dstChain:sourceChannel:dstAddress[:forwardMemo]`, and `sourceChannel` is `""` for unsupported chain pairs. So a malformed/unexpected memo silently signed a packet on **no channel**, and an unparseable `latestBlock` signed a **zero-timeout** packet — an IBC packet with both timeout height and timestamp zero is invalid and can hang until manual refund. None of these threw.

## Fix
Extract a pure, testable `parseIbcTransferParams(memo, latestBlock)` that fails loudly:
- `require` a non-blank `sourceChannel` (mirrors the existing guard in `QBTCTransactionHelper`),
- `require` a non-zero, parseable `timeoutTimestamp` instead of signing a no-timeout packet,
- rejoin the forward memo (`drop(3).joinToString(":")`) so a forward memo containing `:` survives instead of being silently dropped.

The valid path (well-formed 3/4-part memo + parseable `latestBlock`) is unchanged — only previously-malformed cases now error instead of broadcasting a bad packet. Extracting the parse to a pure function keeps it unit-testable without the Trust Wallet Core JNI used in the signing path.

## Test plan
- [x] New `ParseIbcTransferParamsTest` (7 cases): 3-part memo, 4-part memo, forward memo containing `:`, null memo, blank/missing channel, null `latestBlock`, zero/unparseable timeout
- [x] ktfmt clean

Pure validation change (no UI), so no before/after screenshots apply.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation and error handling for IBC (Inter-Blockchain Communication) transfers to ensure memo parameters are correctly parsed with stricter enforcement of required fields and timeout values.

* **Tests**
  * Added comprehensive test coverage for IBC transfer parameter parsing, including validation of edge cases and error scenarios.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CosmosHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CosmosHelper.kt
@@ -147,15 +147,14 @@ class CosmosHelper(
         }
         return when (atomData.transactionType) {
             TransactionType.TRANSACTION_TYPE_IBC_TRANSFER -> {
-                val memoParts =
-                    keysignPayload.memo?.split(":")
-                        ?: throw Exception("To send IBC transaction, memo should be specified")
-                val sourceChannel = memoParts.getOrNull(1) ?: ""
-
-                val memo = if (memoParts.size == 4) memoParts[3] else ""
-
-                val timeouts = atomData.ibcDenomTraces?.latestBlock?.split("_") ?: emptyList()
-                val timeout = timeouts.lastOrNull()?.toLongOrNull() ?: 0L
+                val ibc =
+                    parseIbcTransferParams(
+                        memo = keysignPayload.memo,
+                        latestBlock = atomData.ibcDenomTraces?.latestBlock,
+                    )
+                val sourceChannel = ibc.sourceChannel
+                val memo = ibc.forwardMemo
+                val timeout = ibc.timeoutTimestamp
 
                 val transferMessage =
                     Cosmos.Message.Transfer.newBuilder()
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4875, opened 2026-06-11
fix(solana): confirmed blockhash commitment + client-side rebroadcast loop

### LINKED ISSUE #4863: Solana blockhash uses finalized commitment + one-shot broadcast -> avoidable expiry failures
## Observed

Solana sends/swaps can fail with "Blockhash not found" / "block height exceeded" shortly after the keysign ceremony, and recovery from that state is thin.

## Cause

`data/.../api/SolanaApi.kt` — `getRecentBlockHash()` fetches the blockhash with the slowest commitment:
```kotlin
method = "getLatestBlockhash",
params = buildJsonArray { addJsonObject { put("commitment", "finalized") } },
```
`finalized` is ~32 slots (~13s) behind the tip, burning that much of the ~60–90s validity window up front. `confirmed` is standard for sending.

`broadcastTransaction(tx)` in the same file is one-shot: a single `sendTransaction` with no `maxRetries`, no client-side rebroadcast loop, and no expiry-bounded resend. On any RPC error it simply throws.

No `lastValidBlockHeight` is captured — `RecentBlockHashValueJson` decodes only `blockhash` (`SolanaJson.kt`), so true expiry can only be inferred by string-matching the broadcast error.

## Fix shape

- Switch the blockhash commitment to `confirmed` in `getRecentBlockHash()`.
- Add a client-side rebroadcast loop: resend the same signed tx on a transient "blockhash not found" (RPC node hasn't yet observed our `confirmed` blockhash — propagation lag) with bounded retries + backoff, while treating "block height exceeded" as true expiry that should re-sign with a fresh blockhash rather than resend.
- (Larger follow-up) Capture `lastValidBlockHeight` and bound the rebroadcast loop on current block height; consider durable nonces for the long-pairing case.

## Cross-platform

Equivalent of vultisig-ios#4532 (fixed in vultisig-ios#4551). iOS applied the first two items; `lastValidBlockHeight` threading was deferred there because it requires a `KeysignMessage` proto change shared across devices + Fast Vault server.

---
Heads up — AI-assisted analysis; verified by hand against current Android `main`.

aminsato:
Closes #4863

## Problem
Two avoidable Solana broadcast failure modes:

1. `getRecentBlockHash()` fetched the blockhash with the **`finalized`** commitment. `finalized` lags the tip by ~32 slots (~13s), burning that much of the ~60–90s blockhash validity window before the keysign ceremony even starts — so a slightly slow ceremony can broadcast against an already-near-expired blockhash.
2. `broadcastTransaction()` was **one-shot**: any RPC error threw immediately, including a transient `"blockhash not found"` that just means the RPC node we hit hasn't observed our (confirmed) blockhash yet (propagation lag).

## Fix (ports vultisig-ios#4551, items 1–2)
- Switch the blockhash commitment to **`confirmed`** — the standard for sending; it tracks the tip closely.
- Add a bounded **client-side rebroadcast loop** (3 attempts, 2s backoff):
  - `"blockhash not found"` with retries left → resend the same signed tx after a backoff.
  - `"block height exceeded"`, or an exhausted blockhash-not-found retry → throw `SolanaBlockhashExpiredException` (true expiry; resending can't help).
  - Any other RPC error → rethrown unchanged.
- The error classification is a pure, unit-tested `solanaBroadcastAction(message, attempt, maxAttempts)` returning `RESEND` / `EXPIRED` / `FATAL`. `-32002` is Solana's generic preflight code, so we match on the message text (same as iOS).

## Scope (matches iOS)
Threading `lastValidBlockHeight` for a precise expiry-bounded loop and having the ceremony **re-sign** with a fresh blockhash on expiry is deferred — it needs a cross-device `KeysignMessage` proto change (and Android has no re-sign hook today). The new `SolanaBlockhashExpiredException` is the typed signal that follow-up will consume.

## Test plan
- [x] `SolanaBroadcastActionTest` — classifier: resend-with-retries, exhausted→expired, height-exceeded→expired, generic→fatal, case-insensitivity
- [x] `SolanaApiBodyReadTest` — added end-to-end error paths: expired→`SolanaBlockhashExpiredException`, generic→`IllegalStateException`; existing success path still green
- [x] full `:data:testDebugUnitTest` — green
- [x] ktfmt clean
- [ ] Manual: Solana send + swap broadcast end-to-end (cannot mainnet-test from here)

## Summary by CodeRabbit

* **New Features**
  * Implemented automatic retry mechanism for transaction broadcasting with exponential backoff between attempts
  * Enhanced error handling to classify network failures into recoverable and terminal error types
  * Added specific exception for blockhash expiration scenarios

* **Improvements**
  * Updated block confirmation strategy for faster transaction validation

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
@@ -33,16 +33,21 @@ import io.ktor.client.statement.bodyAsText
 import io.ktor.http.HttpStatusCode
 import java.math.BigInteger
 import javax.inject.Inject
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.delay
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonPrimitive
 import kotlinx.serialization.json.add
 import kotlinx.serialization.json.addJsonArray
 import kotlinx.serialization.json.addJsonObject
 import kotlinx.serialization.json.buildJsonArray
+import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.jsonPrimitive
 import kotlinx.serialization.json.put
 import timber.log.Timber
 
@@ -135,7 +140,11 @@ constructor(
             RpcPayload(
                 jsonrpc = "2.0",
                 method = "getLatestBlockhash",
-                params = buildJsonArray { addJsonObject { put("commitment", "finalized") } },
+                // `confirmed` is the standard commitment for sending: it tracks the tip closely,
+                // whereas `finalized` lags ~32 slots (~13s) behind and burns that much of the
+                // ~60-90s blockhash validity window before the keysign ceremony even starts
+                // (issue #4863).
+                params = buildJsonArray { addJsonObject { put("commitment", "confirmed") } },
                 id = 1,
             )
         val response = httpClient.post(rpcEndpoint) { setBody(payload) }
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4877, opened 2026-06-11
fix(swap): decode bytes32 token name/symbol instead of showing raw hex

### LINKED ISSUE #4873: fix(swap): bytes32 token name/symbol shown as raw hex (e.g. MKR renders as 4d4b52…)
## Summary

In the Swap screen, certain destination tokens render their **name/symbol as a raw `bytes32` hex string** instead of decoded text. Example: MKR (Maker) on Arbitrum shows as `4d4b520000…0000` instead of `MKR`, and its name shows as `4d616b6572…0000` instead of `Maker`.

## Steps to reproduce

1. Open Swap.
2. From: ETH (Arbitrum, native).
3. Select destination token MKR on Arbitrum (`0x2e9a6Df78E42a30712c10a9Dc4b1C8656f8F2879`).
4. Observe the "To" token name and the token row.

## Expected

- Token symbol shows `MKR`.
- Token name shows `Maker`.

## Actual

- Symbol/name render as the undecoded hex string `4d4b5200000000…000000` / `4d616b6572000…000000`.

See screenshot — the "To" field and the destination token row both show the hex string.

## Cause

MKR is a non-standard ERC-20: its `name()` and `symbol()` are declared as `bytes32`, not `string`. The aggregator quote returns these as the raw hex-encoded `bytes32` value, right-padded with zeros:

```
name:    4d616d6572000…000  →  ASCII "Maker"
coinKey: 4d4b520000000…000  →  ASCII "MKR"
```

These bytes32 values reach the UI without being decoded back to text. A handful of other legacy tokens use `bytes32` names/symbols as well, so the fix should generically detect and decode `bytes32`-style hex values into UTF-8 (trimming the zero padding), with a fallback to the `symbol` field when available.

## Notes

- The token itself is legitimate; `verificationStatus: "unverified"` is the aggregator's own listing flag and unrelated to this display bug.
- Affects display only — selection/quote still function.

## Screenshot

swap-mkr-hex-name

aminsato:
Closes #4873

## Problem
Some legacy ERC-20s (e.g. **MKR**) declare `name()` / `symbol()` as **`bytes32`**, not `string`. Bridged deployments re-encode that as an ABI `string` whose content is the 64-char **hex** of the original `bytes32`, right-padded with zeros. `TokenRepository.decodeContractString()` decoded the ABI-string layer correctly but then surfaced that hex text verbatim — so MKR on Arbitrum resolved to the ticker **`4D4B52…00`** instead of **`MKR`**.

## Root cause (verified on-chain)
The real Arbitrum `symbol()` eth_call for MKR (`0x2e9a…2879`) returns:

```
0x …0020 (offset) …0040 (length=64) 34643462353230303030…   ← ASCII "4d4b5200…00"
```

i.e. a valid ABI `string` whose **content is the hex** `4d4b52…00`. `name()` likewise returns the hex of `"Maker"`.

## Fix
Extract the decoder into a testable top-level `decodeErc20MetadataString()`. After the normal ABI-`string` decode, detect a 64-char hex value that decodes to printable ASCII (a `bytes32` re-encoded as hex) and decode it back to UTF-8, trimming the zero padding. Everything else — normal string symbols, and short tickers whose letters happen to be hex (e.g. `FACE`) — passes through unchanged (guarded by the 64-char-length + printable-ASCII checks).

## Verification

Device (Add custom token → Arbitrum → paste MKR `0x2e9a…2879`) — token now resolves to **MKR**:
| ‌Before |
|-------|
| telegram-cloud-document-4-5859592580211155483 |

| After |
|-------|
| mkr-after |

## Test plan
- [x] New `DecodeErc20MetadataStringTest` — fixtures are the **real captured Arbitrum eth_call responses**: MKR `symbol()` → `MKR`, `name()` → `Maker`; plus a normal `USDC` string, a `FACE` false-positive guard, and malformed → null
- [x] Device-verified on emulator (screenshot above)
- [x] ktfmt clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved decoding of ERC‑20 token names and symbols so legacy bytes32-as-hex values are converted back to readable tickers (fixes display/lookup for bridged/legacy tokens) while avoiding mangling short hex-like tickers and returning null for malformed metadata.

* **Tests**
  * Added comprehensive tests covering standard ABI strings, bytes32-as-hex decoding, non-printable/edge cases, and malformed inputs.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
@@ -274,12 +274,23 @@ internal fun decodeErc20MetadataString(raw: String): String? {
         val bytes = raw.removePrefix("0x").chunked(2).map { it.toInt(16).toByte() }.toByteArray()
         val length = BigInteger(bytes.sliceArray(32..63)).toInt()
         String(bytes.sliceArray(64 until 64 + length)).decodeBytes32HexOrSelf()
-    } catch (_: Exception) {
+    } catch (_: NumberFormatException) {
+        // Non-hex characters in the eth_call result, or an empty byte range fed to BigInteger.
+        null
+    } catch (_: IndexOutOfBoundsException) {
+        // Result shorter than the ABI offset/length/data words require.
         null
     }
 }
 
-private fun String.decodeBytes32HexOrSelf(): String {
+/**
+ * Decodes a value that is the 64-char hex of a `bytes32` (right-padded with zeros) back to UTF-8
+ * text, trimming the zero padding. Returns the receiver unchanged when it is not such a value — so
+ * it is safe to apply to any name/symbol string (a normal ticker like `MKR` passes straight
+ * through). Used both for ABI `eth_call` results ([decodeErc20MetadataString]) and for aggregator
+ * token metadata that already arrives as the bare `bytes32` hex (issue #4873).
+ */
+internal fun String.decodeBytes32HexOrSelf(): String {
     if (length != 64 || any { it !in '0'..'9' && it !in 'a'..'f' && it !in 'A'..'F' }) return this
     val raw = chunked(2).map { it.toInt(16).toByte() }.toByteArray()
     val text = raw.takeWhile { it.toInt() != 0 }.toByteArray()
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
@@ -66,7 +66,7 @@ constructor(
         rpcResponses.forEach {
             val result = it.result ?: return null
             if (it.id == CUSTOM_TOKEN_RESPONSE_TICKER_ID)
-                ticker = result.decodeContractString() ?: return null
+                ticker = decodeErc20MetadataString(result) ?: return null
             else decimal = result.decodeContractDecimal().takeIf { dec -> dec != 0 } ?: return null
         }
         val coin =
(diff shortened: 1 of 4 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenRepository.kt
```diff
@@ -66,7 +66,7 @@ constructor(
         rpcResponses.forEach {
             val result = it.result ?: return null
             if (it.id == CUSTOM_TOKEN_RESPONSE_TICKER_ID)
-                ticker = result.decodeContractString() ?: return null
+                ticker = decodeErc20MetadataString(result) ?: return null
```
rkokhatskyi, line 69: This decoder runs only through `getEVMTokenByContract`, which the Swap 'To' list never calls. That list is built in `SelectAssetViewModel` from `GetChainTokensUseCase` and `oneInchApi.getTokens`, so issue #4873's swap-selector symbol does not route through this fix

aminsato: Good catch — you're right, thanks. The selector list is built from `OneInchToCoinsUseCase` (`it.symbol`) and `EvmCoinFinder.fromOneInch` (`token.symbol`), which read the aggregator's raw symbol and never touch `getEVMTokenByContract`, so the bytes32 hex was slipping straight through to the 'To' list.

Fixed in 43aa24d97: promoted the inner `decodeBytes32HexOrSelf()` helper to `internal` and applied it where those two map the aggregator symbol to `Coin.ticker` (it's a safe no-op on normal tickers). `getEVMTokenByContract` keeps the full-ABI `decodeErc20MetadataString` for the custom-token-add / swap-fee path. Added unit tests for the bare-bytes32 case.

Note: the issue's 'Cause' section uses LI.FI terminology (`coinKey`), so if a destination token is ever sourced from a LI.FI quote rather than the 1inch catalog, that mapping would need the same treatment — the current selector is 1inch-backed, which this now covers.

rkokhatskyi (approved):
LGTM

PR #4879, opened 2026-06-11
fix(zcash): fetch live ZIP-243 branch id at send time (#4851)

### LINKED ISSUE #4851: Zcash ZIP-243 consensus branch ID hardcoded, no runtime fetch
## Observed

The Zcash transparent-spend sighash uses a **hardcoded** ZIP-243 consensus branch ID. At each Zcash network upgrade the branch ID changes; until someone manually edits the constant and ships a release, every Zcash send signs with the wrong branch ID and is rejected by the network — Zcash sends break chain-wide at the upgrade boundary with no runtime recovery.

## Cause

`ZcashConstants.kt:11` (`data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/ZcashConstants.kt`):

```kotlin
internal const val ZCASH_ZIP243_BRANCH_ID_HEX = "30f33754"
```

The doc comment itself states it must be "Regenerate[d] per Zcash network upgrade" from `getblockchaininfo -> consensus.nextblock`. It is consumed by both the native send plan (`UtxoHelper`) and `SwapKitZcashSigner`, so a missed update breaks both native sends and Zcash swaps.

## Fix shape

- Fetch the active branch ID at send time from a Zcash RPC (`getblockchaininfo` → `consensus.chaintip` / the upgrade table) and feed it into the signing input, falling back to the constant only if the fetch fails.
- At minimum, add an automated check (CI or runtime warning) that the constant matches the current network branch ID so the breakage is caught before the upgrade activates.

---

johnnyluo:
## Summary

Fixes #4851. The transparent ZEC sighash used a **hardcoded** ZIP-243 consensus branch id (`ZCASH_ZIP243_BRANCH_ID_HEX = "30f33754"`). At each Zcash network upgrade the branch id changes, so every native ZEC send and SwapKit ZEC swap would sign with the wrong id and be rejected chain-wide until someone manually edited the constant and shipped a release.

This resolves the active branch id from the Zcash RPC at send time and carries it on the payload, with **no compiled-in fallback** — if the value can't be resolved, signing refuses rather than producing a network-rejectable transaction.

## How it works

- **`ZcashApi`** (`api.vultisig.com/zcash`) calls `getblockchaininfo`, reads `consensus.nextblock` (big-endian, e.g. `5437f330`) and reverses the four bytes to the little-endian form WalletCore expects (`30f33754`). Result is cached in memory for **1 hour** (the id only changes at an upgrade), so a single keysign's preimage-hash and final-compile passes read the same value.
- The branch id rides on **`BlockChainSpecific.UTXO.zcashBranchId`** instead of being threaded through signing-helper params:
  - **Initiator**: stamped in `BlockChainSpecificRepository.getSpecific` (covers native sends and SwapKit swaps).
  - **Co-signer**: the `UTXOSpecific` proto can't carry the field, so a payload rebuilt from the initiator's QR/relay proto arrives with it null. `JoinKeysignViewModel` re-resolves the same network-global value before computing sighashes, so the whole committee agrees.
  - `UtxoHelper` / `SwapKitZcashSigner` read it straight off the payload.
- **Fail-loud**: `UtxoHelper` and `SwapKitZcashSigner` throw when the branch id is null/empty rather than falling back to a stale constant. `ZcashConstants.kt` is deleted.

## Trade-off

A reachable Zcash RPC is now a hard requirement for ZEC signing. If `api.vultisig.com/zcash` is briefly unavailable, ZEC sends/swaps fail with a descriptive error until it recovers — deliberate: fail loud over sign-stale.

## Test plan

- [x] New `ZcashApiTest` (7): byte-reversal, missing/malformed field, server error, correct URL, cache-hit skips second request, failure isn't cached.
- [x] New `BlockChainSpecificRepositoryImplTest` case: `getSpecific` stamps the live branch id for Zcash.
- [x] New `SwapKitZcashSignerTest` case: missing branch id is rejected (no fallback); existing happy-path/fee tests updated to pass a branch id.
- [x] `UtxoHelperTest`, `SigningHelperCustomMessageTest` still green.
- [x] `:data` + `:app` main, unit-test, and androidTest sources compile.

## Summary by CodeRabbit

* **New Features**
  * Fetches live Zcash ZIP-243 consensus branch ID at send time and threads it into Zcash signing flows and payloads.

* **Bug Fixes**
  * Removed hardcoded branch ID fallback; signing now requires a live branch ID and fails early if unavailable to ensure correct ZEC signatures.

* **Tests**
  * Added tests covering branch-ID fetch/parsing, caching behavior, signer rejection of missing ID, and repository/payload integration.


data/src/main/kotlin/com/vultisig/wallet/data/models/payload/BlockChainSpecfic.kt
```diff
@@ -6,7 +6,18 @@ import vultisig.keysign.v1.SuiCoin
 import vultisig.keysign.v1.TransactionType
 
 sealed class BlockChainSpecific {
-    data class UTXO(val byteFee: BigInteger, val sendMaxAmount: Boolean) : BlockChainSpecific()
+    /**
+     * @param zcashBranchId live ZIP-243 consensus branch id (little-endian hex, e.g. `30f33754`)
+     *   fetched at send time for Zcash; null for every other UTXO chain and when the RPC was
+     *   unreachable, in which case signing falls back to the compiled-in constant. Transient: it is
```
aminsato: Stale comment: there is no compiled-in fallback anymore. `ZcashConstants.kt` (`ZCASH_ZIP243_BRANCH_ID_HEX`) is deleted in this PR, and `UtxoHelper` now `require`s a non-null branch id (throws otherwise). When the RPC is unreachable this stays null and signing **refuses** — it does not fall back. The "falls back to the compiled-in constant" wording contradicts the actual fail-loud behavior and will mislead future maintainers. Suggest: "...null for every other UTXO chain and when the RPC was unreachable, in which case signing refuses (there is no compiled-in fallback)."

data/src/main/kotlin/com/vultisig/wallet/data/models/payload/BlockChainSpecfic.kt
```diff
@@ -97,3 +108,10 @@ sealed class BlockChainSpecific {
     data class Cardano(val byteFee: Long, val sendMaxAmount: Boolean, val ttl: ULong) :
         BlockChainSpecific()
 }
+
+/**
+ * Live ZIP-243 branch id carried on a Zcash payload's UTXO specific, or null for non-Zcash payloads
+ * and when the RPC was unreachable (signing then falls back to the compiled-in constant).
```
aminsato: Same stale "falls back to the compiled-in constant" wording here — the constant is removed and signing now fails loud on a null branch id. Reword to reflect the refuse-to-sign behavior.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SwapKitZcashSigner.kt
```diff
@@ -40,13 +40,18 @@ internal class SwapKitZcashSigner(
     private val coinType = CoinType.ZCASH
     private val utxo = UtxoHelper(coinType, vaultHexPublicKey, vaultHexChainCode)
 
-    /** ZIP-243 preimage hashes (sorted hex) for every input. */
+    /**
+     * ZIP-243 preimage hashes (sorted hex) for every input. [zcashBranchId] is the live branch id
+     * resolved at send time (carried on the source-chain payload's UTXO specific); null falls back
+     * to the compiled-in constant.
```
aminsato: KDoc says null "falls back to the compiled-in constant", but `buildSigningInputData` now throws `SwapKitZcashSignerException` when `zcashBranchId` is null/empty (the dedicated test asserts this). No fallback exists — update the KDoc to say a null branch id is rejected.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -1778,6 +1789,18 @@ constructor(
         }
     }
 
+    /**
+     * Stamps the live ZIP-243 branch id onto a Zcash UTXO payload rebuilt from proto. Returns the
+     * payload unchanged for non-Zcash chains, non-UTXO specifics, or when the RPC is unreachable
+     * (UtxoHelper then falls back to the compiled-in constant).
```
aminsato: "(UtxoHelper then falls back to the compiled-in constant)" is no longer true — UtxoHelper `require`s the branch id and throws. Returning the payload unchanged here leads to a refuse-to-sign, not a fallback. Reword to match the fail-loud behavior the rest of the PR establishes.

rkokhatskyi (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/api/ZcashApi.kt
```diff
@@ -0,0 +1,102 @@
...
+            if (branchIdBigEndian.isNullOrEmpty()) {
+                Timber.w("Zcash getblockchaininfo returned no consensus.nextblock branch id")
+                return null
+            }
+            reverseHexBytes(branchIdBigEndian)
+        } catch (e: Exception) {
+            if (e is kotlinx.coroutines.CancellationException) throw e
+            Timber.e(e, "Failed to fetch Zcash consensus branch id")
+            null
+        }
```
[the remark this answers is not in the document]
johnnyluo: Keeping the broad `catch (e: Exception) -> null` here intentionally. `fetchBranchId()` is a fail-safe resolver: any failure to obtain a valid live branch id (NetworkException from `bodyOrThrow()`, Ktor deserialization/convert errors, malformed payload) must map to `null` so the signing layer (`UtxoHelper` `require`, `SwapKitZcashSigner`) refuses to sign rather than use a stale/guessed branch id. `getConsensusBranchIdHex()` runs inside the signing flow, so letting a non-network exception propagate would turn a graceful refuse-to-sign into a crash. `CancellationException` is already re-thrown, and the real exception is logged via `Timber.e(e, ...)`, so genuine bugs are visible in logs without degrading the fail-safe outcome.

PR #4892, opened 2026-06-13
fix(swap): price tokens via contract address when priceProviderID misses

### LINKED ISSUE #4874: Swap: destination token fiat shows $0.00 (e.g. ezETH on Base) despite non-zero amount
## Summary

In the Swap flow, the destination ("To") token shows a **fiat value of $0.00** even though the token amount is non-zero. Reproduced swapping **ETH → ezETH on Base**: the estimated receive amount renders as `0.00097134 ezETH` but its fiat is `$0.00` on both the **Swap** screen and the **Swap overview / VerifySwap** screen.

## Steps to reproduce

1. Open Swap.
2. From: **ETH (Base)**. To: **ezETH (Base)**.
3. Enter a sell amount (e.g. `0.00105559` ETH).
4. Observe the quoted "To" amount and its fiat value on the Swap screen, then proceed to the Swap overview.

## Expected

The destination amount should display a non-zero fiat value (ezETH has a market price), consistent with how the source token's fiat is shown.

## Actual

- Swap screen "To" section: `0.00097134 ezETH`, fiat `$0.00`.
- Swap overview "min. payout": `0.00097134 ezETH`, fiat `$0.00`.

The source side (ETH) correctly shows `$1.75`.

## Notes on the "0 ezETH" label

The `0 ezETH` text in the top-right of the Swap screen's "To" section is the wallet's current **balance** of ezETH (the vault holds none), which is separate from the estimated receive amount shown below it. The reportable defect here is the **$0.00 fiat value**, which appears on both screens.

## Likely cause (verified against the code)

Token price refresh branches on token metadata. The branch order tries `priceProviderID` before `contractAddress`, and the two are mutually exclusive (`when { ... -> ... }`):

- `ezETH` is defined with `priceProviderID = "ezETH"` and a valid Base `contractAddress`.
- Because `priceProviderID` is non-empty, the token is routed to the CoinGecko `/simple/price?ids=...` lookup. `ezETH` is not a valid CoinGecko coin id, so that lookup returns nothing.
- The contract-address lookup (`/token_price/base?contract_addresses=...`), which *does* return a price for this contract, is never reached for this token.

The result: no price is cached for ezETH, so `convertTokenValueToFiat` yields `$0.00`. This primarily affects swap destination tokens that aren't already in the vault (and therefore weren't price-refreshed up front).

Other Base tokens defined with a `priceProviderID` that CoinGecko doesn't recognize (but with a valid contract address) would hit the same path.

## Possible fixes

- Fall back to the contract-address price lookup when the `priceProviderID` lookup returns no price, or
- Correct `ezETH`'s `priceProviderID` to a CoinGecko-recognized id, or clear it so the contract-address path is used.

## Environment

- Platform: Android
- Chain: Base
- Provider observed in quote: LI.FI

aminsato:
## Summary

Fixes #4874 — in the Swap flow, the destination token showed a fiat value of **$0.00** despite a non-zero amount (reproduced swapping ETH → ezETH on Base).

### Tranaction hash
https://basescan.org/tx/0x1a484b0fc706cf6741f359661494df60db18746e9275f4ef50b9ecbd9640b5a8

**Root cause:** `TokenPriceRepository.refresh()` routed any token with a non-empty `priceProviderID` to the CoinGecko `/simple/price?ids=...` lookup and never to the contract-address lookup — the two branches were mutually exclusive. `ezETH` is defined with `priceProviderID = "ezETH"` (not a CoinGecko-recognized id) **and** a valid Base `contractAddress`, so the provider-id lookup returned nothing and the contract-address path — which does return a price (CoinGecko, with a LI.FI fallback) — was never reached. No price was cached, so `convertTokenValueToFiat` yielded `$0.00`.

**Fix:** After the provider-id lookup completes, any token that has a `contractAddress` but didn't get a price from its `priceProviderID` now falls back to the contract-address lookup. This is generic — it covers ezETH and any other token with an unrecognized `priceProviderID` but a valid contract.

## Changes

- `TokenPriceRepository.refresh()`: add contract-address fallback for tokens whose `priceProviderID` returned no price.
- Add `TokenPriceRepositoryImplTest` covering both the fallback and the no-regression case.

## Test plan

- `./gradlew :data:testDebugUnitTest --tests "*.TokenPriceRepositoryImplTest"` — passes:
  - Falls back to contract-address lookup when `priceProviderID` returns no price → price cached.
  - Does **not** call the contract-address lookup when `priceProviderID` succeeds (no extra network call for working tokens).
- Verified in the running app: destination ezETH now shows a non-zero fiat value.

## After

| Swap screen (ETH → ezETH on Base) |
|---|
| image |

ezETH destination now shows `$1.28` (was `$0.00`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token price refresh by adding an EVM-specific fallback to ensure tokens are priced via contract-address lookup when provider-id pricing is missing.

* **Tests**
  * Added unit tests covering both fallback and non-fallback behaviors for token price retrieval.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
@@ -126,6 +126,21 @@ constructor(
 
         savePrices(pricesWithProviderIds, currency)
 
+        // Fall back to the contract-address lookup for tokens whose priceProviderID returned no
+        // price but which have a valid contract address (e.g. ezETH on Base, whose priceProviderID
+        // is not a CoinGecko-recognized id). Without this, such tokens would never be priced.
+        val pricedTokenIds = pricesWithProviderIds.keys
+        tokens.forEach { token ->
+            if (
+                token.priceProviderID.isNotEmpty() &&
+                    token.contractAddress.isNotEmpty() &&
+                    token.id !in pricedTokenIds
+            ) {
+                val existingChain = chainContractAddresses.getOrPut(token.chain) { mutableListOf() }
+                chainContractAddresses[token.chain] = existingChain + token
+            }
+        }
+
         chainContractAddresses.map { (chain, tokens) ->
             val pricesWithContractAddress =
                 fetchPricesWithContractAddress(
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
@@ -11,6 +11,7 @@ import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.Coins
 import com.vultisig.wallet.data.models.TokenId
+import com.vultisig.wallet.data.models.TokenStandard
 import com.vultisig.wallet.data.models.settings.AppCurrency
 import java.math.BigDecimal
 import java.math.RoundingMode
@@ -78,7 +79,6 @@ constructor(
         val currencies = listOf(currency)
 
         val tokensByPriceProviderIds = tokens.groupBy { it.priceProviderID.lowercase() }
-        val tokensByContractAddress = tokens.associateBy { it.contractAddress.lowercase() }
 
         val priceProviderIds = mutableListOf<String>()
         val chainContractAddresses = mutableMapOf<Chain, List<Coin>>()
@@ -126,13 +126,16 @@ constructor(
 
         savePrices(pricesWithProviderIds, currency)
 
-        // Fall back to the contract-address lookup for tokens whose priceProviderID returned no
+        // Fall back to the contract-address lookup for EVM tokens whose priceProviderID returned no
         // price but which have a valid contract address (e.g. ezETH on Base, whose priceProviderID
         // is not a CoinGecko-recognized id). Without this, such tokens would never be priced.
+        // Restricted to EVM so non-EVM contract formats (e.g. THORChain x/… tokens handled by
+        // fetchThorContractPrices) aren't fanned out to CoinGecko/LI.FI per-contract calls.
         val pricedTokenIds = pricesWithProviderIds.keys
         tokens.forEach { token ->
             if (
-                token.priceProviderID.isNotEmpty() &&
+                token.chain.standard == TokenStandard.EVM &&
+                    token.priceProviderID.isNotEmpty() &&
                     token.contractAddress.isNotEmpty() &&
                     token.id !in pricedTokenIds
             ) {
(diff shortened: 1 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -126,6 +126,21 @@ constructor(
 
         savePrices(pricesWithProviderIds, currency)
 
+        // Fall back to the contract-address lookup for tokens whose priceProviderID returned no
+        // price but which have a valid contract address (e.g. ezETH on Base, whose priceProviderID
+        // is not a CoinGecko-recognized id). Without this, such tokens would never be priced.
+        val pricedTokenIds = pricesWithProviderIds.keys
+        tokens.forEach { token ->
```
rkokhatskyi, line 135: `ezETH` on Arbitrum, Base and Optimism share one `contractAddress`, but `tokensByContractAddress` (line 81) is keyed by address only, so the new fallback collapses all three chains' contract prices onto a single `id` at line 153. The other same-contract instances stay at $0.00

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -126,6 +126,21 @@ constructor(
...
+            if (
+                token.priceProviderID.isNotEmpty() &&
+                    token.contractAddress.isNotEmpty() &&
+                    token.id !in pricedTokenIds
+            ) {
+                val existingChain = chainContractAddresses.getOrPut(token.chain) { mutableListOf() }
+                chainContractAddresses[token.chain] = existingChain + token
+            }
+        }
```
johnnyluo, line 145: This fallback queues **any** token with a non-empty `priceProviderID` + `contractAddress` that missed — including THORChain tokens with `x/nami…` / `x/staking-tcy` contract addresses (handled by `fetchThorContractPrices`) and Thor pool tokens that happen to carry a `priceProviderID`. Those will now also be sent to `getContractsPrice(chain = ThorChain, …)` plus the LI.FI fallback before their dedicated handler runs later in `refresh()`.

Correctness is preserved (the Thor handlers run afterward and overwrite, and `savePrices` filters zero values so no bad price is cached), but it's wasted network work. Also note that a transient empty response from `getCryptoPrices` (rate-limit / blip) would fan every contract-bearing token out to per-contract CoinGecko + LI.FI calls.

Consider scoping the fallback to EVM chains, or excluding non-EVM contract-address formats (e.g. `x/…`). Not a blocker.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Good catch. Fixed in 63f1c0729 — the price lookup now resolves ids from each chain's own token list inside the loop instead of a global address-keyed map, so `ezETH` on Arbitrum/Base/Optimism each keep their own `id` and get priced individually.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Agreed, scoped the fallback to EVM chains in 63f1c0729 (`token.chain.standard == TokenStandard.EVM`). THORChain `x/…` tokens and Thor pool tokens carrying a `priceProviderID` now stay with their dedicated handlers instead of being fanned out to CoinGecko/LI.FI per-contract calls, which also avoids the rate-limit-blip fan-out for non-EVM tokens.

johnnyluo (approved):
LGTM

PR #4893, opened 2026-06-13
fix(swap): clamp inflated destination fiat to source fiat (#4878)

### LINKED ISSUE #4878: Swap: destination fiat uses market price, not the quote rate — inflated for illiquid tokens
## Summary

In the Swap flow, the destination ("To") token's fiat value can be far larger than the source token's fiat — e.g. swapping **ETH → BINU on Base** shows source `0.00317408 ETH = $5.26` but destination `35.9M BINU = $13.18` (~2.5× higher). A swap is value-preserving, so the two fiat figures should be roughly equal (minus fees), never 2.5× apart.

This is **not** a decimals or arithmetic bug — the conversion is correct. It's a **price-source mismatch**.

## Root cause (verified)

The two sides of the swap are valued with two independent price oracles:

- **Source fiat** = `source amount × source token market price` (ETH — liquid, accurate).
- **Destination fiat** = `dst amount × dst token market price` (CoinGecko, via `convertTokenValueToFiat(dstToken, expectedDstValue, currency)`).

The **swap rate itself** (how many dst tokens you receive) is set by the **DEX pool** the aggregator routes through — not by the market oracle. For liquid pairs the pool rate and the market price agree, so dst fiat ≈ src fiat. For illiquid/meme tokens they diverge sharply.

### BINU example — numbers confirmed
- CoinGecko market price for the BINU contract on Base = `$3.67153e-7`.
- `35,900,000 BINU × $3.67153e-7 = $13.18` — exactly the value shown on screen, so the app's conversion is correct.
- The quote's implied execution rate is `$5.26 / 35.9M ≈ $1.46e-7` per token.
- CoinGecko's mark is ~2.5× the rate you can actually trade at, so the screen implies a 2.5× "free lunch" that doesn't exist.

## Expected

Destination fiat should reflect the value you actually receive at the quoted rate, so it stays consistent with the source fiat (minus fees).

## Suggested fix

Derive the destination fiat from the **rate the quote uses** rather than an independent market-price lookup. Options:
- Anchor dst fiat to the source fiat (swaps are value-preserving), or
- Use the swap provider's own reported USD value for the destination amount, if available in the quote response.

## Steps to reproduce

1. Open Swap, From: **ETH (Base)**, To: an illiquid token whose CoinGecko market price diverges from pool liquidity (e.g. **BINU** on Base).
2. Enter a sell amount (e.g. `0.00317408` ETH).
3. Compare the source fiat vs the destination fiat — destination is materially larger.

## Environment

- Platform: Android
- Chain: Base
- Provider observed in quote: LI.FI

aminsato:
## Summary

Fixes #4878. In the Swap flow the destination ("To") fiat could read far higher than the source fiat for illiquid tokens (e.g. **ETH → BINU on Base**: source `$5.26` vs destination `$13.18`, ~2.5× apart). A swap is value-preserving, so the two fiat figures should be roughly equal — the destination should never imply a "free lunch."

## Root cause

Destination fiat was computed with an independent market oracle (CoinGecko):

```kotlin
convertTokenValueToFiat(dstToken, quote.expectedDstValue, currency) // dstAmount × dstMarketPrice
```

For illiquid/meme tokens that mark diverges sharply from the DEX pool rate the aggregator actually routes through, inflating the destination fiat above the source.

## Fix

- The **displayed** destination fiat is now **clamped to the source fiat** when the market value exceeds it — it reflects the rate the quote executes at, not a stale independent oracle.
- Genuine price impact / fees that push the destination *below* the source pass through **unclamped** — that divergence is real and shouldn't be hidden (a clamp, not a hard `dst = src` anchor).
- The unclamped market value is preserved on `estimatedDstFiat`, used only for **cross-provider ranking** (`maxBy { estimatedDstFiat.value }`). Anchoring that would make providers tie and break route selection, so display and ranking are kept separate.
- Guarded on `srcFiatValue > 0` so a cold source price doesn't zero out the destination display.

> Note: the alternative of using the provider's own reported USD value isn't viable — none of the quote response models (LI.FI / 1inch / Kyber / THORChain / Maya / SwapKit) return a fiat figure, only the destination token amount.

## Test plan

- Added unit test: inflated destination clamps to source fiat (`$13.18` → displayed `$5.26`) while ranking still sees `$13.18`.
- Added unit test: a legitimately-below-source value (`$95` vs `$100` source) passes through unclamped.
- All existing `SwapQuoteManagerTest` ranking/quote tests still pass (`:app:testDebugUnitTest`).

## Proof (real device)

ETH → BINU on Base — source and destination fiat now match (`$0.84` / `$0.84`):

proof

On-chain transaction: https://basescan.org/tx/0xad5bcdd55273cfee44f277a36fb1d211316e072a27e92b61558a23c11390aa9e

## Summary by CodeRabbit

* **Bug Fixes**
  * Displayed destination fiat is clamped to the source fiat when market valuations would otherwise show a higher value; underlying market values remain available for cross-provider ranking.
  * UI/confirmation screens now consistently show the clamped destination fiat to prevent inflated marks from appearing.

* **Tests**
  * Added tests verifying clamped display behavior and preservation of unclamped market values for ranking.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -51,7 +51,9 @@ internal data class QuoteFetchResult(
     val providerUiText: UiText,
     val srcFiatValueText: String,
     val estimatedDstTokenValue: String,
+    // Displayed destination fiat: market value clamped to the source fiat (#4878).
     val estimatedDstFiatValue: String,
+    // Unclamped market value (dstAmount × dstMarketPrice) used only for cross-provider ranking.
     val estimatedDstFiat: FiatValue,
     val feeText: String,
     val swapFeeFiat: FiatValue,
@@ -290,8 +292,25 @@ constructor(
 
         val fiatFees = convertTokenValueToFiat(feeCoin, quote.fees, currency)
         val estimatedDstTokenValue = mapTokenValueToDecimalUiString(quote.expectedDstValue)
+
+        // `expectedDstValue × dstMarketPrice` values the destination at an independent oracle
+        // (e.g. CoinGecko), which for illiquid tokens diverges from the DEX pool rate the quote
+        // actually executes at — inflating the destination fiat above the source (#4878). A swap
+        // is value-preserving, so you can never receive more fiat than you put in; clamp the
+        // *displayed* destination fiat to the source fiat so it reflects the quoted rate. Genuine
+        // price impact / fees that push the destination below the source are real and pass through
+        // unclamped. The unclamped market value is still used for cross-provider ranking below,
+        // where it scales with the destination amount and stays comparable across providers.
+        val marketDstFiatValue = convertTokenValueToFiat(dstToken, quote.expectedDstValue, currency)
         val estimatedDstFiatValue =
-            convertTokenValueToFiat(dstToken, quote.expectedDstValue, currency)
+            if (
+                srcFiatValue.value > BigDecimal.ZERO &&
+                    marketDstFiatValue.value > srcFiatValue.value
+            ) {
+                FiatValue(value = srcFiatValue.value, currency = marketDstFiatValue.currency)
+            } else {
+                marketDstFiatValue
+            }
 
         val rawFees =
             when (quote) {
@@ -350,7 +369,7 @@ constructor(
             srcFiatValueText = srcFiatValueText,
             estimatedDstTokenValue = estimatedDstTokenValue,
             estimatedDstFiatValue = fiatValueToString(estimatedDstFiatValue),
-            estimatedDstFiat = estimatedDstFiatValue,
+            estimatedDstFiat = marketDstFiatValue,
             feeText = resolvedFeeText,
             swapFeeFiat = resolvedSwapFeeFiat,
             outboundFeeText = outboundFeeText,
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -295,22 +295,13 @@ constructor(
 
         // `expectedDstValue × dstMarketPrice` values the destination at an independent oracle
         // (e.g. CoinGecko), which for illiquid tokens diverges from the DEX pool rate the quote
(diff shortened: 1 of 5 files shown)


aminsato:
Addressed in 0965d48bc — converted the two new `fetchQuote` clamp tests to Kotest `shouldBe` assertions per the test guideline. Left the pre-existing JUnit `assertEquals` calls in the rest of the file untouched to keep the diff scoped.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -290,8 +292,25 @@ constructor(
...
+        // unclamped. The unclamped market value is still used for cross-provider ranking below,
+        // where it scales with the destination amount and stays comparable across providers.
+        val marketDstFiatValue = convertTokenValueToFiat(dstToken, quote.expectedDstValue, currency)
         val estimatedDstFiatValue =
-            convertTokenValueToFiat(dstToken, quote.expectedDstValue, currency)
+            if (
+                srcFiatValue.value > BigDecimal.ZERO &&
+                    marketDstFiatValue.value > srcFiatValue.value
+            ) {
+                FiatValue(value = srcFiatValue.value, currency = marketDstFiatValue.currency)
```
rkokhatskyi: `SwapTransaction` carries no destination fiat, only `expectedDstTokenValue`, so `SwapTransactionToUiModelMapper.kt:99-106` recomputes the dst fiat unclamped for `VerifySwapViewModel` and `KeysignFlowViewModel`. The clamp reaches only the form, so the inflated #4878 value still shows on the verify and keysign screens the user signs from

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Good catch — fixed in 6b9ae5bfb. Extracted the clamp into a shared `clampDstFiatToSrcFiat()` helper and applied it in `SwapTransactionToUiModelMapper` as well, so the verify and keysign screens (`VerifySwapViewModel` / `KeysignFlowViewModel`) get the value-preserving destination fiat instead of the unclamped market mark. Added mapper unit tests covering the inflated (clamps) and below-source (passes through) cases.

johnnyluo (approved):
LGTM

PR #4897, opened 2026-06-14
fix(swap): min-output limit on THORChain/Maya swaps via tolerance_bps

### LINKED ISSUE #4840: THORChain/Maya swaps signed with no min-output limit (toAmountLimit hardcoded "0", no tolerance_bps)
## Observed

THORChain and MayaChain swaps are signed and broadcast with **no minimum-output (slippage) limit**. If the pool moves adversely between quote and execution — or a sandwich bot front-runs the inbound — the swap still executes at whatever the pool gives, with no on-chain floor protecting the user. The user can receive materially less than the quoted `expected_amount_out` with no revert.

## Cause

Two layers, both unguarded:

1. The quote request never asks THORChain for a tolerance. `ThorChainApiImpl.getSwapQuotes` (`data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt:221-246`) sends `from_asset`, `to_asset`, `amount`, `destination`, `streaming_interval`, optional `streaming_quantity`, and affiliate params — but **no `tolerance_bps`**.

2. The signed payload hardcodes the limit to zero. `SwapTransactionBuilder.build()` (`app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt`) sets, for THORChain (line 104) and MayaChain (line 174):

```kotlin
toAmountLimit = "0",
streamingInterval = "1",   // "3" for Maya
streamingQuantity = "0",
```

`toAmountLimit` flows verbatim into the signed proto via `PayloadToProtoMapper.kt:180,198`, and the native-deposit memo is signed as-is. A `"0"` limit means "accept any output". (`SwapTransactionBuilderTest.kt:110,157` actively pins this to `"0"`.)

## Fix shape

- Derive a min-output from the quote's `expected_amount_out` minus a slippage budget (e.g. the user's configured tolerance, default ~1-3%), and write it into `toAmountLimit` instead of `"0"`. THORChain memos already carry an optional limit field (`:LIM`); populate it.
- Optionally pass `tolerance_bps` to `/thorchain/quote/swap` so the returned memo already embeds the limit, then sign that memo rather than rebuilding it.
- Update the tests that pin `toAmountLimit == "0"` to assert a non-zero, quote-derived floor.

---

aminsato:
## Problem

THORChain and MayaChain swaps were signed and broadcast with **no minimum-output (slippage) floor**. If the pool moved adversely between quote and execution — or a sandwich bot front-ran the inbound — the swap still executed at whatever the pool gave, with no on-chain revert. The user could receive materially less than the quoted `expected_amount_out`. (#4840)

## Root cause

The on-chain floor is the `:LIM` field the node bakes into the swap **memo**, which Android signs verbatim (`quote.data.memo`). Neither `ThorChainApi.getSwapQuotes` nor `MayaChainApi.getSwapQuotes` sent `tolerance_bps`, so the returned memo carried **no limit** (= accept any output).

The hardcoded `toAmountLimit = "0"` in `SwapTransactionBuilder` flagged in the issue is a red herring — verified that nothing in the signing helpers (`THORChainSwaps`, chain/crypto helpers) reads that proto field. The limit lives in the memo.

## Fix

Send `tolerance_bps = 100` (1%) on every THORChain/Maya quote request, mirroring iOS (`SwapService.defaultThorchainToleranceBps`). The node then bakes `LIM = expected_amount_out × (1 − tolerance_bps/10_000)` into the memo we sign.

- `ThorChainSwapQuoteRequest` → add `toleranceBps` (preserved through the streaming `.copy()`)
- `ThorChainApi` / `MayaChainApi` → emit `tolerance_bps` when present
- `DEFAULT_THORCHAIN_TOLERANCE_BPS = 100` shared constant (aligned with the existing `STREAMING_SLIPPAGE_THRESHOLD_BPS`)
- Clarifying comments on the now-cosmetic `toAmountLimit = "0"`

The 1% default matches the existing streaming-upgrade threshold; there is no per-swap user slippage control yet (a follow-up could surface one).

## Test plan

- [x] New unit tests assert both the **rapid** and **streaming-fallback** THORChain requests carry `toleranceBps == 100`
- [x] Updated Maya provider-test mock stubs for the new param
- [x] `:data` affected suites green (`SwapQuoteRepositoryThorChainStreamingTest`, `SwapQuoteRepositoryProvidersTest`, `MayaChainApiBodyReadTest`)
- [x] ktfmt applied
- [ ] Manual mainnet smoke: confirm a THORChain and a Maya swap memo now contains a non-zero `:LIM`

Closes #4840

## Summary by CodeRabbit

* **New Features**
  * Added slippage protection to swap transactions with a configurable tolerance parameter that enforces a minimum output floor.
  * Default tolerance is set to 100 basis points (1%) to guard against adverse pool price movements between quote and execution.

* **Tests**
  * Added test coverage verifying tolerance parameters are correctly included in swap quote requests.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
@@ -101,6 +101,10 @@ constructor(
                                 routerAddress = quote.data.router,
                                 fromAmount = srcTokenValue.value,
                                 toAmountDecimal = dstTokenValue.decimal,
+                                // The on-chain min-output floor is enforced by the `:LIM` field
+                                // the node bakes into `quote.data.memo` (signed verbatim) when the
+                                // quote request carries `tolerance_bps`; this proto field is not
+                                // read at sign time, so it stays "0".
                                 toAmountLimit = "0",
                                 streamingInterval = "1",
                                 streamingQuantity = "0",
@@ -171,6 +175,9 @@ constructor(
                                 routerAddress = quote.data.router,
                                 fromAmount = srcTokenValue.value,
                                 toAmountDecimal = dstTokenValue.decimal,
+                                // See ThorChain branch: the real floor is the memo `:LIM` the node
+                                // adds from `tolerance_bps`; this proto field is unused at sign
+                                // time.
                                 toAmountLimit = "0",
                                 streamingInterval = "3",
                                 streamingQuantity = "0",
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/MayaChainApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/MayaChainApi.kt
@@ -52,6 +52,7 @@ interface MayaChainApi {
         isAffiliate: Boolean,
         bpsDiscount: Int,
         referralCode: String,
+        toleranceBps: Int? = null,
     ): THORChainSwapQuoteDeserialized
 
     suspend fun broadcastTransaction(tx: String): String?
@@ -188,6 +189,7 @@ constructor(
         isAffiliate: Boolean,
         bpsDiscount: Int,
         referralCode: String,
+        toleranceBps: Int?,
     ): THORChainSwapQuoteDeserialized {
         try {
             val affiliateFeeRate =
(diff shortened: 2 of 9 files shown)


johnnyluo (approved):
LGTM

PR #4898, opened 2026-06-14
fix(swap): use standard EVM swap gas limit on Mantle

johnnyluo:
## Summary

Mantle swaps used a dedicated `DEFAULT_MANTLE_SWAP_LIMIT = 3,000,000,000` gas limit in three places, while every other EVM chain uses the standard `DEFAULT_SWAP_LIMIT` (600,000). This PR removes the Mantle special-casing so Mantle is treated like the other EVM chains:

- **`EthereumFeeService.getDefaultLimit`** — Mantle swaps now use `DEFAULT_SWAP_LIMIT` for the fee estimate/bond.
- **`SwapTransactionBuilder`** — the signed swap tx keeps the aggregator's quoted gas (`quote.data.tx.gas`) instead of overriding it on Mantle.
- **`JoinKeysignFeeUtils`** — the joiner fee uses `DEFAULT_SWAP_LIMIT`; the now-redundant `defaultEvmSwapGasLimit(chain)` helper and its `chain` plumbing are removed.
- Dropped the unused `DEFAULT_MANTLE_SWAP_LIMIT` constant.

## ⚠️ Note / risk

The removed code was deliberate: Mantle's per-gas units run ~an order of magnitude higher than other EVM chains, and the 3B limit existed to keep the **displayed fee bond** consistent with the actual on-chain swap cost (and to keep the joiner aligned with the initiator). With this change the displayed network fee on Mantle swaps is sized off 600k, so it may understate the real cost. Proceeding per request to standardize Mantle.

## Test plan

- [x] `:data:testDebugUnitTest --tests EthereumFeeServiceTest` — green
- [x] `:app:testDebugUnitTest --tests SwapTransactionBuilderTest --tests JoinKeysignSendGasFeeTest` — green
- Updated the three affected tests to assert the standard limit / quoted gas for Mantle.

## Summary by CodeRabbit

* **Refactor**
  * Updated join-keysign and swap fee estimation to use the standard Ethereum default swap limit for Ethereum-based swaps, including Mantle (removing Mantle-specific swap gas handling).
  * Improved OneInch swap gas handling: the staged transaction now uses the quote’s gas limit, with a fallback default when the quote gas is zero.
* **Tests**
  * Updated and added unit tests to match the new standard swap-limit behavior and the revised staged gas-limit logic.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
@@ -239,10 +239,9 @@ constructor(
                 val isApprovalRequired = allowance != null && allowance < srcTokenValue.value
 
                 val specific = specificAndUtxo.blockChainSpecific
-                // Aggregators can return tx.gas == 0; fall back to the standard EVM swap unit so
-                // the
-                // signed payload never carries a zero gas limit (matches SwapQuoteManager's fee
-                // path).
+                // Aggregators can return tx.gas == 0; fall back to the standard EVM swap unit
+                // so the signed payload never carries a zero gas limit (matches
+                // SwapQuoteManager's fee path).
                 val gasLimit =
                     quote.data.tx.gas.takeIf { it != 0L } ?: EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT
                 val quoteData =
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
@@ -1,7 +1,5 @@
 package com.vultisig.wallet.ui.models.swap
 
-import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_MANTLE_SWAP_LIMIT
-import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.EVMSwapPayloadJson
 import com.vultisig.wallet.data.models.FiatValue
@@ -240,19 +238,13 @@ constructor(
                 val isApprovalRequired = allowance != null && allowance < srcTokenValue.value
 
                 val specific = specificAndUtxo.blockChainSpecific
-                val gasLimit =
-                    if (srcToken.chain == Chain.Mantle) {
-                        DEFAULT_MANTLE_SWAP_LIMIT.toLong()
-                    } else {
-                        quote.data.tx.gas
-                    }
                 val quoteData =
                     if (specific is BlockChainSpecific.Ethereum) {
                         quote.data.copy(
                             tx =
                                 quote.data.tx.copy(
                                     gasPrice = specific.maxFeePerGasWei.toString(),
-                                    gas = gasLimit,
+                                    gas = quote.data.tx.gas,
                                 )
                         )
                     } else {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
@@ -1,5 +1,6 @@
 package com.vultisig.wallet.ui.models.swap
 
+import com.vultisig.wallet.data.chains.helpers.EvmHelper
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.EVMSwapPayloadJson
 import com.vultisig.wallet.data.models.FiatValue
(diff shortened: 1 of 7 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
```diff
@@ -247,7 +247,6 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
         val chain = transaction.coin.chain
 
         return when {
```
aminsato, line 249: With Mantle falling through to `DEFAULT_SWAP_LIMIT` (600k), the displayed network fee and the native-token amount reserved for gas are now sized off 600k, while Mantle's per-gas units run ~an order of magnitude higher than other EVM chains. The signed tx is fine (it uses the aggregator's real quoted gas), but on a native-MNT swap the reserved-for-gas math could come up short if it keys off the understated display. Worth a sanity check on a real Mantle native swap that spendable-balance reservation isn't squeezed, and a follow-up issue to track the display understatement rather than letting it disappear with the constant.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
```diff
@@ -240,12 +239,12 @@ constructor(
                 val isApprovalRequired = allowance != null && allowance < srcTokenValue.value
 
                 val specific = specificAndUtxo.blockChainSpecific
+                // Aggregators can return tx.gas == 0; fall back to the standard EVM swap unit so
```
aminsato: ktfmt mangled this into an orphaned `// the` line and split `fee` / `// path`. Reflow it next time the file is touched — not worth a re-push on its own.

aminsato (commented):
Clean standardization — the three Mantle branches now size off the same 600k value, so the initiator's displayed fee and the joiner's fee stay aligned (the original reason the special-case existed), and the signed tx now carries the aggregator's real quoted gas instead of a hardcoded 3B, which is strictly more correct on-chain. The new fallback also matches the existing `quote.tx.gas.takeIf { it != 0L } ?: DEFAULT_ETH_SWAP_GAS_UNIT` convention used in OneInchSwap/KyberSwap/SwapQuoteManager. CI is green. One thing worth confirming before merge (inline) plus a cosmetic nit.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
johnnyluo: Verified on a real Mantle native-MNT swap: the standard `DEFAULT_SWAP_LIMIT` (600k) produces a correct displayed network fee and an adequate native-gas reservation — the "max %" reservation isn't squeezed. The historical premise that Mantle's per-gas usage runs ~an order of magnitude higher no longer holds, so the dedicated 3B `DEFAULT_MANTLE_SWAP_LIMIT` isn't needed. Keeping the standardized limit; no display understatement in practice. Resolving.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
johnnyluo: Reflowed the comment (removed the orphaned `// the` line and the split `fee` / `// path`) in 51691a6c8. Resolving.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
```diff
@@ -240,12 +239,11 @@ constructor(
...
+                // Aggregators can return tx.gas == 0; fall back to the standard EVM swap unit
+                // so the signed payload never carries a zero gas limit (matches
+                // SwapQuoteManager's fee path).
                 val gasLimit =
-                    if (srcToken.chain == Chain.Mantle) {
-                        DEFAULT_MANTLE_SWAP_LIMIT.toLong()
-                    } else {
-                        quote.data.tx.gas
-                    }
+                    quote.data.tx.gas.takeIf { it != 0L } ?: EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT
```
rkokhatskyi, line 253: Removing the Mantle branch drops the 3B signed tx gas floor (#2423), so Android signs Mantle swaps at `max(quote.tx.gas, 600k)` while iOS and Windows keep the 3B floor via `max(aggregatorGas, 3B)`. Windows pins a plain Mantle transfer at 90M gas, so a Mantle swap can run out of gas and revert on Android only

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
johnnyluo: Nope ,  IOS & windows/extension will update it as well

rkokhatskyi (approved):
LGTM

PR #4899, opened 2026-06-14
fix(terra): charge Terra Classic burn tax on USTC/LUNC sends

### LINKED ISSUE #4843: TerraClassic burn-tax absent; flat fee under-pays taxed USTC/LUNC sends
## Observed

Terra Classic (LUNC / USTC) sends use a flat fee and never account for the chain's **burn tax** (a proportional stability/burn tax, ~0.5%, charged on the send amount on top of the gas fee). Because the signed transaction does not reserve the tax, the tx under-pays and is rejected by Terra Classic validators ("insufficient fee" / tax not covered). Users cannot send USTC/LUNC.

(Note: the fee *denom* is correctly `uluna` — this is not a denom bug. The missing piece is the proportional tax math.)

## Cause

`CosmosFeeService.calculateDefaultFees()` (`data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt:57-59`) returns a flat amount for TerraClassic with no reference to the transfer amount:

```kotlin
Chain.TerraClassic -> {
    GasFees(limit = gasLimit.toBigInteger(), amount = 10000000L.toBigInteger())
}
```

There is no `burn_tax` / `tax_rate` computation anywhere in the codebase (grep for `burn.?tax` / `tax.?rate` in `data/src/main` and `app/src/main` returns nothing). The flat `10_000_000 uluna` covers gas but not the amount-proportional tax, so any non-trivial USTC/LUNC send is short.

## Fix shape

- Fetch the active tax rate (`/terra/treasury/v1beta1/tax_rate`) and tax cap (`tax_caps`) for the send denom, compute `taxAmount = min(amount * taxRate, taxCap)`, and add it to the fee `amount` (or to the message, per Terra Classic's current tax mechanism).
- Gate this on `Chain.TerraClassic` only; Terra v2 does not charge burn tax.

---

aminsato:
## Problem

Terra Classic (columbus-5) charges a proportional **burn tax** (~0.5%, `x/tax` module) on every `MsgSend`, paid in the send denom on top of the gas fee. `CosmosFeeService` returned a flat `10,000,000 uluna` with no reference to the send amount, so any non-trivial USTC/LUNC send **under-paid and was rejected** by validators (`insufficient fee`). Users could not send USTC/LUNC. (#4843)

## Fix

Fold the burn tax into the single Cosmos fee field, mirroring `vultisig-ios` (`TerraClassicTax` / `BlockChainService`):

- **`TerraClassicTax`** (new) — pure helper: ceil-rounded tax math (so the signed fee never undershoots), fail-closed `0.5%` fallback rate, bank-denom classification, and per-denom base gas (`uluna` 8,497,500 / `uusd` 225,000).
- **`CosmosApi.getTerraClassicBurnTaxRate()`** — live rate from `/terra/tax/v1beta1/params` → `params.burn_tax_rate`; returns `null` on any failure so the caller falls back conservatively rather than signing an under-funded tx.
- **`CosmosFeeService`** — TerraClassic fee = `baseGas(coin) + burnTax(amount, rate)`, folding the tax only when it is paid in the send denom (native LUNC → uluna, USTC bank denom → uusd). CW20 (`terra1…`) / IBC (`ibc/…`) pay the fee in uluna while the send is in the token's own denom, so they get base gas only to avoid mixing denoms.
- **`TerraHelper`** — sign the USTC bank-denom fee as a **single `uusd` coin = tax-folded `gas`** (was a two-coin `uluna=gas` + hardcoded `uusd=1_000_000`), so the signed fee denom matches the base-gas denom. Native LUNC / CW20 branches already read `gas` and pick up the tax automatically.

The displayed fee and signed fee both derive from `CosmosFeeService`, so they stay consistent. MAX-send estimates the tax on a slightly larger amount than is ultimately sent, so the fee always covers the real tax (fail-safe — sends marginally less rather than over-committing).

## Test plan

- [x] `TerraClassicTaxTest` — tax math, ceil rounding, fallback (null/garbage/negative), denom classification, base-gas selection
- [x] `CosmosFeeServiceTest` — LUNC (uluna base + tax), USTC (uusd base + tax), CW20 (base only, no tax), null-rate fallback; updated for the new constructor
- [x] `CosmosApiTest` — endpoint path + `burn_tax_rate` parse, and null-on-failure
- [x] Full `:data` unit suite green; `:app` compiles; ktfmt applied
- [ ] Manual mainnet smoke: send LUNC and USTC, confirm broadcast succeeds (no `insufficient fee` / tax rejection)

Note: for USTC the fee is now correctly paid in `uusd`; the fee *display* still labels in the native LUNC token (pre-existing single-gas-field display nuance, same as iOS) — out of scope for this rejection fix.

Closes #4843

## Release Notes

* **New Features**
  * Terra Classic transactions now use a live-fetched burn tax rate when estimating fees.

* **Bug Fixes**
  * Improved Terra Classic fee calculations for LUNC vs USTC bank denoms, including correct proportional burn tax behavior and safer burn-tax rate handling.
  * Fee reservation and balance checks now account for fee-paying tokens more accurately.
  * Better fallback to a conservative default rate when live data isn’t available.

* **Tests**
  * Added/updated unit tests for Terra Classic tax rate parsing and fee computations, and refreshed the related expected image hash.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
@@ -1,12 +1,15 @@
 package com.vultisig.wallet.data.blockchain.cosmos
 
+import com.vultisig.wallet.data.api.CosmosApiFactory
 import com.vultisig.wallet.data.blockchain.FeeService
 import com.vultisig.wallet.data.blockchain.model.BlockchainTransaction
 import com.vultisig.wallet.data.blockchain.model.Fee
 import com.vultisig.wallet.data.blockchain.model.GasFees
+import com.vultisig.wallet.data.blockchain.model.Transfer
 import com.vultisig.wallet.data.models.Chain
+import java.math.BigInteger
 
-class CosmosFeeService : FeeService {
+class CosmosFeeService(private val cosmosApiFactory: CosmosApiFactory) : FeeService {
 
     companion object {
         internal const val OSMOSIS_MIN_FEE_UOSMO = 25_000L
@@ -55,12 +58,39 @@ class CosmosFeeService : FeeService {
                 GasFees(limit = gasLimit.toBigInteger(), amount = 20000L.toBigInteger())
             }
             Chain.TerraClassic -> {
-                GasFees(limit = gasLimit.toBigInteger(), amount = 10000000L.toBigInteger())
+                GasFees(
+                    limit = gasLimit.toBigInteger(),
+                    amount = terraClassicFeeAmount(transaction),
+                )
             }
             Chain.Dydx -> {
                 GasFees(limit = gasLimit.toBigInteger(), amount = 2500000000000000L.toBigInteger())
             }
             else -> error("Chain Not Supported: ${chain.name}")
         }
+    }
+
+    /**
+     * Terra Classic fee = base gas + proportional burn tax. The base gas is denominated in the same
+     * denom the signer uses for the fee (uluna for native LUNC / CW20 / IBC, uusd for the USTC bank
+     * denom — see [TerraClassicTax.baseGas] and `TerraHelper`). The burn tax (~0.5%, fetched live
+     * with a conservative fallback) is folded into this single fee field only when it is paid in
+     * the send denom (native LUNC or USTC); CW20/IBC sends, whose fee is in uluna while the send is
+     * in the token's own denom, get base gas only to avoid mixing denoms.
+     */
+    private suspend fun terraClassicFeeAmount(transaction: BlockchainTransaction): BigInteger {
+        val coin = transaction.coin
+        val base = TerraClassicTax.baseGas(coin.contractAddress, coin.isNativeToken).toBigInteger()
+
+        val taxable =
+            transaction is Transfer &&
+                TerraClassicTax.taxPaidInSendDenom(coin.contractAddress, coin.isNativeToken) &&
+                transaction.amount > BigInteger.ZERO
+        if (!taxable) return base
+
+        val rawRate =
+            cosmosApiFactory.createCosmosApi(Chain.TerraClassic).getTerraClassicBurnTaxRate()
+        val rate = TerraClassicTax.parseRate(rawRate)
+        return base + TerraClassicTax.burnTax(transaction.amount, rate)
     }
 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
@@ -7,12 +7,25 @@ import com.vultisig.wallet.data.blockchain.model.Fee
 import com.vultisig.wallet.data.blockchain.model.GasFees
 import com.vultisig.wallet.data.blockchain.model.Transfer
 import com.vultisig.wallet.data.models.Chain
+import java.math.BigDecimal
 import java.math.BigInteger
 
 class CosmosFeeService(private val cosmosApiFactory: CosmosApiFactory) : FeeService {
 
+    /**
+     * Terra Classic burn rate only changes via governance, but [calculateFees] runs on every
+     * debounced keystroke during fee display. Cache the parsed rate for a short TTL so the hot path
+     * doesn't hit `/terra/tax/v1beta1/params` on each keypress. A single `@Volatile` reference
+     * makes the read/publish atomic; concurrent refreshes are harmless (idempotent fetch).
+     */
+    @Volatile private var cachedBurnTaxRate: CachedBurnTaxRate? = null
+
+    private data class CachedBurnTaxRate(val rate: BigDecimal, val fetchedAtMs: Long)
+
     companion object {
         internal const val OSMOSIS_MIN_FEE_UOSMO = 25_000L
+
+        private const val BURN_TAX_RATE_TTL_MS = 5 * 60 * 1000L
 
         // `gasLimit * gasPrice` lands below what Akash's validators accept: the
         // chain-registry price (0.025 uakt/gas) on a ~300k-gas delegation yields
@@ -88,9 +101,16 @@ class CosmosFeeService(private val cosmosApiFactory: CosmosApiFactory) : FeeServ
                 transaction.amount > BigInteger.ZERO
         if (!taxable) return base
 
+        return base + TerraClassicTax.burnTax(transaction.amount, burnTaxRate())
+    }
+
+    private suspend fun burnTaxRate(): BigDecimal {
+        val now = System.currentTimeMillis()
+        cachedBurnTaxRate?.let { if (now - it.fetchedAtMs < BURN_TAX_RATE_TTL_MS) return it.rate }
         val rawRate =
             cosmosApiFactory.createCosmosApi(Chain.TerraClassic).getTerraClassicBurnTaxRate()
         val rate = TerraClassicTax.parseRate(rawRate)
-        return base + TerraClassicTax.burnTax(transaction.amount, rate)
+        cachedBurnTaxRate = CachedBurnTaxRate(rate, now)
+        return rate
     }
 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/TerraClassicTax.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/TerraClassicTax.kt
@@ -1,0 +1,89 @@
+package com.vultisig.wallet.data.blockchain.cosmos
+
+import java.math.BigDecimal
+import java.math.BigInteger
+import java.math.RoundingMode
+
+/**
+ * Terra Classic (columbus-5) charges a proportional **burn tax** on every `MsgSend`, paid in the
+ * send denom on top of the gas fee. The rate lives in the chain's `x/tax` module (`burn_tax_rate`,
+ * currently 0.5%) and is fetched live; this helper holds the conservative fallback and the pure tax
+ * math so the signed fee, the validated fee, and the displayed fee stay consistent.
+ *
+ * Ported from vultisig-ios `TerraClassicTax`.
+ */
+object TerraClassicTax {
+
+    /**
+     * Conservative fallback burn-tax rate used when the live `x/tax` params can't be
+     * fetched/decoded. Matches current governance (0.5%). Failing closed (taxing) rather than open
+     * (0%) avoids signing a tx the chain then rejects at broadcast.
+     */
+    val fallbackBurnTaxRate: BigDecimal = BigDecimal("0.005")
+
+    /**
+     * Base gas fee for an `uluna`-denominated Terra Classic send (300k gas × 28.325 uluna/gas =
+     * 8,497,500 uluna ≈ 8.5 LUNC). Paid by native LUNC, CW20 (`terra1…`) and IBC (`ibc/…`) tokens,
+     * whose fee the signer denominates in `uluna`.
+     */
+    const val ULUNA_BASE_GAS: Long = 8_497_500L
+
+    /**
+     * Base gas fee for a `uusd`-denominated Terra Classic send (300k gas × 0.75 uusd/gas = 225,000
+     * uusd). Paid only by the USTC bank denom, whose fee the signer denominates in `uusd`.
+     */
+    const val UUSD_BASE_GAS: Long = 225_000L
+
+    /**
+     * Burn tax on a send [amount] (in the denom's smallest unit) at [rate], rounded **up** so the
+     * signed fee never undershoots the chain's check. Returns 0 for a non-positive amount or rate.
+     */
+    fun burnTax(amount: BigInteger, rate: BigDecimal): BigInteger {
+        if (amount <= BigInteger.ZERO || rate <= BigDecimal.ZERO) return BigInteger.ZERO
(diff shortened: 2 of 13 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TerraHelper.kt
```diff
@@ -139,19 +144,13 @@ class TerraHelper(
                         )
                         .setFee(
                             Cosmos.Fee.newBuilder()
-                                .setGas(1000000)
+                                .setGas(gasLimit)
                                 .addAmounts(
                                     Cosmos.Amount.newBuilder()
-                                        .setDenom("uluna")
+                                        .setDenom(keysignPayload.coin.contractAddress)
```
johnnyluo, line 150: why `setDenom` to `keysignPayload.coin.contractAddress` ?

johnnyluo:
Please attach tx links as well

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TerraHelper.kt
```diff
@@ -139,19 +144,13 @@ class TerraHelper(
                         )
                         .setFee(
                             Cosmos.Fee.newBuilder()
-                                .setGas(1000000)
+                                .setGas(gasLimit)
                                 .addAmounts(
                                     Cosmos.Amount.newBuilder()
```
johnnyluo, line 149: **Functional risk — please verify on mainnet.** Now that the USTC fee is signed purely in `uusd` (this line), the fee no longer comes out of `uluna`/LUNC. But the non-native send validation in `DefaultSendStrategy.kt:265` still gates on the native balance:

```kotlin
} else if (nativeTokenValue < gasFee.value) { // LUNC balance vs a uusd-denominated number
```

Previously USTC paid part of the fee in `uluna` (the old hardcoded `10_000_000`), so requiring LUNC was correct. After this change a USTC holder with little/no LUNC could be **incorrectly blocked** from a send the chain would actually accept (fee is deducted from their USTC). This is the "single-gas-field nuance" noted in the PR, but here it affects *validation behavior*, not just the displayed label. The unchecked manual-smoke-test box is exactly the case to exercise: send USTC from a vault with ~0 LUNC.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
```diff
@@ -55,12 +58,39 @@ class CosmosFeeService : FeeService {
...
+
+        val taxable =
+            transaction is Transfer &&
+                TerraClassicTax.taxPaidInSendDenom(coin.contractAddress, coin.isNativeToken) &&
+                transaction.amount > BigInteger.ZERO
+        if (!taxable) return base
+
+        val rawRate =
+            cosmosApiFactory.createCosmosApi(Chain.TerraClassic).getTerraClassicBurnTaxRate()
+        val rate = TerraClassicTax.parseRate(rawRate)
```
johnnyluo, line 112: **Minor (perf):** this adds a network call to `/terra/tax/v1beta1/params` on every debounced keystroke during fee calculation (other Cosmos chains stay purely local). The burn rate only changes via governance, so consider caching it per-session / short TTL to avoid the latency and network dependency in the fee-display hot path. Non-blocking — the `null` fallback keeps it safe.

johnnyluo (commented):
Inline notes from the review — the core fee math and signing changes look correct and well-tested. A few items to consider, the first being the one I'd resolve before merge.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/TerraClassicTax.kt
```diff
@@ -0,0 +1,89 @@
...
+     * 8,497,500 uluna ≈ 8.5 LUNC). Paid by native LUNC, CW20 (`terra1…`) and IBC (`ibc/…`) tokens,
+     * whose fee the signer denominates in `uluna`.
+     */
+    const val ULUNA_BASE_GAS: Long = 8_497_500L
+
+    /**
+     * Base gas fee for a `uusd`-denominated Terra Classic send (300k gas × 0.75 uusd/gas = 225,000
+     * uusd). Paid only by the USTC bank denom, whose fee the signer denominates in `uusd`.
+     */
+    const val UUSD_BASE_GAS: Long = 225_000L
```
johnnyluo, line 43: **Minor:** `baseGas` returns `UUSD_BASE_GAS` for *any* bank denom, and `TerraHelper` signs the fee in `coin.contractAddress`. Fine for the three default tokens (only `uusd` is a bank denom), but a custom non-`uusd` bank denom (e.g. `ukrw`) would get the `uusd` gas price in its own denom and be mis-priced. A short comment noting "only `uusd` is supported among bank denoms" or a denom guard would future-proof this.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/TerraClassicTax.kt
```diff
@@ -0,0 +1,89 @@
...
+        return BigDecimal(amount).multiply(rate).setScale(0, RoundingMode.CEILING).toBigInteger()
+    }
+
+    /**
+     * Parse a decimal-string `burn_tax_rate` from the LCD into a [BigDecimal], falling back to the
+     * conservative default on null input or any parse failure / negative value.
+     */
+    fun parseRate(raw: String?): BigDecimal {
+        if (raw == null) return fallbackBurnTaxRate
+        return runCatching { BigDecimal(raw.trim()) }.getOrNull()?.takeIf { it >= BigDecimal.ZERO }
```
johnnyluo: **Nit:** `parseRate` guards the lower bound (`>= ZERO`) but has no upper bound — a parseable-but-garbage value like `"5"` would be accepted as a 500% tax and massively overcharge. The endpoint is trusted so this is very unlikely, but a sanity cap (e.g. reject `> 0.1`) would make the fail-safe symmetric.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
```diff
@@ -55,12 +58,39 @@ class CosmosFeeService : FeeService {
...
+     * Terra Classic fee = base gas + proportional burn tax. The base gas is denominated in the same
+     * denom the signer uses for the fee (uluna for native LUNC / CW20 / IBC, uusd for the USTC bank
+     * denom — see [TerraClassicTax.baseGas] and `TerraHelper`). The burn tax (~0.5%, fetched live
+     * with a conservative fallback) is folded into this single fee field only when it is paid in
+     * the send denom (native LUNC or USTC); CW20/IBC sends, whose fee is in uluna while the send is
+     * in the token's own denom, get base gas only to avoid mixing denoms.
+     */
+    private suspend fun terraClassicFeeAmount(transaction: BlockchainTransaction): BigInteger {
+        val coin = transaction.coin
+        val base = TerraClassicTax.baseGas(coin.contractAddress, coin.isNativeToken).toBigInteger()
```
rkokhatskyi, line 96: USTC max sends overspend because the fee is now paid in `uusd` from the send balance yet `GetAvailableTokenBalanceUseCase` reserves no fee for non-native tokens, so the signed `amount + fee` exceeds the balance and the chain rejects it, while iOS reserves it for bank denoms via `terraClassicMaxValue`

rkokhatskyi (dismissed):

aminsato:
@johnnyluo mainnet tx proof for a large LUNC send (where the flat-fee main build would be rejected for insufficient fee):

**LUNC send — 6,217.99 LUNC**
https://finder.terra.money/classic/tx/8C1C3B29FE3141CEFB749C4AA0781B0E1A553EADF0F14888A88B81EFB2B88DCD

Signed fee = `39,587,446 uluna` = base gas `8,497,500` + burn tax `ceil(6,217,989,136 × 0.005) = 31,089,946`, accepted (`code: 0`). The explorer decomposes it into Transaction fee `8.497501 LUNC` (gas) + Tax `31.089945 LUNC` (burn). On main this same send signs only the flat `10 LUNC` fee — far below the ~31 LUNC burn tax — and is rejected.

Note: the USTC (`uusd`) bank-denom path and the ~0-LUNC validation case are not device-verified yet (test vault holds LUNC + a discovered UUSDC bank denom, no USTC).

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
aminsato: Good catch — fixed in 967f2834a. `GetAvailableTokenBalanceUseCase` now reserves the fee out of the token's own balance for Terra Classic bank denoms (gated on `chain == TerraClassic && isBankDenom(...)`, so EVM/other non-native tokens with a non-`terra1` contract address are not misread as bank denoms), mirroring iOS `terraClassicMaxValue`. `AmountFractionManager` also routes bank denoms through the fresh-fee recompute so MAX reserves base gas **plus** the amount-proportional burn tax (not just base). The reserve estimates the tax on a slightly larger amount than is finally sent, so `amount + fee <= balance` holds (fail-safe).

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TerraHelper.kt
aminsato: Fixed in 967f2834a. `DefaultSendStrategy` now validates Terra Classic bank-denom sends against the token's **own** balance covering `amount + fee` (in `uusd`), instead of requiring a native LUNC balance for a `uusd`-denominated fee — so a USTC holder with ~0 LUNC is no longer wrongly blocked. CW20/IBC keep the native-LUNC gate. Verified the native LUNC path end-to-end on mainnet (see tx link comment); the ~0-LUNC USTC case is not device-verified yet (test vault has no USTC).

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/CosmosFeeService.kt
aminsato: Fixed in 967f2834a. `CosmosFeeService` now caches the parsed burn rate in a `@Volatile` field with a 5-minute TTL, so the fee hot path no longer hits `/terra/tax/v1beta1/params` on every debounced keystroke. The `null` fallback path is unchanged.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/TerraClassicTax.kt
aminsato: Fixed in 967f2834a. Added `maxBurnTaxRate = 0.1` — `parseRate` now falls back to the conservative default for any value `< 0` **or** `> 0.1`, so a garbage `"5"` (500%) can no longer be applied. Covered by a new test case.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/cosmos/TerraClassicTax.kt
aminsato: Added the clarifying comment in 967f2834a (`baseGas` now notes that only `uusd` is supported among bank denoms — it's the only non-native bank token in the curated Terra Classic coin list). Heads-up: with the `CosmosBankCoinFinder` auto-discovery (#4500) a wallet can now hold an *uncurated* bank denom (e.g. a `uusdc`), which this branch would mis-price. A hard denom guard / per-denom gas price for discovered bank denoms is out of scope here and left as a known limitation (the curated USTC/`uusd` + LUNC scope is correct).

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TerraHelper.kt
aminsato: For Terra Classic bank-denom tokens, `coin.contractAddress` holds the bank denom itself (USTC → `"uusd"`), not a CW20 contract address. This branch is only reached for bank denoms — the guard above excludes `terra1…` (CW20) and `ibc/…` — so `contractAddress` is the `uusd` denom. Setting the fee denom to it makes the fee (base gas + burn tax, precomputed in `CosmosFeeService` and carried in `atomData.gas`) get paid in the send token's own denom, matching iOS `TerraHelperStruct`.

data/src/test/kotlin/com/vultisig/wallet/data/blockchain/cosmos/TerraClassicTaxTest.kt
```diff
@@ -0,0 +1,83 @@
...
+            TerraClassicTax.parseRate("0.005000000000000000"),
+        )
+    }
+
+    @Test
+    fun `parseRate falls back on null, blank, garbage and negative`() {
+        assertEquals(TerraClassicTax.fallbackBurnTaxRate, TerraClassicTax.parseRate(null))
+        assertEquals(TerraClassicTax.fallbackBurnTaxRate, TerraClassicTax.parseRate("not-a-number"))
+        assertEquals(TerraClassicTax.fallbackBurnTaxRate, TerraClassicTax.parseRate("-0.1"))
+    }
```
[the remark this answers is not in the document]
aminsato: Done in 967f2834a — added explicit `parseRate("")` and `parseRate("   ")` assertions to the fallback test.

johnnyluo (approved):
LGTM

PR #4903, opened 2026-06-15
fix(swap): don't deduct LI.FI fee from source balance on percentage tap

### LINKED ISSUE #4902: [Fix] LI.FI percentage/MAX buttons falsely report insufficient balance and clear the amount [Swap]
## Problem

On the Swap screen, tapping a percentage chip (25/50/75) or MAX while the active route is a LI.FI quote can wrongly report "insufficient balance" and clear the amount field, even when the source-token balance is more than enough. It breaks the convenience path for a common case — a low-decimal source token (e.g. USDC, 6 decimals) swapped into a higher-decimal destination token (e.g. an 18-decimal ERC-20). Manual amount entry still works, so funds aren't at risk, but the buttons are misleading and effectively unusable for those pairs.

## Expected Behavior

The percentage/MAX buttons set a fraction of the source-token balance, the same as manual entry and the iOS/desktop apps. The LI.FI service fee comes out of the amount you receive (it is already reflected in the quoted output), so it must never be deducted from the source balance.

## Steps to Reproduce

1. Open Swap and pick a LI.FI-routed pair where the source token has fewer decimals than the destination (e.g. USDC → an 18-decimal token).
2. Enter an amount so a quote loads.
3. Tap 25% / 50% / 75% / MAX.
4. The field clears and an "insufficient balance" error appears, even though the balance is sufficient.

## Solution

In `SwapFormViewModel.selectSrcPercentage`, stop subtracting the quote's swap fee from the source balance. For LI.FI the quote fee is denominated in the **destination** token's raw units, so subtracting it from a source-token raw balance mixes units/decimals — for a low-decimal source into a high-decimal destination it can far exceed the balance and drive the usable amount negative, which trips the insufficient-balance path. The source-side amount should account only for the source-chain network fee on a native source (matching iOS/desktop), since the swap fee is already taken from the destination amount.

### Must Do
- [ ] Remove the swap-fee subtraction from the source balance in `selectSrcPercentage`.
- [ ] Keep the existing native-source network-fee handling (only when the gas chain matches the source chain).
- [ ] Add a regression test covering a LI.FI quote with a low-decimal source into a high-decimal destination.

### Must NOT Do
- Don't change the swap signing/broadcast path or quote construction.
- Don't alter the cross-chain network-fee handling already covered by #4582.
- Don't refactor surrounding swap-form logic or touch unrelated files.

## Acceptance Criteria
- [ ] `./gradlew :app:testDebugUnitTest --tests "*SwapFormViewModelTest*"` passes.
- [ ] A new regression test fails before the change and passes after it.
- [ ] For a LI.FI quote with a 6-decimal source into an 18-decimal destination, a percentage tap sets a non-zero fraction of the balance (no false "insufficient balance", no reset to empty/0).

rkokhatskyi:
## Description

Fixes #4902

On the Swap form, the percentage chips (25/50/75) and MAX read the active quote's swap fee and subtracted it from the source-token balance before computing the amount. For LI.FI the quote fee is denominated in the **destination** token's raw units, so subtracting it from a source-token raw balance mixes decimals. For a low-decimal source into a high-decimal destination (e.g. USDC, 6dp, into an 18-decimal ERC-20) the destination-denominated fee can far exceed the source balance, driving the usable amount negative — which cleared the field and showed a false "insufficient balance" error.

The swap fee is taken out of the destination amount and is already reflected in the quoted output, so it must not be deducted from the source balance — iOS and the desktop app never deduct it. The existing source-chain network-fee handling for a native source is unchanged. Manual amount entry was already correct, so this only brings the buttons in line with it.

## Which feature is affected?
- [x] Swap - amount-input fix only (computed source amount before quoting); the resulting swap is identical to manually typing the same amount, so there is no new on-chain behavior to link

## Checklist

N/A — this is a ViewModel computation fix. The visible effect (the amount field holding a fraction of the balance instead of clearing with an error) is produced by runtime logic against a live LI.FI quote, which the static preview harness can't reproduce faithfully. The behavior is pinned by a unit test that fails before the change (`expected: <250> but was: <>`) and passes after.

## Additional context

Co-sign is unaffected. `selectSrcPercentage` only writes to the shared source-amount field — the same channel manual typing uses — so the staged swap amount, payload construction, session, approval, and broadcast are all identical to entering the value by hand (which already worked). No signing/co-sign code is touched.

Verification: `./gradlew :app:testDebugUnitTest --tests "*SwapFormViewModelTest"` → 90 passed, 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected swap “max usable” calculations to subtract only the source-chain network fee (when the source is native on its own gas chain), preventing incorrect provider fee deductions that could lead to negative or insufficient balances.
  * Refined fee handling to avoid decimal-mixing and ensure the insufficient-balance check is based on the updated usable amount.

* **Tests**
  * Updated swap quote construction to use LI.FI provider-id mapping.
  * Strengthened and added assertions covering source percentage selection and ensuring LI.FI destination-denominated fees aren’t subtracted from low-decimal sources.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -10,7 +10,6 @@ import androidx.navigation.toRoute
 import com.vultisig.wallet.R
 import com.vultisig.wallet.data.models.Address
 import com.vultisig.wallet.data.models.Chain
-import com.vultisig.wallet.data.models.SwapProvider
 import com.vultisig.wallet.data.models.TokenValue
 import com.vultisig.wallet.data.repositories.SwapTransactionRepository
 import com.vultisig.wallet.data.utils.safeLaunch
@@ -340,13 +339,13 @@ constructor(
         // "insufficient balance" warning on a now-valid amount and mask the live quote/formError.
         _uiState.update { it.copy(error = null) }
 
-        val swapFee =
-            quoteState.quote?.fees?.value.takeIf { quoteState.provider == SwapProvider.LIFI }
-                ?: BigInteger.ZERO
-
+        // Only the source-chain network fee may be deducted from the source balance, and only for a
+        // native source on its own gas chain. The provider swap fee is taken from the destination
+        // amount (for LI.FI it is denominated in the destination token's units), so subtracting it
+        // here would mix decimals and could wrongly drive the usable amount negative for a
+        // low-decimal source into a high-decimal destination.
         val maxUsableTokenAmount =
             srcTokenValue.value -
-                swapFee -
                 (quotePipeline.estimatedNetworkFeeTokenValue.value?.value?.takeIf {
                     srcToken.isNativeToken && quotePipeline.gasFeeChain.value == srcToken.chain
                 } ?: BigInteger.ZERO)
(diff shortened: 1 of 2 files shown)


aminsato (approved):
LGTM

PR #4911, opened 2026-06-15
fix(evm): price OP-stack L1 data fee into EVM tx fees

### LINKED ISSUE #4844: OP-stack L1 data fee never added (calculateLayer1Fees returns ZERO); Mantle missing from isLayer2
## Observed

On OP-stack L2s (Optimism, Base, and others), the **L1 data fee** — the dominant cost component of an L2 transaction — is never added to the displayed/reserved fee. The app under-estimates the true cost, so a send/swap can fail to confirm, or a max-send leaves too little to cover the real L1 portion. Mantle is doubly affected: it isn't even classified as an L2.

## Cause

1. `EthereumFeeService.calculateLayer1Fees()` (`data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt:82-84`) is a stub:

```kotlin
private fun calculateLayer1Fees(): BigInteger {
    return BigInteger.ZERO
}
```

It is called for every `isLayer2` chain (lines 43-48 and 99-106) and contributes ZERO, so the OP-stack `GasPriceOracle.getL1Fee` data-fee component is never priced in.

2. Mantle is missing from the L2 set. `Chain.isLayer2` (`data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt:305-314`) lists only `Arbitrum, Base, Blast, Optimism, ZkSync`:

```kotlin
val Chain.isLayer2: Boolean
    get() = when (this) {
        Chain.Arbitrum, Chain.Base, Chain.Blast, Chain.Optimism, Chain.ZkSync -> true
        else -> false
    }
```

`Chain.Mantle` falls to `else -> false`, so even the (currently no-op) L1 path is skipped for it.

## Fix shape

- Implement `calculateLayer1Fees()` by calling the OP-stack `GasPriceOracle` predeploy (`0x420...0F`) `getL1Fee(bytes)` with the RLP-encoded unsigned tx, and add the result to the fee `amount`.
- Add `Chain.Mantle` to `isLayer2` (Mantle is OP-stack-derived and has an L1 fee component).
- Keep Arbitrum out of the OP-stack oracle path — its L1 surcharge is folded into the gas estimate differently.

---

aminsato:
## Summary

Fixes #4844.

On OP-stack L2s (Optimism, Base, Blast, Mantle) the **L1 data fee** — often the dominant cost of an L2 transaction — was never added to the displayed/reserved fee, so the app under-estimated cost and a max-send could leave too little to cover the real L1 portion. `EthereumFeeService.calculateLayer1Fees()` was a stub returning `BigInteger.ZERO`, and **Mantle wasn't even classified as an L2**.

## What changed

- **`Chain.kt`** — add `Chain.Mantle` to `isLayer2`; introduce `Chain.isOpStackL2` (`Optimism, Base, Blast, Mantle`), the set that actually exposes the OP-stack `GasPriceOracle`. Arbitrum and ZkSync are L2s but price their L1 component differently, so they stay out of the oracle path.
- **`EvmApi.getOpStackL1Fee(...)`** — builds the RLP-encoded unsigned EIP-1559 tx from the supplied fields, ABI-encodes `getL1Fee(bytes)`, `eth_call`s the `GasPriceOracle` predeploy (`0x420…0F`) and decodes the wei result.
- **`EthereumRlpEncoder.kt`** (new) — encodes the `0x02 || rlp([chainId, nonce, …, accessList])` unsigned type-2 payload via Wallet Core's `EthereumRlp`.
- **`EthereumFeeService`** — gates the L1 path on `isOpStackL2`, resolves the real broadcast call context per tx type (native = memo, ERC-20 = transfer calldata to the token contract, swap = router calldata), prices the L1 fee **best-effort** (a failed oracle call degrades to zero rather than blocking the whole fee calc), and adds it to the amount.

The JNI-dependent RLP/ABI/oracle work lives behind the `EvmApi` boundary so `EthereumFeeService` stays pure-JVM unit-testable.

## Testing

- `:data:testDebugUnitTest` green — **9 new tests** in `EthereumFeeServiceTest`: L1 fee added to amount; correct `to`/`value`/`chainId`/calldata size for native, ERC-20 and swap txs; Arbitrum/ZkSync/Ethereum do **not** query the oracle; graceful degradation when the oracle call fails.
- `:data:compileDebugKotlin` and `:app:compileDebugKotlin` succeed; ktfmt applied.
- The exact `getL1Fee(bytes)` call this code makes was verified against the production RPCs the app uses (`/base/`, `/opt/`, `/mantle/`) and returns real non-zero values, confirming the RLP envelope + ABI encoding + oracle address are correct.

## Test plan (device)

1. Build/install debug, open **Send** or **Swap** on Base / Optimism / Blast / Mantle with the native token.
2. The network fee on the verify screen now includes the L1 data fee (higher than `main`, where it was always 0).
3. Confirm Arbitrum / ZkSync / Ethereum are unaffected.

## Note

Adding Mantle to `isLayer2` also makes the Mantle native token show the chain-logo overlay on the swap/send verify screens, consistent with the other L2s.

## Summary by CodeRabbit

* **New Features**
  * Added OP-stack L1 data-availability fee estimation for eligible L2 transactions to improve total fee calculations (Base, Blast, Optimism).

* **Bug Fixes**
  * Updated total fee computation to include the OP-stack L1 component for native transfers, ERC-20 transfers, and swaps, with a safe fallback to the standard EIP-1559 fee when estimation fails.
  * Ensured Mantle and non-OP-stack chains don’t use OP-stack L1 fee estimation.

* **Tests**
  * Expanded coverage for oracle call behavior, calldata sizing/handling, and fallback behavior across transaction types and chains.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
@@ -142,7 +142,17 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
                 L1CallContext(
                     to = transaction.to,
                     value = if (coin.isNativeToken) transaction.amount else BigInteger.ZERO,
-                    data = Numeric.hexStringToByteArray(transaction.callData),
+                    // The router calldata is the dominant part of a swap's L1 data fee, but it is
+                    // not known here: gas estimation runs in parallel with the quote fetch, so
+                    // every
+                    // live caller builds the Swap with an empty callData. Pricing the empty payload
+                    // would drop almost the entire L1 cost, so we fall back to a representative
+                    // swap-calldata payload. If a caller ever supplies real calldata, we honour it.
+                    data =
+                        transaction.callData
+                            .takeIf { it.isNotEmpty() }
+                            ?.let { Numeric.hexStringToByteArray(it) }
+                            ?: REPRESENTATIVE_SWAP_CALLDATA,
                 )
             else ->
                 error("Unsupported transaction type for L1 fee: ${transaction::class.simpleName}")
@@ -331,6 +341,18 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
 
     companion object {
         private val GWEI = BigInteger.TEN.pow(9)
+
+        // Stand-in calldata used to price a swap's OP-stack L1 data fee when the real router
+        // calldata is unavailable at estimation time (see resolveL1CallContext). Sized to a typical
+        // DEX/router swap payload (~512 bytes covers a 1inch/Kyber route or a THORChain
+        // depositWithExpiry memo) and filled with non-zero bytes, which the OP-stack fee formula
+        // charges at the higher 16-gas/byte rate — a deliberately conservative estimate so the
+        // reserved fee does not under-cover the real L1 cost. Post-Ecotone/Fjord L1 fees are small
+        // in absolute terms, so the conservative sizing adds negligible cost while avoiding a gross
+        // underestimate.
+        const val REPRESENTATIVE_SWAP_CALLDATA_BYTES = 512
+        private val REPRESENTATIVE_SWAP_CALLDATA =
+            ByteArray(REPRESENTATIVE_SWAP_CALLDATA_BYTES) { 1 }
 
         // Extra bump applied to the committed base fee on the non-swap send path. With the
         // further 20% added in calculateMaxFeePerGas this yields a baseFee × 1.5 + priorityFee
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
@@ -91,8 +91,12 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
         if (!chain.isOpStackL2) return BigInteger.ZERO
         if (fees !is Eip1559) return BigInteger.ZERO
 
+        // Resolve the call context outside the try so a genuine programming error (unsupported
+        // transaction type) surfaces instead of being masked as a zero L1 fee. Only the best-effort
+        // oracle call below degrades to zero on failure.
+        val context = resolveL1CallContext(transaction)
+
         return try {
-            val context = resolveL1CallContext(transaction)
             evmApi.getOpStackL1Fee(
                 senderAddress = transaction.coin.address,
                 to = context.to,
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
@@ -10,11 +10,14 @@ import com.vultisig.wallet.data.blockchain.model.GasFees
 import com.vultisig.wallet.data.blockchain.model.Swap
 import com.vultisig.wallet.data.blockchain.model.Transfer
 import com.vultisig.wallet.data.models.Chain
-import com.vultisig.wallet.data.models.isLayer2
+import com.vultisig.wallet.data.models.isOpStackL2
+import com.vultisig.wallet.data.models.oneInchChainId
 import com.vultisig.wallet.data.models.supportsLegacyGas
+import com.vultisig.wallet.data.utils.Numeric
 import com.vultisig.wallet.data.utils.increaseByPercent
 import java.math.BigInteger
 import javax.inject.Inject
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import timber.log.Timber
@@ -40,12 +43,7 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
                 calculateEip1559Fees(limit, chain, false, evmApi)
             }
 
-        val l1Fees =
-            if (chain.isLayer2) {
-                calculateLayer1Fees()
-            } else {
-                BigInteger.ZERO
-            }
+        val l1Fees = calculateLayer1Fees(transaction, fees, evmApi)
 
         return fees.addL1Amount(l1Fees)
     }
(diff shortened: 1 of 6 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
```diff
@@ -79,10 +77,86 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
...
+                maxFeePerGas = fees.maxFeePerGas,
+                maxPriorityFeePerGas = fees.maxPriorityFeePerGas,
+                chainId = chain.oneInchChainId().toBigInteger(),
+            )
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.w(e, "L1 data fee estimation failed for %s; treating as zero", chain)
+            BigInteger.ZERO
+        }
```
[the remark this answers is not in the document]
aminsato: Good catch — fixed in 7bfe586. Moved `resolveL1CallContext(transaction)` above the `try` so its `error()` for an unsupported transaction type now propagates as a real programming error instead of being masked as a zero L1 fee. The `try`/`catch` now wraps only the best-effort `getOpStackL1Fee` oracle call, which still degrades to zero on RPC failure.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
```diff
@@ -79,10 +77,86 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
...
+    }
+
+    // ERC-20 transfer(address,uint256) calldata, built with plain hex concatenation (no JNI) so
+    // this
+    // class stays unit-testable. Matches EvmApi.constructERC20TransferData.
+    private fun erc20TransferCallData(recipient: String, amount: BigInteger): ByteArray {
+        val methodId = "a9059cbb"
+        val paddedAddress = recipient.removePrefix("0x").padStart(64, '0')
+        val paddedValue = amount.toString(16).padStart(64, '0')
+        return Numeric.hexStringToByteArray(methodId + paddedAddress + paddedValue)
```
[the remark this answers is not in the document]
aminsato: Skipping this one — the stated rationale does not hold. `stripHexPrefix()` (StringExtensions.kt:71) is `if (startsWith("0x")) substring(2) else this`, which is case-sensitive, exactly like `removePrefix("0x")`. So for a `"0X"`-prefixed string both leave it untouched and produce identical calldata — there is no divergence or mismatch. The recipient address also comes from already-normalized internal sources. Using the shared helper would only be cosmetic, not a correctness fix, so leaving as-is.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
```diff
@@ -308,11 +308,25 @@ val Chain.isLayer2: Boolean
...
 
+// OP-stack–derived L2s that expose the GasPriceOracle predeploy (0x420…0F) and charge a separate
+// L1 data fee on top of L2 execution gas. Arbitrum and ZkSync are L2s too (isLayer2) but price
+// their L1 component differently, so they are deliberately excluded from the OP-stack oracle path.
+val Chain.isOpStackL2: Boolean
+    get() =
+        when (this) {
+            Chain.Base,
+            Chain.Blast,
+            Chain.Mantle,
```
rkokhatskyi: Routing `Mantle` through the oracle here adds a fee denominated in ETH to an MNT amount, since Mantle's `getL1Fee` is `l1GasUsed * l1BaseFee() * scalar()` with no tokenRatio while `fees.amount` is MNT wei, so the Mantle L1 cost is understated by the ETH/MNT tokenRatio

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
```diff
@@ -79,10 +77,86 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
...
+                        to = coin.contractAddress,
+                        value = BigInteger.ZERO,
+                        data = erc20TransferCallData(transaction.to, transaction.amount),
+                    )
+                }
+            is Swap ->
+                L1CallContext(
+                    to = transaction.to,
+                    value = if (coin.isNativeToken) transaction.amount else BigInteger.ZERO,
+                    data = Numeric.hexStringToByteArray(transaction.callData),
```
rkokhatskyi: Every live swap caller builds `Swap(callData = "", to = sender)` (`SwapGasCalculator.kt:90`, `JoinSwapUiModelBuilder.kt:116`), so this branch prices an empty calldata tx and the dominant swap router calldata L1 cost is never added

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
aminsato: Good catch — confirmed. Mantle is OP-stack–derived but its `getL1Fee` is ETH-denominated while the L2 gas (and our `fees.amount`) is MNT, so summing them understates the L1 cost without the chain's ETH/MNT `tokenRatio`. Rather than thread a tokenRatio conversion through, I removed `Chain.Mantle` from `isOpStackL2`, so Mantle no longer routes through the oracle and its L1 component is back to zero (its pre-PR behaviour) instead of being mis-summed against an MNT amount. Updated the doc comment to spell out the exclusion.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
aminsato: Confirmed — both live callers (`SwapGasCalculator.calculateGasFee` and `JoinSwapUiModelBuilder`) build the `Swap` with `callData = ""` and `to = sender`, and the real router calldata genuinely isn't available at estimation time: gas calc runs in parallel with the quote fetch in `SwapQuotePipelineController`, so the quote (which holds the router `tx.data`) only resolves later. Threading it in would mean restructuring that parallel pipeline.

Instead I now price the swap L1 cost against a representative ~512-byte swap-calldata payload (filled with non-zero bytes, so it's charged at the conservative 16-gas/byte rate) so the dominant L1 component is no longer dropped to zero. If a caller ever does supply real calldata, it's honoured. I also replaced the old Mantle-swap test that pinned the unreal `callData="0xabcdef"/to="0xRouter"` shape with one that asserts an OP-stack swap with empty calldata prices L1 against the representative payload.

rkokhatskyi (approved):
LGTM

PR #4917, opened 2026-06-15
fix(address-book): match EVM addresses case-insensitively

### LINKED ISSUE #4916: [Fix] EVM address book lookups are case-sensitive, so checksummed recipients miss saved entries [Address Book]
# [Fix] EVM address book lookups are case-sensitive, so checksummed recipients miss saved entries [Address Book]

## Problem
The address book stores and looks up addresses with exact, case-sensitive matching (`WHERE chainId = :chainId AND address = :address`). EVM addresses use EIP-55 mixed-case checksums, so the same account written with different casing — a checksummed address copied from an explorer (`0xAbC…`) versus the same address typed/pasted in lowercase (`0xabc…`) — is treated as a different entry. As a result a saved recipient is not recognized on the Verify/Sign screen, and the same recipient can be saved twice. Anyone who uses the address book on an EVM chain is affected.

## Expected Behavior
For EVM chains, an address saved under one casing should be found regardless of the casing used at lookup time (existence check, fetch, delete). This matches iOS and Windows, which compare EVM addresses case-insensitively. Non-EVM chains (Bitcoin, Solana, Cosmos, TON, etc.), where casing is significant, must keep exact matching.

## Steps to Reproduce
1. In the address book, save an Ethereum recipient using its checksummed form (e.g. `0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B`).
2. Start a send to the same recipient, pasted in lowercase (`0xab5801a7d398351b8be11c439e05c5b3259aec9b`).
3. On the Verify screen the saved title is not shown and "Save to address book" is offered, as though the recipient were unknown.

## Solution
Make the address-book existence/fetch/delete comparisons case-insensitive for EVM chains only, scoped by the chain's token standard, while leaving non-EVM chains' exact matching untouched. This mirrors iOS/Windows, which lowercase EVM addresses for comparison while preserving the stored value. The stored address is left exactly as the user entered it (so the displayed casing is unchanged), and no database migration or schema-version bump is required.

### Must Do
- [ ] Match `entryExists`, `getEntry`, and `delete` case-insensitively for EVM chains.
- [ ] Keep exact, case-sensitive matching for non-EVM chains.
- [ ] Add a repository unit test covering both the EVM (case-insensitive) and non-EVM (exact) paths.

### Must NOT Do
- Don't normalize or rewrite stored addresses — preserve the user's entered casing for display.
- Don't apply case-insensitive matching to non-EVM chains, where casing is significant.
- Don't add a database migration or bump the Room schema version.
- Don't change unrelated code or refactor surrounding logic.

## Acceptance Criteria
- [ ] `./gradlew :data:testDebugUnitTest` succeeds.
- [ ] New test: an EVM entry saved with mixed-case casing is found and deleted when queried with its lowercased form.
- [ ] New test: a non-EVM entry is NOT matched when queried with a different casing.

rkokhatskyi:
## Description

Fixes #4916

EVM addresses use EIP-55 checksum casing, so the same recipient saved under one capitalization (e.g. a checksummed address copied from an explorer) was treated as a different entry when looked up under another (e.g. the same address pasted in lowercase). The saved contact's title was not shown on the Verify/Sign screen and "Save to address book" was offered as if the recipient were unknown, and the same account could be saved twice.

This matches the address-book existence/fetch/delete by case for EVM chains only — scoped by the chain's token standard — using `COLLATE NOCASE` query variants, while leaving non-EVM chains (Bitcoin, Solana, Cosmos, TON, …), whose address encodings are case-sensitive, on exact matching. The stored value is left exactly as the user entered it, so the displayed casing is unchanged and no database migration or schema-version bump is required. This brings Android in line with iOS and Windows, which both compare EVM addresses case-insensitively while storing them verbatim.

## Which feature is affected?

## Checklist

N/A — no UI surface changes. The fix is in the data layer (address-book repository/DAO); the only user-visible effect is that an already-saved EVM recipient's name now resolves regardless of address casing, which is covered by unit tests.

## Additional context

The address-book lookups feed only the recipient display label and the "Save to address book" hint on the Verify/Keysign screens; they do not influence the signing payload, session, or broadcast, so co-signing is unaffected. The neighbouring vault-name match on those same screens already lowercases EVM addresses via `normalizeAddressForLookup`; this change makes the address-book lookup consistent with it.

## Summary by CodeRabbit

* **New Features**
  * Address book now treats EVM chain addresses as case-insensitive for lookups and deletion, accommodating EIP-55 checksum variations.
  * Non-EVM chains continue using exact case-sensitive matching.

* **Tests**
  * Added comprehensive test suite validating case-insensitive address matching for EVM chains and case-sensitive matching for other chains.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/db/dao/AddressBookEntryDao.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/db/dao/AddressBookEntryDao.kt
@@ -14,13 +14,30 @@ interface AddressBookEntryDao {
     @Query("SELECT * FROM address_book_entry WHERE chainId = :chainId AND address = :address")
     suspend fun getEntry(chainId: String, address: String): AddressBookEntryEntity
 
+    // COLLATE NOCASE folds ASCII case, which is exactly what EIP-55 checksum casing varies, so an
+    // EVM entry is matched regardless of how the address was capitalized when saved or looked up.
+    @Query(
+        "SELECT * FROM address_book_entry WHERE chainId = :chainId AND address = :address COLLATE NOCASE"
+    )
+    suspend fun getEntryIgnoringCase(chainId: String, address: String): AddressBookEntryEntity
+
     @Upsert suspend fun upsert(entry: AddressBookEntryEntity)
 
     @Query("DELETE FROM address_book_entry WHERE chainId = :chainId AND address = :address")
     suspend fun delete(chainId: String, address: String)
 
     @Query(
+        "DELETE FROM address_book_entry WHERE chainId = :chainId AND address = :address COLLATE NOCASE"
+    )
+    suspend fun deleteIgnoringCase(chainId: String, address: String)
+
+    @Query(
         "SELECT EXISTS(SELECT 1 FROM address_book_entry WHERE chainId = :chainId AND address = :address)"
     )
     suspend fun entryExists(chainId: String, address: String): Boolean
+
+    @Query(
+        "SELECT EXISTS(SELECT 1 FROM address_book_entry WHERE chainId = :chainId AND address = :address COLLATE NOCASE)"
+    )
+    suspend fun entryExistsIgnoringCase(chainId: String, address: String): Boolean
 }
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #4733, opened 2026-06-04
feat(ton): decode jetton transfers on dApp keysign

rkokhatskyi:
## Summary

TonConnect signing requests headlined the forwarded gas (e.g. 0.32 TON) instead of the real jetton amount, showed an opaque transfer to the sender's jetton wallet, and hid the BOC payload behind a badge. This decodes the message body so the keysign verify and done screens show the real operation.

- Minimal BOC reader + body decoder (jetton transfer, NFT transfer, excess-gas refund) under `data/crypto/ton`. It returns raw `workchain:hex` addresses so it stays JVM unit-testable; the UI formats them via `TONAddressConverter`.
- Jetton symbol resolution against the vault: the transfer's destination wallet is mapped to its master through the TON indexer, then matched to a held coin. On Android the master lives in `Coin.contractAddress` (the coin's `address` is the owner wallet), so the lookup is needed — unlike iOS, which stores the wallet in `Coin.address`.
- The resolved amount/ticker/logo becomes the keysign hero; each message renders its operation label, real recipient, forwarded TON amount, and a copyable raw payload.

Every Android device in a TonConnect co-sign joins the keysign, so the hero is wired once in `JoinKeysignViewModel` and carried to both the verify and done screens. Blockaid doesn't cover TON, so this replaces the (empty) Blockaid hero on the TON path.

Scope is the transfer opcodes; swap simulation and on-chain metadata for jettons the vault doesn't hold are left out.

## Before / After

| Before | After |
|--------|-------|
| before | after |

Both are the full keysign verify screen with the messages section expanded, same vault/addresses/fee. Before headlines `0.32 TON` with `#1 Transfer` to the sender's jetton wallet; after headlines `100 USDT` with `#1 Jetton Transfer` to the real recipient, a `Forward TON Amount` row, and the copyable raw payload.

## Testing

- `./gradlew :data:testDebugUnitTest` — BOC decoder against canonical @ton/core fixtures (jetton/NFT/excesses, hex BOC, 0x00000000 text-comment prefix, truncation/unknown-opcode/no-ref error cases) + the jetton-master DTO helper.
- `./gradlew :app:testDebugUnitTest` — message-to-row mapping and vault jetton resolution (hit, not-in-vault, unresolved master, case-sensitive match, excesses skip, negative-amount guard).
- Before/after captured on the emulator via PreviewActivity.

Closes #4057

## Summary by CodeRabbit

* **New Features**
[...]

PR #4923, opened 2026-06-15
feat(ton): decode and display TON DEX swaps on dApp keysign (#4737)

### LINKED ISSUE #4737: feat(ton): decode and display TON DEX swaps on dApp keysign
## Problem

PR #4733 (issue #4057, "Phase 1C") shipped the local BOC decoder, jetton/NFT/excess transfer display, and a vault-held jetton HERO on the TonConnect keysign verify/done screens. It intentionally did **not** implement DEX swap display.

A TON swap on STON.fi or DeDust arrives over the wire as a **jetton transfer to the DEX router** (the swap parameters live inside the transfer's forward payload). Because the decoder only classifies the outer jetton-transfer op, a swap currently renders as a plain **"Jetton Transfer" to the router address** — the user sees a transfer to an opaque contract instead of a "You're swapping X → Y" summary.

Concretely: swapping 100 USDT → TON on STON.fi v2 today shows up as `Jetton Transfer · to: EQ…<router>` with no indication that it is a swap, what the output token is, or the minimum received.

A second, related gap: metadata (symbol/decimals/logo) is resolved **only for vault-tracked tokens**. That is sufficient for plain transfers (you transfer tokens you hold, and those are usually auto-discovered into `vault.coins`), but the swap **output token** is something the user does not hold yet and is not in `vault.coins`, so it needs on-chain/registry metadata to render its ticker and icon. These two are really one follow-up: **swap display, with output-token metadata resolution as a sub-part.**

## Co-sign topology (Android is joiner-only)

Android has **no TonConnect/dApp connection layer** — `signTon` requests only ever arrive via the keysign **JOIN** flow (`Route.Keysign.Join`). So for any TON dApp swap:

- **Android is always the joining co-signer**, never the device holding the dApp connection. This work is purely the **joiner-side display** on Android's verify/done screens — no connection/origination, session, or bridge code is needed (or wanted) here.
- The **initiator** (the companion that holds the TonConnect connection and starts the keysign) is the **browser extension** (the primary/reference dApp surface) or **iOS** (which also supports TonConnect co-sign — vultisig/vultisig-ios#4138). Both `extension-initiator + Android-joiner` and `iOS-initiator + Android-joiner` are valid pairings.
- **Out of scope:** in-app TON swaps started from Android's own swap UI go through the native transfer path (no `signTon`), so they are a different code path and not covered here.

This mirrors the topology of the jetton-transfer display already shipped in #4733.

## Scope

Port the swap-decode/emulation/display half of Phase 1C that Windows shipped bundled. Components to add (file references point at the iOS source of truth):

- [ ] **`TonOpcode` swap constants** — add the swap body opcodes the decoder switches on: `stonfiV2Swap 0x6664de2a`, `ptonTransfer 0x01f3835d`, `dedustNativeSwap 0xea06185d`, `dedustJettonSwap 0xe3a0d482` (alongside the already-ported `jettonTransfer 0x0f8a7ea5`, `nftTransfer 0x5fcc3d14`, `excesses 0xd53276db`). Sourced from STON.fi v2 / DeDust docs.
- [ ] **`TonMessageBodyIntent.Swap` model** — extend the decoded-body sealed model with a `Swap(provider stonfi|dedust, offerAsset ton|jetton, offerAmount, minOut?, receiverAddress?, refundAddress?, excessesAddress?, targetAddress?)` case. Amounts as decimal strings / `BigInteger` (jetton supplies exceed `Long`).
- [ ] **Local swap classification in `TonMessageBodyDecoder`** — parse the STON.fi v2 swap inside a jetton-transfer / pTON forward payload and the DeDust native-in swap; read **every** swap field so a valid-prefix/garbage-tail body fails closed (returns null) rather than rendering a fake swap. (Builds on the cell/slice + BOC parser introduced by the jetton-decode work.)
- [ ] **`TonKnownRouters` allowlist (anti-spoofing gate)** — hardcoded `Set` of STON.fi v2 router addresses, STON.fi v2 pTON wallet addresses, and the DeDust mainnet native-TON vault, normalized to user-friendly bounceable form. Swap classification is gated on it: `ptonTransfer` only when the outer destination is a known pTON wallet; `dedustNativeSwap` only when the outer destination is the native vault; STON.fi v2 jetton-swap only when the inner jetton-transfer destination is a known router. Without this gate any contract could spoof a swap card.
- [ ] **Remote `tonapi.io` emulation fallback** — for swaps the local decoder cannot statically classify (notably **DeDust jetton swaps** — DeDust uses one vault per jetton, so the vault set is not statically enumerable and `dedustJettonSwap 0xe3a0d482` is deliberately not locally classified). Build an unsigned wallet-v4R2 external-message BOC (reuse `TonHelper.getPreSignedInputData(payload)` + a 64-byte zero placeholder signature via `TransactionCompiler.compileWithSignatures`), then `POST https://tonapi.io/v2/events/emulate?ignore_signature_check=true` and map the first `status==ok && type==JettonSwap` action into a swap result. Local-decode-first, remote-emulation-fallback.
- [ ] **Output-token metadata resolution** — resolve the swap output token (not held by the user, not in `vault.coins`) via registry + on-chain `/jetton/masters` lookup for ticker/decimals/logo. From emulation, take it straight from TonAPI's `jetton_master_out` preview. Handle TON→jetton, jetton→TON, jetton→jetton directions; tolerate TonAPI numeric fields being int64-or-string; TON native fallback ticker `TON` / decimals `9`.
- [ ] **Sign-display ViewModel/state** — orchestrate: decode each `TonMessage` body locally → entry(message, intent, jettonCoin, swapOutputCoin); resolve metadata per entry; if no local swap and a `keysignPayload` exists, run the emulation simulator. Visibility rules: when emulation surfaces a swap, suppress all per-message panels (the swap card is the whole summary); when a local swap is the hero, hide same-self sidecar messages ≤ 0.01 TON (10_000_000 nanoton). Cache on `(vaultPubKeyECDSA, fromAddress, messages)` to avoid re-running. (`@HiltViewModel` / `StateFlow` feeding the Compose screen.)
- [ ] **`SignTonDisplayView` rewrite (per-message swap card + hero)** — `SignTonDisplayView.kt` today only renders raw to/amount per message. Add a **"You're swapping" hero swap card** (from/to coin icon + formatted amount + ticker, down-arrow divider) for both local-decoded and emulation swaps; per-message labeled panels (jettonTransfer / nftTransfer / excessGasRefund / transfer) with typed rows; a collapsible raw-payload disclosure; and a loading spinner while simulating. The call site must pass `keysignPayload + vault + fromAddress` (today only `signTon` is passed). New `youreSwapping` / `transactionDetails` / `rawPayload` / operation-title strings must be added to all 10 locale files.

## Out of scope

Everything already shipped in #4057 / PR #4733: the local BOC cell/slice parser, plain jetton-transfer / NFT-transfer / excess-gas-refund decoding and display, and the vault-held jetton HERO. This follow-up only adds the swap classification, emulation fallback, output-token metadata, and the swap-card UI on top of that foundation.

## Cross-platform parity

The Android swap-display work mirrors the swap/emulation scope that iOS and Windows shipped bundled into their Phase 1C tickets (neither platform split swap into a separate issue):

- **iOS** — issue [#4148](https://github.com/vultisig/vultisig-ios/issues/4148) ("feat: TON BOC decoding & jetton transfer display (Phase 1C)"); implemented by merged PR [#4308](https://github.com/vultisig/vultisig-ios/pull/4308) ("feat(ton): TonConnect verify display + jetton/swap simulation (Windows parity)"), which folded in and superseded the standalone decoder PR [#4304](https://github.com/vultisig/vultisig-ios/pull/4304). Hero/sentinel foundation: PR [#4145](https://github.com/vultisig/vultisig-ios/pull/4145).
- **Windows** — issue [#3737](https://github.com/vultisig/vultisig-windows/issues/3737) ("feat: TON BOC decoding & jetton transfer display (Phase 1C)"); implemented by merged PR [#3817](https://github.com/vultisig/vultisig-windows/pull/3817) ("feat(ton): decode BOC, display jetton transfers, collapse raw memo"), which shipped the DEX swap decode + remote `tonapi.io` emulation + output-token metadata in the same bundle.

Note: there is no separate dedicated "swap-display" issue on either platform — the swap work rolls up to the Phase 1C tickets above.

## Design decisions (recommended)

Because Android is the **joiner** doing **display-only independent verification** (see *Co-sign topology*), one rule governs all of these: derive what you show from the raw `signTon` you're about to co-sign, and don't phone a third party more than necessary to do it.

**1. Trust boundary — read the bytes, never the initiator. (Hard requirement.)**
Build the swap hero/card from the joiner's own decode of `signTon`; never from `dappMetadata` or any initiator-supplied label. A compromised initiator could otherwise show a benign swap while the signed bytes do something else. Keep classification pure-local and offline — this is the entire point of local decode + the router-allowlist gate.

**2. Emulation — ship local-decode-only; do NOT call public tonapi.io from a co-signer.**
In order of preference:
1. **Local-decode-only (recommended default).** The allowlist-gated decoder already classifies the common swaps — STON.fi v2 (jetton + pTON) and DeDust native — with zero network and zero leak. Ship this and stop. Swaps it can't statically classify (DeDust *jetton* swaps, whose vaults aren't enumerable) degrade to a plain jetton-transfer card, which is safe and honest — it literally is a jetton transfer to that contract.
2. **Only if a swap card for that long tail is judged worth it: route `/v2/events/emulate` through the `api.vultisig.com` proxy** (small backend passthrough — the proxy currently fronts toncenter, not tonapi.io). That keeps the iOS/Windows display while hiding the co-signer's IP from tonapi.io and enabling caching/rate-limiting.
3. **Reject the iOS/Windows approach of calling public `https://tonapi.io/v2/events/emulate` directly.** On a co-signer it exposes the unsigned-but-real tx (wallet address + destinations + amounts) **and the user's IP** to a third party, per device, before the tx is even signed — a privacy regression for a display-only label, and a weak basis for a security screen.

This diverges from iOS/Windows, so ratify it with the team — but for a privacy-first co-signer, local-only (or proxied) is the right default, not direct tonapi.io.

**3. Router allowlist — hardcode it, guard it with a unit test (not a runtime assert).**
Ship `TonKnownRouters` as a checked-in literal `Set` mirrored from the SDK `knownRouters`, with a comment naming the source-of-truth + last-synced date. Don't build a sync pipeline: drift only ever fails *safe* (a real swap degrades to a plain transfer; an attacker still can't earn a swap card), so manual re-sync after a STON.fi/DeDust router redeploy is acceptable. Replace iOS's DEBUG-only assert with a **CI unit test** asserting every entry parses as a valid bounceable address and the set is non-empty — stronger than a runtime assert.

**4. Address normalization — canonicalize both sides through WalletCore, prove it with a test.**
Normalize the allowlist entries (once, at load) **and** every candidate address through `TONAddressConverter.toUserFriendly(addr, bounceable = true, testnet = false)` before any `Set.contains` / equality check, so comparison is always canonical-vs-canonical (a raw `workchain:hex` and its `EQ…` form must compare equal). Keep this in the runtime/VM layer — the pure decoder stays JNI-free and returns raw addresses, exactly as #4733 established. Add a unit test feeding a router in both raw and friendly form and asserting the gate matches both — that closes the silent-gate-failure risk rather than leaving it open.

## Reference

Follow-up to #4057 / PR #4733 (Phase 1C jetton-transfer decoder). Builds directly on the BOC cell/slice parser and display scaffolding landed there.

aminsato:
Closes #4737. Follow-up to the merged Phase-1C jetton decoder (#4733). Decodes TON DEX swaps on the joiner-side keysign verify/done screens so a swap no longer renders as an opaque "Transfer to <router>".

Android is **joiner-only** — `signTon` arrives via the keysign JOIN flow; the initiator is the browser extension or iOS. This is pure joiner-side display: no connection/session/bridge code.

## Verified on device (DeDust TON → jetton)

| Before | After |
|--------|-------|
| before | after |

Before: the swap decoded as a plain **"Transfer"** with a single-sided "You're sending" hero. After: a **"You're swapping TON → JETTON"** hero + a **"Swap"**-labelled message row, on a real on-chain DeDust swap.

## What it does
- **Decoder (data, JNI-free):** swap opcodes (pTON `0x01f3835d`, STON.fi v2 `0x6664de2a`, DeDust native `0xea06185d`), a `Swap` intent, and a forward-payload cursor. Fail-closed: reads every additional-data field (garbage tail → null), referral as Maybe-address, DeDust `given_out` throws, DeDust `SwapParams` read from its `^ref`. DeDust-jetton `0xe3a0d482` is defined but never dispatched.
- **Router allow-list:** `TonKnownRouters` (STON.fi v2 routers + pTON wallets + DeDust native vault), mirrored from the iOS/SDK source-of-truth. iOS's DEBUG assert is replaced with a **CI test** validating every entry as a real bounceable address (base64url → 36 bytes → CRC16).
- **Runtime gate:** normalizes both the allow-list and the candidate through WalletCore so raw/URL-safe forms compare equal, then gates. Ungated/unresolvable → no swap card (falls back to per-message display). Built only from the joiner's own decode, never from initiator-supplied dApp metadata.
- **Output-token metadata:** `getJettonMetadata` (`/v3/jetton/masters`) for STON.fi (target = output jetton wallet); for DeDust the target is the **pool**, so `getDedustPoolOutputMaster` reads the pool's `get_assets` and resolves the non-native asset. Resolution order: vault → `Coins` registry → on-chain. All via the existing api.vultisig.com proxy.
- **UI:** reuses the existing `HeroContent.Swap` "You're swapping X → Y" hero (no bespoke card), taking precedence over the single-sided jetton hero and carrying to the done screen. Per-message list gains a "Swap" row and hides same-self gas sidecars (≤ 0.01 TON). `ton_op_swap` added across all 10 locales.

## ⚠️ Design decision — local-decode only (needs team ratification)
iOS (#4308) and Windows (#3817) call **public `tonapi.io/v2/events/emulate` directly from the co-signer**. This PR **deliberately diverges** to local-decode only (no emulation, no per-device IP/tx leak), per the issue's design section.

**Coverage trade-off, surfaced by on-device testing:** local-decode covers what is statically gateable — **STON.fi v2** and **DeDust native** (both verified). It does **not** cover:
- **STON.fi v1** (`0x25938561`) — un-gateable by a static set (per-jetton router wallets), so it degrades to a plain jetton-transfer card.
- **Omniston / cross-DEX aggregation** — different settlement op, not classifiable locally.

In practice STON.fi/DeDust frequently route through v1/Omniston, so a swap card will not always appear. Closing that gap needs either a network owner-lookup (v1) or proxied `/events/emulate` (everything). Flagging for the team to decide between shipping this honest local-only baseline vs. adding proxied emulation.

## Test plan
- `:data`: STON.fi v2 swap classification (real BOC fixture) + fail-closed cases; `TonKnownRouters` CRC/validity. ✅
- `:app`: gate matches a router in **both raw and friendly** form, rejects an unlisted router, **DeDust native swap resolves from the pool** (real on-chain fixture), suppresses the gas sidecar. ✅
- `:app:assembleDebug`, ktfmt. ✅
- **Device E2E:** real DeDust TON→jetton swap, joined on Android — swap hero + "Swap" row shown (screenshots above). ✅

## Out of scope
Remote tonapi.io emulation (see design decision), in-app TON swaps from Android's own swap UI (native path, no `signTon`), and the Toncoin→Gram native-ticker rebrand (name-only, separate).

## Cross-platform parity
iOS #4308, Windows #3817 (both merged). Parity on opcodes / parse-order / router-gating / fail-closed; **divergence on emulation** as described.

## Summary by CodeRabbit

* **New Features**
  * Added support for TON Swap operations in transaction previews and multi-message signing, including a dedicated Swap label and swap-aware from/to amount presentation.
  * Introduced gated “hero” rendering for trusted DEX swaps (with improved token details via TON jetton metadata).
  * Added multi-language UI text for the Swap operation.

* **Bug Fixes**
  * Improved swap detection/decoding across additional TON opcodes and ensured consistent hero content across the verify/done flow.
  * Suppressed small self-addressed gas “sidecar” rows during trusted swaps and safely fell back to standard transfer rendering for untrusted swaps.

* **Tests**
  * Expanded coverage for swap decoding, hero resolution, gas-sidecar filtering, and router allow-list validation.


app/src/main/java/com/vultisig/wallet/ui/models/keysign/TonMessageDecode.kt
```diff
@@ -12,20 +12,46 @@ import vultisig.keysign.v1.TonMessage
...
+        if (hasSwap && isSwapGasSidecar(message, intent, fromAddress, formatAddress)) {
+            return@mapNotNull null
+        }
         val rawPayload = message.payload?.takeIf { it.isNotEmpty() }
         val hasStateInit = !message.stateInit.isNullOrEmpty()
-        when (val intent = TonMessageBodyDecoder.decode(message.payload)) {
+        when (intent) {
+            is TonMessageBodyIntent.Swap ->
+                TonMessageUiModel(
+                    operation = TonMessageOperation.Swap,
```
rkokhatskyi: `mapTonMessages` classifies a row as `TonMessageOperation.Swap` straight from the ungated `TonMessageBodyDecoder.decode(it.payload)` with no `TonKnownRouters` gate, while the hero gates it in `resolveTonSwapHero`, so a swap-shaped body to any contract earns a trusted Swap label on the signing screen and hides its gas sidecar

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/TonMessageDecode.kt
aminsato: Good catch — fixed in aea457475. Extracted the allow-list gate into a shared `TonSwapGate` so the per-message list and the hero trust exactly the same set of swaps (kept them as one copy since `TonKnownRouters` warns the runtime gate must not drift). `mapTonMessages` now demotes an ungated swap to a plain transfer, so a swap-shaped body to an unknown contract no longer earns a Swap label or hides its self-addressed gas sidecar. Added a regression test (`demotes an ungated swap to a plain transfer and keeps its gas sidecar visible`) and fixed the existing sidecar test, which was passing an identity normalizer yet still asserting a Swap label.

rkokhatskyi (approved):
LGTM

PR #4926, opened 2026-06-15
fix(send): reject amounts with more decimals than the token supports

### LINKED ISSUE #4925: [Fix] Send silently truncates amounts with more decimals than the token supports [Send]
# [Fix] Send silently truncates amounts with more decimals than the token supports [Send]

## Problem
On the Send screen, when a user enters an amount with more fractional digits than the selected token's precision (e.g. `1.1234567` for a 6-decimal token), the extra digits are silently dropped when the amount is converted to base units for signing. `BigDecimal.movePointRight(decimals).toBigInteger()` truncates toward zero, so the amount that gets signed and broadcast differs from what the user typed, with no warning. It affects anyone who types (or pastes) more precision than the asset can represent.

## Expected Behavior
The signed amount should always match what the user entered. If the amount has more decimal places than the token supports, the app should reject it with a clear message instead of silently altering the value.

## Steps to Reproduce
1. Open Send and select a token with few decimals (e.g. a 6-decimal token such as USDC).
2. Enter an amount with more fractional digits than the token supports, e.g. `1.1234567`.
3. Tap Continue.
4. The flow proceeds with `1.123456` — the 7th digit is dropped silently.

## Solution
Validate the entered amount's fractional-digit count against the selected token's decimals before converting to base units. Reject over-precision (with a clear, localized message) in the shared submit validator so every send strategy is covered, and mirror the same check in the inline amount validation so the user gets early feedback on the field. Trailing zeros do not count as real precision, so `1.5000000` on a 6-decimal token is still accepted. This guarantees the signed amount equals the displayed amount.

### Must Do
- [ ] Reject send amounts whose fractional digits exceed the selected token's decimals in the shared submit validation path (covers plain sends and the send-form DeFi actions that route through it).
- [ ] Surface the same rejection inline on the amount field for early feedback.
- [ ] Add a clear, localized error string across all maintained locale files.
- [ ] Add unit tests for the boundary: at-precision passes, over-precision fails, trailing zeros are ignored.

### Must NOT Do
- Don't change the separate Deposit screen amount validator/flow — that is a distinct surface.
- Don't add an input mask or otherwise change how digits are typed into the amount field.
- Don't alter the existing precision clamping used by fiat↔token conversion or the percentage/MAX buttons (they already scale down to the token's precision).
- Don't refactor the send strategies or unrelated validation.

## Acceptance Criteria
- [ ] `./gradlew :app:testDebugUnitTest` succeeds.
- [ ] `AmountManager.validateTokenAmount` returns an error `UiText` for a 6-decimal token with input `1.1234567`.
- [ ] `AccountValidator.validate()` throws for an over-precision amount and returns normally at exactly the token's precision.
- [ ] An amount with trailing zeros beyond the token's precision (e.g. `1.5000000`) is accepted.

rkokhatskyi:
## Description

On the Send screen, entering an amount with more fractional digits than the selected token's precision (e.g. `1.1234567` for a 6-decimal token) silently dropped the surplus when the amount was converted to base units for signing — `movePointRight(decimals).toBigInteger()` truncates toward zero — so the signed/broadcast amount differed from what the user typed, with no warning.

This validates the entered amount's fractional-digit count against the token's decimals and rejects over-precision with a clear, localized message instead of silently truncating:

- `AccountValidator.validate()` — the shared pre-submit check that every send strategy (plain send and the send-form DeFi actions) runs before building the payload, so the truncated amount can never reach signing or the co-signer.
- `AmountManager.validateTokenAmount()` — mirrors the check inline on the amount field for early feedback, alongside the existing positivity/length checks.

Trailing zeros are not treated as real precision, so `1.5000000` on a 6-decimal token is still accepted. The fiat↔token conversion and percentage/MAX buttons already clamp to the token's precision and are unaffected.

Fixes #4925

## Which feature is affected?

No tx link applies: the change only blocks an over-precision amount before any transaction is built or signed. Valid amounts (within the token's precision) are unchanged, so existing sends sign and broadcast exactly as before.

## Checklist

Not applicable — no layout or component change. The new message renders in the Send form's existing amount-error slot (the same one that already shows "Amount must be a positive number" / "Invalid amount"). The prior behavior had no visual manifestation: the amount was silently truncated with the form looking unchanged, so there is no before/after visual to compare.

## Additional context

Co-sign: the precision check runs in `AccountValidator.validate()` before `BlockChainSpecificRepository.getSpecific` and the `KeysignPayload` are built, so the VultiServer co-signer can never receive an over-precision-truncated `toAmount`. No co-sign behavior changes for valid amounts.

Scope is limited to the Send screen. The Deposit screen uses a separate validator (`DepositFieldValidator`) and flow; it is left untouched here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added decimal precision validation for token amounts during the send process. The app now prevents users from entering amounts that exceed the token's supported decimal precision and displays clear error messages.
  * Error messages are localized across German, Spanish, Croatian, Italian, Korean, Dutch, Portuguese, Russian, and Chinese.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountManager.kt
@@ -81,6 +81,13 @@ internal class AmountManager(
         if (decimal == null || decimal <= BigDecimal.ZERO) {
             return UiText.StringResource(R.string.send_error_no_amount)
         }
+        val tokenDecimals = selectedToken.value?.decimal
+        if (tokenDecimals != null && decimal.hasExcessDecimals(tokenDecimals)) {
+            return UiText.FormattedText(
+                R.string.send_error_too_many_decimals,
+                listOf(tokenDecimals),
+            )
+        }
         return null
     }
 
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/SendAmountExtensions.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/SendAmountExtensions.kt
@@ -1,0 +1,13 @@
+package com.vultisig.wallet.ui.models.send
+
+import java.math.BigDecimal
+
+/**
+ * True when this amount carries more fractional digits than the token's [decimals] precision.
+ *
+ * Converting to base units with `movePointRight(decimals).toBigInteger()` rounds toward zero, so
+ * any digit past [decimals] would be dropped silently. Trailing zeros are not real precision —
+ * `1.5000000` fits a 6-decimal token — so they are stripped before the scale is compared.
+ */
+internal fun BigDecimal.hasExcessDecimals(decimals: Int): Boolean =
+    stripTrailingZeros().scale() > decimals
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/submit/AccountValidator.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/submit/AccountValidator.kt
@@ -8,6 +8,7 @@ import com.vultisig.wallet.data.models.TokenValue
 import com.vultisig.wallet.data.models.allowZeroGas
 import com.vultisig.wallet.data.repositories.AddressParserRepository
 import com.vultisig.wallet.ui.models.send.InvalidTransactionDataException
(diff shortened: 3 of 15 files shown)


johnnyluo (approved):
LGTM

PR #4928, opened 2026-06-15
fix(tron): guard cached chain parameters against concurrent fee calculations

### LINKED ISSUE #4927: [Fix] TRON fee service caches chain parameters in an unsynchronized shared field [Send/TRON]
# [Fix] TRON fee service caches chain parameters in an unsynchronized shared field [Send/TRON]

## Problem
`TronFeeService` is provided as an application-wide `@Singleton` (`FeeServiceProvidersModule`), so a single instance is shared by every caller that resolves TRON fees — the send form, swap gas calculation, and the join-keysign fee resolver. It memoizes the result of `tronApi.getChainParameters()` in a plain `private var chainParameters` field via an unsynchronized check-then-set, and reads it back with a `!!` non-null assertion. When two fee calculations run at the same time (e.g. the send screen recomputes a fee while a portfolio/keysign fee resolve is in flight), they race on that field: both can observe `null` and each issue a redundant `getChainParameters()` network call, and the `!!` is an assertion on mutable state that other coroutines can write concurrently.

## Expected Behavior
Chain parameters are fetched at most once and shared safely across concurrent callers, with no non-null assertion on a mutable shared field.

## Steps to Reproduce
The race is timing/JMM-dependent and not reliably reproducible by hand. It is demonstrated by a unit test that launches several `calculateFees` calls concurrently against one service instance while the `getChainParameters()` call is suspended: with the current code each concurrent caller invokes `getChainParameters()`, instead of exactly one fetch being shared.

## Solution
Guard the cached field with the same double-checked locking pattern already used by `ZcashApi.getConsensusBranchIdHex()` in the same module: a `@Volatile` field with a lock-free fast-path read, and a `Mutex.withLock` re-check before fetching. The first caller fetches and caches; concurrent callers wait on the mutex, then return the cached value. This removes the data race and the `!!`. No change to fee math, fetched values, or the `TronApi` interface.

### Must Do
- [ ] Make the TRON chain-parameter cache thread-safe so concurrent `calculateFees` calls trigger a single `getChainParameters()` fetch
- [ ] Remove the `!!` non-null assertion on the mutable cache field
- [ ] Add a concurrency unit test asserting `getChainParameters()` is invoked exactly once across N parallel `calculateFees` calls

### Must NOT Do
- Don't change the fee calculation math or the values read from chain parameters
- Don't modify the `TronApi` interface or any other fee service
- Don't introduce a new caching abstraction — reuse the existing `ZcashApi` double-checked-lock pattern

## Acceptance Criteria
- [ ] New test launching N parallel `calculateFees` calls asserts `tronApi.getChainParameters()` is called exactly once (fails before the fix, passes after)
- [ ] `grep -n '!!' data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronFeeService.kt` returns nothing

rkokhatskyi:
## Description

`TronFeeService` is bound as an application-wide `@Singleton` and is shared by every caller that resolves TRON fees — the send form, swap gas calculation, and the join-keysign fee resolver. It memoised the result of `tronApi.getChainParameters()` in a plain `private var` field through an unsynchronised check-then-set, and read it back with a `!!` non-null assertion. When two fee calculations run concurrently they could race on that field: both observe `null`, each issues a redundant network fetch, and the `!!` asserts non-null on state another coroutine can write at the same time.

This guards the cached field with the same double-checked locking pattern already used by `ZcashApi.getConsensusBranchIdHex()` in this module — a `@Volatile` field with a lock-free fast-path read and a `Mutex.withLock` re-check before fetching. The first caller fetches and caches; concurrent callers wait on the mutex and then reuse the cached value, and the `!!` is gone. Fee math and the values read from chain parameters are unchanged.

Fixes #4927

## Which feature is affected?

_TRX/TRC20 fee estimation is part of the send and swap paths, but this change is internal concurrency correctness only — transaction construction and on-chain output are byte-identical, so there is no new tx link to attach._

## Checklist

No UI surface — this is a data-layer concurrency fix.

## Additional context

The change only affects how chain parameters are cached and synchronised; the fetched values and the fee they feed into are unchanged, so signing payload construction, server co-signing, and broadcast are unaffected — the fee estimate handed to the keysign flow is identical to before. A regression test launches 8 concurrent `calculateFees` calls against one instance and asserts `getChainParameters()` is invoked exactly once (it fails before the fix, which issues eight fetches); a second test confirms sequential calls still reuse the cache. `./gradlew :data:testDebugUnitTest` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced Tron fee calculation reliability with improved concurrency handling and optimized chain parameter caching.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronFeeService.kt
@@ -18,6 +18,8 @@ import javax.inject.Inject
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.supervisorScope
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import timber.log.Timber
 import wallet.core.jni.Base58
 
@@ -66,7 +68,8 @@ import wallet.core.jni.Base58
  */
 class TronFeeService @Inject constructor(private val tronApi: TronApi) : FeeService {
 
-    private var chainParameters: TronChainParametersJson? = null
+    private val chainParametersMutex = Mutex()
+    @Volatile private var chainParameters: TronChainParametersJson? = null
 
     override suspend fun calculateFees(transaction: BlockchainTransaction): TronFees =
         coroutineScope {
@@ -310,12 +313,14 @@ class TronFeeService @Inject constructor(private val tronApi: TronApi) : FeeServ
     }
 
     private suspend fun getCacheTronChainParameters(): TronChainParametersJson {
-        return if (chainParameters == null) {
-            val params = tronApi.getChainParameters()
-            chainParameters = params
-            params
-        } else {
-            chainParameters!!
+        chainParameters?.let {
+            return it
+        }
+        return chainParametersMutex.withLock {
+            chainParameters?.let {
+                return it
+            }
+            tronApi.getChainParameters().also { chainParameters = it }
         }
     }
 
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4936, opened 2026-06-16
fix(network): retry transport errors only for idempotent methods

### LINKED ISSUE #4932: [Fix] Shared HTTP client retries non-idempotent POSTs on transport timeout [networking]
# [Fix] Shared HTTP client retries non-idempotent POSTs on transport timeout [networking]

## Problem
The shared Ktor `HttpClient` installs `HttpRequestRetry` with `retryOnException(maxRetries = 3, retryOnTimeout = true)`, which retries **every** request — including non-idempotent `POST`s — on an `IOException` or read-timeout. The status-based predicate (`retryIf`) already restricts 5xx/429/408 retries to safe methods (GET/HEAD/OPTIONS), but the transport-exception/timeout retry is applied unconditionally. A `POST` that the server actually received but slow-ACKed (for example a transaction broadcast the node already accepted, or a relay message during keygen/keysign) can therefore be silently resent up to three times once the 15s read timeout fires.

## Expected Behavior
Transport-level retries (IOException / read-timeout) apply only to idempotent methods, consistent with the status-based policy that already gates on safe methods, and with the dedicated QBTC proof client which already disables implicit retry for its non-idempotent `/prove` POST. A non-idempotent request surfaces the failure once instead of being auto-resent; the per-chain broadcast recovery (`recoverIfAlreadyBroadcast` / known-hash mapping) handles a transaction that actually landed, and OkHttp's `retryOnConnectionFailure` still covers genuine pre-send connection failures.

## Steps to Reproduce
1. Start a Send and broadcast a transaction over a slow/flaky connection.
2. Have the broadcast `POST` reach the node but its response arrive slower than the 15s read timeout (the node has accepted the tx; the ACK is in flight).
3. `HttpRequestRetry` re-sends the same broadcast `POST` up to three times instead of surfacing the timeout once.

## Solution
Replace `retryOnException(maxRetries = 3, retryOnTimeout = true)` with a method-gated `retryOnExceptionIf` that only retries idempotent (GET/HEAD/OPTIONS) requests. Keep the `CancellationException` guard that the built-in predicate provides, because `HttpRequestRetry` hands the cause straight to the predicate without filtering cancellations, so a blanket retry would break structured concurrency. The status-based `retryIf` predicate, the exponential backoff, and OkHttp's connection-level retry are left untouched. This makes the exception-based and status-based predicates consistent (both gate on idempotency) and is a one-file production change.

### Must Do
- [ ] Gate transport-exception/timeout retries to idempotent methods (GET/HEAD/OPTIONS) via `retryOnExceptionIf`, reusing one `isSafeMethod` helper shared with the status-based `retryIf`.
- [ ] Preserve the existing cancellation semantics (never retry on `CancellationException`).
- [ ] Update the existing test that pins POST-being-retried-on-IOException, and add a read-timeout regression test pair (POST not resent; GET still retried three times).

### Must NOT Do
- Don't change the retry count (3) or the exponential backoff.
- Don't touch the per-chain broadcast or `recoverIfAlreadyBroadcast` recovery logic, the status-based `retryIf`, or the OkHttp engine config (`retryOnConnectionFailure`, timeouts).
- Don't add or alter retry behavior on any other `HttpClient` instance.

## Acceptance Criteria
- [ ] `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.networkutils.HttpClientConfiguratorTest"` passes.
- [ ] A `POST` that fails with an `IOException` / `SocketTimeoutException` is attempted exactly once (no retry) and surfaces a `NetworkException`.
- [ ] A `GET` that fails with an `IOException` / `SocketTimeoutException` is still retried three times before failing.

rkokhatskyi:
## Description

Fixes #4932

The shared `HttpClient`'s `HttpRequestRetry` retried every request on `IOException`/read-timeout via `retryOnException(maxRetries = 3, retryOnTimeout = true)`, including non-idempotent `POST`s. The status-based predicate (`retryIf`) already restricted 5xx/429/408 retries to safe methods, but the transport-exception/timeout retry applied to every method. A `POST` the node already received but slow-ACKed — e.g. a transaction broadcast — could be resent up to three times once the 15s read timeout fired.

This replaces `retryOnException` with a method-gated `retryOnExceptionIf` so transport/timeout retries apply only to idempotent methods (GET/HEAD/OPTIONS), reusing one `isSafeMethod` helper shared with the status predicate. The `CancellationException` guard is kept because `HttpRequestRetry` hands the cause straight to the predicate without filtering coroutine cancellation, so a blanket retry would break structured concurrency. Retry count, exponential backoff, the status predicate, and OkHttp's `retryOnConnectionFailure` are unchanged. It mirrors the dedicated QBTC proof client, which already disables implicit retry for its non-idempotent `/prove` POST.

## Which feature is affected?
- [x] Sending  — broadcast POSTs (timeout-failure path only)
- [x] Swap  — broadcast POSTs (timeout-failure path only)

No happy-path change to attach a tx link for: a successful broadcast is unaffected; only the retry behavior on a transport timeout changes, which is covered by unit tests.

## Checklist

Not applicable — networking/transport-layer change, no UI surface.

## Additional context

Co-sign / keysign: the relay POSTs in `SessionApi` (`startSession`, `sendTssMessage`, the completion markers, `uploadSetupMessage`) and the FastVault co-sign POSTs use the same shared client, so they are also no longer auto-retried on a transport timeout. They are idempotent (session/party-set keyed with server-side dedup), `startSession`/`startWithCommittee` keep their own app-level `withRelayRetry`, OkHttp still retries genuine pre-send connection failures, and the keygen/keysign flow is user-retryable — so correctness is unaffected. This matches iOS, where broadcasts are single-shot and a timeout is treated as inconclusive (verified on-chain, never auto-rebroadcast) rather than resent.

Most broadcast paths already recover a transaction that actually landed via `BroadcastTxUseCase.recoverIfAlreadyBroadcast` (BTC, Solana, Ton, Tron, Ripple, Cardano, Polkadot). For the few without a precomputed-hash recovery path (EVM, Cosmos/THORChain/Maya, Sui, Bittensor), a read-timeout now surfaces an error the user retries — the node then rejects the duplicate (nonce-too-low / mempool-cache / "already known"), which is mapped to success — instead of an implicit resend. Extending `recoverIfAlreadyBroadcast` to those chains is a reasonable follow-up, out of scope here.

Tests: `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.networkutils.HttpClientConfiguratorTest"` is green (8 tests) — the test that pinned POST-being-retried is inverted, and a read-timeout pair is added (POST not resent; GET still retried three times). The full `:data:testDebugUnitTest` suite is green and `:data:ktfmtCheck` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP request retry logic to prevent unnecessary retries of non-idempotent requests (POST, PUT, DELETE)
  * Enhanced timeout and cancellation exception handling to avoid redundant operation retries
  * Safe HTTP methods (GET, HEAD, OPTIONS) continue to retry appropriately for better reliability

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/networkutils/HttpClientConfigurator.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/networkutils/HttpClientConfigurator.kt
@@ -15,6 +15,7 @@ import io.ktor.serialization.kotlinx.json.json
 import io.ktor.util.appendIfNameAbsent
 import java.io.IOException
 import javax.inject.Inject
+import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.serialization.json.Json
 
 /**
@@ -27,8 +28,10 @@ import kotlinx.serialization.json.Json
  * 3. **HttpCallValidator** — converts transport-level [IOException]s into [NetworkException] with
  *    `httpStatusCode = 0` (client-side error). This runs *after* [HttpRequestRetry] exhausts its
  *    retries.
- * 4. **HttpRequestRetry** — retries on [IOException] (up to 3 times with exponential backoff) and
- *    on 5xx/429/408 responses for safe HTTP methods.
+ * 4. **HttpRequestRetry** — retries idempotent (GET/HEAD/OPTIONS) requests up to 3 times with
+ *    exponential backoff: on transport [IOException]s/read-timeouts and on 5xx/429/408 responses.
+ *    Non-idempotent requests (e.g. POST broadcasts) are never auto-retried, so a write the server
+ *    received but slow-ACKed is not silently resent.
  *
  * The retry plugin fires first: if all retries fail, the [IOException] propagates to
  * [HttpCallValidator], which wraps it in a [NetworkException].
@@ -59,16 +62,18 @@ class HttpClientConfigurator @Inject constructor(private val json: Json) {
 
             install(HttpRequestRetry) {
                 exponentialDelay()
-                retryOnException(maxRetries = 3, retryOnTimeout = true)
+
+                // Retry transport failures (IOException, read-timeout) only for idempotent
+                // methods. A non-idempotent request — e.g. a transaction broadcast the node
+                // already received but slow-ACKed — must not be silently resent. The
+                // CancellationException guard is required because HttpRequestRetry passes the
+                // cause straight to this predicate without filtering out coroutine cancellation.
+                retryOnExceptionIf(maxRetries = 3) { request, cause ->
+                    isSafeMethod(request.method) && cause !is CancellationException
+                }
 
                 retryIf { request, response ->
-                    val method = request.method
                     val status = response.status.value
-
-                    val isSafeMethod =
-                        method == HttpMethod.Get ||
-                            method == HttpMethod.Head ||
-                            method == HttpMethod.Options
 
                     val isServerError = status >= 500
                     val isRetriableStatus =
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4937, opened 2026-06-16
fix(swap): take a plain fraction of the balance for 25/50/75% chips

### LINKED ISSUE #4933: [Fix] Swap 25/50/75% chips deduct the network fee before applying the fraction [Swap]
# [Fix] Swap 25/50/75% chips deduct the network fee before applying the fraction [Swap]

## Problem
On the swap form, the 25%/50%/75% chips and MAX all run through one handler that computes a single "max usable" amount (balance minus the source-chain network fee for a native source) and then multiplies it by the fraction. After PR #4903 stopped the provider swap fee from being subtracted on these taps, the source-chain network fee is still subtracted before the fraction is applied — so a 50% tap on a native source yields (balance − network fee) × 0.5 rather than balance × 0.5. iOS and the desktop app take a plain fraction of the full balance for 25/50/75 and only reserve the network fee for MAX, so the same balance produces a different amount on Android.

## Expected Behavior
25/50/75% give a plain fraction of the full source balance (no fee subtraction). MAX reserves the source-chain network fee for a native source on its own gas chain, as it does today.

## Steps to Reproduce
1. Open Swap with a native source token (e.g. ETH) that has a balance and a live quote.
2. Tap 50%.
3. The inserted amount is (balance − network fee) × 0.5 instead of half of the full balance.

## Solution
Reserve the network-fee deduction for the MAX selection only; for the 25/50/75 fractions take the fraction of the full balance. This mirrors manual entry (which was already correct) and the other clients, and leaves the MAX/native behaviour and the insufficient-balance handling unchanged. Same approach as PR #4903, applied to the non-MAX fractions.

### Must Do
- [ ] 25/50/75% compute a plain fraction of the full source balance with no fee subtraction
- [ ] MAX keeps reserving the source-chain network fee for a native source on its own gas chain
- [ ] Add a ViewModel test asserting a non-MAX fraction on a native source does not subtract the network fee

### Must NOT Do
- Don't reintroduce the provider/swap fee subtraction removed in PR #4903
- Don't change MAX behaviour or the insufficient-balance error path
- Don't touch the quote pipeline, signing, or payload construction

## Acceptance Criteria
- [ ] `./gradlew :app:testDebugUnitTest` passes
- [ ] New test: a 0.5 fraction on a native source with a non-zero network fee yields balance × 0.5 (fails before the change, passes after)
- [ ] MAX on the same native source still subtracts the network fee

rkokhatskyi:
## Description

Fixes #4933

On the swap form, the 25%/50%/75% chips and MAX run through one handler that computed a single "max usable" amount (balance minus the source-chain network fee for a native source) and then multiplied it by the fraction. After PR #4903 stopped the provider swap fee from being subtracted here, the source-chain network fee was still subtracted before the fraction was applied — so a 50% tap on a native source produced half of (balance − network fee) instead of half of the balance. iOS and the desktop app take a plain fraction of the full balance for 25/50/75 and reserve the network fee only for MAX, so the same balance produced a different amount on Android.

This gates the network-fee deduction on the MAX selection; the 25/50/75 fractions now take a plain fraction of the full balance, matching manual entry and the other clients. MAX (native source on its own gas chain) still reserves the network fee, and the insufficient-balance handling is unchanged.

## Which feature is affected?
- [x] Swap - amount-input fix only (the computed source amount before quoting); the resulting swap is identical to typing the same amount, so there is no new on-chain behavior to link

## Checklist

N/A — this is a ViewModel computation fix. The visible effect (the amount field holding a plain fraction of the balance) is produced by runtime logic against a live quote, which the static preview harness can't reproduce faithfully. The behavior is pinned by a unit test that fails before the change (`expected: <1> but was: <0.9995>`) and passes after.

## Additional context

Co-sign is unaffected. `selectSrcPercentage` only writes to the shared source-amount field — the same channel manual typing uses — so the staged swap amount, payload construction, session, approval, and broadcast are all identical to entering the value by hand. No signing/co-sign code is touched.

Verification: `./gradlew :app:testDebugUnitTest --tests "*SwapFormViewModelTest"` → all passing (the existing MAX test still pins the native network-fee reservation; the 50% test now pins the plain-fraction behavior).

## Release Notes

* **Bug Fixes**
  * Adjusted swap percentage selection behavior: network fees are now only deducted when selecting the maximum amount. Selecting 50%, 75%, or other partial percentages now reserves the full selected amount without fee deduction.

* **Tests**
  * Updated test cases to reflect revised percentage selection behavior.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -339,16 +339,21 @@ constructor(
         // "insufficient balance" warning on a now-valid amount and mask the live quote/formError.
         _uiState.update { it.copy(error = null) }
 
-        // Only the source-chain network fee may be deducted from the source balance, and only for a
-        // native source on its own gas chain. The provider swap fee is taken from the destination
-        // amount (for LI.FI it is denominated in the destination token's units), so subtracting it
-        // here would mix decimals and could wrongly drive the usable amount negative for a
-        // low-decimal source into a high-decimal destination.
-        val maxUsableTokenAmount =
-            srcTokenValue.value -
-                (quotePipeline.estimatedNetworkFeeTokenValue.value?.value?.takeIf {
+        // The 25/50/75 chips take a plain fraction of the full balance, matching iOS and the
+        // desktop app. Only MAX reserves the source-chain network fee, and only for a native source
+        // on its own gas chain. The provider swap fee is taken from the destination amount (for
+        // LI.FI it is denominated in the destination token's units), so it is never deducted from
+        // the source balance here — that would mix decimals and could wrongly drive the usable
+        // amount negative for a low-decimal source into a high-decimal destination.
+        val reservedNetworkFee =
+            if (percentage >= 1f) {
+                quotePipeline.estimatedNetworkFeeTokenValue.value?.value?.takeIf {
                     srcToken.isNativeToken && quotePipeline.gasFeeChain.value == srcToken.chain
-                } ?: BigInteger.ZERO)
+                } ?: BigInteger.ZERO
+            } else {
+                BigInteger.ZERO
+            }
+        val maxUsableTokenAmount = srcTokenValue.value - reservedNetworkFee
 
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #4938, opened 2026-06-16
refactor(address-book): return null from getEntry instead of throwing on a miss

### LINKED ISSUE #4934: [Fix] Address-book lookup throws on a missing entry, forcing defensive wrappers [Address Book]
# [Fix] Address-book lookup throws on a missing entry, forcing defensive wrappers [Address Book]

## Problem
`AddressBookEntryDao.getEntry(...)` returns a non-null entity and throws when no row matches, so every caller compensates: the verify and keysign screens pair an `entryExists(...)` pre-check with a `runCatching { getEntry(...).title }.getOrNull()`. That is fragile — a future caller that forgets the wrapper crashes on an ordinary "address not saved" miss — and the exists-then-read pair issues two queries for a single lookup. Following the address-book work in PR #4917, this tidies the read path.

## Expected Behavior
`getEntry` returns `null` when there is no matching entry, and callers read `getEntry(...)?.title` with no `runCatching` and no redundant existence pre-check where it only gated the read.

## Steps to Reproduce
Not user-visible today (all current callers guard the throw). It surfaces as a latent crash risk: calling `getEntry` for an address that isn't in the address book throws `NoSuchElementException` instead of returning nothing.

## Solution
Make `getEntry` (and its case-insensitive variant) return a nullable entity through the DAO and repository, and simplify the read sites to `getEntry(...)?.title`. Drop the existence pre-check where it only gated the read; keep `entryExists` where its boolean result is independently used (the "save to address book" hint, the edit-vs-create branch). No change to stored values, casing, the case-insensitive matching from #4917, or any signing/keysign payload — only the recipient display title.

### Must Do
- [ ] `getEntry` returns a nullable entity through `AddressBookEntryDao` and `AddressBookRepository`
- [ ] Read sites use `getEntry(...)?.title`; remove the `runCatching` wrappers
- [ ] Add a test asserting `getEntry` returns null for a missing entry

### Must NOT Do
- Don't change stored address casing/values or add a database migration
- Don't alter the case-insensitive (EVM) matching behaviour added in #4917
- Don't touch signing/keysign payload construction, session, or broadcast (this only affects the displayed address-book title)

## Acceptance Criteria
- [ ] `./gradlew :data:testDebugUnitTest` and `./gradlew :app:testDebugUnitTest` pass
- [ ] New test: `getEntry` returns null for an address that is not in the address book
- [ ] `grep -rn "runCatching" app/src/main/java/com/vultisig/wallet/ui/models | grep getEntry` returns nothing

rkokhatskyi:
## Description

Fixes #4934

`AddressBookEntryDao.getEntry(...)` returned a non-null entity and threw when no row matched, so the verify and keysign screens compensated by pairing an `entryExists(...)` pre-check with `runCatching { getEntry(...).title }.getOrNull()`. That contract is fragile — a future caller that forgets the wrapper would crash on an ordinary "address not saved" miss — and the exists-then-read pair issues two queries for one lookup.

This makes `getEntry` (and its case-insensitive variant) return a nullable entity through the DAO and repository, and reads `getEntry(...)?.title` at the call sites. The existence pre-check is dropped where it only gated the read; `entryExists` stays where its boolean result is independently used (the "save to address book" hint on the keysign screen and the edit-vs-create branch in the address entry screen). No change to stored values/casing, the case-insensitive matching from #4917, or any signing/keysign payload — only the recipient display title.

## Which feature is affected?

## Checklist

N/A — no UI surface change. The address-book lookup only resolves the recipient display label and the "save to address book" hint; both are covered by unit tests.

## Additional context

Co-sign is unaffected. The lookup feeds only the displayed address-book title and the "save to address book" flag on the verify/keysign screens; it does not influence the signing payload, session, approval, or broadcast. The keysign path keeps a graceful fallback — if the entry is removed between the existence check and the read, `getEntry(...)?.title` now yields `null` instead of throwing, exactly as the old `runCatching` did.

Verification: `./gradlew :data:testDebugUnitTest` and `./gradlew :app:testDebugUnitTest --tests "*AddressEntryViewModelTest" --tests "*VerifyTransactionViewModelTest" --tests "*KeysignViewModel*Test"` pass; a new repository test asserts `getEntry` returns null for an address that is not saved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved address book lookup handling to properly handle missing address entries instead of suppressing errors.

* **Refactor**
  * Simplified address resolution logic across transaction verification and keysign flows for better reliability.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VerifyTransactionViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VerifyTransactionViewModel.kt
@@ -340,12 +340,9 @@ constructor(
                         }
                     }
                     ?.name
-            val dstInAddressBook =
-                dstVaultName == null && addressBookRepository.entryExists(chain.id, tx.dstAddress)
             val dstAddressBookTitle =
-                if (dstInAddressBook) {
-                    runCatching { addressBookRepository.getEntry(chain.id, tx.dstAddress).title }
-                        .getOrNull()
+                if (dstVaultName == null) {
+                    addressBookRepository.getEntry(chain.id, tx.dstAddress)?.title
                 } else null
 
             val memo = tx.memo
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSendUiModelBuilder.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSendUiModelBuilder.kt
@@ -193,12 +193,9 @@ constructor(
                     }
                 }
                 ?.name
-        val isSavedInAddressBook =
-            dstVaultName == null && addressBookRepository.entryExists(chain.id, payload.toAddress)
         val dstAddressBookTitle =
-            if (isSavedInAddressBook) {
-                runCatching { addressBookRepository.getEntry(chain.id, payload.toAddress).title }
-                    .getOrNull()
+            if (dstVaultName == null) {
+                addressBookRepository.getEntry(chain.id, payload.toAddress)?.title
             } else null
 
         val isUnlimitedApproval =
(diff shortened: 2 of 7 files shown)


johnnyluo (approved):
LGTM

PR #4942, opened 2026-06-16
fix(keygen): skip QR flash and blank frame in Fast Vault peer discovery

### LINKED ISSUE #4922: Fast Vault: Peer Discovery (Scan QR) screen flashes before Connecting screen
## Summary

In the **Fast Vault (standard) creation flow**, after tapping **Next** on the "Choose a password" screen, the **Peer Discovery ("Scan QR" / "Devices (1/1)") screen flashes briefly** before the app transitions to the "Connecting…" screen. In the Fast Vault flow the VultiServer is the second party, so the peer-discovery / QR screen should **never** be shown to the user — the flow should go straight from the password screen into the "Connecting…" state.

There is also a short **blank (empty dark) screen** between the QR flash and the "Connecting…" animation.

## Steps to reproduce

1. Start creating a new vault → **Fast Vault** (standard flow, with email + password).
2. Reach the **"Choose a password"** screen, enter a password, tap **Next**.
3. Observe the transition.

## Actual behavior

The transition shows three states in ~1.5s:

1. "Scan QR" / **"Devices (1/1)"** peer-discovery screen appears briefly (empty QR box, "This Device" + "Scan with 2nd device").
2. A **blank dark screen** for roughly ~0.3–0.5s.
3. The **"Connecting…"** animation finally appears.

Timeline captured from a screen recording (~2.9s total):

| Moment | Screen |
|--------|--------|
| Tap **Next** on "Choose a password" | password |
| ~1.1s — **Peer Discovery flashes** (should not appear) | scanqr |
| ~1.6s — **blank screen** | blank |
| ~2.9s — **Connecting…** | connecting |

## Expected behavior

For the Fast Vault standard flow, tapping **Next** should transition directly from the password screen into the **"Connecting…"** state. The peer-discovery / "Scan QR" / "Devices" screen should not be shown, and there should be no intermediate blank frame.

## Technical analysis

The peer-discovery screen chooses what to render from `PeerDiscoveryUiModel` state (`KeygenPeerDiscoveryScreen` in `PeerDiscoveryScreen.kt`):

```
error != null            -> Error
warning != null          -> ErrorView
connectingToServer != null -> Connecting (Rive) screen
else                     -> PeerDiscoveryScreen (QR / devices)
```

Two things combine to produce the flash:

1. **The QR/devices flash.** The initial `PeerDiscoveryUiModel` is created with `connectingToServer = null`, so the very first composition renders the `else` branch — the full QR / "Devices (1/1)" screen. `connectingToServer` is only set to non-null inside `startVultiServerConnection()`, which is reached from `loadData()`. `loadData()` runs asynchronously (it awaits the DKLS secret-setting flow, the feature-flags fetch, and optionally a vault lookup) before it ever hits the Fast-Vault branch:

   ```kotlin
   if (!email.isNullOrBlank() && !password.isNullOrBlank()) {
       if (args.action == TssAction.Migrate && signers.count() > 2) { ... }
       else startVultiServerConnection()   // <- only here is connectingToServer set
   } else { startPeerDiscovery() }
   ```

   During that async window the QR/devices screen is on screen — that is the visible flash.

2. **The blank frame.** When `connectingToServer` does flip to non-null, the `connectingToServer != null` branch defers rendering its Rive animation by 300ms:

   ```kotlin
   var showRive by remember { mutableStateOf(false) }
   LaunchedEffect(Unit) { kotlinx.coroutines.delay(300); showRive = true }
   if (showRive) { RiveAnimation(...) }
   ```

   For those 300ms nothing is drawn → the blank dark screen.

## Suggested fix direction

Seed the connecting state up front for the Fast Vault flow so the QR/devices `else` branch is never composed — e.g. initialize `connectingToServer = ConnectingToServerUiModel(false)` in the initial state when `email`/`password` are present and the flow is not the active-vault migrate (`signers > 2`) case (the one path that intentionally shows peer discovery). Separately, consider rendering a non-empty placeholder during the 300ms `showRive` delay so there is no blank frame.

## Environment

- Branch observed: `aminsato/4884_slim_keygen_peerdiscovery_vm`
- Flow: Fast Vault (standard), Internet/relay mode
- Affected files: `app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt`, `app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt`

> Note: the `connectingToServer = null` initial state and async `loadData()` also exist on `main`, so this is likely a pre-existing UX issue rather than a regression introduced by the peer-discovery refactor.

aminsato:
Fixes #4922.

## Problem

In the **Fast Vault (standard) creation flow**, tapping **Next** on the "Choose a password" screen briefly flashed the **Peer Discovery ("Scan QR" / "Devices (1/1)") screen**, then a **blank dark frame**, before the "Connecting…" animation appeared. In the Fast Vault flow the VultiServer is the second party, so the peer-discovery / QR screen should never be shown.

## Root cause

1. **QR/devices flash** — the initial `PeerDiscoveryUiModel` is created with `connectingToServer = null`, so the very first composition renders the `else` branch (the full QR / "Devices" screen). `connectingToServer` is only set non-null inside `startVultiServerConnection()`, reached after `loadData()` finishes its async work (DKLS secret setup, feature-flags fetch, optional vault lookup). The QR screen is on-screen during that window.
2. **Blank frame** — the connecting branch defers its Rive animation by 300ms (`showRive`), drawing nothing in the meantime.

## Fix

- **`KeygenPeerDiscoveryViewModel`** — seed `connectingToServer = ConnectingToServerUiModel(false)` in the initial state when it's a Fast Vault flow (`email` + `password` present), so the QR/devices `else` branch is never composed. Excludes `Migrate`: the active-vault migrate (`signers > 2`) path intentionally shows peer discovery, and the signer count isn't knowable synchronously at construction.
- **`PeerDiscoveryScreen`** — wrap the connecting branch in a `Box` filled with `Theme.v2.colors.backgrounds.primary`, so the 300ms pre-Rive window renders the app backdrop rather than a blank dark frame.

## Test plan

- [ ] Fast Vault (standard, email + password) → password screen → **Next**: transitions directly into "Connecting…" with no QR flash and no blank frame.
- [ ] Secure vault (2/2, 3/3, 4+) keygen still shows the QR / peer-discovery screen.
- [ ] Active-vault migrate (vault with > 2 signers) still shows the QR / peer-discovery screen.
- [ ] Reshare / KeyImport Fast Vault flows go straight to "Connecting…".

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented a blank/empty peer discovery screen during connection initialization by ensuring the initial UI state is properly set before async loading completes.
  * Removed visual gaps during peer discovery startup by rendering a solid background behind the loading animation until it appears.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -160,17 +160,15 @@ constructor(
                 deviceCount = args?.deviceCount,
                 allowsMoreDevices =
                     (args?.deviceCount ?: MIN_KEYGEN_DEVICES) >= UNBOUNDED_KEYGEN_DEVICE_COUNT,
-                // Seed the connecting state for Fast Vault so the QR/devices branch is never
-                // composed during the async loadData() window (it would otherwise flash before
-                // startVultiServerConnection() flips this to non-null). Excludes Migrate: the
-                // active-vault migrate (signers > 2) path intentionally shows peer discovery, and
-                // we can't know the signer count synchronously here.
+                // Seed the connecting state for every Fast Vault flow (credentials present) so the
+                // QR/devices branch is never composed during the async loadData() window (it would
+                // otherwise flash before startVultiServerConnection() flips this to non-null).
+                // loadData() clears this back to null for the only Fast Vault path that shows peer
+                // discovery instead of connecting to the server (active-vault migrate, signers >
+                // 2),
+                // which we cannot detect synchronously here.
                 connectingToServer =
-                    if (
-                        !args?.email.isNullOrBlank() &&
-                            !args.password.isNullOrBlank() &&
-                            args.action != TssAction.Migrate
-                    )
+                    if (!args?.email.isNullOrBlank() && !args.password.isNullOrBlank())
                         ConnectingToServerUiModel(false)
                     else null,
                 enableNotification = false,
@@ -487,6 +485,10 @@ constructor(
                 // join
                 // Also need to request the server to join the upgrade process
                 if (args.action == TssAction.Migrate && signers.count() > 2) {
+                    // This is the only Fast Vault path that shows peer discovery rather than the
+                    // connecting screen, so undo the connecting state seeded in the initial UI
+                    // model.
+                    state.update { it.copy(connectingToServer = null) }
                     startPeerDiscovery()
                     requestVultiServerConnection()
                 } else {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -160,6 +160,19 @@ constructor(
                 deviceCount = args?.deviceCount,
                 allowsMoreDevices =
                     (args?.deviceCount ?: MIN_KEYGEN_DEVICES) >= UNBOUNDED_KEYGEN_DEVICE_COUNT,
+                // Seed the connecting state for Fast Vault so the QR/devices branch is never
+                // composed during the async loadData() window (it would otherwise flash before
+                // startVultiServerConnection() flips this to non-null). Excludes Migrate: the
+                // active-vault migrate (signers > 2) path intentionally shows peer discovery, and
+                // we can't know the signer count synchronously here.
+                connectingToServer =
+                    if (
+                        !args?.email.isNullOrBlank() &&
+                            !args.password.isNullOrBlank() &&
+                            args.action != TssAction.Migrate
+                    )
+                        ConnectingToServerUiModel(false)
+                    else null,
                 enableNotification = false,
             )
         )
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt
@@ -133,13 +133,17 @@ internal fun KeygenPeerDiscoveryScreen(model: KeygenPeerDiscoveryViewModel = hil
                 showRive = true
             }
 
-            if (showRive) {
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -160,6 +160,19 @@ constructor(
...
+                // Seed the connecting state for Fast Vault so the QR/devices branch is never
+                // composed during the async loadData() window (it would otherwise flash before
+                // startVultiServerConnection() flips this to non-null). Excludes Migrate: the
+                // active-vault migrate (signers > 2) path intentionally shows peer discovery, and
+                // we can't know the signer count synchronously here.
+                connectingToServer =
+                    if (
+                        !args?.email.isNullOrBlank() &&
+                            !args.password.isNullOrBlank() &&
+                            args.action != TssAction.Migrate
```
rkokhatskyi: The seed excludes all `Migrate`, but `loadData` sends a 2 signer Fast Vault migrate to `startVultiServerConnection` because `Migrate && signers.count() > 2` is false, so the QR screen this PR removes still flashes during the async load, reachable via Vault Settings migrate through `FastVaultEmailViewModel`

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
aminsato: Good catch — you're right. I was excluding all Migrate, but the 2-signer Fast Vault migrate falls into the `else` and routes to `startVultiServerConnection()`, so it kept flashing the QR. Fixed in 1895014c5: the seed now covers every Fast Vault flow (credentials present), and `loadData()` clears `connectingToServer` back to null only in the `Migrate && signers > 2` branch — the one path that actually shows peer discovery. So the 2-signer migrate now stays on the connecting screen, and the multi-device migrate still shows QR.

johnnyluo (approved):
LGTM

PR #4943, opened 2026-06-16
fix(reshare): require all vault devices to join before start (#4940)

### LINKED ISSUE #4940: Reshare: validate all required vault devices joined the QR before starting
## Summary

In the **reshare** flow, if one of the vault's existing devices (pairs) does not join the reshare QR code, the initiator can still start the reshare. The reshare/signing then fails because not all required parties are present.

There is no pre-start validation that **all** devices of the vault have joined the reshare session before it begins.

## Steps to Reproduce

1. Have a vault with multiple paired devices (e.g. 2-of-3).
2. On the initiating device, start a **Reshare** and show the QR code.
3. Have only some — but not all — of the required existing devices scan/join the QR.
4. Start the reshare.

## Actual Behavior

- Reshare starts even though not every required device joined.
- The reshare/signing fails partway through, leaving the user with a failed operation and no clear reason.

## Expected Behavior

- Before allowing reshare to start, the app should verify that **all** of the vault's required devices have joined the reshare session.
- If any required device is missing, the start action should be blocked with a clear message indicating which/how many devices still need to join.

## Notes

- The peer-discovery screen should reflect the full set of expected devices for the vault and only enable "Start" once the threshold of required participants for a valid reshare are present.
- This affects all platforms — companion issues are being filed on Android, Windows, and iOS for parity.

aminsato:
## Summary

In the reshare flow, the peer-discovery screen kept the default `MIN_KEYGEN_DEVICES` (2) threshold, so the initiator could start a reshare once only **two** devices had joined the QR session — even for a vault with more signers. The reshare then failed partway through because the missing required parties never participated, leaving the user with a failed operation and no clear reason.

Fixes #4940.

## Root cause

`KeygenPeerDiscoveryViewModel.loadData()` only raised `minimumDevices` to the existing vault's signer count for `Migrate` and `SingleKeygen`. `ReShare` was omitted, so it fell back to the default threshold of 2.

## Fix

Add `TssAction.ReShare` to the existing-vault branch that sets `minimumDevices`/`minimumDevicesDisplayed = existingVault.signers.size`, mirroring the already-reviewed `Migrate`/`SingleKeygen` handling.

As a result:
- The **Start / Next** button only enables once every one of the vault's existing devices has rejoined the reshare session.
- The **"X of Y" device counter** reflects the full set of expected devices instead of a hardcoded 2.

## Scope note

This targets the secure-vault peer-discovery path (the issue's 2-of-3 repro). The FastVault auto-start dispatch flow is intentionally left untouched to keep the change focused and low-risk on the keygen/reshare flow; it can be addressed separately if needed.

## Test plan

- [x] Added unit test: reshare requires every existing vault device to join before start is enabled (asserts `minimumDevices`/`minimumDevicesDisplayed` track the vault's signer count).
- [x] `KeygenPeerDiscoveryViewModel` test suite passes (26 tests).
- [x] ktfmt applied.
- [ ] Manual multi-device reshare smoke test (e.g. 2-of-3): confirm Start stays disabled until all devices join.

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted peer-discovery device requirements for the reshare flow to use a signing-quorum threshold derived from the existing committee, rather than requiring the full signer count.
* **Tests**
  * Added/updated coverage for the manual reshare peer-discovery path (with the relevant feature flag state) to confirm minimum device thresholds are clamped to the expected signing quorum.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -457,7 +457,16 @@ constructor(
                 resharePrefix = existingVault.resharePrefix
                 signers = existingVault.signers
 
-                if (args.action == TssAction.Migrate || args.action == TssAction.SingleKeygen) {
+                // Migrate/SingleKeygen and ReShare all operate on an existing vault, so every
+                // one of its devices must rejoin before the session can start. Without this,
+                // ReShare keeps the default MIN_KEYGEN_DEVICES threshold and lets the initiator
+                // start once only 2 devices joined — the reshare then fails partway through
+                // because the missing parties never participated (issue #4940).
+                if (
+                    args.action == TssAction.Migrate ||
+                        args.action == TssAction.SingleKeygen ||
+                        args.action == TssAction.ReShare
+                ) {
                     state.update {
                         it.copy(
                             minimumDevices = existingVault.signers.size,
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -457,22 +457,29 @@ constructor(
                 resharePrefix = existingVault.resharePrefix
                 signers = existingVault.signers
 
-                // Migrate/SingleKeygen and ReShare all operate on an existing vault, so every
-                // one of its devices must rejoin before the session can start. Without this,
-                // ReShare keeps the default MIN_KEYGEN_DEVICES threshold and lets the initiator
-                // start once only 2 devices joined — the reshare then fails partway through
-                // because the missing parties never participated (issue #4940).
-                if (
-                    args.action == TssAction.Migrate ||
-                        args.action == TssAction.SingleKeygen ||
-                        args.action == TssAction.ReShare
-                ) {
-                    state.update {
-                        it.copy(
-                            minimumDevices = existingVault.signers.size,
-                            minimumDevicesDisplayed = existingVault.signers.size,
-                        )
+                when (args.action) {
+                    // Migrate/SingleKeygen recreate the full vault, so every existing signer
+                    // must rejoin before the session can start.
+                    TssAction.Migrate,
+                    TssAction.SingleKeygen ->
+                        state.update {
+                            it.copy(
+                                minimumDevices = existingVault.signers.size,
+                                minimumDevicesDisplayed = existingVault.signers.size,
+                            )
+                        }
+                    // ReShare only needs a signing quorum of the old committee — matching iOS
+                    // getThreshold()+1, keysign discovery, and the reshare_start_screen_warning
+                    // contract. Requiring every signer would deadlock a reshare that intentionally
+                    // drops a lost device, since next() already filters oldCommittee to the
+                    // present signers.
+                    TssAction.ReShare -> {
+                        val threshold = Utils.getThreshold(existingVault.signers.size)
+                        state.update {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -457,7 +457,16 @@ constructor(
...
-                if (args.action == TssAction.Migrate || args.action == TssAction.SingleKeygen) {
+                // Migrate/SingleKeygen and ReShare all operate on an existing vault, so every
+                // one of its devices must rejoin before the session can start. Without this,
+                // ReShare keeps the default MIN_KEYGEN_DEVICES threshold and lets the initiator
+                // start once only 2 devices joined — the reshare then fails partway through
+                // because the missing parties never participated (issue #4940).
+                if (
+                    args.action == TssAction.Migrate ||
+                        args.action == TssAction.SingleKeygen ||
+                        args.action == TssAction.ReShare
```
rkokhatskyi: Adding `ReShare` sets `minimumDevices = existingVault.signers.size`, which deadlocks reshare once any device is lost because `hasEnoughDevices` needs `selectedDevicesSize >= signers.size`, yet `next()` filters `oldCommittee` to present signers so a reduced committee reshare is valid. Use `Utils.getThreshold(signers.size)` to match iOS `getThreshold()+1`, keysign discovery, and the `reshare_start_screen_warning` threshold contract

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -457,7 +457,16 @@ constructor(
                 resharePrefix = existingVault.resharePrefix
                 signers = existingVault.signers
 
-                if (args.action == TssAction.Migrate || args.action == TssAction.SingleKeygen) {
+                // Migrate/SingleKeygen and ReShare all operate on an existing vault, so every
+                // one of its devices must rejoin before the session can start. Without this,
+                // ReShare keeps the default MIN_KEYGEN_DEVICES threshold and lets the initiator
+                // start once only 2 devices joined — the reshare then fails partway through
+                // because the missing parties never participated (issue #4940).
```
rkokhatskyi: The `Without this, ReShare keeps the default ... (issue #4940)` lines narrate prior behavior and the ticket rather than the rule the code enforces, and will go stale, so drop them for a single line invariant

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
aminsato: Agreed, you're right — clamping to `signers.size` deadlocks a reduced-committee reshare since `hasEnoughDevices = selectedDevicesSize >= minimumDevices` while `next()` already filters `oldCommittee` to the present signers. Fixed in e4e87b3ed: ReShare now uses `Utils.getThreshold(existingVault.signers.size)` (= iOS `getThreshold()+1`, matching keysign discovery and the `reshare_start_screen_warning` contract). Migrate/SingleKeygen still require the full signer set since they recreate the whole vault. Updated the unit test to assert the quorum threshold.

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
aminsato: Done — dropped the narration about prior behavior and the ticket. Each branch now carries a single-line invariant describing the rule the code enforces (full signer set for Migrate/SingleKeygen, signing quorum for ReShare).

johnnyluo:
close this one 

Technically , we need to have 2/3 of signers to join to start a reshare , so the changes doesn't make too much difference

Issue #4884, opened 2026-06-12
refactor(keygen): slim down God-object KeygenPeerDiscoveryViewModel


aminsato:
## Summary

`KeygenPeerDiscoveryViewModel` (~900 lines) has grown into a God ViewModel that mixes payload construction, FastVault server dispatch, Android mediator-service lifecycle, participant discovery, QR generation, and key-import extraction in a single class. It's the same shape as the ViewModels recently slimmed on the swap and deposit flows (`SwapFormViewModel`, `DepositFormViewModel`), and would benefit from the same treatment.

`app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt`

### 1. Per-`TssAction` logic is smeared across four sites (highest leverage)
The same 5-way switch over `TssAction` (`KEYGEN` / `KeyImport` / `ReShare` / `Migrate` / `SingleKeygen`) is duplicated in:
- `createKeygenPayload()` — proto payload per action (lines ~610–693)
- `requestVultiServerConnection()` — FastVault dispatch per action (lines ~704–826)
- `next()` — `libType` resolution (lines ~353–358)
- `toLinkTssType()` — link type string (lines ~695–702)

Adding or changing one action requires editing all four places (shotgun surgery). This mirrors the per-option strategy split already done for `DepositFormViewModel`. A `KeygenAction` strategy (one implementation per action) owning `buildPayload`, `joinServer`, `libType`, and `linkType` would collapse all four switches into one place.

### 2. Mutable `var` soup as implicit session state
`hexChainCode`, `localPartyId`, `libType`, `pubKeyEcdsa`, `signers`, `resharePrefix`, `serverUrl`, `isTssBatchEnabled` (lines ~176–203) are all `private var`, populated inside `loadData()` and read across many methods. This creates temporal coupling — anything running before `loadData()` completes observes stale defaults, and the initialization contract is invisible. Replace with a single immutable session/config object produced once by `loadData()`.

### 3. Android/mediator plumbing lives in the ViewModel
`BroadcastReceiver`, `IntentFilter`, `registerReceiver`, `MediatorService`, and the `onCleared()` unregister (lines ~835–898) are framework lifecycle concerns that make the class hard to unit-test. Extract into a dedicated controller / use case.

### 4. Inconsistent coroutine error handling
[...]

PR #4945, opened 2026-06-16
fix(keygen): start a fresh session when switching network mode (#4944)

### LINKED ISSUE #4944: Keygen: switching back to Internet mode shows stale peers from previous session
## Description

In the keygen peer discovery screen, switching the network mode from **Local** to **Internet** does not start a fresh Internet session. Instead, devices that joined during a *previous* Internet-mode session reappear in the device list as stale peers.

## Steps to Reproduce

1. Start keygen and open peer discovery in **Internet** mode.
2. Let one or more devices join over the Internet relay.
3. Switch to **Local** mode.
4. Switch back to **Internet** mode.

https://github.com/user-attachments/assets/bd72f50b-c85a-4994-8243-485eb9fd989a

## Actual Behavior

The old devices from the earlier Internet session show up again in the device list, even though a fresh connection was expected. The session is effectively being reused rather than re-created.

## Expected Behavior

Switching back to Internet mode should establish a **new connection / new session**, so only devices that join the new session appear. Stale peers from a prior session must not reappear.

## Notes

A screen recording demonstrating the issue is available (`Screen_recording_20260616_134822.mp4`) and can be attached to this issue.

aminsato:
## Summary

Closes #4944.

In keygen peer discovery, switching network mode **Internet → Local → Internet** resurfaced devices from a previous Internet session as stale peers.

**Root cause:** `KeygenPeerDiscoveryViewModel` minted `sessionId` once (`private val`) and reused it for the life of the screen. `switchMode()` cleared the visible device list but kept the same id, so reconnecting to the relay:
- found the original session still alive on `VULTISIG_RELAY_URL`,
- `isSessionStarted()` saw it still populated and skipped re-creating it,
- `startParticipantDiscovery()` re-polled the same session id and streamed the old peers back in.

**Fix:** mint a fresh `sessionId` (and cancel in-flight discovery) on every `switchMode()`, so reconnecting always starts a clean session and only peers that join the new session appear.

## Changes
- `sessionId` is now a `var`, regenerated in `switchMode()` alongside the existing device-list reset.
- Cancel `discoverParticipantsJob` before starting the new session to stop the old poller from re-emitting stale peers.

## Test plan
- Added unit test `switching back to Internet mode uses a fresh session id` — captures the `sessionId` passed to `sessionApi.startWithCommittee` across an Internet → Local → Internet cycle and asserts the two Internet sessions differ.
- `./gradlew :app:testReleaseUnitTest --tests "*KeygenPeerDiscoveryViewModelTest"` — all 22 tests green.
- Manual: start keygen in Internet mode, let a device join, switch to Local, switch back to Internet — old device no longer reappears.

## Note
Touches the same file as open PRs #4884 (KeygenPeerDiscoveryViewModel slim-down) and #4922 (Fast Vault peer-discovery flash); will rebase if either merges first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved peer discovery and keygen flows by regenerating and consistently reusing a fresh session ID across relay and mediator connections, avoiding stale peer associations.
  * Mode switching now fully cancels in-flight discovery/participant polling, restarts the relevant services, and uses the new session ID end-to-end.

* **Tests**
  * Enhanced unit tests to verify the correct session IDs are passed during discovery/connection.
  * Added coverage ensuring switching back to Internet mode creates a new committee session ID (Internet → Local → Internet).


app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -164,7 +164,7 @@ constructor(
             )
         )
 
-    private val sessionId = Uuid.random().toHexString()
+    private var sessionId = Uuid.random().toHexString()
```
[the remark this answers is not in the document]
aminsato: Fixed in c9d3f449f. `switchMode()` now cancels the whole previous run via a new `startPeerDiscoveryJob` (not just `discoverParticipantsJob`), captures the fresh id in a local, and threads it through `startPeerDiscovery` → `createKeygenPayload` / `startSessionWithRetry` / `startParticipantDiscovery` / `startMediatorService`, so each discovery run uses a single stable session. The async mediator broadcast reads a captured `mediatorSessionId` instead of the shared field.

johnnyluo:
conflict @aminsato

aminsato:
> conflict @aminsato

@johnnyluo Fixed — merged latest `main` and resolved the conflict in `KeygenPeerDiscoveryViewModel.kt` (Fast Vault peer-discovery block: kept the `sessionId`-threaded `startPeerDiscovery(sessionId)`/`requestVultiServerConnection(sessionId)` calls, since this branch changed those signatures and the no-arg form from `main` no longer compiles). Pushed.

johnnyluo:
@aminsato , can you please test keygen locally using "Local Mode"?   it is not working for me

johnnyluo:
> @aminsato , can you please test keygen locally using "Local Mode"? it is not working for me

I found the issue , raised a PR #4957

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -295,7 +309,7 @@ constructor(
                 selectedDevices = emptyList(),
             )
         }
-        viewModelScope.safeLaunch { startPeerDiscovery() }
+        startPeerDiscoveryJob = viewModelScope.safeLaunch { startPeerDiscovery(newSessionId) }
```
rkokhatskyi, line 312: `switchMode()` mints a fresh `sessionId` but never re-registers the VultiServer, so on the active-vault Migrate path where `loadData()` pairs `startPeerDiscovery(sessionId)` with `requestVultiServerConnection(sessionId)` at 530-531, toggling network strands the server on the dead old session and stalls the migrate. It must also call `requestVultiServerConnection(newSessionId)`

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -282,8 +285,19 @@ constructor(
...
+        //
+        // Switching network mode must also start a brand-new session. Reusing the same sessionId
+        // let the relay (or mediator) hand back peers that joined the previous session — e.g.
+        // Internet -> Local -> Internet would re-read the still-alive relay session and resurface
+        // stale devices (issue #4944). Cancel the whole previous discovery run (not just
+        // participant polling) so an already-running startPeerDiscovery() can't observe the new
+        // sessionId mid-flight, then mint a fresh id captured in a local and threaded through the
+        // run so every step uses the same session even if switchMode() fires again.
         discoverParticipantsJob?.cancel()
+        startPeerDiscoveryJob?.cancel()
```
rkokhatskyi, line 297: `startPeerDiscoveryJob` is assigned only in `switchMode` (312), so the initial run `loadData()` starts via `startPeerDiscovery(sessionId)` at 530/536 is untracked and this cancel is a no-op on the first switch. That leftover run then rebinds `discoverParticipantsJob` to the old relay session, resurfacing the stale peers #4944 targets

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
johnnyluo: when working with vultiserver , only internet mode is allowed , user can't switch to local, so that's ok

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
aminsato: Good catch — confirmed and fixed in 3fbac28c0. The initial discovery run from `loadData()` wasn't tracked, so on the first toggle `startPeerDiscoveryJob?.cancel()` was a no-op and an in-flight `startPeerDiscovery()` (e.g. still retrying in `startSessionWithRetry`) could resume after the switch and rebind `discoverParticipantsJob` to the old session.

`loadData()` now assigns its coroutine to `startPeerDiscoveryJob` (cancelling any prior run first), so `switchMode()` reliably tears down the previous run before starting a fresh session.

johnnyluo (approved):
LGTM

PR #4962, opened 2026-06-18
fix(keygen): reveal Scan QR only after its Rive frame renders (#4954)

### LINKED ISSUE #4954: QR code flashes un-styled before beautifying on keygen Scan QR screen
## Summary

On the keygen **"Scan QR"** (peer discovery) screen, after tapping **Next** on the **"Name your vault"** screen, the QR code first renders in a plain, **un-styled** form and only **"beautifies"** into the final branded/styled QR after about a second. This produces a visible flash where the user sees the raw QR before it gets its proper Vultisig styling.

## Steps to reproduce

1. Start vault creation (Secure vault flow).
2. Enter a name on the **"Name your vault"** screen and tap **Next**.
3. Observe the **"Scan QR"** screen as it appears.

## Actual behavior

For ~1 second the QR code shows as a plain/un-styled QR (see screenshot below), then it re-renders into the final beautified/branded QR.

## Expected behavior

The QR code should render in its final styled form immediately, with no intermediate plain-QR flash.

## Screenshot (un-styled QR shown for the first second)

unstyled-qr

## Notes

- Distinct from #4942 (which addressed the Fast Vault flow showing the QR/peer-discovery screen and a blank frame). This is about the **styling/beautification** of the QR itself flashing on the Secure vault peer-discovery screen.
- Repro recorded on an Android emulator (`sdk_gphone64`).

aminsato:
Closes #4954

## Summary

On the Secure-vault **Scan QR** (peer-discovery) screen the QR code first rendered in a bare, un-styled form and only "beautified" into the branded frame ~1s later (#4954).

**Root cause:** the QR `Image` is pure Compose and faded in the instant its bitmap was ready, while the decorative `riv_qr_scanned` frame is a Rive `AndroidView` whose first frame renders a few hundred ms after composition. So the QR appeared before its frame — there is no second QR generation; it's purely a render-ordering race.

## Fix

- `RiveAnimation.kt` — add an optional `onFirstFrame: () -> Unit` to the `@RawRes` overload. It registers a `RiveFileController.Listener` once in the view `factory` and fires the callback on the first `notifyAdvance` (first rendered frame), then unregisters. Default no-op, so all other callers are unaffected.
- `PeerDiscoveryScreen.kt` (`QrCodeContainer`) — gate the QR's `AnimatedVisibility` on `qrCode != null && frameReady`, where `frameReady` flips on `onFirstFrame`. It is seeded to `!isRiveInitialized` so the QR still shows immediately when Rive is unavailable or in previews. The existing `fireState("isSucces")` device-joined animation is untouched.

The QR and its branded frame now appear together.

## Before / After

| Before | After |
|--------|-------|
| before | after |

## Test plan

- Start Secure-vault creation → name the vault → tap **Next** → observe the **Scan QR** screen: the QR appears already inside its branded frame, with no intermediate un-styled flash.
- Confirm a second device joining still triggers the success animation (`isSucces`).
- Confirm previews / Rive-disabled fallback still render the QR.

## Release Notes

- **Enhancements**
  - Improved coordination between the first rendered Rive frame and QR code visibility during peer discovery.
  - Added a one-time “first frame” callback for Rive animations to support initial UI actions.

- **Bug Fixes**
  - Introduced a guarded Rive initialization flow to prevent repeated crash loops across app restarts.

- **Tests**
  - Updated Rive initialization tests to mock the new initialization entry point and added coverage for when the in-flight marker write fails.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/app/RiveInitializer.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/RiveInitializer.kt
@@ -1,0 +1,70 @@
+package com.vultisig.wallet.app
+
+import android.content.Context
+import android.content.SharedPreferences
+import androidx.annotation.VisibleForTesting
+import app.rive.runtime.kotlin.core.Rive
+import com.vultisig.wallet.BuildConfig
+import timber.log.Timber
+
+internal var isRiveInitialized: Boolean = false
+    private set
+
+@VisibleForTesting
+internal fun resetRiveInitialized() {
+    isRiveInitialized = false
+}
+
+private const val RIVE_GUARD_PREFS = "rive_init_guard"
+private const val KEY_RIVE_INIT_IN_FLIGHT_VERSION = "rive_init_in_flight_version"
+internal const val NO_IN_FLIGHT_VERSION = -1
+
+/**
+ * Returns the [SharedPreferences] backing the Rive crash-loop guard.
+ *
+ * Kept as a plain (unencrypted) prefs file — the only stored value is an int "init attempted"
+ * marker keyed by app version — and resolved synchronously so [initializeRive] can read/write
+ * before invoking the native Rive entry point.
+ */
+@VisibleForTesting
+internal fun riveGuardPrefs(context: Context): SharedPreferences =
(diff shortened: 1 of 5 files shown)


johnnyluo (approved):
LGTM

PR #4964, opened 2026-06-18
fix(keygen): surface the real cause on session-start failure (timeout vs. generic)

### LINKED ISSUE #4956: Keygen: generic 'Something went wrong' on network timeout hides the real cause
## Summary

When starting a keygen session fails due to a network/server timeout, the UI shows a generic **"Something went wrong / An error occurred"** screen that gives the user no indication of what actually went wrong (e.g. a network timeout vs. an internal failure). The underlying error is also mislabeled in logs.

## Steps to reproduce

1. Start a keygen flow (peer discovery).
2. The call to `router/start/<sessionId>` times out (slow/unreachable server, flaky connection).
3. After 3 retry attempts fail, the error screen is shown.

## Actual behavior

- The error screen displays only the generic strings `error_view_default_title` ("Something went wrong") and `error_view_default_description` ("An error occurred"), so the user cannot tell whether it's a network problem, a timeout, or something else — and gets no guidance on what to do.
- In logs the failure surfaces as `NetworkException: No internet connection`, even though the device has connectivity. The real cause is a `SocketTimeoutException` from the `router/start` request. Wrapping every `IOException` as "No internet connection" is misleading both to users and when triaging logs.

## Expected behavior

- The keygen error screen should reflect the actual failure cause. At minimum, distinguish a network/timeout failure ("Couldn't reach the server, check your connection and try again") from a generic error.
- The error message used for an `IOException` should not unconditionally say "No internet connection". A socket/read timeout against the server should be reported as a timeout, not as a lack of internet.

## Relevant log

```
REQUEST https://api.vultisig.com/router/start/<REDACTED_SESSION_ID> failed with exception:
java.net.SocketTimeoutException: Socket timeout has expired [url=.../router/start/<REDACTED_SESSION_ID>, socket_timeout=unknown] ms
E  Failed to start keygen session
E  com.vultisig.wallet.data.utils.NetworkException: No internet connection
   at ...HttpClientConfigurator$configure$1$3$1.invokeSuspend(HttpClientConfigurator.kt:54)
   ...
Caused by: java.net.SocketTimeoutException: Socket timeout has expired [url=.../router/start/<REDACTED_SESSION_ID>, socket_timeout=unknown] ms
   at io.ktor.client.engine.okhttp.OkUtilsKt.mapOkHttpException(OkUtils.kt:82)
   ...
Caused by: java.net.SocketTimeoutException: timeout
   at okhttp3.internal.http2.Http2Stream$StreamTimeout.newTimeoutException(Http2Stream.kt:675)
   ...
```

## Notes

- Two layers contribute:
  - `HttpClientConfigurator` wraps any `IOException` into `NetworkException("No internet connection")`, collapsing distinct transport failures (timeout, DNS, connection refused) into one misleading message.
  - `KeygenPeerDiscoveryViewModel.startSessionWithRetry()` (and `next()`) always fall back to the generic `error_view_default_*` strings, discarding the actual cause.

App version: 1.0.108 (108)

aminsato:
## Summary

Closes #4956.

When starting a keygen session failed due to a network/server timeout, the UI showed the generic **"Something went wrong / An error occurred"** screen and the failure was logged as `NetworkException: No internet connection` — even though the device had connectivity and the real cause was a `SocketTimeoutException` on `router/start`. Two layers contributed, and both are fixed here.

### Layer 1 — transport-failure classification (`data`)
- `HttpClientConfigurator` no longer wraps every `IOException` as `NetworkException("No internet connection")`. It classifies the subtype:
  - `SocketTimeoutException` → **Timeout** ("Connection timed out")
  - `UnknownHostException` → **NoConnectivity** ("No internet connection")
  - everything else → **Transport** ("Network request failed")
- New `NetworkErrorKind` enum + `kind` field on `NetworkException` (with a back-compat constructor, so existing call sites are untouched).
- New `Throwable.networkErrorKind()` walks the cause chain so the UI can pick a cause-specific message.

### Layer 2 — keygen error surfacing (`app`)
- `startSessionWithRetry()`, `startVultiServerConnection()` and `next()` in `KeygenPeerDiscoveryViewModel` now map the actual failure to a cause-specific `ErrorUiModel`. A timeout or connectivity loss tells the user to check their connection and retry; anything that isn't a recognisable transport failure keeps the generic copy.
- New `keygen_error_timeout_*` and `keygen_error_network_*` strings across **all 10 locales**.

## Test plan
- `./gradlew :data:testDebugUnitTest` — extended `HttpClientConfiguratorTest` (timeout/DNS/generic kinds), updated `NetworkStateInterceptorContractTest` to assert `kind`, updated `MockHttpClient` to mirror prod classification, added `NetworkErrorKindTest` (cause-chain walking). ✅
- `./gradlew :app:testDebugUnitTest --tests "*.peer.*"` — keygen peer-discovery suite green. ✅
- `./gradlew :app:compileDebugKotlin` — ✅
- ktfmt applied; all 10 `strings.xml` validated as well-formed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved key generation error handling by classifying failures as connection timeout, no connectivity, or other network issues, and showing clearer user-facing warnings/errors.
* **Localization**
  * Added localized Keygen network error message strings (timeout and general connectivity problems) for German, Spanish, Croatian, Italian, Korean, Dutch, Portuguese, Russian, and Chinese.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -39,8 +39,8 @@ import com.vultisig.wallet.data.usecases.QrShareInfo
 import com.vultisig.wallet.data.usecases.tss.DiscoverParticipantsUseCase
 import com.vultisig.wallet.data.usecases.tss.ParticipantName
 import com.vultisig.wallet.data.utils.NetworkErrorKind
-import com.vultisig.wallet.data.utils.networkErrorKind
 import com.vultisig.wallet.data.utils.safeLaunch
+import com.vultisig.wallet.data.utils.toNetworkErrorKind
 import com.vultisig.wallet.ui.components.errors.ErrorUiModel
 import com.vultisig.wallet.ui.navigation.Destination
 import com.vultisig.wallet.ui.navigation.NavigationOptions
@@ -237,18 +237,18 @@ constructor(
      * copy.
      */
     private fun Throwable.toKeygenErrorUiModel(): ErrorUiModel =
-        when (networkErrorKind()) {
+        when (toNetworkErrorKind()) {
             NetworkErrorKind.Timeout ->
                 ErrorUiModel(
                     title = UiText.StringResource(R.string.keygen_error_timeout_title),
                     description = UiText.StringResource(R.string.keygen_error_timeout_description),
                 )
-            NetworkErrorKind.NoConnectivity,
-            NetworkErrorKind.Transport ->
+            NetworkErrorKind.NoConnectivity ->
                 ErrorUiModel(
                     title = UiText.StringResource(R.string.keygen_error_network_title),
                     description = UiText.StringResource(R.string.keygen_error_network_description),
                 )
+            NetworkErrorKind.Transport,
             NetworkErrorKind.Http,
             null ->
                 ErrorUiModel(
--- a/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
@@ -38,6 +38,8 @@ import com.vultisig.wallet.data.usecases.QrShareField
 import com.vultisig.wallet.data.usecases.QrShareInfo
 import com.vultisig.wallet.data.usecases.tss.DiscoverParticipantsUseCase
 import com.vultisig.wallet.data.usecases.tss.ParticipantName
+import com.vultisig.wallet.data.utils.NetworkErrorKind
+import com.vultisig.wallet.data.utils.networkErrorKind
 import com.vultisig.wallet.data.utils.safeLaunch
 import com.vultisig.wallet.ui.components.errors.ErrorUiModel
 import com.vultisig.wallet.ui.navigation.Destination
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -38,6 +38,8 @@ import com.vultisig.wallet.data.usecases.QrShareField
 import com.vultisig.wallet.data.usecases.QrShareInfo
 import com.vultisig.wallet.data.usecases.tss.DiscoverParticipantsUseCase
 import com.vultisig.wallet.data.usecases.tss.ParticipantName
+import com.vultisig.wallet.data.utils.NetworkErrorKind
```
johnnyluo, line 41: these two are the same , except one is lowercase , Can you please rename `networkErrorKind` to `ToNetworkErrorKind`

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
```diff
@@ -224,6 +226,34 @@ constructor(
...
+    private fun Throwable.toKeygenErrorUiModel(): ErrorUiModel =
+        when (networkErrorKind()) {
+            NetworkErrorKind.Timeout ->
+                ErrorUiModel(
+                    title = UiText.StringResource(R.string.keygen_error_timeout_title),
+                    description = UiText.StringResource(R.string.keygen_error_timeout_description),
+                )
+            NetworkErrorKind.NoConnectivity,
+            NetworkErrorKind.Transport ->
+                ErrorUiModel(
```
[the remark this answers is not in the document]
aminsato: Done in 3b69d1aa0 — `NetworkErrorKind.Transport` now falls through to the generic copy alongside `Http`/`null`, so `keygen_error_network_*` is reserved for `NoConnectivity`.

app/src/main/java/com/vultisig/wallet/ui/models/peer/KeygenPeerDiscoveryViewModel.kt
aminsato: Done in 3b69d1aa0 — renamed `Throwable.networkErrorKind()` to `toNetworkErrorKind()` (def, call site, import, and tests) to avoid the case-only collision with the `NetworkErrorKind` enum.

johnnyluo (approved):
LGTM

PR #4969, opened 2026-06-18
feat(swap): Advanced Swap — Market/Limit tabs (phase 0 of #4858)

### LINKED ISSUE #3389: Missing Feature: Advanced Swap Settings (Slippage, Gas Limit, External Recipient)
## Missing Feature
**Figma Link**: https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=67186-146846

### Description
The Figma design for the swap flow includes an **Advanced Settings** bottom sheet accessible from the swap screen. This bottom sheet provides three configurable options:

1. **Slippage Tolerance** — with options: Auto, 0.5%, 1%, 3%, Custom
2. **Gas Limit** — configurable gas limit with Auto default
3. **Use External Recipient** — toggle to enable sending swap output to a different address (with address input field, paste/scan/contacts buttons)

The swap screen in Figma shows an "Advanced Settings" text link at position y=680 (below the swap component, above the button).

### Current State
The Android implementation in `SwapScreen.kt` has no Advanced Settings link or bottom sheet. There is no slippage tolerance selector, gas limit override, or external recipient option in the swap flow.

### Expected Behavior
- Add an "Advanced Settings" tappable text below the fee details section
- Implement a bottom sheet with the three settings (Slippage Tolerance, Gas Limit, External Recipient)
- Slippage Tolerance sub-sheet with radio-style selection (Auto, 0.5%, 1%, 3%, Custom with text input)
- Gas Limit sub-sheet with numeric input
- External Recipient sub-sheet with address input + paste/scan/contacts action bar

### LINKED ISSUE #4858: Feature: Advanced Swap — slippage, gas limit, external recipient (Figma)
## Summary

Implement the **Advanced Swap** flow per Figma → [Swap - Advanced Swap](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=67186-146846)

The swap screen gains a mode tab row and an "Advanced Settings" entry that opens a sheet with per-swap overrides: slippage tolerance, gas limit, and an external recipient.

## Scope (as shown in the design)

**1. Swap mode tabs**
- `Market` / `Limit` tab row above the swap component (`NFTs – Soon` exists in the design but is hidden). This ticket covers the tab row UI with Market active; Limit-order execution is not specced in this node and is out of scope here.

**2. Advanced Settings entry + sheet**
- "Advanced Settings" link above the Swap button opens the **Advanced swap** sheet with three rows, each showing its current value:
  - Slippage Tolerance — `Auto`
  - Gas Limit — `Auto`
  - Use External Recipient — `Off`

**3. Slippage sheet**
- Options: `Auto` / `0.5%` / `1%` / `3%` / `Custom` (numeric input).
- Helper copy: *"Limits how much the price can change before execution is canceled."*

**4. Gas Limit sheet**
- Manual gas limit numeric input with `Auto` as default. Only applicable on chains where a custom gas limit makes sense (EVM); hide or disable elsewhere.

**5. Use External Recipient**
- "Send to Different Address" input with paste / QR scan / address book actions.
- ⚠️ The external recipient MUST be surfaced on the verify/keysign screen — the user has to see the destination address before signing, never a silent default.

**6. Expanded fee breakdown**
- Total Fees row expands to show Network Fee (gas price + fiat), Swap Fee (incl. rate), and Applied Discounts (VULT tier), plus Price Impact with a quality rating (Good / Average), matching the "Expanded" frames in the design.

## Acceptance criteria

- [ ] Tab row renders with Market active; Limit visible per design
- [ ] Advanced Settings sheet with the three rows and current values
- [ ] Slippage selection (presets + custom) applied to the swap quote/tx
- [ ] Custom gas limit applied on EVM swaps; Auto default preserved
- [ ] External recipient flow works via paste / QR / address book, and the address is shown on the verify screen before signing
- [ ] Settings reset sensibly between swaps (no sticky custom slippage surprises)

aminsato:
## Summary

Implements the **Advanced Swap** flow (#4858) — a `Market | Limit` mode tab row and an **Advanced Settings** sheet with per-swap overrides: slippage tolerance, EVM gas limit, and an external recipient. Built to parity with the full Windows feature graph (UI vultisig/vultisig-windows#4113/#4114/#4117/#4118/#4119, SDK vultisig/vultisig-sdk#757, wiring vultisig/vultisig-windows#4134, verify-surfacing #4152) — entirely in-repo, since Android has no separate swap SDK.

### Transaction hash
https://etherscan.io/tx/0xaef02a315e1e9290133c5ce29b06102ed16123816a1fca61ce2bc0063127c59f

Structured commits:

- [x] **Phase 0 — Market/Limit tab row.**
- [x] **Phase 1 — Advanced Settings link + sheet** (single sheet that pages between menu and each selector).
- [x] **Phase 2 — Slippage tolerance.** `slippageBps` (null = Auto) → THORChain/Maya `tolerance_bps`, 1inch `slippage`, Kyber `slippageTolerance`, **LI.FI** `slippage` (fraction). Part of the quote cache key; a change re-fetches.
- [x] **Phase 3 — EVM gas-limit override.** Replaces both `tx.gas` and `ethereumSpecific.gasLimit` (`OneInchSwap` signs `maxOf(...)`); EVM-only, applied at build time.
- [x] **Phase 4 — External recipient + verify surfacing.** Honoured **only on native THORChain/Maya** (memo `destination`); every general aggregator (1inch/Kyber/Jupiter/LI.FI/SwapKit) is **skip-guarded** when a recipient is set — matching the cross-platform decision (SDK #757 / #4152). Stamped on the tx and shown as a **warning-styled Recipient row** on the verify screen so the destination is never a silent default.

## Parity notes (deep cross-check)

- **External recipient is native-only** (THORChain/Maya), exactly as Windows/iOS/SDK ship it; routing it through aggregators is a deferred SDK follow-up there too.
- **SwapKit slippage:** Windows' SDK forwards a percent to SwapKit; Android's SwapKit integration uses a server-side slippage floor with no quote-time param, so it is a no-op here (integration difference, not a regression).
- **Expanded fee breakdown / price-impact rating (epic item 6):** not shipped by the Windows PRs either; the Good/Average metric does not exist in this codebase, so it is intentionally out of scope.
- **Deferred (as on Windows):** "Select route" sheet, VULT-tier gating, external-recipient QR-scan/address-book entry (paste + manual ship now), joiner-side recipient surfacing.

## Validation

- ktfmt clean; `:data` + `:app` swap unit suites green (slippage cache-key, Kyber default+override, gas-limit override, external-recipient stamping).

## Device verification

Verified end-to-end on a mainnet device — both the **gas-limit** and **slippage** overrides from the Advanced sheet are baked into the signed/executed transaction ([etherscan tx `0xaef0…c59f`](https://etherscan.io/tx/0xaef02a315e1e9290133c5ce29b06102ed16123816a1fca61ce2bc0063127c59f)).

**Slippage proof — overrides match the screenshots and the on-chain calldata.** Decoding the KyberSwap `swap` calldata (selector `0xe21fd0e9`) of the tx above:

| Field | Advanced sheet / quote (screenshot) | On-chain calldata | Match |
|---|---|---|---|
| Gas limit | `501000` | `gas = 501,000` | ✅ |
| Amount in | `0.0022308 ETH` | `value = 0.0022308 ETH` | ✅ |
| Expected out (quote) | `3.787248 USDC` | embedded meta `AmountOut = 3787248` | ✅ |
| Slippage tolerance | `3%` | `minReturnAmount = 3673630` (`0x380e1e`) | ✅ |

The minimum-output amount the contract enforces is exactly `floor(3787248 × (1 − 0.03)) = 3673630` → **a precise 3% floor**, i.e. the 3% slippage selected in the sheet is what protects the swap on-chain. The calldata also embeds `"Source":"vultisig-android"`, confirming the tx originated from this app.

## Issues fixed

Fixes #4858

This PR also fully resolves #3389 (*Missing Feature: Advanced Swap Settings — Slippage, Gas Limit, External Recipient*). #3389 and #4858 describe the **same feature** against the same Figma node (`67186-146846`) — the Advanced Settings sheet with slippage tolerance, gas limit, and external recipient — so the same implementation closes both.

Fixes #3389

## Release Notes

* **New Features**
  * Added swap mode selection (Market/Limit); Limit mode shows "coming soon" placeholder
  * Added Advanced Settings sheet for Market mode with controls for slippage tolerance, gas limit, and external recipient address
  * External recipient now displays on swap verification screen when configured

* **Bug Fixes**
  * Improved "amount too low" error message to clarify fees must be covered

* **Localization**
  * Updated error messages and added new UI strings in German, Spanish, Croatian, Italian, Korean, Dutch, Portuguese, Russian, and Simplified Chinese


johnnyluo:
@aminsato based on figma , the button for advanced setting should be here
CleanShot 2026-06-20 at 11 46 47@2x

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
```diff
@@ -273,12 +294,13 @@ constructor(
                     dstAddress = dstAddress,
                     approveSpender = approveSpender,
                     expectedDstTokenValue = dstTokenValue,
-                    blockChainSpecific = specificAndUtxo,
+                    blockChainSpecific = effectiveSpecificAndUtxo,
                     estimatedFees = quote.fees,
                     gasFees = estimatedNetworkFeeTokenValue ?: gasFee,
                     memo = null,
                     isApprovalRequired = isApprovalRequired,
                     gasFeeFiatValue = gasFeeFiatValue,
```
johnnyluo: **Gas-limit override is not reflected in the displayed / verify-screen network fee.**

The override correctly replaces `tx.gas` and `ethereumSpecific.gasLimit` above, but `gasFeeFiatValue` (and `gasFees`) are still passed straight from the aggregator estimate — they were computed against the *original* estimated gas limit, not the override. So a user who sets gas limit `501000` sees a fee preview derived from the old estimate on both the swap form and the verify screen.

- Raising the limit: actual cost is mostly unaffected (you pay gas *used*, not the limit), so the preview is only slightly off.
- Lowering the limit below what the swap needs: the tx can revert out-of-gas, with a fee preview that still looks fine and **no warning**.

Consider recomputing the displayed max-fee from the override (gasLimit × gasPrice), or surfacing a warning when the override is below the estimated limit, so the verify screen doesn't show a fee that can't match what executes.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
```diff
@@ -169,7 +171,21 @@ internal class SwapQuotePipeline(
...
+            // aggregator (1inch / Kyber / Jupiter / LI.FI / SwapKit) is dropped when a recipient is
+            // set, never silently misrouting funds. This mirrors the cross-platform decision
+            // (vultisig-sdk#757 / vultisig-windows#4152): threading a recipient through the
+            // aggregators is a deferred follow-up there too.
+            val eligibleProviders =
+                if (externalRecipient.isNullOrBlank()) {
+                    allEligibleProviders
+                } else {
+                    allEligibleProviders.filter {
+                        it == SwapProvider.THORCHAIN || it == SwapProvider.MAYA
```
johnnyluo: **SwapKit should be kept here too when an external recipient is set; only the EVM aggregators (1inch / Kyber / Jupiter / LI.FI) should be dropped.**

SwapKit takes the recipient as its quote-time `dstAddress`, so it can route the output to a custom address just like THORChain/Maya — it doesn't bake the vault address into calldata the way the EVM aggregators do. As written, a recipient narrows eligibility to THORChain/Maya only, needlessly excluding valid SwapKit routes (e.g. BTC-leg cross-chain swaps).

To support it:
1. Add `SwapProvider.SWAPKIT` to this filter.
2. Thread `externalRecipient` into `fetchSwapKitQuote` and use it for both the cache-key `dstAddress` and the request `dstAddress` (mirroring the `effectiveDst` / `cacheDstAddress` / `requestDstAddress` handling in the THOR/Maya path).

The `SwapTransactionBuilder` already stamps `externalRecipient` on the SwapKit path, so the verify-screen surfacing is already in place. Also note the comment here (and on the EVM aggregator builder branch) currently lists SwapKit among the "dropped" aggregators — that should be corrected once SwapKit is honoured.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/AdvancedSwapSettings.kt
```diff
@@ -0,0 +1,436 @@
...
+    externalRecipient: String?,
+    errorText: String?,
+    onSelect: (String?) -> Unit,
+) {
+    val state = rememberTextFieldState(externalRecipient.orEmpty())
+
+    LaunchedEffect(Unit) {
+        snapshotFlow { state.text.toString() }
+            .distinctUntilChanged()
+            .collect { text -> onSelect(text.trim().takeIf { it.isNotEmpty() }) }
```
johnnyluo, line 230: **Quote re-fetches on every recipient keystroke can hit THOR/Maya with intermediate/invalid addresses.**

Each edit flows `onSelect` → ViewModel `externalRecipient` StateFlow → the pipeline's `combine(externalRecipient)`, so partially-typed (and currently-invalid) addresses trigger quote requests. The `isValid` gate only blocks at `swap()` time / inline display, not the fetch — so the native quote endpoints get called with malformed destinations mid-typing, which can flash transient quote errors.

The 300ms pipeline debounce + `distinctUntilChanged` mitigate the volume, but consider gating the re-quote on a *valid* address (only push the recipient downstream once it passes `isValid`), so intermediate keystrokes don't fire native quote calls.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
aminsato: Fixed in 60c4fc0dd. The OneInch branch now recomputes the displayed/verify fee from the override: `overriddenFeeWei = gasLimit × maxFeePerGasWei`, re-valued in fiat via the native-token price implied by the matched `(gasFee, gasFeeFiatValue)` baseline pair (fiat ÷ token, which is gas-independent — no price oracle needed). Auto (no override) keeps the aggregator estimate untouched. Added a unit test covering it (300_000 × 10 = 3_000_000 wei → $9.00).

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
aminsato: Done in d179f0c28. SwapKit is now kept in the recipient filter alongside THORChain/Maya, and `externalRecipient` is threaded into `fetchSwapKitQuote` for both the cache key (`cacheDstAddress`) and the request `dstAddress` (`requestDstAddress`), mirroring the `effectiveDst` handling on the THOR/Maya path. Only the EVM aggregators (1inch / Kyber / Jupiter / LI.FI) are dropped when a recipient is set now, and the pipeline comment was corrected to reflect that.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/AdvancedSwapSettings.kt
aminsato: Fixed in ed8e0267b. The ViewModel now feeds the quote pipeline from a separate `quoteRecipient` flow that only receives a present-and-valid address; an invalid or partially-typed value routes quotes to the vault (null) instead of being pushed downstream, so the native quote endpoints are never called with a malformed destination. The typed value still drives the field display and the `swap()` pre-flight gate. Added VM tests asserting an invalid recipient isn't pushed into the pipeline while a valid one is.

aminsato:
@coderabbitai full reiview

app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/AdvancedSwapSettings.kt
```diff
@@ -0,0 +1,476 @@
...
+
+    // Parse percent → bps on each edit and hoist it up; rapid typing coalesces in the quote
+    // pipeline's debounce. Blank/invalid input leaves the selection untouched.
+    LaunchedEffect(Unit) {
+        snapshotFlow { customState.text.toString() }
+            .distinctUntilChanged()
+            .collect { text ->
+                // Normalize comma decimal separators ("0,5") to dot before parsing — locales that
+                // use a comma would otherwise fail toBigDecimalOrNull() and never apply slippage.
+                val percent = text.trim().replace(',', '.').toBigDecimalOrNull() ?: return@collect
```
rkokhatskyi: Emptying the custom slippage field never resets to Auto because `toBigDecimalOrNull() ?: return@collect` skips `onSelect(null)`, so the last entered tolerance keeps driving every quote and the signed swap while the field renders blank, silently loosening slippage protection

app/src/main/java/com/vultisig/wallet/ui/screens/swap/VerifySwapScreen.kt
```diff
@@ -266,6 +266,16 @@ private fun VerifySwapScreen(
                         metadata = tx.src.token.address,
                     )
 
+                    // External recipient must be visible before signing — never a silent default
+                    // (#4858). Only shown when the user routed the output to a custom address, and
+                    // given warning emphasis so it reads as a deliberate deviation, not a fee row.
+                    tx.externalRecipient
```
rkokhatskyi, line 272: This recipient warning only renders for the initiator since the cosigner's `JoinSwapUiModelBuilder.kt:482` builds `SwapTransactionUiModel` without `externalRecipient`, so a cosigning peer blind signs a THOR/Maya memo routing output to a non vault address its verify screen never shows

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -113,15 +133,71 @@ constructor(
...
 
+    /**
+     * Tracks whether a custom gas limit applies to the selected source chain (EVM only) and clears
+     * a stale override when switching to a non-EVM source, so it can never carry over to a chain
+     * that ignores it (#4858).
+     */
+    private fun observeGasLimitApplicability() {
+        viewModelScope.launch {
+            selectedSrc.collect { src ->
+                val applicable = src?.account?.token?.chain?.standard == TokenStandard.EVM
```
rkokhatskyi: `isGasLimitApplicable` gates only on `src.chain.standard == TokenStandard.EVM`, yet the override is consumed solely in the `SwapQuote.OneInch` builder branch, so an EVM token swapped via THORChain/Maya shows an enabled Gas Limit row and accepts a value the signed transaction silently ignores

data/src/main/kotlin/com/vultisig/wallet/data/api/LiFiChainApi.kt
```diff
@@ -87,6 +89,8 @@ constructor(
                 parameter("fromAmount", fromAmount)
                 parameter("fromAddress", fromAddress)
                 parameter("toAddress", toAddress)
+                // LI.FI takes slippage as a fraction (0.01 = 1%); omitted = LI.FI's own default.
+                slippageBps?.let { parameter("slippage", it / 10_000.0) }
```
rkokhatskyi: `it / 10_000.0` is stringified by `Double.toString()`, which emits scientific notation like `1.0E-4` for 1 to 9 bps custom slippage, so LI.FI receives a non numeric `slippage` and the tightest tolerances are silently dropped
```kotlin
slippageBps?.let { parameter("slippage", (it / 10_000.0).toBigDecimal().toPlainString()) }
```

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -324,6 +402,7 @@ constructor(
             srcToken.address,
             dstToken.address,
             currentAmount,
+            slippageBps.value,
```
rkokhatskyi, line 433: `cacheQuote` keys this flip snapshot on `dstToken.address` and omits `externalRecipient`, while the THOR/Maya fetch keys on `effectiveDst` (`SwapQuoteManager.kt:610`), so a set recipient, flip, clear sequence lets a later no recipient `get` serve the recipient routed quote whose memo still pays the cleared address

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -2695,6 +2707,138 @@ internal class SwapFormViewModelTest {
...
+    fun `external recipient on a pair with no native route surfaces a recipient-aware unsupported error`() =
+        runTest(mainDispatcher) {
+            // Setting a recipient drops the aggregators and keeps only THORChain/Maya. When the
+            // pair
+            // has no native route at all, the bare "not supported" must instead name the recipient
+            // as the reason (#4858).
+            every { swapQuoteRepository.getEligibleProviders(any(), any()) } returns
+                listOf(SwapProvider.THORCHAIN)
+            coEvery {
+                swapQuoteManager.fetchBestQuote(
```
rkokhatskyi, line 2812: This stubs `fetchBestQuote` with nine `any()` while the real method takes eleven params, so once a non null `externalRecipient` is set MockK fails to match and the spy runs the real method, leaving `recipientAwareError` unexercised by both recipient tests that name it

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -287,6 +291,12 @@ constructor(
                     showIndicativeRate(input)
                 }
                 .combine(refreshQuoteState) { input, _ -> input }
+                // A slippage change re-fetches the quote silently (like the refresh timer), so the
+                // new tolerance is applied without a spinner flash (#4858).
+                .combine(slippageBps) { input, _ -> input }
+                // An external-recipient change likewise re-fetches so the swap re-routes to the new
+                // address (#4858).
+                .combine(externalRecipient) { input, _ -> input }
```
rkokhatskyi: `combine(externalRecipient)` and `combine(slippageBps)` sit after the `onEach { isLoading = true }` block, so these changes re-fetch without raising `isLoading`, leaving the Swap button enabled over the staged prior quote, so a stale quote whose routing or tolerance differs from the displayed value can be signed

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/AdvancedSwapSettings.kt
aminsato: Fixed in 8fbfa2a0b. Emptying the custom field now resets to Auto: a blank input calls onSelect(null) instead of returning early, while a `hasCustomValue` guard keeps the empty seed shown when a preset is active from clobbering that preset.

data/src/main/kotlin/com/vultisig/wallet/data/api/LiFiChainApi.kt
aminsato: Fixed in 8fbfa2a0b. Slippage is now formatted with `BigDecimal(bps).movePointLeft(4).stripTrailingZeros().toPlainString()`, so 1–9 bps render as plain decimals (e.g. `0.0001`) rather than scientific notation. Added LiFiChainApiSlippageTest pinning the query param.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in 00df154c5. `isGasLimitApplicable` now also requires the resolved route to honor the override (only SwapQuote.OneInch does). The pipeline exposes a tri-state `honorsGasLimitOverride` flow combined with the EVM-source check, so a THOR/Maya route disables the row; before a quote resolves it stays applicable for an EVM source.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in 00df154c5. `cacheCurrentQuote` now keys on the same effective destination the fetch path uses (recipient when set, else the vault address), so a set/flip/clear sequence can't serve the recipient-routed quote to a no-recipient lookup.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Fixed in 00df154c5. Both recipient-test stubs now use 11 `any()` matchers, so MockK matches once a non-null recipient is set instead of defaulting the trailing params to null and running the real method.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in 00df154c5. The slippage and external-recipient `onEach { isLoading = true }` now ride each flow individually before the combine, so those re-fetches disable the Swap button until the new quote lands, while the silent refresh timer still doesn't flash the spinner.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/VerifySwapScreen.kt
aminsato: Fixed in cdc202eda. The cosigner path (JoinSwapUiModelBuilder) now recovers the recipient from the THOR/Maya swap memo's destination and surfaces it on the verify screen when it differs from the vault's own destination address. The keysign proto is in the do-not-edit commondata submodule so it can't carry an explicit field; EVM aggregators bake the vault address into calldata (recipient dropped) and aren't affected.

rkokhatskyi (approved):
LGTM

rkokhatskyi (approved):
LGTM

PR #4977, opened 2026-06-20
fix(swap): guard null "To" token balance while loading

### LINKED ISSUE #4963: Swap: "To" token shows "null GRT" balance while balance is loading
## Summary

On the **Swap** screen, the destination ("To") token displays the literal text **`null GRT`** as its balance for a few seconds after the destination token is selected (while the balance is still loading). It only resolves once the balance finishes loading.

## Steps to Reproduce

1. Open **Swap**.
2. Set **From** to Ethereum → ETH.
3. Set **To** to Ethereum → GRT (any token whose balance isn't cached yet).
4. Observe the top-right balance label of the "To" box.

## Actual Behavior

The "To" balance label reads `null GRT` (literal string "null" + ticker) until the balance loads.

## Expected Behavior

While the balance is unknown/loading, show nothing (empty), a `0`, or a loading skeleton — never the literal text `null`.

## Notes (verified root cause)

`TokenBalanceUiModel.balance` is nullable (`String?`) and is `null` until the token value is loaded (`AccountToTokenBalanceUiModelMapper` sets `balance = tokenValue?.let(...)`).

In `TokenInput.kt` the balance is rendered as:

```kotlin
text = selectedToken?.let { "${it.balance} ${it.title}" } ?: ""
```

The `?.let` only guards against `selectedToken` being null — not `balance`. When `balance` is null, Kotlin string interpolation emits the literal `"null"`, producing `null GRT`.

## Screenshot

null GRT on Swap screen

(Also reproduced in a screen recording; the label shows `null GRT` for a few seconds before resolving.)

johnnyluo:
## Summary
- The Swap screen's "To" token balance label showed the literal text `null <ticker>` (e.g. `null GRT`) for a few seconds after selecting a destination token, until the balance finished loading.
- Root cause: in `TokenInput.kt` the label was built as `selectedToken?.let { "${it.balance} ${it.title}" }`. The `?.let` only guarded `selectedToken` — not its nullable `balance` (`String?`, null until loaded). Kotlin string interpolation then emitted the literal `"null"`.
- Fix: guard `balance` as well, so the label stays empty while the balance is null/loading and only renders `"<balance> <ticker>"` once it resolves.

## Issue
Closes #4963

## Test Plan
- [x] Debug Kotlin compiles (`:app:compileDebugKotlin`)
- [ ] Manual: Swap → From ETH → To GRT (uncached balance); confirm the "To" balance label is empty while loading and never shows `null`, then resolves to the real balance.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token balance display in token input to properly handle cases where balance information is unavailable.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/TokenInput.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/TokenInput.kt
@@ -210,7 +210,10 @@ internal fun TokenInput(
             }
 
             Text(
-                text = selectedToken?.let { "${it.balance} ${it.title}" } ?: "",
+                text =
+                    selectedToken?.let { token ->
+                        token.balance?.let { balance -> "$balance ${token.title}" }
+                    } ?: "",
                 style = Theme.brockmann.supplementary.caption,
                 color = Theme.v2.colors.text.tertiary,
                 textAlign = TextAlign.End,


rkokhatskyi (approved):
LGTM

PR #4982, opened 2026-06-21
fix(defi): show selected currency symbol in DeFi tabs, not device locale

### LINKED ISSUE #4967: [BUG] THORChain defi tab displaying amount in wrong currency symbol
# [Fix] [BUG] THORChain defi tab displaying amount in wrong currency symbol

## Problem
The vault currency setting is set to USD, but the THORChain defi tab is displaying value in £, not $.

## Expected Behavior
N/A

## Steps to Reproduce
1. See description above.

### Must Do
- [ ] Implement the fix/feature described above

### Must NOT Do
- Don't change unrelated code

## Acceptance Criteria
- [ ] Behavior matches expected behavior above

## Context
- Reported by **Jatinkkalra** (@jatinkkalra on Discord) on Discord

cc @realpaaao

aminsato:
Fixes #4967

## Problem

The vault currency was set to USD, but the THORChain DeFi tab displayed amounts in £ instead of $.

## Root cause

`AppCurrencyRepositoryImpl.getCurrencyFormat()` built its `NumberFormat` from the **device locale only**:

```kotlin
cachedCurrencyFormat = NumberFormat.getCurrencyInstance(currentLocale)
```

So the currency **symbol** followed the device locale's default currency, ignoring the user's selected vault currency. On a UK device (`en_GB`) that default is GBP → £. The numeric values were already correct (prices are fetched in the selected currency); only the symbol was wrong.

The correct path elsewhere (`FiatValueToStringMapperImpl`) compensates by cloning the format and setting `currency = Currency.getInstance(...)`. But the DeFi tab ViewModels (THORChain, Maya, Tron, Circle, Cosmos) call `getCurrencyFormat()` → `format()` directly, so they all inherited the wrong symbol. THORChain was the one reported.

## Fix

Apply the selected `AppCurrency` to the format at the root, so every direct caller is correct:

- Set `currency = Currency.getInstance(appCurrency.ticker)` on the format.
- Key the cache on `(locale, currency)` so it rebuilds when the user changes currency.
- Return a clone so concurrent callers never share a mutable `NumberFormat` (thread safety).
- `FiatValueToStringMapperImpl` is unaffected — it still clones and overrides the currency.

One change fixes the reported THORChain bug **and** the identical latent defect in the Maya/Tron/Circle/Cosmos DeFi tabs.

## Test plan

- [x] New `AppCurrencyRepositoryImplTest` (3 cases, all green):
  - UK locale + USD selected → uses `$`, never `£` (the #4967 regression)
  - US locale + EUR selected → `€1,000.00`
  - currency change between calls is reflected
- [x] ktfmt applied

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed currency formatting to match the currently selected app currency, including correct symbols and fraction digits across different device locales.

- **Tests**
  - Added multi-locale and multi-currency formatting tests to ensure output updates when the selected ticker changes and that zero-decimal locales still render the selected currency’s expected decimals.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/AppCurrencyRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/AppCurrencyRepository.kt
@@ -4,9 +4,11 @@ import androidx.datastore.preferences.core.stringPreferencesKey
 import com.vultisig.wallet.data.models.settings.AppCurrency
 import com.vultisig.wallet.data.sources.AppDataStore
 import java.text.NumberFormat
+import java.util.Currency
 import java.util.Locale
 import javax.inject.Inject
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
@@ -31,6 +33,7 @@ internal class AppCurrencyRepositoryImpl @Inject constructor(private val dataSto
 
     private val mutex = Mutex()
     private var cachedLocale: Locale? = null
+    private var cachedCurrency: AppCurrency? = null
     private var cachedCurrencyFormat: NumberFormat? = null
 
     override val currency: Flow<AppCurrency>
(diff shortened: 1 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/AppCurrencyRepository.kt
```diff
@@ -49,18 +52,31 @@ internal class AppCurrencyRepositoryImpl @Inject constructor(private val dataSto
...
+                cachedLocale != currentLocale ||
+                    cachedCurrency != appCurrency ||
+                    cachedCurrencyFormat == null
+            ) {
                 cachedLocale = currentLocale
-                cachedCurrencyFormat = NumberFormat.getCurrencyInstance(currentLocale)
+                cachedCurrency = appCurrency
+                cachedCurrencyFormat =
+                    NumberFormat.getCurrencyInstance(currentLocale).apply {
+                        currency = Currency.getInstance(appCurrency.ticker)
```
rkokhatskyi: `currency = Currency.getInstance(appCurrency.ticker)` sets the symbol but not the fraction digits, which stay at the device locale default currency, so a USD vault on a `ja_JP`/`ko_KR` device renders `$1,235` and drops the cents that iOS keeps by pinning fraction digits
```kotlin
NumberFormat.getCurrencyInstance(currentLocale).apply {
    currency = Currency.getInstance(appCurrency.ticker)
    minimumFractionDigits = currency.defaultFractionDigits
    maximumFractionDigits = currency.defaultFractionDigits
}
```

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AppCurrencyRepository.kt
aminsato: Good catch — fixed in 285bdf17f. Now pinning `minimumFractionDigits`/`maximumFractionDigits` to `selectedCurrency.defaultFractionDigits` so a USD vault keeps its cents on a JPY/KRW (0-decimal) device locale. Added a `ja_JP` test asserting `$1,234.50`.

rkokhatskyi (approved):
LGTM

PR #4988, opened 2026-06-21
feat(swap): show external recipient on swap verify screen (#4972)

### LINKED ISSUE #4972: [Add] Show external recipient address on swap verify screen [Swap/Keysign Verify]
# [Add] Show external recipient address on swap verify screen [Swap/Keysign Verify]

## Problem
The Advanced Swap feature lets a user route swap output to an **external recipient** (an address other than their own vault). That override is honored in the quote/keysign payload (THORChain/MayaChain memo `destination`, CowSwap order `receiver`), but the Android swap/keysign verify screen never displays it.

This is most visible in a **cross-device** flow: when the **extension** initiates a swap with `Use External Recipient` set to a different address and **Android joins the session to co-sign**, Android shows **no on-screen confirmation of where the funds are going** — the external recipient address is visible on the extension but **not** on Android. The signer on Android approves a transaction that sends funds to a different address without seeing it. This is the most security-sensitive of the advanced overrides.

This was already implemented on Windows/extension — see vultisig/vultisig-windows#4152 (and PR vultisig/vultisig-windows#4154). Android needs the same behavior for parity.

## Solution
On the swap/keysign verify screen, when the swap output is routed to an address **other than the current vault's own address for the destination chain**, render an emphasized **warning row** (warning treatment, e.g. amber, with an alert icon and text like "Sending to an external address") showing the recipient address.

Source the displayed address from the **built/received keysign payload** (the bytes that will actually be signed) — native `destination` (parsed from the THORChain/MayaChain memo) / CowSwap order `receiver` — rather than any live form state, so the screen reflects the real destination even when Android only joins to co-sign and never filled the form.

Default (own-address) swaps stay unchanged — no extra row.

### How it was done on Windows (reference)
- Resolve the recipient from the keysign payload: native → memo destination (3rd `:`-separated segment `SWAP:ASSET:DESTINATION:...`); CowSwap → decode the order and read `receiver` (guarded by `provider == 'cowswap'`).
- Compare the resolved recipient against the vault's own address on the destination chain. Comparison is **chain-aware**: case-insensitive for EVM (checksum vs lowercase — CowSwap lowercases the receiver), exact match for case-sensitive formats (Solana base58, UTXO, Cosmos bech32).
- Render the warning row only when they differ.

### Must Do
- [ ] Render a recipient warning row on the Android swap/keysign verify screen when the resolved destination differs from the vault's own address
- [ ] Source the displayed address from the keysign payload (native `destination` / CowSwap `receiver`), not live form state
- [ ] Visually distinguish it as an external-destination warning, not a neutral row
- [ ] Use a chain-aware address comparison (case-insensitive for EVM, exact otherwise)

### Must NOT Do
- Don't show the row for default (own-address) swaps
- Don't change the override wiring / quote logic

### Out of Scope
- Threading external recipient through 1inch/Kyber/LiFi/SwapKit aggregators (separate follow-up)
- Address-book / contact-name resolution for the recipient

## Acceptance Criteria
- [ ] A cross-device THORChain swap (e.g. BTC → USDT.eth) **initiated on the extension** with a custom recipient shows the recipient address on the Android verify screen before signing
- [ ] A normal swap to the user's own vault address shows **no** recipient row on Android
- [ ] The address displayed matches the `destination`/`receiver` in the signed keysign payload (parity with the extension)

## Screenshot (extension reference)

Image

_The extension already shows this warning box on the swap verify screen; Android should match it._

## References
- Windows issue: vultisig/vultisig-windows#4152
- Windows implementation PR: vultisig/vultisig-windows#4154

aminsato:
## Summary
Closes #4972. Shows the **external recipient address** on the swap/keysign verify screen when a swap's output is routed to an address other than the vault's own destination-chain address — parity with vultisig/vultisig-windows#4152.

The recipient is sourced from the **signed keysign payload** (the THORChain/MayaChain memo's `DESTINATION` segment), not live form state. This matters most in the cross-device co-sign flow: when the extension initiates a swap with a custom recipient and Android only **joins to co-sign**, Android previously showed no on-screen confirmation of where the funds were going. Now the signer sees it before approving.

## What changed
- **`SwapExternalRecipient.kt`** (new) — pure resolver: parses the memo destination and compares it chain-aware against the vault's own destination-chain address (case-insensitive for EVM, exact for Solana/UTXO/Cosmos). Returns `null` for default own-address swaps.
- **`VerifySwapViewModel`** — added `externalRecipient` to `SwapTransactionUiModel`; resolved on the initiator path.
- **`JoinSwapUiModelBuilder`** — resolved on the co-sign (join) path for THORChain/MayaChain swaps.
- **`VerifySwapScreen`** — new amber `ExternalRecipientWarning` row, shown only when a non-self recipient is resolved.
- **Strings** — `verify_swap_external_recipient_title` / `_description` across all 10 locales.
- **Tests** — `SwapExternalRecipientTest` (EVM casing, non-EVM exactness, streaming memo, non-swap/blank/own-address → null).

### Scope
Native THORChain/MayaChain memo destination only. CowSwap is not an active swap provider on Android (only a known-contract reference) and the EVM aggregators (1inch/Kyber/LiFi/SwapKit) are explicitly out of scope in the issue.

## Before / After

| Before | After |
|--------|-------|
| before | after |

Real Android renders (PreviewActivity + `adb screencap`). "Before" is a normal own-address swap (no row); "after" shows the external-recipient warning for the same swap.

## Test plan
- [x] `./gradlew :app:testDebugUnitTest --tests "*SwapExternalRecipientTest"` — green
- [x] `assembleDebug` compiles; rendered on device via PreviewActivity (screenshots above)
- [ ] Cross-device THORChain swap (BTC → USDT.eth) initiated on the extension with a custom recipient → recipient shows on Android verify before signing
- [ ] Normal swap to the vault's own address → no recipient row

## Summary by CodeRabbit

* **New Features**
  * Added support for external recipient addresses in swap transactions, allowing users to specify alternative destination addresses instead of their vault address.
  * Enhanced swap confirmation screen to display and validate external recipient information from transaction memos.

* **Tests**
  * Added comprehensive test coverage for external recipient address validation across swap scenarios.


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapExternalRecipient.kt
```diff
@@ -0,0 +1,54 @@
...
+ */
+internal fun parseThorchainMemoDestination(memo: String?): String? {
+    val trimmed = memo?.trim().orEmpty()
+    if (trimmed.isEmpty()) return null
+
+    val segments = trimmed.split(":")
+    if (segments.size < 3) return null
+
+    val verb = segments[0].uppercase()
+    val isSwap = verb == "SWAP" || verb == "=" || verb == "S"
```
rkokhatskyi: `parseThorchainMemoDestination` accepts only `SWAP`/`=`/`S`, so a THORChain `=<` limit swap or `m=<` modify limit memo returns null at `SwapExternalRecipient.kt:51` though its destination sits at the same `segments[2]` Android already reads. A cosigner then sees no external recipient row for those verbs, while the Windows `split(':')[2]` reference applies no verb gate

aminsato: Good catch — fixed. The verb gate now accepts the whole swap-verb family instead of the exact `SWAP`/`=`/`S` allowlist:

```kotlin
val isSwap = verb == "SWAP" || verb == "S" || verb.contains("=")
```

So `=` and its limit (`=<`) / modify-limit (`m=<`) variants all resolve the destination at `segments[2]`, while non-swap verbs (`+` add, `-` withdraw, bond, loan, …) stay rejected so a stray memo can't surface a phantom recipient. Added a test pinning `=<` / `m=<` / `M=<`.

I also reconciled with the Advanced Swap merge (#4969): the cosigner path (`JoinSwapUiModelBuilder.resolveExternalRecipient`) now delegates to this same `resolveExternalSwapRecipient` helper, so the fix flows to the join-keysign row you were worried about — and it picks up the chain-aware (case-sensitive for UTXO/Cosmos) comparison too.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/VerifySwapScreen.kt
```diff
@@ -300,6 +300,10 @@ private fun VerifySwapScreen(
                     )
                 }
 
+                tx.externalRecipient?.let { recipient ->
+                    ExternalRecipientWarning(address = recipient)
```
rkokhatskyi: `ExternalRecipientWarning` renders the same `tx.externalRecipient` that main already shows via `VerifyExternalRecipientRow` at `VerifySwapScreen.kt:272`, so a custom recipient now appears twice, the compact in-card row plus this amber card, while Windows renders it once. Consolidate to a single site and drop the redundant composable and `verify_swap_external_recipient_*` strings

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/screens/swap/VerifySwapScreen.kt
aminsato: Good catch — consolidated to a single site in e3fb8de83.

This was a collision with the Advanced Swap merge (#4969): when I opened this PR only the amber `ExternalRecipientWarning` existed, but #4969 independently added the in-card `VerifyExternalRecipientRow` backed by the same `tx.externalRecipient` field, so both rendered.

Dropped the `ExternalRecipientWarning` composable + its call site + the `verify_swap_external_recipient_*` strings (×10 locales), keeping the in-card row as the single render site — matching Windows. The memo-based resolution (`resolveExternalSwapRecipient`) and initiator/cosigner wiring are unchanged.

johnnyluo (approved):
LGTM

PR #5002, opened 2026-06-22
fix(thorchain): resolve native tx status via node when Midgard has no action

aminsato:
## Problem

Native THORChain/MayaChain transactions stay stuck **"In progress"** in transaction history forever — even after they've clearly succeeded or failed on-chain — while the app keeps polling Midgard in the background.

Observed case: a native THORChain tx that **failed** on-chain (explorer shows `Error — Zeroamount: execute wasm contract failed`) still displays "In progress…" days later. Logcat shows the app repeatedly hitting Midgard and getting an empty result:

```
GET …/thorchain_midgard/v2/actions?txid=1f14…
→ 200 {"actions": [], "count": "0"}
```

## Root cause

`ThorMayaChainStatusProvider` resolves status **only** through Midgard's `/v2/actions`. Midgard only indexes recognized *actions* (swap / addLiquidity / withdraw / savers / loan / bond). A plain native transfer (e.g. sending a secured asset) or a native deposit the node **rejects during execution** never becomes a Midgard action, so `/v2/actions` returns `actions: []`. The provider treated that as `Pending`, so the tx never reached a terminal state and was polled indefinitely.

This affected **all** native THORChain/MayaChain transfers, not just failed ones — a successful native send would also have been stuck "In progress".

## Fix

When Midgard returns no action, fall back to the native node's `cosmos/tx/v1beta1/txs/{hash}` endpoint, which reports the committed tx's result code directly:

- `code == 0` → **Confirmed**
- `code != 0` → **Failed** (surfacing the node's `raw_log`, e.g. "Zeroamount: execute wasm contract failed")
- `404` / not yet committed → **Pending** (keep polling)

Midgard remains the primary source so refund/failed-action semantics are unchanged; the native query is a fallback only used when no action exists. Reuses the provider's existing `HttpClient`, so the constructor and DI wiring are unchanged.

## Test plan

- [x] `./gradlew :data:testDebugUnitTest --tests "*ThorMayaChainStatusProviderTest"` — 18 tests, 0 failures
- Added native-fallback cases: failed code → Failed (with `raw_log`), code 0 → Confirmed, 404 → Pending, blank `raw_log` → default reason
- Existing Midgard action mapping (refund / failed / success / pending) unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced transaction status detection with an improved fallback mechanism to native chain endpoints, providing more reliable and accurate determination of transaction states (confirmed, pending, or failed). Better edge case handling ensures transactions are properly classified even when not immediately indexed on THORChain and MayaChain networks.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
@@ -1,5 +1,6 @@
 package com.vultisig.wallet.data.api.txstatus
 
+import com.vultisig.wallet.data.api.models.cosmos.CosmosEnvelopedTxResponse
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.usecases.txstatus.TransactionResult
 import com.vultisig.wallet.data.usecases.txstatus.TransactionStatusProvider
@@ -7,6 +8,7 @@ import com.vultisig.wallet.data.utils.bodyOrThrow
 import io.ktor.client.HttpClient
 import io.ktor.client.request.get
 import io.ktor.client.request.parameter
+import io.ktor.http.HttpStatusCode
 import javax.inject.Inject
 import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.serialization.SerialName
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5005, opened 2026-06-22
fix(banner): single dismiss motion on swipe-to-dismiss (#5001)

### LINKED ISSUE #5001: Notification banner double-animates on swipe-to-dismiss (re-centers then slides up)
## Summary

The foreground "Join transaction" notification banner double-animates when swiped away. After the horizontal swipe-to-dismiss flicks the banner off-screen, the banner **re-appears centered and then slides upward** before finally disappearing. The dismiss should be a single, consistent motion in the swipe direction.

## Steps to reproduce

1. Trigger the foreground signing notification banner (e.g. join a transaction so the "Join transaction" banner appears at the top of the screen).
2. Swipe the banner horizontally past ~35% of its width to dismiss it.
3. Watch the final moment of the dismissal.

## Observed behavior

1. The banner follows the finger horizontally and slides off-screen (diagonal/horizontal flick).
2. The banner is briefly gone.
3. **At the last second, the banner pops back to center and slides bottom-to-top out of view** — a second, unrelated animation.

Screen recording shows the banner sliding off horizontally, then reappearing at center and re-animating upward.

## Expected behavior

A single dismiss animation in the swipe direction. The banner should not re-appear and should not run a second vertical slide-out after the swipe completes.

## Root cause

There are **two independent dismiss animations** that both fire, plus an offset reset that re-centers the banner mid-exit:

1. **Horizontal swipe** — `ForegroundNotificationBanner.kt` animates `offsetX` off-screen on a successful drag:
   `app/src/main/java/com/vultisig/wallet/ui/components/banners/ForegroundNotificationBanner.kt:116-133`
   ```kotlin
   onDragEnd = {
       scope.launch {
           if (abs(offsetX.value) >= dismissThreshold) {
               val target = if (offsetX.value > 0) size.width.toFloat() else -size.width.toFloat()
               offsetX.animateTo(target)   // (1) banner slides horizontally off-screen
               onDismiss()                 // (2) sets foregroundNotification = null
               offsetX.snapTo(0f)          // (3) snaps banner back to center
           } else {
               offsetX.animateTo(0f)
           }
       }
   }
   ```

2. **Vertical AnimatedVisibility exit** — in `MainActivityContent.kt`, `onDismiss()` clears the state, flipping `visible` to false and triggering `slideOutVertically { -it }` (slide up):
   `app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt:157-180`
   ```kotlin
   AnimatedVisibility(
       visible = foregroundNotification != null,
       enter = slideInVertically { -it },
       exit = slideOutVertically { -it },   // vertical slide-up exit
       ...
   ) { ForegroundNotificationBanner(...) }
   ```

The sequence that produces the glitch:
- `offsetX.animateTo(target)` slides the banner horizontally off-screen.
- `onDismiss()` → `clearForegroundNotification()` sets `_foregroundNotification.value = null`, so `AnimatedVisibility` begins its **vertical** `slideOutVertically` exit. The banner subtree keeps composing for the duration of the exit transition.
- `offsetX.snapTo(0f)` then **re-centers** the still-composing banner horizontally. So the banner — which had just gone off-screen sideways — visibly snaps back to center and rides the vertical exit slide upward.

Net result: horizontal flick → re-center → vertical slide-up = the observed "moves from bottom to top at the last second."

## Suggested direction

Pick a single dismiss path rather than layering two:
- Either drive the dismiss purely via the swipe (`offsetX`) and remove/relax the `AnimatedVisibility` vertical exit for the swipe case, or
- Call `onDismiss()` immediately (skip the horizontal `animateTo`) and let `slideOutVertically` be the only dismiss animation, or
- Avoid the `snapTo(0f)` re-center while the exit transition is still running (e.g. reset the offset only on the next bind via the existing `LaunchedEffect(qrCodeData)` rather than inline after `onDismiss()`).

## Environment

- Branch: `main`
- Affects: foreground signing notification banner (`ForegroundNotificationBanner`)

aminsato:
## Summary

Fixes #5001 — the foreground "Join transaction" notification banner double-animated when swiped away. After the horizontal swipe-to-dismiss flicked the banner off-screen, it **re-appeared centered and slid upward** before disappearing.

## Root cause

Two independent dismiss animations were firing, plus an offset reset that re-centered the banner mid-exit:

1. The horizontal swipe animated `offsetX` off-screen (`ForegroundNotificationBanner.kt`).
2. `onDismiss()` cleared the state, flipping the parent `AnimatedVisibility` to its vertical `slideOutVertically { -it }` exit (`MainActivityContent.kt`). The banner subtree keeps composing for the exit's duration.
3. `offsetX.snapTo(0f)` then **re-centered** the still-composing banner, so it rode the vertical exit back into view.

Net: horizontal flick → re-center → vertical slide-up.

## Fix

Drive the offset reset off "is there a live request" instead of an inline post-dismiss snap:

- Removed the inline `offsetX.snapTo(0f)`. After a successful swipe the banner now stays off-screen, so the parent's vertical exit transition plays out invisibly — a single, consistent motion in the swipe direction.
- Reset the offset only on (re)entry: `LaunchedEffect(qrCodeData, isActive) { if (isActive) offsetX.snapTo(0f) }`, with `isActive = foregroundNotification != null` passed from `MainActivityContent`.

This preserves the re-broadcast guard the old inline snap provided: a new or re-broadcast push produces a fresh `ForegroundNotificationState` instance, flipping `isActive` true, which re-centers the banner before it re-enters. Crucially the reset never fires during the exit, so the glitch is gone.

## Test plan

- [ ] Trigger the foreground signing banner (join a transaction).
- [ ] Swipe horizontally past ~35% of its width — banner slides off in the swipe direction in a single motion, with no re-center or vertical slide-up.
- [ ] Swipe below the threshold — banner springs back to center.
- [ ] Receive a new signing request after dismissing — banner re-appears centered and slides in normally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced foreground notification banner dismissal behavior—the banner now correctly remains dismissed after swiping until a new notification arrives, improving the experience during screen transitions.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
@@ -172,6 +172,7 @@ private fun MainActivityContent(
                         qrCodeData = notification.qrCodeData,
                         vaultName = notification.vaultName,
                         transactionSummary = notification.transactionSummary,
+                        isActive = foregroundNotification != null,
                         onTap = onBannerTap,
                         onDismiss = onBannerDismiss,
                     )
--- a/app/src/main/java/com/vultisig/wallet/ui/components/banners/ForegroundNotificationBanner.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/banners/ForegroundNotificationBanner.kt
@@ -83,6 +83,7 @@ internal fun ForegroundNotificationBanner(
     qrCodeData: String,
     vaultName: String,
     transactionSummary: UiText,
+    isActive: Boolean,
     onTap: () -> Unit,
     onDismiss: () -> Unit,
     modifier: Modifier = Modifier,
@@ -97,9 +98,12 @@ internal fun ForegroundNotificationBanner(
     val offsetX = remember { Animatable(0f) }
     val scope = rememberCoroutineScope()
 
-    // Reset any residual swipe offset when a new (or re-shown) request is bound,
-    // so a previously dismissed banner doesn't reappear off-screen.
-    LaunchedEffect(qrCodeData) { offsetX.snapTo(0f) }
+    // Re-center only when a live request (re)enters. A new or re-broadcast push produces a fresh
+    // state instance, so `isActive` flips true and resets any residual swipe offset. Crucially we
+    // do NOT reset on dismissal: after a successful swipe the banner must stay off-screen while the
+    // parent's vertical exit transition plays out, otherwise the snap-back rides that exit and the
+    // banner visibly re-centers then slides up (#5001).
+    LaunchedEffect(qrCodeData, isActive) { if (isActive) offsetX.snapTo(0f) }
 
     Box(
         modifier =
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #1535, opened 2025-01-01
Add support for swapping on Solana using the LiFi provider

aminsato:
ETH -> SOL
https://etherscan.io/tx/0x35d75b330054c8b6aed11f24c046c28a1b92a0cf1eea1188b99c913c6e47b962

SOL -> WIF
https://solscan.io/tx/2TdXXYbyxGySTbVGLD3zXXJe2ST6tva7sDJmzQM4E8VSpp5m442qF9N3K966QeQovQv8pjoKDw3x3nDUgR5LYUdH

PR #5017, opened 2026-06-23
fix(swap): charge VULT-discounted LiFi integrator fee on Solana routes

### LINKED ISSUE #5012: Swap: LiFi non-EVM (Solana) routes bypass the VULT discount
## Summary

VULT-holder swap discounts do not apply to Solana (Jupiter) swaps. A VULT holder pays the full 50 bps affiliate fee on Solana regardless of tier, while the same holder gets the discount (down to 0 bps at Ultimate) on EVM and native THOR/Maya swaps. This is a token-utility hole and is now inconsistent with the public positioning on vultisig.com ("hold VULT, swap cheaper, free at the top tier").

## Where

`app/src/main/java/com/vultisig/wallet/data/usecases/GetDiscountBpsUseCase.kt:130` — the discount-eligible set excludes Jupiter:

```kotlin
private val supportedProviders =
    setOf(
        SwapProvider.THORCHAIN,
        SwapProvider.MAYA,
        SwapProvider.ONEINCH,
        SwapProvider.LIFI,
        SwapProvider.KYBER,
        SwapProvider.SWAPKIT,
    )
```

`GetDiscountBpsUseCase.kt:44` then returns 0 discount for any provider not in the set, so Jupiter swaps never receive the VULT discount.

Related: LiFi charges/discounts its integrator fee only on EVM routes (non-EVM LiFi skips the fee path), so non-EVM LiFi swaps also bypass the discount. Lower impact (no fee charged there today), but worth aligning for consistency.

## Fix

- Add `SwapProvider.JUPITER` to `supportedProviders` and ensure the post-discount affiliate bps flows into the Jupiter quote/fee path.
- Confirm a VULT holder on Solana sees the discounted bps (and 0 bps at Ultimate tier), at parity with EVM/native swaps.
- Align LiFi non-EVM fee/discount handling so the discount policy is uniform across chains.

## Acceptance
- [ ] Jupiter (Solana) swaps apply the VULT discount, matching EVM/native behavior
- [ ] Ultimate-tier holder swaps Solana at 0 bps
- [ ] Discount eligibility is documented as provider-agnostic (or the exceptions are explicitly justified)

Suggested owner: Amin/Roman. Part of the swap-quality push (see net-metric work in #5011).

aminsato:
Fixes #5012

## Problem

LiFi swaps that involve **Solana** were not charging the Vultisig integrator fee, and VULT holders got **no discount** on those routes — while EVM and native THOR/Maya swaps did. This is the LiFi side of #5012.

Root cause: `LiFiChainApiImpl.getSwapQuote()` omitted the `integrator`/`fee` params whenever Solana was the source or destination:

```kotlin
if (!isSolanaChainInvolved) {
    parameter("integrator", INTEGRATOR_ACCOUNT)
    parameter("fee", updatedFeeIntegrator)
}
```

That guard was added in **#1535 (Jan 2025)** because LiFi *rejected* those params on Solana routes at the time. A side effect: the UI still computed and displayed a ~0.5% fee via `integratorFeeAmount()` that was never actually collected (a display/charge mismatch).

## Fix

**LiFi now supports integrator fees on Solana routes** — verified against the live `li.quest/v1/quote` API: with `integrator=vultisig-android&fee=…` on USDC(Arbitrum)→SOL(Solana), the quote's `LIFI Fixed Fee` scales exactly with the fee param (and with the discount). So the 2025 limitation no longer applies.

This PR **removes the Solana guard** so the integrator fee — reduced by the VULT discount — is sent on **every** route, identical to EVM behavior:
- Solana LiFi swaps now charge the affiliate fee (recovering revenue).
- VULT holders get their tier discount on Solana, at parity with EVM/native (addresses #5012's intent).
- The displayed fee now matches what is actually collected.

## Test plan

- [x] `:data:testDebugUnitTest` — new `LiFiChainApiImplRequestTest` (MockEngine) pins that `integrator`/`fee` are sent on a Solana-destination route and that `fee` reflects the VULT discount (50bps→0.005; Gold −20bps→0.003). Existing `LiFiChainApiIntegratorFeeTest` still green.
- [x] `:app:compileDebugKotlin` green; ktfmt clean.
- [x] Live `li.quest` quote API confirms the fee param is honored on EVM→Solana and scales with the discount.
- [ ] **Mainnet device verification (remaining):** run an EVM→SOL LiFi swap and confirm (a) the quote still signs/broadcasts with the fee present, and (b) the affiliate fee actually settles. Before/after screenshots to follow.

## Jupiter (Solana) half — out of scope, infrastructure-blocked

The other half of #5012 (apply the VULT discount to Jupiter/Solana swaps) cannot be done as a client code change today, and is being tracked/closed as infra-and-product work rather than left open:

- Jupiter currently charges **no** Vultisig fee on either Android or iOS — `JupiterApi` sends no `platformFeeBps`/`feeAccount`, and iOS' Jupiter quote endpoint has none either. With no baseline fee there is nothing to discount (the issue's "0 bps at Ultimate" presupposes a fee that doesn't exist).
- Introducing one requires Jupiter's platform-fee mechanism: a real **Vultisig-owned Solana fee-collection account** + a token account (ATA) **per fee mint**, `platformFeeBps` on the quote + `feeAccount` on the swap, and almost certainly a change to the `api.vultisig.com/jup` backend proxy.
- That's on-chain infrastructure + a product/business decision (whether to monetize Solana, which account collects), not something safe to fabricate in this client PR.

So this PR resolves the **LiFi non-EVM** portion of #5012; the **Jupiter** portion is deferred to dedicated infra/product work.

**Side-finding (separate ticket):** 1inch is in `GetDiscountBpsUseCase.supportedProviders` and is passed `bpsDiscount`, but `OneInchQuoteSource` never reads it — the 1inch discount may itself be a no-op.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated swap quote requests so LI.FI `integrator` and fee parameters are consistently included, ensuring Solana-based routes receive the same fee/discount behavior as other supported chains.
* **Tests**
  * Added coverage to verify outgoing quote URLs include the expected `integrator` and `fee` values, including correct discount application for Solana-destination scenarios.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/LiFiChainApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/LiFiChainApi.kt
@@ -93,10 +93,6 @@ constructor(
                         BigDecimal(it).movePointLeft(4).stripTrailingZeros().toPlainString(),
                     )
                 }
-                // Always collect the (VULT-discounted) integrator fee, Solana routes included.
-                // LI.FI rejected the `integrator`/`fee` params on Solana routes when Solana support
-                // landed (#1535, Jan 2025), so they were omitted there; LI.FI has since added fee
-                // support for Solana, so the fee + discount now apply uniformly across all chains.
                 parameter("integrator", LiFiChainApi.INTEGRATOR_ACCOUNT)
                 parameter("fee", updatedFeeIntegrator)
             }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/LiFiChainApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/LiFiChainApi.kt
@@ -2,8 +2,6 @@ package com.vultisig.wallet.data.api
 
 import com.vultisig.wallet.data.api.models.quotes.LiFiSwapQuoteDeserialized
 import com.vultisig.wallet.data.api.models.quotes.LiFiSwapQuoteError
-import com.vultisig.wallet.data.models.Chain
-import com.vultisig.wallet.data.models.oneInchChainId
 import com.vultisig.wallet.data.utils.LiFiSwapQuoteResponseSerializer
 import io.ktor.client.HttpClient
 import io.ktor.client.request.get
@@ -69,10 +67,6 @@ constructor(
         bpsDiscount: Int,
         slippageBps: Int?,
     ): LiFiSwapQuoteDeserialized {
-        val isSolanaChainInvolved =
-            fromChain.toLong() == Chain.Solana.oneInchChainId() ||
-                toChain.toLong() == Chain.Solana.oneInchChainId()
-
         val bpsDiscountFee = round(bpsDiscount.toDouble()) / LiFiChainApi.BPS_DENOMINATOR_INT
         val updatedFeeIntegrator =
             (round(
@@ -99,10 +93,12 @@ constructor(
                         BigDecimal(it).movePointLeft(4).stripTrailingZeros().toPlainString(),
                     )
                 }
-                if (!isSolanaChainInvolved) {
-                    parameter("integrator", LiFiChainApi.INTEGRATOR_ACCOUNT)
-                    parameter("fee", updatedFeeIntegrator)
-                }
+                // Always collect the (VULT-discounted) integrator fee, Solana routes included.
+                // LI.FI rejected the `integrator`/`fee` params on Solana routes when Solana support
+                // landed (#1535, Jan 2025), so they were omitted there; LI.FI has since added fee
+                // support for Solana, so the fee + discount now apply uniformly across all chains.
+                parameter("integrator", LiFiChainApi.INTEGRATOR_ACCOUNT)
+                parameter("fee", updatedFeeIntegrator)
             }
         return try {
             json.decodeFromString(liFiSwapQuoteResponseSerializer, response.bodyAsText())
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/LiFiChainApiImplRequestTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/LiFiChainApiImplRequestTest.kt
@@ -1,0 +1,65 @@
+package com.vultisig.wallet.data.api
+
+import com.vultisig.wallet.data.utils.LiFiSwapQuoteResponseSerializerImpl
+import io.ktor.client.HttpClient
+import io.ktor.client.engine.mock.MockEngine
+import io.ktor.client.engine.mock.respond
+import io.ktor.http.HttpStatusCode
+import kotlin.test.assertContains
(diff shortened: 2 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/LiFiChainApi.kt
```diff
@@ -99,10 +93,12 @@ constructor(
...
-                if (!isSolanaChainInvolved) {
-                    parameter("integrator", LiFiChainApi.INTEGRATOR_ACCOUNT)
-                    parameter("fee", updatedFeeIntegrator)
-                }
+                // Always collect the (VULT-discounted) integrator fee, Solana routes included.
+                // LI.FI rejected the `integrator`/`fee` params on Solana routes when Solana support
+                // landed (#1535, Jan 2025), so they were omitted there; LI.FI has since added fee
+                // support for Solana, so the fee + discount now apply uniformly across all chains.
+                parameter("integrator", LiFiChainApi.INTEGRATOR_ACCOUNT)
+                parameter("fee", updatedFeeIntegrator)
```
rkokhatskyi, line 97: This now sends `integrator`/`fee` on Solana-source LiFi routes, which `SwapProviderTable` routes through LIFI under `Chain.Solana`, yet only the EVM-source direction was verified and `#1535` guarded the Solana-source one, so confirm a Solana-source quote still returns rather than silently dropping LiFi

data/src/main/kotlin/com/vultisig/wallet/data/api/LiFiChainApi.kt
```diff
@@ -99,10 +93,12 @@ constructor(
                         BigDecimal(it).movePointLeft(4).stripTrailingZeros().toPlainString(),
                     )
                 }
-                if (!isSolanaChainInvolved) {
-                    parameter("integrator", LiFiChainApi.INTEGRATOR_ACCOUNT)
-                    parameter("fee", updatedFeeIntegrator)
-                }
+                // Always collect the (VULT-discounted) integrator fee, Solana routes included.
```
rkokhatskyi: This comment should drop the `#1535, Jan 2025` history and the "uniformly across all chains" claim, which overstates since iOS gates the fee to EVM-source, keeping only the stable fact that LI.FI now accepts the `integrator`/`fee` params on Solana routes

johnnyluo (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/api/LiFiChainApi.kt
johnnyluo: Previously we had an issue with Li.fi when source chain is Solana , with integrator parameter , it returns an error , that has since been fixed on Li.fi configuration , so now we should be able to support solana like normal

we should remove the comments

PR #5021, opened 2026-06-23
feat(qbtc): governance proposals and voting

rkokhatskyi:
Adds a Governance tab to the QBTC DeFi screen for viewing proposals and voting on-chain.

QBTC's DeFi screen gains a **Governance** tab next to **Staked**. It lists proposals grouped into Active / Passed / Rejected sections with the current tally and the voter's own vote. Active proposals whose voting window is still open expose a Vote action (Yes / No / Abstain / No with veto) that runs through the existing verify → keysign flow; passed and rejected proposals are read-only.

Proposals come from the gov v1 RPC. Active proposals read the live `/cosmos/gov/v1/proposals/{id}/tally` endpoint, since cosmos-sdk leaves `final_tally_result` at zero until a proposal closes. The vote re-checks the voting window before staging, so a window that closes while the sheet is open aborts instead of signing a guaranteed-reject vote. Votes pay QBTC's `min_tx_fee` (800) rather than the generic Cosmos fee.

## Screenshots

| Governance tab (QBTC DeFi) |
| --- |
| governance |

## Testing

- `./gradlew :data:test :app:testDebugUnitTest --tests "*CosmosGovApiTest" --tests "*GovernanceViewModelTest"` — 18 tests pass (8 API, 10 ViewModel)
- `./gradlew :app:assembleDebug` — pass
- Live QBTC RPC verified: `/proposals` and `/proposals/{id}/tally` return HTTP 200

## Summary by CodeRabbit

* **New Features**
  * Added a new Governance section for QBTC, showing active, passed, and rejected proposals with vote tallies, statuses, and countdown timing.
  * Enabled viewing and casting votes from the app, including a vote confirmation sheet and “your vote” indicators.
  * Added governance support to the staking screen with a dedicated Governance tab.

* **Localization**
  * Added governance text strings in multiple languages, including vote actions, proposal states, time labels, and error messages.

* **Bug Fixes**
  * Improved handling for missing or closed voting states and empty proposal lists.


johnnyluo:
@rkokhatskyi good job , can you please move Governance to the tab , next to "stake" in Defi tab? 
CleanShot 2026-06-24 at 10 18 57@2x

app/src/main/java/com/vultisig/wallet/ui/models/governance/GovernanceViewModel.kt
```diff
@@ -0,0 +1,427 @@
...
+        return ProposalUi(
+            id = id,
+            title = title.orEmpty(),
+            summary = summary.orEmpty(),
+            status = status,
+            timeLabel = timeLabel(status, end, now, votingOpen),
+            // Votable only while the proposal is in its voting period AND the window is still open,
+            // so a status that lags the on-chain clock can't offer a guaranteed-reject vote.
+            isVotable = status == ProposalStatus.Active && votingOpen,
+            tally = finalTallyResult.toTally(),
```
johnnyluo: **Active proposals render `final_tally_result`, which is all-zeros during the voting period.** In standard cosmos-sdk gov v1, `final_tally_result` is only populated by the EndBlocker once voting closes — during the voting period every count is `"0"`. The live tally lives behind `/cosmos/gov/v1/proposals/{id}/tally`, which is never called here.

Result: an active proposal with real on-chain votes (e.g. 1000 YES) returns an all-zero `final_tally_result`, `toTally()` sees `total <= 0` and returns `TallyUi(hasVotes=false)`, so the Active card (the primary screen) shows the empty grey bar + "No votes yet" + no result headline.

Please verify against the live QBTC RPC. If QBTC doesn't fork the gov module to populate this field live, active proposals need the `/tally` endpoint instead of `final_tally_result`.

app/src/main/java/com/vultisig/wallet/ui/models/ChainDashboardUiModel.kt
```diff
@@ -168,6 +189,20 @@ constructor(
     }
 
     fun back() {
-        viewModelScope.launch { navigator.back() }
+        viewModelScope.launch {
+            // A sub-tab reached via the in-dashboard selector (the dashboard was opened at the
+            // Wallet view) returns to that Wallet view. A dashboard opened directly at a DeFi
+            // position from the DeFi portfolio pops as before, back to where it was opened from.
+            val enteredAtWallet = args.route is ChainDashboardRoute.Wallet
```
johnnyluo: **A `Governance` selection can leak app-wide and break the home toggle.** The connection type is held in the shared `CryptoConnectionTypeRepository` flow, but `back()` only resets it to `Wallet` when the dashboard was *entered at* the Wallet view (`enteredAtWallet`).

Repro: open the QBTC dashboard directly at a DeFi position (via the DeFi portfolio → `enteredAtWallet=false`), tap the Governance pill, press back → `navigator.back()` runs without resetting the flow. Back on the home screen, `CryptoConnectionSelect` is rendered with the default 2-pill `[Wallet, Defi]` and `activeType=Governance`: `indexOf` returns -1 → `coerceAtLeast(0)`, so the highlight sits on Wallet while *both* pills render greyed (`enabled = type == activeType` is false for both), and content shows DeFi accounts (`getAccounts` returns `defiAccounts` for any non-Wallet).

The `effectiveType` patch above only fixes the in-dashboard display, not the persisted flow. Consider resetting Governance → Wallet whenever leaving the QBTC dashboard (or coercing on the home screen).

app/src/main/java/com/vultisig/wallet/ui/models/governance/GovernanceViewModel.kt
```diff
@@ -0,0 +1,427 @@
...
+    fun openVoteSheet(proposal: ProposalUi) {
+        if (!proposal.isVotable) return
+        _state.update { it.copy(voteSheetProposal = proposal) }
+    }
+
+    fun dismissVoteSheet() {
+        _state.update { it.copy(voteSheetProposal = null) }
+    }
+
+    fun castVote(proposalId: String, option: VoteOption) {
```
johnnyluo, line 249: **`castVote` doesn't re-check `isVotable`, so a vote can be staged for a proposal whose window closed while the sheet was open.** `openVoteSheet` guards on `isVotable`, but `castVote` builds and stages the `QBTC_VOTE` deposit unconditionally.

Repro: open the vote sheet on a proposal with ~30s left, deliberate past the deadline, tap Confirm → a full keysign runs and the chain rejects the late vote. Cheap to fix by re-checking the voting window (or the proposal's current `isVotable`) before staging the transaction.

app/src/main/java/com/vultisig/wallet/ui/models/governance/GovernanceViewModel.kt
```diff
@@ -0,0 +1,427 @@
...
+                _state.update {
+                    it.copy(
+                        isLoading = false,
+                        isRefreshing = false,
+                        error = null,
+                        active = activeUi,
+                        passed = passedUi,
+                        rejected = rejectedUi,
+                        selectedTab =
+                            if (userPickedTab) it.selectedTab
```
johnnyluo: **`selectedTab` is silently recomputed on every refresh when `userPickedTab` is false, yanking the user off their current tab.** This runs on every `load()` including pull-to-refresh.

Repro: initial load has no active proposals → tab auto-selects Passed (user never tapped, so `userPickedTab` stays false); a refresh returns a new active proposal → `firstNonEmptyTab` recomputes to Active and the tab snaps from Passed to Active under the user's finger. Make the tab sticky after the first successful load.

app/src/main/java/com/vultisig/wallet/ui/models/governance/GovernanceViewModel.kt
```diff
@@ -0,0 +1,427 @@
...
+        ) {
+            val coin = resolveCoin(vaultId)
+            // QBTC's verified `min_tx_fee` floor. min_gas_price is 0 on the chain, so this flat
+            // amount is the entire fee — and `CosmosFeeService` returns 7_500 for QBTC, which
+            // overpays this 800 floor ~9x (see the QBTC entry in CosmosStakingConfig).
+            val feeAmount = BigInteger.valueOf(CosmosStakingConfig.feeAmountFor(Chain.Qbtc))
+            val gasFee = TokenValue(value = feeAmount, token = coin)
+
+            val specific =
+                blockChainSpecificRepository.getSpecific(
```
johnnyluo: **`getSpecific()` is called without `withContext(ioDispatcher)`, unlike every sibling Cosmos VM.** `CosmosDelegateViewModel` (and Undelegate/Redelegate/WithdrawRewards) wrap the identical call in `withContext(ioDispatcher)`, and `resolveCoin()` in this same file uses `ioDispatcher` too; `getSpecific` itself doesn't switch dispatchers. Here it runs on `viewModelScope` (Main). Ktor offloads the socket round-trip so it's not a hard ANR, but wrap it for consistency with the established pattern and to keep the parsing off Main.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -391,7 +391,9 @@ constructor(
 
         viewModelScope.launch {
             when (uiState.value.cryptoConnectionType) {
-                CryptoConnectionType.Wallet -> {
+                CryptoConnectionType.Wallet,
+                // Governance is a QBTC-only chain-dashboard sub-tab; treat it like Wallet here.
+                CryptoConnectionType.Governance -> {
```
johnnyluo: **Altitude:** `CryptoConnectionType.Governance` is special-cased as "treat like Wallet" across three sites — here, the `loadAccounts` filter below (~line 494), and the dashboard route/`back()` logic. Combined with the home-screen leak noted on `ChainDashboardUiModel.back()`, this scattering is why a stale Governance value is mishandled. A single derived predicate (e.g. `isWalletLikeTab`) or resetting Governance once at the dashboard boundary would localize the special case and close the gap.

johnnyluo (commented):
Automated review of the QBTC governance changes. Six points below, most-severe first — the active-tally one is the only behavioral blocker; the rest are correctness/UX/consistency.

app/src/main/java/com/vultisig/wallet/ui/models/governance/GovernanceViewModel.kt
rkokhatskyi: Confirmed against the live QBTC RPC — /cosmos/gov/v1/proposals/{id}/tally returns the running counts while final_tally_result stays zero through the voting period, so active proposals now read the tally endpoint via a new getGovTally and only closed ones keep final_tally_result.

app/src/main/java/com/vultisig/wallet/ui/models/ChainDashboardUiModel.kt
rkokhatskyi: Resolved by the tab move — Governance is no longer a CryptoConnectionType in the shared flow but a local tab inside the QBTC DeFi screen, so there's nothing to persist onto the home toggle and this back()/clamp special-casing is reverted to main.

app/src/main/java/com/vultisig/wallet/ui/models/governance/GovernanceViewModel.kt
rkokhatskyi: castVote now re-checks the proposal is still an open active one (votingEndTime in the future) before staging the deposit, so a window that closes while the sheet is open aborts with an error instead of burning a keysign on a guaranteed-reject vote.

app/src/main/java/com/vultisig/wallet/ui/models/governance/GovernanceViewModel.kt
rkokhatskyi: Gone with the tab move — the embedded view lists Active/Passed/Rejected as sections rather than status tabs, so there's no selectedTab left to recompute on refresh.

app/src/main/java/com/vultisig/wallet/ui/models/governance/GovernanceViewModel.kt
rkokhatskyi: Wrapped the getSpecific call in withContext(ioDispatcher) to match CosmosDelegateViewModel and the resolveCoin path in the same file.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
rkokhatskyi: Removed the scattering entirely — Governance is no longer a CryptoConnectionType, so all three wallet-like special-cases here, the loadAccounts filter, and the dashboard route/back are reverted; it's a self-contained tab in the QBTC DeFi screen now.

rkokhatskyi:
Done — Governance is now a tab next to Staked inside the QBTC DeFi screen and the bottom pill is gone. Also took the inline notes: active proposals pull the live `/tally` (final_tally_result reads zero mid-vote), the vote re-checks the window before signing, and getSpecific moved off the main thread.

governance tab

app/src/main/java/com/vultisig/wallet/ui/models/governance/GovernanceViewModel.kt
```diff
@@ -0,0 +1,419 @@
...
+                                async {
+                                    val tally =
+                                        runCatching { api.getGovTally(proposal.id) }.getOrNull()
+                                    val vote =
+                                        runCatching { api.getGovVote(proposal.id, coin.address) }
+                                            .getOrNull()
+                                            ?.options
+                                            ?.firstOrNull()
+                                            ?.option
+                                    proposal.id to ActiveExtras(tally, VoteOption.fromWire(vote))
```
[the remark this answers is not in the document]
rkokhatskyi: Dropped the runCatching here — getGovTally/getGovVote already return null on a real failure and now both rethrow CancellationException, so this only removes the redundant catch and the cancellation swallow; the per-proposal enrichment stays best-effort so one flaky tally/vote degrades that card rather than erroring the whole list.

app/src/main/java/com/vultisig/wallet/ui/models/governance/GovernanceViewModel.kt
```diff
@@ -0,0 +1,419 @@
...
+        if (proposal == null || proposal.votingEndTime?.isAfter(Instant.now()) != true) {
+            _state.update {
+                it.copy(
+                    voteSheetProposal = null,
+                    error = R.string.governance_voting_closed.asUiText(),
+                )
+            }
+            return
+        }
+        _state.update { it.copy(isSubmitting = true, error = null) }
```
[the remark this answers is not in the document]
rkokhatskyi: Added an isSubmitting early-return at the top of castVote so rapid taps can't enqueue a second DepositTransaction before Compose disables the button.

app/src/main/java/com/vultisig/wallet/ui/screens/cosmosstaking/CosmosStakingPositionsScreen.kt
```diff
@@ -112,7 +143,32 @@ internal fun CosmosStakingPositionsScreen(
...
+            )
+        }
+        val error = governanceState.error
+        if (error != null) {
+            UiAlertDialog(
+                title = stringResource(R.string.governance_error_title),
+                text = error.asString(),
+                onDismiss = governanceViewModel::dismissError,
+            )
+        }
```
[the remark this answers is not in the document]
rkokhatskyi: Gated the vote sheet and error dialog on governanceActive instead of showGovernance, so a background load error or a stale sheet can't render over the Staked tab.

johnnyluo (approved):
LGTM

PR #5025, opened 2026-06-24
fix(swap): keep THORChain referral split at full VULT discount

### LINKED ISSUE #5024: [Fix] Full VULT discount drops the referral code and pays the referrer 0 bps on THORChain swaps [Swap]
# [Fix] Full VULT discount drops the referral code and pays the referrer 0 bps on THORChain swaps [Swap]

## Problem
THORChain swap quotes build their `affiliate`/`affiliate_bps` query parameters in `ThorChainAffiliateHelper.buildAffiliateParams`. When the user's VULT tier discount reaches the maximum (50 bps, the ULTIMATE tier), the helper short-circuits and returns `affiliate_bps = "0"` with only the Vultisig affiliate address — silently dropping the referral code even when one is present. A user who was referred by someone and happens to hold enough VULT for the top tier stops paying their referrer entirely: the referrer's 10 bps cut disappears from the on-chain swap memo. The referrer keeps that 10 bps on the desktop app and the iOS app for the same swap, so referral payouts diverge by platform.

## Expected Behavior
At any discount tier, a referred THORChain swap keeps the nested affiliate split `affiliate=<code>/<vultisig>` and pays the referrer their fixed 10 bps; only the Vultisig affiliate portion shrinks with the discount and floors at 0. At the full (50 bps) discount a referred swap should emit `affiliate_bps = "10/0"` rather than `"0"` with no code. Swaps without a referral code are unchanged: a single Vultisig affiliate whose bps floors at 0.

## Steps to Reproduce
1. Hold enough VULT to reach the top tier (50 bps discount) on the active vault.
2. Save a friend referral code on that vault.
3. Start a THORChain-routed swap and inspect the affiliate parameters / on-chain memo.
4. The memo carries `affiliate=<vultisig>` and `affiliate_bps=0` — the referral code and the referrer's 10 bps are gone, instead of `affiliate=<code>/<vultisig>` with `affiliate_bps=10/0`.

## Solution
Remove the `discountBps >= 50` early return in `buildAffiliateParams`. The existing referral / no-referral branches already subtract the discount through `maxOf(0, baseBps - discountBps)`, so they produce the correct result at full discount on their own (`10/0` with a referral, `0` without) and match the desktop and iOS clients. No new branching or constants are needed.

### Must Do
- [ ] Remove the `discountBps >= 50` short-circuit so the referral branch runs at full discount and emits `10/0` with the code preserved
- [ ] Keep the no-referral path emitting a single affiliate whose bps floors at 0
- [ ] Update `ThorChainAffiliateHelperTest` so the full-discount cases assert the preserved nested split

### Must NOT Do
- Don't change the affiliate address constant or the base/referred bps constants
- Don't touch the MayaChain affiliate path (separate, intentionally non-nested) or the swap quote/signing pipeline
- Don't alter the no-referral behavior or add new fee tiers

## Acceptance Criteria
- [ ] New/updated test: a referral code at 50 bps discount yields `affiliate = "<code>/<addr>"` and `affiliate_bps = "10/0"` (fails before the change, passes after)
- [ ] No-referral case at 50 bps discount still yields a single affiliate with `affiliate_bps = "0"`

rkokhatskyi:
## Description

Fixes #5024

THORChain swap quotes build their `affiliate`/`affiliate_bps` query parameters in `ThorChainAffiliateHelper.buildAffiliateParams`. The helper short-circuited at `discountBps >= 50` and returned a single Vultisig affiliate with `affiliate_bps = "0"`, dropping the referral code even when one was present. So a referred user who holds enough VULT to reach the top tier (50 bps discount) stopped paying their referrer entirely — the referrer's fixed 10 bps cut disappeared from the on-chain memo, while the desktop and iOS apps keep paying it for the same swap.

The discount is meant to reduce only the Vultisig affiliate portion (it floors at 0), not the referrer's reward. Removing the early return lets the existing referral / no-referral branches run; both already subtract the discount through `maxOf(0, baseBps - discountBps)`, so at full discount a referred swap emits `affiliate = "<code>/va"` with `affiliate_bps = "10/0"` (referrer keeps 10 bps, Vultisig 0), and a swap with no referral still emits a single affiliate with `affiliate_bps = "0"`. This matches the desktop and iOS clients.

## Which feature is affected?

## Checklist

Not applicable — this changes the affiliate query parameters sent to the THORChain quote endpoint; there is no UI surface. The swap form's referral-discount display is driven separately and is unchanged.

## Additional context

The MayaChain affiliate path is intentionally separate (no nested affiliates) and is untouched. The signing/keysign path is unaffected: the affiliate parameters only change what is sent to the quote endpoint; the local device and the VultiServer co-signer continue to sign the identical quote/memo returned by the node, so there is no divergence between regular and co-signed swaps.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed affiliate fee handling for full-discount scenarios so referral-based payouts are still applied when a referral code is present.
  * Updated zero-discount cases to keep the expected flat affiliate output when no referral code is provided.
  * Improved coverage for affiliate calculations to match the new behavior across both empty and populated referral inputs.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/ThorChainAffiliateHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/ThorChainAffiliateHelper.kt
@@ -10,14 +10,6 @@ object ThorChainAffiliateHelper {
      * @param discountBps discount in basis points to subtract from the base affiliate fee
      */
     fun buildAffiliateParams(referralCode: String, discountBps: Int): Map<String, String> {
-        // Full discount: drop all affiliate fees and ignore the referral code.
-        if (discountBps >= 50) {
-            return mapOf(
-                "affiliate" to THORChainSwaps.AFFILIATE_FEE_ADDRESS,
-                "affiliate_bps" to "0",
-            )
-        }
-
         return if (referralCode.isNotEmpty()) {
             val affiliateFeeRateBp =
                 calculateBpsAfterDiscount(
--- a/data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/ThorChainAffiliateHelperTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/ThorChainAffiliateHelperTest.kt
@@ -68,7 +68,7 @@ class ThorChainAffiliateHelperTest {
     }
 
     @Test
-    fun `discount of 50 bps short-circuits to clean referral with zero affiliate fee`() {
+    fun `full discount without referral code zeroes the affiliate fee`() {
         val params =
             ThorChainAffiliateHelper.buildAffiliateParams(referralCode = "", discountBps = 50)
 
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5029, opened 2026-06-24
fix(swap): clamp 1inch referrer fee at zero

### LINKED ISSUE #5026: [Fix] 1inch referrer fee can go negative when the VULT discount exceeds the base fee [Swap/1inch]
## Problem
On 1inch swaps the referrer (affiliate) fee sent to the 1inch API is the base 0.5% fee minus the user's VULT discount, with no lower bound. Today the largest VULT discount (50 bps) exactly matches the base fee, so the value bottoms out at 0%. If a discount tier is ever raised — or any path passes a discount above the base fee — the computed fee turns negative, and 1inch rejects the request, breaking swaps for the highest VULT tier.

## Expected Behavior
The referrer fee should never drop below 0%. When the discount meets or exceeds the base fee, the value sent to 1inch should be exactly 0, matching the other providers (LI.FI, Kyber, SwapKit, THORChain/Maya), which all clamp the discounted fee at zero.

## Steps to Reproduce
1. Hold enough VULT to reach the maximum discount tier (or any tier whose discount ≥ the base referrer fee).
2. Start a 1inch swap.
3. The `fee` query parameter sent to the 1inch quote/swap endpoint is `base − discount`; with `discount > base` this is negative and 1inch rejects the request.

> Today the discount caps at exactly the base fee, so the value lands on 0 rather than going negative — the defect is currently latent but one tier change away from breaking swaps.

## Solution
Clamp the discounted referrer fee at 0 before it is sent to 1inch, the same way every other provider already does. Pull the small fee calculation into a testable helper so the clamp is covered by a unit test.

### Must Do
- [ ] Clamp the 1inch referrer fee at 0 so the `fee` parameter can never be negative.
- [ ] Add a unit test covering discount = 0, discount = base fee, and discount > base fee.

### Must NOT Do
- Don't change the base referrer fee, referrer address, or slippage handling.
- Don't alter the discount-tier values or how the VULT discount is computed.
- Don't touch the fee logic of any other swap provider.

## Acceptance Criteria
- [ ] With a discount greater than the base fee, the computed 1inch fee is `0.0`, not negative.
- [ ] With no discount, the fee remains the base `0.5`.

rkokhatskyi:
## Description

The affiliate fee 1inch is sent for a swap was computed as the base `0.5%` referrer fee minus the user's VULT discount, with no lower bound:

```kotlin
round((ONEINCH_REFERRER_FEE - bpsDiscountTransformed) * 100) / 100.0
```

The highest VULT tier already discounts the full 50 bps, which exactly equals the 0.5% base fee, so today the value bottoms out at `0.0`. But there is no clamp, so any future tier increase — or any path passing a discount above the base fee — would send a negative `fee` to 1inch, which rejects it and breaks the swap for top-tier holders. Every other provider already clamps this (LI.FI, Kyber, SwapKit, THORChain/Maya), and iOS clamps it in `OneInchService.bps(for:)` (`max(0, referredFee - formattedDiscount)`); only the Android 1inch path didn't.

This clamps the discounted referrer fee at zero and pulls the small calculation into a testable helper so the clamp is covered by a unit test. The value on the wire is unchanged for every discount the app can produce today (0 → `0.5`, 50 → `0.0`); the only behavioral change is that a discount above the base fee now yields `0.0` instead of a negative number.

Fixes #5026

## Which feature is affected?

> No swap tx link: the defect is latent (the largest discount the app can reach today lands the fee on exactly `0.0`, so the request 1inch receives is byte-identical before and after this change). The fix is a safety clamp against future discount-tier changes, verified by unit test.

## Checklist

Not applicable — data-layer change with no UI surface.

## Additional context

Co-signer impact: none. The `fee` parameter feeds the 1inch quote/swap request that produces the EVM calldata the VultiServer co-signer later signs, but the value sent is identical for every discount the app can currently produce (max discount 50 bps = base 0.5% → `0.0`), so the signed payload is unchanged. The clamp only diverges from the old behavior for a discount greater than the base fee, which is unreachable today.

`./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.OneInchReferrerFeeTest"` passes (4/4).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap quote fee calculation so affiliate fees now reflect the configured discount correctly.
  * Prevented fees from dropping below zero when discounts meet or exceed the base fee.

* **Tests**
  * Added coverage for fee calculation scenarios, including no discount, partial discount, and full discount cases.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/swapAggregators/OneInchApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/swapAggregators/OneInchApi.kt
@@ -123,10 +123,6 @@ constructor(
         bpsDiscount: Int,
         slippageBps: Int?,
     ) {
-        val bpsDiscountTransformed = round(bpsDiscount.toDouble()) / 100.0
-        val referrerFeeUpdated =
-            round((ONEINCH_REFERRER_FEE - bpsDiscountTransformed) * 100) / 100.0
-
         // bps → percent (100 bps = 1%); Auto (null) keeps 1inch's historical 0.5% default.
         val slippagePercent = slippageBps?.let { it.toDouble() / 100.0 } ?: ONEINCH_DEFAULT_SLIPPAGE
 
@@ -138,7 +134,7 @@ constructor(
         parameter("disableEstimate", true)
         parameter("includeGas", true)
         parameter("referrer", ONEINCH_REFERRER_ADDRESS)
-        parameter("fee", if (isAffiliate) referrerFeeUpdated else "0")
+        parameter("fee", if (isAffiliate) discountedReferrerFee(bpsDiscount) else "0")
     }
 
     override suspend fun getTokens(chain: Chain): OneInchTokensJson =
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

Issue #5013, opened 2026-06-22
Advanced Swap parity (slippage / gas / recipient)


realpaaao:
> **Note:** the slippage **value plumbing** to all providers is split out to #5050; this ticket now covers the Advanced Swap **UI** (slippage / gas / recipient) parity with iOS #4546.

## Summary

Two related gaps:

1. **No Advanced Swap on Android.** iOS has Advanced Swap (user-set slippage, gas limit, external recipient) tracked in vultisig/vultisig-ios#4546. Android has no equivalent.
2. ~~**The existing Android slippage field is a partial no-op.** The swap form already exposes a slippage value (`SendFormUiModel.kt:62`, default `"1.0"`), but it only reaches KyberSwap and 1inch. THORChain, MayaChain, LiFi, Jupiter, and SwapKit ignore it. So the user adjusts a control that silently does nothing for most providers, which is a trust bug on a money path.~~ — moved to #5050

## Desired behavior

- ~~Plumb the user slippage value to every provider that accepts one, converting to each provider's native unit (THOR/Maya `tolerance_bps`, Kyber bps, 1inch percent, LiFi fraction, SwapKit percent). For providers that genuinely cannot take a client-side slippage override, hide/disable the control rather than show a no-op.~~ — moved to #5050
- Bring Advanced Swap to parity with iOS #4546: user slippage, gas limit, and external recipient.

## Acceptance
- [ ] ~~Changing the slippage value measurably changes the min-received / tolerance sent to each provider that supports it (verified per provider)~~ — moved to #5050
- [ ] ~~No provider silently ignores a user-set slippage while the control is shown as active~~ — moved to #5050
- [ ] Advanced Swap (slippage + gas + external recipient) reaches parity with iOS #4546

Ref: iOS #4546. Suggested owner: Amin/Roman. Part of the swap-quality push (companion to net-metric #5011).

rkokhatskyi:
Regading 2, a related ticket under PR #5030  (all providers to be checked through)

rkokhatskyi:
Split the ticket into 2: this one is for UI work and #5050 for slippage values plumbing

johnnyluo:
fixed

PR #5030, opened 2026-06-24
fix(swap): send slippage tolerance on Jupiter quotes

### LINKED ISSUE #5027: [Fix] Jupiter swap quote ignores slippage tolerance [Swap/Solana]
# [Fix] Jupiter swap quote ignores slippage tolerance [Swap/Solana]

## Problem
The Jupiter quote request used for Solana swaps sends only `inputMint`/`outputMint`/`amount` and no slippage tolerance, so the minimum-received that gets baked into the signed swap transaction is whatever default the Jupiter API picks server-side rather than a value the app controls. The swap form already exposes a slippage control (added in #4969) and threads the chosen value to the other direct aggregators — 1inch and Kyber both honor it and fall back to their own default on Auto — but the Jupiter path drops it. So a user who adjusts slippage sees no effect on Solana swaps, and the Auto case commits to an opaque tolerance instead of an explicit one.

## Expected Behavior
Jupiter, a direct aggregator that accepts a quote-time `slippageBps` parameter, should behave like 1inch and Kyber: forward a user-set slippage to the quote, and on Auto pin a deterministic default of 0.5% (50 bps) — matching 1inch's default and Jupiter's own documented default — so the slippage the app commits to is explicit and stable.

## Steps to Reproduce
1. Open the swap form and pick a Solana source token (a route that quotes via Jupiter, e.g. SOL → USDC).
2. Set a slippage value in the swap controls.
3. Observe the outgoing `…/jup/swap/v1/quote` request: it carries no `slippageBps`, so the chosen value never reaches the provider and the minimum-received is set by the API default.

## Solution
Thread the request's slippage through `JupiterQuoteSource` into `JupiterApi.getSwapQuote` and add a `slippageBps` query parameter to the quote call, defaulting to 50 bps when the user is on Auto (`request.slippageBps ?: 50`), mirroring the existing `slippageBps ?: SLIPPAGE_TOLERANCE` pattern in `KyberApi` and the percent fallback in `OneInchApi`. Update the `SwapQuoteRequest.slippageBps` documentation so Jupiter is listed among the providers that honor the override rather than among the server-side-protection ones. This is the Jupiter slice of the broader slippage-plumbing tracked in #5013; LI.FI and SwapKit, which embed slippage in opaque server routing, stay as-is.

### Must Do
- [ ] Add a `slippageBps: Int?` parameter to `JupiterApi.getSwapQuote` and send it as the `slippageBps` query param on the `/swap/v1/quote` request, defaulting to 50 bps on Auto.
- [ ] Pass `request.slippageBps` from `JupiterQuoteSource.fetch` into the API call.
- [ ] Update the `SwapQuoteRequest.slippageBps` KDoc to reflect that Jupiter now honors the override.
- [ ] Add a `JupiterApiTest` asserting the quote request carries `slippageBps=50` on Auto and the user value when set, and update the existing Jupiter mocks for the new signature.

### Must NOT Do
- Don't change LI.FI or SwapKit slippage handling (they apply slippage server-side, out of scope here).
- Don't add or restyle any swap UI; this is a data-layer parity fix with no visible surface change.
- Don't modify the Jupiter swap-build (`/swap/v1/swap`) body, the priority-fee/compute-budget logic, or any JNI/crypto code.

## Acceptance Criteria
- [ ] The Jupiter `/swap/v1/quote` request carries `slippageBps=50` when slippage is on Auto.
- [ ] A user-set slippage value is forwarded verbatim to the Jupiter quote request.

rkokhatskyi:
## Description

The Jupiter quote request used for Solana swaps sent only `inputMint`/`outputMint`/`amount`, so the minimum-received that gets baked into the signed swap transaction was whatever default the Jupiter API applied server-side rather than a value the app controls. The swap form's slippage control already threads the chosen value to the other direct aggregators — 1inch and Kyber both honor it and fall back to their own default on Auto — but the Jupiter path dropped it, so a user-set slippage had no effect on Solana swaps.

This forwards `request.slippageBps` into `JupiterApi.getSwapQuote` and adds a `slippageBps` query parameter on the quote call, defaulting to 50 bps (0.5%) on Auto — mirroring `KyberApi`'s `(slippageBps ?: SLIPPAGE_TOLERANCE)` and `OneInchApi`'s percent default, and matching Jupiter's own documented default. LI.FI and SwapKit, which apply slippage in opaque server-side routing, are unchanged. This is the Jupiter slice of the broader slippage-plumbing tracked in #5013.

Fixes #5027

## Which feature is affected?

> Data-layer change to the Jupiter quote request only; no on-chain tx link is attached because the fix is verified by unit tests asserting the outgoing request, not a live broadcast.

## Checklist

No UI surface changes — the slippage control already exists; this only changes the tolerance value sent to Jupiter on the wire.

## Additional context

Co-signer impact: none. The change is confined to the quote-request parameters in the data layer. The resulting Solana swap transaction is consumed opaquely by the keysign/co-sign flow, which signs the bytes the app builds and does not independently derive or validate slippage — a tighter, app-controlled minimum-received only changes the transaction content, not the signing or broadcast path.

Tests: `JupiterApiTest` (MockEngine) asserts the `/swap/v1/quote` request carries `slippageBps=50` on Auto and the user value when set; `SwapQuoteRepositoryProvidersTest` verifies `JupiterQuoteSource` forwards `request.slippageBps`. `./gradlew :data:testDebugUnitTest` for the two classes passes (JupiterApiTest 2/2, SwapQuoteRepositoryProvidersTest 11/11).

## Summary by CodeRabbit

* **New Features**
  * Swap quotes now accept an explicit quote-time slippage value, applying a default when none is provided.
  * The selected slippage is forwarded through to supported quote flows (including Jupiter).

* **Bug Fixes**
  * Quote caching and request building now consistently include slippage, preventing mismatched cached results between different slippage values.
  * Updated provider slippage behavior messaging to reflect current behavior (including Jupiter).

* **Documentation**
  * Refreshed quote-time slippage notes for provider handling differences.

* **Tests**
  * Added/updated tests covering default vs custom slippage, forwarding, and 429 quote retry behavior.


data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/JupiterQuoteSource.kt
```diff
@@ -21,6 +21,7 @@ internal class JupiterQuoteSource @Inject constructor(private val jupiterApi: Ju
                     toToken = toToken,
                     fromAmount = request.tokenValue.value.toString(),
                     fromAddress = request.srcAddress,
+                    slippageBps = request.slippageBps,
```
johnnyluo, line 24: **The user-set slippage never reaches here — `request.slippageBps` is always `null` for Jupiter.**

The only caller that builds the Jupiter `SwapQuoteRequest` is `SwapQuoteManager.fetchLiFiJupiterQuote` (app module, ~line 849), and it constructs the request *without* a `slippageBps` argument:

```kotlin
SwapQuoteRequest(
    srcToken = srcToken,
    dstToken = dstToken,
    tokenValue = tokenValue,
)
```

Since `SwapQuoteRequest.slippageBps` defaults to `null`, every quote reaching this line has `request.slippageBps == null`, so `JupiterApi` always falls back to the new `DEFAULT_SLIPPAGE_BPS` (50). The `SwapQuoteRepositoryProvidersTest` passes only because it constructs the request directly with `slippageBps = 250` — a real swap from the form never does.

Net effect in production: the only behavior change is that Auto now sends a fixed 50 bps instead of Jupiter's server-side default. The PR's stated goal — that a *user-set* slippage takes effect on Solana swaps — is not achieved. To wire it, pass `slippageBps = slippageBps` in the Jupiter request at `SwapQuoteManager.kt:849` (the param is already in scope), **and** see the cache-key note below.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapQuoteSource.kt
```diff
@@ -25,8 +25,8 @@ data class SwapQuoteRequest(
     /**
      * User-chosen slippage tolerance in basis points (100 = 1%), or null for "Auto" — in which case
      * each provider keeps its own default ([DEFAULT_THORCHAIN_TOLERANCE_BPS] for THORChain/Maya,
-     * 0.5% for 1inch, 1% for Kyber). Honoured only by providers that accept a quote-time slippage
-     * override; LI.FI / SwapKit / Jupiter ignore it (server-side protection). See #4858.
+     * 0.5% for 1inch and Jupiter, 1% for Kyber). Honoured only by providers that accept a
```
johnnyluo: Two follow-on notes tied to this KDoc change (now lists Jupiter as honoring quote-time slippage):

1. **Stale contradicting comment**: `SwapQuoteManager.fetchLiFiJupiterQuote` (~line 816) still reads *"LI.FI accepts a quote-time slippage override (fraction); Jupiter does not"*, which now contradicts this KDoc and the data-layer change.

2. **Cache key is slippage-agnostic for Jupiter**: in that same method, `liFiSlippageBps = slippageBps.takeIf { provider == SwapProvider.LIFI }` is `null` for Jupiter and is what's fed to `getCachedQuoteOrFetch` as the cache-key slippage component. So even after the request is fixed to pass `slippageBps`, two requests differing only in slippage would collide on the cache and return a stale quote. Both need updating for the override to work end-to-end.

johnnyluo (commented):
Automated review. The data-layer plumbing here is correct and well-tested, but a cross-file trace shows the user-set slippage still doesn't reach the wire because the caller (`SwapQuoteManager.fetchLiFiJupiterQuote`) — not part of this diff — never populates `slippageBps` on the Jupiter request. Inline comments anchored to the closest diff lines below.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/JupiterQuoteSource.kt
rkokhatskyi: Confirmed — the data-layer thread was inert because `SwapQuoteManager.fetchLiFiJupiterQuote` built the Jupiter `SwapQuoteRequest` without `slippageBps`, so it now passes `slippageBps = slippageBps` there like every other provider, which makes a user-set tolerance reach the Solana quote instead of only flipping the Auto default to 50 bps, and a new `SwapQuoteManagerTest` captures the Jupiter request and asserts it carries the chosen value.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapQuoteSource.kt
rkokhatskyi: Both addressed — the stale comment now reads that LI.FI and Jupiter both take a quote-time override, and the cache key drops the `liFiSlippageBps` gate to pass `slippageBps` for both providers, which stops two tolerances for the same pair colliding on a stale Jupiter quote, with a new test that fetches one pair at two slippages and asserts a re-quote rather than a cache hit. I also corrected the `SwapQuoteRequest` doc, since `LiFiQuoteSource` forwards the override as well — only SwapKit genuinely applies a server-side floor.

johnnyluo (approved):
LGTM

PR #5031, opened 2026-06-24
fix(swap): clamp quote countdown ring progress to 0f..1f after expiry

### LINKED ISSUE #5028: [Fix] Swap quote countdown ring receives out-of-range progress after expiry [Swap]
## Problem
The swap screen shows a small circular countdown ring next to the advanced-settings button that tracks how long the current quote stays valid (1 minute). The value it feeds `CircularProgressIndicator` is `remainingTime / quoteLifetime`. Once a quote reaches 0:00 before a fresh quote replaces it, `remainingTime` is negative, so the ring is driven with a negative progress — outside the indicator's documented `0f..1f` range. The countdown text next to it is already floored at zero, so the two disagree, and the ring's appearance then depends on the framework's undocumented handling of out-of-range values.

## Expected Behavior
At and past expiry the ring's progress should be `0` — matching the floored countdown text and staying within `CircularProgressIndicator`'s `0f..1f` contract.

## Steps to Reproduce
1. Open Swap in Market mode and let a quote load so the countdown ring appears.
2. Leave the screen idle until the timer reaches 0:00 before a new quote arrives.
3. The progress fed to the ring is now negative (outside `0f..1f`) even though the countdown text reads `00:00`.

## Solution
Clamp the computed fraction to `0f..1f` before it reaches `CircularProgressIndicator`, the same way the countdown text is already floored at zero. Keep the math in a small testable helper so the boundary is covered by a unit test.

### Must Do
- [ ] Clamp the quote-timer progress fraction to `0f..1f` before it reaches `CircularProgressIndicator`.
- [ ] Add a unit test covering negative (expired) and over-range remaining times.

### Must NOT Do
- Don't change the countdown text formatting or the 1-minute quote lifetime.
- Don't alter quote refresh/expiry timing or any swap quoting logic.
- Don't restyle the timer pill or touch unrelated swap UI.

## Acceptance Criteria
- [ ] `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.screens.swap.components.QuoteTimerProgressTest"` passes.
- [ ] After expiry, the value passed to the ring is `0` (no negative progress).
- [ ] A non-expired quote still shows the ring filling and emptying as before.

rkokhatskyi:
## Description

Fixes #5028

The swap quote countdown ring feeds `CircularProgressIndicator` the fraction `remainingTime / quoteLifetime`. Once a quote reaches 0:00 before a fresh quote replaces it, `remainingTime` is negative, so the ring is driven with a negative progress — outside the indicator's documented `0f..1f` range, while the countdown text right next to it is already floored at zero.

This clamps the fraction to `0f..1f` in a small pure helper (`quoteTimerProgress`) so the ring's value matches the floored text and stays inside the indicator's contract. The helper is unit-tested at the boundaries.

## Which feature is affected?

> UI-only change to the swap quote countdown ring; no on-chain transaction is involved.

## Checklist

No visible before/after delta to show. The current Material3 (`material3` 1.4.0) determinate `CircularProgressIndicator` renders a negative progress the same as `0` (empty ring), so the expired-quote ring already looks correct on screen — I verified this by rendering the expired state both with and without the clamp and getting identical output. The fix removes the out-of-contract value the indicator is fed, which the text is already guarded against and which the framework's masking of out-of-range input is not guaranteed to keep doing.

## Additional context

- The countdown text already floors remaining time at zero (`inWholeSeconds.coerceAtLeast(0)`); this brings the ring's progress in line with it.
- Parity: the Windows app clamps the remaining-time value at the source (`Math.max(prev - 1, 0)`), and iOS resets its timer rather than letting it go negative — Android now stays within range too.
- No signing, keysign, payload, or co-sign path is touched; this only affects the value passed to a progress indicator on the swap form.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quote timer progress handling so the circular indicator stays within a valid range, including after a quote expires.
  * Prevented negative or overfilled progress values from being shown in the swap timer.

* **Tests**
  * Added coverage for quote timer progress across full, partial, zero, expired, and over-limit timing cases.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/QuoteTimer.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/QuoteTimer.kt
@@ -43,7 +43,7 @@ internal fun QuoteTimer(expiredAt: Instant, modifier: Modifier = Modifier) {
             val now = Clock.System.now()
             val left = expiredAt - now
             timeLeft = formatDurationAsMinutesSeconds(left)
-            progress = (left / expiredAfter).toFloat()
+            progress = quoteTimerProgress(left, expiredAfter)
         }
     }
 
@@ -85,3 +85,11 @@ private fun formatDurationAsMinutesSeconds(duration: Duration): String {
     val seconds = totalSeconds % 60
     return String.format(Locale.getDefault(), "%02d:%02d", minutes, seconds)
 }
+
+/**
+ * Fraction of the quote's lifetime still remaining, clamped to the 0f..1f range
+ * [CircularProgressIndicator] requires. [remaining] goes negative once the quote expires before it
+ * is replaced, so the raw ratio is coerced rather than passed straight to the indicator.
+ */
+internal fun quoteTimerProgress(remaining: Duration, lifetime: Duration): Float =
+    (remaining / lifetime).toFloat().coerceIn(0f, 1f)
--- a/app/src/test/java/com/vultisig/wallet/ui/screens/swap/components/QuoteTimerProgressTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/screens/swap/components/QuoteTimerProgressTest.kt
@@ -1,0 +1,36 @@
+package com.vultisig.wallet.ui.screens.swap.components
+
+import kotlin.test.assertEquals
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5033, opened 2026-06-24
fix(swap): set real byte fee on THORChain UTXO swap pre-signing input

### LINKED ISSUE #5032: [Fix] THORChain UTXO swap pins a placeholder byte fee before signing [Swap]
## Problem

When building a THORChain swap on a UTXO chain (Bitcoin, Litecoin, Dogecoin, Bitcoin Cash), `UtxoHelper.getSwapPreSigningInputData` hardcodes the transaction's per-byte fee to `1` sat/vByte. The value is harmless today because the only caller immediately re-runs the builder through `getSigningInputData`, which overwrites it with the real fee from the keysign payload. The hardcoded `1` is therefore never used — but it is a trap: any future caller that builds the swap input without that second pass would sign a THORChain swap at 1 sat/vByte, which the network rejects.

## Expected Behavior

The swap pre-signing input should carry the real per-byte fee from the keysign payload (the same `byteFee` the regular send path uses), so the builder produces a correct input on its own and there is no placeholder to accidentally rely on.

## Steps to Reproduce

This is a latent code-correctness issue with no user-visible symptom on the current call path:

1. Open `UtxoHelper.getSwapPreSigningInputData` and note `.setByteFee(1L)`.
2. Its only caller, `THORChainSwaps.getPreSignedInputData`, pipes the result through `getSigningInputData`, which re-sets the fee from `keysignPayload`.
3. The `1` is never used — but a refactor that skips the second pass would silently produce a 1 sat/vByte swap.

## Solution

Set the byte fee in `getSwapPreSigningInputData` from the payload's UTXO `byteFee` (the same way the sibling `getBitcoinSigningInput` builds its input) instead of the hardcoded `1`. The downstream `getSigningInputData` already applies that exact value, so the signed transaction is byte-for-byte identical; this only removes the misleading literal and makes the function correct in isolation.

### Must Do
- [ ] Replace the hardcoded `setByteFee(1L)` in `UtxoHelper.getSwapPreSigningInputData` with the payload's real UTXO byte fee.
- [ ] Add a unit test asserting the swap pre-signing input carries the payload byte fee, not a placeholder.

### Must NOT Do
- Don't change the signing output / transaction bytes — the fee that reaches the wire must stay exactly what `getSigningInputData` applies today.
- Don't remove or alter the `setByteFee` call inside `getSigningInputData`; it finalizes any builder handed to it and is shared with the regular send path.
- Don't touch the unrelated `setByteFee(1L)` placeholders in the SwapKit PSBT signers (they intentionally pair with a frozen transaction plan).
- Don't refactor the surrounding swap/helper logic.

## Acceptance Criteria
- [ ] `UtxoHelper.getSwapPreSigningInputData` no longer contains a hardcoded byte-fee literal.
- [ ] A unit test asserts the returned input's `byteFee` equals the payload's `byteFee`.

rkokhatskyi:
## Description

`UtxoHelper.getSwapPreSigningInputData` built the THORChain swap signing input with a hardcoded `byteFee` of 1 sat/vByte. Its only caller, `THORChainSwaps.getPreSignedInputData`, immediately re-runs the builder through `getSigningInputData`, which re-applies the real per-byte fee from the keysign payload — so the hardcoded `1` was never used. The literal was a latent hazard: any future caller that built the swap input without that second pass would sign a THORChain swap at 1 sat/vByte, which the network rejects.

This sets the byte fee from the payload's UTXO `byteFee` directly — the same value the regular send path uses in `getBitcoinSigningInput`. Because `getSigningInputData` already applies that exact fee, the signed transaction is byte-for-byte identical to before; this only removes the misleading literal and makes the builder correct on its own.

Fixes #5032

## Which feature is affected?

## Checklist

N/A — signing-path logic, no UI surface.

## Additional context

No behavioural change: the per-byte fee that reaches the wire is unchanged (`getSigningInputData` already overwrote the placeholder with `keysignPayload`'s UTXO `byteFee`), so the produced transaction — and therefore every co-signer's pre-image hashes and the final broadcast — is byte-for-byte identical to `main`. This is why no swap tx link is attached: there is no observable on-chain difference to demonstrate.

Co-sign is unaffected: the initiator, peer devices, and VultiServer all rebuild the signing input from the shared keysign payload, and the byte fee is taken from that same payload in both the old and new code.

The added unit test asserts the swap pre-signing input carries the payload's `byteFee` (7) rather than a placeholder. It is JNI-gated like the sibling `UtxoHelper` swap/plan tests (it needs Trust Wallet Core native), so it runs where that library is present and is skipped otherwise.

## Summary by CodeRabbit

* **Bug Fixes**
  * Swap signing now uses the network fee value provided in the transaction payload, instead of a fixed placeholder amount.
  * Added test coverage to confirm swap signing respects the provided fee value.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelper.kt
@@ -72,13 +72,14 @@ class UtxoHelper(
         require(thorChainSwapPayload.data.vaultAddress.isNotEmpty()) {
             "Vault address is required for THORChain swap"
         }
+        val utxo = keysignPayload.blockChainSpecific as BlockChainSpecific.UTXO
         val input =
             Bitcoin.SigningInput.newBuilder()
                 .setHashType(BitcoinScript.hashTypeForCoin(coinType))
                 .setAmount(thorChainSwapPayload.data.fromAmount.toLong())
                 .setToAddress(thorChainSwapPayload.data.vaultAddress)
                 .setChangeAddress(keysignPayload.coin.address)
-                .setByteFee(1L)
+                .setByteFee(utxo.byteFee.toLong())
                 .setCoinType(coinType.value())
                 .setUseMaxAmount(false)
                 .setOutputOpReturn( // output index is the latest by default
--- a/data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelperTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelperTest.kt
@@ -235,26 +235,26 @@ class UtxoHelperTest {
         verifySwapOpReturn(memo)
     }
 
+    @Test
+    fun `getSwapPreSigningInputData - applies the payload byte fee, not a placeholder`() {
+        val helper = UtxoHelper(CoinType.BITCOIN, "", "")
+        val payload =
+            utxoPayload(
+                byteFee = BigInteger.valueOf(7L),
+                memo = "SWAP:BTC.BTC:$btcAddress:0:t:0:30",
+                swapPayload = thorChainSwapPayload(),
+            )
+        try {
+            val input = helper.getSwapPreSigningInputData(payload)
+            assertEquals(7L, input.byteFee)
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5037, opened 2026-06-24
fix(swap): unify Verify Swap consent gate between button and signing guard

### LINKED ISSUE #5036: [Fix] Unify Verify Swap consent gate between Continue button and signing guard [Swap/Verify]
# [Fix] Unify Verify Swap consent gate between Continue button and signing guard [Swap/Verify]

## Problem
On the Verify Swap screen, "all consents given" is encoded twice. The Continue button reads `VerifySwapUiModel.hasAllConsents`, which is `consentAmount && consentReceiveAmount && (consentAllowance || !hasConsentAllowance)`. The signing guard in `VerifySwapViewModel.keysign()` recomputes its own stricter expression `consentReceiveAmount && consentAmount && consentAllowance`, dropping the `|| !hasConsentAllowance` relaxation used for swaps that need no token approval. The two agree today only because `consentAllowance` is initialized to `!isApprovalRequired`, but they are independent expressions and will silently diverge the moment either side is edited.

## Expected Behavior
The signing guard and the Continue button enable on exactly the same condition. There is a single source of truth for consent completeness (`hasAllConsents`), mirroring the sibling Send verify screen, where `VerifyTransactionViewModel.keysign()` already reads `uiState.value.hasAllConsents`.

## Steps to Reproduce
This is a latent correctness risk rather than a currently user-visible bug:
1. Open a swap that requires no token approval (no allowance step).
2. The Continue button enables once amount and receive are confirmed (the allowance consent is implicitly satisfied).
3. The signing path re-derives its own consent condition instead of reusing the button's, so a future change to consent initialization or the no-approval case could enable the button while `keysign()` refuses (a dead-end), or the reverse.

## Solution
Have `keysign()` read `state.value.hasAllConsents` instead of recomputing a parallel condition, so the gate has one definition. This is the pattern the Send verify screen already uses. No behavior change for any state reachable today.

### Must Do
- [ ] Replace the recomputed consent condition in `VerifySwapViewModel.keysign()` with `state.value.hasAllConsents`.
- [ ] Add a focused unit test on the consent gate covering the no-approval relaxation.

### Must NOT Do
- Don't change `hasAllConsents` itself or the consent initialization (`consentAllowance = !isApprovalRequired`).
- Don't alter the swap signing payload, fee/amount math, or the security-scanner flow.
- Don't refactor unrelated parts of `VerifySwapViewModel`.

## Acceptance Criteria
- [ ] `./gradlew :app:testDebugUnitTest` passes.
- [ ] `keysign()` launches the keysign exactly when `hasAllConsents` is true, including the no-approval relaxation.
- [ ] The Continue button and the signing guard share one consent definition.

rkokhatskyi:
## Description

Fixes #5036

The Verify Swap screen encoded "all consents given" twice. The Continue button reads `VerifySwapUiModel.hasAllConsents` — `consentAmount && consentReceiveAmount && (consentAllowance || !hasConsentAllowance)` — while `keysign()` recomputed a stricter `consentReceiveAmount && consentAmount && consentAllowance`, dropping the no-approval relaxation. They agreed only because `consentAllowance` is initialized to `!isApprovalRequired`, so the two could silently diverge on a future edit: the button could enable while signing refuses (a dead-end), or the reverse.

`keysign()` now reads `state.value.hasAllConsents`, so the gate has one definition — the same pattern the sibling Send verify screen (`VerifyTransactionViewModel.keysign()`) already uses. No behavior change for any state reachable today.

## Which feature is affected?

No on-chain tx link: the gate is logically identical to the previous code for every state reachable today, so signing produces the same transactions as before — this only removes a second, divergence-prone copy of the condition.

## Checklist

N/A — no user-visible change; this consolidates an internal signing gate to a single source of truth.

## Additional context

Signing impact: `keysign()` only decides whether to launch the keysign; it does not touch payload construction, session/vault selection, or broadcast. Because the gate is unchanged for every reachable state, the Fast Sign (FastVault co-sign) and paired-device paths behave the same. `VerifySwapViewModelTest` exercises both the QR (`joinKeySign`) and biometric (`authFastSign`) entry points through the unified gate, plus the no-approval relaxation that the old parallel condition would have rejected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed swap verification consent gating to use the app’s overall consent-completeness state, ensuring keysigning proceeds when all required consents are satisfied (including correct handling of approval-required vs non-approval swaps).
* **Tests**
  * Added unit test coverage for consent-gated swap keysigning, validating behavior when consents are missing, when they’re complete for approval-required swaps, and for non-approval swaps.
  * Verified both standard signing and fast-sign flows under the same consent conditions.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
@@ -188,10 +188,7 @@ constructor(
     }
 
     private fun keysign(keysignInitType: KeysignInitType) {
-        val hasAllConsents =
-            state.value.let { it.consentReceiveAmount && it.consentAmount && it.consentAllowance }
-
-        if (hasAllConsents) {
+        if (state.value.hasAllConsents) {
             viewModelScope.launch {
                 launchKeysign(
                     keysignInitType,
--- a/app/src/test/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModelTest.kt
@@ -64,6 +64,7 @@ internal class VerifySwapViewModelTest {
     private lateinit var securityScannerService: SecurityScannerContract
     private lateinit var vaultRepository: VaultRepository
 
+    /** Sets up mocks and test dispatcher before each test. */
     @BeforeEach
     fun setUp() {
         Dispatchers.setMain(testDispatcher)
@@ -82,6 +83,7 @@ internal class VerifySwapViewModelTest {
         coEvery { vaultRepository.get(VAULT_ID) } returns mockk<Vault>(relaxed = true)
     }
 
+    /** Cleans up mocks and resets test dispatcher after each test. */
     @AfterEach
     fun tearDown() {
         unmockkStatic("androidx.navigation.SavedStateHandleKt")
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5063, opened 2026-06-27
Enlarge keygen/keysign QR card to fill screen width

aminsato:
## Summary

Follow-up to #5060 (merged). On the keygen/keysign Scan QR screen the QR was pinned to a fixed `185.dp`, leaving large empty margins inside the framed card. This enlarges the card and QR to the scaffold content width so the code consumes most of the screen width and scans more reliably from a second device — matching the iOS Pair screen.

- `QrCodeContainer` / `QrCodeImage` now stretch to `fillMaxWidth()` with a square `aspectRatio(1f)` instead of a fixed 185.dp footprint.
- The `@Preview` (`PeerDiscoveryScreenPreview`) previously passed `qr = null`, so the card rendered empty in Android Studio. It now encodes a QR inline (same quiet-zone margin) fed a production-sized keygen payload, so the preview shows a dense, representative code with the expand control — no emulator/Hilt context required.

## Before / After

| Before (fixed 185.dp) | After (full width) |
|--------|--------|
| before | after |

## Test plan

- [x] Rendered the real keygen peer-discovery screen on an emulator (dense QR): card fills the content width, expand button stays on the top-right corner clear of the finder pattern
- [x] Android Studio `PeerDiscoveryScreenPreview` renders the QR after Build & Refresh

## Summary by CodeRabbit

* **New Features**
  * Improved the peer discovery QR display so it scales more naturally across available space and stays centered during its fade-in animation.
  * Updated the preview to render a realistic QR code, making the screen preview closer to the actual app experience.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/peer/PeerDiscoveryScreen.kt
@@ -17,6 +17,7 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.aspectRatio
 import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #5069, opened 2026-06-28
feat(home): global TTL-based dismissal for promo banners (#5064)

### LINKED ISSUE #5064: Use global TTL-based dismissal for promo banners
## Problem
Promo banner dismissal is currently inconsistent and partly session-only / partly vault-scoped. The user experience should be global: if a user dismisses a CTA banner, switching vaults should not immediately surface the same CTA again.

Current Android paths checked:
- `FollowXBanner` / migration banners are hidden by `VaultAccountsViewModel.tempRemoveBanner()` via `isBannerVisible = false`, so dismissal is session/view-model scoped.
- `BuyVultBanner` is session-only below Silver via `buyVultSessionDismissed`.
- Once Silver is reached, `BuyVultBanner` persists a boolean under `buy_vult_banner_dismissed/<vaultId>`, so the storage is vault-based and has no expiry.

## Desired behavior
Use one global dismissal model per device/user, keyed by banner intent / CTA, not by vault. Store `dismissedAt` and re-show after the banner-specific TTL expires.

Suggested banner IDs / granularity:
- `upgrade_vault_dkls`
- `buy_vult_swap`
- `follow_x_vultisig`

Suggested TTLs are product decisions; implementation should make them configurable per banner rather than hard-coded into view logic.

## Acceptance checks
- Dismissing a promo banner in one vault does not show the same banner again just because the user switches vaults.
- Each banner can have its own TTL.
- Expired dismissals are ignored and the banner can show again.
- Existing `buy_vult_banner_dismissed/<vaultId>` data is migrated or treated safely.
- Add unit coverage for global dismissal, per-banner TTL, and expired vs active dismissal.

aminsato:
## Summary

Fixes #5064.

Promo-banner dismissal was inconsistent and partly session-only / partly vault-scoped, so dismissing a CTA banner could resurface it just by switching vaults:

- **Upgrade / Follow X** — in-memory `isBannerVisible`, so dismissal lasted only for the current visit.
- **Buy VULT below Silver** — session-only.
- **Buy VULT at Silver** — persisted **per vault** (`buy_vult_banner_dismissed/<vaultId>`) with **no expiry**.

This replaces all three with one **global, TTL-based** model keyed by banner intent (not by vault).

## Changes

- **`PromoBannerDismissalRepository`** (new) — stores a per-banner `dismissedAt` timestamp and reports a banner as dismissed only while `now - dismissedAt < ttl`. Includes a `Clock` seam for deterministic tests.
- **`PromoBanner` enum** — centralizes banner ids + per-banner TTLs so they're tunable in one place rather than hard-coded into view logic.
- **`VaultAccountsViewModel`** — derives `showUpgradeBanner` / `showFollowXBanner` / `showBuyVultBanner` from the repo (upgrade additionally requires a GG20, migration-eligible vault). Dropped the session flag, `tempRemoveBanner()`, and the Silver-tier dependency that only gated persistence.
- **Pager / screen** — each banner's close button now dismisses **that** banner independently (previously one button hid Upgrade + Follow X together).
- **Migration** — `MainDataModule` seeds the global `buy_vult_swap` dismissal if any vault had the legacy per-vault key dismissed, then removes the stale keys. Removed the now-dead per-vault methods from `VaultDataStoreRepository`.

## Acceptance criteria

- ✅ Dismissing in one vault does not resurface the banner on a vault switch (global key)
- ✅ Each banner has its own TTL
- ✅ Expired dismissals are ignored → banner shows again
- ✅ Legacy `buy_vult_banner_dismissed/<vaultId>` data migrated safely
- ✅ Unit coverage for global dismissal, per-banner TTL, expiry, and the GG20 gate

## Note for reviewers

The issue states the TTL values are a **product decision**. The durations in `PromoBanner` (Upgrade 14d / Buy VULT 7d / Follow X 30d) are placeholder defaults — happy to adjust to whatever product prefers; they live in one enum.

## Test plan

- `PromoBannerDismissalRepositoryTest`: not-dismissed, within-TTL, TTL-lapse, per-banner independence.
- `VaultAccountsViewModelTest`: dismiss delegation, show/hide by TTL, GG20 gating of the upgrade banner.
- `:data` + `:app` compile (incl. Hilt); ktfmt applied.

## Summary by CodeRabbit

* **New Features**
  * Promo banners now have independent, global visibility and dismissal for **Upgrade**, **Follow X**, and **Buy VULT**.
  * Dismissals are **TTL-based** and automatically reappear after the timeout.
* **Bug Fixes**
  * Banner visibility now updates correctly when switching vaults.
  * Upgrade banner eligibility is based on the selected vault’s type.
* **Chores**
  * Added data migration to move legacy Buy VULT dismissal state into the new TTL-based system.
* **Tests**
  * Updated and expanded unit tests for TTL dismissal and banner visibility rules.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -184,7 +184,7 @@ constructor(
     private fun collectBannerVisibility(vaultId: VaultId) {
         bannerJob?.cancel()
         bannerJob =
-            viewModelScope.launch {
+            viewModelScope.safeLaunch {
                 // Re-collected per vault so an expired TTL is re-evaluated on vault switch / home
                 // re-entry. The upgrade banner is migration-only (GG20); the others depend solely
                 // on
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -28,6 +28,8 @@ import com.vultisig.wallet.data.repositories.BalanceVisibilityRepository
 import com.vultisig.wallet.data.repositories.CryptoConnectionTypeRepository
 import com.vultisig.wallet.data.repositories.DefaultDeFiChainsRepository
 import com.vultisig.wallet.data.repositories.LastOpenedVaultRepository
+import com.vultisig.wallet.data.repositories.PromoBanner
+import com.vultisig.wallet.data.repositories.PromoBannerDismissalRepository
 import com.vultisig.wallet.data.repositories.RequestResultRepository
 import com.vultisig.wallet.data.repositories.TiersNFTRepository
 import com.vultisig.wallet.data.repositories.VaultDataStoreRepository
@@ -35,7 +37,6 @@ import com.vultisig.wallet.data.repositories.VaultRepository
 import com.vultisig.wallet.data.repositories.vault.VaultMetadataRepo
 import com.vultisig.wallet.data.services.PushNotificationManager
 import com.vultisig.wallet.data.usecases.EnableTokenUseCase
-import com.vultisig.wallet.data.usecases.GetDiscountBpsUseCase
 import com.vultisig.wallet.data.usecases.HasCircleAccountUseCase
 import com.vultisig.wallet.data.usecases.IsGlobalBackupReminderRequiredUseCase
 import com.vultisig.wallet.data.usecases.NeverShowGlobalBackupReminderUseCase
@@ -84,7 +85,6 @@ internal data class VaultAccountsUiModel(
     val showNotificationIntroSheet: Boolean = false,
     val showNotificationVaultSheet: Boolean = false,
     val notificationIntroVaults: List<VaultIntroItem> = emptyList(),
-    val showMigration: Boolean = false,
     val isRefreshing: Boolean = false,
     val totalFiatValue: String? = null,
     val totalDeFiValue: String? = null,
@@ -92,7 +92,11 @@ internal data class VaultAccountsUiModel(
     val accounts: List<AccountUiModel> = emptyList(),
     val defiAccounts: List<AccountUiModel> = emptyList(),
     val searchTextFieldState: TextFieldState = TextFieldState(),
-    val isBannerVisible: Boolean = true,
+    // Per-banner visibility, each gated on a global, TTL-based dismissal (#5064). The upgrade
+    // banner
+    // additionally requires the vault to be GG20 (migration-eligible).
+    val showUpgradeBanner: Boolean = false,
+    val showFollowXBanner: Boolean = false,
     val showBuyVultBanner: Boolean = false,
     val cryptoConnectionType: CryptoConnectionType = CryptoConnectionType.Wallet,
     val scanQrUiModel: ScanQrUiModel = ScanQrUiModel(),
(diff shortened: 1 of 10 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/PromoBannerDismissalRepository.kt
```diff
@@ -0,0 +1,70 @@
...
+ * vault keeps it hidden everywhere, and each banner carries its own [ttl]: once dismissed it stays
+ * hidden until the TTL elapses, then becomes eligible to show again (#5064).
+ *
+ * TTLs are product knobs and are centralized here rather than hard-coded into view logic, so they
+ * can be tuned per banner in one place.
+ */
+enum class PromoBanner(val id: String, val ttl: Duration) {
+    UpgradeVaultDkls(id = "upgrade_vault_dkls", ttl = 14.days),
+    BuyVultSwap(id = "buy_vult_swap", ttl = 7.days),
+    FollowXVultisig(id = "follow_x_vultisig", ttl = 30.days),
```
rkokhatskyi: `FollowXVultisig` ttl is `30.days` and `UpgradeVaultDkls` is `14.days`, but issue #5064's product decision sets every promo banner except `buy_vult_swap` to 15 days, so Follow X stays dismissed twice as long as intended

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/PromoBannerDismissalRepository.kt
aminsato: Good catch — fixed in 3055a016a. Set both `UpgradeVaultDkls` (was 14d) and `FollowXVultisig` (was 30d) to `15.days`, matching the product decision (everything except `buy_vult_swap` resurfaces after 15 days). `BuyVultSwap` stays at 7d.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -182,17 +181,37 @@ constructor(
         collectLastOpenedVault()
     }
 
-    private fun collectBuyVultBannerDismissed(vaultId: VaultId) {
-        buyVultBannerJob?.cancel()
-        buyVultBannerJob =
+    private fun collectBannerVisibility(vaultId: VaultId) {
+        bannerJob?.cancel()
+        bannerJob =
             viewModelScope.launch {
```
johnnyluo: **Use `safeLaunch` here — a deserializing repo call in a plain `viewModelScope.launch` can crash the app.**

This `launch` now runs `withContext(ioDispatcher) { vaultRepository.get(vaultId) }`, which resolves to `vaultDao.loadById(vaultId)?.toVault()` — a DB read plus entity deserialization that can throw (DB corruption, a `toVault()` deserialization failure). In a plain `viewModelScope.launch` that exception is uncaught and propagates to the default handler → crash on home-screen load / vault switch.

The `collectBuyVultBannerDismissed` this replaces only `combine`d flows (no throwing suspend call), so this risk is newly introduced. CLAUDE.md: *"Use `safeLaunch` instead of `viewModelScope.launch` for coroutines that perform network calls or deserialization."* `safeLaunch` is already used for this exact `vaultRepository.get` pattern in `collectLastOpenedVault` and in `dismissPromoBanner`.

Suggest switching `bannerJob = viewModelScope.launch { … }` to `viewModelScope.safeLaunch { … }`.

aminsato: Good catch — switched to `viewModelScope.safeLaunch` in 3bb372491. The `vaultRepository.get` DB read + `toVault()` deserialization can throw, and a plain `launch` would propagate to the default handler. Now consistent with the other usages of this call.

johnnyluo (approved):
LGTM

PR #5070, opened 2026-06-28
fix(swap): align route-not-available error tooltip under the flip button (#5062)

### LINKED ISSUE #5062: Swap: "Swap route not available" error tooltip is misaligned
## Description

On the **Swap** screen, when a swap route cannot be found, the **"Swap route not available"** error tooltip is rendered with incorrect alignment. The error box overlaps the **From** and **To** token rows and its caret/pointer does not line up with the element it is meant to annotate, leaving it floating between the two amount fields.

## Expected behavior

The error tooltip should be positioned cleanly relative to its anchor (e.g. below the swap inputs or attached to the relevant field) without overlapping the From/To rows, and its caret should point to the correct element.

## Actual behavior

The error box is horizontally/vertically misaligned and overlaps the swap input rows, partially hiding the "To" amount and CACAO balance.

## Steps to reproduce

1. Open the **Swap** screen
2. Select **From: Ethereum / UNI** and **To: Mayachain / CACAO (Native)**
3. Enter an amount that has no available route
4. Observe the "Swap route not available" error tooltip alignment

## Screenshot

swap error misalignment

aminsato:
## Summary

Fixes #5062.

On the Swap screen, the **"Swap route not available"** error tooltip rendered misaligned — floating over the From/To rows, hiding the red error circle and the "To" amount.

**Root cause:** the error `HintBox` is offset using the flip button's bottom-center. The flip button reports that position via `boundsInParent()` — relative to **its own parent** (the source-input `Box`) — while the hint is positioned in the **outer `Box`** coordinate space. The missing vertical offset of the content above the input block (the Market/Limit tabs + spacing) pushed the tooltip too high, so it covered the red circle and overlapped the rows.

**Fix:** track the input block's top-left in the outer `Box` space and add it to the flip-button anchor, translating it into the correct space. The tooltip now sits directly under the red error circle, with its caret aligned to it, and no longer overlaps the swap inputs.

## Changes
- `SwapScreen.kt` — translate the error-hint anchor from the flip button's parent space into the outer `Box` space.

## Before / After

| Before | After |
|--------|-------|
| before | after |

Real Android renders captured via `PreviewActivity` + ADB screencap on the emulator.

## Test plan
- [x] `./gradlew :app:assembleDebug` passes
- [x] ktfmt clean (`ktfmtCheckMain`)
- [x] Device-verified: tooltip sits under the red circle, caret aligned, both From/To amounts fully visible

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the placement of swap error hints so they appear correctly aligned under the relevant error indicator in Market mode.
  * Fixed hint positioning when the swap input area is laid out differently, making error messages easier to read and more consistent.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
@@ -30,6 +30,8 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.layout.boundsInParent
+import androidx.compose.ui.layout.onGloballyPositioned
 import androidx.compose.ui.platform.LocalFocusManager
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
@@ -187,7 +189,12 @@ internal fun SwapScreen(
             var bottomCenter by remember { mutableStateOf(Offset.Zero) }
             val space = 8.dp
 
+            // The flip button reports its bottom-center in its own parent's space, but the error
+            // hint is offset in the outer Box's space. Adding the input block's top-left translates
+            // the anchor into that outer space so the hint sits directly under the red error circle
+            // instead of floating too high and overlapping the From/To rows (#5062).
             var flipButtonBottomCenter by remember { mutableStateOf(Offset.Zero) }
+            var inputBlockTopLeft by remember { mutableStateOf(Offset.Zero) }
 
             val error = state.error ?: state.formError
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
```diff
@@ -187,7 +189,12 @@ internal fun SwapScreen(
             var bottomCenter by remember { mutableStateOf(Offset.Zero) }
             val space = 8.dp
 
+            // The flip button reports its bottom-center in its own parent's space, but the error
+            // hint is offset in the outer Box's space. Adding the input block's top-left translates
+            // the anchor into that outer space so the hint sits directly under the red error circle
+            // instead of floating too high and overlapping the From/To rows (#5062).
```
rkokhatskyi: Unnecessary comment, please remove

rkokhatskyi (approved):

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
aminsato: Done — removed the comment in 63ce5600a.

rkokhatskyi (approved):
LGTM

Issue #5041, opened 2026-06-25
[Parity] Native TON nominator-pool staking on the DeFi/Earn tab


gastonm5:
## [Parity] Native TON nominator-pool staking on the DeFi/Earn tab

iOS shipped native **TON nominator-pool staking** on the DeFi tab (vultisig-ios PR #4654 / issue #4653). This issue tracks parity on PLATFORM. The mechanics below were each verified against **live on-chain data** — several differ from common assumptions/docs, so please follow them exactly (a wrong comment or a non-bounceable send can lose user funds).

### Scope
Stake (deposit) + Unstake (withdraw) into TON **nominator pools**, shown on the DeFi/Earn tab with the pool (delegator) name + APR. NOT liquid staking (Tonstakers) — that's a separate, larger follow-up.

### Pool list / picker
- Source: tonapi `GET /v2/staking/pools`. Filter to genuine **nominator** implementations: `whales` and `tf`. **Exclude `liquidTF`** (Tonstakers — different mechanism, needs a jetton/op-code flow). In practice only **Whales** pools are accessible (min **50 TON**); `tf` pools need ~300k TON.
- Each pool entry: `address` (raw `0:` form), `name`, `apy` (percent), `min_stake` (nanotons), `verified`, `implementation`, `cycle_end`.

### Deposit / withdraw mechanism (CRITICAL — verified on-chain)
- Deposits/withdrawals are **TON transfers with a text comment**, but the comment word is **per pool implementation**:
  - `whales` → deposit comment **`"Deposit"`**, withdraw **`"Withdraw"`** (capitalization matters). NOTE: the tonwhales/ton-nominators README says `"Stake"` — that is **stale and rejected on-chain** (exit code 72). Use `"Deposit"`.
  - `tf` (standard `ton-blockchain/nominator-pool`) → deposit `"d"`, withdraw `"w"`.
  - Unknown implementation → block the action (don't guess a comment).
  - The op-code `0xa2065f2c` ("send_stake") seen on these pools is the **validator's cycle message**, NOT a user deposit — do not replicate it.
- **Send BOUNCEABLE.** Pool addresses arrive raw `0:`; convert to the **bounceable user-friendly `EQ…`** form for the destination. A non-bounceable deposit that the pool rejects is **absorbed (lost)** instead of bounced back. (TON docs: *"sending a message to the pool without a bounce flag may result in funds becoming locked."*)
[...]

Issue #4654, opened 2026-05-26
SwapKit Phase 2 hardening: sub-provider persistence, client deviation check, approvalTx fallback


aminsato:
# [Add] SwapKit Phase 2 hardening: sub-provider persistence, client deviation check, approvalTx fallback [Swap / SwapKit]

## Problem
Three deferred follow-ups from the #4619 round-2 review remain before/alongside Phase 2 widening the sub-provider surface. None of them block Phase 1 (feature flag default-off), but they harden the SwapKit integration: the persisted transaction loses the sub-provider label, output-amount deviation is only enforced server-side, and the ERC-20 approval spender has no fallback if `meta.approvalAddress` goes missing. These gaps matter more as Phase 2 adds sub-providers that touch user funds.

## Expected Behavior
- Done screen, history, and joined-device views show `via CHANGE` / `via CHAINFLIP` / `NEAR` / `GARDEN` (the live sub-provider) instead of a collapsed plain `"SwapKit"`.
- Android independently rejects a route whose buy amount deviates beyond tolerance, even if a buggy/compromised proxy skips the server-side check.
- Android can still resolve a correct approval spender if `meta.approvalAddress` is ever absent, by decoding it from `approvalTx.data`.

## Solution
Thread the sub-provider through the persisted swap payload so the canonical `"SwapKit"` provider is no longer the only stored signal, and add two defense-in-depth guards (a cheap client-side deviation check and a spender fallback decoded from approval calldata). This preserves the load-bearing "which third party touches user funds" signal across all persisted views and reduces reliance on trusted proxy infrastructure, without altering the default-off Phase 1 behavior.

### Must Do
- [ ] **Persist the sub-provider (review #4b):** thread the sub-provider through the persisted `EVMSwapPayloadJson` payload and `SwapTransactionToUiModelMapper`, so the done screen, history, and joined-device views display the sub-provider label instead of collapsed canonical `"SwapKit"`. (The cache-hit half is already fixed in #4619 via `SwapQuote.OneInch.subProvider`.)
- [ ] **Client-side deviation defense-in-depth (review #11):** using the route's existing `expectedBuyAmount` and `expectedBuyAmountMaxSlippage`, add a cheap client check — e.g. `if (buyAmount < buyAmountMaxSlippage * (1 - tolerance)) throw QuoteDeviation` — to complement the server-only `outputAmountDeviationTooHigh` enforcement.
[...]

PR #5073, opened 2026-06-28
feat(ton): nominator-pool staking — data + transaction core (#5041)

aminsato:
## Summary

First PR toward native **TON nominator-pool staking** on the DeFi/Earn tab (iOS parity — vultisig-ios #4654, issue #5041). This PR lands the **data layer** and the **fund-safe transaction-building core**. The pool picker, DeFi-tab positions screen, and chain gating land in a **follow-up** PR (which carries the before/after UI screenshots).

Scope was split this way deliberately: the transaction mechanics are fund-safety critical and verified on-chain, so they ship first with full unit coverage; the visual pieces ship next.

### Data layer
- `TonStakingApi` against `tonapi.io` — `/v2/staking/pools`, `/v2/staking/pool/{addr}`, `/v2/staking/nominator/{acct}/pools` + DTOs (these endpoints are served only by tonapi.io, not the Vultisig proxy).
- `TonDeFiBalanceService` — reads the **primary** pool's `amount + pending_deposit`, decorates with APY, persists via `StakingDetailsRepository`. Wired into DI and `BalanceRepository` (dispatch + cache key). It is **dormant** until the gating lands in the follow-up.

### Transaction core (mechanics verified on-chain)
- **Per-implementation comment:** `whales` → `Deposit`/`Withdraw`, `tf` → `d`/`w`. An unknown implementation **blocks** the action rather than guessing a comment.
- **Bounceable send:** the pool address is converted to the user-friendly bounceable `EQ…` form and passed as `dstAddress`, so the block-chain spec resolves `bounceable = true`. Previously `dstAddress` was never passed for stake/unstake, so the flag defaulted to `false` — a deposit a pool rejects would be **absorbed (lost)** instead of bounced back.
- **Amounts:** a deposit must clear `min_stake + ~1 TON` commission (depositing exactly the minimum is rejected); a withdraw carries the fixed **0.2 TON** signal fee (the pool returns the full staked balance).
- Removed the stale hardcoded `"d"/"w"` `DepositMemo` entries.

## Tests
- `TonNominatorPoolTest` (comment resolution per implementation, min-deposit math, fixed withdraw fee).
- Rewritten `StakeStrategyTest` / `UnstakeStrategyTest` — comment, bounceable destination, amount rules, and blocking of unsupported/unknown pools.
- Updated `DepositFormViewModelTest` for the new factory dependency.
- New error strings translated across all 10 locales.

## Follow-up (next PR)
- Pool picker on the stake flow (filtered to `whales`/`tf`, excludes `liquidTF`).
- TON DeFi-tab positions screen + chain gating (`isDeFiSupported` / `hasDeFiPositionsScreen` / `PositionTon` route), pending-withdrawal lock with unlock time, staked TON in the DeFi total.
- Before/after screenshots.

## Test plan
- [x] `:data:testDebugUnitTest` (TON staking tests) green
- [x] `:app:testDebugUnitTest` (Stake/Unstake/DepositForm tests) green
- [x] ktfmt clean on `:data` and `:app`
- [ ] Manual: stake into a Whales pool and unstake (covered by the follow-up UI PR)

## Summary by CodeRabbit

* **New Features**
  * Added TON staking support, including staking pool lookup, deposit/withdraw handling, and TON DeFi balance visibility.
  * Added support for supported pool types and minimum stake checks with clearer error messages.

* **Bug Fixes**
  * Improved TON transaction address handling and memo selection for staking actions.
  * Withdrawals now use the correct fixed fee and reject unsupported pools instead of proceeding with invalid data.

* **Documentation**
  * Expanded localized error text for TON staking across multiple languages.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/StakeStrategy.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/StakeStrategy.kt
@@ -2,16 +2,16 @@ package com.vultisig.wallet.ui.models.deposit.submit
 
 import androidx.compose.foundation.text.input.TextFieldState
 import com.vultisig.wallet.R
+import com.vultisig.wallet.data.api.chains.ton.TonStakingApi
+import com.vultisig.wallet.data.blockchain.ton.TonNominatorPool
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
-import com.vultisig.wallet.data.models.DepositMemo
 import com.vultisig.wallet.data.models.DepositTransaction
 import com.vultisig.wallet.data.models.EstimatedGasFee
 import com.vultisig.wallet.data.models.TokenValue
 import com.vultisig.wallet.data.repositories.AccountsRepository
 import com.vultisig.wallet.data.repositories.BlockChainSpecificAndUtxo
 import com.vultisig.wallet.data.repositories.BlockChainSpecificRepository
-import com.vultisig.wallet.data.repositories.ChainAccountAddressRepository
 import com.vultisig.wallet.ui.models.deposit.DepositFormUiModel
 import com.vultisig.wallet.ui.models.send.InvalidTransactionDataException
 import com.vultisig.wallet.ui.utils.UiText
@@ -20,7 +20,13 @@ import java.math.BigInteger
 import java.util.UUID
 import kotlinx.coroutines.flow.first
 
-/** Builds a Stake [DepositTransaction] for the TON chain. */
+/** Whether a TON nominator-pool transaction deposits into or withdraws from the pool. */
+internal enum class TonStakingAction {
+    DEPOSIT,
+    WITHDRAW,
+}
+
+/** Builds a Stake (deposit) [DepositTransaction] for a TON nominator pool. */
 internal class StakeStrategy(
     private val vaultIdProvider: () -> String?,
     private val chainProvider: () -> Chain?,
@@ -28,7 +34,8 @@ internal class StakeStrategy(
     private val nodeAddressFieldState: TextFieldState,
     private val tokenAmountFieldState: TextFieldState,
     private val accountsRepository: AccountsRepository,
-    private val chainAccountAddressRepository: ChainAccountAddressRepository,
+    private val tonStakingApi: TonStakingApi,
+    private val toBounceableAddress: (String) -> String,
     private val blockChainSpecificRepository: BlockChainSpecificRepository,
     private val calculateGasFee: suspend (Chain, Coin, String) -> TokenValue,
     private val getFeesFiatValue:
@@ -36,15 +43,16 @@ internal class StakeStrategy(
 ) : DepositSubmitStrategy {
 
     override suspend fun build(): DepositTransaction =
-        buildTonDepositTransaction(
-            memo = DepositMemo.Stake,
+        buildTonStakingTransaction(
+            action = TonStakingAction.DEPOSIT,
             vaultIdProvider = vaultIdProvider,
             chainProvider = chainProvider,
             stateProvider = stateProvider,
             nodeAddressFieldState = nodeAddressFieldState,
             tokenAmountFieldState = tokenAmountFieldState,
             accountsRepository = accountsRepository,
-            chainAccountAddressRepository = chainAccountAddressRepository,
+            tonStakingApi = tonStakingApi,
+            toBounceableAddress = toBounceableAddress,
             blockChainSpecificRepository = blockChainSpecificRepository,
             calculateGasFee = calculateGasFee,
             getFeesFiatValue = getFeesFiatValue,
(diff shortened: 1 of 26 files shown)


aminsato:
Thanks for the review — addressed in f8ecc2c.

**`getStakingPool` swallowing failures** — fixed. Replaced `runCatching {}.getOrNull()` with a targeted `catch (NetworkException)` that maps only **HTTP 404** (genuinely unknown pool) to `null` and rethrows everything else. Transient/timeout/5xx failures now propagate, and coroutine cancellation is no longer swallowed.

**Partial pool metadata** — the field that gates the action is `implementation`; when it's absent/unknown the comment resolver returns `null` and the strategy already blocks the deposit/withdraw. I left `min_stake` best-effort (null → 0): the pool enforces its real minimum on-chain, and since this PR sends **bounceable**, a below-minimum deposit bounces back rather than being lost — so it's a UX guard, not a fund-safety one, and I'd rather not over-block on an API field that drops out.

**`TonDeFiBalanceService` error handling** — fixed. It now catches `NetworkException` (which `bodyOrThrow` actually throws — `IOException` alone never caught the real failures) and **falls back to the cached stake** instead of returning an empty list that `BalanceRepository` would cache and hide the position. The APY lookup is isolated in its own best-effort try so a decoration failure can't drop a valid position.

**`persistStakedBalance` broad `catch`** — kept intentionally. It rethrows `CancellationException` and logs the rest, matching the existing `TronDeFiBalanceService` / `MayaDeFiBalanceService` / `CosmosStakingDeFiBalanceService` persistence pattern — a DB write failure shouldn't fail the balance fetch. Narrowing only this one service would make it the odd one out.

**Tests** — both addressed. The `getSpecific` mock now pins the exact bounceable `dstAddress` (`EQ_0:pool`) so the test fails if the strategy stops forwarding it; the unsupported-pool and below-minimum cases now assert the surfaced `UiText` resId. Added `TonDeFiBalanceServiceTest` covering staked-total math, APY-failure tolerance, the cache fallback, and the empty case.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/ton/TonStakingApi.kt
```diff
@@ -0,0 +1,124 @@
...
+            .pools
+
+    override suspend fun getStakingPool(poolAddress: String): TonStakingPoolInfoJson? =
+        runCatching {
+                http
+                    .get("$BASE_URL/v2/staking/pool/$poolAddress")
+                    .bodyOrThrow<TonStakingPoolResponseJson>()
+                    .pool
+            }
+            .getOrNull()
```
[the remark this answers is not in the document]
aminsato: Addressed in d0f65aa. A deposit now requires a positive `min_stake` — a missing/zero value is treated as drifted pool metadata and **blocks** the deposit (`ton_stake_error_unsupported_pool`) rather than silently using the 1 TON commission floor.

I enforced it at the deposit use-site in `StakeStrategy` instead of inside `getStakingPool`, because `getStakingPool` is also used to decorate the held position's APY (`TonDeFiBalanceService`), where `min_stake` is irrelevant — rejecting there would needlessly drop a valid position's APY. `implementation` is still the gate for both deposit and withdraw via the comment resolver. Added a covering test.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ton/TonDeFiBalanceService.kt
```diff
@@ -0,0 +1,117 @@
...
+        } catch (e: IOException) {
+            Timber.w(e, "TonDeFiBalanceService: Network error fetching nominator-pool balance")
+            getCacheDeFiBalance(address, vaultId)
+        }
+    }
+
+    /** Best-effort pool APY; a decoration failure must not drop an otherwise-valid position. */
+    private suspend fun fetchPoolApr(poolAddress: String): Double? =
+        try {
+            tonStakingApi.getStakingPool(poolAddress)?.apy
```
rkokhatskyi: `fetchPoolApr` persists tonapi `apy` raw, but `StakingDetails.apr` is a fraction that `Double.formatPercentage` multiplies by 100, so the TON APR renders 100x high once the DeFi screen reads it. iOS does `apy / 100` here

app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/StakeStrategy.kt
```diff
@@ -81,38 +93,73 @@ internal suspend fun buildTonDepositTransaction(
...
+                val minStakeNano =
+                    poolInfo.minStake?.takeIf { it > 0 }
+                        ?: throw InvalidTransactionDataException(
+                            UiText.StringResource(R.string.ton_stake_error_unsupported_pool)
+                        )
+                resolveDepositAmount(tokenAmountFieldState, selectedToken, minStakeNano)
+            }
+            // The withdraw message carries only the fixed signal fee; the entered amount is
+            // ignored.
+            TonStakingAction.WITHDRAW -> TonNominatorPool.WITHDRAW_FEE
```
rkokhatskyi, line 154: Withdraw ignores the entered amount and always sends `WITHDRAW_FEE`, yet `DepositFormScreen` still shows an editable amount field for `DepositOption.Unstake`, so a user typing a partial amount silently gets a full unstake. iOS shows no amount field for unstake

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/ton/TonStakingApi.kt
```diff
@@ -0,0 +1,132 @@
...
+     * participation returns an empty list.
+     */
+    suspend fun getNominatorPools(address: String): List<TonAccountStakingInfoJson>
+}
+
+internal class TonStakingApiImpl @Inject constructor(private val http: HttpClient) : TonStakingApi {
+
+    override suspend fun getStakingPools(): List<TonStakingPoolEntryJson> =
+        http
+            .get("$BASE_URL/v2/staking/pools") { parameter("include_unverified", false) }
```
rkokhatskyi, line 48: `include_unverified=false` does not restrict tonapi to verified pools and the decoded `verified` flag is never used to filter, so the follow-up picker would surface unverified pools where iOS filters on `verified` client-side

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ton/TonDeFiBalanceService.kt
aminsato: Fixed in 08cfef3. `fetchPoolApr` now scales the tonapi percentage into the fraction `StakingDetails.apr` expects (`apy / 100`), so `formatPercentage`'s ×100 renders correctly. Matches iOS. Added a test assertion (`apr == 0.05` for `apy = 5.0`) to lock it in.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/StakeStrategy.kt
aminsato: Fixed in 08cfef3. Removed `DepositOption.Unstake` from the amount-field show condition in `DepositFormScreen`, so unstake no longer shows an editable amount (it withdraws the full staked balance). `Stake`/`Unstake` are TON-only, so no other chain is affected. Matches iOS.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/ton/TonStakingApi.kt
aminsato: Fixed in 08cfef3. `getStakingPools()` now filters on the decoded `verified` flag client-side, since tonapi does not honor `include_unverified=false`. Mirrors the iOS client-side filter. (No caller yet — the picker is a follow-up — but the filter is in place for when it lands.)

rkokhatskyi (approved):
LGTM

PR #5075, opened 2026-06-28
feat(swap): add Price Impact row to fee breakdown (iOS parity) (#5018)

### LINKED ISSUE #5018: Swap: Android missing Price Impact and VULT (-20 bps) rows (iOS parity)
## Summary

On the **Swap** screen, the Android app does not show the **Price Impact** row, and the **VULT (-20 bps)** affiliate-discount row is also not displayed. iOS shows both rows in the fee breakdown for the same swap. This is a UI/feature parity gap between Android and iOS.

## Comparison

| Android (current) | iOS (expected) |
|--------|--------|
| android | ios |

Same swap on both platforms (Tron TRX → Ton, SwapKit provider).

**Android fee breakdown shows:**
- Provider
- Total Fee
- Network Fee
- Swap Fee

**iOS fee breakdown shows:**
- Provider
- Total Fee
- Network Fee
- **VULT (-20 bps)** → `-$0.03`
- **Price Impact** → `-1.33% (Average)`

## Missing on Android

1. **Price Impact row** — Android does not display price impact at all. iOS shows the percentage and a qualitative label (e.g. `-1.33% (Average)`).
2. **VULT (-20 bps) discount row** — iOS shows the VULT affiliate discount line with its fiat value; Android does not display it in this swap.

## Expected behavior

The Android swap quote breakdown should match iOS:
- Show a **Price Impact** row with the percentage and the qualitative tier label (e.g. Low / Average / High), styled to match Figma.
- Show the **VULT (-20 bps)** discount row with the bps amount and its negative fiat value when an affiliate discount applies.

## Notes for implementation

- The `FeeBreakdown` UI model (`app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormUiModel.kt`) currently has no price-impact field; one would need to be added and populated from the quote pipeline.
- The `SwapFeeBreakdown` composable (`app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/SwapFeeBreakdown.kt`) already renders a VULT tier discount row gated on `discountInfo.vultBpsDiscount`/`vultBpsDiscountFiatValue`, but it is not appearing for this swap — investigate why those values are not populated. There is no price-impact row in this composable.
- Verify final layout, ordering, colors and the qualitative label thresholds against Figma before implementing.

## Platform

- Android
- Screen: Swap (Market tab)

aminsato:
Fixes #5018

## Summary

Adds the **Price Impact** row to the swap fee breakdown, closing the iOS-parity gap reported in #5018.

While investigating, the **VULT (-20 bps) discount row was already implemented** on Android — so the only genuine gap was Price Impact. This PR adds just that.

- THORChain/Maya derive price impact from the node's `slippage_bps`; SwapKit from its route's `totalSlippageBps`. Both are exposed as a fractional `SwapQuote.priceImpact`. EVM aggregators (1inch/Kyber/LiFi/Jupiter) report none, so the row is hidden for them — matching iOS.
- The row shows a signed percentage and a **Good / Average / High** tier coloured green / amber / red, with bands matching iOS `SwapCryptoLogic` (`> -1%` Good, `> -3%` Average, else High).
- Cleared on quote reset so a stale value can't linger after an error or token flip.
- 4 new strings across all 10 locales (`nl`/`ru` hand-translated, the rest ported from iOS).

## Before / After

| Before | After |
|--------|-------|
| before | after |

*Before is the Android screenshot from #5018 (fee breakdown ends at Swap Fee, no Price Impact). After shows the new `+1.00% (Good)` row on a TRX → GRAM SwapKit (NEAR) route.*

## Test plan

- `PriceImpactTest` — formatting/negation/sign and Good/Average/High banding incl. the −1% and −3% boundaries.
- `SwapQuotePriceImpactTest` — THORChain/Maya `slippage_bps` → fraction derivation, null handling, and SwapKit pass-through.
- All swap unit suites in `:app` and `:data` pass; ktfmt clean.

## Summary by CodeRabbit

* **New Features**
  * Added price impact visibility in the expanded swap fee breakdown, showing a formatted percentage plus a qualitative tier (Good/Average/High) when available.
  * Extended swap quote handling to capture and propagate price impact through to the UI, including SwapKit and chain slippage-derived sources.
  * Added “Price Impact” UI strings and tier labels for supported languages.
* **Bug Fixes**
  * Cleared previously shown price impact when resetting swap quote/fee state to avoid stale display.
* **Tests**
  * Added unit tests for price impact formatting and quote-derived price impact behavior (including null and boundary cases).


app/src/test/java/com/vultisig/wallet/ui/models/swap/PriceImpactTest.kt
```diff
@@ -0,0 +1,67 @@
+package com.vultisig.wallet.ui.models.swap
+
+import java.math.BigDecimal
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+import org.junit.jupiter.api.Test
```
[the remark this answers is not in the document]
aminsato: Declining this one. `kotlin.test` is the dominant assertion style in this codebase — 71 files under `app/src/test` use it vs 32 using Kotest — and most neighbors in this very directory (`SwapFormViewModelTest`, `SwapGasCalculatorTest`, `SwapKitErrorMappingTest`, `SwapTransactionBuilderTest`, `SwapValidatorTest`) use `kotlin.test` or raw JUnit assertions. The direct sibling added in this same PR, `SwapQuotePriceImpactTest.kt`, also uses `kotlin.test.assertEquals`/`assertNull`. Converting only this file to Kotest would make it inconsistent with its own sibling and the local convention, so the assertions stay as-is.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -220,6 +220,10 @@ constructor(
             swapResponse.providers.firstOrNull()?.takeIf { it.isNotBlank() }
                 ?: swapResponse.meta.subProvider?.takeIf { it.isNotBlank() }
 
+        // Fractional price impact (e.g. 0.0133 == 1.33%) for the Price Impact row (iOS parity).
+        // SwapKit reports it on the chosen route as totalSlippageBps; map bps → fraction.
+        val priceImpact = best.totalSlippageBps?.let { BigDecimal.valueOf(it).movePointLeft(4) }
```
rkokhatskyi: iOS sources SwapKit price impact from `response.meta.priceImpact` (`SwapQuote.swift:185`), a field it models and selects separately from `totalSlippageBps`. This reads the route's `totalSlippageBps` instead, so the SwapKit value and Good/Average/High tier are not guaranteed to match iOS despite the stated parity

app/src/test/java/com/vultisig/wallet/ui/models/swap/PriceImpactTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        // displayImpact == -0.03 is NOT > -0.03, so the 3% boundary falls into High.
+        val display = formatPriceImpact(BigDecimal("0.03"))!!
+        assertEquals("-3.00%", display.percent)
+        assertEquals(PriceImpactLevel.HIGH, display.level)
+    }
+
+    @Test
+    fun `favorable impact renders an explicit plus sign and Good tier`() {
+        // A negative slippage (favorable) negates to a positive display value.
+        val display = formatPriceImpact(BigDecimal("-0.002"))!!
```
rkokhatskyi, line 66: Both production sources `slippage_bps` and `totalSlippageBps` are non-negative slippage magnitudes, so `impact` is never negative and a favorable `+x.xx%` cannot occur. This pins `-0.002`, an input the quote pipeline never emits

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Good catch — confirmed and fixed. iOS models `meta.priceImpact` and `totalSlippageBps` as two distinct fields and `SwapQuote.swift` deliberately reads `response.meta.priceImpact` (returning it directly as a fraction, no bps conversion). Our `SwapKitRouteMeta` didn't model `priceImpact` at all, so this was reading the wrong source field *and* applying a bps→fraction conversion that iOS doesn't.

Fixed in the latest commit:
- Added `@SerialName("priceImpact") val priceImpact: Double?` to `SwapKitRouteMeta`.
- `SwapKitQuoteSource` now reads `best.meta?.priceImpact?.let { BigDecimal.valueOf(it) }` directly (no `movePointLeft`), so the value and the Good/Average/High tier match iOS.
- Added decode coverage for the new `meta.priceImpact` field and refreshed the stale doc comments that referenced `totalSlippageBps`.

app/src/test/java/com/vultisig/wallet/ui/models/swap/PriceImpactTest.kt
aminsato: You were right for the code as it stood: with both sources being non-negative slippage magnitudes (`slippage_bps`, `totalSlippageBps`), the negative input was unreachable and the `+x.xx%` branch was dead.

The companion fix above changes that. SwapKit's `meta.priceImpact` is a *signed* directional impact (negative == a favorable trade), unlike the `totalSlippageBps` magnitude we used before — so a negative value into `formatPriceImpact` is now a real production case via the SwapKit path, and the `+` prefix branch is genuinely exercised. I've kept the test (it pins the sign-handling that mirrors iOS `SwapCryptoLogic`) and updated its comment to explain that the negative input now comes from SwapKit rather than being hypothetical.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Follow-up after device testing — heads up, reading `meta.priceImpact` **alone** regresses the feature.

On-device, a real SwapKit (NEAR) **BTC → TRX** route shows the Price Impact row as `+2.30% (Good)` when sourced from `totalSlippageBps`, but the row **disappears entirely** once I switch to `meta.priceImpact`. The Vultisig SwapKit proxy (`api.vultisig.com/swapkit-a/v3`) doesn't populate `meta.priceImpact` for those routes (NEAR Intents in particular), so it comes back null and the row is hidden — i.e. iOS would show nothing here either.

Two things this also surfaced:
1. `totalSlippageBps` is **signed** in practice (that BTC→TRX route returned a negative value → the favorable `+2.30%`), which is the opposite of the "non-negative magnitude" premise in your other comment — so the `+x.xx%` path is genuinely reachable in production, not hypothetical.
2. Dropping `totalSlippageBps` would lose a real, useful value for the routes users actually get.

So I've changed it to **prefer `meta.priceImpact` (your parity point — used as soon as the proxy provides it) and fall back to `totalSlippageBps`** when it's absent:

```kotlin
val priceImpact =
    best.meta?.priceImpact?.let { BigDecimal.valueOf(it) }
        ?: best.totalSlippageBps?.let { BigDecimal.valueOf(it).movePointLeft(4) }
```

This matches iOS exactly whenever `meta.priceImpact` is present, and keeps the row populated (current behavior) when it isn't — strictly better than removing the row. Before/after screenshots are in the PR thread. Let me know if you'd rather hard-match iOS and accept the row being hidden for NEAR routes instead.

app/src/test/java/com/vultisig/wallet/ui/models/swap/PriceImpactTest.kt
aminsato: Update with on-device evidence: the favorable `+x.xx%` case is **real in production**, not just a formatter edge. A live SwapKit (NEAR) BTC → TRX route returned a **negative** slippage and rendered `+2.30% (Good)` on screen — so SwapKit's slippage/impact is signed, contrary to the "non-negative magnitude" premise. The test stays (comment updated to cite the device case rather than a hypothetical input). See the companion thread on `SwapKitQuoteSource` for the related source fix.

aminsato:
@rkokhatskyi both review points are addressed in `15e4879c8`:

**1. SwapKit source field (`meta.priceImpact` vs `totalSlippageBps`)** — you're right that iOS reads `meta.priceImpact`, so I modelled that field on `SwapKitRouteMeta` and now prefer it. But on-device it turned out the Vultisig SwapKit proxy leaves `meta.priceImpact` **null** for NEAR Intents routes, so reading it alone hides the row entirely (a live BTC→TRX route lost its `+2.30%`). To keep iOS parity *without* regressing, it now prefers `meta.priceImpact` and falls back to `totalSlippageBps`:

```kotlin
val priceImpact =
    best.meta?.priceImpact?.let { BigDecimal.valueOf(it) }
        ?: best.totalSlippageBps?.let { BigDecimal.valueOf(it).movePointLeft(4) }
```

Matches iOS the moment the proxy supplies `meta.priceImpact`; keeps the row populated otherwise. Happy to hard-match iOS (and accept a hidden row on NEAR routes) instead if you'd prefer — just say the word.

**2. Favorable `+x.xx%` test** — that same BTC→TRX route returned a **negative** slippage and rendered `+2.30% (Good)` on-device, so SwapKit's value is signed and the `+` branch is reachable in production, not hypothetical. Kept the test and updated its comment to cite the device case.

Decode coverage for `meta.priceImpact` added; `:data` + `:app` price-impact suites green. Both inline threads have the detailed follow-ups.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
rkokhatskyi: Fix is right: `best.meta?.priceImpact` is read verbatim as a fraction, matching iOS `Decimal(impact)`, and only falls back to `totalSlippageBps/10000` when the proxy omits it. Keep the fallback rather than hard-matching iOS, showing the route's slippage for NEAR Intents is strictly better than hiding the row, and you still get exact iOS parity whenever `meta.priceImpact` is present. Confirmed the new field decodes (`SwapKitQuoteDecodingTest`) and the swap unit suites pass.

app/src/test/java/com/vultisig/wallet/ui/models/swap/PriceImpactTest.kt
rkokhatskyi: The signed `meta.priceImpact` makes the `+` branch a genuine production path now, so the test pins real sign-handling rather than an unreachable input. Good to keep it with the updated comment.

rkokhatskyi (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/api/models/quotes/SwapKitQuoteJson.kt
```diff
@@ -136,6 +136,13 @@ data class SwapKitWarning(
 data class SwapKitRouteMeta(
     @SerialName("assets") val assets: List<SwapKitMetaAsset> = emptyList(),
     @SerialName("tags") val tags: List<String> = emptyList(),
+    /**
+     * Signed fractional price impact for this route (e.g. `0.0133` == +1.33% cost, a negative value
+     * == a favorable trade). Distinct from [SwapKitRoute.totalSlippageBps]: that is a non-negative
+     * slippage magnitude in bps, whereas this is the directional impact the user actually sees. iOS
+     * reads this same field (`meta.priceImpact`) for the Price Impact row.
```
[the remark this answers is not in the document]
aminsato: Fixed in `c56b2d109`. Updated the `priceImpact` KDoc so it no longer claims `totalSlippageBps` is a non-negative magnitude — it's signed in practice (negative for favorable routes, as seen on the live BTC→TRX route), and the note now reflects that plus the fallback path.

PR #5076, opened 2026-06-28
fix(swap): correct network fee overestimation on EVM aggregator swaps

### LINKED ISSUE #5056: Swap network fee estimate ~5x the actual on-chain fee (same-chain EVM / SwapKit)
## Summary

On a **same-chain ETH → USDT** swap (routed via SwapKit), the swap estimate shows a **Network fee of $3.90 (0.00246998 ETH)**, but the actual on-chain fee was **$0.76 (0.000485817617031168 ETH)** — roughly **5× too high**. Overstating the gas fee on a money path is a trust issue and can scare users away from a swap that is actually cheap.

## Steps to reproduce

1. Swap ETH → USDT on the same chain (small amount; routes via SwapKit).
2. Read the **Network fee** on the swap estimate / Verify screen.
3. Sign & broadcast, then compare with the on-chain fee.

## Screenshots

Image

## Observed vs actual

- **Estimate shown:** Network fee **0.00246998 ETH (≈ $3.90)** (Swap fee ≈ $0.43)
- **Actual tx:** https://etherscan.io/tx/0xee70cd8bf8613f2fca58b31f1e21105802f3a99e663d3d73f3596b1f30a50c70
  - gasUsed = **235,392**
  - effectiveGasPrice = **2.064 Gwei**
  - tx gasLimit = **286,146**
  - **fee = 0.000485817617031168 ETH (≈ $0.76)**

5.08× overstatement.

## Analysis

The displayed network fee is `gasPrice × gas` in `SwapQuoteManager.fetchKyberQuote` (and the same in `fetchOneInchQuote` for 1inch/LiFi/Jupiter):

```kotlin
val gasFees =
    apiQuote.tx.gasPrice.toBigInteger() *
        (apiQuote.tx.gas.takeIf { it != 0L } ?: EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT).toBigInteger()
```

with `EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT = 600000`.

The 5.08× gap decomposes cleanly into two factors:

- **Gas units (the wrong, fixable part):** the numbers indicate the estimate used the flat **600,000** default (the aggregator's `tx.gas` was 0/unavailable at quote time) vs **235,392 actually used** — and even the tx itself was signed with only a **286,146** gas limit. So the display uses ~2× the gas limit the tx actually carries and ~2.5× the gas it uses.
- **Gas price (partly market drift):** quote-time gasPrice (≈ 4.12 Gwei, implied: 0.00246998 ÷ 600000) vs execution effectiveGasPrice 2.064 Gwei. `600000 × 4.116 Gwei = 0.00247 ETH` = the shown estimate.

The gas-price drift between quote and execution is partly expected; the **gas-units fallback to a flat 600k is clearly wrong** — the estimate should reflect the gas the tx is actually built/signed with, not a worst-case constant.

## Suspected fix

Base the displayed estimate on the same gas limit the swap tx is signed with (the aggregator's `tx.gas`, ≈ 286k here) instead of the flat 600k fallback — and ideally show an expected fee (closer to gasUsed) rather than a worst-case limit. The `gasPrice × (tx.gas ?: 600000)` pattern / 600k default appears in:

- `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt` ~L724 (Kyber), ~L788 (1inch/LiFi/Jupiter)
- `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt` ~L266
- `app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt` ~L230
- `data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/KyberQuoteSource.kt` ~L64
- `EvmHelper.DEFAULT_ETH_SWAP_GAS_UNIT = 600000` in `data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/EvmHelper.kt:33`

Worth checking iOS/Windows for parity (whether they show a tighter estimate based on the route's own gas).

## Impact

Same-chain EVM swaps (KyberSwap confirmed; 1inch / LiFi likely share the pattern) display a network fee several × higher than the user actually pays — a trust issue on a money path.

rkokhatskyi:
## Summary

The swap **Network fee** was computed as `maxFeePerGas × DEFAULT_SWAP_LIMIT (600000)` — a flat constant independent of the chosen route — so same-chain EVM aggregator swaps (SwapKit / KyberSwap / 1inch / LI.FI) displayed a network fee several times higher than the actual on-chain cost (#5056).

## What changed

- Re-base the displayed network fee onto the gas limit the swap tx is actually **signed** with — the aggregator route's `tx.gas`, floored by the same per-chain limit the signer applies (`getGasLimit` for native ETH/Arbitrum, otherwise the 600k default) — across all three surfaces:
  - initiator swap form — `SwapQuotePipeline.resolveNetworkFee` → new `SwapGasCalculator.rebaseEvmSwapNetworkFee`;
  - Verify screen / staged transaction — flows through the existing pass-through; the gas-limit-override fiat re-valuation now keys off the matched route-gas pair instead of the flat-600k baseline;
  - co-signer verify screen — `computeJoinKeysignSwapNetworkFee` now uses `maxFee × maxOf(routeGas, gasLimit)`, with the route gas carried in the EVM swap payload.
- Align the Ethereum swap base-fee bump with iOS: `×1.5 → ×1.1` (broadcast ceiling `baseFee × 1.32 + tip` instead of `×1.8`). The previous Ethereum-only bump priced the bond ~36% above iOS for the same route.
- **Signing and payload bytes are unchanged** — only the displayed/reserved fee changes. Non-EVM providers (Jupiter/Solana, UTXO, Cosmos) and THORChain/Maya native deposits are unaffected.
- The same co-sign under-display exists on iOS (its co-sign screen reads the stored gas-limit floor, not the signed route gas) — filed separately as vultisig/vultisig-ios#4680.

## Validation (real swap)

SwapKit ETH→USDT, tx [`0xd8b1…3df6`](https://etherscan.io/tx/0xd8b15a8ff399f06ccfac303908db849aa8e8d1542bd1f9e70036433397c33df6):

| | gas units | gas price | fee (ETH) | vs actual |
|---|---|---|---|---|
| Android estimate | 359,942 (signed limit) | 0.593 Gwei (maxFee) | 0.00021251 | 1.36× |
| Actual on-chain | 275,033 (gasUsed) | 0.568 Gwei (effective) | 0.00015631 | 1.0× |

Before the fix the same swap displayed ~5× the on-chain cost (the flat 600k limit). The residual 1.36× is the inherent bond headroom — the tx reserves more gas than it uses and the price ceiling sits a hair above the effective price — and is never *under*-displayed.

## Testing

- `./gradlew :data:testDebugUnitTest --tests "*EthereumFeeServiceTest"` — 45 passed (incl. new `baseFee × 1.32` ceiling case).
- `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.swap.*" --tests "*JoinKeysignSendGasFeeTest"` — all passed, incl. new `SwapGasCalculatorTest` rebase cases, `SwapQuotePipelineNetworkFeeTest`, co-signer route-gas cases, and the builder gas-override matched-pair case.
- `./gradlew :app:ktfmtCheck :data:ktfmtCheck` — clean.
- Manually verified on a funded vault (emulator + real broadcast); see the on-chain decomposition above.

## Before / After

| Swap fees before this PR | Swap fees after this PR |
|--------|-------|
| Image |  image |

Closes #5056

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved EVM (1inch) swap network fee and fiat display by rebasing estimates to the route’s actual gas usage.
  * Fixed gas-limit override fiat revaluation to use the matched route-gas estimate pair.
  * Updated join-swap fee calculation for EVM routes to support optional route-gas and keep consistent gas flooring behavior.
  * Refined Ethereum swap base-fee bump so stored and broadcast fee calculations remain aligned.
* **Tests**
  * Added regression coverage for route-gas rebasing (including skip/floor conditions) and added new pipeline-level network-fee tests; updated Ethereum fee expectations.


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapGasCalculator.kt
```diff
@@ -295,6 +296,52 @@ constructor(
...
+        val signedGasLimit =
+            maxOf(routeGas.toBigInteger(), getGasLimit(srcToken) ?: DEFAULT_SWAP_LIMIT)
+        if (signedGasLimit == DEFAULT_SWAP_LIMIT) return null
+
+        val nativeCoin =
+            withContext(Dispatchers.IO) { tokenRepository.getNativeToken(srcToken.chain.id) }
+        val rebasedFee =
+            TokenValue(
+                value = baselineGasFee.value * signedGasLimit / DEFAULT_SWAP_LIMIT,
+                token = nativeCoin,
```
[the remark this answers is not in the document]
rkokhatskyi: Good catch — fixed in 5ad72d386 so `rebaseEvmSwapNetworkFee` returns null for OP-stack L2 chains (`srcToken.chain.isOpStackL2`, i.e. Base/Optimism/Blast) since those are exactly the chains where `EthereumFeeService.addL1Amount` folds an L1 data fee into `baselineGasFee`, which the gas-limit ratio would otherwise scale, so rather than the heavier separate-and-recombine I skip the rebase there because those L2 fees are tiny so keeping the gas-pass estimate is acceptable, whereas non-OP-stack EVM (Ethereum/Arbitrum/etc.) carry no L1 component so the rescale stays exact, and I added a regression test (`rebaseEvmSwapNetworkFee skips OP-stack L2 chains to avoid scaling the L1 fee`).

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
```diff
@@ -492,6 +493,31 @@ internal class SwapQuotePipeline(
...
+            // before the quote landed. Re-base it onto the route gas so the estimate matches the
+            // signed/paid fee instead of a worst-case ~2x over-estimate (#5056). A route with no
+            // usable gas (e.g. Jupiter's Solana quotes) leaves the gas-pass value untouched.
+            val currentGasFee = gasFee?.takeIf { gasFeeChain == srcToken.chain }
+            val rebased =
+                currentGasFee?.let {
+                    swapGasCalculator.rebaseEvmSwapNetworkFee(
+                        srcToken = srcToken,
+                        baselineGasFee = it,
+                        routeGas = quote.data.tx.gas,
```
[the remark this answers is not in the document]
rkokhatskyi: Not applicable here — SwapKit EVM routes are materialized as `SwapQuote.OneInch`, not `SwapQuote.SwapKit`, since `SwapQuoteManager.fetchSwapKitQuote` builds a `SwapQuote.OneInch` from the `SwapQuoteResult.Evm` branch and only returns `SwapQuote.SwapKit` for the non-EVM `SwapQuoteResult.Native` routes (BTC/PSBT, TON, Solana) which carry no EVM `tx.gas`, so this branch already covers SwapKit's EVM aggregator routes, which I verified against a live SwapKit ETH→USDT swap where the rebase fired and the estimate matched the on-chain bond, whereas adding `SwapQuote.SwapKit` here wouldn't even compile against `quote.data.tx.gas` since its `data` is `SwapKitSwapPayloadJson` rather than the EVM quote envelope.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
```diff
@@ -278,20 +278,27 @@ constructor(
...
                         val overriddenFeeWei = gasLimit.toBigInteger() * specific.maxFeePerGasWei
                         val overriddenFiat =
                             gasFeeFiatValue.value
                                 .multiply(overriddenFeeWei.toBigDecimal())
-                                .divide(gasFee.value.toBigDecimal(), 10, RoundingMode.HALF_UP)
+                                .divide(
+                                    priceReferenceFee.value.toBigDecimal(),
+                                    10,
+                                    RoundingMode.HALF_UP,
+                                )
```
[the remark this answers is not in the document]
rkokhatskyi: Applied in 5ad72d386, with one correction to the suggested form: pairing `priceReferenceFee = estimatedNetworkFeeTokenValue ?: gasFee` with an independent `priceReferenceFiat = estimatedNetworkFeeFiatValue ?: gasFeeFiatValue` would itself mismatch when the token falls back to `gasFee` while `estimatedNetworkFeeFiatValue` is present (it would divide an estimated fiat by the baseline wei), so instead I pair the token and fiat together (`estimatedNetworkFeeTokenValue` with `estimatedNetworkFeeFiatValue`, else `gasFee` with `gasFeeFiatValue`) which keeps both the override price ratio and the staged fee on a single matched pair, and the existing OneInch-approve test already guards the token-null / fiat-present case.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
```diff
@@ -228,21 +227,13 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
...
-            // Ethereum (× 1.32 elsewhere). This is sized to survive base-fee spikes
-            // during the MPC review + sign window so the tx still lands before the
-            // deadline.
-            val bumpPercent = if (chain == Chain.Ethereum) 50 else 10
-            return baseNetworkPrice.increaseByPercent(bumpPercent)
+            // Swaps embed a short DEX deadline; the 10% base bump plus the 20% in
+            // calculateMaxFeePerGas gives a baseFee × 1.32 + priorityFee ceiling, matching iOS. The
+            // old Ethereum-only × 1.5 over-bumped the bond ~36% above iOS for no gain (issue
+            // #5056).
+            return baseNetworkPrice.increaseByPercent(SWAP_BASE_FEE_BUMP_PERCENT)
```
johnnyluo: **Please revert the Ethereum swap base-fee bump back to ×1.5 (×1.8 broadcast ceiling).**

This change lowers the committed `maxFeePerGas` for **Ethereum mainnet** swaps from `baseFee × 1.8 + priorityFee` to `baseFee × 1.32 + priorityFee` (the per-chain `if (chain == Chain.Ethereum) 50 else 10` is now a flat `10`). That ~36% smaller ceiling leaves only ~32% headroom for the base fee to climb during the MPC review + sign window.

In testing, **ETH → VULT swaps failed because the gas price was too low** — the signed `maxFeePerGas` no longer covered the network base fee by broadcast time, so the tx didn't land before the embedded DEX deadline. The previous `×1.8` ceiling was sized precisely to survive base-fee spikes during the multi-device sign window; the iOS-parity argument isn't worth re-introducing stuck/expired swaps on mainnet.

Suggest restoring the Ethereum-specific 50% bump here (keep the other-chain `10%` as-is). The network-fee **display** re-basing in the rest of this PR is unaffected by reverting this.

johnnyluo: Looks like 1inch swaps , the gas limit is too low ,

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignFeeUtils.kt
```diff
@@ -38,26 +38,26 @@ internal fun computeJoinKeysignNetworkFee(
...
 ): TokenValue =
     when (blockChainSpecific) {
-        is BlockChainSpecific.Ethereum ->
-            TokenValue(
-                value = blockChainSpecific.maxFeePerGasWei * EthereumFeeService.DEFAULT_SWAP_LIMIT,
-                token = nativeCoin,
-            )
+        is BlockChainSpecific.Ethereum -> {
+            val limit =
+                aggregatorRouteGas?.let { maxOf(it, blockChainSpecific.gasLimit) }
```
aminsato: **Cross-device fee mismatch on OP-stack L2s (Base/Optimism/Blast).**

`computeJoinKeysignSwapNetworkFee` rebases onto `maxOf(aggregatorRouteGas, gasLimit)` for *all* EVM chains, but the initiator side (`SwapGasCalculator.rebaseEvmSwapNetworkFee`) added an `if (srcToken.chain.isOpStackL2) return null` guard that keeps the L1-inclusive 600k baseline. The joiner has no such guard, so on an OP-stack native-ETH aggregator swap the joiner displays a fee based on the L2-only route gas (~286k) while the initiator shows the 600k baseline plus the L1 data fee. The two participants in the same keysign show materially different network fees, and the joiner understates the real cost by the L1 fee.

Mirror the `isOpStackL2` skip here so the joiner falls back to the signed `gasLimit` baseline on OP-stack chains, matching the initiator.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapGasCalculator.kt
```diff
@@ -295,6 +297,47 @@ constructor(
...
+        srcToken: Coin,
+        baselineGasFee: TokenValue,
+        routeGas: Long,
+    ): GasCalculationResult? {
+        if (routeGas <= 0L || baselineGasFee.value <= BigInteger.ZERO) return null
+        // Skip OP-stack L2s: their baseline folds in an L1 data fee the gas-limit ratio can't
+        // scale.
+        if (srcToken.chain.isOpStackL2) return null
+        val signedGasLimit =
+            maxOf(routeGas.toBigInteger(), getGasLimit(srcToken) ?: DEFAULT_SWAP_LIMIT)
```
aminsato: **ERC-20 gas-limit floor diverges from the joiner.**

The initiator floors route gas at `getGasLimit(srcToken) ?: DEFAULT_SWAP_LIMIT`, but `getGasLimit` returns non-null only for native ETH/ARB — for ERC-20 (and other EVM chains) it returns null, so the floor is 600k and `rebase` returns null whenever `routeGas <= 600k`, leaving the display at `maxFee × 600000`. The joiner (`computeJoinKeysignSwapNetworkFee`) instead floors at the signed `blockChainSpecific.gasLimit`, which for ERC-20 swaps is `max(600000, onchainEstimate × 1.5)` and can exceed 600k. For an ERC-20 swap whose estimate × 1.5 > 600k the two co-signers again display different fees for the identical tx. The two sides should floor on the same value.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
```diff
@@ -492,6 +497,23 @@ internal class SwapQuotePipeline(
                 networkFee = NetworkFeeUpdate.Clear
                 effectiveNetworkFeeTokenValue = null
             }
+        } else if (quote is SwapQuote.OneInch && srcToken.chain.standard == TokenStandard.EVM) {
+            // Re-base the gas-pass's flat estimate onto the EVM aggregator route's real gas
+            // (#5056);
+            // a route with no usable gas (e.g. Jupiter's Solana quotes) keeps the gas-pass value.
+            val currentGasFee = gasFee?.takeIf { gasFeeChain == srcToken.chain }
```
aminsato, line 508: **Stale fee not cleared in the EVM-aggregator branch.**

When `currentGasFee` is null because `gasFeeChain` lags `srcToken.chain` right after a token switch, `rebased` is null and the `if (rebased != null)` block is skipped — so `networkFee`/`effectiveNetworkFeeTokenValue` keep the previous chain's values. The sibling UTXO branch above (line ~497) explicitly sets `NetworkFeeUpdate.Clear` + `null` in the same situation. Without it, immediately after switching the source token to an EVM chain the form keeps showing and validating against the old chain's fee, which can wrongly enable/disable the swap. Add the same `Clear` for the null-rebase case.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
```diff
@@ -348,4 +335,43 @@ constructor(
...
+    ): Pair<TokenValue, FiatValue> {
+        val referenceFee = estimatedNetworkFeeTokenValue ?: gasFee
+        val referenceFiat =
+            if (estimatedNetworkFeeTokenValue != null)
+                estimatedNetworkFeeFiatValue ?: gasFeeFiatValue
+            else gasFeeFiatValue
+        if (
+            specific !is BlockChainSpecific.Ethereum ||
+                overrideGasLimit == null ||
+                referenceFee.value.signum() <= 0
```
aminsato, line 364: **Relocated positivity guard changes behavior when the estimate is a non-null zero.**

The guard moved from `gasFee.value.signum() > 0` to `referenceFee.value.signum() <= 0`, where `referenceFee = estimatedNetworkFeeTokenValue ?: gasFee`. When `estimatedNetworkFeeTokenValue` is non-null but zero while `gasFee` is positive, the old code applied the user gas-limit override; the new code short-circuits and returns the zero reference instead, so the confirm screen shows a 0 (or non-overridden) fee that can't match what the tx executes with. Worth guarding on `gasFee` (or asserting the estimate is never a non-null zero) to preserve the override path.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapGasCalculator.kt
```diff
@@ -295,6 +297,47 @@ constructor(
...
+        if (signedGasLimit == DEFAULT_SWAP_LIMIT) return null
+
+        val nativeCoin = tokenRepository.getNativeToken(srcToken.chain.id)
```
aminsato: **Redundant native-token lookup.**

`tokenRepository.getNativeToken(srcToken.chain.id)` re-fetches the native coin, but `baselineGasFee.token` is already that chain's native coin (set to `nativeCoin` in `calculateGasFee`). Reuse `baselineGasFee.token` to drop the extra per-refresh flow lookup; correctness is unaffected.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
rkokhatskyi: Reverted in eed1f14ac — restored the Ethereum-specific 50% bump (×1.8 ceiling) and kept other chains at 10%, since you're right that ×1.32 left too little headroom for the base fee to climb across the MPC sign window and caused stuck/expired ETH mainnet swaps, whereas the network-fee display re-basing (the actual #5056 fix) is independent and stays; I also added a regression test pinning the ×1.8 ceiling so this can't silently drop again.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignFeeUtils.kt
rkokhatskyi: Fixed in eed1f14ac — the joiner and the initiator now both go through one shared `evmSwapDisplayGasLimit(srcToken, routeGas)` which returns null for OP-stack L2s, so the joiner falls back to the same 600k baseline the initiator keeps and the two co-signers no longer disagree on OP-stack swaps.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapGasCalculator.kt
rkokhatskyi: Fixed in eed1f14ac — both sides now floor on the same value via that shared `evmSwapDisplayGasLimit` (route gas floored at `getGasLimit ?: 600k`) rather than the joiner flooring at `blockChainSpecific.gasLimit` while the initiator floors at 600k, so an ERC-20 swap displays the identical limit on both (600k when the route is ≤600k, the route gas above it).

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
rkokhatskyi: Fixed in eed1f14ac — the EVM branch now sets `NetworkFeeUpdate.Clear` + null when `currentGasFee` is null (gasFeeChain lagging the source chain after a token switch), mirroring the UTXO branch so the form stops showing and validating against the previous chain's fee; added a test for the lagging-chain case.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
rkokhatskyi: Fixed in eed1f14ac — the reference now ignores a non-null zero estimate via `estimatedNetworkFeeTokenValue?.takeIf { it.value.signum() > 0 }` and falls back to the positive `gasFee`/`gasFeeFiatValue` pair, so the override still applies and `repriceFee` never divides by zero; added a test that passes a zero estimate with an override and asserts the override fiat.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapGasCalculator.kt
rkokhatskyi: Done in eed1f14ac — dropped the `getNativeToken` lookup and reuse `baselineGasFee.token` (already the chain's native coin from `calculateGasFee`) via `baselineGasFee.copy(value = …)`.

johnnyluo (approved):
LGTM

PR #5091, opened 2026-06-29
feat(swap): VULT-scaled Jupiter affiliate fee + Solana routing (#5053)

### LINKED ISSUE #5053: [Swap] Add VULT-scaled affiliate fee to Android Jupiter + route native SOL via THORChain
# [Swap] Add VULT-scaled affiliate fee to Android Jupiter + route native SOL via THORChain

> See `00-shared-spec.md` for the cross-platform routing/fee spec.

## Context
Android already uses **Jupiter as its Solana swap provider** (and currently the *only*
Solana route), but with **zero affiliate fee** — the quote sends only
`inputMint/outputMint/amount/slippageBps` and the swap body has **no `feeAccount`**:

- `data/.../api/JupiterApi.kt` — `getSwapQuote()` (no `platformFeeBps`, no `feeAccount`)
- `data/.../repositories/swap/JupiterQuoteSource.kt` — quote source
- `app/.../ui/models/swap/SwapQuoteManager.kt:549-558` — `providerPriority` (Jupiter = lowest, Solana-only)

Every other provider charges a VULT-scaled 50 bps (`GetDiscountBpsUseCase`,
`THORChainSwaps`, `KyberApi`, `OneInchApi`, `LiFiChainApi`). We want Jupiter to match, and
native SOL cross-chain to prefer THORChain (consistency with iOS/SDK).

## Scope
1. **`JupiterApi.kt`** — add affiliate fee:
   - Add `affiliateBps: Int?` (and the output mint) to `getSwapQuote(...)`.
   - Quote: `parameter("platformFeeBps", affiliateBps)` when > 0.
   - Swap body: add `put("feeAccount", <ATA(feeOwner, outputMint)>)`.
2. **`JupiterQuoteSource.kt`** — pass the VULT-scaled bps through. Compute via
   `GetDiscountBpsUseCase` → `affiliateBps = max(0, 50 − discountBps)` (same as Kyber/1inch/LiFi).
3. **Routing** (`SwapQuoteManager.kt` `providerPriority`):
   - Native SOL ↔ other chain: **THORChain first** (confirm THORChain SOL support is wired
     on Android; the SDK enables it), then SwapKit/LiFi.
   - SOL ↔ SPL / SPL ↔ SPL: **Jupiter first**, LiFi fallback.
   - SPL ↔ other chain: THORChain (if supported) → SwapKit → LiFi (Jupiter not eligible).
   - Keep Jupiter gated to same-chain Solana only.
4. **feeAccount derivation** — ATA of `(owner = fee wallet, mint = output mint)`. Use
   wallet-core / Solana SPL ATA derivation. Create-if-missing or pre-provision common mints.

## Fee account
Fee owner wallet: `8iqhrtBzMcYLR6c6FkzeoMHibedYDkHvLKnX2ArNie5z`.
Not using Jupiter's on-chain referral program (we keep 100% of the fee).

## Files
- `data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt`
- `data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/JupiterQuoteSource.kt`
- `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt`
- `app/.../data/usecases/GetDiscountBpsUseCase.kt` (reuse)

## Tests
- Jupiter quote includes `platformFeeBps` = VULT-scaled bps; swap includes `feeAccount`.
- 0-discount user → 50 bps; tiered user → reduced; never negative.
- Native SOL→BTC/ETH routes via THORChain, not Jupiter.
- SOL↔SPL / SPL↔SPL still route via Jupiter.

## Acceptance
See `00-shared-spec.md`.

---

# Shared spec (cross-platform)

# Jupiter / Solana swap alignment — shared spec (read first)

This is the common spec referenced by all four platform issues. Goal: make Solana
swap routing **identical across SDK (Windows/extension), iOS, and Android**, route
on-Solana token swaps through Jupiter (no aggregator markup), keep native SOL
cross-chain on THORChain, and charge a **VULT-scaled affiliate fee on Jupiter**.

## Routing rules (apply on every platform)

| Swap shape | Provider preference order |
|---|---|
| **Native SOL ↔ other chain** (cross-chain) | THORChain → SwapKit → LiFi |
| **SOL ↔ SPL or SPL ↔ SPL** (stays on Solana) | **Jupiter** → LiFi (fallback) |
| **SPL token ↔ other chain** (cross-chain) | THORChain (if pair supported) → SwapKit → LiFi |

Hard constraint: **Jupiter is Solana-only and same-chain** — it cannot do any
cross-chain route. Only offer Jupiter when `from.chain == Solana && to.chain == Solana`.

## Affiliate fee on Jupiter

Jupiter the aggregator takes **0%**; the only fee charged is the one we add.

- **Quote** (`GET .../swap/v1/quote`): add `platformFeeBps=<affiliateBps>`.
- **Swap** (`POST .../swap/v1/swap`): add `feeAccount=<our fee ATA for the fee mint>`.
- Fee-mint constraint (ExactIn): the fee mint must be the **input or output** mint.
  **Use the OUTPUT mint** as the fee mint. (ExactOut would force the input mint —
  we run ExactIn, so output-mint is fine.)

### Fee account

- **Fee owner wallet (Solana):** `8iqhrtBzMcYLR6c6FkzeoMHibedYDkHvLKnX2ArNie5z`
- `feeAccount` = the **Associated Token Account (ATA)** of `(owner = fee wallet, mint = output mint)`.
  Derive per output mint: `getAssociatedTokenAddressSync(outputMint, feeOwner)`.
- **Provisioning decision: create-if-missing, per swap, per asset.** Each swap prepends an
  **idempotent** create-ATA instruction (`createAssociatedTokenAccountIdempotent`) for the
  fee ATA (output mint, owner = fee wallet). No pre-provisioning. We are **not** using the
  Jupiter on-chain Referral Program (so Jupiter keeps 0% of our fee), just a self-owned ATA.
  - ⚠️ **Jupiter does NOT auto-create the `feeAccount` ATA.** Its `/swap` endpoint auto-creates
    the *user's output* ATA, but the **fee account is a precondition** you must supply already
    initialized. So we MUST prepend our own `createAssociatedTokenAccountIdempotent` for the
    fee ATA. There is no request flag for this. (Confirmed against Jupiter docs.)
  - ⚠️ Rent (~0.002 SOL) is paid by whoever signs the create instruction. Prepended to the
    user's swap tx → **the user pays** (first time per fee mint). If we'd rather absorb it,
    create the fee ATA in a separate Vultisig-funded tx (one-time per mint) instead of per
    swap. Default per this ticket = create-if-missing in the user's tx. Implementer should
    inspect the returned tx instructions to confirm the rent-payer.

### Fee consolidation (USDC) — NOT possible via the fee mechanism
Jupiter requires the **fee mint to be one of the swap-pair mints** (input or output for
ExactIn). So you **cannot** point every swap's `feeAccount` at a single USDC account — for a
pair where neither leg is USDC (e.g. SOL→BONK) the fee can only land in SOL or BONK.
Therefore fees accrue across **many mints** (the output mint of each swap). If consolidating
to USDC is desired later, that's a **separate periodic sweep/swap job**, independent of the
swap fee mechanism — not a config flag. (Confirmed against Jupiter docs.)

## Fee scaling by VULT holding (reuse existing logic — do NOT reinvent)

Every other provider already does: `affiliateBps = max(0, baseBps(50) − vultTierDiscountBps)`.
Apply the **same** value to Jupiter's `platformFeeBps`.

- VULT tiers / discount bps (SDK `swap/affiliate/config.ts`): base `50`; tier discount
  Bronze 5 / Silver 10 / Gold 20 / Platinum 25 / Diamond 35 / Ultimate 50 (NFT = +1 tier).
- SDK: `getSwapAffiliateBps(vultDiscountTier)`.
- Android: `GetDiscountBpsUseCase` → `max(0, 50 − discount)`.
- iOS: `bps(for: vultTierDiscount)` pattern (see `LiFiService` 141-142).

## API base (already proxied — keep using the proxy)

- Quote: `https://api.vultisig.com/jup/swap/v1/quote`
- Swap:  `https://api.vultisig.com/jup/swap/v1/swap`
- Params: `inputMint`, `outputMint`, `amount` (raw atomic units, no decimals),
  `slippageBps` (default 50), **`platformFeeBps` (new)**.
- Swap body: `quoteResponse`, `userPublicKey`, compute-unit config, **`feeAccount` (new)**.
- NOTE: legacy `quote-api.jup.ag/v6` was deprecated 2025-10-01; we already use `swap/v1`.

## Acceptance (all platforms)

- [ ] Native SOL cross-chain quotes come from THORChain first.
- [ ] SOL↔SPL and SPL↔SPL quotes come from Jupiter, LiFi only as fallback.
- [ ] Jupiter quote request includes `platformFeeBps` = VULT-scaled affiliate bps.
- [ ] Jupiter swap request includes `feeAccount` = ATA(feeOwner, outputMint).
- [ ] VULT holder sees reduced bps; 0-discount user sees 50 bps; never negative.
- [ ] Cross-chain-from-Solana never selects Jupiter.
- [ ] Behaviour matches the other two platforms for the same pair.

aminsato:
## Summary
Brings the VULT-scaled affiliate fee to Android's Jupiter Solana swaps and aligns Solana routing with the cross-platform spec (SDK is the reference; iOS not yet implemented). Fixes #5053.

- **Affiliate fee** — the Jupiter quote sends `platformFeeBps = max(0, 50 − VULT tier discount)`; the swap credits the fee (taken in the output mint) to a Vultisig-owned ATA. Jupiter itself takes 0% and we use no on-chain Referral Program, so we keep 100%.
- **Fee-ATA provisioning** — Jupiter's `/swap` auto-creates the *user's* output ATA but never the fee ATA, so a `feeAccount` that doesn't exist makes the tx fail on submit. We prepend an idempotent `createAssociatedTokenAccount` via wallet-core `SolanaTransaction.insertInstruction` (resolving Token vs Token-2022 from the mint owner). The user pays the one-time ~0.002 SOL rent the first time per output mint; it is a no-op thereafter. The co-signer signs the same baked transaction, so MPC byte-parity holds.
- **Routing** — Jupiter is Solana-only and same-chain, so it is excluded from cross-chain pairs by construction; it is now preferred over LI.FI/SwapKit for on-Solana pairs.

## Routing
| Swap shape | Preference order |
|---|---|
| Native SOL ↔ other chain | THORChain → SwapKit → LI.FI |
| SOL ↔ SPL / SPL ↔ SPL (stays on Solana) | **Jupiter** → LI.FI |
| SPL ↔ other chain | THORChain (if supported) → SwapKit → LI.FI |

## On-chain proof (mainnet)
A real SOL → USDC swap executed via Jupiter with the fee collected:
https://solscan.io/tx/4xiCLYvoPWRU4V35UBGFKZWsa8mVxgppjugD2bD7s2vfBzskZ7sSwvJDugsqPgYvqqyzxXVtCpczmUDSYohChTVK

- transaction succeeded, routed through the Jupiter program
- the prepended create-idempotent-ATA executed: fee ATA `tigMQDjwCNAzNndtiX93ZK1p71XaKTTRrQ8mfyp39LS` (owner `8iqhrt…`) went from non-existent to created and credited the affiliate fee

## Test plan
- `JupiterApiTest` — quote carries `platformFeeBps`; swap carries `feeAccount`; **no** fee account when Jupiter floors the amount to 0; quote fails on an unresolvable mint
- `JupiterFeeAtaInstructionTest` — pins the SPL create-idempotent-ATA layout (account order/flags, programId, `data` byte), verified against `@solana/spl-token`
- `SwapProviderTableTest` — Jupiter eligible on-Solana, excluded cross-chain; THORChain kept for native SOL
- `SwapQuoteManagerTest` — Jupiter outranks SwapKit + LI.FI for an on-Solana pair
- All suites green offline; ktfmt clean

## Known limitation
The displayed Solana network-fee estimate does not yet include the one-time fee-ATA rent (~0.002 SOL, first swap per output mint); the SDK adds `ataRentLamports`. The user pays the correct amount on-chain — only the pre-swap estimate is slightly low. Deferred as a fee-display follow-up.

## Summary by CodeRabbit

* **New Features**
  * Expanded swap quote routing to recognize Jupiter, including affiliate/platform fee handling and automatic fee-ATA setup when a non-zero platform fee is quoted.
  * Updated provider selection priority so Jupiter can be preferred for on-Solana routes.
* **Bug Fixes**
  * Avoid unnecessary fee-ATA actions when Jupiter floors the platform fee to zero.
  * Adjusted swap fee reservation to include the Solana fee-ATA rent-exempt minimum for native-SOL scenarios.
* **Tests**
  * Added/updated unit and API tests covering Jupiter fee handling, fee-ATA instruction generation, eligibility rules, and quote tie-breaking.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -453,10 +453,19 @@ constructor(
         // the source balance here — that would mix decimals and could wrongly drive the usable
         // amount negative for a low-decimal source into a high-decimal destination.
         val reservedNetworkFee =
-            if (percentage >= 1f) {
-                quotePipeline.estimatedNetworkFeeTokenValue.value?.value?.takeIf {
-                    srcToken.isNativeToken && quotePipeline.gasFeeChain.value == srcToken.chain
-                } ?: BigInteger.ZERO
+            if (
+                percentage >= 1f &&
+                    srcToken.isNativeToken &&
+                    quotePipeline.gasFeeChain.value == srcToken.chain
+            ) {
+                val baseFee =
+                    quotePipeline.estimatedNetworkFeeTokenValue.value?.value ?: BigInteger.ZERO
+                // A Jupiter affiliate swap may create the destination-mint fee ATA on first use,
+                // whose ~0.002 SOL rent the payer (this wallet) funds but the network-fee estimate
+                // doesn't include. Reserve it on native-SOL MAX so the first such swap can't
+                // underfund and fail on submit; it leaves harmless dust when no ATA is created.
+                if (srcToken.chain == Chain.Solana) baseFee + SOLANA_FEE_ATA_RENT_RESERVE
+                else baseFee
             } else {
                 BigInteger.ZERO
             }
@@ -644,6 +653,11 @@ constructor(
 
         // Upper bound for slippage tolerance: 10_000 bps = 100%.
         private const val MAX_SLIPPAGE_BPS = 10_000
+
+        // Rent-exempt minimum for an SPL token account (~0.00203928 SOL, in lamports). Held back on
+        // native-SOL MAX swaps to cover a first-use Jupiter fee-ATA creation the fee estimate
+        // omits.
+        private val SOLANA_FEE_ATA_RENT_RESERVE = BigInteger.valueOf(2_039_280)
 
         // Grouped, up-to-8-decimal $VULT amount (e.g. "3,000", "6.65648001"); truncates rather than
         // rounds up so a displayed balance never overstates what the vault holds.
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -464,7 +464,11 @@ constructor(
                 // whose ~0.002 SOL rent the payer (this wallet) funds but the network-fee estimate
                 // doesn't include. Reserve it on native-SOL MAX so the first such swap can't
                 // underfund and fail on submit; it leaves harmless dust when no ATA is created.
-                if (srcToken.chain == Chain.Solana) baseFee + SOLANA_FEE_ATA_RENT_RESERVE
+                // Jupiter only handles same-chain Solana, so a SOL→non-Solana MAX (routed via
+                // THORChain/LI.FI) never creates a fee ATA — don't reserve dust for it.
+                val dstChain = selectedDst.value?.account?.token?.chain
+                if (srcToken.chain == Chain.Solana && dstChain == Chain.Solana)
+                    baseFee + SOLANA_FEE_ATA_RENT_RESERVE
                 else baseFee
             } else {
                 BigInteger.ZERO
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -562,19 +562,24 @@ constructor(
         }
 
     /**
-     * Provider preference order for the banded selection. Lower index = preferred. THORChain is
-     * most preferred, then MayaChain, SwapKit, KyberSwap, 1inch, LI.FI; Jupiter (Solana-only,
-     * rarely competing in the same candidate set) ranks last.
+     * Provider preference order for the banded selection. Lower index = preferred. THORChain then
+     * MayaChain lead for cross-chain. Jupiter sits next, above SwapKit/Kyber/1inch/LI.FI: it is
+     * Solana-only and same-chain, so the only candidate set it ever appears in is an on-Solana
+     * SOL↔SPL / SPL↔SPL pair (against LI.FI and a SwapKit/NEAR cross-chain route), where the spec
+     * wants Jupiter — the native Solana DEX aggregator, no markup — preferred. Ranking it above
+     * those globally is safe because it never competes outside that set, and it never co-occurs
+     * with THORChain/Maya (those need a non-Solana or native-SOL leg Jupiter can't be eligible
+     * for).
      */
     private fun providerPriority(provider: SwapProvider): Int =
         when (provider) {
             SwapProvider.THORCHAIN -> 0
             SwapProvider.MAYA -> 1
-            SwapProvider.SWAPKIT -> 2
-            SwapProvider.KYBER -> 3
-            SwapProvider.ONEINCH -> 4
-            SwapProvider.LIFI -> 5
-            SwapProvider.JUPITER -> 6
+            SwapProvider.JUPITER -> 2
+            SwapProvider.SWAPKIT -> 3
+            SwapProvider.KYBER -> 4
+            SwapProvider.ONEINCH -> 5
+            SwapProvider.LIFI -> 6
         }
 
     /**
@@ -893,6 +898,14 @@ constructor(
                                     dstToken = dstToken,
                                     tokenValue = tokenValue,
                                     srcAddress = src.address.address,
+                                    // VULT-scaled affiliate fee (parity with Kyber/1inch/LiFi):
+                                    // 50 bps base, reduced by the holder's tier discount, never
+                                    // negative. Jupiter takes 0% itself, so this is the only fee.
+                                    affiliateBps =
+                                        maxOf(
+                                            0,
+                                            JUPITER_AFFILIATE_FEE_BPS - (vultBPSDiscount ?: 0),
+                                        ),
                                     slippageBps = slippageBps,
                                 ),
                             )
@@ -1272,6 +1285,10 @@ constructor(
         private const val KYBER_AFFILIATE_FEE_BPS = 50
         /** Base bps before tier discount; clamped to SwapKit's documented 0..1000 range. */
         private const val SWAPKIT_AFFILIATE_FEE_BPS = 50
+        /**
+         * Base bps before tier discount for Jupiter's `platformFeeBps` (parity with the others).
+         */
+        private const val JUPITER_AFFILIATE_FEE_BPS = 50
         private const val NATIVE_TOKEN_SENTINEL = "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
         private const val QUOTE_FETCH_TIMEOUT_MS = 15_000L
         /**
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
@@ -75,8 +75,15 @@ constructor(
                 feeAtaService.resolveFeeAccount(toToken)
             else null
 
+        // When Jupiter floored the fee to 0 we resolve no `feeAccount`. Round-tripping the quote's
+        // `platformFee` back into `/swap` without a matching `feeAccount` makes Jupiter reject the
+        // fee-bearing swap, which would drop the now-preferred Jupiter route to a worse provider,
+        // so
+        // strip it. With no fee requested `platformFee` is already null, so this is a no-op there.
+        val quoteResponseForSwap = if (feeAccount == null) body.copy(platformFee = null) else body
+
         val quoteSwapRequestBody = buildJsonObject {
-            put("quoteResponse", json.encodeToJsonElement(body))
+            put("quoteResponse", json.encodeToJsonElement(quoteResponseForSwap))
             put("userPublicKey", fromAddress)
             put("dynamicComputeUnitLimit", true)
             if (feeAccount != null) put("feeAccount", feeAccount.feeAccount)
@@ -100,14 +107,24 @@ constructor(
         }
         val quoteSwapData = swapResponse.bodyOrThrow<QuoteSwapTransactionJson>()
 
-        // Jupiter never initializes the `feeAccount` ATA, so prepend an idempotent create for it.
-        // The user pays the ~0.002 SOL rent the first time per fee mint; it is a no-op thereafter.
-        // The co-signer signs this exact transaction from the keysign payload, so byte-parity
-        // holds. Note: the displayed Solana network-fee estimate does not yet include this one-time
-        // rent (the SDK adds `ataRentLamports`); that is a deferred fee-display refinement.
+        // Jupiter never initializes the `feeAccount` ATA, so prepend an idempotent create for it
+        // the
+        // first time per fee mint (the payer funds the ~0.002 SOL rent; it is skipped once the ATA
+        // exists, see `needsCreate`). The co-signer signs this exact transaction from the keysign
+        // payload, so byte-parity holds.
         val swapTxData =
-            if (feeAccount != null)
-                feeAtaService.prependCreateFeeAta(quoteSwapData.data, feeAccount, fromAddress)
+            if (feeAccount != null && feeAccount.needsCreate)
+                feeAtaService
+                    .prependCreateFeeAta(quoteSwapData.data, feeAccount, fromAddress)
+                    // Jupiter sized `SetComputeUnitLimit` from its `dynamicComputeUnitLimit`
+                    // simulation, which ran before this create-ATA existed, so the baked limit does
+                    // not budget for the ~30k CU a first-time ATA creation costs. Raise the limit
+                    // by
+                    // that headroom so the first swap per fee mint can't revert on an exceeded
+                    // compute budget.
+                    .let { withFeeAta ->
+                        bumpComputeUnitLimit(withFeeAta, CREATE_ATA_COMPUTE_UNITS)
+                    }
             else quoteSwapData.data
 
         val feePrice = (SolanaTransaction.getComputeUnitPrice(swapTxData) ?: "0").toBigInteger()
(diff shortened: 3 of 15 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -893,6 +898,14 @@ constructor(
...
                                     tokenValue = tokenValue,
                                     srcAddress = src.address.address,
+                                    // VULT-scaled affiliate fee (parity with Kyber/1inch/LiFi):
+                                    // 50 bps base, reduced by the holder's tier discount, never
+                                    // negative. Jupiter takes 0% itself, so this is the only fee.
+                                    affiliateBps =
+                                        maxOf(
+                                            0,
+                                            JUPITER_AFFILIATE_FEE_BPS - (vultBPSDiscount ?: 0),
+                                        ),
```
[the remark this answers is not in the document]
aminsato: Verified against the code — this isn't a collision. `getCachedQuoteOrFetch(...)` keys on `srcToken.address` (the vault's own Solana address), and the VULT discount is a pure function of the vault, so two different affiliate-bps values always originate from two different vaults → different `srcAddress` → distinct cache entries. Kyber / 1inch / LI.FI all vary by `vultBPSDiscount` and use this exact key pattern for the same reason. Leaving as-is to stay consistent with the other affiliate providers.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -76,17 +100,23 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Jupite
         }
         val quoteSwapData = swapResponse.bodyOrThrow<QuoteSwapTransactionJson>()
 
-        val feePrice =
-            (SolanaTransaction.getComputeUnitPrice(quoteSwapData.data) ?: "0").toBigInteger()
+        // Jupiter never initializes the `feeAccount` ATA, so prepend an idempotent create for it.
+        // The user pays the ~0.002 SOL rent the first time per fee mint; it is a no-op thereafter.
+        // The co-signer signs this exact transaction from the keysign payload, so byte-parity
+        // holds. Note: the displayed Solana network-fee estimate does not yet include this one-time
+        // rent (the SDK adds `ataRentLamports`); that is a deferred fee-display refinement.
```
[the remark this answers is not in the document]
aminsato: This is a known, deliberate deferral (see the code comment here and the "Known limitation" section in the PR description). The user pays the correct rent on-chain — only the pre-swap network-fee *estimate* omits the one-time ~0.002 SOL ATA rent (the SDK adds `ataRentLamports`). Surfacing it cleanly means threading the "charges fee / prepends ATA" signal from the quote into the Solana network-fee estimator, which is a separate fee-display change; tracking it as a follow-up rather than expanding this PR's surface.

aminsato:
Thanks @coderabbitai — addressed in 0169c7a:

**Nitpicks (fixed):**
- ✅ **SwapQuoteRepositoryProvidersTest** — added `jupiter forwards a positive VULT-scaled affiliate bps to the api`, pinning the non-null wiring (`affiliateBps=30`) alongside the existing null-path test.
- ✅ **JupiterApiTest** — added `a resolved fee prepends the create-ATA onto the swap transaction`: `/swap` now returns a tx so the impl reaches `prependCreateFeeAta`; the fake records its inputs and raises a marker just before the native compute-budget step, asserting the prepend runs on the swap tx with the swap's fee payer and the resolved fee account.

**Major findings (replied inline, no change):**
- **Quote cache key** — not a collision: the key already includes `srcToken.address` (the vault), and the discount is a pure function of the vault, so different fee values never share a cache entry. Same pattern as Kyber/1inch/LI.FI.
- **First-time fee-ATA rent** — known, deliberate deferral (documented in the code comment and the PR's "Known limitation"); the on-chain amount is correct, only the pre-swap estimate omits it. Tracked as a fee-display follow-up.

The end-to-end behavior is verified on mainnet — the linked tx shows the fee ATA created and credited.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -76,17 +100,23 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Jupite
...
+        val swapTxData =
+            if (feeAccount != null)
+                feeAtaService.prependCreateFeeAta(quoteSwapData.data, feeAccount, fromAddress)
+            else quoteSwapData.data
+
+        val feePrice = (SolanaTransaction.getComputeUnitPrice(swapTxData) ?: "0").toBigInteger()
```
rkokhatskyi, line 130: `SetComputeUnitLimit` is sized by Jupiter's `dynamicComputeUnitLimit` simulation that ran before `prependCreateFeeAta`, and only the compute-unit price is re-bumped here, so the first create per fee mint adds ~25k CU that can exceed the baked limit and revert the tx, needs mainnet confirmation

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterFeeAtaService.kt
```diff
@@ -0,0 +1,142 @@
...
+            tokenProgramId = tokenProgramId,
+        )
+    }
+
+    override fun prependCreateFeeAta(
+        txData: String,
+        fee: JupiterFeeAccount,
+        feePayer: String,
+    ): String =
+        SolanaTransaction.insertInstruction(
```
rkokhatskyi, line 95: `prependCreateFeeAta` adds the fee owner and related accounts as new static keys, but wallet-core `insertInstruction` cannot see the route's address-lookup tables, so a dense v0 route can overflow 1232 bytes or hit `AccountLoadedTwice` and be rejected at broadcast, needs mainnet confirmation on a multi-hop ALT route

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -76,17 +100,23 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Jupite
         }
         val quoteSwapData = swapResponse.bodyOrThrow<QuoteSwapTransactionJson>()
 
-        val feePrice =
-            (SolanaTransaction.getComputeUnitPrice(quoteSwapData.data) ?: "0").toBigInteger()
+        // Jupiter never initializes the `feeAccount` ATA, so prepend an idempotent create for it.
+        // The user pays the ~0.002 SOL rent the first time per fee mint; it is a no-op thereafter.
```
rkokhatskyi: The MAX native-SOL reserve at `SwapFormViewModel.kt:455` covers only the network-fee estimate, which excludes this ~0.002 SOL fee-ATA rent, so the first swap into a not-yet-created fee ATA underfunds the wallet and the tx fails on submit, needs runtime confirmation

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -52,10 +66,20 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Jupite
...
+        val feeAccount =
+            if (requestsPlatformFee && quotedFeeAmount > BigInteger.ZERO)
+                feeAtaService.resolveFeeAccount(toToken)
+            else null
+
         val quoteSwapRequestBody = buildJsonObject {
             put("quoteResponse", json.encodeToJsonElement(body))
             put("userPublicKey", fromAddress)
             put("dynamicComputeUnitLimit", true)
+            if (feeAccount != null) put("feeAccount", feeAccount.feeAccount)
```
rkokhatskyi, line 89: When Jupiter floors `platformFee.amount` to 0, line 79 still round-trips `platformFee` into the swap body while `feeAccount` is omitted here, so if Jupiter rejects a fee-bearing quote lacking `feeAccount` the now-preferred Jupiter throws and falls back to a worse provider, needs runtime confirmation

data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
```diff
@@ -390,6 +398,27 @@ constructor(
...
+                    method = "getAccountInfo",
+                    params =
+                        buildJsonArray {
+                            add(account)
+                            addJsonObject { put("encoding", "base64") }
+                        },
+                )
+                .result
+                ?.value
+                ?.owner
```
rkokhatskyi: `getAccountOwner` reads only `result.value.owner` and never inspects the parsed `.error`, while `postRpc` does not throw on a 200 JSON-RPC error body, so a transient RPC error resolves to null with no log and silently drops the now-preferred Jupiter unlike the sibling methods here

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
rkokhatskyi: This comment's last sentence references `ataRentLamports`, which exists nowhere in this repo, and narrates deferred fee-display work, so it should be dropped as stale-prone noise while the first three why-lines stay

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -453,10 +453,19 @@ constructor(
...
+                    quotePipeline.gasFeeChain.value == srcToken.chain
+            ) {
+                val baseFee =
+                    quotePipeline.estimatedNetworkFeeTokenValue.value?.value ?: BigInteger.ZERO
+                // A Jupiter affiliate swap may create the destination-mint fee ATA on first use,
+                // whose ~0.002 SOL rent the payer (this wallet) funds but the network-fee estimate
+                // doesn't include. Reserve it on native-SOL MAX so the first such swap can't
+                // underfund and fail on submit; it leaves harmless dust when no ATA is created.
+                if (srcToken.chain == Chain.Solana) baseFee + SOLANA_FEE_ATA_RENT_RESERVE
+                else baseFee
```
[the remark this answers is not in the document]
aminsato: Fixed in 5ab98e528. Gated the reserve to same-chain Solana: it now adds `SOLANA_FEE_ATA_RENT_RESERVE` only when `srcToken.chain == Chain.Solana && dstChain == Chain.Solana`. A SOL→non-Solana MAX (routed via THORChain/LI.FI) no longer reserves dust.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -96,8 +143,23 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Jupite
...
 
+    /**
+     * Raise the transaction's `SetComputeUnitLimit` by [extraUnits]. Returns [txData] unchanged
+     * when it carries no limit instruction (nothing to rebase) or wallet-core cannot re-encode it.
+     */
+    private fun bumpComputeUnitLimit(txData: String, extraUnits: BigInteger): String {
+        val currentLimit =
+            SolanaTransaction.getComputeUnitLimit(txData)?.toBigIntegerOrNull() ?: return txData
+        return SolanaTransaction.setComputeUnitLimit(txData, (currentLimit + extraUnits).toString())
+            ?: txData
```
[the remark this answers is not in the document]
aminsato: Fixed in 5ab98e528. `bumpComputeUnitLimit` now fails closed: the no-limit case stays a no-op (early return), but if `setComputeUnitLimit` returns null after the fee-ATA create is prepended it throws instead of returning the original tx — broadcasting the un-bumped budget would guarantee an on-chain revert, so dropping Jupiter to a fallback is the safer outcome.

rkokhatskyi (approved):
LGTM

Issue #3911, opened 2026-03-26
[Fix] Chain names are partially hidden in select chains and DeFi chains list [Chains Selection]


braja555:
[Fix]

**Problem**
Chain names are partially hidden in both the search results on the select chains page and the default list on the DeFi chains page, making them difficult to read.

**Expected Behavior**
Chain names should be fully visible and properly aligned in both search results and the default chains list.

**Steps to Reproduce**
1. Go to the Select Chains page.
2. Search for any chain.
3. Observe that the chain names in search results are half hidden.
4. Navigate to DeFi → Select DeFi Chains.
5. Observe that all chain names in the list are partially hidden.

![image](https://github.com/user-attachments/assets/b892bff5-15bc-435a-8c80-2bb7ae5be219)

PR #5095, opened 2026-06-29
fix(chains): clear nav bar inset on Select chains grid bottom (#5084)

### LINKED ISSUE #5084: Select chains: last row label is clipped at bottom of sheet
## Description
On the **Select chains** screen, the label of the final row of chain tiles is partially cut off at the bottom of the bottom sheet. In the attached screenshot the last visible tile (Zcash) shows its icon but the text label "Zcash" is clipped, so the chain name is not fully readable.

## Steps to reproduce
1. Open the app and go to the chain selection screen ("Select chains").
2. Scroll to the very bottom of the chains grid.

## Expected
The last row's tile labels are fully visible, with enough bottom padding so the chain name renders completely above the sheet edge / home indicator.

## Actual
The bottom row's label is clipped — only the top of the text is visible (e.g. "Zcash").

## Likely cause
Missing bottom content padding / scroll content inset on the chains grid, so the last row sits flush against the bottom edge of the sheet.

## Environment
- Platform: Android
- Screen: Select chains (bottom sheet)

_Screenshot available from reporter (10:19 status bar capture showing clipped "Zcash" label)._

aminsato:
## Summary
Fixes #5084 — on the **Select chains** bottom sheet, the last row's chain label (alphabetically *Zcash*) is clipped/dimmed at the bottom of the sheet.

The grid (`TokenSelectionList`) reserved a hardcoded `60.dp` bottom content padding (originally added in #3911 for the same "chain names hidden" class of bug). On devices where the bottom sheet renders **under** the system navigation bar, that fixed 60dp isn't enough to lift the last row clear of a tall nav bar plus the `BottomFadeEffect`, so the final label gets clipped.

## Change
- Make the grid's bottom content padding **nav-bar-inset aware** so it scales with the actual device instead of a fixed value:
  ```kotlin
  contentPadding = PaddingValues(bottom = 60.dp + navigationBarPadding)
  ```
  This benefits every grid built on `TokenSelectionList` (chain selection, DeFi chain selection, token management).
- Adds a `chain_selection` `PreviewActivity` entry that renders the full chain list, for reproduction.

## Reproduction note
The clip does **not** reproduce on a stock Pixel AVD — Material's `ModalBottomSheet` insets its content above the nav bar there. The bug manifests on devices/OEMs where the sheet renders edge-to-edge under the nav bar. The before/after below were captured on **Small_Phone (360×640dp, 3-button nav)** reproducing that under-nav-bar condition; the last row sits tight against the nav bar before, and clears it after.

## Before / After

| Before | After |
|--------|-------|
| before | after |

## Test plan
- [x] `./gradlew :app:assembleDebug` green
- [x] ktfmt clean
- [x] Last-row label clears the nav bar on the affected configuration

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the token selection grid so items no longer sit too close to the bottom navigation area on devices with gesture or system navigation bars.
* **Chores**
  * Expanded the app’s preview coverage with an additional chain-selection preview screen for easier visual checks during development.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -70,8 +70,10 @@ import com.vultisig.wallet.ui.components.v2.fastselection.components.ChainSelect
 import com.vultisig.wallet.ui.components.v2.fastselection.components.SelectPopup
 import com.vultisig.wallet.ui.components.v2.snackbar.rememberVsSnackbarState
 import com.vultisig.wallet.ui.models.AccountUiModel
+import com.vultisig.wallet.ui.models.ChainSelectionUiModel
 import com.vultisig.wallet.ui.models.ChainTokenUiModel
 import com.vultisig.wallet.ui.models.ChainTokensUiModel
+import com.vultisig.wallet.ui.models.ChainUiModel
 import com.vultisig.wallet.ui.models.DeviceMeta
 import com.vultisig.wallet.ui.models.TransactionDetailsUiModel
 import com.vultisig.wallet.ui.models.TransactionScanStatus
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5096, opened 2026-06-29
fix(keysign): guard QR share bitmap scaling against null color space (#5089)

### LINKED ISSUE #5089: Crash: can't create bitmap without a color space in MakeQrCodeBitmapShareFormat
## Summary

`MakeQrCodeBitmapShareFormatImpl.invoke` crashes with `IllegalArgumentException: can't create bitmap without a color space` while rendering the shareable QR code bitmap. The exception originates inside `Bitmap.createScaledBitmap` — reached through the `Bitmap.scale(...)` extension used to clamp the QR bitmap into the allowed size range. When the source bitmap reports a `null` color space, `createScaledBitmap` rejects it and throws.

This happens on the keysign "save QR" path (`KeysignShareViewModel.saveShareQrBitmap`), so affected users cannot save/share the QR image.

## Stack trace

```
Exception java.lang.IllegalArgumentException: can't create bitmap without a color space
    at android.graphics.Bitmap.createBitmap (Bitmap.java:1255)
    at android.graphics.Bitmap.createBitmap (Bitmap.java:1031)
    at android.graphics.Bitmap.createScaledBitmap (Bitmap.java:883)
    at com.vultisig.wallet.data.usecases.MakeQrCodeBitmapShareFormatImpl.invoke (MakeQrCodeBitmapShareFormat.kt:305)
    at com.vultisig.wallet.data.usecases.MakeQrCodeBitmapShareFormatImpl.invoke (MakeQrCodeBitmapShareFormat.kt:86)
    at com.vultisig.wallet.ui.models.keysign.KeysignShareViewModel$saveShareQrBitmap$1$rendered$1.invokeSuspend (KeysignShareViewModel.kt:246)
    at kotlinx.coroutines...
```

## Affected versions / devices

- App version: **108** (per Android Vitals)
- Device: Samsung Galaxy A54 5G
- OS: Android 16 (API 36)
- First seen: 2026-06-28

## Where it happens

The crash is on one of the `Bitmap.scale(...)` calls in `MakeQrCodeBitmapShareFormat.kt` that clamp the QR bitmap to the `[QR_CODE_MIN_SIZE_DP, QR_CODE_MAX_SIZE_DP]` range (and/or the icon/logo scale calls). `scale` delegates to `Bitmap.createScaledBitmap`, which on API 26+ throws `can't create bitmap without a color space` when the source bitmap's color space is `null`.

## Suggested direction

Ensure every bitmap fed into a scale/createScaledBitmap call carries a valid color space (e.g. guard with a copy into `ARGB_8888` / sRGB when `colorSpace == null` before scaling), and wrap the share-bitmap rendering so a failure degrades gracefully instead of crashing the keysign share flow.

aminsato:
## Summary

Fixes #5089 — `IllegalArgumentException: can't create bitmap without a color space` crash on the keysign **save/share QR** path (`KeysignShareViewModel.saveShareQrBitmap` → `MakeQrCodeBitmapShareFormatImpl`).

`Bitmap.scale()` delegates to `Bitmap.createScaledBitmap`, which rejects a source bitmap whose `colorSpace` is `null`. Some devices (reported on Samsung Galaxy A54 5G / Android 16) hand us QR/logo bitmaps with a null color space, so scaling them to clamp into the `[QR_CODE_MIN_SIZE_DP, QR_CODE_MAX_SIZE_DP]` range crashes the share flow.

## Changes

- **`MakeQrCodeBitmapShareFormat.kt`** — new private `Bitmap.scaleWithColorSpace()` helper: when the source `colorSpace == null`, it redraws the bitmap into a fresh `createBitmap(...)` (which defaults to sRGB) before scaling, then recycles the temporary copy. Applied to all scale call sites (QR clamp ×2, value icon, signature logo). Existing `!== original` recycle guards are preserved.
- **`KeysignShareViewModel.saveShareQrBitmap`** — the render is now wrapped in a try/catch (rethrows `CancellationException`, otherwise logs via Timber and returns `null`) so any future render failure degrades gracefully (share bitmap stays unset, user can retry) instead of crashing keysign.

## Test plan

- [x] `:data:compileDebugKotlin` and `:app:compileDebugKotlin` pass
- [x] ktfmt clean
- [ ] Manual: save/share QR from keysign on an affected device renders without crashing

No unit test added: the project has no Robolectric and `Bitmap` is a framework class that can't be exercised on the JVM. The change is a defensive guard around real `Bitmap` behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QR/share bitmap generation to be cancellation- and error-safe, preventing crashes and keeping the share bitmap from being left in an inconsistent state.
  * Fixed QR code, field icon, and logo bitmap scaling on devices where bitmap color space information is missing, reducing rendering failures.


data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
```diff
@@ -59,6 +59,24 @@ interface MakeQrCodeBitmapShareFormat : (Context, Bitmap, Int, Bitmap, QrShareIn
 
 private fun Float.dp(context: Context): Float = this * context.resources.displayMetrics.density
 
+// `Bitmap.scale` delegates to `createScaledBitmap`, which rejects a source whose color space is
+// null with "can't create bitmap without a color space" (some devices hand us such QR/logo
+// bitmaps). Redraw those into a fresh sRGB-backed bitmap before scaling so the share render can
+// never crash on this input.
+private fun Bitmap.scaleWithColorSpace(width: Int, height: Int, filter: Boolean = true): Bitmap {
```
rkokhatskyi: `GenerateQrBitmap.kt:72` scales the embedded logo with raw `logo.scale()`, so the same null color space crash survives on the Share Vault QR screen where `ShareVaultQrScreen.kt:73` passes `decodeResource(ic_qr_vultisig)` and `generateBitmap` catches only `WriterException`

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/usecases/MakeQrCodeBitmapShareFormat.kt
aminsato: Good catch — fixed in 106cf9c06. `GenerateQrBitmap` was scaling the embedded logo with raw `logo.scale()`, and `ShareVaultQrViewModel.generateBitmap()` only catches `WriterException`, so the same null-color-space crash survived on this path (and Keygen/Keysign/QBTC/account QR all share this use case).

I extracted the color-space-safe scale into a shared `BitmapColorSpaceScaling.kt` and the logo now goes through `scaleWithColorSpace(...)`. The two `bitmap.scale(...)` calls there are left raw since that bitmap is created `ARGB_8888` and always has a color space.

rkokhatskyi (approved):
LGTM

PR #5107, opened 2026-06-30
perf(balances): decouple price refresh from balance fetch (#5077)

### LINKED ISSUE #5077: Decouple price fetch from balance fetch (stop prices gating balance display)
## Summary
Each per-address coroutine awaits the **full price refresh before fetching balances**, and `.first()` collapses the reactive price flow so live price updates are discarded. Measured price-batch RTT ≈ 540 ms (plus slow trailing THOR/Maya pool calls), so a slow price feed delays balances even for chains that need no pricing. Price is read from the in-memory `tokenIdToPrice` StateFlow (resolves instantly), so this gate is a **sequencing artifact**, not a real data dependency. (Desktop already decouples these.)

## Current behavior
`data/.../repositories/AccountsRepository.kt:94-161`:
```kotlin
val loadPrices = async { tokenPriceRepository.refresh(vaultCoins) }
addresses.fetchAccountFromDb(); send(... isComplete=false)   // cached snapshot first (good)
addresses.mapIndexed { index, account -> async {
    loadPrices.await()                                       // BLOCKS fresh balances (line 111)
    val newAccounts = account.accounts.map { async {
        val balance = balanceRepository
            .getTokenBalanceAndPrice(address, it.token).first()  // .first() discards live price (line 123)
        it.applyBalance(balance.tokenBalance, balance.price)
    } }.awaitAll()
    addresses[index] = account.copy(accounts = newAccounts); send(... isComplete=false)
} }.awaitAll()
send(... isComplete=true)
```

## Proposed change
1. **Remove `loadPrices.await()` (`:111`)** so balances fetch immediately in parallel with the price refresh (price StateFlow returns cached/zero instantly).
2. Make fiat self-update when fresh prices land — either:
   - **(a, minimal)** keep per-token `.first()`, but after `awaitAll()` (`:153`) do `loadPrices.await()` then re-emit a snapshot recomputing fiat from the now-fresh StateFlow; or
   - **(b, cleaner)** change `getTokenBalanceAndPrice(...).first()` (`:123`, mirror at `:277`, `:430`) to **collect** the flow so a price-StateFlow update re-emits balance×price automatically. The `.first()` is the structural root of the coupling.
3. Apply the same to the single-coin paths at `:277` and `:430`.

## Correctness / risks
- $0-fiat flicker only on first-ever cold start. Rates persist (`tokenPrice` Room + StateFlow), so warm starts never flicker.
- Verify the `channelFlow` still emits `isComplete=true` once balances resolve if option (b) keeps the price flow long-lived (use a take/until, or use option (a)).
- No change to call count/load; independent of Multicall and TTL work.

## Acceptance criteria
- [ ] Crypto amounts render without waiting for the price refresh (balance display ≤ balance-RTT, not + price-RTT).
- [ ] A slow/stubbed price endpoint no longer delays balance amounts for price-independent chains (e.g. BTC).
- [ ] Fiat populates/updates once prices land (no permanent $0 on warm start).
- [ ] Cached-snapshot-first cold-start behavior + `isComplete` terminal emission unchanged.

_Part of a cross-app balance-fetching initiative (also filed for vultisig-ios)._

aminsato:
## Summary

Fixes #5077.

Each per-address coroutine in `AccountsRepository` awaited the **full price refresh before fetching balances**, so a slow price feed (measured batch RTT ≈ 540 ms, plus trailing THOR/Maya pool calls) delayed balances even for chains that need no pricing — and a failed refresh blocked *every* balance fetch. Prices are served from an in-memory `MutableStateFlow` (`tokenIdToPrice`) that resolves instantly, and `getTokenBalanceAndPrice` zips it with a single-shot balance flow, so `.first()` just snapshots whatever price is present at collection time. The `loadPrices.await()` gate was a **sequencing artifact**, not a real data dependency (desktop already decouples these).

## Changes

- **`loadAddressBalances`** — removed the per-chain `loadPrices.await()` gate so balances fetch immediately in parallel with the refresh and stream as each chain resolves. After every chain resolves, fiat is recomputed from the freshly persisted prices via a new `recomputeFiatFromFreshPrices()` helper that reuses the Room-cache read (`getCachedTokenBalanceAndPrice`) — **no extra network calls**. The terminal `isComplete=true` emission is preserved, and a price-refresh failure is caught so it can't block completion.
- **`loadAddress`** — replaced the `supervisorScope`-joined refresh (which serialized refresh → balance) with a `coroutineScope`/`async` so the balance fetch runs concurrently with the refresh, then re-emits from cache once prices land.
- **`loadAccount`** — the refresh now runs as `async` concurrently with the balance fetch (wall-clock `max` instead of sum), then recomputes fiat from cache.

## Correctness / risks

- `$0`-fiat flicker only on a first-ever cold start (empty in-memory price StateFlow). Rates persist in Room + StateFlow, so warm starts never flicker — the recompute is a no-op confirmation there.
- No change to call count / load; independent of the Multicall and TTL work.
- A failed price refresh no longer blocks balance display.

## Acceptance criteria

- [x] Crypto amounts render without waiting for the price refresh.
- [x] A slow/stubbed price endpoint no longer delays balance amounts for price-independent chains.
- [x] Fiat populates/updates once prices land (no permanent `$0` on warm start).
- [x] Cached-snapshot-first cold-start behavior + `isComplete` terminal emission unchanged.

## Test plan

- `AccountsRepositoryImplTest` updated for the cache-recompute path.
- New test **`balances stream before the price refresh completes and fiat updates once prices land`**: gates the price refresh, asserts the crypto amount surfaces with `$0` fiat while prices are still pending, then asserts fiat populates on the terminal emission once prices resolve (no second balance fetch).
- `./gradlew :data:testDebugUnitTest --tests "*AccountsRepositoryImplTest"` green; ktfmt applied.

_Part of a cross-app balance-fetching initiative (also filed for vultisig-ios)._

## Summary by CodeRabbit

* **Bug Fixes**
  * Balance streaming now prioritizes immediate crypto amount updates while token price refresh happens independently, without delaying partial results.
  * When loading completes, fiat values are recalculated from the latest persisted cached balance/price, improving accuracy across address and account views.
  * Token price lookup now falls back to persisted (cached) prices when in-memory data is empty.
  * Token price refresh errors are logged and handled gracefully so completion still completes.

* **Tests**
  * Updated/added repository tests to verify streaming vs. refresh timing, correct fiat recomputation from cache, cold-start price fallback, and resilience to refresh/cache failures.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
@@ -324,7 +324,7 @@ constructor(
                             tokenPriceRepository.refresh(updatedCoins)
                         } catch (e: Exception) {
                             if (e is kotlinx.coroutines.CancellationException) throw e
-                            Timber.e(e, "Price refresh failed for ${chain.id}")
+                            Timber.e(e, "Price refresh failed for %s", chain.id)
                         }
                     }
 
@@ -558,8 +558,12 @@ constructor(
         // cache, so once prices land we recompute fiat from the cache without a second network
         // call.
         val loadPrices = async {
-            runCatching { tokenPriceRepository.refresh(coins) }
-                .onFailure { Timber.e(it, "Failed to refresh token prices for chain: $chain") }
+            try {
+                tokenPriceRepository.refresh(coins)
+            } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
+                Timber.e(e, "Failed to refresh token prices for chain: %s", chain)
+            }
         }
 
         val accountToUpdate =
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
@@ -109,8 +109,12 @@ constructor(
                             async {
                                 try {
                                     val address = account.address
-                                    loadPrices.await()
 
+                                    // Don't gate balances on the price refresh: prices come from an
+                                    // in-memory StateFlow that resolves instantly (cached/zero), so
+                                    // awaiting the refresh here only delayed balances for chains
+                                    // that need no pricing. Fiat is recomputed below once prices
+                                    // land. (Desktop decouples these the same way.)
                                     val newAccounts =
                                         if (account.chain.standard == TokenStandard.EVM) {
                                             // One Multicall3 round-trip per (chain, address) for
@@ -174,6 +178,21 @@ constructor(
                         }
                         .awaitAll()
 
+                    // Balances are already streamed; now wait for the price refresh and recompute
+                    // fiat from the freshly persisted prices. On a warm start the StateFlow already
+                    // held prices, so the per-chain emissions were already correct and this is a
+                    // no-op confirmation; only a cold start (empty StateFlow) updates fiat here.
+                    try {
+                        loadPrices.await()
+                    } catch (e: Exception) {
+                        if (e is kotlinx.coroutines.CancellationException) throw e
+                        Timber.e(
+                            e,
+                            "Price refresh failed; emitting balances with last-known prices",
+                        )
+                    }
+                    addresses.recomputeFiatFromFreshPrices()
+
                     // Terminal emission: every chain has resolved (or failed). Carries the final
                     // snapshot and flags completion so the UI can stop the refresh spinner.
                     send(AddressBalancesUpdate(addresses.toList(), isComplete = true))
@@ -229,6 +248,29 @@ constructor(
         }
     }
 
+    /**
+     * Recomputes each account's fiat value and unit price from the now-refreshed prices, reusing
+     * the Room-backed [BalanceRepository.getCachedTokenBalanceAndPrice] read (the freshly fetched
+     * balance was already persisted there, and so is the fresh price) so no extra network calls are
+     * made. Accounts with no cached balance (e.g. a brand-new coin whose network fetch failed) are
+     * left untouched so a streamed value is never overwritten with null.
+     */
+    private suspend fun MutableList<Address>.recomputeFiatFromFreshPrices() {
+        forEachIndexed { index, account ->
+            val newAccounts =
+                account.accounts.map { acc ->
+                    val balance =
+                        balanceRepository.getCachedTokenBalanceAndPrice(account.address, acc.token)
+                    if (balance.tokenBalance.tokenValue != null) {
+                        acc.applyBalance(balance.tokenBalance, balance.price)
+                    } else {
+                        acc
+                    }
+                }
+            this[index] = account.copy(accounts = newAccounts)
+        }
+    }
+
     private suspend fun getSPLCoins(solanaCoins: List<Coin>, vault: Vault): List<Coin> {
         if (solanaCoins.any { !it.isNativeToken }) return emptyList()
         val solanaAddress = solanaCoins.firstOrNull()?.address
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -109,8 +109,12 @@ constructor(
                             async {
                                 try {
                                     val address = account.address
-                                    loadPrices.await()
 
+                                    // Don't gate balances on the price refresh: prices come from an
+                                    // in-memory StateFlow that resolves instantly (cached/zero), so
+                                    // awaiting the refresh here only delayed balances for chains
+                                    // that need no pricing. Fiat is recomputed below once prices
+                                    // land. (Desktop decouples these the same way.)
```
rkokhatskyi, line 117: This decoupling prices each streamed balance via `getPrice`, which reads the in-memory `tokenIdToPrice` that starts empty on every process start (`TokenPriceRepository.kt:60`), so the per-chain emission overwrites the cached snapshot fiat with $0 until the terminal recompute. iOS fetches prices first and Windows serves last-known, so neither shows the cold-open $0 dip this introduces

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -267,13 +313,30 @@ constructor(
...
+                        try {
+                            tokenPriceRepository.refresh(updatedCoins)
+                        } catch (e: Exception) {
+                            if (e is kotlinx.coroutines.CancellationException) throw e
+                            Timber.e(e, "Price refresh failed for %s", chain.id)
+                        }
+                    }
 
-                loadPrices.await()
+                    emitRefreshAddress(account)
```
rkokhatskyi, line 331: `emitRefreshAddress` here also prices from the empty `tokenIdToPrice` StateFlow, so the coin-detail fiat drops from the cached value to $0 before `emitCachedAddress` restores it, the dip the removed `loadPrices.await()` ordering used to prevent on cold start

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Good catch — confirmed the cold-start $0 dip. Root cause: both this streamed emission and `emitRefreshAddress` price via `getTokenBalanceAndPrice`/`getEvmTokenBalancesAndPrices` → `TokenPriceRepository.getPrice()`, which only reads the in-memory `tokenIdToPrice` StateFlow (empty until a refresh lands).

Fixed centrally in `getPrice` (commit 1c54412de): it now falls back to the persisted Room price (`getCachedPrice`) when the in-memory map has no entry yet, so the last-known fiat holds until the refresh updates the StateFlow — matching the last-known behavior you referenced. This covers both flagged sites (and the DeFi path) without re-coupling the fetch. Added cold-start + in-memory-precedence tests.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Good catch — confirmed the cold-start $0 dip. Root cause: both this streamed emission and `emitRefreshAddress` price via `getTokenBalanceAndPrice`/`getEvmTokenBalancesAndPrices` → `TokenPriceRepository.getPrice()`, which only reads the in-memory `tokenIdToPrice` StateFlow (empty until a refresh lands).

Fixed centrally in `getPrice` (commit 1c54412de): it now falls back to the persisted Room price (`getCachedPrice`) when the in-memory map has no entry yet, so the last-known fiat holds until the refresh updates the StateFlow — matching the last-known behavior you referenced. This covers both flagged sites (and the DeFi path) without re-coupling the fetch. Added cold-start + in-memory-precedence tests.

rkokhatskyi (approved):
LGTM

Issue #5085, opened 2026-06-29
DeFi chains list does not refresh immediately after enabling/disabling chains


aminsato:
# [Fix] DeFi chains list does not refresh after enabling/disabling chains [DeFi Portfolio / Select DeFi chains]

## Problem
After enabling or disabling chains in the **Select DeFi chains** screen, the **DeFi** screen (DeFi Portfolio / Portfolio list) does not update immediately to reflect the change. The user must perform a manual pull-to-refresh before the DeFi chain list shows the updated set of chains. The screen continues showing the previous chain list until then.

## Expected Behavior
The DeFi portfolio chain list updates immediately to reflect the chains that were just enabled/disabled, with no manual pull-to-refresh required.

## Steps to Reproduce
1. Open the **DeFi** tab from the bottom navigation.
2. Open the **Select DeFi chains** screen (chain selector).
3. Enable or disable one or more chains, then confirm (tap the checkmark / close the sheet).
4. Return to the DeFi screen.
5. Observe that the chain list still shows the previous set of chains until a manual pull-to-refresh is performed.

### Must Do
- [ ] Ensure the DeFi Portfolio / Portfolio list observes chain enable/disable changes and re-renders immediately when returning from the **Select DeFi chains** screen (e.g. refresh/invalidate the relevant state or flow on chain-selection confirmation).
- [ ] Verify the Wallet tab for the same stale-list behavior and apply a consistent fix if it is also affected.

### Must NOT Do
- Don't change unrelated code
- Don't alter the chain selection/persistence logic itself — only the refresh/observation of the DeFi (and, if affected, Wallet) chain list
- Don't remove or break the existing pull-to-refresh behavior

## Acceptance Criteria
- [ ] `./gradlew assembleDebug` succeeds
- [ ] After enabling/disabling chains in **Select DeFi chains** and returning to the DeFi screen, the chain list reflects the change immediately without a manual pull-to-refresh
- [ ] Wallet tab behavior is confirmed consistent (either unaffected, or fixed the same way)

## Context
- Reproduced on the DeFi tab ("DeFi Portfolio" / "Portfolio" list).
- The Wallet tab behavior should be checked for the same issue for consistency.
- Repository: vultisig/vultisig-android
- Platform: android
- Issue type: bugfix

PR #5109, opened 2026-06-30
fix(defi): prune disabled chains from persisted DeFi selection

### LINKED ISSUE #5093: Disabling a chain in the Wallet selector does not remove it from the DeFi chain selection
## Description

When a chain is disabled in the **Wallet → Select chains** screen, it is removed from the vault and disappears from the visible DeFi surfaces. However, the persisted **DeFi chain selection** (the `defaultDeFiChains` set backing the DeFi Portfolio filter and the *Select DeFi chains* screen) is **never pruned** for that chain.

As a result the DeFi selection state is left inconsistent: the chain stays a member of the saved DeFi set even though it is no longer in the vault.

## Steps to Reproduce

1. Have a vault with a DeFi-supported chain enabled (e.g. THORChain), and confirm it appears in the DeFi Portfolio.
2. Go to **Wallet → Select chains** and **disable** that chain. Confirm.
3. The chain correctly disappears from the DeFi Portfolio (it derives from the vault's coins).
4. Later, go back to **Wallet → Select chains** and **re-enable** the same chain.

## Expected Behavior

Re-enabling a chain in the Wallet selector should re-add it in a neutral state. The user should explicitly opt it back into DeFi via *Select DeFi chains* if they want it there.

## Actual Behavior

On re-enable, the chain **immediately reappears in the DeFi Portfolio as enabled**, because it was never removed from the saved DeFi chain set when it was disabled. The previous DeFi selection silently resurrects.

## Notes

- The DeFi Portfolio and *Select DeFi chains* both derive their *available* chains from the vault's coins, so a disabled chain is correctly hidden while absent. The inconsistency is in the persisted DeFi selection set, which outlives the chain's removal from the vault.
- Suggested direction: when chains are disabled in the Wallet chain selector commit path, also remove those chains from the persisted DeFi chain set so the two stay in sync.
- Related: #5085 / #5090 (immediate refresh of the DeFi list after editing DeFi chains).

aminsato:
## Summary

Fixes #5093.

Disabling a chain in **Wallet → Select chains** removed it from the vault but left it a member of the persisted DeFi chain set (`defi_chains_<vaultId>`). Because the DeFi Portfolio / *Select DeFi chains* screens derive their *available* chains from the vault's coins, the stale membership stayed dormant while the chain was gone — and silently resurrected (re-appearing as **enabled** in the DeFi Portfolio) when the chain was later re-enabled.

This keeps the two stores in sync: when chains are disabled in the Wallet selector commit path, they are also pruned from the persisted DeFi set. Re-enabling a chain now comes back in a neutral (unselected) state; the user must explicitly opt it back into DeFi via *Select DeFi chains*.

## Changes

- `DefaultDeFiChainsRepository`: new `removeChains(vaultId, chains)` that prunes chains from the effective set and persists **only when something actually changes** — so disabling an unrelated / non-DeFi chain never needlessly materializes defaults or flips the `hasBeenSet` flag.
- `ChainSelectionViewModel.onCommitChanges()`: after disabling accounts, prune the disabled chains from the DeFi set.

## Edge cases handled

- Never-customized vault (default DeFi set `{ThorChain, Ethereum}`): disabling ThorChain persists `{Ethereum}`, so it won't reappear on re-enable.
- Disabling a chain not in the DeFi set (or a non-DeFi chain): no-op, defaults stay un-materialized.

## Test plan

- Added `ChainSelectionViewModelTest` cases: prunes disabled chains on commit; passes an empty set when nothing is disabled.
- `ChainSelectionViewModelTest` + `DeFiChainSelectionViewModelTest` green.
- ktfmt applied.

## Related

- #5085 / #5090 (immediate refresh of the DeFi list after editing DeFi chains)

## Summary by CodeRabbit

* **Bug Fixes**
  * When vault chain selections are reduced, matching DeFi chain selections are now removed too.
  * Empty updates are handled safely without changing saved selections.
  * Added coverage for both removal and no-change scenarios.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/ChainSelectionViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/ChainSelectionViewModel.kt
@@ -14,6 +14,7 @@ import com.vultisig.wallet.data.models.TssKeysignType
 import com.vultisig.wallet.data.models.Vault
 import com.vultisig.wallet.data.models.hasPreGeneratedKey
 import com.vultisig.wallet.data.repositories.ChainAccountAddressRepository
+import com.vultisig.wallet.data.repositories.DefaultDeFiChainsRepository
 import com.vultisig.wallet.data.repositories.RequestResultRepository
 import com.vultisig.wallet.data.repositories.TokenRepository
 import com.vultisig.wallet.data.repositories.VaultRepository
@@ -53,6 +54,7 @@ constructor(
     private val tokenRepository: TokenRepository,
     private val chainAccountAddressRepository: ChainAccountAddressRepository,
     private val discoverTokenUseCase: DiscoverTokenUseCase,
+    private val defaultDeFiChainsRepository: DefaultDeFiChainsRepository,
     private val navigator: Navigator<Destination>,
     private val requestResultRepository: RequestResultRepository,
     private val snackbarFlow: SnackbarFlow,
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5110, opened 2026-06-30
fix(solana): skipPreflight on broadcast to survive blockhash propagation lag

### LINKED ISSUE #5102: [Fix] Solana swap signs a stale route — only the blockhash is refreshed before keysign, not the quote [Swap/Solana]
## Observed
Solana swaps (e.g. Jupiter PYTH→Fartcoin, cf. #2054) intermittently fail at broadcast with `Transaction simulation failed` (custom program error / slippage). More likely with Fast Vaults, where the server round-trip lengthens the quote→broadcast window.

## Cause
`data/.../chains/helpers/SolanaSwap.kt` `getPreSignedInputData` (L41-89) decodes the provider's pre-built transaction and overwrites **only** `recentBlockhash` (L66 / L79: `setRecentBlockhash(recentBlockHash)`), then re-signs that message. The route, slippage bounds, and account/pool state baked into the quote are **not** refreshed. If price/pool state drifts past slippage between quote time and broadcast, preflight simulation rejects it.

`UpdateSolanaKeysignPayloadUseCase` only patches `BlockChainSpecific.Solana.recentBlockHash`; it does not re-fetch the swap quote.

## Fix shape
- Re-fetch the swap quote close to keysign (or shorten/track the quote validity window) instead of only swapping the blockhash.
- Surface slippage/route preflight failures with a distinct, actionable message (ties into the on-chain-error-messaging issue) rather than the generic device-timeout copy.

## Cross-platform
iOS: **Healthier — Android should match.** Re-fetches the *whole* quote (route + slippage + chain-specific) on a ~60s timer and on retry (`Features/Swap/ViewModels/SwapVerifyViewModel.swift:74-134` `refreshData` → `fetchQuote` + `fetchChainSpecific`), and builds the keysign payload from that fresh tx at sign time. The blockhash-only overwrite (`Blockchain/Solana/Signing/SolanaSwaps.swift:38-71`) is only a final touch-up on an already-fresh quote, so route/slippage aren't left stale the way Android leaves them.
Windows: **Affected (same gap), and explicitly acknowledged.** For general (LiFi) Solana swaps the prebuilt quote tx is reused and only `recentBlockhash` is overwritten (`@vultisig/core-mpc` `signingInputs/resolvers/solana/index.ts`); route/slippage/accounts stay as quoted. The SDK even comments it (`chain/swap/general/lifi/api/injectSolanaAtaIfMissing.ts`): *"the upstream fix is to call LiFi closer to broadcast time (or refresh the quote on expiry)… deferred as out-of-scope."* (Native THORChain Solana swaps rebuild fresh on both Android & Windows.) Worth a mirror issue on vultisig-windows. Note provider differs: Android general-swap = Jupiter, Windows general-swap = LiFi.

Related: #2054.

---

aminsato:
**Proof (on-chain, Jupiter swap landed after the fix):** https://orbmarkets.io/tx/5886KTmHC5SwCwMT7TAShkn884UBZW8b6MsbGYdm2YkFDCFs6zdssQTNWEnVY5nYMEaxu2LtUrMVvbsfRMbWdDov

Closes #5102

## Problem

After #5105 (transient `BlockhashNotFound` retry), a Solana broadcast — most visibly a **Jupiter swap** — could still fail with `Transaction simulation failed` / `BlockhashNotFound`. An on-device trace shows the retry working as designed but still exhausting:

```
13:19:32.893  sendTransaction → {"err":"BlockhashNotFound"}
13:19:33.388  Transient blockhash-not-found on broadcast attempt 1/3; resending after backoff
13:19:35.x    attempt 2/3 → BlockhashNotFound
13:19:37.9    attempt 3/3 → BlockhashNotFound
13:19:38.396  Error broadcasting transaction: Transaction simulation failed
13:19:38–44   getSignatureStatuses (real hash) → value:[null]  (recovery works, tx never landed)
13:19:44.288  SolanaBlockhashExpiredException
```

So #5105 did its job (correct classification + clean recovery, no crash), but the broadcast still didn't land.

## Root cause

`sendTransaction` was sent **without `skipPreflight`**, so the receiving (load-balanced) RPC node **simulates the tx against its own blockhash cache** and rejects an otherwise-valid recent blockhash with `BlockhashNotFound`. Every retry hits a different lagging node and fails the same way.

This is worse for **Jupiter swaps**: the blockhash is minted on a different cluster (`api.vultisig.com/jup`) than the broadcast proxy (`api.vultisig.com/solana`), so it propagates to the broadcast nodes *slower* than the ~5s retry window. Native sends recover because their blockhash comes from the same proxy they broadcast to.

## Fix

Set `skipPreflight: true` on `sendTransaction`. The node then skips its local simulation and **forwards the tx straight to the current leader** (at the tip, has the blockhash). The transaction is already simulated upstream — WalletCore for native sends, Jupiter for swaps (its quote carries `simulationError: null`) — and on-chain success is still verified by the existing status polling, so the receiving node's preflight re-simulation was both redundant and the thing failing.

## Notes / scope

- **iOS parity:** iOS does *not* set `skipPreflight` today (it only pins `encoding=base64`). This is an Android-leading change worth porting to iOS.
- `skipPreflight` means a genuinely-invalid tx isn't caught client-side and fails on-chain instead — acceptable here (Solana charges no fee for a failed tx, and swaps keep on-chain min-out/slippage protection).

## Test plan

- [x] `SolanaApiBodyReadTest` passes, incl. a new case asserting the `sendTransaction` request body carries `"skipPreflight":true`
- [ ] Manual: a Jupiter SOL→token swap from a Fast Vault now broadcasts and confirms through node lag

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana transaction broadcasting by sending the RPC request with `preflightCommitment: "confirmed"` alongside the transaction payload.
  * Updated Solana status parsing to treat landed transactions with an `"err"` field as failed, and refined how status fields are extracted and interpreted.

* **Tests**
  * Added request-body capture support to validate outgoing broadcast parameters.
  * Expanded coverage for signature status mapping and `"err"`-driven failure behavior across finalized/processed and error/no-error scenarios.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
@@ -200,16 +200,10 @@ constructor(
                 RpcPayload(
                     jsonrpc = "2.0",
                     method = "sendTransaction",
-                    // skipPreflight: don't let the receiving RPC node simulate the tx against its
-                    // own (load-balanced, possibly lagging) blockhash cache — that re-simulation is
-                    // what rejects an otherwise-valid recent blockhash with "BlockhashNotFound",
-                    // especially for Jupiter swaps whose blockhash is minted on a different cluster.
-                    // The node forwards straight to the current leader (at the tip, has the
-                    // blockhash) instead. On-chain success is still verified by status polling.
                     params =
                         buildJsonArray {
                             add(tx)
-                            addJsonObject { put("skipPreflight", true) }
+                            addJsonObject { put("preflightCommitment", "confirmed") }
                         },
                     id = 1,
                 )
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
@@ -200,7 +200,17 @@ constructor(
                 RpcPayload(
                     jsonrpc = "2.0",
                     method = "sendTransaction",
-                    params = buildJsonArray { add(tx) },
+                    // skipPreflight: don't let the receiving RPC node simulate the tx against its
+                    // own (load-balanced, possibly lagging) blockhash cache — that re-simulation is
+                    // what rejects an otherwise-valid recent blockhash with "BlockhashNotFound",
+                    // especially for Jupiter swaps whose blockhash is minted on a different cluster.
+                    // The node forwards straight to the current leader (at the tip, has the
+                    // blockhash) instead. On-chain success is still verified by status polling.
+                    params =
+                        buildJsonArray {
+                            add(tx)
+                            addJsonObject { put("skipPreflight", true) }
+                        },
                     id = 1,
                 )
             repeat(BROADCAST_MAX_ATTEMPTS) { index ->
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/models/SolanaJson.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/models/SolanaJson.kt
@@ -4,6 +4,7 @@ import java.math.BigInteger
 import kotlinx.serialization.Contextual
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonElement
 import kotlinx.serialization.json.JsonObject
 
 @Serializable
@@ -70,7 +71,13 @@ data class SolanaRpcResponseJson<T>(
     @SerialName("error") val error: RpcError?,
 )
 
-@Serializable data class SolanaSignatureStatus(val confirmationStatus: String? = null)
+@Serializable
+data class SolanaSignatureStatus(
+    val confirmationStatus: String? = null,
+    // Non-null when the tx executed but reverted (e.g. swap slippage, insufficient funds). A Solana
+    // TransactionError is either a string or an object, so keep it as a raw JsonElement.
+    @SerialName("err") val err: JsonElement? = null,
+)
 
 @Serializable
 data class SolanaAccountInfoResponseJson(
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/SolanaStatusProvider.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/SolanaStatusProvider.kt
@@ -13,14 +13,20 @@ internal class SolanaStatusProvider @Inject constructor(private val solanaApi: S
 
     override suspend fun checkStatus(txHash: String, chain: Chain): TransactionResult =
         try {
-            val confirmationStatus =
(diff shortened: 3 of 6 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
```diff
@@ -192,7 +192,17 @@ constructor(
...
+                    // skipPreflight: don't let the receiving RPC node simulate the tx against its
+                    // own (load-balanced, possibly lagging) blockhash cache — that re-simulation is
+                    // what rejects an otherwise-valid recent blockhash with "BlockhashNotFound",
+                    // especially for Jupiter swaps whose blockhash is minted on a different cluster.
+                    // The node forwards straight to the current leader (at the tip, has the
+                    // blockhash) instead. On-chain success is still verified by status polling.
+                    params =
+                        buildJsonArray {
+                            add(tx)
+                            addJsonObject { put("skipPreflight", true) }
```
rkokhatskyi: `skipPreflight=true` lets a tx that fails on execution (Jupiter slippage revert, insufficient funds) land as a failed tx that `SolanaStatusProvider` reports Confirmed, because `SolanaSignatureStatus` (SolanaJson.kt:73) drops `err`. iOS returns failed when `err` is set, so Android shows a green success for a reverted swap that still paid the fee

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
aminsato: Good catch — fixed in 428c7a399. You're right that skipPreflight lets a reverted tx (slippage, insufficient funds) land on-chain, where the old mapping would have shown it green.

`SolanaSignatureStatus` now carries `err`, and `SolanaStatusProvider` reports `TransactionResult.Failed(err)` whenever it's set — taking priority over `confirmationStatus` — so a finalized-but-failed tx is no longer treated as confirmed. Mirrors iOS. Added `SolanaStatusProviderTest` covering finalized-with-err and processed-with-err → Failed, plus the unchanged confirmed/pending paths.

rkokhatskyi (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
johnnyluo: @aminsato instead of skipPreflight, can we send with parameter "preflightCommitment" set it to "confirmed"

johnnyluo:
https://orbmarkets.io/tx/5kVvZBQxmSyqLzHVxgerVk1zYZm1moaRH4sR8Pq6m3hEPvjtjmmGL28bzrnAaaZf9wWbBjVSW9z6vqua1ot5Q1bV

PR #5114, opened 2026-07-01
fix(solana): priority fee conflates total lamports with per-CU price

### LINKED ISSUE #5101: [Fix] Solana priority fee conflates total fee (lamports) with per-compute-unit price [Send/Solana]
## Observed
The Solana priority fee placed in the signed transaction does not match the per-compute-unit price it's supposed to be. Under network congestion the broadcast tx is **under-prioritized** (so it lingers unconfirmed and is more likely to hit blockhash expiry), and the displayed network fee does not equal the fee actually paid.

## Cause
The send path stores the **total fee amount** (lamports) into the field used as the **per-compute-unit price** (microlamports/CU):

- `app/.../send/GasFeeOrchestrator.kt` L186: `TokenValue(value = fees.amount, token = nativeCoin)` — `gasFee.value` = total fee (base + priorityAmount + rent), in lamports.
- `data/.../repositories/BlockChainSpecificRepository.kt` L404: `BlockChainSpecific.Solana(priorityFee = gasFee.value, … priorityLimit = SOLANA_PRIORITY_FEE_LIMIT /*100_000*/)` — total-fee-lamports is assigned to `priorityFee`.
- `data/.../chains/helpers/SolanaHelper.kt` L63-83: `setPriorityFeePrice(price = maxOf(solanaSpecific.priorityFee, 1_000_000))` and `setPriorityFeeLimit(limit = priorityLimit)` — i.e. that lamports value is treated as price (microlamports/CU) and multiplied by 100,000 CU.

`applyGasSettings` / `adjustGasFee` only correct ETH/UTXO/Cardano, never Solana, so the conflation is never fixed downstream. The `maxOf(…, 1_000_000)` floor masks it at low fees (result happens to land at the floor), but it diverges as soon as the dynamic median fee rises.

Net effect (worked through): the actual broadcast priority fee caps around ~0.001 SOL even at the congestion cap, i.e. it **under-pays** relative to the displayed estimate — so it does not cause insufficient-funds, but it does cause slow inclusion (feeding blockhash-expiry failures, see sibling issue) and a display/charged mismatch.

## Fix shape
Pass the priority-fee **price** (the per-CU value from `getMedianPriorityFee`, i.e. `GasFees.price`) into `BlockChainSpecific.Solana.priorityFee`, not the total `fees.amount`. Keep `priorityLimit` as the CU limit. Verify display fee == sum of base + (price × limit / 1e6).

## Cross-platform
iOS: **Correct (not buggy) — Android should match.** Keeps the two fields separate: `priorityFeePrice.price` = per-CU microlamports (median of `getRecentPrioritizationFees`) and `priorityFeeLimit.limit` = 100_000 CU (`Blockchain/Solana/Signing/Solana.swift:54-59,135-140`; `BlockChainSpecific.swift:25` annotated "priority fee is in microlamports"). No conflation with a total lamport amount.
Windows: **Correct (not buggy) — Android should match.** `@vultisig/core-chain` keeps `priorityFeePrice` (µLam/CU, from `getRecentPrioritizationFees` percentile) and `priorityFeeLimit` (100_000 CU) as separate fields (`solanaConfig.ts`, `signingInputs/resolvers/solana/send.ts`). No conflation with a total-lamports amount. **This bug is Android-only — both iOS and Windows are correct.**

---

aminsato:
**Proof (on-chain, Fast Vault SOL send after the fix):** https://orb.helius.dev/tx/3GuQC3eYetR7AK6LF6fxV2WFpjWgjcuiQF7ytxzt3mRFpXHszxyg8saW5Yhw8QmddfT3KUm2qG1dXcZuoMrCyJ7y

Closes #5101

## Problem

The Solana send path stored the **total fee** (lamports) into the field used as the **per-compute-unit price** (microlamports/CU). `BlockChainSpecificRepository` assigned `gasFee.value` (base + priority + rent) to `BlockChainSpecific.Solana.priorityFee`, which `SolanaHelper` feeds into `setPriorityFeePrice(...)` and multiplies by the 100k CU limit. The `maxOf(price, 1_000_000)` floor masked it at low fees, but it diverges as the dynamic median rises.

**Net effect:** under congestion the broadcast tx is **under-prioritized** — it lingers unconfirmed and is more likely to hit blockhash expiry — and the displayed network fee doesn't equal the fee actually paid.

## Fix

In the `TokenStandard.SOL` branch of `getSpecific`, read the dynamic median price via `solanaApi.getMedianPriorityFee(...)` (`getRecentPrioritizationFees`) — the same source the fee **display** uses (`SolanaFeeService`) — and pass it as `priorityFee`, keeping `priorityLimit` as the CU limit. So the signed tx is prioritized correctly and `display fee == base + (price × limit / 1e6)`.

## Cross-platform parity

Verified against the local iOS repo: `BlockChainService.swift` builds `.Solana(priorityFee: dynamicPriorityFee, priorityLimit: SolanaHelper.priorityFeeLimit, …)` where `dynamicPriorityFee = max(fetchRecentPrioritizationFees(), defaultFee)` — identical to this change. Windows keeps the two fields separate too. **The bug was Android-only.**

## Test plan

- [x] `BlockChainSpecificRepositoryImplTest` passes, incl. a new case asserting the SOL `priorityFee` is the per-CU median (`50000`) and **not** the total `gasFee` (`105000`)
- [ ] Manual: send SOL under congestion — priority fee in the signed tx matches the median price × limit, and the charged fee equals the displayed estimate

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
@@ -398,10 +398,22 @@ constructor(
                     val fromAddressPubKeyResult = fromAddressPubKey.await()
                     val toAddressPubKeyResult = toAddressPubKey.await()
                     Timber.d("solana blockhash: $recentBlockHashResult")
+                    // priorityFee is a per-compute-unit PRICE (microlamports/CU), not the total fee
+                    // amount. gasFee.value holds the total (base + priority + rent); assigning it
+                    // here made SolanaHelper multiply the whole fee by the 100k CU limit, mispricing
+                    // the tx. Read the dynamic median price the same way the fee display does
+                    // (getRecentPrioritizationFees) so the signed tx is prioritized correctly and the
+                    // charged fee matches the estimate. iOS/Windows keep these two fields separate.
+                    val priorityAccounts = buildList {
+                        add(token.address)
+                        fromAddressPubKeyResult?.first?.let { add(it) }
+                        toAddressPubKeyResult?.first?.let { add(it) }
+                    }
+                    val priorityFeePrice = solanaApi.getMedianPriorityFee(priorityAccounts)
                     BlockChainSpecificAndUtxo(
                         BlockChainSpecific.Solana(
                             recentBlockHash = recentBlockHashResult,
-                            priorityFee = gasFee.value,
+                            priorityFee = priorityFeePrice,
                             fromAddressPubKey = fromAddressPubKeyResult?.first,
                             toAddressPubKey = toAddressPubKeyResult?.first,
                             programId = fromAddressPubKeyResult?.second == true,
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
@@ -29,6 +29,7 @@ import com.vultisig.wallet.data.blockchain.model.VaultData
 import com.vultisig.wallet.data.blockchain.sui.SuiFeeService.Companion.SUI_DEFAULT_GAS_BUDGET
 import com.vultisig.wallet.data.chains.helpers.CardanoHelper
 import com.vultisig.wallet.data.chains.helpers.SOLANA_PRIORITY_FEE_LIMIT
+import com.vultisig.wallet.data.chains.helpers.SOLANA_PRIORITY_FEE_PRICE
 import com.vultisig.wallet.data.chains.helpers.TronHelper.Companion.TRON_DEFAULT_ESTIMATION_FEE
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
@@ -398,18 +399,21 @@ constructor(
                     val fromAddressPubKeyResult = fromAddressPubKey.await()
                     val toAddressPubKeyResult = toAddressPubKey.await()
                     Timber.d("solana blockhash: $recentBlockHashResult")
-                    // priorityFee is a per-compute-unit PRICE (microlamports/CU), not the total fee
-                    // amount. gasFee.value holds the total (base + priority + rent); assigning it
-                    // here made SolanaHelper multiply the whole fee by the 100k CU limit, mispricing
-                    // the tx. Read the dynamic median price the same way the fee display does
-                    // (getRecentPrioritizationFees) so the signed tx is prioritized correctly and the
-                    // charged fee matches the estimate. iOS/Windows keep these two fields separate.
-                    val priorityAccounts = buildList {
-                        add(token.address)
-                        fromAddressPubKeyResult?.first?.let { add(it) }
-                        toAddressPubKeyResult?.first?.let { add(it) }
-                    }
-                    val priorityFeePrice = solanaApi.getMedianPriorityFee(priorityAccounts)
+                    // priorityFee is a per-compute-unit price (microlamports/CU); priorityLimit is
+                    // the CU limit. SolanaHelper charges price * limit / 1e6 as the priority fee.
+                    // Swaps sign the aggregator's prebuilt tx, which carries its own compute budget,
+                    // so these fields are ignored for swaps — skip the RPC and use the fallback.
+                    val priorityFeePrice =
+                        if (isSwap) {
+                            SOLANA_PRIORITY_FEE_PRICE.toBigInteger()
+                        } else {
+                            val priorityAccounts = buildList {
+                                add(token.address)
+                                fromAddressPubKeyResult?.first?.let { add(it) }
+                                toAddressPubKeyResult?.first?.let { add(it) }
+                            }
+                            solanaApi.getMedianPriorityFee(priorityAccounts)
+                        }
                     BlockChainSpecificAndUtxo(
                         BlockChainSpecific.Solana(
                             recentBlockHash = recentBlockHashResult,
--- a/data/src/test/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepositoryImplTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepositoryImplTest.kt
@@ -27,12 +27,14 @@ import com.vultisig.wallet.data.blockchain.model.GasFees
 import com.vultisig.wallet.data.blockchain.model.Transfer
 import com.vultisig.wallet.data.blockchain.sui.SuiFeeService.Companion.SUI_DEFAULT_GAS_BUDGET
 import com.vultisig.wallet.data.chains.helpers.SOLANA_PRIORITY_FEE_LIMIT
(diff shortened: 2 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
```diff
@@ -398,10 +398,22 @@ constructor(
...
+                    // here made SolanaHelper multiply the whole fee by the 100k CU limit, mispricing
+                    // the tx. Read the dynamic median price the same way the fee display does
+                    // (getRecentPrioritizationFees) so the signed tx is prioritized correctly and the
+                    // charged fee matches the estimate. iOS/Windows keep these two fields separate.
+                    val priorityAccounts = buildList {
+                        add(token.address)
+                        fromAddressPubKeyResult?.first?.let { add(it) }
+                        toAddressPubKeyResult?.first?.let { add(it) }
+                    }
+                    val priorityFeePrice = solanaApi.getMedianPriorityFee(priorityAccounts)
```
rkokhatskyi: `priorityFeePrice` is a fresh median fetch, but `GasFeeSelection.kt:34` still feeds the stale `gasFee` into the balance check and Verify Send's `estimatedFee`/`totalGas`, so under congestion the confirm screen and the signed tx can charge different priority fees.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
```diff
@@ -398,10 +398,22 @@ constructor(
                     val fromAddressPubKeyResult = fromAddressPubKey.await()
                     val toAddressPubKeyResult = toAddressPubKey.await()
                     Timber.d("solana blockhash: $recentBlockHashResult")
+                    // priorityFee is a per-compute-unit PRICE (microlamports/CU), not the total fee
+                    // amount. gasFee.value holds the total (base + priority + rent); assigning it
+                    // here made SolanaHelper multiply the whole fee by the 100k CU limit, mispricing
+                    // the tx. Read the dynamic median price the same way the fee display does
+                    // (getRecentPrioritizationFees) so the signed tx is prioritized correctly and the
+                    // charged fee matches the estimate. iOS/Windows keep these two fields separate.
+                    val priorityAccounts = buildList {
```
rkokhatskyi: `getMedianPriorityFee` now runs for swaps too, but Solana swap signers ignore `priorityFee`/`priorityLimit` since aggregators bake their own compute budget into the prebuilt tx per `SolanaFeeService.kt:58`, so every swap build fires a wasted RPC call.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
```diff
@@ -398,10 +398,22 @@ constructor(
                     val fromAddressPubKeyResult = fromAddressPubKey.await()
                     val toAddressPubKeyResult = toAddressPubKey.await()
                     Timber.d("solana blockhash: $recentBlockHashResult")
+                    // priorityFee is a per-compute-unit PRICE (microlamports/CU), not the total fee
```
rkokhatskyi: This comment narrates the old bug, `gasFee.value` multiplied by the CU limit in `SolanaHelper`, instead of stating the invariant, so it reads as stale history once forgotten. Keep the per-CU price fact and drop the narration.

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
aminsato: Done in 6d7a4a8a1 — reworded to state the invariant (priorityFee is a per-CU price; charged = price × limit / 1e6) instead of narrating the old bug.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
aminsato: Fixed in 6d7a4a8a1 — the median fetch is now gated on `!isSwap`. Swap builds use the fallback floor price with no RPC, since swap signers ignore `priorityFee`/`priorityLimit` (aggregator tx carries its own compute budget). Added a test asserting a swap build makes zero `getMedianPriorityFee` calls.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
aminsato: Good catch — this is a real divergence for native sends (the swap path no longer fetches after the gate above). The display fee flows `GasFeeOrchestrator` → `calculateFees` (median @ T1) → `TokenValue(fees.amount)`, and `GasFeeOrchestrator:186` discards `GasFees.price`; the signed tx now fetches its own median @ T2 in `getSpecific`, so under congestion they can differ.

The clean fix is to compute the per-CU price once and share it: expose `GasFees.price` from `GasFeeOrchestrator`, carry it alongside `gasFee` to `DefaultSendStrategy`, and pass it into `getSpecific` (which falls back to the fetch when absent). That touches the send gas pipeline (`GasFeeOrchestrator` / `GasFeeParams` / `DefaultSendStrategy` / `getSpecific`), so I'd prefer to do it as a focused follow-up rather than widen this PR — it needs an on-device check of the Verify screen fee under congestion.

Want it in this PR, or a follow-up issue? Note the divergence is bounded: both values are medians clamped to [1M, 100M] µLam/CU fetched seconds apart, and under non-congestion both hit the 1M floor and are identical.

johnnyluo (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
johnnyluo: Confirmed the signing path is unaffected, so this is display-only. The join device signs the initiator's payload verbatim: `resolveMessagesToSign` → `SigningHelper.getKeysignMessages(payload, vault)` → `SolanaHelper.getPreSignedImageHash` → `getPreSignedInputData` reads `priorityFee`/`priorityLimit` straight from `keysignPayload.blockChainSpecific` (`getSpecific` is never called on the join path). Every co-signer therefore derives the same sighash from the initiator's per-CU price, and the broadcast tx carries exactly that value.

The divergence is limited to the verify-screen fee **label** — `JoinKeysignFeeResolver` re-fetches via `calculateFees` for Solana instead of reading the payload's `priorityFee`. No signing/broadcast impact. Keeping this PR scoped to the per-CU semantic fix and tracking the display-label alignment (join side) + initiator-side price threading as a follow-up.

aminsato: Thanks for the detailed trace — agreed on all points. The signing/broadcast path is correct on both initiator and join devices (co-signers derive the same sighash from the initiator's per-CU price), so this is display-only. Keeping this PR scoped to the per-CU semantic fix and tracked the two display-label gaps (initiator-side price threading + join-side `JoinKeysignFeeResolver` reading the payload instead of re-fetching) as a follow-up: #5127.

PR #5116, opened 2026-07-01
fix(keysign): on-chain broadcast rejections shown as network errors, not device timeout

### LINKED ISSUE #5099: [Fix] Solana on-chain broadcast failures mislabeled as "one of your devices didn't respond" — on-chain errors must say on-chain [Keysign/Errors]
## Observed
A Solana **send** (and swap) with a Fast Vault fails on the keysign error screen with:

- Title: **Transaction failed**
- Body: **"One of your devices didn't respond in time. Check your connection and try again."**
- "Show exact error" reveals: **`Transaction simulation failed`** (+ a Solana reason)

The body copy tells the user (and support) this is a pairing/connectivity/MPC-timeout problem. It is not. The MPC ceremony **succeeded** — both devices responded, the signature was produced and locally verified — and the failure happened later, at **broadcast time**, when the Solana RPC `sendTransaction` preflight rejected the transaction on-chain.

There is **no client-side simulation** anywhere in the Android send/swap path, so `"Transaction simulation failed"` can only come from the RPC preflight at broadcast.

## Cause
`app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt` → `resolveSigningError(rawMessage)` (≈ L28-96) only recognizes 5 substrings (`Blockhash not found`, `insufficient funds`, the SPL missing-ATA prefix, the mixed-reshare marker, the Cosmos broadcast marker). Everything else falls into the `else` catch-all, which shows `signing_error_transaction_failed_description` = *"One of your devices didn't respond in time…"*.

Common Solana RPC preflight reasons that fall into the generic `else` and get mislabeled:
- `Transaction simulation failed: insufficient lamports …` (system-program — note "lamports" not "funds", so it misses the `insufficient funds` branch)
- `Transaction simulation failed: Attempt to debit an account but found no record of a prior credit`
- `Transaction simulation failed: This transaction has already been processed`
- `… block height exceeded` (blockhash expiry, but worded differently from `Blockhash not found`)

The mapping is also internally inconsistent for the *same* root cause: blockhash expiry surfaces as either `Blockhash not found` (→ timeout copy) or `block height exceeded` (→ generic device copy).

## Also: the on-chain reason is swallowed
`SolanaApi.broadcastTransaction` only surfaces `error["message"]` (`SolanaApi.kt:206`); the richer `error.data.err` / `error.data.logs` (where Solana puts the per-instruction reason + program logs) is parsed into the `error: JsonObject?` model (`SolanaJson.kt:42`) but **discarded**. This is long-standing (true both before and after the #4875 rewrite), not a regression — but combined with the v1.0.110 friendly screen demoting the raw text behind "Show exact error", users now see a bare "Transaction simulation failed" with no reason. Fix: when building the broadcast exception, append `error.data.err` + `error.data.logs` so "Show exact error" actually shows the on-chain cause.

## Fix shape
1. Add an explicit branch so any error containing `Transaction simulation failed` / `block height exceeded` / `insufficient lamports` / `already been processed` is presented as an **on-chain** failure with accurate copy (e.g. "The network rejected this transaction" + the on-chain reason), not a device/connection message.
2. Never route an error containing `simulation failed` into the device-timeout `else`.
3. Keep the raw text under "Show exact error".
4. **Retry UX:** with MPC, silent auto-retry is not possible (the user must sign again). The error screen's **"Try again"** must clearly route back to re-sign (today `KeysignFlowViewModel.tryAgain()` → `back()` → Verify screen, which re-runs `setData` and fetches a fresh blockhash). Confirm this re-sign path works cleanly from this error state and that the button label/expectation matches "sign again", not "retry network".

## Cross-platform
iOS: **Healthier — Android should match.** No friendly-mapping layer; `KeysignViewModel.handleBroadcastError` (`Features/Keysign/ViewModels/KeysignViewModel.swift:1075-1115`) falls through to a `default` that surfaces the raw RPC reason (`SolanaServiceError.rpcError` → "RPC Error: <reason>"), shown via `ErrorView` with a "Show exact error" disclosure. iOS also reclassifies a tx that actually landed (`isAlreadyOnChain` by deterministic hash, `:1107-1111`) as **success** instead of an error — Android does not, and shows "already been processed" as a hard failure. Recommend Android (1) surface the on-chain reason and (2) add the already-on-chain success reclassification.
Windows: **Also affected (shares the mislabel).** `core/ui/mpc/keysign/KeysignSigningStep.tsx` renders `signing_error_description` = *"One of your devices didn't respond in time…"* (`core/ui/i18n/locales/en.ts`) for **every** non-session error, including broadcast-time RPC rejections (signing + `broadcastTx` share one mutation, `useKeysignMutation.ts:332-340`). Milder than Android — no `resolveSigningError`-style remapper, and the raw error is shown via a "Show exact error" disclosure — but the headline is still the device-timeout copy, and "Try again" (`FlowErrorPageContent.tsx`) only navigates back (no rebroadcast). Worth a mirror issue on vultisig-windows.

Related: #4996 (introduced the friendly screen in v1.0.110), #4767, #4863.

---

aminsato:
Closes #5099

## Problem

A Solana send/swap with a Fast Vault that fails **at broadcast** was shown on the keysign error screen as:

- Title: **Transaction failed**
- Body: **"One of your devices didn't respond in time. Check your connection and try again."**
- "Show exact error": **`Transaction simulation failed`** (bare, no reason)

That copy blames pairing/connectivity/MPC. But the MPC ceremony succeeded — the signature was produced and locally verified — and the failure happened later, when the Solana RPC `sendTransaction` preflight rejected the tx on-chain. The app runs **no** client-side simulation, so `Transaction simulation failed` can only come from the node's preflight.

Real example (Jupiter swap, `err = AccountLoadedTwice`): the on-chain reason was discarded and the screen said "device didn't respond".

## Fix

**1. Surface the on-chain reason (`SolanaApi.broadcastTransaction`).** Solana puts the real cause in `error.data.err` (`AccountLoadedTwice`, `BlockhashNotFound`, …) while `error.message` is only the generic `Transaction simulation failed`. The reason was parsed into the model but dropped. Now it's appended to the surfaced message.

**2. Classify broadcast rejections as on-chain (`resolveSigningError`).** Any `simulation failed` now routes to an **on-chain rejection** branch — `Transaction rejected by the network: <reason>` (reusing the existing Cosmos-broadcast strings, no new copy) — instead of the device-timeout `else`. Matching is normalized (lowercase, strip spaces/underscores) so it catches both the RPC wording and Solana's camelCase enums. Also:
- Blockhash expiry is unified: both `Blockhash not found` and `block height exceeded` map to the "sign again" copy (previously inconsistent).
- `insufficient funds` **and** `insufficient lamports` map to the insufficient-funds copy (the lamports wording previously missed the branch).

Raw text stays under "Show exact error". **"Try again"** already routes back to the Verify screen, which re-signs with a fresh blockhash (`KeysignFlowViewModel.tryAgain() → back()`), so the re-sign UX is correct for MPC.

## Cross-platform

Matches iOS, which surfaces the raw RPC reason rather than a friendly device-timeout. (iOS also reclassifies an already-on-chain tx as success — Android's `BroadcastTxUseCase.recoverIfAlreadyBroadcast` already covers the landed-tx case via status polling; a fuller reclassification can be a follow-up.) Windows shares the mislabel — worth a mirror issue.

## Before / After

Same failure (Fast Vault Jupiter swap, on-chain `AccountLoadedTwice` at broadcast), real device renders.

| Before | After |
|--------|-------|
| before | after |

Before: *"One of your devices didn't respond in time…"* (implies pairing/connectivity). After: *"Transaction rejected by the network: AccountLoadedTwice"* (the actual on-chain cause).

## Test plan

- [x] `:data:compileDebugKotlin` + `:app:compileDebugKotlin` pass
- [x] `SolanaApiBodyReadTest` passes, incl. a new case asserting the broadcast exception message contains the on-chain reason (`AccountLoadedTwice`)
- [x] Manual (on-device, Fast Vault Jupiter swap): broadcast rejected with `AccountLoadedTwice`. Logcat: `Error broadcasting transaction: Transaction simulation failed: AccountLoadedTwice` (reason now appended); the error screen reads **"Transaction rejected by the network: AccountLoadedTwice"** — not "device didn't respond".

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction error handling by matching failure messages more reliably across casing and formatting.
  * Enhanced “signing error” descriptions, including clearer guidance for common failures and better support for simulation-related rejection details.
  * Broadcast and send transaction failures now surface more specific on-chain reasons when available, with safe fallbacks to generic messaging.
* **Tests**
  * Expanded coverage for extracting on-chain error reasons across string, structured, and non-object payload formats.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
@@ -25,15 +25,20 @@ internal data class SigningError(
  * use the red critical variant.
  */
 @Composable
-internal fun resolveSigningError(rawMessage: String): SigningError =
-    when {
-        rawMessage.contains("Blockhash not found") ->
+internal fun resolveSigningError(rawMessage: String): SigningError {
+    // Normalize away spaces/underscores/case so a single check matches both the RPC's human wording
+    // ("Blockhash not found", "insufficient lamports") and Solana's camelCase enum forms
+    // ("BlockhashNotFound", "BlockHeightExceeded") that error.data.err carries.
+    val normalized = rawMessage.lowercase().replace(" ", "").replace("_", "")
+    return when {
+        // Blockhash expiry — both RPC wordings map to the same "sign again" copy.
+        normalized.contains("blockhashnotfound") || normalized.contains("blockheightexceeded") ->
             SigningError(
                 title = stringResource(R.string.signing_error_transaction_failed_title),
                 description = stringResource(R.string.signing_error_transaction_timeout),
                 errorState = ErrorState.CRITICAL,
             )
-        rawMessage.contains("insufficient funds") ->
+        normalized.contains("insufficientfunds") || normalized.contains("insufficientlamports") ->
             SigningError(
                 title = stringResource(R.string.error_insufficient_funds_title),
                 description = stringResource(R.string.signing_error_insufficient_funds),
@@ -87,6 +92,15 @@ internal fun resolveSigningError(rawMessage: String): SigningError =
                 )
             }
         }
+        // On-chain rejection at broadcast: the app runs no client-side simulation, so a
+        // "simulation failed" can only come from the node's preflight. This is a network rejection,
+        // NOT a device/connection timeout — surface it as such, with the on-chain reason.
+        normalized.contains("simulationfailed") ->
+            SigningError(
+                title = stringResource(R.string.signing_error_transaction_failed_title),
+                description = onChainRejectionDescription(rawMessage),
+                errorState = ErrorState.CRITICAL,
+            )
         else ->
             SigningError(
                 title = stringResource(R.string.signing_error_transaction_failed_title),
@@ -94,6 +108,30 @@ internal fun resolveSigningError(rawMessage: String): SigningError =
                 errorState = ErrorState.CRITICAL,
             )
     }
+}
+
+/**
+ * Builds the description for an on-chain broadcast rejection, appending the node-supplied reason
+ * (the text after "simulation failed", e.g. `AccountLoadedTwice`) when present. Reuses the same
+ * "rejected by the network" copy as the Cosmos broadcast path.
+ */
+@Composable
+private fun onChainRejectionDescription(rawMessage: String): String {
+    val detail =
+        rawMessage
+            .substringAfter("simulation failed", "")
+            .trimStart(':', ' ')
+            .let { raw ->
+                if (raw.length > BROADCAST_DETAIL_MAX_LENGTH) {
+                    raw.take(BROADCAST_DETAIL_MAX_LENGTH).trimEnd() + "…"
+                } else raw
+            }
+    return if (detail.isBlank()) {
+        stringResource(R.string.signing_error_broadcast_rejected)
+    } else {
+        stringResource(R.string.signing_error_broadcast_rejected_s, detail)
+    }
+}
 
 @Composable
 internal fun KeysignErrorScreen(
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
@@ -117,9 +117,12 @@ internal fun resolveSigningError(rawMessage: String): SigningError {
  */
 @Composable
 private fun onChainRejectionDescription(rawMessage: String): String {
+    // Match the marker the same way the branch does (case + spaces/underscores insensitive) so the
+    // on-chain reason after it isn't lost for wordings like "SimulationFailed: AccountLoadedTwice".
     val detail =
-        rawMessage
-            .substringAfter("simulation failed", "")
+        SIMULATION_FAILED_MARKER.find(rawMessage)
+            ?.let { rawMessage.substring(it.range.last + 1) }
+            .orEmpty()
             .trimStart(':', ' ')
             .let { raw ->
                 if (raw.length > BROADCAST_DETAIL_MAX_LENGTH) {
@@ -159,6 +162,9 @@ internal fun KeysignErrorScreen(
  * truncated with an ellipsis before being interpolated into the user-facing label.
  */
 private const val BROADCAST_DETAIL_MAX_LENGTH = 120
+
+/** Matches the RPC "simulation failed" marker regardless of case and space/underscore wording. */
+private val SIMULATION_FAILED_MARKER = Regex("(?i)simulation[ _]?failed")
 
 @Preview(showBackground = true, name = "KeysignErrorScreen Preview")
 @Composable
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
@@ -38,7 +38,19 @@ internal fun resolveSigningError(rawMessage: String): SigningError {
                 description = stringResource(R.string.signing_error_transaction_timeout),
                 errorState = ErrorState.CRITICAL,
             )
-        normalized.contains("insufficientfunds") || normalized.contains("insufficientlamports") ->
+        // On-chain rejection at broadcast: the app runs no client-side simulation, so a
+        // "simulation failed" can only come from the node's preflight. Checked BEFORE the
+        // insufficient-funds branch so an on-chain reason that merely contains "insufficient funds"
+        // (e.g. "InsufficientFundsForRent") is reported as a network rejection with its reason,
+        // rather than the fund-the-wallet warning.
+        normalized.contains("simulationfailed") ->
+            SigningError(
+                title = stringResource(R.string.signing_error_transaction_failed_title),
+                description = onChainRejectionDescription(rawMessage),
+                errorState = ErrorState.CRITICAL,
+            )
+        // Pre-broadcast insufficient funds (e.g. EVM gas) — actionable "fund the wallet" copy.
+        normalized.contains("insufficientfunds") ->
             SigningError(
                 title = stringResource(R.string.error_insufficient_funds_title),
                 description = stringResource(R.string.signing_error_insufficient_funds),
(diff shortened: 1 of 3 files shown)


aminsato:
Note: the `AccountLoadedTwice` shown in the before/after screenshots is a *separate* underlying bug — this PR only makes it readable. Filed as #5117: Android re-serializes Jupiter's prebuilt v0 tx through WalletCore (`decode → setRawMessage → compileWithSignatures`), which mangles the address-table-lookup layout into a duplicate account. iOS avoids this by splicing the signature into the original bytes and broadcasting verbatim. The fix belongs there; this PR stands on its own for the error-copy/reason-surfacing.

data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
```diff
@@ -48,6 +48,7 @@ import kotlinx.serialization.json.addJsonArray
 import kotlinx.serialization.json.addJsonObject
 import kotlinx.serialization.json.buildJsonArray
 import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.jsonObject
```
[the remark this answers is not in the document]
aminsato: Fixed in 72b67b3bb — switched to `(error["data"] as? JsonObject)` so a primitive/string `data` returns null instead of throwing, and the original RPC message still surfaces. Added a test with `"data": "oops"`.

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
```diff
@@ -87,13 +92,46 @@ internal fun resolveSigningError(rawMessage: String): SigningError =
...
+ * Builds the description for an on-chain broadcast rejection, appending the node-supplied reason
+ * (the text after "simulation failed", e.g. `AccountLoadedTwice`) when present. Reuses the same
+ * "rejected by the network" copy as the Cosmos broadcast path.
+ */
+@Composable
+private fun onChainRejectionDescription(rawMessage: String): String {
+    val detail =
+        rawMessage
+            .substringAfter("simulation failed", "")
+            .trimStart(':', ' ')
```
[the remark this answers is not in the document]
aminsato: Fixed in 72b67b3bb — the extractor now uses `Regex("(?i)simulation[ _]?failed")`, matching the branch's normalization, so the reason is preserved for wordings like `SimulationFailed: AccountLoadedTwice`. (In practice the Solana RPC always sends the lowercase spaced form, but the robustness gap is real.)

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
```diff
@@ -25,15 +25,20 @@ internal data class SigningError(
...
+    return when {
+        // Blockhash expiry — both RPC wordings map to the same "sign again" copy.
+        normalized.contains("blockhashnotfound") || normalized.contains("blockheightexceeded") ->
             SigningError(
                 title = stringResource(R.string.signing_error_transaction_failed_title),
                 description = stringResource(R.string.signing_error_transaction_timeout),
                 errorState = ErrorState.CRITICAL,
             )
-        rawMessage.contains("insufficient funds") ->
+        normalized.contains("insufficientfunds") || normalized.contains("insufficientlamports") ->
```
rkokhatskyi: This branch matches before the new simulationfailed check below, so a real on-chain rejection like InsufficientFundsForRent, whose message contains both insufficientfunds and simulationfailed, still lands here and shows the fund-the-wallet copy instead of the intended network-rejection message

data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
```diff
@@ -216,6 +217,16 @@ constructor(
                     result.error ?: return result.result ?: error("broadcastTransaction error")
 
                 val message = error["message"]?.jsonPrimitive?.contentOrNull ?: error.toString()
+                // Solana puts the actual on-chain reason in error.data.err (e.g. "AccountLoadedTwice",
+                // "BlockhashNotFound"); error.message only carries the generic "Transaction simulation
+                // failed". Append the reason so the failure surfaced to the user names the real cause
+                // instead of a bare, generic message.
+                val reason =
+                    (error["data"] as? JsonObject)?.get("err")?.let { err ->
+                        (err as? JsonPrimitive)?.contentOrNull ?: err.toString()
```
rkokhatskyi: For structured errors like InstructionError, err is a JsonObject so the primitive cast fails and this falls back to err.toString(), embedding raw JSON such as {"InstructionError":[1,{"Custom":6001}]} directly into the always-visible rejection description shown to the user

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/screens/keysign/KeysignErrorScreen.kt
aminsato: Fixed in 3c2b16a4a — the on-chain `simulation failed` branch is now checked before the insufficient-funds branch, so `InsufficientFundsForRent` (which contains "insufficient funds") is reported as "Transaction rejected by the network: InsufficientFundsForRent" rather than the fund-the-wallet warning. Also dropped the `insufficient lamports` alias from the funds branch — it's an on-chain reason handled by the simulation-failed branch (per #5099). The pre-broadcast EVM `insufficient funds` case (no "simulation failed") still gets the fund-the-wallet copy.

data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
aminsato: Fixed in 3c2b16a4a — for a structured `err` (JsonObject) the surfaced reason is now the variant name (the single top-level key, e.g. `InstructionError`) instead of `err.toString()`, so raw JSON like `{"InstructionError":[1,{"Custom":6001}]}` no longer leaks into the user-facing rejection text. Bare string enums (`AccountLoadedTwice`) are unchanged. Added a test asserting the message contains `InstructionError` and not `Custom`.

johnnyluo (approved):
LGTM

PR #5122, opened 2026-07-01
feat(price): compact subscript notation for tiny token prices (#5120)

### LINKED ISSUE #5120: [Add] Compact subscript notation for tiny token prices (e.g. 0.0₇3)
# [Add] Compact subscript notation for tiny token prices [price display]

## Problem
Very small token prices (e.g. LUNC) already avoid rendering as `$0.00` thanks to `FiatValueToStringMapper(asPrice = true)`, but they are shown as long plain decimals (e.g. `$0.00006`) and are capped at 8 fraction digits — so extremely tiny prices like `0.00000003` lose precision or become hard to read. The desktop app and browser extension now render these using compact **subscript notation**, where a small subscript digit encodes the number of leading zeros (e.g. `0.00000003` → `$0.0₇3`, meaning "7 zeros then 3"). Android should match this for a consistent, readable presentation.

## Solution
Extend the `asPrice` branch of `FiatValueToStringMapper` to produce compact subscript notation for tiny prices instead of long plain decimals. Match the contract already shipped in the TypeScript codebase (see references): keep a few significant digits, and once there are **4 or more leading zeros** after the decimal point, collapse them into a subscript count. Prices ≥ one sub-unit (e.g. `$0.01` for USD) and zero keep standard formatting.

Expected output (parity with desktop):

| Value | Now (Android) | Requested |
|---|---|---|
| `0.00000003` | `$0.00000003` (capped) | `$0.0₇3` |
| `0.00001234` | `$0.000012` | `$0.0₄1234` |
| `0.0001234` | `$0.0001234` | `$0.0001234` (unchanged, <4 zeros) |
| `0.00006` | `$0.00006` | `$0.00006` (unchanged, <4 zeros) |
| `0.5`, `1.23` | unchanged | unchanged |

Rounding contract: **round** (HALF_UP), consistent with the existing `asPrice` path.

### Reference implementation (TypeScript)
- Formatting algorithm + subscript contract (threshold = 4 leading zeros, round): **vultisig/vultisig-sdk#918** (`formatAmount` → `$0.0₇3`)
- Legible subscript rendering on desktop/extension (renders the count as a real, slightly larger subscript rather than a tiny Unicode glyph): **vultisig/vultisig-windows#4256**

### Existing code to extend
- `FiatValueToStringMapper` `asPrice = true` branch (used in `TokenDetailViewModel`, `ChainTokensViewModel`, `CustomTokenViewModel`) — add the subscript formatting there.

### Must Do
- [ ] Render tiny prices (≥4 leading zeros) with subscript notation `$0.0ₙX` where n = leading-zero count
- [ ] Keep <4-leading-zero and ≥ sub-unit prices formatted exactly as today
- [ ] Add unit tests in `FiatValueToStringMapperImplTest.kt` covering the subscript cases

### Must NOT Do
- Don't change existing behavior for non-tiny amounts or the `asFee` path

### Out of Scope
- Applying subscript notation to fee/amount formatting (this is per-token price only)

## Acceptance Criteria
- [ ] `0.00000003` renders as `$0.0₇3` in the token price UI
- [ ] The subscript digit is legible (not a tiny Unicode glyph) — see vultisig/vultisig-windows#4256 for the desktop treatment
- [ ] Prices ≥ sub-unit and zero are unchanged

## Screenshot / Reference
Image

## Context
Follow-up to cross-platform parity with vultisig-windows issue #4231 (now resolved on desktop + extension).

aminsato:
## Summary
Closes #5120.

Tiny per-token prices (e.g. LUNC) already avoided rendering as `$0.00`, but were shown as long plain decimals capped at 8 fraction digits, so extremely small prices lost precision and were hard to read. Desktop and the browser extension already render these using **compact subscript notation**, where a subscript digit encodes the number of leading zeros (`0.00000003` → `$0.0₇3`). This brings Android to parity.

The `asPrice` branch of `FiatValueToStringMapper` now ports the shared SDK `formatTinyCurrencyAmount` contract (vultisig/vultisig-sdk#918):

- Keep **4 significant figures**.
- Once there are **≥4 leading zeros** after the decimal point, collapse them into subscript notation.
- Round **HALF_UP**.
- Computed on `BigDecimal` (round to 4 sig figs via `MathContext`, then derive the leading-zero count from `scale()/precision()`), so a rounding carry (`0.0000999999` → `$0.0001`) shifts the count correctly.
- Android keeps its existing sub-unit (`0.01`) gate; the currency symbol and its locale placement are preserved via `formatToCharacterIterator`.

### Behavior

| Value | Before | After |
|---|---|---|
| `0.00000003` | `$0.00000003` (capped) | `$0.0₇3` |
| `0.00001234` | `$0.000012` | `$0.0₄1234` |
| `0.00006` | `$0.00006` | `$0.0₄6` |
| `0.0001234` | `$0.00012` | `$0.0001234` |
| `0.00456` | `$0.00456` | `$0.00456` |
| `≥ 0.01`, `0` | unchanged | unchanged |

> **Note on the issue's spec table:** it marked both `0.00001234` and `0.00006` differently, but both have exactly 4 leading zeros — impossible under a single threshold. Following the issue's directive to "match the TypeScript codebase", the SDK contract is authoritative, so `0.00006` → `$0.0₄6`.

## Test plan
- Updated + added unit tests in `FiatValueToStringMapperImplTest` covering subscript cases, the plain-decimal branch below the threshold, HALF_UP rounding, the rounding-carry shift, and unchanged fallbacks. `:app:testDebugUnitTest` green.
- Device-verified on the Assets screen (LUNC).

## Before / After

| Before | After |
|--------|-------|
| before | after |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved formatting for very small fiat prices, making tiny values display more clearly and consistently.
  * Preserved locale-specific currency placement while switching between compact subscript notation and regular decimal output.
  * Kept fee formatting behavior unchanged.

* **Tests**
  * Expanded coverage for tiny price rendering, rounding behavior, and the transition between subscript and plain decimal formats.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/mappers/FiatValueToStringMapper.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/mappers/FiatValueToStringMapper.kt
@@ -3,7 +3,10 @@ package com.vultisig.wallet.ui.models.mappers
 import com.vultisig.wallet.data.models.FiatValue
 import com.vultisig.wallet.data.repositories.AppCurrencyRepository
 import java.math.BigDecimal
+import java.math.MathContext
 import java.math.RoundingMode
+import java.text.AttributedCharacterIterator
+import java.text.CharacterIterator
 import java.text.NumberFormat
 import java.util.Currency
 import javax.inject.Inject
@@ -17,9 +20,11 @@ internal interface FiatValueToStringMapper {
      * extra precision (up to three extra fraction digits) using [RoundingMode.DOWN] so a sub-unit
      * fee like `$0.001234` displays as `$0.00123` instead of being truncated to `$0.00`.
      *
-     * When [asPrice] is true and the value is a sub-unit price, the result is rendered with enough
-     * fraction digits to reveal its first significant figures, so a tiny per-token price like
-     * LUNC's `$0.00006` displays as `$0.00006` instead of collapsing to `$0.00`.
+     * When [asPrice] is true and the value is a sub-unit price, the result is rendered with a few
+     * significant figures instead of collapsing to `$0.00`. Once there are four or more leading
+     * zeros after the decimal point, they are collapsed into compact subscript notation matching
+     * the desktop/extension contract, so `0.00000003` displays as `$0.0₇3` ("seven zeros then 3")
+     * while shallower prices like `0.0001234` keep their plain decimals as `$0.0001234`.
      *
      * All other values fall through to the currency's standard formatting. [asFee] takes precedence
      * over [asPrice] when both are set.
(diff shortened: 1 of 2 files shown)


rkokhatskyi (changes requested):
TBH, this looks ugly for small numbers, I would do it only for more than 6 decimals or so. This format number is not present in any well-known wallet on the market and will only confuse the users. 

cc: @Ehsan-saradar

johnnyluo:
let's leave it as it is

Ehsan-saradar:
Hey @johnnyluo @rkokhatskyi — since I opened the original issue, wanted to share a bit of context on the reasoning here. Just my perspective, the call is yours 🙂

The key thing is that this subscript notation is meant for a token's **unit price**, not for a holding's fiat **balance** — and the two are worth treating differently.

For a **balance** worth a fraction of a cent, extra precision doesn't really help the user, so most wallets don't render it in full — collapsing sub-cent amounts is the common convention (see the [Satcomma Standard](https://medium.com/coinmonks/the-satcomma-standard-89f1e7c2aede) and the [Bitcoin Design guide on units & symbols](https://bitcoin.design/guide/designing-products/units-and-symbols/)).

A token's **price** is a different case. A unit price stays meaningful no matter how small — for evaluating or trading a low-cap token, the exact figure matters. Showing it precisely in a compact form (rather than flattening it) can be a genuinely useful capability.

That's the split several established wallets already use — subscript notation for tiny **prices**:

- **Rabby** — does exactly this in its formatter (subscript for the token price/amount).
- **Ctrl (formerly XDEFI)** — uses the same subscript style for small prices.

Just wanted to lay out why I think the subscript direction is worthwhile — totally happy with whatever you both decide. 🙏

Issue #5117, opened 2026-07-01
[Fix] Jupiter Solana swaps rejected with AccountLoadedTwice — raw tx re-serialized through WalletCore instead of signed verbatim [Swap/Solana]


aminsato:
# [Fix] Jupiter Solana swaps rejected with AccountLoadedTwice — sign raw tx verbatim instead of re-serializing through WalletCore [Swap/Solana]

## Problem
Jupiter swaps on Solana (Fast Vault) are rejected at broadcast with `AccountLoadedTwice` and never land. Jupiter returns a valid v0 (versioned) transaction using an address-table lookup (ALT), which it also simulates server-side successfully. Android's `SolanaHelper.signRawTransaction` re-serializes this tx through WalletCore's proto re-encoder (`TransactionDecoder.decode → Solana.SigningInput.setRawMessage → TransactionCompiler.compileWithSignatures`), and for a v0 tx with ALTs this round-trip does not reproduce Jupiter's message byte-for-byte, producing a duplicate account and the `AccountLoadedTwice` rejection.

## Expected Behavior
Raw/prebuilt Solana transactions (Jupiter swaps, dApp txs) should be signed and broadcast **verbatim** — the original transaction bytes are preserved, only the signature is spliced in. The swap should broadcast successfully and land, matching iOS behavior (`Blockchain/Solana/Signing/Solana.swift:442-538`), which extracts the raw message bytes, signs them, and splices the signature into the original transaction bytes (`replaceSubrange` at the signer-0 offset) without ever re-compiling.

## Steps to Reproduce
1. Open the Android app with a Fast Vault.
2. Initiate a Solana swap that routes through Jupiter (a v0 transaction with an ALT).
3. Sign and broadcast the swap.
4. Observe the broadcast fails: `sendTransaction → {"code":-32002,"message":"Transaction simulation failed","data":{"err":"AccountLoadedTwice","logs":[]}}` and the transaction never lands.

### Must Do
- [ ] In the raw path (`signRawTransaction` / `getPreSignedImageHashForRaw`), extract the message bytes, sign them, and splice the 64-byte signature into the original transaction bytes at signer index 0, then broadcast verbatim — mirroring iOS.
- [ ] Stop routing raw/prebuilt txs through `setRawMessage` + `compileWithSignatures` (the WalletCore re-encoder round-trip). Reuse the existing shortvec message-extraction helper (from the tx-hash work) to locate the splice offset.

### Must NOT Do
[...]

PR #5130, opened 2026-07-01
fix(solana): honor Jupiter swap simulationError instead of signing a doomed tx

### LINKED ISSUE #5118: [Fix] Jupiter Solana swap ignores simulationError — signs/broadcasts a tx Jupiter already flagged as failing [Swap/Solana]
## Observed

A Jupiter swap on Solana is signed and broadcast even when Jupiter has already **pre-simulated the transaction and reported it will fail**. The `/jup/swap/v1/swap` response carries a `simulationError`, but the app ignores it and proceeds to keysign + broadcast a doomed transaction.

Captured `/jup/swap/v1/swap` response (PYTH → SOL, 3-hop Meteora → Raydium CLMM → Manifest route):

```json
{
  "swapTransaction": "…",
  "lastValidBlockHeight": 408114842,
  "computeUnitLimit": 1400000,
  "simulationError": {
    "errorCode": "TRANSACTION_ERROR",
    "error": "Error processing Instruction 2: custom program error: 0x1789"
  }
}
```

`0x1789 = 6025` — a custom program error from one of the route AMMs (a price/limit/liquidity condition at execution). Jupiter is telling us the tx won't succeed; the user still gets asked to sign it, and it fails at broadcast.

## Cause

The Jupiter swap response DTO parses only the transaction bytes and drops every other field, including `simulationError`:

- `data/.../api/models/quotes/JupiterSwapQuoteJson.kt:16,21` — `quoteSwapData` → `QuoteSwapTransactionJson(@SerialName("swapTransaction") val data: String)`. No `simulationError`, `lastValidBlockHeight`, or `computeUnitLimit`.
- `data/.../repositories/swap/JupiterQuoteSource.kt` consumes only `quote.swapTransaction.data`.

So there is no point at which a non-null `simulationError` can stop the flow before signing.

## Fix shape

1. Add `simulationError` (and `lastValidBlockHeight` / `computeUnitLimit`) to the Jupiter swap response model.
2. When `simulationError` is non-null, do **not** proceed to sign — re-quote or surface a "route temporarily unavailable / try again" state instead of asking the user to sign a transaction Jupiter already flagged as failing.

## Why this matters more now

- With client-side `skipPreflight` on the Solana broadcast (#5110), a transaction that fails simulation is no longer rejected at preflight — it is forwarded to the leader and **lands on-chain as a failed tx that still pays the fee**. Honoring `simulationError` before signing prevents paying for a known-failing swap.
- Independent of #5117 (which is the structural `AccountLoadedTwice` from re-serializing Jupiter's tx). This issue is about execution validity: even a structurally-correct tx should not be signed when Jupiter says it will fail.

## Cross-platform

Verify iOS/Windows honor Jupiter's `simulationError` before signing; if not, mirror there.

Related: #5117 (AccountLoadedTwice / re-serialization), #5110 (skipPreflight), #5099 (broadcast error copy).

---
_Verified against vultisig-android `main`: the response field is dropped by `QuoteSwapTransactionJson`. Response captured from the live `/jup/swap/v1/swap` call._

aminsato:
Closes #5118

## Problem

Jupiter's `/jup/swap/v1/swap` response **pre-simulates the built transaction** and returns a non-null `simulationError` when it will fail on-chain (slippage / min-out / liquidity at execution). The app dropped that field, so it proceeded to keysign + broadcast a transaction Jupiter already flagged as doomed.

Captured (PYTH → SOL, Meteora → Raydium CLMM → Manifest):
```json
{ "swapTransaction": "…",
  "simulationError": { "errorCode": "TRANSACTION_ERROR",
                       "error": "Error processing Instruction 2: custom program error: 0x1789" } }
```

## Fix

- Add `simulationError` to `QuoteSwapTransactionJson` (the `/swap` response DTO).
- In `JupiterApiImpl.getSwapQuote`, when `simulationError` is non-null, throw `SwapException.SwapRouteNotAvailable` **before** building the tx. `swapApiCall` passes typed `SwapException`s through unchanged, so this drops just the Jupiter route — the picker re-quotes or falls back to another provider, and the user sees "route not available" (`R.string.swap_route_not_available`) instead of being asked to sign a doomed tx.

The throw fires right after parsing the `/swap` body, before the WalletCore compute-budget step, so no native JNI is involved on the failing path.

## Why it matters now

With client-side `skipPreflight` on the Solana broadcast (#5110), a sim-failing tx is no longer rejected at preflight — it lands on-chain as a **failed tx that still pays the fee**. Honoring `simulationError` before signing prevents paying for a known-failing swap. Independent of #5117 (structural `AccountLoadedTwice`).

## Test plan

- [x] `JupiterApiTest` passes, incl. a new case asserting a `/swap` `simulationError` throws `SwapRouteNotAvailable` before the native compute-budget step
- [ ] Manual: a Jupiter route whose `/swap` returns `simulationError` is dropped from the picker (no sign prompt), and a healthy route still swaps

## Cross-platform

Verify iOS/Windows honor Jupiter's `simulationError` before signing; mirror if not.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved swap handling by detecting simulation errors in Jupiter swap responses.
  - If a swap response includes a simulation error, the app now stops and marks the route as unavailable instead of proceeding.
  - Added automated test coverage to ensure routes with simulation errors are correctly rejected.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
@@ -110,8 +110,8 @@ constructor(
         // Jupiter pre-simulates the swap tx and reports a non-null `simulationError` when it will
         // fail on-chain (slippage / min-out / liquidity at execution). Don't build or offer a tx
         // Jupiter already knows is doomed — drop the Jupiter route so the picker re-quotes or falls
-        // back to another provider. This matters especially with client-side `skipPreflight`, where
-        // a sim-failing tx would still land on-chain and burn the fee.
+        // back to another provider, instead of taking the user through the full keysign only for
+        // the broadcast to be rejected at preflight.
         quoteSwapData.simulationError?.let { simError ->
             throw SwapException.SwapRouteNotAvailable(
                 "[Jupiter] swap simulation failed: ${simError.error ?: simError.errorCode ?: "unknown"}"
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
@@ -107,6 +107,17 @@ constructor(
         }
         val quoteSwapData = swapResponse.bodyOrThrow<QuoteSwapTransactionJson>()
 
+        // Jupiter pre-simulates the swap tx and reports a non-null `simulationError` when it will
+        // fail on-chain (slippage / min-out / liquidity at execution). Don't build or offer a tx
+        // Jupiter already knows is doomed — drop the Jupiter route so the picker re-quotes or falls
+        // back to another provider. This matters especially with client-side `skipPreflight`, where
+        // a sim-failing tx would still land on-chain and burn the fee.
+        quoteSwapData.simulationError?.let { simError ->
+            throw SwapException.SwapRouteNotAvailable(
+                "[Jupiter] swap simulation failed: ${simError.error ?: simError.errorCode ?: "unknown"}"
+            )
+        }
+
         // Jupiter never initializes the `feeAccount` ATA, so prepend an idempotent create for it
         // the
         // first time per fee mint (the payer funds the ~0.002 SOL rent; it is skipped once the ATA
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/models/quotes/JupiterSwapQuoteJson.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/models/quotes/JupiterSwapQuoteJson.kt
@@ -18,7 +18,19 @@ data class QuoteSwapTotalDataJson(
     @SerialName("routePlan") val routePlan: List<RoutePlanItemJson>,
 )
 
-@Serializable data class QuoteSwapTransactionJson(@SerialName("swapTransaction") val data: String)
+@Serializable
+data class QuoteSwapTransactionJson(
+    @SerialName("swapTransaction") val data: String,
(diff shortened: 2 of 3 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -107,6 +107,17 @@ constructor(
         }
         val quoteSwapData = swapResponse.bodyOrThrow<QuoteSwapTransactionJson>()
 
+        // Jupiter pre-simulates the swap tx and reports a non-null `simulationError` when it will
+        // fail on-chain (slippage / min-out / liquidity at execution). Don't build or offer a tx
+        // Jupiter already knows is doomed — drop the Jupiter route so the picker re-quotes or falls
+        // back to another provider. This matters especially with client-side `skipPreflight`, where
```
rkokhatskyi: `skipPreflight` claim here is stale because `SolanaApi.kt:207` only sets `preflightCommitment: confirmed`, since #5110's merged commit walked skipPreflight back, so the RPC still preflight-simulates and a doomed route should get rejected before paying the fee

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
aminsato: Good catch — that claim is stale. #5110 shipped with preflight left on (`preflightCommitment: confirmed` at `SolanaApi.kt:207`), so a sim-failing tx is rejected at preflight and no fee is burned. Reworded in f09384962: the actual benefit is dropping the route at quote time so the user isn't taken through a full keysign only for the broadcast to be rejected at preflight.

johnnyluo (approved):
LGTM

Issue #4981, opened 2026-06-20
Rebrand TON native token to GRAM (ticker + logo)


Ehsan-saradar:
## Summary

The Open Network community voted (81.22% in favor) to rename the native token **Toncoin (TON)** to **Gram (GRAM)**, effective **2026-06-15**. This is a cosmetic **token** rebrand only — name, ticker, and logo. The **blockchain stays "The Open Network"** and there is **no swap/bridge/claim/migration**: addresses, balances, contracts, NFTs, and staking are unchanged.

This is a cross-platform effort; the data source-of-truth is commondata.

### Do (display-facing)
- [ ] Native TON coin in `data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt` (~L2965): `ticker = "TON"` → `"GRAM"`
- [ ] Add a **GRAM coin drawable** and set `logo = "ton"` → `logo = "gram"` on that coin
- [ ] Audit user-facing strings that hardcode "TON" for the token amount/name

### Decouple the icon (keep the chain logo)
The chain icon must stay TON. `ChainLogo.kt` maps `Chain.Ton -> R.drawable.ton` (and `ton_mono`). Keep `res/drawable/ton.xml` / `ton_mono.xml` for the **chain**, and add a new `gram` drawable for the **coin** logo only.

### Do NOT change (protocol / external identifiers)
- `Chain.Ton` enum value / chain kind — protobuf identity shared cross-platform
- `priceProviderID = "the-open-network"` — keep until CoinGecko renames its id (changing early breaks prices)
- `decimal = 9`, `isNativeToken = true`
- SwapKit / Banxa asset codes that use `"TON"` (e.g. `SwapKitQuoteSource.kt`, `SwapProviderTable.kt`) — third-party identifiers, stay until those providers rebrand

## Acceptance
- Wallet displays GRAM name/ticker/logo for the native token; chain icon + name stay TON ("The Open Network")
- No "swap/migrate TON to GRAM" prompt (community flagged such offers as scams)

## Related
- Data source-of-truth: vultisig/commondata#92
- SDK: vultisig/vultisig-sdk#785
- Windows/extension: vultisig/vultisig-windows#4160 (ref #4129)

PR #5133, opened 2026-07-01
feat(ton): TON nominator-pool staking on the DeFi tab (#5041)

### LINKED ISSUE #5041: [Parity] Native TON nominator-pool staking on the DeFi/Earn tab

aminsato:
### Stake
  https://tonviewer.com/transaction/678a9e99b5a4a3ff65bfb3d09792fd2dde53c8394fa4b22c3ae8399e5eff5cb2
  ### Unstake
 https://tonviewer.com/transaction/b5dcf64dabf0d996208fdf6654d87d2492d96db5b7d08a7a1d9cc48f7f22ab84
 ##

Follow-up to #5073 (merged data + transaction core) that lands the **TON nominator-pool staking UI on the DeFi tab**, ported to match the iOS implementation (vultisig/vultisig-ios#4654). Closes #5041.

TON staking now lives **exclusively on the DeFi tab** (Stake / Unstake), and is removed from the generic Functions/Deposit flow — mirroring iOS, which deleted TON from the Functions flow.

### DeFi positions screen
- New `PositionTon` route + ChainDashboard wiring; `Chain.Ton` added to `isDeFiSupported` / crypto-connection sets.
- Reads the primary nominator pool (name, APY, staked amount) with a **pending-withdrawal lock** + unlock countdown.
- Always renders the position card (zeroed with **Unstake disabled** at 0 staked), so both Stake and Unstake are present — matching iOS/macOS.

### Dedicated Stake screen
- Centered amount + 25/50/75/Max chips + balance available.
- Inline **Pool** row → pool picker sheet (verified nominator pools — `whales`/`tf`, excludes `liquidTF` — sorted by APY, with search).
- Dynamic **"Minimum stake is X TON"** (pool minimum + ~1 TON deposit buffer).

### Dedicated Unstake screen
- Full-withdrawal confirmation (no amount): staked amount + pool address, full-withdrawal notice, insufficient-fee guard.

### Reuse & removal
- Both screens reuse the fund-safe `buildTonStakingTransaction` core (bounceable send, per-implementation comment, min-stake + buffer, fixed 0.2 TON withdraw fee) + `DepositGasFeeHelper`, routing to the existing verify / keysign path — no new crypto or fee code.
- Removed `Chain.Ton` from `isDepositSupported` and dropped Stake/Unstake from the generic deposit option list (this was the source of the "Address is invalid" error when no pool was pasted).

### Localization
- All new strings translated across the 10 supported locales.

## Tests
- `TonDeFiPositionsViewModelTest` (position rendering, lock, stake/unstake routing), `TonStakeViewModelTest` (pool filter/sort/capacity/scale), updated deposit + DeFiNavActions tests.
- `:app:testDebugUnitTest` + `:data:testDebugUnitTest` green; ktfmt clean.

## Test plan
- [x] Unit tests green (`:app` + `:data`)
- [x] ktfmt clean
- [ ] Manual on-device: stake into a Whales pool + unstake (mainnet MPC) — pending
- [ ] Before/after screenshots — pending device verification

## Notes
- **Draft** pending on-device keysign verification + before/after screenshots.
- Native TON coin displays as **GRAM** on Android (per the #4981 rebrand), so the screens read "GRAM" where iOS shows "TON".

## Summary

* **New Features**
  * Added TON DeFi positions with dedicated TON staking and unstaking screens, including pool search/selection, APY display, and a live unlock countdown for pending withdrawals.
  * Added TON-specific staking/unstaking navigation and actions.
* **Bug Fixes**
  * Ensured TON consistently uses the DeFi staking flow (not the generic deposit/send flow), with correct chain-dashboard and vault account routing.
  * Improved staking transaction info so the verify screen can show pool/node details.
* **Localization**
  * Added and updated TON staking/DeFi UI strings across multiple languages.
* **Tests**
  * Added/updated unit tests for TON positions, staking/unstaking behavior, filtering, and navigation.


app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
```diff
@@ -0,0 +1,327 @@
...
+
+    private suspend fun buildPositionUiModel(
+        staked: BigInteger,
+        pendingWithdraw: BigInteger,
+        poolAddress: String,
+        coin: Coin,
+        price: BigDecimal,
+        currencyFormat: NumberFormat,
+    ): TonStakingUiModel {
+        val poolInfo = runCatching { tonStakingApi.getStakingPool(poolAddress) }.getOrNull()
```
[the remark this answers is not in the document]
aminsato: Fixed in e298b1e. The lookup now catches \`CancellationException\` separately and rethrows it; only genuine missing-pool cases (other exceptions) degrade to \`null\` for the display fallback, so a cancelled reload can no longer swallow the error and publish stale \`Success\` state.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonStakeViewModel.kt
```diff
@@ -0,0 +1,364 @@
...
+        ) {
+            _state.update { it.copy(isLoadingPools = true) }
+            val decimals = coin?.decimal ?: com.vultisig.wallet.data.models.Coins.Ton.TON.decimal
+            allPools =
+                withContext(ioDispatcher) {
+                    filterAndSortPools(tonStakingApi.getStakingPools(), decimals)
+                }
+            _state.update { it.copy(isLoadingPools = false, pools = allPools) }
+            prefillExistingPool(decimals)
+        }
```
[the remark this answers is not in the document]
aminsato: Fixed in e298b1e. \`loadPools()\`'s \`onError\` now clears \`isLoadingPools\`, so on failure the picker falls back to its empty state ("no pools found") instead of spinning forever.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonUnstakeViewModel.kt
```diff
@@ -0,0 +1,198 @@
...
+                            )
+                        }
+                    }
+                    .getOrNull()
+            val required =
+                BigDecimal(
+                        TonNominatorPool.WITHDRAW_FEE + (gasFee?.value ?: java.math.BigInteger.ZERO)
+                    )
+                    .movePointLeft(nativeCoin.decimal)
+            val balance =
```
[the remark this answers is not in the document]
aminsato: Fixed in e298b1e. \`loadCoin()\` now fails closed: a null vault or missing native TON coin sets \`hasSufficientBalance = false\` + an explicit error, and the gas-fee call is no longer wrapped in \`runCatching\` so a failure propagates to \`onError\` (which also disables the action) instead of defaulting gas to zero and understating the requirement. The screen suppresses the generic insufficient-balance hint while a hard error is shown.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/ton/TonStakeScreen.kt
```diff
@@ -0,0 +1,378 @@
...
+                        .padding(horizontal = 12.dp, vertical = 12.dp),
+            ) {
+                UiIcon(
+                    drawableResId = R.drawable.ic_search,
+                    size = 16.dp,
+                    tint = Theme.v2.colors.text.tertiary,
+                )
+                BasicTextField(
+                    value = searchTextFieldState.text.toString(),
+                    onValueChange = onSearchQueryChange,
```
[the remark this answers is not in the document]
aminsato: Fixed in e298b1e. Switched the search field to the state-based \`BasicTextField(state = searchTextFieldState, ...)\` and drive filtering from \`searchTextFieldState.textAsFlow()\` in the ViewModel, so the typed query is written back into the bound state and survives recomposition.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/ton/TonUnstakeScreen.kt
```diff
@@ -0,0 +1,130 @@
...
+                            .clip(RoundedCornerShape(16.dp))
+                            .background(Theme.v2.colors.backgrounds.secondary)
+                            .border(1.dp, Theme.v2.colors.border.light, RoundedCornerShape(16.dp))
+                            .padding(16.dp),
+                    verticalArrangement = Arrangement.spacedBy(12.dp),
+                ) {
+                    InfoRow(
+                        title = stringResource(R.string.ton_defi_staked_amount, ticker),
+                        value = state.stakedDisplay,
+                    )
```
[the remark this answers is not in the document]
aminsato: Not changing this one. \`ton_defi_staked_amount\` ("Staked %1\$s") is used identically on \`TonDeFiPositionsScreen.kt\` (also passing \`ticker\`) as a "Staked TON" **label** paired with a separate value cell — here the value cell is \`state.stakedDisplay\`. The placeholder contract in this codebase is the ticker, not the amount; the translations ("Gestaked %1\$s" / "В стейкинге %1\$s") read correctly as "Staked TON". Passing \`stakedDisplay\` here would render "Staked 51 GRAM   51 GRAM" and diverge from the positions screen. Leaving the ticker + separate value cell as-is for consistency.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/StakeStrategy.kt
```diff
@@ -180,6 +180,9 @@ internal suspend fun buildTonStakingTransaction(
         srcToken = selectedToken,
         srcAddress = srcAddress,
         dstAddress = dstAddress,
+        // The verify screen surfaces the pool as the Node Address (iOS/macOS parity); without it
+        // the row is hidden. It equals the bounceable destination for a nominator-pool message.
+        nodeAddress = dstAddress,
```
rkokhatskyi, line 185: TON stake and unstake build a DepositTransaction with isDeposit true, and the joining device's allow-list in `JoinDepositUiModelBuilder.kt:45-52` only accepts MayaChain, THORChain, Ethereum, Cosmos and UTXO, so a Secure Vault co-signer hits its else branch and throws before it can even render a verify screen, blocking every multi-device TON staking ceremony

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
```diff
@@ -294,6 +296,8 @@ val Chain.isDeFiSupported: Boolean
             Chain.MayaChain,
             Chain.Ethereum,
             Chain.Tron,
+            // TON surfaces a nominator-pool staking positions view (stake / unstake).
+            Chain.Ton,
```
rkokhatskyi, line 300: Chain.Ton is added to isDeFiSupported here but the private Coin.isValidForDeFi() allowlist in `AccountsRepository.kt:601` and its defiChainNativeCoins list at `AccountsRepository.kt:422` are never updated, so the vault's aggregate Portfolio DeFi tab drops the TON coin before building any account row and never includes the staked TON value in the total

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonUnstakeViewModel.kt
```diff
@@ -0,0 +1,202 @@
...
+import timber.log.Timber
+import wallet.core.jni.TONAddressConverter
+
+@Immutable
+internal data class TonUnstakeUiState(
+    val ticker: String = "",
+    val poolAddress: String = "",
+    val stakedDisplay: String = "",
+    /** Liquid balance covers the 0.2 TON withdraw signal + network fee. */
+    val hasSufficientBalance: Boolean = true,
```
rkokhatskyi: hasSufficientBalance defaults to true with no isLoading field, so TonUnstakeScreen renders Continue as Enabled on first composition before `loadCoin()` resolves, and submit's guard at `TonUnstakeViewModel.kt:100` reads that same stale default, letting a fast tap push a withdraw the account cannot cover

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonStakeViewModel.kt
```diff
@@ -0,0 +1,374 @@
...
+                    )
+            val nativeCoin =
+                vault.coins.firstOrNull { it.chain == Chain.Ton && it.isNativeToken }
+                    ?: return@safeLaunch setError(
+                        R.string.ton_defi_error_ton_not_in_vault.asUiText()
+                    )
+            coin = nativeCoin
+
+            val gasReservation =
+                BigDecimal(TonNominatorPool.WITHDRAW_FEE).movePointLeft(nativeCoin.decimal)
```
rkokhatskyi: stakeableBalance reserves TonNominatorPool.WITHDRAW_FEE, 0.2 TON, the unstake message's carried amount, instead of the real deposit fee this same PR set at `TonFeeService.kt:50`, 0.05 TON matching iOS defaultFee, so a wallet in the 0.15 TON band above the pool minimum gets clamped below requiredMinStake and Continue stays disabled

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
```diff
@@ -0,0 +1,338 @@
...
+            totalAmountPrice = stakedFiat,
+            ticker = coin.ticker,
+            poolName = poolInfo?.name?.takeIf { it.isNotBlank() } ?: shortPool(poolAddress),
+            stakedDisplay = "${stakedTon.stripTrailingZeros().toPlainString()} ${coin.ticker}",
+            stakedFiatDisplay = stakedFiat,
+            // tonapi `apy` is a percentage (13.27 = 13.27%); `formatPercentage` multiplies by 100,
+            // so scale to a fraction first.
+            apy = poolInfo?.apy?.let { (it / 100).formatPercentage() },
+            hasPosition = true,
+            isActionLocked = isLocked,
```
rkokhatskyi, line 280: isActionLocked is a point-in-time snapshot from the last tonapi fetch with no reset-to-loading and no polling, so a resume-triggered reload right after submitting a Withdraw can still read the old unlocked value, and `onUnstake()`'s own check at line 318 trusts that stale state, letting a second Withdraw fire early

app/src/main/res/values/strings.xml
```diff
@@ -474,6 +474,23 @@
     <string name="error_invalid_chain">Chain is invalid</string>
     <string name="ton_stake_error_unsupported_pool">This staking pool isn\'t supported</string>
     <string name="ton_stake_error_min_amount">Minimum stake for this pool is %1$s TON</string>
```
rkokhatskyi: `ton_stake_error_min_amount` hardcodes the literal word TON while `TonStakeScreen.kt:58` interpolates the coin's real ticker, which for Android's native TON coin is GRAM per `Coins.kt:2965`, so the same screen shows Stake GRAM in the header and Minimum stake for this pool is 51 TON right below it

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonStakeViewModel.kt
```diff
@@ -0,0 +1,374 @@
...
+                }
+            _state.update { it.copy(isLoadingPools = false, pools = allPools) }
+            prefillExistingPool(decimals)
+        }
+    }
+
+    /** For an add-more stake, preselect the position's pool (fetch it if capacity-filtered out). */
+    private suspend fun prefillExistingPool(decimals: Int) {
+        val poolAddress = route.poolAddress?.takeIf { it.isNotBlank() } ?: return
+        if (_state.value.selectedPool != null) return
```
rkokhatskyi, line 313: prefillExistingPool checks selectedPool != null only once before the async `getStakingPool` call, so if the user taps a different pool in the picker while that call is in flight, the unconditional `onPoolSelected(existing)` at line 317 overwrites the manual pick back to the old pool with no re-check and no warning

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
```diff
@@ -0,0 +1,338 @@
...
+            else current
+        }
+    }
+
+    /**
+     * Opens the dedicated Stake screen. An existing position prefills its pool (add-more); a
+     * first-time stake leaves the pool unset so the user picks one there — mirroring macOS "Stake
+     * TON".
+     */
+    fun onStake() {
```
rkokhatskyi, line 357: `onStake()` never reads isActionLocked while `onUnstake()` does at line 318, breaking the documented invariant that a pending withdrawal blocks both actions, and only the button's own disabled state in the screen enforces it today, so any future second stake entry point would silently reintroduce the race this guard exists to prevent

app/src/test/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModelTest.kt
```diff
@@ -0,0 +1,224 @@
...
+            TonStakingPoolInfoJson(name = "Whales", apy = 5.0, cycleEnd = 9_999_999_999L)
+
+        val vm = createViewModel().also { it.setData(VAULT_ID) }
+
+        val data = (vm.state.value as TonDeFiUiState.Success).tonData
+        assertTrue(data.isActionLocked)
+
+        // A locked position must not navigate into the unstake flow.
+        vm.onUnstake()
+        coVerify(exactly = 0) { navigator.route(any<Route.Deposit>()) }
```
rkokhatskyi: This locked-position test verifies `navigator.route(any<Route.Deposit>())` was never called, but onUnstake's real navigation target when unlocked is Route.TonUnstake per `TonDeFiPositionsViewModel.kt:326`, never Route.Deposit, so the assertion is vacuous and would still pass with the isActionLocked guard deleted, leaving the pending-withdrawal lock on unstake untested

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonUnstakeViewModel.kt
```diff
@@ -0,0 +1,202 @@
...
+                    Route.VerifyDeposit(vaultId = route.vaultId, transactionId = transaction.id)
+                )
+                _state.update { it.copy(isSubmitting = false) }
+            } catch (e: InvalidTransactionDataException) {
+                setError(e.text)
+            }
+        }
+    }
+
+    private fun loadCoin() {
```
rkokhatskyi, line 147: loadCoin's fail-closed logic, setting hasSufficientBalance false on a null vault, a missing native coin, or a failed gas-fee call, has zero test coverage anywhere in the suite, unlike sibling TonStakeViewModel and TonDeFiPositionsViewModel, so a future edit reintroducing a swallowed exception or a defaulted zero fee would silently re-enable a doomed withdraw

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
```diff
@@ -0,0 +1,338 @@
...
+        coin: Coin,
+        price: BigDecimal,
+        currencyFormat: NumberFormat,
+    ): TonStakingUiModel {
+        // Pool metadata (name/apy/cycle) is display-only, so a genuine lookup miss degrades to a
+        // short fallback label. Cancellation must still propagate — otherwise a cancelled reload
+        // could swallow the exception and let this coroutine publish a stale Success state.
+        val poolInfo =
+            try {
+                tonStakingApi.getStakingPool(poolAddress)
```
rkokhatskyi, line 247: buildPositionUiModel fetches poolInfo.implementation but only uses it for name, apy and cycleEnd, never checking TonNominatorPool.isNominatorImplementation the way the Stake picker's filterAndSortPools does, so an existing Tonstakers or liquidTF position renders hasPosition true with both Stake and Unstake enabled, only failing deep inside buildTonStakingTransaction after the user completes the whole flow

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
```diff
@@ -0,0 +1,338 @@
...
+            ticker = coin.ticker,
+            poolName = poolInfo?.name?.takeIf { it.isNotBlank() } ?: shortPool(poolAddress),
+            stakedDisplay = "${stakedTon.stripTrailingZeros().toPlainString()} ${coin.ticker}",
+            stakedFiatDisplay = stakedFiat,
+            // tonapi `apy` is a percentage (13.27 = 13.27%); `formatPercentage` multiplies by 100,
+            // so scale to a fraction first.
+            apy = poolInfo?.apy?.let { (it / 100).formatPercentage() },
+            hasPosition = true,
+            isActionLocked = isLocked,
+            pendingWithdrawDisplay =
```
rkokhatskyi, line 281: pendingWithdrawDisplay is computed here with real formatting logic, decimals, ticker, trailing-zero stripping, but no Compose screen reads it, TonUnlockNotice only renders the countdown from unlockEpochMs, so during a pending withdrawal the user sees a countdown to unlock with no indication of how much TON is actually pending

app/src/test/java/com/vultisig/wallet/ui/models/defi/TonStakeViewModelTest.kt
```diff
@@ -0,0 +1,101 @@
+package com.vultisig.wallet.ui.models.defi
+
+import com.vultisig.wallet.data.api.chains.ton.TonStakingPoolEntryJson
+import com.vultisig.wallet.data.models.Coins
+import java.math.BigDecimal
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import org.junit.jupiter.api.Test
+
+internal class TonStakeViewModelTest {
```
rkokhatskyi, line 39: Every test here calls the static filterAndSortPools helper or builds a bare TonPoolUiModel, none construct TonStakeViewModel itself, so loadCoin's stakeableBalance computation, the one reserving TonNominatorPool.WITHDRAW_FEE instead of the real deposit fee, has no regression test, unlike sibling TonDeFiPositionsViewModelTest which does instantiate its ViewModel with mocked dependencies

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
```diff
@@ -0,0 +1,338 @@
...
+        val tempSelectedPositions: List<String> = listOf(TON_KEY),
+    ) : TonDeFiUiState
+}
+
+/**
+ * View-model for the native TON nominator-pool staking position on the DeFi/Earn tab.
+ *
+ * Reads the account's pools from tonapi, treats the largest as the primary position (mirrors
+ * [com.vultisig.wallet.data.blockchain.ton.TonDeFiBalanceService] / vultisig-ios
+ * `TonStakeInteractor`), and decorates it with the pool name + APY. Stake/Unstake route into the
```
rkokhatskyi: This KDoc still says Stake/Unstake route into the existing Deposit flow, but onStake and onUnstake now navigate to the dedicated Route.TonStake and Route.TonUnstake screens this PR adds, never Route.Deposit, so the comment describes the pre-PR architecture and could mislead a future maintainer into thinking DepositStrategyFactory still owns TON staking

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+
+    /**
+     * A pending withdrawal blocks both stake and unstake until the pool releases the balance. Read
+     * the lock from the latest loaded state, and treat an in-flight reload as locked so a
+     * resume-triggered refresh right after a Withdraw can't be raced by a stale unlocked snapshot.
+     */
+    private fun isActionLocked(): Boolean {
+        if (loadJob?.isActive == true) return true
```
rkokhatskyi, line 348: isActionLocked() now also blocks both actions whenever loadJob is active, but the button's enabled state in TonDeFiPositionsScreen.kt still reads only the stale tonData.isActionLocked snapshot, so on every resume-triggered reload the button renders Enabled while a tap silently does nothing

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
```diff
@@ -0,0 +1,371 @@
...
+        // A position in a pool this app can't stake into (liquid-staking Tonstakers/`liquidTF`)
+        // must not offer Stake/Unstake — the `"d"`/`"w"` text-comment deposit only fails deep in
+        // buildTonStakingTransaction. Mirror the Stake picker's filterAndSortPools filter. A
+        // metadata miss (null implementation) is treated permissively so a transient tonapi failure
+        // can't hide a genuine nominator position.
+        if (
+            poolInfo?.implementation != null &&
+                !TonNominatorPool.isNominatorImplementation(poolInfo.implementation)
+        ) {
+            return null
```
rkokhatskyi, line 263: For a Tonstakers liquidTF position this returns null so the card renders hasPosition=false, hiding a real nonzero stake, while TonDeFiBalanceService.getRemoteDeFiBalance still sums the same stakedTotal into the vault's Portfolio DeFi total with no implementation filter, so the same stake reads as zero here and nonzero there

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/StakeStrategy.kt
rkokhatskyi: Fixed in ea0498bfe, BlockChainSpecific.Ton added to the allow-list. Traced every downstream call in build() (fee resolver, fee service composite, memo parser, both mappers) and none has a hidden chain gate that would still throw. Worth confirming with a real 2-device stake/unstake before merge since this class of bug only fully proves out at runtime.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
rkokhatskyi: Fixed in ea0498bfe, Chain.Ton added to both isValidForDeFi() and defiChainNativeCoins. Traced the full pipeline to VaultAccountsViewModel's totalDeFiValue, no other gate blocking it.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonUnstakeViewModel.kt
rkokhatskyi: Fixed in ea0498bfe, isLoading now defaults true, hasSufficientBalance defaults false, and submit() gates on isLoading too. All three loadCoin() exit paths correctly reset isLoading.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonStakeViewModel.kt
rkokhatskyi: Fixed in ea0498bfe, now calls depositGasFeeHelper.calculateGasFee for the real ~0.05 TON deposit fee, matching iOS's TonHelper.defaultFee exactly, and pinned by a new construction test.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
rkokhatskyi: The described race is fixed, isActionLocked() now also treats an in-flight reload as locked and guards both actions. But the fix introduces a new issue, see the new comment on isActionLocked() itself, the button's visual enabled state doesn't know about the in-flight reload so it can render Enabled while a tap silently no-ops.

app/src/main/res/values/strings.xml
rkokhatskyi: Fixed for this string, both call sites now pass the ticker, confirmed correct in all 10 locales. Note ton_staking_min_stake, the pool-picker row caption, still hardcodes TON, a smaller leftover of the same class, not blocking.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonStakeViewModel.kt
rkokhatskyi: Fixed in ea0498bfe, the null-check on selectedPool now happens after the async lookup, and since this runs on Main with no suspension between the check and the write it's a full fix, not just a narrower window.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
rkokhatskyi: Symmetry restored, onStake() now calls the same isActionLocked() as onUnstake(). See the new comment on isActionLocked() though, the fix that restores this symmetry also introduces a button-state desync during reloads.

app/src/test/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModelTest.kt
rkokhatskyi: Fixed, the assertion now checks Route.TonUnstake instead of Route.Deposit, and a new onStake-locked case was added alongside it.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonUnstakeViewModel.kt
rkokhatskyi: Fixed, TonUnstakeViewModelTest.kt now covers all the fail-closed loadCoin paths. One small residual, no test calls submit() itself to pin the isLoading guard directly, though the Compose button already gates the only real entry point.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
rkokhatskyi: Fixed for the original bug, buildPositionUiModel now drops non-nominator implementations instead of offering a doomed Stake/Unstake. New comment on the return null line though, this can now hide a real stake that TonDeFiBalanceService still counts in the aggregate Portfolio total.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
rkokhatskyi: Fixed, now rendered inside TonUnlockNotice alongside the countdown, no overlap.

app/src/test/java/com/vultisig/wallet/ui/models/defi/TonStakeViewModelTest.kt
rkokhatskyi: Fixed, there's now a real construction test pinning the corrected fee reservation (0.95 for a 1 TON balance with a 0.05 TON fee stub), which would fail under the old WITHDRAW_FEE constant.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
rkokhatskyi: Fixed, accurate now. Minor nit: the closing clause attributes the transaction core to buildPositionUiModel's sibling view-models, but buildPositionUiModel is a private helper with no siblings, the actual siblings are TonStakeViewModel/TonUnstakeViewModel.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
aminsato: Fixed in f4747cc. The reload state is now reflected in the UI: added `isReloading` to `TonDeFiUiState.Success` (set when a reload starts over an already-rendered screen, cleared on completion or on a background-refresh failure). The card now disables Stake/Unstake on `tonData.isActionLocked || state.isReloading`, so the buttons move in lockstep with `isActionLocked()`'s `loadJob` guard rather than rendering Enabled over a stale snapshot. The unlock notice still keys off `data.isActionLocked` alone so it only tracks a genuine withdrawal.

app/src/main/java/com/vultisig/wallet/ui/models/defi/TonDeFiPositionsViewModel.kt
aminsato: Fixed in f4747cc. `TonDeFiBalanceService.getRemoteDeFiBalance` now applies the same nominator-implementation filter the ViewModel uses — reusing the `getStakingPool` lookup already made for APY (renamed `fetchPoolApr` → `fetchPoolInfo`). A `liquidTF` primary pool is treated as zero staked so the Portfolio DeFi total matches the DeFi screen, and a metadata miss (null implementation) is treated permissively so a transient tonapi failure can't erase a genuine nominator position. Added a unit test covering the liquidTF exclusion.

johnnyluo (approved):
LGTM

PR #5143, opened 2026-07-02
fix(solana): Jupiter fee-ATA prepend breaks SOL-output swaps (AccountLoadedTwice)

aminsato:
###
## Tranasction hash
  - http://orbmarkets.io/tx/yX9eJa5RfDjawRpeSKoz8FQp4sQDXJ2TB4aYApMN8J1qqcZ628SsXAsM3YNytCDrd7yTZh8tYCzXkCzqgHmPZ3Y

Swapping any SPL token to native SOL (e.g. ~1 USDC → SOL on a Fast vault) fails at broadcast with:

```
Transaction simulation failed: AccountLoadedTwice
```

Keysign completes; the RPC rejects the signed transaction.

## Root cause

Reproduced on an emulator, decoded the failing signed transaction, and resolved its address lookup tables on-chain:

- The affiliate fee is taken in the swap's output mint — wrapped SOL for SOL-output swaps.
- The fee owner's wSOL ATA doesn't exist, so `prependCreateFeeAta` injected an idempotent create-ATA instruction via wallet-core's `SolanaTransaction.insertInstruction`.
- `insertInstruction` appends the instruction's accounts as **new static keys without deduplicating against the transaction's address lookup tables**. Jupiter's route already loads the wSOL mint (`So1111…1112`) through an ALT, so the same account ends up loaded twice — the Solana runtime rejects the transaction before execution.

Decoded evidence from the rejected tx: static key `[17]` = wSOL mint (from our injected instruction) while ALT `5oRF9x…X7Fq` readonly index 0 resolves to the same mint.

Since the rejected transaction is the very one that would create the fee ATA, the pair can never recover: every token→SOL swap fails, forever.

## Fix

Align with iOS (`JupiterService.swift`), which never injects instructions into Jupiter's transaction:

- Remove the create-ATA prepend (and its compute-limit bump). The fee ATA is provisioned off the signed path — a backend responsibility.
- `resolveFeeAccount` now verifies the fee ATA exists on-chain and throws otherwise, failing the Jupiter quote so the provider race falls back to LiFi (which also collects the affiliate fee).
- For native-SOL outputs, take the affiliate fee in the **input mint** (the fee owner holds no wSOL ATA, and collecting in wSOL would need unwrapping) — same rule as iOS.

## Test plan

- [x] `JupiterApiTest` — 8 tests green, including new coverage: fee taken in input mint for wSOL outputs; unprovisioned fee ATA fails the quote (provider falls back).
- [x] `:data` and `:app` compile; ktfmt applied.
- [ ] On-device re-run of the failing USDC→SOL Fast-vault swap (emulator was closed; will verify on the next session).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap fee handling so fees are calculated with the correct token on native SOL outputs.
  * Swaps now fail more cleanly when a fee account can’t be resolved or isn’t available.
  * Removed extra transaction handling that could affect swap request construction, making swap payloads more consistent.


johnnyluo (approved):
LGTM

PR #5144, opened 2026-07-02
Join-transaction banner: swipe-up dismiss + stop covering home content

### LINKED ISSUE #5134: Keysign 'Join transaction' banner: not dismissable by swipe-up + blocks home content
The "Join transaction" foreground-notification banner (shown while a keysign/swap session is live) has two UX problems that make it feel like a stuck, blocking popup.

### 1. Swipe-up doesn't dismiss (only horizontal swipe does)
Dismissal is wired to **horizontal** drag only — `detectHorizontalDragGestures` in `ForegroundNotificationBanner.kt:109-137`. There is no vertical gesture, so the natural "swipe up to dismiss a top banner" action does nothing, and users don't discover the sideways flick. The banner reads as non-dismissable.

**Expected:** swiping the banner **up** dismisses it (add a vertical drag gesture, or accept both axes).

### 2. Banner blocks the view underneath
The banner is an overlay drawn on top of `navContent` (`MainActivityContent.kt:153` for content, `:156-179` for the overlay on top). It floats over the home screen without pushing content down, so while visible it obscures the balance card / content beneath it (see screenshot — top of the balance card is covered).

**Expected:** the banner shouldn't cover home content — either push content down while shown, shrink it, or make it clearly transient.

### Repro
1. Start / join a swap keysign session so the "Join transaction" banner appears on the home screen.
2. Try to swipe the banner up → nothing happens.
3. Note the balance card underneath is partially hidden while the banner is up.

Component: `app/src/main/java/com/vultisig/wallet/ui/components/banners/ForegroundNotificationBanner.kt`
Placement: `app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt`

aminsato:
Fixes #5134

The "Join transaction" foreground-notification banner (shown while a keysign/swap session is live) behaved like a stuck, blocking popup. Two fixes:

## Changes

**1. Swipe-up now dismisses the banner** (`ForegroundNotificationBanner.kt`)
- Dismissal was wired to horizontal drag only (`detectHorizontalDragGestures`), so the natural "swipe up to dismiss a top banner" gesture did nothing.
- Switched to `detectDragGestures` and added an upward-flick dismissal: dragging up past ~25% of the banner height animates it off-screen and reports the dismissal, mirroring the existing horizontal logic. The retained off-screen offset keeps it hidden through the parent's vertical exit transition, so the #5001 double-animation fix stays intact.
- Downward travel is clamped to `0` — the banner is top-anchored, so pushing it down into content is meaningless.

**2. Banner no longer covers home content** (`MainActivityContent.kt`)
- The banner floated as an overlay on top of `navContent` and obscured the top of the balance card.
- Its measured height is now reserved in the content `Column` via a mirrored `AnimatedVisibility` spacer, so content is pushed down while the banner is shown and returns up when it's dismissed.
- The overlay itself is unchanged, preserving its status-bar bleed and rounded-corner gradient reveal.

## Test plan
- [ ] Start/join a swap keysign session so the "Join transaction" banner appears on home.
- [ ] Swipe the banner **up** → it dismisses.
- [ ] Swipe the banner **sideways** → still dismisses (unchanged).
- [ ] Short drag in either direction → springs back.
- [ ] While the banner is shown, the balance card sits fully below it (not covered); when dismissed, content returns to its normal position.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
@@ -2,27 +2,33 @@ package com.vultisig.wallet.app.activity.components
 
 import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.ExperimentalAnimationApi
+import androidx.compose.animation.expandVertically
+import androidx.compose.animation.shrinkVertically
 import androidx.compose.animation.slideInVertically
 import androidx.compose.animation.slideOutVertically
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.safeDrawingPadding
 import androidx.compose.foundation.layout.statusBars
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.key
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

Issue #5125, opened 2026-07-01
Solana: align verify-screen priority-fee label with the signed per-CU price


johnnyluo:
Follow-up from #5114 (review thread https://github.com/vultisig/vultisig-android/pull/5114#discussion_r3503649610).

## Background

#5114 fixed the semantic bug where the Solana `priorityFee` field (a per-compute-unit price, microlamports/CU) was being fed the total fee amount. As part of that, `getSpecific` now fetches the dynamic median (`getMedianPriorityFee`) to populate `priorityFee`.

## Problem (display-only — signing is safe)

The **signed/broadcast** transaction is correct: the join device signs the initiator's payload verbatim (`SigningHelper.getKeysignMessages` → `SolanaHelper.getPreSignedImageHash` → `getPreSignedInputData` reads `priorityFee`/`priorityLimit` straight from `keysignPayload.blockChainSpecific`; `getSpecific` is never called on the join path). Every co-signer derives the same sighash from the initiator's per-CU price.

The divergence is limited to the **fee label** shown on the verify screen, because the median is fetched independently at different times:

- **Initiator confirm screen**: display fee flows `GasFeeOrchestrator` → `SolanaFeeService.calculateFees` (median @ T1) → `TokenValue(fees.amount)`; `GasFeeOrchestrator` discards `GasFees.price`. The signed tx then fetches its own median @ T2 in `getSpecific`.
- **Join verify screen**: `JoinKeysignFeeResolver.resolveFallbackFeeAmount` falls through the EVM/THORChain cases and calls `feeServiceComposite.calculateFees(...)` for Solana — a **fresh** median fetch on the joiner's device (@ T3) — instead of reading `priorityFee`/`priorityLimit` from the payload it already holds. `computeJoinKeysignNetworkFee` (`JoinKeysignFeeUtils.kt`) only special-cases `Ethereum`/`THORChain`; Solana returns the recomputed `fallbackFeeAmount`.

Under congestion these snapshots can differ, so the displayed fee may not equal the fee actually charged. The divergence is bounded: both are medians clamped to [1M, 100M] µLam/CU fetched seconds apart, and under non-congestion both hit the 1M floor and are identical.

## Proposed fix

[...]

PR #5156, opened 2026-07-03
fix(solana): stop double-counting priority fee in send fee estimate

johnnyluo:
## Summary
- `SolanaFeeService.calculateFees` was adding the prioritization fee **twice** to the displayed SOL network fee.
- `getFeeForMessage` already includes the prioritization fee, because the serialized message (`getVersionedMessage`) carries the ComputeBudget `SetComputeUnitPrice` / `SetComputeUnitLimit` instructions. The method then added a separate `priorityAmount` on top, so the estimate was `base + 2×priority + rent` instead of `base + priority + rent`.
- Fix: take the RPC message fee as the full base+priority and only add rent-exemption on top (`amount = messageFee + rentExemptionFee`).

## Verification (mainnet RPC, apiVersion 4.1.0)
Probed `getFeeForMessage` with a hand-built transfer message:

| message | `getFeeForMessage` |
|---|---|
| no ComputeBudget instructions | **5,000** (pure base, 1 signature) |
| with ComputeBudget (limit 100k, price 1,000,000 µLam/CU) | **105,000** (5,000 base + 100,000 priority) |

This confirms the RPC folds the priority fee into the message fee, so the old `+ priorityAmount` was a genuine double count.

## Notes
- Signing/broadcast are unaffected — this only changes the displayed estimate.
- Related area: the Solana priority-fee display work in #5114 / #5125.
- `calculateFees` builds its message via TrustWallet Core JNI, which isn't loaded in JVM unit tests (see `SolanaHelperTest`), so it isn't unit-testable here; the mainnet probe covered the behavior instead.
- **Not addressed here:** `calculateDefaultFees` (swap/estimate path) has a similar shape — its `DEFAULT_COIN_TRANSFER_BASE_FEE` constant (105,000 = base + floor-priority) also gets `priorityAmount` added on top. Left untouched pending a decision on the intended default margin.

## Test Plan
- [ ] Manual: send native SOL and an SPL token, confirm the displayed network fee matches the amount actually charged on-chain (explorer)
- [ ] Manual: verify under congestion the fee tracks base + single priority + rent

test tx link: https://orbmarkets.io/tx/3T3NC2xLTKr1BcN6qZ7AFYRg4X3HAEvAbjKknwgWJyVkwEiYTU1oeC2LadheN4QFYXaSJHdeSRUPy4cRokdCKRq9

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Solana transaction fee calculations so non-swap transactions use the full message fee plus rent exemption, avoiding double-counting of priority fees.
  * Improved fee reporting accuracy while keeping priority fee price and limit values unchanged.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/SolanaFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/SolanaFeeService.kt
@@ -71,22 +71,21 @@ class SolanaFeeService @Inject constructor(private val solanaApi: SolanaApi) : F
 
         val serializedTx = SolanaHelper(vaultHexPubKey).getVersionedMessage(keySignPayload)
 
-        val baseFee = solanaApi.getFeeForMessage(serializedTx)
+        // serializedTx carries the ComputeBudget instructions, so getFeeForMessage already returns
+        // base signature fee + prioritization fee (price × limit / 1e6) — verified against mainnet
+        // RPC (a no-ComputeBudget message returns 5000, the same message with them returns 105000).
+        // Adding a separate priority term here would double-count it, so only rent-exemption is
+        // added on top.
+        val messageFee = solanaApi.getFeeForMessage(serializedTx)
         val rentExemptionFee = calculateRentExemptionForTokens(toAddress, coin)
 
         val solanaPriorityFee =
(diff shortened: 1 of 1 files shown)


rkokhatskyi (approved):
Nice catch, LGTM!

PR #5158, opened 2026-07-03
fix(keysign): stop joined device faking a txid when broadcast genuinely fails

### LINKED ISSUE #5146: [Keysign] Joined device shows a txid that never landed when broadcast fails
On a non-initiating (co-signing) device, `broadcastOrRecover` catches every non-`CancellationException` and calls `recoverJoinedDeviceBroadcast`, which returns `signedTx.transactionHash` for **any** error as long as the hash is non-blank — with no error classification and no on-chain check.

`app/src/main/java/com/vultisig/wallet/data/usecases/BroadcastKeysignUseCase.kt:224-241`

The inner `broadcastTx` layer (`BroadcastTxUseCase`) already returns the hash on the legitimate duplicate-broadcast race (peer landed it) **after** verifying on-chain over 3 attempts for BTC/Solana/Polkadot/Ton/Ripple/Tron/Cardano. So the outer catch is reached **only when verification proved the tx is not on chain** — and it then overrides that negative result with a `Broadcasted(local hash)`.

**Result:** the co-signing device shows a success screen + explorer link for a transaction that verifiably never landed (both peers failed, or a genuine rejection both hit). For EVM (no inner verify) a real "insufficient funds" / "intrinsic gas too low" is swallowed into a fake success with a keccak hash.

### Fix sketch
In `recoverJoinedDeviceBroadcast`, return the local hash only when the error is a known duplicate/nonce-collision (or a fresh on-chain `verify` succeeds); otherwise re-throw so the joined device surfaces the failure like the initiator does.

**Effort:** M · **Risk:** MED (narrowing the catch must not reintroduce a false error on the legit duplicate-broadcast race).

johnnyluo:
## Summary
- On a joined (co-signing) device, `recoverJoinedDeviceBroadcast` caught **every** non-cancellation exception and returned the locally computed hash, showing a success screen + explorer link for a transaction that **never landed** — for EVM a real `insufficient funds` / `intrinsic gas too low` was swallowed into a fake success, and for verify-capable chains the outer catch overrode the inner layer's *proven* "not on chain" verdict.
- Narrowed the recovery to the one duplicate-broadcast signal that legitimately reaches this layer unresolved: a **Cosmos account sequence mismatch** (`codespace=sdk`/`code=32`). Every other chain's real duplicate race is already resolved a layer down — EVM in `EvmApi.sendTransaction`, BTC-family/Solana/Polkadot/Ton/Ripple/Tron/Cardano via the inner on-chain verify in `BroadcastTxUseCase`, and the Cosmos mempool-cache code (19) in the Cosmos broadcast parser.
- Everything else now **re-throws**, so a genuine failure surfaces on the joined device exactly as it does on the initiator instead of a false success.

### Changes
- `CosmosBroadcastException`: added typed `isSequenceMismatch` property.
- `BroadcastKeysignUseCase.recoverJoinedDeviceBroadcast`: recover only when joined device **and** `error.isDuplicateBroadcast()` (a Cosmos sequence mismatch) **and** the precomputed hash is non-blank; otherwise re-throw. Used a typed check rather than fragile EVM message-string matching (`EvmApi` already consumes those), which keeps the classifier precise and avoids reintroducing a false error on the legit duplicate-broadcast race.
- Added `BroadcastKeysignUseCaseTest` covering the recover and re-throw paths.

## Issue
Closes #5146

## Test Plan
- [x] New `BroadcastKeysignUseCaseTest` passes (success passthrough, Cosmos sequence-mismatch recovery, non-sequence-mismatch re-throw, generic EVM failure re-throw, initiator never recovers, blank-hash no-recovery, cancellation propagates)
- [x] Existing `BroadcastTxUseCaseTest` still passes
- [ ] Manual: on a 2-of-2 vault, initiate a keysign that genuinely fails to broadcast (e.g. insufficient funds) and confirm the joined device now shows an error instead of a success screen
- [ ] Manual: confirm the legit duplicate-broadcast race (peer lands the tx first) still shows success with the correct hash on the joined device

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved joined-device broadcast recovery so only recognized Cosmos duplicate-broadcast cases are handled locally.
  * Non-Cosmos errors, non-duplicate Cosmos rejections, initiator failures, and cancellations now surface correctly instead of being masked.
  * Prevented recovery when a transaction has no local hash, reducing incorrect fallback behavior.

* **Tests**
  * Expanded coverage for broadcast recovery scenarios across Cosmos and other transaction types.


app/src/main/java/com/vultisig/wallet/data/usecases/BroadcastKeysignUseCase.kt
```diff
@@ -186,15 +187,24 @@ constructor(
...
+     * In a multi-device vault both peers compute the same signed transaction and call this path;
      * whichever device's broadcast reaches the network first wins. The losing device's broadcast is
-     * then rejected — Substrate in particular surfaces this as `Transaction has a bad signature`
-     * (code 1010) when the initiator's extrinsic has already advanced the nonce. The signed bytes
-     * are byte-identical on both devices, so the locally computed
-     * [SignedTransactionResult.transactionHash] is the canonical on-chain hash, and we use it
-     * instead of failing the joined-device screen.
+     * then rejected. Most chains already resolve that race in a lower layer: EVM in
+     * `EvmApi.sendTransaction` (returns the keccak hash on `nonce too low` / `already known`),
+     * BTC-family, Solana, Polkadot, Ton, Ripple, Tron and Cardano in [BroadcastTxUseCase] via an
```
aminsato, line 196: Bittensor is the one chain this list misses: `BittensorApi.broadcastTransaction` only resolves the mempool-phase duplicate (`Already Imported` → null → `orKnownHash`), and unlike Polkadot it has no `recoverIfAlreadyBroadcast` wrapper in `BroadcastTxUseCase`. If the losing device's broadcast arrives after the extrinsic is already in a block, the node rejects it and the joined device now shows an error for a transaction that landed — the old blanket recovery masked this case by accident.

A false error is far better than the fake success this PR fixes, and the window is narrow, so this shouldn't block. Filing a follow-up to wrap Bittensor in `recoverIfAlreadyBroadcast` with an on-chain verify (same shape as Polkadot) would close the gap.

aminsato (approved):
Verified the load-bearing claim that every other chain's legitimate duplicate-broadcast race is resolved below this layer, and it holds:

- **EVM**: `EvmApi.sendTransaction` returns the keccak hash for all duplicate signals (`already known`, the `nonce too low` variants, `transaction already exists`, mempool/cache codes), so any EVM rejection reaching `broadcastOrRecover` is genuine. Using the typed Cosmos check instead of re-matching EVM message strings in the upper layer was the right call.
- **BTC-family, Solana, Polkadot, Ton, Ripple, Tron, Cardano**: all go through `recoverIfAlreadyBroadcast` with a retried on-chain verify, so a rejection reaching the top layer means the verify proved the tx is not on chain — re-throwing is correct, and the updated Polkadot test documents this well.
- **Cosmos**: mempool-phase duplicate (code 19) is success in `parseCosmosBroadcastResponse`; the committed-phase sequence mismatch (`sdk`/32) is indeed the only race signal that reaches this catch, and `isSequenceMismatch` matches the constants used by `from()`.
- Initiator/blank-hash/cancellation guards are all covered by tests; the suite passes locally, and the re-thrown `CosmosBroadcastException` markers map to the existing localized copy in `KeysignErrorScreen`.

One inline comment on a Bittensor edge worth a follow-up issue — not blocking.

The PR description says "Added `BroadcastKeysignUseCaseTest`" but the tests live in the existing `KeysignViewModelBroadcastRecoveryTest.kt` — update the description so the file is findable.

The two unchecked manual test-plan items (real 2-of-2 broadcast failure, and the legit duplicate race) are the right things to confirm before merge, since the sequence-mismatch recovery can't be exercised against a real node in unit tests.

PR #5160, opened 2026-07-04
fix(swap): rKUJI destination fails with generic 'adjust the amount' error

### LINKED ISSUE #5113: Swap: rKUJI destination always fails with generic 'adjust the amount' error (THORChain 'bad to asset')
## Summary

Requesting a THORChain swap quote **into rKUJI** (e.g. ATOM → rKUJI) always fails. The screen shows a generic, misleading error — *"Swap quote failed. Please try again or adjust the amount."* — but the request can never succeed: THORChain rejects the destination asset as invalid, so no retry or amount change will help.

## Steps to reproduce

1. Open **Swap** (Market tab).
2. From: **Cosmos / ATOM (Native)**.
3. To: **Cosmos / rKUJI**.
4. Enter any amount (e.g. `0.564445` ATOM).

## Expected

Either a valid quote is returned, or — if THORChain genuinely cannot route into rKUJI — the pair should not be offered / the error should clearly say the route is unsupported, rather than telling the user to "adjust the amount".

## Actual

A red **Error** toast appears: *"Swap quote failed. Please try again or adjust the amount."*

Under the hood the THORChain quote endpoint returns **HTTP 400** for both the rapid and streaming-fallback attempts:

```
RESPONSE: 400 Bad Request
{"code":3, "message":"bad to asset: invalid symbol: invalid request", "details":[]}
```

The destination asset sent to THORChain is:

```
to_asset=GAIA.rKUJI-ibc/50A69DC508ACCADE2DAC4B8B09AA6D9C9062FCBFA72BB4C6334367DECD972B06
```

i.e. rKUJI is being addressed as a GAIA (Cosmos) IBC token. THORChain does not recognize this symbol, so it rejects the request. Because the failure is deterministic (not a transient/amount problem), the "try again or adjust the amount" guidance is misleading.

App-side this surfaces as:

```
SwapException$UnkownSwapError: bad to asset: invalid symbol: invalid request
```

## Impact

- rKUJI is selectable as a swap destination but every quote attempt fails.
- The error message points the user toward an action (changing the amount) that cannot resolve the problem.

## Notes

- Two sub-problems here: (1) the rKUJI swap pair appears valid in the UI but is not routable as constructed, and (2) the error messaging for non-retryable / unsupported-route quote failures is generic and misleading.
- Observed on `aminsato/solana_broadcast_skip_preflight` (current dev build), Android.
- Wallet/destination address redacted from logs.

aminsato:
## Summary

Fixes #5113 — requesting a THORChain quote into a Cosmos Hub IBC token (e.g. ATOM → rKUJI) always failed with a deterministic Thornode 400 (`bad to asset: invalid symbol: invalid request`), surfaced as the misleading *"Swap quote failed. Please try again or adjust the amount."*

Both sub-problems from the issue are addressed:

- **Don't offer the unroutable pair**: the provider table offered THORCHAIN for every `GaiaChain` coin, but THORChain's only Cosmos Hub pool is native ATOM. The GaiaChain branch is now gated on THORChain eligibility (static `ATOM` fallback + live `Available` pools, same pattern as the BSC/Avalanche/Base branches). Selecting rKUJI as a destination now resolves to an empty provider set and shows **"Swap route not available"** immediately — no doomed network call.
- **Honest error for unsupported routes**: Thornode's `bad to asset` / `bad from asset` / `invalid symbol` messages now map to `SwapRouteNotAvailable` instead of falling through to `UnkownSwapError`, so any path that still reaches the node reports the route as unsupported instead of suggesting an amount change.

If THORChain ever lists a Cosmos IBC pool, the live pool eligibility cache re-enables the route automatically — no static table change needed.

## Test plan

- New unit tests pin the behavior:
  - `SwapProviderTableTest`: native ATOM keeps THORCHAIN; a GaiaChain IBC token gets no providers; a live `GAIA.RKUJI` Available pool re-enables the route.
  - `HandleSwapExceptionTest`: the exact Thornode message (`bad to asset: invalid symbol: invalid request`), the `bad from asset` variant, and bare `invalid symbol` all map to `SwapRouteNotAvailable`.
- `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest` — all green.
- Manual: Swap ATOM → rKUJI now shows "Swap route not available" instead of the retry/amount toast; ATOM → BTC/ETH quoting unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated swap error handling to recognize additional “invalid asset symbol” conditions from swap failures and surface them as route-unavailable.
  * Improved Gaia/THORChain swap routing by introducing a native fallback token list and requiring eligibility before enabling THORChain for specific Gaia assets.
* **Tests**
  * Added unit tests for “invalid symbol” swap error mapping.
  * Added Gaia/THORChain routing tests to verify provider availability for native ATOM vs. specific IBC tokens based on eligibility.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapException.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/errors/SwapException.kt
@@ -62,6 +62,11 @@ sealed class SwapException(message: String) : Exception(message) {
                     contains("amount less than dust threshold") -> AmountBelowDustThreshold(error)
                     contains("amount less than min swap amount") -> SmallSwapAmount(error)
                     contains("pool does not exist") -> SwapRouteNotAvailable(error)
+                    // Thornode rejects an asset id it can't parse ("bad to asset: invalid
+                    // symbol: invalid request") — deterministic, so never "adjust the amount".
+                    contains("bad to asset") ||
+                        contains("bad from asset") ||
+                        contains("invalid symbol") -> SwapRouteNotAvailable(error)
                     contains("trading is halted") || contains("trading halted") ->
                         TradingHalted(error)
                     contains("timeout") -> TimeOut(error)
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapProviderTable.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapProviderTable.kt
@@ -39,6 +39,7 @@ constructor(private val poolEligibility: SwapPoolEligibilityRepository) : SwapPr
             "YFI",
         )
     private val thorBscTokens = setOf("BNB", "USDT", "USDC")
+    private val thorGaiaTokens = setOf("ATOM")
     private val thorAvaxTokens = setOf("AVAX", "USDC", "USDT", "SOL")
     private val thorBaseTokens = setOf("ETH", "CBBTC", "USDC", "VVV")
     private val mayaEthTokens = setOf("ETH", "USDC", "LLD")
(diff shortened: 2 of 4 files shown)


aminsato:
@coderabbitai full review

johnnyluo (approved):
LGTM

PR #5165, opened 2026-07-04
fix(zcash): re-plan memo txs to the ZIP-317 conventional fee

### LINKED ISSUE #5155: [BUG] ZEC memo txs (Maya-route swaps) plan an underpaying ZIP-317 fee — extension co-sign fails, mobile-only broadcast rejected
## Summary

ZEC transactions **with a memo** (i.e. MayaChain-routed ZEC swaps — the swap instruction rides in an OP_RETURN) are planned with an underpaying ZIP-317 fee on Android. Two user-visible failures:

1. **Extension-initiated swap, Android co-signs → keysign fails every time.** The extension's SDK signer plans the correct (higher) conventional fee; Android plans WalletCore's underpaying `zip0317` fee. Different fee → different change output → different preimage hashes → MPC keysign cannot converge. The user then sees a trailing relay error (reported as "HTTP 404") after the failed sign — that's the torn-down session, not the cause.
2. **Mobile-only signing → tx signs but the network rejects it at broadcast** (fee below the ZIP-317 conventional minimum).

Live user report: ZEC → USDC initiated on extension, co-signed with Android, signing failed, then HTTP 404 on the Android side.

## Root cause

WalletCore's `zip_0317` planner sizes an OP_RETURN output at a flat ~34 bytes — one logical action short of what ZIP-317 charges — and ignores `byteFee` in that mode (verified against WalletCore 4.6.13 in vultisig/vultisig-windows#4145 / vultisig/vultisig-windows#4153). So any memo tx plans exactly one action short (e.g. 15000 zats planned vs 20000 required). WalletCore 4.7.0 does **not** fix this (no Zcash changes in 4.6.15→4.7.0).

The TS SDK fixed this in vultisig/vultisig-sdk#773: when the `zip_0317` plan underpays the byte-accurate conventional fee, re-plan with `zip_0317` **off** (where `byteFee` is honoured) and bump `byteFee` until the fee clears (`packages/core/mpc/keysign/signingInputs/resolvers/utxo.ts`, `planZcashConventionalFee`).

**Android never got the equivalent fix.** `data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelper.kt` sets `setZip0317(coinType == CoinType.ZCASH)` on both the swap path (memo required, `setOutputOpReturn`) and the send path, with no conventional-fee check or re-plan anywhere in the file.

Note: SwapKit ZEC swaps are unaffected — `SwapKitZcashSigner` uses a frozen plan from the SwapKit PSBT, and NEAR-route swaps are plain transfers with no memo. Only the memo path (Maya-native swaps, sends with memo) is broken.

## Fix

Port sdk#773's `planZcashConventionalFee` algorithm into `UtxoHelper`:
1. Plan with `zip0317 = true` as today.
2. Compute the byte-accurate ZIP-317 conventional fee for the planned tx (recipient + change + OP_RETURN with PUSHDATA/CompactSize thresholds — see `packages/core/chain/chains/utxo/fee/zip317.ts` in the SDK for the reference math).
3. If the plan's fee is below it, re-plan with `zip0317 = false` and bump `byteFee` until the fee meets the conventional fee (SDK caps at 5 bumps).
4. Plain (no-memo) sends keep the `zip0317` plan unchanged.

⚠️ The re-planned fee must be **byte-identical to the SDK's result** — MPC co-signing requires all devices to derive the same digests. Divergence here just moves the keysign failure. This is exactly the class of bug the cross-platform signing-vector net (vultisig/commondata#94) should pin; a ZEC-memo vector should be added there as part of this fix.

Long-term, the clean fix is upstreaming the OP_RETURN action-count correction to trustwallet/wallet-core so all platforms converge without workarounds.

## Verification (all reproduced live on 2026-07-03)

- SwapKit rail for ZEC→USDC is healthy on all three proxies (quote + swap 200s) — separate from this bug.
- Maya quote for ZEC→ETH.USDC returns a valid quote, so Maya-route swaps are offered and users do hit this path.
- Extension prod (0.2.22) + desktop v1.0.64+ carry sdk#773; Android (v1.0.112) and iOS carry no equivalent.

Sibling issue filed on vultisig-ios (same gap in `UTXOChainsHelper.swift`).

aminsato:
Fixes #5155

## Problem

WalletCore's `zip_0317` planner sizes an OP_RETURN output at a flat ~34 bytes and ignores `byteFee`, so ZEC **memo** transactions plan exactly one logical action short (e.g. 15,000 zats planned where ZIP-317 requires 20,000). This breaks two flows:

1. **Extension-initiated ZEC swap, Android co-signs → keysign fails every time.** The extension's SDK signer plans the higher conventional fee; Android planned WalletCore's underpaying `zip_0317` fee. Different fee → different change output → different preimage digests → MPC keysign can't converge.
2. **Mobile-only signing → tx signs but the network rejects it at broadcast** (fee below the ZIP-317 conventional minimum).

Maya-native ZEC swaps ride the plain send path with the swap instruction carried in the OP_RETURN memo, so they hit exactly this path. SwapKit ZEC swaps are unaffected (frozen PSBT plan); NEAR-route swaps are plain transfers with no memo.

## Fix

Ports the SDK's `planZcashConventionalFee` (and the merged iOS sibling #4728) into `UtxoHelper`:

1. Plan with `zip_0317 = true` as today.
2. Compute the byte-accurate ZIP-317 conventional fee for the planned tx (recipient + change + OP_RETURN, with PUSHDATA/CompactSize thresholds) via the new `ZcashConventionalFee`.
3. If the plan's fee is below it, re-plan with `zip_0317 = false` (where WalletCore honours `byteFee`) and bump `byteFee` until the fee clears (capped at 5 bumps).
4. Plain (no-memo) sends already meet the fee and keep the `zip_0317` plan unchanged.

All three planning call sites (`getSigningInputData`, `getBitcoinPreSigningInputData`, `getBitcoinTransactionPlan`) route through the new `planTransaction` guard so the fee preview and the signed tx agree.

## Cross-platform parity

The re-plan is a **byte-parity** port — every co-signing device must derive the same plan or the digests diverge and the keysign failure just moves. `ZcashConventionalFee` mirrors the SDK's `zip317.ts` line-for-line, and the plan-level golden vectors mirror iOS against the same **wallet-core 4.7.0** pin this module already uses.

## Tests

- `ZcashConventionalFeeTest` — pure-math golden vectors mirrored from the SDK's `zip317.test.ts` / iOS `ZcashConventionalFeeTests` (12 tests, run on the JVM host, all green).
- `ZcashZip317PlanTest` — real-WalletCore regression: plain send stays at the floor, memo/send-max/long-memo/Maya-swap-shaped/non-ASCII memos re-plan to the exact iOS golden fee & change, and the insufficient-funds plan passes through untouched. Skipped gracefully when the WalletCore native lib is absent in the JVM host (runs on CI/device).

## Follow-up

A ZEC-memo signing vector should be added to the cross-platform signing-vector net (vultisig/commondata#94) to pin this across platforms — tracked separately since `commondata/` is a submodule.

## Test plan

- [x] `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.chains.helpers.*"` green
- [x] `./gradlew :data:ktfmtCheck` clean
- [ ] On device: ZEC → USDC Maya swap, extension-initiated + Android co-sign, converges and broadcasts

## Summary by CodeRabbit

* **Bug Fixes**
  * Zcash transactions now better match the expected network fee rules, including memo/OP_RETURN cases and larger outputs.
  * Transaction planning is more reliable for Bitcoin-family sends, with improved fee and change calculations in edge cases.
* **Tests**
  * Added regression coverage for Zcash fee calculation and planning across common, memo-heavy, and insufficient-funds scenarios.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelper.kt
@@ -35,6 +35,12 @@ class UtxoHelper(
     val vaultHexChainCode: String,
 ) {
     companion object {
+        /**
+         * Max non-ZIP-317 re-plans before giving up raising the Zcash fee. Matches the SDK's cap so
+         * every co-signing device fails identically instead of diverging on the plan.
+         */
+        private const val MAX_ZCASH_FEE_BUMPS = 5
+
         fun getHelper(vault: Vault, coinType: CoinType): UtxoHelper {
             when (coinType) {
                 CoinType.BITCOIN,
@@ -53,6 +59,89 @@ class UtxoHelper(
                 else -> throw Exception("Unsupported chain")
             }
         }
+    }
+
+    /**
+     * Plan the transaction, guarding Zcash plans against WalletCore's underpaying ZIP-317 mode.
+     * Every other coin plans directly. For Zcash the [signingInput] builder is mutated (zip0317
+     * toggled off, byteFee bumped) so the serialized input the digests commit to matches the plan.
+     */
+    private fun planTransaction(
+        signingInput: Bitcoin.SigningInput.Builder
+    ): Bitcoin.TransactionPlan =
+        if (coinType == CoinType.ZCASH) {
+            planZcashConventionalFee(signingInput)
+        } else {
+            AnySigner.plan(signingInput.build(), coinType, Bitcoin.TransactionPlan.parser())
+        }
+
+    /**
+     * Produce a Zcash plan whose fee meets the ZIP-317 conventional fee.
+     *
+     * WalletCore's `zip_0317` planner sizes an OP_RETURN output as a flat ~34 bytes and ignores
+     * `byteFee`, so memo transactions (MayaChain-routed swaps carry the swap instruction in an
+     * OP_RETURN; sends with memo too) plan one logical action short — e.g. 15,000 zats where the
+     * network requires 20,000 — with no way to raise the fee in that mode. When the `zip_0317` plan
+     * underpays the byte-accurate conventional fee, re-plan with `zip_0317` off — where WalletCore
+     * honours `byteFee` — and bump `byteFee` until the fee clears. Plain (no-memo) sends already
+     * meet the fee and keep the `zip_0317` plan.
+     *
+     * An empty plan (no selected UTXOs) is returned untouched: coin selection produced nothing
+     * (insufficient funds), and that flow owns the outcome.
+     *
+     * Byte-parity port of the SDK's `planZcashConventionalFee` (and iOS `UTXOChainsHelper`) — every
+     * co-signing device must derive the same plan or the MPC preimage digests diverge and keysign
+     * fails. Keep in lockstep with the SDK.
+     */
+    private fun planZcashConventionalFee(
+        signingInput: Bitcoin.SigningInput.Builder
+    ): Bitcoin.TransactionPlan {
+        val memoSize = signingInput.outputOpReturn.size()
+
+        fun conventionalFee(plan: Bitcoin.TransactionPlan): Long =
+            ZcashConventionalFee.conventionalFee(
+                inputCount = plan.utxosCount,
+                outputSizes =
+                    ZcashConventionalFee.transparentOutputSizes(
+                        change = plan.change,
+                        memoSize = memoSize,
+                    ),
+            )
+
+        // An empty plan has no shape to charge for; leave the fee flow to the caller.
+        fun meetsConventionalFee(plan: Bitcoin.TransactionPlan): Boolean =
+            plan.utxosCount == 0 || plan.fee >= conventionalFee(plan)
+
+        val zipPlan =
+            AnySigner.plan(signingInput.build(), coinType, Bitcoin.TransactionPlan.parser())
+        if (meetsConventionalFee(zipPlan)) {
+            return zipPlan
+        }
+
+        signingInput.setZip0317(false)
+        var byteFee = 1L
+        var plan = zipPlan
+        repeat(MAX_ZCASH_FEE_BUMPS) {
+            signingInput.setByteFee(byteFee)
+            plan = AnySigner.plan(signingInput.build(), coinType, Bitcoin.TransactionPlan.parser())
+            if (meetsConventionalFee(plan)) {
+                return plan
+            }
+
+            // byteFee mode scales fee linearly with vsize; derive the byteFee that clears the
+            // conventional fee for this plan's (byteFee-independent) vsize.
+            val plannerVsize = if (plan.fee > 0L) plan.fee / byteFee else 1L
+            byteFee =
+                maxOf(
+                    ZcashConventionalFee.ceilDiv(conventionalFee(plan), plannerVsize),
+                    byteFee + 1,
+                )
+        }
+
+        error(
+            "Failed to meet the Zcash minimum network fee (ZIP-317): planned ${plan.fee} zats, " +
+                "required ${conventionalFee(plan)}"
+        )
     }
 
     fun getPreSignedImageHash(keysignPayload: KeysignPayload): List<String> {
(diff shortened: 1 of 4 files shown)


johnnyluo:
@coderabbitai review please

johnnyluo:
Please provide a on chain tx link

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/UtxoHelper.kt
```diff
@@ -55,6 +61,89 @@ class UtxoHelper(
...
+     * meet the fee and keep the `zip_0317` plan.
+     *
+     * An empty plan (no selected UTXOs) is returned untouched: coin selection produced nothing
+     * (insufficient funds), and that flow owns the outcome.
+     *
+     * Byte-parity port of the SDK's `planZcashConventionalFee` (and iOS `UTXOChainsHelper`) — every
+     * co-signing device must derive the same plan or the MPC preimage digests diverge and keysign
+     * fails. Keep in lockstep with the SDK.
+     */
+    private fun planZcashConventionalFee(
```
johnnyluo, line 96: **Downstream: the ZEC swap fee estimate under-reports this new fee.** (anchored here since the caller isn't in the diff)

The re-plan introduced here raises a Maya-native ZEC swap's fee to the ZIP-317 conventional fee (~20,020+ zats). But `SwapGasCalculator.getBitcoinTransactionPlan(...)` estimates the swap network fee with `memo = null` (`SwapGasCalculator.kt:145`), which keeps the cheap `zip_0317` plan (~10,000 zats) and never triggers this re-plan. So the network fee shown to the user during a ZEC swap is roughly half of what the signed tx will actually cost, and for a send-max/full-balance ZEC swap the under-reservation could leave too little for the true fee and fail the real plan.

The `memo = null` predates this PR, but this PR widens the estimate-vs-actual gap in exactly the flow it targets. Worth confirming the swap fee display/reservation stays correct end-to-end, or passing the real swap memo through in `SwapGasCalculator` so the estimate re-plans the same way keysign does.

johnnyluo: Confirmed against the code: `SwapGasCalculator.getBitcoinTransactionPlan(...)` is called with `memo = null` at `SwapGasCalculator.kt:145`. With no memo, `signingInput.outputOpReturn` is empty, so the new `planZcashConventionalFee` sees `memoSize = 0`, the `zip_0317` plan already clears the conventional fee, and it never re-plans — the estimate stays on the cheap plan (~10-15k zats) while the signed Maya-native ZEC swap re-plans to ~30k zats (per this PR's own `ZcashZip317PlanTest` "Maya-swap-shaped memo" vector, 30,160 zats).

So the swap fee shown to the user is roughly half the real cost, and a send-max/full-balance ZEC swap could under-reserve and fail the real plan. The `memo = null` predates this PR, but this PR widens the estimate-vs-actual gap in exactly the flow it targets. Suggest threading the real swap memo through `SwapGasCalculator` so the estimate re-plans the same way keysign does (or otherwise confirming the fee display/reservation stays correct end-to-end). Leaving this thread open to track.

johnnyluo: Traced the full display path — this is **not** a live bug, resolving.

The displayed UTXO swap fee doesn't come from the `memo = null` call in `calculateGasFee`. That runs before any quote exists (no memo available yet), and for UTXO chains its `estimated` is explicitly discarded (`SwapQuotePipelineController.kt:198`). The actual displayed fee and balance/send-max check come from `resolveUtxoPlanFee` (`SwapQuotePipeline.kt:473`), which passes `memo = result.utxoMemo` = `quote.data.memo` (the real Maya swap memo) → `getBitcoinSigningInput` sets `outputOpReturn` → `planZcashConventionalFee` re-plans to the ZIP-317 conventional fee (~30k zats). So the display already matches keysign. No change needed.

johnnyluo (approved):
LGTM

PR #5169, opened 2026-07-04
feat(cardano): attach send memo on-chain as CIP-20 metadata (WalletCore 4.7.0)

### LINKED ISSUE #5124: Cardano: attach CIP-20 memo via WalletCore native auxiliary_data (wallet-core 4.7.0 released)
## Summary

`@trustwallet/wallet-core` **4.7.0** is released and adds native Cardano CIP-20 memo support — the `Cardano.SigningInput.auxiliary_data` proto field ([trustwallet/wallet-core#4784](https://github.com/trustwallet/wallet-core/pull/4784), fixes [#4783](https://github.com/trustwallet/wallet-core/issues/4783), shipped in [4.7.0](https://github.com/trustwallet/wallet-core/releases/tag/4.7.0)).

This unblocks the Android TODO in [`CardanoHelper.kt`](data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt#L54):

```kotlin
// TODO: Implement memo support when WalletCore adds Cardano metadata support
```

We've already implemented this on Extension/Desktop + the SDK (see references below) and verified it on mainnet. This issue proposes the equivalent Android change.

## How it works

A Cardano memo is CIP-20 transaction metadata under label **674**:

```
{ 674: { "msg": ["<chunk1>", "<chunk2>", ...] } }
```

Each `msg` text chunk must be ≤ **64 UTF-8 bytes** (split on codepoint boundaries, never mid-codepoint).

With 4.7.0 the flow is fully native — no client-side CBOR body patching:

1. CBOR-encode the CIP-20 auxiliary data.
2. Set it on `Cardano.SigningInput.auxiliary_data`.
3. WalletCore commits `blake2b-256(auxData)` into tx body key 7 (`auxiliary_data_hash`) and embeds the aux bytes as element `[3]` of the signed-tx array (replacing the `null` / `0xf6` sentinel).

## Proposed changes (Android)

- [ ] **Bump wallet-core** `4.6.13` → `4.7.0` in [`gradle/libs.versions.toml`](gradle/libs.versions.toml#L44) (`walletCore` version ref). Confirm the 4.7.0 Android artifact is published to the resolved repo.
- [ ] **Set auxiliary data** in [`CardanoHelper.kt`](data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt) — when `keysignPayload.memo` is present, build the CIP-20 CBOR (label 674, 64-byte chunking) and set it on `Cardano.SigningInput` (`auxiliary_data`). The existing CBOR helpers in [`CardanoUtils.kt`](data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoUtils.kt) can build the metadata. Remove the TODO at line 54.
- [ ] **Fee**: include the auxiliary-data byte size in fee estimation so the memo bytes are paid for (the current `AnySigner.plan` + `forceFee` flow underprices a memo send otherwise).
- [ ] **Enable the memo UI** for Cardano — the memo field is gated in [`SendFormGraph.kt`](app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormGraph.kt#L456-L458):
  ```kotlin
  val hasMemo =
      (token.isNativeToken || token.chain.standard == TokenStandard.COSMOS) &&
          token.chain != Chain.Sui
  ```
  Cardano (a UTXO chain) evaluates to `false` today; include native Cardano so the input in [`SendFormAmountSection.kt`](app/src/main/java/com/vultisig/wallet/ui/screens/send/SendFormAmountSection.kt#L245) is shown.
- [ ] The `memo` already exists on [`KeysignPayload`](data/src/main/kotlin/com/vultisig/wallet/data/models/payload/KeysignPayload.kt#L23); confirm it's threaded to `CardanoHelper`.

## Cross-device signing note

Cardano signing is byte-parity-sensitive across co-signing devices. Since WalletCore now derives the `auxiliary_data_hash` and envelope natively, iOS/Android/Extension must all produce the **same** CIP-20 CBOR bytes (identical label, chunking, and canonical map encoding) so co-signers agree on the signing hash. Keep the metadata encoding identical across platforms.

## Reference implementation

- SDK signing: [vultisig/vultisig-sdk#722](https://github.com/vultisig/vultisig-sdk/pull/722) — sets `SigningInput.auxiliaryData`, drops the old client-side patching.
- Extension/Desktop: [vultisig/vultisig-windows#4074](https://github.com/vultisig/vultisig-windows/pull/4074) — wallet-core 4.7.0 bump + memo UI re-enable.
- **Verified on mainnet** (Metadata tab → label `674`): https://cardanoscan.io/transaction/f8a1c9e741334db41b954e3cbee16605aae36eb80259bbea3d791a8f8eb60e21

This is a cross-platform gap — the same TODO exists on iOS ([#4326](https://github.com/vultisig/vultisig-ios/issues/4326)).

johnnyluo:
## ⚠️ HIGH-tier Cardano signing path — needs on-device + cross-device review before merge

Attaches the Cardano send memo on-chain as **CIP-20 transaction metadata (label 674)** using WalletCore 4.7.0's native `Cardano.SigningInput.auxiliary_data`. Cardano signing is MPC/TSS, so this is **byte-parity-sensitive**: it needs on-device and cross-device (iOS ↔ Android ↔ Extension) co-signing byte-parity validation on mainnet before it leaves draft. Kept as a **draft** for that reason.

Closes #5124. Cross-platform reference: [vultisig-ios#4709](https://github.com/vultisig/vultisig-ios/pull/4709) (merged, mainnet-verified).

## Problem

The Cardano send memo was typed by the user but **silently dropped at signing time** — the old `CardanoHelper` carried a `// TODO: Implement memo support when WalletCore adds Cardano metadata support`. wallet-core `4.7.0` (already on `main` via #5154) exposes the native `Cardano.SigningInput.auxiliary_data` field, unblocking this.

## What changed

- **`CardanoCIP20`** (new) — canonical CIP-20 encoder: CBOR for `{ 674: { "msg": [chunks] } }`, each `msg` chunk ≤ **64 UTF-8 bytes** split on codepoint boundaries (never mid-codepoint), plus the `blake2b-256` aux hash. The encoding is **byte-parity pinned** to the iOS/SDK golden vector so MPC co-signers agree on the sighash.
- **`CardanoHelper`** — builds the aux CBOR from `keysignPayload.memo` and sets it on `Cardano.SigningInput.auxiliary_data`. WalletCore commits `blake2b-256(auxData)` into tx body **key 7** (`auxiliary_data_hash`) and embeds the aux bytes as signed-tx element **[3]** natively. Android compiles the signed tx via `compileWithSignatures`, so **no manual envelope patching is needed** (unlike iOS, which hand-builds the envelope because `compileWithSignatures` crashes there).
- **Fee** — the memo is threaded into `estimateFee` so the size-based dynamic fee **prices the aux bytes**. Without it a memo send underprices and the node rejects the broadcast with `FeeTooSmallUTxO`.

## No UI gating change (Android differs from iOS)

The Cardano memo input already renders on Android: the Send memo gate is `token.isNativeToken || COSMOS`, and native ADA is `isNativeToken = true`. iOS needed a `supportsMemo` change only because it used a separate gate that explicitly excluded Cardano.

## Byte parity (golden vector)

`{ 674: { "msg": ["hello world"] } }` encodes to `a11902a2a1636d7367816b68656c6c6f20776f726c64` — identical to the iOS `CardanoCIP20Tests` and SDK `buildCip20AuxData` fixtures.

## Test plan

- [x] `CardanoCIP20Test` — 13 tests: chunking (ASCII 64/65-byte boundaries, 2/3/4-byte codepoints never torn, empty memo → single empty chunk), pinned CBOR golden vectors, blake2b-256 aux hash. All green.
- [x] Existing `*Cardano*` suites still green (swap path passes `memo = null` → no aux, unchanged).
- [x] `ktfmtCheck` clean.
- [ ] **On-device mainnet send with a memo** → confirm CIP-20 label 674 metadata on cardanoscan.
- [ ] **Cross-device co-signing** (Android ↔ iOS ↔ Extension) → identical body/sighash, broadcast succeeds.

on chain tx link: 
https://cardanoscan.io/transaction/b555ee5c488bf18aa6471ecad6bb0756b7e0614f0a8a66f86bd068cb403c21eb

## Summary of changes

* **New Features**
  * Optional Cardano memos are now encoded as CIP-20 auxiliary data and included in Cardano signing inputs.

* **Bug Fixes**
  * Memo content now consistently affects Cardano fee estimation and byte-fee reporting across send and join-fee flows.
  * Cardano-specific recalculations and submission now reliably use the memo value confirmed during validation.

* **Tests**
  * Added deterministic CIP-20 encoding/hash tests, including UTF-8 boundary-safe chunking.
  * Added coverage to ensure memo presence changes the Cardano sighash as expected.

### DIFF
--- a/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
@@ -560,6 +560,33 @@ class ChainHelpersTest {
         }
     }
 
+    @Test
+    fun cardanoMemoChangesSighash() {
+        val transactions: List<TransactionData> = loadTransactionData(CARDANO_JSON_FILE)
+        transactions.forEach { transaction ->
+            val internalPayload = transaction.keysignPayload.toInternalKeySignPayload()
+            val cardano = internalPayload.blockChainSpecific as BlockChainSpecific.Cardano
+
+            // WalletCore ignores forceFee for max-amount sends (output = total - fee), so the
+            // memo-driven aux hash can't be isolated from the fee there; assert only on ordinary
+            // sends, matching cardanoSignerForcesTransmittedByteFee.
+            if (cardano.sendMaxAmount) return@forEach
+
+            // Attaching a CIP-20 memo must flow through setAuxiliaryData into the pre-image hash:
+            // WalletCore commits blake2b-256(auxData) into body key 7, so the sighash MUST differ
+            // from the memo-less payload. Mirrors iOS's CardanoCip20SigningTests.
+            val noMemoHash = CardanoHelper.getPreSignedImageHash(internalPayload.copy(memo = null))
+            val memoHash =
+                CardanoHelper.getPreSignedImageHash(internalPayload.copy(memo = "hello world"))
+
+            assertTrue(
+                "Expected 64-char hex hash but got: ${memoHash[0]}",
+                memoHash[0].matches(Regex("[0-9a-f]{64}")),
+            )
+            assertNotEquals(noMemoHash, memoHash)
+        }
+    }
+
     private fun loadTransactionData(jsonFile: String): List<TransactionData> {
         val appContext: Context = InstrumentationRegistry.getInstrumentation().context
         val data =
--- a/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
@@ -550,6 +550,7 @@ class ChainHelpersTest {
                     sendMaxAmount = cardano.sendMaxAmount,
                     ttl = cardano.ttl.toLong(),
                     utxos = internalPayload.utxos,
+                    memo = internalPayload.memo,
                 )
 
             assertTrue(
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
@@ -1,6 +1,7 @@
 package com.vultisig.wallet.data.chains.helpers
 
 import com.google.protobuf.ByteString
+import com.vultisig.wallet.data.crypto.CardanoCIP20
 import com.vultisig.wallet.data.crypto.CardanoUtils
 import com.vultisig.wallet.data.crypto.checkError
 import com.vultisig.wallet.data.models.Chain
@@ -40,6 +41,7 @@ object CardanoHelper {
         ttl: Long,
         utxos: List<UtxoInfo>,
         forceFee: Long,
+        memo: String?,
     ): Cardano.SigningInput.Builder {
         val input =
             Cardano.SigningInput.newBuilder()
@@ -51,8 +53,13 @@ object CardanoHelper {
                         .setChangeAddress(changeAddress)
                         .setForceFee(forceFee)
                 )
-                // TODO: Implement memo support when WalletCore adds Cardano metadata support
                 .setTtl(ttl)
+
+        // CIP-20 memo (label 674). When present, WalletCore commits
+        // blake2b-256(auxDataCbor) into the body at map key 7 and emits the aux
+        // CBOR as element [3] of the signed tx. The encoder is byte-parity pinned
+        // to the SDK golden vector so co-signers agree on the sighash.
+        cip20AuxData(memo)?.let { input.setAuxiliaryData(ByteString.copyFrom(it)) }
 
         // Add UTXOs to the input
         for (inputUtxo in utxos) {
@@ -95,7 +102,19 @@ object CardanoHelper {
             ttl = ttl.toLong(),
             utxos = keysignPayload.utxos,
             forceFee = byteFee,
+            memo = keysignPayload.memo,
         )
+    }
+
+    /**
+     * Canonical CIP-20 auxiliary-data CBOR (label 674) for the payload [memo], or `null` when there
+     * is no memo. Both the pre-sign input (`Cardano.SigningInput.auxiliary_data`) and the
+     * WalletCore-compiled signed envelope derive from these bytes, so the body's key-7 hash and the
+     * embedded aux stay consistent — and byte-identical to the iOS/Extension co-signers.
+     */
+    private fun cip20AuxData(memo: String?): ByteArray? {
+        if (memo.isNullOrEmpty()) return null
+        return CardanoCIP20.buildAuxData(memo).auxDataCbor
     }
 
     /**
@@ -133,6 +152,7 @@ object CardanoHelper {
         sendMaxAmount: Boolean,
         ttl: Long,
         utxos: List<UtxoInfo>,
+        memo: String?,
     ): Long {
         val signingInput =
             buildSigningInput(
@@ -143,6 +163,7 @@ object CardanoHelper {
                     ttl = ttl,
                     utxos = utxos,
                     forceFee = 0,
+                    memo = memo,
                 )
                 .build()
         return plan(signingInput).fee
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
@@ -56,9 +56,10 @@ object CardanoHelper {
                 .setTtl(ttl)
 
         // CIP-20 memo (label 674). When present, WalletCore commits
-        // blake2b-256(auxDataCbor) into the body at map key 7 and emits the aux
-        // CBOR as element [3] of the signed tx. The encoder is byte-parity pinned
-        // to the SDK golden vector so co-signers agree on the sighash.
+        // blake2b-256(auxDataCbor) into the body at map key 7 and embeds the aux
+        // CBOR as the transaction's auxiliary_data element. The encoder is
+        // byte-parity pinned to the SDK golden vector so co-signers agree on the
+        // sighash.
         cip20AuxData(memo)?.let { input.setAuxiliaryData(ByteString.copyFrom(it)) }
 
         // Add UTXOs to the input
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoCIP20.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoCIP20.kt
@@ -13,8 +13,8 @@ import kotlin.experimental.and
  * Each `msg` text chunk is at most 64 UTF-8 bytes, split on Unicode codepoint boundaries (never
  * mid-codepoint). WalletCore 4.7.0 consumes these CBOR bytes via
  * `Cardano.SigningInput.auxiliary_data`: it commits `blake2b-256(auxDataCbor)` into the tx body at
- * map key 7 (auxiliary_data_hash) and embeds the bytes as element [3] of the signed transaction
- * array.
+ * map key 7 (auxiliary_data_hash) and embeds the bytes as the signed transaction's auxiliary_data
+ * element.
  *
  * Cardano signing is MPC/TSS: every co-signing device (iOS / Android / Extension) builds the input
  * independently and the Blake2b sighash must match byte-for-byte, so this encoding MUST be
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoCIP20.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoCIP20.kt
@@ -1,0 +1,143 @@
+package com.vultisig.wallet.data.crypto
+
+import com.vultisig.wallet.data.common.Utils
+import kotlin.experimental.and
+
+/**
+ * Canonical CIP-20 transaction-metadata encoder for Cardano memos.
+ *
+ * A Cardano memo is CIP-20 metadata under the registered label 674:
+ *
+ *     { 674: { "msg": [ "<chunk1>", "<chunk2>", ... ] } }
+ *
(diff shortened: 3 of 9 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
```diff
@@ -289,6 +289,7 @@ constructor(
                                     sendMaxAmount = isMaxAmountEnabled,
                                     ttl = ttl.toLong(),
                                     utxos = utxos,
+                                    memo = memo,
```
rkokhatskyi, line 292: `GasFeeOrchestrator.collectSpecific` calls `getSpecific` with no memo argument, so the Cardano balance gate and displayed fee stay memo-blind while the signed transaction forces the real, memo-inclusive byteFee, understating the fee shown right before signing

rkokhatskyi, line 292: A memo can push this planner past `Error_low_balance`, and the catch here falls back to a flat fee that ignores memo bytes while `CardanoHelper` still force-signs the same memo, risking `FeeTooSmallUTxO` after a full MPC round, needs mainnet confirmation

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
```diff
@@ -51,9 +53,14 @@ object CardanoHelper {
                         .setChangeAddress(changeAddress)
                         .setForceFee(forceFee)
                 )
-                // TODO: Implement memo support when WalletCore adds Cardano metadata support
                 .setTtl(ttl)
 
+        // CIP-20 memo (label 674). When present, WalletCore commits
+        // blake2b-256(auxDataCbor) into the body at map key 7 and emits the aux
+        // CBOR as element [3] of the signed tx. The encoder is byte-parity pinned
```
rkokhatskyi: This comment says WalletCore embeds the aux CBOR as element [3], but `compileWithSignatures` here resolves to `encodeTransactionWithSig`, which builds a 3-element array `[txAux, signatures, auxData]`, so Android's real aux data sits at index [2], not [3]

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
```diff
@@ -51,9 +53,14 @@ object CardanoHelper {
...
+        // to the SDK golden vector so co-signers agree on the sighash.
+        cip20AuxData(memo)?.let { input.setAuxiliaryData(ByteString.copyFrom(it)) }
```
rkokhatskyi, line 63: `DefaultSendStrategy` reads `memoFieldState.text` twice on different dispatchers, once for `Transaction.memo` and again inside `Dispatchers.IO` for fee estimation, so a keystroke landing between the two can force a byteFee sized for a different memo than the one embedded here

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
```diff
@@ -95,9 +102,21 @@ object CardanoHelper {
             ttl = ttl.toLong(),
             utxos = keysignPayload.utxos,
             forceFee = byteFee,
+            memo = keysignPayload.memo,
```
rkokhatskyi, line 106: `JoinKeysignFeeResolver`'s Cardano fallback disagrees with this, hardcoding `UtxoFeeService`'s flat 180000-lovelace default instead of reading `payload.blockChainSpecific.Cardano.byteFee`, so a memo that inflates the real signed fee never reaches the co-signing device's approval screen

data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoCIP20.kt
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Split a memo into chunks of at most 64 UTF-8 bytes, respecting UTF-8 codepoint boundaries.
+     *
+     * A multi-byte codepoint straddling the 64-byte boundary is moved entirely to the next chunk
+     * rather than torn — a torn codepoint decodes to U+FFFD and corrupts the memo on-chain. UTF-8
+     * continuation bytes have the top bits `10xxxxxx`; we walk back off them until the cut lands on
+     * a leading byte. An empty memo yields a single empty-string chunk (matches the SDK, which
+     * always emits at least one `msg` element).
+     */
+    fun memoToChunks(memo: String): List<String> {
```
rkokhatskyi, line 44: `memoToChunks` has no upper bound on memo length or chunk count, and neither the Send memo field nor WalletCore's Cardano signer cap transaction size, so a very long memo can grow the tx past protocol max size with no error until broadcast, needs mainnet confirmation

app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
```diff
@@ -550,6 +550,7 @@ class ChainHelpersTest {
                     sendMaxAmount = cardano.sendMaxAmount,
                     ttl = cardano.ttl.toLong(),
                     utxos = internalPayload.utxos,
+                    memo = internalPayload.memo,
```
rkokhatskyi, line 553: No test threads a non-null memo through the real signing pipeline, since both Cardano fixtures here have memo null, so `buildSigningInputBuilder`'s `setAuxiliaryData` wiring into `getPreSignedImageHash` is never exercised, unlike iOS's `CardanoCip20SigningTests` which assert the sighash changes with a memo

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
johnnyluo: Fixed. Threaded the memo into `GasFeeOrchestrator.collectSpecific` (new `memoFlow`, gated to Cardano only so other chains do not refetch specifics on memo keystrokes), so the displayed byteFee now prices the aux bytes while editing — matching the submit-time and signed fee.

johnnyluo: Acknowledged, leaving as-is. This flat-fee fallback only fires when the planner throws (typically insufficient balance), and in that state the actual signing/plan would fail too, so it never reaches a real broadcast. It predates the memo change and there is no better size-aware fallback once planning has failed. The common path uses `estimateFee(memo)` and is memo-accurate. Tagged for the mainnet checklist rather than a code change.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
johnnyluo: Fixed the comment. The exact array index is WalletCore-internal and not load-bearing here, so I softened both comments (here and `CardanoCIP20`) to say the aux CBOR is embedded as the tx auxiliary_data element without asserting a specific index.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
johnnyluo: Good catch — fixed. `getSpecific` now reuses the `memo` value captured earlier in `submit()` instead of re-reading `memoFieldState.text` inside `Dispatchers.IO`, so the fee-sizing memo and the embedded `Transaction.memo` can no longer diverge on a mid-flight keystroke.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
johnnyluo: Fixed. `computeJoinKeysignNetworkFee` now has a `BlockChainSpecific.Cardano` branch returning `byteFee`, so the co-signing device surfaces the forced memo-inclusive fee instead of the flat 180000 default. Also short-circuited the now-unused fallback fetch for Cardano in `resolveFallbackFeeAmount`.

data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoCIP20.kt
johnnyluo: Leaving as-is for now. The protocol max tx size is ~16KB, far above any realistic memo, and the encoder deliberately never tears codepoints. A hard cap belongs at the Send memo input (UX/validation) rather than in the CBOR encoder, and this is flagged as a mainnet-checklist item. Happy to add a UI-level length guard in a follow-up if desired.

app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
johnnyluo: Fixed. Added `cardanoMemoChangesSighash` to `ChainHelpersTest`, which threads a non-null memo through `getPreSignedImageHash` and asserts the sighash differs from the memo-less payload — exercising the `setAuxiliaryData` -> key-7 hash wiring, mirroring iOS CardanoCip20SigningTests.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
rkokhatskyi: One clarification on the deferral: force_fee just clamps to available balance and never re-derives from size, so the flat fallback (lower than the size-based fee that triggered Error_low_balance) can let signing succeed where the memo-aware estimate failed, rather than failing the same way again. Still narrow and non-fund-losing, just wanted the mainnet-checklist note to reflect the actual mechanism.

rkokhatskyi (approved):
LGTM

PR #5170, opened 2026-07-05
fix(keysign): reach a terminal state on a hashless broadcast

### LINKED ISSUE #5149: [Keysign] Broadcasted result with null txHash never reaches a terminal state (infinite spinner)
In `KeysignViewModel` the `is KeysignBroadcastResult.Broadcasted` branch transitions to a finished/polling state only inside `if (txHash != null)` — there is **no `else`** (`app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt:817-845`).

`BroadcastTxUseCase.invoke` returns `String?` and can yield null (e.g. `orKnownHash` returns null when both the RPC hash and `tx.transactionHash` are blank; `Broadcasted.txHash` is explicitly nullable). When broadcast succeeds but yields no hash, `signingState` stays at the last signing state (`KeysignECDSA`/`EdDSA`) forever — infinite spinner, no success, no error. The user cannot tell if funds moved and may force-retry the whole keysign → **double-send risk**.

### Fix sketch
Add an `else` that transitions to a terminal state — `KeysignFinished(Broadcasted)` if a hashless broadcast is legitimately "submitted", otherwise `Error` — mirroring the non-null path.

**Effort:** S · **Risk:** LOW.

johnnyluo:
## Summary

Fixes #5149.

In `KeysignViewModel`, the `is KeysignBroadcastResult.Broadcasted` branch only transitioned to a terminal state **inside** `if (txHash != null)` — there was no `else`.

`Broadcasted.txHash` is nullable, and a successful broadcast can legitimately yield `null` (e.g. `orKnownHash` returns `null` when both the RPC-returned hash and the local `tx.transactionHash` are blank). When that happened, `signingState` stayed at the last signing state (`KeysignECDSA`/`EdDSA`) forever — an infinite spinner with no success and no error. The user can't tell whether funds moved and may force-retry the whole keysign → **double-send risk**.

## Changes

- Add the missing `else` branch: a hashless-but-submitted broadcast now lands on the terminal `KeysignFinished(TransactionStatus.Broadcasted)` state, mirroring the existing non-polling success path. Because the use case only returns `Broadcasted` on a successful submit, a null hash means "submitted, hash unavailable" — a terminal success, not an error. Without a hash we can't save history or poll status, so those steps are skipped.
- Extract the result→state mapping into an `internal suspend fun applyBroadcastResult(result)` seam so it can be unit-tested without driving the full (static-`SigningHelper`-dependent) broadcast path.

## Test plan

- Added `KeysignViewModelApplyBroadcastResultTest`:
  - `Broadcasted(txHash = null)` → `signingState == KeysignFinished(Broadcasted)` (the bug).
  - `Broadcasted(txHash = "0x…")` on a non-status chain → `KeysignFinished(Broadcasted)` with `txHash` set (guards the refactor).
- `./gradlew :app:testDebugUnitTest --tests "…KeysignViewModelApplyBroadcastResultTest"` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cases where broadcasts that return no transaction hash could leave the app in an unfinished signing flow; such broadcasts now correctly complete.
  * Prevented users from navigating back while signing/broadcasting is in progress to avoid interruption-related issues.
  * Improved broadcast handling to treat successful and benign duplicate outcomes correctly, and to surface meaningful errors on genuine rejections.
* **Tests**
  * Added tests for broadcast-result handling with and without a transaction hash.
  * Updated API tests to assert errors are thrown for genuine rejection cases with the expected message.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
@@ -782,15 +782,18 @@ constructor(
     private suspend fun broadcastTransaction() {
         val payload = keysignPayload ?: return
 
-        when (
-            val result =
-                broadcastKeysign(
-                    vault = vault,
-                    payload = payload,
-                    signatures = signatures,
-                    isInitiatingDevice = isInitiatingDevice,
-                )
-        ) {
+        applyBroadcastResult(
+            broadcastKeysign(
+                vault = vault,
+                payload = payload,
+                signatures = signatures,
+                isInitiatingDevice = isInitiatingDevice,
+            )
+        )
+    }
+
+    internal suspend fun applyBroadcastResult(result: KeysignBroadcastResult) {
+        when (result) {
             is KeysignBroadcastResult.ApprovalNotConfirmed -> {
                 _state.update {
                     it.copy(
@@ -840,6 +843,17 @@ constructor(
                                     KeysignState.KeysignFinished(TransactionStatus.Broadcasted)
                             )
                         }
+                    }
+                } else {
+                    // The broadcast succeeded but produced no hash. Land on the terminal
+                    // "broadcasted" state instead of leaving signingState at the last signing
+                    // state forever (infinite spinner → the user may force-retry and double-send).
+                    // Without a hash we cannot save history or poll status.
+                    _state.update {
+                        it.copy(
+                            signingState =
+                                KeysignState.KeysignFinished(TransactionStatus.Broadcasted)
+                        )
                     }
                 }
             }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
@@ -44,34 +44,39 @@ internal class RippleApiImp @Inject constructor(private val http: HttpClient) : 
 
             val rpcResp = response.bodyOrThrow<RippleBroadcastResponseResponseJson>()
 
+            val engineResult = rpcResp.result.engineResult
             val resultMessage = rpcResp.result.engineResultMessage
+            val hash = rpcResp.result.txJson?.hash
 
-            if (rpcResp.result.engineResult != "tesSUCCESS") {
-                if (
-                    resultMessage?.contains("The transaction was applied", ignoreCase = true) ==
-                        true ||
-                        resultMessage.equals(
-                            "This sequence number has already passed.",
-                            ignoreCase = true,
-                        ) ||
-                        resultMessage.equals("The transaction is redundant.", ignoreCase = true)
-                ) {
-                    if (!rpcResp.result.txJson?.hash.isNullOrBlank()) {
-                        return rpcResp.result.txJson.hash
-                    }
+            // A benign duplicate-broadcast race: the peer's identical transaction was already
+            // applied/queued, so the transaction is (or will be) on-chain and we can recover its
+            // hash rather than treat it as a failure.
+            val alreadyApplied =
+                resultMessage?.contains("The transaction was applied", ignoreCase = true) == true ||
+                    resultMessage.equals(
+                        "This sequence number has already passed.",
+                        ignoreCase = true,
+                    ) ||
+                    resultMessage.equals("The transaction is redundant.", ignoreCase = true)
+
+            if (engineResult == "tesSUCCESS" || alreadyApplied) {
+                if (!hash.isNullOrBlank()) {
+                    return hash
                 }
-                return resultMessage ?: ""
-            } else {
-                val hash = rpcResp.result.txJson?.hash
-                return if (!hash.isNullOrBlank()) {
-                    hash
-                } else {
-                    resultMessage ?: ""
-                }
+                // Submitted but the node returned no hash. Don't invent one from the message; let
+                // the caller's on-chain recovery confirm using the locally computed hash instead.
+                error("XRP broadcast returned no transaction hash (engine_result=$engineResult)")
             }
+
+            // Any other engine result is a genuine rejection (e.g. temBAD_FEE, tecUNFUNDED). Throw
+            // the node's message instead of returning it as a fake txid, so the keysign surfaces
+            // the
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -782,15 +782,18 @@ constructor(
...
+            broadcastKeysign(
+                vault = vault,
+                payload = payload,
+                signatures = signatures,
+                isInitiatingDevice = isInitiatingDevice,
+            )
+        )
+    }
+
+    internal suspend fun applyBroadcastResult(result: KeysignBroadcastResult) {
```
rkokhatskyi, line 795: `if (txHash != null)` further down allows unsafe behavior because `RippleApi.kt:63` returns the engine's rejection message as a fake non-null hash on real failures, while iOS's `RippleService.swift` throws instead of treating that text as a txid.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -782,15 +782,18 @@ constructor(
...
+        when (result) {
             is KeysignBroadcastResult.ApprovalNotConfirmed -> {
```
rkokhatskyi, line 797: This branch never invalidates the balance cache even though `BroadcastKeysignUseCase.kt:109`'s `evmApi.sendTransaction` already broadcast the approval and spent gas, so every approval timeout or revert leaves a stale native-token balance until an unrelated refresh.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -841,6 +844,17 @@ constructor(
                             )
                         }
                     }
+                } else {
+                    // The broadcast succeeded but produced no hash. Land on the terminal
+                    // "broadcasted" state instead of leaving signingState at the last signing
+                    // state forever (infinite spinner → the user may force-retry and double-send).
```
rkokhatskyi, line 850: This still allows a force-retry double-send because the in-progress `Keysign` composable in `KeysignScreen.kt` has no `BackHandler`, unlike `JoinKeysignView.kt:364`, so system back pops the nav entry and cancels `viewModelScope` before this terminal state ever lands.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
johnnyluo: Good catch — fixed in f1bf513af. RippleApiImp.broadcastTransaction now throws on a genuine engine rejection (anything that isn't tesSUCCESS or a benign already-applied/redundant/sequence-passed race) instead of returning engineResultMessage as a fake txid, matching iOS RippleService. BroadcastTxUseCase's recoverIfAlreadyBroadcast then verifies on-chain and re-throws if the tx never landed, so the keysign surfaces the failure rather than persisting the rejection text as a hash. Updated the characterization test to expect the throw.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
johnnyluo: Fixed in f1bf513af. Added BackHandler(enabled = state.isInProgress) {} to KeysignView, so while signing/broadcasting is in progress system back is consumed and can't pop the nav entry / cancel the ViewModel scope before the terminal state lands. Back is re-enabled once the flow reaches a finished/error state, mirroring JoinKeysignView.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
johnnyluo: Not acting on this one. The un-invalidated branch is the ERC-20 approval path, and an approval only spends gas on the native token — it doesn't move the balance being sent. The next balance refresh reconciles the small gas delta, so a dedicated cache invalidation here isn't warranted. Resolving.

data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
```diff
@@ -44,34 +44,39 @@ internal class RippleApiImp @Inject constructor(private val http: HttpClient) :
...
+                    ) ||
+                    resultMessage.equals("The transaction is redundant.", ignoreCase = true)
+
+            if (engineResult == "tesSUCCESS" || alreadyApplied) {
+                if (!hash.isNullOrBlank()) {
+                    return hash
                 }
+                // Submitted but the node returned no hash. Don't invent one from the message; let
+                // the caller's on-chain recovery confirm using the locally computed hash instead.
+                error("XRP broadcast returned no transaction hash (engine_result=$engineResult)")
```
rkokhatskyi, line 68: This can't actually recover because `RippleHelpter.kt:155` hardcodes the local `transactionHash` to `""`, so `recoverIfAlreadyBroadcast`'s `hash.isNotBlank()` guard is always false and a genuine hashless success here now throws instead of ever reaching on-chain recovery.

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
johnnyluo: Resolved by the #5171 merge that's now in this branch. That PR populates the local canonical hash (RippleHelpter.kt:157, `transactionHash = calculateTransactionHash(signedTransaction)`) instead of the old hardcoded "". So recoverIfAlreadyBroadcast's `tx.transactionHash.isNotBlank()` guard now passes: a thrown hashless success reaches isAlreadyOnChain and recovers the canonical hash rather than dead-ending, and genuine rejections still surface as errors. No code change needed.

rkokhatskyi (approved):
LGTM

PR #5171, opened 2026-07-05
fix(ripple): populate the canonical transaction hash on signed XRP txs

### LINKED ISSUE #5150: [Send] XRP signed transactions carry an empty transactionHash (untrackable)
`RippleHelpter.getSignedTransaction` returns `SignedTransactionResult(rawTransaction = ..., transactionHash = "")` — the hash is hardcoded empty (`data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt:153-156`), unlike every other chain helper which returns a real hash.

**Consequence:** for every XRP send, downstream status polling, explorer deep-links, and the "view transaction" affordance have no hash to work with — the user cannot track or confirm the send. `PolkadotHelper.kt:138-142` documents this exact failure class in a comment ("wrong/empty hash broke status lookups and the tx link").

### Fix sketch
Derive the canonical XRPL tx hash (SHA-512Half of the `TransactionType`-prefixed signed blob, or read it from WalletCore's `Ripple.SigningOutput` if exposed) and populate `transactionHash`.

**Effort:** S · **Risk:** LOW (additive; does not change the signed bytes).

johnnyluo:
## Summary
- `RippleHelper.getSignedTransaction` returned `transactionHash = ""`, unlike every other chain helper.
- The empty hash left XRP sends untrackable (status polling, explorer deep-links) **and** silently disabled the duplicate-broadcast recovery path in `BroadcastTxUseCase`, which only recovers when `tx.transactionHash` is non-blank.
- WalletCore's `Ripple.SigningOutput` exposes only `encoded`/`error`/`errorMessage` — no hash. So the fix derives the canonical XRPL transaction ID locally: **SHA-512Half (first 32 bytes of SHA-512) of the signed blob prefixed with `0x54584E00` (`TXN\0`), uppercased**.
- Additive change — it does not alter the signed bytes.

## Issue
Closes #5150

## Test Plan
- [x] New unit test `RippleHelperTest` verifies the derivation against a **real validated ledger transaction** (blob + node-assigned hash fetched from the XRPL cluster) — passes
- [ ] Manual: send XRP, confirm the tx hash appears and the explorer deep-link + status polling resolve
- [ ] Manual: verify duplicate-broadcast recovery no longer silently fails for XRP

on chain tx link: https://xrpscan.com/tx/055F94F6B8883A733C2AA59AA0D4E7571AA2C6B64BB808A2DBAD8D9C1456D48D

## Summary by CodeRabbit

* **Bug Fixes**
  * Ripple transactions now produce the correct canonical transaction hash (no longer returning an empty value).
  * Signed Ripple transaction data is now properly retained and used to derive the transaction identifier for tracking.

* **Tests**
  * Added JUnit coverage to confirm the derived Ripple transaction hash matches the expected canonical XRPL transaction ID.

### DIFF
--- a/data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelperTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelperTest.kt
@@ -1,0 +1,26 @@
+package com.vultisig.wallet.data.chains.helpers
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+@OptIn(ExperimentalStdlibApi::class)
+class RippleHelperTest {
+
+    /**
+     * Canonical XRPL transaction ID derivation, checked against a real validated transaction
+     * (ledger tx `005899AE…C974`). The blob is the binary form returned by the node's `tx` command,
+     * and the expected hash is the id the node and explorers index it under. This locks in that we
+     * hash SHA-512Half over the `TXN `-prefixed blob and uppercase the result.
+     */
+    @Test
+    fun `calculateTransactionHash matches the canonical XRPL transaction id`() {
+        val signedBlob =
+            "1200072200000000240F81187C20190F81186B201B064808B764D5484DDC1BC9340000000000000000000000000045555200000000002ADB0B3959D60A6E6991F729E1918B716392523065400000008841978068400000000000000A73210253C1DFDCF898FE85F16B71CCE80A5739F7223D54CC9EBA4749616593470298C574473045022100A92E57F5E483E3DBBE6D88F75EB8A0193C97C853095868140CE83EFBDCB356C002206E7857B5EC8F0482D884E6D38C4548820C53B1EA81A4562D3DDB41253C287502811472A3DE6B0973062D5F2FE77383EF02F0C17901AB"
+                .hexToByteArray()
+
+        assertEquals(
+            "005899AE70C8A8E0148C331956CF37D216596BB757764EA11F3B19392609C974",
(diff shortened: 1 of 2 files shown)


data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelperTest.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.junit.jupiter.api.Test
+
+@OptIn(ExperimentalStdlibApi::class)
+class RippleHelperTest {
+
+    /**
+     * Canonical XRPL transaction ID derivation, checked against a real validated transaction
+     * (ledger tx `005899AE…C974`). The blob is the binary form returned by the node's `tx` command,
+     * and the expected hash is the id the node and explorers index it under. This locks in that we
+     * hash SHA-512Half over the `TXN `-prefixed blob and uppercase the result.
```
rkokhatskyi: This says the blob is TXN prefixed with a trailing space but the fourth prefix byte is 0x00 not a space, matching RippleHelpter.kt:175's correct description of TXN mnemonic plus a zero byte

rkokhatskyi (approved):

data/src/test/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelperTest.kt
johnnyluo: Fixed in 89ded42c8 — corrected the wording to `0x54584E00` (`TXN` + zero byte), matching the prefix in RippleHelpter.kt:175. Thanks.

rkokhatskyi (approved):
LGTM

PR #5172, opened 2026-07-05
fix(swap): show refund/failure reason on the transaction done screen

### LINKED ISSUE #5152: [Swap] Refund/failure reason is dropped on the done screen (shown only in history)
When a swap is refunded or fails, the app resolves a human-readable reason (e.g. "deposits are paused for asset (ETH.USDT…)") and carries it on the status object — but the done screen the user is looking at renders only a generic label and discards the reason. The reason is shown only later, in the separate transaction-history detail sheet.

### Evidence
- The reason is plumbed onto the status: `TransactionStatus.Refunded(val reason: UiText)` / `Failed(reason)`, built from the polled result at `app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt:962-963` (reason = `metadata.refund.reason`).
- The done screen renders only static strings, ignoring `reason`:
  - Title: `TxDoneScaffold.kt:163-165` → `stringResource(R.string.transaction_status_refunded)` / `transaction_failed`.
  - Status row: `TxDoneScaffold.kt:268-273` maps `Refunded`/`Failed` to a fixed label, not the payload.
- The reason *is* rendered elsewhere — the history detail sheet's `RefundReasonBanner` (`TransactionDetailBottomSheet.kt`).

### Impact
A user whose swap just refunded (or failed) sees only "Refunded" / "Transaction failed" with no explanation, at the exact moment they're looking for one. They must know to dig into transaction history to find out why (paused pool, slip exceeded, etc.). The most teachable moment — right when the money flow resolves — is silent.

### Fix sketch
In `TxDoneScaffold`, when status is `Refunded`/`Failed` and `reason` is non-blank, show the reason beneath the status label — reuse the existing `RefundReasonBanner` component. Data is already resolved and rendered elsewhere, so this is display wiring only.

**Effort:** S · **Risk:** LOW (display-only).

aminsato:
## Summary

Closes #5152.

When a swap is **refunded or fails**, the app already resolves a human-readable reason (e.g. *"deposits are paused for asset (ETH.USDT…)"*) and carries it on `TransactionStatus.Refunded(reason)` / `Failed(cause)`. But the **done screen** — the exact moment the user is looking for an explanation — rendered only a generic *"Refunded"* / *"Transaction failed"* label and dropped the reason. The reason surfaced only later, buried in the separate transaction-history detail sheet.

This wires the reason onto the done screen:

- Promoted the existing `RefundReasonBanner` from `private` → `internal` so it can be reused within the `transaction` package (no duplicate component).
- In `TxDoneScaffold`, when the status is `Refunded`/`Failed` and the reason/cause is non-blank, render the banner beneath the status label.

Display-only wiring — the data is already resolved. `TxDoneScaffold` is the shared scaffold behind both the send and swap done screens, so both benefit.

## Before / After

| Before | After |
|--------|-------|
| before | after |

Same `Refunded` state on the done screen — **before** drops the reason, **after** surfaces it in the amber banner.

## Test plan

- [x] Rendered the done screen in `Refunded` state via PreviewActivity on emulator; reason banner appears beneath the label.
- [x] `Broadcasted`/`Confirmed`/`Signed` states unchanged (banner only shows when a non-blank reason exists).
- [x] Banner hides when the transaction-detail view is expanded (reason already shown there), so no duplication.

## Summary by CodeRabbit

* **New Features**
  * Added support for showing the refund reason on the transaction completion screen (only when the transaction is refunded and the reason is not blank).
  * The banner displays the reason with the first character title-cased for improved readability.

* **Bug Fixes**
  * Improved transaction result visibility so users can better understand the refund outcome when a reason is available.


app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
```diff
@@ -132,58 +133,76 @@ private fun SuccessTransaction(
...
-                        animation = R.raw.riv_transaction_error,
-                        modifier = Modifier.fillMaxWidth(),
-                    )
-                } else {
-                    Image(
-                        painter = painterResource(R.drawable.img_tx_overview_bg),
-                        contentDescription = null,
-                        alignment = Alignment.Center,
-                        modifier = Modifier.padding(horizontal = 48.dp).fillMaxWidth(),
+            Column {
```
rkokhatskyi, line 136: `RefundReasonBanner` sits inside `AnimatedVisibility(isTransactionDetailVisible.not())` while `SendTxOverviewScreen.kt:123` and `SwapTransactionOverviewScreen.kt:127` detailContent only render `TransactionStatusRow`, so tapping Transaction Details deletes the reason this PR just added instead of showing it there too

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
```diff
@@ -132,58 +133,76 @@ private fun SuccessTransaction(
...
-                            .align(Alignment.BottomCenter)
-                            .padding(bottom = 48.dp),
-                )
+
+                // Surface the protocol's refund/failure reason (paused pool, slip exceeded, etc.)
+                // right on the done screen, not only later in transaction-history detail.
+                val statusReason =
+                    when (val status = transactionStatus) {
+                        is TransactionStatus.Refunded -> status.reason
+                        is TransactionStatus.Failed -> status.cause
```
rkokhatskyi: `status.cause` can surface a stale refund reason since THORChain/Maya classifies Failed before the refund posts, `KeysignTxStatusPoller.kt:182` then stops polling and `TransactionDao.kt:80` excludes Failed rows from repoll forever, so a refunded user reads their funds failed

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
```diff
@@ -132,58 +133,76 @@ private fun SuccessTransaction(
...
+                        else -> null
+                    }
+                statusReason
+                    ?.asString()
+                    ?.takeIf { it.isNotBlank() }
+                    ?.let { reason ->
+                        RefundReasonBanner(reason = reason)
```
rkokhatskyi: `status.cause` for Solana is `status.err.toString()` (`SolanaStatusProvider.kt:21`) on a raw JsonElement, so a reverted swap shows literal JSON like InstructionError/Custom braces in this banner instead of a human reason, same class of bug for Cosmos rawLog and Ripple/Tron codes

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
```diff
@@ -132,58 +133,76 @@ private fun SuccessTransaction(
...
+                // Surface the protocol's refund/failure reason (paused pool, slip exceeded, etc.)
+                // right on the done screen, not only later in transaction-history detail.
+                val statusReason =
+                    when (val status = transactionStatus) {
+                        is TransactionStatus.Refunded -> status.reason
+                        is TransactionStatus.Failed -> status.cause
+                        else -> null
+                    }
+                statusReason
+                    ?.asString()
```
rkokhatskyi: `status.cause` for a poll timeout is the hardcoded literal at `KeysignViewModel.kt:965`, Confirmation taking longer than expected, wrapped as DynamicString which asString() never localizes, so non English users see this one English sentence on an otherwise fully translated screen

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TransactionDetailBottomSheet.kt
```diff
@@ -399,7 +399,7 @@ private fun DetailInfoRows(
  * inspect the on-chain refund.
  */
 @Composable
-private fun RefundReasonBanner(reason: String) {
+internal fun RefundReasonBanner(reason: String) {
```
rkokhatskyi, line 403: `RefundReasonBanner` renders reason verbatim with no capitalize transform, and Midgard reasons like deposits are paused for asset start lowercase per `ThorMayaChainStatusProviderTest.kt`, so this banner reads as a lowercase sentence fragment, replaceFirstChar titlecase would match the codebase convention

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
```diff
@@ -132,58 +133,76 @@ private fun SuccessTransaction(
...
-                        Theme.brockmann.body.l.medium.copy(
-                            brush = Theme.v2.colors.gradients.primary
-                        ),
-                    modifier =
-                        Modifier.fillMaxWidth()
-                            .align(Alignment.BottomCenter)
-                            .padding(bottom = 48.dp),
-                )
+
+                // Surface the protocol's refund/failure reason (paused pool, slip exceeded, etc.)
```
rkokhatskyi: This comment claims the reason is shown later in transaction detail, but `TransactionDetailBottomSheet.kt:117` only checks Refunded never Failed, so for Failed transactions this banner is the only place the reason ever appears, the comment should say so instead

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
aminsato: Good catch — the banner was inside `AnimatedVisibility(isTransactionDetailVisible.not())` and `detailContent` only renders `TransactionStatusRow`, so expanding detail dropped it. Moved the banner out into the outer Column in f369480 so it stays visible in both collapsed and expanded states.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
aminsato: Agreed — surfacing `Failed.cause` here is wrong. THORChain/Maya classify Failed before the refund posts and those rows never repoll, so a refunded user would read "failed." Fixed by dropping the `Failed` branch entirely; the done-screen banner now only shows `Refunded.reason` (the clean Midgard text). f369480

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
aminsato: Right — `SolanaStatusProvider` builds `Failed(status.err.toString())` from a raw JsonElement (same class of raw string for Cosmos rawLog / Ripple / Tron). Rather than sanitize each provider, the banner no longer shows `Failed.cause` at all — Refunded-only now. f369480

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
aminsato: Correct — the TimedOut → `"Confirmation taking longer than expected".asUiText()` literal at `KeysignViewModel:965` is a hardcoded English DynamicString and would never localize. It's no longer reachable by the banner now that Failed causes aren't shown. f369480

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TransactionDetailBottomSheet.kt
aminsato: Done — `RefundReasonBanner` now titlecases via `reason.replaceFirstChar { ch -> ch.titlecase(Locale.ROOT) }`, matching the `JoinKeysignEvmDisplayUtils` convention. f369480

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
aminsato: Fixed the comment. With the banner now Refunded-only, `Refunded.reason` *is* also shown in the detail sheet (`TransactionDetailBottomSheet:117`), so I rewrote the comment to explain the Refunded-only rationale instead of the old misleading claim. f369480

aminsato:
Thanks for the thorough review @rkokhatskyi — all six were spot on. Pushed f369480:

- **Refunded-only banner.** Dropped the `Failed.cause` branch, which resolves the stale-"failed" for THORChain refunds, the raw Solana JSON, and the untranslated timeout literal in one move. Only `Refunded.reason` (clean Midgard text) is shown now.
- Moved the banner out of the detail-collapsed `AnimatedVisibility` so it survives expanding Transaction Details.
- Titlecased the reason; corrected the misleading comment.

**Scope note:** the issue mentioned "refund/**failure** reason," but given the three problems with `Failed.cause`, I've kept failure causes off the done screen for now. If we do want to surface failures, I think that's a separate change to sanitize/localize `Failed.cause` into a proper `UiText` at the provider layer — happy to open a follow-up if you'd like it.

rkokhatskyi (approved):
LGTM

PR #5181, opened 2026-07-05
fix(solana): fall back to default compute-unit limit on a zero override

### LINKED ISSUE #5178: [Fix] Solana compute-budget limit silently zeroes out on a zero override
# [Fix] Solana compute-budget limit silently zeroes out on a zero override [Send / Solana]

## Problem
When building a Solana transaction, the compute-unit priority limit taken from the signing payload is passed straight into the `SetComputeUnitLimit` instruction with no positive-value check. If the incoming value is `0` (e.g. a payload built by a peer/joining device with a missing or zero override), the signed transaction ends up with a zero compute-budget limit instead of falling back to the app's default.

## Expected Behavior
If the payload's priority-fee compute limit is not a positive value, the code should fall back to the app's default Solana compute-unit limit constant instead of encoding `0`.

## Steps to Reproduce
1. Construct a Solana keysign payload whose priority-fee compute limit is `0`.
2. Run it through `SolanaHelper.getPreSignedInputData`.
3. Inspect the resulting `SetComputeUnitLimit` instruction — its value is `0` instead of the default limit.

## Solution
Guard the value before it's applied to the compute-budget instruction, falling back to the existing default limit constant whenever the provided value is not positive.

### Must Do
- [ ] Add a positive-value guard before the compute-limit is set, falling back to the default limit constant
- [ ] Add a unit test asserting a payload with limit `0` produces a signing input using the default limit

### Must NOT Do
- Don't change the default limit constant's value
- Don't touch unrelated Solana fee-estimation logic

## Acceptance Criteria
- [ ] New unit test passes (and fails without the fix)
- [ ] Existing Solana signing tests remain green

rkokhatskyi:
## Summary
`SolanaHelper.getPreSignedInputData` set the Solana compute-unit limit straight from `BlockChainSpecific.Solana.priorityLimit` with no positive-value check. A zero value (e.g. a joining device reconstructing the payload from a peer's proto whose `compute_limit` came through as `"0"`) encoded a zero compute budget instead of falling back to the app's default (`SOLANA_PRIORITY_FEE_LIMIT` = 100000). Guarded the value with the same `takeIf { it > BigInteger.ZERO } ?: default` idiom already used elsewhere in this codebase for peer-supplied numeric overrides.

Closes #5178

## Testing
- `./gradlew :data:testDebugUnitTest` (full `data` module: 1870 passed, 0 failed, 39 skipped — pre-existing + these two gracefully skip locally since WalletCore's native lib is Android-only and unavailable to the host JVM, same as the rest of this file's JNI-dependent tests)
- `./gradlew :data:lintDebug`

## Summary by CodeRabbit

* **Bug Fixes**
  * Solana swap transactions now handle priority fee limits more reliably by using a default value when none is provided and preventing overly large values from being used.
* **Tests**
  * Added coverage for Solana swap payloads to verify both default fallback behavior and preserving valid custom priority limits.
  * Improved test handling when native libraries are unavailable, allowing unsupported environments to skip cleanly.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SolanaHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SolanaHelper.kt
@@ -64,6 +64,12 @@ class SolanaHelper(private val vaultHexPublicKey: String) {
         }
         val toAddress = AnyAddress(keysignPayload.toAddress, coinType)
 
+        val effectivePriorityLimit =
+            solanaSpecific.priorityLimit
+                .takeIf { it > BigInteger.ZERO }
+                ?.min(BigInteger.valueOf(Int.MAX_VALUE.toLong()))
+                ?.toInt() ?: SOLANA_PRIORITY_FEE_LIMIT
+
         val input =
             Solana.SigningInput.newBuilder()
                 .setV0Msg(true)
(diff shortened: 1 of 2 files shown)


johnnyluo:
We should respect the value in keysign payload, priority fee limit can be 0

PR #5182, opened 2026-07-05
fix(qbtc): block claim co-sign when the vault password is empty

### LINKED ISSUE #5179: [Fix] QBTC Fast Vault claim allows an empty password through to the server co-sign
# [Fix] QBTC Fast Vault claim allows an empty password through to the server co-sign [QBTC Claim]

## Problem
On the QBTC Fast Vault claim flow, submitting the password field with no client-side non-empty check lets an empty password reach the server co-sign call. There's no guard blocking submission when the password field is blank.

## Expected Behavior
Submitting with an empty password should show a validation error and must never trigger the server co-sign request.

## Steps to Reproduce
1. Open QBTC claim on a Fast Vault.
2. Leave the password field empty.
3. Trigger confirm/submit.
4. Observe the co-sign call still fires (or the confirm control isn't reliably disabled).

## Solution
Add an explicit non-empty check on the password field before invoking the co-sign orchestrator, and keep the confirm control disabled while the field is blank.

### Must Do
- [ ] Add an empty-password guard before invoking the co-sign orchestrator
- [ ] Ensure the confirm control stays disabled while the password field is blank
- [ ] Add a unit test asserting an empty-password submit emits a validation error and never calls the orchestrator

### Must NOT Do
- Don't change server-side validation
- Don't touch unrelated steps of the QBTC claim flow

## Acceptance Criteria
- [ ] New unit test passes (and fails without the fix)
- [ ] Manual check: confirm control disabled with an empty password field

rkokhatskyi:
## Summary
Fixes #5179. The QBTC Fast Vault claim password sheet had no validation at all: the Confirm button had no disabled state, and the keyboard "Go" action called straight into `QbtcClaimViewModel.confirm()` regardless of whether the field was blank, so an empty password could reach the server co-sign call.

- `FastVaultPasswordSheet` now disables Confirm while the password field is blank and guards the shared submit path so the IME action can't bypass it either.
- `QbtcClaimViewModel.confirm()` adds a defense-in-depth check: an empty password sets a `passwordError` on the `Selecting` state (reusing the existing `password_should_not_be_empty` string) and returns without ever launching the round runner.

## Screenshots

| After (blank password, Confirm disabled) |
|---|
| after |

A matching "before" capture (button not yet disabled) couldn't be produced — the local emulator hit repeated ANRs from unrelated host load while trying to interact with the sheet, reproducing identically on the unmodified pre-fix build, so it's an environment issue rather than a regression from this change.

## Testing
- `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.qbtc.QbtcClaimViewModelTest"` — 12/12 passing, including the new empty-password test
- `./gradlew :app:ktfmtCheck`
- `./gradlew :app:lintDebug`
- `./gradlew :app:assembleDebug`

## Summary by CodeRabbit

* **Bug Fixes**
  * Added Fast Vault password validation in the QBTC claim flow.
  * Empty or whitespace-only passwords now display an error message/footnote and stop the claim from proceeding.
  * The password input and submit CTA now reflect validation (submit enabled only when the password is non-empty).

* **Tests**
  * Added unit tests covering empty and whitespace-only password validation, ensuring the claim action is not triggered when validation fails.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/qbtc/QbtcClaimViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/qbtc/QbtcClaimViewModel.kt
@@ -241,7 +241,7 @@ constructor(
         val claim = readyClaim() ?: return
         // The password sheet already disables Confirm while the field is blank, so this is a
         // defense-in-depth guard against ever reaching the co-sign call with an empty password.
-        if (fastVaultPassword.isEmpty()) {
+        if (fastVaultPassword.isBlank()) {
             passwordError = UiText.StringResource(R.string.password_should_not_be_empty)
             emitSelecting()
             return
--- a/app/src/main/java/com/vultisig/wallet/ui/models/qbtc/QbtcClaimViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/qbtc/QbtcClaimViewModel.kt
@@ -11,6 +11,7 @@ import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import androidx.navigation.toRoute
+import com.vultisig.wallet.R
 import com.vultisig.wallet.data.api.SessionApi
 import com.vultisig.wallet.data.blockchain.cosmos.qbtc.claim.ClaimableUtxo
 import com.vultisig.wallet.data.blockchain.cosmos.qbtc.claim.LoadClaimableQbtcUtxosUseCase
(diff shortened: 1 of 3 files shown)


rkokhatskyi:
Both nitpicks point at the same real gap, so fixed together: `isEmpty()` let a password consisting only of spaces reach the co-sign call in both `QbtcClaimViewModel.confirm()` and the sheet's `canSubmit`, whereas the guard's whole point is to stop invalid passwords from reaching the server, so switched both to `isBlank()` in a96ea2869 and added a dedicated whitespace-only unit test alongside the existing empty-password one. I skipped the trimming part of the suggestion since iOS has the identical `password.isEmpty()` check with no trim (`FastVaultEnterPasswordView.swift:90`, `QBTCClaimViewModel.swift:295`) and Windows' zod schema only enforces `.min(1)` on the raw string, so trimming the value that's actually sent to the server co-sign would be an Android-only behavior change beyond what this ticket is about — the blank check itself is purely a client-side UX guard and doesn't touch what gets sent once a real password is present.

johnnyluo (approved):
LGTM

PR #5190, opened 2026-07-06
test(solana): guard send fee estimate against priority double-count

### LINKED ISSUE #5115: [Fix] Solana displayed network fee double-counts the priority fee (~2× actual) [Send/Solana]
## Observed

On a native SOL send, the app displays a **Network Fee of `0.000205` SOL**, but the transaction actually pays **`0.000105` SOL** on-chain — the displayed fee is ~2× the real fee.

Example (Fast Vault SOL send):
- Tx: https://orb.helius.dev/tx/3GuQC3eYetR7AK6LF6fxV2WFpjWgjcuiQF7ytxzt3mRFpXHszxyg8saW5Yhw8QmddfT3KUm2qG1dXcZuoMrCyJ7y
- Explorer fee breakdown: Base `0.000005` + Priority `0.0001` = **Total `0.000105` SOL**
- App "Transaction complete" → Network Fee: **`0.000205` SOL**

## Cause (verified)

`0.000205 = DEFAULT_COIN_TRANSFER_BASE_FEE (0.000105) + priorityAmount (0.0001)` — the `SolanaFeeService.calculateDefaultFees` formula:

- `data/.../blockchain/solana/SolanaFeeService.kt:185` — `DEFAULT_COIN_TRANSFER_BASE_FEE = CoinType.SOLANA.toUnit("0.000105")` = 105,000 lamports.
- `data/.../blockchain/solana/SolanaFeeService.kt:180` — `amount = DEFAULT_COIN_TRANSFER_BASE_FEE + priorityAmount`.

The real base fee is `0.000005` SOL (5,000 lamports / signature, per `getFeeForMessage`). The `0.000105` constant is **not** a base fee — it already bundles `0.000005` base + `0.0001` priority. `calculateDefaultFees` then adds `priorityAmount` (another `0.0001`) on top, so the **priority fee is counted twice** → `0.000205` displayed vs `0.000105` paid.

The exact-calc path (`calculateFees`, which uses `getFeeForMessage` for the true base) would produce `0.000105`, matching the explorer — so a native send should not be showing the default estimate. Two things to fix/confirm:
1. `DEFAULT_COIN_TRANSFER_BASE_FEE` should be the real base (~`0.000005`), so the default path stops double-counting priority.
2. Confirm why a native send's displayed/stored fee went through `calculateDefaultFees` rather than `calculateFees` (possible fallback when the exact calc throws).

## Expected

Displayed network fee == on-chain total == `base + (priorityPrice × limit / 1e6)`.

## Cross-platform

Android-only. iOS/Windows keep base and priority separate and don't bundle a priority fee into a "base" constant.

## Related

Sibling of #5101 (signed-tx priority-fee conflation, fixed in #5114). That fix corrected the **signed** priority fee; this is a separate defect in the **displayed** fee estimate.

---
_Verified by hand against vultisig-android `main` and the on-chain tx above (arithmetic + the cited constant)._

### LINKED ISSUE #5174: test(solana): add coverage for send priority-fee estimate (regression guard for #5115)
## Context
PR #5156 fixed #5115 — the Solana send fee estimate was double-counting the priority fee (~2× the actual on-chain fee). The fix shipped in v1.0.112 **with no test coverage**.

## Ask
Add a unit test around the Solana send fee-estimate path that locks the correct behavior:
- priority fee is counted **once**, not conflated with base/compute-unit pricing
- displayed network fee matches the signed tx's fee (no double-count)
- covers the per-CU-price vs total-lamports distinction that #5101/#5114 also touched

## Why it matters
This is a money-path display value on the Send verify screen, and the surrounding Solana fee/priority-fee logic has churned repeatedly this sprint (#5101, #5114, #5115, #5125, #5127). A regression here silently misstates fees to users. A cheap guard test pays for itself.

Ref: fix PR #5156, original bug #5115.

aminsato:
Closes #5174.

## Context
PR #5156 fixed #5115 — the Solana **send** fee estimate double-counted the priority fee (~2× the real on-chain fee) — but shipped in v1.0.112 with no test coverage. This adds the regression guard.

`getFeeForMessage` already folds the ComputeBudget priority fee into the returned message fee (base signature fee + `price × limit / 1e6`). The fix stopped adding a separate priority term, so the displayed send fee is `messageFee + rentExemption`.

## Why it wasn't testable before
- Constructing `SolanaFeeService` triggered a WalletCore **JNI** call at class load (`CoinType.SOLANA.toUnit(...)` in the companion), which throws `UnsatisfiedLinkError` in JVM unit tests.
- The exact-fee path serializes the message via `SolanaHelper.getVersionedMessage()`, also JNI-backed.

## Changes
Two behavior-preserving, testability-only production tweaks in `SolanaFeeService`:
1. Compute `DEFAULT_COIN_TRANSFER_BASE_FEE` (0.000105 SOL = 105,000 lamports, SOL has 9 decimals) without `CoinType`, so the service constructs off-device. Value is identical.
2. Extract the versioned-message serializer into a constructor seam. The `@Inject` constructor keeps using the real `SolanaHelper`; an `internal` constructor lets the test stub it.

## Tests (`SolanaFeeServiceTest`, 4 cases)
- Native SOL send → `amount == messageFee` (priority **not** double-counted); no rent RPC.
- SPL send, recipient has **no** token account → `amount == messageFee + rentExemption` (each counted once).
- SPL send, recipient **has** a token account → no rent charged.
- Swap default path → locks the micro-lamports-per-CU → lamports conversion (`price × limit ÷ 1e6 + base`), the per-CU-price vs total-lamports distinction from #5101/#5114.

## Test plan
- [x] `./gradlew :data:testDebugUnitTest --tests "*SolanaFeeServiceTest"` — all 4 pass.
- [x] Mutation check: re-introducing the `+ priorityFee` double-count fails 3 tests; reverting makes them pass.
- [x] ktfmt applied.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/SolanaFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/SolanaFeeService.kt
@@ -11,14 +11,12 @@ import com.vultisig.wallet.data.chains.helpers.SolanaHelper
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.payload.BlockChainSpecific
 import com.vultisig.wallet.data.models.payload.KeysignPayload
-import com.vultisig.wallet.data.utils.toUnit
 import java.math.BigDecimal
 import java.math.BigInteger
 import java.math.RoundingMode
 import javax.inject.Inject
 import kotlinx.coroutines.async
 import kotlinx.coroutines.coroutineScope
-import wallet.core.jni.CoinType
 
 /**
  * Implementation of [FeeService] for Solana blockchain transactions.
@@ -52,7 +50,23 @@ import wallet.core.jni.CoinType
  *
  * @property solanaApi API interface for interacting with Solana blockchain RPC endpoints.
  */
-class SolanaFeeService @Inject constructor(private val solanaApi: SolanaApi) : FeeService {
+class SolanaFeeService
+internal constructor(
+    private val solanaApi: SolanaApi,
+    // Serializes a versioned Solana message for [getFeeForMessage]. Backed by WalletCore JNI in
+    // production; overridable so the fee-composition arithmetic can be unit-tested without the
+    // native library.
+    private val serializeVersionedMessage:
+        (vaultHexPubKey: String, payload: KeysignPayload) -> String,
+) : FeeService {
+
+    @Inject
+    constructor(
+        solanaApi: SolanaApi
+    ) : this(
+        solanaApi,
+        { vaultHexPubKey, payload -> SolanaHelper(vaultHexPubKey).getVersionedMessage(payload) },
+    )
 
     override suspend fun calculateFees(transaction: BlockchainTransaction): Fee {
         // Jupiter builds its own versioned tx at signing time. Serializing a fake SPL/SOL transfer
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5195, opened 2026-07-06
test(swap): pin SwapKit EVM network fee to the oracle bond (#5121)

### LINKED ISSUE #5121: SwapKit EVM swaps: verify-screen Network Fee is understated (uses SwapKit's inbound estimate/placeholder, not the oracle-priced broadcast fee)
## Summary

On the swap verify screen, the **Network Fee** shown for EVM SwapKit routes is understated by orders of magnitude — for some sub-providers it renders as effectively zero (e.g. `0.00000000000000013 ETH`, i.e. a fraction of a cent on an ERC-20 swap). The displayed fee is taken from SwapKit's wire `fees[]` `inbound` estimate, which (a) is a near-zero placeholder for some EVM providers and (b) does not match the oracle-priced gas the tx is actually broadcast with. The **broadcast tx is priced normally** — so this is a display + gas-sufficiency-check bug, not a stuck-tx bug.

## Where

`data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt`

- Header doc (~L54–58): *"The route's `fees[]` entry with `type=="inbound"` for the source chain becomes `tx.swapFee` (raw native units) so the displayed Network Fee reconciles against the actual on-chain debit — especially on Solana / Chainflip / NEAR where the EVM gas envelope reads zero."*
- `inboundFeeRawUnits(...)` (~L339) reads `fees[].type=="inbound"` and scales by the source chain's **native** decimals.
- That value is assigned to the EVM swap payload's `swapFee` (~L415, L471, L494), which becomes the displayed Network Fee.

The intent (reconcile the displayed fee with the real on-chain debit) is correct for **non-EVM** sources (BTC PSBT, Solana, TRON), where the wire `inbound` entry is the real native miner fee. For **EVM** sources it's wrong: the real debit is the oracle-priced gas the broadcast commits to, not SwapKit's `inbound` estimate.

## Root cause / evidence

SwapKit's `inbound` estimate for EVM is computed at SwapKit's own (stale/low) gas price and, for some providers, is a placeholder. Live via the Vultisig proxy:

- `ETH.USDC → ARB.USDC` (NEAR): `fees[]` `inbound` = **0.000005455 ETH ≈ $0.009**.
- The `tx` envelope for the same route: `gas = 0x186a0` (100k), `gasPrice = 0x4946111` ≈ **0.077 gwei** (mainnet base fee is 1–30 gwei).
- FLASHNET-on-EVM is documented (iOS `SwapKitService`) to return a near-zero `inbound` placeholder (~hundreds of wei) → renders as `0.0000000000000001x ETH` / `$0.00`.

Meanwhile the swap flows through the `oneinchSwapPayload` broadcast path, where the EVM gas price is bumped to the chain fee oracle before signing (parity with iOS `OneInchSwaps.getPreSignedInputData`, which does `gasPrice = max(swapKit, oracle maxFeePerGas)`), so the user is actually charged the real (higher) fee.

## Impact

- Misleading fee on verify — a SwapKit swap looks nearly free; FLASHNET routes show ~$0.00.
- **Total Fee (fiat) under-reported.**
- **Gas-coin sufficiency check uses the understated fee** → a user with barely enough native gas passes validation, then the real (oracle-priced) tx needs more → on-chain "insufficient funds for gas" failure. This is the actionable risk.

## Suggested fix

For **EVM** sources, display the fee the broadcast actually commits to — the oracle-bumped `gasLimit × maxFeePerGas` (same numbers used at signing), not the SwapKit `inbound` estimate. Keep the `inbound`-from-`fees[]` path only for **non-EVM** (BTC/Solana/TRON) where it is the real native miner fee.

Add tests against the FLASHNET / NEAR / Chainflip EVM fixtures asserting the displayed fee ≈ the oracle-priced gas (not SwapKit's `inbound` placeholder), including the FLASHNET near-zero case.

## Scope / related

- Related but distinct from #5076 (`fix(swap): correct network fee overestimation on EVM aggregator swaps`) — that addressed EVM aggregator **over**estimation; this is SwapKit-route **under**estimation via the `inbound` display path.
- iOS has the same class of bug via a different path (`SwapKitService.evmNetworkFee` uses `tx.gasPrice × tx.gas`) — tracked separately in vultisig-ios.
- **Not affected:** Windows/extension (`vultisig-sdk`) — it discards SwapKit's `gasPrice` and computes the fee from the live RPC oracle.

_Verified against `vultisig-android` @ `ae7decbd6`._

aminsato:
## Summary

#5121 reports that the SwapKit EVM verify-screen **Network Fee** is understated — allegedly taken from SwapKit's near-zero `fees[].inbound` placeholder rather than the oracle-priced broadcast gas, with the same understated value feeding the gas-sufficiency check (the actionable "insufficient funds for gas at broadcast" risk).

Tracing the Android code, **this does not reproduce**. A SwapKit EVM route is wrapped as `SwapQuote.OneInch` (`SwapQuoteManager.fetchSwapKitQuote`) and its Network Fee is the oracle bond from `SwapGasCalculator.rebaseEvmSwapNetworkFee` — `oracle maxFeePerGas × routeGasLimit`, the same number the tx is signed with in `OneInchSwap`. This was already true at `ae7decbd6` (PR #5076), the exact commit the issue was verified against. The `inbound` placeholder only feeds the separate **"Estimated Fees"** row, where its ~$0 value is harmless to the Total.

This PR is **tests only** — it adds regression coverage that reproduces the issue's exact FLASHNET fixture and pins the correct behavior, so the invariant can't silently regress. It's the Android mirror of the iOS parity test in vultisig-ios#4723 (which closes the iOS sibling #4707 — a real bug there because iOS *excluded* SwapKit EVM from the oracle path; Android does not).

## Tests added

Fixture from the issue: `tx.gas=0x186a0` (100k), `tx.gasPrice=0x4946111` (≈0.077 gwei), inbound placeholder `130 wei`.

- **`SwapGasCalculatorTest`** — the rebased Network Fee is the oracle bond (`20 gwei × 100k = 0.002 ETH`), asserted `!=` the 130-wei placeholder and `!=` SwapKit's stale `gasPrice × gas` seed. `rebaseEvmSwapNetworkFee` doesn't even take the placeholder as an input.
- **`SwapQuotePipelineNetworkFeeTest`** — a `provider="SwapKit"` EVM quote takes the *same* oracle-rebase branch as 1inch (the branch keys on `OneInch && chain==EVM`, never reads `quote.fees`), so the placeholder reaches neither the displayed Network Fee nor the balance/gas-sufficiency check.
- **`SwapTransactionToUiModelMapperFeeBreakdownTest`** — the placeholder renders only in the "Estimated Fees" (swap-fee) row (~$0.00); Network Fee and Total are the oracle bond → Total is **not** under-reported. This pinpoints where the reporter's `0.0000…13 ETH` actually came from.

## Test plan

```bash
./gradlew :app:testDebugUnitTest \
  --tests "*SwapGasCalculatorTest" \
  --tests "*SwapQuotePipelineNetworkFeeTest" \
  --tests "*SwapTransactionToUiModelMapperFeeBreakdownTest"
```

`BUILD SUCCESSFUL` — `SwapGasCalculatorTest` 50/0/0, `SwapQuotePipelineNetworkFeeTest` 4/0/0, `SwapTransactionToUiModelMapperFeeBreakdownTest` 3/0/0.

## Recommendation

Close #5121 as already-addressed by #5076 / not-reproducible on Android; these tests stand as the proof and guard against regression. No production code changed.

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for swap fee and network fee calculations in EVM swap scenarios.
  * Verified that near-zero placeholder fees are ignored and the correct oracle-based fee is shown instead.
  * Added checks to ensure the total and fee breakdown match the expected swap route values.

### DIFF
--- a/app/src/test/java/com/vultisig/wallet/ui/models/mappers/SwapTransactionToUiModelMapperFeeBreakdownTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/mappers/SwapTransactionToUiModelMapperFeeBreakdownTest.kt
@@ -2,12 +2,17 @@
 
 package com.vultisig.wallet.ui.models.mappers
 
+import com.vultisig.wallet.data.api.models.quotes.EVMSwapQuoteJson
+import com.vultisig.wallet.data.api.models.quotes.OneInchSwapTxJson
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.models.EVMSwapPayloadJson
 import com.vultisig.wallet.data.models.FiatValue
+import com.vultisig.wallet.data.models.SwapProvider
 import com.vultisig.wallet.data.models.SwapTransaction.RegularSwapTransaction
 import com.vultisig.wallet.data.models.THORChainSwapPayload
 import com.vultisig.wallet.data.models.TokenValue
+import com.vultisig.wallet.data.models.getSwapProviderId
 import com.vultisig.wallet.data.models.payload.SwapPayload
 import com.vultisig.wallet.data.models.settings.AppCurrency
 import com.vultisig.wallet.data.repositories.AppCurrencyRepository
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5199, opened 2026-07-06
Scope tx-status notification manager as a singleton

### LINKED ISSUE #5193: [Fix] Transaction-status notification isn't cancelled on completion
# [Fix] Transaction-status notification isn't cancelled on completion [Send / Tx Status]

## Problem
When a tracked transaction reaches a terminal state, the code responsible for dismissing its foreground status notification calls `complete()` on a different instance of the notification manager than the one actually bound to the running foreground service/poller. Because the manager isn't scoped so both sides share one instance, the cancel call is effectively a no-op and the notification is left showing.

## Expected Behavior
Completing a transaction should reliably cancel/dismiss its foreground status notification.

## Steps to Reproduce
1. Start a transaction that uses the foreground tx-status notification/poller (e.g. a slower-confirming chain).
2. Let it reach a terminal state (success or failure).
3. Observe the notification remains visible instead of being dismissed.

## Solution
Ensure the poller and the code that calls `complete()` share the same manager instance (e.g. by scoping it as a singleton), so cancellation actually reaches the live notification.

### Must Do
- [ ] Scope the notification manager so the poller and the completion path use the same instance
- [ ] Add a test/assertion proving completion cancels the notification that was actually started

### Must NOT Do
- Don't change unrelated notification content or formatting
- Don't change the polling cadence/logic

## Acceptance Criteria
- [ ] New test/assertion passes (and fails without the fix)
- [ ] Manual check: notification disappears when a tracked transaction reaches a terminal state

rkokhatskyi:
Fixes #5193

`KeysignTxStatusPoller` (which binds the running foreground service) and `KeysignFlowViewModel.complete()` (which cancels its notification) each injected their own `TransactionStatusServiceManager`. With no scope, Hilt handed each site a separate instance, so `complete()` cancelled a manager that was never bound to the running service and the notification stayed visible after the transaction settled. Scoped the manager `@Singleton` so both sides share the same bound instance, matching how `PushNotificationManager` and other cross-ViewModel shared-state classes are already scoped in this codebase.

## Testing
- `./gradlew :app:testDebugUnitTest` (new `TransactionStatusServiceManagerTest` plus full suite)
- `./gradlew :app:assembleDebug`
- `./gradlew :app:lintDebug`
- `./gradlew :app:ktfmtCheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured the transaction status manager is treated as a single shared instance, improving consistency across the app.

* **Tests**
  * Added a regression test to verify the transaction status manager remains correctly configured as a singleton.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/services/TransactionStatusServiceManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/services/TransactionStatusServiceManager.kt
@@ -9,11 +9,18 @@ import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.usecases.txstatus.TransactionResult
 import dagger.hilt.android.qualifiers.ApplicationContext
 import javax.inject.Inject
+import javax.inject.Singleton
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
 
+/**
+ * Bound to the running foreground [TransactionStatusService], so it must be a singleton: every
+ * injection site (the poller that starts it and the screen that completes it) needs to observe and
+ * cancel the same bound service, not a fresh, never-bound instance.
+ */
+@Singleton
 class TransactionStatusServiceManager
 @Inject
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5200, opened 2026-07-06
fix(thorchain): unbond MAX uses bonded amount, not wallet balance

### LINKED ISSUE #5196: [Android] TC unbond MAX button fills wallet balance (max bondable) instead of bonded amount
## Summary
On Android, the THORChain **unbond** form's MAX / percentage buttons populate the wallet's spendable RUNE balance (the max you could *bond*) instead of the amount currently *bonded* to the selected node (the max you can unbond).

## Steps to reproduce
1. DeFi → Bonded → tap a bonded node → **Unbond**.
2. Tap **MAX** (or a percentage) on the amount field.

**Result:** amount fills with the wallet's spendable RUNE (max *bondable*).
**Expected:** amount fills with the node's *bonded* amount (max *unbondable*), i.e. 100% = full bond.

## Root cause (traced)
Unbond reuses the generic Send form with the wallet RUNE coin, and the per-node bonded amount is dropped at navigation:

1. `ThorchainDefiPositionsViewModel.onClickUnBond()` routes to `Route.Send(type=UNBOND, tokenId=runeCoin.id, address=nodeAddress)` — only the node address is passed, not the bonded amount (`Route.Send.amount` exists but is unused here).
2. `AccountsLoader` sends UNBOND down the `else` branch → `loadDeFiAddresses()`, which fills the RUNE account with the on-chain **wallet** RUNE balance (`AccountsRepository.loadDeFiAddresses` → `getDefiTokenBalanceAndPrice`). Bonded RUNE isn't in spendable balance, so this is exactly "max available to bond".
3. `AmountFractionManager.chooseMaxTokenAmount()` → `fetchPercentageOfAvailableBalance()` → `getAvailableTokenBalance(account, ZERO)` returns that wallet balance → MAX = wallet-spendable RUNE.
4. Same wrong source gates submit in `send/submit/UnbondStrategy.kt` (insufficient-balance check compares the entered unbond amount against the wallet balance).

The correct value already exists in `BondedNodePosition.amount` (shown as `node.bondedAmount` in the Bonded tab) — it's just never threaded into the form.

## Expected behavior (already correct on iOS + Windows)
The MAX/percentage base **and** the submit validation should use the node's **bonded** amount:
- **iOS** `UnbondTransactionViewModel`: resolves `coin.bondedNodes.first { $0.address == bondAddress }.bond`, pre-fills it on load, slider computes `availableAmount * pct` off that bond.
- **Windows** `useDepositBalance` / `useUnbondableBalanceQuery` + `UnbondForm`: slider computes `(pct / 100) * unbondableBalance`.

## Suggested fix
Thread the node's bonded amount through `Route.Send` (or resolve it in the UNBOND load path) and make MAX/percentage + the submit validation use the bonded amount instead of `getAvailableTokenBalance`. iOS's pre-fill-full-bond behavior is the cleanest target.

## Scope
Android only. iOS ✅ correct · Windows ✅ correct · SDK ➖ n/a (no TC unbond MAX surface).

---
_Filed after a cross-platform source trace. "Urgent" mapped to `high-priority` (no `urgent` label in this repo)._

aminsato:
## Summary

Fixes #5196. The THORChain **unbond** form's MAX / percentage buttons (and the submit-time balance check) filled the wallet's spendable RUNE balance (max *bondable*) instead of the amount currently *bonded* to the selected node (max *unbondable*). iOS and Windows already unbond against the bonded amount.

### Root cause
Unbond reuses the generic Send form with the wallet RUNE account. Only the node address was threaded through `Route.Send`; the per-node bonded amount was dropped, so the RUNE account's balance stayed the wallet-spendable balance. Both `AmountFractionManager` (MAX/percentage) and `UnbondStrategy` (submit validation) read that balance via `GetAvailableTokenBalanceUseCase`.

### Fix
Thread the node's bonded amount into `Route.Send` and override the RUNE account's balance in the UNBOND load path, so both consumers read the bonded amount from the selected account — a single point of correction.

- `Route.Send` — new `bondedAmount: String?` (base-unit).
- `ThorchainDefiPositionsViewModel` — cache the active bonded nodes; `onClickUnBond` resolves the selected node's amount by address and passes it.
- `SendFormGraph` — seed `bondedAmount` from args and expose it to the loader.
- `AccountsLoader` — dedicated `UNBOND` branch overrides the RUNE account's `tokenValue` with the bonded amount (recomputes fiat from price; falls back to the wallet balance if the amount is absent).

### Scope
Android only. Maya unbond uses a different form (`Route.Deposit`) and is untouched; all other DeFi load paths are unchanged.

## Test plan
- [x] `AccountsLoaderTest` — 18/18 pass, incl. 3 new UNBOND cases (balance override, null fallback, fiat recompute + non-RUNE pass-through).
- [x] `:app:compileDebugUnitTestKotlin` compiles clean.
- [ ] Manual: DeFi → Bonded → node → Unbond → MAX fills the node's full bond; percentages compute off the bond; unbond > bond is rejected.

## Summary by CodeRabbit

* **New Features**
  * Unbonding now pre-fills the send form using the selected THORChain bonded amount.
  * Send navigation now carries an optional bonded amount to drive unbond requests.
* **Bug Fixes**
  * RUNE token and fiat estimates are now recalculated from the bonded amount; non-RUNE balances remain unchanged, with a safe zero fallback when missing.
  * Bonded-node loading is more reliable by cancelling in-flight updates before refreshing.
* **UI/UX**
  * The bond/unbond destination address can be selectively locked to prevent retargeting.
* **Tests**
  * Expanded unbond scenarios coverage for token/fiat recalculation.


app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -251,4 +258,52 @@ internal class AccountsLoader(
             publishLoaded(emptyList(), generation)
         }
     }
+
+    // Unbond draws from the RUNE already bonded to the selected node, not the wallet's spendable
+    // RUNE. loadDeFiAddresses returns the wallet balance, so override the RUNE account's balance
+    // with the node's bonded amount (threaded via bondedAmountProvider). This makes the
+    // MAX/percentage base and the submit-time balance check operate on the bonded amount — matching
+    // iOS/Windows.
+    private suspend fun loadUnbondAccount(vaultId: VaultId, generation: Long) {
```
johnnyluo, line 268: @aminsato , I think this is incorrect , when user trying to unbond ,  user have to specify an amount that they would like to unbond , but when user click the percentage button , then we will need to send a request to THORChain to find out how much user can unbond.

For example , if user want to unbond node `thor15qcqwmwu002gfu4ynwee5nhqy62j268e3dmw0q`, the a request should be sent to https://gateway.liquify.com/chain/thorchain_api/thorchain/node/thor15qcqwmwu002gfu4ynwee5nhqy62j268e3dmw0q 

and then based on the bond address , you will be able to find out how much user have bonded , and then calculate how much user can unbond , 

When user click 100% , then you can let user to unbond the `"total_bond": "120971835768699"`

Keep in mind , bond provider can unbond as well

johnnyluo (dismissed):
can you please see the comment

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
aminsato: Thanks @johnnyluo — this is already what the PR does. The amount threaded into the unbond form isn't the wallet balance; it's the wallet's own bond on the selected node, fetched from the node API.

`onClickUnBond` reads `activeBondedNodes[node].amount`, and that value is computed in `ThorchainBondUseCase.calculateBondMetrics()`, which calls `getNodeDetails(nodeAddress)` (the `/thorchain/node/{address}` endpoint you linked) and pulls `myBond` from the `bond_providers` entry matching our bond address:

```kotlin
for (provider in bondProviders) {
    val providerBond = provider.bond.toBigIntegerOrNull() ?: BigInteger.ZERO
    if (provider.bondAddress == myBondAddress) {
        myBond = providerBond
    }
    totalBond += providerBond
}
```

So MAX / 100% fills `myBond`, which already handles the bond-provider case (a provider can only unbond their own bond, not `total_bond`). The `AccountsLoader` override just swaps the RUNE account's balance for that already-resolved per-node bond so the MAX/percentage base and submit-time check operate on it. No wallet-balance value is used in the unbond path.

Let me know if you'd prefer the value re-fetched at unbond time rather than reusing the amount already loaded on the Bonded tab — happy to add that if freshness is the concern.

realpaaao (commented):
Reviewed the fix by source trace + running the unit tests. **Looks correct on paper** — but flagging up front that I don't currently have a bonding position to exercise the manual on-chain path (DeFi → Bonded → node → Unbond → MAX), so that checkbox in the test plan stays unverified by me.

### What I traced (all holds)
- **Resolution:** `onClickUnBond` matches on `activeBondedNodes.firstOrNull { it.node.address == nodeAddress }`, and the button passes `node.fullAddress` (raw), which is exactly `BondedNodePosition.node.address` — so the lookup resolves, not a silent miss on the truncated display address.
- **Units:** `BondedNodePosition.amount` = `myBond` = `provider.bond.toBigInteger()` straight from the THORChain node-details API — base-unit RUNE (1e8), same scale as `TokenValue.value` / RUNE's 8 decimals. No scaling bug.
- **Single point of correction:** the `AccountsLoader.UNBOND` branch overrides only the RUNE account's `tokenValue` (+ recomputes fiat) and publishes to the shared `accountsState`. That same `selectedAccount` feeds **both** consumers — `AmountFractionManager` (MAX/percentage, zero gas subtraction, early-returns before any fee recompute) and `send/submit/UnbondStrategy` (submit check rejects `entered > bonded`). `GetAvailableTokenBalanceUseCase` reads `account.tokenValue` directly, so the override genuinely propagates to both.
- **Scope is clean:** Maya (`Route.Deposit`) untouched; the `else` branch still serves the other DeFi types.

### Tests
Ran `AccountsLoaderTest` on this branch: **18/18 pass, 0 failures**, including the 3 new UNBOND cases (balance override, null → wallet fallback, fiat recompute + non-RUNE pass-through).

### Minor, non-blocking
- Fiat goes null when `account.price` is null (cosmetic; amount stays correct — the documented fallback handles it).
- `activeBondedNodes` is a plain `var` written in the bonded-load collector and read in `onClickUnBond`; safe under main-dispatcher confinement (as the comment notes), and in the real flow the Bonded tab has already rendered before Unbond is tappable.

**Verdict:** LGTM on code + tests. Someone with an active bond should still run the manual MAX/percentage/over-bond check before merge to close the on-chain loop.

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -251,4 +258,52 @@ internal class AccountsLoader(
...
+
+    private fun publishUnbond(
+        accounts: List<Account>,
+        bondedAmount: BigInteger?,
+        generation: Long,
+    ) {
+        if (generation != currentGeneration) return
+        // Without a bonded amount (e.g. a stale deep link) fall back to the wallet balance rather
+        // than showing a zero balance that would block the form entirely.
+        if (bondedAmount == null) {
```
rkokhatskyi: The null-bondedAmount fallback claims to preserve the wallet balance, but `loadDeFiAddresses(vaultId,false)` never reaches that branch; `ThorchainDeFiBalanceService.kt:172-187` sums every bonded node's amount instead. A multi-node vault can then MAX or submit for this node up to that combined total, not its actual bond

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -211,6 +211,10 @@ constructor(
     private var lpDialogJob: Job? = null
     private var loadLpJob: Job? = null
 
+    // Latest bonded nodes, kept so onClickUnBond can resolve the selected node's raw bonded amount
+    // (the UI model only carries the formatted string). Read/written on the main dispatcher.
+    private var activeBondedNodes: List<BondedNodePosition> = emptyList()
```
rkokhatskyi, line 217: `activeBondedNodes` now backs a real balance ceiling, but `loadBondedNodes()` (called from `loadSavedPositions` and `onPositionSelectionDone`) has no Job tracking like `lpDialogJob`. Repeated refreshes stack concurrent collectors that overwrite it out of order, so `onClickUnBond` can resolve a stale bond

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1137,13 +1142,16 @@ constructor(
             val runeCoin = vault.coins.find { it.ticker == "RUNE" && it.chain == Chain.ThorChain }
 
             if (runeCoin != null) {
+                val bondedAmount =
```
rkokhatskyi, line 1161: `bondedAmount` is frozen to the tapped node at navigation time, but the destination address field in the Send form stays fully editable with no re-resolution. A user can retarget the address to a different node after MAX prefills this node's bond and still pass validation

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
aminsato: Good catch. Since \`loadDeFiAddresses\` returns the combined bond across every node, the old fallback let a multi-node vault MAX/submit against the aggregate. When the per-node amount is missing we can't derive the real ceiling, so it now zeroes the RUNE balance (blocks the form) rather than exposing another node's bond. Comment and \`AccountsLoaderTest\` updated. Fixed in e28e6d753.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Right — \`getActiveNodes\` never completes, so repeated refreshes stacked collectors that wrote \`activeBondedNodes\` out of order. Added a \`loadBondedNodesJob\` that's cancelled before each relaunch (same as \`lpDialogJob\`) and dropped the redundant \`launch { }\` wrappers. Fixed in e28e6d753.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Agreed. The unbond destination address is now locked: \`AddressInputSection\` gained an \`editable\` flag (disables the field and hides paste/scan/address-book) and \`SendFormScreen\` passes \`isAddressEditable = defiType != UNBOND\`, so it can no longer be retargeted after the ceiling is frozen. Bond/regular send stay editable. Fixed in e28e6d753.

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -251,4 +258,51 @@ internal class AccountsLoader(
             publishLoaded(emptyList(), generation)
         }
     }
+
+    // Unbond draws from the RUNE already bonded to the selected node, not the wallet's spendable
+    // RUNE. loadDeFiAddresses returns the wallet balance, so override the RUNE account's balance
```
rkokhatskyi: This comment still says loadDeFiAddresses returns the wallet balance, contradicting the corrected one in `publishUnbond` below (`ThorchainDeFiBalanceService.kt:174` sums every bonded node). Worth updating so a future edit doesn't trust the stale wording and reopen the fixed bug

rkokhatskyi (approved):

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -401,71 +406,77 @@ constructor(
...
                     return@launch
                 }
 
-                val address = runeCoin.address
+                state.update { it.copy(bonded = it.bonded.copy(isLoading = true)) }
+
+                // Load selected positions, if disabled then show nothing
+                try {
+                    val vault = withContext(Dispatchers.IO) { vaultRepository.get(vaultId) }
+                    val runeCoin = vault?.coins?.find { it.chain.id == Chain.ThorChain.id }
```
johnnyluo: THORChain can have many different coins , like RUNE , RUJI , TCY etc

aminsato:
@Vaulty-bot pick-up

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
johnnyluo: Fixed in 89ba76ae4 — corrected the `loadUnbondAccount` comment so it no longer claims `loadDeFiAddresses` returns the wallet balance; it now states it returns the RUNE account carrying the *combined* bond across every node, matching `publishUnbond` below.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
johnnyluo: Good point — fixed in 89ba76ae4. `runeCoin` was resolved via `find { it.chain.id == Chain.ThorChain.id }`, which picks the first THORChain coin (could be RUJI/TCY). Now matches the ticker explicitly (`it.ticker == "RUNE" && it.chain == Chain.ThorChain`) in `loadBondedNodes`, `loadStakingPositions`, and `loadLpPositions`, consistent with `onClickBond`/`onClickUnBond`.

PR #5204, opened 2026-07-06
fix(banner): restore Join-transaction banner slide-in for new requests

### LINKED ISSUE #5201: [Android] Join-transaction banner appears with no entrance animation (slide-in never plays for new requests)
## Summary
When a signing request comes through, the **"Join transaction"** foreground banner appears **abruptly with no entrance animation**, which is jarring. It should slide in smoothly from the top.

An enter animation (`slideInVertically { -it }`) is already defined in code, but it **never plays when a new request arrives** — see root cause below.

## Root cause (verified)
`MainActivityContent.kt` wraps the banner's `AnimatedVisibility` in `key(lastNotification?.qrCodeData)`:

```kotlin
key(lastNotification?.qrCodeData) {
    AnimatedVisibility(
        visible = foregroundNotification != null,
        enter = slideInVertically { -it },   // slide down from top — intended
        exit = slideOutVertically { -it },
        ...
    ) { ForegroundNotificationBanner(...) }
}
```

Two things combine to kill the enter animation:
1. When a **new** request arrives, `lastNotification.qrCodeData` changes → the `key(...)` **disposes and recreates** the `AnimatedVisibility` as a brand-new node.
2. `AnimatedVisibility(visible: Boolean, ...)` **does not run its enter transition on initial composition** when `visible` is already `true`. Because the re-keyed node is created with `visible == true` from the start, the `slideInVertically` enter is skipped → the banner just pops in.

So for the real-world path ("a fresh signing request comes in"), the enter animation is effectively dead. (Exit/slide-out and swipe-to-dismiss still work, since those happen while the node is alive.)

The `key(qrCodeData)` was added in `d019ae7a3d` (2026-03-22) and is in shipped builds (currently 1.0.112), so this has been silently missing for months.

## Suggested fix
Seed the transition so the enter animation plays even on the freshly-keyed node — use the `MutableTransitionState` overload:

```kotlin
key(lastNotification?.qrCodeData) {
    val visibleState = remember { MutableTransitionState(false) }
    visibleState.targetState = foregroundNotification != null
    AnimatedVisibility(
        visibleState = visibleState,
        enter = slideInVertically { -it } /* + fadeIn() for polish */,
        exit = slideOutVertically { -it },
        ...
    ) { ... }
}
```

This keeps the `key(qrCodeData)` (needed to reset swipe offset per the banner's `LaunchedEffect(isActive)`, ref #5001/#5134) while letting the initial slide-in actually run.

Alternative: don't key the `AnimatedVisibility` on `qrCodeData` (move the `key` inside the content around `ForegroundNotificationBanner`) so the node persists across requests and the `false→true` transition animates — but verify this doesn't reintroduce the swipe snap-back/re-center regressions from #5001/#5134.

Mirror the same treatment on the height-reserving `AnimatedVisibility` above it (`expandVertically()`) so the reserved space grows in step with the slide-in.

## Key file
- `app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt:180-195`
- Banner composable: `app/src/main/java/com/vultisig/wallet/ui/components/banners/ForegroundNotificationBanner.kt`

## Scope
Android. Foreground "Join transaction" keysign banner.

aminsato:
## Summary
The foreground **"Join transaction"** banner popped in with **no entrance animation** whenever a new signing request arrived — the intended `slideInVertically { -it }` never played. Fixes #5201.

## Root cause
`MainActivityContent.kt` wraps the overlay `AnimatedVisibility` in `key(lastNotification?.qrCodeData)`:
- A new request changes `qrCodeData` → the `key(...)` **disposes and recreates** the `AnimatedVisibility` as a fresh node.
- `AnimatedVisibility(visible: Boolean, ...)` **does not run its enter transition on initial composition** when `visible` is already `true`. The re-keyed node is born `visible == true`, so the `slideInVertically` enter is skipped and the banner just pops in.

Exit/slide-out and swipe-to-dismiss were unaffected (they run while the node is alive). Silently broken since `d019ae7a3d`.

## Fix
Seed both `AnimatedVisibility` blocks from a `MutableTransitionState(false)` and drive them via `targetState = foregroundNotification != null`, so the `false → true` transition actually plays on the freshly keyed node:
- **Overlay banner** — the slide-in now runs on every new request.
- **Height-reserving spacer** (`expandVertically`) — mirrored so the reserved space grows in step, including the cold-start case where a notification is already present on the first frame (same enter-skip trap).

The `key(qrCodeData)` is kept, so the per-request swipe-offset reset behavior from #5001 / #5134 is preserved.

## Test plan
- [ ] New signing request arrives → banner slides in from the top (no pop-in)
- [ ] A second request arriving while one is shown → new banner slides in, content stays pushed down
- [ ] Swipe-to-dismiss (horizontal + upward) still settles off-screen with no snap-back / re-center (#5001 / #5134 regression check)
- [ ] Exit/slide-out on tap-through and clear still plays

## Before / After
_Device before/after pending — animation capture on emulator to follow._

## Summary by CodeRabbit

* **Bug Fixes**
  * Foreground notification banners now animate in more reliably when they appear.
  * The space reserved under the status bar now expands and collapses in sync with the banner.
  * If a notification is already present when the banner appears, the slide-in transition now plays correctly.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
@@ -2,6 +2,7 @@ package com.vultisig.wallet.app.activity.components
 
 import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.ExperimentalAnimationApi
+import androidx.compose.animation.core.MutableTransitionState
 import androidx.compose.animation.expandVertically
 import androidx.compose.animation.shrinkVertically
 import androidx.compose.animation.slideInVertically
@@ -165,9 +166,13 @@ private fun MainActivityContent(
             OfflineBanner(isOffline)
             // Reserve the banner's height so content sits below it while it is shown. This
             // AnimatedVisibility mirrors the overlay's enter/exit so the reserved space grows and
-            // collapses in step with the banner sliding in and out.
+            // collapses in step with the banner sliding in and out. Seed the transition from a
+            // false MutableTransitionState so the expand runs even when a notification is already
+            // present on the first composition, keeping it in step with the overlay's slide-in.
+            val reserveState = remember { MutableTransitionState(false) }
+            reserveState.targetState = foregroundNotification != null
             AnimatedVisibility(
-                visible = foregroundNotification != null,
+                visibleState = reserveState,
                 enter = expandVertically(),
                 exit = shrinkVertically(),
             ) {
@@ -178,8 +183,14 @@ private fun MainActivityContent(
 
         // Banner drawn as overlay on top of navContent.
         key(lastNotification?.qrCodeData) {
+            // key(qrCodeData) disposes and recreates this node whenever a new request arrives, so
+            // it is born with visible == true and AnimatedVisibility(visible: Boolean) would skip
+            // its enter transition. Seeding a false MutableTransitionState and flipping targetState
+            // to true makes the slide-in run on the freshly keyed node.
+            val bannerState = remember { MutableTransitionState(false) }
+            bannerState.targetState = foregroundNotification != null
             AnimatedVisibility(
-                visible = foregroundNotification != null,
+                visibleState = bannerState,
                 enter = slideInVertically { -it },
                 exit = slideOutVertically { -it },
                 modifier =


johnnyluo (approved):
LGTM

PR #5206, opened 2026-07-06
fix(swap): show Track button on feeless swaps (decouple link from swapFee)

### LINKED ISSUE #5202: [Android] Swap done screen missing Track button (Solana/LiFi track link gated behind parseable swapFee)
## Summary
On the swap **"Transaction complete / pending"** done screen, the **Track** button is missing for (at least) same-chain **Solana** swaps (e.g. Bonk → SOL). The screen has a working tx-hash explorer link (↗), but the dedicated **Track** button that should point at the swap tracker doesn't appear.

## Reporter repro
Bonk → SOL swap → done screen shows "Transaction pending…", From/To cards, Transaction Hash, Transaction Details, **Done** — but no **Track** button.

## Root cause (verified in code)
The Track button renders only when the progress link is non-blank:
`app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SwapTransactionOverviewScreen.kt:168`
```kotlin
if (!progressLink.isNullOrBlank()) { VsButton(label = "Track", ...) }
```

`progressLink` = `swapProgressLink` from `ExplorerLinkRepository.getSwapProgressLink()` (built in `BroadcastKeysignUseCase.kt:160`). The Solana branch *intends* to return a Helius tracker but is gated behind `swapFee` parsing:
`data/src/main/kotlin/com/vultisig/wallet/data/repositories/ExplorerLinkRepository.kt`
```kotlin
is SwapPayload.EVM -> {
    if (payload.data.quote.tx.swapFee.toBigIntegerOrNull() != null) {
        if (fromCoin.chain == toCoin.chain && fromCoin.chain == Chain.Solana)
            "https://orb.helius.dev/tx/$tx"     // intended tracker for Solana swaps
        else "https://scan.li.fi/tx/$tx"
    } else null                                 // ← link silently dropped
}
```

`EVMSwapQuoteJson.swapFee` **defaults to `""`** (`data/src/main/kotlin/com/vultisig/wallet/data/api/models/quotes/EVMSwapQuoteJson.kt:39`), and `"".toBigIntegerOrNull() == null`. So any swap whose quote omits a swap fee (feeless routes) hits the `else null` branch → `swapProgressLink == null` → **Track button suppressed**, despite a valid tracker URL existing for that swap.

**Caveat:** the exact runtime `swapFee` for this Bonk→SOL quote isn't confirmed (data-dependent), but the field defaults to empty and the tracker link is explicitly gated on it parsing — that's the mechanism. Not a stuck tx; the ↗ tx-hash link still opens the explorer.

## Suggested fix
Decouple the tracker URL from `swapFee`. Build the Solana (Helius) / LiFi track link based on the src chain + tx hash (which are always available for a broadcast swap); `swapFee` has no bearing on whether a tracker exists. Optionally, if there's genuinely no dedicated tracker for a route, fall back to the source-chain explorer link so the Track button is always present for swaps.

## Key files
- `app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SwapTransactionOverviewScreen.kt:168` — Track button visibility
- `data/src/main/kotlin/com/vultisig/wallet/data/repositories/ExplorerLinkRepository.kt` — `getSwapProgressLink()` swapFee gate
- `data/src/main/kotlin/com/vultisig/wallet/data/api/models/quotes/EVMSwapQuoteJson.kt:39` — `swapFee: String = ""` default
- `app/src/main/java/com/vultisig/wallet/data/usecases/BroadcastKeysignUseCase.kt:160` — where `swapProgressLink` is built

## Scope
Android. Same-chain Solana swaps confirmed by code path; the same `swapFee`-gate also affects the LiFi (`scan.li.fi`) cross-chain track link whenever a quote's `swapFee` is empty — worth checking those too.

aminsato:
## Summary
Fixes #5202. On the swap **"Transaction pending / complete"** done screen, the **Track** button was missing for feeless swaps — notably same-chain **Solana** swaps (e.g. Bonk → SOL).

The Track button renders only when `getSwapProgressLink()` returns a non-blank URL:
```kotlin
if (!progressLink.isNullOrBlank()) { VsButton(label = "Track", ...) }
```
For the `SwapPayload.EVM` branch the tracker URL was gated behind the affiliate `swapFee` parsing as a `BigInteger`. But `EVMSwapQuoteJson.swapFee` defaults to `""`, and `"".toBigIntegerOrNull() == null`, so any quote that omits a swap fee hit the `else null` branch → the valid tracker URL was silently dropped and the Track button disappeared.

## Fix
Drop the `swapFee` gate in the EVM branch. The tracker exists independently of the affiliate fee, so the link is now derived purely from the source chain + tx hash (both always present for a broadcast swap):
- same-chain Solana → `https://orb.helius.dev/tx/<hash>`
- otherwise → `https://scan.li.fi/tx/<hash>`

SwapKit-routed swaps are unaffected — they return earlier from the `track.swapkit.dev` branch before reaching this code.

## Test plan
- [x] `ExplorerLinkRepositoryImplTest`: updated the case that asserted the old buggy `null`; added regression coverage for same-chain Solana (Helius) and non-SwapKit EVM (LiFi) with empty `swapFee`.
- [ ] Device before/after: the done screen with vs without the Track button (pending a build environment with the Trust Wallet Core PAT + emulator).

> Note: unit tests could not be executed in the authoring environment (`androidx.work:work-testing` not fetchable / no GitHub Packages PAT). All cases were verified by tracing the logic; please confirm green in CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Swap progress tracking links are now shown more reliably for swaps even when fee details are unavailable.
  * Same-chain Solana swaps now open the correct tracking link.
  * EVM swap routes now consistently resolve to a transaction tracker instead of appearing missing.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/ExplorerLinkRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/ExplorerLinkRepository.kt
@@ -49,16 +49,15 @@ internal class ExplorerLinkRepositoryImpl @Inject constructor() : ExplorerLinkRe
             is SwapPayload.MayaChain ->
                 "https://www.explorer.mayachain.info/tx/${tx.removePrefix("0x")}"
             is SwapPayload.EVM -> {
-                if (payload.data.quote.tx.swapFee.toBigIntegerOrNull() != null) {
-                    if (
-                        payload.data.fromCoin.chain == payload.data.toCoin.chain &&
-                            payload.data.fromCoin.chain == Chain.Solana
-                    ) {
-                        "https://orb.helius.dev/tx/${tx}"
-                    } else "https://scan.li.fi/tx/${tx}"
-                } else {
-                    null
-                }
+                // The tracker exists independently of the affiliate `swapFee`; gating on it dropped
+                // the Track button for feeless routes (e.g. same-chain Solana Bonk -> SOL). Derive
+                // the link from src chain + tx hash, both always present for a broadcast swap.
+                if (
+                    payload.data.fromCoin.chain == payload.data.toCoin.chain &&
+                        payload.data.fromCoin.chain == Chain.Solana
+                ) {
+                    "https://orb.helius.dev/tx/${tx}"
+                } else "https://scan.li.fi/tx/${tx}"
             }
 
             else -> null
--- a/data/src/test/kotlin/com/vultisig/wallet/data/repositories/ExplorerLinkRepositoryImplTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/repositories/ExplorerLinkRepositoryImplTest.kt
@@ -105,9 +105,10 @@ class ExplorerLinkRepositoryImplTest {
     }
 
     @Test
-    fun `SwapKit-routed EVM on untrackable source chain falls back to source-chain path`() {
+    fun `SwapKit-routed EVM on untrackable source chain falls back to LiFi tracker`() {
         // Hyperliquid is EVM-shaped but outside SwapKit's /track catalogue, so the SwapKit branch
-        // is skipped and the existing EVM source-chain logic applies (empty swapFee -> null).
+        // is skipped and the EVM source-chain logic applies. The tracker no longer depends on
+        // swapFee, so a feeless route still gets a Track link.
         val link =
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

Issue #5186, opened 2026-07-06
fix(ripple): reject hex/scientific notation in XRP amount parser


realpaaao:
## What
Android XRP amount parsing uses `BigDecimal` which **accepts hex/scientific notation** (`RippleHelpter.kt:48`), unlike the SDK which guards against it. A malformed/scientific amount can yield a wrong signed amount.

## Fix
Reject hex/scientific notation in the XRP amount parser (mirror the SDK's regex guard); accept only plain decimal.

## Acceptance
- [ ] Hex/scientific inputs rejected before signing
- [ ] Plain-decimal amounts unaffected
- [ ] Unit test covering hex/scientific rejection

Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

## References
- XRP currency/amount formats (drops = 6 decimals, no hex/scientific): https://xrpl.org/docs/references/protocol/data-types/currency-formats

PR #5207, opened 2026-07-06
fix(ripple): stop returning the error string as the tx hash on broadcast failure

### LINKED ISSUE #5185: fix(ripple): stop returning the error string as the tx hash on broadcast failure (port SDK result-code handling)
## What
On XRP broadcast, Android returns the **error message string as the transaction hash** on non-recovered failures (`RippleApi.kt:63`) — producing an untrackable/garbage hash and masking the real failure. It only recovers hashes on "already applied/passed/redundant" strings; other tec/tef/tem codes are mishandled.

## Fix
Port the SDK's ripple broadcast classification: gate on `tesSUCCESS`; on `tefALREADY`/`tefPAST_SEQ` verify-by-hash; on any other failure return a **typed error**, never the error string as a hash.

## Acceptance
- [ ] Never returns an error string in place of a tx hash
- [ ] tec/tef/tem codes classified; real reason surfaced
- [ ] tefALREADY / already-applied → verify-by-hash, no double-broadcast
- [ ] Unit test on the failure paths

Ref: XRPL reliable submission https://xrpl.org/docs/concepts/transactions/reliable-transaction-submission · Part of the XRP chain-hardening set (tracker: vultisig-sdk#979).

aminsato:
## Summary

Fixes #5185. On an XRPL `submit` rejection the app previously returned the engine **error message as the transaction hash**, so the keysign flow persisted a garbage txid instead of surfacing the failure. This ports the iOS `RippleService` engine-result classification into Android.

- Extracts a pure, testable `resolveBroadcastHash(result)` (top-level, `internal`).
  - Returns the echoed `tx_json.hash` **only** for `tesSUCCESS` or a benign duplicate-broadcast race (`tefPAST_SEQ` / `tefALREADY`, or an already-applied/redundant message) so on-chain recovery can verify by hash without re-broadcasting.
  - Every other engine result (`tem*`, `tec*`, other `tef*`) throws a typed `RippleBroadcastException` carrying the engine code + message — never returns the message as a fake hash. Message format mirrors iOS: `Ripple broadcast failed (code): message`.
- `broadcastTransaction` performs the HTTP call inside a `try` (rewrapping non-broadcast errors), then classifies the result outside it. Genuine rejections are still logged with `Timber.e` at the API layer.

## Test plan

- New `ResolveBroadcastHashTest` (9 classification cases): `tesSUCCESS`/duplicate return the echoed hash; `tem*`/`tec*`/other `tef*` throw the typed exception with the correct `engineResult`/`engineResultMessage`; `tec*`-with-hash still throws; blank-hash success throws instead of inventing a hash; code-only fallback message.
- Updated `RippleApiBodyReadTest` rejection case to assert `RippleBroadcastException` and its fields.
- `:data:compileDebugKotlin` and ktfmt pass locally. (Unit tests run on CI — the local env cannot resolve `androidx.work:work-testing`.)

## Notes
Ports iOS `RippleService.broadcastTransaction`. Part of the XRP hardening set (#5184, #5186); related to #5170.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction submission handling so duplicate or already-applied broadcasts are treated more reliably.
  * Rejection failures now return clearer, structured error details instead of a generic failure.
  * Prevented blank or invented transaction hashes from being returned on unsuccessful submissions.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
@@ -1,6 +1,7 @@
 package com.vultisig.wallet.data.api
 
 import RippleBroadcastResponseResponseJson
+import RippleBroadcastResponseResponseResultJson
 import RippleBroadcastSuccessResponseJson
 import com.vultisig.wallet.data.api.models.RpcPayload
 import com.vultisig.wallet.data.models.Coin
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5214, opened 2026-07-07
fix(send): reject hex/scientific-notation amounts before signing (#5186)

### LINKED ISSUE #5186: fix(ripple): reject hex/scientific notation in XRP amount parser

aminsato:
Closes #5186

Transaction hash
https://xrpscan.com/tx/D7C5B13CB977CF154F6CF2FCDF6C3D326CD8308BE7E1AEE569EAFDE3D7F7D2FB

## Problem
Send amount parsing uses `String.toBigDecimalOrNull()`, which accepts **scientific notation** (`1e5` → `100000`). The numeric amount keyboard can't type `e`/`x`, and `TokenAmountInput`'s `InputTransformation` strips non-decimal characters from typed and pasted input — but that filter only runs on IME/paste edits.

A `vultisig://send?...&amount=1e5` deeplink writes the amount into the field **programmatically** (`SendFormGraph` → `setTextAndPlaceCursorAtEnd`), bypassing the filter. `DeepLinkHelper.getAmount()` returns the raw query param with no validation, so a crafted link could feed a malformed/scientific value straight into the signed transaction. This is the same exposure fixed in `vultisig-sdk#973` (which routed parsing through the SDK's decimal-to-base-unit converter for all chains).

Hex (`0x1F`) already fails `toBigDecimalOrNull`; scientific notation was the live gap.

## Fix
Add a single plain-decimal contract and apply it at every amount-parse choke point:

- `String.isPlainDecimal()` / `String.toPlainBigDecimalOrNull()` in `SendAmountExtensions.kt` — digits and at most one decimal point; rejects hex, scientific, signs, separators.
- `AmountManager.validateTokenAmount()` — gates the UI error state and Continue button.
- All send/DeFi submit strategies that re-parse the field before building the signed amount: `DefaultSendStrategy`, `BondStrategy`, `UnbondStrategy`, `StakeStrategy`, `UnstakeStrategy`, `MintStrategy`, `RedeemStrategy`, `WithdrawUsdcCircleStrategy`, and the shared `AccountValidator`.

Plain decimals (including `.5`, `5.`, trailing zeros) are unaffected.

## Test plan
- [x] `SendAmountExtensionsTest` — plain accepted; `1e5` / `1E+2` / `2.5e3` / `0x1F` / `-5` / `+5` / `1,5` / `1.2.3` rejected.
- [x] `AmountManagerTest` — new cases assert scientific/hex/signed inputs return the invalid-amount error; existing valid-amount and precision tests still pass (22/22).
- [x] `./gradlew :app:compileDebugKotlin` and `:app:testDebugUnitTest --tests "*ui.models.send.*"` green.
- [ ] Device: `adb shell am start -a android.intent.action.VIEW -d "vultisig://send?assetChain=Ripple&assetTicker=XRP&toAddress=r...&amount=1e5"` — amount rejected, Continue disabled; `&amount=10.5` still works.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved send amount parsing to use plain-decimal validation, making token/fiat inputs more consistent across normal entry and other input paths.
  * Now rejects unsupported number formats such as scientific notation, hex notation, and signed values during amount validation.
* **Tests**
  * Added/expanded unit tests for plain-decimal parsing utilities and for rejecting invalid formats during send amount validation.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountManager.kt
@@ -108,7 +108,7 @@ internal class AmountManager(
     }
 
     private suspend fun handleTokenInput(token: Coin, tokenString: String) {
-        val tokenDecimal = tokenString.toBigDecimalOrNull()
+        val tokenDecimal = tokenString.toPlainBigDecimalOrNull()
         _isMaxAmount.value = tokenDecimal?.compareTo(maxAmount) == 0 && maxAmount > BigDecimal.ZERO
 
         val fiatValue =
@@ -151,7 +151,7 @@ internal class AmountManager(
             return
         }
 
-        val tokenDecimal = tokenValue.toBigDecimalOrNull()
+        val tokenDecimal = tokenValue.toPlainBigDecimalOrNull()
         _isMaxAmount.value = tokenDecimal?.compareTo(maxAmount) == 0 && maxAmount > BigDecimal.ZERO
 
         lastTokenValueUserInput = tokenValue
@@ -164,7 +164,7 @@ internal class AmountManager(
         token: Coin,
         transform: (value: BigDecimal, price: BigDecimal, token: Coin) -> BigDecimal,
     ): String? {
-        val decimalValue = value.toBigDecimalOrNull() ?: return ""
+        val decimalValue = value.toPlainBigDecimalOrNull() ?: return ""
 
         val price =
             try {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountManager.kt
@@ -77,7 +77,7 @@ internal class AmountManager(
         if (value.length > TextFieldUtils.AMOUNT_MAX_LENGTH) {
             return UiText.StringResource(R.string.send_from_invalid_amount)
         }
-        val decimal = value.toBigDecimalOrNull()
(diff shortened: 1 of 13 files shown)


aminsato:
Thanks @coderabbitai — addressed both nitpicks:

- **AmountManager conversion paths**: applied. `handleTokenInput`, `handleFiatInput`, and `convertValue` now use `toPlainBigDecimalOrNull()` so the fiat/isMax preview matches the plain-decimal contract enforced at submit (commit 6bfa951).
- **Duplicate parse via `ValidatedAccount`**: valid, but deferring to a separate refactor. It's a pre-existing pattern across all 8 strategies, and folding it in here would leave several with unused `BigDecimal.ZERO` imports (CI lint aborts on that) and broaden a focused security fix. Tracking as follow-up.

johnnyluo (approved):
LGTM

PR #5215, opened 2026-07-07
fix(ripple): stop double-counting XRP reserve in reaping warning (#5208)

### LINKED ISSUE #5208: fix(ripple): XRP reaping warning double-counts the reserve (spurious warning, always fires on MAX)
## What

The XRP reaping warning double-counts the account reserve, so it fires spuriously — and **always fires on a MAX send** — even when the send is valid on-chain.

`RippleApiImp.getBalance` already returns the **owner-aware, reserve-net** balance:

```kotlin
// data/.../api/RippleApi.kt:106-111
val balance = accountInfo?.getBalance() ?: BigInteger.ZERO
val ownerCount = accountInfo?.getOwnerCount() ?: BigInteger.ZERO
val accountReservedBalance =
    reservedBalance.getBaseReserve() + (ownerCount * reservedBalance.getIncReserve())
maxOf(balance - accountReservedBalance, BigInteger.ZERO)   // reserve already subtracted
```

That reserve-net value flows into `getAvailableTokenBalance` → the MAX/percentage amount, so the MAX **amount** is correct.

But `ChainValidationService.checkIsReapable` then subtracts a **second, static 1 XRP** on top of that already-reserve-net balance:

```kotlin
// app/.../send/ChainValidationService.kt:183-190
selectedChain == Chain.Ripple && selectedToken.ticker == Coins.Ripple.XRP.ticker ->
    RippleHelper.DEFAULT_EXISTENTIAL_DEPOSIT.toBigInteger()   // = 1_000_000 drops = 1 XRP
...
if (balance - (gasFee.value + tokenAmountInt) >= existentialDeposit) return null   // warns below +1 XRP
```

Because `balance` here is already net of the true on-chain reserve, comparing the post-send remainder against a second static 1 XRP double-counts. The warning fires across the whole band `[true reserve, true reserve + 1 XRP]`, and on a **MAX send** the remainder is 0, so `0 >= 1 XRP` is false and the reaping warning **always** shows — even though a MAX send that lands exactly at the reserve is valid (the fee is reserve-exempt).

## Severity

The reaping check drives `AmountManager.reapingError`, which renders as a **display-only warning banner** (`SendFormScreen.ReapingErrorMessage`) — it does **not** block the send. So this is a confusing false-positive warning, not fund loss. Lower severity than the iOS sibling, but a real correctness/UX bug and a parity gap.

## Contrast with iOS (#4744 / PR #4748)

iOS had the same static-1-XRP-on-top-of-reserve-net double-count, but there it also **stranded 1 XRP on MAX** and **hard-blocked** valid sends. iOS fix (draft PR #4748): unify on the owner-aware reserve as the single source of truth, zero the second static existential deposit for XRP, and remove the XRP reap block. Android's `getBalance` is already the owner-aware source, so Android only needs the reaping/existential-deposit side corrected.

## Suggested fix

For XRP, the remainder-below-existential-deposit reaping test is redundant with the reserve already netted out in `getBalance`. Either:
- drop `Chain.Ripple` from the static `DEFAULT_EXISTENTIAL_DEPOSIT` reaping path (Polkadot genuinely needs its ED and should stay), or
- compare the XRP post-send remainder against `0` (the available balance already excludes the true reserve), so the warning only fires when a send would actually exceed the spendable balance.

Keep the reserve source owner-aware and live from `server_state` (already the case in `getBalance`); do not reintroduce a hardcoded reserve.

## Acceptance

- [ ] A MAX XRP send from an account with `OwnerCount > 0` does **not** show a spurious reaping warning
- [ ] The reaping warning for XRP reflects the true owner-aware on-chain reserve, not a second static 1 XRP
- [ ] Polkadot reaping behavior unchanged
- [ ] Unit test in `ChainValidationServiceTest` covering XRP with `OwnerCount > 0`

Part of the XRP chain-hardening set (tracker: vultisig-sdk#979). Sibling of iOS #4744 (draft PR vultisig-ios#4748). Related Android XRP hardening: #5184, #5185, #5186.

aminsato:
## Summary

Fixes #5208. The XRP reaping warning double-counted the account reserve, so it fired spuriously — and **always on a MAX send** — even when the send was valid on-chain.

`RippleApi.getBalance` already returns the **owner-aware, reserve-net** balance (`baseReserve + OwnerCount * incReserve`, both live from `server_state`), so the account `tokenValue` reaching `ChainValidationService.checkIsReapable` is already net of the true on-chain reserve.

`checkIsReapable` then subtracted a **second, static 1 XRP** existential deposit on top of it. The warning fired across the whole `[reserve, reserve + 1 XRP]` band, and on a MAX send (remainder ≈ 0) the `0 >= 1 XRP` test always failed → the banner always showed.

The warning is display-only (`SendFormScreen.ReapingErrorMessage`) — it never blocked the send, so this is a false-positive UX bug / iOS parity gap, not fund loss.

## Changes

- **`ChainValidationService.checkIsReapable`** — drop XRP from the static existential-deposit reaping path so it no longer produces a warning. XRP's reserve is already netted out upstream in `getBalance`, so the check is redundant.
- **Polkadot unchanged** — its `tokenValue` is the raw free balance and the ED is only netted into the MAX amount (`GetAvailableTokenBalanceUseCase`), so subtracting the ED once in `checkIsReapable` is the single, correct subtraction there.
- Remove the now-unused `send_form_ripple_reaping_warning` string from **all 10** locale files.
- Add unit tests covering the XRP MAX send and the reserve-band case.

## Why not "compare against 0"

The alternative (compare the XRP remainder against `0`) leaves a dead code path — since `getBalance` already prevents the spendable amount from dropping the account below reserve, the XRP reaping check can never legitimately trigger. Removing it entirely is the clean fix.

## Test plan

- `./gradlew :app:testDebugUnitTest --tests ChainValidationServiceTest` → green
  - `checkIsReapable - xrp MAX send with owner reserves does not warn` ✅
  - `checkIsReapable - xrp remainder within one XRP band does not warn` ✅
  - Polkadot below-ED / sufficient-balance tests unchanged and passing ✅
- ktfmt applied.

## Acceptance criteria

- [x] MAX XRP send with `OwnerCount > 0` shows no spurious reaping warning
- [x] XRP reaping reflects the true owner-aware reserve, not a second static 1 XRP
- [x] Polkadot reaping behavior unchanged
- [x] Unit test in `ChainValidationServiceTest` covering XRP with `OwnerCount > 0`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved XRP send validation so the app no longer shows an incorrect “reaping” warning for normal XRP transactions.
  * MAX send flows now correctly account for XRP reserve rules, reducing false warnings when sending the maximum available balance.
  * Updated localization strings to match the refined XRP sending behavior across supported languages.


rkokhatskyi (approved):
LGTM

johnnyluo (approved):
LGTM

Issue #4762, opened 2026-06-06
Cosmos staking (LUNA/LUNC): rename 'Next Award' → 'Claimable rewards'


realpaaao:
# [Fix] Rename Cosmos staking 'Next Award' → 'Claimable rewards' [LUNA/LUNC Staking Position Card]

## Problem
On the LUNA/LUNC staking position card, the 🏆 row is labeled **"Next Award"**, but the value shown is the **currently-accrued, claimable** reward (`position.pendingReward`) — withdrawable right now via the existing **Claim** button. Cosmos-SDK staking rewards accrue continuously (there is no scheduled "next" award), so the label is misleading. It should read **"Claimable rewards"**.

## Expected Behavior
The 🏆 row on the Cosmos (LUNA/LUNC) staking position card displays the label **"Claimable rewards"** next to the accrued `position.pendingReward` value, accurately reflecting that the amount is immediately withdrawable via the Claim button.

## Steps to Reproduce
1. Open the app and navigate to the Cosmos (LUNA/LUNC) staking screen.
2. Locate a staking position with a non-zero pending reward.
3. Observe the 🏆 row is labeled "Next Award" even though the value is the currently-claimable reward and a Claim button is available.

### Must Do
- [ ] Change the value of string `cosmos_staking_next_award` from "Next Award" to "Claimable rewards" (`app/src/main/res/values/strings.xml:1327`).
- [ ] Verify the updated label renders correctly in `CosmosStakingPositionsScreen.kt:480` (label) alongside the value at `:486` (`formatStakeAmount(position.pendingReward)`).
- [ ] Check iOS `ValidatorCard`/staking position for the same Cosmos label and flag/rename if the identical issue exists there.

### Must NOT Do
- Don't change unrelated code.
- Do **NOT** rename the generic `next_award` (id 49) used by THORChain **TCY / RUJI** staking — those have an actual scheduled payout ("Next Payout" date + "Next Award"), so "Next Award" is correct there.
- Don't change the string **id** (`cosmos_staking_next_award`) or any references to it — only the displayed string value.

## Acceptance Criteria
- [ ] `./gradlew assembleDebug` succeeds.
- [ ] The Cosmos (LUNA/LUNC) staking position card 🏆 row displays "Claimable rewards" instead of "Next Award".
- [ ] THORChain TCY/RUJI staking still displays "Next Award" (generic `next_award` id 49 unchanged).

## Context
[...]

PR #5218, opened 2026-07-07
fix(terraClassic): price send fee at the effective gas limit so Verify matches signed fee

johnnyluo:
On chain tx link: 
Initiate from android , join from android
https://finder.terra.money/classic/tx/8C04F35CD4C51866FB1E8073A2B35E6656C8661976A7269A34E59698EC614F5E
Initiate from MacOS , join from android
https://finder.terra.money/classic/tx/5935ABBAAC9933436AD5F0C711E83BF9F983958E0B1D135A0E088347018CDA71

## Summary
Ports iOS PR [vultisig/vultisig-ios#4762](https://github.com/vultisig/vultisig-ios/pull/4762) to Android.

When sending on **Terra Classic**, the Send **Verify** screen read `chainSpecific.gas` (priced at a fixed limit) while the signer re-derived the amount at signing time via `TerraClassicTax.scaledSendFee` — so the **displayed** fee could diverge from the **signed** fee once a dynamic `gas_wanted` was honored.

Android was already halfway there: `TerraClassicTax.baseGas(contractAddress, isNativeToken, gasLimit)` already prices the base at the effective gas limit up front, so `chainSpecific.gas` is the final fee. This PR removes the now-redundant re-derivation:

- **`TerraClassicTax`**: drop `scaledSendFee(...)` and the unused `scaled(...)` helper; `baseGas` is the single source of the fee.
- **`TerraHelper`**: use `atomData.gas` **verbatim** as the fee amount (no re-scaling), still honoring the relayed dynamic `gas_wanted` via `effectiveGasLimit`. Removed the now-dead `isTerraClassic` constructor param.
- **`SigningHelper`**: drop the `isTerraClassic` argument at both call sites.
- **Tests**: removed the `scaledSendFee`/`scaled` unit tests (the `baseGas` tests already cover per-limit pricing); updated `ChainHelpersTest` call sites + comment.

Result: the fee shown on Verify / Send Details / keysign-confirm now equals the signed fee.

## Parity / safety
- At the static 300k limit the base is **unscaled**, so the non-simulated path is **byte-identical** and cross-device MPC parity is unchanged.
- The relayed `gas_limit` write side is still off on Android (honor-only, per #5112/#5191), and Terra Classic isn't simulated, so today `gas` is priced at 300k and `effectiveGasLimit` resolves to 300k — no behavior change.
- When an iOS initiator (post-#4762) relays `gas` already priced at a dynamic limit plus `gas_limit`, Android now echoes it verbatim instead of double-scaling it from the static base — the correct cross-device behavior.
- Plain Terra (phoenix-1) pays a flat fee (never repriced) — unchanged.
- dYdX has the same latent display gap (`CosmosGasPricedFee` is the shared infra for that); left as a separate follow-up, matching iOS scoping.

## Test Plan
- [x] `:data:testDebugUnitTest --tests TerraClassicTaxTest` passes
- [x] `:app:compileDebugAndroidTestKotlin` compiles (ChainHelpersTest)
- [ ] Instrumented `ChainHelpersTest` (`sendTerra`, `terraClassicRelayedGasLimitHonoredInSignedBytes`) on a device/emulator
- [ ] Manual: send LUNC on Terra Classic; Verify fee matches the broadcast/signed fee

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Terra and Terra Classic (LUNC) signing so fees consistently use the fully priced gas fee amount while honoring the effective (possibly relayed) gas limit.
  * Updated relayed gas limit handling to better match expected signed transaction bytes.
  * Adjusted join-keysign fee logic for Cosmos so network fees are taken directly from the provided gas value, with the fee-service fallback behavior updated accordingly.
* **Tests**
  * Updated Terra Classic and Cosmos fee test coverage and added a Terra Classic (LUNC) fixture to match the new fee semantics.


app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignFeeUtils.kt
```diff
@@ -39,6 +39,14 @@ internal fun computeJoinKeysignNetworkFee(
         // would understate a memo send on the co-signing device's approval screen.
         is BlockChainSpecific.Cardano ->
             TokenValue(value = blockChainSpecific.byteFee.toBigInteger(), token = nativeCoin)
+        // Cosmos signs `chainSpecific.gas` verbatim as the fee amount (see CosmosHelper /
+        // TerraHelper), so the joiner must surface that exact value — the initiator already priced
+        // it at the effective gas limit. Recomputing it locally via the fee service reprices Terra
+        // Classic's base at the static 300k limit and re-derives its burn tax from the joiner's own
+        // amount, so the co-signer's Verify screen diverges from the signed fee (issue: 8.5525
+        // LUNC).
+        is BlockChainSpecific.Cosmos ->
```
aminsato, line 50: The class KDoc up top (line 18) still reads *"Ethereum returns…, THORChain returns…, and every other chain returns `fallbackFeeAmount`."* That's now stale for this Cosmos branch and the existing Cardano one — both read straight from `BlockChainSpecific` and never touch the fallback. Worth updating the summary so the next reader isn't misled into thinking Cosmos still goes through the fee service.

The branch itself is correct: `CosmosHelper.buildCosmosFee` sets `fee.amount = atomData.gas` verbatim for every Cosmos chain, so surfacing `blockChainSpecific.gas` here always matches the signed bytes (same rationale as Cardano). For non-TerraClassic Cosmos this also replaces a recomputed fee-service estimate that could drift from the initiator's signed fee — a real improvement, not just the Terra Classic fix.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TerraHelper.kt
```diff
@@ -96,7 +71,13 @@ class TerraHelper(
...
             PublicKey(keysignPayload.coin.hexPublicKey.hexToByteArray(), PublicKeyType.SECP256K1)
         val effectiveGasLimit = effectiveGasLimit(atomData)
-        val feeAmount = feeAmount(keysignPayload, atomData, effectiveGasLimit)
+        // `atomData.gas` already carries the final fee AMOUNT priced for the effective gas limit:
+        // the initiator pins the Terra Classic base to that limit up front
+        // (`TerraClassicTax.baseGas`
+        // in CosmosFeeService), and plain Terra pays a flat fee. Either way the signer echoes it
+        // verbatim, so the signed fee equals the fee shown on Verify, while still honoring the
+        // relayed dynamic `gas_wanted`. Mirrors vultisig-ios `TerraHelperStruct`.
+        val feeAmount = atomData.gas.toString()
```
rkokhatskyi, line 80: `atomData.gas` now signs verbatim, but iOS's live `TerraHelperStruct` (its #4762 fix is unmerged) rescales via `scaledSendFee` since `chainSpecific.gas` stays priced at the static 300k limit while relaying a dynamic `gasLimit` (tx 5935ABBA signed 9332883 for gas_limit 327551, not 8552500), so co-signers hash a different fee and the signature never forms.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignFeeUtils.kt
rkokhatskyi, line 50: this same verbatim-echo assumption is about to break a third platform: vultisig-windows is pinned to `@vultisig/core-mpc@1.8.17` (no `gasLimit` field at all), but its queued `1.8.19` bump adds `resolveCosmosGasFee`, which rescales from the static base exactly like the `scaledSendFee` this PR deletes, so an Android and a post-bump Windows co-signer will hash different fees too.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignFeeResolver.kt
```diff
@@ -88,7 +88,8 @@ constructor(
         if (
             blockChainSpecific is BlockChainSpecific.Ethereum ||
                 blockChainSpecific is BlockChainSpecific.THORChain ||
-                blockChainSpecific is BlockChainSpecific.Cardano
+                blockChainSpecific is BlockChainSpecific.Cardano ||
+                blockChainSpecific is BlockChainSpecific.Cosmos
```
rkokhatskyi, line 92: for a Cosmos dApp fee entirely in a non-native denom, `dappSuppliedNativeFee` returns null (per its own tests) and this Cosmos arm falls back to displaying `blockChainSpecific.gas`, but `CosmosHelper.buildAminoSigningInput` builds the signed fee purely from `signAmino.fee`, never `cosmosSpecific.gas`, so the co-signer approves a fee/denom unrelated to what actually gets signed.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignFeeUtils.kt
johnnyluo: Good catch — updated the class KDoc in e39e182 so the summary now reflects that Cardano returns the transmitted `byteFee` and Cosmos returns the initiator-priced `blockChainSpecific.gas`, both read straight from `BlockChainSpecific` (only the remaining chains fall back to `fallbackFeeAmount`). Thanks for confirming the branch rationale.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TerraHelper.kt
johnnyluo: This is safe on the currently-deployed paths. The relayed dynamic gas-limit **write side is still off** on Android (honor-only, per #5112/#5191) and Terra Classic is not simulated, so `chainSpecific.gas` is always priced at the static 300k limit and `effectiveGasLimit` resolves to 300k. At the 300k base `TerraClassicTax.baseGas` is unscaled, so echoing `atomData.gas` verbatim produces **byte-identical** SignDoc bytes to the old `scaledSendFee`-at-300k path — cross-device MPC parity is unchanged.

The divergence you describe only occurs if an initiator relays `gas` priced at a *dynamic* limit ≠ 300k. That is exactly the scenario the tx in the PR body exercises: `5935ABBA` was **initiated from macOS, joined from this Android build**, signed 9332883 for gas_limit 327551, and broadcast successfully on-chain — the signature *did* form, which it couldn't if the two devices had hashed different fees. When a post-#4762 iOS initiator relays `gas` already priced at the dynamic limit, echoing it verbatim is the correct behavior; double-scaling it from the static base (the old code) is what would diverge.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignFeeUtils.kt
johnnyluo: Same reasoning as the TerraHelper thread. The joiner surfaces the fee the initiator actually signs — for Cosmos that is `chainSpecific.gas` verbatim (`CosmosHelper` / `TerraHelper` set `fee.amount = atomData.gas`). Whichever platform initiates, the joiner echoes the transmitted `gas`; it does not independently rescale, so an Android joiner and a Windows (1.8.19 `resolveCosmosGasFee`) or iOS (#4762 `scaledSendFee`) *initiator* still hash the same fee because the joiner reads the initiator's own value rather than recomputing it. The only device that prices the base is the **initiator**, and on the deployed Android paths that price is at the static 300k limit (write-side off), so it matches the unscaled base. If a future dynamic-write-side lands, that pricing lives in `CosmosFeeService` and stays the single cross-device source.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignFeeResolver.kt
johnnyluo: You're right about the mechanics — `buildAminoSigningInput` builds the signed fee purely from `signAmino.fee`, and a fee entirely in a non-native denom makes `dappSuppliedNativeFee` return null so this arm shows `blockChainSpecific.gas`. But this is a **pre-existing** display gap, not something this PR introduces: before this change the Cosmos arm fell through to `else -> fallbackFeeAmount` (the fee-service estimate), which is *also* a native-denom value unrelated to a non-native amino fee. So both the old and new code display a native-denom number for that edge case; neither surfaces the actual non-native signed fee.

Surfacing a multi-denom / non-native dApp fee correctly on Verify is a broader change (the fee row is single-value, native-denom today) and is out of scope for this Terra Classic fix. Leaving this thread open and tracking it as a separate follow-up rather than expanding this PR.

rkokhatskyi (approved):

aminsato:
LGTM

PR #2331, opened 2025-08-12
Update function pairing screen

aminsato:
## Description

Please include a summary of the change and which issue is fixed. 

Fixes #2315

## Which feature is affected?
- [ ] Create vault ( Secure / Fast) - Please ensure you created a Secure vault & fast vault
- [ ] Sending  - Please attach a tx link here
- [ ] Swap - Please attach a tx link for the swap here
- [ ] New Chain / Chain related feature  -  Please attach tx link here

## Checklist

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

## Summary by CodeRabbit

* **New Features**
  * Redesigned Verify Deposit screen: card-based layout, optional top app bar, chain logo + network row.
  * Hold-to-confirm paired-sign flow with fallback to standard confirm button.

* **Improvements**
  * Deposit UI model fields renamed to Source/Destination; destination hidden when empty.
  * Network fees now shown in token and fiat with live currency-aware conversion; amounts unified across views.

* **Localization**
  * Added verify-deposit strings for DE, ES, HR, IT, NL, PT, RU, EN.

PR #5221, opened 2026-07-07
fix(deposit): remove duplicate Amount row on Function overview

### LINKED ISSUE #5216: Function overview: redundant "Amount" row duplicates the "You're sending" amount
The **Function overview** (transaction verify) screen shows the send amount twice.

The amount already appears prominently in the header row (`1 RUNE` / `$0.41`) at the top of the summary card. Further down the card, a separate **Amount** row repeats the same value (`1`), which is redundant and adds visual noise.

The **Amount** row should be removed from this screen since the amount is already displayed in the "You're sending" header.

## Screenshot

overview

## Steps to reproduce
1. Start a BOND (deposit/function) transaction on THORChain
2. Proceed to the Function overview screen
3. Observe the amount is shown both in the header ("You're sending") and in the "Amount" row

## Expected
Amount is shown once (in the "You're sending" header). The duplicate "Amount" row is not displayed.

## Actual
Amount appears twice — in the header and in a separate "Amount" row.

aminsato:
## Description

Fixes #5216

The **Function overview** (Verify Deposit) screen showed the send amount twice:

- The amount already appears in the **"You're sending"** header (`SwapToken`) at the top of the summary card (`1 RUNE` / fiat).
- A separate **Amount** row lower in the card repeated the same `tx.token.value`, adding redundant visual noise.

This removes the duplicate row so the amount is shown once, in the header.

### Changes
- Remove the conditional `Amount` `VerifyCardDetails` block (and its leading divider) from `VerifyDepositScreen.kt`.
- Remove the now-unused `java.math.BigDecimal` import.

### Scope / safety
- The shared `verify_transaction_amount_title` string is **kept** — still used by `TransactionDoneScreen` and `SignTonDisplayView`.
- `SendTxOverviewScreen` is intentionally **not** touched — it has no "You're sending" header, so its Amount row is not redundant.
- Purely presentational removal; no ViewModel/data/logic changes.

## Which feature is affected?
- [x] New Chain / Chain related feature — THORChain BOND/deposit Function overview (verify) screen

## Screenshots

_Before/after device render pending (build env)._ The amount now renders only in the "You're sending" header.

## Additional context

Note: the duplicate row was introduced by #2425, which added it after the "You're sending" header (added earlier in #2331) already displayed the amount — so it was redundant from the point it landed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified the deposit verification screen so transaction details no longer rely on token amount parsing; the divider now appears only when source/destination address or memo content is present.
  * Removed the previously shown “deposit amount” row from the send transaction overview for native value transfers, so the UI no longer displays that amount section before the unlimited approval area.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt
@@ -55,7 +55,6 @@ import com.vultisig.wallet.ui.screens.swap.VerifyCardDetails
 import com.vultisig.wallet.ui.screens.swap.VerifyCardDivider
 import com.vultisig.wallet.ui.theme.Theme
 import com.vultisig.wallet.ui.utils.asString
-import java.math.BigDecimal
 
 @Composable
 internal fun VerifyDepositScreen(
@@ -227,21 +226,6 @@ internal fun VerifyDepositScreen(
                             title = stringResource(R.string.verify_transaction_memo_title),
                             subtitle = tx.memo,
                             showAllContent = true,
-                        )
-                    }
-
-                    if (
-                        tx.token.value.isNotEmpty() &&
-                            try {
-                                tx.token.value.toBigDecimal() > BigDecimal.ZERO
-                            } catch (e: Exception) {
-                                false
-                            }
-                    ) {
-                        VerifyCardDivider(0.dp)
-                        VerifyCardDetails(
-                            title = stringResource(R.string.verify_transaction_amount_title),
-                            subtitle = (tx.token.value),
                         )
                     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt
```diff
@@ -230,21 +229,6 @@ internal fun VerifyDepositScreen(
                         )
                     }
 
-                    if (
-                        tx.token.value.isNotEmpty() &&
-                            try {
-                                tx.token.value.toBigDecimal() > BigDecimal.ZERO
```
rkokhatskyi, line 236: `SendTxOverviewScreen` has the same duplicate amount display this PR fixes here, its `VsOverviewToken` header at `SendTxOverviewScreen.kt:108-118` already shows the value while the Amount row at `SendTxOverviewScreen.kt:180` repeats it whenever `functionName` is null, true for every Deposit and plain Send

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt
aminsato: Good catch — verified and fixed in 62d1d3be7. You're right that `VsOverviewToken` (SendTxOverviewScreen.kt:108-118) already renders `value + ticker`, so the Amount row at :180 was the same duplicate for every Deposit and plain Send. Removed the row and the now-unused `BigDecimal` import. (The PR's earlier commit message claiming this screen had no header was wrong — it does.)

rkokhatskyi (approved):
LGTM

PR #5228, opened 2026-07-08
fix(solana): use finalized commitment for Fast Vault pre-ceremony blockhash refresh

### LINKED ISSUE #5225: [Fix] Solana pre-signing blockhash refresh doesn't use finalized commitment
# [Fix] Solana pre-signing blockhash refresh doesn't use finalized commitment

## Problem
The only blockhash-fetch method Solana signing uses hardcodes "confirmed" commitment, including the refresh that runs immediately before starting a keysign ceremony. A confirmed blockhash can be unknown to a load-balanced broadcast node at signing time (producing a preflight "blockhash not found" error seconds after fetching it), whereas a finalized blockhash is guaranteed to be known everywhere, with still enough of its validity window left for a normal ceremony.

## Expected Behavior
The pre-ceremony blockhash refresh (right before signing starts) should fetch a blockhash at "finalized" commitment; other callers that just need a quote/fee estimate can keep using "confirmed".

## Steps to Reproduce
1. Start a Solana keysign (send/swap), especially via a Fast Vault where a server round-trip widens the gap between blockhash fetch and broadcast.
2. Observe an occasional blockhash-expired/not-found failure that the existing rebroadcast-retry loop doesn't always absorb.

## Solution
Add a "finalized" commitment variant of the blockhash-fetch call and use it specifically for the pre-ceremony refresh, keeping "confirmed" for the initial quote/fee-estimate path.

### Must Do
- [ ] Add a finalized-commitment blockhash-fetch method
- [ ] Use it in the pre-ceremony blockhash refresh, not the general confirmed-commitment method
- [ ] Leave the quote/fee-estimate blockhash fetch on confirmed commitment

### Must NOT Do
- Don't change the existing rebroadcast/retry loop
- Don't change blockhash handling for dApp-supplied raw transactions

## Acceptance Criteria
- [ ] New/updated tests confirm the pre-ceremony refresh requests finalized commitment
- [ ] Existing fee-estimate tests still pass on confirmed commitment

rkokhatskyi:
## Summary

The Solana blockhash refresh that runs right before a keysign ceremony starts used `confirmed` commitment for every vault type. A `confirmed` blockhash can still be unknown to the load-balanced RPC node that ends up handling the broadcast, which occasionally surfaced as a "blockhash not found" failure shortly after signing — most visible on Fast Vault, where the extra device-server round trip widens the gap between the blockhash fetch and the broadcast.

`SolanaApi` gets a `getFinalizedBlockHash()` method alongside the existing `getRecentBlockHash()` (confirmed). `UpdateSolanaKeysignPayloadUseCase` now branches on vault type: Fast Vault (server auto-co-signs, ceremony is always short) uses `finalized`, since it's guaranteed known to every RPC node at the cost of ~13s of the blockhash's ~60-90s validity window — a cost a Fast Vault ceremony never gets close to spending. Secure Vault (waits on a human to approve on another device, duration unpredictable) keeps `confirmed` to preserve the fuller validity window. The quote/fee-estimate call sites (`BlockChainSpecificRepository`, `SolanaFeeService`) are untouched.

Closes #5225.

## Testing
- `./gradlew assembleDebug`
- `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest`
- `./gradlew :app:lintDebug :data:lintDebug`

## Summary by CodeRabbit

* **New Features**
  * Solana signing now selects the appropriate blockhash based on vault type (finalized vs confirmed), improving transaction validity.
  * Added support for fetching a finalized blockhash alongside the existing recent blockhash.
* **Bug Fixes**
  * Solana keysign payload updates are now derived using the selected vault before generating signing messages.
* **Tests**
  * Expanded coverage for blockhash commitment request/response details.
  * Added tests for Solana vs non-Solana payloads and null payload handling.


rkokhatskyi:
Fixed the nitpick on `fetchBlockHash`'s log/error messages (e879039a8) — they now name the actual RPC method and commitment level instead of hardcoding "recent blockhash", and use Timber format specifiers per house style.

johnnyluo:
Currently , when we broadcast transaction , we already using preflightCommitment: Confirmed , so it is standard to fetch the recentBlockHash using commitment: Confirmed as well 

I don't think this PR is needed

rkokhatskyi:
> Currently , when we broadcast transaction , we already using preflightCommitment: Confirmed , so it is standard to fetch the recentBlockHash using commitment: Confirmed as well
> 
> I don't think this PR is needed


Agree, not worth doing (a bit of over-engineering)

PR #5229, opened 2026-07-08
fix(solana): throw on blockhash RPC error instead of returning empty string

### LINKED ISSUE #5226: [Fix] Solana blockhash fetch silently returns an empty string on RPC error
# [Fix] Solana blockhash fetch silently returns an empty string on RPC error

## Problem
When the blockhash RPC call returns a JSON-RPC error, the code logs it and returns an empty string instead of throwing, unlike the adjacent case where a malformed/missing result correctly throws. The empty string then flows unchecked into the transaction payload and eventually into signing, surfacing later as an opaque WalletCore error instead of a clear "network error, please retry" message — after potentially wasting an entire multi-device keysign round-trip.

## Expected Behavior
An RPC error while fetching the blockhash should throw a clear, retryable error before peer discovery/signing starts, the same way a malformed response already does.

## Steps to Reproduce
1. Trigger a Solana send/swap while the blockhash RPC call returns a JSON-RPC error response (e.g. simulate a transient node error).
2. Observe the flow continues with an empty blockhash instead of failing immediately, eventually surfacing a confusing low-level signing error later.

## Solution
Make the blockhash fetch throw on a JSON-RPC error response the same way it already throws on a missing/malformed result, so the failure surfaces immediately and clearly instead of propagating an empty string downstream.

### Must Do
- [ ] Throw instead of returning an empty string when the blockhash RPC call returns an error
- [ ] Add a unit test asserting an RPC error response throws rather than returning ""

### Must NOT Do
- Don't change the handling of a malformed/missing result (already throws correctly)
- Don't add retry logic here — just fail clearly

## Acceptance Criteria
- [ ] A simulated RPC error surfaces as a clear failure before signing starts, not an empty blockhash

rkokhatskyi:
## Description

`getRecentBlockHash` logged a JSON-RPC error response and returned `""` instead of throwing, unlike the adjacent malformed/missing-result branch a few lines below it, which already throws correctly. The empty string then flowed unchecked into the keysign payload and only surfaced later as an opaque WalletCore error during signing, after potentially wasting a multi-device keysign round-trip.

Now the RPC-error branch throws immediately, matching the existing malformed-result handling in the same function and the RPC-error idiom used elsewhere in the api package (`DashApi`, `CardanoApi`).

Fixes #5226

## Which feature is affected?
- [x] Sending - no tx link, this is a pre-signing failure-mode fix (no on-chain behavior change on the happy path)
- [ ] Swap

## Checklist

N/A - backend-only change, no UI surface affected.

## Additional context

Sibling PR #5228 (issue #5225, still open) also touches `SolanaApi.kt` (adds a new `getFinalizedBlockHash()` method) - non-conflicting overlap, just a heads up for whoever rebases second.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for recent Solana block hash requests so RPC error responses now surface as exceptions with clearer details instead of returning an empty value.
  * Added automated coverage to verify the thrown exception type and that the error message includes the transient node error text.


johnnyluo (approved):
LGTM

PR #5230, opened 2026-07-08
fix(cosmos): coalesce per-token bank-balance requests (#5161)

### LINKED ISSUE #5161: Cosmos balance endpoint called once per token on Select asset sheet
## Description

Opening the **Select asset** sheet for the Cosmos chain (swap flow) fires the same balance request many times within a couple of seconds — one identical HTTP call per token listed on the sheet (~10 for Cosmos: ATOM, KUJI, NAMI, USDC, rKUJI, FUZN, LVN, NSTK, USK, WINK):

```
GET https://cosmos-rest.publicnode.com/cosmos/bank/v1beta1/balances/<address>
```

This endpoint returns **all** denominations for the address in a single response, so every call after the first is redundant.

## Steps to reproduce

1. Open a vault with the Cosmos chain enabled (with several Cosmos tokens added)
2. Start a swap and tap the token field to open **Select asset** with Cosmos selected
3. Observe network traffic

**Actual:** N identical `GET /cosmos/bank/v1beta1/balances/{address}` requests fire back-to-back (one per token).

**Expected:** a single request per address; the response already contains every denom needed to populate all token rows.

## Where it happens

- `AccountsRepository.emitRefreshAddress` fans out per token: `address.accounts.map { balanceRepository.getTokenBalanceAndPrice(...) }` — all Cosmos tokens share the same address.
- `BalanceRepository.getTokenValue` (Cosmos branch: `GaiaChain, Kujira, Dydx, Osmosis, Terra, Noble, Akash`) calls `cosmosApi.getBalance(address)` for **each** token, fetches the full denom list, picks one denom with `.find { it.hasValidDenom(coin) }`, and discards the rest.
- `CosmosApiImp.getBalance` issues the GET with no caching, request deduplication, or in-flight sharing.

Other balance paths already avoid this: EVM chains are batched through Multicall3 (`getEvmTokenBalancesAndPrices`), and the DeFi balance path uses a `SimpleCache` plus a per-key mutex for in-flight sharing. The plain Cosmos bank-balance path has neither.

## Impact

- Wasted bandwidth and battery; slower balance population (requests run sequentially)
- Unnecessary load on the public RPC endpoint, risking rate limiting for users with many tokens

Applies to all Cosmos-SDK chains that go through this branch (Cosmos Hub, Kujira, dYdX, Osmosis, Terra, Noble, Akash).

aminsato:
## Summary

Opening the swap **Select asset** sheet for a Cosmos chain fired ~N identical
`GET /cosmos/bank/v1beta1/balances/{address}` requests — one per token (~10 for Cosmos: ATOM, KUJI, NAMI, USDC, …). The bank endpoint returns **every** denom for an address in a single response, so every call after the first was pure duplication.

Root cause: `BalanceRepository.getTokenValue` creates a **fresh `CosmosApiImp` per token** and calls `getBalance` on each, so no per-instance state could dedupe them.

## Fix

Add a shared `CosmosBalanceCache` in `CosmosApi.kt`:

- `ConcurrentHashMap` store + **per-key `Mutex`** (concurrent in-flight sharing) + **short TTL (10s)**, keyed by `"$rpcEndpoint|$address"`.
- The `@Singleton` `CosmosApiFactoryImp` holds **one** shared cache and injects it into every `CosmosApiImp`, so the per-token fan-out for an address collapses to a single round-trip — whether the calls arrive concurrently (mutex shares the in-flight fetch) or sequentially (TTL serves the first result).
- Mirrors the existing `SimpleCache` + per-key-mutex pattern already used by the DeFi balance path, but stores in a `ConcurrentHashMap` so reads across different keys stay thread-safe.
- The constructor param defaults to a fresh cache, so existing callers and tests are unaffected.

**Why it's safe:** `getBalance` is used only for balance display and token discovery (`CosmosBankCoinFinder`) — never in the signing/fee path (which uses `simulate` / `getAccountNumber`). On a fetch error nothing is cached, so it fails open and the next caller retries. A refresh past the TTL window still hits the network.

Applies to all Cosmos-SDK chains through this branch (Cosmos Hub, Kujira, dYdX, Osmosis, Terra, Noble, Akash, QBTC).

## Test plan

Added 4 unit tests to `CosmosApiTest.kt` (all green):

- [x] 10 sequential `getBalance` calls for one address → **1** HTTP request
- [x] 10 concurrent callers → **1** HTTP request, all receive the balance (in-flight sharing)
- [x] two different addresses → **2** requests (cache keyed by address)
- [x] `ttlMs = 0` → **2** requests on repeat (proves TTL gating; no stale masking)

`./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.CosmosApiTest"` passes; ktfmt applied.

Closes #5161

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced duplicate balance requests by reusing recent Cosmos balance responses for a short time.
  * Improved handling of simultaneous balance lookups so multiple callers share the same in-flight request.
  * Ensured balance reads are kept separate per address, with fresh data fetched after cache expiry.

* **Tests**
  * Added coverage for sequential caching, concurrent request sharing, address-based separation, and cache expiration behavior.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/CosmosApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/CosmosApi.kt
@@ -38,8 +38,11 @@ import io.ktor.http.HttpStatusCode
 import io.ktor.http.contentType
 import io.ktor.util.encodeBase64
 import java.net.URLEncoder
+import java.util.concurrent.ConcurrentHashMap
 import javax.inject.Inject
 import kotlin.coroutines.cancellation.CancellationException
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.buildJsonObject
@@ -126,6 +129,11 @@ constructor(
     private val cosmosThorChainResponseSerializer: CosmosThorChainResponseSerializer,
     private val customRpcRepository: CustomRpcRepository,
 ) : CosmosApiFactory {
+    // Shared across every CosmosApiImp this @Singleton factory hands out so the per-token
+    // bank-balance requests for one address collapse to a single round-trip. See
+    // CosmosBalanceCache.
+    private val balanceCache = CosmosBalanceCache()
+
     override fun createCosmosApi(chain: Chain): CosmosApi {
         val defaultApiUrl = CustomRpcDefaultEndpoint.cosmosUrl(chain)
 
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5233, opened 2026-07-08
fix(solana): gate tx-status err check to finalized commitment only

### LINKED ISSUE #5231: [Fix] Solana tx-status polling treats err as terminal Failed at any confirmation level
# [Fix] Solana tx-status polling treats err as terminal Failed at any confirmation level, not gated to finalized like iOS

## Problem
`SolanaStatusProvider.checkStatus` treats a non-null `err` on the signature status as an immediately terminal `Failed` result regardless of confirmation level (processed/confirmed/finalized). Roughly 5% of Solana blocks at processed/confirmed level never finalize (they can be forked out and re-included with a different outcome), so a transient `err` observed at that early, unsettled level can produce a permanently wrong "Transaction Failed" verdict for a transaction that goes on to actually confirm. Once emitted this is terminal — the polling loop stops with no re-check. The code's own comment claims this "matches iOS," but it does not: iOS gates the `err` check strictly inside the `finalized` case.

## Expected Behavior
A non-null `err` should only be treated as terminal Failed once the status has reached `finalized` commitment. At `processed`/`confirmed`, the transaction should report Pending (matching iOS) even if `err` is currently set, since that snapshot can still change before finality.

## Steps to Reproduce
1. Broadcast a Solana transaction and poll its status shortly after submission (processed/confirmed level).
2. Have the RPC report a transient `err` at that early level that later clears once the block finalizes.
3. Observe the app immediately shows a terminal "Transaction Failed" and stops polling, even though the transaction later actually confirms.

## Solution
Nest the `err`-based Failed check inside the `finalized` branch of the confirmation-status `when`, mirroring iOS's `SolanaTransactionStatusProvider.swift`. At `confirmed`/`processed`/`null`, return Pending regardless of `err`.

### Must Do
- [ ] Gate the `err != null` → Failed check to `confirmationStatus == "finalized"` only
- [ ] Update/add tests: finalized+err → Failed, finalized without err → Confirmed, confirmed+err → Pending, processed+err → Pending, processed without err → Pending, null → Pending
- [ ] Correct the code comment that currently claims this matches iOS

### Must NOT Do
- Don't change the retry/broadcast logic in SolanaApi.kt
- Don't touch the unrelated `demoteForReorg` mechanism in TransactionHistoryRepository.kt (dead code, out of scope)

## Acceptance Criteria
- [ ] Updated tests confirm `err` at processed/confirmed returns Pending, and only finalized+err returns Failed
- [ ] Existing finalized-without-err → Confirmed behavior still passes

rkokhatskyi:
## Summary

`SolanaStatusProvider.checkStatus` treated a non-null `err` on the signature status as an immediately terminal `Failed` result regardless of confirmation level. Roughly 5% of Solana blocks at processed/confirmed level never finalize (they can be forked out and re-included with a different outcome), so a transient `err` observed at that early, unsettled level could permanently mis-report a transaction that goes on to actually confirm — the polling loop treats `Failed` as terminal and never re-checks. The code's own comment claimed this matched iOS; it did not, since iOS nests its `err` check strictly inside the `finalized` case.

`err` is now only consulted once `confirmationStatus == "finalized"`; `confirmed`/`processed`/`null` all report `Pending` regardless of `err`, matching iOS.

Closes #5231.

## Testing
- `./gradlew assembleDebug`
- `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest`
- `./gradlew :app:lintDebug :data:lintDebug`

## Summary by CodeRabbit

* **Bug Fixes**
  * Transaction status handling now treats non-final Solana confirmations as pending, even when an error is present.
  * Only finalized transactions with an error are marked as failed; unrecognized statuses are reported as not found.
* **Tests**
  * Updated coverage to reflect the new status behavior for processed, confirmed, and unrecognized transaction states.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/SolanaStatusProvider.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/SolanaStatusProvider.kt
@@ -14,19 +14,20 @@ internal class SolanaStatusProvider @Inject constructor(private val solanaApi: S
     override suspend fun checkStatus(txHash: String, chain: Chain): TransactionResult =
         try {
             val status = solanaApi.checkStatus(txHash)?.result?.value?.firstOrNull()
-            when {
+            when (status?.confirmationStatus) {
                 // A landed tx with `err` set executed but reverted (swap slippage, insufficient
                 // funds, etc.). Report failure instead of treating a finalized-but-failed tx as
-                // confirmed (matches iOS).
-                status?.err != null -> TransactionResult.Failed(status.err.toString())
-                else ->
-                    when (status?.confirmationStatus) {
-                        "finalized" -> TransactionResult.Confirmed
-                        "confirmed",
-                        "processed",
-                        null -> TransactionResult.Pending
-                        else -> TransactionResult.NotFound
-                    }
+                // confirmed (matches iOS). Only finalized is terminal: roughly 5% of blocks at
+                // processed/confirmed never finalize, so an `err` seen at those earlier levels
+                // can still change before finality and must not be reported as a permanent
+                // Failed (matches iOS, which nests its own err check inside `finalized` too).
+                "finalized" ->
+                    if (status.err != null) TransactionResult.Failed(status.err.toString())
+                    else TransactionResult.Confirmed
+                "confirmed",
+                "processed",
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5240, opened 2026-07-08
fix(keysign): sha256-hash cosmos custom messages to match iOS/Windows

ahdzib-maya:
## Problem

`SigningHelper.getKeysignMessages(CustomMessagePayload)` keccak256's the message for every non-EdDSA chain. But cosmos-family chains (THORChain, Maya, Cosmos, …) sign the **sha256** of the message — Keplr ADR-36 `signArbitrary` over the `StdSignDoc` bytes — so the Android digest, and the `md5(message)` relay message-ID derived from it, diverges from:

- **iOS** — `CustomMessagePayload.keysignMessages`
- **Windows / extension** — `getCustomMessageHex.ts` (`match(getChainKind, { evm: keccak256, cosmos: sha256, … })`)
- the CLI initiator

As a result an Android device **cannot co-sign a cosmos custom message** with any of those platforms: it fetches the relay setup/round messages under a different message-ID than the initiator wrote, so the keysign rounds 404 and the session times out.

## Fix

Route the digest by chain kind, mirroring `getCustomMessageHex.ts`:

| chain kind | digest |
|---|---|
| cosmos-family (`COSMOS`, `THORCHAIN`) | sha256 |
| EdDSA (Solana, TON, …) | raw (precomputed digest) |
| everything else (EVM, Tron) | keccak256 |

Adds `ByteArray.toSha256ByteArray()`. Behavior for EVM / EdDSA / typed-data is unchanged; only cosmos-family custom messages change (keccak256 → sha256), bringing Android to parity with iOS and Windows.

## Testing

Added unit tests in `SigningHelperCustomMessageTest` (cosmos → sha256, and cosmos ≠ keccak256). Not built locally (no Android SDK in this environment) — relying on CI for the full build + test run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated message signing so Cosmos-family and THORChain messages use the correct hashing method.
  * Improved handling for EdDSA-based chains to preserve their expected signing bytes.
  * Keccak-256 remains the default for other supported chains.

* **Tests**
  * Added coverage for Cosmos and THORChain signing behavior to verify the new hashing rules.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
@@ -5,12 +5,14 @@ package com.vultisig.wallet.data.chains.helpers
 import com.vultisig.wallet.data.api.swapAggregators.OneInchSwap
 import com.vultisig.wallet.data.common.toHexBytes
 import com.vultisig.wallet.data.common.toKeccak256ByteArray
+import com.vultisig.wallet.data.common.toSha256ByteArray
 import com.vultisig.wallet.data.crypto.SuiHelper
 import com.vultisig.wallet.data.crypto.ThorChainHelper
 import com.vultisig.wallet.data.crypto.TonHelper
 import com.vultisig.wallet.data.models.Chain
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5243, opened 2026-07-09
fix(send): hide empty Memo row on Send overview

### LINKED ISSUE #5234: Send overview shows empty Memo row when memo is blank
## Description

On the **Send overview** (transaction confirmation) screen, the **Memo** row is always rendered — even when the transaction has no memo. This results in an empty label with a blank value taking up vertical space in the summary card.

The Memo row should only be shown when a memo value is actually present. When the memo is empty, the row (label + divider) should be hidden entirely.

## Steps to Reproduce

1. Start a send with no memo (e.g. a plain THORChain / RUNE transaction without a memo).
2. Proceed to the **Send overview** screen.
3. Observe the **Memo** row rendered with an empty value.

## Expected Behavior

- If the memo is empty/blank, the Memo row is not displayed at all.

## Actual Behavior

- The Memo row is displayed with an empty value, leaving a blank field in the summary.

## Screenshot

Send overview empty memo

## Environment

- Screen: Send overview / transaction confirmation
- Example chain: THORChain

johnnyluo:
## Summary

On the **Send overview** (transaction confirmation) screen, the **Memo** row was rendered whenever `tx.memo` was non-null. Because an empty memo is an empty string (`""`) rather than `null`, the row showed up as a blank **Memo** label with no value, wasting vertical space in the summary card.

This changes the guard from `tx.memo != null` to `!tx.memo.isNullOrBlank()`, so the Memo row (and its divider) only appears when a memo is actually present.

Fixes #5234

## Changes

- `VerifySendScreen.kt` — guard the Memo row with `isNullOrBlank()`.
- `PreviewActivity.kt` — add a `verify_send_empty_memo` preview covering the empty-memo THORChain send case.

## Before / After

| Before | After |
|--------|-------|
| before | after |

In *Before*, notice the empty **Memo** row between **To** and **Network**. In *After*, the row is gone and **To** flows directly into **Network**.

## Test Plan

- [x] Rendered `VerifySendScreen` with an empty memo via PreviewActivity — empty Memo row no longer appears.
- [x] Existing previews with a real memo still render the Memo row (unchanged behavior for non-blank memos).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved send confirmation screens so empty or whitespace-only memos no longer appear as memo details.
  * Added a new preview for the empty-memo send flow to help verify the updated display.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -280,6 +280,7 @@ class PreviewActivity : ComponentActivity() {
                     "vault_detail_mldsa" -> VaultDetailPreview(withMldsa = true)
                     "error_screen_after" -> ErrorScreenAfterPreview()
                     "chain_selection" -> ChainSelectionClipPreview()
+                    "verify_send_empty_memo" -> VerifySendEmptyMemoPreview()
                     else -> SwapConfirmPreview()
                 }
             }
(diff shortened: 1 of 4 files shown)


rkokhatskyi (approved):
LGTM

PR #5244, opened 2026-07-09
fix(solana): fall back to default on malformed keysign compute_limit

### LINKED ISSUE #5236: [Fix] Solana inbound proto mapper crashes on malformed compute_limit but silently zeros malformed priority_fee
# [Fix] Solana inbound proto mapper crashes on malformed compute_limit but silently zeros malformed priority_fee

## Problem
`KeysignPayloadProtoMapper.kt`'s inbound Solana mapping parses the two priority-fee fields inconsistently: `priorityFee` uses `it.priorityFee.toBigIntegerOrNull() ?: BigInteger.ZERO` (fails soft), while `computeLimit` uses a throwing `it.computeLimit?.toBigInteger() ?: SOLANA_PRIORITY_FEE_LIMIT.toBigInteger()` — the `?.` only guards a null/absent field, `toBigInteger()` itself throws uncaught on a malformed non-numeric string. This is inbound data crossing a peer/system boundary during a TSS keysign ceremony, so a malicious or buggy co-signing device sending a garbled `compute_limit` string crashes the receiving app instead of gracefully defaulting, unlike the sibling field.

## Expected Behavior
Both fields should use the same fail-soft pattern (fall back to a sane default) so malformed inbound data from a peer never crashes the receiving app during a keysign ceremony.

## Steps to Reproduce
1. Have a peer device send a `KeysignPayloadProto` with a non-numeric `compute_limit` string during a Solana keysign session join.
2. Observe the receiving Android app throws an uncaught `NumberFormatException` instead of defaulting.

## Solution
Change `computeLimit` parsing to `toBigIntegerOrNull() ?: SOLANA_PRIORITY_FEE_LIMIT.toBigInteger()`, matching `priorityFee`'s existing fail-soft pattern.

### Must Do
- [ ] Make `computeLimit` parsing fail soft (`toBigIntegerOrNull() ?: default`) instead of throwing on malformed input
- [ ] Add a unit test asserting a malformed `compute_limit` string produces the default value instead of throwing

### Must NOT Do
- Don't change `priorityFee`'s existing fail-soft behavior
- Don't change `SolanaHelper`'s own floor/clamp logic for priorityFee

## Acceptance Criteria
- [ ] A malformed `compute_limit` from a peer no longer crashes the app

rkokhatskyi:
Closes #5236

## Problem
The inbound keysign payload mapper parsed the Solana `compute_limit` field with a throwing `toBigInteger()`. The `?.` only guards a null field — `toBigInteger()` itself throws on a non-numeric string. Since this is untrusted data deserialized from a peer at the start of a keysign ceremony, a buggy or malicious co-signer sending a garbled `compute_limit` crashed the receiving device with an uncaught `NumberFormatException` — while the sibling `priority_fee` field on the same message already fails soft.

## Fix
Parse `compute_limit` with `toBigIntegerOrNull()`, matching `priority_fee`. Malformed or empty input now falls back to the existing `SOLANA_PRIORITY_FEE_LIMIT` default instead of throwing.

| `compute_limit` on the wire | Before | After |
|---|---|---|
| absent | default | default |
| `"0"` | `0` | `0` |
| `"5000"` | `5000` | `5000` |
| `"abc"` / `""` | **crash** | default |

Only unparseable input changes behavior. Every valid value — including an explicit `0` — still maps through verbatim, so this does not reintroduce the payload override rejected in #5178.

## Testing
- `./gradlew :data:testDebugUnitTest` — passes, including the new `KeysignPayloadProtoMapperSolanaComputeLimitTest` (malformed / empty / absent / well-formed). The malformed and empty cases throw `NumberFormatException` without the fix.
- `./gradlew :app:assembleDebug` — passes.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/mappers/KeysignPayloadProtoMapper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/mappers/KeysignPayloadProtoMapper.kt
@@ -208,7 +208,7 @@ internal class KeysignPayloadProtoMapperImpl @Inject constructor() : KeysignPayl
                                 toAddressPubKey = it.toTokenAssociatedAddress,
                                 programId = it.programId == true,
                                 priorityLimit =
-                                    it.computeLimit?.toBigInteger()
+                                    it.computeLimit?.toBigIntegerOrNull()
                                         ?: SOLANA_PRIORITY_FEE_LIMIT.toBigInteger(),
                             )
                         }
--- a/data/src/test/kotlin/com/vultisig/wallet/data/mappers/KeysignPayloadProtoMapperSolanaComputeLimitTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/mappers/KeysignPayloadProtoMapperSolanaComputeLimitTest.kt
@@ -1,0 +1,96 @@
+package com.vultisig.wallet.data.mappers
+
+import com.vultisig.wallet.data.chains.helpers.SOLANA_PRIORITY_FEE_LIMIT
+import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.models.payload.BlockChainSpecific
+import com.vultisig.wallet.data.models.payload.KeysignPayload
+import java.math.BigInteger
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5245, opened 2026-07-09
fix(solana): confirm SPL token account by derivation on indexer lag

### LINKED ISSUE #5224: [Fix] Solana token-account existence check has no fallback when the indexer lags
# [Fix] Solana token-account existence check has no fallback when the indexer lags

## Problem
Checking whether a wallet's or recipient's associated token account (ATA) exists makes a single `getTokenAccountsByOwner` RPC call; if it returns empty (e.g. due to indexer/replica lag right after the account was created) or throws, the code immediately concludes the account doesn't exist with no further verification. This produces a false "missing token account" error for the sender, or causes the app to attempt to (re)create an ATA that already exists for the recipient — which the on-chain program rejects, since account creation isn't idempotent.

## Expected Behavior
When the primary lookup comes back empty or fails, the app should double-check by deriving the associated-token-account address directly and confirming with a direct account-info lookup before concluding the account doesn't exist.

## Steps to Reproduce
1. Attempt to send an SPL token shortly after the sender's or recipient's associated token account was created (or during any transient RPC/indexer delay).
2. Observe a "missing token account" error for the sender, or a transaction that fails at broadcast/simulation because it tries to recreate an ATA that already exists for the recipient.

## Solution
When the account-lookup RPC call returns empty or throws, derive both the classic and Token-2022 associated-token addresses directly and confirm existence with a direct account-info RPC call before deciding the account truly doesn't exist.

### Must Do
- [ ] Add a fallback account-info lookup when the primary token-account lookup returns empty or throws
- [ ] Derive both classic and Token-2022 associated-token addresses for the fallback check
- [ ] Add a unit test covering the fallback path

### Must NOT Do
- Don't change the primary lookup's happy-path behavior when it succeeds normally
- Don't add retries/fallbacks to unrelated RPC calls

## Acceptance Criteria
- [ ] A simulated empty/failed primary lookup no longer immediately reports "account doesn't exist" without the fallback check running

rkokhatskyi:
## Summary

Sending an SPL token relies on `getTokenAssociatedAccountByOwner`, which made a single `getTokenAccountsByOwner` call and treated an empty or failed response as "the token account doesn't exist". When the RPC indexer lags — common right after an associated token account (ATA) is created — that caused two failures:

- the **sender** got a false "missing token account" error, and
- the **recipient** path tried to recreate an ATA that already exists, which the on-chain program rejects (account creation isn't idempotent).

Now, when the primary lookup comes back empty or fails, we derive the classic-SPL and Token-2022 associated-token addresses and confirm existence with a direct `getAccountInfo` call before concluding the account is missing. A successful primary lookup behaves exactly as before.

## Details

- The derivation reuses the same WalletCore calls (`defaultTokenAddress` / `token2022Address`) the signing path already uses to build recipient ATAs, so it yields the canonical, wallet-agnostic address regardless of how the vault was created — generated, imported seed, or Phantom-path import (that choice only affects the owner address, which is already resolved before this point).
- The derivation is injected as a constructor lambda (same pattern as `SolanaFeeService`) so the fallback is unit-tested without the native library; Hilt still builds the impl through the unchanged `@Inject` constructor.
- Follows iOS `SolanaService`; the Android fallback additionally covers the case where the primary call errors (iOS only handles an empty result). Windows has no fallback.

## Testing

- `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.SolanaAssociatedTokenAccountTest"` — 8 tests pass
- `./gradlew :data:ktfmtFormat` — clean
- `./gradlew :app:assembleDebug` — passes

Closes #5224

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
@@ -51,6 +51,7 @@ import kotlinx.serialization.json.contentOrNull
 import kotlinx.serialization.json.jsonPrimitive
 import kotlinx.serialization.json.put
 import timber.log.Timber
+import wallet.core.jni.SolanaAddress
 
 interface SolanaApi {
     suspend fun getBalance(address: String): BigInteger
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5247, opened 2026-07-10
feat(ripple): dedicated XRP Destination Tag field (+ X-addr autofill, RequireDest gate)

### LINKED ISSUE #5184: feat(ripple): dedicated Destination Tag field under Memo in XRP send (+ X-addr autofill, RequireDest warning)
## What
The XRP send view has **no dedicated destination-tag field** — the tag is overloaded onto the memo field, which is error-prone and hides a fund-loss risk. Add a proper destination-tag UX for XRP sends.

## Fix
1. **Add a dedicated "Destination Tag" field as a second field directly under the Memo field** in the XRP send view (numeric, optional).
2. **X-addresses:** accept `X...` address input; auto-populate the destination-tag field from the decoded tag and lock/show it (consumes vultisig-sdk#981).
3. **RequireDestTag warning:** when the destination requires a tag and none is entered, block the send with a clear warning (consumes vultisig-sdk#980).

Android: RippleHelpter.kt builds the Payment; tag currently overloaded onto memo.

## Acceptance
- [ ] Dedicated Destination Tag field rendered as a second field under Memo in the XRP send flow
- [ ] Pasting an X-address auto-fills the tag from the decoded value
- [ ] Sending to a RequireDest account with no tag is blocked with a clear warning
- [ ] Tag is carried into the signed Payment (not via memo overload)

Ref: Figma send view · XRPL dest tags https://xrpl.org/docs/concepts/transactions/source-and-destination-tags · Part of the XRP chain-hardening set (tracker: vultisig-sdk#979). Depends on vultisig-sdk#980, vultisig-sdk#981.

johnnyluo:
## Summary
Adds first-class XRPL **Destination Tag** support to the XRP send flow, mirroring the iOS combo design ([vultisig-ios#4749](https://github.com/vultisig/vultisig-ios/pull/4749)).

- **First-class carrier**: syncs the `commondata` submodule to include `RippleSpecific.destination_tag` (already merged on commondata `main`, PR #95). The tag rides that proto field, **independent** of the free-text memo.
- **Send form**: a collapsible **Destination Tag** field under Memo for XRP (numeric, optional). Auto-expands + locks when an X-address paste autofills it.
- **X-address (XLS-5d)**: hand-rolled codec (WalletCore exposes none on JNI) — paste an `X…` address to normalize to the classic `r…` address and autofill the embedded tag; a tag-0 flag-1 X-address is rejected outright.
- **RequireDest gate**: blocks a tagless send to an account with `lsfRequireDestTag` set (fails closed on lookup error, matching the existing reserve check).
- **Signing** (`RippleHelper`): prefers the tag field; supports tag+memo via rawJSON `DestinationTag`+`Memos`; legacy numeric-memo→tag and swap text-memo→Memos paths preserved.
- **Verify** (initiator **and** join-device — both via the shared `TransactionToUiModelMapper` + `VerifySendScreen`): shows a Destination Tag row from the proto field alongside Memo, when non-empty.
- New strings localized across all 10 locales.

## Issue
Closes #5184

## Design note / tradeoff
The tag is carried **only** in the `destination_tag` proto field (no memo dual-write — impossible now that memo carries independent free-text). A keysign ceremony with a device on a **pre-`destination_tag` build** will **fail to sign** a tagged XRP send rather than silently drop the tag — fail-safe (no fund loss, no wrong tx), but all devices must update. The tag+memo combo path likewise requires all co-signers on this build.

## Test Plan
- [x] `RippleDestinationTagTest` — canonical parser + X-address codec vectors (with/without tag, uint32-max, testnet reject, tag-0 reject, corrupt checksum)
- [x] `AddressManagerTest` — X-address autofill/lock, no-tag leaves editable, clearing releases lock and drops derived tag
- [x] Unit tests: mapper / DefaultSendStrategy / ChainValidationService / VerifyTransaction green
- [ ] `./gradlew lint` (running)
- [ ] Manual: send XRP with a tag → confirm `DestinationTag` on-chain; paste X-address → autofill/lock; tagless send to a RequireDest account → blocked; verify tag shows on both initiator and joining-device Verify screens

On chain test tx link:
https://xrpscan.com/tx/5624D7D4CDC445492C84DF38384196CA67AD4261CB93499B19CEAD408D7950C9
https://xrpscan.com/tx/7A0E125D5BFA1885019E861A3AB4BF7B95EFA9AAABE0B1A9D81B2CAC86D8E94D
https://xrpscan.com/tx/72DD773D9216731A204BA97AF5E6C348030354711A41CDE09BFF919A1193DAF1

## Summary by CodeRabbit

* **New Features**
  * Added an XRP-only “destination tag” field to the send flow (numeric, optional hinting) with XRP-specific validation.
  * Decodes destination tags from pasted XRP X-addresses, normalizes the classic address, and auto-locks the tag when derived.
  * Displays the destination tag as a dedicated row during send verification, suppressing duplicate display when it mirrors the memo.
* **Bug Fixes**
  * Prevents XRP sends when a required destination tag is missing or invalid.
  * Updates XRP fee calculation to remove an unintended destination activation reserve component.
* **Localization**
  * Added destination-tag UI text and XRP destination-tag error messages across supported languages.


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt
```diff
@@ -48,51 +49,110 @@ object RippleHelper {
...
-                                                }
-                                        )
-                                )
-                            ),
+            }
+
+            memo != null -> {
+                // No first-class tag: a purely numeric memo is the legacy destination-tag carrier
+                // (older payloads and swap contracts); any other memo is an on-chain Memos blob.
+                val memoAsLong = memo.toLongOrNull()
```
rkokhatskyi: the legacy memo-as-tag fallback signs `memo.toLongOrNull()` unchecked, while iOS's shipped `RippleDestinationTag.parseCanonical`/`parseTag` rejects `0`, leading zeros and out of uint32 range, so memo `0` or `007` produce a different transaction shape here than on iOS for the same payload

app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToUiModelMapper.kt
```diff
@@ -28,6 +29,8 @@ constructor(
             dstAddress = from.dstAddress,
             dstLabel = from.dstLabel,
             memo = from.memo,
+            destinationTag =
```
rkokhatskyi: `destinationTag` here only reads the dedicated field, never `RippleHelpter.kt`'s legacy memo-as-tag branch. A plain numeric Memo shows under the Memo row on Verify while the broadcast tx actually carries it as `DestinationTag` with no Memos entry

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -197,6 +215,7 @@ internal class DefaultSendStrategy(
                                     chain,
                                 )
                             }
+                            .let { applyRippleDestinationTag(it, destinationTag) }
```
rkokhatskyi, line 235: `applyRippleDestinationTag` never mirrors the tag into `memo`, unlike iOS's shipped `dualWritingRippleTag`, even though commondata's own `destination_tag` commit documents a dual-write rollout so old readers stay byte-identical. A Windows or not-yet-updated co-signer independently rebuilds an untagged payment for the same KeysignPayload

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt
```diff
@@ -48,51 +49,110 @@ object RippleHelper {
...
+        // When the memo merely echoes the tag (memo == the tag's canonical decimal), it is not a
+        // distinct memo — drop it and sign tag-only. This keeps a co-signer that predates the
+        // destination_tag field byte-identical: it reads the numeric memo and builds the same
+        // DestinationTag, so the pre-image matches instead of diverging on a Memos blob.
+        val memoEchoesTag = memo != null && memo == destinationTag?.toString()
+
+        when {
+            // Tag + a distinct free-text memo: WalletCore's OperationPayment can't carry both, so
+            // hand-build a Payment that sets DestinationTag and a Memos blob.
+            destinationTag != null && memo != null && !memoEchoesTag ->
```
rkokhatskyi, line 66: `memoEchoesTag` only checks exact string equality, with no equivalent to iOS's `RippleMemoError.tagMemoConflict`, so a tag field plus a different canonical numeric memo, e.g. tag `100` memo `200`, silently signs a combo rawJson tx here that iOS's shipped signer refuses to build at all

app/src/main/java/com/vultisig/wallet/ui/models/send/ChainValidationService.kt
```diff
@@ -246,6 +247,41 @@ internal class ChainValidationService @Inject constructor(private val rippleApi:
...
+     * No-ops for non-Ripple sends and non-native tokens. Fails closed on a lookup failure, matching
+     * [validateRippleDestinationReserve]: skipping the check would reopen the exact loss it guards.
+     */
+    suspend fun validateRippleDestinationTag(
+        selectedToken: Coin,
+        dstAddress: String,
+        destinationTag: UInt?,
+    ) {
+        if (selectedToken.chain != Chain.Ripple || !selectedToken.isNativeToken) return
+        if (destinationTag != null) return
```
rkokhatskyi, line 264: `validateRippleDestinationTag` only checks the dedicated field via `destinationTag != null`, never `memoFieldState`, so a user who types their required tag into Memo instead (the legacy carrier `RippleHelpter.kt` still honors) gets falsely blocked with destination tag required

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/xrp/RippleFeeService.kt
```diff
@@ -25,43 +24,26 @@ import kotlinx.coroutines.supervisorScope
...
-            } else {
-                BigInteger.ZERO
-            }
-
-        RippleFees(
-            networkFee = networkFee,
-            accountActivationFee = accountActivationFee,
-            amount = networkFee + accountActivationFee,
-        )
+        RippleFees(networkFee = networkFee, amount = networkFee)
```
rkokhatskyi, line 46: this drops the live `reserve_base` fetch entirely, leaving `RippleHelper.DEFAULT_EXISTENTIAL_DEPOSIT`, a hardcoded 1 XRP constant, as the sole gate in `validateRippleDestinationReserve`, with no cross-check against the network's actual current reserve if an XRPL amendment changes it again

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/xrp/RippleFeeService.kt
johnnyluo: because count the activation fee as fee is a mistake , user just need to send more than 1 XRP to activate an address ,  activation fee is not charged along the gas fee of the transaction , that's why we need to remove from it.

On sending screen , we will detect whether an address need to activate and will make sure user send more than 1 if it is not

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt
johnnyluo: Fixed in e4d56faca — the legacy memo-as-tag branch now uses `RippleDestinationTag.parseCanonicalDestinationTag` (the same parser the dedicated field uses), so `0`, leading zeros and out-of-uint32 values fall through to a plain Memos blob instead of being reinterpreted or overflowing a DestinationTag. Matches iOS's parseCanonical/parseTag.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
johnnyluo: Fixed in e4d56faca — `DefaultSendStrategy` now dual-writes the tag into the memo when a tag is set and there's no distinct memo, so a not-yet-updated co-signer that only reads the legacy memo-as-tag carrier rebuilds the same DestinationTag (byte-identical sighash). `memoEchoesTag` drops the echo from the signed tx, so no on-chain Memos blob is added. Added a regression test (`submit dual-writes the XRP destination tag into the memo`).

app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToUiModelMapper.kt
johnnyluo: Fixed in e4d56faca — the mapper now surfaces the tag whether it rides in the first-class proto field or a canonical-numeric memo, and suppresses that memo when it merely echoes the tag (dual-write / legacy carrier) so it isn't shown twice on Verify.

app/src/main/java/com/vultisig/wallet/ui/models/send/ChainValidationService.kt
johnnyluo: Fixed in e4d56faca — `validateRippleDestinationTag` now also treats a canonical numeric memo as a present tag (`destinationTag ?: parseCanonicalDestinationTag(memo)`), so a RequireDest send whose tag is carried in the memo is no longer falsely blocked, matching what RippleHelper actually signs.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt
johnnyluo: Discussed with the author — decision is to keep signing both here. A tag field of 100 with a distinct numeric memo of 200 is a valid XRPL payment carrying both, and reflects exactly what the user entered in the two separate fields; we'd rather honor both than refuse. Not adding a tagMemoConflict error. Happy to revisit if you feel strongly.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -129,7 +131,36 @@ internal class DefaultSendStrategy(
...
+                        } else null
+
+                    // Dual-write the tag into the memo when a tag is set and there's no distinct
+                    // memo, so a not-yet-updated co-signer that only reads the legacy memo-as-tag
+                    // carrier rebuilds the same DestinationTag (byte-identical sighash). Mirrors
+                    // iOS
+                    // dualWritingRippleTag; RippleHelper's memoEchoesTag drops this echo from the
+                    // signed tx, so no on-chain Memos blob is added.
+                    val memo =
+                        if (chain == Chain.Ripple && destinationTag != null && userMemo == null) {
```
rkokhatskyi, line 163: `userMemo == null` here relies on `isNotEmpty()` at line 134, so a single-space memo is non-null and skips the dual-write, while `RippleHelpter.kt` treats that same memo as blank via `isNotBlank()`. A not-yet-updated co-signer reading the un-dual-written memo rebuilds an untagged payment

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt
```diff
@@ -48,51 +49,113 @@ object RippleHelper {
...
-                            ),
+            }
+
+            memo != null -> {
+                // No first-class tag: a canonical uint32 memo is the legacy destination-tag carrier
+                // (older payloads and swap contracts); any other memo is an on-chain Memos blob.
+                // Use the same canonical parser as the dedicated field, so "0", leading zeros and
+                // out-of-uint32 values are treated as a plain memo (matching iOS) instead of being
+                // silently reinterpreted — or overflowed — into a DestinationTag.
+                val memoAsTag = RippleDestinationTag.parseCanonicalDestinationTag(memo)
```
rkokhatskyi, line 90: `parseCanonicalDestinationTag` folds shape and nonzero into one check, so memo `0` falls to the plain-text branch and signs successfully with no tag. iOS's `parseCanonical`/`parseTag` split those and throw `invalidMemo` for exactly `0`, so this comment's claimed iOS parity doesn't hold for that value

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/xrp/RippleFeeService.kt
rkokhatskyi: traced the causal chain through git history: DEFAULT_EXISTENTIAL_DEPOSIT and validateRippleDestinationReserve both predate this PR and never referenced the fee-side getBaseReserve() fetch this PR removes, they were always two independent, unreconciled reserve numbers. dropping this, the fee/reserve separation itself is correct and matches iOS

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt
rkokhatskyi: checked iOS's own stated reason for rejecting this shape: the concern isn't user confusion, it's that the wire carries two independent real conventions for a destination tag, the native field and the legacy numeric-memo convention this PR's own dual-write exists to serve for old readers, so DestinationTag=100 plus Memos="200" is two valid-looking, mutually exclusive interpretations depending on which convention the receiving system trusts, not cosmetic redundancy. it's also more reachable than deliberate dual-typing since SendFormGraph.kt prefills memo from deep-link/nav args, so a prefilled memo plus a hand-typed tag hits this silently, and XRP tag/memo mixups are the most common real-world XRP fund-loss failure mode industry-wide. still think this is worth closing to match iOS, happy to hear more if you disagree

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
johnnyluo: Fixed in 6fb379974 — `userMemo` now uses `isNotBlank()` (not `isNotEmpty()`), matching `RippleHelper`'s own `isNotBlank()`. A whitespace-only memo is treated as absent, so the tag dual-write kicks in and a not-yet-updated co-signer rebuilds the same tagged payment.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt
johnnyluo: You're right that the comment overclaimed exact iOS parity for `0`. Per the earlier decision on this PR we intentionally keep a bare `0`/leading-zero/out-of-range memo as free text (a plain Memos blob) rather than reject it — the safety property that matters is that it's never reinterpreted or overflowed into a DestinationTag. Corrected the comment in 6fb379974 to state that explicitly instead of claiming iOS parity.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt
johnnyluo: this situation is temporary , memo is only useful for THORChain swaps ,  I think no one is actually use memo field except THORChain swaps 

if user has a way to specify a desitnation tag , which means they already updated the app to a version can support both destination tag and memo , so it is fine for us just treat it separately 

Very soon , IOS will also remove dual write logic , once current version get released

rkokhatskyi (approved):
LGTM

PR #5253, opened 2026-07-10
fix(ripple): gate Confirmed on validated flag + meta.TransactionResult

### LINKED ISSUE #5248: XRP: RippleStatusProvider marks Confirmed without validated flag or meta.TransactionResult — failed/unvalidated payments show Confirmed
## Problem — XRP txs shown as Confirmed that failed or were never validated

`RippleStatusProvider` marks a tx **Confirmed** solely on `tx.result.status == "success"` (`data/.../api/txstatus/RippleStatusProvider.kt:19`). For the XRPL `tx` method (api_version 2, `RippleApi.kt:63-80`), RPC-level `status: "success"` only means *the tx was found* — it says nothing about outcome. Two concrete false-success modes:

1. **On-chain FAILED payments show Confirmed.** A validated tx with a `tec*` result (`tecUNFUNDED_PAYMENT`, `tecDST_TAG_NEEDED`, `tecNO_DST_INSUF_XRP`, …) burned the fee and delivered NO funds — but XRPL still returns `status:"success"` for it. The model never even deserializes `meta` (`RippleJson.kt:24-30` parses only hash/status/tx_json/validated), so `meta.TransactionResult` is never checked.
2. **Unvalidated txs show Confirmed — terminally.** The `validated` flag IS parsed (`RippleJson.kt:29`) but never read (`RippleStatusProvider.kt:17-21`). A tx found in a not-yet-validated ledger yields Confirmed; `updateToConfirmed` then drops the row from the pending poll set (`TransactionHistoryRepository.kt:71-72`), so if the tx is later dropped (LastLedgerSequence expiry, failed consensus) it stays Confirmed forever.

## Fix

Gate Confirmed on `tx.result.validated == true` AND `meta.TransactionResult == "tesSUCCESS"` (mirror SolanaStatusProvider's "finalized" gate); `tec*` → Failed with the code surfaced; unvalidated → keep polling.

## Context
Native parity audit 2026-07-10 (adversarially verified, S2 ×2). Same false-success class as vultisig-sdk#1119. Report: `native-parity-matrix-report-2026-07-10.md`.

johnnyluo:
## Summary
- `RippleStatusProvider` marked a tx **Confirmed** on RPC-level `status == "success"`, which for the XRPL `tx` method only means the tx was *found* — not that it succeeded.
- Now Confirmed requires `validated == true` **AND** `meta.TransactionResult == "tesSUCCESS"`.
- Validated `tec*` results (`tecUNFUNDED_PAYMENT`, `tecDST_TAG_NEEDED`, …) → `Failed(code)` instead of a false-green Confirmed.
- Not-yet-validated or meta-absent → keep polling (`Pending`) instead of a terminal Confirmed that could later be dropped (LastLedgerSequence expiry / failed consensus).
- Added `meta` to the `tx` response model (`RippleTxMetaJson`); it was parsed nowhere before.
- Mirrors the Solana `finalized` gate.

## Issue
Closes #5248

On Chain Tx Link https://xrpscan.com/tx/F5E925EEF91BF62AB5487330B2D8365AB5C2DADCF01E12CC3CBAFDBB89C39CEA

## Test Plan
- [x] Added `RippleStatusProviderTest` covering both false-success modes (validated tec*, unvalidated tesSUCCESS, missing meta)
- [x] Lint / ktfmt passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Ripple transaction status tracking by waiting for ledger validation before confirming transactions.
  * Correctly distinguishes successful, failed, pending, incomplete, and unavailable transaction responses.
  * Prevents missing transaction metadata or API errors from being incorrectly reported as failures.

* **Tests**
  * Added coverage for validated, unvalidated, failed, incomplete, null, and error responses.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/RippleStatusProvider.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/RippleStatusProvider.kt
@@ -13,11 +13,20 @@ class RippleStatusProvider @Inject constructor(private val rippleApi: RippleApi)
 
     override suspend fun checkStatus(txHash: String, chain: Chain): TransactionResult =
         try {
-            val tx = rippleApi.getTsStatus(txHash)
+            val result = rippleApi.getTsStatus(txHash)?.result
+            val engineResult = result?.meta?.transactionResult
             when {
-                tx == null -> TransactionResult.Pending
-                tx.result.status == "success" -> TransactionResult.Confirmed
-                else -> TransactionResult.Failed(tx.result.status)
+                // Not found yet, or found only in a not-yet-validated ledger: keep polling. A tx in
+                // an unvalidated ledger can still be dropped (LastLedgerSequence expiry, failed
+                // consensus), so it must not be reported terminally.
+                result == null || !result.validated -> TransactionResult.Pending
+                // Validated but outcome unknown (meta absent): don't claim success, keep polling.
+                engineResult == null -> TransactionResult.Pending
+                // Only tesSUCCESS delivered funds. A validated tec* result (tecUNFUNDED_PAYMENT,
+                // tecDST_TAG_NEEDED, …) burned the fee and delivered nothing — it is a real
+                // failure.
+                engineResult == "tesSUCCESS" -> TransactionResult.Confirmed
+                else -> TransactionResult.Failed(engineResult)
             }
         } catch (e: CancellationException) {
             throw e
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/models/RippleJson.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/models/RippleJson.kt
@@ -27,7 +27,11 @@ data class RippleBroadcastSuccessResultJson(
     @SerialName("status") val status: String,
     @SerialName("tx_json") val txJson: RippleBroadcastSuccessTransactionJson,
     @SerialName("validated") val validated: Boolean,
+    @SerialName("meta") val meta: RippleTxMetaJson? = null,
 )
+
+@Serializable
+data class RippleTxMetaJson(@SerialName("TransactionResult") val transactionResult: String? = null)
 
 @Serializable
 data class RippleBroadcastSuccessTransactionJson(
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/txstatus/RippleStatusProviderTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/txstatus/RippleStatusProviderTest.kt
@@ -1,0 +1,89 @@
+package com.vultisig.wallet.data.api.txstatus
+
+import RippleBroadcastSuccessResponseJson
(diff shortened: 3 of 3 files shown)


aminsato (commented):
The core fix is correct — gating Confirmed on `validated == true` AND `meta.TransactionResult == "tesSUCCESS"` matches XRPL semantics, and validated `tec*` → `Failed` is right. Two non-blocking notes below on pre-existing polling behavior (not introduced by this PR), just for awareness.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/RippleStatusProvider.kt
```diff
@@ -13,11 +13,20 @@ class RippleStatusProvider @Inject constructor(private val rippleApi: RippleApi)
...
+            val result = rippleApi.getTsStatus(txHash)?.result
+            val engineResult = result?.meta?.transactionResult
             when {
-                tx == null -> TransactionResult.Pending
-                tx.result.status == "success" -> TransactionResult.Confirmed
-                else -> TransactionResult.Failed(tx.result.status)
+                // Not found yet, or found only in a not-yet-validated ledger: keep polling. A tx in
+                // an unvalidated ledger can still be dropped (LastLedgerSequence expiry, failed
+                // consensus), so it must not be reported terminally.
+                result == null || !result.validated -> TransactionResult.Pending
```
aminsato, line 22: Non-blocking, pre-existing: a permanently-dropped/expired tx (`txnNotFound`) fails to deserialize into `RippleBroadcastSuccessResultJson` → throws → caught below → `Pending`. So a dropped tx is indistinguishable from a transient network blip and polls forever. Not introduced by this PR (relies on the caller's own timeout), but worth noting since this branch now formalizes the not-found → Pending contract.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/RippleStatusProvider.kt
```diff
@@ -13,11 +13,20 @@ class RippleStatusProvider @Inject constructor(private val rippleApi: RippleApi)
...
+                // Validated but outcome unknown (meta absent): don't claim success, keep polling.
+                engineResult == null -> TransactionResult.Pending
+                // Only tesSUCCESS delivered funds. A validated tec* result (tecUNFUNDED_PAYMENT,
+                // tecDST_TAG_NEEDED, …) burned the fee and delivered nothing — it is a real
+                // failure.
+                engineResult == "tesSUCCESS" -> TransactionResult.Confirmed
+                else -> TransactionResult.Failed(engineResult)
```
aminsato, line 29: Non-blocking: `tef*`/`tel*`/`tem*` codes never reach a validated ledger, so they never land in `else -> Failed` — they sit at `Pending` until the poller gives up. Consistent with the drop-handling above and acceptable, just flagging the tail behavior.

rkokhatskyi (approved):
LGTM

Issue #5250, opened 2026-07-10
Broadcast dedup/recovery: string-match without on-chain verify swallows rejections as success (EVM/Cardano/Sui/QBTC/Tron — one shared surface)


realpaaao:
## Problem — the broadcast dedup/recovery table swallows rejections as success (5 chains, one shared surface)

`BroadcastTxUseCase` gives BTC/Solana/Polkadot/Ton/Ripple/Tron/Cardano a `recoverIfAlreadyBroadcast` verify pattern (3 attempts, `BroadcastTxUseCase.kt:198-249`) — but the per-chain dedup heuristics around it accept **string evidence without on-chain verification**, and two chains are outside the pattern entirely:

1. **EVM: `contains("known")` matches "unknown".** `EvmApi.kt:484` swallows any `eth_sendRawTransaction` rejection whose message contains `ERROR_KNOWN = "known"` (`EvmApi.kt:736`) — including "unknown sender", "unknown block", gateway "unknown method" — and returns the locally computed keccak hash (`EvmApi.kt:498-499`) with **no on-chain verify anywhere downstream** (`BroadcastTxUseCase.kt:102-117` has no verify call). User sees success with a hash that will never exist.
2. **EVM: "nonce too low" (+10 substrings) accepted with no hash check.** If the nonce was consumed by a DIFFERENT tx (stale nonce from spec-build, activity from another wallet on the same address), the rejection is reported as success (`EvmApi.kt:481-503`).
3. **Cardano: recovery returns the WRONG hash.** The 400/`unknownOutputReferences` recovery extracts `error.data.unknownOutputReferences[0].transaction.id` — the txid of the **parent** tx that created the spent input (an outpoint reference, see `OgmiosTransactionResponse.kt:27-38`), not the tx being broadcast (`CardanoApi.kt:118-127`). A genuine concurrent-spend rejection is reported as success under an unrelated txid.
4. **Sui: excluded from recovery entirely.** Direct broadcast, no recover wrapper (`BroadcastTxUseCase.kt:158-160`); the joined-device fallback needs a non-blank precomputed hash + a Cosmos-specific exception (`BroadcastKeysignUseCase.kt:245-248, 263-264`) — a second-device duplicate submission shows a hard failure for a tx that landed.
[...]

PR #5254, opened 2026-07-10
fix(broadcast): verify tx on-chain before treating dedup rejections as success

johnnyluo:
## Summary
The per-chain broadcast dedup heuristics accepted string evidence without an on-chain check, so genuine rejections could surface as success (issue #5250). This fixes the four self-contained cases:

- **EVM — bare `"known"` matched `"unknown"`** (`EvmApi.kt`): dropped the bare `"known"` substring, which also matched `"unknown sender/block/method"`. Such rejections now surface as failures instead of a fabricated success hash.
- **EVM — `"nonce too low"` accepted blind** (`EvmApi.kt`): *already-known* errors (our exact signed bytes are already held → the keccak hash is canonical) still return immediately; *nonce-too-low* is ambiguous (a **different** tx may have consumed the nonce), so we now confirm our own hash has a receipt (`isTxMined`, 3× retry) before reporting success, otherwise surface the error.
- **Cardano — recovery returned the wrong hash** (`CardanoApi.kt`): a 400 no longer extracts the parent txid from `unknownOutputReferences` (an unrelated outpoint reference). It throws so `BroadcastTxUseCase` verifies our *actual* `transactionHash` on-chain.
- **Tron — verify window couldn't work** (`TronApi.kt`): verify now hits the full node (`wallet/gettransactionbyid`) instead of `walletsolidity` (~60s solidify lag > the ~4–6s verify loop), so a just-broadcast tx confirms instead of false-failing.

### Deferred (need dedicated changes, not partial patches)
- **Sui**: the digest isn't precomputable from the raw tx, so `transactionHash` is blank and recovery structurally can't work — needs SDK support to expose the digest pre-broadcast.
- **Cosmos/QBTC code-32 joined-device recovery**: the blind trust lives in `BroadcastKeysignUseCase.recoverJoinedDeviceBroadcast` (sync, no API access). Doing it right means moving the on-chain verify into `BroadcastTxUseCase` and removing the separate joined-device path — a behavioral refactor of a money-path use case with existing tests.

## Issue
Refs #5250

## Test Plan
- [x] New unit tests: `EvmApiBroadcastTest` (5 cases — already-known, unknown-sender surfaced, nonce-too-low mined vs. not-mined, plain success)
- [ ] Lint passes (CI)
- [ ] Debug build succeeds (CI)

On chain txs: 
EVM: https://etherscan.io/tx/0xb3458c8dc8c79cf2f2fd6df24291e27a4c2bcb5393c5c3bf8d10e71558b635c6
Tron: https://tronscan.org/#/transaction/e6658f1b78a04af297d52e76e38f2e00b11f8e11323dc90a1e5401db30838469
Cardano: https://cardanoscan.io/transaction/c8c32f265bbbf8246234e3bbb1ea04c999ca71da5278f894e518a04eacf338f2

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved EVM transaction broadcast handling for already-broadcast and “nonce too low” cases, including receipt verification when needed.
  * Updated Cardano broadcast behavior so BadRequest with `unknownOutputReferences` is treated as a real rejection (not a success).
  * Refined Cardano confirmation checks to rely on actual confirmation count.
  * Improved Tron transaction status checks by querying the correct full-node endpoint.
  * Enhanced ERC-20 approval broadcast fallback to reduce failures during duplicate-broadcast races.
* **Tests**
  * Added EVM broadcast recovery/deduplication coverage and tightened rejection/confirmation scenarios.
  * Updated Cardano broadcast tests to match the new rejection behavior.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
@@ -116,14 +116,17 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Cardan
                 }
 
                 HttpStatusCode.BadRequest -> {
+                    // A 400 (e.g. unknownOutputReferences) is a genuine rejection. The txid inside
+                    // unknownOutputReferences is the PARENT tx that created the spent input, not
+                    // the
+                    // tx we broadcast — returning it would report success under an unrelated hash.
+                    // Throw instead so BroadcastTxUseCase can verify our actual hash on-chain and
+                    // only treat a real duplicate submission as success.
                     val ogmiosError =
                         json.decodeFromString<OgmiosTransactionResponse>(response.bodyAsText())
-                    ogmiosError.error?.data?.unknownOutputReferences?.firstOrNull()?.transaction?.id
-                        ?: run {
-                            val errorMessage = ogmiosError.error?.message ?: "Unknown error"
-                            Timber.e("Cardano transaction submission failed: $errorMessage")
-                            error("Failed to broadcast transaction: $errorMessage")
-                        }
+                    val errorMessage = ogmiosError.error?.message ?: "Unknown error"
+                    Timber.e("Cardano transaction submission failed: $errorMessage")
+                    error("Failed to broadcast transaction: $errorMessage")
                 }
 
                 else -> {
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
@@ -19,7 +19,6 @@ import io.ktor.http.path
 import java.math.BigInteger
 import javax.inject.Inject
 import kotlin.coroutines.cancellation.CancellationException
-import kotlinx.coroutines.delay
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.buildJsonObject
 import kotlinx.serialization.json.put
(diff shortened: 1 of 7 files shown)


johnnyluo:
@coderabbitai review

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -480,30 +481,33 @@ class EvmApiImp(
...
-            ) {
-                // even the server returns an error , but this still consider as success
-                return Numeric.hexStringToByteArray(signedTransaction).toKeccak256()
-            } else {
-                throw Exception(responseBody)
+            val localHash = Numeric.hexStringToByteArray(signedTransaction).toKeccak256()
+            return when {
+                // The node already holds our exact signed bytes, so the deterministic keccak hash
+                // is the canonical on-chain id — safe to report without re-verifying.
+                ALREADY_BROADCAST_ERRORS.any { message.contains(it) } -> localHash
```
aminsato, line 493: Dropping the bare `"known"` substring fixes the `"unknown sender"` false-positive, but it also narrows real dedup coverage — and the EVM broadcast path in `BroadcastTxUseCase` calls `sendTransaction` directly with **no** `recoverIfAlreadyBroadcast` fallback, so any already-broadcast message that no longer matches becomes a user-facing hard failure on a tx that is actually in the mempool.

Two legitimate dedup responses now fall through to `throw Exception(responseBody)`:
- `"known transaction: 0x…"` — emitted by some geth-based / Erigon nodes; matched before via the bare `"known"`, not by `"already known"`.
- `"Already known"` (capital A, e.g. Nethermind) — `message.contains("already known")` is case-sensitive and misses it; the old lowercase `"known"` substring caught it.

Suggest matching on the concept while excluding only the `unknown` collision instead of enumerating exact strings:
```kotlin
message.contains("known", ignoreCase = true) && !message.contains("unknown", ignoreCase = true) -> localHash
```
That keeps `"already known"`, `"known transaction: …"`, and `"Already known"` as success while still surfacing `"unknown sender/block/method"` as a real rejection. Worth a test case pinning `"known transaction: 0x…"` → success alongside the existing `"unknown sender"` → surfaced case.

aminsato (dismissed):
The nonce-too-low `isTxMined` gate, the Cardano throw-and-verify path (wired through `recoverIfAlreadyBroadcast`), and the Tron full-node switch all check out. One change requested on the EVM dedup set below.

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
```diff
@@ -116,14 +116,17 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Cardan
...
                         json.decodeFromString<OgmiosTransactionResponse>(response.bodyAsText())
-                    ogmiosError.error?.data?.unknownOutputReferences?.firstOrNull()?.transaction?.id
-                        ?: run {
-                            val errorMessage = ogmiosError.error?.message ?: "Unknown error"
-                            Timber.e("Cardano transaction submission failed: $errorMessage")
-                            error("Failed to broadcast transaction: $errorMessage")
-                        }
+                    val errorMessage = ogmiosError.error?.message ?: "Unknown error"
+                    Timber.e("Cardano transaction submission failed: $errorMessage")
+                    error("Failed to broadcast transaction: $errorMessage")
```
rkokhatskyi, line 129: Every 400 now routes to `BroadcastTxUseCase.kt:194`'s verify, `getTxStatus(hash)?.txHash?.isNotBlank()`, but txHash is non-nullable so it's true for any Koios row, not proof of confirmation, unlike `CardanoStatusProvider.kt:19`'s own `numConfirmations > 0` bar. Every genuine rejection now reports fabricated success.

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
```diff
@@ -116,14 +116,17 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Cardan
                 }
 
                 HttpStatusCode.BadRequest -> {
```
rkokhatskyi, line 118: The recovery verify (`BroadcastTxUseCase.kt:225-249`) allows only ~4-6s (3 attempts, 2s backoff) before giving up, but Cardano blocks land roughly every 20s, so a genuinely successful duplicate broadcast from a losing co-signer can still surface as a hard failure, needs mainnet confirmation.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -480,30 +481,33 @@ class EvmApiImp(
...
+                // "nonce too low" is ambiguous: either our own tx already mined, or a DIFFERENT tx
+                // consumed the nonce (stale spec-build nonce, another wallet on the same address).
+                // Only report success if OUR hash is actually on chain; otherwise surface the
+                // error.
+                NONCE_TOO_LOW_ERRORS.any { message.contains(it) } ->
```
rkokhatskyi: `BroadcastKeysignUseCase.kt:109` broadcasts the ERC-20 approve tx with no try/catch, unlike the main tx's `broadcastOrRecover`. A losing co-signer's approve dedup, previously always swallowed, can now throw here and skip `ApprovalNotConfirmed`, landing on a generic error screen with the approve hash never surfaced.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -733,17 +737,28 @@ class EvmApiImp(
...
-        private const val ERROR_TX_IN_MEMPOOL = "tx already in mempool"
-        private const val ERROR_EXISTING_TX = "existing tx"
-        private const val ERROR_TX_IN_CACHE = "tx already exists in cache"
-        private const val ERROR_CODE_10055 = "code=10055"
+        // Node already has our exact signed tx; the keccak hash of our bytes is canonical.
+        // ("already known" replaces the old bare "known", which also matched "unknown …".)
+        private val ALREADY_BROADCAST_ERRORS =
+            listOf(
+                "already known",
+                "Transaction is temporarily banned",
```
rkokhatskyi: "Transaction is temporarily banned" sits in the zero-verification list while `NONCE_TOO_LOW_ERRORS` two lines below requires `isTxMined()`. Substrate's tx-pool bans on any validation failure, not only an already-mined tx, so a permanently rejected transaction can still report a fabricated success hash here.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -480,30 +481,33 @@ class EvmApiImp(
...
         }
         return jsonObject.result ?: error("send transaction failed")
     }
 
+    // Confirms a specific tx hash landed by looking up its receipt (present only once mined, which
+    // is exactly the "nonce too low" case). Retries briefly to absorb receipt propagation lag.
+    private suspend fun isTxMined(txHash: String): Boolean {
+        repeat(TX_MINED_ATTEMPTS) { attempt ->
+            if (attempt > 0) delay(TX_MINED_BACKOFF_MS)
+            if (getTxStatus(txHash)?.result != null) return true
```
rkokhatskyi, line 511: `isTxMined` budgets only 3 attempts, 2s apart (~4s), across all 13 EVM chains sharing `api.vultisig.com` as a multi-upstream proxy. If the receipt-serving replica lags the one that returned the error, an already-mined tx (nonce already consumed) is hard-failed, needs mainnet confirmation.

data/src/test/kotlin/com/vultisig/wallet/data/api/EvmApiBroadcastTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    @Test
+    fun `unknown sender rejection is surfaced, not swallowed`() = runTest {
+        // "unknown sender" contains "known" — the old bare match wrongly reported it as success.
+        val client =
+            MockHttpClient.respondingWith(HttpStatusCode.OK, errorResponse("unknown sender").second)
+        assertFailsWith<Exception> { api(client).sendTransaction(signedTx) }
+    }
+
+    @Test
+    fun `nonce too low is success only when our hash is actually mined`() = runTest {
```
rkokhatskyi, line 61: This only exercises `isTxMined()`'s first attempt since `respondingWithSequence`'s 2nd response already returns a receipt. A regression to `TX_MINED_ATTEMPTS = 1` or a broken backoff would still pass. A 3-response sequence (error, not-yet-mined, mined) would actually cover the retry.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -733,17 +737,28 @@ class EvmApiImp(
...
-        private const val ERROR_TEMPORARILY_BANNED = "Transaction is temporarily banned"
-        private const val ERROR_NONCE_TOO_LOW_NEXT = "nonce too low: next nonce"
-        private const val ERROR_NONCE_TOO_LOW_RANGE = "nonce too low. allowed nonce range:"
-        private const val ERROR_TX_ALREADY_EXISTS = "transaction already exists"
-        private const val ERROR_NONCE_TOO_LOW_ADDRESS = "nonce too low: address"
-        private const val ERROR_TX_IN_MEMPOOL = "tx already in mempool"
-        private const val ERROR_EXISTING_TX = "existing tx"
-        private const val ERROR_TX_IN_CACHE = "tx already exists in cache"
-        private const val ERROR_CODE_10055 = "code=10055"
+        // Node already has our exact signed tx; the keccak hash of our bytes is canonical.
```
rkokhatskyi, line 745: This comment restates the when-branch explanation at line 486 almost verbatim and narrates removed code ("the old bare known"), which is gone from the file and will rot. Keep one explanation at the usage site and drop this copy.

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
johnnyluo: Intentional trade-off, leaving as-is. With the verify now requiring `numConfirmations > 0`, a legit duplicate that hasn't landed within the ~4-6s window surfaces as a *failure*, not a fabricated success. That's the safe side per this PR's whole premise (#5250): a losing co-signer seeing a transient error and retrying is far better than reporting success for a tx that never confirmed. Widening the window only trades that safety for convenience; happy to revisit if mainnet shows false-failures are common.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
johnnyluo: Same reasoning as the Cardano window thread: the short `isTxMined` budget errs toward false-failure, never false-success. If the receipt replica lags, we surface the node's error and the user retries — safe. Extending retries just to catch replica lag reintroduces the ambiguity this PR removes. Leaving as-is pending mainnet signal that false-failures are actually happening.

rkokhatskyi (approved):
LGTM

PR #5257, opened 2026-07-10
fix(ton): apply wallet-level bounceable to every TonConnect message

aminsato:
**Proof (fixed swap, broadcast on-chain):** https://tonviewer.com/transaction/1f88b7cd71d55f9c544e2f3937dfed5a1772652ea606b64039574f6a11cec28e

## Problem

Multi-message TonConnect swaps initiated from the browser extension (e.g. a STON.fi swap) never finished co-signing. The mobile co-signer sat forever on `GET /router/setup-message` returning `404`, and the signing spinner never resolved.

## Root cause

In DKLS keysign the setup message is stored under a key of `md5(pre-image-hash)`. The initiator uploads under `md5(its hash)`; each co-signer fetches under `md5(its own hash)`. If a co-signer computes a **different** pre-image hash than the initiator, it polls a message-id that was never uploaded and 404s indefinitely.

The hashes diverged on the TON `bounceable` flag:

- **Initiator (browser extension / desktop)** applies the single wallet-level `tonSpecific.bounceable` to **every** TonConnect message.
- **Android** derived `bounceable` **per-message** from the friendly-address tag (`EQ → true`, `UQ → false`).

A swap whose message targets a `UQ` (non-bounceable) address — STON.fi attaches a self/peer message to a `UQ...` address — made Android flip that transfer to `bounceable=false` while the initiator kept `true`. Different transfer cell → different pre-image hash → different `md5` message-id → permanent `404` on the setup message.

Multi-message TonConnect is only ever initiated by the extension/desktop, so mobile is always the co-signer and must reproduce the initiator's hash.

## Fix

`TonHelper.buildTonConnectTransfer` now uses `tonSpecific.bounceable` for every TonConnect transfer instead of parsing the `EQ`/`UQ` prefix, matching the initiator's signing input byte-for-byte.

## Test plan

- Added `tonConnectBounceableFollowsWalletFlagNotAddressPrefix` regression: a `bounceable=true` payload whose second message targets a `UQ` address now yields `bounceable=true` on both transfers (previously `false` on the `UQ` one).
- Existing `tonConnectPayloadAndStateInitAreThreadedIntoSigningInput` and `sendTONTest` still pass (unaffected — they use `bounceable=false`).
- Verified end-to-end: STON.fi GRAM→USDT swap now signs and broadcasts (proof link above).

## Cross-platform note

iOS `Ton.swift` `isBounceable` has the same per-address derivation and would hit the same `404` co-signing an extension-initiated `UQ` swap. Tracking a matching iOS fix separately so all platforms stay hash-consistent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed TON Connect transfers to consistently use the wallet’s bounceable setting for all signing messages, without deriving bounceability from destination address prefixes.
  * Improved nominator-pool comment transfers so bounceable is correctly set based on both the wallet/TON setting and supported comment types.
* **Tests**
  * Added a regression instrumentation test to verify bounceable behavior in TON signing payloads.
  * Expanded unit test coverage for recognizing valid nominator-pool transfer comments.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/TonHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/TonHelper.kt
@@ -3,6 +3,7 @@
 package com.vultisig.wallet.data.crypto
 
 import com.google.protobuf.ByteString
+import com.vultisig.wallet.data.blockchain.ton.TonNominatorPool
 import com.vultisig.wallet.data.common.toHexByteArray
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.SignedTransactionResult
@@ -113,11 +114,18 @@ object TonHelper {
 
         val mode = calculateSendMode(tonSpecific.sendMaxAmount)
 
+        // Nominator-pool deposits/withdrawals MUST be sent bounceable so a message the pool
+        // rejects (e.g. an uninitialized or mis-funded pool) bounces back instead of being
+        // absorbed (lost). tonSpecific.bounceable resolves to false for an uninitialized
+        // destination, so force the flag for pool comments — matching the initiating device
+        // which does the same — otherwise the pre-image hash diverges and the joiner 404s.
+        val bounceable = tonSpecific.bounceable || TonNominatorPool.isTransferComment(payload.memo)
+
         return TheOpenNetwork.Transfer.newBuilder()
             .setDest(toAddress.description())
             .setAmount(ByteString.copyFrom(amount.toHexString().toHexByteArray()))
             .setMode(mode)
-            .setBounceable(tonSpecific.bounceable)
+            .setBounceable(bounceable)
             .apply { payload.memo?.let { setComment(it) } }
             .build()
     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/TonHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/TonHelper.kt
@@ -83,15 +83,14 @@ object TonHelper {
         val toAddress = AnyAddress(msg.to, CoinType.TON)
         val amount = msg.amount.toLongOrNull() ?: 0L
         val mode = calculateSendMode(sendMaxAmount = false)
-        // TonConnect addresses encode bounceability in the user-friendly prefix:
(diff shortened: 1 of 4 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/crypto/TonHelper.kt
```diff
@@ -83,15 +83,14 @@ object TonHelper {
...
-                else -> tonSpecific.bounceable
-            }
+        // Apply the wallet-level bounceable flag from tonSpecific to every TonConnect message,
+        // matching the initiating device (browser extension / desktop) which builds the setup
+        // message that all co-signers must reproduce. It does NOT derive bounceability per-address
+        // from the EQ/UQ friendly-address tag. Deriving it per-address here diverged the pre-image
+        // hash from the initiator whenever a message targeted a UQ (non-bounceable) address — e.g.
+        // STON.fi's self/peer message in a swap — so the md5 message-id differed, the joiner 404'd
+        // on the setup message, and co-signing never completed.
+        val bounceable = tonSpecific.bounceable
```
rkokhatskyi, line 94: Windows forces bounceable=true whenever memo is `d` or `w` (nominator-pool stake), overriding tonSpecific.bounceable at the same call site this fix targets (`ton/index.js:12,22,28`). The native path at `TonHelper.kt:120` has no such override, so a stake tx to an uninitialized pool can still diverge the hash and reproduce this exact 404

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/crypto/TonHelper.kt
aminsato: Addressed in d81ea112b. `buildNativeTransfer` now forces the bounce flag for nominator-pool comments: `bounceable = tonSpecific.bounceable || TonNominatorPool.isTransferComment(payload.memo)`, covering both the tf (`d`/`w`) and whales (`Deposit`/`Withdraw`) comments. Root cause was `BlockChainSpecificRepository` forcing `bounceable=false` for an uninitialized pool destination, which diverged the native pre-image hash from the initiator. Added `TonNominatorPool.isTransferComment` plus test coverage.

johnnyluo (approved):
LGTM

Issue #4794, opened 2026-06-08
Refactor PRs are net-additive; SwapFormViewModel god-method survived #4778


aminsato:
# [Fix] SwapFormViewModel god-method survived net-additive refactor [Swap]

## Problem
Recent "refactor" PRs are consistently net-additive — adding 2–4× more lines than they remove — while the core complexity they target survives. `SwapFormViewModel` (#4778) is the clearest example: the PR was titled "extract quote-fetch pipeline and shrink the god-method," but after the change the file is still ~1150 lines and `calculateFees()` is still ~390 lines. Collaborators were extracted but the orchestration stayed inline, so the headline "shrink the god-method" goal was not met.

## Expected Behavior
A refactor of `SwapFormViewModel` should leave total complexity flat or lower. The quote pipeline should be a meaningfully smaller, single-responsibility unit; its method should be named for what it actually does; and extracted collaborators should expose return-value seams instead of mutating ViewModel internals.

## Steps to Reproduce
1. Open `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt`.
2. Observe the file is ~1150 lines and `calculateFees()` is ~390 lines.
3. Note the single `collectLatest` block still does provider lookup, per-provider discount fan-out, best-quote resolution, THORCHAIN referral handling, Vult BPS discounts, the SwapKit-UTXO fee special case, UTXO plan-fee computation, balance validation, and four catch arms — nested ~6 levels deep.
4. Note that collaborators (`SwapQuoteManager`, `SwapValidator`, `SwapGasCalculator`, `SwapDiscountChecker`, `SwapTransactionBuilder`) mutate the ViewModel's internals via passed-in `MutableStateFlow`s, a `Job`, `uiState`, and `viewModelScope` rather than returning values (e.g. `collectSelectedTokens(addresses, selectedSrcId, selectedDstId, selectedSrc, selectedDst, chain, selectTokensJob, viewModelScope)`).
5. Note `resetQuoteState()` must poke both `QuoteStateHolder` and the still-scattered loose flows (`swapFeeFiat`, `estimatedNetworkFeeTokenValue`, `estimatedNetworkFeeFiatValue`, `gasFee`, `gasFeeChain`, `referralCode`).

### Must Do
- [ ] Extract the `collectLatest` body into a `SwapQuotePipeline` that takes inputs and returns a sealed result, leaving the ViewModel to wire flows to UI state.
[...]

PR #5261, opened 2026-07-12
fix(thorchain): read on-chain sRUJI receipt for Staked RUJI amount

### LINKED ISSUE #5258: [BUG] DeFi → Staked: "Staked RUJI" reads bonded amount, shows 0 when the vault holds sRUJI receipts
## Summary
On **DeFi → Staked**, the **Staked RUJI** amount is read from the Rujira staking API's `bonded` field, which can return `0` even when the vault holds sRUJI staking-receipt tokens (`x/staking-x/ruji`). Same bug as iOS (vultisig/vultisig-ios#4794); already fixed in vultisig-windows.

## Where
`data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt` — `getRujiStakeBalance`:
```kotlin
val stake = response.data?.node?.stakingV2?.firstOrNull() ?: return RujiStakeBalances()
val stakeAmount = stake.bonded.amount.toBigIntegerOrNull() ?: BigInteger.ZERO
```
`stake.bonded.amount` can be `0` even when receipts are held on-chain; the on-chain receipt balance is the real source of truth.

## Reference fix (vultisig-windows)
Read the on-chain sRUJI balance directly and prefer it over `bonded`, mirroring how **sTCY** derives its amount from `x/staking-tcy`:
`.../cosmos/bank/v1beta1/balances/{address}/by_denom?denom=x/staking-x/ruji`

- Windows issue: vultisig/vultisig-windows#4329
- Windows PR: vultisig/vultisig-windows#4337
- Windows fix location: `core/ui/defi/chain/queries/services/thorchainStake/rujiStakeService.ts` (reads receipt balance, falls back to `bonded` only when nothing is held).

## Suggested change
Derive the Staked RUJI amount from the on-chain `x/staking-x/ruji` balance (like sTCY / `x/staking-tcy`), falling back to `bonded` only when nothing is held.

## Related (worth verifying)
`DEFI_ONLY_THORCHAIN_DENOMS` in `TokenRepository.kt` currently excludes `x/staking-ruji`, but the on-chain denom appears to be `x/staking-x/ruji` (see iOS #4318). Worth checking whether the receipt is matched/excluded under the correct denom on Android.

## Notes
Found while implementing Send-from-DeFi for sRUJI in vultisig-windows.

aminsato:
## What

Fixes #5258 — **DeFi → Staked → "Staked RUJI"** showed `0` (or a wrong value) even when the vault held sRUJI staking-receipt tokens on-chain.

## Root causes

1. **`bonded` is unreliable.** `ThorChainApi.getRujiStakeBalance` read the staked amount from the Rujira GraphQL `bonded` field, which can report `0` even when `x/staking-x/ruji` receipt tokens are held on-chain.
2. **Wrong `stakingV2` entry.** The code used `stakingV2.firstOrNull()` with no filter. The GraphQL returns multiple positions (e.g. **TCY and RUJI**), so it often grabbed the TCY position (bonded `0`) and reported it as *Staked RUJI* — corrupting amount, rewards, and APR.
3. **Denom exclusion never matched.** `DEFI_ONLY_THORCHAIN_DENOMS` listed only `x/staking-ruji`, but the on-chain receipt denom is `x/staking-x/ruji`, so the receipt leaked into the wallet token list.

## Fix

- Select the **RUJI** position from `stakingV2` by symbol instead of taking the first.
- Prefer the **on-chain `x/staking-x/ruji` receipt balance** (the source of truth for what the vault holds), falling back to `bonded` only when the balance read **fails** — a successful zero stays zero. Mirrors the sTCY / `x/staking-tcy` pattern.
- Correct `DEFI_ONLY_THORCHAIN_DENOMS` to cover both `x/staking-ruji` and `x/staking-x/ruji`.

## Cross-platform parity

Matches the intended **vultisig-windows #4337** logic (`rujiStakeService.ts`): read the receipt via the balances endpoint and use `heldAmount ?? bondedAmount` with failure-only fallback; the receipt raw amount is displayed with RUJI decimals. Sibling of iOS #4794. Windows still uses `stakingV2[0]` — this PR additionally filters for the RUJI position so rewards/APR/ticker are read from the correct entry (displayed amount is unchanged in the common receipt path).

## Verification

Verified against a **live thornode** account: the receipt denom is exactly `x/staking-x/ruji`; a vault with `TCY(bonded 0)` + `RUJI(bonded 7875733)` positions holds receipt `41952462`.

## Test plan

- [x] 5 new unit tests in `ThorChainApiImplTest` (URL-routing mock: GraphQL stake POST vs Cosmos balances GET):
  - prefers receipt over bonded when receipts are held
  - keeps a successful zero receipt as zero (does not fall back to bonded)
  - falls back to bonded only when the balances read fails
  - selects the RUJI position (not the leading TCY) on fallback
  - prefers receipt over RUJI bonded even with TCY listed first
- [x] `./gradlew :data:testDebugUnitTest --tests ThorChainApiImplTest` → 16/16 green
- [ ] On-device check: DeFi → Staked → "Staked RUJI" shows the real staked amount

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RUJI staking balance accuracy by preferring on-chain receipt balances when available.
  * Preserved genuine zero receipt balances without falling back to bonded amounts.
  * Added safer fallbacks to bonded values if on-chain balance reads fail or are unparseable.
  * Correctly selects the RUJI staking entry when multiple assets appear in the same staking list.
  * Updated DeFi-only RUJI denom exclusions to cover both legacy and receipt denom spellings.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
@@ -438,14 +438,25 @@ constructor(
             throw Exception("Could not fetch balances: ${response.errors}")
         }
 
-        val stake = response.data?.node?.stakingV2?.firstOrNull() ?: return RujiStakeBalances()
+        // stakingV2 can hold several positions (e.g. TCY and RUJI); pick the RUJI one rather than
+        // the first, or the amount/rewards get read from an unrelated position.
+        val stake =
+            response.data?.node?.stakingV2?.firstOrNull {
+                it.bonded.asset.metadata?.symbol.equals(RUJI_STAKE_SYMBOL, ignoreCase = true)
+            }
 
-        val stakeAmount = stake.bonded.amount.toBigIntegerOrNull() ?: BigInteger.ZERO
-        val stakeTicker = stake.bonded.asset.metadata?.symbol ?: ""
-        val rewardsAmount = stake.pendingRevenue?.amount?.toBigIntegerOrNull() ?: BigInteger.ZERO
-        val rewardsTicker = stake.pendingRevenue?.asset?.metadata?.symbol ?: DEFAULT_REWARDS_TICKER
+        val bondedAmount = stake?.bonded?.amount?.toBigIntegerOrNull() ?: BigInteger.ZERO
+        // The on-chain x/staking-x/ruji receipt is the source of truth for what the vault holds;
+        // the GraphQL `bonded` field can report 0 even when receipts are held. Prefer the receipt
+        // (a successful zero stays zero), falling back to `bonded` only when the balance read
+        // failed. Mirrors vultisig-windows #4337.
+        val stakeAmount = readRujiReceiptBalance(address) ?: bondedAmount
+
+        val stakeTicker = stake?.bonded?.asset?.metadata?.symbol ?: RUJI_STAKE_SYMBOL
+        val rewardsAmount = stake?.pendingRevenue?.amount?.toBigIntegerOrNull() ?: BigInteger.ZERO
+        val rewardsTicker = stake?.pendingRevenue?.asset?.metadata?.symbol ?: DEFAULT_REWARDS_TICKER
         val apr =
-            runCatching { (stake.pool?.summary?.apr?.value ?: "0.0").toDouble() }.getOrDefault(0.0)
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5263, opened 2026-07-12
fix(ton): surface action-phase failures instead of false Confirmed

### LINKED ISSUE #5249: TON: IGNORE_ACTION_PHASE_ERRORS on every transfer + aborted-only status check — failed transfers (incl. jettons) show Confirmed
## Problem — every TON transfer is signed to ignore its own failure

All TON send paths OR `IGNORE_ACTION_PHASE_ERRORS` into the send mode: `TonHelper.kt:165-175` (`calculateSendMode`, used by native :115 and TonConnect :85) and hardcoded for jettons at `TonHelper.kt:149-151` (`PAY_FEES_SEPARATELY | IGNORE_ACTION_PHASE_ERRORS`, added "to prevent retry loops").

Consequences:
1. **Native TON:** if the action phase fails (e.g. balance insufficient for value+fees under PAY_FEES_SEPARATELY), the wallet tx completes **un-aborted** — seqno consumed, no transfer executed. `TonStatusProvider.kt:16-21` inspects only `description.aborted` on the external wallet tx, so the user sees **Confirmed** while no funds moved.
2. **Jettons:** the transfer failure happens in the jetton-wallet contract's SEPARATE transaction, which the status check never inspects (no out_msgs / action-result / jetton-tx follow-up) — same false Confirmed.

## Fix direction
- Drop `IGNORE_ACTION_PHASE_ERRORS` where safe (solve the retry-loop concern via the status check instead), or
- Make the status provider outcome-aware: check the action-phase result / out_msgs on the wallet tx, and for jettons follow the jetton-wallet tx before declaring success.

## Context
Native parity audit 2026-07-10 (adversarially verified, S2). False-success class — sibling of vultisig-sdk#1119. Report: `native-parity-matrix-report-2026-07-10.md`.

aminsato:
## Problem

Every TON send is signed with `IGNORE_ACTION_PHASE_ERRORS` (`TonHelper.calculateSendMode` + the hardcoded jetton mode). When a transfer can't be paid for — e.g. `value + fees` exceeds balance under `PAY_FEES_SEPARATELY` — the action is **silently skipped** instead of aborting the transaction. The wallet tx lands **un-aborted** with the seqno consumed but **no funds moved**.

`TonStatusProvider` inspected only `description.aborted`, so the user was shown a false **Confirmed** while nothing was actually transferred.

Fixes #5249.

## Fix

Make the status provider **outcome-aware**. This is a **read-side change only** — the signing path (`TonHelper`) is deliberately left untouched so the TSS pre-image hash stays byte-identical across co-signing devices (changing send modes would diverge the hash and 404 joiners).

- Model the compute (`compute_ph`) and action phases of the transaction description.
- Report **Failed** when:
  - the compute phase reverted — exit code not in `{0, 1}` (matching the iOS/SDK resolver's TVM success convention), or
  - the action phase failed / ran out of funds / skipped an action (`success == false`, `no_funds == true`, or `skipped_actions > 0`).
- `aborted` handling and the "still indexing → keep polling" behaviour are unchanged, so existing successful flows are unaffected.

### Cross-platform note
Neither iOS nor the SDK currently inspect the **action** phase — they check only `aborted` + `compute_ph.exit_code` — so this also closes a cross-platform gap (sibling `vultisig-sdk#1119`).

### Known limitation
This catches native false-success and jetton sends that fail at the **wallet** level (can't afford the 0.08 TON carrier). It does **not** follow the jetton-wallet contract's *separate* transaction, so a failure inside the jetton-wallet (e.g. insufficient jetton balance → bounce) is out of scope here and would need out-message following.

## Test plan

- `./gradlew :data:testDebugUnitTest --tests "*.TonStatusProviderTest"` — 13 tests, all green.
- New coverage:
  - the exact bug: `aborted: false` + `no_funds: true` → **Failed**
  - compute revert (exit 37) → Failed; exit 0/1 → Confirmed
  - action `skipped_actions > 0` / `success == false` → Failed
  - fully-successful compute + action → Confirmed
  - a **verbatim production toncenter payload** decoded through the real kotlinx pipeline, proving the `compute_ph`/`action` `@SerialName` mappings match live data and unknown keys are tolerated.

### Reproduction note
Not reproducible via the native send form: it reserves a flat 0.05 TON and blocks `amount > balance − 0.05`, so the action phase always has ample headroom. The bug class manifests on paths that bypass that guard (TonConnect/dApp-supplied amounts, or a fee under-estimate). Verified via the unit tests + live-JSON parse rather than a device screenshot, which is the appropriate bar for a read-side status change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TON transaction status detection for compute-phase reverts and action-phase failures.
  * Transactions with insufficient funds, skipped actions, or unsuccessful actions are now correctly marked as failed.
  * Added clearer failure details, including relevant compute exit codes.
  * Successful TON transactions continue to be confirmed when all phases complete successfully.

* **Tests**
  * Expanded coverage for TON transaction phase outcomes and real-world response handling.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/chains/ton/TonApiDto.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/chains/ton/TonApiDto.kt
@@ -201,4 +201,28 @@ data class TransactionJson(
 data class TonTransactionDescriptionJson(
     @SerialName("aborted") val aborted: Boolean? = null,
     @SerialName("destroyed") val destroyed: Boolean? = null,
+    @SerialName("compute_ph") val computePhase: TonComputePhaseJson? = null,
+    @SerialName("action") val actionPhase: TonActionPhaseJson? = null,
+)
+
+/**
+ * TVM compute phase of a TON transaction. A `null` [exitCode] means the message carried no compute
+ * phase (a plain transfer to a wallet), which is not a failure. Exit codes 0 and 1 are the TVM
+ * success conventions; any other code is a revert.
+ */
+@Serializable data class TonComputePhaseJson(@SerialName("exit_code") val exitCode: Int? = null)
+
+/**
+ * Action phase of a TON transaction — where the wallet contract actually emits its outgoing
+ * transfer message(s). Because every Vultisig TON send is signed with `IGNORE_ACTION_PHASE_ERRORS`,
+ * an action that can't be carried out (e.g. insufficient balance for value+fees) is silently
+ * skipped instead of aborting the transaction, so these fields — not `aborted` — are what reveal a
+ * transfer that never moved any funds.
+ */
+@Serializable
+data class TonActionPhaseJson(
+    @SerialName("success") val success: Boolean? = null,
+    @SerialName("no_funds") val noFunds: Boolean? = null,
+    @SerialName("result_code") val resultCode: Int? = null,
+    @SerialName("skipped_actions") val skippedActions: Int? = null,
 )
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5265, opened 2026-07-12
feat(solana): sign and broadcast dApp transaction batches

### LINKED ISSUE #5238: [Fix] Solana signAllTransactions ceremony completes then hard-fails assembling multi-transaction output
# [Fix] Solana signAllTransactions ceremony completes then hard-fails assembling multi-transaction output

## Problem
When a dApp calls `signAllTransactions`/`signAndSendAllTransactions` with more than one raw Solana transaction, `SolanaHelper.kt`'s `getPreSignedImageHash` already loops over all N raw transactions and includes all their hashes in the keysign ceremony, so the full multi-device TSS ceremony — including the user's physical approval on their phone — completes successfully. But `getSignedTransaction` has `require(signSolana.rawTransactions.size == 1)` and only assembles the first transaction, so the ceremony's result can never be delivered: the whole batch is discarded with an opaque "Signing Error" after the user has already completed the entire approval flow.

## Expected Behavior
At minimum, detect the N>1 case before the ceremony starts (not after), and show a clear "batch signing not yet supported" message, so the user is never asked to approve a signing request that can't succeed.

## Steps to Reproduce
1. From a dApp connected via the browser extension, call `signAllTransactions`/`signAndSendAllTransactions` with more than one transaction.
2. Complete the full multi-device keysign ceremony, including approving on the paired phone.
3. Observe the ceremony completes, then assembly fails with an opaque Signing Error and the whole batch is discarded.

## Solution
Move the existing `require(rawTransactions.size == 1)` guard to fire at request-acceptance time (before starting peer discovery/the ceremony), with a clear, actionable error message, instead of failing only after the ceremony completes.

### Must Do
- [ ] Detect `signSolana.rawTransactions.size > 1` before starting the keysign ceremony and show a clear "not yet supported" error
- [ ] Add a test asserting the guard fires before ceremony start for a multi-transaction payload

### Must NOT Do
- Don't attempt full batch-signing support in this ticket (see follow-up comment for that plan)
- Don't change the single-transaction `signSolana` path

## Acceptance Criteria
- [ ] New test passes
- [ ] A multi-transaction `signSolana` request is rejected with a clear message before the ceremony starts, not after

rkokhatskyi:
Closes #5238

## Summary

A dApp `signAllTransactions` / `signAndSendAllTransactions` request with more than one raw Solana transaction already completed the full keysign ceremony — `getPreSignedImageHash` hashes every transaction and all N hashes get signed — but assembly (`SolanaHelper.getSignedTransaction`) only delivered the first one, so the whole batch was discarded with an opaque Signing Error after the user had approved on the phone. This adds batch support end to end instead of the fail-fast guard the ticket originally scoped:

- `SolanaHelper.getSignedTransactions` assembles one signed transaction per raw transaction (empty batches rejected with a clear message); `SigningHelper.getSignedTransactions` routes `signSolana` payloads through it and wraps every other payload's single result, so no other chain changes behavior
- `BroadcastKeysignUseCase` broadcasts the batch sequentially in payload order, fail-fast, matching the extension's `chainPromises` semantics; the per-tx duplicate-broadcast recovery still applies, so racing the extension's copies of the same batch is safe
- The first hash stays the primary one driving the done screen and status polling; the remaining hashes are persisted to transaction history, each with its own explorer link
- The verify screen now shows every transaction of the batch (`Transaction.signSolana` is now `List<String>`); previously a joiner reviewing a 3-tx batch only saw the first transaction. `SignSolanaDisplayView` takes the raw-transaction list directly (matching its SignTon/SignSui siblings) and separates instruction rows with a hairline divider
- `signAllTransactions` (`skipBroadcast=true`) is untouched — the phone never assembles or broadcasts on that path

The wire format is unchanged (`SignSolana.raw_transactions` was already `repeated`), and the extension needs no changes — it compiles and returns its own copies to the dApp. iOS has the same single-transaction assembly guard (`Solana.swift`, `rawTransactions.count == 1`) and needs the equivalent fix for extension + iOS-phone pairings.

## Before / After

Verify screen on the joining device for a 2-transaction dApp batch. Before, only the first transaction of the batch was surfaced for review; now every transaction's instructions and raw bytes are shown, with a hairline divider separating the instruction rows.

| | Before | After |
|---|--------|-------|
| Collapsed (entry point) | before-collapsed | after-collapsed |
| Expanded | before-expanded | after-expanded |

## Testing

- `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest` — 3516 tests, green
- `./gradlew assembleDebug` — green
- `./gradlew :app:lintDebug` — green

New tests:
- `SolanaHelperTest` — empty batch rejected before any WalletCore call; single-transaction assembly keeps its strict one-tx contract
- `BroadcastKeysignBatchTest` — batch broadcasts every signed transaction in payload order, first hash primary + rest as `additionalTxHashes`, mid-batch failure propagates and skips the rest
- `KeysignViewModelApplyBroadcastResultTest` — every hash of a batch lands in transaction history with its own explorer link

Known follow-ups (out of scope here): dependent batches (tx2 consuming state created by tx1) can fail phone-side preflight before tx1 confirms — the shared Solana broadcast path would need a confirmation gap or skipPreflight for later batch items; the raw-transaction display flattens instruction numbering across the batch without per-transaction headers; the done screen shows only the primary hash.

## Summary by CodeRabbit

* **New Features**
  * Added support for signing and broadcasting multiple Solana transactions from a single request.
  * Solana verification screens now display all transactions in a batch, with expandable details.
  * Additional transaction hashes are retained in transaction history with explorer links.

* **Bug Fixes**
  * Improved handling of empty or invalid Solana transaction batches.
  * Batch broadcasts now stop immediately if one transaction fails.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -202,6 +202,9 @@ class PreviewActivity : ComponentActivity() {
                     "choose_vault" -> SelectVaultTypeScreenPreview()
                     "content_row" -> ContentRowPreview()
                     "solana_display" -> SolanaDisplayPreview()
+                    "solana_batch_verify" -> SolanaBatchVerifySendPreview(transactionCount = 2)
+                    "solana_batch_verify_before" ->
+                        SolanaBatchVerifySendPreview(transactionCount = 1)
                     "ton_display_single" -> TonDisplayPreview(messageCount = 1)
                     "ton_display_multi" -> TonDisplayPreview(messageCount = 4)
                     "verify_ton_jetton_before" -> VerifyTonJettonPreview(decoded = false)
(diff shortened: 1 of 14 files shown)


rkokhatskyi:
@coderabbitai Moved the additionalTxHashes persistence out of the primary-hash guard in d02253c6c — each batch hash is now saved to history with its own explorer link even when the primary hash is null, and the terminal-state handling for the no-polling/no-hash cases is unified. Added a test pinning it (broadcasted batch with null primary hash still persists additional hashes).

rkokhatskyi:
@coderabbitai Strengthened the null-primary test in e02220756 — it now captures the recorded history data and asserts the additional hash carries its own explorer link, matching the batch test above it.

rkokhatskyi:
# Testing and screenshots

https://solscan.io/tx/4XSpeXPDjtgU7qQhKEt5FUh8kt4WjQgvaLEzUgrp2tv4Cva2nKHSsuUjcUG1QPk59KakbjUPkFJ75h82dHQm1jzJ
Tx confirmation Android 
image


Instructions and DApp 

image

johnnyluo (approved):
LGTM

PR #5266, opened 2026-07-12
fix(broadcast): on-chain verify Cosmos/QBTC code-32 before reporting success

johnnyluo:
## Summary
Follow-up to #5254, closing the remaining fixable item of #5250 (item 5).

The joined-device duplicate-broadcast recovery in `BroadcastKeysignUseCase` trusted a Cosmos `code=32` (account sequence mismatch) **blindly** and returned the locally computed hash as success — with no on-chain check. If the sequence was consumed by an *unrelated* tx (not our byte-identical duplicate), the user saw a success screen and an explorer link for a transaction that never landed.

**Fix:** move the recovery down into `BroadcastTxUseCase`, wrapping Cosmos/QBTC in the same `recoverIfAlreadyBroadcast` on-chain-verify path every other chain already uses (`getTxStatus(hash) != null`, 3× retry with backoff). A `code=32` means the peer's byte-identical tx **already committed and advanced the sequence**, so a hash lookup cleanly separates "our tx landed" from "a different tx consumed the sequence".

With the verify one layer down, the blind joined-device path had no remaining signal to handle, so `broadcastOrRecover` / `recoverJoinedDeviceBroadcast` / `isDuplicateBroadcast` are removed (−180 lines). Every chain now satisfies the issue's invariant: *any dedup heuristic must be followed by an on-chain existence check of the actual tx hash before reporting success* — EVM inside `sendTransaction` (`isTxMined`), everything else via `recoverIfAlreadyBroadcast`.

### Still deferred
- **Sui (#5250 item 4)**: WalletCore's `Sui.SigningOutput` exposes no transaction digest (the Rust core doesn't compute one either), so `transactionHash` is blank and recovery structurally cannot verify a hash pre-broadcast. Needs SDK support to expose the digest. Issue #5250 should stay open, scoped down to just Sui.

## Issue
Refs #5250

## Test Plan
- New `BroadcastTxUseCaseTest` cases: Cosmos recovers with local hash when a `code=32` rejection is already on-chain; rethrows (after 3 verify attempts) when our tx is **not** on-chain
- Removed obsolete `KeysignViewModelBroadcastRecoveryTest` (tested the deleted `broadcastOrRecover`)
- [x] `./gradlew :data:testDebugUnitTest --tests "*BroadcastTxUseCaseTest"` passes
- [x] `./gradlew :app:compileDebugKotlin` succeeds
- [ ] Lint passes (CI)
- [ ] Debug build succeeds (CI)

On chain tx link: 
https://finder.terra.money/classic/tx/4E758480309055634289660C6D2693288DCCE0AEA4C6895E38D0374C6C1D65B1

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved broadcast recovery for joined-device “already broadcast” and sequence-mismatch scenarios by verifying on-chain status.
  - ThorChain recovery now treats multiple terminal outcomes (e.g., confirmed/refunded/failed) as success, while non-terminal states do not.
  - Cosmos recovery now requires LCD transaction execution `code == 0`, and rethrows the original error when recovery can’t be verified.
  - Updated Cosmos transaction status parsing to treat missing `code` fields from LCD/gateway responses as success (`0`).

- **Tests**
  - Expanded sequence-mismatch recovery test coverage across ThorChain and Cosmos (Kujira), including retry behavior and success/failure conditions.
  - Removed an outdated broadcast-recovery test suite.


data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -135,7 +135,17 @@ constructor(
...
+                    broadcast = {
+                        cosmosApi.broadcastTransaction(tx.rawTransaction).orKnownHash(tx)
+                    },
+                    // A code=32 sequence mismatch on a joined co-signer means the peer's
+                    // byte-identical tx already advanced the account sequence in a committed
+                    // block. The LCD returns 404 (null) until our hash is committed, so a non-null
+                    // status proves our tx — not an unrelated one — consumed the sequence.
+                    verify = { hash -> cosmosApi.getTxStatus(hash) != null },
+                )
             }
```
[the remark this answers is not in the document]
johnnyluo: Good catch — fixed in 968fbfe. A Cosmos tx can be committed yet fail execution, so verify now requires `txResponse.code == 0` before recovering the byte-identical duplicate; an included-but-failed tx surfaces as an error instead of a false success. Added a test covering the non-zero-code case.

johnnyluo:
Re the nitpick ("recovery kicks in for any exception, not only code=32"): leaving as-is. This is how the shared `recoverIfAlreadyBroadcast` helper works for every other chain in the file (BTC/Solana/Polkadot/Ton/Ripple/Tron/Cardano) — catch, then verify the hash on-chain. A genuine failure (insufficient funds, etc.) never lands, so `getTxStatus` returns null/non-zero and it's rethrown — not masked, just delayed ~4s by the verify loop, same as every sibling chain. Special-casing only Cosmos to pre-filter on `code=32` would break that consistency for a trivial latency win on a rare path.

app/src/main/java/com/vultisig/wallet/data/usecases/BroadcastKeysignUseCase.kt
```diff
@@ -163,7 +163,7 @@ constructor(
                 nonceAcc = nonceAcc,
             )
 
-        val txHash = broadcastOrRecover(chain, signedTx, isInitiatingDevice)
+        val txHash = broadcastTx(chain = chain, tx = signedTx)
```
rkokhatskyi, line 166: ThorChain and MayaChain also throw `CosmosBroadcastException` and were recovered by the deleted `broadcastOrRecover`'s generic `isSequenceMismatch` check. Their `BroadcastTxUseCase` branches stay bare, unwrapped in `recoverIfAlreadyBroadcast`, so a joined device now hard fails on the exact race this PR just fixed for Cosmos

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -135,7 +135,18 @@ constructor(
...
+                    tx = tx,
+                    broadcast = {
+                        cosmosApi.broadcastTransaction(tx.rawTransaction).orKnownHash(tx)
+                    },
+                    // A code=32 sequence mismatch on a joined co-signer means the peer's
+                    // byte-identical tx already advanced the account sequence in a committed
+                    // block. The LCD returns 404 (null) until our hash is committed. A committed tx
+                    // can still have failed execution (non-zero code), so require code==0 — an
+                    // included-but-failed tx moved no funds and must not be reported as success.
+                    verify = { hash -> cosmosApi.getTxStatus(hash)?.txResponse?.code == 0 },
```
rkokhatskyi, line 160: `TxResponse.code` is non-nullable with no default (`CosmosTxStatusJson.kt:17`), unlike the sibling broadcast model's `code: Int? ?: COSMOS_TX_SUCCESS_CODE`. If an LCD ever omits a zero-value `code`, deserialization throws, retries exhaust, and a landed tx gets reported as failed, needs runtime confirmation

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/data/usecases/BroadcastKeysignUseCase.kt
johnnyluo: Good catch — real regression. THORChain and MayaChain are now wrapped in `recoverIfAlreadyBroadcast` (commit 29901a3), so a joined-device sequence-mismatch race verifies the peer's byte-identical tx actually committed via `TransactionStatusRepository.checkTransactionStatus == Confirmed` before returning the local hash — otherwise it rethrows. Added a THOR recover test and a Maya rethrow-when-not-confirmed test.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
johnnyluo: Fixed in 29901a3. `CosmosTxStatusJson.TxResponse.code` now defaults to 0 — since the Cosmos LCD (gRPC-gateway) omits zero-value fields, a successfully-executed (code=0) tx would otherwise miss the field and throw `MissingFieldException`, exhaust retries, and be reported as not-on-chain. Defaulting to 0 is transparent to the existing consumers (`CosmosStatusProvider`, the verify lambda) which already treat 0 as success.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -243,6 +271,10 @@ constructor(
         return false
     }
 
+    private suspend fun isConfirmedOnChain(hash: String, chain: Chain): Boolean =
+        transactionStatusRepository.checkTransactionStatus(hash, chain) ==
+            TransactionResult.Confirmed
```
rkokhatskyi: `isConfirmedOnChain` requires `TransactionResult.Confirmed`, but `ThorMayaChainStatusProvider.mapAction` returns `Refunded` for a landed refunded swap, a common outcome. Retries exhaust, the sequence mismatch is rethrown, and a landed tx shows as failed, risking a double send retry, any terminal Midgard result should count as landed

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
johnnyluo: Good catch — fixed in 9da48d5. Renamed `isConfirmedOnChain` → `isLandedOnChain`; it now treats any terminal on-chain outcome (`Confirmed`, `Refunded`, or `Failed`) as landed, since all three prove the peer's byte-identical tx committed and consumed the sequence. Only `Pending`/`NotFound`/`TimedOut` propagate the rejection, so a landed-but-refunded swap no longer surfaces as failed / risks a double send. Added a THOR refunded-recovers test.

rkokhatskyi (approved):
LGTM

PR #5267, opened 2026-07-13
fix(cardano): emit Conway-era 4-element tx envelope with is_valid flag

### LINKED ISSUE #5268: Cardano: native ADA sends rejected by Conway-era nodes (3-element tx envelope missing is_valid flag)
## Problem

Submitting a native ADA send fails at the node with:

```json
{
  "jsonrpc": "2.0",
  "method": "submitTransaction",
  "error": {
    "code": 3005,
    "message": "Failed to submit the transaction in the current era. This may happen when trying to submit a transaction near an era boundary (i.e. at the moment of a hard-fork). Retrying should help.",
    "data": { "ledgerEra": "conway", "queryEra": "dijkstra" }
  },
  "id": 1
}
```

## Root cause

WalletCore's `TransactionCompiler.compileWithSignatures` emits the **legacy 3-element** Cardano transaction envelope:

```
[ transaction_body, transaction_witness_set, auxiliary_data ]   // 0x83
```

Conway-era ledgers require the **4-element** form with an `is_valid` boolean (introduced in Alonzo):

```
[ transaction_body, transaction_witness_set, is_valid, auxiliary_data ]   // 0x84
```

The submitted CBOR starts with `0x83`, so the node rejects it as not valid for the current era.

Note: the SwapKit Cardano path already emits the correct 4-element envelope (`0x84`); only the native send path (`CardanoHelper.getSignedTransaction`) was affected.

## Fix

Splice `is_valid = true` (`0xf5`) into the signed envelope before broadcast, producing the 4-element form. This is safe post-signing — the vkey signature and tx_id cover only the transaction body (element 0), which stays byte-for-byte unchanged.

Fixed in #5267.

johnnyluo:
## Summary
- Native ADA sends failed to submit on Conway-era nodes: WalletCore's `compileWithSignatures` emits the legacy 3-element Cardano envelope `[body, witness_set, aux_data]` (`0x83`), which current nodes reject.
- Added `CardanoUtils.addIsValidFlag` to splice `is_valid = true` (`0xf5`) into the envelope, producing the required 4-element form `[body, witness_set, is_valid, aux_data]` (`0x84`), and applied it in `CardanoHelper.getSignedTransaction` before broadcast.
- Safe post-signing: the vkey signature and tx_id cover only the body (element 0), which is left byte-for-byte untouched. Already-4-element envelopes (e.g. the SwapKit path) pass through unchanged.

## Issue
Closes #5268

## Test Plan
- [x] `CardanoUtilsTest` — upgrades a real rejected envelope, tx_id unchanged, idempotent on 4-element input
- [x] Manual: sign & broadcast a native ADA send against a Conway/Dijkstra node

On chain tx link: https://cardanoscan.io/transaction/c1160d3a7e19f3c097d04703b4d10f9827734cf37e56e8ec1936697fea26bf4b

## Summary by CodeRabbit

* **Bug Fixes**
  * Cardano transactions are now encoded in the current format with the required validity flag.
  * Transaction hashes remain consistent while signed transaction data is updated.
  * Improved handling of tagged transaction data and already-updated transaction formats.

* **Tests**
  * Added coverage for legacy, current, and tagged Cardano transaction envelopes.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
@@ -243,10 +243,12 @@ object CardanoHelper {
             )
 
         val output = Cardano.SigningOutput.parseFrom(compileWithSignature).checkError()
-        var transactionHash =
-            CardanoUtils.calculateCardanoTransactionHash(output.encoded.toByteArray())
+        // WalletCore emits the legacy 3-element envelope; Conway-era nodes require the 4-element
+        // form with an is_valid flag. Splicing it in doesn't touch the signed body (element 0).
+        val encoded = CardanoUtils.addIsValidFlag(output.encoded.toByteArray())
+        val transactionHash = CardanoUtils.calculateCardanoTransactionHash(encoded)
         return SignedTransactionResult(
-            rawTransaction = output.encoded.toByteArray().toHexString(),
+            rawTransaction = encoded.toHexString(),
             transactionHash = transactionHash,
         )
     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoUtils.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/CardanoUtils.kt
@@ -53,6 +53,39 @@ object CardanoUtils {
         return Bech32.encode("addr", addressData)
     }
 
+    /**
+     * Upgrades a legacy 3-element Cardano transaction envelope `[body, witness_set,
+     * auxiliary_data]` (emitted by WalletCore) to the Conway-era 4-element form `[body,
+     * witness_set, is_valid, auxiliary_data]` by splicing `is_valid = true` (0xf5) before the
+     * auxiliary_data. Safe post-signing: the vkey signature and tx_id cover only the body (element
+     * 0), which is left byte-for-byte untouched. Already-4-element envelopes are returned
+     * unchanged.
+     */
+    fun addIsValidFlag(encoded: ByteArray): ByteArray {
+        require(encoded.isNotEmpty()) { "empty Cardano transaction" }
+        when (encoded[0].toInt() and 0xFF) {
+            0x84 -> return encoded // already Conway-era 4-element array
+            0x83 -> {} // legacy 3-element array — needs is_valid
+            else -> error("unexpected Cardano tx array header: 0x%02x".format(encoded[0]))
+        }
+
+        val afterBody = findEndOfCBORItem(encoded, 1)
+        val afterWitness = findEndOfCBORItem(encoded, afterBody)
+
+        return ByteArray(encoded.size + 1).also { out ->
+            out[0] = 0x84.toByte()
+            System.arraycopy(encoded, 1, out, 1, afterWitness - 1) // body + witness_set
+            out[afterWitness] = 0xF5.toByte() // is_valid = true
+            System.arraycopy(
+                encoded,
+                afterWitness,
+                out,
+                afterWitness + 1,
+                encoded.size - afterWitness,
+            )
+        }
+    }
+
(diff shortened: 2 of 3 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
```diff
@@ -243,10 +243,12 @@ object CardanoHelper {
             )
 
         val output = Cardano.SigningOutput.parseFrom(compileWithSignature).checkError()
-        var transactionHash =
-            CardanoUtils.calculateCardanoTransactionHash(output.encoded.toByteArray())
+        // WalletCore emits the legacy 3-element envelope; Conway-era nodes require the 4-element
+        // form with an is_valid flag. Splicing it in doesn't touch the signed body (element 0).
+        val encoded = CardanoUtils.addIsValidFlag(output.encoded.toByteArray())
```
rkokhatskyi, line 248: `findEndOfCBORItem` has no case for CBOR tags (major type 6), so anything beyond plain ints, strings, arrays and maps throws here after the tx is already signed. `SwapKitCardanoSigner` a few files over already handles tags in its own CBOR walker, worth mirroring.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
```diff
@@ -243,10 +243,12 @@ object CardanoHelper {
             )
 
         val output = Cardano.SigningOutput.parseFrom(compileWithSignature).checkError()
```
rkokhatskyi, line 245: iOS and Windows never call `compileWithSignatures` for Cardano (`CardanoHelper.kt:238`), both hand-build the envelope instead because of a known `AddressV2::isValid` crash in that call, at the same wallet-core version Android pins. This PR only patches the output, not the call itself, needs runtime confirmation on Android's binding.

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
johnnyluo: Good catch — fixed in 52fb772. Added the major-type-6 (tag) case to `findEndOfCBORItem`, mirroring `SwapKitCardanoSigner`'s walker: skip the tag number's argument bytes, then recurse on the single tagged item. This matters for Conway-era set tag 258 (`0xd90102`), which would otherwise throw in `addIsValidFlag` *after* signing. Also hardens `extractCardanoTransactionBody` (same shared walker). Added a test that walks a tag-258 set embedded in the witness.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/CardanoHelper.kt
johnnyluo: Thanks — I dug into this. This PR does **not** touch the `compileWithSignatures` call; it only post-processes `output.encoded`. The call itself is pre-existing — it's been the Android Cardano signing path since PR #2024 (`eed0d334`), so it isn't new risk this PR introduces.

On the `AddressV2::isValid` crash: Android's pinned wallet-core binding doesn't hit it on this path. The runtime evidence is the bug report that opened this PR — the submitted `cbor` (`83a4…` with a fully-formed witness set containing the vkey + 64-byte signature) is the *output* of `compileWithSignatures` on Android. If the crash affected Android's binding, there'd be no signed envelope to submit at all — signing would have crashed instead of producing bytes the node rejected purely for the era/is_valid reason. So the compile succeeds on Android; the only gap was the missing `is_valid` flag, which this PR adds.

Aligning Android onto the hand-built-envelope approach iOS/Windows use is a reasonable larger refactor, but out of scope here since the current call demonstrably works.

rkokhatskyi (approved):
LGTM

PR #4530, opened 2026-05-16
fix(swap): normalise THORChain secured-asset name to Thornode form

aminsato:
## Summary
- THORChain secured-asset coins (`btc-btc`, `eth-eth`, `eth-usdc-0xa0b8…`, etc.) were being sent to Thornode as the raw lowercase hyphenated `contractAddress`, which fails with `bad to asset: chain id len is less than 3: invalid request`. Multi-segment EVM secured assets like `eth-usdc-0x…` fell to the else branch and became `THOR.USDC` — also wrong (Thornode expects the L1 chain prefix, not THOR).
- `swapAssetName()` for the secured-asset branch now derives `CHAIN.SYMBOL[-0xHex]`: chain and symbol uppercased, EVM hex tail preserved verbatim so EIP-55 checksum casing survives the round-trip.

## Mapping

| Input contractAddress | Before | After |
|---|---|---|
| `btc-btc` | `btc-btc` | `BTC.BTC` |
| `eth-eth` | `eth-eth` | `ETH.ETH` |
| `bsc-bnb` | `bsc-bnb` | `BSC.BNB` |
| `eth-usdc-0xa0b86991…` | `THOR.USDC` (else branch) | `ETH.USDC-0xa0b86991…` |
| `ruji` | `THOR.RUJI` ✅ | `THOR.RUJI` (unchanged) |

## Cross-platform parity
- **SDK** — same bug class, fixed upstream by [vultisig-sdk#455](https://github.com/vultisig/vultisig-sdk/pull/455) (v0.22.7, 2026-05-13). This PR mirrors `toNativeSwapAsset`'s test matrix.
- **iOS** — not affected; already uses dedicated `THORChainHelper.securedAssetChain/Symbol`.
- **Windows / extension** — picks up the SDK fix on next bump.

## Closes
- Closes #4480
- Closes #4489 (duplicate, closed by @johnnyluo in favour of #4480)

## Test plan
- [x] `./gradlew :data:ktfmtFormat :app:ktfmtFormat` — clean
- [x] `./gradlew :app:compileDebugKotlin` — green
- [x] `./gradlew :data:testDebugUnitTest` — green, 4 new tests + 3 updated assertions in `CoinSwapAssetNameTest`
- [ ] Manual: open Swap, From = THORChain secured BTC (`BTC.BTC`) → To = RUNE, confirm quote succeeds (no `chain id len is less than 3` error)
- [ ] Manual: same for ETH secured USDC → RUNE / BTC (covers #4489's reported failure)
[...]

PR #5271, opened 2026-07-13
Make THORChain secured assets discoverable in the swap picker

### LINKED ISSUE #5251: Swap: make THORChain secured assets discoverable + swappable as destinations
## Goal

In the swap view, a user can discover and swap **into any** THORChain secured asset — not only the ones already held in the vault.

## Why this is small

Swapping into a secured asset already works end-to-end: THORChain mints the secured asset as the swap's settlement output (quote memo `=:ETH-USDC:thor1…`). Android already handles this:
- Provider assignment already covers secured assets: `SwapProviderTable.providersFor` maps `Chain.ThorChain → {THORCHAIN, MAYA}` unconditionally (`data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapProviderTable.kt:119`).
- Swap signing handles secured via `ThorChainHelper` / `ThorchainSwapHelper` (`isSecuredAsset` / `securedAssetChain` / `securedAssetSymbol`).

The only missing piece is **discoverability**.

## Current state (verified)

The swap asset picker (`SelectAssetViewModel`, `app/src/main/java/com/vultisig/wallet/ui/screens/select/SelectAssetViewModel.kt`) builds its list from held tokens + a `getChainTokens(chain)` catalog (non-held entries are shown `isDisabled=true` and added via `enableTokenUseCase` on selection). The static `Coins.kt` catalog contains **0 secured-asset denoms** (verified — no `btc-btc` / `eth-usdc-0x…` entries). So only secured assets the user already holds appear; unheld ones are undiscoverable.

## Scope / acceptance

- Secured assets are selectable swap **destinations** in the picker, clearly labeled (e.g. a "Secured" group/badge).
- Add secured assets to the THORChain chain-tokens catalog (`getChainTokens(Chain.ThorChain)` / `Coins.kt`) so they surface as addable (`isDisabled`) destinations, then flow through the existing enable → swap path.
- Selecting one routes through the **existing** swap → mint-on-settle path — **no new keysign/tx type**; reuse the existing pending → tx-history pipeline.
- Same-underlying-asset case (hold BTC → want secured BTC) routes to **SECURE+ mint** (existing Deposit flow) rather than a wasteful 1:1 pool swap.
- Zero-balance secured destinations are allowed (you are minting, not spending).

## Source of truth

Populate from a **dynamic** secured-asset list derived from THORChain pools (complete, never stale); static list only as an offline fallback. The shared list is owned by the SDK issue in the coordinated set below — Android should consume it rather than hardcode a list.

## Out of scope

Rujira / FIN orderbook trading (separate track).

Suggested owner: Android (Amin / Roman).

## Coordinated set (all-platform, 1:1)
- iOS: vultisig/vultisig-ios#4788
- Android: vultisig/vultisig-android#5251
- SDK (shared source of truth + Windows/ext): vultisig/vultisig-sdk#1130

SDK owns the dynamic secured-asset list; iOS + Android consume it.

rkokhatskyi:
## Summary

Secured assets already worked end-to-end as swap destinations (THORChain mints them on settlement), but they only appeared in the swap picker if the vault already held a balance — the THORChain token catalog had no secured-asset entries at all. This adds a live catalog derived from THORChain's pool list (`ThorChainSecuredAssetRepository`, with a small static offline fallback), wires it into `GetChainTokensUseCase` for `Chain.ThorChain`, and tags secured rows with a "Secured" badge in the picker.

Also: picking a secured asset whose underlying matches the currently-selected swap source (e.g. holding BTC, picking secured BTC as destination) now routes to the existing SECURE+ deposit mint flow instead of a wasteful 1:1 pool swap.

## Screenshots

Swap destination picker, THORChain tab.

| Before | After |
|--------|-------|
| before | after |

## Testing

- `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.repositories.ThorChainSecuredAssetRepositoryImplTest" --tests "com.vultisig.wallet.data.usecases.chaintokens.GetChainTokensUseCaseTest"`
- `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.swap.*"` (261 tests, including the same-underlying-asset predicate test and the full existing `SwapFormViewModelTest` suite)
- `./gradlew :data:lintDebug :app:lintDebug`
- `./gradlew :app:ktfmtFormat :data:ktfmtFormat`
- Manual verification via `PreviewActivity` on an emulator (screenshots above)

Closes #5251

## Summary by CodeRabbit

* **New Features**
  * Added secured-asset discovery and display in the asset selector, including a localized “Secured” badge (multiple languages) and secured source-chain pills.
  * Extended debug previews with secured-asset “before” and “after” modes.
  * Updated swap token selection to recognize secured THORChain forms and route accordingly.
  * Updated ThorChain token lists to include secured-asset entries.
* **Bug Fixes**
  * Improved secured-underlying matching and token de-duplication to prevent incorrect hiding/selection.
* **Tests**
  * Added coverage for secured-asset parsing, refresh/caching, underlying-to-secured matching, and merge/override behavior.


app/src/test/java/com/vultisig/wallet/ui/models/swap/IsUnderlyingOfSecuredAssetTest.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        assertFalse(eth.isUnderlyingOfSecuredAsset(securedBtc))
+    }
+
+    @Test
+    fun `a ticker outside the SECURE+ eligible set never matches`() {
+        val sol = nativeCoin(Chain.Solana, ticker = "SOL")
+        val securedSol = securedCoin(ticker = "SOL", contractAddress = "avax-sol-0xfe6b19286")
+
+        assertFalse(sol.isUnderlyingOfSecuredAsset(securedSol))
```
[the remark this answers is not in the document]
rkokhatskyi: Good catch — fixed. The old fixture used `avax-sol-0xfe6b19286` as the secured coin's contract, so the assertion was failing on the chain-code mismatch (AVAX vs SOL) as much as on eligibility. Changed to `sol-sol` so it isolates `isSecuredAssetEligible()`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorChainSecuredAssetRepository.kt
```diff
@@ -0,0 +1,92 @@
...
+    suspend fun getSecuredAssetCoins(): List<Coin>
+}
+
+@Singleton
+internal class ThorChainSecuredAssetRepositoryImpl
+@Inject
+constructor(private val thorChainApi: ThorChainApi) : ThorChainSecuredAssetRepository {
+
+    @Volatile private var cache: List<Coin> = emptyList()
+    @Volatile private var lastRefreshMs: Long = 0L
```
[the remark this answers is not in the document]
rkokhatskyi: Agreed, this was a real gap against the ticket's "static list only as an offline fallback" requirement. Fixed: the repository now seeds/falls back to a small static catalog built from the existing `ThorChainHelper.SECURE_ASSETS_TICKERS` (BTC/ETH/BCH/LTC/DOGE/AVAX/BNB) instead of an empty list, used only until a live pool fetch first succeeds — a later failure still keeps the last-good live snapshot rather than reverting to it. Added tests for both the cold-start fallback and the BSC chain-code (bsc-bnb, not bnb-bnb).

data/src/test/kotlin/com/vultisig/wallet/data/repositories/ThorChainSecuredAssetRepositoryImplTest.kt
```diff
@@ -0,0 +1,127 @@
...
+    fun `falls back to the static catalog when the very first fetch fails`() = runTest {
+        coEvery { thorChainApi.getPools() } throws RuntimeException("boom")
+
+        val coins = repo().getSecuredAssetCoins()
+
+        assertEquals(
+            setOf("BTC", "ETH", "BCH", "LTC", "DOGE", "AVAX", "BNB"),
+            coins.map { it.ticker }.toSet(),
+        )
+        assertTrue(coins.all { it.isSecuredAsset() })
```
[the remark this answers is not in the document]
rkokhatskyi: Confirmed — the second call stayed inside the 5-minute TTL and never hit the failing mock, so it only proved the TTL cache hit. Fixed by adding an internal `refresh()` seam (bypasses TTL, mirrors `SwapPoolEligibilityRepositoryImpl`'s own public one) and using it to force a real refetch attempt in the test.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/ThorChainSecuredAssetRepositoryImplTest.kt
```diff
@@ -0,0 +1,127 @@
...
+            val repo = repo()
+            val liveSnapshot = repo.getSecuredAssetCoins()
+            assertEquals(listOf("ETH"), liveSnapshot.map { it.ticker })
+
+            coEvery { thorChainApi.getPools() } throws RuntimeException("boom")
+            // Still inside the TTL window, so this read serves the cached live snapshot without
+            // re-fetching (and therefore without hitting the now-failing mock).
+            val secondRead = repo.getSecuredAssetCoins()
+
+            assertEquals(liveSnapshot, secondRead)
```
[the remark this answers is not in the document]
rkokhatskyi: Same fix as the other thread — this is the duplicate finding on the same test. Now uses the internal `refresh()` seam to force a real (failing) refetch instead of relying on TTL timing.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorChainSecuredAssetRepository.kt
```diff
@@ -0,0 +1,113 @@
...
+     * (`btc-btc`, `eth-usdc-0xa0b8…`), or null when malformed or THOR-native (e.g. `THOR.TCY`,
+     * which uses the `x/`-prefixed LP/staking convention, not the secured-asset one).
+     */
+    private fun toSecuredAssetCoin(poolAsset: String): Coin? {
+        val dot = poolAsset.indexOf('.')
+        if (dot <= 0 || dot >= poolAsset.length - 1) return null
+        val chainPart = poolAsset.substring(0, dot)
+        if (chainPart.equals("THOR", ignoreCase = true)) return null
+        val rest = poolAsset.substring(dot + 1)
+        val ticker = rest.substringBefore('-').uppercase()
```
johnnyluo, line 77: [P1] Preserve distinct secured assets that share a ticker

Every generated coin uses the pool ticker, but `Coin.id` is `ticker-chainId` and all of these coins are on THORChain. Assets such as `ETH.USDC` and `AVAX.USDC` therefore both become `USDC-THORChain`; the later `distinctBy { it.id }` silently drops one, so the full secured-asset catalog is not discoverable. Please use an identity that includes the secured contract or underlying chain, and add a test with two pools sharing a ticker.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorChainSecuredAssetRepository.kt
```diff
@@ -0,0 +1,113 @@
...
+                chain = Chain.ThorChain,
+                ticker = ticker,
+                logo = ticker.lowercase(),
+                address = "",
+                // Every THORChain secured asset uses RUNE's 8-decimal precision, not the
+                // underlying chain's native decimals.
+                decimal = SECURED_ASSET_DECIMALS,
+                hexPublicKey = "",
+                priceProviderID = "",
+                contractAddress = "${chainCode.lowercase()}-${denomTail.lowercase()}",
```
johnnyluo, line 109: [P1] Quote the secured asset instead of its underlying L1 asset

These catalog coins are now selectable swap targets, but `ThorChainQuoteSource` passes `dstToken.swapAssetName()` to `to_asset`. For a secured coin with `contractAddress = avax-usdt-...`, the current `Coin.swapAssetName()` path rewrites it to dotted `AVAX.USDT-...`, which requests the real cross-chain asset rather than the THORChain secured asset. With a THORChain destination, the dotted request currently returns HTTP 500; the correct `AVAX-USDT-...` request returns a valid quote and a secured-asset memo. Please preserve dash notation for secured-asset quote requests and add a quote-source/request test for this target.

johnnyluo:
Given THORChain support multiple USDC/USDT , like ETH.USDC , ETH.USDT and also TRON.USDT , on the following screen , please also show which source chain it is , like IOS
CleanShot 2026-07-18 at 11 13 35@2x

IOS: 

image

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorChainSecuredAssetRepository.kt
rkokhatskyi: Confirmed and fixed. Added Coin.catalogKey (contract-qualified for secured assets, e.g. USDC-thorchain-eth-usdc-0xa0b8... vs USDC-thorchain-avax-usdc-0xb97e...) as a dedup identity separate from the persisted Coin.id, mirroring iOS's own id/uniqueId split — so Room/price-cache/disabled-coin persistence (all keyed on Coin.id) is untouched. Used it in the catalog dedup, the LazyColumn key, and also fixed a second instance of the same bug in SelectAssetViewModel's held-vs-catalog exclusion check, which would otherwise hide a not-yet-held secured asset if a different one sharing its ticker was already held. Added a regression test with two pools sharing a ticker, plus restored a source-chain pill in the picker (matching iOS) so rows stay visually distinguishable, not just structurally.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ThorChainSecuredAssetRepository.kt
rkokhatskyi: Confirmed via trace and fixed — swapAssetName() now returns the raw contractAddress verbatim for secured assets instead of thorChainSecuredAssetSwapName()'s dot form. Checked this against current iOS (Coin.swapAsset) and the Windows SDK (toNativeSwapAsset): both pass the denom through unchanged, dash notation, no dot-conversion, confirmed by iOS's own passing test suite. This function was introduced by #4530 in May citing iOS as the reference — that citation was wrong then too. Kept thorChainSecuredAssetSwapName() for swapAssetComparisonName()'s same-asset guard, which needs the dot form to match a secured coin's identity against its native-chain counterpart, so that guard isn't affected. Added a ThorChainQuoteSource test asserting the raw toAsset sent for a secured destination, and updated the two existing Coin tests that were asserting the buggy dot output.

rkokhatskyi:
@johnnyluo Added the source-chain pill next to "Secured" matching iOS

secured asset picker with source-chain pills

data/src/main/kotlin/com/vultisig/wallet/data/models/Coin.kt
```diff
@@ -143,14 +158,20 @@ private fun Coin.thorChainSecuredAssetSwapName(): String {
...
+ * checksum-casing differences from QR payloads; THORChain secured assets are also lowercased so the
+ * THORChain side (`THORCHAIN` standard) matches its native EVM counterpart. Other non-EVM chains
+ * (e.g. Cosmos ibc/, Kujira factory/) use case-sensitive canonical forms as returned by the
+ * THORChain API and are not altered.
  */
 fun Coin.swapAssetComparisonName(): String {
-    val name = swapAssetName()
+    val name =
+        if (chain == Chain.ThorChain && isSecuredAsset()) thorChainSecuredAssetSwapName()
+        else swapAssetName()
```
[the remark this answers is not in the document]
rkokhatskyi: Confirmed and fixed. swapAssetComparisonName() now only lowercases a secured asset when its underlying chain is EVM (checked via Chain.swapAssetName(), the same chain-code table the native-coin side already uses), instead of unconditionally lowercasing every ThorChain secured asset. Added a BTC test case and a secured-BTC-vs-native-BTC equality test alongside the existing Ethereum ones.

data/src/main/kotlin/com/vultisig/wallet/data/models/Coin.kt
```diff
@@ -79,6 +79,15 @@ fun Coin.isSecuredAsset(): Boolean {
...
 }
 
+/**
+ * A dedup/list-key identity, distinct from the persisted [Coin.id]. Secured assets on different
+ * underlying chains can share a ticker (e.g. `ETH.USDC` and `AVAX.USDC` both have ticker `USDC`),
+ * which would otherwise collide on `id` (always `<ticker>-thorchain`) and silently drop one from
+ * catalogs/lists keyed on it.
+ */
+val Coin.catalogKey: String
+    get() = if (isSecuredAsset()) "$id-$contractAddress" else id
```
johnnyluo: [P1] Propagate contract-qualified identity beyond the catalog

catalogKey makes ETH.USDC and AVAX.USDC distinct only in catalog/UI lists. Both still have Coin.id == USDC-thorchain; SwapTokenSelector stores result.token.id, AccountsRepository resolves by id, and CoinEntity uses id + vaultId as its primary key. Selecting a second secured USDC can therefore replace the first persisted coin or resolve the picker choice back to the wrong account, potentially quoting or swapping the wrong secured denom. Please carry the contract-qualified identity through persistence, account selection, balances, and prices, or prevent same-ticker secured assets from coexisting.

rkokhatskyi:
Found few other things while reviewing, Claude missed the migration and few other important aspects. It has been fixed now

data/src/main/kotlin/com/vultisig/wallet/data/models/Coin.kt
rkokhatskyi: Fixed at the root: Coin.id is now contract-qualified for secured assets (not a separate catalogKey), so persistence, account resolution, and dedup all share one identity. A follow-up self-review caught two related bugs this would've otherwise left in place — the balance cache (TokenValueEntity) and the picker's request id (SwapTokenSelector) needed the same fix. Fixed both, added tests for all three, full suite green.

johnnyluo (approved):
LGTM

PR #5280, opened 2026-07-14
fix(keysign): skip branded splash on notification cold start (#5269)

### LINKED ISSUE #5269: [Android] Keysign notification cold-start replays full splash before keysign — skip splash when opened from a keysign intent
## Summary
Opening a keysign push notification **occasionally** cold-restarts the app before the keysign flow appears, even when the app was recently unlocked/foregrounded:

1. The full branded **`AnimatedSplash` replays** (looks like the app restarting from scratch).
2. Then the keysign start animation plays and the Send/Swap overview comes up.

The keysign always completes correctly — the only defect is the jarring **full-splash replay** on the cold path. This is a UX-hardening ask: when the app is cold-started *by a keysign notification*, skip/abbreviate the splash and route straight into keysign.

## Root cause (verified)
The keysign notification is an FCM push (`VultisigFirebaseMessagingService`). There are three delivery paths; only the third is affected:

1. **App foreground + alive** → local broadcast (`PUSH_NOTIFICATION_ACTION`) → in-app `ForegroundNotificationBanner`. No restart.
2. **Backgrounded, process alive** → tap fires `MainActivity.onNewIntent()` (warm) → `mainViewModel.onPushNotificationReceived()` (`MainActivity.kt:157`). No restart.
3. **Backgrounded, process reclaimed by the OS** → tap cold-starts `MainActivity.onCreate` → `installSplashScreen()` (`MainActivity.kt:82`) + Compose `AnimatedSplash(...)` (`MainActivity.kt:116`) replay → then `onCreate` reads the `qr_code_data` extra and calls `mainViewModel.onPushNotificationReceived(it)` (`MainActivity.kt:86-88`, comment: *"Handle notification tap when app was killed — ViewModel awaits navigation readiness"*).

**The intermittency is Android background process death, not an app self-restart.** Confirmed no app-side restart cause: no `recreate()`, `finishAffinity()`, `finishAndRemoveTask()`, `killProcess`, and no app-lock/auto-lock feature that tears down the activity (greps empty on `origin/main`). When the OS has reclaimed the process, tap = cold start; when it hasn't, `CLEAR_TOP | SINGLE_TOP` gives a warm `onNewIntent`. "Even if unlocked before" = the prior foreground session's process was reclaimed while backgrounded.

The payload survives on both paths (it rides in the notification intent extra `qr_code_data`), which is why keysign always resumes.

## Deterministic repro
"Occasionally" → "every time" by forcing process death:
- Developer Options → **Don't keep activities** = ON (or `adb shell am kill com.vultisig.wallet` while the app is backgrounded), then
- Trigger a keysign request from the co-signer and tap the notification.
- Observe: full `AnimatedSplash` replays, then keysign.

## Proposed fix (scoped — app-side, small)
When `MainActivity` is cold-started with a `qr_code_data` extra present (i.e. launched from a keysign notification), **skip or abbreviate `AnimatedSplash`** and route directly into the keysign flow, so the cold path visually matches the warm `onNewIntent` path instead of looking like a restart.
- Key file: `app/src/main/java/com/vultisig/wallet/app/activity/MainActivity.kt` (`onCreate` splash block `:82-116`, extra read `:86-88`).

## Out of scope (possible follow-up)
Reducing how *often* the cold path happens at all — e.g. a short-lived foreground service to keep the process alive while a keysign request is pending (there's currently no FGS covering incoming keysign; `TransactionStatusService` only covers outbound tx status). Larger change; not required for this fix.

## Scope
Android. Keysign push-notification cold-start path.

aminsato:
## Summary
Closes #5269.

When the OS has reclaimed the process, tapping a keysign push notification **cold-starts** `MainActivity`, which replays the full branded `AnimatedSplash` before the keysign flow appears — so the cold path looks like the app restarting from scratch. The warm `onNewIntent` path (process still alive) shows no splash at all.

This makes the cold notification path match the warm path:
- Track whether `onCreate` was launched from a keysign notification (the `qr_code_data` extra is present) and, if so, **skip `AnimatedSplash`**.
- Because `SetupNavGraph`'s `NavHost` captures its start destination on first composition, hold on a plain background while the destination resolves (`isLoading`) rather than starting on the stale `Route.Home` default — then route straight into the keysign flow via the existing `onPushNotificationReceived` (which awaits navigation readiness).

Only the keysign cold-start path changes; every other launch (normal app open) still shows the splash exactly as before.

## Scope
Single file: `app/src/main/java/com/vultisig/wallet/app/activity/MainActivity.kt`.

## Test plan
Reproduced path 3 from the issue on an emulator by simulating the FCM `PendingIntent` after killing the process:

```
adb shell am force-stop com.vultisig.wallet
adb shell am start -n com.vultisig.wallet/.app.activity.MainActivity \
  --es message "<keysign-deeplink>" -f 0x34000000   # NEW_TASK|CLEAR_TOP|SINGLE_TOP
```

Burst screencaps confirm:
- **Normal cold start (control):** branded `AnimatedSplash` plays (logo → "Vultisig" wordmark).
- **Notification cold start (fixed):** no branded splash — plain background briefly while the start destination resolves, then straight into content; the notification handler runs end-to-end (a non-matching test vault correctly surfaces the "Vault not found" snackbar; a real payload routes into `Keysign.Join`).

Verified `:app:compileDebugKotlin` green and ktfmt applied.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app startup when opened from a key-signing notification.
  * Prevented the branded splash screen from appearing during this notification-based launch flow.
  * Added a smoother loading state before navigating to the appropriate screen.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/app/activity/MainActivity.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/MainActivity.kt
@@ -16,6 +16,9 @@ import androidx.activity.enableEdgeToEdge
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import androidx.compose.animation.ExperimentalAnimationApi
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.material3.windowsizeclass.ExperimentalMaterial3WindowSizeClassApi
 import androidx.compose.material3.windowsizeclass.calculateWindowSizeClass
 import androidx.compose.runtime.Composable
@@ -25,6 +28,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.toArgb
 import androidx.core.content.ContextCompat
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
@@ -84,9 +88,13 @@ class MainActivity : AppCompatActivity() {
         super.onCreate(savedInstanceState)
 
         // Handle notification tap when app was killed — ViewModel awaits navigation readiness.
-        intent?.getStringExtra(VultisigFirebaseMessagingService.QR_CODE_DATA)?.let {
-            mainViewModel.onPushNotificationReceived(it)
-        }
+        // Track whether this cold start was launched by a keysign notification, so we can skip
+        // the branded AnimatedSplash replay and make the cold path match the warm onNewIntent path.
+        val launchedFromKeysignNotification =
+            intent?.getStringExtra(VultisigFirebaseMessagingService.QR_CODE_DATA)?.let {
+                mainViewModel.onPushNotificationReceived(it)
+                true
+            } ?: false
 
         val systemBarStyle =
             SystemBarStyle.auto(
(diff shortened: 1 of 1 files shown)


rkokhatskyi (approved):
LGTM

PR #5283, opened 2026-07-14
fix(tron): operation-aware Verify header for freeze/unfreeze (#5274)

### LINKED ISSUE #5274: [Android] Verify screen shows generic "You're sending" for non-send operations (e.g. TRX unfreeze) — surface operation-aware Verify headers
## Summary
Non-send operations fall back to the generic **"You're sending"** Verify screen. The clearest case: **unfreezing TRX** (unstaking — releasing frozen balance back to available) shows *"You're sending [amount] TRX"*, which misrepresents the operation. The Verify screen is the last human checkpoint before signing a money-moving action, so it must state the true operation — not "sending."

This issue is to **stop defaulting non-send operations to the generic send header**, starting with TRON staking (freeze/unfreeze/withdraw).

## Root cause (verified)
1. **Both Verify screens hard-code the "You're sending" header**, regardless of operation:
   - `VerifySendScreen.kt:240` — `header = stringResource(R.string.verify_deposit_sending)`
   - `VerifyDepositScreen.kt:144` — same string
   - `verify_deposit_sending` = *"You're sending"* (`strings.xml:764`)

2. **The operation type is known at the form but lost by Verify.** TRON staking is routed through the Send form. The form screen renders the correct title — `SendFormScreen.kt:220` → `tron_unfreeze_screen_title` = *"Unfreeze TRON"* (`strings.xml:79`; freeze = *"Freeze TRON"*, `:78`). But the operation is carried onward **only as an internal memo prefix** `"UNFREEZE:<resource>"` / `"FREEZE:<resource>"`:
   - `tronStakingMemo()` → `"${op.memoPrefix}:${resource.name}"` (`TronStaking.kt:8-20`)
   - `TronStakingService.applyStakingMemo()` (`TronStakingService.kt:79-88`)
   - `TronHelper` consumes that memo to build `UnfreezeBalanceV2` (`TronHelper.kt:51`, `:279-280`)

   By the Verify layer the transaction is just a send with a memo — no operation signal survives — so it falls back to the generic "You're sending" header.

3. **Likely secondary leak (please confirm):** `VerifySendScreen.kt:274` renders a **Memo row for any non-blank `tx.memo`**. If the internal `"UNFREEZE:BANDWIDTH"` prefix reaches `tx.memo` at the Verify UI-model layer, it surfaces to users as a *"Memo: UNFREEZE:BANDWIDTH"* row — internal plumbing leaking into a signing screen. (Verified the display path; not yet traced whether the staking memo populates the send-Verify `tx.memo` vs. only `keysignPayload.memo`.)

## Impact
- Users freezing/unfreezing TRX see "You're sending" on the pre-signing Verify screen — the last checkpoint misstates the operation.
- Systemic, not TRX-only: the "You're sending" header is shared by both `VerifySendScreen` and `VerifyDepositScreen`, so other non-send operations routed through these screens are affected the same way.

## Proposed fix (matches scope: kill generic send-only Verify)
- **Thread the operation type (not just a memo prefix) through to the Verify UI model** so the Verify layer can render intent.
- **Derive the Verify header/title from the operation** instead of hard-coding "You're sending." For TRON staking, mirror the form labels (*Unfreeze TRON* / *Freeze TRON* / *Withdraw*), and **don't render the internal memo prefix as a Memo row.**
- Generalize the header so non-send operations (staking, resource freeze/unfreeze, votes, withdraws, contract interactions) get an intent-revealing Verify header rather than the generic send fallback.

## Scope
Android. **Concrete first cut:** TRON staking (freeze / unfreeze / withdraw) Verify screen. **Follow-up:** audit which other operations currently fall through to "You're sending" via the shared header (deposit-verify path) and give them operation-aware headers too.

## Key files
- `app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt:240` (+ memo row `:274`)
- `app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt:144`
- `app/src/main/java/com/vultisig/wallet/ui/models/send/TronStakingService.kt:79-88`
- `data/src/main/kotlin/com/vultisig/wallet/data/blockchain/tron/TronStaking.kt:8-20`
- `data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TronHelper.kt:51,279-280`
- `app/src/main/java/com/vultisig/wallet/ui/screens/send/SendFormScreen.kt:220` (reference: form already shows the right label)

aminsato:
## Summary
The Verify screen is the last human checkpoint before signing. TRON freeze/unfreeze is routed through the Send form and carries its operation only as an internal memo prefix (`FREEZE:<resource>` / `UNFREEZE:<resource>`). By the Verify layer that prefix populates `tx.memo` and no operation signal survives, so the screen:

- showed the generic **"You're sending"** header — misrepresenting a stake/unstake as a plain send, and
- leaked the raw prefix as a **"Memo: UNFREEZE:ENERGY"** row (internal plumbing on a signing screen).

Fixes #5274.

## Changes
- `TransactionDetailsUiModel` — added `@StringRes headerTitleRes: Int?` (null → generic "You're sending" fallback).
- `TransactionToUiModelMapper` — detects the TRON staking memo via the existing `TRON_STAKING_MEMO_REGEX`, maps the operation to the same labels the Send form uses (`Freeze TRON` / `Unfreeze TRON`), and nulls the internal memo so it no longer renders as a Memo row.
- `VerifySendScreen` — header now reads `tx.headerTitleRes ?: R.string.verify_deposit_sending`.

## Scope
First cut is the Send-form path (TRON freeze/unfreeze), matching the issue. `VerifyDepositScreen` uses a different model with its own `operation` field and is untouched. `TronStakingOperation` currently has only FREEZE/UNFREEZE (no WITHDRAW), so those two are covered.

## Test plan
- New unit test `TransactionToUiModelMapperTronStakingTest` (3 cases): unfreeze → unfreeze header + memo hidden; freeze → freeze header + memo hidden; plain TRON send → generic header + memo preserved. `:app:testDebugUnitTest` green.
- Manual: Vault → DeFi tab → TRON → Unfreeze → amount → Continue → Verify.

## Before / After

| Before | After |
|--------|-------|
| before | after |

## Summary by CodeRabbit

* **New Features**
  * Added operation-specific headers for TRON staking transactions (Freeze/Unfreeze) and display resource-focused memos.
* **Bug Fixes**
  * Updated transaction verification to use the new header when available, otherwise fall back to the standard title.
  * Ensured TRC20 staking memos are not misclassified as TRON staking.
* **Tests**
  * Added automated coverage for TRON Freeze/Unfreeze memo parsing and standard memo passthrough behavior.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToUiModelMapper.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToUiModelMapper.kt
@@ -1,5 +1,8 @@
 package com.vultisig.wallet.ui.models.mappers
 
+import com.vultisig.wallet.R
+import com.vultisig.wallet.data.blockchain.tron.TRON_STAKING_MEMO_REGEX
+import com.vultisig.wallet.data.blockchain.tron.TronStakingOperation
 import com.vultisig.wallet.data.chains.helpers.RippleDestinationTag
 import com.vultisig.wallet.data.mappers.SuspendMapperFunc
 import com.vultisig.wallet.data.models.Chain
@@ -28,8 +31,34 @@ constructor(
                 ?: if (from.token.chain == Chain.Ripple)
                     RippleDestinationTag.parseCanonicalDestinationTag(from.memo)?.toString()
                 else null
+        // TRON freeze/unfreeze is routed through the Send form and carries its operation only as an
+        // internal memo prefix ("FREEZE:<resource>" / "UNFREEZE:<resource>"). Surface that as an
+        // operation-aware Verify header and drop the raw prefix so it doesn't leak into the Memo
+        // row.
+        val tronStakingOperation =
+            if (from.token.chain == Chain.Tron) {
+                from.memo
+                    ?.let { TRON_STAKING_MEMO_REGEX.matchEntire(it) }
+                    ?.groupValues
+                    ?.get(1)
+                    ?.let { prefix ->
+                        TronStakingOperation.entries.firstOrNull { it.memoPrefix == prefix }
+                    }
+            } else null
+
         val memo =
-            if (from.token.chain == Chain.Ripple && from.memo == destinationTag) null else from.memo
+            when {
+                from.token.chain == Chain.Ripple && from.memo == destinationTag -> null
+                tronStakingOperation != null -> null
+                else -> from.memo
+            }
+
+        val headerTitleRes =
+            when (tronStakingOperation) {
+                TronStakingOperation.FREEZE -> R.string.tron_freeze_screen_title
+                TronStakingOperation.UNFREEZE -> R.string.tron_unfreeze_screen_title
+                null -> null
+            }
 
         return TransactionDetailsUiModel(
             token =
@@ -42,6 +71,7 @@ constructor(
             dstAddress = from.dstAddress,
             dstLabel = from.dstLabel,
             memo = memo,
+            headerTitleRes = headerTitleRes,
             destinationTag = destinationTag,
             signAmino = from.signAmino,
             signDirect = from.signDirect,
--- a/app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToUiModelMapper.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToUiModelMapper.kt
@@ -33,23 +33,23 @@ constructor(
                 else null
         // TRON freeze/unfreeze is routed through the Send form and carries its operation only as an
         // internal memo prefix ("FREEZE:<resource>" / "UNFREEZE:<resource>"). Surface that as an
-        // operation-aware Verify header and drop the raw prefix so it doesn't leak into the Memo
-        // row.
+        // operation-aware Verify header, but keep the captured resource (BANDWIDTH/ENERGY) in the
+        // Memo row so the signed contract field stays visible. Gate on the native token so a TRC20
+        // send that merely echoes a matching memo isn't mislabeled as a staking operation.
+        val tronStakingMatch =
+            if (from.token.chain == Chain.Tron && from.token.isNativeToken)
+                from.memo?.let { TRON_STAKING_MEMO_REGEX.matchEntire(it) }
+            else null
         val tronStakingOperation =
-            if (from.token.chain == Chain.Tron) {
-                from.memo
-                    ?.let { TRON_STAKING_MEMO_REGEX.matchEntire(it) }
-                    ?.groupValues
-                    ?.get(1)
-                    ?.let { prefix ->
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToUiModelMapper.kt
```diff
@@ -28,8 +31,34 @@ constructor(
...
+                    ?.let { prefix ->
+                        TronStakingOperation.entries.firstOrNull { it.memoPrefix == prefix }
+                    }
+            } else null
+
         val memo =
-            if (from.token.chain == Chain.Ripple && from.memo == destinationTag) null else from.memo
+            when {
+                from.token.chain == Chain.Ripple && from.memo == destinationTag -> null
+                tronStakingOperation != null -> null
```
rkokhatskyi: `tronStakingOperation != null -> null` nulls the whole memo, discarding the resource the regex captured, so Verify shows `Freeze TRON` with no BANDWIDTH/ENERGY row, a signed contract field that iOS and the pre-PR memo row both surfaced before signing

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToUiModelMapper.kt
aminsato: Good catch — fixed in fefa9077c. Instead of nulling the memo, the Verify Memo row now shows the captured resource (`BANDWIDTH`/`ENERGY`, regex group 2) while the header carries the operation, so the signed contract field stays visible before signing.

app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToUiModelMapper.kt
```diff
@@ -28,8 +31,34 @@ constructor(
...
+                else -> from.memo
+            }
+
+        val headerTitleRes =
+            when (tronStakingOperation) {
+                TronStakingOperation.FREEZE -> R.string.tron_freeze_screen_title
+                TronStakingOperation.UNFREEZE -> R.string.tron_unfreeze_screen_title
+                null -> null
+            }
```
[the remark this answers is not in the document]
aminsato: Fixed in fefa9077c. Detection is now gated on `from.token.isNativeToken` in addition to `Chain.Tron`, so a TRC20 send that merely echoes a `FREEZE:`/`UNFREEZE:` memo keeps its generic header and memo instead of being mislabeled as a staking operation. Added a TRC20-echo test to cover it.

rkokhatskyi (approved):
LGTM

johnnyluo (approved):
LGTM

PR #5264, opened 2026-07-12
feat(thorchain): add bRUNE & ybRUNE (display, pricing, LP exclusion)

aminsato:
## Summary

Closes #5241 — full bRUNE / ybRUNE support on THORChain, matching iOS parity for these Rujira liquid-bonding assets.

bRUNE ("Bonded RUNE") and its auto-compounding staking receipt ybRUNE were only surfacing through the generic THORChain bank-denom auto-discovery, which uppercased their tickers (`BRUNE` / `YBRUNE`), had no logo, and left them unpriced. This PR turns them into first-class, properly rendered tokens.

### 1. Picker exclusion (`isLpToken`)
`Coin.isLpToken` matched `contractAddress == "x/brune"` under the **`Chain.MayaChain`** branch, but `x/brune` is a **THORChain** denom — so the check was unreachable dead code and bRUNE was never excluded from the swap/asset pickers. Moved the check to the `Chain.ThorChain` branch (ybRUNE `x/staking-x/brune` is already covered by the existing `x/staking-` prefix). Consumed by `SelectAssetViewModel` / `FastSelectionPopupSharedViewModel` (`filterNot { it.isLpToken }`).

### 2. Display
- Curated `Coins.ThorChain` entries: **bRUNE** (`x/brune`, 8 dec) and **ybRUNE** (`x/staking-x/brune`, 8 dec).
- Added the real bRUNE logo (`brune.png`) and mapped `brune` / `ybrune` in `CoinLogo.kt`.
- Restored proper ticker casing in the THORChain discovery path so they render as `bRUNE` / `ybRUNE` with the correct icon instead of the generic uppercased fallback.

### 3. Pricing (`TokenPriceRepository.fetchThorContractPrices`)
- **bRUNE → RUNE parity.** It's ≥1:1 RUNE-backed with no THORChain pool, so it tracks RUNE; no NAV query needed.
- **ybRUNE → NAV × bRUNE.** NAV = `liquid_bond_size / liquid_bond_shares` from the standard `rujira-staking` `{"status":{}}` smart query on contract `thor179fex2rxd45caedmz4hxsnu42sw20lu0djyh4yukyh965sq8muuqptru2g` — the exact sTCY mechanism, with RUNE as the base asset instead of TCY.
- Extracted the shared NAV math into `navPerShareFromStatus(...)`, now reused by both sTCY and ybRUNE (behavior-preserving for sTCY).

### Deliberate approximations
[...]

PR #5284, opened 2026-07-14
fix(thorchain): bRUNE/ybRUNE follow-ups — discovery, pricing & swap gating (#5275)

### LINKED ISSUE #5275: [Fix] bRUNE/ybRUNE follow-ups: dead auto-discovery, stale legacy ticker, and THORChain price-helper bugs (post #5264)
# [Fix] bRUNE/ybRUNE follow-ups: dead auto-discovery, stale legacy ticker, and THORChain price-helper bugs (post #5264)

## Problem
PR #5264 added bRUNE/ybRUNE display, LP-exclusion, and pricing, and merged before a requested-changes review could land (it went through the merge queue first). The review surfaced several confirmed gaps in the merged code that are still open:

1. **Auto-discovery never fires for a new holder.** `getRefreshTokens`'s `enabledDenoms` gate (`TokenRepository.kt:98`) drops any bank denom not already in `vault.coins`, and a fresh vault only ever seeds native RUNE, so the new canonicalization block (`TokenRepository.kt:151-161`) is unreachable the first time a user actually receives bRUNE/ybRUNE. Today it only activates via a manual "Select Assets" toggle.
2. **Legacy pre-PR tickers never correct.** Anyone who already had bRUNE/ybRUNE auto-discovered before this PR shipped keeps the old uppercase "BRUNE"/"YBRUNE" ticker forever: `TokenRefreshWorker.kt:109`'s dedup is case-insensitive (`id.equals(..., ignoreCase = true)`) and returns before ever reaching the replace-on-mismatch path.
3. **`isLpToken`'s ThorChain check is case-sensitive.** `Coin.kt:49`'s `contractAddress == "x/brune"` is an exact match, unlike the `.lowercase()` guard the same PR added in `TokenPriceRepository.kt`. A coin with different persisted casing (from before this PR, or any future non-canonical on-chain denom) stays swappable.
4. **`runePriceUsd()` reads the cache with the wrong key.** `TokenPriceRepository.kt:492-494` looks up the price cache by `Coins.ThorChain.RUNE.priceProviderID` ("thorchain"), but every cache row is keyed by `Coin.id` ("RUNE-THORChain"). The cache branch can never hit, so every refresh redundantly re-hits CoinGecko live, and a transient empty/rate-limited response persists a $0 price for bRUNE/ybRUNE with nothing logged.
5. **The same helper double-applies FX for non-USD currencies.** Its live fallback returns RUNE's price in the user's selected app currency (not USD), but `TokenPriceRepository.kt:472` multiplies that value by `tetherPrice` a second time. A JPY user, for example, would see bRUNE valued at roughly 150x its real value.
6. **A malformed NAV response silently persists $0.** `navPerShareFromStatus` (`TokenPriceRepository.kt:501`) returns a real NAV of 0 (not the intended no-bonds fallback of 1) if `liquid_bond_size` fails to parse while `liquid_bond_shares` parses positive, and `savePrices` only guards against an empty map, not a zero value, unlike the sibling contract-address path at line 172.
7. **`decimal` isn't reset alongside ticker/contractAddress.** The override block at `TokenRepository.kt:155` restores `symbol`/`contractAddress` to the curated identity but leaves `decimal` at whatever the metadata ladder above computed. If THORChain ever registers denom metadata for these two denoms with a non-8 exponent, the discovered coin would silently disagree with the curated 8-decimal definition.
8. **LP-exclusion is incomplete outside the asset pickers.** `TokenDetailViewModel.kt:87`'s Swap button routes with `srcTokenId = tokenId` unconditionally, with no `isLpToken` check anywhere in that path, so bRUNE/ybRUNE (like sTCY/yRUNE/yTCY before them) can still be picked as a swap source directly from their own Token Detail page.
9. **The new NAV contract has no fixture coverage.** `BRUNE_STAKING_CONTRACT` (`TokenPriceRepository.kt:515`) is a brand-new address with zero in-repo test coverage, and every field in `VaultRedemptionDataJson` defaults to an empty string, so a wrong address or a renamed response field would deserialize cleanly and silently price ybRUNE at exact RUNE parity instead of failing loudly.

## Expected Behavior
- A bRUNE/ybRUNE balance auto-surfaces the first time a wallet holds it, without a manual enable step.
- A wallet that already held bRUNE/ybRUNE before this PR shows the corrected ticker/casing after the next refresh.
- `isLpToken` and the swap entry points agree: bRUNE/ybRUNE can't be selected as a swap asset from any screen, not just the two pickers.
- bRUNE/ybRUNE price in any app currency matches RUNE's real USD-equivalent price, using the cache correctly and never persisting a $0/wrong-magnitude value on a transient failure.

## Solution
- Gate #1: either seed `enabledDenoms` from the live balance response instead of only `vault.coins` on first discovery, or otherwise make first-time bank-denom discovery reachable without a manual toggle.
- Gate #2: add a coin-identity correction path (similar in spirit to iOS's `CoinService.migrateOldContractAddresses`) so a ticker/contractAddress fix can overwrite a stale persisted entry instead of being permanently skipped by the case-insensitive dedup.
- #3: match `Coin.kt:49`'s check casing to the `.lowercase()` pattern already used in `TokenPriceRepository.kt`.
- #4/#5: key `runePriceUsd()`'s cache lookup by `Coins.ThorChain.RUNE.id` and route its live fallback through the same currency-normalizing helper the rest of the file uses, instead of assuming the fallback result is USD.
- #6: apply the same `value != BigDecimal.ZERO` guard used at `TokenPriceRepository.kt:172` to the `fetchThorContractPrices` save path.
- #7: reset `decimal` to the curated value inside the same override block that already resets `symbol`/`contractAddress`.
- #8: add an `isLpToken` check to `TokenDetailViewModel`'s swap-eligibility gate (or a single shared choke point all swap entry points go through).
- #9: add a unit test with a fixture response for `BRUNE_STAKING_CONTRACT`'s `{"status":{}}` query to pin the expected shape.

### Must Do
- [ ] Make first-time bRUNE/ybRUNE (and other never-before-tracked ThorChain bank-denom) auto-discovery actually reachable
- [ ] Add a migration/correction path so a stale pre-PR ticker gets fixed instead of permanently skipped
- [ ] Make `Coin.isLpToken`'s ThorChain match case-insensitive
- [ ] Fix `runePriceUsd()`'s cache key to use `Coin.id`, not `priceProviderID`
- [ ] Fix the FX double-conversion in the live-fallback path for non-USD currencies
- [ ] Add a zero-value guard to the `fetchThorContractPrices` save path
- [ ] Reset `decimal` in the bRUNE/ybRUNE canonicalization override
- [ ] Add an `isLpToken` (or equivalent) gate to the Token Detail Swap button
- [ ] Add a fixture test for `BRUNE_STAKING_CONTRACT`'s expected response shape

### Must NOT Do
- Don't touch the bRUNE = RUNE parity pricing assumption itself (a disclosed, deliberate approximation from #5264, not a bug)
- Don't refactor the whole pricing pipeline; keep each fix scoped to its own function
- Don't change pricing/discovery behavior for already-shipped tokens (RUJI/KUJI/TCY/sTCY/yRUNE/yTCY) beyond reusing the same guard patterns

## Acceptance Criteria
- [ ] `./gradlew :data:testDebugUnitTest` passes, including new tests for each item above
- [ ] A vault receiving its first-ever bRUNE/ybRUNE balance shows it automatically on the next refresh
- [ ] A vault with a pre-#5264 "BRUNE"/"YBRUNE" entry shows the corrected ticker after one refresh
- [ ] bRUNE/ybRUNE cannot be selected as a swap source from the Token Detail screen
- [ ] bRUNE/ybRUNE USD value is correct for a non-USD app currency

## Context
- Follow-up from the deep review of #5264 (merged c1bb9b5 before the review posted, via the merge queue)
- See inline comments on https://github.com/vultisig/vultisig-android/pull/5264 for the exact code citations and reasoning behind each item above

aminsato:
Closes #5275

Post-merge cleanup of the confirmed review gaps from #5264 (which merged via the queue before the requested-changes review landed). Each fix is scoped to its own function; the bRUNE = RUNE parity approximation and the already-shipped RUJI/KUJI/TCY paths are deliberately left untouched.

## What & why

**Discovery / identity** (`TokenRepository`, `TokenRefreshWorker`)
- Auto-discovery never fired for a new holder: `getRefreshTokens`' `enabledDenoms` gate is seeded only from `vault.coins`, and a fresh vault only seeds native RUNE, so a received bRUNE/ybRUNE denom was dropped before reaching the canonicalization block. Curated liquid-bonding denoms now bypass the gate.
- Legacy pre-#5264 entries kept the old uppercase `BRUNE`/`YBRUNE` forever — the refresh dedup matches ids case-insensitively and returned early. Added an identity-correction path that overwrites a persisted entry whose ticker/contractAddress/decimal has drifted.
- `decimal` is now reset in the canonicalization override alongside ticker/contractAddress.

**Swap gating** (`Coin.isLpToken`, `TokenDetailViewModel`)
- `isLpToken`'s ThorChain check was case-sensitive (`== "x/brune"`); now compares on the lowercased denom, matching the pricing path.
- The Token Detail Swap button had no `isLpToken` gate, so bRUNE/ybRUNE could be picked as a swap source from their own detail page even though the asset pickers exclude them. `canSwap` now also requires `!isLpToken`.

**Pricing** (`TokenPriceRepository`)
- `runePriceUsd()` read the cache by `priceProviderID` ("thorchain") while rows are keyed by `Coin.id` ("RUNE-THORChain"), so the cache branch never hit and every refresh re-fetched live. Now keyed by `Coin.id`.
- The live fallback returned RUNE in the app currency, which callers then multiplied by `tetherPrice` again (a JPY user saw ~150x). The fallback now fetches RUNE explicitly in USD.
- A transient failure (unparseable NAV, rate-limited RUNE) yields a 0 price, and `savePrices` only guarded an empty currency map — so a $0 overwrote the last-known good price. Zero prices are now dropped on the save path (via `signum()`, so a scaled `0E-8` from a NAV division still counts).

## Tests
- `TokenPriceRepositoryImplTest` — bRUNE parity & ybRUNE NAV×RUNE, cache-read-by-id, live-USD FX-applied-once, malformed-NAV zero-guard, redemption-JSON shape fixture.
- `TokenRepositoryImplTest` — auto-discovery bypass, decimal canonicalization.
- `TokenRefreshWorkerTest` — stale-ticker correction, no-op when already correct.
- `CoinIsLpTokenTest` — non-canonical casing still detected as LP.

`./gradlew :data:testDebugUnitTest` passes; `:app` and `:data` compile.

## Test plan
- [x] `./gradlew :data:testDebugUnitTest`
- [x] `:app:compileDebugKotlin` / `:data` compile
- [ ] Device: receive first-ever bRUNE/ybRUNE → auto-surfaces on next refresh
- [ ] Device: pre-existing `BRUNE` entry shows corrected `bRUNE` after one refresh
- [ ] Device: bRUNE/ybRUNE Swap button disabled on the Token Detail screen
- [ ] Device: bRUNE/ybRUNE fiat value correct for a non-USD app currency

## Summary by CodeRabbit

- **Bug Fixes**
  - Disabled swap actions for THORChain LP receipt tokens, including improved detection for non-canonical casing.
  - Prevented $0 (zero) derived prices from overwriting existing cached prices.
  - Improved THORChain bRUNE/ybRUNE and RUNE/sTCY pricing, including more consistent USD handling and better cache/refresh behavior.
  - Corrected token identity drift during refresh and ensured curated THORChain tokens surface automatically with accurate decimals.

- **Tests**
  - Added/expanded coverage for LP detection, curated token discovery/decimals, refresh identity correction, and bRUNE/ybRUNE pricing with NAV redemption edge cases.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
@@ -422,6 +422,22 @@ constructor(
                         fetchTetherPrice()
                     }
 
+                // bRUNE and ybRUNE both price off RUNE-in-USD. Fetch it once up front (only when a
+                // RUNE-backed denom is present) so concurrent per-token async blocks don't each
+                // miss
+                // a cold cache and fire a duplicate live CoinGecko call for the same value.
+                val runeUsdPrice =
+                    if (
+                        matchingTokens.any {
+                            val denom = it.contractAddress.lowercase()
+                            denom == BRUNE_DENOM || denom == YBRUNE_DENOM
+                        }
+                    ) {
+                        runePriceUsd()
+                    } else {
+                        BigDecimal.ZERO
+                    }
+
                 val tokenIdToPrices = coroutineScope {
                     contracts
                         .zip(tokenIds)
@@ -434,7 +450,7 @@ constructor(
                                         when (token.contractAddress.lowercase()) {
                                             // bRUNE is ≥1:1 RUNE-backed with no THORChain pool, so
                                             // it tracks RUNE at parity.
-                                            BRUNE_DENOM -> runePriceUsd()
+                                            BRUNE_DENOM -> runeUsdPrice
                                             // ybRUNE is the auto-compounding bRUNE staking receipt:
                                             // NAV (liquid_bond_size / liquid_bond_shares) × bRUNE,
                                             // and bRUNE ≈ RUNE. Same mechanism as sTCY.
@@ -445,7 +461,7 @@ constructor(
                                                             contract
                                                         )
                                                     )
-                                                nav * runePriceUsd()
+                                                nav * runeUsdPrice
                                             }
                                             "x/staking-tcy" -> {
                                                 val tcyPriceUSD =
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
@@ -433,7 +433,17 @@ constructor(
                             denom == BRUNE_DENOM || denom == YBRUNE_DENOM
                         }
                     ) {
-                        runePriceUsd()
+                        // This runs outside the per-token async guards, so a failure here would
+                        // abort the whole batch (NAMI, sTCY, …). Contain it and fall back to 0 —
+                        // the zero-price filter below then drops bRUNE/ybRUNE for this cycle
+                        // instead of overwriting their last-known good price.
+                        try {
+                            runePriceUsd()
+                        } catch (e: Exception) {
+                            if (e is kotlinx.coroutines.CancellationException) throw e
+                            Timber.e(e, "Failed to fetch shared RUNE-in-USD price")
+                            BigDecimal.ZERO
+                        }
                     } else {
                         BigDecimal.ZERO
                     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
@@ -438,7 +438,7 @@ constructor(
                         // the zero-price filter below then drops bRUNE/ybRUNE for this cycle
                         // instead of overwriting their last-known good price.
                         try {
-                            runePriceUsd()
+                            runePriceUsd(currency)
                         } catch (e: Exception) {
                             if (e is kotlinx.coroutines.CancellationException) throw e
                             Timber.e(e, "Failed to fetch shared RUNE-in-USD price")
@@ -474,9 +474,7 @@ constructor(
                                                 nav * runeUsdPrice
                                             }
                                             "x/staking-tcy" -> {
-                                                val tcyPriceUSD =
-                                                    getCachedPrice("tcy", AppCurrency.USD)
-                                                        ?: getPriceByPriceProviderId("tcy")
+                                                val tcyPriceUSD = tcyPriceUsd(currency)
                                                 val nav =
                                                     navPerShareFromStatus(
                                                         thorApi.getThorchainTokenPriceByContract(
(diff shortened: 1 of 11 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/workers/TokenRefreshWorker.kt
```diff
@@ -106,11 +106,21 @@ constructor(
...
+            enabledCoinIds.firstOrNull { enabledCoinId ->
                 enabledCoinId.id.equals(refreshToken.id, ignoreCase = true)
             }
-        if (isTokenExist) return
+        if (existing != null) {
+            // The token is already tracked. The id match is case-insensitive, so a stale entry
+            // whose curated identity has since been corrected (e.g. a ticker recased from "BRUNE"
+            // to "bRUNE", or a canonicalized contractAddress/decimal) would otherwise be kept
+            // forever. Overwrite it with the freshly derived identity when they disagree.
+            if (existing.needsIdentityCorrection(refreshToken)) {
```
rkokhatskyi, line 119: `needsIdentityCorrection` fires on any contractAddress difference while `Coin.id` is ticker based (`Coin.kt:21`), so a second CoinGecko listed token that 1inch reports under an existing ticker deletes the persisted coin and re-adds a different contract under that id, flipping the send target on each refresh. Gate on `contractAddress.equals(refreshToken.contractAddress, ignoreCase = true)` so only the same contract bRUNE recasing survives

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -490,8 +496,23 @@ constructor(
...
-        getCachedPrice(Coins.ThorChain.RUNE.priceProviderID, AppCurrency.USD)
-            ?: getPriceByPriceProviderId(Coins.ThorChain.RUNE.priceProviderID)
+        // Cache rows are keyed by Coin.id ("RUNE-THORChain"), not priceProviderID ("thorchain"), so
+        // the lookup must use the id or it can never hit and every call re-fetches live. The live
+        // fallback fetches RUNE explicitly in USD: getPriceByPriceProviderId returns the app
+        // currency, and callers multiply this result by tetherPrice (currency-per-USD), so a
+        // non-USD value here would double-apply FX.
+        getCachedPrice(Coins.ThorChain.RUNE.id, AppCurrency.USD) ?: fetchRunePriceUsdLive()
+
+    private suspend fun fetchRunePriceUsdLive(): BigDecimal =
```
rkokhatskyi, line 540: This RUNE fix is not mirrored to its sTCY twin at `TokenPriceRepository.kt:451`, which still reads `getCachedPrice("tcy", USD) ?: getPriceByPriceProviderId("tcy")`: that key never hits since rows key on `TCY-THORChain`, and the fallback returns the app currency which line 472 then multiplies by tetherPrice, so a JPY user sees sTCY roughly 150x too high

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -490,8 +496,23 @@ constructor(
...
 
     private suspend fun runePriceUsd(): BigDecimal =
-        getCachedPrice(Coins.ThorChain.RUNE.priceProviderID, AppCurrency.USD)
-            ?: getPriceByPriceProviderId(Coins.ThorChain.RUNE.priceProviderID)
+        // Cache rows are keyed by Coin.id ("RUNE-THORChain"), not priceProviderID ("thorchain"), so
+        // the lookup must use the id or it can never hit and every call re-fetches live. The live
+        // fallback fetches RUNE explicitly in USD: getPriceByPriceProviderId returns the app
+        // currency, and callers multiply this result by tetherPrice (currency-per-USD), so a
+        // non-USD value here would double-apply FX.
+        getCachedPrice(Coins.ThorChain.RUNE.id, AppCurrency.USD) ?: fetchRunePriceUsdLive()
```
rkokhatskyi: `getCachedPrice(RUNE.id, USD)` reads a RUNE `usd` row that is only written while the app currency is USD and is never invalidated on a currency change, so a EUR user who once used USD prices bRUNE/ybRUNE off a frozen stale RUNE quote. Gate the cache hit on the current currency being USD, else fetch live

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -479,6 +479,12 @@ constructor(
                         }
                         .awaitAll()
                         .filterNotNull()
+                        // Drop zero prices rather than persist them. A transient failure (an
+                        // unparseable NAV, a rate-limited RUNE price) yields 0, and savePrices
+                        // only guards an empty currency map, so a $0 would overwrite the
+                        // last-known good price. signum() (not `!= ZERO`) is used so a scaled
+                        // zero like 0E-8 from a NAV division still counts as zero.
+                        .filter { (_, prices) -> prices.values.any { it.signum() != 0 } }
```
rkokhatskyi, line 511: The comment says an unparseable NAV yields 0 and is dropped, but `navPerShareFromStatus` returns 1 when `liquid_bond_shares` is empty (it defaults to "" in `VaultRedemptionResponseJson.kt:16`), so a malformed 2xx prices ybRUNE/sTCY at nonzero RUNE parity that this filter keeps, overwriting the accrued NAV price

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/workers/TokenRefreshWorker.kt
aminsato: Good catch — fixed in dac839202. `needsIdentityCorrection` now returns early unless the persisted and refreshed `contractAddress` match case-insensitively, so a different contract sharing an existing ticker can no longer replace the row and flip the send target. Only same-token recasing/decimal fixes survive. Added a regression test (`doWork does not hijack a token when a different contract shares an existing ticker`) alongside the existing recasing tests.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 6c73c0cb8. Added `tcyPriceUsd(currency)` mirroring `runePriceUsd`: it reads the cache by `Coins.ThorChain.TCY.id` (not the `tcy` provider id) and falls back to a live TCY-in-USD fetch, so the value is USD-clean before NAV and the caller apply their multipliers — no more double FX. Covered by `sTCY prices off the TCY cache keyed by coin id, not the tcy price provider` and `sTCY live TCY fallback fetches TCY in USD and applies FX only once`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 6c73c0cb8. Both `runePriceUsd` and `tcyPriceUsd` now consult the USD cache row only when the current app currency is USD; otherwise they fetch live. A non-USD user can no longer read a frozen stale USD row left over from a previous USD session.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 6c73c0cb8. `navPerShareFromStatus` now returns `BigDecimal.ZERO` (dropped by the zero-filter) when `liquid_bond_size`/`liquid_bond_shares` are unparseable/empty, instead of falling through to `1`. Parity (`1`) is reserved for a genuine, fully-parsed pre-bond state where shares is present and zero. Regression test: `does not persist a nonzero parity price when the NAV status is empty`.

rkokhatskyi (approved):
LGTM

PR #5285, opened 2026-07-14
fix(address-book): list vaults without the chain enabled in My Vaults (#5282)

### LINKED ISSUE #5282: Send: vaults without the chain enabled are missing from Address Book → My Vaults
### Description

When sending a token, the **Address Book → My Vaults** tab only lists vaults that already have the selected asset's chain **enabled**. If a destination vault owns an address on that chain but hasn't added/enabled the chain, it does **not** appear in the list — so the user cannot pick it as a recipient even though the vault can receive that token.

Every vault deterministically derives an address for every supported chain from its public keys, so a vault should be selectable as a destination regardless of whether it has explicitly enabled that chain.

### Steps to Reproduce

1. Have (at least) two vaults on the device.
2. In **Vault A**, start a **Send** of an asset (e.g. XRP).
3. On the Send screen, open **Address Book** and switch to the **My Vaults** tab.
4. Make sure **Vault B** does **not** have that chain (XRP) enabled.

### Expected

**Vault B** is listed under **My Vaults**, using its derived address for the selected chain, so it can be chosen as the destination.

### Actual

**Vault B** is not shown. Only vaults that already have the chain enabled appear in the **My Vaults** list.

### Technical note (verified)

`AddressBookBottomSheetViewModel` builds the "My Vaults" list from each vault's **already-enabled coins**:

```kotlin
vaultRepository.getAll().mapNotNull { vault ->
    vault.coins
        .find { it.chain.id == args.chainId }   // only enabled chains have a coin
        ?.address
        ?.let { existingAddress -> ... }
}
```

Vaults without a coin for `args.chainId` are dropped by `mapNotNull`. The fix should derive the address from the vault's public key (e.g. via `ChainAccountAddressRepository`) as a fallback when the chain isn't enabled, rather than relying solely on `vault.coins`.

### Environment

- App: Vultisig Android
- Screen: Send → Address Book → My Vaults

aminsato:
## Summary

Fixes #5282.

In **Send → Address Book → My Vaults**, only vaults that already had the selected asset's chain **enabled** were listed. A destination vault that can receive the token — every vault deterministically derives an address for every supported chain — was dropped if it hadn't explicitly enabled that chain, so it could not be picked as a recipient.

## Root cause

`AddressBookBottomSheetViewModel` built the "My Vaults" list solely from each vault's already-enabled `vault.coins`. Vaults with no coin for `args.chainId` were silently removed by `mapNotNull`.

## Fix

- Keep the enabled-coin address as the preferred source.
- Fall back to deriving the address from the vault's public key via `ChainAccountAddressRepository.getAddress(chain, vault)` when the chain isn't enabled, so every vault is selectable using its derived address.
- The derivation rethrows `CancellationException` and skips only genuinely un-derivable vaults, so one failure can't drop the whole list.

## Test plan

- [ ] Two vaults on device; Vault B does **not** have XRP enabled.
- [ ] Start a Send of XRP from Vault A → Address Book → My Vaults.
- [ ] Verify Vault B now appears, showing its derived XRP address, and can be selected as the destination.
- [ ] Verify vaults with the chain already enabled still show their existing address (no regression).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved address book behavior so eligible vaults can be shown and selected even if the chain-specific coin address hasn’t been enabled yet.
  * Added a smarter fallback for address display: if an existing address isn’t available, the app derives one to ensure the correct entries appear.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/transaction/AddressBookBottomSheetViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/transaction/AddressBookBottomSheetViewModel.kt
@@ -17,6 +17,7 @@ import com.vultisig.wallet.ui.navigation.Route
 import com.vultisig.wallet.ui.navigation.back
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
@@ -73,23 +74,30 @@ constructor(
         }
 
         viewModelScope.launch {
+            val chain = Chain.fromRaw(args.chainId)
             val vaults =
                 vaultRepository.getAll().mapNotNull { vault ->
-                    vault.coins
-                        .find { it.chain.id == args.chainId }
-                        ?.address
-                        ?.let { existingAddress ->
-                            AddressEntryUiModel(
-                                model =
-                                    AddressBookEntry(
-                                        chain = Chain.fromRaw(args.chainId),
-                                        address = existingAddress,
-                                        title = vault.name,
-                                    ),
-                                title = vault.name,
-                                subtitle = existingAddress,
-                            )
-                        }
+                    // Prefer the already-enabled coin's address; otherwise derive the address
+                    // from the vault's public key so vaults that haven't enabled this chain are
+                    // still selectable as a destination.
+                    val address =
+                        vault.coins.find { it.chain.id == args.chainId }?.address
+                            ?: try {
+                                chainAccountAddressRepository.getAddress(chain, vault).first
+                            } catch (e: CancellationException) {
+                                throw e
+                            } catch (_: Exception) {
+                                null
+                            }
+
+                    address?.let {
+                        AddressEntryUiModel(
+                            model =
+                                AddressBookEntry(chain = chain, address = it, title = vault.name),
+                            title = vault.name,
+                            subtitle = it,
+                        )
+                    }
                 }
 
             state.update { it.copy(vaults = vaults) }
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/transaction/AddressBookBottomSheetViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/transaction/AddressBookBottomSheetViewModel.kt
@@ -6,6 +6,7 @@ import androidx.lifecycle.viewModelScope
 import androidx.navigation.toRoute
 import com.vultisig.wallet.data.models.AddressBookEntry
 import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.hasPreGeneratedKey
 import com.vultisig.wallet.data.models.logo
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/vultisig/wallet/ui/screens/transaction/AddressBookBottomSheetViewModel.kt
```diff
@@ -73,23 +74,30 @@ constructor(
...
-                                    ),
-                                title = vault.name,
-                                subtitle = existingAddress,
-                            )
-                        }
+                    // Prefer the already-enabled coin's address; otherwise derive the address
+                    // from the vault's public key so vaults that haven't enabled this chain are
+                    // still selectable as a destination.
+                    val address =
+                        vault.coins.find { it.chain.id == args.chainId }?.address
```
rkokhatskyi, line 92: Lists KeyImport vaults for chains they hold no pre-generated key for, yet `hasPreGeneratedKey` blocks enabling those chains (`ChainSelectionViewModel.kt:202`), so an ETH-only import shows as a BTC destination it can never view or spend in app, unlike Windows which skips such vaults
```kotlin
if (!vault.hasPreGeneratedKey(chain)) return@mapNotNull null
```

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/AddressBookBottomSheetViewModel.kt
```diff
@@ -73,23 +74,30 @@ constructor(
...
+                            ?: try {
+                                chainAccountAddressRepository.getAddress(chain, vault).first
```
rkokhatskyi: This runs `Tss.getDerivedPubKey` and WalletCore per vault on `viewModelScope`'s main dispatcher with no `withContext`, once for every vault that has not enabled the chain, so opening the sheet with many vaults can block the UI thread, needs device confirmation, wrap in `withContext(Dispatchers.Default)`

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/AddressBookBottomSheetViewModel.kt
aminsato: Good catch — added `if (!vault.hasPreGeneratedKey(chain)) return@mapNotNull null` so KeyImport vaults without a pre-generated key for the chain are skipped, matching the Windows behavior. Fixed in df75c1f.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/AddressBookBottomSheetViewModel.kt
aminsato: Wrapped the derivation in `withContext(Dispatchers.Default)` so `Tss.getDerivedPubKey`/WalletCore no longer runs on the main dispatcher. Fixed in df75c1f — still worth a device smoke test with many vaults to confirm no jank.

johnnyluo:
Let's not do that , if the vault doesn't have the chain you are trying to send , don't show it , keep current behaviour

PR #5288, opened 2026-07-14
fix(ripple): don't show $0.00 on a balance-fetch network failure (#5276)

### LINKED ISSUE #5276: XRP balance shows $0.00 on network timeout instead of a loading/error state
## Summary

When the balance request for a chain fails due to a network error (e.g. `SocketTimeoutException`), the coin/chain screen renders the balance as **zero** instead of surfacing a loading or error state. The user cannot distinguish "we couldn't reach the network" from "your account is empty", which is dangerous — it looks like the funds disappeared.

Observed on **Ripple (XRP)**, but this is a general balance-loading concern that likely affects other chains that fall back to a zero balance on failure.

## Steps to reproduce

1. Open a vault that holds XRP.
2. Open the Ripple coin screen while the network to the RPC node (`xrplcluster.com`) is slow/unreachable so the balance call times out.
3. Observe the balance.

## Actual behavior

- Total shows `$0.00`
- XRP row shows `0 XRP`
- The XRP price still loads correctly (`$1.07`), so only the balance failed.

## Expected behavior

- On a failed balance fetch, keep showing a **loading indicator** (or the last known cached balance), or show an explicit error/retry affordance — do **not** silently display `0`.
- A network failure must be visually distinct from a genuinely empty account.

## Root cause (verified)

`RippleApi.getBalance()` swallows every non-cancellation exception and returns `BigInteger.ZERO`:

```
} catch (e: Exception) {
    Timber.e("Error in getBalance: ${e.message}")
    BigInteger.ZERO   // <-- network failure is indistinguishable from empty account
}
```

`data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt`

A timeout/connection error should propagate (or resolve to a distinct "unknown" state) so the UI can keep loading / show an error instead of committing a zero balance. Other chain balance implementations should be audited for the same `catch -> return ZERO` pattern.

## Relevant logs (redacted)

```
Ktor Client  REQUEST https://xrplcluster.com failed with exception:
    java.net.SocketTimeoutException: Socket timeout has expired [url=https://xrplcluster.com]
RippleApiImp            E  Error in fetchTokenAccountsByOwner: Connection timed out
RippleApiImp$getBalance E  Error in getBalance: Connection timed out
Ktor Client  REQUEST https://api.vultisig.com/coingeicko/.../simple/price?ids=ripple...
    -> RESPONSE 200 {"ripple":{"usd":1.07}}   // price OK, balance failed
```

## Environment

- Chain: Ripple (XRP)
- App build from branch under active development

aminsato:
## Summary

Closes #5276.

`RippleApi.getBalance()` caught **every** non-cancellation exception and returned `BigInteger.ZERO`. A network failure (e.g. `SocketTimeoutException` to `xrplcluster.com`) was therefore indistinguishable from a genuinely empty account, and the UI committed a false **`$0.00` / `0 XRP`** — which reads to the user as "the funds disappeared".

## Change

- **`getBalance()`** — no longer swallows failures into `ZERO`; the exception propagates. An unfunded account (`actNotFound`) still returns HTTP 200 with a null `account_data`, so it resolves to a real zero **without throwing** — the one legitimate zero.
- **`fetchAccountsInfo()`** — rethrows the original exception instead of flattening it into an `IllegalStateException` with no cause, so the transport kind (timeout / no connectivity) survives for `toNetworkErrorKind()` classification. Also fixed a stale copy-paste log tag.

## Why this is the right fix

`AccountsRepository.loadAddressBalances` already `catch`es a thrown balance error and **keeps the cached balance** rather than overwriting it, and both `ChainValidationService` XRP callers catch generic `Exception`. So propagating gives the exact behavior the issue asks for, and matches how EVM/Cosmos balances already behave (they never swallow to `ZERO`).

## Resulting UX on a network failure

| Situation | Before | After |
|---|---|---|
| Warm start (balance cached in Room) | `$0.00` | Last-known balance retained |
| Cold start / brand-new coin (no cache) | `$0.00` | Loading placeholder stays (no false zero) |

The issue's expected behavior listed *keep loading*, *show last-known cached*, or *explicit error/retry*. This PR delivers the first two; it does not add a dedicated error/retry indicator (that would require threading a per-chain error status through the balance UI).

## Tests

New `RippleApiBalanceTest` (4 tests, all green):
- funded account nets balance against reserves
- unfunded account resolves to `0` without throwing
- a socket timeout propagates as `NetworkException(kind = Timeout)`
- `fetchAccountsInfo` preserves the original error kind

Existing `RippleApiBodyReadTest` (5 tests) still pass. `./gradlew :data:testDebugUnitTest` green.

## Scope note

The issue notes other chains may share the `catch → return ZERO` pattern (Solana, Polkadot, Tron, Cardano, Bittensor, ThorChain, Ton). This PR is intentionally scoped to XRP, the reported chain; happy to follow up on the others.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Ripple balance calculations to subtract required reserves and clamp results to non-negative values.
  * Unfunded accounts now return a zero balance without throwing.
  * Network/timeout failures are no longer swallowed; the original error type and timeout classification are propagated for consistent handling.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
@@ -79,25 +79,24 @@ internal class RippleApiImp @Inject constructor(private val http: HttpClient) : 
         return response.bodyOrThrow<RippleBroadcastSuccessResponseJson>()
     }
 
+    // A network failure (timeout / no connectivity) must propagate so the balance layer can keep
+    // the last-known value or surface a loading/error state — never swallow it into ZERO, which is
+    // indistinguishable from a genuinely empty account and reads as "the funds disappeared". An
+    // unfunded account returns an HTTP 200 with a null `account_data` (actNotFound), so it still
+    // resolves to zero here without throwing.
     override suspend fun getBalance(coin: Coin): BigInteger = supervisorScope {
-        try {
-            val accountInfoDeferred = async { fetchAccountsInfo(coin.address) }
-            val reservedBalanceDeferred = async { fetchServerState() }
+        val accountInfoDeferred = async { fetchAccountsInfo(coin.address) }
+        val reservedBalanceDeferred = async { fetchServerState() }
 
-            val accountInfo = accountInfoDeferred.await()
-            val reservedBalance = reservedBalanceDeferred.await()
+        val accountInfo = accountInfoDeferred.await()
+        val reservedBalance = reservedBalanceDeferred.await()
 
-            val balance = accountInfo?.getBalance() ?: BigInteger.ZERO
-            val ownerCount = accountInfo?.getOwnerCount() ?: BigInteger.ZERO
-            val accountReservedBalance =
-                reservedBalance.getBaseReserve() + (ownerCount * reservedBalance.getIncReserve())
+        val balance = accountInfo?.getBalance() ?: BigInteger.ZERO
+        val ownerCount = accountInfo?.getOwnerCount() ?: BigInteger.ZERO
+        val accountReservedBalance =
+            reservedBalance.getBaseReserve() + (ownerCount * reservedBalance.getIncReserve())
 
-            maxOf(balance - accountReservedBalance, BigInteger.ZERO)
-        } catch (e: Exception) {
-            if (e is kotlinx.coroutines.CancellationException) throw e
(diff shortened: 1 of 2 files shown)


rkokhatskyi (approved):
LGTM

PR #5297, opened 2026-07-15
fix(swap): stop placeholder skeletons blinking on empty amount

### LINKED ISSUE #5296: Swap form: destination/fee placeholders blink when amount field is empty
### Describe the bug

When the swap form is opened, or when the source/destination token is changed, the destination amount and fee **skeleton placeholders blink** (a shimmer flashes on and then off) even though the amount field is empty and there is nothing to load.

### Root cause

In `SwapQuotePipelineController` the loading spinner (`isLoading`) is raised unconditionally at three points in the quote pipeline, regardless of whether there is actually a quote to fetch:

- The main input `onEach` fires for **every** emission of the amount flow, including the empty-field emission that occurs on form open / pair change.
- The `slippageBps` and `externalRecipient` StateFlows emit their initial values the moment the pipeline subscribes (form open), and each one flips `isLoading = true` via its inline `onEach`.

Roughly ~300ms later, `collectLatest` observes the empty field, `resolveQuote` returns `Empty`, and `resetQuoteState()` flips `isLoading = false`. That `true → false` round-trip with no data behind it is exactly the visible "blink" of the destination/fee skeletons.

### Expected behavior

The loading skeletons should only appear when there is genuinely a quote being fetched — i.e. a routable pair **and** a positive source amount. With an empty/zero amount field, the placeholders must stay quiet (no blink) on form open, on a bare pair/destination change, and on a slippage/recipient change.

### Steps to reproduce

1. Open the Swap form (do not type an amount).
2. Observe the destination amount / fee placeholders shimmer briefly.
3. Change the destination token (still without an amount) — the placeholders blink again.

### Affected area

- `app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt`

aminsato:
## Summary

Fixes the swap-form skeleton placeholders that **blink** (shimmer flashes on, then off) when the form is opened or when the source/destination token is changed while the amount field is empty — even though there is no data to load.

**Root cause** (`SwapQuotePipelineController`): `isLoading` was raised unconditionally at three points in the quote pipeline:
- The main input `onEach` fired for every emission, including the empty-field emission on form open / pair change.
- The `slippageBps` and `externalRecipient` StateFlows emit their initial values on subscription (form open), each flipping `isLoading = true`.

~300ms later `collectLatest` sees the empty field, `resolveQuote` returns `Empty`, and `resetQuoteState()` flips `isLoading = false`. That `true → false` round-trip with no data behind it is the visible blink.

**Fix:** a guarded helper `startLoadingIfQuotable()` that only raises the spinner when there is genuinely something to quote — a routable pair **and** a positive source amount — used at all three sites. Empty/zero fields now stay quiet on form open, on a bare pair/destination change, and on a slippage/recipient change. Real typing (positive amount) still shows the skeleton and fetches a quote exactly as before.

## Test plan

- [x] `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.swap.SwapFormViewModelTest"` — green
- [x] New regression test `empty amount never raises the loading spinner on form open` asserts `isLoading` stays `false` across the full debounce window and no quote is fetched for an empty field
- [x] Existing `calculateFees debounces amount changes` (spinner true while a real amount is loading) still passes — behavior preserved for positive amounts

Closes #5296

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced brief swap loading/spinner flickers during form setup and when adjusting pair, slippage, or recipient while the source amount is empty or zero.
  * Loading and quote UI now trigger only for routable token pairs; when inputs become non-quotable, the UI clears stale quote/fee state and prevents late results from showing.
* **Tests**
  * Expanded debounce and edge-case coverage for spinner behavior, routability gating, and correct state clearing.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -10,6 +10,7 @@ import androidx.navigation.toRoute
 import com.vultisig.wallet.R
 import com.vultisig.wallet.data.models.Address
 import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.Coins
 import com.vultisig.wallet.data.models.TokenStandard
 import com.vultisig.wallet.data.models.TokenValue
@@ -365,12 +366,27 @@ constructor(
                 selectedDstId = selectedDstId,
                 addresses = addresses,
                 uiState = _uiState,
-                // With an empty/zero amount there is nothing to quote, so loading a not-yet-held
-                // token's account must not raise the quote skeletons — that reproduces the blink
-                // this gate removes on the pipeline side (#5296 review).
-                hasQuotableAmount = srcAmount?.let { it > BigDecimal.ZERO } ?: false,
+                // Raise the quote skeletons while loading a not-yet-held token's account only when
+                // the pair the pick forms could actually be quoted — a positive amount AND a
+                // routable (distinct, provider-backed) pair, mirroring the pipeline's own
+                // isPairRoutable && amount>0 gate. A bare amount>0 check would still blink over an
+                // unroutable or same-token pair before updatePairSupport catches it (#5296 review).
+                isSelectionQuotable = { selectedToken ->
+                    isSelectionQuotable(targetArg, selectedToken)
+                },
             )
         }
+    }
+
+    private fun isSelectionQuotable(targetArg: String, selectedToken: Coin): Boolean {
+        val amount = srcAmount ?: return false
+        if (amount <= BigDecimal.ZERO) return false
+        val (src, dst) =
+            when (targetArg) {
+                ARG_SELECTED_SRC_TOKEN_ID -> selectedToken to selectedDst.value?.account?.token
+                else -> selectedSrc.value?.account?.token to selectedToken
+            }
+        return src != null && dst != null && quotePipeline.isPairRoutable(src, dst)
     }
 
     fun flipSelectedTokens() {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
@@ -365,6 +365,10 @@ constructor(
                 selectedDstId = selectedDstId,
                 addresses = addresses,
                 uiState = _uiState,
+                // With an empty/zero amount there is nothing to quote, so loading a not-yet-held
+                // token's account must not raise the quote skeletons — that reproduces the blink
+                // this gate removes on the pipeline side (#5296 review).
+                hasQuotableAmount = srcAmount?.let { it > BigDecimal.ZERO } ?: false,
             )
         }
     }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
@@ -145,6 +145,10 @@ constructor(
     // SwapIsNotSupported only after the user typed an amount and waited out the debounce).
     private var isPairSupported = true
 
+    // A same-token pair is "supported" (no "no route" guidance while mid-pick) but has no provider
+    // and can never be quoted, so the loading gate keys off routability, not mere support (#5296).
+    private var isPairRoutable = false
+
     private val pairNotSupportedError = UiText.StringResource(R.string.swap_route_not_available)
 
     private val srcAmount: BigDecimal?
@@ -317,12 +321,13 @@ constructor(
                 }
                 .combine(refreshQuoteState) { input, _ -> input }
                 // A slippage or external-recipient change re-fetches with a different tolerance /
-                // routing, so raise isLoading to disable the Swap button until the new quote lands
-                // —
-                // otherwise the prior, differently-routed quote could be signed (#4858, review
-                // #4969). The onEach rides each flow individually, so the silent refresh timer
-                // above
-                // doesn't flash the spinner.
+                // routing, so — when there is actually something to quote — raise isLoading to
+                // disable the Swap button until the new quote lands, otherwise the prior,
+                // differently-routed quote could be signed (#4858, review #4969). Routed through
+                // startLoadingIfQuotable so an empty/zero amount stays quiet instead of blinking
+                // the
+                // skeletons (#5296). The onEach rides each flow individually, so the silent refresh
+                // timer above doesn't flash the spinner.
                 .combine(slippageBps.onEach { startLoadingIfQuotable() }) { input, _ -> input }
                 .combine(externalRecipient.onEach { startLoadingIfQuotable() }) { input, _ ->
                     input
@@ -463,15 +468,22 @@ constructor(
     }
 
     /**
-     * Raise the loading spinner only when there is genuinely a quote to fetch — a routable pair with
-     * a positive source amount. An empty or zero field has nothing to load, so gating here stops the
-     * destination/fee skeletons from flashing (isLoading true→false) on form open, a bare pair
-     * change, or a slippage/recipient change while the amount field is still empty (#4712).
+     * Reconciles quote-driven UI state with the current input on a trigger (typing, pair, slippage,
+     * or recipient change), ahead of the debounced fetch:
+     * - Routable pair (a real provider, so same-token is excluded) with a positive source amount:
+     *   raise the spinner so the destination/fee skeletons and disabled Swap button lead the fetch.
+     * - Nothing to quote (empty/zero field or an unroutable pair): leave the spinner off so the
+     *   skeletons never flash true→false (blink) on form open or a bare pair/slippage/recipient
+     *   change (#4712, #5296). If a resolved quote is still on screen, clear it now so a cleared or
+     *   zeroed amount disables Swap and drops the stale destination/fee immediately instead of
+     *   leaving them tappable until the 300ms debounce runs resetQuoteState (#5296 review).
      */
     private fun startLoadingIfQuotable() {
         val amount = srcAmount
-        if (isPairSupported && amount != null && amount > BigDecimal.ZERO) {
+        if (isPairRoutable && amount != null && amount > BigDecimal.ZERO) {
             isLoading = true
+        } else if (uiState.value.quoteDisplay.hasQuote) {
+            resetQuoteState()
         }
     }
 
@@ -569,9 +581,10 @@ constructor(
     private fun updatePairSupport(src: SendSrc, dst: SendSrc) {
         val srcToken = src.account.token
         val dstToken = dst.account.token
-        isPairSupported =
-            srcToken == dstToken ||
+        isPairRoutable =
+            srcToken != dstToken &&
                 swapQuoteRepository.getEligibleProviders(srcToken, dstToken).isNotEmpty()
+        isPairSupported = srcToken == dstToken || isPairRoutable
         if (!isPairSupported) {
             resetQuoteState(error = pairNotSupportedError, cause = null, tag = null)
         } else if (uiState.value.formError == pairNotSupportedError) {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
@@ -155,6 +155,19 @@ constructor(
     private val srcAmount: BigDecimal?
         get() = srcAmountState.text.toString().toBigDecimalOrNull()
 
+    /**
+     * True when the live field currently forms a quotable request: a routable pair (a real
+     * provider, so same-token is excluded) with a positive source amount. Shared by
+     * [startLoadingIfQuotable] and the late-result guard in [applyQuoteResult] so both agree on
+     * what "quotable" means, keeping zero/invalid amounts and unroutable pairs quiet on both the
+     * loading and result paths (#5296).
+     */
+    private val isLiveInputQuotable: Boolean
+        get() {
+            val amount = srcAmount
+            return isPairRoutable && amount != null && amount > BigDecimal.ZERO
+        }
+
     private var isLoading: Boolean
         get() = uiState.value.isLoading
         set(value) {
@@ -385,13 +398,15 @@ constructor(
         input: QuoteInput,
         result: SwapQuotePipelineResult.Success,
     ) {
-        // isAmountFieldEmpty is read once when resolveQuote is called, but the field can be cleared
+        // isAmountFieldEmpty is read once when resolveQuote is called, but the live input can
+        // change
         // while this fetch — queued on a prior debounce cycle — is still in flight, and
         // collectLatest
-        // can't cancel it until the empty input clears the debounce. Re-read the live field here so
+        // can't cancel it until the next input clears the debounce. Re-check the live input here so
         // a
-        // late-landing fetch drops the now-stale quote instead of resurrecting it (#5296 review).
-        if (srcAmountState.text.isEmpty()) {
+        // late-landing fetch for a now non-quotable field (cleared, zeroed, or an unroutable pair)
+        // drops the stale quote instead of resurrecting it (#5296 review).
+        if (!isLiveInputQuotable) {
             resetQuoteState()
             return
         }
@@ -491,8 +506,7 @@ constructor(
      *   leaving them tappable until the 300ms debounce runs resetQuoteState (#5296 review).
      */
     private fun startLoadingIfQuotable() {
-        val amount = srcAmount
-        if (isPairRoutable && amount != null && amount > BigDecimal.ZERO) {
+        if (isLiveInputQuotable) {
             isLoading = true
         } else if (uiState.value.quoteDisplay.hasQuote || uiState.value.isLoading) {
             // Clear a resolved quote OR a spinner we raised while a firm quote was still pending:
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
@@ -4,6 +4,7 @@ import androidx.compose.foundation.text.input.TextFieldState
 import com.vultisig.wallet.R
 import com.vultisig.wallet.data.IoDispatcher
 import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.FiatValue
 import com.vultisig.wallet.data.models.SwapQuote
 import com.vultisig.wallet.data.models.TokenStandard
@@ -384,6 +385,17 @@ constructor(
         input: QuoteInput,
         result: SwapQuotePipelineResult.Success,
     ) {
+        // isAmountFieldEmpty is read once when resolveQuote is called, but the field can be cleared
+        // while this fetch — queued on a prior debounce cycle — is still in flight, and
+        // collectLatest
+        // can't cancel it until the empty input clears the debounce. Re-read the live field here so
+        // a
+        // late-landing fetch drops the now-stale quote instead of resurrecting it (#5296 review).
+        if (srcAmountState.text.isEmpty()) {
+            resetQuoteState()
+            return
+        }
+
         val (src, _) = input.address
 
         quoteState.provider = result.provider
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -460,6 +462,19 @@ constructor(
         }
     }
 
+    /**
+     * Raise the loading spinner only when there is genuinely a quote to fetch — a routable pair with
+     * a positive source amount. An empty or zero field has nothing to load, so gating here stops the
+     * destination/fee skeletons from flashing (isLoading true→false) on form open, a bare pair
+     * change, or a slippage/recipient change while the amount field is still empty (#4712).
+     */
+    private fun startLoadingIfQuotable() {
```
rkokhatskyi, line 508: `startLoadingIfQuotable` only gates this pipeline's own isLoading writes, but `SwapTokenSelector.checkTokenSelectionResponse` still does `uiState.update { isLoading = true }` unconditionally whenever a not-yet-held token is picked (`SwapTokenSelector.kt:196`), so picking a new swap destination with an empty amount reproduces the same blink.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -312,7 +312,7 @@ constructor(
                     // Unroutable pair: the "no route" guidance already showed on selection (#4710),
                     // so don't spin or fetch an indicative estimate for a pair we can't quote.
                     if (!isPairSupported) return@onEach
-                    isLoading = true
+                    startLoadingIfQuotable()
```
rkokhatskyi, line 333: `startLoadingIfQuotable` only ever sets `isLoading`, never clears it, and no-ops once `amount<=0`, so clearing or zeroing the field over an already-resolved quote leaves the Swap button enabled and the stale destination/fee numbers on screen until the 300ms debounce runs `resetQuoteState`.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -323,8 +323,10 @@ constructor(
                 // #4969). The onEach rides each flow individually, so the silent refresh timer
                 // above
                 // doesn't flash the spinner.
-                .combine(slippageBps.onEach { isLoading = true }) { input, _ -> input }
-                .combine(externalRecipient.onEach { isLoading = true }) { input, _ -> input }
+                .combine(slippageBps.onEach { startLoadingIfQuotable() }) { input, _ -> input }
```
rkokhatskyi, line 345: This centralizes the slippage/recipient blink-guard from #4858/#4969 into `startLoadingIfQuotable`, but no test asserts `isLoading` still gets raised here with a positive amount already entered, so a future regression to this shared gate that silently drops the Swap-button-disable safety net would pass CI.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -460,6 +462,19 @@ constructor(
         }
     }
 
+    /**
+     * Raise the loading spinner only when there is genuinely a quote to fetch — a routable pair with
```
rkokhatskyi: This docstring calls the gate `isPairSupported` "a routable pair," but `updatePairSupport`'s own doc further down this file says same-token pairs are deliberately marked supported with no route at all, so the wording overclaims what the check actually guarantees.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -323,8 +323,10 @@ constructor(
                 // #4969). The onEach rides each flow individually, so the silent refresh timer
                 // above
                 // doesn't flash the spinner.
-                .combine(slippageBps.onEach { isLoading = true }) { input, _ -> input }
-                .combine(externalRecipient.onEach { isLoading = true }) { input, _ -> input }
+                .combine(slippageBps.onEach { startLoadingIfQuotable() }) { input, _ -> input }
+                .combine(externalRecipient.onEach { startLoadingIfQuotable() }) { input, _ ->
```
rkokhatskyi, line 346: The comment above still says a slippage/recipient change unconditionally "raise[s] isLoading," but the call here now goes through `startLoadingIfQuotable`, which no-ops for an empty or zero amount, so the comment no longer describes this code's behavior for that case.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Good catch — fixed in da92d3f39. `checkTokenSelectionResponse` now only raises the shared `isLoading` while loading a not-yet-held token's account when there is a quotable amount (`hasQuotableAmount`, threaded from `SwapFormViewModel.navigateToSelectToken`). Picking a new destination with an empty field no longer flashes the quote skeletons; the account still loads silently.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in da92d3f39. `startLoadingIfQuotable` now also does `else if (uiState.value.quoteDisplay.hasQuote) resetQuoteState()`, so clearing or zeroing the amount over a resolved quote disables Swap and drops the stale destination/fee immediately instead of waiting out the 300ms debounce. Covered by the new `clearing the amount over a resolved quote disables swap and drops the stale quote at once` test.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Added in da92d3f39: `a slippage change with a positive amount still raises the loading spinner` — enters an amount, lets the quote settle, then changes slippage and asserts `isLoading` goes back to true. That guards the #4858/#4969 Swap-button-disable safety net against a future regression to the shared gate.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in da92d3f39. The gate now keys off a new `isPairRoutable` flag (`srcToken != dstToken && getEligibleProviders(...).isNotEmpty()`) rather than `isPairSupported`, and the docstring no longer conflates "supported" with "routable" — same-token pairs are supported but explicitly not routable, so they never spin.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in da92d3f39 — reworded the comment to say the slippage/external-recipient change routes through `startLoadingIfQuotable`, which stays quiet for an empty/zero amount instead of unconditionally raising `isLoading`.

rkokhatskyi:
@aminsato build is failing

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -931,6 +932,166 @@ internal class SwapFormViewModelTest {
...
+
+            val vm = createViewModelWithSwapTokens()
+            advanceUntilIdle()
+
+            vm.srcAmountState.setTextAndPlaceCursorAtEnd("10")
+            Snapshot.sendApplyNotifications()
+            advanceTimeBy(400)
+            advanceUntilIdle()
+            assertTrue(vm.uiState.value.quoteDisplay.hasQuote)
+            assertFalse(vm.uiState.value.isSwapDisabled)
```
rkokhatskyi, line 1092: This test fails as written: `createViewModelWithSwapTokens` defaults `ethBalance` to 1 ETH but the test types "10", tripping `SwapValidator`'s insufficient-balance guard, so `isSwapDisabled` is `true` here, not `false`; bump the fixture balance.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
```diff
@@ -931,6 +932,166 @@ internal class SwapFormViewModelTest {
...
+
+            vm.srcAmountState.setTextAndPlaceCursorAtEnd("10")
+            Snapshot.sendApplyNotifications()
+            advanceTimeBy(400)
+            advanceUntilIdle()
+            assertFalse(vm.uiState.value.isLoading)
+
+            vm.setSlippageBps(300)
+            runCurrent()
+            assertTrue(vm.uiState.value.isLoading)
```
rkokhatskyi, line 1055: This also fails: `setTextAndPlaceCursorAtEnd("10")` from empty is a 2-char jump that trips `amountChanges`'s paste heuristic, caching `immediate=true` so the slippage-triggered refetch also debounces at 0ms and resolves before this assertion runs.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTokenSelector.kt
```diff
@@ -190,19 +192,25 @@ constructor(
...
     ) {
         val result = requestResultRepository.request<AssetSelected>(targetArg) ?: return
 
         if (result.isDisabled) {
-            uiState.update { it.copy(isLoading = true) }
+            // Only raise the shared loading flag when there is an amount to quote; otherwise
+            // loading
+            // a not-yet-held token's account would flash the destination/fee skeletons true→false
+            // over an empty field — the same blink the pipeline gate removes (#5296 review).
+            val showLoading = hasQuotableAmount
```
rkokhatskyi: `hasQuotableAmount` only checks `srcAmount>0`, unlike the pipeline's own gate `isPairRoutable && amount>0` (`SwapQuotePipelineController.kt:483`), so picking a not-yet-held token forming an unroutable or same-token pair still blinks `isLoading` during the account load before `updatePairSupport` catches it.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -365,6 +365,10 @@ constructor(
                 selectedDstId = selectedDstId,
                 addresses = addresses,
                 uiState = _uiState,
+                // With an empty/zero amount there is nothing to quote, so loading a not-yet-held
+                // token's account must not raise the quote skeletons — that reproduces the blink
+                // this gate removes on the pipeline side (#5296 review).
+                hasQuotableAmount = srcAmount?.let { it > BigDecimal.ZERO } ?: false,
```
rkokhatskyi: No test calls `selectSrcToken()`/`selectDstToken()` anywhere, so this `hasQuotableAmount` gate — the fix for a previously flagged Major finding — has zero regression coverage; a future revert or inversion here would pass CI silently.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -460,6 +467,26 @@ constructor(
...
+     *   skeletons never flash true→false (blink) on form open or a bare pair/slippage/recipient
+     *   change (#4712, #5296). If a resolved quote is still on screen, clear it now so a cleared or
+     *   zeroed amount disables Swap and drops the stale destination/fee immediately instead of
+     *   leaving them tappable until the 300ms debounce runs resetQuoteState (#5296 review).
+     */
+    private fun startLoadingIfQuotable() {
+        val amount = srcAmount
+        if (isPairRoutable && amount != null && amount > BigDecimal.ZERO) {
+            isLoading = true
+        } else if (uiState.value.quoteDisplay.hasQuote) {
```
rkokhatskyi: This only resets when `hasQuote` is already true. Typing a quotable amount sets `isLoading=true`, then clearing it before the firm quote resolves leaves `hasQuote` false, so this branch no-ops and the spinner stays stuck true for the rest of the debounce.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -460,6 +467,26 @@ constructor(
...
+            resetQuoteState()
```
rkokhatskyi, line 516: This eager reset can be undone: `isAmountFieldEmpty` (line 355) is captured once per `resolveQuote` call and never rechecked, so a fetch from a prior debounce cycle still in flight when the field clears can finish afterward and `applyQuoteResult` resurrects the stale quote.

rkokhatskyi (dismissed):

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Fixed in 7196ce1bf. `setTextAndPlaceCursorAtEnd("10")` from empty is a 2-char jump that trips the paste heuristic, caching `immediate=true` so the slippage refetch debounced at 0ms and resolved before the assertion. The test now types a single-char `"1"` (free typing → 300ms debounce), so `runCurrent()` observes the spinner still up before the debounce elapses.

app/src/test/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModelTest.kt
aminsato: Fixed in 7196ce1bf. Bumped the fixture to 10 ETH (`ethBalance = 10000000000000000000`) and type `"1"`, so the insufficient-balance guard no longer disables Swap and `isSwapDisabled` is `false` at that assertion.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTokenSelector.kt
aminsato: Fixed in 7196ce1bf. The `hasQuotableAmount` boolean is replaced with an `isSelectionQuotable: (Coin) -> Boolean` predicate. `SwapFormViewModel` builds it from the *newly picked* token as `amount > 0 && quotePipeline.isPairRoutable(newPair)`, mirroring the pipeline’s own `isPairRoutable && amount>0` gate (new public `SwapQuotePipelineController.isPairRoutable(src, dst)`, reused inside `updatePairSupport`). So picking a not-yet-held token that forms an unroutable or same-token pair no longer raises `isLoading` during the account load — it stays silent until `updatePairSupport` runs.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in 7196ce1bf. Added three tests that drive `selectDstToken()` / `selectSrcToken()` end-to-end, parking the account load on a `CompletableDeferred` so the assertion lands while only the selection gate (not the follow-up re-quote) has touched `isLoading`: a routable pick raises the spinner, and an unroutable pick and a same-token pick never do. A future revert or inversion of the gate now fails CI.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in 7196ce1bf. The reset branch is now `else if (uiState.value.quoteDisplay.hasQuote || uiState.value.isLoading) resetQuoteState()`. So when a quotable amount raises the spinner but is cleared before the firm quote lands (hasQuote still false), the `isLoading` term still triggers the reset instead of stranding the spinner for the rest of the debounce. Covered by the new `clearing a quotable amount before its quote resolves clears the loading spinner at once` test.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in 7196ce1bf. `applyQuoteResult` now re-reads the live field up front — `if (srcAmountState.text.isEmpty()) { resetQuoteState(); return }` — so a fetch queued on a prior debounce cycle that lands after the field is cleared drops the result instead of resurrecting the stale quote (collectLatest cannot cancel it until the empty input clears the debounce). Covered by `a fetch that lands after the field is cleared never resurrects the stale quote`, which gates the fetch on a `CompletableDeferred` and asserts `hasQuote` never emits true across the landing.

aminsato:
Pushed 7196ce1bf addressing the second review round:

- **Token-pick loading gate** now keys off routability, not just `amount>0` — replaced the `hasQuotableAmount` boolean with an `isSelectionQuotable` predicate that mirrors the pipeline’s `isPairRoutable && amount>0`, plus a new public `SwapQuotePipelineController.isPairRoutable(src, dst)` reused in `updatePairSupport`.
- **`startLoadingIfQuotable`** resets on `hasQuote || isLoading`, so clearing a quotable amount before its firm quote lands no longer strands the spinner.
- **`applyQuoteResult`** re-reads the live field and drops a late-landing fetch instead of resurrecting the stale quote.
- Fixed the two flaky tests (paste-heuristic `immediate` flag; 1 ETH fixture vs `"10"` balance guard) and added 5 tests covering the select-token gate (routable raises / unroutable + same-token don’t), the clear-before-quote reset, and the late-fetch resurrection guard.

`:app:testDebugUnitTest --tests SwapFormViewModelTest` → 111 passing, 0 failures; ktfmt applied.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -377,6 +385,17 @@ constructor(
...
+        // isAmountFieldEmpty is read once when resolveQuote is called, but the field can be cleared
+        // while this fetch — queued on a prior debounce cycle — is still in flight, and
+        // collectLatest
+        // can't cancel it until the empty input clears the debounce. Re-read the live field here so
+        // a
+        // late-landing fetch drops the now-stale quote instead of resurrecting it (#5296 review).
+        if (srcAmountState.text.isEmpty()) {
+            resetQuoteState()
+            return
+        }
```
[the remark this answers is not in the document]
aminsato: Fixed in 427eabf6e. Extracted a shared `isLiveInputQuotable` predicate (`isPairRoutable && amount != null && amount > BigDecimal.ZERO`) and use it in **both** `startLoadingIfQuotable()` and this `applyQuoteResult` guard, so the loading and result paths agree on what "quotable" means. The guard is now `if (!isLiveInputQuotable) { resetQuoteState(); return }` instead of the empty-only check — a fetch queued on a prior debounce cycle that lands while the live field is `"0"`/invalid or forms an unroutable pair now drops the stale quote instead of resurrecting it.

Added `a fetch that lands after the amount is zeroed never resurrects the stale quote`, mirroring the existing cleared-field test but zeroing the amount mid-flight, so the new zero/unroutable branch has direct regression coverage.

On the form-open observation duplicate in the same review: the test class uses an `UnconfinedTestDispatcher`, so the quote pipeline runs eagerly during `SwapFormViewModel` construction (it starts in `init`). By the time a collector can attach to the hot `uiState` StateFlow, any construction-time transient has already settled, so post-subscription collection can't observe a pre-subscription emission regardless of `advanceUntilIdle()` placement — a queued `StandardTestDispatcher` doesn't fix this either, since construction dispatches the pipeline before the test can dispatch a collector, so the collector still loses the race and sees only the conflated latest value. Genuinely catching a form-open transient would require deferring the pipeline start out of `init`, which is out of scope here. The catchable transient — the post-subscribe `"0"`/slippage/recipient triggers — is already asserted by `empty or zero amount never raises the loading spinner across the debounce window`, and the settled form-open state is asserted false.

johnnyluo (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -145,11 +146,28 @@ constructor(
...
         get() = srcAmountState.text.toString().toBigDecimalOrNull()
 
+    /**
+     * True when the live field currently forms a quotable request: a routable pair (a real
+     * provider, so same-token is excluded) with a positive source amount. Shared by
+     * [startLoadingIfQuotable] and the late-result guard in [applyQuoteResult] so both agree on
+     * what "quotable" means, keeping zero/invalid amounts and unroutable pairs quiet on both the
+     * loading and result paths (#5296).
+     */
+    private val isLiveInputQuotable: Boolean
```
rkokhatskyi, line 165: `isLiveInputQuotable` guards `applyQuoteResult` against a stale result but never compares to `input.amount`, so a fetch resolved for an earlier amount can still land once a newer, still-positive amount is typed for the same pair, applying a quote whose memo/expected-output was priced for the wrong amount.

rkokhatskyi (changes requested):

PR #5291, opened 2026-07-14
Terra send overcharges: size the fee to actual gas, not the 300k ceiling (#5279)

aminsato:
## Summary

A Terra (LUNA, phoenix-1) native send is billed a flat **7500 uluna** — the static 300k-gas ceiling priced at 0.025 uluna/gas — even though a `MsgSend` only meters ~78k gas. Cosmos deducts the declared `fee.amount` with no refund of unused gas, so every Terra send overpays ~2.6×.

This PR sizes the Terra fee to the actual simulated gas:
- Simulates the native send (like the other Cosmos chains), prices `fee.amount` at the simulated gas limit instead of the 300k ceiling, and relays that gas limit so the broadcast declares it.
- Fixes the on-device simulate path, which previously threw and silently reverted to the static fee.

## ⚠️ Status: DRAFT — blocked on cross-platform `gas_limit` honoring

**This must not merge until the extension (Windows/SDK) and iOS *release* builds honor `CosmosSpecific.gas_limit` for Terra.** Device testing on a real 2-device vault (Android + browser extension) showed:

| Scenario | Result |
|---|---|
| **Android initiator, extension pair** | ❌ **Signing hangs forever** — Android signs/relays a dynamic `gas_limit` (~67k) but the extension co-signer builds the SignDoc with the static 300k → MPC hash mismatch → stuck (fail-closed, no funds move). |
| **Extension initiator, Android pair** | ✅ Succeeds, but `fee.amount` is still the flat 7500 (extension has no fix). |

To lower Terra's `fee.amount` below 7500 you **must** lower `gas_wanted` below 300k, and `gas_wanted` is part of the SignDoc — so **every** co-signer must build the identical lowered value or the MPC signature fails (the proto comment on `CosmosSpecific.gas_limit` states this explicitly). There is no Android-only fix.

**Safe** for: FastVault / vultiserver vaults (server signs the raw hash blindly) and Android+Android.
**Breaks** for: Android + a non-honoring extension/iOS peer.

## On-chain evidence (both extension-initiated txs)

| Tx | Extension displayed | On-chain `fee.amount` (burned) | `gas_limit` | `gas_used` |
|----|----|----|----|----|
| `79C9E911…451CB1` | 0.001944 LUNA | **7500 uluna** | 300000 | 77779 |
| `5CE8A7DC…3D3093` | 0.001944 LUNA | **7500 uluna** | 300000 | 77779 |

[...]

Issue #5279, opened 2026-07-14
Terra send network fee: Android bills static 300k-gas fee (0.0075 LUNA) vs Windows actual gas (0.001944 LUNA)


aminsato:
## Summary

On the **Transaction complete** screen after broadcasting a Terra (LUNA) send, Android shows a **Network Fee of 0.0075 LUNA**, while the Windows client shows **0.001944 LUNA** for the *same* transaction.

Android bills the **flat fee derived from the static 300,000 gas limit** rather than the gas the transaction actually uses. Windows sizes the fee from simulated/actual gas, so it charges ~4× less. Because Cosmos does not refund unused gas, the higher declared fee is fully debited — Android users overpay on every Terra send.

The value Android shows is *displayed* correctly (it matches what was debited on-chain); the bug is that the fee baked into the tx is too high in the first place.

## Cross-platform comparison (same tx)

Tx: `79C9E911A...492451CB1` — Terra (phoenix-1), 2.705074 LUNA send.

| | Android | Windows |
|---|---|---|
| Network Fee | **0.0075 LUNA** ($0.00035) | **0.001944 LUNA** ($0.00) |

| Android (Transaction complete) | Windows (Done) |
|--------|--------|
| ![android](https://i.imgur.com/htKTeml.png) | ![windows](https://i.imgur.com/WuNp90y.png) |

## Root cause

In `data/.../blockchain/cosmos/CosmosFeeService.kt`:

- `SIMULATION_SUPPORTED_CHAINS` is `{ GaiaChain, Kujira, Dydx, Osmosis, Noble, Akash }` — **Terra and TerraClassic are excluded** (documented as intentional: they route through `TerraHelper` / burn-tax rather than a native `MsgSend`, so a `CosmosHelper`-built simulation tx wouldn't match the broadcast tx).
- `simulatedGasUsed()` therefore returns `null` for Terra, and `calculateDefaultFees()` short-circuits to `staticGasFees()`.
- For Terra that is `terraFeeAmount(300_000)` = `0.025 uluna/gas × 300,000` = **7,500 uluna (0.0075 LUNA)**, a fixed amount independent of the real transaction.

Every simulated chain instead sizes the fee from `gas_used × 1.3` headroom, so their fees track actual usage — which is what the Windows client already does here (0.001944 LUNA ≈ ~78k gas × 0.025).

## Expected

[...]

PR #5299, opened 2026-07-15
Terra fee: show gas_used × price on done screen (display parity with extension) (#5279)

aminsato:
## Summary

Display-only alternative to #5291 for the Terra (LUNA) fee overcharge (#5279). After a Terra send confirms, this replaces the transaction-done **Network Fee** with `floor(gas_used × 0.025 uluna/gas)` — the same number the vultisig **extension** and **iOS** already show (e.g. **0.001944 LUNA** instead of the flat **0.0075**).

**Signing is untouched.** This is purely a display change, so it has none of the cross-platform risk of #5291.

## ⚠️ Important: this does NOT lower the fee you actually pay

Terra (Cosmos) does **not** refund unused gas — the amount actually deducted is the declared `fee.amount` (the static **7500 uluna**), regardless of `gas_used`. On-chain both recent test txs confirm it:

| Tx | On-chain `fee.amount` (burned) | `gas_limit` | `gas_used` |
|----|----|----|----|
| `79C9E911…451CB1` | 7500 uluna | 300000 | 77779 |
| `5CE8A7DC…3D3093` | 7500 uluna | 300000 | 77779 |

So this PR makes Android show `77779 × 0.025 = 0.001944` to **match the extension/iOS**, even though 0.0075 is what leaves the balance. It is **cross-platform consistency, not a real fee reduction** — the number shown is intentionally lower than what was deducted (exactly what the extension does today). The genuine fix (lowering the deducted fee) requires every co-signer to honor a relayed `gas_limit` and lives on `aminsato/5279_terra_dynamic_fee` (#5291), which is blocked until the extension/iOS release builds honor it.

## Why offer this

The real fix (#5291) is blocked and, worse, causes stuck signing when Android initiates against a non-honoring extension/iOS peer. This branch is a safe interim: it removes the visual inconsistency users see when comparing devices (extension 0.001944 vs Android 0.0075), with zero signing changes and zero regression risk.

**Decision needed (see #5291 thread):** is showing a fee lower than what's deducted acceptable for cross-platform parity? If not, close this and wait for the real fix.

## Changes

- `UpdateTerraDisplayFeeUseCase`: fetch the confirmed tx's `gas_used` (retrying briefly) and price it at Terra's min gas price (`floor`, to match the extension). Mirrors `UpdateEvmActualFeeUseCase` — except EVM genuinely refunds unused gas, so there the recomputed value really is what was burned.
- `KeysignViewModel`: inject `CosmosApiFactory`; run `tryUpdateTerraDisplayFee` alongside `tryUpdateEvmActualFee` once a terminal status is reached. Both initiator and pair devices poll status, so both refresh the shown fee.

## Verification

- Unit tests: `KeysignViewModelTryUpdateTerraDisplayFeeTest` (Terra prices at `gas_used × 0.025` = 1944 and updates the model; non-Terra chain is skipped; missing `gas_used` leaves the declared fee). Existing keysign VM tests updated for the new constructor param. All green.
- Path verified by inspection: Terra is in `TxStatusConfiguration`, and both devices reach `startForegroundPolling` (broadcast is not skipped for a normal send).

## Test plan

- [ ] Terra send (extension initiator + Android pair) → Android done screen shows `gas_used × 0.025` (~0.001944), matching the extension
- [ ] Terra send where status never confirms → done screen keeps the declared fee (0.0075), no crash
- [ ] Non-Terra Cosmos send (e.g. ATOM) → unaffected

## Summary by CodeRabbit

- **New Features**
  - Terra transaction fees now update on the completion screen using the confirmed transaction’s actual gas usage.
  - Terra native transfers now use dynamically calculated gas limits for more accurate fee handling.
  - Terra transaction simulation now uses chain-specific transaction construction.

- **Bug Fixes**
  - Prevented incorrect gas simulation for non-native Terra sends.
  - Corrected Cosmos LCD response parsing for gas and log fields, improving reliability of fee updates.

- **Tests**
  - Added coverage for Terra fee math, gas-limit safety, display-fee updates, and non-Terra/no-update scenarios.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/usecases/UpdateTerraDisplayFeeUseCase.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/usecases/UpdateTerraDisplayFeeUseCase.kt
@@ -1,0 +1,76 @@
+package com.vultisig.wallet.data.usecases
+
+import com.vultisig.wallet.data.api.CosmosApiFactory
+import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.models.EstimatedGasFee
+import com.vultisig.wallet.data.models.GasFeeParams
+import com.vultisig.wallet.data.models.TokenValue
+import java.math.BigDecimal
+import java.math.BigInteger
+import java.math.RoundingMode
+import javax.inject.Inject
+import kotlinx.coroutines.delay
+
+/**
+ * Recomputes the Terra network fee shown on the transaction-done screen from the confirmed tx's
+ * actual `gas_used`, mirroring the vultisig extension / iOS "gas used × min gas price" display.
+ *
+ * DISPLAY-ONLY. Terra (Cosmos) does NOT refund unused gas — the amount actually DEDUCTED is the
+ * declared `fee.amount` (the static 7500 uluna), not this number. This intentionally shows less
+ * than what left the balance, purely for cross-platform parity with the other clients' done
+ * screens; it does not change the signed fee. The real fix (lowering the deducted fee) needs every
+ * co-signer to honor a lowered gas_limit — see issue #5279. Mirrors [UpdateEvmActualFeeUseCase],
+ * except EVM genuinely refunds unused gas so there the recomputed value IS what was burned.
+ */
+internal class UpdateTerraDisplayFeeUseCase
+@Inject
+constructor(
+    private val cosmosApiFactory: CosmosApiFactory,
+    private val gasFeeToEstimatedFee: GasFeeToEstimatedFeeUseCase,
+) {
+
+    /**
+     * Returns the `gas_used × min-gas-price` fee estimate for [txHash], or null when not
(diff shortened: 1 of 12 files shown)


aminsato:
Found and fixed the root cause of the done screen not updating (commit 33c7ef308).

`CosmosTxStatusJson.TxResponse` annotated the gas fields camelCase — `@SerialName("gasUsed"/"gasWanted"/"rawLog")` — but the Cosmos LCD serializes them **snake_case** (`gas_used`/`gas_wanted`/`raw_log`). So they always deserialized to `null`, `UpdateTerraDisplayFeeUseCase` got `gasUsed == null` and bailed, and the fee stayed at the declared 0.0075. `CosmosStatusProvider` only reads `height`/`code`, so this went unnoticed (it also silently dropped the chain's `raw_log` from failure messages — fixed too).

Fixed the `@SerialName` values to match the wire format + added `CosmosTxStatusJsonTest` as a parse guard. Pending on-device re-test.

PR #5300, opened 2026-07-15
fix(sei): map SEI to the Ethereum coin type

aminsato:
### Transaction hash
https://seiscan.io/tx/0xfb1126f760c33caedb276a48621e63d777a45aff52e913c866a37cc22d89f479

SEI is EVM-compatible (chainId **1329**, secp256k1, `0x` addresses, derivation path `m/44'/60'/0'/0/0`), but it was mapped to WalletCore's native `CoinType.SEI`, whose raw `derivationPath()` is the **Cosmos** path `m/44'/118'/0'/0/0`. That mismatch was patched over with `compatibleType` / `compatibleDerivationPath` / `compatibleChainId` overrides — but only some call sites applied them, so the raw Cosmos path could leak through code paths that used `derivationPath()` directly.

Fix it at the root: map `Chain.Sei -> CoinType.ETHEREUM` (`Chain.kt`), exactly like `Chain.Hyperliquid` right below it, and matching **iOS and desktop**, which also resolve SEI to the Ethereum coin type. The SEI EVM chainId (1329) is applied through `compatibleChainId`'s ETHEREUM branch, mirroring Hyperliquid's `999`.

## Why it's safe (address-preserving)

Address derivation already routed SEI through `compatibleDerivationPath` (→ `m/44'/60'`) + `compatibleType` (→ `ETHEREUM`), so the derived SEI address does **not** change — this only corrects the leaky raw-path call sites and keeps the derivation path consistent across platforms.

## Changes

- `Chain.kt`: `Chain.Sei -> CoinType.ETHEREUM`
- `CoinType.kt`: move the `1329` chainId override into `compatibleChainId`'s ETHEREUM branch (alongside Hyperliquid's `999`)
- `SeiEvmCoinTypeTest`: pins `Chain.Sei -> ETHEREUM`, path `m/44'/60'/0'/0/0`, and chainId `1329`

## Test plan

- [x] compiles (`:app` / `:data` / androidTest)
- [x] SEI send signs and broadcasts on device (tx above)
- [x] iOS ↔ Android SEI co-sign works

> The separate iOS/Android ↔ extension/desktop SEI co-sign failure is a different issue on the desktop side, tracked at vultisig/vultisig-windows#4369, and is not addressed here.

### DIFF
--- a/app/src/androidTest/java/com/vultisig/wallet/data/utils/SeiEvmCoinTypeTest.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/data/utils/SeiEvmCoinTypeTest.kt
@@ -1,0 +1,43 @@
+package com.vultisig.wallet.data.utils
+
+import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.coinType
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import wallet.core.jni.CoinType
+
+/**
+ * Pins the SEI EVM coin-type contract.
+ *
+ * SEI is EVM-compatible and reuses the Ethereum coin type (secp256k1, derivation path
+ * `m/44'/60'/0'/0/0`, 0x address format), exactly like Hyperliquid — matching iOS and desktop. Its
+ * real EVM chainId (1329) is applied through [compatibleChainId] rather than WalletCore's default.
+ * This guards against SEI regressing to WalletCore's native SEI coin type (Cosmos path
+ * `m/44'/118'/…`, chainId defaulting to Ethereum's `1`), which would derive a different key and
+ * break cross-platform co-signing.
+ */
+class SeiEvmCoinTypeTest {
+
+    @Test
+    fun sei_mapsToEthereumCoinType() {
+        assertEquals(CoinType.ETHEREUM, Chain.Sei.coinType)
+    }
+
+    @Test
+    fun sei_usesEthereumDerivationPath() {
+        assertEquals("m/44'/60'/0'/0/0", Chain.Sei.coinType.derivationPath())
+    }
+
+    @Test
+    fun sei_usesSeiEvmChainId() {
+        // 1329 is applied via the override, not WalletCore's Ethereum default of "1".
+        assertEquals("1329", Chain.Sei.coinType.compatibleChainId(Chain.Sei))
+    }
+
+    @Test
+    fun ethereumAndHyperliquid_chainIdsUnaffected() {
+        // Regression guard for the shared ETHEREUM branch in compatibleChainId.
+        assertEquals("1", Chain.Ethereum.coinType.compatibleChainId(Chain.Ethereum))
+        assertEquals("999", Chain.Hyperliquid.coinType.compatibleChainId(Chain.Hyperliquid))
+    }
+}
--- a/data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
@@ -122,7 +122,12 @@ val Chain.coinType: CoinType
             Chain.Zcash -> CoinType.ZCASH
             Chain.Cardano -> CoinType.CARDANO
             Chain.Mantle -> CoinType.MANTLE
-            Chain.Sei -> CoinType.SEI
+            // SEI is EVM-compatible: it reuses the Ethereum coin type (secp256k1, derivation
+            // path m/44'/60'/0'/0/0, 0x address format) exactly like Hyperliquid below. Its EVM
+            // chainId (1329) is applied via compatibleChainId. Mapping straight to ETHEREUM keeps
+            // the derivation path/address/coin type consistent with iOS and desktop, which also
+            // map SEI to the Ethereum coin type.
+            Chain.Sei -> CoinType.ETHEREUM
             Chain.Hyperliquid -> CoinType.ETHEREUM
             Chain.Qbtc -> CoinType.COSMOS
         }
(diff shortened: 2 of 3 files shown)


aminsato:
Added instrumented regression coverage in `CompatibleDerivationPathTest` (androidTest, where the wallet-core native lib is available):

- SEI resolves to `m/44'/60'/0'/0/0`
- SEI's compatible path diverges from the raw `derivationPath()` — the reason the override exists
- non-SEI coins (ETH, BSC, BTC) fall through to the untouched wallet-core path
- the `CoinType.ETHEREUM` fallback is preserved

Kept it on the extension function rather than the private `buildJoinRequest`, since that function *is* the contract both the request and the EVM signer rely on.

app/src/androidTest/java/com/vultisig/wallet/data/utils/CompatibleDerivationPathTest.kt
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    @Test
+    fun ethereumFallback_matchesEthereumPath() {
+        // buildJoinRequest falls back to CoinType.ETHEREUM when no coin is available.
+        assertEquals(
+            CoinType.ETHEREUM.derivationPath(),
+            CoinType.ETHEREUM.compatibleDerivationPath(),
+        )
```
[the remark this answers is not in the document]
aminsato: Good catch — the test didn't exercise the null-coin path through `buildJoinRequest`, and its name overpromised. Since `nonSeiCoins_compatiblePathMatchesRawPath` already covers `CoinType.ETHEREUM` (the fallback target), I dropped the redundant test and folded the fallback rationale into that test's comment. Properly driving `buildJoinRequest(null coin)` would mean standing up the full Hilt ViewModel, which isn't worth it for asserting a one-line `?:` — the contract that matters is that ETHEREUM's compatible path equals its raw path, which stays covered.

aminsato:
@coderabbitai full review

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
```diff
@@ -369,8 +369,21 @@ class DKLSKeysign(
                         .let { Base64.decode(it) }
             }
 
+            Timber.tag("EvmHelperTrace")
```
johnnyluo: There are too many debug print out in this PR , I think it is for debug purpose , can you please remove it?

aminsato: Done — removed all the debug logging. It was temporary instrumentation I'd added to trace where SEI keysign was diverging; force-pushed the branch back to just the fix + tests (no more `EvmHelperTrace` / debug prints anywhere).

For the record, the tracing paid off: the SEI cosign failure was **not** on Android — iOS↔Android co-signs SEI fine. It reproduces only when the extension/desktop is a co-signer, and all three clients resolve SEI to the same Ethereum path `m/44'/60'/0'/0/0`, so it's not a derivation-path mismatch. Filed it against the desktop repo: vultisig/vultisig-windows#4369.

This PR now stands purely as a small correctness cleanup (the join request shouldn't advertise the raw Cosmos path for SEI — no functional consumer reads it today, so it's not a behavior fix).

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
```diff
@@ -377,7 +378,8 @@ constructor(
             messages = messagesToSign,
             sessionId = _sessionID,
             hexEncryptionKey = _encryptionKeyHex,
-            derivePath = (_keysignPayload?.coin?.coinType ?: CoinType.ETHEREUM).derivationPath(),
+            derivePath =
+                (_keysignPayload?.coin?.coinType ?: CoinType.ETHEREUM).compatibleDerivationPath(),
```
johnnyluo: I believe this is the wrong fix , instead of here update the derivationPath, you can update https://github.com/vultisig/vultisig-android/blob/main/data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt#L125 

When Sei , return Chain.Ethereum

johnnyluo: on IOS , it return sei's coinType as Ethereum

aminsato: Good call — done. Fixed it at the root: `Chain.Sei -> CoinType.ETHEREUM` (Chain.kt), exactly like `Chain.Hyperliquid` right below it, and matching iOS/desktop which also resolve SEI to the Ethereum coin type. The SEI EVM chainId (1329) now goes through `compatibleChainId`'s ETHEREUM branch alongside Hyperliquid's 999.

Reverted the `buildJoinRequest` change entirely — with the root fix, plain `derivationPath()` already yields `m/44'/60'/0'/0/0` for SEI, so that file is untouched now. Squashed the branch to a single commit; net diff is just Chain.kt + CoinType.kt + a test pinning Chain.Sei→ETHEREUM / path m/44'/60' / chainId 1329.

Address-preserving: derivation already routed SEI through `compatibleDerivationPath` (m/44'/60') + `compatibleType` (ETHEREUM), so the derived SEI address doesn't change. (`compatibleType`/`compatibleDerivationPath` still carry a now-unreachable `CoinType.SEI` branch — harmless dead code; happy to strip those too if you'd like, kept the diff focused for now.)

johnnyluo (approved):
LGTM

PR #5303, opened 2026-07-15
feat(keysign): co-sign dApp XRPL transactions (SignRipple) #5298

### LINKED ISSUE #5298: [Add] Co-signer support for XRPL dApp transactions (SignRipple keysign) [Keysign/Ripple]
# [Add] Co-signer support for XRPL dApp transactions (SignRipple keysign) [Keysign/Ripple]

## Problem

XRPL dApps (via the GemWallet API) hand the wallet a full transaction as JSON — an `OfferCreate` (DEX swap) or a cross-currency `Payment`. The browser extension now initiates these keysigns; for a **Secure Vault** the peer device (Android) must co-sign. Today Android cannot: `RippleHelper` (`chains/helpers/RippleHelpter.kt`) rebuilds the transaction from `coin.address` / `toAddress` / `toAmount` / `BlockChainSpecific.Ripple` and **never reads a `signData`/`signRipple` field**; no `signRipple` / `SignRipple` exists anywhere in the repo. So a relayed dApp XRPL keysign is either unrepresentable or gets re-derived into a *different* transaction — the co-sign produces a non-matching signature, and the reviewer approves blind.

## Context — already shipped on the other platforms

- **Schema:** `SignRipple { string raw_json = 1 }` added at `sign_ripple = 46` in `KeysignPayload.sign_data` — vultisig/commondata#99 (merged).
- **Signing:** `@vultisig/core-mpc` signs `signData.signRipple.rawJson` verbatim via WalletCore's Ripple `rawJson` path, and **fails closed** — it rejects any transaction whose `Account` ≠ the signing vault's XRP address — vultisig/vultisig-sdk#1190 (published as `@vultisig/core-mpc` 1.11.0).
- **Initiator + confirmation UI:** vultisig/vultisig-windows#4340 (merged). Verified end-to-end on-chain: [cross-currency Payment swap, `tesSUCCESS`](https://xrpscan.com/tx/BC9A70E5C49C0E45609CC12984399FAB1D42BDC3E6679F53F1A46C702452A9E3).

This issue mirrors that behavior on Android so a Secure Vault co-signer can participate.

## Solution

Carry the dApp's raw XRPL transaction to the signer verbatim and sign it deterministically, matching the extension/SDK byte-for-byte so the MPC signature is valid. Reuse the same passthrough pattern Android already has for Sui dApp PTBs (`SuiHelper` + `signSui`).

1. **Pick up the schema** — advance the `commondata` git submodule pointer to a commit including `sign_ripple` so the build-time `protobuf` codegen emits `SignRipple` (package `vultisig.keysign.v1`). Note the codegen flattens the oneof into independent nullable fields (there is no `SignDataCase`), so this surfaces as a nullable `signRipple`.
2. **Model + mappers** — add a `signRipple` field to the domain `KeysignPayload` (`data/models/payload/KeysignPayload.kt`) and wire **both** mappers: `KeysignPayloadProtoMapper.kt` (proto→domain, and set `BlockChainSpecific.Ripple` on this branch) **and `PayloadToProtoMapper.kt` (domain→proto — the outbound relay to peers).** The relay mapper currently omits some signData fields (`signTon`/`signSui`/`signBitcoin`); `signRipple` **must** be relayed or the co-signer never receives it.
3. **Sign verbatim** — teach `RippleHelper` to, when `keysignPayload.signRipple != null`, build the `Ripple.SigningInput` via `setRawJson(rawJson)` and sign as-is (analogous to `SuiHelper` reading `signSui.unsignedTxMsg`), instead of reconstructing an `OperationPayment` from `toAddress`/`toAmount`. Every device must produce identical bytes through the same `TransactionCompiler` path.
4. **Fail closed** — before signing, parse `rawJson` and reject unless its `Account` equals this vault's derived XRP `coin.address` (from `ChainAccountAddressRepository`). Same defense as `core-mpc` 1.11.0: a co-signer must never sign a transaction spending an account other than its own vault.
5. **Decode for review** — decode the transaction in the co-signer verify UI (`JoinKeysignView.kt` / `JoinKeysignViewModel.kt` / `JoinSendUiModelBuilder.kt` / `TransactionToUiModelMapper.kt`) so the peer approves readable terms — type, destination, `Amount`, `SendMax`/`DeliverMin`, `TakerGets`/`TakerPays`, issuer — with a **raw-JSON fallback** when decoding fails (a signing screen must never go blank).

### Must Do
- [ ] Advance the `commondata` submodule so codegen emits `SignRipple`
- [ ] Add a `signRipple` field to the domain `KeysignPayload`
- [ ] Wire both proto mappers, including the **outbound relay** (`PayloadToProtoMapper.kt`) so peers receive `signRipple`
- [ ] Sign `signRipple.rawJson` verbatim in `RippleHelper` so co-signed bytes match the extension/SDK
- [ ] Fail closed: reject when `rawJson.Account` ≠ this vault's XRP address
- [ ] Decode the transaction in the co-signer review UI, with a raw-JSON fallback
- [ ] Keep the existing native XRP send keysign path unchanged

### Must NOT Do
- Don't reconstruct the transaction from `toAddress`/`toAmount` on the `signRipple` path — sign the raw JSON as-is
- Don't add a dApp browser provider to Android — this is the co-signer/peer role only
- Don't broadcast anything that isn't the raw transaction the user reviewed

### Out of Scope
- An Android-side dApp connection/injection surface (the extension is the initiator)
- XRPL transaction types beyond what the extension allowlists (`Payment` / `OfferCreate` / `OfferCancel` / `TrustSet`)

## Acceptance Criteria
- [ ] A Secure Vault XRPL dApp swap initiated on the extension co-signs successfully from Android and lands on-chain (`tesSUCCESS`)
- [ ] Android produces the same signing bytes as the extension/SDK for an identical `signRipple` payload
- [ ] A payload whose `rawJson.Account` ≠ the vault's XRP address is rejected before signing
- [ ] The co-signer review screen shows decoded terms (type + amounts + destination), or the raw JSON if it can't be decoded — never a blank/misleading screen

aminsato:
## Summary

Adds Secure Vault **co-signer** support for XRPL dApp transactions (`SignRipple`).

**Closes #5298.**

XRPL dApps hand the wallet a full transaction as JSON (an `OfferCreate` DEX swap, a cross-currency `Payment`, etc.) via the extension's GemWallet provider; the browser extension initiates the keysign and the Android peer must co-sign the **exact same bytes**.

Previously `RippleHelper` always rebuilt the transaction from `toAddress`/`toAmount`/`BlockChainSpecific.Ripple` and never read a raw-JSON field, so a relayed dApp keysign produced a non-matching signature. This carries the dApp's raw XRPL transaction to the signer verbatim and signs it deterministically.

Mirrors the already-shipped pieces on the other platforms:
- Schema — `SignRipple { raw_json }` at `sign_ripple = 46` (commondata#99, merged)
- Signing — `@vultisig/core-mpc` 1.11.0 signs `signRipple.rawJson` verbatim and fails closed (vultisig-sdk#1190)
- Initiator + confirmation UI — vultisig-windows#4340 (merged)

Follows the same passthrough pattern Android already has for Sui dApp PTBs (`SuiHelper` + `signSui`).

## What changed

**Signing (data)**
- Advance the `commondata` submodule to the `SignRipple` schema
- Add `signRipple` to the domain `KeysignPayload`
- Wire both proto mappers, including the **outbound relay** (`PayloadToProtoMapper`) so peers actually receive `signRipple`
- `RippleHelper`: when `signRipple` is present, build the `Ripple.SigningInput` via `setRawJson(...)` (only rawJson + the vault pubkey) instead of reconstructing an `OperationPayment`; the native XRP send path is untouched

**Review UI**
- `RippleDappTransactionDecoder`: pure, never-throwing decode of the raw XRPL JSON into Type / Selling / Buying / Issuer / amounts (drops → XRP) / Fee, with a raw-JSON fallback so the screen is never blank
- `SignRippleDisplayView`: decoded "Transaction summary" card, matching the extension's confirmation UI
- Verify and Done screens render the operation instead of a misleading **"0 XRP"** hero, expand the summary, and drop the empty native "To" row
- Notification banner decodes the operation (e.g. `OfferCreate: 1 XRP → 2.5 USD`) instead of "Send 0 XRP"

## Review fixes

Follow-up commits addressing reviewer feedback (@rkokhatskyi) and CodeRabbit:

**Security / correctness**
- **Fail closed on a TransactionType allowlist** — only `Payment` / `OfferCreate` / `OfferCancel` / `TrustSet` are co-signable, and signing-mechanics fields (`TxnSignature`/`Signers`/`SigningPubKey`) are rejected, mirroring the Windows `sanitizeRippleDappTx` defense. A `SetRegularKey` whose `Account` is the vault's own (public) address can no longer be co-signed into an account takeover. This also neutralizes the "EscrowCreate renders as a Payment" concern — off-allowlist types can never reach signing.
- **Fee is verified, not re-estimated** — the Verify screen now surfaces the `Fee` actually baked into the signed rawJson (drops → XRP) instead of a live `RippleFeeService` estimate, so an inflated dApp fee is visible.
- **No decoy recipient** — the wire-derived native `To` row is suppressed for `signRipple`; the decoded rawJson `Destination` (the value being signed) is authoritative.
- **Honour the wire `rippleSpecific`** — the initiator always populates it, so the proto mapper no longer discards it for an empty placeholder.
- **`Account`** is still pinned to the vault's derived XRP address (unchanged).

**i18n / UX**
- Decoded field rows now carry semantic `RippleDappTxFieldKey` identifiers (with distinct issuer keys) resolved to localized strings in Compose; `ripple_field_*` added across all 10 locales.
- The dApp summary fallback is localized (`R.string.ripple_dapp_transaction`) instead of a hardcoded English literal.
- Co-signed `OfferCreate`/`TrustSet` rows persist a decoded `dappSummary` in transaction history, so they keep their meaning after a restart instead of downgrading to a blank native-XRP row.

## Resolved: co-signer Done screen stuck "pending"

Root-caused on-device: the `tx` status response was deserialized into a Payment-shaped model whose `tx_json` fields (`Destination`, `DeliverMax`, `Flags`, …) were non-nullable. A co-signed `OfferCreate` carries `TakerGets`/`TakerPays` and no `Destination`, so the strict decoder threw `MissingFieldException` on every poll; `RippleStatusProvider` swallowed it as `Pending` — leaving the Done screen "pending" forever even though the transaction settled `tesSUCCESS` on-chain.

Fixed by making every `tx_json` field nullable/defaulted (status only reads `validated` + `meta.TransactionResult`, so the shape is display-only). Verified against a real `OfferCreate` `tx` response; status now resolves to Confirmed.

## On-chain proof

Secure Vault XRPL `OfferCreate` initiated on the extension and **co-signed from Android**, landed on mainnet:

**[`0750BAFBE1788859141021DA743B5D434B8F1D90A05D6595F7C48C6DFC1407B1`](https://xrpscan.com/tx/0750BAFBE1788859141021DA743B5D434B8F1D90A05D6595F7C48C6DFC1407B1)** — `tesSUCCESS`, validated.

A successful co-sign is itself the byte-parity proof: if Android's signing input didn't match the extension's, the DKLS session would 404 and never complete.

## Co-signer review screen

| Android verify (decoded) |
|--------------------------|
| verify |

## Test plan

- [x] Unit tests: fail-closed guard (allowlist / `SetRegularKey`-on-own-account / `EscrowCreate` / tampered field / Account mismatch / missing / blank), rawJson parse, decoder (native drops→XRP, issued-currency + distinct issuers, OfferCreate, TrustSet LimitAmount, Fee, malformed→fallback), one-line `summarize`, and non-Payment `tx` status deserialization
- [x] E2E on device: extension-initiated XRPL dApp OfferCreate co-signed from Android → `tesSUCCESS` on mainnet (hash above), Done screen resolves to Confirmed

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end support for XRPL dApp SignRipple flows, including an expandable verification card with decoded details and raw transaction data.
  * Added Ripple-specific transaction banners, done-screen/details views, and consent checkbox for dApp transaction verification.
  * Added transaction-history support with a decoded one-line dApp summary and improved destination display handling.
* **Bug Fixes**
  * Fixed dApp pushes/decoding so they no longer show misleading generic send information when recipient details differ.
  * Improved XRPL response handling when returned JSON omits optional fields.
* **Tests**
  * Added/expanded unit tests for decoding, summarization, account validation, and malformed/partial JSON cases.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSendUiModelBuilder.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSendUiModelBuilder.kt
@@ -161,6 +161,7 @@ constructor(
 
         val signSolana = payload.signSolana?.rawTransactions?.firstOrNull() ?: ""
         val signSui = payload.signSui?.unsignedTxMsg?.takeIf { it.isNotEmpty() }
+        val signRipple = payload.signRipple?.rawJson?.takeIf { it.isNotBlank() }
         val transaction =
             Transaction(
                 id = UUID.randomUUID().toString(),
@@ -180,6 +181,7 @@ constructor(
                 signDirect = signDirect,
                 signSolana = signSolana,
                 signSui = signSui,
+                signRipple = signRipple,
             )
 
         val transactionToUiModel = mapTransactionToUiModel(transaction)
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSendUiModelBuilder.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSendUiModelBuilder.kt
@@ -2,6 +2,7 @@ package com.vultisig.wallet.ui.models.keysign
 
 import com.vultisig.wallet.data.blockchain.model.Transfer
 import com.vultisig.wallet.data.blockchain.model.VaultData
+import com.vultisig.wallet.data.chains.helpers.RippleDappTransactionDecoder
 import com.vultisig.wallet.data.chains.helpers.UtxoHelper
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.GasFeeParams
@@ -107,21 +108,34 @@ constructor(
             )
 
         val nativeCoin = withContext(Dispatchers.IO) { tokenRepository.getNativeToken(chain.id) }
+        // A dApp XRPL tx is signed verbatim, so the fee that is actually paid is the `Fee` baked
+        // into its raw JSON — not a live re-estimate. Surface that exact value so an inflated Fee
+        // is
+        // visible on the co-signer's Verify screen instead of being masked by a normal-looking
+        // RippleFeeService estimate.
+        val rippleDappFeeDrops =
+            payload.signRipple?.rawJson?.let { RippleDappTransactionDecoder.feeDrops(it) }
         val gasFee =
-            if (chain.standard == TokenStandard.UTXO && chain != Chain.Cardano) {
-                val utxoHelper = UtxoHelper.getHelper(vault, payloadToken.coinType)
-                val plan = utxoHelper.getBitcoinTransactionPlan(payload)
-                if (plan.error != SigningError.OK) {
-                    Timber.e("UTXO plan error: ${plan.error.name}")
+            when {
+                chain.standard == TokenStandard.UTXO && chain != Chain.Cardano -> {
+                    val utxoHelper = UtxoHelper.getHelper(vault, payloadToken.coinType)
+                    val plan = utxoHelper.getBitcoinTransactionPlan(payload)
+                    if (plan.error != SigningError.OK) {
+                        Timber.e("UTXO plan error: ${plan.error.name}")
+                    }
+                    TokenValue(value = BigInteger.valueOf(plan.fee), token = nativeCoin)
                 }
-                TokenValue(value = BigInteger.valueOf(plan.fee), token = nativeCoin)
-            } else {
-                feeResolver.resolveJoinKeysignNetworkFee(
-                    payload = payload,
-                    chain = chain,
-                    nativeCoin = nativeCoin,
-                    blockchainTransaction = blockchainTransaction,
-                )
+
+                rippleDappFeeDrops != null ->
+                    TokenValue(value = rippleDappFeeDrops, token = nativeCoin)
+
+                else ->
+                    feeResolver.resolveJoinKeysignNetworkFee(
+                        payload = payload,
+                        chain = chain,
+                        nativeCoin = nativeCoin,
+                        blockchainTransaction = blockchainTransaction,
+                    )
             }
 
         val totalGasAndFee =
--- a/app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToHistoryDataMapper.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToHistoryDataMapper.kt
@@ -1,5 +1,7 @@
 package com.vultisig.wallet.ui.models.mappers
 
+import com.vultisig.wallet.data.chains.helpers.RippleDappTransactionDecoder
+import com.vultisig.wallet.data.chains.helpers.RippleDappTxFieldKey
 import com.vultisig.wallet.data.mappers.MapperFunc
 import com.vultisig.wallet.data.models.SendTransactionHistoryData
 import com.vultisig.wallet.ui.models.TransactionDetailsUiModel
@@ -11,15 +13,26 @@ internal interface SendTransactionHistoryDataMapper :
 internal class SendTransactionHistoryDataMapperImpl @Inject constructor() :
     SendTransactionHistoryDataMapper {
 
-    override fun invoke(from: TransactionDetailsUiModel) =
-        SendTransactionHistoryData(
+    override fun invoke(from: TransactionDetailsUiModel): SendTransactionHistoryData {
+        // A dApp XRPL tx (signRipple) has native amount 0 and — for offers/trust lines — no wire
+        // recipient. Persist the decoded one-line summary and the JSON's real Destination so the
+        // history row keeps its true meaning after a restart instead of downgrading to "0 XRP → ".
+        val rippleDapp = from.signRipple
+        val dappSummary = rippleDapp?.let { RippleDappTransactionDecoder.summarize(it.rawJson) }
+        val toAddress =
+            rippleDapp?.value(RippleDappTxFieldKey.TO)?.takeIf { it.isNotBlank() }
+                ?: from.dstAddress
+
+        return SendTransactionHistoryData(
             fromAddress = from.srcAddress,
-            toAddress = from.dstAddress,
+            toAddress = toAddress,
             amount = from.token.value,
             token = from.token.token.ticker,
             tokenLogo = from.token.token.logo,
             feeEstimate = from.networkFeeTokenValue,
             memo = from.memo.orEmpty(),
             fiatValue = from.token.fiatValue,
+            dappSummary = dappSummary,
         )
+    }
 }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToHistoryDataMapper.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToHistoryDataMapper.kt
@@ -19,9 +19,17 @@ internal class SendTransactionHistoryDataMapperImpl @Inject constructor() :
         // history row keeps its true meaning after a restart instead of downgrading to "0 XRP → ".
         val rippleDapp = from.signRipple
         val dappSummary = rippleDapp?.let { RippleDappTransactionDecoder.summarize(it.rawJson) }
+        // For a dApp XRPL tx the recipient must come only from the decoded JSON `Destination`.
+        // Offers/trust lines have none, so leave it empty rather than falling back to the wire
+        // `dstAddress` — a relay could set that to a trusted label/vault while the signed JSON
+        // sends elsewhere, making the history row's recipient a decoy. Native sends keep
+        // dstAddress.
         val toAddress =
-            rippleDapp?.value(RippleDappTxFieldKey.TO)?.takeIf { it.isNotBlank() }
-                ?: from.dstAddress
+            if (rippleDapp != null) {
+                rippleDapp.value(RippleDappTxFieldKey.TO).orEmpty()
+            } else {
+                from.dstAddress
+            }
 
         return SendTransactionHistoryData(
             fromAddress = from.srcAddress,
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
@@ -288,18 +288,26 @@ internal fun VerifySendScreen(
                         bracketValue = tx.srcVaultName?.let { tx.srcAddress },
                     )
 
-                    VerifyCardDivider(0.dp)
+                    // For a dApp XRPL tx the native `To` comes from the wire `toAddress`, which is
+                    // not the value being signed — the real recipient (if any) lives in the raw
+                    // JSON `Destination` and is rendered by SignRippleDisplayView below. Showing
+                    // the
+                    // wire address here would let a relay display a trusted label/vault while the
+                    // signed JSON sends elsewhere, so suppress the native To row for signRipple.
+                    if (tx.signRipple == null) {
+                        VerifyCardDivider(0.dp)
 
-                    val toDstLabel =
-                        tx.dstVaultName
-                            ?: tx.dstAddressBookTitle
-                            ?: tx.dstContractLabel
-                            ?: tx.dstLabel
-                    VerifyCardDetails(
-                        title = stringResource(R.string.verify_transaction_to_title),
-                        subtitle = toDstLabel ?: tx.dstAddress,
-                        bracketValue = toDstLabel?.let { tx.dstAddress },
-                    )
+                        val toDstLabel =
+                            tx.dstVaultName
+                                ?: tx.dstAddressBookTitle
+                                ?: tx.dstContractLabel
+                                ?: tx.dstLabel
+                        VerifyCardDetails(
+                            title = stringResource(R.string.verify_transaction_to_title),
+                            subtitle = toDstLabel ?: tx.dstAddress,
+                            bracketValue = toDstLabel?.let { tx.dstAddress },
+                        )
+                    }
 
                     // A memo that merely echoes the destination tag is signed as tag-only, so it is
                     // not shown as a separate Memo row (it appears in the Destination Tag row
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
@@ -107,6 +107,7 @@ internal fun VerifySendScreen(viewModel: VerifyTransactionViewModel = hiltViewMo
         confirmTitle = stringResource(R.string.keysign_sign_transaction),
         onConsentAddress = viewModel::checkConsentAddress,
         onConsentAmount = viewModel::checkConsentAmount,
+        onConsentDappTransaction = viewModel::checkConsentDappTransaction,
         onConfirm = viewModel::joinKeySign,
         onBackClick = viewModel::back,
         onFastSignClick = viewModel::fastSign,
@@ -126,6 +127,7 @@ internal fun VerifySendScreen(
     dappMetadata: DAppMetadata? = null,
     onConsentAddress: (Boolean) -> Unit = {},
     onConsentAmount: (Boolean) -> Unit = {},
+    onConsentDappTransaction: (Boolean) -> Unit = {},
     onBackClick: () -> Unit = {},
     onConfirmScanning: () -> Unit = {},
     onDismissScanning: () -> Unit = {},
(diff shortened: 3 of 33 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleDappTransaction.kt
```diff
@@ -0,0 +1,131 @@
...
+        val obj =
+            try {
+                json.parseToJsonElement(rawJson).jsonObject
+            } catch (e: Exception) {
+                Timber.w("Failed to decode SignRipple rawJson for display: %s", e.message)
+                return RippleDappTx(transactionType = null, fields = emptyList(), rawJson = rawJson)
+            }
+
+        val transactionType = obj.stringOrNull("TransactionType")
+        val fields = buildList {
```
rkokhatskyi, line 81: `decode()` never extracts `Fee`, and `dappSuppliedNativeFee` only covers `cosmosNativeDenom` chains, so Ripple falls to `RippleFeeService`'s live network estimate in `EstimatedNetworkFee`, unrelated to the actual `Fee` in `rawJson` that gets signed verbatim. A malicious inflated Fee shows as a normal small fee.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt
```diff
@@ -157,6 +177,61 @@ object RippleHelper {
...
+            .build()
+            .toByteArray()
+    }
+
+    /**
+     * Fail-closed guard for a dApp-supplied [SignRipple] transaction: throws unless the JSON's
+     * `Account` equals this vault's derived XRP [expectedAccount]. Pure (no JNI), so it is
+     * unit-testable independently of WalletCore.
+     */
+    internal fun verifyDappTransactionAccount(rawJson: String, expectedAccount: String) {
```
rkokhatskyi: `verifyDappTransactionAccount` only checks `Account`, never `TransactionType`, unlike Windows' `sanitizeRippleDappTx` allowlist. A `SetRegularKey` tx with `Account` equal to the vault's own address (public info) passes this guard and gets co-signed, installing an attacker's key with no warning since `decode()` never reads `RegularKey`.

app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
```diff
@@ -334,6 +364,14 @@ internal fun VerifySendScreen(
                             )
                         }
 
+                    tx.signRipple?.let {
```
rkokhatskyi, line 377: The native `To` row (`VerifySendScreen.kt:293`, `toDstLabel ?: tx.dstAddress`) still renders unconditionally from wire `toAddress`, never checked against `rawJson`'s `Destination`. A relay can set `toAddress` to a trusted vault or contact label while `rawJson.Destination` actually sends elsewhere, making the headline recipient a decoy.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleDappTransaction.kt
```diff
@@ -0,0 +1,131 @@
...
+            transactionType?.let { add(RippleDappTxField("Type", it)) }
+            obj.stringOrNull("Account")?.let { add(RippleDappTxField("From", it)) }
+            obj.stringOrNull("Destination")?.let { add(RippleDappTxField("To", it)) }
```
rkokhatskyi: `decode()` reuses the generic `Destination`/`Amount` rules for any type that happens to share those field names, so `EscrowCreate` renders identically to a plain Payment (From/To/Amount) while `Condition`/`FinishAfter`, the fields that actually govern release, are silently dropped, creating false confidence rather than an honest gap.

data/src/main/kotlin/com/vultisig/wallet/data/mappers/KeysignPayloadProtoMapper.kt
```diff
@@ -147,6 +148,22 @@ internal class KeysignPayloadProtoMapperImpl @Inject constructor() : KeysignPayl
...
 
+                    // A dApp-supplied XRPL transaction (`signRipple`) is signed verbatim from its
+                    // raw JSON via WalletCore's `rawJson` path — Fee / Sequence /
+                    // LastLedgerSequence are already encoded in the JSON, so the initiator omits
+                    // `rippleSpecific`. Stand in an empty placeholder so [RippleHelper], which
+                    // casts
+                    // `blockChainSpecific` to `Ripple`, has a value to read; its fields go unused
+                    // on
+                    // the signRipple path.
+                    from.signRipple != null ->
```
rkokhatskyi: This comment's premise is wrong: Windows always populates `rippleSpecific` for signRipple too (`build.ts` runs `getChainSpecific` for every chain before setting `signData`), so this branch discards real live-fetched sequence/fee data the wire already carries, unlike the analogous Sui branch above where `SuiSpecificSchema` really is empty.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSendUiModelBuilder.kt
```diff
@@ -180,6 +181,7 @@ constructor(
                 signDirect = signDirect,
                 signSolana = signSolana,
                 signSui = signSui,
+                signRipple = signRipple,
```
rkokhatskyi, line 198: Once signed, this `signRipple` data has nowhere to go: `SendTransactionHistoryDataMapperImpl` (`TransactionToHistoryDataMapper.kt:14`) only carries fromAddress/toAddress/amount/memo, so a co-signed OfferCreate/TrustSet/cross-currency Payment permanently downgrades to a blank or generic native-XRP row in transaction history after the app restarts.

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleDappTransaction.kt
aminsato: Good catch — fixed in 28d9067a7. The decoder now extracts `Fee` and exposes `feeDrops(rawJson)`, and `JoinSendUiModelBuilder` drives the Network Fee row from the signed rawJson `Fee` (drops→XRP) instead of the `RippleFeeService` live estimate. An inflated Fee is now visible on the Verify screen. Test: `decodes the signed Fee as XRP`.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleHelpter.kt
aminsato: Fixed in 28d9067a7. `verifyDappTransactionAccount` → `verifyDappTransaction` now enforces a `TransactionType` allowlist (`Payment`/`OfferCreate`/`OfferCancel`/`TrustSet`, mirroring Windows `sanitizeRippleDappTx`) and rejects signing-mechanics fields (`TxnSignature`/`Signers`/`SigningPubKey`). A `SetRegularKey` whose `Account` equals the vault own address is now refused. Test: `verifyDappTransaction rejects a SetRegularKey even for the vault own account`.

app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
aminsato: Fixed in 28d9067a7 — the native `To` row is now suppressed when `tx.signRipple != null`. The recipient the co-signer sees is the decoded rawJson `Destination` inside `SignRippleDisplayView` (the value actually being signed), so a relay can no longer show a trusted label/vault while the JSON sends elsewhere.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleDappTransaction.kt
aminsato: Addressed by the allowlist in 28d9067a7 — `EscrowCreate` (and any other type that happens to share `Destination`/`Amount` field names) is now rejected at signing time, so it can never reach the co-signer rendered as a plain Payment. For the allowlisted types the decoded card also always keeps the raw JSON visible as the source of truth.

data/src/main/kotlin/com/vultisig/wallet/data/mappers/KeysignPayloadProtoMapper.kt
aminsato: Agreed — fixed in 28d9067a7. The `signRipple` placeholder branch is now gated on `from.rippleSpecific == null`; when the wire carries `rippleSpecific` (which the Windows initiator always populates via `getChainSpecific`) it is honored through the existing `rippleSpecific` branch instead of being zeroed out.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSendUiModelBuilder.kt
aminsato: Fixed in 28d9067a7 — added a nullable `dappSummary` to `SendTransactionHistoryData` (default null, serialized into the JSON payload column, no Room migration), populated in the history mapper from the decoded summary + the rawJson `Destination`. `SendTransactionCard` renders it in place of the `0 XRP`/blank row, so a co-signed OfferCreate/TrustSet keeps its meaning after an app restart.

app/src/main/java/com/vultisig/wallet/ui/models/mappers/TransactionToHistoryDataMapper.kt
```diff
@@ -11,15 +13,26 @@ internal interface SendTransactionHistoryDataMapper :
...
-        SendTransactionHistoryData(
+    override fun invoke(from: TransactionDetailsUiModel): SendTransactionHistoryData {
+        // A dApp XRPL tx (signRipple) has native amount 0 and — for offers/trust lines — no wire
+        // recipient. Persist the decoded one-line summary and the JSON's real Destination so the
+        // history row keeps its true meaning after a restart instead of downgrading to "0 XRP → ".
+        val rippleDapp = from.signRipple
+        val dappSummary = rippleDapp?.let { RippleDappTransactionDecoder.summarize(it.rawJson) }
+        val toAddress =
+            rippleDapp?.value(RippleDappTxFieldKey.TO)?.takeIf { it.isNotBlank() }
+                ?: from.dstAddress
```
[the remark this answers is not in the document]
aminsato: Fixed in e23b5022b. For a dApp XRPL tx the history `toAddress` now comes only from the decoded JSON `Destination` (`.orEmpty()`), so OfferCreate/TrustSet leave it blank instead of falling back to the wire `dstAddress`. The `from.dstAddress` fallback is retained exclusively for ordinary sends, so a relay can no longer surface a trusted-looking decoy recipient in history.

app/src/main/java/com/vultisig/wallet/ui/screens/send/VerifySendScreen.kt
```diff
@@ -236,16 +237,45 @@ internal fun VerifySendScreen(
...
+                                header =
+                                    stringResource(
+                                        tx.headerTitleRes ?: R.string.verify_deposit_sending
+                                    ),
+                                valuedToken = tx.token,
+                                shape = RoundedCornerShape(0.dp),
+                                modifier = Modifier.fillMaxWidth(),
+                                withContainer = false,
+                            )
+                        }
```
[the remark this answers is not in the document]
aminsato: Fixed in e23b5022b. For `signRipple` the Verify screen no longer shows the address/amount consent pair (both meaningless for an OfferCreate — no recipient, 0 native amount) and now shows a single `verify_transaction_consent_dapp_transaction` ("I have reviewed the transaction details") consent instead. `hasAllConsents` gates on that single consent for dApp txs while native sends keep the two-checkbox flow. The native To row and 0-XRP hero were already suppressed for signRipple; the decoded terms remain the source of truth in `SignRippleDisplayView`.

johnnyluo (approved):
LGTM

PR #5306, opened 2026-07-16
fix(thorchain): Unbond confirmation label + formatted From/To

### LINKED ISSUE #5301: THORChain Unbond confirmation: wrong "You're sending" label and raw From/To addresses
When performing a THORChain **Unbond**, the confirmation screen ("Function overview") mislabels the action and shows the vault address unformatted.

unbond-confirmation

## Problems

1. **Wrong action label** — the summary reads **"You're sending"**. This is an unbond, not a send, so the label is misleading. It should read **"Unbonding"** (matching the operation being performed).

2. **Unformatted From / To addresses** — both **From** and **To** display the current vault address as a raw string (`thor1...`). They should be formatted as **`VaultName (address)`**, consistent with how the vault address is presented elsewhere.

## Steps to reproduce

1. Go to a THORChain node bond position and choose **Unbond**.
2. Enter an amount and tap **Continue**.
3. Observe the **Function overview** confirmation screen.

## Expected

- Action label reads **"Unbonding"** instead of **"You're sending"**.
- **From** and **To** show **`VaultName (address)`** instead of the raw address.

## Actual

- Label reads **"You're sending"**.
- **From** and **To** show the raw vault address.

aminsato:
## Summary

The THORChain **Unbond** "Function overview" confirmation mislabeled the action as **"You're sending"** and rendered the **From**/**To** vault address as a raw `thor1…` string.

The node-management Unbond flow builds its `DepositTransaction` on the send side, which leaves `operation` blank — so the header can't key off `operation` alone. This resolves the title from the operation **or** the `UNBOND` memo prefix, and formats From/To as **`VaultName (address)`** the same way the Send verify screen does (the destination of a self-operated node is the vault itself, so it resolves to the vault name).

Closes #5301

## Changes

- **`DepositTransactionToUiModelMapper`** — add `depositVerifyTitleRes(operation, memo)` → `verify_deposit_unbonding` for Unbond, else the generic `verify_deposit_sending`; expose it as a new `titleRes` on the UI model.
- **`VerifyDepositViewModel`** — resolve `srcVaultName` / `dstVaultName` / `dstAddressBookTitle` via the shared `resolveDstVaultName` util (injected `VaultRepository`, `AddressBookRepository`, `ChainAccountAddressRepository`, `@IoDispatcher`).
- **`VerifyDepositScreen`** — render `tx.titleRes` and format From/To as `VaultName (address)` using `VerifyCardDetails`' `bracketValue`.
- New string `verify_deposit_unbonding` across all 10 locales.
- Unit tests for the title resolver; `PreviewActivity` before/after previews.

Non-Unbond deposits keep "You're sending" and simply gain the cleaner `VaultName (address)` From/To formatting.

## Before / After

| Before | After |
|--------|-------|
| before | after |

## Test plan

- [x] `./gradlew :app:testDebugUnitTest` — new title-resolver tests + updated `VerifyDepositViewModelTest` green.
- [x] Device render (before/after above) via `PreviewActivity` on emulator.
- [ ] Manual mainnet smoke: unbond from a THORChain node position and confirm the "Function overview" reads "Unbonding" with `VaultName (address)` From/To.

## Summary by CodeRabbit

- **New Features**
  - Added THORChain unbond verification preview support for new “before” and “after” screens.
  - Deposit verification UI now uses a dynamic header title (Unbonding vs Sending) and improved source/destination labeling via vault and address-book names.
- **Bug Fixes**
  - Refined “from/to” subtitles and bracket values for clearer context.
  - The node address row now shows only when it’s relevant (non-empty and different from the destination address).
- **Localization**
  - Added `verify_deposit_unbonding` string across supported languages.
- **Tests**
  - Expanded verification view model and mapper tests to cover unbonding title selection and label resolution.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
@@ -117,7 +117,7 @@ constructor(
                         nodeAddress = transaction.nodeAddress,
                         pairedAddress = transaction.pairedAddress,
                         pool = transaction.pool,
-                        titleRes = depositVerifyTitleRes(transaction.operation, transaction.memo),
+                        titleRes = depositVerifyTitleRes(transaction.operation),
                     )
 
                 state.update {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
@@ -127,7 +127,7 @@ constructor(
                 val depositTransactionUiModel =
                     mapTransactionToUiModel(transaction)
                         .withVaultLabels(
-                            vaultId = vaultId!!,
+                            vaultId = transaction.vaultId,
                             chain = transaction.srcToken.chain,
                             dstAddress = transaction.dstAddress,
                         )
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
@@ -1,24 +1,32 @@
 package com.vultisig.wallet.ui.models.deposit
 
+import androidx.annotation.StringRes
 import androidx.lifecycle.SavedStateHandle
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import androidx.navigation.toRoute
 import com.vultisig.wallet.R
+import com.vultisig.wallet.data.IoDispatcher
 import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.repositories.AddressBookRepository
 import com.vultisig.wallet.data.repositories.BalanceRepository
+import com.vultisig.wallet.data.repositories.ChainAccountAddressRepository
 import com.vultisig.wallet.data.repositories.DepositTransactionRepository
 import com.vultisig.wallet.data.repositories.VaultPasswordRepository
+import com.vultisig.wallet.data.repositories.VaultRepository
 import com.vultisig.wallet.data.usecases.IsVaultHasFastSignByIdUseCase
 import com.vultisig.wallet.ui.models.keysign.KeysignInitType
 import com.vultisig.wallet.ui.models.mappers.DepositTransactionToUiModelMapper
+import com.vultisig.wallet.ui.models.mappers.depositVerifyTitleRes
 import com.vultisig.wallet.ui.models.swap.ValuedToken
 import com.vultisig.wallet.ui.navigation.Route
 import com.vultisig.wallet.ui.navigation.SendDst
 import com.vultisig.wallet.ui.navigation.util.LaunchKeysignUseCase
 import com.vultisig.wallet.ui.utils.UiText
+import com.vultisig.wallet.ui.utils.resolveDstVaultName
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.first
@@ -45,6 +53,7 @@ internal data class DepositTransactionUiModel(
     val pairedAddress: String = "",
     val pool: String = "",
     val validatorName: String = "",
+    @StringRes val titleRes: Int = R.string.verify_deposit_sending,
 )
 
 internal data class VerifyDepositUiModel(
@@ -70,6 +79,10 @@ constructor(
     private val depositTransactionRepository: DepositTransactionRepository,
     private val balanceRepository: BalanceRepository,
     private val vaultPasswordRepository: VaultPasswordRepository,
+    private val vaultRepository: VaultRepository,
+    private val addressBookRepository: AddressBookRepository,
+    private val chainAccountAddressRepository: ChainAccountAddressRepository,
+    @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
     private val launchKeysign: LaunchKeysignUseCase,
     private val isVaultHasFastSignById: IsVaultHasFastSignByIdUseCase,
 ) : ViewModel() {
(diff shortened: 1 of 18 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
```diff
@@ -104,13 +117,20 @@ constructor(
...
 
                 state.update {
                     it.copy(isLoading = true, depositTransactionUiModel = initialTransaction)
                 }
 
-                val depositTransactionUiModel = mapTransactionToUiModel(transaction)
+                val depositTransactionUiModel =
+                    mapTransactionToUiModel(transaction)
+                        .withVaultLabels(
+                            vaultId = vaultId!!,
```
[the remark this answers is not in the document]
aminsato: Addressed in 123974f32 — resolve the From label from the loaded transaction's non-null `vaultId` (`DepositTransaction.vaultId: String`) instead of force-unwrapping the nullable route arg.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
```diff
@@ -104,13 +117,20 @@ constructor(
...
                         pool = transaction.pool,
+                        titleRes = depositVerifyTitleRes(transaction.operation, transaction.memo),
                     )
 
                 state.update {
                     it.copy(isLoading = true, depositTransactionUiModel = initialTransaction)
                 }
 
-                val depositTransactionUiModel = mapTransactionToUiModel(transaction)
+                val depositTransactionUiModel =
```
rkokhatskyi, line 127: Confirmed cross-path gap: `withVaultLabels` in `VerifyDepositViewModel.kt` (initiator, ~line 127) resolves srcVaultName/dstVaultName/dstAddressBookTitle, but `JoinDepositUiModelBuilder.build()` (joiner, ~line 123) calls `mapDepositTransactionToUiModel` directly and returns, never populating those fields, so `VerifyDepositScreen`'s null-fallback renders raw thor1... addresses on the joining device for the same Unbond the initiator sees labeled. `JoinDepositUiModelBuilder` already has `vault`, `chain`, and `dstAddress` in scope to mirror this, same as `JoinSendUiModelBuilder.kt:202-213` does for Send per #5278. Title fix (Unbonding vs sending) does reach the joiner fine since it lives in the shared mapper. Worth a fast follow-up before this ships as complete parity with Send

app/src/main/java/com/vultisig/wallet/ui/models/mappers/DepositTransactionToUiModelMapper.kt
```diff
@@ -12,6 +15,19 @@ import kotlinx.coroutines.flow.first
...
     SuspendMapperFunc<DepositTransaction, DepositTransactionUiModel>
 
+/**
+ * Resolves the "Function overview" header for a deposit: an Unbond reads "Unbonding" rather than
+ * the generic "You're sending", since it isn't a send (issue #5301). The send-side node-management
+ * flow leaves [operation] blank, so the Unbond memo prefix is honored as a fallback signal.
+ */
+@StringRes
+internal fun depositVerifyTitleRes(operation: String, memo: String): Int =
+    if (operation == OPERATION_UNBOND || memo.trimStart().startsWith("UNBOND", ignoreCase = true)) {
```
rkokhatskyi: depositVerifyTitleRes matches any memo starting with UNBOND case-insensitively, with no operation check. `CustomStrategy.kt` builds free-text memos straight from user input, so a Custom deposit whose memo happens to start with 'unbond' gets mislabeled 'Unbonding' on the Verify screen even though it's an unrelated generic deposit, not a node unbond

app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt
```diff
@@ -187,9 +188,11 @@ internal fun VerifyDepositScreen(
                         VerifyCardDivider(0.dp)
                     }
                     if (tx.dstAddress.isNotEmpty()) {
+                        val toDstLabel = tx.dstVaultName ?: tx.dstAddressBookTitle
                         VerifyCardDetails(
                             title = stringResource(R.string.verify_transaction_to_title),
-                            subtitle = tx.dstAddress,
+                            subtitle = toDstLabel ?: tx.dstAddress,
+                            bracketValue = toDstLabel?.let { tx.dstAddress },
```
rkokhatskyi, line 195: UnbondStrategy.kt (deposit-form) sets both dstAddress and nodeAddress to the same node address, so the newly-formatted Unbond Verify screen shows the address twice: nicely as 'To: VaultName (address)' and again raw under the 'Node address' row in `VerifyDepositScreen.kt`. The PR's own preview never sets nodeAddress, so this gap doesn't show in its before/after screenshots

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
aminsato: Fixed in 744b25bbc. `JoinDepositUiModelBuilder.build()` now resolves `srcVaultName` (the joining vault) plus `dstVaultName`/`dstAddressBookTitle` via `resolveDstVaultName` + `AddressBookRepository`, exactly mirroring `JoinSendUiModelBuilder`, and copies them onto the deposit UI model. The joining device now renders `VaultName (address)` for the same Unbond the initiator labels — full parity with Send, in this PR rather than a follow-up.

app/src/main/java/com/vultisig/wallet/ui/models/mappers/DepositTransactionToUiModelMapper.kt
aminsato: Fixed in 744b25bbc. Dropped the memo-prefix heuristic entirely — `depositVerifyTitleRes` now keys off the structured `operation` alone (`operation == OPERATION_UNBOND`), so a Custom deposit whose free-text memo happens to start with "unbond" is no longer mislabeled. The one path that previously relied on the memo fallback (send-flow `UnbondStrategy`) now sets `operation = OPERATION_UNBOND`; the deposit-form strategy and the joiner (via `ThorchainMemoParser`) already set it, so every real Unbond still reads "Unbonding".

app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt
aminsato: Fixed in 744b25bbc. The Verify screen now guards the Node address row with `tx.nodeAddress != tx.dstAddress`, so when Unbond sets both to the same node address it renders once as the "To" row and the redundant "Node address" row is suppressed. Done at the display layer so it also covers the joiner, where the parser sets `nodeAddress` and `dstAddress` to the same value.

johnnyluo (approved):
LGTM

PR #5311, opened 2026-07-17
fix(swap): drop late quote resolved for a different amount on same pair

### LINKED ISSUE #5310: Swap: stale quote for an earlier amount can be applied and signed over a newer, different amount (same pair)
## Summary

`SwapQuotePipelineController.isLiveInputQuotable` (added in #5297 to stop a stale, no-longer-valid
quote from being resurrected after the amount field is cleared/zeroed or the pair becomes
unroutable) only checks **whether something is currently quotable** — it never checks whether the
landing fetch was actually resolved **for the amount currently in the field**. A quote fetch queued
for an earlier amount can still be applied after the user has typed a different, still-positive
amount for the same pair, showing (and making signable) a quote priced for the wrong amount.

## Root cause

`app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt`

```kotlin
private val isLiveInputQuotable: Boolean
    get() {
        val amount = srcAmount
        return isPairRoutable && amount != null && amount > BigDecimal.ZERO
    }
```

`applyQuoteResult` gates on this before applying a resolved quote:

```kotlin
private suspend fun applyQuoteResult(
    input: QuoteInput,
    result: SwapQuotePipelineResult.Success,
) {
    if (!isLiveInputQuotable) {
        resetQuoteState()
        return
    }
    ...
```

The guard re-reads the *live* amount/pair but never compares it against `input.amount` (the amount
the fetch was actually resolved for). So it's blind to the specific case where the live amount is
different from, but still as valid as, the amount the in-flight fetch was started for.

**Why the race is reachable:** `startLoadingIfQuotable()`/`showIndicativeRate()` run in an `.onEach`
upstream of `.debounce { ... }`, so every keystroke is processed immediately — only the actual
`resolveQuote()` fetch inside `.collectLatest { ... }` waits out the 300ms debounce.
`collectLatest` only cancels an in-flight fetch when the debounced flow delivers a **new** value to
it, i.e. only after the *next* input's own 300ms quiet period elapses. If a user:

1. types an amount (e.g. `5`),
2. pauses long enough (>300ms) for the debounce to fire and a real quote fetch to start,
3. then resumes typing before that fetch resolves (e.g. changes the field to `56`, same pair, still
   positive),

the `5`-priced fetch can complete and land inside the window before `56`'s own debounce elapses.
`isLiveInputQuotable` sees `isPairRoutable == true` and `amount ("56") > 0`, so the guard passes and
`applyQuoteResult` applies the stale `5`-priced quote — `hasQuote = true`, `isSwapDisabled = false`,
the displayed destination amount/fee, and `quoteState.quote` all now reflect `5`, while the field
shows `56`.

This is an everyday interaction pattern (type a rough amount, glance at the indicative estimate,
then refine it), not a contrived edge case, and swap-quote fetches (parallel provider calls to
THORChain/MayaChain/1inch/etc.) commonly take longer than the 300ms debounce window.

## Impact

Nothing downstream cross-checks the quote's basis amount against the amount actually being staged
for signing:

- `SwapFormViewModel.swap()` passes `srcAmount = srcAmountState.text.toString()` (the live text)
  alongside `quote = quoteState.quote` (possibly stale) into `SwapInputCollector.collect(...)`.
- `SwapInputCollector.collect` only checks the quote is non-null and `expectedDstValue != 0` — it
  never compares the quote's own basis amount to the live `srcAmountInt`.
- `SwapTransactionBuilder.build` then uses `srcTokenValue` (built fresh from the live amount) as the
  actual on-chain transfer amount, while `quote.data.memo` / `quote.expectedDstValue` (THORChain
  `:LIM:` floor, EVM aggregator calldata, etc.) still reflect the stale, smaller/larger amount.
- `VerifySwapViewModel` does not re-quote or re-validate before keysign.

Net effect: a transaction can be staged (and, if tapped quickly, signed) where the amount actually
transferred doesn't match the amount the quote/memo/slippage-floor was computed for — e.g. a
THORChain swap's `:LIM:` minimum-output floor calibrated for a much smaller amount than what is
actually deposited, silently defeating slippage protection for that swap.

## Scope note

This predates #5297 — `applyQuoteResult` had **zero** staleness guard of any kind before that PR
(any resolved quote was applied unconditionally, regardless of amount/pair). #5297 narrowed the
exposure (closing the empty/zeroed/unroutable-pair sub-cases) but wasn't scoped to close this
same-pair/different-amount sub-case, since it wasn't part of the original blink bug (#5296) being
fixed. Flagged as its own ticket rather than reopening #5297.

## Suggested fix

Compare the live input against what the fetch actually resolved for, not just "is something
quotable" — e.g. thread a monotonic request/generation id through `QuoteInput`/`resolveQuote`, or
compare `input.amount == srcAmount` (and the resolved pair) inside the `applyQuoteResult` guard, and
drop the result if it no longer matches.

## References

- Introduced/narrowed by #5297 (see review at
  https://github.com/vultisig/vultisig-android/pull/5297#pullrequestreview-4713535532)
- Related safety-net context: #4858, #4969

johnnyluo:
## Summary
- `applyQuoteResult`'s late-result guard only checked `isLiveInputQuotable` ("is *something* quotable now"), never whether the resolved fetch matched the amount currently in the field.
- A quote fetch queued for an earlier amount could resolve **after** the user typed a different, still-positive amount for the **same pair** (before that amount's own 300ms debounce fired), and be applied — since `collectLatest` can't cancel the in-flight fetch until the next debounced value arrives.
- Net effect: a swap could be staged (and, if tapped quickly, signed) where the on-chain transfer amount didn't match the amount the quote/memo/slippage-floor was computed for — e.g. a THORChain `:LIM:` minimum-output floor calibrated for a much smaller amount, **silently defeating slippage protection**.

## Fix
- Require the live amount to still match `input.amount` (scale-safe `compareTo`) before applying a resolved quote; drop it via `resetQuoteState()` otherwise. One extra clause on the existing choke-point guard every quote flows through.

## Issue
Closes #5310

## Test Plan
- [x] Added regression test `a fetch that lands for an earlier amount never applies over a newer amount on the same pair` — gates the amount-`1` fetch so it lands while `56` is still in the debounce, asserts the `1`-priced quote is never applied. Passes with the fix; without it the guard would flip `hasQuote` true.
- [x] Full `SwapFormViewModelTest` suite passes (`./gradlew :app:testDebugUnitTest`)
- [x] Manual: type an amount, pause >300ms, then extend it before the quote lands — verify the displayed/staged quote always matches the field

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented late or superseded swap quote results from overwriting newer UI state after amount, token pair, slippage, or external recipient changes.
  * Strengthened stale-quote detection (including endpoint changes and amount/null mismatches) and discards only outdated quote data while preserving the newest loading/indicative destination estimate.
  * Improved behavior when a resolved quote is no longer valid by clearing quote-derived details appropriately.

* **Tests**
  * Added regression coverage for delayed results after source amount changes, token pair flips, slippage updates, and external recipient updates.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
@@ -319,7 +319,13 @@ constructor(
                     updatePairSupport(src, dst)
                 }
                 .combine(amountChanges) { address, immediate ->
-                    QuoteInput(address = address, amount = srcAmount, immediate = immediate)
+                    QuoteInput(
+                        address = address,
+                        amount = srcAmount,
+                        slippageBps = slippageBps.value,
+                        externalRecipient = externalRecipient.value,
+                        immediate = immediate,
+                    )
                 }
                 // Fires on real user intent (typing, paste, percentage, token change) but not on
                 // the
@@ -342,9 +348,14 @@ constructor(
                 // the
                 // skeletons (#5296). The onEach rides each flow individually, so the silent refresh
                 // timer above doesn't flash the spinner.
-                .combine(slippageBps.onEach { startLoadingIfQuotable() }) { input, _ -> input }
-                .combine(externalRecipient.onEach { startLoadingIfQuotable() }) { input, _ ->
-                    input
+                .combine(slippageBps.onEach { startLoadingIfQuotable() }) { input, liveSlippageBps
+                    ->
+                    input.copy(slippageBps = liveSlippageBps)
+                }
+                .combine(externalRecipient.onEach { startLoadingIfQuotable() }) {
+                    input,
+                    liveExternalRecipient ->
+                    input.copy(externalRecipient = liveExternalRecipient)
                 }
                 // Percentage / Max / paste skip the debounce (0ms); free typing still coalesces at
                 // 300ms so rapid edits fire a single quote fetch.
@@ -371,8 +382,8 @@ constructor(
                                 referralCode = referralCode.value,
                                 currentDiscountInfo = uiState.value.discountInfo,
                                 selectedSrcTokenTitle = uiState.value.selectedSrcToken?.title,
-                                slippageBps = slippageBps.value,
-                                externalRecipient = externalRecipient.value,
+                                slippageBps = input.slippageBps,
+                                externalRecipient = input.externalRecipient,
                             )
                     ) {
                         SwapQuotePipelineResult.Empty -> resetQuoteState()
@@ -426,7 +437,9 @@ constructor(
                 liveAmount == null ||
                 liveAmount.compareTo(input.amount) != 0 ||
                 !inputSrc.hasSameQuoteEndpointAs(selectedSrc.value) ||
-                !inputDst.hasSameQuoteEndpointAs(selectedDst.value)
+                !inputDst.hasSameQuoteEndpointAs(selectedDst.value) ||
+                input.slippageBps != slippageBps.value ||
+                input.externalRecipient != externalRecipient.value
         if (isSuperseded) {
             discardSupersededQuoteResult()
             return
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
@@ -406,7 +406,20 @@ constructor(
         // a
         // late-landing fetch for a now non-quotable field (cleared, zeroed, or an unroutable pair)
         // drops the stale quote instead of resurrecting it (#5296 review).
-        if (!isLiveInputQuotable) {
+        //
+        // isLiveInputQuotable only asks "is something quotable now"; it can't catch a fetch that
+        // resolved for an earlier, still-positive amount on the same pair (type 5, pause past the
+        // debounce, resume to 56 before 5's fetch lands). Also require the live amount to still
+        // match the amount this fetch was resolved for, so a quote priced for the old amount can't
+        // be applied — and then signed with a mismatched memo / slippage floor — over the new one
+        // (#5310).
+        val liveAmount = srcAmount
+        if (
+            !isLiveInputQuotable ||
+                input.amount == null ||
+                liveAmount == null ||
+                liveAmount.compareTo(input.amount) != 0
+        ) {
             resetQuoteState()
             return
         }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
@@ -412,15 +412,23 @@ constructor(
         // debounce, resume to 56 before 5's fetch lands). Also require the live amount to still
         // match the amount this fetch was resolved for, so a quote priced for the old amount can't
         // be applied — and then signed with a mismatched memo / slippage floor — over the new one
-        // (#5310).
-        val liveAmount = srcAmount
-        if (
-            !isLiveInputQuotable ||
-                input.amount == null ||
-                liveAmount == null ||
-                liveAmount.compareTo(input.amount) != 0
-        ) {
+        // (#5310). Match the source/destination quote endpoints as well: changing to another
+        // routable pair with the same amount creates the same pre-debounce landing window.
+        if (!isLiveInputQuotable) {
             resetQuoteState()
+            return
+        }
+
+        val liveAmount = srcAmount
+        val (inputSrc, inputDst) = input.address
+        val isSuperseded =
+            input.amount == null ||
+                liveAmount == null ||
+                liveAmount.compareTo(input.amount) != 0 ||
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -406,11 +406,32 @@ constructor(
...
         }
 
+        val liveAmount = srcAmount
+        val (inputSrc, inputDst) = input.address
+        val isSuperseded =
+            input.amount == null ||
+                liveAmount == null ||
+                liveAmount.compareTo(input.amount) != 0 ||
+                !inputSrc.hasSameQuoteEndpointAs(selectedSrc.value) ||
+                !inputDst.hasSameQuoteEndpointAs(selectedDst.value)
```
aminsato: The supersession guard covers `amount` + source/dest endpoints, but not `slippageBps` / `externalRecipient` — and those are the same class of quote-shaping input this PR is guarding. They're read live at fetch-start (`slippageBps.value` / `externalRecipient.value` in `resolveQuote`, L374-375) and aren't carried on `QuoteInput`, so a slippage change with the amount and pair unchanged hits the identical pre-debounce landing window: the in-flight fetch was started with the old slippage (which set the `:LIM:` floor / memo), the user tightens slippage, and before the new value clears the 300ms debounce the old-slippage fetch lands, matches amount+endpoints, and is applied — staging exactly the mismatched floor #5310 is about, just via a different trigger.

Since the guard has no record of which slippage/recipient the fetch used (they aren't on `QuoteInput`), it can't currently detect this. Is the slippage/recipient variant intentionally out of scope for #5310, or worth a follow-up? To cover it, slippage/recipient would need to be captured on `QuoteInput` and compared here the same way.

aminsato (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -406,11 +406,32 @@ constructor(
...
+        // (#5310). Match the source/destination quote endpoints as well: changing to another
+        // routable pair with the same amount creates the same pre-debounce landing window.
         if (!isLiveInputQuotable) {
             resetQuoteState()
             return
         }
 
+        val liveAmount = srcAmount
+        val (inputSrc, inputDst) = input.address
+        val isSuperseded =
```
rkokhatskyi, line 435: `isSuperseded` checks only amount and pair, never `externalRecipient`, so a quote fetched under an abandoned recipient can land and apply after the address changes. OneInch/SwapKit bake that stale destination into the signed payload (`SwapTransactionBuilder.kt:228,242`) while Verify still shows the newly typed address.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
rkokhatskyi: `isSuperseded` also never compares `slippageBps`, so a quote resolved under an abandoned tolerance can still land and apply after the user tightens it. THORChain/Maya's `:LIM` min-output floor is signed verbatim from that stale tolerance (`SwapTransactionBuilder.kt:111`), silently reverting the protection the user just changed.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
johnnyluo: Valid finding, and I do not think pushing it to a follow-up would be defensible. Slippage and recipient are quote-shaping inputs just like amount and pair: accepting an old result can sign a stale :LIM floor or stale routing payload. Commit c46bd675d now snapshots both values on QuoteInput, uses that exact snapshot for resolveQuote, and rejects the result if either live value changed. Dedicated gated regression tests cover both pre-debounce races; the full test suite and lint pass.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
johnnyluo: Agreed. Treating recipient as out of scope would leave a real signing-integrity hole. Commit c46bd675d captures externalRecipient in the debounced QuoteInput, passes that captured value into quote resolution, and compares it against the live valid recipient before applying success. A gated regression test changes the recipient while the old fetch is in flight and proves that result never sets hasQuote while the replacement request remains loading.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
johnnyluo: Agreed. This is the same stale-request invariant, not a separate feature. Commit c46bd675d captures slippageBps in QuoteInput, resolves the quote with that captured tolerance, and rejects the result when live slippage differs. The regression test tightens slippage while the Auto fetch is gated and verifies the abandoned result never becomes signable.

PR #5312, opened 2026-07-17
fix(evm): stop persisting a failed balance read as a real 0 (#5308)

### LINKED ISSUE #5308: [Fix] EVM: a failed balance read is persisted as a real `0` balance, indistinguishable from an empty wallet [Balances / Home]
# [Fix] EVM: a failed balance read is written to the DB as a real `0` balance [Balances / Home]

## Problem

A funded EVM account displays **`0` / `$0.00`** after a single transient network failure, and that zero is **persisted to Room**, so it survives an app restart and stays until a *successful* refresh replaces it. To the user it reads as "my funds disappeared".

This is the same defect class as #5276 (XRP, fixed in #5288) — but on the EVM path, which #5288's commit message explicitly assumed was **already correct**:

> Let the failure propagate: AccountsRepository already catches a thrown balance error and keeps the cached balance (warm start) or leaves the row in its loading placeholder (cold start), **matching how EVM/Cosmos balances behave.**

That premise is **true for Cosmos and false for EVM**. `CosmosApiImp.getBalance` uses `bodyOrThrow` and lets failures propagate. `EvmApiImp` swallows them into `BigInteger.ZERO`. #5276 closed by asking that "other chain balance implementations should be audited for the same `catch -> return ZERO` pattern" — this issue is that audit's result for EVM.

Affects all 13 EVM chains (Ethereum, BSC, Avalanche, Base, Arbitrum, Polygon, Optimism, Mantle, Blast, Cronos, ZkSync, Sei, Hyperliquid).

All line numbers below are on `origin/main` @ `d63067931`.

## The trace: how a failed read reaches persistence and the UI

This is the part that makes it a real bug rather than a defensive-default in an API layer. **The zero is written into Room over the last-known-good value, with no guard.**

**1 — `EvmApiImp` converts a network failure into `ZERO` instead of throwing** (`EvmApi.kt:153-168`):

```kotlin
override suspend fun getBalance(coin: Coin): BigInteger {
    return try {
        if (coin.isNativeToken) getETHBalance(coin.address)
        else getERC20Balance(coin.address, coin.contractAddress)
    } catch (e: SocketTimeoutException) {
        Timber.d("request time out, message: ${e.message}")
        BigInteger.ZERO          // <-- transport failure == empty account
    } catch (e: NetworkException) {
        Timber.d("RPC error fetching balance: status=%d message=%s", e.httpStatusCode, e.message)
        BigInteger.ZERO          // <-- HTTP/RPC failure == empty account
    }
}
```

**2 — `BalanceRepositoryImpl.getEvmTokenBalancesAndPrices` persists it unconditionally** (`BalanceRepository.kt:611-621`):

```kotlin
return coins.associate { coin ->
    val rawBalance = balances[coin.contractAddress] ?: BigInteger.ZERO   // <-- and again here

    tokenValueDao.insertTokenValue(                    // <-- unconditional write, no compare-to-cache
        TokenValueEntity(
            chain = coin.chain.id,
            address = address,
            ticker = coin.ticker,
            tokenValue = rawBalance.toString(),
        )
    )
```

The non-batch path does the same: `getTokenValue(...)` emits the balance and `.onEach { tokenValueDao.insertTokenValue(...) }` writes it (`BalanceRepository.kt:446-590`, write at `581-589`).

**3 — the write destroys the good value.** `TokenValueDao.insertTokenValue` is `@Insert(onConflict = OnConflictStrategy.REPLACE)` on `TokenValueEntity`, whose primary key is `["chain", "address", "ticker"]` — so the zero **replaces** the previously-known non-zero row rather than being added alongside it.

**4 — the safety net that saves XRP cannot fire for EVM.** `AccountsRepository.loadAddressBalances` seeds the UI from the DB (`AccountsRepository.kt:104`, `fetchAccountFromDb()`), then per chain:

```kotlin
} catch (e: Exception) {                              // AccountsRepository.kt:172-176
    Timber.e(e)
    // ignore                                          // <-- leaves the cached row intact
}
```

That `catch` is exactly the mechanism #5288 relies on to preserve XRP's cached balance. **Because the EVM path returns normally instead of throwing, this catch never runs** — the zero flows through `applyBalance(...)` (`AccountsRepository.kt:125-135`) into the emitted `AddressBalancesUpdate` and onto the Home/coin screens. Identical in shape to the iOS bug: it doesn't throw, so the existing error handling is dead.

**5 — it persists.** On the next cold start `fetchAccountFromDb()` reads back the poisoned `0`. If the device is still offline, the refresh fails again and the account continues to show `0` indefinitely.

## Which failure paths are actually reachable

Five `ZERO` returns exist on the EVM balance path. They are **not** equally real — this matters for the fix and for review:

| # | Path | Reachable? |
|---|------|-----------|
| 1 | `getBalance` → `catch (NetworkException)` (`EvmApi.kt:160`) | **Yes — the live one.** Catches *every* transport failure. |
| 2 | `getERC20Balance` → JSON-RPC `error` non-null (`EvmApi.kt:187-191`) | **Yes.** Rate limiting / `execution reverted` / node errors return HTTP 200 with an `error` object. |
| 3 | `getETHBalance` → JSON-RPC `error` non-null (`EvmApi.kt:313-316`) | **Yes.** Same, for native balances. |
| 4 | `getBalances` → Multicall3 sub-call `success == false` (`EvmApi.kt:264`) | **Yes — and this is the exact iOS bug.** See below. |
| 5 | `getERC20Balance` → decoder throws / null result (`EvmApi.kt:194-205`) | Narrow. Needs a malformed 200 response. Theoretical in practice. |

Two notes that raise the severity of #1:

- **`catch (e: SocketTimeoutException)` (`EvmApi.kt:157`, and again at `269-296` in `fetchBalancesPerToken`) is effectively dead code.** `HttpClientConfigurator` installs an `HttpCallValidator` that converts every `IOException` into a classified `NetworkException` before it escapes the client (`HttpClientConfigurator.kt:55-61`, `100-107`) — a socket timeout arrives as `NetworkException(kind = Timeout)`, so the `NetworkException` branch is what actually fires. A fix that only touches the `SocketTimeoutException` branch would change nothing.
- **EVM balance reads get no retry.** `HttpRequestRetry.retryOnExceptionIf` is gated on `isSafeMethod(request.method)` = GET/HEAD/OPTIONS only (`HttpClientConfigurator.kt:63-80`, `90-93`). EVM RPC is `http.post(rpcUrl)` (`EvmApi.kt:629-637`), so it is **excluded from the retry policy**. One transient timeout goes straight to a persisted zero, with no retry to save it.

`NetworkErrorKind` already distinguishes `Timeout` / `NoConnectivity` / `Transport` / `Http` (`RpcExtensions.kt:98-106`) — the classification needed to do this correctly already exists and is simply discarded here.

### The Multicall3 partial-failure path (#4)

`getBalances` batches via `aggregate3` and maps each sub-call result (`EvmApi.kt:261-266`):

```kotlin
val r = decoded[index]
contract to if (r.success) Multicall3.decodeUint256Word(r.returnData) else BigInteger.ZERO
```

The surrounding code falls back to `fetchBalancesPerToken` whenever the multicall **as a whole** errors, decodes short, or returns empty (`EvmApi.kt:236-258`) — that handling is good. But a **partial** failure, where the aggregate call succeeds and an individual sub-call returns `success = false`, is not an error: it silently becomes a `0` balance for that one token and is persisted like any other. This is precisely the iOS defect, mechanism for mechanism.

## Expected Behavior

- A failed balance read must **never** be persisted or displayed as `0`.
- On failure, keep the last-known cached balance (warm start) or stay in the loading/error state (cold start) — the behavior #5288 established for XRP.
- A genuinely empty account (`0x0` from a healthy node, or a Multicall3 sub-call that legitimately returns zero) must remain distinguishable from a read that failed.

## Steps to Reproduce

1. Open a vault holding a funded EVM asset (e.g. ETH or USDC on Ethereum) and let the balance load correctly.
2. Make the EVM RPC endpoint unreachable — enable airplane mode, use a network-link conditioner, or point the chain at a black-hole RPC via the custom-RPC setting.
3. Pull to refresh the Home screen.
4. The row and the portfolio total show `0` / `$0.00` (the price still resolves — only the balance failed).
5. Force-quit and reopen the app while still offline: the row still shows `0`, read back from Room. The pre-failure balance is gone.

## Root cause — the design defect

**`BigInteger` cannot express "unknown".** Every EVM balance function returns a non-nullable `BigInteger`, so "the read failed" and "this account holds nothing" are *the same value*, and no caller downstream can tell them apart at any layer. The persistence and UI layers are not doing anything wrong — they are faithfully storing and rendering a number the API layer told them was real. No amount of caller-side defensiveness fixes this while the type can't carry the distinction.

The codebase already has the right pattern in two places:

- `SPLTokenRepository.getBalance(coin): BigInteger?` returns **nullable**, and `BalanceRepository.getTokenValue` consumes it as `splTokenRepository.getBalance(coin) ?: splTokenRepository.getCachedBalance(coin)` (`BalanceRepository.kt:553-559`) — fall back to cache, never invent a zero.
- `RippleApi.getBalance` (post-#5288) propagates the failure and lets `AccountsRepository` preserve the cached value.

EVM is the outlier.

## Solution

Make failure representable, then preserve last-known values. Either shape works — the second matches what already shipped for XRP and needs the least new machinery:

1. **Nullable / `Result` return:** `getBalance`/`getERC20Balance`/`getBalances` return `BigInteger?` (or `Result<BigInteger>`), with `null` = "read failed". Mirror the SPL pattern in `BalanceRepository`: on `null`, skip the `insertTokenValue` write entirely and keep the cached row. For `getBalances`, omit failed keys from the returned map instead of mapping them to `ZERO`, and drop the `?: BigInteger.ZERO` at `BalanceRepository.kt:612`.
2. **Propagate (XRP parity):** delete the `catch → ZERO` blocks so the failure reaches `AccountsRepository`'s existing `catch` at `:172`, which already preserves the cached balance. For the Multicall3 partial failure, either fall back to `fetchBalancesPerToken` for the failed subset or propagate.

Whichever is chosen, **the guard must live at or above the persistence write** (`BalanceRepository.kt:614` and `:582`) so no failed read can reach Room, and a genuine on-chain zero must still be distinguishable from a failure.

### Must Do
- [ ] Make a failed EVM balance read distinguishable from a genuine zero (nullable/`Result`, or propagate).
- [ ] Stop `BalanceRepository` persisting a failed read — never `insertTokenValue` a zero that came from a failure; keep the cached row.
- [ ] Handle the Multicall3 **partial** sub-call failure (`EvmApi.kt:264`) — do not map `success = false` to `ZERO`.
- [ ] Remove or correct the dead `catch (e: SocketTimeoutException)` branches (`EvmApi.kt:157`, `:280`), which `HttpCallValidator` makes unreachable.
- [ ] Unit tests: failed read preserves the cached balance; a genuine zero still persists as zero; a Multicall3 partial failure does not zero the other tokens. `RippleApiBalanceTest.kt` (added by #5288) is the model.

### Must NOT Do
- Don't change unrelated code.
- Don't touch the non-EVM balance paths (Cosmos/Ripple/SPL already behave correctly) beyond reusing their pattern.
- Don't alter the existing whole-multicall → per-token fallback (`EvmApi.kt:236-258`) — it is correct.

## Acceptance Criteria
- [ ] A funded EVM account whose balance read times out keeps showing its last-known balance, not `0`.
- [ ] The failed read is not written to Room — after force-quit + reopen, the last-known balance is still shown.
- [ ] A genuinely empty EVM account still displays `0`.
- [ ] A Multicall3 partial sub-call failure does not zero that token's balance.
- [ ] `./gradlew :data:test` passes, including new coverage for the three cases above.

## Cross-platform note

iOS has the **same class of bug** via a different mechanism: EVM balances are batched through Multicall3 `aggregate3` with `allowFailure: true`, a partial sub-call failure decodes to `nil`, and the caller maps `?? 0` — a real zero for a funded coin. It doesn't throw, so iOS's fallback error handling never fires either. A PR is in flight there. Android reaches the same end state from a blunter trigger (a plain network timeout), and additionally **persists** it. Worth fixing consistently across platforms — and worth noting that `EvmApi.kt:264` is the *identical* Multicall3 partial-failure mechanism, so Android has both variants.

## Severity

**High.** No fund loss and no incorrect signing — this is display/cache correctness. But: it is triggered by an ordinary transient timeout (not a rare edge case), it is not mitigated by retry (POST is excluded from the retry policy), it affects all 13 EVM chains including the most-used assets, it **persists across app restarts**, and the user-visible symptom — a funded wallet reporting `0` — is the single most alarming thing a wallet can show. #5276 was filed and fixed for exactly this symptom on one chain; the EVM path is the larger surface and was left unaudited on a false premise.

---

*Static analysis of `origin/main` @ `d63067931`; not reproduced on-device.*

johnnyluo:
## Summary

A funded EVM account showed **`0` / `$0.00`** after a single transient network failure, and that zero was **written to Room** over the last-known balance — surviving app restarts until a *successful* refresh replaced it, reading as "my funds disappeared".

Root cause: `EvmApiImp` swallowed every transport/RPC failure into `BigInteger.ZERO`, so "read failed" and "account is empty" became the *same value*. The persistence and UI layers then faithfully stored and rendered a fake `0`. Same defect class as #5276 (XRP, fixed in #5288), which wrongly assumed the EVM path already propagated failures.

Changes:
- **`getBalance`** now propagates instead of catching into `ZERO`, so `AccountsRepository`'s existing per-chain `catch` keeps the cached value (XRP parity). Dead `SocketTimeoutException` branches removed — `HttpCallValidator` makes them unreachable.
- **`getETHBalance` / `getERC20Balance`** throw on a JSON-RPC `error` (rate limiting / node errors) rather than returning `ZERO`.
- **`getBalances`** omits a Multicall3 sub-call that reports `success == false`, and any per-token failure, instead of mapping it to `ZERO` — siblings and genuine on-chain zeros are untouched.
- **`BalanceRepository.getEvmTokenBalancesAndPrices`** skips the `insertTokenValue` write for a token missing from the map (failed read), preserving the cached row; genuine zeros still persist.

## Issue

Closes #5308

## Test Plan

- [x] `./gradlew :data:testDebugUnitTest` passes, incl. new `EvmApiBalanceTest` cases:
  - a failed read (403 / timeout) propagates instead of becoming `0`
  - a genuine on-chain `0x0` still resolves to zero
  - a Multicall3 partial sub-call failure is omitted, siblings unaffected
  - a genuine zero sub-call stays a real zero
  - a failed per-token read (no-Multicall3 chain) is omitted
- [ ] On-device: funded EVM asset, kill RPC (airplane mode / black-hole RPC), pull-to-refresh → row keeps last-known balance; force-quit + reopen while offline → still shown (not `0`)

## Summary by CodeRabbit

- **Bug Fixes**
  - Balance read failures are no longer incorrectly displayed or saved as zero.
  - Genuine zero balances remain correctly recognized and displayed.
  - Failed token reads are skipped while existing cached balances are preserved.
  - Partial Multicall results now retain successful token balances without inventing values for failed requests.
  - Invalid or incomplete balance data is handled safely.
  - QR code cancellation handling no longer incorrectly reports cancellations as invalid QR errors.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -261,14 +261,16 @@ class EvmApiImp(
                 return fetchBalancesPerToken(address, contractAddresses)
             }
 
-        // A partial failure — the aggregate call succeeded but an individual sub-call reports
-        // `success == false` — is a failed read for that one token, not a zero balance. Omit it
-        // from the map (rather than mapping to ZERO) so the caller keeps its cached value and never
-        // persists a fake 0 (#5308).
+        // A failed read for a single token — the sub-call reports `success == false`, or reports
+        // success but carries empty/malformed data that can't be decoded (a non-standard token, a
+        // proxy returning no data without reverting) — is not a zero balance. Omit it from the map
+        // (rather than mapping to ZERO) so the caller keeps its cached value and never persists a
+        // fake 0 (#5308). A genuine on-chain zero is a full zero word and decodes to ZERO.
         return contractAddresses
             .mapIndexedNotNull { index, contract ->
                 val r = decoded[index]
-                if (r.success) contract to Multicall3.decodeUint256Word(r.returnData) else null
+                if (!r.success) null
+                else Multicall3.decodeUint256WordOrNull(r.returnData)?.let { contract to it }
             }
             .toMap()
     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -28,7 +28,6 @@ import io.ktor.client.request.post
 import io.ktor.client.request.setBody
 import io.ktor.client.statement.bodyAsText
 import java.math.BigInteger
-import java.net.SocketTimeoutException
 import javax.inject.Inject
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
@@ -89,9 +88,11 @@ interface EvmApi {
      * balance (via `Multicall3.getEthBalance`); non-empty entries request `ERC20.balanceOf`.
      *
      * Returns a map keyed by each input contract address (the native "" key included) to its
-     * balance, with failed sub-calls decoded as [BigInteger.ZERO]. Falls back to per-token
+     * balance. A sub-call that fails is omitted from the map (never mapped to [BigInteger.ZERO]) so
+     * the caller keeps its cached value instead of persisting a fake 0 (#5308); a genuine on-chain
+     * zero is present with value [BigInteger.ZERO]. Falls back to per-token
      * [getERC20Balance]/[getBalance] on chains without a canonical Multicall3, for a single
-     * address, or when the multicall errors — keeping results byte-identical to the per-token path.
+     * address, or when the multicall errors.
      */
     suspend fun getBalances(
         address: String,
@@ -150,22 +151,14 @@ class EvmApiImp(
     private val chain: Chain,
 ) : EvmApi {
 
-    override suspend fun getBalance(coin: Coin): BigInteger {
-        return try {
-            if (coin.isNativeToken) getETHBalance(coin.address)
-            else getERC20Balance(coin.address, coin.contractAddress)
-        } catch (e: SocketTimeoutException) {
-            Timber.d("request time out, message: ${e.message}")
-            BigInteger.ZERO
-        } catch (e: NetworkException) {
-            Timber.d(
-                "RPC error fetching balance: status=%d message=%s",
-                e.httpStatusCode,
-                e.message,
-            )
-            BigInteger.ZERO
-        }
-    }
+    // A failed read must propagate, never collapse into ZERO: a zero is indistinguishable from a
+    // genuinely empty account and would be persisted over the last-known balance, reading as "my
+    // funds disappeared" (#5308). Letting it throw lets AccountsRepository's per-chain catch keep
+    // the cached value. A healthy node returning `0x0` for an empty account still resolves to a
+    // real zero inside the helpers below.
+    override suspend fun getBalance(coin: Coin): BigInteger =
+        if (coin.isNativeToken) getETHBalance(coin.address)
+        else getERC20Balance(coin.address, coin.contractAddress)
 
     override suspend fun getERC20Balance(address: String, contractAddress: String): BigInteger {
         val rpcResp =
@@ -183,10 +176,14 @@ class EvmApiImp(
                 },
             )
         if (rpcResp.error != null) {
-            Timber.d(
-                "get erc20 balance,contract: $contractAddress,address: $address error: ${rpcResp.error.message}"
+            // An RPC-level error (rate limiting, node error) is a failed read, not an empty
+            // balance — surface it so the balance layer keeps the cached value (#5308).
+            throw NetworkException(
+                httpStatusCode = 0,
+                message =
+                    "erc20 balance rpc error, contract=$contractAddress address=$address: " +
+                        "${rpcResp.error.message}",
             )
-            return BigInteger.ZERO
         }
         return rpcResp.result?.let {
             try {
@@ -257,11 +254,14 @@ class EvmApiImp(
                 return fetchBalancesPerToken(address, contractAddresses)
             }
 
+        // A partial failure — the aggregate call succeeded but an individual sub-call reports
+        // `success == false` — is a failed read for that one token, not a zero balance. Omit it
+        // from the map (rather than mapping to ZERO) so the caller keeps its cached value and never
+        // persists a fake 0 (#5308).
         return contractAddresses
-            .mapIndexed { index, contract ->
+            .mapIndexedNotNull { index, contract ->
                 val r = decoded[index]
-                contract to
-                    if (r.success) Multicall3.decodeUint256Word(r.returnData) else BigInteger.ZERO
+                if (r.success) contract to Multicall3.decodeUint256Word(r.returnData) else null
             }
             .toMap()
     }
@@ -273,31 +273,26 @@ class EvmApiImp(
         contractAddresses
             .map { contract ->
                 async {
-                    contract to
-                        try {
+                    try {
+                        contract to
                             if (contract.isEmpty()) getETHBalance(address)
                             else getERC20Balance(address, contract)
-                        } catch (e: SocketTimeoutException) {
-                            Timber.d(
-                                "per-token balance timeout, contract=%s address=%s message=%s",
-                                contract,
-                                address,
-                                e.message,
-                            )
-                            BigInteger.ZERO
-                        } catch (e: NetworkException) {
-                            Timber.d(
-                                "per-token balance RPC error, contract=%s address=%s status=%d message=%s",
-                                contract,
-                                address,
-                                e.httpStatusCode,
-                                e.message,
-                            )
-                            BigInteger.ZERO
-                        }
+                    } catch (e: NetworkException) {
+                        // Failed read — omit the token so its cached balance is preserved rather
+                        // than overwritten with a fake 0 (#5308).
+                        Timber.d(
+                            "per-token balance failed, contract=%s address=%s status=%d message=%s",
+                            contract,
+                            address,
+                            e.httpStatusCode,
+                            e.message,
+                        )
+                        null
+                    }
                 }
             }
             .awaitAll()
+            .filterNotNull()
             .toMap()
     }
 
@@ -311,8 +306,12 @@ class EvmApiImp(
                 },
             )
         if (rpcResp.error != null) {
-            Timber.d("get balance ,address: $address error: ${rpcResp.error.message}")
-            return BigInteger.ZERO
+            // A failed read, not an empty account — propagate so the cached balance is kept
+            // (#5308).
+            throw NetworkException(
+                httpStatusCode = 0,
+                message = "eth balance rpc error, address=$address: ${rpcResp.error.message}",
+            )
         }
         return rpcResp.result.convertToBigIntegerOrZero()
     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -185,20 +185,27 @@ class EvmApiImp(
                         "${rpcResp.error.message}",
             )
         }
-        return rpcResp.result?.let {
-            try {
-                EthereumFunction.balanceErc20Decoder(it)
-            } catch (e: Exception) {
-                Timber.d("get erc20 balance,contract: $contractAddress,address: $address error: $e")
-                BigInteger.ZERO
-            }
-        }
-            ?: run {
-                Timber.d(
-                    "get erc20 balance,contract: $contractAddress,address: $address error: response is null"
+        // A null result or a decode failure is a malformed read, not a genuine zero (a healthy node
+        // returns a 32-byte zero word for an empty balance, which decodes cleanly). Propagate so
+        // the
+        // cached balance is kept rather than overwritten with a fake 0 (#5308).
+        val result =
+            rpcResp.result
+                ?: throw NetworkException(
+                    httpStatusCode = 0,
+                    message =
+                        "erc20 balance null result, contract=$contractAddress address=$address",
                 )
-                BigInteger.ZERO
-            }
+        return try {
+            EthereumFunction.balanceErc20Decoder(result)
+        } catch (e: Exception) {
+            throw NetworkException(
+                httpStatusCode = 0,
+                message =
+                    "erc20 balance decode failed, contract=$contractAddress address=$address: " +
+                        "${e.message}",
+            )
+        }
     }
 
     override suspend fun getBalances(
@@ -313,7 +320,24 @@ class EvmApiImp(
                 message = "eth balance rpc error, address=$address: ${rpcResp.error.message}",
             )
         }
-        return rpcResp.result.convertToBigIntegerOrZero()
+        // A null/malformed result with no explicit error is still a failed read, not an empty
+        // account (a healthy node returns "0x0" for zero) — propagate so the cached balance is kept
+        // (#5308).
+        val cleaned = rpcResp.result?.removePrefix("0x")
+        if (cleaned.isNullOrEmpty()) {
+            throw NetworkException(
+                httpStatusCode = 0,
+                message = "eth balance null result, address=$address",
+            )
+        }
+        return try {
+            BigInteger(cleaned, 16)
+        } catch (e: NumberFormatException) {
+            throw NetworkException(
+                httpStatusCode = 0,
+                message = "eth balance malformed result, address=$address: ${rpcResp.result}",
+            )
+        }
     }
 
     override suspend fun getNonce(address: String): BigInteger {
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/Multicall3.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/Multicall3.kt
@@ -110,7 +110,10 @@ object Multicall3 {
      */
     fun decodeUint256WordOrNull(returnData: String): BigInteger? {
         val cleaned = returnData.removePrefix("0x")
-        return if (cleaned.isEmpty()) null else cleaned.toBigIntegerOrNull(16)
+        // Only a full 32-byte word is a valid uint256 result. A short/truncated word (e.g. a 4-byte
+        // "0x0000000a") is malformed data — omit it (failed read) rather than decode it to a bogus
+        // balance, matching the per-token `balanceErc20Decoder`, which rejects the same bytes.
+        return if (cleaned.length != HEX_PER_WORD) null else cleaned.toBigIntegerOrNull(16)
     }
 
     private fun encodeCall3Tuple(call: Call3): String {
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/Multicall3.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/Multicall3.kt
@@ -1,6 +1,5 @@
 package com.vultisig.wallet.data.chains.helpers
 
-import com.vultisig.wallet.data.common.convertToBigIntegerOrZero
 import com.vultisig.wallet.data.common.remove0x
 import com.vultisig.wallet.data.models.Chain
 import java.math.BigInteger
@@ -103,8 +102,16 @@ object Multicall3 {
         }
     }
 
-    /** Decodes a single left-padded uint256 return word (e.g. a `balanceOf` result). */
-    fun decodeUint256Word(returnData: String): BigInteger = returnData.convertToBigIntegerOrZero()
+    /**
+     * Decodes a single left-padded uint256 return word (e.g. a `balanceOf` result), or null when
+     * the word is empty/malformed. A successful sub-call that carries no decodable data is a failed
+     * read, not a genuine zero (a real zero balance is a full 32-byte zero word) — returning null
+     * lets the caller omit it and keep the cached balance rather than persist a fake 0 (#5308).
+     */
+    fun decodeUint256WordOrNull(returnData: String): BigInteger? {
+        val cleaned = returnData.removePrefix("0x")
+        return if (cleaned.isEmpty()) null else cleaned.toBigIntegerOrNull(16)
+    }
 
     private fun encodeCall3Tuple(call: Call3): String {
         val data = call.callData.remove0x()
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/BalanceRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/BalanceRepository.kt
@@ -608,36 +608,44 @@ constructor(
 
         val currency = appCurrencyRepository.currency.first()
 
-        return coins.associate { coin ->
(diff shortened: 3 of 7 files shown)


johnnyluo:
Addressed both review findings in 82f6a725c:

- **ERC-20 / ETH null & malformed results** — `getERC20Balance` now throws `NetworkException` on a decode failure or null `result`, and `getETHBalance` throws on a null/malformed `result`, instead of collapsing to `ZERO`. Same #5308 defect class as the RPC-error path — a healthy node returns a 32-byte zero word / "0x0" for a genuine empty account, which still resolves to a real zero.
- **Tests** — switched `EvmApiBalanceTest` from `runBlocking` to `runTest`.

`:data:testDebugUnitTest` and `:data:lintDebug` pass.

aminsato (commented):
Review of the failed-balance-read fix. The core change (propagate instead of collapsing to 0) is right and the tests are solid. Five points below — the first three are behavioral gaps where a fake/blank 0 can still surface or the fix over-reaches.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -150,22 +151,14 @@ class EvmApiImp(
...
-            )
-            BigInteger.ZERO
-        }
-    }
+    // A failed read must propagate, never collapse into ZERO: a zero is indistinguishable from a
+    // genuinely empty account and would be persisted over the last-known balance, reading as "my
+    // funds disappeared" (#5308). Letting it throw lets AccountsRepository's per-chain catch keep
+    // the cached value. A healthy node returning `0x0` for an empty account still resolves to a
+    // real zero inside the helpers below.
+    override suspend fun getBalance(coin: Coin): BigInteger =
```
aminsato, line 159: `getBalance` throwing (instead of returning ZERO) fixes the batch path but regresses the single-chain refresh. `AccountsRepository.emitRefreshAddress` (loadAddress → ChainTokens / TokenDetail screens) refreshes tokens in a sequential `accounts.map { getTokenBalanceAndPrice(...).first() }`, whose EVM branch reaches this method via `getTokenValue`. One token's failed RPC read now throws out of the whole `.map`, so the refreshed `Address` is never emitted and the flow lands in the ViewModel's `.catch` — **every** sibling token on that chain stays on stale cached values, not just the failing one. Before this change only the failing token showed a fake 0 and siblings refreshed normally. Isolate per-token failures on that path (per-token try/catch that keeps the cached account) so it matches the batch path's behavior.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -257,11 +261,14 @@ class EvmApiImp(
...
+        // `success == false` — is a failed read for that one token, not a zero balance. Omit it
+        // from the map (rather than mapping to ZERO) so the caller keeps its cached value and never
+        // persists a fake 0 (#5308).
         return contractAddresses
-            .mapIndexed { index, contract ->
+            .mapIndexedNotNull { index, contract ->
                 val r = decoded[index]
-                contract to
-                    if (r.success) Multicall3.decodeUint256Word(r.returnData) else BigInteger.ZERO
+                if (r.success) contract to Multicall3.decodeUint256Word(r.returnData) else null
```
aminsato: The Multicall3 batch path still collapses a malformed success into a fake 0, so the #5308 invariant isn't fully enforced here. `decodeUint256Word` delegates to `convertToBigIntegerOrZero`, which returns ZERO for empty or truncated `returnData`. A sub-call with `success == true` but empty/short data (non-standard token, a proxy that returns no data without reverting, a corrupted aggregate response) still persists 0 over the cached balance — the exact fake-zero this PR targets, reachable through aggregate3. Treat an undecodable success word as a failed read (omit it) rather than 0.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -311,10 +313,31 @@ class EvmApiImp(
...
+            // (#5308).
+            throw NetworkException(
+                httpStatusCode = 0,
+                message = "eth balance rpc error, address=$address: ${rpcResp.error.message}",
+            )
+        }
+        // A null/malformed result with no explicit error is still a failed read, not an empty
+        // account (a healthy node returns "0x0" for zero) — propagate so the cached balance is kept
+        // (#5308).
+        val cleaned = rpcResp.result?.removePrefix("0x")
```
aminsato, line 328: This now treats `"0x"` as a failed read. The replaced `convertToBigIntegerOrZero()` resolved `"0x"` (empty hex after stripping the prefix) to a genuine ZERO; here `cleaned` becomes `""` and throws. Some nodes/proxies return `"0x"` rather than `"0x0"` for an empty native account, so a real zero balance is misclassified as a failure and omitted — on a cache-less new account it then renders blank instead of 0. Accept empty-hex as zero.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
```diff
@@ -183,25 +176,36 @@ class EvmApiImp(
...
             )
         if (rpcResp.error != null) {
-            Timber.d(
-                "get erc20 balance,contract: $contractAddress,address: $address error: ${rpcResp.error.message}"
+            // An RPC-level error (rate limiting, node error) is a failed read, not an empty
+            // balance — surface it so the balance layer keeps the cached value (#5308).
+            throw NetworkException(
+                httpStatusCode = 0,
+                message =
+                    "erc20 balance rpc error, contract=$contractAddress address=$address: " +
```
aminsato, line 184: `httpStatusCode = 0` is the sentinel for a transport failure where no HTTP response was received, but these six new throws fire on a 200 response with an RPC-level error / null / malformed body. A retry or UI layer keying off `httpStatusCode == 0` to detect offline/transport conditions will misclassify an application-level RPC error (the node did respond) as offline. Use a distinct signal for RPC-body errors.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BalanceRepository.kt
```diff
@@ -608,36 +608,44 @@ constructor(
...
-                TokenValue(value = rawBalance, unit = coin.ticker, decimals = coin.decimal)
-            val price = tokenPriceRepository.getPrice(coin, currency).first()
+        return coins
+            .mapNotNull { coin ->
+                // A missing key means the balance read failed (getBalances omits failed tokens):
+                // skip
+                // the write and the emission so the cached row is preserved rather than overwritten
+                // with a fake 0 (#5308). A genuine on-chain zero is present in the map and still
+                // persists below.
+                val rawBalance = balances[coin.contractAddress] ?: return@mapNotNull null
```
aminsato, line 618: For a token with no cached `TokenValueEntity` row yet, omitting it on a failed first read skips both the write and the emission, so `AccountsRepository:130` falls back to the cache-less `acc` and the token renders blank until a later refresh succeeds — the pre-fix path showed 0. This is the trade-off of the omit approach; flagging it for a newly-added EVM token whose first load is flaky.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
johnnyluo: Fixed in 9df0544df. You're right — the single-chain `loadAddress` path goes through `emitRefreshAddress`, which maps per-token with `getTokenBalanceAndPrice(...).first()`; the EVM branch reaches the now-throwing `getBalance`, so one failure dropped the whole emission and left every sibling unrefreshed. Wrapped each per-token read in try/catch, falling back to the cached value for just the failing token — now matches the batch path's isolation.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
johnnyluo: Fixed in 9df0544df. Correct — `decodeUint256Word` delegated to `convertToBigIntegerOrZero`, which collapses empty/truncated `returnData` to 0. Replaced it with `decodeUint256WordOrNull`, which returns null for an empty/malformed word; the aggregate3 path now omits an undecodable success word (via `mapIndexedNotNull`) instead of persisting a fake 0. A genuine on-chain zero is a full 32-byte zero word and still decodes to ZERO. Added coverage at both the helper level and the `getBalances` level (`success == true` with empty data → omitted).

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
johnnyluo: Keeping as-is. For `eth_getBalance` the JSON-RPC spec encodes a QUANTITY as minimal hex with zero as `"0x0"` — a compliant node never returns `"0x"` for a real zero balance. `"0x"` therefore only comes from a non-compliant/error response, and treating it as a failed read (keep cached) is exactly the #5308 invariant. Accepting `"0x"` as zero would reintroduce the fake-zero for the native token whenever a proxy returns `"0x"` on error. The flagged scenario needs both a non-compliant node AND no cache, and self-heals on the next refresh — not worth reopening the bug this PR closes.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
johnnyluo: Keeping as-is. The offline/transport sentinel is `NetworkException.kind`, not `httpStatusCode`. `toNetworkErrorKind()` explicitly ignores `kind == Http` and only returns transport kinds (Timeout/NoConnectivity/Transport), which are set solely by `HttpClientConfigurator`. These new throws use the default `kind = Http`, so they can never be classified as offline, and the only `toNetworkErrorKind` consumer is `KeygenPeerDiscoveryViewModel` (not the balance path). No code keys off `httpStatusCode == 0` for offline detection — `0` here just means 'no meaningful HTTP status for an RPC-body error on a 200'.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BalanceRepository.kt
johnnyluo: Acknowledged — this is the intended trade-off, not a regression. For a brand-new EVM token with no cached row, a flaky first read renders blank briefly and fills in on the next successful refresh. The alternative (show 0) is precisely the #5308 fake-zero this PR removes; persisting a fabricated 0 for a token the user just added is worse than a momentary blank. Leaving as-is.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/Multicall3.kt
```diff
@@ -103,8 +102,16 @@ object Multicall3 {
...
-    fun decodeUint256Word(returnData: String): BigInteger = returnData.convertToBigIntegerOrZero()
+    /**
+     * Decodes a single left-padded uint256 return word (e.g. a `balanceOf` result), or null when
+     * the word is empty/malformed. A successful sub-call that carries no decodable data is a failed
+     * read, not a genuine zero (a real zero balance is a full 32-byte zero word) — returning null
+     * lets the caller omit it and keep the cached balance rather than persist a fake 0 (#5308).
+     */
+    fun decodeUint256WordOrNull(returnData: String): BigInteger? {
+        val cleaned = returnData.removePrefix("0x")
+        return if (cleaned.isEmpty()) null else cleaned.toBigIntegerOrNull(16)
```
rkokhatskyi: `decodeUint256WordOrNull` disagrees with the per-token decoder for malformed data: it accepts any non-empty hex via `toBigIntegerOrNull(16)` with no length check, so a 4-byte `0x0000000a` becomes balance 10, while `balanceErc20Decoder` (EthereumFunctions.kt:68) rejects the same bytes and gets omitted, so the two paths return a different value for the same response

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
rkokhatskyi, line 159: `getBalance` now throws on any transient RPC failure, but `AccountsRepository.loadAccount` (`AccountsRepository.kt:591`) calls `getTokenBalanceAndPrice(...).first()` with no catch, unlike the emitRefreshAddress fix in this same PR. Its only caller, `SwapTokenSelector.kt:214`, swallows the throw and returns without setting selectedSrcId, so picking a new swap token can silently no-op

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/Multicall3.kt
johnnyluo: Fixed in a4c0286b3. You're right — the two decoders disagreed. `decodeUint256WordOrNull` accepted any non-empty hex via `toBigIntegerOrNull(16)`, so a truncated 4-byte word decoded to a bogus balance, while `balanceErc20Decoder` (ABI decode) rejects the same bytes and the token is omitted. Added a full-word length check (`cleaned.length != HEX_PER_WORD → null`), so a short/malformed word is now treated as a failed read and omitted — matching the per-token path and the KDoc's 'full 32-byte word' contract. A genuine on-chain zero is a full 32-byte zero word and still decodes to ZERO. Test coverage added for the truncated case.

data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
johnnyluo: Fixed in a4c0286b3. Correct — `loadAccount` called `getTokenBalanceAndPrice(...).first()` with no catch, so the now-throwing `getBalance` propagated out; `SwapTokenSelector`'s `catch` then returned early, before setting `selectedSrcId`/`selectedDstId`, silently dropping the pick. Wrapped the live read in try/catch inside `loadAccount`, falling back to `getCachedTokenBalanceAndPrice` on failure (CancellationException rethrown) — same isolation as the batch/emitRefreshAddress paths. `loadAccount` no longer throws on a transient read, so the selection registers and the row shows the cached value instead of a fake 0.

rkokhatskyi (approved):
LGTM

PR #5324, opened 2026-07-18
fix(keysign): recover DKLS-family signature from relay on failure (#5314)

### LINKED ISSUE #5314: DKLS keysign: device stuck on "Signing" when a relay message is lost — no completion-check fallback
# [Fix] DKLS-family keysign has no completion-check fallback when a relay message is lost [Keysign / "Signing" screen]

## Problem
During a DKLS 2-of-2 keysign, a single relay message lost to a transport error (a ktor `HttpRequestRetry` connection-reuse race producing `EOFException: Channel is already closed` on `GET .../router/message/...`) leaves one device permanently stuck on the "Signing" screen while the peer completes successfully. The stuck device's local DKLS session never reaches `isFinished`, and `DKLSKeysign.keysignOneMessageWithRetry`'s catch block only retries or rethrows — it never calls `KeysignVerify.checkKeysignComplete()` to fetch the completed signature that the successful peer already posted to the relay via `markLocalPartyKeysignComplete`. The device times out (60s no-message timeout, 3 retries) and fails with "signatures empty" even though the valid 2-of-2 signature has been on the relay the whole time. This is especially serious when the stuck device is the one responsible for broadcasting: a fully-signed transaction may never be broadcast.

## Expected Behavior
When a DKLS-family signing attempt fails or times out, the device should check the relay for an already-completed signature (as the legacy GG20 path does in `KeysignViewModel.signMessageWithRetry`) and, if one exists, use it and complete successfully instead of failing a ceremony the network as a whole completed.

## Steps to Reproduce
1. Start a DKLS 2-of-2 keysign between two devices (relay: `api.vultisig.com/router`).
2. On one device (observed on the initiate device), induce a transport error on a `GET .../router/message/...` request at the final message round, so the response body is dropped mid-`HttpRequestRetry` (`EOFException: Channel is already closed`). The lost message is not redelivered; every subsequent `GET` returns `[]`.
3. Observe: the peer device completes and shows the success screen; the affected device stays on "Signing", hits its 60s no-message timeout, retries 3× against a peer that has already left the ceremony, and fails with "signatures empty".

Note: reliably reproducing depends on inducing the transport error at the final round; a unit test simulating the condition is the practical repro (see Scope).

### Must Do
- [ ] In `DKLSKeysign.keysignOneMessageWithRetry` (and the equivalent retry/catch paths in `SchnorrKeysign` and `MldsaKeysign`), on a signing failure/timeout call `KeysignVerify.checkKeysignComplete(messageToSign)` before retrying/rethrowing; if a signature is returned, store it in `signatures` and return successfully — mirroring the existing GG20 fallback in `KeysignViewModel.signMessageWithRetry`.
- [ ] Apply the fallback consistently across all three DKLS-family keysign classes (`DKLSKeysign`, `SchnorrKeysign`, `MldsaKeysign`).
- [ ] Add a unit test simulating "local `isFinished` never arrives, but the relay has the completed signature" and asserting the device recovers the signature instead of failing.

### Must NOT Do
- Don't change unrelated code.
- Don't modify the legacy GG20 path (`KeysignViewModel.signMessageWithRetry`) — it already has the correct behavior and serves as the reference pattern.
- Don't attempt to fix the underlying ktor `HttpRequestRetry`/connection-reuse race or relay redelivery semantics in this issue — the fix is the completion-check fallback.

## Acceptance Criteria
- [ ] On signing failure/timeout, each DKLS-family keysign class queries `checkKeysignComplete` and completes with the relay-recovered signature when one is present
- [ ] New unit test covering the "relay has signature, local session never finishes" recovery path passes

## Context
- Observed configuration: DKLS 2-of-2 keysign; the **initiate** device was the one stuck on "Signing" while the joiner succeeded. Relay: `api.vultisig.com/router`.
- Root-cause stack trace from the stuck device:

```
DKLSKeysign  E  Failed to get messages
  com.vultisig.wallet.data.utils.NetworkException: Network request failed
  Caused by: java.io.EOFException: Channel is already closed
    at io.ktor.utils.io.ByteReadChannelOperationsKt.readFully(...)
    at io.ktor.client.plugins.HttpRequestRetryKt$HttpRequestRetry$2$1.invokeSuspend(HttpRequestRetry.kt:304)
```

- Mechanics: the peer that finishes posts the full signature to the relay via `KeysignVerify.markLocalPartyKeysignComplete`; `KeysignVerify.checkKeysignComplete()` exists to read it back, but the DKLS read side never calls it. `pullInboundMessages` only exits on local `isFinished`.
- The same structural gap exists in iOS `DKLSKeysign.swift` (catch block is retry-or-throw, no `checkKeysignComplete`), so this bug is cross-platform — this issue covers the Android fix only; a corresponding iOS issue should be filed in the iOS repo.
- Repository: vultisig/vultisig-android.

aminsato:
## Summary

Fixes #5314.

### Transaction hash
https://explorer.zksync.io/tx/0x16e66d34ac74a01dd52f2b436958911f829618e31cad15ce3078940d7e8e9cd1

During a DKLS-family keysign, a single relay message lost to a transport error (a ktor `HttpRequestRetry` connection-reuse race → `EOFException: Channel is already closed`) can leave one device permanently stuck on the "Signing" screen while the peer completes successfully. The stuck device's local session never reaches `isFinished`, so it times out and fails with **"signatures empty"** — even though the completed 2-of-2 signature has been on the relay the whole time (the successful peer posts it via `markLocalPartyKeysignComplete`). This is especially serious when the stuck device is the one responsible for broadcasting.

The root cause: the `catch` blocks in `DKLSKeysign`, `SchnorrKeysign`, and `MldsaKeysign` only **retry or rethrow** — they never query the relay for an already-completed signature, unlike the legacy GG20 path (`KeysignViewModel.signMessageWithRetry`) which already does this correctly.

## Changes

- On a signing failure/timeout, each of the three DKLS-family keysign classes now calls `KeysignVerify.checkKeysignComplete(msgHash)` **before** retrying/rethrowing. If the relay already holds the signature, it is stored in `signatures` and the ceremony completes successfully.
- The lookup uses **`msgHash`** (not the raw `messageToSign`), matching each class's `markLocalPartyKeysignComplete(msgHash, …)` write key.
- Mirrors the existing GG20 fallback pattern; the GG20 path is left unchanged.

## Test plan

- [x] `./gradlew assembleDebug` compiles (`:app:compileDebugKotlin` green)
- [x] New regression test `DklsFamilyRelayRecoveryTest` — asserts that on a signing failure the relay **is** queried, with the correct `msgHash` key, across all three keysign classes. Without the fix `checkKeysignComplete` is never called and the test fails.

### Note on test scope

The test pins the *wiring* (the relay lookup happens with the right key) rather than asserting the recovered signature is stored, because `tss.KeysignResponse` is a JNI/native type whose static initializer loads the native `gojni` library and cannot be constructed — or even mocked — on the host JVM. This is the same constraint that makes the existing `SessionApiBodyReadTest` skip `checkKeysignComplete`. Full positive-path coverage would require an instrumented (emulator) test.

## Cross-platform note

The same structural gap exists in iOS `DKLSKeysign.swift` (catch block is retry-or-throw, no `checkKeysignComplete`). This PR covers the Android fix only; a corresponding iOS issue should be filed separately.

## Summary by CodeRabbit

- **Bug Fixes**
  - After a signing failure in DKLS, Schnorr, and MLDSA, the app now re-checks whether the signature was already completed remotely; if so, it saves the result and stops further retries.
  - Improved reliability when publishing “local party keysign complete” by retrying transient failures with a short backoff.
  - Swap Kit BTC transactions now verify all expected MPC signatures match the derived vault public key before compiling.
  - TON transactions now explicitly validate the signature cryptographically before compiling.
- **Tests**
  - Added a regression suite covering DKLS-family relay lookup on signing failure and confirming completion checks use the message hash.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
@@ -414,6 +414,11 @@ class DKLSKeysign(
             throw e
         } catch (e: Exception) {
             println("Failed to sign message ($messageToSign), error: ${e.localizedMessage}")
+            val keySignVerify = KeysignVerify(mediatorURL, sessionID, sessionApi)
+            keySignVerify.checkKeysignComplete(msgHash)?.let {
+                signatures[messageToSign] = it
+                return
+            }
             val maxRetries = if (heardFromEver.isEmpty()) 1 else 3
             if (attempt < maxRetries) {
                 keysignOneMessageWithRetry(attempt + 1, messageToSign)
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
@@ -417,6 +417,13 @@ class DKLSKeysign(
             val keySignVerify = KeysignVerify(mediatorURL, sessionID, sessionApi)
             keySignVerify.checkKeysignComplete(msgHash)?.let {
                 signatures[messageToSign] = it
+                // Recovery bypasses the normal pullInboundMessages success path, so clear any
+                // "waiting for peer" state ourselves — otherwise the screen stays pinned on
+                // WaitingForPeer through broadcast.
+                if (waitingNotified) {
+                    waitingNotified = false
+                    onPeersResumed?.invoke()
+                }
                 return
             }
             val maxRetries = if (heardFromEver.isEmpty()) 1 else 3
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/MldsaKeysign.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/MldsaKeysign.kt
@@ -125,6 +125,13 @@ class MldsaKeysign(
             val keySignVerify = KeysignVerify(mediatorURL, sessionID, sessionApi)
             keySignVerify.checkKeysignComplete(msgHash)?.let {
                 signatures[messageToSign] = it
+                // Recovery bypasses the normal pullInboundMessages success path, so clear any
+                // "waiting for peer" state ourselves — otherwise the screen stays pinned on
+                // WaitingForPeer through broadcast.
+                if (waitingNotified) {
+                    waitingNotified = false
+                    onPeersResumed?.invoke()
+                }
                 return
             }
             val maxRetries = if (heardFromEver.isEmpty()) 1 else MAX_PROTOCOL_RETRIES
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/MldsaKeysign.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/MldsaKeysign.kt
@@ -122,6 +122,11 @@ class MldsaKeysign(
             throw e
         } catch (e: Exception) {
             Timber.e(e, "Failed to sign message (%s)", messageToSign)
+            val keySignVerify = KeysignVerify(mediatorURL, sessionID, sessionApi)
+            keySignVerify.checkKeysignComplete(msgHash)?.let {
+                signatures[messageToSign] = it
+                return
+            }
             val maxRetries = if (heardFromEver.isEmpty()) 1 else MAX_PROTOCOL_RETRIES
             if (attempt < maxRetries) {
                 keysignOneMessage(attempt + 1, messageToSign)
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
@@ -417,6 +417,11 @@ class SchnorrKeysign(
             throw e
         } catch (e: Exception) {
             println("Failed to sign message ($messageToSign), error: ${e.localizedMessage}")
+            val keySignVerify = KeysignVerify(mediatorURL, sessionID, sessionApi)
+            keySignVerify.checkKeysignComplete(msgHash)?.let {
+                signatures[messageToSign] = it
+                return
+            }
             val maxRetries = if (heardFromEver.isEmpty()) 1 else 3
             if (attempt < maxRetries) {
                 keysignOneMessageWithRetry(attempt + 1, messageToSign)
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
@@ -420,6 +420,13 @@ class SchnorrKeysign(
             val keySignVerify = KeysignVerify(mediatorURL, sessionID, sessionApi)
             keySignVerify.checkKeysignComplete(msgHash)?.let {
                 signatures[messageToSign] = it
+                // Recovery bypasses the normal pullInboundMessages success path, so clear any
+                // "waiting for peer" state ourselves — otherwise the screen stays pinned on
+                // WaitingForPeer through broadcast.
+                if (waitingNotified) {
+                    waitingNotified = false
+                    onPeersResumed?.invoke()
+                }
                 return
             }
             val maxRetries = if (heardFromEver.isEmpty()) 1 else 3
--- a/app/src/test/java/com/vultisig/wallet/data/keygen/DklsFamilyRelayRecoveryTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/data/keygen/DklsFamilyRelayRecoveryTest.kt
@@ -5,9 +5,9 @@ import com.vultisig.wallet.data.common.md5
 import com.vultisig.wallet.data.mediator.Message
 import com.vultisig.wallet.data.models.KeyShare
 import com.vultisig.wallet.data.models.Vault
+import io.kotest.assertions.throwables.shouldThrow
+import io.kotest.matchers.collections.shouldContain
 import io.mockk.mockk
-import kotlin.test.assertFailsWith
-import kotlin.test.assertTrue
 import kotlinx.coroutines.test.runTest
 import org.junit.jupiter.api.Test
 
@@ -151,13 +151,9 @@ class DklsFamilyRelayRecoveryTest {
                 encryption = mockk(relaxed = true),
             )
 
-        assertFailsWith<RuntimeException> { keysign.keysignWithRetry() }
+        shouldThrow<RuntimeException> { keysign.keysignWithRetry() }
 
-        assertTrue(
-            expectedMsgHash in sessionApi.completionLookups,
-            "expected the relay to be queried with msgHash=$expectedMsgHash, " +
-                "but lookups were ${sessionApi.completionLookups}",
-        )
+        sessionApi.completionLookups shouldContain expectedMsgHash
     }
 
     @Test
@@ -176,13 +172,9 @@ class DklsFamilyRelayRecoveryTest {
                 encryption = mockk(relaxed = true),
             )
 
-        assertFailsWith<RuntimeException> { keysign.keysignWithRetry() }
+        shouldThrow<RuntimeException> { keysign.keysignWithRetry() }
 
-        assertTrue(
-            expectedMsgHash in sessionApi.completionLookups,
-            "expected the relay to be queried with msgHash=$expectedMsgHash, " +
-                "but lookups were ${sessionApi.completionLookups}",
-        )
+        sessionApi.completionLookups shouldContain expectedMsgHash
     }
 
     @Test
@@ -201,12 +193,8 @@ class DklsFamilyRelayRecoveryTest {
                 encryption = mockk(relaxed = true),
             )
 
-        assertFailsWith<RuntimeException> { keysign.keysignWithRetry() }
+        shouldThrow<RuntimeException> { keysign.keysignWithRetry() }
 
-        assertTrue(
-            expectedMsgHash in sessionApi.completionLookups,
-            "expected the relay to be queried with msgHash=$expectedMsgHash, " +
-                "but lookups were ${sessionApi.completionLookups}",
-        )
+        sessionApi.completionLookups shouldContain expectedMsgHash
     }
 }
--- a/app/src/test/java/com/vultisig/wallet/data/keygen/DklsFamilyRelayRecoveryTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/data/keygen/DklsFamilyRelayRecoveryTest.kt
@@ -1,0 +1,212 @@
+package com.vultisig.wallet.data.keygen
+
+import com.vultisig.wallet.data.api.SessionApi
+import com.vultisig.wallet.data.common.md5
+import com.vultisig.wallet.data.mediator.Message
+import com.vultisig.wallet.data.models.KeyShare
+import com.vultisig.wallet.data.models.Vault
+import io.mockk.mockk
+import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Test
+
+/**
+ * Regression tests for issue #5314: when a DKLS-family signing attempt fails or times out (e.g. a
+ * relay message is lost to a transport error), the device must consult the relay for an
+ * already-completed signature via `KeysignVerify.checkKeysignComplete` before giving up — mirroring
+ * the legacy GG20 fallback in `KeysignViewModel.signMessageWithRetry`. Without this fallback a
+ * device whose local session never reaches `isFinished` fails with "signatures empty" even though
+ * the peer already posted the full signature to the relay.
+ *
+ * These tests pin the *wiring*: on failure the relay IS queried, and with the correct `msgHash` key
+ * (matching each class's `markLocalPartyKeysignComplete(msgHash, …)` write) — not the raw
+ * `messageToSign`. They deliberately do not assert the recovered signature is stored, because
+ * `tss.KeysignResponse` is a JNI/native type whose static initializer loads the native `gojni`
+ * library and therefore cannot be constructed — or even mocked — on the host JVM. This is the same
+ * constraint that makes `SessionApiBodyReadTest` skip `checkKeysignComplete`, and the reason the
+ * [FakeSessionApi] below is hand-written rather than a mockk (mockk eagerly resolves the
+ * `tss.KeysignResponse` return type and crashes with UnsatisfiedLinkError).
+ */
+class DklsFamilyRelayRecoveryTest {
+
+    private val message = "abc123deadbeef"
+    private val expectedMsgHash = message.md5()
+
+    private fun vault() =
+        Vault(
+            id = "vault-id",
+            name = "test-vault",
+            pubKeyECDSA = "pub-ecdsa",
+            pubKeyEDDSA = "pub-eddsa",
+            pubKeyMLDSA = "pub-mldsa",
+            localPartyID = "deviceA",
+            keyshares = listOf(KeyShare("pub-ecdsa", "share")),
+        )
+
+    /**
+     * A [SessionApi] whose setup-message fetch always fails (simulating the failed/lost round that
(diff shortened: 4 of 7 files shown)


app/src/test/java/com/vultisig/wallet/data/keygen/DklsFamilyRelayRecoveryTest.kt
```diff
@@ -0,0 +1,212 @@
+package com.vultisig.wallet.data.keygen
+
+import com.vultisig.wallet.data.api.SessionApi
+import com.vultisig.wallet.data.common.md5
+import com.vultisig.wallet.data.mediator.Message
+import com.vultisig.wallet.data.models.KeyShare
+import com.vultisig.wallet.data.models.Vault
+import io.mockk.mockk
+import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
```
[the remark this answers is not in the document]
aminsato: Done in 8e7fa7ae9 — swapped `assertFailsWith`/`assertTrue` for Kotest `shouldThrow`/`shouldContain` across all three test methods.

johnnyluo:
Please provide on chain tx link

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
```diff
@@ -414,6 +414,11 @@ class DKLSKeysign(
             throw e
         } catch (e: Exception) {
             println("Failed to sign message ($messageToSign), error: ${e.localizedMessage}")
+            val keySignVerify = KeysignVerify(mediatorURL, sessionID, sessionApi)
```
rkokhatskyi, line 417: `markLocalPartyKeysignComplete` in `KeysignVerify.kt:10` swallows failure with zero retry while this GET is auto retried by the shared Ktor client, so if the completing peer's own report POST drops under the same transport blip this fix targets, the relay never learns and this call still returns null, hitting signatures empty anyway

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
```diff
@@ -414,6 +414,11 @@ class DKLSKeysign(
...
+            keySignVerify.checkKeysignComplete(msgHash)?.let {
```
rkokhatskyi, line 418: This GET rides `HttpClientConfigurator.kt:68`'s shared retry stack, up to 3 attempts with backoff toward a 60 second cap, and fires on every caught exception, so under the exact degraded relay conditions this fix targets it can add close to a minute per attempt before falling through to retry, needs runtime confirmation

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
```diff
@@ -414,6 +414,11 @@ class DKLSKeysign(
...
+                signatures[messageToSign] = it
```
rkokhatskyi, line 419: `SwapKitBtcSigner.kt`, which signs BTC and LTC SwapKit deposits through this class, splices `derSignatureBySighash` straight into the witness with no verify call anywhere in the file, unlike `UtxoHelper.getSignedTransaction` which its sibling signers reuse, so a relay sourced signature here has no cryptographic gate before broadcast

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
```diff
@@ -414,6 +414,11 @@ class DKLSKeysign(
...
+                return
```
rkokhatskyi, line 427: This early return skips the `waitingNotified = false` reset at `DklsKeysign.kt:239`, which `pullInboundMessages` performs on normal success, so a relay recovered signature leaves `signingState` pinned on `WaitingForPeer` through broadcast, same gap in Schnorr and Mldsa

app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
```diff
@@ -417,6 +417,11 @@ class SchnorrKeysign(
             throw e
         } catch (e: Exception) {
             println("Failed to sign message ($messageToSign), error: ${e.localizedMessage}")
+            val keySignVerify = KeysignVerify(mediatorURL, sessionID, sessionApi)
```
rkokhatskyi, line 420: Unlike `DklsKeysign.kt:344` and `MldsaKeysign.kt:148`, both gated on attempt == 0, this file's initiator branch at line 346 re uploads a fresh setup message on every retry, which can desync a peer still mid protocol on the old one

app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
```diff
@@ -417,6 +417,11 @@ class SchnorrKeysign(
...
+            keySignVerify.checkKeysignComplete(msgHash)?.let {
+                signatures[messageToSign] = it
```
rkokhatskyi, line 422: `TonHelper.kt:226` builds a `PublicKey` from the vault key but never calls `.verify()` on it before broadcast, unlike every sibling EdDSA helper, so a signature this fallback recovers from the relay reaches that same sink with no local cryptographic check

app/src/main/java/com/vultisig/wallet/data/keygen/MldsaKeysign.kt
```diff
@@ -122,6 +122,11 @@ class MldsaKeysign(
             throw e
         } catch (e: Exception) {
             Timber.e(e, "Failed to sign message (%s)", messageToSign)
+            val keySignVerify = KeysignVerify(mediatorURL, sessionID, sessionApi)
```
rkokhatskyi, line 125: This five line block is now byte identical in `DklsKeysign.kt` and `SchnorrKeysign.kt` too, and a near duplicate of it landed in the unrelated PR #5315, worth extracting into one shared helper over mediatorURL, sessionID, sessionApi, msgHash and signatures so the three copies cannot silently diverge

app/src/main/java/com/vultisig/wallet/data/keygen/MldsaKeysign.kt
```diff
@@ -122,6 +122,11 @@ class MldsaKeysign(
...
+            keySignVerify.checkKeysignComplete(msgHash)?.let {
```
rkokhatskyi, line 126: `QBTCTransactionHelper.kt:35` splices `sig.derSignature` into the broadcast tx with no verify call, and the vendored ML DSA native library exports no verify function at all, so a signature recovered here has no cryptographic gate unlike roughly fifteen other chain helpers

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
aminsato: Good catch — the write everything recovers from was the least protected call. Fixed in 07877735e.

The completion POST (`markLocalPartyKeysignComplete`) is non-idempotent, so `HttpRequestRetry` skips it (only GET/HEAD/OPTIONS are retried), and `KeysignVerify` was swallowing its failure. So under the exact blip this branch targets, the peer's signature POST could drop, the relay never stores it, and every stuck peer's recovery GET returns null → "signatures empty" anyway.

Since reposting the same signature under the same `messageId` is idempotent at the relay, the write now retries itself (3 attempts, 1s backoff) before giving up. This also hardens the GG20 path, which shares this method.

This shrinks the window rather than eliminating it — if a finishing peer dies or loses connectivity for the whole retry window, the relay still never learns. Fully closing that would need any finishing party to be able to mark the ceremony complete, which is a larger change worth a separate issue.

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
aminsato: Confirmed and fixed in da53bb5e2. `SwapKitBtcSigner.getSignedTransaction` now runs the same gate `UtxoHelper.getSignedTransaction` does (`UtxoHelper.kt:359-368`): each DER signature is `verifyAsDER`-checked against the derived vault pubkey over its sighash, and a mismatch throws before anything reaches the witness.

I put the check in the public `getSignedTransaction` (already a WalletCore/JNI boundary) rather than in `compileSignedTransaction`, so the pure byte-for-byte serialization tests that feed a fake DER stay headless.

Worth noting for scope: this was a pre-existing gap, not introduced here — but relay recovery does widen it, since the signature can now come from the relay instead of being locally computed, which is exactly why the gate belongs here. It is defense-in-depth rather than an anti-theft fix: the relay holds no key share, so it cannot forge a valid signature; the worst it can return is a mismatched one, which without this gate would have been spliced into a broadcast the network then rejects. Now it fails locally with a clear error.

app/src/main/java/com/vultisig/wallet/data/keygen/MldsaKeysign.kt
aminsato: Confirmed on both counts: `QBTCTransactionHelper.getSignedTransaction` splices `sig.derSignature` with no gate, and the vendored ML-DSA native library exports no signature-verify entry point (`verifySetupMessage` in `MldsaKeysign` checks the setup-message transcript hash, not the signature), so unlike SwapKit BTC this one cannot be verified on-device today — there is no `verifyAsDER` equivalent to call.

The bounded-risk argument from the SwapKit thread still holds and is what keeps this safe in the meantime: the relay has no key share and cannot forge a valid ML-DSA-44 signature, so the only thing it can return is a mismatched one, which the QBTC chain rejects at broadcast rather than accepting. So this is a missing fail-fast/defense-in-depth gate, not a fund-loss path.

Closing it properly requires the ML-DSA AAR to export a verify function; I will file a follow-up issue to add on-device verification once that lands, rather than pull the native-lib change into this PR.

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
aminsato: Good flag, and the mechanics check out. `checkKeysignComplete` is a GET, so it rides the shared `HttpRequestRetry` stack (`HttpClientConfigurator.kt:63`, 3 attempts, `exponentialDelay()`), and there is no `HttpTimeout` installed on that client — only OkHttp's 15s connect timeout. So against a hanging relay a single recovery lookup can stall for tens of seconds, and because it fires on every caught exception inside `keysignOneMessageWithRetry` (which itself retries up to 3x), the worst-case cost compounds across attempts.

The mitigating factor is that this only runs on the already-failing/timed-out path, so it trades latency on a ceremony that was going to fail anyway for a chance to recover a completed signature — but the compounding is real and the "close to a minute per attempt" estimate is in the right ballpark; it wants a real timing run under a degraded relay to pin the exact number, which I have not done yet.

I left this one as-is for now since it is a latency/UX tradeoff on the failure path rather than a correctness issue, but if you would prefer I can bound it — either a tight `HttpTimeout` scoped to the recovery GET, or moving the check to fire once after retries are exhausted instead of on every attempt. Happy to do either; let me know which you prefer.

app/src/main/java/com/vultisig/wallet/data/keygen/DklsKeysign.kt
aminsato: Confirmed and fixed in 6154346b4 across all three classes. `signWithKeyType` only clears `WaitingForPeer` back to the active state via `onPeersResumed` (`KeysignViewModel.kt:561-566`), and the recovery early-return skipped it, so the screen stayed on "waiting for peer" through broadcast whenever recovery fired after the 10s silence threshold. The recovery branch now runs the same guarded reset `pullInboundMessages` does on resume — `if (waitingNotified) { waitingNotified = false; onPeersResumed?.invoke() }` — before returning.

app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
aminsato: Confirmed and fixed in 6154346b4. `TonHelper.getSignedTransaction` built `PublicKey(pubKeyData, ED25519)` and discarded it — the `verify()` gate every sibling EdDSA helper applies (Solana, Cosmos, Ripple, …) had been dropped. It now verifies the signature against the vault key over `preSigningOutput.data` and throws on mismatch before `compileWithSignatures`, same as the SwapKit gate — so a relay-recovered signature is checked cryptographically instead of trusted.

app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
aminsato: Confirmed: `SchnorrKeysign`'s initiator branch is `if (isInitiateDevice)` with no `&& attempt == 0`, unlike `DklsKeysign.kt:344` and `MldsaKeysign.obtainSetupMessage`, so each Schnorr retry regenerates and re-uploads a fresh setup message and can desync a peer still on the previous one.

This one is pre-existing — it is not introduced or touched by this PR (which only adds the relay-recovery lookup to the catch block). To keep this PR scoped to the recovery fix, I would rather address it in a dedicated PR with a keysign smoke test across a real retry, since it changes the live Schnorr protocol upload path. I will file a follow-up issue. Shout if you would prefer it folded in here instead.

app/src/main/java/com/vultisig/wallet/data/keygen/MldsaKeysign.kt
aminsato: Agree the block is now identical in all three classes, and with the two follow-up fixes it grew to ~10 lines each (recovery lookup + the WaitingForPeer reset from 6154346b4), so the duplication argument is stronger now, not weaker.

I am leaving the extraction out of this PR deliberately: a clean shared helper wants to cover the near-duplicate that landed in #5315 too, and doing that here would either depend on #5315 or force a later reconciliation. Better to extract once, after #5315 merges, over both call sites at the same time. I will file a follow-up issue to do that so the three (four) copies do not drift in the meantime.

data/src/main/kotlin/com/vultisig/wallet/data/api/KeysignVerify.kt
```diff
@@ -1,21 +1,39 @@
...
+                    Timber.tag("KeysignVerify")
+                        .d(
+                            "markLocalPartyKeysignComplete error (attempt %d): %s",
+                            attempt + 1,
+                            e.stackTraceToString(),
+                        )
+                    if (attempt < MAX_MARK_COMPLETE_RETRIES - 1) {
+                        delay(MARK_COMPLETE_BACKOFF_MS)
+                    }
+                }
```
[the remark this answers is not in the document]
aminsato: Not applying this one — rethrowing here would regress the completing device's success path.

`markLocalPartyKeysignComplete` is a best-effort publish that runs *after* the completing device already has its own signature, and its callers do not have a publisher-side recovery path that a throw would unlock:

- DKLS/Schnorr/MLDSA (`DklsKeysign.kt:410-411`): `markLocalPartyKeysignComplete(msgHash, resp)` is immediately followed by `signatures[messageToSign] = resp`. If the publish throws, that local store is skipped, the device drops into its own catch, and `checkKeysignComplete` then queries the relay for the very POST that just failed (→ null) — so it would throw away a valid signature it already computed and re-run the whole ceremony.
- GG20 (`KeysignViewModel.kt:749-750`): the local store happens *before* the publish, so a throw just forces an unnecessary re-sign of an already-completed message.

Rethrowing also does nothing for the stuck peer this fix targets: if the POST genuinely failed after 3 attempts, the signature is not on the relay whether or not the publisher throws. The retry loop already raises the odds the publish lands; swallowing after it is exhausted is the intended semantics (and `CancellationException` is still propagated). Leaving as-is.

aminsato:
> Please provide on chain tx link

https://explorer.zksync.io/tx/0x16e66d34ac74a01dd52f2b436958911f829618e31cad15ce3078940d7e8e9cd1

johnnyluo (approved):
LGTM

PR #5325, opened 2026-07-18
fix(send): re-clamp Max to current balance/fee at submit (#5316)

### LINKED ISSUE #5316: [Send] Max is a one-shot snapshot — never re-clamped on balance/fee refresh; max sends fail with 'insufficient balance' (user: must leave USDT behind)
## User report (2026-07-18)

> please fix/improve the sending max feature. it doesnt really works, even trying to send max non gas tokens. i tried to send all usdt and i need to leave some usdt in the wallet in order to work. happens alot with swapping too

## What is NOT the bug (verified on main @ ebf956478)

The max-amount math for non-gas tokens is exact end-to-end:

- `GetAvailableTokenBalanceUseCase` returns the **full raw balance** for non-native tokens — no fee is subtracted.
- `AmountFractionManager.fetchPercentageOfAvailableBalance` scales at full token precision (`setScale(token.decimal, DOWN)`).
- Submit validation is strict `<`, so `amount == balance` passes ([DefaultSendStrategy.kt#L339](https://github.com/vultisig/vultisig-android/blob/ebf956478/app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt#L339)).
- The text-field round-trip is exact `BigDecimal`, locale-independent.

So a max USDT send can only fail when the balance/fee captured **at tap time** no longer matches the one used **at submit time**. Two such races exist, and nothing re-clamps the field in between.

## Race 1 — token balance: Max is a one-shot stale snapshot

`chooseMaxTokenAmount()` ([AmountFractionManager.kt#L73](https://github.com/vultisig/vultisig-android/blob/ebf956478/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt#L73)) reads the balance once and writes the field via `markMax` ([AmountManager.kt#L59](https://github.com/vultisig/vultisig-android/blob/ebf956478/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountManager.kt#L59)). `AccountsLoader.loadAddresses` first emits the **cached** balance, then re-emits after network hydration. If the balance corrects downward (stale cache, pending spend from another device), the field keeps the old larger number — `maxAmount` is only reset on token switch, never on balance refresh. At submit, `DefaultSendStrategy` re-reads the *current* `selectedAccount.tokenValue` and `selectedTokenValue.value < tokenAmountInt` throws.

**Scenario:** cached USDT balance `123.456789` → Max fills `123.456789` → hydration corrects balance to `123.400000` → Continue → "insufficient". User types `123.4` → works. That is exactly "I need to leave some USDT in the wallet in order to work."

## Race 2 — native coins: EVM Max reuses the cached gas fee

[AmountFractionManager.kt#L200](https://github.com/vultisig/vultisig-android/blob/ebf956478/app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt#L200): for EVM native coins with a cached `gasFee.value`, Max shortcuts to `balance − cachedFee` and skips the fresh fee round-trip that non-EVM native coins get. If gas rises before Continue, submit's `availableTokenBalance = balance − freshFee` drops below the filled amount → insufficient.

**Scenario:** ETH balance `1.0`, cached fee `0.0010` → Max = `0.9990`. Gas spikes, fresh fee `0.0015` → available `0.9985 < 0.9990` → fails.

## Aggravating factor — misleading error string

The token-exceeds-token-balance failure ([DefaultSendStrategy.kt#L339-L346](https://github.com/vultisig/vultisig-android/blob/ebf956478/app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt#L339-L346)) throws `send_error_insufficient_native_balance_with_fees`. The "with fees" phrasing teaches users the wrong mental model — that they must reserve *tokens* for fees — which is precisely how the user described the bug.

## Suggested fix

When `isMaxAmount` is set, re-derive the amount from the **current** balance and **current** fee at submit (or re-clamp the field whenever the balance/fee flows emit). One change covers both races. Also give the token-balance case its own error string without "with fees".

For reference: iOS re-fetches balance+fee and recomputes max on the Verify screen; Windows re-caps native max at keysign build (`refineKeysignAmount`). Android is currently the only platform with no submit-time re-clamp.

aminsato:
Fixes #5316

## Problem

"Max" on the Send screen was a **one-shot snapshot** — it captured the balance and gas fee at tap time, wrote them into the amount field, and never re-clamped when either moved before submit. Nothing re-derives the field between tap and Continue, so two races made a max send fail with "insufficient balance":

1. **Non-native tokens (e.g. USDT):** a stale-high *cached* balance fills the field, then network hydration corrects the balance downward. The field keeps the old larger number, so submit's strict balance check rejects it. This is exactly the reported *"I need to leave some USDT in the wallet in order to work."*
2. **EVM native coins:** Max reuses the *cached* gas fee (`balance − cachedFee`). If gas rises before Continue, submit's `balance − freshFee` drops below the filled amount → rejected.

An aggravating factor: the non-native token-balance shortfall surfaced the `..._native_balance_with_fees` error string, whose "with fees" framing wrongly teaches users to reserve *tokens* for gas — precisely the wrong mental model the user described.

## Fix

- When the amount equals the Max snapshot on a **standard send**, re-derive it from the **current** balance and fee at submit via `getAvailableTokenBalance`, clamping **down only** so we never send more than the user saw. One change covers both races. DeFi flows (staking / unbond / frozen TRX) are excluded — they carry different balance semantics and compute their own max.
- Route the non-native token-balance shortfall to a new `send_error_insufficient_token_balance` string (added to all 10 locales) — same visible wording, but without the misleading "with fees" key.

For reference, iOS re-fetches balance+fee and recomputes max on the Verify screen; Windows re-caps native max at keysign build (`refineKeysignAmount`). Android was the only platform with no submit-time re-clamp.

## Test plan

- `./gradlew :app:compileDebugKotlin` — passes
- Added 3 regression tests to `DefaultSendStrategyTest`:
  - non-native Max re-clamps down when the balance corrects downward (Race 1)
  - native EVM Max re-clamps down when gas rises (Race 2)
  - non-native over-entry surfaces the new token-balance error string
- Full class: **11 tests, 0 failures** (1 pre-existing XRP JNI test skipped in the JVM env)

## Summary by CodeRabbit

- **Bug Fixes**
  - “Max” send amounts are now rechecked against the latest available balance, including updated gas fees, before submission.
  - Improved validation for insufficient token balances with clearer, token-specific error messages.
  - Added localized messages for supported languages.

- **Tests**
  - Added coverage for stale “Max” amounts, increased gas costs, and token over-entry scenarios.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
@@ -176,11 +176,33 @@ internal class DefaultSendStrategy(
                         )
                     }
 
-                    val tokenAmountInt =
-                        tokenAmount.movePointRight(selectedToken.decimal).toBigInteger()
-
                     val srcAddress = selectedToken.address
                     val isMaxAmount = tokenAmount.compareTo(amountManager.currentMaxAmount) == 0
+
+                    // "Max" is a one-shot snapshot: it captures the balance and gas fee at tap
+                    // time and never re-clamps if either moves before submit (a cached balance
+                    // correcting downward after network hydration, or EVM gas rising). A stale-high
+                    // snapshot then exceeds what's actually spendable and submit validation rejects
+                    // it as "insufficient" — which is why a max send can force the user to leave
+                    // some tokens behind. Re-derive the amount from the CURRENT balance and fee,
+                    // clamping DOWN only so we never send more than the user saw. Standard sends
+                    // only: DeFi flows (staking/unbond/frozen TRX) carry different balance
+                    // semantics and compute their own max.
+                    val tokenAmountInt =
+                        tokenAmount.movePointRight(selectedToken.decimal).toBigInteger().let {
+                            entered ->
+                            if (isMaxAmount && defiTypeProvider() == null) {
+                                val available =
+                                    getAvailableTokenBalance(selectedAccount, gasFee.value)?.value
+                                if (available != null && available > BigInteger.ZERO) {
+                                    entered.coerceAtMost(available)
+                                } else {
+                                    entered
+                                }
+                            } else {
+                                entered
+                            }
+                        }
 
                     if (chain == Chain.Tron) {
                         val isTronStakingOp =
@@ -337,9 +359,12 @@ internal class DefaultSendStrategy(
                                 )
 
                         if (selectedTokenValue.value < tokenAmountInt) {
+                            // Token gas is paid in the native coin, not this token, so this is a
+                            // pure token-balance shortfall — keep the message free of any "with
+                            // fees" framing that would wrongly suggest reserving tokens for gas.
                             throw InvalidTransactionDataException(
                                 UiText.FormattedText(
-                                    R.string.send_error_insufficient_native_balance_with_fees,
+                                    R.string.send_error_insufficient_token_balance,
                                     listOf(selectedToken.ticker),
                                 )
                             )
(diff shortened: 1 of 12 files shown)


johnnyluo (approved):
LGTM

PR #5331, opened 2026-07-19
fix(swap): hide MAX whenever the source token is native (#5317)

### LINKED ISSUE #5317: [Swap] Native max fee race — reservedNetworkFee captured at tap; fresh quote fee pushes amount+fee over balance
## Problem

Native-coin Max in the swap form reserves only the fee estimate captured at tap time. Part of the user report "happens alot with swapping too" (see the send-side issue for the full report).

[SwapFormViewModel.kt#L489-L537](https://github.com/vultisig/vultisig-android/blob/ebf956478/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt#L489-L537): `reservedNetworkFee = estimatedNetworkFeeTokenValue.value` at tap time; `maxUsableTokenAmount = srcTokenValue.value - reservedNetworkFee` (L510). Tapping Max then triggers an immediate quote refetch (`markImmediateFetch()`), which recomputes the fee. Validation (`SwapValidator.kt#L81-L87`) checks `srcAmountValue + feeValue > balance` against the **new** fee — if it came back higher, the just-filled Max amount fails with `swap_error_insufficient_balance_and_fees`.

**Scenario:** ETH balance `1.0`, fee at tap `0.0010` → Max = `0.9990`. Fresh quote's fee `0.0015` → `0.9990 + 0.0015 > 1.0` → error, even though the user just tapped Max.

The Solana ATA-rent reserve for SOL→SOL (L498-506) is correct and deliberately native-only — not part of this bug.

## Suggested fix

Same principle as the send-side fix: when the max flag is set, re-clamp `srcAmount` to `balance − currentFee` whenever the fee updates (or add a small headroom buffer to the reserved fee for native max).

aminsato:
### Transaction hash
N/A — UI visibility change only, no transaction is built or signed differently.

Fixes #5317. A native source token pays the swap gas out of the same balance being swapped, so a **Max** amount is never fully spendable: the fee estimate lands, `amount + fee > balance`, and **Swap** goes disabled with nothing on screen explaining why.

The Max gate only hid the shortcut when *both* sides were native, so the common case — ETH → USDC, exactly the one in #5317 — still offered it.

```kotlin
- enableMaxAmount = (isSrcNative && isDstNative).not()
+ enableMaxAmount = isSrcNative.not()
```

Max is now hidden whenever the source is native, and still shown for token sources (USDC → ETH), where gas comes out of a separate native balance and the full token balance really is spendable.

## Note on the change of approach

This replaces the earlier approach on this branch, which kept native Max working by re-clamping `balance − fee` whenever the fee estimate landed late. Per review the intended behaviour is simply *no Max for a native source*, so that work is dropped in favour of the one-line gate change.

## Test plan

`SwapFormViewModelTest`:

- `collectSelectedAccounts sets enableMaxAmount false when both native` — existing, unchanged.
- `collectSelectedAccounts sets enableMaxAmount false when only the source is native` — new, covers ETH → USDC (the #5317 case).
- `collectSelectedAccounts sets enableMaxAmount true when the source is a token` — new, covers USDC → ETH.

`./gradlew :app:testDebugUnitTest --tests "*SwapFormViewModelTest"` passes; ktfmt clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated swap form behavior so the **MAX** option is hidden when the source asset is the network’s native token.
  * MAX remains unavailable until a source token is selected.
  * MAX is available when swapping from a non-native token to the native token.


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -529,13 +536,102 @@ constructor(
...
+    /**
+     * Re-clamps an in-flight native MAX to the still-current balance − fee whenever the reserved
+     * fee changes (gas landing after the tap, or a higher fee from the fresh quote), so the filled
+     * Max amount can't sit above what's spendable and trip
+     * `swap_error_insufficient_balance_and_fees` (#5317). No-ops once the field no longer holds our
+     * filled value (a manual edit) or the source token changed — either ends the session.
+     */
+    private fun observeMaxFeeReclamp() {
+        viewModelScope.launch {
+            quotePipeline.estimatedNetworkFeeTokenValue.collect { reclampNativeMaxAmount() }
```
rkokhatskyi: observeMaxFeeReclamp only listens to estimatedNetworkFeeTokenValue, not selectedSrc, so a balance correction alone (the cached-then-live race #5325 already fixed on send) leaves the MAX amount unreclamped, since calculateGasFee builds its fee tx with amount=BigInteger.ZERO (SwapGasCalculator.kt:109), reproducing the exact #5317 disabled-swap symptom this fixes for fee changes only.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -529,13 +536,102 @@ constructor(
...
+            return BigInteger.ZERO
+        }
+        val baseFee = quotePipeline.estimatedNetworkFeeTokenValue.value?.value ?: BigInteger.ZERO
+        // A Jupiter affiliate swap may create the destination-mint fee ATA on first use, whose
+        // ~0.002 SOL rent the payer (this wallet) funds but the network-fee estimate doesn't
+        // include. Reserve it on native-SOL MAX so the first such swap can't underfund and fail on
+        // submit; it leaves harmless dust when no ATA is created. Jupiter only handles same-chain
+        // Solana, so a SOL→non-Solana MAX (routed via THORChain/LI.FI) never creates a fee ATA —
+        // don't reserve dust for it.
+        val dstChain = selectedDst.value?.account?.token?.chain
```
rkokhatskyi: nativeMaxReservedFee reads selectedDst live to add the Solana ATA rent, but nothing re-runs it on a destination-only switch, since observeMaxFeeReclamp only fires on fee changes and a Solana source never rewrites estimatedNetworkFeeTokenValue from a dst change alone (SwapQuotePipeline.kt:509). MAX with SOL, then switch destination to Solana, and the reserve never gets added.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -529,13 +536,102 @@ constructor(
...
+     * The source-chain network fee reserved on a native MAX: the current estimate, plus the Solana
+     * fee-ATA rent reserve for a SOL→SOL swap. Zero unless the source is native on its own gas
+     * chain (so a stale cross-chain fee is never subtracted). Shared by the initial MAX fill and
+     * the fee-change re-clamp so both reserve identically (#5317).
+     */
+    private fun nativeMaxReservedFee(srcToken: Coin): BigInteger {
+        if (!srcToken.isNativeToken || quotePipeline.gasFeeChain.value != srcToken.chain) {
+            return BigInteger.ZERO
+        }
+        val baseFee = quotePipeline.estimatedNetworkFeeTokenValue.value?.value ?: BigInteger.ZERO
```
rkokhatskyi: nativeMaxReservedFee defaults a null fee to BigInteger.ZERO, so when a UTXO InsufficientUtxos result clears estimatedNetworkFeeTokenValue to null, the reclamp jumps the field up to the full unreserved balance and refetches, which hits InsufficientUtxos again, permanently stranding MAX at the wrong amount.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -529,13 +536,102 @@ constructor(
...
+        val srcTokenValue = account.tokenValue ?: return
+        val srcToken = account.token
+        // A manual edit (which changes the text) or a source-token switch ends the session: only
+        // re-clamp while the field still holds the exact amount we filled for this token.
+        if (srcToken.id != maxSessionSrcId || srcAmountState.text.toString() != maxAmountText) {
+            endMaxSession()
+            return
+        }
+
+        val maxUsable = srcTokenValue.value - nativeMaxReservedFee(srcToken)
```
rkokhatskyi: reclampNativeMaxAmount overwrites the field whenever the derived amount differs from what's shown, with no floor against the current value, so a fee estimate that later drops silently grows the filled amount upward, unlike the sibling send-side fix's explicit clamp-down-only re-derivation.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -125,6 +125,18 @@ constructor(
 
     private var selectTokensJob: Job? = null
 
+    // A native-source MAX reserves the source-chain fee off the top, but that fee is only a
+    // snapshot
+    // of whatever estimate had landed at tap time — often none yet (gas still loading) or a later,
+    // higher one. These track an in-flight MAX so a fee change can re-clamp the filled amount to
+    // the
+    // still-current balance − fee instead of leaving the stale snapshot to trip the
+    // insufficient-balance validation once a higher fee lands (#5317). Cleared the moment the field
```
rkokhatskyi: This comment should say the session ends lazily, not "the moment" it happens: the manual-edit/token-switch check only runs inside reclampNativeMaxAmount, itself gated on the next estimatedNetworkFeeTokenValue emission, so a stale session can persist indefinitely if no fee update ever arrives.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in 37a1f41. The collector now combines \`selectedSrc\` + \`selectedDst\` + \`estimatedNetworkFeeTokenValue\`, so a balance-only correction re-clamps even though \`calculateGasFee\` estimates with amount=0 and never re-emits a fee for a balance change. Added a `balance corrects downward` test.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in 37a1f41. \`selectedDst\` is now one of the combined triggers, so a SOL→SOL destination switch re-runs `nativeMaxReservedFee` and clamps the amount down by the fee-ATA rent.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in 37a1f41. The re-clamp is now clamp-DOWN only: when the fee is cleared to null, `maxUsable` becomes the full balance (≥ the currently filled amount) so it is skipped rather than raising the field and re-fetching into `InsufficientUtxos` again.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in 37a1f41. Clamp-DOWN only now — it rewrites solely when `maxUsable < currentRaw` (the currently filled amount), so a later fee drop never grows what the user saw, matching the send-side re-derivation. Added an `balance corrects upward` test asserting no growth.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Fixed in 37a1f41. Reworded: a flip or a non-MAX/exhausted tap tears the session down eagerly, but a manual edit / source-token switch is detected lazily on the next balance/fee/destination update.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -529,13 +536,102 @@ constructor(
...
+     * fee changes (gas landing after the tap, or a higher fee from the fresh quote), so the filled
+     * Max amount can't sit above what's spendable and trip
+     * `swap_error_insufficient_balance_and_fees` (#5317). No-ops once the field no longer holds our
+     * filled value (a manual edit) or the source token changed — either ends the session.
+     */
+    private fun observeMaxFeeReclamp() {
+        viewModelScope.launch {
+            quotePipeline.estimatedNetworkFeeTokenValue.collect { reclampNativeMaxAmount() }
+        }
+    }
```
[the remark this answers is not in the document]
aminsato: Added `.catch { Timber.e(it) }` on the collector in 37a1f41. Also consolidated the three session flags into a single nullable `MaxSession` per the other suggestion.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
rkokhatskyi: Fixed. selectedDst is now part of the combine trigger, so a destination switch re-derives the ATA reserve through nativeMaxReservedFee. No dedicated test for this specific path, but the mechanism traces correctly both directions.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
rkokhatskyi: Fixed as a side effect of the new clamp-down-only guard (maxUsable >= currentRaw returns early), so a fee clearing to null on InsufficientUtxos can no longer jump the field up to the full balance.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
rkokhatskyi: Fixed. The new down-only guard blocks any growth in the derived amount, confirmed against the new "does not grow" test and by tracing the rounding direction of both conversions.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -529,13 +535,108 @@ constructor(
...
+     * three inputs, not fees alone, because `calculateGasFee` builds its estimate with amount = 0
+     * ([SwapGasCalculator]) so a balance-only change never re-emits a fee.
+     */
+    private fun observeMaxFeeReclamp() {
+        viewModelScope.launch {
+            combine(selectedSrc, selectedDst, quotePipeline.estimatedNetworkFeeTokenValue) { _, _, _
+                    ->
+                    Unit
+                }
+                .catch { Timber.e(it) }
```
rkokhatskyi: This catch sits between combine and collect, so per Flow's exception transparency it never catches a throw from reclampNativeMaxAmount inside the collect block, unlike calculateGas's own inner try/catch a few lines above. Move the call into onEach before catch.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Good catch — Flow exception transparency, the `.catch` sat upstream of the collector. Fixed in dc6546c62: moved the guard into an inner try/catch inside `collect { }` (rethrowing `CancellationException`), mirroring calculateGas.

aminsato:
@coderabbitai on the two test nitpicks (dc6546c62):

- **Quote refresh coverage** — added. The two balance-correction tests now assert `markImmediateFetch` (called only by the MAX tap and the re-clamp): **2** calls after a downward correction (tap + re-clamp) and **1** after an upward one (tap only, no re-clamp), so a regression that drops the re-clamp's refresh is caught. I verify `markImmediateFetch` rather than adding a routable dst + full quote round-trip because the re-clamp writes `srcAmountState` from inside a coroutine, which the test harness can't pair with a Compose snapshot notification deterministically — `markImmediateFetch` is the exact debounce-bypass signal the pipeline keys off, so it's the robust unit-level assertion.
- **Kotest assertions** — skipping: this whole file uses `kotlin.test` (`assertEquals`/`assertTrue`/`assertNull`) throughout; the new tests match the surrounding convention. Switching 5 tests to Kotest in an otherwise `kotlin.test` file would be inconsistent.

johnnyluo (dismissed):
So we decided a while ago that when the source token is native token , we don't have `Max` button , we don't support it , because the calculation is rather complicated

so I am not sure what this is trying to achieve , can you please explain? 

If `Max` button show up when it is native source token is selected , then please fixed that

`Max` button for non-native token , we just use the whole balance , because gas is always paid in native tokens

aminsato:
Good question — but the MAX button does show for a native source right now. The only place it's gated is `SwapTokenSelector.kt:109`:

```kotlin
enableMaxAmount = (isSrcNative && isDstNative).not()
```

So it's hidden only when *both* sides are native (e.g. ETH → BTC). Native → token still shows it. ETH → USDC is the obvious one, and that's exactly the case in #5317. Here it is live:

image

So the "no MAX for native source" decision is only half in place today — it kicks in when the destination is also native, but not otherwise.

Two ways to go, it's really your call:

1. Keep native-source MAX working (this PR): reserve the source-chain gas and re-clamp `balance − fee` when the fee estimate lands late, so tapping MAX doesn't end up with `amount + fee > balance` and disable Swap.
2. If we want to stick with "no MAX when the source is native", the real fix is one line — change the gate to `isSrcNative.not()` so it's hidden whenever the source is native. Then this PR isn't needed.

Which do you prefer? If it's option 2 I'll close this and put up the one-liner instead.

johnnyluo:
> Good question — but the MAX button does show for a native source right now. The only place it's gated is `SwapTokenSelector.kt:109`:
> 
> ```kotlin
> enableMaxAmount = (isSrcNative && isDstNative).not()
> ```
> 
> So it's hidden only when _both_ sides are native (e.g. ETH → BTC). Native → token still shows it. ETH → USDC is the obvious one, and that's exactly the case in #5317. Here it is live:
> 
> image
> So the "no MAX for native source" decision is only half in place today — it kicks in when the destination is also native, but not otherwise.
> 
> Two ways to go, it's really your call:
> 
> 1. Keep native-source MAX working (this PR): reserve the source-chain gas and re-clamp `balance − fee` when the fee estimate lands late, so tapping MAX doesn't end up with `amount + fee > balance` and disable Swap.
> 2. If we want to stick with "no MAX when the source is native", the real fix is one line — change the gate to `isSrcNative.not()` so it's hidden whenever the source is native. Then this PR isn't needed.
> 
> Which do you prefer? If it's option 2 I'll close this and put up the one-liner instead.

Can you please make sure when it is native token , then Max button doesn't show

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTokenSelector.kt
```diff
@@ -100,13 +100,15 @@ constructor(
...
+                    // amount is never actually spendable — hide the shortcut instead of letting it
+                    // fill an amount the fee then invalidates.
                     val isSrcNative = src?.account?.token?.isNativeToken ?: false
-                    val isDstNative = dst?.account?.token?.isNativeToken ?: false
                     uiState.update {
                         it.copy(
                             selectedSrcToken = srcUiModel,
                             selectedDstToken = dstUiModel,
-                            enableMaxAmount = (isSrcNative && isDstNative).not(),
+                            enableMaxAmount = isSrcNative.not(),
```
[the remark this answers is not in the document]
aminsato: Valid — fixed in 1aa200f7f. `enableMaxAmount` now comes from `src?.account?.token?.isNativeToken == false`, so a null source keeps MAX hidden instead of falling through to `true`. Covered by `collectSelectedAccounts keeps enableMaxAmount false when no source is selected`, which drives the selector directly with a null src (the ViewModel always resolves an account, so it cannot reach that state).

aminsato:
@johnnyluo @rkokhatskyi ready for another look — the branch is a different change than the one you reviewed.

The re-clamp approach is gone. Per @johnnyluo's call, the whole diff is now the Max gate plus tests:

```kotlin
- enableMaxAmount = (isSrcNative && isDstNative).not()
+ enableMaxAmount = src?.account?.token?.isNativeToken == false
```

Max is hidden whenever the source is native (the #5317 case, ETH → USDC), and also while no source is selected yet. It still shows for a token source, where gas comes out of a separate native balance.

@rkokhatskyi the Flow exception-transparency point you raised no longer applies — `SwapFormViewModel.kt` is untouched on this branch now.

1aa200f7f also addresses the null-source case CodeRabbit flagged. Four `enableMaxAmount` tests cover both-native, native-source-only, token-source, and no-source.

rkokhatskyi (dismissed):
Two inline comments below need a fix or an explicit call before merge: the isNativeToken re-derivation duplicating srcUiModel, and the now-dead native-source fee reservation in SwapFormViewModel.kt with a comment that still describes it as live.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTokenSelector.kt
```diff
@@ -100,13 +100,16 @@ constructor(
...
+                    // amount is never actually spendable — hide the shortcut instead of letting it
+                    // fill an amount the fee then invalidates. With no source selected yet there is
+                    // nothing to max out either, so MAX stays hidden until a token source arrives.
+                    val isSrcToken = src?.account?.token?.isNativeToken == false
                     uiState.update {
                         it.copy(
                             selectedSrcToken = srcUiModel,
                             selectedDstToken = dstUiModel,
-                            enableMaxAmount = (isSrcNative && isDstNative).not(),
+                            enableMaxAmount = isSrcToken,
```
rkokhatskyi, line 112: `enableMaxAmount` now stays false for a native source, so `selectSrcPercentage`'s reservation for `percentage>=1f && srcToken.isNativeToken` at `SwapFormViewModel.kt:505` can never fire from the MAX chip anymore, and its comment there still describes that branch as live

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTokenSelector.kt
```diff
@@ -100,13 +100,16 @@ constructor(
             combine(selectedSrc, selectedDst) { src, dst ->
                     val srcUiModel = src?.let { accountToTokenBalanceUiModelMapper(it) }
                     val dstUiModel = dst?.let { accountToTokenBalanceUiModelMapper(it) }
-                    val isSrcNative = src?.account?.token?.isNativeToken ?: false
-                    val isDstNative = dst?.account?.token?.isNativeToken ?: false
+                    // A native source pays the gas out of the same balance being swapped, so a MAX
+                    // amount is never actually spendable — hide the shortcut instead of letting it
+                    // fill an amount the fee then invalidates. With no source selected yet there is
+                    // nothing to max out either, so MAX stays hidden until a token source arrives.
+                    val isSrcToken = src?.account?.token?.isNativeToken == false
```
rkokhatskyi: `isSrcToken` walks `src.account.token.isNativeToken` again even though `srcUiModel` just above already carries the same `isNativeToken` field, so the two can drift apart if one is edited without the other. `srcUiModel?.isNativeToken == false` reads it directly instead

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTokenSelector.kt
rkokhatskyi: Fixed in fe42b7e31 — the comment now says this combination is no longer reachable from the picker and reads as a guard, not a live path.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTokenSelector.kt
rkokhatskyi: Fixed in fe42b7e31 — now reads srcUiModel?.isNativeToken, and it also caught a relaxed-mock gap that was masking this (mock now mirrors the real mapper).

rkokhatskyi (approved):
LGTM

PR #5332, opened 2026-07-19
fix(swap): keep SwapKit swap fee zeroed on Cardano, hide the row (#5321)

### LINKED ISSUE #5321: Cardano classified as TokenStandard.UTXO: SwapKit swap fee displayed as zero on ADA (sibling branch already excludes Cardano)
## Summary

`isSwapKitUtxoSwap` treats Cardano as a UTXO chain and zeroes the displayed swap fee for it. Its sibling branch four lines down already excludes Cardano; this one doesn't.

## Evidence

```kotlin
// SwapQuotePipeline.kt:389-395
val isSwapKitUtxoSwap = quote is SwapQuote.SwapKit && srcToken.chain.standard == TokenStandard.UTXO
```

No `&& srcToken.chain != Chain.Cardano`. The compensating plan-fee path at `:417-420`, which the comment above `:389` relies on to justify zeroing the display, **does** exclude Cardano.

Cardano SwapKit swaps are real and reachable — `SigningHelper.kt:342-344`, `:418-420`, `TX_TYPE_CARDANO_PREBUILT`.

Net effect: understated swap fee shown to the user on ADA SwapKit swaps. Display/accounting only — signing and fund movement are unaffected.

## Root cause: Cardano is classified as a UTXO chain

`Chain.kt:48` — `Cardano("Cardano", UTXO, "ADA/vbyte")`. Cardano is in `TokenStandard.UTXO` alongside Bitcoin, and even the fee-unit string is wrong (Cardano has no vbytes).

Cardano actually signs ed25519 + CBOR, not secp256k1 + PSBT:
- `Chain.kt:150` — `TssKeyType.EDDSA`
- `CardanoHelper.kt:217` — `PublicKey(spendingKeyData, PublicKeyType.ED25519)`
- `CardanoHelper.kt:248` — Conway CBOR envelope

Signing dispatch is safe because it switches on `when (chain)` and never consults the predicate (`SigningHelper.kt:341`, `:590-596`), and most predicate sites already carry explicit `chain != Chain.Cardano` guards (`GasFeeSelection.kt:22-23`, `GasFeeOrchestrator.kt:252`, `:300`, `DefaultSendStrategy.kt:283`, `:462`, `BlockChainSpecificRepository.kt:272-273`, `TransactionStatusProvider.kt:49-51`). The pattern is "UTXO means Cardano too, except everywhere it doesn't" — which is why one site was missed.

For comparison, iOS models this correctly: `ChainType` has a first-class `.Cardano` case separate from `.UTXO` (`ChainType.swift:10-22`), `UTXOChainsHelper.getHelper` returns `nil` for anything not `.UTXO` (`UTXOChainsHelper.swift:121-131`), and there is a regression test guarding against widening `.UTXO` to include Cardano.

## Proposed fix

1. Add `&& srcToken.chain != Chain.Cardano` at `SwapQuotePipeline.kt:389`, matching `:417-420`.
2. Fix the `"ADA/vbyte"` fee-unit string at `Chain.kt:48`.
3. Add defensive Cardano guards at two currently-unreachable sites before someone routes ADA through them: `Swaps.kt:22-28` (would parse a Cardano pre-image as a Bitcoin proto) and `SecurityScannerTransactionFactory.kt:35` (gated off today only by the Blockaid supported-chain list, `BlockaidScannerService.kt:106-119`).
4. Longer term, consider splitting `TokenStandard.UTXO` into a coin-selection/fee model and a signing family, so "UTXO-based" stops implying "secp256k1 + PSBT".

## Acceptance
- [ ] ADA SwapKit swaps display the correct fee
- [ ] Test covering a Cardano SwapKit quote
- [ ] `Chain.kt:48` fee unit corrected
- [ ] Defensive guards at the two latent sites

Found during a cross-repo signing-stack review.

aminsato:
## Summary

SwapKit's fee for UTXO-family sources (Bitcoin PSBT, Cardano CBOR) is the **source-chain deposit cost**, which is already surfaced on the **Network Fee** row. Counting it again as a Swap Fee double-counts the same on-chain cost.

The earlier approach carved Cardano out of the swap-fee zeroing (`isSwapKitUtxoSwap`), which reopened that double count and diverged from iOS. Verified against the running macOS app: a Cardano → ETH SwapKit swap shows **no Swap Fee row** — just Network Fee (the ADA deposit cost, shown once), the VULT discount, and Price Impact.

This PR:
- **Re-zeroes** the Cardano SwapKit swap fee (treats it like Bitcoin), so the ADA deposit cost is counted once as Network Fee — no double count.
- **Hides** the swap-fee breakdown row entirely when zeroed (empty `feeText`) instead of rendering a redundant `$0.00`, mirroring iOS `SwapCryptoLogic.fee` / `swapFeeString`. Applies to Bitcoin SwapKit too.
- Removes the now-unused `fiatValueToString` dependency from `SwapQuotePipeline`.
- Keeps the independent correctness fixes: Cardano `feeUnit` → `Lovelace`, and the Cardano guards in `Swaps.getPreSignedImageHash` / `SecurityScannerTransactionFactory`.

## iOS parity

`SwapCryptoLogic.fee` (iOS) for UTXO/Cardano SwapKit returns the transaction-plan fee (Network Fee) and discards the wire inbound fee; `inboundFeeDecimal` returns `.zero`, so the Swap Fee row is empty. This PR mirrors that behaviour.

## Test plan

- `SwapQuotePipelineSwapKitFeeTest` — Bitcoin and Cardano both assert the swap fee is zeroed and `feeText == ""` (row hidden).
- `SwapQuotePipelineNetworkFeeTest` — updated for the constructor change.
- Full `ui.models.swap.*` unit tests green.
- macOS app cross-check: Cardano → ETH SwapKit shows Network Fee only, no Swap Fee row.

Closes #5321

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected SwapKit swap-fee handling for UTXO-family quotes: swap-fee amount is zeroed and the fee breakdown row is hidden instead of showing “$0.00”.
  * Updated Cardano fee unit display to **Lovelace**.
  * Improved handling of Cardano in security-scanner and pre-signed image hash flows to prevent unsupported Bitcoin-style processing.
* **UI**
  * Only renders the expanded “estimated fees” row when the fee value is non-blank.
* **Tests**
  * Added/updated coverage for SwapKit fee and network-fee scenarios.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
@@ -16,7 +16,6 @@ import com.vultisig.wallet.data.repositories.SwapQuoteRepository
 import com.vultisig.wallet.data.usecases.ConvertTokenAndValueToTokenValueUseCase
 import com.vultisig.wallet.data.usecases.GetDiscountBpsUseCase
 import com.vultisig.wallet.data.usecases.getTierType
-import com.vultisig.wallet.ui.models.mappers.FiatValueToStringMapper
 import com.vultisig.wallet.ui.models.send.SendSrc
 import com.vultisig.wallet.ui.utils.UiText
 import java.math.BigDecimal
@@ -138,7 +137,6 @@ internal class SwapQuotePipeline(
     private val swapDiscountChecker: SwapDiscountChecker,
     private val swapGasCalculator: SwapGasCalculator,
     private val swapValidator: SwapValidator,
-    private val fiatValueToString: FiatValueToStringMapper,
 ) {
 
     /**
@@ -378,24 +376,25 @@ internal class SwapQuotePipeline(
                 tierType = vultResult.tierType,
             )
 
-        // SwapKit BTC settles by broadcasting the provider's PSBT, whose miner fee is the only
-        // network cost — and it is already surfaced as the UTXO plan network fee below. SwapKit
-        // reports that same deposit cost as its inbound fee, so counting it again as a swap fee
-        // would
-        // double-count the BTC network cost in the headline total (iOS shows it once). Zero the
-        // swap-fee contribution and the breakdown row so Total reconciles to Network Fee alone; the
+        // SwapKit UTXO-family sources (Bitcoin PSBT deposit, Cardano CBOR deposit) settle by
+        // broadcasting a deposit whose on-chain miner fee is the only network cost, already
+        // surfaced on the Network Fee row (BTC via the UTXO plan below; Cardano via the flat
+        // send-style plan fee — Cardano is excluded from the plan path in `isUtxoSwap`). SwapKit
+        // reports that same deposit cost as its wire inbound fee, so counting it again as a swap
+        // fee would double-count the source-chain network cost in the headline total. iOS discards
+        // the wire inbound fee for both families and shows the cost once as Network Fee
+        // (`SwapCryptoLogic.fee`), so zero the swap-fee contribution and its breakdown row; the
         // affiliate fee is already baked into expectedDstValue.
         val quote = quoteResult.quote
         val isSwapKitUtxoSwap =
-            quote is SwapQuote.SwapKit &&
-                srcToken.chain.standard == TokenStandard.UTXO &&
-                srcToken.chain != Chain.Cardano
+            quote is SwapQuote.SwapKit && srcToken.chain.standard == TokenStandard.UTXO
         val effectiveSwapFeeFiat =
             if (isSwapKitUtxoSwap) FiatValue(BigDecimal.ZERO, quoteResult.swapFeeFiat.currency)
             else quoteResult.swapFeeFiat
-        val feeText =
-            if (isSwapKitUtxoSwap) fiatValueToString(effectiveSwapFeeFiat, asFee = true)
-            else quoteResult.feeText
+        // Empty text hides the swap-fee breakdown row entirely (iOS `swapFeeString` returns
+        // `.empty` here): the deposit cost is already surfaced as the Network Fee, so there is no
+        // separate swap fee to show rather than a redundant "$0.00" row.
+        val feeText = if (isSwapKitUtxoSwap) "" else quoteResult.feeText
 
         val priceImpactDisplay = formatPriceImpact(quoteResult.priceImpact)
 
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
@@ -331,7 +331,7 @@ internal class SwapQuotePipeline(
     }
 
     /** Builds the display-ready [SwapQuotePipelineResult.Success] from the winning quote. */
-    private suspend fun buildSuccess(
+    internal suspend fun buildSuccess(
         bestQuote: BestQuote,
         src: SendSrc,
         srcTokenValue: BigInteger,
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
```diff
@@ -387,7 +387,9 @@ internal class SwapQuotePipeline(
         // affiliate fee is already baked into expectedDstValue.
         val quote = quoteResult.quote
         val isSwapKitUtxoSwap =
-            quote is SwapQuote.SwapKit && srcToken.chain.standard == TokenStandard.UTXO
+            quote is SwapQuote.SwapKit &&
+                srcToken.chain.standard == TokenStandard.UTXO &&
+                srcToken.chain != Chain.Cardano
```
rkokhatskyi: Un-zeroing this for Cardano reopens a double count because `UtxoFeeService.kt:27`'s flat 180000 lovelace Network Fee is never replaced by a real plan like `SwapGasCalculator.kt:124` does for other UTXO chains, so Total Fee now adds that guess to SwapKit's real inbound fee for the same on-chain cost

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
```diff
@@ -387,7 +387,9 @@ internal class SwapQuotePipeline(
         // affiliate fee is already baked into expectedDstValue.
         val quote = quoteResult.quote
         val isSwapKitUtxoSwap =
-            quote is SwapQuote.SwapKit && srcToken.chain.standard == TokenStandard.UTXO
+            quote is SwapQuote.SwapKit &&
```
rkokhatskyi: utxoFeeData's SwapKit branch at `SwapQuotePipeline.kt:413` still checks only `chain.standard == TokenStandard.UTXO`, unlike this check and `isUtxoSwap` below, so a Cardano quote gets a non-null utxoDstAddress and utxoMemo on Success, harmless only because both readers already gate on isUtxoSwap

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
aminsato: You're right — I confirmed the double count and reversed course. SwapKit's fee for Cardano is source-native (the ADA deposit cost), the same on-chain cost as the flat network fee, so un-zeroing it double-counted in the total. I verified the intended behavior against the running macOS app: a Cardano→ETH SwapKit swap shows no Swap Fee row at all — just Network Fee (the ADA cost, once), the VULT discount, and Price Impact (iOS `SwapCryptoLogic.fee` returns the plan fee and `inboundFeeDecimal` returns `.zero`).

Fixed in b841c41cb: dropped the Cardano carve-out so its swap fee is zeroed like Bitcoin, and additionally hide the row entirely when zeroed (empty `feeText`) instead of showing `$0.00`, matching iOS.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
aminsato: Good catch. Left as-is for now: it's a pre-existing condition (not introduced by this PR — the same on `main`), and it's harmless because both readers (`utxoDstAddress`/`utxoMemo`) gate on `isUtxoSwap`, which does exclude Cardano. Happy to tighten the `utxoFeeData` SwapKit branch to also exclude Cardano in a follow-up if you'd prefer the consistency.

johnnyluo (approved):
LGTM

PR #5338, opened 2026-07-19
fix(substrate): stop reporting success on a malformed broadcast body

### LINKED ISSUE #5320: Substrate broadcast reports success on a malformed RPC response: `(null, null)` body becomes a fabricated success hash (Polkadot + Bittensor)
## Summary

`broadcastTransaction` returns `String?` for both Substrate chains, and `null` is overloaded to mean two incompatible things:

- **"duplicate broadcast — the MPC peer already submitted this, resolve to the known hash"** (intended)
- **"the response body had neither `result` nor `error` and we have no idea what happened"** (unintended)

Both land on the same `.orKnownHash(tx)` path and are reported to the user as a **successful broadcast with a locally-computed hash**. Nothing verifies on-chain.

## The chain of events

`Json { explicitNulls = false }` is the production config (`DataModule.kt:35-40`, injected via `HttpClientConfigurator.kt:43`). `PolkadotBroadcastTransactionJson` declares `result: String?` / `error: …?` with no defaults (`PolkadotResponse.kt:31-35`).

So a truncated, malformed, or gateway-mangled body decodes cleanly to `(null, null)`:

```kotlin
// PolkadotApi.kt:181-191
val responseContent = response.body<PolkadotBroadcastTransactionJson>()
if (responseContent.error != null) {
    if (responseContent.error.code == 1012 || responseContent.error.code == 1013) {
        return null          // intended: duplicate → orKnownHash
    }
    throw Exception(...)
}
return responseContent.result   // (null, null) lands here → also null
```

Then `orKnownHash` (`BroadcastTxUseCase.kt:292-293`) turns that `null` into `tx.transactionHash` — a hash we computed locally, for a transaction that may never have reached a node.

`recoverIfAlreadyBroadcast` (`BroadcastTxUseCase.kt:228-248`) only runs its `verify` callback inside `catch`. The `(null, null)` path returns **normally**, so `isExtrinsicInChain` never runs.

Same shape on Bittensor (`BittensorApi.kt:119-142`, `:141`), which additionally has **no `recoverIfAlreadyBroadcast` wrapper at all** (`BroadcastTxUseCase.kt:182-184` vs Polkadot's `:173-180`).

## Impact

User sees "sent", gets a hash, and the extrinsic may not exist. On Bittensor there is no recovery path whatsoever.

## Secondary: Bittensor idempotency is case-sensitive and code-blind

`BittensorApi.kt:134` — `message?.contains("Already Imported")`. No `1012`/`1013` code check, no `already known` / `temporarily banned`, case-sensitive. Polkadot matches on numeric code only (`PolkadotApi.kt:183`) with no string fallback — which is fragile in the other direction, since Polkadot goes through the `api.vultisig.com/dot/` proxy (`PolkadotApi.kt:293`): a proxy that normalizes JSON-RPC codes while preserving messages would defeat the check entirely.

The repo's own fixture shows the code is on the wire and ignored for Bittensor (`BittensorApiBodyReadTest.kt:90-98`).

## Proposed fix

1. Disambiguate the return. Either a sealed result (`Broadcast.Accepted(hash)` / `Broadcast.Duplicate` / `Broadcast.Unknown`), or throw on `(null, null)` so `recoverIfAlreadyBroadcast` engages and verifies on-chain.
2. Wrap Bittensor in `recoverIfAlreadyBroadcast` with an `isExtrinsicInChain` verify, as Polkadot has.
3. Make idempotency detection uniform across both chains: numeric codes `1012`/`1013` **and** case-insensitive string fallbacks for `already imported` / `already known` / `temporarily banned`.

Best form is one shared Substrate broadcast classifier both API classes call — iOS did exactly this (`RpcService.BroadcastErrorClassifier`) and has zero drift between the two chains as a result. Android duplicated the logic across two classes and drifted on three axes.

## Acceptance
- [ ] `(null, null)` body no longer reports success
- [ ] Bittensor verifies on-chain before claiming success
- [ ] Shared idempotency classifier used by both chains
- [ ] Tests: truncated body, case-variant "already imported", code 1012/1013 with an unexpected message

Related: vultisig-sdk#1391 (the SDK's Bittensor resolver has the same missing-`result` hole; Android's version additionally reaches Polkadot).

aminsato:
Closes #5320

### Transaction hash
https://taostats.io/extrinsic/0xcafd2c16b579ec3e0c2ac1e38d08a148c8f242052c339f9f70beaad520091df0

## Problem

`broadcastTransaction` on both Substrate chains returns `String?`, and `null` was overloaded to mean two incompatible things:

- **intended** — a duplicate rebroadcast (the MPC peer already submitted this) → resolve to the known hash
- **unintended** — the response body had neither `result` nor `error` and we have no idea what happened

Under the production `Json { explicitNulls = false }` config, a truncated / gateway-mangled `author_submitExtrinsic` body decodes cleanly to `(null, null)`. Both cases flowed through `.orKnownHash(tx)` and were reported to the user as a **successful broadcast with a locally-computed hash** — for a transaction that may never have reached a node. Nothing verified on-chain. On Bittensor there was **no recovery/verify path at all**.

Bittensor's idempotency check was also weaker than Polkadot's: it matched only the case-sensitive string `"Already Imported"` and ignored the numeric code on the wire, while Polkadot matched only the numeric code with no message fallback.

## Fix

One shared classifier both chains call (mirrors iOS's `RpcService.BroadcastErrorClassifier`, so the two chains cannot drift):

- `result` present → return the hash
- idempotent error — codes `1012`/`1013` **or** case-insensitive `already imported` / `already known` / `temporarily banned` → return `null`
- `(null, null)` or any other error → **throw**

- **`SubstrateBroadcast.kt`** (new) — the classifier + `SubstrateBroadcastException`.
- **`PolkadotApi` / `BittensorApi`** — both `broadcastTransaction` delegate to it (uniform idempotency detection).
- **`BroadcastTxUseCase`** — wrapped **Bittensor** in `recoverIfAlreadyBroadcast` with an on-chain `verify` (via the Bittensor status provider), so a thrown malformed body now checks the extrinsic actually landed before claiming success — as Polkadot already did.

## Acceptance criteria

- [x] `(null, null)` body no longer reports success
- [x] Bittensor verifies on-chain before claiming success
- [x] Shared idempotency classifier used by both chains
- [x] Tests: truncated body, case-variant "already imported", codes 1012/1013 with an unexpected message

## Test plan

- `SubstrateBroadcastTest` (8 cases) — accept / `(null,null)` throws / codes 1012+1013 / case-variant message fallbacks / non-idempotent error throws.
- `PolkadotApiBodyReadTest` + `BittensorApiBodyReadTest` — added truncated-body-throws and code/message-fallback cases through the real kotlinx decode path (`explicitNulls = false`).
- All `:data` unit tests pass.

## Device verification (happy path, both chains)

Real Vultisig-signed self-sends broadcast and confirmed successful on-chain — the newly-wrapped Bittensor recover/verify path does not break normal sends:

| Chain | Call | Result | Explorer |
|---|---|---|---|
| Bittensor | `Balances.transfer_allow_death` | ✅ success (block 8654541) | https://taostats.io/extrinsic/0x9125336c984a40a3acbe6e954e9e05cb5cd1cf2dcd9551724524d7c93a4261fd |
| Polkadot AssetHub | `Balances.transfer_keep_alive` | ✅ success (block 18397944) | https://assethub-polkadot.subscan.io/extrinsic/0x9b9d5163ea147d78b76c5ab7711f3177886a77c1b3089b5099588528519ab29b |

Related: vultisig-sdk#1391 (the SDK's Bittensor resolver has the same missing-`result` hole).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction broadcast handling across supported Substrate-based networks.
  - Idempotent “already imported/known” submissions (e.g., JSON-RPC code **1013**) are now treated as non-failures.
  - “Temporarily banned” submissions (e.g., JSON-RPC code **1012**) and malformed/ambiguous responses now correctly fail instead of being misreported as success.
  - Bittensor broadcasts now only report success after the transaction is confirmed on-chain.

- **Tests**
  - Expanded coverage for duplicate handling, temporarily banned cases, malformed responses, non-2xx idempotent bodies, and classifier edge cases.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/SubstrateBroadcast.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/SubstrateBroadcast.kt
@@ -1,0 +1,61 @@
+package com.vultisig.wallet.data.api
+
+import com.vultisig.wallet.data.api.models.cosmos.PolkadotBroadcastTransactionErrorJson
+import com.vultisig.wallet.data.api.models.cosmos.PolkadotBroadcastTransactionJson
+
+/** Thrown when a Substrate `author_submitExtrinsic` response cannot be interpreted as success. */
+internal class SubstrateBroadcastException(message: String) : Exception(message)
+
+/**
+ * Shared classifier for Substrate `author_submitExtrinsic` responses (Polkadot + Bittensor).
+ *
+ * The JSON-RPC envelope is overloaded three ways and each must be handled distinctly, otherwise a
+ * malformed body silently becomes a fabricated success:
+ * - `result` present -> the node accepted the extrinsic; return its hash.
+ * - idempotent error -> a peer / second signing device already submitted this extrinsic; return
+ *   null so the caller resolves to the locally known hash.
+ * - anything else -> unknown or failed: throw so the caller's recover-if-already-broadcast path
+ *   verifies on-chain instead of reporting success blindly.
+ *
+ * The `(null, null)` case — neither `result` nor `error`, e.g. a truncated or proxy-mangled body
+ * decoded under the production `explicitNulls = false` config — falls into the last bucket and must
+ * not be reported as success.
+ *
+ * Both API classes share one classifier so idempotency detection cannot drift between the chains.
+ */
+internal object SubstrateBroadcast {
+
+    // Substrate transaction-pool JSON-RPC error codes for an already-known / already-imported
+    // extrinsic (a harmless duplicate rebroadcast).
+    private val IDEMPOTENT_CODES = setOf(1012, 1013)
+
+    // Case-insensitive message fallbacks for nodes / proxies that normalize the numeric code away
+    // (e.g. Polkadot's api.vultisig.com/dot proxy) while preserving the human-readable message.
+    private val IDEMPOTENT_MESSAGES =
+        listOf("already imported", "already known", "temporarily banned")
+
+    /**
+     * @return the broadcast hash, or null for an idempotent duplicate.
+     * @throws SubstrateBroadcastException on a node error or an uninterpretable `(null, null)`
+     *   body.
+     */
+    fun classify(response: PolkadotBroadcastTransactionJson): String? {
+        val error = response.error
+        if (error != null) {
+            if (isIdempotent(error)) return null
+            throw SubstrateBroadcastException(
+                "Substrate broadcast failed: ${error.data ?: error.message}"
+            )
+        }
+        return response.result
+            ?: throw SubstrateBroadcastException(
+                "Substrate broadcast returned neither a result nor an error"
+            )
+    }
+
+    private fun isIdempotent(error: PolkadotBroadcastTransactionErrorJson): Boolean {
+        if (error.code in IDEMPOTENT_CODES) return true
+        val message = error.message?.lowercase() ?: return false
+        return IDEMPOTENT_MESSAGES.any { message.contains(it) }
+    }
+}
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/SubstrateBroadcast.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/SubstrateBroadcast.kt
@@ -25,14 +25,22 @@ internal class SubstrateBroadcastException(message: String) : Exception(message)
  */
 internal object SubstrateBroadcast {
 
-    // Substrate transaction-pool JSON-RPC error codes for an already-known / already-imported
-    // extrinsic (a harmless duplicate rebroadcast).
-    private val IDEMPOTENT_CODES = setOf(1012, 1013)
+    // Substrate transaction-pool JSON-RPC error code for an already-imported extrinsic (a harmless
+    // duplicate rebroadcast). Only 1013 (`AlreadyImported`) is idempotent. 1012
+    // (`TemporarilyBanned`) is deliberately excluded: upstream bans a transaction after it was
+    // reported invalid, so it is retriable and potentially invalid — treating it as a duplicate
+    // would fabricate success and skip the on-chain verification path. It must throw and flow
+    // through the caller's recover-if-already-broadcast path instead.
+    private val IDEMPOTENT_CODES = setOf(1013)
 
-    // Case-insensitive message fallbacks for nodes / proxies that normalize the numeric code away
-    // (e.g. Polkadot's api.vultisig.com/dot proxy) while preserving the human-readable message.
-    private val IDEMPOTENT_MESSAGES =
-        listOf("already imported", "already known", "temporarily banned")
+    // Canonical duplicate messages for nodes / proxies that normalize the numeric code away (e.g.
+    // Polkadot's api.vultisig.com/dot proxy) while preserving the human-readable message. Matched
+    // with word boundaries (case-insensitive) so an unrelated error merely containing these words
+    // is not misclassified as a harmless duplicate.
+    private val IDEMPOTENT_MESSAGE_PATTERNS =
+        listOf("already imported", "already known").map {
+            Regex("\\b${Regex.escape(it)}\\b", RegexOption.IGNORE_CASE)
+        }
 
     /**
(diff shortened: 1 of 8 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/SubstrateBroadcast.kt
```diff
@@ -0,0 +1,61 @@
...
+ * decoded under the production `explicitNulls = false` config — falls into the last bucket and must
+ * not be reported as success.
+ *
+ * Both API classes share one classifier so idempotency detection cannot drift between the chains.
+ */
+internal object SubstrateBroadcast {
+
+    // Substrate transaction-pool JSON-RPC error codes for an already-known / already-imported
+    // extrinsic (a harmless duplicate rebroadcast).
+    private val IDEMPOTENT_CODES = setOf(1012, 1013)
```
johnnyluo: **[P1] Do not treat `TemporarilyBanned` / code 1012 as a successful duplicate.**

Upstream Substrate maps 1012 to `TemporarilyBanned`, not `AlreadyImported` (1013), and explicitly describes temporarily banned transactions as retriable and potentially invalid: https://github.com/paritytech/polkadot-sdk/blob/3f7ecbc466872c43259a6d2b707fbe28a4e058bd/substrate/client/rpc-api/src/author/error.rs and https://github.com/paritytech/polkadot-sdk/blob/3f7ecbc466872c43259a6d2b707fbe28a4e058bd/substrate/client/transaction-pool/api/src/error.rs

Returning `null` here makes `.orKnownHash(tx)` report success immediately and skips the new on-chain verification path, preserving the fabricated-success failure mode for a transaction that was banned after an invalid submission. Please remove code 1012 and the `temporarily banned` message from the idempotent set so they throw and flow through `recoverIfAlreadyBroadcast`; only 1013 / already-imported should resolve directly as a duplicate. The new 1012 test should be updated accordingly.

aminsato: Fixed in feb9aaccf. Removed `1012` from `IDEMPOTENT_CODES` (now `setOf(1013)`) and dropped `"temporarily banned"` from the message set, so `TemporarilyBanned` now throws and flows through `recoverIfAlreadyBroadcast` → on-chain verification. Only `1013`/already-imported resolves directly as a duplicate.

Tests updated accordingly: the old "1012 resolves to null" cases now assert they throw (`SubstrateBroadcastTest`, `Bittensor/PolkadotApiBodyReadTest`), and I added Bittensor `BroadcastTxUseCase` branch coverage confirming a `SubstrateBroadcastException` recovers to the known hash when on-chain and rethrows otherwise.

data/src/main/kotlin/com/vultisig/wallet/data/api/SubstrateBroadcast.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    // Case-insensitive message fallbacks for nodes / proxies that normalize the numeric code away
+    // (e.g. Polkadot's api.vultisig.com/dot proxy) while preserving the human-readable message.
+    private val IDEMPOTENT_MESSAGES =
+        listOf("already imported", "already known", "temporarily banned")
```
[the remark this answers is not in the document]
aminsato: Addressed in feb9aaccf. Removing `"temporarily banned"` (per the maintainer's P1) shrinks the surface, and the remaining `already imported` / `already known` phrases are now matched with word-boundary, case-insensitive regex (`\balready imported\b`) instead of raw `contains`, so an unrelated error merely containing those words no longer classifies as a duplicate. I kept regex rather than `equals`/`startsWith` because the proxy fallback messages arrive as full sentences (e.g. "Transaction is already imported."). Added a regression test that an unrelated `imported`-mentioning error throws.

johnnyluo (approved):
The 1012/TemporarilyBanned finding is resolved: it now throws into on-chain recovery, with accepted, duplicate, recovered, and rethrow paths covered. Latest build, lint, and instrumented-test checks pass.

PR #5340, opened 2026-07-19
fix(swap): show chain indicator on both swap-overview token rows

### LINKED ISSUE #5339: Swap overview: From token row missing chain indicator shown on To row
### Description

On the **Swap overview** screen, the **From** token row does not show a chain indicator, while the **To** token row does.

The To row displays a small chain badge + label (e.g. `on ETH`) next to the destination asset, making it clear which chain the received token lives on. The From row shows only the amount, symbol, and fiat value — with no equivalent `on <chain>` indicator. This is inconsistent between the two sides of the swap.

### Expected

The From token row should show a chain indicator (badge + `on <chain>` label) consistent with the To token row, so both sides clearly communicate the chain of each asset.

### Actual

- **From** row: amount + symbol + fiat value only, no chain indicator.
- **To** row: amount + symbol + fiat value **and** `on <chain>` indicator.

### Screen

`ui/screens/swap/VerifySwapScreen.kt` (Swap overview / "You're swapping").

### Labels

ui-mismatch

aminsato:
## Summary

On the **Swap overview** screen the **From** token row omitted the `on <chain>` indicator while the **To** row showed it, so the two sides of the swap communicated their chain inconsistently.

The indicator was gated on `isSwap && (!token.isNativeToken || token.chain.isLayer2)`, which hid it for native assets on their own L1. On a native→token swap (e.g. BTC → USDC on Ethereum) the native From asset had no chain label while the token To asset did.

## Changes

- `VerifySwapScreen.kt` — gate the `on <chain>` indicator on `isSwap` alone, so both swap rows always render it. Removed the now-unused `isLayer2` import.
- `PreviewActivity.kt` — added a `swap_confirm_chain` debug preview (native From / token To) used to capture the renders below.

The only other `SwapToken` caller (verify-deposit) uses the default `isSwap = false` and is unaffected.

## Before / After

| Before | After |
|--------|-------|
| before | after |

In *before*, only the To row (`USDC`) shows `on ETH`; the From row (`BTC`) has no indicator. In *after*, the From row shows `on BTC`, matching the To row.

## Test plan

- Open a swap with a native source and a token destination (e.g. BTC → USDC), reach the Swap overview, and confirm both rows show the `on <chain>` indicator.

Closes #5339

## Summary by CodeRabbit

* **New Features**
  * Added a swap confirmation preview showing chain indicators for both native and token assets.
  * Swap rows now consistently display the relevant network, including native assets.

* **Bug Fixes**
  * Improved chain visibility in swap confirmation screens for clearer transaction details.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -188,6 +188,7 @@ class PreviewActivity : ComponentActivity() {
             OnBoardingComposeTheme {
                 when (screen) {
                     "swap_confirm" -> SwapConfirmPreview()
+                    "swap_confirm_chain" -> SwapChainIndicatorPreview()
                     "swap_confirm_disabled" -> SwapConfirmPreview(allConsents = false)
                     "swap_confirm_external_recipient" ->
                         SwapConfirmPreview(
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
Reviewed locally: implementation is correctly scoped; Gradle tests and diff checks pass.

PR #5342, opened 2026-07-20
fix: check swap inbound status before signing

johnnyluo:
## What changed

- add a live sign-time inbound status preflight for native THORChain and MayaChain swaps
- query the source protocol directly and match the source chain's inbound entry
- block signing when `halted`, `global_trading_paused`, or `chain_trading_paused` is set
- fail closed when inbound status cannot be fetched or decoded
- disable signing controls while the preflight is running
- add MayaChain `/mayachain/inbound_addresses` API support and tolerate pause fields omitted when false

## Why

The quote flow can detect trading halts, but the quote may become stale before the user signs. Android previously launched keysign from the verify screen without re-checking live inbound status, which could spend source-chain gas on a deposit that THORChain or MayaChain would refund. This brings Android in line with the iOS sign-time safety gate.

## User impact

Users are prevented from signing native THORChain/MayaChain swaps when the source chain has halted after quote creation. If live status cannot be verified, signing stays blocked and the existing trading-halted message is shown so the user can retry later.

## Validation

- `./gradlew test`
- `./gradlew ktfmtCheck`
- focused ViewModel, halt-gate, and Maya inbound decoding tests

on chain tx link: https://runescan.io/tx/B49D413263AF186CBB7FB2C5EEC3AE552B9682B4D2847D1551504DE95230C4BC

## Summary

* **New Features**
  * Added inbound trading preflight checks for native swaps to block signing when the source chain is halted or unavailable.
  * Added MayaChain inbound status fetching to support halt validation.
* **Bug Fixes**
  * Implemented “fail closed” verification: signing is blocked when inbound status can’t be confirmed.
* **UI Improvements**
  * Disabled swap CTAs while signing is in progress (including fast-sign paired buttons).
* **Tests**
  * Added/extended unit tests for halted/unavailable behavior and signing prevention.
* **Chores**
  * Updated network caching behavior and moved VerifyDepositViewModel background work to the injected I/O dispatcher.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapInboundHaltPreflight.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapInboundHaltPreflight.kt
@@ -1,0 +1,55 @@
+package com.vultisig.wallet.ui.models.swap
+
+import com.vultisig.wallet.data.api.MayaChainApi
+import com.vultisig.wallet.data.api.ThorChainApi
+import com.vultisig.wallet.data.api.errors.SwapException
+import com.vultisig.wallet.data.api.models.thorchain.THORChainInboundAddress
+import com.vultisig.wallet.data.models.SwapTransaction
+import com.vultisig.wallet.data.models.payload.SwapPayload
+import com.vultisig.wallet.data.models.swapAssetName
+import javax.inject.Inject
+import kotlinx.coroutines.CancellationException
+import timber.log.Timber
+
+/**
+ * Live, sign-time safety gate for native THORChain and MayaChain swaps.
+ *
+ * Quotes can become stale between the swap form and signing. Fetching the protocol's current
+ * inbound set here prevents spending source-chain gas on a deposit that the protocol will refund.
+ * The check deliberately fails closed when the inbound status cannot be verified.
+ */
+internal class SwapInboundHaltPreflight
+@Inject
+constructor(private val thorChainApi: ThorChainApi, private val mayaChainApi: MayaChainApi) {
+
+    suspend fun assertSourceChainNotHalted(transaction: SwapTransaction) {
+        val fetchInboundAddresses: suspend () -> List<THORChainInboundAddress> =
+            when (transaction.payload) {
+                is SwapPayload.ThorChain -> thorChainApi::getTHORChainInboundAddresses
+                is SwapPayload.MayaChain -> mayaChainApi::getInboundAddresses
+                else -> return
+            }
+
+        val inboundAddresses =
+            try {
+                fetchInboundAddresses()
+            } catch (e: CancellationException) {
+                throw e
+            } catch (e: Exception) {
+                Timber.w(e, "Unable to verify swap inbound status; blocking native swap")
+                throw SwapException.TradingHalted(SIGNING_BLOCKED_MESSAGE)
+            }
+
+        val sourceChain = transaction.srcToken.chain.swapAssetName()
+        val inbound =
+            inboundAddresses.firstOrNull { it.chain.equals(sourceChain, ignoreCase = true) }
+
+        if (inbound?.let { it.halted || it.globalTradingPaused || it.chainTradingPaused } == true) {
+            throw SwapException.TradingHalted(SIGNING_BLOCKED_MESSAGE)
+        }
+    }
+
+    private companion object {
+        const val SIGNING_BLOCKED_MESSAGE = "Source-chain trading is halted or unavailable"
+    }
+}
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
@@ -5,6 +5,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import androidx.navigation.toRoute
 import com.vultisig.wallet.R
+import com.vultisig.wallet.data.api.errors.SwapException
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.Coins
 import com.vultisig.wallet.data.models.SwapTransaction
@@ -224,10 +225,17 @@ constructor(
         state.update { it.copy(isSigning = true) }
         viewModelScope.safeLaunch(
             onError = { e ->
-                Timber.w(e, "Swap sign-time inbound preflight blocked signing")
-                state.update {
-                    it.copy(errorText = UiText.StringResource(R.string.swap_error_trading_halted))
-                }
+                // Only the preflight fails closed as a halt. A keysign/navigation failure must not
+                // masquerade as "trading halted", so it surfaces a generic error instead.
+                val errorMessage =
+                    if (e is SwapException.TradingHalted) {
+                        Timber.w(e, "Swap sign-time inbound preflight blocked signing")
+                        R.string.swap_error_trading_halted
+                    } else {
+                        Timber.e(e, "Swap keysign launch failed")
+                        R.string.error_view_default_title
+                    }
+                state.update { it.copy(errorText = UiText.StringResource(errorMessage)) }
             }
         ) {
             try {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
@@ -103,6 +103,7 @@ internal data class VerifySwapUiModel(
     val txScanStatus: TransactionScanStatus = TransactionScanStatus.NotStarted,
     val showScanningWarning: Boolean = false,
     val vaultName: String = "",
+    val isSigning: Boolean = false,
 ) {
     val hasAllConsents: Boolean
         get() =
@@ -122,6 +123,7 @@ constructor(
     private val isVaultHasFastSignById: IsVaultHasFastSignByIdUseCase,
     private val securityScannerService: SecurityScannerContract,
     private val vaultRepository: VaultRepository,
+    private val inboundHaltPreflight: SwapInboundHaltPreflight,
 ) : ViewModel() {
 
     val state = MutableStateFlow(VerifySwapUiModel())
@@ -130,6 +132,7 @@ constructor(
     private val vaultId: VaultId = args.vaultId
     private val transactionId: String = args.transactionId
     private var _fastSign = false
+    private var transaction: SwapTransaction? = null
 
     private val _fastSignFlow = Channel<Boolean>()
     val fastSignFlow = _fastSignFlow.receiveAsFlow()
@@ -142,6 +145,7 @@ constructor(
                         navigator.back()
                         return@launch
                     }
+            this@VerifySwapViewModel.transaction = transaction
             val vault = vaultRepository.get(vaultId)
             val vaultName = vault?.name
             if (vaultName == null) {
@@ -204,23 +208,41 @@ constructor(
     }
 
     private fun keysign(keysignInitType: KeysignInitType) {
-        if (state.value.hasAllConsents) {
-            viewModelScope.launch {
-                launchKeysign(
-                    keysignInitType,
-                    transactionId,
-                    password.value,
-                    Route.Keysign.Keysign.TxType.Swap,
-                    vaultId,
-                )
-            }
-        } else {
+        if (!state.value.hasAllConsents) {
             state.update {
                 it.copy(
                     errorText =
                         UiText.StringResource(R.string.verify_transaction_error_not_enough_consent)
                 )
             }
+            return
+        }
+        if (state.value.isSigning) return
+
+        val transaction = transaction ?: return
+        state.update { it.copy(isSigning = true) }
+        viewModelScope.launch {
+            try {
+                inboundHaltPreflight.assertSourceChainNotHalted(transaction)
+            } catch (e: kotlinx.coroutines.CancellationException) {
+                throw e
+            } catch (e: Exception) {
+                Timber.w(e, "Swap sign-time inbound preflight blocked signing")
+                state.update {
+                    it.copy(
+                        errorText = UiText.StringResource(R.string.swap_error_trading_halted),
+                        isSigning = false,
+                    )
+                }
+                return@launch
+            }
+            launchKeysign(
+                keysignInitType,
+                transactionId,
+                password.value,
+                Route.Keysign.Keysign.TxType.Swap,
+                vaultId,
+            )
         }
     }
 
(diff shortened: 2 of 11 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
```diff
@@ -204,23 +208,41 @@ constructor(
...
+                return@launch
+            }
+            launchKeysign(
+                keysignInitType,
+                transactionId,
+                password.value,
+                Route.Keysign.Keysign.TxType.Swap,
+                vaultId,
+            )
         }
```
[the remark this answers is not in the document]
johnnyluo: Fixed in 63c0ea77c. The signing block now uses a finally clause to reset isSigning after successful navigation, failures, and coroutine cancellation; the focused ViewModel tests also assert the state is restored.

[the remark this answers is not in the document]
johnnyluo: Fixed in 63c0ea77c. The network preflight now runs through viewModelScope.safeLaunch, retaining cancellation propagation and the existing user-facing error handling.

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/THORChainInboundAddress.kt
```diff
@@ -9,9 +9,11 @@ data class THORChainInboundAddress(
...
     @SerialName("halted") val halted: Boolean,
-    @SerialName("global_trading_paused") val globalTradingPaused: Boolean,
-    @SerialName("chain_trading_paused") val chainTradingPaused: Boolean,
-    @SerialName("chain_lp_actions_paused") val chainLPActionsPaused: Boolean,
+    // MayaChain omits pause fields when they are false, while THORChain currently emits them.
+    // Defaults keep the shared inbound model compatible with both protocols.
+    @SerialName("global_trading_paused") val globalTradingPaused: Boolean = false,
+    @SerialName("chain_trading_paused") val chainTradingPaused: Boolean = false,
+    @SerialName("chain_lp_actions_paused") val chainLPActionsPaused: Boolean = false,
     @SerialName("gas_rate") val gasRate: String,
```
aminsato: Default `gasRate`/`gasRateUnits` as well. This model is now shared with MayaChain, which omits `false`/empty fields; if mayanode ever drops `gas_rate` for a chain, decoding throws and the preflight blocks that swap as a false halt. These fields aren't used by the halt gate, so defaulting them (`= ""`) removes the last non-safety decode-failure vector.

```kotlin
@SerialName("gas_rate") val gasRate: String = "",
@SerialName("gas_rate_units") val gasRateUnits: String = "",
```

app/src/test/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModelTest.kt
```diff
@@ -222,6 +228,45 @@ internal class VerifySwapViewModelTest {
...
+                IllegalStateException("network unavailable")
+            val vm = createViewModel()
+            advanceUntilIdle()
+
+            vm.consentAmount(true)
+            vm.consentReceiveAmount(true)
+            vm.authFastSign()
+            advanceUntilIdle()
+
+            vm.state.value.errorText.shouldNotBeNull()
```
aminsato, line 266: Assert `isSigning shouldBe false` here too, matching the QR-path test. The `finally` reset is the guarantee that a blocked attempt can be retried, and this is the biometric path that should confirm it.

aminsato (commented):
A couple of actionable inline points below. Overall the gate is correct and well-tested — it fails closed on fetch/decode errors, `ignoreUnknownKeys` covers Maya's extra fields, and the default-`false` pause fields are the right fix.

app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
```diff
@@ -204,22 +209,40 @@ constructor(
...
+            }
+            return
+        }
+        if (state.value.isSigning) return
+
+        val transaction = transaction ?: return
+        state.update { it.copy(isSigning = true) }
+        viewModelScope.safeLaunch(
+            onError = { e ->
+                Timber.w(e, "Swap sign-time inbound preflight blocked signing")
```
rkokhatskyi: `onError` treats every exception from this try block, including `launchKeysign`, as a halt, so a navigation failure unrelated to the preflight surfaces as `trading halted` and logs a misleading `blocked signing` message, should discriminate on `SwapException.TradingHalted` specifically

app/src/main/java/com/vultisig/wallet/ui/screens/swap/VerifySwapScreen.kt
```diff
@@ -183,6 +185,8 @@ private fun VerifySwapScreen(
     onContinueAnyway: () -> Unit,
     onDismissRequest: () -> Unit,
 ) {
+    val isSignEnabled = (!isConsentsEnabled || hasAllConsents) && !isSigning
```
rkokhatskyi, line 188: `isSigning` only drives `VsButtonState` here, never `VsButton`'s `isLoading` like `VerifyDepositScreen` and the Cosmos staking screens do for their own in-flight state, so the new multi-second (timeout plus retries) preflight fetch just greys the button out with zero progress feedback

app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
```diff
@@ -204,22 +209,40 @@ constructor(
...
+                it.copy(
+                    errorText =
+                        UiText.StringResource(R.string.verify_transaction_error_not_enough_consent)
+                )
+            }
+            return
+        }
+        if (state.value.isSigning) return
+
+        val transaction = transaction ?: return
```
rkokhatskyi, line 224: `val transaction = transaction ?: return` silently no-ops the tap with no `errorText` and no `isSigning` update, unlike the `hasAllConsents` guard right above it, so a stale in-memory `SwapTransactionRepository` miss leaves Sign doing nothing with zero feedback

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapInboundHaltPreflight.kt
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Quotes can become stale between the swap form and signing. Fetching the protocol's current
+ * inbound set here prevents spending source-chain gas on a deposit that the protocol will refund.
+ * The check deliberately fails closed when the inbound status cannot be verified.
+ */
+internal class SwapInboundHaltPreflight
+@Inject
+constructor(private val thorChainApi: ThorChainApi, private val mayaChainApi: MayaChainApi) {
+
+    suspend fun assertSourceChainNotHalted(transaction: SwapTransaction) {
```
rkokhatskyi, line 25: This gate only runs on the initiating device, `JoinKeysignViewModel.joinKeysign` (`JoinKeysignViewModel.kt:944`) wires the same `VerifySwapScreen` to a co-sign path with no halt check at all, so any additional signer on a multi-device vault can still co-sign a deposit to a chain that halted after the initiator's own check

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/THORChainInboundAddress.kt
```diff
@@ -9,9 +9,11 @@ data class THORChainInboundAddress(
...
     @SerialName("address") val address: String,
     @SerialName("halted") val halted: Boolean,
-    @SerialName("global_trading_paused") val globalTradingPaused: Boolean,
-    @SerialName("chain_trading_paused") val chainTradingPaused: Boolean,
-    @SerialName("chain_lp_actions_paused") val chainLPActionsPaused: Boolean,
-    @SerialName("gas_rate") val gasRate: String,
-    @SerialName("gas_rate_units") val gasRateUnits: String,
+    // MayaChain omits pause fields when they are false, while THORChain currently emits them.
+    // Defaults keep the shared inbound model compatible with both protocols.
+    @SerialName("global_trading_paused") val globalTradingPaused: Boolean = false,
```
rkokhatskyi, line 14: These new defaults also decode `SecuredAssetLoader`'s THORChain reads (`fetchThorChainInboundForChain`, `SecuredAssetLoader.kt:106-113`), so a response that ever omits `chainTradingPaused`/`chainLPActionsPaused` now silently decodes as false instead of throwing, turning its fail-closed `FetchFailed` into an `Available` deposit address for an unconfirmed-pause chain

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapInboundHaltPreflight.kt
```diff
@@ -0,0 +1,55 @@
...
+        val fetchInboundAddresses: suspend () -> List<THORChainInboundAddress> =
+            when (transaction.payload) {
+                is SwapPayload.ThorChain -> thorChainApi::getTHORChainInboundAddresses
```
rkokhatskyi, line 28: This live fetch has no explicit no-store/cache-bypass directive, unlike iOS's equivalent gate which passes `bypassCache true` at the call site, it silently relies on THORNode/Mayanode continuing to omit `Cache-Control` on the shared `HttpCache`-enabled client (`NetworkModule.kt:43`), needs runtime confirmation if upstream ever adds caching headers

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
johnnyluo: Fixed in 3c64b5044. `onError` now discriminates: only `SwapException.TradingHalted` (which the preflight throws for both an actual halt and a fail-closed fetch failure) shows the trading-halted message; any other exception from `launchKeysign` is logged at error level and surfaces a generic `error_view_default_title`.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/VerifySwapScreen.kt
johnnyluo: Fixed in 3c64b5044. Threaded `isLoading = isSigning` through the standalone `VsButton` and `FastSignPairedButtons` so the preflight fetch shows a spinner instead of just grey, matching `VerifyDepositScreen` / the Cosmos staking screens.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapInboundHaltPreflight.kt
johnnyluo: Fixed in 3c64b5044. Both `inbound_addresses` reads now send `Cache-Control: no-cache, no-store`, matching iOS's `bypassCache`. Since `getTHORChainInboundAddresses` is shared with `SecuredAssetLoader` (also a halt-safety read), the header hardens both call sites against upstream ever adding cache headers to the `HttpCache`-enabled client.

app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
johnnyluo: Leaving as-is. The stale-repository case is already handled at load time: `init` does `swapTransactionRepository.getTransaction(transactionId)` and calls `navigator.back()` on a miss (VerifySwapViewModel.kt:143-148), so the screen has already navigated away. By the time Sign is tappable, `transaction` is non-null; the `?: return` is an unreachable defensive guard, not a live no-feedback path. Unlike the consent guard, `isSigning` hasn't been set yet at that line, so there's nothing to reset either. Adding an `errorText` here would only surface an impossible state.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapInboundHaltPreflight.kt
johnnyluo: Out of scope for this PR, intentionally. This gate matches iOS, which places the sign-time check on the initiating device. In the TSS flow the initiator owns the session start and broadcast; a co-signer only ever joins a session the initiator already opened after passing its own gate, so the initiator is the control point for 'don't spend source-chain gas on a refunded deposit'. A redundant per-co-signer check is defense-in-depth worth a follow-up, but it isn't required for the safety property this PR delivers and would expand scope well beyond the iOS-parity fix.

and it should be fine if we did the check on the initiate device

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/THORChainInboundAddress.kt
johnnyluo: This is safe and intentional. In both protocols an omitted pause field means *false* (not-paused): MayaChain omits false pause flags, and THORChain omits-when-false too, so decoding a missing field as `false` is semantically correct for both, not a fail-open. A genuinely malformed/truncated response still fails closed, because `chain`, `address` and `halted` remain required (no defaults) and will throw → `FetchFailed`. The defaults only relax the fields whose absence already means 'not paused', so `SecuredAssetLoader`'s fail-closed guarantee is preserved. Reverting them would break MayaChain decoding, which is the reason the shared model needs them.

app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
rkokhatskyi: Fixed, the branches now discriminate on `SwapException.TradingHalted` specifically and fall back to a generic error otherwise, so a launchKeysign/navigation failure no longer shows trading halted

app/src/main/java/com/vultisig/wallet/ui/screens/swap/VerifySwapScreen.kt
rkokhatskyi: Fixed, isSigning now drives isLoading on both VsButton and FastSignPairedButtons (which now exposes the param), and VsButton renders its loading indicator independently of state so it shows even while disabled

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapInboundHaltPreflight.kt
rkokhatskyi: Fixed with an explicit no-cache, no-store request header on both calls, verified against Ktor 3.0.3 source that no-cache forces shouldValidate into ShouldValidate, which always triggers a live network round trip rather than serving a cached body

aminsato (approved):
LGTM

PR #5350, opened 2026-07-20
fix(chain): make Coins the single source for chain identity

### LINKED ISSUE #5323: Chain identity: unknown ticker silently resolves to ThorChain on a withdrawal path; second decimals source disagrees by up to 100×
## 1. Unknown ticker silently resolves to ThorChain (live)

`CoinExtensions.kt:54` ends a ticker→`Chain` `when` with:

```kotlin
else -> Chain.ThorChain
```

The keys are display tickers (`"DOT"`, `"TAO"`, `"LUNA"`, `"LUNC"`, `"USDC"` → Noble, …). **Any ticker not in the list resolves to ThorChain** rather than failing.

Caller: `WithdrawSecuredAssetStrategy.kt:67` — a withdrawal path.

This is fragile for a second reason: it couples a fund-path lookup to a ticker registry that has already drifted elsewhere in the repo (`UATOM` vs `ATOM`, `MATIC` vs `POL`). A ticker rename anywhere upstream silently reroutes a withdrawal to ThorChain instead of erroring.

**Fix:** make it exhaustive or throw on unknown. A silent wrong-chain default on a withdrawal path should not exist.

## 2. Two disagreeing decimals sources (latent, one guard from being live)

`Coins.kt` is the correct source. But `CoinType.kt:12-13` exposes `CoinType.decimals = CoinTypeConfiguration.getDecimals(this)`, reachable via `Chain.coinType` (`Chain.kt:87`) and `Chain.toValue()` (`Chain.kt:518-520`). Because several chains share a WalletCore CoinType, the two disagree:

| Chain | `Coins.kt` (correct) | via `Chain.coinType` | error |
|---|---|---|---|
| MayaChain | 10 (CACAO) | `THORCHAIN` → 8 | 100× |
| Bittensor | 9 (TAO) | `POLKADOT` → 10 | 10× |
| Qbtc | 8 | `COSMOS` → 6 | 100× |

**Currently unreachable for amounts:** the `toValue` call sites (`SecuredAssetStrategy.kt:250`, `ChainValidationService.kt:144`) sit inside `validateBtcLikeAmount`, which first calls `chain.getDustThreshold` — and that throws for any non-UTXO chain (`Chain.kt:461-466`). So an unrelated guard is the only thing preventing a 100× amount error on MayaChain.

**But live for tickers today:** `chain.coinType.symbol` on the adjacent line renders **RUNE for MayaChain, DOT for Bittensor, ATOM for Qbtc**.

iOS and Windows share the identical CoinType mapping but never derive decimals or symbols from it, so they are immune.

**Fix:** make `Coins.kt` the single source. Either remove the `CoinType.decimals`/`symbol` accessors or mark them clearly as WalletCore-internal and unsafe for display or amount math.

## Why both of these matter together

Neither would surface in a diff review, and both are the same shape: a per-chain lookup with a silent fallback instead of a failure. The one thing protecting the second is a guard that exists for an unrelated reason and could be relaxed by someone with no idea it was load-bearing.

## Acceptance
- [ ] `CoinExtensions.kt:54` throws or is exhaustive
- [ ] `Coins.kt` is the only source of decimals and symbols for display/amount math
- [ ] Test asserting MayaChain/Bittensor/Qbtc decimals and symbols via every reachable accessor
- [ ] Test asserting an unknown ticker does not resolve to ThorChain

Found during a cross-repo chain-identity review. Related: vultisig-ios#4843, vultisig-windows (companion filed separately).

aminsato:
Closes #5323

## Problem

Two same-shaped chain-identity bugs — a per-chain lookup with a silent wrong-value fallback instead of a failure:

1. **`String.getChain()` silently resolved any unknown ticker to `Chain.ThorChain`.** Its only caller is `WithdrawSecuredAssetStrategy` — a fund path. A ticker rename anywhere upstream (the repo already has `UATOM`/`ATOM`, `MATIC`/`POL` drift) would silently reroute a secured-asset withdrawal to ThorChain instead of erroring.

2. **`coinType.symbol` / `coinType.decimals` disagree with `Coins.kt`.** Several chains share one WalletCore `CoinType`, so the WalletCore-derived values are wrong:

   | Chain | Coins.kt (correct) | via `coinType` | error |
   |---|---|---|---|
   | MayaChain | CACAO / 10 | RUNE / 8 | 100× |
   | Bittensor | TAO / 9 | DOT / 10 | 10× |
   | Qbtc | QBTC / 8 | ATOM / 6 | 100× |

   The amount side (`Chain.toValue`) is currently guarded only by an unrelated `getDustThreshold` check; the symbol side is live today for the network-search filter and the min-amount validation error.

## Fix

- `getChain()` now `error()`s on an unrecognized ticker. `DepositFormViewModel`'s catch turns that into a user-facing error rather than a misrouted withdrawal.
- Made `Coins.kt` the single source: added `Chain.nativeToken` / `Chain.nativeTokenTicker`, redefined `Chain.toValue` to use the Coins native decimals, and routed the three chain-variable `coinType.symbol` display sites (`SecuredAssetStrategy`, `ChainValidationService`, `SelectNetworkViewModel`) through `nativeTokenTicker`.
- Marked `CoinType.symbol` / `CoinType.decimals` with KDoc as WalletCore-internal and unsafe for display/amount math. Hardcoded `Chain.ThorChain.coinType.*` / `CoinType.THORCHAIN.*` usages are left as-is (correct by construction).

## Test plan

- New `ChainIdentitySourceTest` (7 tests, `:data:testDebugUnitTest`) — green:
  - `getChain` resolves known tickers and throws on unknown (asserting it does **not** become ThorChain).
  - MayaChain/Bittensor/Qbtc decimals (10/9/8), tickers (CACAO/TAO/QBTC), and `Chain.toValue` conversion via the reachable accessors.
  - Every chain has exactly one native token in `Coins`.
- `:app:compileDebugKotlin` and `:data` compile clean; ktfmt applied.

## Acceptance
- [x] `getChain()` throws instead of defaulting to ThorChain
- [x] `Coins.kt` is the single source of decimals and symbols for display/amount math
- [x] Test asserting MayaChain/Bittensor/Qbtc decimals and symbols via reachable accessors
- [x] Test asserting an unknown ticker does not resolve to ThorChain

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected native token symbols and decimal handling for chains that share underlying coin types.
  * Minimum amount validation and amount conversions now use each chain’s registered native token details.
  * Unknown token tickers now produce a clear error instead of potentially selecting the wrong chain.
* **Improvements**
  * Network search now matches native token tickers more accurately.
  * Added safeguards to ensure each chain has a single, correctly configured native token.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/SecuredAssetStrategy.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/SecuredAssetStrategy.kt
@@ -14,10 +14,10 @@ import com.vultisig.wallet.data.models.GasFeeParams
 import com.vultisig.wallet.data.models.OPERATION_MINT
 import com.vultisig.wallet.data.models.TokenStandard
 import com.vultisig.wallet.data.models.TokenValue
-import com.vultisig.wallet.data.models.coinType
 import com.vultisig.wallet.data.models.getDustThreshold
 import com.vultisig.wallet.data.models.getPubKeyByChain
 import com.vultisig.wallet.data.models.isSecuredAssetEligible
+import com.vultisig.wallet.data.models.nativeTokenTicker
 import com.vultisig.wallet.data.models.payload.BlockChainSpecific
 import com.vultisig.wallet.data.models.payload.UtxoInfo
 import com.vultisig.wallet.data.models.toValue
@@ -25,7 +25,6 @@ import com.vultisig.wallet.data.repositories.BlockChainSpecificAndUtxo
 import com.vultisig.wallet.data.repositories.BlockChainSpecificRepository
 import com.vultisig.wallet.data.repositories.TokenRepository
 import com.vultisig.wallet.data.repositories.VaultRepository
-import com.vultisig.wallet.data.utils.symbol
 import com.vultisig.wallet.ui.models.send.InvalidTransactionDataException
 import com.vultisig.wallet.ui.utils.UiText
 import com.vultisig.wallet.ui.utils.asUiText
@@ -245,7 +244,7 @@ internal class SecuredAssetStrategy(
     private fun validateBtcLikeAmount(tokenAmountInt: BigInteger, chain: Chain) {
         val minAmount = chain.getDustThreshold
         if (tokenAmountInt < minAmount) {
-            val symbol = chain.coinType.symbol
+            val symbol = chain.nativeTokenTicker
             val name = chain.raw
             val formattedMinAmount = chain.toValue(minAmount).toString()
             throw InvalidTransactionDataException(
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/WithdrawSecuredAssetStrategy.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/WithdrawSecuredAssetStrategy.kt
@@ -64,6 +64,9 @@ internal class WithdrawSecuredAssetStrategy(
 
         val selectedSecureAsset = stateProvider().selectedSecuredAsset
 
+        // getChain() throws on an unrecognised ticker rather than silently defaulting to a chain —
+        // the DepositFormViewModel catch block turns that into a user-facing error instead of
+        // misrouting the withdrawal to the wrong chain.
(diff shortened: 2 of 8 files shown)


johnnyluo (approved):
LGTM

PR #5353, opened 2026-07-20
fix(keysign): reuse Schnorr setup message on retry

### LINKED ISSUE #5327: SchnorrKeysign re-uploads a fresh setup message on every retry
Surfaced during review of PR #5324 (thread https://github.com/vultisig/vultisig-android/pull/5324#discussion_r3607926141). Pre-existing; not introduced by that PR.

## Problem

In `SchnorrKeysign.keysignOneMessageWithRetry`, the initiator branch is:

```kotlin
if (isInitiateDevice) {
    keysignSetupMsg = getKeysignSetupMessage(messageToSign)
    sessionApi.uploadSetupMessage(...)
}
```

There is no `&& attempt == 0` guard, unlike the DKLS and MLDSA equivalents:
- `DklsKeysign.kt`: `if (isInitiateDevice && attempt == 0)`
- `MldsaKeysign.obtainSetupMessage`: `if (isInitiateDevice && attempt == 0) createAndUploadSetupMessage(...) else downloadSetupMessage(...)`

So on each Schnorr retry the initiator regenerates and re-uploads a fresh setup message. A peer still mid-protocol on the previous setup message can then desync (it is signing against a different session).

## Fix

Gate the Schnorr initiator upload on `attempt == 0` to match DKLS/MLDSA; retries should reuse the setup message already on the relay.

## Test

Keysign smoke test across a forced retry (EdDSA/Schnorr path), since this changes the live protocol upload path — verify a retrying initiator does not desync a peer.

aminsato:
## Summary

`SchnorrKeysign.keysignOneMessageWithRetry` regenerated and re-uploaded a fresh setup message on **every** retry, because the initiator's upload branch was missing the `attempt == 0` guard that `DklsKeysign` and `MldsaKeysign` already use:

- `DklsKeysign.kt`: `if (isInitiateDevice && attempt == 0)`
- `MldsaKeysign.obtainSetupMessage`: `if (isInitiateDevice && attempt == 0) …`

On a retry the initiator would push a brand-new setup message to the relay, so a peer still mid-protocol on the previous one desynced (it was signing against a different session).

The fix gates the initiator upload on the first attempt only. On any retry the initiator now falls through to the existing `else` branch and downloads the setup message already on the relay — identical to the DKLS/MLDSA behavior.

Pre-existing bug, surfaced during review of #5324. Closes #5327.

## Changes

- `SchnorrKeysign.kt`: `if (isInitiateDevice)` → `if (isInitiateDevice && attempt == 0)`
- Added `SchnorrKeysignSetupMessageRetryTest` pinning the upload-vs-download decision on a retry.

## Test plan

- [x] `SchnorrKeysignSetupMessageRetryTest` — on `attempt = 1` the initiator calls `getSetupMessage` (download) and never calls `uploadSetupMessage`. Verified the test **fails** when the `attempt == 0` guard is removed and **passes** with the fix.
- [x] Existing `DklsFamilyRelayRecoveryTest` still green.
- [ ] Multi-device EdDSA/Schnorr keysign smoke test across a forced retry (e.g. Solana/Polkadot send with two devices, toggling connectivity to trigger a retry) — verify a retrying initiator does not desync a peer. Requires physical multi-device signing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Schnorr signing retries so subsequent attempts reuse the existing setup message instead of uploading it again.
  * Reduced the risk of retry failures caused by unnecessary setup-message uploads.

* **Tests**
  * Added coverage confirming the correct download behavior during signing retries.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
@@ -343,7 +343,7 @@ class SchnorrKeysign(
         try {
             val keysignSetupMsg: ByteArray
 
-            if (isInitiateDevice) {
+            if (isInitiateDevice && attempt == 0) {
                 keysignSetupMsg = getKeysignSetupMessage(messageToSign)
 
                 sessionApi.uploadSetupMessage(
--- a/app/src/test/java/com/vultisig/wallet/data/keygen/SchnorrKeysignSetupMessageRetryTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/data/keygen/SchnorrKeysignSetupMessageRetryTest.kt
@@ -1,0 +1,179 @@
+@file:OptIn(ExperimentalEncodingApi::class)
+
+package com.vultisig.wallet.data.keygen
+
+import com.vultisig.wallet.data.api.SessionApi
+import com.vultisig.wallet.data.common.md5
+import com.vultisig.wallet.data.mediator.Message
+import com.vultisig.wallet.data.models.KeyShare
+import com.vultisig.wallet.data.models.Vault
+import com.vultisig.wallet.data.usecases.Encryption
+import io.kotest.assertions.throwables.shouldThrowAny
+import io.kotest.matchers.collections.shouldBeEmpty
+import io.kotest.matchers.collections.shouldContain
+import kotlin.io.encoding.Base64
+import kotlin.io.encoding.ExperimentalEncodingApi
+import kotlinx.coroutines.test.runTest
+import org.junit.jupiter.api.Test
+
+/**
+ * Regression test for issue #5327: the Schnorr initiator must only create and upload a fresh setup
+ * message on the first attempt. On a retry it has to reuse the setup message already on the relay
+ * by downloading it — otherwise a peer still mid-protocol on the previous setup message desyncs
+ * against a brand-new session. This mirrors the `isInitiateDevice && attempt == 0` guard used by
+ * `DklsKeysign` and `MldsaKeysign`.
+ *
+ * The test pins the upload-vs-download *decision* on a retry: with `attempt = 1` the initiator must
+ * call `getSetupMessage` (download) and must NOT call `uploadSetupMessage`. Were the `attempt == 0`
+ * guard removed, the initiator would take the upload branch and `getSetupMessage` would never be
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/vultisig/wallet/data/keygen/SchnorrKeysign.kt
```diff
@@ -343,7 +343,7 @@ class SchnorrKeysign(
         try {
             val keysignSetupMsg: ByteArray
 
-            if (isInitiateDevice) {
+            if (isInitiateDevice && attempt == 0) {
```
rkokhatskyi, line 346: gating the initiator upload to attempt == 0 means a failed `uploadSetupMessage` on attempt 0 now makes attempt 1 download and find nothing instead of silently re-uploading, burning `SessionApi`'s 10x 1s retry loop before failing, needs multi-device confirmation

app/src/test/java/com/vultisig/wallet/data/keygen/SchnorrKeysignSetupMessageRetryTest.kt
```diff
@@ -0,0 +1,179 @@
...
+ * message on the first attempt. On a retry it has to reuse the setup message already on the relay
+ * by downloading it — otherwise a peer still mid-protocol on the previous setup message desyncs
+ * against a brand-new session. This mirrors the `isInitiateDevice && attempt == 0` guard used by
+ * `DklsKeysign` and `MldsaKeysign`.
+ *
+ * The test pins the upload-vs-download *decision* on a retry: with `attempt = 1` the initiator must
+ * call `getSetupMessage` (download) and must NOT call `uploadSetupMessage`. Were the `attempt == 0`
+ * guard removed, the initiator would take the upload branch and `getSetupMessage` would never be
+ * called — so this test fails. The signing rounds that follow the download touch the goschnorr JNI
+ * library, which is unavailable on the host JVM; the ceremony therefore throws once the decision
```
rkokhatskyi, line 30: this comment attributes the throw to goschnorr JNI being unavailable, but it actually throws at `SchnorrKeysign.kt:371`'s double Base64 decode of the fixture value before JNI is ever reached, so a more realistic fixture would silently exercise a different path than documented

rkokhatskyi (dismissed):

johnnyluo (approved):
LGTM

PR #5354, opened 2026-07-20
fix(deposit): correct From/To/pool on Mint LP transaction-complete screen

### LINKED ISSUE #5351: Deposit (Mint LP) complete screen: To empty, From not vault-formatted, mint asset missing
### Summary

On the **Transaction complete** screen for a MayaChain deposit that mints an LP position (`Mint` operation), several fields are wrong or missing compared to the **Function overview** screen shown before signing. The transaction itself succeeds — this is purely a display problem on the completion screen.

### Steps to reproduce

1. Perform a MayaChain deposit that adds liquidity / mints an LP position (e.g. CACAO into an `ARB.GLD` LP pool, memo `+:ARB.GLD-…`, Operation = `Mint`).
2. Sign and complete the transaction.
3. Open the **Transaction complete** screen.

### Actual behavior (Transaction complete)

- **To** field is **empty**.
- **From** field shows the raw chain address (`maya12a9r…um6z3zm`) instead of the vault-formatted value `VaultName (chain address)`.
- The **mint asset / pool** is not shown at all — the overview screen shows `→ ARB.GLD LP CACAO` and a `Pool` row, but the completion screen only shows the deposited `1 CACAO`.

### Expected behavior

Match the pre-sign **Function overview** screen:

- **From** should be vault-formatted, e.g. `AE_QBTC_SAM (maya12a9r…um6z3zm)`.
- **To** should be populated (or the row hidden if genuinely not applicable — but it should not render as an empty value).
- The **mint asset / target LP pool** (`ARB.GLD LP CACAO` / `Pool` row) should be displayed so the user can see what was minted.

### Screenshots

| Function overview (before signing) | Transaction complete (bug) |
|---|---|
| overview | complete |

### Notes

- Network: MayaChain, Operation: `Mint`, Memo: `+:ARB.GLD-…`
- Addresses/hashes in screenshots are already truncated by the UI.

aminsato:
## Summary

Closes #5351.

The MayaChain Mint (add-liquidity) deposit **Transaction complete** screen disagreed with the pre-sign **Function overview** in three ways. The transaction itself always succeeded — these were display-only defects.

The deposit done screen renders via `SendTxOverviewScreen` (fed by `TransactionTypeUiModel.toUiTransactionInfo()`). Three independent root causes:

- **From showed a raw address** instead of `VaultName (address)`. The initiator (`KeysignViewModel`) resolved the signing vault name only *inside* the destination-gated label block, which a mint deposit skips because it carries no destination. `srcVaultName` is now resolved up front, so From is vault-formatted regardless of whether a destination exists. (The join path already did this.)
- **To rendered as an empty row.** A mint deposit has no destination — the pool is the target — so the row is now hidden entirely rather than shown blank.
- **The minted pool was missing.** `operation` + `pool` are now carried through to the done screen, which renders the `→ POOL LP TOKEN` line under the token hero and a **Pool** row, matching the pre-sign overview.

## Changes

- `KeysignViewModel` — resolve `srcVaultName` before the destination-gated address-book lookup.
- `SendTxOverviewScreen` — hide the To row when there is no destination; add the LP-destination line and Pool row.
- `UiModelOverviewExtensions` / `UiTransactionInfo` — carry `operation` and `pool` for deposits.
- `PreviewActivity` — `deposit_mint_done` preview for on-device inspection.

## Test plan

- `DepositUiTransactionInfoTest` extended with a Mint LP case (operation/pool carried, destination empty). Full class green.
- On-device (PreviewActivity, `deposit_mint_done`): To row hidden, `→ ARB.GLD LP CACAO` line and Pool row present.
- Note: the vault-formatted From is a `KeysignViewModel` runtime fix; end-to-end confirmation of that label needs a real two-device MayaChain Mint keysign.

## Before / After

| Before | After |
|--------|-------|
| before | after |

_Real Android renders (PreviewActivity, `deposit_mint_done`). Before: empty **To** row, no pool. After: **To** hidden, `→ ARB.GLD LP CACAO` line + **Pool** row. (The mock supplies `srcVaultName`, so From is vault-formatted in both — that fix lives in `KeysignViewModel` and is confirmed by unit test.)_

## Summary by CodeRabbit

* **New Features**
  * Added a deposit-mint preview/completion screen.
  * Enhanced mint/deposit transaction overviews to display the minted pool (and related operation details) when applicable.
* **Bug Fixes**
  * Improved “From” vault labeling for destination-less deposit/mint transactions.
  * Refined when “To”/destination rows are shown to better match decoded transaction details.
  * Ensured deposit ceremony verification consistently shows the same “From” label across steps.
* **Tests**
  * Added unit coverage for deposit-mint transaction info including operation and pool fields.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
@@ -353,14 +353,7 @@ constructor(
         // skip the destination-gated label block below and would show a raw address (issue #5351).
         transactionTypeUiModel?.let { model ->
             _state.update {
-                it.copy(
-                    transactionUiModel =
-                        model.withResolvedLabels(
-                            srcVaultName = vault.name,
-                            dstVaultName = null,
-                            dstAddressBookTitle = null,
-                        )
-                )
+                it.copy(transactionUiModel = model.withResolvedSrcVaultName(vault.name))
             }
         }
 
@@ -1037,6 +1030,26 @@ private fun TransactionTypeUiModel.addressBookTarget(): AddressBookTarget? {
         }
     return if (dstAddress.isBlank()) null else AddressBookTarget(chain, dstAddress)
 }
+
+/**
+ * Sets only the source vault name, leaving any destination labels the joiner already resolved
+ * intact. Used up front for destination-less deposits (e.g. a Mint LP add-liquidity) so the From
+ * row renders "VaultName (address)" without a raw-address flash on the To row.
+ */
+private fun TransactionTypeUiModel.withResolvedSrcVaultName(
+    srcVaultName: String
+): TransactionTypeUiModel =
+    when (this) {
+        is TransactionTypeUiModel.Send ->
+            TransactionTypeUiModel.Send(tx.copy(srcVaultName = srcVaultName))
+        is TransactionTypeUiModel.Deposit ->
+            TransactionTypeUiModel.Deposit(
+                depositTransactionUiModel.copy(srcVaultName = srcVaultName)
+            )
+        is TransactionTypeUiModel.Swap ->
+            TransactionTypeUiModel.Swap(swapTransactionUiModel.copy(srcVaultName = srcVaultName))
+        else -> this
+    }
 
 /** Returns a copy of this model with the resolved From/To labels applied (Send or Deposit only). */
 private fun TransactionTypeUiModel.withResolvedLabels(
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
@@ -348,6 +348,22 @@ constructor(
         )
 
     init {
+        // Resolve the signing vault's name up front so the From row renders "VaultName (address)"
+        // even for deposits with no destination — e.g. a Mint LP add-liquidity — which otherwise
+        // skip the destination-gated label block below and would show a raw address (issue #5351).
+        transactionTypeUiModel?.let { model ->
+            _state.update {
+                it.copy(
+                    transactionUiModel =
+                        model.withResolvedLabels(
+                            srcVaultName = vault.name,
+                            dstVaultName = null,
+                            dstAddressBookTitle = null,
+                        )
+                )
+            }
+        }
+
         // Both Send and Deposit done-screens resolve the same destination labels + "Add to Address
         // Book" affordance, so a Cosmos staking deposit renders identically on the initiator and a
         // joining device (issue #4939).
--- a/app/src/test/java/com/vultisig/wallet/ui/screens/transaction/DepositUiTransactionInfoTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/screens/transaction/DepositUiTransactionInfoTest.kt
@@ -79,8 +79,8 @@ internal class DepositUiTransactionInfoTest {
 
         info.operation shouldBe "Mint"
         info.pool shouldBe "ARB.GLD-0xAFD091"
-        // No destination on an add-liquidity mint — the pool above is the target, so the done
-        // screen hides the "To" row rather than rendering an empty value (issue #5351).
+        // An add-liquidity mint has no destination, so the mapper yields an empty `to` and a
+        // null `toLabel`.
         info.to shouldBe ""
         info.toLabel shouldBe null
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -348,6 +348,22 @@ constructor(
...
     init {
+        // Resolve the signing vault's name up front so the From row renders "VaultName (address)"
+        // even for deposits with no destination — e.g. a Mint LP add-liquidity — which otherwise
+        // skip the destination-gated label block below and would show a raw address (issue #5351).
+        transactionTypeUiModel?.let { model ->
+            _state.update {
+                it.copy(
+                    transactionUiModel =
+                        model.withResolvedLabels(
+                            srcVaultName = vault.name,
```
rkokhatskyi: srcVaultName disagrees with VerifyDepositScreen's From label because this uses the corrected _currentVault while JoinDepositUiModelBuilder.kt:74,132 resolves it from the stale route vaultId, so after a vault auto-switch the two screens show different vault names for the same ceremony.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -348,6 +348,22 @@ constructor(
...
+                            dstVaultName = null,
```
rkokhatskyi: withResolvedLabels unconditionally nulls dstVaultName/dstAddressBookTitle here even when the joiner already resolved them in JoinDepositUiModelBuilder.kt:149-155, so the To row can briefly flash the raw address before the async block restores it; only null when the incoming model lacks them.

app/src/test/java/com/vultisig/wallet/ui/screens/transaction/DepositUiTransactionInfoTest.kt
```diff
@@ -62,4 +62,26 @@ internal class DepositUiTransactionInfoTest {
...
+                    pool = "ARB.GLD-0xAFD091",
+                )
+            )
+
+        val info = model.toUiTransactionInfo()
+
+        info.operation shouldBe "Mint"
+        info.pool shouldBe "ARB.GLD-0xAFD091"
+        // No destination on an add-liquidity mint — the pool above is the target, so the done
+        // screen hides the "To" row rather than rendering an empty value (issue #5351).
```
rkokhatskyi: This comment should be trimmed to what's tested because it describes the Compose hasDestination gate in SendTxOverviewScreen.kt, but this test only asserts toUiTransactionInfo() output, so it overstates what's verified here.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
aminsato: Good catch. `loadTransaction` runs after `checkIsVaultCorrect`/`switchToCorrectVault` may have reassigned `_currentVault`, and the deposit builder was the only one still passing the stale route `vaultId`. Changed the call site to `joinDepositUiModelBuilder.build(payload, _currentVault.id)` so it matches the swap builder (`vault = _currentVault`) and send builder (`srcVaultName = _currentVault.name`) — one From name on both verify and done screens after an auto-switch. Fixed in c9e745bd1.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
aminsato: Agreed — the up-front pass was clobbering the destination labels the joiner already resolved, and the async block only restored them a moment later, hence the flash. Added a `withResolvedSrcVaultName` helper that sets only the source vault name and leaves `dstVaultName`/`dstAddressBookTitle` untouched; the destination-less Mint LP case still gets its From label. Fixed in c9e745bd1.

app/src/test/java/com/vultisig/wallet/ui/screens/transaction/DepositUiTransactionInfoTest.kt
aminsato: Trimmed — the comment now only describes what `toUiTransactionInfo()` asserts (empty `to`, null `toLabel`) and no longer references the Compose row-hiding gate. Fixed in c9e745bd1.

johnnyluo (approved):
LGTM

PR #5359, opened 2026-07-21
fix(solana): count swap priority fee once, not twice

### LINKED ISSUE #5347: [Fix] Solana swap fee estimate double-counts the priority fee, showing roughly double the real cost
# [Fix] Solana swap fee estimate double-counts the priority fee, showing roughly double the real cost

## Problem
The default fee calculation used for Solana swap estimates adds a freshly-computed priority fee on top of a base-fee constant that already includes a priority fee at its floor value. In the common case this makes the displayed swap fee come out to roughly double the real fee.

## Expected Behavior
The Solana swap fee estimate should show the actual base fee plus the priority fee computed once, not the priority component counted twice.

## Steps to Reproduce
1. Request a Solana-source swap quote under normal (non-congested) network conditions.
2. Compare the displayed network fee against the real base signature fee plus a single priority-fee component.
3. Observe the displayed fee is roughly double what it should be, because the "base fee" constant already bundles the floor priority fee and a second priority fee is added on top.

## Solution
Give the swap default-fee calculation a genuine base-only constant and add the computed priority fee exactly once, instead of adding it on top of a constant that already includes a priority component. The existing unit test that currently pins the doubled value as "correct" needs its expected value (and its comment) corrected in the same change.

### Must Do
- [ ] Fix the swap default-fee calculation so the priority fee is only counted once
- [ ] Correct the existing test's expected value and comment to match the fixed (non-doubled) amount

### Must NOT Do
- Don't change the already-correct non-swap fee calculation path
- Don't change how the priority fee itself is fetched/computed

## Acceptance Criteria
- [ ] Updated test asserts the corrected (non-doubled) fee amount

johnnyluo:
## Summary
- `SolanaFeeService.calculateDefaultFees` (used for Solana swap fee estimates) added a freshly-computed priority fee on top of `DEFAULT_COIN_TRANSFER_BASE_FEE` (105000 lamports), which already bundled the base signature fee (5000) plus the floor priority fee (100000) — roughly doubling the displayed swap fee.
- Replaced that constant with a genuine base-only `BASE_SIGNATURE_FEE` (5000 lamports / 0.000005 SOL) so the priority fee is added exactly once.
- Corrected the regression test's expected value (`205_000` → `105_000`) and comment.
- Non-swap `calculateFees` path and priority-fee fetching left unchanged.

## Issue
Closes #5347

## Test Plan
- [x] `./gradlew :data:testDebugUnitTest --tests "*SolanaFeeServiceTest"` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Solana default fee estimation to use the base signature fee plus the computed priority component (instead of the prior base assumption).
  * Updated the swap default fee test expectation to match the corrected fee amount calculation.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/SolanaFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/SolanaFeeService.kt
@@ -195,13 +195,12 @@ internal constructor(
     }
 
     private companion object {
-        // Base signature fee only: 0.000005 SOL (5000 lamports) per signature. The computed
-        // priority
-        // fee is added on top exactly once in calculateDefaultFees. Expressed in lamports (SOL has
-        // 9
-        // decimals) without CoinType so the arithmetic can be exercised in JVM unit tests, which
-        // have
-        // no WalletCore native library loaded.
+        // Base signature fee for the single-signer swap estimate: 5000 lamports (0.000005 SOL).
+        // The swap path signs one Jupiter-built tx with the vault key, so one signature is assumed;
+        // the priority fee is added on top exactly once in calculateDefaultFees.
+        // Expressed in raw lamports (SOL has 9 decimals) rather than via CoinType so the fee
+        // arithmetic can be exercised in JVM unit tests, which have no WalletCore native lib
+        // loaded.
         val BASE_SIGNATURE_FEE: BigInteger =
             "0.000005".toBigDecimal().movePointRight(9).toBigInteger()
     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/SolanaFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/SolanaFeeService.kt
@@ -190,15 +190,19 @@ internal constructor(
         return GasFees(
             price = dynamicFeePrice,
             limit = SOLANA_PRIORITY_FEE_LIMIT.toBigInteger(),
-            amount = DEFAULT_COIN_TRANSFER_BASE_FEE + priorityAmount,
+            amount = BASE_SIGNATURE_FEE + priorityAmount,
         )
     }
 
     private companion object {
-        // 0.000105 SOL expressed in lamports (SOL has 9 decimals). Computed without CoinType so the
-        // service can be constructed and its fee arithmetic exercised in JVM unit tests, which have
+        // Base signature fee only: 0.000005 SOL (5000 lamports) per signature. The computed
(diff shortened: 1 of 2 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/SolanaFeeService.kt
```diff
@@ -190,15 +190,19 @@ internal constructor(
...
+        // priority
+        // fee is added on top exactly once in calculateDefaultFees. Expressed in lamports (SOL has
+        // 9
+        // decimals) without CoinType so the arithmetic can be exercised in JVM unit tests, which
+        // have
         // no WalletCore native library loaded.
-        val DEFAULT_COIN_TRANSFER_BASE_FEE: BigInteger =
-            "0.000105".toBigDecimal().movePointRight(9).toBigInteger()
+        val BASE_SIGNATURE_FEE: BigInteger =
+            "0.000005".toBigDecimal().movePointRight(9).toBigInteger()
```
aminsato, line 205: The comment describes this as a *per-signature* fee (5000 lamports/signature), but here it's added as a single flat amount. A Solana swap transaction can carry more than one required signature, in which case the on-chain base fee is `n * 5000`. Is the single-signature assumption guaranteed for the swap path this fee estimate covers? If a swap can have extra signers, this will under-estimate the base by 5000 lamports each.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/SolanaFeeService.kt
```diff
@@ -190,15 +190,19 @@ internal constructor(
...
             limit = SOLANA_PRIORITY_FEE_LIMIT.toBigInteger(),
-            amount = DEFAULT_COIN_TRANSFER_BASE_FEE + priorityAmount,
+            amount = BASE_SIGNATURE_FEE + priorityAmount,
         )
     }
 
     private companion object {
-        // 0.000105 SOL expressed in lamports (SOL has 9 decimals). Computed without CoinType so the
-        // service can be constructed and its fee arithmetic exercised in JVM unit tests, which have
+        // Base signature fee only: 0.000005 SOL (5000 lamports) per signature. The computed
```
aminsato: ktfmt has wrapped this comment mid-phrase, leaving `priority`, `9`, and `have` dangling on their own lines, which makes it hard to read. Reflowing to whole clauses per line (e.g. break after `per signature.` and after `added on top exactly once in calculateDefaultFees.`) keeps the intent legible.

rkokhatskyi (approved):
LGTM

PR #5360, opened 2026-07-21
fix(cardano): recover loser device on duplicate-broadcast rejection

### LINKED ISSUE #5337: Cardano multi-device keysign: loser device shows 'Transaction failed' though swap succeeds on-chain
### Summary

On a multi-device Cardano (ADA) keysign — reproduced with an ADA→ETH SwapKit swap — the two signing devices disagree on the outcome. The device that wins the broadcast race submits the transaction successfully and shows **Transaction complete**. The other device tries to broadcast the *same* signed transaction, is rejected by the node with `All inputs are spent. Transaction has probably already been included`, fails duplicate-broadcast recovery, and shows **Transaction failed** — even though the transaction (and the swap) actually completed on-chain.

So the user sees a scary "Transaction failed" error on one device for a transaction that in reality succeeded.

### Steps to reproduce

1. Two-device vault.
2. Start an ADA → ETH swap (SwapKit, NEAR provider) and co-sign it from both devices.
3. Both devices finish keysign and both broadcast the identical signed CBOR to the Ogmios endpoint.

### Actual behavior

- **Device A (winner):** `submitTransaction` → 200, swap tracked to `completed`, screen shows **Transaction complete / Successful**.
- **Device B (loser):** `submitTransaction` → **400**:
  ```
  {"error":{"code":3997,"message":"The transaction couldn't be added to the mempool...",
   "data":{"error":"All inputs are spent. Transaction has probably already been included"}}}
  ```
  App throws `Failed to broadcast transaction` and shows **Transaction failed – "One of your devices didn't respond in time. Check your connection and try again."**

### Expected behavior

Both devices should report success when the transaction actually lands on-chain. A duplicate-broadcast rejection on the losing device of a multi-device keysign is expected and must be recovered into a success, exactly as the other chains do.

### Root cause

The duplicate-broadcast recovery in `BroadcastTxUseCase` exists for Cardano, but its verify predicate is too strict for the moment it runs:

`data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt` (Cardano branch):

```kotlin
verify = { hash -> (cardanoApi.getTxStatus(hash)?.numConfirmations ?: 0) > 0 },
```

Recovery (`recoverIfAlreadyBroadcast` → `isAlreadyOnChain`) only retries **3 times with a 2s backoff (~4–6s total)**. Right after the winning device broadcasts, the tx is in the mempool but has **0 confirmations** — a Cardano block takes ~20s. So every verify attempt sees `num_confirmations: 0`, `isAlreadyOnChain` returns `false`, and the original 400 is re-thrown as a hard failure.

This is exactly what the logs show: after the 400, the loser polls Koios `tx_status` repeatedly and keeps getting `num_confirmations: 0`, then gives up and surfaces the error.

Two compounding problems:
1. **Verify requires `> 0` confirmations** but the recovery window (~5s) is far shorter than a Cardano block time (~20s), so a genuinely-accepted tx can't be confirmed in time.
2. The node's own response — `All inputs are spent. Transaction has probably already been included` — is itself a strong "already broadcast" signal that is currently ignored; only the confirmation count is consulted.

Note the `> 0` requirement was chosen deliberately because Koios `tx_status` echoes the queried hash back even for unknown txs (so a returned row alone doesn't prove inclusion). Any fix needs to keep distinguishing "known/accepted" from "unknown" without waiting a full block.

### Suggested direction (for discussion)

- Treat the `All inputs are spent / already been included` Ogmios error on the loser device as a duplicate-broadcast signal and recover to the locally computed hash, and/or
- Extend the Cardano verify window (more attempts / longer backoff) so a 0-conf-but-accepted tx has time to confirm, and/or use a mempool-aware check that can distinguish "accepted, pending" from "unknown".

### Environment

- Branch context: relates to the recent Cardano swap-fee / initiator-vs-joined-device work (#5321, #5329/#5330).
- Chain: Cardano (ADA), SwapKit swap via NEAR provider.
- Device B stack trace: `CardanoApiImpl.broadcastTransaction(CardanoApi.kt:129)` → re-wrapped at `CardanoApi.kt:140` → surfaced by `KeysignViewModel`.

*(Wallet addresses and Cloudflare trace IDs redacted from the attached logs.)*

johnnyluo:
## Summary
On a multi-device Cardano (ADA) keysign, the device that loses the broadcast race rebroadcasts the peer's byte-identical signed tx. Ogmios rejects it with `All inputs are spent. Transaction has probably already been included`. The old recovery required `num_confirmations > 0`, but a Cardano block takes ~20s — far longer than the ~5s verify window — so the accepted tx never confirmed in time and the loser device showed **Transaction failed** for a swap that actually succeeded on-chain.

This treats that Ogmios error as the duplicate-broadcast signal it is:
- `CardanoApi.broadcastTransaction` throws a typed `CardanoTransactionAlreadyBroadcastException` when Ogmios's `data.error` reports the inputs are already spent / already included (surfaced past the outer catch, like `CancellationException`).
- The Cardano branch of `BroadcastTxUseCase` catches it and recovers to the locally computed hash — no need to wait a full block for a confirmation that can't arrive in the verify window. Since the peer's tx is byte-identical, its spent inputs prove our hash is canonical.
- Genuine rejections (`unknownOutputReferences`, other errors) still throw and fall through to on-chain verification unchanged.

On chain tx link: https://cardanoscan.io/transaction/077d743bd64fe0a966f435d1653d71d0a44e2b663d20bf9fa9c74a98992e5efb
## Issue
Closes #5337

## Test Plan
- [x] New unit tests pass (`./gradlew :data:testDebugUnitTest --tests "*CardanoApiBroadcastTest" --tests "*BroadcastTxUseCaseTest"`)
  - `CardanoApiBroadcastTest`: spent-inputs 400 → `CardanoTransactionAlreadyBroadcastException`; `unknownOutputReferences` and other 400s still throw `IllegalStateException`
  - `BroadcastTxUseCaseTest`: duplicate-broadcast rejection recovers the local hash (no `getTxStatus` wait); non-duplicate rejection propagates
- [ ] Manual: ADA→ETH SwapKit swap co-signed from two devices — both report success

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Cardano transaction broadcasting by recognizing duplicate-broadcast and “already been included/inputs already spent” rejections.
  * Duplicate-broadcast responses now emit a dedicated already-broadcast signal so the use case can recover using the locally computed transaction hash when available.
  * Hardened local transaction hash calculation to “fail closed” (returns blank) when CBOR parsing fails.
  * Improved parsing of additional Ogmios error details from bad-request responses.

* **Tests**
  * Added/extended coverage for spent-input, duplicate-broadcast, and transaction-hash parsing failure scenarios.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
@@ -117,8 +117,17 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Cardan
 
                     ogmiosResponse.result?.transaction?.id
                         ?: run {
-                            val errorMessage = ogmiosResponse.error?.message ?: "Unknown error"
+                            // Ogmios can convey a submission error under HTTP 200. Mirror the 400
+                            // handling so a duplicate-broadcast rejection recovers regardless of
+                            // the
+                            // status Ogmios uses to report it.
+                            val dataError = ogmiosResponse.error?.data?.error
+                            val errorMessage =
+                                dataError ?: ogmiosResponse.error?.message ?: "Unknown error"
                             Timber.e("Cardano transaction submission failed: $errorMessage")
+                            if (dataError?.isAlreadyBroadcast() == true) {
+                                throw CardanoTransactionAlreadyBroadcastException(errorMessage)
+                            }
                             error("Failed to broadcast transaction: $errorMessage")
                         }
                 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
@@ -24,6 +24,14 @@ import kotlinx.serialization.json.buildJsonObject
 import kotlinx.serialization.json.put
 import timber.log.Timber
 
+/**
+ * The Ogmios node rejected our broadcast because the inputs are already spent — its own message
+ * ("Transaction has probably already been included") is the duplicate-broadcast signal on the
+ * losing device of a multi-device keysign. Since the peer's byte-identical tx spent those exact
+ * inputs, our locally computed hash is canonical (issue #5337).
+ */
+class CardanoTransactionAlreadyBroadcastException(message: String) : Exception(message)
+
 interface CardanoApi {
     suspend fun getBalance(coin: Coin): BigInteger
 
@@ -124,8 +132,12 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Cardan
                     // only treat a real duplicate submission as success.
                     val ogmiosError =
                         json.decodeFromString<OgmiosTransactionResponse>(response.bodyAsText())
-                    val errorMessage = ogmiosError.error?.message ?: "Unknown error"
+                    val dataError = ogmiosError.error?.data?.error
+                    val errorMessage = dataError ?: ogmiosError.error?.message ?: "Unknown error"
                     Timber.e("Cardano transaction submission failed: $errorMessage")
+                    if (dataError?.isAlreadyBroadcast() == true) {
+                        throw CardanoTransactionAlreadyBroadcastException(errorMessage)
+                    }
                     error("Failed to broadcast transaction: $errorMessage")
                 }
 
@@ -136,10 +148,17 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Cardan
             }
         } catch (t: Throwable) {
             if (t is CancellationException) throw t
+            if (t is CardanoTransactionAlreadyBroadcastException) throw t
             Timber.e(t, "Failed to broadcast Cardano transaction")
             error("Failed to broadcast transaction : ${t.message}")
         }
     }
+
+    // Ogmios reports spent inputs as "All inputs are spent. Transaction has probably already been
+    // included" — the loser of a duplicate-broadcast race, not a genuine rejection.
+    private fun String.isAlreadyBroadcast(): Boolean =
+        contains("already been included", ignoreCase = true) ||
+            contains("inputs are spent", ignoreCase = true)
 
     private suspend fun getCurrentSlot(): ULong {
         val response = httpClient.get(url) { url { path(apiV1Path, "tip") } }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
@@ -5,6 +5,7 @@ import com.vultisig.wallet.data.api.models.cardano.CardanoSlotResponseJson
 import com.vultisig.wallet.data.api.models.cardano.CardanoTxStatusResponseJson
 import com.vultisig.wallet.data.api.models.cardano.CardanoUtxoRequestJson
 import com.vultisig.wallet.data.api.models.cardano.CardanoUtxoResponseJson
+import com.vultisig.wallet.data.api.models.cardano.OgmiosError
 import com.vultisig.wallet.data.api.models.cardano.OgmiosTransactionResponse
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.payload.UtxoInfo
@@ -50,6 +51,11 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Cardan
     private val url: String = "https://api.koios.rest"
     private val apiV1Path: String = "api/v1"
     private val ogmiosUrl = "https://api.vultisig.com/ada/"
+
+    private companion object {
+        // Ogmios "UnknownOutputReference": the tx spends inputs the ledger no longer knows.
+        const val OGMIOS_UNKNOWN_OUTPUT_REFERENCE_CODE = 3117
+    }
 
     override suspend fun getBalance(coin: Coin): BigInteger {
 
@@ -121,11 +127,11 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Cardan
                             // handling so a duplicate-broadcast rejection recovers regardless of
                             // the
                             // status Ogmios uses to report it.
-                            val dataError = ogmiosResponse.error?.data?.error
+                            val submitError = ogmiosResponse.error
                             val errorMessage =
-                                dataError ?: ogmiosResponse.error?.message ?: "Unknown error"
+                                submitError?.data?.error ?: submitError?.message ?: "Unknown error"
                             Timber.e("Cardano transaction submission failed: $errorMessage")
-                            if (dataError?.isAlreadyBroadcast() == true) {
+                            if (submitError?.isAlreadyBroadcast() == true) {
                                 throw CardanoTransactionAlreadyBroadcastException(errorMessage)
                             }
                             error("Failed to broadcast transaction: $errorMessage")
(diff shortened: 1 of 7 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
```diff
@@ -124,8 +132,12 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Cardan
                     // only treat a real duplicate submission as success.
                     val ogmiosError =
                         json.decodeFromString<OgmiosTransactionResponse>(response.bodyAsText())
-                    val errorMessage = ogmiosError.error?.message ?: "Unknown error"
+                    val dataError = ogmiosError.error?.data?.error
+                    val errorMessage = dataError ?: ogmiosError.error?.message ?: "Unknown error"
                     Timber.e("Cardano transaction submission failed: $errorMessage")
+                    if (dataError?.isAlreadyBroadcast() == true) {
```
aminsato: The duplicate-broadcast detection lives only in the `BadRequest` branch, but the `HttpStatusCode.OK` branch above (line 120) also decodes and reports an error body — so the code already assumes Ogmios can surface a submission error under HTTP 200 (JSON-RPC-over-HTTP commonly does). If the spent-inputs rejection ever arrives as `200 OK` with an error object instead of `400`, the loser device falls into the `?: run { … error(…) }` path and hard-fails again — the exact bug this PR fixes, left unhandled on one of the two error paths.

Worth mirroring the check in the OK branch, e.g. read `ogmiosResponse.error?.data?.error` and throw `CardanoTransactionAlreadyBroadcastException` there too, so recovery is independent of which status Ogmios uses to convey the rejection.

aminsato:
Please attach a Cardano transaction link

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
```diff
@@ -136,11 +157,21 @@ constructor(private val httpClient: HttpClient, private val json: Json) : Cardan
...
+            if (t is Error) throw t
             Timber.e(t, "Failed to broadcast Cardano transaction")
             error("Failed to broadcast transaction : ${t.message}")
         }
     }
 
+    // Ogmios reports spent inputs as "All inputs are spent. Transaction has probably already been
+    // included" — the loser of a duplicate-broadcast race, not a genuine rejection.
+    private fun String.isAlreadyBroadcast(): Boolean =
+        contains("already been included", ignoreCase = true) ||
```
rkokhatskyi: isAlreadyBroadcast matches only the post-inclusion text and never reads error.code, missing Ogmios's already-in-mempool rejection (code 3117) that iOS's CardanoService.swift:200 and Windows's cardano.js:21 both check as the primary duplicate-broadcast signal against this same backend, so the more likely near-simultaneous race still falls through unhandled

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -224,9 +225,17 @@ constructor(
...
+                        try {
+                            cardanoApi
+                                .broadcastTransaction(chain.name, tx.rawTransaction)
+                                .orKnownHash(tx)
+                        } catch (e: CardanoTransactionAlreadyBroadcastException) {
+                            // Ogmios says the inputs are already spent by the peer's byte-identical
+                            // tx, so our locally computed hash is canonical. A Cardano block takes
+                            // ~20s, far longer than the verify window, so this node signal is the
+                            // only timely proof the tx landed (issue #5337).
+                            tx.transactionHash.takeIf { it.isNotBlank() } ?: throw e
```
rkokhatskyi, line 237: This returns tx.transactionHash with zero on-chain check, but that hash comes from CardanoHelper.kt:249's hand-rolled CBOR walker, which silently falls back to hashing the full envelope on any parse error. That same parser already broke once in 965aebc72, so a fault here now ships as a permanent phantom success

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
```diff
@@ -224,9 +225,17 @@ constructor(
...
                     tx = tx,
                     broadcast = {
-                        cardanoApi
-                            .broadcastTransaction(chain.name, tx.rawTransaction)
-                            .orKnownHash(tx)
+                        try {
+                            cardanoApi
+                                .broadcastTransaction(chain.name, tx.rawTransaction)
+                                .orKnownHash(tx)
+                        } catch (e: CardanoTransactionAlreadyBroadcastException) {
```
rkokhatskyi, line 232: This catch means the shared verify callback below never runs for this case, so a wrong hash here can never resolve. CardanoStatusProvider returns NotFound forever and KeysignViewModel maps that to StillConfirming, leaving the tx or any SwapKit swap stuck pending instead of showing a failure

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/api/CardanoApi.kt
johnnyluo: Added in ac642c52b — `isAlreadyBroadcast` now also matches `error.code == 3117`, so Android lines up with iOS and the shared SDK against this backend.

One correction for the record: 3117 is not "already in mempool" — in the Ogmios spec it's `3117/UnknownOutputReference` (see the local-tx-submission error table; the iOS comment mislabels it, and Ogmios has no "already in mempool" code at all). The near-simultaneous race you're describing actually surfaces as `3997/UnexpectedMempoolError` with "All inputs are spent. Transaction has probably already been included" — the exact payload captured in #5337, which this PR already handled. 3117 is the *post-inclusion* loser: the winner's tx is in a block and the ledger no longer knows our inputs. Both stages of the race are now covered.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
johnnyluo: The fallback hash couldn't actually reach the network on this path, on three counts:

- The #5337 scenario is a swap, and `SwapKitCardanoSigner` computes its hash with **no fallback** — `blake2b256(parsed.body)` or signing throws.
- On the send path, `addIsValidFlag` walks the identical body bytes with the same walker *immediately before* the hash is computed, and it throws with no fallback. WalletCore emits 3-element envelopes today (that's why the splice exists), so any body-parse fault aborts signing before broadcast.
- 965aebc72 is precisely that fail-closed behavior in action: the walker bug **threw pre-broadcast** in `addIsValidFlag` — it never produced a wrong hash.

That said, the silent fallback was a landmine waiting for the day WalletCore emits 4-element envelopes directly (early-return, walker unexercised), so ac642c52b removes it: `calculateCardanoTransactionHash` now fails closed to blank, and every recovery path (`orKnownHash`, `recoverIfAlreadyBroadcast`, the new catch's `takeIf { isNotBlank() }`) rethrows on blank. A wrong hash can no longer be reported as success, structurally.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/BroadcastTxUseCase.kt
johnnyluo: Skipping verify here is the fix, not a gap. The verify predicate requires `num_confirmations > 0` inside a ~5s window (3 attempts × 2s) against a ~20s block time — it mathematically cannot succeed at the moment the loser gets rejected. That's the root cause of #5337: routing this case through verify is what produced the false "Transaction failed".

Both sibling platforms already ship exactly this behavior: iOS `CardanoService` returns the precomputed hash on 3117 with zero verification, and the SDK broadcast resolver does the same (`if (rpcErrorCode === 3117) return getCardanoTxHash(tx)`).

On the stuck-StillConfirming scenario: it requires the wrong hash from your other comment, which ac642c52b makes impossible (hash-parse failure now yields blank → the catch rethrows the original rejection instead of recovering). With a correct hash and a true duplicate, the peer's tx is byte-identical → same body → same txid, so status tracking resolves as soon as the block lands. One block of "confirming" is the accurate state — and strictly better than the guaranteed-false failure it replaces.

Issue #5335, opened 2026-07-19
Swap: "Total Fee" understates real swap cost (excludes slippage), unreconcilable with explorer or value delta


aminsato:
### Summary

On the swap **Transaction complete** screen, the **Total Fee** value does not reconcile with either the on-chain network fee **or** the actual value the user lost in the swap. Second cross-chain data point for the fee-composition problem in #5334, this time showing the opposite direction of error (Total Fee **understates** the real cost).

### Concrete example (LTC → ADA swap)

- App **Total Fee:** `$0.00051`
- Actual on-chain LTC network fee for the same transaction: **348 litoshi = 0.00000348 LTC ≈ $0.00016**
- App **From:** `0.09442129 LTC` = `$4.45`  →  **To:** `26.428548 ADA` = `$4.38`
  - Value actually lost in the swap ≈ **$0.07**

So on this screen:
- Total Fee ($0.00051) is ~3× the on-chain network fee ($0.00016), **and**
- Total Fee ($0.00051) is ~137× smaller than the value the user actually gave up ($0.07).

The user cannot reconcile "$0.00051" with anything they can observe — not the block explorer fee, not the from/to value delta.

### Why

As traced in #5334, `totalFee = feesFiatForTotal + gasFeeFiatValue`. The liquidity/slippage component is intentionally excluded (it is baked into the received amount). For this route that excluded component is essentially the entire real cost (~$0.07), so the displayed Total Fee collapses to a near-network-fee number that badly understates what the swap cost the user.

Reference: `app/src/main/java/com/vultisig/wallet/ui/models/mappers/SwapTransactionToUiModelMapper.kt` (`feesFiatForTotal + gasFeeFiatValue`), rendered in `TransactionDoneScreen.kt`.

### Transaction facts (match the screenshot)

- Sender matches the app's **From** address; on-chain outputs go to a swap-provider inbound address + change back to sender (not the ADA `addr1v855…` destination) — expected for a cross-chain swap.
- Status "In progress…" — the ADA outbound leg had not settled yet.

### Expected / suggested

- Total Fee should be reconcilable with what the user can see. Either:
  - include the slippage/liquidity/price-impact component so Total Fee ≈ the from→to value delta, or
  - relabel it (e.g. "Network + Swap Fee") and add a separate "Price impact / slippage" line so the ~$0.07 is accounted for.
- At minimum, show a breakdown so the number is explainable.

### Notes

[...]

Issue #5334, opened 2026-07-19
Swap: "Total Fee" bundles provider fee + network fee, doesn't match on-chain network fee


aminsato:
# [Fix] Clarify "Total Fee" as composite of Swap Fee + Network Fee on Swap completion screen [Swap → Transaction complete]

## Problem
On the swap **Transaction complete** screen, the **Total Fee** value is a composite of the swap-provider fee **plus** the on-chain network (gas) fee, but its label reads as if it were a single/network fee. Because the same screen also shows a separate **Network Fee** row (in a token amount), users cross-checking Total Fee against the actual on-chain fee in a block explorer see a large, unexplained gap. This is a clarity/labeling/reconciliation problem, not necessarily a math bug — the composite value appears legitimate.

## Expected Behavior
- It is explicit that **Total Fee = Swap Fee + Network Fee** (e.g. via a tooltip or an on-screen breakdown), so the composite is discoverable.
- Fee rows (Swap Fee / Network Fee / Total Fee) use consistent fiat values so they visibly add up on screen, rather than mixing a token amount (Network Fee) with fiat (Total Fee).
- The semantics of "Total Fee" vs "Max. Total Fee" are confirmed and consistent across the overview/verify screens and the completion screen.

## Steps to Reproduce
1. Perform a swap across chains (e.g. XRP → ATOM).
2. On the **Transaction complete** screen, note the **Total Fee** value (example: `$0.28`).
3. Note the separate **Network Fee** row, shown as a token amount (example: `0.0004 XRP`).
4. Look up the same transaction in a block explorer and compare the on-chain network/burn fee (example: 400 drops = 0.0004 XRP ≈ `$0.0004`).
5. Observe that Total Fee (`$0.28`) is ~640× the real on-chain network fee and cannot be reconciled — because it silently also includes the swap-provider fee.

### Must Do
- [ ] Make the composite explicit on the completion screen — surface that Total Fee = Swap Fee + Network Fee (tooltip and/or breakdown rows) in `SwapTransactionDetail` within `TransactionDoneScreen.kt`.
- [ ] Present fee rows in consistent units (fiat) so Swap Fee + Network Fee visibly sum to Total Fee, rather than mixing token amount and fiat.
- [ ] Confirm and align the "Total Fee" vs "Max. Total Fee" labels across the verify/overview screens and the completion screen.

### Must NOT Do
- Don't change unrelated code.
[...]

PR #5364, opened 2026-07-21
fix(swap): correct Swap Fee display across form, verify, and join

### LINKED ISSUE #5358: Swap fee: 1inch route shows gas as "Swap Fee"; verify screen drops percentage and tier discount
## Context

Android's swap **form** is the reference implementation for native routes. `SwapQuoteManager.kt:325-373` itemizes `fees.affiliate` as `"Swap Fee (0.50%)"`, separates outbound, and — uniquely among the clients — excludes the liquidity component from the headline total so the visible rows actually sum (`:360-367`). Keep that. Three gaps remain.

## 1. Pure 1inch shows the wrong number (MUST FIX)

`SwapQuoteManager.kt:801-856` never reads `apiQuote.tx.swapFee`. It assigns `gasPrice × gas` to the fee row, so the user sees **gas cost labeled "Swap Fee"** — duplicating the Network Fee row and misrepresenting what Vultisig charges.

Kyber, LI.FI, Jupiter and SwapKit all correctly call `resolveSwapFee(...)`. Only the pure-1inch branch is wrong.

## 2. Verify screen degrades

`VerifySwapScreen.kt:289-292` uses `swap_form_estimated_fees_title` ("Swap Fee") instead of the `_with_percent_` variant at `strings.xml:252` (`swap_form_estimated_fees_with_percent_title = "Swap Fee (%1$s)"`). The file contains **no discount or tier row at all** — the user cannot tell their VULT tier was applied at the moment of approval.

## 3. Aggregator routes are opaque

`rawFees == null` (`:325-330`, `:368-372`) collapses the display to one unlabeled fee — no percentage, no outbound row, no discount rows.

## Already correct — do not regress

At Ultimate tier the row **stays visible** showing `Swap Fee (0.00%) $0.00` (`FiatValueToStringMapper.kt:49-50` formats `<= 0` as standard currency, so the `isNotBlank()` guard at `SwapFeeBreakdown.kt:172` passes). This is the target behavior for all clients — iOS currently hides the row instead.

## Blocked by

vultisig/vultisig-sdk#1450 — affiliate and protocol fees as separate fields on the SDK fee surface.

## Acceptance criteria

- [ ] 1inch route populates the fee row from `tx.swapFee` — never gas. If 1inch does not return it, render `Vultisig Fee (0.50%) — included in quoted rate`
- [ ] Verify screen uses `swap_form_estimated_fees_with_percent_title` and renders the VULT tier + referral rows, matching the form
- [ ] Aggregator routes render `Vultisig Fee (X.XX%)` using the `affiliate_bps` sent on the quote request
- [ ] Percentage sourced from sent bps rather than the client-side fiat division at `SwapQuoteManager.kt:348-357`, which returns `null` when source fiat is unavailable
- [ ] Keep existing Ultimate-tier behavior (`$0.00` row stays visible)

## Supersedes / relates to

Supersedes #5080 (`[Verify] Swap fee display correctness across all providers`, currently `agent:blocked`) — that's a verification checklist; this is the build spec. Motivating symptom reports: #5334, #5335 (Total Fee unreconcilable with on-chain).

aminsato:
## What & why

Fixes #5358 — the swap **Swap Fee** row was wrong or incomplete across the form, verify, and co-signer screens.

### 1inch showed gas as the Swap Fee (and double-counted it)
`fetchOneInchQuote` put `gasPrice × gas` in the fee slot, so gas was mislabeled as the Swap Fee — duplicating the Network Fee row and doubling it in the total (motivating #5334, #5335). 1inch never returns the affiliate fee separately (it is baked into the quoted rate), so the row now reads **"included in quoted rate"** and adds nothing to the total.

### Verify screen dropped the percentage and tier discount
The verify screen showed a bare "Swap Fee" — no percentage, no VULT tier / referral rows — so the user couldn't tell their tier was applied at the moment of approval. It now carries the form's fee/discount context and renders the same breakdown.

### Percentage sourced from the sent bps
The Swap Fee % is the net affiliate bps actually sent (base 50 bps − VULT discount, clamped ≥ 0) instead of a client-side fiat division that returned null when source fiat was unavailable. Uniform across providers: `0.50%` normally, `0.00%` at the Ultimate tier (row stays visible).

### SwapKit UTXO double-count (found during device testing)
For a SwapKit BTC/Cardano deposit the deposit cost *is* the Network Fee. The form already hides the Swap Fee row for these; verify and the join co-signer didn't, inflating the total (e.g. **$0.51 → $1.04**). They now hide it too.

## Screenshots

**Verify — percentage + VULT tier row now shown (SwapKit ONEINCH)**
| Form | Verify |
|--------|--------|
| form | verify |

**SwapKit UTXO (BTC) — Swap Fee correctly hidden, total = network fee ($0.51)**
| Form | Verify |
|--------|--------|
| form | verify |

**Join co-signer — matches the initiator (fee itemized for aggregator, hidden for UTXO)**
| ETH→USDC | BTC→ETH |
|--------|--------|
| cosigner-eth | cosigner-btc |

## Test plan
- New unit tests: `formatAffiliatePercent` cases; 1inch included-in-rate (fee $0, total = network); percentage reflects VULT discount; THORChain % from bps; verify-mapper 1inch total is network-only; SwapKit-UTXO hides the row + total stays at network fee. All green (`:app` + `:data` `testDebugUnitTest`).
- Device-verified on initiator **and** join co-signer across SwapKit-aggregator and SwapKit-UTXO routes (screenshots above) — totals reconcile, no double-counts.
- Localized `swap_form_estimated_fees_included_in_rate` across all 10 locales.

Closes #5358

## Summary by CodeRabbit

* **New Features**
  * Swap fee breakdowns now indicate when fees are **included in the quoted rate**, with consistent messaging across supported languages.
  * The swap verify screen now shows **VULT tier** and **referral** discount rows, in addition to the updated fee details.
* **Bug Fixes**
  * Fixed swap total and fee-row calculations to prevent **double-counting**, including **SwapKit UTXO/Cardano** scenarios where the swap-fee row may be hidden.
  * Improved “Swap Fee” presentation for providers that **bake fees into the rate**, ensuring totals and labels stay aligned.


app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
```diff
@@ -287,12 +299,11 @@ constructor(
...
-                        totalFee =
-                            fiatValueToStringMapper(
-                                estimatedFee + networkGasFeeFiatValue,
-                                asFee = true,
-                            ),
+                        totalFee = fiatValueToStringMapper(swapFeeForTotal, asFee = true),
                         provider = provider,
+                        swapFeeIncludedInRate = isOneInchIncludedInRate,
+                        swapFeePercent =
+                            if (isOneInchIncludedInRate) formatAffiliatePercent(null) else null,
```
rkokhatskyi: swapFeePercent disagrees with the initiator's own verify screen because this uses formatAffiliatePercent(null), always 0.50%, while SwapQuoteManager.kt:361 uses the real vultBPSDiscount, so the two co-signers see a different fee percent for the same swap

app/src/main/java/com/vultisig/wallet/ui/screens/swap/VerifySwapScreen.kt
```diff
@@ -286,10 +289,27 @@ private fun VerifySwapScreen(
...
 
-                    VerifyCardDetails(
-                        title = stringResource(R.string.swap_form_estimated_fees_title),
-                        subtitle = tx.providerFee.fiatValue,
-                    )
+                    // Swap Fee row mirrors the form (#5358): percentage in the title when known,
+                    // and "included in quoted rate" instead of a fiat amount for providers (1inch)
+                    // that bake the affiliate fee into the quoted destination amount. Hidden
+                    // entirely for a SwapKit UTXO deposit, whose cost is already the Network Fee.
+                    if (!tx.swapFeeHidden) {
```
rkokhatskyi, line 300: This screen now gates on swapFeeHidden and swapFeeIncludedInRate, but TransactionDoneScreen.kt:254 still renders providerFee.fiatValue unconditionally, so right after signing the same 1inch or SwapKit UTXO swap reverts to the pre-#5358 double-counted fee this PR just fixed here

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -328,9 +353,22 @@ constructor(
                 is SwapQuote.MayaChain -> quote.data.fees
                 else -> null
             }
+        // Displayed affiliate percentage sourced from the net bps actually sent to the provider —
+        // base 50 bps reduced by the VULT tier discount, clamped at zero — not the client-side
+        // fiat division this used to do, which returned null whenever source fiat was unavailable
+        // and rounded oddly (#5358). 0.50% with no discount; 0.00% at the Ultimate tier (full 50
+        // bps discount). Every provider charges the same base affiliate, so this is uniform.
+        val swapFeePercent = formatAffiliatePercent(vultBPSDiscount)
```
rkokhatskyi, line 358: This comment should be trimmed since it restates formatAffiliatePercent's own KDoc a few lines up almost verbatim and narrates the deleted fiat-division implementation, which is PR history that will read as meaningless once #5358 is old

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
```diff
@@ -245,6 +246,17 @@ constructor(
 
                 val estimatedFee = convertTokenValueToFiat(feeToken, estimatedTokenFees, currency)
 
+                // Literal 1inch bakes the affiliate fee into the quoted rate and never itemizes it,
+                // so the co-signer shows "included in quoted rate" — matching the initiator —
+                // rather
+                // than the gas placeholder the else-branch above computes for `value` (#5358,
+                // #5329). The join device can't know the initiator's VULT discount, so the row uses
+                // the base rate. Other providers carry a real fee and are unaffected.
```
rkokhatskyi: This comment breaks mid sentence with an orphaned rather on its own line and staples two ticket numbers, burying the one real point that the join device can't know the initiator's VULT discount, which is the actual cause of the bug two lines below

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
```diff
@@ -476,6 +487,11 @@ constructor(
                         providerFeeToken = srcToken,
                         currency = currency,
                         providerLabel = providerLabel,
+                        // BTC/Cardano SwapKit deposits report the deposit cost as their inbound
+                        // fee,
+                        // already shown as the Network Fee — hide the Swap Fee row so it isn't
+                        // double-counted, matching the initiator and the form (#5358, #5321).
+                        swapFeeHidden = srcToken.chain.standard == TokenStandard.UTXO,
```
johnnyluo: When `swapFeeHidden` is true, the `getSwapKitInboundFee` network fetch above (lines 458–477) is wasted — the fetched fee is neither displayed (row hidden) nor added to the total. Consider skipping the fetch for UTXO sources and passing a zero `TokenValue` instead; that's one fewer network round-trip on the co-signer's verify path.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -66,11 +66,36 @@ internal data class QuoteFetchResult(
...
     val priceImpact: BigDecimal? = null,
 )
 
+/**
+ * Base Vultisig affiliate fee, in basis points, charged before any VULT tier discount. Every swap
+ * provider quotes against the same 50 bps base (THORChain/Maya `AFFILIATE_FEE_RATE`, the aggregator
+ * `*_AFFILIATE_FEE_BPS` constants, and 1inch's 0.5% `referrer` fee), so the displayed rate is
+ * uniform.
+ */
+internal const val BASE_AFFILIATE_FEE_BPS = 50
```
johnnyluo, line 85: The claim that every provider quotes against the same 50 bps base is now load-bearing for the displayed percentage: if any per-provider `*_AFFILIATE_FEE_BPS` constant (or THORChain's `AFFILIATE_FEE_RATE`) diverges from 50 later, this constant silently goes stale and the row overstates/understates the fee. Consider deriving it from (or adding a `check` against) one of those constants so a divergence fails loudly.

Related: the percentage is now the *intended* net rate rather than the *observed* one — if a provider ever skips or reduces the affiliate (e.g. dust-sized swaps), the row will overstate. Net improvement over the fiat division (which returned null on missing source fiat), just worth being aware of.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
johnnyluo: This makes the join co-signer show the base "0.50%" while the initiator with a VULT tier may show e.g. "0.30%" — a visible cross-device discrepancy during the same signing ceremony. The comment documents why (the join device can't know the tier), but showing no percentage here — like the other providers on the join path — would avoid the mismatch entirely. Worth confirming product prefers the base rate over omission.

app/src/main/res/values-es/strings.xml
```diff
@@ -166,6 +166,7 @@
     <string name="swap_form_from_title">De</string>
     <string name="swap_form_estimated_fees_title">Tarifa de intercambio</string>
     <string name="swap_form_estimated_fees_with_percent_title">Tarifa de intercambio (%1$s)</string>
+    <string name="swap_form_estimated_fees_included_in_rate">incluida en la tarifa cotizada</string>
```
johnnyluo: Since the fee itself is "Tarifa de intercambio", "tarifa cotizada" reads closer to "quoted fee" than "quoted rate". Suggest `incluida en el tipo de cambio cotizado` (or `en la tasa cotizada`) to keep fee vs. rate unambiguous. The other nine locales read fine.

johnnyluo (commented):
Reviewed — solid, well-scoped fix; correct on all the paths I traced (the `swapFeeHidden` flag is properly scoped to the SwapKit branch, the tier derivation on verify matches the form, and the fee-display snapshot is race-free). Inline comments below; only the first two are worth acting on before merge, the rest are minor.

Also cosmetic: a few comments have formatter-orphaned words on their own line (e.g. "— rather / than" in JoinSwapUiModelBuilder.kt:251, "fee, / already" at :491, similar in SwapTransactionToUiModelMapper.kt) — a quick rewrap would help readability.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
aminsato: Fixed in 6a2af84b3. The join device can't recover the initiator's VULT tier discount from the signed payload (no net-bps field in the 1inch payload proto, and commondata is a submodule we don't touch here), so rather than show a *wrong* base 0.50% it now omits the percent entirely — keeping only "included in quoted rate", the same as every other provider on the join path. No more cross-device discrepancy.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/VerifySwapScreen.kt
aminsato: Good catch — fixed in 6a2af84b3. TransactionDoneScreen's Swap Fee row now mirrors the verify screen: hidden for a SwapKit UTXO deposit (cost already in the Network Fee) and "included in quoted rate" for 1inch, instead of rendering providerFee.fiatValue unconditionally. So the post-signing done screen no longer reverts to the double-counted fee.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Trimmed in 6a2af84b3 — dropped the KDoc restatement and the deleted-fiat-division narration; it now just points at formatAffiliatePercent.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
aminsato: Rewrapped in 6a2af84b3 — no more orphaned "rather", and the comment now leads with the real point (the join device can't know the initiator's VULT discount).

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
aminsato: Done in 6a2af84b3. Computed swapFeeHidden up front and skip the getSwapKitInboundFee /v3/quote call entirely for UTXO sources — the fetched fee is neither displayed (row hidden) nor added to the total, so it was a pure wasted round-trip. Passes a zero TokenValue in that path.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Addressed in 6a2af84b3. The aggregator KYBER/SWAPKIT/JUPITER *_AFFILIATE_FEE_BPS constants now derive from BASE_AFFILIATE_FEE_BPS directly, and I added an init check that fails loudly if THORChain's AFFILIATE_FEE_RATE_BP (separate module) ever diverges from it. On the second point — agreed the displayed % is now the intended net rate rather than the observed one; noted as a known trade-off, and it's still strictly better than the old fiat division that returned null on missing source fiat.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
aminsato: Went with the omission you suggested (6a2af84b3): the join path now shows no percentage rather than the base 0.50%, matching the other providers on this path and avoiding the cross-device mismatch. True parity would need the net bps carried in the payload, which means a commondata proto change — out of scope for this PR.

app/src/main/res/values-es/strings.xml
aminsato: Reworded in 6a2af84b3 to "incluida en el tipo de cambio cotizado" so fee vs. rate stays unambiguous. Thanks for the locale check.

data/src/main/kotlin/com/vultisig/wallet/data/models/SwapTransaction.kt
```diff
@@ -40,6 +40,21 @@ sealed interface SwapTransaction {
...
+     * default to "unavailable", so a co-signer rebuilding the tx from the signed payload — which
+     * cannot know the initiator's tier — degrades gracefully to no discount rows.
+     */
+    val swapFeePercent: String?
+    /** True when the affiliate fee is baked into the quoted rate (1inch) — see [swapFeePercent]. */
+    val swapFeeIncludedInRate: Boolean
+    val vultBpsDiscount: Int?
+    val vultBpsDiscountFiatValue: String?
+    val referralBpsDiscount: Int?
+    val referralBpsDiscountFiatValue: String?
```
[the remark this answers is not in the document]
aminsato: Resolved in 6a2af84b3, but via omission rather than payload metadata. Carrying the affiliate %/discount through the keysign payload would require a new field in 1inch_swap_payload.proto, and commondata/ is a submodule we don't edit from this repo. So instead of transmitting the values, the join path now shows *no* percentage (keeping only "included in quoted rate"), which removes the wrong-0.50% discrepancy and matches the graceful-degradation contract already documented on SwapTransaction. True cross-device parity can be a follow-up in the commondata repo if product wants the exact net rate on the join device.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -128,6 +154,17 @@ constructor(
...
+    init {
+        // The displayed affiliate percentage (formatAffiliatePercent) assumes every provider quotes
+        // the same BASE_AFFILIATE_FEE_BPS. The aggregator constants derive from it directly; THOR's
+        // rate lives in another module, so guard it here — a future divergence fails loudly instead
+        // of silently overstating/understating the Swap Fee row (#5358 review).
+        check(THORChainSwaps.AFFILIATE_FEE_RATE_BP == BASE_AFFILIATE_FEE_BPS) {
+            "THORChain affiliate rate ${THORChainSwaps.AFFILIATE_FEE_RATE_BP} bps diverged from " +
+                "displayed base $BASE_AFFILIATE_FEE_BPS bps; update formatAffiliatePercent"
+        }
+    }
```
johnnyluo: This `check()` is comparing two `const val`s (`THORChainSwaps.AFFILIATE_FEE_RATE_BP` and `BASE_AFFILIATE_FEE_BPS`). Both get inlined as literal `50` at compile time, and `SwapQuoteManager` (app module) already depends on `THORChainSwaps` (data module), so Gradle's Kotlin incremental compiler recompiles this file whenever the data-module constant changes — the two sides of this comparison can never legitimately drift apart within a normal build. In practice this reduces to `check(50 == 50)`: permanently true, and in the one scenario where it *wouldn't* be (a broken incremental build reusing stale classes), the check itself would be reading the same stale inlined value it's meant to catch, so it can't actually detect that case either.

More importantly, `SwapQuoteManager` has no `@Singleton` scope — it's re-created (and this `init` re-run) every time the swap form screen opens. If these constants are ever legitimately changed together but a stale-build edge case does somehow slip one past, this throws `IllegalStateException` in `init`, crashing the swap screen for every user on open, in production — rather than failing at build/CI time.

Suggest replacing this with a unit test asserting `THORChainSwaps.AFFILIATE_FEE_RATE_BP == BASE_AFFILIATE_FEE_BPS` (same 'fails loudly on drift' guarantee, but it fails on the PR instead of on a user's device), or downgrading to a `Timber.e` warning if you want it to stay runtime-observable.

johnnyluo (changes requested):
The Swap Fee display fixes look solid overall (1inch included-in-rate, SwapKit UTXO double-count, discount-row extraction). Requesting changes on one item — see inline comment on SwapQuoteManager.kt.

rkokhatskyi (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Agreed on both counts — fixed in b901296fb by removing the `init` check and moving the assertion into a unit test.

You're right that it reduced to `check(50 == 50)`: both sides are `const val`s inlined at compile time, and since the app module already depends on `THORChainSwaps`, the incremental compiler recompiles this file whenever the data-module constant changes — so the two sides can't legitimately drift within a normal build, and in the stale-classes case the check would be reading the same stale inlined value it was meant to catch.

The scoping point is the more serious one, and I'd missed it: `SwapQuoteManager` isn't `@Singleton`, so `init` re-runs on every swap-form open — a spurious throw would crash the swap screen for every user on open, in production, which is a far worse failure mode than the drift it was guarding against.

Went with your first suggestion over the `Timber.e` downgrade: `SwapQuoteManagerTest` now has `every provider quotes the same base affiliate rate as the one displayed`, asserting `THORChainSwaps.AFFILIATE_FEE_RATE_BP shouldBe BASE_AFFILIATE_FEE_BPS`. Same fails-loudly-on-drift guarantee, but it fails on the PR instead of on a user's device. The aggregator `*_AFFILIATE_FEE_BPS` constants still derive from `BASE_AFFILIATE_FEE_BPS` directly, so THORChain's cross-module rate was the only one needing an explicit assertion.

johnnyluo (approved):
LGTM

PR #5368, opened 2026-07-21
fix(send): reject an invalid recipient with a visible error (#5361)

### LINKED ISSUE #5361: fix(send): reject an invalid recipient with a visible error across all chains
## Problem

A send flow can accept an invalid recipient into its state and leave the next action unavailable without explaining why. Recipient validation must be a first-class, chain-general user-facing gate, not an implicit failure later in the flow.

This surfaced while preparing a **USDC send on Ethereum**: a transaction ID was entered where the recipient address belongs. The app should explicitly say the value is not a valid Ethereum recipient address.

## Required behaviour

For every supported chain and send asset:

- validate the recipient as soon as it is captured or selected;
- show a clear inline error when invalid: `That recipient is not a valid address for this chain.` Include the expected format where useful;
- keep the next action unavailable only with that visible reason and a direct correction path;
- allow the user to replace the value and continue without restarting;
- never build, sign, or broadcast from an invalid recipient.

## Acceptance criteria

- [ ] Invalid-recipient UI and recovery are shared across native send flows, not limited to a chain-specific screen.
- [ ] Coverage includes an Ethereum/USDC send with a txid substituted for the address, plus representative Cosmos, UTXO, Solana and XRP-family invalid inputs.
- [ ] Each case renders a human-readable error and prevents transaction construction/signing.
- [ ] Valid replacement enables the flow normally.

aminsato:
## Summary

A native send flow could accept an invalid recipient and leave **Continue** unavailable without explaining why. This surfaced with a **USDC send on Ethereum** where a transaction id was pasted where the recipient address belongs.

The inline-error slot (`dstAddressError` → `AddressInputSection`) already existed end-to-end but nothing ever populated it. This wires it, chain-general, on the single shared send path.

- **`AddressManager`** exposes `invalidAddress: StateFlow<Boolean>` — `true` when a non-empty recipient fails chain validation (`chainAccountAddressRepository.isValid`) **and** can't resolve to a valid address (ENS/name), `false` while empty, resolving, valid, or decoded from an XRP X-address.
- **`SendFormGraph`** maps `invalidAddress` onto `dstAddressError`, so the address field renders the inline error for every native send chain.
- **`isContinueDisabled()`** now also gates on `dstAddressError != null`, so the flow can't proceed from an invalid recipient — the inline error is the visible reason, and correcting the field clears it and re-enables Continue.
- New string `send_error_invalid_recipient_address` = _"That recipient is not a valid address for this chain."_ added to all 10 locales.

Submit already rejected an invalid recipient (`DefaultSendStrategy`), kept as defense in depth, so nothing builds, signs, or broadcasts from one.

## Acceptance criteria

- [x] Invalid-recipient UI and recovery are shared across native send flows, not a chain-specific screen (single `AddressManager` / `DefaultSendStrategy` path).
- [x] Covers an Ethereum/USDC send with a txid substituted for the address, plus Cosmos, UTXO, Solana and XRP invalid inputs.
- [x] Each case renders a human-readable error and prevents transaction construction/signing.
- [x] Valid replacement enables the flow normally.

## Test plan

- Send → Ethereum → paste a txid into **Send to** → inline error shows, red border, **Continue** disabled → replace with a valid address → error clears, Continue enables.
- Same on Cosmos (ATOM), Bitcoin, Solana, XRP with junk input.

Unit tests: `AddressManagerTest` — 5 new cases (eth txid, 4-chain matrix, valid clears, empty clears, resolver-failure). **18/18 green**.

## Screenshot

| Ethereum send, txid pasted |
|--------|
| after |

Closes #5361

## Summary by CodeRabbit

* **Bug Fixes**
  * Added an explicit invalid-recipient error state in the send form, including correct clearing when the input becomes valid or is removed.
  * “Continue” is now blocked for invalid recipients, with correct exceptions for locked unbond flows.
  * Destination address “complete” indicators and amount section behavior no longer treat invalid recipients as complete.

* **Localization**
  * Added the “invalid recipient address” message to supported languages.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
@@ -61,8 +61,10 @@ internal class AddressManager(
     val isDstAddressComplete: StateFlow<Boolean> = _isDstAddressComplete.asStateFlow()
 
     // True when a non-empty recipient failed chain validation and couldn't resolve to a valid
-    // address. Drives the inline "not a valid address for this chain" error; false while empty,
-    // resolving, or valid. Chain-general: every send chain validates through the same path below.
+    // address. Drives the inline "not a valid address for this chain" error; false once the input
+    // is valid or empty. While name resolution is in flight the flag holds its previous value, so a
+    // standing error stays put until the new input resolves instead of blinking off and back on.
+    // Chain-general: every send chain validates through the same path below.
     private val _invalidAddress = MutableStateFlow(false)
     val invalidAddress: StateFlow<Boolean> = _invalidAddress.asStateFlow()
 
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
@@ -60,6 +60,12 @@ internal class AddressManager(
     private val _isDstAddressComplete = MutableStateFlow(false)
     val isDstAddressComplete: StateFlow<Boolean> = _isDstAddressComplete.asStateFlow()
 
+    // True when a non-empty recipient failed chain validation and couldn't resolve to a valid
+    // address. Drives the inline "not a valid address for this chain" error; false while empty,
+    // resolving, or valid. Chain-general: every send chain validates through the same path below.
+    private val _invalidAddress = MutableStateFlow(false)
+    val invalidAddress: StateFlow<Boolean> = _invalidAddress.asStateFlow()
+
     private val _onAddressValidated = MutableSharedFlow<Unit>(extraBufferCapacity = 1)
     val onAddressValidated: SharedFlow<Unit> = _onAddressValidated.asSharedFlow()
 
@@ -144,6 +150,7 @@ internal class AddressManager(
                     _dstAddressLabel.value = null
                 }
                 _resolvedDstAddress.value = addressStr
+                _invalidAddress.value = false
                 _onAddressValidated.tryEmit(Unit)
             }
             addressStr.isNotEmpty() -> {
@@ -155,6 +162,7 @@ internal class AddressManager(
             else -> {
                 _resolvedDstAddress.value = null
                 _dstAddressLabel.value = null
+                _invalidAddress.value = false
             }
         }
     }
@@ -186,6 +194,7 @@ internal class AddressManager(
             addressFieldState.setTextAndPlaceCursorAtEnd(decoded.classicAddress)
         }
         _resolvedDstAddress.value = decoded.classicAddress
+        _invalidAddress.value = false
         // Surface the pasted X-address as the label so Verify/Done show what the user entered.
         _dstAddressLabel.value = originalInput
         _onAddressValidated.tryEmit(Unit)
@@ -200,11 +209,13 @@ internal class AddressManager(
             if (chainAccountAddressRepository.isValid(chain, resolved)) {
                 _dstAddressLabel.value = addressStr
                 _resolvedDstAddress.value = resolved
+                _invalidAddress.value = false
                 addressFieldState.setTextAndPlaceCursorAtEnd(resolved)
                 _onAddressValidated.tryEmit(Unit)
             } else {
                 _resolvedDstAddress.value = null
                 _dstAddressLabel.value = null
+                _invalidAddress.value = true
             }
         } catch (e: CancellationException) {
             throw e
@@ -214,6 +225,7 @@ internal class AddressManager(
             Timber.w(e, "Failed to resolve address %s on %s", addressStr, chain)
             _resolvedDstAddress.value = null
             _dstAddressLabel.value = null
+            _invalidAddress.value = true
         }
     }
 }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormGraph.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormGraph.kt
@@ -436,6 +436,23 @@ internal class SendFormGraph(
             }
         }
         scope.launch {
+            addressManager.invalidAddress.collect { invalid ->
+                uiState.update {
+                    it.copy(
+                        dstAddressError =
+                            if (invalid) {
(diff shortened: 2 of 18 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormGraph.kt
```diff
@@ -435,6 +435,23 @@ internal class SendFormGraph(
                 uiState.update { it.copy(destinationTagLocked = locked) }
             }
         }
+        scope.launch {
+            addressManager.invalidAddress.collect { invalid ->
+                uiState.update {
+                    it.copy(
+                        dstAddressError =
```
rkokhatskyi, line 442: New `dstAddressError` can go invisible: `FoldableAmountWidget.onToggle` in `SendFormAmountSection.kt:103` collapses Address on `isDstAddressComplete` alone, not this field, so tapping Amount while invalid hides the error and shows a false complete checkmark, with Continue silently disabled.

app/src/main/res/values-de/strings.xml
```diff
@@ -193,6 +193,7 @@
     <string name="camera_permission_open_settings">Einstellungen öffnen</string>
     <string name="verify_transaction_value">Wert</string>
     <string name="send_error_no_address">Adresse ist ungültig</string>
+    <string name="send_error_invalid_recipient_address">Dieser Empfänger ist keine gültige Adresse für dieses Netzwerk.</string>
```
rkokhatskyi: This translates 'chain' as 'Netzwerk' but the app's established German term is 'Blockchain' (`error_invalid_chain`, `custom_rpc_select_chain`), which keeps a separate word for network elsewhere. The same swap from chain to network repeats in es/hr/it/ko/nl/pt/zh-rCN, only ru already uses network for chain app-wide.

app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
```diff
@@ -60,6 +60,12 @@ internal class AddressManager(
     private val _isDstAddressComplete = MutableStateFlow(false)
     val isDstAddressComplete: StateFlow<Boolean> = _isDstAddressComplete.asStateFlow()
 
+    // True when a non-empty recipient failed chain validation and couldn't resolve to a valid
+    // address. Drives the inline "not a valid address for this chain" error; false while empty,
+    // resolving, or valid. Chain-general: every send chain validates through the same path below.
```
rkokhatskyi: This comment claims the flag is false while resolving, but the `addressStr.isNotEmpty()` branch at line 156 never touches `_invalidAddress` before awaiting `tryResolveName`, so it silently keeps its prior value instead, contradicting the stated invariant.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormUiModel.kt
```diff
@@ -109,10 +109,15 @@ internal data class SendFormUiModel(
...
  *
+ * An invalid recipient ([dstAddressError] set) also disables Continue so the flow can't proceed
+ * from an address that failed chain validation — the inline error is the visible reason, and
+ * correcting the field clears it and re-enables Continue.
+ *
  * @return true when Continue must stay disabled.
  */
 internal fun SendFormUiModel.isContinueDisabled(): Boolean =
     isLoading ||
+        dstAddressError != null ||
```
rkokhatskyi: This also hard-gates the non-editable Unbond node-address field (`SendFormAddressSection.kt:271`, `isAddressEditable = false`), so if that prefilled address is ever flagged invalid, Continue is permanently blocked with no retype or paste path to recover.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormGraph.kt
aminsato: Confirmed and fixed in 4fe89bdcc. `FoldableAmountWidget.onToggle` now also requires `!state.isDstAddressBlocking`, so Address stays expanded while the error is the reason Continue is disabled. Both address widgets also pass `complete = state.isDstAddressComplete && state.dstAddressError == null`, so the checkmark no longer appears on a filled-in but invalid recipient.

app/src/main/res/values-de/strings.xml
aminsato: Right, fixed in 4fe89bdcc. Each locale now uses the term it already established for chain (checked against `error_invalid_chain` / `select_chain_chain_title` / `custom_rpc_select_chain` in each file): de Blockchain, es cadena, hr lanac, it catena, ko 체인, nl keten, pt cadeia, zh 链. ru is unchanged since сеть is its app-wide word for chain.

app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
aminsato: Correct — the comment was wrong, not the code. The `addressStr.isNotEmpty()` branch deliberately leaves the flag alone so a standing error doesn't blink off and back on while resolution is in flight; it settles on the resolver's outcome. Comment rewritten in 4fe89bdcc to say exactly that.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormUiModel.kt
aminsato: Good catch — that was a real dead end, since `AddressInputSection` also hides the paste/scan/address-book row when the field isn't editable. Fixed in 4fe89bdcc: `isDstAddressBlocking` (`dstAddressError != null && isDstAddressEditable`) limits the gate to a recipient the user can actually correct, so a locked Unbond node address no longer blocks Continue. The inline error still renders, and `UnbondStrategy.kt:85` still rejects an invalid node address at submit with `send_error_no_address`, so nothing invalid gets built or signed. `isDstAddressEditable` also replaced the inline `defiType != UNBOND` check at `SendFormScreen.kt:435` so the lock and the gate can't drift apart. Covered by the new `SendFormContinueGateTest`.

johnnyluo (approved):
LGTM

PR #5375, opened 2026-07-22
fix(solana): average two central samples for even-length priority-fee sets

### LINKED ISSUE #5344: [Fix] Solana priority-fee median calculation is wrong for even-length sample sets
# [Fix] Solana priority-fee median calculation is wrong for even-length sample sets

## Problem
When computing the median of recent prioritization-fee samples, the code picks the upper-middle element of the sorted list instead of averaging the two central elements when the sample count is even. This makes the computed priority fee skew higher than the true median whenever the RPC returns an even number of non-zero fee samples.

## Expected Behavior
The median should be computed correctly for both odd- and even-length sample sets — averaging the two central values when the count is even.

## Steps to Reproduce
1. Trigger a Solana priority-fee calculation when the RPC returns an even number of non-zero prioritization-fee samples (e.g. four samples: 100, 200, 800, 900).
2. Observe the computed median is the upper-middle sample (800) instead of the average of the two central samples (500).

## Solution
Fix the median calculation to average the two central elements when the sorted sample list has an even length, instead of indexing straight into the middle.

### Must Do
- [ ] Correct the median calculation for even-length sample sets
- [ ] Add a unit test covering both odd- and even-length sample sets

### Must NOT Do
- Don't change the fallback/clamp logic around the median calculation
- Don't change how the samples are fetched or filtered

## Acceptance Criteria

rkokhatskyi:
Fixes #5344

`getMedianPriorityFee` indexed straight into `nonZeroFees[size / 2]`, which is the upper-middle element of the sorted list. That's correct for an odd sample count but wrong for an even one — for `[100, 200, 800, 900]` it returned 800 instead of the true median (200+800)/2 = 500, skewing the priority fee higher than intended.

Extracted the calculation into `solanaMedianPriorityFee` and fixed the even-length case to average the two central elements, matching iOS (`SolanaService.fetchRecentPrioritizationFees`).

## Test plan
- New unit tests in `SolanaMedianPriorityFeeTest`: empty set, single sample, odd counts, even counts (including the ticket's exact 4-sample regression case), an even-sum-is-odd case to confirm truncation rather than rounding, and large `BigInteger` values.
- `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.*" --tests "com.vultisig.wallet.data.blockchain.solana.*"` — all green.
- `./gradlew :data:ktfmtCheck` — clean.

Note: `EthereumFeeService.kt` has the identical `getOrNull(size / 2)` pattern for its own priority-fee median, affecting Polygon and other EVM chains. Left out of this PR since the ticket is scoped to Solana; filed as #5376. iOS has the same EVM bug (`RpcEvmService.swift` / `EvmServiceStruct.swift`), filed as vultisig-ios#4878. iOS and Windows Solana priority-fee code already averages correctly for even-length sets, so no cross-platform fix is needed there.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
@@ -715,7 +715,7 @@ internal fun solanaMedianPriorityFee(
     if (size == 0) return fallback
     val mid = size / 2
     return if (size % 2 == 0) {
-        (sortedFees[mid - 1] + sortedFees[mid]) / BigInteger.TWO
+        (sortedFees[mid - 1] + sortedFees[mid]) / BigInteger.valueOf(2)
     } else {
         sortedFees[mid]
     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
@@ -224,7 +224,7 @@ internal class SolanaApiImp(
                     ?.filter { it > BigInteger.ZERO }
                     ?.sorted()
                     .orEmpty()
-            val median = if (nonZeroFees.isEmpty()) fallback else nonZeroFees[nonZeroFees.size / 2]
+            val median = solanaMedianPriorityFee(nonZeroFees, fallback)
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5378, opened 2026-07-22
feat(sui): decode dApp PTBs on the verify screen

### LINKED ISSUE #5372: [Fix] Sui dApp-supplied PTB verify screen shows only raw bytes — no decoding, essentially blind-signing
# [Fix] Sui dApp-supplied PTB verify screen shows only raw bytes — no decoding, essentially blind-signing

## Problem
When a Secure Vault co-signs a Sui dApp-supplied Programmable Transaction Block (PTB), the verify/co-sign screen (`SignSuiDisplayView`) renders only the sender address and the raw base64 BCS bytes — no decoding of the PTB's commands (MoveCall, TransferObjects, SplitCoins, etc.), inputs, or gas budget. A malicious or compromised dApp can construct a PTB that drains owned objects to an attacker address or invokes an arbitrary MoveCall, and the co-signing device shows nothing a human can meaningfully review before approving.

This is a regression against the feature's own written spec: the tracking issue for this exact feature (#4881) explicitly required "render the decoded PTB (commands / inputs / sender / gas) ... instead of a '0 SUI / no recipient' transfer card" as acceptance criteria. The merged PR shipped raw-bytes-only, justified as "mirrors `SignSolanaDisplayView`" — a claim that doesn't hold, since Solana's view does decode instruction summaries for the equivalent flow.

There is also no working compensating control: the Blockaid security scan wired to this same screen validates a different, fabricated transaction, not the real PTB. `SecurityScannerTransactionFactory.createSUISecurityScannerTransaction` builds a scanner payload that never sets `signSui`, so `SuiHelper.getPreSignedInputData` falls through to reconstructing a generic `Pay`/`PaySui` transfer from placeholder fields instead of taking the "sign the real PTB verbatim" branch — meaning the security badge shown on this screen never actually reflects the real PTB being signed.

## Expected Behavior
The verify/co-sign screen should decode and render the real PTB's commands, inputs, sender, and gas budget — matching iOS and Windows, which both already fully decode Sui PTBs for this exact flow. The Blockaid scan on this screen should validate the actual PTB the user is about to sign, not a reconstructed placeholder transaction.

## Steps to Reproduce
1. Initiate a Sui dApp transaction (e.g. a swap or any MoveCall) from a browser extension against a Secure Vault.
2. On the Android co-signing device, open the verify screen.
3. Observe only a sender address and a raw base64 blob — no command/recipient/amount decoding, and the security badge (if shown) does not reflect the actual signed bytes.

## Solution
Port a BCS `TransactionData::V1` parser (iOS/Windows already have a well-specified reference implementation) into pure Kotlin, and render Commands/Inputs/sender/gas in `SignSuiDisplayView`, falling back to the raw-bytes view only when parsing fails. Separately, fix `SecurityScannerTransactionFactory.createSUISecurityScannerTransaction` to set `signSui` on the scanner payload so the Blockaid scan validates the real PTB instead of a fabricated one.

### Must Do
- [ ] Add a BCS PTB decoder and render decoded commands/inputs/sender/gas on the Sui verify/co-sign screen
- [ ] Fall back to the existing raw-bytes view only when decoding fails
- [ ] Fix `createSUISecurityScannerTransaction` to set `signSui` so Blockaid scans the actual PTB, not a placeholder transfer

### Must NOT Do
- Don't change the underlying signing path (`SuiHelper.buildSignDirectInputData`) — this is a display/scan-input fix, not a signing-logic change
- Don't remove the raw-bytes fallback entirely — keep it for the case decoding fails

## Acceptance Criteria
- [ ] A MoveCall, TransferObjects, and SplitCoins PTB each render their real decoded contents on the verify screen
- [ ] The Blockaid scan result reflects the real PTB (verified via a test asserting `signSui` is set on the scanner-built payload)

rkokhatskyi:
## Summary

Closes #5372. The Sui verify/co-sign screen for a dApp-supplied Programmable Transaction Block (PTB) only showed the sender address and raw base64 bytes — no decoding of commands, inputs, or gas, so a co-signer had no meaningful way to review what they were about to sign. Separately, the Blockaid scan on that same screen validated a reconstructed Pay/PaySui built from unrelated RPC coins instead of the real PTB.

- Added `SuiPtbParser`, a pure-Kotlin BCS decoder for Sui's `TransactionData::V1` format (no WalletCore/JNI involved), covering all seven PTB command variants (MoveCall, TransferObjects, SplitCoins, MergeCoins, Publish, MakeMoveVec, Upgrade), both `CallArg` input variants (Pure, Object with its three sub-kinds), and `Argument`/`TypeTag` recursion for generics.
- `SignSuiDisplayView` now renders the decoded sender, gas budget/price, commands, and inputs, falling back to the existing raw-bytes view only when a PTB fails to decode.
- `SecurityScannerTransactionFactory.createSUISecurityScannerTransaction` now builds a payload with `signSui` set when the transaction carries a dApp PTB, so the Blockaid scan reflects the real signed bytes and skips the RPC coin lookup entirely for that case; the native-transfer path (no PTB) is unchanged.
- Added the 9 new label strings to all 10 locales.

## Screenshots

| Before | After |
|--------|-------|
| before | after |

## Testing

- `./gradlew :data:testDebugUnitTest` — new `SuiPtbParserTest` (real cross-platform-verified PTB fixture plus hand-built fixtures per command/input variant and malformed-input cases) and `SecurityScannerTransactionFactoryTest` (asserts the dApp-PTB payload signs the PTB verbatim and skips the RPC coin lookup; asserts the native-transfer path is unaffected), plus the full existing suite.
- `./gradlew :app:testDebugUnitTest` — full suite, green.
- `./gradlew :app:compileDebugKotlin` — green.
- `./gradlew :app:lintDebug :data:lintDebug` — green.

## Summary by CodeRabbit

* **New Features**
  * Expanded Sui transaction detail views with sender, gas info, commands, arguments, and inputs rendered in a readable format.
  * Added a raw fallback display when decoding is unavailable.
  * Added a best-effort Sui PTB decoder and improved security scanning payload construction for dApp-provided and native Sui transactions.
* **Localization**
  * Added new Sui transaction detail and address-mismatch label translations across supported languages.
* **Bug Fixes**
  * Improved resilience to decode failures and invalid/incomplete transaction payloads.
* **Tests**
  * Added unit tests for the PTB decoder and security scanner transaction factory.


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SuiPtbParser.kt
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+    private fun readStructTag(reader: SuiBcsReader): String {
+        val address = reader.readAddress()
+        val module = reader.readString()
+        val name = reader.readString()
+        val typeParams = reader.readVector { readTypeTag(reader) }
+        val generics = if (typeParams.isEmpty()) "" else "<${typeParams.joinToString(", ")}>"
+        return "$address::$module::$name$generics"
+    }
```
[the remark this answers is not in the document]
rkokhatskyi: Good catch, fixed — added a depth guard (MAX_TYPE_TAG_DEPTH = 32) to readTypeTag/readStructTag that throws IllegalStateException past the limit, which SignSuiDisplayView's existing fallback already catches. Added two regression tests covering the boundary (31 levels succeeds) and beyond it (40 levels throws).

rkokhatskyi:
@coderabbitai review

app/src/main/java/com/vultisig/wallet/ui/components/SignSuiDisplayView.kt
```diff
@@ -46,6 +58,26 @@ fun SignSuiDisplayView(
 ) {
     var isExpanded by rememberSaveable { mutableStateOf(initiallyExpanded) }
 
+    // SuiPtbParser is pure Kotlin BCS decoding, not JNI, but still non-trivial work; deferring it
+    // until first expansion mirrors SignSolanaDisplayView so non-expanding users never pay the
+    // parse cost on the composition thread.
+    val parsedTransaction =
```
johnnyluo: **Eager parsing, not deferred to expansion**

The comment above this block says parsing is deferred until first expansion, "mirrors SignSolanaDisplayView", but `SignSolanaDisplayView` places its `remember(rawTransactions) { SolanaTransactionParser.parse(tx) }` *inside* `AnimatedVisibility(visible = isExpanded)`, while here `parsedTransaction` is computed above `AnimatedVisibility(visible = isExpanded)` (line 106) — outside the gate.

As written, every composition of this verify screen parses the full PTB (BCS traversal through all commands/inputs, up to the 32-level `TypeTag` recursion) on the main/composition thread, whether or not the card is ever expanded. Since the PTB is dApp-supplied and untrusted, this pays the full decode cost unconditionally instead of only when the user opts to review details.

Suggest moving the `remember` block inside the `AnimatedVisibility` content to match the stated intent (and the Solana precedent).

rkokhatskyi: Good catch, you're right — moved the remember block inside AnimatedVisibility's content so parsing only happens once the card is actually expanded, matching SignSolanaDisplayView's real placement (the comment was aspirational, not accurate to what I'd written).

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SuiPtbParser.kt
```diff
@@ -0,0 +1,334 @@
...
+            1 ->
+                if (bytes[0].toInt() == 0 || bytes[0].toInt() == 1) {
+                    SuiPureValue("bool", (bytes[0].toInt() == 1).toString())
+                } else {
+                    SuiPureValue("u8", (bytes[0].toInt() and 0xFF).toString())
+                }
+            8 -> SuiPureValue("u64", bytes.toLittleEndianBigInteger().toString())
+            16 -> SuiPureValue("u128", "0x" + bytes.toLittleEndianBigInteger().toString(16))
+            32 -> SuiPureValue("address", "0x" + bytes.toHexString())
+            else -> SuiPureValue("bytes(${bytes.size})", "0x" + bytes.toHexString())
```
aminsato, line 198: Pure inputs are untyped on the wire, so this infers a type from byte length and the UI then prints it as fact (`Pure(address): 0x…`). A 32-byte pure is equally a `u256` or a `vector<u8>` with 31 payload bytes; an 8-byte one is equally a 7-byte `vector<u8>`. On a screen whose only job is telling the co-signer what they are about to sign, a confident wrong label is worse than no label — qualify it (`32 bytes (address?)`) or always append the raw hex next to the guess.

Line 197 also renders u128 as hex while u64 is decimal, so a u128 amount shows up as `0x2386f26fc10000`. Make it decimal to match.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SuiPtbParser.kt
```diff
@@ -0,0 +1,334 @@
...
+        }
+}
+
+data class ParsedSuiTransaction(
+    val sender: String,
+    val gasPrice: BigInteger,
+    val gasBudget: BigInteger,
+    val gasPaymentCount: Int,
```
aminsato: `gasPaymentCount` is decoded and exposed but never rendered — `DecodedPtbSections` only reads sender, gasBudget and gasPrice. Either surface it in `GasSummarySection` or drop the field; the payment vector still has to be read to advance the cursor either way.

app/src/main/java/com/vultisig/wallet/ui/components/SignSuiDisplayView.kt
```diff
@@ -72,52 +84,326 @@ fun SignSuiDisplayView(
...
         }
     }
 }
 
+@Composable
+private fun DecodedPtbSections(transaction: ParsedSuiTransaction) {
+    if (transaction.sender.isNotBlank()) {
+        VerifyCardJsonDetails(
+            title = stringResource(R.string.sui_sender),
+            subtitle = transaction.sender,
```
aminsato, line 132: The decoded path renders the PTB's own sender and drops the `sender` param entirely (only `RawBytesFallback` still uses it). If a dApp supplies a PTB whose sender is not this vault, the card shows the dApp's claim with nothing to compare it against — exactly the case a co-signer needs to catch. Compare `transaction.sender` against `sender` and mark a mismatch.

aminsato (commented):
A few points on the decode path.

johnnyluo (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SuiPtbParser.kt
rkokhatskyi: On the u128/u64 mismatch: fixed, u128 is decimal now too. On the confident-label point: verified this isn't a new gap — iOS's decodePure does the identical unhedged byte-length classification, and Windows falls back to the exact same unhedged guess whenever its ABI/RPC lookup misses or doesn't apply to that input. Since none of the three platforms qualify the label today, I'd rather not add a divergent hedge here without a matching product decision across platforms; happy to revisit if we want to push that change everywhere.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SuiPtbParser.kt
rkokhatskyi: Agreed, fixed — dropped gasPaymentCount and surfaced gasOwner instead, which now backs the sender/gas-owner mismatch check below. The payment vector is still walked to advance the cursor, just not kept anymore.

app/src/main/java/com/vultisig/wallet/ui/components/SignSuiDisplayView.kt
rkokhatskyi: Good catch, fixed — the decoded sender now shows a warning color plus an explicit mismatch line against the vault's own address when they differ. Went a step further too: gasData.owner was being read to advance the cursor but silently discarded, which is the same risk in a worse shape — a sponsored-transaction PTB could name a different gas payer than this vault with zero visibility, so that's now surfaced and checked against the vault address the same way.

PR #5379, opened 2026-07-22
fix(evm): average two central samples for even-length priority-fee sets

### LINKED ISSUE #5376: [Fix] EVM priority-fee median picks upper-middle sample instead of averaging (Polygon and other EVM chains)
## Problem
`List<BigInteger>.median()` in `data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt:394` computes the median as `getOrNull(size / 2)` — the upper-middle element of a sorted list. That's correct for an odd-length list but wrong for an even-length one, where the true median is the average of the two central elements.

`calculateMaxPriorityFeePerGas` (`EthereumFeeService.kt:253-313`) calls this `.median()` extension for the Polygon branch and the catch-all `else` branch (every EVM chain other than Avalanche, Ethereum-swap, Arbitrum/Mantle, Blast, and Base/Optimism). The input list comes from `EvmApi.getFeeHistory()` (`EvmApi.kt:594-612`), which requests `eth_feeHistory` with a fixed block count of **10** — an even number that never changes at runtime. So this branch hits the buggy even-length path on effectively every quote for Polygon and every other chain routed through the `else` branch, not just as an intermittent edge case.

This is the same bug class as #5344 (Solana's `getMedianPriorityFee`), found as a sibling while fixing that ticket.

## Expected Behavior
The median should average the two central elements when the sorted sample list has an even length, matching the fix applied to the Solana median calculation in PR #5375.

### Must Do
- Fix `List<BigInteger>.median()` (or its call sites) in `EthereumFeeService.kt` to average the two central elements for an even-length list
- Add a unit test covering both odd- and even-length sample sets, including the real 10-sample `getFeeHistory` shape

### Must NOT Do
- Don't change how `getFeeHistory` fetches or filters samples
- Don't change the priority-fee floor/clamp logic around the median calculation

## Acceptance Criteria

rkokhatskyi:
Fixes #5376

`median()` indexed straight into `getOrNull(size / 2)`, which is the upper-middle element of the sorted `eth_feeHistory` reward list. That's correct for an odd sample count but wrong for an even one. `getFeeHistory()` always requests a fixed 10-block window, so this hit the buggy even-length path on effectively every quote for Polygon and every other chain routed through the catch-all `else` branch, not just as an edge case.

Fixed to average the two central elements when the sample count is even — same fix as the Solana sibling in #5375.

An existing test (`two-element fee history picks the upper midpoint`) was pinning the buggy behavior (asserting `gwei(8)` for `[2, 8]`); renamed and corrected to assert the true average (`gwei(5)`).

`solanaMedianPriorityFee` (Solana, #5375) and this EVM fix independently implemented the identical even-length-averaging math. Consolidated both into `List<BigInteger>.median()` in `BigIntegerExtensions.kt` so the two chains delegate to one implementation instead of maintaining two copies that could drift apart.

## Test plan
- Fixed the pre-existing regression test plus two new ones: a 10-element case matching the real `getFeeHistory` window shape, and an even-sum-is-odd case confirming truncation rather than rounding.
- New `BigIntegerExtensionsTest` (8 cases) is now the canonical test for the shared median algorithm; `SolanaMedianPriorityFeeTest` trimmed to the two cases specific to that wrapper (fallback-on-empty, delegation).
- `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeServiceTest" --tests "com.vultisig.wallet.data.api.SolanaMedianPriorityFeeTest" --tests "com.vultisig.wallet.data.utils.BigIntegerExtensionsTest" --tests "com.vultisig.wallet.data.api.SolanaApiBodyReadTest" --tests "com.vultisig.wallet.data.api.SolanaBroadcastActionTest"` — all green (77 tests).
- `./gradlew :data:lintDebug` — clean (checked explicitly this time, since the Solana sibling PR broke CI lint on `BigInteger.TWO` requiring API 33; this fix uses `BigInteger.valueOf(2)`, API 1).
- `./gradlew :data:ktfmtCheck` — clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fee median calculations for both Ethereum and Solana when the sample count is even by averaging the two central values.
  * Preserved correct behavior for odd-sized samples and ensured empty inputs still use the configured fallback.
  * Ensured integer truncation behavior matches expected results for fractional averages.
* **Tests**
  * Expanded and updated median edge-case coverage to prevent regressions in fee-history calculations.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/utils/BigIntegerExtensions.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/utils/BigIntegerExtensions.kt
@@ -36,3 +36,17 @@ fun BigInteger.increaseByPercent(percent: Int): BigInteger =
 
 fun BigInteger.toValue(decimals: Int): BigDecimal =
     this.toBigDecimal().divide(BigDecimal.TEN.pow(decimals))
+
+/**
+ * Median of this list (already sorted ascending), or null when empty. Averages the two central
+ * elements for an even-length list instead of picking a single upper-middle one.
+ */
+fun List<BigInteger>.median(): BigInteger? {
+    if (isEmpty()) return null
+    val mid = size / 2
+    return if (size % 2 == 0) {
+        (this[mid - 1] + this[mid]) / BigInteger.valueOf(2)
+    } else {
+        this[mid]
+    }
+}
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/SolanaMedianPriorityFeeTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/SolanaMedianPriorityFeeTest.kt
@@ -1,6 +1,6 @@
 package com.vultisig.wallet.data.api
 
-import java.math.BigInteger
+import com.vultisig.wallet.data.utils.median
 import kotlin.test.assertEquals
 import org.junit.jupiter.api.Test
 
@@ -14,43 +14,8 @@ internal class SolanaMedianPriorityFeeTest {
     }
 
     @Test
-    fun `single sample returns that sample`() {
-        assertEquals(
-            500.toBigInteger(),
-            solanaMedianPriorityFee(listOf(500.toBigInteger()), fallback),
-        )
-    }
-
-    @Test
-    fun `odd sample count returns the middle element unchanged`() {
-        val samples = listOf(100, 200, 300).map { it.toBigInteger() }
-        assertEquals(200.toBigInteger(), solanaMedianPriorityFee(samples, fallback))
-    }
-
-    @Test
-    fun `even sample count averages the two central elements`() {
-        // Ticket regression case: upper-middle (800) is wrong, true median is (200+800)/2 = 500.
+    fun `non-empty sample set delegates to the shared median calculation`() {
         val samples = listOf(100, 200, 800, 900).map { it.toBigInteger() }
-        assertEquals(500.toBigInteger(), solanaMedianPriorityFee(samples, fallback))
-    }
-
-    @Test
-    fun `two samples averages both`() {
-        val samples = listOf(100, 300).map { it.toBigInteger() }
-        assertEquals(200.toBigInteger(), solanaMedianPriorityFee(samples, fallback))
-    }
-
-    @Test
-    fun `even sample count with an odd sum truncates down instead of rounding`() {
-        val samples = listOf(100, 100, 201, 300).map { it.toBigInteger() }
-        // (100 + 201) / 2 = 150 (BigInteger division truncates toward zero for non-negative
-        // values).
-        assertEquals(150.toBigInteger(), solanaMedianPriorityFee(samples, fallback))
-    }
-
-    @Test
-    fun `large values do not overflow`() {
-        val samples = listOf(BigInteger("9999999999999999999"), BigInteger("10000000000000000001"))
-        assertEquals(BigInteger("10000000000000000000"), solanaMedianPriorityFee(samples, fallback))
+        assertEquals(samples.median(), solanaMedianPriorityFee(samples, fallback))
     }
 }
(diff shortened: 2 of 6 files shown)


johnnyluo (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/utils/BigIntegerExtensions.kt
```diff
@@ -36,3 +36,17 @@ fun BigInteger.increaseByPercent(percent: Int): BigInteger =
...
+ */
+fun List<BigInteger>.median(): BigInteger? {
+    if (isEmpty()) return null
+    val mid = size / 2
+    return if (size % 2 == 0) {
+        (this[mid - 1] + this[mid]) / BigInteger.valueOf(2)
+    } else {
+        this[mid]
+    }
+}
```
aminsato, line 52: The sorted-ascending precondition now lives only in the KDoc, and the function is public across the module — any `List<BigInteger>` gets `.median()` in autocomplete. On an unsorted list `this[mid - 1] + this[mid]` averages two arbitrary samples and returns a plausible-looking wrong number, which is the same silent failure mode this PR exists to fix. The two current callers happen to sort first; the third one won't necessarily.

Sorting inside is O(n log n) on a 10-element window and removes the trap:

```kotlin
fun List<BigInteger>.median(): BigInteger? {
    if (isEmpty()) return null
    val sorted = sorted()
    val mid = size / 2
    return if (size % 2 == 0) {
        (sorted[mid - 1] + sorted[mid]) / BigInteger.valueOf(2)
    } else {
        sorted[mid]
    }
}
```

Both call sites keep their own `.sorted()` regardless — `getFeeHistory()` needs it so `EthereumFeeService` can use `lastOrNull()` as the window max for Ethereum swaps — so this is purely defensive for future callers. If the caller-sorts contract is worth keeping, `medianOfSorted()` at least puts it in the name.

data/src/test/kotlin/com/vultisig/wallet/data/api/SolanaMedianPriorityFeeTest.kt
```diff
@@ -14,43 +14,8 @@ internal class SolanaMedianPriorityFeeTest {
...
-        // (100 + 201) / 2 = 150 (BigInteger division truncates toward zero for non-negative
-        // values).
-        assertEquals(150.toBigInteger(), solanaMedianPriorityFee(samples, fallback))
-    }
-
-    @Test
-    fun `large values do not overflow`() {
-        val samples = listOf(BigInteger("9999999999999999999"), BigInteger("10000000000000000001"))
-        assertEquals(BigInteger("10000000000000000000"), solanaMedianPriorityFee(samples, fallback))
+        assertEquals(samples.median(), solanaMedianPriorityFee(samples, fallback))
```
aminsato, line 19: `samples.median()` on the expected side makes this assertion self-referential — it holds for any implementation of `median()`, including the upper-middle behavior this PR removes. Asserting the literal keeps the Solana wrapper pinned to a real value on its own:

```kotlin
assertEquals(500.toBigInteger(), solanaMedianPriorityFee(samples, fallback))
```

That drops the last use of the `median` import at the top of the file.

aminsato (commented):
Two points on the shared helper and the trimmed Solana test.

data/src/main/kotlin/com/vultisig/wallet/data/utils/BigIntegerExtensions.kt
rkokhatskyi: Fair point, and confirmed against the merged source — both current callers (EvmApi.getFeeHistory and SolanaApi.getMedianPriorityFee) sort before calling, so this isn't live today, no third caller exists yet either. The public-visibility footgun for a future caller is real though. Leaving it as-is for now since nothing currently hits it; will sort defensively or rename to medianOfSorted if a new caller shows up. No change needed right now.

data/src/test/kotlin/com/vultisig/wallet/data/api/SolanaMedianPriorityFeeTest.kt
rkokhatskyi: You're right, that assertion is self-referential and doesn't actually protect against a regression since both sides call the same function on the same input. The real math is still covered by BigIntegerExtensionsTest's literal-value cases though, so no correctness risk. Leaving this test as-is for now, can tighten it to assert the literal value later.

PR #5384, opened 2026-07-23
fix(swap): account for price impact so Total Fee is reconcilable (#5335)

### LINKED ISSUE #5335: Swap: "Total Fee" understates real swap cost (excludes slippage), unreconcilable with explorer or value delta

aminsato:
Closes #5335

## Problem

On the swap **Transaction complete** screen, Total Fee reconciles with neither the on-chain network fee nor the value the user actually lost. The reported LTC → ADA swap showed **$0.00051** against a **~$0.07** from/to delta — 137× smaller than what the swap really cost.

Investigating turned up **two** independent causes:

**1. Price Impact never survived the swap form.** The liquidity/slippage cost is baked into the quoted destination amount, so it is deliberately excluded from Total Fee (per #5334's must-not). But it was only ever *shown* on the swap form, because it lived on the quote pipeline and never reached `SwapTransaction` — so after signing, the dominant cost on a thin route had no row anywhere.

**2. The completion screen's rows don't add up.** #5373 (the #5334 fix) and #5364 (the #5358 fix) merged the same day. #5373 added the fee breakdown to `SwapTransactionOverviewScreen` rendering `providerFee.fiatValue` unconditionally, while #5364 taught the verify screen to honor `swapFeeHidden` / `swapFeeIncludedInRate`. The result: on a SwapKit UTXO deposit the completion screen lists a Swap Fee that contributes nothing to Total Fee — and that is precisely the LTC → ADA route in this issue.

## Approach

Of the two options the issue proposes, this takes the second — **keep the fee math, give the missing cost its own row**. Folding slippage into Total Fee is explicitly ruled out by #5334 ("do not change the `feesFiatForTotal + gasFeeFiatValue` computation or the intentional exclusion of the liquidity/slippage component"). **No fee value changes in this PR.**

## Changes

| File | What |
|---|---|
| `data/…/models/SwapTransaction.kt` | new in-memory `priceImpact: BigDecimal?`, carried like the #5358 fee/discount context |
| `SwapTransactionBuilder.kt` | `priceImpact = quote.priceImpact` in all four provider branches |
| `SwapTransactionToUiModelMapper.kt` | formats it via the existing `formatPriceImpact()` |
| `VerifySwapViewModel.kt` | `priceImpactPercent` / `priceImpactLevel` on `SwapTransactionUiModel` |
| `PriceImpactRow.kt` *(new)* | shared row; the swap form's inline block collapses into it |
| `VerifySwapScreen.kt` | Price Impact row above Total Fee |
| `SwapTransactionOverviewScreen.kt` | Price Impact row **+** the Swap Fee row now honors hidden / included-in-rate |

`SwapQuote.priceImpact` was already populated — THORChain/Maya derive it from `slippage_bps`, SwapKit from the route's `meta.priceImpact` — so there is no new network work. **No new strings**: the four the form uses (`swap_form_price_impact_title`, `swap_price_impact_good/average/high`) already exist in all 10 locales.

The co-signer path (`JoinSwapUiModelBuilder`) has no quote to read the impact from, so the row hides there — the same graceful degradation the VULT-tier and referral rows use.

## Testing

- 336 mapper + swap unit tests green offline, including 2 new ones in `SwapTransactionToUiModelMapperFeeBreakdownTest`:
  - price impact reaches the UI model (`-1.58%` / Average) **while `totalFee` stays `1.20`** — pinning that the fee math is untouched
  - a provider reporting no impact leaves both fields null, hiding the row
- `./gradlew :app:lintDebug` clean
- ktfmt applied

**In-app check** — swap **LTC → ADA** (the reported route), sign, then expand details on Transaction complete: there should be **no Swap Fee row**, **Total Fee should equal the Network Fee** exactly, and **Price Impact** should show what the route cost. A THORChain pair (e.g. BTC → ETH) shows the Price Impact row on both Verify and Transaction complete.

## Known limitation

1inch / Kyber / LiFi / Jupiter report no price impact, so the row stays hidden on those routes and Total Fee remains the only number there. That is an upstream data gap and matches iOS behavior.

## Checklist

- [x] Lint clean
- [x] Unit tests green
- [x] Self-reviewed against requirements
- [x] No secrets committed
- [x] Conventional commit messages

## Summary by CodeRabbit

- **New Features**
  - Added price impact details to swap verification, fee breakdowns, and transaction overview screens.
  - Displays the price impact percentage with a clear status indicator: good, average, or high.
  - Shows swap fee percentages and whether fees are included in the exchange rate.
  - Hides fee and price impact rows when relevant information is unavailable.

- **Bug Fixes**
  - Preserved total fee calculations while displaying price impact separately.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/mappers/SwapTransactionToUiModelMapper.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/mappers/SwapTransactionToUiModelMapper.kt
@@ -12,6 +12,7 @@ import com.vultisig.wallet.data.usecases.ConvertTokenValueToFiatUseCase
 import com.vultisig.wallet.ui.models.swap.SwapTransactionUiModel
 import com.vultisig.wallet.ui.models.swap.ValuedToken
 import com.vultisig.wallet.ui.models.swap.clampDstFiatToSrcFiat
+import com.vultisig.wallet.ui.models.swap.formatPriceImpact
 import com.vultisig.wallet.ui.models.swap.formatSwapKitProviderLabel
 import javax.inject.Inject
 import kotlinx.coroutines.flow.first
@@ -93,6 +94,12 @@ constructor(
             } else {
                 quotesFeesFiat
             }
+
+        // The liquidity component stays out of the fee total (it is already reflected in the
+        // destination amount), which on a thin route leaves most of what the user gave up
+        // unexplained. Surface it as its own Price Impact row, formatted as the swap form does
+        // (#5335).
+        val priceImpactDisplay = formatPriceImpact(from.priceImpact)
 
         // SwapTransaction carries no destination fiat, so it is recomputed here for the verify and
         // keysign screens. Apply the same value-preserving clamp the swap form uses (#4878) so an
@@ -177,6 +184,8 @@ constructor(
             vultBpsDiscountFiatValue = from.vultBpsDiscountFiatValue,
             referralBpsDiscount = from.referralBpsDiscount,
             referralBpsDiscountFiatValue = from.referralBpsDiscountFiatValue,
+            priceImpactPercent = priceImpactDisplay?.percent,
+            priceImpactLevel = priceImpactDisplay?.level,
         )
     }
 }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapTransactionBuilder.kt
@@ -123,6 +123,7 @@ constructor(
                     vultBpsDiscountFiatValue = feeDisplay.vultBpsDiscountFiatValue,
                     referralBpsDiscount = feeDisplay.referralBpsDiscount,
                     referralBpsDiscountFiatValue = feeDisplay.referralBpsDiscountFiatValue,
+                    priceImpact = quote.priceImpact,
                     payload =
                         SwapPayload.ThorChain(
                             THORChainSwapPayload(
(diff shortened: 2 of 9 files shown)


johnnyluo (approved):
LGTM

PR #4734, opened 2026-06-04
test(deposit): DepositFormViewModel + DeFiNavActions coverage (#4186)

rkokhatskyi:
## Summary

Adds the test coverage requested in #4186. Refs #4186 (also changed the stale things for scope)

- `ThorchainFunctionsTest` already covers **all 7** memo builders (`stakeRUJI`, `unstakeRUJI`, `claimRujiRewards`, `mintYToken`, `redeemYToken`, `stakeTcyCompound`, `unStakeTcyCompound`) — 22 tests, green.
- `DepositFormViewModelTest` already existed (18 tests), and per-action transaction construction is covered by `BondStrategyTest` / `UnbondStrategyTest` / `LeaveStrategyTest`.
- The RUJI/TCY/yToken/TRX/Circle flows moved out of the deposit VM into the **Send** module; in the deposit VM, `loadData` now maps only `BOND`/`UNBOND`/`STAKE_CACAO`/`UNSTAKE_CACAO`/`ADD_LP`/`REMOVE_LP` to a `DepositOption`.

This PR fills the **genuine remaining gaps** against the ticket's intent, with **no production changes**.

### `DepositFormViewModelTest` (+23 tests → 41 total)
- **Per-chain deposit-option scoping**: ThorChain (Bond/Unbond/Leave/Custom/Merge/UnMerge/WithdrawSecuredAsset), GaiaChain (TransferIbc/Switch), Ton (Stake/Unstake), alongside the existing MayaChain (Leave/Custom) case.
- **`selectDepositOption` default-token mapping** for Unbond/Leave/AddLiquidity on ThorChain (RUNE).
- **Previously-untested validators**: provider/operator address, node address, operator-fee basis points (incl. the intentional blank no-op), slippage (required/invalid/format/valid), asset chars, LP-unit chars, and the Switch-scoped thor-address no-op.

### `DeFiNavActionsTest` (new, 10 tests)
- `parseDepositType` round-trips **every** `DeFiNavActions` value, plus aliases, case-folding, `unknown → null`, and the USDC-circle `valueOf` fallback path.
- `getStakeDeFiNavAction` / `getUnstakeDeFiNavAction` coin → action mappings, incl. unsupported coin → throws.

## Notes / follow-ups (production deliberately untouched per the ticket)

[...]

PR #5385, opened 2026-07-23
fix(sui): bound coin selection for cross-device keysign parity

### LINKED ISSUE #5252: fix(sui): match iOS #4734 coin-object selection — cross-device keysign parity
## Summary

iOS PR [vultisig/vultisig-ios#4734](https://github.com/vultisig/vultisig-ios/pull/4734) changes how a Sui send selects coin objects: it now embeds/references only the **fewest largest objects that cover the send** (bounded by 255, one under Sui's 256-gas-payment cap), instead of every owned object. This fixes real failures on wallets whose SUI balance is scattered across many coin objects:

- **Keysign can't start:** oversized payload → co-signer poll `HTTP 404`, initiator `Transaction data expired`.
- **Broadcast rejected:** `Size limit exceeded: serialized transaction size exceeded maximum of 131072 is 303638` (Sui's 128 KiB tx limit).
- `InsufficientCoinBalance in command 0` on token sends.

**Android must ship the byte-identical selection algorithm — not just a similar fix.** This is a cross-device consensus requirement, not only a bug port (see below).

## Why this is a parity blocker, not a nice-to-have

In MPC keysign every co-signing device **independently recomputes** the Sui transaction (and its sighash) from the shared `KeysignPayload`, and all devices must produce **byte-identical** transaction bytes or the signatures don't combine and the send fails.

Android currently references **every** coin object from the payload, in RPC order:
- `data/src/main/kotlin/com/vultisig/wallet/data/crypto/SuiHelper.kt:50-95` — builds `suiObjectRefs` / `nonSuiObjectRef` from all payload coins; no covering-subset selection.
- `data/src/main/kotlin/com/vultisig/wallet/data/blockchain/sui/SuiFeeService.kt:106-108` — `coins = allCoinsDeferred.await()` embeds every owned coin, unbounded, in RPC order.

iOS (post-#4734) sorts by balance desc and selects a covering subset. So once #4734 ships:

- **Android/older device initiates** a Sui send from a multi-object wallet → payload carries all objects in RPC order → an updated **iOS co-signer trims to a balance-sorted subset** → different tx bytes → **keysign fails**. This is the common case (most wallets hold several SUI objects; a typical send needs one).
- Even a near-max send that needs *all* objects still diverges, because iOS emits them balance-sorted while Android emits RPC order.

Before #4734, iOS and Android agreed (both referenced all objects in payload order). #4734 introduces the divergence, so Android has to match to keep cross-platform vaults working.

## Scope for Android

Port the **exact** deterministic algorithm from #4734 (`SuiConstants.swift` → `SuiCoinType`) so Android builds byte-identical Sui transactions:

1. **Input-coin selection** (`selectInputCoins`): fewest largest objects covering `target`, sorted **balance desc, then `objectID` ascending** as the tie-break; cap at **255**. Native `PaySui` target = `amount + gasBudget`; token `Pay` target = `amount`.
2. **Payload bounding** (`selectPayloadCoins` in the payload builder, `SuiFeeService`): embed only the covering objects (native) / covering token objects + the largest-few native SUI gas candidates (token). This also fixes the relay-404 / oversized-payload symptom on Android.
3. **Gas object:** Android already picks the smallest-covering SUI object (`SuiHelper.kt:151-153 selectSuiGasCoin`, #3989) — keep it, just confirm the tie-break/ordering matches iOS exactly.

The sort key, tie-break, cap value, and gas-pick must be **identical** to iOS down to the byte, or cross-platform keysigns still diverge.

## Verification (required)

Add a **cross-platform interop test**: a fixed owned-coins fixture + `amount` + `gasBudget` → assert the built `Sui.SigningInput` bytes are **identical** to iOS's for the same fixture (native small send, native near-max, 800-object dusty wallet, token send). #4734 ships `SuiHelperInputDataTests.swift` with these exact fixtures — mirror them.

## Coordination

Do **not** merge #4734 to a public iOS release until this lands (or all platforms gate the new behavior behind a version/capability flag). Track: iOS #4734 · this issue (Android) · vultisig-sdk parity issue (link once filed).

---
_Receipts verified against `vultisig-android@main` and `vultisig-ios` PR #4734 head (`6834d99`)._

aminsato:
## Summary

A Sui send referenced **every** coin object the wallet owned, in RPC order. This PR selects only the fewest largest objects that cover the send, using an ordering that is byte-identical to the other platforms.

Two problems, one fix:

- **Size.** On a wallet whose balance is scattered across many objects, referencing all of them overruns Sui's 128 KiB transaction-size limit and — for a native `PaySui`, whose entire input set is the gas payment — the 256-object gas cap. The payload becomes too large to relay (co-signer polls `404`, initiator reports `Transaction data expired`), or the broadcast is rejected with `Size limit exceeded: serialized transaction size exceeded maximum of 131072`.
- **Consensus.** In MPC keysign every co-signing device independently rebuilds the transaction from the shared `KeysignPayload`. If two devices pick different objects, or the same objects in a different order, the transaction bytes differ and the signatures don't combine.

Closes #5252

## What changed

- `SuiHelper.selectInputCoins` — fewest largest objects covering the target, sorted **balance descending, `coinObjectId` ascending** as tie-break, capped at **255** (one under Sui's 256-object gas cap). Native `PaySui` target is `amount + gasBudget`; token `Pay` target is `amount`.
- `SuiHelper.selectPayloadCoins` — bounds what gets embedded in the payload: covering native objects for a native send; covering token objects plus the 5 largest native SUI gas candidates for a token send. Wired into both `BlockChainSpecificRepository` (real keysign) and `SuiFeeService` (dry run), which is what fixes the oversized-payload relay failure.
- `SuiHelper.selectSuiGasCoin` — still the smallest covering native object, now with `coinObjectId` ascending as tie-break.
- **Coin-type normalization** — Sui accepts a coin type's address in short (`0x2`) or zero-padded long (`0x000…002`) form and RPC nodes are inconsistent about which they return. The previous raw string compare would classify a native object as a token object on one device and not another; because the send branches on `tokenObjects.isNotEmpty()`, that silently turns a token `Pay` into a native `PaySui`. Only the address is normalized — `::module::Struct` identifiers stay case-sensitive, since `::coin::USDC` and `::coin::usdc` are genuinely different types.

## Cross-platform parity

The reference implementation is **[vultisig-sdk#1275](https://github.com/vultisig/vultisig-sdk/pull/1275)**, merged 2026-07-16, which also covers Windows via `@core`. The same algorithm is proposed for iOS in [vultisig-ios#4734](https://github.com/vultisig/vultisig-ios/pull/4734), which is held open pending this PR — merging here unblocks it.

Sort key, tie-breaks, the 255 cap, the 5-object gas-candidate count, and the gas-object pick all match the SDK.

**One deliberate deviation:** the SDK *throws* on an under-funded selection and when no object covers the gas budget; this PR stays best-effort (returns the capped set / `null`). `SuiFeeService` calls `selectPayloadCoins` during live fee estimation as the user types an amount, so throwing there would break the fee row instead of letting the send form's own insufficient-balance validation surface the problem. Byte-parity is unaffected — the selections are identical whenever the send is funded. Happy to switch to throwing if reviewers prefer strict parity.

## Test plan

**Unit** — `SuiCoinSelectionTest`, 21 tests, green. Mirrors the SDK's `suiCoinSelection.test.ts` cases: covering selection, both tie-breaks, the 255 cap, normalization (long-form native, case-sensitive struct ids), and payload bounding for native and token sends. Full `:data:testDebugUnitTest` is 2202 tests, 0 failures.

**Mainnet, signed end-to-end on device** — vault `0x3a91…c65c`:

| Path | Transaction | What it exercises |
|---|---|---|
| Native SUI (`PaySui`) | [`FGSi1DUa…LcEoA`](https://suiscan.xyz/mainnet/tx/FGSi1DUasU6CkiDSSKwJbNpN4fHBM5Dve6oPGJGLcEoA) | `SplitCoins` from `GasCoin` → `TransferObjects`; bounded selection resolved to a **single** gas-payment object |
| Sui token (`Pay`) | [`7WD6TGJC…9EtL3`](https://suiscan.xyz/mainnet/tx/7WD6TGJCdtBjoYK64tJpeMz9kGBY9g8CER2eMq89EtL3) | ETH-on-Sui token object `0x7ec3…6ea0` split, gas paid from a **separate** native object — the branch the normalization change affects |

Both succeeded. The token send is the important one: it confirms the coin-type comparison still routes a token to `Pay` rather than falling through to `PaySui`.

**Not covered by these transactions, stated plainly:**

- The signing vault holds only 3 native SUI objects, so these txs do **not** exercise the scattered-balance case the size fix targets. That path is covered by unit tests (800-object dusty-wallet fixture) but not by an on-chain send.
- Cross-device byte parity is **not** demonstrated. It needs an iOS build carrying #4734, which is unmerged and unreleased, so no counterpart device exists to disagree with Android yet. Parity here rests on unit tests plus a line-by-line audit against the merged SDK.
- `SuiHelperInputDataTest` (8 byte-level `Sui.SigningInput` assertions) compiles but has not been run — it needs the WalletCore JNI on a device/emulator.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Sui payment and fee-simulation accuracy for fragmented balances by embedding only the required coins for transfers.
  - Deterministic native/token coin selection with native coin-type normalization, capped reference counts, and exclusion of look-alike coin types.
  - Added stricter handling for missing or underfunded token selections (fail fast instead of fallback).
  - Updated Sui coin fetching to cursor-based pagination to reliably retrieve all coins.

- **Tests**
  - Added extensive coverage for native/token coin selection, gas and payload selection rules, type matching, coin limits, and failure/rejection scenarios.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/SuiHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/SuiHelper.kt
@@ -25,7 +25,49 @@ object SuiHelper {
     private val coinType = CoinType.SUI
     private val suiContractAddress = "0x2::sui::SUI"
 
-    private fun getPreSignedInputData(keysignPayload: KeysignPayload): ByteArray {
+    /**
+     * Sui accepts a coin type's address part in short (`0x2`) or zero-padded long (`0x000…002`)
+     * form, and RPC nodes are not consistent about which they return. Comparing the raw strings
+     * would classify the same coin differently on two devices — a native object read as a token
+     * object silently turns a `Pay` into a `PaySui`. Only the address is normalized (leading zeros
+     * stripped, lowercased); Move module and struct identifiers stay case-sensitive because
+     * `::coin::USDC` and `::coin::usdc` are genuinely different types. Mirrors iOS
+     * `SuiCoinType.normalize` and SDK `normalizeSuiCoinType` (vultisig-sdk#1275).
+     */
+    internal fun normalizeSuiCoinType(coinType: String): String {
+        val addressEnd = coinType.indexOf("::")
+        if (addressEnd < 0) return normalizeSuiAddress(coinType)
+        return normalizeSuiAddress(coinType.substring(0, addressEnd)) +
+            coinType.substring(addressEnd)
+    }
+
+    internal fun isSameSuiCoinType(lhs: String, rhs: String): Boolean =
+        normalizeSuiCoinType(lhs) == normalizeSuiCoinType(rhs)
+
+    private fun normalizeSuiAddress(address: String): String {
+        val hex = address.lowercase().removePrefix("0x").trimStart('0')
+        return "0x" + hex.ifEmpty { "0" }
+    }
+
+    private fun SuiCoin.isNativeSui(): Boolean = isSameSuiCoinType(coinType, suiContractAddress)
+
+    /**
+     * Upper bound on the coin objects a single send may reference. Sui rejects a transaction whose
+     * serialized size exceeds 128 KiB, and a `PaySui` send uses its entire input set as the gas
+     * payment — which Sui caps at 256 objects. Staying one under that cap keeps every send safely
+     * within both limits. Mirrors iOS `SuiConstants.maxInputCoinObjects`.
+     */
+    internal const val MAX_INPUT_COIN_OBJECTS = 255
+
+    /**
+     * How many of the largest native SUI objects to embed as gas candidates for a token send. The
+     * signer picks one to pay gas; carrying the largest few (rather than all, or just one) keeps
+     * the payload small while guaranteeing a covering object survives a re-estimated gas budget.
+     * Mirrors iOS `SuiConstants.gasCandidateObjectCount`.
+     */
+    internal const val GAS_CANDIDATE_OBJECT_COUNT = 5
+
+    internal fun getPreSignedInputData(keysignPayload: KeysignPayload): ByteArray {
         require(keysignPayload.coin.chain == Chain.Sui) { "Coin is not SUI" }
 
         // dApp-supplied Programmable Transaction Block (Sui Wallet Standard): the base64
@@ -47,47 +89,46 @@ object SuiHelper {
 
         val toAddress = AnyAddress(keysignPayload.toAddress, coinType)
 
-        val (nonSuiObjectRef, suiObjectRefs) =
-            coins
-                .partition {
-                    it.coinType.equals(keysignPayload.coin.contractAddress) &&
-                        !it.coinType.equals(suiContractAddress)
-                }
-                .let { (notSui, sui) ->
-                    notSui.map {
-                        Sui.ObjectRef.newBuilder()
-                            .setObjectId(it.coinObjectId)
-                            .setVersion(it.version.toLong())
-                            .setObjectDigest(it.digest)
-                            .build()
-                    } to
-                        sui.filter { it.coinType.equals(suiContractAddress) }
-                            .map {
-                                Sui.ObjectRef.newBuilder()
-                                    .setObjectId(it.coinObjectId)
-                                    .setVersion(it.version.toLong())
-                                    .setObjectDigest(it.digest)
-                                    .build()
-                            }
-                }
-        val gascoin = selectSuiGasCoin(coins, gasBudget)
+        val suiObjects = coins.filter { it.isNativeSui() }
+        val tokenObjects =
+            coins.filter {
+                isSameSuiCoinType(it.coinType, keysignPayload.coin.contractAddress) &&
+                    !it.isNativeSui()
+            }
 
         val input =
-            if (nonSuiObjectRef.isNotEmpty()) {
+            if (tokenObjects.isNotEmpty()) {
+                    // Token send (Pay): reference only the largest token objects covering the
+                    // amount, and pay gas from a single native SUI object that covers the budget.
+                    val gasCoin = selectSuiGasCoin(coins, gasBudget)
+                    val gasObjectRef =
+                        selectSuiGasObjectRef(
+                            suiObjects.map { it.toObjectRef() },
+                            gasCoin?.coinObjectId,
+                        )
+                    val tokenInputCoins =
+                        selectInputCoins(tokenObjects, keysignPayload.toAmount).map {
+                            it.toObjectRef()
+                        }
                     Sui.SigningInput.newBuilder()
                         .setPay(
                             Sui.Pay.newBuilder()
-                                .setGas(selectSuiGasObjectRef(suiObjectRefs, gascoin?.coinObjectId))
-                                .addAllInputCoins(nonSuiObjectRef)
+                                .setGas(gasObjectRef)
+                                .addAllInputCoins(tokenInputCoins)
                                 .addAllRecipients(listOf(toAddress.description()))
                                 .addAllAmounts(listOf(keysignPayload.toAmount.toLong()))
                                 .build()
                         )
                 } else {
+                    // Native send (PaySui): the input set also pays gas (Sui gas-smashes it into
+                    // one coin), so cover amount + gas with the fewest largest objects.
+                    val target = keysignPayload.toAmount + gasBudget
+                    val nativeInputCoins =
+                        selectInputCoins(suiObjects, target).map { it.toObjectRef() }
                     Sui.SigningInput.newBuilder()
                         .setPaySui(
                             Sui.PaySui.newBuilder()
-                                .addAllInputCoins(suiObjectRefs)
+                                .addAllInputCoins(nativeInputCoins)
                                 .addAllRecipients(listOf(toAddress.description()))
                                 .addAllAmounts(listOf(keysignPayload.toAmount.toLong()))
                                 .build()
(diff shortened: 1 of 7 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/crypto/SuiHelper.kt
```diff
@@ -47,47 +89,46 @@ object SuiHelper {
...
+        val suiObjects = coins.filter { it.isNativeSui() }
+        val tokenObjects =
+            coins.filter {
+                isSameSuiCoinType(it.coinType, keysignPayload.coin.contractAddress) &&
+                    !it.isNativeSui()
+            }
 
         val input =
-            if (nonSuiObjectRef.isNotEmpty()) {
+            if (tokenObjects.isNotEmpty()) {
```
rkokhatskyi: This branches Pay vs PaySui on `tokenObjects.isNotEmpty()`, but `SuiApi.getAllCoins` (`SuiApi.kt:80-107`) never paginates and Sui's `suix_getAllCoins` defaults to 50 objects per page, confirmed live against the same RPC host. A wallet with 50+ native SUI objects can have its whole page be native coins, so a genuine token send finds zero token objects here and silently signs a native PaySui transfer of the raw amount instead, wrong asset, wrong value, same recipient.

data/src/main/kotlin/com/vultisig/wallet/data/crypto/SuiHelper.kt
```diff
@@ -147,10 +188,103 @@ object SuiHelper {
...
+        var accumulated = BigInteger.ZERO
+        for (coin in sorted) {
+            // Keep at least one object so a zero/near-zero-amount send still has an input;
+            // otherwise stop once the target is covered.
+            if (selected.isNotEmpty() && accumulated >= target) break
+            if (selected.size >= maxObjects) break
+            selected.add(coin)
+            accumulated += coin.balanceOrZero()
+        }
+        return selected
```
rkokhatskyi, line 255: `selectInputCoins` never checks that `accumulated` reached `target` after capping at `maxObjects`, so a fragmented balance returns silently short, unlike the sibling `selectSuiGasObjectRef` which throws on an equivalent gap. `getKeysignMessages` succeeds regardless, so the full multi-device ceremony runs to completion before Sui rejects the tx at broadcast for insufficient balance.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
```diff
@@ -603,7 +604,17 @@ constructor(
...
+                            // Embed only the coin objects the send needs, not every owned object —
+                            // an unbounded set bloats the pairing QR / TSS relay payload and fails
+                            // to relay on wallets whose balance is scattered across many objects.
+                            coins =
+                                SuiHelper.selectPayloadCoins(
+                                    coinsDeferred.await(),
+                                    isNativeToken = token.isNativeToken,
+                                    contractAddress = token.contractAddress,
+                                    amount = tokenAmountValue ?: BigInteger.ZERO,
+                                    gasBudget = suiFees.limit,
```
rkokhatskyi: Payload coins are re-selected against the refined `suiFees.limit`, but only the baseline-budget-sized payload in `SuiFeeService.kt:108-118` was ever dry-run priced. If the refined budget needs more native objects than that baseline selection, the broadcast tx carries an unpriced, larger input set, needs mainnet confirmation of the real cost delta.

data/src/main/kotlin/com/vultisig/wallet/data/crypto/SuiHelper.kt
```diff
@@ -25,7 +25,49 @@ object SuiHelper {
...
 
-    private fun getPreSignedInputData(keysignPayload: KeysignPayload): ByteArray {
+    /**
+     * Sui accepts a coin type's address part in short (`0x2`) or zero-padded long (`0x000…002`)
+     * form, and RPC nodes are not consistent about which they return. Comparing the raw strings
+     * would classify the same coin differently on two devices — a native object read as a token
+     * object silently turns a `Pay` into a `PaySui`. Only the address is normalized (leading zeros
+     * stripped, lowercased); Move module and struct identifiers stay case-sensitive because
+     * `::coin::USDC` and `::coin::usdc` are genuinely different types. Mirrors iOS
+     * `SuiCoinType.normalize` and SDK `normalizeSuiCoinType` (vultisig-sdk#1275).
```
rkokhatskyi: This says the case-sensitive match mirrors iOS `SuiCoinType.normalize`, but shipped iOS lowercases the whole coin type and is case-insensitive by design, per its own `testNormalizeIsCaseInsensitive`. It actually mirrors the SDK, not iOS; the false parity claim could mislead someone into aligning Android to iOS and collapsing distinct Move types.

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/crypto/SuiHelper.kt
aminsato: Good catch — this was the actual root cause: `SuiApi.getAllCoins` only read the first `suix_getAllCoins` page (~50 objects) while iOS paginates. Fixed in e5d7b8074 by following `nextCursor`/`hasNextPage` to completion (mirrors `SuiService.getAllCoins`), so a token whose objects land on a later page is no longer invisible. As defense-in-depth I also switched the branch to `keysignPayload.coin.isNativeToken` (the same signal `selectPayloadCoins` uses) and made a token send with no token objects fail via `check(...)` instead of falling through to `PaySui`.

data/src/main/kotlin/com/vultisig/wallet/data/crypto/SuiHelper.kt
aminsato: Fixed in e5d7b8074. `selectInputCoins` stays best-effort (payload embedding must stay lossless), but the signing path now enforces coverage via a new `checkCoinsCover` helper — the same fail-fast contract as `selectSuiGasObjectRef` — for both the native and token input sets, so an under-funded selection is rejected before the ceremony instead of at broadcast.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
aminsato: Fixed in e5d7b8074. The broadcast payload now selects coins against `SUI_DEFAULT_GAS_BUDGET` — the budget the dry-run priced against — instead of the refined `suiFees.limit`, so the embedded/broadcast object set is always a subset of what was simulated and never carries an unpriced, larger input set. The refined limit still becomes the tx's gas ceiling. One deliberate trade-off: an extreme near-full-balance send where the padded budget would need a coin the default budget didn't select now fails the coverage check (fail-fast) rather than broadcasting an unpriced set — flagging in case mainnet shows that edge matters.

data/src/main/kotlin/com/vultisig/wallet/data/crypto/SuiHelper.kt
aminsato: You're right — verified against iOS `SuiConstants.swift:42` (`coinType.lowercased()`) and `testNormalizeIsCaseInsensitive`. The doc was wrong, not the code. Corrected in e5d7b8074 to state it mirrors the SDK (case-sensitive on module/struct) and *deliberately diverges* from iOS (case-insensitive), since matching iOS would collapse distinct Move types like `::coin::USDC` vs `::coin::usdc`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
```diff
@@ -603,7 +604,24 @@ constructor(
...
+                            // the broadcast transaction never carries an unpriced, larger input set
+                            // than the dry-run measured. The refined limit still becomes the tx's
+                            // gas ceiling above.
+                            coins =
+                                SuiHelper.selectPayloadCoins(
+                                    coinsDeferred.await(),
+                                    isNativeToken = token.isNativeToken,
+                                    contractAddress = token.contractAddress,
+                                    amount = tokenAmountValue ?: BigInteger.ZERO,
+                                    gasBudget = SUI_DEFAULT_GAS_BUDGET,
```
rkokhatskyi: Embedded coins are now selected against the static `SUI_DEFAULT_GAS_BUDGET`, but `gasBudget` at `line 606` is still the uncapped, refined `suiFees.limit`. Signing computes `target` from that refined value while `suiObjects` only has the baseline-sized embedded set, so `checkCoinsCover` can reject a fully-funded, fragmented-balance native send as insufficient balance.

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
aminsato: You're right, and it's worse than an edge case — the fallback path hits it every time. `paddedDefaultSuiFees()` returns `SUI_DEFAULT_GAS_BUDGET.increaseByPercent(15)` = 3,450,000, which is *by definition* above the 3,000,000 the selection targeted, so every fee-estimation failure produced the mismatch. The dry-run path reaches it too: `(computationCost + storageCost) * 1.15` routinely lands over 3,000,000 for a real transfer.

Fixed in 78930142b by selecting against `maxOf(SUI_DEFAULT_GAS_BUDGET, suiFees.limit)`. That resolves this without regressing your earlier unpriced-input-set comment rather than trading one for the other:

- **Refined budget under the default** (common case): selection stays at the default, so the embedded set is exactly the one `SuiFeeService` dry-run priced — no unpriced inputs.
- **Refined budget above the default**: selection widens by only the objects that higher ceiling genuinely needs. Sui requires a `PaySui` input set to hold `amount + gasBudget`, so signing against a budget the selection never targeted builds a transaction the network would reject anyway — the narrower set was never actually the safer option.

Three tests pin the invariant the signing path checks (`embedded total >= amount + gasBudget`): refined-budget-above-default, the padded fallback budget, and one pinning set-identity with the dry-run priced selection when the refined budget is lower. Verified they catch it — reverting the one-line change fails the first two.

rkokhatskyi (approved):
LGTM

PR #5386, opened 2026-07-23
feat(swap): screen the external recipient on non-EVM-source swaps

### LINKED ISSUE #5348: [Feat] Recipient-only security scan fallback for non-EVM-source swaps
# [Add] Recipient-only security scan fallback for non-EVM-source swaps

## Problem
When a swap's source chain doesn't support full transaction scanning (e.g. Solana as the source), the security scanner skips screening entirely for that swap — not even the destination address gets checked. A swap sending proceeds to an external recipient from a non-EVM source chain currently gets zero malicious/scam-address screening, unlike the equivalent EVM-source swap.

## Solution
Add a fallback path that, when the source chain can't be fully scanned, still screens the destination/recipient address on its own — build a minimal recipient-only scan request and surface that verdict the same way the existing full-transaction scan verdict is shown. Start with Solana as the source chain, since it's a real, currently-supported swap source with no coverage today.

### Must Do
- [ ] Add a recipient-only security scan fallback for swaps whose source chain doesn't support full transaction scanning
- [ ] Wire it in for Solana-source swaps
- [ ] Surface the recipient-scan verdict through the existing scan-result UI

### Must NOT Do
- Don't change the existing EVM-source full-transaction scan path
- Don't change verdict-handling/blocking behavior beyond adding this new signal

### Out of Scope
- Extending the fallback to every non-EVM source chain in this issue — start with Solana, other chains can follow separately

## Acceptance Criteria
- [ ] A Solana-source swap to an external recipient triggers a recipient-only security scan
- [ ] Existing EVM-source swap scanning behavior is unaffected

aminsato:
Closes #5348

## Problem

When a swap's source chain can't be scanned as a full transaction, the verify screen skipped security screening entirely — not even the destination address was checked. A Solana-source swap sending its output to an external recipient reached the sign step with **zero** malicious/sanctioned-address screening, unlike the equivalent EVM-source swap.

## Solution

Add a recipient-only fallback that screens the swap's external recipient on the **destination** chain, independently of whether the source-chain transaction can be scanned:

- `SecurityScannerTransactionFactory.createRecipientSecurityScannerTransaction()` models the recipient as a native transfer on the destination chain (`from` = the vault's destination-chain address, `to` = the external recipient, amount 0, data `0x`).
- `VerifySwapViewModel` falls through to it when the existing source-chain gate fails, publishing the verdict through the same `TransactionScanStatus.Scanned` the EVM path uses — so the existing Blockaid badge and the pre-sign warning sheet pick it up with no UI changes.

Only EVM destinations are modelled: that is the one address-shaped scan Blockaid takes. The Bitcoin/Solana/Sui endpoints need a serialized transaction, which doesn't exist for an outbound leg the vault never signs — those are skipped silently, with provider-side AML screening remaining the backstop. The fallback is also skipped when the swap lands back in the vault's own address, since there is no external address to screen.

Mirrors the iOS implementation (`SecurityScannerTransactionFactory.createRecipientSecurityScanner` / `SecurityScannerViewModel.scanRecipientOnly`).

**The EVM-source full-transaction path is unchanged**, as is verdict-handling and blocking behaviour — the fallback only adds a signal where there was none.

Incidental: the IO dispatcher is now injected rather than hardcoded, matching the send-side `VerifyTransactionViewModel`, so the scan hop runs on the test scheduler and the resulting state is assertable.

## Device verification

Mainnet SwapKit **SOL → ETH** swap with an external recipient (`0x0bd4…eF3Fe6`). The **"Transaction scanned by Blockaid"** badge only renders on `TransactionScanStatus.Scanned`, and a Solana source can never reach the full-transaction path — so the verdict comes from the new recipient-only scan on the Ethereum side. On `main` this same screen shows no scan row at all.

after

## Test plan

- [x] `./gradlew :app:lintDebug :data:lintDebug`
- [x] `SecurityScannerRecipientTransactionTest` (5 cases): EVM destination builds the expected scan; missing/blank recipient rejected; non-EVM destination rejected and carries the chain; missing vault address on the destination chain rejected
- [x] `VerifySwapViewModelTest` (9 cases, 3 new): Solana-source + external EVM recipient → `Scanned`; no external recipient → `NotStarted`, no scan call; non-EVM destination → `NotStarted`, no scan call
- [x] Device: Solana → Ethereum swap with an external recipient shows the Blockaid badge (screenshot above)

## Known limitation

Solana → Bitcoin / Sui / Cosmos destinations remain unscanned — there is no address-shaped scan for those chains at the provider. Extending coverage to them is out of scope here, matching the issue.

## Summary by CodeRabbit

* **New Features**
  * Added recipient-only security scanning for swaps when full source-chain transaction scanning isn’t supported.
  * Supported EVM destination checks using the swap’s external recipient.
  * Security scanning now skips unsupported chains, missing recipients, and unsupported destinations safely.

* **Bug Fixes**
  * Improved security-scan handling for swaps originating on non-EVM chains.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
@@ -5,6 +5,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import androidx.navigation.toRoute
 import com.vultisig.wallet.R
+import com.vultisig.wallet.data.IoDispatcher
 import com.vultisig.wallet.data.api.errors.SwapException
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.Coins
@@ -17,6 +18,7 @@ import com.vultisig.wallet.data.repositories.VaultPasswordRepository
 import com.vultisig.wallet.data.repositories.VaultRepository
 import com.vultisig.wallet.data.securityscanner.BLOCKAID_PROVIDER
 import com.vultisig.wallet.data.securityscanner.SecurityScannerContract
+import com.vultisig.wallet.data.securityscanner.SecurityScannerSupport
 import com.vultisig.wallet.data.securityscanner.isChainSupported
 import com.vultisig.wallet.data.usecases.IsVaultHasFastSignByIdUseCase
 import com.vultisig.wallet.data.utils.safeLaunch
@@ -32,7 +34,7 @@ import com.vultisig.wallet.ui.utils.UiText
 import com.vultisig.wallet.ui.utils.handleSigningFlowCommon
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
-import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.receiveAsFlow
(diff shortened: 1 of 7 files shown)


rkokhatskyi (approved):
LGTM

PR #5387, opened 2026-07-23
feat(ripple): show XRP trust-line token balances in the asset list

### LINKED ISSUE #5210: feat(ripple): show XRP token balances in the asset list
## What
Show XRP trust-line **token balances** in the asset list (reuse the existing token-discovery/asset-list pattern from EVM/Solana/Cosmos — no new UI states).

Android: RippleHelpter.kt / asset list.

Consumes the SDK logic (vultisig-sdk#997). Part of the **XRP token support** feature set (all platforms, low-prio backlog); 1:1 with the SDK issue.

## Acceptance
- [ ] XRP issued-currency tokens appear in the asset list with correct balances (issuer + currency)
- [ ] Reuses existing token/asset-list patterns; no bespoke states

## References
- Fungible tokens: https://xrpl.org/docs/concepts/tokens/fungible-tokens
- Reserves: https://xrpl.org/docs/concepts/accounts/reserves
- account_lines: https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/account-methods/account_lines

aminsato:
## What

XRP issued-currency (trust-line) tokens now appear in the asset list with their balances, discovered from `account_lines`.

Discovery routes through the existing `TokenRepository.getTokensWithBalance` seam that the EVM and Cosmos-bank finders already use, so the chain-detail screen and the background refresh worker pick these up with no bespoke UI states.

Closes #5210 · SDK counterpart: vultisig/vultisig-sdk#997

## Decisions worth reviewing

Android is the first platform to implement this — iOS and the extension have nothing here yet — so the token model was defined in this PR rather than matched to an existing one.

**Identity is `"<currency>.<issuer>"` in `contractAddress`.** An XRPL currency code is only unique per issuer, and the raw on-chain code is stored (3-char ASCII or 40-char hex) so it round-trips into `account_lines` comparisons unchanged; the display ticker is derived separately.

**`Coin.id` is contract-qualified for these.** Several independent issuers each mint their own `USD`. On the plain `ticker-chainId` id they would collide, and Room's REPLACE-on-conflict insert means enabling the second would silently overwrite the first's persisted row — the same hazard THORChain secured assets are already qualified for. Native XRP is untouched.

**Decimals are a wallet-side choice.** Unlike ERC-20/SPL, an XRPL issued currency has no on-chain decimals field: amounts travel as decimal strings with 16 significant digits. Pinned to 15, truncating down, so a displayed balance never rounds up past what the ledger holds.

**Zero and negative lines are dropped.** A zero-balance line is an opt-in with nothing in it (holding one is how an account subscribes to a currency before ever receiving it); a negative balance means the account *issues* the currency rather than holds it.

**No separate owner-reserve work was needed** for the SDK issue's second acceptance item: XRPL's `OwnerCount` already counts trust lines, so the existing native-XRP reserve subtraction in `RippleApi.getBalance` already folds them into spendable XRP.

**`account_lines` reads are coalesced.** One response carries every line for an account, but the balance layer asks per token — N held currencies would otherwise fire N identical paginated request chains at the public cluster. `RippleAccountLinesCache` uses a per-address mutex + 10s TTL, mirroring the existing `CosmosBalanceCache`.

## Slightly beyond the ticket: these assets are read-only

`Ripple.OperationPayment` carries a **drops-denominated** amount, so the payment builder can only move native XRP. Now that these tokens are reachable in the asset list, routing one into send would have signed away `toAmount` *drops of XRP* instead of the token the user picked — a silent loss of unrelated funds.

Moving an issued currency needs a `Payment` carrying a `CurrencyAmount` (currency + issuer + value), which is separate work. Until then they are gated: the token-detail send/swap buttons and both asset pickers are closed via `Coin.isReadOnlyAsset`, and `RippleHelper` refuses outright so no future entry point can reach the drops path unnoticed. That guard sits *after* the dApp raw-JSON branch, which signs verbatim and stays free to submit `TrustSet` and issued-currency payments.

Shipping the display without this would have been unsafe, so it is included rather than deferred.

## Test plan

- `:data:testDebugUnitTest` and `:app:testDebugUnitTest` — full suites green
- `:app:lintDebug` green; ktfmt applied
- 36 new tests: `RippleAccountLinesTest` (11 — parsing, marker pagination, repeated-marker bail-out, per-token lookup, issuer disambiguation, read coalescing), `RippleTokenTest` (17 — identity round-trip, id collisions, hex currency decoding in both shapes, fixed-point scaling and truncation), `RippleTokenFinderTest` (8 — zero/negative/XRP filtering, two issuers of one currency, failure tolerance), plus a Ripple-delegation case in `TokenRepositoryImplTest`
- Each of the three commits compiles independently (`:data` + `:app`)

**Not verified on device or mainnet.** Reaching this state needs a vault whose XRP account holds a trust line with a positive balance, which requires a `TrustSet` the app has no screen for — it has to come through the extension's XRPL dApp path. Reviewers with such an account, the two things to eyeball are the balance precision on a fractional holding and two issuers of the same currency code rendering as separate rows.

## Summary by CodeRabbit

* **New Features**
  * Added XRPL issued-token (trust-line) discovery and balance conversion, including RLUSD.
  * Added support for detecting issued tokens as read-only assets.

* **Bug Fixes**
  * Excluded read-only assets from token selection and fast token pickers.
  * Hid the Send action for read-only assets and tightened send capability gating.
  * Improved native XRP vs issued-token balance handling, with fail-closed protection for unsupported issued-currency transfers.

* **Tests**
  * Added/expanded coverage for trust-line parsing, pagination/caching, token discovery, balance conversion, and read-only behavior.


app/src/main/java/com/vultisig/wallet/ui/models/TokenDetailViewModel.kt
```diff
@@ -38,6 +39,7 @@ internal data class TokenDetailUiModel(
     val isRefreshing: Boolean = false,
     val canDeposit: Boolean = false,
     val canSwap: Boolean = false,
+    val canSend: Boolean = true,
```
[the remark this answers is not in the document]
aminsato: Applied in a6a4be051 — `canSend` now defaults to `false`, matching `canDeposit`/`canSwap`, so SEND stays hidden until a matching account resolves.

data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
```diff
@@ -112,6 +171,58 @@ internal class RippleApiImp @Inject constructor(private val http: HttpClient) :
...
+                    .post(BASE_XRP_CLUSTER) { setBody(payload) }
+                    .bodyOrThrow<RippleAccountLinesResponseJson>()
+                    .result
+
+            // An unfunded account answers with `actNotFound` and no `lines` — a legitimately empty
+            // holding, not a failure, so it falls through to an empty list.
+            result?.lines?.let(lines::addAll)
+
+            val next = result?.marker ?: return lines
+            if (next == marker) return lines
```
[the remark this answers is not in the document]
aminsato: Applied in a6a4be051. `fetchAllAccountLinePages` now throws `RippleRpcException` on any `result.error` other than `actNotFound`, which stays the legitimate empty-account case. Because the throw escapes `RippleAccountLinesCache.getOrFetch` before the entry is written, the failure is not cached — covered by two new tests in `RippleAccountLinesTest`: one asserting the throw on `slowDown`, one asserting the next read still fetches fresh lines.

data/src/main/kotlin/com/vultisig/wallet/data/models/RippleToken.kt
```diff
@@ -0,0 +1,120 @@
...
+    return parseRippleTokenIdentity(contractAddress)
+}
+
+fun parseRippleTokenIdentity(contractAddress: String): RippleTokenIdentity? {
+    val separator = contractAddress.indexOf(RIPPLE_TOKEN_SEPARATOR)
+    if (separator <= 0 || separator == contractAddress.lastIndex) return null
+    return RippleTokenIdentity(
+        currency = contractAddress.substring(0, separator),
+        issuer = contractAddress.substring(separator + 1),
+    )
```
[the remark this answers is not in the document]
aminsato: Applied in a6a4be051 — `parseRippleTokenIdentity` now rejects any address with more than one separator, with a `USD.<issuer>.extra` case added to `RippleTokenTest`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RippleTokenFinder.kt
```diff
@@ -0,0 +1,94 @@
...
+                Timber.e(e, "XRPL account_lines timed out")
+                return emptyList()
+            } catch (e: NetworkException) {
+                Timber.e(e, "XRPL account_lines failed: status=%d", e.httpStatusCode)
+                return emptyList()
+            } catch (e: CancellationException) {
+                throw e
+            } catch (e: Exception) {
+                Timber.e(e, "XRPL account_lines failed")
+                return emptyList()
```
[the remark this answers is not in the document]
aminsato: Keeping this one, deliberately. `CosmosBankCoinFinder` — the sibling this finder is modelled on — has the same `CancellationException` rethrow followed by a broad catch, for the same reason: `getTokensWithBalance` is consumed by `GetChainTokensUseCase`, which does not guard the Ripple branch, so anything escaping here propagates through the flow and breaks the whole Choose Tokens screen rather than degrading to "no discovered tokens". Discovery is best-effort by contract and the failure is logged at error level, so a regression is still visible in logs without taking the screen down.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RippleTokenFinder.kt
```diff
@@ -0,0 +1,94 @@
...
+                chain = Chain.Ripple,
+                ticker = rippleCurrencyTicker(currency),
+                logo = "",
+                address = "",
+                decimal = RIPPLE_TOKEN_DECIMALS,
+                hexPublicKey = "",
+                priceProviderID = "",
+                contractAddress = contractAddress,
+                isNativeToken = false,
+            )
```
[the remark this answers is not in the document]
aminsato: Not a defect — the address is already applied one layer up. `getTokensWithBalance` is only reached through `TokenRepositoryImpl.getRefreshTokens`, which ends with `.map { token -> token.copy(address = address, hexPublicKey = derivedPublicKey) }` (TokenRepository.kt:263), stamping the derived vault address and public key onto every discovered coin before it reaches the balance layer. `EvmCoinFinder` and `CosmosBankCoinFinder` return address-less coins for the same reason; setting it inside the finder would duplicate that and leave `hexPublicKey` empty regardless.

rkokhatskyi (changes requested):
I think we should do the full support for tokens, not just discovery. This is incomplete feature and needs more work. Few things in mind: 

1. Sending XRPL tokens doesn't work
2. We need to control the list of tokens as anyone can create a trustline, call it USDT and you don't know where's correct one. XRP doesn't have many tokens, we could add them all
3. No pricing for XRPL tokens
4. We cache balances keys on `(chain, address, ticker)`, not the full issuer-qualified id, so two same-ticker issuers can collide (REPLACE-on-conflict)

aminsato:
Thanks Roman — agree the full token experience is the goal. I've split it so this PR lands correct, safe discovery now, with sending and pricing as the next steps.

**Addressed in this PR (749eb1418):**

**2. Controlled token list.** Discovery is now gated to the curated `Coins` catalog. The finder surfaces a trust line only when its `currency/issuer` pair matches a catalog entry and drops everything else, so a stranger's line labelled `USD`/`USDT` — or the right currency code from the wrong issuer — never shows. New tokens are onboarded by adding them to `Coins.Ripple.all` (with logo + `priceProviderID`); RLUSD is in there today and I'll expand the list. Covered by `RippleTokenFinderTest`.

**4. Cache-key collision.** Root cause was `TokenValueEntity.tokenId`: it only issuer-qualified THORChain secured assets, so for Ripple it collapsed to the plain `ticker-chain` id and no longer mirrored the already-issuer-qualified `Coin.id`. Two issuers of the same currency then resolved to one balance row and `BalanceRepository.getCachedTokenBalances`' `coins.find { it.id == tokenEntity.tokenId }` missed (decimals → 0). `tokenId` now mirrors `Coin.id` for issued currencies, so each issuer keeps its own row. Covered by `TokenValueEntityTest` (two issuers → distinct ids; native XRP stays plain).

**Follow-ups (separate PRs under the XRP token support set):**

**1. Sending XRPL tokens.** Needs a `Payment` carrying a `CurrencyAmount` (currency + issuer + value) rather than the drops-only path `RippleHelper` builds today. That's why issued currencies are gated read-only for now — routing one into send/swap would sign an XRP transfer of the numeric balance. I'll build the issued-currency signing path as its own PR and lift the read-only gate there.

**3. Pricing.** I'll wire XRPL issued currencies into the price provider (keyed per issuer via each catalog entry's `priceProviderID`) so discovered tokens show a fiat value, as a follow-up.

Keeping this PR to correct, curated, read-only discovery keeps it reviewable; happy to track 1 and 3 as linked issues if you'd like.

data/src/main/kotlin/com/vultisig/wallet/data/models/Coin.kt
```diff
@@ -71,6 +77,18 @@ val Coin.isLpToken: Boolean
...
+/**
+ * True when the wallet can read this coin's balance but cannot yet move it.
+ *
+ * XRPL issued currencies are the only such coins today: `account_lines` gives their balances, but
+ * transferring one needs a `Payment` carrying a `CurrencyAmount` (currency + issuer + value), while
+ * [com.vultisig.wallet.data.chains.helpers.RippleHelper] only builds drop-denominated XRP payments.
+ * Routing one into send or swap would sign an XRP transfer of the token's numeric balance, so both
+ * actions stay closed until issued-currency signing lands.
+ */
+val Coin.isReadOnlyAsset: Boolean
```
rkokhatskyi, line 89: `isReadOnlyAsset` gates the three pickers but not send preselection: `TokenPreselectionService.kt:85` matches on raw `Coin.id` equality, and `VaultListViewModel.kt:157` builds that id from unvalidated deep link params, so a crafted vultisig://send link pins RLUSD as the send source until keysign throws

data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
```diff
@@ -112,6 +171,66 @@ internal class RippleApiImp @Inject constructor(private val http: HttpClient) :
...
+        // same marker can't spin here forever.
+        repeat(MAX_ACCOUNT_LINES_PAGES) {
+            val payload =
+                RpcPayload(
+                    method = "account_lines",
+                    params =
+                        buildJsonArray {
+                            addJsonObject {
+                                put("account", walletAddress)
+                                put("ledger_index", "validated")
```
rkokhatskyi: Marker paging resends the `validated` shorthand on every page instead of pinning page one's `ledger_index`, which XRPL documents as returning incomplete results. A trust line dropped at a page boundary then reads as zero and gets persisted by `BalanceRepository.kt:592`

data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
```diff
@@ -112,6 +171,66 @@ internal class RippleApiImp @Inject constructor(private val http: HttpClient) :
...
+                                put("limit", ACCOUNT_LINES_PAGE_SIZE)
```
rkokhatskyi, line 214: This walk pulls up to 10000 lines while the only consumer keeps curated issuers at `RippleTokenFinder.kt:83`. Anyone can inflate an account's line count by opening trust lines against it, so `ignore_default: true` filters those server side in one request

data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
```diff
@@ -112,6 +171,66 @@ internal class RippleApiImp @Inject constructor(private val http: HttpClient) :
         maxOf(balance - accountReservedBalance, BigInteger.ZERO)
     }
 
+    override suspend fun getTokenBalance(coin: Coin): BigInteger {
+        val identity = coin.rippleTokenIdentity() ?: return BigInteger.ZERO
+        val line =
+            fetchAccountLines(coin.address).firstOrNull { it.matches(identity) }
+                ?: return BigInteger.ZERO
+        return line.balance.toRippleTokenUnits()
```
rkokhatskyi: `getTokenBalance` scales every line at the fixed `RIPPLE_TOKEN_DECIMALS` while the value is rendered at `coin.decimal` in `BalanceRepository.kt:588`, and nothing ties them together. The coin is already in hand here, so scaling by `coin.decimal` removes the unwritten invariant

data/src/main/kotlin/com/vultisig/wallet/data/models/RippleToken.kt
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Human-readable ticker for a raw XRPL currency code.
+ *
+ * Standard codes are three ASCII characters and are used verbatim. Longer names are carried as the
+ * hex of a 160-bit code whose first byte is `0x00`, followed by the zero-padded ASCII name — decode
+ * those back to text. Anything else (demurrage or XLS-14 codes, unprintable payloads) has no
+ * meaningful text form, so a short hex prefix keeps the row distinguishable.
+ */
+fun rippleCurrencyTicker(currency: String): String {
```
rkokhatskyi: `rippleCurrencyTicker` has no production caller because `RippleTokenFinder.kt:70` returns curated catalog coins whose ticker is hardcoded, so it plus `isHexDigit` and both hex constants are unreachable outside tests while `RippleApi.kt:398` still names it the display path

data/src/main/kotlin/com/vultisig/wallet/data/models/RippleToken.kt
```diff
@@ -0,0 +1,124 @@
...
+        currency = contractAddress.substring(0, separator),
+        issuer = contractAddress.substring(separator + 1),
+    )
+}
+
+/**
+ * Human-readable ticker for a raw XRPL currency code.
+ *
+ * Standard codes are three ASCII characters and are used verbatim. Longer names are carried as the
+ * hex of a 160-bit code whose first byte is `0x00`, followed by the zero-padded ASCII name — decode
```
rkokhatskyi: This inverts the XRPL rule: a `0x00` lead byte marks the standard three character form, and a longer name must never start with it, which RLUSD at `Coins.kt:2280` shows by beginning `0x52`. State it that way so the next catalog entry is encoded correctly

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/models/Coin.kt
aminsato: Fixed in d497b467a. `findPreselectedToken` now skips `isReadOnlyAsset` coins on both the id-match and the first-token fallback, so the unvalidated deep-link token id can no longer pin an XRPL issued currency as the send source.

data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
aminsato: Fixed in d497b467a. Added a `ledger_index` field to `RippleAccountLinesResultJson`; page one still sends the `validated` shorthand, then every later page pins the concrete index that first page resolved to, so a marker is never resumed against a shifted ledger. Covered by a new test asserting the second request carries the pinned index.

data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
aminsato: Fixed in d497b467a. The request now sends `ignore_default: true`, so default (zero-balance) trust lines are filtered server-side — anyone opening throwaway lines against the account no longer inflates the paged walk. Same test asserts the flag is present.

data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
aminsato: Fixed in d497b467a. `toRippleTokenUnits` now takes a `decimals` parameter and `getTokenBalance` passes `coin.decimal`, so the parsed units are scaled at the same decimals the balance is rendered at instead of a standalone constant. Behaviour is unchanged for RLUSD (its catalog `decimal` is already `RIPPLE_TOKEN_DECIMALS`), but the invariant is now explicit and travels with the coin.

data/src/main/kotlin/com/vultisig/wallet/data/models/RippleToken.kt
aminsato: Fixed in d497b467a. Removed `rippleCurrencyTicker` along with `isHexDigit` and both hex constants — they had no production caller (the finder returns curated catalog coins with hardcoded tickers), only tests referenced them. Updated the `RippleApi.kt` doc that named it the display path.

data/src/main/kotlin/com/vultisig/wallet/data/models/RippleToken.kt
aminsato: Resolved by deletion in d497b467a — the inverted `0x00` note lived in `rippleCurrencyTicker`, which is now gone. RLUSD stays encoded from its ASCII lead byte (`0x52`) via `rippleTokenContractAddress` in the catalog.

data/src/main/kotlin/com/vultisig/wallet/data/api/RippleApi.kt
```diff
@@ -112,6 +171,88 @@ internal class RippleApiImp @Inject constructor(private val http: HttpClient) :
...
+                                val pinned = pinnedLedgerIndex
+                                if (pinned != null) {
+                                    put("ledger_index", pinned)
+                                } else {
+                                    put("ledger_index", "validated")
+                                }
+                                put("limit", ACCOUNT_LINES_PAGE_SIZE)
+                                // Trust lines still in their default state hold nothing and are
+                                // dropped downstream anyway; filtering them server-side keeps the
+                                // walk short when anyone spams zero-balance lines against the
```
rkokhatskyi: `ignore_default` does not shorten the walk: rippled increments the page counter and pins the marker before the filter runs (`AccountLines.cpp:207` ahead of `:229`), so the marker comes back identical with and without it. This wording should say it shrinks the response only

rkokhatskyi (changes requested):

aminsato:
@Vaulty-bot pick-up

aminsato:
@Vaulty-bot pick-up

rkokhatskyi (approved):
LGTM

PR #5389, opened 2026-07-23
fix(folder): vertically center vault name in Edit Folder rows

### LINKED ISSUE #5388: Vault name not vertically centered in Edit Folder rows
## Description

On the **Edit Folder** screen (add/remove vaults from a folder), the vault **name** in each row under **Active Vaults** is not vertically centered — it sits slightly above the row's center, misaligned with the vault icon and the toggle.

## Root cause

The active-vault rows build a `VaultCeilUiModel` **without** a `balance`. In `VaultCeil`, the subtitle line is still rendered as `Text(text = model.balance ?: "")` — an empty, invisible line that nonetheless reserves a full footnote line-height plus a 2.dp spacer. Because the row is laid out `CenterVertically`, it centers the whole `Column` (name + invisible subtitle), which pushes the visible name above the icon's center.

The **Available** row has a real balance (e.g. `US$3.14`), so it shows both lines and looks balanced — which is why only the active rows appear off.

## Expected

The vault name should be vertically centered within each row, aligned with the vault icon and toggle.

## Steps to reproduce

1. Open a folder → tap edit (pencil)
2. Observe the rows under **Active Vaults**: the names are shifted upward relative to the icon/toggle

## Screenshot

See the misaligned names ("Secure Vault #1", "Ahmad-Ehsan", "sss", "Fast-DKLS") sitting above row center.

aminsato:
## Summary

On the **Edit Folder** screen, the vault **name** in each **Active Vaults** row was not vertically centered — it sat above the row's center, misaligned with the vault icon and toggle.

### Root cause
The active-vault rows build a `VaultCeilUiModel` **without** a `balance`. `VaultCeil` still rendered the subtitle as `Text(text = model.balance ?: "")` — an empty, invisible line that reserved a full footnote line-height plus a 2.dp spacer. Because the row is laid out `CenterVertically`, it centered the whole `Column` (name + invisible subtitle), pushing the visible name above the icon's center. The **Available** row has a real balance, so it looked balanced — which is why only the active rows appeared off.

### Fix
Only render the subtitle line (and its spacer) when there is something to show — an active-vault name, a folder count, or a non-empty balance. Balance-less rows now collapse to just the name and center correctly. Folder rows and vaults with a balance are unchanged (`hasSubtitle` stays true).

### Also
- Exposed the inner `CreateFolderScreen` overload as `internal` and added an `edit_folder` debug preview in `PreviewActivity` to render the screen with mock data for before/after capture.

## Before / After

| Before | After |
|--------|-------|
| before | after |

## Test plan
- Open a folder → tap edit (pencil) → verify each **Active Vaults** row's name is vertically centered with the icon and toggle.
- Verify the **Available** row (with a balance) and folder rows are unchanged.

Closes #5388

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed vault row layout so subtitle areas are shown only when relevant information is available.
  * Improved vertical centering for vault rows without balances.

* **Preview**
  * Added an edit-folder preview showcasing vault rows with and without balances for easier visual verification.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -55,6 +55,7 @@ import com.vultisig.wallet.data.usecases.GenerateQrBitmap
 import com.vultisig.wallet.data.usecases.MakeQrCodeBitmapShareFormat
 import com.vultisig.wallet.data.usecases.QrShareField
 import com.vultisig.wallet.data.usecases.QrShareInfo
+import com.vultisig.wallet.data.usecases.VaultAndBalance
 import com.vultisig.wallet.ui.components.SignMessageCard
 import com.vultisig.wallet.ui.components.SignTonDisplayView
 import com.vultisig.wallet.ui.components.UiIcon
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

Issue #1541, opened 2025-01-06
[BUG] Signing TON creates error


realpaaao:
**Describe the bug**
When co-signing TON and initating from windows, this error message gets displayed even though the transaction is broadcasted

**To Reproduce**
1. Send TON
2. Wait keysign to complete
3. See error

**Expected behavior**
Just display error if broadcast was not possible

**Screenshots**
![Image](https://github.com/user-attachments/assets/c4de006a-8702-4a9c-a302-0a49e4785eeb)

 - Version 36

PR #5390, opened 2026-07-24
feat: add Robinhood Chain (Arbitrum Orbit L2, chain id 4663)

realpaaao:
## Summary
- Adds Robinhood Chain following the Sei/Mantle template: enum entry, `CoinType.ROBINHOODCHAIN` (wallet-core 4.7.0), direct mainnet RPC, Blockscout explorer, canonical Multicall3 (deployment verified on-chain), custom-RPC eligibility, tx-status polling, zero priority fee (Orbit sequencer ignores tips).
- Swaps: LiFi + KyberSwap, both live-verified on 4663 (Kyber slug derives from the enum raw name; router matches the standard MetaAggregationRouterV2). 1inch off; Blockaid stays list-gated off pending their 4663 coverage.
- Token catalog: native ETH + USDG/USDe/WETH/LINK (verified on-chain) + all 96 Robinhood stock tokens with per-stock icons; `EvmCoinFinder`'s curated-catalog discovery now reads balances through the Multicall3-batched `EvmApi.getBalances` (~100 RPCs → 1) instead of per-token calls.

## Changes
| Area | Change |
|------|--------|
| `models/Chain.kt`, `Coins.kt`, `Vault.kt` | enum + coinType + swap/ticker maps; native ETH + majors + `STOCK_TOKENS` |
| `api/CustomRpcDefaultEndpoint.kt`, `RpcHealthProbe.kt`, `models/CustomRpcSupportedChains.kt` | RPC endpoint + health probe id + custom-RPC eligibility |
| `chains/helpers/Multicall3.kt`, `SigningHelper.kt` | canonical Multicall3 group; EVM signing groups |
| `usecases/EvmCoinFinder.kt` (+test) | curated-catalog discovery batched via `getBalances` |
| `repositories/*` (Balance/ExplorerLink/SwapProviderTable) | balance dispatch, Blockscout links, LIFI+KYBER provider set |
| `usecases/BroadcastTxUseCase.kt`, `txstatus/TxStatusConfiguration.kt` | broadcast + polling config |
| `blockchain/.../EthereumFeeService` (fee logic) | zero priority fee on Robinhood |
| `ChainLogo.kt` + drawables | official chain mark (color + mono, rounded disc) |

## Context
Chain-detail discovery on chains without 1inch coverage scans the curated catalog for balances — an empty catalog meant nothing could be discovered. Stock tokens come from Robinhood's public registry; regeneration tooling lives in vultisig-sdk (`scripts/generate-robinhood-stock-tokens.mjs`). Companion PRs: vultisig-sdk#1541, vultisig-ios#4898. Note: this branch aligns the vendored `commondata` checkout to the already-recorded `c91f2ea` pointer (SignRipple protos) — no pointer change.

## Test plan
- [x] `./gradlew assembleDebug testDebugUnitTest` green (JDK 21)
- [x] Forced test re-run (`--rerun`) — both `:data` and `:app` unit suites executed against the final tree
- [ ] Manual chain-detail pass on device (add chain → discover tokens → LiFi/Kyber quote) — screenshots to follow

## receipts
```
BUILD SUCCESSFUL  # assembleDebug testDebugUnitTest
> Task :data:testDebugUnitTest   # executed, green (forced --rerun)
> Task :app:testDebugUnitTest    # executed, green (forced --rerun)
```
Multicall3 canonical deployment verified on-chain before enabling batched balances:
```
$ eth_getCode 0xcA11bde05977b3631167028862bE2a173976CA11 → 3808 bytes — DEPLOYED
```
No screenshots yet — flagged in the test plan rather than substituted.
https://claude.ai/code/session_012boHdofAN4VPBhAFcionHq

## Summary by CodeRabbit

* **New Features**
  * Added support for the Robinhood network, including logos/icons, token and stock-token listings, balance discovery, transaction signing/broadcasting, and transaction status tracking.
  * Added Robinhood RPC defaults and explorer links, plus Robinhood to multcall support for improved on-chain querying.
  * Enabled swaps on Robinhood with the supported provider set.
* **Bug Fixes**
  * Fixed Robinhood fee behavior to use the correct EIP-1559 priority fee handling.
  * Corrected Robinhood asset-id, chain-id, and explorer/contract mapping behaviors for better integrations.
* **Tests**
  * Expanded Robinhood coverage for routing, fees, swap eligibility, and hybrid token discovery.

On chain tx link: https://robinhoodchain.blockscout.com/tx/0x4aedf3719f3e104aa2835c62667f693268333c2755de892688de5b786695514c


realpaaao:
@coderabbitai review

rkokhatskyi:
@realpaaao please attach tx sample link

data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt
```diff
@@ -2400,6 +2400,1128 @@ object Coins {
...
+                ticker = "LINK",
+                logo = "link",
+                address = "",
+                decimal = 18,
+                hexPublicKey = "",
+                priceProviderID = "chainlink",
+                contractAddress = "0x492641F648a4986844848E0beFE66D14817bCE34",
+                isNativeToken = false,
+            )
+        val STOCK_TOKENS = listOf(
```
rkokhatskyi, line 2487: The 96 stock tokens added here implement the ERC 8056 scaled UI extension but balances are read with the raw `balanceOf` selector at `Multicall3.kt:24`. CRWD's uiMultiplier is 4.0 on chain 4663 today, so a 0.6 share position renders as 0.15 everywhere

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapProviderTable.kt
```diff
@@ -118,6 +118,8 @@ constructor(private val poolEligibility: SwapPoolEligibilityRepository) : SwapPr
 
             Chain.Mantle -> setOf(SwapProvider.LIFI, SwapProvider.KYBER)
 
+            Chain.Robinhood -> setOf(SwapProvider.LIFI, SwapProvider.KYBER)
```
rkokhatskyi: LIFI is enabled here but `Chain.kt:349` oneInchChainId has no Robinhood arm, so `LiFiQuoteSource.kt:31` throws SwapRouteNotAvailable before any request. Cross chain pairs drop KYBER as same chain only, leaving no provider even though LI.FI lists chain 4663

data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt
```diff
@@ -3296,6 +4418,7 @@ object Coins {
             Chain.Zcash to Zcash.all,
             Chain.Cardano to Cardano.all,
             Chain.Sei to Sei.all,
+            Chain.Robinhood to Robinhood.all,
```
rkokhatskyi, line 4444: The 96 stock tokens carry an empty priceProviderID so they price by contract address, but `CoinGeckoApi.kt:98` has no Robinhood arm and the error is swallowed into an empty map, so every stock token stays at $0.00 while CoinGecko serves platform robinhood

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
```diff
@@ -206,6 +208,7 @@ val Chain.isSwapSupported: Boolean
                 Chain.Cardano,
                 Chain.Ton,
                 Chain.Hyperliquid,
+                Chain.Robinhood,
```
rkokhatskyi, line 218: Robinhood is missing from isLayer2 at `Chain.kt:319`, so `VsOverviewToken.kt:77` suppresses the chain badge on its native ETH and the transaction done card renders it identically to mainnet ETH, unlike every other native ETH L2

data/src/main/kotlin/com/vultisig/wallet/data/usecases/EvmCoinFinder.kt
```diff
@@ -92,18 +89,8 @@ constructor(private val oneInchApi: OneInchApi, private val evmApiFactory: EvmAp
...
-                    async {
-                        coin.takeIf {
-                            balanceOrZero(evmApi, address, coin.contractAddress) > BigInteger.ZERO
-                        }
-                    }
-                }
-                .awaitAll()
-                .filterNotNull()
-        }
+        val balances = evmApi.getBalances(address, curated.map { it.contractAddress })
```
rkokhatskyi, line 112: The KDoc chain list at `EvmCoinFinder.kt:26` enumerates the chains that take this fallback and omits Robinhood, which is absent from `ONE_INCH_SUPPORTED_CHAINS` at line 160, so the list no longer tells a maintainer which chains depend on the curated catalog

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
```diff
@@ -36,6 +36,7 @@ enum class Chain(val raw: ChainId, val standard: TokenStandard, val feeUnit: Str
     Mantle("Mantle", EVM, "Gwei"),
     Sei("Sei", EVM, "Gwei"),
     Hyperliquid("Hyperliquid", EVM, "Gwei"),
+    Robinhood("Robinhood", EVM, "Gwei"),
```
rkokhatskyi, line 39: The `blockTimeMs` KDoc at `Chain.kt:475` says every EVM chain supported today is listed explicitly, but Robinhood has no arm in the when below it and takes the 4000ms fallback, so approval polling runs 4x slower than on Arbitrum

rkokhatskyi (changes requested):

johnnyluo:
Please provide onchain tx link

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapProviderTable.kt
realpaaao: Confirmed and fixed in 04d4e1a8 — `Chain.Robinhood -> 4663` added to `oneInchChainId()`.

Verified LI.FI genuinely serves the chain before wiring it rather than dropping LIFI from the table:

```
$ curl -s https://li.quest/v1/chains | jq '.chains[] | select(.id==4663)'
{ "key": "out", "name": "Robinhood Chain", "id": 4663, "mainnet": true,
  "diamondAddress": "0xB477751B76CF82d00a686A1232f5fCD772414Af3",
  "permit2": "0x000000000022D473030F116dDEE9F6B43aC78BA3",
  "multicallAddress": "0xcA11bde05977b3631167028862bE2a173976CA11" }
```

One correction to the "leaving no provider" part: KyberSwap does serve the chain today, so same-chain pairs were working —

```
$ curl -s "https://aggregator-api.kyberswap.com/robinhood/api/v1/routes?tokenIn=0xeee…eee&tokenOut=0x5fc5360D0400a0Fd4f2af552ADD042D716F1d168&amountIn=1000000000000000"
{"code":0,"message":"successfully","data":{"routeSummary":{"amountInUsd":"1.8575093085714678","amountOut":"1854939","amountOutUsd":"1.8571449096019483", …
```

`KyberApi` keys off `chain.raw.lowercase()`, not `oneInchChainId()`, which is why it was unaffected. So the real gap was cross-chain pairs, where LI.FI was the only provider and threw before issuing a request. That's now closed.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
realpaaao: Confirmed and fixed in 04d4e1a8 — `Chain.Robinhood` added to `isLayer2`.

Traced it to the exact site you flagged: `VsOverviewToken.kt` gates the badge on
`chainLogo.takeIf { !token.isNativeToken || token.chain.isLayer2 }`, so native ETH on 4663 was
rendering with no chain badge and was indistinguishable from mainnet ETH on the done card.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/EvmCoinFinder.kt
realpaaao: Confirmed and fixed in 04d4e1a8 — the KDoc now reads
`Blast / Cronos / Hyperliquid / Mantle / Robinhood / Sei / zkSync`.

Robinhood is absent from `ONE_INCH_SUPPORTED_CHAINS`, so it does take the curated-catalog
fallback, and the list was the only thing telling a maintainer that.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
realpaaao: Confirmed and fixed in 04d4e1a8 — `Chain.Robinhood` added to the 1s arm.

Measured the real cadence rather than assuming Orbit defaults:

```
$ # eth_getBlockByNumber deltas on rpc.mainnet.chain.robinhood.com
18790595->18791595:  1000 blocks /  101s = 101 ms/block
18771595->18790595: 19000 blocks / 1913s = 101 ms/block
```

101 ms/block, so the 4s fallback was 40x the real cadence. Grouped with Arbitrum at `1_000L`
rather than adding a sub-second tier, since `MIN_POLL_DELAY_MS` floors everything at 2s anyway and
Arbitrum (~250ms) is already rounded the same way.

Worth noting *why* this slipped: the guard test enumerated chains by name in `@EnumSource`, so a
newly added chain was simply never covered. Replaced it with a real exhaustiveness guard so the
KDoc's claim is enforced rather than asserted:

```kotlin
@Test
fun `every EVM chain has an explicit block time arm`() {
    // 4s is the fallback, not a real cadence — hitting it means a new EVM chain has no arm.
    val onFallback =
        Chain.entries.filter {
            it.standard == TokenStandard.EVM && it.blockTimeMs == 4_000L
        }

    assertTrue(onFallback.isEmpty(), "EVM chains missing a blockTimeMs arm: $onFallback")
}
```

Verified all 14 EVM chains have explicit arms today, so it passes by construction and fails the
build on the next omission.

data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt
realpaaao: Confirmed and fixed in 04d4e1a8 — `Chain.Robinhood -> "robinhood"` added to `coinGeckoAssetId`.

The mechanism was exactly as you describe: `coinGeckoAssetId` had no arm, so `error()` threw,
`getContractsPrice` caught it into `emptyMap()`, and the failure was invisible. CoinGecko does
serve the platform:

```
$ curl -s https://api.coingecko.com/api/v3/asset_platforms | jq '.[] | select(.id=="robinhood")'
{ "id": "robinhood", "chain_identifier": 4663, "name": "Robinhood", "native_coin_id": "ethereum" }
```

One correction on the blast radius, since it changes how urgent this is — "every stock token stays
at $0.00" isn't quite right. I measured coverage for all 100 contract-priced tokens through the
proxy we actually ship against:

```
PRICED   97/100
UNPRICED 3/100: CRWD, SKHY, LINK
```

LINK is a false positive there — it carries `priceProviderID = "chainlink"`, so
`TokenPriceRepository` routes it through `getCryptoPrices` by id and never hits the contract path.
So the real outcome is **94 of 96 stock tokens price correctly** once the arm exists; only CRWD and
SKHY have no CoinGecko entry (and neither has a Chainlink feed on 4663 either, so no oracle would
recover them). Those two will read $0.00 — happy to hide zero-priced tokens or drop them from the
catalog if you'd prefer, just say which.

Also worth flagging for cross-client consistency: iOS #4898 (`case .robinhood: return "robinhood"`)
and SDK #1541 (`[EvmChain.Robinhood]: 'robinhood'`) already carried their equivalent arms — Android
was the only client missing it, which is why this looked chain-wide rather than platform-specific.

data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt
realpaaao: Good catch on the extension being live — I verified that part directly. `uiMultiplier()` is real on
these contracts, and CRWD is 4.0:

```
$ # eth_call uiMultiplier() [0xa60bf13d] on 0xea72Ecca…4233931 (CRWD), chain 4663
0x0000000000000000000000000000000000000000000000003782dace9d900000   = 4.0

$ # swept all 96 stock tokens — non-unity multipliers:
CRWD 4.0   MU 1.0000748   SGOV 1.0009575    (the other 93 are exactly 1.0)
```

But I don't think the conclusion follows, and I'd like to leave the balance path as-is. The reason
is that the price feeds on this chain are quoted **per token, not per share** — Robinhood's own
integration docs are explicit:

> The feed returns the price of one token, which is the underlying share price times the multiplier.
> `latestRoundData()` returns this directly, so you don't apply the multiplier yourself.

I checked that our actual price sources agree with the on-chain oracle, using AAPL (multiplier 1.0)
to calibrate:

| source | AAPL |
|---|---|
| Chainlink `latestRoundData()` on 4663 (`0x6B22A786…f2cD0`) | $333.22 |
| CoinGecko `robinhood` platform | $331.98 |
| LI.FI | $332.02 |

Sub-0.4% spread, i.e. ordinary inter-source drift. Since the Chainlink feed is per-token by
definition and the other two track it, `rawBalance × price` already yields the correct fiat value
and applying the multiplier a second time would *introduce* a 4x error on CRWD rather than remove
one. The docs land on the same point:

> The economic value is identical in every case — only the presentation differs.

So what's left is genuinely a presentation choice — "0.15 CRWD tokens" vs "0.6 share-equivalents" —
and the docs list both as valid conventions (`balance * uiMultiplier() / 1e18` is the third of
three). Everything downstream stays in raw token units end to end, so send/swap/approve remain
internally consistent.

Two things that make me want to defer it rather than build it here:

1. CRWD has no CoinGecko entry and no Chainlink feed on 4663, so it reads $0.00 regardless of how we
   handle the multiplier. The one token where the 4x is visible is also the one we can't price.
2. Doing it properly means scaling on read *and* unscaling on send/swap/approve. That's an
   irreversible path, and I'd rather it be its own PR with its own test pass than ride along here.

Happy to be wrong on this — if you're reading the feed semantics differently, or you'd rather we
adopt the share-equivalent convention for display, I'll open a follow-up and implement it there.

realpaaao:
## Review fixes — 04d4e1a8

Five of the six comments are addressed; the sixth (`uiMultiplier`) I've replied to with a case for
leaving the balance path alone. Details are in each thread.

| Comment | Resolution |
|---|---|
| `oneInchChainId` missing → LI.FI throws | Fixed — `Chain.Robinhood -> 4663` |
| `coinGeckoAssetId` missing → prices swallowed | Fixed — `Chain.Robinhood -> "robinhood"` |
| `isLayer2` missing → no chain badge on native ETH | Fixed |
| `blockTimeMs` missing → 4s fallback | Fixed — 1s arm + real exhaustiveness guard |
| `EvmCoinFinder` KDoc omits Robinhood | Fixed |
| ERC-8056 `uiMultiplier` not applied to balances | Pushing back — feeds are per-token, see thread |

## Scope

**Changed:** four lookup tables that key off `Chain` and silently degraded when Robinhood was absent
from them, plus the block-time guard that let the omission through.

**Deliberately not touched:**
- `LiQuestApi.lifiChainId` maps Ethereum only, so the LI.FI price fallback is dead for every chain,
  not just this one. Pre-existing and out of scope here — CoinGecko covers 94/96 stock tokens, so
  it isn't load-bearing for Robinhood.
- The 96-token catalog itself, pending the `uiMultiplier` discussion.
- Windows has no Robinhood support at all — no PR open. Flagging, not fixing.

**Potential concerns:** CRWD and SKHY have no price on any source (no CoinGecko entry, no Chainlink
feed on 4663) and will read $0.00. Say the word if you'd rather hide zero-priced tokens or drop
those two from the catalog.

## receipts

Everything below is a real run against live endpoints. **I could not compile or run the test suite
locally** — `com.trustwallet:wallet-core` 401s without `TRUSTWALLET_PAT`, which isn't available in
my environment — so CI is the compile/test gate for this commit, not a local green build. Calling
that out rather than implying a clean lint.

```bash
$ # chain id — confirms 4663 is what the RPC actually reports
$ curl -s -X POST https://rpc.mainnet.chain.robinhood.com/ \
    -d '{"jsonrpc":"2.0","id":1,"method":"eth_chainId","params":[]}'
{"jsonrpc":"2.0","id":1,"result":"0x1237"}          # 4663

$ # LI.FI serves the chain — justifies wiring the id rather than dropping the provider
$ curl -s https://li.quest/v1/chains | jq '.chains[] | select(.id==4663) | {name, diamondAddress}'
{"name":"Robinhood Chain","diamondAddress":"0xB477751B76CF82d00a686A1232f5fCD772414Af3"}

$ # block cadence — 101 ms/block, vs the 4s fallback it was taking
18790595->18791595:  1000 blocks /  101s = 101 ms/block
18771595->18790595: 19000 blocks / 1913s = 101 ms/block

$ # price coverage through the proxy we ship against, all 100 contract-priced tokens
PRICED   97/100
UNPRICED 3/100: CRWD, SKHY, LINK       # LINK prices by priceProviderID, so 94/96 stock tokens OK

$ # per-token vs per-share, calibrated on AAPL (multiplier 1.0)
Chainlink latestRoundData() on 4663 : $333.22
CoinGecko robinhood platform        : $331.98
LI.FI                               : $332.02
```

Verified separately that all 14 EVM chains have an explicit `blockTimeMs` arm, so the new
exhaustiveness guard passes by construction.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/EvmCoinFinder.kt
```diff
@@ -26,8 +23,8 @@ import timber.log.Timber
  * `/balance/v1.2/{chain}/balances/{address}` (no Alchemy 100-token cap) and metadata from
  * `/token/v1.2/{chain}/custom`. Legitimacy is gated by the CoinGecko-provider allowlist, replacing
  * the empty-logo heuristic that was dropping legit small-caps (see #4555). For EVM chains outside
- * 1inch's surface — Blast / Cronos / Hyperliquid / Mantle / Sei / zkSync — discovery falls back to
- * `balanceOf`-iterating the curated [Coins] catalog.
+ * 1inch's surface — Blast / Cronos / Hyperliquid / Mantle / Robinhood / Sei / zkSync — discovery
```
rkokhatskyi: Robinhood is not outside 1inch's surface: `/1inch/balance/v1.2/4663` returns 200 where Blast, Cronos and Hyperliquid all return 404, and its token list covers 296 contracts. Routing it to the 100 entry curated fallback leaves 197 indexed tokens undiscoverable

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
```diff
@@ -358,6 +362,8 @@ fun Chain.oneInchChainId(): Long =
         Chain.ZkSync -> 324
         Chain.Mantle -> 5000
         Chain.Hyperliquid -> 999
+        // LI.FI-only: 1inch never quotes 4663, but LiFiQuoteSource needs the id resolved.
```
rkokhatskyi: This is false: `api.vultisig.com/1inch/swap/v6.1/4663/quote` returns 200 with a live dstAmount and `/swap` returns executable calldata, while chain 9999 returns 404. Robinhood staying LI.FI and Kyber only is a provider table choice, not a 1inch coverage gap

rkokhatskyi (changes requested):

aminsato (commented):
Chain wiring looks complete — I re-ran the table sweep (every file mentioning `Sei`/`Mantle`/`Blast` and not `Robinhood`) and the only remaining gaps are deliberate ones: `supportsFeeService` (dead set, also missing Sei/Hyperliquid), Blockaid's chain lists (gated off, as the description says), `FeeService`, and `String.getChain()` (THORChain secured-asset tickers only, unreachable here).

Verified live against chain 4663 rather than taking the description's word for it:

- `eth_chainId` → `0x1237` (4663), Multicall3 deployed at the canonical address.
- All **100** catalog contracts: `symbol()` matches every `ticker` and `decimals()` matches every `decimal` — zero mismatches, no duplicate tickers or contracts. I ran this as a single 100-sub-call `aggregate3` built with this PR's own encoder, which also confirms the batched `EvmCoinFinder` path returns in one round trip at full catalog size rather than falling back to 100 concurrent `eth_call`s.
- CoinGecko asset platform `robinhood` exists and prices these contracts (AAPL token → $331.66, USDG → $0.9999), so the empty `priceProviderID` on stock tokens resolves through the contract path.
- KyberSwap answers on the `chain.raw.lowercase()` = `robinhood` slug, and LI.FI quotes 4663 — both provider entries are real.
- wallet-core 4.7.0's registry has `robinhoodchain`, chainId `4663`, `m/44'/60'/0'/0/0`, so `compatibleChainId`/`compatibleDerivationPath` need no special-casing.
- `robinhoodchain.blockscout.com` serves `/tx/`; wallet-core's own explorer URL redirects to it.
- Cross-platform parity with vultisig-ios#4898 is exact: same chain name, ticker, `ROBINHOOD` swap asset, `robinhood` slug, and an identical 101-token catalog.

Three comments inline. Non-blocking beyond those:

- No before/after render of the chain tile or the token list. Repo convention is real screenshots for anything UI-visible, and this adds two drawables plus a 101-token list; the description flags them as to-follow, which is the right call, but they're still owed before merge.
- Robinhood is the second chain (after Sei) to point at a vendor RPC directly instead of an `api.vultisig.com` route, so balance and broadcast traffic reaches Robinhood with the user's IP and address attached. Worth a proxy route rather than a precedent.
- Swaps ship enabled on a chain Blockaid can't scan. Fail-closed is correct and the gate does that, but it's a swap-capable chain with no transaction screening at all.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
```diff
@@ -358,6 +362,8 @@ fun Chain.oneInchChainId(): Long =
...
+        Chain.Robinhood -> 4663
```
aminsato, line 372: This makes `oneInchChainId()` do double duty. Its `else` arm throws `SwapRouteNotAvailable`, so a resolved id reads as "1inch routes this chain" — and two call sites act on that reading:

- `GetChainTokensUseCase.emitEvmTokens` calls `oneInchApi.getTokens(chain)` for every EVM chain. Before this arm it threw while building the URL, so no request left the device. Now every open of the Robinhood token list fires `api.vultisig.com/1inch/swap/v6.1/4663/tokens`, which can only fail, then falls back through `runCatching`. Harmless in outcome, but it's a guaranteed-doomed round trip on a screen open.
- `ContractAbiRepository.evmChainId` documents the inverse invariant in its KDoc: "[Chain.Sei] is the one EVM chain Sourcify indexes that 1inch doesn't route (adding it to [oneInchChainId] would wrongly signal swap support), so it stays the sole documented exception." That is now false, and the KDoc is unchanged.

A separate `Chain.evmChainId` for the LI.FI and Sourcify callers, leaving `oneInchChainId()` as the 1inch-routability signal, keeps both honest and lets Sei's exception fold into it. If that's too much for this PR, the ContractAbiRepository KDoc still needs to stop claiming Sei is the sole exception.

data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt
```diff
@@ -2400,6 +2400,1128 @@ object Coins {
...
+                hexPublicKey = "",
+                priceProviderID = "chainlink",
+                contractAddress = "0x492641F648a4986844848E0beFE66D14817bCE34",
+                isNativeToken = false,
+            )
+        val STOCK_TOKENS = listOf(
+            Coin(
+                chain = Chain.Robinhood,
+                ticker = "AAOI",
+                logo = "https://financialmodelingprep.com/image-stock/AAOI.png",
```
aminsato, line 2491: 91 of the 96 stock icons hotlink `financialmodelingprep.com`; `main` has 5 absolute logo URLs in this entire file. The endpoint is unauthenticated today (I checked — `/image-stock/AAPL.png` and `/TSLA.png` both 200), but nothing pins that: FMP is a commercial data vendor, and if they gate hotlinking or rate-limit the path, the whole stock catalog renders iconless with no fallback. It also means the icon fetch itself tells a third party which tickers a Vultisig user is browsing.

Mirroring these to the Vultisig asset host — the way the chain marks ship as bundled drawables — removes both the availability and the disclosure. The regeneration script in vultisig-sdk is the natural place to emit the mirrored URLs.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
```diff
@@ -275,9 +275,10 @@ class EthereumFeeService @Inject constructor(private val evmApiFactory: EvmApiFa
             maxOf(capped, ETHEREUM_SWAP_PRIORITY_FEE_FLOOR)
         } else {
             when (chain) {
-                // Arb and Mantle requires no miner tip
+                // Arb, Mantle and Robinhood (Arbitrum Orbit) require no miner tip
                 Chain.Arbitrum,
-                Chain.Mantle -> BigInteger.ZERO
+                Chain.Mantle,
+                Chain.Robinhood -> BigInteger.ZERO
```
aminsato, line 281: The other two arms of this branch each have a dedicated test — `Arbitrum priority fee is always zero` and `Mantle priority fee is always zero` in `EthereumFeeServiceTest` — and `EthereumFeeServiceTest` is untouched here. Without one, the ~10x fee overstatement this commit fixes silently returns the next time someone restructures this `when`. It's the same three lines with `transfer(Chain.Robinhood)`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/EvmCoinFinder.kt
realpaaao: You're right that `/balance/v1.2/4663` is live — confirmed, and discovery now uses it (f7ff4afeb). But pure 1inch routing would have been a regression, which is why this ships as a union rather than a rewire:

- The discovery pipeline is two-stage: `/balance` → `/token/v1.2/custom` metadata + CoinGecko-provider gate. I ran all 101 catalog tokens through stage 2: **only 4 resolve; all 96 stock tokens return 404** (e.g. `/1inch/token/v1.2/4663/custom?addresses=0xaf3d…93f9` → `{"message":"Not Found"}` for AAPL). The 1inch path alone discovers a wallet's USDG and drops its AAPL.
- Of the 296-token list: 197 aren't in our catalog, and ~139 of those fail the same CG gate every other chain applies (`TESTIHOOD`, `r0b`, …). The real delta 1inch adds is **58 CG-verified tokens** (PEPE, WIF, DOGE, GME, VIRTUAL, syrupUSDG…) — the union now surfaces them.
- LINK (`0x4926…ce34`) is absent even from 1inch's balance index, so the curated scan also stays load-bearing for it.

Robinhood is now the first `HYBRID_DISCOVERY` chain: both paths run concurrently, curated entries win the dedupe. Tests pin the union, the metadata-404 case, and the 1inch-down case.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
realpaaao: Confirmed — `/quote` returns live pricing and `/swap` returns executable calldata to the deployed router (`0x5a70…89c7`, 24.5KB code on-chain; my first `/swap` probe 403'd only because I used a junk `from` address). The comment is deleted and 1inch is now a real provider in ab99f24e1: Robinhood = `{ONEINCH, LIFI, KYBER}`, pinned by a provider-contract test (exact set, LiFi alone cross-chain). For the record, the original exclusion came from following the Sei/Mantle template without probing the proxy — your probe was the correction.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
realpaaao: Resolved by the provider change rather than a workaround: Robinhood is now genuinely 1inch-routed (ab99f24e1), so a resolved id signals exactly what the `ContractAbiRepository.evmChainId` KDoc invariant says it should, and the `/tokens` call `GetChainTokensUseCase` fires is a real request (296 tokens on 4663). Sei stays the sole documented exception.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/EthereumFeeService.kt
realpaaao: Added in 183802bd0 — `Robinhood priority fee is always zero`, same three lines as the Arbitrum/Mantle arms. Mutation-verified: dropping `Chain.Robinhood` from the zero arm fails it.

data/src/test/kotlin/com/vultisig/wallet/data/usecases/EvmCoinFinderTest.kt
```diff
@@ -242,9 +243,8 @@ internal class EvmCoinFinderTest {
                 !it.isNativeToken && it.contractAddress.isNotEmpty()
             }
         every { evmApiFactory.createEvmApi(Chain.Blast) } returns evmApi
-        coEvery { evmApi.getERC20Balance(ADDRESS, any()) } returns BigInteger.ZERO
-        coEvery { evmApi.getERC20Balance(ADDRESS, curated.contractAddress) } returns
-            BigInteger.valueOf(7)
+        coEvery { evmApi.getBalances(ADDRESS, any()) } returns
+            mapOf(curated.contractAddress to BigInteger.valueOf(7))
```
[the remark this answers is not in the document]
realpaaao: Fixed in f7ff4afeb — the stub now matches the exact curated Blast contract list and verifies exactly one `getBalances` call.

data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt
realpaaao: Checked the obvious alternative first — Robinhood's own CDN (`cdn.robinhood.com/ncw_assets/logos/{address}.png`, the URLs 1inch's token list uses): all 91 return 200, but every one is the **same 4,058-byte placeholder** (identical md5 `7458db13`). The chain's own CDN has no real stock icons; FMP is currently the only source that does.

So FMP stays for now, and mirroring the icons to the Vultisig asset host via the SDK regeneration script — your suggestion — is the follow-up that kills both the availability and the disclosure concern. Tracking it on the SDK side.

realpaaao:
Parity note: the same design shipped on the other two surfaces, so the 1inch decision was made once and applied everywhere —

- **ios#4898**: hybrid discovery + `.oneinch` provider + provider-contract test (also resolves its merge conflicts).
- **sdk#1541**: hybrid `findEvmCoins` + Robinhood in `oneInchSwapEnabledChains` + a **chain-scoped router allowlist entry** — 4663's 1inch router is a chain-specific deployment (`0x5a70…89c7`), the second after zkSync, so the AGG-02 fund-safety allowlist got a scoped arm + the same test trio. A live CLI `swap-quote` on 4663 ranked 1inch first through that gate (receipt on the SDK PR).

Provider sets now match 1:1: Robinhood = {1inch, LiFi, Kyber} on all three (+ gated SwapKit on iOS per its unconditional-arm convention).

data/src/main/kotlin/com/vultisig/wallet/data/usecases/EvmCoinFinder.kt
```diff
@@ -39,14 +40,26 @@ constructor(private val oneInchApi: OneInchApi, private val evmApiFactory: EvmAp
...
+        if (heldContracts.isEmpty()) return emptyList()
+        return fetchMetadataAndFilter(chain, heldContracts)
+    }
 
-        return vultTopUpOnEthereum(chain, address, discovered)
+    private suspend fun findHybrid(chain: Chain, address: String): List<Coin> = coroutineScope {
+        val oneInch = async { findOneInch(chain, address) }
+        val curated = async { findFallback(chain, address) }
+        // Curated first so its hand-verified entries win the dedupe on contracts both paths find.
+        (curated.await() + oneInch.await()).distinctBy { it.contractAddress.lowercase() }
```
rkokhatskyi: Both GME contracts on 4663 clear this dedupe because the key is the contract, not the ticker based `Coin.id`. The 1inch `0x7e8638` entry then REPLACEs the curated `0x1b0E31` row via `VaultDao.kt:58`, so the vault's GME points at a contract the user never chose

data/src/test/kotlin/com/vultisig/wallet/data/usecases/EvmCoinFinderTest.kt
```diff
@@ -55,6 +55,7 @@ internal class EvmCoinFinderTest {
                 Chain.Mantle,
                 Chain.Sei,
                 Chain.Hyperliquid,
+                Chain.Robinhood,
```
rkokhatskyi: This entry now asserts the opposite of the same commit: the test says Robinhood has no 1inch surface and must take the fallback path, while `EvmCoinFinder.kt:189` states 1inch indexes 4663 and `find` routes it through findHybrid

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/usecases/EvmCoinFinder.kt
realpaaao: Fixed in f1532c0cf — the dedupe now keys on `Coin.id` (the Room PK the REPLACE collides on) instead of the contract, curated list first, so the curated GME `0x1b0E31…` is the row that survives and the 1inch `0x7e8638…` entry is dropped before it can reach `insertCoins`. Regression test added and mutation-verified: restoring the contract-keyed dedupe fails it.

One deliberate platform divergence to flag: iOS and the SDK keep both GME rows because their coin ids include the contract (`chain-ticker-contract` / contract-as-id), so there's no overwrite there — dropping the second contract on those platforms would hide funds the address genuinely holds. If we want ticker-collision policy unified cross-platform, that's a product call worth its own issue.

data/src/test/kotlin/com/vultisig/wallet/data/usecases/EvmCoinFinderTest.kt
realpaaao: Fixed in f1532c0cf — Robinhood is dropped from the fallback-path list (that test now covers only the six true fallback chains), and the hybrid test additionally asserts `HYBRID_DISCOVERY_CHAINS` is disjoint from `ONE_INCH_SUPPORTED_CHAINS`, so the routing claim and the code can't drift apart again. Full `:data:testDebugUnitTest` suite green (2,290 tests).

johnnyluo (approved):
LGTM

PR #5392, opened 2026-07-25
fix(home): adopt V3 icons for the home action buttons

aminsato:
## Summary

The five glyphs behind `AssetActionButton` had drifted from the design. The row still rendered the legacy pack — a double-arrow swap, a shopping bag for **Buy**, and the shapes cluster for **Function** — while Figma had moved the whole row to the V3 icon set.

Each drawable is redrawn from the V3 component used in the design frame:

| Button | Was | Now (V3 component) |
|---|---|---|
| Swap | double horizontal arrows | `arrows-rotate-center` |
| Send | diagonal arrow | `arrow-to-corner-top-right` |
| Buy | shopping bag + plus | `plus` |
| Function | plus / circle / square / cross cluster | `grid-plus` |
| Receive | arrow over a baseline | `arrow-down-from-line` |

**Figma:** [Main 4 Actions row](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=48948-111455)

## Notes

- Geometry is taken at the size the design *places* the icons — a 20-unit box with a 2-unit stroke — not the 24-unit component master. The design resizes the instances without scaling the stroke, so authoring against the master would have rendered the row ~17% too thin. Verified by measuring the exported Figma PNG: the stroke is exactly 2 units in a 20-unit box.
- Edited in place under the existing filenames. These five drawables are referenced only by `AssetActionButton`, and the names describe the action rather than the glyph, so there are no Kotlin changes and no new/orphaned resources.
- Colour stays `#F0F4FC`, consistent with the rest of the V3 pack; `UiIcon` tints it at the call site anyway.

## Test plan

- Open Home → the Swap / Send / Buy / Function / Receive row renders the V3 glyphs.
- Confirm the selected (Swap) button still shows the glyph correctly against the blue CTA fill.
- Geometry was checked by rendering the five drawables headlessly and diffing them against the Figma export of the same frame — shape, weight and placement line up on all five.

## Summary by CodeRabbit

* **Style**
  * Refined buy, functions, receive, send, and swap icons with updated dimensions and refreshed vector artwork.
  * Improved icon clarity and consistency through standardized sizing, line weights, and stroke styling.


johnnyluo (approved):
LGTM

PR #5395, opened 2026-07-25
feat(swap): THORChain limit-swap foundation + Place form (#4154)

aminsato:
## Summary

THORChain **limit swap** — Place flow (#4154), behind the remote `limit-swap` feature flag **and** a THORChain-routable pair. The Market swap path is untouched. This covers the full flow: foundation → form UI → **signable placement** → **limit-specific confirmation** → **local record**.

### Transasctoin hash 
https://etherscan.io/tx/0x23b3175c971379b3d23da478db9bcf4a01e587bc4e2ecafedc77104f24e6d467
https://gateway.liquify.com/chain/thorchain_midgard/v2/actions?txid=23B3175C971379B3D23DA478DB9BCF4A01E587BC4E2ECAFEDC77104F24E6D467&limit

## Foundation (data — pure, JUnit-tested against the SDK's published memo vectors)

- **`LimitSwapMemo`** builds the `=<` memo: source amount rescaled to THORChain's **1e8 fixed point** (an 18-decimal source fed raw would encode a LIM `1e10x` off), integer-floored LIM that **throws on 0** (THORChain reads it as an unprotected market order), UTXO **80-byte OP_RETURN cap** dropping the affiliate only when forced. Affiliate stays `va` (Android's market thorname) — the one intentional deviation from the SDK vectors (byte-lengths identical). `assertLimitSwapMemo` fail-closes the signer against a market/corrupted memo or zero LIM; `parse` recovers the fields for the local record.
- **`thorchainMemoAsset`** uses a dedicated prefix map so **Noble routes as `NOBLE`**, not its `USDC` fee ticker.
- **Availability gate** on `EnableAdvSwapQueue` (fail-closed, exact `1`), **routable supported-chains**, **reference-price probe**. 12-vector fixture suite asserts the memo byte-for-byte.

## Form UI (Execute-when card, Figma `74341-118736`)

- `$`/asset display toggle, **Market/+1/+5/+10%** presets, **12h/24h/3d** expiry, live **% from market**, below-/far-above-market warnings. Canonical quantity is **buy-units per sell-unit** (what the LIM consumes) so the toggle never changes the signed value. Additive VM wiring in `SwapFormViewModel`.

## Placement, confirmation & storage (reuses the market verify → keysign pipeline)

- **`BuildLimitSwapTransactionUseCase`** mirrors the SDK's `buildLimitSwapKeysignPayload`: native **RUNE → MsgDeposit**, native **gas asset → Asgard inbound** transfer, **ERC20 → router `depositWithExpiry`** + approve when the allowance is short. Produces a `RegularSwapTransaction` (`SwapPayload.ThorChain` + `=<` memo) that flows through the existing **Blockaid + "The swap amount is correct" consent + keysign** unchanged.
- **Fail-closed on both ends:** `EnableAdvSwapQueue` mimir checked at placement **and** re-checked at sign time (`SwapInboundHaltPreflight`, keyed off the `=<` memo).
- **Confirmation** is now limit-aware: "You're placing a limit order", the destination as the **min. payout** (the memo's LIM), and a **"Target Price: 1 BTC = $…" / expiry** row. Labels are computed once in the use-case and carried on the in-memory `RegularSwapTransaction`, so a co-signer degrades to the plain swap overview.
- **Storage:** a `pending_limit_order` Room table (**migration 36→37**) records every placed order on successful broadcast, keyed off the signed `=<` memo in `KeysignViewModel` — fires only for limit orders, and a storage failure can't break the keysign success path.
- `THORChainInboundAddress` gained the `router` field (the market path reads it off the quote; a locally-built order has none).

## Tests
- `:data` — memo (build / `assertLimitSwapMemo` / `parse`) / amount / asset / availability / supported-chains / market-price / `PendingLimitOrderRepository`, all incl. SDK fixtures.
- `:app` — `LimitOrderPricingTest`, `BuildLimitSwapTransactionUseCaseTest` (mimir fail-closed, RUNE vs gas-asset routing, `=<` memo), `SwapInboundHaltPreflightTest` (+ sign-time mimir), `SwapFormViewModelTest` (121), `KeysignViewModel*` construction tests updated — green.
- `:app:lintDebug` ✅; strings translated to all 10 locales.

The Limit tab previously rendered a static "Limit orders coming soon" placeholder (the production state while the flag is off). Coin logos show as letter placeholders in the isolated previews and load normally in the real screen.

## Summary
* **New Features**
  * Added Market/Limit swap mode with an “Execute when” limit-order form (target presets, expiry options, and a price-unit toggle) plus market-comparison warnings.
  * Limit orders now show target price and expiry details on the verification screen and include additional signing-time safety checks for unsupported advanced limit conditions.
  * Newly placed limit orders are locally tracked to support improved verification.
* **Localization**
  * Added Limit Swap translations across multiple languages.
* **Tests**
  * Added coverage for limit pricing, memo validation, supported-chain detection, market probing, and limit transaction construction.


aminsato:
Thanks @coderabbitai — addressed in `29fabd25b`:

**Correctness / fund-safety**
- 🔴 `BuildLimitSwapTransactionUseCase`: now **fails closed when an EVM ERC20 source has no router** (a plain transfer would drop the memo and strand the tokens), and rejects `globalTradingPaused` / `chainTradingPaused` inbounds at placement too, not just `halted`.
- `SwapFormViewModel`: the limit-swap flag is now a `MutableStateFlow` inside the reactive `combine`, so a late-resolving remote flag re-triggers the market-price fetch (no more "enabled with no price").
- `LimitSwapMemo.parse` now enforces the same grammar as `assertLimitSwapMemo` (trade-target shape, **positive LIM**, affiliate bps) — a malformed memo can't be recorded with a zero target price.
- Confirmation now classifies a tx as a limit order only when the memo **and** both labels are present, keeping the title and Target-Price/expiry row atomic.

**i18n / a11y**
- Limit-form fiat now honors the selected app currency (`ConvertTokenValueToFiat` + `FiatValueToStringMapper`) instead of hardcoding `$`.
- "% from market" moved to a string resource; added `contentDescription`s and a 40dp touch target for the pencil. All new strings propagated across the 10 locales.

**Cancellation / nitpicks**
- Rethrow `CancellationException` in the `KeysignViewModel` pending-order write and `LimitSwapSupportedChains`.
- `requireNotNull` for the memo-asset precondition and the fixture resource; Kotest `shouldBe` in `PendingLimitOrderRepositoryTest`; refreshed the stale `SwapScreen` CTA comment.

**New tests**: ERC20-no-router fail-closed, `parse` grammar rejection, and the sign-time advanced-queue gate (enabled/disabled).

`:app:lintDebug` ✅ and all affected unit suites green.

aminsato:
Addressed the remaining review threads + nitpicks in `e7faf0fd5`:

- **Guard `placeLimitOrder()` re-entry** — added an early `if (isLoadingNextScreen) return`, so a fast double-tap can't start a duplicate on-chain deposit.
- **Gate the Limit CTA on form readiness** — the "Place Limit Order" button is now disabled until there's a resolved target price and a positive sell amount (`limitOrder.isPlaceOrderEnabled`), instead of relying on the placement to fail later.
- **`safeLaunch` for the network call** — `fetchMarketPrice` now uses `safeLaunch` per the repo convention instead of `launch` + manual `runCatching`.
- **Nitpick** — dropped the repeated `!!` on exception messages in the limit-swap tests (`error?.message.orEmpty().contains(...)`).

The two still-open threads on **`LimitSwapForm` a11y** (contentDescription + 40dp touch target) and **atomic limit-order classification** (`isLimitOrder` only true when both labels are present) were already fixed in `29fabd25b` — those threads are anchored to the pre-fix line numbers. Earlier outside-diff/nitpicks (stale CTA comment, `requireNotNull`, Kotest, the sign-time-gate test) are also in as of `29fabd25b`.

`:app:lintDebug` ✅ and all affected unit suites green.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/LimitSwapForm.kt
```diff
@@ -0,0 +1,377 @@
...
+        contentAlignment = Alignment.Center,
+    ) {
+        Icon(
+            painter = painterResource(icon),
+            contentDescription = null,
+            tint = if (selected) colors.text.primary else colors.text.secondary,
+            modifier = Modifier.size(16.dp),
+        )
+    }
+}
```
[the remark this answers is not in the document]
aminsato: Fixed in `29fabd25b`: both `UnitToggleButton` and the pencil now pass a `contentDescription` (`limit_swap_toggle_price_unit` / `limit_swap_edit_assets`), and the pencil is wrapped in a 40dp clickable `Box` (16dp icon centered) for the touch target. This thread is anchored to the pre-fix line numbers.

app/src/main/java/com/vultisig/wallet/ui/models/swap/VerifySwapViewModel.kt
```diff
@@ -108,6 +108,11 @@ internal data class SwapTransactionUiModel(
     // rebuilds the tx from the payload without a quote — the row is then hidden (#5335).
     val priceImpactPercent: String? = null,
     val priceImpactLevel: PriceImpactLevel? = null,
+    // THORChain limit order (#4154): drives the "You're placing a limit order" title and the Target
+    // Price / expiry row on the confirmation screen. False (and null labels) for market swaps.
+    val isLimitOrder: Boolean = false,
+    val limitTargetPriceLabel: String? = null,
+    val limitExpiryLabel: String? = null,
```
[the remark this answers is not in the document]
aminsato: Fixed in `29fabd25b`: the mapper now sets `isLimitOrder = limitTargetPriceLabel != null && limitExpiryLabel != null`, so the UI-model classification is only true when both confirmation labels are present — the incomplete state can't reach the signing screen. Anchored to pre-fix lines.

aminsato:
@coderabbitai full review

aminsato:
Addressed review 4781181903 in `126b18d85`:

**Correctness / safety**
- `ThorMimirRepository.isAdvancedSwapQueueEnabled` now rethrows `CancellationException` (was swallowed by `runCatching`), keeping the fail-closed default for real failures.
- `PendingLimitOrderRepository` no longer persists an empty `source_asset` on a mapping failure — it propagates to the caller's best-effort log instead of writing an unusable record.
- `getLimitSwapSupportedChains` returns **only THORChain** on a successful all-halted response (paused chains aren't re-offered); the static-routable fallback moved to the repository and now applies **only on a failed fetch**.
- **Limit-mode errors** now render an inline surface in `SwapScreen` — the Market error box is absolutely positioned off the market form, so limit failures were silent.

**Quality / a11y**
- `VerifySwap` target-price row: `maxLines`/`ellipsis`/`weight` so a long price can't crowd out the expiry chip.
- `requireNotNull`/`error()` for preconditions (dest-address validator, `intervalBlocks`).
- New test: placement rejects a trading-paused inbound.

**Deferred, with reasons:**
- *Extract the limit concern into its own state holder* — a large refactor of a ~1,100-line, quote-pipeline-owning ViewModel; deferred as you flagged ("fine to defer past this draft").
- *Pencil edits both assets* — Phase 1 opens the sell-asset picker; a dedicated pair-edit (buy asset from the limit form) is a follow-up.
- *Swap the BTC/LTC segwit regex for full bech32 checksum* — the pure decoder in this file is bech32 (v0) only, so this would **reject bech32m/taproot** (`bc1p`/`ltc1p`) addresses. The validator is defensive over the vault's own already-derived address, and the shape regex matches the SDK's own validators. Left as-is to avoid the regression.

`:app:lintDebug` ✅ and the affected unit suites green.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -194,6 +231,323 @@ constructor(
...
+                    val dstCoin = dst?.account?.token
+                    if (
+                        mode == SwapMode.Limit &&
+                            flagEnabled &&
+                            srcCoin != null &&
+                            dstCoin != null &&
+                            isThorchainRoutable(srcCoin.chain) &&
+                            isThorchainRoutable(dstCoin.chain)
+                    ) {
+                        fetchMarketPrice(srcCoin, dstCoin)
```
rkokhatskyi, line 285: `observeLimitForm`'s pair-change branch calls `fetchMarketPrice` without clearing `limitTargetPrice`, so the old pair's price stays live and placeable until the fetch resolves, letting `placeLimitOrder` sign a stale, wrong-pair rate into the `=<` memo's LIM.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
```diff
@@ -194,6 +231,323 @@ constructor(
...
+                        vaultId = inputs.vaultId,
+                        srcToken = inputs.srcToken,
+                        dstToken = inputs.dstToken,
+                        srcAddress = inputs.srcAddress,
+                        srcTokenValue = inputs.srcTokenValue,
+                        // Fail-safe: the memo only accepts 8 fractional digits, and rejects
+                        // anything longer with an exception rather than rounding it itself.
+                        targetPrice = LimitOrderPricing.toMemoScale(targetPrice),
+                        expiryHours = limitExpiry.value.hours,
+                        destinationAddress = externalRecipient.value ?: inputs.dstToken.address,
```
rkokhatskyi, line 451: `placeLimitOrder` reads `externalRecipient.value` straight into `destinationAddress` without calling `externalRecipientError()`, the checksum gate `swap()` enforces via `chain.coinType.validate`, so a stale or unchecksummed external recipient can be signed into the `=<` memo's dest_addr for a limit order.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapInboundHaltPreflight.kt
```diff
@@ -49,7 +57,23 @@ constructor(private val thorChainApi: ThorChainApi, private val mayaChainApi: Ma
...
+    /**
+     * Re-checks the `EnableAdvSwapQueue` mimir at sign time for a THORChain limit order (memo
+     * starts with `=<`). The mimir can flip while the user sits on the confirmation screen, and a
+     * `=<` order placed while the queue is disabled can execute as an unprotected market swap — so
+     * the gate is re-run here, fail-closed, just before signing.
+     */
+    private suspend fun assertAdvancedSwapQueueEnabledForLimitOrder(transaction: SwapTransaction) {
+        val memo = transaction.memo ?: return
+        if (!memo.startsWith(LimitSwapMemo.PREFIX)) return
+        if (!thorMimirRepository.isAdvancedSwapQueueEnabled()) {
```
rkokhatskyi: `isAdvancedSwapQueueEnabled` reads `ThorMimirRepository`'s shared 30s-TTL singleton cache, so this sign-time recheck returns the same value the placement-time check already read whenever Sign happens within 30s of Place, silently defeating the fail-closed recheck the code's own comments describe.

app/src/main/java/com/vultisig/wallet/ui/models/swap/BuildLimitSwapTransactionUseCase.kt
```diff
@@ -0,0 +1,237 @@
...
+                    )
+                ),
+        )
+    }
+
+    /** "1 <buy> = $65,800.13" (fiat when the sell price is known, else in sell-asset units). */
+    private fun targetPriceLabel(srcToken: Coin, dstToken: Coin, targetPrice: BigDecimal): String {
+        val fiat = LimitOrderPricing.fiatPricePerBuyUnit(targetPrice, srcToken.usdPrice)
+        return if (fiat != null) {
+            "1 ${dstToken.ticker} = $${fiatFormat.format(fiat)}"
```
rkokhatskyi: `targetPriceLabel` hardcodes `$` and formats raw `srcToken.usdPrice` directly, ignoring the app currency, the same bug `SwapFormViewModel.formatFiat` was fixed for elsewhere in this PR, so a non-USD user sees a mislabeled USD price on the limit-order sign-confirmation screen.

app/src/main/java/com/vultisig/wallet/ui/models/swap/BuildLimitSwapTransactionUseCase.kt
```diff
@@ -0,0 +1,237 @@
...
+                    BigDecimal.ZERO
+                }
+            "1 ${dstToken.ticker} = ${assetFormat.format(sellPerBuy)} ${srcToken.ticker}"
+        }
+    }
+
+    private fun expiryLabel(expiryHours: Int): String =
+        when (expiryHours) {
+            72 -> "3d"
+            else -> "${expiryHours}h"
```
rkokhatskyi: `expiryLabel` returns a raw `"${hours}h"`/`"3d"` literal instead of the `R.string.limit_swap_expiry_*` resources the form's own pills use for the identical value, so the confirmation screen's expiry text can silently diverge from the localized form if a locale ever translates these differently.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Confirmed and fixed in b9e535220.

The window is actually wider than the refetch: `fetchMarketPrice` only reseeds a *null* target (`preset != null || limitTargetPrice.value == null`), so a custom preset-less price survived the pair change indefinitely, not just until the fetch resolved.

The collector now drops `marketTargetPrice`/`limitTargetPrice` and cancels the in-flight job whenever the pair key changes, so `isPlaceOrderEnabled` goes false until the new pair prices. Regression test: `SwapFormViewModelTest.changing the pair drops the previous pair's target price until the new one resolves`.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapFormViewModel.kt
aminsato: Confirmed and fixed in b9e535220 — `placeLimitOrder` now runs the same `externalRecipientError()` gate as `swap()` before collecting inputs.

Worth noting the memo builder was not a sufficient backstop: `assertValidLimitSwapDestinationAddress` is shape-only for EVM (`^0x[0-9a-fA-F]{40}$`) and legacy base58, so an unchecksummed mixed-case EVM address passed it. And since the typed recipient is never cleared when the destination chain changes, a leftover address from the Market tab reached the builder unvalidated.

Regression test: `SwapFormViewModelTest.placeLimitOrder is refused when the external recipient is invalid`.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapInboundHaltPreflight.kt
aminsato: Confirmed and fixed in b9e535220. `ThorMimirRepositoryImpl` is bound `@Singleton` with a 30s TTL, so any sign inside that window returned the value the placement-time check had already read.

`isAdvancedSwapQueueEnabled` now takes `forceRefresh`, which bypasses and re-seeds the cache, and the sign-time gate passes `true` (the LP validators keep the cached read they were added for).

Two tests cover it: `ThorMimirRepositoryTest.forceRefresh re-reads the queue flag inside the TTL`, and the preflight tests now stub `isAdvancedSwapQueueEnabled(forceRefresh = true)` so a cached read fails them.

app/src/main/java/com/vultisig/wallet/ui/models/swap/BuildLimitSwapTransactionUseCase.kt
aminsato: Confirmed and fixed in b9e535220.

Rather than currency-fixing the label at build time, the transaction now carries the raw value (`limitOrderTargetPrice: BigDecimal?`) and `SwapTransactionToUiModelMapper` formats it — it already holds the currency, `ConvertTokenValueToFiatUseCase` and `FiatValueToStringMapper`. So the confirmation screen prices one buy unit through the same path as the form (`convertTokenValueToFiat` on one whole sell unit → `fiatPricePerBuyUnit` → `fiatValueToStringMapper(asPrice = true)`), and it also picks up a currency change made after the order was staged. The sell-asset-units fallback is unchanged.

app/src/main/java/com/vultisig/wallet/ui/models/swap/BuildLimitSwapTransactionUseCase.kt
aminsato: Confirmed and fixed in b9e535220. The transaction now carries `limitOrderExpiryHours: Int?` and the mapper resolves it through `LimitExpiryOption.labelRes` — the exact resources the form's pills render — so the two can no longer diverge per locale.

That made `SwapTransactionUiModel.limitExpiryLabel` a `UiText`, resolved with `asString()` in `VerifySwapScreen`.

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/THORChainInboundAddress.kt
```diff
@@ -16,4 +16,8 @@ data class THORChainInboundAddress(
     @SerialName("chain_lp_actions_paused") val chainLPActionsPaused: Boolean = false,
     @SerialName("gas_rate") val gasRate: String = "",
     @SerialName("gas_rate_units") val gasRateUnits: String = "",
+    // The chain's router contract (EVM chains only), used to build ERC20 limit-order deposits. The
+    // market swap path reads the router off the swap quote instead; a locally-built limit order has
+    // no quote, so it comes from here. Absent for non-EVM inbounds.
+    @SerialName("router") val router: String? = null,
```
johnnyluo: This approach is wrong , limit swaps also use THORChain normal swap endpoint , so when the inbound asset is ERC20 token , then the router will be set , thus you should not rely on this field to get it

data/src/main/kotlin/com/vultisig/wallet/data/api/models/quotes/LimitSwapReferenceQuoteJson.kt
```diff
@@ -0,0 +1,15 @@
...
+import kotlinx.serialization.Serializable
+
+/**
+ * Minimal view of a THORChain `/quote/swap` response for the limit-swap reference price. Only
+ * `expected_amount_out` (1e8 fixed point) is read; every other quote field is irrelevant to seeding
+ * the form's price and is ignored.
+ */
+@Serializable
+data class LimitSwapReferenceQuoteJson(
+    @SerialName("expected_amount_out") val expectedAmountOut: String? = null,
```
johnnyluo: the full response is like 
```
RESPONSE: 200 OK
                                                                                                    METHOD: HttpMethod(value=GET)
                                                                                                    FROM: https://gateway.liquify.com/chain/thorchain_api/thorchain/quote/swap?from_asset=ETH.USDC-0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48&to_asset=BTC.BTC&amount=1000000000&destination=bc1qgvrzaskwqauth0653x0dnw4p58cu89snfcys32&streaming_interval=1&streaming_quantity=1&affiliate=va&affiliate_bps=50
                                                                                                    COMMON HEADERS
                                                                                                    -> access-control-allow-headers: Content-Type, Authorization, solana-client, x-client-id, grpc-metadata-x-cosmos-block-height, x-server-time, Origin, X-Requested-With, Accept
                                                                                                    -> access-control-allow-methods: GET, POST, PUT, DELETE, OPTIONS
                                                                                                    -> access-control-allow-origin: *
                                                                                                    -> access-control-expose-headers: grpc-metadata-x-cosmos-block-height, x-server-time
                                                                                                    -> content-type: application/json
                                                                                                    -> date: Tue, 28 Jul 2026 11:44:47 GMT
                                                                                                    -> grpc-metadata-x-cosmos-block-height: 27193724
                                                                                                    -> server: fasthttp
                                                                                                    -> vary: Accept-Encoding
                                                                                                    -> x-server-time: 1785239088
                                                                                                    BODY Content-Type: application/json
                                                                                                    BODY START
                                                                                                    {
                                                                                                      "inbound_address": "0xe3536ba9559966c357f551ceccccf38b533aa171",
                                                                                                      "inbound_confirmation_blocks": 2,
                                                                                                      "inbound_confirmation_seconds": 24,
                                                                                                      "outbound_delay_blocks": 0,
                                                                                                      "outbound_delay_seconds": 0,
                                                                                                      "fees": {
                                                                                                        "asset": "BTC.BTC",
                                                                                                        "affiliate": "79",
                                                                                                        "outbound": "1053",
                                                                                                        "liquidity": "31",
                                                                                                        "total": "1163",
                                                                                                        "slippage_bps": 19,
                                                                                                        "total_bps": 688
                                                                                                      },
                                                                                                      "router": "0xD37BbE5744D730a1d98d8DC97c42F0Ca46aD7146",
                                                                                                      "expiry": 1785239978,
                                                                                                      "warning": "Do not cache this response. Do not send funds after the expiry.",
                                                                                                      "notes": "Base Asset: Send the inbound_address the asset with the memo encoded in hex in the data field. Tokens: First approve router to spend tokens from user: asset.approve(router, amount). Then call router.depositWithExpiry(inbound_address, asset, amount, memo, expiry). Asset is the token contract address. Amount should be in native asset decimals (eg 1e18 for most tokens). Do not swap to smart contract addresses.",
                                                                                                      "dust_threshold": "1000",
                                                                                                      "recommended_min_amount_in": "392571090",
                                                                                                      "recommended_gas_rate": "15",
                                                                                                      "gas_rate_units": "gwei",
                                                                                                      "memo": "=:b:bc1qgvrzaskwqauth0653x0dnw4p58cu89snfcys32:0/1/1:va:50",
                                                                                                      "expected_amount_out": "14604",
                                                                                                      "max_streaming_quantity": 1,
                                                                                                      "streaming_swap_blocks": 0,
                                                                                                      "total_swap_seconds": 24
                                                                                                    }
```

It is using the same swap quote endpoint , so you shouldn't actually need it

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
```diff
@@ -212,17 +231,51 @@ internal fun SwapScreen(
...
 
                     if (selectedMode == SwapMode.Limit) {
-                        Text(
-                            text = stringResource(R.string.swap_limit_coming_soon),
-                            style = Theme.brockmann.body.m.medium,
-                            color = Theme.v2.colors.text.tertiary,
-                            modifier = Modifier.padding(vertical = 48.dp),
-                        )
+                        val limitOrder = state.limitOrder
+                        if (limitOrder != null) {
```
johnnyluo, line 241: limit swaps only support on THORChain , so if the swap provider is not THORChain , user should not get into the limit swap tab

johnnyluo:
Also , please attach on chain tx link

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/THORChainInboundAddress.kt
aminsato: Done in dd7406563 — the `router` field is off `THORChainInboundAddress` again, and the ERC20 router now comes from `/thorchain/quote/swap` via the existing `getSwapQuotes`, exactly as the market path resolves it.

The inbound-addresses call is still made, but only for what the quote does not carry: the `halted` / `global_trading_paused` / `chain_trading_paused` flags the placement-time gate fails closed on. If you would rather that gate keyed off the quote erroring instead, say so and I will drop the second call entirely.

data/src/main/kotlin/com/vultisig/wallet/data/api/models/quotes/LimitSwapReferenceQuoteJson.kt
aminsato: Agreed — removed in dd7406563. `LimitSwapReferenceQuoteJson` and `getLimitSwapReferenceExpectedOut` are gone; `LimitSwapMarketPriceRepository` now calls `getSwapQuotes` and reads `expected_amount_out` off the same `THORChainSwapQuote` the market path decodes.

Two behaviour changes fall out of that, and I think both are improvements:

- The probe used to deliberately omit the affiliate to get a "clean" rate. Going through `getSwapQuotes` means it now quotes with `va:50` — the same affiliate the `=<` memo itself carries — so the form no longer prices the order against a rate it can never actually fill at.
- Assets are identified with `swapAssetName()` (full contract, `ETH.USDC-0xA0b8…`) like every other quote, instead of the memo-side abbreviated form.

app/src/main/java/com/vultisig/wallet/ui/screens/swap/SwapScreen.kt
aminsato: Correct, and it was worse than an ungated tab: `isLimitTabEnabled` was computed but never actually read by the screen, so the Limit tab was tappable for any pair at all.

Fixed in dd7406563. The flag is now wired into `SwapModeTabs` (`VsTab` already takes `isEnabled`, so the tab renders disabled rather than disappearing and shifting the layout), and the condition behind it is THORChain actually being an eligible provider for the pair — `SwapProvider.THORCHAIN in swapQuoteRepository.getEligibleProviders(src, dst)`. That is the live, pool-aware, per-token check rather than the chain-level `isThorchainRoutable` it used before, which would have offered the tab for e.g. an ETH token with no THORChain pool.

The pair-eligibility version is now part of the limit form trigger too, so a pair picked before the live pools land is re-evaluated once they do — and a pair that loses THORChain routing while the Limit form is open drops the user back to Market instead of stranding them on a hidden tab. Two regression tests cover both.

aminsato:
> Also , please attach on chain tx link

https://etherscan.io/tx/0x23b3175c971379b3d23da478db9bcf4a01e587bc4e2ecafedc77104f24e6d467
https://gateway.liquify.com/chain/thorchain_midgard/v2/actions?txid=23B3175C971379B3D23DA478DB9BCF4A01E587BC4E2ECAFEDC77104F24E6D467&limit

johnnyluo (approved):
LGTM

PR #5396, opened 2026-07-26
refactor(defi): share tab models and fiat conversion across Thorchain and Maya

aminsato:
## Summary

Removes the duplication between the Thorchain and Maya DeFi position ViewModels, and puts a test net under both so the remaining deduplication can be done safely.

These two files were the largest untested ViewModels in the DeFi area (1343 and 811 lines) and are structural twins — roughly 17 of Maya's methods have an identically-named Thorchain counterpart, so a fix in one has to be repeated in the other.

**No user-visible change.** This is a refactor plus tests; the screens render exactly as before.

## What changed

**`DefiTabUiModels.kt` (new)** — the six shared tab models (`BondedTabUiModel`, `StakingTabUiModel`, `LpTabUiModel`, `LpPositionUiModel`, `StakePositionUiModel`, `BondedNodeUiModel`) moved out of `ThorchainDefiPositionsViewModel.kt`, where they only *looked* Thorchain-specific. Same package, so no import changes anywhere.

**`DefiFiatValueCalculator.kt` (new)** — `createFiatValue` was identical in both ViewModels, down to the `CancellationException` handling. It now exists once as an injected collaborator; `createFiatValueFromPoolAsset` moved alongside it.

**`@IoDispatcher` seam** — both ViewModels drop hardcoded `Dispatchers.IO` for the qualifier ten other ViewModels already use. Same dispatcher in production; it makes the load paths deterministic under test.

Net: **61 insertions, 178 deletions** in the two ViewModels.

| File | Before | After |
|---|---|---|
| `ThorchainDefiPositionsViewModel.kt` | 1343 | 1257 |
| `MayachainDefiPositionsViewModel.kt` | 811 | 778 |

## Deliberately not unified

`calculateStakingFiatPrice` looks like a twin but **Thorchain catches only `java.io.IOException` while Maya catches `Exception`** — merging them would silently change Thorchain's failure behavior. `calculateBondedFiatPrice` likewise differs by coin and decimal scale (RUNE/8 vs CACAO/10). Both belong in their own change, made deliberately.

## Tests

49 characterization tests. They pin the subtle behavior, not the obvious paths:

- Maya's saved-selection fallback, including the dotted LP key (`BTC.BTC`)
- Thorchain applying its stored selection **verbatim with no fallback** — an asymmetry with Maya that any future shared base must preserve
- RUNE-by-ticker resolution in all three places the vault is searched (the fixture lists RUJI first on purpose)
- The `updateExistingPosition` withdraw invariant
- Unbond resolving a node's raw bonded amount from the live list
- The `onNavigateToFunctions` token mapping (redeem burns the yToken, mint spends the underlying)
- LP pool-share math and the loading guards on both LP tabs

### Why Thorchain had no tests

Not neglect. Constructing its UI model reaches `wallet.core.jni.CoinTypeConfiguration`, which has no host-JVM binary:

```
java.lang.UnsatisfiedLinkError: 'java.lang.String
wallet.core.jni.CoinTypeConfiguration.getSymbol(wallet.core.jni.CoinType)'
```

The codebase already knew — `ChainValidationService.kt:37` hardcodes a constant to dodge the same call. Only the two JNI-backed extension properties are stubbed, pinned to THORChain's real values (`RUNE`, 8 decimals). `toValue`, `formatAmount` and `formatRuneReward` all run their genuine arithmetic.

### Mutation-checked

Every assertion was verified to actually constrain something: five defects seeded one at a time into the Thorchain ViewModel, three into Maya. Each was killed by the test naming that behavior.

One test failed that bar on the first pass — it asserted a constructor default rather than the invariant its name claimed, because `createTCYStakePosition` already hardcodes `canWithdraw = false`. It was split into two tests that call `updateExistingPosition` directly.

## Test plan

- [x] `./gradlew :app:testDebugUnitTest` — **1663 tests, 0 failures** (up from 1638)
- [x] Mutation testing on all 8 seeded defects, run individually
- [x] `./gradlew :app:ktfmtCheckMain :app:ktfmtCheckTest` — clean
- [ ] Manual smoke: THORChain and MayaChain → **DeFi** toggle → Bonded / Staked / LP tabs, Manage Positions, and the bond/unbond/stake routes. Worth doing on a vault with real bonded nodes and an LP position — the fiat figures on both screens are the thing to watch, since they now flow through one shared calculator.

## Summary by CodeRabbit

* **New Features**
  * Added shared fiat-value conversion for DeFi token amounts and pool-asset legs.
  * Introduced reusable UI models for bonded, staking, and liquidity-pool tabs.
* **Bug Fixes**
  * Improved pricing fallback behavior and error handling (including safe zero-amount short-circuiting).
  * Enhanced DeFi loading/formatting and LP display fallback when pool stats are missing.
* **Tests**
  * Added and expanded characterization tests for DeFi position view models and fiat conversion behavior.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
@@ -1,0 +1,73 @@
+package com.vultisig.wallet.ui.models.defi
+
+import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.models.FiatValue
+import com.vultisig.wallet.data.models.settings.AppCurrency
+import com.vultisig.wallet.data.repositories.TokenPriceRepository
+import java.math.BigDecimal
+import java.math.RoundingMode
+import javax.inject.Inject
+import kotlin.coroutines.cancellation.CancellationException
+import timber.log.Timber
+
+/**
+ * Prices DeFi position amounts in the user's currency.
+ *
+ * The Thorchain and Maya position screens each carried their own identical copy of this conversion;
+ * it lives here so both share one implementation. A price lookup that fails yields a zero value
+ * rather than propagating, so a position whose price is momentarily unavailable still renders its
+ * underlying token amount instead of collapsing the whole tab.
+ */
+internal class DefiFiatValueCalculator
+@Inject
+constructor(private val tokenPriceRepository: TokenPriceRepository) {
+
+    suspend fun createFiatValue(amount: BigDecimal, coin: Coin, currency: AppCurrency): FiatValue =
+        convert(amount, currency) {
+            tokenPriceRepository.getCachedPrice(tokenId = coin.id, appCurrency = currency)
+                ?: tokenPriceRepository.getPriceByContactAddress(
+                    coin.chain.id,
+                    coin.contractAddress,
+                )
+        }
+
+    /**
+     * Prices a pool asset the vault does not hold as a [Coin], so there is no coin id to look up —
+     * the cache key is rebuilt from the pool's ticker and chain instead.
+     */
+    suspend fun createFiatValueFromPoolAsset(
+        amount: BigDecimal,
+        chain: Chain,
+        ticker: String,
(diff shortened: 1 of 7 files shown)


johnnyluo (approved):
LGTM

PR #5409, opened 2026-07-26
fix(evm): point Ethereum POL at the real POL contract, not legacy MATIC's

### LINKED ISSUE #5404: [Fix] Ethereum's curated "POL" token points to the deprecated legacy MATIC contract, not the real Polygon Ecosystem Token
# [Fix] Ethereum's curated "POL" token points to the deprecated legacy MATIC contract, not the real Polygon Ecosystem Token

## Problem
`Coins.kt` defines both `Ethereum.MATIC` and `Ethereum.POL` with the identical `contractAddress` (`0x7d1afa7b718fb893db30a3abc0cfc608aacfebb0` — the legacy MATIC contract, confirmed live via `eth_call`). The real, official POL contract on Ethereum is a separate deployed contract (`0x455e53CBB86018Ac2B8092FdCd39d8444aFFC3F6`). A prior fix (PR #2811) only corrected the price-provider ID/DB ticker/logo for POL, never the contract address, and a Room migration renamed existing users' tracked MATIC row to ticker="POL" in place without changing the contract address — so pre-migration users are now mislabeled "POL" while still pointing at the legacy contract.

## Expected Behavior
The "POL" entry should point at the real POL contract; the legacy MATIC entry should remain unchanged.

## Steps to Reproduce
1. Enable "POL" on Ethereum from Select Tokens.
2. A user holding real POL (e.g. from a CEX withdrawal) sees a zero balance, since the app queries the legacy contract.
3. A user sending legacy MATIC via the "POL" entry may have the transfer arrive at a destination (CEX/bridge) that only recognizes the real POL contract.

## Solution
Update `Coins.kt`'s `Ethereum.POL` entry to `contractAddress = "0x455e53CBB86018Ac2B8092FdCd39d8444aFFC3F6"`, keep the legacy MATIC entry unchanged, and add a Room migration correcting the `contractAddress` column for existing rows where `chain='Ethereum' AND ticker='POL' AND contractAddress=<legacy address>`.

### Must Do
- [ ] Fix `Coins.kt`'s `Ethereum.POL` contract address
- [ ] Add a Room migration correcting existing mislabeled rows

### Must NOT Do
- Don't change the legacy MATIC entry
- Don't touch any other curated token's contract address

## Acceptance Criteria
- [ ] "POL" on Ethereum resolves balance/transfers against the real POL contract
- [ ] Migration test confirms existing mislabeled rows are corrected

rkokhatskyi:
## Summary

`Coins.kt` defined `Ethereum.MATIC` and `Ethereum.POL` with the identical `contractAddress` (the legacy MATIC contract `0x7d1afa7b718fb893db30a3abc0cfc608aacfebb0`). The real POL contract on Ethereum is a separate deployment (`0x455e53CBB86018Ac2B8092FdCd39d8444aFFC3F6`); sharing the address meant enabling POL resolved balances and built transfers against MATIC's contract instead.

Coins are persisted per-vault and reconstructed straight from the stored columns, so the registry fix alone only reaches newly enabled coins. Added `MIGRATION_36_37` to correct the `contractAddress` for already-persisted `chain='Ethereum', ticker='POL'` rows still pointing at the legacy address. The `tokenValue` balance cache is keyed by contractAddress too, so it simply misses once for previously-cached POL balances and self-heals on the next live read (same pattern already established by `MIGRATION_34_35`).

`Ethereum.MATIC` and the unrelated native `Polygon.POL` (different chain, different object block) are untouched.

Closes #5404

## Tests

- `data/src/test/kotlin/.../CoinsPolContractAddressTest.kt` — pins `Ethereum.POL`'s contract address to the real value, asserts it's distinct from `Ethereum.MATIC`'s, and pins MATIC's address stays the legacy value.
- `app/src/test/java/.../Migration36To37Test.kt` — captures the exact SQL `MIGRATION_36_37` issues and asserts it targets `Coins.Ethereum.POL.contractAddress` (not a duplicated literal, so the migration and the registry can't silently drift apart) with the legacy-address filter matching `Coins.Ethereum.MATIC.contractAddress`.

## Test plan

- [x] `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest` — 2252 + 1615 tests, 0 failures
- [x] `./gradlew assembleDebug` — succeeds
- [x] `./gradlew :app:lintDebug :data:lintDebug` — clean, no new issues
- [x] `./gradlew :app:ktfmtCheckMain :app:ktfmtCheckTest :data:ktfmtCheckMain :data:ktfmtCheckTest` — clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the Ethereum POL token to use its correct contract address.
  - Existing wallets are migrated automatically so stored POL entries reflect the corrected address.
  - Preserved the legacy MATIC contract address for MATIC tokens.

- **Tests**
  - Added coverage to verify the corrected POL address and database migration behavior.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/db/AppDatabase.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/db/AppDatabase.kt
@@ -61,7 +61,7 @@ import com.vultisig.wallet.data.db.models.VaultOrderEntity
             VaultNotificationSettingsEntity::class,
             TransactionHistoryEntity::class,
         ],
-    version = 36,
+    version = 37,
     exportSchema = false,
 )
 @TypeConverters(
--- a/app/src/main/java/com/vultisig/wallet/data/db/DatabaseModule.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/db/DatabaseModule.kt
@@ -43,6 +43,7 @@ import com.vultisig.wallet.data.db.migrations.MIGRATION_32_33
 import com.vultisig.wallet.data.db.migrations.MIGRATION_33_34
 import com.vultisig.wallet.data.db.migrations.MIGRATION_34_35
 import com.vultisig.wallet.data.db.migrations.MIGRATION_35_36
+import com.vultisig.wallet.data.db.migrations.MIGRATION_36_37
 import com.vultisig.wallet.data.db.migrations.MIGRATION_3_4
 import com.vultisig.wallet.data.db.migrations.MIGRATION_4_5
 import com.vultisig.wallet.data.db.migrations.MIGRATION_5_6
@@ -108,6 +109,7 @@ internal interface DatabaseModule {
                     MIGRATION_33_34,
                     MIGRATION_34_35,
                     MIGRATION_35_36,
+                    MIGRATION_36_37,
                 )
                 .build()
 
--- a/app/src/main/java/com/vultisig/wallet/data/db/migrations/MIGRATIONS.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/db/migrations/MIGRATIONS.kt
@@ -880,3 +880,25 @@ internal val MIGRATION_35_36 =
             )
         }
     }
+
+// Corrects the Ethereum POL contract address from the legacy MATIC contract it was mistakenly
+// sharing (#5404). The registry fix in Coins.kt only reaches newly enabled coins; coins are
+// persisted per-vault and reconstructed straight from the stored `contractAddress` column, so
+// vaults that already enabled POL keep resolving balances/transfers against the legacy MATIC
+// contract until the stored row itself is corrected. The cached balance in `tokenValue` is keyed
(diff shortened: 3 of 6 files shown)


rkokhatskyi:
Re the CodeRabbit nitpick on `Migration36To37Test.kt` (executing against real SQLite instead of a mock): fair point in principle, but this repo has no infrastructure for that yet — no Robolectric, no JVM SQLite driver dependency, `exportSchema = false`, and zero `MigrationTestHelper` precedent anywhere (checked repo-wide), so building that out for the first time is a genuinely heavy lift for a one-column data-correction migration, which the review itself tags as such. The current test pins the exact SQL text against the live `Coins.Ethereum.POL`/`MATIC` values so it can't silently drift from the registry, and the `WHERE chain='Ethereum' AND ticker='POL' AND contractAddress=<legacy>` clause makes it structurally impossible to touch MATIC or the native Polygon POL row, so I'm leaving it as is.

johnnyluo (approved):
LGTM

PR #5411, opened 2026-07-26
fix(send): sum base + priority fee into the signed EVM gas price

### LINKED ISSUE #5397: [Fix] Advanced Gas Settings save path never sums base fee + priority fee — zeros BSC's gasPrice, drops priority fee on every other EVM chain
# [Fix] Advanced Gas Settings save path never sums base fee + priority fee — zeros BSC's gasPrice, drops priority fee on every other EVM chain

## Problem
`DefaultSendStrategy.applyGasSettings()` sets `maxFeePerGasWei = gs.baseFee` alone, never `gs.baseFee + gs.priorityFee`. Because BSC's on-chain base fee is hardcoded to 0 (BEP-226), any BSC send routed through Advanced Gas Settings signs with `gasPrice = 0` — rejected by BSC's mempool every time, 100% reproducible just by opening the dialog and tapping Save. On the other 12 EIP-1559 chains, the user's chosen priority fee never reaches the actually-paid `maxFeePerGas` cap, and since `priorityFeeWei` can independently exceed the uncombined `maxFeePerGasWei`, the signed fields can become protocol-invalid (`priority > max`) with no cross-field validation.

## Expected Behavior
`maxFeePerGasWei` should equal `baseFee + priorityFee`, matching the default (non-override) fee path, `calculateMaxFeePerGas`, and Windows' `evm.js` which does this explicitly.

## Steps to Reproduce
1. Open Advanced Gas Settings for a BSC send and tap Save without changing anything.
2. The signed transaction uses `gasPrice = 0` and is rejected by BSC's mempool.
3. On any other EVM chain, set a priority fee and observe it's dropped from the actually-paid `maxFeePerGas`.

## Solution
In `applyGasSettings()`, compute `maxFeePerGasWei = gs.baseFee + gs.priorityFee` instead of `gs.baseFee` alone; apply the same fix to `GasFeeSelection`'s estimate. For BSC (`chain.supportsLegacyGas`), either hide the Base-Fee/Priority-Fee split entirely or present a single legacy "Gas Price" field sourced from `eth_gasPrice`. Add cross-field validation so the invariant holds regardless of entry point.

### Must Do
- [ ] Fix `applyGasSettings()` to sum base fee + priority fee for `maxFeePerGasWei`
- [ ] Hide or adapt the Base-Fee/Priority-Fee split for legacy-gas chains (BSC)
- [ ] Add cross-field validation preventing `priorityFee > maxFeePerGas` from ever being signed

### Must NOT Do
- Don't change the default (non-override) fee calculation path — it's already correct
- Don't change how the raw Gwei text fields are parsed/validated for format

## Acceptance Criteria
- [ ] A BSC send via Advanced Gas Settings signs with a non-zero gasPrice
- [ ] `maxFeePerGasWei` includes the priority fee on all 13 EVM chains
- [ ] A regression test asserts `maxFeePerGasWei >= priorityFeeWei` after save

rkokhatskyi:
## Summary

Advanced Gas Settings saved `maxFeePerGasWei` as the base fee alone, dropping the priority fee from what actually gets signed and broadcast on every EVM chain. On BSC this went further: its "Max base fee" field was populated from `eth_getBlockByNumber` (pinned near zero by BEP-226) instead of `eth_gasPrice`, so saving the dialog without changing anything signed a transaction with `gasPrice = 0`, rejected by BSC's mempool every time.

## What changed

- `GasSettings.Eth.maxFeePerGasWei` is now a computed `baseFee + priorityFee`, used both when signing (`DefaultSendStrategy.applyGasSettings`) and when estimating the fee shown to the user (`GasFeeSelection.selectGasFeeForFeeEstimation`) — one source of truth so the two can't disagree.
- Legacy-gas chains (currently BSC) get a single "Gas price" field sourced from `eth_gasPrice`, with the priority-fee input hidden entirely instead of showing a meaningless base-fee/priority-fee split.
- `save()` clamps both fields to non-negative, so a pasted negative value can never push the signed fee below the priority fee.
- Both fields are blanked before every chain's fetch, so a value left over from a previously-opened chain (the dialog's ViewModel is scoped to the Send screen, not to one dialog open) can't leak into a save for a different chain.

## Screenshots

BSC's Advanced Gas Settings dialog, same build, `isLegacyGas` toggled to show before/after:

| Before | After |
|--------|-------|
| before | after |

Other EVM chains keep the existing two-field layout unchanged; only the underlying sum changed for them.

## Testing

- `./gradlew :app:testDebugUnitTest :data:testDebugUnitTest` — green
- `./gradlew :app:ktfmtCheckMain :app:ktfmtCheckTest :app:ktfmtCheckDebug :data:ktfmtCheck` — clean
- `./gradlew :app:lintDebug` — clean
- New/updated tests: `GasSettingsViewModelTest` (legacy-chain sourcing, negative-input clamping, stale-field-across-chain-switch regression), `SelectGasFeeForFeeEstimationTest` (sum instead of base-fee-alone), `DefaultSendStrategyTest` (signed `maxFeePerGasWei` matches the displayed estimate for the same `GasSettings.Eth`)

Closes #5397

## Summary by CodeRabbit

* **New Features**
  * Added legacy-gas handling in gas settings: shows a single gas price and hides priority-fee controls when applicable.
  * Updated the gas settings experience so loading state properly blocks saving during fee retrieval.
* **Localization**
  * Added translated “Gas price (GWEI)” labels across supported languages.
* **Bug Fixes**
  * Fixed Ethereum gas-fee math to consistently use the maximum fee (base + priority) for fee estimation and transaction signing.
  * Prevent negative fee inputs and avoid carrying over stale fee values when switching chains.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -48,6 +48,7 @@ import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.Coins
 import com.vultisig.wallet.data.models.getCoinLogo
 import com.vultisig.wallet.data.models.logo
+import com.vultisig.wallet.data.models.payload.BlockChainSpecific
 import com.vultisig.wallet.data.models.payload.DAppMetadata
 import com.vultisig.wallet.data.securityscanner.SecurityRiskLevel
 import com.vultisig.wallet.data.securityscanner.SecurityScannerResult
@@ -109,6 +110,7 @@ import com.vultisig.wallet.ui.models.peer.PeerDiscoveryUiModel
 import com.vultisig.wallet.ui.models.qbtc.QbtcClaimMaturingUtxoUiModel
 import com.vultisig.wallet.ui.models.qbtc.QbtcClaimUiState
 import com.vultisig.wallet.ui.models.qbtc.QbtcClaimUtxoUiModel
+import com.vultisig.wallet.ui.models.send.GasSettingsUiModel
 import com.vultisig.wallet.ui.models.swap.SwapFormUiModel
 import com.vultisig.wallet.ui.models.swap.SwapTransactionUiModel
 import com.vultisig.wallet.ui.models.swap.ValuedToken
@@ -139,6 +141,7 @@ import com.vultisig.wallet.ui.screens.referral.EmptyReferralBanner
 import com.vultisig.wallet.ui.screens.select.AssetUiModel
 import com.vultisig.wallet.ui.screens.select.SelectAssetScreen
 import com.vultisig.wallet.ui.screens.select.SelectAssetUiModel
+import com.vultisig.wallet.ui.screens.send.GasSettingsScreen
 import com.vultisig.wallet.ui.screens.send.VerifySendScreen
 import com.vultisig.wallet.ui.screens.settings.DiscountTiersScreenPreview
 import com.vultisig.wallet.ui.screens.settings.TierType
@@ -315,6 +318,8 @@ class PreviewActivity : ComponentActivity() {
                     "unbond_verify_before" -> UnbondVerifyPreview(after = false)
                     "unbond_verify_after" -> UnbondVerifyPreview(after = true)
                     "edit_folder" -> EditFolderPreview()
+                    "gas_settings_bsc_before" -> GasSettingsBscPreview(isLegacyGas = false)
+                    "gas_settings_bsc_after" -> GasSettingsBscPreview(isLegacyGas = true)
                     else -> SwapConfirmPreview()
                 }
             }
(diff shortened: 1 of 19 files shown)


rkokhatskyi:
Both fixed in e5c3b02c0. The Save-while-loading gap was real: blanking the fields on chain switch closed the stale-value leak but left Save pressable during a slow fetch or after a failed one, and a blank field saves as zero via toBigDecimalOrZero() — the same bug through a different door. Added isLoadingEthFee to GasSettingsUiModel, set true before the fetch and cleared only on success, and wired it into the Save button's isLoading (which already disables its click handler). On failure it stays true so Save stays disabled rather than let a zero fee through, with tests for both the success-clears and failure-stays-disabled cases. Also added the require() non-negative guard on GasSettings.Eth's constructor as suggested, plus a test asserting it throws for a negative baseFee or priorityFee.

johnnyluo (approved):
LGTM

PR #5413, opened 2026-07-27
fix(keysign): guard startForegroundService against rejection

### LINKED ISSUE #5407: Crash-adjacent: unguarded startForegroundService in TransactionStatusServiceManager.startPolling can strand the keysign done screen on the spinner
## Problem
`TransactionStatusServiceManager.startPolling()` calls `context.startForegroundService(intent)` (`TransactionStatusServiceManager.kt:62`) with no guard. On Android 12+ (API 31+), `Context.startForegroundService()` throws `ForegroundServiceStartNotAllowedException` (a subclass of `IllegalStateException`) when the app has no active foreground state — the exact restriction #5393/#5394 just fixed for the teardown path, still present on the startup path.

`startPolling` is called from `KeysignTxStatusPoller.pollStatusService`, itself called from `KeysignViewModel.startForegroundPolling()` right after a successful broadcast (`KeysignViewModel.kt:863`). That call runs inside `viewModelScope.safeLaunch { ... }`, and `safeLaunch` catches every `Throwable` except `CancellationException` (`data/utils/SafeLaunch.kt`). So a background broadcast (e.g. the app was backgrounded while the tx was in flight and the broadcast completes right after) throws inside `safeLaunch`, is silently swallowed, and `onStatus(TransactionResult.Pending)` never runs.

## Impact
The keysign done screen's `signingState` stays at `KeysignECDSA`/`isInProgress = true` — the UI never leaves the signing spinner even though the transaction was already broadcast successfully. The user has no way to see the tx status without leaving and re-entering the flow (if that's even possible), and no error is surfaced.

## Steps to Reproduce
1. Start a keysign for a chain with tx-status support (`txStatusConfigurationProvider.supportTxStatus(chain)` true).
2. Right after the transaction broadcasts, background the app (before `startForegroundPolling` → `startPolling` runs) — e.g. tap Home or lock the screen at the moment the broadcast result returns.
3. `startForegroundService` throws `ForegroundServiceStartNotAllowedException`; `safeLaunch` swallows it.
4. Return to the app: the done screen is stuck showing the signing spinner, not the broadcasted/pending state.

## Suggested Fix
Wrap `context.startForegroundService(intent)` in the same `runCatching` pattern already used a few lines below for `bindService`, and have `startPolling` return `false` (or a distinct tri-state) on failure so `KeysignTxStatusPoller.pollStatusService` can fall back to showing `TransactionResult.Broadcasted`/`Pending` from the already-known broadcast result instead of hanging on `Pending` forever, mirroring how a rejected `bindService()` is already handled.

## Context
Found during code review of #5394 (fix for the teardown-side `BackgroundServiceStartNotAllowedException`, `TransactionStatusServiceManager.kt:98` prior to that fix). This is the equivalent startup-side gap: same restriction class, different call (`startForegroundService` vs the old `startService`), pre-existing and unmodified by #5394. Nine independent review passes across two review rounds on #5394 confirmed this is real and reachable but out of scope for that PR since the line is byte-identical to what shipped before it.

johnnyluo:
## Summary
- `TransactionStatusServiceManager.startPolling()` called `context.startForegroundService(intent)` unguarded. On API 31+, this throws `ForegroundServiceStartNotAllowedException` when the app has no active foreground state (e.g. backgrounded right as a broadcast completes).
- The call runs inside `safeLaunch`, which swallows the exception, stranding the keysign done screen on the signing spinner with no error surfaced.
- Wrapped the call in `runCatching`, mirroring the existing `bindService` guard a few lines below, and return `false` on failure.
- No caller changes needed: `KeysignTxStatusPoller.pollStatusService` already treats `bound == false` as "emit `Pending`, return `null`" (added in #5394's fix for the equivalent teardown-side gap), so this closes the startup-side gap the same way.

## Issue
Closes #5407

## Test Plan
- [x] `./gradlew compileDebugKotlin`
- [x] `./gradlew lint`
- [ ] Manual: start a keysign for a chain with tx-status support, background the app right as the broadcast completes, confirm the done screen doesn't get stuck on the signing spinner

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction status polling reliability when background service startup is restricted.
  * Polling now fails gracefully instead of throwing an error when the service cannot start.
  * Callers receive a clear failure result and are no longer left waiting for an unavailable service connection.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/services/TransactionStatusServiceManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/services/TransactionStatusServiceManager.kt
@@ -49,8 +49,10 @@ constructor(@param:ApplicationContext private val context: Context) {
     /**
      * Starts the foreground status service for [txHash] on [chain] and binds to it.
      *
-     * @return `false` when the system rejected the binding: `onServiceConnected()` never fires in
-     *   that case, so [serviceReady] stays `false` and callers must not wait on it.
+     * @return `false` when the system rejected the start (e.g.
+     *   `ForegroundServiceStartNotAllowedException` on API 31+ with no foreground state) or the
+     *   binding: `onServiceConnected()` never fires in that case, so [serviceReady] stays `false`
+     *   and callers must not wait on it.
      */
     fun startPolling(txHash: String, chain: Chain): Boolean {
         val intent =
@@ -59,7 +61,11 @@ constructor(@param:ApplicationContext private val context: Context) {
                 putExtra(TransactionStatusService.EXTRA_TX_HASH, txHash)
                 putExtra(TransactionStatusService.EXTRA_CHAIN, chain.raw)
             }
-        context.startForegroundService(intent)
+        val started =
+            runCatching { context.startForegroundService(intent) }
+                .onFailure { Timber.w(it, "Failed to start transaction status service") }
+                .isSuccess
+        if (!started) return false
         if (isBound) return true
 
         // Marked bound on request, not on connect: a binding that has not connected yet still


rkokhatskyi (approved):
LGTM

PR #5414, opened 2026-07-27
fix(price): add Mantle to CoinGecko contract-price chain map

### LINKED ISSUE #5406: [Fix] CoinGecko contract-price chain map omits Mantle — Android-only pricing gap vs iOS
# [Fix] CoinGecko contract-price chain map omits Mantle — Android-only pricing gap vs iOS

## Problem
`CoinGeckoApi`'s `coinGeckoAssetId` map lists Ethereum, Avalanche, Base, Blast, Arbitrum, Polygon, Optimism, BscChain, ZkSync, and Solana, then falls through to `else -> error(...)`. `Chain.Mantle` is absent, so `getContractsPrice(chain=Mantle, ...)` always throws and returns an empty map — an unconditional miss, not merely conditional on token coverage. Because the LI.FI fallback (EVM-013) also only maps Ethereum, both price sources fail for Mantle, leaving non-native token prices permanently at $0. iOS's `CryptoPriceService.coinGeckoPlatform` already includes Mantle.

## Expected Behavior
Mantle should resolve a real CoinGecko platform identifier, matching iOS.

## Steps to Reproduce
1. Hold any non-native ERC-20 on Mantle.
2. Observe no USD price is shown, while the identical iOS user sees a real price for the same token/contract.

## Solution
Add `Chain.Mantle -> "mantle"` to `CoinGeckoApi.coinGeckoAssetId`, matching iOS's `coinGeckoPlatform`.

### Must Do
- [ ] Add the Mantle entry to `coinGeckoAssetId`

### Must NOT Do
- Don't add entries for CronosChain/Sei/Hyperliquid in this ticket — those are absent from iOS too (a shared gap, not this ticket's scope)

## Acceptance Criteria
- [ ] A non-native Mantle ERC-20's USD price resolves via CoinGecko

rkokhatskyi:
## Summary
`coinGeckoAssetId` had no entry for `Chain.Mantle`, so `getContractsPrice` hit the catch-all `else -> error(...)` before ever reaching the network and non-native Mantle tokens never resolved a USD price. iOS's `CryptoPriceService.coinGeckoPlatform` already maps Mantle to the same `"mantle"` platform id, so this adds the equivalent entry on Android.

Overlap note: PR #5390 (https://github.com/vultisig/vultisig-android/pull/5390) touches the same `when` block to add a Robinhood entry; low conflict risk since it's a one-line addition in an unrelated ~20-file feature PR (Robinhood chain support), not the same behavior.

## Test plan
- `./gradlew --no-daemon :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.*"` — passing, including a new regression test that fails with `UninitializedPropertyAccessException` when the fix is reverted (proving `coinGeckoAssetId` previously threw before any request was sent) and a sibling test confirming an unmapped chain still degrades to an empty map.
- `./gradlew --no-daemon :data:ktfmtFormat` — no diff

Closes #5406

## Summary by CodeRabbit

- **Bug Fixes**
  - Added cryptocurrency price support for the Mantle network through CoinGecko.
  - Ensured contract price requests use the correct Mantle platform identifier.
  - Preserved graceful handling for networks without supported price mappings.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/CoinGeckoApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/CoinGeckoApi.kt
@@ -94,6 +94,7 @@ internal class CoinGeckoApiImpl @Inject constructor(private val http: HttpClient
                 Chain.BscChain -> "binance-smart-chain"
                 Chain.ZkSync -> "zksync"
                 Chain.Solana -> "solana"
+                Chain.Mantle -> "mantle"
 
                 else -> error("No CoinGecko asset id for chain $this")
             }
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/CoinGeckoApiContractPriceChainTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/CoinGeckoApiContractPriceChainTest.kt
@@ -1,0 +1,56 @@
+package com.vultisig.wallet.data.api
+
+import com.vultisig.wallet.data.models.Chain
+import io.ktor.client.HttpClient
+import io.ktor.client.engine.mock.MockEngine
+import io.ktor.client.engine.mock.respond
+import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
+import io.ktor.http.HttpStatusCode
+import io.ktor.serialization.kotlinx.json.json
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5415, opened 2026-07-27
fix(price): resolve LI.FI chain id for all EVM chains

### LINKED ISSUE #5405: [Fix] LiFi price-fallback chain map only covers Ethereum, silently fails for 12 other EVM chains with no log
# [Fix] LiFi price-fallback chain map only covers Ethereum, silently fails for 12 other EVM chains with no log

## Problem
`LiQuestApi`'s `lifiChainId` resolver only maps `Chain.Ethereum -> "eth"`; every other chain hits `else -> error(...)`. This is the resolver for the LI.FI price-fallback call used whenever a held ERC-20's price isn't in CoinGecko — a common case for smaller-cap/L2-native tokens, and the guaranteed case for every non-native token on CronosChain/Mantle/Sei/Hyperliquid. The thrown error is swallowed to `null` inside a try/catch with no `Timber` log at all, unlike every other fallback in the file — making the failure invisible in production.

## Expected Behavior
The LI.FI price fallback should resolve a real chain identifier for all 13 EVM chains, matching iOS's `fetchLifiTokenPrice`, which draws from `Chain.ticker` and attempts a lookup for every chain.

## Steps to Reproduce
1. Hold a non-Ethereum EVM ERC-20 that CoinGecko has no contract-price data for.
2. The portfolio/balance view shows a silent $0 valuation, no error surfaced anywhere.

## Solution
Build a complete `Chain -> LI.FI chain identifier` map mirroring `oneInchChainId()`/`swapAssetName()`'s completeness, add a `Timber` warning in the catch branch, and add a unit test asserting non-Ethereum chains resolve a real identifier instead of throwing.

### Must Do
- [ ] Build a complete per-chain LI.FI identifier map for all 13 EVM chains
- [ ] Add a `Timber` warning when the fallback fails, matching the rest of the file's logging convention

### Must NOT Do
- Don't change the CoinGecko price path (tracked separately as EVM-014)
- Don't change the LI.FI swap-quote chain-id resolution, only the price-fallback resolver

## Acceptance Criteria
- [ ] All 13 EVM chains resolve a real LI.FI chain identifier for price fallback
- [ ] New test asserts non-Ethereum chains no longer throw

rkokhatskyi:
## Summary
`LiQuestApi`'s `Chain.lifiChainId` resolver only mapped `Chain.Ethereum -> "eth"` and threw for every other chain, so the LI.FI price-fallback call (used when CoinGecko has no contract price) failed for all 12 other EVM chains. The failure was also swallowed silently in `TokenPriceRepository` with no log, so an affected token just showed a $0 valuation with no trace.

- `lifiChainId` now resolves the chain's real numeric EVM chain id via a new `Chain.evmChainId()` extension, matching iOS's `fetchLifiTokenPrice`, which sends `String(chain.chainID)`. Verified against the live `li.quest` API directly: 7 of the 13 EVM chains reject a ticker-style value (`avax`, `base`, `blast`, `op`, `bnb`, `zk`, `hype`) with HTTP 400, while every chain accepts its numeric id.
- `evmChainId()` reuses the existing `oneInchChainId()` table (gated on `TokenStandard.EVM` so non-EVM chains correctly return null despite `oneInchChainId()` also carrying a non-EVM Solana pseudo-id) plus the same `Sei -> "1329"` override `ContractAbiRepository` already used for Sourcify lookups. That repository's private copy of this exact table is now replaced with a call to the shared extension instead of carrying two near-identical private tables.
- Added the missing `Timber.e` log (using `%s` format args) in `TokenPriceRepository.getLifiContractPriceInUsd`'s catch block.
- Non-EVM chains still throw (unchanged), since this fallback is EVM-only; the CoinGecko price path and the separate LI.FI swap-quote chain-id resolver are untouched.

Closes #5405

## Test plan
- `./gradlew --no-daemon :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.LiQuestApiImplTest" --tests "com.vultisig.wallet.data.repositories.TokenPriceRepositoryImplTest" --tests "com.vultisig.wallet.data.repositories.ContractAbiRepositoryTest"` — passing, including a new test asserting all 13 EVM chains resolve the correct numeric chain id without throwing, and that non-EVM chains still throw
- `./gradlew --no-daemon :data:ktfmtCheck` — clean
- Verified all 13 numeric chain ids return HTTP 200 from the live `li.quest/v1/token` endpoint

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Li.Fi token price requests by deriving the correct LiQuest network identifier for EVM networks.
  * Added detailed error logging when token price lookups fail, including the relevant chain and contract info.
  * Ensured non-EVM networks fail clearly instead of sending an incorrect identifier.
* **Tests**
  * Added tests verifying EVM network identifier resolution and that non-EVM requests throw an exception.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/LiQuestApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/LiQuestApi.kt
@@ -2,8 +2,7 @@ package com.vultisig.wallet.data.api
 
 import com.vultisig.wallet.data.api.models.LiQuestResponseJson
 import com.vultisig.wallet.data.models.Chain
-import com.vultisig.wallet.data.models.TokenStandard
-import com.vultisig.wallet.data.models.ticker
+import com.vultisig.wallet.data.models.evmChainId
 import io.ktor.client.HttpClient
 import io.ktor.client.call.body
 import io.ktor.client.request.get
@@ -31,9 +30,5 @@ internal class LiQuestApiImpl @Inject constructor(private val http: HttpClient) 
             .body()
 
     private val Chain.lifiChainId: String
-        get() =
-            when {
-                standard == TokenStandard.EVM -> ticker().lowercase()
-                else -> error("lifi chain id not found for chain $this")
-            }
+        get() = evmChainId() ?: error("lifi chain id not found for chain $this")
 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/LiQuestApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/LiQuestApi.kt
@@ -2,6 +2,8 @@ package com.vultisig.wallet.data.api
 
 import com.vultisig.wallet.data.api.models.LiQuestResponseJson
 import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.TokenStandard
+import com.vultisig.wallet.data.models.ticker
 import io.ktor.client.HttpClient
 import io.ktor.client.call.body
 import io.ktor.client.request.get
@@ -30,8 +32,8 @@ internal class LiQuestApiImpl @Inject constructor(private val http: HttpClient) 
 
     private val Chain.lifiChainId: String
         get() =
-            when (this) {
-                Chain.Ethereum -> "eth"
+            when {
+                standard == TokenStandard.EVM -> ticker().lowercase()
                 else -> error("lifi chain id not found for chain $this")
             }
 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
@@ -271,7 +271,7 @@ constructor(
             BigDecimal(liQuestApi.getLifiContractPriceUsd(chain, contract).priceUsd)
         } catch (e: Exception) {
             if (e is kotlinx.coroutines.CancellationException) throw e
-            Timber.e(e, "Failed to fetch Lifi price for $chain contract: $contract")
+            Timber.e(e, "Failed to fetch Lifi price for %s contract: %s", chain, contract)
             null
         }
 
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
@@ -271,6 +271,7 @@ constructor(
             BigDecimal(liQuestApi.getLifiContractPriceUsd(chain, contract).priceUsd)
         } catch (e: Exception) {
             if (e is kotlinx.coroutines.CancellationException) throw e
+            Timber.e(e, "Failed to fetch Lifi price for $chain contract: $contract")
             null
         }
 
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/LiQuestApiImplTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/LiQuestApiImplTest.kt
@@ -1,0 +1,82 @@
+package com.vultisig.wallet.data.api
+
+import com.vultisig.wallet.data.models.Chain
+import io.ktor.client.HttpClient
+import io.ktor.client.engine.mock.MockEngine
+import io.ktor.client.engine.mock.respond
+import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
+import io.ktor.client.request.HttpRequestData
+import io.ktor.http.ContentType
+import io.ktor.http.HttpHeaders
+import io.ktor.http.HttpStatusCode
+import io.ktor.http.headersOf
+import io.ktor.serialization.kotlinx.json.json
+import kotlin.test.assertFailsWith
+import kotlinx.coroutines.test.runTest
+import kotlinx.serialization.json.Json
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+/**
+ * Pins the `chain` query param LI.FI's price-fallback endpoint receives for every EVM chain,
+ * matching iOS's `chain.ticker.lowercased()`. Every non-Ethereum EVM chain used to throw instead of
+ * resolving a real identifier.
+ */
+class LiQuestApiImplTest {
+
+    private val jsonHeaders =
+        headersOf(HttpHeaders.ContentType, ContentType.Application.Json.toString())
+
+    private fun apiCapturing(onRequest: (HttpRequestData) -> Unit): LiQuestApi =
+        LiQuestApiImpl(
+            HttpClient(
+                MockEngine { request ->
+                    onRequest(request)
+                    respond(
+                        content = """{"priceUSD":"1.23"}""",
+                        status = HttpStatusCode.OK,
+                        headers = jsonHeaders,
+                    )
+                }
+            ) {
+                install(ContentNegotiation) { json(Json { ignoreUnknownKeys = true }) }
+            }
+        )
+
+    @Test
+    fun `every EVM chain resolves a lifi chain id matching iOS ticker lowercased`() = runTest {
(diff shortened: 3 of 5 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -271,6 +271,7 @@ constructor(
             BigDecimal(liQuestApi.getLifiContractPriceUsd(chain, contract).priceUsd)
         } catch (e: Exception) {
             if (e is kotlinx.coroutines.CancellationException) throw e
+            Timber.e(e, "Failed to fetch Lifi price for $chain contract: $contract")
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed the Timber call to use %s placeholders instead of string interpolation, matching the repo convention. Left the broad catch (e: Exception) as-is though: it's pre-existing (I only added the log line inside it), and the identical broad-catch-to-null pattern already exists in this same file's three other price-fallback catches (fetchThorPoolPrices, fetchMayaPoolPrices, the THOR contract-price catch a few lines below), so narrowing just this one would make it inconsistent with its siblings; tightening error handling across all of them is a bigger, separate change outside this chain-id ticket's scope.

data/src/test/kotlin/com/vultisig/wallet/data/api/LiQuestApiImplTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import io.ktor.client.engine.mock.MockEngine
+import io.ktor.client.engine.mock.respond
+import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
+import io.ktor.client.request.HttpRequestData
+import io.ktor.http.ContentType
+import io.ktor.http.HttpHeaders
+import io.ktor.http.HttpStatusCode
+import io.ktor.http.headersOf
+import io.ktor.serialization.kotlinx.json.json
+import kotlin.test.assertFailsWith
```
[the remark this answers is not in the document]
rkokhatskyi: This isn't actually broken: kotlin.test.assertFailsWith is inline, so its lambda gets spliced into the caller and inherits the enclosing runTest suspend context rather than needing a real suspend function type, and the repo already relies on this exact pattern in several other tests (e.g. MayachainBondRepositoryImplTest.kt:124, VultiSignerRepositoryImplBackupCodeTest.kt:84, both call a suspend repository function directly inside assertFailsWith). This test file itself compiles and passes (./gradlew :data:testDebugUnitTest --tests LiQuestApiImplTest, 2/2 green), so leaving it as-is.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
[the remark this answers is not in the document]
rkokhatskyi: No need, thanks. It's an existing pattern across the file rather than something new this PR introduced, so I'd rather not open a tracking issue on its behalf.

data/src/main/kotlin/com/vultisig/wallet/data/api/LiQuestApi.kt
```diff
@@ -30,8 +32,8 @@ internal class LiQuestApiImpl @Inject constructor(private val http: HttpClient)
 
     private val Chain.lifiChainId: String
         get() =
-            when (this) {
-                Chain.Ethereum -> "eth"
+            when {
+                standard == TokenStandard.EVM -> ticker().lowercase()
```
johnnyluo: **iOS does not actually do this, and this breaks 7 of the 13 target chains.**

I checked `vultisig-ios`'s `TokenPriceSource.swift` (`fetchLifiTokenPrice`) directly: it does **not** send `chain.ticker.lowercased()`. It sends the numeric EVM chain ID as a string:

```swift
guard let chainID = chain.chainID else { ... }
CryptoPriceAPI.lifiTokenPrice(network: String(chainID), address: contract)
```

`Chain.chainID: Int?` in `Chain.swift` is a separate per-chain table (Ethereum→1, Avalanche→43114, Base→8453, Blast→81457, Arbitrum→42161, Polygon→137, Optimism→10, BSC→56, Cronos→25, zkSync→324, Mantle→5000, Hyperliquid→999, Sei→1329) — the exact kind of table this PR's description claims iOS avoids. The `.lowercased()` in the Kotlin/iOS request builder is a no-op on a numeric string; it's not evidence of a ticker-based scheme.

I confirmed against the live API (`https://li.quest/v1/token?chain=<value>&token=0x0…`):

| `chain` param | Result |
|---|---|
| `eth`, `arb`, `pol`, `cro`, `mnt`, `sei` (ticker) | 200 OK — coincidentally valid LI.FI chain keys |
| `avax`, `base`, `blast`, `op`, `bnb`, `zk`, `hype` (ticker) | **400** `{"message":"/chain must be number, /chain must be equal to one of the allowed values..."}` |
| any numeric chain ID (1, 43114, 8453, 56, 25, 999, ...) | 200 OK for every EVM chain tested |

So with this change, 5 chains gain working prices (arb, pol, cro, mnt, sei) but 7 remain broken (avax, base, blast, op, bnb, zk, hype) — they'll now fail with a logged `Timber.e` instead of silently, but the $0 valuation bug persists for the majority of chains this PR claims to fix.

Suggest resolving to the numeric EVM chain ID (matching iOS's actual `chainID` table) instead of `ticker().lowercase()`. Worth checking whether the codebase already has an EVM chain-ID table elsewhere (RPC/gas config, WalletCore chainId, etc.) to reuse rather than adding a new one.

johnnyluo (commented):
Cross-checked this against the actual iOS source and the live LI.FI API — see inline comments. The core claim in the PR description ("matching iOS's `fetchLifiTokenPrice`, which sends `chain.ticker.lowercased()`") doesn't hold up, and as a result about half the target chains are still broken, just with a logged error now instead of a silent one.

data/src/test/kotlin/com/vultisig/wallet/data/api/LiQuestApiImplTest.kt
```diff
@@ -0,0 +1,82 @@
+package com.vultisig.wallet.data.api
```
johnnyluo, line 1: Since the chain-id values asserted here (eth, avax, base, ...) are the same wrong values being sent to LI.FI (see comment on `LiQuestApi.kt`), this test suite passes while 7 of the 13 chains it covers would still get a 400 from the real API. Worth adding a note or fixing alongside the underlying resolver so the test doesn't lock in the incorrect mapping.

data/src/main/kotlin/com/vultisig/wallet/data/api/LiQuestApi.kt
rkokhatskyi: Confirmed and fixed, thank you. My iOS checkout was 10 days stale; PR #4897 upstream ("use numeric chain ID for LiFi token price requests") switched fetchLifiTokenPrice from ticker to chain.chainID after that, and the S12 refactor (#4913) moved it into TokenPriceSource.swift, which is why I'd read the old CryptoPriceService.swift version. Pulled origin/main on vultisig-ios and confirmed chain.chainID's table directly, then independently curled li.quest myself and got the same 400s you did on avax/base/blast/op/bnb/zk/hype and 200 on every numeric id. Now resolving via a new Chain.evmChainId() (Ethereum through 1, ..., Sei through the same 1329 override) and, per your reuse suggestion, replaced ContractAbiRepository's identical private table with a call to that shared extension instead of leaving two copies. Pushed in 236c14bad.

data/src/test/kotlin/com/vultisig/wallet/data/api/LiQuestApiImplTest.kt
rkokhatskyi: Fixed: the test now asserts the numeric chain id for each of the 13 EVM chains (1, 43114, 8453, ..., 1329 for Sei) instead of the wrong ticker values, matching what's actually sent to li.quest now.

johnnyluo (approved):
LGTM

PR #5417, opened 2026-07-27
fix(solana): fall back to cached rent-exempt reserve on RPC failure

### LINKED ISSUE #5345: [Fix] Solana rent-exemption fee estimate silently drops to zero on RPC failure
# [Fix] Solana rent-exemption fee estimate silently drops to zero on RPC failure

## Problem
When the call to fetch the minimum balance required for rent exemption fails, the code catches the exception and returns zero instead of a safe fallback. That zero value feeds directly into both the displayed fee estimate and the pre-send balance-sufficiency check, so a transient RPC failure can understate the real fee needed for an SPL token send that requires creating a new associated token account.

## Expected Behavior
An RPC failure while fetching the rent-exemption amount should not silently resolve to zero — it should fall back to a known-safe minimum value so the displayed fee estimate and balance check stay accurate.

## Steps to Reproduce
1. Send an SPL token to a recipient who needs a new associated token account created, while the rent-exemption RPC call fails.
2. Observe the displayed fee and the pre-send balance check both silently omit the rent-exemption amount.

## Solution
Cache the last successfully-fetched rent-exemption value and use that as the fallback on failure, instead of zero — self-updating, so it tracks the live value whenever the RPC call has succeeded at least once, rather than depending on a single hardcoded literal. For the cold-start case (no successful fetch yet this session), bootstrap with the standard 165-byte SPL Token Account rent-exempt minimum: **2,039,280 lamports (~0.00203928 SOL)**.

This constant isn't an arbitrary guess: it's exactly reproducible from Solana's own rent formula — `(128-byte account overhead + 165-byte token account data) × 3,480 lamports/byte-year × 2-year exemption threshold = 2,039,280`. Independently confirmed live against both `https://api.vultisig.com/solana/` (the exact endpoint this app calls) and Solana's public `api.mainnet-beta.solana.com`, both returning `2039280` for `getMinimumBalanceForRentExemption(165)` right now. iOS already uses this identical constant (`SolanaHelper.ataRentLamports = 2_039_280`) as its *only* value for this fee component, with no live RPC call at all — so a cached-value-with-this-bootstrap approach is strictly more robust than iOS's own existing approach, not a downgrade.

### Must Do
- [ ] Cache the last successfully-fetched rent-exemption value; use it as the fallback on a subsequent RPC failure
- [ ] Bootstrap with 2,039,280 lamports when no successful fetch has happened yet (cold start)
- [ ] Add a unit test asserting the cached value is used when the RPC call fails after a prior success, and the bootstrap constant is used on a cold-start failure

### Must NOT Do
- Don't change the happy-path RPC call behavior
- Don't change unrelated fee-calculation logic

## Acceptance Criteria

rkokhatskyi:
## Summary
`getMinimumBalanceForRentExemption()` caught any RPC failure and returned zero, which fed directly into both the displayed fee estimate and the pre-send balance check for an SPL send that needs to create a new associated token account — a transient RPC hiccup could understate the real fee needed.

- The no-arg overload now delegates to the existing byte-length overload and treats a non-positive result the same way `SolanaDelegateViewModel.fetchRentReserve` already treats its own rent-exempt read: fall back instead of trusting a zero.
- The fallback is the last successfully-fetched value this session, cached in a `@Volatile` field (same pattern as `CosmosFeeService`/`TronFeeService`), so it tracks the live network value once a read has succeeded.
- Before any call has succeeded (cold start), it falls back to the standard 165-byte SPL Token Account rent-exempt minimum (2,039,280 lamports) — the same constant iOS's `SolanaHelper.ataRentLamports` uses as its only source, with no live RPC call at all.
- The 200-byte staking overload (`getMinimumBalanceForRentExemption(dataLength)`) is unchanged; it still returns zero on failure and its one caller already supplies its own fallback constant.

Closes #5345

## Test plan
- `./gradlew --no-daemon :data:testDebugUnitTest` — passing, including new tests asserting the live value is returned on success, the cached value is used when a later fetch fails, and the bootstrap constant is used on a cold-start failure
- `./gradlew --no-daemon :data:ktfmtCheck` — clean
- `./gradlew --no-daemon assembleDebug` — successful

## Summary by CodeRabbit

* **Improvements**
  * Solana rent-exemption estimates now use the latest available network value.
  * If the network is temporarily unavailable, the app retains the last successful value or uses a standard fallback for the initial request.

* **Tests**
  * Added coverage for successful lookups, temporary RPC failures, and first-use fallback behavior.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/SolanaApi.kt
@@ -66,6 +66,13 @@ import wallet.core.jni.SolanaAddress
 interface SolanaApi {
     suspend fun getBalance(address: String): BigInteger
 
+    /**
+     * Live rent-exempt reserve for a new 165-byte SPL Associated Token Account. The last
+     * successfully-fetched value is cached and reused as the fallback when the underlying RPC read
+     * fails, so a transient error never silently drops the fee to zero. Before any call has
+     * succeeded this session, falls back to the standard 165-byte rent-exempt minimum (2,039,280
+     * lamports).
+     */
     suspend fun getMinimumBalanceForRentExemption(): BigInteger
 
     suspend fun getRecentBlockHash(): String
@@ -101,9 +108,10 @@ interface SolanaApi {
     suspend fun checkStatus(txHash: String): SolanaRpcResponseJson<SolanaSignatureStatusesResult>?
 
     /**
-     * Minimum lamports for rent exemption of an account of the given byte length. The no-arg
-     * overload targets SPL token accounts (165 bytes); staking passes the 200-byte stake-account
-     * size. Returns [BigInteger.ZERO] on RPC failure.
+     * Minimum lamports for rent exemption of an account of [dataLength] bytes (e.g. the 200-byte
+     * stake-account size used by staking). Returns [BigInteger.ZERO] on RPC failure; callers that
+     * need a non-zero fallback supply their own (see the no-arg [getMinimumBalanceForRentExemption]
+     * for the SPL-account overload that caches one internally).
      */
     suspend fun getMinimumBalanceForRentExemption(dataLength: Int): BigInteger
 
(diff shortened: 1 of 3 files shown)


rkokhatskyi:
On the broad-catch nitpick: that catch(Exception) block belongs to the existing dataLength overload (unchanged by this PR) and is the same shape used by every other method in this file — getBalance, getMedianPriorityFee, getSPLTokens, getAccountOwner, checkStatus, and so on all catch Exception and rethrow only CancellationException, and the staking caller (SolanaDelegateViewModel.fetchRentReserve) already relies on that same overload returning zero on failure rather than throwing. Narrowing it here would diverge from that established convention and would also change behavior for the staking path, which is out of scope for this fix. It's also not silent: the catch block already logs the real exception via Timber.e before falling back, so a genuine deserialization or programming bug still surfaces in logs, it just doesn't crash a fee estimate that runs on every debounced keystroke. A repo-wide pass on exception granularity across SolanaApiImp would be a separate, larger change, so leaving this as-is for now.

johnnyluo (approved):
LGTM

Issue #5399, opened 2026-07-26
[Fix] getGasPrice() returns zero on RPC error, feeding a 0 wei legacy gasPrice into a signed BSC transaction


rkokhatskyi:
# [Fix] getGasPrice() returns zero on RPC error, feeding a 0 wei legacy gasPrice into a signed BSC transaction

## Problem
`EvmApi.getGasPrice()` returns `BigInteger.ZERO` when `rpcResp.error != null`. BSC is the only chain with `supportsLegacyGas = true`, so it's the only chain reaching `EthereumFeeService.calculateLegacyGasFees()`, which uses the returned gas price with no floor or sanity check before signing.

## Expected Behavior
An RPC error while fetching the gas price should not silently resolve to zero — it should throw (matching iOS's `fetchGasPrice()`, which throws via the decode-failure path) so the caller can surface a retryable error before staging the transaction.

## Steps to Reproduce
1. Send or swap on BSC while `eth_gasPrice` errors (transient RPC failure).
2. `getGasPrice()` returns 0 instead of throwing.
3. `calculateLegacyGasFees` signs a transaction with `gasPrice = 0`.
4. The full MPC ceremony completes, then broadcast fails as underpriced.

## Solution
Throw on `rpcResp.error != null` in `getGasPrice()` rather than silently continuing, matching the pattern already established for other EVM RPC-error-swallow fixes in this codebase (#5308/#5312 fixed the identical pattern for `getBalance`/`getBalances`).

### Must Do
- [ ] `getGasPrice()` throws instead of returning `BigInteger.ZERO` on RPC error

### Must NOT Do
- Don't change the happy-path gas price parsing
- Don't change any other RPC method's error handling in this file beyond `getGasPrice()`

## Acceptance Criteria
- [ ] `./gradlew assembleDebug` succeeds
- [ ] A BSC send with a failing `eth_gasPrice` call surfaces a retryable error instead of signing a zero-price transaction
- [ ] New test exercises the RPC-error path for `getGasPrice()`

PR #5418, opened 2026-07-27
fix(evm): throw on base-fee/priority-fee RPC error instead of signing a zeroed tx

### LINKED ISSUE #5400: [Fix] getBaseFee()/getMaxPriorityFeePerGas() return zero on RPC error, can sign an EIP-1559 tx priced below the real base fee
# [Fix] getBaseFee()/getMaxPriorityFeePerGas() return zero on RPC error, can sign an EIP-1559 tx priced below the real base fee

## Problem
`EvmApi.getBaseFee()` and `getMaxPriorityFeePerGas()` both return `BigInteger.ZERO` on RPC error. `EthereumFeeService.calculateEip1559Fees()` computes `maxFeePerGas = baseFee.increaseByPercent(20).add(priorityFee)` directly off `getBaseFee()` with no floor or error check — if `baseFee` silently becomes 0, `maxFeePerGas` collapses to roughly just the priority-fee component, virtually guaranteed below the real current base fee. Avalanche compounds this: its priority-fee calculation calls the same zero-swallowing `getMaxPriorityFeePerGas()` directly, so it can independently double-zero both fee components.

## Expected Behavior
An RPC error while fetching base fee or priority fee should throw (matching iOS's `RpcEvmService.swift`, which throws in both cases) rather than degrading to zero, so the failure can be surfaced as a retryable error before staging the transaction.

## Steps to Reproduce
1. Send or swap on any EIP-1559 chain (Ethereum, Arbitrum, Optimism, Base, Blast, Polygon, Avalanche, CronosChain, Mantle, Sei, Hyperliquid) while `eth_getBlockByNumber` or the priority-fee RPC call errors.
2. `maxFeePerGas` computes from a silently-zeroed base fee.
3. Complete a full multi-device MPC signing ceremony with a normal-looking fee estimate shown.
4. Broadcast rejects as underpriced.

## Solution
Throw on `rpcResp.error != null` in both `getBaseFee()` and `getMaxPriorityFeePerGas()`; let existing callers propagate the failure as a retryable UI error before staging the transaction.

### Must Do
- [ ] `getBaseFee()` throws instead of returning `BigInteger.ZERO` on RPC error
- [ ] `getMaxPriorityFeePerGas()` throws instead of returning `BigInteger.ZERO` on RPC error

### Must NOT Do
- Don't change the happy-path fee calculation logic
- Don't change ZkSync's separate fee path (tracked separately as EVM-003/EVM-007)

## Acceptance Criteria
- [ ] An EIP-1559 send with a failing base-fee or priority-fee RPC call surfaces a retryable error instead of signing an underpriced transaction
- [ ] New test exercises the RPC-error path for both methods, including the Avalanche double-zero case

rkokhatskyi:
## Summary
- `EvmApi.getBaseFee()` and `getMaxPriorityFeePerGas()` silently returned `BigInteger.ZERO` on an `eth_getBlockByNumber`/`eth_maxPriorityFeePerGas` RPC error. `EthereumFeeService.calculateEip1559Fees()` built `maxFeePerGas` directly off `getBaseFee()` with no floor check, so a transient RPC error could sign a transaction priced near zero. Avalanche is hit twice, since its priority fee also comes straight from `getMaxPriorityFeePerGas()`.
- Both methods now throw `NetworkException` on `rpcResp.error != null`, matching the propagate-don't-swallow pattern already used by `getBalance`/`getBalances` (#5308/#5312) and iOS's throwing `RpcEvmService`. The happy-path parsing is untouched.

Overlap note: PR #5416 (https://github.com/vultisig/vultisig-android/pull/5416) also touches `EvmApi.kt`, for the sibling `getGasPrice()` RPC-error fix (issue #5399); low conflict risk since it edits `getGasPrice()`/`estimateGasForCallDataTransfer()`, non-adjacent to this diff's `getBaseFee()`/`getMaxPriorityFeePerGas()`.

Closes #5400

## Test plan
- `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.EvmApiFeeTest"` — new tests: RPC error propagates for both methods, happy path still parses, and an Avalanche-chain instance fails closed on both calls for the same RPC outage
- `./gradlew :data:testDebugUnitTest` — full data module suite green
- `./gradlew :data:ktfmtCheck` — clean
- `./gradlew assembleDebug` — succeeds

## Summary by CodeRabbit

* **Bug Fixes**
  * RPC failures while retrieving EIP-1559 fee data now surface as network errors instead of being treated as zero values.
  * Prevents transactions from being prepared with potentially invalid or underpriced fee components.

* **Tests**
  * Added coverage for successful fee parsing and error handling across supported EVM networks.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -459,8 +459,13 @@ class EvmApiImp(
     override suspend fun getMaxPriorityFeePerGas(): BigInteger {
         val rpcResp = fetch<RpcResponse>("eth_maxPriorityFeePerGas", buildJsonArray {})
         if (rpcResp.error != null) {
-            Timber.d("get max priority fee per gas , error: ${rpcResp.error.message}")
-            return BigInteger.ZERO
+            // A failed read must propagate, never collapse into ZERO: a zero priority fee can
+            // combine with a zeroed base fee (below) to sign an EIP-1559 tx priced under the real
+            // network fee, which broadcast then rejects as underpriced (#5400).
+            throw NetworkException(
+                httpStatusCode = 0,
+                message = "max priority fee per gas rpc error: ${rpcResp.error.message}",
+            )
         }
 
         return rpcResp.result.convertToBigIntegerOrZero()
@@ -585,8 +590,11 @@ class EvmApiImp(
                     },
             )
         if (response.error != null) {
-            Timber.d("get base fee error: ${response.error.message}")
-            return BigInteger.ZERO
+            // Same rationale as getMaxPriorityFeePerGas above (#5400).
+            throw NetworkException(
+                httpStatusCode = 0,
+                message = "base fee rpc error: ${response.error.message}",
+            )
         }
         return response.result?.baseFeePerGas.convertToBigIntegerOrZero()
     }
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/EvmApiFeeTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/EvmApiFeeTest.kt
@@ -1,0 +1,85 @@
+package com.vultisig.wallet.data.api
+
+import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.testutils.MockHttpClient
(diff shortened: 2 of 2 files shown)


rkokhatskyi:
@coderabbitai review

johnnyluo (approved):
LGTM

Issue #4865, opened 2026-06-11
refactor(swap): finish slimming SwapFormViewModel & encapsulate uiState


aminsato:
# [Fix] Finish slimming SwapFormViewModel & encapsulate uiState [Swap form]

## Problem
`SwapFormViewModel` (~979 lines) is the remaining item in the ViewModel-decomposition series (#4861 KeysignViewModel, #4862 DepositFormViewModel). Most logic was already extracted in the #4836 swap split (`SwapQuotePipeline`, `SwapQuoteManager`, `SwapGasCalculator`, `SwapTokenSelector`, `SwapTransactionBuilder`, `SwapValidator`, `SwapDiscountChecker`, `QuoteStateHolder`), but the ViewModel still exposes a public `MutableStateFlow` `uiState` (line ~106) that any collaborator can mutate, and `swap()` is ~120 lines (lines ~217–335) of input-gathering and validation guards. A spread of private `MutableStateFlow` fields for selection/fee state must also be kept mutually consistent by hand.

## Expected Behavior
The ViewModel ends up a thin state holder: `uiState` is exposed as a read-only `StateFlow` backed by a private `_uiState` (per the project's state-management convention, closing the same encapsulation gap flagged in #4861), and `swap()` is reduced to coordination — input collection/validation extracted into a helper that returns a validated-inputs object or surfaces an error, removing the doubled try/catch + `isLoadingNextScreen` reset paths before delegating to `swapTransactionBuilder.build(...)`.

## Steps to Reproduce
1. Open `SwapFormViewModel` (~979 lines) in the vultisig-android repository.
2. Observe `val uiState = MutableStateFlow(SwapFormUiModel())` at line ~106 — public and mutable, so collaborators can mutate UI state directly.
3. Observe `swap()` at lines ~217–335 — ~120 lines of vault/src/dst/gas/amount/quote null-checks and preflight before delegating to `swapTransactionBuilder.build(...)`, with doubled try/catch and `isLoadingNextScreen` reset paths.
4. Observe fragmented private `MutableStateFlow` fields (`selectedSrc`/`selectedDst`/`selectedSrcId`/`selectedDstId`, `gasFee`, `gasFeeChain`, `estimatedNetworkFeeTokenValue`/`estimatedNetworkFeeFiatValue`, `addresses`, `refreshQuoteState`) that must be kept mutually consistent.

### Must Do
[...]

Issue #4862, opened 2026-06-11
refactor(deposit): finish splitting DepositFormViewModel into per-option strategy files


aminsato:
# [Fix] Finish splitting DepositFormViewModel into per-option DepositSubmitStrategy files [Deposit form]

## Problem
`DepositFormViewModel` is a ~2966-line god object with 82 functions and ~24 injected dependencies. It owns every deposit option's transaction-building logic, all field validation, address/gas/balance loading, and UI state — the same class of problem as #4861 (KeysignViewModel). The strategy refactor is already started (`BondStrategy`, `UnbondStrategy`, `LeaveStrategy` are extracted and wired through the `depositStrategies` map), but ~13 remaining options still delegate to inline builders inside the ViewModel (lines ~1554–2551), which make up the bulk of the file.

## Expected Behavior
Each deposit option's transaction builder lives in its own `DepositSubmitStrategy` file (same shape as `BondStrategy`/`UnbondStrategy`/`LeaveStrategy`), wired through the existing `depositStrategies` map. Validation, loading/preflight, and gas/fee/amount helpers are extracted into dedicated unit-testable units. The ViewModel becomes a thin coordinator focused on state + coordination, with strategy-specific mutable state scoped to the strategy that uses it.

## Steps to Reproduce
1. Open `DepositFormViewModel` in the vultisig/vultisig-android codebase.
2. Observe the ~2966-line file containing ~13 inline `create*Transaction` builders, ~20 `validate*` / `*ErrorOrNull` functions, loading/preflight logic, and gas/fee helpers all in one class.
3. Compare with the already-extracted `BondStrategy`/`UnbondStrategy`/`LeaveStrategy` files to see the intended target pattern.

### Must Do
- [ ] Migrate each remaining inline builder into its own per-option `DepositSubmitStrategy` file, following the `BondStrategy`/`UnbondStrategy`/`LeaveStrategy` shape and keeping the `depositStrategies` map wiring in the ViewModel: `createAddCacaoPoolTransaction`, `createCustomTransaction`, `createStakeTransaction`, `createUnstakeTransaction`, `createTransferIbcTx`, `createSwitchTx`, `createMergeTx`, `createUnMergeTx`, `createRemoveCacaoPoolTransaction`, `createAddLiquidityTransaction`, `createRemoveLiquidityTransaction`, `createSecuredAssetTransaction`, `createWithdrawSecuredAssetTransaction` (plus `createTonDepositTransaction`).
[...]

PR #5423, opened 2026-07-27
fix(defi): surface RUJI's bonded and auto-compounding positions as separate cards

### LINKED ISSUE #5419: [Fix] Staked RUJI shows 0 in DeFi tab — bonded and sRUJI positions collapsed into one card
## Observed

Two users independently reported in Discord (2026-07-24) that their staked RUJI reads zero:

- `#💻-general-vs` — "Is anybody experiencing their staked ruji disappeared from the Vultisig UI?"
- `#bug-reports-android` — "my DeFi account shows zero, although I have Ruji staked - i feel this is a repeating issue"

## Why: RUJI has two positions, Android renders one card

Rujira's RUJI staking pool gives every account **two independent positions**, and an account can hold both at once:

| | **Bonded** ("Standard") | **Auto-compounding** |
|---|---|---|
| Rewards | USDC, claimed manually (`account.claim`) | USDC auto-buys more RUJI into the position |
| Receipt token | none — contract state only | **sRUJI**, denom `x/staking-x/ruji` |
| Stake | `{"account":{"bond":{}}}` funded with `x/ruji` | `{"liquid":{"bond":{}}}` funded with `x/ruji` |
| Unstake | `{"account":{"withdraw":{"amount":"…"}}}`, no funds | `{"liquid":{"unbond":{}}}` funded with sRUJI **shares** |
| Rujira GraphQL | `stakingV2.bonded.amount` | `stakingV2.liquidSize.amount` |
| APR / claimable USDC | belongs here | none — revenue compounds into the amount |

Android collapses both into a single `StakingDetails` row and picks one side.

## Defects

**1. Bonded-only holders see `0`.** `data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt:453-459`

```kotlin
val bondedAmount = stake?.bonded?.amount?.toBigIntegerOrNull() ?: BigInteger.ZERO
// The on-chain x/staking-x/ruji receipt is the source of truth ...
// Prefer the receipt (a successful zero stays zero), falling back to `bonded`
// only when the balance read failed. Mirrors vultisig-windows #4337.
val stakeAmount = readRujiReceiptBalance(address) ?: bondedAmount
```

`readRujiReceiptBalance` deliberately returns a genuine `BigInteger.ZERO` when the wallet holds no sRUJI receipt. For a bonded-only staker — the default "Standard" pool — that successful zero wins and `bondedAmount` is discarded. `RujiStakingService.getStakingDetailsFromNetwork` (`data/src/main/kotlin/com/vultisig/wallet/data/blockchain/thorchain/RujiStakingService.kt:70-80`) emits that single value as the position, so the DeFi tab shows 0 for a funded account.

The cited reference, vultisig-windows #4337, was the same wrong-shaped fix; Windows has since moved past it.

**2. Auto-compounding amount is a raw share count, not a value.** The GraphQL query (`ThorChainApi.kt:742-772`) never requests `liquidSize`, and `StakingV2` (`data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/RujiGraphQLResponse.kt:31-36`) has no field for it. The sRUJI receipt balance is displayed verbatim. sRUJI's share price is above 1 RUJI and rising, so compounders are understated by roughly the accrued yield.

**3. No `liquid.unbond` path for RUJI at all.** `ThorchainFunctions.unstakeRUJI` (`data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/ThorchainFunctions.kt:27-41`) only emits `account.withdraw`. `ExecMsg.liquidBond()` / `liquidUnbond()` exist but are reachable only through `stakeTcyCompound` / `unStakeTcyCompound`, hardcoded to `DENOM_STAKING_TCY` and `STAKING_TCY_COMPOUND_CONTRACT`. On the stake side, `DeFiNavActions.STAKE_RUJI` always routes to `stakeRUJI` → `accountBond` (`app/src/main/java/com/vultisig/wallet/ui/models/send/submit/StakeStrategy.kt:121-122, 208-210`); the `isAutocompound` toggle is TCY-only.

Net effect: an auto-compound-only holder's Unstake fires `account.withdraw` against an empty bonded position, and there is no way to enter or exit the auto-compounding position from Android.

## Repro

Bonded cohort — stake RUJI via the Standard pool (`account.bond`), hold no sRUJI, open DeFi → THORChain → Staking. Position shows 0.

Auto-compound cohort — hold sRUJI only. Amount renders as the share count rather than its RUJI value, and Unstake builds an `account.withdraw` against a bonded position that does not exist.

## Definition of done

Both positions surface as **separate cards**, each independently visible and transactable:

- **RUJI card** — `stakingV2.bonded.amount`, carries the APR and claimable USDC. Stake `account.bond`, unstake `account.withdraw`.
- **sRUJI card** — `stakingV2.liquidSize.amount` (already priced in RUJI; do not display the raw share count). Stake `liquid.bond` funded with `x/ruji`, unstake `liquid.unbond` funded with sRUJI shares.
- A holder of one, the other, or both sees exactly what they hold. Neither cohort's fix re-breaks the other.
- Add the `x/staking-x/ruji` receipt coin to `Coins.kt` following the existing sTCY (`Coins.kt:2943`) and ybRUNE (`Coins.kt:2997`) precedent — the compound-receipt-as-its-own-coin pattern is already established in this codebase, just never applied to sRUJI.

## Reference implementations

- **iOS** — vultisig/vultisig-ios#4865 (fixes vultisig/vultisig-ios#4862). Emits both positions, models the compound side as its own `.compound` coin, adds the `liquid.bond`/`liquid.unbond` builders.
- **Windows** — already correct. `core/ui/defi/chain/queries/services/thorchainStake/rujiStakeService.ts` reads `bonded` and `liquidSize` separately and converts to shares only when building the `liquid.unbond` tx.

Android is the last platform on the broken shape.

aminsato:
Closes #5419

## RUJI has two positions; Android rendered one card

Rujira's RUJI staking pool gives every account **two independent positions**, and an account can hold both at once:

|  | **Bonded** ("Standard") | **Auto-compounding** |
|---|---|---|
| Rewards | USDC, claimed manually (`account.claim`) | USDC auto-buys more RUJI into the position |
| Receipt token | none — contract state only | **sRUJI**, denom `x/staking-x/ruji` |
| Stake | `{"account":{"bond":{}}}` funded with `x/ruji` | `{"liquid":{"bond":{}}}` funded with `x/ruji` |
| Unstake | `{"account":{"withdraw":{"amount":"…"}}}`, no funds | `{"liquid":{"unbond":{}}}` funded with sRUJI **shares** |
| GraphQL | `stakingV2.bonded.amount` | `stakingV2.liquidSize.amount` |
| APR / claimable USDC | belongs here | none — revenue compounds into the amount |

### The regression, and why repointing the check would not fix it

This is the mirror image of #5258. That fix (PR #5261, mine) made the card **prefer** the on-chain `x/staking-x/ruji` receipt, so a successful read — *including a genuine zero* — won. But the Standard pool has no receipt token at all, so a zero receipt is the normal state for a bonded staker: their balance was discarded and the card read `0`. Simply repointing the check back at `bonded` would re-break the auto-compounders #5258 was filed for.

So this stops picking one and reports both. iOS went through exactly the same two-step regression (#4794 → #4797 → #4862 → **#4865, merged**), and vultisig-windows' `rujiStakeService.ts` already reads `bonded` and `liquidSize` separately. Android was the last platform on the broken shape.

### Two further defects, both live

**The auto-compounding amount was a raw share count.** `liquidSize` was never requested. sRUJI's share price is above 1 and rises as revenue compounds, so compounders were understated by roughly the accrued yield. The card now shows `liquidSize` — the receipt priced in RUJI, i.e. what Rujira itself displays.

**There was no `liquid.unbond` path for RUJI at all.** `ExecMsg.liquidBond()` / `liquidUnbond()` existed but were reachable only via `stakeTcyCompound` / `unStakeTcyCompound`, hardcoded to TCY's denom and contract. An auto-compound-only holder's Unstake fired `account.withdraw` against a bonded position that does not exist.

## Live verification before any code

`liquidSize` and `liquidShares` on the endpoint Android already calls (`api.vultisig.com/ruji/api/graphql`), for an account holding both positions:

```
bonded:       1638238990000   (16382.38990000 RUJI)
liquidShares: 1385594365632   == the on-chain x/staking-x/ruji balance
liquidSize:   1406899113878   (14068.99113878 RUJI)
```

`liquidShares` equals the bank balance to the base unit, and `bonded` moves independently of both — it is contract state with no receipt token, so it cannot be derived from any balance. `liquidSize / liquidShares` is the pool share price.

## The fix

- **RUJI card** — `bonded`, carries the APR and the claimable USDC. Stake `account.bond`, unstake `account.withdraw`. Unchanged.
- **sRUJI card** — `liquidSize`, stat-free. Stake `liquid.bond` funded with `x/ruji`, unstake `liquid.unbond` funded with receipt shares.
- Both are RUJI-denominated and price off RUJI, so the tab's RUJI total is their sum.

### Decisions worth reviewing

**Shares come from the same response as the size they back**, so the conversion ratio is self-consistent; the on-chain `by_denom` read is kept only as a fallback for a response missing `liquidShares` (the two are equal by construction). That fallback is gated on there being a position to size at all, so an account with no auto-compounding balance — including every account holding nothing but bonded RUJI — never pays for the request.

**Amounts fail closed.** A present-but-unparseable `bonded` or `liquidSize` now throws instead of coercing to zero — coercing is precisely the defect this issue is about, and the service already falls back to its cached position. An *absent* position stays a genuine zero and does not throw.

**So does the share count.** The on-chain read reports a failed lookup and an empty balance differently, and the caller preserves that: a live position whose share count cannot be read throws rather than reporting zero shares, which the unbond would otherwise read as insufficient balance and silently refuse. The unstake form translates that failure into a localized error instead of surfacing the API's status text.

**The unstake form is denominated in RUJI**, so its MAX matches the card the user tapped, and `UnstakeStrategy` converts to shares at submit. Redeeming the whole position sends the exact share balance rather than a rounded conversion, so no dust is stranded; rounding is downward everywhere else, so the redemption cannot exceed holdings even if the share price moves between the form loading and the submit.

**sRUJI stays out of the Wallet tab.** `DEFI_ONLY_THORCHAIN_DENOMS` is unchanged and the coin is not in `Coins.ThorChain.all`, so the receipt is not discoverable as a spendable token — matching windows' `isRujiStakingReceiptCoin` filter, and avoiding the "send it away and silently empty the staked position" hazard iOS still carries. Since it is therefore absent from both address flows, the unstake form synthesizes its account from the cached position, the same way `loadRewardsAccount` already does for the USDC rewards coin.

**No migration is needed.** A single RUJI position tile gates both cards, so existing stakers keep their position without an opt-in. iOS needed a one-shot migration because its cards are individually selectable; windows uses the same single-tile gating this does.

## Testing

`:app:testDebugUnitTest` and `:data:testDebugUnitTest` green, `:app:lintDebug` `BUILD SUCCESSFUL`, ktfmt applied.

The six `ThorChainApiImplTest` cases that pinned the receipt-preferring behaviour are rewritten, and ~20 tests added across the new surface:

- bonded-only, compound-only, and both-held cohorts, each asserted not to suppress the other
- the auto-compounding amount is `liquidSize`, not the share count
- fail-closed on unparseable `bonded` / `liquidSize`; genuine zero for an absent position
- `liquidShares` falling back to the on-chain receipt, failing closed when that read also fails, and skipped entirely for a bonded-only position
- a failed position read at unbond submit surfacing a translated error rather than the API's status text
- the RUJI position is selected by symbol, not by `stakingV2[0]`
- `liquid.bond` / `liquid.unbond` payload shape, denoms, and dust guards
- the share conversion: partial redemption at the pool ratio, and 100% sending the exact share balance
- the empty-position guard refusing to build a redemption
- the synthesized sRUJI account and its zero-ceiling fallback

## Not verified

**On-device mainnet broadcast of `liquid.bond` and `liquid.unbond` has not been performed.** These are two new money-moving execute paths that have never been broadcast from Android. The message shape is verified against the deployed contract — `rujira-staking/liquid.unbond` is present in the bytecode of this exact contract, and iOS broadcasts the identical `rujira_rs::interfaces::staking` interface against the bRUNE pool on mainnet — but that is not a substitute for a real signature. Worth gating merge on it.

## Summary by CodeRabbit

* **New Features**
  * Added staking and unstaking support for auto-compounding sRUJI (“liquid bond”/“liquid unbond”), including correct action routing and token selection.
  * RUJI and sRUJI now display as separate staking cards, with the RUJI tab total reflecting both positions.
  * Unstake now uses live sRUJI position details to determine redeemable shares.
* **Bug Fixes**
  * Improved RUJI/sRUJI aggregation, rewards/valuation display, and default/placeholder position behavior.
  * More resilient handling of missing accounts and incomplete staking data.
* **Tests**
  * Expanded unit and UI coverage for the new RUJI/sRUJI flows and transaction payloads.


aminsato:
@coderabbitai full review

data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
```diff
@@ -446,15 +446,24 @@ constructor(
...
+        // compounds.
+        val bondedAmount = stake?.bonded?.amount.toStakeAmountOrThrow(BONDED_FIELD)
+        val autoCompoundAmount = stake?.liquidSize?.amount.toStakeAmountOrThrow(LIQUID_SIZE_FIELD)
+        // Shares come from the same response as the size they back, so their ratio is consistent;
+        // the on-chain receipt (identical by construction) only covers a response missing the
+        // field.
+        val autoCompoundShares =
+            stake?.liquidShares?.amount?.toBigIntegerOrNull()
+                ?: readRujiReceiptBalance(address)
+                ?: BigInteger.ZERO
```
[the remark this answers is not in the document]
aminsato: Fixed — both halves of this were real.

The receipt read is now gated on there actually being an auto-compounding position to size, and a failed read on a live position propagates instead of collapsing to zero:

```kotlin
val autoCompoundShares =
    if (autoCompoundAmount <= BigInteger.ZERO) {
        BigInteger.ZERO
    } else {
        stake?.liquidShares?.amount?.toBigIntegerOrNull()
            ?: readRujiReceiptBalance(address)
            ?: error("RUJI auto-compounding share count is unavailable")
    }
```

Gating on `autoCompoundAmount` rather than on `stake != null` covers one case the suggested diff does not: a **bonded-only** staker has a non-null `stake` but no liquid position, so `stake?.let { … }` would still fire the bank read and — with the added `error(…)` — would let a transient balance failure sink a perfectly good bonded amount into the cache fallback. Size zero implies shares zero, so skipping the read there is safe and it also drops the round trip for every account with no RUJI position at all, which was your second point.

Two `ThorChainApiImplTest` cases added: `liquidShares` absent + a 500 on the bank read now throws, and a bonded-only position with the same failing bank read still reports its bonded amount with zero shares.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/UnstakeStrategy.kt
```diff
@@ -236,6 +249,84 @@ internal class UnstakeStrategy(
...
+        val stakeBalances =
+            withContext(Dispatchers.IO) { thorChainApi.getRujiStakeBalance(srcAddress) }
+        val positionValue = stakeBalances.autoCompoundAmount
+        val heldShares = stakeBalances.autoCompoundShares
+
+        if (positionValue <= BigInteger.ZERO || heldShares <= BigInteger.ZERO) {
+            throw InvalidTransactionDataException(
+                UiText.StringResource(R.string.send_error_insufficient_balance)
+            )
+        }
```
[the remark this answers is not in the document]
aminsato: Fixed. The fetch is wrapped and its failure mapped to `InvalidTransactionDataException(R.string.dialog_default_error_body)`, with the cause logged rather than dropped, and `CancellationException` rethrown so a cancelled submit isn't reported as an error:

```kotlin
val stakeBalances =
    withContext(Dispatchers.IO) { runCatching { thorChainApi.getRujiStakeBalance(srcAddress) } }
        .getOrElse { error ->
            if (error is CancellationException) throw error
            Timber.e(error, "Failed to read the RUJI staking position for the unbond")
            throw InvalidTransactionDataException(
                UiText.StringResource(R.string.dialog_default_error_body)
            )
        }
```

This matters more now than when you first raised it: the sibling fix on `ThorChainApi` makes an unreadable share count throw instead of reporting zero, so this path is the one that renders that failure. Regression test added — `submit UNSTAKE_SRUJI translates a failed position read instead of leaking its message` asserts the string resource and that no deposit is persisted.

aminsato:
Nitpicks — four of the five applied, one skipped with a reason.

**Timber format specifiers** — applied. All five call sites in `RujiStakingService` and `ThorchainDeFiBalanceService` now pass `%d`/`%s` arguments instead of interpolating.

**`ThorchainFunctionsTest` negative amount** — applied. `stakeRujiCompound` now asserts `BigInteger.valueOf(-1)` throws alongside the zero case.

**`ThorchainFunctionsTest` `!!`** — applied, but the premise was off: `payload.coins` is `List<CosmosCoin?>`, so `single()` returns a *nullable* — it doesn't remove the `!!` on its own. Bound it as `requireNotNull(payload.coins.single())` instead, which also drops the now-redundant `assertEquals(1, payload.coins.size)` since `single()` enforces the count and gives a better failure message than `!!` would.

**`AccountsLoader` find-or-bail** — applied. Extracted `List<Account>.findSourceOrPublishEmpty(tokenId, generation, missingReason)`, collapsing four call sites across `publishRewards`, `publishCircleUsdc` and `publishAutoCompoundRuji`. `missingReason` is optional deliberately: `publishRewards` runs against a pre-hydration snapshot where a missing RUNE/RUJI account is expected, and logging it at error level there would flood the log on every load — the same concern the Circle USDC path already documents for its `Timber.w`.

**Route `UnstakeStrategy` through the shared staking service** — skipped. `RujiStakingService` returns `List<StakingDetails>`, which carries `stakeAmount` but not `autoCompoundShares`, and shares are precisely what this call site needs to size the `liquid.unbond`. Plumbing shares through `StakingDetails` would mean carrying them into the Room entity and the mapper, and the service's first emission is the *cached* one — sizing a redemption from a stale share count is exactly the failure the direct read avoids. The direct call is deliberate here; error handling is now centralized at the call site instead, per the comment above.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/UnstakeStrategy.kt
```diff
@@ -236,6 +250,96 @@ internal class UnstakeStrategy(
...
+
+        val shares =
+            if (tokenAmountInt >= positionValue) {
```
rkokhatskyi: The MAX ceiling comes from `AccountsLoader`'s cached staking details while this share conversion uses a fresh `thorChainApi.getRujiStakeBalance` call, mixing two independently sourced values. When the live position sits below the stale ceiling a partial typed amount can trigger full redemption while the Verify screen still shows the original amount

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/UnstakeStrategy.kt
```diff
@@ -129,6 +132,17 @@ internal class UnstakeStrategy(
                                     chain = chain,
                                 )
 
+                            DeFiNavActions.UNSTAKE_SRUJI ->
```
rkokhatskyi, line 146: This new UNSTAKE_SRUJI branch reuses the insufficient balance guard at `UnstakeStrategy.kt:113`, which compares the typed amount against a stake value captured once at form load and never refreshed, while the actual redemption below fetches the live position fresh. A valid larger amount, if the position grew meanwhile, gets wrongly rejected as insufficient balance

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/thorchain/RujiStakingService.kt
```diff
@@ -21,16 +21,23 @@ constructor(
...
                 val cachedDetails =
-                    stakingDetailsRepository.getStakingDetailsByCoindId(
-                        vaultId,
-                        Coins.ThorChain.RUJI.id,
-                    )
-                if (cachedDetails != null) {
+                    stakingDetailsRepository.getStakingDetails(vaultId).filter {
+                        it.coin.id in RUJI_POSITION_COIN_IDS
+                    }
+                if (cachedDetails.isNotEmpty()) {
```
rkokhatskyi, line 33: `getStakingDetails` emits the persisted cache with no check that it matches the new dual position schema. A pre upgrade row, where old code stored either a false zero for bonded only stakers or a raw sRUJI share count, renders onto the new bonded RUJI card until the next successful fetch overwrites it

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1169,8 +1194,22 @@ constructor(
         private const val RUJI_REWARDS_SYMBOL = "USDC"
         private const val POSITION_DISPLAY_SCALE = 4
 
+        // The bonded and auto-compounding RUJI positions render as two independent cards.
+        private val RUJI_POSITION_COIN_IDS =
```
rkokhatskyi: RUJI_POSITION_COIN_IDS is declared twice with identical contents, once here and again in `RujiStakingService.kt:114`, with no shared source even though app depends on data. Future edits to the RUJI and sRUJI pair would need both updated or the loading gate and cache filter silently diverge

app/src/main/java/com/vultisig/wallet/ui/models/send/TokenPreselectionService.kt
```diff
@@ -146,6 +146,11 @@ internal class TokenPreselectionService(
             DeFiNavActions.STAKE_RUJI,
             DeFiNavActions.UNSTAKE_RUJI -> Coins.ThorChain.RUJI
 
+            // The auto-compounding position is funded with RUJI and redeemed out of the sRUJI
+            // receipt, so the two directions start from different tokens.
+            DeFiNavActions.STAKE_SRUJI -> Coins.ThorChain.RUJI
+            DeFiNavActions.UNSTAKE_SRUJI -> Coins.ThorChain.sRUJI
```
rkokhatskyi, line 155: The empty accounts guard at `TokenPreselectionService.kt:121` only nulls WITHDRAW_RUJI and WITHDRAW_USDC_CIRCLE, so when the RUNE account is missing this UNSTAKE_SRUJI branch still selects the static sRUJI template. Submission still fails safely via AccountValidator, but the Asset section wrongly shows complete with a blank balance instead of surfacing the missing prerequisite

data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
```diff
@@ -463,44 +479,55 @@ constructor(
...
     }
 
+    /**
+     * Parses a staking position amount, failing closed on a malformed value.
+     *
+     * An absent amount means the account holds no such position, which is a genuine zero. A
+     * *present* but unparseable one is a partial response: coercing it to zero would erase a live
+     * position from the UI, so it throws and lets the caller keep its cached position instead.
+     */
+    private fun String?.toStakeAmountOrThrow(field: String): BigInteger {
```
rkokhatskyi: The Rujira GraphQL schema marks bonded and liquidSize amounts as required, so a missing value only signals a degraded response, not an empty position. `toStakeAmountOrThrow` treats a missing amount as zero and persists it, silently zeroing and disabling Unstake for a live RUJI or sRUJI holder

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/thorchain/RujiStakingService.kt
```diff
@@ -39,13 +46,15 @@ constructor(
...
                         Timber.e(
                             e,
-                            "RujiStakingService: Error fetching RUJI staking details for vault $vaultId",
+                            "RujiStakingService: Error fetching RUJI staking details for vault %s",
+                            vaultId,
                         )
 
-                        if (cachedDetails != null) {
-                            Timber.d("RujiStakingService: Using cached RUJI position due to error")
+                        if (cachedDetails.isNotEmpty()) {
```
rkokhatskyi, line 53: For a vault staked before this PR shipped, the cache holds only the legacy RUJI row, so a failed fresh fetch here emits that single row again instead of throwing. The sRUJI card stays stuck loading and the aggregate DeFi balance omits the compounding position until the next full reload

data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
```diff
@@ -446,15 +446,31 @@ constructor(
...
-        // (a successful zero stays zero), falling back to `bonded` only when the balance read
-        // failed. Mirrors vultisig-windows #4337.
-        val stakeAmount = readRujiReceiptBalance(address) ?: bondedAmount
+        // The bonded and auto-compounding positions are independent and are reported side by side:
+        // neither may substitute for the other. Preferring one (as this used to, reading the
+        // on-chain receipt over `bonded`) reports a funded account as empty whenever it holds only
+        // the other kind. `liquidSize` — not the raw receipt balance — is the auto-compounding
+        // amount, because the receipt is denominated in shares whose price rises as revenue
+        // compounds.
+        val bondedAmount = stake?.bonded?.amount.toStakeAmountOrThrow(BONDED_FIELD)
```
rkokhatskyi: `getRujiStakeBalance` parses bonded and auto compound amounts sequentially in one function, so a malformed value in either discards both, even a successfully parsed sibling amount. A vault whose RUJI positions were never cached before has no fallback at `RujiStakingService.kt:56`, so a funded position renders as an empty card that stops loading

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/UnstakeStrategy.kt
aminsato: Confirmed and fixed in `380d9831b`.

The ceiling came from the cached `liquidSize` `AccountsLoader` seeded the form with, so a position that had compounded past it made a valid amount look insufficient. That check is now skipped for `UNSTAKE_SRUJI` only — the live read below is the sole authority, and it clamps to the whole position, so it can never over-redeem. Covered by `submit UNSTAKE_SRUJI honours an amount the cached ceiling has fallen behind`.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/UnstakeStrategy.kt
aminsato: Confirmed and fixed in `380d9831b`, together with the ceiling check on the other side of the same gap.

The redemption is now clamped to the live position, and the transaction carries the clamped value rather than the typed one, so Verify shows the amount the contract will actually return instead of the stale larger figure. Full redemption still sends the exact share balance, so no dust is stranded. Covered by `submit UNSTAKE_SRUJI redeems and displays only what the live position still holds`.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/thorchain/RujiStakingService.kt
aminsato: Confirmed and fixed in `90bebbf4b`.

The cache is now trusted only when it holds both position rows. Every write persists the pair, so a single-row set can only have come from the old collapsed logic — treating it as no cache keeps a false zero or a raw share count off the bonded card, and the fresh fetch rewrites it. Self-heals on the first good response, so no migration. Covered by `ignores a pre-upgrade cache holding only the bonded row`.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/thorchain/RujiStakingService.kt
aminsato: Confirmed and fixed by the same change in `90bebbf4b`.

It was slightly worse than described: the flow returned *normally* after emitting the lone legacy row, so the tab's `.catch` never ran and the compounding card kept its placeholder spinning indefinitely, while `_totalValueRujiStake` summed only the bonded amount. With a partial cache treated as empty the failure now reaches the tab, which clears both cards. Covered by `surfaces the failure rather than a legacy half cache when the network read fails`.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed in `8d1ee5f3a` — `RUJI_POSITION_COIN_IDS` now lives once in `RujiStakingService.Companion` and the tab imports it.

app/src/main/java/com/vultisig/wallet/ui/models/send/TokenPreselectionService.kt
aminsato: Confirmed and fixed in `ffd7c152a` — `UNSTAKE_SRUJI` joins the WITHDRAW types in the empty-accounts branch. It belongs to the same group: its account is synthesized on top of the vault's RUNE account, so an empty load means that prerequisite is missing. Covered by `preSelect UNSTAKE_SRUJI - Loaded(empty) does not preselect static template coin`.

data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
aminsato: Confirmed and fixed in `a1ad7f166`. I introspected the live endpoint to check the premise:

```
StakingAccount.bonded       -> Balance!
StakingAccount.liquidSize   -> Balance!
StakingAccount.liquidShares -> Balance!
Balance.amount              -> Bigint!
```

So a missing amount inside a present node is a degraded response, exactly as you said. The parse now takes the node: only its absence is a genuine zero, and within a present one a missing amount fails closed like a malformed one. Two fixtures had to be corrected — they omitted `liquidSize` on a present position, which the schema does not allow. Covered by `throws rather than zeroing an amount a present position omits`.

data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
aminsato: Fixed the part that genuinely crosses sources, in `a1ad7f166`.

The share count is the one value that does not arrive with the amounts — it falls back to a separate bank query — so an outage there was discarding two perfectly good displayable positions. It is now reported as unknown rather than thrown, and the unbond (its only reader) fails closed on the null, so a live position is never redeemed against a guessed count.

I kept the two GraphQL amounts atomic. They arrive in one response whose schema marks them non-null, so a null in either means that response is degraded for both — and GraphQL null-propagation makes the partial case largely unreachable anyway: a non-null field that fails nulls out the whole `stakingV2` element and populates `errors`, which we already reject before parsing. Splitting them would mean partially-null balances rippling through the service, the cache and the unbond to buy robustness against a shape the server cannot send. Happy to do it if you would still rather have it. Covered by `reports unreadable shares as unknown without losing the amounts`.

aminsato:
@coderabbitai the two outside-diff items from your last round:

- **RUJI refresh should preserve claimable USDC rewards** — the divergence between the cached and refreshed paths was real, so both now build the bonded balance through one helper (`a4a601f70`). Worth noting there is no user-visible effect today: nothing reads `DeFiBalance.Balance.coinRewards`/`rewardsAmount`, and the card's claimable USDC comes from `StakingDetails.rewards` via the view model. The fix is to stop the two paths disagreeing about the same balance.
- **Seed a loading placeholder for `sRUJI`** — already done in `933c28cf0`, before this round: `loadDefaultStakingPositions()` carries an sRUJI card gated on the same RUJI tile, which is also what pins its slot in the list.

johnnyluo:
## Manual testing: unstaking sRUJI fails to generate the QR code

Tried this on-device — attempting to **unstake sRUJI** fails before it can produce the keysign QR code.

### Root cause

`Coins.ThorChain.sRUJI` (`data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt`) is a static template coin with `hexPublicKey = ""` — it's never discoverable as a wallet token, so it never goes through the normal flow that populates a real vault-derived public key.

For `UNSTAKE_SRUJI`, `AccountsLoader.publishAutoCompoundRuji` synthesizes the sRUJI account like this:

```kotlin
// app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt:311
val sRuji = Coins.ThorChain.sRUJI.copy(address = thorchainAccount.token.address)
```

Only `address` is copied from the real THORChain account — `hexPublicKey` is left at `""`. This synthesized coin becomes `selectedToken` (via `TokenPreselectionService`'s `UNSTAKE_SRUJI -> Coins.ThorChain.sRUJI` mapping), then `DepositTransaction.srcToken`, and ultimately `KeysignPayload.coin` for the signing request.

Building the pre-signing input calls:

```kotlin
// data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt:128
PublicKey(keysignPayload.coin.hexPublicKey.hexToByteArray(), PublicKeyType.SECP256K1)
```

An empty hex string decodes to a zero-length byte array, which WalletCore's `PublicKey` constructor rejects (it requires a valid 33/65-byte SECP256K1 key) — this is what aborts QR generation.

This explains why it's specific to **unstaking**: `STAKE_SRUJI`'s `selectedToken` is `Coins.ThorChain.RUJI` (a real, fully-loaded coin with a valid `hexPublicKey`), so staking never touches this path. Only `UNSTAKE_SRUJI` routes a synthesized, template-sourced coin into the signer.

**Suggested fix**: also copy `hexPublicKey = thorchainAccount.token.hexPublicKey` onto the synthesized `sRuji` coin at `AccountsLoader.kt:311`, alongside `address`.

---

## Requesting on-chain verification

The PR description already flags that `liquid.bond`/`liquid.unbond` haven't been broadcast on mainnet. Given the bug above turned up on the first manual attempt at one of the four flows, could you provide real broadcast tx links for all four before this merges?

- [ ] Stake RUJI (bonded)
- [ ] Stake sRUJI (auto-compounding / `liquid.bond`)
- [ ] Unstake RUJI (bonded)
- [ ] Unstake sRUJI (auto-compounding / `liquid.unbond`)

johnnyluo (changes requested):
please provide stake/unstake onchain tx link for both RUJI and sRUJI

aminsato:
Thanks for catching this — your diagnosis was exactly right, and I've fixed it in 203237fe7.

## The `hexPublicKey` fix

Confirmed the whole chain: `Coins.ThorChain.sRUJI` is a template with `hexPublicKey = ""` and lives in `Coins.defiOnly` (deliberately out of `ThorChain.all`), so it is never vault-derived. `publishAutoCompoundRuji` copied only `address`, and that coin rides `selectedToken` → `DepositTransaction.srcToken` → `KeysignPayload.coin` → `ThorChainHelper.kt:128`, where WalletCore rejects the zero-length key.

Applied your suggestion — the RUNE account already resolved there for the address and the gas fee supplies the key, since every THORChain token shares the chain's derived key:

```kotlin
val sRuji =
    Coins.ThorChain.sRUJI.copy(
        address = thorchainAccount.token.address,
        hexPublicKey = thorchainAccount.token.hexPublicKey,
    )
```

Worth recording why it was unique to sRUJI, since I checked the neighbours: every other DeFi token in this flow is in `Coins.all` and therefore vault-derived (`sTCY` included), `WITHDRAW_RUJI` re-resolves the real RUJI account at submit time, and only the ThorChain and TON helpers read `coin.hexPublicKey` at all — EVM doesn't, so the synthesized Circle USDC coin was never at risk. Regression assert added in `AccountsLoaderTest`; 322 tests in the `send` package are green.

## On-chain verification — all four flows

All `code: 0`, all from the same vault to the Rujira staking contract `thor13g83nn…rml5cr`.

| Flow | Tx | Height | Message |
|---|---|---|---|
| Stake RUJI (bonded) | [`3A89BC3C…1D06`](https://gateway.liquify.com/chain/thorchain_api/cosmos/tx/v1beta1/txs/3A89BC3C5A2EF0CE7979B9F0ED36FA4A393CF59FD83A73074C6B0AB46A051D06) | 27207817 | `{"account":{"bond":{}}}` — funds `6834300 x/ruji` |
| Stake sRUJI (`liquid.bond`) | [`82FFF572…7412`](https://gateway.liquify.com/chain/thorchain_api/cosmos/tx/v1beta1/txs/82FFF57204952539509A9369488E5A1F2A2B80B964021E3918FA435E72FC7412) | 27207779 | `{"liquid":{"bond":{}}}` — `9112401` → `8973220` shares |
| Unstake RUJI (bonded) | [`170160DC…2146`](https://gateway.liquify.com/chain/thorchain_api/cosmos/tx/v1beta1/txs/170160DCC93B85F0F3B5E7E18C22CC319434742D1E999D9009EC8204D6E42146) | 27207834 | `{"account":{"withdraw":{"amount":"3677508"}}}` |
| Unstake sRUJI (`liquid.unbond`) | [`EE770B19…D220`](https://gateway.liquify.com/chain/thorchain_api/cosmos/tx/v1beta1/txs/EE770B1937ECC83DC0F930270445848A5FE88AE2293A81DC19A8F4715182D220) | 27207617 | `{"liquid":{"unbond":{}}}` — `18476230` shares → `returned 18762759` |

**Heads-up on the links:** these are thornode REST URLs rather than runescan, because **runescan cannot render Rujira contract txs** — it shows type `Undefined`, `Dec 31 1969` (unix epoch 0) and a blank height, which reads like a failed transaction. Midgard indexes the hashes correctly (4 actions each, all `status: success`, right height and date); the explorer front end just has no mapping for the `contract` action type. Every flow here is a `MsgExecuteContract`, so runescan is misleading for all four.

### The ledger closes to the unit

Against live chain state after all four:

```
bonded    7875733 + 6834300 − 3677508 = 11032525   ✅
shares   18476232 + 8973220           = 27449452   ✅
x/ruji   36449604 − 9112401 − 6834300 + 3677508 = 24180411   ✅
```

Three details that are stronger evidence than a bare success code:

- **The share conversion is exact and floors correctly.** The unbond redeemed 18476230 of 36952462 shares and `returned` 18762759 — marginally *above* half the displayed position, because the share price ticked from 1.015506 to 1.015511 between the cached read and execution. That is precisely why `UNSTAKE_SRUJI` skips the stale ceiling and clamps to `min(typed, positionValue)`.
- **The percentage buttons floor rather than round.** 25% of 27337203 → `6834300` (`.75` dropped), 25% of 14710033 → `3677508` (`.25` dropped). Flooring is deliberate: a redemption can never exceed holdings.
- **The bonded unstake sized itself against 14710033 — the *post-bond* figure**, not the stale pre-bond 7875733. That is direct proof the position cache refreshes between transactions.

One piece of contract behaviour worth noting for reviewers: `account.bond` **auto-claims pending revenue** — 15578 `eth-usdc` was returned to the vault inside `3A89BC3C`, so `pendingRevenue` went to 0 and the later `account.withdraw` correctly reported `rewards: 0`. That is Rujira's contract, not the app. The `wasm-rujira-fin/trade` and `wasm-rujira-thorchain-swap/swap` events riding along in these txs are likewise the autocompounder converting its own USDC revenue into RUJI — contract-internal, not a swap the app fires.

## Separate pre-existing bug found while verifying

Both RUJI cards render `$0.00` fiat. That is **not** from this PR — `Coins.ThorChain.RUJI.priceProviderID = "ruji"`, but CoinGecko's id for the token is `rujira`. Verified against our own proxy:

```
GET api.vultisig.com/coingeicko/api/v3/simple/price?ids=ruji    → {}
GET api.vultisig.com/coingeicko/api/v3/simple/price?ids=rujira  → {"rujira":{"usd":0.191}}
```

Nothing recovers it downstream: the fallback `getPriceByContactAddress` fans `x/ruji` out to CoinGecko-contract and LI.FI (neither covers THORChain), and `fetchThorContractPrices` whitelists only `x/nami*`, `x/staking-tcy`, bRUNE and ybRUNE. So RUJI is unpriced app-wide, including the wallet balance — it just becomes visible here because these cards are new. iOS already has it right (`TokensStore.swift:2175/2185` uses `"rujira"` for both `ruji` and `sruji`).

It is a one-word fix in `data/`. Happy to fold it into this PR or file it separately — let me know which you prefer.

johnnyluo (approved):
LGTM

PR #5434, opened 2026-07-28
test(keysign): close the golden signing-fixture parity gap and make it structural

### LINKED ISSUE #5420: Golden signing-fixture corpus drift: 3 files missing vs iOS/TS core; hardcoded filenames in ChainHelpersTest hide new fixtures
## Summary

The golden signing-fixture corpus in `app/src/androidTest/assets/` has drifted out of sync with the equivalent corpora in vultisig-ios (`VultisigAppTests/TestData/`, 23 files) and the TS core (`vultisig-sdk` `packages/core/mpc/keysign/tests/fixtures/mobile/`, 23 files): Android has **20 of 23 files**. The structural cause is that `ChainHelpersTest.kt` loads fixtures through hardcoded filename constants, so a fixture added to the other corpora is invisible here until someone also edits the test class.

## Missing files (present in both iOS and the TS core)

- `cosmos-sdk-sign-amino.json` — SignAmino: ATOM multi-message, THORChain memo
- `cosmos-sdk-sign-direct.json` — SignDirect: THORChain secure-asset LTC, ATOM multi-send
- `solana-sign-data.json` — Solana sign-data case

These pin exactly the hand-rolled/alternate signing paths (Amino vs Direct, sign-data) where cross-platform divergence is most likely, and MPC keysign requires all co-signing platforms to derive byte-identical pre-image hashes — a payload class covered on two platforms but not the third can drift silently and only surface as a live keysign failure (see vultisig/vultisig-windows#4464 for this week's field example of that failure class).

## Suggested fix

1. Copy the 3 files from either sibling corpus and wire test methods for them.
2. Consider enumerating the assets directory (the way the TS runner `readdirSync`s its fixtures dir) instead of per-file constants, so file-set parity with the other repos becomes structural rather than manual.
3. Add an executed-case-count assertion (TS asserts its file count) so a silently-skipped file can't pass — the iOS suite is currently red-flagged for exactly that failure mode (vultisig/vultisig-ios#4931).

aminsato:
Closes #5420

## The issue's premise needs a correction

#5420 reports that Android is missing 3 golden-fixture **files** (`cosmos-sdk-sign-amino.json`, `cosmos-sdk-sign-direct.json`, `solana-sign-data.json`) and therefore 3 uncovered payload classes. That isn't what's happening. **Android keeps those cases inline in the per-chain files; iOS splits them into dedicated files.** The file count differs (20 vs 23); the coverage largely doesn't.

Case-level audit of both corpora, matching on name + expected hash:

| | |
|---|---|
| Android cases | 66 (now 67) |
| iOS cases | 65 |
| Cases in the 3 "missing" files | 5 — **4 already present in Android**, with byte-identical expected hashes |
| Genuinely absent | **1** — `SignAmino - THORChain transaction memo` |

`SignAmino - ATOM` was already in `cosmos.json`; both `SignDirect` cases in `cosmos.json` / `thorchain.json`; `Send Solana USDC` in `solana.json`. Copying the 3 files verbatim as the issue suggests would have produced **duplicate cases in two files each**, free to drift apart — worse than the status quo. So this PR adds the one case that was actually missing and then fixes the two structural problems the issue correctly identifies underneath the file-count framing.

## What changed

**1. `SignAmino - THORChain transaction memo` → `thorchain.json`** (67 cases total)

Placed next to the `SignDirect` THORChain case already living there, following this repo's inline-per-chain convention. Picked up automatically by `sendTHORchain`, which iterates every case in the file.

The expected hash is verified independently of WalletCore — sha256 over the canonical amino `StdSignDoc` (`chain_id=thorchain-1`, sequence 425, account 139521) reproduces `5792e34e…` exactly. The same oracle reproduces the already-green ATOM case (`235ed178…`), which is what validates the oracle.

**2. Canonical field spellings now parse** (`ChainHelperModels.kt`)

`SolanaSpecific` is `program_id` / `compute_limit` in commondata `blockchain_specific.proto` and in both sibling corpora; this corpus spelled them `has_program_id` / `priority_limit`. The Tron trigger payload used camelCase where the proto and iOS use snake_case.

Since the fixture reader sets `ignoreUnknownKeys = true`, a case copied verbatim from vultisig-ios would have had those fields **deserialize to their defaults while the test still passed** — silently changing the signed message with no failing assertion. Exactly the failure class this corpus exists to catch. Both spellings are now accepted via `@JsonNames`; no existing fixture was edited and no hash moved.

**3. File-set parity is now structural** (`fixtureCorpusMatchesDeclaredFileSet`)

Fixtures were reachable only through hardcoded filename constants, so a `.json` added to `androidTest/assets` stayed invisible until someone also edited the test class. The new test asserts the declared set and the assets directory agree **in both directions**, that no fixture parses to zero cases, and that the corpus totals a pinned `EXPECTED_CASE_COUNT` — mirroring the file-count assertion the TS runner makes over its own fixtures directory.

## Test plan

`ChainHelpersTest` on `Medium_Phone` (emulator-5554): **28 tests, 0 failures, 0 errors, 0 skipped.**

- `sendTHORchain` — pass. Now 11 cases including the new amino case; its expected hash reproduces on Android.
- `fixtureCorpusMatchesDeclaredFileSet` — pass at 67 cases across 20 files.
- `sendSolana`, `sendTronTest` — pass, confirming the `@JsonNames` aliases changed no existing fixture's parse.

The guard was checked negatively too — dropping a stray `zz_stray.json` into the assets dir fails it as intended:

```
java.lang.AssertionError: Fixture files present in androidTest/assets but not declared in
FIXTURE_FILES — declare each one and exercise it from a test, or the corpus grows untested
expected:<[]> but was:<[zz_stray.json]>
```

Stray file removed; suite re-run green.

```
./gradlew :app:connectedDebugAndroidTest \
  -Pandroid.testInstrumentationRunnerArguments.class=com.vultisig.wallet.data.blockchain.ChainHelpersTest
```

## Follow-ups found along the way (not in this PR)

- **vultisig-ios `mayaswap.json` is malformed JSON** — trailing comma before the closing `]` (line 142). Both Python and `JSONSerialization` reject it, so those 2 MAYA swap cases may not be loading in the iOS suite at all. Worth an issue upstream.
- **`./gradlew assembleDebugAndroidTest` fails on `main` independently of this PR** — `:data:mergeDebugAndroidTestJavaResource` hits a duplicate `META-INF/versions/9/OSGI-INF/MANIFEST.MF` between `bcprov-jdk18on:1.79` and `jspecify:1.0.0`; needs a `packaging { resources { excludes += … } }` block in `data/build.gradle.kts`. Scoping runs to `:app:` routes around it.
- **`Send Ray token` in `solana.json`** carries a stale `from_address_pub_key` whose value differs from the canonical `from_token_associated_address` the model actually reads. Harmless today, a trap for anyone who "fixes" the key name.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with legacy fixture JSON field name spellings during payload deserialization (snake_case and alternate legacy keys).

* **Tests**
  * Added a THORChain RUNE signing fixture covering a custom transaction memo.
  * Enhanced fixture validation to ensure the declared fixture set matches the available assets, verifies case counts, and confirms all fixture-to-test bindings are exercised.
  * Added a test to confirm alternate JSON field spellings deserialize correctly.

### DIFF
--- a/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelperModels.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelperModels.kt
@@ -151,10 +151,16 @@ data class TonSpecific(
 
 /**
  * `program_id` / `compute_limit` are the field names in `SolanaSpecific` (commondata
- * `blockchain_specific.proto`) and are what the iOS and TS-core fixture corpora use; this corpus
- * historically spelled them `has_program_id` / `priority_limit`. Both spellings are accepted so a
- * fixture can be copied verbatim between repos without a field silently deserializing to its
- * default — which would change the signed message while the test still passed.
+ * `blockchain_specific.proto`); this corpus historically spelled them `has_program_id` /
+ * `priority_limit`. Both spellings are accepted so a fixture can be copied verbatim between repos
+ * without a field silently deserializing to its default — which would change the signed message
+ * while the test still passed.
+ *
+ * Note this is a defensive accommodation, not proof of cross-platform correctness: iOS's own
+ * `KeysignPayloadCodable` only ever decodes `priority_limit`, so its `solana-sign-data.json`
+ * fixture — which spells the field `compute_limit` — silently loses that value on iOS too.
+ * Accepting `compute_limit` here means Android won't drop it the way iOS does, not that the value
+ * is verified identical across platforms.
  */
 @OptIn(ExperimentalSerializationApi::class)
 @Serializable
--- a/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelperModels.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelperModels.kt
@@ -1,6 +1,8 @@
 import com.vultisig.wallet.data.models.payload.SignDirect
+import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonNames
 import vultisig.keysign.v1.SignTon
 
 @Serializable
@@ -29,12 +31,14 @@ data class KeysignPayload(
     var triggerSmartContractPayload: TriggerSmartContractPayload? = null,
 )
 
+/** Snake-case aliases mirror `TronTriggerSmartContractPayload` in commondata and the iOS corpus. */
+@OptIn(ExperimentalSerializationApi::class)
 @Serializable
 data class TriggerSmartContractPayload(
-    val ownerAddress: String,
-    val contractAddress: String,
-    val callValue: String? = null,
-    val callDataValue: String? = null,
+    @JsonNames("owner_address") val ownerAddress: String,
+    @JsonNames("contract_address") val contractAddress: String,
+    @JsonNames("call_value") val callValue: String? = null,
+    @JsonNames("call_token_value") val callDataValue: String? = null,
     val data: String,
 )
 
@@ -145,14 +149,22 @@ data class TonSpecific(
     val bounceable: Boolean,
 )
 
+/**
+ * `program_id` / `compute_limit` are the field names in `SolanaSpecific` (commondata
+ * `blockchain_specific.proto`) and are what the iOS and TS-core fixture corpora use; this corpus
+ * historically spelled them `has_program_id` / `priority_limit`. Both spellings are accepted so a
+ * fixture can be copied verbatim between repos without a field silently deserializing to its
+ * default — which would change the signed message while the test still passed.
+ */
+@OptIn(ExperimentalSerializationApi::class)
 @Serializable
 data class SolanaSpecific(
     @SerialName("recent_block_hash") val recentBlockHash: String,
     @SerialName("priority_fee") val priorityFee: String,
-    @SerialName("has_program_id") val hasProgramId: Boolean? = false,
+    @SerialName("has_program_id") @JsonNames("program_id") val hasProgramId: Boolean? = false,
     @SerialName("from_token_associated_address") val fromAddressPubKey: String? = null,
     @SerialName("to_token_associated_address") val toAddressPubKey: String? = null,
-    @SerialName("priority_limit") val priorityLimit: String? = null,
+    @SerialName("priority_limit") @JsonNames("compute_limit") val priorityLimit: String? = null,
 )
 
 @Serializable
--- a/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
@@ -718,6 +718,53 @@ class ChainHelpersTest {
         }
     }
 
+    /**
+     * The golden corpus is shared with vultisig-ios (`VultisigAppTests/TestData`) and the TS core
+     * (`packages/core/mpc/keysign/tests/fixtures/mobile`). MPC keysign requires every co-signing
+     * platform to derive byte-identical pre-image hashes, so a payload class pinned on two
+     * platforms but not the third can drift silently and only surface as a live keysign failure.
+     *
+     * Fixtures are loaded through [FIXTURE_FILES] rather than the asset directory itself, so
+     * dropping a `.json` into `androidTest/assets` without wiring it to a helper would otherwise be
+     * invisible. This asserts the declared set and the directory agree in both directions, that
+     * every file holds at least one case, and that the corpus totals [EXPECTED_CASE_COUNT] — so a
+     * fixture that silently stops contributing cases can't pass. Mirrors the file-count assertion
+     * the TS runner makes over its own fixtures directory (issue #5420).
+     */
+    @Test
+    fun fixtureCorpusMatchesDeclaredFileSet() {
+        val assetFiles =
+            InstrumentationRegistry.getInstrumentation()
+                .context
+                .assets
+                .list("")
+                .orEmpty()
+                .filter { it.endsWith(".json") }
+                .toSet()
+
+        assertEquals(
+            "Fixture files present in androidTest/assets but not declared in FIXTURE_FILES — " +
+                "declare each one and exercise it from a test, or the corpus grows untested",
+            emptySet<String>(),
+            assetFiles - FIXTURE_FILES,
+        )
+        assertEquals(
+            "Fixture files declared in FIXTURE_FILES but absent from androidTest/assets",
+            emptySet<String>(),
+            FIXTURE_FILES - assetFiles,
+        )
+
+        val caseCountsByFile = FIXTURE_FILES.associateWith { loadTransactionData(it).size }
+        val emptyFiles = caseCountsByFile.filterValues { it == 0 }.keys
+        assertEquals("Fixture files that parsed to zero cases", emptySet<String>(), emptyFiles)
+        assertEquals(
+            "Golden fixture case count changed — update EXPECTED_CASE_COUNT deliberately, and " +
+                "keep the case in step with vultisig-ios and the TS core",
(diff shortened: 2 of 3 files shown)


app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
```diff
@@ -718,6 +718,53 @@ class ChainHelpersTest {
...
+        )
+
+        val caseCountsByFile = FIXTURE_FILES.associateWith { loadTransactionData(it).size }
+        val emptyFiles = caseCountsByFile.filterValues { it == 0 }.keys
+        assertEquals("Fixture files that parsed to zero cases", emptySet<String>(), emptyFiles)
+        assertEquals(
+            "Golden fixture case count changed — update EXPECTED_CASE_COUNT deliberately, and " +
+                "keep the case in step with vultisig-ios and the TS core",
+            EXPECTED_CASE_COUNT,
+            caseCountsByFile.values.sum(),
```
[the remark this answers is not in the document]
aminsato: Valid — fixed in b4342ee4d.

The guard counted cases *parsed*, not cases *signed*, so a fixture could be declared, parse cleanly, count toward `EXPECTED_CASE_COUNT`, and still have nothing hashing it. That's the same silent-skip the guard exists to prevent, one level in.

`FIXTURE_FILES` is now `FIXTURE_TESTS`, mapping each fixture to the test that signs it, and the guard asserts every bound name resolves to a real `@Test` on the class. Verified negatively — pointing `sui.json` at a nonexistent test fails as intended:

```
java.lang.AssertionError: Fixture files bound to a test that doesn't exist or isn't annotated
@Test — the fixture would be parsed but never signed
expected:<{}> but was:<{sui.json=sendSuiButNoSuchTest}>
```

I stopped short of the full executor-registry refactor. The 20 signing tests have genuinely heterogeneous shapes — `sendThorchainSwapTest` accumulates approve + swap hashes with a nonce increment, `sendMayaChainSwapTest` goes through `SigningHelper` with a `Vault`, `sendCardano` makes bespoke tamper assertions — so collapsing them behind one `(payload) -> List<String>` signature would flatten real differences for a marginal gain over the binding assertion. Happy to revisit if you'd still prefer it.

28/28 green on emulator.

app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelperModels.kt
```diff
@@ -145,14 +149,22 @@ data class TonSpecific(
     val bounceable: Boolean,
 )
 
+/**
+ * `program_id` / `compute_limit` are the field names in `SolanaSpecific` (commondata
+ * `blockchain_specific.proto`) and are what the iOS and TS-core fixture corpora use; this corpus
```
rkokhatskyi: This comment claims iOS's fixture corpora use `compute_limit`, but iOS's decoder only ever reads `priority_limit` in `KeysignPayloadCodable.swift:262`, and iOS's own `solana-sign-data.json` case spelled `compute_limit` silently loses that field there too.

app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelperModels.kt
```diff
@@ -145,14 +149,22 @@ data class TonSpecific(
...
 @Serializable
 data class SolanaSpecific(
     @SerialName("recent_block_hash") val recentBlockHash: String,
     @SerialName("priority_fee") val priorityFee: String,
-    @SerialName("has_program_id") val hasProgramId: Boolean? = false,
+    @SerialName("has_program_id") @JsonNames("program_id") val hasProgramId: Boolean? = false,
     @SerialName("from_token_associated_address") val fromAddressPubKey: String? = null,
     @SerialName("to_token_associated_address") val toAddressPubKey: String? = null,
-    @SerialName("priority_limit") val priorityLimit: String? = null,
+    @SerialName("priority_limit") @JsonNames("compute_limit") val priorityLimit: String? = null,
```
rkokhatskyi, line 173: None of the five new `@JsonNames` aliases (`owner_address`/`contract_address`/`call_value`/`call_token_value` on `TriggerSmartContractPayload`, `program_id`/`compute_limit` here) are exercised by any fixture case that reaches an asserted hash. The one case spelling `compute_limit` takes the `signSolana` raw transaction shortcut in `SolanaHelper.kt:283`, so a wrong alias would go undetected.

app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
```diff
@@ -718,6 +718,71 @@ class ChainHelpersTest {
...
+        )
+        assertEquals(
+            "Fixture files declared in FIXTURE_TESTS but absent from androidTest/assets",
+            emptySet<String>(),
+            declaredFiles - assetFiles,
+        )
+
+        // Parsing a fixture is not the same as signing it: without this, a file could be declared,
+        // parse cleanly, count toward the total, and still have no test deriving hashes from it.
+        val signingTests =
```
rkokhatskyi, line 765: `signingTests` here only checks that a `FIXTURE_TESTS` value names some `@Test` method, never that the method body loads that key's file. A future edit like `TERRA_JSON_FILE to "sendUTXO"` leaves this guard green while terra.json goes unsigned.

rkokhatskyi (approved):

app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelperModels.kt
johnnyluo: Confirmed — checked iOS's `KeysignPayloadCodable.swift`: `CodingKeys` only ever maps `priorityLimit` to `"priority_limit"`, and its own `solana-sign-data.json` fixture spells the field `compute_limit`, so iOS silently drops it too. The comment overclaimed. Fixed in b6f5bab06 — reworded to state accurately that accepting `compute_limit` here is a defensive accommodation, not proof the value is honored identically cross-platform.

app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelperModels.kt
johnnyluo: Confirmed — traced this. `solana.json`'s only `compute_limit` case (`Send Solana USDC`) carries `sign_solana.raw_transactions`, so `SolanaHelper.getPreSignedImageHash` takes the raw-tx shortcut and never touches `priorityLimit`. `tron.json`'s only `TriggerSmartContractPayload` case uses the camelCase spelling, not the snake_case alias. Fixed in b6f5bab06 by adding `jsonNamesAcceptBothFieldSpellings`, which decodes both models directly against the snake_case spellings and asserts the fields land correctly — cheaper than wiring new cases through to an asserted hash, and it actually exercises the aliases that were previously untested.

app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
johnnyluo: Confirmed — this was a real gap: the existence check only proved the bound name resolves to *some* `@Test`, not that its body loads that specific file, so `TERRA_JSON_FILE to "sendUTXO"` would have stayed green. Fixed in b6f5bab06: `loadTransactionData` now records its caller via stack trace into `loadedFilesByTestMethod`, and a new `@AfterClass` check (`verifyFixtureBindingsWereActuallyLoaded`) asserts every `FIXTURE_TESTS` binding was actually observed loading its file. It only flags a binding whose test *did* run but didn't touch the file, so running a single test in isolation still works. Verified negatively by rebinding `TERRA_JSON_FILE` to `"sendUTXO"` locally — fails with `expected:<{}> but was:<{terra.json=sendUTXO}>` as intended, then reverted. 29/29 green on two devices after the fix.

PR #5437, opened 2026-07-28
fix(review): ask for a store review only after a transaction succeeds

### LINKED ISSUE #5427: Add in-app review prompt after successful transaction (ratings velocity is our store-conversion bottleneck)
## Problem

Vultisig Android has **440 ratings** (Play Store) @ 5.0★ — healthier than iOS (62) but still thin, and nearly flat run-over-run. Ratings count + velocity gate both search rank and store-listing conversion on Play; before keyword/metadata work pays off, this is the ceiling.

## Proposal

Trigger the native in-app review flow (Google Play In-App Review API — `ReviewManager.requestReviewFlow()` → `launchReviewFlow()`) at the moment of highest satisfaction:

- **After a successful send/swap confirmation** — transaction confirmed on-chain, user on the success screen. Not at app open, never after an error.
- App-side throttling before requesting: ≥3 successful transactions AND ≥7 days since install AND not prompted in the last 120 days. (Play enforces its own quota and the flow may no-op silently — that's expected; never assume it displayed.)
- Native flow only — no custom "rate us" chrome, no incentives, never gate content on a rating (Play policy requires the standard API).

## Why now

ASO keyword/metadata iteration is underway on the marketing side; its conversion impact is capped by ratings volume. Single highest-leverage product change for organic Play growth right now.

## Acceptance

- In-app review flow requested on the success screen under the throttle rules.
- Analytics event logged when `launchReviewFlow` is invoked (so marketing can correlate ratings velocity to the trigger; the API intentionally does not report whether the user rated).

aminsato:
Closes #5427

## Summary

The issue asks for an in-app review prompt after a successful transaction. The app already had one — added in #883 — but it was wired to fire at the wrong moments and had no throttling, so the work is fixing *when* it asks rather than adding the call.

Three problems on `main`:

1. **It asked for a rating after failed transactions.** `KeysignState.KeysignFinished` is not a success signal — it carries a `TransactionStatus` that may be `Failed` or `Refunded`. A user whose swap was refunded got a rating prompt.
2. **It asked when the user copied their receive address.** Routine, repeatable, and not a satisfaction peak — and it burned Play's quota so the prompt was unavailable at the moment that actually earns a rating.
3. **No throttling of any kind.** Every completed keysign requested the flow.

## Changes

**Throttle** — new `InAppReviewRepository` gates on all three conditions from the issue: ≥3 successful transactions, ≥7 days since install, and ≥120 days since the last prompt. Install age is read from `PackageManager.firstInstallTime` rather than a stored timestamp, so users who installed before this gate existed count as long-standing instead of restarting the clock; an unreadable install date fails closed. The counter is incremented inside a single DataStore transaction so concurrent successes can't clobber each other.

The cooldown starts when the prompt is *requested*, not when the user rates. Play never reports whether the dialog appeared, so treating a request as spent is the only safe reading.

**Trigger** — `KeysignViewModel` now watches for the first terminal status where funds actually moved (`Broadcasted` / `Confirmed`) and emits an event the screen collects. Excluded: `Failed` and `Refunded`, `Signed` (PSBT co-signing never reaches a network), `Pending` and `StillConfirming` (no on-chain result yet), and dApp message signing (no transaction at all).

First-match-only matters — status polling re-emits `KeysignFinished` on every tick, so a confirmed transaction would otherwise ask on every poll and inflate the counter.

**Plumbing** — launching the flow needs an `Activity`, so it stays in the UI behind a small `InAppReviewEffect`. That also replaces an unchecked `context as Activity` cast with `LocalActivity`, and wires the co-signing device, which previously never saw the prompt at all.

## Not included

The issue's second acceptance criterion — an analytics event when `launchReviewFlow` is invoked — is **not** implemented. This repo has no analytics stack: no Firebase Analytics, Amplitude, Mixpanel, or event-logging abstraction of any kind (Firebase is present for push messaging only). Satisfying it means introducing first-party telemetry into the wallet, which is a product and privacy decision rather than a technical one. Happy to add it in a follow-up once that call is made.

## Test plan

- 15 new unit tests: 7 covering each throttle gate at its boundary, 8 covering the status matrix and the once-only guarantee.
- Full suite green: 3,996 tests, 0 failures.
- `:app:lintDebug`, `:data:lintDebug` and `assembleDebug` all pass.

Manual, on a device updated **over** an existing install (a clean install resets `firstInstallTime`, which makes the 7-day gate unreachable for a week):

- Copy a receive address repeatedly on a chain screen → snackbar only, no rating sheet.
- Complete a swap that gets refunded → refund reason on the done screen, no rating sheet.
- Complete three sends/swaps → the sheet can appear on the third, subject to Play's own quota.

Note that Play decides whether the dialog actually renders and no-ops silently when its quota is exhausted, so an absent dialog is not by itself evidence of a defect.

## Open

The prompt has not yet been observed firing on the **co-signing** device. Still being investigated — it may simply be that device's own counter not having reached three transactions, since each install throttles independently. The initiator path is unaffected either way.

## Summary by CodeRabbit

* **New Features**
  * Added in-app review prompts after keysigning finishes with eligible successful on-chain transaction outcomes.
  * Review requests are throttled using successful-transaction count, install age, and a cooldown window.
* **Bug Fixes**
  * Removed review prompting from address-copy actions in Chain Tokens.
  * Updated review behavior to be driven by a shared Compose effect instead of direct UI calls.
* **Tests**
  * Added comprehensive tests for review eligibility and throttling, plus updated ViewModel/repository wiring coverage.
* **Documentation**
  * N/A (not changed)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -204,6 +210,23 @@ sealed interface TransactionStatus {
     data class Refunded(val reason: UiText) : TransactionStatus
 }
 
+/**
+ * True only where the user's funds actually moved on-chain. [TransactionStatus.Signed] never
```
rkokhatskyi: `isSuccessful`'s doc claims `Broadcasted` means funds actually moved on chain, but `Broadcasted` (`KeysignViewModel.kt:176`) only promises submission to the network. That is weaker than the mempool inclusion `Pending` describes, yet `Pending` is excluded as unsuccessful

rkokhatskyi (approved):

PR #5438, opened 2026-07-28
fix(zksync): clamp EIP-1559 priority fee to max fee

### LINKED ISSUE #5401: [Fix] ZkSync EIP-1559 spec never clamps priority fee to max fee, unlike iOS
# [Fix] ZkSync EIP-1559 spec never clamps priority fee to max fee, unlike iOS

## Problem
Both `BlockChainSpecificRepository`'s ZkSync branch (the real send path) and `ZkFeeService` (the live preview path) build the EIP-1559 spec straight from `zks_estimateFee`'s raw response with no comparison between `maxFeePerGasWei` and `priorityFeeWei`. WalletCore's encoding performs no `priority <= max` validation either. Per protocol rules (go-ethereum's `ErrTipAboveFeeCap`), a tx with `priority > max` is invalid and rejected at broadcast. iOS explicitly clamps this (`BlockChainService.swift`) specifically because ZkSync's fee pair comes from an external RPC rather than being algorithmically derived.

## Expected Behavior
The priority fee should never exceed the max fee in the constructed spec — matching iOS.

## Steps to Reproduce
1. Send on ZkSync when `zks_estimateFee` returns a priority fee greater than the max fee (an observed real-world response shape, not purely theoretical — iOS's defensive clamp exists because of it).
2. The signed tx has `priority > max`.
3. Broadcast rejects at `eth_sendRawTransaction`, after a completed MPC ceremony, with no automatic retry.

## Solution
Clamp with `adjustedPriority = if (maxPriorityFeePerGas > maxFeePerGas) maxFeePerGas else maxPriorityFeePerGas` before constructing the EIP-1559 spec in both `BlockChainSpecificRepository`'s ZkSync branch and `ZkFeeService`, mirroring iOS exactly. Extract into one shared helper both paths call so they can't drift.

### Must Do
- [ ] Add the priority-fee-clamp in `BlockChainSpecificRepository`'s ZkSync branch
- [ ] Add the identical clamp in `ZkFeeService`
- [ ] Extract into one shared helper used by both

### Must NOT Do
- Don't change the clamp logic for any other chain
- Don't change ZkSync's zero-on-RPC-error handling (tracked separately as EVM-003)

## Acceptance Criteria
- [ ] A ZkSync fee estimate with `priority > max` is clamped before signing, on both the preview and real send paths
- [ ] New test asserts the clamp

aminsato:
## Description

zkSync is the one EVM chain whose EIP-1559 fee pair arrives verbatim from an external RPC (`zks_estimateFee`) rather than being derived locally, so an inverted pair (`priority > max`) is representable here in a way it is not on any other chain.

Nothing downstream catches it. `EthereumGasHelper` copies `priorityFeeWei` straight into WalletCore's `maxInclusionFeePerGas`, and WalletCore performs no `priority <= max` validation while encoding. An inverted pair would therefore be rejected only at `eth_sendRawTransaction` (geth's `ErrTipAboveFeeCap`) — **after the MPC ceremony has already completed**, with no automatic retry.

This adds the clamp in a single shared helper called by both zkSync fee paths so they cannot drift:

```kotlin
internal fun ZkGasFee.clampedPriorityFee(): BigInteger =
    maxPriorityFeePerGas.coerceAtMost(maxFeePerGas)
```

- `BlockChainSpecificRepository` — the real send path
- `ZkFeeService` — the live fee-preview path

### Honest scoping

This is a **defensive parity fix, not a live-bug fix.** In normal operation `zks_estimateFee` returns `maxPriorityFeePerGas = 0`, so the clamp is a no-op in production (confirmed on mainnet below). It is worth landing because Android was the only platform without a guard, and because the failure mode it prevents is unusually expensive — a rejection that surfaces only after a completed keysign.

Closes #5401

## Which feature is affected?
- [x] Sending - Please attach a tx link here
- [ ] Swap

Mainnet tx (zkSync Era): https://explorer.zksync.io/tx/0x12887b165c01110cccc2d24102e3332660883a38661413a172e10c278ea69153

## Parity

Android was the only platform lacking a guard on this path. All three siblings already have one:

- **iOS:** `n/a` — already fixed. `BlockChainService.swift`, `.zksync` branch:
  ```swift
  // Ensure priority fee does not exceed max fee
  let adjustedPriority = maxPriorityFeePerGas > maxFeePerGas ? maxFeePerGas : maxPriorityFeePerGas
  ```
  This PR ports that behaviour to Android.
- **Windows + extension:** `n/a` — covered via the SDK resolver below; no separate zkSync fee path of their own.
- **SDK:** `n/a` — already guarded, by construction rather than by clamp. `getEvmFeeQuote.ts` stores `baseFeePerGas = bigIntMax(0n, maxFeePerGas - maxPriorityFeePerGas)` and rebuilds `maxFeePerGas` downstream as `baseFeePerGas + clamp(priority)`, so `priority <= maxFee` holds structurally. Its comment names this case explicitly — *"A compromised RPC returning a malformed tuple (maxFee < priority, the exact clamp-attack case)"* — citing SDK2-01 / #1078.

The SDK reaching the same conclusion independently, on the same `zks_estimateFee` tuple, is the strongest available corroboration that the concern is real.

### Known gap, deliberately not addressed here

The SDK also applies a second, orthogonal guard Android has no equivalent of: an absolute per-chain priority-fee ceiling (`clampEvmPriorityFee`, 50 gwei for zkSync) to catch a compromised RPC returning a wildly inflated tip. That is a separate concern from the `priority <= max` invariant and out of scope for this issue — worth a follow-up ticket.

## Verification

**Unit tests — 3 new, all green.**

`ZkFeeServiceTest` (preview path):
- `priority fee is clamped to max fee when the RPC returns a tip above the cap` — a `max=7 / priority=9` estimate yields `priority=7`
- `priority fee below the max fee is left untouched` — a `max=7 / priority=2` estimate is passed through unchanged

`BlockChainSpecificRepositoryImplTest` (real send path):
- `zkSync specific clamps a priority fee that exceeds the max fee` — a `max=77 / priority=99` estimate yields `priorityFeeWei=77`

```
:data:testDebugUnitTest   BUILD SUCCESSFUL   (full module suite)
assembleDebug             BUILD SUCCESSFUL
```

**Mainnet no-regression proof.** Real send on zkSync Era from a device build of this branch, decoded from `eth_getTransactionByHash` / `eth_getTransactionReceipt`:

| Field | On-chain value | Advanced gas fee sheet |
|---|---|---|
| `maxFeePerGas` | 45,250,000 wei = 0.04525 gwei | `0.04525` ✅ |
| `maxPriorityFeePerGas` | 0 | `0` ✅ |
| `gas` | 133,156 | `133156` ✅ |
| `value` | 0.000410715242139985 ETH | matches amount sent ✅ |
| `type` | `0x2` (EIP-1559 envelope) | — |
| `status` | `0x1` — success | — |

All three gas fields are byte-exact between the in-app fee sheet and the signed transaction, which is the point worth proving: `spec.priorityFeeWei` — the value the clamp writes — is what actually reaches WalletCore and gets broadcast, unmangled. Broadcast was accepted with no `ErrTipAboveFeeCap`.

Fee display also reconciles: 133,156 × 0.04525 gwei = 0.00000602 ETH, matching the Network Fee row. Actual charge was 0.0000043850 ETH since `gasUsed` (96,907) came in under the limit — expected.

**What this proves:** the clamp is correctly wired into the real send path with no regression.
**What it does not prove:** the clamp firing. zkSync returned a zero priority fee, so `coerceAtMost` was a no-op. The firing case is covered by the unit tests only — and, per the scoping note above, is not expected to occur in production.

## Checklist

## Additional context

The issue's `files:` header lists stale paths. The real locations are `data/.../repositories/BlockChainSpecificRepository.kt` and `data/.../blockchain/ethereum/ZkFeeService.kt`.

Per the issue's Must-NOT-Do, `EvmApi.zkEstimateFee`'s zero-on-RPC-error behaviour (EVM-003) is left untouched. The interaction is benign: on error the entire `ZkGasFee` is zeros, so the clamp no-ops.

- **Agent**: Claude Code
- **Issue**: #5401
- **Confidence**: high
- **Needs human review for**: none — the change is a two-line clamp on a single chain's fee path, covered by tests and a mainnet send. Worth a second opinion only on whether the defensive-fix framing justifies landing it at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved zkSync/EIP-1559 fee handling by clamping priority fees to never exceed the maximum fee cap.
  * Prevents invalid fee pairs (including negative fee values) from being used in transaction fee calculations.
  * Ensures parsed JSON-RPC quantities are treated as non-negative; invalid or signed values now fall back to zero.
* **Tests**
  * Expanded fee estimation and repository tests to cover priority fee clamping, zero flooring for negatives, and corrected quantity parsing.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeePolicy.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeePolicy.kt
@@ -1,0 +1,14 @@
+package com.vultisig.wallet.data.blockchain.ethereum
+
+import com.vultisig.wallet.data.api.models.ZkGasFee
+import java.math.BigInteger
+
+/**
+ * zkSync's fee pair arrives verbatim from `zks_estimateFee` instead of being derived locally, so it
+ * can carry a priority fee above the max fee. That pair is invalid under EIP-1559 (geth rejects it
+ * as `ErrTipAboveFeeCap`) and WalletCore performs no such validation while encoding, so the
+ * rejection would only surface at broadcast — once the MPC ceremony has already completed and with
+ * no automatic retry.
+ */
+internal fun ZkGasFee.clampedPriorityFee(): BigInteger =
+    maxPriorityFeePerGas.coerceAtMost(maxFeePerGas)
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeePolicy.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeePolicy.kt
@@ -9,6 +9,11 @@ import java.math.BigInteger
  * as `ErrTipAboveFeeCap`) and WalletCore performs no such validation while encoding, so the
  * rejection would only surface at broadcast — once the MPC ceremony has already completed and with
  * no automatic retry.
+ *
+ * The floor holds the pair non-negative independently of how it was parsed: ordering alone would
+ * pass a negative tip straight through, and WalletCore encodes a negative as two's-complement bytes
+ * it reads back unsigned. `convertToBigIntegerOrZero` already rejects a signed RPC quantity at the
+ * parse boundary, so this is the second of two layers rather than the only one.
  */
 internal fun ZkGasFee.clampedPriorityFee(): BigInteger =
-    maxPriorityFeePerGas.coerceAtMost(maxFeePerGas)
+    maxPriorityFeePerGas.coerceIn(BigInteger.ZERO, maxFeePerGas.coerceAtLeast(BigInteger.ZERO))
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeeService.kt
@@ -26,7 +26,7 @@ class ZkFeeService @Inject constructor(private val evmApiFactory: EvmApiFactory)
 
         return Eip1559(
             limit = feeEstimate.gasLimit,
-            maxPriorityFeePerGas = feeEstimate.maxPriorityFeePerGas,
+            maxPriorityFeePerGas = feeEstimate.clampedPriorityFee(),
             maxFeePerGas = feeEstimate.maxFeePerGas,
             networkPrice = feeEstimate.maxFeePerGas,
             amount = feeEstimate.maxFeePerGas * feeEstimate.gasLimit,
--- a/data/src/main/kotlin/com/vultisig/wallet/data/common/StringExtensions.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/common/StringExtensions.kt
@@ -106,14 +106,21 @@ fun String.remove0x(): String {
 
 fun String.isNotEmptyContract(): Boolean = isNotEmpty() && !equals(ZERO_ADDRESS, ignoreCase = true)
 
+/**
+ * Parses an Ethereum JSON-RPC `QUANTITY`, which is unsigned by spec, so a signed value is malformed
+ * input and yields zero like any other. The sign floor is load-bearing: `BigInteger(String, radix)`
+ * accepts a leading `-`, so `"-5"` (or `"0x-5"`, since only the prefix is stripped) would otherwise
+ * parse into a real negative that passes every downstream magnitude check and reaches WalletCore as
+ * two's-complement bytes it reads back unsigned — `-5` encodes to `0xfb`, or 251.
+ */
 fun String?.convertToBigIntegerOrZero(): BigInteger {
     val cleanedInput = this?.removePrefix("0x")
     return if (cleanedInput.isNullOrEmpty()) {
         BigInteger.ZERO
     } else {
         try {
-            BigInteger(cleanedInput, 16)
-        } catch (e: NumberFormatException) {
+            BigInteger(cleanedInput, 16).coerceAtLeast(BigInteger.ZERO)
+        } catch (_: NumberFormatException) {
             BigInteger.ZERO
         }
     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/BlockChainSpecificRepository.kt
@@ -22,6 +22,7 @@ import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion
 import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_COIN_TRANSFER_LIMIT
 import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_SWAP_LIMIT
 import com.vultisig.wallet.data.blockchain.ethereum.EthereumFeeService.Companion.DEFAULT_TOKEN_TRANSFER_LIMIT_WITH_MARGIN
+import com.vultisig.wallet.data.blockchain.ethereum.clampedPriorityFee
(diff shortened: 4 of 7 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeePolicy.kt
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * zkSync's fee pair arrives verbatim from `zks_estimateFee` instead of being derived locally, so it
+ * can carry a priority fee above the max fee. That pair is invalid under EIP-1559 (geth rejects it
+ * as `ErrTipAboveFeeCap`) and WalletCore performs no such validation while encoding, so the
+ * rejection would only surface at broadcast — once the MPC ceremony has already completed and with
+ * no automatic retry.
+ */
+internal fun ZkGasFee.clampedPriorityFee(): BigInteger =
+    maxPriorityFeePerGas.coerceAtMost(maxFeePerGas)
```
rkokhatskyi: `clampedPriorityFee` fixes ordering but not sign since `coerceAtMost` never rejects negatives, and `convertToBigIntegerOrZero` (`StringExtensions.kt:109`) parses an unprefixed negative hex fee straight into a real BigInteger. A malicious custom zkSync RPC can trigger this, and WalletCore signs the resulting bytes unsigned, reproducing the stuck or rejected tx this fix targets

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeePolicy.kt
aminsato: Confirmed, and worse than the ordering case in one respect — thanks for catching it.

Traced it end to end before fixing. `removePrefix("0x")` means both `"-5"` and `"0x-5"` reach `BigInteger(cleaned, 16)`, which honours the leading `-`, so both parse to a real `-5` (only `"-0x5"` fails). `coerceAtMost` is blind to it because a negative is *already* below the cap — `min(-5, 100) = -5`. Then `EthereumGasHelper.kt:53,58` calls `priorityFeeWei.toByteArray()`, and WalletCore reads the two's-complement bytes back unsigned: `-5` → `0xfb` → **251**. So the tip comes out inflated rather than merely wrong-signed, which is the same post-keysign rejection this PR set out to prevent.

Fixed in `35ddd7524`, in two layers — the sign hole was never confined to the priority fee, since both zkSync call sites also pass `maxFeePerGas` and `gasLimit` through raw:

1. **`StringExtensions.kt:109`, the root cause you pointed at** — `.coerceAtLeast(BigInteger.ZERO)` on the parse result. Every one of the 16 callers parses an Ethereum JSON-RPC `QUANTITY`, unsigned by spec, so a signed value is malformed input and zero matches the existing `OrZero` contract. This closes it for gas, nonce, balance and `baseFeePerGas` at the same time.
2. **`clampedPriorityFee` itself** — `coerceIn(ZERO, maxFeePerGas.coerceAtLeast(ZERO))`, so the invariant holds for any caller regardless of how the pair was parsed. The inner `coerceAtLeast` is load-bearing rather than belt-and-braces: a bare `coerceIn(ZERO, negativeMax)` throws `IllegalArgumentException`, which would just trade one failure for another.

6 new tests — sign rejection at the parse boundary (`"-5"`, `"0x-5"`, `"-2540be400"`), plus negative-tip and negative-max-fee cases through `ZkFeeService`. Full `:data` suite green: 2310 tests, 0 failures.

One judgement call worth your view: fix (1) lands in a shared helper, so it reaches well beyond this PR's stated zkSync scope. I went that way because fixing only the fee helper would leave `maxFeePerGas` and `gasLimit` exposed to the identical bug, and a per-call-site guard would be 16 copies of the same floor. Happy to split it into its own PR and leave this one as the zkSync clamp alone if you'd rather keep the blast radius tight.

johnnyluo (approved):
LGTM

PR #5451, opened 2026-07-28
fix(sui): surface JSON-RPC error details and terminal tx-status failures

### LINKED ISSUE #5444: [Fix] SuiApiImpl discards the JSON-RPC error envelope, surfacing one hardcoded generic message for every real failure
# [Fix] SuiApiImpl discards the JSON-RPC error envelope, surfacing one hardcoded generic message for every real failure

## Problem
`getBalance`, `getReferenceGasPrice`, `getAllCoins`, `executeTransactionBlock`, and `dryRunTransaction` all null-check `response.result` and throw a hardcoded generic string on failure, never reading `response.error` (which already exposes `.message`/`.code`). A malformed or rejected RPC call (e.g. insufficient gas, a locked/equivocated gas object, invalid signature) returns a real, actionable JSON-RPC error, but the user only ever sees the generic string. For `checkStatus`, any non-not-found RPC error (rate limit, malformed digest, indexer outage) is collapsed into the same "keep polling" outcome as a genuine not-found, so a persistent error is masked as "Pending" until the full timeout instead of surfacing immediately.

## Expected Behavior
Thrown errors should include the real RPC error message/code, matching iOS's `Errors.broadcastFailed(error)` pattern, and `checkStatus` should distinguish a genuine not-found from a terminal RPC error the way iOS's `SuiTransactionStatusProvider` does.

## Steps to Reproduce
1. Trigger a Sui broadcast or dry-run that the node rejects with a specific error (e.g. a stale gas-object version conflict).
2. The app shows a generic "Failed to execute transaction block"/"Failed to dry run transaction block" message with no indication of the real cause.

## Solution
In each `SuiApiImpl` method, check `response.error` before falling back to the generic message, and include `response.error?.message` (and code) in the thrown exception. For `checkStatus`, thread the `RpcError` back to `SuiStatusProvider` and branch on code: the not-found-style code maps to retry, any other code maps to a terminal failure.

### Must Do
- [ ] Surface the real RPC error message/code in `getBalance`/`getReferenceGasPrice`/`getAllCoins`/`executeTransactionBlock`/`dryRunTransaction`
- [ ] Have `checkStatus` distinguish a genuine not-found from a terminal RPC error

### Must NOT Do
- Don't change the happy-path response parsing
- Don't change polling intervals/timeouts

## Acceptance Criteria
- [ ] A rejected broadcast/dry-run surfaces the real RPC error message to the user
- [ ] New test asserts a non-not-found `checkStatus` error surfaces as terminal, not endless retry

rkokhatskyi:
## Summary

`SuiApiImpl` discarded the JSON-RPC `error` envelope on every failed call and threw a single hardcoded generic message instead of the node's real error. `checkStatus` also collapsed every RPC error into the same retryable outcome as a genuine not-found digest, so a persistent failure (rate limit, indexer outage) was masked as pending until the full poll timeout instead of surfacing.

- `getBalance`/`getReferenceGasPrice`/`getAllCoins`/`executeTransactionBlock`/`dryRunTransaction` now include the RPC error's message and code in the thrown exception when present, falling back to the original message otherwise.
- `checkStatus` now distinguishes the not-found RPC code (`-32602`, matching iOS's `SuiTransactionStatusProvider`) from any other error. A not-found code still resolves to a retryable outcome; any other code now throws, which `SuiStatusProvider` maps to a terminal `TransactionResult.Failed` so the poll loop stops immediately instead of retrying until timeout.

Closes #5444

## Testing

- `./gradlew :data:testDebugUnitTest` (new `SuiApiTest` and `SuiStatusProviderTest`, plus the full data module suite)
- `./gradlew :data:ktfmtCheck :data:lintDebug`
- `./gradlew assembleDebug`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Sui transaction status handling for pending, not-found, confirmed, and failed transactions.
  - Terminal RPC errors now surface clear error messages and codes instead of being treated as pending.
  - Not-yet-landed transactions continue to be handled as pending.

- **Tests**
  - Added coverage for Sui RPC errors, successful responses, transaction status mapping, and retry behavior.


johnnyluo (approved):
LGTM

PR #5452, opened 2026-07-29
fix(token detail): stop the sheet re-anchoring while its data loads

aminsato:
## Description

Reported via a screen recording: opening a token from a chain screen makes the detail sheet appear, sit noticeably too low with empty skeletons, and then lurch upward once the balance arrives.

Frame-by-frame on the original recording (91 frames, 1080x2400): the sheet appeared fully placed in a single frame with no slide-in, held ~73dp below its resting place for ~600 ms with the Network row clipped under the gesture bar, then snapped upward. It reached rest **880 ms after it first appeared**.

Three things combine to cause it:

- **The load only started after the open animation settled.** `TokenDetailScreen` wired `onExpand` to `refresh()`, and `DottyBottomSheet` fires `onExpand` only once `sheetState.currentValue` settles — so the sheet was always composed, measured and drawn before any data existed.
- **The action row measured 0 dp on the first frame.** `canSwap` / `canSend` / `canBuy` / `canDeposit` all default to `false` and only resolve once the account loads, so the row went from 0 dp to ~76 dp in one recomposition. That was the single biggest height jump.
- **`LoadableValue` kept re-measuring after the values landed.** Its `AnimatedContent` used the default `transitionSpec`, which carries a `SizeTransform`; four of them in one sheet meant the content height was still changing for hundreds of ms, and `ModalBottomSheet` re-derives its anchors on every size change. The skeleton's flat `defaultMinSize(20.dp, 16.dp)` also didn't reserve the loaded line's height.

Fixes:

- `TokenDetailViewModel` loads in `init` instead of on expand, so the fetch is in flight while the sheet animates in.
- The action row reserves `heightIn(min = assetActionButtonHeight)`. That value is exported from `AssetActionButton` and derived from the same `52.dp` icon box, `8.dp` gap and caption line height the button itself uses, so it can't drift from the real thing.
- `LoadableValue` cross-fades only (`fadeIn() togetherWith fadeOut() using null`) and sizes its placeholder from the passed style's line height, so skeleton and loaded text occupy identical space.
- Dropped `LaunchedEffect(Unit) { sheetState.expand() }` in `DottyBottomSheet`. `ModalBottomSheet` already animates to Expanded on first composition and `skipPartiallyExpanded` leaves nowhere else to settle; expanding on the same frame short-circuited that animation, which is why the sheet had no slide-in at all.

### Original report

The recording the bug was reported with, on a real device. The sheet pops in with empty skeletons, holds too low with the Network row clipped under the gesture bar, then lurches up once the balance lands. The wallet address is blurred.

reported bug

## Which feature is affected?

None of the above — presentation only. No keysign, broadcast or fee path is touched.

## Checklist

No unit test: the defect is layout timing inside `ModalBottomSheet`'s anchor calculation, which a JVM test can't observe. Verified instead by recording the same launch on both builds and measuring the sheet position per frame (numbers below). A debug preview is included so the check is repeatable.

## Screenshots

Captured on `emulator-5554` (1080x2400, animation scales at 1.0) via a new debug preview, `PreviewActivity -e screen token_detail_sheet_loading`, which composes the sheet with an empty model and lands the account 700 ms later — the real sequence. Both clips are trimmed to the same window around the sheet opening.

| Before | After |
|--------|-------|
| before | after |

The loading frame is where it shows clearest — same moment in both builds, sheet at two different heights:

| Before (loading) | After (loading) |
|--------|-------|
| before loading | after loading |

| Before (loaded) | After (loaded) |
|--------|-------|
| before loaded | after loaded |

Drag-handle y tracked frame by frame across both recordings:

| | Before | After |
|---|---|---|
| enter animation | none — fully placed in one frame | slides in over ~215 ms (469 → 268) |
| position while loading | y=320, ~73 dp below its resting place | y=268, already final |
| when the data lands | single-frame snap 320 → 270 | stays at 268, no movement |
| time to settle | ~615 ms after appearing | settles once and stays |

## Additional context

- `assetActionButtonHeight` is new public-in-module API on `AssetActionButton`. Any other row built from asynchronously-resolved capability flags can reserve it the same way.
- The inner `TokenDetailScreen(uiModel, …)` composable went from `private` to `internal` with defaulted callbacks so the debug preview can drive it directly. Happy to keep it private and duplicate the content in the preview instead if that's preferred.
- `LoadableValue` is shared with `VaultAccountsScreen`, `BalanceBanner`, `AccountItem`, `ChainAccount`, `ChainTokensScreen` and `VaultListScreen`. The change makes every one of them size-stable across a load; no call site needed edits.
- `TokenAddressQrBottomSheet` has the identical late-load pattern (`onExpand = viewModel::loadData`). Left alone to keep this diff focused — worth a follow-up.
- The equivalent sheet on iOS is a separate implementation and wasn't inspected; if it shows the same jump it needs its own ticket.

- **Agent**: Claude Code
- **Issue**: none — reported by screen recording
- **Confidence**: high
- **Needs human review for**: the `private` → `internal` widening of `TokenDetailScreen`, and whether the `LoadableValue` cross-fade (no size animation) is wanted everywhere it's used

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -21,7 +21,11 @@ import androidx.compose.foundation.text.input.rememberTextFieldState
 import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.geometry.Offset
@@ -82,6 +86,7 @@ import com.vultisig.wallet.ui.models.ChainTokenUiModel
 import com.vultisig.wallet.ui.models.ChainTokensUiModel
 import com.vultisig.wallet.ui.models.ChainUiModel
 import com.vultisig.wallet.ui.models.DeviceMeta
+import com.vultisig.wallet.ui.models.TokenDetailUiModel
 import com.vultisig.wallet.ui.models.TransactionDetailsUiModel
 import com.vultisig.wallet.ui.models.TransactionScanStatus
 import com.vultisig.wallet.ui.models.VaultDetailUiModel
@@ -128,6 +133,7 @@ import com.vultisig.wallet.ui.models.swap.VultTierGateUiModel
 import com.vultisig.wallet.ui.models.toNetworkUiModel
 import com.vultisig.wallet.ui.models.v3.ReviewVaultDevicesUiState
 import com.vultisig.wallet.ui.screens.ChainSelectionScreen
+import com.vultisig.wallet.ui.screens.TokenDetailScreen
 import com.vultisig.wallet.ui.screens.TransactionDoneView
 import com.vultisig.wallet.ui.screens.VaultDetailScreen
 import com.vultisig.wallet.ui.screens.cosmosstaking.CosmosStakingPositionsContent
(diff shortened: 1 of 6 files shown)


johnnyluo (approved):
LGTM

Issue #5439, opened 2026-07-28
FCM token re-registration races service teardown — a lost race permanently kills pushes


realpaaao:
`VultisigFirebaseMessagingService.onNewToken` launches the re-register network call on `serviceScope`, which `onDestroy` cancels — and a `FirebaseMessagingService` stops itself once the callback returns, so the HTTP re-registration races service teardown. Lose the race → the server keeps the dead token while the local toggle stays ON.

No retry anywhere: `reRegisterAllOptedInVaults` swallows per-vault failures with a log line, `refreshTokenIfNeeded` early-returns whenever any token is already stored, and nothing calls it at app start — so a failed re-registration never heals. Classic "pushes worked, then stopped forever".

**Fix:** move re-registration to WorkManager (backoff, survives teardown); add a startup reconcile that re-registers all enabled vaults.

Receipts: `data/services/VultisigFirebaseMessagingService.kt` (`onNewToken`, `onDestroy`), `data/services/PushNotificationManager.kt` (`refreshTokenIfNeeded`, `reRegisterAllOptedInVaults`).

Issue #3302, opened 2026-03-11
[Push Notifications] notifyVaultDevices gated on sender's own opt-in status


rkokhatskyi:
> **This issue was generated by Claude (AI). Introduced in PR #3271.**

# [Fix] notifyVaultDevices gated on sender's own opt-in status [Push Notifications]

## Problem

`PushNotificationManager.notifyVaultDevices()` checks `isVaultOptedIn(vault.id)` before sending the notification to the server. This checks whether the **initiating** device (the sender) has opted in to notifications. If device A starts a keysign but never enabled notifications for itself, device B (which did opt in and is registered on the server) will never receive the push notification.

The server is responsible for deciding which registered devices to notify. The sender should always call the notify endpoint regardless of its own opt-in status. iOS does not gate sending on the sender's opt-in. The same check also exists in `KeysignFlowViewModel.sendNotification()` at line 405.

## Expected Behavior

The notify API call should be sent whenever a keysign is initiated, regardless of the sender's own notification opt-in status. The server handles delivery to registered devices.

## Steps to Reproduce
1. Set up a 2-of-3 vault with devices A, B, C
2. On device B and C, enable push notifications for the vault
3. On device A, do NOT enable push notifications
4. On device A, initiate a keysign transaction
5. Observe that devices B and C never receive a push notification

## Solution

Remove the `isVaultOptedIn(vault.id)` check from `PushNotificationManager.notifyVaultDevices()`. Also remove the duplicate check in `KeysignFlowViewModel.sendNotification()`. The `enableNotification` UI flag in `KeysignFlowUiState` can still control whether the resend button is visible but should not gate the initial automatic notification send.

### Must Do
- [ ] Remove `isVaultOptedIn` check from `PushNotificationManager.notifyVaultDevices()`
- [ ] Remove duplicate `isVaultOptedIn` check from `KeysignFlowViewModel.sendNotification()`
- [ ] Keep the resend button visibility gated on opt-in status (UI only)
- [ ] Always send the initial notification when keysign starts

### Must NOT Do
- Don't change unrelated code
- Don't refactor surrounding logic
- Don't modify the registration/unregistration flow

## Acceptance Criteria
- [ ] `./gradlew assembleDebug` succeeds
- [ ] Keysign notification is sent even if the initiating device has not opted in
- [ ] Resend button still only appears when the device has opted in
- [ ] Other vault participants who opted in receive the push notification

## References
[...]

PR #5454, opened 2026-07-29
fix: notify vault devices without a local FCM token

### LINKED ISSUE #5440: notifyVaultDevices aborts when the sender has no stored token — while the UI reports 'notifications sent'
`PushNotificationManager.notifyVaultDevices` early-returns when the *local* device has no stored FCM token — but the initiator doesn't need a token to notify its peers, and iOS has no such gate (#3302's fix removed the sibling opt-in gate here but left this one).

Worse, the early return is a normal return, so `KeysignFlowViewModel` still shows `push_notifications_sent` and starts the 30s resend cooldown — the user is told the push went out and is blocked from retrying, while nothing was sent.

**Fix:** drop the gate (or if it's a deliberate "this device isn't in the push system" guard, keep it but stop claiming success); surface notify failures instead of unconditional success.

Receipts: `data/services/PushNotificationManager.kt` (`notifyVaultDevices` token guard), `ui/models/keysign/KeysignFlowViewModel.kt` (`sendNotification` success banner + cooldown).

aminsato:
Closes #5440

## Problem

`PushNotificationManager.notifyVaultDevices` early-returned when the **initiating** device had no stored FCM token, so a keysign started from that device never reached `/notify` — and peers that *had* opted in were never pushed.

The guard checked a value the request never uses. `NotifyRequest` carries only `vault_id`, `vault_name`, `local_party_id` and `qr_code_data`; the server selects recipients from the devices registered under `vault_id`. The initiator's own token has no bearing on the call.

This is the same gate #3302 already removed from this function for the sender's opt-in flag — the token check was its surviving sibling, both introduced together in #3271.

Because the skip was a plain `return`, `KeysignFlowViewModel.sendNotification` carried on as if it had succeeded: it recorded the QR as notified, showed **"Notification sent successfully!"**, and started the 30s resend cooldown. The user was told the push went out *and* was locked out of retrying, while nothing had been sent.

## Fix

Remove the guard.

That also fixes the reporting half on its own — `notificationApi.notify` throws via `throwIfUnsuccessful()`, so with the silent return gone a real failure reaches `sendNotification`'s `safeLaunch(onError = …)`, which shows **"Push notification failed"** and leaves both `lastNotifiedQrData` and the cooldown untouched, so the resend button stays live. `safeLaunch` rethrows `CancellationException`, so navigating away isn't misreported as a push failure.

## How the no-token state is reachable

`getStoredToken()` / `refreshTokenIfNeeded()` have no callers outside `PushNotificationManager`. The token is written only by:

1. `VultisigFirebaseMessagingService.onNewToken` → `serviceScope.launch { … }`, a scope `onDestroy` cancels (#5439) — the write can be lost
2. `refreshTokenIfNeeded()`, reached only from `setVaultOptIn` / `setVaultsOptIn` **when enabling**

So a user who never opted in, or who lost the #5439 race, has no token — and then every keysign silently skipped the notify while claiming success. #5439 and #5440 compound each other.

## Cross-platform

iOS `PushNotificationManager.swift:248` has no token gate, so this brings Android to parity.

Worth noting separately: iOS swallows the error (`catch { logger.error(…) }`) and `KeysignDiscoveryView.swift:313` sets the success banner unconditionally, so iOS has the false-success half of this bug. Not addressed here — needs a sibling iOS issue.

## Test plan

New `PushNotificationManagerTest` — 4/4 green via `./gradlew :app:testDebugUnitTest`:

- notify fires when there is **no** stored token (the regression — fails on the old code, where `notify` was never called)
- notify fires when there is a stored token
- the request carries the SHA-256 vault id, vault name, local party id and QR data, and no token
- a throwing `notify` propagates rather than returning normally, so the caller can't report success

### Manual

Two Secure-vault devices, A = initiator, B = peer. B has push notifications on for the vault; A has no FCM token (fresh install first launched offline, and Settings → Notifications never opened, since that screen back-fills the token).

A → Send → Sign, landing on peer discovery:

| | Before | After |
|---|---|---|
| Phone A | "Notification sent successfully!" + 30s cooldown | same |
| **Phone B** | nothing, ever | **"Sign transaction — A new signing request is waiting for you"** |

Failure surfacing: on A's peer-discovery screen, enable airplane mode and tap **Resend notification** → "Push notification failed", button stays tappable with no cooldown lockout.

Regression: A with notifications **on** → B still receives the push, A still shows success + cooldown. Unchanged.

No UI changes in this PR — no visual diff to show.

## Summary by CodeRabbit

* **Bug Fixes**
  * Vault device notifications are now sent regardless of whether a local push token is available.
  * Notification failures are properly reported instead of being treated as successful when local token data is missing.

* **Tests**
  * Added coverage for notification delivery with and without a local push token.
  * Added verification of notification payload details and failure handling.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/services/PushNotificationManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/services/PushNotificationManager.kt
@@ -187,12 +187,15 @@ constructor(
         setVaultsOptIn(allVaults.map { it.id to enabled })
     }
 
+    /**
+     * Asks the server to push the keysign QR to this vault's registered devices.
+     *
+     * Deliberately independent of the local device's own push state: [NotifyRequest] carries no
+     * token, and the server picks the recipients from the devices registered under `vaultId`. An
+     * initiator that never opted in — or whose token write was lost — must still be able to notify
+     * peers that did. Failures propagate so the caller can report them rather than claim success.
+     */
     suspend fun notifyVaultDevices(vault: Vault, qrCodeData: String) {
-        val token = getStoredToken()
-        if (token == null) {
-            Timber.d("No FCM token available, skipping notification")
-            return
-        }
         notificationApi.notify(
             NotifyRequest(
                 vaultId = notificationVaultId(vault),
--- a/app/src/test/java/com/vultisig/wallet/data/services/PushNotificationManagerTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/data/services/PushNotificationManagerTest.kt
@@ -1,0 +1,104 @@
+package com.vultisig.wallet.data.services
+
+import android.content.SharedPreferences
+import com.vultisig.wallet.data.api.NotificationApi
+import com.vultisig.wallet.data.api.NotifyRequest
+import com.vultisig.wallet.data.db.dao.VaultNotificationSettingsDao
+import com.vultisig.wallet.data.models.Vault
+import com.vultisig.wallet.data.repositories.VaultRepository
+import io.kotest.matchers.shouldBe
+import io.mockk.coEvery
+import io.mockk.coVerify
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.slot
+import java.io.IOException
+import java.security.MessageDigest
+import kotlinx.coroutines.test.runTest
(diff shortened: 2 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5456, opened 2026-07-29
fix: heal FCM re-registration and deliver pushes on the right channel

### LINKED ISSUE #5439: FCM token re-registration races service teardown — a lost race permanently kills pushes

### LINKED ISSUE #5441: Background pushes land on FCM's auto-created 'Miscellaneous' channel at default importance
Backend keysign messages carry a `notification` block, so when the app is backgrounded/killed the FCM SDK renders the tray notification itself — `onMessageReceived` never runs, and the app's `keysign_requests_channel` (IMPORTANCE_HIGH) is effectively unused in exactly the case that matters.

There is no `com.google.firebase.messaging.default_notification_channel_id` meta-data in the manifest, so users get a default-importance notification (no heads-up peek) on a channel literally named "Miscellaneous" — easy to mute, easy to miss for a time-sensitive keysign request.

Also: nothing detects revoked POST_NOTIFICATIONS or a blocked channel after opt-in (`areNotificationsEnabled()` is never called) — the server keeps sending, the device drops everything, the UI says ON.

**Fix:** add the manifest meta-data pointing at a pre-created high-importance channel (server-side `channel_id` in vultisig/notification#31 completes it); check `areNotificationsEnabled()` in a startup reconcile.

Receipts: `AndroidManifest.xml` (no default channel meta-data), `data/services/VultisigFirebaseMessagingService.kt` (`showSystemNotification` only runs from `onMessageReceived`).

aminsato:
Closes #5439
Closes #5441

Two ends of the same failure: a push that is never re-registered, and a push that arrives but lands silently. They share one fix — a startup reconcile — so they ship together. One commit per issue.

## #5439 — re-registration raced service teardown

`onNewToken` launched the HTTP re-registration on the service's own `CoroutineScope`, which `onDestroy` cancels — and a `FirebaseMessagingService` stops itself as soon as the callback returns. Lose that race and the server keeps a dead token while the local toggle still reads ON.

Nothing healed it afterwards:

- `reRegisterAllOptedInVaults` swallowed every per-vault failure with a log line
- `refreshTokenIfNeeded` early-returned whenever *any* token was already stored
- nothing called it at app start

Classic "pushes worked, then stopped forever".

**Fix.** `onNewToken` is now fully synchronous — the token is persisted on the spot, the network half goes to `PushRegistrationWorker` — so neither half can be lost to teardown. The worker retries with exponential backoff behind a `CONNECTED` constraint, capped at 5 attempts so a permanent failure stops rescheduling. `reRegisterOptedInVaults` reports whether every vault landed instead of silently dropping the ones that didn't, and a failing vault no longer strands the vaults behind it.

`PushNotificationInitializer` enqueues the same worker on every launch — the path back for a device whose registration was already lost. It skips minting a token when no vault has opted in, so users who never asked for pushes are never registered.

Token acquisition moves behind `FcmTokenProvider` so the path is testable off-device.

## #5441 — pushes landed on "Miscellaneous"

Backend keysign messages carry a `notification` block, so a backgrounded or killed app never reaches `onMessageReceived` — the FCM SDK renders the tray entry itself. `keysign_requests_channel` (IMPORTANCE_HIGH) was created *inside* `showSystemNotification`, reachable only from `onMessageReceived`, so in exactly the case that matters it did not exist. With no `default_notification_channel_id` in the manifest either, the SDK fell back to an auto-created channel named "Miscellaneous" at default importance: no heads-up peek, trivially muted, easy to miss for a time-critical signing request.

**Fix.** The channel id is a string resource shared by the manifest meta-data and `KeysignNotificationChannel`, so the two can't drift. The channel is created by `PushNotificationInitializer` at app start — a push that starts the process runs ContentProviders before the messaging service, so the channel exists by the time the SDK renders.

**Second half:** nothing detected a revoked `POST_NOTIFICATIONS` or a muted channel after opt-in — `areNotificationsEnabled()` was never called anywhere. The server kept sending, the device dropped everything, the toggle read ON. The notifications screen now re-reads OS state on every resume (neither revocation raises a callback) and shows a warning that opens the app's system notification settings.

New string in all 10 locales, reusing each locale's existing wording for "system settings" (`notifications_vault_sheet_subtitle`) and "signing request" (`keysign_notification_body`). The channel id itself is `translatable="false"`.

## Test plan

`./gradlew :app:testDebugUnitTest` — **1739 tests, 0 failures**, including 16 new. `./gradlew :app:lintDebug` clean.

`PushRegistrationWorkerTest` (5) — re-registers against the current token; **retries when a vault fails**; retries when no token can be obtained; gives up at the attempt cap; does nothing when no vault has opted in.

`PushNotificationTokenTest` (8) — `storeToken` persists synchronously; `currentToken` prefers the stored token, mints and persists when absent, returns null (not throws) when minting fails; `reRegisterOptedInVaults` registers every vault with the new token, reports failure, and keeps going past a failure; `hasOptedInVaults`.

`NotificationsSettingsViewModelTest` (3) — flags blocked notifications while the in-app toggle still reads ON; stays quiet when the system allows them; keeps the flag when the settings flow re-emits.

### Manual

**#5441:** opt a Secure vault in, background the app, have a peer start a keysign. Expect a heads-up peek under **Keysign Requests** in the app's notification categories — not a silent entry under "Miscellaneous".

**#5441 second half:** with the toggle ON, revoke notifications in system settings, return to Settings → Notifications. Expect the warning; tapping it opens system notification settings; re-enabling and returning clears it.

**#5439:** hard to force by hand — it needs a token rotation whose re-registration fails. Closest approximation: opt in offline so registration fails, then restore network and relaunch; the startup worker should re-register (`adb shell dumpsys jobscheduler | grep push_registration`).

## Not verified on device

No `TRUSTWALLET_PAT` in this environment, so `assembleDebug` can't run and none of this is device-verified — unit tests and lint only, off the Gradle cache. The manual checks above are the ones that still need running, in particular the heads-up peek, which is the whole point of #5441.

## Note

The startup reconcile re-registers on **every** launch — one POST per opted-in vault, typically one or two. That's what #5439 asks for and registration is idempotent server-side, but say the word if you'd rather it were throttled.

## Summary by CodeRabbit

* **New Features**
  * Added automatic push-notification initialization on app startup, including scheduled re-registration for opted-in vaults.
  * Introduced support for a high-priority signing notification channel.
  * Added an in-app settings warning when notifications are blocked by the system, with a one-tap shortcut to system notification settings.
* **Bug Fixes**
  * Improved FCM token handling and notification channel eligibility checks, with more reliable token re-registration and capped retry behavior.
* **Documentation**
  * Added translated “blocked by system” warning text across supported languages.
* **Tests**
  * Added/updated automated tests for token storage, registration retry behavior, and the system-blocked UI state.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -149,6 +149,7 @@ dependencies {
     testImplementation(libs.kotest.assertions.core)
     testImplementation(libs.mockk)
     testImplementation(libs.kotlinx.coroutines.test)
+    testImplementation(libs.androidx.work.testing)
     androidTestImplementation(libs.androidx.junit)
     androidTestImplementation(libs.androidx.test.runner)
     androidTestImplementation(libs.androidx.test.rules)
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -136,6 +136,10 @@
             <meta-data
                 android:name="com.vultisig.wallet.app.startup.WorkManagerInitializer"
                 android:value="androidx.startup" />
+
+            <meta-data
+                android:name="com.vultisig.wallet.app.startup.PushNotificationInitializer"
+                android:value="androidx.startup" />
         </provider>
 
         <service
@@ -151,6 +155,15 @@
                 <action android:name="com.google.firebase.MESSAGING_EVENT" />
             </intent-filter>
         </service>
+
+        <!--
+             A backgrounded message carrying a notification block is rendered by the FCM SDK itself,
+             so onMessageReceived never runs. Without this the SDK invents a "Miscellaneous" channel
+             at default importance and a keysign request arrives with no heads-up peek.
+        -->
+        <meta-data
+            android:name="com.google.firebase.messaging.default_notification_channel_id"
+            android:value="@string/keysign_notification_channel_id" />
 
     </application>
 
--- a/app/src/main/java/com/vultisig/wallet/app/startup/PushNotificationInitializer.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/startup/PushNotificationInitializer.kt
@@ -1,0 +1,31 @@
+package com.vultisig.wallet.app.startup
+
+import android.content.Context
+import androidx.startup.Initializer
+import com.vultisig.wallet.data.services.KeysignNotificationChannel
+import com.vultisig.wallet.data.services.PushRegistrationWorker
+
+/**
+ * Startup reconcile for push delivery.
+ *
+ * Creates the keysign channel before any push can arrive — a backgrounded message is rendered by
+ * the FCM SDK, which never reaches the app's own code, so a channel created lazily on first
+ * notification does not exist when it is needed.
+ *
+ * Then queues a re-registration. Nothing else re-registers after the initial opt-in, so a device
+ * whose token rotated while a registration failed kept a dead token on the server with no path back
+ * — pushes stopped permanently. Running this every launch is the healing path.
+ */
+internal class PushNotificationInitializer : Initializer<Unit> {
+
+    override fun create(context: Context) {
+        val appContext = context.applicationContext
+        KeysignNotificationChannel.ensureCreated(appContext)
+        // KEEP: a run already in flight is registering the same token; restarting it would only
+        // reset its backoff.
+        PushRegistrationWorker.enqueue(appContext, replaceExisting = false)
+    }
+
+    override fun dependencies(): List<Class<out Initializer<*>>> =
+        listOf(WorkManagerInitializer::class.java)
+}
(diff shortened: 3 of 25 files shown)


johnnyluo (approved):
LGTM

Issue #5398, opened 2026-07-26
[Fix] ZkSync fee calculation collapses to all-zero on any RPC error, showing a normal-looking $0.00 fee


rkokhatskyi:
# [Fix] ZkSync fee calculation collapses to all-zero on any RPC error, showing a normal-looking $0.00 fee

## Problem
`EvmApi.zkEstimateFee()` returns an all-zero `ZkGasFee` on any JSON-RPC error. Neither consumer (`ZkFeeService`, `BlockChainSpecificRepository`'s ZkSync branch) applies a floor — they feed `gasLimit`/`maxFeePerGas`/`maxPriorityFeePerGas` straight through with no `try/catch` or `maxOf()`-against-floor, unlike every other EVM chain's fee path. ZkSync is user-overridable via custom RPC, making transient/unsupported-endpoint errors realistic.

## Expected Behavior
An RPC error while estimating ZkSync fees should not silently resolve to zero — it should either throw (blocking the send with a visible error, matching iOS) or fall back to a safe non-zero floor.

## Steps to Reproduce
1. Send native ETH or an ERC-20 on ZkSync while `zks_estimateFee` errors (e.g. via a custom RPC that doesn't support the method, or a transient proxy hiccup).
2. Observe the fee preview shows $0.00.
3. Complete the full multi-device MPC signing ceremony.
4. Broadcast rejects with `gasLimit=0`/`maxFeePerGas=0` (intrinsic gas too low) — the ceremony was wasted.

## Solution
Add the same `maxOf(estimate, floor)` guard used elsewhere to both `ZkFeeService` and `BlockChainSpecificRepository`'s ZkSync branch (floor `gasLimit` at 200000 to match iOS/Windows, plus a sane non-zero `maxFeePerGas`/priority fallback), or make `zkEstimateFee()` throw on JSON-RPC error so the existing catch-and-fallback-to-default path kicks in (which itself then needs the same floor).

### Must Do
- [ ] Stop `zkEstimateFee()` from silently returning all-zero on RPC error
- [ ] Add a floor/fallback in both `ZkFeeService` and `BlockChainSpecificRepository`'s ZkSync branch

### Must NOT Do
- Don't change the happy-path RPC response handling
- Don't change fee floors for any chain other than ZkSync

## Acceptance Criteria
- [ ] `./gradlew assembleDebug` succeeds
[...]

PR #5459, opened 2026-07-29
fix(zksync): price the fee and the signed gas limit from one estimate

### LINKED ISSUE #5425: [Fix] ZkSync's two fee-building paths query zks_estimateFee with different placeholder calldata — preview fee can diverge from the signed gas limit
# [Fix] ZkSync's two fee-building paths query zks_estimateFee with different placeholder calldata — preview fee can diverge from the signed gas limit

## Problem
`BlockChainSpecificRepository`'s ZkSync branch (the real send/signing path) hex-encodes the ASCII bytes of the literal string `"0xffffffff"` — a 10-byte payload — before calling `zks_estimateFee`. `ZkFeeService` (the live preview path) instead passes the same string as a real 4-byte hex payload. Since zkSync's gas/pubdata pricing scales with calldata byte count, these two calls can return different `gasLimit`/fee values for the same recipient.

## Expected Behavior
Both the live preview and the actual signed transaction should compute the fee estimate from identical calldata, so what's shown while typing matches what gets signed.

## Steps to Reproduce
1. Start a ZkSync send or swap and observe the "Est. network fee" shown while typing (computed via `ZkFeeService`'s 4-byte payload).
2. Complete the send — the actual signed `gasLimit` comes from `BlockChainSpecificRepository`'s inline 10-byte payload via a separate `zkEstimateFee` call.
3. The two values can differ, since they query the RPC with different-sized placeholder calldata.

## Solution
Make `ZkFeeService` use the exact same placeholder-construction logic as `BlockChainSpecificRepository`'s inline path (extract one shared helper), or better, remove the duplication and have `BlockChainSpecificRepository`'s ZkSync branch call through `FeeServiceComposite`/`ZkFeeService` like every other EVM chain.

### Must Do
- [ ] Unify the placeholder-calldata construction between the two ZkSync fee-estimate call sites
- [ ] Prefer a single shared code path over duplicated logic

### Must NOT Do
- Don't change the zero-on-RPC-error handling (tracked separately as issue #5398 / EVM-003)
- Don't change the priority-fee clamp behavior (tracked separately as issue #5401 / EVM-007)

## Acceptance Criteria
- [ ] The live preview fee and the actually-signed gas limit are computed from identical calldata for ZkSync
- [ ] New test asserts both call sites build the same payload

aminsato:
## Summary

zkSync's two fee-building paths queried `zks_estimateFee` with **different placeholder calldata**, so the fee shown while typing could be derived from a different estimate than the gas limit the transaction was actually signed with.

| Call site | `data` sent to `zks_estimateFee` | size |
|---|---|---|
| `BlockChainSpecificRepository` (signing) | `0x30786666666666666666` — UTF-8 bytes of `"0xffffffff"`, hex-encoded | 10 bytes |
| `ZkFeeService` (live preview) | `0xffffffff` | 4 bytes |

zkSync prices gas *and pubdata* by calldata size, so the two calls can return different `gasLimit`/fee values for the same recipient.

iOS (`RpcEvmService.getGasInfoZk`) does the UTF-8 → hex encoding, so the repository's payload is the cross-platform-correct one and `ZkFeeService` was the divergent side.

## Changes

Rather than duplicate the payload construction in two places, this takes the route the issue prefers — a single code path:

- **`BlockChainSpecificRepository`** — the ZkSync branch no longer makes its own `zkEstimateFee` call. The fee lookup is hoisted above the zk/non-zk split, and ZkSync's gas limit now comes from the same `Eip1559` result that priced the preview, so the two cannot diverge. Routes through `FeeServiceComposite` → `ZkFeeService` like every other EVM chain. Also removes one duplicate RPC round-trip per send.
- **`ZkFeeService`** — the placeholder lives in one place as `PLACEHOLDER_CALL_DATA`, built the iOS way.
- Non-zkSync EVM logic is unchanged — only relocated. Most of the diff is re-indentation.

### Interaction with #5438

#5438 (merged while this was in progress) added `clampedPriorityFee()` to **both** zkSync call sites — exactly the duplication this issue is about. After unification the clamp lives in `ZkFeeService` alone; #5438's regression test `zkSync specific clamps a priority fee that exceeds the max fee` still passes through the new path, confirming the behavior is preserved rather than bypassed.

### Deliberately out of scope

- Zero-on-RPC-error handling (#5398 / EVM-003) — untouched.
- Priority-fee clamp behavior (#5401 / EVM-007) — untouched.
- ZkSync still ignores the caller-supplied `gasLimit` override, exactly as before.

## Test plan

- `./gradlew :data:testDebugUnitTest` — **2435 tests, 0 failures**.
- New `BlockChainSpecificRepositoryImplTest` case asserts both entry points hit `zkEstimateFee` with byte-identical calldata.
- New `ZkFeeServiceTest` case pins `PLACEHOLDER_CALL_DATA` to `0x30786666666666666666`, so a regression back to the 4-byte literal fails CI.
- Existing zkSync repository and clamp tests pass unchanged.
- ktfmt clean.

### Manual verification

The calldata delta is 6 bytes of pubdata — sub-cent on zkSync, below the app's display precision — so this is not distinguishable by eye in the fee text. The observable is the **Gas Limit** field in the Send screen's gas-settings dialog, which is prefilled from `spec.gasLimit` (the signing path):

1. zkSync → Send → recipient + amount → open gas settings → Gas Limit populates with a live `zks_estimateFee` value (not blank, not `0`, not a generic default).
2. Broadcast, then confirm the explorer's gas limit equals the value the dialog showed — i.e. the previewed number is the signed number.
3. Regression: Ethereum native send, zkSync ERC-20 send, and a zkSync swap all still price and sign.

Closes #5425

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved zkSync fee estimation by deriving the same call targets and calldata (including proper empty memo data) for both fee preview and signing.
  * Refined EVM fee/gas computation to correctly handle EIP-1559 vs non–EIP-1559 networks and updated zkSync gas-limit handling.
  * Ensured ERC-20 fee calldata is built for `transfer(address,uint256)` rather than a simple recipient transfer.

* **Tests**
  * Expanded fee estimation tests to validate memo payload encoding/decoding and swap/transfer calldata construction across scenarios.
  * Added assertions that zkSync fee estimation is invoked with byte-identical calldata for preview and signing.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeeService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeeService.kt
@@ -5,6 +5,10 @@ import com.vultisig.wallet.data.blockchain.FeeService
 import com.vultisig.wallet.data.blockchain.model.BlockchainTransaction
 import com.vultisig.wallet.data.blockchain.model.Eip1559
 import com.vultisig.wallet.data.blockchain.model.Fee
+import com.vultisig.wallet.data.blockchain.model.Swap
+import com.vultisig.wallet.data.blockchain.model.Transfer
+import com.vultisig.wallet.data.common.add0x
+import com.vultisig.wallet.data.utils.Numeric
 import javax.inject.Inject
 
 class ZkFeeService @Inject constructor(private val evmApiFactory: EvmApiFactory) : FeeService {
@@ -15,16 +19,14 @@ class ZkFeeService @Inject constructor(private val evmApiFactory: EvmApiFactory)
         estimateFee(transaction)
 
     private suspend fun estimateFee(transaction: BlockchainTransaction): Fee {
-        val chain = transaction.coin.chain
-        val coin = transaction.coin
-        val toAddress = transaction.to
-        val evmApi = evmApiFactory.createEvmApi(chain)
+        val evmApi = evmApiFactory.createEvmApi(transaction.coin.chain)
+        val call = resolveCall(transaction)
 
         val feeEstimate =
             evmApi.zkEstimateFee(
-                srcAddress = coin.address,
-                dstAddress = toAddress,
-                data = PLACEHOLDER_CALL_DATA,
+                srcAddress = transaction.coin.address,
+                dstAddress = call.to,
+                data = call.data,
             )
 
         return Eip1559(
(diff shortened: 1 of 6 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeeService.kt
```diff
@@ -32,4 +35,19 @@ class ZkFeeService @Inject constructor(private val evmApiFactory: EvmApiFactory)
             amount = feeEstimate.maxFeePerGas * feeEstimate.gasLimit,
         )
     }
+
+    companion object {
+        private const val PLACEHOLDER_MEMO = "0xffffffff"
```
rkokhatskyi: We need to build calldata instead of guessing, longer memos consume more bytes

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/ethereum/ZkFeeService.kt
aminsato: Agreed — fixed in 98a7eba19. The placeholder is gone; `ZkFeeService.resolveCall()` now builds the call the transaction will actually broadcast:

| Transaction | `to` | `data` |
|---|---|---|
| Native send | recipient | the memo bytes (`0x` when there is no memo) |
| ERC-20 send | **token contract** | `transfer(address,uint256)` |
| Swap | router | the router calldata (`0x` when the caller has none yet) |

Two things worth flagging:

**The memo is encoded the way the signing path encodes it.** `EvmHelper` puts the memo in the payload via `toByteStringOrHex()`, so a hex-looking memo (`0xdeadbeef`) is 4 bytes, not the 10 characters it is written with. The new `memoCallData` helper applies the same rule — otherwise "build the real calldata" would still mis-size hex memos.

**The ERC-20 path had a second, larger version of this bug.** A zkSync token send signs `to = contractAddress` with `erc20Transfer` (`ERC20Helper.buildErc20TokenTransfer`), but the estimate was aimed at the recipient EOA — so it priced a bare value transfer and the signed gas limit could not cover the token transfer. It now estimates against the contract with the real transfer calldata.

`erc20TransferCallData` moved verbatim out of `EthereumFeeService` into a shared `EvmCallData.kt` so both fee services build it from one definition. I left `EthereumFeeService.resolveL1CallContext`'s memo handling alone — changing OP-stack L1 fee sizing felt out of scope here; happy to follow up separately if you want them unified.

Note this now diverges from iOS `RpcEvmService.getGasInfoZk`, which still defaults to `memo: "0xffffffff"` and `toAddress: .zeroAddress`. Worth an iOS ticket.

Tests: 6 new `ZkFeeServiceTest` cases (no memo / text memo / longer memo / hex memo / ERC-20 / swap with and without callData), and the repository parity test now pushes a memo through `getSpecific` and pins both entry points to the memo-derived payload. `:data` suite green at 2441 tests.

johnnyluo (approved):
LGTM

PR #5461, opened 2026-07-29
fix(limit order): cosigner's confirmation reads as a plain swap

aminsato:
## Summary

Follow-up to #4154 (THORChain limit swap — Place flow).

### Transaction hash
https://mempool.space/tx/9b8973d49f50033d5436996ec5b69bb8e43b9b345f93abcf924f32e5808264b5

A `=<` limit order reaches the joining device as an ordinary THORChain swap payload plus its memo. The target price and the lifetime only ever existed as form state on the placing device, so the cosigner was asked to approve **"You're swapping"** — no price floor, no expiry, nothing distinguishing it from a market swap.

- Recover both from the memo, which *is* the order: LIM against the sold amount gives the target price, the interval blocks give the lifetime (`LimitSwapMemo.targetPrice` / `expiryHours`).
- Format them through the same use case the placing device uses, extracted for the purpose — one formatter is what makes the two screens read identically.
- Fix the fee rows, which disagreed for a separate reason (see below).

## Fee parity

The limit builder passed the source-chain **gas fee** as `estimatedFees`, a field the verify mapper reads as a *destination-token* amount for THORChain — a BTC gas fee priced as ETH, shown as the Swap Fee and then added on top of the Network Fee in the total. It also never asked for the affiliate and outbound fees a resting order actually pays, while the cosigner did.

Both devices now read one THORChain quote, through the same repository call with the same request shape (vault's own destination address, no referral code, no bps discount). The discount is omitted deliberately: it's known only to the placing device, so including it would put the two screens back out of agreement.

Every route fetches that quote now, not just ERC20 placements, so a quote failure degrades to no fee breakdown rather than blocking an order whose memo is already fully determined. ERC20 still fails closed on a missing router.

## Before / After

Joined device — the reported bug:

| Before | After |
|--------|-------|
| before | after |

Placing device — fee rows:

| Before | After |
|--------|-------|
| before | after |

Real Android renders (PreviewActivity on an emulator), not design mockups. The placing device's existing confirmation screen is the reference for the joined one — this is a behavioral gap, not a visual mismatch, so there is no separate Figma node.

## Changes

| | Before (cosigner) | After (cosigner) |
|---|---|---|
| Title | "You're swapping" | "You're placing a limit order" |
| Target Price / expiry row | absent | `1 ETH = $1,838.58` · `12h` |

| | Before (placing device) | After (placing device) |
|---|---|---|
| Swap Fee | $0.01 — gas fee read as the bought asset | $0.04 — quote's affiliate fee |
| Outbound Fee | absent | $0.25 |
| Max. Total Fee | $0.47 — network fee counted twice | $0.74 |

Known remaining difference: the two devices still show a different **Network Fee** (the placing device shows the swap form's projection, the cosigner recomputes the UTXO transaction plan from the payload). That is generic join-path behavior for every UTXO swap, not limit-specific, so it is out of scope here.

## Test plan

- [x] `:app` and `:data` unit tests pass (1729 tests, 0 failures)
- [x] New coverage: memo → target price / lifetime round-trip and its null cases; fee breakdown rescaled from THORChain's 1e8 into the bought asset's decimals; degraded path when the quote fails
- [x] `ktfmtCheck` clean, `assembleDebug` succeeds
- [x] Joined-device and placing-device confirmations captured on emulator
- [ ] End-to-end: place a BTC→ETH limit order on a 2-of-2 vault and confirm both devices show the same title, target price, expiry, and fee rows
- [ ] Market swaps unaffected on both devices (no limit memo → unchanged screen)
- [ ] ERC20 limit order still fails closed when THORChain reports no router

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added limit-order details to join-device and transaction confirmation screens, including target price and expiry.
  - Added clearer fee breakdowns for limit-order confirmations, including provider, outbound, and total fees.
  - Improved limit-order price and expiry display across supported swap flows.
  - Enhanced transaction construction with route-aware fee information when available.

- **Bug Fixes**
  - Improved handling when quote data or limit-order metadata is unavailable, allowing transactions to continue with appropriate fallback values.

- **Tests**
  - Added coverage for limit-order metadata, pricing, expiry conversion, fee calculations, and memo round-tripping.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -217,6 +217,12 @@ class PreviewActivity : ComponentActivity() {
                     "limit_swap_form_assets" ->
                         LimitSwapFormPreview(expandedSection = LimitFormSection.Asset)
                     "limit_order_confirm" -> LimitOrderConfirmPreview()
+                    "limit_order_fees_before" -> LimitOrderFeesPreview(carriesRouteFees = false)
+                    "limit_order_fees_after" -> LimitOrderFeesPreview(carriesRouteFees = true)
+                    "join_limit_order_confirm_before" ->
+                        JoinLimitOrderConfirmPreview(recognizesLimitOrder = false)
+                    "join_limit_order_confirm_after" ->
+                        JoinLimitOrderConfirmPreview(recognizesLimitOrder = true)
                     "limit_order_done" -> LimitOrderDonePreview()
                     "swap_confirm" -> SwapConfirmPreview()
                     "swap_confirm_chain" -> SwapChainIndicatorPreview()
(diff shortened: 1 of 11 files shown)


johnnyluo (approved):
LGTM

PR #5462, opened 2026-07-29
feat: Add price chart and market data to token detail screen

### LINKED ISSUE #5428: [Add] Price chart + market data on the coin detail screen [Token Detail]
# [Add] Price chart + market data on the coin detail screen [Token Detail]

## Problem

`TokenDetailScreen` shows ticker, balance, spot price and chain name only — no price history, no 24h change, no market cap. Every mainstream wallet shows a price chart here, and users have to leave the app to get it.

This is a parity follow-up to vultisig-ios#4968, which adds the same feature to iOS's `CoinDetailScreen` and is being built as the reference implementation (iOS out-of-scope'd Android/Windows parity deliberately, to keep that issue shippable on its own).

## Solution

Add an animated price chart (1D / 1W / 1M / 1Y / ALL) plus market stats, price extremes and token info to `TokenDetailScreen`, sourced from the CoinGecko proxy the app already calls via `CoinGeckoApi.kt` / `TokenPriceRepository.kt`. No new provider, no backend work, no API key.

Match the iOS reference implementation's endpoint choices once it lands (`market_chart` for the id and contract variants, `coins/markets` for stats) — check vultisig-ios#4968 for the exact verified paths, case-sensitivity gotcha on `/coins/{id}` lookups, and the point-count/downsampling behavior of CoinGecko's free tier before implementing, so both clients agree on cache TTLs and range-to-`days` mapping.

Assets priced only via pool data (no CoinGecko `priceProviderId` — THORChain/Maya/Solana pool tokens) won't have a chart or stats source; hide that section for those tokens rather than adding a fallback provider (mirrors iOS's decision).

### Must Do
- [ ] Chart/stats data source layered on `CoinGeckoApi.kt` — id-chart, contract-chart (for tokens without a `priceProviderId`), and markets-stats calls, reusing `TokenPriceRepository`'s existing id/contract resolution logic
- [ ] Cache with sensible TTLs per range (short for 1D, longer for 1Y/ALL) — reuse whatever caching primitive the repo layer already uses for prices
- [ ] ViewModel state for chart / stats / selected range; range switch cancels the in-flight fetch without clearing the currently displayed series
- [ ] Price chart composable (Compose) with range picker (1D/1W/1M/1Y/ALL), drag-to-scrub showing price + date, tinted green/red by the selected range's sign
- [ ] Market stats section (market cap, rank, FDV, 24h volume, circulating/max supply), price extremes (24h low–high, ATH/ATL + dates), token info (contract address, decimals, network, explorer link)
- [ ] Downsample chart series to a bounded point count before rendering (CoinGecko's free tier returns wildly different point counts per range)
- [ ] Loading state (skeleton/shimmer), and a no-chart state for pool-priced tokens with no `priceProviderId`
- [ ] Wire into `TokenDetailScreen`
- [ ] Unit tests: chart decoding, range-to-days mapping, cache hit/refetch/fail-open, id-vs-contract routing, downsampler bounds, ViewModel range-switch cancellation
- [ ] Localization strings for all supported locales

### Must NOT Do
- Don't add a new networking library — use the existing Ktor client / `CoinGeckoApi.kt`
- Don't add a new backend host — route through the existing proxy
- Don't poll — fetch on open and on range change only
- Don't touch existing price-fetch behavior used elsewhere (e.g. balance/portfolio pricing) beyond extracting shared id/contract resolution if needed

### Out of Scope
- Fallback price providers (DeFiLlama, Midgard) for pool-priced assets without a CoinGecko id
- Candlestick/OHLC charts
- Portfolio-level (vault total) charts
- Price alerts / watchlists

## Acceptance Criteria
- [ ] Lint passes with no new warnings
- [ ] Build succeeds
- [ ] `./gradlew test` green, including new tests
- [ ] Opening a token with a CoinGecko id shows a chart defaulting to 1D; switching ranges doesn't collapse the layout to a spinner
- [ ] Dragging across the chart scrubs price + date
- [ ] Chart tint reflects sign of the selected range's change
- [ ] Switching the app currency re-renders chart and stats in the new currency
- [ ] A pool-priced token with no `priceProviderId` renders the screen with chart/stats sections absent — no spinner, no empty box, no crash
- [ ] Localization strings present for all supported locales

rkokhatskyi:
## Summary

- Adds a CoinGecko-backed price chart (1D/1W/1M/1Y/ALL ranges, drag-to-scrub) to the token detail screen
- Adds market stats (market cap, rank, FDV, volume, supply) and price extremes (24h low/high, all-time high/low) sections
- Adds a token info section (contract address, decimals)
- Chart/stats sections are gated to coins with a genuine CoinGecko source (price-provider id, or a contract address on a chain CoinGecko's contract-lookup endpoint actually supports) — pool-priced and unsupported-chain tokens never show an empty/broken chart box
- Chart and stats fetches are cached per (coin, range, currency) with in-flight request coalescing, and fail open to the last-good series on a transient failure without ever relabeling stale data as belonging to a different range or currency
- A CoinGecko response with fewer than 2 price points (new/unindexed coins) is treated the same as no data

## Screenshots from testing

| Screenshot 1 | Screenshot 2 | Screenshot 3 | Screenshot 4 |
| :---: | :---: | :---: | :---: |
| Screenshot 5 | Screenshot 6 | Screenshot 7 | Screenshot 8 |

## Test plan

- [ ] Open a token with a CoinGecko price-provider id or supported contract address — verify the chart loads, range switching works, and drag-to-scrub shows price/date
- [ ] Switch app currency, verify it re-fetches in the new currency
- [ ] Open a pool-priced token (e.g. a THORChain/Maya pool asset) — verify no chart/stats section appears
- [ ] Verify `:app:testDebugUnitTest` and `:data:testDebugUnitTest` pass

## Summary by CodeRabbit

* **New Features**
  * Added an interactive token price chart with selectable time ranges and drag-to-scrub.
  * Added token market data sections: market stats, 24h & all-time price extremes, and token info (including copyable contract address).
  * Migrated the token details screen to a new scaffold layout with updated sections and callbacks.
  * Added localized UI text for the chart/market data, limit-order flow, and send/notification messaging.

* **Bug Fixes**
  * Improved chart and market-stats loading resilience with stale-data fallback and better handling of missing/empty data.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/components/chart/MarketStatsSection.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/chart/MarketStatsSection.kt
@@ -62,7 +62,6 @@ internal fun MarketStatsSection(
         }
     if (rows.isEmpty()) return
 
-    UiSpacer(size = 24.dp)
     Column(modifier = modifier.fillMaxWidth()) {
         Text(
             text = stringResource(R.string.token_details_market_stats_title),
--- a/app/src/main/java/com/vultisig/wallet/ui/components/chart/MarketStatsSection.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/chart/MarketStatsSection.kt
@@ -1,0 +1,84 @@
+package com.vultisig.wallet.ui.components.chart
+
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.dp
+import com.vultisig.wallet.R
+import com.vultisig.wallet.ui.components.UiHorizontalDivider
+import com.vultisig.wallet.ui.components.UiSpacer
+import com.vultisig.wallet.ui.components.v2.containers.TopShineContainer
+import com.vultisig.wallet.ui.components.v2.tokenitem.TokenMetaRow
+import com.vultisig.wallet.ui.models.MarketStatsUiModel
+import com.vultisig.wallet.ui.theme.Theme
+
+/**
+ * Market cap, rank, FDV, 24h volume and supply. While [isLoading] and [stats] is still null, shows
+ * a placeholder skeleton (fixed row set) so the sheet's layout doesn't jump once the fetch
+ * resolves; once loaded, only the fields CoinGecko actually returned are shown. Renders nothing
+ * once loaded if every field came back null.
+ */
+@Composable
+internal fun MarketStatsSection(
+    stats: MarketStatsUiModel?,
+    isLoading: Boolean,
+    modifier: Modifier = Modifier,
+) {
+    val rows: List<Pair<String, String?>> =
+        when {
+            stats != null ->
+                listOfNotNull(
+                    stats.marketCap?.let {
+                        stringResource(R.string.token_details_market_cap) to it
+                    },
+                    stats.marketCapRank?.let {
+                        stringResource(R.string.token_details_market_cap_rank) to it
+                    },
+                    stats.fullyDilutedValuation?.let {
+                        stringResource(R.string.token_details_fully_diluted_valuation) to it
+                    },
+                    stats.volume24h?.let {
+                        stringResource(R.string.token_details_volume_24h) to it
+                    },
+                    stats.circulatingSupply?.let {
+                        stringResource(R.string.token_details_circulating_supply) to it
+                    },
+                    stats.maxSupply?.let { stringResource(R.string.token_details_max_supply) to it },
+                )
+            isLoading ->
+                listOf(
+                    stringResource(R.string.token_details_market_cap) to null,
+                    stringResource(R.string.token_details_market_cap_rank) to null,
+                    stringResource(R.string.token_details_fully_diluted_valuation) to null,
+                    stringResource(R.string.token_details_volume_24h) to null,
+                    stringResource(R.string.token_details_circulating_supply) to null,
+                    stringResource(R.string.token_details_max_supply) to null,
+                )
+            else -> emptyList()
+        }
+    if (rows.isEmpty()) return
+
+    UiSpacer(size = 24.dp)
+    Column(modifier = modifier.fillMaxWidth()) {
+        Text(
+            text = stringResource(R.string.token_details_market_stats_title),
+            style = Theme.brockmann.body.m.medium,
+            color = Theme.v2.colors.text.primary,
+        )
+        UiSpacer(size = 12.dp)
+        TopShineContainer(backgroundColor = Theme.v2.colors.backgrounds.surface1) {
+            Column(horizontalAlignment = Alignment.CenterHorizontally) {
+                rows.forEachIndexed { index, (key, value) ->
+                    TokenMetaRow(key = key, value = value)
+                    if (index != rows.lastIndex) {
+                        UiHorizontalDivider(modifier = Modifier.fillMaxWidth())
+                    }
+                }
+            }
+        }
+    }
+}
--- a/app/src/main/java/com/vultisig/wallet/ui/components/chart/PriceChartView.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/chart/PriceChartView.kt
@@ -2,6 +2,7 @@ package com.vultisig.wallet.ui.components.chart
 
 import androidx.compose.foundation.Canvas
 import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.gestures.detectDragGestures
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
@@ -15,6 +16,7 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.geometry.Offset
@@ -69,8 +71,10 @@ internal fun PriceChartSection(
             points = chart.points,
             isPositive = chart.isPositive,
             isLoading = chart.isLoading,
+            isStale = chart.isStale,
             changePercentText = chart.changePercentText,
             onScrub = { scrubbedPoint = it },
+            onRetry = { onRangeSelected(chart.selectedRange) },
         )
         UiSpacer(size = 16.dp)
         ChartRangePicker(selectedRange = chart.selectedRange, onRangeSelected = onRangeSelected)
@@ -162,8 +166,10 @@ private fun PriceChartCanvas(
     points: List<ChartPointUiModel>,
     isPositive: Boolean,
     isLoading: Boolean,
+    isStale: Boolean,
     changePercentText: String,
     onScrub: (ChartPointUiModel?) -> Unit,
+    onRetry: () -> Unit,
     modifier: Modifier = Modifier,
 ) {
     val tint = if (isPositive) Theme.v2.colors.alerts.success else Theme.v2.colors.alerts.error
@@ -277,6 +283,27 @@ private fun PriceChartCanvas(
                             .clip(RoundedCornerShape(12.dp))
                             .background(placeholderColor)
                 )
+            }
+
+            // A range switch that fails before any data has ever loaded for it lands here
+            // (isStale=true, points cleared) — surface a retry affordance rather than a
+            // silent blank box, since nothing else on screen signals the fetch failed.
+            isStale -> {
+                Box(
+                    modifier =
+                        Modifier.fillMaxWidth()
+                            .height(CHART_HEIGHT)
+                            .clip(RoundedCornerShape(12.dp))
+                            .background(placeholderColor)
+                            .clickable(onClick = onRetry),
+                    contentAlignment = Alignment.Center,
+                ) {
+                    Text(
+                        text = stringResource(R.string.price_chart_retry_hint),
+                        style = Theme.brockmann.body.s.medium,
+                        color = Theme.v2.colors.text.tertiary,
+                    )
+                }
             }
 
             else -> Unit
--- a/app/src/main/java/com/vultisig/wallet/ui/components/chart/PriceChartView.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/chart/PriceChartView.kt
@@ -1,0 +1,320 @@
+package com.vultisig.wallet.ui.components.chart
+
+import androidx.compose.foundation.Canvas
+import androidx.compose.foundation.background
+import androidx.compose.foundation.gestures.detectDragGestures
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberUpdatedState
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Path
+import androidx.compose.ui.graphics.StrokeCap
+import androidx.compose.ui.graphics.StrokeJoin
+import androidx.compose.ui.graphics.drawscope.Stroke
+import androidx.compose.ui.input.pointer.pointerInput
+import androidx.compose.ui.layout.onSizeChanged
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.IntSize
+import androidx.compose.ui.unit.dp
+import com.vultisig.wallet.R
+import com.vultisig.wallet.data.models.ChartRange
+import com.vultisig.wallet.ui.components.UiSpacer
+import com.vultisig.wallet.ui.models.ChartPointUiModel
+import com.vultisig.wallet.ui.models.ChartUiModel
+import com.vultisig.wallet.ui.theme.Theme
+import java.time.Instant
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+import java.time.format.FormatStyle
+import java.util.Locale
+import kotlin.math.roundToInt
+
+private val CHART_HEIGHT = 160.dp
+
+/**
+ * Price chart with a range picker and drag-to-scrub, tinted green/red by [ChartUiModel.isPositive].
+ */
+@Composable
+internal fun PriceChartSection(
+    chart: ChartUiModel,
+    onRangeSelected: (ChartRange) -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    var scrubbedPoint by remember(chart.points) { mutableStateOf<ChartPointUiModel?>(null) }
+
+    Column(modifier = modifier.fillMaxWidth()) {
+        ChartHeader(
+            isPositive = chart.isPositive,
+            changePercentText = chart.changePercentText,
+            scrubbedPoint = scrubbedPoint,
+        )
+        UiSpacer(size = 12.dp)
+        PriceChartCanvas(
+            points = chart.points,
+            isPositive = chart.isPositive,
+            isLoading = chart.isLoading,
+            changePercentText = chart.changePercentText,
+            onScrub = { scrubbedPoint = it },
+        )
+        UiSpacer(size = 16.dp)
+        ChartRangePicker(selectedRange = chart.selectedRange, onRangeSelected = onRangeSelected)
+    }
+}
+
+@Composable
+private fun ChartHeader(
+    isPositive: Boolean,
+    changePercentText: String,
+    scrubbedPoint: ChartPointUiModel?,
+) {
+    val tint = if (isPositive) Theme.v2.colors.alerts.success else Theme.v2.colors.alerts.error
+    Column {
+        if (scrubbedPoint != null) {
+            Text(
+                text = scrubbedPoint.priceText,
+                style = Theme.satoshi.price.title2,
+                color = Theme.v2.colors.text.primary,
+            )
+            Text(
+                text = scrubbedPoint.timestampMillis.toScrubDateText(),
+                style = Theme.brockmann.body.s.medium,
+                color = Theme.v2.colors.text.tertiary,
+            )
+        } else {
+            Text(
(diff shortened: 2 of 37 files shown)


app/src/main/java/com/vultisig/wallet/ui/components/chart/PriceChartView.kt
```diff
@@ -0,0 +1,320 @@
...
+        when {
+            points.size >= 2 -> {
+                var scrubX by remember(points) { mutableStateOf<Float?>(null) }
+                var canvasSize by remember { mutableStateOf(IntSize.Zero) }
+                val latestPoints by rememberUpdatedState(points)
+
+                fun updateScrub(x: Float, width: Float) {
+                    val current = latestPoints
+                    scrubX = x
+                    onScrub(current[nearestPointIndex(x, width, current.size)])
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed — added `rememberUpdatedState(onScrub)` alongside the existing `points` one, so both drag handlers and `updateScrub` read the latest callback. 2088ecb1d.

app/src/main/java/com/vultisig/wallet/ui/models/TokenDetailViewModel.kt
```diff
@@ -171,16 +291,211 @@ constructor(
...
+                val stats = tokenPriceChartRepository.getStats(coin, currency)
+                uiState.update {
+                    it.copy(
+                        statsLoading = false,
+                        marketStats =
+                            stats?.toMarketStatsUiModel(currency.ticker) ?: it.marketStats,
+                        priceExtremes =
+                            stats?.toPriceExtremesUiModel(currency.ticker) ?: it.priceExtremes,
+                    )
+                }
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed — added `lastLoadedStatsCurrency` mirroring `lastLoadedChartKey`; a failed re-fetch now only falls back to held stats when they match the current currency. 2088ecb1d.

data/src/main/kotlin/com/vultisig/wallet/data/api/CoinGeckoApi.kt
```diff
@@ -81,22 +102,48 @@ internal class CoinGeckoApiImpl @Inject constructor(private val http: HttpClient
...
+        http
+            .get(
+                "https://api.vultisig.com/coingeicko/api/v3/coins/" +
+                    "${chain.coinGeckoAssetId}/contract/$contractAddress/market_chart"
+            ) {
+                parameter("vs_currency", currency)
+                parameter("days", days)
+                header("Content-Type", "application/json")
             }
+            .bodyOrThrow()
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed with `appendPathSegments(..., encodeSlash = true)` — worth noting Ktor 3.0.3's `encodeSlash` defaults to false, so the plain suggested diff wouldn't have actually encoded anything; verified against the bytecode and added a regression test with a contractAddress containing '/'. 2088ecb1d.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceChartRepository.kt
```diff
@@ -0,0 +1,142 @@
...
+        val key = ChartCacheKey(coin.id, range, currency.ticker)
+        val result =
+            try {
+                chartCache.getOrPut(key, range.cacheTtlMillis) { fetchChart(coin, range, currency) }
+            } catch (e: CancellationException) {
+                throw e
+            } catch (e: Throwable) {
+                Timber.e(e, "Failed to fetch market chart for %s", coin.id)
+                // Fail open to the last-good series rather than a blank/error chart.
+                chartCache.peekStale(key)
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed — narrowed to `catch (e: NetworkException)` for `getChart`, and `NetworkException` + `NoSuchElementException` for `getStats` (the latter covers CoinGecko's genuine empty-markets-array case, which isn't a failure to fail open from). Verified `bodyOrThrow`/`HttpCallValidator` wrap all real transport failures as `NetworkException`, so a real bug now propagates instead of being swallowed as "stale." Updated the two existing tests to mock realistic `NetworkException` failures and added a test per method proving a non-network exception now propagates.

data/src/main/kotlin/com/vultisig/wallet/data/utils/TtlCache.kt
```diff
@@ -0,0 +1,97 @@
...
+
+        return when (lookup) {
+            is Lookup.Hit -> lookup.value
+            is Lookup.Await -> lookup.deferred.await()
+            is Lookup.Start -> {
+                try {
+                    val value = loader()
+                    mutex.withLock {
+                        entries[key] = Entry(value, now + ttlMillis)
+                        inFlight.remove(key)
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed — `getOrPut` now takes `nowMillis: () -> Long` instead of a single `now: Long` snapshot, and reads it again after `loader()` completes for the entry's `expiresAt`, so a slow fetch no longer shrinks its own cache entry's freshness window. Added a test with a clock that advances mid-load to pin this.

rkokhatskyi:
Addressed the nitpicks from the latest review:

- `TokenMetaRow.kt` — added `verticalAlignment = Alignment.CenterVertically`, matching `TokenMetaAddressRow`.
- `PriceChartView.kt` — deduped `yFor` into a single top-level function both `ChartGeometry.yFor` and `buildChartGeometry`'s path construction call.
- `ChartRangePicker.kt` — switched to `Modifier.selectable` (role + selected + click in one) and added `heightIn(min = 48.dp)` for a proper touch target.
- `TokenDetailViewModel.kt` — renamed the shadowed `it` to `state` in the stats `uiState.update` block.
- `TokenDetailViewModelTest.kt` — dropped the redundant class-level `@OptIn`, renamed the probe test, removed its unreachable stub.

One nitpick declined: moving `MarketStatsSection`/`PriceExtremesSection`'s internal `UiSpacer(24.dp)` back to the caller, to match `PriceChartSection`/`TokenInfoSection`. That asymmetry is intentional, not an oversight — `PriceChartSection`/`TokenInfoSection` are gated by `chart != null` / `tokenInfo != null`, and once those objects exist they always render visible content, so a caller-owned spacer is safe. `MarketStatsSection`/`PriceExtremesSection` are gated by `statsLoading || marketStats != null`, and `marketStats` can be non-null with every field null (CoinGecko returns such entries for stale/inactive coins) — in that case the section renders nothing via its own `rows.isEmpty()` guard. A caller-owned spacer would still fire on that gate, leaving an orphaned 24dp gap — which is the exact bug this internal-spacer placement was added to fix.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceChartRepository.kt
```diff
@@ -0,0 +1,148 @@
...
+            null
+        }
+    }
+
+    private suspend fun fetchStats(coin: Coin, currency: AppCurrency): CoinMarketStats {
+        val json =
+            coinGeckoApi
+                .getMarketStats(coin.priceProviderID, currency.ticker.lowercase())
+                .firstOrNull()
+                ?: throw NoSuchElementException("No market stats for ${coin.priceProviderID}")
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed — added a private `NoMarketStatsException` and catch only that sentinel, so an unrelated bug that happens to throw `NoSuchElementException` (from mapping or anywhere else) now propagates instead of being swallowed. Also good catch by detekt on the unused catch variable, renamed to `_`.

rkokhatskyi:
Ran a dedicated blocking-call/performance pass over this PR's diff (4 independent lenses, each finding adversarially cross-checked 3x): confirmed one real issue — `TokenPriceChartRepositoryImpl.fetchChart()` decoded and sorted the full raw CoinGecko series (up to ~4800 points for the ALL range) inline on the caller's dispatcher, which traces back to `Dispatchers.Main.immediate` with no `withContext` switch anywhere in the chain (ViewModel → repository → TtlCache → API). Fixed by wrapping the decode+downsample in `withContext(Dispatchers.Default)`, matching the pattern already used elsewhere in this codebase (`BackupPasswordViewModel`, `VaultDetailViewModel`) for CPU-bound work. One other candidate finding (fresh `Stroke`/`Color` allocation per draw frame in `PriceChartView.kt`) was checked and rejected — negligible cost, not worth the added complexity.

data/src/main/kotlin/com/vultisig/wallet/data/api/CoinGeckoApi.kt
```diff
@@ -81,22 +103,54 @@ internal class CoinGeckoApiImpl @Inject constructor(private val http: HttpClient
...
-
-                else -> error("No CoinGecko asset id for chain $this")
+    override suspend fun getMarketChart(
+        id: String,
+        currency: String,
+        days: String,
+    ): MarketChartResponseJson =
+        http
+            .get(
+                "https://api.vultisig.com/coingeicko/api/v3/coins/${id.lowercase()}/market_chart"
```
aminsato: `getContractMarketChart` below builds its path with `appendPathSegments(..., encodeSlash = true)`, but `id` is interpolated straight into the path string here.

`priceProviderID` is not always from a hardcoded table — `SPLTokenRepository.kt:132`/`151` take it from the remote Solana token list's `extensions.coingeckoId`, so a list entry containing `/` or `../` retargets the request path on `api.vultisig.com`. Give it the same treatment:

```kotlin
http.get("https://api.vultisig.com/coingeicko/api/v3/coins") {
    url {
        appendPathSegments(listOf(id.lowercase(), "market_chart"), encodeSlash = true)
    }
    parameter("vs_currency", currency)
    parameter("days", days)
    header("Content-Type", "application/json")
}
```

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceChartRepository.kt
```diff
@@ -0,0 +1,161 @@
...
+        currency: AppCurrency,
+    ): MarketChart? {
+        if (!coin.hasMarketDataSource) return null
+        val key = ChartCacheKey(coin.id, range, currency.ticker)
+        val result =
+            try {
+                chartCache.getOrPut(key, range.cacheTtlMillis) { fetchChart(coin, range, currency) }
+            } catch (e: CancellationException) {
+                throw e
+            } catch (e: NetworkException) {
```
aminsato, line 72: `TtlCache` documents that a coalesced follower receives `CoalescedFetchCancelledException` when the initiating fetch is cancelled, specifically so the follower can "fail open". Nothing here does: that type is a plain `Exception`, not a `NetworkException`, so it escapes `getChart` without ever reaching `peekStale` and lands in `loadChart`'s `safeLaunch(onError)`, which calls `toFallback(range, hadPriorPoints, currencyTicker = null)` — clearing the points and marking the chart stale, the opposite of the documented fail-open. `getStats` at line 120 has the same gap.

It also can't be fixed at this call site as written, since `CoalescedFetchCancelledException` is a private nested class of `TtlCache`. Two options: make it public and catch it alongside `NetworkException`, or have `TtlCache` retry the load for the follower rather than failing it when only the initiator was cancelled.

data/src/main/kotlin/com/vultisig/wallet/data/utils/MarketChartDecoder.kt
```diff
@@ -0,0 +1,35 @@
...
+ * the range's headline change. Zero when there are fewer than 2 points or the first price is zero
+ * (avoids a division by zero).
+ */
+fun changePercent(points: List<MarketChartPoint>): Double {
+    val first = points.firstOrNull()?.price ?: return 0.0
+    val last = points.lastOrNull()?.price ?: return 0.0
+    if (first.signum() == 0) return 0.0
+    return last
+        .subtract(first)
+        .divide(first, 10, java.math.RoundingMode.HALF_UP)
```
aminsato: `java.math.RoundingMode` spelled out inline — import it. This file already imports `java.math.BigDecimal` at the top, and the repo style is to import `java.math.*` types rather than fully qualify them at the use site.

data/src/main/kotlin/com/vultisig/wallet/data/utils/ChartDownsampler.kt
```diff
@@ -0,0 +1,21 @@
...
+fun <T> downsampleChartPoints(points: List<T>, maxPoints: Int = DEFAULT_MAX_CHART_POINTS): List<T> {
+    require(maxPoints >= 2) { "maxPoints must be at least 2" }
+    if (points.size <= maxPoints) return points
+
+    val lastIndex = points.size - 1
+    val result =
+        List(maxPoints) { i -> points[(i * lastIndex.toDouble() / (maxPoints - 1)).toInt()] }
+            .toMutableList()
+    result[0] = points[0]
+    result[result.lastIndex] = points[lastIndex]
```
aminsato: These two writes are already what the generator produces, so `.toMutableList()` and both assignments are dead:

- at `i = 0` the index expression is `0 * lastIndex / (maxPoints - 1)` = `0`
- at `i = maxPoints - 1` it is `(maxPoints - 1) * lastIndex / (maxPoints - 1)`, whose exact quotient is `lastIndex` and is representable as a `Double` for any list size here, so correctly-rounded IEEE-754 division returns it exactly

The body reduces to returning the `List(maxPoints) { ... }` directly. If you'd rather keep them as a guard against the float reasoning above, a test asserting `first`/`last` are preserved for a few awkward `(size, maxPoints)` pairs would document that intent better than the silent overwrite.

app/src/main/java/com/vultisig/wallet/ui/components/chart/PriceChartView.kt
```diff
@@ -0,0 +1,320 @@
...
+                Box(
+                    modifier =
+                        Modifier.fillMaxWidth()
+                            .height(CHART_HEIGHT)
+                            .clip(RoundedCornerShape(12.dp))
+                            .background(placeholderColor)
+                )
+            }
+
+            else -> Unit
```
aminsato, line 309: This branch is reachable in a real failure path, not just as a no-op default. When a range switch fails before any data has ever loaded for that range, `toFallback` sets `isStale = true` and empties `points` while leaving `chart` non-null, so the user gets: a 160dp blank box here, a `ChartHeader` rendering `" "` because `changePercentText` was also emptied, and a range picker still highlighting the range that failed.

Nothing on screen says the fetch failed or that re-tapping the range retries — which is the retry affordance `onChartRangeSelected`'s `isStale` check was built for. Render an error/retry state for `points.isEmpty() && !isLoading` instead of `Unit`.

app/src/main/java/com/vultisig/wallet/ui/components/chart/MarketStatsSection.kt
```diff
@@ -0,0 +1,84 @@
...
+                    stringResource(R.string.token_details_fully_diluted_valuation) to null,
+                    stringResource(R.string.token_details_volume_24h) to null,
+                    stringResource(R.string.token_details_circulating_supply) to null,
+                    stringResource(R.string.token_details_max_supply) to null,
+                )
+            else -> emptyList()
+        }
+    if (rows.isEmpty()) return
+
+    UiSpacer(size = 24.dp)
```
aminsato: This leading spacer is emitted before the `Column`, so it sits outside the caller-supplied `modifier` and lands directly in the parent's layout. It's also inconsistent with the other three sections: in `TokenDetailsContent` the 24dp spacers for `PriceChartSection` and `TokenInfoSection` come from the call site, while this one and `PriceExtremesSection.kt:56` bake it in — so spacing for the four sections is set in two different places.

Move it to the call site with the others, guarded by the same `statsLoading || marketStats != null` condition that already wraps the call.

aminsato (commented):
Reviewed the full diff. The architecture holds up well: the `hasMarketDataSource` gate is the right place to keep pool-priced and unindexed-chain tokens from ever rendering a broken chart box, the `(range, currency)` keying on both the cache and `lastLoadedChartKey` genuinely prevents stale data being relabelled, moving decode/downsample to `Dispatchers.Default` is the correct call for the ALL range, and the test coverage across `TtlCache`, the downsampler, the decoder, the repository and the ViewModel is thorough. I verified every id in `coinGeckoAssetPlatformId()` against CoinGecko's `/asset_platforms` — all 13 including `robinhood` are real platform ids.

Six inline points below, two of which are worth fixing before merge (the unencoded path segment and the coalesced-cancellation gap).

johnnyluo (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/api/CoinGeckoApi.kt
rkokhatskyi: Good catch — priceProviderID genuinely isn't always from a hardcoded table, SPLTokenRepository.kt:132/151 do pull it from a remote Solana token list's extensions.coingeckoId. Applied the same appendPathSegments(..., encodeSlash = true) treatment to getMarketChart, and added a regression test with an id containing '/'. Also fixed the KDoc above it while I was there — it still claimed callers must lowercase id, which contradicted the id.lowercase() already in the implementation. e7f23cbea.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceChartRepository.kt
rkokhatskyi: Confirmed, and it wasn't fixable at this call site as you noted since CoalescedFetchCancelledException is private. Went with your second option: TtlCache's Await branch now catches it directly and retries via a recursive getOrPut call rather than propagating it — a follower whose leader got cancelled just becomes a fresh attempt (or coalesces with a newer one), instead of failing on an outcome that was never really its own. Updated the existing cancellation test to assert the follower now succeeds via retry instead of just 'fails with something that isn't CancellationException'. e7f23cbea.

data/src/main/kotlin/com/vultisig/wallet/data/utils/MarketChartDecoder.kt
rkokhatskyi: Fixed — imported RoundingMode. e7f23cbea.

data/src/main/kotlin/com/vultisig/wallet/data/utils/ChartDownsampler.kt
rkokhatskyi: Your math checks out — verified both index expressions are exact under IEEE-754 for any realistic point count. Removed the overwrites and .toMutableList(), and went with your suggested test to document the intent: added a case covering several deliberately awkward (size, maxPoints) pairs (including CoinGecko's own documented raw/target sizes for ALL/1W) asserting first/last are always exact. e7f23cbea.

app/src/main/java/com/vultisig/wallet/ui/components/chart/PriceChartView.kt
rkokhatskyi: Agreed, this was a real gap in the retry affordance onChartRangeSelected's isStale check was built for. Added an isStale branch rendering a retry hint (re-taps the currently-selected range via the same onRangeSelected callback the range tabs use), instead of falling through to the blank else -> Unit. e7f23cbea.

app/src/main/java/com/vultisig/wallet/ui/components/chart/MarketStatsSection.kt
rkokhatskyi: Moved it to the call site as you suggested, but changed the gating condition rather than keeping the bare statsLoading || marketStats != null: added MarketStatsUiModel/PriceExtremesUiModel.hasAnyValue() and gate on that instead. Reason: marketStats can be non-null with every field null (CoinGecko returns such an entry for stale/inactive coins), and the bare non-null check you quoted would let the caller-owned spacer fire while the section itself renders nothing via its own rows.isEmpty() guard — the exact orphaned-gap bug that motivated putting the spacer inside in the first place. This gets the consistency you wanted (one placement mechanism, same as chart/token-info) without reintroducing that. e7f23cbea.

PR #5464, opened 2026-07-30
feat(limit-swap): limit orders in TX History + cancel a resting order

aminsato:
## Summary

A placed THORChain limit order had nowhere to live in the app. The inbound deposit landed in TX History as an ordinary swap row — which confirms within minutes — while the order itself could rest unfilled for up to three days, and there was no way to close it early. Phase 1 (#5395) recorded orders into a write-only table; this reads that table back and adds the cancel.

Two user-facing changes:

- **A Limit Orders tab in TX History**, backed by THORChain's advanced-swap queue: live status, target price, expiry countdown and partial-fill split, for open and closed orders alike.
- **Cancel order**, from the card, routed through the existing deposit verify → keysign flow.

### Transaction Hash
https://mempool.space/api/address/bc1qrncuculen6deh35at408fv95ng9kx3ve70gjjx/txs/mempool

## Tracking

`queue/limit_swaps?sender=` returns every resting order for an address in one request, so the tracker polls a **list, not a hash**, and an order's **absence** is the only terminal signal it gets.

That signal is not always true — the queue is reached through a load-balancing gateway whose backends are not always in sync, and one answering "no resting orders" for an order that never left is a well-formed, present response nothing at the decoding layer can tell from a real closure. So:

- a closure is acted on only after **two consecutive polls** agree;
- a **reappearance resets the streak** (reappearing is the stale-backend signature itself);
- a response **missing the `limit_swaps` key** is treated as "no information", never as an empty queue — reading it as empty would close every one of a vault's orders at once on the strength of a response we did not understand.

The queue never says *why* an order left, so the outcome is resolved separately from Midgard's refund reason (`limit swap expired` / `limit swap cancelled`, THORChain's own words). Anything unrecognised stays `refunded` — the observable fact, with no cause attached — rather than being labelled expired or cancelled on a guess. Nothing revisits a terminal order, so a guess there is permanent.

## Cancelling

`m=<:<amt><SRC>:<target><TGT>:0`, sent either as a `MsgDeposit` (THORChain-funded order) or as a memo-bearing dust transfer to the Asgard inbound (any other routable source), then through the existing deposit verify → keysign pipeline unchanged.

Three details are load-bearing, because each fails **silently** — the chain accepts the memo, charges for it, and cancels nothing:

| | Why |
|---|---|
| Assets spelled in **full** | `m=<` is the one inbound memo type THORChain does not run through `fuzzyAssetMatch`, so the placement memo's 6-character contract suffix keys a bucket no order was indexed under. The long form is captured at signing, because the abbreviation cannot be reversed. |
| Both amounts **exact** | THORNode addresses an order by a bucket key derived from `(sourceAmount × 1e8) / tradeTarget`. One unit of drift lands in a different bucket. |
| L1 dust clears **both** floors | The signer's own floor is in native units; THORChain's `dust_threshold` is in *its* 1e8 fixed point on every chain. Rescaling is the entire difference between 2e13 wei and 2000 wei — and the 8-decimal UTXO chains cannot show the bug. A ceiling bounds how much a remote value can cause to be donated, since nothing attached to an `m=<` is refundable. |

Eligibility fails closed at every unknown, and cross-checks what was signed against what the queue reports — **amounts and assets** — blocking on a disagreement rather than signing one of two values that cannot both be right. A blocked order says *why* on the card instead of silently withholding the button.

A broadcast cancel moves the order to the non-terminal `cancelling`: that is a claim about *our* transaction, never about the order. THORChain accepts a cancel that matches nothing, so the order keeps its place in the list, keeps its countdown, and keeps being polled — and a cancel the chain later refuses has its record withdrawn, so the button comes back rather than the order being pre-labelled closed. Orders that share a ratio bucket are flagged, because THORChain modifies the first match and the one the user tapped may not be the one that closes.

The `EnableAdvSwapQueue` mimir gate deliberately does **not** apply to a cancel. Placement is gated and must be — a `=<` signed while the queue is off can execute as an unprotected market swap. A cancel has the opposite risk profile: if the mimir flips off while an order rests, the order still exists and the user still needs a way out.

## Changes

**Data**
- `LimitSwapCancelMemo` — the `m=<` builder, the abbreviation guard, and THORNode's queue-index bucket key (18-char ratio normalisation reproduced, pad *and* truncate).
- `LimitOrderCancelDust` — the two floors, the 1e8 ↔ native rescale, the observable minimum and the per-chain ceiling.
- `LimitOrderCancelEligibility` — the fail-closed gate, plus duplicate-bucket detection and memo→order attribution.
- `RefreshLimitOrdersUseCase` — queue reconciliation, absence corroboration, cancel verification/self-heal.
- `MidgardLimitOutcomeResolver`, `ThorchainLimitSwapQueue` wire models (decodes an asset in both its flat-string and protobuf-JSON shapes), `dust_threshold` on `THORChainInboundAddress`.
- `pending_limit_order` widened (migration 39 → 40, all columns nullable/defaulted so existing rows survive; a null cancel input leaves an older order uncancellable rather than cancelled with guessed values).

**UI**
- `LIMIT` tab in `TransactionHistoryScreen` + `LimitOrderCard`; `BuildLimitOrderCancelTransactionUseCase`.
- Strings added to all 10 locales.

## Test plan

- `:data:testDebugUnitTest` and `:app:testDebugUnitTest` pass. 48 new tests covering the memo grammar, the dust math, the eligibility gate, and the tracker's state machine — including the absence-corroboration and never-close-on-an-unknown-envelope rules.
- `:app:lintDebug` passes.
- `:data:lintDebug` passes (it caught a `BigInteger.TWO` — API 33 against minSdk 26 — that the `:app` run could not see).
- Screens captured on the emulator, below. A mainnet cancel rehearsal on both routes (THORChain `MsgDeposit` and an L1 dust send) is still worth doing before this ships behind the flag.

## Known follow-up

The inbound deposit still appears under Swaps/Overview and will read "Successful" once it confirms, even while the order rests. That row is accurate about the *transaction*, and the Limit Orders tab now carries the order's real state — but arbitrating the swap row's status against the queue is a separate change to the tracking-provider gate and is deliberately not in this PR.

---

## Update — first on-device run found a real bug

Cancelling a BTC-funded order failed with a generic `Failed to broadcast transaction`. The node's actual answer:

```
insufficient fee, rejecting replacement <txid>;
new feerate 0.00005000 BTC/kvB <= old feerate 0.00005000 BTC/kvB
```

The order's **placement deposit was still in the mempool**. Blockchair only marks an output spent once the spending tx confirms, so the cancel was built from a UTXO set that still listed the inputs the pending placement was spending — the two transactions conflict, and at an equal fee rate the node rejects the cancel. At a *higher* rate it would have been worse: the cancel would have **replaced the placement**, destroying the order by a mechanism the user never asked for and that no screen would have explained.

**The gate is the fix, not a fee bump.** An order THORChain has not observed resting has nothing for a `m=<` to match in the first place, so cancelling it could never have worked on any chain — the UTXO conflict is just the loudest symptom. A new `PlacementNotObserved` blocker keys off the first queue observation, so it clears itself on the next poll and says so on the card instead of greying the button out in silence.

Two things the bug also exposed, both fixed:

- BlockChair's mempool broadcast swallowed the node's rejection and threw a fixed string. These reasons are specific and actionable on every such failure, and the thrown message is all the user ever sees — it now carries the body.
- The affordability pre-check compared the balance against `gasFee`, which on a UTXO chain is a per-byte **rate**, not a total. It now prices the real transaction plan over the exact inputs and OP_RETURN the cancel will carry.

## Update — cancel presentation, ported from iOS

A cancel used to render through the generic deposit vocabulary — "You're sending 0 RUNE", or "You're sending 0.00002 BTC" of dust that is about to be donated. Matching [vultisig-ios#4857](https://github.com/vultisig/vultisig-ios/commit/105c0573824a32afa00b21cdc47e745c2a89bd55), including its copy:

- **Verify** leads with "You're cancelling a limit order", the order's `SRC → TGT` pair, and what cancelling actually does to the filled and unfilled parts.
- **Done** reads "Cancellation **sent** — Your order stays open until THORChain confirms it closed."

"Sent", never "complete", and that is the whole point: THORChain accepts a cancel that addresses the wrong ratio bucket, charges for it, and closes nothing. Only the chain's own account of the closure, read by the tracker, may promote an order to Cancelled.

Both are keyed off the `m=<` memo rather than a transaction type, so a co-signer — which never sees the initiator's transaction, only the payload it is asked to sign — reads exactly the same thing.

Strings added to all 10 locales, lifted from the iOS translations where they exist (de/es/hr/it/ko/pt/zh); nl and ru written to match the terminology already in those files.

Previews for screenshots: `limit_order_cancel_verify`, `limit_order_cancel_done`, `limit_orders_tab`, `limit_orders_tab_empty`.

## Screens

Real emulator renders via `PreviewActivity` (`limit_orders_tab`, `limit_order_cancel_verify`, `limit_order_cancel_done`).

| Limit Orders tab | Cancel — Verify | Cancel — Done |
|---|---|---|
| tab | verify | done |

The tab shows the three states that matter together: an open order with its countdown and partial fill and a live Cancel; one mid-cancellation, styled as in-flight rather than done and explaining why it stays open; and a closed one.

## Summary by CodeRabbit

* **New Features**
  * Added a **Limit** tab to Transaction History with limit-order cards (status, expiry, and fill progress).
  * Added THORChain limit-order tracking and a **cancel** flow, including verification and completion screens.

* **Bug Fixes**
  * Improved limit-order cancellation eligibility and clearer blocked/error messaging.
  * Refreshed limit orders and pending transactions independently to prevent cross-impact on failures.
  * Improved cancel fee affordability calculations for UTXO chains.

* **Documentation/Localization**
  * Added/updated localized UI strings for the limit-orders experience, including cancel/verification/completion and status labels.


app/src/main/java/com/vultisig/wallet/ui/models/limitorder/BuildLimitOrderCancelTransactionUseCase.kt
```diff
@@ -0,0 +1,307 @@
...
+                )
+                .fee
+                .toBigInteger()
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.w(e, "Could not plan the cancel transaction on %s; pricing off the rate", chain)
+            gasFee.value
+        }
+    }
```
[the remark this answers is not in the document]
aminsato: Valid, fixed in c2671aee6.

Verified the premise first: `UtxoHelper.planTransaction` returns `AnySigner.plan(...)` unchecked, so a short coin selection comes back as a `TransactionPlan` carrying `Error_not_enough_utxos` with **fee = 0** rather than throwing. That zero flowed straight into the affordability comparison and made any balance look sufficient — so an unaffordable cancel passed the guard, went through the whole signing ceremony, and failed at broadcast. Same shape as the bug this pre-check exists to prevent.

`totalFee` now draws the distinction the original code missed — between a planner that **blew up** and one that **answered**:

- exception → still falls back to the rate. It told us nothing, and refusing a cancel on a failed local estimate is worse than under-stating a fee the node has the last word on anyway.
- `plan.error != SigningError.OK` or `utxosCount == 0` → `InsufficientBalance`. That is not a failed estimate, it is the answer.

Structured so the plan call sits in its own `try` and the error check happens after it, otherwise the generic `catch` would have swallowed the exception the check raises.

data/src/main/kotlin/com/vultisig/wallet/data/swap/limit/LimitOrderCancelEligibility.kt
```diff
@@ -0,0 +1,301 @@
...
+    if (LimitOrderStatus.fromRaw(order.status).isTerminal) {
+        return LimitOrderCancelEligibility.Blocked(LimitOrderCancelBlocker.Terminal)
+    }
+    if (order.cancelBroadcastHash != null) {
+        return LimitOrderCancelEligibility.Blocked(LimitOrderCancelBlocker.CancelAlreadyBroadcast)
+    }
+    // The order has to be RESTING before it can be cancelled — see [PlacementNotObserved]. Any
+    // field the queue writes proves it was there; `expiryObservedAt` is the one written on every
+    // resting poll, so it is the honest witness rather than a field that only some responses carry.
+    if (order.expiryObservedAt == null && order.depositAmount == null) {
```
johnnyluo: **Worth confirming before merge:** this gate clears `PlacementNotObserved` as soon as `expiryObservedAt` is non-null, but `RefreshLimitOrdersUseCase`/the DAO write `expiry_observed_at` independently of `depositAmount`/`observedTradeTarget` (both come off the nullable `state` block on the queue entry, see `observeResting`). If a genuinely resting order's queue entry ever lacks a `state` block, this gate clears purely off `expiryObservedAt` while `depositAmount`/`observedTradeTarget` stay permanently null — which means the amount/asset cross-check at lines 184/189 (the PR's headline safety net against signing a doomed cancel) never has data to run against, and the cancel proceeds on signed-only data with zero on-chain corroboration. Can we confirm `state` is always populated on a real resting entry from THORNode?

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RefreshLimitOrdersUseCase.kt
```diff
@@ -0,0 +1,290 @@
...
+            // Reading it as empty would close every one of this sender's orders at once.
+            Timber.e("Limit queue response carried no limit_swaps key — treating as unknown")
+            return
+        }
+
+        // Hex case is not semantic, and the queue's casing need not match the hash we broadcast
+        // under, so match case-insensitively.
+        val restingByHash = resting.associateBy { it.swap.tx.id.uppercase() }
+
+        orders.forEach { order ->
```
johnnyluo, line 148: Minor: the network fetch above (line 85) is try/caught so one sender's failure doesn't abort others, but this `orders.forEach` isn't — `recordStatus`/`confirmCancelBroadcast`/`clearCancelBroadcast` (plain Room suspend calls reached via `observeResting`/`observeAbsent`/`verifyPendingCancel`) are unguarded. A DB exception here would propagate out of `invoke(vaultId)` and abort processing of any remaining `sourceAddress` groups in the outer `groupBy(...).forEach`, which is inconsistent with the deliberate per-sender fault isolation the rest of the class provides.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RefreshLimitOrdersUseCase.kt
```diff
@@ -0,0 +1,290 @@
...
+     * Everything else stays [LimitOrderStatus.Refunded] — the observable fact, with no cause
+     * attached. An order rejected at placement (halted pool, bad memo) also refunds, seconds in,
+     * and inventing a cause here would be inventing it for that user too.
+     */
+    private fun refundedOrCredited(order: PendingLimitOrderEntity): LimitOrderStatus {
+        if (order.cancelBroadcastHash == null || !order.cancelConfirmed) {
+            return LimitOrderStatus.Refunded
+        }
+        val elapsedMs = System.currentTimeMillis() - order.createdAt
+        val ttlMs = order.expiryBlocks.toLong() * THORCHAIN_BLOCK_MS
```
johnnyluo, line 269: Nit: `ttlMs = order.expiryBlocks * THORCHAIN_BLOCK_MS` assumes an exact 6s block time to decide whether a no-reason refund should read as `Cancelled` vs `Refunded`. Real block time drifts, so near the TTL boundary this heuristic could misclassify one for the other. Low impact — cosmetic status label only, not funds — but might be worth a short comment acknowledging the approximation, similar to the reasoning already given elsewhere in this file for the unrecognised-reason default.

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -237,6 +263,7 @@ constructor(
             _uiState.update { it.copy(isRefreshing = true) }
             try {
                 refreshPendingTransactions(vaultId, chainId)
+                refreshLimitOrders(vaultId)
```
johnnyluo: Minor: `refreshOnEnter()` (line ~278) fires `refreshPendingTransactions` and `refreshLimitOrders` as two independent `safeLaunch` calls specifically so one failing doesn't take the other down (per the comment there). This `refresh()` (manual pull-to-refresh) instead runs them sequentially inside one `try` — if `refreshPendingTransactions` throws, `refreshLimitOrders` for that pull never runs. Worth splitting the same way, or a comment noting this is an accepted inconsistency for pull-to-refresh specifically.

app/src/main/java/com/vultisig/wallet/ui/models/limitorder/BuildLimitOrderCancelTransactionUseCase.kt
```diff
@@ -0,0 +1,321 @@
...
+        decimals: Int,
+        inboundDustThreshold: String?,
+    ): BigInteger =
+        try {
+            limitOrderCancelDustAmount(
+                localDustFloor = limitOrderCancelLocalDustFloor(chain),
+                inboundDustThreshold = inboundDustThreshold,
+                decimals = decimals,
+                ceiling =
+                    limitOrderCancelDustCeiling(chain).movePointRight(decimals).toBigIntegerExact(),
```
johnnyluo, line 220: Nit: `limitOrderCancelDustCeiling(chain).movePointRight(decimals).toBigIntegerExact()` throws `ArithmeticException` if a chain's ceiling `BigDecimal` scale doesn't divide evenly into its native decimals. Not reachable with today's chain table, but it isn't caught by the surrounding `catch (e: LimitOrderCancelDustError)`, so it would degrade to a generic/unhandled error instead of the intended `DustUnavailable`. A future chain addition with an awkward decimal scale could regress this silently — worth a defensive catch or switching to explicit integer arithmetic.

app/src/test/java/com/vultisig/wallet/ui/models/transaction/TransactionHistoryViewModelTest.kt
```diff
@@ -86,6 +103,11 @@ internal class TransactionHistoryViewModelTest {
             savedStateHandle = SavedStateHandle(),
             transactionHistoryRepository = transactionHistoryRepository,
             refreshPendingTransactions = refreshPendingTransactions,
+            pendingLimitOrderRepository = pendingLimitOrderRepository,
+            refreshLimitOrders = refreshLimitOrders,
+            mapLimitOrderToUiModel = mapLimitOrderToUiModel,
+            buildLimitOrderCancelTransaction = buildLimitOrderCancelTransaction,
```
johnnyluo, line 135: Test coverage gap: this file only received constructor-wiring changes for the new limit-order dependencies — no test exercises `cancelLimitOrder()`, `dismissCancelError()`, `observeLimitOrders()`'s state update, or the LIMIT tab's filtering/mapping. That's a contrast with the very thorough data-layer coverage (48 new tests) elsewhere in this PR. Since cancel routes a real signed/broadcast transaction, a happy-path test plus one blocked-eligibility test at this layer would be worth adding before merge.

johnnyluo (commented):
Reviewed via a 3-way split (core money/state-machine logic, UI/ViewModel layer, tests+localization). Overall: unusually careful, deliberately fail-closed money-handling code, well tested at the data layer. Left 6 inline notes below — one worth explicit confirmation before merge (eligibility cross-check), the rest are minor robustness/coverage gaps. No blockers found beyond the mainnet cancel rehearsal the PR description already calls out as a pre-ship gate.

data/src/main/kotlin/com/vultisig/wallet/data/swap/limit/LimitOrderCancelDust.kt
```diff
@@ -0,0 +1,218 @@
...
+        Chain.BitcoinCash,
+        Chain.Dash,
+        Chain.Zcash -> BigDecimal("0.001")
+        Chain.GaiaChain,
+        Chain.Noble,
+        Chain.Kujira -> BigDecimal("0.5")
+        // ETH (0.00002 attach), BSC (0.0002) and anything else. Immaterial in fiat on every
+        // supported chain while still leaving room for a threshold that moves by an order of
+        // magnitude.
+        else -> BigDecimal("0.001")
```
rkokhatskyi, line 218: THORChain publishes `dust_threshold` 1e8 for XRP, 1e7 for TRON and 1e5 for SOL, so the doubled attach lands 2000x, 200x and 2x above this `0.001` ceiling and every cancel from those chains dies as `ExceedsCeiling` while the card still shows a live Cancel button

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RefreshLimitOrdersUseCase.kt
```diff
@@ -0,0 +1,290 @@
...
+            Timber.e("Limit queue response carried no limit_swaps key — treating as unknown")
+            return
+        }
+
+        // Hex case is not semantic, and the queue's casing need not match the hash we broadcast
+        // under, so match case-insensitively.
+        val restingByHash = resting.associateBy { it.swap.tx.id.uppercase() }
+
+        orders.forEach { order ->
+            val entry = restingByHash[order.inboundTxHash.uppercase()]
```
rkokhatskyi: EVM broadcast hashes are stored `0x`-prefixed by `EvmApi.sendTransaction` while THORNode reports `swap.tx.id` as bare 64 char hex, so this lookup never matches an ETH, AVAX, BSC or BASE funded order and Cancel stays blocked on `PlacementNotObserved` forever

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/MidgardLimitOutcomeResolver.kt
```diff
@@ -0,0 +1,122 @@
...
+ * attached. An order rejected at placement (halted pool, malformed memo) also refunds seconds in
+ * with no TTL elapsed, so inventing a cause here would be inventing it for that user too.
+ */
+class MidgardLimitOutcomeResolver @Inject constructor(private val httpClient: HttpClient) {
+
+    suspend fun resolveOutcome(inboundTxHash: String): LimitOrderOutcome =
+        try {
+            val response: MidgardLimitActionsResponse =
+                httpClient
+                    .get(THORCHAIN_MIDGARD_ACTIONS_URL) { parameter(TXID_PARAM, inboundTxHash) }
```
rkokhatskyi: Midgard returns zero actions for a `0x`-prefixed txid and one for the bare form, so an EVM order's stored hash resolves no outcome and the order stays resting forever. iOS strips the prefix in `THORChainTransactionStatusAPI.swift:59`

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/MidgardLimitOutcomeResolver.kt
```diff
@@ -0,0 +1,122 @@
...
+            throw e
+        } catch (e: Exception) {
+            Timber.w(e, "Limit-order outcome lookup failed for %s", inboundTxHash)
+            LimitOrderOutcome.Unresolved
+        }
+
+    /** Null when this action says nothing about the order's fate. */
+    private fun classify(action: MidgardLimitAction): LimitOrderOutcome? =
+        when (action.type) {
+            ACTION_TYPE_SWAP ->
```
rkokhatskyi: Midgard types a limit fill as `limit_swap`, not `swap`, and many filled orders return that action alone with no companion `swap`, so `classify` yields null and the order stays `Unresolved` and resting long after it actually filled

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RefreshLimitOrdersUseCase.kt
```diff
@@ -0,0 +1,290 @@
...
+     * Until then the order keeps whatever state it already has: nothing is written, so the last
+     * good resting observation — its fill split and expiry countdown — is left exactly as it was,
+     * and no third "possibly gone" state is invented.
+     */
+    private suspend fun observeAbsent(order: PendingLimitOrderEntity) {
+        // Clamped once corroborated: an order can stay absent for many polls without being released
+        // — an outcome Midgard has not indexed yet — and the streak is only ever compared against
+        // the threshold, so counting past it measures nothing and grows without bound.
+        val streak =
+            absentPollStreaks.compute(order.inboundTxHash) { _, current ->
```
rkokhatskyi, line 217: `refreshOnEnter` never sets `isRefreshing`, so a pull to refresh during the init poll runs two concurrent invocations of this singleton and both increment the same streak, letting one stale backend round reach the threshold of 2 and close a live order permanently

app/src/main/java/com/vultisig/wallet/ui/models/limitorder/BuildLimitOrderCancelTransactionUseCase.kt
```diff
@@ -0,0 +1,321 @@
...
+        memo: String,
+    ) {
+        val balance =
+            accountsRepository
+                .loadAddress(vaultId, chain)
+                .first()
+                .accounts
+                .firstOrNull { it.token.isNativeToken }
+                ?.tokenValue
+                ?.value ?: return
```
rkokhatskyi: `loadAddress(...).first()` is the mapper's first emission, where `ChainAndTokensToAddressMapper` builds every account with `tokenValue = null`, so the elvis always returns and the whole affordability guard never runs, including the `plan.error` check added in c2671ae

app/src/main/java/com/vultisig/wallet/ui/models/limitorder/BuildLimitOrderCancelTransactionUseCase.kt
```diff
@@ -0,0 +1,321 @@
...
+        val gasFee =
+            depositGasFeeHelper.calculateGasFee(
+                vaultId = vaultId,
+                chain = sourceChain,
+                token = signingCoin,
+                srcAddress = signingCoin.address,
+            )
+
+        val isThorchainSourced = sourceChain == Chain.ThorChain
+        val destination = if (isThorchainSourced) null else resolveInboundVault(sourceChain)
```
rkokhatskyi, line 129: For any non-THORChain source this keeps the payload send shaped, so `JoinKeysignViewModel.kt:654` returns false and the co-signer gets the plain send screen with no cancel title, pair or dust disclosure. iOS keys that off the memo on the join screen instead

app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt
```diff
@@ -82,58 +87,74 @@ internal fun TransactionDoneView(
...
-                                    shape = RoundedCornerShape(24.dp),
+                Column(
+                    modifier = Modifier.verticalScroll(rememberScrollState()).fillMaxWidth(),
+                    verticalArrangement = Arrangement.spacedBy(16.dp),
+                ) {
+                    // A cancel is the one completed transaction whose success is NOT the user's
+                    // answer: THORChain accepts a cancel that matches no order, so the honest
+                    // headline is that the request was sent, and the order stays open until the
+                    // chain says otherwise. Keyed off the memo so a co-signer reads the same thing.
+                    limitCancelPair?.let { pair ->
```
rkokhatskyi, line 98: A cancel always resolves to `TransactionTypeUiModel.Deposit`, which `Keysign.kt:147` routes to `SendTxOverviewScreen`, so `TransactionDoneView` never receives it and this header plus the `limit_order_cancel_done_*` strings never render in the app

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RefreshLimitOrdersUseCase.kt
```diff
@@ -0,0 +1,290 @@
...
+    private val settledCancelHashes = ConcurrentHashMap.newKeySet<String>()
+
+    suspend operator fun invoke(vaultId: String) {
+        val open = repository.getOpenOrders(vaultId)
+        if (open.isEmpty()) return
+
+        open
+            .groupBy { it.sourceAddress }
+            .forEach { (sender, orders) ->
+                if (sender.isNullOrBlank()) {
```
rkokhatskyi, line 115: Orders placed by phase 1, which shipped in `v1.0.116`, have `source_address` NULL and `MIGRATION_39_40` backfills nothing, so this skips them on every poll while the new tab renders them as Open with a reason promising they become cancellable once the deposit confirms

app/src/main/java/com/vultisig/wallet/ui/models/limitorder/LimitOrderHistoryUiModel.kt
```diff
@@ -0,0 +1,201 @@
...
+                    ?.takeUnless { it == LimitOrderCancelBlocker.Terminal }
+                    ?.toUiText(),
+            hasCancelDuplicate =
+                isCancellable && duplicateRestingLimitOrders(order, all).isNotEmpty(),
+        )
+    }
+
+    private fun formatSourceAmount(order: PendingLimitOrderEntity): String {
+        val amount = order.sourceAmount.toBigIntegerOrNull() ?: return order.sourceAmount
+        val decimals = order.sourceDecimals ?: return order.sourceAmount
```
rkokhatskyi: `source_decimals` is NULL for the phase 1 rows already shipped in `v1.0.116` and `MIGRATION_39_40` adds no backfill, so this returns `sourceAmount` verbatim and a 0.5 BTC order renders as `50000000 BTC` beside a correct target price

app/src/main/java/com/vultisig/wallet/ui/models/limitorder/LimitOrderHistoryUiModel.kt
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/**
+ * How much of the deposit has been swapped, as a whole percentage — or null when nothing has been
+ * observed or nothing has filled.
+ *
+ * An order fills via streaming sub-swaps, so a partial fill is a real, stable state and the split
+ * is what explains a two-leg settlement: part paid out in the target asset, the remainder refunded.
+ */
+private fun PendingLimitOrderEntity.fillPercent(): Int? {
```
rkokhatskyi, line 229: `fillPercent` has no terminal status gate, unlike `expiryText` directly above it, and `recordStatus` never clears the fill columns, so a closed order keeps its last resting snapshot and the card reads `Filled` alongside `Filled 40%`

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -246,20 +273,92 @@ constructor(
...
+     * whether it is still live, and the inbound deposit that placed it confirms within minutes
+     * regardless.
+     *
+     * Not filtered by [chainId]: an order is a claim about a PAIR, and hiding one because the user
+     * opened history from the destination chain would hide exactly the order they came looking for.
+     */
+    private fun observeLimitOrders() {
+        viewModelScope.launch {
+            pendingLimitOrderRepository.observeOrders(vaultId).collect { orders ->
+                val uiModels = mapLimitOrderToUiModel.map(orders, System.currentTimeMillis())
```
rkokhatskyi: `nowMs` is sampled only when the Room flow re-emits and the only writers are the on enter poll and pull to refresh, so the countdown baked in at `LimitOrderHistoryUiModel.kt:170` freezes and a card keeps reading `Expires in 4m` after the order expired

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TransactionHistoryScreen.kt
```diff
@@ -191,23 +213,68 @@ private fun TransactionHistoryScreen(
...
-                            Modifier.fillMaxWidth().padding(top = 27.dp, start = 16.dp, end = 16.dp)
-                    )
-                } else {
-                    TransactionGroupedList(
-                        groups = state.groups,
-                        onItemClick = onItemClick,
-                        modifier = Modifier.fillMaxSize(),
-                    )
+                when {
+                    state.selectedTab == TransactionHistoryTab.LIMIT ->
```
rkokhatskyi, line 222: The asset filter chips and search entry render above this on every tab, but the LIMIT branch renders `state.limitOrders` with no `selectedAssetIds` filtering, so filtering to ETH still lists every RUNE to BTC order under an active ETH chip

app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt
```diff
@@ -146,10 +146,35 @@ internal fun VerifyDepositScreen(
...
 
                     SwapToken(valuedToken = tx.token, isLoading = state.isLoading)
 
+                    // Stated before signing, not after: cancelling refunds only what has not
+                    // filled,
+                    // and on an L1 route the amount above is donated with no refund path.
+                    if (tx.limitCancelPair != null) {
+                        UiSpacer(12.dp)
+                        Text(
+                            text = stringResource(R.string.limit_order_cancel_explanation),
```
rkokhatskyi: On the L1 route the amount above this is dust THORChain donates to the pool with no refund path, 2 DOGE on Dogecoin, while this is the only cost sentence shown and it says the cancel costs one THORChain network fee

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TransactionHistoryScreen.kt
```diff
@@ -161,6 +177,12 @@ private fun TransactionHistoryScreen(
                             onClick = { onTabSelected(TransactionHistoryTab.SEND) },
                         )
                     }
+                    tab {
```
rkokhatskyi: Placing a limit order needs both the remote `limit-swap` flag and the local `LIMIT_SWAP_ENABLED_KEY`, which defaults to false, but this tab is added unconditionally, so every user gets a fourth permanently empty tab for a feature they cannot reach

app/src/main/java/com/vultisig/wallet/ui/models/mappers/DepositTransactionToUiModelMapper.kt
```diff
@@ -23,11 +24,20 @@ internal interface DepositTransactionToUiModelMapper :
  * UnbondStrategy paths on the initiator, and the joiner via ThorchainMemoParser.
  */
 @StringRes
-internal fun depositVerifyTitleRes(operation: String): Int =
-    if (operation == OPERATION_UNBOND) {
-        R.string.verify_deposit_unbonding
-    } else {
-        R.string.verify_deposit_sending
+/**
```
rkokhatskyi, line 19: This KDoc sits between `@StringRes` and the declaration, so tooling still binds the block above the annotation, the one promising the title is keyed off `operation` alone and never the raw memo. Merge the two above `@StringRes`

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/swap/limit/LimitOrderCancelDust.kt
aminsato: Valid, fixed in 83b9bc7.

XRP, TRON and SOL are all in `thorchainMemoAssetChainPrefix`, so they are offered — and all three fell through to the `else` branch, which is sized for ETH-like units. Added rows: XRP `10`, TRON `1`, SOL `0.02` (joining LTC/AVAX, whose threshold is the same).

The reason this happened is that the table was sized chain-by-chain against the thresholds I had in front of me, and nothing failed when a chain was added to the routable set without one. So the fix is a table entry plus a guard: a test now walks `staticLimitSwapSupportedChains`, fails if a routable chain has no pinned `dust_threshold`, and asserts each ceiling admits the dust its own threshold doubles to. A future chain fails in CI rather than in a user's hands.

One caveat worth stating plainly: I had no network access from this environment, so I could not re-read `inbound_addresses` myself. The pinned table is your numbers. If any is off, that test is the single place to correct it.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RefreshLimitOrdersUseCase.kt
aminsato: Valid, fixed in 83b9bc7.

Traced the premise before changing anything: `EvmApi.sendTransaction` returns either the node's `eth_sendRawTransaction` result or `Numeric.toHexString(localHash)` on the already-known path, both `0x`-prefixed, and that value goes straight into `inbound_tx_hash` via `recordPendingLimitOrderIfNeeded`. THORNode reports `swap.tx.id` bare. So the map lookup could never hit, and the order was never seen resting — which is worse than a blocked Cancel button, because it is also what feeds the `expiry_observed_at` the eligibility gate keys on.

New `thorchainTxId()` in `data/api/txstatus/`, applied to BOTH sides of the map — the queue's own ids as well as ours — so it is symmetric and idempotent rather than a one-sided patch. Regression test added: a `0x`-prefixed order matched against a bare, uppercase queue entry.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/MidgardLimitOutcomeResolver.kt
aminsato: Valid, fixed in 83b9bc7 — same helper as the queue lookup, and I took the uppercase from `THORChainTransactionStatusAPI.midgardTxid` as well rather than only stripping the prefix, since Midgard keys on uppercase hex and the two normalisations belong together.

Thanks for the iOS pointer; it is also what settled the `limit_swap` question in the thread below.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/MidgardLimitOutcomeResolver.kt
aminsato: I looked into this one hard and I don't think I can take it — iOS says the opposite, and its tests pin it.

`LimitOrderTrackingSeams.swift:279-310` documents the action set as: a `limit_swap` when the order is **placed**, a `swap` when it executes, a `refund` when it closes unfilled. The enum there carries an explicit note — "`limitSwap` is the PLACEMENT and is deliberately not an outcome — an order that is still resting has exactly this and nothing else" — and `MidgardLimitOutcomeResolverTests.testThePlacementActionAloneIsNotAnOutcome` asserts `.unresolved` against a captured mainnet body of exactly `{"type":"limit_swap","status":"success"}`.

If Android read that as `Filled`, every live resting order would close as filled the moment Midgard first indexed its placement — terminally, since nothing revisits a terminal order. That is the failure this resolver exists to prevent, so I'd rather leave the order `Unresolved` and ask again.

I could not check mainnet directly from this environment, so if you have a capture of a filled order returning `limit_swap` with no companion `swap`, that would settle it and I'll change it immediately.

What I did fix here, in 83b9bc7, is a real bug next to it: `classify` ran through `firstNotNullOfOrNull`, so the outcome depended on Midgard's list order, and the comment above it claimed a settled swap should **outrank** a refund. It is the other way round — an order that partially filled and then closed has both indexed, and what closed it is the refund; reading the older fill would report it terminally Filled on the strength of a fill that was only part of the story. A refund now outranks a fill explicitly, matching `MidgardLimitOutcomeResolver.swift`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RefreshLimitOrdersUseCase.kt
aminsato: Valid, fixed in 83b9bc7 — though not the way the comment suggests, and I want to be explicit about why.

Setting `isRefreshing` in `refreshOnEnter` would stop the two callers overlapping, but not the harm. Two polls a few hundred milliseconds apart land on the same load-balanced backend with near certainty, so even run strictly back-to-back they are not the independent observations `ABSENT_POLLS_BEFORE_CLOSING = 2` assumes — one stale round would still count twice.

So refreshes are now COALESCED per vault: a caller arriving while a poll is in flight joins it via a `CompletableDeferred` and returns when it completes, rather than starting a second. Per vault, not global, so another vault's refresh is never swallowed. This is close to what iOS gets structurally from one poll task per sender.

app/src/main/java/com/vultisig/wallet/ui/models/limitorder/BuildLimitOrderCancelTransactionUseCase.kt
aminsato: Valid, and the worst of the batch. Fixed in 83b9bc7.

Confirmed the mechanism: `ChainAndTokensToAddressMapperImpl` builds every `Account` with `tokenValue = null`, and `loadAddress` emits that mapper output before any balance is fetched — so `.first()` was, unconditionally, the pre-balance snapshot. The elvis fired on every call. The guard has never run in production, and c2671ae landed inside a branch that was already unreachable.

Two changes:

- The balance now comes from `accountsRepository.loadAccount(vaultId, signingCoin)`, which fetches live and falls back to the Room cache, wrapped so a failed read returns null rather than throwing.
- The fee is priced **before** the balance is read. A failed balance lookup still means "don't block" — refusing a cancel on a failed lookup is the wrong direction — but a planner that answers `Error_not_enough_utxos` is a refusal in its own right, and it must not be skipped just because the balance behind it was unavailable. That ordering is what makes the c2671ae check actually reachable.

app/src/main/java/com/vultisig/wallet/ui/models/limitorder/BuildLimitOrderCancelTransactionUseCase.kt
aminsato: Valid, fixed in 83b9bc7.

`isDepositPayload` now checks the memo before it looks at the specific: `LimitOrderCancelPresentation.isCancel(payload.memo)` returns true → deposit. That covers the Cosmos-sourced case too, which the `when` would otherwise have sent down the `MsgSend` branch. The PSBT exclusion still exits first, so a `SECURE+:` co-sign is unaffected.

Keying it off the memo rather than a wire flag is the same reasoning the presentation layer already used — the memo is all a joining device ever sees, and it is what THORChain itself reads to decide what the transaction is.

app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt
aminsato: Valid, fixed in 83b9bc7. The header and its strings had never rendered once.

`Keysign.kt`'s `Deposit`/`Send` branch now routes to `TransactionDoneView` when the model carries a `limitCancelPair`, via a small `TransactionTypeUiModel?.isLimitOrderCancel()` next to the rest of the cancel presentation. The send overview is wrong for a cancel for the same reason the Verify screen needed its own title: it headlines the attached dust as though the user had sent it somewhere.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RefreshLimitOrdersUseCase.kt
aminsato: Valid, fixed in 83b9bc7 — the reason shown, not the tracking, and I want to be clear about which.

Those rows cannot be tracked at all, and there is no backfill available: `source_chain` is NULL alongside `source_address`, so there is nothing to derive a sender address from. `source_asset` is not a substitute — a secured-asset source is THORChain-placed but carries a bare denom with no `THOR.` prefix, which is exactly why `source_chain` was added as its own column.

What was actually broken is what the card said. `limitOrderCancelEligibility` reached `PlacementNotObserved` first, so these orders showed "cancellable once its deposit confirms" — a promise that can never come true, since they are skipped by every poll. There is now a `LegacyOrderNotTrackable` blocker checked ahead of the resting gate, with copy saying the order was placed by an older version and will still refund automatically at expiry. iOS has the same concept (`limitSwap.cancel.unavailableLegacyOrder`).

app/src/main/java/com/vultisig/wallet/ui/models/limitorder/LimitOrderHistoryUiModel.kt
aminsato: Valid, fixed in 83b9bc7.

`sellAmount` is now nullable and the card renders the pair alone when it is null — `RUNE → BTC`, never a number in the coin's smallest units. Where the precision is genuinely recoverable it still shows an amount: if the placement memo names a chain's own native asset (`BTC.BTC`, `THOR.RUNE`), the decimals come from `Chain.nativeToken`. A token leg (`ETH.USDC-0X…`) carries its own precision that the memo does not spell, so those render without an amount rather than with a wrong one.

app/src/main/java/com/vultisig/wallet/ui/models/limitorder/LimitOrderHistoryUiModel.kt
aminsato: Valid, fixed in 3a24285.

Gated on `Filled` specifically, rather than on `isTerminal` like `expiryText` above it. The contradiction you describe is real and is particular to that status: the stored split is the last RESTING observation, taken before the fill that closed the order, and `recordStatus` does not revise those columns — so the card asserts two different things about the same order.

The other terminals keep it deliberately. The fill columns are persisted precisely because a terminal order disappears from the queue taking its fill state with it, and an order that expired 40% filled paid out that part and refunded the rest — that split is the only explanation the user gets for a two-leg settlement.

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
aminsato: Valid, fixed in 83b9bc7.

`observeLimitOrders` now combines the Room flow with a 15s tick, so `nowMs` is resampled on a timer instead of only when something writes to the table. Well under the minute the label is granular to, and it costs a pure re-map of a short list — no network, no database. When nothing has changed the mapped list compares equal and the StateFlow drops the emission, so it does not churn recomposition either.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TransactionHistoryScreen.kt
aminsato: Valid, fixed in 83b9bc7 — the ViewModel filters `limitOrders` by `selectedAssetIds` in the same `combine` that maps them, so the screen keeps rendering what it is given.

An order matches on EITHER leg: filtering to ETH should surface a RUNE→ETH order as readily as an ETH→BTC one, since a pair is what an order is about. Matched on the ticker rather than the full `chain:ticker` id, because only the order's SOURCE chain is recorded — qualifying the buy leg by chain isn't possible, and qualifying just one of the two would be arbitrary.

app/src/main/java/com/vultisig/wallet/ui/screens/deposit/VerifyDepositScreen.kt
aminsato: Valid, fixed in 3a24285. That sentence was true only on the THORChain route.

The L1 route now gets its own copy naming the donation: the amount above is the minimum THORChain needs to see the request, it goes to the pool and cannot be refunded, and the network fee is on top of it. Ten locales.

Chosen off the transaction's amount rather than the source chain — a THORChain-sourced cancel is a `MsgDeposit` carrying no coins, everything else carries dust — because the amount is precisely what the user is about to part with, and it is the value rendered directly above the sentence.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TransactionHistoryScreen.kt
aminsato: Valid, fixed in 3a24285.

The tab is now gated on the same conjunction the swap form gates placement on: the remote `limit-swap` flag AND `LimitSwapConfig.isFeatureEnabled`. It is last in the group, so the other three keep their ordinals when it is absent.

One addition beyond the report, because gating on the flags alone would have introduced a different bug: the tab is also shown when the vault already HAS orders. Otherwise flipping the kill switch — or a user toggling Advanced Settings off — would hide a resting order along with the only way to cancel it. Two tests cover both halves.

app/src/main/java/com/vultisig/wallet/ui/models/mappers/DepositTransactionToUiModelMapper.kt
aminsato: Valid, fixed in 3a24285 — merged into one block above `@StringRes`.

The stale half was worse than misplaced: it stated the title is keyed off `operation` alone and never the raw memo, which is the exact opposite of what the function now does and why. Anyone reading it in tooling would have concluded the memo branch was a mistake.

data/src/main/kotlin/com/vultisig/wallet/data/swap/limit/LimitOrderCancelEligibility.kt
aminsato: Good question, and I've left the gate as it is — here's the reasoning, and I'd like your read on it.

You're right that the two fields come off different parts of the entry: `expiry_observed_at` is written on every resting poll, `deposit_amount` only when the queue entry carries a `state` block. So an entry without `state` clears the gate on `expiryObservedAt` alone and leaves the amount cross-check comparing against null, which `agreesWithChain` treats as agreement.

That is deliberate rather than accidental — absence is not disagreement, and refusing to cancel an order until the first poll that happens to carry a `state` block would be a worse failure than the one the check prevents. But you've identified the residual risk correctly: in that state the cancel proceeds on signed-only data.

What keeps it from being open-ended is that the signed data is not a guess — `source_amount_1e8` and `trade_target` are captured at signing from the exact integers the memo was built from, and a cancel that reproduces them lands in the bucket the placement created. The cross-check is corroboration, not the primary source.

I could not verify against mainnet from this environment (no network access this session), so I have not been able to confirm whether a genuinely resting entry ever lacks `state`. Every sample I have — including the iOS fixtures captured from mainnet — carries it. If you know of a case that doesn't, say so and I'll make `PlacementNotObserved` require `depositAmount` rather than either field.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RefreshLimitOrdersUseCase.kt
aminsato: Valid, fixed in 83b9bc7 — a per-order `try`/`catch` inside the loop, rethrowing `CancellationException`, so one order's failed Room write no longer abandons its siblings or the remaining senders. It now matches the per-sender isolation the fetch above it already had, which is what made the gap visible.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/RefreshLimitOrdersUseCase.kt
aminsato: Valid, and documented in 83b9bc7 rather than changed.

The KDoc on `refundedOrCredited` now states the approximation and why it is tolerated: the two outcomes differ only in the label on an already-closed order — the funds took the same path either way — and the alternative, an exact block height, is not knowable for an order that has already left the queue. Worth having written down next to the constant rather than inferred from it.

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
aminsato: Valid, fixed in 83b9bc7. Split into two `launch`es inside a `coroutineScope`, each with its own catch, so pull-to-refresh has the same fault isolation `refreshOnEnter` does. `isRefreshing` still clears in the `finally` once both settle.

It was not an accepted inconsistency, just an oversight — there was no reason for the manual path to be less robust than the automatic one.

app/src/main/java/com/vultisig/wallet/ui/models/limitorder/BuildLimitOrderCancelTransactionUseCase.kt
aminsato: Valid, fixed in 83b9bc7 — `resolveDust` now catches `ArithmeticException` alongside `LimitOrderCancelDustError` and maps it to the same `DustUnavailable`.

Not reachable today, and it stayed unreachable with the new rows: I checked each added ceiling against its chain's precision (XRP/TRON at 6, SOL at 9) and all divide evenly. But the failure mode you describe is exactly the kind that would arrive silently with a future chain, so the defensive catch is worth its two lines.

app/src/test/java/com/vultisig/wallet/ui/models/transaction/TransactionHistoryViewModelTest.kt
aminsato: Fair, and fixed. Five tests at this layer in 83b9bc7 and 3a24285:

- `cancelLimitOrder` happy path — asserts the built transaction is stored and that navigation targets the SAME id, since routing to an id the repository never received would land the user on an empty verify screen.
- a refused cancel — surfaces the right message, stores nothing, navigates nowhere, and `dismissCancelError` clears it.
- the LIMIT tab's filtering — an active chip matching on either leg.
- two for the new tab-visibility gate.

`observeLimitOrders`' own state update is exercised by the last three, since they all read `uiState.limitOrders` after the flow settles.

aminsato:
Thanks both — 22 threads, all replied inline. Two commits: `83b9bc7` and `3a24285`.

**21 of 22 fixed.** The three that mattered most were all things that had never worked in production rather than edge cases:

- the affordability guard read the balance off `loadAddress().first()`, which the mapper always builds with `tokenValue = null`, so the whole check — including the `plan.error` branch added in c2671ae — short-circuited on every call;
- EVM hashes are stored `0x`-prefixed and both THORNode's queue and Midgard key on the bare form, so no ETH/AVAX/BSC/BASE order ever matched its own queue entry or resolved an outcome;
- XRP, TRON and SOL fell through the dust ceiling's `else`, so every cancel from them died as `ExceedsCeiling` with a live Cancel button still on the card.

**One declined:** `limit_swap` as a fill (`MidgardLimitOutcomeResolver.kt:75`). iOS documents and test-pins it as the **placement** action — a still-resting order has exactly that and nothing else — so reading it as `Filled` would terminally close every live order the moment Midgard first indexed it. Full reasoning and the iOS citations are in that thread; a mainnet capture showing otherwise would settle it and I'll change it. I did fix the real ordering bug next to it: a refund now outranks a fill regardless of Midgard's list order.

**Two guards added so these classes of bug fail in CI instead:** a test that walks `staticLimitSwapSupportedChains` and fails if a routable chain has no sized dust ceiling, and five `TransactionHistoryViewModel` tests covering `cancelLimitOrder`, its error path, the tab filter and the new visibility gate.

**One caveat that spans several threads:** I had no network access while working on this, so I could not re-read `inbound_addresses` or Midgard directly. The pinned `dust_threshold` table in `LimitOrderCancelDustTest` is @rkokhatskyi's numbers, and it is the single place to correct any of them. Everything else was verified against the code or against the iOS implementation.

`:app` and `:data` suites green, `:app:lintDebug` green, ktfmt clean.

rkokhatskyi (commented):
> **One caveat that spans several threads:** I had no network access while working on this, so I could not re-read `inbound_addresses` or Midgard directly. The pinned `dust_threshold` table in `LimitOrderCancelDustTest` is @rkokhatskyi's numbers, and it is the single place to correct any of them. Everything else was verified against the code or against the iOS implementation.

@aminsato this needs to be checked carefully

johnnyluo:
@aminsato please provide on chain tx link to place a limit swap and cancel it

johnnyluo:
CleanShot 2026-07-31 at 20 34 52@2x
Refer to the snapshot above, I was trying to place a limit swap order , swap 5 RUNE -> DOGE

you can see the price is both confusing and wrong, 

5 RUNE should be able to swap around 27 DOGE , why it is 13.23 ?

Also it should be 1 RUNE worth x amount of DOGE , please refer to the implementation in IOS/MacOS

aminsato:
On-chain proof of the full place → cancel → refund loop, all on Dogecoin.

**1. Place** — [`8ee3e04a…fdaa`](https://blockchair.com/dogecoin/transaction/8ee3e04a82253e0abc967ce754efb993e11e1d55ab3fa7ee70f9f207edfefdaa) (09:04:29 UTC)

15 DOGE to the inbound vault, `OP_RETURN`:

```
=<:THOR.RUNE:thor12a9…ua5t:265857540/7200/0:va:50
```

It then rested: `GET /thorchain/queue/limit_swaps?sender=DKwcZ…P1zx` returned `total: 1` with `time_to_expiry_blocks: 7148` of 7200.

**2. Cancel** — [`e5836ba6…7e14`](https://blockchair.com/dogecoin/transaction/e5836ba66120bac55168a2584114e0be817ffaa16c1face3c42e00a1cd337e14) (09:31:14 UTC)

2 DOGE dust to the same inbound vault, `OP_RETURN` (44 bytes):

```
m=<:1500000000DOGE.DOGE:265857540THOR.RUNE:0
```

Both amounts match the placement exactly — `1500000000` is the deposit, `265857540` the trade target — which is the pair THORNode keys the resting order on.

**3. Refund** — [`2c1b1e6a…2a01`](https://blockchair.com/dogecoin/transaction/2c1b1e6a45dab34201b47605af0aa361e7cc28924647256f673df10511772a01) (09:33:10 UTC)

The queue went to `total: 0` and THORChain refunded `11.42132005 DOGE` to the sender under `REFUND:8EE3E04A…FDAA` — the 15 DOGE deposit less the ~3.58 DOGE Dogecoin outbound fee.

So the L1 cancel route works end to end: `m=<` sent from the chain that funded the order, matched by THORNode, order dropped from the queue, funds returned.

aminsato:
Both halves of this were real, and the `13.23` was the more serious one — you were right that the number itself was wrong, not just its label.

### The rate was wrong (d4fb729)

`fetchMarketPrice` sized the reference quote from `Coin.usdPrice`, which nothing populates outside the SPL-token path. For RUNE it was `null → ZERO`, so `getMarketProbeAmount` fell back to quoting **one whole unit** — the exact case its ~$100 notional was written to avoid. A 1 RUNE probe into DOGE grosses ~6.1 DOGE and pays ~3.6 of it in Dogecoin outbound fee, so the price came back ~2.5 DOGE/RUNE against an actual ~6.1. Your "5 RUNE should get ~27 DOGE" was correct; the form was showing roughly half.

This was not cosmetic — the presets, the below-market warning and the LIM signed into the memo all derived from that number, so orders rested at about half the intended price. That is also why they kept filling immediately and drawing the "price already available, consider a market swap" warning.

Fixed by pricing the source through `TokenPriceRepository`, explicitly in USD: the notional is a fixed `100`, so a JPY-denominated price would size the probe ~150× too small and reproduce the same fee-dominated quote for anyone who merely changed their display currency.

Verified against mainnet — THORChain quotes `100 RUNE → 610.25017841 DOGE` (6.1025/RUNE); the form now reads `6.12238842` where it read `2.55210159`.

### The card was quoting three different quantities (7b0daea, 9338fe3, 3dfdb6b)

The header named the **buy** asset (`1 DOGE`), the large number was the fiat price of one DOGE, and the line beneath it was the **total output** for the sell amount — three unrelated figures stacked under one heading, which is what made it unreadable.

All three now describe the one quantity you entered, per the iOS/macOS convention:

```
5 RUNE
$2.15
30.61194210 DOGE
```

The asset line equals the buy leg of the Asset section by construction, and the fiat line values the destination amount, so it moves with the +1/+5/+10% pills instead of sitting frozen.

The confirmation screen had the same inversion — it read `1 <buy> = $x`, the reciprocal of the rate the memo's LIM actually encodes, on the screen a co-signer approves. It now states the rate in asset terms: `1 RUNE = 6.12238842 DOGE`.

One note: this diverges from the Figma mock (node `78798:74520`), which expresses the price as `1 <buy asset> = $<usd>`. iOS deliberately does not follow it either — happy to revisit if design wants the other convention.

johnnyluo (approved):
LGTM

PR #5349, opened 2026-07-20
feat(ripple): SHA-512-half digest for XRPL custom-message co-signing

aminsato:
## Summary

- Adds the missing **Ripple** branch to the custom-message keysign digest so a Secure (multi-device) Vault can co-sign an XRPL off-chain `signMessage` initiated by the browser extension.
- XRPL (GemWallet `signMessage`) messages sign **SHA-512-half** — the first 32 bytes of SHA-512 — of the message bytes. Android had no Ripple branch, so a `chain: Ripple` message fell through to the `else → keccak256` path. The digest (and the md5 message-ID derived from it) diverged from the initiator, so the Android co-signer could not locate the setup message and cross-platform co-signing 404'd — the same failure class already documented in this method for `eth_signTypedData_v4` / Cosmos.
- Mirrors the shared digest contract in vultisig-windows `getCustomMessageHex.ts` (`ripple: () => sha512(bytes).slice(0, 32)`), added in vultisig/vultisig-windows#4399.

## Changes

- `ByteArrayExtensions.kt`: add `toSha512ByteArray()` alongside `toSha256ByteArray()`.
- `SigningHelper.getKeysignMessages(CustomMessagePayload)`: add `chain == Chain.Ripple -> processedBytes.toSha512ByteArray().copyOf(32)` before the `else`. No behavior change for any other chain.

## Scope notes

- Only affects **Secure Vaults** (extension initiator + Android co-signer). Fast Vaults co-sign with VultiServer, so Android is never invoked.
- Co-signer digest only — DER encoding + uppercasing remain the initiator's responsibility; no signature-format change here.
- Out of scope: originating `signMessage` on Android, XRPL dApp **transaction** co-signing (already shipped via `RippleHelper`), and any confirmation-screen display.

## Test plan

- `./gradlew :data:testDebugUnitTest --tests "*.SigningHelperCustomMessageTest"` — green.
- New unit tests:
  - `Ripple plain-text message is SHA-512-half hashed, not keccak256` — behavioral assertion.
  - `Ripple digest is byte-identical to the vultisig-windows extension contract` — pins two reference vectors (UTF-8 and `0x`-hex inputs) computed independently, proving byte-for-byte parity with the merged Windows contract.

Closes #5341

## Summary by CodeRabbit

* **New Features**
  * Added Ripple-specific message hashing using SHA-512-half for signing message IDs.

* **Bug Fixes**
  * Corrected Ripple message digest generation for both plain-text and hexadecimal message inputs.

* **Tests**
[...]

PR #5465, opened 2026-07-30
fix(sui): intent-wrap the custom-message digest so co-signing converges

### LINKED ISSUE #5442: [Fix] Sui signPersonalMessage/signMessage custom-message co-signing computes the wrong digest
# [Fix] Sui signPersonalMessage/signMessage custom-message co-signing computes the wrong digest

## Problem
`SigningHelper.getKeysignMessages(CustomMessagePayload)` treats any EdDSA-family chain (which includes Sui) as needing no digest transformation — it signs the raw message bytes verbatim. Sui's Wallet Standard `signPersonalMessage`/`signMessage` actually requires `blake2b_256(intent[3,0,0] || bcs(message))` — an intent-wrapped digest. This generic EdDSA passthrough was introduced for all EdDSA chains without a Sui-specific carve-out.

The identical bug shape was already fixed for Cosmos (PR #5240) and Ripple (PR #5349) in this exact function — Sui is simply the next chain not yet covered.

## Expected Behavior
A Sui `signPersonalMessage`/`signMessage` request should be hashed as `blake2b_256([0x03,0x00,0x00] || uleb128(len) || message)`, matching the Windows extension's `getSuiDigestHex` and the Sui Wallet Standard spec.

## Steps to Reproduce
1. A Sui dApp calls `signPersonalMessage`/`signMessage` through the Windows extension against a Secure Vault with a mobile co-signer.
2. The initiator (Windows) derives the message-to-sign as the intent-wrapped blake2b digest.
3. The joining Android device independently derives the raw message bytes instead.
4. The two parties' derived hashes diverge — the MPC threshold-signing round cannot converge, and the session hangs/fails with no diagnostic pointing at the real cause.

## Solution
Add a Sui-specific branch to `SigningHelper.getKeysignMessages(CustomMessagePayload)`: BCS-encode the message as `vector<u8>` (uleb128 length prefix + bytes), prepend the `[0x03,0x00,0x00]` PersonalMessage intent, and blake2b-256 hash the result (Bouncy Castle `Blake2bDigest`, already used in `SwapKitSuiSigner.kt`).

### Must Do
- [ ] Add the Sui-specific intent-wrapped blake2b digest branch for custom-message signing
- [ ] Add a regression test pinning the digest against a known Sui test vector

### Must NOT Do
- Don't change the EdDSA passthrough behavior for other EdDSA chains
- Don't change the `signSui`/PTB co-signing path (unrelated — that's a different, already-correct pipeline)

## Acceptance Criteria
- [ ] A Sui `signPersonalMessage` request produces the correct intent-wrapped blake2b digest, matching Windows
- [ ] New test asserts the digest against a known-correct value

aminsato:
Closes #5442

## Problem

Sui is an EdDSA chain, so `SigningHelper.getKeysignMessages(CustomMessagePayload)` matched the generic `chain?.TssKeysignType == TssKeyType.EDDSA -> processedBytes` branch and signed the **raw message bytes**.

The Sui Wallet Standard `signPersonalMessage` / `signMessage` digest is not the raw bytes. It is:

```
blake2b_256( [PersonalMessage(3), V0(0), Sui(0)] || uleb128(len) || message )
```

So when a Sui dApp signs through the Windows extension against a Secure Vault, the initiator derives the intent-wrapped digest while the joining Android device derives the raw bytes. The two parties never agree on what to sign — the MPC threshold round cannot converge, and the session hangs with no diagnostic pointing at the real cause.

This is the same bug shape already fixed for Cosmos (#5240) and Ripple (#5349) in this exact `when`-expression. Sui was simply the next chain not yet carved out.

## Fix

- `SigningHelper.kt` — add `chain == Chain.Sui -> processedBytes.toSuiPersonalMessageDigest()` ahead of the EdDSA branch, mirroring the Ripple line directly above it.
- `ByteArrayExtensions.kt` — new `toSuiPersonalMessageDigest()` implementing the intent prefix, BCS `vector<u8>` ULEB128 length prefix, and blake2b-256.

The helper lives in `ByteArrayExtensions` alongside the other hashes this dispatch already uses (`toSha256ByteArray`, `toSha512ByteArray`, `toKeccak256ByteArray`) rather than in `SuiHelper` — that object initializes `CoinType.SUI` at class load, which would drag WalletCore JNI into the unit tests and make them unrunnable on the JVM.

The `signSui` / PTB co-signing path is untouched. It uses the `TransactionData(0)` intent scope with no BCS wrap and was already correct; see `SwapKitSuiSigner.digest`.

## Cross-platform contract

The intent encoding was confirmed from `@mysten/sui`'s BCS definitions, not from memory:

| Field | Value |
|---|---|
| `IntentScope` | `TransactionData=0, TransactionEffects=1, CheckpointSummary=2, PersonalMessage=3` |
| `IntentVersion` | `V0 = 0` |
| `AppId` | `Sui = 0` |

giving the `[0x03, 0x00, 0x00]` prefix, matching `@vultisig/core-chain`'s `suiPersonalMessageIntent`.

The extension sends the message as `0x`-prefixed hex (`clients/extension/src/inpage/providers/sui/sign.ts`), so it lands on the existing hex-decode branch — deliberately, to avoid UTF-8 round-tripping corrupting binary payloads. Both the hex and UTF-8 branches are covered.

## Testing

**Test vectors are pinned against the canonical initiator implementation** — `@vultisig/core-chain`'s `getSuiPersonalMessageDigest`, the exact function the extension's `getCustomMessageHex.ts` `sui` branch calls — rather than derived from this code, so the test can't agree with a bug it shares.

| Input | Expected digest |
|---|---|
| `"Hello Vultisig"` (UTF-8) | `4a140032aaf9e5699dc678b2caf2687349a2a6ac1103cdba6a2d3f462ca21182` |
| `0x56756c7469736967` (hex) | `245f3ecbbcca443e43d26989bab5ba55cf03c6db0f9b28a4e12d072b894e83d6` |
| 200 x `0x41` | `a90b2a8fe4149193029c26b2e2000730da6e64f6dee07ccf939142d02ef7ceb6` |

The 200-byte case forces the BCS length prefix to spill to two bytes (`0xC8 0x01`). A naive single-byte encoder passes every short vector and silently diverges on exactly the messages most likely to be real — dApp login challenges routinely exceed 127 bytes.

A third test asserts Sui no longer produces the raw-passthrough result, pinning the regression itself.

- All 3 new tests **fail with the branch removed and pass with it** — verified explicitly, not assumed.
- Full `:data` suite: **2470 tests, 0 failures**.
- `./gradlew assembleDebug` succeeds.
- ktfmt clean.

## Not covered

The live multi-device ceremony is **not** verified. That needs a Sui dApp driving the Windows extension as initiator against a Secure Vault with an Android co-signer. The digest now matches the canonical implementation byte-for-byte, but convergence of the real MPC round is untested here.

Note there is no Android-only way to exercise this: the in-app Sign-custom-message screen (`SignMessageFormViewModel`) never sets `chain`, so it always falls to the keccak256 default and cannot reach the Sui branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Sui personal message signing by applying the correct Wallet Standard digest format.
  * Added support for UTF-8, hexadecimal, and longer messages during Sui signing.

* **Tests**
  * Added coverage for Sui digest reference values, extended message lengths, and separation from raw EdDSA behavior.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/SigningHelper.kt
@@ -8,6 +8,7 @@ import com.vultisig.wallet.data.common.toHexBytes
 import com.vultisig.wallet.data.common.toKeccak256ByteArray
 import com.vultisig.wallet.data.common.toSha256ByteArray
 import com.vultisig.wallet.data.common.toSha512ByteArray
+import com.vultisig.wallet.data.common.toSuiPersonalMessageDigest
 import com.vultisig.wallet.data.crypto.SuiHelper
 import com.vultisig.wallet.data.crypto.ThorChainHelper
 import com.vultisig.wallet.data.crypto.TonHelper
@@ -70,12 +71,16 @@ object SigningHelper {
         //  - Ripple (XRPL GemWallet signMessage / ripple-keypairs) signs SHA-512-half — the
         //    first 32 bytes of SHA-512 — of the message bytes. keccak256 here diverges from
         //    the extension initiator's digest and cross-platform co-signing 404s.
+        //  - Sui is EdDSA but is *not* a raw passthrough: the Wallet Standard
+        //    signPersonalMessage/signMessage digest is intent-wrapped and BCS-length-prefixed,
+        //    so it must be carved out ahead of the EdDSA branch below.
         //  - Everything else (EVM, Tron) signs the keccak256.
         val bytes =
             when {
                 chain?.standard == TokenStandard.COSMOS ||
                     chain?.standard == TokenStandard.THORCHAIN -> processedBytes.toSha256ByteArray()
                 chain == Chain.Ripple -> processedBytes.toSha512ByteArray().copyOf(32)
+                chain == Chain.Sui -> processedBytes.toSuiPersonalMessageDigest()
                 chain?.TssKeysignType == TssKeyType.EDDSA -> processedBytes
                 else -> processedBytes.toKeccak256ByteArray()
             }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/common/ByteArrayExtensions.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/common/ByteArrayExtensions.kt
@@ -3,6 +3,7 @@ package com.vultisig.wallet.data.common
 import com.google.protobuf.ByteString
 import com.vultisig.wallet.data.utils.Numeric
 import java.security.MessageDigest
+import org.bouncycastle.crypto.digests.Blake2bDigest
 import org.bouncycastle.jcajce.provider.digest.Keccak
 
 fun ByteArray.toKeccak256(): String {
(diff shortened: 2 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5467, opened 2026-07-31
fix(zksync): propagate zk fee estimate RPC errors instead of zeroing

### LINKED ISSUE #5398: [Fix] ZkSync fee calculation collapses to all-zero on any RPC error, showing a normal-looking $0.00 fee

johnnyluo:
## Summary
- `EvmApi.zkEstimateFee()` collapsed a JSON-RPC error (or a null `result`) into an all-zero `ZkGasFee`, and neither consumer (`ZkFeeService`, the ZkSync branch in `BlockChainSpecificRepository`) applied a floor — so a transient/unsupported-RPC error on `zks_estimateFee` produced a normal-looking $0.00 fee preview and a `gasLimit=0` signed transaction.
- Broadcast then rejects the zero-fee tx as intrinsic-gas-too-low, after the full multi-device MPC signing ceremony already ran.
- Fixed by making `zkEstimateFee()` throw `NetworkException` on RPC error / missing result, matching the existing `getBaseFee()`/`getMaxPriorityFeePerGas()` convention (#5400) — `FeeServiceComposite.calculateFees` already catches and retries via `calculateDefaultFees`, and a repeated failure now surfaces as a visible error instead of silently signing a zero-fee tx.

## Test plan
- [x] `./gradlew :data:testDebugUnitTest --tests "com.vultisig.wallet.data.api.EvmApiFeeTest" --tests "com.vultisig.wallet.data.blockchain.ethereum.ZkFeeServiceTest" --tests "com.vultisig.wallet.data.repositories.BlockChainSpecificRepositoryImplTest"` — pass
- [x] Added tests covering `zkEstimateFee`'s RPC-error path, missing-result path, and happy path

Fixes #5398

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ZkSync fee estimation error handling.
  - RPC failures and missing fee results now report a network error instead of returning an invalid zero fee.
  - Successful responses continue to provide the expected gas limit and fee details.

- **Tests**
  - Added coverage for failed, incomplete, and successful fee estimation responses.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -672,19 +672,21 @@ class EvmApiImp(
                 },
             )
         return if (response.error != null) {
-            Timber.d(
-                "zk estimate fee, srcAddress: $srcAddress,dstAddress: $dstAddress," +
-                    "data: $data error: ${response.error.message}"
+            // A failed estimate must propagate, never collapse into ZERO: a zeroed gasLimit/
+            // maxFeePerGas signs a tx that broadcast then rejects as intrinsic-gas-too-low,
+            // after the ceremony already ran (#5398).
+            throw NetworkException(
+                httpStatusCode = 0,
+                message =
+                    "zk estimate fee rpc error, srcAddress: $srcAddress,dstAddress: $dstAddress," +
+                        "data: $data error: ${response.error.message}",
             )
-            ZkGasFee(BigInteger.ZERO, BigInteger.ZERO, BigInteger.ZERO, BigInteger.ZERO)
         } else {
             val resultJson =
                 response.result
-                    ?: return ZkGasFee(
-                        BigInteger.ZERO,
-                        BigInteger.ZERO,
-                        BigInteger.ZERO,
-                        BigInteger.ZERO,
+                    ?: throw NetworkException(
+                        httpStatusCode = 0,
+                        message = "zk estimate fee rpc returned no result",
                     )
             ZkGasFee(
                 gasLimit = resultJson.gasLimit.convertToBigIntegerOrZero(),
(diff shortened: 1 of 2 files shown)


rkokhatskyi (approved):
LGTM

PR #5468, opened 2026-07-31
fix(keysign): relay signTon, signSui and signBitcoin to co-signers

### LINKED ISSUE #5445: [Fix] Outbound proto mapper never serializes signSui (or signSolana/signBitcoin) — silent relay gap if ever re-serialized
# [Fix] Outbound proto mapper never serializes signSui (or signSolana/signBitcoin) — silent relay gap if ever re-serialized

## Problem
`PayloadToProtoMapper` (outbound: domain → proto) maps every other chain-specific field, including `signAmino`/`signDirect` for Cosmos, but has no branch for `signSui` (nor `signSolana`/`signBitcoin`). The inbound mapper (`KeysignPayloadProtoMapper`) *does* read `signSui` from an incoming proto. This is currently unreachable in practice — no Android code path originates a `signSui`-populated `KeysignPayload` today (no Sui dApp connector exists on Android) — but the moment any future code re-serializes a received `signSui`-populated payload (a native Sui dApp connector, a resend/session-recovery utility, anything that persists and replays a `KeysignPayload`), the dropped field causes a silent hash mismatch: the peer would fall through to rebuilding a generic Pay/PaySui from RPC coins instead of the dApp's actual PTB, signing a different transaction than requested.

## Expected Behavior
The outbound mapper should be symmetric with the inbound one for `signSui` (and `signSolana`/`signBitcoin`), mirroring the existing `signAmino`/`signDirect` round-trip treatment.

## Steps to Reproduce
Currently unreachable via any shipped Android flow. Latent: would surface the moment a future feature re-serializes a `signSui`-populated `KeysignPayload`.

## Solution
Add `signSui`, `signSolana`, and `signBitcoin` passthrough fields to `PayloadToProtoMapper`'s `KeysignPayloadProto(...)` construction, mirroring the existing `signAmino`/`signDirect` treatment.

### Must Do
- [ ] Add `signSui`/`signSolana`/`signBitcoin` passthrough to the outbound mapper
- [ ] Add a round-trip test asserting `outbound(inbound(proto))` preserves these fields unchanged

### Must NOT Do
- Don't change the inbound mapper (already correct)
- Don't add new dApp-connector functionality — this is purely the proto round-trip symmetry fix

## Acceptance Criteria
- [ ] New test asserts `signSui`/`signSolana`/`signBitcoin` round-trip through the outbound mapper unchanged

rkokhatskyi:
## Summary

`PayloadToProtoMapper` (domain → proto, used by the keysign initiator in `BuildKeysignMessageUseCase`) wrote only four of the seven `sign_data` oneof members. A relayed payload therefore reached peer devices with `signTon`, `signSui` and `signBitcoin` stripped, even though the inbound `KeysignPayloadProtoMapper` has always read all seven.

A peer that loses one of these rebuilds a default signing input instead — a single plain TON transfer rather than the TON Connect messages (`TonHelper.kt:69-77`), a `Pay`/`PaySui` rather than the dApp's PTB (`SuiHelper.kt:81`) — so its message hash stops matching the initiator's and the `md5(hash)`-keyed DKLS setup message 404s. `signBitcoin` fails harder: it carries the only PSBT marker, since `BlockChainSpecific.BitcoinPSBT` is a sibling of `UTXO` and deliberately maps to no chain-specific block. Dropping it leaves nothing for the inbound mapper to resolve and it hits its terminal `error(...)`.

iOS models `sign_data` as one enum and switches over it exhaustively (`Cosmos+ProtoMappable.swift:38-55`); the extension builds the protobuf oneof directly (`keysignPayload/build.ts:529-547`). Both emit all seven. Android flattened the oneof into seven independent nullable fields that the mapper lists by hand, which is how three came to be missed.

No shipped flow changes behaviour: Android does not originate a `signTon`/`signSui`/`signBitcoin` keysign today, so the gap is latent until such a payload is re-serialized. `signSolana` (named in the issue) and `signRipple` were already covered by #5239 and #5303.

Also folds the three near-duplicate comment blocks in that region into one and drops the stale note that still listed these three as a deferred follow-up.

Closes #5445

## Notes

- No `BitcoinPSBT` branch is added to the outbound chain-specific chain, deliberately: the variant has no fields, and the inbound mapper resolves it from `signBitcoin` before it looks at any `*Specific` block. Fabricating a `utxo_specific` would reintroduce exactly the stale `byteFee`/`sendMaxAmount` the inbound mapper documents discarding.
- `signBitcoin` is the one member whose round-trip is still not wire-compatible with iOS: `KeysignPayloadProtoMapper.kt:136` collapses the wire's `utxoSpecific` into the dataless `BitcoinPSBT` marker, so an Android-relayed PSBT payload carries no `blockchain_specific` and iOS rejects it at `KeysignMessage+ProtoMappable.swift:80`. That is unchanged by this PR — before it, the same payload failed on iOS *and* on Android — and unreachable today. Fixing it means giving `BitcoinPSBT` a `byteFee`/`sendMaxAmount`, which touches the inbound mapper this ticket scopes out; worth a follow-up.
- A re-serialized `signSui` payload emits a zeroed `suicheSpecific` from the inbound placeholder. That is inert — the inbound `signSui` branch runs ahead of `suicheSpecific` on every platform — and the test pins that a second pass is a fixed point.

## Testing

- `./gradlew :data:testDebugUnitTest` — 2517 tests, 0 failures
- `./gradlew :data:ktfmtCheck :data:lintDebug`
- `./gradlew :app:compileDebugKotlin`

The three new round-trip tests were confirmed to fail against the unpatched mapper (`signBitcoin`, `signSui`, `signTon` all red; the pre-existing tests in those files stayed green).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signing payload handling for Bitcoin, Sui, and TON transactions.
  * Preserved transaction details, messages, BOC fields, and chain-specific markers during data conversion.

* **Tests**
  * Added coverage confirming Bitcoin, Sui, and TON signing payloads remain intact through conversion round trips.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/mappers/PayloadToProtoMapper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/mappers/PayloadToProtoMapper.kt
@@ -258,12 +258,13 @@ internal class PayloadToProtoMapperImpl @Inject constructor() : PayloadToProtoMa
                 } else null,
             wasmExecuteContractPayload = keysignPayload.wasmExecuteContractPayload,
             // Every `sign_data` variant carries pre-built signing bytes a peer cannot reconstruct.
-            // Relaying one as null makes the peer rebuild a default input (bank send / plain
-            // transfer / Pay / OperationPayment) whose message hash diverges from the initiator's,
-            // so the DKLS setup message — keyed by md5(hash) — 404s and keysign never completes.
-            // `signBitcoin` also carries the only PSBT marker: drop it and the payload reaches the
-            // peer with no `blockchain_specific` member at all, which the inbound mapper rejects.
-            // The proto keeps the seven in a `oneof`, so at most one is ever set.
+            // Relaying one as null makes the peer fall back to rebuilding a default input (bank
+            // send / plain transfer / Pay / OperationPayment): at best its message hash diverges
+            // from the initiator's and the md5(hash)-keyed DKLS setup message 404s; at worst there
+            // is nothing to rebuild from and it throws — a `signSui` payload carries no toAddress,
+            // and `signBitcoin` carries the only PSBT marker, so without it the payload reaches the
+            // peer with no `blockchain_specific` member at all. The proto keeps the seven in a
+            // `oneof`, so at most one is ever set.
             signAmino = keysignPayload.signAmino,
             signDirect = keysignPayload.signDirect,
             signSolana = keysignPayload.signSolana,
--- a/data/src/main/kotlin/com/vultisig/wallet/data/mappers/PayloadToProtoMapper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/mappers/PayloadToProtoMapper.kt
@@ -257,37 +257,19 @@ internal class PayloadToProtoMapperImpl @Inject constructor() : PayloadToProtoMa
                     )
                 } else null,
             wasmExecuteContractPayload = keysignPayload.wasmExecuteContractPayload,
-            // Cosmos SignDoc artefacts MUST round-trip to peer devices. Without these, a relayed
-            // payload arrives with `signDirect`/`signAmino == null` on the peer, which then
-            // rebuilds
-            // a default (bank-send) signing input — its message hash diverges from the initiator's,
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #5470, opened 2026-08-01
fix: drop stale bundled Cloudflare intermediate cert pin

johnnyluo:
## Summary
- `network_security_config.xml` pinned `Cloudflare TLS Issuing ECC CA 1` as an extra trust anchor for `vultisig.com` (#2567). Cloudflare has since rotated the issuing intermediate for the zone to `Cloudflare TLS Issuing ECC CA 3`, making the bundled cert stale.
- Devices that rely on the app's bundled intermediate (instead of validating purely via system trust anchors) fail with `javax.net.ssl.SSLHandshakeException: Chain validation failed`, since the pinned cert no longer matches what's served.
- Verified live that `api.vultisig.com` now serves a complete chain (leaf → CA3 → SSL.com Transit ECC CA R2 → SSL.com TLS ECC Root CA 2022 → cross-signed by the long-trusted legacy `AAA Certificate Services`/Comodo root), so system trust anchors alone are sufficient.
- Removed the `vultisig.com` domain-config override and the now-unused raw cert resource; base-config (system trust anchors only) covers all domains.

## Test plan
- [x] `./gradlew :app:processDebugResources` succeeds with the updated config
- [x] Manual: confirm TLS to `api.vultisig.com` still succeeds on a real device/emulator after this change

## Summary by CodeRabbit

* **Security**
  * Removed the special network security configuration for `vultisig.com` and its subdomains.
  * The app now relies on the global network security settings for these connections.


rkokhatskyi (approved):
LGTM

PR #5477, opened 2026-08-02
fix(token detail): rest the sheet at a third and expand it as you scroll

### LINKED ISSUE #5474: Token detail opens as a full screen since the price chart landed — restore a partial-height sheet with a faded chart edge
On `main`, token detail no longer opens as a bottom sheet at all — it's a full-screen route you have to navigate back out of. **Not released yet:** #5462 merged 2026-07-30 13:07 UTC, ~12h after v1.0.116 was cut, so installed builds still show the old sheet.

**Cause** — #5462 (price chart + market data) changed both the presentation and the container:
- `ui/navigation/NavGraph.kt:177` — `dialog<TokenDetail>` → `composable<TokenDetail>`
- `ui/screens/TokenDetailScreen.kt` — the content used to be wrapped in `DottyBottomSheet`; it's now a `V3Scaffold` filling the screen.

**Ask:** put it back in a bottom sheet resting at ~1/3 height, expandable to full.

Note that restoring `DottyBottomSheet` as-is will *not* give you that — it hardcodes `rememberModalBottomSheetState(skipPartiallyExpanded = true)` (`v2/bottomsheets/DottyBottomSheet.kt:40`). It looked like a third before only because the sheet wraps its content and the content was four short rows; with the chart plus stats/extremes/info it would open full height — i.e. exactly the symptom, just via a sheet instead of a route. So the fix needs an explicit partial rest position:

- `skipPartiallyExpanded = false` with a ~1/3 rest anchor (Material3's partial state sits at ~half by default, so a third needs custom anchors), drag-up to expand to full.
- At the partial height, fade/blur the content at the sheet's bottom edge so the cut-off chart reads as "there's more" rather than as a clipped view.
- Swipe-down dismisses from the partial height — that's the main thing the full-screen route costs.

Same regression on iOS from its own chart PR: vultisig/vultisig-ios#5014.

**Done when:** opens partial by default; drag up expands to full and back; content visibly fades at the cut edge in the partial state; swipe-down dismisses without a back navigation.

aminsato:
Closes #5474

## Summary

#5462 (price chart + market data) moved token detail from a bottom sheet to a full-screen route — `dialog<TokenDetail>` → `composable<TokenDetail>`, and `DottyBottomSheet` → `V3Scaffold`. The screen has to be navigated back out of instead of swiped away. It merged ~12h after v1.0.116 was cut, so no released build shows it.

This restores the sheet as the hybrid @realpaaao asked for: **rests at a third, expands to full as you scroll, and is easy to close from the top.**

> do the 1/3 sheet as before and expand to full sheet when scrolled. Important is that the sheet is easily closed when at the top and scrolling down further.

That direction also answers @rkokhatskyi's objection on the earlier revision — that a modal is worse than a page for long content, because closing it means a drag back and a tap outside. It ends up full screen, and leaving is one continuous downward scroll rather than a two-step dance.

## Why this needed a new component

`ModalBottomSheet` cannot rest at a third. Material3 1.4.0 hardcodes the partial anchor inside a private `draggableAnchors` block:

```kotlin
if (sheetSize.height > (fullHeight / 2) && !sheetState.skipPartiallyExpanded) {
    PartiallyExpanded at fullHeight / 2f
}
```

No parameter, no `SheetState` hook, no detent list — half is the only rest position it can produce. So this adds `ExpandingBottomSheet`, an `AnchoredDraggable` sheet that owns its scroll.

The gesture handoff is the whole point of it. A nested-scroll connection spends an upward scroll on **expanding the sheet before the content moves at all**; once expanded, the reverse holds — the content scrolls back to its top, and the pull that keeps going collapses the sheet and then dismisses it. One continuous gesture in each direction, which is what "easily closed when at the top and scrolling down further" asks for.

## Rest height

The third is the design. The rest anchor is `max(window / 3, measuredContent + cutEdgeFade)`, where the measured block ends at the **action row** — the price/network card, the chart and the stats all sit below the fold. The measurement is only a floor: it raises the sheet on a screen too short, or a font scale too large, for the actions to clear the fade within a third, and never lets it open taller just because there is more to show.

Lands at ~36% on a 1080×2400 emulator, chartless and full alike.

An earlier revision measured the whole readable block (identity + balance + actions + price/network card) and rested at ~51% — half again, not a third. That is fixed here.

`CutEdgeFadeHeight` drops 96dp → 48dp and is now the clearance the measurement reserves, so the fade can never dim the action labels. At 96dp it did.

## Two things that weren't obvious

**`skipPartiallyExpanded = false` on its own does nothing.** `ModalBottomSheet` drives its own `show()` from a `LaunchedEffect` that runs *before the first layout*, when the sheet has no anchors and therefore no partial one to aim at. `show()` falls through to `Expanded`, and the anchor-change handler preserves that target once the anchors arrive. This is what made the bespoke sheet the cheaper path in the end, rather than a workaround stacked on `ModalBottomSheet`.

**The chart swallowed vertical drags.** `PriceChartView` scrubbed with `detectDragGestures`, which claims the pointer on slop in *any* direction and consumes every change, so a vertical drag starting on the canvas never reached the parent. Now `detectHorizontalDragGestures`; scrubbing is unchanged. Less critical now that the chart sits below the fold at rest, but it still breaks the drag once the sheet is open.

## Cross-platform

iOS shipped the same fix independently in vultisig-ios#5031 (merged 2026-08-03), and it is worth reading side by side because the two platforms landed in different places:

- **They measured a third and rejected it.** On iPhone 16 Pro the content above the chart is ~301pt and a third of the screen is ~276pt, which cuts through the action labels. They rest at `.fraction(0.50)` — ~46% of the screen — to clear the balance, the action row *and* the chart's price/change chip. Android's floor protects the action labels instead, so a third is safe here, but the sheet shows actions and no price where theirs shows both. **If product wants the two to match, this is the decision to make.**
- **They had to disable scrolling at the partial detent** (`.scrollDisabled`) because the `ScrollView` swallowed the expand drag — `.presentationContentInteraction(.resizes)` did not help; `.refreshable` was the blocker — and then rewind the scroll position on collapse so the user is not stranded mid-content. The nested-scroll handoff here makes both unnecessary.
- Both fade the cut edge rather than letting it read as clipped, which is the shared answer to "a partial sheet just opens below the fold".

## Notes

- `DottyBottomSheet`'s `skipPartiallyExpanded` parameter defaults to the old value, so `TokenAddressQrBottomSheet` is behaviourally untouched.
- The host window's dim is zeroed when the sheet is a navigation dialog destination, so it does not sit under the sheet's own scrim at full strength while the sheet slides away.
- Expanded stops below the status bar rather than under it, so the clock never overlaps the sheet header and the rounded top edge stays visible at every anchor.

## Test plan

- `:app:testDebugUnitTest`, `:app:lintDebug` and `ktfmt` pass.
- Verified on an emulator through the debug-only `token_detail_sheet_full` and `token_detail_sheet_loading` preview keys: opens at a third, scrolling up expands to full before the content moves, scrolling back down returns to the top then collapses then dismisses, and the cut-edge fade draws only while content is actually cut.
- Chartless tokens rest at the same third — the anchor is content-independent above the floor.
- A vertical drag started **on the chart** moves the sheet instead of being swallowed.

Worth a human pass on a short device and at a large font scale, where the measured floor rather than the fraction decides the rest height.

## Summary by CodeRabbit

* **New Features**
  * Token details now open in a polished, expandable bottom sheet.
  * Added an in-sheet explorer shortcut and improved spacing, sizing, and navigation-bar support.
  * Added smooth expanded and resting states with gesture, scrim, and dismissal support.
  * Added smoother chart scrubbing and consistent dotted bottom-sheet styling.

* **Accessibility**
  * Added localized close-action labels in multiple languages.

* **Tests**
  * Added coverage for smooth sheet expansion when content height changes.


rkokhatskyi (changes requested):
hey @realpaaao,

We agreed with @johnnyluo to make it full screen. Token details is super long and makes the UX worse in the modal as it is harder to close it (you need to drag it back and then tap outside of the modal - when testing I found this UX is bad compared to full screen). Also if we check top 5 wallets on the market  - they do full screen as well. This change was intentional from my side

cc: @aminsato

realpaaao:
I agree that a full-screen detail view is the right destination for long token data. A hybrid could retain a fast initial glance without restoring the old modal UX: open at roughly one-third height with the primary action and key information, then expand to full screen as the user scrolls. Fade the chart out during that transition; once the content is back at the top, a further pull-down collapses it again.

That gives the main data immediate focus, keeps a clear full-screen reading mode, and avoids requiring users to drag down and tap outside just to leave. cc @aminsato

aminsato:
@rkokhatskyi @realpaaao — I need a direction call before taking this further.

Where we are: `main` is a full-screen route, this PR restores a sheet resting at half height, and @realpaaao's suggestion above is a hybrid that rests at ~1/3 and expands to full screen as you scroll.

@rkokhatskyi — your close-affordance objection lands on the sheet-only version, but the hybrid ends up full screen, where back or swipe leaves it like any other page. Does that change your read, or do you still want a plain page?

One cost note so the choice is informed: the hybrid isn't a flag. Material3's `ModalBottomSheet` hardcodes its partial anchor at half the screen with no parameter or hook, so ~1/3 plus scroll-driven expansion means a bespoke `AnchoredDraggable` sheet — a new shared component rather than a tweak to this PR.

So: plain full-screen page, the half-height sheet as it stands, or the hybrid? Happy to build any of the three.

realpaaao:
@aminsato  Please do the 1/3 sheet as before and expand to full sheet when scrolled. Important is that the sheet is easily closed when at the top and scrolling down further.

aminsato:
Marked ready. Two things changed since the last review, and both are direction rather than polish.

**It now rests at a third and expands as you scroll** — @realpaaao's hybrid, which is not what the earlier revision on this PR did. The half-height sheet is gone, and so is the reasoning I gave for it: `ModalBottomSheet` genuinely cannot rest at a third, so this adds `ExpandingBottomSheet`, an `AnchoredDraggable` sheet that owns its scroll. An upward scroll is spent expanding the sheet before the content moves; scrolling back down returns the content to its top, then collapses the sheet, then dismisses it.

@rkokhatskyi your CHANGES_REQUESTED still stands from the half-height revision and I have not marked it addressed. I think the hybrid dissolves the objection behind it — you and @johnnyluo wanted a page because closing a modal over long content meant a drag back and a tap outside, and this ends up full screen with leaving as one continuous downward scroll. But that is your call, not mine, so it would be good to have your read before this goes further.

**iOS shipped the same fix independently** in vultisig/vultisig-ios#5031 (merged 2026-08-03) and landed on a different rest height. They measured a third on an iPhone 16 Pro at ~276pt against ~301pt of content above the chart — it cuts through the action labels — and rest at ~46% of the screen instead, showing the balance, the actions *and* the chart's price/change chip. Android is safe at a third because the rest anchor takes a floor from the measured action row, but the sheet shows actions and no price where theirs shows both. If the two platforms should match, that is worth deciding now rather than after both ship — details in the Cross-platform section of the body.

app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
```diff
@@ -3428,6 +3435,66 @@ private fun TokenDetailSheetLoadingPreview() {
...
+ * extremes and token info. That is the content that outgrows the screen, so this is the preview
+ * that actually exercises the partial rest position, the drag up to full, and the faded cut edge.
+ */
+@Composable
+private fun TokenDetailSheetFullPreview() {
+    val points =
+        List(48) { index ->
+            val price = 73.5 + sin(index / 5.0) * 6.0 + index * 0.12
+            ChartPointUiModel(
+                timestampMillis = 1_752_000_000_000L + index * 1_800_000L,
```
rkokhatskyi: What is this number for? Why that specific value?

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
```diff
@@ -0,0 +1,346 @@
...
+
+    LaunchedEffect(windowHeight, restHeight, expandedTop, cutEdgeFade) {
+        if (windowHeight <= 0) return@LaunchedEffect
+        val hiddenOffset = windowHeight.toFloat()
+        // A third of the window is the resting height the design asks for. The measured content
+        // only ever raises it, on a screen too short — or a font scale too large — for the part
+        // that has to be readable to clear the fade within a third; it never lets the sheet open
+        // taller than that just because there is more to show.
+        val visibleAtRest = maxOf(hiddenOffset * RestFraction, restHeight + cutEdgeFade)
+        state.updateAnchors(
```
rkokhatskyi: restHeight can grow after the sheet settles at Rest since neither LoadableValue call at `TokenDetailScreen.kt:160` nor `:169` sets maxLines, so a balance that wraps to two lines retriggers `ExpandingBottomSheet.kt:115`'s updateAnchors. AndroidX's trySnapTo writes offset synchronously with no animation, so the already visible sheet jumps to its new position instead of resizing smoothly.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
aminsato: You're right, and it's worse than a theory — I reproduced it. `updateAnchors` calls `trySnapTo`, which goes straight to `dragTo`: no animation, offset written in place. On a 1080x2400 emulator the edge teleports **526px in a single frame** when the resting height grows under a settled sheet.

Fixed in `bf20b265a`. The offset is read either side of `updateAnchors`. A moved offset means the snap actually landed, and that only happens when nothing else holds the sheet — an entrance, a drag or a fling keeps its own animation and blocks the snap — so it isolates exactly the case worth animating. `dispatchRawDelta` puts the edge back synchronously, which runs before that frame is laid out so the jump is never drawn, and `anchoredDrag` then walks it over on the same spec the sheet's other settles use.

I went for the mechanism rather than `maxLines = 1` on the two `LoadableValue`s, since wrapping is only one way the measured block can change height — a late-loading balance does it too, and so would a font-scale change.

There's a test on it now, `ExpandingBottomSheetTest`, which samples the edge frame by frame with the clock held. It fails on the pre-fix code with `expected:<1600.0> but was:<1074.0>`, so the regression is pinned rather than just described.

app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
aminsato: Fair — it was an epoch-millis base for the mock chart's points, and nothing said so. It only ever surfaces when you scrub the chart, which prints the date of the point under your finger.

Named in `bf20b265a`: `CHART_FIRST_POINT_MILLIS`, `CHART_POINT_INTERVAL = 30.minutes`, `CHART_POINT_COUNT`, with a line saying why it's a fixed instant rather than the clock — two captures of the preview then scrub to the same dates, so they stay comparable. Half-hourly across the 24H range the chart opens on. The date itself carries no meaning beyond being a plausible one.

aminsato:
@rkokhatskyi both of your comments are answered in the threads and fixed — the anchor one was real and reproducible, 526px of teleport in one frame, now pinned by a test that fails on the old code. Also picked up CodeRabbit's scrim nitpick: `detectTapGestures` ignores accessibility-dispatched actions, so the scrim carries a description and an `onClick` semantics action now.

One more since your review, reported separately: **opening the sheet sometimes flashed the screen black** for a frame or two before the sheet appeared. Not anything the sheet draws — it's the platform dim. This destination passes `decorFitsSystemWindows = false`, which themes the dialog with Compose's `FloatingDialogWindowTheme`, and that's the style carrying `android:backgroundDimEnabled`. Clearing the dim was already the intent here; the timing was wrong. Compose shows the dialog from a `DisposableEffect`, and effects run after the content composes, so a `LaunchedEffect` could only clear it *after* the window had been added with the dim and composited. Setting it during composition lands before the window is ever shown. Frame data from the report: two frames at zero luma, both before any part of the sheet had drawn.

Worth being straight about the limits of that last one — it's a race ("sometimes"), and it did not reproduce on my emulator in the run I recorded, so the fix rests on the frame evidence plus the theme source rather than on a before/after capture of the flash itself.

The direction question above is still the open one, and it's yours: this now rests at a third and expands to full as you scroll, so leaving it is one continuous downward scroll rather than the drag-back-and-tap-outside that you and @johnnyluo objected to. Your CHANGES_REQUESTED is still the one from the half-height revision — I haven't touched it.

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
```diff
@@ -0,0 +1,396 @@
...
+        // it, so a balance that loads late and wraps to a second line grows the resting height
+        // under a sheet the reader is looking at and its edge jumps. That write only lands when
+        // nothing else holds the sheet — an entrance, a drag or a fling keeps its own animation and
+        // leaves the offset alone — so an offset that moved here is precisely the case worth
+        // animating. Undoing it is synchronous and runs before this frame is laid out, so the jump
+        // is never drawn; the sheet then travels the distance.
+        val offsetAfterUpdate = state.offset
+        if (!offsetBeforeUpdate.isNaN() && offsetAfterUpdate != offsetBeforeUpdate) {
+            state.dispatchRawDelta(offsetBeforeUpdate - offsetAfterUpdate)
+            state.anchoredDrag(target) { anchors, latestTarget ->
```
rkokhatskyi: During a live finger drag, a restHeight change runs the new correction animation at `ExpandingBottomSheet.kt:157`, but `dispatchRawDelta` in the touch path never checks its mutex, so both write state.offset uncoordinated. Since target resolves to the nearest anchor to the drag position, not Rest, the sheet can dismiss or snap full screen mid gesture.

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
```diff
@@ -0,0 +1,396 @@
...
+        // A third of the window is the resting height the design asks for. The measured content
+        // only ever raises it, on a screen too short — or a font scale too large — for the part
+        // that has to be readable to clear the fade within a third; it never lets the sheet open
+        // taller than that just because there is more to show.
+        val visibleAtRest = maxOf(hiddenOffset * RestFraction, restHeight + cutEdgeFade)
+        // Where the sheet is headed has to be read before the rewind below moves it away from its
+        // anchor: read afterwards, it is whichever anchor happens to be nearest the old offset, and
+        // once the resting place climbs past the halfway mark of the window that is the bottom of
+        // the screen — so the sheet would leave rather than grow.
+        val target = state.targetValue
```
rkokhatskyi: A second restHeight change while the first correction animation at `ExpandingBottomSheet.kt:157` is still running can reopen the self dismiss bug 65d7c06bd just fixed, since target here reads a nearest anchor fallback off a frozen mid tween offset instead of the live dragTarget. Needs device confirmation of the exact coroutine cancellation timing.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
aminsato: Confirmed, and it's the worse of the two — reproduced at **947px of movement under a held finger** on a 1080x2400 emulator.

You're right about the mechanism down to the lock. In foundation 1.11.2, `dispatchRawDelta` (`AnchoredDraggable.kt:1245`) writes the offset through `anchoredDragScope.dragTo` without going near `dragMutex`, and the nested-scroll path is the one that uses it — so a scroll in the content owns the sheet while announcing nothing. `trySnapTo` (`:1259`) is a `tryMutate`, so it succeeds underneath, and the correction starts against a finger that has no way to cancel it. It's asymmetric in the other direction too: a drag arriving *after* the correction starts can't cancel it either, so the animation wins every frame it emits.

One correction to add to yours: `target` didn't merely fail to resolve to Rest, it resolved to the anchor nearest wherever the finger happened to be — `targetValue` is `dragTarget ?: calculateTargetValue(offset)` (`:871`) and `dragTarget` is null on that path. So mid-drag it was reading the drag position as an intention.

Fixed in `487785415` by removing the assumption rather than patching the symptom. The sheet now keeps its own `destination` instead of asking the state for one, anchor updates wait for the reader to let go — of the sheet via the drag `interactionSource`, of the content via `scrollState.isScrollInProgress` — and a travel already under way stands down when a finger arrives, since standing down is what following the finger looks like.

`aRestHeightThatGrowsUnderAFingerWaitsForItToLift` holds a gesture open with `down()` + `moveBy()`, changes the resting height, and asserts the sheet hasn't moved before `up()`. It fails on the pre-fix code with `expected:<1421.0> but was:<474.0>`.

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
aminsato: The self-dismissal is real and reproduces — a test on the pre-fix code fails with *"the sheet dismissed itself instead of resizing"*. You were right to want device confirmation of the cancellation timing; it isn't just possible, it's the ordering that actually happens.

One part of the mechanism is off, and it's worth stating because it changes where the hole is rather than whether there is one. `dragTarget` **is** set during the correction: `anchoredDrag(targetValue)` assigns it at `AnchoredDraggable.kt:1205`, so `targetValue` reads that and not a nearest-anchor fallback. The hole opens on the way out. A second height change cancels the effect the correction runs in, the `finally` at `:1218` clears `dragTarget`, and that cancellation resumes on the main dispatcher *before* the newly launched effect body does — so by the time the next pass reads `targetValue`, `dragTarget` is null and the offset is frozen mid-tween. Then `calculateTargetValue` picks the nearest anchor to that offset, and early enough in the travel the nearest anchor is the bottom of the screen. Same outcome you predicted, one step further along than described.

Fixed in `487785415` on both counts. The destination is no longer inferred from the offset at all, and the keyed effect is gone: a single `snapshotFlow { geometry }.collect` means a height that changes again queues behind the travel instead of cancelling it half way, so the frozen mid-tween offset stops existing as a state anything reads.

A third one turned up while checking this, same family: the no-target `anchoredDrag` (`:1154`) never clears the `dragTarget` that `updateAnchors` hands it, so anchors updated during a fling left `targetValue` stale for good. Also gone, since nothing consumes it now.

Worth flagging one thing the fix nearly broke, since it argues for the test rather than the reasoning: `destination` has to start on **Hidden**. The first `updateAnchors` snaps to it, so naming the resting place any earlier puts the sheet there outright and leaves the entrance animating a distance of nothing. The old code got Hidden for free from `targetValue`'s NaN fallback. I found that by writing the test, not by reading the code, and it's pinned now — `theSheetSlidesInRatherThanAppearingAtItsRestingPlace` fails on it with the edge sitting at 1600.0 for all ten sampled frames.

aminsato:
@rkokhatskyi both comments were real and both are fixed in `487785415`, with replies in the threads. Short version:

- The live-drag race is the worse one — **947px of movement under a held finger**. `dispatchRawDelta` takes no lock, so a scroll in the content owns the sheet invisibly and the correction animation can neither be cancelled by the finger nor see it.
- The mid-travel one reopens the self-dismissal exactly as you guessed. Your mechanism was one step off — `dragTarget` *is* set during the correction; it's the cancellation clearing it that opens the hole — but the outcome you predicted is the outcome I reproduced.
- A third turned up while checking: the no-target `anchoredDrag` never clears the drag target `updateAnchors` gives it, so anchors updated during a fling left `targetValue` stale permanently.

All three were the same root — the sheet asking `AnchoredDraggableState` where it was headed, when that question only has an answer while nothing is touching it. It keeps its own destination now, anchor updates wait for the reader to let go, and one collector replaces the keyed effect so a second height change queues behind the travel instead of cancelling it mid-tween.

`ExpandingBottomSheetTest` is at four tests, each verified to fail on the code it fixes. One of them exists because the fix nearly broke the entrance: the first `updateAnchors` snaps to the destination, so starting it at Rest rather than Hidden puts the sheet at its resting place outright with nothing left to animate. Caught by writing the test, not by reading the code.

The direction question is still the open one and it's still yours. This rests at a third and expands to full as you scroll, so leaving it is one continuous downward scroll rather than the drag-back-and-tap-outside you and @johnnyluo objected to. Your CHANGES_REQUESTED is still the one from the half-height revision — I haven't touched it. Also still unanswered: iOS shipped the same fix at **~46%** rest height (vultisig/vultisig-ios#5031) where this rests at a third, so if the platforms should match, better decided now than after both ship.

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
```diff
@@ -0,0 +1,493 @@
...
+ * Hands the sheet the anchors [geometry] asks for and takes it along to the one it belongs on,
+ * rather than leaving it where the old anchors put it.
+ *
+ * A sheet nothing else has hold of is moved to its new anchor by a write with no animation behind
+ * it, so a balance that loads late and wraps to a second line grows the resting height under a
+ * sheet the reader is looking at and its edge jumps. Undoing that write is synchronous and runs
+ * before the frame it landed in is laid out, so the jump is never drawn; the sheet then travels the
+ * distance instead.
+ */
+private suspend fun AnchoredDraggableState<ExpandingSheetValue>.moveAnchors(
```
rkokhatskyi, line 332: `moveAnchors` takes `geometry` as a plain parameter that never refreshes during its own multi frame `travelTo` animation, unlike `destination` which is a live closure. A second restHeight change landing during that tween queues a whole extra `moveAnchors` call, so the sheet visibly settles at the stale height then immediately reanimates to the correct one.

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
```diff
@@ -0,0 +1,493 @@
...
+    // taller that it was on its way out. It starts on Hidden because that is where the sheet is
+    // before it slides in: name Rest any earlier and the first anchors the sheet is given put it
+    // there outright, with nothing left to animate.
+    var destination by remember { mutableStateOf(ExpandingSheetValue.Hidden) }
+
+    // Two ways a reader can have hold of the sheet, neither of them visible from the state itself:
+    // a drag on the sheet takes its lock but announces nothing else, and a scroll in the content
+    // drives the sheet through dispatchRawDelta, which takes no lock at all.
+    val isSheetDragged = dragInteractions.collectIsDraggedAsState()
+    val isHeldByReader = { isSheetDragged.value || scrollState.isScrollInProgress }
```
rkokhatskyi: `isHeldByReader` never detects an in flight `flingToAnchor`: dispatchPreFling runs before doFlingAnimation's own scroll block starts, so isScrollInProgress reads false throughout it. A restHeight change landing there changes the shared anchors mid fling, restarting the fling with stale velocity so it can land on the wrong anchor, undoing an intended dismiss.

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
```diff
@@ -0,0 +1,493 @@
...
+    // that answer an accident of timing — the reading that once told a sheet which had only grown
+    // taller that it was on its way out. It starts on Hidden because that is where the sheet is
+    // before it slides in: name Rest any earlier and the first anchors the sheet is given put it
+    // there outright, with nothing left to animate.
+    var destination by remember { mutableStateOf(ExpandingSheetValue.Hidden) }
+
+    // Two ways a reader can have hold of the sheet, neither of them visible from the state itself:
+    // a drag on the sheet takes its lock but announces nothing else, and a scroll in the content
+    // drives the sheet through dispatchRawDelta, which takes no lock at all.
+    val isSheetDragged = dragInteractions.collectIsDraggedAsState()
```
rkokhatskyi: `dragInteractions` never receives a DragInteraction since the child verticalScroll always wins touch before anchoredDraggable's own gesture node runs, so `isSheetDragged` stays false and `isHeldByReader` reduces to `scrollState.isScrollInProgress` alone. The interactionSource wiring and the OR branch can be removed.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
aminsato: Confirmed, and it's the queueing from `487785415` doing exactly what you'd expect of it. `snapshotFlow` conflates rather than drops, so a second height landing mid-travel isn't lost — but it only gets applied after the travel it's waiting on has run to a height that has already been withdrawn. The sheet arrives there, settles, and then sets off again. Two `LoadableValue`s resolving a few hundred milliseconds apart is enough to see it.

`aRestHeightSupersededMidTravelIsDroppedRatherThanArrivedAtFirst` goes short → tall → mid and samples the edge frame by frame. On the pre-fix code:

```
[1593, 1566, 1513, … , 480, 475, 474, 474, 478, 492, 520, … , 963]
```

474 is the tall anchor. It reaches it, sits there for two frames, and turns round.

Making `geometry` live the way `destination` is doesn't work, and it's worth saying why: `anchoredDrag` re-runs its block whenever the anchors change (`AnchoredDraggable.kt:1159`, `:1206`) and the block animates from a captured `from`, so re-reading the anchors mid-tween restarts the travel from where it *began* rather than from where it has got to — a jump backwards in place of a settle at the wrong height.

So the travel stands down instead. `moveAnchors` takes an `isOutOfDate`, the travel abandons the moment a newer measurement arrives, and the next pass carries on from wherever the sheet is — the same mechanism as the hand-off to a finger, and for the same reason: arriving is only worth it if the place is still the place. There's also an early return before `updateAnchors` for the case where the height moves on while we're still waiting for the reader to let go, so we don't swap anchors for a height nobody is asking for any more and then have to walk the sheet back from it.

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
aminsato: Confirmed, and the source is exact on the mechanism: `dispatchPreFling` is called at `Scrollable.kt:864`, outside the `scroll {}` block that `doFlingAnimation` opens at `:888`, and `dispatchPostFling` (`:870`) runs after that block has closed. So `isScrollInProgress` reads false across the whole of the sheet's own fling — through both entry points, not only the pre-fling one.

The landing is worse than "wrong anchor". `updateAnchors` can't snap — the fling holds `dragMutex`, so `trySnapTo` fails and the offset comparison returns early — but the anchors have already been swapped, and `restartable(inputs = { anchors })` re-runs the fling block with the velocity it began with. `snapFlingBehavior` then recomputes the target from wherever the sheet has got to, and once the resting anchor has moved above that point, the next anchor in the fling direction is Hidden. A flick aimed at the resting place dismisses.

`aRestHeightThatGrowsMidFlingDoesNotSendTheSheetPastWhereItWasThrown` throws the sheet from full back towards rest, catches it 12 frames in and grows the resting height. It fails on the pre-fix code with *"the sheet was flung off the screen rather than to its anchor"*.

Fixed in `ebe9e02a7`: the sheet's own fling raises a flag `isHeldByReader` reads, so an anchor update waits it out. One thing fell out of that and is worth flagging on its own — where the fling landed has to be recorded *before* the flag comes down. The settle collector is a frame behind, and the anchor update the fling has been holding up reads the destination the instant it is released; a destination still naming where the sheet was before the flick takes it straight back there.

One gap left, deliberately. `onDragStopped` launches `onScrollStopped` in a coroutine (`Scrollable.kt:387`), so a dispatch-sized window remains between the drag's scroll session closing and the flag going up. An update landing exactly there still swaps anchors, but the outcome is bounded: the snap lands, it is undone synchronously, and the fling then takes the mutex at equal priority and cancels the correction — a redundant correction, not a wrong-anchor dismissal. Closing it would mean latching "held" from the first user-input scroll until `onPostFling` returns, and a latch that never receives its release event freezes the anchors for good. That is a worse failure than the one it prevents.

app/src/main/java/com/vultisig/wallet/ui/components/v2/bottomsheets/ExpandingBottomSheet.kt
aminsato: Confirmed, and by probe rather than by argument — I mirrored the structure (a parent `anchoredDraggable` with an interaction source, a `verticalScroll` child filling it) and collected from the source across a swipe up, a swipe down and a slow drag. Nothing arrives. The child claims the Main pass before the gesture underneath it is offered anything, so `isSheetDragged` was false for the whole of the sheet's life and `isHeldByReader` was `scrollState.isScrollInProgress` wearing a second condition.

Removed in `ebe9e02a7` — and I took the `anchoredDraggable` modifier with it rather than only the wiring. An inert gesture with nothing watching it is the same hole one round later: it costs nothing today and becomes an unseen holder of the offset the moment someone changes how the content is composed. With it gone, every touch reaches the sheet through the nested-scroll connection, which is what turns `isHeldByReader` into an answer rather than a guess — two conditions over paths the sheet owns, instead of a survey of the gestures that might be running.

`aPullDownAtTheTopTakesTheSheetOffTheScreen` pins the gesture that removal would have broken if the reading were wrong: the pull-down close, which only ever reaches the sheet as the part of a scroll the content had no use for.

aminsato:
@rkokhatskyi all three were real and all three are fixed in `ebe9e02a7`, with the detail in the threads. Short version:

- **The fling was the worst of them.** `dispatchPreFling` runs outside the scroll session the child opens for its own fling and `dispatchPostFling` after it has closed, so `isScrollInProgress` is false across the whole of the sheet's — anchors swapped there restart it with the velocity it began with, and the next anchor in that direction is the bottom of the screen. The test fails on the old code with *"the sheet was flung off the screen rather than to its anchor"*.
- **The stale geometry showed up exactly as you described it**: the edge climbs to the withdrawn height, sits there for two frames, then turns round. The measurement can't just be read live — `anchoredDrag` restarts its block on an anchor change and the block animates from a captured offset — so the travel stands down instead and the next pass carries on from where the sheet got to.
- **The interaction source was dead**, confirmed with a probe. I took the `anchoredDraggable` modifier with it rather than just the wiring: an inert gesture with nothing watching it is the same hole one round later. Everything now reaches the sheet through the nested-scroll connection.

7 androidTests, all green; each new one verified to fail on the code it fixes. One residual is written up in the fling thread rather than left implied.

Still outstanding from before, and it's the only thing in the way now: you never answered on direction. Paaao asked for this hybrid over the plain full-screen page, and the close-affordance objection behind your original CHANGES_REQUESTED does dissolve here — the sheet ends up full screen, and a pull-down at the top closes it in one gesture rather than the drag-back-then-tap-outside dance. If you're still against it I'd rather settle that than keep polishing a component we might not want.

rkokhatskyi (approved):
LGTM

PR #5486, opened 2026-08-03
fix(tron): drop the memo fee from freeze/unfreeze estimates

### LINKED ISSUE #5481: [Fix] Freeze/Unfreeze fee estimate wrongly includes a ~1 TRX phantom "memo fee" that's never actually charged on-chain
# [Fix] Freeze/Unfreeze fee estimate wrongly includes a ~1 TRX phantom "memo fee" that's never actually charged on-chain

## Problem
`TronStakingService.applyStakingMemo()` writes an internal routing signal (`"FREEZE:BANDWIDTH"`/`"UNFREEZE:ENERGY"`) into the shared memo field state on every Freeze/Unfreeze. `TronFeeService.calculateNativeTrxFee(hasMemo = !memo.isNullOrEmpty())` then adds the real ~1 TRX memo-fee chain parameter whenever `hasMemo` is true — which is always true for freeze/unfreeze. But `TronHelper.buildFreezeBalanceV2`/`buildUnfreezeBalanceV2` never actually call `.setMemo(...)` — the string is parsed for its prefix and discarded. Per Tron's real chain rule, the memo fee is only charged when `raw_data.data` is non-empty, so the extra ~1 TRX is never actually charged on-chain — it's a pure estimation-side phantom.

## Expected Behavior
The fee estimate for Freeze/Unfreeze should not include a memo fee, since no memo is ever written to the actual transaction.

## Steps to Reproduce
1. Open Freeze or Unfreeze for TRX.
2. Observe the fee estimate is inflated by ~1 TRX over the true near-zero cost.
3. With a liquid TRX balance between the real fee and the phantom ~1 TRX, submitting Unfreeze is blocked by an "insufficient balance for gas" error before the MPC ceremony even starts, despite the real on-chain fee being able to clear easily.

## Solution
Pass `memo = null` for the fee-estimation `Transfer` model on `FREEZE_TRX`/`UNFREEZE_TRX` in `GasFeeOrchestrator.kt` (and `AmountFractionManager.kt`'s MAX/percentage calculation), or have `TronFeeService` special-case `hasMemo` off whenever the internal staking-memo prefix matches.

### Must Do
- [ ] Exclude the memo fee from Freeze/Unfreeze fee estimation
- [ ] Fix the corresponding MAX/percentage-amount reservation in `AmountFractionManager`

### Must NOT Do
- Don't change memo-fee behavior for genuine memo-bearing sends
- Don't change the internal staking-memo routing signal itself, only how it's treated for fee estimation

## Acceptance Criteria
- [ ] Freeze/Unfreeze fee estimate no longer includes the phantom ~1 TRX memo fee
- [ ] New test covers the frozen-zero/hasMemo-true case explicitly

rkokhatskyi:
Closes #5481

## Summary

Tron Freeze/Unfreeze carry their resource type through the send form's memo field (`FREEZE:BANDWIDTH`, `UNFREEZE:ENERGY`), which `TronHelper` parses to pick between `FreezeBalanceV2` and `UnfreezeBalanceV2`. Those two are the only branches of `getPreSignedInputData` that never call `setMemo` — the string is read for its prefix and discarded. Tron only burns the memo fee when `raw_data.data` is non-empty, so the ~1 TRX `TronFeeService` added for every freeze/unfreeze was never actually charged.

The fee gate now asks the same question the signing dispatch asks, through one shared `tronStakingIntent(coin, toAddress, memo)`, so the estimate and the signed transaction cannot drift apart.

## Why it isn't only cosmetic

- MAX/percentage on Freeze reserved the phantom 1 TRX out of the selectable balance, so a max freeze quietly left ~1 TRX behind.
- Unfreeze validates liquid TRX against the estimate (`DefaultSendStrategy`), so a vault with most of its TRX frozen could be blocked from unfreezing with "insufficient balance for gas" over a fee that does not exist.

Both are fixed by the estimate itself — `GasFeeOrchestrator` and `AmountFractionManager` needed no change, they just consume the corrected fee. The same holds for the co-signer: Tron falls through `computeJoinKeysignNetworkFee`'s `else` branch and the joiner re-estimates via `TronFeeService`, so fixing at the service keeps both devices showing the same number. Fixing at the two send-form call sites instead would have left the joiner on the old value.

## Signed bytes are untouched

Tron's `feeLimit` comes from `BlockChainSpecificRepository`, which uses the constant `TRON_DEFAULT_ESTIMATION_FEE` (800,000) for native tokens and never consults the fee service. The keysign preimage is unchanged, so this is safe mid-flight for an in-progress ceremony.

## Notes

- The two builders now take the parsed `TronResourceType` instead of re-deriving the grammar with `removePrefix("FREEZE:")` + a hardcoded `require`. `resource.name` is byte-identical to what the old parse produced, since the regex only matches enum names. The dispatch's inner `when` is exhaustive, so a new `TronStakingOperation` fails the build rather than silently falling through.
- A genuine user memo on a native TRX send still pays the memo fee; only the internal routing signal is exempt.
- MAX on Freeze now offers the full balance, since the true fee is zero while the free bandwidth quota covers the transfer. That is the same exposure a plain MAX TRX send already has.

## Testing

- `./gradlew :data:testDebugUnitTest` — 2614 tests, all passing
- `./gradlew :app:testDebugUnitTest --tests "*Tron*" --tests "*SendForm*" --tests "*GasFeeOrchestrator*" --tests "*DefaultSendStrategy*" --tests "*AmountFraction*" --tests "*JoinKeysign*" --tests "*JoinSend*"` — 142 tests, all passing
- `./gradlew :data:lintDebug`, `:data:ktfmtFormat`

New `TronMemoFeeMatchesSignedMemoTest` drives both paths from one input table and asserts the fee prices a memo exactly when the built `SigningInput` carries one, over 11 inputs including the lowercase/`FREEZE:`/trailing-space near-misses. Reverting just the fee gate fails it plus two `TronFeeServiceTest` cases.

No screenshots: the diff contains no UI, layout, or string files. The only user-visible difference is a fee figure derived from live TronGrid state and the vault's on-chain bandwidth, so a `PreviewActivity` capture would only show numbers hand-written into the mock rather than evidence of the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected TRON fee estimation so staking-related memos remain fee-free.
  * Ensured genuine transfer memos continue to incur the applicable memo fee.
  * Improved validation for staking transfers, including destination and memo requirements.

* **Tests**
  * Added coverage for valid and invalid staking operations, memo handling, signed transactions, and full-balance TRX staking scenarios.


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TronHelper.kt
```diff
@@ -31,25 +34,27 @@ class TronHelper(
...
-                TRON_STAKING_MEMO_REGEX.matches(keysignPayload.memo) &&
-                keysignPayload.memo.startsWith("FREEZE:") ->
-                buildFreezeBalanceV2(keysignPayload, tronSpecific)
-            keysignPayload.coin.isNativeToken &&
-                keysignPayload.coin.address == keysignPayload.toAddress &&
-                keysignPayload.memo != null &&
-                TRON_STAKING_MEMO_REGEX.matches(keysignPayload.memo) &&
-                keysignPayload.memo.startsWith("UNFREEZE:") ->
-                buildUnfreezeBalanceV2(keysignPayload, tronSpecific)
+            staking != null ->
```
johnnyluo: Please don't remove the explicit `startsWith("FREEZE:")` / `startsWith("UNFREEZE:")` checks here — this can cause a cross-device keysign issue.

johnnyluo (changes requested):
Please don't remove the explicit FREEZE:/UNFREEZE: prefix checks — this can cause a cross-device keysign issue.

johnnyluo (approved):
LGTM

PR #5509, opened 2026-08-03
Send: adjust a typed amount to balance − fee instead of dead-ending on "Insufficient balance"

### LINKED ISSUE #5493: Send: auto-adjust amount to balance − fee when amount + fee exceeds balance, instead of erroring (iOS #5020 sibling)
## What
Sibling of vultisig/vultisig-ios#5020. When a manually-typed amount is within balance but `amount + fee` exceeds it, the send flow hard-errors ("Insufficient balance") instead of auto-adjusting the amount down to `balance − fee` and showing the adjusted value.

Android already has the right precedent for max sends: `DefaultSendStrategy.kt:190-213` re-derives the amount from the current balance and fee at submit time and clamps down (`coerceAtMost(available)`) — but only when `isMaxAmount` (typed amount equals the max snapshot). A user who types a near-max amount by hand gets the error dead-end.

## Ask
Extend the clamp to the non-max native-coin case when `amount ≤ balance < amount + fee`: adjust to `balance − fee`, update the visible amount before Verify (never sign a different amount than displayed), and keep the error for the case where the amount alone exceeds balance.

## Where
- `app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt:190-213` (existing max-only clamp)
- Balance/fee validation producing the error: `send_error_insufficient_native_balance_with_fees` / `send_error_insufficient_balance` usages in the send submit path

aminsato:
Closes #5493

Sibling of vultisig/vultisig-ios#5020.

## Problem

Type an amount that fits your balance but not `amount + fee`, and Send dead-ends on **"Insufficient ETH balance"**. There is no way forward from there: the fee is only known once the form has an amount, so the user is left guessing a smaller number until one sticks.

Submit already handled the mirror-image case — a stale **Max** snapshot whose balance or fee moved after the tap gets re-derived from the current values and clamped down (#5316). A hand-typed near-max amount got the error instead.

## Fix

Extend that same clamp to the typed case, and reflect the result back into the amount fields.

`DefaultSendStrategy` now routes both cases through `clampToSpendableBalance()`, which clamps to `GetAvailableTokenBalanceUseCase` — the one place that already knows what "spendable" means per chain (nets the fee, plus Polkadot's existential deposit; XRP's account reserve is netted upstream in `RippleApi.getBalance`). It only ever reduces, so a send can never exceed what the user asked for.

Three things it deliberately does **not** touch:

- **An amount over the balance itself** — that's a real over-entry, not a fee edge, and still errors.
- **Non-native tokens** — gas comes out of the native coin, so a token balance is fully spendable and an amount equal to it goes out unchanged.
- **DeFi flows** — staking/unbond/frozen TRX carry different balance semantics and compute their own max.

The clamp runs before the Cardano minimum-send, UTXO-dust and Ripple-reserve checks, so the adjusted amount is re-validated by all of them rather than slipping past.

### The adjustment is made visible

`showAdjustedAmount()` writes the clamped value back into the token amount field and scales the fiat field by the same ratio. This is load-bearing, not cosmetic: `Transaction.fiatValue` is read straight off that fiat field, so a clamp that skipped it left Verify showing a fiat figure that no longer matched the token amount beside it — a pre-existing wart on the Max path that this fixes too. It also re-snapshots the max when the clamp came from a Max tap, so the form's 100% chip doesn't deselect itself just because the fee moved.

### Left alone on purpose

TerraClassic bank denoms (USTC) hit the same dead-end and also pay their fee in their own denom — but their burn tax means `balance − gasFee` isn't the real ceiling, so clamping there could still fail at broadcast. They keep erroring.

## Test plan

`DefaultSendStrategyTest` — 15/15 green, 325/325 across the send package.

Three new cases:

- a hand-typed native amount adjusts to `balance − fee`, asserted on the persisted `tokenValue`, on both visible fields, and on `Transaction.fiatValue`
- a non-native amount equal to the balance goes out unchanged
- a DeFi native amount is not adjusted and still errors

One reworked: `submit blocks native token send exceeding the available balance` → `…exceeding the balance itself`. Its old fixture (balance 1.0 ETH, entered 0.5, available 0.4) is exactly the case that now legitimately clamps, so the balance dropped to 0.4 to keep it a genuine over-entry.

### On device

1. Chain row → **Send** → paste a recipient.
2. Type your **full native balance** by hand — no Max, no % chip.
3. Wait for the fee row, tap **Continue**.

The amount field visibly drops to `balance − fee`, the fiat line drops with it, and Verify opens on that reduced amount. Typing *more* than the balance still errors; tapping 100% behaves as before with the chip still selected; a USDC send of your exact full balance is untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Send amounts are automatically adjusted when network fees reduce the available balance.
  * Token and fiat values remain synchronized with precise downward rounding.
  * Text-field cursor positions are preserved during updates.
  * Maximum-amount selections stay consistent after recalculation.
  * DeFi transfers and non-native token amounts remain unchanged when balances are insufficient.
  * UTXO transactions are replanned when amount adjustments are required.
  * Effective, chain-appropriate gas fees are used for balance validation and estimation.
  * Additional validation now covers minimum send amounts and post-adjustment balance limits.


app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -454,6 +450,66 @@ internal class DefaultSendStrategy(
...
+        account: Account,
+        balance: BigInteger,
+        gasFee: TokenValue,
+        isMaxAmount: Boolean,
+    ): BigInteger {
+        if (defiTypeProvider() != null) return entered
+        // Non-native tokens pay gas in the native coin, so their shortfall is never fee-caused.
+        val isFeeOnlyShortfall = account.token.isNativeToken && entered <= balance
+        if (!isMaxAmount && !isFeeOnlyShortfall) return entered
+        val available = getAvailableTokenBalance(account, gasFee.value)?.value ?: return entered
```
rkokhatskyi, line 512: clampToSpendableBalance and the later balance recheck both size against gasFee.value, which adjustGasFee never updates for GasSettings.Eth, while applyGasSettings does patch the signed maxFeePerGasWei and gasLimit. Raising Advanced Gas Settings then sending near-full balance yields a clean adjusted amount that can still be rejected at broadcast for insufficient funds.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -454,6 +450,66 @@ internal class DefaultSendStrategy(
...
+        return entered.coerceAtMost(available)
+    }
+
+    /**
+     * Reflects an [adjusted] amount back into the amount fields so the form — and the Verify screen
+     * built from them — shows what will actually be signed. The user must never sign an amount they
+     * weren't shown. The fiat mirror is scaled by the same ratio instead of re-fetched so it stays
+     * consistent even before AmountManager's price-driven conversion catches up.
+     */
+    private fun showAdjustedAmount(
```
rkokhatskyi, line 553: showAdjustedAmount's ratio-scaled fiat write races AmountManager.collectConversion, which shares the same field and scope and recomputes fiat from live price at submit's later suspension points, blanking the field on a price-fetch failure. Transaction.fiatValue reads that field afterward, so Verify can show a wrong or $0.00 estimate next to a correctly signed crypto amount.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -187,30 +189,24 @@ internal class DefaultSendStrategy(
...
-                            }
-                        }
+                        clampToSpendableBalance(
+                            entered = enteredAmountInt,
+                            account = selectedAccount,
+                            balance = selectedTokenValue.value,
+                            gasFee = gasFee,
+                            isMaxAmount = isMaxAmount,
+                        )
+                    if (tokenAmountInt < enteredAmountInt) {
```
rkokhatskyi: showAdjustedAmount rewrites the amount and fiat fields as soon as the clamp reduces them, before the Cardano/BTC/XRP post-clamp validators run. A 1.5 ADA balance with a 1.45 ADA fee clamps to 0.05, gets displayed, then validateCardanoUTXORequirements rejects it for being under the 1.4 ADA minimum, leaving the field on an amount the user never typed.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -187,30 +189,24 @@ internal class DefaultSendStrategy(
...
-                    // correcting downward after network hydration, or EVM gas rising). A stale-high
-                    // snapshot then exceeds what's actually spendable and submit validation rejects
-                    // it as "insufficient" — which is why a max send can force the user to leave
-                    // some tokens behind. Re-derive the amount from the CURRENT balance and fee,
-                    // clamping DOWN only so we never send more than the user saw. Standard sends
-                    // only: DeFi flows (staking/unbond/frozen TRX) carry different balance
-                    // semantics and compute their own max.
+                    val enteredAmountInt =
+                        tokenAmount.movePointRight(selectedToken.decimal).toBigInteger()
                     val tokenAmountInt =
```
rkokhatskyi, line 210: applyBitcoinPlan only rebuilds the WalletCore plan when planBtc.value is null, but collectPlanFee already fills planBtc.value on every keystroke for the raw, pre-clamp amount, with no debounce gating that assignment. validateBtcLikeAmount then checks the newly clamped amount against that stale over-fee plan and throws insufficient_utxos_error, defeating this fix for hand-typed full-balance BTC/LTC/DOGE/BCH/DASH/ZEC sends.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -454,6 +450,66 @@ internal class DefaultSendStrategy(
...
+    private fun showAdjustedAmount(
+        adjusted: BigInteger,
+        entered: BigInteger,
+        token: Coin,
+        isMaxAmount: Boolean,
+    ) {
+        val adjustedDecimal = BigDecimal(adjusted).movePointLeft(token.decimal).stripTrailingZeros()
+        // The clamped value is still the true maximum, so re-snapshot it — otherwise the form's
+        // "100%" selection would drop just because the fee moved.
+        if (isMaxAmount) {
```
rkokhatskyi, line 562: The isMaxAmount gate around amountManager.markMax has no test coverage: the two Max-path tests only assert the clamped Transaction value, never the field text or a markMax call, and the hand-typed test never checks whether markMax was skipped. A regression that always, never, or wrongly invokes it would pass the whole suite unnoticed.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Confirmed and fixed in f724e70e6. `adjustGasFee` has branches for `GasSettings.UTXO` and Cardano but none for `GasSettings.Eth`, so a raised cap never reached `gasFee` while `applyGasSettings` still patched the signed `maxFeePerGasWei`/`gasLimit` — the adjusted amount looked clean and reserved ~1/3 of what the tx actually spends.

Added `withEvmGasSettings()`, reserving `maxFeePerGasWei × gasLimit` (the same worst-case figure `gasFeeToEstimatedFee` already shows the user), and routed both the clamp and the native balance check through it rather than only the clamp — otherwise the check would still wave through an amount that cannot pay the raised fee.

Covered by `submit sizes the adjustment on raised advanced gas settings, not the stale gas fee`: 200 gwei × 21000 on a 1.0 ETH balance adjusts to 0.9958, not the 0.9985 the stale estimate gives.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Right, and it was worse than a race on the write — the read happened after the `withContext(Dispatchers.IO)` `getSpecific` hop, so `AmountManager.handleTokenInput` had a full I/O round-trip to land in, and `convertValue` returning null on a price-fetch failure blanks the field outright.

Fixed by removing the read entirely: the fiat mirror is now captured once, before the clamp, and carried down to `Transaction.fiatValue` (scaled by the same ratio when the amount is adjusted). Nothing re-reads that field past a suspension point.

Worth noting this closes the hole on the unclamped path too — every send was building `fiatValue` from a post-I/O read of a field another coroutine owns, not just adjusted ones.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Agreed — the ADA case is exactly right, and stranding the form on a number the user never typed is worse than the error they started with.

`showAdjustedAmount` now runs after `transactionRepository.addTransaction`, i.e. once every balance/dust/reserve validator has passed and the send is committed. On any rejection the fields keep what was typed. The clamp itself still happens early, because `getSpecific` and the BTC plan both need the final amount.

Covered by `submit leaves the amount fields untouched when a post-clamp validator rejects the send` — a 20 XRP entry clamps to 19.9996, then the destination-tag validator rejects it, and both fields still read `20`/`40`. I used the XRP tag validator rather than the Cardano floor because `Chain.Cardano.getDustThreshold` goes through WalletCore JNI, which is not reliably loadable in a plain JVM unit test.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Confirmed — `collectPlanFee` assigns `planBtc.value = plan` on every emission of the amount field with no debounce gating that write, and `applyBitcoinPlan` only rebuilt when the value was null. So the clamped amount got validated against a plan built for the pre-clamp amount, and `validateBtcLikeAmount` threw `insufficient_utxos_error` — the fix would have been dead on arrival for exactly the hand-typed full-balance UTXO send it targets.

`applyBitcoinPlan` now takes `forceReplan`, set when the clamp reduced the amount, and re-plans for the amount actually being sent. That also refreshes `planFee`, so the fee shown on Verify matches the re-planned transaction instead of the discarded one.

Covered by `submit re-plans a clamped UTXO send instead of reusing the pre-clamp plan`, which seeds a stale plan and asserts `getPlan` is called with the clamped value.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Fair — the gate was asserted nowhere. Both Max tests now check the field text and `verify { amountManager.markMax(BigDecimal("0.9985")) }`, and the hand-typed test asserts `verify(exactly = 0) { amountManager.markMax(any()) }`. A regression that always, never, or wrongly fires it now fails.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -454,6 +450,66 @@ internal class DefaultSendStrategy(
...
+        isMaxAmount: Boolean,
+    ): BigInteger {
+        if (defiTypeProvider() != null) return entered
+        // Non-native tokens pay gas in the native coin, so their shortfall is never fee-caused.
+        val isFeeOnlyShortfall = account.token.isNativeToken && entered <= balance
+        if (!isMaxAmount && !isFeeOnlyShortfall) return entered
+        val available = getAvailableTokenBalance(account, gasFee.value)?.value ?: return entered
+        if (available <= BigInteger.ZERO) return entered
+        return entered.coerceAtMost(available)
+    }
```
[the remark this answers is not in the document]
aminsato: Not taking this one. The two calls are `GetAvailableTokenBalanceUseCaseImpl`, which is pure arithmetic over the `Account` already in hand — no I/O and no network, so there is nothing to save.

The TOCTOU argument points the other way: the suggested `ClampedAmount` caches the balance *across* the `withContext(Dispatchers.IO)` `getSpecific` call, so the amount would be validated against a reading taken before that hop rather than after it. Fetching again after the I/O is the narrower window, not the wider one. Both calls also read the same `gasFee` captured once by `accountValidator.validate()`, so they cannot disagree about the fee.

What was genuinely wrong nearby is that both sites sized against `gasFee` while Advanced Gas Settings never reach it — fixed in f724e70e6.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -454,6 +450,66 @@ internal class DefaultSendStrategy(
...
+
+        val enteredFiat = fiatAmountFieldState.text.toString().toPlainBigDecimalOrNull()
+        if (enteredFiat == null || enteredFiat.signum() <= 0 || entered.signum() <= 0) return
+        val adjustedFiat =
+            enteredFiat
+                .multiply(BigDecimal(adjusted))
+                .divide(BigDecimal(entered), token.decimal, RoundingMode.DOWN)
+                .stripTrailingZeros()
+        fiatAmountFieldState.setTextAndPlaceCursorAtEnd(adjustedFiat.toPlainString())
+    }
```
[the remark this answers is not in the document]
aminsato: Added the non-terminating-ratio test — `submit bounds the adjusted fiat when the clamp ratio does not divide evenly` asserts the fiat string carries no more than the token’s own precision and that the field and `Transaction.fiatValue` agree.

Kept the scale at `token.decimal` rather than switching to `Currency.defaultFractionDigits`. `AmountManager.handleTokenInput` writes that same field at `token.decimal` moments later, so a different scale here would make this write visibly disagree with the one that immediately follows it; and Verify renders through `FiatValueToStringMapper`, which formats to the currency’s digits anyway, so the long tail never reaches the user. Changing the app-wide fiat scale is a reasonable idea but belongs in its own PR — it touches the conversion every keystroke goes through, not just the adjusted path.

aminsato:
Thanks @rkokhatskyi — all five verified against the code and all five were real. Pushed as f724e70e6.

| Comment | Fix |
|---|---|
| Advanced Gas Settings never reach `gasFee` | `withEvmGasSettings()` reserves `maxFeePerGasWei × gasLimit`, used by the clamp **and** the balance check |
| Fiat write races `AmountManager` | fiat captured once up front and carried to `Transaction.fiatValue`; the field is never re-read past a suspension point |
| Fields rewritten before the validators | `showAdjustedAmount` moved after `addTransaction` — a rejection leaves what the user typed |
| Stale `planBtc` from `collectPlanFee` | `applyBitcoinPlan(forceReplan)` re-plans for the clamped amount, refreshing `planFee` too |
| `markMax` gate untested | both Max tests assert `markMax(0.9985)` + field text; hand-typed asserts `exactly = 0` |

The gas-settings one was the sharpest — sized on the stale estimate, a 200 gwei cap adjusted a 1.0 ETH balance to 0.9985 instead of 0.9958, so the "fix" would have handed back a clean number that fails at broadcast. Both call sites now reserve the real figure.

The fiat one turned out broader than the adjusted path: the read sat after the `getSpecific` I/O hop, so *every* send was building `fiatValue` from a post-suspension read of a field `AmountManager` blanks on price-fetch failure. That is closed for all sends now, not just clamped ones.

19/19 in `DefaultSendStrategyTest`, none skipped; send + deposit packages green. Four new tests, one per finding above.

One I did not take, from CodeRabbit — caching the available balance across the `getSpecific` I/O hop to save a second `getAvailableTokenBalance` call. It is pure arithmetic over an `Account` already in hand, so there is nothing to save, and caching it across the hop widens the TOCTOU window rather than closing it. Reasoning left inline.

Device verification still outstanding.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -454,6 +470,86 @@ internal class DefaultSendStrategy(
...
+     * folds UTXO byte fees and Cardano fees back into `gasFee` but deliberately leaves
+     * `GasSettings.Eth` out, while [applyGasSettings] still patches the signed `maxFeePerGasWei`
+     * and `gasLimit` — so an amount sized on `gasFee` alone reserves less than the transaction
+     * spends and can be adjusted to a clean-looking value the chain then rejects for insufficient
+     * funds.
+     */
+    private fun withEvmGasSettings(gasFee: TokenValue): TokenValue {
+        val eth = gasSettings.value as? GasSettings.Eth ?: return gasFee
+        return gasFee.copy(value = eth.maxFeePerGasWei * eth.gasLimit)
+    }
```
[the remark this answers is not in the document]
aminsato: Good catch — confirmed and fixed in 02a9dd941.

`graph.gasSettings.value` is written only by `saveGasSettings` and never reset; nothing clears it on a token or chain change. So a `GasSettings.Eth` from an Ethereum send survives a switch to Bitcoin, and `withEvmGasSettings` would have read its wei-denominated `maxFeePerGasWei × gasLimit` as satoshis — 4.2e15 against a 1e6 balance, leaving nothing spendable and failing *every* BTC send as "insufficient balance". That is a hard regression, not a rounding edge.

I gated on `chain.standard != TokenStandard.EVM` rather than clearing the state on chain change, to match what the other two consumers already do: `applyGasSettings` only applies `GasSettings.Eth` alongside a `BlockChainSpecific.Ethereum`, and `selectGasFeeForFeeEstimation` reaches its `evmGasSettings` branch only after the Cardano and UTXO branches have returned. My helper was the odd one out in keying off the type alone.

Clearing `gasSettings` on token change is still worth doing — it is latent today only because all three consumers now guard — but it belongs in its own PR since it touches the form state lifecycle rather than this submit path.

Regression test: `submit ignores stale EVM gas settings on a non-EVM chain` — a 200 gwei × 21000 leftover on a 0.005 BTC send, asserting the amount goes out untouched and sized against the 2000 sat fee.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -187,29 +189,31 @@ internal class DefaultSendStrategy(
...
-                    // some tokens behind. Re-derive the amount from the CURRENT balance and fee,
-                    // clamping DOWN only so we never send more than the user saw. Standard sends
-                    // only: DeFi flows (staking/unbond/frozen TRX) carry different balance
-                    // semantics and compute their own max.
+                    // Read the fiat mirror once, here, and carry it down to the Transaction rather
+                    // than re-reading the field past the suspension points below. AmountManager
+                    // owns that field too and blanks it whenever a price fetch fails, so a late
+                    // read can put a $0.00 estimate beside a correctly signed amount.
+                    val enteredFiat = fiatAmountFieldState.text.toString().toPlainBigDecimalOrNull()
+                    val spendableGasFee = withEvmGasSettings(chain, gasFee)
```
rkokhatskyi, line 206: spendableGasFee (line 197) only reaches clampToSpendableBalance and the native-token balance recheck. The sibling non-native token branch at DefaultSendStrategy.kt:384 still does nativeTokenValue < gasFee.value, so a raised Advanced Gas Settings fee is not reserved before an ERC-20 send and can fail at broadcast after a full MPC keysign.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -454,6 +470,91 @@ internal class DefaultSendStrategy(
...
+        token: Coin,
+        isMaxAmount: Boolean,
+    ) {
+        val adjustedDecimal = BigDecimal(adjusted).movePointLeft(token.decimal).stripTrailingZeros()
+        // The clamped value is still the true maximum, so re-snapshot it — otherwise the form's
+        // "100%" selection would drop just because the fee moved.
+        if (isMaxAmount) {
+            amountManager.markMax(adjustedDecimal)
+        }
+        tokenAmountFieldState.setTextAndPlaceCursorAtEnd(adjustedDecimal.toPlainString())
```
rkokhatskyi: showAdjustedAmount rewrites the amount fields after the transaction is already persisted, but AmountManager's collectConversion is still running on those same fields in the same scope and treats the write as user input, re-fetching price and blanking fiatAmountFieldState on failure. Resubmitting without retyping can then persist a zero fiatValue next to a correctly signed amount.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -187,29 +189,31 @@ internal class DefaultSendStrategy(
...
-                    // it as "insufficient" — which is why a max send can force the user to leave
-                    // some tokens behind. Re-derive the amount from the CURRENT balance and fee,
-                    // clamping DOWN only so we never send more than the user saw. Standard sends
-                    // only: DeFi flows (staking/unbond/frozen TRX) carry different balance
-                    // semantics and compute their own max.
+                    // Read the fiat mirror once, here, and carry it down to the Transaction rather
+                    // than re-reading the field past the suspension points below. AmountManager
+                    // owns that field too and blanks it whenever a price fetch fails, so a late
+                    // read can put a $0.00 estimate beside a correctly signed amount.
+                    val enteredFiat = fiatAmountFieldState.text.toString().toPlainBigDecimalOrNull()
```
rkokhatskyi: enteredFiat is read right after accountValidator.validate(), but validate() itself suspends: awaitGasFee() can wait up to five seconds when the gas fee hasn't loaded, and resolveName() hits the network for ENS or THORName addresses. AmountManager's independent price coroutine can blank the fiat field during that window, landing a $0.00 estimate in the persisted transaction.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -454,6 +470,91 @@ internal class DefaultSendStrategy(
...
+     * `GasSettings.Eth` out, while [applyGasSettings] still patches the signed `maxFeePerGasWei`
+     * and `gasLimit` — so an amount sized on `gasFee` alone reserves less than the transaction
+     * spends and can be adjusted to a clean-looking value the chain then rejects for insufficient
+     * funds.
+     */
+    private fun withEvmGasSettings(chain: Chain, gasFee: TokenValue): TokenValue {
+        // saveGasSettings never clears the stored settings, so a GasSettings.Eth set on an EVM
+        // chain outlives a switch to, say, Bitcoin. Its wei-denominated product would then be read
+        // as satoshis and swallow the whole balance. The other two consumers (applyGasSettings,
+        // selectGasFeeForFeeEstimation) already gate on the chain or the spec type; do the same.
```
rkokhatskyi: The comment claims selectGasFeeForFeeEstimation already gates on chain or spec type, but it only special-cases Cardano and UTXO chains, for Cosmos, THORChain, Solana, Substrate, Sui, Ripple, and TRC20 chains a stale GasSettings.Eth still overwrites the displayed fee via the unconditional evmGasSettings != null branch.

app/src/test/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategyTest.kt
```diff
@@ -739,6 +749,10 @@ internal class DefaultSendStrategyTest {
             assertNull(lastError, "Expected no error; got $lastError")
             // Clamped down from 0.999 to the current 0.9985 available.
             assertEquals(BigInteger("998500000000000000"), captured.captured.tokenValue.value)
+            assertEquals("0.9985", tokenAmountFieldState.text.toString())
+            // Still a Max — the snapshot moves onto the adjusted value so the form's 100%
+            // selection survives the fee moving under it.
```
rkokhatskyi, line 758: This sibling non-native Max re-clamp test (DefaultSendStrategyTest.kt:662) hits the identical showAdjustedAmount call this test does, with isMaxAmount and isAmountAdjusted both true, but never adds the equivalent verify { amountManager.markMax(...) } this test has here. A regression dropping or misvaluing markMax on that branch would pass the whole suite silently.

app/src/test/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategyTest.kt
```diff
@@ -823,65 +837,810 @@ internal class DefaultSendStrategyTest {
...
+            }
+        }
+
+    /**
+     * The clamp runs before the Cardano/BTC/XRP validators, so an adjusted amount can still be
+     * rejected downstream. When that happens the fields must be left holding what the user actually
+     * typed — stranding the form on a number they never entered is worse than the original error.
+     */
+    @Test
+    fun `submit leaves the amount fields untouched when a post-clamp validator rejects the send`() =
```
rkokhatskyi, line 1193: No test drives validateCardanoUTXORequirements through submit, though the code comment right above the showAdjustedAmount call names a Cardano floor rejection as its motivating case. Cardano appears only in comments in this file, never in a test body; a future edit special-casing that branch would pass the whole suite.

rkokhatskyi (dismissed):

app/src/test/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategyTest.kt
rkokhatskyi: Confirmed — 123400000 at 6 decimals is 123.4, matching the assertion, and it's a real verify{}, not a no-op. This gate is now covered on both the EVM and non-native Max paths.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Confirmed and fixed in af73a6a8a.

You are right that `spendableGasFee` stopped short. It reached `clampToSpendableBalance` and the native recheck, but the non-native branch still read `nativeTokenValue < gasFee.value` — so an ERC-20 send with a raised cap could clear a native-balance check it cannot actually afford, and the user only finds out at broadcast, after paying for a full keysign round. Now gated on `spendableGasFee.value`.

Worth noting this is the same reserve, not a second one: the ERC-20 transfer's gas comes out of the native coin, so the raised `maxFeePerGasWei x gasLimit` is exactly what has to be there.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Right — I moved the read to avoid the later suspensions and then left it behind one. `validate()` suspends twice: `awaitGasFee()` waits up to `GAS_FEE_TIMEOUT_MS` (5s) when the fee has not loaded, and `resolveName()` goes to the network for an ENS/THORName address. Plenty of room for the price coroutine to blank the field.

Fixed in af73a6a8a: both fields are snapshotted at the top of the launch block, before `validate()`, and everything downstream parses those strings rather than re-reading the state. The token amount comes from the same snapshot now too, so the amount and its fiat mirror are guaranteed to be the pair the user was actually looking at when they hit Continue — previously they could be read seconds apart.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: The comment was wrong and so was my reasoning for leaving that function alone — thanks for reading it rather than taking my word. `selectGasFeeForFeeEstimation` returns early only for Cardano and `TokenStandard.UTXO`; Cosmos, THORChain, Solana, Substrate, Sui, Ripple and TRC20 all fall through to the unconditional `evmGasSettings != null` branch and take a wei-denominated `maxFeePerGasWei` as their displayed fee. Only `applyGasSettings` was genuinely safe, and only because it pairs the settings with a `BlockChainSpecific.Ethereum` that no other chain produces.

Fixed in af73a6a8a by resolving the settings once, at the boundary, instead of restating the guard at each site:

```kotlin
internal fun GasSettings?.evmSettingsFor(chain: Chain): GasSettings.Eth? =
    (this as? GasSettings.Eth)?.takeIf { chain.standard == TokenStandard.EVM }
```

Applied at all three call sites — `GasFeeOrchestrator`'s fee estimate, `DefaultSendStrategy`'s `totalGasAndFee`, and `withEvmGasSettings`. `grep 'as? GasSettings.Eth'` over `app/src/main` now returns only the helper itself, so a new consumer cannot reintroduce the bare cast without deliberately writing one. The misleading comment is gone.

`saveGasSettings` still never clears its state, which is what makes all of this reachable; that reset belongs in the form lifecycle and I would rather not fold it into this PR.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
aminsato: Confirmed. Fixed in af73a6a8a, and chasing it turned up a second failure underneath.

First part is as you describe: `AmountManager.markProgrammaticAmount` now pre-arms `lastTokenValueUserInput`/`lastFiatValueUserInput` with the pair about to be written, so the collector recognises the write as its own and never re-runs the price fetch.

That was not sufficient. The test for it failed with the **token** field empty, not the fiat one. `collectConversion` combined two separate `snapshotFlow`s, and `combine` delivers each source's emission on its own — so the collector woke on the new token amount paired with the *old* fiat text, took the `lastFiatValueUserInput != fiatString` branch, ran `handleFiatInput` on stale text and cleared the token field. The clamp erased the very amount it was trying to display. `Snapshot.withMutableSnapshot` around the two writes does not help, because the interleaving is in the flow plumbing rather than the snapshot.

So `collectConversion` now reads both fields in one `snapshotFlow`:

```kotlin
snapshotFlow {
    tokenAmountFieldState.text.toString() to fiatAmountFieldState.text.toString()
}
```

which can only ever report a consistent pair. The writes are still batched into a single apply so the settled pair arrives as one emission. This also closes the same interleaving for ordinary fast typing, where it would have shown up as a field clearing itself mid-entry — pre-existing, and I suspect the cause of some of the flakier amount-field behaviour.

Covered by `markProgrammaticAmount keeps a failed price fetch from blanking the written fiat` (price fetch throws, both fields survive, `getPrice` never called) and `markProgrammaticAmount re-derives isMaxAmount against the max snapshot`.

app/src/test/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategyTest.kt
aminsato: Fair — the comment named a case nothing exercised. Added in af73a6a8a:

`submit surfaces the Cardano minimum-send floor on a clamped amount and keeps the typed fields` — a 1.5 ADA balance against a 1.45 ADA fee, so the clamp lands at 0.05 ADA, under the 1.4 ADA floor. Asserts `minimum_send_amount_is_ada` is raised and that both fields still read the typed `1.5`/`0.6`. That is the exact scenario from your earlier comment, now driven through `submit` rather than described in prose.

It carries the same `assumeTrue` JNI guard as the XRP and BTC tests, since `Chain.Cardano.getDustThreshold` resolves through `coinType.getDustThreshold` in WalletCore. It runs unskipped locally and on CI.

aminsato:
Second round pushed as af73a6a8a — all five verified, all five real, all fixed.

| Comment | Fix |
|---|---|
| `spendableGasFee` missed the non-native branch | ERC-20 native-balance check now gates on `spendableGasFee.value` |
| `enteredFiat` read after `validate()` suspends | both fields snapshotted before `validate()`; nothing re-reads the state |
| clamp write races `collectConversion` | `markProgrammaticAmount` + single-`snapshotFlow` conversion (below) |
| `selectGasFeeForFeeEstimation` gate overclaimed | new `evmSettingsFor(chain)` at every call site; the bare cast is gone |
| no Cardano test through `submit` | 1.5 ADA / 1.45 ADA fee clamps under the floor, rejected, fields untouched |

Two things worth pulling out.

**The `selectGasFeeForFeeEstimation` one — you were right and my comment was wrong.** It returns early only for Cardano and UTXO, so Cosmos, THORChain, Solana, Substrate, Sui, Ripple and TRC20 all fell through to the unconditional `evmGasSettings != null` branch. Only `applyGasSettings` was genuinely safe, and only because it pairs with a `BlockChainSpecific.Ethereum` no other chain produces. Resolved once at the boundary now — `grep "as? GasSettings.Eth"` over `app/src/main` returns only the helper itself.

**The conversion race had a second failure under it.** Pre-arming the caches was not enough: the test failed with the *token* field empty. `combine` delivers each source separately, so the collector woke on the new amount paired with the old fiat text, read that as a cleared field, and cleared its counterpart — the clamp erasing the amount it was trying to show. `Snapshot.withMutableSnapshot` does not help; the interleaving is in the flow plumbing. `collectConversion` now reads both fields in one `snapshotFlow`, so it can only observe a consistent pair. That also closes the same hole for ordinary fast typing, where it would surface as a field clearing itself mid-entry.

Full app unit suite green. Send package 333 tests, 0 failures, 0 skipped — `DefaultSendStrategyTest` 21, `AmountManagerTest` 24.

One thing I deliberately left out: `saveGasSettings` still never clears its state, which is what makes the stale-settings class of bug reachable at all. All consumers guard now, so it is latent — but the reset belongs in the form lifecycle and I would rather not fold it into this PR.

Device verification is still the outstanding gap.

johnnyluo (approved):
LGTM

PR #5512, opened 2026-08-03
fix(Sui): price auto-discovered tokens via CoinGecko contract lookup

### LINKED ISSUE #5508: Sui: auto-discovered tokens never get a fiat price — no Sui contract-price route
## What
Auto-discovered Sui tokens (#5469, `SuiTokenFinder`) get `priceProviderID = ""` (`SuiTokenFinder.kt:127`) and then never price: the contract-address price path is EVM-restricted (`TokenPriceRepository.kt:139-155`) and `CoinGeckoApi.coinGeckoAssetId` has no Sui entry (`CoinGeckoApi.kt:84-98`). Result: discovered tokens show balances at $0.00 indefinitely.

## Fix
Add a Sui contract-price route (CoinGecko sui platform id; iOS additionally has a Cetus-router price fallback worth porting). Sibling: iOS vultisig/vultisig-ios#5028.

rkokhatskyi:
## Summary

Fixes #5508. Auto-discovered Sui tokens (`SuiTokenFinder`) carry an empty `priceProviderID` and only a contract address, so they depend entirely on `TokenPriceRepository`'s generic contract-price route. That route resolves its request through `coinGeckoAssetPlatformId()`, which had no entry for Sui — `CoinGeckoApiImpl` threw internally before any network call, degrading to an empty map, and the LI.FI fallback is EVM-only so it couldn't help either. The token was never priced and showed $0.00 indefinitely.

Adds the missing `"sui"` CoinGecko platform id — the same class of fix already shipped for Mantle in #5406. Verified against the live CoinGecko API that `"sui"` is the correct platform id (it's also the exact key CoinGecko uses in `usd-coin`'s own `platforms` map, matching our curated USDC contract address byte-for-byte).

The rest of the pipeline (`TokenPriceRepositoryImpl`, `CoinGeckoApiImpl`, `Coin.hasMarketDataSource`) is already chain-agnostic and needed no changes — the same code path already works for Solana.

Note: iOS additionally routes Sui through a Cetus on-chain router as its primary (not fallback) price source — a materially different, standalone integration with no existing footprint in this codebase. Out of scope here; this closes the reported gap for any token CoinGecko indexes.

## Before / After

Real device renders (chain page for a Sui vault holding a discovered token with no `priceProviderID`), fetched live through the actual `TokenPriceRepository`/`CoinGeckoApi` production code:

| Before | After |
|--------|-------|
| before | after |

## Testing

- `./gradlew :data:testDebugUnitTest` — full suite green, including two new tests: `CoinGeckoApiContractPriceChainTest` (pins the "sui" platform id resolution) and `TokenPriceRepositoryImplTest` (pins the dispatch of a discovered Sui token into the contract-price route).
- `./gradlew :data:ktfmtCheck` and `:data:lintDebug` — pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Added support for retrieving Sui token prices through CoinGecko.
  - Sui tokens without a configured price-provider ID now fall back to contract-address pricing.
  - Corrected Sui platform mapping to ensure accurate USD price retrieval.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
@@ -457,6 +457,7 @@ fun Chain.coinGeckoAssetPlatformId(): String? =
         Chain.Robinhood -> "robinhood"
         Chain.Mantle -> "mantle"
         Chain.CronosChain -> "cronos"
+        Chain.Sui -> "sui"
         else -> null
     }
 
--- a/data/src/test/kotlin/com/vultisig/wallet/data/api/CoinGeckoApiContractPriceChainTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/api/CoinGeckoApiContractPriceChainTest.kt
@@ -38,6 +38,27 @@ class CoinGeckoApiContractPriceChainTest {
     }
 
     @Test
+    fun `getContractsPrice resolves Sui to the sui CoinGecko platform id`() = runTest {
+        lateinit var requestedUrl: String
+        val engine = MockEngine { request ->
+            requestedUrl = request.url.toString()
+            respond(content = "{}", status = HttpStatusCode.OK)
+        }
+        val api = CoinGeckoApiImpl(HttpClient(engine) { install(ContentNegotiation) { json() } })
+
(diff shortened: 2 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5513, opened 2026-08-04
fix(transaction): collapse hero on scroll, fix bottom-bar nav padding

### LINKED ISSUE #5490: Swap tx overview: scrolling should collapse the pending hero; Track/Done bottom buttons missing navigation-bar padding
## What
Two problems on the swap transaction overview / done screen (user screenshot: pending swap, USDC→BTC):

1. **Scrolling should collapse the "Transaction pending…" hero.** The pending animation + title is a fixed `heightIn(min = 300.dp)` block inside the scrollable column, and the only interaction that collapses it is tapping "Transaction Details". Scrolling just moves it — with the detail rows expanded the hero still occupies the top of the viewport and the actual content is squeezed into the lower half of the screen. Expected: scrolling down collapses/dismisses the pending hero (collapsing-header behavior), so the details get the screen.

2. **Bottom Track/Done buttons have wrong padding.** The bottom-bar row applies only `padding(horizontal = 24.dp, vertical = 12.dp)` and no navigation-bar insets, so with edge-to-edge the buttons sit ~12dp from the physical bottom edge, colliding with the gesture nav area. Horizontal padding is also 24dp while the screen content uses 16dp.

## Where
- Hero block: `app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt:166` (`AnimatedVisibility(isTransactionDetailVisible.not())` — the only collapse trigger) and `:191-192` (`heightIn(min = 300.dp)` hero box inside the `verticalScroll` column at `:163`)
- Bottom bar: `app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SwapTransactionOverviewScreen.kt:246-249` — `Row(... .padding(horizontal = 24.dp, vertical = 12.dp))`, no `navigationBarsPadding()`; other screens (referral flows, bottom sheets) apply insets.
- The bottom-bar fix likely belongs in the shared `TxDoneScaffold` `bottomBar` slot (`TxDoneScaffold.kt:121-138`) so Send/Swap/QBTC done screens all pick it up — `SendTxOverviewScreen.kt` has the same pattern.

## Notes
In the screenshot the pending hero and the expanded detail rows are visible at the same time; per the current `TxDoneScaffold` logic the hero should hide while details are expanded — worth confirming which entry point (initiator done screen vs joining-device flow) produced that state while fixing the scroll behavior.

johnnyluo:
## Summary
- Scrolling the tx-done content (Send/Swap/QBTC done screens share `TxDoneScaffold`) now collapses the pending hero the same way tapping "Transaction Details" does, instead of leaving the fixed 300dp hero pinned at the top while the details get squeezed into the lower half of the screen.
- The shared Track/Done bottom bar now applies `navigationBarsPadding()` so the buttons clear the gesture-nav area under edge-to-edge instead of sitting ~12dp from the physical bottom edge.
- Bottom-bar horizontal padding aligned from 24dp to 16dp to match the rest of the screen content (Send + Swap done screens).

## Issue
Closes #5490

## Test plan
- [ ] Manual: open a pending swap tx overview, scroll down, confirm hero collapses and details get full viewport
- [ ] Manual: confirm Track/Done buttons sit above the gesture-nav bar with proper spacing on Send, Swap, and QBTC claim done screens

## Summary by CodeRabbit

* **UI Improvements**
  * Adjusted horizontal spacing for transaction completion banners, buttons, and action bars.
  * Added navigation-bar spacing to the transaction completion bottom bar.
  * Transaction details now expand automatically after scrolling, while still supporting manual expansion.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
@@ -23,6 +23,7 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.heightIn
+import androidx.compose.foundation.layout.navigationBarsPadding
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.wrapContentSize
@@ -34,8 +35,10 @@ import androidx.compose.material3.SnackbarHost
 import androidx.compose.material3.SnackbarHostState
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalContext
@@ -126,6 +129,7 @@ internal fun TxDoneScaffold(
             }
             AnimatedVisibility(
                 visibleState = bottomBarVisibility,
+                modifier = Modifier.navigationBarsPadding(),
                 enter =
                     fadeIn(animationSpec = tween(durationMillis = 300, delayMillis = 120)) +
                         slideInVertically(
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
```diff
@@ -157,11 +161,22 @@ private fun SuccessTransaction(
...
 
+    val scrollState = rememberScrollState()
+
+    // Scrolling the content collapses the pending hero, same as tapping "Transaction Details" —
+    // otherwise the fixed-height hero keeps hogging the top of the viewport while the user tries
+    // to read the rows below it (#5490).
+    LaunchedEffect(scrollState) {
+        snapshotFlow { scrollState.value }
+            .collect { offset ->
+                if (offset > 0 && !isTransactionDetailVisible) {
```
rkokhatskyi: The scroll-collapse fires on the first pixel with no threshold, and `onTransactionDetailVisibleChange` is never called with `false` anywhere in the codebase. Because `Keysign.kt`'s `AnimatedContent` groups every `KeysignFinished` sub-state under one `contentKey`, this flag survives pending to confirmed/failed, permanently hiding the success or failure hero after one accidental scroll.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TxDoneScaffold.kt
johnnyluo: Fixed in 95ba5f71f — replaced the unbounded `offset > 0` collect with a 24dp threshold and a one-shot `first{}`, so a stray touch/overscroll bounce no longer triggers it. Also fixes the stale-capture issue CodeRabbit flagged in the neighboring thread by keying the effect on `isTransactionDetailVisible` instead of reading it inside the collector.

Left the "stays collapsed across pending → confirmed/failed" behavior as-is for now — that matches the pre-existing tap-to-expand semantics (same flag, not new in this PR), and decoupling scroll-triggered collapse from status transitions felt like a separate product decision rather than a bug fix. Happy to revisit if you'd rather the hero always reappear on a terminal status.

rkokhatskyi (approved):
LGTM

PR #5514, opened 2026-08-04
fix(sui): add a Sui branch to custom token search

### LINKED ISSUE #5507: Sui: custom token add always fails — SearchTokenUseCase has no Sui branch while the tile is offered
## What
Adding a custom Sui token is dead but the UI offers it: `SearchTokenUseCase.kt:41-48` dispatches only EVM/SOL/Kujira/Terra with `else -> null`, while the "+ Custom" tile is hardcoded on (`TokenSelectionScreen.kt:38,120`, rendered `:94-99`) — so `CustomTokenViewModel.kt:56-60` always lands in `showError()` for Sui.

## Fix
Add a Sui branch: metadata by coin type via `suix_getCoinMetadata` (call exists in `SuiApi`), struct-tag validation via the existing `isSameSuiCoinType` normalization. Windows behavior is the reference. Sibling: iOS vultisig/vultisig-ios#5026.

johnnyluo:
## Description

`SearchTokenUseCase` dispatched EVM/SOL/Kujira/Terra but had no Sui case, so the "+ Custom" tile (offered on every chain) always fell through to `CustomTokenViewModel.showError()` on Sui.

Added a `SearchSuiTokenUseCase` that resolves the pasted struct tag via `suix_getCoinMetadata` (existing `SuiApi` call), gated by a coin-type shape check (`address::module::struct`) that also excludes the native SUI type — it's already on the vault by default, never a "custom" token to add.

Fixes #5507

## Which feature is affected?
- [x] New Chain / Chain related feature  -  Please attach tx link here

N/A — this only affects custom-token resolution (metadata read), not a send/swap transaction.

## Parity

- iOS: linked vultisig-ios#5026
- Windows + extension: n/a — issue notes Windows already has this working; used as the reference
- SDK: n/a — reuses existing `SuiHelper` coin-type normalization, no SDK change needed

## Checklist

N/A — data-layer fix, no UI change (the "+ Custom" tile already existed).

- **Agent**: Claude Code
- **Issue**: #5507
- **Confidence**: high
- **Needs human review for**: Sui coin-type shape validation regex/segments (no live Sui RPC hit in this environment; verified via unit tests only)

## Summary by CodeRabbit

* **New Features**
  * Added support for searching Sui tokens using valid Sui coin types.
  * Matching catalog tokens now include available provider pricing.
  * Uncatalogued tokens can be resolved using on-chain metadata with an initial zero price.
  * Invalid, malformed, native SUI, or unavailable token data is handled gracefully.
  * Added support for displaying SVG images throughout the app.

* **Tests**
  * Added coverage for valid searches, rejected coin types, catalog matching, metadata fallback, and unavailable token data.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/SearchSuiTokenUseCase.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/SearchSuiTokenUseCase.kt
@@ -1,6 +1,7 @@
 package com.vultisig.wallet.data.usecases
 
 import com.vultisig.wallet.data.api.chains.SuiApi
+import com.vultisig.wallet.data.api.chains.SuiCoinMetadata
 import com.vultisig.wallet.data.crypto.SuiHelper
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
@@ -8,6 +9,8 @@ import com.vultisig.wallet.data.models.Coins
 import com.vultisig.wallet.data.repositories.TokenPriceRepository
 import java.math.BigDecimal
 import javax.inject.Inject
+import kotlin.coroutines.cancellation.CancellationException
+import timber.log.Timber
 
 internal interface SearchSuiTokenUseCase : suspend (String) -> CoinAndPrice?
 
@@ -24,7 +27,7 @@ constructor(private val suiApi: SuiApi, private val tokenPriceRepository: TokenP
             )
         }
 
-        val metadata = suiApi.getCoinMetadata(coinType) ?: return null
+        val metadata = fetchMetadata(coinType) ?: return null
         val coin =
             Coin(
                 chain = Chain.Sui,
@@ -39,6 +42,22 @@ constructor(private val suiApi: SuiApi, private val tokenPriceRepository: TokenP
             )
         return CoinAndPrice(coin, BigDecimal.ZERO)
     }
+
+    /**
+     * Mirrors [SuiTokenFinder]'s handling of the same call: a bad node response or transient
+     * network failure must surface as "not found" rather than crash the custom-token screen, the
+     * same contract [SearchTerraTokenUseCase] and [SearchKujiraTokenUseCase] give their own RPC
+     * reads.
+     */
+    private suspend fun fetchMetadata(coinType: String): SuiCoinMetadata? =
+        try {
+            suiApi.getCoinMetadata(coinType)
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.d(e, "Sui coin metadata fetch failed for %s", coinType)
+            null
+        }
 
     /**
      * The curated [Coins] catalog entry for [coinType], matched through
--- a/data/src/main/kotlin/com/vultisig/wallet/data/usecases/SearchSuiTokenUseCase.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/usecases/SearchSuiTokenUseCase.kt
@@ -1,17 +1,29 @@
 package com.vultisig.wallet.data.usecases
 
 import com.vultisig.wallet.data.api.chains.SuiApi
+import com.vultisig.wallet.data.crypto.SuiHelper
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.models.Coins
+import com.vultisig.wallet.data.repositories.TokenPriceRepository
 import java.math.BigDecimal
 import javax.inject.Inject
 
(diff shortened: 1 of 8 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/usecases/SearchTokenUseCase.kt
```diff
@@ -80,6 +84,20 @@ constructor(
...
+     * input before spending a network round-trip. The native SUI type is excluded: it's already on
+     * the vault by default, never a "custom" token to add.
+     */
+    private fun String.isSuiCoinTypeShape(): Boolean {
+        val segments = split("::")
+        return segments.size == 3 &&
+            segments.all { it.isNotBlank() } &&
+            segments[0].startsWith("0x") &&
+            !SuiHelper.isNativeSuiCoinType(this)
+    }
```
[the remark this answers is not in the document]
johnnyluo: we do some basic validation here , and defer the real validator to the RPC call

aminsato:
A newly added custom token doesn't always show up in the Select Tokens list right away — you have to leave the screen and reopen it before it appears.

**Steps**
1. Vault → Sui → Select tokens → Custom
2. Paste a Sui coin type (SCB in the recording) → token resolves → **Add SCB Token**
3. Sheet dismisses and you're back on Select tokens

**Actual:** the list is unchanged — SCB is not there.
**Expected:** SCB appears in the list (enabled) as soon as it's added.

Closing Select tokens and reopening it shows SCB correctly, so the token is saved — it's the list on screen that isn't refreshed after the add.

repro

Full recording: https://i.imgur.com/LZTyeDW.mp4

data/src/main/kotlin/com/vultisig/wallet/data/usecases/SearchSuiTokenUseCase.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    override suspend operator fun invoke(coinType: String): CoinAndPrice? {
+        curatedCoin(coinType)?.let { coin ->
+            return CoinAndPrice(
+                coin,
+                tokenPriceRepository.getPriceByPriceProviderId(coin.priceProviderID),
+            )
+        }
+
+        val metadata = suiApi.getCoinMetadata(coinType) ?: return null
```
rkokhatskyi: `suiApi.getCoinMetadata` throws on any RPC error uncaught since `isSuiCoinTypeShape` never checks the address is hex, so a malformed paste like `0xnothex::coin::COIN` crashes the custom-token screen, while `SearchTerraTokenUseCase` and `SearchKujiraTokenUseCase` both catch this same class of failure

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/usecases/SearchSuiTokenUseCase.kt
johnnyluo: Good catch — fixed in 38998191f. `fetchMetadata` now wraps `suiApi.getCoinMetadata` the same way `SuiTokenFinder`, `SearchTerraTokenUseCase`, and `SearchKujiraTokenUseCase` already handle their own RPC reads (rethrow on cancellation, otherwise log and return null), and added a regression test asserting an RPC throw surfaces as `null` rather than propagating.

johnnyluo:
> A newly added custom token doesn't always show up in the Select Tokens list right away — you have to leave the screen and reopen it before it appears.
> 
> **Steps**
> 
> 1. Vault → Sui → Select tokens → Custom
> 2. Paste a Sui coin type (SCB in the recording) → token resolves → **Add SCB Token**
> 3. Sheet dismisses and you're back on Select tokens
> 
> **Actual:** the list is unchanged — SCB is not there. **Expected:** SCB appears in the list (enabled) as soon as it's added.
> 
> Closing Select tokens and reopening it shows SCB correctly, so the token is saved — it's the list on screen that isn't refreshed after the add.
> 
> repro [ repro ](https://camo.githubusercontent.com/40610efd8c81f8d0f9b459244bdec4527cbcc4cbd3f06f4896e551cd4dc2b939/68747470733a2f2f692e696d6775722e636f6d2f6c4c6c3057414e2e676966) [ ](https://camo.githubusercontent.com/40610efd8c81f8d0f9b459244bdec4527cbcc4cbd3f06f4896e551cd4dc2b939/68747470733a2f2f692e696d6775722e636f6d2f6c4c6c3057414e2e676966)
> 
> Full recording: https://i.imgur.com/LZTyeDW.mp4

@aminsato I did it locally , it works fine for me , could be network

rkokhatskyi (approved):
LGTM

PR #5515, opened 2026-08-04
fix(keysign): done screen stuck on "Pending" when status tracking never reports

### LINKED ISSUE #5510: Keysign done screen can stuck on "Pending" forever when the tx-status foreground service fails to bind
## What
Split out of #5488 (mechanism 3 — a real bug, but a different symptom from the hang reported there).

On chains with tx-status support (ThorChain included), `applyBroadcastResult` hands the terminal state entirely to an async poller: `KeysignViewModel.kt:944-945` takes the `startForegroundPolling` branch, so the defensive `KeysignFinished(Broadcasted)` else-branch at `:947-955` never runs.

`KeysignTxStatusPoller.pollStatusService` then starts the foreground status service, emits one `Pending`, and returns `null` if the binding was rejected (`KeysignTxStatusPoller.kt:77-84`). No further status ever arrives, so the done screen sits on "Pending" for the rest of the session even after the tx confirms.

Reachable when the app is backgrounded at the moment polling starts and `startForegroundService` is refused on API 31+ (`TransactionStatusServiceManager.startPolling` returns `false`).

**Impact is cosmetic, not a correctness risk.** `onStatus(Pending)` fires *before* the `if (!bound) return null`, and any `KeysignFinished` maps to the done phase (`Keysign.kt:112`) — so the user does land on the done screen with the tx hash and explorer link, just with a status that never advances. `saveTransactionHistory` also ran before polling started, so the background tx-history poller still confirms the row (`KeysignTxStatusPoller.kt:107-109`).

## Fix direction
Land on `KeysignFinished(Broadcasted)` when `pollStatusService` returns `null` without a terminal result, so the screen never advertises a status the app has stopped tracking. Small enough to fold into the #5488 fix if that PR ends up touching this path.

## Where
- `app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt:944-955, 1038-1055`
- `app/src/main/java/com/vultisig/wallet/data/services/KeysignTxStatusPoller.kt:71-101`
- `app/src/main/java/com/vultisig/wallet/data/services/TransactionStatusServiceManager.kt:57-84`

### LINKED ISSUE #5516: Keysign done screen can still hang on "Pending" when the status service binds but never reports
## What

Split out of #5510 / PR #5515, which only covers the case where the status service **never starts** (`startForegroundService` refused → `startPolling` returns `false` → `TxStatusPollOutcome.NotTracked` → done screen lands on `Broadcasted`).

Two sibling routes produce the identical "Pending forever" symptom and are still open, because in both of them `pollStatusService` never returns at all:

1. **Bind accepted, connection never made.** `TransactionStatusServiceManager.startPolling` returns `true` once `bindService` returns `true`, but `onNullBinding` / `onBindingDied` are not overridden, so if `onServiceConnected` never fires, `serviceReady` stays `false` and `serviceReady.filter { it }.first()` (`KeysignTxStatusPoller.kt:99-101`) suspends for the rest of the session.

2. **Service bound, its own polling never started.** `TransactionStatusService.onStartCommand` catches `SecurityException` from `startForeground` and `stopSelf()`s *before* calling `startPolling` (`TransactionStatusService.kt:68-81`). `BIND_AUTO_CREATE` keeps the instance alive and `onBind` still hands back the binder, so `getStatusFlow()` is non-null but the flow sits at its constructor value `TransactionResult.Pending` forever — `.first { it.isTerminal() }` never completes.

Both are pre-existing (unchanged by #5515) and both are cosmetic in the same way #5510 was: the tx hash and explorer link are on screen, `saveTransactionHistory` already ran, so the history row is still settled later.

## Fix direction

An outcome case cannot express these — the coroutine never gets to return one. A wall-clock deadline is needed around the two waits: `withTimeout` on the `serviceReady` wait and on the first-terminal wait, falling through to the same `TxStatusPollOutcome.NotTracked` the rejected-binding path now uses. The chain's own `maxWaitSeconds` (`TxStatusConfiguration`) is the natural budget — it is already the service's internal poll budget, so the outer deadline just needs headroom over it.

Overriding `onNullBinding` / `onBindingDied` in `TransactionStatusServiceManager` would additionally close route 1 at the source.

## Where

- `app/src/main/java/com/vultisig/wallet/data/services/KeysignTxStatusPoller.kt:99-116`
- `app/src/main/java/com/vultisig/wallet/data/services/TransactionStatusServiceManager.kt:34-47, 57-84`
- `app/src/main/java/com/vultisig/wallet/data/services/TransactionStatusService.kt:68-81`

rkokhatskyi:
## Summary

The keysign done screen could sit on "Transaction pending…" for the rest of the session — the transaction confirmed, the screen never said so.

Every chain has tx-status support, so `applyBroadcastResult` always takes the polling branch and the defensive `KeysignFinished(Broadcasted)` else-branch only ever covers a hashless broadcast. That left `pollStatusService` as the only thing that could finish the screen, and it had four ways to fail without doing so:

- `startForegroundService` is refused — API 31+ with the app backgrounded at the moment the broadcast lands. It emitted one `Pending`, returned, and nothing wrote the screen again. This is the route the ticket describes.
- The binding is accepted but `onServiceConnected` never fires, so the wait on `serviceReady` never resolves.
- The binder is gone by the time the service reports ready.
- The service binds but its own `startForeground` throws, so it stops before starting its poll job and its status flow sits on the initial `Pending` — the wait for a terminal status never resolves.

All four now land on `Broadcasted`, the same state a clean broadcast with nothing to poll already reaches. The middle two needed deadlines rather than a return value, because the coroutine never got far enough to return anything: the bind wait is short (the service runs in this process), and the status wait is the chain's own `maxWaitSeconds` plus headroom. That budget is what makes the deadline safe — the service starts its own clock in `onStartCommand` and emits `TimedOut` when it reaches it, so a healthy poll always settles first and is never torn down.

The rejected-start path also stops publishing the `Pending` it knows nothing can advance, so that case goes straight from the signing spinner to the done screen.

### Why `poll()` needed a new return type

Its nullable `TransactionResult` had a `null` that meant two contradictory things: the status service never reported, and the SwapKit foreground budget lapsed while the swap was still settling. Only the first should finish the screen — mapping `null` to `Broadcasted` would flip an unsettled cross-chain swap to "Transaction successful". They are now separate `TxStatusPollOutcome` cases, handled exhaustively:

- `Terminal` — refresh the actual EVM fee (unchanged).
- `NotTracked` — land on `Broadcasted`.
- `HandedOff` — leave the last observed status; the tx-history poller settles the row.

### Parity

iOS seeds its done-screen poller with `.broadcasted` and writes `.pending` only once an RPC actually answered (`ChainPoller.swift`, `TransactionStatusViewModel.swift`), so it cannot get stuck this way, and its SwapKit give-up (`failureGiveUpInterval = 30 * 60`) deliberately stays `.pending`. This lands Android on both. Windows degrades to `pending` forever when its status query fails — same bug class, still open there.

## Before / After

Send done screen after a successful broadcast the status service could not track. Same build, device and mock data; only the status differs.

| Before — stuck on "Transaction pending…" indefinitely | After — "Transaction successful" |
|--------|-------|
| before | after |

## Testing

`./gradlew :app:ktfmtCheck :data:ktfmtCheck testDebugUnitTest lintDebug assembleDebug` — passing.

11 new tests. `KeysignTxStatusPollerTest` (new file) covers each of the four failure routes plus the settled and SwapKit paths. `KeysignViewModelApplyBroadcastResultTest` gains four, including one that drives both the "nothing to poll" and "the poller never started" paths and asserts they land on the same status, and one that pins the SwapKit handoff to `Pending` so the two outcomes cannot be collapsed. Three were confirmed to fail with the fix reverted; the two deadline tests hang without it.

Overlap note: PR #5513 (https://github.com/vultisig/vultisig-android/pull/5513) touches the transaction done screens; no shared files, low conflict risk.

Closes #5510
Closes #5516

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -247,6 +247,10 @@ class PreviewActivity : ComponentActivity() {
                     "camera_button" -> CameraButton(onClick = {})
                     "banner" -> BannerPreview()
                     "send_tx_done" -> SendTxDonePreview()
+                    "send_tx_done_untracked_before" ->
+                        SendTxDonePreview(status = TransactionStatus.Pending)
+                    "send_tx_done_untracked_after" ->
+                        SendTxDonePreview(status = TransactionStatus.Broadcasted)
                     "deposit_mint_done" -> DepositMintDonePreview()
                     "transaction_history_empty" -> TransactionHistoryEmptyState()
                     "limit_orders_tab" -> LimitOrdersTabPreview()
@@ -1433,7 +1437,7 @@ private fun tonJettonSendState(decoded: Boolean): VerifyTransactionUiModel {
 }
 
 @Composable
-private fun SendTxDonePreview() {
+private fun SendTxDonePreview(status: TransactionStatus = TransactionStatus.Broadcasted) {
     val ethCoin = Coins.Ethereum.ETH
 
     SendTxOverviewScreen(
@@ -1441,14 +1445,14 @@ private fun SendTxDonePreview() {
         showSaveToAddressBook = true,
         transactionHash = "0x1a2b3c...d4e5f6",
         transactionLink = "https://etherscan.io/tx/0x1a2b3c",
-        transactionStatus = TransactionStatus.Broadcasted,
+        transactionStatus = status,
         onComplete = {},
         onBack = {},
         onAddToAddressBook = {},
         tx =
             UiTransactionInfo(
                 type = UiTransactionInfoType.Send,
-                token = ValuedToken(token = ethCoin, value = "1.5", fiatValue = "$3,847.50"),
+                token = ValuedToken(token = ethCoin, value = "0.05", fiatValue = "$128.25"),
                 from = "0xAbCdEf1234567890AbCdEf1234567890AbCdEf12",
                 to = "0x9876543210FeDcBa9876543210FeDcBa98765432",
                 memo = "",
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -247,10 +247,6 @@ class PreviewActivity : ComponentActivity() {
                     "camera_button" -> CameraButton(onClick = {})
                     "banner" -> BannerPreview()
                     "send_tx_done" -> SendTxDonePreview()
-                    "send_tx_done_untracked_before" ->
-                        SendTxDonePreview(status = TransactionStatus.Pending)
-                    "send_tx_done_untracked_after" ->
-                        SendTxDonePreview(status = TransactionStatus.Broadcasted)
                     "deposit_mint_done" -> DepositMintDonePreview()
                     "transaction_history_empty" -> TransactionHistoryEmptyState()
                     "limit_orders_tab" -> LimitOrdersTabPreview()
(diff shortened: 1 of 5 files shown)


johnnyluo (approved):
LGTM

PR #5517, opened 2026-08-04
fix(ton): show GRAM instead of stale TON label in stake/unstake UI

### LINKED ISSUE #5499: TON→GRAM rebrand: leftover "TON" labels in the TON DeFi/staking UI
## What
The Toncoin → GRAM rebrand is incomplete on Android — **chain-level display labels still read "TON"**:

- `Chain.kt:61`: `Ton("Ton", TON, "TON")` — `feeUnit` is documented as a UI label (`Chain.kt:527` comment) and still reads "TON"
- `Chain.ticker()` (`Chain.kt:436`) also returns "TON"

Both render in fee rows / chain-level displays. Keep protocol identifiers untouched (swap-asset names, Banxa codes — display-only rebrand, map at the boundary; SwapKit still knows the asset as "TON").

## Correction (2026-08-03)
The original version of this issue also claimed existing vault coins never migrate. **That was wrong** — the Room migration exists and is thorough: `app/src/main/java/com/vultisig/wallet/data/db/migrations/MIGRATIONS.kt:777-796` rewrites `vaultCoin` id/ticker/logo `TON-Ton → GRAM-Ton` including `disabledCoin` rows (#4981). The initial search looked in `data/src/main/.../db/` instead of `app/src/main/.../db/migrations/`. Scope of this issue is the chain-label leak only.

## Cross-platform status (checked)
- **Windows/extension**: done correctly (chainFeeCoin GRAM + `CoinsMetadataManager` persisted-coin metadata refresh on launch).
- **iOS**: has its own leftovers — vultisig/vultisig-ios#5022 (ChainConfig ticker/feeUnit + defiPositions not migrated).

rkokhatskyi:
## Summary

The Toncoin → GRAM rebrand left three leftover "TON" leaks in the TON DeFi/staking UI. The chain-level "TON" labels (the network name, e.g. the DeFi banner header, `Chain.ticker()`, `feeUnit`) are correct as-is and untouched — the network is still TON, only the native coin's ticker changed to GRAM.

- `TonStakeScreen`/`TonUnstakeScreen`: the amount fields fell back to a hardcoded `"TON"` while `state.ticker` is still empty (before `loadCoin()` resolves the real native coin), so the label briefly showed the stale ticker on every open. Fallback is now `"GRAM"`, matching what it settles to.
- The stake pool list's minimum-stake hint (`ton_staking_min_stake`) had `"TON"` baked into the string with only one substitution arg for the amount. Added a second `%2$s` arg and threaded the real ticker through `TonPoolPickerSheet`/`TonPoolRow`.
- The coin-not-found error (`ton_defi_error_ton_not_in_vault`, shown by `TonStakeViewModel`/`TonUnstakeViewModel`/`TonDeFiPositionsViewModel` when the native coin can't be resolved from the vault) still named "TON"; changed to "GRAM" to match the sibling `solana_staking_error_sol_not_in_vault` pattern (hardcoded correct-for-chain ticker, no param).

All 9 other locale `strings.xml` files got the equivalent structural edit — same placeholder shape, word order/grammar preserved per locale, GRAM kept untranslated (matching how the existing `ton_forward_ton_amount` string already handles GRAM in every locale).

Closes #5499

## Testing

- `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.defi.*"` — passing, including two new regression tests (`TonStakeViewModelTest`, `TonUnstakeViewModelTest`) asserting the resolved ticker is `"GRAM"`, not the stale `"TON"` label.
- `./gradlew :app:ktfmtCheck` — passing.
- `./gradlew :app:lintDebug` — passing.

## Summary by CodeRabbit

- **Updates**
  - TON staking and unstaking screens now display **GRAM** when no ticker is provided.
  - Staking titles, amounts, minimum-stake messages, pool selection, and pool details consistently use the correct symbol.
  - Updated localized messages to reference GRAM and support dynamic asset symbols.

- **Bug Fixes**
  - Corrected outdated TON labels in staking and unstaking interfaces.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/ton/TonStakeScreen.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/ton/TonStakeScreen.kt
@@ -53,9 +53,10 @@ import java.text.DecimalFormat
 @Composable
 internal fun TonStakeScreen(viewModel: TonStakeViewModel = hiltViewModel()) {
     val state by viewModel.state.collectAsState()
+    val ticker = state.ticker.ifEmpty { "GRAM" }
 
     V2Scaffold(
-        title = stringResource(R.string.ton_stake_title, state.ticker.ifEmpty { "TON" }),
+        title = stringResource(R.string.ton_stake_title, ticker),
         onBackClick = viewModel::back,
     ) {
         val amountText = viewModel.amountFieldState.text.toString()
@@ -69,7 +70,7 @@ internal fun TonStakeScreen(viewModel: TonStakeViewModel = hiltViewModel()) {
                 verticalArrangement = Arrangement.spacedBy(12.dp),
             ) {
                 StakingAmountCard(
-                    ticker = state.ticker.ifEmpty { "TON" },
+                    ticker = ticker,
                     amountFieldState = viewModel.amountFieldState,
                     available = state.stakeableBalance,
                     percentageSelected = state.percentageSelected,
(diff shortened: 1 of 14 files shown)


johnnyluo (approved):
LGTM

PR #5518, opened 2026-08-04
fix(defi): drop the hardcoded "APY (approx.) 1%" from bonded headers

### LINKED ISSUE #5494: DeFi bonded header always shows hardcoded "APY (approx.) 1%" — remove the row when no real value is fetched
### What happens

The Bonded tab header (THORChain and Maya) always renders **"APY (approx.) 1%"**. The value is a hardcoded default parameter — no APY is ever fetched or passed for this row — and it directly contradicts the per-node cards below it, which show the real computed APY (currently ~20%).

### Receipts

- `BondedTabContent` → `HeaderDeFiWidget` (no apy argument): `app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/BondedTabComponents.kt:45`
- `HeaderDeFiWidget` renders `ApyApproxWithHint()` unconditionally: `app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt:636` (also the two-action overload at `:787`)
- The hardcoded default: `ApyApproxWithHint(modifier, apy: String = "1%")` at `DeFiComponents.kt:679`, `ApyApprox(..., apy: String = "1%")` at `:652`
- iOS shows no APY row in the bonded header at all (`DefiChainBondedView.swift`), which is why iOS "shows correctly"
- `SolanaStakingPositionsScreen.kt:412` already documents cloning the widget just to omit this "baked-in" row

### Expected

Remove the APY row from the header unless a real fetched value is passed in. Make the row opt-in (`apy: String? = null`, render only when non-null). The Circle USDC screen is the one legitimate user of a static "1%" (`CircleDeFiPositionsScreen.kt:105/116`, matching iOS `CircleYieldProvider.swift` staticApy) — it should pass the value explicitly.

aminsato:
Closes #5494

## Summary

The Bonded tab header on THORChain and Maya always rendered **"APY (approx.) 1%"**. No APY was ever fetched or passed for that row — the value came from a default parameter — and it sat directly above node cards showing the real computed APY (~20%), so the header contradicted the list beneath it. iOS shows no APY row in this header at all (`DefiChainBondedView.swift`), which is why the same screen reads correctly there.

## What changed

- **`DeFiComponents.kt`** — both `HeaderDeFiWidget` overloads take `apy: String? = null` and render `ApyApproxWithHint` (plus its trailing spacer) only when a caller supplies a value. The divider above the row is unaffected either way.
- Removed the `apy: String = "1%"` defaults from `ApyApprox` and `ApyApproxWithHint`. This is what stops the bug recurring — the value can now only originate at a call site, so a new screen cannot silently inherit a fabricated APY.
- **Bonded tab** passes nothing, so the row disappears for both THORChain and Maya (they share `BondedTabContent`).
- **Circle** is the one legitimate static-APY caller: it publishes no APY feed, and iOS carries the same fixed value as `CircleYieldProvider.staticApyText` (`circleStaticApy = "1%"`, identical across all iOS locales). It now passes `CIRCLE_STATIC_APY` explicitly at both call sites.
- **`SolanaTotalStakedCard` deleted.** It was a byte-for-byte copy of the single-action `HeaderDeFiWidget` minus the APY row, and its doc comment said it existed only to omit that "baked-in" row. With the row opt-in the clone is redundant, so the Solana screen uses the shared widget and ~65 lines of duplication go away.

## Test plan

- `:app:compileDebugKotlin` clean; `:app:lintDebug` BUILD SUCCESSFUL; existing DeFi unit tests pass; ktfmt applied.
- **THORChain** → DeFi toggle → THORChain card → Bonded tab (default): the "Total Bonded RUNE" header no longer shows an APY row; per-node APY on the cards below is unchanged.
- **Maya** → DeFi toggle → MayaChain card → Bonded tab: same.
- **Circle** → DeFi toggle → Circle card: header **still** shows "APY (approx.) 1%", in both the open-account and Open Account states.
- **Solana** → DeFi toggle → Solana card: the "Total Staked SOL" card renders exactly as before (no APY row); this is the screen to eyeball, since it swapped from the local clone to the shared widget.

No unit test added: the repo's only Compose test harness is `androidTest` (instrumented, emulator-required), so covering a render-gating change with no logic would mean standing up new test infrastructure.

## Summary by CodeRabbit

* **New Features**
  * APY information is now shown on DeFi headers when available.
  * Circle DeFi positions display a fixed 1% APY.
  * Solana staking now uses the shared DeFi header experience for staked totals and validator actions.
* **Improvements**
  * DeFi and staking screens provide a more consistent layout and interaction experience.
  * APY hints no longer appear when no APY value is available.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
@@ -604,6 +604,7 @@ internal fun HeaderDeFiWidget(
     onClickAction: () -> Unit,
     totalAmount: String,
     totalPrice: String = "",
+    apy: String? = null,
     isLoading: Boolean = false,
     isBalanceVisible: Boolean = true,
     buttonState: VsButtonState = VsButtonState.Enabled,
@@ -666,9 +667,11 @@ internal fun HeaderDeFiWidget(
 
         UiSpacer(16.dp)
 
-        ApyApproxWithHint()
+        if (apy != null) {
+            ApyApproxWithHint(apy = apy)
 
-        UiSpacer(16.dp)
+            UiSpacer(16.dp)
+        }
 
         VsButton(
             label = buttonText,
@@ -681,8 +684,8 @@ internal fun HeaderDeFiWidget(
 
 @Composable
 private fun ApyApprox(
+    apy: String,
     modifier: Modifier = Modifier,
-    apy: String = "1%",
     onMoreInfoClick: (() -> Unit)? = null,
     onMoreInfoIconModifier: Modifier = Modifier,
 ) {
@@ -709,7 +712,7 @@ private fun ApyApprox(
 }
 
 @Composable
-private fun ApyApproxWithHint(modifier: Modifier = Modifier, apy: String = "1%") {
+private fun ApyApproxWithHint(apy: String, modifier: Modifier = Modifier) {
     var showHint by remember { mutableStateOf(false) }
     var iconPosition by remember { mutableStateOf(Offset.Zero) }
     var boxCoords by remember { mutableStateOf<LayoutCoordinates?>(null) }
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5519, opened 2026-08-04
fix(defi): show the RUJI staking APR instead of dropping it

### LINKED ISSUE #5498: RUJI staking APR is fetched but never displayed (apr = null) — and the parser lacks the 1e12 scaling iOS applies
### What happens

The RUJI staking card shows no APR on Android; iOS displays it (e.g. `1.16%`). The API layer fetches the value, but the service drops it.

### Receipts

- `ThorChainApi` fetches `pool.summary.apr.value` (GraphQL query, `data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt:828-834`) and parses it at `:506-507`
- `RujiStakingService` hardcodes `apr = null` for both the bonded and auto-compounding positions — `data/src/main/kotlin/com/vultisig/wallet/data/blockchain/thorchain/RujiStakingService.kt:103`, `:115` (origin/main)
- Display path exists and is unit-correct: `ThorchainDefiPositionsViewModel.kt:579` → `details.apr?.formatPercentage()`
- iOS: `THORChainStakingService.swift:140` → `AccountRootData.swift:77-82` `fractionalRate`

### Two landmines to fix when wiring it up

1. **Missing 1e12 scaling.** The API field is a 12-decimal bigint string (e.g. `"11623890337"` = 1.16%). The Android parse at `ThorChainApi.kt:507` is a bare `.toDouble()` — feeding that to `formatPercentage()` (×100) would render `1162389033.70%`. iOS divides by `pow(10, 12)` (`AccountRootData.swift:80`).
2. **Missing `status` field.** The GraphQL query doesn't request the pool `status`; iOS suppresses the APR whenever the pool isn't `AVAILABLE` (`AccountRootData.swift:78`).

aminsato:
## Summary

The RUJI staking APR was already being fetched — `RujiStakingService` then hardcoded `apr = null` on both positions, so the **Staked RUJI** card rendered without the APY row iOS shows.

Wiring it through needed two corrections on the way, or the row would have been worse than absent:

- **12-decimal scaling.** The Rujira APR is a bigint scalar carrying 12 decimal places, not a percentage: `"11623890337"` is `1.16%`. The parse was a bare `.toDouble()`, which `formatPercentage()` would have multiplied by another 100 and rendered as `1162389033.70%`. `Apr.fractionalRate` now does the `movePointLeft(12)` that iOS does in `AccountRootData.fractionalRate`.
- **Pool status.** The GraphQL query never asked for the pool's `status`. iOS suppresses the rate whenever the pool isn't `AVAILABLE`, so a `SOON` / `NOT_APPLICABLE` pool would have shown its placeholder rate as if it were live. The field is now requested and honoured; an absent `status` still means unqualified, not unavailable.

`RujiStakeBalances.apr` became nullable in the process. A missing rate must not arrive as `0.0` — the card's APY row is null-guarded, so absent hides the row while zero would assert the position earns nothing. The parse stays lenient for the same reason the revenue read above it is: the rate decorates an otherwise complete card and must not take the balances down with it.

The Room cache already round-trips `apr` in both mapper directions, so a cached card keeps its rate.

## Scope note

Only the **bonded** position carries the rate. The auto-compounding (sRUJI) one reinvests its revenue into its own amount and publishes no rate of its own — which is what its existing contract in `RujiStakeBalances` and `RujiStakingService` already says. iOS renders one card holding both amounts and a single APR; Android splits them into two cards, so the rate lands on the primary one.

## Changes

| File | Change |
|---|---|
| `RujiGraphQLResponse.kt` | `Apr` gains `status` + a `fractionalRate` getter (12-decimal scaling, non-`AVAILABLE` suppression) |
| `ThorChainApi.kt` | query requests `status`; parse reads `fractionalRate` instead of `.toDouble()` |
| `RujiStakeBalances.kt` | `apr: Double = 0.0` → `apr: Double? = null` |
| `RujiStakingService.kt` | bonded position passes `rujiStakeInfo.apr` through |

## Test plan

7 new unit tests, all green:

**`ThorChainApiImplTest`**
- scales the APR out of its 12-decimal bigint (`11623890337` → `0.011623890337`)
- publishes no APR while the pool is not `AVAILABLE`
- keeps the balances when the APR is missing or unreadable
- reads the APR of a pool that omits its `status`

**`RujiStakingServiceTest`**
- carries the pool APR onto the bonded position
- leaves the auto-compounding position without an APR of its own
- reports no APR when the pool publishes none

```
./gradlew :data:testDebugUnitTest --tests "*ThorChainApiImplTest*" --tests "*RujiStakingServiceTest*"
```

**Manual:** open a vault holding a bonded RUJI position → tap **THORChain** → **DeFi** toggle → **Staked** tab → the **Staked RUJI** card now shows an **APY** row reading ~`1.16%`, matching iOS.

Closes #5498

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RUJI staking APR accuracy using the network-provided rate and correct decimal scaling.
  * APR now appears as unavailable when missing, invalid, or not available instead of displaying 0%.
  * Bonded RUJI positions display available APR correctly.
  * Auto-compounded sRUJI positions continue to omit APR where it does not apply.

* **Tests**
  * Added coverage for available, unavailable, missing, invalid, and status-less APR responses.


data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/RujiGraphQLResponse.kt
```diff
@@ -69,5 +69,34 @@ data class StakingV2(
...
+    /**
+     * The fractional rate (`0.0116` for 1.16%), or null when the pool publishes no usable one.
+     *
+     * Null rather than zero on every failure — an absent rate hides the APR row, where a zero would
+     * assert the position earns nothing.
+     */
+    val fractionalRate: Double?
+        get() {
+            if (status != null && status != APR_STATUS_AVAILABLE) return null
+            val raw = value?.toBigDecimalOrNull() ?: return null
```
rkokhatskyi: `fractionalRate` parses `value` with `toBigDecimalOrNull` here, accepting decimal-point and exponent strings even though the field is an integer-only bigint scalar; sibling amount fields use `toBigIntegerOrNull` instead. A value like "1162.389" parses cleanly and silently scales to a wrong finite percentage rather than null

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/RujiGraphQLResponse.kt
```diff
@@ -69,5 +69,34 @@ data class StakingV2(
...
+            return raw.movePointLeft(APR_DECIMALS).toDouble()
```
rkokhatskyi: `movePointLeft(APR_DECIMALS)` is unguarded here: a `value` string with an extreme exponent like "1E2147483643" parses as a valid BigDecimal but throws an uncaught ArithmeticException, propagating through RujiStakingService's catch-rethrow and dropping the whole RUJI card instead of just hiding the APR row

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/RujiGraphQLResponse.kt
```diff
@@ -69,5 +69,34 @@ data class StakingV2(
...
+) {
+    /**
+     * The fractional rate (`0.0116` for 1.16%), or null when the pool publishes no usable one.
+     *
+     * Null rather than zero on every failure — an absent rate hides the APR row, where a zero would
+     * assert the position earns nothing.
+     */
+    val fractionalRate: Double?
+        get() {
+            if (status != null && status != APR_STATUS_AVAILABLE) return null
```
rkokhatskyi: This treats a null status as available, matching iOS but diverging from Windows's `rujiStakeService.ts`, deliberately hardened in PR #4361 to treat null status as unavailable. Neither documents a real response causing this; whether a not-yet-live pool can return null status with a stale value needs runtime confirmation

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/RujiGraphQLResponse.kt
aminsato: Confirmed and fixed in a4a174661. You are right on both the schema and the precedent — the Rujira `Bigint` scalar documents itself as *"a signed large integer… always returned as a string to preserve precision"*, and every sibling amount in `ThorChainApi` (`liquidShares` :500, `pendingRevenue` :504, :534, :554) reads it with `toBigIntegerOrNull`. The APR was the only one parsed as a decimal.

Reproduced your example: `"1162.389"` parses cleanly and scales to `1.162389E-9`, i.e. a rendered `0.0000001%`. Worse than the bug this PR set out to fix, because that one was obviously wrong on screen and this one isn't.

Now `toBigIntegerOrNull`, which rejects the decimal point and the exponent form outright. That also closes your next comment at the source — see there.

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/RujiGraphQLResponse.kt
aminsato: Confirmed — it throws exactly as you described, and I ran your string to be sure rather than reason about `checkScale` from memory:

```
"1E2147483643"  -> ArithmeticException: BigInteger would overflow supported range
"1E-2147483640" -> ArithmeticException: Underflow
"1E400"         -> Infinity (no throw)
```

The negative exponent is the second door into the same failure. And the consequence is what you said: `RujiStakingService` catch-rethrows at :122-125, so one bad APR string took the whole Staked RUJI card down instead of hiding one row — the exact inversion of this PR's intent, which is that the rate decorates the card and must never take the balances with it.

Fixed in a4a174661 by removing the unbounded scale rather than guarding the call: a `BigInteger` always carries scale 0, so `movePointLeft(12)` lands on 12 and can neither overflow nor underflow. The `toBigIntegerOrNull` switch from your other comment is what makes that hold — it is one fix, not two.

The `1E400` row is why CodeRabbit's finite check stays on top of this: a plain 400-digit integer is a legal `Bigint` and still reaches `Infinity`, which `formatPercentage` would render as `∞%`. Covered by `keeps the card when the APR overflows a double`.

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/RujiGraphQLResponse.kt
aminsato: You asked for runtime confirmation, so I introspected the endpoint we actually call (`api.vultisig.com/ruji/api/graphql`):

```json
{"__type":{"name":"Apr","fields":[
  {"name":"status","type":{"kind":"NON_NULL","ofType":{"kind":"ENUM","name":"AprStatus"}}},
  {"name":"value","type":{"kind":"SCALAR","name":"Bigint"}}]}}
```

`status` is non-null, `AprStatus` is `AVAILABLE | NOT_APPLICABLE | SOON`, and we select the field explicitly — so a conforming server cannot produce the null-status response, and no not-yet-live pool can reach that path with a stale value. The lenient branch was unreachable rather than wrong, which is also why neither side documents a response causing it.

That makes Windows's reading free to adopt, so I have adopted it in a4a174661: `status != AVAILABLE` suppresses the rate, absent included. A response that omits a field its own schema marks required is not attesting the rate it carries, and of the two failure modes — hiding a real APR versus rendering a placeholder as live yield — the second is the one that misinforms someone about what their position earns. That is the same reasoning the file already applies to null-versus-zero.

`value` is the genuinely nullable one, so an `AVAILABLE` pool publishing no rate yet is the case worth covering; CodeRabbit asked for that test in a sibling thread and it is in this commit.

Flipped `reads the APR of a pool that omits its status` to `publishes no APR when the pool omits its status`. iOS's `AccountRootData.fractionalRate` still has the lenient guard — same dead branch, no behavioural difference against this schema, so I have left it rather than open a drive-by iOS PR. Happy to file it if you would rather all three read alike.

data/src/main/kotlin/com/vultisig/wallet/data/api/models/thorchain/RujiGraphQLResponse.kt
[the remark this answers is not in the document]
aminsato: Confirmed and taken in a4a174661, with the guard kept even though the parse now rejects exponent strings: a plain 400-digit integer is a legal `Bigint`, scales to `Infinity`, and would reach `formatPercentage` as `∞%`. Verified `"1E400"` → `Infinity` before adding it.

Covered by `getRujiStakeBalance keeps the card when the APR overflows a double`, which also asserts the balances survive it.

data/src/test/kotlin/com/vultisig/wallet/data/api/ThorChainApiImplTest.kt
```diff
@@ -560,6 +569,78 @@ class ThorChainApiImplTest {
...
+                            apr = aprJson(value = "not-a-rate"),
+                        ),
+                    balancesBody = """{"balances":[]}""",
+                )
+
+            val result = api.getRujiStakeBalance("thor1abc")
+
+            assertNull(result.apr)
+            assertEquals(BigInteger("7875733"), result.stakeAmount)
+        }
```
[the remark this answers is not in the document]
aminsato: Fair — and it is the reachable half of the pair, which the existing test did not cover. Schema introspection confirms `value: Bigint` is nullable while `status: AprStatus!` is not, so an `AVAILABLE` pool that has not published a rate yet is a real response, where the null-status case is not.

Added `getRujiStakeBalance keeps the balances when the APR carries no value` using `aprJson(value = null)` — asserts `result.apr` is null and `stakeAmount` is still `BigInteger("7875733")`, as you specified.

Two more from the same round: `publishes no APR for a rate that is not an integer` (`"1162.389"`, which used to scale to a finite wrong 1.16e-9) and `keeps the card when the APR overflows a double`.

johnnyluo (approved):
LGTM

data/src/test/kotlin/com/vultisig/wallet/data/api/ThorChainApiImplTest.kt
```diff
@@ -560,6 +569,128 @@ class ThorChainApiImplTest {
...
+            newRujiApi(
+                stakeBody = rujiStakeBody(bondedAmount = "7875733", apr = aprJson(value = null)),
+                balancesBody = """{"balances":[]}""",
+            )
+
+        val result = api.getRujiStakeBalance("thor1abc")
+
+        assertNull(result.apr)
+        assertEquals(BigInteger("7875733"), result.stakeAmount)
+    }
```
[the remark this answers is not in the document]
johnnyluo: Skipping — the whole file uses `runBlocking` uniformly (every existing test, ~20+ of them, no `runTest` import anywhere in it). Switching just these 4 new tests to `runTest` would make the file inconsistent for no functional gain: none of them use delays or need virtual-time control, so the two runners behave identically here. A file-wide `runBlocking` → `runTest` pass would be a reasonable follow-up, but it's out of scope for this PR.

PR #5520, opened 2026-08-04
fix(send): stop UTXO max sends failing with insufficient-UTXOs error

### LINKED ISSUE #5504: Max send plans against a lost `sendMaxAmount` flag and a stale failed plan — "Not enough available UTXOs" with a fully funded wallet
Max sends on UTXO chains fail with `Not enough available UTXOs to cover the selected amount` even when the wallet holds a single healthy UTXO. Three defects in the send flow compound into this; each is independently reproducible.

### 1. Max amount is computed with the sat/vB rate, not a fee

`AmountFractionManager.chooseMaxTokenAmount()` → `GetAvailableTokenBalanceUseCase`:

```kotlin
// GetAvailableTokenBalanceUseCase.kt:41-43
value = tokenValue.value.minus(gasCost).minus(reserve).coerceAtLeast(BigInteger.ZERO)
```

For UTXO chains `gasCost` is `gasFee.value` — the byte fee from `UtxoFeeService.kt:23-38` (Blockchair suggested sat/vB × 2.5). Max therefore fills in `balance − ~10 sats` while the real fee is 100–200× larger. This lands the amount in the one band WalletCore's exact selector can never satisfy (`inputs < amount + fee`, and the amount is just below the input sum so the planner's own force-max fallback doesn't trigger). The fee *displayed* is the correct `planFee` (`GasFeeSelection.kt:25-32`) — sizing and display use different quantities.

### 2. The `sendMaxAmount` flag is lost to a race and then clobbered

`GasFeeOrchestrator.collectMaxAmount()` patches the flag onto `specific` as a side channel:

```kotlin
// GasFeeOrchestrator.kt:302-306
val spec = specific.value?.blockChainSpecific as? BlockChainSpecific.UTXO
    ?: return@collect            // no-op while specific is still null
```

- **First Max tap of a session:** `specific.value` is still null (no gas fee had been emitted yet), the patch no-ops, and `isMaxAmountFlow` is a deduping StateFlow so it never re-fires. Deterministic.
- **Any later re-emit:** `collectSpecific` calls `getSpecific(..., isMaxAmountEnabled = false)` — hardcoded (`GasFeeOrchestrator.kt:441`) — and overwrites the patch whenever the byte fee moves.

`GasFeeOrchestratorTest.kt:238-256` pre-seeds `specific.value`, so the test passes over both holes.

### 3. Submit reuses the cached failed plan

```kotlin
// DefaultSendStrategy.kt:497-506
planBtc.value
    ?: bitcoinPlanService.getPlan(vaultId, selectedToken, dstAddress, tokenAmountInt, specific, memo)
        .also { plan -> planBtc.value = plan; planFee.value = plan.fee }
```

The submit path builds a *corrected* specific (`isMaxAmountEnabled = isMaxAmount`, `DefaultSendStrategy.kt:247`) but the elvis short-circuits on the failed plan the orchestrator already cached. `planBtc` is never cleared, and `collectPlanFee` swallows its own exceptions (`GasFeeOrchestrator.kt:284-288`), so the error is sticky at Continue.

### 4. Every plan error becomes the same string

```kotlin
// ChainValidationService.kt:152-155
if (plan?.error != SigningError.OK) {
    throw InvalidTransactionDataException(R.string.insufficient_utxos_error.asUiText())
}
```

`Error_not_enough_utxos`, `Error_missing_input_utxos`, `Error_dust_amount_requested`, `Error_zero_amount_requested` all collapse into "Not enough available UTXOs", which is why this symptom has been re-reported since #668 without converging on the design.

### Repro

1. Fresh send screen, BTC, tap Max (first tap of the session), Continue → error, single healthy UTXO.
2. Tap Max, wait for a byte-fee refresh, Continue → same.

### Fix direction

- Compute Max from a real plan: run the WalletCore plan with `useMaxAmount = true` and use `plan.amount` / `plan.fee` for the field and the signed tx. Kill the `balance − byteFee` math.
- Set `isMaxAmountEnabled` from actual UI state inside `collectSpecific`; delete the side-channel patch.
- Invalidate `planBtc` whenever specific/amount changes; re-plan at submit.
- Map each `SigningError` to a distinct message.

Verified against WalletCore 4.7 wasm: single 1,000,000-sat UTXO, `amount = total − 550, useMaxAmount = false` → `Error_not_enough_utxos`; same input with the flag true plans fine (`fee 550, amount 999450`).

johnnyluo:
## Description

Max sends on UTXO chains (BTC, LTC, DOGE, DASH, ZEC, BCH) could fail with "Not enough available UTXOs to cover the selected amount" even with a single, fully funded UTXO. Three compounding defects in the send flow caused this:

1. **Wrong quantity subtracted for Max.** `AmountFractionManager` computed Max as `balance − byteFee`, where `byteFee` is a sat/vB *rate* (from `UtxoFeeService`), not the transaction's total fee — so Max filled in an amount only a few sats below the real balance while the actual fee is 100–200x larger. Max now runs a real WalletCore plan with `sendMaxAmount=true` and uses `plan.amount`/`plan.fee`, falling back to the old approximation only if planning isn't possible yet (e.g. `specific` hasn't loaded).
2. **The max-amount flag was lost/clobbered.** `GasFeeOrchestrator` patched `sendMaxAmount` onto the UTXO specific via a side channel that no-op'd on the first Max tap of a session (`specific` was still null) and got overwritten by every later specific refresh, which hardcoded `isMaxAmountEnabled = false`. `isMaxAmountEnabled` is now sourced live from UI state when the specific is built, and the side-channel patch is deleted.
3. **Submit reused a stale cached plan.** `DefaultSendStrategy` only re-planned the BTC transaction if `planBtc.value` was still null, so a background-collected plan computed against a wrong flag/amount was reused as-is at submit time. Submit now always re-plans against the corrected `specific`.

Fixes #5504

on chain tx link: https://blockchair.com/dogecoin/transaction/88f391b6e9d50c9be267b0f2ba638464203e4fc90f59d957665fb47698275377

https://blockchair.com/zcash/transaction/b3bdde36b1c80ef2358d82966abf933ae85b4db84f51ff51c7a88dc66c02590a

## Which feature is affected?

## Parity

Touches Send's fee/amount computation for UTXO chains. I don't have access to the iOS/Windows/SDK repos from this environment to check whether they share the same Max-amount computation path — flagging for maintainer triage rather than guessing:

- iOS: not verified — please check if `chooseMaxTokenAmount` there has the same balance-minus-byteFee issue
- Windows + extension: not verified
- SDK: not verified

## Checklist

N/A — logic-only fix in the send amount/fee pipeline, no UI changes.

- **Agent**: Claude Code
- **Issue**: #5504
- **Confidence**: medium — the fix matches the issue's own "fix direction" and WalletCore verification, and is covered by new unit tests (all passing), but I could not exercise the real send flow against a live wallet/testnet.
- **Needs human review for**: the Max-amount real-plan path in `AmountFractionManager.calculateUtxoMaxAmount` (new WalletCore plan call on the amount-selection hot path) and the "always re-plan at submit" change in `DefaultSendStrategy` (extra plan call on every UTXO submit)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved maximum-amount calculations for Bitcoin and other UTXO-based transactions.
  * Uses accurate transaction plans when available, with a reliable fee-based fallback if planning fails.
  * Applies current maximum-amount settings during fee estimation and transaction preparation.
  * Recalculates Bitcoin transaction plans at submission time to prevent stale plan or fee data.
  * Ensures maximum-amount sends use the planner’s actual spendable amount.

* **Tests**
  * Added regression coverage for maximum-amount selection, fee estimation, and transaction submission.


app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -494,20 +494,22 @@ internal class DefaultSendStrategy(
...
+                specific = specific,
+                memo = memo,
+            )
+            .also { plan ->
+                planBtc.value = plan
+                planFee.value = plan.fee
+            }
 
         return chainValidationService.selectUtxosIfNeeded(
             chain = chain,
```
aminsato: The fresh plan is published to the shared `planBtc` flow and then read back out of it (`plan = planBtc.value`) to feed `selectUtxosIfNeeded`, which reopens the stale-plan window this change is closing.

`GasFeeOrchestrator.collectPlanFee` writes the same `planBtc`/`planFee` flows from a background collector that re-fires on `specific`, `recalculateGasFee`, and the address/amount/memo text flows. Its `getPlan` call is suspending, so one that was already in flight when submit started can land between the write on line 510 and the read on line 517 — UTXO selection then runs against exactly the stale-flag/stale-amount plan the comment above says we no longer trust.

The two plans can also differ without any race: submit clamps `tokenAmountInt` with `coerceAtMost(available)` before planning, so the background plan is computed for a different amount.

Use the local value instead:

```kotlin
val plan =
    bitcoinPlanService.getPlan(
        vaultId = vaultId,
        selectedToken = selectedToken,
        dstAddress = dstAddress,
        tokenAmountInt = tokenAmountInt,
        specific = specific,
        memo = memo,
    )
planBtc.value = plan
planFee.value = plan.fee

return chainValidationService.selectUtxosIfNeeded(
    chain = chain,
    specific = specific,
    plan = plan,
)
```

app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt
```diff
@@ -240,6 +259,43 @@ internal class AmountFractionManager(
...
+                    bitcoinPlanService.getPlan(
+                        vaultId = vault.id,
+                        selectedToken = token,
+                        dstAddress = addressFieldState.text.asAddressInput(),
+                        tokenAmountInt = balance,
+                        specific = maxSpecific,
+                        memo = memoFieldState.text.toString().takeIf { it.isNotBlank() },
+                    )
+                }
+            if (plan.error != SigningError.OK) return null
```
rkokhatskyi, line 288: `calculateUtxoMaxAmount` falls back to the old balance minus byteFee rate math whenever the plan fails, e.g. a blank destination when Max is tapped, so `Transaction.tokenValue` at `DefaultSendStrategy.kt:418` stages that wrong amount. sendMaxAmount=true still reaches WalletCore's planner, which silently signs a different real sweep amount

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -494,20 +494,22 @@ internal class DefaultSendStrategy(
...
+                vaultId = vaultId,
+                selectedToken = selectedToken,
+                dstAddress = dstAddress,
+                tokenAmountInt = tokenAmountInt,
+                specific = specific,
+                memo = memo,
+            )
+            .also { plan ->
+                planBtc.value = plan
+                planFee.value = plan.fee
```
rkokhatskyi: planFee.value set here gets read again for the Verify screen's fee at `DefaultSendStrategy.kt:400`, but the unconditional `withContext(Dispatchers.IO)` hop at lines 321 to 337 sits between them, letting `GasFeeOrchestrator.collectPlanFee` overwrite it with a different plan first. Signed bytes stay correct, but the displayed fee can mismatch what's charged

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
rkokhatskyi: I don't think this is reachable, so no change needed here. `viewModelScope` is `Dispatchers.Main.immediate`, single threaded and cooperative, and there's no suspend call between the `.also` write and either read since `selectUtxosIfNeeded` and `validateBtcLikeAmount` are both plain non-suspend functions. So `collectPlanFee` can only write before or after that window, never inside it, and `applyBitcoinPlan`'s own write always lands last right before the read runs. Also checked `getAvailableTokenBalance`, the one call between the write and the later `validateBtcLikeAmount` read: it's declared suspend but the bound implementation never actually yields, so it can't create a gap either.

johnnyluo: Fixed in 0603017: `applyBitcoinPlan` now returns the plan alongside the updated `specific` and the caller uses that local value for `selectUtxosIfNeeded` directly, instead of reading it back out of `planBtc.value`.

@rkokhatskyi's reachability analysis for this specific write→read span looks right to me (no suspension point between them), but using the local value removes the need to reason about that invariant at all — it's free and closes the discussion either way.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
johnnyluo: Fixed in 0603017. This one is a real, reachable race — unlike the `selectUtxosIfNeeded` read a few lines below, there's a genuine `withContext(Dispatchers.IO)` (the Ripple destination validation) sitting between `applyBitcoinPlan` writing `planFee.value` and this call reading it back, and `GasFeeOrchestrator.collectPlanFee`'s background collector writes the same flow.

`applyBitcoinPlan` now returns the plan alongside the specific, and submit uses `btcPlan?.fee ?: planFee.value` directly instead of re-reading the flow after the IO hop.

app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt
johnnyluo: Good catch — fixed the downstream symptom in 0603017 rather than `calculateUtxoMaxAmount`'s fallback itself (that fallback is intentional: it's the loading-state estimate shown while the plan can't be computed yet, e.g. a still-blank destination).

By submit time the destination is guaranteed non-blank, so `applyBitcoinPlan`'s fresh WalletCore plan always succeeds for a real send. Submit now stages `Transaction.tokenValue` from that plan's own `amount` for a Max UTXO send instead of the approximate `tokenAmountInt`, so what gets persisted/shown on Verify always matches what the plan (and therefore the signer) actually sweeps — added a regression test (`submit stages the plan's real sweep amount for a Max BTC send...`) pinning this with a plan amount that deliberately differs from the estimate.

rkokhatskyi:
@johnnyluo the PR has conflicts now

app/src/main/java/com/vultisig/wallet/ui/models/send/AmountFractionManager.kt
rkokhatskyi: Fixed in 0603017. `submit()` now stages `Transaction.tokenValue` from the plan's own amount (`stagedAmountInt`, using `btcPlan.amount` whenever `isMaxAmount` and `btcPlan.error == SigningError.OK`) instead of the approximate estimate, so the fallback no longer leaks a wrong displayed amount. New test asserts the plan's real sweep amount lands in the transaction.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
rkokhatskyi: Fixed in 0603017. `applyBitcoinPlan` now returns the plan directly and `submit()` uses `btcPlan?.fee ?: planFee.value` for the estimate instead of re-reading the shared flow, closing the window this comment flagged.

rkokhatskyi (approved):
LGTM

Issue #4999, opened 2026-06-22
refactor: move loadData init-time flow wiring into DepositDataLoader (Phase 10c of #4631)


aminsato:
# [Refactor] Phase 10c — move `loadData` init-time flow wiring into `DepositDataLoader`

Sub-issue of #4631 (decompose `DepositFormViewModel`).

## Context

Phase 10b (`DepositOptionCoordinator`, #4991) has landed via #4992 — option-selection orchestration is out of the VM, which is now **1,109 lines**. The remaining bulk of `loadData` is the **init-time flow wiring**: the option-list build, the initial state seed, and three sequence-sensitive flow collectors started at screen entry.

This is the **biggest and highest-risk** Phase 10 step. The collectors in `loadData` are exactly the ones the umbrella's *Key risk* note is about — reordering `combine`/`onEach` chains here is what caused the stale-emission regression fixed in #4601 on the Send side. They are deliberately left for last. This PR moves them into `DepositDataLoader` (which already owns `loadAddress` / `setMetadataInfo`) behind a single wiring entry point, while keeping `viewModelScope` ownership and collector order bit-identical.

## Scope — this PR only

Move the `loadData` init-time wiring (current line numbers on `main`) into `DepositDataLoader`:

- the `depositOptions` chain-`when` build + the initial `_state` seed (`depositMessage` / `depositOptions` / `depositOption` / `depositChain` / `selectedToken`) — L382–L427
- the `coinList` derivation + its `_state` seed — L429–L440
- the `dataLoader.loadAddress(vaultId, chain)` call — already on the loader; keep the call site ordering
- the `address.filterNotNull().onEach { … }.launchIn(viewModelScope)` collector (native-token selection + ADD_LP gas-fee display) — L442–L463
- **combine collector #1** — `selectedCoin × address × depositOption × selectedToken` → `updateTokenAmount` + the `SecuredAsset → collectSecuredAssetAddresses` / `setMetadataInfo` `collect` — L465–L515
- **combine collector #2** — `selectedCoin × depositOption` → `dstChainList` build + `selectDstChain(first)` for IBC/Switch — L517–L552

[...]

PR #5521, opened 2026-08-05
test(chains): golden vectors for Cosmos matrix, Dash/Zcash, Bittensor, QBTC, limit orders

johnnyluo:
## Summary
- Adds `cosmos-chain-matrix.json`: sends for Osmosis/Dydx/Noble/Akash + one genuine IBC transfer (`transaction_type: 3`, actually routed through `CosmosHelper`'s IBC branch, not a plain send that merely looks like one)
- Adds `bittensor.json` and `qbtc.json`: the two hand-rolled signing paths in this repo (no WalletCore `TransactionCompiler` involved) — the class of divergence #5421 flags as highest-risk
- Extends `utxo.json` with Dash + Zcash sends
- Extends `thorchainswap.json` with a THORChain limit-order "Place" case — confirmed via `BuildLimitSwapTransactionUseCase` that placing an order is an ordinary THORChain swap deposit, differing only in the `=<` memo vs a market swap's `=`
- Fixes two pre-existing bugs found while building this out:
  - `ChainHelperModels.kt`'s `UtxoSpecific` had no `zcash_branch_id` field at all, so no Zcash fixture could ever have worked
  - `ChainHelperModelExtension.kt`'s Cosmos-family mapping hardcoded `transactionType = UNSPECIFIED` regardless of the fixture's `transaction_type` field, so no Cosmos-family IBC fixture — including the pre-existing `kujira.json` one — ever actually exercised the IBC code path. Fixed by reusing the existing `getTransactionType()` helper (already used correctly for THORChain).

Item 2 (SEI) is deliberately not covered here — PR #4999 found Swift and the TS core produce genuinely different transaction envelopes for SEI, so no vector can land until that's resolved upstream.

Every new `expected_image_hash` was captured by actually running this repo's Kotlin signer (`connectedDebugAndroidTest` on a Pixel 9 Pro emulator) — none were hand-authored. Per #5421's rule, a hash needs agreement from at least two of the three real signers (Kotlin/Swift/TS) before it's ground truth; only the Kotlin side has run so far, so **Swift/TS corroboration is still pending** before this should be treated as fully verified.

## Issue
Part of #5421 (items 4, 6, 7)

## Test plan
- [x] `./gradlew :app:compileDebugAndroidTestKotlin` — compiles clean
- [x] `./gradlew lint` — passes
- [x] `./gradlew :app:connectedDebugAndroidTest --tests "com.vultisig.wallet.data.blockchain.ChainHelpersTest"` on a Pixel 9 Pro emulator — all 35 tests pass, including `fixtureCorpusMatchesDeclaredFileSet` (corpus now 87 cases / 25 files)
- [ ] Cross-signer corroboration (Swift and/or TS core) against the same payloads

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction handling for Cosmos, THORChain, and Zcash transfers.
  * Added support for correctly processing Zcash transaction branch information.

* **Tests**
  * Expanded signing coverage for Bittensor, QBTC, Cosmos chains, THORChain swaps, Dash, and Zcash.
  * Added validation for Cosmos IBC transfers and additional transaction scenarios.


app/src/androidTest/assets/utxo.json
```diff
@@ -207,5 +207,74 @@
...
+        "hex_public_key": "023e4b76861289ad4528b33c2fd21b3a5160cd37b3294234914e21efb6ed4a452b",
+        "logo": "zcash.png"
+      },
+      "to_address": "t1XFwpecw7qoLEL5NfFb5hufJYUipVx91hP",
+      "to_amount": "49900000",
+      "BlockchainSpecific": {
+        "UtxoSpecific": {
+          "byte_fee": "100",
+          "send_max_amount": false,
+          "zcash_branch_id": "c2d6d0b4"
```
aminsato, line 264: `c2d6d0b4` is the **big-endian** NU5 branch id — the raw `consensus.nextblock` value as the RPC returns it. A real payload never carries that form: `ZcashApiImpl.fetchBranchId()` pipes the RPC value through `reverseHexBytes()` before it ever reaches `BlockChainSpecific.UTXO` (`data/.../api/ZcashApi.kt:77`, and the KDoc on `BlockChainSpecfic.kt:10` states the field is little-endian hex). `UtxoHelper` then feeds it to `setBranchId(ByteString.fromHex(branchId))` unmodified (`UtxoHelper.kt:322`), so for NU5 the value on the wire is `b4d0d6c2`.

That makes this vector pin a ZIP-243 sighash computed over a byte-swapped branch id — the one input the app can't produce, so the case can't regress-catch anything in the branch-id path, and Swift/TS corroboration would have to hardcode the same wrong-endian value to agree with it.

Swap to `"zcash_branch_id": "b4d0d6c2"` and recapture `expected_image_hash` so the vector locks in the digest the signer actually computes.

rkokhatskyi (approved):
LGTM

app/src/androidTest/assets/utxo.json
johnnyluo: it is test file , so it should be ok

PR #5522, opened 2026-08-05
fix(defi): price RUJI off its real CoinGecko id instead of $0.00

aminsato:
## Summary

A staked RUJI position rendered as `$0.00` — `0.18930143 RUJI` is roughly `$0.035`. RUJI turned out to have **no working price source anywhere in the app**.

`Coins.ThorChain.RUJI` and `sRUJI` carried `priceProviderID = "ruji"`, which CoinGecko does not resolve — it indexes the token as `rujira`:

```
simple/price?ids=ruji    → {}
simple/price?ids=rujira  → {"rujira":{"usd":0.186663}}
```

So the bulk `simple/price` refresh saved nothing and no `RUJI-ThorChain` row was ever written to the price cache. None of the other routes cover it either:

- the EVM contract fallback in `TokenPriceRepository` is gated on `TokenStandard.EVM`
- `fetchThorPoolPrices` finds no match — there is no RUJI pool on THORChain
- `fetchThorContractPrices` only lists `x/nami*`, `x/staking-tcy` and BRUNE/yBRUNE

With the cache empty, the DeFi cards fell through to `getPriceByContactAddress`, which for an `x/…` denom can only ever return zero: THORChain has neither a CoinGecko asset-platform id (`Chain.coinGeckoAssetPlatformId()`) nor a LI.FI chain id. Both lookups throw and are swallowed, so every refresh also logged two stack traces:

```
IllegalStateException: No CoinGecko asset id for chain ThorChain
IllegalStateException: lifi chain id not found for chain ThorChain
```

## Changes

- `Coins.kt` — `priceProviderID` on RUJI and sRUJI is now `rujira`, matching iOS (`TokensStore.swift:3174`, `:3184`).
- `MIGRATION_40_41` — coins are persisted per vault when added, so correcting the static definitions alone would leave every already-added RUJI stuck at `$0.00` until the user removed and re-added it. The migration rewrites `priceProviderId = 'ruji'` → `'rujira'` in the `coin` table.
- `DefiFiatValueCalculator` / `ThorchainDefiPositionsViewModel.priceFor` — try the coin's price-provider id before the contract address. A coin carrying a CoinGecko id should never be priced through a route its chain cannot support; this also stops the two exception stacks per refresh.

## Test plan

- [x] `DefiFiatValueCalculatorTest` — two new cases pin the ordering: an uncached coin resolves through its price-provider id without touching the contract route, and a provider id that resolves to nothing still falls through to the contract address. Existing cases unchanged.
- [x] `./gradlew :app:testDebugUnitTest --tests DefiFiatValueCalculatorTest --tests ThorchainDefiPositionsViewModelTest` passes.
- [x] `rujira` verified live against both CoinGecko and the `api.vultisig.com/coingeicko` proxy.
- [ ] Manual: open the THORChain DeFi tab on a vault with staked RUJI and confirm the Compounded RUJI / Staked RUJI cards show a non-zero fiat value, and that logcat no longer reports the two `ThorChain` price-lookup exceptions.
- [ ] Manual: upgrade an install that already has RUJI added (DB v40 → v41) and confirm the stored price-provider id is rewritten without a data loss.

No UI code changed — this is a pricing/data-layer fix, so the rendered value changes but no layout, style or composable does.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DeFi fiat value calculations by checking provider-based prices before contract-address pricing.
  * Added fallback pricing when cached values are unavailable.
  * Corrected RUJI and sRUJI price-provider mappings for more reliable token valuations.
  * Updated stored coin data automatically during the database upgrade.

* **Tests**
  * Added coverage for provider-based pricing and contract-address fallbacks.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
@@ -26,6 +26,7 @@ constructor(private val tokenPriceRepository: TokenPriceRepository) {
     suspend fun createFiatValue(amount: BigDecimal, coin: Coin, currency: AppCurrency): FiatValue =
         convert(amount, currency) {
             tokenPriceRepository.getCachedPrice(tokenId = coin.id, appCurrency = currency)
+                ?: priceByProviderId(coin)
                 ?: tokenPriceRepository.getPriceByContactAddress(
                     coin.chain.id,
                     coin.contractAddress,
@@ -49,6 +50,18 @@ constructor(private val tokenPriceRepository: TokenPriceRepository) {
                 appCurrency = currency,
             ) ?: tokenPriceRepository.getPriceByContactAddress(chain.id, contractAddress)
         }
+
+    /**
+     * The CoinGecko-id route, tried before the contract-address one. THORChain has neither a
+     * CoinGecko asset-platform id nor a LI.FI chain id, so the contract route can only ever return
+     * zero for `x/…` denoms like RUJI — a position whose coin carries a price-provider id must go
+     * through it or the card renders $0.00.
+     */
+    private suspend fun priceByProviderId(coin: Coin): BigDecimal? =
+        coin.priceProviderID
+            .takeIf { it.isNotEmpty() }
+            ?.let { tokenPriceRepository.getPriceByPriceProviderId(it) }
+            ?.takeIf { it > BigDecimal.ZERO }
 
     private suspend fun convert(
         amount: BigDecimal,
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
@@ -957,6 +957,10 @@ constructor(
     private suspend fun priceFor(coin: Coin, currency: AppCurrency): BigDecimal =
         try {
             tokenPriceRepository.getCachedPrice(tokenId = coin.id, appCurrency = currency)
+                ?: coin.priceProviderID
+                    .takeIf { it.isNotEmpty() }
+                    ?.let { tokenPriceRepository.getPriceByPriceProviderId(it) }
+                    ?.takeIf { it > BigDecimal.ZERO }
                 ?: tokenPriceRepository.getPriceByContactAddress(
                     coin.chain.id,
                     coin.contractAddress,
--- a/app/src/main/java/com/vultisig/wallet/data/db/AppDatabase.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/db/AppDatabase.kt
@@ -64,7 +64,7 @@ import com.vultisig.wallet.data.db.models.VaultOrderEntity
             TransactionHistoryEntity::class,
             PendingLimitOrderEntity::class,
         ],
-    version = 40,
+    version = 41,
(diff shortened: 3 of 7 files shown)


johnnyluo (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
```diff
@@ -26,6 +26,7 @@ constructor(private val tokenPriceRepository: TokenPriceRepository) {
     suspend fun createFiatValue(amount: BigDecimal, coin: Coin, currency: AppCurrency): FiatValue =
         convert(amount, currency) {
             tokenPriceRepository.getCachedPrice(tokenId = coin.id, appCurrency = currency)
+                ?: priceByProviderId(coin)
```
rkokhatskyi, line 29: `createFiatValueFromPoolAsset` disagrees with the just-fixed `createFiatValue` because it only tries `getPriceByContactAddress`, never `priceProviderID`, so a THOR.TCY or MAYA.* LP position not held as a vault coin still prices at $0.00 via `ThorchainDefiPositionsViewModel.priceForPoolAsset` too.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -957,6 +957,10 @@ constructor(
     private suspend fun priceFor(coin: Coin, currency: AppCurrency): BigDecimal =
         try {
             tokenPriceRepository.getCachedPrice(tokenId = coin.id, appCurrency = currency)
+                ?: coin.priceProviderID
```
rkokhatskyi, line 960: This fallback chain duplicates `DefiFiatValueCalculator.priceByProviderId` verbatim, though this ViewModel already injects `fiatValueCalculator` and calls it elsewhere, risking one copy getting fixed and the other silently missed. `ThorchainDefiPositionsViewModelTest.kt:107`'s blanket `getCachedPrice` stub means this exact branch has no test coverage today.

app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
```diff
@@ -50,6 +51,18 @@ constructor(private val tokenPriceRepository: TokenPriceRepository) {
...
+    /**
+     * The CoinGecko-id route, tried before the contract-address one. THORChain has neither a
+     * CoinGecko asset-platform id nor a LI.FI chain id, so the contract route can only ever return
+     * zero for `x/…` denoms like RUJI — a position whose coin carries a price-provider id must go
+     * through it or the card renders $0.00.
+     */
+    private suspend fun priceByProviderId(coin: Coin): BigDecimal? =
+        coin.priceProviderID
+            .takeIf { it.isNotEmpty() }
+            ?.let { tokenPriceRepository.getPriceByPriceProviderId(it) }
```
rkokhatskyi, line 63: `getPriceByPriceProviderId` (`TokenPriceRepository.kt:205-209`) never calls `savePrices` like its sibling `getPriceByContactAddress` does (`:191-203`), so every DeFi-tab visit re-fetches sRUJI/sTCY live from CoinGecko with no self-healing, since these synthetic receipt coins never enter the cached bulk `refresh()` path.

rkokhatskyi (changes requested):

Issue #5496, opened 2026-08-03
DeFi fiat pricing: contract-price fallback is dead code on non-EVM chains; cache rows exist only for vault-held coins


realpaaao:
### What happens

For DeFi position tokens on non-EVM chains (sTCY, yRUNE, yTCY, sRUJI, Maya pool assets), the fiat value silently resolves to zero/blank. Two mechanisms compound:

**1. The contract-price fallback is dead code on every non-EVM chain.**
Every DeFi fiat helper does cache-lookup → `getPriceByContactAddress` (`DefiFiatValueCalculator.kt:26-34`). That fallback:
- `CoinGeckoApi.coinGeckoAssetId` has no THORChain/Maya/TON/Cosmos branch → `else -> error(...)` (`CoinGeckoApi.kt:99`), swallowed to `emptyMap()` at `:56-58`
- LI.FI fallback is EVM-only (`LiQuestApi.kt:33`), swallowed at `TokenPriceRepository.kt:249-256`
- The resulting `null` is mapped to a **non-empty** `mapOf(currency to ZERO)` (`TokenPriceRepository.kt:243-246`), so `getPriceByContactAddress` (`:180-192`) returns zero *and writes a poisoned `price = "0"` row into Room*

Net: for THORChain/Maya DeFi the Room cache is the only price source — there is no live fallback.

**2. Cache rows only exist for coins the vault holds — DeFi position tokens usually aren't.**
`TokenPriceRepository.refresh(tokens)` is only ever called with vault coins (`AccountsRepository.kt:99`, `:425`, `:580`), and THORChain auto-discovery excludes these denoms (`TokenRepository.kt:104-109`). sRUJI is excluded from `Coins.coins` entirely and stripped on refresh (`TokenRefreshWorker.kt:97-102`). So a user who staked sTCY/yTCY/yRUNE without manually enabling those tokens has no cache row, and by (1) no fallback. The cache key is the token id, so sTCY having `priceProviderID = "tcy"` (`Coins.kt:4091`) doesn't help — its row is never populated.

iOS avoids both: rates are keyed by priceProviderId/contract independent of vault membership (`RateProvider.swift:38-48`), sTCY reads TCY's rate for free (`TokensStore.swift:3121`), and THORChain/Maya contracts price via dedicated sources incl. pool depth (`TokenPriceSource.swift:15-29`, `:83-186`).

### Expected

[...]

PR #5523, opened 2026-08-05
fix(defi): show zero or unavailable fiat instead of hiding the row

### LINKED ISSUE #5497: DeFi cards hide fiat values instead of showing $0.00/unavailable — failures read as missing dollar values
### What happens

When a fiat value can't be computed, the DeFi UI hides the row entirely instead of showing a zero or an "unavailable" state — so any pricing/API failure reads as "the dollar value is missing" (the current user reports). During today's THORNODE gateway outage, staking cards degrade to a bare ticker with no amount and no fiat.

### Receipts

Hide gates:
- `if (state.stakedFiatDisplay.isNotEmpty()) { Text(...) }` — `app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/StakingTabComponents.kt:120`
- `if (totalPrice.isNotEmpty()) { ... }` — `DeFiComponents.kt:617` (bonded header)

Producers of `""`:
- `calculateBondedFiatPrice` returns `""` on any exception — `ThorchainDefiPositionsViewModel.kt:425` (origin/main); Maya equivalents at `MayachainDefiPositionsViewModel.kt:429`, `:445`
- Model defaults — `DefiTabUiModels.kt:20` (`totalBondedPrice`), `:48`; `emptyBondedTabUiModel()` never sets it (`DeFiExtensions.kt:77-82`)
- Largest producer: the placeholder-card path. `loadDefaultStakingPositions()` seeds cards without fiat (`ThorchainDefiPositionsViewModel.kt:1207-1293`), and every service `.catch` handler just flips `isLoading = false` without filling amount/fiat (`:514-531` RUJI, `:594-611` TCY, `:650-671` yRUNE/yTCY/sTCY; `MayachainDefiPositionsViewModel.kt:341-352`)

Related header/card disagreement:
- `.catch` handlers never flip the internal `_isLoadingTotalAmount`, and failed legs stay ZERO — the total header can read the hardcoded `$0.00` (`ThorchainDefiPositionsViewModel.kt:1193`, a USD literal that also ignores the locale currency formatting from #5458) while cards show real values
- LP positions are excluded from the total entirely (`:257-263`) — a user with only LPs sees `$0.00` above populated LP cards

### Expected

- Render a currency-formatted zero or an explicit "price unavailable" state instead of hiding rows; remove the `isNotEmpty()` gates
- Fill or mark placeholder cards on `.catch` instead of leaving them blank
- Flip the loading flag in every `.catch`, include LP in the total, derive the zero-balance default from `appCurrencyRepository.getCurrencyFormat()`

aminsato:
Closes #5497

## Problem

When a DeFi position's fiat value couldn't be produced, the UI dropped the price line entirely. A pricing or API failure therefore read as *"the dollar value is missing"* rather than *"the price isn't available"* — and during a THORNODE outage the staking cards degraded to a bare ticker with no amount and no fiat at all.

Three `isNotEmpty()` gates did the hiding; the blanks feeding them came from placeholder cards that were never filled in, `.catch` handlers that only flipped a spinner, and model defaults of `""`.

## Approach

Fiat display fields are now nullable. `null` means *"we could not resolve a price"* and renders as an explicit unavailable marker (`—`); a resolved price is always a formatted string. That keeps a genuine zero distinguishable from a pricing failure, and neither is ever hidden.

## Changes

- **Removed all three hide gates** — `StakingTabComponents.kt`, and both `HeaderDeFiWidget` overloads in `DeFiComponents.kt`.
- **Placeholder staking cards** carry `"0 TICKER"` and a formatted zero instead of a bare ticker. `selectionKey()` was keying the Manage-Positions filter off that same display string, so it now reads the ticker off the coin.
- **Every `.catch`** fills the amount and price it was leaving blank, and releases the total-loading flag it was leaving set — the bonded flow's handler never cleared it, which stranded the header.
- **LP joins the header total.** `TotalDefiValue` carried bond/RUJI/TCY/default-stake but no LP, so a holder whose value sat in LP saw a total that disagreed with the cards under it.
- **The zero balance is formatted through the app currency** instead of a hardcoded `"$0.00"`, which was wrong for every non-USD user. Constants removed from both ViewModels, `DefiUiModel`, and the Circle screen.

Because the total header now starts in a loading state, every early bail-out has to settle it — a vault missing its RUNE coin with staking selected would otherwise shimmer forever. That path is covered by a test.

## Scope note

This is the presentation half of the problem. #5496 covers the data half — why some non-EVM positions resolve to a zero price in the first place. They're independent: a pricing failure there comes back as a formatted `$0.00` and is displayed, whereas this PR is about values that were blank and hidden. Landing this first means unpriceable positions read as "no price" rather than as a confident `$0.00`.

## Test plan

- 8 new unit tests across `ThorchainDefiPositionsViewModelTest` (27 → 32) and `MayachainDefiPositionsViewModelTest` (16 → 19), covering: LP reaching the header total on both chains, a failed bond load settling the total instead of stranding it, a failed staking load pricing at zero rather than blank, a German-locale zero containing no dollar sign, and the shimmer regression above.
- Full `:app` unit test suite passes.
- `./gradlew :app:lintDebug` passes.
- ktfmt applied.

Manual, on device:
- With airplane mode on, open **DeFi → THORChain → Staked**: cards read `0 RUJI` / `$0.00` instead of a bare ticker with no fiat line. Same for the Bonded header, and for **DeFi → MayaChain**.
- **Settings → Currency → EUR**, then reopen the DeFi screen: every zero renders `0,00 €` instead of `$0.00`.
- **Manage Positions**: toggling staking positions off and on still filters to the expected cards (regression check for the `selectionKey` change).

## Summary by CodeRabbit

* **Bug Fixes**
  * DeFi totals now include liquidity pool values across supported networks.
  * Totals remain loading until all balance sources are resolved, preventing incomplete or stale values.
  * Unavailable fiat prices display an em dash instead of blank or misleading zero values.
  * Currency changes refresh liquidity pool, bonded, and staking prices correctly.
  * Empty and failed states show properly formatted zero balances where applicable.
  * Staking placeholders and balance displays now use consistent currency formatting.

* **Tests**
  * Expanded coverage for aggregation, repricing, failures, synchronization, and zero-value formatting.


app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -252,7 +259,13 @@ constructor(
                 .catch { t ->
                     Timber.e(t)
                     _totalBondedRaw.value = BigInteger.ZERO
-                    updateModel { it.copy(bonded = it.bonded.copy(isLoading = false)) }
+                    val zero = zeroFiat()
+                    updateModel {
+                        it.copy(
+                            isTotalAmountLoading = false,
+                            bonded = it.bonded.copy(isLoading = false, totalBondedPrice = zero),
```
johnnyluo: This `.catch` block correctly settles `totalBondedPrice` to a formatted zero on failure — but two nearby sibling paths in the same `loadBondedNodes()` were not updated the same way:

1. The `if (cacaoCoin == null)` branch just above (line 253) only clears `bonded.isLoading`; `totalBondedPrice` stays at its default `null`, so the card renders `—` (unavailable) instead of `$0.00`. Compare `ThorchainDefiPositionsViewModel.kt`'s identical "vault does not have RUNE coin" branch, which this PR explicitly updated to set `totalBondedPrice = zeroFiat()`.
2. The outer `catch (t: Throwable)` a bit further down (lines 307-314, wrapping the whole `vaultRepository.get()` call) has the same gap — it clears `isLoading` but never sets `totalBondedPrice`.

Since "no CACAO coin" genuinely means nothing is bonded, `$0.00` reads better than an unavailable marker here, matching the rationale this PR uses elsewhere.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -348,17 +409,33 @@ constructor(
...
+                        .catch { t ->
+                            Timber.e(t)
+                            // The bond leg is one input to the header total. Leaving the flag set
+                            // would strand the header on its pre-load value, so settle it at zero
+                            // instead of letting a failed leg look like a total that never arrived.
+                            val zero = zeroFiat()
+                            state.update {
+                                it.copy(
+                                    bonded =
+                                        it.bonded.copy(isLoading = false, totalBondedPrice = zero)
```
johnnyluo, line 556: This `.catch` on the `flatMapLatest` collector correctly settles the bonded price and calls `updateTotalValueStatus`. However the outer `catch (t: Throwable)` further down (around lines 466-475), which wraps the whole `try` block including `vaultRepository.get(vaultId)`, was not given the same treatment — it only does:
```kotlin
state.update {
    it.copy(isTotalAmountLoading = false, bonded = it.bonded.copy(isLoading = false))
}
```
No `totalBondedPrice` fallback and no `updateTotalValueStatus(...)` call to release `_isLoadingTotalAmount`. This is pre-existing code the diff did not touch, but it is the exact bug class this PR sets out to close ("every early bail-out has to settle it", "every `.catch` fills the amount and price it was leaving blank"). Low likelihood trigger (`vaultRepository.get()` throwing), but worth closing for consistency with the sibling catch here and in `loadStakingPositions()`'s outer catch (which does release the flag).

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -648,24 +682,10 @@ constructor(
...
-                                            } else {
-                                                position
-                                            }
-                                        }
-                                )
-                        )
+                    settleStakingPositions {
+                        it.coin.id == Coins.ThorChain.yRUNE.id ||
+                            it.coin.id == Coins.ThorChain.yTCY.id ||
+                            it.coin.id == Coins.ThorChain.sTCY.id
```
johnnyluo, line 834: Minor: a bit further down in this same function, the generated position uses `stakedFiatDisplay = stakedFiatByCoinId[coin.id].orEmpty()` (line 734). Now that `stakedFiatDisplay` is `String?`, this coerces a missing/failed lookup to `""` instead of `null` — which would render as a blank fiat line rather than the intended `—` (unavailable) marker. In practice the map is always populated for every `coin.id` in `loadedPositions` by construction, so this is unreachable today, but it is the one call site touched by this PR that does not honor the new "null = unavailable" contract. Suggest dropping `.orEmpty()` so a future refactor of the matching logic fails safely into `—` rather than silently into a blank string.

johnnyluo (commented):
Follow-up review comments from a closer read — flagging a couple of spots where the same failure-handling pattern applied elsewhere in this PR was missed nearby. Not blocking, just worth a look.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -252,7 +259,13 @@ constructor(
                 .catch { t ->
                     Timber.e(t)
                     _totalBondedRaw.value = BigInteger.ZERO
```
rkokhatskyi, line 399: The `cacaoCoin == null` branch just above (`:250-254`) never sets `totalBondedPrice`, so it stays null and renders as the unavailable dash instead of a real zero. This catch block does call `zeroFiat()`, and Thorchain's identical `runeCoin == null` branch (`ThorchainDefiPositionsViewModel.kt:410-415`) does too, so this one path was missed

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -348,17 +409,33 @@ constructor(
...
+                            Timber.e(t)
+                            // The bond leg is one input to the header total. Leaving the flag set
+                            // would strand the header on its pre-load value, so settle it at zero
+                            // instead of letting a failed leg look like a total that never arrived.
+                            val zero = zeroFiat()
+                            state.update {
+                                it.copy(
+                                    bonded =
+                                        it.bonded.copy(isLoading = false, totalBondedPrice = zero)
+                                )
```
rkokhatskyi, line 557: The outer `catch (t: Throwable)` a few lines below (`:469-473`) sets `isTotalAmountLoading = false` directly instead of calling `updateTotalValueStatus` like this catch does. `totalAmountPrice`, only ever set from that pipeline, can stay null forever while the header spinner is already gone

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -890,7 +917,7 @@ constructor(
             lpAssetLogoRes(parsed.chain, parsed.ticker, parsed.contractAddress) ?: (logo as? Int)
         return LpPositionUiModel(
             titleLp = "$ticker Pool",
-            totalPriceLp = DEFAULT_ZERO_BALANCE,
+            totalPriceLp = zeroBalance.value,
```
rkokhatskyi: `totalPriceLp = zeroBalance.value` reads a StateFlow the async write at `:175` may not have resolved yet, with no synchronization between them. When that races, the RUNE-coin-missing (`:842-849`) and fetch-failure (`:902-909`) branches freeze the null into the terminal LP state, permanently showing the dash instead of a formatted zero

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -474,7 +509,7 @@ constructor(
                 val assetTicker = pool.ticker.substringAfter("/")
                 LpPositionUiModel(
                     titleLp = "${pool.ticker} Pool",
-                    totalPriceLp = MayachainDefiPositionsUiModel.DEFAULT_ZERO_BALANCE,
+                    totalPriceLp = zeroBalance.value,
```
rkokhatskyi: Same unsynchronized race as the Thorchain sibling: `totalPriceLp = zeroBalance.value` here can still be null when the async fill at `:177` hasn't resolved. The cacaoCoin-missing (`:538-543`) and address-invalid (`:549-554`) branches freeze that null into the terminal LP state instead of a formatted zero

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -224,6 +236,9 @@ constructor(
                         isLoading = isLoading,
                     )
                 }
+                .combine(totalValueLpFiat) { totalValue, lpFiat ->
```
rkokhatskyi: This merges the new LP leg into the total but never gates loading on it. `_isLoadingTotalAmount` (set at lines 481/516/543/565/590/670/764) is flipped false by bond and staking alone, so the header can settle a total that excludes a still-loading LP position, then silently jump once it resolves

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -181,8 +186,10 @@ constructor(
 
     private fun observeTotalRaw(): Job =
         viewModelScope.launch {
-            combine(_totalBondedRaw, _totalStakingRaw) { bonded, staking -> bonded + staking }
-                .collect { totalRaw -> updateTotalFiatValue(totalRaw) }
+            combine(_totalBondedRaw, _totalStakingRaw, _totalLpFiat) { bonded, staking, lpFiat ->
```
rkokhatskyi, line 257: This combine has no loading gate at all, unlike Thorchain's version. It fires on the seeded zero StateFlows the instant observeTotalRaw subscribes, so updateTotalFiatValue commits $0.00 with isTotalAmountLoading=false before bond, staking, or LP have actually loaded, defeating the loading-default flip this PR just made

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
```diff
@@ -648,15 +657,15 @@ internal fun HeaderDeFiWidget(
...
 
-                        Text(
-                            text = if (isBalanceVisible) totalPrice else HIDE_BALANCE_CHARS,
-                            style = Theme.brockmann.supplementary.footnote,
-                            color = Theme.v2.colors.text.tertiary,
-                        )
-                    }
+                    Text(
+                        text =
+                            if (isBalanceVisible) totalPrice ?: FIAT_VALUE_UNAVAILABLE
```
rkokhatskyi, line 664: Removing the isNotEmpty() gate here helps Thorchain/Maya but regresses `SolanaStakingPositionsScreen.kt:133`, an untouched consumer of this same widget whose `totalStakedFiatDisplay` still defaults to "" on load failure (`SolanaStakingPositionsViewModel.kt:97,291-299`). That empty string skips the `?: FIAT_VALUE_UNAVAILABLE` elvis and now renders a blank price line instead of the old hidden row

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -512,21 +574,7 @@ constructor(
...
-                                            if (position.coin.id in RUJI_POSITION_COIN_IDS) {
-                                                position.copy(isLoading = false)
-                                            } else {
-                                                position
-                                            }
-                                        }
-                                )
-                        )
-                    }
+                    settleStakingPositions { it.coin.id in RUJI_POSITION_COIN_IDS }
```
rkokhatskyi, line 709: None of the three staking `.catch` blocks (here, `:641-643`, `:683-690`) release `_isLoadingTotalAmount`, unlike loadBondedNodes' catch. Since this PR flips the loading default to true, a staking failure with no bond selected strands the header spinner forever; adding an isTotalAmountLoading assertion to the PR's own failed-staking test fails, stuck true

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
```diff
@@ -80,7 +80,6 @@ constructor(
     private val _state =
         MutableStateFlow(
             DefiUiModel(
-                totalAmountPrice = "$0.00",
                 isTotalAmountLoading = true,
```
rkokhatskyi, line 83: Removing the hardcoded "$0.00" default with no replacement regresses the no-account path: loadCirclePositions' branch at `:169-174` never sets totalAmountPrice or totalDepositCurrency, so every brand-new Circle user now sees the unavailable dash on first load instead of a genuine zero

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -254,8 +269,10 @@ constructor(
                         fiatValueCalculator.createFiatValue(decimalAmount, position.coin, currency)
                     }
 
+                val lpFiatValue = FiatValue(totalValue.lpFiatValue, currency.ticker)
```
rkokhatskyi: `_totalValueLpFiat` caches a magnitude priced under whichever currency was active when reloadLpTab last finished; this line relabels it with a freshly-read currency instead of reconverting. A currency change plus any bond or staking refresh mixes the stale LP amount into the new total until LP catches up

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -403,7 +421,7 @@ constructor(
                     withContext(ioDispatcher) { appCurrencyRepository.getCurrencyFormat() }
                 updateModel {
                     it.copy(
-                        totalAmountPrice = currencyFormat.format(fiatValue.value),
+                        totalAmountPrice = currencyFormat.format(fiatValue.value.add(lpFiat)),
```
rkokhatskyi: Same currency-staleness bug as the Thorchain sibling, but with no guard at all since this is a plain BigDecimal.add, not FiatValue.plus. _totalLpFiat keeps its old-currency magnitude until reloadLpTab's network fetch (`~557-580`) completes, so a currency change plus a refresh mixes two currencies into one total

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Fixed. This was the root of most of the others, so I changed the shape rather than patching the call site.

The header total was guarded by one shared `_isLoadingTotalAmount` that any leg could clear — the wrong model for a value summed from independent loads, since whichever leg finished first declared the header done for everyone. Each leg now holds a nullable total where `null` means "has not reported yet", and the total is priced only once every leg is non-null. The seeded-zero emission you spotted no longer produces anything, because bond, stake and LP all start unreported instead of at zero.

The property that made it worth doing: the value and the settle signal are now the same write, so a bail-out cannot report one without the other.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed. LP is now one of the five nullable legs, so it gates the total like the rest — the header cannot settle while LP is outstanding, and there is no jump when it lands.

Worth flagging one case the gate needed on top of that: an unselected staking source never runs its loader, so it reports zero up front rather than leaving the leg parked forever.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed. That catch now reports the leg (`_totalValueBond.update { BigInteger.ZERO }`) and prices the card at zero, exactly like the sibling `.catch` above it.

`updateTotalValueStatus` is gone — it was the only way to set the flag and the leg separately, which is what allowed them to drift. Writing the leg is now the only way to settle it.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed. The `zeroBalance` StateFlow is gone from both ViewModels — the placeholders are now built inside the LP job, after `zeroFiat()` has actually resolved, so there is nothing left to race against.

Covered by `a failed LP load leaves the placeholder priced at zero, not unavailable` in both test classes.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Fixed, the same way as the Thorchain sibling: `zeroBalance` deleted, placeholders built inside the job after `zeroFiat()` resolves. Both the cacaoCoin-missing and address-invalid branches now freeze in a real formatted zero.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed. All three now report their raw total in the `.catch` (`_totalValueRujiStake` / `_totalValueTCYStake` / `_totalValueDefaultStake`), so a failed refresh no longer prices the header off the previous run's amount.

`_isLoadingTotalAmount` is gone entirely — the legs are nullable now and the header is priced only once all five have reported, so releasing a shared flag is no longer something a handler can forget.

Your suggested assertion is in as `a failed RUJI load resets its leg instead of keeping the previous total`, which loads a real amount first and then fails a refresh, so it covers the stale-total half too.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Good catch, fixed. `_totalValueLpFiat` is a `FiatValue` now, so it carries the currency it was priced in instead of being relabelled with whatever is active at total time.

Two parts to it:
- A currency change drops the LP leg back to unreported and reloads it, so the header parks on its spinner until LP re-prices rather than showing a mixed-currency figure.
- The total refuses a leg whose currency no longer matches, which closes the window between the change and the reload landing.

`FiatValue.plus` already had the `require(currency == other.currency)` guard for exactly this — relabelling was walking straight past it.

Covered by `switching currency re-prices LP instead of relabelling the old magnitude`.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Fixed the same way as the Thorchain sibling — and you are right that this one was worse, since a plain `BigDecimal.add` has no guard to walk past in the first place.

`_totalLpFiat` is a `FiatValue` now, a currency change invalidates and reloads the leg, and `updateTotalFiatValue` returns early if the LP leg's currency does not match the one it is pricing in.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
aminsato: You are right, this regressed an untouched consumer. Fixed by making `SolanaStakingPositionsUiState.totalStakedFiatDisplay` nullable, so it follows the same contract as every other fiat field here: `null` is "could not be priced", and the empty-string default that slipped past the elvis is gone.

`SolanaHeaderBanner` in the same screen took `String` too and had the identical problem, so it is nullable now as well.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/circle/CircleDeFiPositionsViewModel.kt
aminsato: Fixed. That branch now resolves a formatted zero and sets both `totalAmountPrice` and `totalDepositCurrency`.

Holding no Circle account is a genuine zero rather than a price we failed to resolve, so it reads better as a real value than as the unavailable marker — same reasoning as the "no CACAO coin" bond branch elsewhere in this PR.

Covered by `a user with no Circle account yet sees a real zero, not the unavailable dash`.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Both fixed. The `cacaoCoin == null` branch and the outer `catch` now price the card at zero, matching the Thorchain branch you pointed at.

Agreed on the reasoning — "no CACAO coin" means nothing is bonded, which is a real zero rather than a price we failed to resolve, so `$0.00` is the honest reading and the dash would not be.

Covered by `a vault without CACAO prices the bond card at zero rather than unavailable` and `a bond load that throws before the flow starts still prices the card`.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed — and it turned out to be worth more than the low trigger likelihood suggested, because it was a symptom of the wrong shape rather than a one-off gap.

The header total was guarded by a single shared `_isLoadingTotalAmount` that any leg could clear, so a handler forgetting it was always going to be possible. Each leg now holds a nullable total where `null` means "has not reported yet", and the total is priced only once every leg is non-null. The value and the settle signal are the same write, so this catch cannot clear the spinner without also reporting the leg.

Covered by `a bond load that throws before the flow starts still settles the header`.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed — `.orEmpty()` dropped, so a missed lookup stays `null` and renders as the unavailable marker instead of a blank line.

Agreed it is unreachable today; taking it because it was the one touched call site not honouring the new contract, and a blank line is the exact failure mode this PR exists to remove.

aminsato:
Thanks both — every finding was real and all are addressed in `00414fe` and `919857a`. Replies are on the individual threads; the shape change is worth summarising here because it is what most of the review resolved to.

**The header total no longer uses a shared loading flag.** `_isLoadingTotalAmount` was the wrong model for a value summed from several independent loads: whichever leg finished first declared the header done for everyone, so its siblings either published a partial figure that jumped as the rest landed, or — when the leg that owned the flag bailed out without clearing it — left the header on a permanent dash with no spinner to explain it.

Each leg now holds a nullable total where `null` means "this leg has not reported yet", and the header is priced only once every leg is non-null. The value and the settle signal are the same write, so a bail-out cannot report one without the other.

That covers the outer bond catch, Maya's combine firing on seeded zeros, LP being summed but never gating, and the three staking `.catch` handlers keeping stale raw totals.

Three cases the gate needed handling explicitly, all found by tests rather than by reading:

- A flow that completes without emitting is done, not pending — neither `.catch` nor `.collect` fires there, so every leg settles in `.onCompletion`.
- An unselected staking source never runs its loader, so it reports zero up front.
- A failing LP dialog fetch parks `reloadLpTab` before it can report; the tab deliberately stays loading there, but the leg still reports.

**Separately, the LP leg now carries the currency it was priced in.** The raw legs reconvert on every total, but LP joins already converted and was being relabelled rather than reconverted — `FiatValue.plus` guards exactly that, and relabelling walked past the guard. A currency change now invalidates and reloads the leg.

**Two consumers outside these tabs** were relying on behaviour this PR changed, both caught by @rkokhatskyi: Solana staking's empty-string fiat default drew a blank line once the `isNotEmpty()` gate went away, and Circle's no-account path never set a price after the hardcoded `$0.00` was removed. Both fixed.

66 tests green across the four ViewModels (Thorchain 37, Maya 24, Circle 5), full `:app` unit suite and `lintDebug` pass.

---

One CodeRabbit finding I have **not** taken: making `LpPositionUiModel.totalFiatValue` carry its own unavailable state. The leak it names is real — `toUiModel` falls back to `BigDecimal.ZERO` when an asset price lookup fails, so a pool contributes a RUNE-only partial to the total — but that is pre-existing pricing behaviour rather than something this PR introduced, and it is the subject of #5496. Fixing it properly means giving the header total an unavailable state too, which is a larger change than this issue calls for. Happy to pick it up there if you would rather it move now.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
rkokhatskyi: Confirmed fixed in the gate-the-header-total commit — `MayachainDefiPositionsViewModel.kt:299-309` now calls `zeroFiat()` and sets `totalBondedPrice` on this branch too, matching the sibling below it and Thorchain's equivalent.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -214,80 +241,183 @@ constructor(
...
+     * rather than on a figure mixing two currencies.
+     */
+    private fun observeCurrencyChanges(): Job =
+        viewModelScope.launch {
+            appCurrencyRepository.currency
+                .map { it.ticker }
+                .distinctUntilChanged()
+                .drop(1)
+                .collect {
+                    _totalValueLpFiat.value = null
```
rkokhatskyi: observeCurrencyChanges nulls the LP leg but never resets isTotalAmountLoading. Since the combine only emits once every leg is non-null, handleTotalValueUpdate cannot re-run while LP reprices, so the header keeps showing the stale prior-currency total with no spinner until the network refetch lands

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -177,12 +190,51 @@ constructor(
...
+     * rather than on a figure mixing two currencies.
+     */
+    private fun observeCurrencyChanges(): Job =
+        viewModelScope.launch {
+            appCurrencyRepository.currency
+                .map { it.ticker }
+                .distinctUntilChanged()
+                .drop(1)
+                .collect {
+                    _totalLpFiat.value = null
```
rkokhatskyi: Same gap as the Thorchain sibling: nulling _totalLpFiat here starves observeTotalRaw's combine until reloadLpTab's network fetch resolves, but nothing resets isTotalAmountLoading, so the header keeps the pre-switch total with no loading indicator for the whole round trip

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -455,6 +552,7 @@ constructor(
         if (selectedLpKeys.isEmpty()) {
             loadLpJob?.cancel()
             updateModel { it.copy(lp = LpTabUiModel(isLoading = false, positions = emptyList())) }
+            loadLpJob = viewModelScope.launch { reportLpFiat(BigDecimal.ZERO) }
```
rkokhatskyi, line 658: The sibling branch above reports LP correctly; the lpPositionsDialog.isEmpty() check three lines below can't tell not-yet-loaded from loaded-and-genuinely-empty. With a persisted LP key selected and getMayaNodePools() legitimately returning [], _totalLpFiat never reports and isTotalAmountLoading is stuck true forever, surviving pull-to-refresh

app/src/test/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModelTest.kt
```diff
@@ -271,6 +275,138 @@ internal class MayachainDefiPositionsViewModelTest {
...
+        successData(vm).totalAmountPrice shouldBe "$0.40"
+
+        coEvery { appCurrencyRepository.getCurrencyFormat() } returns
+            NumberFormat.getCurrencyInstance(Locale.GERMANY)
+        currency.value = AppCurrency.EUR
+
+        // Re-priced under the new currency rather than carrying the old number across.
+        val settled = successData(vm)
+        settled.isTotalAmountLoading shouldBe false
+        settled.totalAmountPrice?.contains("$") shouldBe false
```
rkokhatskyi: This assertion only checks the total lacks a $. Mutating observeCurrencyChanges to relabel the stale LP magnitude with the new ticker instead of re-pricing still passes it, since the German-formatter mock never emits $ regardless of the value. The named regression is undetectable

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -214,80 +241,183 @@ constructor(
...
+        _totalValueLpFiat.value = FiatValue(value, currency.ticker)
+    }
+
+    /**
+     * Re-prices LP when the user switches currency. The bond and staking legs are held as raw chain
+     * amounts and convert afresh on every total, but LP is stored already converted and cannot be
+     * re-based — so it drops back to unreported and reloads, which parks the header on its spinner
+     * rather than on a figure mixing two currencies.
+     */
+    private fun observeCurrencyChanges(): Job =
```
rkokhatskyi, line 393: observeCurrencyChanges only re-prices the LP leg. calculateBondedFiatPrice and calculateStakingFiatPrice run once per load from one-shot flows and are never re-invoked here, so after a currency switch the header total updates correctly but the Bonded and Staking cards permanently keep the old currency

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -177,12 +190,51 @@ constructor(
...
+        _totalLpFiat.value = FiatValue(value, currency.ticker)
     }
 
+    /**
+     * Re-prices LP when the user switches currency. The bond and staking legs are held as raw chain
+     * amounts and convert afresh on every total, but LP is stored already converted and cannot be
+     * re-based — so it drops back to unreported and reloads, which parks the header on its spinner
+     * rather than on a figure mixing two currencies.
+     */
+    private fun observeCurrencyChanges(): Job =
```
rkokhatskyi, line 223: Same shape as the Thorchain sibling: this only nulls and reloads the LP leg. calculateBondedFiatPrice and calculateStakingFiatPrice run once from one-shot flows and are never re-triggered here, so a currency switch updates the header total but leaves the Bonded and CACAO Staking cards on the old currency

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -866,31 +979,39 @@ constructor(
                                     currencyFormat,
                                 )
                             } else {
-                                dialogPool.toPlaceholderUiModel()
+                                dialogPool.toPlaceholderUiModel(zero)
                             }
                         }
 
+                    _totalValueLpFiat.value =
```
rkokhatskyi, line 1063: A transient per-pool failure is caught and swallowed inside GetThorChainLpPositionsUseCase, leaving that pool's placeholder totalFiatValue at zero. This fold sums it into a confident header total anyway, flipping isTotalAmountLoading false with no error indicator and silently understating the real total

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed. This is the leg gate's own failure mode, and you named it exactly: the combine cannot emit while a leg is null, so dropping the LP leg quietly turned "in flight" into "settled, in the currency you just left".

`observeCurrencyChanges` goes through `resetTotalsToPending()` now, which drops every leg **and** clears `totalAmountPrice` / raises `isTotalAmountLoading` in the same step. The value and the pending signal move together, which is the same rule the leg gate itself follows — half a state change was the bug. The currency-mismatch branch in `handleTotalValueUpdate` sets the same pending state as a fallback.

Covered by `switching currency parks the header on its spinner, not on the old total`, which holds the LP re-price so the mid-switch state is actually observable. I reverted the fix to confirm the test fails without it.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Fixed the same way as the Thorchain sibling. `observeCurrencyChanges` calls `resetTotalsToPending()`, which drops all three legs and clears `totalAmountPrice` / raises `isTotalAmountLoading` in one step, so the header cannot sit on a settled pre-switch figure while `reloadLpTab` is out on the network. `updateTotalFiatValue`'s mismatch branch sets the same state defensively.

Covered by `switching currency parks the header on its spinner, not on the old total` in the Maya test class.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Fixed. `lpPositionsDialog.isEmpty()` is gone — there is an explicit `lpDialogLoaded` flag now that flips true when the pool fetch returns, whatever it returns. That is the same shape Thorchain already had with its nullable `availablePools`; Maya was the one conflating the two states.

A chain that legitimately reports no pools now falls through to the `selectedPools.isEmpty()` branch and reports the LP leg as zero, so the header settles instead of spinning through pull-to-refresh.

Covered by `a chain reporting no LP pools still settles the header`, which selects a persisted LP key and has `getMayaNodePools()` return an empty list — the exact case you described.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed, and the diagnosis was right: the header was the only thing re-pricing because LP is the only leg that stores a currency. `calculateBondedFiatPrice` and `calculateStakingFiatPrice` format once per load out of one-shot flows and have no re-entry point at all.

A currency change is a full reload now — bond, staking and LP — so every card re-converts alongside the header.

One thing worth recording since it shaped the fix: nothing observed currency changes before this PR (currency was only ever read with `.first()`), so the stale cards are not new. What was new is the header and the cards *disagreeing*, which is what re-pricing LP alone introduced.

Covered by `switching currency re-prices the bonded and staking cards, not just the header`, which stubs a different EUR price so re-converting and merely re-formatting produce different numbers.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Fixed the same way as the Thorchain sibling — a currency change reloads bond, staking and LP rather than only nulling the LP leg, so the Bonded and CACAO Staking cards re-convert along with the header.

Covered by `switching currency re-prices the bonded and staking cards, not just the header` in the Maya test class, asserting both `bonded.totalBondedPrice` and the staking position's `stakedFiatDisplay`.

app/src/test/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModelTest.kt
aminsato: You are right, and it is the second time an assertion in this file has been satisfied by the formatter rather than by the value.

Both currency tests now stub a **different** price per currency, so a magnitude carried across unchanged and one genuinely re-priced are different numbers instead of the same number in a different format. Maya asserts `0,60 €` against the `$0.40` it started at; Thorchain asserts `9,00 €` against `$6.00`.

I verified it by the mutation you described — made `observeCurrencyChanges` relabel the stale LP magnitude with the new ticker instead of re-pricing it, and confirmed both currency tests fail where the old `contains("$")` assertion passed. I did the same for the other four fixes in this round: reverted each one and checked that exactly the intended tests fail.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Taken, and it needed a change in the use case rather than at the fold. `GetThorChainLpPositionsUseCase` returned nothing both for a pool that errored and for a pool the user holds no liquidity in, so the ViewModel had no way to tell them apart and read the second meaning for the first.

It returns `ThorChainLpPositions(positions, failedPools)` now, and the LP leg gained a third state — `LpLegTotal.Priced | Unavailable`, with `null` still meaning "has not reported yet" — so the header shows the unavailable marker instead of a confident understated figure. Only failures among the pools the user actually selected count: the use case queries every available pool, so an unrelated pool's outage must not blank a total that priced perfectly well.

Maya has the same shape without the per-pool granularity: a `getMemberDetails` failure fell back to an empty `MayaMemberDetails`, which reads as "this vault holds no liquidity anywhere". That reports unavailable too. A `getLpPoolStats` failure deliberately stays priced — the added-amounts branch it falls back to is still a real, if approximate, valuation, and marking it unavailable would regress a working display.

Covered by `a pool that failed to load makes the header unavailable rather than understated`, `a failure in a pool the user did not select leaves the total priced`, and Maya's `member details that fail to load make the header unavailable rather than zero`.

For transparency: this is the finding I pushed back on from CodeRabbit last round as belonging to #5496. With you raising it independently I took it here instead — #5496 still covers the other half, `toUiModel` falling back to zero on a failed asset-price lookup.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -214,80 +241,183 @@ constructor(
...
-                val currencyFormat =
-                    withContext(ioDispatcher) { appCurrencyRepository.getCurrencyFormat() }
+            val lpFiatValue = totalValue.lpFiatValue
+            if (lpFiatValue.currency != currency.ticker) {
+                // The currency changed after LP was priced. The raw legs re-convert on every run,
+                // but LP is stored already converted, so adding it now would sum two currencies.
+                // observeCurrencyChanges has already dropped the leg and asked for a re-price;
+                // leave the header on its previous value until that lands.
+                return
+            }
```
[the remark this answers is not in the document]
aminsato: Valid, and fixed — Roman flagged the same gap independently. `observeCurrencyChanges` now calls `resetTotalsToPending()`, which drops every leg and clears `totalAmountPrice` / raises `isTotalAmountLoading` in the same step, and the currency-mismatch branch sets the same pending state as the fallback you suggested.

I went further than the suggested diff on one point: re-pricing only LP left the Bonded and Staking cards in the previous currency, since their fiat strings are formatted once per load from one-shot flows. A currency change is a full reload of bond, staking and LP now.

The delayed-LP test you asked for is `switching currency parks the header on its spinner, not on the old total` in both ViewModel test classes.

app/src/test/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModelTest.kt
```diff
@@ -271,6 +275,138 @@ internal class MayachainDefiPositionsViewModelTest {
...
+    }
```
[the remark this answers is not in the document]
aminsato: Valid, and fixed along the lines you suggested — Roman raised the same point. Both currency tests now stub a different price per currency so a stale magnitude and a re-priced one are different numbers: Maya asserts `0,60 €` against the `$0.40` it started at, Thorchain `9,00 €` against `$6.00`.

Verified by mutation: relabelling the stale magnitude with the new ticker instead of re-pricing now fails both tests, where the old `contains("$")` assertion passed it.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1066,7 +1273,7 @@ constructor(
 
             loadBondedNodes()
 
-            launch { loadStakingPositions() }
+            loadStakingPositions()
```
rkokhatskyi, line 1288: onPositionSelectionDone never calls resetTotalsToPending the way observeCurrencyChanges does (`:400`). Adding a new LP pool while bond and staking stay unchanged leaves totalValueLpFiat and isTotalAmountLoading untouched, so the header keeps showing the pre-add total with no spinner until the new pool's async fetch lands, silently understating it

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -177,12 +193,74 @@ constructor(
...
+     *
+     * Legs keep their last value across a reload, so without this the header would go on showing a
+     * settled total — in the old currency — for as long as the refetch takes. Clearing the value
+     * and the legs together is what makes the spinner honest.
+     */
+    private fun resetTotalsToPending() {
+        _totalBondedRaw.value = null
+        _totalStakingRaw.value = null
+        _totalLpFiat.value = null
+        updateModel { it.copy(totalAmountPrice = null, isTotalAmountLoading = true) }
```
rkokhatskyi, line 252: loadSavedPositions's collector, retriggered by every onPositionSelectionDone DataStore write (`:288-293`), reloads bond, staking and LP without calling this first. Adding a new LP pool leaves _totalLpFiat at its pre-selection value during the async fetch, so the header shows a settled total missing the new leg until it lands

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Confirmed, and the asymmetry is exactly where you put it. `onPositionSelectionDone` runs the same three-leg reload as `observeCurrencyChanges` — bonded refresh trigger, `loadBondedNodes`, `loadStakingPositions`, `reloadLpTab` — and legs keep their last value across a reload, so all three sat on their pre-add values while the new pool's fetch was in flight. The header read as settled and short by exactly the pool just added, with no spinner to say more was coming.

Fixed in `234997af3`: `resetTotalsToPending()` now runs in `onPositionSelectionDone` before the triggers, same as the currency path.

I checked the settle side before adding it, since a reset is only safe if every leg reports again. `reloadLpTab` reports on all of them — the `selectedPools.isEmpty()` early return, the missing-RUNE bail, the catch, and both the priced and `Unavailable` outcomes. The one branch that doesn't is `pools == null`, and it calls `ensureAvailablePoolsLoaded()`, which re-enters `reloadLpTab` once the fetch lands; `onPositionSelectionDone` already calls it too. So the header can't be stranded on a leg that never arrives.

The test holds the LP fetch, adds `ETH.ETH` through the dialog on top of a settled `$6.00`, and asserts `totalAmountPrice == null` and `isTotalAmountLoading == true` before releasing it to `$12.00`. It fails on the pre-fix code.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Confirmed, and you traced the retrigger correctly — it's the same bug as the Thorchain one, just reached through DataStore instead of inline. `onPositionSelectionDone` writes, the `getSelectedPositions` collector re-fires, and it reloads LP, bonded and staking with all three legs still holding their pre-selection values. The header showed a settled total missing the new leg for the length of the fetch.

Fixed in `234997af3`. The reset goes in the collector rather than in `onPositionSelectionDone`, since the collector is the one point every Maya reload passes through — putting it at the call site would leave any other write to that key with the same problem. On the initial load it's a no-op against legs that are already unreported.

Same settle check as the Thorchain side: Maya's `reloadLpTab` reports the leg on every path — both `emptyList()` early returns, the missing/invalid CACAO bails, the `onError`, and the priced/`Unavailable` outcomes — so the reset can't strand the header.

The test backs `getSelectedPositions` with a flow that `saveSelectedPositions` writes into, so the reload is driven by the DataStore retrigger you described rather than by a direct call. It holds `getMemberDetails`, adds `BTC.BTC` to a settled `$0.00`, asserts unpriced and spinning, then lets it settle at `$0.40`. Fails on the pre-fix code.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1062,11 +1269,16 @@ constructor(
 
             loadedTabs.clear()
 
+            // Same reload set as a currency switch, so the header owes the same honesty: legs keep
+            // their pre-selection values, and without this a newly added pool would leave the total
+            // reading as settled — and short by that pool — until its fetch lands.
+            resetTotalsToPending()
```
rkokhatskyi, line 1282: onPositionSelectionDone calls resetTotalsToPending unconditionally with no check against the previous selection, and the Done button has no enabled gate. Opening Manage Positions and tapping Done without changing anything now blanks a settled header to a spinner and re-fetches bond, staking and LP for nothing

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -207,6 +285,14 @@ constructor(
                 updateModel {
                     it.copy(selectedPositions = positions, tempSelectedPositions = positions)
                 }
+
+                // Every onPositionSelectionDone write re-runs this collector, so it reloads all
+                // three legs for the same reason a currency switch does and owes the header the
+                // same honesty: legs keep their pre-selection values, and without this a newly
+                // added pool would leave the total reading as settled — and short by that pool —
+                // until its fetch lands.
+                resetTotalsToPending()
```
rkokhatskyi: This fires on every getSelectedPositions emission. Repeat identical writes are deduped by DataStore itself, but the first-ever save for a vaultId flips the key from absent to present, which DataStore can't treat as unchanged, so a vault's very first Manage Positions confirmation flashes an already-settled header to its spinner for no reason

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Real, and fixed in `320d8db20`.

Done is reachable without touching a checkbox and I reset unconditionally, so an unchanged confirm blanked a settled header and refetched all three legs for an identical selection. Guarded at the top of `onPositionSelectionDone` instead of gating the button — the button gate would still leave the ViewModel willing to do the work, and this covers any other caller:

```kotlin
if (selectedPositions.toSet() == state.value.selectedPositions.toSet()) {
    state.update { it.copy(showPositionSelectionDialog = false) }
    return@launch
}
```

Compared as sets because only membership decides what gets loaded. The save is skipped too — nothing changed, so there is nothing to persist. The `ensureAvailablePoolsLoaded()` retry isn't lost by the early return: `setPositionSelectionDialogVisibility(true)` already runs it when the dialog opens.

Test: `confirming a selection nothing changed in leaves the header settled` — holds the LP refetch so a reload would be visible as a spinner, then confirms without a change and asserts the header stays at $6.00 and nothing was written. Verified to fail on the pre-fix code.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Real, and your mechanism is exactly right — fixed in `320d8db20`.

DataStore does dedupe the repeat write (`DataStoreImpl` compares before writing), but the first-ever save flips the key from absent to present, which it can't read as unchanged, so the same selection came back as an emission and reloaded all three legs.

Deduped on the *derived* positions rather than the raw set, since that's the value the reload actually depends on:

```kotlin
.map { saved -> if (hasMayaPositions) saved.toList() else MAYA_DEFAULT_SELECTED_POSITIONS }
.distinctUntilChanged { old, new -> old.toSet() == new.toSet() }
```

Sets, not lists — the stored set doesn't have to come back in the order the default list has, and the absent→present flip is precisely the case where the two sides can differ only in order. Note this also covers emissions the write didn't come from: the store key is per-vault, so any write to that file re-emits here.

Added the same no-op guard to Maya's `onPositionSelectionDone` as well, which has a second benefit — an unchanged confirm no longer writes Maya's defaults over a selection this screen only ever reads through its Maya fallback.

Two tests, both verified to fail on the pre-fix code: `a store emission that leaves the selection alone keeps the header settled` (re-emits an equivalent set in a different order while the staking leg is held, asserts no spinner) and `confirming a selection nothing changed in writes nothing`.

Separately, while checking this: `DefiPositionsRepository` keys only on `selected_positions_$vaultId`, so the THORChain and Maya screens share one key per vault and either screen's save overwrites the other's selection. Pre-existing and out of scope here — want me to open a ticket?

aminsato:
@rkokhatskyi both were real and both are fixed in `320d8db20`, with the detail in the threads.

The round-4 reset was right for a selection that changed and wrong for one that didn't, and each screen reached that state by a different route:

- **Thor** reloads inline, so an unchanged Done reset and refetched directly. Guarded in `onPositionSelectionDone` rather than by gating the button — the ViewModel shouldn't be willing to do the work in the first place.
- **Maya** reloads off the saved-positions collector, and your read of the store is correct: the repeat write is deduped, but the first-ever save flips the key from absent to present, which can't be read as unchanged. Deduped on the derived positions there, plus the same no-op guard.

Compared as sets in both places: only membership decides what gets loaded, and a stored set doesn't have to come back in the order the default list has — which is exactly the absent→present case.

3 tests added, each verified to fail against the pre-fix code; 79 DeFi tests green.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -195,37 +273,54 @@ constructor(
...
+                        saved.any { it in MAYA_STATIC_POSITION_KEYS || it.contains(".") }
+                    if (hasMayaPositions) saved.toList() else MAYA_DEFAULT_SELECTED_POSITIONS
                 }
-                reloadLpTab()
+                // The store re-emits for writes that leave this vault's selection alone — the
+                // first-ever save flips the key from absent to present, which it cannot read as
+                // unchanged — and a reload for one of those would blank a settled header for
+                // nothing. Compared as sets: only membership decides what gets loaded, and a stored
+                // set can come back in a different order than the default list.
+                .distinctUntilChanged { old, new -> old.toSet() == new.toSet() }
```
rkokhatskyi, line 303: Deselecting every Maya position (Bond+Stake, no LP pools) re-derives via the empty-selection fallback to the same {BOND,STAKE} set the collector last emitted, so this swallows the reload: tabs correctly empty but the header total freezes at the pre-deselection amount until an unrelated reload trigger

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
rkokhatskyi: Yes please, worth its own ticket — it's a real cross-screen bug independent of this refactor. Everything in this round is otherwise solid; one new finding posted separately (the Maya dedup swallowing a full-deselect, since it collapses to the same default-fallback set the collector last saw).

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Real, and the mechanism is exactly as you describe — fixed in `e7404b36`.

Clearing everything writes an empty set, the `hasMayaPositions` fallback re-derived it to the same `{BOND, STAKE}` the collector last emitted, and the dedup dropped it. What made it visible rather than merely wasteful is that `onPositionSelectionDone` had already written the raw empty list into state: the tabs cleared while the header kept the total of the positions just removed, and the two stayed out of step until the ViewModel was recreated.

The general form of what round 5 introduced: **a dedup on a derived value swallows any change that derives back to what it replaced.** So the fix is to stop deriving empty away rather than to weaken the dedup:

```kotlin
saved.isEmpty() -> emptyList()
saved.any { it in MAYA_STATIC_POSITION_KEYS || it.contains(".") } -> saved.toList()
else -> MAYA_DEFAULT_SELECTED_POSITIONS
```

`getSelectedPositions` returns its own (non-empty) `DEFAULT_POSITIONS` for an absent key, so an empty set can only be one this screen cleared on purpose — the same invariant the Thorchain screen already leans on by using `saved` verbatim. The dedup stays as it is, and the state and the collector can no longer disagree about what is selected.

**One thing your comment assumes that was not true: the tabs were not correctly empty.** `loadBondedNodes` never looked at the selection at all, so unchecking Bond in Manage Positions did nothing — the card stayed and the header went on counting the leg. That is pre-existing and not from this PR, but the deselect you found only settles honestly once it is fixed, so I gated it here rather than leave the fix half-done:

```kotlin
if (!currentModel.selectedPositions.hasBondPositions()) {
    _totalBondedRaw.value = BigInteger.ZERO
    ...
    return
}
```

Reported as a real zero, not left unreported, so the leg still settles. It does not reuse `emptyBondedTabUiModel()` — that renders `0 ${Chain.ThorChain.coinType.symbol}`, i.e. `0 RUNE` on a CACAO card, and that symbol is a JNI call that throws `UnsatisfiedLinkError` under JVM unit tests. Say so if you would rather have the bond gate as its own PR.

Three tests, each verified to fail against the reverted ViewModel and only those three: the derivation rule driven from the store (`an emptied stored selection is honoured, not re-derived to the defaults`), the whole path through the dialog from a settled $200.00 to a settled $0.00 with both tabs empty, and the bond gate on its own. 125 DeFi tests green.

Ticket for the shared key is #5533.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Ticket opened: #5533 — the reproduction (deselect on one screen, confirm on the other, come back), why it currently reads as a silent reset rather than as data loss, and a per-chain key with a migration of the existing value onto the THORChain key so nobody's selection is cleared by the upgrade.

Worth noting there that the shared key is also what forces the `hasMayaPositions` heuristic, which is what your new finding fell out of — keying per chain would let both screens use the stored set verbatim and drop the fallback entirely.

rkokhatskyi (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
rkokhatskyi: Confirmed fixed in e7404b36. Good catch on the bond gate too — left it in the same PR since it's the exact scenario the deselect fix has to be honest about; splitting it would leave the same bug half-fixed. While re-checking this I found two more concrete manifestations of #5533 (clearing THORChain's selection now wipes Maya's tabs, and the else-fallback can silently revive a Bond position the user turned off) — added them to that ticket rather than reopening scope here, since the per-chain key fix you proposed resolves both. Approving.

PR #5526, opened 2026-08-06
fix(swap): show the provider logo wherever a provider is named

### LINKED ISSUE #5524: Swap provider logo missing on 3 of 5 provider surfaces (Figma) — plus no CowSwap icon and no compile-time exhaustiveness
## What

Per Figma, every place we surface a swap **provider** must show the provider **logo** next to its name — not the name as plain text.

Reference: [Vultisig App — node 45201-39285](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=45201-39285&t=BziM8pmI3KCoHvqF-0)

Already shipped on Windows/Extension in **[vultisig-windows#4572](https://github.com/vultisig/vultisig-windows/pull/4572)** (issue [#4568](https://github.com/vultisig/vultisig-windows/issues/4568)) — this is the Android side of the same design.

## Current Android state

A resolver already exists — `getProviderLogo(providerName)` in [`CoinLogo.kt`](app/src/main/java/com/vultisig/wallet/data/models/CoinLogo.kt) — but it is wired into only **two** of the five provider surfaces, and it does not cover every provider.

**Renders the logo today:**
- `VerifyProviderRow` — [`VerifySwapScreen.kt:674`](app/src/main/java/com/vultisig/wallet/ui/screens/swap/VerifySwapScreen.kt#L674)
- `ViaBadge` on the history card — [`SwapTransactionCard.kt:170`](app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SwapTransactionCard.kt#L170) (fed by `TransactionHistoryViewModel.kt:407`)

**Plain text, no logo:**
- Swap form `Provider` row — [`SwapFeeBreakdown.kt:77`](app/src/main/java/com/vultisig/wallet/ui/screens/swap/components/SwapFeeBreakdown.kt#L77) — `FormDetails2(value = quoteDisplay.provider.asString())`
- Transaction-done screen — [`TransactionDoneScreen.kt:247`](app/src/main/java/com/vultisig/wallet/ui/screens/TransactionDoneScreen.kt#L247) — `OtherField(value = …)`
- Transaction detail sheet — [`TransactionDetailBottomSheet.kt:379`](app/src/main/java/com/vultisig/wallet/ui/screens/transaction/TransactionDetailBottomSheet.kt#L379) — `DetailValuePill(text = provider)`, even though the UI model already carries `providerLogo`

The inconsistency is user-visible: the history **card** shows a logo while the detail **sheet** for the same swap does not.

## Gaps in the resolver itself

1. **CowSwap is missing.** `getProviderLogo` maps `thorchain`, `maya`/`mayachain`, `jupiter`, `uni`, `1inch`, `kyber`/`kyberswap`, `li.fi`, and `swapkit*` — no `cowswap` branch, and there is no CowSwap drawable in `app/src/main/res/drawable/` (`jup.webp`, `kyberswap.png`, `lifi.png`, `oneinch.png`, `swapkit.xml` exist).
2. **No compile-time exhaustiveness.** It matches a lowercased `String` in a `when` with `else -> null`, so a provider added to the swap layer silently renders with no icon instead of failing the build. Windows solved this with a `Record<GeneralSwapProvider, string>` — the Android equivalent is keying off the `SwapProvider` enum rather than the display string.
3. **Route-hint labels.** `formatSwapKitProviderLabel` produces `SwapKit (CHAINFLIP)`; that survives only because of the `startsWith("swapkit")` special case. Strip the trailing ` (…)` hint once, before matching, so the rule holds for every provider instead of one.

## Scope

- [ ] Add a CowSwap drawable and map it in `getProviderLogo`
- [ ] Key the resolver off the `SwapProvider` enum so a new provider is a compile error, not a missing icon
- [ ] Strip the trailing route hint (`SwapKit (CHAINFLIP)` → `SwapKit`) before lookup, dropping the `startsWith` special case
- [ ] Render logo + name in the swap form `Provider` row (`SwapFeeBreakdown.kt`)
- [ ] Render logo + name on `TransactionDoneScreen`
- [ ] Render logo + name in `TransactionDetailBottomSheet` (reuse the `providerLogo` already on the UI model)
- [ ] Keep the form row's loading placeholder height unchanged — the icon must not shift layout when the quote resolves

### Out of scope
- How the provider **name** is computed (`providerLabel` / `SwapQuoteManager`) — unchanged
- Provider selection UI or alternative-route display
- Swap routing / quote logic

## Acceptance criteria
- [ ] All five surfaces — swap form, verify, transaction-done, history card, detail sheet — show the same logo + name for the same swap
- [ ] General providers (1Inch, LI.FI, KyberSwap, SwapKit, CowSwap, Jupiter) get an icon, not just the native swap chains (THORChain, MayaChain)
- [ ] `SwapKit (CHAINFLIP)` and similar route-hint labels still resolve their logo
- [ ] No layout shift in the form `Provider` row while the quote loads; long names still crop to one line

aminsato:
## Summary

Per [Figma node 45201-39285](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=45201-39285), every surface that names a swap **provider** shows its **logo** next to the name. Three of the five did not — so the same swap showed a logo on its history **card** and none in the detail **sheet** opened from that card.

The detail sheet was the starkest case: its UI model already carried `providerLogo`, and the value was dropped one line above the row that needed it.

Android side of [vultisig-windows#4572](https://github.com/vultisig/vultisig-windows/pull/4572).

## Changes

All five surfaces now route through a single `SwapProviderLabel` composable (16dp logo + 4dp gap + name, per Figma) instead of four near-copies that could drift apart again.

| Surface | File | Change |
|---|---|---|
| Swap form | `SwapFeeBreakdown.kt` | plain text → logo + name (`valueComposable` overload) |
| Transaction done | `TransactionDoneScreen.kt` | plain text → logo + name (new composable-value `OtherField` overload) |
| Detail sheet | `TransactionDetailBottomSheet.kt` | now passes the `providerLogo` the model already carried |
| Verify | `VerifySwapScreen.kt` | inline logo code replaced by the shared component |
| History card | `SwapTransactionCard.kt` | unchanged — already correct |

`getProviderLogo` is now keyed off the `SwapProvider` enum through the existing `swapProviderFromWireId`, so **a provider added to the swap layer is a compile error here** rather than a silently missing icon. The route hint is stripped once up front, generalising what `startsWith("swapkit")` did for one provider — `SwapKit (CHAINFLIP)` and any future hinted label resolve by the same rule.

## Before / After

Real Android renders (emulator, `PreviewActivity` + `adb screencap`). Identical preview code in both builds; only the fix differs.

**Swap form** — `Provider` row in the fee panel

| Before | After |
|--------|-------|
| before | after |

**Transaction done** — also shows the route hint resolving: the label reads `SwapKit (CHAINFLIP)` and still finds the SwapKit drawable

| Before | After |
|--------|-------|
| before | after |

**Detail sheet** — the pill that disagreed with its own history card

| Before | After |
|--------|-------|
| before | after |

**Verify** — the control. Included because this PR refactors it; before/after are identical, so the shared component introduced no regression on the surface that was already right.

| Before | After |
|--------|-------|
| before | after |

The swap-form `Provider` row sits at the same Y in both shots — the 16dp logo matches the caption line height and the loading skeleton, so nothing shifts as a quote resolves.

## Three deliberate deviations from the issue checklist

1. **CowSwap is not added.** It is not a provider on Android at all — absent from the `SwapProvider` enum and from `swapProviderFromWireId`; the only reference in the repo is a settlement-contract address in `KnownEvmContracts.kt`. A drawable plus mapping would be unreachable code. Supporting CowSwap as a provider is worth its own issue.
2. **`"uni" -> R.drawable.uni` removed as dead.** No swap provider named Uni exists — the Uniswap hits in the repo are EVM *router contract labels*, unrelated. `uni.webp` stays; `getCoinLogo` still uses it.
3. **`VerifyProviderRow` lost its 1dp logo border.** Figma shows no border on the provider logo and the history card's `ViaBadge` never had one. Acceptance criterion #1 asks all five surfaces to match, so this unifies rather than preserving the odd one out. This is the only change to a surface the issue called already-correct — see the Verify comparison above.

## Test plan

- `CoinLogoTest` — 6 new tests: every `SwapProvider` display label resolves a logo (the check no compiler can do for the string → enum step), display labels → drawable, case-insensitivity, wire aliases the history layer stores, route-hint stripping including on a non-SwapKit provider, unknown → null. 23/23 green.
- Full `:app:testDebugUnitTest` suite green.
- `:app:lintDebug` green.
- Four `PreviewActivity` cases added for manual re-checks: `provider_swap_form`, `provider_tx_done`, `provider_detail_sheet`, `provider_verify`.

Closes #5524

## Summary by CodeRabbit

- **New Features**
  - Added provider logos and names throughout swap verification, fee breakdowns, completed transactions, and transaction details.
  - Provider details now support optional logos with graceful text-only fallback.
  - Added previews for provider-based swap and transaction states.

- **Bug Fixes**
  - Improved provider recognition across display names, aliases, capitalization differences, and route hints.
  - Unsupported providers continue to display without an incorrect logo.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -53,6 +53,7 @@ import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.Coins
 import com.vultisig.wallet.data.models.getCoinLogo
+import com.vultisig.wallet.data.models.getProviderLogo
 import com.vultisig.wallet.data.models.logo
 import com.vultisig.wallet.data.models.payload.BlockChainSpecific
 import com.vultisig.wallet.data.models.payload.DAppMetadata
@@ -90,9 +91,11 @@ import com.vultisig.wallet.ui.models.TokenDetailUiModel
 import com.vultisig.wallet.ui.models.TokenSelectionUiModel
 import com.vultisig.wallet.ui.models.TokenUiModel
 import com.vultisig.wallet.ui.models.TransactionDetailsUiModel
+import com.vultisig.wallet.ui.models.TransactionHistoryItemUiModel
 import com.vultisig.wallet.ui.models.TransactionHistoryTab
 import com.vultisig.wallet.ui.models.TransactionHistoryUiState
 import com.vultisig.wallet.ui.models.TransactionScanStatus
+import com.vultisig.wallet.ui.models.TransactionStatusUiModel
 import com.vultisig.wallet.ui.models.VaultAccountsUiModel
 import com.vultisig.wallet.ui.models.VaultDetailUiModel
 import com.vultisig.wallet.ui.models.VerifyTransactionUiModel
@@ -181,10 +184,12 @@ import com.vultisig.wallet.ui.screens.swap.preview.AdvancedGasLimitPreview
 import com.vultisig.wallet.ui.screens.swap.preview.AdvancedMenuConfiguredPreview
 import com.vultisig.wallet.ui.screens.swap.preview.AdvancedMenuPreview
 import com.vultisig.wallet.ui.screens.swap.preview.AdvancedSlippagePreview
+import com.vultisig.wallet.ui.screens.swap.preview.SwapFormProviderPreview
 import com.vultisig.wallet.ui.screens.swap.preview.SwapFormQuoteLoadingPreview
 import com.vultisig.wallet.ui.screens.swap.preview.SwapToolbarPreview
 import com.vultisig.wallet.ui.screens.transaction.SendTxOverviewScreen
 import com.vultisig.wallet.ui.screens.transaction.SwapTransactionOverviewScreen
+import com.vultisig.wallet.ui.screens.transaction.TransactionDetailBottomSheet
 import com.vultisig.wallet.ui.screens.transaction.TransactionHistoryEmptyState
 import com.vultisig.wallet.ui.screens.transaction.TransactionHistoryScreen
 import com.vultisig.wallet.ui.screens.transaction.UiTransactionInfo
@@ -276,6 +281,10 @@ class PreviewActivity : ComponentActivity() {
                     "swap_error_before" -> SwapErrorBeforePreview()
                     "swap_error" -> SwapErrorPreview()
                     "swap_quote_loading" -> SwapFormQuoteLoadingPreview()
+                    "provider_swap_form" -> SwapFormProviderPreview()
+                    "provider_tx_done" -> SwapProviderTxDonePreview()
+                    "provider_detail_sheet" -> SwapProviderDetailSheetPreview()
+                    "provider_verify" -> SwapProviderVerifyPreview()
                     "swap_toolbar" -> SwapToolbarPreview()
                     "swap_advanced_locked" -> SwapAdvancedLockedPreview()
                     "custom_rpc_gate" -> CustomRpcGatePreview()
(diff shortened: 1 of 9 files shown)


johnnyluo (approved):
LGTM

PR #5528, opened 2026-08-06
Passcode app-lock 1/5: derive and hold the data key

aminsato:
Part 1 of 5 for #5343 (passcode app-lock + at-rest encryption, parity with Windows/extension).

Establishes the key hierarchy the rest of the feature builds on. **No UI and no behaviour change** — nothing calls into this yet.

## What

The passcode never encrypts vault data directly. A random 256-bit data key does, and only that key is wrapped under PBKDF2(passcode). Changing the passcode therefore rewraps 44 bytes instead of re-encrypting every keyshare, and the same data key can later be wrapped a second time under a biometric-gated Keystore key without keeping a second plaintext copy of the data. AES-256-GCM authenticates the wrap, so a failed unwrap *is* the passcode check — there is no separate verifier to drift out of sync.

## Choices worth reviewing

**210k PBKDF2 iterations, not the 600k used for exported backups.** A backup travels off the device and faces unlimited offline guessing, so it buys every iteration it can afford. The wrapped passcode key never leaves the device — it lives inside the AndroidKeyStore-encrypted preferences, so reaching it already means defeating hardware-backed key storage, and guessing it online is throttled. 210k is the OWASP floor for PBKDF2-HMAC-SHA256 and keeps unlock responsive on low-end devices, which matters once auto-lock can fire every minute.

**Throttling carries real weight at 5 digits.** 100k combinations is not many, so penalties escalate 30s → 1m → 5m → 15m → 1h from the fifth wrong attempt, persist across relaunches, and refuse to expire early when the system clock is wound backwards.

## Test plan

- `./gradlew :data:testDebugUnitTest` — 37 new tests covering the cipher round-trip, wrong-passcode and tampered-blob paths, the full lockout ladder including clock rollback, and concurrent unlock attempts on real threads.
- No user-visible surface to exercise; the feature becomes reachable in part 5.

## Summary by CodeRabbit

* **New Features**
  * Added passcode protection with setup, unlock, change, removal, and manual locking.
  * Added configurable auto-lock durations, including immediate locking and preset intervals.
  * Added escalating lockouts after repeated incorrect passcode attempts.
  * Preserved passcode credentials and lockout state across app relaunches.

* **Bug Fixes**
  * Added safeguards for invalid or tampered stored passcode data.

* **Tests**
  * Added coverage for encryption, validation, persistence, lockouts, and concurrent unlock attempts.


aminsato:

PR #5529, opened 2026-08-06
Passcode app-lock 2/5: encrypt vault keyshares at rest

aminsato:
Part 2 of 5 for #5343. Stacked on #5528 — **review that one first**; this PR's diff is against it.

Keyshares are now stored encrypted under the passcode-derived data key. Still no UI: nothing enables a passcode yet, so every existing install keeps behaving exactly as before.

## What

Encryption happens at the two points where `VaultRepository` already converts between the domain vault and its rows, so the rest of the app is untouched.

Stored shares carry a `vlpc1:` tag, which lets plaintext and ciphertext coexist in one table. That is what makes the feature survivable: users who never set a passcode keep plaintext rows forever, and a bulk conversion killed halfway through leaves a mix the next launch still reads correctly, then finishes on retry.

## Ordering, which is the part that matters

Both directions are ordered so no interruption can strand a share:

- **Enabling** writes the wrapped key first and encrypts second, so ciphertext never exists without a recorded key.
- **Disabling** decrypts everything into memory and validates it before writing, and only then drops the credentials — so a share that will not open aborts the whole operation while the passcode is still protecting the rest.
- **Changing** the passcode rewraps the data key and touches no shares at all.

## Reads and writes while locked

Reads drop the shares and log rather than throwing. The lock screen keeps the user away from signing, but background work such as notifications and balance sync still reads vaults and needs only public keys and addresses. The write path refuses to persist a keyshare while locked, so a locked read can never be echoed back as data loss.

## Test plan

- `./gradlew :data:testDebugUnitTest` — 14 new tests plus 8 added to `VaultRepositoryImplTest`, covering the tagged round-trip, plaintext passthrough for passcode-less users, half-migrated tables, the abort-before-write path on an undecryptable share, and the locked read/write behaviour above.
- End-to-end signing is exercised in part 5, where a passcode can actually be set.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/db/dao/VaultDao.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/db/dao/VaultDao.kt
@@ -42,6 +42,11 @@ interface VaultDao {
     @Query("SELECT coinId FROM disabledCoin WHERE vaultId = :vaultId")
     suspend fun loadDisabledCoinIds(vaultId: VaultId): List<String>
 
+    /**
+     * Every keyshare across every vault, used to re-key them when the passcode is set or removed.
+     */
+    @Query("SELECT * FROM keyShare") suspend fun loadAllKeyShares(): List<KeyShareEntity>
+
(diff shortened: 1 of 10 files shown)


aminsato:

PR #5530, opened 2026-08-06
Passcode app-lock 3/5: the App Locked entry screen

aminsato:
Part 3 of 5 for #5343. Stacked on #5529.

Builds the lock screen to the Figma frame [79816-126646](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=79816-126646). Nothing routes to it yet — the guard that shows it and the settings that enable a passcode land in parts 4 and 5.

## Design

The five cells, the card that holds them, the type ramp and the backdrop wash all come from the design; tokens map 1:1 (`backgrounds.surface1`, `border.light`, `primary.accent4`, `headings.largeTitle`, `supplementary.footnote`). The card tint and the 10% white hairlines have no token, so they are named constants.

Only the filled and error states are mine: the frame ships a Default state alone, so a filled cell reuses the design's own accent treatment and a rejected passcode borrows the existing `alerts.error` token rather than inventing a colour.

## Why a new component

Entry is masked. `VsCodeInputField` renders the digits it holds, which is right for a backup code mailed to the user and wrong for a passcode guarding their vaults. This keeps that component's mechanic — an invisible 1dp `BasicTextField` owning input and keyboard, visible cells drawn from its text — so the two behave the same under focus and IME.

Verification fires on the fifth digit rather than behind a button, which is what the design implies with no CTA on the frame.

## Fixes found on device

The second commit fixes three things caught by running this on hardware:

- **The error never appeared.** Rejecting a passcode empties the field, and that empty is itself an emission from the field's flow — the collector read it as the user starting over and cleared the error in the same frame it was set. Only a non-empty field counts as starting over now. The original tests passed through this because `snapshotFlow` does not re-emit in a JVM test until snapshot changes are published, and they only published their own edits; they now publish the view model's clear too, and fail without the fix.
- **The prompt did not move when the keypad opened**, leaving it low and half-covered. It now insets by the IME and centres in what is visible.
- **Locking did not reset the prompt.** The guard hosts this view model outside the nav graph, so it is activity-scoped and one instance serves every lock; a half-typed attempt was still in the cells next time.

## Test plan

- `./gradlew :app:testDebugUnitTest` — 8 view-model tests: auto-submit on the fifth digit, attempt countdown, error persistence, throttle countdown, reset on re-lock.
- Rendered on an emulator via the debug preview harness (`adb shell am start -n com.vultisig.wallet/com.vultisig.wallet.debug.PreviewActivity -e screen passcode_lock_error`), keyboard open and closed.
- Strings land in all ten locales, using each one's established wording for a passcode rather than reusing the wallet-password term.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -143,6 +143,9 @@ import com.vultisig.wallet.ui.models.swap.VerifySwapUiModel
 import com.vultisig.wallet.ui.models.swap.VultTierGateUiModel
 import com.vultisig.wallet.ui.models.toNetworkUiModel
 import com.vultisig.wallet.ui.models.v3.ReviewVaultDevicesUiState
+import com.vultisig.wallet.ui.models.passcode.PasscodeLockError
+import com.vultisig.wallet.ui.models.passcode.PasscodeLockUiModel
+import com.vultisig.wallet.ui.screens.passcode.PasscodeLockScreen
 import com.vultisig.wallet.ui.screens.ChainSelectionScreen
 import com.vultisig.wallet.ui.screens.TokenDetailScreen
 import com.vultisig.wallet.ui.screens.TokenSelectionScreen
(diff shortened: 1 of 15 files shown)


aminsato:

PR #5531, opened 2026-08-06
Passcode app-lock 4/5: gate the app and re-lock on idle

aminsato:
Part 4 of 5 for #5343. Stacked on #5530.

The commit that makes the feature real: the lock screen from part 3 now guards the app, and the app re-locks itself after the configured idle timeout. Setting a passcode is still only possible from part 5, so an install that has never set one is unaffected.

## Replacing, not stacking, the existing gate

A configured passcode takes over from the device-credential prompt rather than sitting on top of it. Two consecutive unlocks to open the app would be worse than either alone, and the passcode is the stronger of the two because it also holds the key to the encrypted keyshares. **With no passcode set, `BiometryAuthScreen` behaves exactly as before.**

While the persisted state is still loading the guard shows an opaque cover rather than nothing, otherwise a locked app flashes a frame or two of its contents before the lock screen arrives.

## Auto-lock

Runs off the **process** lifecycle, not an activity's, so rotating or re-entering the task is not mistaken for leaving the app.

It locks two ways because neither alone is enough: a timer fires in the background so the data key is not held a moment longer than the user asked for, and the elapsed time is re-checked on return to cover a process that was frozen or killed before that timer could run. The elapsed check reads `elapsedRealtime`, which keeps counting through deep sleep and cannot be wound backwards from Settings.

The chosen timeout is mirrored into a field rather than read when backgrounding, because reading the flow at that moment suspends and the process can be frozen before the read returns.

## Test plan

- `./gradlew :app:testDebugUnitTest` — 8 tests: immediate lock, timed lock firing and not firing early, cancellation on early return, the frozen-process path, a wound-back clock, and a timeout change taking effect without a restart.
- Manual, once part 5 lands: set a passcode, background the app past each timeout, confirm it locks; return within the timeout and confirm it does not.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/app/activity/MainActivity.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/MainActivity.kt
@@ -33,6 +33,7 @@ import androidx.compose.ui.graphics.toArgb
 import androidx.core.content.ContextCompat
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.ProcessLifecycleOwner
 import androidx.lifecycle.compose.LocalLifecycleOwner
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.lifecycle.lifecycleScope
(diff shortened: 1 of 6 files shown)


aminsato:

PR #5534, opened 2026-08-06
fix(keysign): default missing chain to Ethereum for KeyImport custom messages

### LINKED ISSUE #5527: [Fix] KeyImport custom messages still use the root keyshare when Android initiates [Keysign]
# [Fix] KeyImport custom messages still use the root keyshare when Android initiates [Keysign]

## Problem

PR #5525 (merged, from external contributor @arpicato) made `startKeysignKeyImport()` resolve the signing chain from `CustomMessagePayload.chain`, which fixed custom-message co-signing for sessions **iOS or the Windows extension** start. Two paths were left behind and still select the wrong keyshare.

A KeyImport vault holds N+2 MPC keyshares — root ECDSA at BIP32 path `m`, root EdDSA, and one per imported chain (`KeygenViewModel.kt:641-702`). Signing passes `chainPath = ""` (`DklsKeysign.kt` sets `chainPathSlice = null` for KeyImport), so there is no in-ceremony BIP32 derivation to correct a wrong pick — whichever share is selected is the one that signs.

1. **Android as initiator.** `SignMessageFormViewModel.kt:52-57` builds a `CustomMessagePayload` without `chain`. `resolveKeysignChain` (`KeysignViewModel.kt:107-114`) returns `null` when the field is absent, so `startKeysignKeyImport()` (`:538-539`) falls back to `vault.pubKeyECDSA` — the path-`m` share. An iOS or Windows co-signer defaults a missing chain to Ethereum and loads the per-chain share instead. The two devices then load shares of **different keys**, and DKLS aborts: the sign session folds `setup.keyshare().final_session_id` into its session hash and hard-errors on mismatch (`dkls23-rs/src/sign/dsg.rs:147-152, 265-267`).

2. **Fast-vault / VultiServer leg — confirmed broken server-side.** `KeysignFlowViewModel.buildJoinRequest` (`:372-386`) sends `derivePath = (_keysignPayload?.coin?.coinType ?: CoinType.ETHEREUM).derivationPath()` (`:380`) and `chain = _keysignPayload?.coin?.chain?.raw ?: ""` (`:383`). Both fall to their defaults for custom messages, so VultiServer is told `chain = ""`.

   Verified against the server source (vultisig/vultiserver @ `2a6390f`, `service/keysign_dkls.go:65-78`):

   ```go
   if !req.Mldsa && req.Chain != "" && localStateAccessor.Vault.LibType == v1.LibType_LIB_TYPE_KEYIMPORT {
       for _, chainInfo := range localStateAccessor.Vault.ChainPublicKeys {
           if strings.EqualFold(chainInfo.Chain, req.Chain) { publicKey = chainInfo.PublicKey; found = true; break }
       }
       if !found { return nil, fmt.Errorf("public key for chain %s not found in vault", req.Chain) }
       // when it is KeyImport vault , DerivePath need to be ignored
       req.DerivePath = "m"
   }
   ```

   The per-chain selection is **gated on `req.Chain != ""`**. With an empty chain the whole branch is skipped, so the server keeps `publicKey = req.PublicKey` — the vault **root** key Android sends at `:376` — and loads that keyshare (`GetLocalState(publicKey)`, `keysign_dkls.go:175`). Post-#5525 the Android device resolves the per-chain key for any custom message that carries a chain, so the two sides load different keyshares and the ceremony aborts on a key-id mismatch.

   For a custom message Android itself initiates, the payload carries no chain, so **both** sides currently fall back to the root share and the ceremony *does* complete — producing a signature under BIP32 path `m`, which recovers to no address the user holds. So the fast-vault symptom today is a silently useless signature rather than a hard failure.

   `req.DerivePath` is **not** part of this mechanism. The server is always a joiner: it builds its session from the initiator's setup message (`SignSessionFromSetup`, `keysign_dkls.go:217`), and Android passes `chainPathSlice = null` for KeyImport, so no BIP32 derivation happens in the ceremony on either side. `DerivePath` is used only for a non-empty guard (`:153`), two log lines, and a post-hoc `KeyshareDeriveChildPublicKey` + `ecdsa.Verify` whose mismatch only *logs* "Signature is invalid" (`:276-293`). Updating it is consistency/log-correctness, not a functional requirement — **`chain` is the field that matters.**

   Sending the chain is what fixes it: the server then loads the same per-chain keyshare the device uses.

   ⚠️ **Send a chain the vault actually holds, not a hardcoded `Ethereum`.** The server's lookup is exact-match only (`strings.EqualFold`, `:68`) with no derivation-path fallback and no root fallback — an unknown chain hard-errors with `public key for chain %s not found in vault` (`:74-76`). Because KeyImport chain selection is balance-driven and never forces Ethereum (`KeyImportChainsSetupViewModel.kt:135` `isSelected = hasBalance`; `continueWithSelection()` only rejects an empty list), a BTC-only or BSC-only KeyImport vault is creatable. Blindly sending `"Ethereum"` for such a vault makes the server abort inside its async worker — `/sign` has already returned 200 (`api/server.go:171` enqueues via asynq) and the client swallows join failures (`KeysignSessionCoordinator.startSession`, `catch (e: Exception) { Timber…e(e, "startSession failed") }`), so the user gets an indefinite stall with no error. Prefer Ethereum when `chainPublicKeys` has it, otherwise pick an entry sharing the EVM derivation path. iOS resolves whatever chain name is sent by exact match, so naming the held chain works on every platform.

   This is reachable and not exotic: `VerifySignMessageViewModel.kt:85-107` offers biometry/password fast-sign for custom messages, and 1 device (fast vault) is the default at vault creation (`ChooseDeviceCountViewModel.kt:31`). iOS (`FastVaultKeysignBootstrap`) and Windows (`FastKeysignServerStep.tsx`) both send the resolved chain already.

Affects KeyImport vaults only. Normal DKLS/GG20 vaults are unaffected — they have one share and derive in-ceremony.

## Expected Behavior

An Android-initiated custom message on a KeyImport vault completes against an iOS or Windows co-signer, using the same per-chain keyshare both peers pick, and the fast-vault server request names the same chain the local device signs with.

## Steps to Reproduce

1. Create a KeyImport vault (import a seed phrase) with 2 devices — one Android, one iOS or Windows extension.
2. On Android: Vault Settings → Sign Custom Message. Enter method `personal_sign` and any message.
3. Start the session and join from the iOS / Windows device.
4. The ceremony never completes. Android surfaces `fail to get keysign signature LIB_SIGNGEN_ERROR`; a Windows peer shows "Vault Share Mismatch".

## Solution

Make Android's missing-chain behaviour identical to the other two platforms rather than falling through to the root key. iOS defaults `targetChain` to `.ethereum` (`KeysignViewModel.swift:449`) and Windows defaults to `customMessageDefaultChain = Chain.Ethereum` (`core/ui/mpc/keysign/customMessage/chains.ts`). Do the same in `resolveKeysignChain`, and additionally set the field explicitly on our own initiator so the payload we put on the wire is self-describing for any joiner — iOS already does this (`SettingsCustomMessageView.swift:221` sets `chain: Chain.ethereum.name`). Then thread the resolved chain into the fast-vault request so the server leg agrees with the local device.

Note the root-key fallback in `Vault.getEcdsaSigningKey` should stay as-is — it is the BIP32 fallback root for non-imported chains and is load-bearing elsewhere.

### Must Do
- [ ] Set `chain` on the `CustomMessagePayload` built at `SignMessageFormViewModel.kt:52-57` — to a chain the vault actually holds (prefer `Chain.Ethereum` when present in `chainPublicKeys`, else an entry sharing the EVM derivation path). Do **not** hardcode Ethereum; see the server exact-match note above.
- [ ] Default a missing/unparseable chain to `Chain.Ethereum` in `resolveKeysignChain` (`KeysignViewModel.kt:107-114`) so it matches iOS and Windows.
- [ ] Use the resolved chain for `derivePath` (`KeysignFlowViewModel.kt:380`) and `chain` (`:383`) in `buildJoinRequest` when there is no `keysignPayload`.
- [ ] Add a unit test that a KeyImport vault with a `chainPublicKeys[Ethereum]` entry selects that key — not `pubKeyECDSA` — for a custom message with no chain set.

### Must NOT Do
- Don't change unrelated code
- Don't remove or alter the root-key fallback in `Vault.getEcdsaSigningKey` / `getEddsaSigningKey`
- Don't touch the transaction signing path — `keysignPayload.coin.chain` already takes precedence and is correct

### Out of Scope
- `startKeysignDkls()` hardcoding `DEFAULT_ETHEREUM_DERIVATION_PATH` for custom messages on non-KeyImport DKLS vaults (`KeysignViewModel.kt:529`) — same root cause class, separate issue.
- A Windows-style `customMessageSupportedChains` allowlist in `SigningHelper` — hygiene, separate issue.

## Acceptance Criteria
- [ ] A custom message initiated on Android for a KeyImport vault carries a non-empty `chain` on the wire
- [ ] `resolveKeysignChain(null, CustomMessagePayload())` returns `Chain.Ethereum`, and a payload chain still takes precedence over the default
- [ ] `buildJoinRequest` sends the resolved chain and its derivation path for custom messages instead of `""` / hardcoded Ethereum
- [ ] An Android-initiated custom message on a KeyImport vault co-signs successfully with an iOS or Windows peer

## References
- PR #5525 — fixed the joiner direction; merged, external contributor @arpicato
- iOS: `VultisigApp/Features/Keysign/ViewModels/KeysignViewModel.swift:449-458`
- Windows: `core/ui/mpc/keysign/action/KeysignActionProvider.tsx`, `core/ui/mpc/keysign/customMessage/chains.ts`

johnnyluo:
## Summary
- `SignMessageFormViewModel` now sets `chain = Chain.Ethereum.raw` on the outgoing `CustomMessagePayload`, matching iOS/Windows behavior.
- `resolveKeysignChain` defaults a missing/unparseable chain to `Chain.Ethereum` instead of returning `null`, so `startKeysignKeyImport()` picks the per-chain keyshare instead of falling back to the vault's root ECDSA key.
- `KeysignFlowViewModel.buildJoinRequest` now threads the resolved chain into both `derivePath` and `chain` for the fast-vault/VultiServer join request instead of sending `""`, so the server also selects the per-chain key and forces `DerivePath = "m"`.

## Issue
Closes #5527

## Test Plan
- [x] `./gradlew lint` passes
- [x] `./gradlew testDebugUnitTest --tests "com.vultisig.wallet.ui.models.keysign.ResolveKeysignChainTest"` passes (added cases: missing chain defaults to Ethereum, explicit chain still takes precedence)
- [ ] Manual: Android-initiated custom message on a KeyImport vault co-signs successfully with an iOS or Windows peer

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved custom message signing by correctly resolving the selected blockchain.
  * Defaults custom message signing to Ethereum when no valid chain is specified.
  * Preserves explicitly selected chains, including Solana.
  * Improved ECDSA chain selection based on vault configuration.

* **Tests**
  * Added coverage for default, explicit, invalid, and vault-specific signing chains.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
@@ -382,7 +382,7 @@ constructor(
             derivePath = (chain?.coinType ?: CoinType.ETHEREUM).derivationPath(),
             isEcdsa = tssKeysignType == TssKeyType.ECDSA,
             password = password,
-            chain = chain?.raw ?: "",
+            chain = resolveJoinRequestChainRaw(_keysignPayload, customMessagePayload),
             mldsa = tssKeysignType == TssKeyType.MLDSA,
         )
     }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
@@ -19,6 +19,7 @@ import com.vultisig.wallet.data.models.TransactionHistoryData
 import com.vultisig.wallet.data.models.TssKeyType
 import com.vultisig.wallet.data.models.TssKeysignType
 import com.vultisig.wallet.data.models.Vault
+import com.vultisig.wallet.data.models.coinType
 import com.vultisig.wallet.data.models.isSecureVault
 import com.vultisig.wallet.data.models.payload.KeysignPayload
 import com.vultisig.wallet.data.models.tokenLogoRes
@@ -372,15 +373,16 @@ constructor(
     private fun buildJoinRequest(vault: Vault): JoinKeysignRequestJson? {
         val password = password
         if (password.isNullOrBlank()) return null
+        val chain = resolveKeysignChain(_keysignPayload, customMessagePayload)
         return JoinKeysignRequestJson(
             publicKeyEcdsa = vault.pubKeyECDSA,
             messages = messagesToSign,
             sessionId = _sessionID,
             hexEncryptionKey = _encryptionKeyHex,
-            derivePath = (_keysignPayload?.coin?.coinType ?: CoinType.ETHEREUM).derivationPath(),
+            derivePath = (chain?.coinType ?: CoinType.ETHEREUM).derivationPath(),
             isEcdsa = tssKeysignType == TssKeyType.ECDSA,
             password = password,
-            chain = _keysignPayload?.coin?.chain?.raw ?: "",
+            chain = chain?.raw ?: "",
             mldsa = tssKeysignType == TssKeyType.MLDSA,
         )
     }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
@@ -25,6 +25,7 @@ import com.vultisig.wallet.data.models.TokenStandard
 import com.vultisig.wallet.data.models.TransactionHistoryData
 import com.vultisig.wallet.data.models.TssKeyType
 import com.vultisig.wallet.data.models.Vault
+import com.vultisig.wallet.data.models.coinType
 import com.vultisig.wallet.data.models.getEcdsaSigningKey
 import com.vultisig.wallet.data.models.getEddsaSigningKey
 import com.vultisig.wallet.data.models.getSwapProviderId
@@ -110,8 +111,21 @@ internal fun resolveKeysignChain(
 ): Chain? =
     keysignPayload?.coin?.chain
         ?: customMessagePayload?.let { payload ->
-            runCatching { Chain.fromRaw(payload.chain.orEmpty()) }.getOrDefault(Chain.Ethereum)
+            val raw = payload.chain
+            if (raw.isNullOrBlank()) {
+                Chain.Ethereum
+            } else {
+                runCatching { Chain.fromRaw(raw) }
+                    .onFailure { Timber.w("Unknown custom message chain %s", raw) }
+                    .getOrNull()
+            }
         }
+
+/** The `chain` field sent to VultiServer when joining a keysign session. */
+internal fun resolveJoinRequestChainRaw(
+    keysignPayload: KeysignPayload?,
+    customMessagePayload: CustomMessagePayload?,
+): String = resolveKeysignChain(keysignPayload, customMessagePayload)?.raw ?: ""
 
 /** UI state for an in-progress or completed keysign session. */
 internal sealed class KeysignState {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
@@ -109,8 +109,8 @@ internal fun resolveKeysignChain(
     customMessagePayload: CustomMessagePayload?,
 ): Chain? =
     keysignPayload?.coin?.chain
-        ?: customMessagePayload?.chain?.let { raw ->
-            runCatching { Chain.fromRaw(raw) }.getOrNull()
+        ?: customMessagePayload?.let { payload ->
+            runCatching { Chain.fromRaw(payload.chain.orEmpty()) }.getOrDefault(Chain.Ethereum)
         }
 
 /** UI state for an in-progress or completed keysign session. */
--- a/app/src/main/java/com/vultisig/wallet/ui/models/sign/SignMessageFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/sign/SignMessageFormViewModel.kt
@@ -5,6 +5,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.VaultId
+import com.vultisig.wallet.data.models.resolveEcdsaChain
 import com.vultisig.wallet.data.repositories.CustomMessagePayloadDto
 import com.vultisig.wallet.data.repositories.CustomMessagePayloadRepo
 import com.vultisig.wallet.data.repositories.VaultRepository
@@ -55,7 +56,7 @@ constructor(
                             message = messageFieldState.text.toString(),
                             vaultPublicKeyEcdsa = vault.pubKeyECDSA,
                             vaultLocalPartyId = vault.localPartyID,
-                            chain = Chain.Ethereum.raw,
+                            chain = vault.resolveEcdsaChain(Chain.Ethereum).raw,
                         ),
                 )
             customMessagePayloadRepo.add(payload)
--- a/app/src/main/java/com/vultisig/wallet/ui/models/sign/SignMessageFormViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/sign/SignMessageFormViewModel.kt
@@ -3,6 +3,7 @@ package com.vultisig.wallet.ui.models.sign
 import androidx.compose.foundation.text.input.TextFieldState
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.VaultId
 import com.vultisig.wallet.data.repositories.CustomMessagePayloadDto
 import com.vultisig.wallet.data.repositories.CustomMessagePayloadRepo
@@ -54,6 +55,7 @@ constructor(
                             message = messageFieldState.text.toString(),
                             vaultPublicKeyEcdsa = vault.pubKeyECDSA,
                             vaultLocalPartyId = vault.localPartyID,
+                            chain = Chain.Ethereum.raw,
                         ),
                 )
             customMessagePayloadRepo.add(payload)
--- a/app/src/test/java/com/vultisig/wallet/ui/models/keysign/ResolveJoinRequestChainFieldsTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/keysign/ResolveJoinRequestChainFieldsTest.kt
@@ -1,0 +1,41 @@
+package com.vultisig.wallet.ui.models.keysign
+
+import com.vultisig.wallet.data.models.Chain
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import vultisig.keysign.v1.CustomMessagePayload
+
+class ResolveJoinRequestChainFieldsTest {
+    @Test
+    fun `custom message with no chain set sends Ethereum as the raw chain`() {
+        assertEquals(
+            Chain.Ethereum.raw,
+            resolveJoinRequestChainRaw(
+                keysignPayload = null,
+                customMessagePayload = CustomMessagePayload(),
+            ),
+        )
+    }
+
+    @Test
+    fun `custom message chain is threaded into the raw chain sent to the server`() {
+        assertEquals(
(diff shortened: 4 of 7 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
```diff
@@ -372,15 +373,16 @@ constructor(
...
             publicKeyEcdsa = vault.pubKeyECDSA,
             messages = messagesToSign,
             sessionId = _sessionID,
             hexEncryptionKey = _encryptionKeyHex,
-            derivePath = (_keysignPayload?.coin?.coinType ?: CoinType.ETHEREUM).derivationPath(),
+            derivePath = (chain?.coinType ?: CoinType.ETHEREUM).derivationPath(),
             isEcdsa = tssKeysignType == TssKeyType.ECDSA,
             password = password,
-            chain = _keysignPayload?.coin?.chain?.raw ?: "",
+            chain = chain?.raw ?: "",
```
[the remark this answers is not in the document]
johnnyluo: chain should not be empty , use "m" vs "" , technically should no difference

app/src/main/java/com/vultisig/wallet/ui/models/sign/SignMessageFormViewModel.kt
```diff
@@ -54,6 +55,7 @@ constructor(
                             message = messageFieldState.text.toString(),
                             vaultPublicKeyEcdsa = vault.pubKeyECDSA,
                             vaultLocalPartyId = vault.localPartyID,
+                            chain = Chain.Ethereum.raw,
```
rkokhatskyi: Hardcoding `Chain.Ethereum.raw` breaks any KeyImport vault without an Ethereum entry. That shape is reachable since import selection is balance driven (`KeyImportChainsSetupViewModel.kt:135`) and `continueWithSelection()` only rejects an empty list. VultiServer looks up the chain by exact match with no fallback and errors with `public key for chain %s not found in vault` (`keysign_dkls.go:68-76`). The user never sees it because `/sign` already returned 200 (`api/server.go:171`) and `KeysignSessionCoordinator.startSession` swallows join failures, so the session just stalls. Windows has the same exact match and throws on a miss (`KeysignActionProvider.tsx:45-56`). Set `chain` to a chain the vault actually holds. Prefer `Chain.Ethereum` when it is in `chainPublicKeys` and otherwise any entry on the same EVM derivation path. iOS already resolves path siblings after vultisig/vultisig-ios#5071 so naming the held chain works everywhere. Server gap filed at https://github.com/vultisig/vultiserver/issues/159

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
rkokhatskyi: Please cover `buildJoinRequest` by tests. Nothing asserts which chain it sends for a custom message payload. Reverting `chain = chain?.raw ?: ""` to `_keysignPayload?.coin?.chain?.raw ?: ""` keeps the suite green.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -109,8 +109,8 @@ internal fun resolveKeysignChain(
     customMessagePayload: CustomMessagePayload?,
 ): Chain? =
     keysignPayload?.coin?.chain
-        ?: customMessagePayload?.chain?.let { raw ->
-            runCatching { Chain.fromRaw(raw) }.getOrNull()
+        ?: customMessagePayload?.let { payload ->
+            runCatching { Chain.fromRaw(payload.chain.orEmpty()) }.getOrDefault(Chain.Ethereum)
```
aminsato: `getOrDefault(Chain.Ethereum)` collapses two cases that should stay distinct:

1. `chain` is unset — legacy payloads from before the field existed. Ethereum is the right default, and that's what this PR is for.
2. `chain` is set to a value this build can't parse — a chain iOS/extension already supports but Android hasn't added to the enum yet, or a malformed value. `Chain.fromRaw` uses `entries.first { ... }`, so it throws and we silently substitute Ethereum.

Case 2 is the problem: `startKeysignKeyImport` then signs with `vault.getEcdsaSigningKey(Chain.Ethereum)`, so this device returns a signature under the wrong keyshare. The initiator doesn't see a chain mismatch, it just sees an invalid signature further down the flow — with no log to point back here.

Scoping the default to the unset case keeps the fix and preserves the loud-ish failure for unknown chains:

```kotlin
?: customMessagePayload?.let { payload ->
    val raw = payload.chain
    if (raw.isNullOrBlank()) {
        Chain.Ethereum
    } else {
        runCatching { Chain.fromRaw(raw) }
            .onFailure { Timber.w("Unknown custom message chain %s", raw) }
            .getOrNull()
    }
}
```

This also keeps it consistent with `SigningHelper.getKeysignMessages`, which still resolves the same field with `getOrNull()` — as written, an unrecognized chain gets an EVM keccak256 digest there but an Ethereum keyshare here, and the two will drift as soon as either fallback changes.

Worth a third test alongside the two added ones: `CustomMessagePayload(chain = "NotAChain")`, pinning whichever behavior you settle on.

app/src/main/java/com/vultisig/wallet/ui/models/sign/SignMessageFormViewModel.kt
johnnyluo: Good catch. Added `Vault.resolveEcdsaChain(preferred)` in `Vault.kt`: prefers `Chain.Ethereum` when the vault holds it, otherwise picks any ECDSA `chainPublicKeys` entry sharing Ethereum's derivation path (BSC/Polygon/Arbitrum/...), falling back to `Ethereum` only when the vault has no EVM-family key at all. `SignMessageFormViewModel` now calls `vault.resolveEcdsaChain(Chain.Ethereum).raw` instead of hardcoding Ethereum. See d442debb8.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignFlowViewModel.kt
johnnyluo: Extracted `resolveJoinRequestChainRaw(keysignPayload, customMessagePayload)` in `KeysignViewModel.kt` — it's the exact `chain` computation `buildJoinRequest` now calls, and it's unit tested in `ResolveJoinRequestChainFieldsTest`. Reverting the wiring in `buildJoinRequest` back to the old field now fails that test. (`derivePath` itself still calls into native Trust Wallet Core and can't be unit tested here, same as the rest of this file's chain-resolution helpers.) See d442debb8.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
johnnyluo: Implemented exactly as suggested: `resolveKeysignChain` now only defaults to Ethereum when the chain is unset/blank; an unparseable chain returns null and logs `Timber.w("Unknown custom message chain %s", raw)` instead of silently substituting Ethereum. Added the `CustomMessagePayload(chain = "NotAChain")` test case you suggested, asserting it resolves to null. See d442debb8.

app/src/test/java/com/vultisig/wallet/ui/models/keysign/ResolveJoinRequestChainFieldsTest.kt
```diff
@@ -0,0 +1,41 @@
+package com.vultisig.wallet.ui.models.keysign
+
+import com.vultisig.wallet.data.models.Chain
+import kotlin.test.Test
+import kotlin.test.assertEquals
```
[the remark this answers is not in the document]
johnnyluo: Good catch — switched to `org.junit.jupiter.api.Test` and `io.kotest.matchers.shouldBe`, matching this project's test conventions. See d147fc957.

aminsato (approved):
LGTM

PR #5542, opened 2026-08-08
fix(defi): price non-EVM position tokens instead of caching a zero

### LINKED ISSUE #5496: DeFi fiat pricing: contract-price fallback is dead code on non-EVM chains; cache rows exist only for vault-held coins

aminsato:
Fixes #5496

## Problem

DeFi position tokens on THORChain and Maya (sTCY, yRUNE, yTCY, sRUJI, Maya pool assets) resolved to $0.00. Two gaps compounded.

**The contract-price fallback could not answer for any non-EVM chain — but it answered anyway.** CoinGecko has no asset-platform id for those chains and LI.FI has no chain id, and the unresolved LI.FI price was mapped to `ZERO`. A zero is indistinguishable from a real "worth nothing" quote, so it was both returned to the caller *and* written into Room. On chains with no working contract source, that poisoned row was the only price the token would ever have.

**The price cache is only ever refreshed for coins the vault holds**, and position tokens usually aren't held. A user who staked sTCY or yTCY without separately enabling the token had no cache row and, by the above, no live fallback either.

## Changes

- A missing CoinGecko asset platform is now an ordinary miss (`emptyMap()`) rather than an `error()` swallowed into one, and the LI.FI leg is skipped on chains it doesn't index — neither fabricates a zero
- THORChain contract lookups route through the sources that do cover them: index NAV for the bRUNE / ybRUNE / sTCY / NAMI receipts, pool TOR price for everything else. Maya prices off pool depth
- A zero is never returned or persisted as if it were a quote
- The staking screen refreshes prices for its own position coins instead of depending on vault membership
- A pool asset resolves to its curated coin first, so it picks up that coin's CoinGecko id

That last one covers a case the issue didn't name but which is the same bug: both `createFiatValueFromPoolAsset` and the THORChain LP path went straight to the contract route, which *cannot* price a native asset — BTC has no contract address to look up. This is what the "Maya pool assets" symptom actually was.

The third bullet of the issue (key sTCY off TCY's rate) had already landed on `main` via `priceProviderID = "tcy"` plus the provider-id step in `DefiFiatValueCalculator`. This PR only extracts the shared denom/contract constants so `fetchThorContractPrices` and the new contract route can't drift apart.

## Test plan

- `:data:testDebugUnitTest` and `:app:testDebugUnitTest` pass, including 5 new repository tests and 3 new calculator tests covering the NAV route, the pool fallback, the never-cache-a-zero rule, and the LI.FI gate
- `:app:lintDebug` and `:data:lintDebug` pass
- Two existing tests asserted the old behaviour and were updated: `ThorchainDefiPositionsViewModelTest` explicitly asserted the BTC LP leg prices to zero (now $2), and the calculator's contract-address fallback case now resolves through curated USDC

No before/after screenshots: this is a pricing-logic fix with no layout or design component. Reproducing it visually needs a real vault with a staked position plus live network, and a PreviewActivity render would only show fiat constructed in the preview — which would be fake evidence rather than proof.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DeFi fiat value calculations for native assets, pool assets, staking tokens, and receipt tokens.
  * Added more reliable pricing across supported chains, including THORChain and MayaChain.
  * Added fallback pricing when primary THORChain data is unavailable.
  * Improved asset matching with curated coin data and case-insensitive metadata.
  * Prevented invalid, missing, or zero-price results from being cached.
  * Preserved cached prices when refreshes fail.
  * Improved LP totals and unsupported-chain handling.
* **Tests**
  * Expanded coverage for contract matching, fallback pricing, caching, and unsupported-chain scenarios.


app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -842,6 +842,22 @@ constructor(
...
+                    withContext(ioDispatcher) {
+                        try {
+                            tokenPriceRepository.refresh(loadedPositions.map { it.coin })
+                        } catch (t: Throwable) {
+                            if (t is CancellationException) throw t
+                            // Pricing is cosmetic here: a failed refresh leaves the previous
+                            // cached price in place rather than blocking the cards.
+                            Timber.e(t, "Failed to refresh staking position prices")
+                        }
+                    }
```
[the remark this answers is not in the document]
aminsato: Not applying this one. `safeLaunch` is a replacement for `launch` — it starts a coroutine. There's no `launch` here: this is a sequential call inside an existing `collect`, and the refresh has to complete before the fiat strings are formatted below it.

Using it would mean `safeLaunch { ... }.join()`, which adds a child coroutine and a join to get semantics identical to what's already there — the `try`/`catch` rethrows `CancellationException` and logs everything else, which is exactly `safeLaunch`'s body.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -192,14 +193,115 @@ constructor(
...
+                    ?.values
+                    ?.firstOrNull()
+
+        // Only a real quote is worth keeping. A miss used to arrive here as a zero and be written
+        // straight into Room, so every later cache read served it back as a confident "this token
+        // is worth nothing" — and on the chains with no working contract source that poisoned row
+        // was the only price the token would ever have.
+        if (price == null || price.signum() == 0) return BigDecimal.ZERO
+
+        savePrices(mapOf(contractAddress to mapOf(currency to price)), currency)
```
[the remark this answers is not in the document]
aminsato: Fixed in 7bf123f — this was a real gap. All three sites now reject anything not strictly positive:

- the direct contract path guards `signum() <= 0` rather than `== 0`
- a Lifi quote of literal `0` is dropped alongside the null case
- `mayaPoolPrice` rejects an empty CACAO side (it divides out to a price of 0) and a ratio that rounds to zero at 8dp

The last one predates this branch, but persisting a zero is precisely what this PR is removing, so it belongs here.

aminsato:
## Found the live root cause from a device logcat

A logcat from a vault holding all four position tokens showed:

| token | shown | should be |
|---|---|---|
| yRUNE | **$0.00** | ~$0.6276 |
| yTCY | **$0.00** | ~$0.1081 |
| sTCY | $0.1267 | ~$0.1327 |
| RUNE | $0.4508 | ✅ |

Every NAV call was failing:

```
GET https://thorchain.ibs.team/api/cosmwasm/wasm/v1/contract/{c}/smart/eyJzdGF0dXMiOiB7fX0=
500 {"code":2,"message":"codespace sdk code 26: invalid height: context did not
     contain latest block height in either check state or finalize block state (27275817)"}
```

`getThorchainTokenPriceByContract` was the **only** call in `ThorChainApi` pinned to `thorchain.ibs.team` — every other THORChain request goes through the shared liquify gateway. That node has fallen ~65k blocks behind (chain head was 27341296 in the same log). I confirmed it's persistent, not a blip, and that the gateway answers the identical query correctly:

```
$ curl .../thorchain.ibs.team/api/cosmwasm/.../smart/eyJzdGF0dXMiOiB7fX0=
{"code":2,"message":"...invalid height...(27275817)"}

$ curl .../gateway.liquify.com/chain/thorchain_api/cosmwasm/.../smart/eyJzdGF0dXMiOiB7fX0=
{"data":{"liquid_bond_shares":"735579287087061","liquid_bond_size":"771186485285645",...}}
```

Since that query is the *only* price source for the index receipts, one stale node prices the whole class at zero. Fixed in 8023cf2 by asking the gateway first and keeping ibs.team as the fallback.

Worth calling out: **sTCY was hiding this.** It didn't render $0.00 because its `priceProviderID = "tcy"` fallback covered for the dead NAV call — but that yields bare TCY parity, silently dropping the ~4.8% NAV premium (NAV is currently 1.0484). A degradation with no visible failure.

Also contained the NAV failure in the THORChain contract route so a dead NAV host costs a denom its preferred price rather than its turn at the pool route.

Two new tests pin the host order and the fallback.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1178,18 +1196,35 @@ constructor(
...
+            Coins.allResolvable.firstOrNull { coin ->
+                coin.chain == chain &&
+                    coin.ticker.equals(ticker, ignoreCase = true) &&
+                    (contractAddress.isEmpty() ||
+                        coin.contractAddress.equals(contractAddress, ignoreCase = true))
+            }
+        if (curated != null) return priceFor(curated, currency)
+
+        return try {
             tokenPriceRepository.getPriceByContactAddress(chain.id, contractAddress)
```
rkokhatskyi: priceForPoolAsset's fallback for an asset not in the catalogue (`ThorchainDefiPositionsViewModel.kt:1220-1221`) calls getPriceByContactAddress directly, skipping the getCachedPrice check its DefiFiatValueCalculator sibling still runs at `DefiFiatValueCalculator.kt:58-61`. This LP leg can render $0.00 on a contract miss while Maya's equivalent leg succeeds from cache

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -192,14 +193,126 @@ constructor(
...
+
+        // Only a real quote is worth keeping. A miss used to arrive here as a zero and be written
+        // straight into Room, so every later cache read served it back as a confident "this token
+        // is worth nothing" — and on the chains with no working contract source that poisoned row
+        // was the only price the token would ever have. Anything not strictly positive is a miss:
+        // a negative quote is nonsense no source should produce, and persisting one would be worse
+        // than the zero this replaces.
+        if (price == null || price.signum() <= 0) return BigDecimal.ZERO
+
+        savePrices(mapOf(contractAddress to mapOf(currency to price)), currency)
```
rkokhatskyi: getPriceByContactAddress caches a resolved price under the raw contract address here, but every reader, e.g. `DefiFiatValueCalculator.kt:66`, queries by coin.id, never a bare contract address. The write is unreachable, so a position card without a vault-held coin repeats the live lookup on every reload instead of hitting cache

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1135,7 +1151,9 @@ constructor(
 
         val assetPrice =
             assetCoin?.let { priceFor(it, currency) }
```
rkokhatskyi, line 1156: This vault coin match at `ThorchainDefiPositionsViewModel.kt:1147` compares ticker with ignoreCase true, but Maya's identical lookup at `MayachainDefiPositionsViewModel.kt:822` uses exact equals. A held token whose ticker case differs from the pool string, Kujira CW20 metadata isn't uppercased like EVM tokens are, misses on Maya and shows zero where Thorchain resolves it

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -290,6 +410,14 @@ constructor(
 
     private suspend fun fetchTetherPrice() = getPriceByPriceProviderId(TETHER_PRICE_PROVIDER_ID)
 
+    /**
+     * Currency-per-USD, for converting the USD-quoted sources (pool TOR prices, index NAV) into the
+     * app currency. Skips the USDT round trip when the app currency already is USD.
+     */
+    private suspend fun tetherPriceFor(currency: String): BigDecimal =
```
rkokhatskyi, line 502: thorContractPrice multiplies a valid priceUsd by tetherPriceFor(currency), which returns BigDecimal.ZERO on any CoinGecko miss instead of failing, see the return at `TokenPriceRepository.kt:257`. A transient tether price failure zeroes an otherwise correct non USD price, and the signum check can't tell that apart from a genuine miss

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -842,6 +842,22 @@ constructor(
...
                     val loadedPositions = defaultPositions.filter { it.coin.id in coinsToLoad }
 
+                    // sTCY, yTCY and yRUNE are position tokens, not wallet balances: unless the
+                    // user separately enabled them the vault does not hold them, and the periodic
+                    // price refresh only ever covers vault coins. Refresh them here — nothing else
+                    // will — or their cards fall through to a contract lookup with no cache row
+                    // behind it and render $0.00.
+                    withContext(ioDispatcher) {
+                        try {
+                            tokenPriceRepository.refresh(loadedPositions.map { it.coin })
```
rkokhatskyi: sTCY shares TCY's priceProviderID tcy, so the provider batch write at `TokenPriceRepository.kt:138`, first reached for sTCY via this new refresh call, can cache TCY's raw price before the NAV correction in fetchThorContractPrices runs. A failed correction leaves sTCY stuck at plain TCY parity

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -384,13 +493,42 @@ constructor(
...
 
+    /**
+     * A Maya asset's price in [currency], derived from its pool's depth: the pool quotes the asset
+     * in CACAO, and CACAO's own price comes from the cache (it is a CoinGecko-priced native coin).
+     * Null when the pool is empty or CACAO has no price yet, so the caller drops the token rather
+     * than recording a zero.
+     */
+    private suspend fun mayaPoolPrice(token: Coin, currency: String): BigDecimal? {
+        val userCurrency = appCurrencyRepository.currency.first()
+        val cacaoPrice = getCachedPrice(Coins.MayaChain.CACAO.id, userCurrency) ?: return null
```
rkokhatskyi: mayaPoolPrice only reads CACAO's price from cache and gives up on a miss, unlike runePriceUsd and tcyPriceUsd, which live fetch, compare `TokenPriceRepository.kt:504` to `:647`. The new single request mayaContractPrice caller has no guarantee CACAO was refreshed first, so a cold cache renders the leg at $0.00 where iOS and Windows recover live

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -192,14 +193,126 @@ constructor(
...
+                thorReceiptPriceUsd(contractAddress, currency)
+            } catch (e: Exception) {
+                if (e is kotlinx.coroutines.CancellationException) throw e
+                Timber.w(e, "NAV price unavailable for %s, trying the pool", contractAddress)
+                null
+            }
+
+        val priceUsd = receiptPrice ?: thorPoolPriceUsd(contractAddress) ?: return null
+        if (priceUsd.signum() <= 0) return null
+        return priceUsd * tetherPriceFor(currency)
```
rkokhatskyi: tetherPriceFor(currency) uses its argument only to pick skip or fetch; when it fetches, it rereads appCurrencyRepository.currency itself at `TokenPriceRepository.kt:319`, ignoring the parameter. If currency changes between the caller's capture at `TokenPriceRepository.kt:198` and this call, thorContractPrice multiplies a USD price by another currency's tether rate and persists it under the stale label

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -384,13 +493,42 @@ constructor(
...
     }
 
+    /**
+     * A Maya asset's price in [currency], derived from its pool's depth: the pool quotes the asset
+     * in CACAO, and CACAO's own price comes from the cache (it is a CoinGecko-priced native coin).
+     * Null when the pool is empty or CACAO has no price yet, so the caller drops the token rather
+     * than recording a zero.
+     */
+    private suspend fun mayaPoolPrice(token: Coin, currency: String): BigDecimal? {
+        val userCurrency = appCurrencyRepository.currency.first()
```
rkokhatskyi: mayaPoolPrice takes a currency parameter but ignores it, rereading appCurrencyRepository.currency itself here. Both callers still save and return the result under their own earlier captured currency string, e.g. `TokenPriceRepository.kt:214`, so a price computed under one currency can get persisted and displayed as another if the app currency changes mid call

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -305,10 +433,7 @@ constructor(
                     return@supervisorScope
                 }
 
-            val tickerUsd = AppCurrency.USD.ticker.lowercase()
-            val tetherPrice =
-                if (currency.equals(tickerUsd, ignoreCase = true)) 1.toBigDecimal()
-                else fetchTetherPrice()
+            val tetherPrice = tetherPriceFor(currency)
```
rkokhatskyi, line 525: fetchThorPoolPrices never got the positive price guard the rest of this PR added. A pool's assetTorPrice of zero survives the guard that only checks null at `TokenPriceRepository.kt:452` and reaches savePrices at `:463`, which only filters empty maps, so a cached zero permanently blocks a later real price

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -242,8 +355,12 @@ constructor(
                     coinGeckoContractsPrice.keys.any { key -> key.equals(address, false) }
                 }
 
-            notInCoinGeckoTokens.takeIf { it.isNotEmpty() }
-                ?: return@coroutineScope coinGeckoContractsPrice
+            // LI.FI only indexes EVM chains, so asking it about a THORChain/Maya/Cosmos contract
```
rkokhatskyi, line 418: This comment claims the caller's non EVM route handles THORChain, Maya and Cosmos contracts, but nativeChainContractPrice's when block at `TokenPriceRepository.kt:231-234` only covers Chain.ThorChain and Chain.MayaChain, defaulting to null for every other chain. Kujira, Gaia, Osmosis and Terra contracts have no such route despite the claim

app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
```diff
@@ -45,12 +44,29 @@ constructor(private val tokenPriceRepository: TokenPriceRepository) {
         currency: AppCurrency,
     ): FiatValue =
         convert(amount, currency) {
-            tokenPriceRepository.getCachedPrice(
-                tokenId = "$ticker-${chain.id}",
-                appCurrency = currency,
-            ) ?: tokenPriceRepository.getPriceByContactAddress(chain.id, contractAddress)
+            val curated =
+                Coins.allResolvable.firstOrNull { coin ->
```
rkokhatskyi: priceForPoolAsset at `ThorchainDefiPositionsViewModel.kt:1211-1217` reimplements the identical curated coin match already added here, even though that file already injects and calls fiatValueCalculator elsewhere. Both copies rely on Coins.kt never giving two coins on the same chain the same ticker, an invariant nothing enforces or tests

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -842,6 +842,22 @@ constructor(
                 .collect { defaultPositions ->
                     val loadedPositions = defaultPositions.filter { it.coin.id in coinsToLoad }
```
rkokhatskyi, line 843: This refresh call fires even when loadedPositions is empty, reachable whenever a vault deselects sTCY, yRUNE and yTCY but keeps RUJI or TCY selected. `TokenPriceRepository.kt:128` has no empty list guard, so it issues a real CoinGecko proxy request with an empty ids parameter on every Staking tab load for no benefit

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed in 0ac7d64. The catalogue-miss path now reads `getCachedPrice("\$ticker-\${chain.id}")` before paying for a live contract lookup, same order as the `DefiFiatValueCalculator` sibling.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 0ac7d64. Now resolves the contract to its curated coin and caches under `coin.id`. A contract the catalogue does not carry has no id to key on, so it is simply not cached rather than written somewhere unreachable. Test: `a resolved contract price is cached under the coin id readers query by`.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed in 0ac7d64 — Maya now matches the ticker with `ignoreCase = true`, same as the THORChain tab.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 0ac7d64. `thorContractPrice` now treats a non-positive FX rate as a failed lookup and returns null instead of multiplying through to a zero. Test: `a non-USD price is left unresolved when the FX rate cannot be fetched`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 0ac7d64. Added a `cacaoPrice` helper with the cache-then-live shape `runePriceUsd`/`tcyPriceUsd` already use, so a cold cache no longer zeroes the leg.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 0ac7d64. `mayaPoolPrice` now takes an `AppCurrency` and uses it. Worth noting the conversion needed `currency.uppercase()` — `AppCurrency.fromTicker` matches the enum ticker exactly, and `currency` travels lowercased through this class, so the naive version resolved to null every time.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 0ac7d64 — `fetchThorPoolPrices` now drops a pool quoting zero rather than letting it reach `savePrices`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 0ac7d64. The comment now says the Cosmos chains have no native route, so a contract there stays unpriced — but unpriced and unrecorded, rather than the zero it used to write down as a real quote.

app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
aminsato: Fixed in 0ac7d64 — extracted to `Coins.findCurated` (and `findCuratedByContract` for the contract-only case), so the match exists once.

On the invariant: you are right that nothing enforces unique chain+ticker. I have documented the matching rules on the helper rather than added a guard, since a duplicate would be a catalogue bug worth failing loudly on, not something a price lookup should paper over. Happy to add a test asserting uniqueness across `Coins.allResolvable` if you want it pinned.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed in 0ac7d64 — guarded with `loadedPositions.isNotEmpty()`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Documented rather than restructured, in 0ac7d64.

You are right that the parameter only selects skip-vs-fetch and the fetch resolves the app currency live. I left the behaviour alone because it predates this PR and every caller refetches wholesale on a currency change (`observeCurrencyChanges`), which is what keeps the window closed — but the parameter reads as though it pins the rate, so the KDoc now says explicitly that it does not, and that it is not safe to widen the gap between capture and use on the strength of it. The `mayaPoolPrice` instance you flagged separately was a real one and is fixed properly.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Not changing this one — the ordering is the graceful degradation, not a defect.

You are right about the sequence: the provider batch writes TCY parity for sTCY, then `fetchThorContractPrices` corrects it to NAV x TCY. If the correction fails, sTCY sits at parity.

Parity is the better failure mode though. The alternative — withhold the write until NAV lands — shows $0.00 whenever the NAV host is down, and that is strictly worse than a price about 5% light. This is not hypothetical: the device logcat on this PR shows the NAV host stale and returning `invalid height` for every contract, and sTCY was the only one of the four still showing a number, precisely because of this ordering. yRUNE and yTCY have no provider id and rendered $0.00.

Where I agree with you is that the degradation is silent. NAV is currently 1.0484, so parity understates by ~4.6% with nothing on screen saying so. I have fixed the cause (8023cf2 moves the NAV query off the single stale node), but if you want the symptom surfaced too — a staleness marker on the card when the correction fails — that is worth its own issue rather than widening this one.

data/src/main/kotlin/com/vultisig/wallet/data/api/ThorChainApi.kt
```diff
@@ -615,15 +615,36 @@ constructor(
...
-        val url = "$IBS_TEAM_URL/api/cosmwasm/wasm/v1/contract/$contract/smart/$STATUS_QUERY_BASE64"
-        return httpClient
+        val path = "cosmwasm/wasm/v1/contract/$contract/smart/$STATUS_QUERY_BASE64"
+        return try {
+            fetchContractStatus("$THORNODE_BASE/$path")
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.w(e, "Gateway status query failed for %s, falling back to ibs.team", contract)
+            fetchContractStatus("$IBS_TEAM_URL/api/$path")
```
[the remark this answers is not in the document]
aminsato: Fixed in 5be37b9. Both URLs now go through `appendPathSegments(..., encodeSlash = true)`, matching what `CoinGeckoApiImpl` already does for its contract lookups.

Worth noting this is more than theoretical here: THORChain denoms routinely carry slashes (`x/staking-x/brune` is a curated one), and the contract reaching this call is derived from a denom. Added the regression test you asked for — `a contract containing a slash stays inside its own path segment` feeds `thor1abc/../../evil` and asserts the slashes come back percent-encoded inside one segment and the path still starts with the contract route.

data/src/test/kotlin/com/vultisig/wallet/data/api/ThorChainApiContractStatusHostTest.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @Test
+    fun `the shared gateway answers the status query and ibs team is left alone`() = runBlocking {
+        val response =
+            api(HttpStatusCode.OK, HttpStatusCode.InternalServerError)
+                .getThorchainTokenPriceByContract("thor1contract")
+
+        assertEquals("200", response.data.liquidBondSize)
+        assertEquals(listOf(false), requestedHosts.map { it.contains("ibs.team") })
```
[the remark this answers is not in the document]
aminsato: Fixed in 5be37b9 — both tests now assert the exact ordered host list (`[gateway]` and `[gateway, ibs.team]`), so gateway-first routing is enforced rather than merely permitted.

data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt
```diff
@@ -4479,4 +4479,40 @@ object Coins {
...
+     * checksummed EVM addresses disagree on case; an empty one matches on chain and ticker alone,
+     * which is how native assets arrive. Ticker matching is likewise case-insensitive: chain
+     * metadata is not uniformly uppercased the way the EVM token lists are.
+     */
+    fun findCurated(chain: Chain, ticker: String, contractAddress: String): Coin? =
+        allResolvable.firstOrNull { coin ->
+            coin.chain == chain &&
+                coin.ticker.equals(ticker, ignoreCase = true) &&
+                (contractAddress.isEmpty() ||
+                    coin.contractAddress.equals(contractAddress, ignoreCase = true))
```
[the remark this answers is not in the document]
aminsato: Not applying — the premise is sound but the conclusion does not hold here, and the change would cost more than it buys.

For case-insensitive matching to resolve the *wrong* coin, the catalogue would need two entries on one chain whose contract addresses differ only in case. I checked rather than assumed: `CoinsCuratedLookupTest` in 5be37b9 groups `Coins.allResolvable` by chain + lowercased contract (and by chain + lowercased ticker) and asserts no group has more than one member. Both pass, so no query can resolve ambiguously today, and the test now fails the build if anyone adds an entry that would make it possible.

Going chain-aware would also introduce a failure mode rather than remove one. The case-insensitivity is load-bearing for EVM: THORChain and Maya pools name the contract in upper case (`ETH.USDC-0XA0B86991…`) while the catalogue carries the checksummed mixed-case form, and that mismatch is exactly what this PR fixes. Requiring exact matching on Solana/Sui buys nothing measurable — no collision exists — while turning any future casing difference in a pool string into a silent $0.00, which is the bug being fixed.

Happy to revisit if a real collision ever shows up; the test will be what surfaces it.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -323,8 +475,11 @@ constructor(
...
                         }
 
-                        // If still no price found, skip this token
-                        if (priceUsd == null) {
+                        // No price, or a pool quoting zero: skip the token either way. savePrices
+                        // only filters empty maps, so a zero would land in Room and every later
+                        // cache read would serve it as a real "worth nothing" — permanently, since
+                        // nothing invalidates a cached price.
+                        if (priceUsd == null || priceUsd.signum() <= 0) {
                             return@mapNotNull null
```
rkokhatskyi, line 552: fetchThorPoolPrices guards priceUsd here but never checks the product against tetherPrice at `TokenPriceRepository.kt:489`. tetherPriceFor returns ZERO on a CoinGecko miss, so a real pool price like RUJI's gets multiplied to zero and cached as a confident quote, the exact case thorContractPrice was just fixed for

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -344,37 +499,17 @@ constructor(
...
 
-            val cacaoToken =
-                tokenList.find { it.chain == Chain.MayaChain && it.isNativeToken }
-                    ?: return@supervisorScope
-
-            val userCurrency = appCurrencyRepository.currency.first()
-            val cacaoPrice = getCachedPrice(cacaoToken.id, userCurrency) ?: return@supervisorScope
-            if (cacaoPrice <= BigDecimal.ZERO) return@supervisorScope
-
+            val appCurrency = appCurrencyRepository.currency.first()
```
rkokhatskyi: fetchMayaPoolPrices rereads the currency fresh here to price the token, but savePrices persists the result under the currency captured much earlier at `TokenPriceRepository.kt:89`. A switch mid refresh caches a price computed in the new currency labeled with the old one, corrupting that row until a later refresh happens to match

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -192,14 +193,144 @@ constructor(
...
+                null
+            }
+
+        val priceUsd = receiptPrice ?: thorPoolPriceUsd(contractAddress) ?: return null
+        if (priceUsd.signum() <= 0) return null
+
+        // fetchTetherPrice answers a CoinGecko miss with ZERO, so multiplying blind would turn a
+        // perfectly good USD price into a $0.00 that reads exactly like "we have no price" — and
+        // for a non-USD user, only ever on the FX leg. Treat a missing rate as a failed lookup.
+        val fxRate = tetherPriceFor(currency)
```
rkokhatskyi, line 298: tetherPriceFor's non USD branch rereads the currency independently, ignoring what getPriceByContactAddress captured at `TokenPriceRepository.kt:198`. Since price() checks cache before calling this path again, a mislabeled row from a currency switch mid fetch gets replayed indefinitely for an LP leg no other refresh ever revisits

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -384,13 +519,63 @@ constructor(
         }
     }
 
+    /**
```
rkokhatskyi: These two stacked doc comments both sit above cacaoPrice. The first, about pool depth and an empty pool returning null, actually describes mayaPoolPrice at `TokenPriceRepository.kt:551`, left behind when the function split. mayaPoolPrice itself now carries no doc comment of its own

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -842,6 +842,25 @@ constructor(
...
 
+                    // sTCY, yTCY and yRUNE are position tokens, not wallet balances: unless the
+                    // user separately enabled them the vault does not hold them, and the periodic
+                    // price refresh only ever covers vault coins. Refresh them here — nothing else
+                    // will — or their cards fall through to a contract lookup with no cache row
+                    // behind it and render $0.00. Skipped when nothing loaded: refresh has no
+                    // empty-list guard of its own and would spend a live request on no ids.
+                    if (loadedPositions.isNotEmpty()) {
+                        withContext(ioDispatcher) {
+                            try {
```
rkokhatskyi, line 853: sTCY still shares TCY's priceProviderID, so refresh(), triggered by this call, can cache TCY's raw price under sTCY's row at `TokenPriceRepository.kt:138` before the NAV correction runs. Raised in the prior review and still open, if that correction fails sTCY stays priced at bare TCY parity

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 23b925c. `fetchThorPoolPrices` now rejects a non-positive rate up front and abandons the batch rather than multiplying every pool price through it.

You were right that this was the same case, and it was not the only other one — the LI.FI leg in `fetchPricesWithContractAddress` and the `usdPrice` branch of `refresh` had the identical shape, so both got the same guard. Test: `a THORChain pool price is not cached when the FX rate cannot be fetched`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 23b925c. `fetchMayaPoolPrices` now derives its `AppCurrency` from the caller's `currency` (the same `fromTicker(currency.uppercase())` conversion `mayaContractPrice` uses) instead of rereading, so the price cannot be computed in one currency and filed under another.

Test: `a Maya pool price is computed in the currency it is persisted under` — it drives the currency flow to emit USD first and EUR after, so a reread is observable, then asserts CACAO is never priced in EUR while the row is labelled `usd`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 23b925c. There is now a `fetchTetherPrice(currency)` that asks CoinGecko for the rate in the currency it was handed, so `tetherPriceFor` honours its parameter all the way down.

The reread came from delegating to `getPriceByPriceProviderId`, which resolves the app currency itself — correct for that public method, wrong underneath a caller that already captured one. `getPriceByPriceProviderId` is unchanged. I had documented this window in the previous round instead of closing it; your point about `price()` serving the cache first is what makes that insufficient, since the mislabeled row is never revisited.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in 23b925c — the pool-depth doc moved back above `mayaPoolPrice`, and `cacaoPrice` keeps only its own.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed in 23b925c. `refresh` now filters the NAV-priced denoms out of both generic routes, leaving `fetchThorContractPrices` as the only thing that writes those rows — so a failed NAV correction leaves the last-known price alone instead of exposing an uncorrected TCY-parity row.

The predicate is extracted as `isNavPricedDenom` and shared with `fetchThorContractPrices`, so the exclusion list and the claim list cannot drift apart.

I fixed it in the pricing route rather than changing `sTCY.priceProviderID` in `Coins.kt`: the shared id is not wrong as catalogue data (sTCY really is TCY-denominated), it is only wrong as an instruction to the provider batch. Test: `sTCY does not inherit TCY's price through their shared provider id`, which fails the NAV call and asserts nothing is written.

aminsato:
Pushed 23b925c addressing the review.

@coderabbitai the two nitpicks from the last two runs are done in the same commit:

- `ThorChainApiContractStatusHostTest` now uses `runTest` — all three tests, not only the new one, so the file is consistent.
- Added `a pool quoting zero is a miss, not a valuation worth caching` to `TokenPriceRepositoryImplTest`. I kept it as its own test rather than folding it into the empty-pool one, since a pool that answers `0` and a pool that does not exist are different paths and both are worth pinning.

Each new test was verified to fail with its fix reverted, so none of them pass vacuously.

app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
```diff
@@ -45,12 +44,23 @@ constructor(private val tokenPriceRepository: TokenPriceRepository) {
...
+                tokenPriceRepository.getCachedPrice(
+                    tokenId = "$ticker-${chain.id}",
+                    appCurrency = currency,
+                ) ?: tokenPriceRepository.getPriceByContactAddress(chain.id, contractAddress)
+            }
         }
 
+    private suspend fun price(coin: Coin, currency: AppCurrency): BigDecimal =
+        tokenPriceRepository.getCachedPrice(tokenId = coin.id, appCurrency = currency)
+            ?: priceByProviderId(coin)
```
rkokhatskyi: price() falls to priceByProviderId here before the NAV correct contract route. sTCY shares TCY's priceProviderID tcy, `Coins.kt:4093`, so on any cache miss this returns TCY's raw price mislabeled as sTCY's, the same duplicate exists at `ThorchainDefiPositionsViewModel.kt:1188-1191`

app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
```diff
@@ -45,12 +44,23 @@ constructor(private val tokenPriceRepository: TokenPriceRepository) {
...
+            ?: tokenPriceRepository.getPriceByContactAddress(coin.chain.id, coin.contractAddress)
```
rkokhatskyi: Neither priceByProviderId nor getPriceByContactAddress here take the currency this function already captured, both self determine it internally in `TokenPriceRepository.kt:350` and `:212`. If currency changes before either resolves, the returned number and its label can disagree, and getPriceByContactAddress persists that mismatch to cache

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
```diff
@@ -406,22 +646,15 @@ constructor(
...
                 val tokenIds = matchingTokens.map { it.id }
 
-                val tickerUsd = AppCurrency.USD.ticker.lowercase()
-                val tetherPrice =
-                    if (currency.equals(tickerUsd, ignoreCase = true)) {
-                        BigDecimal.ONE
-                    } else {
-                        fetchTetherPrice()
-                    }
+                val tetherPrice = tetherPriceFor(currency)
```
rkokhatskyi, line 679: fetchThorContractPrices multiplies by tetherPrice here with no signum guard, unlike every sibling call site of tetherPriceFor. When the FX rate is unavailable it still fires a live NAV or RUNE price call per token, wasted work the batch's existing zero filter only discards after the fact

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
aminsato: Fixed in d8be900. `price()` now skips the provider-id route for a NAV-priced denom and goes straight to the contract route, which is the only one that applies the NAV correction.

The list of those denoms moved to `Coins.isNavPricedDenom` so the pricing repository, the refresh batch and the position screens all read the same one — it was already duplicated between `refresh` and `fetchThorContractPrices`, and a fourth copy in the UI layer would have been the thing that drifts. Pinned by a catalogue test that asserts the exact membership (sTCY, bRUNE, ybRUNE, yTCY, yRUNE) and by a calculator test that fails if sTCY ever reaches `getPriceByPriceProviderId`.

The duplicate at `ThorchainDefiPositionsViewModel` is gone: `priceFor`/`priceForPoolAsset` now delegate to `DefiFiatValueCalculator.priceOf`/`priceOfPoolAsset`, keeping only their own catch so a failed leg still renders at zero instead of collapsing the card.

app/src/main/java/com/vultisig/wallet/ui/models/defi/DefiFiatValueCalculator.kt
aminsato: Fixed in d8be900. Both `getPriceByContactAddress` and `getPriceByPriceProviderId` now take the caller's `appCurrency`, and a shared `currencyTicker` helper only rereads `appCurrencyRepository.currency` when the caller has none of its own — the same treatment `tetherPriceFor` and `mayaPoolPrice` already got. The calculator passes the currency it captured down every route, so the quote and the label `convert` stamps on it can no longer come from two different currencies, and the contract route can no longer persist that mismatch.

The parameter is nullable with a default so the callers that genuinely have no currency of their own (`SearchEvmTokenUseCase`, `SearchSolTokenUseCase`, `SearchSuiTokenUseCase`) keep resolving it live rather than having one threaded through them for no reason.

Covered by a repository test asserting a lookup passed EUR quotes in EUR and caches under `eur` while the app currency is USD.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TokenPriceRepository.kt
aminsato: Fixed in d8be900. `fetchThorContractPrices` now gives up on a non-positive rate right after `tetherPriceFor`, before the shared RUNE fetch and the per-token NAV calls, matching the guard the other three call sites have. Nothing is persisted, so the last-known prices stay in place.

Test asserts that with no FX rate available `getThorchainTokenPriceByContract` is never called and nothing reaches the DAO.

johnnyluo (approved):
LGTM

PR #5544, opened 2026-08-09
feat(design-system): add the corner-radius token scale

aminsato:
Part of #5503 — **phase 1 of 4.** This PR defines the token layer only. It does not close the issue.

`ui/theme/` has `Colors`, `Typography` and `Theme` and **no radius concept at all**, so every rounded surface in the app spells its geometry out as a numeric literal: 384 of them across 23 distinct values, plus six different ways to say "fully round". This adds the missing layer and migrates a small, deliberately-chosen slice of call sites to prove the API works end to end.

**Zero visual change is the acceptance criterion for this phase**, and it is measured rather than asserted — see below.

Ports vultisig-ios#5040. The scale is the same one; the API is not, because Compose is not SwiftUI — the differences are called out where they matter.

---

## The scale

| Token | dp | Surface class | Source |
|---|---|---|---|
| `xs` | 4 | progress tracks, skeleton bars | inferred |
| `sm` | 8 | small inline tags | inferred |
| `md` | 12 | list rows, compact containers | Figma — asset list row |
| `lg` | 16 | input fields, inner icon tiles | Figma — input, inner tile |
| `xl` | 24 | cards, banners, list containers | Figma — banner |
| `pill` | — | chips, search fields, capsules | Figma — 50 / 77 / 99 |

**There is deliberately no sheet token, and the scale stops at 24.** The two sheet frames in the design file measure 38 and 28, but both are stock design-kit components pasted into the file rather than authored Vultisig surfaces — 38 is Apple's iOS 26 sheet, 28 is Material 3's default. Neither is ours, and adding a step for them would encode a platform default as a design decision. A comment at the scale declaration records this so nobody "completes" the scale later.

### Where these came from

Read off rendered frames in Figma file `puB2fsVpPrBx3Sup7gaa3v`, section **"Main View (Mobile) - iOS 26"** — these were re-measured for this PR rather than taken from the iOS port:

- banner `68734:97487` → `rounded-[24px]`
- inner icon tile `68734:97489` and input field `26557:15834` → `rounded-[16px]`
- asset list row `51141:203613` → `rounded-tl-[12px] rounded-tr-[12px]`
- vault-type chip `rounded-[99px]`, coin logo `rounded-[50px]`, glass-close `rounded-[77px]`

Two caveats worth stating plainly:

1. **Radii are not Figma variables.** `get_variable_defs` on that section returns colours, fonts, a stroke width and `dimensions/3 = 8` — there is no authored radius scale to import. These values were measured off rendered frames.
2. **`4` and `8` are lower-confidence than the rest** — inferred from existing app usage and that `dimensions/3`, not read off a frame. Any outlier found during phase 2 should go back to the designer rather than be rounded to the nearest token.

Values are shared with iOS and Windows — identical numbers, units per platform.

---

## The token API

```kotlin
Theme.v2.radius.xl              // the token — reads like Theme.v2.colors.*
Modifier.clip(Theme.v2.radius.xl)
Modifier.background(color, Theme.v2.radius.xl)
Card(shape = Theme.v2.radius.xl)

Theme.v2.radius.md.size         // Dp, only where a Shape will not fit (Canvas, insets)
Theme.v2.radius.lg.shape.copy(...)   // derive partial corners from the token
```

**The token *is* a `Shape`.** That is the one design decision everything else follows from: a call site names neither a radius nor a corner geometry, so changing the app's corner shape is one line inside `Radii` instead of another 384-site sweep. It also means the token drops into `shape =`, `clip`, `background` and `border` with no wrapper and no `.shape` noise at the point of use.

`Radius.Fixed.size` stays for the cases a `Shape` cannot reach — a `Canvas` corner radius, an inset computed off the surface. `Radius.shape` is a `CornerBasedShape` rather than a bare `Shape` so partial-corner surfaces can `copy(...)` from the token instead of restating its number. Both have a call site in this PR; nothing ships unused.

### `pill` is `CircleShape`, and that is better than the iOS equivalent

iOS needs a clamping *bound* — a 100,000pt constant, measured to still produce a capsule at 5120×2880 — because SwiftUI clamps a large radius down to half the smaller dimension. Compose's `CircleShape` is `RoundedCornerShape(percent = 50)`, which is defined *relative to the surface*. It cannot stop being round no matter how large the surface gets, so the Android token needs no magic number at all.

The app's fully-round idioms (50, 70, 77, 88, 99, 100 dp and `percent = 50`) all collapse onto it.

---

## `CornerType` is replaced by the token type

`V2Container` had its own two-case corner vocabulary — `CornerType.Circular` and `CornerType.RoundedCornerShape(size = 12.dp)` — across 17 call sites. It is deleted; `V2Container` now takes `radius: Radius` directly, so there is one corner concept in the app rather than two. `Circular` → `pill` is identical by construction (both produce `CircleShape`), and each numeric site maps to its own step.

---

## Migrated sites — every one pixel-neutral

26 token references across 21 files, chosen so each site's current literal **already equals** the token's value. Nothing was re-valued.

| File | Before | After |
|---|---|---|
| `components/v2/containers/V2Container.kt` | `CornerType` | `radius: Radius`, defaulting to `md` |
| 17 `V2Container` call sites | `CornerType.Circular` ×10, `12.dp` ×3, `24.dp` ×2, `8.dp` ×2 | `pill` / `md` / `xl` / `sm` |
| `components/SignTonDisplayView.kt` | `RoundedCornerShape(4.dp)` | `xs` |
| `components/library/form/Form.kt` | `RoundedCornerShape(4.dp)` | `xs` |
| `components/errors/ErrorView.kt` ×2 | `RoundedCornerShape(16.dp)` | `lg` |
| `components/dapp/DappRequestBanner.kt` | `RoundedCornerShape(16.dp)` | `lg` |
| `components/v2/fastselection/…/FastSelectionModalContent.kt` | `clip(RoundedCornerShape(24.dp))` | `clip(xl)` |
| `screens/transaction/TransactionHistoryScreen.kt` ×2 | `CornerRadius(12.dp.toPx())` + `RoundedCornerShape(12.dp)` | `md.size` + `md` |
| `screens/cosmosstaking/CosmosWithdrawRewardsScreen.kt` ×2 | `RoundedCornerShape(topStart = 16.dp, topEnd = 16.dp)` | `lg.shape.copy(...)` |

Between them these cover every scale step the app uses, and every form a radius takes in Compose: `Card(shape =)`, `Modifier.clip`, `Modifier.background`, `Modifier.border`, a `DrawScope` corner radius, and the partial-corner case.

### Two of them fixed latent drift

Both are the bug class that no lint rule will ever catch, because it is invisible in the declaration and only shows at the call site:

- **`TransactionHistoryEmptyState`** drew its border on a canvas at `CornerRadius(12.dp.toPx())` and its background at `RoundedCornerShape(12.dp)` — two independent literals for one surface. The stroke would have stopped tracing the card the first time either moved. Both now read one token.
- **`CosmosWithdrawRewardsScreen`**'s docked footer restated `topStart = 16.dp, topEnd = 16.dp` in its clip and again in its border. Now derived once from `lg`.

### Deliberately not migrated

- **Sheets.** Out of scope entirely, not just out of phase 1 — the scale has no step for them.
- **Text inputs.** Figma says 16; the app's fields are at 12. Real drift, needs a designer call rather than a silent rounding.
- **Fully-round literals on unbounded surfaces.** See the clamping test below — converting one of those is a visual change, not a rename.

---

## How pixel-neutrality is proven

**1. Geometry, not eyeballs.** `RadiusTokenTest` (9 cases) asserts each token's `createOutline` is identical to the outline of the exact literal expression it replaced, pins the scale values and their ordering, and checks the derived partial-corner shape against the literal it replaces.

It also **measures the clamping bound** rather than assuming it, which is the rule phase 2 needs: a fully-round literal only collapses to a capsule while the surface's smaller dimension stays under twice the literal. `RoundedCornerShape(99.dp)` on a 320×240 surface draws 99, not a capsule — so migrating a fully-round literal is only safe on a surface that is provably bounded. That is why this PR converts `CornerType.Circular` (already `CircleShape`, identical by construction) and leaves the numeric fully-round literals for phase 2.

Two of these tests failed on their first run and both were worth having:
- `Outline` equality is float-exact, and a 70dp literal clamps to 22.000002 rather than 22 — so the pill assertions compare drawn radii with a sub-pixel tolerance instead of comparing outline objects.
- An outline with unequal corners builds an `android.graphics.Path`, which a plain JVM test cannot touch — so the partial-corner case reads the four corner sizes off the shape instead.

**2. A pixel diff of the whole preview surface.** Both APKs were built, installed on the same emulator, and driven through **all 128 `PreviewActivity` screens** with the status bar frozen via SystemUI demo mode. The two capture sets were then compared pixel by pixel.

A raw before/after comparison is not conclusive on its own, because some previews render non-deterministically — animated heroes, async coin logos, the soft keyboard. So the branch was captured **twice**, giving a measured noise floor to compare the result against:

| | |
|---|---|
| Preview screens compared | 128 |
| Unstable across two runs of **identical** code | 11 |
| Deterministic screens | 117 |
| Of those, differing `main` → this branch | **0** |
| Differences not explained by that instability | **0** |

Every screen that differs between `main` and this branch also differs between two runs of this branch against itself. The 11 unstable ones are animation phase (`keysign_signing_lunc`, `review_vault_devices`' pulsing ring), async image loads (`blockaid_hero_done_*`, `sui_token_selection_*`, `swap_toolbar`) and the IME (`edit_folder`'s diff is entirely the keyboard suggestion strip and text cursor).

All 128 were compared rather than only the screens I touched, because the failure mode that matters for a token layer is not the site you edited — it is a shared component you did not realise fed a dozen other screens.

---

## What phases 2–4 cover

**Phase 2 — migrate by surface class**, one PR per batch, ordered by descending confidence. The two largest populations in the app are 12 (×138 including `size =` form) and 16 (×74), and **most of them are not containers**, so blanket-converting either would be visually destructive. The iOS phase-2 ruling — *containers take 24, anything nested inside one takes an explicitly smaller step, and the two levels must differ* — ports directly and is derived from call sites, not declarations.

**Phase 3 — the lint rule** banning new numeric radius literals, which is the only mechanism that makes the issue's "drift can't re-accumulate" actually true. It can only land after the outliers are reconciled.

**Phase 4 — corner geometry.** Compose has no continuous-corner primitive, so the iOS finding (the SDK silently moved every surface to `.continuous`) has no Android equivalent — Android corners are circular arcs and always have been. If the design language wants squircles, that becomes a custom `CornerBasedShape` swapped in at one line inside `Radii`, which is exactly what this layer exists to make possible.

---

## Gate

- `:app:testDebugUnitTest` — full suite, **BUILD SUCCESSFUL**, including 9 new `RadiusTokenTest` cases.
- `:app:lintDebug` — **BUILD SUCCESSFUL**.
- `ktfmt` applied.
- The 128-preview pixel diff above.

## Summary by CodeRabbit

- **UI Improvements**
  - Standardized corner rounding across cards, buttons, search bars, banners, dialogs, snackbars, and empty states.
  - Updated surfaces to use consistent theme-defined radius sizes, including small, medium, large, extra-large, and pill shapes.
  - Improved consistency for partially rounded sections, such as withdrawal summaries and bordered error areas.
  - Refined visual alignment across wallet, token, staking, transaction, and home screens.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/components/SignTonDisplayView.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/SignTonDisplayView.kt
@@ -194,7 +194,7 @@ private fun BadgeText(text: String) {
         modifier =
             Modifier.background(
                     color = Theme.v2.colors.variables.bordersLight,
-                    shape = RoundedCornerShape(4.dp),
+                    shape = Theme.v2.radius.xs,
                 )
                 .padding(horizontal = 6.dp, vertical = 2.dp),
         color = Theme.v2.colors.text.primary,
--- a/app/src/main/java/com/vultisig/wallet/ui/components/VaultCeil.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/VaultCeil.kt
@@ -24,7 +24,6 @@ import androidx.compose.ui.unit.dp
 import com.vultisig.wallet.R
 import com.vultisig.wallet.ui.components.library.form.FormCard
 import com.vultisig.wallet.ui.components.v2.containers.ContainerType
-import com.vultisig.wallet.ui.components.v2.containers.CornerType
 import com.vultisig.wallet.ui.components.v2.containers.V2Container
 import com.vultisig.wallet.ui.components.v2.icons.VaultIcon
 import com.vultisig.wallet.ui.theme.Theme
@@ -62,10 +61,7 @@ internal fun VaultCeil(
 
     val (logoBackground, logoBorder) = logo
     FormCard {
-        V2Container(
-            type = containerType,
-            cornerType = CornerType.RoundedCornerShape(size = 12.dp),
-        ) {
+        V2Container(type = containerType, radius = Theme.v2.radius.md) {
             Row(
                 verticalAlignment = Alignment.CenterVertically,
                 modifier =
--- a/app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/dapp/DappRequestBanner.kt
@@ -9,7 +9,6 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.CircleShape
-import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.Icon
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
@@ -63,7 +62,7 @@ internal fun DappRequestBanner(metadata: DAppMetadata, modifier: Modifier = Modi
                 .fillMaxWidth()
                 .background(
                     color = Theme.v2.colors.backgrounds.surface2,
-                    shape = RoundedCornerShape(16.dp),
+                    shape = Theme.v2.radius.lg,
                 )
                 .padding(20.dp)
                 .semantics(mergeDescendants = true) { contentDescription = announcement },
--- a/app/src/main/java/com/vultisig/wallet/ui/components/errors/ErrorView.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/errors/ErrorView.kt
@@ -12,7 +12,6 @@ import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
@@ -119,12 +118,12 @@ private fun ShowExactErrorRow(modifier: Modifier = Modifier, onClick: () -> Unit
                 .fillMaxWidth()
                 .background(
                     color = Theme.v2.colors.backgrounds.surface1,
-                    shape = RoundedCornerShape(16.dp),
+                    shape = Theme.v2.radius.lg,
                 )
                 .border(
                     width = 1.dp,
                     color = Theme.v2.colors.border.light,
-                    shape = RoundedCornerShape(16.dp),
+                    shape = Theme.v2.radius.lg,
                 )
                 .clickable(onClick = onClick)
                 .padding(20.dp),
--- a/app/src/main/java/com/vultisig/wallet/ui/components/library/form/Form.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/library/form/Form.kt
@@ -119,7 +119,7 @@ internal fun TokenCard(
                             .border(
                                 width = 1.dp,
                                 color = Theme.v2.colors.neutrals.n50,
-                                shape = RoundedCornerShape(4.dp),
+                                shape = Theme.v2.radius.xs,
                             )
                             .align(BottomEnd),
                 )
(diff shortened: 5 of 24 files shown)


johnnyluo (approved):
LGTM

PR #5545, opened 2026-08-09
fix(defi): key the Manage Positions selection per chain, not just per vault

### LINKED ISSUE #5533: DeFi: THORChain and Maya share one Manage Positions key per vault, so either screen overwrites the other
Split out of the review on #5523 at @rkokhatskyi's request — pre-existing, independent of that PR.

## What happens

`DefiPositionsRepository` stores the Manage Positions selection under one key per vault:

```kotlin
private fun selectedPositionsKey(vaultId: String) =
    stringSetPreferencesKey("selected_positions_$vaultId")
```

Both `ThorchainDefiPositionsViewModel` and `MayachainDefiPositionsViewModel` read and write that
same key, so for a vault holding both chains each screen's save clobbers the other's selection.

## Steps to reproduce

1. Open a vault with both THORChain and MayaChain positions.
2. THORChain DeFi → Manage Positions → deselect TCY → Done.
3. MayaChain DeFi → Manage Positions → deselect CACAO staking → Done.
4. Back to THORChain DeFi and reopen Manage Positions.

TCY is selected again: step 3 wrote Maya's selection over the THORChain one.

## Why it is not visible as data loss today

The Maya screen carries a fallback — a stored set holding no Maya key (`maya:bond:CACAO`,
`maya:stake:CACAO`, or a dotted LP asset) is treated as the other screen's and replaced with the
Maya defaults. So Maya degrades to its defaults rather than showing nothing, and THORChain shows
whatever Maya last wrote. The user's choice is lost either way, silently, on the next visit to the
other screen.

That fallback is also why a genuinely empty selection is awkward to represent on the Maya side: it
has to lean on the store returning its (non-empty) defaults for an absent key to tell "cleared on
purpose" apart from "belongs to the other screen". Keying per chain removes the need for the
heuristic entirely.

## Suggested fix

Key the selection per chain as well as per vault — e.g. `selected_positions_${chainId}_$vaultId` —
and migrate the existing value to the THORChain key on first read so nobody's current selection is
reset by the upgrade. The Maya `hasMayaPositions` fallback in `loadSavedPositions` can then go.

## Affected files

- `data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt`
- `app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt`
- `app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt`

aminsato:
Closes #5533

`DefiPositionsRepository` stored the Manage Positions selection under one key per vault:

```kotlin
private fun selectedPositionsKey(vaultId: String) =
    stringSetPreferencesKey("selected_positions_$vaultId")
```

Both `ThorchainDefiPositionsViewModel` and `MayachainDefiPositionsViewModel` read and write it, so on a vault holding both chains each screen's save clobbers the other's selection.

## Why the Maya fallback cannot be made to work

Nothing looks like data loss today only because the Maya screen guesses ownership from the shape of the stored keys:

```kotlin
saved.isEmpty() -> emptyList()
saved.any { it in MAYA_STATIC_POSITION_KEYS || it.contains(".") } -> saved.toList()
else -> MAYA_DEFAULT_SELECTED_POSITIONS
```

Each branch is wrong in a different direction, and the comment above it already concedes the premise — *"The THORChain screen writes this same key."*

**A plain-ticker set falls to the defaults, reviving a position the user turned off.** Maya: uncheck Bond, Done → stores `{maya:stake:CACAO}`. THORChain tab, same vault: uncheck RUJI, Done → overwrites the shared key with a set holding no Maya member. Maya's next read has `hasMayaPositions == false`, falls to `MAYA_DEFAULT_SELECTED_POSITIONS`, and Bond is checked again — its live balance refetched and re-added to the header, with zero Maya-side action. (Reported by @rkokhatskyi.)

**An empty set zeroes the other screen.** Since round 6 of #5523, `saved.isEmpty()` is read unconditionally as "Maya cleared its own selection". `onPositionSelectionDone` has no minimum-selection guard on either screen, so unchecking all six THORChain positions persists an empty set, and the Maya tab drops Bond, Staking and LP to $0.00 on real CACAO positions the user never touched. (Also reported by @rkokhatskyi.)

**The dotted branch is ambiguous the other way.** THORChain LP pool keys are dotted too — `positionKey = pool`, e.g. `BTC.BTC`. A THORChain selection containing any pool passes `it.contains(".")`, so Maya adopts that entire selection verbatim and renders THORChain's pools.

## The fix

The key carries the chain — `defi_selected_positions_${chain.id}_$vaultId` — so there is nothing left to guess and the whole `when` block goes.

**The read now returns `Flow<Set<String>?>`, and that is what actually retires the heuristic.** The store previously answered both "never chose" and "cleared on purpose" with the same default set; telling those apart is the entire reason the `when` block existed. Null is now "never chose" and it is the only case defaults apply to, so an empty set means empty. Defaults move to the screens, where they already differ per chain — and the store's THORChain set was a duplicate of `defaultSelectedPositionsDialog()`, so that duplication goes too.

## Migration

A `DataMigration` moves each pre-upgrade value onto the THORChain key at store creation, so no one's current selection is reset by the upgrade.

The new keys take a distinct prefix (`defi_selected_positions_`) rather than extending the old one. That keeps the two namespaces from overlapping, so `shouldMigrate` cannot recognise its own output and re-run — a second pass would overwrite a selection made *after* the upgrade with the pre-upgrade one. There is a test for exactly that.

**One deliberate trade-off.** THORChain receives the legacy value because it is the screen that writes the shared key on any vault that has one; Maya only ever persisted through it after the user opened a second screen. A vault whose last write happened to come from the Maya screen loses that one selection here — once, at upgrade — instead of carrying the ambiguity forward forever. Guessing the author from key shape is what this PR is removing, so the migration does not do it either. It is documented at the migration.

## Tests

**6 new migration tests** (`DefiPositionsRepositoryMigrationTest`) covering the move onto the THORChain key, multiple vaults in one pass, unrelated preferences left alone, idempotency, and post-upgrade keys not being mistaken for legacy ones. One is a real edge case: a **cleared** selection has to survive as an empty set, because dropping it would read back as "never chose" and put every position the user had unchecked in front of them again on the first launch after the upgrade.

**A test that pinned the bug is now inverted.** `a saved selection holding no Maya key falls back to the Maya defaults` asserted precisely the revive behaviour above. It is now `a stored selection is used verbatim even when no key in it is Maya-specific`, which fails on `main`.

Also added: null-path coverage on both screens, and a cross-chain isolation test driving a THORChain write (empty, then plain-ticker) past the Maya screen. That one is weaker by construction — the mock is keyed by chain, so it verifies the ViewModels pass the right chain rather than proving two keys coexist in a real store. The migration tests are where the store-level logic is exercised.

## Gate

- `:data:testDebugUnitTest` and `:app:testDebugUnitTest` — full suites, **BUILD SUCCESSFUL**. 91 tests across the three touched classes: 38 Maya, 47 THORChain, 6 migration.
- `:app:lintDebug` — **BUILD SUCCESSFUL**.
- `ktfmt` applied.

## Manual test plan

Requires a vault with THORChain and MayaChain enabled. Install with `adb install -r` — `uninstall` wipes the DataStore and the migration never runs.

1. **The reported repro.** THORChain → Manage Positions → uncheck TCY → Done. MayaChain → Manage Positions → uncheck CACAO staking → Done. Back to THORChain → Manage Positions: TCY is checked again on `main`, stays unchecked here.
2. **Revive.** Maya → uncheck Bond → Done. THORChain → uncheck RUJI → Done. Back to Maya: Bond is checked again on `main`, stays unchecked here.
3. **Zeroing.** THORChain → uncheck all six → Done. Open MayaChain: header drops to $0.00 with all three tabs empty on `main`; shows its own selection and real total here. Best seen on a vault with actual CACAO.
4. **Upgrade.** Install `main`, uncheck TCY and yRUNE on THORChain, force-stop, then `adb install -r` this branch. The THORChain selection survives; Maya opens on its defaults, having no stored selection of its own yet.

## Summary by CodeRabbit

* **Bug Fixes**
  * DeFi position selections are now stored separately for each blockchain and vault.
  * MayaChain and THORChain selections no longer affect one another.
  * Explicitly cleared selections remain empty instead of being replaced by defaults.
  * Existing saved selections are migrated automatically while preserving unrelated preferences.
  * Default positions are applied only when no saved selection exists.

* **Tests**
  * Added coverage for migration, empty selections, defaults, cross-chain isolation, and clearing all positions.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
@@ -45,15 +45,26 @@ private fun selectedPositionsKey(chain: Chain, vaultId: String) =
  * the Bonded tab and its header leg would stay on with no row in the dialog able to switch them
  * off.
  *
- * Pool keys are the one thing the shape cannot attribute — both chains write `CHAIN.TICKER` and
- * both list pools like `BTC.BTC` — so they go to both chains. Landing on a chain the user did not
- * pick them on shows a pool card they can uncheck; dropping them would silently shorten the header
- * total instead.
+ * A chain's key is only written when the set holds something that chain can own. Left out, it reads
+ * back as null — never chose — and the screen answers with its own defaults, which is what that
+ * vault was already being shown. Writing an empty set instead would call it a clearing and blank a
+ * screen the vault never configured: a set of only `maya:` keys lit both THORChain legs before the
+ * upgrade, since `hasBondPositions` and `hasStakingPositions` match the Cacao keys too.
  *
- * MayaChain is skipped for the one shape it cannot have authored: a set of plain tickers and
- * nothing else, which is what that screen already answered with its defaults before the upgrade. An
- * empty set is not that shape. It is a selection cleared on purpose, and leaving it out would hand
- * back the Bond and Staking rows the user had switched off.
+ * An empty set is written to both. That one is a clearing both screens saw, and leaving it out
+ * would hand back every row the user had switched off.
+ *
+ * Pool keys are the one thing the shape cannot attribute — both chains write `CHAIN.TICKER` and
+ * both list pools like `BTC.BTC` — so they ride along to MayaChain whenever a `maya:` key is
+ * already carrying it, and a set of nothing but pools counts as a MayaChain choice on its own,
+ * because that is exactly what the Maya dialog leaves behind when both Cacao rows are unchecked.
+ * Alongside plain tickers the set reads as THORChain's, and MayaChain's own Bond and Staking
+ * defaults serve that vault better than a lone pool card.
+ *
+ * What that costs is worth naming: the Maya dialog seeds itself from the stored set and only
+ * toggles its own rows, so a Maya selection can end up as plain tickers it never picked, and
+ * nothing in the shape tells it apart from a THORChain-authored one. It gets MayaChain's defaults —
+ * the same answer that screen's own read gave this shape before the upgrade.
  */
 internal object SelectedPositionsPerChainMigration : DataMigration<Preferences> {
     override suspend fun shouldMigrate(currentData: Preferences): Boolean =
@@ -79,14 +90,18 @@ internal fun migrateSelectedPositionsPerChain(currentData: Preferences): Prefere
         val thorchainPositions = positions - mayaPositions
         val poolPositions = thorchainPositions.filterTo(mutableSetOf()) { it.isPoolPositionKey() }
 
-        // Written even when empty: the user cleared every THORChain position on purpose, and an
-        // absent key would read back as "never chose" and hand them the defaults again.
-        migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = thorchainPositions
+        // A set of only `maya:` keys says nothing about THORChain, and recording it as a clearing
+        // would blank a screen that was showing both its legs before the upgrade.
+        if (positions.isEmpty() || thorchainPositions.isNotEmpty()) {
+            migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = thorchainPositions
+        }
 
-        // Unchecking both Cacao rows leaves a set with no `maya:` key in it, so their presence
-        // cannot be what decides this. A pool or an emptied set is just as much a MayaChain choice,
-        // and only a set of plain tickers rules that screen out.
-        if (positions.isEmpty() || mayaPositions.isNotEmpty() || poolPositions.isNotEmpty()) {
+        // A `maya:` key is the only attribution the shape carries. Pools are the exception the
+        // Maya dialog forces: unchecking both Cacao rows leaves a set with no `maya:` key at all,
+        // so a set of nothing but pools is a MayaChain choice too. Pools sitting next to plain
+        // tickers are not — that set reads as THORChain's, and Maya's defaults beat a lone pool.
+        val isPoolOnlySelection = positions.isNotEmpty() && positions == poolPositions
+        if (positions.isEmpty() || mayaPositions.isNotEmpty() || isPoolOnlySelection) {
             migrated[selectedPositionsKey(Chain.MayaChain, vaultId)] = mayaPositions + poolPositions
         }
     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
@@ -50,9 +50,10 @@ private fun selectedPositionsKey(chain: Chain, vaultId: String) =
  * pick them on shows a pool card they can uncheck; dropping them would silently shorten the header
  * total instead.
  *
- * MayaChain is only written when a `maya:` key proves a choice was made there. Writing it for a
- * THORChain-only vault would leave a screen the user has never opened with nothing selected rather
- * than with its defaults.
+ * MayaChain is skipped for the one shape it cannot have authored: a set of plain tickers and
+ * nothing else, which is what that screen already answered with its defaults before the upgrade. An
+ * empty set is not that shape. It is a selection cleared on purpose, and leaving it out would hand
+ * back the Bond and Staking rows the user had switched off.
  */
 internal object SelectedPositionsPerChainMigration : DataMigration<Preferences> {
     override suspend fun shouldMigrate(currentData: Preferences): Boolean =
@@ -76,13 +77,17 @@ internal fun migrateSelectedPositionsPerChain(currentData: Preferences): Prefere
 
         val mayaPositions = positions.filterTo(mutableSetOf()) { it.isMayaPositionKey() }
         val thorchainPositions = positions - mayaPositions
+        val poolPositions = thorchainPositions.filterTo(mutableSetOf()) { it.isPoolPositionKey() }
 
         // Written even when empty: the user cleared every THORChain position on purpose, and an
         // absent key would read back as "never chose" and hand them the defaults again.
         migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = thorchainPositions
-        if (mayaPositions.isNotEmpty()) {
-            migrated[selectedPositionsKey(Chain.MayaChain, vaultId)] =
-                mayaPositions + thorchainPositions.filter { it.isPoolPositionKey() }
+
+        // Unchecking both Cacao rows leaves a set with no `maya:` key in it, so their presence
+        // cannot be what decides this. A pool or an emptied set is just as much a MayaChain choice,
+        // and only a set of plain tickers rules that screen out.
+        if (positions.isEmpty() || mayaPositions.isNotEmpty() || poolPositions.isNotEmpty()) {
+            migrated[selectedPositionsKey(Chain.MayaChain, vaultId)] = mayaPositions + poolPositions
         }
     }
     return migrated
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
@@ -31,6 +31,13 @@ private const val SELECTED_POSITIONS_PREFIX = "defi_selected_positions_"
 private const val MAYA_POSITION_KEY_PREFIX = "maya:"
 private const val POOL_POSITION_KEY_SEPARATOR = "."
 
+// What each screen answers a vault that never chose with, frozen at the shape they had when this
+// migration shipped for the same reason the key prefixes are. Only ever written next to a pool: it
+// is how an unattributable pool is kept without the vault being recorded as having chosen anything
+// else, and a vault that reaches this migration with no pool at all is left without a key instead.
+private val MAYA_DEFAULT_POSITION_KEYS = setOf("maya:bond:CACAO", "maya:stake:CACAO")
+private val THORCHAIN_DEFAULT_POSITION_KEYS = setOf("RUNE", "RUJI", "TCY", "sTCY", "yRUNE", "yTCY")
+
 private fun selectedPositionsKey(chain: Chain, vaultId: String) =
     stringSetPreferencesKey("$SELECTED_POSITIONS_PREFIX${chain.id}_$vaultId")
 
@@ -45,30 +52,31 @@ private fun selectedPositionsKey(chain: Chain, vaultId: String) =
  * the Bonded tab and its header leg would stay on with no row in the dialog able to switch them
  * off.
  *
- * A chain's key is only written when the set holds something that chain can own. Left out, it reads
- * back as null — never chose — and the screen answers with its own defaults, which is what that
- * vault was already being shown. Writing an empty set instead would call it a clearing and blank a
- * screen the vault never configured: a set of only `maya:` keys lit both THORChain legs before the
- * upgrade, since `hasBondPositions` and `hasStakingPositions` match the Cacao keys too.
+ * Only the `maya:` keys and the plain tickers attribute themselves: the first pair can only have
+ * come from the Maya dialog, the second only from THORChain's. A chain with none of its own keys in
+ * the set has not been heard from, and its key is left unwritten, which reads back as null — never
+ * chose — so the screen answers with its own defaults. Writing an empty set instead would call it a
+ * clearing and blank a screen the vault never configured: a set of only `maya:` keys lit both
+ * THORChain legs before the upgrade, since `hasBondPositions` and `hasStakingPositions` match the
+ * Cacao keys too.
  *
  * An empty set is written to both. That one is a clearing both screens saw, and leaving it out
  * would hand back every row the user had switched off.
  *
- * Pool keys are the one thing the shape cannot attribute — both chains write `CHAIN.TICKER` and
- * both list pools like `BTC.BTC` — so every pool is kept on both chains, a pool-only set included.
- * Landing on a chain the user did not pick it on shows a pool card they can uncheck; dropping it
- * would move a THORChain LP position onto the Maya screen and take the header total with it.
+ * Pool keys attribute nothing — both chains write `CHAIN.TICKER` and both list pools like `BTC.BTC`
+ * — so a pool is never evidence that its chain chose, and never dropped either. It rides along to
+ * both, next to whatever each chain has of its own, or next to that chain's defaults where it has
+ * nothing. Landing on a chain the user did not pick it on shows a pool card they can uncheck;
+ * dropping it would empty an LP tab that was showing it before the upgrade.
  *
- * What a pool cannot settle by itself is whether MayaChain chose at all. It counts as MayaChain
- * evidence only when the set holds nothing else, because that is exactly what the Maya dialog
- * leaves behind when both Cacao rows are unchecked. Sitting next to plain tickers the set reads as
- * THORChain's, and MayaChain's own Bond and Staking defaults serve that vault better than a lone
- * pool card.
+ * A set of nothing but pools is the exception on both sides: it is the whole selection each screen
+ * had, and it is what the Maya dialog leaves behind when both Cacao rows are unchecked, so pairing
+ * it with the defaults would hand back the rows that had to come off to leave that shape.
  *
- * What that costs is worth naming: the Maya dialog seeds itself from the stored set and only
- * toggles its own rows, so a Maya selection can end up as plain tickers it never picked, and
- * nothing in the shape tells it apart from a THORChain-authored one. It gets MayaChain's defaults —
- * the same answer that screen's own read gave this shape before the upgrade.
+ * What that costs is worth naming: each dialog seeds itself from the stored set and only toggles
+ * its own rows, so a Maya selection can end up as plain tickers it never picked, and nothing in the
+ * shape tells it apart from a THORChain-authored one. It gets MayaChain's defaults — the same
+ * answer that screen's own read gave this shape before the upgrade.
  */
 internal object SelectedPositionsPerChainMigration : DataMigration<Preferences> {
     override suspend fun shouldMigrate(currentData: Preferences): Boolean =
@@ -93,26 +101,43 @@ internal fun migrateSelectedPositionsPerChain(currentData: Preferences): Prefere
         val mayaPositions = positions.filterTo(mutableSetOf()) { it.isMayaPositionKey() }
         val thorchainPositions = positions - mayaPositions
         val poolPositions = thorchainPositions.filterTo(mutableSetOf()) { it.isPoolPositionKey() }
+        val tickerPositions = thorchainPositions - poolPositions
 
-        // A set of only `maya:` keys says nothing about THORChain, and recording it as a clearing
-        // would blank a screen that was showing both its legs before the upgrade.
-        if (positions.isEmpty() || thorchainPositions.isNotEmpty()) {
-            migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = thorchainPositions
-        }
+        selectionFor(positions, tickerPositions, poolPositions, THORCHAIN_DEFAULT_POSITION_KEYS)
+            ?.let { migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = it }
 
-        // A `maya:` key is the only attribution the shape carries. Pools are the exception the
-        // Maya dialog forces: unchecking both Cacao rows leaves a set with no `maya:` key at all,
-        // so a set of nothing but pools is a MayaChain choice too — a choice it shares with
-        // THORChain above, since the pool itself stays unattributable. Pools sitting next to plain
-        // tickers do not count — that set reads as THORChain's, and Maya's defaults beat a lone
-        // pool.
-        val isPoolOnlySelection = positions.isNotEmpty() && positions == poolPositions
-        if (positions.isEmpty() || mayaPositions.isNotEmpty() || isPoolOnlySelection) {
-            migrated[selectedPositionsKey(Chain.MayaChain, vaultId)] = mayaPositions + poolPositions
+        selectionFor(positions, mayaPositions, poolPositions, MAYA_DEFAULT_POSITION_KEYS)?.let {
+            migrated[selectedPositionsKey(Chain.MayaChain, vaultId)] = it
         }
     }
     return migrated
 }
+
+/**
+ * What one chain takes from a pre-upgrade set, or null to leave its key unwritten so the screen
+ * answers with its own defaults. [ownKeys] is the part of the set only this chain can have written:
+ * the `maya:` keys for MayaChain, the plain tickers for THORChain. Pools belong to neither and are
+ * never evidence for either — they are only ever carried.
+ */
+private fun selectionFor(
+    positions: Set<String>,
+    ownKeys: Set<String>,
+    poolPositions: Set<String>,
+    defaults: Set<String>,
+): Set<String>? =
+    when {
+        // A clearing both screens saw.
+        positions.isEmpty() -> emptySet()
+        // Nothing but pools is the whole selection on both screens, and pairing it with the
+        // defaults would hand back the rows that had to be unchecked to leave this shape behind.
+        positions == poolPositions -> poolPositions
+        ownKeys.isNotEmpty() -> ownKeys + poolPositions
+        // No say of its own, but a pool that may still be this chain's. Dropping it would empty an
+        // LP tab that was showing it; keeping it alone would read as a choice this vault never
+        // made.
+        poolPositions.isNotEmpty() -> defaults + poolPositions
+        else -> null
+    }
 
 private fun String.isMayaPositionKey() = startsWith(MAYA_POSITION_KEY_PREFIX)
 
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
@@ -1,49 +1,88 @@
 package com.vultisig.wallet.data.repositories
 
 import android.content.Context
+import androidx.datastore.core.DataMigration
+import androidx.datastore.preferences.core.Preferences
 import androidx.datastore.preferences.core.edit
 import androidx.datastore.preferences.core.stringSetPreferencesKey
 import androidx.datastore.preferences.preferencesDataStore
-import com.vultisig.wallet.data.models.Coins
+import com.vultisig.wallet.data.models.Chain
 import dagger.hilt.android.qualifiers.ApplicationContext
 import javax.inject.Inject
 import javax.inject.Singleton
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
 
-private val Context.dataStore by preferencesDataStore(name = "defi_positions_preferences")
+private val Context.dataStore by
+    preferencesDataStore(
+        name = "defi_positions_preferences",
+        produceMigrations = { listOf(SelectedPositionsPerChainMigration) },
+    )
+
+// Every DeFi screen used to share one key per vault, so each screen's save overwrote the other's
+// selection. The prefix changes with the shape so the two namespaces cannot overlap and the
+// migration below can recognise a pre-upgrade key by name alone.
+internal const val LEGACY_SELECTED_POSITIONS_PREFIX = "selected_positions_"
+private const val SELECTED_POSITIONS_PREFIX = "defi_selected_positions_"
+
+private fun selectedPositionsKey(chain: Chain, vaultId: String) =
+    stringSetPreferencesKey("$SELECTED_POSITIONS_PREFIX${chain.id}_$vaultId")
+
+/**
+ * Moves every pre-upgrade selection onto the THORChain key so nobody's current choice is reset by
+ * the upgrade.
+ *
+ * THORChain is where the value goes because it is the screen that writes the shared key on any
+ * vault that has one: MayaChain only ever persisted through it after the user opened a second
+ * screen, and a Maya-authored value read back on the Maya key would be indistinguishable from a
+ * genuine Maya choice. A vault whose last write happened to come from the Maya screen loses that
+ * one selection here — once, at upgrade — instead of carrying the ambiguity forward forever.
+ */
+internal object SelectedPositionsPerChainMigration : DataMigration<Preferences> {
+    override suspend fun shouldMigrate(currentData: Preferences): Boolean =
+        currentData.asMap().keys.any { it.name.startsWith(LEGACY_SELECTED_POSITIONS_PREFIX) }
+
+    override suspend fun migrate(currentData: Preferences): Preferences =
+        migrateSelectedPositionsToThorchain(currentData)
+
+    override suspend fun cleanUp() = Unit
+}
+
+internal fun migrateSelectedPositionsToThorchain(currentData: Preferences): Preferences {
+    val migrated = currentData.toMutablePreferences()
+    currentData.asMap().forEach { (key, value) ->
+        if (!key.name.startsWith(LEGACY_SELECTED_POSITIONS_PREFIX)) return@forEach
+        migrated.remove(stringSetPreferencesKey(key.name))
+
+        val vaultId = key.name.removePrefix(LEGACY_SELECTED_POSITIONS_PREFIX)
+        val positions = (value as? Set<*>)?.filterIsInstance<String>()?.toSet()
+        if (vaultId.isNotEmpty() && positions != null) {
+            migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = positions
+        }
+    }
+    return migrated
+}
 
 @Singleton
 class DefiPositionsRepository
 @Inject
 constructor(@ApplicationContext private val context: Context) {
-    companion object {
-        private fun selectedPositionsKey(vaultId: String) =
-            stringSetPreferencesKey("selected_positions_$vaultId")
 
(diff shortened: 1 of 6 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
```diff
@@ -1,49 +1,88 @@
...
+}
+
+internal fun migrateSelectedPositionsToThorchain(currentData: Preferences): Preferences {
+    val migrated = currentData.toMutablePreferences()
+    currentData.asMap().forEach { (key, value) ->
+        if (!key.name.startsWith(LEGACY_SELECTED_POSITIONS_PREFIX)) return@forEach
+        migrated.remove(stringSetPreferencesKey(key.name))
+
+        val vaultId = key.name.removePrefix(LEGACY_SELECTED_POSITIONS_PREFIX)
+        val positions = (value as? Set<*>)?.filterIsInstance<String>()?.toSet()
```
rkokhatskyi, line 98: Nothing filters `maya:bond:CACAO` out of the migrated set, and `hasBondPositions` matches it at `DeFiExtensions.kt:61` while no THORChain dialog row owns that key. The Bonded tab and its header leg stay on with every checkbox unchecked, and `onPositionSelectionChange` cannot remove a key no row owns

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
```diff
@@ -1,49 +1,88 @@
...
+        if (vaultId.isNotEmpty() && positions != null) {
+            migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = positions
```
rkokhatskyi: The MayaChain key is never written, so `MayachainDefiPositionsViewModel.kt:287` reads null and falls to `MAYA_DEFAULT_SELECTED_POSITIONS`. A Maya-last-writer vault loses its selection at upgrade, and `selectedLpKeys` at :643 goes empty, dropping a chosen pool's fiat from the header total

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
```diff
@@ -1,49 +1,88 @@
...
+
+private fun selectedPositionsKey(chain: Chain, vaultId: String) =
+    stringSetPreferencesKey("$SELECTED_POSITIONS_PREFIX${chain.id}_$vaultId")
+
+/**
+ * Moves every pre-upgrade selection onto the THORChain key so nobody's current choice is reset by
+ * the upgrade.
+ *
+ * THORChain is where the value goes because it is the screen that writes the shared key on any
+ * vault that has one: MayaChain only ever persisted through it after the user opened a second
```
rkokhatskyi: `MayaChain only ever persisted through it after the user opened a second screen` describes no code path, because main's `MayachainDefiPositionsViewModel.onPositionSelectionDone` writes the shared key behind only a no-op guard, so a Maya-only vault's legacy value is always Maya-authored

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
rkokhatskyi: One scope correction: it is not only a Maya-last-writer vault. The THORChain dialog can only toggle its own tickers and pools, and `setPositionSelectionDialogVisibility` seeds `tempSelectedPositions` from the stored set, so the `maya:` keys survive every later THORChain save intact. The migrated value therefore still carries Maya's last selection even when THORChain wrote last, and that selection is discarded for every vault whose Maya choice ever differed from the Maya defaults

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: Confirmed and fixed in 9ceec28f5.

You are right on every leg of it: `hasBondPositions` at `DeFiExtensions.kt:61` matches `MAYA_BOND_CACAO_KEY`, so `ThorchainDefiPositionsViewModel.kt:504` and `ThorchainDefiPositionsScreen.kt:181` both light up off a key no Thorchain dialog row owns — and `onPositionSelectionChange` at `:1245` only adds or removes keys its own rows carry, so there was genuinely no way back off it.

The migration now reads ownership off each key instead of moving the set wholesale: `maya:` keys go to the MayaChain key, plain tickers to Thorchain. A `maya:` key can no longer reach the Thorchain side at all, which is asserted directly in `a Maya key never lands on Thorchain`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: Confirmed and fixed in 9ceec28f5 — and your scope correction below is right too, so I am answering both here.

I checked the path you describe: `setPositionSelectionDialogVisibility` seeds `tempSelectedPositions` from the stored set at `ThorchainDefiPositionsViewModel.kt:1231`, and `onPositionSelectionChange` can only toggle keys its own rows own, so `maya:` keys do survive every later Thorchain save intact. The stored value therefore still carried Maya's last selection whoever wrote last, and the loss was not limited to a Maya-last-writer vault the way I had scoped it. The `selectedLpKeys` consequence at `:643` follows exactly as you describe, since `MAYA_DEFAULT_SELECTED_POSITIONS` is the two static keys and nothing else.

The migration now writes the MayaChain key whenever a `maya:` key is present, so a Maya choice is carried across rather than reset.

One thing I could not resolve cleanly, so I want it on the record rather than buried in the KDoc. Pool keys cannot be attributed from the string: both screens write `CHAIN.TICKER` and both list pools like `BTC.BTC`, so there is nothing in the key to say who chose it. I send them to both chains. A pool landing on the chain it was not chosen on has a dialog row and can be unchecked; dropping it would silently shorten the header total, which is the failure you flagged. I verified neither screen hangs on a selected pool key that matches no dialog row — `ThorchainDefiPositionsViewModel.kt:974` and `MayachainDefiPositionsViewModel.kt:661` both settle on a reported zero.

The related trap, which is the mirror of your first comment: MayaChain is only written when a `maya:` key proves a choice was made there. Writing it unconditionally would hand a Thorchain-only vault a Maya screen with nothing selected instead of its defaults, because `hasBondPositions` matches the `RUNE` ticker as well.

Covered by `a Maya selection is kept instead of being reset to the Maya defaults`, `a pool the shape cannot attribute is kept on both chains`, and `a vault that never chose on Maya is left without a Maya key`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: You are right, and the sentence is gone in 9ceec28f5.

I checked `main` before rewriting it: `MayachainDefiPositionsViewModel.onPositionSelectionDone` writes the shared key behind only the no-op guard comparing `tempSelectedPositions` against `selectedPositions`, with nothing gating it on a second screen having been opened. So a Maya-only vault's legacy value is Maya-authored, exactly as you say, and the justification I built on top of that claim did not hold.

The rewritten KDoc drops the argument entirely rather than patching it, since the migration no longer needs a reason to prefer one chain — it splits by key ownership, so there is no last-writer guess left to justify.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
```diff
@@ -1,49 +1,117 @@
...
+        val positions = (value as? Set<*>)?.filterIsInstance<String>()?.toSet()
+        if (vaultId.isEmpty() || positions == null) return@forEach
+
+        val mayaPositions = positions.filterTo(mutableSetOf()) { it.isMayaPositionKey() }
+        val thorchainPositions = positions - mayaPositions
+
+        // Written even when empty: the user cleared every THORChain position on purpose, and an
+        // absent key would read back as "never chose" and hand them the defaults again.
+        migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = thorchainPositions
+        if (mayaPositions.isNotEmpty()) {
```
rkokhatskyi: `mayaPositions.isNotEmpty()` reads an absent maya: key as never chose, but unchecking both `MAYA_BOND_CACAO_KEY` and `MAYA_STAKE_CACAO_KEY` on Maya can leave a pool-only or empty legacy set. `MayachainDefiPositionsViewModel.kt:287` then reverts that vault to `MAYA_DEFAULT_SELECTED_POSITIONS`, reviving Bond and Staking and losing the chosen pool, while the same value is written to THORChain's key instead.

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: Confirmed and fixed in `bbbd0fb7f`.

You are right that the presence of a `maya:` key cannot be what decides this. `onPositionSelectionChange` removes the key outright, so unchecking both Cacao rows leaves a set with no `maya:` key in it at all — a pool-only set, or an empty one — and `mayaPositions.isNotEmpty()` read both of those as never chose. `MayachainDefiPositionsViewModel.kt:287` then handed the vault `MAYA_DEFAULT_SELECTED_POSITIONS`, so Bond and Staking came back on a screen the user had switched them off on, and the pool they kept went only to the THORChain key.

What settled the rule for me was `main`'s own behaviour, since the migration owes each screen what it was already showing. Pre-upgrade the Maya screen read the shared key as `saved.isEmpty() -> emptyList()`, then `saved.any { it in MAYA_STATIC_POSITION_KEYS || it.contains(".") } -> saved.toList()`, and fell to the defaults only for a set of plain tickers. The gate I wrote matched the middle branch alone, which is exactly why the other two shapes changed under it.

The condition now covers all three:

```kotlin
if (positions.isEmpty() || mayaPositions.isNotEmpty() || poolPositions.isNotEmpty()) {
    migrated[selectedPositionsKey(Chain.MayaChain, vaultId)] = mayaPositions + poolPositions
}
```

A set of plain tickers still skips the MayaChain key, which keeps the case from your earlier thread intact: a screen the vault has never opened gets its defaults rather than an empty dialog. `unchecking both Cacao rows is a Maya choice, not an absent one` asserts the pool-only set now, and the cleared-selection test asserts an empty MayaChain key instead of an absent one. The never-chose test was itself carrying a pool in its fixture, so it was asserting the bug; it holds plain tickers now.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
```diff
@@ -1,49 +1,122 @@
...
+        val poolPositions = thorchainPositions.filterTo(mutableSetOf()) { it.isPoolPositionKey() }
+
+        // Written even when empty: the user cleared every THORChain position on purpose, and an
+        // absent key would read back as "never chose" and hand them the defaults again.
+        migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = thorchainPositions
+
+        // Unchecking both Cacao rows leaves a set with no `maya:` key in it, so their presence
+        // cannot be what decides this. A pool or an emptied set is just as much a MayaChain choice,
+        // and only a set of plain tickers rules that screen out.
+        if (positions.isEmpty() || mayaPositions.isNotEmpty() || poolPositions.isNotEmpty()) {
```
rkokhatskyi: `poolPositions.isNotEmpty()` treats a dotted key as proof of a Maya choice, but THORChain pool keys are dotted too. A vault that only ticked a THORChain pool gets a Maya key holding just that pool, so `hasBondPositions` at `DeFiExtensions.kt:61` no longer matches and the bonded CACAO that main showed leaves the Maya header

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
```diff
@@ -1,49 +1,122 @@
...
+        val positions = (value as? Set<*>)?.filterIsInstance<String>()?.toSet()
+        if (vaultId.isEmpty() || positions == null) return@forEach
+
+        val mayaPositions = positions.filterTo(mutableSetOf()) { it.isMayaPositionKey() }
+        val thorchainPositions = positions - mayaPositions
+        val poolPositions = thorchainPositions.filterTo(mutableSetOf()) { it.isPoolPositionKey() }
+
+        // Written even when empty: the user cleared every THORChain position on purpose, and an
+        // absent key would read back as "never chose" and hand them the defaults again.
+        migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = thorchainPositions
```
rkokhatskyi: The THORChain write has no guard, so a legacy set of only `maya:` keys leaves `thorchainPositions` empty and records that as cleared. `ThorchainDefiPositionsViewModel.kt:479` then reads a present empty set rather than null, skips `defaultSelectedPositionsDialog()` and blanks a real RUNE bond main showed

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
```diff
@@ -1,49 +1,122 @@
...
+ * `maya:` key handed to THORChain would be worse than dropped: `hasBondPositions` matches it, so
+ * the Bonded tab and its header leg would stay on with no row in the dialog able to switch them
+ * off.
+ *
+ * Pool keys are the one thing the shape cannot attribute — both chains write `CHAIN.TICKER` and
+ * both list pools like `BTC.BTC` — so they go to both chains. Landing on a chain the user did not
+ * pick them on shows a pool card they can uncheck; dropping them would silently shorten the header
+ * total instead.
+ *
+ * MayaChain is skipped for the one shape it cannot have authored: a set of plain tickers and
```
rkokhatskyi: `the one shape it cannot have authored` is wrong because the Maya dialog seeds `tempSelectedPositions` from the stored set at `MayachainDefiPositionsViewModel.kt:899` and only toggles its own rows, so unchecking both Cacao rows and every pool persists exactly a plain ticker set

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: Confirmed and fixed in `51cf21065`.

Every leg of it holds. `hasStakingPositions` at `DeFiExtensions.kt:65` matches `MAYA_STAKE_CACAO_KEY` the same way `hasBondPositions` matches the bond one, so a legacy set of only `maya:` keys had *both* THORChain legs loading before the upgrade, not just the bonded one. Writing the empty remainder handed `ThorchainDefiPositionsViewModel.kt:479` a present empty set, so it skipped `defaultSelectedPositionsDialog()` and the screen went to `NoPositionsContainer` with `_totalValueBond` zeroed — a real RUNE bond gone from a screen the vault had never chosen on.

What settled the direction was that a maya-only set is not a THORChain statement at all. It is reachable with no THORChain dialog interaction whatsoever: the Maya dialog's first save on a vault with no stored key writes exactly `{maya:bond:CACAO, maya:stake:CACAO}`. And in the other authoring path — a user clearing every THORChain row — main did not honour that clearing either, since the `maya:` keys kept both legs alive regardless. Neither path supports reading it as cleared.

The write is now guarded on the set holding something THORChain can own:

```kotlin
if (positions.isEmpty() || thorchainPositions.isNotEmpty()) {
    migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = thorchainPositions
}
```

An empty legacy set still goes through, since that one is a clearing both screens saw. Covered by `a vault that never chose on Thorchain is left without a Thorchain key`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: Fixed in `51cf21065`, with one thing I want to put in front of you rather than decide silently.

The finding holds for the shape a THORChain user actually leaves behind. With `{RUNE, BTC.BTC}`, main's Maya read took the dotted branch and returned the set at face value, so `hasBondPositions` matched on `RUNE` and the Maya header carried its bonded leg. The old guard handed Maya `{BTC.BTC}`, nothing matched, and the leg went off a screen that vault had never chosen on. `poolPositions.isNotEmpty()` was standing in as proof of a Maya choice on a key shape the KDoc itself admits cannot attribute.

Where I could not follow you literally is `a vault that only ticked a THORChain pool`. Read strictly as a pool and nothing else, that is the same shape as your last finding, which needed a pool-only set to stay a Maya choice — unchecking both Cacao rows leaves exactly that. Taken strictly it also does not produce the consequence you describe: for `{BTC.BTC}` alone, main's Maya read returned `["BTC.BTC"]`, so `hasBondPositions` was already false and no bonded CACAO was on that header to lose.

So I took the reading that satisfies both findings — a pool counts for MayaChain only when the set holds nothing else, and never as proof alongside plain tickers:

```kotlin
val isPoolOnlySelection = positions.isNotEmpty() && positions == poolPositions
if (positions.isEmpty() || mayaPositions.isNotEmpty() || isPoolOnlySelection) {
```

Pools still ride along into the value whenever a `maya:` key is already carrying it. Covered by `a Thorchain pool next to tickers is not read as a Maya choice`, alongside the pool-only case from last round. If you did mean the strict version and want a pool-only set to go to THORChain alone, say so and I will drop the carve-out with the earlier test.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: Confirmed and fixed in `51cf21065`.

I traced the path you describe: `setPositionSelectionDialogVisibility` seeds `tempSelectedPositions` from `selectedPositions` at `MayachainDefiPositionsViewModel.kt:899`, and main's Maya read returned the stored set at face value on the dotted-or-static branch, so THORChain tickers were sitting in that temp set. `onPositionSelectionChange` only ever adds or removes keys Maya's own rows carry, so unchecking both Cacao rows and every pool persists precisely a plain ticker set. The sentence was wrong.

The skip stays, but the reason is attribution rather than authorship, and the cost is now named instead of denied. A `maya:` key is the only marker the shape carries; a plain ticker set holds nothing pointing at MayaChain, and that vault gets the Maya defaults — the same answer that screen's own read gave this shape before the upgrade, since main fell through to `MAYA_DEFAULT_SELECTED_POSITIONS` for a set of plain tickers too. So the vault in your path is not losing anything to the upgrade that it was not already losing on the next read.

Writing an empty Maya key instead would be far worse: every THORChain-only vault that ever opened the positions dialog would land on a blank Maya screen, Bond and Staking included, on a chain it had never touched. The KDoc now carries that trade-off in the open.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
```diff
@@ -1,49 +1,137 @@
...
+        if (vaultId.isEmpty() || positions == null) return@forEach
+
+        val mayaPositions = positions.filterTo(mutableSetOf()) { it.isMayaPositionKey() }
+        val thorchainPositions = positions - mayaPositions
+        val poolPositions = thorchainPositions.filterTo(mutableSetOf()) { it.isPoolPositionKey() }
+
+        // A set of only `maya:` keys says nothing about THORChain, and recording it as a clearing
+        // would blank a screen that was showing both its legs before the upgrade.
+        if (positions.isEmpty() || thorchainPositions.isNotEmpty()) {
+            migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = thorchainPositions
```
johnnyluo: The doc comment above this function says a set of nothing-but-pools "counts as a MayaChain choice **on its own**" (i.e. exclusively Maya's), but this branch doesn't exclude that case.

`thorchainPositions` is `positions - mayaPositions`, with no carve-out for pool-only sets. So for a legacy value like `{"BTC.BTC"}` (no `maya:` prefix, no plain ticker — exactly what's left when a user unchecks both Cacao rows on the Maya screen, keeping only an LP pool):

```
mayaPositions = {}
thorchainPositions = {"BTC.BTC"}   // nonempty -> this branch fires
```

`migrated[thorchainKey]` ends up `{"BTC.BTC"}`, and (via `isPoolOnlySelection` below) `migrated[mayaKey]` also ends up `{"BTC.BTC"}` — the pool is written to **both** chains.

Consequence: post-migration, `ThorchainDefiPositionsViewModel.loadSavedPositions` reads a non-null set for a vault that never touched the THORChain screen, so it skips `defaultSelectedPositionsDialog()` and shows only the LP pool — RUNE/RUJI/TCY/sTCY/yRUNE/yTCY silently disappear from a screen the user never configured.

Suggest guarding this branch with `&& !isPoolOnlySelection` (or reordering so `isPoolOnlySelection` is computed first and excluded here) so pool-only selections go to Maya exclusively, matching the doc comment.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepositoryMigrationTest.kt
```diff
@@ -0,0 +1,204 @@
...
+
+    @Test
+    fun `unchecking both Cacao rows is a Maya choice, not an absent one`() = runTest {
+        // A Maya selection down to a pool holds no maya: key at all. Reading that as "never chose"
+        // would revive Bond and Staking and drop the pool the user had kept.
+        val before = preferencesWithLegacySelection(VAULT_ID, setOf("BTC.BTC"))
+
+        val after = migrateSelectedPositionsPerChain(before)
+
+        after[mayachainKey(VAULT_ID)] shouldBe setOf("BTC.BTC")
```
johnnyluo, line 134: This test asserts the Maya side of the pool-only case but not the THORChain side — which is exactly the assertion that would catch the bug noted on `DefiPositionsRepository.kt`. Every other branch in this file (e.g. `a pool the shape cannot attribute is kept on both chains`) checks both keys; this one should too:

```kotlin
after[thorchainKey(VAULT_ID)].shouldBeNull()
```

As written, this currently passes even though `thorchainKey(VAULT_ID)` also gets `{"BTC.BTC"}` written to it, contradicting the "MayaChain choice on its own" intent described above.

johnnyluo (dismissed):
Strong fix for the core clobbering bug and very well-documented migration, but there's a mismatch between the migration's documented intent and its actual behavior for pool-only legacy selections that isn't caught by the current tests. Left two inline comments with a concrete repro.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: Thanks — the doc line was genuinely misleading, and it is fixed in `81316370a`. The code is deliberate though, so let me put the reasoning next to it.

`on its own` meant "with no `maya:` key to vouch for it", not "exclusively MayaChain's". A pool is unattributable whichever set it arrives in — both screens write `CHAIN.TICKER` and both list `BTC.BTC` — so it is kept on both chains, which is the rule the neighbouring test `a pool the shape cannot attribute is kept on both chains` already pins. A pool-only set settles one narrower question: whether MayaChain chose at all. The KDoc now asks those two questions separately.

On the consequence, I checked what main showed this vault before the upgrade. `getSelectedPositions` on main is `preferences[key] ?: DEFAULT_POSITIONS`, and the key is present holding `{"BTC.BTC"}`, so the THORChain screen read exactly `["BTC.BTC"]` — `hasBondPositions` and `hasStakingPositions` both false, LP tab showing the pool. RUNE/RUJI/TCY/sTCY/yRUNE/yTCY were already not on that screen, so the migration is not taking them away; it is handing the screen back what it was showing.

Guarding with `&& !isPoolOnlySelection` would break the other authoring path for the same shape. A pool-only set is just as reachable from the THORChain dialog — uncheck every ticker, keep one LP pool — and that vault would then get its LP position written only to the Maya key, so the pool leaves the screen it was picked on, while the six defaults it had deliberately cleared come back. Reviving cleared rows is the failure this PR has been closing round after round, so I would rather keep the pool on both screens, where an unwanted card can be unchecked on the chain that did not want it.

I took your coverage point either way — the pool-only test now asserts both keys.

data/src/test/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepositoryMigrationTest.kt
aminsato: Agreed on the substance — that case should assert both keys like the ones around it, and it now does in `81316370a`.

I pinned it to the behaviour rather than to `shouldBeNull()`:

```kotlin
after[mayachainKey(VAULT_ID)] shouldBe setOf("BTC.BTC")
after[thorchainKey(VAULT_ID)] shouldBe setOf("BTC.BTC")
```

The reasoning is in the reply on `DefiPositionsRepository.kt`: the pool is unattributable in this set exactly as it is in every other, so both chains keep it, and `on its own` in the KDoc was about whether a pool proves MayaChain chose — not about the pool being MayaChain's alone. That sentence has been rewritten. With the assertion in place the rule is now stated in the test rather than left to the doc comment, so a future change in either direction has to argue with a failing test.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
```diff
@@ -1,49 +1,143 @@
...
+        val positions = (value as? Set<*>)?.filterIsInstance<String>()?.toSet()
+        if (vaultId.isEmpty() || positions == null) return@forEach
+
+        val mayaPositions = positions.filterTo(mutableSetOf()) { it.isMayaPositionKey() }
+        val thorchainPositions = positions - mayaPositions
+        val poolPositions = thorchainPositions.filterTo(mutableSetOf()) { it.isPoolPositionKey() }
+
+        // A set of only `maya:` keys says nothing about THORChain, and recording it as a clearing
+        // would blank a screen that was showing both its legs before the upgrade.
+        if (positions.isEmpty() || thorchainPositions.isNotEmpty()) {
```
rkokhatskyi: `thorchainPositions` holds pool keys, so an unattributable pool counts as THORChain evidence here while line 110 refuses that same pool as MayaChain evidence. A Maya first vault that ticked one LP pool stores `{maya:bond:CACAO, maya:stake:CACAO, BTC.BTC}` with no THORChain interaction, so THORChain gets `{BTC.BTC}`, skips `defaultSelectedPositionsDialog()` and hides a real bonded RUNE at `ThorchainDefiPositionsViewModel.kt:504`

```kotlin
val hasThorchainOwnKey = (thorchainPositions - poolPositions).isNotEmpty()
if (positions.isEmpty() || hasThorchainOwnKey || isPoolOnlySelection) {
    migrated[selectedPositionsKey(Chain.ThorChain, vaultId)] = thorchainPositions
}
```

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
rkokhatskyi: No, your reading is right and the carve-out should stay. `a vault that only ticked a THORChain pool` was loose wording for the `{RUNE, BTC.BTC}` shape you fixed, not for a pool-only set. Keeping a pool-only set on THORChain is correct for exactly the reason you gave Johnny.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
```diff
@@ -1,49 +1,143 @@
...
+        }
+
+        // A `maya:` key is the only attribution the shape carries. Pools are the exception the
+        // Maya dialog forces: unchecking both Cacao rows leaves a set with no `maya:` key at all,
+        // so a set of nothing but pools is a MayaChain choice too — a choice it shares with
+        // THORChain above, since the pool itself stays unattributable. Pools sitting next to plain
+        // tickers do not count — that set reads as THORChain's, and Maya's defaults beat a lone
+        // pool.
+        val isPoolOnlySelection = positions.isNotEmpty() && positions == poolPositions
+        if (positions.isEmpty() || mayaPositions.isNotEmpty() || isPoolOnlySelection) {
```
rkokhatskyi: Withholding the MayaChain key here also withholds the pool, so a Maya picked LP pool sitting next to plain tickers leaves the Maya LP tab empty where main showed it verbatim on the dotted branch. Writing the Maya defaults plus `poolPositions` for that shape keeps both

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: Confirmed and fixed in `8f2af6444`.

The asymmetry was real and I had it backwards on your side of it: `thorchainPositions` is everything that is not a `maya:` key, pools included, so a pool was standing as THORChain evidence in the same breath the Maya branch refused it. For `{maya:bond:CACAO, maya:stake:CACAO, BTC.BTC}` — reachable with the THORChain dialog never opened — THORChain took `{BTC.BTC}`, `hasBondPositions` stopped matching the Cacao key that had been keeping that leg alive on main, and the bonded RUNE went off the screen.

I took your snippet with one extension, which comes from your other comment rather than from me: withholding the key withholds the pool too, and if that pool was THORChain's, its LP tab empties. So instead of dropping it, a chain with none of its own keys in the set takes the pool next to its own defaults:

```kotlin
positions.isEmpty() -> emptySet()
positions == poolPositions -> poolPositions
ownKeys.isNotEmpty() -> ownKeys + poolPositions
poolPositions.isNotEmpty() -> defaults + poolPositions
else -> null
```

`ownKeys` is the `maya:` keys for MayaChain and the plain tickers for THORChain, so both sides run the same four branches — the rule is now stated once instead of twice. The pool-only shape stays the exception you named last round: it is the whole selection each screen had, so it goes across alone rather than next to the defaults.

The cost of the extension is that both screens' default sets are now spelled out in the migration, frozen at today's shape, so a vault migrated through that branch carries an explicit key and would not pick up a later change to a screen's defaults. I think that is the right trade for not losing a pool, but if you would rather keep your snippet as written and let the pool go, say so and I will drop the `defaults + poolPositions` branch.

Covered by `a Maya picked pool does not silence the Thorchain screen`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: Confirmed and fixed in `8f2af6444`, and I applied the same argument to the THORChain side in your other comment, since withholding that key drops the pool exactly the same way.

A chain with none of its own keys in the set now takes the pool alongside its own defaults rather than nothing:

```kotlin
poolPositions.isNotEmpty() -> defaults + poolPositions
```

So `{RUNE, BTC.BTC}` gives MayaChain `{maya:bond:CACAO, maya:stake:CACAO, BTC.BTC}` — the LP tab keeps the pool if it was Maya-picked, and the vault is not recorded as having chosen anything of its own on a screen it may never have opened. `a pool next to tickers is kept on Maya without being read as its choice` covers it, and the pool-only shape still goes across alone, so unchecking both Cacao rows is not undone by this.

One consequence worth flagging, since it is the price of writing defaults from :data: that vault now has an explicit MayaChain key holding the defaults as they are today, so a later change to `MAYA_DEFAULT_SELECTED_POSITIONS` will not reach it. The alternative was leaving the key unwritten and losing the pool, which is what you are pointing at here.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
aminsato: Heads up that this branch moved under your comment — the rule now runs through a single `selectionFor` for both chains, after @rkokhatskyi found the mirror image of your case: a pool was counting as THORChain evidence while being refused as MayaChain evidence.

Pools are now evidence for neither, and are dropped by neither: a chain with none of its own keys in the set takes the pool next to that chain's defaults. The pool-only shape you asked about is unchanged and still goes to both chains alone, for the reason in my reply above — pairing it with the defaults would hand back the Cacao rows that had to be unchecked to leave that shape behind. The doc comment that misread as "exclusively MayaChain's" is gone either way.

rkokhatskyi (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/repositories/DefiPositionsRepository.kt
rkokhatskyi: Keep the `defaults + poolPositions` branch. Dropping the pool loses a real LP position now, while the frozen defaults only cost a vault that never chose on that chain a later addition to the list — and any vault that taps Done is pinned the same way. The literals match `defaultSelectedPositionsDialog()` and `MAYA_DEFAULT_SELECTED_POSITIONS` exactly today.

PR #5546, opened 2026-08-09
fix(ui): align top bar back button and show Continue as loading

aminsato:
## Summary

Two fixes reported on the DeFi withdraw flow.

**Top bar padding.** `VsTopAppBar` dropped its back and action buttons straight into the Material3 icon slots, which only reserve 4dp, so the back button sat against the screen edge instead of the app's 16dp margin. It now applies the same 12dp inset `V2Topbar` uses, so both bars line up. Affects `VerifyDepositScreen` ("Function overview") and `VerifySignMessageScreen`.

**Continue button greyed out instead of loading.** Tapping a percentage chip runs the fraction calculation and then writes the amount, which arms a gas-fee recompute. Both windows disabled Continue with nothing to explain the wait — and the Circle withdraw screen hides the fee row entirely, so the shimmering fee wasn't even on screen. Continue now shows its loading indicator for that work.

- `isContinueLoading()` drives the spinner: submit in flight, percentage/MAX calculation, or the fee recompute that selection triggers.
- The fee window is gated on a new `hasAmountInput` flag, because `isGasFeeLoading` also holds its initial `true` before any amount exists — no estimate is armed there, so Continue stays plainly disabled until the user enters an amount rather than spinning forever on entry.
- `isAmountSelectionLoading` was added to `isContinueDisabled()` too. It keeps the button one colour across the whole wait instead of flipping blue → grey → blue, and closes a real hole: the soft-keyboard submit path could fire mid-calculation and sign the amount the chip was about to overwrite.

## Before / After

Top bar — back button at 4dp vs the app's 16dp margin:

| Before | After |
|--------|-------|
| before | after |

Continue while the fee recomputes after tapping 50%:

| Before | After |
|--------|-------|
| before | after |

## Test plan

- 7 new cases in `SendFormContinueGateTest` cover each window: percentage selection, the fee recompute with an amount present, the pre-entry state (disabled, no spinner), submit in flight, and an invalid recipient (disabled, no spinner).
- `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.send.*"` — 344 tests, 0 failures.
- Screenshots captured on the emulator via `PreviewActivity`; a `withdraw_usdc_circle` entry was added there for the Circle withdraw form.
- Worth a manual pass on Send and the TRON freeze/unfreeze screens, which share the same Continue gate.

## Summary by CodeRabbit

- **New Features**
  - Added a preview for withdrawing USDC through Circle, including account selection, amount entry, and fee loading states.
- **Bug Fixes**
  - Improved the Continue button behavior across send flows:
    - Prevents continuation when amount selection or required validation is incomplete.
    - Shows loading feedback during amount selection, fee updates, and submission.
    - Avoids displaying misleading loading indicators during initial fee estimation.
- **Style**
  - Added consistent spacing around top-bar navigation and action controls.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -28,6 +28,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
 import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.FilterQuality
@@ -138,7 +139,9 @@ import com.vultisig.wallet.ui.models.peer.PeerDiscoveryUiModel
 import com.vultisig.wallet.ui.models.qbtc.QbtcClaimMaturingUtxoUiModel
 import com.vultisig.wallet.ui.models.qbtc.QbtcClaimUiState
 import com.vultisig.wallet.ui.models.qbtc.QbtcClaimUtxoUiModel
+import com.vultisig.wallet.ui.models.send.AmountFraction
 import com.vultisig.wallet.ui.models.send.GasSettingsUiModel
+import com.vultisig.wallet.ui.models.send.SendFormUiModel
 import com.vultisig.wallet.ui.models.send.SendSrc
 import com.vultisig.wallet.ui.models.send.TokenBalanceUiModel
 import com.vultisig.wallet.ui.models.swap.LimitExpiryOption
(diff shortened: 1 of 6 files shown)


aminsato:
@coderabbitai Not applying this one — the guideline is real but the fix as scoped would make the file worse.

The comment covers lines 73-124, which is 6 of the 19 tests in `SendFormContinueGateTest.kt`. The other 13 already use `kotlin.test.assert*`, and the file imports it at the top. Converting only the new block leaves one file asserting two different ways, which is a worse outcome than either consistent choice.

Converting all 19 is the coherent version of the request, but this PR is a top bar alignment fix plus a Continue spinner — rewriting an unrelated 27-assertion test file would bury the change it is meant to show.

The surrounding code is also not ambiguous about which style is local. In `app/src/test/java/com/vultisig/wallet/ui/models/send`, 27 of the 28 test files use `kotlin.test`/JUnit assertions; `AddressManagerTest.kt` is the only Kotest one. Repo-wide it is 114 of 184 files in `app/src/test` and 268 of 295 in `data/src/test`. Kotest is available on the classpath, it is just not what this area uses, so the new tests match their neighbours.

Happy to do a standalone conversion pass if we want the guideline enforced for real, but it should be its own PR rather than a partial edit smuggled into this one.

johnnyluo (approved):
LGTM

PR #5552, opened 2026-08-09
fix(passcode): non-destructive credential reads, durable writes

### LINKED ISSUE #5549: [Fix] readCredentials() permanently wipes the passcode credentials on a read that merely failed [Passcode]
## Problem

`SharedPreferencesPasscodeStore.readCredentials()` calls the irreversible `clearCredentials()` from
inside a **read** path when exactly one of the two stored halves comes back null
(`PasscodeStore.kt:70`). It cannot tell "this key is absent" from "this key is present but did not
decrypt on this launch", because `EncryptingSharedPreferences.getString` swallows every decryption
failure and returns the default: `decryptOrNull` catches bare `Exception` and returns null
(`SecureSharedPreferences.kt:159-165`), and `readTyped` turns that into the caller's default
(`:174-175, :190-191`).

So a transient AndroidKeyStore fault on one of the two `decryptRaw` calls destroys the only record of
the data-encryption key that opens every at-rest-encrypted keyshare. The next launch finds ciphertext
with no credentials, resolves to `PasscodeState.KeyUnavailable` (`PasscodeRepository.kt:257`) and
tells the user to reinstall the app — permanent loss of every vault on the device. This codebase
already treats transient keystore faults as real: the 3-second lock timeout, `KeyStoreException`
routing and the in-memory fallback in `MainDataModule` exist for exactly that (#4403, #5106).

**The destructive line has no upside in any case.** Walk both:

- *Permanent corruption* (bad Base64, truncated payload, `AEADBadTagException`): `readCredentials`
  returns null with or without the clear, `resolveInitialState` yields `KeyUnavailable` either way,
  and the wrap is unrecoverable regardless. Clearing changes nothing.
- *Transient fault*: without the clear the credentials are still on disk and the next launch reads
  them fine. With the clear they are gone forever.

The comment justifying it — *"Drop the remaining half so the app reports 'no passcode' consistently
rather than leaving a stray salt that a later write would pair with a mismatched key"* — describes a
state nothing produces. `writeCredentials` writes both keys in one `edit` (`:76-81`) and
`clearCredentials` removes both (`:83-88`); no path writes one alone. `SharedPreferencesImpl` also
persists via a whole-file write plus backup-file rename, so a torn write cannot leave one key behind.
The repair defends against a case that cannot occur, and can only ever cause the loss it is meant to
tidy up after.

## Expected Behavior

A failed credential read reports that the credentials are unreadable. It never deletes them. The
existing `PasscodeState.StoreUnavailable` already carries exactly that meaning — "unreadable this
launch, restart, do not reinstall".

## Steps to Reproduce

Not reproducible on demand from the UI, because the trigger is a keystore fault. It is directly
observable in the store, which is where it should be tested:

1. Set a passcode so `passcode_salt` and `passcode_wrapped_data_key` are both written.
2. Make exactly one of the two values fail to decrypt on read (a test double over the
   `SharedPreferences` seam, or a corrupted single entry).
3. Call `readCredentials()`.
4. Both keys are now removed from the store. The passcode is unrecoverable and the next launch shows
   the reinstall dead end, even if the underlying fault was momentary.

## Solution

Delete the `clearCredentials()` call from `readCredentials()`. Reading must not mutate. If the
half-present case is worth distinguishing at all, distinguish it with `prefs.contains(key)` — which
is already documented to be true only when the value decrypts (`SecureSharedPreferences.kt:213-222`)
— and report the unreadable case as `StoreUnavailable` rather than repairing it destructively.

### Must Do
- [ ] Remove the `clearCredentials()` call from `readCredentials()` in
      `SharedPreferencesPasscodeStore`
- [ ] Replace the stale comment with what the code now guarantees: a read never destroys credentials
- [ ] Add a test asserting that a half-readable store leaves both underlying keys intact and simply
      reports no readable credentials
- [ ] Update any existing test that pins the old wiping behaviour

### Must NOT Do
- Don't change `writeCredentials` / `clearCredentials` semantics — `clearCredentials` is correct on
  the `disablePasscode` path
- Don't change unrelated code

## Acceptance Criteria
- [ ] A store where one of the two credential values fails to decrypt still holds both keys
      afterwards, and a later launch where both decrypt unlocks normally with the original passcode
- [ ] Setting, changing and disabling a passcode are unchanged

### LINKED ISSUE #5550: [Fix] The wrapped data key is written with apply(), so keyshare ciphertext outlives its key [Passcode]
## Problem

`setPasscode` makes the encrypted keyshares durable before the key that opens them. If the process
dies in between, every vault on the device is permanently unreadable.

The order in `PasscodeRepositoryImpl.setPasscode` is:

1. `store.writeLockout(cleared())` — `commit = true`, synchronous (`PasscodeStore.kt:101`)
2. `persistWrappedKey(key, passcode)` (`PasscodeRepository.kt:299`) → `writeCredentials`, which uses
   `prefs.edit { }` (`PasscodeStore.kt:77`) — androidx `edit {}` defaults to `commit = false` and
   calls `apply()`, so the disk write is deferred to `QueuedWork`
3. `protectAllOrLog(key)` (`:304`) → `protectAll` encrypts every keyshare and writes them in one
   Room `@Upsert` transaction, which is durable when it returns

`EncryptingSharedPreferences.edit()` is a straight passthrough to the real editor
(`SecureSharedPreferences.kt:225`), over a file-backed `SharedPreferencesImpl`
(`MainDataModule.kt:144-145`), so nothing upstream makes the write synchronous. `apply()` survives
lifecycle transitions — `QueuedWork.waitToFinish()` runs at activity pause/stop — but **not** a
process kill: an uncaught exception on any thread, or a low-memory kill, drops the queued write while
the Room transaction is already on disk.

Result: ciphertext with no wrap. The next launch resolves to `PasscodeState.KeyUnavailable` and tells
the user to reinstall. This is the same "credentials that do not survive the launch" loss the
`isPersistent()` gate at `PasscodeRepository.kt:269` was added to prevent, reached by a different
route on a perfectly persistent store.

The asymmetry is the giveaway: the **failed-attempt counter** is committed synchronously, with a
comment explaining why durability matters there, while the **key protecting every vault** is not.

The window is short — `apply()` typically flushes within milliseconds — so this is not a
high-frequency failure. It is filed because the consequence is unrecoverable, the fix is one
parameter, and the file already establishes the correct pattern one method away.

## Expected Behavior

The wrapped data key is on disk before any ciphertext encrypted under it is. A process kill at any
point during `setPasscode` leaves either "no passcode and plaintext shares" or "passcode and
recoverable shares", never "ciphertext and no key".

## Steps to Reproduce

Timing-dependent, so best pinned as a store-level test rather than by hand:

1. Enable the passcode feature, set a passcode on a vault that has keyshares.
2. Kill the process (`adb shell am kill`, or an uncaught exception) in the window between
   `protectAll`'s Room transaction committing and `QueuedWork` flushing the queued prefs write.
3. Relaunch. The keyshares are `vlpc1:`-encrypted, the credentials are absent, and the app shows
   "Vaults unavailable — reinstall the app".

## Solution

Use `prefs.edit(commit = true)` in `writeCredentials`, matching `writeLockout` directly below it, so
the wrap is durable before `protectAll` runs. Both already execute on `@DefaultDispatcher` inside
`withContext(dispatcher)` (`PasscodeRepository.kt:490-494`), never on the main thread, so the
synchronous write does not risk an ANR. `clearCredentials` is worth the same treatment for symmetry,
though its failure mode is benign: a lost clear leaves the passcode configured over plaintext shares,
which the next unlock's `protectAll` sweep re-encrypts correctly.

### Must Do
- [ ] Change `writeCredentials` to `prefs.edit(commit = true) { ... }`
- [ ] Change `clearCredentials` to match
- [ ] State in the comment why credential writes are synchronous: ciphertext written under this key
      becomes durable immediately afterwards

### Must NOT Do
- Don't move these calls onto the main thread or add a dispatcher hop — they already run on
  `@DefaultDispatcher` via their callers
- Don't change the ordering inside `setPasscode`; the wrap-before-encrypt sequence is already correct
- Don't change unrelated code

## Acceptance Criteria
- [ ] Setting, changing and disabling a passcode all still work, and a vault signs afterwards
- [ ] No new main-thread disk I/O (no StrictMode disk-write violation on the passcode flows)

rkokhatskyi:
Closes #5549
Closes #5550

Both changes are in `SharedPreferencesPasscodeStore`, so they ship together rather than conflicting on rebase.

**A failed read no longer deletes the credentials (#5549).** `readCredentials()` called `clearCredentials()` whenever exactly one of the two stored halves came back null. `EncryptingSharedPreferences.getString` swallows a decryption failure into null, so a transient AndroidKeyStore fault on one of the two values destroyed the only record of the data key that opens every at-rest-encrypted keyshare — the next launch resolved to `KeyUnavailable` and told the user to reinstall. The repair also guarded a state nothing produces: `writeCredentials` writes both keys in one editor and `clearCredentials` removes both, so no path leaves a stray half behind. A read now reports unreadable credentials by returning nothing.

**Credential writes are durable, and a refused write changes nothing (#5550).** `setPasscode` persists the wrap and then encrypts every keyshare in one Room transaction, so a queued `apply()` dropped by a process kill left durable ciphertext with no key. Both credential mutations now commit synchronously through the editor rather than androidx's `edit(commit = true)`, which discards the `Boolean`.

Reporting the refusal was not enough on its own. `SharedPreferencesImpl.commit` runs `commitToMemory()` before it writes the file and leaves the map updated when that write fails, while `writeToFile` deletes the partly-written file and leaves the backup for the next load to restore. The in-memory copy was therefore the only thing that had diverged, and it made the recovery paths lie: a failed change accepted only the new passcode until the next launch, and a failed disable rejected the correct one outright. Both mutations now put the two keys back before throwing.

`setPasscode` already turned that failure into `Failed` before encrypting anything; `changePasscode` and `disablePasscode` now do the same instead of letting it escape into a `viewModelScope.launch`. `disablePasscode` decrypts the shares before dropping the credentials, so its failure path reseals them under the same key, inside `NonCancellable` so a cancellation cannot leave them in the clear.

`writeLockout` keeps `edit(commit = true)`: nothing downstream depends on that write having landed, so a lost attempt counter costs one uncharged attempt rather than the keyshares.

iOS lands on the same two rules: `KeychainReadResult` keeps "absent" apart from "unreadable" so a read failure never deletes, and `storeWrappedDataKey` persists the wrapper synchronously — and verifies it by read-back — before any share is sealed.

## Testing

```
./gradlew :data:testDebugUnitTest    # 2734 pass, 0 failures
./gradlew :data:ktfmtCheck :data:lintDebug
```

The three tests that pinned the old wiping behaviour now assert the store is left untouched, and new tests cover a half-readable store that recovers on the next read, a refused write and a refused clear over a store that applies edits to memory and then reports the disk write as refused, and the three repository paths that report them. Checked against the pre-change store: the read tests and both refusal tests fail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved incomplete or malformed credential data instead of deleting it during reads.
  * Improved reliability when credential data is temporarily unavailable, allowing successful retries.
  * Ensured credential saves and clears are completed immediately and reliably.
  * Improved passcode setup, change, and disable behavior when credential persistence fails.
  * Prevented failed operations from leaving passcode data in an inconsistent state and provided clearer failure results.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
@@ -356,8 +356,8 @@ internal class PasscodeRepositoryImpl(
                 // The data key is unchanged, so stored keyshares stay valid: only the wrap is
                 // rewritten. This is why changing the passcode is instant on a large vault set.
                 //
-                // Nothing has changed yet at this point, so a wrap that never reached the disk
-                // simply leaves the old passcode in force.
+                // Nothing else has been touched yet, so a wrap that did not land leaves the old
+                // passcode in force.
                 val persisted = runCatching { persistWrappedKey(key, newPasscode) }
                 persisted.exceptionOrNull()?.let { cause ->
                     if (cause is CancellationException) throw cause
@@ -391,10 +391,6 @@ internal class PasscodeRepositoryImpl(
                 // One unit, uncancellable. Stopping between the two leaves the key live and the
                 // state Unlocked with no wrap on disk: every keyshare written afterwards would be
                 // sealed under a key the next launch has no way to recover.
-                //
-                // A clear that never reached the disk throws before any of this, so the passcode
-                // stays in force — but unprotectAll has already put every share back in the
-                // clear, so they have to be sealed again before the failure is reported.
                 val cleared = runCatching {
                     withContext(NonCancellable) {
                         withContext(dispatcher) { store.clearCredentials() }
@@ -409,6 +405,7 @@ internal class PasscodeRepositoryImpl(
                         cause,
                         "Refusing to disable the passcode: the credentials are still there",
                     )
+                    // The shares are already back in the clear under a passcode still in force.
                     protectAllOrLog(key)
                     key.fill(0)
                     return@verifyLocked PasscodeUnlockResult.Failed
--- a/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
@@ -356,12 +356,15 @@ internal class PasscodeRepositoryImpl(
                 // The data key is unchanged, so stored keyshares stay valid: only the wrap is
                 // rewritten. This is why changing the passcode is instant on a large vault set.
                 //
-                // Nothing else has been touched yet, so a wrap that did not land leaves the old
-                // passcode in force.
-                val persisted = runCatching { persistWrappedKey(key, newPasscode) }
-                persisted.exceptionOrNull()?.let { cause ->
-                    if (cause is CancellationException) throw cause
-                    Timber.e(cause, "Failed to store the new passcode")
+                // Nothing else has been touched yet, and the store puts back what it had, so a
+                // wrap that did not land leaves the old passcode in force.
+                try {
+                    persistWrappedKey(key, newPasscode)
+                } catch (e: CancellationException) {
+                    key.fill(0)
+                    throw e
+                } catch (e: Exception) {
+                    Timber.e(e, "Failed to store the new passcode")
                     key.fill(0)
                     return@verifyLocked PasscodeUnlockResult.Failed
                 }
@@ -391,23 +394,25 @@ internal class PasscodeRepositoryImpl(
                 // One unit, uncancellable. Stopping between the two leaves the key live and the
                 // state Unlocked with no wrap on disk: every keyshare written afterwards would be
                 // sealed under a key the next launch has no way to recover.
-                val cleared = runCatching {
+                try {
                     withContext(NonCancellable) {
                         withContext(dispatcher) { store.clearCredentials() }
                         swapDataKey(null)
-                        key.fill(0)
                         _state.value = PasscodeState.Disabled
+                        key.fill(0)
                     }
-                }
-                cleared.exceptionOrNull()?.let { cause ->
-                    if (cause is CancellationException) throw cause
-                    Timber.e(
-                        cause,
-                        "Refusing to disable the passcode: the credentials are still there",
-                    )
-                    // The shares are already back in the clear under a passcode still in force.
-                    protectAllOrLog(key)
+                } catch (e: CancellationException) {
                     key.fill(0)
+                    throw e
+                } catch (e: Exception) {
+                    Timber.e(e, "Refusing to disable the passcode: the credentials are still there")
+                    // unprotectAll has already put every share back in the clear, and the passcode
+                    // that guards them is still in force. Uncancellable because they stay exposed
+                    // until this finishes.
+                    withContext(NonCancellable) {
+                        protectAllOrLog(key)
+                        key.fill(0)
+                    }
                     return@verifyLocked PasscodeUnlockResult.Failed
                 }
                 PasscodeUnlockResult.Success
--- a/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
@@ -355,7 +355,16 @@ internal class PasscodeRepositoryImpl(
             verifyLocked(currentPasscode) { key ->
                 // The data key is unchanged, so stored keyshares stay valid: only the wrap is
                 // rewritten. This is why changing the passcode is instant on a large vault set.
-                persistWrappedKey(key, newPasscode)
+                //
+                // Nothing has changed yet at this point, so a wrap that never reached the disk
+                // simply leaves the old passcode in force.
+                val persisted = runCatching { persistWrappedKey(key, newPasscode) }
+                persisted.exceptionOrNull()?.let { cause ->
+                    if (cause is CancellationException) throw cause
+                    Timber.e(cause, "Failed to store the new passcode")
+                    key.fill(0)
+                    return@verifyLocked PasscodeUnlockResult.Failed
+                }
                 swapDataKey(key)
                 publishUnlockedUnlessLocked()
                 PasscodeUnlockResult.Success
@@ -382,11 +391,26 @@ internal class PasscodeRepositoryImpl(
                 // One unit, uncancellable. Stopping between the two leaves the key live and the
                 // state Unlocked with no wrap on disk: every keyshare written afterwards would be
                 // sealed under a key the next launch has no way to recover.
-                withContext(NonCancellable) {
-                    withContext(dispatcher) { store.clearCredentials() }
-                    swapDataKey(null)
+                //
+                // A clear that never reached the disk throws before any of this, so the passcode
+                // is still in place — over shares unprotectAll has already put back in the clear,
+                // which the next unlock re-encrypts.
+                val cleared = runCatching {
+                    withContext(NonCancellable) {
+                        withContext(dispatcher) { store.clearCredentials() }
+                        swapDataKey(null)
+                        key.fill(0)
+                        _state.value = PasscodeState.Disabled
+                    }
+                }
+                cleared.exceptionOrNull()?.let { cause ->
+                    if (cause is CancellationException) throw cause
+                    Timber.e(
+                        cause,
+                        "Refusing to disable the passcode: the credentials are still there",
+                    )
                     key.fill(0)
-                    _state.value = PasscodeState.Disabled
+                    return@verifyLocked PasscodeUnlockResult.Failed
                 }
                 PasscodeUnlockResult.Success
             }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
@@ -393,8 +393,8 @@ internal class PasscodeRepositoryImpl(
                 // sealed under a key the next launch has no way to recover.
                 //
                 // A clear that never reached the disk throws before any of this, so the passcode
-                // is still in place — over shares unprotectAll has already put back in the clear,
-                // which the next unlock re-encrypts.
+                // stays in force — but unprotectAll has already put every share back in the
+                // clear, so they have to be sealed again before the failure is reported.
                 val cleared = runCatching {
                     withContext(NonCancellable) {
                         withContext(dispatcher) { store.clearCredentials() }
@@ -409,6 +409,7 @@ internal class PasscodeRepositoryImpl(
                         cause,
                         "Refusing to disable the passcode: the credentials are still there",
                     )
+                    protectAllOrLog(key)
                     key.fill(0)
                     return@verifyLocked PasscodeUnlockResult.Failed
                 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
@@ -72,21 +72,36 @@ constructor(private val prefs: SharedPreferences) : PasscodeStore {
     }
 
     override fun writeCredentials(credentials: PasscodeCredentials) {
-        // The wrap has to be on disk before setPasscode seals the keyshares under it. androidx's
-        // edit(commit = true) discards commit()'s result, so the editor is driven directly.
-        val committed =
-            prefs
-                .edit()
-                .putString(KEY_SALT, encode(credentials.salt))
-                .putString(KEY_WRAPPED_KEY, encode(credentials.wrappedDataKey))
-                .commit()
-        check(committed) { "Passcode credentials were not written to disk" }
+        // The wrap has to be on disk before setPasscode seals the keyshares under it.
+        commitBothHalves("Passcode credentials were not written to disk") {
+            putString(KEY_SALT, encode(credentials.salt))
+            putString(KEY_WRAPPED_KEY, encode(credentials.wrappedDataKey))
+        }
     }
 
     override fun clearCredentials() {
         // A clear that did not land brings back a passcode the user removed.
-        val committed = prefs.edit().remove(KEY_SALT).remove(KEY_WRAPPED_KEY).commit()
-        check(committed) { "Passcode credentials were not removed from disk" }
+        commitBothHalves("Passcode credentials were not removed from disk") {
+            remove(KEY_SALT)
+            remove(KEY_WRAPPED_KEY)
+        }
+    }
+
+    /**
+     * Applies [mutation] to both halves, or puts back what they held and throws [failure].
+     *
+     * `commit` updates the in-memory map before it writes the file and leaves it updated when that
+     * write fails, while the file keeps its previous contents. Putting the map back is what makes a
+     * refused write change nothing.
+     */
+    private fun commitBothHalves(failure: String, mutation: SharedPreferences.Editor.() -> Unit) {
+        val salt = prefs.getString(KEY_SALT, null)
+        val wrapped = prefs.getString(KEY_WRAPPED_KEY, null)
+        val editor = prefs.edit()
+        editor.mutation()
+        if (editor.commit()) return
+        prefs.edit().putString(KEY_SALT, salt).putString(KEY_WRAPPED_KEY, wrapped).apply()
+        error(failure)
     }
 
     override fun readLockout(): PasscodeLockoutState =
--- a/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
@@ -37,8 +37,10 @@ internal interface PasscodeStore {
 
     fun readCredentials(): PasscodeCredentials?
 
+    /** @throws IllegalStateException if the credentials did not reach the disk. */
     fun writeCredentials(credentials: PasscodeCredentials)
 
+    /** @throws IllegalStateException if the credentials are still on the disk afterwards. */
     fun clearCredentials()
 
     fun readLockout(): PasscodeLockoutState
@@ -70,20 +72,24 @@ constructor(private val prefs: SharedPreferences) : PasscodeStore {
     }
 
     override fun writeCredentials(credentials: PasscodeCredentials) {
-        // commit, not apply: setPasscode durably seals every keyshare under this key straight
-        // afterwards, so an apply() lost to a process kill leaves ciphertext with no wrap.
-        prefs.edit(commit = true) {
-            putString(KEY_SALT, encode(credentials.salt))
-            putString(KEY_WRAPPED_KEY, encode(credentials.wrappedDataKey))
-        }
+        // commit, not apply, and the result is checked: setPasscode durably seals every keyshare
+        // under this key straight afterwards, so a write that was only queued — or refused
+        // outright — leaves ciphertext with no key. androidx's edit(commit = true) discards
+        // commit()'s result.
+        val committed =
+            prefs
+                .edit()
+                .putString(KEY_SALT, encode(credentials.salt))
+                .putString(KEY_WRAPPED_KEY, encode(credentials.wrappedDataKey))
+                .commit()
+        check(committed) { "Passcode credentials were not written to disk" }
     }
 
     override fun clearCredentials() {
-        // Committed too: a lost clear brings back a passcode the user removed.
-        prefs.edit(commit = true) {
-            remove(KEY_SALT)
-            remove(KEY_WRAPPED_KEY)
-        }
+        // Checked for the same reason: a clear that did not land brings back a passcode the user
+        // removed.
+        val committed = prefs.edit().remove(KEY_SALT).remove(KEY_WRAPPED_KEY).commit()
+        check(committed) { "Passcode credentials were not removed from disk" }
     }
 
     override fun readLockout(): PasscodeLockoutState =
--- a/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
@@ -65,17 +65,15 @@ constructor(private val prefs: SharedPreferences) : PasscodeStore {
     override fun readCredentials(): PasscodeCredentials? {
         val salt = prefs.getString(KEY_SALT, null)?.let(::decodeOrNull)
         val wrapped = prefs.getString(KEY_WRAPPED_KEY, null)?.let(::decodeOrNull)
-        // A read never deletes. A half that failed to decrypt this launch reads exactly like one
-        // that was never stored, and the wrap it pairs with opens every encrypted keyshare.
+        // A half that failed to decrypt is indistinguishable from one that was never stored, so it
+        // is never grounds for deleting the other.
         if (salt == null || wrapped == null) return null
         return PasscodeCredentials(salt = salt, wrappedDataKey = wrapped)
     }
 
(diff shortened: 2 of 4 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
```diff
@@ -63,25 +63,24 @@ constructor(private val prefs: SharedPreferences) : PasscodeStore {
...
     }
 
     override fun writeCredentials(credentials: PasscodeCredentials) {
-        prefs.edit {
+        // commit, not apply: setPasscode durably seals every keyshare under this key straight
+        // afterwards, so an apply() lost to a process kill leaves ciphertext with no wrap.
+        prefs.edit(commit = true) {
             putString(KEY_SALT, encode(credentials.salt))
             putString(KEY_WRAPPED_KEY, encode(credentials.wrappedDataKey))
         }
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed in 4329e818b. Both credential mutations now go through the editor directly and `check()` the `Boolean` that `commit()` returns, so a refused write throws instead of being reported as stored. `setPasscode` already caught that and returns `Failed` before `protectAll` runs, and `changePasscode`/`disablePasscode` now do the same rather than letting it escape into a `viewModelScope.launch` — covered by two store tests and three repository tests. `writeLockout` is deliberately left as it is: nothing downstream depends on that write having landed, so a lost attempt counter costs one uncharged attempt rather than the keyshares, and failing an unlock outright on a flaky store would lock the owner out of their own app.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
```diff
@@ -382,11 +391,26 @@ internal class PasscodeRepositoryImpl(
...
+                }
+                cleared.exceptionOrNull()?.let { cause ->
+                    if (cause is CancellationException) throw cause
+                    Timber.e(
+                        cause,
+                        "Refusing to disable the passcode: the credentials are still there",
+                    )
                     key.fill(0)
-                    _state.value = PasscodeState.Disabled
+                    return@verifyLocked PasscodeUnlockResult.Failed
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed in e9aa5d792. That path now calls `protectAllOrLog(key)` before zeroing the key and returning `Failed`, so the shares are sealed again under the same data key the still-present credentials wrap. The test asserts the full sequence is protect, unprotect, protect. The sibling branch above it needs nothing: `unprotectAll` decrypts everything in memory and throws before the upsert, so a share that will not open leaves the table untouched.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
```diff
@@ -63,28 +65,28 @@ constructor(private val prefs: SharedPreferences) : PasscodeStore {
...
-        }
+        // The wrap has to be on disk before setPasscode seals the keyshares under it. androidx's
+        // edit(commit = true) discards commit()'s result, so the editor is driven directly.
+        val committed =
+            prefs
+                .edit()
+                .putString(KEY_SALT, encode(credentials.salt))
+                .putString(KEY_WRAPPED_KEY, encode(credentials.wrappedDataKey))
+                .commit()
+        check(committed) { "Passcode credentials were not written to disk" }
```
aminsato: A failed `commit()` does not mean nothing was written. `SharedPreferencesImpl.commitToMemory()` applies both values to the in-memory map *before* the disk write is attempted, so after this `check` throws, `prefs.getString(KEY_SALT/KEY_WRAPPED_KEY)` returns the new values for the rest of the process.

Concretely: disk is full, user taps "Set passcode". `writeCredentials` throws, `setPasscode` catches and returns `Failed` without encrypting anything — correct with respect to disk. But the next attempt hits the `store.readCredentials() != null` guard at `PasscodeRepository.kt:276` and logs "Refusing to set a passcode: one is already configured", returning `Failed` forever until the app is force-restarted, with no message explaining why.

The `changePasscode` case is worse: the comment promises "a wrap that did not land leaves the old passcode in force", but the in-memory prefs hold the **new** salt/wrap, so after a reported failure the app accepts only the new passcode (the old one returns `Wrong`), then silently flips back to the old one after a restart.

To make the recovery paths true, the store needs to restore the previous in-memory values when the commit fails.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
```diff
@@ -63,28 +65,28 @@ constructor(private val prefs: SharedPreferences) : PasscodeStore {
...
     }
 
     override fun clearCredentials() {
-        prefs.edit {
-            remove(KEY_SALT)
-            remove(KEY_WRAPPED_KEY)
-        }
+        // A clear that did not land brings back a passcode the user removed.
+        val committed = prefs.edit().remove(KEY_SALT).remove(KEY_WRAPPED_KEY).commit()
+        check(committed) { "Passcode credentials were not removed from disk" }
```
aminsato: Mirror image of the `writeCredentials` issue: a false `commit()` means the removal *already took effect in memory*, so `disablePasscode`'s "the credentials are still there" recovery is acting on a false premise.

When the disk write fails during "Turn off passcode": `clearCredentials` throws, `disablePasscode` re-encrypts the keyshares with `protectAllOrLog(key)` and leaves `state = Unlocked` — but the in-memory prefs no longer contain `KEY_SALT`/`KEY_WRAPPED_KEY`. The user backgrounds the app, `lock()` flips to `Locked`, and `verifyLocked` (`PasscodeRepository.kt:443`) gets `null` from `readCredentials()` and returns `Wrong(ATTEMPTS_BEFORE_LOCKOUT)` for the correct passcode. The user is locked out of their own wallet — every entry reads "wrong passcode" — until they force-kill and relaunch, at which point the on-disk credentials reappear.

data/src/test/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepositoryImplTest.kt
```diff
@@ -629,11 +674,13 @@ internal class FakePasscodeStore : PasscodeStore {
 
     @Synchronized
     override fun writeCredentials(credentials: PasscodeCredentials) {
+        writeFailure?.let { throw it }
```
aminsato, line 678: The fake throws *before* mutating, so these tests assert an all-or-nothing store that `SharedPreferencesPasscodeStore` does not provide.

`changePasscode leaves the old passcode in force when the new wrap is not stored` asserts `assertEquals(credentials, store.readCredentials())` after the failed write, and `disablePasscode reseals the keyshares when the credentials are not removed` asserts the credentials survive the failed clear. Both pass only because of the ordering here. Back the store with a real `SharedPreferences` whose `commit()` returns false (memory updated, disk not) and both assertions fail.

So the regression this suite is meant to lock in — "a write that did not land changes nothing" — is untested, which is why the lockout-on-correct-passcode failures flagged in `PasscodeStore.kt` ship green. Mutating first and then throwing would make the fake match production and turn these into real regression tests.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
```diff
@@ -382,11 +391,24 @@ internal class PasscodeRepositoryImpl(
...
+                    }
+                }
+                cleared.exceptionOrNull()?.let { cause ->
+                    if (cause is CancellationException) throw cause
+                    Timber.e(
+                        cause,
+                        "Refusing to disable the passcode: the credentials are still there",
+                    )
+                    // The shares are already back in the clear under a passcode still in force.
+                    protectAllOrLog(key)
```
aminsato: The re-seal runs *outside* `withContext(NonCancellable)` — the block closes at line 401 and the handler starts at 402 — so a cancellation arriving while the shares sit decrypted aborts the recovery instead of completing it.

`disablePasscode` runs in a `viewModelScope`. `unprotectAll()` has decrypted every vault keyshare to plaintext in Room; `store.clearCredentials()` then throws; control reaches `protectAllOrLog(key)`, whose first suspension point throws `CancellationException` if the settings screen was destroyed in the meantime. `protectAllOrLog` rethrows `CancellationException` by design, so the re-encryption never happens and `key.fill(0)` on the next line is skipped. Every keyshare is left in the clear on disk while a passcode is still configured — the exact at-rest exposure this feature prevents — until the user happens to complete a successful unlock, the only other caller of the sweep. The plaintext data key also stays resident unzeroed.

Pulling the handler's recovery inside the `NonCancellable` block would close this.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
```diff
@@ -355,7 +355,16 @@ internal class PasscodeRepositoryImpl(
             verifyLocked(currentPasscode) { key ->
                 // The data key is unchanged, so stored keyshares stay valid: only the wrap is
                 // rewritten. This is why changing the passcode is instant on a large vault set.
-                persistWrappedKey(key, newPasscode)
+                //
+                // Nothing else has been touched yet, so a wrap that did not land leaves the old
+                // passcode in force.
+                val persisted = runCatching { persistWrappedKey(key, newPasscode) }
+                persisted.exceptionOrNull()?.let { cause ->
+                    if (cause is CancellationException) throw cause
```
aminsato: This rethrow exits before `key.fill(0)` on line 365, so the recovered plaintext data key is never zeroed on the cancellation path.

If `changePasscode` is cancelled (user backgrounds the app / leaves the screen) while `persistWrappedKey` is writing, the unwrapped 32-byte AES data key stays live in a heap `ByteArray` that nothing owns — `swapDataKey(key)` never ran, so `lock()` cannot zero it either. It survives until GC, against the file's own invariant that key material is zeroed on every exit path; compare `unlock()`'s `try`/`finally` at 425-428, which does zero it. Same omission in the `disablePasscode` handler at line 403.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
```diff
@@ -382,11 +391,24 @@ internal class PasscodeRepositoryImpl(
...
                 // state Unlocked with no wrap on disk: every keyshare written afterwards would be
                 // sealed under a key the next launch has no way to recover.
-                withContext(NonCancellable) {
-                    withContext(dispatcher) { store.clearCredentials() }
-                    swapDataKey(null)
+                val cleared = runCatching {
+                    withContext(NonCancellable) {
+                        withContext(dispatcher) { store.clearCredentials() }
+                        swapDataKey(null)
+                        key.fill(0)
```
aminsato: Zeroing the key here, inside the guarded block, while the failure handler at 402 re-encrypts every keyshare with that same array, encodes an invariant nothing enforces.

`key.fill(0)` runs before `_state.value = PasscodeState.Disabled` on the next line, and the handler calls `protectAllOrLog(key)` at 409. If any statement after this line ever throws, every keyshare in every vault gets sealed under 32 zero bytes — valid AES key material that no passcode can reopen, i.e. permanent silent vault loss with no error shown. Today only `clearCredentials()` (the first statement) throws, so it is latent rather than live, but the guard costs one line: move `key.fill(0)` out of the block to a single site after the handler.

The new test at `PasscodeRepositoryImplTest.kt:343` asserts only the call sequence `listOf("protect","unprotect","protect")` and never that `protection.protectedWith` still equals the original key, so it cannot catch this.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
```diff
@@ -382,11 +391,24 @@ internal class PasscodeRepositoryImpl(
                 // One unit, uncancellable. Stopping between the two leaves the key live and the
                 // state Unlocked with no wrap on disk: every keyshare written afterwards would be
                 // sealed under a key the next launch has no way to recover.
-                withContext(NonCancellable) {
-                    withContext(dispatcher) { store.clearCredentials() }
-                    swapDataKey(null)
+                val cleared = runCatching {
```
aminsato: `runCatching`'s catch is `catch (e: Throwable)`, so `Error`s get downgraded to `PasscodeUnlockResult.Failed`.

An `OutOfMemoryError` or `StackOverflowError` raised inside `persistWrappedKey` (361) or `store.clearCredentials()` (396) is logged as "Failed to store the new passcode" / "the credentials are still there" and returned as `Failed`. The user sees a routine "could not save" and keeps using a process that is out of memory — and in the disable case the handler then runs a full `protectAllOrLog` re-encryption sweep over every vault in that degraded process. `setPasscode`'s `catch (e: Exception)` at line 308 gets this right for the identical operation.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
```diff
@@ -355,7 +355,16 @@ internal class PasscodeRepositoryImpl(
             verifyLocked(currentPasscode) { key ->
                 // The data key is unchanged, so stored keyshares stay valid: only the wrap is
                 // rewritten. This is why changing the passcode is instant on a large vault set.
-                persistWrappedKey(key, newPasscode)
+                //
+                // Nothing else has been touched yet, so a wrap that did not land leaves the old
+                // passcode in force.
+                val persisted = runCatching { persistWrappedKey(key, newPasscode) }
```
aminsato: This is the third copy of the same six-line shape — `runCatching { … }.exceptionOrNull()?.let { rethrow-cancel; Timber.e; key.fill(0); return Failed }` — alongside the one already at 384-390 and the new one at 394-412. The same file also handles the identical concern with a second idiom: `try { } catch (e: CancellationException) { throw e } catch (e: Exception) { }` at `setPasscode:306`, `initialize:230`, and `protectAllOrLog:331`.

The cost is concrete rather than stylistic: the `key.fill(0)` obligation is hand-copied at each exit and, as a direct result, was omitted on both new rethrow paths (see the comment at 363). A private `suspend fun failOnError(key: ByteArray, message: String, block: suspend () -> Unit)` — or matching `setPasscode`'s `try`/`catch` with a shared `finally` — expresses it once.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
rkokhatskyi: Right, and confirmed against AOSP rather than taken on trust: `commit()` runs `commitToMemory()` before `enqueueDiskWrite` and returns `mcr.writeToDiskResult`, so the map keeps the new values when the write fails. Fixed in 02f84269b — both mutations go through `commitBothHalves`, which snapshots the two keys, and on a refused commit puts them back before throwing. The file itself needs no undo: `writeToFile` deletes the partly-written file and leaves the backup, which the next load restores, so only the in-memory copy had diverged.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeStore.kt
rkokhatskyi: Same fix covers this one, since the clear goes through `commitBothHalves` too — the removal is put back in memory before the throw, so `readCredentials()` keeps returning the credentials and the correct passcode still unlocks. Worth noting the recovery you describe was doing the right thing on the wrong premise rather than the wrong thing: re-encrypting the shares is still correct, it just needed the credentials to actually still be readable for the state to hang together.

data/src/test/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepositoryImplTest.kt
rkokhatskyi: Fair, and the fake is now faithful rather than lenient — the store really is all-or-nothing, so throwing before mutating is what production does. Added two tests that pin it at the seam you pointed at: `a write the disk refused is reported and changes nothing` and `a clear the disk refused is reported and changes nothing` run over a real `InMemorySharedPreferences` behind a `RefusingPreferences` decorator that applies the edit to memory and then reports the disk write as refused, exactly like `SharedPreferencesImpl`. Both fail against the previous commit with a state mismatch, not a missing throw. Building that decorator turned up a trap worth flagging: `SharedPreferences.Editor by editor` delegates `putString`, so a chained call returns the inner editor and escapes the wrapper — the mutators return `this` now.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
rkokhatskyi: Fixed in 02f84269b — the recovery moved inside `withContext(NonCancellable)`, so the re-encryption and the zeroing both complete. One correction to the impact: the exposure ends at the next unlock rather than being open-ended, because the passcode is still configured so the user has to unlock to use the app at all and `unlock` runs the same sweep. Still the wrong window to leave open, so the fix stands.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
rkokhatskyi: Fixed — `key.fill(0)` now runs before the rethrow on both paths. One thing your reasoning got slightly wrong though: the file has no invariant here to violate. The sibling handler directly above (`unprotectAll`, line 385) rethrows the same way without zeroing, and pre-diff `changePasscode` did not zero on any throw path either, so this was the established shape rather than a lapse from it. Taking it anyway because it costs a line reorder.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
rkokhatskyi: Fixed by making `key.fill(0)` the last statement in the block, so nothing that can throw follows it and the handler always gets a live key. Agreed it was latent rather than live — `_state.value = ...` was the only statement after it — but the handler I added is what made the array load-bearing after that point, so the ordering had to change with it. Also added `assertContentEquals(repository.dataKeyOrNull(), protection.protectedWith)` to the reseal test, which is the assertion that would have caught the zeroed-key version.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
rkokhatskyi: Fixed — both are plain `try`/`catch (e: CancellationException)` + `catch (e: Exception)` now, so an `Error` propagates instead of being reported as `Failed`. This also matches the idiom the file already uses three times (`initialize`, `setPasscode`, `protectAllOrLog`) for the same concern, so it is one less shape rather than one more.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
rkokhatskyi: Half of this is fixed and half I am declining. The concrete cost you named — the hand-copied `key.fill(0)` obligation being dropped — is real and is now handled at every exit on both new paths, and switching them to `try`/`catch` removes one of the two competing idioms rather than adding a third. I am not extracting a `failOnError` helper on top of that: the file expresses this concern with a plain `try`/`catch` in three existing places, so a wrapper would be the odd one out at five sites, and it would have to take the key, the message and a suspend block to earn its keep. The centralisation that would actually pay is zeroing the key once in `verifyLocked` rather than in each `onVerified` body, but that needs `changePasscode` to stop taking ownership of the array via `swapDataKey(key)`, which is a change to the key-lifetime contract and does not belong in this PR.

johnnyluo (approved):
LGTM

Issue #5548, opened 2026-08-09
[Fix] Keygen join completes behind the passcode lock and the new keyshare is discarded [Passcode / Keygen]


rkokhatskyi:
## Problem

A keygen or reshare ceremony can run to completion while the app is passcode-locked, and its freshly
generated keyshare is then refused at save and lost. The other devices commit normally, so the
committee considers the vault created (or resharded) while this device holds nothing and can never
sign for it.

`JoinKeygenViewModel` auto-advances with no user interaction: `waitForKeygenToStart` polls
`while (isActive)` (`JoinKeygenViewModel.kt:382`) and, once the initiator adds this device to the
committee, routes straight to `Route.Keygen.Generating` (`:397-399`). `KeygenViewModel.init` calls
`generateKey()` immediately on construction (`KeygenViewModel.kt:186-187`). None of the passcode
gate's protections apply, because all three of them gate *input*: `PasscodeGuard` swallows pointer
events, installs a `BackHandler` and applies `clearAndSetSemantics` — while `SetupNavGraph` is
composed unconditionally in `MainActivityContent` and both routing collectors run regardless of lock
state. Navigation and ViewModel construction are never consulted.

At the end, `saveVault()` (`KeygenViewModel.kt:349`) reaches `VaultRepositoryImpl.toVaultDb`, whose
`check(!(isLocked && dataKey == null))` (`VaultRepository.kt:291`) throws. `generateKey`'s
`catch (e: Exception)` (`:351`) swallows it into a generic error screen. The share existed only in
the ViewModel, nothing re-saves after unlock, and `tryAgain()` only navigates back — so it is gone.

**`AutoLockHold` was built for exactly this and is taken one screen too late.** Its KDoc says
*"locking before saveVault() leaves this device with no share for a vault the rest of the group has
already created"*. But the hold is taken inside `KeygenViewModel.generateKey` (`:325`), and the only
other reference in the app is `PasscodeAutoLock`. The join wait that precedes it holds nothing, and
`withHold` only postpones a *future* lock — it cannot recover a data key `lock()` has already zeroed.

The keysign twin is already protected: `JoinKeysignViewModel.kt:382` calls
`vaultRepository.awaitKeySharesReadable()`, with a comment noting that this screen composes behind
the passcode lock. That mitigation was never mirrored to the keygen path.

## Expected Behavior

A keygen or reshare join either does not start while the vault is unwritable, or waits for the
[...]

PR #5554, opened 2026-08-10
fix(passcode): draw the app lock in a window of its own

### LINKED ISSUE #5551: [Fix] Content drawn in its own window stays visible above the passcode lock screen [Passcode]
## Problem

The passcode gate is drawn inside the activity window (`MainActivityContent`'s `overlayContent`), so
anything that renders in a **separate** window sits above it. The current mitigation pops nav
`dialog<>` destinations when the gate closes, but it misses two shapes, and in both the user sees
wallet content over the lock screen and can still tap it.

**1. In-screen `ModalBottomSheet`s are never popped.** `onLocked` only drains entries whose
`navigatorName == "dialog"` (`MainActivityContent.kt:143-147`, `:317`). Material3
`ModalBottomSheet` is not a nav destination — in the pinned material3-android 1.4.0, `ModalBottomSheet`
calls `ModalBottomSheetDialog` unconditionally and `ModalBottomSheetDialogWrapper` extends
`ComponentDialog`, so it always gets its own window and no call site can opt out. Around 30 files
under `app/src/main/java/com/vultisig/wallet/ui/` use it, including `TransactionDetailBottomSheet`
and `AddressBookBottomSheet`. Lock with one open and it stays on screen, above the lock, live.

**2. `onLocked` is a one-shot edge.** `PasscodeGuard` fires it from
`LaunchedEffect(isLocked) { if (isLocked) onLocked() }` (`PasscodeGuard.kt:55`). The key is a
`Boolean`, so a `true → true` transition never re-runs the block: it drains only the back stack that
existed at the moment the gate closed. Anything routed *after* that is never popped, and both routing
collectors (`mainViewModel.destination` and `mainViewModel.route`) run unconditionally inside
`navContent`, so routing continues behind the closed gate.

Neither of the other two protections helps: `clearAndSetSemantics` is applied to the nav-content
`Box` (`MainActivityContent.kt:132`), which does not contain a separate window, and the guard's
`swallowPointerInput` only covers its own node in the activity window.

## Expected Behavior

Once the gate closes, nothing but the lock screen is visible or interactive, regardless of which
window it was drawn in or when it was created.

## Steps to Reproduce

**Bottom sheet**

1. Enable the passcode feature, set a passcode, set Auto-lock to **After 1 minute**.
2. Open a screen with a bottom sheet — e.g. tap a transaction to open `TransactionDetailBottomSheet`.
3. Background the app for over a minute, then return.
4. The lock screen is up, and the sheet is still drawn above it showing transaction details, and
   still responds to taps.

**Late-routed dialog**

1. Same setup, let the app lock.
2. Trigger anything that routes a `dialog<>` destination while locked (a deep link, or a flow that
   emits to `mainViewModel.destination`).
3. The dialog mounts above the lock screen and is never popped.

## Solution

Make the gate a property of the window stack rather than a one-time cleanup. Preferred: render the
lock screen itself in its own full-screen `Dialog` while the gate is closed, so it is above any
`ComponentDialog` the app opens, and dismissal is impossible while locked — that fixes both shapes
without enumerating sheet call sites. If the current in-window overlay is kept instead, then the pop
must be driven by the gate *state* rather than the closing edge, and in-screen sheets must also be
dismissed, which means routing sheet visibility through something the guard can reach.

### Must Do
- [ ] Ensure the lock screen renders above content drawn in its own window (`ModalBottomSheet`,
      `dialog<>` destinations)
- [ ] Make dismissal of that content driven by the gate being closed, not by the closing edge, so
      anything routed while already locked is covered too
- [ ] Verify on device with both repro paths above

### Must NOT Do
- Don't remove the existing `dialog<>` pop until the replacement covers it
- Don't change unrelated code

## Acceptance Criteria
- [ ] Locking with `TransactionDetailBottomSheet` open leaves only the lock screen visible, and taps
      do not reach the sheet
- [ ] A `dialog<>` destination routed while the app is already locked does not appear above the lock
      screen
- [ ] With no passcode configured, bottom sheets and dialog destinations behave exactly as before

rkokhatskyi:
## Problem

The lock was drawn inside the activity's window. A `ModalBottomSheet` and a `dialog<>` destination each get a window of their own, and one activity's windows stack in the order they are added — so nothing drawn in the activity's window can be above them. Lock the app with a transaction sheet open and the sheet stayed on top, readable and still taking taps. The `dialog<>` pop that mitigated this fired once, on the edge the gate closed, so a destination routed while already locked was never cleared.

## Fix

- `PasscodeGuard` hosts the lock in a full-screen `Dialog`, opened when the app locks, which puts it over anything already open. The in-window cover stays for the frame before that window is added and for the state that is still loading.
- The `dialog<>` pop follows the locked state instead of the edge, so a destination routed behind a closed gate is popped too.
- `VsSnackBar` is a `Dialog` as well and the global `SnackbarFlow` has async producers, so it is no longer composed while the gate is closed.
- `isGateClosed` is derived from a new `isLocked` rather than repeating the truth table, so the cover and the lock cannot disagree.

## Before / After

| Before | After |
|--------|-------|
| before | after |

Same emulator and mock state: the vault screen with a transaction sheet open, then locked. Before, the sheet draws over a dimmed lock screen and holds focus; after, the lock covers it and owns the keyboard.

## Testing

- `:app:assembleDebug`, `:app:lintDebug`, `:app:ktfmtCheck`, `:data:ktfmtCheck`
- `:app:testDebugUnitTest` — full suite, 1941 tests
- `:app:connectedDebugAndroidTest` — new `PopDialogDestinationsWhileLockedTest`, 4/4
- On device, both repro paths from the issue on API 35 and API 26: the lock covers an open sheet, the keypad opens and takes digits inside the dialog window, and back is swallowed.

Closes #5551

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app locking behavior by preventing interaction with underlying content while the lock screen is displayed.
  * Automatically closes open dialogs when the app becomes locked.
  * Ensures dialogs opened after unlocking remain available.
  * Delays snackbar messages until the app is ready for interaction.

* **Tests**
  * Added coverage for dialog handling across locked, unlocked, and transitioning app states.
  * Added validation for passcode lock and gate status behavior.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
@@ -93,6 +93,12 @@ internal fun MainActivityContent(
         snackbarFlow.collectMessage { (message, type) ->
             val resolved = message.asString(context)
             if (resolved.isBlank()) return@collectMessage
+            // Held rather than shown: the bar is a window of its own, so one raised while the gate
+            // is closed would be added after the lock's and drawn over it. Held rather than
+            // dropped, too — the bar's progress runs off its own scope whether or not anything is
+            // composed, so a message shown behind the lock would tick out unseen. The channel
+            // behind collectMessage buffers, so waiting here keeps it until there is a reader.
+            passcodeGuardModel.state.first { !it.isGateClosed }
             snackbarState.show(resolved, type)
         }
     }
@@ -142,14 +148,10 @@ internal fun MainActivityContent(
         overlayContent = {
             PasscodeGuard(model = passcodeGuardModel)
 
-            // The snackbar is a window of its own too, so one raised while the gate is up is added
-            // after the lock's and drawn over it.
-            if (!isGateClosed) {
-                VsSnackBar(
-                    modifier = Modifier.align(Alignment.BottomCenter),
-                    snackbarState = snackbarState,
-                )
-            }
+            VsSnackBar(
+                modifier = Modifier.align(Alignment.BottomCenter),
+                snackbarState = snackbarState,
+            )
         },
     )
 }
--- a/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
@@ -39,6 +39,7 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.navigation.NavBackStackEntry
+import androidx.navigation.NavController
 import androidx.navigation.NavDestination.Companion.hasRoute
 import androidx.navigation.NavHostController
 import com.vultisig.wallet.R
@@ -80,7 +81,11 @@ internal fun MainActivityContent(
     // tree at the same moment. Swallowing pointer input is only half a lock: everything behind the
     // gate stays composed, and TalkBack would happily walk it, read it out and activate it.
     val passcodeGuardModel: PasscodeGuardViewModel = hiltViewModel()
-    val isGateClosed = passcodeGuardModel.state.collectAsStateWithLifecycle().value.isGateClosed
+    val guardState by passcodeGuardModel.state.collectAsStateWithLifecycle()
+    val isGateClosed = guardState.isGateClosed
+    val isLocked = guardState.isLocked
+
+    PopDialogDestinationsWhileLocked(navController, isLocked)
 
     val context = LocalContext.current
     val snackbarState = mainViewModel.snackbarState
@@ -135,22 +140,16 @@ internal fun MainActivityContent(
             }
         },
         overlayContent = {
-            PasscodeGuard(
-                model = passcodeGuardModel,
-                // Dialog destinations render in their own window, above this overlay. Popping them
-                // as the gate closes is what stops a bottom sheet — a receive address, a QR — from
-                // sitting on top of the lock screen and still accepting taps.
-                onLocked = {
-                    while (navController.currentBackStackEntry.isDialog()) {
-                        if (!navController.popBackStack()) break
-                    }
-                },
-            )
+            PasscodeGuard(model = passcodeGuardModel)
 
-            VsSnackBar(
-                modifier = Modifier.align(Alignment.BottomCenter),
-                snackbarState = snackbarState,
-            )
+            // The snackbar is a window of its own too, so one raised while the gate is up is added
+            // after the lock's and drawn over it.
+            if (!isGateClosed) {
+                VsSnackBar(
+                    modifier = Modifier.align(Alignment.BottomCenter),
+                    snackbarState = snackbarState,
+                )
+            }
         },
     )
 }
@@ -306,6 +305,26 @@ private fun MainActivityContentPreview() {
             )
         },
     )
+}
+
+/**
+ * Keeps `dialog<>` destinations off the back stack for as long as [isLocked].
+ *
+ * They draw in a window of their own, and the passcode lock's window only covers the windows that
+ * were open before it, so one routed after the app locks opens on top of the lock — still visible,
+ * still taking input. Driven by the state rather than by the moment it changes, because routing
+ * carries on behind a closed gate.
+ */
+@Composable
+internal fun PopDialogDestinationsWhileLocked(navController: NavController, isLocked: Boolean) {
+    LaunchedEffect(navController, isLocked) {
+        if (!isLocked) return@LaunchedEffect
+        navController.currentBackStackEntryFlow.collect {
+            while (navController.currentBackStackEntry.isDialog()) {
+                if (!navController.popBackStack()) break
+            }
+        }
+    }
 }
 
 /**
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/passcode/PasscodeGuard.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/passcode/PasscodeGuard.kt
@@ -47,38 +47,50 @@ internal fun PasscodeGuard(model: PasscodeGuardViewModel = hiltViewModel()) {
     // which is what puts it first in line.
     BackHandler(enabled = state.isGateClosed) {}
 
-    // The whole gate while the persisted state is still loading, and the cover over the app for
-    // the frame between the state that raises the lock and the lock's own window being added.
     if (state.isGateClosed) {
+        // Covers the app for the frame between the state that closes the gate and the window below
+        // being added, which is the only moment the window cannot cover for itself.
         Box(
             Modifier.fillMaxSize()
                 .background(Theme.v2.colors.backgrounds.primary)
                 .swallowPointerInput()
         )
+
+        // One call site for the whole closed gate rather than one per state: a second would be a
+        // second composition slot, and moving between them would take the window down and put a new
+        // one up at exactly the moment the lock appears.
+        LockWindow { GateContent(state.passcodeState) }
     }
 
-    when (state.passcodeState) {
-        PasscodeState.Locked -> LockWindow { PasscodeLockScreen() }
+    // The passcode was just turned off, so the user has already identified themselves this session;
+    // sending them straight into a device-credential prompt would be a non sequitur.
+    if (state.passcodeState == PasscodeState.Disabled && !state.isIdentityProven) {
+        BiometryAuthScreen()
+    }
+}
+
+/** What the gate shows, once there is something to show. */
+@Composable
+private fun GateContent(passcodeState: PasscodeState) {
+    when (passcodeState) {
+        PasscodeState.Locked -> PasscodeLockScreen()
         PasscodeState.KeyUnavailable ->
-            LockWindow {
-                DeadEndScreen(
-                    title = stringResource(R.string.passcode_key_unavailable_title),
-                    message = stringResource(R.string.passcode_key_unavailable_message),
-                )
-            }
+            DeadEndScreen(
+                title = stringResource(R.string.passcode_key_unavailable_title),
+                message = stringResource(R.string.passcode_key_unavailable_message),
+            )
         PasscodeState.StoreUnavailable ->
-            LockWindow {
-                DeadEndScreen(
-                    title = stringResource(R.string.passcode_store_unavailable_title),
-                    message = stringResource(R.string.passcode_store_unavailable_message),
-                )
-            }
-        // Persisted state is still loading; the cover above is the gate until it resolves.
+            DeadEndScreen(
+                title = stringResource(R.string.passcode_store_unavailable_title),
+                message = stringResource(R.string.passcode_store_unavailable_message),
+            )
+        // Persisted state is still loading. The window is up regardless, because a dialog
+        // destination restored with the back stack opens its own window while this read is still
+        // running, and only another window can cover that one.
         PasscodeState.Unknown -> Unit
-        PasscodeState.Unlocked -> Unit
-        // The passcode was just turned off, so the user has already identified themselves this
-        // session; sending them straight into a device-credential prompt would be a non sequitur.
-        PasscodeState.Disabled -> if (!state.isIdentityProven) BiometryAuthScreen()
+        // The gate is open in both, so this is not composed.
+        PasscodeState.Unlocked,
+        PasscodeState.Disabled -> Unit
     }
 }
 
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/passcode/PasscodeGuard.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/passcode/PasscodeGuard.kt
@@ -5,13 +5,12 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.safeDrawingPadding
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
```diff
@@ -80,7 +81,11 @@ internal fun MainActivityContent(
     // tree at the same moment. Swallowing pointer input is only half a lock: everything behind the
     // gate stays composed, and TalkBack would happily walk it, read it out and activate it.
     val passcodeGuardModel: PasscodeGuardViewModel = hiltViewModel()
-    val isGateClosed = passcodeGuardModel.state.collectAsStateWithLifecycle().value.isGateClosed
+    val guardState by passcodeGuardModel.state.collectAsStateWithLifecycle()
+    val isGateClosed = guardState.isGateClosed
+    val isLocked = guardState.isLocked
+
+    PopDialogDestinationsWhileLocked(navController, isLocked)
```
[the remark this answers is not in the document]
rkokhatskyi: Deliberate. `isGateClosed` also covers `PasscodeState.Unknown`, which every launch passes through whether or not a passcode is configured, so popping there would delete a `dialog<>` destination restored after process death for users who have no passcode — against the ticket's "with no passcode configured, bottom sheets and dialog destinations behave exactly as before". Nothing new can be routed during `Unknown` either: both routing collectors only start after `currentBackStackEntryFlow.first()` and `CheckDeeplink` is gated on `isNavigationReady`, so the only thing that window leaves uncovered is a restored destination, and it behaves exactly as it did before this PR. Closing it without touching the back stack means hosting the gate in its own window for the whole closed gate rather than only while locked, which adds and removes a window on every cold start — happy to do that instead if you would rather have it.

rkokhatskyi:
This class of bug is not Android-only, so the ports are filed: iOS is vultisig/vultisig-ios#5082 and Windows/Extension is vultisig/vultisig-windows#4596. iOS already documents the gap in a comment on `passcodeGate` and names the same fix ("the lock hosted in a window above the presentation layer"); Windows has it undocumented, where every modal portals to `document.body` at z-index 1-1100 and escapes `#root`'s `isolation: isolate`.

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
```diff
@@ -308,6 +307,26 @@ private fun MainActivityContentPreview() {
...
+ *
+ * They draw in a window of their own, and the passcode lock's window only covers the windows that
+ * were open before it, so one routed after the app locks opens on top of the lock — still visible,
+ * still taking input. Driven by the state rather than by the moment it changes, because routing
+ * carries on behind a closed gate.
+ */
+@Composable
+internal fun PopDialogDestinationsWhileLocked(navController: NavController, isLocked: Boolean) {
+    LaunchedEffect(navController, isLocked) {
+        if (!isLocked) return@LaunchedEffect
```
aminsato, line 323: This effect is armed by `isLocked`, but `isLocked` is false for `PasscodeState.Unknown` — and `Unknown` is exactly the window where the only thing covering the app is the in-activity-window `Box` in `PasscodeGuard`, which by this PR's own reasoning cannot cover another window.

After process death with `dialog<Route.AddressQr>` / `dialog<Route.Receive>` on the back stack, the NavHost restores that dialog destination on relaunch and its window opens immediately while `passcodeRepository.initialize()` is still decoding keystore material. `if (!isLocked) return@LaunchedEffect` skips the pop, so the receive-address / QR sheet is fully visible and tappable over the launch cover until the state resolves to `Locked`.

`isGateClosed` is the signal that matches what the cover actually protects — keying the effect (and the parameter) on it closes the pre-read frame.

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
```diff
@@ -308,6 +307,26 @@ private fun MainActivityContentPreview() {
...
+/**
+ * Keeps `dialog<>` destinations off the back stack for as long as [isLocked].
+ *
+ * They draw in a window of their own, and the passcode lock's window only covers the windows that
+ * were open before it, so one routed after the app locks opens on top of the lock — still visible,
+ * still taking input. Driven by the state rather than by the moment it changes, because routing
+ * carries on behind a closed gate.
+ */
+@Composable
+internal fun PopDialogDestinationsWhileLocked(navController: NavController, isLocked: Boolean) {
```
aminsato, line 321: `LockWindow`'s KDoc names two window-owning things — "A `ModalBottomSheet` and a `dialog<>` destination each get a window too" — but this only pops nav dialog entries. A `ModalBottomSheet` driven from screen state has no equivalent defence.

`VaultAccountsScreen` shows `MonthlyBackupReminder`, `NotificationsIntroBottomSheet` and `VaultNotificationOptInBottomSheet` from async ViewModel state (`showGlobalBackupReminder()` / `checkNotificationPrompt()` in `VaultAccountsViewModel.loadVault`, both `safeLaunch`ed). On a cold start with a passcode set, or on resume after auto-lock, the passcode read can finish first and put the lock window up; `showMonthlyBackupReminder` then flips a moment later and the sheet's window is added *after* the lock's. The backup reminder renders on top of the passcode lock and still takes taps — "Back up now" (`viewModel::backupVault`) walks the user into the vault backup flow from a locked app.

Three per-window-type patches (this pop, the snackbar gate, the in-window cover) each defend one window kind; a single mechanism that suppresses window-owning content while the gate is closed would cover the sheets too.

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
```diff
@@ -135,22 +140,16 @@ internal fun MainActivityContent(
...
-            )
+            PasscodeGuard(model = passcodeGuardModel)
 
-            VsSnackBar(
-                modifier = Modifier.align(Alignment.BottomCenter),
-                snackbarState = snackbarState,
-            )
+            // The snackbar is a window of its own too, so one raised while the gate is up is added
+            // after the lock's and drawn over it.
+            if (!isGateClosed) {
```
aminsato: This drops snackbar messages rather than deferring them. `VSSnackbarState` launches its progress pipeline in `init` with `.launchIn(coroutineScope)` — 60 steps over `duration` (1s default), then `emit(ProgressState(isVisible = false, ...))` — entirely independent of whether `VsSnackBar` is composed, and `MainViewModel` holds it for the process lifetime.

So a message raised while the gate is closed (a broadcast or keysign failure arriving through `snackbarFlow` during auto-lock, or anything raised in the `Unknown` frame at launch) ticks to `isVisible = false` behind the lock. The user unlocks a few seconds later and never sees it — and for those failures the snackbar is the only surface.

Holding the message and showing it once the gate opens, rather than not composing the bar, keeps the notification.

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
```diff
@@ -308,6 +307,26 @@ private fun MainActivityContentPreview() {
...
+ * were open before it, so one routed after the app locks opens on top of the lock — still visible,
+ * still taking input. Driven by the state rather than by the moment it changes, because routing
+ * carries on behind a closed gate.
+ */
+@Composable
+internal fun PopDialogDestinationsWhileLocked(navController: NavController, isLocked: Boolean) {
+    LaunchedEffect(navController, isLocked) {
+        if (!isLocked) return@LaunchedEffect
+        navController.currentBackStackEntryFlow.collect {
+            while (navController.currentBackStackEntry.isDialog()) {
```
aminsato, line 325: The pop keeps no record of what it dropped, and nothing replays it on unlock — so a dialog destination routed while locked is destroyed for the session rather than deferred.

`VaultAccountsViewModel.showVerifyFastVaultPasswordReminderIfRequired` calls `navigator.route(Route.FastVaultPasswordReminder(vaultId))` from a `safeLaunch` as soon as the vault loads, which on a locked cold start lands while `isLocked` is true. It gets popped, and the routing collectors are not replayed on unlock (`aDialogDestinationRoutedAfterUnlockingSurvives` only covers routes issued *after* unlocking), so the reminder never appears that session.

The same holds for any queued flow step landing on a dialog destination (`dialog<Keysign.Password>`, `dialog<FastVaultVerification>`). `AutoLockHold` is used only by `KeygenViewModel`, so a signing flow that routes to its password dialog while locked loses that step and stalls after unlock.

Recording the popped entry and re-routing it when the gate opens turns this from a drop into a deferral.

app/src/main/java/com/vultisig/wallet/ui/screens/passcode/PasscodeGuard.kt
```diff
@@ -35,65 +36,86 @@ import com.vultisig.wallet.ui.theme.Theme
...
+ * Hosts [content] in a window of its own.
+ *
+ * A `ModalBottomSheet` and a `dialog<>` destination each get a window too, and one activity's
+ * windows stack in the order they were added — so nothing drawn in the activity's own window can be
+ * above them, however opaque. Adding the lock's window when the app locks is what puts it over a
+ * sheet that was already open.
+ */
+@Composable
+private fun LockWindow(content: @Composable () -> Unit) {
+    Dialog(
```
aminsato, line 107: Moving the lock out of the activity's window looks like it drops the `FLAG_SECURE` coverage the passcode screen used to inherit for free.

`DialogProperties.securePolicy` defaults to `SecureFlagPolicy.Inherit`, which resolves against the parent window only in `DialogWrapper.init` / `updateParameters` (a `SideEffect`). `MainActivity.observePreventScreenshots()` collects `preventScreenshotsRepository.isEnabled` from DataStore asynchronously, and `PasscodeGuardViewModel` resolves `Locked` from its own async read. If the passcode read wins, this `Dialog` is created while the activity window still has no `FLAG_SECURE`, so the lock window is created non-secure. `PasscodeGuardUiModel` then stays constant while the user sits on the lock screen, so `PasscodeGuard` never recomposes and the `SideEffect` that would re-apply the policy never runs.

A screenshot, screen recording, or screen-share at that point captures the whole passcode entry screen while the activity window behind it is correctly blacked out. Passing the prevent-screenshots state into `securePolicy` explicitly, instead of relying on inheritance, removes the race.

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
rkokhatskyi: **Fixed — but not by rearming on `isGateClosed`.**

- Real hole, agreed: a restored `dialog<Receive>` / `dialog<AddressQr>` opens its window while `initialize()` is still reading, and the in-window cover can't reach it.
- `isGateClosed` also covers `Unknown`, which **every** launch passes through — popping there destroys a restored destination for users with no passcode, against the ticket's "behave exactly as before".
- Instead the gate now hosts its window for the **whole** closed gate. The restored dialog is covered, not destroyed, and the back stack is untouched either way.
- One call site, not one per state — a second slot would take the window down and put a new one up at the moment the lock appears.

Verified on device: the `Unknown` frame is an opaque window with nothing in it, and a no-passcode cold start still reaches the start screen with no flash (the theme gives dialog windows no enter/exit animation).

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
rkokhatskyi: **Fixed — good catch.**

- Correct: `VSSnackbarState` ticks its progress off `MainViewModel`'s scope regardless of composition, so gating composition turned a leak into a silent drop.
- The collector now waits for the gate to open before calling `show()`. `SnackbarFlow` is a `Channel(BUFFERED)`, so the message simply keeps until there's a reader.
- That lets the composition gate come out entirely — the bar is composed unconditionally again, and a message raised while locked is shown on unlock instead of lost.

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
rkokhatskyi: **Agreed — this is the residual named in the PR description, tracked as #5560** (rescoped to exactly this after review).

- Your two extras are new to that ticket and I'll add them: `NotificationsIntroBottomSheet` and `VaultNotificationOptInBottomSheet` via `checkNotificationPrompt()`'s `safeLaunch` — same async shape as the backup reminder.
- The "Back up now" point is sharper than what's written there: it's an **action** reachable from a locked app, not just something visible.
- It's two mechanisms now, not three — the snackbar moved to the producer side in this push.

On the single mechanism: Compose has no hook to suppress a `ModalBottomSheet` from outside it, so it means every sheet call site reading a gate-aware wrapper. The only alternative — not composing `navContent` while locked — disposes the `NavHost` and kills in-flight work, which is what #5548 / #5553 exist to protect. That belongs in #5560.

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
rkokhatskyi: **Not taking this one — the mechanism is right, the trade doesn't pay.**

- Popping predates this PR: `onLocked` already drained dialog entries on the closing edge. The only **newly** lost case is a destination routed while already locked.
- Confirmed instance is `FastVaultPasswordReminder` — a reminder that returns on the next launch.
- `Keysign.Password` goes through `LaunchKeysignUseCase` off the user's sign tap, so the app would have to lock in the gap between tap and route.
- The one I'd actually watch is `FastVaultVerification` (`KeygenViewModel.kt:1009`) — keygen holds auto-lock and the lock can fire as the hold releases. That's #5548 / #5553 territory, not this PR's.

Replay means re-issuing a route with its args and options against a back stack and vault that may have moved on — a lot of moving parts for a reminder. The principled version is not routing at all while the gate is closed, which is what iOS does with deeplinks and what #5559 asks for. Better done there, across all routing rather than the dialog subset.

app/src/main/java/com/vultisig/wallet/ui/screens/passcode/PasscodeGuard.kt
rkokhatskyi: **Mechanism is right, but the payload is empty — declining.**

- Agreed on the race: `Inherit` is snapshotted in `DialogWrapper.init` and re-applied only from the `SideEffect`, and `PasscodeGuardUiModel` is constant while `Locked`, so the window can stay non-secure.
- But the lock window holds a title, a subtitle and six masked cells. The activity window keeps its **own** `FLAG_SECURE` and is redacted independently, so a capture in that state yields the lock screen over a redacted app — no wallet content.
- The race is also narrower than it looks: `observePreventScreenshots` starts at `onStart`, while the guard's view model is only constructed once `MainActivityContent` composes after the splash.

Wiring `PreventScreenshotsRepository` into the guard to protect a screen with nothing on it is coupling without a payload. Worth revisiting the moment anything sensitive lands on the gate.

johnnyluo (approved):
LGTM

PR #5567, opened 2026-08-10
fix(ripple): reject and flag XRPL partial payments that hide their delivery

### LINKED ISSUE #5555: XRPL dApp signing: hidden Flags/Paths let a tx deliver less than the confirmation shows
## Summary
For dApp-supplied XRPL transactions (co-signer path), `Flags` and `Paths` are signed verbatim and omitted from the **decoded** confirmation. A `Payment` with `tfPartialPayment` and no `DeliverMin` is approved as an exact payment but can settle for far less than the displayed `Amount`. Severity: **Medium**. (Android is partially mitigated — see below.)

## Background (XRPL semantics)
`tfPartialPayment` (`0x00020000`) turns `Amount` from a guaranteed delivery into a **maximum**; the delivered value is decided at execution. Without `DeliverMin` there is no floor. `Paths` let the initiator steer routing.

## What's wrong
- `data/.../chains/helpers/RippleHelpter.kt` → `buildDappRawJsonInputData` calls `setRawJson(rawJson)` — the whole inbound JSON is signed. Its gate `verifyDappTransaction` only checks the type allowlist, forbids `TxnSignature`/`Signers`/`SigningPubKey`, and requires `Account == expectedAccount`. **`Flags` and `Paths` are never read.**
- `data/.../chains/helpers/RippleDappTransaction.kt` decodes only type, `Account`, `Destination`, `DestinationTag`, `Amount`, `SendMax`, `DeliverMin`, `TakerGets`, `TakerPays`, `LimitAmount`, `Fee` — no `Flags`/`Paths`.
- The keysign notification/summary (`GetKeysignTransactionSummaryUseCase` → `RippleDappTransactionDecoder.summarize()`) shows `Amount` alone — the number tfPartialPayment makes misleading.

**Partial mitigation:** `ui/components/SignRippleDisplayView.kt` always appends the full raw JSON (expanded by default) via `VerifyCardJsonDetails`. The signed bytes are technically on screen — but unlabeled, below every decoded row, inside a scrollable box. That downgrades "hidden" to "unlabeled"; it is not a fix, since it relies on a co-signer scrolling to a blob and knowing `"Flags":131072` means partial payment.

## Impact
Same self-swap scenario (Destination = vault, inflated `Amount`, large `SendMax`, `Flags: 131072`, no `DeliverMin`): passes every check, the decoded card + notification show the attractive `Amount`, and it can deliver dust while spending up to `SendMax`. No Blockaid backstop — XRPL is not scanned.

## Fix
1. **Display (show the correct details):** add `Flags` and `Paths` to `RippleDappTxFieldKey` and decode them in `RippleDappTransaction.kt`; render an explicit warning row when `tfPartialPayment` is set and indicate `Paths` presence. Fix the notification summary so it doesn't present `Amount` as guaranteed when partial-payment is set.
2. **Fail closed:** in `verifyDappTransaction`, reject a `Payment` with `tfPartialPayment` set when `DeliverMin` is absent (same fail-closed pattern already used for forbidden fields/types).

## Acceptance
- A `Payment` with `Flags: 131072` and no `DeliverMin` is rejected in `verifyDappTransaction`, or clearly flagged as partial-payment in the decoded confirmation and notification.

aminsato:
Closes #5555

## The problem

A dApp-supplied XRPL transaction is signed verbatim (`setRawJson`), so `verifyDappTransaction` is the only gate between the wire and the signer — and it never read `Flags` or `Paths`.

`tfPartialPayment` (`0x00020000`) turns a `Payment`'s `Amount` from a guaranteed delivery into a **maximum**; the delivered value is decided at execution and, without `DeliverMin`, has no floor. A `Payment` with that flag set therefore passed the type allowlist, the `Account` check and the forbidden-field check, while the decoded confirmation card and the keysign notification both presented `Amount` as an exact payment. It can deliver dust and still spend up to `SendMax`. XRPL is not scanned by Blockaid, so there is no second line of defence.

The raw JSON was technically on screen — unlabeled, below every decoded row, inside a scroll box. That made the flag *unlabeled* rather than *hidden*; it did not make it visible.

## The fix

**Fail closed** — `verifyDappTransaction` now refuses a `Payment` that sets `tfPartialPayment` without a positive `DeliverMin` floor. Three details:

- **Scoped to `Payment`.** `OfferCreate` reuses `0x00020000` as `tfImmediateOrCancel`; reading the bit generically would have broken every XRPL DEX swap. Locked in by a test.
- **An unreadable `Flags` is refused** (non-integer, or explicit null) — the bit cannot be ruled out, so it fails closed rather than assuming unset.
- **`DeliverMin` must be a positive amount**, not merely present. A native floor is an integer count of drops, an issued floor is `currency` + `issuer` + a decimal `value`, and all of them are string-encoded — so `""`, `"0"`, `"-1"`, `"abc"`, a bare number and `{"currency":"USD"}` naming no issuer are each treated as no floor. This is field-for-field what `isPositiveRippleAmount` does in the extension's `sanitizeRippleDappTx`, the initiator-side half of the same defense.

**Show the real terms** — `Flags` and `Paths` are decoded into their own rows (`Flags` as its raw bitfield so it matches the JSON below it; `Paths` as a route count; both omitted when zero/absent). The partial-payment warning is rendered **outside** the collapsible summary — a warning behind a chevron would repeat the bug being fixed. Styling matches the existing unlimited-approval warning.

**Stop the banner implying a guarantee** — `summarize()` renders a partial-payment amount as a ceiling: `Payment: ≤ 1 XRP`, or `Payment: ≥ 0.5 XRP, ≤ 1 XRP` when a floor is set. Comparison symbols rather than wording like "up to", because this string is built in the data layer and would otherwise be untranslatable English.

Strings added to all 10 locales. `Flags` is intentionally left untranslated everywhere — it is the literal XRPL field name and its value is a raw number the co-signer matches against the JSON directly below.

## Known limitation

Rejection keys on a **positive** `DeliverMin`, so `"0"`, `"-1"`, `"abc"` and an issued amount naming no issuer are all refused as the non-floors they are. What remains open is magnitude: a dApp can still satisfy the gate with `DeliverMin: "1"` — one drop, functionally the same attack — though it now surfaces as a visible `Deliver min: 0.000001 XRP` row beside the warning, which is a judgment a co-signer can actually make. Enforcing a ratio (`DeliverMin` ≥ some fraction of `Amount`) would close it, but that is a policy choice beyond the issue and should match whatever Windows and iOS decide.

## Test plan

Automated (green):

- `./gradlew :data:testDebugUnitTest --tests '*Ripple*'` — 36 tests, 0 failures: 20 in `RippleHelperTest`, 16 in `RippleDappTransactionDecoderTest`
  - **gate:** rejects a partial `Payment` with no floor; rejects a `DeliverMin` that is present but is not a well-formed, strictly positive XRPL amount — `"   "`, `null`, `"0"`, `"-1"`, `"abc"`, fractional drops `"1.5"`, bare-number drops, `{"value":"25"}` and `{"currency":"USD","value":"25"}` naming no issuer, and the zero/negative/non-numeric issued forms; accepts a one-drop floor and a sub-unit issued floor; passes unrelated flags (`tfNoRippleDirect`); passes an `OfferCreate` reusing the same bit; rejects an unreadable `Flags`
  - **decode:** `Flags`/`Paths` rows; no row for a zero/absent `Flags`; the partial-payment bit is only read for a `Payment`; `summarize()` ceiling and floor+ceiling forms
  - the positivity test was checked against a revert of `hasPositiveAmount` — it fails there (`20 tests completed, 1 failed`), so it pins the behaviour rather than passing vacuously
- `./gradlew :app:mergeDebugResources` — passes; compiles all 10 locale files, which is what catches an escaping error in the added strings
- `./gradlew :app:lintDebug` — passes
- `ktfmt` applied

Manual — **the playground does not reproduce this, and cannot.**

The earlier plan here assumed vultisig/community-tools#38 would make this a one-click repro. #38 has merged, and it does not: the initiator refuses these transactions itself, before a keysign exists.

1. `sanitizeRippleDappTx.ts` rejects a partial `Payment` unless `isPositiveRippleAmount(DeliverMin)` — the same rule this PR adds to Android, and the reason the fixture returns *"Ripple partial payment must specify a positive DeliverMin"* in the browser
2. `@vultisig/sdk` then re-checks `Account` / `Destination` / `Amount` against the reviewed payload at keysign time
3. only after both does a keysign QR exist for a phone to scan

So nothing this guard would reject can reach an Android co-signer through the official initiator. That is precisely what the guard is for — it is the last line against a **tampering relay**, not against our own extension — but it does mean there is no honest end-to-end path that exercises it. Reaching it requires deliberately stubbing the check in `sanitizeRippleDappTx.ts` and rebuilding the extension, which stages the failure rather than observing it. The unit tests above are the real verification, and the implementation is kept field-for-field in step with `isPositiveRippleAmount` so the two halves of the defense agree.

One unrelated blocker found while attempting the run, worth its own issue: Android cannot sign **any** XRPL transaction carrying `Paths` — wallet-core 4.7.0 fails with `Unknown/unsupported XRPL type 'PathSet'`, downstream of this guard having already passed. Every cross-currency fixture in the playground carries `Paths`. Pre-existing and untouched by this PR, which nonetheless decodes a `Paths` row.

## Summary by CodeRabbit

* **New Features**
  * Added clearer XRP transaction details, including flags and routing paths.
  * Partial-payment transactions now display a prominent warning that the shown amount may be a maximum.
  * Partial-payment summaries include minimum-delivery information when available.
  * Added localized messaging across supported languages.

* **Bug Fixes**
  * Improved validation for partial payments, including minimum-delivery amounts and transaction flags.
  * Rejected invalid, missing, unreadable, blank, or non-positive payment amounts.


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleDappTransaction.kt
```diff
@@ -3,12 +3,72 @@ package com.vultisig.wallet.data.chains.helpers
...
+ */
+internal fun JsonObject.flagsOrNull(): Long? =
+    (this["Flags"] as? JsonPrimitive)?.contentOrNull?.toLongOrNull()
+
+/**
+ * The value of [key] when it is a non-blank JSON *string*, else null. XRPL encodes every amount as
+ * a string, and a blank one names nothing: it must read the same as an absent field so a floor the
+ * verify screen would not render cannot pass validation.
+ */
+private fun JsonObject.stringOrNull(key: String): String? =
```
rkokhatskyi: `stringOrNull` at `RippleDappTransaction.kt:36` is shadowed inside the decoder by the member at `:309`, so the identical expression `stringOrNull("currency")` runs a strict validator in `hasPositiveAmount` and a lenient one in `addAmount`, which a rename would make visible

rkokhatskyi (approved):

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/RippleDappTransaction.kt
johnnyluo: Already fixed by 2418ce180 ("name the two XRPL string readers apart") — the two helpers are now `nonBlankStringOrNull` and `displayStringOrNull`, no more shadowing. Resolving.

PR #5568, opened 2026-08-10
refactor(design-system): put the remaining corner radii on the scale (phase 3)

aminsato:
Phase 3 of 4 for #5503. Phase 1 (#5544) added the token scale, phase 2 (#5547) moved the container level onto 24. This is the rest of the sweep: every corner literal that was already a step on the scale now reads the token, so the scale is the only place a radius is written down. Phase 4 is the lint rule that keeps it that way.

**383 raw corner literals become 27**, and all 27 are named below.

## What changed

**Plain renames — 4, 8, 12 → `xs`, `sm`, `md`.** No value moves. The 12s are the largest population in the app and most are rows and fields rather than containers, which is why phase 2 left them: blanket-converting that group would have been a restyle, and reading them as `md` is not.

**Fully-round literals → `pill`.** The app had ten spellings of "round": 50, 60, 70, 77, 88, 99, 100 and 999 dp, `percent = 50` and `percent = 100`. They all draw the same capsule — but only because Compose clamps a corner to half the surface's smaller side, so the equivalence is a property of the *surface*, not of the number. Every site was checked against its own surface first. The bound is real: on a 320×240 surface `RoundedCornerShape(99.dp)` draws 99 and is not a capsule.

That check found the spelling a census misses — a number that doesn't read as fully-round but clamps anyway: **30 on a 50dp carousel item, 18 on a 32dp toggle, 3 on a 6dp dot**. Those were already drawing capsules; they now say so.

`RoundedCornerShape(percent = 100)` is the one collapse that isn't obvious. Compose clamps each corner separately rather than scaling all four, so asking for 100% takes the whole smaller dimension and leaves the opposite corner nothing — a very different shape from asking for half of it twice. It survives only because the leftover rule then splits the short edge evenly. That spelling carries `VsButton`, the app's primary button, so it is **measured in a test** rather than argued.

**`V2SheetShape` hoisted.** The V2 sheet frame was spelled out three times in one package. That's the drift the scale exists to stop, token or no token.

## What stays off the scale, and why

| Site | Value | Why |
|---|---|---|
| Bottom sheets | 32 / 34 / 38 / 40 | The scale stops at 24 — the design file's sheet frames are stock design-kit components (Apple's 38, Material 3's 28), so there is no step to reference. Phase-1 ruling. |
| QR frame (peer discovery) | 24.75 / 18.56 | Concentric by construction; the inner corner sits a fixed inset inside the outer one and rounding either breaks it. |
| Key-import chain row | 20 | Figma value sits equidistant from 16 and 24. iOS left the same row. |
| Search field | 10 | See below. |
| Checkbox, skeleton bars | 2 | Below `xs`; snapping to 4 is visible at that size. |
| Governance tally bar | 5 / 3 | See below. |
| ChooseVault container, limit-swap toggle | 15, 20 | No design source at all. Left bare on purpose so phase 4's rule surfaces them. |
| `Shapes.large`, square corners | 0 | A zero corner is the absence of a radius, not a step. |

## Two things that need a design call — not fixed here

**1. The input radius.** There is no single answer in the app: the search field is **10**, `VsTextInputField` and the passcode cells **12**, the code-entry boxes **24**, every v2 search surface **fully round** — and the design file's input measures **16**. Snapping the one off-scale input to a token would pick a winner for the whole family by accident, so it's left at 10 with the inventory recorded at the site. Worth settling before phase 4.

**2. A real drift bug in the governance tally bar.** It rounds at 5 when empty and 3 once voted, so the corner visibly changes the moment the first vote lands. Recorded in code rather than fixed, because which one is right is a design question.

## Test plan

- `:app:compileDebugKotlin`, `:app:testDebugUnitTest`, `:app:lintDebug` — all green, 0 lint errors.
- `RadiusTokenTest` grew from 9 to 11 cases: one pins `percent = 100` against `pill` on a button-sized surface, one pins the four clamped call sites (carousel item, limit-swap toggle, referral dot, sheet grabber).
- **Pixel diff, 128 PreviewActivity screens, main vs branch on one emulator.** A rename sweep should move nothing, so the whole point is the noise floor: the branch was captured **twice** to find out what differs between two runs of identical code. 65 of 128 screens are unstable — status-bar icons, animations, a system clipboard toast, an IME tooltip. Exactly one screen (`swap_toolbar`) differed against main but not against itself; six fresh captures of it from the same APK differ from each other in the *same three pixels*, because it draws a live `CircularProgressIndicator` off `Clock.System.now()`.

  **Result: 0 screens differ beyond the measured noise.**

No before/after screenshots because there is nothing to show — that is the claim being made, and the diff above is the evidence for it.

## Summary by CodeRabbit

* **Style**
  * Standardized corner radii across wallet screens, forms, cards, buttons, charts, banners, and bottom sheets.
  * Improved visual consistency by applying shared theme shapes for medium, small, and pill-style elements.
  * Preserved intentional exceptions for specialized controls such as search fields and QR frames.

* **Tests**
  * Expanded coverage for recognizing equivalent fully rounded and capsule-shaped elements.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -1271,7 +1271,7 @@ private fun SolanaDisplayPreview() {
                 .fillMaxSize()
                 .background(
                     color = Theme.v2.colors.variables.bordersLight,
-                    shape = RoundedCornerShape(12.dp),
+                    shape = Theme.v2.radius.md,
                 )
                 .padding(8.dp),
         verticalArrangement = Arrangement.spacedBy(12.dp),
@@ -1321,10 +1321,7 @@ private fun SolanaInstructionMock(
     Column(
         modifier =
             Modifier.fillMaxWidth()
-                .background(
-                    shape = RoundedCornerShape(8.dp),
-                    color = Theme.v2.colors.backgrounds.dark,
-                )
+                .background(shape = Theme.v2.radius.sm, color = Theme.v2.colors.backgrounds.dark)
                 .padding(6.dp),
         verticalArrangement = Arrangement.spacedBy(6.dp),
     ) {
@@ -1674,7 +1671,7 @@ private fun DeFiAccountListPreview() {
             modifier =
                 Modifier.background(
                     color = Theme.v2.colors.backgrounds.secondary,
-                    shape = RoundedCornerShape(12.dp),
+                    shape = Theme.v2.radius.md,
                 )
         ) {
             AccountList(
--- a/app/src/main/java/com/vultisig/wallet/ui/components/Modifiers.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/Modifiers.kt
@@ -3,7 +3,6 @@ package com.vultisig.wallet.ui.components
 import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.background
 import androidx.compose.foundation.border
-import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.composed
 import androidx.compose.ui.unit.dp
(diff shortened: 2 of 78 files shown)


johnnyluo (approved):
LGTM

PR #5569, opened 2026-08-10
fix(backup): refuse a vault whose keyshares only partly decoded

### LINKED ISSUE #5565: Backup export accepts a partial keyshare set — produces a .vult that restores cleanly and cannot sign
Required by vultisig/vultisig-sdk#1838: a `.vult` must always be complete and importable anywhere.

## What's wrong

Export only refuses when **all** keyshares are missing. It should refuse when **any** are missing.

`MapVaultToProto.kt:47`:
```kotlin
if (from.keyshares.isEmpty()) throw VaultKeysharesUnavailableException()
```

The comment above that line already describes the right behaviour. The check doesn't do it.

## Why it matters

When the passcode key isn't available, `toKeyShares` (`VaultRepository.kt:183-197`) drops the encrypted rows and returns the plaintext ones. It warns, it doesn't throw. A half-full list isn't empty, so the check passes.

You get a `.vult` that saves fine, imports fine, and can't sign. You find out when you need it.

## When it happens

Any time plaintext and `vlpc1:` rows sit in the table together — a `protectAll` that died halfway, or a share written while the key was unavailable.

## Fix

Refuse the export unless every stored row decoded. `toKeyShares` has to report the drop instead of swallowing it — compare rows read against shares returned.

Also check `isBackedUp` is only set after a successful export, so a refused one doesn't mark the vault as backed up.

## Done when

- [ ] Export throws if any row failed to decode, not only if all did
- [ ] `toKeyShares` reports a partial read instead of warning and continuing
- [ ] `isBackedUp` set only after a complete export
- [ ] Test: one plaintext row, one encrypted row, no key — export refused

aminsato:
Closes #5565

## Problem

`MapVaultToProto` refuses an export when the keyshare list is empty. A half-full list is not empty, so it passes.

The list can be half-full because a keyshare table holding both plaintext and `vlpc1:` rows is a state the app supports — `VaultKeyShareProtection`'s own contract says an interrupted `protectAll` "leaves a readable mix rather than a corrupt table". Reading that mix without the data key dropped the encrypted rows and returned the plaintext ones (`VaultRepository.kt:183-197`), warning rather than reporting.

Nothing downstream can tell a partial vault from a complete one:

- **Export** writes a `.vult` that saves, restores cleanly, and can never sign — the worst shape a backup failure takes, because it is only discovered when the backup is needed.
- **Reshare** merges the survivors at `KeygenViewModel.kt:502` and writes the truncated set back over the real rows. That one is permanent.

## Fix

Make the locked read all-or-nothing: if any row fails to decode while the data key is unavailable, return none.

Empty is the shape callers already check for, so the existing export guard starts working with no new exception type or plumbing, and the reshare merge sees an empty set — which an empty upsert leaves alone rather than overwriting. Background sync, the reason a locked read returns a vault at all, only ever needed public keys and addresses.

A failure *with the key in hand* still throws. That is a damaged row or one written for another vault — a real defect, not an expected read — and hiding it behind the same silence would be wrong.

`MapVaultToProto` gains a comment and no logic: the count of shares a vault *should* have exists only in the repository, so completeness cannot be checked at the mapper.

## Not currently reachable through the UI

Worth stating plainly, since the issue reads as though it is live. Every route from a partial read to an export is already closed:

- `Locked` — the lock screen covers the app.
- `KeyUnavailable` / `StoreUnavailable` — `PasscodeGuard.kt:80` shows a full-screen dead end.
- `Disabled` — `resolveInitialState` only returns it when no encrypted rows exist, so nothing can fail to decode.
- Stale vault read behind the lock screen — all three backup ViewModels call `awaitKeySharesReadable()` before reading (`BackupPasswordViewModel.kt:127`, `BackupVaultViewModel.kt:115`, `BackupPasswordRequestViewModel.kt:102`).

This is the last line of defence being made correct, not a live user-facing bug. It matters because it is the layer that is supposed to hold when one of those upstream guards is changed or missed.

## On the issue's checklist

Three of four items are addressed here. The fourth — `isBackedUp` set only after a complete export — needed no change: all three backup paths already gate `setBackupStatus(id, true)` behind `backupSuccess` (`BackupPasswordViewModel.kt:319`, `BackupVaultViewModel.kt:201`, `BackupPasswordRequestViewModel.kt:196`), so a refused export could never mark a vault backed up.

## Test plan

Four new tests. In `VaultRepositoryImplTest`:

- mixed table, no key → **no** shares (the bug)
- mixed table, key available → both shares load, so the rule costs nothing in the normal case
- mixed table, wrong key → still throws

In `MapVaultToProtoImplTest`: pins the mapper's half of the contract, with a comment pointing at the repository test that owns the other half — `VaultRepositoryImpl` is `internal` to `:data`, so the two halves cannot meet in one test.

Falsification check: removing `return emptyList()` fails `get drops every keyshare when only some can be read while locked` and nothing else — the other 38 tests in the class stay green.

Ran green:

```
./gradlew :data:testDebugUnitTest :app:testDebugUnitTest --offline
./gradlew :data:lintDebug :app:lintDebug --offline
```

Not manually verified on a device — reaching the state requires forcing it, per the section above. Suggested manual pass, since the change touches every vault read: backup and reimport a vault both with and without a passcode set, send a transaction after unlocking (signing needs a complete set), and confirm balances still refresh while locked.

## Summary by CodeRabbit

* **Bug Fixes**
  * Vault reads now follow an all-or-nothing approach for key shares.
  * If any encrypted share cannot be read while a vault is locked, all shares are discarded instead of returning incomplete data.
  * Unlocked vaults continue to return available shares, while unreadable shares correctly trigger an error.
  * Improved handling prevents partially reconstructed vault data from being exported.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/data/mappers/MapVaultToProto.kt
+++ b/app/src/main/java/com/vultisig/wallet/data/mappers/MapVaultToProto.kt
@@ -44,6 +44,11 @@ internal class MapVaultToProtoImpl @Inject constructor() : MapVaultToProto {
         // of storage because the data key was unavailable. Exporting that would hand the user a
         // .vult that reports success, restores cleanly, and can never sign — the worst shape a
         // backup failure can take, because it is only discovered when the backup is needed.
+        //
+        // Empty is enough to test for because the read is all-or-nothing: a vault that could not
+        // be fully decrypted arrives with none of its shares rather than the ones that happened to
+        // open. Storage owns that guarantee — see VaultRepositoryImpl.toKeyShares — because the
+        // count of shares a vault *should* have exists only there.
         if (from.keyshares.isEmpty()) throw VaultKeysharesUnavailableException()
         return VaultProto(
             name = from.name,
--- a/app/src/test/java/com/vultisig/wallet/data/mappers/MapVaultToProtoImplTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/data/mappers/MapVaultToProtoImplTest.kt
@@ -146,4 +146,22 @@ internal class MapVaultToProtoImplTest {
         assertNull(mapper.exportableOrNull(vault().copy(keyshares = emptyList())))
         assertNotNull(mapper.exportableOrNull(vault()))
     }
+
+    /**
+     * A vault whose stored rows were part plaintext, part encrypted, read with no data key, is
+     * refused.
+     *
+     * The other half of this claim lives in `VaultRepositoryImplTest.get drops every keyshare when
+     * only some can be read while locked`, which is where that vault is actually produced —
+     * `VaultRepositoryImpl` is internal to the data module, so the two halves cannot meet in one
+     * test. This one pins the side the mapper owns: given the all-or-nothing read, checking for
+     * emptiness is what refuses a partial keyshare set, and it must keep doing so.
+     */
+    @Test
+    fun `a partially readable keyshare set reaches the mapper as empty and is refused`() {
+        val fromPartialLockedRead = vault().copy(keyshares = emptyList())
+
+        assertFailsWith<VaultKeysharesUnavailableException> { mapper(fromPartialLockedRead) }
+        assertNull(mapper.exportableOrNull(fromPartialLockedRead))
+    }
 }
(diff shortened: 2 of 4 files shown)


johnnyluo (approved):
LGTM

PR #5577, opened 2026-08-10
fix(sui): migrate off JSON-RPC to GraphQL RPC before mid-Oct decommission

### LINKED ISSUE #5506: Sui: migrate off JSON-RPC before mid-Oct decommission (gRPC/GraphQL + failover) — HIGH, hard deadline
## What
Migrate Android Sui transport off JSON-RPC before Sui decommissions it (**wind-down began week of 2026-07-27; full decommission mid-October 2026** — documented in vultisig-sdk `core/chain/chains/sui/config.ts:8-21`; the SDK already moved to gRPC in vultisig-sdk#1583). Android is 100% JSON-RPC on the hardcoded `https://sui-rpc.publicnode.com` (`SuiApi.kt:53`, all `suix_*/sui_*` methods). **Untouched, Sui breaks in the app ~mid-October.**

## Scope
- Move to gRPC (`fullnode.mainnet.sui.io:443`) or GraphQL (`graphql.mainnet.sui.io` — the SDK's React Native path proves the surface).
- Add endpoint failover (today: single host, no fallback, no custom-RPC support for Sui — `CustomRpcSupportedChains.kt:16-40` is EVM+Cosmos only).

## Priority
High — hard external deadline. Backlogged for CW 34/35. Sibling: iOS vultisig/vultisig-ios#5025.

aminsato:
## Description

Closes #5506

Sui is retiring JSON-RPC. Shutdown on Foundation mainnet full nodes began the week of **2026-07-27** and full decommission (code removal) lands **mid-October 2026**, after which no provider can serve it. Android was 100% JSON-RPC on a single hardcoded host (`sui-rpc.publicnode.com`) with no fallback, so untouched, every Sui balance, fee estimate and broadcast breaks in about nine weeks.

All eight calls now go through Sui GraphQL RPC, and the transport gained endpoint failover.

| was | now |
|---|---|
| `suix_getBalance` | `address.balance(coinType)` |
| `suix_getReferenceGasPrice` | `epoch.referenceGasPrice` |
| `suix_getAllCoins` | `address.objects(filter: { type: "0x2::coin::Coin" })`, paginated |
| `suix_getCoinMetadata` | `coinMetadata(coinType)` |
| `sui_executeTransactionBlock` | `mutation executeTransaction` |
| `sui_dryRunTransactionBlock` | `simulateTransaction` |
| `sui_getTransactionBlock` | `transaction(digest)` |
| `sui_getLatestCheckpointSequenceNumber` | `checkpoint.sequenceNumber` |

### Why GraphQL and not gRPC

Both are supported replacements. GraphQL is a plain JSON POST that the existing Ktor client already speaks; gRPC would pull a `grpc-okhttp` / `grpc-kotlin` stack into a module whose protobuf plugin only generates kotlinx-serialization classes today. The SDK reached the same conclusion for its non-streaming platforms (vultisig-sdk#1583).

### Response differences absorbed at the boundary

The public `SuiApi` surface is unchanged, so no call sites move. Three differences would have failed silently rather than loudly:

- **Coin types** arrive wrapped as `0x2::coin::Coin<T>` and zero-padded. Unwrapped and stripped, they match `suix_getAllCoins` byte-for-byte. This matters beyond comparison: `SuiTokenFinder` *persists* this string as a discovered token's `contractAddress`, so without it the same token rediscovers as a second entry differing only by padding.
- **Execution status** is `SUCCESS`/`FAILURE` where the app matches the lowercase JSON-RPC spelling — unhandled, every confirmed transaction reads as pending until the poll times out.
- **A never-held coin** resolves to `null`, which is a genuine zero, not an upstream fault. GraphQL answers HTTP 200 even when it refuses, so `data: null` with a populated `errors` array is the real failure signal.

### Failover

Scoped to transport faults (unreachable host, TLS, 5xx). A populated `errors` array is the node's considered answer to a well-formed request, so it is raised immediately rather than replayed against every remaining host. Retrying a broadcast is safe — Sui identifies a transaction by the digest of its signed bytes, so re-submitting identical bytes is idempotent.

## Which feature is affected?

- [x] Send — Sui (balances, token discovery, fee estimation, broadcast, status)

Sui transaction sent through this build:
**https://suiscan.xyz/mainnet/tx/42j1UPdbzENXw9ZQUkh6yUza8AmMWrKGAigYH95QFSd9**

## Test plan

**Unit tests** — 2753 `:data` tests pass, lint clean. 32 in `SuiApiTest` (error surfacing, coin-type unwrap and padding strip, pagination, status lowercasing, not-found vs refusal, metadata nulls, response-shape rejection) and 9 new in `SuiGraphQlTransportTest` (failover on 5xx and on unreachable host, no failover on a GraphQL error payload, cancellation propagation, empty endpoint list).

**Validated against mainnet** — beyond mocked tests:

- Diffed both transports for the same live address: `version`, `balance`, `digest` and `previousTransaction` match byte-for-byte, and coin types match exactly once normalized.
- Ran the production `simulateTransaction` document against real on-chain transaction bytes and got `status: SUCCESS` with the gas summary in the mapped shape — this confirms the dry-run envelope, where the BCS bytes now travel inside a JSON-encoded `sui.rpc.v2.Transaction`.
- Broadcast and status confirmed by the real send linked above (`SUCCESS`, checkpoint 309017858).

**In-app** — Sui balances render, token discovery lists held coins with correct tickers and decimals, custom-token lookup resolves, fee estimate populates on the verify screen, send broadcasts and reaches confirmed.

## Additional context

Two follow-ups this surfaced, neither blocking:

1. **The failover list ships with one host.** `graphql.mainnet.sui.io` is the only public Sui GraphQL endpoint that answers; the other candidates I probed 403 or fail TLS, and `fullnode.mainnet.sui.io/graphql` is a 404 (gRPC only). The transport walks the whole list, so adding a host is a one-line change — but I would rather ship one verified endpoint than seed the list with hosts that do not respond.
2. **`api.vultisig.com/sui` is itself JSON-RPC.** Verified live — it answers `suix_getReferenceGasPrice`. Our own Sui proxy is subject to the same mid-October decommission and is the natural second endpoint here, but that is backend work in another repo. Worth filing separately.

Also worth flagging: the iOS sibling vultisig/vultisig-ios#5025 is still open and unassigned, and `Endpoint.swift` still hardcodes the same retired host.

`getLatestCheckpointSequenceNumber` has no call sites anywhere in the app. It is migrated for completeness but is dead surface — a candidate for removal in a separate cleanup.

## Summary by CodeRabbit

* **New Features**
  * Sui network operations now use GraphQL for balances, coins, metadata, transactions, simulations, status checks, and checkpoints.
  * Added paginated coin retrieval, improved coin-type normalization, endpoint failover, and clearer error reporting.

* **Bug Fixes**
  * Improved handling of malformed, incomplete, unavailable, or failed Sui responses.
  * Preserved cancellation during network requests.
  * Improved transaction status reporting and failure messages.
  * Added safeguards for repeated or missing pagination data.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
@@ -20,6 +20,13 @@ import kotlinx.serialization.json.putJsonObject
 import timber.log.Timber
 import vultisig.keysign.v1.SuiCoin
 
+/**
+ * Backstop on the coin-object walk in [SuiApi.getAllCoins]: at 50 objects per page this is 5000
+ * coin objects, far beyond any real wallet, so reaching it means the connection is misbehaving
+ * rather than that someone genuinely holds that many.
+ */
+internal const val SUI_MAX_COIN_PAGES = 100
+
 interface SuiApi {
 
     suspend fun getBalance(address: String, contractAddress: String): BigInteger
@@ -70,7 +77,12 @@ internal class SuiApiImpl @Inject constructor(http: HttpClient, private val json
         // An address that has never held the coin type resolves to null rather than an error —
         // that is a genuine zero, not an upstream fault. A malformed address is refused by the node
         // with a populated `errors` array, which the transport has already raised by this point.
-        return response.address?.balance?.totalBalance?.toBigIntegerOrNull() ?: BigInteger.ZERO
+        val totalBalance = response.address?.balance?.totalBalance ?: return BigInteger.ZERO
+
+        // A present-but-unparsable amount is a node contract violation, not an empty wallet, so it
+        // is raised rather than collapsed into the same zero an absent balance returns.
+        return totalBalance.toBigIntegerOrNull()
+            ?: throw SuiRpcException("unparsable balance: $totalBalance")
     }
 
     override suspend fun getReferenceGasPrice(): BigInteger =
@@ -84,6 +96,7 @@ internal class SuiApiImpl @Inject constructor(http: HttpClient, private val json
     override suspend fun getAllCoins(address: String): List<SuiCoin> {
         val allCoins = mutableListOf<SuiCoin>()
         var cursor: String? = null
+        var pagesFetched = 0
 
         // The object connection is paginated. Follow hasNextPage/endCursor so a wallet whose
         // objects span multiple pages doesn't return a truncated set — otherwise a token whose
@@ -117,7 +130,24 @@ internal class SuiApiImpl @Inject constructor(http: HttpClient, private val json
                 )
             }
 
-            cursor = objects.pageInfo.endCursor.takeIf { objects.pageInfo.hasNextPage }
+            pagesFetched++
+
+            // Termination cannot rest on the node alone: a `hasNextPage` that never flips, or an
+            // `endCursor` that never advances, would spin here forever while `allCoins` grows
+            // without bound. Stop on a repeated cursor, and cap the walk at [SUI_MAX_COIN_PAGES].
+            val nextCursor = objects.pageInfo.endCursor.takeIf { objects.pageInfo.hasNextPage }
+            if (
+                nextCursor != null && (nextCursor == cursor || pagesFetched >= SUI_MAX_COIN_PAGES)
+            ) {
+                Timber.w(
+                    "Sui coin pagination stopped after %d pages with more reported; cursor %s",
+                    pagesFetched,
+                    if (nextCursor == cursor) "did not advance" else "budget exhausted",
+                )
+                cursor = null
+            } else {
+                cursor = nextCursor
+            }
         } while (cursor != null)
 
         return allCoins
@@ -178,7 +208,7 @@ internal class SuiApiImpl @Inject constructor(http: HttpClient, private val json
 
         val error = effects.errorMessage()
         if (error.isNotEmpty()) {
-            throw Exception("Simulation Error: $error")
+            error("Simulation Error: $error")
         }
 
         val gasSummary =
@@ -252,7 +282,9 @@ internal class SuiApiImpl @Inject constructor(http: HttpClient, private val json
         try {
             json.decodeFromJsonElement<T>(this)
         } catch (e: SerializationException) {
-            throw SuiRpcException("unexpected response shape: ${e.message}")
+            // Carry the cause: the message alone does not say which field of which selection set
+            // failed, and a decode failure against a live node is hard to place without the trace.
+            throw SuiRpcException("unexpected response shape: ${e.message}", cause = e)
         }
 
     private companion object {
@@ -488,11 +520,14 @@ private fun TransactionEffectsFields.errorMessage(): String =
  * `::coin::USDC` and `::coin::usdc` are genuinely distinct Move types.
  *
  * Returns null for anything that isn't a generic instantiation, which the type filter excludes.
+ *
+ * Internal rather than private so tests that build coins from a raw GraphQL payload normalize
+ * through this exact function instead of a copy that can drift from it.
  */
-private fun unwrapCoinType(repr: String): String? {
+internal fun unwrapCoinType(repr: String): String? {
     val open = repr.indexOf('<')
     val close = repr.lastIndexOf('>')
-    if (open < 0 || close <= open) return null
+    if (open !in 0..<close) return null
     val coinType = repr.substring(open + 1, close)
 
     val addressEnd = coinType.indexOf("::")
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
@@ -3,6 +3,7 @@ package com.vultisig.wallet.data.api.chains
 import com.vultisig.wallet.data.api.models.SuiExecutionStatus
 import com.vultisig.wallet.data.api.models.SuiTransactionBlockEffects
 import com.vultisig.wallet.data.api.models.SuiTransactionBlockResponse
+import com.vultisig.wallet.data.crypto.SuiHelper
 import io.ktor.client.HttpClient
 import java.math.BigInteger
 import javax.inject.Inject
@@ -23,7 +24,8 @@ import vultisig.keysign.v1.SuiCoin
 /**
  * Backstop on the coin-object walk in [SuiApi.getAllCoins]: at 50 objects per page this is 5000
  * coin objects, far beyond any real wallet, so reaching it means the connection is misbehaving
- * rather than that someone genuinely holds that many.
+ * rather than that someone genuinely holds that many — which is why exhausting it raises instead of
+ * returning the objects collected so far.
  */
 internal const val SUI_MAX_COIN_PAGES = 100
 
@@ -74,9 +76,13 @@ internal class SuiApiImpl @Inject constructor(http: HttpClient, private val json
                 )
                 .decode<BalanceResponse>()
 
-        // An address that has never held the coin type resolves to null rather than an error —
-        // that is a genuine zero, not an upstream fault. A malformed address is refused by the node
-        // with a populated `errors` array, which the transport has already raised by this point.
+        // An address that has never held the coin type resolves to a real zero `Balance`, not to
+        // null — checked against mainnet. Null is the node's answer when it has no checkpoint in
+        // scope to read the balance at, and there is nothing to show for that but zero either: it
+        // is an absent reading, not a smaller one, and the alternative is an error banner on an
+        // ordinary wallet screen. A malformed address is a different case entirely — the node
+        // refuses it with a populated `errors` array, which the transport has already raised by
+        // this point.
         val totalBalance = response.address?.balance?.totalBalance ?: return BigInteger.ZERO
 
         // A present-but-unparsable amount is a node contract violation, not an empty wallet, so it
@@ -118,11 +124,19 @@ internal class SuiApiImpl @Inject constructor(http: HttpClient, private val json
 
             objects.nodes.forEach { node ->
                 val coinType = node.contents?.type?.repr?.let(::unwrapCoinType) ?: return@forEach
+                // The three fields that make up the signed `Sui.ObjectRef` are dropped with the
+                // coin when any is absent, the same way an undescribable coin type is. Defaulting
+                // them would be worse than losing the object: a blank digest is signed into the
+                // bytes every device in the ceremony commits to and is only rejected at broadcast,
+                // and a blank version throws NumberFormatException inside SuiHelper instead.
+                val coinObjectId = node.address ?: return@forEach
+                val version = node.version ?: return@forEach
+                val digest = node.digest ?: return@forEach
                 allCoins.add(
                     SuiCoin(
-                        coinObjectId = node.address.orEmpty(),
-                        version = node.version?.toString().orEmpty(),
-                        digest = node.digest.orEmpty(),
+                        coinObjectId = coinObjectId,
+                        version = version.toString(),
+                        digest = digest,
                         balance = node.contents.json?.balance.orEmpty(),
                         previousTransaction = node.previousTransaction?.digest.orEmpty(),
                         coinType = coinType,
@@ -134,20 +148,25 @@ internal class SuiApiImpl @Inject constructor(http: HttpClient, private val json
 
             // Termination cannot rest on the node alone: a `hasNextPage` that never flips, or an
             // `endCursor` that never advances, would spin here forever while `allCoins` grows
-            // without bound. Stop on a repeated cursor, and cap the walk at [SUI_MAX_COIN_PAGES].
+            // without bound. Both bounds raise rather than return what was collected — a partial
+            // coin list is not a smaller wallet. Coin selection would read it as one, and the
+            // caller gets an "insufficient balance" rejection with nothing pointing at the walk;
+            // the log that used to carry that detail is compiled out of release builds, where the
+            // only planted Timber tree is the debug one.
             val nextCursor = objects.pageInfo.endCursor.takeIf { objects.pageInfo.hasNextPage }
-            if (
-                nextCursor != null && (nextCursor == cursor || pagesFetched >= SUI_MAX_COIN_PAGES)
-            ) {
-                Timber.w(
-                    "Sui coin pagination stopped after %d pages with more reported; cursor %s",
-                    pagesFetched,
-                    if (nextCursor == cursor) "did not advance" else "budget exhausted",
+            if (nextCursor != null && nextCursor == cursor) {
+                throw SuiRpcException(
+                    "coin pagination stalled: the node reported more pages but returned the same " +
+                        "cursor after $pagesFetched pages"
                 )
-                cursor = null
-            } else {
-                cursor = nextCursor
             }
+            if (nextCursor != null && pagesFetched >= SUI_MAX_COIN_PAGES) {
+                throw SuiRpcException(
+                    "coin pagination exceeded the $SUI_MAX_COIN_PAGES page budget with more pages " +
+                        "still reported after ${allCoins.size} coin objects"
+                )
+            }
+            cursor = nextCursor
         } while (cursor != null)
 
         return allCoins
@@ -382,6 +401,11 @@ internal class SuiApiImpl @Inject constructor(http: HttpClient, private val json
  * `null` rather than omitting it, and each query selects only part of a shared type (a simulation
  * asks for `gasEffects`, a status poll for `checkpoint`), so no single field is guaranteed present.
  *
+ * The exception is a connection's own `pageInfo`/`nodes`, which the schema declares non-null and
+ * every page of the walk selects. Giving those a default would decode a response missing them as a
+ * complete empty page and silently truncate the coin list; without one the decode fails and
+ * [SuiApiImpl.decode] raises it as the node fault it is.
+ *
  * Numeric fields follow the schema's scalar, not convenience: `BigInt` arrives as a JSON string
  * (and can exceed [Long]), `UInt53` as a JSON number. The injected [Json] is not lenient, so typing
  * one as the other fails to decode.
@@ -401,13 +425,9 @@ internal class SuiApiImpl @Inject constructor(http: HttpClient, private val json
 @Serializable private data class CoinObjectOwner(val objects: CoinObjectConnection? = null)
 
 @Serializable
-private data class CoinObjectConnection(
-    val pageInfo: PageInfo = PageInfo(),
-    val nodes: List<CoinObjectNode> = emptyList(),
-)
+private data class CoinObjectConnection(val pageInfo: PageInfo, val nodes: List<CoinObjectNode>)
 
-@Serializable
-private data class PageInfo(val hasNextPage: Boolean = false, val endCursor: String? = null)
+@Serializable private data class PageInfo(val hasNextPage: Boolean, val endCursor: String? = null)
 
 @Serializable
 private data class CoinObjectNode(
@@ -506,18 +526,23 @@ private fun TransactionEffectsFields.errorMessage(): String =
  * Two adjustments, both required to keep the transport swap invisible to the rest of the app:
  * 1. The object connection reports the wrapper struct, whereas the app compares against the bare
  *    `T`. A wrapper string would match no known coin and turn every native send into a token send.
- * 2. GraphQL always spells the address zero-padded (`0x000…002::sui::SUI`) where JSON-RPC returned
- *    it stripped (`0x2::sui::SUI`). Comparisons already go through
+ * 2. GraphQL always spells addresses zero-padded (`0x000…002::sui::SUI`) where JSON-RPC returned
+ *    them stripped (`0x2::sui::SUI`). Comparisons already go through
  *    [SuiHelper.isSameSuiCoinType][com.vultisig.wallet.data.crypto.SuiHelper] and tolerate either,
  *    but [SuiTokenFinder][com.vultisig.wallet.data.usecases.SuiTokenFinder] *persists* this exact
  *    string as a discovered token's `contractAddress` — so stripping here keeps a token discovered
  *    after the migration identical to the same token discovered before it, rather than a second
  *    entry that only differs by padding.
  *
+ * The stripping is [SuiHelper.normalizeSuiCoinType][com.vultisig.wallet.data.crypto.SuiHelper]
+ * itself rather than a local copy of it: the two must agree exactly — this function produces the
+ * `contractAddress` that function later compares against — and an inline reimplementation is free
+ * to drift from it, which is how nested generic addresses came to be normalized in one and not the
+ * other. Both leave Move module and struct identifiers case-sensitive, because `::coin::USDC` and
+ * `::coin::usdc` are genuinely distinct Move types.
+ *
  * Verified against a live mainnet address: with this normalization the returned coin types match
- * `suix_getAllCoins` exactly. Only the address is normalized — Move module and struct identifiers
- * stay case-sensitive, matching [SuiHelper][com.vultisig.wallet.data.crypto.SuiHelper], because
- * `::coin::USDC` and `::coin::usdc` are genuinely distinct Move types.
+ * `suix_getAllCoins` exactly.
  *
  * Returns null for anything that isn't a generic instantiation, which the type filter excludes.
  *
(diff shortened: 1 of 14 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
```diff
@@ -46,196 +40,468 @@ interface SuiApi {
...
+                )
+                .decode<BalanceResponse>()
 
-        return response.result?.jsonObject?.get("totalBalance")?.jsonPrimitive?.content?.let {
-            BigInteger(it)
-        } ?: error(response.error.describe("Failed to get sui balance"))
+        // An address that has never held the coin type resolves to null rather than an error —
+        // that is a genuine zero, not an upstream fault. A malformed address is refused by the node
+        // with a populated `errors` array, which the transport has already raised by this point.
+        return response.address?.balance?.totalBalance?.toBigIntegerOrNull() ?: BigInteger.ZERO
```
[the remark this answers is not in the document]
aminsato: Valid — fixed in cb5dffae3.

Right that both cases collapsed: an absent `totalBalance` is a genuine zero (an address that never held the coin type), but a present-but-unparsable one is a node contract violation being shown to someone as an empty wallet — the exact failure `decode()` exists to prevent. The absent case returns early with `BigInteger.ZERO`; a value that fails to parse now raises `SuiRpcException` carrying the offending string. Covered by `getBalance rejects an unparsable balance instead of reading it as zero`.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
```diff
@@ -46,196 +40,468 @@ interface SuiApi {
...
+                        previousTransaction = node.previousTransaction?.digest.orEmpty(),
                         coinType = coinType,
                     )
                 )
             }
 
-            val hasNextPage = result["hasNextPage"]?.jsonPrimitive?.booleanOrNull ?: false
-            cursor = if (hasNextPage) result["nextCursor"]?.jsonPrimitive?.content else null
+            cursor = objects.pageInfo.endCursor.takeIf { objects.pageInfo.hasNextPage }
         } while (cursor != null)
```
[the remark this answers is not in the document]
aminsato: Valid — fixed in cb5dffae3.

Termination rested entirely on data from the other side, and you are right that the fixture hid it: both pages returned `cursor-1`, so the mock answered by call order and the test would have passed against a loop that never advanced.

`getAllCoins` now stops when `endCursor` does not advance and caps the walk at `SUI_MAX_COIN_PAGES` (100 pages x 50 objects = 5000, far past any real wallet), logging when it truncates so a short list is never silent. Three tests: pages now carry distinct cursors and the test asserts request two actually sends `page-1-cursor`; a repeated cursor terminates; and a node that advances the cursor forever stops at the budget.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
```diff
@@ -46,196 +40,468 @@ interface SuiApi {
...
+                .decode<SimulateTransactionResponse>()
+                .simulateTransaction
+                ?.effects ?: throw SuiRpcException("failed to dry run transaction")
 
-        if (dryRunResponse.effects.status.error.isNotEmpty()) {
-            throw Exception("Simulation Error: ${dryRunResponse.effects.status.error}")
+        val error = effects.errorMessage()
+        if (error.isNotEmpty()) {
+            throw Exception("Simulation Error: $error")
         }
```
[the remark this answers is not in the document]
aminsato: Valid — fixed in cb5dffae3, now `error("Simulation Error: $error")`. Message text and the `assertFailsWith<Exception>` assertion are unchanged, and it lands on the same `IllegalStateException` the rest of the file raises.

Also took the related nitpick: `SuiRpcException` gained a `cause` parameter and `decode()` passes the `SerializationException` through, since the message alone does not identify which field of which selection set failed.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiGraphQl.kt
```diff
@@ -0,0 +1,129 @@
...
+        variables: JsonObject = JsonObject(emptyMap()),
+    ): JsonObject {
+        var lastTransportFailure: Exception? = null
+
+        for (endpoint in endpoints) {
+            val response =
+                try {
+                    http
+                        .post(endpoint) { setBody(SuiGraphQlRequest(document, variables)) }
+                        .bodyOrThrow<SuiGraphQlResponse>()
```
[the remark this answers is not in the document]
aminsato: Skipping this one, with a note for when a second endpoint lands.

The premise checks out — `NetworkModule` sets OkHttp `connectTimeout`/`readTimeout`/`writeTimeout` at 15s each and no `HttpTimeout` plugin is installed anywhere, so there is no end-to-end cap. But a per-request `timeout {}` only takes effect if `HttpTimeout` is installed on the *shared* client, which would put a ceiling on every network call in the app — including the keysign relay long-polls, which are deliberately open far longer than 15s. That is a much larger change than this PR should carry.

Bounded in practice today: one endpoint, and GraphQL goes out as POST, which `HttpRequestRetry` excludes from both `retryOnExceptionIf` and `retryIf` (`isSafeMethod`), so there is no retry multiplier on top of the phase timeouts. Worth revisiting as its own change when a second host is added, since that is when the durations start compounding.

Took the sibling nitpick on this block though: failover was documented as transport-only but caught `Exception`, so a 4xx was replayed against every remaining host. It now rethrows `NetworkException` with a 4xx status immediately — client-side faults arrive as status 0 and still fail over, as does 5xx. New test `does not fail over on a 4xx`.

aminsato:
Went through the nitpicks as well — all six were real and are in cb5dffae3.

**`SuiApi.kt` — preserve the `SerializationException` as the cause.** `SuiRpcException` gained a `cause` parameter and `decode()` passes it through.

**`SuiGraphQl.kt` — failover fired on every non-cancellation exception.** The KDoc promised transport-level faults only while the code caught `Exception`, so a deterministic 4xx was replayed against every remaining host. Now rethrown immediately; client-side faults (status 0) and 5xx still fail over. Test: `does not fail over on a 4xx`.

**`SuiGraphQlTransportTest` — the test name promised more than it verified.** Renamed to `attempts every endpoint when all are unreachable`, which is what it asserts, and dropped the trailing unasserted sanity call — the 5xx test above it already covers recovery.

**`SuiGraphQlTransportTest` — local `jsonObject()` cast helper.** Removed; imports `kotlinx.serialization.json.jsonObject`.

**`SuiHelperTest` — `parseCoins` did not mirror `getAllCoins`.** This was the sharpest one: the parser unwrapped `Coin<T>` but skipped the address normalization, and the fixture used the already-stripped spelling, so the test passed by agreeing with itself rather than by matching the node. Took the long-term option you suggested — `unwrapCoinType` is `internal` now and the test calls the production function, so the two cannot drift again — and padded the first fixture coin to the real GraphQL spelling so the normalization is actually exercised.

**`SuiApiTest` — pagination test could not detect a non-advancing cursor.** Distinct `endCursor` per page, and the test now asserts request two sends `page-1-cursor`. `MockHttpClient.RequestCapture` keeps every body in order (`capturingRequestSequence`) instead of only the last, since asserting the last body cannot show that page two carried page one's cursor. Added the repeated-cursor and page-budget cases alongside it.

Full `:data` suite: 2757 tests, 0 failures.

data/src/test/kotlin/com/vultisig/wallet/data/api/chains/SuiGraphQlTransportTest.kt
```diff
@@ -0,0 +1,173 @@
...
+    fun `falls over to the next endpoint when the first answers 5xx`() = runTest {
+        val client =
+            MockHttpClient.respondingWithSequence(
+                HttpStatusCode.InternalServerError to "upstream down",
+                HttpStatusCode.OK to successBody,
+            )
+
+        val data = SuiGraphQlTransport(client, endpoints).query(QUERY)
+
+        assertEquals("100", data["epoch"]!!.jsonObject["referenceGasPrice"]!!.jsonPrimitive.content)
```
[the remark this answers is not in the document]
aminsato: Valid — fixed in 23263d8aa.

The `!!` predates this PR, but the point stands and this is the test where it costs the most: the regression it exists to catch is failover returning a body without an `epoch`, and that would have surfaced as a `NullPointerException` on the assertion line rather than as `expected "100" but was null`. Safe calls now, so `assertEquals` names what is missing.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiGraphQl.kt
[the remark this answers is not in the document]
aminsato: Thanks — no follow-up issue for now, please.

There is no second GraphQL endpoint on the horizon (the Vultisig proxy still speaks JSON-RPC and is subject to the same decommission), so an issue with no trigger would sit unactioned. The reasoning is captured in this thread and in the `SUI_GRAPHQL_ENDPOINTS` KDoc, which is where anyone adding a second host will be reading — that is the moment the end-to-end cap becomes worth designing, and it should be scoped then against whatever the relay long-poll requirements look like at the time.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
```diff
@@ -46,196 +47,496 @@ interface SuiApi {
...
                 allCoins.add(
                     SuiCoin(
-                        coinObjectId = obj["coinObjectId"]?.jsonPrimitive?.content ?: "",
-                        version = obj["version"]?.jsonPrimitive?.content ?: "",
-                        digest = obj["digest"]?.jsonPrimitive?.content ?: "",
-                        balance = obj["balance"]?.jsonPrimitive?.content ?: "",
-                        previousTransaction =
-                            obj["previousTransaction"]?.jsonPrimitive?.content ?: "",
+                        coinObjectId = node.address.orEmpty(),
+                        version = node.version?.toString().orEmpty(),
```
rkokhatskyi: `getAllCoins` defaults a null `version`/`digest` to `""` instead of dropping the coin, unlike the `coinType` check one line above. A null `digest` bakes a blank `objectDigest` into the signed `Sui.ObjectRef` (`SuiHelper.kt:319`), risking a wasted TSS ceremony rejected only at broadcast; a null `version` throws `NumberFormatException` instead.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
```diff
@@ -46,196 +47,496 @@ interface SuiApi {
...
+ */
+internal fun unwrapCoinType(repr: String): String? {
+    val open = repr.indexOf('<')
+    val close = repr.lastIndexOf('>')
+    if (open !in 0..<close) return null
+    val coinType = repr.substring(open + 1, close)
+
+    val addressEnd = coinType.indexOf("::")
+    if (addressEnd < 0) return coinType
+    val address = coinType.substring(0, addressEnd).lowercase().removePrefix("0x").trimStart('0')
```
rkokhatskyi: `unwrapCoinType` and `SuiHelper.normalizeSuiCoinType` only strip the address before the first `::`. GraphQL zero-pads every nested address recursively (live-confirmed on `Coin<Balance<SUI>>`), so a nested-generic coin type no longer string-matches a pre-migration `contractAddress`, and `SuiHelper.kt:113`'s `check(tokenObjects.isNotEmpty())` blocks that token's send entirely.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
```diff
@@ -46,196 +47,496 @@ interface SuiApi {
...
-            val hasNextPage = result["hasNextPage"]?.jsonPrimitive?.booleanOrNull ?: false
-            cursor = if (hasNextPage) result["nextCursor"]?.jsonPrimitive?.content else null
+            pagesFetched++
+
+            // Termination cannot rest on the node alone: a `hasNextPage` that never flips, or an
+            // `endCursor` that never advances, would spin here forever while `allCoins` grows
+            // without bound. Stop on a repeated cursor, and cap the walk at [SUI_MAX_COIN_PAGES].
+            val nextCursor = objects.pageInfo.endCursor.takeIf { objects.pageInfo.hasNextPage }
+            if (
+                nextCursor != null && (nextCursor == cursor || pagesFetched >= SUI_MAX_COIN_PAGES)
```
rkokhatskyi: `getAllCoins` silently truncates past its shared 5000-object budget, pooled across every coin type an address holds (`SuiApi.kt:138-150`), logging only via `Timber.w`, a no-op in release builds. `BlockChainSpecificRepository.kt:522`'s coin selection then sees a generic "insufficient balance" rejection with no way to diagnose truncation as the cause.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
```diff
@@ -46,196 +47,496 @@ interface SuiApi {
...
+
+@Serializable private data class EpochData(val referenceGasPrice: String? = null)
+
+@Serializable private data class AllCoinsResponse(val address: CoinObjectOwner? = null)
+
+@Serializable private data class CoinObjectOwner(val objects: CoinObjectConnection? = null)
+
+@Serializable
+private data class CoinObjectConnection(
+    val pageInfo: PageInfo = PageInfo(),
```
rkokhatskyi: `CoinObjectConnection`'s `pageInfo`/`nodes` default to `PageInfo()`/`emptyList()`, so a response missing the schema-NON_NULL `objects.pageInfo`/`objects.nodes` keys decodes as a complete empty page instead of failing. Silently truncates coin pagination with no log, unlike the `address?.objects` null check one line above.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
```diff
@@ -46,196 +47,496 @@ interface SuiApi {
...
                     },
-            )
+                )
+                .decode<BalanceResponse>()
 
-        return response.result?.jsonObject?.get("totalBalance")?.jsonPrimitive?.content?.let {
-            BigInteger(it)
-        } ?: error(response.error.describe("Failed to get sui balance"))
-    }
+        // An address that has never held the coin type resolves to null rather than an error —
```
rkokhatskyi: This comment says a never-held coin type resolves `balance` to null. Sui's GraphQL docs say the opposite: `Address.balance` returns null only when no checkpoint is in scope; a never-held type returns a real zero `Balance` instead. No functional bug since both paths yield `BigInteger.ZERO`, but the reasoning is backwards.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiGraphQl.kt
```diff
@@ -0,0 +1,148 @@
...
+@Serializable
+internal data class SuiGraphQlErrorExtensions(@SerialName("code") val code: String? = null)
+
+private const val UNKNOWN_GRAPHQL_ERROR = "unknown error"
+
+/**
+ * A Sui GraphQL application error — the node parsed the request and refused it (bad address,
+ * undecodable transaction, indexer outage).
+ *
+ * Extends [IllegalStateException] because the JSON-RPC implementation this replaces reported the
```
rkokhatskyi: This KDoc says `SuiRpcException` extends `IllegalStateException` because its JSON-RPC predecessor used `error(...)`, naming `SuiStatusProvider` as the caller this matters for. The old `SuiRpcException` actually extended plain `Exception` and was thrown via `throw SuiRpcException(error)`, never `error(...)`, exactly for `checkStatus` — the rationale is backwards there.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
```diff
@@ -46,196 +47,496 @@ interface SuiApi {
...
+internal fun unwrapCoinType(repr: String): String? {
+    val open = repr.indexOf('<')
+    val close = repr.lastIndexOf('>')
+    if (open !in 0..<close) return null
+    val coinType = repr.substring(open + 1, close)
+
+    val addressEnd = coinType.indexOf("::")
+    if (addressEnd < 0) return coinType
+    val address = coinType.substring(0, addressEnd).lowercase().removePrefix("0x").trimStart('0')
+    return "0x" + address.ifEmpty { "0" } + coinType.substring(addressEnd)
```
rkokhatskyi: `unwrapCoinType` reimplements `SuiHelper.normalizeSuiCoinType`'s address normalization inline instead of calling it, linked only by a prose comment. Same module, no import cycle — `SuiHelperTest` already avoids this exact drift risk by importing the production parser instead of reimplementing it. Should delegate to `SuiHelper.normalizeSuiCoinType` directly.

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
aminsato: Fixed in fd979dafb.

Confirmed, and the inconsistency with the line above was the giveaway. `coinObjectId`, `version` and `digest` are exactly the three fields `SuiHelper.toObjectRef()` turns into the signed `Sui.ObjectRef`, so a blank digest gets committed to by every device in the ceremony and comes back only as a broadcast rejection, and a blank version does not even reach that — `version.toLong()` throws first.

All three now drop the coin object the way an undescribable `coinType` does.

Worth noting for the record that this one is inherited rather than introduced here: the JSON-RPC implementation had the same `?: ""` on all four fields. The GraphQL schema declares `address`/`version`/`digest` non-null on `Object`, so in practice it should not fire either way — but the failure mode if it ever does is bad enough that a defensive default is the wrong default.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
aminsato: Fixed in fd979dafb — this was the most serious of the seven.

Reproduced your live finding directly:

```
$ curl -s -X POST https://graphql.mainnet.sui.io/graphql -d \
  '{"query":"query { type(type: \"0x2::coin::Coin<0x2::table::Table<0x2::sui::SUI, 0x2::sui::SUI>>\") { repr } }"}'

"0x000…002::coin::Coin<0x000…002::table::Table<0x000…002::sui::SUI,0x000…002::sui::SUI>>"
```

Every address padded, and `suix_getAllCoins` on the JSON-RPC host still returns `0x2::sui::SUI` short — so for any coin type that is itself a generic instantiation (LP tokens being the obvious case), the two spellings diverge on everything but the head, and `isSameSuiCoinType` was blind to it.

`normalizeSuiCoinType` now normalizes every address in the type rather than only the leading one. It splits on the generic delimiters and normalizes each qualified segment's address; a segment with no `::` is a primitive argument (`u64`, `bool`, `vector`) and passes through untouched, since rewriting `u64` as an address would invent a type that names nothing. Whitespace is dropped too, because `, ` and `,` between type arguments are two nodes' spelling of the same separator and a stored pre-migration string may carry either.

Four tests: nested padding compares equal, nested struct names stay case-sensitive, primitives survive, and the `getAllCoins` path produces `0x2::spot_dex::LP<0x2::sui::SUI,0x2b::coin::COIN>` from the fully-padded repr.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
aminsato: Fixed in fd979dafb.

You are right that the log was doing no work — `VoltixApplication` plants `Timber.DebugTree()` under `if (BuildConfig.DEBUG)` and nothing else, so in release the one artifact explaining a short list did not exist. Which leaves coin selection reading a truncated wallet as a smaller wallet, and the user getting `checkCoinsCover`'s insufficient-balance message for a wallet that is not short of anything.

Both bounds now raise a `SuiRpcException` instead of returning what was collected:

- stalled cursor — `"coin pagination stalled: the node reported more pages but returned the same cursor after N pages"`
- budget exhausted — `"coin pagination exceeded the 100 page budget with more pages still reported after N coin objects"`

The budget one is the judgement call: a wallet genuinely holding >5000 coin objects now fails its send outright where before it would have tried with an arbitrary 5000-object subset. I think that is the right trade — the connection ordering is not by balance, so the subset is not the "best 5000" and a send built on it is already a coin flip, and 5000 is far enough past any real Vultisig wallet that reaching it says more about the node than about the holder. An honest error that names the walk beats a silently wrong one.

`SuiTokenFinder` is unaffected in shape: it already catches and logs to an empty list, so a misbehaving node costs discovery that cycle rather than wiping tokens.

Both existing tests flipped from asserting a truncated size to asserting the raise and its message.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
aminsato: Fixed in fd979dafb.

Both defaults are gone — `CoinObjectConnection(val pageInfo: PageInfo, val nodes: List<CoinObjectNode>)`, and `PageInfo.hasNextPage` loses its `false` default for the same reason (a missing `hasNextPage` reading as "no more pages" is the same silent truncation one level down). `endCursor` keeps its null, which the schema does allow.

A response missing either now fails to decode and `decode()` raises it as `"unexpected response shape"`, which is the treatment the rest of the file's malformed-response cases already get. The class KDoc explains why these two are the carve-out from the everything-is-nullable rule that holds for the rest of the models: those are nullable because each query selects a subset of a shared type, whereas a connection's `pageInfo`/`nodes` are non-null in the schema and selected on every page of the walk.

Test covers both directions — `objects` with only `nodes`, and `objects` with only `pageInfo`.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
aminsato: Fixed in fd979dafb — comment only, as you say.

Checked it against mainnet rather than taking either reading on trust:

```
$ curl -s -X POST https://graphql.mainnet.sui.io/graphql -d '{"query":"query { address(
    address: \"0x…005\") { balance(coinType: \"0xdee9::deep::DEEP\") { totalBalance } } }"}'

{"data":{"address":{"balance":{"totalBalance":"0"}}}}
```

A real zero `Balance` for a type that address has never held, exactly as the docs say. So null means the node had no checkpoint in scope, not "never held it".

The comment now says that, and gives the actual reason the null still maps to `BigInteger.ZERO`: it is an absent reading rather than a smaller one, there is nothing to show for it but zero, and the alternative is an error banner on an ordinary wallet screen. The test's comment was making the same wrong claim and is updated with it.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiGraphQl.kt
aminsato: Fixed in fd979dafb.

Checked `git show 068c0aacf^` and you are right about which half is wrong:

```kotlin
internal class SuiRpcException(val rpcError: RpcError) :
    Exception("Sui RPC error ${rpcError.code}: ${rpcError.message}")
```

thrown once, via `throw SuiRpcException(error)` in `checkStatus` — the one path that was *not* `error(...)`. The `error(...)` half of the rationale is sound but belongs to the other six call sites (`getBalance`, `getReferenceGasPrice`, `getAllCoins`, `getCoinMetadata`, `executeTransactionBlock`, `dryRunTransaction` all ended in `?: error(...)`), and `SuiStatusProvider` was the one caller it did not apply to.

The KDoc now keeps the two apart: `IllegalStateException` because that is what those six threw, and a second paragraph noting `checkStatus`'s predecessor extended plain `Exception` and that `SuiStatusProvider` catches this type by name rather than by supertype, so it is unaffected by the widening either way.

data/src/main/kotlin/com/vultisig/wallet/data/api/chains/SuiApi.kt
aminsato: Fixed in fd979dafb — `unwrapCoinType` is now three lines and ends in `SuiHelper.normalizeSuiCoinType(...)`.

Agreed on the reasoning, and this thread and the nested-generics one are the same bug seen from two sides: the drift the duplication invited is precisely that the two implementations disagreed about nested addresses. That is also why the delegation direction is this one rather than the reverse — this function produces the `contractAddress` that `normalizeSuiCoinType` is later asked to compare against, so they have to be the same code, not merely agree today.

The KDoc says that outright now instead of gesturing at it in prose, and cites `SuiHelperTest`'s precedent of importing the production parser.

johnnyluo (approved):
LGTM

PR #5579, opened 2026-08-10
fix(backup): a vault nobody exported no longer reads as backed up

### LINKED ISSUE #5578: Vault model has no backup flag — Android cannot tell whether a vault was ever exported
Android has no way to tell whether a vault was ever backed up. iOS and Windows both do.

`data/src/main/kotlin/com/vultisig/wallet/data/models/Vault.kt` has no backup field — no `isBackedUp`, nothing equivalent. iOS has `Vault.isBackedUp` and Windows has `isBackedUp` on its stored vault.

## Why it matters now

Under vultisig/vultisig-sdk#1838 we keep encrypting shares at rest, which means the key can be lost — Android invalidates its keystore key when the device lock or biometrics change. That is survivable, but only if the user has a `.vult`.

So we need to know who does. Without the flag we can't:
- gate turning on the passcode behind a backup (vultisig/vultisig-sdk#1845, screen 1)
- prompt a user who has never exported
- tell someone in the key-is-gone state whether they can actually recover (#5566)

## What to build

Add the flag to the Vault model and its Room entity.

Two rules, both already right on the other clients:

- **Set it only after a complete, verified export.** Not on opening the backup screen, not on a partial write. Depends on #5565 — right now a partial export can succeed, so the flag would be recording a dead backup.
- **Clear it on reshare.** A reshare changes the key shares, which makes any existing `.vult` stale. iOS does this at `KeygenView.swift:334`; Windows starts reshared vaults at `isBackedUp: false`.

Also worth matching iOS's backup-reminder banner behaviour once the flag exists.

## Done when

- [ ] `Vault` and its Room entity carry a backup flag, migrated for existing vaults (default false)
- [ ] Set only after a complete export that passed the #5565 check
- [ ] Cleared on reshare, key import and any flow that changes the shares
- [ ] Existing vaults default to not-backed-up rather than being assumed safe

Ref: vultisig/vultisig-sdk#1838, vultisig/vultisig-sdk#1845

aminsato:
Closes #5578

## One correction to the issue first

The issue says `Vault` has "no backup field — no `isBackedUp`, nothing equivalent". Android has had one since #1239, in preferences at `vault_backup/<vaultId>`: set by all three export paths and by `.vult` import, cleared on reshare, and read by the home-screen warning banner. So the reminder banner the issue asks for already exists, and the flag was already there to read.

What was broken is the part that matters. Three things:

**The default was `true`.** `readBackupStatus` answered `true` for any vault with no stored flag. iOS declares `var isBackedUp: Bool = false` (`Vault.swift:27`) and Windows starts reshared vaults at false.

**Two save paths never wrote a flag,** so they inherited that `true`:

- fast-vault creation hands the vault to `temporaryVaultRepository` and defers the save to OTP verification, which wrote nothing — **a freshly created fast vault read as backed up and never showed the warning banner**
- the `SingleKeygen` MLDSA merge replaces `keyshares` and left the flag standing

**Preferences was the wrong home.** Not on the row means no cascade on delete (orphaned keys), and nothing can read the flag alongside the vault.

That inverted default is what blocks vultisig-sdk#1845: a gate asking "has this vault been exported" can't be built on a flag that answers yes for a vault that hasn't been.

## What this does

- `Vault.isBackedUp` + `VaultEntity.isBackedUp`, DB 41 → 42 (`ALTER TABLE vault ADD COLUMN isBackedUp INTEGER NOT NULL DEFAULT 0`)
- `VaultRepository.setBackupStatus` writes that one column rather than rewriting the row — the export paths hold a vault they read before writing the file, and an upsert from it would echo back whatever changed since, including keyshares a lock dropped mid-export
- the preferences flag and its `true` default are deleted outright, so the inverted default has no way back
- `VaultBackupStatusBackfill` carries the old flags across: SQL cannot read DataStore, so without a code-side pass everyone already holding a `.vult` would be told to back up again. Once per install under a mutex; row writes ahead of the edit that drops the legacy keys and records completion, so a run that dies midway stays re-derivable from the keys still in place. A vault with no stored flag stays `false` — unknown is not the same as safe. Swept by key prefix, which also clears the flags orphaned by deleted vaults
- the repository runs that pass **before** every vault read and before `upsert`/`setBackupStatus`. A read that ran first would return a vault flagged not-backed-up on the strength of a column the pass hadn't filled in; a backfill landing behind a reshare's write would put the old `true` back over it
- both MLDSA merges (`KeygenViewModel`, `FastVaultVerificationViewModel`) clear the flag. A `.vult` from before an MLDSA ceremony holds no MLDSA share, so restoring it yields a vault that cannot sign on QBTC — stale the same way a reshared vault's backup is
- keygen, reshare and migrate need nothing new: they upsert a freshly built vault, which carries the cleared default. That is also what fixes fast vaults
- the home banner reads `!vault.isBackedUp` off the vault it already loaded, one read fewer

## Two flag writes lost their `withContext(Dispatchers.IO)`

Room's suspend DAO calls already dispatch off the main thread, so the hop bought nothing — and it was leaking a coroutine past `resetMain` in the ViewModel tests. The `:app` suite failed roughly two runs in three before it went; it is green three consecutive runs now.

## Test plan

- `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest` — **4,702 pass, 0 fail** (2,748 `:data`, 1,954 `:app`), full `:app` suite run three times for the flakiness above
- `./gradlew :app:lintDebug :data:lintDebug` — clean; ktfmt clean
- New coverage:
  - `VaultBackupStatusBackfillTest` (12) — a legacy `true` is carried over; no stored flag and an explicit `false` both stay not-backed-up; legacy keys dropped and completion recorded; an orphan from a deleted vault dropped; unrelated preferences untouched; a later run in a new process does nothing; repeat and concurrent callers run the pass once; a failed run neither propagates nor gives up
  - `Migration41To42Test` — pins `DEFAULT 0`; a default of 1 would hand every row a backup it may not have
  - `VaultRepositoryImplTest` (+5) — flag round-trips both ways, a row with no flag reads false, `setBackupStatus` delegates, and the backfill is ordered ahead of the read
  - `VaultAccountsViewModelTest` (+2) — banner shows for a never-exported vault, hides for an exported one
  - `FastVaultVerificationViewModelTest` (+1) — the MLDSA merge clears the flag on the vault it writes back

Manual: fresh fast vault → home shows the backup warning (it did not before); export → warning clears; reshare → warning returns.

## Notes

- `#5565` (partial-keyshare export) is #5569, approved. This does not depend on it to land, but the "set only after a complete export" guarantee is only true once it does — the flag is set on export success, and #5569 is what makes that success mean the file is complete.
- Untouched: the monthly global backup reminder (`IsGlobalBackupReminderRequiredUseCase`), which is a separate, time-based nag.

## Summary by CodeRabbit

* **New Features**
  * Vault backup status is now stored with each vault.
  * Existing backup statuses are migrated automatically.
  * Backup status updates are preserved when vaults are saved or modified.
  * Backup warnings now reflect whether a vault has been backed up.

* **Bug Fixes**
  * Vault changes and key-share merges now correctly clear outdated backup status.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keygen/BackupVaultViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keygen/BackupVaultViewModel.kt
@@ -11,7 +11,6 @@ import com.vultisig.wallet.data.mappers.exportableOrNull
 import com.vultisig.wallet.data.models.TssAction
 import com.vultisig.wallet.data.models.Vault
 import com.vultisig.wallet.data.models.getVaultPart
-import com.vultisig.wallet.data.repositories.VaultDataStoreRepository
 import com.vultisig.wallet.data.repositories.VaultRepository
 import com.vultisig.wallet.data.usecases.CreateVaultBackupUseCase
 import com.vultisig.wallet.data.usecases.backup.CreateVaultBackupFileNameUseCase
@@ -52,7 +51,6 @@ constructor(
     private val isFileExtensionValid: IsVaultBackupFileExtensionValidUseCase,
     private val createVaultBackup: CreateVaultBackupUseCase,
     private val mapVaultToProto: MapVaultToProto,
-    private val vaultDataStoreRepository: VaultDataStoreRepository,
     private val snackbarFlow: SnackbarFlow,
     private val saveBackupToUri: SaveBackupToUriUseCase,
     private val deleteBackupDocument: DeleteBackupDocumentUseCase,
@@ -202,9 +200,7 @@ constructor(
             val vaultId = args.vaultId
 
             if (backupSuccess) {
-                withContext(Dispatchers.IO) {
-                    vaultDataStoreRepository.setBackupStatus(args.vaultId, true)
-                }
+                vaultRepository.setBackupStatus(args.vaultId, true)
 
                 snackbarFlow.showMessage(
                     UiText.StringResource(R.string.vault_settings_success_backup_message)
--- a/data/src/main/kotlin/com/vultisig/wallet/data/db/migrations/VaultBackupStatusBackfill.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/db/migrations/VaultBackupStatusBackfill.kt
@@ -1,0 +1,100 @@
+package com.vultisig.wallet.data.db.migrations
+
+import androidx.datastore.preferences.core.booleanPreferencesKey
+import com.vultisig.wallet.data.db.dao.VaultDao
+import com.vultisig.wallet.data.models.VaultId
+import com.vultisig.wallet.data.sources.AppDataStore
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import timber.log.Timber
+
+/**
+ * Carries the per-vault backup flag from DataStore onto the vault row, once per install.
+ *
+ * The flag used to live at `vault_backup/<vaultId>` in preferences and was read with a default of
+ * `true`, so a vault nobody had written a flag for counted as backed up. Two save paths never wrote
+ * one — fast-vault creation and the MLDSA merge — which made "never exported" indistinguishable
+ * from "exported" for precisely the vaults that had never been exported.
+ *
+ * [MIGRATION_41_42] replaces it with a column defaulting to `false`. Restoring the flags of users
+ * who *have* exported is why this class exists: SQL cannot read DataStore, so without a code-side
+ * pass everyone who already holds a `.vult` would be told to back up again. A vault with no stored
+ * flag stays `false` — unknown is not the same as safe.
+ *
+ * Called from [com.vultisig.wallet.data.repositories.VaultRepository] ahead of every read of a
+ * vault and every write that could carry the flag, so no caller can observe the pre-backfill state
+ * and no reshare can be undone by a backfill that lands after it. Safe to call from anywhere: the
+ * work happens under a [Mutex] at most once, and concurrent callers await that same run rather than
+ * repeating it. After it completes, calls cost a volatile read.
+ *
+ * A failure leaves the flags where they are and lets the next caller retry, rather than propagating
+ * — a preferences read that goes wrong must not take every vault read down with it. The order of
+ * writes is what makes the retry safe: rows are written first, and the legacy keys are dropped in
+ * the same edit that records completion, so a run that dies midway is always re-derivable from the
+ * keys still in place.
+ */
+@Singleton
+internal class VaultBackupStatusBackfill
+@Inject
+constructor(private val vaultDao: VaultDao, private val appDataStore: AppDataStore) {
+
+    private val mutex = Mutex()
+
+    @Volatile private var isCompleted = false
+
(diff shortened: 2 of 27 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keygen/BackupVaultViewModel.kt
```diff
@@ -202,9 +200,7 @@ constructor(
             val vaultId = args.vaultId
 
             if (backupSuccess) {
-                withContext(Dispatchers.IO) {
-                    vaultDataStoreRepository.setBackupStatus(args.vaultId, true)
-                }
+                vaultRepository.setBackupStatus(args.vaultId, true)
```
rkokhatskyi, line 203: `vaultRepository.setBackupStatus` runs inside a bare `viewModelScope.launch` with no CoroutineExceptionHandler anywhere in the app, so a Room UPDATE failure here crashes right after a successful backup instead of showing an error. `ImportFileViewModel.kt:216` already wraps the identical call in `runBestEffort`, this and two sibling ViewModels never do

data/src/main/kotlin/com/vultisig/wallet/data/db/migrations/VaultBackupStatusBackfill.kt
```diff
@@ -0,0 +1,100 @@
...
+            }
+            isCompleted = true
+        }
+    }
+
+    private suspend fun backfill() {
+        val vaultIds = vaultDao.loadAllIds()
+        val backedUp =
+            vaultIds.filter { appDataStore.readData(legacyBackupStatusKey(it), false).first() }
+        backedUp.forEach { vaultDao.setBackupStatus(vaultId = it, isBackedUp = true) }
```
rkokhatskyi, line 73: `backfill()` copies a legacy `vault_backup/<id>=true` straight onto `isBackedUp` with no check that keyshares changed since. Every app version before this PR's MLDSA merge fix (`KeygenViewModel.kt`, `FastVaultVerificationViewModel.kt`) never cleared that legacy flag, so a vault exported before adding quantum resistance backfills to true despite holding an unexported MLDSA share

data/src/main/kotlin/com/vultisig/wallet/data/models/Vault.kt
```diff
@@ -22,6 +22,21 @@ data class Vault(
     var coins: List<Coin> = emptyList(),
     var chainPublicKeys: List<ChainPublicKey> = emptyList(),
     var libType: SigningLibType = SigningLibType.GG20,
+    /**
+     * True once this vault has been exported to a `.vult` that is known to be complete.
+     *
+     * Defaults to false, so a vault nobody has exported is never mistaken for one that is safe to
+     * lose — matching iOS and Windows. Set it only through
+     * [com.vultisig.wallet.data.repositories.VaultRepository.setBackupStatus], and only after an
```
rkokhatskyi, line 30: This KDoc says `isBackedUp` is set only through `VaultRepository.setBackupStatus`, but its own next paragraph describes a path that clears it directly instead. `KeygenViewModel.kt:893` and `FastVaultVerificationViewModel.kt:149` both assign `existingVault.isBackedUp = false` then upsert, never calling `setBackupStatus`

rkokhatskyi (changes requested):

johnnyluo:
I am going to close this one , android just have a different implementation , which is the implementation detail , there is no need to add these , user doesn't get any more benefit from it.

PR #5580, opened 2026-08-11
feat(swap): Select route picker in the Advanced swap sheet

aminsato:
## What

Adds the missing **Select route** feature to the Advanced swap sheet, porting the route picker from the iOS Advanced Swap work (vultisig-ios#4570).

The Advanced swap sheet gains a **Select route** row (between Gas Limit and Use External Recipient) that opens a page listing **every fetched provider quote**, ranked by net output amount with the active route pinned to the top. Each row shows the provider logo and name, a "Fee $X · ~Ns" subtitle, the approximate destination output with its fiat value, and a green check on the active route. Tapping a route applies it immediately and returns to the menu.

### Transaction hash
https://etherscan.io/tx/0xfc087edcefc819690f0c321f271cf04715193b972050e3b767feb42891701a80

## Behavior

- **Ranking**: routes are ordered best→worst by the same net-output metric (`comparableDstFiat`) the auto-selection ranks on, so the top row always shows the largest output. The auto winner can differ from the top row when the banded provider preference applies — the list order deliberately stays output-ranked.
- **Override semantics (iOS parity)**: a manual pick overrides the automatic best until the next quote refresh (expiry / amount / pair / slippage / recipient change), which resets the route to Auto. The menu row reads "Auto" until a manual pick is made, then the provider name.
- **No extra fetch**: the per-provider fan-out already fetched every candidate and discarded the losers. Picking a route rebuilds display state for the already-fetched candidate and re-applies it through the same `applyQuoteResult` path — fees, discounts, UTXO plan-fee verification, balance validation, and the refresh timer all follow the pick. Picks that land after the input changed are dropped by the existing superseded guards.
- **Row details**: the ETA segment only renders for THORChain/Maya (the only providers exposing `total_swap_seconds`); the fee segment is dropped for 1inch, whose affiliate fee is baked into the quoted rate. The row is disabled while fewer than two routes exist.
- The row lives inside the Advanced sheet, so it inherits the existing Silver-tier gate.

## Changes

- `SwapQuoteManager.fetchBestQuote` returns `RankedQuotes(best, ranked)`; `QuoteResolution.Success` / `SwapQuotePipelineResult.Success` carry the ranked candidate set through `buildSuccess`.
- `SwapQuotePipelineController`: keeps a `RouteContext(input, ranked)`, builds `SwapRouteUiModel` rows, and exposes `selectRoute(provider)`; route state clears on every reset/supersede path.
- `SwapFormUiModel` gains `routeOptions` + `isRouteManuallySelected`; `SwapFormViewModel.selectRoute` delegates to the controller.
- `AdvancedSwapSettingsSheet`: new `SelectRoute` page + menu row; new `ic_route.xml` vector (ported from the iOS `route.svg` asset).
- Strings added to English and all 9 locale files, matching each locale's established route/fee terminology.
- Preview: `AdvancedSelectRoutePreview` + PreviewActivity `swap_advanced_select_route` entry.

## UI (verified on emulator, described in text)

- **Before**: Advanced swap menu shows Slippage Tolerance / Gas Limit / Use External Recipient only.
- **After**: a fourth "Select route — Auto ›" row with the route icon appears between Gas Limit and Use External Recipient; the Select route page renders the ranked list (e.g. THORChain pinned first with green check, "Fee $0.00 · ~606s", "~21.83561311 RUNE / $9.47"; Maya protocol below with an empty radio), matching the reference design.

## Tests

- New: `fetchBestQuote ranks the full candidate set by net output regardless of the winner` — pins that the picker list stays output-ordered even when the banded preference selects a lower-output provider as best.
- Updated `SwapQuoteManagerTest` / `SwapFormViewModelTest` for the `RankedQuotes` return shape.
- `:app:testDebugUnitTest` swap package: 332 tests green. `:app:lintDebug` + `:data:lintDebug` green. ktfmt clean.

## New Features
- Added advanced swap route selection in Advanced Swap Settings.
- Compare providers by output amount, fees, and estimated completion time.
- Select a preferred route without refetching quotes.
- Routes are ranked by expected destination value with provider branding and selection states.
- Added localized route-selection text across supported languages.

## Tests
- Expanded coverage for route ranking, provider selection, output formatting, and network-fee handling.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -511,39 +511,60 @@ constructor(
             else selected
         }
 
+        val metric = rankingMetric(successes)
         return RankedQuotes(
             best = selectBestQuote(successes),
+            // Ordered by the same metric the winner is chosen with, so the picker can never rank a
+            // route above the one it puts the check on. The raw amount breaks a tie between two
+            // routes whose fiat rounds to the same figure.
             ranked =
                 successes.sortedWith(
-                    // Fiat first (matches the best-quote metric), then the raw destination amount —
-                    // every candidate quotes the same dst token, so amounts compare directly. The
-                    // tie-break keeps the list output-ordered when the dst token has no price and
-                    // comparableDstFiat collapses to zero for every candidate at once.
-                    compareByDescending<BestQuote> { it.result.comparableDstFiat }
-                        .thenByDescending { it.result.quote.expectedDstValue.decimal }
+                    compareByDescending<BestQuote>(metric).thenByDescending {
+                        it.result.quote.expectedDstValue.decimal
+                    }
                 ),
         )
     }
 
     /**
-     * Picks the winning quote across providers. The ranking metric is net destination output
-     * ([QuoteFetchResult.comparableDstFiat]) — every provider in a candidate set swaps to the same
-     * destination token, so the values are directly comparable. Among quotes within
-     * [PROVIDER_PREFERENCE_BAND] of the best net output (economically tied on rate) a banded
-     * preference picks the winner by the iOS canonical rule: source gas decides only when both
-     * quotes expose it (same-chain EVM aggregators), then provider priority, then higher net
-     * output. Only the aggregators expose gas and their priorities form one contiguous block, so no
-     * gas-unknown provider's priority falls between two gas-exposing ones — the comparator is a
-     * strict weak order for every realizable candidate set and the winner is independent of input
-     * order. Anything outside the band loses on output.
+     * What a candidate set is ranked by: net destination fiat
+     * ([QuoteFetchResult.comparableDstFiat]) whenever any candidate has one, else the raw
+     * destination amount.
+     *
+     * Every candidate quotes the same destination token, so the amount is directly comparable and
+     * says exactly what the fiat would have said. Without the fallback an unpriced destination
+     * collapses every candidate's metric to zero at once, which reads as "every route is
+     * economically tied" and hands the whole decision to provider priority — so the winner could be
+     * a route another one beats outright on output.
+     */
+    private fun rankingMetric(successes: List<BestQuote>): (BestQuote) -> BigDecimal =
+        if (successes.any { it.result.comparableDstFiat.signum() > 0 }) {
+            { it.result.comparableDstFiat }
+        } else {
+            { it.result.quote.expectedDstValue.decimal }
+        }
+
+    /**
+     * Picks the winning quote across providers. The ranking metric is [rankingMetric] — net
+     * destination output, in fiat where the destination has a price and in destination tokens where
+     * it doesn't. Every provider in a candidate set swaps to the same destination token, so the
+     * values are directly comparable either way. Among quotes within [PROVIDER_PREFERENCE_BAND] of
+     * the best net output (economically tied on rate) a banded preference picks the winner by the
+     * iOS canonical rule: source gas decides only when both quotes expose it (same-chain EVM
+     * aggregators), then provider priority, then higher net output. Only the aggregators expose gas
+     * and their priorities form one contiguous block, so no gas-unknown provider's priority falls
+     * between two gas-exposing ones — the comparator is a strict weak order for every realizable
+     * candidate set and the winner is independent of input order. Anything outside the band loses
+     * on output.
      *
      * Assumes [successes] is non-empty (the caller has already surfaced the all-failed case).
      */
     internal fun selectBestQuote(successes: List<BestQuote>): BestQuote {
-        val best = successes.maxBy { it.result.comparableDstFiat }
-        val floor = best.result.comparableDstFiat * (BigDecimal.ONE - PROVIDER_PREFERENCE_BAND)
+        val metric = rankingMetric(successes)
+        val best = successes.maxBy(metric)
+        val floor = metric(best) * (BigDecimal.ONE - PROVIDER_PREFERENCE_BAND)
         return successes
-            .filter { it.result.comparableDstFiat >= floor }
+            .filter { metric(it) >= floor }
             .minWithOrNull(
                 // Source gas decides only when both quotes expose it; otherwise this ties and
                 // selection falls through to provider priority, then higher net output.
@@ -553,7 +574,7 @@ constructor(
                         if (lhsGas != null && rhsGas != null) lhsGas.compareTo(rhsGas) else 0
                     }
                     .thenBy { providerPriority(it.candidate.provider) }
-                    .thenByDescending { it.result.comparableDstFiat }
+                    .thenByDescending(metric)
             ) ?: best
     }
 
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -513,7 +513,15 @@ constructor(
 
         return RankedQuotes(
             best = selectBestQuote(successes),
-            ranked = successes.sortedByDescending { it.result.comparableDstFiat },
+            ranked =
+                successes.sortedWith(
+                    // Fiat first (matches the best-quote metric), then the raw destination amount —
+                    // every candidate quotes the same dst token, so amounts compare directly. The
+                    // tie-break keeps the list output-ordered when the dst token has no price and
+                    // comparableDstFiat collapses to zero for every candidate at once.
+                    compareByDescending<BestQuote> { it.result.comparableDstFiat }
+                        .thenByDescending { it.result.quote.expectedDstValue.decimal }
+                ),
         )
     }
 
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -105,13 +105,21 @@ internal data class QuoteCandidate(
 internal data class BestQuote(val candidate: QuoteCandidate, val result: QuoteFetchResult)
 
 /**
+ * Every successfully fetched quote for one request: the banded-preference winner ([best]) plus the
+ * full candidate set ordered best→worst by net destination output ([ranked]). [ranked] always
+ * contains [best]; it drives the Select-route picker, where the user can override the automatic
+ * winner with any other fetched route.
+ */
+internal data class RankedQuotes(val best: BestQuote, val ranked: List<BestQuote>)
+
+/**
  * Outcome of resolving the best quote for a pair: either a [Success] holding the winning
  * [BestQuote], or a [Failure] whose typed swap error has already been mapped to a renderable
  * [Failure.formError] so the ViewModel only has to surface it.
  */
 internal sealed interface QuoteResolution {
-    /** A winning quote was fetched. */
-    data class Success(val best: BestQuote) : QuoteResolution
+    /** A winning quote was fetched; [ranked] is the full output-ordered candidate set. */
+    data class Success(val best: BestQuote, val ranked: List<BestQuote>) : QuoteResolution
 
     /** The fetch failed; [formError] is the mapped message, [cause]/[tag] are for logging. */
     data class Failure(val formError: UiText, val cause: Throwable, val tag: String) :
@@ -432,7 +440,7 @@ constructor(
         amount: BigDecimal,
         slippageBps: Int? = null,
         externalRecipient: String? = null,
-    ): BestQuote {
+    ): RankedQuotes {
         if (candidates.isEmpty()) {
             throw SwapException.SwapIsNotSupported("Swap is not supported for this pair")
         }
@@ -503,7 +511,10 @@ constructor(
             else selected
         }
 
-        return selectBestQuote(successes)
+        return RankedQuotes(
+            best = selectBestQuote(successes),
+            ranked = successes.sortedByDescending { it.result.comparableDstFiat },
+        )
     }
 
     /**
@@ -561,7 +572,7 @@ constructor(
         externalRecipient: String? = null,
     ): QuoteResolution =
         try {
-            QuoteResolution.Success(
+            val fetched =
                 fetchBestQuote(
                     candidates = candidates,
                     src = src,
@@ -575,7 +586,7 @@ constructor(
                     slippageBps = slippageBps,
                     externalRecipient = externalRecipient,
                 )
-            )
+            QuoteResolution.Success(best = fetched.best, ranked = fetched.ranked)
         } catch (e: SwapException) {
             QuoteResolution.Failure(
                 formError = mapSwapExceptionToFormError(e, srcToken, selectedSrcTokenTitle),
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
@@ -27,6 +27,7 @@ import dagger.assisted.Assisted
 import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
 import java.math.BigDecimal
+import java.math.RoundingMode
 import java.text.DecimalFormat
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineDispatcher
@@ -154,12 +155,14 @@ constructor(
     // reset cancels this first, keeping exactly one writer of quote state at a time.
     private var selectRouteJob: Job? = null
 
-    private data class RouteContext(val input: QuoteInput, val ranked: List<BestQuote>)
+    // Ticket taken by every writer of quote state — each pipeline apply, each manual pick, each
+    // reset. Cancelling the pick job only protects one direction: a pipeline apply that suspended
+    // in resolveNetworkFee resumes with no idea a pick has landed since, and would stamp its own
+    // network fee and isSwapDisabled over the quote now on screen. Whoever wrote the quote last
+    // owns its fee, so a writer resuming on a stale ticket drops its outcome.
+    private var quoteApplyGeneration = 0
 
-    // Route-row output formatter: full precision with grouping, unlike the display mapper, which
-    // abbreviates above 1e6 and would collapse the differences the picker column compares.
-    // Main-confined like the rest of this class (DecimalFormat is not thread-safe).
-    private val routeOutputFormat = DecimalFormat("#,##0.########")
+    private data class RouteContext(val input: QuoteInput, val ranked: List<BestQuote>)
 
     // Suppresses the quote-refresh timer while the form isn't the foreground screen. The form's
     // scope (= viewModelScope) stays alive on the back stack once the flow proceeds to
@@ -498,6 +501,7 @@ constructor(
 
         val (src, dst) = input.address
 
+        val generation = ++quoteApplyGeneration
         routeContext = RouteContext(input, result.rankedQuotes)
         val routeOptions =
             buildRouteOptions(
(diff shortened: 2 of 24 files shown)


rkokhatskyi:
@aminsato please attach screenshots before / after

aminsato:
> @aminsato please attach screenshots before / after

https://etherscan.io/tx/0xfc087edcefc819690f0c321f271cf04715193b972050e3b767feb42891701a80

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -495,6 +516,82 @@ constructor(
...
+                    bestQuote = candidate,
+                    src = src,
+                    srcTokenValue = srcTokenValue,
+                    tokenValue = tokenValue,
+                    currentDiscountInfo = uiState.value.discountInfo,
+                    rankedQuotes = ctx.ranked,
+                )
+            // applyQuoteResult re-runs the superseded/quotable guards against the live input, so a
+            // pick that lands after the user changed amount/pair/settings is dropped, not applied.
+            applyQuoteResult(ctx.input, result, isManualRoutePick = true)
```
rkokhatskyi: Picking THORChain or Maya on a native EVM source keeps the aggregator's re-based gas because `resolveNetworkFee` writes a fee only in the UTXO and `SwapQuote.OneInch` branches (`SwapQuotePipeline.kt:485-528`), so the same route shows a different Network Fee and Total than when it wins a fresh fetch

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -495,6 +516,82 @@ constructor(
...
+     * picked, already-fetched candidate and applies it through the same path as an auto-resolved
+     * quote, so fees, discounts, balance validation, and the refresh timer all follow the pick. No
+     * network fetch — the pick only chooses among the quotes of the last resolution, and the next
+     * refresh re-defaults the route to the automatic best (iOS parity).
+     */
+    fun selectRoute(provider: SwapProvider) {
+        val ctx = routeContext ?: return
+        if (quoteState.provider == provider) return
+        val candidate = ctx.ranked.firstOrNull { it.candidate.provider == provider } ?: return
+        scope.safeLaunch(onError = { Timber.e(it, "selectRoute") }) {
```
rkokhatskyi: `applyQuoteResult` was serialized by the `collectLatest` at line 374 and this second caller is never cancelled, so a pick suspended in `resolveNetworkFee` can write its own network fee and `isSwapDisabled` over the quote an expiry refresh applied meanwhile

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -495,6 +516,82 @@ constructor(
...
+            val result =
+                swapQuotePipeline.buildSuccess(
+                    bestQuote = candidate,
+                    src = src,
+                    srcTokenValue = srcTokenValue,
+                    tokenValue = tokenValue,
+                    currentDiscountInfo = uiState.value.discountInfo,
+                    rankedQuotes = ctx.ranked,
+                )
+            // applyQuoteResult re-runs the superseded/quotable guards against the live input, so a
```
rkokhatskyi: `ctx.input` is captured before the launch, so when an input changed meanwhile the late pick reaches `discardSupersededQuoteResult` and clears the newer quote that had already landed, leaving the form with no quote, no error, no spinner and no armed refresh

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -495,6 +516,82 @@ constructor(
...
+     * Applies a manual route pick from the Select-route sheet: rebuilds the display state for the
+     * picked, already-fetched candidate and applies it through the same path as an auto-resolved
+     * quote, so fees, discounts, balance validation, and the refresh timer all follow the pick. No
+     * network fetch — the pick only chooses among the quotes of the last resolution, and the next
+     * refresh re-defaults the route to the automatic best (iOS parity).
+     */
+    fun selectRoute(provider: SwapProvider) {
+        val ctx = routeContext ?: return
+        if (quoteState.provider == provider) return
+        val candidate = ctx.ranked.firstOrNull { it.candidate.provider == provider } ?: return
```
rkokhatskyi, line 641: The candidate is resolved by provider alone and neither `buildRouteOptions` nor the supersede guard reads `expiredAt`, which each source stamps at its own parse time, so a row that lapsed while the sheet was open applies with Swap left enabled

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -495,6 +516,82 @@ constructor(
...
+                    }
+                SwapRouteUiModel(
+                    provider = candidate.candidate.provider,
+                    name = fetch.providerUiText,
+                    logo = getProviderLogo(candidate.candidate.provider.getSwapProviderId()),
+                    // 1inch bakes the affiliate fee into the quoted rate — there is no separate
+                    // amount to show, so the fee segment is dropped rather than showing "$0.00".
+                    feeText =
+                        if (fetch.swapFeeIncludedInRate) null
+                        else fiatValueToString(fetch.swapFeeFiat, asFee = true),
```
rkokhatskyi, line 613: Row fee reads the raw `fetch.swapFeeFiat`, bypassing the SwapKit UTXO zeroing in `buildSuccess` (`SwapQuotePipeline.kt:401-410`), so a BTC SwapKit row advertises a swap fee that vanishes from the breakdown the moment that route is picked

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -495,6 +516,82 @@ constructor(
...
+                    etaText =
+                        etaSeconds?.let {
+                            UiText.FormattedText(R.string.swap_route_eta, listOf(it))
+                        },
+                    outputText = "~${fetch.estimatedDstTokenValue} $dstTicker",
```
rkokhatskyi: `estimatedDstTokenValue` abbreviates above 1e6 to one decimal, so 1,051,000 and 1,149,999 both render `~1.1M` in the column the picker exists to compare, while iOS keeps two decimals (`DecimalExtension.swift:106`)

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -503,7 +511,10 @@ constructor(
             else selected
         }
 
-        return selectBestQuote(successes)
+        return RankedQuotes(
+            best = selectBestQuote(successes),
+            ranked = successes.sortedByDescending { it.result.comparableDstFiat },
```
rkokhatskyi: Ranking on `comparableDstFiat` collapses when the destination token has no price, since `TokenPriceRepository.getPrice` falls back to zero for every candidate at once, leaving the rows in provider-table order under a helper that promises they are ranked by output

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in dd050327b. calculateGas now stores its display estimate as the EVM baseline, and resolveNetworkFee gained a third branch: a non-aggregator quote on an EVM source restores that stored estimate verbatim, so the fee is byte-identical to what a fresh fetch with that winner shows (restoring the exact stored value rather than recomputing avoids introducing the reverse mismatch). Covered by the new `restores the gas-pass estimate for a THOR quote on an EVM source` test.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in dd050327b with a `selectRouteJob` handle: every pipeline-owned apply and both reset paths cancel an in-flight pick before writing, so a pick suspended in resolveNetworkFee is cancelled instead of resuming over the refreshed quote. The pick also re-checks `routeContext` identity after the buildSuccess suspension and drops itself when a newer apply replaced the context.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in dd050327b. The manual path now returns silently from both guards — `!isLiveInputQuotable` and the superseded check — instead of running resetQuoteState/discardSupersededQuoteResult, which belong to pipeline-owned results where the on-screen state IS the stale state. A stale pick leaves the newer request's quote, spinner, and refresh timer untouched.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in dd050327b. selectRoute now refuses a candidate whose quote's expiredAt has lapsed and bumps refreshQuoteState instead, so the whole list is replaced by a fresh candidate set rather than applying a quote that can no longer be signed at its rate.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in dd050327b. buildRouteOptions now mirrors buildSuccess's fee treatment: a SwapKit quote on a UTXO-family source hides the fee segment (its deposit cost is already the Network Fee), alongside the existing 1inch included-in-rate case — so a row never advertises a fee that vanishes when picked.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in dd050327b. The output column is now formatted from the raw quote's expectedDstValue with a full-precision grouping formatter (up to 8 fraction digits) instead of the abbreviating display mapper, so 1,051,000 and 1,149,999 render distinctly.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Fixed in dd050327b. The ranked list now tie-breaks on the raw destination amount — every candidate quotes the same dst token, so amounts compare directly — which keeps it output-ordered when comparableDstFiat collapses to zero for all candidates. Pinned by the new `ranks by token amount when the dst token has no price` test.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -129,10 +134,33 @@ constructor(
...
+    // applies, but a pick runs on an independent coroutine — so every pipeline-owned apply and
+    // reset cancels this first, keeping exactly one writer of quote state at a time.
+    private var selectRouteJob: Job? = null
+
+    private data class RouteContext(val input: QuoteInput, val ranked: List<BestQuote>)
+
+    // Route-row output formatter: full precision with grouping, unlike the display mapper, which
+    // abbreviates above 1e6 and would collapse the differences the picker column compares.
+    // Main-confined like the rest of this class (DecimalFormat is not thread-safe).
+    private val routeOutputFormat = DecimalFormat("#,##0.########")
```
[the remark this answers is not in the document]
aminsato: Not changing this: 8 fraction digits is the deliberate cap, matching iOS exactly — the same picker column there goes through `formatDecimalToLocale`, which sets `maximumFractionDigits = 8` (DecimalExtension.swift:82) for every sub-1M value. Deriving the digit count from `value.scale()` would print up to 18 dust digits for EVM tokens and diverge from the reference implementation, while two quotes that agree to 1e-8 of the same token are economically indistinguishable (their fiat lines render identically too). Above 1M this format is already strictly more precise than iOS, which abbreviates to 2 decimals — that gap was the substance of the earlier review point.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -489,12 +552,115 @@ constructor(
...
+                        else fiatValueToString(fetch.swapFeeFiat, asFee = true),
+                    etaText =
+                        etaSeconds?.let {
+                            UiText.FormattedText(R.string.swap_route_eta, listOf(it))
+                        },
+                    // Formatted from the raw quote amount, NOT estimatedDstTokenValue: the display
+                    // formatter abbreviates above 1e6 ("~1.1M"), which collapses the differences
+                    // this column exists to compare (iOS keeps full precision here).
+                    outputText =
+                        "~${routeOutputFormat.format(fetch.quote.expectedDstValue.decimal)} " +
```
rkokhatskyi: Dropping the abbreviation makes `outputText` unbounded while the output Column has no weight and the name Column has `weight(1f)`, so `~375,000,000.12345678 PEPE` measures 198dp of the ~200dp two-column budget at 360dp and clips the provider name to nothing. iOS abbreviates here too (`DecimalExtension.swift:136`), so two decimals would fix both

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -424,7 +459,18 @@ constructor(
         // be applied — and then signed with a mismatched memo / slippage floor — over the new one
         // (#5310). Match the source/destination quote endpoints as well: changing to another
         // routable pair with the same amount creates the same pre-debounce landing window.
+        // A pipeline-owned result supersedes any pick still applying: cancel it so a pick
+        // suspended in resolveNetworkFee can't write its network fee / isSwapDisabled over the
+        // quote this fresh result is about to apply.
+        if (!isManualRoutePick) {
```
rkokhatskyi, line 468: Cancellation only runs in this direction, and a pipeline apply already suspended in `resolveNetworkFee` has no generation check on resume, so it stamps its own network fee and `isSwapDisabled` over a pick that completed meanwhile

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -503,7 +511,18 @@ constructor(
...
+        return RankedQuotes(
+            best = selectBestQuote(successes),
+            ranked =
+                successes.sortedWith(
+                    // Fiat first (matches the best-quote metric), then the raw destination amount —
+                    // every candidate quotes the same dst token, so amounts compare directly. The
+                    // tie-break keeps the list output-ordered when the dst token has no price and
+                    // comparableDstFiat collapses to zero for every candidate at once.
+                    compareByDescending<BestQuote> { it.result.comparableDstFiat }
+                        .thenByDescending { it.result.quote.expectedDstValue.decimal }
```
rkokhatskyi: The amount tie-break lands on `ranked` but not on `selectBestQuote`, which still resolves an all-zero-fiat set by provider priority, so for an unpriced destination the picker pins a green check on a route the row below it beats on output

rkokhatskyi (dismissed):

johnnyluo (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in 5d2fb31. You are right that this swung too far the other way — full precision made the column unbounded, and it is the one with no weight while the name has `weight(1f)`, so it wins the measure and the name gets what's left.

Now formatted the way iOS formats its own (`formatForDisplay`): full precision with grouping below 1e6, two decimals and a K/M/B/T suffix from there up, truncating rather than rounding so a row can't advertise more output than the quote pays. `~375,000,000.12345678 PEPE` becomes `~375M PEPE`; the sub-million case that motivated dropping the mapper keeps all eight decimals, so two routes differing in the last digit still read as different rows.

New `RouteOutputFormatTest` pins both ends: eight-decimal precision below the threshold, abbreviation and truncation above it.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in 5d2fb31. Cancelling the pick job only ever protected the pick→pipeline direction; the other way round there was nothing to cancel, since the pipeline apply is the `collectLatest` coroutine itself, and on resume from `resolveNetworkFee` it had no way to know a pick had landed.

Added a generation ticket taken by every writer of quote state — each pipeline apply, each manual pick — right before it writes the quote. After `resolveNetworkFee` returns, a writer whose ticket is no longer current returns without applying the outcome, so the network fee, `isSwapDisabled` and the refresh timer all stay with whoever wrote the quote last. `resetQuoteState` and `discardSupersededQuoteResult` bump it too, so a fee resolving for a quote that has since been cleared can't land on empty state either.

The pick-side cancellation stays: it stops a pick from doing the work at all, which is still worth avoiding.

Not unit-tested — `resolveNetworkFee` runs on the `SwapQuotePipeline` the controller constructs internally, so there is no seam to suspend it at from a test. Same reason the pick-race fixes in dd05032 went in without one.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Fixed in 5d2fb31. The tie-break was the wrong shape for this — it patched the ordering while `selectBestQuote` kept ranking on fiat alone, so the two could disagree exactly where you said.

Both now go through one `rankingMetric(successes)`: net destination fiat when any candidate has one, the raw destination amount when none does (every candidate quotes the same dst token, so it says what the fiat would have). That also fixes the band, which was the real damage — with every metric at zero the floor is zero, so the whole set counts as economically tied and provider priority decides on its own. On the amount metric the band means what it means everywhere else, and a route that loses materially on output is out of it.

`ranked` keeps the amount as a secondary tie-break for two routes whose fiat rounds to the same figure.

New test asserts `best` is SwapKit's 200 rather than THORChain's 100 on the all-zero-fiat set, and that it equals `ranked.first()`. Note this does not make `best == ranked.first()` an invariant in general — a route inside the band can still win on priority with slightly less output, which is the intended behaviour and is pinned by the existing test above it.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -440,11 +489,27 @@ constructor(
...
+            // and no armed refresh.
+            if (isManualRoutePick) return
             discardSupersededQuoteResult()
             return
         }
 
-        val (src, _) = input.address
+        val (src, dst) = input.address
+
+        val generation = ++quoteApplyGeneration
```
rkokhatskyi, line 504: The ticket is taken before `buildRouteOptions`, which suspends on `fiatValueToString` reading `currency.first()` from DataStore, and the quote write after it is unguarded, so an overtaken writer still lands its display and only its fee is dropped. Re-check the generation after `buildRouteOptions` too

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Fixed in e2588266b: the generation is now re-checked immediately after buildRouteOptions, before the quote/display write, so a writer overtaken during that suspension lands nothing — same treatment as the resolveNetworkFee window.

rkokhatskyi (approved):
LGTM

PR #5581, opened 2026-08-11
fix(passcode): don't open bottom sheets over the lock screen

### LINKED ISSUE #5560: ModalBottomSheet composed after the lock is raised renders above it
**Rescoped 2026-08-10** after review — see the comment thread. The `FLAG_SECURE`-on-the-lock-window claim was withdrawn (Compose inherits it via `SecureFlagPolicy.Inherit`), and the recents-snapshot question was split out as a product decision. What remains is the sheet residual below.

## Problem

PR #5554 hosts the lock in its own full-screen `Dialog` and pops nav `dialog<>` destinations while locked. Android windows within a token stack in **creation order**, so the lock covers what was already open — but a window created *afterwards* still lands on top. `PopDialogDestinationsWhileLocked` keys on `navigatorName == "dialog"`, a proxy for "creates a window" that misses `ModalBottomSheet` composed inside an ordinary `composable<>` screen.

This is the known residual named in #5554's own description.

## Reachable set

Not every sheet call site — only those whose visibility is flipped by **async ViewModel work** that can land after the lock is up, with no user action:

- **`MonthlyBackupReminder`** — `showGlobalBackupReminder()` is a `safeLaunch` off vault load (`VaultAccountsViewModel.kt:320-324`), so its result can arrive after the lock. Confirmed reachable.
- **`ScanQrScreen`** — `composable<ScanQr>` at `NavGraph.kt:315` whose body is a `ModalBottomSheet`, so it creates a window the `navigatorName` test misses. Structurally valid, tap-routed in practice.

Not affected: `AddressBookBottomSheet` (declared `dialog<AddressBook>` at `NavGraph.kt:424`, already popped) and `TransactionDetailBottomSheet` (tap-routed; while locked the lock window holds focus and the in-window cover consumes the rest).

Accessibility has the same root cause rather than being a separate defect: a popup window's semantics subtree is outside the `clearAndSetSemantics {}` at `MainActivityContent.kt:130-134`, so TalkBack can walk a sheet raised while locked. Handling the composition handles the semantics with it.

## Severity

Visibility and input only — **not** a fund-loss path. The data key is dropped on lock, so a stale or late sheet can neither read a key share nor sign.

## Fix options

There is no generic "suppress `ModalBottomSheet` while locked" hook in Compose. Two shapes:

1. Stop composing `navContent` while the gate is closed — **rejected**: disposes the `NavHost`, tearing down screen state and in-flight work, which is what #5548 / PR #5553 exist to protect.
2. Route sheet call sites through a gate-aware wrapper (or a `CompositionLocal` they read) — enumeration, but behind one shared abstraction rather than in the popper's type test. **Recommended**, subject to the implementer's judgement.

## Acceptance criteria

- [ ] A sheet whose visibility flips from async work after the lock is raised does not render above it
- [ ] The same handling covers the semantics subtree (TalkBack cannot walk it)
- [ ] `MonthlyBackupReminder` covered by a regression test

Ports of this class: iOS #5082, Windows/extension #4596. Related: #5551, PR #5554.

rkokhatskyi:
Closes #5560.

Windows stack in the order they're added, so a bottom sheet that opens while the app is locked lands on top of the lock. #5554 handled the sheets that were already open and the dialog destinations; what's left is a sheet opened by background work after the lock is already up.

`OnceUnlocked` waits for the passcode gate before composing its content. One that was already open stays put — the lock covers it anyway, and closing it would throw away the screen underneath.

Applied to the three sheets on Home that show up on their own rather than on a tap: the monthly backup reminder and the two notification prompts. The rest either need a tap, which can't land while the lock has focus, or are dialog destinations #5554 already pops. iOS does the same thing in `presentsWhenUnlocked`.

## Before / After

| Before | After | After, once unlocked |
|--------|-------|----------------------|
| before | after | unlocked |

## Testing

- `:app:connectedDebugAndroidTest` for `OnceUnlockedTest` + `PopDialogDestinationsWhileLockedTest` — 10 tests, 0 failures on a Pixel 9 Pro API 35 emulator. The three hold cases fail without the gate.
- `:app:testDebugUnitTest`, `:app:lintDebug`, `:app:ktfmtCheck`

## Summary by CodeRabbit

* **New Features**
  * Added passcode-aware handling for backup reminders and notification prompts.
  * Reminders and prompts become available after the wallet is unlocked and remain available if the app is locked again.
  * Prevented gated overlays from remaining accessible while the passcode screen is closed.

* **Tests**
  * Added coverage for reminder visibility, reopening behavior, persistence, and pending reminder handling.

### DIFF
--- a/app/src/androidTest/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlockedTest.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlockedTest.kt
@@ -1,0 +1,146 @@
+package com.vultisig.wallet.ui.screens.passcode
+
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onAllNodesWithText
+import androidx.compose.ui.test.onNodeWithText
+import androidx.test.platform.app.InstrumentationRegistry
+import com.vultisig.wallet.R
+import com.vultisig.wallet.ui.screens.home.MonthlyBackupReminder
+import org.junit.Rule
+import org.junit.Test
+
+/**
+ * Driven through `MonthlyBackupReminder`, the sheet the home screen raises off a background read
+ * with no user action — so its window can be added after the passcode lock's and draw above it.
+ * Composed here the way the home screen composes it: the caller's own condition, then the gate.
+ */
+class OnceUnlockedTest {
+
+    @get:Rule val compose = createComposeRule()
+
+    private var isGateClosed by mutableStateOf(false)
+    private var isReminderWanted by mutableStateOf(false)
+
+    @Test
+    fun theReminderShowsWhileTheGateIsOpen() {
+        start(gateClosed = false)
+
+        wantReminder(true)
+
+        assertReminderShows()
+    }
+
+    @Test
+    fun theReminderIsHeldWhileTheGateIsClosed() {
+        start(gateClosed = true)
+
+        wantReminder(true)
+
+        assertReminderHeld()
+    }
+
+    @Test
+    fun theHeldReminderShowsOnceTheGateOpens() {
+        start(gateClosed = true)
+        wantReminder(true)
+        assertReminderHeld()
+
+        setGateState(closed = false)
+
+        assertReminderShows()
+    }
+
+    /**
+     * The lock's own window covers a sheet that was already up, and dropping it would discard the
+     * screen state behind it.
+     */
+    @Test
+    fun theReminderAlreadyUpWhenTheGateClosesStaysUp() {
+        start(gateClosed = false)
+        wantReminder(true)
+        assertReminderShows()
+
+        setGateState(closed = true)
+
+        assertReminderShows()
+    }
+
+    /** Held, not queued: what is no longer wanted by the time the gate opens is not raised. */
+    @Test
+    fun theHeldReminderIsDroppedWhenItStopsBeingWanted() {
+        start(gateClosed = true)
+        wantReminder(true)
+
+        wantReminder(false)
+        setGateState(closed = false)
+
+        assertReminderHeld()
+    }
+
+    /**
+     * The hold is per showing, so dismissing and being raised again behind the lock still holds.
+     */
+    @Test
+    fun theReminderRaisedAgainAfterTheGateClosesIsHeld() {
+        start(gateClosed = false)
+        wantReminder(true)
+        assertReminderShows()
+        wantReminder(false)
+
+        setGateState(closed = true)
+        wantReminder(true)
+
+        assertReminderHeld()
+    }
+
+    private fun start(gateClosed: Boolean) {
+        isGateClosed = gateClosed
+        compose.setContent {
+            CompositionLocalProvider(LocalIsGateClosed provides isGateClosed) {
+                if (isReminderWanted) {
+                    OnceUnlocked {
+                        MonthlyBackupReminder(onDismiss = {}, onBackup = {}, onDoNotRemind = {})
+                    }
+                }
+            }
+        }
+    }
+
+    private fun wantReminder(wanted: Boolean) {
+        isReminderWanted = wanted
+        compose.waitForIdle()
+    }
+
+    private fun setGateState(closed: Boolean) {
+        isGateClosed = closed
+        compose.waitForIdle()
+    }
+
+    private fun assertReminderShows() {
+        compose.waitUntil(SettleTimeoutMillis) { isReminderComposed() }
+
+        compose.onNodeWithText(reminderTitle).assertExists()
+    }
+
+    private fun assertReminderHeld() {
+        compose.waitForIdle()
+
+        compose.onNodeWithText(reminderTitle).assertDoesNotExist()
+    }
+
+    private fun isReminderComposed(): Boolean =
+        compose.onAllNodesWithText(reminderTitle).fetchSemanticsNodes().isNotEmpty()
+
+    private companion object {
+        val reminderTitle: String =
+            InstrumentationRegistry.getInstrumentation()
+                .targetContext
+                .getString(R.string.monthly_backup_reminder_title)
+
+        const val SettleTimeoutMillis = 5_000L
+    }
+}
--- a/app/src/androidTest/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlockedTest.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlockedTest.kt
@@ -1,6 +1,7 @@
 package com.vultisig.wallet.ui.screens.passcode
 
 import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.setValue
@@ -10,6 +11,7 @@ import androidx.compose.ui.test.onNodeWithText
 import androidx.test.platform.app.InstrumentationRegistry
 import com.vultisig.wallet.R
 import com.vultisig.wallet.ui.screens.home.MonthlyBackupReminder
+import org.junit.Assert.assertEquals
 import org.junit.Rule
 import org.junit.Test
 
@@ -24,6 +26,9 @@ class OnceUnlockedTest {
 
     private var isGateClosed by mutableStateOf(false)
     private var isReminderWanted by mutableStateOf(false)
+
+    /** Counts entries into composition, so a held reminder can be asserted never to have run. */
+    private var reminderCompositions = 0
 
     @Test
     fun theReminderShowsWhileTheGateIsOpen() {
@@ -69,16 +74,17 @@ class OnceUnlockedTest {
         assertReminderShows()
     }
 
-    /** Held, not queued: what is no longer wanted by the time the gate opens is not raised. */
+    /** Held, not queued: what stops being wanted behind the gate is never raised at all. */
     @Test
-    fun theHeldReminderIsDroppedWhenItStopsBeingWanted() {
+    fun theHeldReminderIsNeverComposedIfItStopsBeingWanted() {
         start(gateClosed = true)
         wantReminder(true)
 
         wantReminder(false)
         setGateState(closed = false)
 
-        assertReminderHeld()
+        compose.waitForIdle()
+        assertEquals(0, reminderCompositions)
     }
 
     /**
@@ -103,6 +109,10 @@ class OnceUnlockedTest {
             CompositionLocalProvider(LocalIsGateClosed provides isGateClosed) {
                 if (isReminderWanted) {
                     OnceUnlocked {
+                        DisposableEffect(Unit) {
+                            reminderCompositions++
+                            onDispose {}
+                        }
                         MonthlyBackupReminder(onDismiss = {}, onBackup = {}, onDoNotRemind = {})
                     }
                 }
--- a/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
+++ b/app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
@@ -19,6 +19,7 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.safeDrawingPadding
 import androidx.compose.foundation.layout.statusBars
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.key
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
```diff
@@ -137,12 +139,19 @@ internal fun MainActivityContent(
...
 
-            Box(
-                modifier =
-                    Modifier.fillMaxSize()
-                        .then(if (isGateClosed) Modifier.clearAndSetSemantics {} else Modifier)
-            ) {
-                SetupNavGraph(navController = navController, startDestination = startDestination)
+            // Neither the cover nor the semantics reset below reaches a window of its own, so
+            // content that opens one holds itself back on this instead. See OnceUnlocked.
+            CompositionLocalProvider(LocalIsGateClosed provides isGateClosed) {
```
aminsato, line 144: `isGateClosed` is `isLocked || passcodeState == Unknown`, and `isLocked` is false for `PasscodeState.Disabled`. Disabled is the default configuration — no app passcode set — and it is exactly the case where `PasscodeGuard` composes `BiometryAuthScreen()` into `overlayContent`. So on cold start with device credentials and no app passcode, the local is false, `OnceUnlocked` releases immediately, and `loadData`'s `showGlobalBackupReminder()` / `checkNotificationPrompt()` raise their sheets over the biometric screen with live "Back up" and "Do not remind me again" buttons, tappable before authentication.

That is the bug this PR is fixing, unfixed for the majority configuration. The local needs a predicate that is also true while the device-credential gate is up (`PasscodeGuardUiModel` knows it via `isIdentityProven`), not `isGateClosed` as it stands.

aminsato, line 144: Separate point about line 90, which isn't in the diff: `PopDialogDestinationsWhileLocked(navController, isLocked)` is still driven by `isLocked` while the new sheet gate is driven by `isGateClosed`. The two differ on `Unknown` and on `Disabled`.

`VaultAccountsViewModel.showVerifyFastVaultPasswordReminderIfRequired()` routes `Route.FastVaultPasswordReminder`, registered as `dialog<FastVaultPasswordReminder>` (NavGraph.kt:311). If it lands while the state is `Unknown`, the sweep's `LaunchedEffect` returns early and never collects, so the dialog's window is added after `LockWindow`'s and draws over the passcode gate — vault name plus a password prompt. In the passcode-disabled configuration `isLocked` is false for the whole session, so it is never popped at all. Both gates should key on the same predicate.

app/src/main/java/com/vultisig/wallet/ui/screens/home/VaultAccountsScreen.kt
```diff
@@ -87,30 +88,38 @@ internal fun VaultAccountsScreen(viewModel: VaultAccountsViewModel = hiltViewMod
         }
     }
 
+    // These three raise themselves off background work rather than off a tap, so each can open its
+    // window after the passcode lock has opened its own, and land above it.
     if (state.showMonthlyBackupReminder) {
```
aminsato, line 93: The month is consumed at check time, not at display time: `IsGlobalBackupReminderRequiredUseCaseImpl.invoke()` calls `setGlobalBackupReminderStatus(currentEpochMonth)` in the same branch that returns true.

With a passcode set, cold start runs `showGlobalBackupReminder()` behind the lock — the flag is written — and `OnceUnlocked` withholds the sheet. If the user backs out at the passcode screen or the process is killed while locked, the sheet is never composed but the month is already recorded, and there is no backup reminder for the rest of that month. Before this change the sheet at least got raised in that session.

Holding is only safe here if the consume moves to when the sheet is actually shown (or is rolled back when the hold never releases).

app/src/main/java/com/vultisig/wallet/ui/screens/home/VaultAccountsScreen.kt
```diff
@@ -87,30 +88,38 @@ internal fun VaultAccountsScreen(viewModel: VaultAccountsViewModel = hiltViewMod
         }
     }
 
+    // These three raise themselves off background work rather than off a tap, so each can open its
```
aminsato, line 91: Three of the four prompts `VaultAccountsViewModel.loadData` raises off background work are covered here. The fourth, `showVerifyFastVaultPasswordReminderIfRequired` (VaultAccountsViewModel.kt:327), routes `Route.FastVaultPasswordReminder` — a `dialog<>` destination, so `PopDialogDestinationsWhileLocked` drops it instead of holding it.

`loadData` doesn't re-run on unlock, and `setFastVaultPasswordReminderShownDate` is only written after a successful verification, so for fast-vault users with a passcode the reminder is swallowed on every launch: they are never prompted to verify their fast-vault password and can lose access to the server share without warning. The hold added here would have preserved it; the pop does not.

aminsato, line 91: About the permission collector just above (lines 81-89, outside the diff): it isn't gate-held.

`onNotificationEnable()` runs `markVaultPrompted` for every vault — suspending DataStore writes — before sending to `_requestNotificationPermission`. If the user taps "Enable" and immediately backgrounds the app or the screen times out, the coroutine finishes while locked, the channel emits, and the collector (still active, Home stays composed behind the gate) calls `permissionLauncher.launch(POST_NOTIFICATIONS)`. The system "Allow Vultisig to send you notifications?" dialog then sits over the lock and can be answered without unlocking; its result flips `showNotificationVaultSheet`. The sheet is held but the system prompt that precedes it isn't.

app/src/main/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlocked.kt
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Latched: once composed, [content] stays through a later lock, whose window covers it anyway.
+ * Callers keep their own condition around this, so one that lapses while the gate is closed raises
+ * nothing once it opens.
+ */
+@Composable
+internal fun OnceUnlocked(content: @Composable () -> Unit) {
+    val isGateClosed = LocalIsGateClosed.current
+    var isReleased by remember { mutableStateOf(!isGateClosed) }
+    LaunchedEffect(isGateClosed) { if (!isGateClosed) isReleased = true }
```
aminsato: The latch keeps an already-open sheet composed when the gate closes again, which is the right call for screen state — but that sheet lives in its own window, and the `clearAndSetSemantics {}` added for the gate is on the `Box` around `SetupNavGraph` in the activity's window, so it can't reach it.

With TalkBack on and `VaultNotificationOptInBottomSheet` open (it lists every secure vault by name), auto-lock on backgrounding leaves `isReleased` true, the sheet composed, and its content in the accessibility tree while the passcode screen is up. Visually covered, still readable — the same leak the guard's own comment at MainActivityContent.kt:82-84 exists to prevent. Worth clearing semantics on the held content when the gate closes.

aminsato: The release rule is encoded twice — the `remember` initializer and the effect both mean "open means released" — and the effect relaunches on every gate flip forever after the latch is set. A later change to the gate condition has to be made in both places or the two silently diverge.

The file being edited alongside this one already uses the simpler composition-phase idiom (`if (foregroundNotification != null) lastNotification = foregroundNotification`, MainActivityContent.kt:79-80). One encoding of the rule, no effect that keeps restarting, and no one-frame gap between the gate opening and the content composing:

```suggestion
    var isReleased by remember { mutableStateOf(false) }
    if (!isGateClosed) isReleased = true
```

(`LaunchedEffect` then drops out of the imports.)

app/src/androidTest/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlockedTest.kt
```diff
@@ -0,0 +1,146 @@
...
+        compose.waitForIdle()
+    }
+
+    private fun assertReminderShows() {
+        compose.waitUntil(SettleTimeoutMillis) { isReminderComposed() }
+
+        compose.onNodeWithText(reminderTitle).assertExists()
+    }
+
+    private fun assertReminderHeld() {
```
aminsato, line 139: `assertReminderShows()` concedes the sheet isn't composed synchronously — it polls `waitUntil(5_000)`. `assertReminderHeld()` asserts absence after a bare `waitForIdle()`.

So a regression where the gate releases one idle cycle late (release moved onto a lifecycle- or animation-driven signal, say) still leaves all four hold assertions green, and the sheet-over-the-lock bug this file exists to guard can be reintroduced without the instrumented tests failing. The hold assertions need the same settle window as the show assertions before they conclude nothing is there.

app/src/androidTest/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlockedTest.kt
```diff
@@ -0,0 +1,146 @@
...
+        assertReminderShows()
+
+        setGateState(closed = true)
+
+        assertReminderShows()
+    }
+
+    /** Held, not queued: what is no longer wanted by the time the gate opens is not raised. */
+    @Test
+    fun theHeldReminderIsDroppedWhenItStopsBeingWanted() {
```
aminsato: This one passes with `OnceUnlocked`'s body reduced to a bare `content()`: `wantReminder(false)` removes the whole subtree before the gate opens, so nothing can be shown either way. The property asserted is the caller's `if`, not the gate.

To actually cover "held, not queued" it needs the reminder still wanted at some sibling call site, or an assertion that the held content was never composed (a composition counter) rather than that it's absent after removal. As written it costs a `connectedDebugAndroidTest` run and guards nothing.

aminsato (commented):
The `OnceUnlocked` latch itself is sound, but the predicate it keys on and the set of prompts it covers leave the original bug reachable. Details inline — the first two are the ones I'd want resolved before merge.

app/src/main/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlocked.kt
rkokhatskyi: Fixed in 581b855.

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
rkokhatskyi: Declining — they differ on purpose. #5554 keyed the sweep on `isLocked` because `isGateClosed` would pop a restored dialog destination during `Unknown`, which every launch passes through. The `Unknown` case self-heals: the effect re-runs at `Locked` and `currentBackStackEntryFlow` emits on subscribe, covered by `theDialogDestinationsOpenWhenTheLockArrivesAreAllPopped`.

app/src/main/java/com/vultisig/wallet/ui/screens/home/VaultAccountsScreen.kt
rkokhatskyi: Unchanged by this PR — the stamp is at query time on `main` too. Worse: that write goes to the same DataStore that feeds `lastOpenedVaultId`, with no `distinctUntilChanged` anywhere, so `loadData` re-runs and clears the flag milliseconds later regardless. Both deserve their own ticket.

app/src/main/java/com/vultisig/wallet/ui/screens/home/VaultAccountsScreen.kt
rkokhatskyi: You are right it is lost every launch — `setFastVaultPasswordReminderShownDate` only writes from `back()`/`verify()`. But this hold cannot reach it: `DialogNavigator` creates the window at route time. The fix is holding the route, in a file this PR does not touch — I will take it separately.

rkokhatskyi: Real, but not ours to cover — the permission dialog is a system window, above everything the app owns. Only not launching it helps. Narrow path and the user just tapped Enable, so: separate ticket.

app/src/androidTest/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlockedTest.kt
rkokhatskyi: Declining — a late release holds the sheet longer, which is the safe direction. The failing direction is caught: with `OnceUnlocked` reduced to `content()`, four of six fail. A longer wait would just be an arbitrary sleep.

app/src/androidTest/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlockedTest.kt
rkokhatskyi: Fair, it was asserting the caller `if`. Fixed in 581b855 — the content bumps a composition counter, so it fails `expected:<0> but was:<1>` without the hold.

johnnyluo (approved):
LGTM

app/src/main/java/com/vultisig/wallet/app/activity/components/MainActivityContent.kt
rkokhatskyi: Not here — the hole is real, the predicate is wrong. `Disabled && !isIdentityProven` holds for the whole session of every no-passcode user, so it would suppress all three sheets forever. The only true signal is `isAuthorized`, a `rememberSaveable` inside `BiometryAuthPrompt.kt:46`; hoisting it redefines the gate for the semantics reset and the dialog sweep too. Separate issue.

app/src/main/java/com/vultisig/wallet/ui/screens/passcode/OnceUnlocked.kt
rkokhatskyi: Predates the latch — a sheet open when the lock arrives stayed composed after #5554 too. And `clearAndSetSemantics` here cannot reach it: the sheet content is its own `AndroidComposeView` behind `ModalBottomSheetDialog`.

Issue #5571, opened 2026-08-10
EvmApi.getNonce returns 0 on RPC error and uses "latest" instead of "pending"


aminsato:
## Summary

`EvmApi.getNonce` has two defects: it returns `BigInteger.ZERO` on any RPC error instead of failing, and it queries with the `"latest"` block tag instead of `"pending"`. Both can lead to failed ceremonies or silently replaced transactions.

## Details

`data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt:345-360`:

```kotlin
override suspend fun getNonce(address: String): BigInteger {
    val rpcResp =
        fetch<RpcResponse>(
            "eth_getTransactionCount",
            buildJsonArray {
                add(address)
                add("latest")
            },
        )
    if (rpcResp.error != null) {
        Timber.d("get nonce ,address: $address error: ${rpcResp.error.message}")
        return BigInteger.ZERO
    }
    return BigInteger(rpcResp.result?.removePrefix("0x") ?: "0", 16)
}
```

**Defect A — error swallowed to zero.** On any RPC error the method returns `BigInteger.ZERO` rather than throwing. That zero nonce is stamped into `BlockChainSpecific.Ethereum`, shown nowhere, and a full MPC keysign ceremony runs before broadcast fails with "nonce too low". This mirrors the pattern that was hardened for `getGasPrice` in #5399. On a genuinely fresh account (real nonce 0) it coincidentally succeeds, masking the bug.

**Defect B — `"latest"` instead of `"pending"`.** `eth_getTransactionCount(address, "latest")` counts only mined transactions. Sending a second EVM transaction while the first is still pending assigns both the same nonce. The second broadcast is then either rejected as an underpriced replacement or — if the user raised gas in Advanced Gas Settings — *replaces* the first payment, silently cancelling a transfer the user believes was sent.

Example: send #1 (nonce 5, 20 gwei, pending), immediately send #2 (also nonce 5, user set 40 gwei) → send #1 never lands.

## Suggested remediation

- Throw on RPC error (as `getGasPrice` now does) instead of returning `BigInteger.ZERO`.
- Use the `"pending"` block tag so consecutive sends serialize correctly.

## Severity

Medium — correctness bug that can cause burned keysign ceremonies and silently replaced/cancelled transactions.

johnnyluo:
Investigated both defects:

[...]

PR #5595, opened 2026-08-12
Fix EVM nonce RPC error handling

johnnyluo:
## Summary

Fixes defect A from #5571 by making `EvmApi.getNonce` propagate JSON-RPC errors instead of silently returning `BigInteger.ZERO`.

## Details

`eth_getTransactionCount` only enters the RPC error branch on an actual RPC failure such as rate limiting or node errors. Returning zero there can cause the app to sign with an invalid nonce and only fail after the keysign ceremony. This now throws `NetworkException`, matching the hardened `getGasPrice` behavior.

This intentionally does not change the block tag from `latest` to `pending`; the issue comment confirms `latest` is intended so resend/replace-by-fee flows can reuse the mined nonce.

## Validation

- `./gradlew :data:testDebugUnitTest --tests com.vultisig.wallet.data.api.EvmApiNonceTest`
- `./gradlew :data:ktfmtCheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when retrieving transaction nonces: RPC errors now report a network error instead of returning an invalid zero value.
  * Successfully returned nonces continue to be parsed correctly using the latest block information.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/EvmApi.kt
@@ -352,8 +352,10 @@ class EvmApiImp(
                 },
             )
         if (rpcResp.error != null) {
-            Timber.d("get nonce ,address: $address error: ${rpcResp.error.message}")
-            return BigInteger.ZERO
+            throw NetworkException(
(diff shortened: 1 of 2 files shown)


rkokhatskyi (approved):
LGTM

PR #5596, opened 2026-08-14
fix(defi): use the cross-platform 8k2mz Kamino attribution tag

aminsato:
## Summary

The attribution memo on Kamino deposits and withdraws carried `vs`. iOS writes `8k2mz` ([vultisig-ios#5166](https://github.com/vultisig/vultisig-ios/pull/5166)), so nothing this app originated was being attributed.

Kamino's kvault API takes no referrer or partner parameter, which is why attribution is an SPL Memo in the first place, and why the bytes are the whole point: the tag is the filter every downstream measurement of Vultisig-originated deposits keys on — Kamino's own attribution and the public Dune queries alike — and it has to be byte-identical on iOS, Android and Windows.

- `KaminoAttributionMemo.TAG` → `8k2mz`. The injector's base58 `data`, the validator's exact-match and its rejection message all read that one constant, so they move together.
- The golden test and the near-miss rejection list stay pinned to the literal rather than reading `TAG`. That is deliberate: if they derived it, a drift to any other valid string would leave the whole suite agreeing with itself and green while the attribution matched nothing.
- The near-miss list now follows iOS's — `8k2m`, `8k2mz1`, a trailing space, a leading space, `8K2MZ`. The tag is matched whole and as bytes, so each of those is a different memo that no downstream filter counts.

No behaviour changes beyond the bytes: the memo is still one instruction, naming no accounts, appended last after the compute-budget pair, and still the only memo the app will sign.

## Test plan

- [x] `./gradlew :data:testDebugUnitTest --tests "*Kamino*"` — passes, including the rewritten near-miss rejection cases
- [x] `./gradlew :data:assembleDebugAndroidTest` — compiles
- [ ] `KaminoAttributionMemoTest` (instrumented) — **not run locally, no device available.** It is the test that exercises the real WalletCore insertion against a live deposit fixture and pins the base58 encoding `7NBh6Z7`. That value was computed offline, not observed on device, so it needs the instrumented run before merge.
- [ ] One live deposit on device, confirming the on-chain memo reads `8k2mz`

## Summary by CodeRabbit

* **Updates**
  * Updated Kamino transaction attribution to use the `8k2mz` memo tag.
  * Improved validation coverage for formatted, case-variant, empty, and unrelated memo values.
  * Updated transaction attribution checks to reflect the new five-byte memo format.

### DIFF
--- a/data/src/androidTest/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoAttributionMemoTest.kt
+++ b/data/src/androidTest/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoAttributionMemoTest.kt
@@ -17,7 +17,7 @@ import wallet.core.jni.TransactionDecoder
 import wallet.core.jni.proto.Solana
 
 /**
- * Golden coverage for the `vs` attribution memo against a real Kamino deposit.
+ * Golden coverage for the `8k2mz` attribution memo against a real Kamino deposit.
  *
  * The fixture is a live `POST /ktx/kvault/deposit` response for the Steakhouse USDC vault: a v0
  * versioned transaction whose four instructions address 9 static account keys plus 18 loaded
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5609, opened 2026-08-15
fix(defi): make the Kamino compute budget agree with iOS

### LINKED ISSUE #5598: Kamino compute-unit price instruction is appended instead of following the limit
## Summary

`KaminoTransactionPreparer` injects the ComputeBudget pair with two WalletCore helpers that place their instruction differently: `setComputeUnitLimit` inserts at the front, `setComputeUnitPrice` appends at the back. Every prepared Kamino deposit and withdraw therefore carries the price last instead of right after the limit, and an iOS co-signer refuses to join. Introduced by #5588.

## Details

`data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparer.kt:162-171` calls `setComputeUnitLimit` then `setComputeUnitPrice`. At wallet-core 4.7.0, with no ComputeBudget instruction already present, the first inserts at index 0 and the second appends. Kamino's responses carry none, so the signed layout is:

```
[limit, ATA create, kvault deposit, farms init, farms stake, price, memo]
```

iOS emits both at positions 0 and 1 (`SolanaV0Transaction.swift:414-429`) and matches instructions positionally, so it finds the ATA create where the price should be, marks the transaction unreadable, and blocks the join. Swapping the call order does not help: the limit still lands at index 0.

## Impact

An Android-initiated Kamino deposit or withdraw cannot be co-signed from an iPhone — the Join button silently does nothing. No funds are at risk, and neither platform has shipped a store build with Kamino Earn yet.

## Suggested remediation

- Keep `setComputeUnitLimit`, but place the price with an explicit `SolanaTransaction.insertInstruction(tx, 1, ...)` carrying the ComputeBudget program id and `SetComputeUnitPrice` payload, the path the attribution memo already uses.
- Add an instrumented test pinning the order (limit 0, price 1, memo last); no existing test distinguishes the layouts.

aminsato:
Fixes #5598

### Transaction hash
https://orb.helius.dev/tx/nvE6t5pzCzSnsGjRhwfHUGc7bbpRqcaHyJn3MjBCKDHuLBrT1krTcAgcHz33g69dC71kDeW8BY2ASc8JELL7kFB

## Summary

An Android-initiated Kamino deposit or withdraw could not be co-signed from an iPhone. The reported cause was instruction ordering, but fixing that exposed three more disagreements with iOS in the same area — each of which independently blocks the join, and each with the same practical symptom, so they cannot be told apart from the co-signer's screen.

All four are fixed here. No funds are at risk and neither platform has shipped a store build with Kamino Earn.

## 1. The compute-unit price was appended instead of following the limit

`KaminoTransactionPreparer` injected the ComputeBudget pair with two WalletCore helpers that place their instruction differently: `setComputeUnitLimit` inserts at the front, `setComputeUnitPrice` appends. Kamino's responses carry no ComputeBudget instruction, so every prepared deposit and withdraw came out as:

```
[limit, ATA create, kvault deposit, farms init, farms stake, price, memo]
```

iOS emits the pair at 0 and 1 and matches the instructions that follow *by position*, so it found the ATA create where the price should be, marked the transaction unreadable and refused to join. Swapping the call order does not help — the limit still lands at index 0.

The price is now built locally (borsh: discriminator 3 followed by a little-endian `u64`) and inserted at index 1 with `SolanaTransaction.insertInstruction`, the path the attribution memo already uses. The ComputeBudget program key is already in the static keys after the limit, so WalletCore reuses it and no account indices shift.

## 2. The unit price had no ceiling

`unitPriceFor` only floored. iOS clamps the sample into `[20_000, 1_000_000]` and its decoder **refuses** a transaction priced outside that range (`KaminoTransactionDecoder.swift:383-384`), so an unclamped price is a transaction no iPhone will join.

That is not a hypothetical: the sample reaching this code comes from `SolanaApi.getMedianPriorityFee`, which returns `maxOf(median, 1_000_000)` capped at `100_000_000`. Every reading it produces is at or above the ceiling iOS enforces. `unitPriceFor` now clamps into the same range `KaminoComputeBudget.clampedUnitPrice` uses on iOS.

## 3. A response carrying its own compute budget would have been double-priced

Replacing `setComputeUnitPrice` with `insertInstruction` drops the in-place overwrite. Kamino's responses carry no ComputeBudget instruction today, but should one ever arrive with its own *price*, `setComputeUnitLimit` would leave it untouched and the insert would produce two `SetComputeUnitPrice` instructions — which the chain rejects, after a full signing ceremony has been spent on it.

The read-back that already ran after the limit now collects **every** ComputeBudget index and requires exactly `[0]`. One decode covers both the position contract and the absence of anything else, and it also catches a response supplying its own limit.

## 4. The payload recorded a different fee from the one in the bytes

The relayed `KeysignPayload` carried the generic Solana fee fields — the raw network sample and the app-wide `SOLANA_PRIORITY_FEE_LIMIT` of 100,000 units — while the bytes carried the Kamino budget just injected. `KaminoVerifyPresentation.priorityFeeAgrees` compares the two for **equality** before iOS will co-sign, so the join screen showed:

> The network fee inside this transaction is not the one shown above. Do not sign it.

and disabled the Join button. The limit alone was enough to fail it: 100,000 recorded against 320,000 in the bytes.

The same mismatch made the fee on the verify screen wrong on both platforms — priced off 100,000 units, it quoted 3,134,280 lamports of priority fee for a transaction whose real priority fee is 320,000, overstating the cost by an order of magnitude.

`BuildKaminoKeysignPayloadUseCase` now records both fields from the same functions the transaction is built with. `unitPriceFor` is idempotent, so the payload and the bytes agree by construction rather than by coincidence. Safe for signing: the raw path never reads these fields — they only feed `applyPriorityFee` on the transfer and staking signing inputs, and the Kamino bytes are already built.

## Also

- The ComputeBudget program id in the validator's allow-list points at `KaminoComputeBudget.PROGRAM_ID` instead of repeating the literal.
- `setUnitPriceData` drops a redundant mask and accumulator; `toInt().toByte()` already narrows to the low eight bits.
- The androidTest fixtures move to a shared `KaminoFixtures`, so the payload coverage and the preparer coverage assert against the same captured responses instead of drifting copies.

## Test plan

`:data:connectedDebugAndroidTest` — 36/36 pass on an Android 13 emulator. `:data:testDebugUnitTest` and `:data:lintDebug` pass.

Every fix was verified to be a real regression guard by reverting it and confirming the intended test fails:

| Reverted | Test | Failure |
|---|---|---|
| `insertInstruction` → `setComputeUnitPrice` | `the_compute_budget_leads_the_transaction_as_limit_then_price` | `expected:<ComputeBudget111...> but was:<ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL>` — the ATA create sitting where the price belongs |
| clamp → floor-only | `the_unit_price_is_capped_at_the_ceiling_an_iPhone_co_signer_will_accept` | `expected:<1000000> but was:<100000000>` |
| index list → `indexOfFirst` | `a_response_that_already_carries_a_compute_budget_is_refused_rather_than_double_priced` | no exception thrown |
| recorded budget → relayed generic values | `the_recorded_fee_is_what_the_user_will_actually_be_charged` | `expected:<320000> but was:<3134280>` — the exact priority fee behind the 0.00313928 SOL an iPhone refused |

Unit tests additionally pin the borsh encoding across the full `u64` range with its out-of-range rejections, and the price clamp at both bounds.

Known gap: the withdraw fixture is the **unstaked** shape, captured from a wallet holding no position. The staked shape — two extra `farms` instructions — has no fixture, so the withdraw compute limit rests on the iOS mainnet measurements rather than on anything this suite proves.

## Follow-up, not in this PR

The Kamino path takes its price sample from `getMedianPriorityFee`, whose own 1,000,000 floor sits exactly at the ceiling above. The clamp is therefore always binding and the 20,000 fallback is unreachable in practice, costing roughly 400,000 lamports of priority fee per transaction rather than 8,000. iOS avoids this by sampling `getRecentPrioritizationFees` raw and clamping that, instead of reusing the app-wide Solana value. Giving the Kamino path its own sample is a separate change.

## Bug Fixes
- Corrected Solana transaction ordering so compute limits and pricing appear before Kamino instructions, with memos placed last.
- Rejected duplicate or unexpected compute-budget instructions.
- Added safe fallback and maximum limits for compute-unit pricing.
- Improved compute-budget encoding and validation.
- Ensured signing payloads record accurate Kamino compute limits and fees.
- Corrected fee calculations to use native Solana fees and preserve blockchain-specific transaction data.
- Prevented phantom token-account rent charges.

## Tests
- Expanded coverage for ordering, memo placement, encoding, boundary values, price capping, malformed responses, and invalid inputs.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoComputeBudget.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoComputeBudget.kt
@@ -1,6 +1,10 @@
 package com.vultisig.wallet.data.blockchain.solana.kamino
 
 import java.math.BigInteger
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.put
+import wallet.core.jni.Base58
 
 /**
  * Which side of a vault a transaction is on. Selects the compute-unit budget and the validator's
@@ -21,6 +25,15 @@ enum class KaminoAction {
  * the transaction actually needs.
  */
 object KaminoComputeBudget {
+
+    /** The ComputeBudget program both injected instructions invoke. */
+    const val PROGRAM_ID = "ComputeBudget111111111111111111111111111111"
+
+    /** `ComputeBudgetInstruction::SetComputeUnitLimit`, borsh discriminator 2, then a `u32`. */
+    const val SET_UNIT_LIMIT_DISCRIMINATOR = 2
+
+    /** `ComputeBudgetInstruction::SetComputeUnitPrice`, borsh discriminator 3, then a `u64`. */
+    const val SET_UNIT_PRICE_DISCRIMINATOR = 3
 
     /**
      * Compute units these transactions actually consume, measured on mainnet via
@@ -95,4 +108,44 @@ object KaminoComputeBudget {
     }
 
     private val MICRO_LAMPORTS_PER_LAMPORT = BigInteger.valueOf(1_000_000)
+
+    /**
+     * The `SetComputeUnitPrice` instruction as WalletCore's JSON instruction format, ready to
+     * insert at a chosen position.
+     *
+     * Built here rather than left to `SolanaTransaction.setComputeUnitPrice`, which appends: the
+     * two budget instructions have to sit together at the front, and the helper would leave the
+     * price behind every instruction Kamino built. See [KaminoTransactionPreparer] for why the
+     * position is part of the cross-platform contract.
+     *
+     * The instruction takes no accounts, and `data` is base58 — the encoding WalletCore's JSON
+     * format expects, not base64.
+     */
+    fun setUnitPriceInstructionJson(price: BigInteger): String =
+        buildJsonObject {
+                put("programId", PROGRAM_ID)
+                put("accounts", JsonArray(emptyList()))
+                put("data", Base58.encodeNoCheck(setUnitPriceData(price)))
+            }
+            .toString()
+
+    /**
+     * Borsh-encoded `SetComputeUnitPrice`: the discriminator byte followed by the micro-lamport
+     * price as a little-endian `u64`.
+     */
+    fun setUnitPriceData(price: BigInteger): ByteArray {
+        require(price.signum() >= 0 && price.bitLength() <= 64) {
+            "compute-unit price $price does not fit a u64"
+        }
+        val data = ByteArray(9)
+        data[0] = SET_UNIT_PRICE_DISCRIMINATOR.toByte()
+        var remaining = price
+        for (offset in 1..8) {
+            data[offset] = remaining.and(BYTE_MASK).toInt().toByte()
+            remaining = remaining.shiftRight(8)
+        }
+        return data
+    }
+
+    private val BYTE_MASK = BigInteger.valueOf(0xFF)
 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoComputeBudget.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoComputeBudget.kt
@@ -75,6 +75,22 @@ object KaminoComputeBudget {
      */
     val FALLBACK_UNIT_PRICE: BigInteger = BigInteger.valueOf(20_000)
 
+    /**
+     * Ceiling on the sampled price, in micro-lamports per compute unit.
+     *
+     * The sample is a number a remote node hands us, multiplied by a six-figure limit to produce
+     * lamports the user pays, so unbounded it is an unbounded fee. At this ceiling and the largest
+     * limit the priority fee tops out at 400,000 lamports (0.0004 SOL).
+     *
+     * It is also a cross-platform contract, not just a spend cap: iOS clamps into the same `[floor,
+     * ceiling]` range and its decoder *refuses* a transaction priced outside it
+     * (`KaminoTransactionDecoder.swift`), so anything above this is a transaction an iPhone
+     * co-signer will not join. That is not a hypothetical here — the sample this receives comes
+     * from `SolanaApi.getMedianPriorityFee`, which floors at the app-wide 1,000,000 and caps at
+     * 100,000,000, so every congested-network sample would land above the ceiling unclamped.
+     */
+    val MAX_UNIT_PRICE: BigInteger = BigInteger.valueOf(1_000_000)
+
     fun unitLimitFor(vault: KaminoVault, action: KaminoAction): BigInteger =
         when (action) {
             KaminoAction.WITHDRAW -> WITHDRAW_UNIT_LIMIT
@@ -86,9 +102,15 @@ object KaminoComputeBudget {
                 }
         }.let(BigInteger::valueOf)
 
-    /** Never price below the floor, however stale or absent the network sample is. */
+    /**
+     * Brings a sampled network price inside `[FALLBACK_UNIT_PRICE, MAX_UNIT_PRICE]`, the same range
+     * iOS clamps into and the only range its decoder accepts.
+     *
+     * The floor is a floor rather than a default: a sample below it would under-tip a transaction
+     * that has to land before its blockhash expires.
+     */
     fun unitPriceFor(networkPrice: BigInteger?): BigInteger =
-        networkPrice?.takeIf { it > FALLBACK_UNIT_PRICE } ?: FALLBACK_UNIT_PRICE
+        (networkPrice ?: FALLBACK_UNIT_PRICE).coerceIn(FALLBACK_UNIT_PRICE, MAX_UNIT_PRICE)
 
     /**
      * The priority fee this transaction will actually be charged, in lamports.
@@ -134,18 +156,11 @@ object KaminoComputeBudget {
      * price as a little-endian `u64`.
      */
     fun setUnitPriceData(price: BigInteger): ByteArray {
-        require(price.signum() >= 0 && price.bitLength() <= 64) {
+        // The sign is checked separately: `bitLength` ignores it, so -1 would clear the bound.
+        require(price.signum() >= 0 && price.bitLength() <= Long.SIZE_BITS) {
             "compute-unit price $price does not fit a u64"
         }
-        val data = ByteArray(9)
-        data[0] = SET_UNIT_PRICE_DISCRIMINATOR.toByte()
-        var remaining = price
-        for (offset in 1..8) {
-            data[offset] = remaining.and(BYTE_MASK).toInt().toByte()
-            remaining = remaining.shiftRight(8)
-        }
-        return data
+        return byteArrayOf(SET_UNIT_PRICE_DISCRIMINATOR.toByte()) +
+            ByteArray(Long.SIZE_BYTES) { byte -> price.shiftRight(8 * byte).toInt().toByte() }
     }
-
-    private val BYTE_MASK = BigInteger.valueOf(0xFF)
 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparer.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparer.kt
@@ -169,17 +169,26 @@ class KaminoTransactionPreparer @Inject constructor(private val kaminoApi: Kamin
                 "WalletCore could not set the Kamino compute-unit limit"
             }
 
-        // WalletCore chooses where the limit goes — index 0, unless the first instruction is
-        // `AdvanceNonceAccount`, which Kamino never builds — so the position is observed rather
-        // than
-        // assumed. Anything but 0 is not the layout iOS emits, and is refused here rather than sent
-        // to a co-signer that would silently decline it.
-        val limitIndex =
-            KaminoTransactionDecoder.decode(withLimit).instructions.indexOfFirst {
-                it.programId == KaminoComputeBudget.PROGRAM_ID
+        // Two assumptions are read back here rather than trusted, because the insert below adds a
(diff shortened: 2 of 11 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparer.kt
```diff
@@ -166,8 +169,41 @@ class KaminoTransactionPreparer @Inject constructor(private val kaminoApi: Kamin
...
+        // Kamino's responses carry no compute budget, so that pair leaves the price behind every
+        // instruction Kamino built, and an iOS co-signer — which emits the two at 0 and 1 and reads
+        // the rest positionally — finds the wrong instruction where the price belongs, marks the
+        // transaction unreadable and refuses to join. Swapping the call order does not help; the
+        // limit still lands at index 0.
+        return checkNotNull(
+            SolanaTransaction.insertInstruction(
+                withLimit,
+                UNIT_PRICE_INDEX,
+                KaminoComputeBudget.setUnitPriceInstructionJson(price),
```
rkokhatskyi, line 205: `price` here has no ceiling because `KaminoComputeBudget.unitPriceFor` only floors, and `getMedianPriorityFee` already returns `maxOf(median, 1_000_000)` while iOS refuses anything above `1_000_000` at `KaminoTransactionDecoder.swift:384`, so an iPhone still declines every transaction this builds

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparer.kt
```diff
@@ -166,8 +169,41 @@ class KaminoTransactionPreparer @Inject constructor(private val kaminoApi: Kamin
...
+
+        // Deliberately not `SolanaTransaction.setComputeUnitPrice`. The two helpers place their
+        // instruction differently: the limit is *inserted at the front*, the price is *appended*.
+        // Kamino's responses carry no compute budget, so that pair leaves the price behind every
+        // instruction Kamino built, and an iOS co-signer — which emits the two at 0 and 1 and reads
+        // the rest positionally — finds the wrong instruction where the price belongs, marks the
+        // transaction unreadable and refuses to join. Swapping the call order does not help; the
+        // limit still lands at index 0.
+        return checkNotNull(
+            SolanaTransaction.insertInstruction(
```
rkokhatskyi, line 202: Replacing `setComputeUnitPrice` with `insertInstruction` drops the in place overwrite, so a response carrying its own compute budget now signs two `SetComputeUnitPrice` and the chain rejects it after the ceremony
```kotlin
check(KaminoTransactionDecoder.decode(base64Transaction).instructions.none {
    it.programId == KaminoComputeBudget.PROGRAM_ID
}) { "Kamino returned a compute budget the app did not ask for" }
```

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoComputeBudget.kt
```diff
@@ -95,4 +108,44 @@ object KaminoComputeBudget {
...
+            "compute-unit price $price does not fit a u64"
+        }
+        val data = ByteArray(9)
+        data[0] = SET_UNIT_PRICE_DISCRIMINATOR.toByte()
+        var remaining = price
+        for (offset in 1..8) {
+            data[offset] = remaining.and(BYTE_MASK).toInt().toByte()
+            remaining = remaining.shiftRight(8)
+        }
+        return data
```
rkokhatskyi: The mask and the running accumulator are unnecessary because `toInt().toByte()` already narrows to the low eight bits, so the body collapses to one expression and `BYTE_MASK` can be deleted
```suggestion
        // A negative price passes the bound alone because bitLength ignores the sign.
        require(price.signum() >= 0 && price.bitLength() <= Long.SIZE_BITS) {
            "compute-unit price $price does not fit a u64"
        }
        return byteArrayOf(SET_UNIT_PRICE_DISCRIMINATOR.toByte()) +
            ByteArray(Long.SIZE_BYTES) { i -> price.shiftRight(8 * i).toInt().toByte() }
```

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparer.kt
aminsato: Confirmed, and worse than a spend cap — `getMedianPriorityFee` returns `maxOf(median, 1_000_000).coerceAtMost(100_000_000)`, so the sample handed to `prepare` is *always* at or above the ceiling iOS enforces at `KaminoTransactionDecoder.swift:384`. Every congested-network reading would have built a transaction no iPhone joins.

`unitPriceFor` now clamps into `[FALLBACK_UNIT_PRICE, MAX_UNIT_PRICE]` = `[20_000, 1_000_000]`, the same range `KaminoComputeBudget.clampedUnitPrice` uses on iOS, with `MAX_UNIT_PRICE` added alongside the floor. `priorityFeeLamports` runs through the same function, so the Max-amount reserve moves with it.

Covered by a unit test on the clamp and by `the_unit_price_is_capped_at_the_ceiling_an_iPhone_co_signer_will_accept` end-to-end. Reverting `unitPriceFor` to the floor-only form fails the latter with `expected:<1000000> but was:<100000000>`.

One consequence worth naming: because that 1,000,000 floor sits exactly at the ceiling, the price is now always the ceiling and the 20,000 fallback is unreachable in practice — about 400,000 lamports of priority fee per transaction rather than 8,000. iOS avoids it by sampling `getRecentPrioritizationFees` raw and clamping that, instead of reusing the app-wide Solana value. Giving the Kamino path its own sample is a separate change; happy to open it if you want it before this merges.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparer.kt
aminsato: Right — an insert cannot overwrite, so a response carrying its own `SetComputeUnitPrice` would have signed two and been rejected on chain after the ceremony.

Rather than a separate pre-check, I folded it into the read-back that was already there after the limit: it now collects *every* ComputeBudget index and requires exactly `[0]`. That covers both the position contract and the absence of anything else in one decode, and it also catches a response supplying its own *limit* — `setComputeUnitLimit` would overwrite that in place wherever it sits, leaving the indices something other than `[0]`.

```kotlin
val instructions = KaminoTransactionDecoder.decode(withLimit).instructions
val budgetIndices =
    instructions.indices.filter { instructions[it].programId == KaminoComputeBudget.PROGRAM_ID }
check(budgetIndices == listOf(UNIT_LIMIT_INDEX)) {
    "expected the app's compute-unit limit alone at index $UNIT_LIMIT_INDEX, " +
        "found ComputeBudget instructions at $budgetIndices"
}
```

`a_response_that_already_carries_a_compute_budget_is_refused_rather_than_double_priced` builds a fixture with its own price appended and asserts `prepare` throws. Putting the old `indexOfFirst` check back fails it on the `assertThrows`.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoComputeBudget.kt
aminsato: Taken as written — `BYTE_MASK` is gone and the accumulator with it:

```kotlin
// The sign is checked separately: `bitLength` ignores it, so -1 would clear the bound.
require(price.signum() >= 0 && price.bitLength() <= Long.SIZE_BITS) {
    "compute-unit price $price does not fit a u64"
}
return byteArrayOf(SET_UNIT_PRICE_DISCRIMINATOR.toByte()) +
    ByteArray(Long.SIZE_BYTES) { byte -> price.shiftRight(8 * byte).toInt().toByte() }
```

The existing encoding test still pins the same bytes across the full `u64` range and both out-of-range rejections.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparer.kt
```diff
@@ -166,8 +169,41 @@ class KaminoTransactionPreparer @Inject constructor(private val kaminoApi: Kamin
...
+                withLimit,
+                UNIT_PRICE_INDEX,
+                KaminoComputeBudget.setUnitPriceInstructionJson(price),
+            )
+        ) {
             "WalletCore could not set the Kamino compute-unit price"
         }
```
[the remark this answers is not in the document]
aminsato: Valid — fixed, though the check landed one step later than suggested.

Instead of rejecting on `base64Transaction` before mutation, the read-back that already ran after `setComputeUnitLimit` now collects every ComputeBudget index and requires exactly `[0]`. Same guarantee in one decode rather than two, and it additionally catches a response supplying its own *limit*, which `setComputeUnitLimit` would overwrite in place wherever it sat.

On the validator: it deliberately does not count or position ComputeBudget instructions, because the preparer is the only thing that puts them there and the transaction is refused above before validation ever sees it. `a_response_that_already_carries_a_compute_budget_is_refused_rather_than_double_priced` pins that, and fails if the check is relaxed back to `indexOfFirst`.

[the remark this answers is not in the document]
aminsato: Thanks — the recorded learning matches the intent, with one clarification worth pinning: the check is `budgetIndices == listOf(UNIT_LIMIT_INDEX)`, an equality on the full index list rather than a count, so it also fails if the single ComputeBudget instruction sits anywhere other than index 0. That position is the cross-platform half of the contract — iOS emits limit-then-price at 0 and 1 and matches everything after them positionally — so it has to fail as loudly as a duplicate does.

Resolving this thread.

rkokhatskyi (approved):
LGTM

PR #5612, opened 2026-08-16
Fix Kamino SOL max deposit reserve

### LINKED ISSUE #5599: Kamino Max deposit under-reserves SOL and is always rejected at preflight
## Summary

The Max/100% chip on the Allez SOL Kamino vault offers more than the transaction it builds can pay for. It reserves 3,046,280 lamports against a first deposit that actually costs 11,727,640 — an 8,681,360-lamport shortfall, whatever the balance is. Every first Max deposit into the SOL vault is co-signed and broadcast, then rejected at preflight. Introduced by #5588.

## Details

`app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt:229-255` builds the reserve as base fee (1,000,000) + priority fee (7,000) + wrapped-SOL rent (2,039,280). Two of those three are wrong.

The priority fee is computed with `networkPrice = null`, which takes `KaminoComputeBudget`'s 20,000 µlamports/CU fallback. The build path instead uses `solanaSpecific.priorityFee` (`BuildKaminoKeysignPayloadUseCase.kt:63`), floored at 1,000,000 µlamports/CU, so the charged fee is 350,000 lamports, not 7,000.

The wSOL ATA is also not the only account a first deposit creates: it also creates the share ATA and the farm `UserState` account, neither of them reserved. Repeat deposits are unaffected — `createIdempotent` charges rent only when the account is absent.

## Impact

The wallet spends a full MPC ceremony on a transaction that cannot succeed. No funds are lost: preflight is on (`SolanaApi.kt:253` does not set `skipPreflight`), so nothing lands.

## Suggested remediation

- Pass the same unit price the build path uses (`solanaSpecific.priorityFee`) instead of `null`, and reserve rent for the share ATA and the farm `UserState` as well as the wSOL ATA.
- Better, follow iOS (`KaminoTransactionPreparer.swift:245-264`): simulate a probe deposit and read the payer's post-execution balance back from the node, which gets every rent term right.

johnnyluo:
## Summary
- calculate Kamino SOL max deposit priority headroom from the same Solana priority price used by the payload build path
- reserve first-deposit rent for missing wSOL ATA, share ATA, and Kamino farm UserState accounts
- pin the first-deposit SOL max amount regression in Kamino amount tests

Fixes #5599

on chain tx link: https://orbmarkets.io/tx/3JcwWazxR9hg32mCc3rgweetVK6nzEaiKw3bRFah7GSoFbeNtEw7NCqB2yFmBdfjziVs2GZrzTk4eEpztYyiJUpa
## Testing
- ./gradlew :app:testDebugUnitTest --tests com.vultisig.wallet.ui.models.solanastaking.KaminoAmountViewModelTest
- ./gradlew :app:ktfmtCheck

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SOL deposit balance calculations by reserving required network fees and account-creation rent.
  * Added fallback handling when fee, rent, account, or position data cannot be retrieved.
  * Increased accuracy for first-time Kamino deposits, including newly created account costs.

* **Tests**
  * Updated deposit scenarios and balance expectations to reflect revised fee and rent calculations.
  * Added coverage for priority fees and first-deposit account creation costs.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
@@ -39,6 +39,7 @@ import com.vultisig.wallet.data.repositories.BlockChainSpecificRepository
 import com.vultisig.wallet.data.repositories.ChainAccountAddressRepository
 import com.vultisig.wallet.data.repositories.DepositTransactionRepository
 import com.vultisig.wallet.data.repositories.VaultRepository
+import com.vultisig.wallet.data.utils.runCatchingCancellable
 import com.vultisig.wallet.data.utils.safeLaunch
 import com.vultisig.wallet.ui.navigation.Destination
 import com.vultisig.wallet.ui.navigation.Navigator
@@ -195,7 +196,8 @@ constructor(
             }
 
             val available = spendableBalance(vault, coin)
-            val vaultState = runCatching { kaminoApi.getVaultState(vault.address) }.getOrNull()
+            val vaultState =
+                runCatchingCancellable { kaminoApi.getVaultState(vault.address) }.getOrNull()
 
             _state.update {
                 it.copy(
@@ -239,27 +241,22 @@ constructor(
             .movePointLeft(coin.decimal)
     }
 
+    /**
+     * Reads the network's median priority price directly rather than through
+     * [BlockChainSpecificRepository.getSpecific], which also fetches a recent blockhash and ATA
+     * lookups unrelated to pricing. Any one of those failing would fail the whole call, nulling out
+     * the price too and dropping the floor from the app-wide 1,000,000 to
+     * [KaminoComputeBudget.FALLBACK_UNIT_PRICE]'s 20,000 — under-reserving the exact way #5599
+     * failed, just at a smaller scale. [SolanaApi.getMedianPriorityFee] already falls back to
+     * 1,000,000 internally and never throws but for cancellation, so nothing here needs its own
+     * fallback.
+     */
     private suspend fun kaminoDepositPriorityFee(vault: KaminoVault, coin: Coin): BigInteger {
-        val solanaSpecific =
-            runCatching {
-                    blockChainSpecificRepository
-                        .getSpecific(
-                            chain = Chain.Solana,
-                            address = coin.address,
-                            token = coin,
-                            gasFee = TokenValue(SolanaHelper.DefaultFeeInLamports, coin),
-                            isSwap = false,
-                            isMaxAmountEnabled = false,
-                            isDeposit = true,
-                        )
-                        .blockChainSpecific as? BlockChainSpecific.Solana
-                }
-                .getOrNull()
-
+        val networkPrice = solanaApi.getMedianPriorityFee(listOf(coin.address))
         return KaminoComputeBudget.priorityFeeLamports(
             vault = vault,
             action = KaminoAction.DEPOSIT,
-            networkPrice = solanaSpecific?.priorityFee,
+            networkPrice = networkPrice,
         )
     }
 
@@ -281,19 +278,33 @@ constructor(
     }
 
     private suspend fun tokenAccountRentReserve(): BigInteger =
-        runCatching { solanaApi.getMinimumBalanceForRentExemption() }
+        runCatchingCancellable { solanaApi.getMinimumBalanceForRentExemption() }
             .getOrNull()
             ?.takeIf { it.signum() > 0 } ?: SPL_TOKEN_ACCOUNT_RENT_LAMPORTS
 
     private suspend fun tokenAccountExists(walletAddress: String, mint: String): Boolean =
-        runCatching {
+        runCatchingCancellable {
                 solanaApi.getTokenAssociatedAccountByOwner(walletAddress, mint).first != null
             }
             .getOrDefault(false)
 
+    /**
+     * Whether the wallet already has a farm UserState account for this vault, so a redeposit does
+     * not need to pay its rent again.
+     *
+     * Goes through [KaminoWithdrawEligibility.resolve] rather than checking for a bare entry in the
+     * response: the same endpoint keeps a zero-share row for a wallet that has fully exited, and
+     * the withdraw flow already treats that row as no position. Checking entry presence alone would
+     * disagree with that and waive the rent on an account that may no longer exist.
+     */
     private suspend fun hasKaminoVaultPosition(walletAddress: String, vault: KaminoVault): Boolean =
-        runCatching {
-                kaminoApi.getUserPositions(walletAddress).any { it.vaultAddress == vault.address }
+        runCatchingCancellable {
+                val entry =
+                    kaminoApi.getUserPositions(walletAddress).firstOrNull {
+                        it.vaultAddress == vault.address
+                    }
+                KaminoWithdrawEligibility.resolve(entry, vault.sharesDecimals) is
+                    KaminoWithdrawEligibility.Withdrawable
             }
             .getOrDefault(false)
 
@@ -305,7 +316,8 @@ constructor(
      * is denominated in is derived from those, never the other way round.
      */
     private suspend fun loadWithdrawState(vault: KaminoVault, coin: Coin) {
-        val positions = runCatching { kaminoApi.getUserPositions(coin.address) }.getOrNull()
+        val positions =
+            runCatchingCancellable { kaminoApi.getUserPositions(coin.address) }.getOrNull()
         val eligibility =
             if (positions == null) {
                 // A failed read is neither "nothing to withdraw" nor "everything": refuse it.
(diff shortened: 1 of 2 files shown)


rkokhatskyi:
@johnnyluo please attach tx link for this change

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -220,43 +220,83 @@ constructor(
...
+                            isDeposit = true,
+                        )
+                        .blockChainSpecific as? BlockChainSpecific.Solana
+                }
+                .getOrNull()
+
+        return KaminoComputeBudget.priorityFeeLamports(
+            vault = vault,
+            action = KaminoAction.DEPOSIT,
+            networkPrice = solanaSpecific?.priorityFee,
```
rkokhatskyi: `networkPrice` falls to null whenever the unrelated `getRecentBlockHash()` call inside the same `getSpecific()` fails, so `unitPriceFor` floors at `KaminoComputeBudget`'s own 20,000 instead of the app wide 1,000,000. That reserves 7,000 lamports instead of 350,000, reproducing #5599's preflight rejection at a smaller scale

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -220,43 +220,83 @@ constructor(
...
+     * For a first SOL-vault deposit that is several things, not one: the base fee, the charged
+     * Kamino priority fee (price × compute limit), and rent for every account the deposit creates:
+     * wSOL ATA, share ATA and farm user-state. Reserving only the transfer-shaped fee leaves a Max
+     * deposit unable to fund those accounts and it fails at preflight.
      *
      * A token deposit pays all of that in SOL, so the token balance is spendable in full.
      */
     private suspend fun spendableBalance(vault: KaminoVault, coin: Coin): BigDecimal {
         val balance = balanceRepository.getTokenValue(coin.address, coin).first().value
         if (!coin.isNativeToken) return BigDecimal(balance).movePointLeft(coin.decimal)
```
rkokhatskyi, line 234: This early return skips all SOL side reservation for USDC vaults, so a first Steakhouse or RWA deposit never checks native SOL against the same share ATA and farm UserState rent this PR reserves for Allez. A low SOL USDC holder hits the identical preflight rejection this PR fixes

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -220,43 +220,83 @@ constructor(
...
-                    .getOrNull()
-                    ?.takeIf { it.signum() > 0 } ?: SPL_TOKEN_ACCOUNT_RENT_LAMPORTS
-            } else {
-                BigInteger.ZERO
-            }
+        val priorityFee = kaminoDepositPriorityFee(vault, coin)
+        val rentReserve = kaminoNativeDepositRentReserve(vault, coin)
 
-        val headroom = SolanaHelper.DefaultFeeInLamports + priorityFee + wrappedSolRent
+        val headroom = SolanaHelper.DefaultFeeInLamports + priorityFee + rentReserve
```
rkokhatskyi, line 239: `headroom` now reserves up to 10.4M lamports of ATA and farm rent that `submit()`'s `gasFee` at line 482 never includes, since that field is still base plus priority fee only. A first Max SOL deposit now shows an amount and fee that no longer sum to the starting balance

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -220,43 +220,83 @@ constructor(
...
+
+    private suspend fun tokenAccountExists(walletAddress: String, mint: String): Boolean =
+        runCatching {
+                solanaApi.getTokenAssociatedAccountByOwner(walletAddress, mint).first != null
+            }
+            .getOrDefault(false)
+
+    private suspend fun hasKaminoVaultPosition(walletAddress: String, vault: KaminoVault): Boolean =
+        runCatching {
+                kaminoApi.getUserPositions(walletAddress).any { it.vaultAddress == vault.address }
```
rkokhatskyi: `hasKaminoVaultPosition` waives farm rent on bare entry presence, while `KaminoWithdrawEligibility.resolve` treats a zero share entry from the same endpoint as no position. A fully exited wallet whose position row persists would under reserve its next deposit and hit preflight rejection again

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -220,43 +220,83 @@ constructor(
...
+                            address = coin.address,
+                            token = coin,
+                            gasFee = TokenValue(SolanaHelper.DefaultFeeInLamports, coin),
+                            isSwap = false,
+                            isMaxAmountEnabled = false,
+                            isDeposit = true,
+                        )
+                        .blockChainSpecific as? BlockChainSpecific.Solana
+                }
+                .getOrNull()
```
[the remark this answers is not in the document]
johnnyluo: Fixed — replaced every suspend-wrapping `runCatching` in this file with the existing cancellation-safe `runCatchingCancellable` helper (`data/src/main/kotlin/.../data/utils/RunCatchingCancellable.kt`), including the two sites you flagged.

app/src/test/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModelTest.kt
```diff
@@ -283,14 +300,9 @@ internal class KaminoAmountViewModelTest {
...
-        // 7,000 lamports) less the base fee — so strictly below 0.998, and far below a naive
-        // fee-only headroom.
-        assertTrue(
-            available < BigDecimal("0.998"),
-            "expected rent + priority fee reserved, was $available",
-        )
-        assertTrue(available > BigDecimal("0.99"), "reserved too much, was $available")
+        // 1 SOL less base fee (1,000,000), priority fee (1,000,000 µlamports x 350,000 CU =
+        // 350,000), two token-account rents (2,039,280 each) and farm UserState rent (6,299,080).
+        assertEquals(0, BigDecimal("0.98827236").compareTo(available))
```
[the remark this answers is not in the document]
johnnyluo: This file uses `kotlin.test.assertEquals` throughout (no Kotest dependency in this test), so switching one assertion to Kotest's `shouldBe` would introduce a new test framework for a single line rather than removing an inconsistency. Skipping this one.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
johnnyluo: Good catch — confirmed: `getSpecific()`'s `coroutineScope` throws as a whole if `getRecentBlockHash()` fails, which nulled the whole `solanaSpecific` including the unrelated priority price. Fixed by reading the price directly via `solanaApi.getMedianPriorityFee()`, which has its own internal fallback to the app-wide 1,000,000 and no longer depends on the blockhash/ATA lookups succeeding.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
johnnyluo: Agreed this is real, but it's a different bug than #5599 (SOL-vault only) — fixing it needs a second, SOL-denominated balance check layered on top of a USDC-denominated `available`, which is a bigger change than this PR's scope. Filed as a follow-up: https://github.com/vultisig/vultisig-android/issues/5617

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
johnnyluo: Fixed — `submit()`'s `gasFee` now folds in the same first-deposit rent reserve `spendableBalance()` computed, so amount + fee sums back to the wallet's starting balance on a Max native-SOL deposit. Added a test asserting the exact lamport total.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
johnnyluo: Fixed — `hasKaminoVaultPosition` now goes through `KaminoWithdrawEligibility.resolve` and requires `Withdrawable` (i.e. real nonzero spendable shares) instead of trusting bare entry presence, so a zero-share row left behind by a full exit no longer waives the farm rent.

rkokhatskyi (approved):
LGTM

PR #5613, opened 2026-08-16
fix(defi): send a Kamino withdraw to the signer, not to the vault

### LINKED ISSUE #5605: Kamino withdraw sets the destination address to the vault instead of the signer's own account
## Summary

A Kamino Earn withdraw builds its keysign payload with `toAddress` set to the Kamino vault address, the same value a deposit uses. A withdraw pays the signer's own Solana account; the vault is the source, not the destination. Introduced by #5588.

## Details

`data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/BuildKaminoKeysignPayloadUseCase.kt:66-70`:

```kotlin
return KeysignPayload(
    coin = coin,
    toAddress = vault.address,
```

The `action` parameter (`KaminoAction.DEPOSIT` / `WITHDRAW`) is in scope but nothing branches on it. `KaminoAmountViewModel.kt:428` mirrors the same value into the `DepositTransaction` the verify screen reads. On the initiator, `VerifyDepositScreen.kt:237-245` renders the `To` row from that field under a "You're withdrawing" title. A Kamino payload carries no memo, so `JoinKeysignViewModel.isDepositPayload` returns false on an Android co-signer and the vault address lands on the generic Send verify screen with no withdraw framing. iOS branches on the action: `KaminoKeysignPayloadFactory.swift:81` uses `coin.address` for a withdraw.

## Impact

Both devices of a co-signed pair are shown the vault address as the destination of a withdrawal. No funds are at risk: routing lives in the relayed transaction bytes. The defect is what the approving user is told. No released build ships Kamino Earn.

## Suggested remediation

- Branch `toAddress` on `action` in `BuildKaminoKeysignPayloadUseCase.kt`: `coin.address` for `WITHDRAW`, `vault.address` for `DEPOSIT`.
- Apply the same branch to `dstAddress` at `KaminoAmountViewModel.kt:428` — the verify screen reads the stored `DepositTransaction`, not the returned payload.
- Add a withdraw case to `KaminoAmountViewModelTest` asserting `dstAddress == coin.address`.

aminsato:
Fixes #5605

## Summary

A Kamino Earn withdraw is paid out of the vault into the user's own Solana account, but both the keysign payload and the stored `DepositTransaction` named the vault as the destination — the same value a deposit uses. The verify screen therefore told the approving user the money was going to the place it was coming from.

Nothing was at risk and nothing was misrouted: routing lives in the relayed transaction bytes, and this field is display only. What it affected is what the user was asked to approve.

## Changes

The branch lives in one function, used by both sites:

```kotlin
fun kaminoDestinationAddress(vault: KaminoVault, action: KaminoAction, signerAddress: String) =
    when (action) {
        KaminoAction.DEPOSIT -> vault.address
        KaminoAction.WITHDRAW -> signerAddress
    }
```

- `BuildKaminoKeysignPayloadUseCase` uses it for `toAddress`.
- `KaminoAmountViewModel` uses it for the `DepositTransaction`'s `dstAddress`.

Deliberately not two copies of the same `when`. Those two sites drifting is what produced the bug, and the stored transaction is the one the verify screen reads — branching only the payload would have changed nothing the user sees.

iOS branches the same way, for the same stated reason (`KaminoKeysignPayloadFactory.swift:81`: *"A withdraw pays the user's own account, so that is the destination — not the vault, which is where the funds come from."*).

## Test plan

Five new tests, each verified to be a real regression guard by reverting the fix and confirming it fails:

| Test | Fails when |
|---|---|
| `a withdraw is destined for the signer's own account, not the vault` | the ViewModel site is reverted |
| `a deposit is still destined for the vault` | — (pins the unchanged direction) |
| `the payload carries the branched destination, not the vault for both directions` | the use-case site is reverted |
| `a deposit is destined for the vault` | — |
| `a withdraw is destined for the signer, because the vault is where it comes from` | — |

The new `KaminoDestinationAddressTest` mocks the preparer, so it covers the payload branch without WalletCore or a captured response. That matters here: the captured withdraw fixture carries the `u64::MAX` full-exit sentinel and is refused by the validator by design, so no instrumented test can build a withdraw payload end to end.

`:app:testDebugUnitTest`, `:data:testDebugUnitTest`, `:data:lintDebug` and ktfmt pass.

**Not run:** `:data:connectedDebugAndroidTest` — no emulator available at the time of writing. Nothing in that suite asserts on `toAddress`, so the risk is low, but it has not been run against this change.

## Out of scope

The issue also notes that a Kamino payload carries no memo, so `JoinKeysignViewModel.isDepositPayload` returns false and an Android co-signer renders the generic Send verify screen with no withdraw framing. The destination it shows is now correct, but the framing is still wrong. That is a larger change than the remediation this issue asks for and is left for a follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Kamino transaction destinations based on the action:
    * Deposits are directed to the vault.
    * Withdrawals are directed to the wallet.

* **Tests**
  * Added coverage confirming destination addresses in transactions and signing payloads.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
@@ -24,6 +24,7 @@ import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoWithdrawEligibili
 import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoWithdrawLiquidity
 import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoWithdrawMath
 import com.vultisig.wallet.data.blockchain.solana.kamino.coin
+import com.vultisig.wallet.data.blockchain.solana.kamino.kaminoDestinationAddress
 import com.vultisig.wallet.data.chains.helpers.SolanaHelper
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
@@ -458,7 +459,9 @@ constructor(
                     srcAddress = coin.address,
                     srcTokenValue = TokenValue(value = keysignPayload.toAmount, token = coin),
                     memo = "",
-                    dstAddress = vault.address,
+                    // The verify screen reads this stored copy, not the payload above, so the
+                    // withdraw branch has to be applied here too or nothing the user sees changes.
+                    dstAddress = kaminoDestinationAddress(vault, action, coin.address),
                     estimatedFees = gasFee,
                     estimateFeesFiat = "",
                     // The payload's, not `specific`'s. `KeysignShareViewModel` rebuilds the relayed
--- a/app/src/test/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModelTest.kt
@@ -15,6 +15,7 @@ import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoWithdrawEligibili
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.DepositTransaction
 import com.vultisig.wallet.data.models.OPERATION_KAMINO_DEPOSIT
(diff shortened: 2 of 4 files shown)


johnnyluo (approved):
LGTM

PR #2807, opened 2025-11-17
[Home]: Support DeFi Balances (Home Screen)

JaimeToca:
## Description

- Support Home Screen DeFi Balances
- Preload Defi cache values when user is "wallet"
- When "DeFi" is clicked, show cache values and refresh in the background.
- Add Flow code, so that its easy to plug in another chains

## Which feature is affected?
- [ ] Create vault ( Secure / Fast) - Please ensure you created a Secure vault & fast vault
- [ ] Sending  - Please attach a tx link here
- [ ] Swap - Please attach a tx link for the swap here
- [ ] New Chain / Chain related feature  -  Please attach tx link here

## Checklist

- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

## Release Notes

* **New Features**
  * DeFi balance support now available; users can view DeFi account balances separately from regular wallet accounts.
  * Added support for ThorChain DeFi assets including staking positions.

* **Bug Fixes**
  * Improved token price lookup with enhanced case-insensitive handling.

PR #5615, opened 2026-08-16
fix(defi): total the auto-compounding RUJI position on the DeFi tab

### LINKED ISSUE #5614: DeFi Portfolio reads $0.00 for an auto-compounding RUJI position
## Summary

The DeFi Portfolio row for THORChain, and the portfolio total above it, exclude the auto-compounding RUJI position entirely. A vault whose RUJI is all auto-compounding reads "THORChain / $0.00" on the DeFi tab while the THORChain DeFi screen one tap deeper shows that same position at its correct fiat value. Reported on v1.0.116 and still present on v1.0.117.

## Details

`AccountsRepositoryImpl.loadDeFiAddresses` builds each chain's accounts from `vault.coins` (`data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt:421`), and the chain's DeFi balances are then matched onto those accounts — by ticker on the cached path (`:474`) and by coin id on the network path (`BalanceRepository.getDefiTokenBalanceAndPrice`). A balance with no account to match is discarded, and every unmatched account falls through to the explicit zero-fill beside it.

The auto-compounding position is reported under `Coins.ThorChain.sRUJI`, which can never appear in `vault.coins`: it is held out of the catalogue on purpose (`Coins.kt:4478`, `defiOnly`) and its denom is excluded from token discovery (`TokenRepository.kt:291`, `DEFI_ONLY_THORCHAIN_DENOMS`) so it is never offered as a wallet token. Its value therefore has no account to land on and never reaches the row.

The row's asset count comes from that same list — `assetsSize = from.accounts.size` (`AddressToUiModelMapper.kt:37`) — which is what the reported screen shows: "4 Assets / $0.00", i.e. four THORChain wallet tokens each carrying a zero DeFi balance, with the one funded position absent from the list. The DeFi Portfolio header is the sum of the rows, so it inherits the same hole.

The THORChain DeFi screen is unaffected because it reads `RujiStakingService` directly and prices sRUJI off RUJI (`ThorchainDefiPositionsViewModel.kt:751`) — hence the correct value one level deeper.

### Regression

Introduced by #5423 (`a3b3fd1aa`, 30 Jul), which split RUJI's bonded and auto-compounding positions onto separate coins. Before it, `RujiStakeBalances.stakeAmount` was the on-chain sRUJI receipt balance reported under `Coins.ThorChain.RUJI` — a real wallet token — so it matched an account and produced a value.

### Reproduction

1. Hold an auto-compounding (`liquid.bond`) RUJI position on THORChain, with nothing bonded and no liquid RUJI.
2. Open the DeFi tab. The THORChain row reads $0.00 and so does the DeFi Portfolio header.
3. Tap the row. The Staked tab shows the compounded RUJI position at its correct fiat value.

Reproduced in a unit test against `loadDeFiAddresses`: given no bond, zero bonded RUJI and an sRUJI position worth $250, the THORChain address comes back as `[(RUNE, $0), (RUJI, $0)]` and totals $0 — the position is not in the list at all.

## Impact

Every vault holding the auto-compounding RUJI position has its DeFi Portfolio total understated by exactly that amount, and reads $0.00 outright when it is the only position held. No funds are at risk and no flow is blocked: the position stays visible and redeemable from the screen below.

The general shape is wider than this one coin — any DeFi position keyed to a coin the vault does not also carry as a wallet token is dropped the same way — but sRUJI is the case guaranteed to hit it, since it is kept out of discovery by design.

aminsato:
Closes #5614

## Problem

`loadDeFiAddresses` builds each chain's accounts from `vault.coins` and matches the chain's DeFi balances onto them, so a position whose coin the vault does not carry as a wallet token has nowhere for its balance to land and is discarded. `sRUJI` is exactly that by construction — kept out of the catalogue (`Coins.defiOnly`) and out of token discovery (`DEFI_ONLY_THORCHAIN_DENOMS`) so it is never offered as a wallet token.

An auto-compounding staker therefore read "THORChain / $0.00" over a funded position, and the DeFi Portfolio header, being the sum of the rows, inherited the hole. The THORChain DeFi screen one tap deeper was correct throughout — it reads `RujiStakingService` directly and prices sRUJI off RUJI.

Regressed in #5423, which split RUJI's bonded and auto-compounding positions apart. Before it the compounding amount was reported under `Coins.ThorChain.RUJI` — a real wallet token, so it had an account to match.

## Changes

**`AccountsRepository.loadDeFiAddresses`**
- Injects `Coins.defiOnly` into each chain's token list, stamped with the address and `hexPublicKey` the chain's tokens share. Both balance paths then resolve it unchanged — `getDeFiCachedTokeBalanceAndPrice` was already returning the sRUJI balance, it simply had nowhere to go.
- Prices them in the refresh. A price is cached under its own coin's token id, and that is the row the cached path reads, so leaving them out would keep valuing a resolved position at $0.00.
- Drops the ones that resolve to nothing. An account is injected before any balance is known, so without this every THORChain row's asset count would gain one for a position the vault does not hold.

**`Coins.isDefiOnly(coin)`** — matched on id, since these coins are stamped with the holding vault's address before they reach an account.

**`AccountsLoader`** — the three `loadDeFiAddresses` call sites go through `spendableAccounts()`, which filters those accounts back out. They are not holdings a form can draw on: the receipt is redeemed through `UNSTAKE_SRUJI`, which synthesizes its own account. The token picker is unchanged.

## Test plan

Five new tests, each verified to fail against the code it guards:

`AccountsRepositoryImplTest`
- cached DeFi addresses carry the auto-compounding position, and the chain's total includes it — before the fix the THORChain address came back `[(RUNE, $0), (RUJI, $0)]`, totalling $0 against a $250 position
- a refresh fetches **and** prices it (`coVerify` on the refresh list)
- a vault holding none carries no account for it, so the row's asset count is untouched
- the injected coin carries the chain's address and public key

`AccountsLoaderTest`
- the DeFi form path drops accounts that only ever back a position

`:data` and `:app` unit suites pass (599 tests in `:app`). `ktfmtCheck` clean on both modules.

Not covered by an automated test: the on-device render. Worth a manual pass on a vault holding a compounded RUJI position — the DeFi tab's THORChain row and the header should both show its value, and a vault without one should be unchanged.

## Summary by CodeRabbit

* **New Features**
  * Improved DeFi account and balance handling across supported chains.
  * DeFi positions now receive refreshed, underlying-asset pricing and relevant native-token details.
  * ThorChain auto-compounding positions are reflected in balances and chain totals.

* **Bug Fixes**
  * Unfunded or unavailable DeFi-only assets are automatically excluded.
  * Non-spendable DeFi-only positions no longer appear as sendable accounts.
  * Improved price consistency for cached and refreshed auto-compounding positions.


data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -420,14 +420,21 @@ constructor(
...
+                .groupBy { it.chain }
+                .mapValues { (chain, tokens) -> tokens + defiOnlyTokens(chain, tokens) }
+
+        // Price the DeFi-only positions alongside the vault's own coins: a price is cached under
+        // its own coin's token id, and that is the row the cached path reads, so a refresh that
+        // left them out would keep valuing a resolved position at $0.00.
         val loadPrices =
             if (isRefresh) {
-                async { tokenPriceRepository.refresh(defiCoins) }
+                async { tokenPriceRepository.refresh(coins.values.flatten()) }
```
rkokhatskyi, line 440: The injected sRUJI is priced only from a `sRUJI-THORChain` row that nothing but this `isRefresh`-gated refresh writes. The cold-start DeFi load passes false (`VaultAccountsViewModel.kt:278`), so the cached emission values a funded position at $0.00 (`BalanceRepository.kt:275`) and the ticket's symptom survives

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -496,7 +503,10 @@ constructor(
                             }
                         val canBeDeFiProvider = address.chain.isDeFiSupported
 
-                        address.copy(accounts = updatedAccounts, isDefiProvider = canBeDeFiProvider)
+                        address.copy(
+                            accounts = updatedAccounts.dropUnfundedDefiOnly(),
```
rkokhatskyi, line 532: `balancesByTicker` is keyed by bare ticker across every DeFi chain, so the funded Circle deposit reported as `Coins.Ethereum.USDC` also lands on the Solana or Tron USDC account at `AccountsRepository.kt:481` and is counted once per chain in the DeFi total

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Confirmed, and fixed in 89fdf1c25.

You're right that nothing outside this gated refresh writes `sRUJI-THORChain`. The receipt can never be in `vault.coins`, so unlike every other DeFi coin it also missed the *ungated* refresh — `loadAddressBalances` passes `vaultCoins` and nothing else — which is what keeps the rest of these rows warm across a cold start.

Rather than ungate this one (the cached emission exists so the tab paints without waiting on the network, and awaiting a refresh here would take that away), I let the receipts ride along on that ungated refresh instead:

```kotlin
val loadPrices = async { tokenPriceRepository.refresh(vaultCoins + Coins.defiOnly) }
```

`loadAccounts` runs it on every launch, and sRUJI shares RUJI's `rujira` provider id, so the CoinGecko batch gains no round trip.

One correction on the impact, for the record: the ticket's symptom itself doesn't survive — the position now reaches the row with its token amount, and the account is no longer absent. What survived was its *fiat* reading $0.00 on the cached emission, until the refresh landed. And the DeFi tab is only ever entered through `setCryptoConnectionType` → `loadDeFiBalances(vaultId, true)`, so that refresh is always seconds behind. Still worth closing, since the cached emission is what paints first and the row persists once written.

Covered by `the wallet price refresh keeps the DeFi-only receipts priced`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Verified and fixed in 89fdf1c25.

`isValidForDeFi()` is chain-level, so every vault token on Ethereum, Solana and Tron gets a DeFi account, and `getDeFiCachedTokeBalanceAndPrice` drops the coin on the way out — the balances arrive carrying a ticker and nothing else. Pooled into one map, the funded Circle deposit reported as `Coins.Ethereum.USDC` matched all three USDC accounts, and the portfolio total counted the one deposit once per chain.

The lookup is now built from the address's own chain:

```kotlin
val balancesByTicker =
    cacheBalancesByChain[address.chain]
        ?.associateBy { it.tokenBalance.tokenValue?.unit?.lowercase() }
        .orEmpty()
```

Accounts with no match still zero-fill exactly as before, and the emit guard is the same condition spread over the per-chain lists.

For the record this one predates the PR — `balancesByTicker` came in with #2807 — but it's on the block this change touches and the fix is contained, so it's here rather than in a follow-up. Covered by `a cached DeFi balance stays on the chain that reported it`, which totals `DEPOSIT_FIAT` once instead of twice.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -650,6 +678,40 @@ constructor(
...
+     * address the balance is fetched against. A chain whose native token is absent is skipped: the
+     * mapper builds no address for it either, so the accounts would have nothing to hang from.
+     */
+    private fun defiOnlyTokens(chain: Chain, tokens: List<Coin>): List<Coin> {
+        val nativeToken = tokens.firstOrNull { it.isNativeToken } ?: return emptyList()
+        return Coins.defiOnly
+            .filter { position ->
+                position.chain == chain && tokens.none { it.id.equals(position.id, true) }
+            }
+            .map { it.copy(address = nativeToken.address, hexPublicKey = nativeToken.hexPublicKey) }
```
rkokhatskyi, line 711: The compounding leg is valued off `sRUJI-THORChain` while the bonded leg beside it and `ThorchainDefiPositionsViewModel.kt:751` both use RUJI's row, which `fetchThorPoolPrices` overwrites with the THOR.RUJI pool price. Live now that is 0.17177730 against CoinGecko's 0.171154 for the same RUJI

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -96,7 +96,15 @@ constructor(
...
                 supervisorScope {
-                    val loadPrices = async { tokenPriceRepository.refresh(vaultCoins) }
+                    // The DeFi-only receipts ride along: no vault carries one, so this refresh —
+                    // the only ungated one — is all that keeps their price rows warm.
+                    // loadDeFiAddresses refreshes them too, but only on an explicit reload, and
+                    // its cached emission reads those rows, so a cold start would otherwise value
+                    // a funded position at $0.00. They price off the token they mirror and share
+                    // its provider id, so the lookup below gains no round trip.
+                    val loadPrices = async {
+                        tokenPriceRepository.refresh(vaultCoins + Coins.defiOnly)
```
rkokhatskyi: Appending `Coins.defiOnly` here defeats the `thorTokens.isEmpty()` bail at `TokenPriceRepository.kt:510`, so a vault with no THORChain token now holds its `isComplete` emission behind a thornode /thorchain/pools fetch that can only miss for sRUJI

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
```diff
@@ -496,7 +518,10 @@ constructor(
                             }
                         val canBeDeFiProvider = address.chain.isDeFiSupported
 
-                        address.copy(accounts = updatedAccounts, isDefiProvider = canBeDeFiProvider)
+                        address.copy(
+                            accounts = updatedAccounts.dropUnfundedDefiOnly(),
```
rkokhatskyi, line 532: An unmatched account in this cached emission is stamped `AppCurrency.USD.ticker` at `AccountsRepository.kt:513` while the network emission uses the app currency, and `calculateAccountsPartialFiatValue` keeps the last element's currency, so a EUR total renders with a dollar sign

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Confirmed, and fixed in b5a6c3543.

Traced it through: sRUJI is not a NAV-priced denom, so it goes into the provider batch on its borrowed `rujira` id and its row gets CoinGecko's quote. `fetchThorPoolPrices` then runs *after* that batch and overwrites RUJI's row with the pool price — and it can't do the same for the receipt, because `mapThorPoolAsset("x/staking-x/ruji")` yields `staking.x/ruji` and the ticker fallback yields `thor.sruji`, neither of which is a pool. So the two rows were always going to disagree, and both legs are reported in RUJI (`rujiPositionUiModel` prices both off `Coins.ThorChain.RUJI`).

The fix is to leave the receipt no row of its own rather than to keep two in sync:

```kotlin
private val priceRowAliases: Map<String, Coin> =
    mapOf(ThorChain.sRUJI.id.lowercase() to ThorChain.RUJI)

fun pricedAs(coin: Coin): Coin = priceRowAliases[coin.id.lowercase()] ?: coin
```

Applied inside `TokenPriceRepositoryImpl` at the three places a row is read or written — `getPrice`, `getCachedPrice`, `refresh` — so the cached DeFi emission, the network emission, the position card and the unstake form all land on the same number, and a new reader can't drift off on its own. It sits on the catalogue next to `isNavPricedDenom` for the reason that one gives: the pricing repository and everything valuing a position have to agree on which rows exist.

Covered by `the auto-compounding receipt is priced off RUJI's row, not one of its own` (pool 0.17177730 wins over CoinGecko 0.171154, and nothing is written under the receipt's id) and `a cached price read for the receipt is served from RUJI's row`, which stubs a stale sRUJI row in Room and asserts it is not preferred.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Verified and fixed in b5a6c3543.

You're right — `Coins.defiOnly` is a THORChain non-native token, so appending it unconditionally made `thorTokens` non-empty for every vault, and `loadPrices.await()` gates the terminal emission. A Bitcoin-only vault paid for a `/thorchain/pools` fetch that, as you say, could only miss for sRUJI.

The receipts now ride along only where one can actually be injected:

```kotlin
private fun defiOnlyReceiptsFor(vaultCoins: List<Coin>): List<Coin> =
    Coins.defiOnly.filter { receipt -> vaultCoins.any { it.chain == receipt.chain } }
```

`defiOnlyTokens` hangs a receipt off the chain's native token, so a vault without the chain can never carry the position and its row is one nothing reads. That restores the `thorTokens.isEmpty()` bail for those vaults.

One residual cost, so it isn't a surprise: a vault that holds THORChain but only RUNE now does trigger a pools fetch it previously skipped — after the pricing fix above, the pool row *is* where the position's price comes from, so it's a fetch that resolves a price rather than one that can only miss. Narrowing it further would mean gating on cached staking details, which pulls a new dependency into this repository and would leave a first-ever position unpriced until the DeFi refresh; say the word if you'd rather have that trade.

Covered by `a vault without the receipt's chain does not pay to price it`, which asserts no DeFi-only coin reaches the refresh for a Solana-only vault, alongside the existing THORChain case.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
aminsato: Verified and fixed in b5a6c3543.

`calculateAccountsPartialFiatValue` seeds on the first resolved value but keeps `fiatValue.currency` through the fold, so the label is the *last* summed account's — one USD-stamped zero at the end of the flattened list renames a EUR total. The zero-fill now takes the currency the resolved balances came back in:

```kotlin
val cachedCurrency =
    cacheBalancesByChain.values.asSequence().flatten().firstNotNullOfOrNull {
        it.tokenBalance.fiatValue?.currency
    } ?: AppCurrency.USD.ticker
```

Read off this emission's own balances rather than re-resolving the app currency, so the label can't disagree with the numbers beside it when the currency changes mid-load — `getDeFiCachedTokeBalanceAndPrice` captured it once for the whole batch.

For the record the hardcoded USD predates this PR, but the per-chain keying in 89fdf1c25 is what made unmatched accounts common (that was the point — the Solana and Tron USDC accounts stop borrowing Ethereum's deposit and now zero-fill), so it belongs in this branch rather than a follow-up.

Covered by `an unmatched cached DeFi account keeps the currency the balances resolved in`: one funded EUR deposit on Ethereum, and the zero-filled ETH account beside it must not come back USD.

rkokhatskyi (approved):
LGTM

PR #5619, opened 2026-08-16
fix(defi): total both Solana DeFi positions in the chain header

### LINKED ISSUE #5602: Kamino Earn positions are missing from the Solana DeFi header total
## Summary

The Solana DeFi screen's chain header sums native staking only, so a vault with a Kamino Earn position sees "Solana / $0.00" directly above a card showing that position's real value. Introduced by #5588, which added the Earn tab without extending the header above it.

## Details

`app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/SolanaStakingPositionsScreen.kt:143-147` passes `state.totalStakedFiatDisplay` to `SolanaHeaderBanner`, and that value folds stake accounts only (`SolanaStakingPositionsViewModel.kt:330-336`). `kaminoState` is already collected in the same composable at line 75 and used 34 lines below, so both figures are in scope. The banner is chain-labelled and sits outside the tab branch, so it presents as the chain's DeFi total rather than a staking-specific one, and `selectedTab` defaults to `DeFiTab.EARN`, so the mismatch is on the tab the user lands on. Kamino share mints are also excluded from SPL discovery (`SPLTokenRepository.kt:56-62`) on the premise that the Earn tab values them, so a deposit reduces the wallet total with no offsetting increase on this surface.

## Impact

Every vault holding a Kamino position has its Solana DeFi header understated by exactly that amount; the zero-stake case just shows it most sharply. No funds at risk and no flow blocked.

## Suggested remediation

Add the Kamino total to the header value. iOS sums both (`DefiBalanceService.swift:38-43`).

aminsato:
Fixes #5602

## Problem

The Solana DeFi banner is labelled with the chain and sits above the tab row, but it was fed
`state.totalStakedFiatDisplay` — the native-staking half on its own. A vault holding a Kamino Earn
position saw a figure short by exactly that amount, and a vault with no stake accounts read
"Solana / $0.00" directly above a funded card. `selectedTab` defaults to `EARN`, so the mismatch is
on the tab the screen opens on.

Both figures were already in scope: `kaminoState` is collected in the same composable and used a few
lines below.

## Changes

- `KaminoEarnUiModel.totalFiatValue` — the sum the Earn card already shows, kept unformatted so the
  header can add it. Zero when no vault is enabled, which is a read that succeeded and found
  nothing; null only while genuinely unresolved.
- The screen hands that to `SolanaStakingPositionsViewModel` as it resolves, and the view-model
  folds the two halves into one `chainTotalFiatDisplay`. The sum happens there rather than in the
  composition because that is where the user's currency format lives.
- Unknown wins over known: while either half is unresolved the banner shows the unavailable marker
  rather than the other half, since a number short by a real position reads as the chain's whole
  DeFi balance. The staking-only case still shows its figure, because "no Kamino vault enabled"
  resolves to zero rather than unknown.
- The banner's loading state waits on both loads, so it cannot print the staking total and then jump.

Neither tab loses anything: the Staked tab still shows Total Staked SOL and the Earn tab still shows
the Kamino total. The banner was the only thing claiming to be the chain's figure.

iOS sums the same two in `DefiBalanceService.totalBalanceInFiat`.

## Test plan

Seven tests, five of them in a new `SolanaStakingPositionsViewModelTest`:

- the header adds Earn to native staking
- either half may arrive first (Earn resolves before the stake accounts)
- no Kamino vault enabled still leaves the header showing the staked total
- an unresolved Earn total leaves the header unavailable rather than short
- a vault without SOL reports no total at all
- `KaminoEarnViewModelTest`: nothing enabled hands the header a resolved zero, not an unknown; the
  raw total matches the formatted one the tab renders

Checked on a vault holding both: the banner reads $75.91 against $75.36 staked and $0.55 in Earn,
and it stays put when switching between the two tabs.

## Summary by CodeRabbit

* **New Features**
  * Solana staking screens now display a combined total for native staking and enabled Kamino Earn positions.
  * Combined totals update when the selected currency or Earn vaults change.
  * Empty Kamino Earn configurations correctly contribute a zero value.

* **Bug Fixes**
  * Unavailable balances and prices are no longer reported as zero.
  * Totals remain loading until required staking, Earn, and currency data is resolved.
  * Currency formatting remains accurate during currency changes and partial refreshes.
  * Correctly handles accounts without SOL and failed data loads.

* **Tests**
  * Added coverage for totals, currency changes, loading states, zero balances, and failed vault loads.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
@@ -29,6 +29,7 @@ import com.vultisig.wallet.data.utils.safeLaunch
 import com.vultisig.wallet.ui.navigation.Destination
 import com.vultisig.wallet.ui.navigation.Navigator
 import com.vultisig.wallet.ui.navigation.Route
+import com.vultisig.wallet.ui.screens.v2.defi.solana.KaminoEarnTotal
 import com.vultisig.wallet.ui.utils.UiText
 import com.vultisig.wallet.ui.utils.asUiText
 import com.vultisig.wallet.ui.utils.formatPercent
@@ -44,6 +45,7 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.update
 import timber.log.Timber
@@ -108,10 +110,12 @@ internal data class SolanaStakingPositionsUiState(
     val chainTotalFiatDisplay: String? = null,
     /**
      * The two halves [chainTotalFiatDisplay] is made of, kept raw because they are loaded by
-     * separate view-models and either may arrive first.
+     * separate view-models and either may arrive first. The Kamino half is stored exactly as it was
+     * handed over, currency tag included, so the screen can tell "Earn's figure has not reached
+     * this view-model yet" from "Earn has no figure".
      */
     val stakedFiatValue: BigDecimal? = null,
-    val kaminoFiatValue: BigDecimal? = null,
+    val kaminoTotal: KaminoEarnTotal? = null,
     val positions: List<SolanaStakePositionRow> = emptyList(),
     val error: UiText? = null,
 )
@@ -144,13 +148,52 @@ constructor(
 
     private var vaultId: VaultId = ""
     private var loadJob: Job? = null
+    private var currencyJob: Job? = null
     private var solCoin: Coin? = null
     private var accountsByPubkey: Map<String, SolanaStakeAccount> = emptyMap()
     private var isBuildingStakingTx: Boolean = false
 
+    /**
+     * The currency the figures on screen were priced in, together with its format. Held as one
+     * reference so a reader never sees a format from one currency next to the ticker of another,
+     * and null until the first load has priced anything.
+     */
+    @Volatile private var pricing: Pricing? = null
+
+    private data class Pricing(val currency: AppCurrency, val format: NumberFormat)
+
     fun setData(vaultId: VaultId) {
         this.vaultId = vaultId
-        loadData()
+        observeCurrency()
+    }
+
+    /**
+     * Loads on entry and again on every display-currency change: every figure here is priced, so a
+     * switch invalidates them rather than merely relabelling them. The collector's first emission
+     * performs the initial load.
+     */
+    private fun observeCurrency() {
+        currencyJob?.cancel()
+        currencyJob =
+            viewModelScope.safeLaunch(
+                onError = { Timber.e(it, "Failed to watch the app currency") }
+            ) {
+                appCurrencyRepository.currency.distinctUntilChanged().collect { currency ->
+                    if (pricing?.currency?.equals(currency) == false) {
+                        // Drop the old-currency figures instead of leaving them under a new
+                        // symbol while the reload runs.
+                        pricing = null
+                        _state.update {
+                            it.copy(
+                                isLoading = true,
+                                totalStakedFiatDisplay = null,
+                                chainTotalFiatDisplay = null,
+                            )
+                        }
+                    }
+                    loadData(currency)
+                }
+            }
     }
 
     fun refresh() {
@@ -166,18 +209,12 @@ constructor(
      *
      * Null is "not resolved", which is not zero: reporting the staking half alone would present a
      * number that is short by a real position as the chain's whole DeFi balance.
+     *
+     * Applied synchronously on purpose. Formatting asynchronously here let two handovers finish out
+     * of order, so an earlier total could land last and sit on the banner as the current one.
      */
-    fun onKaminoTotalChanged(total: BigDecimal?) {
-        viewModelScope.safeLaunch(onError = { Timber.w(it, "Failed to total the Solana header") }) {
-            val currencyFormat = appCurrencyRepository.getCurrencyFormat()
-            _state.update {
-                it.withChainTotal(
-                    staked = it.stakedFiatValue,
-                    kamino = total,
-                    format = currencyFormat,
-                )
-            }
-        }
+    fun onKaminoTotalChanged(total: KaminoEarnTotal?) {
+        _state.update { it.withChainTotal(staked = it.stakedFiatValue, kamino = total) }
     }
 
     fun onStake() {
@@ -321,7 +358,7 @@ constructor(
         }
     }
 
-    private fun loadData() {
+    private fun loadData(selectedCurrency: AppCurrency? = null) {
         loadJob?.cancel()
         loadJob =
             viewModelScope.safeLaunch(
@@ -349,8 +386,9 @@ constructor(
                 }
 
                 val isBalanceVisible = balanceVisibilityRepository.getVisibility(vaultId)
-                val currency = appCurrencyRepository.currency.first()
+                val currency = selectedCurrency ?: appCurrencyRepository.currency.first()
                 val currencyFormat = appCurrencyRepository.getCurrencyFormat()
+                pricing = Pricing(currency = currency, format = currencyFormat)
                 val price = cachedPrice(solCoin.id, currency)
 
                 this@SolanaStakingPositionsViewModel.solCoin = solCoin
@@ -384,11 +422,7 @@ constructor(
                             positions = rows,
                             error = null,
                         )
-                        .withChainTotal(
-                            staked = stakedFiatValue,
-                            kamino = it.kaminoFiatValue,
-                            format = currencyFormat,
-                        )
+                        .withChainTotal(staked = stakedFiatValue, kamino = it.kaminoTotal)
                 }
             }
     }
@@ -460,18 +494,25 @@ constructor(
      * printing the other half, which would read as the chain total while being short by whatever
      * the unread side holds. A user with no Kamino vault enabled has a resolved zero there, so the
      * common staking-only case still shows a figure.
+     *
+     * A Kamino half priced in some other currency counts as unresolved too. The two view-models
+     * read the selected currency independently, so a mid-session switch reaches one before the
+     * other; the banner waits that window out rather than adding euros to dollars.
      */
     private fun SolanaStakingPositionsUiState.withChainTotal(
         staked: BigDecimal?,
-        kamino: BigDecimal?,
-        format: NumberFormat,
-    ): SolanaStakingPositionsUiState =
-        copy(
+        kamino: KaminoEarnTotal?,
+    ): SolanaStakingPositionsUiState {
+        val pricing = pricing
+        val kaminoValue = kamino?.takeIf { it.currency == pricing?.currency }?.value
+        return copy(
             stakedFiatValue = staked,
-            kaminoFiatValue = kamino,
+            kaminoTotal = kamino,
             chainTotalFiatDisplay =
-                if (staked == null || kamino == null) null else format.format(staked.add(kamino)),
+                if (staked == null || kaminoValue == null || pricing == null) null
+                else pricing.format.format(staked.add(kaminoValue)),
         )
+    }
 
     private suspend fun cachedPrice(tokenId: String, currency: AppCurrency): BigDecimal =
         tokenPriceRepository.getCachedPrice(tokenId = tokenId, appCurrency = currency)
--- a/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
@@ -100,6 +100,18 @@ internal data class SolanaStakingPositionsUiState(
     // string would slip past that and draw a blank line where the price belongs.
     val totalStakedFiatDisplay: String? = null,
     val totalStakedSolDisplay: String = "",
+    /**
+     * The chain's whole DeFi holding — native staking plus Kamino Earn — for the header banner,
+     * which is labelled with the chain rather than with the selected tab. Null renders as the
+     * unavailable marker, same as [totalStakedFiatDisplay].
+     */
+    val chainTotalFiatDisplay: String? = null,
+    /**
+     * The two halves [chainTotalFiatDisplay] is made of, kept raw because they are loaded by
+     * separate view-models and either may arrive first.
+     */
+    val stakedFiatValue: BigDecimal? = null,
+    val kaminoFiatValue: BigDecimal? = null,
     val positions: List<SolanaStakePositionRow> = emptyList(),
     val error: UiText? = null,
 )
@@ -145,6 +157,27 @@ constructor(
         if (vaultId.isEmpty()) return
         _state.update { it.copy(isReloading = true) }
         loadData()
+    }
+
+    /**
+     * The Kamino Earn total for this vault, handed over by the screen. Earn is loaded by its own
+     * view-model, so the header banner — which is labelled with the chain, not with the selected
+     * tab — can only be their sum once both sides have reported.
+     *
+     * Null is "not resolved", which is not zero: reporting the staking half alone would present a
+     * number that is short by a real position as the chain's whole DeFi balance.
+     */
+    fun onKaminoTotalChanged(total: BigDecimal?) {
+        viewModelScope.safeLaunch(onError = { Timber.w(it, "Failed to total the Solana header") }) {
+            val currencyFormat = appCurrencyRepository.getCurrencyFormat()
+            _state.update {
+                it.withChainTotal(
+                    staked = it.stakedFiatValue,
+                    kamino = total,
+                    format = currencyFormat,
+                )
+            }
+        }
     }
 
     fun onStake() {
@@ -335,19 +368,27 @@ constructor(
                             acc + account.delegatedStake.toBigDecimal()
                         }
                         .movePointLeft(solCoin.decimal)
-                val totalFiat = currencyFormat.format(totalStakedSolAmount.multiply(price))
+                val stakedFiatValue = totalStakedSolAmount.multiply(price)
+                val totalFiat = currencyFormat.format(stakedFiatValue)
 
                 _state.update {
                     it.copy(
-                        isLoading = false,
-                        isReloading = false,
-                        isBalanceVisible = isBalanceVisible,
-                        totalStakedFiatDisplay = totalFiat,
-                        totalStakedSolDisplay =
-                            totalStakedSolAmount.stripTrailingZeros().formatTokenAmount(SOL_TICKER),
-                        positions = rows,
-                        error = null,
-                    )
+                            isLoading = false,
+                            isReloading = false,
+                            isBalanceVisible = isBalanceVisible,
+                            totalStakedFiatDisplay = totalFiat,
+                            totalStakedSolDisplay =
+                                totalStakedSolAmount
+                                    .stripTrailingZeros()
+                                    .formatTokenAmount(SOL_TICKER),
+                            positions = rows,
+                            error = null,
+                        )
+                        .withChainTotal(
+                            staked = stakedFiatValue,
+                            kamino = it.kaminoFiatValue,
+                            format = currencyFormat,
+                        )
                 }
             }
     }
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
```diff
@@ -147,6 +159,27 @@ constructor(
...
 
+    /**
+     * The Kamino Earn total for this vault, handed over by the screen. Earn is loaded by its own
+     * view-model, so the header banner — which is labelled with the chain, not with the selected
+     * tab — can only be their sum once both sides have reported.
+     *
+     * Null is "not resolved", which is not zero: reporting the staking half alone would present a
+     * number that is short by a real position as the chain's whole DeFi balance.
+     */
+    fun onKaminoTotalChanged(total: BigDecimal?) {
```
rkokhatskyi: `onKaminoTotalChanged` launches an uncancelled coroutine per call and closes over its `total` argument, so two calls can finish their `getCurrencyFormat` suspensions out of order, letting an earlier stale value permanently overwrite the fresher total in `chainTotalFiatDisplay`

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/SolanaStakingPositionsScreen.kt
```diff
@@ -141,8 +149,10 @@ internal fun SolanaStakingPositionsContent(
             horizontalAlignment = CenterHorizontally,
         ) {
             SolanaHeaderBanner(
-                totalValue = state.totalStakedFiatDisplay,
-                isLoading = state.isLoading,
+                totalValue = state.chainTotalFiatDisplay,
+                // Both halves gate the banner: showing the staking total the moment it lands would
+                // print a figure that then jumps as Earn resolves.
+                isLoading = state.isLoading || kaminoState.isLoading,
```
rkokhatskyi: The `isLoading` gate reads `kaminoState.isLoading` directly and flips false the instant Kamino's own load settles, while `chainTotalFiatDisplay` only updates once `onKaminoTotalChanged`'s separate coroutine finishes an async `getCurrencyFormat` lookup, so the banner can show the unavailable dash for a frame even though both totals are already known

rkokhatskyi: This OR-gate couples the banner to `kaminoState.isLoading`, which `KaminoEarnViewModel.load()` resets true on every refresh and vault-toggle reload, not just first load, so pulling to refresh replaces the already-correct `chainTotalFiatDisplay` with a skeleton even though the value never changed

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
```diff
@@ -266,10 +270,8 @@ constructor(
                                 current.rows.filter { row -> row.vaultAddress in enabled }
                             },
                         totalFiat =
-                            resolved
-                                .takeIf { it.isNotEmpty() }
-                                ?.let { currencyFormat.format(totalFiatValue(it)) }
-                                ?: current.totalFiat,
+                            resolvedTotal?.let { currencyFormat.format(it) } ?: current.totalFiat,
+                        totalFiatValue = resolvedTotal ?: current.totalFiatValue,
```
rkokhatskyi: `fiatValue` collapses a resolved-but-unpriced or unread position into `BigDecimal.ZERO` instead of null, while `depositedFiat` for the same row correctly stays null and shows unavailable. `totalFiatValue` sums `fiatValue` regardless, so a transient positions-read or price-lookup failure silently drops that vault from the new chain header total

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
```diff
@@ -412,6 +453,26 @@ constructor(
...
 
+    /**
+     * Folds the two halves of the chain's DeFi holding into the banner value.
+     *
+     * Unknown wins over known: while either half is unresolved the banner says so rather than
+     * printing the other half, which would read as the chain total while being short by whatever
+     * the unread side holds. A user with no Kamino vault enabled has a resolved zero there, so the
+     * common staking-only case still shows a figure.
+     */
+    private fun SolanaStakingPositionsUiState.withChainTotal(
```
rkokhatskyi, line 520: `withChainTotal` adds `stakedFiatValue` and `kaminoFiatValue` as raw BigDecimals with no shared currency check, and each is priced by an independently-timed currency read in its own view-model. After a mid-session currency change lands in one half before the other, the header sums two different currencies as one figure

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
aminsato: Verified and fixed in `37b10773c`.

Real: the call site is a `LaunchedEffect` keyed on the Earn total, but the coroutine it starts lives on `viewModelScope` and is never cancelled, so two handovers could both be parked on `getCurrencyFormat()` and resume in either order. The earlier one resuming last wrote its stale `total` and stayed there until something else moved.

The suspension is gone rather than serialised. The view-model now caches `Pricing(currency, format)` from the load that priced everything on screen, so the handover is a synchronous `_state.update` — there is no window to finish out of order in. There is a test for it: `the newest Kamino total wins, whatever order the handovers land in`.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/SolanaStakingPositionsScreen.kt
aminsato: Verified and fixed in `37b10773c`.

Real, and it was the same frame both ways round: `kaminoState.isLoading` goes false during composition, while `onKaminoTotalChanged` only runs in the `LaunchedEffect` after that frame is composed — so the banner painted the unavailable dash with both halves already known.

The gate now also treats "Earn has a total that has not reached the staking view-model yet" as loading, by comparing what Earn is holding against what was handed over:

```kotlin
isLoading =
    state.chainTotalFiatDisplay == null &&
        (state.isLoading ||
            kaminoState.isLoading ||
            kaminoState.totalValue != state.kaminoTotal)
```

That is why the staking state keeps the handed-over half verbatim, currency tag included — so "not arrived yet" is distinguishable from "arrived, and is nothing".

aminsato: Verified and fixed in `37b10773c`.

Real — `load()` sets `isLoading = true` on every entry, and it is re-entered on pull-to-refresh, on a vault toggle and now on a currency change, so an already-correct banner dropped back to a skeleton each time.

The gate is now conditioned on the total never having resolved, so it only covers the first load:

```kotlin
isLoading =
    state.chainTotalFiatDisplay == null && (state.isLoading || kaminoState.isLoading || ...)
```

Once a figure is up it stays up through a refresh, which is the same rule the Earn tab already applies to its own card (`state.isLoading && state.totalFiat == null`) and to its rows.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
aminsato: Verified and fixed in `37b10773c`. This was the worst of the five — the zero was survivable while it only fed the Earn card, and this PR promoted it into the chain total.

`KaminoEarnRow.fiatValue` is now `BigDecimal?` and carries the same claim `depositedFiat` already did, so an unread or unpriced position is unknown rather than nothing:

```kotlin
fiatValue = tokenAmount?.let { amount -> coin?.let { fiatValueOrNull(amount, it, currency) } }
```

The sum follows it — `totalFiatValue` returns null as soon as one row is unpriced, and a total is only published when every *enabled* vault produced a row, so a vault that dropped out of the round cannot silently shrink the figure either. A previous total survives a failed refresh only while it still covers the same selection and the same currency, tracked in `totalCoverage`; otherwise it is dropped before the reload rather than carried as a fallback, which also closes the path where the load throws before reaching the update.

Covered by `an unpriced vault leaves the total unknown rather than counting it as zero` and `a total from an earlier selection is dropped rather than reused for a new one`.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
aminsato: Verified and fixed in `37b10773c`.

Real, and checking it turned up a wider version of the same thing: neither view-model was watching the currency at all, so after a switch in settings both tabs kept their old figures under the new symbol until the screen was left and re-entered. The sum was just the loudest symptom.

Two parts:

- A half now travels as `KaminoEarnTotal(value, currency)`, and `withChainTotal` treats a half priced in another currency as unresolved. Mid-switch the banner waits instead of adding euros to dollars.
- Both view-models reload on a currency change — the staking one collects `appCurrencyRepository.currency` (its first emission is what performs the initial load), the Earn one folds it into the existing selection collector via `combine`. So the figures are re-priced rather than relabelled, and the unresolved window closes on its own.

The emitted currency is threaded into the load rather than re-read inside it, so the load prices with the currency that triggered it.

`a currency change re-prices the screen instead of relabelling it` walks the whole sequence: $110 → unresolved the moment USD becomes EUR while Earn still holds a dollar figure → €100 once Earn re-reports.

app/src/test/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModelTest.kt
```diff
@@ -0,0 +1,197 @@
...
+import com.vultisig.wallet.ui.navigation.Navigator
+import io.mockk.coEvery
+import io.mockk.every
+import io.mockk.mockk
+import java.math.BigDecimal
+import java.math.BigInteger
+import java.text.NumberFormat
+import java.util.Locale
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
```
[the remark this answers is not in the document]
aminsato: Applied in `37b10773c` — `assertEquals` → `shouldBe`, `assertNull` → `shouldBeNull`, Jupiter annotations kept.

`KaminoEarnViewModelTest` is left on `kotlin.test`: it is pre-existing and entirely in that style, and converting only the assertions this PR touches would leave one file written two ways.

aminsato:
@coderabbitai on the outside-diff finding for `KaminoEarnViewModel.kt` 260-274 (no thread to reply in) — verified and fixed in `37b10773c`.

Both halves were right:

- A partial sum is no longer published. A total is only produced once every *enabled* vault contributed a row, and — separately — only when every row is priced, since `KaminoEarnRow.fiatValue` is now nullable instead of collapsing an unread or unpriced position to zero.
- A previous total is retained only while it still covers the same enabled set and the same currency, tracked in `totalCoverage`. A stale one is dropped at the start of the load rather than used as a fallback at the end, which also covers a load that throws before reaching the state update.

Tests: `an unpriced vault leaves the total unknown rather than counting it as zero` and `a total from an earlier selection is dropped rather than reused for a new one`.

One part I did not test as written: a row *failing to build*. Every call inside `buildRow` is individually `runCatching`-wrapped, so the outer `runCatching` has nothing left to catch and that path is unreachable. The `resolved.size == vaults.size` guard stays as defence; the reachable form of the same bug — a vault that builds but cannot be priced — is what the first test drives.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
```diff
@@ -412,6 +487,33 @@ constructor(
...
+     * A Kamino half priced in some other currency counts as unresolved too. The two view-models
+     * read the selected currency independently, so a mid-session switch reaches one before the
+     * other; the banner waits that window out rather than adding euros to dollars.
+     */
+    private fun SolanaStakingPositionsUiState.withChainTotal(
+        staked: BigDecimal?,
+        kamino: KaminoEarnTotal?,
+    ): SolanaStakingPositionsUiState {
+        val pricing = pricing
+        val kaminoValue = kamino?.takeIf { it.currency == pricing?.currency }?.value
```
rkokhatskyi, line 526: `kaminoValue` is checked against `pricing.currency` here but `staked` never is, so a currency switch can combine a stale, old-currency `stakedFiatValue` with a fresh-currency Kamino total. Kamino's zero-vault path needs no network call, so it can win that race and relabel the old total under the new symbol

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
```diff
@@ -132,13 +148,52 @@ constructor(
...
+                onError = { Timber.e(it, "Failed to watch the app currency") }
+            ) {
+                appCurrencyRepository.currency.distinctUntilChanged().collect { currency ->
+                    if (pricing?.currency?.equals(currency) == false) {
+                        // Drop the old-currency figures instead of leaving them under a new
+                        // symbol while the reload runs.
+                        pricing = null
+                        _state.update {
+                            it.copy(
+                                isLoading = true,
```
rkokhatskyi, line 192: This reset nulls `totalStakedFiatDisplay` and `chainTotalFiatDisplay` but leaves `state.positions` untouched, so every stake-account card keeps rendering its old-currency `stakedFiatDisplay` string for the whole reload, and permanently if that reload then errors since `onError` never touches `positions`. `KaminoEarnViewModel`'s currency reset has the same gap for `rows`

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
```diff
@@ -300,6 +338,7 @@ constructor(
         walletAddress: String,
         position: KaminoUserPositionJson?,
         positionsAreKnown: Boolean,
+        currency: AppCurrency,
```
rkokhatskyi, line 369: The new `currency` param threads through this call, but two lines below, a present position whose `totalShares` fails to parse still collapses `shares` to `BigDecimal.ZERO` instead of null. That reports a genuine deposit as a resolved zero and hides Withdraw, while `KaminoWithdraw.kt` treats the same field failing to parse as a hard failure

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
aminsato: Verified and fixed in `6615b7835`.

Real, and reachable exactly as you describe. `pricing` was published before `fetchStakeAccounts`, so for the whole of a reload after a switch it already said EUR while `stakedFiatValue` was still the dollar figure — and Kamino's no-vault-enabled path resolves without a network call, so it can hand its euro half over inside that window. Both sides then passed the one check there was, and the old staked number went onto the banner under the new symbol.

Two changes:

- A half is now `DefiFiatTotal(value, currency)` on **both** sides, not just Kamino's, and `withChainTotal` holds each of them to the currency the screen is reading. The type outgrew its old name (`KaminoEarnTotal`) once staking used it too, so it moved to `ui/screens/v2/defi/DefiFiatTotal.kt`.
- `pricing` is published with the figures it priced rather than ahead of them, which is what its own KDoc claimed it was. A load still in flight no longer claims a currency.

`a Kamino half that re-priced first is not added to a staked half that has not` drives it: the reload is parked on `fetchStakeAccounts` while Earn hands over a euro half, the banner waits, and it resolves to €100 once the stake accounts land.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/SolanaStakingPositionsViewModel.kt
aminsato: Verified and fixed in `6615b7835`. Both halves of it, including the Kamino one you pointed at.

The reset dropped the two totals and left every card's own fiat exactly as it was, so each row wore the new symbol over an old-currency number for the whole reload — and permanently when that reload failed, since `onError` never rebuilds a row on either view-model, and Kamino additionally keeps its previous rows when every vault fails.

- `SolanaStakePositionRow.stakedFiatDisplay` is nullable now and cleared alongside the totals; the screen renders that as the unavailable dash, same as the banner does.
- `KaminoEarnViewModel` clears `depositedFiat` and `fiatValue` on the same trigger. It needed its own marker (`pricedCurrency`) rather than `totalCoverage`: the rows outlive a load that resolves no total, so the total's currency does not tell you what the cards are priced in.

What each position holds — the staked SOL, the deposited token amount — stays on screen. That is priced in neither currency.

Two tests, one per view-model, both named `a currency change clears the fiat on each card, and a failed reload leaves it clear`: they switch currency with the reload rigged to fail, then assert the fiat is gone and the amount is not.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
aminsato: Verified and fixed in `6615b7835`.

Real, and the same zero-means-two-things bug as the `fiatValue` one from the last round, one layer earlier: `decimalOrNull(position?.totalShares) ?: BigDecimal.ZERO` folded "the wallet has no position in this vault" together with "the position is there and its size would not parse". Only the first is a zero. The second reported a real deposit as nothing, took Withdraw away with it via `hasPosition`, and — now that this feeds the chain header — shrank the whole chain's figure without saying so. `KaminoWithdraw` refuses to build against the same unparsed field, so the two disagreed about the same string.

```kotlin
val shares =
    when {
        !positionsAreKnown -> null
        position == null -> BigDecimal.ZERO
        else -> KaminoPositionMath.decimalOrNull(position.totalShares)
    }
```

An unparseable balance is now null all the way out: no deposited amount, no fiat, no total, and `hasPosition` stays true so the way out of the position is not removed on the strength of a read that did not succeed.

`a position whose share balance cannot be read is unknown rather than zero`.

app/src/test/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModelTest.kt
```diff
@@ -127,6 +129,87 @@ internal class KaminoEarnViewModelTest {
...
+    @Test
+    fun `nothing enabled hands the chain header a resolved zero, not an unknown`() = runTest {
+        // The header adds this to native staking. Null there would read as "not loaded" and blank
+        // the whole banner for everyone who never turned Earn on.
+        coEvery { selectionRepository.getSelectedVaults(VAULT_ID) } returns flowOf(emptySet())
+
+        val state = viewModel().apply { setData(VAULT_ID) }.state.value
+
+        assertEquals(0, BigDecimal.ZERO.compareTo(state.totalValue?.value))
+        assertEquals(AppCurrency.USD, state.totalValue?.currency)
```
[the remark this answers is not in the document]
aminsato: Applied in `6615b7835`, and for the whole file rather than only the added assertions — 48 of them, `assertEquals` → `shouldBe`, `assertNull`/`assertNotNull` → `shouldBeNull`/`shouldNotBeNull`, `assertTrue`/`assertFalse` → `shouldBe true`/`false`, and the one assertion carrying a message to `withClue`. Jupiter annotations kept.

That was my objection last round — converting only the new lines would have left one file written two ways — so the file is migrated as a whole instead.

aminsato:
@coderabbitai on the outside-diff finding for `KaminoEarnViewModel.kt` 424-438 (no thread to reply in) — verified and fixed in `6615b7835`.

`appCurrency = currency` is passed now. The repository's own KDoc asks for exactly that: a caller holding a currency must pass it, or the quote can be filed and returned under whichever currency happens to be selected by the time the lookup lands.

Checking it turned up a second hole in the same call. `getPriceByContactAddress` reports a miss as `BigDecimal.ZERO` rather than by throwing, so `runCatching { ... }.getOrNull() ?: return null` never fired for one — and a native token has no contract address, so the call returns zero without reaching the network at all. An unpriced vault was therefore still being folded into the total as a zero, which is the thing last round's nullable `fiatValue` was meant to stop. Only a strictly positive quote counts as a price now, which is the same rule the repository applies before caching one.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
```diff
@@ -380,32 +449,50 @@ constructor(
...
+                .getOrNull()
+                ?.takeIf { it.signum() > 0 } ?: return null
         return amount.multiply(price).setScale(2, RoundingMode.DOWN)
     }
 
-    private suspend fun fiatOrNull(amount: BigDecimal, coin: Coin): String? {
-        val value = fiatValueOrNull(amount, coin) ?: return null
+    private suspend fun fiatOrNull(amount: BigDecimal, coin: Coin, currency: AppCurrency): String? {
+        val value = fiatValueOrNull(amount, coin, currency) ?: return null
         return runCatching { appCurrencyRepository.getCurrencyFormat().format(value) }.getOrNull()
```
rkokhatskyi: fiatOrNull prices `value` with the pinned `currency` param but formats it via `getCurrencyFormat()`, an independent live read. A currency switch mid-buildRow, before the reload's cancellation lands, leaves the value priced in the old currency but stamped with the new one's symbol

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
aminsato: Fixed in 8766a598c.

`AppCurrencyRepository` now has a `getCurrencyFormat(appCurrency)` overload, so the symbol and the price come from the same currency. `fiatOrNull` asks for the format of the currency it priced in rather than reading the selection again; the no-arg overload delegates to it, so the locale and fraction-digit rules stay in one place. Each row still takes its own format instance — the repository clones per call under its mutex, and rows are built concurrently.

`SolanaStakingPositionsViewModel` read the format the same way, where `Pricing` pairs a format with a currency that could then disagree with itself, so it is pinned there too.

A test stubs the pinned overload to euros and the no-arg one to yen: it fails as `"￥100" should include substring "€"` without the change.

rkokhatskyi (approved):
LGTM

PR #5620, opened 2026-08-16
Fix Kamino SOL deposit shown as bogus SOL->WSOL swap on join device

johnnyluo:
## Summary
- A Kamino SOL-vault deposit wraps SOL and spends it in the same transaction, so its wSOL account nets down to just its rent-exempt residual (2,039,280 lamports / 0.00203928 SOL) in Blockaid's simulation diff — the wrap-in and spend-out cancel out and never show as their own diff.
- `BlockaidSimulationParser.parseSolanaMulti` picked that residual up as the "received" leg of a `Swap`, so a 0.059435 SOL deposit into Allez SOL was shown on the joining device as `0.059435 SOL -> 0.00203928 WSOL` instead of the real amount being sent.
- Native SOL and wrapped SOL are already normalised to the same mint address in this parser (the WSOL mint doubles as the sentinel for native SOL), so this adds a same-asset guard mirroring the existing EVM same-asset check a few lines above: when both legs resolve to the same asset, collapse to a `Transfer` of the real outgoing amount instead of a `Swap`.
- Not Kamino-specific: this is a general fix for any transaction that wraps and spends SOL in one shot.

## Testing
- `./gradlew :data:testDebugUnitTest --tests com.vultisig.wallet.data.securityscanner.blockaid.BlockaidSimulationParserTest` (added a regression test reproducing the reported numbers)
- `./gradlew :data:ktfmtCheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solana transaction interpretation for swaps, transfers, and SOL/WSOL conversions.
  * Prevented temporary native or wrapped SOL amounts from being incorrectly identified as swap destinations.
  * Correctly filters out withdrawal-like asset changes.
  * Raw dApp transactions now display an unverified status when simulation details are unavailable.

* **Tests**
  * Added coverage for SOL wrapping, WSOL withdrawals, batched wrapping, genuine swaps, and unverified raw dApp transactions.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
@@ -720,7 +720,13 @@ constructor(
                     buildHeroContent(
                         simulation = result.simulation,
                         decodedFunctionName = decodedFunctionName,
-                        didLoadSimulation = true,
+                        // A network failure or an undecodable payload never produced a real
+                        // Blockaid verdict (see [BlockaidKeysignScanResult.didLoadSimulation]) —
+                        // only a completed scan (including a legitimate "no balance change" empty
+                        // one) is eligible to flip the hero to Unverified below. Otherwise a
+                        // Blockaid outage would hide the trustworthy native amount on every
+                        // in-app raw-Solana flow (staking, Kamino) that also sets [signSolana].
+                        didLoadSimulation = result.didLoadSimulation,
                         // Solana never decodes a function name (no ABI to decode against), so a
                         // raw dApp-signed Solana transaction (Kamino, other program calls) is the
                         // only other signal that the payload's wire amount is unverified rather
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
@@ -721,6 +721,11 @@ constructor(
                         simulation = result.simulation,
                         decodedFunctionName = decodedFunctionName,
                         didLoadSimulation = true,
+                        // Solana never decodes a function name (no ABI to decode against), so a
+                        // raw dApp-signed Solana transaction (Kamino, other program calls) is the
+                        // only other signal that the payload's wire amount is unverified rather
+                        // than a plain transfer's trustworthy amount.
+                        isRawDappTransaction = payload.signSolana != null,
                     )
                 updateSendUiModel(verifyUiModel) { current ->
                     current.copy(transaction = current.transaction.copy(heroContent = hero))
--- a/data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
@@ -219,28 +219,38 @@ internal object BlockaidSimulationParser {
         val outAmount = parseRawAmount(outRaw) ?: return null
         val fromCoin = buildSolanaCoin(outSource.asset, outSource.assetType) ?: return null
 
-        val inSource = inSources.firstOrNull()
+        // Native SOL and wrapped SOL are normalised to the same mint in buildSolanaCoin (the WSOL
+        // address doubles as the native-SOL sentinel), so a same-asset in-leg is wrap/unwrap noise,
+        // not a real swap destination — e.g. a Kamino deposit's wSOL rent-exempt residual nets out
+        // as a small "incoming" diff. Prefer an in-leg that is a genuinely different asset (the
+        // same
+        // search order parseEvmSwap uses), so a batched signAllTransactions that also contains a
+        // real swap doesn't get shadowed by an unrelated SOL/WSOL leg landing first in `inSources`.
+        val inSource =
+            inSources.firstOrNull {
+                val coin = buildSolanaCoin(it.asset, it.assetType)
+                coin != null && !coin.address.equals(fromCoin.address, ignoreCase = true)
+            } ?: inSources.firstOrNull()
         val inRaw = inSource?.incoming?.rawValue?.toRawValueString()
         val inAmount = inRaw?.let(::parseRawAmount)
         val toCoin = inSource?.let { buildSolanaCoin(it.asset, it.assetType) }
-
-        // Native SOL and wrapped SOL are normalised to the same mint in buildSolanaCoin (the WSOL
-        // address doubles as the native-SOL sentinel), so a SOL-out/WSOL-in pair is the same asset
-        // on both sides — never a genuine swap. It shows up when a transaction wraps SOL and spends
-        // it in the same instruction (e.g. a Kamino deposit): the wSOL account's rent-exempt
-        // residual nets out as a small "incoming" diff, which would otherwise be read as the swap's
-        // destination amount. Falls back to a Transfer of the real out leg instead.
         val sameAsset = toCoin != null && toCoin.address.equals(fromCoin.address, ignoreCase = true)
 
-        return if (inAmount != null && toCoin != null && !sameAsset) {
-            BlockaidSimulationInfo.Swap(
-                fromCoin = fromCoin,
-                toCoin = toCoin,
-                fromAmount = outAmount,
-                toAmount = inAmount,
-            )
-        } else {
-            BlockaidSimulationInfo.Transfer(fromCoin = fromCoin, fromAmount = outAmount)
+        return when {
+            inAmount != null && toCoin != null && !sameAsset ->
+                BlockaidSimulationInfo.Swap(
+                    fromCoin = fromCoin,
+                    toCoin = toCoin,
+                    fromAmount = outAmount,
+                    toAmount = inAmount,
+                )
+            // Same asset on both legs with a bigger in-leg is a Kamino-style withdraw: the
+            // "outgoing" leg is just the temp wSOL account being debited before it closes, and the
+            // amount the user actually receives is the in-leg, not the out-leg. Pure-incoming diffs
+            // are intentionally not represented in the hero (see parseSolanaTransfer), so fall back
+            // to the generic title instead of showing a misleading outgoing transfer.
+            sameAsset && inAmount != null && inAmount > outAmount -> null
+            else -> BlockaidSimulationInfo.Transfer(fromCoin = fromCoin, fromAmount = outAmount)
         }
     }
 
--- a/data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
@@ -215,10 +215,19 @@ internal object BlockaidSimulationParser {
         // Regular swap path: pick by amount, not position. Blockaid does not contractually order
         // diffs, so a batch with a real outgoing leg plus a small network-fee leg (both
         // outgoing-only native SOL) must not have `firstOrNull` land on whichever happens to come
-        // first — the fee is always far smaller than a genuine outgoing leg.
+        // first — the fee is always far smaller than a genuine outgoing leg. Amounts are compared
+        // decimal-normalised (not raw): Blockaid nets one row per asset, so out rows always carry
+        // different decimals (e.g. a 2,039,280-lamport wSOL rent residual at 9 decimals is only
+        // 0.00203928 SOL, smaller than a 2,000,000-raw USDC leg at 6 decimals, i.e. $2 — but the
+        // raw
+        // integers alone rank the residual first).
         val outSource =
-            outSources.maxByOrNull {
-                it.outgoing?.rawValue?.toRawValueString()?.let(::parseRawAmount) ?: BigInteger.ZERO
+            outSources.maxByOrNull { diff ->
+                val raw =
+                    diff.outgoing?.rawValue?.toRawValueString()?.let(::parseRawAmount)
+                        ?: BigInteger.ZERO
+                val decimals = diff.asset.decimals?.clampDecimals() ?: 0
+                raw.toBigDecimal().movePointLeft(decimals)
             } ?: return null
         val outRaw = outSource.outgoing?.rawValue?.toRawValueString() ?: return null
         val outAmount = parseRawAmount(outRaw) ?: return null
--- a/data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
@@ -224,7 +224,15 @@ internal object BlockaidSimulationParser {
         val inAmount = inRaw?.let(::parseRawAmount)
         val toCoin = inSource?.let { buildSolanaCoin(it.asset, it.assetType) }
 
-        return if (inAmount != null && toCoin != null) {
+        // Native SOL and wrapped SOL are normalised to the same mint in buildSolanaCoin (the WSOL
+        // address doubles as the native-SOL sentinel), so a SOL-out/WSOL-in pair is the same asset
+        // on both sides — never a genuine swap. It shows up when a transaction wraps SOL and spends
+        // it in the same instruction (e.g. a Kamino deposit): the wSOL account's rent-exempt
+        // residual nets out as a small "incoming" diff, which would otherwise be read as the swap's
+        // destination amount. Falls back to a Transfer of the real out leg instead.
+        val sameAsset = toCoin != null && toCoin.address.equals(fromCoin.address, ignoreCase = true)
+
+        return if (inAmount != null && toCoin != null && !sameAsset) {
             BlockaidSimulationInfo.Swap(
                 fromCoin = fromCoin,
                 toCoin = toCoin,
--- a/data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
@@ -75,16 +75,28 @@ internal object BlockaidSimulationParser {
         // silently lose the user's actual swap result. Both `asset.type == "SOL"` and
         // `assetType == "SOL"` are checked because Blockaid is inconsistent about which field
         // carries the marker.
+        //
+        // Among outgoing-only native-SOL candidates, the fee is the SMALLEST one, not the first:
+        // a Kamino SOL deposit can emit [big SOL out (real leg), WSOL in (rent residual), tiny SOL
+        // out (fee)] — both the real leg and the fee are outgoing-only native SOL, and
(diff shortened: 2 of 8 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
```diff
@@ -224,7 +224,15 @@ internal object BlockaidSimulationParser {
...
-        return if (inAmount != null && toCoin != null) {
+        // Native SOL and wrapped SOL are normalised to the same mint in buildSolanaCoin (the WSOL
+        // address doubles as the native-SOL sentinel), so a SOL-out/WSOL-in pair is the same asset
+        // on both sides — never a genuine swap. It shows up when a transaction wraps SOL and spends
+        // it in the same instruction (e.g. a Kamino deposit): the wSOL account's rent-exempt
+        // residual nets out as a small "incoming" diff, which would otherwise be read as the swap's
+        // destination amount. Falls back to a Transfer of the real out leg instead.
+        val sameAsset = toCoin != null && toCoin.address.equals(fromCoin.address, ignoreCase = true)
+
+        return if (inAmount != null && toCoin != null && !sameAsset) {
```
rkokhatskyi: `buildSolanaCoin` maps native SOL onto the WSOL mint, so the guard also fires on a WSOL-out/SOL-in pair and this line drops `inAmount`. A Kamino SOL-vault withdraw closes the payout account (`KaminoComputeBudget.kt:59`), so its hero shows an outgoing WSOL leg instead of the SOL received

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
```diff
@@ -224,7 +224,15 @@ internal object BlockaidSimulationParser {
...
         val toCoin = inSource?.let { buildSolanaCoin(it.asset, it.assetType) }
 
-        return if (inAmount != null && toCoin != null) {
+        // Native SOL and wrapped SOL are normalised to the same mint in buildSolanaCoin (the WSOL
+        // address doubles as the native-SOL sentinel), so a SOL-out/WSOL-in pair is the same asset
+        // on both sides — never a genuine swap. It shows up when a transaction wraps SOL and spends
+        // it in the same instruction (e.g. a Kamino deposit): the wSOL account's rent-exempt
+        // residual nets out as a small "incoming" diff, which would otherwise be read as the swap's
+        // destination amount. Falls back to a Transfer of the real out leg instead.
+        val sameAsset = toCoin != null && toCoin.address.equals(fromCoin.address, ignoreCase = true)
```
rkokhatskyi, line 263: `sameAsset` only inspects `inSources.firstOrNull()` while `parseEvmSwap` searches past same-asset candidates at `BlockaidSimulationParser.kt:121`, so a `signAllTransactions` batch scanned as one diff set drops the real destination whenever the SOL/WSOL leg comes first, rendering a swap as a one-way send

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
```diff
@@ -212,27 +212,50 @@ internal object BlockaidSimulationParser {
...
 
-        // Regular swap path: pick by field presence rather than position. Blockaid does not
-        // contractually order diffs.
-        val outSource = outSources.firstOrNull() ?: return null
+        // Regular swap path: pick by amount, not position. Blockaid does not contractually order
+        // diffs, so a batch with a real outgoing leg plus a small network-fee leg (both
+        // outgoing-only native SOL) must not have `firstOrNull` land on whichever happens to come
+        // first — the fee is always far smaller than a genuine outgoing leg.
+        val outSource =
+            outSources.maxByOrNull {
```
rkokhatskyi: `maxByOrNull` ranks out legs by `raw_value` while `decimals` varies per diff, and Blockaid nets one row per asset so multiple out rows are always different assets. A wSOL rent row of 2039280 then outranks a 2 USDC leg and heroes the residual this PR exists to suppress

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
```diff
@@ -212,27 +212,50 @@ internal object BlockaidSimulationParser {
...
+                    fromAmount = outAmount,
+                    toAmount = inAmount,
+                )
+            // Same asset on both legs with a bigger in-leg is a Kamino-style withdraw: the
+            // "outgoing" leg is just the temp wSOL account being debited before it closes, and the
+            // amount the user actually receives is the in-leg, not the out-leg. Pure-incoming diffs
+            // are intentionally not represented in the hero (see parseSolanaTransfer), so fall back
+            // to the generic title instead of showing a misleading outgoing transfer.
+            sameAsset && inAmount != null && inAmount > outAmount -> null
+            else -> BlockaidSimulationInfo.Transfer(fromCoin = fromCoin, fromAmount = outAmount)
```
rkokhatskyi, line 279: `inAmount > outAmount` at line 257 never fires on a wrap or unwrap because the fee is charged natively, leaving the in leg one fee below the out leg. This line then heroes the gross out leg, so a 1 wSOL unwrap reads as sending 1.00203928 WSOL

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
```diff
@@ -212,27 +212,50 @@ internal object BlockaidSimulationParser {
...
+                    toCoin = toCoin,
+                    fromAmount = outAmount,
+                    toAmount = inAmount,
+                )
+            // Same asset on both legs with a bigger in-leg is a Kamino-style withdraw: the
+            // "outgoing" leg is just the temp wSOL account being debited before it closes, and the
+            // amount the user actually receives is the in-leg, not the out-leg. Pure-incoming diffs
+            // are intentionally not represented in the hero (see parseSolanaTransfer), so fall back
+            // to the generic title instead of showing a misleading outgoing transfer.
+            sameAsset && inAmount != null && inAmount > outAmount -> null
```
rkokhatskyi, line 278: There is no generic title to fall back to on Solana because `JoinKeysignFeeResolver.kt:47` returns null for all but EVM chains, so a null simulation drops to the legacy card headed You're sending. A Kamino withdraw then reads as sending the amount it pays out

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
johnnyluo: Good catch — fixed in e756d8b91 by normalising each out-leg's raw amount by its own `decimals` before comparing (`raw.toBigDecimal().movePointLeft(decimals)`), plus a regression test with a mixed-decimals wSOL-rent-vs-USDC batch.

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
johnnyluo: Agreed this is a real gap, but confirming the exact behavior needs real Blockaid wrap/unwrap trace data (whether the wSOL "out" leg is the SPL token-amount delta or the full account-lamport delta including rent) rather than a guess from the code alone. Leaving this open as a follow-up rather than shipping a fix built on an unverified assumption.

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
johnnyluo: Confirmed — `BuildHeroContentUseCase` only emits `Unverified` when `decodedFunctionName != null`, and `JoinKeysignFeeResolver.getTransactionFunctionInfo` only decodes for EVM chains, so a `null` Solana simulation does fall through to the legacy native-amount hero instead of a safe generic title. Giving Solana its own generic-title fallback is a cross-cutting change beyond this PR's scope (touches `BuildHeroContentUseCase`'s contract and needs a product decision on the copy) — leaving this open as a follow-up rather than bundling it in here.

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
```diff
@@ -212,27 +212,59 @@ internal object BlockaidSimulationParser {
...
+            else -> BlockaidSimulationInfo.Transfer(fromCoin = fromCoin, fromAmount = outAmount)
```
rkokhatskyi, line 279: A wrap has no wSOL out leg, so the rent attribution question does not apply and `BlockaidSimulationParser.kt:266` cannot fire because the native row funds the wSOL row plus the fee. Wrapping 1 SOL into an existing ATA shows 1.000005 SOL sent when only the fee leaves

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
```diff
@@ -212,27 +212,59 @@ internal object BlockaidSimulationParser {
...
+                    toCoin = toCoin,
+                    fromAmount = outAmount,
+                    toAmount = inAmount,
+                )
+            // Same asset on both legs with a bigger in-leg is a Kamino-style withdraw: the
+            // "outgoing" leg is just the temp wSOL account being debited before it closes, and the
+            // amount the user actually receives is the in-leg, not the out-leg. Pure-incoming diffs
+            // are intentionally not represented in the hero (see parseSolanaTransfer), so fall back
+            // to the generic title instead of showing a misleading outgoing transfer.
+            sameAsset && inAmount != null && inAmount > outAmount -> null
```
rkokhatskyi, line 278: Returning null here renders the withdraw as the outgoing transfer this arm exists to prevent: Solana never decodes a function name, so `VerifySendScreen.kt:272` falls back to You're sending with `payload.toAmount`. On main the same two diffs showed the inbound leg

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
johnnyluo: Fixed in b6eeb5c39: added `isRawDappTransaction` to `BuildHeroContentUseCase`, set from `payload.signSolana != null`. A null simulation on a raw dApp Solana transaction (Kamino withdraw and friends) now renders the safe `Unverified` hero instead of falling through to the native-amount hero with the payload's untrustworthy wire amount — the same treatment EVM already gets from a decoded function name. Plain native Solana transfers are unaffected since `payload.signSolana` is null for those.

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
johnnyluo: Agreed on the analysis — for a wrap, out (SOL debited) is always fee-more than in (WSOL credited), so `inAmount > outAmount` structurally can't fire and it falls to the gross-amount Transfer. Deferring this one though: fixing it correctly means asserting what Blockaid's wrap/unwrap diff actually represents (SPL delta vs. full account-lamport delta including rent) without a real trace to confirm against, same reservation as the unwrap case above. Tracking as a follow-up alongside it rather than shipping a guess.

rkokhatskyi, line 279: A wrap always reaches this line because out carries the wrapped amount plus rent and fee, so `BlockaidSimulationParser.kt:266` can never fire. Netting needs no model choice: its error is at most the fixed rent while the gross leg's grows with every unspent wrapped lamport

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
```diff
@@ -721,6 +721,11 @@ constructor(
                         simulation = result.simulation,
                         decodedFunctionName = decodedFunctionName,
                         didLoadSimulation = true,
+                        // Solana never decodes a function name (no ABI to decode against), so a
+                        // raw dApp-signed Solana transaction (Kamino, other program calls) is the
+                        // only other signal that the payload's wire amount is unverified rather
+                        // than a plain transfer's trustworthy amount.
+                        isRawDappTransaction = payload.signSolana != null,
```
aminsato, line 734: `signSolana` is not a "raw dApp transaction" marker — every in-app Solana flow that pre-builds raw transactions sets it: `SolanaDelegateViewModel:289`, `SolanaUnstakeViewModel:113`, `SolanaMoveStakeViewModel:185`, `SolanaFinishMoveViewModel:184`, `SolanaStakingPositionsViewModel:359`, `KaminoAmountViewModel:561`. Combined with `didLoadSimulation = true` being passed unconditionally, this flips the hero to `Unverified` ("Unverified function" + warning triangle, amount hidden) for all of them whenever the scan produced no simulation for a reason unrelated to the transaction: `BlockaidSimulationServiceImpl.scan` returns `EMPTY` on any `NetworkException`, and `scanSolana` returns `EMPTY` (cached, since it is not a failure) when the base64→base58 transcode fails.

Concrete: a co-signer joins a plain 0.5 SOL native-stake delegate while Blockaid is unreachable. Before: "0.5 SOL". After: "Unverified function — review the details below", with the real amount gone from the hero. The signal also loses meaning by firing on ordinary in-app staking.

Gating on something that actually identifies an externally-supplied payload (or at least distinguishing "simulation returned no balance change" from "scan failed") would keep the warning meaningful.

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
```diff
@@ -212,27 +212,59 @@ internal object BlockaidSimulationParser {
...
+        // raw
+        // integers alone rank the residual first).
+        val outSource =
+            outSources.maxByOrNull { diff ->
+                val raw =
+                    diff.outgoing?.rawValue?.toRawValueString()?.let(::parseRawAmount)
+                        ?: BigInteger.ZERO
+                val decimals = diff.asset.decimals?.clampDecimals() ?: 0
+                raw.toBigDecimal().movePointLeft(decimals)
+            } ?: return null
```
aminsato, line 243: The "real outgoing leg plus a small network-fee leg" case this ranking is justified by is unreachable at 3 diffs, because `parseSolana` has already dropped the real leg before `parseSolanaMulti` runs: the pre-filter at line 79-89 takes `indexOfFirst { isNativeSol && outgoing-only }`, i.e. the *first* outgoing-only native-SOL diff, not the smallest.

Concrete — a Kamino SOL deposit where Blockaid emits the fee as its own row:
```
[0] SOL  out 59,435,000   (real leg)
[1] WSOL in  2,039,280    (rent residual)
[2] SOL  out 5,000        (fee)
```
`all.size == 3` → `solFeeIndex = 0` → the 59,435,000 leg is discarded as "the fee". `parseSolanaMulti` then sees out = 5,000 / in = 2,039,280, hits the new same-asset branch (`inAmount > outAmount`) and returns null, so the deposit renders as "Unverified function" instead of the 0.059435 SOL transfer this PR is fixing. `maxByOrNull` never sees the two out rows.

The two hunks also assume opposite wire shapes: the new comment states Blockaid nets one row per asset (which is why the out rows "always carry different decimals"), while the 3-diff pre-filter exists precisely because it does not net the SOL fee. If netting does hold, two native-SOL out rows cannot co-occur and the fee rationale for `maxByOrNull` — plus the fixtures in the two new batch tests — describe a response Blockaid never sends. Picking the *smallest* outgoing-only native diff in the pre-filter would make both hunks consistent.

johnnyluo: Confirmed and fixed in b51c66d82. Reproduced your concrete example as a regression test (`solana three-diff SOL deposit with a separate fee leg keeps the real out amount`): the 3-diff pre-filter now picks the **smallest** outgoing-only native-SOL diff as the fee, not `indexOfFirst`. With `[59435000 SOL out, 2039280 WSOL in, 5000 SOL out]`, it now correctly drops the 5000-lamport fee and keeps the real 59435000-lamport leg, so the deposit renders as a 0.059435 SOL transfer instead of falling through to null/Unverified.

Left the "Blockaid nets one row per asset" vs. "3-diff prefilter assumes it doesn't net" tension as-is since both hunks now agree in practice (amount-based selection in both places), but appreciate you flagging the inconsistency — that's what surfaced the bug.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinKeysignViewModel.kt
johnnyluo: Fixed in b51c66d82, but not by narrowing the raw-dApp-tx marker — `dappMetadata` looked like the obvious signal but it's null for in-app Kamino/staking flows too (they build the payload locally, not via the extension), so gating on it would break the Kamino-withdraw fix this PR exists for. Went with your second suggestion instead: `BlockaidKeysignScanResult` now carries `didLoadSimulation`, false on a `NetworkException` or an undecodable Solana raw transaction, true for a completed scan (including a legitimate "no balance change" empty one). `JoinKeysignViewModel` now passes `result.didLoadSimulation` instead of a hardcoded `true`, so `isRawDappTransaction` can only flip the hero to Unverified when Blockaid actually returned a verdict — a network blip or decode failure now falls back to the native-amount hero, same as before this PR. Added regression tests for both failure paths in `BlockaidSimulationServiceImplTest`.

data/src/main/kotlin/com/vultisig/wallet/data/securityscanner/blockaid/BlockaidSimulationParser.kt
johnnyluo: Tried this: use the WSOL "in" amount instead of the gross native-SOL "out" amount whenever same-asset and in ≤ out. It broke the existing `solana SOL-out WSOL-in pair is a transfer, not a swap` test — that fixture (the Kamino-deposit case this PR fixes) has the exact same shape (same-asset, small in-leg ≤ big out-leg), but there the small "in" leg is rent-residual dust and the *out* leg is the real value, the opposite of the wrap case. Same-asset + in≤out can't tell a genuine wrap (in = the real wrapped amount, out inflated by fee/rent) apart from a deposit-and-immediately-spend (out = the real value, in = leftover dust) — both produce the identical diff shape. So this still needs something Blockaid's response doesn't give us from the shape alone (real trace data on how the two cases actually differ, if they do) — reverted, leaving it as the known gap it already was.

aminsato (approved):
LGTM

PR #5621, opened 2026-08-16
fix(defi): count Kamino Earn deposits in the DeFi portfolio

### LINKED ISSUE #5616: Kamino Earn deposits are missing from the DeFi Portfolio total
## Summary

A Kamino Earn deposit leaves the DeFi Portfolio untouched, so the vault's Solana row and the
portfolio header are both short by the deposited amount. Combined with the share-mint filter added
alongside the feature, a deposit disappears from every total in the app except the Earn tab it was
made from.

## Details

The DeFi Portfolio's rows and header both come from `AccountsRepository.loadDeFiAddresses`
(`data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt:416`), whose
per-chain balances are resolved by `BalanceRepository` — Solana routes to
`SolanaStakingDeFiBalanceService` on both the cached (`BalanceRepository.kt:256`) and the refresh
(`:431`) path, and that service sums delegated stake accounts and nothing else
(`SolanaStakingDeFiBalanceService.kt:35-50`). No `DeFiService` produces a Kamino balance: the six
implementations are THORChain, Maya, Solana staking, TON, Tron and Circle, and
`data/.../blockchain/solana/kamino/` is transaction-building and registry code only. So
`VaultAccountsViewModel.kt:598-609`, which totals what those accounts carry, never sees the
position.

The deposit is not counted as a wallet token either. #5588 filters Kamino share mints out of SPL
discovery (`SPLTokenRepository.kt:53-62`) on the premise that the Earn tab values them, which is
correct as far as double-counting goes but means the balance leaves the wallet total at deposit
time with nothing on the DeFi side to receive it.

Net effect for a vault holding, say, 1,000 USDC in Steakhouse USDC: the wallet total drops by
$1,000 when the deposit lands, the DeFi Portfolio header and its Solana row stay where they were,
and the only surface in the app that reports the money is the Earn tab of the Solana DeFi screen.

iOS counts it on both surfaces: `DefiBalanceService.swift:38-43` makes the Solana chain total
`solanaStakingTotalBalanceFiatDecimal + kaminoEarnTotalBalanceFiatDecimal`, and `:86-87` adds
`kaminoEarnPositionCount` to the chain's position count.

## Impact

Every vault with a Kamino Earn position under-reports its holdings by that amount on the DeFi
Portfolio, and reports it nowhere at all outside one tab. No funds at risk and no flow blocked —
the position itself, its Withdraw action and the on-chain state are all fine.

Related but distinct: #5602 is the same hole one screen deeper, in the Solana DeFi chain header.
That one is a UI-layer sum of two view-models; this one needs a balance producer.

## Suggested remediation

1. A Kamino `DeFiService` that reads the enabled vaults (`KaminoVaultSelectionRepository`) and the
   wallet's positions (`KaminoApi.getUserPositions` + `tokensPerShare`) and returns one
   `DeFiBalance.Balance` per vault, denominated in that vault's underlying coin — folded into the
   Solana branch of `BalanceRepository` alongside the staking service rather than replacing it.
2. A persisted snapshot of the last known position, so `getCacheDeFiBalance` has something to
   answer with. iOS reads a stored `KaminoPosition` for exactly this reason; Solana staking already
   does the same through `StakingDetailsRepository`.
3. Somewhere for the balance to land. A vault can hold a Kamino USDC position without carrying
   USDC-on-Solana as a wallet token, and `loadDeFiAddresses` builds its accounts from `vault.coins`
   — the same hole #5615 closes for sRUJI by injecting `Coins.defiOnly`. That mechanism is scoped
   to a fixed catalogue, so it needs generalizing rather than reusing as-is.
4. The cached path keys balances by ticker alone, across every chain at once
   (`AccountsRepository.kt:464`). A Kamino USDC balance on Solana and a Circle USDC balance on
   Ethereum would collide there and be applied to each other's account, so that key needs the chain
   in it before a second USDC-denominated DeFi source exists.

aminsato:
Fixes #5616

## Problem

A Kamino Earn deposit left the DeFi Portfolio untouched: no `DeFiService` produced a Kamino balance
at all, so `loadDeFiAddresses` never saw the position and neither the Solana row nor the header
counted it. Combined with #5588 keeping the share mints out of SPL discovery — correct, or the same
deposit would count twice — the money left the wallet total when the deposit landed and arrived
nowhere. The Earn tab was the only surface in the app reporting it.

## Changes

**`KaminoDeFiBalanceService`** — reads the enabled vaults and the wallet's positions and reports each
one in its vault's underlying token, since shares have no price of their own and are deliberately
not wallet tokens. Gated on the same per-vault opt-in the Earn tab uses, so the two surfaces cannot
disagree. One balance per token rather than per vault: two curated vaults are USDC, and the pipeline
resolves a chain's position by coin, so a second entry for the same coin would go unread. A failed
read falls back to the last known size — reporting zero on a real deposit reads as a loss — and zero
shares resolve without a metrics call, so a wallet holding nothing cannot be left unknown by one
failing.

**`KaminoPositionCacheRepository`** — persists the derived amount per kVault, so the cache-first read
has an answer before the network lands. Kamino has no cheap balance endpoint (a position is shares
times a live share price), which is why the derived figure is kept rather than the inputs; iOS
persists the same snapshot. Keyed by vault address, not token, since two vaults share the USDC
ticker.

**`SolanaDeFiBalanceService`** — Solana's DeFi position is two independent things, so this sums
native staking and Kamino per coin. Not a concatenation: the SOL Earn vault reports in the coin
native staking already uses, and the pipeline matches the first entry on the refresh path and the
last one on the cached path, so leaving them separate would drop one of the two SOL figures and make
a cold start disagree with a refresh.

Two pieces of plumbing this needed:

- **Somewhere for the balance to land.** A wallet can hold a position in a token it holds none of —
  it may have deposited its whole balance, or deposited on another device. Accounts are built from
  `vault.coins`, so the balance resolved to an account that was not there and was dropped. The
  curated vaults' tokens are now injected when missing, marked `Account.isPositionOnly`, and dropped
  again once they resolve to nothing, so a vault with no position gains no empty row. `AccountsLoader`
  filters them out of the form paths: their balance is a position, never something to spend.
- **Per-chain balance keys.** The cached path keyed balances by ticker across every chain at once.
  USDC is a Circle position on Ethereum and a Kamino one on Solana, so the two collided and one
  chain's balance was applied to the other chain's account. Keyed per chain now.

iOS totals the same two in `DefiBalanceService.totalBalanceInFiat`, and counts Kamino in
`defiPositionCount`.

## Test plan

16 new tests:

`KaminoDeFiBalanceServiceTest` (8) — a position reported in its underlying token; two vaults sharing
a token reported as one balance; a vault not enabled is never read; a wallet holding nothing reports
nothing and records that; a failed positions read falls back to the last known size; a failed share
price keeps the last known size; the cached read answers from the snapshot without touching the
network; a vault switched off stops counting even while its snapshot survives.

`SolanaDeFiBalanceServiceTest` (5) — staked SOL and a SOL Earn position summed rather than one
dropped; different tokens keep their own balances; a staking-only vault reported exactly as before;
neither held reports nothing; the cached read sums both sides too.

`AccountsRepositoryImplTest` (2) — a cached DeFi balance lands on its own chain's account, not
another chain's (verified to fail against the current code, where the Ethereum USDC account took the
Solana figure); a Kamino position lands on an account even when the vault holds none of the token;
and a token held neither in the wallet nor in a position adds no account.

`AccountsLoaderTest` (1) — the DeFi form path drops accounts that only ever hold a position.

Checked against a vault holding 1 staked SOL and 0.555968 USDC in Steakhouse USDC: the Solana row
went from "$75.36 · 1 SOL" to "$75.81 · 2 Assets", matching the chain screen one tap deeper, and
disabling the vault under Manage Positions drops it back to the staking figure.

## Note for review

Touches `AccountsRepository.loadDeFiAddresses` and `AccountsLoader`, which #5615 also edits. Whichever
lands second needs a rebase. `dropUnfundedPositionOnly` / `isPositionOnly` here and
`dropUnfundedDefiOnly` / `Coins.isDefiOnly` there are the same mechanism for two different reasons —
worth collapsing into one once both are in.

## Summary by CodeRabbit

- **New Features**
  - Added support for viewing Kamino Earn positions alongside Solana native staking balances.
  - Added reliable DeFi balance aggregation, including cached values when network data is unavailable.
  - Added persistence for DeFi position balances between sessions.
  - Added eligible position tokens to portfolio displays, even when they are not directly spendable.

- **Bug Fixes**
  - Prevented position-only accounts from appearing as spendable accounts.
  - Improved token grouping across chains to prevent balance conflicts.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
@@ -1,0 +1,151 @@
+package com.vultisig.wallet.data.blockchain.solana.kamino
+
+import com.vultisig.wallet.data.api.KaminoApi
+import com.vultisig.wallet.data.api.KaminoUserPositionJson
+import com.vultisig.wallet.data.blockchain.DeFiService
+import com.vultisig.wallet.data.blockchain.model.DeFiBalance
+import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.repositories.KaminoPositionCacheRepository
+import com.vultisig.wallet.data.repositories.KaminoVaultSelectionRepository
+import java.math.BigDecimal
+import java.math.BigInteger
+import kotlin.coroutines.cancellation.CancellationException
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.supervisorScope
+import timber.log.Timber
+
+/**
+ * DeFi-balance provider for the Kamino Earn vaults on Solana.
+ *
+ * A position is held as vault shares, which are worth a live, per-vault multiple of the underlying
+ * token, so the balance reported here is the token amount those shares redeem for — denominated in
+ * the vault's own underlying (USDC, SOL) rather than in shares, which the app has no price for and
+ * deliberately keeps out of wallet-token discovery.
+ *
+ * Gated on the per-vault opt-in, matching the Earn tab: a vault the user has switched off shows no
+ * card there and must not show a balance here either, or the two surfaces disagree.
+ *
+ * Mirrors iOS `DefiBalanceService.kaminoEarnTotalBalanceFiatDecimal`.
+ */
+class KaminoDeFiBalanceService(
+    private val kaminoApi: KaminoApi,
+    private val selectionRepository: KaminoVaultSelectionRepository,
+    private val positionCache: KaminoPositionCacheRepository,
+) : DeFiService {
+
+    override suspend fun getRemoteDeFiBalance(address: String, vaultId: String): List<DeFiBalance> {
+        val enabled = enabledVaults(vaultId)
+        if (enabled.isEmpty()) return emptyList()
+
+        return try {
+            // One call for the whole wallet. An empty list is a real answer — the wallet holds no
+            // position — while a throw is not knowing, which is why it is caught below rather than
+            // collapsed into zero.
+            val positions = kaminoApi.getUserPositions(address).associateBy { it.vaultAddress }
+            val cached = positionCache.getPositions(vaultId)
+            Timber.d(
+                "KaminoDeFiBalanceService: %d enabled vaults, %d positions held",
+                enabled.size,
+                positions.size,
+            )
+
+            val amounts =
+                supervisorScope {
+                        enabled
+                            .map { vault ->
+                                async {
+                                    // A vault whose own read failed keeps its last known size:
+                                    // reporting zero would read as a deposit that has gone.
+                                    val amount =
+                                        amountOf(vault, positions[vault.address])
+                                            ?: cached[vault.address]
+                                    amount?.let { vault to it }
+                                }
+                            }
+                            .awaitAll()
+                    }
+                    .filterNotNull()
+                    .toMap()
+
+            positionCache.savePositions(vaultId, amounts.mapKeys { it.key.address })
+            balancesOf(amounts)
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.w(e, "KaminoDeFiBalanceService: failed to read positions")
+            getCacheDeFiBalance(address, vaultId)
+        }
+    }
+
+    override suspend fun getCacheDeFiBalance(address: String, vaultId: String): List<DeFiBalance> {
+        return try {
+            val enabled = enabledVaults(vaultId)
+            if (enabled.isEmpty()) return emptyList()
+            val cached = positionCache.getPositions(vaultId)
+            balancesOf(
+                enabled.mapNotNull { vault -> cached[vault.address]?.let { vault to it } }.toMap()
+            )
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.w(e, "KaminoDeFiBalanceService: failed to read the cached positions")
+            emptyList()
+        }
+    }
+
+    private suspend fun enabledVaults(vaultId: String): List<KaminoVault> {
+        val enabled = selectionRepository.getSelectedVaults(vaultId).first()
+        return KaminoVaultRegistry.ALLOW_LIST.filter { it.address in enabled }
+    }
+
+    /**
+     * What the wallet's shares in [vault] redeem for, or null when that cannot be established.
+     *
+     * Zero shares need no share price, so a wallet holding nothing in a vault resolves without a
+     * metrics call — and cannot be left unknown by one failing.
+     */
+    private suspend fun amountOf(
+        vault: KaminoVault,
+        position: KaminoUserPositionJson?,
+    ): BigInteger? {
+        val shares = KaminoPositionMath.decimalOrNull(position?.totalShares) ?: BigDecimal.ZERO
+        if (shares.signum() == 0) return BigInteger.ZERO
+
+        val tokensPerShare =
+            runCatching { kaminoApi.getVaultMetrics(vault.address) }
+                .getOrNull()
+                ?.let { KaminoPositionMath.decimalOrNull(it.tokensPerShare) } ?: return null
+
+        return KaminoPositionMath.tokenAmount(shares, tokensPerShare, vault.tokenDecimals)
+            .movePointRight(vault.tokenDecimals)
+            .toBigInteger()
+    }
+
+    /**
+     * One balance per underlying token, not per vault: two of the curated vaults are USDC, and the
+     * balance pipeline resolves a chain's DeFi position by coin, so two entries for the same coin
+     * would leave whichever came second unread.
+     */
+    private fun balancesOf(amounts: Map<KaminoVault, BigInteger>): List<DeFiBalance> {
+        val balances =
+            amounts
+                .mapNotNull { (vault, amount) ->
+                    vault.coin?.takeIf { amount.signum() > 0 }?.let { coin -> coin to amount }
+                }
+                .groupBy { (coin, _) -> coin.id.lowercase() }
+                .map { (_, entries) ->
+                    DeFiBalance.Balance(
+                        coin = entries.first().first,
+                        amount =
+                            entries.fold(BigInteger.ZERO) { running, (_, amount) ->
+                                running + amount
+                            },
+                    )
+                }
+
+        return if (balances.isEmpty()) emptyList()
+        else listOf(DeFiBalance(chain = Chain.Solana, balances = balances))
+    }
+}
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
@@ -25,7 +25,12 @@ import timber.log.Timber
  * deliberately keeps out of wallet-token discovery.
  *
  * Gated on the per-vault opt-in, matching the Earn tab: a vault the user has switched off shows no
- * card there and must not show a balance here either, or the two surfaces disagree.
+ * card there and must not count here either.
+ *
+ * On a failed read the two surfaces do part company, deliberately. A card can say it does not know
+ * the position and the user reads that; a portfolio total has no such affordance, so leaving the
+ * position out would quietly understate every figure above it. This side therefore keeps the last
+ * known size while the card shows the position as unavailable.
  *
  * Mirrors iOS `DefiBalanceService.kaminoEarnTotalBalanceFiatDecimal`.
  */
@@ -36,22 +41,25 @@ class KaminoDeFiBalanceService(
 ) : DeFiService {
 
     override suspend fun getRemoteDeFiBalance(address: String, vaultId: String): List<DeFiBalance> {
-        val enabled = enabledVaults(vaultId)
-        if (enabled.isEmpty()) return emptyList()
+        val amounts =
+            try {
+                // The opt-in read is inside the guard like every other read here: this service is
+                // one of two the Solana provider awaits side by side, so an escaping throw would
+                // take the healthy staking figure down with it.
+                val enabled = enabledVaults(vaultId)
+                if (enabled.isEmpty()) return emptyList()
 
-        return try {
-            // One call for the whole wallet. An empty list is a real answer — the wallet holds no
-            // position — while a throw is not knowing, which is why it is caught below rather than
-            // collapsed into zero.
-            val positions = kaminoApi.getUserPositions(address).associateBy { it.vaultAddress }
-            val cached = positionCache.getPositions(vaultId)
-            Timber.d(
-                "KaminoDeFiBalanceService: %d enabled vaults, %d positions held",
-                enabled.size,
-                positions.size,
-            )
+                // One call for the whole wallet. An empty list is a real answer — the wallet holds
+                // no position — while a throw is not knowing, which is why it is caught below
+                // rather than collapsed into zero.
+                val positions = kaminoApi.getUserPositions(address).associateBy { it.vaultAddress }
+                val cached = positionCache.getPositions(vaultId)
+                Timber.d(
+                    "KaminoDeFiBalanceService: %d enabled vaults, %d positions held",
+                    enabled.size,
+                    positions.size,
+                )
 
-            val amounts =
                 supervisorScope {
                         enabled
                             .map { vault ->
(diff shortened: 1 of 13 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
```diff
@@ -0,0 +1,151 @@
...
+            )
+
+            val amounts =
+                supervisorScope {
+                        enabled
+                            .map { vault ->
+                                async {
+                                    // A vault whose own read failed keeps its last known size:
+                                    // reporting zero would read as a deposit that has gone.
+                                    val amount =
```
rkokhatskyi, line 68: On a getVaultMetrics failure for one vault, `amountOf` returns null and the caller falls back to the persisted cache, so the portfolio shows a stale balance while the Earn tab, which keeps no cache, shows Value unavailable for the same position, contradicting this file's own claim the two surfaces cannot disagree.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
```diff
@@ -0,0 +1,151 @@
...
+     * metrics call — and cannot be left unknown by one failing.
+     */
+    private suspend fun amountOf(
+        vault: KaminoVault,
+        position: KaminoUserPositionJson?,
+    ): BigInteger? {
+        val shares = KaminoPositionMath.decimalOrNull(position?.totalShares) ?: BigDecimal.ZERO
+        if (shares.signum() == 0) return BigInteger.ZERO
+
+        val tokensPerShare =
```
rkokhatskyi, line 156: `amountOf` never checks tokensPerShare is positive, only that it parses. A metrics response with tokensPerShare=0 parses to BigDecimal.ZERO, so it returns zero instead of null, skipping the cache fallback, and `savePositions` at `KaminoDeFiBalanceService.kt:72` then persists that zero, erasing the real cached position until a later good read.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
```diff
@@ -0,0 +1,151 @@
...
+ * Mirrors iOS `DefiBalanceService.kaminoEarnTotalBalanceFiatDecimal`.
+ */
+class KaminoDeFiBalanceService(
+    private val kaminoApi: KaminoApi,
+    private val selectionRepository: KaminoVaultSelectionRepository,
+    private val positionCache: KaminoPositionCacheRepository,
+) : DeFiService {
+
+    override suspend fun getRemoteDeFiBalance(address: String, vaultId: String): List<DeFiBalance> {
+        val enabled = enabledVaults(vaultId)
```
rkokhatskyi: `enabledVaults(vaultId)` here runs before the try block, unlike the same call properly guarded inside getCacheDeFiBalance. A DataStore read failure throws uncaught, and since `SolanaDeFiBalanceService.kt:26-29` runs staking and kamino as sibling asyncs in one coroutineScope, it cancels the healthy staking result too, dropping the whole Solana row for that refresh.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/KaminoPositionCacheRepository.kt
```diff
@@ -0,0 +1,66 @@
...
+        return stored
+            .mapNotNull(::decode)
+            .filter { KaminoVaultRegistry.isAllowed(it.first) }
+            .toMap()
+    }
+
+    /** Replaces the snapshot wholesale, so a vault that no longer holds anything stops counting. */
+    suspend fun savePositions(vaultId: String, positions: Map<String, BigInteger>) {
+        context.kaminoPositionDataStore.edit { preferences ->
+            preferences[positionsKey(vaultId)] =
```
rkokhatskyi: savePositions wholesale replaces the on disk snapshot with only the currently enabled vaults, this assigns rather than merges, and `KaminoDeFiBalanceService.kt:56-72` builds that map only from enabled vaults. Disabling a vault then re-enabling it wipes its cached amount, so the next cached-first render under-reports the DeFi total until the following network refresh corrects it.

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
aminsato: Real, and worse than a wrong figure: the zero was persisted. `tokensPerShare="0"` parses, so `amountOf` answered zero instead of null, skipped the cache fallback, and `savePositions` wrote the zero over the last known size — the position stayed gone until a later good read.

Only a positive share price counts as an answer now; zero joins unparseable and unread as "not knowing", so the position keeps its size and nothing is persisted over it.

`a share price of zero counts as unknown rather than as an empty position` covers both halves — the balance and that the zero never reaches the snapshot. Pushed in `14c72d88c`.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
aminsato: Real. Moved inside the guard.

Confirmed the blast radius you describe: `SolanaDeFiBalanceService` awaits this service and staking as siblings in one `coroutineScope`, so the escaping throw cancelled the staking result too and the whole Solana row went. `getCacheDeFiBalance` re-reads the opt-in inside its own guard, so the fallback path degrades to an empty list rather than throwing again.

`an opt-in read that fails leaves the rest of the chain's load standing` locks it in. Pushed in `14c72d88c`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/KaminoPositionCacheRepository.kt
aminsato: Real. It merges now.

The wholesale replace only ever saw the switched-on vaults, so disabling one dropped its amount at the next refresh and re-enabling it left the cache-first render short until the network answered. Merging keeps what a write says nothing about, and a vault reported as holding nothing still arrives as an explicit zero, so "no longer holds anything stops counting" still holds.

To get this under test the store moved from its own `Context` DataStore to `AppDataStore`, the seam `PromoBannerDismissalRepositoryImpl` already uses. No migration: the class is new in this PR, so nothing is persisted in production. New `KaminoPositionCacheRepositoryTest` covers the merge, the zero, per-vault isolation and the allow-list drop. Pushed in `14c72d88c`.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
aminsato: The divergence is real — I checked the other side: on a metrics failure `KaminoEarnViewModel` leaves `tokenAmount` null and the card renders `FIAT_VALUE_UNAVAILABLE`, while this side hands back the persisted amount.

I kept the behaviour and fixed the claim, because the two surfaces are not the same kind of thing. A card owns one position and can say it does not know it — the user reads that and knows to retry. A portfolio total has no such affordance: dropping the position silently understates the Solana row, the chain header and the vault total, with nothing anywhere to say a figure is missing. Between a stale number and a quietly wrong one, the stale number is the one the user can act on.

So the KDoc no longer claims general parity. It scopes the "matching the Earn tab" claim to the opt-in gate, which really is enforced on both sides, and states the failure-path difference as deliberate.

Making them actually agree is possible — the Earn tab would have to read this snapshot rather than keep none — but that is a change to `KaminoEarnViewModel`, which #5619 is currently rewriting, so it would collide. Happy to file it as a follow-up if you want the card falling back too. Pushed in `14c72d88c`.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
```diff
@@ -0,0 +1,151 @@
...
+                    .filterNotNull()
+                    .toMap()
+
+            positionCache.savePositions(vaultId, amounts.mapKeys { it.key.address })
+            balancesOf(amounts)
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            Timber.w(e, "KaminoDeFiBalanceService: failed to read positions")
+            getCacheDeFiBalance(address, vaultId)
```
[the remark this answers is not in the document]
aminsato: Confirmed and fixed. `savePositions` sat inside the read's fallback, so a store that refused the write threw into the same catch as a failed network read and the fresh amounts were discarded in favour of the stale cache.

The snapshot is now persisted after the read resolves, in its own guard that rethrows `CancellationException` and logs anything else — no broad swallow. A snapshot that could not be kept costs the next cold start a stale figure; discarding the read would have cost this one its whole position.

`a snapshot the store refuses to keep does not discard the balance just read` covers it. Pushed in `14c72d88c`.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
```diff
@@ -0,0 +1,151 @@
...
+        vault: KaminoVault,
+        position: KaminoUserPositionJson?,
+    ): BigInteger? {
+        val shares = KaminoPositionMath.decimalOrNull(position?.totalShares) ?: BigDecimal.ZERO
+        if (shares.signum() == 0) return BigInteger.ZERO
+
+        val tokensPerShare =
+            runCatching { kaminoApi.getVaultMetrics(vault.address) }
+                .getOrNull()
+                ?.let { KaminoPositionMath.decimalOrNull(it.tokensPerShare) } ?: return null
```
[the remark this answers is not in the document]
aminsato: Confirmed and fixed. `runCatching` catches `Throwable`, so a cancelled metrics call came back as "no metrics" and the vault quietly resolved from cache while the load carried on inside a scope that was already going away.

Replaced with an explicit `try`/`catch` that rethrows `CancellationException` and logs other failures, keeping the null fallback for those. `a cancelled share-price read stops the load rather than falling back` asserts the cancellation now propagates out of `getRemoteDeFiBalance`. Pushed in `14c72d88c`.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
```diff
@@ -0,0 +1,185 @@
...
+     *
+     * Zero shares need no share price, so a wallet holding nothing in a vault resolves without a
+     * metrics call — and cannot be left unknown by one failing.
+     */
+    private suspend fun amountOf(
+        vault: KaminoVault,
+        position: KaminoUserPositionJson?,
+    ): BigInteger? {
+        val shares = KaminoPositionMath.decimalOrNull(position?.totalShares) ?: BigDecimal.ZERO
+        if (shares.signum() == 0) return BigInteger.ZERO
```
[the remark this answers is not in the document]
aminsato: Confirmed and fixed — this is the share-price finding from the last round mirrored onto the shares side, and it contradicted `KaminoPositionMath.decimalOrNull`'s own doc ("a malformed number is treated as missing rather than as zero").

`?: BigDecimal.ZERO` collapsed three different situations into "holds nothing": no entry for the vault, a shares string that will not parse, and a negative one. Only the first is an answer. For the other two the zero skipped the `?: cached[vault.address]` fallback and was then persisted over the last known size, so one malformed payload emptied the position from the Solana row, the chain header and the vault total, and kept it empty across cold starts.

Taken with a guard shaped like the `tokensPerShare` one right below it rather than the diff as proposed:

```kotlin
if (position == null) return BigInteger.ZERO

val shares =
    KaminoPositionMath.decimalOrNull(position.totalShares)?.takeIf { it.signum() >= 0 }
        ?: return null
if (shares.signum() == 0) return BigInteger.ZERO
```

Unknown shares now skip the metrics call too — there is no point pricing a quantity that is not known.

Four tests: unparseable, negative and absent `totalShares` each keep the cached size and write it back unchanged, and a fourth pins the other half of the rule — an explicit `"0"` must still erase the old figure, or a withdrawn position would linger on the portfolio for good. The three behaviour ones were verified to fail against the reverted code. Pushed in `c0318615f`.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * A position is held as vault shares, which are worth a live, per-vault multiple of the underlying
+ * token, so the balance reported here is the token amount those shares redeem for — denominated in
+ * the vault's own underlying (USDC, SOL) rather than in shares, which the app has no price for and
+ * deliberately keeps out of wallet-token discovery.
+ *
+ * Gated on the per-vault opt-in, matching the Earn tab: a vault the user has switched off shows no
+ * card there and must not count here either.
+ *
+ * On a failed read the two surfaces do part company, deliberately. A card can say it does not know
```
rkokhatskyi, line 29: This comment claims the divergence is contained to a stale portfolio total against an unavailable card, but `KaminoEarnViewModel.kt:365-367` defaults fiatValue to BigDecimal.ZERO on the same null tokenAmount, and `:407-408` folds that into totalFiat, so the Earn tab's own section total silently zeroes on the identical failure.

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoDeFiBalanceService.kt
johnnyluo: Checked against the current file — this claim no longer holds. `KaminoEarnRow.fiatValue` is nullable and only set when `tokenAmount` resolves (KaminoEarnViewModel.kt:436-439: `tokenAmount?.let { amount -> coin?.let { fiatValueOrNull(amount, it, currency) } }`), and `totalFiatValue` returns null as soon as any row's `fiatValue` is null rather than folding it in as zero (KaminoEarnViewModel.kt:499-505). That null-propagation was landed by #5619 (merged to main), and this branch has since been rebased onto it, so the Earn tab's own section total goes unavailable on the same failure rather than silently zeroing — the KDoc's claim is accurate as written. No code change needed here.

johnnyluo (approved):
LGTM

PR #5622, opened 2026-08-16
fix(defi): re-read the DeFi list when home comes back to the front

### LINKED ISSUE #5618: DeFi Portfolio keeps pre-change figures until the tab is switched
## Summary

The DeFi Portfolio keeps the figures it loaded before a position changed. Enable a Kamino vault,
toggle a THORChain position, or sign a deposit — all of which happen a screen deeper — press back,
and the rows and the header still show the pre-change values. They only correct themselves on a tab
switch, so for as long as the user stays on the DeFi tab it disagrees with the chain screen one tap
away.

## Details

`VaultAccountsViewModel.loadDeFiBalances` (`app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt:536`)
has exactly three callers:

- `loadData` (`:278`), on app start or a vault switch — and with `isRefresh = vaultChanged`, so the
  first load of a session is cache-only,
- `setCryptoConnectionType` (`:764`), when the user switches onto the DeFi tab,
- the chain-picker sheet closing (`:741`).

Returning from a chain screen is none of those, so nothing asks the list to look again.

Pull-to-refresh does not cover it either: `refreshData` (`:346`) calls `loadAccounts` (`:349`) — the
wallet list — and leaves the DeFi rows the user is pulling on untouched.

The data itself is fine. `BroadcastKeysignUseCase.kt:184` already invalidates the chain's DeFi
balance cache after every broadcast, so a re-read would return the new value; nothing performs one.

## Impact

Cosmetic but persistent, and it undermines a fix one level down: any correction to what the DeFi
rows report is invisible until the user happens to switch tabs. Worst on the position toggles, where
the user has just told the app what they want to see and the list still shows the old answer. No
funds at risk.

## Suggested remediation

Re-read the DeFi list when home comes back to the front while the DeFi tab is active, and make
pull-to-refresh on that tab refresh the list being pulled. The first load being cache-only means a
resume read is not a duplicate fetch — it upgrades that render to a real one.

aminsato:
Fixes #5618

## Problem

The DeFi Portfolio kept the figures it loaded before a position changed. `loadDeFiBalances` has three
callers — app start / vault switch (cache-only on the first load), switching onto the DeFi tab, and
the chain picker closing. Returning from a chain screen is none of them, so enabling a Kamino vault,
toggling a THORChain position or signing a deposit and pressing back left the rows and the header on
their pre-change values until the user happened to switch tabs.

Pull-to-refresh did not cover it either: `refreshData` reloaded the wallet accounts, which is the
other list, leaving the one being pulled on untouched.

The data was already fine — `BroadcastKeysignUseCase` invalidates the chain's DeFi balance cache
after every broadcast — nothing performed the re-read.

## Changes

- `onScreenResumed()`, called from a `LifecycleResumeEffect` on the home screen, re-reads the DeFi
  list when home comes back to the front. Guarded to the DeFi tab, so the wallet's streaming load is
  untouched.
- `refreshData()` now also refreshes the DeFi rows when that tab is the one in view.

The first load is cache-only, so the resume read is not a duplicate fetch — it upgrades that first
cached render to a real one.

## Test plan

Three tests in a new `VaultAccountsViewModelTest`:

- returning to the DeFi tab re-reads the positions (a network read, not the cached one the first
  load settles for)
- returning to the wallet tab leaves the DeFi list alone
- pulling to refresh on the DeFi tab refreshes the list being pulled

Checked on device with a Kamino position: unticking the vault under Manage Positions and pressing
back drops the Solana row to $75.16 · 1 SOL immediately, matching the chain header, and ticking it
back restores $75.72 · 2 Assets — both without switching tabs or pulling.

## Summary by CodeRabbit

* **Bug Fixes**
  * DeFi balances now refresh when returning to the home screen, reopening a DeFi account, or manually refreshing from the DeFi tab.
  * Account information is re-read after navigating back from deeper screens.
  * Refresh indicators accurately reflect completed, interrupted, or failed wallet and DeFi loads.
  * Automatic refreshes are throttled, duplicate vault entries are avoided, and backup warnings update promptly.

* **Tests**
  * Added coverage for refresh behavior, throttling, navigation, and backup status updates.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -599,7 +599,7 @@ constructor(
                             .loadDeFiAddresses(vaultId, readsNetwork)
                             .map { addresses -> addresses.sortByAccountsTotalFiatValue() }
                             .catch { error ->
-                                Timber.e(error, "Error loading DeFi balances for vault: $vaultId")
+                                Timber.e(error, "Error loading DeFi balances for vault: %s", vaultId)
                             }
                             // This load, not the wallet stream, is what a pull on the DeFi tab is
                             // waiting on. A cancelled one is superseded by the load that replaced
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -347,6 +347,26 @@ constructor(
         val vaultId = vaultId ?: return
         updateRefreshing(true)
         loadAccounts(vaultId)
+        // A pull on the DeFi tab has to reload the DeFi rows: they come from their own load, so
+        // refreshing only the wallet accounts left the list the user was pulling on untouched.
+        if (uiState.value.cryptoConnectionType == CryptoConnectionType.Defi) {
+            loadDeFiBalances(vaultId, isRefresh = true)
+        }
+    }
+
+    /**
+     * Re-reads the DeFi list when home comes back to the front.
+     *
+     * Positions are changed a screen deeper — a Kamino vault switched on under Manage Positions, a
+     * deposit signed — and nothing on the way back asks this list to look again, so it kept
+     * rendering the figures it had loaded before the change. The wallet tab has its own streaming
+     * load and is left alone.
+     */
+    fun onScreenResumed() {
+        val vaultId = vaultId ?: return
+        if (uiState.value.cryptoConnectionType == CryptoConnectionType.Defi) {
+            loadDeFiBalances(vaultId, isRefresh = true)
+        }
     }
 
     fun send() {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -56,10 +56,12 @@ import com.vultisig.wallet.ui.utils.throttleLatest
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
 import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.combine
@@ -69,6 +71,7 @@ import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.receiveAsFlow
 import kotlinx.coroutines.flow.update
@@ -166,6 +169,15 @@ constructor(
     private var loadAccountsJob: Job? = null
     private var loadDeFiBalancesJob: Job? = null
     private var bannerJob: Job? = null
+    private var deFiRefreshThrottleJob: Job? = null
+
+    // The tab a pull-to-refresh was started on, and the only load allowed to take the spinner back
+    // down. On the DeFi tab the rows under the user's finger come from loadDeFiBalances, so the
+    // wallet stream reaching isComplete says nothing about whether they are still stale.
+    private var refreshOwner: CryptoConnectionType? = null
+
+    // True while a network DeFi read is recent enough to stand in for a resume-triggered one.
+    private var isDeFiRefreshThrottled = false
 
     // Last merged snapshot per stream (wallet / DeFi), keyed in-memory so a chain that comes back
     // with a null balance mid-refetch can carry its previously-shown value forward (#4768). Reset
@@ -345,11 +357,13 @@ constructor(
 
     fun refreshData() {
         val vaultId = vaultId ?: return
+        val pulledTab = uiState.value.cryptoConnectionType
+        refreshOwner = pulledTab
         updateRefreshing(true)
         loadAccounts(vaultId)
         // A pull on the DeFi tab has to reload the DeFi rows: they come from their own load, so
         // refreshing only the wallet accounts left the list the user was pulling on untouched.
-        if (uiState.value.cryptoConnectionType == CryptoConnectionType.Defi) {
+        if (pulledTab == CryptoConnectionType.Defi) {
             loadDeFiBalances(vaultId, isRefresh = true)
         }
     }
@@ -361,12 +375,17 @@ constructor(
      * deposit signed — and nothing on the way back asks this list to look again, so it kept
      * rendering the figures it had loaded before the change. The wallet tab has its own streaming
      * load and is left alone.
+     *
+     * Every return to home resumes, including ones that changed nothing (Receive, the vault list)
+     * and ones that already asked for the list themselves ([openAddChainAccount] reloads once the
+     * chain picker's result lands), so a read that recent stands in for this one — the same
+     * throttled-on-appear iOS puts on its DeFi screen.
      */
     fun onScreenResumed() {
         val vaultId = vaultId ?: return
-        if (uiState.value.cryptoConnectionType == CryptoConnectionType.Defi) {
-            loadDeFiBalances(vaultId, isRefresh = true)
-        }
+        if (uiState.value.cryptoConnectionType != CryptoConnectionType.Defi) return
+        if (isDeFiRefreshThrottled) return
+        loadDeFiBalances(vaultId, isRefresh = true)
     }
 
     fun send() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -347,6 +347,26 @@ constructor(
         val vaultId = vaultId ?: return
         updateRefreshing(true)
         loadAccounts(vaultId)
+        // A pull on the DeFi tab has to reload the DeFi rows: they come from their own load, so
+        // refreshing only the wallet accounts left the list the user was pulling on untouched.
+        if (uiState.value.cryptoConnectionType == CryptoConnectionType.Defi) {
+            loadDeFiBalances(vaultId, isRefresh = true)
```
rkokhatskyi, line 371: `loadDeFiBalances` clears `isRefreshing` only in its `.catch`, so `refreshData`'s spinner is cleared solely by `loadAccounts`'s `it.isComplete` for the wallet stream, unrelated to whether this DeFi reload finished, so it can read done while the visible rows are still stale

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -347,6 +347,26 @@ constructor(
...
+     *
+     * Positions are changed a screen deeper — a Kamino vault switched on under Manage Positions, a
+     * deposit signed — and nothing on the way back asks this list to look again, so it kept
+     * rendering the figures it had loaded before the change. The wallet tab has its own streaming
+     * load and is left alone.
+     */
+    fun onScreenResumed() {
+        val vaultId = vaultId ?: return
+        if (uiState.value.cryptoConnectionType == CryptoConnectionType.Defi) {
+            loadDeFiBalances(vaultId, isRefresh = true)
```
rkokhatskyi: `onScreenResumed` reloads unconditionally on every resume with no throttle, so closing `Receive`, `VaultList`, or the chain-picker dialog from the DeFi tab also triggers it, racing `openAddChainAccount`'s own reload at `:761` for a guaranteed duplicate fetch on that path

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Real, fixed in cd59da089.

The spinner now belongs to the tab the pull was started on. `refreshData` records that tab in `refreshOwner`, and `finishRefreshing(source)` is a no-op for the tab that does not own it — so the wallet stream reaching `isComplete` no longer speaks for a pull on the DeFi tab, whose visible rows are `defiAccounts` and come from `loadDeFiBalances` alone.

The DeFi side now clears on flow completion (`.onCompletion { cause -> if (cause == null) ... }`) instead of only in `.catch`, so a successful reload takes the spinner down too. A cancelled load deliberately does not clear: it was superseded by the load that replaced it, and that one clears in its turn.

Test `pulling on the DeFi tab keeps the spinner up until the DeFi list lands` drives exactly your scenario — the wallet stream emits `isComplete = true` while the DeFi flow is still open — and asserts `isRefreshing` is still true, then false once the DeFi flow closes. It fails on a revert of the fix. `pulling on the wallet tab still ends with the wallet stream` guards the other direction.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Real, fixed in cd59da089.

The `openAddChainAccount` race is worse than a race, it is deterministic in the wrong direction: `DeFiChainSelectionViewModel.saveSelection` calls `respond(REFRESH_CHAIN_DATA)` **before** `navigator.back()`, so the reload at `:761` always starts first and the resume then cancelled it mid-fetch and refetched from scratch.

A network DeFi read taken in the last 15s now stands in for a resume-triggered one, so the picker path fetches once and returning from `Receive` / the vault list no longer refetches on the way back. 15s is the interval iOS already throttles the same screen by — `DefiMainScreen.swift:84`, `.throttledOnAppear(interval: 15.0, action: refresh)`.

Only `isRefresh = true` starts the window: the cold-start load is cache-only and fetches nothing, so it must not stand in for a network read, and the first resume still upgrades that cached render.

Tradeoff worth being explicit about: a position change whose whole round trip finishes inside 15s of the last read will not re-read until the next resume, pull, or tab switch. That is the same characteristic iOS has, but say the word and I will shorten the window.

Tests: `resuming right after a read reuses it instead of fetching again` (fails on revert) and `resuming once the read has gone stale fetches again`.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
rkokhatskyi: Confirmed fixed, traced refreshOwner/finishRefreshing end to end including error paths and cross-tab interleaving

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
rkokhatskyi: Confirmed, and the ordering is deterministic not lucky: respond() completes the deferred while openAddChainAccount's awaiting continuation is Main.immediate, so it resumes inline and throttleDeFiRefresh runs before saveSelection's navigator.back() even executes

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -535,15 +576,23 @@ constructor(
...
                             .loadDeFiAddresses(vaultId, isRefresh)
                             .map { addresses -> addresses.sortByAccountsTotalFiatValue() }
                             .catch { error ->
-                                updateRefreshing(false)
                                 Timber.e(error, "Error loading DeFi balances for vault: $vaultId")
+                            }
+                            // This load, not the wallet stream, is what a pull on the DeFi tab is
+                            // waiting on. A cancelled one is superseded by the load that replaced
+                            // it, which clears the spinner in its turn.
+                            .onCompletion { cause ->
```
rkokhatskyi, line 609: `finishRefreshing(Defi)` in `.onCompletion` fires regardless of `isRefresh`, and `loadDeFiBalancesJob?.cancel()` at `:578` runs before that check, so an unrelated write through the shared `AppDataStore` (e.g. `toggleBalanceVisibility`) during a pull cancels the network fetch and clears the spinner off a cache-only read

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
johnnyluo: Real, fixed in 90a66148c.

Gated `finishRefreshing(Defi)` on `isRefresh` in the `.onCompletion` block, so a cache-only reload (e.g. one that races in via a shared AppDataStore write while a refresh's network fetch is still in flight) can no longer complete first and clear the spinner off the cache instead of the network read it's actually waiting on.

rkokhatskyi: Half fixed: the premature-clear is gone, but the gate now also excludes the cache-only replacement that races in, so neither the cancelled job nor its replacement ever calls finishRefreshing and the spinner sticks instead. Root cause is still collectLastOpenedVault's missing distinctUntilChanged, filed as a fresh comment below

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -535,15 +576,29 @@ constructor(
...
+                            }
+                            // This load, not the wallet stream, is what a pull on the DeFi tab is
+                            // waiting on. A cancelled one is superseded by the load that replaced
+                            // it, which clears the spinner in its turn. Gated on isRefresh so a
+                            // cache-only reload (e.g. one an unrelated AppDataStore write races in
+                            // while a refresh's network fetch is still in flight) can't complete
+                            // first and clear the spinner off the cache instead of the network
+                            // read.
+                            .onCompletion { cause ->
+                                if (cause == null && isRefresh) {
```
rkokhatskyi: `.onCompletion`'s new `isRefresh` gate also blocks the cache-only replacement that `loadDeFiBalancesJob?.cancel()` at `:578` lets an unrelated `AppDataStore` write race in, so neither the cancelled network job nor its replacement calls `finishRefreshing`, and the spinner sticks until a later pull or tab switch

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Real, fixed in 6caf20ff7 — and the root cause you named is where the fix went.

`AppDataStoreImpl.readData` maps `dataStore.data`, which re-emits the whole preferences snapshot on every write whatever key it touched, so `collectLastOpenedVault` took each one as a vault to load and re-ran `loadData` with the id already on screen. That is what lets a cache-only DeFi load cancel a pull's fetch at `:578`. The id is read `distinctUntilChanged()` now, so only an actual vault change reloads — which also stops every preference write from cancelling and restarting the wallet balance stream, not just the DeFi one.

Removing that churn broke something that was living off it: the backup warning was read once per `loadData` (`readBackupStatus(...).first()`), and backing the vault up on another screen only took the banner down because the write re-emitted the id. It is collected now, so it comes down on its own rather than waiting for an unrelated write.

I also kept the spinner honest independently of the trigger: a load that replaces the one an outstanding pull is waiting on carries that pull's network read forward (`readsNetwork = isRefresh || refreshOwner == Defi`), so the spinner is always owned by a load that actually fetches — never cleared off the cache, never left up. That path survives the `distinctUntilChanged`: pull on the DeFi tab, switch to the wallet tab, and its chain picker's `loadData` still cancels the fetch the pull is waiting on.

Tests, each verified to fail on a revert of its own fix:
- `an unrelated preference write does not reload the vault` — emits the same id twice from a SharedFlow (a StateFlow would conflate the repeat away before the screen saw it, which is not what DataStore does) and asserts one wallet load and one DeFi load.
- `a reload landing mid-pull still takes the spinner down` — the wallet-tab chain picker cancelling a DeFi pull's fetch; asserts the replacement fetches and the spinner comes down when it lands.
- `backing the vault up takes the warning down`.

10 tests in the file, full `:app` unit suite green.

johnnyluo (approved):
LGTM

Issue #5623, opened 2026-08-16
A vault whose passcode key is gone cannot be recovered in-app — the .vult re-import is refused as a duplicate


rkokhatskyi:
## Summary

`PasscodeState.KeyUnavailable` is a terminal state. The screen it shows offers no action, and the one recovery the copy points at — re-importing the vault's `.vult` — is refused as a duplicate. There is no in-app way out; today only reinstalling the app clears it. iOS closed the same gap with vultisig/vultisig-ios#5106 (an action on the recovery screen) and vultisig/vultisig-ios#5107 (an import that supersedes the unopenable row), both merged 2026-08-11.

Split out of #5566, which covers the copy only.

## Details

**The screen has no way out.** `PasscodeGuard.kt:77-81` renders `DeadEndScreen` for `KeyUnavailable`; the composable (`:139-165`) is two `Text`s and nothing else. It is hosted in a `Dialog` with `dismissOnBackPress = false` and `dismissOnClickOutside = false` (`:113-116`), over a `BackHandler(enabled = isGateClosed) {}` (`:48`) and a pointer-swallowing underlay (`:56`), so the nav host behind it is unreachable. A `.vult` opened from a file manager routes to `Route.ImportVault` and lands behind the same gate.

**Re-importing the backup is refused.** `VaultRepositoryImpl.toKeyShares` returns `emptyList()` rather than null or a throw when the data key is missing (`VaultRepository.kt:191-211`), so the orphaned row still resolves to a non-null `Vault`:

```kotlin
vaultRepository.getByEcdsa(vault.pubKeyECDSA)?.let {
    Timber.d("saveVault: vault already exists, updating")
    throw DuplicateVaultException()
}
```

`SaveVaultUseCase.kt:31-34`. `ImportFileViewModel.kt:131` turns that into `SaveResult.Duplicate` → `showDuplicateError()`, a hard error with no override offered. The only callers that pass `shouldOverrideVault = true` are keygen reshare and migrate (`KeygenViewModel.kt:902`).

[...]

Issue #5558, opened 2026-08-10
Never clear credentials on a read that merely failed — prove absence before any destructive clear


realpaaao:
Required by vultisig/vultisig-sdk#1838 invariant 2: **a read that fails must never be treated as a value that is absent, and nothing is deleted unless absence is proven.**

## The live bug (still on `main`)

`PasscodeStore.readCredentials()` — `data/.../passcode/PasscodeStore.kt:63-74`:

```kotlin
val salt = prefs.getString(KEY_SALT, null)?.let(::decodeOrNull)
val wrapped = prefs.getString(KEY_WRAPPED_KEY, null)?.let(::decodeOrNull)
if (salt == null || wrapped == null) {
    if (salt != null || wrapped != null) clearCredentials()   // deletes BOTH
    return null
}
```

`clearCredentials()` removes `passcode_salt` **and** `passcode_wrapped_data_key`.

## Why it fires on healthy data

`SecureSharedPreferences.decryptOrNull` (`:158-165`) catches **every** exception from Base64-decode + AES-GCM and returns null; `readTyped` (`:173-175`) and `getString` (`:189-191`) chain that null through to the caller default. So a value that is **intact on disk** but momentarily undecryptable — the class's own KDoc (`:42-46`) documents keystore-daemon stalls on certain Pixel/Samsung devices — is indistinguishable from "never written."

`readCredentials()` runs on **every cold start** (`PasscodeRepository.kt:248`) and **every unlock attempt** (`:444`). A transient fault on one of the two halves makes the app conclude "torn write" and permanently delete both.

`SharedPreferencesPasscodeStoreTest.kt:62-101` tests this as intended behaviour, but mocks `getString()` returning null directly — genuine on-disk corruption. It never models a spurious null injected by the encrypting decorator over a healthy value. That is why CI is green.

## Not the same as `recoverAndRetry`

`MainDataModule.kt:159-204` is a separate, **correctly gated** destructive path (whole-file + keystore-alias wipe, fires only on master-key-level exceptions at DI construction). Per-value decrypt failures never escape `decryptOrNull` to reach it. Only the `PasscodeStore` path is this bug.

## Fix

1. Let genuine `GeneralSecurityException` propagate from `SecureSharedPreferences` (or add a `getStringOrThrow`) so callers can tell **present-but-unreadable** from **never-written**.
2. Only run a destructive clear when `prefs.contains(key)` is provably false — never merely because a read returned null.
[...]

PR #5624, opened 2026-08-16
fix(passcode): correct the unreadable-keyshare copy and add a retry

### LINKED ISSUE #5566: When the passcode key is genuinely gone, tell the user to restore from .vult — not to reinstall
Under vultisig/vultisig-sdk#1838 we keep at-rest encryption, so the key can legitimately go missing. Android invalidates the keystore key when the device lock or biometrics change, and `MainDataModule.recoverAndRetry` (`:159-204`) then deletes the prefs file holding the salt and wrapped key. That path is correctly gated, and deleting is reasonable at that point.

What happens next isn't.

## What the user sees

`PasscodeState.KeyUnavailable` (`PasscodeRepository.kt:247-260`) shows copy telling them to **reinstall the app** and restore from a backup (`strings.xml:289-290`).

Reinstalling is the one action that destroys a vault whose data is still sitting on disk. If the state was reached wrongly — see #5558 — following that instruction is what causes the loss.

## What it should say

Your vaults are encrypted, the key is gone on this device, restore from your `.vult`. No reinstall. Don't clear anything.

## Fix

- Rewrite the `KeyUnavailable` copy to point at the `.vult`
- Only show it when the key is provably absent (#5558 covers not landing here by mistake)
- Check no other copy in the app tells a user to reinstall or clear data to fix a passcode problem

## Done when

- [ ] `KeyUnavailable` names the `.vult` as the recovery path
- [ ] No copy anywhere suggests reinstalling to fix a passcode problem
- [ ] The state only appears when the key is provably absent

rkokhatskyi:
Closes #5566.

## Summary

`PasscodeState.KeyUnavailable` told the user to reinstall the app. Reinstalling clears the Room database, which removes every vault on the device, not only the ones affected: `hasEncryptedKeyShares()` closes the gate app-wide off a single sealed row (`VaultDao.kt:59`), and plaintext shares can sit beside sealed ones (`VaultRepository.kt:222-223`, per #5553).

Neither unreadable state had an exit either. Back and outside taps belong to the gate, so the only way off the screen was a force stop.

## Changes

- Rewrote `passcode_key_unavailable_title` / `_message` and the sibling `passcode_store_unavailable_title` / `_message` across all 10 locales. The reinstall instruction is gone and the `.vult` backup is named as the recovery path. The sibling pair is included because the same composable renders it.
- `DeadEndScreen` now delegates to the shared `ErrorView` instead of two `Text` nodes, so both states get the standard hero, typography and layout.
- Added `PasscodeRepository.retry()` behind a **Try again** button. Both states are decided by one keystore read that `initialize()` latches for the process; `retry()` re-runs it. On `StoreUnavailable` that is the fix. On `KeyUnavailable` it also recovers a transient read fault, which `isPersistent()` cannot distinguish from a lost key because it is a DI type check rather than a readability check.
- Added `passcode_settings_encryption_explanation`, a caption under the Passcode Encryption toggle. The enable flow said nothing about the encryption being bound to one device. It renders outside the `if (state.isPasscodeEnabled)` block so it also shows before a passcode is set.
- `PasscodeSettingsScreen`'s body moves into a scrolling `Column` to host the caption, matching `NotificationsSettingsScreen`. Most of that file's diff is ktfmt re-indentation from the added nesting.

The German, Spanish, Croatian, Italian and Portuguese translations previously rendered "key shares" in the file-sharing sense; the rewrite corrects that. The same mistranslation remains in unrelated keys and is left alone.

Not included: `#5558`'s proven-absence criterion, closed NOT_PLANNED. The `.vult` import is still refused as a duplicate (`SaveVaultUseCase.kt:31-34`, tracked in #5623), so the copy names the backup without instructing an import.

## Before / After

| | Before | After |
|---|---|---|
| Key unavailable | before | after |
| Storage unavailable | before | after |
| Passcode settings | before | after |

## Testing

```
./gradlew :app:assembleDebug
./gradlew :app:testDebugUnitTest
./gradlew :data:testDebugUnitTest
./gradlew :app:lintDebug
```

All pass. `lintDebug` is what enforces `MissingTranslation` across the 10 locales.

New tests: `PasscodeRepositoryImplTest` covers `retry` re-reading a store that has come back and leaving an already-read state alone; `PasscodeSettingsExplanationTest` asserts the caption renders with the passcode both off and on.

### DIFF
--- a/app/src/androidTest/java/com/vultisig/wallet/ui/screens/passcode/PasscodeSettingsExplanationTest.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/ui/screens/passcode/PasscodeSettingsExplanationTest.kt
@@ -1,0 +1,56 @@
+package com.vultisig.wallet.ui.screens.passcode
+
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onNodeWithText
+import androidx.test.platform.app.InstrumentationRegistry
+import com.vultisig.wallet.R
+import com.vultisig.wallet.data.passcode.AutoLockTimeout
+import com.vultisig.wallet.ui.models.passcode.PasscodeSettingsUiModel
+import org.junit.Rule
+import org.junit.Test
+
+/**
+ * The rows under the switch are conditional on a passcode existing, so the caption sits one indent
+ * away from inheriting that condition, and it is worth most in the state the condition excludes.
+ */
+class PasscodeSettingsExplanationTest {
+
+    @get:Rule val compose = createComposeRule()
+
+    private val explanation =
+        InstrumentationRegistry.getInstrumentation()
+            .targetContext
+            .getString(R.string.passcode_settings_encryption_explanation)
+
+    @Test
+    fun theExplanationShowsBeforeAPasscodeIsSet() {
+        start(passcodeEnabled = false)
+
+        compose.onNodeWithText(explanation).assertIsDisplayed()
+    }
+
+    @Test
+    fun theExplanationShowsOnceAPasscodeIsSet() {
+        start(passcodeEnabled = true)
+
+        compose.onNodeWithText(explanation).assertIsDisplayed()
+    }
+
+    private fun start(passcodeEnabled: Boolean) {
+        compose.setContent {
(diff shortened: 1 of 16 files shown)


johnnyluo (approved):
LGTM

PR #5625, opened 2026-08-16
fix(defi): match the Kamino action buttons to their Figma component

aminsato:
## Summary

The Kamino Earn card's **Withdraw** button had no visible pill at all — an icon and a label floating on the card, with a stray hairline where its edge should have been. It was painted `backgrounds.tertiary` (#0B1A3A) on a card painted `backgrounds.secondary` (#061B3A), a difference the eye cannot find, and the icon's circle was set to the card colour exactly.

Both buttons now take their values from the two `DeFi Button` variants in Figma rather than from anything chosen here.

- Withdraw — [node 80871:75942](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=80871-75942)
- Deposit — [node 80871:75943](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=80871-75943)

## Changes

| Property | Figma | Before | After |
|---|---|---|---|
| Withdraw background | `backgrounds/surface-2` #11284A | `backgrounds.tertiary` #0B1A3A | `backgrounds.tertiary_2` |
| Withdraw border | 1dp `borders/extra-light` rgba(255,255,255,.03) | none | matches |
| Deposit background | `buttons/cta-(primary)` #0B4EFF | `primary.accent4` | `buttons.ctaPrimary` |
| Deposit border | 1dp `primary/accent-3` #2155DF | none | matches |
| Icon circle fill | rgba(255,255,255,.12) | `backgrounds.secondary` / `primary.accent3` | matches |
| Icon circle / glyph size | 34dp / 16dp | 30dp / 12dp | matches |
| Glyph | `circle-minus-filled`, `circle-plus-filled` | outline `circle_minus`, `circle_plus` | new filled drawables |
| Row spacing | — | 16dp | 8dp, as the sibling DeFi card |

The glyphs were the other half of the defect: the existing drawables are outlines — a stroked ring with a transparent centre — where the design calls for filled ones. Both are added from the exported vectors. The minus is translated by 1.6 because its path carries no inset of its own; the plus already spans 1.6..14.4 of its 16-unit box, which is the 10% inset the design applies.

### Scope of the shared component

The 34dp/16dp measurement lives in the shared `ActionButton`, so it is **passed by the Kamino cards rather than changed there**. The defaults stay exactly what every other DeFi screen was already drawing (30dp/12dp) and none of them move — the diff to `DeFiComponents.kt` is two parameters and their use.

Whether the rest of the DeFi buttons should follow the component is a real question, but it is answered on screens this change has no way to verify, so it is left alone. The other DeFi action row also still passes the outline `circle_minus`; same reasoning.

## Verification

Rendered on an emulator through `PreviewActivity`'s `solana_defi_earn` screen, before and after, with the same mock data. Both card shapes were checked: Steakhouse USDC, which has a position and shows both buttons, and Allez SOL, which has none and shows Deposit alone across the full width.

No screenshots are attached here by preference; the change is described by the value table above and is quickest to confirm by running that preview screen.

`:app:testDebugUnitTest`, `:app:lintDebug` and ktfmt pass.

## Summary by CodeRabbit

* **New Features**
  * Added refreshed deposit and withdrawal action buttons in Kamino Earn, including clearer filled icons, borders, spacing, and updated backgrounds.
  * Added new plus and minus circle icons for improved DeFi action controls.
  * Improved flexibility of action button sizing to support a more consistent visual experience.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
@@ -49,6 +49,7 @@ import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.IntOffset
 import androidx.compose.ui.unit.IntSize
 import androidx.compose.ui.unit.dp
@@ -271,6 +272,12 @@ fun ActionButton(
     contentColor: Color,
     iconCircleColor: Color,
     enabled: Boolean = true,
+    // The DeFi Button component measures these at 34dp and 16dp, which is what the Kamino cards
+    // pass. The defaults are the sizes every other caller was already drawing, left alone
+    // deliberately rather than corrected in passing: they reach screens this change has no business
+    // touching.
+    iconCircleSize: Dp = 30.dp,
+    iconSize: Dp = 12.dp,
     onClick: () -> Unit,
 ) {
     Button(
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

PR #5626, opened 2026-08-16
fix(defi): stop folding failed Kamino reads into zero

### LINKED ISSUE #5601: Kamino Earn: a failed read renders a real position as zero
## Summary

Kamino reads that fail are rendered as real zeros instead of "not read". A user with a live deposit can be shown a $0.00 Earn total, an available balance of 0 on the withdraw form, and a vault card with no Withdraw button, with no error shown. Introduced by #5588.

## Details

`KaminoPositionMath.kt:16-22` states the rule: a malformed number is missing, not zero. Three sites fold an unresolved value into zero anyway.

- `KaminoAmountViewModel.loadWithdrawState` swallows the `/metrics` failure at `:277`, leaving `rate` and `maximum` null, then publishes `eligibility = Withdrawable` with `available = maximum.toDecimalOrZero(...)` = 0.
- `KaminoEarnViewModel.buildRow:365-367` emits a row with `fiatValue = ... ?: BigDecimal.ZERO` when metrics are missing; the row counts as resolved and folds into `totalFiat` as zero.
- `KaminoEarnViewModel.kt:318-323` applies `?: BigDecimal.ZERO` to an unparseable `totalShares`, making `hasPosition` false and dropping the Withdraw button.

## Impact

No funds at risk — nothing is signed and the on-chain position is untouched. The user is told their deposit is worth nothing, and in the third case the only in-app route out of the vault disappears. `load()` runs once from `init` with no retry, so re-entering the screen is the only recovery.

## Suggested remediation

- Treat a null `rate` (or a null `maximum` on a non-empty position) as `Unreadable` plus a load error, and add a retry.
- Make `KaminoEarnRow.fiatValue` nullable, retain the previous row's value when a refresh cannot resolve one, and mark the total as partial when any row is unresolved.
- Drop the `?: BigDecimal.ZERO` on `shares` so an unreadable position keeps its Withdraw button.

johnnyluo:
## Summary
- `KaminoAmountViewModel.loadWithdrawState`: a real position whose rate could not be read (`/metrics` failure) now becomes `Unreadable` instead of `Withdrawable` with a zero maximum — a live deposit no longer reads as a $0 available balance.
- `KaminoEarnViewModel.buildRow`: an unparseable `totalShares` on a position entry that *is* present is now treated as unreadable (keeps the Withdraw button) rather than folded into the same zero as a confirmed empty position (absent entry).
- `KaminoEarnRow.fiatValue` is now nullable. A row whose fiat value fails to resolve on a refresh keeps its last known value instead of zeroing, and the screen-wide total is left at its last confirmed value rather than being recomputed short by an unresolved row.

## Issue
Closes #5601

## Test Plan
- [x] Added unit tests for all three sites: metrics outage on a real position, unparseable-but-present share count, partial-refresh total
- [x] `./gradlew :app:testDebugUnitTest --tests "com.vultisig.wallet.ui.models.solanastaking.KaminoAmountViewModelTest" --tests "com.vultisig.wallet.ui.screens.v2.defi.solana.KaminoEarnViewModelTest"`
- [x] `./gradlew lint`
- [ ] Manual verification on device/emulator against a live Kamino position

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Solana Kamino staking balance and withdrawal eligibility handling when vault pricing data is unavailable or invalid.
  - Prevented unavailable values from appearing as zero, avoiding understated portfolio totals.
  - Preserved previously displayed values during temporary pricing failures.
  - Correctly displays confirmed zero-balance positions and clears stale totals after a full withdrawal.
- **Tests**
  - Added coverage for unavailable metrics, partial pricing failures, zero balances, and stale total cleanup.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
@@ -337,9 +337,15 @@ constructor(
         // A real position whose rate could not be read is not "nothing to withdraw" — it is a
         // failed read, same as the positions call failing outright. Without this, a `/metrics`
         // outage would publish `Withdrawable` with a zero maximum, which reads as an emptied
-        // position rather than an unresolved one.
+        // position rather than an unresolved one. `rate == null` alone misses a `tokensPerShare`
+        // of `"0"`: `KaminoRate.parse` accepts it as a valid non-positive rate, so it must be
+        // caught via `!rate.isPositive` too, or `maximumTokens` rejecting it downstream leaves the
+        // same zero-maximum-with-no-warning fold this guard exists to close.
         val resolvedEligibility =
-            if (eligibility is KaminoWithdrawEligibility.Withdrawable && rate == null) {
+            if (
+                eligibility is KaminoWithdrawEligibility.Withdrawable &&
+                    (rate == null || !rate.isPositive)
+            ) {
                 KaminoWithdrawEligibility.Unreadable
             } else {
                 eligibility
--- a/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
@@ -334,6 +334,17 @@ constructor(
         val rate = KaminoRate.parse(metrics?.tokensPerShare)
         val held = eligibility.withdrawableShares?.maximum
 
+        // A real position whose rate could not be read is not "nothing to withdraw" — it is a
+        // failed read, same as the positions call failing outright. Without this, a `/metrics`
+        // outage would publish `Withdrawable` with a zero maximum, which reads as an emptied
+        // position rather than an unresolved one.
+        val resolvedEligibility =
+            if (eligibility is KaminoWithdrawEligibility.Withdrawable && rate == null) {
+                KaminoWithdrawEligibility.Unreadable
+            } else {
+                eligibility
+            }
+
         val maximum =
             if (held != null && rate != null) {
                 KaminoWithdrawMath.maximumTokens(held, rate, vault.tokenDecimals)
@@ -370,7 +381,7 @@ constructor(
                 ticker = coin.ticker,
                 available = maximum.toDecimalOrZero(vault.tokenDecimals),
                 minimum = minimum?.let { m -> BigDecimal(m.baseUnits).movePointLeft(m.decimals) },
-                eligibility = eligibility,
+                eligibility = resolvedEligibility,
             )
         }
     }
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
@@ -256,20 +256,36 @@ constructor(
 
                 val resolved = rows.filterNotNull()
                 _state.update { current ->
+                    // A row whose fiat value did not resolve this refresh keeps its last known
+                    // value rather than going blank or zero — same principle as keeping the whole
+                    // card standing on a failed vault call, one row at a time.
+                    val previousFiatByVault =
+                        current.rows.associate { it.vaultAddress to it.fiatValue }
+                    val merged =
+                        resolved.map { row ->
+                            if (row.fiatValue != null) row
+                            else
+                                previousFiatByVault[row.vaultAddress]?.let { fiatValue ->
+                                    row.copy(fiatValue = fiatValue)
+                                } ?: row
+                        }
                     current.copy(
                         isLoading = false,
                         // Keep the previous rows if every vault failed, rather than blanking
                         // positions the user holds because one refresh did not land — but only for
                         // vaults still enabled, or disabling one would leave its card on screen.
                         rows =
-                            resolved.ifEmpty {
+                            merged.ifEmpty {
                                 current.rows.filter { row -> row.vaultAddress in enabled }
                             },
+                        // Null when any row is still unresolved after the merge above: a total
+                        // short by one row's real value is not a smaller total, so the last
+                        // confirmed total is kept on screen instead.
                         totalFiat =
-                            resolved
+                            merged
                                 .takeIf { it.isNotEmpty() }
-                                ?.let { currencyFormat.format(totalFiatValue(it)) }
-                                ?: current.totalFiat,
+                                ?.let { totalFiatValue(it) }
+                                ?.let { currencyFormat.format(it) } ?: current.totalFiat,
                     )
                 }
             }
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -334,6 +334,17 @@ constructor(
         val rate = KaminoRate.parse(metrics?.tokensPerShare)
         val held = eligibility.withdrawableShares?.maximum
 
+        // A real position whose rate could not be read is not "nothing to withdraw" — it is a
+        // failed read, same as the positions call failing outright. Without this, a `/metrics`
+        // outage would publish `Withdrawable` with a zero maximum, which reads as an emptied
+        // position rather than an unresolved one.
+        val resolvedEligibility =
+            if (eligibility is KaminoWithdrawEligibility.Withdrawable && rate == null) {
```
rkokhatskyi: `resolvedEligibility` only checks `rate == null`, but `KaminoRate.parse` accepts `tokensPerShare = "0"` as a valid non-positive rate, so eligibility stays `Withdrawable` while `maximumTokens` rejects it via `!rate.isPositive`. That leaves `available = 0` with no unreadable message, the exact fold this PR set out to fix.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
```diff
@@ -316,12 +332,15 @@ constructor(
...
-                KaminoPositionMath.decimalOrNull(position?.totalShares) ?: BigDecimal.ZERO
-            } else {
-                null
+            when {
+                !positionsAreKnown -> null
+                position == null -> BigDecimal.ZERO
+                else -> KaminoPositionMath.decimalOrNull(position.totalShares)
             }
         val tokensPerShare = KaminoPositionMath.decimalOrNull(metrics?.tokensPerShare)
         val tokenAmount =
```
rkokhatskyi, line 410: `tokenAmount` requires `tokensPerShare` even when `shares` is a confirmed `BigDecimal.ZERO`, so a real withdrawal, whether `position == null` or `totalShares` parses to `"0"`, still resolves to null whenever `getVaultMetrics` fails that refresh, showing unavailable instead of a real zero deposit.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
```diff
@@ -256,20 +256,36 @@ constructor(
...
+                    // value rather than going blank or zero — same principle as keeping the whole
+                    // card standing on a failed vault call, one row at a time.
+                    val previousFiatByVault =
+                        current.rows.associate { it.vaultAddress to it.fiatValue }
+                    val merged =
+                        resolved.map { row ->
+                            if (row.fiatValue != null) row
+                            else
+                                previousFiatByVault[row.vaultAddress]?.let { fiatValue ->
+                                    row.copy(fiatValue = fiatValue)
```
rkokhatskyi, line 333: The merge only splices `fiatValue` back from the previous row, never `depositedDisplay`, so after a real withdrawal that confirms zero shares, a `getVaultMetrics` failure on that same refresh carries the prior fiat value into `totalFiat` while the card itself correctly renders empty, overstating the total for money the user no longer holds.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
johnnyluo: Confirmed — `KaminoRate.parse("0")` returns a valid non-null `KaminoRate(0, 0)`, so `rate == null` alone missed it. Fixed in 520d53d32 by escalating to `Unreadable` on `rate == null || !rate.isPositive`, plus a regression test with `tokensPerShare = "0"`.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
johnnyluo: Confirmed — a confirmed-zero share count still waited on `tokensPerShare`. Fixed in 520d53d32: `tokenAmount` now short-circuits to zero when `shares.signum() == 0`, independent of whether the metrics call succeeded that refresh, plus a regression test covering a real full withdrawal with a failed metrics call.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
johnnyluo: Traced this rather than patching the merge directly: `fiatValueOrNull` short-circuits to `BigDecimal.ZERO` (not null) when the amount is exactly zero. Once the sibling comment's fix makes a confirmed-zero-shares row resolve `tokenAmount`/`fiatValue` to a real zero regardless of `tokensPerShare`, the row's `fiatValue` is non-null again on that refresh, so `if (row.fiatValue != null) row` in the merge keeps the fresh zero instead of falling through to the stale-value splice — the exact scenario you described. Added a two-refresh regression test (funded → full withdrawal + failed metrics call) asserting the total drops to $0.00 rather than keeping the pre-withdrawal figure, in 520d53d32. Didn't find a residual case that still needs the merge itself touched — flag it if you had a different one in mind.

rkokhatskyi, line 333: The merge still only splices `fiatValue` back, not `depositedDisplay`/`depositedFiat`, so a still-held nonzero position whose `getVaultMetrics` call merely fails on an ordinary refresh shows unavailable on its own card while `totalFiat` keeps silently counting the spliced stale value for that same row.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/solana/KaminoEarnViewModel.kt
johnnyluo: Good catch on the general shape — but tracing it through, this exact scenario doesn't reproduce on the current code anymore.

The splice (`merged`) only ever feeds the `rows` list shown on each card; `totalFiat`/`totalValue` are computed separately from `resolved` (the raw, unspliced rows) via `resolvedTotal`, and only replace the previous total when *every* row resolves fresh together (`resolved.size == vaults.size` and none of their `fiatValue`s are null). If Steakhouse's metrics call fails on an ordinary refresh, `resolvedTotal` comes back null and the total falls back to `current.totalValue` — the last fully-confirmed figure — rather than being recomputed from a mix of this round's fresh rows and last round's spliced ones. So a card can show "Unavailable" while the total is simply frozen, never blended.

Added a regression test, `a total never blends a fresh row with another row's stale splice`, that exercises exactly the case you described: Allez grows to a genuinely new value while Steakhouse's metrics call fails, and asserts the total stays byte-for-byte at the prior confirmed figure rather than summing Allez's fresh number with Steakhouse's stale one.

rkokhatskyi (approved):
LGTM

PR #5627, opened 2026-08-17
fix(defi): pad Kamino deposit minimum past the program's check

### LINKED ISSUE #5600: Kamino deposit at exactly the advertised minimum is refused by the kVault program
## Summary

The Kamino deposit form shows and accepts the vault's published `minDepositAmount`, but the kVault program applies its minimum check to a slightly smaller derived amount and refuses. A user who reads "Minimum is 0.1 USDC" and types 0.1 runs a full two-device MPC ceremony that the RPC then rejects at preflight. Introduced by #5588.

## Details

`app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt:206-210` uses the published figure unchanged; it drives both the on-screen message (`KaminoAmountScreen.kt:94-99`) and the submit gate at `KaminoAmountViewModel.kt:382-384`, which is a strict `<`, so the advertised amount passes the form and gets signed.

The program does not check the instruction's amount. It subtracts `crank_funds_to_deposit` (allocated reserves x `crank_fund_fee_per_reserve`) from `max_amount` and checks the remainder against `min_deposit_amount`. All three allow-listed vaults fail at their advertised minimum with `Custom(7026)` (`DepositAmountBelowMinimum`) and succeed 6-10 base units higher. The shortfall tracks the vault's reserve count, which curators change, so a fixed constant is the wrong fix.

## Impact

The most natural first-time input — exactly the advertised minimum — always fails. Preflight is on, so nothing lands and no funds move; the cost is a wasted two-device ceremony and an opaque error right after the app said the amount was acceptable.

## Suggested remediation

- Add a margin to `minDepositAmount` before it reaches `state.minimum`, so the same adjusted figure is displayed and enforced.
- Match iOS (`KaminoService.swift:203-218`): proportional margin of one part in 1,000, rounded up, with a floor of 16 base units.
- Update `KaminoAmountViewModelTest.kt:154-160`, which pins the unadjusted value.

johnnyluo:
## Summary
- The kVault program subtracts crank funds (allocated reserves × `crank_fund_fee_per_reserve`) from a deposit before comparing it to `min_deposit_amount`, so depositing exactly the vault's advertised minimum always fails preflight (`Custom(7026) DepositAmountBelowMinimum`).
- Added `KaminoPositionMath.depositMinimumWithMargin`, which pads the published minimum by a proportional margin (1/1000, rounded up, floored at 16 base units) before it becomes both the on-screen figure and the submit gate — matching iOS's `KaminoService.swift` behavior. A fixed constant would drift since the shortfall tracks each vault's reserve count.
- `KaminoAmountViewModel` now calls the padded function instead of the raw `baseUnitsToDecimal`.

## Issue
Closes #5600

## Test Plan
- [x] `./gradlew :data:testDebugUnitTest --tests "*.KaminoPositionMathTest"` — added a case covering the proportional margin and the 16-unit floor
- [x] `./gradlew :app:testDebugUnitTest --tests "*.KaminoAmountViewModelTest"` — updated the pinned deposit-minimum test to the padded value
- [x] `./gradlew lint`

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated Kamino staking deposit minimums to include a safety margin, helping ensure deposits meet vault requirements.
  - Minimum amounts now account for token precision and small-value rounding.

- **Tests**
  - Added coverage for proportional margins, minimum margin handling, rounding, and missing minimum values.
  - Updated expected minimum deposit display from `0.1 USDC` to `0.1001 USDC`.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoPositionMath.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoPositionMath.kt
@@ -68,10 +68,13 @@ object KaminoPositionMath {
      * rounded up, floored at 16 base units.
      */
     fun depositMinimumWithMargin(baseUnits: String?, tokenDecimals: Int): BigDecimal? =
-        decimalOrNull(baseUnits)?.let { minimum ->
-            val margin = minimum.divide(BigDecimal(1000), 0, RoundingMode.UP).max(BigDecimal(16))
-            minimum.add(margin).movePointLeft(tokenDecimals)
-        }
+        decimalOrNull(baseUnits)
+            ?.takeIf { it.signum() >= 0 && it.remainder(BigDecimal.ONE).signum() == 0 }
+            ?.let { minimum ->
+                val margin =
+                    minimum.divide(BigDecimal(1000), 0, RoundingMode.UP).max(BigDecimal(16))
+                minimum.add(margin).movePointLeft(tokenDecimals)
+            }
 
     private val ONE_HUNDRED = BigDecimal(100)
 }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoPositionMath.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoPositionMath.kt
@@ -57,5 +57,21 @@ object KaminoPositionMath {
     fun baseUnitsToDecimal(baseUnits: String?, tokenDecimals: Int): BigDecimal? =
         decimalOrNull(baseUnits)?.movePointLeft(tokenDecimals)
 
+    /**
+     * A vault's published `minDepositAmount`, padded so the figure the form displays and enforces
+     * also clears the kVault program's check.
+     *
+     * The program subtracts crank funds (reserve count × fee) from the deposit before comparing it
+     * to `min_deposit_amount`, so depositing exactly the published minimum always falls short by a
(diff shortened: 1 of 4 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoPositionMath.kt
```diff
@@ -57,5 +57,21 @@ object KaminoPositionMath {
...
+     * A vault's published `minDepositAmount`, padded so the figure the form displays and enforces
+     * also clears the kVault program's check.
+     *
+     * The program subtracts crank funds (reserve count × fee) from the deposit before comparing it
+     * to `min_deposit_amount`, so depositing exactly the published minimum always falls short by a
+     * few base units. That shortfall tracks the vault's reserve count, which curators change, so a
+     * fixed constant would drift — a proportional margin does not. Matches iOS: one part in 1,000,
+     * rounded up, floored at 16 base units.
+     */
+    fun depositMinimumWithMargin(baseUnits: String?, tokenDecimals: Int): BigDecimal? =
```
rkokhatskyi, line 70: `depositMinimumWithMargin` has no positivity check, so a negative `minDepositAmount` yields a negative `state.minimum`, and `amount < minimum` at `KaminoAmountViewModel.kt:441` stays false for any positive amount, letting a doomed deposit run a full MPC ceremony before the kVault program rejects it at preflight.

rkokhatskyi (changes requested):

rkokhatskyi (approved):

PR #5628, opened 2026-08-17
fix(defi): read Kamino minWithdrawAmount as tokens, not shares

### LINKED ISSUE #5608: Kamino withdraw reads minWithdrawAmount as shares; the field is token-denominated
## Summary

The Kamino withdraw form reads the vault's `minWithdrawAmount` as share base units and projects it through `tokensPerShare` to get the minimum it displays and enforces. The field is token-denominated, so the minimum on the withdraw screen is off by the share rate. Introduced by #5588.

## Details

`app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt:288-300` builds a `KaminoShareAmount` from `minWithdrawAmount` and converts it to tokens via `KaminoWithdrawMath.minimumTokens`. The kVault program compares the field against a liquidity sum — `available_to_send_to_user + invested_liquidity_to_send_to_user <= vault.min_withdraw_amount` — so it is token base units, which is how iOS reads it (`KaminoService.swift:120-125`).

The resulting `state.minimum` is both displayed (`KaminoAmountScreen.kt:94-99`) and enforced as a hard refusal (`KaminoAmountViewModel.kt:382-383`, `BELOW_MINIMUM`). All three vaults publish `min_withdraw_amount = 1000`, and the real floor is about twice that in either reading, so the advertised minimum is roughly half the enforced one.

## Impact

The band is sub-cent today, so this is a wrong figure on screen plus a wasted MPC ceremony for anyone withdrawing dust. No funds are at risk. The error is rate-dependent and would flip to blocking legitimate withdraws on a vault whose `tokensPerShare` exceeds ~2.

## Suggested remediation

- Read `minWithdrawAmount` as token base units (`KaminoTokenAmount`) and derive the share figure from it, not the other way round.
- Apply a margin above the published figure, as iOS does with its 3x multiple; do not hard-code an observed boundary, it is a live-chain value.
- Update the tests that pin the share reading in `KaminoAmountViewModelTest.kt` and `KaminoWithdrawTest.kt`.

johnnyluo:
## Summary
- The kVault program checks `minWithdrawAmount` against the token value being withdrawn (`available_to_send_to_user + invested_liquidity_to_send_to_user`), not against a share count — but `KaminoAmountViewModel.loadWithdrawState` read it as SHARE base units, so the withdraw form's displayed and enforced minimum was off by the vault's share rate (about 930x on the SOL vault). Matches iOS's reading (`KaminoService.swift:120-125`).
- Added `KaminoWithdrawMath.effectiveMinimumShares`, which parses the field as `KaminoTokenAmount`, applies iOS's tested margin (3x the published figure, since the real on-chain floor sits above it and is a live value rather than a fixed boundary), and converts to the share count a withdraw has to name, rounded up via a new `KaminoTokenAmount.shareAmountRoundedUp`.
- `KaminoAmountViewModel` now reads `minWithdrawAmount` as tokens and calls the new function instead of constructing a `KaminoShareAmount` directly from the raw field.

## Issue
Closes #5608

## Test Plan
- [x] `./gradlew :data:testDebugUnitTest --tests "*.KaminoWithdrawTest"` — added a case covering the token reading, 3x margin, and round-up conversion for both USDC and SOL-shaped vaults
- [x] `./gradlew :app:testDebugUnitTest --tests "*.KaminoAmountViewModelTest"` — updated the two tests pinning the old share-denominated reading
- [x] `./gradlew lint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Kamino withdrawal minimum calculations to use token units.
  * Added appropriate rate conversion, safety margin, and upward rounding.
  * Ensured positions at the minimum threshold remain valid for withdrawal.
  * Improved handling of positions below the minimum to show accurate limits.

* **Documentation**
  * Clarified that published withdrawal minimums are specified in token base units.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
@@ -311,9 +311,10 @@ constructor(
     /**
      * Resolves what the withdraw form may offer, in shares first and tokens only as a projection.
      *
-     * Shares are the primary unit throughout: the endpoint takes shares, the vault's minimum is in
-     * shares, and a full withdraw must send the exact held share count. The token figure the form
-     * is denominated in is derived from those, never the other way round.
+     * Shares are the primary unit for the balance and the endpoint: the endpoint takes shares, and
+     * a full withdraw must send the exact held share count, so the token figure the form is
+     * denominated in is derived from that, never the other way round. The vault's *minimum*,
+     * though, is published in tokens — see [KaminoWithdrawMath.effectiveMinimumShares].
      */
     private suspend fun loadWithdrawState(vault: KaminoVault, coin: Coin) {
         val positions =
@@ -343,11 +344,22 @@ constructor(
 
         val vaultState =
             runCatchingCancellable { kaminoApi.getVaultState(vault.address) }.getOrNull()
-        // Share base units, not token — comparing it against a token amount would be wrong by the
-        // whole share rate (about 930x on the SOL vault).
-        val minimumShares =
+        // Token base units, not shares — the program compares this against the token value being
+        // withdrawn. Reading it as shares would be wrong by the whole share rate (about 930x on the
+        // SOL vault).
+        val publishedMinimum =
             vaultState?.state?.minWithdrawAmount?.let { raw ->
-                runCatching { KaminoShareAmount(BigInteger(raw), vault.sharesDecimals) }.getOrNull()
+                runCatching { KaminoTokenAmount(BigInteger(raw), vault.tokenDecimals) }.getOrNull()
+            }
+        val minimumShares =
+            if (publishedMinimum != null && rate != null) {
+                KaminoWithdrawMath.effectiveMinimumShares(
+                    publishedMinimum,
+                    rate,
+                    vault.sharesDecimals,
+                )
+            } else {
+                null
             }
         val minimum =
             if (minimumShares != null && rate != null) {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
@@ -362,8 +362,14 @@ constructor(
                 null
             }
         val minimum =
-            if (minimumShares != null && rate != null) {
-                KaminoWithdrawMath.minimumTokens(minimumShares, rate, vault.tokenDecimals)
+            if (held != null && minimumShares != null && maximum != null && rate != null) {
+                KaminoWithdrawMath.effectiveMinimumTokens(
+                    held,
+                    minimumShares,
+                    maximum,
+                    rate,
+                    vault.tokenDecimals,
+                )
             } else {
                 null
             }
--- a/app/src/test/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModelTest.kt
@@ -205,9 +205,10 @@ internal class KaminoAmountViewModelTest {
         // 999999999 share base units (one below the balance, stepping back from the API's sentinel)
         // × 1.0544278224860290217, truncated to the token's 6 decimals.
         assertEquals(0, BigDecimal("1054.427821").compareTo(state.available))
-        // minWithdrawAmount is 1000 SHARE base units, so in tokens it is 0.001055 rounded up — not
-        // the 0.001 a token-denominated reading would give.
-        assertEquals(0, BigDecimal("0.001055").compareTo(state.minimum!!))
+        // minWithdrawAmount is 1000 TOKEN base units (0.001 USDC). The program's real floor sits
+        // above that, so the margined figure is 3x it converted to shares and back: 0.003001, not
+        // the 0.001 the raw published amount would suggest.
+        assertEquals(0, BigDecimal("0.003001").compareTo(state.minimum!!))
     }
 
     @Test
@@ -394,9 +395,10 @@ internal class KaminoAmountViewModelTest {
     }
 
     @Test
-    fun `the withdraw minimum is read as shares, not as the token amount`() = runTest {
-        // minWithdrawAmount is 1000 SHARE base units. Treated as token base units it would be
-        // 0.001 USDC; converted properly at the Steakhouse rate it is 0.001055.
+    fun `the withdraw minimum is read as tokens, padded past the program's real floor`() = runTest {
+        // minWithdrawAmount is 1000 TOKEN base units — 0.001 USDC. Read as shares it would convert
+        // to 0.001055; the program's real floor sits above the published figure, so this pads it 3x
+        // (matching iOS) before converting to shares and back: 0.003001.
         givenTokenBalance("0")
         coEvery { kaminoApi.getUserPositions(WALLET) } returns
             listOf(
@@ -411,7 +413,7 @@ internal class KaminoAmountViewModelTest {
             KaminoVaultMetricsJson(tokensPerShare = "1.0544278224860290217")
 
         val minimum = viewModel(isWithdraw = true).state.value.minimum
-        assertEquals(0, BigDecimal("0.001055").compareTo(minimum!!))
+        assertEquals(0, BigDecimal("0.003001").compareTo(minimum!!))
     }
 
     @Test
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -343,11 +344,22 @@ constructor(
...
+        // withdrawn. Reading it as shares would be wrong by the whole share rate (about 930x on the
+        // SOL vault).
+        val publishedMinimum =
             vaultState?.state?.minWithdrawAmount?.let { raw ->
-                runCatching { KaminoShareAmount(BigInteger(raw), vault.sharesDecimals) }.getOrNull()
+                runCatching { KaminoTokenAmount(BigInteger(raw), vault.tokenDecimals) }.getOrNull()
+            }
+        val minimumShares =
+            if (publishedMinimum != null && rate != null) {
+                KaminoWithdrawMath.effectiveMinimumShares(
```
rkokhatskyi, line 356: `effectiveMinimumShares` ceils `state.minimum` twice while `state.available` stays floor-rounded, so at `held == minimumShares` (2846 shares on the USDC vault) `available` is 3000 base units but `minimum` is 3001, and no amount including Max clears both `submit()` checks. iOS caps the minimum at `available` and gates in share-space instead (`KaminoWithdrawViewModel.swift:169`, `KaminoWithdraw.swift:310`).

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -343,11 +344,22 @@ constructor(
...
         val vaultState =
             runCatchingCancellable { kaminoApi.getVaultState(vault.address) }.getOrNull()
-        // Share base units, not token — comparing it against a token amount would be wrong by the
-        // whole share rate (about 930x on the SOL vault).
-        val minimumShares =
+        // Token base units, not shares — the program compares this against the token value being
+        // withdrawn. Reading it as shares would be wrong by the whole share rate (about 930x on the
+        // SOL vault).
+        val publishedMinimum =
             vaultState?.state?.minWithdrawAmount?.let { raw ->
```
aminsato, line 351: The KDoc on the field this reads still asserts the opposite unit. `KaminoApi.kt:186-188` says:

> In the **share** base units, matching the withdraw endpoint's own denomination — not the token's. Comparing it against a token amount is wrong by the share rate.

That is now exactly backwards, and it sits on the model definition — the first place anyone touching this field will look. The unit of `minWithdrawAmount` is the entire subject of this PR and the response type is the one artifact that still documents the old reading, so the next change here is invited to undo it. Please update that comment (and the one on `minDepositAmount`, which is fine but reads as the contrast case) in the same PR.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -343,11 +344,22 @@ constructor(
...
+            if (publishedMinimum != null && rate != null) {
+                KaminoWithdrawMath.effectiveMinimumShares(
+                    publishedMinimum,
+                    rate,
+                    vault.sharesDecimals,
+                )
+            } else {
+                null
             }
         val minimum =
```
aminsato, line 364: `minimum` and `available` round in opposite directions, so a position sitting exactly on the effective minimum renders a minimum one base unit above its own maximum and becomes unwithdrawable through the app.

Steakhouse USDC, published `minWithdrawAmount` 1000, rate 1.0544278224860290217, and a spendable position of exactly 2846 share base units — the value `effectiveMinimumShares` returns here:

- `available` = `maximumTokens` = floor(2846 x 1.05442782...) = **3000** -> 0.003000
- `minimum` = `minimumTokens` = ceil(2846 x 1.05442782...) = **3001** -> 0.003001

`KaminoAmountScreen.kt:94-95` then refuses everything: any amount above 0.003000 is `exceeds_available`, and 0.003000 itself is `below_minimum`, so `canContinue` is never true. On chain the withdraw is valid — a Max sends the held 2846 shares verbatim (`sharesForTokens`'s exact-maximum branch), worth 3000 token units, which is exactly the 3x floor this PR targets.

iOS handles precisely this case in `KaminoWithdrawViewModel.minimumWithdraw`: when the position clears the minimum *in shares* but its token maximum lands below the rounded-up token minimum, it caps the displayed minimum at `maximumTokens`. Where the position genuinely holds less than the minimum, the real figure stands so the user can see what they are short of. Worth porting that clamp, or gating the submit on shares (`shares >= minimumShares`) the way iOS does rather than on the rounded-up token projection.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
aminsato, line 1: Outside this PR's diff, so file-level rather than inline: the deposit minimum at `KaminoAmountViewModel.kt:211` is still `minDepositAmount` read verbatim, which is the same class of bug this PR fixes on the withdraw side.

iOS applies `effectiveMinimumDeposit` — published + `max(ceil(published / 1000), 16)` base units — because the kVault program refuses a deposit *at* the published figure. Measured in `KaminoService.swift:180-218`: Steakhouse USDC refuses 100,007 and accepts 100,008 against a published 100,000; Allez SOL refuses 10,000,009 and accepts 10,000,010 against 10,000,000.

So a user who taps the displayed 0.1 USDC minimum passes the app gate at `KaminoAmountViewModel.kt:451-453` and gets `DepositAmountBelowMinimum` (custom error 7026) on chain.

Fine to leave for a follow-up — flagging it here since the withdraw fix lands in the same file.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
johnnyluo: Confirmed, and fixed in 5ff19d3b0.

Traced it through with your exact numbers: Steakhouse USDC, minWithdrawAmount 1000, rate 1.0544278224860290217, held == minimumShares == 2846. `maximumTokens` (floor) comes back 3000, `minimumTokens` (ceil) comes back 3001 — same product, opposite rounding, one base unit apart, and nothing between them for the form to offer.

Added `KaminoWithdrawMath.effectiveMinimumTokens`: clamps the displayed/gated minimum to `maximumTokens` whenever `held >= minimumShares` (the vault would accept the withdraw) but the raw minimum still landed above the raw maximum. Where the position genuinely holds fewer shares than the minimum, the unclamped figure stands so the user sees what they're short of — matching iOS's behavior as you described it, though implemented as a clamp on the token-space minimum rather than a share-space gate, since the screen's validation already runs entirely in token `BigDecimal`.

Covered by `a position sitting exactly on the effective minimum still has a submittable amount` (reproduces your 2846/3000/3001 numbers exactly) and `a position short of the effective minimum still shows the real figure`.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
johnnyluo: Fixed in 5ff19d3b0.

You're right — this now says exactly the opposite of what the response type meant, and it's the field the whole PR is about. Reworded to say token base units, matching `minDepositAmount`, with a note that reading it as shares (the previous claim) is wrong by the share rate — so a reader who lands here mid-debug gets the correction, not the bug.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
johnnyluo: Confirmed and fixed in 5ff19d3b0, together with the same report from @rkokhatskyi above — same root cause (ceil-rounded minimum vs. floor-rounded maximum on the same non-terminating product), same fix: `KaminoWithdrawMath.effectiveMinimumTokens` clamps the minimum to the maximum once `held >= minimumShares`, and leaves the real minimum in place when the position is genuinely short. Went with the clamp-in-token-space form rather than gating `canContinue` on shares directly, since `KaminoAmountScreen`'s validation is entirely token `BigDecimal` today and this keeps that single source of truth instead of introducing a second, share-space gate that could drift from it.

Test added reproduces your exact 2846/3000/3001 figures.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
johnnyluo: Agreed this is real and the same class of bug — but leaving it for a follow-up rather than pulling it into this PR, which is scoped to the withdraw-side unit/rounding fix. Will track separately rather than expand this diff.

rkokhatskyi (approved):
LGTM

PR #5629, opened 2026-08-17
fix(passcode): recover a vault whose keyshare key the device lost

### LINKED ISSUE #5623: A vault whose passcode key is gone cannot be recovered in-app — the .vult re-import is refused as a duplicate

rkokhatskyi:
`PasscodeState.KeyUnavailable` had no way out. The screen named the `.vult` backup as the way back but had no way to reach it, and re-importing that backup was refused as a duplicate — the orphaned row still resolves to a non-null `Vault`, so the import's collision check finds it, while its sealed keyshares hold every later launch in the same state.

The gate now shows a recovery screen whose primary action opens the file picker. It launches from inside the lock's own window rather than routing to `Route.ImportVault`, which is an ordinary `composable<>` and draws under the lock, behind its cover. Whatever the file holds — one share, or the archive "back up all vaults" writes — is stored on the spot, then the gate re-resolves. It opens once no sealed share is left, which takes a backup per vault.

`SupersedeUnopenableVaultUseCase` decides when the unopenable row may be replaced, porting iOS's `orphanedVault(collidingWith:in:)` clause for clause: the credentials are re-read and still gone (so a keystore that came back refuses before the delete, not after), exactly one stored vault collides on a public key, all of its shares are sealed, every identity it holds is declared identically by the backup, and the backup carries a share for every key either side names. `VaultDao.replace` then deletes and inserts in one `@Transaction`. The ordering row and the last-opened vault id are re-pointed with it — both name the row by id and neither is a foreign key.

Also refuses to store a keyshare that already carries the `vlpc1:` marker. Only the cipher writes it, so one arriving from a backup file would leave ciphertext no key opens, which every later launch reads as credentials gone — on a device that may never have had a passcode. The sealed-share clause reads that marker as proof, so it has to be unforgeable.

The parse-and-store half of the import moves into `StoreImportedVaultUseCase` so both screens share it, keeping the `libType` correction and the QBTC token re-add on the recovery path.

## Before / After

Same device, same vault, same state — a sealed vault whose keystore credentials were deleted.

| Before | After |
|--------|-------|
| before | after |

## Testing

- `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest` — 4988 tests, 0 failures.
- `./gradlew :app:lintDebug :data:lintDebug`, `./gradlew ktfmtFormat`.
- End to end on an API 36 emulator: imported a `.vult`, turned on passcode encryption, deleted the keystore-backed prefs file, relaunched into `KeyUnavailable`, restored. Verified the database then holds one vault row, zero `vlpc1:` rows, no stale ordering row, and a `last_opened_vault_id` matching the row that exists. Repeated with the same backup inside a `.zip`.
- Checked the guards by mutation: disabling `preserves`, the declared-superset clause, the MLDSA collision arm, or either `fileName` seam each fails at least one test.

Closes #5623

## Summary by CodeRabbit

* **New Features**
  * Added keyshare recovery from individual backup files and ZIP archives when passcode access is unavailable.
  * Added password prompts, progress indicators, retry actions, and clearer recovery error handling.
  * Improved imported-vault restoration, including legacy backup handling and token recovery.
  * Vault backups can now replace eligible unreadable vaults while preserving supported data.

* **Improvements**
  * Added localized recovery messaging across supported languages.
  * Prevented invalid encrypted keyshares from being stored.
  * Added loading states to error-screen actions.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
@@ -1,0 +1,214 @@
+package com.vultisig.wallet.ui.models.passcode
+
+import android.net.Uri
+import androidx.annotation.StringRes
+import androidx.compose.foundation.text.input.TextFieldState
+import androidx.compose.foundation.text.input.clearText
+import androidx.compose.runtime.Immutable
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.vultisig.wallet.R
+import com.vultisig.wallet.data.DefaultDispatcher
+import com.vultisig.wallet.data.common.AppZipEntry
+import com.vultisig.wallet.data.models.Vault
+import com.vultisig.wallet.data.passcode.PasscodeRepository
+import com.vultisig.wallet.data.passcode.PasscodeState
+import com.vultisig.wallet.data.usecases.DuplicateVaultException
+import com.vultisig.wallet.data.usecases.MalformedVaultException
+import com.vultisig.wallet.data.usecases.ParseVaultFromStringUseCase
+import com.vultisig.wallet.data.usecases.StoreImportedVaultUseCase
+import com.vultisig.wallet.data.usecases.WrongPasswordException
+import com.vultisig.wallet.data.usecases.file.UriFileReaderUseCase
+import com.vultisig.wallet.data.utils.safeLaunch
+import dagger.hilt.android.lifecycle.HiltViewModel
+import javax.inject.Inject
+import kotlin.coroutines.cancellation.CancellationException
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.withContext
+import timber.log.Timber
+
+@Immutable
+internal data class KeyshareRecoveryUiModel(
+    /**
+     * What the screen says below its title: the standing instruction until a backup has been tried,
+     * and what came of that attempt afterwards.
+     */
+    @StringRes val message: Int = R.string.passcode_key_unavailable_message,
+    val isRestoring: Boolean = false,
+    val isPasswordPromptVisible: Boolean = false,
+    val isPasswordObfuscated: Boolean = true,
+    @StringRes val passwordError: Int? = null,
+)
+
+/**
+ * Restores vaults from their backup file while the gate is closed over an unreadable keystore.
+ *
+ * The import cannot be routed to from here — the import screen draws in the window this one covers
+ * — so the file is read and stored on the spot, and the gate is asked to resolve again afterwards.
+ * It opens once no sealed keyshare is left on the device, which takes a backup per vault.
+ */
+@HiltViewModel
+internal class KeyshareRecoveryViewModel
+@Inject
+constructor(
+    private val uriFileReader: UriFileReaderUseCase,
+    private val parseVaultFromString: ParseVaultFromStringUseCase,
+    private val storeImportedVault: StoreImportedVaultUseCase,
+    private val passcodeRepository: PasscodeRepository,
+    @DefaultDispatcher private val defaultDispatcher: CoroutineDispatcher,
+) : ViewModel() {
+
+    private val _state = MutableStateFlow(KeyshareRecoveryUiModel())
+    val state: StateFlow<KeyshareRecoveryUiModel> = _state.asStateFlow()
+
+    val passwordTextFieldState = TextFieldState()
+
+    /** The shares the picked file holds, kept while a password for them is being asked for. */
+    private var picked: List<AppZipEntry> = emptyList()
+
+    fun onBackupPicked(uri: Uri?) {
+        uri ?: return
+        viewModelScope.safeLaunch {
+            val shares = read(uri)
+            if (shares.isEmpty()) {
+                report(R.string.import_file_not_supported)
+                return@safeLaunch
+            }
+            picked = shares
+            // Every share in an archive was exported together, under one password, and the prompt
+            // below opens on whatever the field holds.
+            passwordTextFieldState.clearText()
+            restore(password = null)
+        }
+    }
+
+    fun onPasswordEntered() {
+        if (picked.isEmpty()) return
+        viewModelScope.safeLaunch { restore(passwordTextFieldState.text.toString()) }
+    }
+
+    fun onPasswordPromptDismissed() {
+        _state.update { it.copy(isPasswordPromptVisible = false, passwordError = null) }
+    }
+
+    fun onPasswordVisibilityToggled() {
+        _state.update { it.copy(isPasswordObfuscated = !it.isPasswordObfuscated) }
+    }
+
+    fun onRetry() {
+        viewModelScope.safeLaunch { passcodeRepository.retry() }
+    }
+
+    /**
+     * The vault shares [uri] holds: the entries of a multi-vault archive, or the one file itself.
+     *
+     * An archive is what "back up all vaults" writes, and the gate needs a share per vault before
+     * it opens — so refusing one here would name the only file the user has as the wrong file. Each
+     * share keeps its own name, which is what tells a mislabelled older backup from a GG20 one; see
+     * [StoreImportedVaultUseCase].
+     */
+    private suspend fun read(uri: Uri): List<AppZipEntry> {
+        val shares =
+            if (uriFileReader.isValidZip(uri)) {
+                uriFileReader.extractZipEntries(uri).entries
+            } else {
+                val content = uriFileReader.readContent(uri) ?: return emptyList()
+                listOf(AppZipEntry(name = uriFileReader.readName(uri).orEmpty(), content = content))
+            }
+        return shares.filter { it.content.isNotBlank() }
+    }
+
+    /**
+     * Stores every share the picked file holds, and asks the gate to resolve again if any landed.
+     *
+     * One outcome for the whole file rather than one per share: an archive holds a share per vault,
+     * and the ones it carries for vaults this device never lost are duplicates by design.
+     */
+    private suspend fun restore(password: String?) {
+        // Deriving an address per default chain takes seconds. Without this, the second tap of a
+        // double tap starts a second restore of the same file, which finds the row the first one
+        // just wrote and reports a duplicate over its success. Read and set before this suspends,
+        // on a scope that dispatches to one thread, so the two cannot both pass.
+        if (state.value.isRestoring) return
+        _state.update {
+            it.copy(message = R.string.passcode_key_unavailable_message, isRestoring = true)
+        }
+        try {
+            var restored: Vault? = null
+            var refusal: Int? = null
+            for (share in picked) {
+                val vault =
+                    try {
+                        withContext(defaultDispatcher) {
+                            parseVaultFromString(share.content, password)
+                        }
+                    } catch (e: WrongPasswordException) {
+                        // Asked once for the whole file, since one password covers all of it.
+                        Timber.d(e, "Wrong or missing backup password")
+                        promptForPassword(isRetry = password != null)
+                        return
+                    } catch (e: MalformedVaultException) {
+                        Timber.w(e, "Backup share is malformed")
+                        refusal = refusal ?: R.string.import_file_not_supported
+                        continue
+                    }
+                when (val outcome = store(vault, share.name)) {
+                    null -> restored = vault
+                    else -> refusal = refusal ?: outcome
+                }
+            }
+            settle(restored, refusal)
+        } finally {
+            _state.update { it.copy(isRestoring = false) }
+        }
+    }
+
+    /** Stores [vault], answering null when it landed and the message to show when it did not. */
+    private suspend fun store(vault: Vault, fileName: String): Int? =
+        try {
+            storeImportedVault(vault, fileName)
+            null
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: DuplicateVaultException) {
+            Timber.d(e, "Backup share does not restore a vault this device is missing")
+            R.string.import_file_screen_duplicate_vault
+        } catch (e: Exception) {
+            Timber.e(e, "Restoring from a backup failed unexpectedly")
+            R.string.dialog_default_error_body
+        }
+
+    private suspend fun settle(restored: Vault?, @StringRes refusal: Int?) {
+        if (restored == null) {
+            report(refusal ?: R.string.import_file_not_supported)
+            return
+        }
+        picked = emptyList()
+        passcodeRepository.retry()
+        // Only the last vault's restore takes the gate down. Said after the state has settled, so
+        // it
+        // is never claimed over a gate that is already lifting.
+        if (passcodeRepository.state.value == PasscodeState.KeyUnavailable) {
+            report(R.string.passcode_key_unavailable_restored)
+        }
+    }
+
+    private fun promptForPassword(isRetry: Boolean) {
+        _state.update {
+            it.copy(
+                isPasswordPromptVisible = true,
+                passwordError = R.string.import_file_screen_password_error.takeIf { _ -> isRetry },
+            )
+        }
+    }
+
+    private fun report(@StringRes message: Int) {
+        _state.update {
+            it.copy(message = message, isPasswordPromptVisible = false, passwordError = null)
+        }
+    }
+}
--- a/app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
@@ -9,6 +9,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.vultisig.wallet.R
 import com.vultisig.wallet.data.DefaultDispatcher
+import com.vultisig.wallet.data.common.AppZipContents
 import com.vultisig.wallet.data.common.AppZipEntry
 import com.vultisig.wallet.data.models.Vault
 import com.vultisig.wallet.data.passcode.PasscodeRepository
@@ -33,10 +34,7 @@ import timber.log.Timber
 
 @Immutable
 internal data class KeyshareRecoveryUiModel(
-    /**
-     * What the screen says below its title: the standing instruction until a backup has been tried,
-     * and what came of that attempt afterwards.
-     */
+    /** The standing instruction until a backup has been tried, then what came of that attempt. */
     @StringRes val message: Int = R.string.passcode_key_unavailable_message,
     val isRestoring: Boolean = false,
     val isPasswordPromptVisible: Boolean = false,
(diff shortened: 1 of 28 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
```diff
@@ -0,0 +1,214 @@
...
+                val vault =
+                    try {
+                        withContext(defaultDispatcher) {
+                            parseVaultFromString(share.content, password)
+                        }
+                    } catch (e: WrongPasswordException) {
+                        // Asked once for the whole file, since one password covers all of it.
+                        Timber.d(e, "Wrong or missing backup password")
+                        promptForPassword(isRetry = password != null)
+                        return
```
aminsato, line 145: A `WrongPasswordException` here returns from `restore()` without ever reaching `settle()`, so shares that already landed earlier in this same loop are never settled: `picked` is not cleared and `passcodeRepository.retry()` is never called.

Concrete case: a "back up all vaults" archive where share 1 is unencrypted and restores the device's only sealed vault, and share 2 is password-protected. Share 1 supersedes the orphaned row, share 2 throws here, the password sheet opens for a backup the user may not have the password for. They dismiss it — and the gate is still closed, still showing `passcode_key_unavailable_message` ("Restore from your .vult backup"), even though the restore that was needed already succeeded and a single `retry()` would have lifted it. The only way out is guessing that "Try again" now works.

Settling what has landed before prompting (or prompting without abandoning the pass) would avoid stranding the user on a gate that is ready to open.

app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
```diff
@@ -0,0 +1,214 @@
...
+    private suspend fun store(vault: Vault, fileName: String): Int? =
+        try {
+            storeImportedVault(vault, fileName)
+            null
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: DuplicateVaultException) {
+            Timber.d(e, "Backup share does not restore a vault this device is missing")
+            R.string.import_file_screen_duplicate_vault
+        } catch (e: Exception) {
```
aminsato, line 177: `store()` treats "threw" as "did not land", but `saveVault` can throw *after* `SupersedeUnopenableVaultUseCase` has already committed the replacement. Once `supersede` returns true the vault row is deleted and the backup inserted; execution then continues into the default-coins block of `SaveVaultUseCaseImpl`, which does `vaultRepository.get(...)`, `tokenRepository.nativeTokens.first()` and a `chainAccountAddressRepository.getAddress()` per default chain — any of which can throw.

When it does: `outcome != null`, so `restored` stays null, `settle()` takes the `restored == null` branch, reports `dialog_default_error_body`, and — the part that matters — never calls `passcodeRepository.retry()`. The user is told the restore failed while their vault was in fact replaced, the superseded row is gone, and the gate stays closed even though it would now resolve.

Same class of problem the KDoc on `StoreImportedVaultUseCase` already guards against for the post-save steps; the derivation work inside `saveVault` is not covered by it.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/SupersedeUnopenableVaultUseCase.kt
```diff
@@ -0,0 +1,142 @@
...
+    private val keyShareCipher: KeyShareCipher,
+    private val passcodeRepository: PasscodeRepository,
+    @IoDispatcher private val dispatcher: CoroutineDispatcher,
+) : SupersedeUnopenableVaultUseCase {
+
+    override suspend fun invoke(backup: Vault): Boolean =
+        withContext(dispatcher) {
+            val superseded = supersededBy(backup) ?: return@withContext false
+
+            vaultRepository.replace(superseded.id, backup)
```
aminsato, line 49: `vaultRepository.replace` is atomic on its own, but the two pointer fix-ups that follow it are separate suspending writes on a cancellable scope, so the invariant the comment describes is only held if nothing interrupts lines 50-58.

The caller is `viewModelScope.safeLaunch` in `KeyshareRecoveryViewModel`; a cancellation (or process death) landing after `replace` has committed but before `setLastOpenedVaultId` leaves `last_opened_vault_id` pointing at the row that was just deleted. `LastOpenedVaultRepositoryImpl` only falls back to `getAll().firstOrNull()` when the stored value is **null**, not when it dangles — so the stored-but-deleted id is returned as-is, which is exactly the "screen behind the gate stays bound to a vault that no longer exists" state this block exists to prevent.

Wrapping the post-`replace` writes in `withContext(NonCancellable)` closes the cancellation half of that window.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/SupersedeUnopenableVaultUseCase.kt
```diff
@@ -0,0 +1,142 @@
...
+            // Two pointers to the row that just went, neither of them a foreign key. The ordering
+            // row would outlive it — the delete flow removes it by hand for the same reason, and
+            // the replacement is given its place in the list the way any new vault is. The
+            // last-opened id is what the home screen resolves the vault it shows from, and it is
+            // read once: left dangling, the screen behind the gate stays bound to a vault that no
+            // longer exists, and sends and swaps route with its id.
+            vaultOrderRepository.delete(parentId = null, name = superseded.id)
+            lastOpenedVaultRepository.setLastOpenedVaultId(backup.id)
```
aminsato: `setLastOpenedVaultId` runs unconditionally, but the justification above it only applies when the superseded vault *was* the last-opened one.

Scenario: the device holds vault B (intact, plaintext shares, the one the user last opened) and vault A (sealed, unopenable). The user restores A from its backup. `last_opened_vault_id` is silently repointed from B to the restored A, so the home screen opens A once the gate lifts — a vault switch the user never asked for, on a device where nothing about B changed.

Guarding on `lastOpenedVaultRepository.lastOpenedVaultId.first() == superseded.id` keeps the dangling-pointer fix while leaving an unrelated choice alone.

app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
```diff
@@ -0,0 +1,214 @@
...
+     *
+     * An archive is what "back up all vaults" writes, and the gate needs a share per vault before
+     * it opens — so refusing one here would name the only file the user has as the wrong file. Each
+     * share keeps its own name, which is what tells a mislabelled older backup from a GG20 one; see
+     * [StoreImportedVaultUseCase].
+     */
+    private suspend fun read(uri: Uri): List<AppZipEntry> {
+        val shares =
+            if (uriFileReader.isValidZip(uri)) {
+                uriFileReader.extractZipEntries(uri).entries
```
aminsato: `AppZipContents.isComplete` is dropped here. `Uri.processZip` stops early on `MAX_IMPORT_ARCHIVE_ENTRIES`, the inflate cap, or a read error and reports that through `isComplete` — `ImportFileViewModel.acceptZip` surfaces it as `import_file_zip_incomplete` for exactly this reason.

On this screen a truncated archive instead restores whatever subset was extracted and then says `passcode_key_unavailable_restored` — "Another vault on this device still needs its backup" — which sends the user hunting for a backup file they are in fact already holding, when the real cause is that the app stopped reading their archive. Worth reporting the incomplete extraction rather than silently restoring a partial list, given the gate only lifts once every sealed vault has been replaced.

app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
```diff
@@ -0,0 +1,214 @@
...
+                    } catch (e: WrongPasswordException) {
+                        // Asked once for the whole file, since one password covers all of it.
+                        Timber.d(e, "Wrong or missing backup password")
+                        promptForPassword(isRetry = password != null)
+                        return
+                    } catch (e: MalformedVaultException) {
+                        Timber.w(e, "Backup share is malformed")
+                        refusal = refusal ?: R.string.import_file_not_supported
+                        continue
+                    }
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed, though by restructuring rather than by settling early: the loop now parses every share before it stores any, so the password prompt is reached before anything has landed and there is no partial state to strand. Same two-phase shape iOS's ProtectedVaultImporter uses, and a new test pins that a mixed archive stores nothing until the password is given.

app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
```diff
@@ -0,0 +1,214 @@
...
+    private suspend fun read(uri: Uri): List<AppZipEntry> {
+        val shares =
+            if (uriFileReader.isValidZip(uri)) {
+                uriFileReader.extractZipEntries(uri).entries
+            } else {
+                val content = uriFileReader.readContent(uri) ?: return emptyList()
+                listOf(AppZipEntry(name = uriFileReader.readName(uri).orEmpty(), content = content))
+            }
+        return shares.filter { it.content.isNotBlank() }
+    }
```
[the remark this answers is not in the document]
rkokhatskyi: Fixed. `read` now returns `AppZipContents` rather than just its entries, and a truncated archive reports `import_file_zip_incomplete` — including when part of it did restore, since that is the actionable cause and "another vault still needs its backup" would send the user looking for a file they already hold.

data/src/main/kotlin/com/vultisig/wallet/data/db/dao/VaultDao.kt
```diff
@@ -93,6 +99,19 @@ interface VaultDao {
...
+     * The delete cascades every row belonging to the vault it removes, keyshares included. Both
+     * statements land together or neither does, so process death part-way cannot leave the device
+     * holding neither the vault it took away nor the one that replaces it.
+     */
+    @Transaction
+    suspend fun replace(supersededVaultId: String, vault: VaultWithKeySharesAndTokens) {
+        delete(supersededVaultId)
+        insert(vault)
+    }
+
```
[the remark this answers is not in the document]
rkokhatskyi: Declining: `pending_limit_order` has no foreign key on `main` either, so `ConfirmDeleteViewModel.delete()` already leaks the identical rows for every vault deletion — this is not something the replacement introduces. The rows are also inert: every query on that table is vault-scoped (`getOpenOrders(vaultId)`, `observeOrders(vaultId)`), so rows for an id that no longer exists are never read. Adding the cascade means an `@Entity` change and a Room migration for a table this PR does not otherwise touch, which is worth its own change.

app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
rkokhatskyi: Good catch, fixed. Rather than settling early I split the loop in two: every share is parsed first, and only once they all parse is anything stored. The password prompt is now reached before any share has landed, so there is nothing to strand — and it matches what iOS's ProtectedVaultImporter does for the same reason. Covered by a test that a mixed archive stores nothing until the password is given.

app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
rkokhatskyi: Confirmed and fixed. `SaveVaultUseCaseImpl` does continue into the default-coins block after the supersede has committed, and `getAddress` per default chain can throw there, so "threw" really did not mean "did not land". `settle` now asks `passcodeRepository.retry()` whatever the outcome — if the row was in fact replaced the gate lifts and the error is never seen, and if nothing landed the re-read is the same no-op the Try again button performs. New test pins the retry on the throwing path.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/SupersedeUnopenableVaultUseCase.kt
rkokhatskyi: Fair, wrapped both in `withContext(NonCancellable)`. The row is already deleted by the time they run, so they are not the caller's to cancel — same reasoning `PasscodeRepositoryImpl` uses around its own paired writes. Worth noting the consequence is narrower than a dangling read: `VaultAccountsViewModel.collectLastOpenedVault` does fall back to `getAll().firstOrNull()` when `get(id)` comes back null, so the harm was the missing DataStore emission leaving `loadData` never re-run, not the id itself being followed.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/SupersedeUnopenableVaultUseCase.kt
rkokhatskyi: Keeping this one unconditional. The ordinary import already does exactly the same thing — `ImportFileViewModel.navigateToHome` routes to `Route.Home(openVaultId = vault.id)`, which `VaultAccountsViewModel.updateLastOpenedVault` turns into `setLastOpenedVaultId` — so landing on the vault you just restored is the app's existing behaviour for restoring a vault from a file, not a switch this path invents. Guarding on the superseded id would also make the write conditional on the one thing it exists to fix, for a preference the user is about to see the result of either way.

app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
rkokhatskyi: Fixed. `read` now returns `AppZipContents` rather than just its entries, and `settle` reports `import_file_zip_incomplete` whenever the archive was truncated — including when part of it did restore, since that is the actionable cause and "another vault still needs its backup" would send the user looking for a file they are already holding. Two tests cover it: a truncated archive with no usable entries, and one where a share did land.

app/src/main/java/com/vultisig/wallet/ui/models/passcode/KeyshareRecoveryViewModel.kt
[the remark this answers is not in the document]
rkokhatskyi: The revision you checked predates the fix — it is in 3e1c01045, pushed after this thread. `restore` now parses every share into a `parsed` list first and only stores once the whole file has parsed, so the `WrongPasswordException` return happens before anything can land. The test is `a share needing a password stores nothing until it is given`, which asserts `coVerify(exactly = 0) { storeImportedVault(any(), any()) }` and no `retry()` for a two-entry archive whose second share needs a password.

johnnyluo (approved):
LGTM

Issue #5585, opened 2026-08-11
[Add] ybRUNE staked-position card with NAV-based value [DeFi]


realpaaao:
## Problem
`ThorchainDefiPositionsViewModel.kt` builds bond, RUJI, TCY and LP positions — no bRUNE/ybRUNE card. Holders of ybRUNE see their bonded position nowhere.

## Spec
Add a **"Compounded bRUNE"** staking card to the THORChain DeFi tab:

- **Amount**: vault's ybRUNE (`x/staking-x/brune`) bank balance
- **Value**: balance × NAV × RUNE price. NAV = `liquid_bond_size / liquid_bond_shares` from the contract's `{"status":{}}` query — already fetched for ybRUNE pricing (`TokenPriceRepository.kt`, `BRUNE_STAKING_CONTRACT`); reuse it
- **Stake / Unstake** route into the liquid-bond wasm flow (#5584)
- **Send** moves the ybRUNE receipt token, not liquid bRUNE

Depends on #5583 (hide receipt from wallet) and #5584 (flow).

## Refs
- Windows: `bruneStakeService.ts` + `stakeUiResolver.ts` (vultisig/vultisig-windows#4407)
- iOS: `THORChainStakeInteractor.swift:154` (vultisig/vultisig-ios#4778)

PR #5635, opened 2026-08-18
fix(wallet): hide the ybRUNE receipt from the token list

### LINKED ISSUE #5583: [Fix] Hide ybRUNE from the wallet token list — it is a DeFi receipt [Wallet]
## Problem
ybRUNE (`x/staking-x/brune`) is the auto-compounding receipt for bonded bRUNE — a DeFi position, not a spendable wallet token. iOS and Windows hide it from the wallet list (iOS `defiOnlyTickers`, `CoinExtension.swift:17`; Windows `isBondedRuneReceiptCoin.ts`). Android shows it as a normal token.

## Cause
`TokenRepository.kt:291`:

```kotlin
internal val DEFI_ONLY_THORCHAIN_DENOMS = setOf("x/staking-ruji", "x/staking-x/ruji")
```

`x/staking-x/brune` is missing, so the filter at `TokenRepository.kt:99` never drops it.

## Fix
Add `x/staking-x/brune` to the set. `TokenRefreshWorker.kt:100` filters on the same set, so the receipt also stops counting toward wallet totals once the DeFi position card (separate issue, linked below) surfaces its value.

aminsato:
## Summary

ybRUNE (`x/staking-x/brune`) is the auto-compounding receipt for bonded bRUNE — a DeFi position, not a spendable wallet token. iOS (`defiOnlyTickers`) and Windows (`isBondedRuneReceiptCoin.ts`) keep it out of the wallet list; Android surfaced it as a normal token.

Two surfaces offered it, so both are closed:

**Balance discovery.** Added `x/staking-x/brune` to `DEFI_ONLY_THORCHAIN_DENOMS`, which only covered the sRUJI receipt. THORChain discovery now drops it, and `TokenRefreshWorker.cleanupDeFiOnlyTokens` deletes it from vaults that picked it up while it was still discoverable. Both gates compare the lowercased denom now — the DeFi-only filter was exact-match while the auto-discover gate directly below it already lowercased, and for this denom the asymmetry mattered, since the canonicalization override exists precisely because node casing drifts.

**The token-selection sheet.** The picker reads the wallet catalogue rather than discovery, so ybRUNE was still offered there and could be enabled by hand straight back into the wallet list. It now lives in `Coins.defiOnly` alongside the sRUJI receipt: out of `Coins.coins` (so neither the picker nor discovery offers it), still resolvable by id, and still hung off the DeFi addresses so the position card in #5585 has something to read. NAV pricing follows it out of the wallet catalogue — `fetchThorContractPrices` scans `allResolvable`, since the contract route is the only one that can price ybRUNE.

Also dropped two ybRUNE paths the new discovery filter makes unreachable: its entry in `AUTO_DISCOVER_THORCHAIN_DENOMS` and its branch in the canonicalization `when`.

Closes #5583.

## Notes

- The receipt's value leaves the wallet total until the position card in #5585 surfaces it. That split is what #5583 describes.
- The same leak exists in the macOS/iOS token-selection sheet: vultisig/vultisig-ios#5194.

## Test plan

- `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest` — passes. The one test change is a removal: `getTokensWithBalance canonicalizes ybRUNE decimal to the curated value` asserted the receipt surfaces as a wallet token, which is exactly the behaviour this PR takes away.
- Manual: THORChain → Select tokens no longer lists ybRUNE (searching `yb` returns nothing), the chain list no longer shows it, and an entry persisted before this change is removed on the next token refresh.

## Summary by CodeRabbit

- **New Features**
  - Added support for recognizing and pricing THORChain’s ybRUNE as a DeFi-only receipt token.
  - Included ybRUNE alongside other supported DeFi receipt positions while keeping it out of standard wallet token lists and automatic token discovery.
- **Bug Fixes**
  - Improved THORChain token matching to work reliably regardless of capitalization.
  - Updated liquid-bonding handling to focus on supported bRUNE tokens and prevent ybRUNE from appearing as a regular wallet asset.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/models/Coins.kt
@@ -4136,6 +4136,8 @@ object Coins {
                 isNativeToken = false,
             )
 
+        // The auto-compounding receipt for bonded bRUNE — a DeFi position, never a wallet token,
+        // so it is kept out of [all] and reached through [defiOnly] instead.
         val ybRUNE =
             Coin(
                 chain = Chain.ThorChain,
@@ -4150,22 +4152,7 @@ object Coins {
             )
 
         val all =
-            listOf(
-                RUNE,
-                TCY,
-                RUJI,
-                KUJI,
-                FUZN,
-                NSTK,
-                WINK,
-                LVN,
-                RKUJI,
-                sTCY,
-                yRUNE,
-                yTCY,
-                bRUNE,
-                ybRUNE,
-            )
+            listOf(RUNE, TCY, RUJI, KUJI, FUZN, NSTK, WINK, LVN, RKUJI, sTCY, yRUNE, yTCY, bRUNE)
     }
 
     object Ton {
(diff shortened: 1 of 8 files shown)


johnnyluo (approved):
LGTM

PR #5636, opened 2026-08-18
fix(defi): drop the inner bevel that drew a line across action buttons

aminsato:
## Summary

Every DeFi action button had a horizontal line drawn through it.

The cause was `Modifier.actionButtonInnerBevel` in `DeFiComponents.kt`: it painted a 1dp white-10% strip along the button's top edge and a 1dp `#0F1C3E` strip along the bottom, clipped to the pill. Across the flat top and bottom of the pill those strips run edge to edge, so they read as a stray line drawn through the button rather than as a bevel.

- Removed the bevel from `ActionButton` — Kamino Withdraw/Deposit, Bond/Unbond, and every other DeFi action button.
- Removed the same effect from the Tron Freeze/Unfreeze buttons (`tronFreezeButtonBevel`), where it was drawn as a gradient perimeter stroke, so the DeFi buttons stay consistent.
- Dropped the now-unused colour constants and imports in both files.

No layout, size, colour, or border change — only the overlay strips are gone.

## Before / After

Solana DeFi — Earn tab (the reported case: Withdraw / Deposit)

| Before | After |
|--------|-------|
| before | after |

## Test plan

- [x] `:app:compileDebugKotlin` passes
- [x] Real Android renders on emulator (API 33) via `PreviewActivity`, before and after, for the three Solana DeFi previews above
- [ ] Tron DeFi positions screen — Freeze/Unfreeze buttons render without the line

## Summary by CodeRabbit

* **Style**
  * Simplified DeFi action button visuals by removing custom bevel effects.
  * Simplified Tron freeze action button visuals by removing gradient bevel effects.
  * Preserved existing button sizing, behavior, and remaining styling.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiComponents.kt
@@ -33,16 +33,9 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
-import androidx.compose.ui.draw.drawWithContent
-import androidx.compose.ui.geometry.CornerRadius
 import androidx.compose.ui.geometry.Offset
-import androidx.compose.ui.geometry.Rect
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5637, opened 2026-08-18
THORChain LP: surface pending half-deposits, show both remove legs

### LINKED ISSUE #4331: THORChain LP: surface pending half-deposits and refund timeout in LP UI
## Problem

When a user performs a symmetric add-liquidity on THORChain (memo `+:ASSET:<paired-address>`) by sending the RUNE side first, THORChain holds that deposit as **pending liquidity** until the matching asset-side deposit arrives from the paired address. During this window:

- No LP units are minted.
- No outbound is created.
- The tx shows as "pending" on block explorers (e.g. runescan).
- After `PendingLiquidityAgeLimit` (~17,280 blocks / ~1 day on default mimir), THORChain auto-refunds the deposited side.

Today the Vultisig LP UI does not surface this state. A user who deposits one side and then opens the app sees nothing actionable — no "waiting for matching deposit" indicator, no refund countdown, no CTA to complete the second side.

### Real example

Tx: `492BE81A5053E6FA7B410E25BE69FD0EDD24BCCD3DD75381671EC785D0C493F1`
- Memo: `+:BASE.VVV-0xACFE6019ED1A7DC6F7B508C02D1B04EC88CC21BF:0x14F6Ed6CBb27b607b0E2A48551A988F1a19c89B6`
- 2 RUNE deposited, observed and finalized.
- Pool `BASE.VVV` shows `pending_inbound_rune: 400000000` (this user's 2 RUNE included).
- LP record under paired asset address `0x14F6…89B6` shows `units: 0`, `pending_rune: 0` — i.e. no position opened yet.
- Matching BASE.VVV deposit from `0x14F6…89B6` has not been observed. RUNE will be refunded if it doesn't arrive in time.

## Proposed UX

In the LP screen / position list:

1. **Pending half-deposit state**: when the user's recent LP deposit corresponds to a pool-level `pending_inbound_*` and there is no matching LP record yet, show a distinct "Waiting for matching `<ASSET>` deposit from `<address>`" card.
2. **Refund countdown**: estimate time-to-refund based on `PendingLiquidityAgeLimit` mimir value and current block height, and show it on the pending card.
3. **CTA to complete the second side**: deep-link / pre-fill the deposit flow with the correct asset, amount (based on current pool ratio for clean symmetric add), inbound vault address, and memo.
4. **Detection sources**:
   - `/thorchain/pool/{asset}` → `pending_inbound_rune` / `pending_inbound_asset`
   - `/thorchain/pool/{asset}/liquidity_provider/{addr}` → `pending_rune`, `pending_asset`, `pending_tx_id`
   - User's recent LP-related tx history to correlate.

## Acceptance criteria

- [ ] LP screen detects and renders pending half-deposits for the user's vault addresses.
- [ ] Pending card shows asset, amount deposited, paired address, and estimated refund time.
- [ ] CTA flows the user into completing the matching asset-side deposit with prefilled values.
- [ ] Covered by unit tests for the detection logic (LP record + pool pending fields).

## Notes

- Related to ongoing LP UI work on branch `aminsato/2141_thorchain_lp_ui`.
- Same logic likely applies to the iOS app — worth cross-filing if confirmed.

aminsato:
Closes #4331

## Summary

Two related gaps in the THORChain LP flows, both of which hide money the user has already committed.

**Pending half-deposits.** A symmetric add (`+:ASSET:<paired-address>`) mints no LP units until both sides arrive. In between, the LP record reads as an empty position, so every position lookup drops it — including ours, which treats a unit-less record as "no position". A user who sends the RUNE side and reopens the app sees nothing at all, and ~1 day later THORChain refunds the deposit with no indication anything happened.

The LP tab now shows a card for each half-deposit: what was deposited, the address the missing half has to come from, the time left before the refund, and a CTA into the deposit flow for the half still owed, on that half's own chain with the pool pre-selected.

**Remove LP second leg.** Withdrawing returns RUNE *and* the pool asset, but the form only rendered the RUNE leg, so the asset side stayed invisible until the confirmation screen. Both legs now render stacked with a divider, each with its own percentage, matching the design and iOS.

## Implementation notes

- Detection starts from `/thorchain/pools`, whose `pending_inbound_rune` / `pending_inbound_asset` say which pools hold anyone's pending liquidity. That collapses a ~100-pool fan-out into one request plus a per-user lookup for the handful of live candidates. Both sides of each candidate are tried, since a symmetric add is keyed by whichever address opened it.
- Refund deadline: `last_add_height + PendingLiquidityAgeLimit - currentHeight` at 6s/block. The limit comes from mimir (`PENDINGLIQUIDITYAGELIMIT`) and falls back to the 17,280 constant. A failed read renders "Unknown" rather than inventing a deadline.
- The pending cards intentionally bypass the Manage-Positions selection filter, including the `NoPositionsContainer` branch. A deposit stuck in a pool the user never selected is precisely the one that would otherwise expire unseen.
- Fixed a related shadowing bug: an empty rune-side LP record used to suppress the asset-side lookup entirely.
- The asset leg on Remove LP keeps thornode's full 1e8 precision — at the three decimals that suit RUNE and CACAO, a redeem of `0.07149251 VVV` rendered as `0.071`. Trailing zeros are trimmed on both legs so amounts read as `1.84`, not `1.840`.
- Maya's symmetric pools render both legs too: `loadRemoveLpData` reads `pool.assetDepth` alongside `cacaoDepth`, sharing `removeLpUnitsDivisor` with the CACAO leg.
- The THORChain LP preflight covers the RUNE side and the asset-side completion leg (which arrives on the pool's own asset chain). Maya's add-LP reuses the same strategy but is excluded — thornode's pause keys and pool status say nothing about a Maya pool.
- New strings are translated across all 10 locales.

## Test plan

- [x] `./gradlew :app:testDebugUnitTest :data:testDebugUnitTest` — green
- [x] `./gradlew :app:lintDebug` — clean
- [x] Both screens rendered on an emulator via `PreviewActivity` (`remove_thor_lp`, `thor_lp_pending`): the two Remove LP legs match the design and iOS, and the pending card renders above the LP list with no pools selected
- [ ] Not verified against a live pending pool — see below

## Open question for review

Issue #4331 reports `pending_rune: 0` on the LP record queried at the paired asset address, which would defeat LP-record-based attribution. Per THORChain's add-liquidity handler the pending amount *should* live on the LP record, so I'd expect that record to sit under the RUNE address instead — but I couldn't reach thornode to confirm. Worth a check against a currently-pending pool before merge; if thornode really reports zero on both sides, only the pool-level aggregate carries the signal and per-user attribution has no source.

## New Features
* View pending THORChain liquidity-pool deposits with amounts, paired addresses, token details, and refund countdowns.
* Complete eligible pending deposits directly from the liquidity positions screen.
* Improved liquidity removal screens to show both withdrawal assets and amounts.
* Refresh pending deposit information when returning to the liquidity positions screen.

## Bug Fixes
* Improved loading, retry, empty, and deposit validation states.

## Localization
* Added pending-deposit translations across supported languages.

## Testing
* Updated DeFi position coverage for pending deposits.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
@@ -17,6 +17,8 @@ import com.vultisig.wallet.data.models.Coins
 import com.vultisig.wallet.data.models.FiatValue
 import com.vultisig.wallet.data.models.ImageModel
 import com.vultisig.wallet.data.models.ThorChainLpPosition
+import com.vultisig.wallet.data.models.ThorChainPendingLpDeposit
+import com.vultisig.wallet.data.models.Vault
 import com.vultisig.wallet.data.models.VaultId
 import com.vultisig.wallet.data.models.coinType
 import com.vultisig.wallet.data.models.getCoinLogo
@@ -30,6 +32,7 @@ import com.vultisig.wallet.data.repositories.DefiPositionsRepository
 import com.vultisig.wallet.data.repositories.TokenPriceRepository
 import com.vultisig.wallet.data.repositories.VaultRepository
 import com.vultisig.wallet.data.usecases.GetThorChainLpPositionsUseCase
+import com.vultisig.wallet.data.usecases.GetThorChainPendingLpDepositsUseCase
 import com.vultisig.wallet.data.usecases.ThorchainBondUseCase
 import com.vultisig.wallet.data.utils.safeLaunch
 import com.vultisig.wallet.data.utils.toValue
@@ -55,6 +58,7 @@ import com.vultisig.wallet.ui.screens.v2.defi.thorchainSupportStakingDeFi
 import com.vultisig.wallet.ui.screens.v2.defi.toUiModel
 import com.vultisig.wallet.ui.utils.UiText
 import com.vultisig.wallet.ui.utils.formatTokenAmount
+import com.vultisig.wallet.ui.utils.lpRefundsInUiText
 import dagger.hilt.android.lifecycle.HiltViewModel
 import java.math.BigDecimal
 import java.math.BigInteger
@@ -133,6 +137,7 @@ constructor(
     private val defaultStakingPositionService: DefaultStakingPositionService,
     private val balanceVisibilityRepository: BalanceVisibilityRepository,
     private val getThorChainLpPositionsUseCase: GetThorChainLpPositionsUseCase,
+    private val getThorChainPendingLpDepositsUseCase: GetThorChainPendingLpDepositsUseCase,
     @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
 ) : ViewModel() {
 
@@ -173,6 +178,7 @@ constructor(
     private var currencyJob: Job? = null
     private var lpDialogJob: Job? = null
     private var loadLpJob: Job? = null
+    private var loadPendingLpJob: Job? = null
     private var loadBondedNodesJob: Job? = null
     private var loadStakingPositionsJob: Job? = null
 
@@ -489,7 +495,89 @@ constructor(
             loadStakingPositions()
 
             reloadLpTab()
+
+            loadPendingLpDeposits()
         }
+    }
+
+    /**
+     * Loads half-finished symmetric adds. Deliberately independent of [reloadLpTab]: those are
+     * gated on the pools the user picked in Manage Positions, and a deposit stuck in a pool they
+     * never selected is exactly the one that would otherwise be refunded unseen.
+     */
+    private fun loadPendingLpDeposits() {
+        loadPendingLpJob?.cancel()
+        loadPendingLpJob =
+            viewModelScope.launch {
+                try {
+                    val vault = withContext(ioDispatcher) { vaultRepository.get(vaultId) }
+                    val runeCoin =
+                        vault?.coins?.find { it.ticker == "RUNE" && it.chain == Chain.ThorChain }
+                            ?: return@launch
+
+                    val pending =
+                        withContext(ioDispatcher) {
+                            getThorChainPendingLpDepositsUseCase(
+                                runeAddress = runeCoin.address,
+                                resolveAssetAddress = { poolId ->
+                                    vault.assetAddressForPool(poolId)
+                                },
+                            )
+                        }
+
+                    val models = pending.map { it.toUiModel(vault.assetAddressForPool(it.pool)) }
+                    state.update { it.copy(lp = it.lp.copy(pendingDeposits = models)) }
+                } catch (e: Throwable) {
+                    if (e is CancellationException) throw e
+                    Timber.e(e, "Failed to load pending THORChain LP deposits")
+                }
+            }
+    }
+
+    private fun String.shortenForDisplay(): String =
+        if (length > 12) "${take(6)}…${takeLast(4)}" else this
+
+    /** The vault's address on a pool's non-RUNE chain, or `null` when it holds no account there. */
+    private fun Vault.assetAddressForPool(poolId: String): String? {
+        val assetChain = parseThorChainPool(poolId).chain ?: return null
+        if (assetChain == Chain.ThorChain) return null
+        val coin =
+            coins.firstOrNull { it.chain == assetChain && it.isNativeToken }
+                ?: coins.firstOrNull { it.chain == assetChain }
+        return coin?.address
+    }
+
+    private fun ThorChainPendingLpDeposit.toUiModel(
+        assetAddress: String?
+    ): PendingLpDepositUiModel {
+        val parsed = parseThorChainPool(pool)
+        val assetTicker = parsed.ticker
+        val runeTicker = Coins.ThorChain.RUNE.ticker
+        val depositedAmount =
+            if (isRunePending) {
+                CoinType.THORCHAIN.toValue(pendingRune)
+                    .stripTrailingZeros()
+                    .formatTokenAmount(runeTicker)
+            } else {
+                CoinType.THORCHAIN.toValue(pendingAsset)
+                    .stripTrailingZeros()
+                    .formatTokenAmount(assetTicker)
+            }
+
+        return PendingLpDepositUiModel(
+            poolId = pool,
+            icon =
+                lpAssetLogoRes(parsed.chain, assetTicker, parsed.contractAddress)
+                    ?: getCoinLogo(assetTicker.lowercase()),
+            chainLogo = parsed.chain?.monoToneLogo,
+            awaitedTicker = if (isRunePending) assetTicker else runeTicker,
+            depositedAmount = depositedAmount,
+            pairedAddress = pairedAddress?.shortenForDisplay(),
+            refundsIn = blocksUntilRefund?.let { lpRefundsInUiText(it * THORCHAIN_BLOCK_SECONDS) },
+            // Completing means sending the missing side, which needs an account on its chain. When
+            // RUNE is the missing half the vault always has one.
+            canComplete = !isRunePending || assetAddress != null,
+        )
     }
 
     @OptIn(ExperimentalCoroutinesApi::class)
@@ -1235,6 +1323,29 @@ constructor(
         }
     }
 
+    /**
+     * Sends the user into the deposit flow for the half they still owe, on that half's own chain,
+     * with the pool pre-selected. The memo the flow builds is what THORChain matches against the
+     * pending record, so this completes the add rather than opening a second one.
+     */
+    fun onClickCompletePendingLp(poolId: String) {
+        val pending = state.value.lp.pendingDeposits.find { it.poolId == poolId } ?: return
+        val missingSideChain =
+            if (pending.awaitedTicker == Coins.ThorChain.RUNE.ticker) Chain.ThorChain
+            else parseThorChainPool(poolId).chain ?: return
+
+        viewModelScope.safeLaunch {
+            navigator.route(
+                Route.Deposit(
+                    vaultId = vaultId,
+                    chainId = missingSideChain.id,
+                    depositType = DeFiNavActions.ADD_LP.type,
+                    poolId = poolId,
+                )
+            )
+        }
+    }
+
     fun onClickRemoveLp(poolId: String) {
         viewModelScope.safeLaunch {
             navigator.route(
@@ -1451,6 +1562,7 @@ constructor(
         private const val RUJI_SYMBOL = "RUJI"
         private const val RUJI_REWARDS_SYMBOL = "USDC"
         private const val POSITION_DISPLAY_SCALE = 4
+        private const val THORCHAIN_BLOCK_SECONDS = 6L
 
         /**
          * The Manage-Positions key a placeholder is gated on. Both RUJI positions are toggled by
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
@@ -550,11 +550,27 @@ constructor(
     }
 
     /**
-     * Marks the scan as settled without touching [LpTabUiModel.pendingDeposits]. A failed reload
-     * must not erase a list an earlier one found — the refund timer is still running on it.
+     * Marks the scan as settled without erasing [LpTabUiModel.pendingDeposits]: a failed reload
+     * must still show a list an earlier one found, because the refund timer is running on it and
+     * the user needs to know. Completion is withdrawn instead — the whole point of re-scanning is
+     * that THORChain may already have refunded the deposit, and nothing downstream re-checks:
+     * onClickCompletePendingLp reads this list directly, and the add-liquidity preflight only asks
+     * about pool-wide pause and status, never about this record. A card the scan could not confirm
+     * therefore stays visible but cannot spend inbound gas on a dead deposit.
      */
     private fun markPendingLpDepositsSettled() {
-        state.update { it.copy(lp = it.lp.copy(pendingDepositsLoaded = true)) }
+        state.update {
+            it.copy(
+                lp =
+                    it.lp.copy(
+                        pendingDeposits =
+                            it.lp.pendingDeposits.map { deposit ->
+                                deposit.copy(canComplete = false)
+                            },
+                        pendingDepositsLoaded = true,
+                    )
+            )
+        }
     }
 
     /**
@@ -1362,6 +1378,9 @@ constructor(
      */
     fun onClickCompletePendingLp(poolId: String) {
         val pending = state.value.lp.pendingDeposits.find { it.poolId == poolId } ?: return
+        // Also enforced by the card's disabled button; held here too so a card the last scan
+        // could not confirm can never route into a deposit, whatever the UI does.
+        if (!pending.canComplete) return
         val missingSideChain =
             if (pending.awaitedTicker == Coins.ThorChain.RUNE.ticker) Chain.ThorChain
             else parseThorChainPool(poolId).chain ?: return
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
@@ -508,29 +508,24 @@ constructor(
     private fun loadPendingLpDeposits() {
         loadPendingLpJob?.cancel()
         loadPendingLpJob =
-            viewModelScope.launch {
-                try {
-                    val vault = withContext(ioDispatcher) { vaultRepository.get(vaultId) }
-                    val runeCoin =
-                        vault?.coins?.find { it.ticker == "RUNE" && it.chain == Chain.ThorChain }
-                            ?: return@launch
+            viewModelScope.safeLaunch(
+                onError = { Timber.e(it, "Failed to load pending THORChain LP deposits") }
+            ) {
+                val vault = withContext(ioDispatcher) { vaultRepository.get(vaultId) }
+                val runeCoin =
+                    vault?.coins?.find { it.ticker == "RUNE" && it.chain == Chain.ThorChain }
+                        ?: return@safeLaunch
 
-                    val pending =
-                        withContext(ioDispatcher) {
-                            getThorChainPendingLpDepositsUseCase(
-                                runeAddress = runeCoin.address,
-                                resolveAssetAddress = { poolId ->
-                                    vault.assetAddressForPool(poolId)
-                                },
-                            )
-                        }
+                val pending =
+                    withContext(ioDispatcher) {
+                        getThorChainPendingLpDepositsUseCase(
+                            runeAddress = runeCoin.address,
+                            resolveAssetAddress = { poolId -> vault.assetAddressForPool(poolId) },
+                        )
+                    }
 
-                    val models = pending.map { it.toUiModel(vault.assetAddressForPool(it.pool)) }
-                    state.update { it.copy(lp = it.lp.copy(pendingDeposits = models)) }
-                } catch (e: Throwable) {
-                    if (e is CancellationException) throw e
-                    Timber.e(e, "Failed to load pending THORChain LP deposits")
-                }
+                val models = pending.map { it.toUiModel(vault.assetAddressForPool(it.pool)) }
+                state.update { it.copy(lp = it.lp.copy(pendingDeposits = models)) }
             }
     }
 
@@ -1081,7 +1076,7 @@ constructor(
 
         if (selectedPools.isEmpty()) {
             loadLpJob?.cancel()
-            state.update { it.copy(lp = LpTabUiModel(isLoading = false, positions = emptyList())) }
+            state.update { it.copy(lp = it.lp.copy(isLoading = false, positions = emptyList())) }
             loadLpJob = viewModelScope.launch { reportLpFiat(BigDecimal.ZERO) }
             return
         }
@@ -1099,7 +1094,7 @@ constructor(
                 // has no liquidity yet should be visible so the Add button is reachable.
                 val placeholders = selectedPools.map { it.toPlaceholderUiModel(zero) }
                 state.update {
-                    it.copy(lp = LpTabUiModel(isLoading = true, positions = placeholders))
+                    it.copy(lp = it.lp.copy(isLoading = true, positions = placeholders))
                 }
 
                 try {
@@ -1114,7 +1109,7 @@ constructor(
                         Timber.e("Vault does not have RUNE coin for LP positions")
                         reportLpFiat(BigDecimal.ZERO)
                         state.update {
-                            it.copy(lp = LpTabUiModel(isLoading = false, positions = placeholders))
+                            it.copy(lp = it.lp.copy(isLoading = false, positions = placeholders))
                         }
                         return@launch
                     }
@@ -1186,15 +1181,13 @@ constructor(
                             // understate the header total rather than admit a value is missing.
                             LpLegTotal.Unavailable
                         }
-                    state.update {
-                        it.copy(lp = LpTabUiModel(isLoading = false, positions = merged))
-                    }
+                    state.update { it.copy(lp = it.lp.copy(isLoading = false, positions = merged)) }
                 } catch (e: Throwable) {
                     if (e is CancellationException) throw e
                     Timber.e(e, "Failed to load THORChain LP positions")
                     reportLpFiat(BigDecimal.ZERO)
                     state.update {
-                        it.copy(lp = LpTabUiModel(isLoading = false, positions = placeholders))
+                        it.copy(lp = it.lp.copy(isLoading = false, positions = placeholders))
                     }
                 }
             }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
@@ -104,9 +104,11 @@ internal data class ThorchainDefiPositionsUiModel(
     val bondPositionsDialog: List<PositionUiModelDialog> = defaultPositionsBondDialog(),
     val stakingPositionsDialog: List<PositionUiModelDialog> = defaultPositionsStakingDialog(),
     val lpPositionsDialog: List<PositionUiModelDialog> = emptyList(),
-    // Flips true once the available-pools fetch returns. Until then the LP tab should stay in
-    // loading state instead of flashing the empty/no-positions UI when the user has selected
-    // positions whose keys don't yet match the (empty) dialog list.
+    // Flips true once the available-pools fetch settles, success or failure. Until then the LP tab
+    // should stay in loading state instead of flashing the empty/no-positions UI when the user has
+    // selected positions whose keys don't yet match the (empty) dialog list. A failed fetch settles
+    // too: an unclearable spinner is worse than the no-positions container, which at least carries
+    // the Manage Positions retry.
     val lpDialogLoaded: Boolean = false,
     val selectedPositions: List<String> = defaultSelectedPositionsDialog(),
     val tempSelectedPositions: List<String> = defaultSelectedPositionsDialog(),
@@ -219,10 +221,14 @@ constructor(
                 Timber.e(e, "Failed to load THORChain LP pools for dialog")
                 // Leave availablePools null so the next user interaction (e.g. opening Manage
                 // Positions or saving a selection) retries instead of soft-locking.
-                // The tab deliberately stays in its loading state (see above), but the header total
-                // is a separate concern: reloadLpTab parks the LP leg unreported while it waits for
-                // this dataset, so report it as zero here or the total would never see all five
-                // legs and would spin forever.
+                // lpDialogLoaded means "settled", not "succeeded": leaving it false parks the tab
+                // in a spinner that nothing can ever clear, while any pending half-deposit still
+                // renders through it. Settling drops the tab to its no-positions container, whose
+                // Manage Positions button is the retry, and lets a pending card stand on its own.
+                state.update { it.copy(lpDialogLoaded = true) }
+                // The header total is a separate concern: reloadLpTab parks the LP leg unreported
+                // while it waits for this dataset, so report it as zero here or the total would
+                // never see all five legs and would spin forever.
                 reportLpFiat(BigDecimal.ZERO)
             }
         }
@@ -509,24 +515,48 @@ constructor(
         loadPendingLpJob?.cancel()
         loadPendingLpJob =
             viewModelScope.safeLaunch(
-                onError = { Timber.e(it, "Failed to load pending THORChain LP deposits") }
+                onError = {
+                    Timber.e(it, "Failed to load pending THORChain LP deposits")
+                    markPendingLpDepositsSettled()
+                }
             ) {
                 val vault = withContext(ioDispatcher) { vaultRepository.get(vaultId) }
                 val runeCoin =
                     vault?.coins?.find { it.ticker == "RUNE" && it.chain == Chain.ThorChain }
-                        ?: return@safeLaunch
+                if (runeCoin == null) {
+                    markPendingLpDepositsSettled()
+                    return@safeLaunch
+                }
 
                 val pending =
                     withContext(ioDispatcher) {
-                        getThorChainPendingLpDepositsUseCase(
-                            runeAddress = runeCoin.address,
-                            resolveAssetAddress = { poolId -> vault.assetAddressForPool(poolId) },
-                        )
+                        getThorChainPendingLpDepositsUseCase(runeAddress = runeCoin.address)
                     }
 
                 val models = pending.map { it.toUiModel(vault.assetAddressForPool(it.pool)) }
-                state.update { it.copy(lp = it.lp.copy(pendingDeposits = models)) }
+                state.update {
+                    it.copy(lp = it.lp.copy(pendingDeposits = models, pendingDepositsLoaded = true))
+                }
             }
+    }
+
+    /**
+     * Marks the scan as settled without touching [LpTabUiModel.pendingDeposits]. A failed reload
+     * must not erase a list an earlier one found — the refund timer is still running on it.
+     */
+    private fun markPendingLpDepositsSettled() {
+        state.update { it.copy(lp = it.lp.copy(pendingDepositsLoaded = true)) }
+    }
+
+    /**
+     * Re-scans for pending half-deposits when the screen comes back to the foreground. They sit on
+     * a refund timer that keeps running while the app is backgrounded, so a stale list can offer
+     * Complete Deposit on a deposit THORChain has already refunded. Positions are left to
+     * pull-to-refresh; only this leg goes stale on its own.
+     */
+    fun onScreenResumed() {
+        if (!::vaultId.isInitialized) return
+        loadPendingLpDeposits()
     }
 
     private fun String.shortenForDisplay(): String =
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
@@ -614,13 +614,20 @@ constructor(
                     .formatTokenAmount(assetTicker)
             }
 
+        // The card is about the side that has not arrived — its title names that ticker — so the
+        // icon and chain badge have to follow it. Pinning them to the pool's asset put an ETH logo
+        // next to "Waiting for matching RUNE deposit".
+        val awaitedChain = if (isRunePending) parsed.chain else Chain.ThorChain
+        val awaitedTicker = if (isRunePending) assetTicker else runeTicker
+        val awaitedContractAddress = if (isRunePending) parsed.contractAddress else ""
+
         return PendingLpDepositUiModel(
             poolId = pool,
             icon =
-                lpAssetLogoRes(parsed.chain, assetTicker, parsed.contractAddress)
-                    ?: getCoinLogo(assetTicker.lowercase()),
-            chainLogo = parsed.chain?.monoToneLogo,
-            awaitedTicker = if (isRunePending) assetTicker else runeTicker,
+                lpAssetLogoRes(awaitedChain, awaitedTicker, awaitedContractAddress)
+                    ?: getCoinLogo(awaitedTicker.lowercase()),
+            chainLogo = awaitedChain?.monoToneLogo,
+            awaitedTicker = awaitedTicker,
             depositedAmount = depositedAmount,
             pairedAddress = pairedAddress?.shortenForDisplay(),
             refundsIn = blocksUntilRefund?.let { lpRefundsInUiText(it * THORCHAIN_BLOCK_SECONDS) },
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
@@ -513,6 +513,15 @@ constructor(
      */
     private fun loadPendingLpDeposits() {
         loadPendingLpJob?.cancel()
+        // Re-arm the gate only when there is a list that could now be wrong. A deposit found before
+        // the app was backgrounded may already have been refunded, and Complete Deposit on a
+        // refunded one just burns inbound gas — so hide it behind the spinner until this scan
+        // confirms it. With nothing pending there is nothing to invalidate, and re-arming would
+        // flash the whole tab to a spinner on every resume for the users who have no half-deposit
+        // at all.
+        if (state.value.lp.pendingDeposits.isNotEmpty()) {
+            state.update { it.copy(lp = it.lp.copy(pendingDepositsLoaded = false)) }
+        }
         loadPendingLpJob =
             viewModelScope.safeLaunch(
                 onError = {
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/RemoveLpCalculator.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/RemoveLpCalculator.kt
@@ -10,9 +10,9 @@ internal object RemoveLpCalculator {
     const val RUNE_DECIMALS = 8
     const val DISPLAY_SCALE = 3
     // Precision kept on `userUnits * poolDepth / totalUnits` before we shift by
-    // 10^decimals and round to DISPLAY_SCALE. Must be >= max decimals + DISPLAY_SCALE
-    // (= 13) to avoid truncation; 18 matches the standard fixed-point scale and leaves
-    // headroom.
+    // 10^decimals and round to DISPLAY_SCALE. Must be >= decimals + scale of any call or the
+    // first DOWN division discards digits the second one still needs; 18 matches the standard
+    // fixed-point scale and leaves headroom over the largest current call (8 + 8).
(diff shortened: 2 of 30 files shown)


aminsato:
Thanks — pushed 32e659a5a.

**Fixed**

- **Pending deposits erased on LP tab reload.** Confirmed and real, and worse than it looks: `reloadLpTab` rebuilt `LpTabUiModel` from scratch at five points, and it runs concurrently with the pending load off `loadSavedPositions`, with `loadLpPositionsForDialog` triggering another reload once the pool list arrives. The card could vanish moments after appearing — including in the exact case it exists for, where no pools are selected and it is the only thing on the tab. Each site now copies the existing tab state.
- **`safeLaunch` for the pending load.** Switched, and dropped the hand-rolled try/catch since `safeLaunch` already rethrows cancellation and logs the rest.
- **Duplicated icon/badge block.** Extracted `AssetIconWithChainBadge`, now shared by both LP cards. Re-rendered the card on an emulator after the refactor to confirm it is pixel-identical.

**Not doing**

- **Tests for the detection use case and the calculator paths.** Deliberate for this PR — noted in the description. It leaves an acceptance criterion on #4331 open.
- **`@Immutable` on `PendingLpDepositUiModel`.** Nothing else in `DefiTabUiModels.kt` carries it, including the `LpPositionUiModel` and `LpTabUiModel` this sits beside. Annotating one model in a file where none of its siblings are annotated would be inconsistent; if the file should adopt `@Immutable`, that is worth doing to all of them at once rather than to the newest addition only.

Green locally: unit tests, lint, and both screens re-verified on an emulator.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1098,15 +1181,13 @@ constructor(
                             // understate the header total rather than admit a value is missing.
                             LpLegTotal.Unavailable
                         }
-                    state.update {
-                        it.copy(lp = LpTabUiModel(isLoading = false, positions = merged))
-                    }
+                    state.update { it.copy(lp = it.lp.copy(isLoading = false, positions = merged)) }
                 } catch (e: Throwable) {
                     if (e is CancellationException) throw e
                     Timber.e(e, "Failed to load THORChain LP positions")
```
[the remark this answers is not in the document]
aminsato: Not changing this one.

That `catch (e: Throwable)` is pre-existing (`6546b72aa1`) — this PR only touched the `state.update` line inside it, which is why the finding anchors here. Narrowing exception semantics on an established load path is unrelated to surfacing pending deposits, and doing it in this PR would put a behaviour change where reviewers are not looking for one.

It is also the convention rather than an oversight: all four `catch (e: Throwable)` blocks in this ViewModel follow the same `CancellationException`-rethrow-then-log shape, and `safeLaunch` — the wrapper this project mandates for network coroutines, and which you correctly asked me to adopt for the pending load in the same review — catches `Throwable` itself:

```kotlin
    throw e
} catch (e: Throwable) {
    onError(e)
}
```

So narrowing this single site would make it inconsistent with both the file and the helper the rest of the codebase routes through.

On the substance: if an `OutOfMemoryError` or `StackOverflowError` fires here, letting it propagate out of a `viewModelScope` coroutine takes down the process rather than restoring any real invariant — the LP tab is not where that gets recovered. Worth doing as a deliberate codebase-wide pass on the idiom (including `safeLaunch`) if the team wants it; not worth a one-off divergence here.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -489,9 +495,86 @@ constructor(
...
+                val pending =
+                    withContext(ioDispatcher) {
+                        getThorChainPendingLpDepositsUseCase(
+                            runeAddress = runeCoin.address,
+                            resolveAssetAddress = { poolId -> vault.assetAddressForPool(poolId) },
+                        )
+                    }
+
+                val models = pending.map { it.toUiModel(vault.assetAddressForPool(it.pool)) }
+                state.update { it.copy(lp = it.lp.copy(pendingDeposits = models)) }
```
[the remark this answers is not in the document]
aminsato: Same call as the other two coverage findings on this PR, answered in https://github.com/vultisig/vultisig-android/pull/5637#issuecomment-5327085296: no tests are being added here, by decision rather than oversight.

You are right that #4331 lists detection coverage as an acceptance criterion, and it is not met — that is stated plainly in the PR description and stays open for the maintainers to call. The paths you name are the right ones if it gets picked up: RUNE-side hit, asset-side fallback when the RUNE-side record is an empty husk, no RUNE account, and the refund-countdown zero-floor.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/RemoveLpCalculator.kt
```diff
@@ -40,19 +40,28 @@ internal object RemoveLpCalculator {
...
             .multiply(poolDepth.toBigDecimal())
             .divide(totalPoolUnits.toBigDecimal(), LP_UNITS_INTERMEDIATE_SCALE, RoundingMode.DOWN)
-            .divide(BigDecimal.TEN.pow(decimals), DISPLAY_SCALE, RoundingMode.DOWN)
+            .divide(BigDecimal.TEN.pow(decimals), scale, RoundingMode.DOWN)
             .toPlainString()
     }
+
+    /** Drops the padding zeros a fixed [scale] leaves behind: `"1.840"` reads as `"1.84"`. */
+    fun trimTrailingZeros(display: String): String =
+        display.toBigDecimalOrNull()?.stripTrailingZeros()?.toPlainString() ?: display
```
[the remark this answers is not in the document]
aminsato: Not adding tests in this PR — same decision as the other two coverage findings, explained in https://github.com/vultisig/vultisig-android/pull/5637#issuecomment-5327085296.

The two cases you name are the right ones, and both were verified by hand rather than by assertion. The 8-decimal path is exactly why the `scale` parameter exists: a real withdrawal of `0.07149251 VVV` renders as `0.071` at the default three decimals, which reads as "nothing to withdraw". I rendered that value on an emulator to confirm it survives, and `trimTrailingZeros` is what turns the RUNE leg from `1.840` into `1.84` next to it — the mismatch was visible enough on screen that it is what prompted trimming both legs rather than just the asset one.

`RemoveLpCalculatorTest` is the natural home if this gets picked up: a sub-0.001 amount at `scale = RUNE_DECIMALS`, plus padded and zero inputs to `trimTrailingZeros`.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1235,6 +1316,29 @@ constructor(
...
+    /**
+     * Sends the user into the deposit flow for the half they still owe, on that half's own chain,
+     * with the pool pre-selected. The memo the flow builds is what THORChain matches against the
+     * pending record, so this completes the add rather than opening a second one.
+     */
+    fun onClickCompletePendingLp(poolId: String) {
+        val pending = state.value.lp.pendingDeposits.find { it.poolId == poolId } ?: return
+        val missingSideChain =
+            if (pending.awaitedTicker == Coins.ThorChain.RUNE.ticker) Chain.ThorChain
+            else parseThorChainPool(poolId).chain ?: return
```
rkokhatskyi, line 1409: `onClickCompletePendingLp` opens the asset-chain completion via the normal add-liquidity flow, but `resolvePairedAddress` (`DepositFormViewModel.kt:571`) returns null for non-ThorChain chains. The memo then carries no paired address, so thornode mints an unintended new position instead of completing the pending deposit.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainPendingLpDepositsUseCase.kt
```diff
@@ -0,0 +1,195 @@
...
+        (pendingRune.toBigIntegerOrNull()?.signum() ?: 0) > 0 ||
+            (pendingAsset.toBigIntegerOrNull()?.signum() ?: 0) > 0
+
+    /**
+     * Reads an LP record as a half-finished symmetric add, or `null` when nothing is pending on it.
+     * Both sides are checked: the user may have sent either half first. A record that already holds
+     * units is a live position, not a pending deposit, even if a later add is still settling.
+     */
+    private fun ThorChainLiquidityProviderJson.toPendingFetch(pool: String): PendingFetch? {
+        if ((units.toBigIntegerOrNull()?.signum() ?: 0) > 0) return null
```
rkokhatskyi: `toPendingFetch` discards any LP record whose units are already nonzero, but thornode keeps staging PendingRune/PendingAsset on existing records the same way. A returning LP's top-up deposit becomes invisible here and silently refunds about a day later with no warning shown anywhere in the app.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/load/LiquidityDataLoader.kt
```diff
@@ -166,6 +167,11 @@ constructor(
                 removeLpPoolDepth = BigInteger.ZERO,
                 removeLpPercent = 0f,
                 removeLpCacaoDisplay = "",
+                // Maya pools withdraw CACAO only; clear any asset leg a previous THORChain
+                // selection left behind so the second amount does not linger on this form.
+                removeLpAssetDisplay = "",
+                removeLpAssetSymbol = "",
```
rkokhatskyi, line 178: Maya's `loadRemoveLpData` reads `pool.cacaoDepth` but never `pool.assetDepth`, so `removeLpAssetSymbol` stays empty and `RemoveLpScreen.kt:138`'s dual-leg row never renders for Maya. Users removing a CACAO pool position never see the paired asset amount they are about to receive on withdrawal.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -489,9 +495,86 @@ constructor(
...
+            loadPendingLpDeposits()
         }
     }
 
+    /**
+     * Loads half-finished symmetric adds. Deliberately independent of [reloadLpTab]: those are
+     * gated on the pools the user picked in Manage Positions, and a deposit stuck in a pool they
+     * never selected is exactly the one that would otherwise be refunded unseen.
+     */
+    private fun loadPendingLpDeposits() {
```
rkokhatskyi, line 514: `loadPendingLpDeposits` only runs once per screen entry with no resume refresh, so a long-backgrounded app can keep showing an already-refunded pending deposit with Complete Deposit still enabled. Tapping it just opens a fresh add-liquidity deposit, wasting gas on a redundant pending cycle.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/thorchain/ThorchainDefiPositionsScreen.kt
```diff
@@ -218,14 +220,20 @@ internal fun ThorchainDefiPositionScreenContent(
...
                                     onClickAdd = onClickAddLp,
                                     onClickRemove = onClickRemoveLp,
+                                    onClickCompletePending = onClickCompletePendingLp,
                                 )
-                            !state.selectedPositions.hasLpPositions(state.lpPositionsDialog) ->
+                            // A pending half-deposit is on a refund timer and belongs to no
+                            // selected pool, so it must survive the empty-selection branch —
+                            // otherwise the one user who has to act sees "no positions".
+                            !state.selectedPositions.hasLpPositions(state.lpPositionsDialog) &&
+                                state.lp.pendingDeposits.isEmpty() ->
```
rkokhatskyi, line 240: `state.lp.pendingDeposits.isEmpty()` gates this branch with no loaded-flag of its own, unlike `lpDialogLoaded` for positions. `loadPendingLpDeposits` (`ThorchainDefiPositionsViewModel.kt:508`) does more round trips than the dialog fetch, so a vault with no selected pools and a real pending deposit can flash `NoPositionsContainer` first.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/maya/RemoveLpScreen.kt
```diff
@@ -195,6 +195,32 @@ internal fun RemoveLpScreenContent(
...
+                if (amount.isEmpty()) stringResource(R.string.remove_pool_zero_amount, symbol)
+                else stringResource(R.string.remove_pool_amount_format, amount, symbol),
+            style = Theme.brockmann.headings.largeTitle,
+            color = Theme.v2.colors.text.primary,
+            textAlign = TextAlign.Center,
+        )
+
+        Text(
+            text = percentLabel,
+            style = Theme.brockmann.headings.subtitle,
```
rkokhatskyi: This extraction swaps the percent caption from `Theme.brockmann.body.m.medium` (16sp/24sp) to `Theme.brockmann.headings.subtitle` (15sp/17sp, -0.18sp tracking), a font regression the commit never mentions. `UnstakeCacaoScreen.kt` still pairs the identical amount/caption pattern with `body.m.medium`, so this caption now reads smaller and tighter than every sibling screen.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainPendingLpDepositsUseCase.kt
```diff
@@ -0,0 +1,195 @@
...
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
+import timber.log.Timber
+
+/** Mimir key for the number of blocks THORChain holds a half-deposit before refunding it. */
+private const val PENDING_LIQUIDITY_AGE_LIMIT_MIMIR = "PENDINGLIQUIDITYAGELIMIT"
+
+/** `PendingLiquidityAgeLimit`'s constant value, used when mimir carries no override (~1 day). */
+private const val DEFAULT_PENDING_LIQUIDITY_AGE_LIMIT = 17_280L
```
rkokhatskyi: `DEFAULT_PENDING_LIQUIDITY_AGE_LIMIT` is 17,280 blocks, but thornode's live mimir and its `constants_v1.go` base value are both 100,800, off by 5.8x. When `getMimir()` succeeds but omits the key, the countdown understates the real ~7 day window by the same factor with no estimate qualifier shown.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/LpTabComponents.kt
```diff
@@ -25,21 +25,36 @@ import androidx.compose.ui.unit.dp
...
 internal fun LpTabContent(
     state: LpTabUiModel,
     onClickAdd: (String) -> Unit,
     onClickRemove: (String) -> Unit,
+    onClickCompletePending: (String) -> Unit = {},
 ) {
     Column(modifier = Modifier.fillMaxWidth(), verticalArrangement = Arrangement.spacedBy(16.dp)) {
+        // Above the positions: a half-deposit is on a refund timer, so it is the one thing on this
+        // tab the user may need to act on before it expires.
+        state.pendingDeposits.forEach { pending ->
```
rkokhatskyi, line 51: This renders `pendingDeposits` unconditionally, so when `loadLpPositionsForDialog` (`ThorchainDefiPositionsViewModel.kt:217`) fails, `lpDialogLoaded` stays false forever and the header total pins at zero, yet a real pending card still shows through the tab's permanent not-yet-loaded state with no retry.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainPendingLpDepositsUseCase.kt
```diff
@@ -0,0 +1,195 @@
...
+    private suspend fun fetchPending(
+        pool: String,
+        runeAddress: String,
+        assetAddress: String?,
+    ): PendingFetch? =
+        try {
+            val runeSide = thorChainApi.getLiquidityProvider(pool, runeAddress)
+            val record =
+                runeSide?.takeIf { it.hasPending() }
+                    ?: assetAddress
```
rkokhatskyi: This `assetAddress` fallback can't find a pending symmetric add the `runeAddress` lookup missed: thornode's `fetchAddr := runeAddr; if empty { fetchAddr = assetAddr }` always prefers RUNE when present, so a record is never keyed by the asset side alone. Drop the fallback, its doc comment is stale too.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/RemoveLpCalculator.kt
```diff
@@ -40,19 +40,28 @@ internal object RemoveLpCalculator {
...
         poolDepth: BigInteger,
         totalPoolUnits: BigInteger,
         decimals: Int,
+        scale: Int = DISPLAY_SCALE,
     ): String? {
         if (totalPoolUnits.signum() <= 0) return null
         return selectedUnits
             .toBigDecimal()
             .multiply(poolDepth.toBigDecimal())
             .divide(totalPoolUnits.toBigDecimal(), LP_UNITS_INTERMEDIATE_SCALE, RoundingMode.DOWN)
```
rkokhatskyi, line 66: `computeAmountDisplay` has no guard for `decimals + scale > 18`, so the two-step DOWN rounding can silently return the wrong value, not just less precision: `decimals=0, scale=19` yields `0.3333333333333333330` where the exact value is `...3333`. Every current call site stays under 18, but nothing enforces it.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
rkokhatskyi: Same root cause also crashes outright for pool-asset chains outside `SECURE_ASSETS_TICKERS` (BTC/ETH/BCH/LTC/DOGE/AVAX/BNB) — including the PR's own BASE.VVV example. `DepositDataLoader.kt`'s `wireInitialState` builds an empty `depositOptions` list for those chains and `depositOptions.first()` throws `NoSuchElementException` before the user can enter an amount.

aminsato: Both confirmed and fixed in a259ca0b0.

The crash: `DepositDataLoader.wireInitialState` no longer calls `depositOptions.first()`. The initial option is seeded from the pending deep-link action, falling back to `firstOrNull()`. Deep-linked options were never in the per-chain dropdown anyway — `setMetadataInfo` overrode the seed a moment later — so the list having no head is now simply not a question the seed asks. `Chain.Base -> "BASE"` and `Chain.Ripple -> "XRP"` were both outside `SECURE_ASSETS_TICKERS`, so BASE and XRP pools were reachable crashes.

The paired address: `resolvePairedAddress` now resolves the *other* half's chain whichever side is being deposited — ThorChain -> asset chain, asset chain -> ThorChain — instead of returning null for everything that isn't ThorChain. `AddLiquidityStrategy` now requires it on both halves of a symmetric pool rather than only the RUNE side.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainPendingLpDepositsUseCase.kt
aminsato: Fixed in a259ca0b0. Dropped the `units > 0` early return in `toPendingFetch`; a nonzero `units` no longer rules a record out, since thornode stages a top-up's pending halves onto the same record that already holds the live position. Doc comment updated to say why.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/load/LiquidityDataLoader.kt
aminsato: Fixed in a259ca0b0. `loadMayaRemoveLpData` now reads `pool.assetDepth` alongside `cacaoDepth` and publishes `removeLpAssetRedeemBase`/`removeLpAssetSymbol`, so the dual-leg row renders for Maya too. It shares `removeLpUnitsDivisor` with the CACAO leg (`selectedUnits * assetDepth / totalPoolUnits`). The "Maya pools withdraw CACAO only" comment that justified the empty symbol is gone — it was the wrong premise.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed in a259ca0b0. Added `onScreenResumed()`, wired from `LifecycleEventEffect(Lifecycle.Event.ON_RESUME)`, mirroring `CosmosStakingPositionsScreen`. It re-runs only the pending scan — positions stay on pull-to-refresh, since this is the leg that goes stale on its own while backgrounded.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/thorchain/ThorchainDefiPositionsScreen.kt
aminsato: Fixed in a259ca0b0. Added `LpTabUiModel.pendingDepositsLoaded`, set once the scan settles either way, and the tab now waits on `!state.lpDialogLoaded || !state.lp.pendingDepositsLoaded` before it will render any definitive branch. Before that flag, "nothing pending" was indistinguishable from "not asked yet".

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/maya/RemoveLpScreen.kt
aminsato: Fixed in a259ca0b0 — reverted to `Theme.brockmann.body.m.medium`. Unintentional; it rode along with extracting `RemoveLpLeg` and should have stayed matched to `UnstakeCacaoScreen`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainPendingLpDepositsUseCase.kt
aminsato: Fixed in a259ca0b0. Confirmed against live mimir on the app's own endpoint (`gateway.liquify.com/chain/thorchain_api/thorchain/mimir`): `"PENDINGLIQUIDITYAGELIMIT": 100800`. Constant corrected to `100_800L` and the "~1 day" doc comment corrected to ~7 days at ~6s blocks.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/LpTabComponents.kt
aminsato: Fixed in a259ca0b0, in two parts.

`lpDialogLoaded` now means *settled*, not *succeeded* — the catch in `loadLpPositionsForDialog` sets it too. A failed fetch previously parked the tab in a spinner nothing could clear; now it falls through to `NoPositionsContainer`, whose Manage Positions button is the existing retry (`ensureAvailablePoolsLoaded`). `availablePools` still stays null so that retry re-fetches.

And the loading branch now passes `pendingDeposits = emptyList()`, so a pending card can no longer render through a tab that is still declaring itself unloaded.

The test that asserted the old contract (`the LP tab stays loading until the available-pool fetch resolves`) is updated to the new one.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainPendingLpDepositsUseCase.kt
aminsato: Fixed in a259ca0b0. Dropped the fallback and the stale doc comment. Since it was the only consumer, `resolveAssetAddress` is gone from the use case's signature too, and `hasPending()` with it — `fetchPending` is now a single lookup by RUNE address.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/RemoveLpCalculator.kt
aminsato: Fixed in a259ca0b0. Added `require(decimals + scale <= LP_UNITS_INTERMEDIATE_SCALE)` with the reason inline, and corrected the constant's comment — it claimed the bound was `max decimals + DISPLAY_SCALE`, but `scale` is a parameter, so the real bound is per-call. Largest current call is the asset leg at 8 + 8.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/load/LiquidityDataLoader.kt
aminsato: Correction on the symbol name above: the function is `loadRemoveLpData` (LiquidityDataLoader.kt:148), as you had it — not `loadMayaRemoveLpData`. The change itself is as described.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/AddLiquidityStrategy.kt
```diff
@@ -79,15 +79,16 @@ internal class AddLiquidityStrategy(
...
+        // Either half of a symmetric add MUST name the other half's address, or THORChain opens a
+        // separate asymmetric position rather than crediting the pair: a RUNE-side add carries the
+        // asset address, and the asset-side add that completes a pending half-deposit carries the
+        // RUNE address.
+        val assetChain = parseThorChainPool(poolId).chain
+        val isSymmetricPool = assetChain != null && assetChain != Chain.ThorChain
+        if (isSymmetricPool && pairedAddress == null) {
+            throw InvalidTransactionDataException(
+                UiText.StringResource(R.string.send_error_no_address)
+            )
```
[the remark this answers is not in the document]
aminsato: Good catch — fixed in 5806f6d46. The preflight guard is now `chain == Chain.ThorChain || isSymmetricPool`, with `assetChain`/`isSymmetricPool` hoisted above it. It validates the pool's status and pause keys, which say nothing about which side the inbound comes from, so the asset-side completion is refundable on identical terms and was skipping exactly the check that exists to spare that gas. Asymmetric pools keep their current behaviour.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/load/LiquidityDataLoader.kt
```diff
@@ -166,6 +167,11 @@ constructor(
                 removeLpPoolDepth = BigInteger.ZERO,
                 removeLpPercent = 0f,
                 removeLpCacaoDisplay = "",
+                // Clear the asset leg a previous selection left behind; the Maya pool's own asset
+                // depth fills it back in below, once loaded.
+                removeLpAssetDisplay = "",
+                removeLpAssetSymbol = "",
+                removeLpAssetRedeemBase = BigInteger.ZERO,
```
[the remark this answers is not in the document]
aminsato: Fixed in 5806f6d46. The null-pool branch now clears `removeLpAssetDisplay`, `removeLpAssetSymbol` and `removeLpAssetRedeemBase`, matching `loadThorChainRemoveLpData`'s equivalent branch. It returns before the reset below it, so it has to do its own clearing.

app/src/test/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModelTest.kt
```diff
@@ -453,16 +457,19 @@ internal class ThorchainDefiPositionsViewModelTest {
...
 
-        // availablePools stays null so the tab must not flash an empty state.
-        assertTrue(vm.state.value.lp.isLoading)
-        assertFalse(vm.state.value.lpDialogLoaded)
+        // lpDialogLoaded means settled, not succeeded: leaving it false parks the tab in a spinner
+        // no retry can clear, while a pending half-deposit still renders through it. Settled, the
+        // tab falls through to its no-positions container and its Manage Positions retry.
+        assertTrue(vm.state.value.lpDialogLoaded)
+        // availablePools stays null so the next interaction re-fetches instead of soft-locking.
+        assertTrue(vm.state.value.lpPositionsDialog.isEmpty())
```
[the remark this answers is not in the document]
aminsato: Fixed in 5806f6d46 — both are `shouldBe true` now. `shouldBe` was already imported; I'd matched the `assertTrue`/`assertFalse` the original test used, but Kotest is the stated convention.

aminsato:
@coderabbitai — the two outside-diff findings, since they have no inline thread:

**`ThorchainDefiPositionsViewModel.kt` 514-521 — stale pending deposits during the resume scan.** Valid, fixed in 5806f6d46, though not quite as proposed. Resetting `pendingDepositsLoaded` unconditionally would flash the entire LP tab to a spinner on every resume for every user, including the large majority who have no pending deposit and therefore nothing stale to hide. The reset is now conditional on `pendingDeposits.isNotEmpty()` — a card that could have been refunded while backgrounded goes behind the spinner until the scan confirms it, and a tab with nothing pending is left alone. Your second point was already handled: `markPendingLpDepositsSettled()` only touches the flag, never the list, so a failed refresh keeps whatever the last successful scan found.

**`DefiTabUiModels.kt` 51-85 — `@Immutable` on the LP model graph.** Not doing this here. Neither `LpTabUiModel` nor `PendingLpDepositUiModel` carries `@Immutable` today and this PR does not add it — the only change to that file is one `Boolean` field. The work you describe is real but codebase-wide: `ImageModel` is `typealias ImageModel = Any` and `UiText` takes `List<Any>`, so constraining them reaches well beyond the DeFi models. That belongs in its own change, not in a PR fixing THORChain LP half-deposits.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/AddLiquidityStrategy.kt
```diff
@@ -69,25 +69,29 @@ internal class AddLiquidityStrategy(
...
+        val isSymmetricPool = assetChain != null && assetChain != Chain.ThorChain
+
         // Preflight against THORChain network state — pool status and the relevant mimir pause
         // keys. Refuses to build the keysign payload when the network would refund the inbound,
-        // sparing the user the inbound gas spend.
-        if (chain == Chain.ThorChain) {
+        // sparing the user the inbound gas spend. It asks about the pool, not about the inbound's
+        // source chain, so it must also cover the asset-side add that completes a pending
+        // half-deposit — that inbound is refundable on exactly the same terms.
+        if (chain == Chain.ThorChain || isSymmetricPool) {
```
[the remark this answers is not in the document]
aminsato: Confirmed and fixed in 90bd34f90.

Traced the path: `MayachainDefiPositionsViewModel.onNavigateToLp` routes with `chainId = Chain.MayaChain.id`, which lands on `DepositOption.AddLiquidity` and therefore this strategy. A Maya pool id like `BTC.BTC` parses to `Chain.Bitcoin`, so `isSymmetricPool` was true and the guard fired — asking thornode about a pool it does not own. `PAUSELP` is global on THORChain, and `BTC.BTC` exists on both networks, so a THOR-side pause or a `Staged` THOR pool would have rejected a valid CACAO add.

Reverting to `chain == Chain.ThorChain` would drop what the guard was widened for, though: the asset-side inbound that completes a pending half-deposit is on e.g. `Chain.Ethereum`, and it is refundable on exactly the same terms. That leg always arrives on the pool's own asset chain, so the guard now keys on that:

```kotlin
val isThorChainLpAdd = chain == Chain.ThorChain || (isSymmetricPool && chain == assetChain)
```

Maya never satisfies it (`chain` is `MayaChain`, `assetChain` is `Bitcoin`), and the completion leg still does. `isSymmetricPool` stays as-is for the paired-address requirement below.

aminsato:
On the outside-diff finding for `LiquidityDataLoader.kt:238-243` (keep Maya CACAO removals single-leg) — not applying it. The Maya asset leg is there deliberately: @rkokhatskyi asked for it in https://github.com/vultisig/vultisig-android/pull/5637#discussion_r3812273222, on the grounds that a symmetric Maya withdrawal returns both halves and users were never shown the paired asset amount they were about to receive. It landed in a259ca0b0, and the "Maya pools withdraw CACAO only" comment that had justified the empty symbol was removed with it, since that premise was wrong.

The PR description line about Maya staying single-leg predates that commit and is now stale — updating it.

The inline finding on `AddLiquidityStrategy.kt` was valid and is fixed in 90bd34f90; details in the thread.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/AddLiquidityStrategy.kt
```diff
@@ -69,25 +69,33 @@ internal class AddLiquidityStrategy(
...
-            if (assetChain != null && assetChain != Chain.ThorChain) {
-                throw InvalidTransactionDataException(
-                    UiText.StringResource(R.string.send_error_no_address)
-                )
-            }
+        // Either half of a symmetric add MUST name the other half's address, or THORChain opens a
+        // separate asymmetric position rather than crediting the pair: a RUNE-side add carries the
+        // asset address, and the asset-side add that completes a pending half-deposit carries the
+        // RUNE address.
+        if (isSymmetricPool && pairedAddress == null) {
```
rkokhatskyi: Widening this to `isSymmetricPool` breaks every Maya add-liquidity, not just the asset-side completion it was meant for: `resolvePairedAddress`'s `when` has no `Chain.MayaChain` branch, so it always returns null for Maya, and `isSymmetricPool` is true for any real Maya pool like BTC.BTC. The preflight guard right above already solved this with `isThorChainLpAdd` — reuse it here.

```kotlin
if (isThorChainLpAdd && pairedAddress == null) {
```

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -489,9 +501,119 @@ constructor(
...
+                    it.copy(lp = it.lp.copy(pendingDeposits = models, pendingDepositsLoaded = true))
+                }
+            }
+    }
+
+    /**
+     * Marks the scan as settled without touching [LpTabUiModel.pendingDeposits]. A failed reload
+     * must not erase a list an earlier one found — the refund timer is still running on it.
+     */
+    private fun markPendingLpDepositsSettled() {
```
rkokhatskyi, line 561: On a failed resume re-scan this marks the scan settled without refreshing `pendingDeposits`, so a stale, possibly-already-refunded card keeps its `canComplete` button enabled. Nothing downstream re-checks freshness: `onClickCompletePendingLp` reads the stale list directly, and `thorChainLpPreflight` only checks pool-wide pause/status, not this specific record.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
rkokhatskyi: Confirmed fixed in a259ca0b0 for the asset-side completion path this raised. `resolvePairedAddress` now resolves the RUNE address when completing from the asset chain, and deposit-option seeding no longer crashes on chains outside `SECURE_ASSETS_TICKERS`. A related regression from the same commit surfaced separately at `AddLiquidityStrategy.kt:95`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainPendingLpDepositsUseCase.kt
rkokhatskyi: Confirmed fixed. `toPendingFetch` no longer excludes records with existing units, so a returning LP's top-up is now detected as pending.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/load/LiquidityDataLoader.kt
rkokhatskyi: Confirmed fixed. `loadRemoveLpData` now reads `pool.assetDepth` and populates the asset leg for Maya, sharing the pool-wide unit divisor with the CACAO leg. Checked the 8-decimal conversion against live Maya Midgard data — `assetDepth` is genuinely 1e8 while `cacaoDepth` is 1e10, so the two legs correctly use different decimals.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
rkokhatskyi: The original gap is fixed, `onScreenResumed` now re-scans on `ON_RESUME`. Found a narrower residual in the failure path, posted separately at `ThorchainDefiPositionsViewModel.kt:556`.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/thorchain/ThorchainDefiPositionsScreen.kt
rkokhatskyi: Confirmed fixed. The tab now gates on both `lpDialogLoaded` and `pendingDepositsLoaded`, and forces `pendingDeposits` empty while either is unloaded, so this can't flash anymore.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/maya/RemoveLpScreen.kt
rkokhatskyi: Confirmed fixed, reverted to `body.m.medium`.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainPendingLpDepositsUseCase.kt
rkokhatskyi: Confirmed fixed at 100,800, matches live thornode mimir.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/LpTabComponents.kt
rkokhatskyi: Confirmed fixed. `loadLpPositionsForDialog`'s catch now sets `lpDialogLoaded=true`, and the tab's loading branch forces `pendingDeposits` empty so a card can't render through it.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/GetThorChainPendingLpDepositsUseCase.kt
rkokhatskyi: Confirmed fixed, the fallback and its comment are gone and `fetchPending` is a single RUNE-address lookup.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/RemoveLpCalculator.kt
rkokhatskyi: Confirmed fixed. Checked every current call site, none can trip it (max is 8+8=16).

app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/AddLiquidityStrategy.kt
aminsato: Fixed in c8d2220fc. Reused `isThorChainLpAdd` as you asked, with `isSymmetricPool` kept alongside it:

```kotlin
if (isSymmetricPool && isThorChainLpAdd && pairedAddress == null) {
```

`isThorChainLpAdd` on its own would swap one break for another: it is true for any `chain == Chain.ThorChain` add, including a `THOR.*` pool where `assetChain == ThorChain`, so `isSymmetricPool` is false and `resolvePairedAddress` correctly returns null because there is no paired side to name. That pool would then start failing with "no address". The conjunction reduces to "symmetric pool, inbound on either of its two halves", which is exactly the set that must carry the paired address, and Maya satisfies neither half.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed in c8d2220fc. `markPendingLpDepositsSettled` now clears `canComplete` on everything it carries forward, so a failed scan leaves the card visible but not actionable:

```kotlin
pendingDeposits = it.lp.pendingDeposits.map { deposit -> deposit.copy(canComplete = false) },
pendingDepositsLoaded = true,
```

Kept the card rather than dropping it — its refund timer is real and the user still needs to know the deposit exists — but withdrew completion, since the scan is the only thing that would have caught a refund and it did not run. A later successful scan rebuilds the list and recomputes `canComplete`, so it re-arms on the next good resume.

Also added the same check in `onClickCompletePendingLp` rather than relying on the disabled button alone, since that is the path you pointed at as reading the list directly.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -489,9 +501,135 @@ constructor(
...
+                    .formatTokenAmount(runeTicker)
+            } else {
+                CoinType.THORCHAIN.toValue(pendingAsset)
+                    .stripTrailingZeros()
+                    .formatTokenAmount(assetTicker)
+            }
+
+        return PendingLpDepositUiModel(
+            poolId = pool,
+            icon =
```
rkokhatskyi, line 626: This card's icon and chain logo always show the pool's asset side, even when RUNE is the side still awaited. When the asset deposits first, the title correctly says "Waiting for matching RUNE deposit" but the icon next to it still shows the asset's logo, not RUNE's.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/AddLiquidityStrategy.kt
rkokhatskyi: Confirmed fixed. Traced all 4 cases by hand: THOR RUNE-side add, asset-side completion, Maya add, and a THOR.*-only pool — the combined `isSymmetricPool && isThorChainLpAdd` guard threads all of them correctly. My suggested one-liner would have broken THOR.*-only pools, gating on `isThorChainLpAdd` alone catches both.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
rkokhatskyi: Confirmed fixed. Every held card is disarmed on a failed re-scan, a later successful scan re-derives canComplete from scratch, and onClickCompletePendingLp now refuses to route on a disarmed card independent of the button state. Worth a look separately: a disarmed-by-failed-scan card renders identically to a permanently-blocked one, no visual distinction, but it's self-healing and non-blocking.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Good catch — the icon was pinned to the pool asset while the title and the icon's own `fallbackTicker` follow the awaited side, so an asset-first deposit showed the asset logo next to "Waiting for matching RUNE deposit".

Icon and chain badge now follow the awaited side: RUNE logo + THORChain badge when RUNE is missing, pool asset + its chain badge when the asset is. Fixed in ebdf3f75b.

rkokhatskyi (approved):

PR #5638, opened 2026-08-18
feat(defi): bond bRUNE / unbond ybRUNE and show its position card

### LINKED ISSUE #5584: [Add] bRUNE stake / ybRUNE unstake — Rujira liquid bond wasm flow [DeFi]
## Problem
Android has no bRUNE staking path. This issue is the transaction flow only — the position card is a separate issue (linked below).

## Spec
Rujira liquid-bond contract `thor179fex2rxd45caedmz4hxsnu42sw20lu0djyh4yukyh965sq8muuqptru2g`:

- **Stake**: wasm execute `{"liquid":{"bond":{}}}`, funds `x/brune` (8 decimals) → mints ybRUNE
- **Unstake**: wasm execute `{"liquid":{"unbond":{}}}`, funds `x/staking-x/brune` (8 decimals) → returns bRUNE

No memo — these are wasm executes, not deposit memos. Unstake max = the vault's ybRUNE bank balance.

Same wasm shape as the sTCY auto-compounder Android already implements.

## Refs
- Windows: `core/ui/vault/deposit/staking/resolvers/brune.ts` (vultisig/vultisig-windows#4395)
- iOS: `BRUNEStakeTransactionBuilder.swift` (vultisig/vultisig-ios#4778)

### LINKED ISSUE #5585: [Add] ybRUNE staked-position card with NAV-based value [DeFi]

aminsato:
## Summary

Ships the Rujira bRUNE liquid bond: a working bond / unbond flow, and a home on the DeFi tab for the position the ybRUNE receipt stands for. Two standalone commits, one per issue — each compiles and passes both suites on its own.

Closes #5584
Closes #5585

Transction hash
https://runescan.io/tx/8691634A81D44036B42931A9AD0F57A8B89DE7D7391113D5D1ADF24A35FED61E

**Depends on #5635**, which takes the ybRUNE receipt out of the wallet token list (#5583). That change used to be the first commit here; it now ships on its own so this PR carries only the bond flow and the card. Merge #5635 first — until it lands, a holder sees the position twice (once as a wallet token, once as the card below).

### 1. Bond bRUNE / unbond ybRUNE (#5584)

- `stakeBRune` emits `{"liquid":{"bond":{}}}` funded with `x/brune`; `unstakeBRune` emits `{"liquid":{"unbond":{}}}` funded with the receipt, both against `thor179fex…`.
- No share conversion on the way in. The sRUJI redemption converts because that position is reported in RUJI; this one is reported in receipt units, so the amount the form carries is already what funds the execute.
- `AccountsLoader` synthesizes the unbond form's account off the RUNE account (address, derived key, gas), since the receipt is not a wallet token. The bond direction spends bRUNE and stays on the ordinary wallet accounts.
- The contract address moves to `ThorchainStakingContracts`: two layers now name it — the executes built here, and the `{"status":{}}` NAV the pricing repository already read to value the receipt.

### 2. Compounded bRUNE card (#5585)

- ybRUNE joins the bank-denom positions `DefaultStakingPositionService` reads and passes both DeFi-balance filters, so it reaches the chain row and the portfolio total as well as its own card.
- Value comes from the NAV price row the receipt already had: balance × (`liquid_bond_size` / `liquid_bond_shares`) × RUNE.
- Selectable in Manage Positions. A vault that already saved a selection keeps it, so the card is opt-in there — as with every position added since that dialog shipped.

## iOS parity

The transaction layer is taken from iOS (`BRUNEStakeTransactionBuilder` / `BRUNEUnstakeTransactionBuilder`, vultisig/vultisig-ios#4778): same contract, same execute messages, same funding denoms, same "no conversion" rule, and the same position read (`fetchBRuneAutoCompoundAmount`). iOS likewise lists ybRUNE — but not bRUNE — among the selectable staking positions.

Two deliberate differences:

| | iOS | here |
|---|---|---|
| Card title | `"Compounded ybRUNE"` (position ticker) | `"Compounded bRUNE"` — the string #5585 asks for, and how the sibling sRUJI card is titled |
| Unstake input | percentage slider on the bRUNE coin | amount field on ybRUNE, MAX at the position — matches the sRUJI unstake and labels the units actually sent |

The signed bytes are the same either way.

## Test plan

- `./gradlew :data:testDebugUnitTest :app:testDebugUnitTest` — 2963 + 2087 green.
- Both commits checked out separately, compiled and run through both suites.
- Merged with #5635 locally: no conflicts, both suites green on the merge result.
- 19 new tests; every one verified to fail when its production change is reverted. They cover the wasm payloads (denoms, messages, guard rails), the synthesized unbond account, the preselection and nav actions, the strategy submits, the position read, and the card's title/flags.

Manual: DeFi tab → THORChain → Manage Positions → tick **ybRUNE** → Done. The "Compounded bRUNE" card shows the position; **Stake** opens the form titled "Stake" with bRUNE preselected, **Unstake** opens it with ybRUNE preselected and MAX at the full position.

## Summary by CodeRabbit

* **New Features**
  * Added THORChain bRUNE staking and ybRUNE unstaking.
  * Added ybRUNE receipt balances and positions to DeFi views.
  * Added automatic token selection, navigation, and transaction handling for bRUNE and ybRUNE.
  * Added updated staking visuals and preserved THORChain position selections when new positions are introduced.

* **Bug Fixes**
  * Improved handling of missing, stale, zero-value, and refreshed ybRUNE balances.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
@@ -903,17 +903,16 @@ constructor(
                                 } else {
                                     R.string.defi_header_staked
                                 }
-                            // The bonded position is named after what was bonded, not after the
-                            // receipt that tracks it: "Compounded bRUNE", the way the sRUJI card
-                            // is titled off RUJI.
-                            val headerTicker =
-                                if (isBondedRuneReceipt) Coins.ThorChain.bRUNE.ticker
-                                else coin.ticker
+                            // Titled in the same unit the amount below it is counted in. The
+                            // sRUJI card can say RUJI because its amount is the pool's RUJI
+                            // liquidSize; this one is the raw receipt balance, and a share is
+                            // worth more than one bRUNE, so naming it bRUNE would understate the
+                            // position by the compounding it exists to earn.
                             val position =
                                 StakePositionUiModel(
                                     coin = defaultPosition.coin,
                                     stakeAssetHeader =
-                                        UiText.FormattedText(headerResId, listOf(headerTicker)),
+                                        UiText.FormattedText(headerResId, listOf(coin.ticker)),
                                     stakedAmountDisplay =
                                         stakeAmount.formatTokenAmount(coin.ticker),
                                     // No .orEmpty(): a missed lookup means "we have no price",
@@ -1484,7 +1483,6 @@ constructor(
             val stcy = Coins.ThorChain.sTCY
             val ytcy = Coins.ThorChain.yTCY
             val yrune = Coins.ThorChain.yRUNE
-            val brune = Coins.ThorChain.bRUNE
             val ybrune = Coins.ThorChain.ybRUNE
 
             return listOf(
@@ -1567,7 +1565,10 @@ constructor(
                 StakePositionUiModel(
                     coin = ybrune,
                     stakeAssetHeader =
-                        UiText.FormattedText(R.string.defi_header_compounded, listOf(brune.ticker)),
+                        UiText.FormattedText(
+                            R.string.defi_header_compounded,
+                            listOf(ybrune.ticker),
+                        ),
                     stakedAmountDisplay = "0 ${ybrune.ticker}",
                     apy = null,
                     canWithdraw = false,
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
@@ -835,7 +835,8 @@ constructor(
                     settleStakingPositions {
                         it.coin.id == Coins.ThorChain.yRUNE.id ||
                             it.coin.id == Coins.ThorChain.yTCY.id ||
-                            it.coin.id == Coins.ThorChain.sTCY.id
+                            it.coin.id == Coins.ThorChain.sTCY.id ||
+                            it.coin.id == Coins.ThorChain.ybRUNE.id
                     }
                 }
                 .onCompletion { cause ->
@@ -888,21 +889,31 @@ constructor(
 
                             val canTransfer = coin.ticker.contains("stcy", ignoreCase = true)
 
+                            val isBondedRuneReceipt =
+                                coin.id.equals(Coins.ThorChain.ybRUNE.id, true)
+
                             val headerResId =
                                 if (supportsMint) {
                                     R.string.defi_header_minted
                                 } else if (
-                                    defaultPosition.coin.id.equals(Coins.ThorChain.sTCY.id, true)
+                                    defaultPosition.coin.id.equals(Coins.ThorChain.sTCY.id, true) ||
+                                        isBondedRuneReceipt
                                 ) {
                                     R.string.defi_header_compounded
                                 } else {
                                     R.string.defi_header_staked
                                 }
+                            // The bonded position is named after what was bonded, not after the
+                            // receipt that tracks it: "Compounded bRUNE", the way the sRUJI card
+                            // is titled off RUJI.
+                            val headerTicker =
+                                if (isBondedRuneReceipt) Coins.ThorChain.bRUNE.ticker
+                                else coin.ticker
                             val position =
                                 StakePositionUiModel(
                                     coin = defaultPosition.coin,
                                     stakeAssetHeader =
-                                        UiText.FormattedText(headerResId, listOf(coin.ticker)),
+                                        UiText.FormattedText(headerResId, listOf(headerTicker)),
                                     stakedAmountDisplay =
                                         stakeAmount.formatTokenAmount(coin.ticker),
                                     // No .orEmpty(): a missed lookup means "we have no price",
@@ -1406,6 +1417,9 @@ constructor(
                     DeFiNavActions.UNSTAKE_TCY -> Coins.ThorChain.TCY.id
                     DeFiNavActions.STAKE_STCY -> Coins.ThorChain.TCY.id
                     DeFiNavActions.UNSTAKE_STCY -> Coins.ThorChain.sTCY.id
+                    // Bonding spends bRUNE; only the unbond starts from the receipt.
+                    DeFiNavActions.STAKE_YBRUNE -> Coins.ThorChain.bRUNE.id
+                    DeFiNavActions.UNSTAKE_YBRUNE -> Coins.ThorChain.ybRUNE.id
                     DeFiNavActions.MINT_YTCY -> Coins.ThorChain.TCY.id
                     DeFiNavActions.REDEEM_YTCY -> Coins.ThorChain.yTCY.id
                     DeFiNavActions.MINT_YRUNE -> Coins.ThorChain.RUNE.id
@@ -1470,6 +1484,8 @@ constructor(
             val stcy = Coins.ThorChain.sTCY
             val ytcy = Coins.ThorChain.yTCY
             val yrune = Coins.ThorChain.yRUNE
+            val brune = Coins.ThorChain.bRUNE
+            val ybrune = Coins.ThorChain.ybRUNE
 
             return listOf(
                 StakePositionUiModel(
@@ -1540,6 +1556,19 @@ constructor(
                     coin = yrune,
                     stakeAssetHeader = UiText.StringResource(R.string.staked_yrune_header),
                     stakedAmountDisplay = "0 ${yrune.ticker}",
+                    apy = null,
+                    canWithdraw = false,
+                    canStake = true,
+                    canUnstake = false,
+                    rewards = null,
+                    nextReward = null,
+                    nextPayout = null,
+                ),
+                StakePositionUiModel(
+                    coin = ybrune,
+                    stakeAssetHeader =
+                        UiText.FormattedText(R.string.defi_header_compounded, listOf(brune.ticker)),
+                    stakedAmountDisplay = "0 ${ybrune.ticker}",
                     apy = null,
                     canWithdraw = false,
                     canStake = true,
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
@@ -887,10 +887,16 @@ constructor(
                                 coin.ticker.contains("yrune", ignoreCase = true) ||
                                     coin.ticker.contains("ytcy", ignoreCase = true)
 
-                            val canTransfer = coin.ticker.contains("stcy", ignoreCase = true)
-
                             val isBondedRuneReceipt =
                                 coin.id.equals(Coins.ThorChain.ybRUNE.id, true)
+
+                            // Both compounding receipts are plain THORChain bank denoms, so the
+                            // vault can move either one to another address — #5585 asks for it on
+                            // ybRUNE, and iOS offers Transfer on every compound position. Matched
+                            // on the coin id: a substring of the ticker is what left this card
+                            // with the wrong logo, since ybRUNE contains none of the others.
+                            val canTransfer =
+                                coin.id.equals(Coins.ThorChain.sTCY.id, true) || isBondedRuneReceipt
 
                             val headerResId =
                                 if (supportsMint) {
@@ -1448,14 +1454,16 @@ constructor(
         }
     }
 
-    fun onClickTransfer() {
+    /**
+     * Opens the plain send form on [coin], the position's own receipt.
+     *
+     * Carries the position's coin rather than a fixed one: sTCY was the only transferable card when
+     * this was written, and the ybRUNE receipt would otherwise have sent the wrong token.
+     */
+    fun onClickTransfer(coin: Coin) {
         viewModelScope.launch {
             navigator.route(
-                Route.Send(
-                    vaultId = vaultId,
-                    chainId = Chain.ThorChain.id,
-                    tokenId = Coins.ThorChain.sTCY.id,
-                )
+                Route.Send(vaultId = vaultId, chainId = coin.chain.id, tokenId = coin.id)
             )
         }
     }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
@@ -1,6 +1,7 @@
 package com.vultisig.wallet.ui.models.send
 
 import com.vultisig.wallet.data.blockchain.model.StakingDetails.Companion.generateId
+import com.vultisig.wallet.data.blockchain.thorchain.DefaultStakingPositionService
 import com.vultisig.wallet.data.blockchain.thorchain.RujiStakingService.Companion.RUJI_REWARDS_COIN
 import com.vultisig.wallet.data.models.Account
 import com.vultisig.wallet.data.models.Address
@@ -14,6 +15,7 @@ import com.vultisig.wallet.data.utils.safeLaunch
 import com.vultisig.wallet.ui.screens.v2.defi.model.DeFiNavActions
 import java.math.BigDecimal
 import java.math.BigInteger
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.Flow
@@ -26,6 +28,7 @@ internal class AccountsLoader(
     private val accountsState: MutableStateFlow<AccountsLoadState>,
     private val accountsRepository: AccountsRepository,
     private val stakingDetailsRepository: StakingDetailsRepository,
+    private val defaultStakingPositionService: DefaultStakingPositionService,
     private val defiTypeProvider: () -> DeFiNavActions?,
     private val mscaAddressProvider: () -> String?,
     private val bondedAmountProvider: () -> BigInteger?,
@@ -254,21 +257,55 @@ internal class AccountsLoader(
 
     /**
      * The ybRUNE receipt is not a wallet token either, so the unbond form has no account to draw on
-     * until one is synthesized from the position the DeFi tab cached.
+     * until one is synthesized from the vault's receipt balance.
      *
      * Denominated in receipt units — the vault's `x/staking-x/brune` bank balance, which is what
      * the position reports — so the amount the form carries is already what funds the unbond. No
      * conversion happens at submit, unlike the RUJI receipt above, whose position is reported in
      * RUJI.
+     *
+     * Cached first, then re-read from the chain, the way the balance rows around it hydrate. The
+     * cached figure is whatever the DeFi tab last stored, and this account is what MAX fills the
+     * form from: left at the cache, a bond made since that refresh would cap MAX below the true
+     * position and quietly redeem less than everything, with the submit-time clamp — which only
+     * ever lowers an amount — unable to notice.
      */
     private suspend fun loadBondedRuneReceiptAccount(vaultId: VaultId, generation: Long) {
-        val cachedDetails =
-            stakingDetailsRepository.getStakingDetailsByCoindId(vaultId, Coins.ThorChain.ybRUNE.id)
+        var receiptAmount =
+            stakingDetailsRepository
+                .getStakingDetailsByCoindId(vaultId, Coins.ThorChain.ybRUNE.id)
+                ?.stakeAmount
+        var hydrated = false
         accountsRepository
             .loadAddresses(vaultId)
             .map { addrs -> addrs.flatMap { it.accounts } }
             .collect { accountsLoaded ->
-                publishBondedRuneReceipt(accountsLoaded, cachedDetails?.stakeAmount, generation)
+                publishBondedRuneReceipt(accountsLoaded, receiptAmount, generation)
+
+                val address =
+                    accountsLoaded
+                        .find { it.token.id.equals(Coins.ThorChain.RUNE.id, true) }
+                        ?.token
+                        ?.address
+                if (hydrated || address.isNullOrEmpty()) return@collect
+                hydrated = true
+
+                // A failed read leaves the cached figure standing rather than emptying the form:
+                // the submit clamp reads the balance again and refuses to build on a guess.
+                val live =
+                    try {
+                        defaultStakingPositionService.getReceiptBalance(
+                            address,
+                            Coins.ThorChain.ybRUNE,
+                        )
+                    } catch (t: Throwable) {
+                        if (t is CancellationException) throw t
+                        Timber.e(t, "Failed to read the live ybRUNE receipt balance")
+                        return@collect
+                    }
+
+                receiptAmount = live
+                publishBondedRuneReceipt(accountsLoaded, live, generation)
             }
     }
 
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
@@ -7,6 +7,7 @@ import com.vultisig.wallet.data.models.Account
 import com.vultisig.wallet.data.models.Address
 import com.vultisig.wallet.data.models.Coins
 import com.vultisig.wallet.data.models.FiatValue
+import com.vultisig.wallet.data.models.TokenId
 import com.vultisig.wallet.data.models.TokenValue
 import com.vultisig.wallet.data.models.VaultId
 import com.vultisig.wallet.data.repositories.AccountsRepository
@@ -29,6 +30,7 @@ internal class AccountsLoader(
     private val stakingDetailsRepository: StakingDetailsRepository,
     private val defaultStakingPositionService: DefaultStakingPositionService,
     private val defiTypeProvider: () -> DeFiNavActions?,
+    private val preselectedTokenIdProvider: () -> TokenId?,
     private val mscaAddressProvider: () -> String?,
     private val bondedAmountProvider: () -> BigInteger?,
 ) {
@@ -88,10 +90,7 @@ internal class AccountsLoader(
                 DeFiNavActions.REDEEM_YTCY,
                 DeFiNavActions.FREEZE_TRX ->
                     scope.safeLaunch(onError = ::onLoadError) {
-                        accountsRepository
-                            .loadAddresses(vaultId)
-                            .map { addrs -> addrs.flatMap { it.accounts } }
-                            .collect { publishLoaded(it, generation) }
+                        loadWalletAccounts(vaultId, generation)
                     }
 
                 else ->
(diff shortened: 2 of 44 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/send/submit/UnstakeStrategy.kt
```diff
@@ -154,6 +155,17 @@ internal class UnstakeStrategy(
...
+                            DeFiNavActions.UNSTAKE_YBRUNE ->
+                                createBRuneUnstakeDepositTransaction(
+                                    vaultId = vaultId,
+                                    selectedToken = selectedToken,
+                                    srcAddress = srcAddress,
+                                    dstAddress = dstAddress,
+                                    tokenAmountInt = tokenAmountInt,
+                                    gasFee = gasFee,
+                                    chain = chain,
+                                )
```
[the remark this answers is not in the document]
aminsato: Verified and fixed in `6a0c11e8c`.

Confirmed: the ybRUNE account balance comes solely from `stakingDetailsRepository` — `AccountsLoader.publishBondedRuneReceipt` synthesizes it from `cachedDetails?.stakeAmount`, read once when the form opens. Now that the receipt is no longer a wallet token, nothing hydrates it between DeFi refreshes, so the ceiling lags the bank balance in both directions: a bond made since the last refresh leaves it too low and rejects an amount the vault can honour, and an unbond made since leaves it too high — and `liquid.unbond` carries the receipt as *funds* on the execute, so too high means the chain refuses the transaction only after the user has paid for the keysign.

The redemption now takes the same treatment the sRUJI one already has in this file: `UNSTAKE_YBRUNE` skips the cached ceiling, and `createBRuneUnstakeDepositTransaction` reads the live `x/staking-x/brune` balance via `thorChainApi.getBalance` — the same denom `DefaultStakingPositionService` builds the position card from, so the two can never disagree — and clamps the entered amount to it.

Clamp rather than reject, matching the sRUJI precedent: rejecting would dead-end MAX whenever the cache reads high, since the form fills it from that same stale figure. The read fails closed — an unreadable balance raises the generic error rather than being taken as an empty position, since sizing a redemption off a guess is what the clamp exists to prevent. A denom absent from the response is a genuine zero (the bank drops empty balances) and stops the submit with `send_error_insufficient_balance`.

4 new tests in `UnstakeStrategyTest`, including your requested cached-vs-live case: cached ceiling 1.0 / live 0.25 / entered 0.4 clamps to 25000000, and the mirror case where the live balance is above a stale-low ceiling. Both verified to fail on revert.

app/src/main/java/com/vultisig/wallet/ui/models/send/TokenPreselectionService.kt
```diff
@@ -167,6 +167,11 @@ internal class TokenPreselectionService(
             DeFiNavActions.WITHDRAW_USDC_CIRCLE -> Coins.Ethereum.USDC
             DeFiNavActions.STAKE_STCY -> Coins.ThorChain.TCY
             DeFiNavActions.UNSTAKE_STCY -> Coins.ThorChain.sTCY
+
+            // The liquid bond is funded with bRUNE and redeemed out of the ybRUNE receipt, so the
+            // two directions start from different tokens — as with the RUJI pair above.
+            DeFiNavActions.STAKE_YBRUNE -> Coins.ThorChain.bRUNE
+            DeFiNavActions.UNSTAKE_YBRUNE -> Coins.ThorChain.ybRUNE
```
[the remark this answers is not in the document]
aminsato: Verified and fixed in `6a0c11e8c`, taking your diff as written.

The path holds: `AccountsLoader.publishBondedRuneReceipt` synthesizes the receipt account on top of the vault's RUNE account and calls `findSourceOrPublishEmpty`, so a vault with no THORChain account publishes `Loaded(emptyList())`. `defaultDefiCoin()` then returned the static `Coins.ThorChain.ybRUNE` template, and `SendFormGraph.collectSelectedAccount` falls back to `Account(token, tokenValue = null, ...)` for a token the account list does not carry — leaving a complete-looking Asset row with a blank balance, an empty address and an empty `hexPublicKey`, which is exactly the gap the guard above was written to close for sRUJI.

`UNSTAKE_YBRUNE` joins that branch, and the comment above it now names both receipt redemptions rather than sRUJI alone. Covered by a new `TokenPreselectionServiceTest` case mirroring the sRUJI one, verified to fail on revert.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -888,21 +889,31 @@ constructor(
 
                             val canTransfer = coin.ticker.contains("stcy", ignoreCase = true)
 
+                            val isBondedRuneReceipt =
```
rkokhatskyi, line 890: `StakingTabComponents.kt:308`'s getHeaderIcon matches stakedAmountDisplay against yrune/ytcy/stcy/ruji/tcy/cacao and falls to R.drawable.om otherwise, but ybRUNE contains none of those substrings since the b breaks yrune. Every Compounded bRUNE card renders the unrelated Mantra logo instead of bRUNE's, unlike CoinLogo.kt which already maps both.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiExtensions.kt
```diff
@@ -36,6 +37,7 @@ internal val thorchainSupportStakingDeFi: List<Coin>
             Coins.ThorChain.sTCY,
             Coins.ThorChain.yRUNE,
             Coins.ThorChain.yTCY,
+            Coins.ThorChain.ybRUNE,
```
rkokhatskyi, line 40: This adds ybRUNE to defaultSelectedPositionsDialog(), but coinsToLoad (`ThorchainDefiPositionsViewModel.kt:672`) filters cards by the persisted selectedPositions set, which DefiPositionsRepository returns verbatim with no reconciliation. Any vault with a selection saved before this PR lacks a bRUNE entry, so a real bond mints ybRUNE while the card and tab total stay silently hidden.

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -245,6 +252,56 @@ internal class AccountsLoader(
...
+     * The ybRUNE receipt is not a wallet token either, so the unbond form has no account to draw on
+     * until one is synthesized from the position the DeFi tab cached.
+     *
+     * Denominated in receipt units — the vault's `x/staking-x/brune` bank balance, which is what
+     * the position reports — so the amount the form carries is already what funds the unbond. No
+     * conversion happens at submit, unlike the RUJI receipt above, whose position is reported in
+     * RUJI.
+     */
+    private suspend fun loadBondedRuneReceiptAccount(vaultId: VaultId, generation: Long) {
+        val cachedDetails =
```
rkokhatskyi: MAX on the ybRUNE unbond fills from this cached balance, read once at screen load, while `UnstakeStrategy.kt:415` only clamps the typed amount down to a fresh live balance, never up. If the live position grew since the cache was taken, MAX silently redeems less than the true maximum, with no warning.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -888,21 +889,31 @@ constructor(
...
+                                        isBondedRuneReceipt
                                 ) {
                                     R.string.defi_header_compounded
                                 } else {
                                     R.string.defi_header_staked
                                 }
+                            // The bonded position is named after what was bonded, not after the
+                            // receipt that tracks it: "Compounded bRUNE", the way the sRUJI card
+                            // is titled off RUJI.
+                            val headerTicker =
```
rkokhatskyi: The ybRUNE card titles itself Compounded bRUNE via this headerTicker override, but stakedAmountDisplay a few lines below still formats with the raw coin.ticker, so the same card shows Compounded bRUNE above an amount reading ybRUNE. The sRUJI card keeps both in RUJI.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/ThorchainFunctions.kt
```diff
@@ -208,6 +260,7 @@ private object Memo {
     const val TCY_UNSTAKE_PREFIX = "TCY-:"
     const val DENOM_STAKING_TCY = "x/staking-tcy"
     const val DENOM_STAKING_RUJI = "x/staking-x/ruji"
+    const val DENOM_STAKING_BRUNE = "x/staking-x/brune"
```
rkokhatskyi: This hardcodes x/staking-x/brune separately from Coins.ThorChain.ybRUNE.contractAddress, though both live in the same module and the contract address just above already got a shared constant for this. An edit to one without the other would fund the unbond execute with a denom the balance clamp no longer matches, silently breaking unstake.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Verified and fixed in `aa87d7774`.

Confirmed exactly as described: `ybRUNE` lowercases to `ybrune`, which holds none of `yrune` / `ytcy` / `stcy` / `ruji` / `tcy` / `cacao` — the `b` breaks `yrune` — so every compounded card fell to `R.drawable.om`, while `CoinLogo.kt:146` has mapped both `brune` and `ybrune` to `R.drawable.brune` since #5387.

Fixed the mismatch rather than only the symptom: the lookup now takes the position's own `coin.ticker` instead of `stakedAmountDisplay`, and `ybrune` maps to `R.drawable.brune`. The display string is only ever `<amount> <ticker>`, so no existing card changes drawable — but the formatted amount was never the right thing to pattern-match, and it is what let this slip through. The prefixed tickers stay ahead of the asset they receipt, or `ytcy`/`stcy` would both match `tcy`.

New `StakingTabHeaderIconTest` pins ybRUNE→brune and the six that were already working, verified to fail on revert. Between this and the two screens that use the widget, the `om` fallback is now unreachable for every position either one can carry.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiExtensions.kt
aminsato: Verified and fixed in `aa87d7774`. This was the worst of the five — a real bond minting a receipt the tab never showed.

The path holds end to end: `loadSavedPositions` treats only `null` as "never chose" and takes the stored set verbatim otherwise, `coinsToLoad` filters `thorchainSupportStakingDeFi` by it, and `DefiPositionsRepository` has no reconciliation — `THORCHAIN_DEFAULT_POSITION_KEYS` in the per-chain migration is deliberately frozen at that migration's shipped shape, so it cannot serve. Adding ybRUNE to the defaults only ever reaches a vault that has never opened the dialog.

Added `NewThorchainPositionsMigration`: positions introduced after vaults could already have saved, keyed by the version that introduced them (`mapOf(1 to setOf("ybRUNE"))`), switched on for every stored THORChain selection. Notes on the shape:

- A version counter in the same store, rather than "add it if missing". Without it this would be a standing rule that switched the position back on at the next cold start after the user had switched it off. `shouldMigrate` is consulted on every store creation.
- It runs after `SelectedPositionsPerChainMigration` in the same pass, so a vault whose selection is still in the legacy blob gets moved first and then topped up. DataStore feeds each migration the previous one's output.
- A selection cleared to nothing is skipped. That shape is a deliberate "show me none", and handing it a row the user never asked for is the same override in the other direction.
- A vault that never chose is left keyless, so it keeps reading back as `null` and takes the screen's defaults — which already carry ybRUNE.
- The next position to join the dialog is a one-line addition under a new version.

8 tests in `DefiPositionsRepositoryMigrationTest`; 4 of them fail on revert (the other 4 pin the shapes it must not touch).

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
aminsato: Verified and fixed in `aa87d7774`.

Confirmed: the account MAX fills from is synthesized here off `stakingDetailsRepository`, read once when the form opens, and `UnstakeStrategy` only ever `coerceAtMost`s down to the live balance. A bond made since the last DeFi refresh therefore capped MAX below the position and redeemed less than everything, silently — the clamp cannot see it, and this receipt has nothing else hydrating it now that it is no longer a wallet token.

Fixed at the ceiling rather than by teaching the clamp to raise: an amount the user typed should never be raised, only MAX should grow, and the strategy has no way to tell the two apart. So the synthesized account now hydrates from the chain the way the balance rows around it do — the cached figure is published first so the form paints without waiting on the network, then the live `x/staking-x/brune` balance replaces it. That also fixes the mirror case the clamp could never reach: a stale-high cache no longer offers a ceiling the vault cannot honour.

A failed read leaves the cached figure standing rather than emptying the form — a fetch that failed says nothing about the position, and the submit-time read still fails closed, so nothing can be built on a guess.

The read itself moved to `DefaultStakingPositionService.getReceiptBalance`, which is what builds the position card, so the card, this ceiling and the submit clamp now size the receipt through one place instead of three (see also your `ThorchainFunctions` comment). 2 new `AccountsLoaderTest` cases — live replaces a stale-low cache, and a failed read keeps the cache — the first verified to fail on revert.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Verified and fixed in `aa87d7774` — though I went the other way from the sRUJI parity, and want to flag why.

The disagreement is real: header "Compounded bRUNE" over an amount formatted with `coin.ticker`, "ybRUNE". But sRUJI is not the precedent it looks like. That card can say RUJI on both lines because its amount *is* RUJI — `rujiPositionUiModel` formats the pool's `liquidSize` with `RUJI_SYMBOL`, not a share count. ybRUNE's amount is the raw `x/staking-x/brune` bank balance, and a share is worth more than one bRUNE and drifts further with every compounding. Titling that "bRUNE" would have understated the position by exactly the yield the position exists to earn — and it is also what funds the unbond execute, so the number has to stay in receipt units.

So both lines now name the receipt. That also matches iOS, which titles the same card off `position.coin.ticker` (`DefiChainStakedPositionView.defaultTitle`, `.compound` → "Compounded ybRUNE") over the same raw receipt amount, and the Manage Positions tile, which is labelled `ybRUNE`. The zero placeholder in `loadDefaultStakingPositions` had the same split and is fixed with it.

The alternative was to convert the amount into bRUNE at the bond's NAV — `TokenPriceRepository` already reads `navPerShare(BRUNE_STAKING_CONTRACT)` to price the receipt — and keep the bRUNE title. I did not take it: the unbond would then have to convert back to shares on submit, which is the rounding-sensitive dance the sRUJI redemption has to do and this one deliberately avoids. Say the word if you would rather have the bRUNE-denominated card and I will do it properly, conversion and all.

Existing VM test updated to assert header and amount name the same unit, verified to fail on revert.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/ThorchainFunctions.kt
aminsato: Verified and fixed in `aa87d7774`. `Memo.DENOM_STAKING_BRUNE` is now `Coins.ThorChain.ybRUNE.contractAddress`, so the denom that funds the execute and the denom the balance clamp measures cannot drift apart — same reasoning as `ThorchainStakingContracts.BRUNE_LIQUID_BOND` above it, and `ThorchainFunctionsTest` already asserted against `ybRUNE.contractAddress`, so it was only the literal that was loose.

The clamp side of that pairing also collapsed into one place while I was here: `UnstakeStrategy` no longer matches the denom itself but reads through `DefaultStakingPositionService.getReceiptBalance`, which is the same service the position card is built from.

I left `DENOM_STAKING_TCY` and `DENOM_STAKING_RUJI` as literals — same hazard, but they predate this PR and are not what it touches. Happy to convert both in a follow-up if you want the object consistent.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
rkokhatskyi: Fixed. getHeaderIcon now keys off coin.ticker with an explicit ybrune branch ordered ahead of the shorter substrings, verified against all 8 tickers with no collisions and pinned by StakingTabHeaderIconTest.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/DeFiExtensions.kt
rkokhatskyi: Fixed with NewThorchainPositionsMigration. Verified the migration-chaining order, the version gate's re-disable-survives-relaunch case, and that a never-chose vault is left alone.

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/ThorchainFunctions.kt
rkokhatskyi: Fixed, confirmed the single call site and that val vs const val is inert here.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
rkokhatskyi: Went the other direction from what I suggested and it holds up. Verified sRUJI's amount really is NAV converted before display while ybRUNE's is raw receipt units, and iOS titles this same card off the receipt ticker too.

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
rkokhatskyi: The live hydrate narrows this a lot, but hydrated is set true before the try/catch around the read, so one transient failure permanently disables the retry for the rest of the sheet visit and MAX can silently undershoot again. New comment below.

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -245,6 +255,90 @@ internal class AccountsLoader(
...
+            .collect { accountsLoaded ->
+                publishBondedRuneReceipt(accountsLoaded, receiptAmount, generation)
+
+                val address =
+                    accountsLoaded
+                        .find { it.token.id.equals(Coins.ThorChain.RUNE.id, true) }
+                        ?.token
+                        ?.address
+                if (hydrated || address.isNullOrEmpty()) return@collect
+                hydrated = true
```
rkokhatskyi: hydrated = true here runs before the try/catch below it, so once the one live read throws, hydration never retries on later loadAddresses emissions for that sheet session. MAX keeps filling from the stale cache, and since UnstakeStrategy's coerceAtMost only lowers the amount, a grown position is silently undershot until the sheet reopens.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -888,16 +889,25 @@ constructor(
 
                             val canTransfer = coin.ticker.contains("stcy", ignoreCase = true)
```
rkokhatskyi: canTransfer only matches tickers containing stcy, so the new ybRUNE card never satisfies the gate that shows the Transfer button at StakingTabComponents.kt:206. Issue #5585 lists Send as a card action, and iOS's CoinService.addToChain genuinely wires it for this same compound position, so Android users have no way to move the receipt except unbonding it.

rkokhatskyi (changes requested):

johnnyluo (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
aminsato: Verified and fixed in `f372964cf`.

Confirmed exactly as described: `hydrated = true` sat above the `try`, and the `catch` does `return@collect`, so the flag was already set when the read failed. Every later `loadAddresses` emission then took the `if (hydrated ...) return@collect` early exit — hydration was retired for the rest of the sheet visit, not just for that emission. MAX went on filling from the cached `StakingDetails` ceiling, and since `UnstakeStrategy` only ever `coerceAtMost`s, a position grown since the last DeFi refresh was silently undershot with no way back short of reopening the sheet.

The flag now moves below the read, so it records "a read landed", not "a read was attempted". The failure path is otherwise unchanged: the cached figure still stands rather than emptying the form, and the submit-time read still fails closed.

Two tests, both verified to fail on revert:
- a first read that throws followed by a second emission ends on the live balance
- once a read succeeds, later emissions do not refetch (`coVerify(exactly = 1)`) — the retry must not turn into a fetch per emission

Thanks for catching this one; it was created by the fix for your last round's finding on the same file.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Verified and fixed in `f372964cf`. Both of your references check out, and the gate was only the first of three things standing in the way.

`canTransfer` was `coin.ticker.contains("stcy")`, so `ybRUNE` never matched. #5585 does list it — "**Send** moves the ybRUNE receipt token, not liquid bRUNE" — and iOS renders `PrimaryButton("transfer")` unconditionally for `.compound` positions (`DefiChainStakedPositionView.swift:208`), resolving the coin through `CoinService.addToChain` having put it in `vault.coins`. The gate now matches on coin id (`sTCY` or the receipt) rather than a ticker substring, for the same reason `getHeaderIcon` moved off substrings last round.

Two things the button alone would have hit:

1. **The form had no account to send from.** `onClickTransfer` was pinned to `Coins.ThorChain.sTCY.id`, so the ybRUNE card would have opened the send form on someone else's token. It now carries the position's own coin. But even then, a plain send loads through `loadAddresses` (built from `vault.coins`, which the receipt is deliberately kept out of), and every DeFi path drops it via `spendableAccounts`' `Coins.isDefiOnly` filter — so the form would have had no ybRUNE account at all. `AccountsLoader` now takes the route's preselected token id: a plain send opened on the receipt publishes the wallet accounts **with** the synthesized receipt appended, hydrated live like the unbond form's. No other send changes, and a vault without a THORChain account keeps its wallet accounts rather than being emptied — the rest is still perfectly sendable.

2. **It would then have crashed at keysign.** `KeysignShareViewModel.loadTransaction` re-resolves the coin with `vault.coins.find { ... }!!` — a row this receipt can never have, and one that `aa7806109` actively migrates away for vaults that picked it up while it was still discoverable. Tapping Continue would have NPE'd. It now falls back to the token the form staged (which carries the chain's own address and derived key) for `Coins.isDefiOnly` coins only, and fails with a message rather than a bare `!!` otherwise. The unbond path was never affected: deposits sign `srcToken` directly.

Signing itself needed nothing — `ThorChainHelper.buildThorchainSendMessage` already uses `contractAddress` as the denom for non-native coins, so the send is funded with `x/staking-x/brune`, the same denom the position card and the unbond clamp read.

5 new tests plus one corrected: `assertFalse(ybRune.canTransfer)` was pinning the bug. Each behaviour test verified to fail on revert — the gate, the route, the account path, and two of the three keysign ones.

aminsato:
@coderabbitai the two findings that came in as review bodies rather than threads, so there was nowhere to reply inline. Both verified, both fixed in `f372964cf`.

**Outside-diff, `AccountsLoader.kt:71-84` — "Add a ybRUNE account path for ordinary sends" (Major).** Real, and it lands together with @rkokhatskyi's finding that the position card never showed its Transfer button. A plain send loads through `loadAddresses`, built from `vault.coins`, which the receipt is deliberately kept out of — and `aa7806109` migrates away the rows vaults picked up while it was still discoverable — so there was no account to select. `AccountsLoader` now takes the route's preselected token id: a send opened on the receipt publishes the wallet accounts with the synthesized receipt appended, hydrated from the chain like the unbond form's. Other sends are untouched, and a vault with no THORChain account keeps its wallet accounts rather than being emptied.

One correction to the finding's reasoning: it is not `isPositionOnly` that drops the receipt — that mark is set only for Solana/Kamino tokens (`AccountsRepository.positionOnlyTokens` returns early for every other chain). The receipt is injected by `defiOnlyTokens` and dropped by `spendableAccounts`' `Coins.isDefiOnly` filter. Same outcome, different filter, and it matters for anyone reading this later.

The finding also stopped one step short of the actual crash: `KeysignShareViewModel.loadTransaction` re-resolves the coin with `vault.coins.find { ... }!!`, so an ybRUNE send would have NPE'd at the share screen even with an account path in place. That now falls back to the token the send form staged — which carries the chain's own address and derived key — for `Coins.isDefiOnly` coins only. Covered by a new `KeysignShareViewModelReceiptSendTest`, including that a vault token still resolves from the vault and that an unknown token still fails loudly.

**Nitpick, `StakingTabHeaderIconTest.kt:5-6` — Kotest assertions.** Fixed; the file is now `shouldBe` / `shouldNotBe` throughout. `AccountsLoaderTest` stays on `kotlin.test` where its new cases live — that file is uniformly `assertEquals`, and mixing the two styles inside one file is worse than either; same disposition as the last round.

app/src/main/java/com/vultisig/wallet/ui/models/send/AccountsLoader.kt
```diff
@@ -245,6 +277,129 @@ internal class AccountsLoader(
...
+        // As with sRUJI: the RUNE account carries the address the unbond is sent from, funds the
+        // gas fee, and supplies the derived key the receipt's catalogue entry has no copy of.
+        val thorchainAccount =
+            if (alongsideWalletAccounts) {
+                // A plain send that merely started on the receipt. Without a THORChain account
+                // there is nothing to synthesize it from, but the rest of the vault is still
+                // perfectly sendable — leave the form the accounts it does have rather than
+                // emptying it over a token it can simply drop.
+                accountsLoaded.find { it.token.id.equals(Coins.ThorChain.RUNE.id, true) }
+                    ?: run {
```
rkokhatskyi, line 366: This fallback publishes silently with no Timber log, unlike every sibling missing-RUNE path in this file (AccountsLoader.kt:371, loadAutoCompoundRuji). If ThorChain is ever removed from the vault after bonding, the receipt quietly drops from the send picker with no diagnostic trail.

rkokhatskyi (approved):

Issue #5607, opened 2026-08-14
Kamino deposit and withdraw never check the wallet holds enough SOL for fees


rkokhatskyi:
## Summary

Neither a Kamino deposit from a token vault nor any Kamino withdraw checks that the wallet holds enough native SOL to pay for the transaction. A vault holding only USDC passes the amount form, the verify screen and the full MPC ceremony, and fails only at broadcast. Introduced by #5588.

## Details

In `app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt:229-231`, `spendableBalance()` returns early with the whole token balance for any non-native `coin`, so the native SOL balance is never read, even though the transaction pays a base fee plus the ComputeBudget priority fee this flow injects. `submit()` validates against the vault minimum and the held share count, never against SOL, and the withdraw path reserves nothing at all.

Nothing downstream closes the gap: `KaminoTransactionPreparer.kt:106-131` never simulates or reads the payer balance, and `VerifyDepositViewModel.kt:233-236` returns immediately for any chain that is not QBTC, leaving Sign enabled. The ordinary deposit form makes this check at `DepositAmountHelper.kt:229-236`; the Kamino form builds its own `DepositTransaction` and never routes through it.

## Impact

A wallet holding USDC and no SOL runs the whole ceremony, then gets the node's raw rejection at broadcast with no indication that SOL is missing. No funds are at risk — preflight means nothing lands — but the ceremony is wasted.

## Suggested remediation

- Read the native SOL balance in `KaminoAmountViewModel` and refuse Continue when it cannot cover the SOL cost, on both the deposit token-vault branch and the withdraw path, with a `KaminoAmountRefusal` entry reusing `R.string.insufficient_native_token`.
- Drop the QBTC-only gate in `VerifyDepositViewModel.checkFeeAffordability` so a shortfall also disables Sign.

PR #5646, opened 2026-08-19
fix(defi): describe a Kamino transaction the same way on both devices

### LINKED ISSUE #5644: Kamino deposit: joining device shows a different amount than initiator
## Summary

On a Kamino SOL-vault deposit the verify screen shows a different amount on the initiating device than on the joining device. The joining device also loses the deposit framing entirely, so a co-signer cannot tell what they are approving.

Observed on two emulators, both running the same build.

## Observed

| | Initiating device | Joining device |
|---|---|---|
| Title | Function overview | Send overview |
| Amount | **0.42921903 SOL** ($33.05) | **0.440591671 SOL** |
| Label | "You're depositing" | — |
| Vault | Allez SOL | not shown |
| Operation | KaminoDeposit | not shown |
| Network Fee | 0.01172764 SOL ($0.90) | 0.000105 SOL ($0.00796) |

The joining device's amount is the deposit principal plus roughly the initiator's fee:

```
0.440591671 − 0.42921903 = 0.011372641 SOL
initiator's displayed fee = 0.01172764 SOL
```

So the joining device appears to be surfacing the transaction's total native-SOL debit (principal + account rents + priority fee) where the initiator surfaces the deposit principal with those costs broken out into the fee row.

## Expected

Both devices describe the same transaction, so both should show:

- the same deposit amount — whichever of the two definitions is chosen, it must be consistent across devices
- the same network fee
- enough context on the joining device to identify the action (destination vault and operation), since the co-signer is being asked to approve a deposit into a specific vault, not a plain send

## Steps to reproduce

1. Two devices holding the same multi-device vault with SOL.
2. Device A: DeFi tab → Solana → Kamino → Allez SOL vault → Deposit → enter an amount → continue to the verify screen.
3. Device B: join the signing session and open its verify screen.
4. Compare the amount, fee and context rows on both screens.

No signature is needed — the mismatch is visible before signing.

## Notes

This is not introduced by #5620. That PR fixes a separate misread where a Kamino SOL deposit rendered as a bogus `SOL → WSOL` swap. I checked out the parser as it stood immediately before that fix and it returns the same 0.440591671 for both plausible simulation-diff shapes, so this amount predates the fix and comes from a path that PR does not change.

Possibly related on the fee side: #5606, #5607.

aminsato:
Closes #5644

## The problem

A Kamino deposit read differently on the two devices of a co-signed vault. The initiating device showed `You're depositing 0.42921903 SOL` into Allez SOL with a `0.01172764 SOL` fee. The joining device showed a **Send overview** of `0.440591671 SOL` with a `0.000105 SOL` fee, no vault and no operation — so a co-signer could not tell what they were approving, and the two screens disagreed on both figures.

Three separate causes, none of them a wrong number in one place:

| | Cause |
|---|---|
| Framing | `JoinKeysignViewModel.isDepositPayload` has no case a Kamino payload can match — the specific is `Solana`, the memo is `null` — so it fell to the send branch. `operation` and the vault name live only in the initiating device's local `DepositTransaction`; nothing about them travels on the wire. |
| Amount | `loadBlockaidSimulation` overwrote the hero with Blockaid's simulated **total native-SOL debit**, which for a SOL-vault deposit is the principal plus the rent for the accounts the transaction creates. The initiator shows the principal alone. |
| Fee | `computeJoinKeysignNetworkFee` has no Solana branch, so the fee was a fresh estimate of a **plain transfer** — 5,000 lamports plus a median price over 100,000 compute units — for a transaction carrying its own 320,000–400,000 unit budget in the bytes being signed. |

The fee and framing causes are not Kamino-specific — every in-app raw-Solana flow (`SolanaDelegate`, `SolanaUnstake`, `SolanaMoveStake`, `SolanaFinishMove`) hits the same two paths. This PR fixes them for Kamino; the staking flows are unchanged and still show a send overview.

## The fix

The joining device recognises Kamino **from the transaction itself**, not from the display fields relayed beside it:

- reads the single kVault instruction's Anchor discriminator — `deposit` and `withdraw_from_available`
- names the vault by the signer's derived **share account**, because the vault's own address travels in an address lookup table and is not among the static account keys
- refuses the pairing unless the payload's destination, and a deposit's amount, match what that instruction carries
- runs the same `KaminoTransactionValidator` the initiating device ran before relaying — **which the joining device never did**: it was signing these bytes unchecked

Anything unrecognised falls back to today's screen rather than blocking a ceremony over a shape the reader does not know.

The fee and the rent move into one shared derivation (`kaminoNetworkFeeLamports`, `KaminoDepositRentReserve`) that **both** devices call. Two screens agreeing on a number is a property of deriving it once, not of two call sites happening to match today.

## Judgement calls for the reviewer

1. **Rent stays inside the displayed network fee.** The initiating device folds it in deliberately, so `amount + fee` reconciles against the wallet balance, and #5607 will want it for the affordability check. Both devices now agree on that definition rather than on a changed one. Splitting rent into its own row is the more honest UI and also an iOS-parity decision — happy to file it as a follow-up if you'd rather.
2. **Direction comes from the discriminator, the vault from the share account.** The alternative was trusting the relayed `toAddress`, which is display material. This way the pairing is checked rather than assumed, at the cost of two pinned discriminator constants (unit-tested against their `sha256("global:…")` derivation, and against live fixture bytes on device).
3. **No Blockaid badge on a recognised Kamino transaction.** Parity: the initiating device's `VerifyDepositViewModel` has never scanned one, and the app's own validator now runs on both sides.

## Test plan

25 new tests. 5 verified to fail when the fix is reverted.

- `KaminoRelayedTransactionTest` (14) — discriminators pinned against their Anchor derivation; deposit/withdraw read; refusals for no kVault instruction, two of them, an unknown discriminator, an unnamed vault, a truncated amount, a multi-transaction payload, undecodable bytes, and a transaction missing the attribution memo; the fee arithmetic
- `KaminoJoinIntentTest` (5) — a payload whose amount or destination disagrees with its own bytes loses the deposit framing; a withdraw's shares are never compared against tokens
- `JoinDepositKaminoTest` (3) — the co-signer's `DepositTransaction` carries the operation, the vault name, the principal, the bytes, and the relayed-budget fee; the transfer estimate is never consulted
- `KaminoRelayedTransactionAndroidTest` (3, emulator) — the same recognition against the **live** `KaminoFixtures` bytes, including that another wallet cannot name the vault. This is what proves the discriminators and the share-account derivation are what those transactions actually carry

Full `:app` and `:data` unit suites, `:app:lintDebug`, `:data:lintDebug` and ktfmt all green.

**Manual:** two devices on one multi-device vault → DeFi → Kamino Earn → Allez SOL → Deposit → sign on A → scan on B. B should read `Function overview` / `You're depositing` with A's amount, `Vault: Allez SOL`, and a network fee string identical to A's. A plain SOL send and a Solana staking delegate must both still join as a Send overview.

## Summary by CodeRabbit

* **New Features**
  * Added support for recognizing and signing relayed Kamino Solana deposits and withdrawals.
  * Added validation for vault, destination, asset, amount, signature, and transaction fee details.
  * Added Kamino-specific operation details, fee calculation, deposit rent handling, and compute-budget validation.
  * Withdrawal verification now displays unverified vault-share amounts when the token amount cannot be confirmed.

* **Bug Fixes**
  * Improved rejection of malformed, ambiguous, or incorrectly targeted transactions.

* **Tests**
  * Added comprehensive unit and integration coverage for Kamino transaction flows.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinDepositUiModelBuilder.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinDepositUiModelBuilder.kt
@@ -2,8 +2,15 @@ package com.vultisig.wallet.ui.models.keysign
 
 import com.vultisig.wallet.data.blockchain.model.Transfer
 import com.vultisig.wallet.data.blockchain.model.VaultData
+import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoAction
+import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoDepositRentReserve
+import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoRelayedIntent
+import com.vultisig.wallet.data.blockchain.solana.kamino.kaminoNetworkFeeLamports
+import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.models.DepositTransaction
 import com.vultisig.wallet.data.models.GasFeeParams
+import com.vultisig.wallet.data.models.OPERATION_KAMINO_DEPOSIT
+import com.vultisig.wallet.data.models.OPERATION_KAMINO_WITHDRAW
 import com.vultisig.wallet.data.models.TokenValue
 import com.vultisig.wallet.data.models.getPubKeyByChain
 import com.vultisig.wallet.data.models.payload.BlockChainSpecific
@@ -40,13 +47,23 @@ constructor(
     private val addressBookRepository: AddressBookRepository,
     private val chainAccountAddressRepository: ChainAccountAddressRepository,
     private val feeResolver: JoinKeysignFeeResolver,
+    private val depositRentReserve: KaminoDepositRentReserve,
 ) {
 
     /**
      * Builds the deposit [JoinKeysignVerifyResult] from [payload] for the vault identified by
      * [vaultId], resolving fees and parsing the THOR memo.
+     *
+     * @param kamino what the relayed Solana bytes turned out to be, when they were recognised as a
+     *   Kamino Earn deposit or withdraw. It carries everything a THOR memo carries for the chains
+     *   above — which vault, which direction — none of which a Kamino payload states in a field
+     *   (issue #5644).
      */
-    suspend fun build(payload: KeysignPayload, vaultId: String): JoinKeysignVerifyResult {
+    suspend fun build(
+        payload: KeysignPayload,
+        vaultId: String,
+        kamino: KaminoRelayedIntent? = null,
+    ): JoinKeysignVerifyResult {
         when (payload.blockChainSpecific) {
             is BlockChainSpecific.MayaChain,
             is BlockChainSpecific.THORChain,
@@ -56,6 +73,14 @@ constructor(
             // to render its verify screen so multi-device staking ceremonies aren't blocked.
             is BlockChainSpecific.Ton,
             is BlockChainSpecific.UTXO -> Unit
+
+            // Solana arrives here only as a recognised Kamino transaction. A raw Solana payload
+            // nobody could read is still a send, and rendering it as a deposit would put a
+            // deposit's framing around bytes this device cannot describe.
+            is BlockChainSpecific.Solana ->
+                requireNotNull(kamino) {
+                    "a Solana deposit must be a recognised Kamino transaction"
+                }
 
             else -> error("BlockChainSpecific ${payload.blockChainSpecific} is not supported")
         }
@@ -89,12 +114,13 @@ constructor(
 
         val nativeCoin = withContext(Dispatchers.IO) { tokenRepository.getNativeToken(chain.id) }
         val estimatedTokenFees =
-            feeResolver.resolveJoinKeysignNetworkFee(
-                payload = payload,
-                chain = chain,
-                nativeCoin = nativeCoin,
-                blockchainTransaction = blockchainTransaction,
-            )
+            kamino?.let { kaminoNetworkFee(it, payload, nativeCoin) }
+                ?: feeResolver.resolveJoinKeysignNetworkFee(
+                    payload = payload,
+                    chain = chain,
+                    nativeCoin = nativeCoin,
+                    blockchainTransaction = blockchainTransaction,
+                )
 
         val totalGasAndFee =
             gasFeeToEstimatedFee(
@@ -119,11 +145,17 @@ constructor(
                 estimatedFees = estimatedTokenFees,
                 estimateFeesFiat = totalGasAndFee.formattedFiatValue,
                 blockChainSpecific = payload.blockChainSpecific,
-                operation = parsedThorMemo?.operation.orEmpty(),
+                operation = kamino?.let(::kaminoOperation) ?: parsedThorMemo?.operation.orEmpty(),
                 nodeAddress = parsedThorMemo?.nodeAddress.orEmpty(),
                 pairedAddress = parsedThorMemo?.pairedAddress.orEmpty(),
                 thorAddress = parsedThorMemo?.thorAddress.orEmpty(),
                 pool = parsedThorMemo?.pool.orEmpty(),
+                // The verify screen labels this row as the vault for a Kamino operation, so the
+                // co-signer reads the destination by name rather than as a bare program address.
+                validatorName = kamino?.vault?.fallbackName,
+                // Renders the instruction breakdown the initiating device shows, which for a
+                // pre-built transaction is the only place its real work is visible.
+                signSolana = payload.signSolana,
             )
         // Resolve the same From/To labels the initiator renders (issue #5301), so the joining
         // co-signer sees "VaultName (address)" instead of raw thor1… addresses. Mirrors
@@ -157,6 +189,48 @@ constructor(
             verifyUiModel = VerifyUiModel.Deposit(VerifyDepositUiModel(depositTransactionUiModel)),
             transactionTypeUiModel = TransactionTypeUiModel.Deposit(depositTransactionUiModel),
             transactionHistoryData = mapDepositTransactionHistoryData(depositTransactionUiModel),
+        )
+    }
+
+    private fun kaminoOperation(kamino: KaminoRelayedIntent): String =
+        when (kamino.action) {
+            KaminoAction.DEPOSIT -> OPERATION_KAMINO_DEPOSIT
+            KaminoAction.WITHDRAW -> OPERATION_KAMINO_WITHDRAW
+        }
+
+    /**
+     * The fee the initiating device quotes for these bytes, derived here from the relayed compute
+     * budget rather than re-estimated.
+     *
+     * A fee service can only price the transfer it is handed, and this is not one: it is a
+     * pre-built transaction carrying its own compute-unit limit — 320,000 to 400,000 against a
+     * transfer's 100,000 — plus, on a first native-SOL deposit, the rent for the accounts it
+     * creates. Estimating it as a transfer is how the two devices came to quote fees a hundredfold
+     * apart for one transaction (issue #5644).
+     */
+    private suspend fun kaminoNetworkFee(
+        kamino: KaminoRelayedIntent,
+        payload: KeysignPayload,
+        nativeCoin: Coin,
+    ): TokenValue {
+        val specific = payload.blockChainSpecific as BlockChainSpecific.Solana
+        val rentReserve =
+            if (kamino.action == KaminoAction.DEPOSIT) {
+                withContext(Dispatchers.IO) {
+                    depositRentReserve(kamino.vault, payload.coin.address)
+                }
+            } else {
+                BigInteger.ZERO
+            }
+        return TokenValue(
+            value =
+                kaminoNetworkFeeLamports(
+                    vault = kamino.vault,
+                    action = kamino.action,
+                    relayedUnitPrice = specific.priorityFee,
+                    rentReserve = rentReserve,
+                ),
+            token = nativeCoin,
         )
     }
 }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinDepositUiModelBuilder.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinDepositUiModelBuilder.kt
@@ -24,6 +24,7 @@ import com.vultisig.wallet.data.usecases.ThorchainMemoParser
 import com.vultisig.wallet.ui.models.deposit.VerifyDepositUiModel
 import com.vultisig.wallet.ui.models.mappers.DepositTransactionHistoryDataMapper
 import com.vultisig.wallet.ui.models.mappers.DepositTransactionToUiModelMapper
+import com.vultisig.wallet.ui.models.mappers.TokenValueToDecimalUiStringMapper
 import com.vultisig.wallet.ui.utils.resolveDstVaultName
 import java.math.BigInteger
 import java.util.UUID
@@ -48,6 +49,7 @@ constructor(
     private val chainAccountAddressRepository: ChainAccountAddressRepository,
     private val feeResolver: JoinKeysignFeeResolver,
     private val depositRentReserve: KaminoDepositRentReserve,
+    private val mapTokenValueToDecimalUiString: TokenValueToDecimalUiStringMapper,
 ) {
 
     /**
@@ -184,6 +186,7 @@ constructor(
                     srcVaultName = srcVaultName,
                     dstVaultName = dstVaultName,
                     dstAddressBookTitle = dstAddressBookTitle,
+                    unverifiedWithdrawShares = kamino?.let(::withdrawShares),
                 )
         return JoinKeysignVerifyResult(
             verifyUiModel = VerifyUiModel.Deposit(VerifyDepositUiModel(depositTransactionUiModel)),
@@ -191,6 +194,22 @@ constructor(
             transactionHistoryData = mapDepositTransactionHistoryData(depositTransactionUiModel),
         )
     }
+
+    /**
+     * The share figure a withdraw's kVault instruction carries, formatted, or null for a deposit —
+     * whose headline amount this device already checked against the same instruction.
+     *
+     * Sized by the vault's own share scale, which is not the token's: Allez SOL's shares carry six
+     * decimals against the token's nine.
+     */
+    private fun withdrawShares(kamino: KaminoRelayedIntent): String? =
+        kamino
+            .takeIf { it.action == KaminoAction.WITHDRAW }
+            ?.let {
+                mapTokenValueToDecimalUiString(
+                    TokenValue(value = it.amount, unit = "", decimals = it.vault.sharesDecimals)
+                )
+            }
 
     private fun kaminoOperation(kamino: KaminoRelayedIntent): String =
         when (kamino.action) {
@@ -213,7 +232,6 @@ constructor(
         payload: KeysignPayload,
         nativeCoin: Coin,
     ): TokenValue {
-        val specific = payload.blockChainSpecific as BlockChainSpecific.Solana
         val rentReserve =
             if (kamino.action == KaminoAction.DEPOSIT) {
                 withContext(Dispatchers.IO) {
@@ -227,7 +245,10 @@ constructor(
                 kaminoNetworkFeeLamports(
                     vault = kamino.vault,
                     action = kamino.action,
-                    relayedUnitPrice = specific.priorityFee,
+                    // From the instructions, not from the field relayed beside them. The two are
+                    // pinned equal by KaminoRelayedIntent.takeIfDescribedBy before this runs, and
+                    // the one inside the bytes is the one the runtime charges.
+                    relayedUnitPrice = kamino.priorityFee?.price,
                     rentReserve = rentReserve,
                 ),
             token = nativeCoin,
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KaminoJoinIntent.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KaminoJoinIntent.kt
@@ -1,0 +1,34 @@
+package com.vultisig.wallet.ui.models.keysign
+
+import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoAction
+import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoRelayedIntent
+import com.vultisig.wallet.data.blockchain.solana.kamino.kaminoDestinationAddress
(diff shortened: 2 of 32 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionCost.kt
```diff
@@ -0,0 +1,116 @@
...
+    vault: KaminoVault,
+    action: KaminoAction,
+    relayedUnitPrice: BigInteger?,
+    rentReserve: BigInteger = BigInteger.ZERO,
+): BigInteger =
+    SolanaHelper.DefaultFeeInLamports +
+        KaminoComputeBudget.priorityFeeLamports(
+            vault = vault,
+            action = action,
+            networkPrice = relayedUnitPrice,
```
rkokhatskyi, line 54: The joining device prices its fee row from the relayed `priorityFee` field (`JoinDepositUiModelBuilder.kt:230`) while ComputeBudget arguments are never checked, only allow-listed (`KaminoTransactionValidator.kt:52`), so a peer can show a clamped low fee while the signed bytes charge far more. iOS's priorityFeeAgrees refuses exactly this

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinDepositUiModelBuilder.kt
```diff
@@ -57,6 +74,14 @@ constructor(
             is BlockChainSpecific.Ton,
             is BlockChainSpecific.UTXO -> Unit
 
+            // Solana arrives here only as a recognised Kamino transaction. A raw Solana payload
+            // nobody could read is still a send, and rendering it as a deposit would put a
+            // deposit's framing around bytes this device cannot describe.
+            is BlockChainSpecific.Solana ->
+                requireNotNull(kamino) {
```
rkokhatskyi, line 83: takeIfDescribedBy skips the amount check for WITHDRAW (`KaminoJoinIntent.kt:29`), so JoinDepositUiModelBuilder.kt:91 renders the adversary-chosen payload.toAmount as the headline under a validated vault name, with no Blockaid scan running to flag it Unverified. iOS surfaces .amountUnverifiable plus the real decoded share count instead

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KaminoJoinIntent.kt
```diff
@@ -0,0 +1,34 @@
...
+ * be signed. Only the bytes say what the transaction does, so the two are compared before the
+ * payload is given a Kamino deposit's framing: a payload naming a destination the transaction does
+ * not pay, or a deposit amount the kVault instruction does not carry, keeps the generic screen
+ * rather than being dressed up as something this device could not confirm (issue #5644).
+ *
+ * Withdraw amounts are deliberately not compared — the instruction carries the share count where
+ * the payload carries the tokens those shares are worth, and the rate between them is not something
+ * a joining device can pin down.
+ */
+internal fun KaminoRelayedIntent.takeIfDescribedBy(payload: KeysignPayload): KaminoRelayedIntent? {
```
rkokhatskyi, line 32: This compares toAddress/toAmount but never the coin, and JoinDepositUiModelBuilder.kt:91 scales that raw amount with the wire's coin.decimal and ticker, so a relayed payload claiming decimal=6 for a 9-decimal Allez SOL deposit renders the headline a thousandfold low. Pin kamino.vault.coin instead, as iOS's asset check does

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoRelayedTransaction.kt
```diff
@@ -0,0 +1,169 @@
...
+     *   a batch of any other size is somebody else's payload.
+     * @param signerAddress the vault's own Solana address — the same address on both devices, which
+     *   is what makes the initiating device's derivations reproducible here.
+     */
+    operator fun invoke(
+        rawTransactions: List<String>,
+        signerAddress: String,
+    ): KaminoRelayedIntent? {
+        val rawTransaction = rawTransactions.singleOrNull() ?: return null
+        val decoded = runCatching { decode(rawTransaction) }.getOrNull() ?: return null
```
rkokhatskyi, line 147: KaminoDecodedTransaction (`KaminoTransactionPreparer.kt:20`) drops the message header, so nothing checks numRequiredSignatures before the join screen labels these bytes a verified Kamino deposit, and the raw-signing splice only fills slot 0, leaving a second signer's slot untouched. iOS gates the same decode with validateUnsignedSingleSigner

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionCost.kt
aminsato: Right, and it was worse than a display disagreement: the row clamps into `MAX_UNIT_PRICE` while the instruction carries a bare `u64`, so the ceiling was on the *figure*, not on the charge.

Fixed in `524425f7e`, on both devices rather than only the joining one.

`KaminoComputeBudget` now reads the budget back out of the instructions (`readFrom`, plus `unitLimitArgument`/`unitPriceArgument`). `KaminoTransactionValidator.rejectAForeignComputeBudget` refuses anything but the pair this app builds: the two instructions at indices 0 and 1, `limit == unitLimitFor(vault, action)`, and a price already inside `[FALLBACK_UNIT_PRICE, MAX_UNIT_PRICE]` (expressed as `price == unitPriceFor(price)`, since the clamp is idempotent). `KaminoRelayedIntent` carries what it read, and `takeIfDescribedBy` compares that against the payload's `(priorityFee, priorityLimit)` — nil-vs-nil agrees, one-sided does not, which is `priorityFeeAgrees`'s rule.

Two consequences worth calling out:

* `JoinDepositUiModelBuilder` now prices the row from `kamino.priorityFee?.price` rather than `specific.priorityFee`. Same number once the two are pinned equal, but sourced from the thing that gets signed.
* Because the rule lives in the shared validator, the initiating device fails closed on the same shapes. That is why the validator fixtures changed — they carried a bare placeholder `ComputeBudget` instruction mid-list, which is not what the preparer emits.

A behaviour note: an initiator that relays no `computeLimit` now falls through to the generic send screen instead of the Kamino one. iOS records both fields so it does not arise in practice, and falling back is the documented contract for anything unrecognised — but it is a change, and I would rather you knew than found it.

8 new validator tests, all verified to fail on revert, plus one androidTest that pins the read against WalletCore's own `setComputeUnitLimit` encoding rather than one the tests wrote.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinDepositUiModelBuilder.kt
aminsato: Confirmed, and the missing Blockaid scan makes it worse rather than better — the one affordance that might have qualified the number is exactly the one this PR removed for a recognised Kamino transaction.

Fixed in `524425f7e`, taking the `.amountUnverifiable` shape rather than a refusal, for the reason your own `KaminoVerifyPresentation` gives: refusing would break every legitimate peer withdraw.

`DepositTransactionUiModel` gains `unverifiedWithdrawShares`, set only by `JoinDepositUiModelBuilder` and only for a withdraw — the initiating device knows the share count it asked for, so it has nothing to disclaim. Under the hero the screen now reads "This device can't confirm this amount. The transaction withdraws 0.5 vault shares."

The share figure is scaled by `vault.sharesDecimals`, not the token's: Allez SOL is 6 against 9, so reusing the coin's scale would have printed a wrong number in the sentence that exists to give a right one.

`kamino_verify_withdraw_unverified` added to all 10 locales, each using the term that file already uses for a vault (most keep it untranslated; ko `볼트`, zh `金库`).

Two tests: the withdraw discloses `"0.5"` from a 500,000-share instruction (fails on revert), and a deposit discloses nothing, since its amount *was* checked against the same instruction.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KaminoJoinIntent.kt
aminsato: Real, and the reason the amount check does not catch it is the part worth writing down: base units are the same number whatever scale is claimed for them, so `amount == payload.toAmount` passes and the *scale applied to them* is what moves.

Fixed in `524425f7e`. `takeIfDescribedBy` now runs `describesAsset`, comparing `payload.coin` against `KaminoVault.coin` on decimals, contract address and ticker (case-insensitive) — the same three your asset arm checks, and skipped the same way when no wallet coin maps to the mint, since an unestablished fact is not a contradiction.

One correction to the example, only so we do not chase the wrong sign later: `decimal=6` on a 9-decimal Allez SOL deposit renders `1e9 / 1e6` = a thousandfold **high**, not low. `decimal=12` is the thousandfold-low case. Wrong in both directions, so the finding stands either way.

Two tests, both failing on revert: a coin with the wrong scale, and one carrying the right mint under a false ticker — the hero reads "You're depositing &lt;amount&gt; &lt;ticker&gt;", so a true figure beside the wrong asset name is still a false claim.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoRelayedTransaction.kt
aminsato: Confirmed on both halves — the header was dropped, and `SolanaHelper.signRawTransaction` splices slot 0 and leaves the rest exactly as received, which is right for a dApp co-sign and wrong for a transaction this screen has just called a verified Kamino deposit.

Fixed in `524425f7e` as `validateUnsignedSingleSigner`'s three assertions: one required signature, every declared slot still an all-zero placeholder, and the fee payer already checked by `rejectAnotherAccountsAuthority`.

The one design choice I would flag: rather than parsing the envelope a second time inside the Kamino path, I extracted `SolanaSignatureEnvelope` into `:data/blockchain/solana/` and pointed `SolanaHelper.parseRawTransaction` at it. The check now reads the same parse the splice writes into. A parallel reader would have been smaller, and it could have agreed with itself while describing different bytes than the ones being signed — which is the class of bug this whole PR is about.

`KaminoDecodedTransaction` carries `requiredSignatures` and `isUnsigned` with **no defaults**. A default would have to be the permissive value, and "nothing was checking this" is precisely what you found.

Two unit tests fail on revert, plus a `SolanaSignatureEnvelopeTest` covering the parse itself: a filled slot, a second *empty* slot (emptiness is not the question — the splice still leaves it unsigned), zero declared signatures, a truncated envelope, and a multi-byte shortvec count that a single-byte reader would call 0. The 30 Solana androidTests confirm the real prepared bytes satisfy the gate.

aminsato:
## CodeRabbit nitpicks

They live in the review body rather than in threads, so the dispositions are collected here. All seven verified against the branch first; five acted on, two declined. Pushed in `524425f7e`.

**Taken**

* **`KaminoTransactionCost.kt:69-84` — sequential reads.** Correct, and it sits in front of a verify screen that cannot quote a fee until it answers, on both devices. `KaminoDepositRentReserve.invoke` now runs the rent-exemption read, the two token-account lookups and the positions call inside `coroutineScope` with `async`.
* **`KaminoTransactionCost.kt:86-116` — silent fallbacks.** Correct: a failed positions read re-charges the farm rent and changes the quoted fee with no trace. `Timber.w` on all three. `runCatchingCancellable` already rethrows `CancellationException`, so the logs are real failures only.
* **`JoinKeysignViewModel.kt:613-634` — eager resolve.** Correct — a decode plus one ATA derivation per allow-listed vault, discarded on the next line for any swap payload. Now `if (swapPayload == null) resolveKaminoIntent(payload) else null`.
* **`JoinDepositUiModelBuilder.kt:211-224` / `KaminoAmountViewModel.kt:486` — unchecked casts.** Half right, and the reachable half is the joining device: recognition gates on `payload.coin.chain`, which does not settle the shape of `blockChainSpecific`, so a `chain=Solana` coin under an `Ethereum` specific passed the `when` and reached the cast. The builder cast is gone entirely — `takeIfDescribedBy` now refuses a non-Solana specific up front, which it has to do anyway to compare the compute budget. `KaminoAmountViewModel` gets a named `checkNotNull`; the payload there is locally built, so that one is cosmetic.
* **`JoinDepositKaminoTest.kt:99-112` — no test for the `requireNotNull(kamino)` refusal.** A real gap. Added: a Solana payload with no recognised intent is refused rather than rendered as a deposit.

**Declined**

* **androidTest `runBlocking` → `runTest`.** The cited guideline is about ViewModels and blocking the main thread; six androidTest files in this repo use `runBlocking`, which is the ordinary shape for an instrumentation test that is already off the main thread. Changing one file would make it the odd one out.
* **Deriving 350,000 / 400,000 from `KaminoComputeBudget.unitLimitFor` in the fee tests.** These assert the *value*, and reading it through the code under test makes the assertion true by construction — the limits have a direct on-chain consequence (too low and the transaction aborts on compute exhaustion), so pinning them is the point. Where a test does assert a relationship, it derives: the new `the intent carries the budget from the instructions` case reads the limit through `unitLimitFor`, because there the claim is "these agree", not "this is 350,000".

app/src/main/res/values-ru/strings.xml
```diff
@@ -543,6 +543,7 @@
     <string name="verify_deposit_depositing">Вы вносите</string>
     <string name="verify_deposit_withdrawing">Вы выводите</string>
     <string name="kamino_verify_vault">Vault Kamino</string>
+    <string name="kamino_verify_withdraw_unverified">Это устройство не может подтвердить сумму. Транзакция выводит долей vault‑а: %1$s.</string>
```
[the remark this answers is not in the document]
aminsato: Not taking the suggested diff — it would introduce the bug it is trying to fix — but the wording was awkward and is now changed, in `591c73d56`.

Russian selects the noun's form from the number in front of it: `1 доля`, `2 доли`, `5 долей`, and — because a fractional quantity takes the genitive **singular** — `0,5 доли`. `%1$s` carries all of those. A share figure is fractional most of the time (the test case is 0.5), so `выводит %1$s долей` would be wrong for the commonest value and right only for 5 and up.

That is why the string was built around a colon in the first place: the placeholder governs nothing, so nothing has to agree with it. The search result quoted above says the same thing from the other side — reordering is available for focus, and the noun stays genitive plural because a quantifier is what would otherwise change it.

The old phrasing did read badly, though. `Транзакция выводит долей vault‑а: %1$s` puts a bare genitive plural after a transitive verb, which is not how anyone would say it. Now:

```
Количество выводимых долей vault‑а: %1$s.
```

A label and a value, which is the ordinary way to render a count in Russian UI and works for every value the placeholder can hold.

Worth noting for the other nine: the `%1$s` + plural noun pattern stands there. German, Spanish, Croatian, Italian, Dutch and Portuguese all take the plural for a decimal quantity, so `0,5 Vault-Anteile` and `0,5 participaciones` are correct; Korean and Chinese do not inflect for number at all. Russian is the one language in the set where the split is by value rather than by plurality.

aminsato:
## CodeRabbit round 2

One inline finding (answered in its thread) and two nitpicks in the review body. Pushed in `591c73d56`.

**Taken — `KaminoTransactionValidator.kt:73-77`, duplicated layout constants.** Fair, and it was mine from the previous commit: I added `UNIT_LIMIT_INDEX`/`UNIT_PRICE_INDEX` to the validator without noticing `KaminoTransactionPreparer` already had its own pair. Both encode the same positional contract with iOS, so both now read them from `KaminoComputeBudget`, alongside the program id and the discriminators they belong with. The observation that a drift would fail at build time is right — `prepare` validates its own output — but that only means the duplication was cheap to get away with, not that two definitions of one contract are worth keeping.

**Declined — androidTest `runBlocking` → `runTest`.** Same answer as round 1, for the same reason: the cited guideline is about ViewModels and blocking the main thread, and six androidTest files in this repo use `runBlocking`. It is the ordinary shape for an instrumentation test, which is already off the main thread. Converting this one file would make it the outlier rather than the example.

**Inline, on the Russian string — declined as suggested, reworded instead.** Moving `%1$s` in front of `долей` would be wrong for the commonest value: Russian takes the genitive singular after a fractional quantity (`0,5 доли`), and the share figure is fractional most of the time. The full reasoning is in [the thread](https://github.com/vultisig/vultisig-android/pull/5646#discussion_r3819578107); the string is now `Количество выводимых долей vault-а: %1$s`, a label and a value, which holds for every number the placeholder can carry.

`:data` and `:app` unit suites, both lint tasks and ktfmt green.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionCost.kt
rkokhatskyi: Verified — describesPriorityFee cross-checks the wire fields against KaminoComputeBudget.readFrom's bytes-derived value, rejectAForeignComputeBudget pins position, limit and price on both devices, and the fee row now prices off kamino.priorityFee instead of the wire field. Closed

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinDepositUiModelBuilder.kt
rkokhatskyi: Verified — unverifiedWithdrawShares is scaled by vault.sharesDecimals, set only for withdraw, and renders under the hero on the join screen. Closed

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KaminoJoinIntent.kt
rkokhatskyi: Verified — describesAsset checks decimal, contractAddress and ticker against vault.coin. Thanks for the correction on the example direction, the finding stands regardless. Closed

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoRelayedTransaction.kt
rkokhatskyi: Verified — rejectAnythingButOneEmptySignatureSlot checks requiredSignatures == 1 and isUnsigned, sourced from the same SolanaSignatureEnvelope the signing splice reads. Good call sharing the parse instead of duplicating it. Closed

rkokhatskyi (approved):

PR #5647, opened 2026-08-19
fix(defi): make the Kamino instruction guards allow-lists

### LINKED ISSUE #5603: Kamino validator guards are deny-lists, not allow-lists
## Summary

`KaminoTransactionValidator`'s SPL Token and System Program guards are deny-lists: each refuses a couple of hand-picked opcodes and lets every other instruction of the same program through unchecked. Introduced by #5588.

## Details

In `data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionValidator.kt:206-239`, both System checks sit inside `if (discriminator == SYSTEM_TRANSFER)`, so every other `SystemInstruction` falls through while the System program stays allow-listed at `:48`. `Assign` (1) on message key 0 reassigns the wallet's system account to another program, and `CreateAccountWithSeed` (3) with `base == from == the wallet` moves an attacker-chosen lamport amount into an attacker-owned account. Both need only the fee-payer signature the message already carries.

`:190-204` rejects only when the first data byte is in `TOKEN_TRANSFER_DISCRIMINATORS = setOf(3, 12)`. There is no `else` and `instruction.accounts` is never read, so `Approve`, `SetAuthority`, `CloseAccount` and Token-2022 extensions pass. A withdraw on the Allez SOL vault legitimately contains a top-level `CloseAccount` (the wSOL unwrap), so a response need only change its `accounts[1]` to redirect the whole withdrawn amount.

## Impact

A missing control, not an observed exploit: it requires a compromised `api.kamino.finance` or a TLS-terminating proxy. Unlike the Kamino-supplied price and vault state, these instructions reach the wallet's whole balance in that token.

## Suggested remediation

- System: `if (discriminator != SYSTEM_TRANSFER) reject`, then the existing vault and destination checks unindented.
- Token: refuse any discriminator outside a known-good set, validated against a real SOL-vault deposit fixture as well as the USDC one.

aminsato:
Closes #5603

## The problem

`KaminoTransactionValidator`'s SPL Token and System guards were deny-lists. Each refused a couple of hand-picked opcodes and let every other instruction of the same program through, while both programs stayed allow-listed:

```kotlin
// Token: no else, and instruction.accounts never read
if (discriminator in TOKEN_TRANSFER_DISCRIMINATORS) reject(…)

// System: every check nested inside the transfer branch
if (discriminator == SYSTEM_TRANSFER) { …vault and destination checks… }
```

A deny-list is no control against a compromised response, because the same response that adds the instruction also picks its opcode. And the opcodes that reach furthest are not transfers: `Assign` (1) on the signer's own system account reassigns it to another program, `CreateAccountWithSeed` (3) funds an account the response chose, `SetAuthority` (6) hands over a token account, `Approve` (4) delegates it. Each needs only the fee-payer signature the message already carries; none moves a lamport itself.

The Allez SOL withdraw makes it concrete — it legitimately carries a top-level `CloseAccount` to unwrap, so rewriting only that instruction's second account redirects the entire withdrawn amount while every other instruction still reads as the withdraw the user asked for.

Missing control rather than an observed exploit: it takes a compromised `api.kamino.finance` or a TLS-terminating proxy.

## What changed

Both guards are now allow-lists.

**System** — `Transfer` (2) only, keeping the existing wrapped-SOL-vault gate and the destination check against the locally derived wSOL account. Every other `SystemInstruction` is refused by name.

**Token / Token-2022** — `SyncNative` (17) and `CloseAccount` (9) only. Everything else refused, Token-2022 extensions included.

The two are checked differently on purpose:

- **`SyncNative` is pinned by subject** — the wrapped-SOL account derived locally from the signer, and only on the wrapped-SOL vault. Its deposit was captured, so the shape is known exactly.
- **`CloseAccount` is pinned by recipient** — both the payout destination and the closing authority must be the signer, while the account being closed is left free.

That last one is the deliberate deviation from the issue's suggested remediation, and the reason is coverage: only the *unstaked* withdraw could be captured, and a staked exit may close accounts no fixture shows. Pinning the subject would refuse a real staked withdraw and buy nothing — the wallet can only close accounts it owns, and every released lamport lands on the wallet either way. Pinning the recipient still refuses the redirect the check exists for.

**Lookup-table accounts fail comparisons rather than skipping them.** These are v0 messages where most accounts resolve through a lookup table and `KaminoTransactionDecoder` names them unresolved. A check that treats "cannot tell" as "not a mismatch" is the same hole in a different place, so every account comparison here is exact.

## Evidence

The suite recorded the wrapped-SOL shape as impossible to observe without depositing real funds. It isn't: Kamino's build endpoints construct a transaction for a wallet holding nothing — the same property that produced the existing `WITHDRAW` fixture — so both Allez SOL shapes were captured directly and are now fixtures:

```
ALLEZ_SOL_DEPOSIT   AssociatedToken → System::Transfer(wallet → wSOL ATA)
                  → Token::SyncNative(wSOL ATA) → AssociatedToken
                  → kVault::deposit → Farms → Farms

ALLEZ_SOL_WITHDRAW  AssociatedToken → kVault::withdraw
                  → Token::CloseAccount(wSOL ATA, wallet, wallet)
```

Neither USDC fixture carries a top-level System or Token instruction at all — that vault moves its tokens entirely through the kVault program's own CPI — which is why the allow-lists had to be sized against these and not against the shapes already in the repo.

The same capture also retires the note at `KaminoTransactionPreparerTest.kt:132`, where the SOL-vault compute-budget test had to assert against a USDC fixture for want of a SOL one.

A confirmed mainnet Allez SOL deposit was decoded independently and checked rule by rule against the new guards: its `System::Transfer` destination and its `SyncNative` account both equal the locally derived wSOL ATA, its memo is the sole final instruction, and its kVault amount is not the sentinel. It passes unchanged.

## Test plan

- `./gradlew :data:testDebugUnitTest` — **2999 tests green**
- `./gradlew :data:compileDebugAndroidTestKotlin :app:compileDebugKotlin` — green
- 11 new unit tests (33 in the validator, up from 23). **All 10 behaviour tests verified to fail on revert** — the old validator was restored and the suite re-run to confirm. The two allow-path tests pass under both, which is the point: they guard against over-tightening, which is the real risk in this change.
- 2 new instrumented tests pin the allow-lists against the real captures. The withdraw one asserts that validation reaches the *sentinel* refusal, which runs after the value-movement rules — so getting there is proof the unwrap passed them rather than that nothing looked.

Manual, on device:

1. DeFi → Solana → Manage Positions → enable **Allez SOL** → Deposit a small amount → Sign. Expect it to sign and broadcast as before; a refusal naming "SPL token instruction 17" or "System instruction 2" would mean the allow-list is too tight.
2. Withdraw part of it → Sign. This is the `CloseAccount` path.
3. Repeat with **Max** — the full-exit shape is the one variant that cannot be captured.
4. **Steakhouse USDC** deposit and withdraw, unchanged, as the plain-token regression.

## Summary by CodeRabbit

- **Bug Fixes**
  - Strengthened Kamino Solana transaction validation for deposits and withdrawals, including wrapped SOL, vaults, authorities, derived accounts, and farm state.
  - Rejected malformed, unsupported, duplicated, unauthorized, or incomplete instructions.
  - Improved amount handling for token and share actions, including fractional, invalid, and non-positive values.

- **Tests**
  - Added realistic coverage for deposits, withdrawals, synchronization, wrapping, unwrapping, and account closing.
  - Expanded checks for instruction allow-lists, ownership, account derivation, lookup tables, and mismatched transaction details.

### DIFF
--- a/data/src/androidTest/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparerTest.kt
+++ b/data/src/androidTest/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparerTest.kt
@@ -36,6 +36,7 @@ class KaminoTransactionPreparerTest {
         vault: KaminoVault = KaminoVaultRegistry.STEAKHOUSE_USDC,
         action: KaminoAction = KaminoAction.DEPOSIT,
         networkUnitPrice: BigInteger? = null,
+        amount: String = AMOUNT,
     ): String = runBlocking {
         val resolved =
             api
@@ -48,10 +49,47 @@ class KaminoTransactionPreparerTest {
                 vault = vault,
                 action = action,
                 walletAddress = KaminoFixtures.WALLET,
-                amount = "1",
+                amount = amount,
                 networkUnitPrice = networkUnitPrice,
             )
     }
+
+    /** The wallet's own associated token account for [mint], as the preparer derives it. */
+    private fun ownAccount(mint: String): String =
+        SolanaAddress(KaminoFixtures.WALLET).defaultTokenAddress(mint)
+
+    /** Validates captured bytes the way the preparer does, with every account derived locally. */
+    private fun validate(
+        transaction: String,
+        vault: KaminoVault,
+        action: KaminoAction,
+        signerAddress: String = KaminoFixtures.WALLET,
+        amountBaseUnits: BigInteger? = BigInteger.ONE,
+    ) =
+        validateDecoded(
+            decoded = KaminoTransactionDecoder.decode(KaminoAttributionMemo.append(transaction)),
+            vault = vault,
+            action = action,
+            signerAddress = signerAddress,
+            amountBaseUnits = amountBaseUnits,
+        )
+
+    private fun validateDecoded(
+        decoded: KaminoDecodedTransaction,
+        vault: KaminoVault,
+        action: KaminoAction = KaminoAction.DEPOSIT,
+        signerAddress: String = KaminoFixtures.WALLET,
+        amountBaseUnits: BigInteger? = BigInteger.valueOf(1_000_000),
+    ) =
+        KaminoTransactionValidator.validate(
+            decoded = decoded,
+            vault = vault,
+            action = action,
+            signerAddress = signerAddress,
+            tokenAccount = ownAccount(vault.tokenMint),
+            shareAccount = ownAccount(vault.sharesMint),
+            amountBaseUnits = amountBaseUnits,
+        )
 
     @Test
     fun a_prepared_deposit_validates_and_its_memo_is_the_final_instruction() {
@@ -131,15 +169,19 @@ class KaminoTransactionPreparerTest {
 
     @Test
     fun a_SOL_vault_deposit_gets_the_larger_budget_it_needs_to_wrap_first() {
-        // Note the fixture: this is the captured **USDC** deposit prepared against the SOL vault,
-        // because no ALLEZ_SOL response has been captured — doing so needs a wallet depositing real
-        // SOL. So what this pins is that the vault selects the larger limit and that the limit
-        // reaches the wire; the wrap-and-sync instruction shape a real SOL deposit carries is not
-        // exercised here. Inventing bytes for it would assert against a transaction Kamino never
-        // sent. The 350,000 value itself rests on the iOS mainnet measurement of 287,029.
-        val prepared = prepare(vault = KaminoVaultRegistry.ALLEZ_SOL)
+        // Against the captured SOL deposit, so the limit is pinned to the wrap-and-sync shape it is
+        // sized for rather than to a USDC transaction standing in for it. The 350,000 value itself
+        // rests on the iOS mainnet measurement of 287,029.
+        val prepared = prepareSolDeposit()
         assertEquals(350_000L, SolanaTransaction.getComputeUnitLimit(prepared)?.toLong())
     }
+
+    private fun prepareSolDeposit(amount: String = SOL_AMOUNT) =
+        prepare(
+            api = KaminoFixtureApi(KaminoFixtures.ALLEZ_SOL_DEPOSIT),
+            vault = KaminoVaultRegistry.ALLEZ_SOL,
+            amount = amount,
+        )
 
     @Test
     fun the_unit_price_floors_rather_than_trusting_a_low_network_sample() {
@@ -308,16 +350,13 @@ class KaminoTransactionPreparerTest {
 
     @Test
     fun a_withdraw_is_refused_when_it_is_not_the_wallet_s_own_transaction() {
-        // The fixture's fee payer is ; validating it against a different signer must fail,
-        // which is what stops the app signing on another account's behalf.
+        // The fixture's fee payer is KaminoFixtures.WALLET; validating it against a different
+        // signer must fail, which is what stops the app signing on another account's behalf.
         val other = "HDsayqAsDWy3QvANGqh2yNraqcD8Fnjgh73Mhb3WRS5E"
         val rejection =
             assertThrows(KaminoTransactionRejected::class.java) {
-                KaminoTransactionValidator.validate(
-                    decoded =
-                        KaminoTransactionDecoder.decode(
-                            KaminoAttributionMemo.append(KaminoFixtures.WITHDRAW)
-                        ),
+                validate(
+                    transaction = KaminoFixtures.WITHDRAW,
                     vault = KaminoVaultRegistry.STEAKHOUSE_USDC,
                     action = KaminoAction.WITHDRAW,
                     signerAddress = other,
@@ -335,11 +374,7 @@ class KaminoTransactionPreparerTest {
         // account, then a SyncNative over it. Both are opcodes the validator has to permit by name,
         // which is the half of an allow-list that a deny-list never had to get right — too strict a
         // rule here refuses every real SOL deposit rather than letting an extra one through.
-        val prepared =
-            prepare(
-                api = KaminoFixtureApi(KaminoFixtures.ALLEZ_SOL_DEPOSIT),
-                vault = KaminoVaultRegistry.ALLEZ_SOL,
-            )
+        val prepared = prepareSolDeposit()
         val instructions = KaminoTransactionDecoder.decode(prepared).instructions
 
         // The address the validator derives locally, cross-checked against the one Kamino chose.
@@ -381,12 +416,10 @@ class KaminoTransactionPreparerTest {
 
         val rejection =
             assertThrows(KaminoTransactionRejected::class.java) {
-                KaminoTransactionValidator.validate(
-                    decoded = decoded,
+                validate(
+                    transaction = KaminoFixtures.ALLEZ_SOL_WITHDRAW,
                     vault = KaminoVaultRegistry.ALLEZ_SOL,
                     action = KaminoAction.WITHDRAW,
-                    signerAddress = KaminoFixtures.WALLET,
-                    wrappedSolAccount = KaminoFixtures.WALLET_WRAPPED_SOL,
                 )
             }
         assertTrue(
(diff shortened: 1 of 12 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionValidator.kt
```diff
@@ -178,66 +179,184 @@ object KaminoTransactionValidator {
...
+     * kVault program's own CPI.
      */
     private fun rejectValueMovementAwayFromTheSigner(
         instructions: List<KaminoTxInstruction>,
         vault: KaminoVault,
+        signerAddress: String?,
         wrappedSolAccount: String?,
     ) {
         instructions.forEachIndexed { index, instruction ->
             when (instruction.programId) {
```
rkokhatskyi, line 446: This `when` has no Farms arm, so `farms::transfer_ownership` rides through unchecked and its `new_owner` never signs, which moves the whole staked position on the signature the message already carries. iOS allow-lists exactly four farms discriminators at `KaminoInstructionSequence.swift:301`

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionValidator.kt
```diff
@@ -128,7 +129,7 @@ object KaminoTransactionValidator {
             reject("memo must be the final instruction")
         }
 
-        rejectValueMovementAwayFromTheSigner(instructions, vault, wrappedSolAccount)
+        rejectValueMovementAwayFromTheSigner(instructions, vault, signerAddress, wrappedSolAccount)
```
rkokhatskyi: Nothing binds the kVault instruction to this vault, so a substituted `vault_state` deposits the balance into a vault the response made through the permissionless `init_vault` while the screen shows the registry address. Pinning account slot 7 to `defaultTokenAddress(vault.sharesMint)` settles it offline, as `KaminoTransactionDecoder.swift:90` does

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionValidator.kt
```diff
@@ -178,66 +179,184 @@ object KaminoTransactionValidator {
...
+     * message already carries, and neither one a transfer.
+     */
+    private fun rejectUnlessWrap(
+        index: Int,
+        instruction: KaminoTxInstruction,
+        vault: KaminoVault,
+        wrappedSolAccount: String?,
+    ) {
+        val discriminator =
+            littleEndianInt(instruction.data.take(4))
```
rkokhatskyi, line 839: `rejectUnlessWrap` reads only the four discriminator bytes, so the transfer's lamports stay unbound and a response can wrap the whole SOL balance while the screen shows the typed amount. iOS pins `lamports == amount` from the same bytes at `KaminoTransactionDecoder.swift:398`

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionValidator.kt
aminsato: Real, and fixed in `65f8eb065`.

Every launch vault has a farm, so the program had to be allow-listed and its instructions then went unread — the one gap where an instruction moves the whole position rather than an amount. `farms::transfer_ownership` needs only the fee-payer signature the message already carries.

The arm now names the same four you pointed at, each derived as `sha256("global:<name>")[0..8]` and then found in the captured bytes rather than pasted out of them:

- `initialize_user` `6f11b9fa3c7a26fe`
- `stake` `ceb0ca12c8d1b36c`
- `unstake` `5a5f6b2acd7c32e1`
- `withdraw_unstaked_deposits` `2466bb31dc248443`

Anything else is refused, `transfer_ownership` (`41b1d749352d632f`) included. On top of the discriminator, what stays static in the captured shapes is pinned too: the authority at slot 0, the wallet's own share account at slot 4 for the stake and slot 3 for the unstaked withdrawal, and the stake argument at `u64::MAX` — which is what Kamino always passes, meaning "stake the whole balance". Both captured deposits carry exactly `ceb0ca12c8d1b36c ffffffffffffffff`.

Two things I did not mirror from iOS, both deliberate:

- The farm and the farms **user state** are lookup-table entries in every captured response and this decode resolves no tables, so they cannot be compared here. iOS recomputes the user state as a program address over `(farm, owner)`; that needs PDA derivation this module does not have, and it is JNI-free on purpose. Left out rather than faked.
- The unstake amount is bounded by the shares the withdraw burns rather than pinned to `requested − alreadyUnstaked`. Android's preparer does not hold that split — the API sizes it — so the exact figure is not available here. The bound needs no balance and still refuses a release larger than the withdraw, which is the shape that would strand shares outside the farm. It is read as the `u128` it is, scaled by `10^18`; reading it as a `u64` would compare the low half and a number that means nothing.

Action-gated the way your template is: a deposit stakes, a withdraw unstakes, neither does both. That rests on `KaminoInstructionSequence.expected` rather than on a captured Android fixture — no staked withdraw has been captured, since it needs a funded position.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionValidator.kt
aminsato: Real, and fixed in `65f8eb065`. Slot 7 settles it, as you said.

I decoded all four captured responses offline to check the premise before writing the check, and it holds in every one: slot 7 is a **static** key — never a lookup-table entry — and it equals `SolanaAddress(wallet).defaultTokenAddress(vault.sharesMint)`:

| fixture | slot 7 | derived share ATA |
|---|---|---|
| `DEPOSIT` / `WITHDRAW` (Steakhouse) | `GSayQpRaoh1LFdBbja4vensNKDfihcixzCcQShKMCdMJ` | same |
| `ALLEZ_SOL_DEPOSIT` / `ALLEZ_SOL_WITHDRAW` | `Hq6N6sNE638VLULNEeAZRTMFmYtsG9ZLLPJYefxwPNWf` | same |

The preparer derives it (and the token account) locally and the validator compares exactly, so an account the response hides behind its own table decodes as unresolved and fails rather than skips.

Two additions on the same instruction, since the check was being added anyway:

- **Slot 0**, the authority the move runs under, pinned to the signer.
- **Slot 6 on a deposit / slot 5 on a withdraw**, the wallet's token account for the vault's underlying mint. That is where a deposit's tokens come from and where a withdraw *pays out* — the same substitution as yours, on the leg that hands over the proceeds rather than the position. Static in all four fixtures too.

The vault slot itself (1) is still not compared: it is a lookup-table entry everywhere, so comparing it would be a check that passes by being unreadable. The instruction is identified by its discriminator now as well — `deposit` `f223c68952e1f2b6` for a deposit, either `withdraw` `b712469c946da122` or `withdraw_from_available` `1383709baadc2239` for a withdraw, which share an account prefix by IDL construction — so a deposit cannot arrive as a withdraw or be read under the wrong layout.

`every_captured_shape_puts_the_wallets_own_accounts_where_the_layout_says` pins the indices against all four responses, so a layout drift shows there rather than as a refusal on someone's device. And `a_response_that_names_another_vault_is_refused_even_though_the_screen_says_this_one` runs the captured Steakhouse deposit against the Allez SOL vault: refused twice over, first at the account it creates, then — with that instruction dropped — at the kVault share slot.

One assumption worth naming: `defaultTokenAddress` derives the classic Token-program ATA. All three curated vaults use it, confirmed against the fixtures. A Token-2022 mint added to the registry would be refused rather than mis-checked, and the layout test above would say so.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionValidator.kt
aminsato: Real, and fixed in `65f8eb065`. The lamports are pinned to the deposit now, as `KaminoTransactionDecoder.swift:398` does.

The amount comes from the very string the app handed the API, converted at the vault's own scale — so it is a comparison against the *request*, not a restatement of the response. The amount screen already quantises input to the token's precision before it gets here, so the conversion is exact; it floors rather than throwing on sub-unit precision, which is unreachable in practice but is the direction the rest of the flow rounds.

Same treatment for the kVault `u64` while I was there, since the wrap is only half of it: on a plain-token vault there is no transfer at all, so the amount on screen was unbound there too. `deposit` takes tokens and `withdraw` takes shares, so the expected figure is scaled by `tokenDecimals` or `sharesDecimals` per action.

The captured Allez SOL deposit is the evidence both are right: its wrap carries `0200000080f0fa0200000000` — 50,000,000 lamports — and its `kvault::deposit` carries the same 50,000,000. `the_wrap_moves_exactly_what_is_being_deposited` asserts the figure and then asks for 0.5 SOL against those bytes, which is refused with "wraps 50000000 lamports rather than the 500000000 this app asked to deposit".

Two consequences that follow from checking amounts instruction by instruction, both closed in the same commit:

- **A second kVault instruction** carrying the figure on screen would pass its own check and move the money twice. Exactly one now.
- **A second wrap** would do the same with lamports. At most one System instruction.

The withdraw-everything sentinel now runs ahead of the walk: it fails the plain amount comparison too, but it is the refusal worth naming.

aminsato:
Both CodeRabbit nitpicks were real and are fixed in `65f8eb065` — replying here because review-body nitpicks carry no thread.

**Empty instruction data and Token-2022 dispatch.** Added, and the branch it named as reachable really was:

- `an SPL token instruction carrying no discriminator at all is refused`
- `a System instruction carrying no discriminator at all is refused`
- `Token-2022 is held to the same allow-list as Token` — an instruction the Token program may not carry must not become permitted by arriving under the newer program id.

**The stale `ALLEZ_SOL` comment.** Correct, and it had become load-bearing: the compute-budget test still prepared the *USDC* fixture against the SOL vault, with a comment saying no SOL response had been captured — which stopped being true when the fixture landed in this PR. It now runs against `KaminoFixtures.ALLEZ_SOL_DEPOSIT` at the 0.05 SOL that response actually wraps, so the 350,000-unit limit is pinned to the wrap-and-sync shape it is sized for. The comment is gone rather than reworded.

Everything else in that review round was addressed with Roman's three findings, in the threads above.

Green: 3028 `:data` unit tests, 33 Kamino instrumented tests on an emulator, `:app:compileDebugKotlin`. The validator suite went 33 → 62 tests and the preparer suite 17 → 22; removing the farms allow-list, the share-account pin or the lamports pin each fails at least one of them.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionValidator.kt
```diff
@@ -160,84 +266,467 @@ object KaminoTransactionValidator {
...
+                    FARMS_STAKE,
+                    FARMS_UNSTAKE,
+                    FARMS_WITHDRAW_UNSTAKED_DEPOSITS,
+                )
+                .any { data.startsWith(it) }
+        if (!isOneWePerform) {
+            reject(
+                "instruction $index is a farms instruction no Kamino deposit or withdraw performs"
+            )
+        }
```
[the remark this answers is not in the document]
aminsato: Real, and fixed in `a03285429`.

The reasoning that produced the kVault and System counts applies here and I only carried half of it over: the farms arm got its per-instruction checks and not the count, so the harm the `unstake` bound is written to prevent is reachable by repeating the instruction it bounds. Two releases each carrying the requested figure both sit within the bound, the withdraw burns that figure once, and the surplus leaves the farm to sit in the share account — earning nothing, and invisible on a screen showing one amount.

I took the stricter form you named rather than the `> 2` diff, because that is what iOS does. Every farms step in `KaminoInstructionSequence.expected` is `repeatable: false`, and `match` fails an instruction left over as `unexpectedInstruction` — so a deposit's `initialize_user` alongside its `stake` stays legal while two of either do not, and `> 2` would have let two unstakes through in a transaction that carried nothing else.

The other three are worth counting for their own reasons, not just for symmetry:

- `initialize_user` — creation is funded by the fee payer, so each copy is the wallet paying rent for a farm user state. The farm account is a lookup-table entry this decode does not resolve, so it need not be this vault's; same shape as the Associated-Token rule already in the file.
- `stake` — the argument is the whole-balance sentinel, so a second one stakes into whatever farm it names, and only the pinned share account and authority constrain it.
- `withdraw_unstaked_deposits` — moves released shares a second time.

The four discriminators now live in one `FARMS_INSTRUCTIONS` list that both the count guard and the walk's allow-list read, so a farms instruction cannot become permitted in one place and unbounded in the other.

Two new tests, both verified to fail with the guard reverted: `a second unstake is refused rather than releasing twice what the withdraw burns` and `each of the four farms instructions is bounded to one, not merely checked one at a time`. The existing `the farms instructions a real deposit and withdraw carry are allowed` already pins the other direction — one unstake plus one unstaked-share withdrawal in one transaction still passes — so the counting cannot have made the real staked withdraw a refusal.

Green: 3030 `:data` unit tests, 33 Kamino instrumented tests, `:app:compileDebugKotlin`.

aminsato:
The remaining nitpick was real too, and is fixed in `a03285429` — replying here because review-body nitpicks carry no thread.

**The two `amountBaseUnits` defaults.** They disagreed, and the smaller one was dead rather than merely inconsistent: both of `validate`'s callers reject before the amount is ever compared — one on the fee payer, one on the withdraw-everything sentinel — so `BigInteger.ONE` was never held against a fixture that moves 1 USDC. A later test reaching `rejectUnlessAmountIs` through that helper would have failed on the stand-in amount instead of on the property it was written for, and a `contains` assertion would still have gone green on the wrong refusal. Both helpers now take `CAPTURED_AMOUNT_BASE_UNITS`, sitting next to the `AMOUNT = "1"` it is the base-unit form of.

The Major finding on the farms counts is answered in its own thread.

Green: 3030 `:data` unit tests, 33 Kamino instrumented tests on an emulator, `:app:compileDebugKotlin`. The validator suite is 64 tests; reverting the farms count guard fails two of them.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionValidator.kt
```diff
@@ -160,84 +297,463 @@ object KaminoTransactionValidator {
...
+     * else.
+     *
+     * A deposit stakes its shares into the vault's farm, so the shares never land in the wallet as
+     * a token balance — which is why the farms program is allow-listed as a program at all. That
+     * made it the one program here whose instructions went unread: `transfer_ownership` hands the
+     * whole staked position to an account that never signs, and nothing about it looks like a
+     * transfer. iOS names exactly these four at `KaminoInstructionSequence.swift:301`, and the
+     * shapes match its live template — a deposit stakes, a withdraw unstakes, neither does both.
+     *
+     * The farm and the farms user state are lookup-table entries in every captured response, so
```
rkokhatskyi: Nothing binds a farms instruction to this vault, and the user state is not a lookup table entry: it is a static key equal to `find_program_address(["user", vault.farm, signer])`. So `farms::stake` sweeps the whole share balance into a farm the response created, which iOS refuses at `KaminoTransactionValidator.swift:1006`

data/src/androidTest/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionPreparerTest.kt
```diff
@@ -328,10 +368,219 @@ class KaminoTransactionPreparerTest {
...
+    }
+
+    @Test
+    fun the_captured_Allez_SOL_withdraw_unwraps_to_the_wallet_and_nowhere_else() {
+        // The instruction #5603 is about. A compromised response need only re-address this one
+        // account to take the whole withdrawn amount, while every other instruction still reads as
+        // the withdraw the user asked for.
+        //
+        // Asserted twice over: the shape itself, and that validating gets as far as the sentinel
+        // refusal — which sits after the value-movement rules, so reaching it is proof the unwrap
```
rkokhatskyi: This commit moved the sentinel check to `KaminoTransactionValidator.kt:261`, ahead of the value movement walk. Redirecting this fixture's `CloseAccount` to an attacker now yields that same sentinel rejection, so the second half of the test proves nothing about the unwrap

rkokhatskyi (changes requested):

johnnyluo (approved):
LGTM

rkokhatskyi (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionValidator.kt
aminsato: Following up on this thread, because I got one of its facts wrong.

I said the farms **user state** was a lookup-table entry and so could not be compared offline, and left it out on that basis. It is not. Decoding all four captured responses, it is a *static* key in both deposits, and it equals the address derived from the farm the registry already pins:

| fixture | instruction | slot | on the wire | derived from `(farm, owner)` |
|---|---|---|---|---|
| `DEPOSIT` | `initialize_user` | 4 | `A1b83WVHAKXeRQAHsdAzJY23ShXCPjqKshzmFFXwGP4Z` | same |
| `DEPOSIT` | `stake` | 1 | `A1b83WVHAKXeRQAHsdAzJY23ShXCPjqKshzmFFXwGP4Z` | same |
| `ALLEZ_SOL_DEPOSIT` | `initialize_user` | 4 | `8ULTfRg47DWt5VBDT7UURTPW6P5Fc5vPMfncfPKpZc3J` | same |
| `ALLEZ_SOL_DEPOSIT` | `stake` | 1 | `8ULTfRg47DWt5VBDT7UURTPW6P5Fc5vPMfncfPKpZc3J` | same |

It is the *farm* slot beside it that is the lookup-table entry — `stake` slot 2, unresolved in both. So the account I dismissed is the only thing that binds a farms instruction to this vault offline, and pinning it is now in `a65db2f8c`.

That matters here rather than being tidiness. With only the authority pinned, the response chose the farm, and `farms::stake` carries the whole-balance sentinel — so a substituted farm takes the wallet's entire share balance, not an amount. The allow-list closed `transfer_ownership`; this closes the same outcome reached through an instruction that *is* on the list.

Deriving it needs `find_program_address`, and WalletCore exposes only the associated-token special case, so the general form and the Base58 under it are added in pure Kotlin — JNI-free deliberately, since these are the checks most worth running in plain unit tests rather than only on a device. It is cross-checked against three associated token accounts WalletCore itself derives (the wrapped-SOL account and both share accounts, all three already trusted elsewhere in this PR). Two of them land on a bump below 255, so the ed25519 on-curve test is pinned by those vectors rather than merely present.

Two deliberate differences from `KaminoTransactionValidator.swift:1006`:

- **Pinned on all four instructions, not the two iOS names.** iOS pins the user state on `unstake` and `withdraw_unstaked_deposits` and pins the farm slot directly on all four, which it can because it resolves lookup tables online. Offline that slot reads as unresolved, so the user state has to carry the binding on `initialize_user` and `stake` as well.
- **Slot 4 on `initialize_user`, slot 1 elsewhere.** That instruction creates the account, so the IDL puts the authority, the payer, the owner and the delegatee ahead of it — consistent with your `FarmsInitializeUser.farm = 5`. Both indices are asserted against the captured responses so a layout drift shows in CI rather than as a refusal on someone's device.

Still not derived here: the farm address itself, which stays a registry constant. And `unstake` / `withdraw_unstaked_deposits` take slot 1 from your account tables rather than from an Android fixture — no staked withdraw has been captured, same caveat as the action gating.

5 new tests assert the pin (including the captured Steakhouse deposit refused against this wallet's state in the Allez SOL farm), each verified to fail with it reverted; 12 more cover the derivation and Base58. Green: 3046 `:data` unit tests, 35 Kamino instrumented, `:app:compileDebugKotlin`.

PR #5654, opened 2026-08-20
fix(defi): check the wallet holds the SOL a Kamino transaction costs

### LINKED ISSUE #5607: Kamino deposit and withdraw never check the wallet holds enough SOL for fees

aminsato:
Closes #5607

## The problem

Every Kamino transaction is paid for in SOL, whatever the vault's underlying token is — but the token form only ever reads the token balance. `spendableBalance` returns that balance in full for a non-native coin, `submit()` validates the amount against the vault minimum and the available balance and nothing else, and the withdraw path reserves nothing at all. Downstream closed none of it: `VerifyDepositViewModel.checkFeeAffordability` returned immediately for any chain that is not QBTC, so Sign stayed enabled.

A vault holding USDC and no SOL therefore passed the amount form, the verify screen and the entire MPC ceremony, and learned about it from the node's rejection at broadcast. No funds at risk — preflight means nothing lands — but the ceremony is spent.

## The rent is most of the cost

The old reserve treated rent as a wrapped-SOL special case (`if (vault.tokenMint != WRAPPED_SOL_MINT) return ZERO`). That is right for the *deposit form's* headroom — a token deposit spends no rent out of the token balance — and wrong for the question "can this wallet pay?".

Decoding the mainnet USDC deposit already checked in as a fixture (`KaminoFixtures.DEPOSIT`) shows what a token vault actually sends:

```
[0] AssociatedToken  createIdempotent      <- share account
[1] KvauGMsp…        kVault::deposit
[2] FarmsPZp…        farms::initialize_user <- farms user-state
[3] FarmsPZp…        farms::stake
```

Both created accounts are paid for in SOL. At the pinned figures that is 2,039,280 + 6,299,080 lamports against a fee of 1,320,000 — roughly seven times the fee — so a fee-only check would still wave through a wallet that cannot fund the deposit. Only the wrap itself stays gated on the mint, and each term is waived when the account already exists, so a redeposit costs the fee alone. `KaminoFixtures.WITHDRAW` opens with one `createIdempotent` before `kVault::withdraw`, so a withdraw reserves one token-account rent; per that fixture's own note the *staked* withdraw shape carries a second creation nobody has captured, so this is a floor rather than a promise.

Deterministic thresholds, since `getMedianPriorityFee` floors at 1,000,000 and `MAX_UNIT_PRICE` caps at the same value:

| Case | SOL required |
|---|---|
| First deposit (fee + share account + user-state) | 0.00965836 |
| Redeposit | 0.00132 |
| Withdraw, destination account exists | 0.0014 |
| Withdraw, destination account gone | 0.00343928 |

## What changed

- `KaminoAmountViewModel.refuseIfSolCannotCover` on the shared submit path, so it covers deposit and withdraw. On a native-SOL deposit the amount is required on top; a withdraw's amount arrives from the position, so it is not.
- The refusal reuses `R.string.insufficient_native_token` ("Not enough %1$s to cover gas fees") rather than adding a Kamino phrasing of the same sentence to ten locale files, so `KaminoAmountRefused` now carries format arguments. **No new strings, no localisation work.**
- Rent generalised past the wrapped-SOL gate, and given a withdraw branch.
- `checkFeeAffordability` now checks the fee against the balance the fee is denominated in: same token → fee + amount out of one balance (QBTC's shape, unchanged); different token → the fee alone against the chain's native balance.

## Two judgement calls worth a look

**1. Fail-open outside QBTC.** Dropping the gate while keeping fail-closed would make one flaky RPC read disable Sign on every deposit flow in the app. QBTC keeps failing closed — it mirrors iOS `canCoverVoteFee` and guards a vote the chain refuses outright — and everywhere else an unresolved balance leaves Sign enabled, with the amount already sized against the same balance by the form. The issue says to drop the gate full stop, so say the word if you want it strict everywhere.

**2. The displayed network fee still omits token-vault rent.** That figure is a cross-device contract: a co-signer derives it from the relayed payload and iOS derives base + price × limit (#5644). Quoting a term the other device does not would put the two screens back out of step. The wallet is still checked against the full cost before the form lets anything through.

## Testing

- 6 new tests in `KaminoAmountViewModelTest`, 3 new/rewritten in `VerifyDepositViewModelTest`. The behavioural ones were each confirmed to fail under three separate reverts: gate removed, rent re-gated to wrapped-SOL only, and the QBTC-only gate restored.
- Full `:app:testDebugUnitTest` green, `:app:lintDebug` green, ktfmt run on `:app`.
- On device: a vault with USDC and under 0.00966 SOL is refused at Continue with "Not enough SOL to cover gas fees"; at 0.009 it is still refused (fee covered, rent not); above 0.00966 it reaches Verify. A wallet with a position redeposits at 0.00132.

## Overlaps #5646

That PR rewrites `spendableBalance` and extracts `KaminoDepositRentReserve` / `kaminoNetworkFeeLamports` into `:data`. Whichever merges second rebases; on top of #5646 this check should size itself through `kaminoNetworkFeeLamports` instead of its own local sum. Happy to take that order either way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Deposit verification now checks fee affordability against the selected fee token, including source and native tokens.
  * Insufficient-balance messages identify the token required for fees.
  * Unavailable balance data is no longer treated as zero.
  * Solana Kamino transactions now accurately account for network fees, rent reserves, and required SOL balances.

* **Improvements**
  * Transactions are blocked when available balances cannot cover required fees or deposits.
  * Recorded fees now reflect the actual amount charged when it differs from the estimate.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
@@ -250,6 +250,9 @@ constructor(
      * Solana fee in SOL, so a vault holding only USDC passed this screen and the entire MPC
      * ceremony before being rejected at broadcast (#5607).
      *
+     * Weighed against [DepositTransaction.chargedFees] where the transaction sets it, since a
+     * quoted fee is not always the charge — see that field.
+     *
      * An unresolved balance fails closed on QBTC alone. There the gate mirrors iOS
      * `canCoverVoteFee` and guards a vote the chain refuses outright, so an unverified balance must
      * not start it. Reading one flaky RPC call as "cannot afford" everywhere else would disable
@@ -257,7 +260,11 @@ constructor(
      */
     private suspend fun checkFeeAffordability(transaction: DepositTransaction) {
         val srcToken = transaction.srcToken
-        val fee = transaction.estimatedFees
+        // The charge, not the quote, wherever the transaction distinguishes them. A Kamino fee row
+        // is padded to the 1,000,000-lamport placeholder both platforms display against a runtime
+        // charge of 5,000, and demanding the padding back disabled Sign on the withdraw that exits
+        // a Max deposit — the padding is exactly what such a deposit leaves in the wallet.
+        val fee = transaction.chargedFees ?: transaction.estimatedFees
         // Tickers rather than coins: TokenValue records the unit it was built with, which is the
         // only statement the transaction makes about which balance pays the fee.
         val feeLeavesSourceBalance = fee.unit.equals(srcToken.ticker, ignoreCase = true)
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
@@ -33,7 +33,6 @@ import java.math.BigInteger
 import javax.inject.Inject
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -88,9 +87,9 @@ internal data class VerifyDepositUiModel(
     val hasFastSign: Boolean = false,
     val isLoading: Boolean = false,
     /**
-     * Whether the source account balance can cover the required network fee (plus any sent amount).
-     * When false the Sign button is disabled and an inline fee-balance error is shown so the
-     * keysign ceremony cannot start for an account that can never pay the fee (issue #5044).
+     * Whether the balance the network fee is drawn from can cover it. When false the Sign button is
+     * disabled and an inline fee-balance error is shown so the keysign ceremony cannot start for an
+     * account that can never pay the fee (issues #5044, #5607).
      */
     val hasEnoughBalance: Boolean = true,
     val insufficientBalanceError: UiText? = null,
@@ -226,16 +225,19 @@ constructor(
     }
 
     /**
-     * Verifies the wallet can cover what the transaction spends, before the keysign ceremony can be
-     * started.
+     * Verifies the wallet can pay the network fee, before the keysign ceremony can be started.
      *
-     * The amount and the fee are not always drawn from one balance. Where the fee is denominated in
-     * the source token they are, and the balance has to carry both — QBTC's vote is that shape, and
-     * its whole cost is the fee (#5044 / #5043). Where they differ the fee is checked on its own
-     * against the chain's native balance, and the amount is left to the form that already sized it
-     * against the source balance. A Kamino USDC deposit is the second shape: it pays its Solana fee
-     * in SOL, so a vault holding only USDC passed this screen and the entire MPC ceremony before
-     * being rejected at broadcast (#5607).
+     * Only the fee: the amount is left to the form that sized it against the source balance in the
+     * first place. A deposit's `srcTokenValue` is not reliably money leaving the wallet — a Kamino
+     * withdraw, a stake withdraw, Move Stake, Finish Move and a Cosmos undelegate all stage the
+     * position's own size in it, and it arrives rather than departs. Adding it here disabled Sign
+     * on every one of those whenever the position was bigger than the liquid balance.
+     *
+     * What varies is which balance the fee comes out of. Where it is denominated in the source
+     * token it leaves that one — QBTC's vote is that shape, and its whole cost is the fee (#5044
+     * / #5043). Otherwise it leaves the chain's native balance: a Kamino USDC deposit pays its
+     * Solana fee in SOL, so a vault holding only USDC passed this screen and the entire MPC
+     * ceremony before being rejected at broadcast (#5607).
      *
      * An unresolved balance fails closed on QBTC alone. There the gate mirrors iOS
      * `canCoverVoteFee` and guards a vote the chain refuses outright, so an unverified balance must
@@ -252,8 +254,6 @@ constructor(
         val feeToken =
             if (feeLeavesSourceBalance) srcToken
             else srcToken.chain.nativeToken.copy(address = transaction.srcAddress)
-        val requiredSpend =
-            if (feeLeavesSourceBalance) fee.value + transaction.srcTokenValue.value else fee.value
 
         val balance =
             readBalance(
@@ -262,7 +262,7 @@ constructor(
                 failClosed = srcToken.chain == Chain.Qbtc,
             ) ?: return
 
-        if (balance < requiredSpend) {
+        if (balance < fee.value) {
             state.update {
                 it.copy(
                     hasEnoughBalance = false,
@@ -279,29 +279,28 @@ constructor(
     /**
      * The balance in [token], or null when it could not be read — disabling Sign on the way out
      * when [failClosed], so an unverified balance cannot start a ceremony that must not run.
+     *
+     * Read through [BalanceRepository.getBalanceOrNull] rather than `getTokenValue`, which on
+     * Solana cannot say "unknown": `SolanaApi.getBalance` answers every RPC error with zero, so a
+     * flaky node read as an empty wallet and refused a funded one.
      */
     private suspend fun readBalance(
         address: String,
         token: Coin,
         failClosed: Boolean,
-    ): BigInteger? =
-        try {
-            balanceRepository.getTokenValue(address, token).first().value
-        } catch (e: kotlinx.coroutines.CancellationException) {
-            throw e
-        } catch (e: Exception) {
-            Timber.e(e)
-            if (failClosed) {
-                state.update {
-                    it.copy(
-                        hasEnoughBalance = false,
-                        insufficientBalanceError =
-                            UiText.StringResource(R.string.network_connection_lost),
-                    )
-                }
+    ): BigInteger? {
+        val balance = balanceRepository.getBalanceOrNull(address, token)
+        if (balance == null && failClosed) {
+            state.update {
+                it.copy(
+                    hasEnoughBalance = false,
+                    insufficientBalanceError =
+                        UiText.StringResource(R.string.network_connection_lost),
+                )
             }
-            null
         }
+        return balance
+    }
 
     fun dismissError() {
         state.update { it.copy(errorText = null) }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
@@ -8,6 +8,9 @@ import androidx.navigation.toRoute
 import com.vultisig.wallet.R
 import com.vultisig.wallet.data.IoDispatcher
 import com.vultisig.wallet.data.models.Chain
+import com.vultisig.wallet.data.models.Coin
+import com.vultisig.wallet.data.models.DepositTransaction
+import com.vultisig.wallet.data.models.nativeToken
 import com.vultisig.wallet.data.repositories.AddressBookRepository
 import com.vultisig.wallet.data.repositories.BalanceRepository
 import com.vultisig.wallet.data.repositories.ChainAccountAddressRepository
@@ -26,6 +29,7 @@ import com.vultisig.wallet.ui.navigation.util.LaunchKeysignUseCase
 import com.vultisig.wallet.ui.utils.UiText
 import com.vultisig.wallet.ui.utils.resolveDstVaultName
 import dagger.hilt.android.lifecycle.HiltViewModel
+import java.math.BigInteger
 import javax.inject.Inject
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.flow.MutableStateFlow
@@ -222,42 +226,41 @@ constructor(
     }
 
     /**
-     * Verifies the source account holds enough native balance to cover the required network fee
-     * (plus any amount being sent) before the keysign ceremony can be started. Scoped to QBTC,
-     * whose unfunded accounts otherwise stage a vote that the chain rejects at broadcast (#5044
-     * / #5043). Fails closed: if the balance cannot be resolved (a network error, or an empty
-     * lookup result) the Sign button is left disabled rather than defaulting to affordable —
-     * mirroring iOS `canCoverVoteFee` — so an unverified balance can never start the doomed
-     * ceremony this guards against.
+     * Verifies the wallet can cover what the transaction spends, before the keysign ceremony can be
+     * started.
+     *
+     * The amount and the fee are not always drawn from one balance. Where the fee is denominated in
+     * the source token they are, and the balance has to carry both — QBTC's vote is that shape, and
+     * its whole cost is the fee (#5044 / #5043). Where they differ the fee is checked on its own
+     * against the chain's native balance, and the amount is left to the form that already sized it
+     * against the source balance. A Kamino USDC deposit is the second shape: it pays its Solana fee
+     * in SOL, so a vault holding only USDC passed this screen and the entire MPC ceremony before
+     * being rejected at broadcast (#5607).
+     *
+     * An unresolved balance fails closed on QBTC alone. There the gate mirrors iOS
+     * `canCoverVoteFee` and guards a vote the chain refuses outright, so an unverified balance must
+     * not start it. Reading one flaky RPC call as "cannot afford" everywhere else would disable
+     * Sign on transactions that are perfectly fundable.
      */
-    private suspend fun checkFeeAffordability(
-        transaction: com.vultisig.wallet.data.models.DepositTransaction
-    ) {
-        if (transaction.srcToken.chain != Chain.Qbtc) return
+    private suspend fun checkFeeAffordability(transaction: DepositTransaction) {
+        val srcToken = transaction.srcToken
+        val fee = transaction.estimatedFees
+        // Tickers rather than coins: TokenValue records the unit it was built with, which is the
+        // only statement the transaction makes about which balance pays the fee.
+        val feeLeavesSourceBalance = fee.unit.equals(srcToken.ticker, ignoreCase = true)
+
+        val feeToken =
+            if (feeLeavesSourceBalance) srcToken
+            else srcToken.chain.nativeToken.copy(address = transaction.srcAddress)
+        val requiredSpend =
+            if (feeLeavesSourceBalance) fee.value + transaction.srcTokenValue.value else fee.value
 
         val balance =
-            try {
-                balanceRepository
-                    .getTokenValue(transaction.srcAddress, transaction.srcToken)
-                    .first()
-                    .value
-            } catch (e: kotlinx.coroutines.CancellationException) {
-                throw e
-            } catch (e: Exception) {
-                Timber.e(e)
-                // Fail closed: an unresolved balance keeps Sign disabled so a QBTC voter can't
-                // start a doomed ceremony on a balance we never confirmed can cover the fee
-                // (#5044).
-                state.update {
-                    it.copy(
-                        hasEnoughBalance = false,
-                        insufficientBalanceError =
-                            UiText.StringResource(R.string.network_connection_lost),
-                    )
-                }
-                return
-            }
-        val requiredSpend = transaction.estimatedFees.value + transaction.srcTokenValue.value
+            readBalance(
+                address = transaction.srcAddress,
+                token = feeToken,
+                failClosed = srcToken.chain == Chain.Qbtc,
+            ) ?: return
 
         if (balance < requiredSpend) {
             state.update {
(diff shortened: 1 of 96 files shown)


johnnyluo (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
```diff
@@ -222,42 +226,41 @@ constructor(
...
+        val fee = transaction.estimatedFees
+        // Tickers rather than coins: TokenValue records the unit it was built with, which is the
+        // only statement the transaction makes about which balance pays the fee.
+        val feeLeavesSourceBalance = fee.unit.equals(srcToken.ticker, ignoreCase = true)
+
+        val feeToken =
+            if (feeLeavesSourceBalance) srcToken
+            else srcToken.chain.nativeToken.copy(address = transaction.srcAddress)
+        val requiredSpend =
+            if (feeLeavesSourceBalance) fee.value + transaction.srcTokenValue.value else fee.value
```
rkokhatskyi: `requiredSpend` adds `srcTokenValue` to the fee, but on a Kamino Allez SOL withdraw and at `SolanaStakingPositionsViewModel.kt:352` that value is lamports arriving from a vault or a stake account. Move Stake and Finish Move stage the same shape, so Sign is disabled on all four whenever the position exceeds liquid SOL

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -656,7 +727,56 @@ constructor(
...
+     * cannot be submitted at all.
+     */
+    private suspend fun refuseIfSolCannotCover(vault: KaminoVault, coin: Coin, amount: BigDecimal) {
+        val solCoin = Coins.Solana.SOL.copy(address = coin.address)
+        val balance =
+            runCatchingCancellable {
+                    balanceRepository.getTokenValue(coin.address, solCoin).first().value
+                }
+                .onFailure { Timber.w(it, "Kamino SOL balance read failed, fee check skipped") }
+                .getOrNull() ?: return
```
rkokhatskyi: This fallback never fires because `SolanaApi.getBalance` turns every RPC error into `BigInteger.ZERO`, so a flaky node reads as zero SOL and refuses a funded wallet. `readBalance` at `VerifyDepositViewModel.kt:292` has the same hole

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -656,7 +727,56 @@ constructor(
...
+
+        val cost = solCostLamports(vault, coin, action)
```
rkokhatskyi: `solCostLamports` charges rent whenever a probe cannot answer, because `tokenAccountExists` and `hasKaminoVaultPosition` both `getOrDefault(false)`, so one failed Kamino or Solana read raises `required` by up to 8,338,360 lamports and refuses a redeposit the wallet can pay for

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -252,22 +276,50 @@ constructor(
...
+     *
+     * Each term is waived when the account already exists, which is what makes a redeposit cost the
+     * fee alone. `createIdempotent` charges nothing for an account that is already there.
+     */
+    private suspend fun depositRentReserve(vault: KaminoVault, coin: Coin): BigInteger {
         val tokenAccountRent = tokenAccountRentReserve()
         val wrappedSolRent =
-            tokenAccountRent.takeUnless { tokenAccountExists(coin.address, vault.tokenMint) }
-                ?: BigInteger.ZERO
+            if (vault.tokenMint == KaminoVaultRegistry.WRAPPED_SOL_MINT) {
```
rkokhatskyi: Charging this rent on token vaults makes `hasKaminoVaultPosition` decisive, and it answers the wrong question: `KaminoWithdraw.kt:364` resolves `Empty` for the one base unit a max withdraw leaves staked, so a returning depositor pays 6,299,080 lamports for a `UserState` that still exists

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -533,9 +597,16 @@ constructor(
             // the same balance, on the wSOL ATA, share ATA and farm UserState it creates. Folded in
             // here too, or the amount plus this fee would undercount the wallet's starting balance
             // by exactly that rent on the verify screen.
+            //
+            // A token deposit's rent is deliberately left out, even though it spends it: this
+            // figure is a cross-device contract — a co-signer derives the same number from the
```
rkokhatskyi: This comment claims the other device derives the same figure, but the branch below adds up to 10,377,640 lamports of rent on a native SOL deposit while iOS `PrebuiltPayloadFee.swift:47` derives base plus price times limit, so the claim holds for token vaults only

app/src/test/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModelTest.kt
```diff
@@ -198,19 +205,82 @@ internal class VerifyDepositViewModelTest {
...
+    @Test
+    fun `a deposit is signable when the native balance covers the fee`() =
+        runTest(testDispatcher) {
+            givenTransaction(
+                chain = Chain.Solana,
+                feeValue = 9_658_360,
+                srcTicker = "USDC",
+                feeUnit = "SOL",
+            )
+            every { balanceRepository.getTokenValue(SRC_ADDRESS, any()) } returns
```
rkokhatskyi: `any()` here matches the source token read as well and `givenTransaction` pins `srcTokenValue` to zero, so nothing exercises the `fee.value + srcTokenValue` term and reverting it leaves the suite green

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
aminsato: Real, and it reaches further than the four. `checkFeeAffordability` now weighs the fee alone — pushed in 6ca0009aa.

Also staging an amount that arrives rather than leaves: `CosmosUndelegateViewModel.kt:262` and `CosmosRedelegateViewModel.kt:359`, both with `gasFee = TokenValue(..., token = coin)`, so Sign went dead whenever the stake exceeded the liquid balance. Move Stake is the sharpest case of all — the transaction it stages is a *deactivate*, so nothing leaves the wallet at any point.

Dropping the amount term rather than trying to tell an outgoing amount from an incoming one: QBTC's vote stages `srcTokenValue = ZERO` (`GovernanceViewModel.kt:304`), so its threshold is unchanged, and the KDoc already claimed this split — "the amount is left to the form that already sized it against the source balance". That is true of the forms that do spend the amount: `KaminoAmountViewModel.spendableBalance` subtracts the SOL headroom, `SolanaDelegateViewModel` computes `stakeableBalance = balance - (rent + fee)` and guards funding again at submit. A whitelist of "outgoing" operations would have been guesswork that the next flow silently falls out of.

New test pins it: `a withdraw is signable on the fee alone though its amount dwarfs the balance` — 5 SOL amount, 0.002 SOL balance, fee 0.00132. Verified it fails if the amount term goes back in.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
aminsato: Confirmed, in both places — the fallback was unreachable code. `SolanaApi.getBalance` returns `BigInteger.ZERO` from the `rpcResp.error != null` branch and from the catch-all, and `BalanceRepository.kt:561` sends native SOL straight to it, so a flaky node arrived as a well-formed zero and both gates read it as an empty wallet.

Fixed at the read rather than at either gate (6ca0009aa):

- `SolanaApi.getBalanceOrNull` reports the failure as null. `getBalance` now delegates to it with `?: BigInteger.ZERO`, so every caller that only displays a figure behaves exactly as before.
- `BalanceRepository.getBalanceOrNull(address, coin)` routes Solana native through that, and wraps every other chain's flow in `runCatchingCancellable` — those report a failed read by throwing, which is the same "unknown" once caught, and cancellation still propagates.

`refuseIfSolCannotCover` and `VerifyDepositViewModel.readBalance` both ask through the repository method now, so the QBTC fail-closed and the fail-open everywhere else key off a null that can actually occur.

Tests: `SolanaBalanceReadTest` pins that a json-rpc error and an unreachable node are null to `getBalanceOrNull` and zero to `getBalance` (verified to fail if the error branch goes back to returning zero), and `BalanceRepositoryBalanceOrNullTest` pins the routing, including that the zero-returning overload is never called as a stand-in and that a `CancellationException` is not swallowed into a null.

One adjacent case I left alone as out of scope for this PR: the same zero reaches the *display* paths, where `spendableBalance` and `SolanaDelegateViewModel.stakeableBalance` show "0 available" on a hiccup. Happy to file it.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -252,22 +276,50 @@ constructor(
...
+        action: KaminoAction,
+    ): BigInteger {
         val networkPrice = solanaApi.getMedianPriorityFee(listOf(coin.address))
         return KaminoComputeBudget.priorityFeeLamports(
             vault = vault,
-            action = KaminoAction.DEPOSIT,
+            action = action,
             networkPrice = networkPrice,
         )
     }
```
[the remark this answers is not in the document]
aminsato: Checked this one against `SolanaApi.kt:216-242` and it does not hold. `getMedianPriorityFee` wraps its whole body in a `try`, rethrows `CancellationException`, and returns `SOLANA_PRIORITY_FEE_PRICE` (1,000,000) from the catch — it cannot throw into `solCostLamports`, so `kamino_error_build_failed` is not reachable this way and the documented fallback at lines 742-745 stands.

The KDoc directly above `kaminoPriorityFee` (lines 269-278) says exactly this, and also why the guard would be actively harmful: `runCatchingCancellable { ... }.getOrNull()` is a no-op here, but if a null ever did surface it would pass null to `KaminoComputeBudget.priorityFeeLamports`, dropping the floor from 1,000,000 to `FALLBACK_UNIT_PRICE`'s 20,000 — the under-reserve that #5599 was filed for, just at a smaller scale.

Not applied. The other two findings on this review were real and are fixed in 6ca0009aa.

aminsato:
Pushed `6ca0009aa` — both review findings fixed, plus the three body-only nitpicks answered here since they have no thread of their own.

**Kotest assertions in the new tests** — applied to the tests this PR adds, leaving the file's pre-existing JUnit assertions where they are, as the finding itself suggested. Churning 60-odd untouched assertions would bury the behaviour changes under review.

**`readBalance` → `runCatchingCancellable`** — absorbed by the fix for the balance finding. `readBalance` no longer catches anything itself: `BalanceRepository.getBalanceOrNull` owns that, using `runCatchingCancellable` for the chains that signal a failed read by throwing.

**Caching the SOL cost computed during load** — not applied. The expensive term is the priority fee, and a cached one is a staleness risk at the moment it matters most; the reads it would save are not free of that trade. The displayed fee is deliberately derived from the compute budget the payload records rather than from this cost, for cross-device parity with iOS, so the two figures are not one cached value in the first place.

Test totals after the change: 2096 `:app` + 2997 `:data`, all green, plus `:app:lintDebug`. The two new behaviour tests were each verified to fail under a revert of their fix.

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
```diff
@@ -613,7 +644,58 @@ constructor(
...
+    private suspend fun refuseIfSolCannotCover(vault: KaminoVault, coin: Coin, amount: BigDecimal) {
+        val solCoin = Coins.Solana.SOL.copy(address = coin.address)
+        val balance =
+            balanceRepository.getBalanceOrNull(coin.address, solCoin)
+                ?: run {
+                    Timber.w("Kamino SOL balance unreadable, fee check skipped")
+                    return
+                }
+
+        val cost = solCostLamports(vault, coin, action)
```
rkokhatskyi: A Max deposit into Allez SOL leaves exactly 995,000 lamports, since `spendableBalance` reserves the padded `DefaultFeeInLamports` where the chain charges 5,000. This gate then demands 1,400,000 for a withdraw that really costs 405,000, so the position can only be exited by funding SOL from outside the app

app/src/main/java/com/vultisig/wallet/ui/models/deposit/VerifyDepositViewModel.kt
```diff
@@ -233,57 +236,83 @@ constructor(
...
-            }
-        val requiredSpend = transaction.estimatedFees.value + transaction.srcTokenValue.value
+            readBalance(
+                address = transaction.srcAddress,
+                token = feeToken,
+                failClosed = srcToken.chain == Chain.Qbtc,
+            ) ?: return
 
-        if (balance < requiredSpend) {
+        if (balance < fee.value) {
```
rkokhatskyi: `SolanaDelegateViewModel.kt:178` reserves `DefaultFeeInLamports` as headroom for a Max stake while the chain charges 5,000, so the wallet is left holding 995,000 and this check disables Sign on every unstake, stake withdraw and move stake, which really cost 5,000

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/solanastaking/KaminoAmountViewModel.kt
aminsato: Confirmed, to the lamport, and fixed in d978a8663 — thank you, this was a bad one: the gate refused the only transaction that could unwind the position it had just helped open.

The withdraw priority fee is 400,000 units × the clamped 1,000,000 µL price = 400,000 lamports, so the gate demanded 1,000,000 + 400,000 against a charge of 5,000 + 400,000. And the 995,000 gap is not a coincidence in your example — it is the same padding on both sides of the deposit: `spendableBalance` reserves `DefaultFeeInLamports` and the runtime deducts 5,000, so a Max deposit leaves exactly the padding behind and the gate then asked for exactly the padding back.

One thing you did not name that had to move with it: the verify screen ran the *same* padded check one step later (`checkFeeAffordability` reads `estimatedFees`), so fixing only the form would have moved the lock-out one screen along rather than removing it.

What landed:

- `kaminoChargeableLamports` beside `kaminoNetworkFeeLamports`, sharing every term but the base — 5,000 rather than the placeholder. Both affordability gates read it. The quoted figure is untouched, since it still has to equal what an iPhone co-signer derives from the relayed payload.
- `spendableBalance` deliberately **keeps** the padded reserve. Reserving more than the chain takes is headroom the wallet keeps, and it is the headroom the exit is later paid out of — trimming it to the charge would leave a Max deposit with ~0 lamports and strand the position for a different reason.
- The verify screen reads the transaction, not the form, so the charge travels on it: new `DepositTransaction.chargedFees`, null everywhere but Kamino. Named for the charge rather than a minimum on purpose — it lands either side of the quote: below it on the padding, above it on a token deposit's account rent, which neither device quotes because iOS cannot derive it from the payload.

Tests: `a withdraw is affordable on the padding a Max deposit leaves, not on the quoted fee` (995,000 lamports, accounts already the wallet's) and `a withdraw is refused one lamport below what the chain charges for it` (404,999), plus the same pair on the verify screen. Each verified to fail on a revert of the gate.

data/src/main/kotlin/com/vultisig/wallet/data/blockchain/solana/kamino/KaminoTransactionCost.kt
```diff
@@ -28,71 +28,113 @@ val FARM_USER_STATE_RENT_LAMPORTS: BigInteger = BigInteger.valueOf(6_299_080)
...
+     * ours covers, so this is a floor rather than a promise — see the fixture note in
+     * `KaminoFixtures`. Under-reserving here costs a wasted ceremony, which is the same failure the
+     * figure exists to reduce, not a worse one.
+     */
+    private suspend fun withdrawRent(vault: KaminoVault, walletAddress: String): BigInteger =
+        coroutineScope {
+            val tokenAccountRent = async { tokenAccountRentReserve() }
+            val hasDestination = async { tokenAccountExists(walletAddress, vault.tokenMint) }
+
+            if (hasDestination.await()) BigInteger.ZERO else tokenAccountRent.await()
```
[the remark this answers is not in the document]
aminsato: Valid, and fixed in d978a8663 — and the finding is better than the KDoc it was aimed at. I had written the floor as unavoidable because no capture of ours covers the staked shape, but the second account is derivable without one: `KaminoTransactionValidator.rejectUnlessCreatingOneOfTheWalletsOwnAccounts` only ever permits a Kamino transaction to create two accounts, the wallet's token account and its share account. A staked exit runs `farms::unstake` and `farms::withdraw_unstaked_deposits` ahead of `kVault::withdraw`, and those shares have to land in the wallet's own share account — so the extra `createIdempotent` can only be that one. Anything else is a transaction this app refuses to sign in the first place.

`withdrawRent` now reserves both terms, each waived when the account already exists.

One deviation from your prompt: the share term is **not** gated on the position being staked. Shares that are already unstaked are sitting in that very account, so a wallet without one has no unstaked shares to withdraw — the staked-ness check would only add a positions read that cannot change the answer. `createIdempotent` charges nothing for an account that is already there, which is what keeps the ordinary exit costing the fee alone.

Exact-balance tests as asked, either side of the boundary: `a withdraw reserves the share account a staked exit releases into` at 4,483,559 lamports (405,000 fee + both 2,039,280 rents, less one) refuses, and `a withdraw goes through on exactly the fee and both account rents` at 4,483,560 passes. The first fails on a revert of the share term.

Worth noting alongside this: @rkokhatskyi found the opposite defect on the same code in this round — the gates were demanding the *padded* 1,000,000-lamport quote where the chain charges 5,000, which stranded a maxed-out position. Both are fixed in the same commit, and they do not collide: the rent is real SOL that has to be there, the padding never was.

aminsato:
Round 2 pushed as `d978a8663`. Both inline findings were real and are answered in their own threads; this covers the body-only nitpick and the shape of the change.

**@rkokhatskyi's finding, in one line:** the affordability gates were weighing the wallet against the fee we *quote* rather than the fee the chain *charges*. A Kamino fee row is padded to `SolanaHelper.DefaultFeeInLamports` — 1,000,000 lamports where the runtime deducts 5,000 — so that an iPhone co-signer reading the relayed payload quotes the same figure as the device that built it. `spendableBalance` reserves the padded base, so a Max deposit into the SOL vault leaves exactly 995,000 lamports of padding behind, and both gates then demanded 1,400,000 back for the withdraw that exits the position, which costs 405,000. The position could only be unwound with SOL from outside the app.

`kaminoChargeableLamports` now sits beside the quoted figure sharing every term but the base, and both gates read it. The quoted figure is untouched — it is still the cross-device contract. The Max reserve keeps its padding on purpose: that headroom is what the exit is paid out of.

CodeRabbit's Major was real too, and pointed the other way — a staked withdraw's second account creation was unreserved. It turned out to be derivable rather than the floor I had documented, since the validator only permits two creations and the staked shape's extra one can only be the wallet's share account.

**Nitpick (`JoinDepositKaminoTest`):** taken — `assertThrows` → Kotest `shouldThrow`.

Green: `:app` 2155 tests, `:data` 3110 tests, `:app:lintDebug`, ktfmt on both modules. 7 new tests, and each of the three fixes verified to fail under its own revert.

johnnyluo (approved):
LGTM

Issue #5319, opened 2026-07-18
Swap: THORChain/Maya inbound vault never revalidated at sign time, and no sign-time halt gate at all


realpaaao:
## Summary

Two gaps on the swap path, both verified absent:

1. **The THORChain/Maya inbound vault is never revalidated** between quote and signing.
2. **There is no sign-time halt/pause check at all.** iOS has this half; Android has neither half.

## 1. Inbound vault not revalidated

The signed vault comes from the quote and is read back verbatim at keysign with no network call:
- `SwapTransactionBuilder.kt:107` — `vaultAddress = quote.data.inboundAddress ?: srcAddress` (Maya `:133-138`, `:184`)
- `SwapTransactionBuilder.kt:58` — `quote.data.router ?: quote.data.inboundAddress ?: srcAddress`
- read back at `KeysignShareViewModel.kt:122-123`, `:153-158`
- `BroadcastTxUseCase.kt:81-215` ignores the payload entirely

Whole-repo proof that no swap path ever fetches it: `ThorChainApi.getTHORChainInboundAddresses()` (`ThorChainApi.kt:90`, `:347-349`) has **exactly one production caller** — `SecuredAssetLoader.kt:103`, which is the deposit path. There is no Maya equivalent anywhere in the repo.

Note the `?: srcAddress` fallback: a quote missing both `router` and `inboundAddress` addresses the swap to the user's own address rather than failing closed.

## 2. No sign-time halt gate

The halt/pause checks at `SecuredAssetLoader.kt:108-111` are deposit-only. Nothing on the swap path checks `halted` / `global_trading_paused` / `chain_trading_paused` before signing. iOS runs `assertSourceChainNotHalted` with a cache-bypassing fetch immediately before payload build (`DefaultSwapInteractor.swift:111-165`); Android has no equivalent.

## Staleness window — worse than iOS

The quote TTL is 1 minute (`SwapQuote.kt:80`), but the refresh job is **cancelled when the user leaves the form**: `SwapQuotePipelineController.kt:600-608` (`pause()`) ← `SwapFormViewModel.kt:217-219` ← `SwapScreen.kt:92` (`onPauseOrDispose`).

So the quote — and the vault address in it — freezes the instant the user taps into verify, and stays frozen through pairing, QR scan, the MPC ceremony and broadcast. Unbounded.

The asgard set rotates on churn. Funds sent to a retired vault have no recovery path.

## Proposed fix

[...]

PR #5659, opened 2026-08-21
fix(swap): show trading-paused copy instead of a quote timeout

### LINKED ISSUE #5656: [Fix] Swap shows timeout when THORChain trading is paused
When THORChain trading is paused, the swap form shows "Swap request timed out. Please try again". ETH<>RUNE and BTC<>RUNE still quote.

Expected: trading-paused copy (`swap_error_trading_halted`). Don't wait for a timeout.

This string is Android-only. iOS already maps pause/halt on the quote error. Android has the halted copy and a sign-time inbound gate, but quote still times out.

Could be related to #4749 / #5319.

aminsato:
Closes #5656

## Problem

When THORChain pauses trading, the swap form shows **"Swap request timed out. Please try again"** — copy that invites a retry which cannot succeed while trading is down. The correct string (`swap_error_trading_halted`) already exists and is already translated into all 10 locales; it just never gets selected.

Two things combine to produce the wrong message:

1. **The pause wording isn't recognised.** The nodes word a pause several ways — `"trading paused for this asset"`, `"the pool is paused"` — but `handleSwapException` only matched `"trading is halted"` / `"trading halted"`. Everything else fell through to `UnkownSwapError`.
2. **An unclassified error loses to a sibling's timeout.** `SwapQuoteManager.swapFailurePriority` ranks `UnkownSwapError` at tier 3 and `TimeOut` at tier 2, so when every provider fails, `minBy` surfaces the timeout from whichever provider stalled — not the halt from THORChain.

Classifying the halt correctly is not enough on its own: `TradingHalted` and `TimeOut` shared tier 2, and ties resolve by provider order, so the timeout still won whenever THORChain wasn't first in the eligible-provider list.

## Changes

**`SwapException.kt`** — extract `TRADING_HALT_MARKERS` (`trading is halted`, `trading halted`, `trading paused`, `_trading_paused`, `is paused`) and match it *ahead of every other class*. Order matters: the nodes report a pause alongside a fee clause (`"trading paused; not enough asset to pay for fees"`), which previously classified as an amount error and sent the user off to adjust an amount that can't help. This is the same marker set and the same check order iOS uses, so both stacks now classify identical node text identically.

**`SwapQuoteManager.kt`** — give `TradingHalted` its own tier, ahead of the transient group. A halt explains the whole outage; a sibling provider routing through the same paused protocol just stalls until its fetch times out, and that symptom shouldn't outrank the cause.

**`ThorChainQuoteSource.kt`** — a rapid quote that failed on a halt now short-circuits instead of firing the streaming fallback. A halt is a property of the pool or the chain, not of the swap interval, so the second round-trip can only be refused the same way — and that wasted request was part of what pushed the candidate past the 15s fetch timeout. This addresses the issue's "don't wait for a timeout" directly.

No string changes: `swap_error_trading_halted` is already present in `values/` and all nine locale files.

## Test plan

- [x] `./gradlew :data:testDebugUnitTest` — pass
- [x] `./gradlew :app:testDebugUnitTest` — pass
- [x] `./gradlew :data:lintDebug :app:lintDebug` — pass
- [x] ktfmt clean

Related: #4749 (introduced the `TradingHalted` class and the copy), #5319 (added the sign-time inbound halt gate). This PR closes the remaining gap at quote time.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection and classification of trading-halted conditions during swaps.
  - Trading halts now take priority over transient or unavailable-route errors.
  - Prevented unnecessary fallback quote requests when trading is halted.
  - Improved handling of additional trading-halt messages from swap providers.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -685,8 +685,8 @@ constructor(
     /**
      * Ranks a failed-quote [error] so the most actionable failure is surfaced when every provider
      * fails. Lower values win. Amount-related errors mean the pair is routable and the user can
-     * recover by adjusting the amount, so they rank above recoverable/transient errors, which in
-     * turn rank above generic "no route" fallbacks.
+     * recover by adjusting the amount, so they rank above an upstream trading halt, which in turn
+     * ranks above the remaining recoverable/transient errors and then generic "no route" fallbacks.
      */
     private fun swapFailurePriority(error: Throwable): Int =
         when (error) {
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5661, opened 2026-08-22
fix(security): enable passcode app-lock by default

johnnyluo:
## Description

The security passcode app-lock has been feature-flagged off by default while under test, only reachable via the hidden Advanced Settings toggle. This flips the default to enabled so the feature is available out of the box, without removing the toggle (users can still turn it off from Advanced Settings).

Change: `PasscodeConfig.isFeatureEnabled` now defaults to `true` (was `false`) in `data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeConfig.kt`. No other logic changes — the existing toggle, persistence, and "keep the entry visible once a passcode is set" behavior are untouched.

## Which feature is affected?
- [ ] Sending
- [ ] Swap

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works (existing `SettingsViewModelTest` / `SecretViewModelTest` cover the flag's visibility and toggle behavior; they mock the flag explicitly so they're unaffected by the default change)

## Additional context

Existing unit tests (`data:testDebugUnitTest`, `app:testDebugUnitTest`) pass unchanged.

## Agent Metadata
- **Agent**: Claude Code
- **Issue**: none
- **Confidence**: high
- **Needs human review for**: none — single default-value flip, well covered by existing tests

## Summary by CodeRabbit

* **New Features**
  * App lock with a passcode is now enabled by default.
  * Users can opt out of passcode protection if desired.
  * Updated descriptions clarify that the feature is generally available.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeConfig.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeConfig.kt
@@ -6,16 +6,16 @@ import javax.inject.Inject
 import kotlinx.coroutines.flow.Flow
 
 /**
- * DataStore-backed feature flag for the passcode app-lock. Default is `false` — the feature is
- * opt-in via the hidden Advanced Settings screen while it is still under test.
+ * DataStore-backed feature flag for the passcode app-lock. Default is `true` — the feature is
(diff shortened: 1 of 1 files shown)


aminsato (approved):
LGTM

PR #5665, opened 2026-08-22
feat(tron): claim TRX whose unfreeze period has matured

### LINKED ISSUE #5640: [Add] Implement TRON unfreeze withdraw/claim transaction
## Problem
On TRON, unfreezing TRX (`UnfreezeBalanceV2`) starts a 14-day unlock timer. Once that timer expires, the TRX is not automatically returned to the spendable balance — a separate `WithdrawExpireUnfreezeContract` transaction must be signed and broadcast to actually claim it.

The app currently implements freeze/unfreeze (`buildUnfreezeBalanceV2` in `data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TronHelper.kt:275`), and correctly tracks/displays pending withdrawal status. `TronPendingWithdrawalRow` in `app/src/main/java/com/vultisig/wallet/ui/screens/v2/defi/tron/TronDeFiPositionsScreen.kt:297-372` shows a "Ready to claim" badge once `expiryEpochMs` has passed, but the badge has no `onClick`/action wired to it — there's no way to actually broadcast a claim.

Notably, a comment in `data/src/main/kotlin/com/vultisig/wallet/data/api/models/TronResponseJson.kt:149-150` already documents the gap:
> `UnfreezeBalanceV2` moves it out of `frozenV2` but only `WithdrawExpireUnfreeze` returns it to the spendable `balance`

No `WithdrawExpireUnfreezeContract` builder exists anywhere in the repo (confirmed via grep).

Cross-platform note: this same gap exists in vultisig-ios (see [vultisig/vultisig-ios#5204](https://github.com/vultisig/vultisig-ios/issues/5204)) and vultisig-windows — filing as a parallel issue there.

## Expected Behavior
Once a pending withdrawal's unlock period has expired ("Ready to claim"), the user should be able to tap it (or a dedicated action) to sign and broadcast a `WithdrawExpireUnfreezeContract` transaction via TSS, after which the TRX becomes spendable and the entry clears from the pending list.

## Steps to Reproduce
1. Freeze some TRX for TRON Power on a Vultisig vault, then unfreeze it (`UnfreezeBalanceV2`).
2. Wait 14 days for the unlock period to expire (or wait for `expiryEpochMs` to pass).
3. Open the TRON DeFi positions screen — the pending withdrawal shows "Ready to claim", but tapping it does nothing.
4. TRX remains locked out of spendable balance; there is no way to withdraw it through the app.

## Solution
Add a `buildWithdrawExpireUnfreeze` contract builder in `TronHelper.kt` (mirroring the existing `buildUnfreezeBalanceV2` pattern) to construct a `WithdrawExpireUnfreezeContract`. Wire a claim action into `TronPendingWithdrawalRow` (`TronDeFiPositionsScreen.kt`) that's enabled only when `isClaimable` is true, routing through the standard send/sign flow so it goes through TSS like other Tron transactions.

### Must Do
- [ ] Add `WithdrawExpireUnfreezeContract` construction in `TronHelper.kt`
- [ ] Wire a claim action on "Ready to claim" pending withdrawal rows, gated on `isClaimable`
- [ ] Route the claim through the existing TSS sign/broadcast flow
- [ ] Refresh the pending withdrawals list after a successful claim

### Must NOT Do
- Don't change freeze/unfreeze logic or countdown/expiry handling
- Don't add extra dependencies

## Acceptance Criteria
- [ ] Lint/build passes
- [ ] A pending withdrawal past its unlock time can be claimed and signed via TSS, resulting in a broadcast `WithdrawExpireUnfreezeContract` transaction
- [ ] After claiming, the TRX balance updates and the entry no longer appears in the pending withdrawals list
- [ ] Non-expired pending withdrawals have no claim action available

aminsato:
Closes #5640

### Transaction hash
https://tronscan.org/transaction/dae62a1502769adc4a958468f23ba8f885d524863f175332e486c2d7eba4d904/overview

## Problem

`UnfreezeBalanceV2` only moves TRX out of `frozenV2`; the chain returns it to the spendable balance solely on a `WithdrawExpireUnfreezeContract` transaction, and no builder for that contract existed anywhere in the repo. The pending-withdrawal row already flipped to a "Ready to claim" badge once `expiryEpochMs` passed, but the badge had no action wired to it — so once the 14-day unlock elapsed the TRX was stranded, visible but unreachable.

## Changes

**Signing (`data/`)**
- `TronHelper.buildWithdrawExpireUnfreeze`, mirroring the existing `buildUnfreezeBalanceV2`. The contract carries only `ownerAddress`, so the payload's `toAmount` never reaches the signed bytes. Dispatch rejects a claim memo aimed at a foreign address, exactly as freeze/unfreeze already do.
- `TRON_WITHDRAW_EXPIRE_UNFREEZE_MEMO` as the routing signal, folded into `isTronStakingTransfer` so `TronFeeService` keeps treating it as the signalling channel it is rather than charging the 1 TRX memo fee for a memo the signed contract never carries (the invariant #5481 established).

**Claim flow (`app/`)**
- `TronDeFiPositionsViewModel.onClaimExpiredWithdrawals()` prices the fee through `FeeServiceComposite`, fetches the Tron `blockChainSpecific`, stages a `Transaction` and routes to `Route.VerifySend` — the same Verify → TSS → broadcast path every other Tron transaction takes.
- The contract sweeps *every* matured entry at once, so the amount shown on Verify is the sum of all expired rows.
- Guards: one claim in flight at a time; nothing staged when the claimable total is zero; and a pre-flight check that the spendable balance covers the fee — a vault with everything staked would otherwise burn a full keysign ceremony that could only fail at broadcast. Failures surface in a dialog instead of replacing the rendered positions.
- `TronDeFiSnapshotCache.clear()` on stage, so the resume-time reload re-reads the chain instead of repainting a row the transaction has already settled.
- A refresh that lands mid-claim no longer rebuilds the screen state from scratch, which used to re-enable the claim button while the claim was still in flight and dismiss an error dialog the user hadn't read (it also reset the position picker on every refresh).

**UI — mirrors iOS**

The pending-withdrawals section is rebuilt to match `TronDashboardView.pendingWithdrawalsCard` from vultisig/vultisig-ios#5207: one card carrying a header with the unfreezing total, a divider, one row per entry (amount, "Ready to claim" or the countdown, and a check-circle / hourglass), and a single **"Claim <sum> TRX"** button at the bottom when anything has matured.

The button is what changes the interaction. Because the contract sweeps every expired entry in one transaction, a per-row action implies claims can be made individually — and the pill it replaced was a small target that stayed inert for the ~4 s the fee and block lookups take, which read as broken. Row amounts now strip trailing zeros (`1 TRX`, `0.25 TRX`) the way iOS formats them, and the pill is demoted back to plain status text.

Supporting changes: `TronDeFiActionButton` extracted out of `TronFreezePositionCard` so Freeze / Unfreeze / Claim share one component, and one ticker for the whole card instead of one per row, so a row flipping to claimable and the button offering to claim it read the same clock.

**Localization**
- `tron_claim_screen_title` and `tron_defi_claim_button` added to all 10 locale files, each using the claim verb that locale already uses for `tron_defi_ready_to_claim` (de "abrufen", pt "resgatar", ru "получить", and so on).

## Test plan

- `./gradlew assembleDebug lintDebug :data:test :app:testDebugUnitTest :app:ktfmtCheck :data:ktfmtCheck` — all pass.
- The withdraw case was added to the existing `TronMemoFeeMatchesSignedMemoTest` table. It passing is itself the proof the new dispatch branch is reached: a fall-through to `buildCoinTransfer` would have written the routing memo into the signed transaction and failed the memo/fee parity assertion.
- Walked on an emulator against a vault with three matured entries: the card renders the `2.25 TRX` total and three ready rows, the claim button stages the transaction, and Verify shows "Claim TRON", `2.25 TRX`, self-addressed, with the `WITHDRAW_EXPIRE_UNFREEZE` memo row — matching the iOS Verify screen field for field.

## Cross-platform

The memo is a wire format shared with co-signers. vultisig-ios#5207 landed on 2026-08-20 with the identical `WITHDRAW_EXPIRE_UNFREEZE` constant and the same self-addressed-native-TRX dispatch guard, so iOS ↔ Android co-signing agrees. vultisig-windows has not implemented it yet, so a claim co-signed by a desktop peer will still fail keysign on mismatched hashes.

## Summary by CodeRabbit

* **New Features**
  * Added support for claiming expired TRON withdrawals with fee checks, balance validation, transaction verification, and progress feedback.
  * Added a pending-withdrawals card with totals, individual amounts, countdowns, and claim-ready status.
  * Added localized TRON claim labels in multiple languages.
* **Bug Fixes**
  * Improved recognition and display of expired TRON withdrawal transactions.
* **Tests**
  * Added coverage for claim state handling, concurrent claims, transaction routing, and fee consistency.


data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TronHelper.kt
```diff
@@ -301,6 +307,35 @@ class TronHelper(
...
+                .setOwnerAddress(keysignPayload.coin.address)
+                .build()
+
+        val txBuild =
+            Tron.Transaction.newBuilder()
+                .setWithdrawExpireUnfreeze(contract)
+                .setTimestamp(tronSpecific.timestamp.toLong())
+                .setBlockHeader(buildBlockHeader(tronSpecific))
+                .setExpiration(tronSpecific.expiration.toLong())
+                .setFeeLimit(tronSpecific.gasFeeEstimation.toLong())
```
rkokhatskyi: `buildWithdrawExpireUnfreeze` sets `feeLimit = gasFeeEstimation` (800000), but `fee_limit` is a smart-contract Energy field the real `wallet/withdrawexpireunfreeze` node RPC never populates for this contract, and iOS's `buildTronWithdrawExpireUnfreezeInput` (`Tron.swift:426`) already hardcodes it to 0 to match. Android should too, or the two platforms sign different bytes for the same claim

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TronDeFiSnapshotCache.kt
```diff
@@ -27,4 +27,13 @@ class TronDeFiSnapshotCache @Inject constructor() {
     fun write(tronAddress: String, snapshot: TronDeFiSnapshot) {
         snapshots[tronAddress] = snapshot
     }
+
+    /**
+     * Drops the snapshot for an address whose on-chain state is about to change under us — a signed
+     * freeze/unfreeze/claim — so the next open re-reads the chain instead of painting a position
+     * the transaction has already settled.
+     */
+    fun clear(tronAddress: String) {
```
rkokhatskyi, line 36: This comment claims a signed freeze/unfreeze also triggers `clear()`, but the only call site is `onClaimExpiredWithdrawals` at `TronDeFiPositionsViewModel.kt:487`; `onTronAction` routes freeze/unfreeze through `Route.Send` with no cache reference, so it should say only the claim clears it

rkokhatskyi (dismissed):

aminsato:
## ✅ All Review Comments Addressed

Thanks for the thorough review, @coderabbitai! I've addressed all three points in commit `388d1e5dc`:

### 1. **Boilerplate Extraction** ✅
Extracted the shared Tron transaction construction logic into a new `buildStakingTransaction()` helper function that:
- Takes a contract-setting lambda to apply specific contract types
- Centralizes `setTimestamp()`, `setBlockHeader()`, `setExpiration()`, and `setFeeLimit()` calls
- Reduces duplication across `buildFreezeBalanceV2`, `buildUnfreezeBalanceV2`, and `buildWithdrawExpireUnfreeze`

**File:** `data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TronHelper.kt`

### 2. **Test Coverage for onClaimExpiredWithdrawals** ✅
Added three comprehensive test cases covering the claim flow's safety guards:
- `claim sets isClaimingWithdrawal flag during processing` - Verifies concurrent-claim prevention
- `claim with no expired withdrawals does nothing` - Verifies zero-value claim guard
- `concurrent claim attempts are ignored while isClaimingWithdrawal is true` - Verifies concurrent attempt blocking

**File:** `app/src/test/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModelTest.kt`

### 3. **Preserve Claim-Related Fields During Refresh** ✅
Updated `publishLoaded()` to explicitly retain claim state across data refreshes:
- `isClaimingWithdrawal` flag preserved when copying Success state
- `claimError` dialog state preserved during pull-to-refresh
- Prevents claim button re-enabling or error dismissal mid-claim

**File:** `app/src/main/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModel.kt`

All changes compile cleanly and follow project conventions. Ready for next review!

aminsato:
## ✅ Follow-up Review Comments Addressed

Thanks for the additional review! I've addressed the two new CodeRabbit findings in commit `3b25fd04d`:

### 1. **WithdrawExpireUnfreeze Zero Fee Limit** ✅
Updated the `buildStakingTransaction()` helper to accept an optional `feeLimit` parameter:
- Freeze/Unfreeze operations continue using `tronSpecific.gasFeeEstimation` (default)
- WithdrawExpireUnfreeze now explicitly sets `feeLimit = 0L` since the operation doesn't consume chain resources
- Preserves the shared timestamp, block header, and expiration logic

**File:** `data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TronHelper.kt`

### 2. **Concurrent-Claim Test Validation** ✅
The test verifies the concurrent-claim guard by:
- Starting a claim with `tronApi.getAccount` suspended to keep the flow in flight
- Confirming `isClaimingWithdrawal` remains `true` after the second call
- This assertion proves the `claimJob?.isActive` guard is working (second call was ignored)

**File:** `app/src/test/java/com/vultisig/wallet/ui/models/defi/TronDeFiPositionsViewModelTest.kt`

All changes compile cleanly and are ready for final review!

aminsato:
## ✅ Review Comments Addressed (Clean Implementation)

All CodeRabbit review comments have been addressed in commit `335fc98c5`:

### 1. **Boilerplate Extraction in TronHelper.kt** ✅
- Created `buildStakingTransaction()` inline helper that:
  - Takes contract-setting lambda for flexibility
  - Applies common timestamp, block header, expiration configuration
  - Supports configurable feeLimit (default uses gasFeeEstimation)
- All three builders now use this helper, reducing ~80 lines of duplication

### 2. **WithdrawExpireUnfreeze Zero Fee Limit** ✅
- Updated `buildWithdrawExpireUnfreeze` to pass `feeLimit = 0L`
- Correctly reflects that this operation doesn't consume chain resources
- Freeze/Unfreeze operations continue using estimated gas fees

### 3. **State Preservation During Refresh** ✅
- Updated `publishLoaded()` to explicitly preserve:
  - `isClaimingWithdrawal` flag (claim in flight)
  - `claimError` (transient claim failure dialog)
- Prevents loss of state during pull-to-refresh or screen resume mid-claim

✅ All existing tests pass
✅ Code compiles cleanly with no warnings
✅ Changes are minimal and focused on the review feedback

johnnyluo (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/chains/helpers/TronHelper.kt
johnnyluo: Already fixed — buildWithdrawExpireUnfreeze now calls buildStakingTransaction(tronSpecific, feeLimit = 0L) { ... }, so feeLimit is hardcoded to 0 for this contract, matching iOS. This landed in 335fc98c5 before this comment was posted; thread is stale.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/TronDeFiSnapshotCache.kt
johnnyluo: Fixed in e0f63ec65 — the KDoc now says only the staged claim clears the cache; freeze/unfreeze route through Route.Send with no cache reference.

PR #5676, opened 2026-08-22
fix(defi): order the DeFi list by the amount each row shows

### LINKED ISSUE #5671: [Fix] DeFi Portfolio chain list does not sort by displayed balance [DeFi]
# [Fix] DeFi Portfolio chain list does not sort by displayed balance [DeFi]

## Problem
Live Android DeFi Portfolio (Dedo / fastandroid) is not in balance order:

Circle Yield $0.00, TerraClassic $4.24, Tron $1.75, THORChain $0.61, Solana $11.03, Maya $0.00, Terra $0.00.

Solana is the largest row and sits fifth.

`loadDeFiBalances` does call `sortByAccountsTotalFiatValue()` (`compareBy` on `accounts.calculateAccountsTotalFiatValue`, then chain name). That is not the order the screenshot shows. Either the sort key is not the DeFi total the row paints (Kamino / staking / Circle vs wallet-token fiat), or Circle is effectively pinned and the rest are not re-sorted after filter/merge.

Wallet tab uses the same helper against wallet balances and does order by amount. DeFi should use the number on the row.

## Solution
Sort `defiAccounts` by the same fiat the DeFi row displays, descending. Include Circle / `isDefiProvider` rows in that sort — a $0 Circle must not sit above a funded chain. Tie-break by chain name.

### Must Do
- [ ] Sort key = displayed DeFi fiat (Kamino, staking, LPs, Circle yield)
- [ ] Circle is not pinned to the top
- [ ] Re-sort after `retainLastKnownBalances` so a cached merge cannot restore enable-order
- [ ] Unit: Solana $11 outranks TerraClassic $4; $0 Circle is below both

### Must NOT Do
- Don't change Wallet-tab sort
- Don't sort by enable-order or chain enum order

## Acceptance Criteria
- [ ] Dedo's list would render Solana, TerraClassic, Tron, THORChain, then the zeros

Siblings: iOS / Windows (filed next)

aminsato:
## Problem

The DeFi Portfolio chain list was not in balance order. From the report:

Circle Yield $0.00, TerraClassic $4.24, Tron $1.75, THORChain $0.61, Solana $11.03, Maya $0.00, Terra $0.00 — the largest row fifth, behind a provider row showing nothing.

## Cause

Two things, both in `VaultAccountsViewModel`.

**The sort key was not the figure the row paints.** It sorted on `calculateAccountsTotalFiatValue`, the strict all-or-nothing fold that returns `null` as soon as *any one* account in the row is still resolving; `compareBy` ranks `null` first. The row itself renders `calculateAccountsPartialFiatValue` (the lenient sum, #4768). So a row could show `$0.00` off its one resolved account, score `null` on the account still in flight, and take the top of the list. That is the Circle row in the report — and it is why the ordering looked arbitrary rather than merely reversed.

**The sort ran before the merge.** `.map { sortByAccountsTotalFiatValue() }` was applied to the incoming emission, while `retainLastKnownBalances` restores last-known values later, inside `updateUiStateFromList`. A chain was therefore ranked on figures it never displayed, and the restored values landed in that stale order — a cached merge could leave a $20 row under a $5 one.

## Fix

- Dropped the pre-merge sort from the DeFi stream.
- New `sortByDisplayedFiatValue()` orders on `calculateAccountsPartialFiatValue` — the same sum `AddressToUiModelMapper` paints — descending, `nullsLast` so a row with nothing resolved sits at the bottom rather than the top, chain name as the tie-break.
- It runs **after** `retainLastKnownBalances`, so a cached merge is re-ordered by what the rows end up showing rather than restoring enable-order.
- Provider rows (Circle) sort on that same key: a `$0.00` position can no longer outrank a funded chain.
- Gated on `isDefi`, so the wallet tab keeps its own sort untouched.

## Tests

4 new tests in `VaultAccountsViewModelTest` (31 in the class):

- the reported list renders Solana, TerraClassic, Tron, THORChain, then the three zeros;
- a row with nothing resolved yet sits below the funded ones;
- rows restored from the cache are re-ordered by the value they show ($20 Terra above $5 Solana — the old key collapsed both onto the chain-name tie-break);
- a pin that the wallet list order is left alone.

The first three were each verified to fail when the ViewModel change is reverted. Full suite: 2159 `:app` unit tests, 0 failures.

## Test plan

On the DeFi tab of a vault funded across several protocols:

1. Home → **DeFi** toggle: rows read top to bottom in descending order of the amount printed on each row, every $0.00 row at the bottom, Circle included.
2. Pull to refresh: each row keeps the amount it was showing and stays in place while the figures refetch; the settled order matches step 1.
3. Cold start into the DeFi tab while balances are still loading: a chain with no figure yet sits at the bottom, then slides into position when its amount lands.
4. Manage Positions → switch a chain off and back on: it returns to the position its amount deserves, not to the end of the list.
5. Wallet tab is unchanged.

Closes #5671

## Summary by CodeRabbit

* **Bug Fixes**
  * DeFi account rows are now ordered by their displayed fiat values, with the highest values first.
  * Rows with unavailable values appear last, with chain name used to resolve ties.
  * Restored cached balances are now included before sorting, ensuring displayed order matches current values.
  * Wallet account ordering remains unchanged.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -17,6 +17,7 @@ import com.vultisig.wallet.data.models.CryptoConnectionType
 import com.vultisig.wallet.data.models.SigningLibType
 import com.vultisig.wallet.data.models.Vault
 import com.vultisig.wallet.data.models.VaultId
+import com.vultisig.wallet.data.models.calculateAccountsPartialFiatValue
 import com.vultisig.wallet.data.models.calculateAccountsTotalFiatValue
 import com.vultisig.wallet.data.models.calculateAddressesPartialFiatValue
 import com.vultisig.wallet.data.models.isFastVault
@@ -54,6 +55,7 @@ import com.vultisig.wallet.ui.utils.pushNotificationErrorUiText
 import com.vultisig.wallet.ui.utils.textAsFlow
 import com.vultisig.wallet.ui.utils.throttleLatest
 import dagger.hilt.android.lifecycle.HiltViewModel
+import java.math.BigDecimal
 import javax.inject.Inject
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.seconds
@@ -597,7 +599,6 @@ constructor(
                 combine(
                         accountsRepository
                             .loadDeFiAddresses(vaultId, readsNetwork)
-                            .map { addresses -> addresses.sortByAccountsTotalFiatValue() }
                             .catch { error ->
                                 Timber.e(
                                     error,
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5678, opened 2026-08-23
fix(swap): stop a THORChain halt from speaking for the whole pair

### LINKED ISSUE #5672: [Fix] Swap surfaces THORChain halt instead of using LiFi/SwapKit [Swap]
# [Fix] Swap surfaces THORChain halt instead of using LiFi/SwapKit [Swap]

## Problem
iOS ETH→SOL showed "Trading Halted"; retry then quoted. Same ranking lives on Android.

`fetchBestQuote` already fetches every candidate in parallel with a **15s** `QUOTE_FETCH_TIMEOUT_MS`. When all fail, `swapFailurePriority` ranks `TradingHalted = 2` **above** timeout/network `= 3` (#5656). That was correct for "don't show timeout when the node is paused" on a THOR-only wait. It is wrong for ETH→SOL: LiFi/SwapKit are eligible, a 15s cold miss is retryable, and halt is what the user sees.

#5656 also short-circuits THOR streaming on halt in `ThorChainQuoteSource` — keep that. The remaining gap is: halt from THOR must not beat a retryable aggregator miss, and aggregators should be retried once instead of painting halt.

## Solution
Same as iOS: halt is per-provider. If any aggregator is eligible, retry it once on timeout/network after a native halt. Surface halt only when no other provider can route.

Do not undo the streaming short-circuit.

### Must Do
- [ ] `swapFailurePriority`: halt does not outrank aggregator timeout when LiFi/SwapKit/Jupiter/1inch/Kyber were in the candidate set
- [ ] One automatic retry of those aggregators after native halt + transient miss
- [ ] Halt copy only when the pair has no non-native provider, or they all failed structurally
- [ ] Unit: THOR halt + LiFi timeout + LiFi retry success → quote, no halt form error

### Must NOT Do
- Don't re-introduce the streaming round-trip on halt (#5656)
- Don't hide halt on THOR-only pairs
- Don't change the sign-time inbound halt gate

## Acceptance Criteria
- [ ] ETH→SOL, THOR halted, LiFi quotes: form shows LiFi
- [ ] LiFi 15s timeout then success: automatic retry, no halt tooltip

Siblings: iOS / SDK+Windows (filed next)

aminsato:
Closes #5672

## Problem

When every provider in a candidate set fails, `SwapQuoteManager.fetchBestQuote` surfaces the most actionable failure rather than the first by provider order. #5656 placed `TradingHalted` above the transient tier, and on a THORChain-only pair that is right: every provider routes through the paused protocol, so a sibling's timeout is a symptom of the halt and "try again" sends the user round a loop the pause will never let them leave.

It does not hold on a pair an aggregator can also route. ETH → SOL carries THORChain, LI.FI and SwapKit. A halt there describes THORChain and nothing else, while LI.FI missing a cold 15s window says nothing about whether the pair routes at all — yet the halt outranked it, so the form painted "Trading is halted" and a manual retry then quoted immediately. Same defect iOS saw.

## Solution

Halt becomes per-provider, in the all-failed branch only:

- **One retry.** A halt from a native protocol (THORChain / MayaChain) plus a transient miss from an eligible aggregator gives exactly those aggregators one more fetch window before any copy is chosen. Bounded to a single extra pass.
- **Demotion.** If the second pass is also empty, a new `PRIORITY_HALT_DEMOTED` drops the halt below the transient tier so the "try again" that is literally true decides the message.
- **Unchanged where it was right.** With no aggregator eligible, or with every eligible one answering with a verdict of its own (no route, not supported, amount), the halt keeps the tier #5656 gave it.

Two supporting changes: `isAggregator` is an exhaustive `when` so a new provider has to be placed on one side rather than silently inheriting a halt's meaning, and failures now travel paired with the candidate that produced them — the ranking cannot tell whose halt it is looking at from a bare throwable. That re-pairing is most of the diff; the fetch body itself is only reindented.

The extra window is reached only when the first pass produced no quote at all, so a run that already holds a route never pays for it. Worst-case latency on a halted all-fail run goes 15s → 30s.

Out of scope, deliberately untouched: the streaming short-circuit in `ThorChainQuoteSource` (#5656) and the sign-time inbound halt gate in `SwapInboundHaltPreflight`.

## Test plan

`./gradlew :app:testDebugUnitTest` — **2174 tests, 0 failures** (8 skipped, pre-existing). `:app:ktfmtCheck` clean, no new compiler warnings.

Five new tests in `SwapQuoteManagerTest`. The two behaviour ones were verified to fail on revert of the source change:

| Test | Pins |
|---|---|
| `retries an aggregator once after a native halt and takes its quote` | THOR halt + LI.FI timeout + LI.FI retry success → a LI.FI quote, exactly 2 LI.FI attempts (**fails on revert**) |
| `surfaces the aggregator timeout, not the halt, when the retry misses too` | Second window also empty → `SwapException.TimeOut`, not the halt (**fails on revert**) |
| `keeps the halt when no aggregator is eligible` | MAYA timeout + THOR halt → halt still surfaces; #5656 intact on THOR-only pairs |
| `keeps the halt when every eligible aggregator answered with a verdict` | LI.FI "no route" + THOR halt → halt surfaces, no retry fired |
| `does not retry an aggregator when no native provider is halted` | THOR "no route" + LI.FI timeout → one LI.FI attempt; the second window is the halt's price, not every failed run's |

Manual: the fixed path needs a live THORChain halt, which is not reproducible on demand. Verified no regression on the healthy path — ETH → BTC quotes from THORChain, ETH → SOL quotes from SwapKit (NEAR), both with unchanged provider, fee and timing.

Siblings: iOS / SDK+Windows

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved swap quote recovery when providers experience temporary failures.
  - Retries eligible aggregator quotes after a native provider reports a trading halt.
  - Returns successful retry quotes when available, while preserving accurate timeout and trading-halt errors.
  - Refined failure prioritization to provide more relevant swap error messages.

- **Tests**
  - Added coverage for provider retries, repeated timeouts, trading halts, and cases where retries are not applicable.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -445,64 +445,115 @@ constructor(
             throw SwapException.SwapIsNotSupported("Swap is not supported for this pair")
         }
 
-        val results: List<Result<BestQuote>> = coroutineScope {
-            candidates
+        // Local so the retry pass below can rerun exactly the same fetch without threading every
+        // request parameter through a helper. Outcomes come back paired with the candidate that
+        // produced them, and in the order they were asked for, so a failure stays attributable to
+        // its provider once the results are flattened.
+        suspend fun attempt(
+            subset: List<QuoteCandidate>
+        ): List<Pair<QuoteCandidate, Result<BestQuote>>> = coroutineScope {
+            subset
                 .map { candidate ->
                     async {
-                        runCatching {
-                                withTimeout(QUOTE_FETCH_TIMEOUT_MS) {
-                                    BestQuote(
-                                        candidate = candidate,
-                                        result =
-                                            fetchQuote(
-                                                provider = candidate.provider,
-                                                src = src,
-                                                dst = dst,
-                                                srcToken = srcToken,
-                                                dstToken = dstToken,
-                                                srcTokenValue = srcTokenValue,
-                                                tokenValue = tokenValue,
-                                                currency = currency,
-                                                vultBPSDiscount = candidate.vultBPSDiscount,
-                                                referral = candidate.referral,
-                                                amount = amount,
-                                                slippageBps = slippageBps,
-                                                externalRecipient = externalRecipient,
-                                            ),
+                        candidate to
+                            runCatching {
+                                    withTimeout(QUOTE_FETCH_TIMEOUT_MS) {
+                                        BestQuote(
+                                            candidate = candidate,
+                                            result =
+                                                fetchQuote(
+                                                    provider = candidate.provider,
+                                                    src = src,
+                                                    dst = dst,
+                                                    srcToken = srcToken,
+                                                    dstToken = dstToken,
+                                                    srcTokenValue = srcTokenValue,
+                                                    tokenValue = tokenValue,
+                                                    currency = currency,
+                                                    vultBPSDiscount = candidate.vultBPSDiscount,
+                                                    referral = candidate.referral,
+                                                    amount = amount,
+                                                    slippageBps = slippageBps,
+                                                    externalRecipient = externalRecipient,
+                                                ),
+                                        )
+                                    }
+                                }
+                                .onFailure { e ->
+                                    // TimeoutCancellationException extends CancellationException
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #5681, opened 2026-08-23
feat(banners): let a dismissed promo banner stay gone for good

### LINKED ISSUE #5669: [Add] Per-banner dismiss policy: never-return vs return [Home banners]
# [Add] Per-banner dismiss policy: never-return vs return [Home banners]

## Problem
Ops needs to assign, per home promo banner, whether a dismiss is forever or the banner comes back. Android today only has TTL in `PromoBanner`:

- UpgradeVaultDkls 15d
- FollowXVultisig 15d
- BuyVultSwap 7d

Every dismiss returns. There is no never-return rule, and the assignment is code-only.

## Solution
Add a permanent dismiss rule next to TTL on `PromoBanner`. Keep the assignment on the enum (one place), not in the pager. When the campaign admin (notification #34) serves a policy, honor it; until then keep today's hardcoded TTLs.

Visual restyle is a different ticket: #5642.

### Must Do
- [ ] `PromoBanner` (or its dismissal rule) supports never-return as well as TTL
- [ ] Permanent dismissals persist globally and never expire
- [ ] Honor remote policy from notification #34 when present; hardcoded fallback otherwise
- [ ] Unit coverage: permanent stays hidden, TTL still expires, banners stay independent

### Must NOT Do
- Don't restyle the banners here (#5642)
- Don't change existing TTL lengths unless product assigns new values on this ticket
- Don't vault-scope dismissal (already global)

### Out of Scope
- The admin tab itself (notification #34)
- Adding Rujira / Backup / Referral / Kamino banners (#5642)

## Acceptance Criteria
- [ ] `./gradlew :data:testDebugUnitTest` succeeds
- [ ] A banner assigned never-return stays hidden after dismiss across vaults and launches
- [ ] TTL banners still resurface after their interval

Siblings: iOS (next) · Windows (next) · Admin tab https://github.com/vultisig/notification/issues/34

aminsato:
## Problem

Every home promo banner dismissal expired. `PromoBanner` carried a single `ttl` knob (upgrade 15d, Follow X 15d, Buy \$VULT 7d), so the only question the store could answer was whether the TTL window had elapsed — there was no way to run a campaign whose dismiss is final.

## Change

- `ttl: Duration` becomes `dismissPolicy: DismissPolicy` — `Ttl(duration)` or `Permanent` — still assigned on the enum, so the per-banner rule stays in one place instead of leaking into the pager.
- `isDismissed` branches on the policy with an exhaustive `when`, so a third kind can't be added silently.
- The served-policy seam is `isDismissed(banner, policy)`; the single-arg overload supplies the banner's own assignment, so call sites need no change when a policy arrives from elsewhere. An overload rather than a default argument, because mockk resolves default arguments outside the proxy and would silently unstub existing callers.

Existing TTLs are carried over unchanged, per the ticket's "don't change existing TTL lengths".

## Notes

**No storage change was required.** DataStore already persists `banner_dismissed_at/<id>` indefinitely; what made a banner return was purely the in-memory comparison. So a permanent dismissal survives relaunch and vault switching with no storage work, and `dismiss()` writes the same timestamp under either policy — a policy can be reassigned later without rewriting what is already stored.

This matches the shape Windows shipped in vultisig/vultisig-windows#4718 (`{ ttl } | { permanent }` union, optional policy input, TTLs untouched). iOS #5238 has no PR yet; its `.session` rule has no Android counterpart, since the banner it exists for (backup reminder) isn't shipped here — so Android's policy is two cases, not three.

## Not included

Two ticket items are blocked on decisions outside this repo:

- **Honor remote policy from vultisig/notification#34** — that issue was closed with "no changes on the notification service" (see vultisig/notification#33), so there is no endpoint or payload shape to honor. The `policy` argument is the seam to wire one in once a contract exists.
- **Which banners should be never-return** — product has not assigned any, and the ticket forbids changing TTLs without an assignment. The mechanism ships here; the assignment is a one-line change per banner.

## Testing

`./gradlew :data:testDebugUnitTest` — 3119 tests green. `./gradlew :app:testDebugUnitTest` — 2169 green (one unrelated flake, `EnterVaultInfoViewModelTest`, failing on "uncaught exceptions before the test started" from another test's scope; passes 10/10 in isolation).

Three tests added: a permanent dismissal outlasting a 3650-day gap that the TTL policy lapses at, a permanent-policy banner that was never dismissed still showing, and a permanent banner leaving its neighbours on their TTL. Verified they bite — flipping `Permanent -> true` to `false` fails two of them.

Closes #5669

## Summary by CodeRabbit

* **New Features**
  * Promo banners now support either time-limited or permanent dismissal.
  * Permanent dismissals remain hidden indefinitely, while time-limited banners reappear after their configured period.
  * Banner dismissal behavior is independently applied to each banner.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -96,9 +96,9 @@ internal data class VaultAccountsUiModel(
     val accounts: List<AccountUiModel> = emptyList(),
     val defiAccounts: List<AccountUiModel> = emptyList(),
     val searchTextFieldState: TextFieldState = TextFieldState(),
-    // Per-banner visibility, each gated on a global, TTL-based dismissal (#5064). The upgrade
-    // banner
-    // additionally requires the vault to be GG20 (migration-eligible).
+    // Per-banner visibility, each gated on a global dismissal whose lifetime is the banner's own
+    // policy (#5064). The upgrade banner additionally requires the vault to be GG20
+    // (migration-eligible).
     val showUpgradeBanner: Boolean = false,
     val showFollowXBanner: Boolean = false,
     val showBuyVultBanner: Boolean = false,
@@ -435,10 +435,9 @@ constructor(
 
     fun dismissFollowXBanner() = dismissPromoBanner(PromoBanner.FollowXVultisig)
 
-    // Global, TTL-based dismissal: the banner stays hidden across vaults until its TTL elapses,
-    // then
-    // becomes eligible again (#5064). Writing the timestamp re-emits the dismissal flow, so the
-    // banner hides reactively without a session flag.
+    // Global dismissal: the banner stays hidden across vaults for as long as its policy says —
+    // until a TTL elapses, or for good (#5064). Writing the timestamp re-emits the dismissal flow,
+    // so the banner hides reactively without a session flag.
     private fun dismissPromoBanner(banner: PromoBanner) {
         viewModelScope.safeLaunch { promoBannerDismissalRepository.dismiss(banner) }
     }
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5696, opened 2026-08-24
fix(ton): stop a rejecting contract from absorbing a transfer

### LINKED ISSUE #5686: [Fix] TON: bounceability decided by address prefix — UQ/raw destinations can absorb funds [TON / S1]
**Problem:** Bounceability of a TON destination is decided by whether the address starts with "E". Non-bounceable (UQ) and raw (`0:`) addresses always go out non-bounceable — a rejecting contract absorbs the transfer instead of refunding it.
**Fix:** Parse the address flag properly instead of string-matching the prefix; force bounceable for swap deposit addresses.
**Receipts:** `data/.../BlockChainSpecificRepository.kt:596-602`.
Sibling of vultisig-sdk#2181 (S1 item 2).

aminsato:
Closes #5686

## Problem

Bounceability of a TON destination was decided by `dstAddress.startsWith("E")`. Two kinds of transfer went out non-bounceable that should not have — and a non-bounceable message a contract rejects is absorbed, not refunded.

**Swap deposits, unconditionally.** `SwapTransactionBuilder` calls `getSpecificAndUtxo` without a `dstAddress`, so the null check returned `false` before any flag was read. TON SwapKit is not signed from payload bytes the way its siblings are — `SigningHelper.kt:201` routes `TX_TYPE_TON` through the plain `TonHelper` transfer path, which reads exactly this flag. So every TON swap reached its router or escrow contract non-bounceable. An expired quote or a paused pool absorbed the deposit instead of refunding it.

**Raw `0:hex` destinations.** They carry no bounceability tag at all, and fell through the prefix check to the unsafe default.

## Fix

`TonAddressFlags` reads the tag out of the address bytes — `tag(1) | workchain(1) | hash(32) | crc16(2)`, `0x11` bounceable / `0x51` non-bounceable, testnet bit masked — in both base64url and standard base64 spellings. It reports a third state, `UNSPECIFIED`, for the raw form, which is the case a prefix check cannot express: raw declares neither, so the caller decides. Pure JVM, no WalletCore, so it unit-tests on the host.

In `BlockChainSpecificRepository`:

- `isSwap` forces bounceable, so a deposit contract that rejects the message refunds it.
- Raw destinations default to bounceable — the safe side for a contract.
- An explicit `UQ…` still opts out, and an undeployed destination is still non-bounceable; a bounceable message there would only bounce back and fail the transfer.

Sibling of vultisig/vultisig-sdk#2181 (S1 item 2). iOS carries the identical defect at `BlockChainService.swift:1038`, so there is no iOS behaviour to match here.

## Not in scope

Overriding an explicit `UQ…` when the destination is a contract. Honouring the flag is what the TON standard says it means; going further needs an on-chain code-hash lookup and is a separate change.

## Test plan

12 new tests, all verified to actually run (`tests=` in the XML reports, not just a green build).

- `TonAddressFlagsTest` — 7 tests. Every vector is the same STON.fi router account re-encoded with a different tag or spelling, so the reader is pinned on the tag alone: EQ, UQ, raw, base64url vs standard base64, both testnet tags, whitespace, and six unclassifiable inputs that must not claim a flag they never declared.
- `BlockChainSpecificRepositoryImplTest` — 5 tests: swap-with-no-destination, EQ, UQ, raw, uninitialized.

Reverted the fix and re-ran: **"raw spelling is bounceable"** and **"swap deposit is bounceable"** fail against the old code. The other three pass on both, as regression locks.

Full `:data` suite: 3136 tests, 0 failures. `:app:compileDebugKotlin` clean. ktfmt applied.

### On device

Verified on TON mainnet as a controlled pair — same wallet, same account as the destination, same 0.01 TON, 13 minutes apart, only the spelling of the destination differing. Out-message `bounce` flag read from the indexer:

| tx | destination spelling | `bounce` |
|---|---|---|
| [`11aaeb52…`](https://tonviewer.com/transaction/11aaeb5234ced096b962a6b9f86d23be3bae50fcaf9ea52bbc2d52061e616014) | `UQCmBAnv…` | `false` |
| [`10e88ea6…`](https://tonviewer.com/transaction/10e88ea6f166e06dab4f3be512405affb2f6d0d2aac0c07ef490f08d470adf6b) | `0:a60409ef…` | **`true`** |

Both report `orig_status: active`, so the destination was deployed in each and the uninitialized branch is ruled out. The first is the regression lock — old and new code agree that an explicit `UQ…` opts out. The second is the change: on the old code `startsWith("E")` is false for a raw address, so it could not have produced `bounce = true`.

The swap leg has no on-device transaction — SwapKit returns `tokenPriceUnavailable` for `TON.TON`, and `SwapProviderTable.kt:177` makes SwapKit the only provider for the chain. It rests on the unit test verified to fail against the pre-fix code.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/ton/TonAddressFlags.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/ton/TonAddressFlags.kt
@@ -1,0 +1,57 @@
+package com.vultisig.wallet.data.crypto.ton
+
+import java.util.Base64
+
+/** Bounceability a TON destination address declares, or [UNSPECIFIED] when it declares none. */
+enum class TonBounceability {
+    /** User-friendly `EQ…` form — a message the destination rejects is refunded. */
+    BOUNCEABLE,
+    /** User-friendly `UQ…` form — a message the destination rejects is absorbed. */
+    NON_BOUNCEABLE,
+    /** Raw `0:hex` form, or anything this reader cannot classify. Carries no flag at all. */
+    UNSPECIFIED,
+}
+
+/**
+ * Reads the bounceable tag out of a TON address rather than inferring it from the leading
+ * character.
+ *
+ * A user-friendly address is base64 of `tag(1) | workchain(1) | hash(32) | crc16(2)`, where the tag
+ * is `0x11` bounceable or `0x51` non-bounceable, optionally OR'd with the `0x80` testnet-only bit.
+ * The raw `0:hex` spelling of the same account carries no tag, which is the case a prefix check
+ * cannot express: it is neither bounceable nor non-bounceable, and the caller has to decide.
+ *
+ * Structural validity beyond the length and the tag (the CRC in particular) is not re-checked here
+ * — a destination reaches this reader only after `ChainAccountAddressRepository.isValid`, and an
+ * unrecognized input degrades to [TonBounceability.UNSPECIFIED] rather than to a flag it did not
+ * declare. Pure JVM (no WalletCore) so it stays unit testable.
+ */
+object TonAddressFlags {
+
+    private const val USER_FRIENDLY_LENGTH = 48
+    private const val USER_FRIENDLY_SIZE = 36
+    private const val TAG_MASK = 0x7f
+    private const val TAG_BOUNCEABLE = 0x11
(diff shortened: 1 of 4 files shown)


johnnyluo (approved):
LGTM

Issue #5435, opened 2026-07-28
Noble send: memo over 256 chars is signed then rejected at broadcast (memo too large)


aminsato:
# [Fix] Enforce Cosmos memo length limit before signing [Send / Noble]

## Problem
Sending on Noble (USDC) with a memo longer than the chain's `max_memo_characters` limit (256 on Noble) passes the entire Send → Verify → Keysign flow with no warning, then fails at broadcast with `memo too large`. The user completes a full TSS keysign with all devices before learning the memo was never going to be accepted. The app performs no length check on the memo field, so any input up to the field's capacity is signed and broadcast.

## Expected Behavior
The memo length limit is enforced *before* signing, so keysign is never started for a memo that will be rejected:
- The Send form rejects (or caps/truncates input at) a memo over the chain's limit, with an inline error naming the limit and the current length.
- The Continue button is disabled while the memo is over the limit.
- The limit is per-chain rather than a single hardcoded number — Cosmos chains expose `max_memo_characters` in their auth params (256 on Noble), and other families have their own ceilings.

## Steps to Reproduce
1. Open a vault and choose Send on Noble (USDC).
2. Paste a memo longer than 256 characters (in the captured case, 26,342 characters).
3. Continue to Verify — no warning is shown.
4. Complete keysign with all devices — signing succeeds.
5. Broadcast fails with "Transaction rejected by the network: maximum number of characters is 256 but received 26342 characters: memo too large", landing the user on "Transaction failed".

### Must Do
- [ ] Add memo-length validation to the Send form that blocks continuing (disables Continue) and shows an inline error naming the limit and the current length when the memo exceeds the chain's cap.
- [ ] Source the limit per-chain from the chain's memo ceiling (Cosmos `max_memo_characters`, 256 on Noble) rather than a single hardcoded number, counting by the same rule the node enforces (characters per `Params.max_memo_characters`, not assuming ASCII/UTF-8 bytes).
- [ ] Confirm the same validation covers the other Cosmos chains sharing this send path (Cosmos Hub, Osmosis, Kujira, THORChain, dYdX, Terra, etc.).

### Must NOT Do
- Don't change unrelated code.
- Don't alter the broadcast error-parsing path (`CosmosBroadcastException` / `CosmosApi`) as the primary fix — the enforcement must happen before keysign, not only surface a better broadcast error.
- Don't hardcode 256 as a global constant that ignores per-chain differences.

## Acceptance Criteria
[...]

PR #5698, opened 2026-08-24
fix(solana): stop a token account from passing as a send recipient

### LINKED ISSUE #5689: [Fix] Solana: ATA/PDA pasted as recipient passes validation — funds unrecoverable [SOL / S1]
**Problem:** A token account (ATA) or program address (PDA) pasted as a Solana recipient passes address validation. An SPL send then derives an ATA-of-an-ATA nobody controls — tokens unrecoverable.
**Fix:** Use the existing private `isOnCurve` (used only for PDA derivation today) inside recipient validation; warn or block off-curve recipients.
**Receipts:** `data/.../ChainAccountAddressRepository.kt:130-141` (bare validate); `data/.../crypto/SolanaProgramDerivedAddress.kt:83` (isOnCurve exists, private).
Sibling of vultisig-sdk#2182 (S1 item 1).

aminsato:
Closes #5689

## The problem

Solana address validation is WalletCore's base58 length check and nothing more, so an associated token account or a program address passes it. Paste one as a recipient and the SPL send path does this:

- `getTokenAccountsByOwner(ata, mint)` returns nothing — a token account owns no token accounts
- the derivation fallback computes `ATA(ata, mint)` and finds no such account on chain
- `toAddressPubKey` comes back null, so `SolanaHelper` takes the `CreateAndTransferToken` branch and **creates** the ATA-of-an-ATA, then transfers into it

That account's owner is the first token account, which holds no key and whose program cannot sign for it. The tokens are unrecoverable. S1 in the Solana money-path hardening epic (vultisig/vultisig-sdk#2182, item 1).

Verified offline against a real derivation:

```
owner  9WzDXwBbmkg8ZTbNMqUxvQRAyrZzDsGYdLVL9zYtAWWM   32 bytes, on-curve
ATA    FGETo8T8wMcN2wCjav8VK6eh3dLk63evNDPxzLSJra8B   32 bytes, off-curve  ← passes validation today
```

## The check

Every program-derived address is off the ed25519 curve by construction — that is what lets the owning program sign for it, and precisely why no user can. `SolanaProgramDerivedAddress` already computes that for derivation, so the guard reuses it: `isWalletAddress` decodes base58, requires 32 bytes, and asks whether they decompress to a point.

Pure Kotlin, no JNI, so it is covered by ordinary unit tests.

## Why it is not part of `isValid`

`SearchTokenUseCase` validates **mint** addresses through the same `isValid`, and mints are routinely program-derived. All three Kamino share mints in `KaminoVaultRegistry` are off-curve:

```
7D8C5pDFxug58L9zkwK7bCiDg4kD4AygzbcZUmf5usHS
DgHN3q3dSYAchNX7V3D4aYiTWMx8RHTgHbfPiwiqBkE9
FiM4VQdXXnTXL7GgChryf9zHNG9cmvKECwf34L2y3CkN
```

Folding the rule into `isValid` would have made exactly the tokens #5588 shipped unsearchable. So `validateRecipient` sits beside it and returns which of the two failures happened — `InvalidForChain` or `NotAWalletAddress` — leaving `isValid` untouched. There is a test pinning that distinction so the shortcut is not taken later.

## What changed

- `SolanaProgramDerivedAddress.isWalletAddress` — the curve check, promoted from the private derivation path
- `ChainAccountAddressRepository.validateRecipient` → `RecipientValidity` (`Valid` / `InvalidForChain` / `NotAWalletAddress`); `isValid` unchanged
- Wired into the three surfaces where a human names a recipient: the send form (`AddressManager` — `invalidAddress: StateFlow<Boolean>` became `addressError: StateFlow<RecipientValidity?>` so the reason survives to the UI), the swap external recipient, and the address book entry screen
- A token account is now rejected outright instead of being sent round the name resolver, which has nothing to find for a well-formed address
- New shared string `error_recipient_not_a_wallet_address`, translated into all 10 locales

It **blocks** rather than warns: a warning tapped through still loses the funds, and no sibling platform has shipped a reference behaviour to match yet.

## Scope boundary

Off-curve is not a general "dangerous recipient" test. The Token Program, ATA Program and System Program addresses are all **on**-curve, so the burn/program-address denylist (sdk#2182 item 2) is separate work and is not filed on Android yet.

## On device

A Kamino share mint — a program-derived address — pasted into Send on Solana. It is a well-formed
address that base58 validation accepts, and before this change it went straight through to the
amount step.

Send screen rejecting a program address with the not-a-wallet error

## Test plan

10 new tests; full suite green at **5329 tests, 0 failures**, lint clean.

- `SolanaProgramDerivedAddressTest` — the three ATAs it already pins are rejected; a real wallet and a program id are accepted; non-32-byte input is rejected; the Kamino share mints are rejected, documenting why the guard cannot live in `isValid`
- `ChainAccountAddressRepositoryRecipientTest` — the ordering: chain validation reported as itself, the curve rule applied after it and only to Solana
- `AddressManagerTest` — a token account sets `NotAWalletAddress`, resolves nothing, emits no validation and never reaches the resolver; a wallet address after it clears the error
- `SwapFormViewModelTest` — the external recipient gets the specific message, not the generic one

The two behaviour tests were confirmed to fail when the guard and the reject branch are reverted.

Manual, on a vault with a Solana account: pasting `FGETo8T8wMcN2wCjav8VK6eh3dLk63evNDPxzLSJra8B` as a recipient in Send, in the swap external-recipient field and in the address book is refused with the new message, while `9WzDXwBbmkg8ZTbNMqUxvQRAyrZzDsGYdLVL9zYtAWWM` is accepted as before. Custom-token search for the Kamino share mint above still resolves — that is the regression this shape exists to avoid. EVM, Bitcoin and XRP sends are untouched.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved recipient address validation during sends, swaps, and address entry.
  - Displays distinct errors for addresses invalid on the selected chain and valid-format addresses that are not wallet addresses.
  - Prevents token accounts or program addresses from being treated as wallet recipients.
  - Prevents outdated validation results from affecting updated recipient addresses.
  - Supports recovery when users replace an invalid recipient with a valid wallet address.

- **Localization**
  - Added translations for the new non-wallet recipient error in supported languages.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
@@ -8,6 +8,7 @@ import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.models.Coin
 import com.vultisig.wallet.data.repositories.AddressParserRepository
 import com.vultisig.wallet.data.repositories.ChainAccountAddressRepository
+import com.vultisig.wallet.data.repositories.RecipientValidity
 import com.vultisig.wallet.data.usecases.RequestAddressBookEntryUseCase
 import com.vultisig.wallet.data.utils.safeLaunch
 import com.vultisig.wallet.ui.utils.asAddressInput
@@ -60,13 +61,12 @@ internal class AddressManager(
     private val _isDstAddressComplete = MutableStateFlow(false)
     val isDstAddressComplete: StateFlow<Boolean> = _isDstAddressComplete.asStateFlow()
 
-    // True when a non-empty recipient failed chain validation and couldn't resolve to a valid
-    // address. Drives the inline "not a valid address for this chain" error; false once the input
-    // is valid or empty. While name resolution is in flight the flag holds its previous value, so a
+    // Why a non-empty recipient was rejected, or null once the input is accepted or empty. Drives
+    // the inline recipient error. While name resolution is in flight the value is held, so a
     // standing error stays put until the new input resolves instead of blinking off and back on.
     // Chain-general: every send chain validates through the same path below.
-    private val _invalidAddress = MutableStateFlow(false)
-    val invalidAddress: StateFlow<Boolean> = _invalidAddress.asStateFlow()
+    private val _addressError = MutableStateFlow<RecipientValidity?>(null)
+    val addressError: StateFlow<RecipientValidity?> = _addressError.asStateFlow()
 
     private val _onAddressValidated = MutableSharedFlow<Unit>(extraBufferCapacity = 1)
     val onAddressValidated: SharedFlow<Unit> = _onAddressValidated.asSharedFlow()
@@ -144,27 +144,36 @@ internal class AddressManager(
             }
         }
 
-        when {
-            chainAccountAddressRepository.isValid(chain, addressStr) -> {
+        if (addressStr.isEmpty()) {
+            _resolvedDstAddress.value = null
+            _dstAddressLabel.value = null
+            _addressError.value = null
+            return
+        }
+
+        when (chainAccountAddressRepository.validateRecipient(chain, addressStr)) {
+            RecipientValidity.Valid -> {
                 // Only clear ENS label if the user typed a new raw address,
                 // not when we programmatically set the field to the resolved address.
                 if (addressStr != _resolvedDstAddress.value) {
                     _dstAddressLabel.value = null
                 }
                 _resolvedDstAddress.value = addressStr
-                _invalidAddress.value = false
+                _addressError.value = null
                 _onAddressValidated.tryEmit(Unit)
             }
-            addressStr.isNotEmpty() -> {
+            // A token account or program address is a well-formed address, so there is no name for
+            // the resolver to find — reject it here instead of sending it round that path.
+            RecipientValidity.NotAWalletAddress -> {
+                _resolvedDstAddress.value = null
+                _dstAddressLabel.value = null
+                _addressError.value = RecipientValidity.NotAWalletAddress
+            }
+            RecipientValidity.InvalidForChain -> {
                 // Clear stale resolved address while async resolution is in-flight
                 _resolvedDstAddress.value = null
                 _dstAddressLabel.value = null
                 tryResolveName(addressStr, token)
-            }
-            else -> {
-                _resolvedDstAddress.value = null
-                _dstAddressLabel.value = null
-                _invalidAddress.value = false
             }
         }
     }
@@ -196,7 +205,7 @@ internal class AddressManager(
             addressFieldState.setTextAndPlaceCursorAtEnd(decoded.classicAddress)
         }
         _resolvedDstAddress.value = decoded.classicAddress
-        _invalidAddress.value = false
+        _addressError.value = null
         // Surface the pasted X-address as the label so Verify/Done show what the user entered.
         _dstAddressLabel.value = originalInput
         _onAddressValidated.tryEmit(Unit)
@@ -208,16 +217,17 @@ internal class AddressManager(
             val resolved = addressParserRepository.resolveName(addressStr, chain)
             // Ignore stale result if user changed input while resolving
             if (addressFieldState.text.asAddressInput() != addressStr) return
-            if (chainAccountAddressRepository.isValid(chain, resolved)) {
+            val validity = chainAccountAddressRepository.validateRecipient(chain, resolved)
+            if (validity == RecipientValidity.Valid) {
                 _dstAddressLabel.value = addressStr
                 _resolvedDstAddress.value = resolved
-                _invalidAddress.value = false
+                _addressError.value = null
                 addressFieldState.setTextAndPlaceCursorAtEnd(resolved)
                 _onAddressValidated.tryEmit(Unit)
             } else {
                 _resolvedDstAddress.value = null
                 _dstAddressLabel.value = null
-                _invalidAddress.value = true
+                _addressError.value = validity
             }
         } catch (e: CancellationException) {
             throw e
@@ -227,7 +237,7 @@ internal class AddressManager(
             Timber.w(e, "Failed to resolve address %s on %s", addressStr, chain)
             _resolvedDstAddress.value = null
             _dstAddressLabel.value = null
-            _invalidAddress.value = true
+            _addressError.value = RecipientValidity.InvalidForChain
         }
     }
 }
--- a/app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
@@ -20,6 +20,7 @@ import com.vultisig.wallet.data.models.settings.AppCurrency
 import com.vultisig.wallet.data.repositories.BlockChainSpecificAndUtxo
 import com.vultisig.wallet.data.repositories.BlockChainSpecificRepository
 import com.vultisig.wallet.data.repositories.ChainAccountAddressRepository
+import com.vultisig.wallet.data.repositories.RecipientValidity
 import com.vultisig.wallet.data.repositories.TransactionRepository
 import com.vultisig.wallet.data.usecases.GasFeeToEstimatedFeeUseCase
 import com.vultisig.wallet.data.usecases.GetAvailableTokenBalanceUseCase
@@ -134,10 +135,22 @@ internal class DefaultSendStrategy(
                                 chainAccountAddressRepository.isValid(chain, dstAddress)
                         }
 
-                    if (!chainAccountAddressRepository.isValid(chain, dstAddress)) {
-                        throw InvalidTransactionDataException(
-                            UiText.StringResource(R.string.send_error_no_address)
-                        )
+                    // The form's own recipient check is debounced and drives nothing but the
+                    // Continue button, and this path re-resolves the address from the raw field
+                    // rather than reading what the form resolved. Re-check here so a submit that
+                    // races that debounce still can't build a transaction for a recipient the
+                    // form would have refused — a Solana token account above all, which strands
+                    // an SPL transfer for good.
+                    when (chainAccountAddressRepository.validateRecipient(chain, dstAddress)) {
+                        RecipientValidity.Valid -> Unit
+                        RecipientValidity.InvalidForChain ->
+                            throw InvalidTransactionDataException(
+                                UiText.StringResource(R.string.send_error_no_address)
+                            )
+                        RecipientValidity.NotAWalletAddress ->
+                            throw InvalidTransactionDataException(
+                                UiText.StringResource(R.string.error_recipient_not_a_wallet_address)
+                            )
                     }
 
                     val selectedTokenValue =
--- a/app/src/test/java/com/vultisig/wallet/ui/models/transaction/AddressEntryViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/transaction/AddressEntryViewModelTest.kt
@@ -9,6 +9,7 @@ import com.vultisig.wallet.data.models.AddressBookEntry
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.repositories.AddressBookRepository
 import com.vultisig.wallet.data.repositories.ChainAccountAddressRepository
+import com.vultisig.wallet.data.repositories.RecipientValidity
 import com.vultisig.wallet.data.repositories.RequestResultRepository
 import com.vultisig.wallet.data.repositories.order.OrderRepository
 import com.vultisig.wallet.data.usecases.RequestQrScanUseCase
@@ -61,6 +62,17 @@ internal class AddressEntryViewModelTest {
         requestQrScan = mockk(relaxed = true)
         addressBookRepository = mockk(relaxed = true)
         chainAccountAddressRepository = mockk(relaxed = true)
+        // The screen validates through validateRecipient, which only parts ways with isValid for
+        // an off-curve Solana recipient; deriving one from the other keeps each test's own isValid
+        // stub as the thing that decides the outcome.
+        every { chainAccountAddressRepository.validateRecipient(any(), any()) } answers
+            {
+                if (chainAccountAddressRepository.isValid(firstArg(), secondArg())) {
+                    RecipientValidity.Valid
+                } else {
+                    RecipientValidity.InvalidForChain
+                }
+            }
         orderRepository = mockk(relaxed = true)
         requestResultRepository = mockk(relaxed = true)
     }
--- a/app/src/test/java/com/vultisig/wallet/ui/models/transaction/AddressEntryViewModelTest.kt
+++ b/app/src/test/java/com/vultisig/wallet/ui/models/transaction/AddressEntryViewModelTest.kt
@@ -67,7 +67,7 @@ internal class AddressEntryViewModelTest {
         // The screen validates through validateRecipient, which only parts ways with isValid for
         // an off-curve Solana recipient; deriving one from the other keeps each test's own isValid
         // stub as the thing that decides the outcome.
-        every { chainAccountAddressRepository.validateRecipient(any(), any()) } answers
+        coEvery { chainAccountAddressRepository.validateRecipient(any(), any()) } answers
             {
                 if (chainAccountAddressRepository.isValid(firstArg(), secondArg())) {
                     RecipientValidity.Valid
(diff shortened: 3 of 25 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
```diff
@@ -60,13 +61,12 @@ internal class AddressManager(
...
-    // True when a non-empty recipient failed chain validation and couldn't resolve to a valid
-    // address. Drives the inline "not a valid address for this chain" error; false once the input
-    // is valid or empty. While name resolution is in flight the flag holds its previous value, so a
+    // Why a non-empty recipient was rejected, or null once the input is accepted or empty. Drives
+    // the inline recipient error. While name resolution is in flight the value is held, so a
     // standing error stays put until the new input resolves instead of blinking off and back on.
     // Chain-general: every send chain validates through the same path below.
-    private val _invalidAddress = MutableStateFlow(false)
-    val invalidAddress: StateFlow<Boolean> = _invalidAddress.asStateFlow()
+    private val _addressError = MutableStateFlow<RecipientValidity?>(null)
```
johnnyluo, line 68: This state only drives the reactive UI error (disables Continue) and is debounced 300ms upstream (`.debounce(300)` in `collectResolvedAddress`). `DefaultSendStrategy.submit()` (`app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt:124-137`) never reads `resolvedDstAddress` or this error state — it independently re-resolves the destination from the raw text field via `AccountValidator.validate()` and gates only on `chainAccountAddressRepository.isValid(chain, dstAddress)`, not `validateRecipient`.

Concretely: paste a Solana token-account address, tap Continue inside the 300ms debounce window (e.g. re-entering the recipient with an amount already filled from a prior send) — `addressError` still holds its previous non-blocking value, Continue is enabled, and `submit()` builds the transaction with the token account as recipient, reproducing the exact bug this PR fixes.

The PR's own code shows this pattern is already known to be needed: the memo-length re-check a few lines below the address check in `DefaultSendStrategy.submit()` exists specifically "so a submit that races the form's per-keystroke validation still can't start a keysign" (issue #5435). The swap path in this same PR gets it right too — `SwapFormViewModel.swap()`/`.placeLimitOrder()` call `validateRecipient` (via `externalRecipientError()`) as a hard gate right before building the transaction. `DefaultSendStrategy` is the one path where that submit-time gate is missing.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ChainAccountAddressRepository.kt
```diff
@@ -30,6 +31,29 @@ interface ChainAccountAddressRepository {
...
     fun isValid(chain: Chain, address: String): Boolean
+
+    /**
+     * Whether [address] can be handed funds on [chain], and why not when it can't.
+     *
+     * Stricter than [isValid] on purpose, and separate from it on purpose: [isValid] also screens
+     * contract and mint addresses, which are routinely program-derived, so the recipient rule
+     * cannot live there without rejecting them.
+     */
+    fun validateRecipient(chain: Chain, address: String): RecipientValidity
```
johnnyluo: Requesting changes: `DefaultSendStrategy.submit()` — the actual authoritative gate before a Send transaction is built — still calls `isValid(chain, dstAddress)` at `app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt:137`, not `validateRecipient`. It also re-resolves `dstAddress` itself from the raw address field rather than consuming `AddressManager.resolvedDstAddress`, so it doesn't inherit the UI-level block either. That leaves a race window (the form's address validation is debounced 300ms) where a token-account/PDA recipient can still reach the transaction builder on the plain Send flow — exactly the scenario this PR sets out to prevent.

Could you update `DefaultSendStrategy.submit()` to gate on `validateRecipient(chain, dstAddress) == RecipientValidity.Valid`, and add a `DefaultSendStrategyTest` case pinning that a Solana token-account recipient is rejected at submit? (`DefaultSendStrategyTest.kt` wasn't touched by this PR, so nothing currently covers this path.)

Worth a quick check too on the other submit-time `isValid(chain, dstAddress)` callers under `send/submit/` (`BondStrategy`, `RedeemStrategy`, `UnstakeStrategy`, `StakeStrategy`, `MintStrategy`, `WithdrawUsdcCircleStrategy`) in case any of them can target a Solana destination — likely out of scope for this PR but the same gap would apply.

johnnyluo (dismissed):
The curve-check itself, and the wiring into the send form / swap external recipient / address-book entry screens, all look solid and well-tested. One gap before this can merge: the actual Send submit path isn't gated by the new check, so the exact bug this PR fixes is still reachable through it. See inline comments.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ChainAccountAddressRepository.kt
```diff
@@ -140,6 +164,14 @@ internal class ChainAccountAddressRepositoryImpl @Inject constructor() :
             else -> chain.coinType.validate(address)
         }
 
+    override fun validateRecipient(chain: Chain, address: String): RecipientValidity =
+        when {
+            !isValid(chain, address) -> RecipientValidity.InvalidForChain
+            chain == Chain.Solana && !SolanaProgramDerivedAddress.isWalletAddress(address) ->
```
rkokhatskyi: `validateRecipient` rejects every off-curve Solana address because `isWalletAddress` is only the curve test, so a Squads vault or DAO treasury PDA is blocked on send, swap and the address book. Those PDAs hold SPL tokens through `ATA(pda, mint)` and spend them via invoke_signed, and iOS `AddressService.swift:177` accepts them

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ChainAccountAddressRepository.kt
aminsato: You're right, and I checked both halves: iOS `AddressService.validateAddress` is bare `coinType.validate` with no curve test at all (it ships `SolanaProgramDerivedAddress` but only uses it for Kamino derivation), and a Squads vault or DAO treasury really does receive SPL through `ATA(pda, mint)` — nothing about being off-curve makes an address unspendable by the program that owns it.

Fixed in 1cfc2876f by making the curve test the question rather than the verdict. An off-curve Solana address now costs one `getAccountInfo`, and only what the cluster reports as owned by SPL Token or Token-2022 is refused — that is the account that actually strands an SPL transfer. Owned by any other program is Valid.

That needed an answer `null` can't carry, since `getAccountOwner` returns it for both a missing account and a failed lookup, so there's a new `getAccountOwnership` returning `Owned` / `Missing` / `Unavailable` (`getAccountOwner` is now a projection of it, callers unaffected).

`Missing` is Valid deliberately — it's your exact example. A vault that has only ever held SPL never needs an account of its own, because the tokens live in the ATAs derived from it, so "no account at this address" is the normal state of the destination you named.

`Unavailable` is still refused: the send path draws the blockhash and the fee from the same cluster, so nothing can be built while it's unreachable anyway, and the address book can be retried.

One narrowing worth stating plainly: an off-curve address with no account on-chain that is genuinely garbage (a corrupted paste decoding to 32 off-curve bytes) is now allowed where the blanket rule blocked it. I think that's the right price — address-poisoning vanity addresses are generated from real keypairs and are on-curve — but it is a real loosening, not just a false-positive fix.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ChainAccountAddressRepository.kt
aminsato: Confirmed and fixed in 6ce2ca622 — you were right that this was the path that mattered. `AccountValidator.validate()` re-resolves `dstAddress` from the raw field and never reads `AddressManager.resolvedDstAddress`, so nothing at submit inherited the form's block, and `isValid` is exactly the base58 length check a token account passes.

`DefaultSendStrategy.submit()` now gates on `validateRecipient`, with the verdict picking the message: `InvalidForChain` still reads as an invalid address, `NotAWalletAddress` says the address is well-formed but isn't a wallet. Two `DefaultSendStrategyTest` cases pin it — a Solana token account is refused at submit with `addTransaction` never called, and a malformed address still reports as an invalid address so the new branch didn't swallow the old one.

On the other `send/submit/` callers: checked, and they can't reach a pasted Solana recipient. `SendStrategies.submitFor` routes only the THORChain/Maya/Circle actions to them (RUJI/TCY/yRUNE/yTCY staking, bond/unbond, Circle USDC withdraw) and those destinations come from the protocol, not the address field. Everything else — including ADD_LP, REMOVE_LP and the TRX freeze actions — falls through to `default`, which is the one now gated. So every Solana send, plain or DeFi, goes through this check; I've left their `isValid` calls alone rather than widen the diff.

Worth flagging since it changes the rule you reviewed: per @rkokhatskyi's comment below, 1cfc2876f narrowed `validateRecipient` so off-curve alone is no longer the verdict — it now asks the cluster and refuses only an address owned by SPL Token or Token-2022 (plus an unreachable lookup). A Squads vault or DAO treasury PDA passes. The submit gate is on top of that narrowed rule.

app/src/main/java/com/vultisig/wallet/ui/models/send/AddressManager.kt
aminsato: Your read of this state is right: it drives the inline error and the Continue gate and nothing else, and behind `.debounce(300)` it can be one keystroke stale. I left it that way rather than trying to make it authoritative — a per-keystroke flow can always lag the tap — and put the real gate at submit in 6ce2ca622, alongside the memo re-check you pointed at.

One thing that did change here in 1cfc2876f: `validateRecipient` now suspends, since an off-curve address costs a `getAccountInfo` before it's judged. That falls out naturally in this collector — it's already inside `mapLatest`, so a verdict for an address the user has moved on from is cancelled rather than landing late.

app/src/test/java/com/vultisig/wallet/ui/models/transaction/AddressEntryViewModelTest.kt
```diff
@@ -117,6 +131,28 @@ internal class AddressEntryViewModelTest {
...
+            vm.addressTextFieldState.edit { replace(0, length, SOL_TOKEN_ACCOUNT) }
+
+            vm.saveAddress()
+            advanceUntilIdle()
+
+            vm.state.value.addressError shouldBe
+                UiText.StringResource(R.string.error_recipient_not_a_wallet_address)
+            coVerify(exactly = 0) { addressBookRepository.add(any()) }
+        }
+
```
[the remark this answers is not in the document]
aminsato: Applied in 1cfc2876f. The test now stubs the route with `chainId = "Solana"` and `address = SOL_TOKEN_ACCOUNT` so the VM actually selects Solana, scopes the stub to `validateRecipient(Chain.Solana, SOL_TOKEN_ACCOUNT)` instead of `any()`, and adds a `coVerify` on that exact invocation. The existing error and `add`-not-called assertions are unchanged.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ChainAccountAddressRepository.kt
aminsato: Flagging a follow-up, since the rule you reviewed moved again in 9aa255a61 — your PDAs still pass, but not for the reason they passed yesterday.

CodeRabbit pointed out that the curve test was still deciding *who gets asked*, and that leaves the hole this PR exists to close. A token account is usually an associated one and so off-curve, but the auxiliary form — `spl-token create-account` against a fresh keypair — is on the curve like any wallet. That is how token accounts were made before ATAs existed and how some exchanges still hold them, and it strands an SPL transfer identically. Under yesterday's rule it was waved straight through.

So every well-formed Solana address is asked about now. What the curve still decides is only how an unreachable cluster resolves, and it has to resolve opposite ways on the two sides:

| | token program | other program | missing | unavailable |
|---|---|---|---|---|
| off-curve | refuse | valid | valid | **refuse** |
| on-curve | refuse | valid | valid | **valid** |

The bottom-right cell is the one worth arguing about. Keeping `Unavailable` a blanket refusal while querying everything would mean any RPC blip blocks every Solana send, including plain wallets — that is a much worse failure than the one this guard prevents. On the curve someone can hold the key, which is the ordinary case for every address a user pastes, so a failed lookup falls back to what the app did before this PR: presume a wallet. Off the curve nobody can hold the key, so an unanswered lookup still leaves a real chance the address is a token account, and it stays refused.

Net effect on your case: a Squads vault or DAO treasury is off-curve and either `Missing` or owned by its own program, so it is `Valid` on the answer, exactly as after 1cfc2876f. The cost is one `getAccountInfo` on every Solana recipient rather than only the off-curve ones — behind `debounce(300)` + `mapLatest` on the form, and one call at submit.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ChainAccountAddressRepository.kt
aminsato: The submit gate you asked for is unchanged in 6ce2ca622 — flagging that the rule underneath it moved again in 9aa255a61, since it changes what that gate actually refuses.

CodeRabbit caught that the curve test was still deciding who gets asked, so an *auxiliary* token account — one opened with `spl-token create-account` against a fresh keypair rather than derived as an ATA — is on-curve and was returning `Valid` without the cluster ever being consulted. That is the pre-ATA form, still held by some exchanges, and it strands an SPL transfer exactly like an ATA does. Every well-formed Solana address is asked about now.

The curve survives only as the `Unavailable` fallback, and it flips by side: off-curve unreachable stays refused, on-curve unreachable is `Valid`. Querying everything while keeping `Unavailable` a blanket refusal would block every Solana send on any RPC blip, which is a worse failure than the one being prevented — on the curve, presuming a wallet is what the app did before this PR.

For `DefaultSendStrategy.submit()` specifically: it now costs one `getAccountInfo` on every Solana send rather than only when the pasted address happened to be off-curve. `getAccountOwnership` catches its own failures and answers `Unavailable`, so it cannot throw into the submit path. The two `DefaultSendStrategyTest` cases are unaffected.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/ChainAccountAddressRepository.kt
```diff
@@ -140,6 +184,38 @@ internal class ChainAccountAddressRepositoryImpl @Inject constructor() :
             else -> chain.coinType.validate(address)
         }
 
+    override suspend fun validateRecipient(chain: Chain, address: String): RecipientValidity =
+        when {
+            !isValid(chain, address) -> RecipientValidity.InvalidForChain
+            chain != Chain.Solana -> RecipientValidity.Valid
+            SolanaProgramDerivedAddress.isWalletAddress(address) -> RecipientValidity.Valid
```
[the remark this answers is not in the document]
aminsato: Confirmed and fixed in 9aa255a61. You were right that curve status doesn't identify an account: a token account opened with `spl-token create-account` against a fresh keypair — the pre-ATA form some exchanges still hold — is on the curve like any wallet, so line 191 was waving through the exact account this guard exists for.

Every well-formed Solana address is asked about now, and there's a test stubbing an on-curve address as `Owned(TOKEN_PROGRAM)` expecting `NotAWalletAddress`, as you suggested.

One narrowing your comment didn't cover, and it was the blocker on just widening the query: `Unavailable` was a blanket refusal. Asking about every address while keeping that would mean any RPC blip refuses every Solana recipient, plain wallets included — a far worse failure than the one being prevented. So the curve test survives, but only as the fallback for an unanswered lookup, and it resolves opposite ways on the two sides:

| | token program | other program | missing | unavailable |
|---|---|---|---|---|
| off-curve | refuse | valid | valid | **refuse** |
| on-curve | refuse | valid | valid | **valid** |

Off the curve only the owning program can sign, so silence still leaves a real chance the address is a token account. On the curve someone can hold the key — the ordinary case for every pasted address — so silence falls back to what the app did before this PR.

app/src/main/java/com/vultisig/wallet/ui/models/send/submit/DefaultSendStrategy.kt
```diff
@@ -134,10 +135,22 @@ internal class DefaultSendStrategy(
...
+                    when (chainAccountAddressRepository.validateRecipient(chain, dstAddress)) {
+                        RecipientValidity.Valid -> Unit
+                        RecipientValidity.InvalidForChain ->
+                            throw InvalidTransactionDataException(
+                                UiText.StringResource(R.string.send_error_no_address)
+                            )
+                        RecipientValidity.NotAWalletAddress ->
+                            throw InvalidTransactionDataException(
+                                UiText.StringResource(R.string.error_recipient_not_a_wallet_address)
+                            )
```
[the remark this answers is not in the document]
aminsato: Applied at two of the three sites in 43d274e63, and skipped at the anchor.

`SwapFormViewModel.setExternalRecipient` and `AddressEntryViewModel.saveAddress` are now `safeLaunch`. The swap one carries a handler that clears `quoteRecipient` rather than only logging: a verdict that never arrives shouldn't leave quotes routing to an address nothing vouched for, and the pre-flight gate asks again before anything is built. `externalRecipientRoutingJob` still holds the returned `Job`, so the cancel-previous behaviour is intact.

`DefaultSendStrategy` is left as it was, which matches your own note — it already has the `try/catch` that calls `showError` and `hideLoading`, and wrapping it in `safeLaunch` would add a second handler over one that already does the right thing.

Worth stating that neither migration fixes a reachable crash: `SolanaApi.getAccountOwnership` catches `Exception` and answers `Unavailable`, re-throwing only `CancellationException`, which is ordinary cancellation. So `validateRecipient` cannot throw into these coroutines today. They're `safeLaunch` because they perform network calls and the repo's convention says so, not because there's something waiting to be caught.

johnnyluo (approved):
LGTM

PR #5721, opened 2026-08-26
fix(home): don't flash "No chains enabled" before accounts load

### LINKED ISSUE #5710: Home flashes "No chains enabled — You've disabled all chains" on every cold start before accounts load
## Observed

Cold-launching v1.0.118 (debug build from tag `37f13e2df`, Pixel 9 Pro API 36 emulator, one Fast Vault `fastandroid` with 12+ chains enabled), the home screen paints the **"No chains enabled — You've disabled all chains. Make sure that at least one chain is enabled."** empty state for ~2–3 s before the portfolio loads. Frame series from `adb screencap` after `am force-stop` + launch: 2 s blank, 4 s splash, **6 s empty-state card visible (vault chip has no name, balance skeleton)**, 9 s portfolio fully rendered. Reproduced 3/3 cold starts. Nothing was disabled — the vault selector and portfolio are intact once loading completes.

## Cause

`VaultAccountsScreen.kt:343` renders the empty state on `state.getAccounts.isEmpty()` with no loading guard, so the first emission (before the accounts query returns) is indistinguishable from "user disabled every chain". #5622 (re-read the DeFi list on foreground) touched this file this release; the string itself dates from #2629.

## Expected

While accounts are loading, show the skeleton/empty list — the "you've disabled all chains" copy only after the accounts flow has emitted at least once and is empty. An "all chains disabled" message flashing on every cold start is alarming copy for a self-custody wallet.

## Env

v1.0.118 tag, emulator Pixel_9_Pro_2 API 36, debug build. Found during the v1.0.118 release sweep (`release-qa-templates/maestro/reports/v1.0.118-android/`).

aminsato:
Closes #5710

## Summary

The home list rendered the **"No chains enabled — You've disabled all chains"** empty state on `state.getAccounts.isEmpty()` with no loading guard. That predicate is also true before the accounts flow has emitted anything, so on a cold start — where deriving addresses for a dozen chains and reading their cached balances takes seconds — every launch flashed the empty state over an intact portfolio. Switching vaults did the same, since `loadData` clears the list first.

Telling a self-custody user their chains are gone, on every launch, is the wrong copy to guess with.

## Changes

- `VaultAccountsUiModel` tracks whether each list's flow has emitted (`hasLoadedAccounts` / `hasLoadedDeFiAccounts`), exposed through an `areAccountsLoaded` getter that mirrors the existing `getAccounts` wallet/DeFi switch.
- The flag is set in `updateUiStateFromList` — the single funnel both tabs go through. It fires on the cached-DB emission, which lands before any network work, so a vault that really does have every chain disabled still shows the message promptly rather than waiting on a fetch.
- Cleared on vault switch alongside the lists it describes, so the guard re-arms.
- `VaultAccountsScreen` only renders the empty state once the flag is set; until then the list renders empty. Deliberately *not* reset on the chain-picker reload, so disabling every chain there still reads correctly.

## Notes

The last touch on this block was the ktfmt sweep (#3265) — the missing guard has been there since the v2 home screen landed and simply got more visible as the chain count grew, rather than being introduced by #5622 as the issue supposed.

If `loadAddressBalances` ever fails outright, its `catch` completes without emitting and the list stays empty instead of showing the wrong copy forever. A real error state there is worth a separate ticket.

## Test plan

Three regression tests added to `VaultAccountsViewModelTest`:

- an empty list is not the no-chains-enabled state until the accounts flow emits, and flips as soon as it does
- switching vaults re-arms the guard until the next vault's accounts land
- the DeFi tab waits on its own flow rather than on the wallet one

`./gradlew :app:testDebugUnitTest --tests "*VaultAccountsViewModelTest"` — 13 + 31 tests, 0 skipped, 0 failures. ktfmt applied.

Manual: cold-launch a multi-chain vault; the chain card stays empty until the rows land, with no "No chains enabled" flash. Then disable every chain from the picker and confirm the message still appears.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -95,6 +95,10 @@ internal data class VaultAccountsUiModel(
     val isBalanceValueVisible: Boolean = true,
     val accounts: List<AccountUiModel> = emptyList(),
     val defiAccounts: List<AccountUiModel> = emptyList(),
+    // Set once the corresponding flow has emitted, so an empty list can be read as "no chains
+    // enabled" rather than "not loaded yet".
+    val hasLoadedAccounts: Boolean = false,
+    val hasLoadedDeFiAccounts: Boolean = false,
     val searchTextFieldState: TextFieldState = TextFieldState(),
     // Per-banner visibility, each gated on a global dismissal whose lifetime is the banner's own
     // policy (#5064). The upgrade banner additionally requires the vault to be GG20
@@ -116,6 +120,14 @@ internal data class VaultAccountsUiModel(
                 accounts
             } else {
                 defiAccounts
+            }
+
+    val areAccountsLoaded: Boolean
+        get() =
+            if (cryptoConnectionType == CryptoConnectionType.Wallet) {
+                hasLoadedAccounts
+            } else {
+                hasLoadedDeFiAccounts
             }
 }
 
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #5722, opened 2026-08-26
feat(swap): stop allowlisting chains for SwapKit; add HOOD + HyperEVM

### LINKED ISSUE #5699: [Add] SwapKit: Robinhood + HyperEVM, and stop closed-listing chains [Swap]
# [Add] SwapKit: Robinhood + HyperEVM, and stop closed-listing chains [Swap]

SwapKit went live with Robinhood (4663) and HyperEVM. Android already holds both and already swaps them through other providers. SwapKit is not offered for either, because Android keeps its own closed SwapKit maps (`SwapProviderTable` + `chainPrefix`).

This ticket is **Android-owned**. Do not wait on the SDK.

## Current state

- `SwapProviderTable`: `Chain.Robinhood -> ONEINCH, LIFI, KYBER`. `Chain.Hyperliquid -> LIFI`.
- `SwapProviderTableTest` pins Hyperliquid in the "must not offer SWAPKIT" set.
- `chainPrefix` in `SwapKitQuoteSource` is `else -> null` for both, so even if the table offered SwapKit the quote source would drop the pair.
- Every new SwapKit corridor currently needs a hand-edit of the provider table + prefix + `/providers` id + `/track`. That is the thing to stop.

## Must Do

Immediate two chains:

- [ ] Confirm SwapKit chain ids and asset prefixes empirically (`/providers`, `/v3/tokens`). Docs: Robinhood = `4663`. HyperEVM may be `hype`, not `"999"`. Do not guess `HOOD` / `HYPE`.
- [ ] Add `SWAPKIT` to `providersFor` for `Chain.Robinhood` and `Chain.Hyperliquid`.
- [ ] Map both in `chainPrefix`, the `/providers` id mapper, and `/track`.
- [ ] Update `SwapProviderTableTest` (remove Hyperliquid from the "must not offer SWAPKIT" set).
- [ ] If we cannot reputation-check a Robinhood *source* (Blockaid does not list 4663 today), keep Robinhood as SwapKit dest-only. Do not fail the pair.

Open the closed list (this is in scope, not a later spike):

- [ ] For any chain Android already holds and can sign, attempt a SwapKit quote without a new allowlist row.
- [ ] If `/providers` says the chain is off, the quote misses, the tx type is unknown, or we have no signer: **skip SwapKit only**. Leave 1inch / Kyber / LiFi / THOR / Maya standing. Never fail the whole pair because SwapKit had nothing.
- [ ] Prefix still has to be right. Pull it from SwapKit's token catalogue, or skip that chain's SwapKit attempt.
- [ ] Pin tests: Robinhood + HyperEVM succeed (or dest-only as above); an unsupported/unsignable SwapKit miss does not hide other providers.

`Chain.Hyperliquid` is HyperEVM 999 (`/hyperevm/` RPC). Do not treat it as HyperCore.

## Out of Scope

- Aleo
- Hyperliquid Core / HyperCore (android#5227 is Core staking, a different chain)
- Changing 1inch / Kyber / LiFi behaviour
- SDK / iOS / Windows code

## Acceptance Criteria

- [ ] SwapKit quotes are attempted for Robinhood and HyperEVM
- [ ] Provider table tests updated
- [ ] A SwapKit miss does not hide 1inch / Kyber / LiFi
- [ ] New SwapKit corridors on chains we already hold do not require a fresh table PR just to *try* the quote
- [ ] HyperEVM is not treated as HyperCore

aminsato:
Closes #5699

## Why

SwapKit was named per chain in `SwapProviderTable`, so every new corridor needed a client edit before a quote was even attempted. Robinhood (4663) and HyperEVM (999) were therefore unreachable, despite the wallet already holding both and swapping them through other providers.

Rather than add two more rows, this ports iOS' shape: there is no SwapKit chain list any more.

## What changed

**The list is gone, not extended.** New `SwapKitCapability` decides receivability as a property — every EVM chain except Sei, plus the non-EVM chains that have a signer or deposit path — and `providersFor` appends `SWAPKIT` centrally. Robinhood and HyperEVM needed no entry; zkSync, Mantle, Blast and Cronos came along with them, all gated by `/providers` at quote time.

**Source eligibility is directional.** An EVM SwapKit route signs calldata against a router we did not build, so only chains the security scanner covers may originate one. Blockaid does not index 4663, so Robinhood is destination-only — `eligibleProvidersFor` drops SwapKit alone, never the pair. HyperEVM differs only in being scannable, which is why `Chain.Hyperliquid` is added to `BlockaidScannerService`: a gate claiming a check that never runs is worse than no gate. **This is the one change outside swap.**

## Three latent defects surfaced while porting

- `SwapKitProviderCache` unioned `supportedChainIds` with no provider filter, where iOS uses `enabledChainIds` minus THORChain/Maya. Against the live response that is six chains wrong (`cosmoshub-4`, `kaiyo-1`, `thorchain-1`, `mayachain-mainnet-v1`, `hype`, `stellar`) — harmless while the chain map knew 8 chains, incorrect the moment it is broadened, which this PR does.
- The cache was bound in DI but **never called**, so no enablement gate existed at all. A comment claimed iOS fails open on a cache miss; it does not — `isEnabled` fails closed. Now wired in, raising the existing `ProviderNotEnabled` variant that was already mapped downstream but unreachable.
- `txType` was dispatched without checking it against the source chain. The allowlist made most mismatches unreachable; opening it makes that check load-bearing, so `signerMatches` ports iOS' `canSign`.

`/track` chain ids are now derived from the enum for EVM instead of listed, so HyperEVM stops falling back to the source-chain link.

## Confirmed, not guessed

The ticket asked for empirical confirmation. Against `/providers` and `/tokens`:

| | chainId | asset prefix |
|---|---|---|
| Robinhood | `4663` | `HOOD` (`HOOD.ETH`, `HOOD.TSLA-0x…`) |
| HyperEVM | `999` | `HYPEREVM` (`HYPEREVM.HYPE`) |

The catalogue's separate `HYPE` bucket is HyperCore — `USDC:0x…` style addresses, `supportedChainIds` only, never enabled. It must not resolve to `Chain.Hyperliquid`, and a test pins that.

## Test plan

- `:data` and `:app` unit tests green; `assembleDebug`, `ktfmtCheck`, lint clean
- New `SwapKitCapabilityTest` (receivability as a property over the whole enum; the receive/quote asymmetry) and `SwapKitChainIdentifierTest` (EVM ids derived; 4663 and 999)
- `SwapKitProviderCacheTest` extended: enabled-vs-supported, the THOR/Maya filter, HyperCore's `hype` never resolving to HyperEVM
- `SwapProviderTableTest`: Robinhood destination-but-not-source, a blocked source keeping every other provider, HyperEVM both ways

Verified on device against a real vault:

- `SOL → ETH` quotes `SwapKit (NEAR)` and wins the route picker on output
- SwapKit toggle off → SwapKit gone and the picker greys out; back on → returns
- `Robinhood → Cardano` errors **on selection** ("Swap route not available"), while `Cardano → Robinhood` gets past selection and fails only **after** the quote ("No routes available for this pair"). Same two chains, opposite directions — that asymmetry is the destination-only gate
- `Cardano → Ethereum` quotes `SwapKit (NEAR)`, SwapKit being the only shared provider
- HyperEVM Send shows "Transaction scanned by Blockaid"

A SwapKit route was also signed and broadcast end to end — `SOL → ETH`, source leg
[`3mwzANRf…JYJsj`](https://orb.helius.dev/tx/3mwzANRfqNQfQrJavLuBitnu8XPgGzzCDGpoGukKKQyjjMAaAd1UEf5rZFxw5bNCRhQYQ6uALdqnbdc4tCqJYJsj)
(slot 441844037, `err: null`, 0.05 SOL, fee 0.000005 SOL). It is a bare system transfer with no DEX
program invocation, which is the expected shape for a NEAR Intents route — SwapKit hands out a
per-route deposit address and settlement happens on the destination leg. A LI.FI Solana swap would
carry Jupiter/DEX instructions instead, so the instruction shape is itself evidence the route went
through SwapKit.

## Known limitation

FLASHNET is the only non-1inch provider with 4663/999 enabled and currently returns `quoteError` on every cross-chain pair for both, at every size. So SwapKit routes for Robinhood and HyperEVM are wired end to end but not yet reachable. The point of the change is that they will light up with no client release when upstream recovers.

## Summary by CodeRabbit

* **New Features**
  * Expanded SwapKit route support across additional EVM and non-EVM networks.
  * Added support for Robinhood, HyperEVM, and Hyperliquid transaction tracking and scanning.
  * Added validation for compatible source and destination chains.
  * Improved provider availability detection using currently enabled chains.

* **Bug Fixes**
  * Improved chain identification and routing across supported networks.
  * Prevented unsupported or non-routable chains from appearing as eligible providers.
  * Preserved alternative providers when SwapKit is unavailable.
  * Retained the last successful provider snapshot during refresh failures, with controlled retries.
  * Improved inbound fee retrieval for enabled routes.


data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitCapability.kt
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * True when the app can hold a SwapKit destination asset on [chain].
+     *
+     * Every EVM chain qualifies except [Chain.Sei], which the wallet holds but does not swap on at
+     * all (it is the one EVM chain absent from iOS' `isSwapAvailable`). No allowlist beyond that —
+     * a chain SwapKit does not actually route is dropped later by the `/providers` gate in
+     * [SwapKitQuoteSource], and an asset SwapKit does not list is dropped by `/v3/quote`.
+     */
+    fun canReceiveOn(chain: Chain): Boolean =
+        if (chain.standard == TokenStandard.EVM) chain != Chain.Sei
```
rkokhatskyi: `canReceiveOn` admits ZkSync, Mantle, Blast and Cronos while `chainPrefix` in `SwapKitQuoteSource.kt:983` has no entry for any of them, so a pair like Cardano to ZkSync now passes `isPairRoutable` in `SwapQuotePipelineController.kt:821` and loses the immediate not supported guidance for a quote that always throws

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
```diff
@@ -108,6 +110,12 @@ constructor(
         if (!config.isFeatureEnabled.first()) {
             return TokenValue(BigInteger.ZERO, request.srcToken)
         }
+        if (
```
rkokhatskyi: This provider cache gate cannot block a bad pair because `fetchInboundFee` is only reached from `JoinSwapUiModelBuilder.kt:480` for a `SwapPayload.SwapKit` a device already quoted. Its only reachable effect is a confident $0.00 swap fee row on the joiner for Tron, Sui, Ton and Ripple, where the initiator shows the real one

rkokhatskyi (dismissed):

johnnyluo (approved):
LGTM

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitCapability.kt
aminsato: Confirmed, and fixed — thanks. One detail on the mechanism: today the pair dies one step earlier than the missing prefix. I re-checked `/providers` against the proxy and `enabledChainIds` has no `324`, `5000`, `81457` or `25`, so `assertChainsEnabled` throws `ProviderNotEnabled` before `assetIdentifier` is ever reached. The missing `chainPrefix` is the second guaranteed failure, waiting for the day SwapKit lights one of them up. Either way your point stands: `isPairRoutable` is a synchronous table lookup and sees neither gate, so Cardano/TON/SUI/BTC/Zcash → ZkSync lost the instant "Swap route not available" and got a spinner plus a late error instead.

Fixed by making the offer honest at the table level: `chainPrefix` moved out of `SwapKitQuoteSource` into `SwapKitAssetPrefix`, and `canReceiveOn` now requires an entry there. A chain with no prefix cannot address a quote at all, so it is no longer offered — ZkSync, Mantle, Blast and Cronos drop back out, while the PR's point survives: a new network still needs exactly one entry, in the map that is load-bearing anyway (it is the same entry HOOD and HYPEREVM needed to be quotable). Blast also stops being a SwapKit *source* even though Blockaid indexes it, since it is no longer receivable.

Pinned in `SwapKitCapabilityTest` (`EVM chains SwapKit has no asset identifier for are not offered at all`) and moved those four to the not-offered list in `SwapProviderTableTest`. `assetIdentifier` keeps its `NoRoutes` throw as defense-in-depth.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Correct on both halves — the gate is removed.

You are right that it can never veto a pair here: the only caller is `JoinSwapUiModelBuilder` for a `SwapPayload.SwapKit` another device already quoted, so the routing decision is long since made. Its one reachable effect was replacing the initiator's real fee with a confident $0.00 whenever this device's snapshot said otherwise, on any non-UTXO source. Letting `/v3/quote` answer restores parity with the initiator; a genuine failure still degrades to zero through the caller's `catch (e: SwapKitError)`, which was already there.

Chasing how the snapshot goes missing turned up a second bug worth fixing in the same pass: `SwapKitProviderCacheImpl.ensureFresh` returned `null` on a failed refresh **even when a good snapshot existed**, so one bad `/providers` call after the 24h TTL reported every chain disabled — SwapKit hidden app-wide, and exactly the state that made this $0.00 reachable. iOS serves last-good there (`SwapKitProviderCache.swift`: "Once a snapshot exists, `providers(now:)` serves it as last-good on a later fetch failure"), so Android now does too. Only the genuine no-data edge — never fetched, or after `invalidate()` — still fails closed.

Tests added: `fetchInboundFee is not gated on the providers snapshot`, `fetch throws ProviderNotEnabled when either leg is not in the providers snapshot` (the quote-side gate had no coverage at all), and two cache tests for last-good vs. the no-data edge.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitCapability.kt
aminsato: Pushed as cbb56e4fc — `SwapKitAssetPrefix.kt` (new), `SwapKitCapability.canReceiveOn` gated on it, and the four chains moved to the not-offered list in `SwapProviderTableTest`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitQuoteSource.kt
aminsato: Pushed as cbb56e4fc — gate removed from `fetchInboundFee`, and `SwapKitProviderCacheImpl` now serves the last good snapshot when a refresh fails.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitProviderCache.kt
```diff
@@ -90,19 +100,28 @@ internal class SwapKitProviderCacheImpl @Inject constructor(private val api: Swa
...
-                // at the API layer. Fail-closed so SwapKit is skipped until the next refresh.
-                null
+                // at the API layer. Serve the last good answer if there is one.
+                lastGoodOrNull()
             } catch (e: Exception) {
                 // Unexpected (mapping/parse regression, programmer error). Surface it in logs
                 // rather than silently treat SwapKit as "disabled" forever.
                 Timber.w(e, "SwapKit providers refresh failed unexpectedly")
-                null
+                lastGoodOrNull()
```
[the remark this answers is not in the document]
aminsato: Valid, and fixed in 2ee787859.

Confirmed the mechanism: serving the last good snapshot leaves `fetchedAtMillis` behind the TTL, so nothing in `ensureFresh` records that an attempt was made and every later call re-enters the refresh path. The cost is a little worse than one request — `assertChainsEnabled` tests both legs of the pair, so it is two dead round-trips in front of each quote, repeated on every quote-expiry refresh for as long as the proxy is down.

One correction to the framing: this cannot overload the endpoint. `isEnabled` is only reached from the quote path, which is user-driven and quote-expiry-timed, not a loop. The real cost is latency on the swap screen during a `/providers` outage.

Fixed as suggested — a `retryAfterMillis` deadline armed in both catch branches, checked alongside the TTL in a shared `isServable(now)`, cleared by a successful refresh and by `invalidate()`. Backoff is 5 minutes: short enough that a recovered endpoint is picked up within a session, long enough that a sustained outage stops taxing the quote path.

One deliberate narrowing: the deadline is only armed when a last-good snapshot exists. With no snapshot the call already returns `false`, so a retry costs nothing a caller can observe and is the only route back — backing off there would just extend a cold-start outage the next call might have cleared. That edge stays eager, and `network failure short-circuits to false without caching the failure` still pins it.

Two tests added in `SwapKitProviderCacheTest`: the stale snapshot is re-served without re-hitting `/providers` while the window stands, and the refresh resumes — picking up a recovered endpoint — once it lapses.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/SwapKitProviderCache.kt
```diff
@@ -67,42 +92,72 @@ internal class SwapKitProviderCacheImpl @Inject constructor(private val api: Swa
...
+                fetchedAtMillis = now
                 chains
             } catch (e: CancellationException) {
                 throw e
             } catch (e: SwapKitError) {
                 // Expected transport/decoding failure from the SwapKit proxy — already classified
-                // at the API layer. Fail-closed so SwapKit is skipped until the next refresh.
-                null
+                // at the API layer. Serve the last good answer if there is one.
+                backOffAndServeLastGood(now)
```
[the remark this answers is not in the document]
aminsato: Valid, and fixed in `fd555e010`.

Confirmed `e` is dead in that branch — the failure is already classified at the API layer, and the branch only decides whether a last-good snapshot can stand in. Bound as `_`, which matches the repo convention and stops it reading as a swallowed exception. The two neighbouring catches keep their bindings: `CancellationException` rethrows `e`, and the generic one logs it.

Took the test nitpick from the review body in the same commit, with two corrections.

**The count is 3, not 2.** The fetch that seeds the snapshot before `invalidate()` is itself an `api.providers()` call, so the sequence is seed, failed refresh, eager retry.

**The mechanism is not the one cited.** The suggestion credits the narrowing in `backOffAndServeLastGood` (deadline armed only when a snapshot exists). I checked that by mutation: arming `retryAfterMillis` unconditionally, before the null check, leaves the new assertion green. It has to — `isServable` gates *both* its branches on `fetchedAtMillis != 0L`, and `invalidate()` cleared it, so no armed deadline is reachable on that path whatever `backOffAndServeLastGood` does. The narrowing is belt-and-braces here; what actually keeps the post-invalidate edge eager is the `fetchedAtMillis` gate.

What the assertion does bite on is ungating the retry branch (`(fresh && fetched) || now < retryAfterMillis`) — that turns it red, along with the existing cold-start test, and it is the regression worth guarding: SwapKit stranded off for the full window with nothing to serve in the meantime. The comment credits the gate rather than the narrowing, so the next reader is not misled about which line is load-bearing.

Still worth adding despite the overlap with `network failure short-circuits to false without caching the failure` — that one covers the never-fetched edge, and the KDoc claims two.

rkokhatskyi (approved):
LGTM

PR #5723, opened 2026-08-26
feat(home): rebuild the promo carousel on Figma's 2026 banner

### LINKED ISSUE #5642: [Update] Home promo banners: match Figma NEW 2026
Figma "Banners NEW 2026" — home carousel.

https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=68734-97320

Target (caption / title). Whole banner is the tap target — no extra CTA button.

| Banner | Caption | Title |
| Upgrade | Sign faster than ever before | Upgrade your vault now |
| Rujira | Stake & earn | Rujira staking live now |
| Follow X | Vultisig is building with you | Follow us on X |
| Backup | Back Up Your Vault | Request full vault custody |
| Referral | Don't miss rewards | Add your referral code |
| Buy $VULT | Buy $VULT | And save on swap fees |
| Kamino | New on Solana | Start earning with Kamino |

Also: 3D art on the right, icon tile on the left, glass close.

New vs today: Rujira, referral, Kamino. Backup title/subtitle as above (iOS subtitle is still "Keep access safe, store online").

Siblings: iOS https://github.com/vultisig/vultisig-ios/issues/5206 · Windows https://github.com/vultisig/vultisig-windows/issues/4680

aminsato:
Closes #5642

The home carousel still showed the three tall banners with their own CTA buttons. Figma's "Banners NEW 2026" replaced that with a compact 81 dp card whose whole surface is the tap target, and added four promos Android never had.

## Design

[Figma — Banners NEW 2026](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=68734-97320), the target all seven are built against:

Figma: Banners NEW 2026, all seven variants

## What this does

One `HomeBanner` shell draws all seven. The surface is `surface1` with a per-promo tint at 69 % over the trailing half, a 24 dp radius and a `borders/light` edge; a 41 dp icon tile leads, caption and title follow, and a 3D render sits in a fixed window pinned to the trailing edge under a glass close.

| Banner | Caption / Title | Shown when | Opens |
|---|---|---|---|
| Kamino | New on Solana / Start earning with Kamino | vault has Solana | Solana DeFi (opens on Earn) |
| Upgrade | Sign faster than ever before / Upgrade your vault now | vault is GG20 | migration onboarding |
| Rujira | Stake & earn / Rujira staking live now | vault has THORChain | THORChain DeFi |
| Follow X | Vultisig is building with you / Follow us on X | always | X, out of app |
| Backup | Back Up Your Vault / Request full vault custody | vault not backed up | backup password request |
| Referral | Don't miss rewards / Add your referral code | no referred code yet | referred-code form |
| Buy $VULT | Buy $VULT / And save on swap fees | always | swap to VULT |

Order is Figma's, and the two chain-gated promos are gated because they open that chain's DeFi screen — on a vault without it there would be nowhere for them to go. The separate CTA buttons are gone, so `upgrade_banner_upgrade_now`, `invite_to_x_banner_button` and `buy_vult_banner_button` are deleted from all ten locale files, and the three old full-bleed backgrounds are deleted with them.

`VaultAccountsUiModel` carries `banners: List<HomeBannerType>` instead of three booleans, so adding a promo no longer means widening the state, the screen signature and a `combine` in three places.

## Two things that had to move for an 81 dp card to work

**The artwork cannot take part in layout.** It is a 125 dp render inside an 81 dp card, so it sits in a `matchParentSize()` layer — laid out normally it sizes the card to itself, which is exactly what happened first time round (the card measured 122 dp). Both the window and the source need `requiredSize`, not `size`: `size` coerces into the incoming constraints and silently shrinks Figma's crop back to its container.

**`VsPager` never shrank.** `maxHeight` started at a hardcoded `140.dp` and only ever grew (`max(new, old)`), so 81 dp banners would have sat in a 140 dp box with 59 dp of dead space. It is derived from the measured pages now, and the map is cleared when the page set changes so stale indices cannot pin the pager to a page that is gone.

## Dismissal

Backup takes a new `DismissPolicy.Session`, held in memory on the (singleton) repository. iOS gives that reminder a session rule and this follows it: a fifteen-day TTL on a vault that is still not backed up would suppress the one banner whose condition is a standing risk rather than an announcement. The other three new banners take the fifteen-day TTL their siblings already use. This brushes #5669, which scopes dismiss-return policy out of the restyle — flagging it rather than hiding it.

## Assets

The seven renders come from the exported originals (2160²) downscaled to 512² WebP: **110 KB for seven, against 132 KB for the three they replace**. Six icons are new vector drawables; the Rujira one reuses `ruji.webp`, which already ships the same glyph.

## On matching Figma

The CSS that `get_design_context` emits for a cropped image fill is lossy, so every artwork transform was solved numerically instead — fitting each source asset against Figma's own rendered node — and the fitted scale and offsets agree with the exported values to within about a dp on all seven. The same fit put the artwork's actual blur spread at roughly half a dp, not the 2 px the export names: Compose's blur radius maps to a much larger sigma, and passing `2.dp` turned the render to mush.

The glass close is Figma's, not the app's previous circle button: a 40 dp disc inset 10 dp, filled white at 1 %, with the card's own drawing re-drawn blurred and clipped to the circle behind it — the fill alone is invisible, the blur is what makes it read as glass.

## Test plan

- `:app:testDebugUnitTest` — 37 in `VaultAccountsViewModelTest`, 0 skipped, 11 of them new: per-banner eligibility (GG20, Solana, THORChain, referred code), Backup dropping live when the vault is backed up, dismissal recording, and carousel order.
- `:data:testDebugUnitTest` — 10 in `PromoBannerDismissalRepositoryTest`, 3 new for `Session` (hides for the process, does not survive a relaunch, independent of the others).
- `:app:lintDebug` clean.
- All seven banners rendered on an emulator through `PreviewActivity --es screen banner` and compared against the Figma nodes; no screenshots attached here, per standing preference.

## Manual check

Home → the carousel should show the promos this vault qualifies for, in the order above. Tap anywhere on a card to follow it; the close disc dismisses. Switch vaults and the set should re-evaluate. Back the vault up on another screen and the Backup card should disappear without leaving home.

## Summary by CodeRabbit

* **New Features**
  * Added a unified home banner carousel for upgrades, backups, referrals, staking, and earnings.
  * Banners now support consistent tap and dismiss interactions based on wallet status and available features.
  * Added localized banner text and new visual icons.
* **Bug Fixes**
  * Improved pager sizing to adapt to visible content.
  * Banner dismissals now behave consistently during the current app session.
* **Tests**
  * Expanded coverage for banner ordering, eligibility, interactions, and dismissal behavior.


app/src/main/java/com/vultisig/wallet/ui/screens/v2/home/pager/banner/HomeBanner.kt
```diff
@@ -0,0 +1,447 @@
...
+    val spec = banner.spec()
+    val backdrop = rememberGraphicsLayer()
+    val frost = rememberGraphicsLayer()
+    val glassTint = Theme.v2.colors.neutrals.n50
+
+    Box(modifier = modifier.fillMaxWidth()) {
+        Box(
+            modifier =
+                Modifier.fillMaxWidth()
+                    .heightIn(min = BannerMinHeight)
```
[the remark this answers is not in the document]
aminsato: Valid — fixed in 215668c63.

The mechanism is right: with `heightIn(min = 81.dp)` and `padding(vertical = 20.dp)` the copy only had the 41 dp the icon tile occupies. Caption + one title line is 38 dp and fits; two lines need 58, so the card grew to ~98 dp and took the pager with it. de/es/hr/it/ru all wrap at this width, so it was reachable in shipping locales, not just in theory.

I did not take the clip, though — that drops half the sentence in those locales. iOS pins the frame at 81 and pads horizontally only, so the text box gets the whole card and two lines fit inside a fixed 81. Same here now: `height(BannerHeight)` on the card, `fillMaxSize().padding(horizontal = 16.dp)` on the content row, still centred vertically. A one-line banner is pixel-identical to before — a 41 dp tile centred in 81 dp leaves the same 20 dp of clearance the padding did.

app/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -1256,8 +1255,15 @@
     <string name="send_tx_overview_add_to_address_book">添加到地址簿</string>
     <string name="invite_to_x_banner_title">Vultisig 正在与您一起构建。</string>
     <string name="invite_to_x_banner_desc">在 X 上关注我们</string>
-    <string name="invite_to_x_banner_button">关注 %s</string>
     <string name="buy_vult_banner_desc">并节省交换费用</string>
+    <string name="rujira_banner_caption">质押并赚取收益</string>
+    <string name="rujira_banner_title">Rujira 质押现已上线</string>
+    <string name="backup_banner_caption">备份您的保险库</string>
+    <string name="backup_banner_title">申请完整保管您的保险库</string>
```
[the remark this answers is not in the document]
aminsato: Leaving this one as is.

The English source string is `Request full vault custody`, not "take full custody" — so 申请 ("request / apply for") is carrying a verb that is actually there. The proposed 完全掌控您的保险库 would drop it.

The string is also verbatim what iOS ships for this banner (`backupBannerSubtitle` in `zh-Hans.lproj`), and this PR is a port of that screen, so changing it here would put the two platforms out of sync on the same banner.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -443,17 +469,46 @@ constructor(
...
+        when (banner) {
+            HomeBannerType.UpgradeVault -> migrate()
+            HomeBannerType.BuyVult -> buyVult()
+            HomeBannerType.BackupVault -> backupVault()
+            // Straight to the chain's DeFi screen: Solana opens on Earn and THORChain on staking,
+            // which is what each banner advertises.
+            HomeBannerType.KaminoEarn ->
+                viewModelScope.launch {
+                    navigator.route(
+                        Route.ChainDashboard(ChainDashboardRoute.PositionSolana(vaultId = vaultId))
```
rkokhatskyi, line 504: `KaminoEarn` and `RujiraStaking` open the chain's wallet token list instead of its DeFi screen because `ChainDashboardViewModel` rebuilds the rendered route from `activeCryptoConnectionFlow` at `ChainDashboardUiModel.kt:143`, and the carousel only shows while that flow reads `Wallet`. Both taps need `setActiveCryptoConnection(Defi)` first

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -443,17 +469,46 @@ constructor(
...
+                    )
+                }
+            HomeBannerType.RujiraStaking ->
+                viewModelScope.launch {
+                    navigator.route(
+                        Route.ChainDashboard(ChainDashboardRoute.PositionTokens(vaultId = vaultId))
```
rkokhatskyi: `PositionTokens` lands on the Bonded tab because `ThorchainDefiPositionsViewModel.kt:95` seeds `selectedTab = R.string.defi_tab_bonded`, while Rujira staking lives on Staked, so the card titled `Rujira staking live now` opens the node bonding list

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
```diff
@@ -210,33 +211,57 @@ constructor(
...
-                            !buyVultDismissed,
-                        )
+                // Re-evaluated per vault so an expired TTL, a newly enabled chain or a referral
+                // code entered elsewhere are all picked up on vault switch / home re-entry.
+                val vault = withContext(ioDispatcher) { vaultRepository.get(vaultId) }
+                val chains = vault?.coins?.mapTo(mutableSetOf()) { it.chain }.orEmpty()
+                val hasReferralCode =
+                    withContext(ioDispatcher) {
+                        !referralCodeSettingsRepository
+                            .getExternalReferralBy(vaultId)
```
rkokhatskyi, line 222: `hasReferralCode` is read once here and only `loadData` re-runs it, which `collectLastOpenedVault` gates behind `distinctUntilChanged` at `VaultAccountsViewModel.kt:292`, so the Referral card keeps asking for a code the user just entered through that same card

rkokhatskyi (dismissed):

johnnyluo (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Confirmed — fixed in 2f4356af5.

Traced it exactly where you said: `collectActiveRoute` rebuilds the rendered route from `activeCryptoConnectionFlow`, so `PositionSolana` / `PositionTokens` were overwritten with `Wallet(chainId)` on the first emission. Every other caller of a DeFi position route is `openAccount`, which only runs when the toggle already reads `Defi`, so the banners were the first entry point from the wallet side and the only ones that hit this.

Both taps now `setActiveCryptoConnection(Defi)` before `navigator.route`. Called on the repository rather than through `setCryptoConnectionType`, which would also fire a network `loadDeFiBalances` on a home screen that is about to go to the back stack; `collectCryptoConnectionType` picks the change up for the UI state either way.

One consequence worth flagging: the toggle is a process-wide singleton, so home comes back on the DeFi tab after either banner. That already happens when the toggle is flipped from the chain dashboard, so it is consistent rather than new, but it is a behaviour change for these two taps.

Covered by `Kamino banner switches to DeFi before opening Solana positions`.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Confirmed — fixed in 2f4356af5. `ThorchainDefiPositionsUiModel.selectedTab` seeds `R.string.defi_tab_bonded`, so the card titled "Rujira staking live now" opened the node bonding list.

`ChainDashboardRoute.PositionTokens` now carries an optional `tab`. Two details it needed:

- the dashboard rebuilds that route on every connection emission, which drops anything the caller put on it, so the tab is held in `ChainDashboardViewModel` and put back on each rebuild;
- the tab is applied once in the view model. The screen leaves and re-enters composition every time the wallet / DeFi toggle flips, and re-seeding on each entry would throw away whichever tab the user had picked since.

The banner passes `DeFiTab.STAKED`; nothing else passes a tab, so every other entry still opens on Bonded.

Worth knowing: iOS lands on Bond for this same banner — `getDefiPositionTypes()` returns `[.bond, .stake, .liquidityPool]` for `.thorChain` and `DefiChainMainViewModel` does `selectedPosition = positionTypes.first`. So this puts Android ahead of iOS rather than in line with it. I can raise the matching iOS ticket if you want the two to agree.

Covered by `Rujira banner opens THORChain positions on the Staked tab`.

app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
aminsato: Confirmed — fixed in 2f4356af5. The comment above that block claimed the code was picked up on "vault switch / home re-entry", but only the first half was actually wired: `loadData` is the sole caller, and `collectLastOpenedVault` gates it behind `distinctUntilChanged`, so returning from the referral screen re-ran nothing.

`onScreenResumed` now re-evaluates banner eligibility before its DeFi early-return. Home already resumes on the way back from that screen, and the work is a vault read plus a prefs read on the IO dispatcher. That route rather than a new flow on the repository: `getExternalReferralBy` is backed by encrypted `SharedPreferences` with no change stream, and the same resume pass also picks up a chain enabled a screen deeper, which the chain-gated promos read.

It works in both directions — removing the code offers the card again on the next return, which a one-way "hide once entered" flag would not have done. Both are covered: `onScreenResumed drops the Referral banner once a code has been entered` and `... restores the Referral banner once the code is removed`.

PR #5726, opened 2026-08-26
fix(ton): drop IGNORE_ACTION_PHASE_ERRORS, sign explicit MAX amount

### LINKED ISSUE #5685: [Fix] TON: send mode still sets IGNORE_ACTION_PHASE_ERRORS after the #5249 status fix [TON / S1]
**Problem:** The status side was fixed in PR #5263 (we now detect action-phase failure), but the send side still sets `IGNORE_ACTION_PHASE_ERRORS` on every transfer — with a comment saying it prevents retry loops.
**Why it matters:** We now correctly *report* the failure we ourselves told the contract to ignore. Dropping the flag on app-initiated sends makes the transfer fail atomically instead of half-executing.
**Fix:** Remove the flag from `calculateSendMode` for app-built sends (keep it only if a dApp explicitly requests it).
**Receipts:** `data/.../crypto/TonHelper.kt:172-182`.
Sibling of vultisig-sdk#2181 (S1 item 1).

aminsato:
Closes #5685

## Summary

- **Drops `IGNORE_ACTION_PHASE_ERRORS` from all three TON builders** — native, jetton and TonConnect. With the flag set, a transfer the wallet contract can't pay for is skipped while the transaction still lands un-aborted: seqno consumed, no funds moved. #5263 taught the status side to detect exactly that, so the app was reporting a failure it had itself asked the contract to ignore.
- **MAX now signs the explicit amount** instead of `amount=0` + `ATTACH_ALL_CONTRACT_BALANCE`. The form displayed `balance − fee` but signed a full sweep, so ~0.045 TON more left the wallet than the user approved. The reserved fee now doubles as the account's storage reserve.
- `calculateSendMode` is gone; a single `SEND_MODE` constant feeds all three builders, which also folds in the jetton path that had been hardcoding its own mode.
- `sendMaxAmount` stays in the payload but no longer affects signing — same as iOS, whose `buildTransfers` destructures it away.
- Reworded the three comments asserting "every Vultisig TON send is signed with `IGNORE_ACTION_PHASE_ERRORS`", which is no longer true. The status-side checks stay — this path still reads transactions the app didn't sign: older sends, and sends built by peers or other clients.

## Cross-platform coordination

Both changes alter the signed bytes, and every co-signer recomputes the signing hash from the payload, so this only works if the platforms move together:

- **iOS already shipped this** in vultisig-ios#5259 (merged 2026-08-24). Android↔iOS TON co-signing is therefore broken *today* — this PR repairs it.
- **Windows / extension still diverge** until vultisig-sdk#2181 (S1 item 1) lands and the `@vultisig/core-chain` pin moves off 2.39.0. Same window iOS is already sitting in.

One note for the issue as written: the "keep it only if a dApp explicitly requests it" clause isn't implementable — `TonMessage` in commondata carries only `to` / `amount` / `payload` / `state_init`, no send mode, and iOS hardcodes pay-fees-separately for TonConnect too.

## Goldens

The Android and iOS `ton.json` fixtures are content-identical, and their pre-change hashes agreed byte for byte. So the re-recorded hashes here are **copied from the iOS commit** rather than freshly generated — `sendTONTest` passing is a byte-level check that the two platforms still produce the same signing hash, not just similar-looking code.

## On-chain verification

A MAX send from a build carrying this change, on mainnet: [`be28ed84…04970`](https://tonviewer.com/transaction/be28ed84d0e339659b08e6bc4756618ffa4c19940572cd661d3d7cf596804970) (wallet v4, seqno 41, 2026-08-26 13:03 UTC, `success: true`).

The decoded signed body carries **`mode: 1`** — pay-fees-separately alone. No `+2` (`IGNORE_ACTION_PHASE_ERRORS`) and no `128` (`ATTACH_ALL_CONTRACT_BALANCE`), which is the whole change visible in the bytes the network actually accepted.

The amounts confirm the MAX half:

| | nanoton | TON |
|---|---|---|
| Balance before | 8 775 999 527 | 8.775999527 |
| Signed amount (`grams`) | 8 725 999 527 | 8.725999527 |
| Fee actually charged | 458 683 | 0.000458683 |
| **Balance after** | **49 541 317** | **0.049541317** |

The signed amount is `balance − 0.05 TON` to the nanoton — the displayed MAX, sent verbatim rather than swept. The wallet kept **0.049541317 TON**; under the old mode-128 sweep it would have been left at ≈0, with ~0.0495 TON more leaving than the form showed. Real fees came to 0.00046 TON, so the 0.05 reserve is comfortable.

## Test plan

- [x] `sendTONTest` green against iOS's re-recorded hashes (native, jetton, memo/stake vectors)
- [x] New `tonSendModeIsPayFeesSeparatelyOnEveryBuilder` — sweeps every fixture, asserts mode 1 on each message
- [x] New `tonMaxSendSignsExplicitAmountNotSweep` — `sendMaxAmount=true` signs the explicit amount with mode 1
- [x] Existing TonConnect tests green, now also asserting the bare mode
- [x] 50 `:data` JVM TON tests green, 0 failures
- [x] Mainnet MAX send signs the displayed amount with mode 1 and leaves the reserve behind — see On-chain verification above
- [x] Mainnet native send completes keysign, broadcasts and confirms
- [ ] Mainnet jetton + TonConnect sends complete keysign and arrive
- [ ] TON send co-signed with an iOS build carrying vultisig-ios#5259

### DIFF
--- a/app/src/androidTest/assets/ton.json
+++ b/app/src/androidTest/assets/ton.json
@@ -29,7 +29,7 @@
       "vault_public_key_ecdsa": "023e4b76861289ad4528b33c2fd21b3a5160cd37b3294234914e21efb6ed4a452b",
       "lib_type": "DKLS"
     },
-    "expected_image_hash": ["aefa16a3825645bcfcf305be112bc2da400d213cb8c20d87e0e43d4eb214a5f8"]
+    "expected_image_hash": ["db6bfdb5473ece22c91a7a497c83e8a5d9009a03ad820f46839472c89f9f9235"]
   },
   {
     "name": "Send USDT TON",
@@ -63,7 +63,7 @@
       "vault_public_key_ecdsa": "023e4b76861289ad4528b33c2fd21b3a5160cd37b3294234914e21efb6ed4a452b",
       "lib_type": "DKLS"
     },
-    "expected_image_hash": ["44cbfbae7924a26eec5e6338370ff5d1dbde3fe63f67b1e8937cff1fc28aa2f3"]
+    "expected_image_hash": ["145673a02a4f1b091d6f9a34bf84490309ce6ef70ae9a2c2de0e5d9c1299cca9"]
   },
   {
     "name": "TON Stake with memo",
@@ -95,6 +95,6 @@
       "vault_public_key_ecdsa": "023e4b76861289ad4528b33c2fd21b3a5160cd37b3294234914e21efb6ed4a452b",
       "lib_type": "DKLS"
     },
-    "expected_image_hash": ["299ecf1da0682c338130b7f3f890484100e41c8f0d2b024b1fd0214df6e310dd"]
+    "expected_image_hash": ["f132a896ec6ea24c25309f2c220da4efcbd99c210323e266a8dafa8c5e290b78"]
   }
 ]
--- a/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
@@ -209,6 +209,85 @@ class ChainHelpersTest {
     }
 
     /**
+     * Every app-built TON transfer carries mode PAY_FEES_SEPARATELY and nothing else - in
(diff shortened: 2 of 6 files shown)


rkokhatskyi (approved):
LGTM

PR #5727, opened 2026-08-26
Remove Kujira chain support

### LINKED ISSUE #5724: Kujira (KUJI) chain appears permanently offline — remove support
## Problem
Kujira wound down operations in 2025 — the team requested exchanges suspend KUJI trading (e.g. MEXC halted deposits/withdrawals/trading May 4, 2025 UTC), as part of a broader wave of Cosmos-ecosystem project shutdowns. No working public RPC/LCD remains, so balances, sends, swaps, and staking on Kujira can no longer function for users.

Verified no working public RPC/LCD remains:
- `kujira-rpc.polkachu.com`, `lcd-kujira.whispernode.com`, `kujira-rpc.ibs.team` — DNS records deleted (NXDOMAIN)
- `rpc-kujira.starsquid.io` — resolves, but the RPC port doesn't respond
- PublicNode's Kujira gateway — 404s on every path
- `ping.pub/kujira` — 404s (no longer tracked by this standard Cosmos explorer)

## Expected Behavior
Kujira should no longer be offered as a chain to add/use, since it cannot function.

## Steps to Reproduce
1. Add Kujira as a chain in a vault
2. Attempt to fetch balance, send, swap, or stake
3. Requests fail — no live RPC/LCD endpoint exists

## Suggested Fix
Remove Kujira chain support (chain registry entry, RPC/LCD config, and any UI wiring that assumes it's usable — deposit/IBC routing, DeFi/staking display, address-prefix detection, etc.).

aminsato:
Closes #5724

## Why

Kujira wound down operations in 2025 and no working public RPC/LCD remains, so balances, sends, swaps and staking on the chain cannot function. Re-verified while writing this:

- `kujira-rpc.polkachu.com`, `lcd-kujira.whispernode.com`, `kujira-rpc.ibs.team` — NXDOMAIN
- `rpc-kujira.starsquid.io` — resolves, RPC port does not respond
- PublicNode's Kujira gateway — 404 on every path

Offering the chain to add is worse than not offering it, so this removes it.

## What

Dropping the `Chain.Kujira` enum entry is what makes it disappear from the add-chain list and the address-book network picker (both are built off `Chain.entries`), and from `Coins.coins`, so `nativeTokens` stops offering KUJI. Everything that hung off it goes with it:

| Area | Removed |
|---|---|
| Chain registry | enum entry, `coinType`, `canSelectTokens`, `isSwapSupported`, `isDepositSupported`, `ticker()`, `swapAssetName()`, `cosmosNativeDenom`, `maxMemoCharacters`, ECDSA pubkey mapping |
| Network config | `kujira-api.polkachu.com` default + custom-RPC entry, `kaiyo-1` SwapKit id, `finder.kujira.network` explorer links, tx-status poll config |
| Signing / fees | both `SigningHelper` branches, Cosmos fee + simulation entries, IBC denom-trace branch |
| Tokens | the `Coins.Kujira` object (13 tokens), `SearchKujiraTokenUseCase` + DI binding + its dispatch branch and now-dead `factory/` canonicalization |
| Deposit / IBC | the Kujira IBC channel table, the Gaia→Kujira channel, `TransferIbc` on the deposit form, Kujira as an IBC destination |
| Swap | swap-provider entry, THORChain memo prefix, limit-order address validator and dust floor |
| UI | Maya pool-prefix mapping, chain logos |

### Kept deliberately

`Coins.ThorChain.KUJI`/`RKUJI` (RUJI merge tokens) and `Coins.GaiaChain.KUJI`/`rKUJI` (IBC balances) both live on chains that still work — removing them would hide real holdings. `kuji.webp` stays for them; only `kuji_mono`, `mnta` and `auto_token_kujira` were left unreferenced and deleted.

`CosmosBalance.isValidKujiraFactoryDenom` was renamed `isValidFactoryDenom` rather than removed — it is load-bearing for Osmosis' `factory/osmo1…/ulvn`.

## The part that isn't just deletion

Removing an enum constant strands rows that were saved under it, and `Chain.fromRaw` throws for those. `VaultRepository` and the key-import paths already caught it. Two did not:

- **`AddressBookRepository.getEntries()`** mapped every row, so a single saved Kujira contact would have thrown and blanked the entire address book.
- **`DefaultChainsRepository`** caught the throw but fell back to the *whole* stock 5-chain list, silently resetting a user's default-chain selection.

Both now skip just the unknown id. Covered by new tests.

`scripts/generate-tokens.js` still mapped `kujira → Chain.Kujira`, so a regeneration would have written a `Coins.kt` that no longer compiles. Without the entry the script skips the chain via its existing unknown-chain warning.

## Cross-platform note

iOS, the extension and Windows all still ship Kujira — Android is first here, which inverts the usual parity direction. Consequence worth knowing: a peer on iOS/desktop can still initiate a Kujira keysign, and this device will now fail to parse that payload rather than co-sign it. Correct given the chain is dead, but the other repos probably want matching tickets.

## Test plan

- `./gradlew :data:test :app:testDebugUnitTest` — green
- `./gradlew :app:lintDebug` — green (CI gate, `abortOnError = true`)
- `./gradlew :app:assembleDebug` — green
- `./gradlew :data:ktfmtCheck :app:ktfmtCheck` — green
- New: 2 tests in `AddressBookRepositoryImplTest` (retired-chain row dropped from the list, reads back absent), 3 in a new `DefaultChainsRepositoryImplTest` (retired id dropped without discarding the rest, all-retired falls back to defaults, unreadable value falls back to defaults). Verified in the XML reports that all 5 actually ran rather than being skipped.
- Removed the `sendKUJIRATest` instrumented fixture; `cosmos.json` / `thorchain.json` keep their KUJI cases since those tokens remain.

Manual: open a vault that had Kujira enabled → the chain is gone from the list, the rest of the vault loads normally, the address book still opens with its other entries intact, and the default-chain selection keeps its non-Kujira picks.

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed support for the Kujira blockchain and associated assets, transfers, staking, swaps, transaction status, explorer links, and custom RPC endpoints.
  * Kujira addresses and transactions are no longer supported.
  * Imports and deep links referencing retired chains now show an error or safely fall back when appropriate.

* **Bug Fixes**
  * Address-book entries for retired chains are safely ignored.
  * Invalid saved chain selections fall back to the default chain list.


app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
```diff
@@ -1055,7 +1036,6 @@ class ChainHelpersTest {
                 QBTC_JSON_FILE to "sendQBTCTest",
                 SUI_JSON_FILE to "sendSUI",
                 TRON_JSON_FILE to "sendTronTest",
-                KUJIRA_JSON_FILE to "sendKUJIRATest",
                 CARDANO_JSON_FILE to "sendCardano",
```
rkokhatskyi, line 1039: `EXPECTED_CASE_COUNT` at `ChainHelpersTest.kt:1054` still reads 88 while the remaining fixtures now sum to 85, because `kujira.json` carried 3 cases. `fixtureCorpusMatchesDeclaredFileSet` therefore fails on every `app:connectedDebugAndroidTest` run

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
```diff
@@ -47,7 +47,6 @@ enum class Chain(val raw: ChainId, val standard: TokenStandard, val feeUnit: Str
     Zcash("Zcash", UTXO, "ZEC/vbyte"),
     Cardano("Cardano", UTXO, "Lovelace"),
     GaiaChain("Cosmos", COSMOS, "uatom"),
```
rkokhatskyi, line 49: `args.chains.mapNotNull` at `KeygenViewModel.kt:578` silently drops `Kujira` from a peer's key import list, while `KeyImportChainsSetupViewModel.swift:125` still offers it, so iOS opens a keygen for that chain which this device never joins and both sides fail after root keygen

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -1024,7 +1024,6 @@ private fun mayaPoolChainPrefixToChain(prefix: String): Chain? =
         "BTC" -> Chain.Bitcoin
         "ETH" -> Chain.Ethereum
         "DASH" -> Chain.Dash
-        "KUJI" -> Chain.Kujira
         "MAYA" -> Chain.MayaChain
```
rkokhatskyi, line 1029: This table also omits `THOR`, `ADA` and `ZEC`, which `/mayachain/pools` currently returns as Available, so those pools hit `else -> null` and price their asset leg at zero through `MayachainDefiPositionsViewModel.kt:845`, halving the card total and the header sum

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
```diff
@@ -47,7 +47,6 @@ enum class Chain(val raw: ChainId, val standard: TokenStandard, val feeUnit: Str
...
-    Kujira("Kujira", COSMOS, "ukuji"),
     Dydx("Dydx", COSMOS, "adydx"),
     Osmosis("Osmosis", COSMOS, "uosmo"),
```
rkokhatskyi, line 51: `Chain.fromRaw(preSelectedChainId)` at `TokenNetworkSelectionDelegate.kt:290` is unguarded, so a `vultisig://send?assetChain=Kujira&toAddress=...` link that worked before now crashes the app, while the sibling arm at `SendFormGraph.kt:390` compares chain ids as strings and degrades cleanly

rkokhatskyi (changes requested):

johnnyluo (approved):
LGTM

app/src/androidTest/java/com/vultisig/wallet/data/blockchain/ChainHelpersTest.kt
aminsato: Confirmed and fixed in 2d3050cbd — `kujira.json` carried 3 of the 88, and the remaining fixtures sum to 85. The CI failure said exactly that (`expected:<88> but was:<85>`); I had only run the unit suites locally, which do not touch the corpus. `EXPECTED_CASE_COUNT` is now 85, with a note that the other two corpora still carry the file, so the sync check reports it as missing here rather than as a hash disagreement.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
aminsato: Real, and the worse half is that it fails silently: the initiator opens one keygen session per chain, routed by `chain.raw`, and waits for this device in each — so a skipped chain leaves that session a party short and both sides die on a timeout that names nothing. Verified `keyImportEnabledChains` at `Chain.swift:247` excludes only cardano / thorChainChainnet / thorChainStagenet / polygonV2 / qbtc, so iOS does still offer `.kujira`. Same holds for an older Android initiator, so it is version skew rather than an iOS quirk.

Android cannot complete that ceremony either way once the enum is gone, so 2d3050cbd makes the joining device refuse up front and name the chain instead of hanging. iOS still needs the matching removal from `keyImportEnabledChains` for the ceremony to stop being offered at all — happy to open that PR against vultisig-ios if you want it tied to this one.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Checked against `mayanode.mayachain.info/mayachain/pools`: `THOR.RUNE`, `ADA.ADA` and `ZEC.ZEC` are all Available today, and `KUJI` is gone from the list entirely — which is a nice independent confirmation that the line this PR removed was dead. Each of the three resolved to no chain, so `assetCoin` could never match and the leg fell to `FiatValue.ZERO`, halving the card and the header.

Added in 2d3050cbd, with the mapping parameterized over the three pools in `MayachainDefiPositionsViewModelTest` — it asserts $0.40 and fails at $0.20 without the fix. Pre-existing rather than caused by this PR, so say the word if you would rather it went out on its own.

data/src/main/kotlin/com/vultisig/wallet/data/models/Chain.kt
aminsato: Confirmed — `Chain.fromRaw` is `entries.first { … }`, and the id reaches it raw from `DeepLinkHelper.kt:43` → `VaultListViewModel.kt:155` → `Route.Send.chainId` → `SendFormGraph.initialize:382`, which is not a suspending path, so an old `vultisig://send?assetChain=Kujira&toAddress=…` took the form down on the way in.

2d3050cbd resolves it through the new `Chain.fromRawOrNull` and falls through to matching the address against the chains that do exist, so it degrades the same way the `SendFormGraph.kt:390` arm you point at does. Regression test in `SendFormViewModelInitTest`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AddressBookRepository.kt
```diff
@@ -41,7 +42,7 @@ constructor(private val addressBookEntryDao: AddressBookEntryDao) : AddressBookR
             } else {
                 addressBookEntryDao.getEntry(lookup.chainId, address)
             }
-        return entity?.toAddressBookEntry()
+        return entity?.toAddressBookEntryOrNull()
```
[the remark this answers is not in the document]
aminsato: The inconsistency is real, though not reachable today: both `Route.AddressEntry` call sites pass a live `chain.id` (`AddressBookViewModel.kt:147` from a list `getEntries` has already filtered, and `KeysignViewModel.kt:1168` from the transaction target), so no retired id gets that far. Closed it anyway in 2d3050cbd — `entryExists` now returns false for a chain id no entry resolves for, so it cannot disagree with `getEntry`. Test added in `AddressBookRepositoryImplTest`.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/AddressBookRepository.kt
```diff
@@ -77,8 +78,17 @@ constructor(private val addressBookEntryDao: AddressBookEntryDao) : AddressBookR
...
-        AddressBookEntry(chain = Chain.fromRaw(chainId), address = address, title = title)
+    // Rows outlive the chains they were saved for: a retired chain leaves its entries behind, and
+    // Chain.fromRaw throws for them. Drop those rows rather than fail the whole address book.
+    private fun AddressBookEntryEntity.toAddressBookEntryOrNull(): AddressBookEntry? {
+        val chain =
+            runCatching { Chain.fromRaw(chainId) }
+                .onFailure {
+                    Timber.w(it, "Dropping address book entry on unknown chain %s", chainId)
+                }
+                .getOrNull() ?: return null
```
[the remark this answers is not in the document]
aminsato: Fixed in 2d3050cbd, though by removing the catch rather than narrowing it: `Chain.fromRawOrNull` is now the lookup for readers that expect ids to outlive their entries, and `fromRaw` delegates to it — still `NoSuchElementException`, now with the id in the message, so the callers that catch that type are unaffected. Both sites here and in `DefaultChainsRepository` use the nullable lookup directly, so nothing broad is caught in either.

On the `SearchTokenUseCase.kt:46` finding in the review body: that one is not reachable either — `SwapQuoteManager` passes `srcNativeToken.chain.id`, and `CustomTokenViewModel` would already have thrown one line earlier at `Chain.fromRaw(chainId).logo`. Made it total regardless, so it answers "not found" instead of throwing out of its caller.

rkokhatskyi (approved):
LGTM

Issue #5023, opened 2026-06-23
THORChain/Maya Auto swaps rejected at quote time when price impact exceeds 1%


johnnyluo:
## Problem
THORChain/Maya **Auto** swaps are rejected at quote time when price impact exceeds 1%, with the node error:

```
emit asset X less than price limit Y
```

## Root cause
Android sends a hardcoded `tolerance_bps` default of 100 (1%) on every THORChain/Maya quote request (`DEFAULT_THORCHAIN_TOLERANCE_BPS`). The node gates `tolerance_bps` against the **single-swap (full-size) emit**, not the streamed output — so a 1% default is unreachable for any non-trivial swap and the quote is refused. The node only reports *streamed* slippage, so no protective-but-achievable default can be derived.

## Fix
Default `tolerance_bps` to 0 for Auto; when 0 (or null) the param is omitted, so the node sets no `LIM` and never rejects the quote. A user-set slippage still flows straight through. Slippage stays mitigated by the rapid→streaming upgrade (`STREAMING_SLIPPAGE_THRESHOLD_BPS`) and the node's own auto-streaming.

## References
- iOS fix: vultisig/vultisig-ios#4640
- Affected: `data/.../api/ThorChainApi.kt`, `data/.../api/MayaChainApi.kt`, `data/.../repositories/swap/SwapQuoteSource.kt`

PR #5734, opened 2026-08-27
fix(swap): stop calling the expected output a minimum payout

### LINKED ISSUE #5711: Swap overview labels the unenforced expected output "min. payout" when no tolerance/limit is in the memo
## Observed

Swap overview for a THORChain RUNE→TCY swap (v1.0.118, Auto slippage, no user slippage set) shows **"min. payout 3.24554248 TCY"**. The signed memo was `=:THOR.TCY:thor1uet6…qz79tu::va:40` — the limit field is empty — and the swap settled at **3.23347754 TCY** (`thorchain/tx/status/FB9AE98A3B93566A9F29B28A5580079CDD71FE8BA5E490464BF92F6710AB3278`), 0.37 % under the number labelled as the minimum.

## Cause

`DEFAULT_THORCHAIN_TOLERANCE_BPS = 0` (`SwapQuoteSource.kt:88`) is the decided default after #5023 (a node-side 1 % floor was unreachable), so unless the user sets slippage the quote carries no `tolerance_bps` and the memo has no `LIM`. That is documented in `ThorChainSwapQuoteRequest.kt:12-17`. But `VerifySwapScreen.kt:499` still labels `expectedDstDecimal` with `swap_form_min_pay` ("min. payout"), so the user is shown an enforced-looking floor that is not enforced.

## Expected

When no limit is in the memo, label the number "expected payout" (or "est. payout") and, if the Advanced sheet has a slippage value, show the real limit as "min. payout". The label should only say "min." when the memo actually carries one. Same label exists in transaction history (`transaction_history_min_payout_label`).

## Env

v1.0.118 tag, emulator, Fast Vault. Not a regression in 118 — the label predates it; the on-chain receipt above is what made the mismatch concrete.

aminsato:
Closes #5711

## Problem

The verify screen labelled the destination amount **"min. payout"** on every swap, but almost none of them enforce one. `DEFAULT_THORCHAIN_TOLERANCE_BPS = 0` (the decided default after #5023), so a THORChain/Maya quote left on Auto slippage carries no `tolerance_bps` and the node returns a memo with an empty `LIM` — the swap accepts any output. The receipt on the issue settled **0.37% under** the number the screen was calling the minimum.

The in-progress history card carried the same label over the same expected amount.

## What changed

- **`ThorchainMemoLimit.assertedLimit()`** (`:data`) reads the floor back out of the memo rather than re-deriving it from the tolerance we sent — the node owns the rounding and the streaming split, and that memo is what gets signed byte-for-byte (`SwapTransactionBuilder` passes `quote.data.memo` through; `UtxoHelper` writes it into the OP_RETURN unchanged). It refuses a non-swap action, a zero or absent limit, and any `LIM/INTERVAL/QUANTITY` triple it cannot read whole — failing toward showing no minimum, which is the safe direction for a label that promises one.
- **`signedMinimumOutput()`** gates that on native routes. Aggregators (1inch / KyberSwap / LI.FI / Jupiter / SwapKit) sign opaque calldata, so whatever floor their router enforces is not visible to us and none is claimed.
- **Verify screen**: the caption is now "expected payout" over the amount, and a real floor gets its own `min. payout: 3.21308705 TCY` line beneath the fiat. A `=<` limit order is the opposite case and keeps "min. payout": its displayed amount *is* the memo's LIM (`BuildLimitSwapTransactionUseCase` builds it from there), so there is no second number to spell out.
- **Co-signer** (`JoinSwapUiModelBuilder`) reads the floor out of the literal bytes it is being asked to sign, not the initiator's quote — the only way both devices can be shown to consent to the same guarantee.
- **Limit orders** are classified off the memo's target asset, not a bare parse of it. A `=<` order's displayed amount IS the floor its memo enforces, so `min. payout` over it is a guarantee — and on a cosigner the memo and the payload's destination coin are two independently decoded halves of one request. `signedLimitOrder()` drops an order whose target asset is confidently not the destination coin, one-sided in the same direction as the market path.
- **History card**: "expected payout" for market swaps. Telling the two apart there needs the limit-ness persisted, so the row now records `isLimitOrder` — a JSON payload field defaulting to `false`, no Room migration, and `false` is the direction that claims nothing. Rows written before this build read as market swaps.
- **Strings**: 3 new keys across all 10 locale files.

## iOS parity

This mirrors what iOS already ships — `ThorchainMemoLimit`, `SwapCryptoLogic.signedMinimumOutput`, the `expectedPayout` key, and `destinationCaptionKey = isLimit ? minPayout : expectedPayout` — down to the translations, so both platforms describe the same swap the same way. Two deliberate deviations:

- iOS's memo **compression** is not ported. Android signs the node's memo verbatim, which is why the LIM read off it is exact and why scientific notation is refused here. (That does leave a latent gap on Android — no 80-byte OP_RETURN guard on the market-swap path — worth its own issue.)
- The consent checkbox "I agree with the min. amount I'll receive" is left alone; iOS kept its equivalent (`swapVerifyCheckbox2Description`) unchanged.

## Test plan

49 new unit tests, all verified to actually run (`tests=15/12/8/14`, `skipped=0`):

- `ThorchainMemoLimitTest` — the issue's real Auto-slippage memo → no floor; a memo with a LIM → that LIM; zero limit, non-swap action, `=<` order, short memo, unreadable triple, non-integer/non-ASCII, scientific notation, ≥2^256, zero-padding.
- `ThorchainMemoAssetMatchTest` — the memo's own target asset vs. the payload's destination coin: layer-1 natives, synth/trade/secured notations, the abbreviated-contract form the memo builder emits, same ticker at another contract, Cosmos and Maya assets, and the unreadable cases that must not be read as a mismatch.
- `SwapTransactionToUiModelMapperMinPayoutTest` — the #5711 swap with and without a tolerance, zero limit, aggregator route, limit order.
- `SwapMinPayoutTest` — 1e8 rescaling into 8- and 18-decimal destinations, Maya, no memo, provider-built route; and for `signedLimitOrder`, a `=<` memo asserting no separate floor, a matching vs. mismatched target asset, an unresolvable spelling, and the `=` / null memos that place no order.

Green after merging `main`: full `:app:testDebugUnitTest` (2289 tests) + `:data:testDebugUnitTest` (3406 tests), 0 failures, and `:app:lintDebug`. ktfmt applied.

**Manual verification** — device, Gold-tier vault, THORChain **ETH → RUNE**:

- **Auto slippage** — the verify screen reads `expected payout` over `8.97170991 RUNE`, with no floor line. The memo (`=:r:thor1…:0/0/3`) carries `LIM = 0`, so nothing is enforced. This is the #5711 case: on `main` the same screen said "min. payout" over that same number.
- **3% slippage** (toolbar sliders → Slippage → 3%) — caption stays `expected payout` over `8.961249 RUNE`, with `min. payout: 8.72730424 RUNE` beneath the fiat.

That floor is **2.61%** below the expected amount rather than the 3% requested, which is exactly why this reads the memo instead of re-deriving from the tolerance. Live thornode, same pair and amount:

```
no affiliate       → LIM 872730873   expected_amount_out 898821632
affiliate va/30bps → LIM 872730873   expected_amount_out 896125362   ← what the app sends
affiliate va/10bps → LIM 872730873   expected_amount_out 897923005
```

The node pegs `LIM` to the full-size, pre-affiliate emit — identical across all three — while `expected_amount_out` drops with the affiliate cut. `expected × (1 − 300/10_000)` would have displayed `8.6924`, a floor the signed memo does not enforce.

- **Aggregator route** (ETH → USDC) — `expected payout` with no floor line, which is the intended behaviour: those sign opaque calldata and expose no floor.

The limit-order caption and the history-card label are covered by unit tests and were not re-driven on device.

## Summary by CodeRabbit

- **New Features**
  - Swap verification shows signed minimum payouts when available.
  - Limit orders display minimum payout details, while market swaps show expected payout.
  - Transaction history identifies limit orders and refreshes in-flight transactions automatically.

- **Bug Fixes**
  - Improved validation for mismatched or invalid swap assets and memo data.

- **Localization**
  - Added payout-related translations across supported languages.

- **Tests**
  - Expanded coverage for payout calculations, memo parsing, limit orders, and invalid routes.


app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapMinPayout.kt
```diff
@@ -0,0 +1,43 @@
...
+ */
+internal fun signedMinimumOutput(payload: SwapPayload, memo: String?, dstToken: Coin): TokenValue? {
+    if (payload !is SwapPayload.ThorChain && payload !is SwapPayload.MayaChain) return null
+    if (memo == null) return null
+    val limit = ThorchainMemoLimit.assertedLimit(memo) ?: return null
+    val targetAsset = ThorchainMemoLimit.targetAsset(memo)
+    if (
+        targetAsset != null && dstToken.compareToMemoAsset(targetAsset) == MemoAssetMatch.DIFFERENT
+    ) {
+        return null
```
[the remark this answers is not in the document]
aminsato: Half of this was real and is fixed in 7b68dd85f; the other half was already handled by design.

**Real:** `isLimitOrder` was derived from a bare `LimitSwapMemo.parse(memo)` with no comparison against the destination coin, in both `SwapTransactionToUiModelMapper` and `JoinSwapUiModelBuilder`. A limit order's displayed destination amount *is* the floor its memo enforces, so `min. payout` over it is a guarantee — and on a cosigner the memo and the payload's destination coin are two independently decoded halves of one request. A memo placing an order for CACAO could therefore put `min. payout` over a TCY amount. Both call sites now go through one helper:

```kotlin
internal fun signedLimitOrder(memo: String?, dstToken: Coin): LimitSwapMemo.Parsed? {
    val parsed = memo?.let(LimitSwapMemo::parse) ?: return null
    if (dstToken.compareToMemoAsset(parsed.targetAsset) == MemoAssetMatch.DIFFERENT) return null
    return parsed
}
```

The comparison stays one-sided, as it already is for market swaps: `UNREADABLE` keeps the order classified rather than silently demoting a valid `=<` to a market swap.

**Not real:** `SwapMinPayout.kt#L33-38` never runs for a `=<` memo. `signedMinimumOutput` short-circuits three lines earlier on `ThorchainMemoLimit.assertedLimit(memo) ?: return null`, and `assertedLimit` rejects any non-swap action — `=<` included — so no LIM from a limit memo is ever converted into `dstToken` and the asset check is unreachable on that path. No change needed there; a test now pins it so it can't silently start mattering.

**Tests** (all verified to run, `skipped=0`): six new cases in `SwapMinPayoutTest` — a `=<` memo asserts no separate floor, a matching target parses as the order, a mismatched target does not, an unresolvable spelling stays an order, and a `=` memo / a null memo place none — plus `a limit memo naming another asset is not this destination's order` in `SwapTransactionToUiModelMapperMinPayoutTest` asserting `isLimitOrder == false` end to end.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapMinPayout.kt
```diff
@@ -0,0 +1,63 @@
...
+    }
+    // The LIM is in the protocol's own fixed point — THORChain's 1e8 whatever the destination
+    // chain, Maya's 1e10 for CACAO — which is exactly what convertToTokenValue rescales from.
+    return dstToken.convertToTokenValue(limit.toString())
```
rkokhatskyi, line 43: `signedMinimumOutput` shows the memo LIM as the floor on the user's receipt, but thornode checks the LIM against the pre-affiliate emit and skims `fees.affiliate` from the outbound after, so the real floor is `LIM − affiliate`. Live at the 0.5% preset LIM `31739612` even lands above `expected_amount_out 31657653`

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapMinPayout.kt
```diff
@@ -0,0 +1,63 @@
...
+ * floor its memo enforces, so labelling it "min. payout" under a ticker the memo does not name
+ * would promise a minimum in the wrong asset. A cosigner decodes the memo and the payload's
+ * destination coin as two independent halves of the same request, which is where they can disagree.
+ *
+ * [LimitSwapMemo.parse] enforces the memo's own grammar; the asset comparison is one-sided, so a
+ * spelling this app cannot resolve leaves the order classified rather than silently demoting it to
+ * a market swap.
+ */
+internal fun signedLimitOrder(memo: String?, dstToken: Coin): LimitSwapMemo.Parsed? {
+    val parsed = memo?.let(LimitSwapMemo::parse) ?: return null
```
rkokhatskyi, line 60: `signedLimitOrder` classifies through `LimitSwapMemo.parse`, whose `^\d+/\d+/\d+$` trade target rejects the `<mantissa>e<exponent>` LIM iOS always emits from `compressLim` (`LimitSwapMemoBuilder.swift:144`), so an Android co-signer now captions an iOS order's enforced floor "expected payout" where main said "min. payout"

data/src/main/kotlin/com/vultisig/wallet/data/swap/ThorchainMemoLimit.kt
```diff
@@ -0,0 +1,90 @@
+package com.vultisig.wallet.data.swap
+
+import java.math.BigInteger
+
+/**
+ * Reads the minimum-output floor (`LIM`) a THORChain / MayaChain swap memo asserts.
+ *
+ * The node bakes `LIM = expected_amount_out × (10_000 − tolerance_bps) / 10_000` into the memo it
```
rkokhatskyi: This comment should state what the node actually computes because `LIM = expected_amount_out × (10_000 − tolerance_bps) / 10_000` is not it: a live `tolerance_bps=50` quote returned LIM `31739612` against `expected_amount_out 31657653`, which that formula cannot produce

data/src/main/kotlin/com/vultisig/wallet/data/swap/ThorchainMemoLimit.kt
```diff
@@ -0,0 +1,90 @@
...
+    fun targetAsset(memo: String): String? {
+        val fields = memo.split(":")
+        if (fields.size < 2 || !isSwapAction(fields[0])) return null
+        return fields[1].takeIf { it.isNotBlank() }
+    }
+
+    /**
+     * THORChain and MayaChain both spell the swap action `SWAP`, `=` or `s`, case-insensitively.
+     * Every other action (`ADD`, `WITHDRAW`, `LOAN+`, the `=<` limit order …) lays its fields out
+     * differently, so its 4th field is not a `LIM/INTERVAL/QUANTITY` triple and must not be read as
```
rkokhatskyi: This comment should give the real reason `=<` is excluded because a limit memo's 4th field IS a `LIM/INTERVAL/QUANTITY` triple at the same index (`LimitSwapMemo.build` emits `=<:TARGET:DEST:LIM/INTERVAL/0`), so only the action check keeps it out

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapMinPayout.kt
aminsato: Confirmed, and it's worse than the affiliate alone: the outbound fee comes off after that check too, so what reaches the wallet is bounded below by `LIM − outbound_fee − affiliate`, not by LIM. Your 0.5% quote is the visible form of it — the screen would render `min. payout: 31739612` directly beneath an expected payout of `31657653`.

Your numbers match mine from the other direction: same pair and amount at 3% tolerance, LIM `872730873` unchanged across 0 / 10 / 30 bps affiliate while `expected_amount_out` moves 898821632 → 897923005 → 896125362. The LIM's basis is the emit before both deductions.

I've left the displayed number as it is in this PR, and I don't think it should be patched here:

- the honest floor needs the outbound fee, which exists only in the quote. A cosigner has the memo bytes and nothing else, so it cannot compute the same number — and the initiator and the cosigner rendering the *same* number is the reason this line exists at all. Subtracting initiator-side only would break exactly the property it was built for.
- iOS displays the same raw LIM (`SwapCryptoLogic.signedMinimumOutput`), so changing only Android would put the two platforms on different numbers for the same swap.

What I think actually fixes it is the label rather than the arithmetic: the row is truthful as the floor the chain checks the swap against, and false only as a promise about the receipt. That's one string across 10 locales plus the iOS half, which is why I'd rather it be its own change on both platforms than a rider on this one. Happy to do it here instead if you'd rather it not wait.

e0eb538b0 at least documents the pre-affiliate, pre-outbound basis in `ThorchainMemoLimit` so the next reader doesn't have to re-derive it from quotes.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapMinPayout.kt
aminsato: Right — fixed in e0eb538b0. `LimitSwapMemo.decodeLim` now reads both spellings the chain accepts, the plain decimal and `<mantissa>e<exponent>`, and `parse` / `assertLimitSwapMemo` both go through it.

The caption wasn't the only casualty: `PendingLimitOrderRepository.record` bails on `parse == null`, so an Android cosigner was also dropping the iOS-placed order out of its pending list entirely.

Bounds mirror iOS's own decoder (`decodeCompressedLim`): field ≤ 40 chars, exponent ≤ 80, and the value under THORNode's `cosmos.Uint` ceiling — `ThorchainMemoLimit.LIMIT_UPPER_BOUND`, now `internal` so both memo grammars refuse the same magnitudes the chain does. The pattern is ASCII-only on purpose: `BigInteger` and `Character.digit` both accept non-ASCII numerals, so `١٦` would otherwise decode to 16. `build` still writes the plain form.

Tests: a 15-case `decodeLim` table, both spellings parsing to the same `Parsed`, `assertLimitSwapMemo` accepting `16e8` while still rejecting `0e8` with the zero-LIM message, and a `signedLimitOrder` case proving a `32e7` memo classifies as an order. `:data` 3423 / `:app` 2290 tests, 0 failures; `:app:lintDebug` green.

The market path (`ThorchainMemoLimit.assertedLimit`) still refuses the shorthand deliberately: Android signs the node's memo verbatim so it never writes one there, and for an Android cosigner of an iOS UTXO market swap refusing it means rendering no floor line at all — which fails in the safe direction rather than captioning one.

data/src/main/kotlin/com/vultisig/wallet/data/swap/ThorchainMemoLimit.kt
aminsato: Fixed in e0eb538b0. The comment now states what the node actually applies the tolerance to — the swap's own emit, before the outbound fee and before the affiliate cut — and carries the evidence rather than a formula: same pair and amount at 3% tolerance, LIM `872730873` unchanged at 0 / 10 / 30 bps affiliate while `expected_amount_out` moves 898821632 → 897923005 → 896125362, plus your 50 bps case where LIM `31739612` lands above `expected_amount_out 31657653`.

It now ends where your other comment starts: the LIM is the level the chain checks the swap against, not the amount that reaches the user.

Worth flagging that iOS carries the same wrong formula in the doc comment on `SwapCryptoLogic.signedMinimumOutput`.

data/src/main/kotlin/com/vultisig/wallet/data/swap/ThorchainMemoLimit.kt
aminsato: Fixed in e0eb538b0. The comment now says the action check is the only thing keeping a `=<` order out: its 4th field is a `LIM/INTERVAL/QUANTITY` triple at the same index — exactly what `LimitSwapMemo.build` emits (`=<:TARGET:DEST:LIM/INTERVAL/0`) — so the field layout alone would read one happily.

It also says why it must stay out, which the old text never did: a limit order's displayed destination amount already IS its LIM, so returning it here would spell the same floor out a second time as if it were a market swap's minimum. The "other actions lay their fields out differently" half survives only where it's true, for `ADD` / `WITHDRAW` / `LOAN+`.

johnnyluo (approved):
LGTM

PR #5735, opened 2026-08-27
fix(tx-history): keep checking a pending row instead of asking once

### LINKED ISSUE #5729: History keeps a mined ETH send as "In progress" for 49+ min after the receipt succeeded
## Observed

Android v1.0.118 (debug build from tag `37f13e2df`). A 0.0001 ETH self-send broadcast at 07:32 local — `0x9530f403c3298e943ec06c0ebfe009db84d0363e0264debf91c54cf6b76c8014`, mined in block 25839408 with `status=0x1` — is still listed as **"In progress… 49m 40s ago"** on the History → Overview tab 49 minutes later, while later transactions (RUNE send, THORChain swap, a LI.FI swap error) show final states.

Logcat shows the app polling `eth_getTransactionReceipt` for that hash right after broadcast and `TransactionStatusService` being stopped by the FGS timeout ~8 s later; the row never gets a second chance.

## Cause (suspected)

`TransactionStatusService` is a foreground service that the system stops on timeout before the receipt lands; there is no re-check when History is opened, so a tx whose receipt arrived after the service died stays "In progress" forever (or until whatever periodic refresh exists runs — none did within 49 min on an app that was force-stopped and relaunched several times).

## Expected

Opening History (or the row) re-queries the receipt for any "In progress" row older than a block time and flips it to Successful/Error. F-189 ("pending tx auto-refreshes to confirmed") in the release checklist fails today.

## Env

Emulator Pixel 9 Pro API 36, v1.0.118 debug. Screenshot of the row is in the release-qa run notes (`hist1.png`); on-chain receipt via `ethereum-rpc.publicnode.com`.

aminsato:
## Problem

A mined ETH send stayed listed as "In progress" for 49+ minutes after its receipt succeeded — `0x9530f403…`, block 25839408, `status=0x1`.

The issue's stated cause ("there is no re-check when History is opened") turned out not to be the defect. `TransactionHistoryViewModel.refreshOnEnter()` has polled from `init` since #3233, and the receipt query itself is fine — querying the app's own default endpoint (`CustomRpcDefaultEndpoint.evmUrl(Chain.Ethereum)`) with that hash returns `status:"0x1"` and every field `EvmRpcResponseJson<EvmTxStatusJson>` requires, so `EvmStatusProvider` would have returned `Confirmed`.

The defect is the cadence. That check ran **exactly once, at ViewModel construction**. On a row broadcast seconds earlier it lands before the receipt exists, writes PENDING, and nothing is scheduled behind it. Meanwhile `InProgressPill` re-renders its elapsed label every second, so a frozen row reads as live.

## Changes

- `refreshOnEnter()` moves out of `init` into `onScreenResumed()`, driven by a `LifecycleResumeEffect` — every return to the screen re-queries, not just the first construction.
- `observeInFlightRows()` polls on a 15s timer, gated on *screen visible* **and** *any row still in flight*, combined into a single collector. An idle history screen issues no requests; a backgrounded one stops rather than polling from behind the home screen. Visibility is a `StateFlow` rather than a `Job?` field, so the resume and pause callbacks share no mutable handle.
- New `RefreshPendingTransactionsUseCase.refreshOne(chain, txHash)` behind `openDetail`, so opening a pending row asks the chain about that one transaction. It deliberately skips the sweep's exponential backoff: the backoff paces unattended polling, and someone opening a specific row is asking a direct question about it. Settled and missing rows are no-ops.
- `TransactionStatus.isInFlight`, exhaustive over the enum so a future status has to declare which side it falls on rather than defaulting to terminal and going unpolled forever.

## Not in scope

`TransactionHistoryDao.getAllPendingTransactions()` and its repository wrapper have zero callers — dead code. So nothing settles a pending row while the user sits on Home or a chain screen. The issue's "Expected" names only History and the row; a background poller carries its own battery and multi-vault questions and belongs in a separate change. Happy to file it.

## Tests

11 new unit tests, no device needed.

**`RefreshPendingTransactionsUseCaseRefreshOneTest`** (4) — checks the chain and persists the result; leaves a settled row alone; no-op when the row is gone; and the contrast case that gives the method its reason to exist: the same backed-off row that `invoke()` skips is still checked by `refreshOne`.

**`TransactionHistoryViewModelTest`** (7 new) — `onScreenResumed` refreshes; the timer fires while visible; the timer stops on pause; no timer runs when every row has settled; both `openDetail` branches. The existing `refresh invokes refreshPendingTransactions` moves from `atLeast = 2` to `exactly = 1` now that construction alone no longer polls.

Full suites green — app 2247 tests, data 3188, 0 failures, 0 errors (XML reports confirm the new cases ran rather than being silently skipped). `:app:lintDebug` clean. ktfmt applied.

## Manual check

Ethereum → Send a small amount to your own address → tap Done immediately → open History (Overview) and stay on it. The row flips to Successful on its own within ~15s of the transaction being mined, with no pull-to-refresh. Backgrounding the app with History open and returning also settles it, as does tapping the pending row to open its detail sheet. A failed transaction must still land on Failed and stay there.

Closes #5729

## Summary by CodeRabbit

* **New Features**
  * Transaction history now refreshes active transactions while the screen is open.
  * Returning to the screen triggers a refresh.
  * Opening an active transaction triggers an immediate status update.
  * Pending transactions affected by temporary indexer delays continue to be tracked.
  * Open transaction details stay synchronized with refreshed status updates.

* **Bug Fixes**
  * Settled transactions are no longer repeatedly refreshed.
  * Individual transaction refreshes now bypass periodic refresh delays.
  * Temporary refresh failures no longer stop ongoing status updates.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
@@ -356,7 +356,7 @@ constructor(
      * and a backgrounded one stops rather than polling from behind the home screen.
      */
     private fun observeInFlightRows() {
-        viewModelScope.launch {
+        viewModelScope.safeLaunch(onError = { t -> Timber.w(t, "In-flight polling stopped") }) {
             transactionHistoryRepository
                 .observeTransactions(
                     vaultId = vaultId,
@@ -542,9 +542,34 @@ constructor(
                         }
                 }
                 .collect { groups ->
-                    _uiState.update { it.copy(groups = groups, isLoading = false) }
+                    _uiState.update {
+                        it.copy(
+                            groups = groups,
+                            isLoading = false,
+                            selectedItem = it.selectedItem.reconciledWith(groups),
+                        )
+                    }
                 }
         }
+    }
+
+    /**
+     * Re-points an open detail sheet at the row this emission just carried.
+     *
+     * [TransactionHistoryUiState.selectedItem] is a snapshot taken when the sheet was presented, so
+     * nothing else reaches it: a sheet left open while the poll settles the transaction keeps
+     * reading "In progress" above a row that has already flipped to Confirmed behind it.
+     *
+     * A row that is no longer in the list — filtered out, or on a tab that does not carry it —
+     * keeps the snapshot rather than closing the sheet under the user.
+     */
+    private fun TransactionHistoryItemUiModel?.reconciledWith(
+        groups: List<TransactionHistoryGroupUiModel>
+    ): TransactionHistoryItemUiModel? {
+        val current = this ?: return null
+        return groups.firstNotNullOfOrNull { group ->
+            group.transactions.firstOrNull { it.id == current.id }
+        } ?: current
     }
 
     private fun TransactionHistoryItemUiModel.matchesAssetIds(assetIds: Set<String>): Boolean =
--- a/app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
@@ -9,6 +9,7 @@ import androidx.navigation.toRoute
 import com.vultisig.wallet.R
 import com.vultisig.wallet.data.db.models.TransactionHistoryEntity
 import com.vultisig.wallet.data.db.models.TransactionStatus
+import com.vultisig.wallet.data.db.models.isInFlight
 import com.vultisig.wallet.data.models.ImageModel
 import com.vultisig.wallet.data.models.SendTransactionHistoryData
 import com.vultisig.wallet.data.models.SwapTransactionHistoryData
@@ -50,6 +51,7 @@ import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.flatMapLatest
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -268,8 +295,19 @@ constructor(
...
+     * to someone who just asked about this specific transaction, without having looked, is what the
+     * sweep's backoff would otherwise do. One status call is a fair price for a current answer.
+     */
     fun openDetail(item: TransactionHistoryItemUiModel) {
         _uiState.update { it.copy(selectedItem = item) }
+        if (!item.status.isInFlight()) return
+        viewModelScope.safeLaunch(
+            onError = { t -> Timber.w(t, "Detail-sheet status re-check failed") }
+        ) {
+            refreshPendingTransactions.refreshOne(item.chain, item.txHash)
```
rkokhatskyi, line 309: `openDetail` fires `refreshOne` but `selectedItem` at `TransactionHistoryViewModel.kt:304` is a snapshot nothing rebuilds, so the sheet keeps reading In progress after the row behind it flips to Confirmed, while iOS reconciles the same snapshot in `TransactionHistoryViewModel.swift:142`

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
aminsato: Right, and it is worse with the poller this PR adds than it was before it — `selectedItem` had exactly two writers, `openDetail` and `dismissDetail`, so the sheet held its snapshot across every emission that settled the row underneath it.

Reconciled on every list emission now, matched on `id`, the same shape as `reconcileSelectedDetail()` on iOS. A row that is not in the emission — another tab, or filtered out by an asset chip — keeps its snapshot rather than having the sheet closed out from under a read in progress.

`6fee8a62d`, with two tests: the sheet follows its row Broadcasted → Confirmed, and stays put when the row leaves the list.

app/src/main/java/com/vultisig/wallet/ui/models/TransactionHistoryViewModel.kt
```diff
@@ -311,6 +349,35 @@ constructor(
...
+                .distinctUntilChanged()
+                .collectLatest { shouldPoll ->
+                    if (!shouldPoll) return@collectLatest
+                    while (true) {
+                        delay(IN_FLIGHT_POLL_INTERVAL)
+                        runCatchingRefresh("in-flight transactions") {
+                            refreshPendingTransactions(vaultId, chainId)
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
aminsato: Valid — `runCatchingRefresh` wraps only the refresh call, so a throw from `observeTransactions`, `map` or `combine` escaped uncaught. Switched to `viewModelScope.safeLaunch(onError = ...)` in `6fee8a62d`.

rkokhatskyi (approved):
LGTM

PR #5739, opened 2026-08-28
fix(swap): keep Jupiter quotes when the fee ATA is missing

realpaaao:
## Summary
- Jupiter no longer fails the quote when the Vultisig affiliate fee ATA is unprovisioned.
- Missing ATA or a fee-bearing `/quote` 4xx requotes without `platformFeeBps` / `feeAccount`, so long-tail Solana tokens still route.
- A fresh no-fee quote is fetched rather than stripping `platformFee` off a fee-bearing quote JSON (Jupiter rejects that).

## Changes
| File | Change |
|------|--------|
| `JupiterApi.kt` | `fetchRouteQuote` 4xx retry; requote without fee when ATA resolve throws |
| `JupiterFeeAtaService.kt` | Probe still throws; caller now requotes instead of dropping Jupiter |
| `JupiterApiTest.kt` | Replaced "fails the quote" with requote + 4xx-retry cases |

## Context
iOS reproduced `JupiterError error 4` (`feeAccountNotProvisioned`) on SOL to BANDOS. Android has the same drop: `resolveFeeAccount` throw failed the Jupiter candidate so a meme with no LiFi index had no route. Jupiter's own `/quote` returned 200 for that pair.

Sibling iOS PR: https://github.com/vultisig/vultisig-ios/pull/5289
Windows/SDK already prepend the fee ATA create and are unchanged.

## Test plan
- [x] `JupiterApiTest` — unprovisioned ATA requotes without fee, no `feeAccount` on `/swap`
- [x] `JupiterApiTest` — fee-bearing 4xx requotes without fee
- [x] Existing fee-present / fee-zero / native-SOL-output cases still pass
- [x] `ktfmtCheckMain` + `ktfmtCheckTest`
- [ ] Live SOL to long-tail SPL on a device (no on-chain receipt in this PR)

## receipts
```
$ JAVA_HOME=/opt/homebrew/opt/openjdk@21 ./gradlew :data:testDebugUnitTest --tests com.vultisig.wallet.data.api.JupiterApiTest
BUILD SUCCESSFUL in 16s
29 actionable tasks: 12 executed, 17 up-to-date
```

Live Jupiter probe: SOL to BANDOS HG3sZ52NRmNpm2BueL1MSBM3krr5M6YJtQohjNyhpump `/quote` HTTP 200; fee ATA not provisioned.

No signed broadcast / explorer tx for this change. Do not request review until a device swap is captured, or a maintainer accepts tests as enough.

## Summary by CodeRabbit

* **New Features**
  * Swap quotes now include platform fee amount and currency details when applicable.
  * Swap fee information is calculated using the destination token for supported swaps.

* **Bug Fixes**
  * Improved quote reliability when affiliate fee accounts are unavailable.
  * Fee-enabled quote errors now surface correctly without issuing an unnecessary swap request.
  * Improved handling of native SOL and wrapped SOL fee calculations.

* **Tests**
  * Expanded coverage for fee fallback, error handling, and fee reporting scenarios.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -312,7 +312,6 @@ constructor(
                         srcTokenValue,
                         tokenValue,
                         vultBPSDiscount,
-                        srcNativeToken,
                         slippageBps,
                     )
 
@@ -1078,7 +1077,6 @@ constructor(
         srcTokenValue: BigInteger,
         tokenValue: TokenValue,
         vultBPSDiscount: Int?,
-        srcNativeToken: Coin,
         slippageBps: Int?,
     ): Pair<SwapQuote, UiText> {
         // LI.FI and Jupiter both take a quote-time slippage override, so slippage is part of the
@@ -1142,11 +1140,10 @@ constructor(
                             )
                         Pair(feeWei, dstToken)
                     } else {
-                        resolveSwapFee(
-                            apiQuote.tx.swapFeeTokenContract,
-                            apiQuote.tx.swapFee,
-                            srcNativeToken,
-                            apiQuote.tx.swapFee.toBigInteger(),
+                        // Jupiter ExactIn platform fee is output-mint base units.
+                        Pair(
+                            apiQuote.tx.swapFee.toBigIntegerOrNull() ?: BigInteger.ZERO,
+                            dstToken,
                         )
                     }
                 val updatedTx = apiQuote.tx.withResolvedSwapFee(feeAmount, feeCoin)
--- a/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
@@ -334,10 +334,8 @@ constructor(
             when (provider) {
                 SwapProvider.MAYA,
                 SwapProvider.THORCHAIN,
-                SwapProvider.LIFI -> dstToken
-                // SwapKit inbound fee is source-native (like 1inch/Kyber/Jupiter, not LiFi's
-                // destination-side integrator model).
-                SwapProvider.SWAPKIT -> srcNativeToken
+                SwapProvider.LIFI,
+                SwapProvider.JUPITER -> dstToken
                 else -> srcNativeToken
             }
 
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
@@ -51,30 +51,18 @@ constructor(
         val feeMint = if (toToken == SOLANA_DEFAULT_CONTRACT_ADDRESS) fromToken else toToken
         val slippage = slippageBps ?: DEFAULT_SLIPPAGE_BPS
 
-        var body =
+        // Probe the fee ATA before quoting. A missing account must never take a fee-bearing
+        // /quote — outAmount would be net of a fee we cannot collect, so the UI would lie.
+        val resolvedFeeAccount = requestedFeeBps?.let { resolveFeeAccountOrNull(feeMint) }
+        val body =
             fetchRouteQuote(
                 fromToken = fromToken,
                 toToken = toToken,
                 fromAmount = fromAmount,
                 slippageBps = slippage,
-                platformFeeBps = requestedFeeBps,
+                platformFeeBps = requestedFeeBps.takeIf { resolvedFeeAccount != null },
             )
-        var feeAccount = affiliateFeeAccount(feeMint, requestedFeeBps, body)
-        if (
-            requestedFeeBps != null && feeAccount == null && quotedFeeAmount(body) > BigInteger.ZERO
-        ) {
-            // Fresh no-fee quote: stripping platformFee off a fee-bearing quote JSON makes Jupiter
-            // reject /swap.
-            body =
-                fetchRouteQuote(
-                    fromToken = fromToken,
-                    toToken = toToken,
-                    fromAmount = fromAmount,
-                    slippageBps = slippage,
-                    platformFeeBps = null,
-                )
-            feeAccount = null
-        }
+        val feeAccount = resolvedFeeAccount.takeIf { quotedFeeAmount(body) > BigInteger.ZERO }
 
         val outAmount = body.outAmount
         val routePlan = body.routePlan
@@ -127,10 +115,13 @@ constructor(
                 swapTxData
             }
 
+        val platformFeeAmount = body.platformFee?.amount.takeIf { feeAccount != null }
         return QuoteSwapTotalDataJson(
             swapTransaction = quoteSwapData.copy(data = updatedSwapTx),
             dstAmount = outAmount,
             routePlan = routePlan,
+            platformFeeAmount = platformFeeAmount,
+            platformFeeMint = feeMint.takeIf { platformFeeAmount != null },
         )
     }
 
@@ -167,20 +158,14 @@ constructor(
         return response.bodyOrThrow()
     }
 
-    private suspend fun affiliateFeeAccount(
-        feeMint: String,
-        requestedFeeBps: Int?,
-        body: SwapRouteResponseJson,
-    ): String? {
-        if (requestedFeeBps == null || quotedFeeAmount(body) <= BigInteger.ZERO) return null
-        return try {
+    private suspend fun resolveFeeAccountOrNull(feeMint: String): String? =
+        try {
             feeAtaService.resolveFeeAccount(feeMint)
         } catch (cancelled: CancellationException) {
             throw cancelled
         } catch (_: Exception) {
             null
         }
-    }
 
     private fun quotedFeeAmount(body: SwapRouteResponseJson): BigInteger =
         body.platformFee?.amount?.toBigIntegerOrNull() ?: BigInteger.ZERO
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
@@ -14,6 +14,7 @@ import io.ktor.client.request.setBody
 import io.ktor.http.HttpStatusCode
 import java.math.BigInteger
 import javax.inject.Inject
+import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.buildJsonObject
 import kotlinx.serialization.json.encodeToJsonElement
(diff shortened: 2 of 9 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -143,6 +134,57 @@ constructor(
...
+        body: SwapRouteResponseJson,
+    ): String? {
+        if (requestedFeeBps == null || quotedFeeAmount(body) <= BigInteger.ZERO) return null
+        return try {
+            feeAtaService.resolveFeeAccount(feeMint)
+        } catch (cancelled: CancellationException) {
+            throw cancelled
+        } catch (_: Exception) {
+            null
+        }
```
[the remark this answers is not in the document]
realpaaao: Declining a dedicated exception type for the same reason as the iOS sibling: unexpected probe failures (RPC, missing mint, unprovisioned ATA) must requote without the fee. Catching only a narrow type would fail the Jupiter candidate again and leave long-tail pairs with no route. `CancellationException` is already rethrown.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -46,47 +47,37 @@ constructor(
...
+                toToken = toToken,
+                fromAmount = fromAmount,
+                slippageBps = slippage,
+                platformFeeBps = requestedFeeBps,
+            )
+        var feeAccount = affiliateFeeAccount(feeMint, requestedFeeBps, body)
+        if (
+            requestedFeeBps != null && feeAccount == null && quotedFeeAmount(body) > BigInteger.ZERO
+        ) {
+            // Fresh no-fee quote: stripping platformFee off a fee-bearing quote JSON makes Jupiter
```
rkokhatskyi: This comment states the inverse of Jupiter's rule: `/swap` accepts a `platformFee`-stripped quote and rejects only one that keeps it with no `feeAccount`, which is exactly what line 81 already relies on. It should say the requote exists to get the gross `outAmount`

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -46,47 +47,37 @@ constructor(
...
+
+        var body =
+            fetchRouteQuote(
+                fromToken = fromToken,
+                toToken = toToken,
+                fromAmount = fromAmount,
+                slippageBps = slippage,
+                platformFeeBps = requestedFeeBps,
+            )
+        var feeAccount = affiliateFeeAccount(feeMint, requestedFeeBps, body)
```
rkokhatskyi: `affiliateFeeAccount` runs after the quote, so the missing-ATA path spends a second `/quote` that iOS avoids by resolving `feeMint` before quoting in `JupiterService.swift`, and this requote duplicates the no-fee fallback already inside `fetchRouteQuote`

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -143,6 +134,57 @@ constructor(
...
+            feeAtaService.resolveFeeAccount(feeMint)
+        } catch (cancelled: CancellationException) {
+            throw cancelled
+        } catch (_: Exception) {
+            null
+        }
+    }
+
+    private fun quotedFeeAmount(body: SwapRouteResponseJson): BigInteger =
+        body.platformFee?.amount?.toBigIntegerOrNull() ?: BigInteger.ZERO
```
rkokhatskyi, line 168: The real affiliate fee read here never leaves this class, so the fee row falls back to `routePlan[].swapInfo.feeAmount` at `JupiterQuoteSource.kt:31`, a field Jupiter's current schema no longer defines, leaving every Jupiter swap rendering `Estimated Fees (0.50%)` as `$0.00`

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
realpaaao: Fixed — that comment is gone with the requote. ATA is resolved first; `/swap` still strips `platformFee` when there is no `feeAccount` so `outAmount` is the gross figure.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
realpaaao: Fixed — `resolveFeeAccountOrNull` runs before `fetchRouteQuote`. A missing ATA quotes once with no `platformFeeBps`. The 4xx no-fee retry inside `fetchRouteQuote` stays for a fee-bearing quote Jupiter itself rejects.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
realpaaao: Fixed — `QuoteSwapTotalDataJson` now carries `platformFeeAmount` / `platformFeeMint`. `JupiterQuoteSource` uses those for `tx.swapFee` instead of `routePlan[].swapInfo.feeAmount`.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -136,13 +115,61 @@ constructor(
                 swapTxData
             }
 
+        val platformFeeAmount = body.platformFee?.amount.takeIf { feeAccount != null }
         return QuoteSwapTotalDataJson(
             swapTransaction = quoteSwapData.copy(data = updatedSwapTx),
             dstAmount = outAmount,
             routePlan = routePlan,
+            platformFeeAmount = platformFeeAmount,
+            platformFeeMint = feeMint.takeIf { platformFeeAmount != null },
```
rkokhatskyi, line 126: `platformFeeMint` is the input mint on a native-SOL destination while Jupiter denominates `platformFee.amount` in the output mint, so USDC to SOL prices a lamport amount as USDC and the row reads about 9.6x the real fee. The fee owner's wSOL ATA does exist, so the line 51 special case can go

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/JupiterQuoteSource.kt
```diff
@@ -41,8 +45,9 @@ internal class JupiterQuoteSource @Inject constructor(private val jupiterApi: Ju
                         gas = 0,
                         value = "0",
                         gasPrice = "0",
-                        swapFee = feeRoute?.swapInfo?.feeAmount ?: "0",
-                        swapFeeTokenContract = feeRoute?.swapInfo?.feeMint.orEmpty(),
+                        swapFee = platformFeeAmount ?: "0",
```
rkokhatskyi, line 48: `estimatedFees` is always `srcNativeToken` out of `resolveSwapFee`, but `SwapTransactionToUiModelMapper.kt:67` prices Jupiter with `from.srcToken`, so this now-nonzero fee renders wrong on Verify and Done for every SPL-source swap and disagrees with the form

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/JupiterQuoteSource.kt
```diff
@@ -41,8 +45,9 @@ internal class JupiterQuoteSource @Inject constructor(private val jupiterApi: Ju
...
+                        swapFeeTokenContract =
+                            if (platformFeeAmount != null) quote.platformFeeMint.orEmpty() else "",
```
rkokhatskyi, line 50: `swapFeeTokenContract` is now a base58 mint, so `resolveSwapFee` takes its `else` branch and on a `searchToken` miss the catch returns `Pair(fallbackFee, srcNativeToken)` at `SwapQuoteManager.kt:1368`, reinterpreting the fee mint's base units as SOL lamports

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
aminsato, line 126: `platformFeeMint` is set to `feeMint`, but `feeMint` is the *input* mint whenever the output is native SOL (`toToken == SOLANA_DEFAULT_CONTRACT_ADDRESS`). Jupiter's quote `platformFee.amount` for an ExactIn quote is denominated in the **output** mint's base units — the quote is issued before any `feeAccount` is known, so it can only be output-denominated. Pairing an output-mint amount with the input mint's contract mis-denominates the fee downstream.

Concrete: swap 1000 USDC -> SOL at ~$200/SOL. `platformFee.amount` ~= 24,875,000 (lamports, ~0.0249 SOL, ~$5). We report `swapFee=24875000`, `swapFeeTokenContract=<USDC mint>`. `SwapQuoteManager.resolveSwapFee` then looks USDC up (6 decimals, ~$1) and renders ~24.9 USDC ~= $25 in the Swap Fee row — a ~5x overstatement — and `withResolvedSwapFee` stamps that onto the keysign payload, so the co-signer sees it too.

iOS handles exactly this case by reporting zero: `let feeOnInput = feeMint != outputMint; let platformFee = feeOnInput ? .zero : platformFeeDecimal(...)` (`JupiterService.swift`), with `platformFeeDecimal` documented as "output-mint raw base units" and converted with `toCoin`. Suggest matching that: only surface `platformFeeAmount` when `feeMint == toToken`, and label it with `toToken`.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -136,13 +115,61 @@ constructor(
...
+                parameter("slippageBps", slippageBps)
+                if (platformFeeBps != null && platformFeeBps > 0) {
+                    parameter("platformFeeBps", platformFeeBps)
+                }
+            }
+        if (response.status == HttpStatusCode.TooManyRequests) {
+            throw SwapException.RateLimitExceeded("[Jupiter] Too many requests")
+        }
+        val chargedFee = platformFeeBps != null && platformFeeBps > 0
+        if (chargedFee && response.status.value in 400..499) {
```
aminsato: The requote fires on *any* 4xx, not just fee-related rejections, and it discards the original error body without reading it. Jupiter answers an unroutable pair with `400 {"error":"Could not find any route"}`; with `affiliateBps > 0` (the normal case) that now costs two `/quote` round-trips before `bodyOrThrow` finally throws on the second one — doubling latency and request volume for every unroutable Solana pair, on a hot path that races providers under `QUOTE_FETCH_TIMEOUT_MS`. Same for 401/403 from the `api.vultisig.com/jup` proxy.

Since `resolveFeeAccountOrNull` already proves the fee ATA exists before this call, the fee-specific 4xx this retry targets should be rare. Narrowing it (e.g. only retry when the error body mentions the platform fee / fee account) or dropping it in favour of the probe would avoid the double request on the common failure.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
```diff
@@ -136,13 +115,61 @@ constructor(
...
+        }
+        return response.bodyOrThrow()
+    }
+
+    private suspend fun resolveFeeAccountOrNull(feeMint: String): String? =
+        try {
+            feeAtaService.resolveFeeAccount(feeMint)
+        } catch (cancelled: CancellationException) {
+            throw cancelled
+        } catch (_: Exception) {
```
aminsato: This swallows every non-cancellation exception with no Timber log. `JupiterFeeAtaServiceImpl` throws for three very different reasons — an unsupported/unknown mint, a **transient Solana RPC failure** in `getAccountOwner`, and a genuinely unprovisioned ATA — and all three now silently degrade to a fee-less quote. A Solana RPC outage or a proxy hiccup therefore turns off the affiliate fee for every Jupiter swap for as long as it lasts, with zero signal anywhere: the quote succeeds, the UI shows a $0 swap fee, and nothing is logged.

At minimum log the throwable (`Timber.w(t, "...")`) so the drop is diagnosable; the repo convention is to log swallowed errors rather than discard them.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
realpaaao: Fixed — fee mint is always the output mint (including wSOL). `platformFee.amount` and the ATA share that denomination. `SwapQuoteManager` stamps the amount onto `dstToken`; Verify/Done uses `dstToken` too.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/JupiterQuoteSource.kt
realpaaao: Fixed — Jupiter `estimatedFees` is now `dstToken` (same as LiFi). `SwapTransactionToUiModelMapper` prices Jupiter with `from.dstToken`, so form / Verify / Done agree.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/swap/JupiterQuoteSource.kt
realpaaao: Fixed — Jupiter no longer goes through `resolveSwapFee` / `searchToken`. The output-mint amount is applied as `dstToken` base units, so a miss cannot reinterpret them as SOL lamports.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
realpaaao: Fixed — dropped the input-mint special case. ExactIn fee is collected and displayed in the output mint, including wSOL, so a USDC→SOL quote is no longer priced as USDC.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
realpaaao: Fixed — the blanket 4xx no-fee requote is gone. A 400 (including "no route") fails on the first `/quote`. The ATA probe still decides whether to send `platformFeeBps`.

data/src/main/kotlin/com/vultisig/wallet/data/api/JupiterApi.kt
realpaaao: Fixed — `resolveFeeAccountOrNull` logs the throwable with `Timber.w` before falling back to a no-fee quote. `CancellationException` still rethrows.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -1142,11 +1140,10 @@ constructor(
...
                     } else {
-                        resolveSwapFee(
-                            apiQuote.tx.swapFeeTokenContract,
-                            apiQuote.tx.swapFee,
-                            srcNativeToken,
-                            apiQuote.tx.swapFee.toBigInteger(),
+                        // Jupiter ExactIn platform fee is output-mint base units.
+                        Pair(
+                            apiQuote.tx.swapFee.toBigIntegerOrNull() ?: BigInteger.ZERO,
+                            dstToken,
```
rkokhatskyi, line 1144: `dstToken` here but the form's fee-coin `when` at line 341 still routes Jupiter to `else -> srcNativeToken`, so `convertTokenValueToFiat` scales by the destination decimals and prices with SOL, inflating a SOL to USDC fee roughly 100x and contradicting the Verify screen

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
realpaaao: Fixed — the swap form now prices Jupiter fees with `dstToken`, matching quote construction and the done screen. Regression: `fetchQuote prices Jupiter fees with the destination token`.

data/src/test/kotlin/com/vultisig/wallet/data/api/JupiterApiTest.kt
```diff
@@ -123,23 +122,67 @@ class JupiterApiTest {
...
+                        } else {
+                            captured.swapBody = (request.body as? TextContent)?.text
+                            respond(content = "", status = HttpStatusCode.TooManyRequests)
+                        }
+                    },
+            )
 
-        assertThrows(IllegalStateException::class.java) {
+        assertThrows(NetworkException::class.java) {
             runBlocking {
```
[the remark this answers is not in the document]
realpaaao: Fixed — the reviewed coroutine case now uses `runTest` with the existing data-module coroutine-test dependency.

rkokhatskyi (approved):
LGTM

Issue #5700, opened 2026-08-24
Review prompt: expand eligibility beyond outbound transactions (2-event threshold) + per-version gate + Settings rate link


realpaaao:
Mirrors the iOS issue (vultisig-ios). #5427 shipped, but `InAppReviewRepository.kt` gates on `MIN_SUCCESSFUL_TRANSACTIONS = 3` + `MIN_INSTALL_AGE = 7.days` + `PROMPT_COOLDOWN = 120.days`, and `InAppReviewEffect` only mounts on the Keysign screens (`KeysignScreen.kt`, `JoinKeysignView.kt`). A hold-heavy vault's users mostly receive rather than send, so the eligible pool is near zero by design.

**Changes**

1. Broaden qualifying positive events and lower the threshold to **2**: successful outbound tx (existing), first sighting of a confirmed **incoming** tx (skip dust/spam), vault backup completed, successful vault restore on a new device, device pairing completed.
2. Replace `PROMPT_COOLDOWN = 120.days` with a per-`versionCode` gate ("not yet prompted for this app version") + a 14-day floor. Play's quota is opaque and the flow silently no-ops, so the request stays fire-and-forget; upstream throttling per version is the correct shape.
3. Keep the 7-day install age (`firstInstallMillis()`).
4. Add a Settings row **"Rate the App"** → deep link to the Play listing (`https://play.google.com/store/apps/details?id=com.vultisig.wallet`). Per Play policy the in-app review API must NOT be triggered by a button, so the row links out instead. Design: https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=81598-153356
5. Instrument counters (policy-evaluated / request-sent) via Timber + local storage.

**Guardrails:** no pre-prompt questions ("enjoying the app?" is verbatim-banned by Play in-app-review policy), no incentives, surface the Play card unmodified.

PR #5743, opened 2026-08-29
feat(review): broaden review-prompt eligibility beyond outbound txs

aminsato:
## Problem

The review prompt shipped in #5427 asked only after **three confirmed outbound transactions**, behind a 7-day install age and a 120-day cooldown, and was mounted on the two keysign screens only. A hold-heavy vault's users mostly receive rather than send, so the eligible pool was near zero by design.

## Fix

**Five qualifying moments instead of one, threshold lowered to two.** A confirmed outbound transaction (existing), funds arriving, a completed vault backup, a restore onto a device that had no vaults, and a completed device pairing. Each event carries a stable namespaced identity (`outbound:<hash>`, `backup:<vaultId>`, …) persisted in a 100-entry ring, so every producer stays safe to call again after a re-entry, a recomposition or a process restart.

**Per-version gate replaces the 120-day cooldown.** At most one ask per released marketing version, plus a 14-day floor so a new release cannot ask on the heels of the previous one. Play's quota is opaque and the flow no-ops silently, so the request stays fire-and-forget and all throttling happens upstream. The 7-day install age is unchanged, and it is still read from the package manager rather than a stored timestamp, so users who installed before the gate existed are treated as long-standing.

**Presentation moved to the app root** (`InAppReviewHost` in `MainActivity`), replacing `InAppReviewEffect` on `KeysignScreen`/`JoinKeysignView`. Recording an event and presenting the card are now separate steps — a pairing counts towards eligibility while navigating straight into keygen, which is no place for a store card. The host waits 2 s for the screen to settle and refuses to claim unless the lifecycle is RESUMED: a claim is spent whether or not Play shows anything, so it must never be spent in the background.

**An ask is only spent off a newly counted event.** `recordAndOfferPrompt` composes record-then-offer in one place rather than leaving the ordering to each call site, so a re-entered success screen cannot burn a version's single opportunity with no new milestone behind it. This is the bug the iOS port caught in review.

**Incoming funds are inferred from the balance.** Android has no confirmed-incoming-transaction feed, so an arrival is a spendable-balance increase between two sightings of the same coin on the token detail screen — the same shape iOS uses on its coin detail screen. The first sighting of a coin never counts (a balance the device is only now learning about is not something that just happened). Dust and spam are turned away by a fiat floor for priced assets, and a native-token-only rule for unpriced ones.

**Settings gains a "Rate the App" row** that links out to the Play listing. Play's policy forbids a button from triggering the in-app review flow, and that flow no-ops once its quota is spent, which would leave the row looking broken.

**Instrumentation:** persisted policy-evaluation and prompt-claim counters, logged through Timber on every evaluation and every claim.

## Upgrade path

`in_app_review/successful_transactions` from #5427 is still read whenever the broadened counter has never been written, so upgraders keep the transactions they already earned. The existing `last_prompted_at` key is reused as the anchor for the 14-day floor; the version gate opens for them on the first evaluation because no version has been recorded yet.

## Producers

| Event | Where |
|---|---|
| Confirmed outbound transaction | `KeysignViewModel` — first successful terminal status |
| Funds arrived | `TokenDetailViewModel` — balance delta vs. the previous sighting |
| Vault backup completed | `BackupVaultViewModel`, `BackupPasswordViewModel`, `BackupPasswordRequestViewModel` |
| Vault restored | `ImportFileViewModel`, only when the device holds no other vault |
| Device pairing completed | `KeygenPeerDiscoveryViewModel.next`, `KeysignFlowViewModel.startKeysign` — record only |

A bulk "back up all vaults" counts once, not once per vault: one tap is one milestone. Fast-sign is excluded from pairing — the co-signer is the server, not a device the user had to pair with.

## UI

One new row in Settings → Support, between "Check for Updates" and "Share the App": a sparkle leading icon, the label "Rate the App", and the standard trailing chevron. It reuses the existing `SettingItem` visuals, so spacing, typography and the row divider are unchanged. New drawable `rate_the_app.xml` from Figma node `81598:153356`.

**One deliberate deviation worth a second opinion:** Figma renders that sparkle white (`#F0F4FC`); every other Settings row icon is tinted `primary.accent4` blue by `SettingItem`. This ships with the standard accent4 tint so the row does not stand out from its siblings — happy to switch it if the white is intended.

`settings_screen_rate_the_app` is translated in all ten locales, matching each one's existing wording for the app (iOS's strings reused where they exist; ru and nl written to match the terminology already in those files).

## Verification

- `:app` + `:data` unit tests — 2241 tests, 0 failures.
- `ktfmtCheck` clean, `:app:lintDebug` + `:data:lintDebug` clean.
- The two existing in-app-review test files were updated to the new API; no new test files were added. Seven other `*ViewModelTest` files gained the new constructor parameter.

## Manual validation / risk

- **Play may decline to show a claimed prompt.** Its API reports no display result, so the evaluation and claim counters are recorded separately and the ask is spent on request. This is accepted by the issue.
- Worth exercising by hand: two qualifying events on an install older than seven days; a new version code after fourteen days; the Settings deep link on a Play-enabled build.
- The incoming-funds path needs a real arrival on a token detail screen to exercise end-to-end — the balance seam is unit-tested at its boundaries but the observation itself is only reachable with a live refresh.

## Which feature is affected?
- [x] Sending
- [x] Swap

## Checklist

## Agent Metadata
- **Agent**: Claude Code
- **Issue**: #5700
- **Ports**: vultisig/vultisig-ios#4984, vultisig/vultisig-ios#5282
- **Confidence**: high on the policy, the storage shapes and the claim ordering, all directly unit-tested
- **Needs human review for**: the Play card itself was never observed — Play throttles and often suppresses it with no API feedback, so its presentation is not meaningfully verifiable and no claim is made about it. Also worth a second opinion: the white-vs-blue sparkle above, whether a bulk backup should count once or per vault, and whether the balance-delta heuristic is the right stand-in for a confirmed incoming transaction.

https://claude.ai/code/session_011rkQQHDBNjgLxyeZbKgz1A

## New Features
- Added in-app review prompts after meaningful milestones, including backups, restores, device pairing, confirmed transactions, and incoming funds.
- Added a “Rate the App” option in Settings that opens the app’s store listing.
- Review prompts are centrally managed, deduplicated, and limited to appropriate opportunities.

## Style
- Reorganized Settings into clearer Support, Community, and About sections.

## Localization
- Added “Rate the App” translations for German, Spanish, Croatian, Italian, Korean, Dutch, Portuguese, Russian, and Simplified Chinese.


app/src/main/java/com/vultisig/wallet/ui/models/BackupPasswordViewModel.kt
```diff
@@ -317,16 +322,21 @@ constructor(
             if (backupSuccess) {
                 when (backupType) {
                     BackupType.AllVaults -> {
-                        awaitVaults().forEach { vault ->
+                        val vaults = awaitVaults()
+                        vaults.forEach { vault ->
                             vaultDataStoreRepository.setBackupStatus(vault.id, true)
                         }
+                        inAppReviewRepository.recordFirstVaultBackupCompleted(vaults)
```
johnnyluo: recordFirstVaultBackupCompleted/recordAndOfferPrompt calls here aren't wrapped in try/catch, unlike the identical code path in BackupPasswordRequestViewModel.kt and BackupVaultViewModel.kt. If inAppReviewRepository throws (e.g. a DataStore IO error), the exception propagates uncaught out of the viewModelScope.launch, skipping the subsequent success snackbar even though the backup file was already written and setBackupStatus already persisted.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/InAppReviewRepository.kt
```diff
@@ -69,50 +177,221 @@ constructor(
...
+            val eventId = event.storageId ?: return@withContext false
+            var counted = false
+            appDataStore.editData { preferences -> counted = countEvent(preferences, eventId) }
+            counted
+        }
 
-            val now = clock.nowMillis()
-            if (now - installTime.firstInstallMillis() < MIN_INSTALL_AGE.inWholeMilliseconds) {
-                return@withContext false
+    override suspend fun onConfirmedBalanceObserved(
```
johnnyluo, line 213: rememberBalance() writes to DataStore unconditionally on every call. TokenDetailViewModel.recordIncomingFundsIfNeeded runs on every emission of the account-refresh polling flow, so as long as a token detail screen is open, every tick triggers appDataStore.editData and rewrites the whole Preferences file, even when the balance hasn't changed — needless disk I/O on a polling hot path.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/InAppReviewRepository.kt
```diff
@@ -69,50 +177,221 @@ constructor(
...
+        withContext(Dispatchers.IO) {
+            val id = coinId.trim()
+            if (id.isEmpty() || balance.signum() < 0) return@withContext false
+
+            var counted = false
+            // One transaction for read-compare-write: two concurrent refreshes of the same coin
+            // would otherwise both see the old balance and both count the same arrival.
+            appDataStore.editData { preferences ->
+                val previous = lastSeenBalances(preferences)[id]?.toBigIntegerOrNull()
+                rememberBalance(preferences, id, balance)
```
johnnyluo, line 229: The incoming-funds dedupe key embeds the absolute resulting balance, so a balance that returns to a previously-seen exact value is silently treated as already counted. Example: 0 -> 200 (counted, storageId "incoming:coinId:200" persisted) -> 150 (spend) -> 200 again (a genuine new arrival): the second rise to 200 produces the identical storageId, countEvent() sees it already in COUNTED_EVENT_IDS_KEY, and the arrival silently fails to count or offer a review prompt.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/InAppReviewRepository.kt
```diff
@@ -2,35 +2,138 @@ package com.vultisig.wallet.data.repositories
...
+    if (record(event)) {
+        requestPromptEvaluation()
+    }
+}
+
+/**
+ * Records a vault backup milestone for only the first vault in a bulk backup operation. One tap is
+ * one milestone: a bulk export of five vaults is a single good moment, not five.
+ */
+suspend fun InAppReviewRepository.recordFirstVaultBackupCompleted(vaults: List<*>) {
```
johnnyluo: recordFirstVaultBackupCompleted takes List<*> and does a runtime 'is Vault' check with a silent no-op fallthrough, instead of just List<Vault> — Vault lives in the same data module so there's no dependency reason for the type erasure. Both call sites (BackupPasswordViewModel, BackupPasswordRequestViewModel) already have List<Vault> on hand; a future caller passing a non-Vault list would silently no-op via 'else -> return@let' instead of failing to compile.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/InAppReviewRepository.kt
aminsato: Thanks for catching this! Fixed in the latest commit:

`rememberBalance()` now checks if the balance actually changed before writing:
```kotlin
if (existingEntry == "$coinId$BALANCE_SEPARATOR$balance") return
```

This prevents unnecessary disk I/O when `onConfirmedBalanceObserved` is called repeatedly with the same balance during polling.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/InAppReviewRepository.kt
aminsato: Thanks for the detailed explanation! Fixed in the latest commit:

Changed the dedupe key from embedding the final balance to embedding the millisecond timestamp captured at event creation time:

```kotlin
data class ConfirmedIncomingFunds(
    val coinId: String,
    val balance: BigInteger,
    val observationTimeMillis: Long = System.currentTimeMillis(),
) : AppReviewEvent {
    override val storageId: String?
        get() = storageId("incoming", "$coinId:$observationTimeMillis")
}
```

Now:
- Balance 0 → 200 (created at t=100) → storageId "incoming:coinId:100" (counted)
- Spend to 150 → New arrival to 200 (created at t=200) → storageId "incoming:coinId:200" (distinct, counts again ✓)

The timestamp captures the arrival event identity at detection time, preventing both balance-based duplicates and minute-boundary issues.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/InAppReviewRepository.kt
```diff
@@ -2,35 +2,141 @@ package com.vultisig.wallet.data.repositories
...
+
+    data class ConfirmedIncomingFunds(val coinId: String, val balance: BigInteger) :
+        AppReviewEvent {
+        // Use coinId with balance hash: if balance goes 0 -> 200 -> 150 -> 200,
+        // both increases should count as separate arrivals, not duplicates.
+        // Hash with timestamp to ensure distinct events at the same balance.
+        override val storageId: String?
+            get() {
+                val timestamp = System.currentTimeMillis() / 60000 // Round to minute for stability
+                return storageId("incoming", "$coinId:$timestamp")
```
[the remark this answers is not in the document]
aminsato: Thanks for the detailed review! Fixed in the latest commit:

Replaced minute-rounded timestamps with full millisecond precision captured at event creation time:

```kotlin
data class ConfirmedIncomingFunds(
    val coinId: String,
    val balance: BigInteger,
    val observationTimeMillis: Long = System.currentTimeMillis(),
) : AppReviewEvent {
    override val storageId: String?
        get() = storageId("incoming", "$coinId:$observationTimeMillis")
}
```

This fixes both issues:
1. **Multiple arrivals in same minute** — Each `ConfirmedIncomingFunds` created at different times gets a different `observationTimeMillis`, so two arrivals at different moments (even within the same minute) produce different storageIds
2. **Minute-boundary re-counting** — The timestamp is captured at event *creation time*, not computed lazily in `storageId`, so it remains stable and won't change after a minute passes

The identity is now atomic and persistent per-arrival event.

johnnyluo:
This adds a lot of complexity for what should be a simple flow. Suggest scoping this way instead:

- Ask for a review at one of two moments: right after a vault is created, or right after the user sends their first transaction — whichever comes first.
- Make it one-off: if the user reviews, never ask again. If they decline/dismiss, it's fine to ask again later (e.g. ~1 month out).
- Drop the five qualifying-event tracking, per-version gating, and balance-delta inference — that's a lot of surface area (ring buffers, cross-vault state, multiple producers across ViewModels) for a prompt that just needs to fire once and respect a no.

Given the size of the rework needed, let's close this off and redo it with the simpler approach — happy to sync offline on details.

PR #5760, opened 2026-08-30
feat(review): ask for a review at vault creation or a first send

### LINKED ISSUE #5700: Review prompt: expand eligibility beyond outbound transactions (2-event threshold) + per-version gate + Settings rate link

aminsato:
Closes #5700

Redo of #5743, which was closed as over-built. This follows the closing comment's shape rather than the issue's original one.

## Problem

The prompt shipped in #5427 waited for **three** confirmed outbound transactions behind a 7-day install age and a 120-day cooldown, and existed only on `KeysignScreen` / `JoinKeysignView`. A hold-heavy vault's users mostly receive rather than send, so the eligible pool was near zero by construction.

## Fix

**Two moments, whichever comes first.** A vault has just been created, or a transaction has just gone through. That is the whole eligibility rule — no event ring, no per-version gate, no balance-delta inference.

**One cooldown, 30 days.** Play reports neither whether the card appeared nor what the user did with it, and it suppresses the card by itself for anyone who has already rated. So the only case the app has to own is a dismissal, and a single 30-day floor covers it. An ask is spent when it is requested, not when it is shown.

**The 7-day install age is gone.** The issue asks to keep it, but "right after a vault is created" is day 0 for every new install — keeping the gate would have left the new trigger dead for exactly the users it targets. Flagging it explicitly since it contradicts item 3 of the issue.

**Presentation moved to the app root.** `InAppReviewHost` in `MainActivity` replaces `InAppReviewEffect` on the two keysign screens. Keygen navigates straight from success into the backup flow, so the screen that earns the card cannot host it. The pending flag is persisted, so a user who backgrounds the app during the settle delay is asked on the next launch instead of losing the ask.

**Vault creation has two producers, not one.** A fast vault is parked in `TemporaryVaultRepository` until the emailed code verifies, so it is not a real vault when `KeygenViewModel` finishes — `FastVaultVerificationViewModel` records the fast path, `KeygenViewModel.saveVault()` the secure one. Both gate on `TssAction.KEYGEN`, so reshare, migrate and key-import are untouched.

**Settings gains a "Rate the App" row** under Support, linking out to the Play listing. Play's policy forbids a button from triggering the in-app flow, and that flow no-ops once its quota is spent, which would leave the row looking broken. Translated in all ten locales.

## Upgrade path

The `in_app_review/last_prompted_at` key from #5427 is reused as-is, so an upgrader already prompted under the old rules keeps cooling down from that prompt. The old `in_app_review/successful_transactions` key is left orphaned rather than migrated — nothing reads it any more.

## Worth a second opinion

- **The card can land on "back up your vault"** ~2 s after a vault is created. That is what "right after a vault is created" means in Android's flow, since creation navigates straight into backup. Making the host wait for Home instead would mean watching the nav backstack — happy to add it if that is preferred.
- **This diverges from iOS on purpose.** `AppReviewPolicy.swift` on iOS `main` ships the version #5743 was porting (2 events + 7-day install + per-version gate + 14-day floor). Android is now deliberately the simpler platform.

## Verification

- `:app` + `:data` unit tests: 5563 tests, 0 failures.
- `:app:lintDebug` + `:data:lintDebug` clean, `ktfmt` applied.
- `InAppReviewRepositoryImplTest` rewritten (8 tests) for the new policy; `FastVaultVerificationViewModelTest` gains two covering the fast-vault creation moment and its absence on an upgrade. The secure-vault call site needs the TSS natives to reach, so it is manual-only.

## Which feature is affected?
- [x] Create vault ( Secure / Fast)
- [x] Sending
- [ ] Swap

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added a “Rate the App” option in Settings that opens the Google Play listing.
  - Added localized labels for supported languages.
  - Added in-app review prompts after eligible vault creation or successful transactions.
  - Review prompts now appear only when the app is active and unlocked.

- **Improvements**
  - Reorganized Settings into clearer Support, Community, and About sections.
  - Review tracking failures no longer interrupt key generation, verification, or transaction flows.


app/src/main/java/com/vultisig/wallet/ui/utils/InAppReviewHost.kt
```diff
@@ -0,0 +1,79 @@
...
+    val isPromptPending by viewModel.isPromptPending.collectAsStateWithLifecycle(false)
+
+    LaunchedEffect(isPromptPending, reviewManager) {
+        if (!isPromptPending) return@LaunchedEffect
+
+        delay(PROMPT_SETTLE_DELAY)
+
+        // An ask is spent whether or not Play shows anything, so it must never be spent while the
+        // app is in the background: the prompt stays pending and is presented on the next launch.
+        if (!lifecycle.currentState.isAtLeast(Lifecycle.State.RESUMED)) return@LaunchedEffect
```
rkokhatskyi: `launchReviewFlow` starts `PlayCoreDialogWrapperActivity` via `Activity.startActivity`, so it draws above the lock's `Dialog` window while `MainActivity` is still RESUMED, and `InAppReviewHost()` at `MainActivity.kt:157` is mounted outside `LocalIsGateClosed`, so a pending card lands on the passcode screen and spends the ask there

app/src/main/java/com/vultisig/wallet/app/activity/MainActivity.kt
```diff
@@ -150,6 +151,11 @@ class MainActivity : AppCompatActivity() {
 
                             CheckUpdates()
 
+                            // Hosted at the root, not on the screen that earned it: creating a
+                            // vault navigates straight on into backup, which is no place for a
```
rkokhatskyi: This comment should state the real reason because the fast path calls `onVaultCreated()` at `FastVaultVerificationViewModel.kt:195`, routes to `Route.BackupVault` 400ms later, then requests the card 1.6s after that, so root hosting does not keep it off backup

data/src/main/kotlin/com/vultisig/wallet/data/repositories/InAppReviewRepository.kt
```diff
@@ -1,118 +1,94 @@
...
         }
+        Timber.i("In-app review: prompt requested")
+    }
 
     /**
-     * Increments the counter inside a single DataStore transaction and returns the new total, so
-     * concurrent successes can't read-modify-write over each other.
+     * Marks the card as owed unless a previous ask is still cooling down.
+     *
+     * The read and the write share one transaction so two moments landing together cannot both
```
rkokhatskyi: This comment should be removed because the transform reads `LAST_PROMPTED_AT_KEY` but only writes `PROMPT_PENDING_KEY`, and the timestamp is written solely in `onPromptRequested`, so two moments landing together both read the same value and both decide the cooldown has elapsed

app/src/test/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModelInAppReviewTest.kt
```diff
@@ -106,10 +96,10 @@ internal class KeysignViewModelInAppReviewTest {
             coVerify(exactly = 0) { inAppReviewRepository.onTransactionSucceeded() }
         }
 
-    // Status polling re-emits KeysignFinished on every tick; counting each one would inflate the
-    // success counter and re-ask on every poll.
+    // Status polling re-emits KeysignFinished on every tick; recording each one would re-arm the
```
rkokhatskyi: `repeated terminal emissions record only once` passes with `.collect` too because `KeysignUiState` is a data class and the second `finishWith(Confirmed)` produces an equal value that `StateFlow` conflates, so the `.first` guard at `KeysignViewModel.kt:456` has no test behind it

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/utils/InAppReviewHost.kt
aminsato: Confirmed, and fixed in 7935b8d.

`InAppReviewHost()` moves out of `MainActivity` into `MainActivityContent`, where `guardState` already lives, and is composed only while the gate is open. Locking cancels an in-flight prompt; unlocking recomposes the host and restarts the settle delay, and the pending flag is persisted, so the ask waits rather than being lost.

That gate only means something if cancellation actually stops the launch, so `requestReviewFlow()` is now awaited inside the effect (`suspendCancellableCoroutine`) rather than launched from a completion listener — the listener fired whatever became of the coroutine that registered it, including after the app had locked or backgrounded. `withResumed` now wraps the launch, not the request.

No test behind it: the gating is structural at the call site, and an instrumented test would need `ReviewManagerFactory` talking to real Play services.

app/src/main/java/com/vultisig/wallet/app/activity/MainActivity.kt
aminsato: Right — I checked the timings and the comment was wrong. `onVaultCreated()` at `FastVaultVerificationViewModel.kt:195`, `Route.BackupVault` 400ms later (`FAST_VAULT_VERIFICATION_SUCCESS_DELAY`), card due 1.6s after that, so it lands on backup either way.

The comment is gone with the move. The one at the new call site in `MainActivityContent` states the two reasons that are actually true: the screen which earned the moment is gone by the time the card is due, and the card is an activity of its own drawn over the lock's window.

data/src/main/kotlin/com/vultisig/wallet/data/repositories/InAppReviewRepository.kt
aminsato: Correct — the transform reads `LAST_PROMPTED_AT_KEY` and only ever writes `PROMPT_PENDING_KEY`; the timestamp is written solely in `onPromptRequested`, so one transaction buys nothing here. Two moments landing together do both read the same value and both conclude the cooldown elapsed — they just both set the same flag, so the claim was wrong rather than the behaviour.

Comment removed, one-line summary kept.

app/src/test/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModelInAppReviewTest.kt
aminsato: Confirmed empirically, in both directions. Swapping `.first` for `.filter {}.collect {}` in `KeysignViewModel`: the old `Confirmed → Confirmed` sequence still passed, so the guard had nothing behind it exactly as you said.

The sequence is now `Pending → Broadcasted → Confirmed` — two distinct successful statuses, which is also the real polling path — and that fails against the `.collect` implementation and passes against `.first`. The comment above the test now says why the statuses have to differ, so it does not regress back to two identical ones.

johnnyluo (approved):
LGTM

PR #5761, opened 2026-08-30
fix(keysign): sign Maya dApp requests with the Maya chain id

### LINKED ISSUE #5756: [Fix] MayaChain dApp signDirect/signAmino are signed with THORChain's chain_id [Keysign]
# [Fix] MayaChain dApp signDirect/signAmino are signed with THORChain's chain_id [Keysign]

## Problem

`ThorChainHelper` serves both THORChain and MayaChain, but its two dApp co-signing builders take the chain id from a hardcoded companion constant instead of the per-instance network id. In `data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt`:

- `coinType` is declared on the companion object at line 75 as `private val coinType: CoinType = CoinType.THORCHAIN`, and both factories — `maya()` (lines 55-63) and `thor()` (lines 65-73) — hang off that same companion, so neither can override it.
- `buildSignDirectSigningInput` (line 488) and `buildAminoSigningInput` (line 513) both call `.setChainId(coinType.chainId())`.
- `CoinType.THORCHAIN.chainId()` resolves to `"thorchain-1"` in the shipped wallet-core registry. MayaChain is not a wallet-core coin at all — it is modelled as `CoinType.THORCHAIN` with `hrp = "maya"` — so there is no value of `coinType` that could ever yield `"mayachain-mainnet-v1"`.

The correct value is already available on the instance. `networkId` is a constructor parameter (line 43) set to `"mayachain-mainnet-v1"` for Maya (line 61) and to `THORCHAIN_NETWORK_ID` for THOR (line 71), and the same file uses it correctly on the native deposit/wasm/send paths at lines 176, 212 and 233. Only these two dApp builders bypass it.

This is reachable in production, not theoretical: `SigningHelper.kt:332` (`getPreSignedImageHash`) and `SigningHelper.kt:608` (`getSignedTransaction`) both construct `ThorChainHelper.maya(...)` for `Chain.MayaChain`, and the signAmino/signDirect early-returns in `getPreSignInputData` (lines 130 and 143) fire before any other message construction.

Result: every MayaChain dApp signAmino/signDirect request produces a signature over `chain_id = "thorchain-1"`. It is self-consistent but fails the Cosmos SDK ante-handler's chain-id check at broadcast — and it fails only *after* a full MPC keysign ceremony has completed, so the user pays a full multi-device round trip to reach an opaque error.

Two related gaps on the same two lines:

1. **The dApp's own chain id is never checked.** `buildSignDirectSigningInput` reads only `signDirect.bodyBytes` (line 467) and `signDirect.authInfoBytes` (line 477); it never reads the request's `chainId`, and the account number comes from wallet state (`cosmosSpecific.accountNumber`, line 489) rather than the request. There is no comparison between a dApp-supplied chain id and the resolved network id anywhere in the codebase — the only production readers of `signDirect.chainId` are a non-blank assertion in `ParseCosmosMessageUseCase.kt:33` and a copy into a display model at `:43`. (This applies to signDirect only: the `SignAmino` message carries no chainId field, so the amino builder has nothing to validate — it is simply the sole source of the value and currently supplies the wrong one.)

2. **THORChain is masked today but has silently lost its live chain-id tracking.** `THORCHAIN_NETWORK_ID` is a mutable companion `var` (line 50) that `InitializeThorChainNetworkIdUseCase` refreshes from cache and then from the network at app start. Because the dApp builders read `coinType.chainId()` instead, they ignore that refresh entirely and use wallet-core's baked-in constant. It happens to match the live value right now, so nothing is visibly broken for THORChain — but a THORChain chain-id rotation would break THOR dApp signing with no code change, and the whole point of that use case is that THORChain's chain id has rotated before.

For contrast, `CosmosHelper` makes the same-looking `coinType.chainId()` calls (lines 100, 215, 240, 256, 420, 446), but there `coinType` is a constructor parameter (`CosmosHelper.kt:35`), so it is per-instance and correct. `ThorChainHelper` is the only place where one hardcoded `CoinType` serves two distinct networks.

Both builders were introduced together and the file has not been touched since 2026-03-27.

## Expected Behavior

A MayaChain dApp signDirect/signAmino request is signed with `mayachain-mainnet-v1`, and a THORChain one with whatever `THORCHAIN_NETWORK_ID` currently resolves to, so the signature passes the ante-handler's chain-id verification. When a signDirect request carries a chain id that disagrees with the resolved network id, the request is rejected before the keysign ceremony starts rather than silently overridden.

## Steps to Reproduce

1. Set up a vault with a MayaChain address.
2. From a Keplr-compatible MayaChain dApp, trigger a `signAmino` or `signDirect` request (for example a governance vote or a bank interaction).
3. Complete the co-signing ceremony on all devices.
4. The broadcast is rejected by the node for a chain-id mismatch. Inspecting the signing input shows `chain_id = "thorchain-1"` where the network expects `mayachain-mainnet-v1`.

## Solution

Replace `coinType.chainId()` with the instance field `networkId` at `ThorChainHelper.kt:488` and `:513`, matching the pattern the same file already uses at lines 176, 212 and 233. This is a two-line change that fixes Maya outright and simultaneously reconnects THORChain to its runtime-refreshed id. Then add a pre-ceremony guard in `buildSignDirectSigningInput` that rejects the request when the dApp-supplied `chainId` is non-blank and does not equal the resolved `networkId`, so a mismatch surfaces before an MPC round trip instead of after it.

### Must Do
- [ ] Change `.setChainId(coinType.chainId())` to use the instance `networkId` in `buildSignDirectSigningInput` (`ThorChainHelper.kt:488`).
- [ ] Change `.setChainId(coinType.chainId())` to use the instance `networkId` in `buildAminoSigningInput` (`ThorChainHelper.kt:513`).
- [ ] In `buildSignDirectSigningInput`, reject the request when the dApp-supplied `chainId` is present and disagrees with `networkId`, failing before the keysign ceremony with a clear message.
- [ ] Add unit tests asserting that a Maya-constructed helper produces `chain_id = "mayachain-mainnet-v1"` and a THOR-constructed one produces the current `THORCHAIN_NETWORK_ID`, for both the signDirect and the signAmino builder.
- [ ] Add a unit test asserting a mismatched dApp `chainId` is rejected.

### Must NOT Do
- Don't change unrelated code
- Don't touch the native deposit/wasm/send paths at lines 176, 212 and 233 — they are already correct and are the reference
- Don't change how `THORCHAIN_NETWORK_ID` is fetched or cached
- Don't change `CosmosHelper` — its `coinType` is a constructor parameter and is correct

## Acceptance Criteria
- [ ] A MayaChain dApp signDirect request builds a signing input with `chain_id = "mayachain-mainnet-v1"`
- [ ] A MayaChain dApp signAmino request builds a signing input with `chain_id = "mayachain-mainnet-v1"`
- [ ] A THORChain dApp request builds a signing input carrying the runtime-resolved `THORCHAIN_NETWORK_ID`, so overriding that value in a test changes the signed chain id
- [ ] A signDirect request whose `chainId` disagrees with the resolved network id is rejected before any keysign ceremony begins

aminsato:
## Summary

`ThorChainHelper` serves both THORChain and MayaChain, but its two dApp co-signing builders took the chain id from the companion `coinType` constant instead of the instance's `networkId`:

- `coinType` is `CoinType.THORCHAIN` on the companion object, so neither the `maya()` nor the `thor()` factory can override it, and `CoinType.THORCHAIN.chainId()` resolves to `"thorchain-1"`.
- MayaChain is not a wallet-core coin at all — it is modelled as `CoinType.THORCHAIN` with `hrp = "maya"` — so no value of `coinType` could ever yield `mayachain-mainnet-v1`.

Every MayaChain dApp `signAmino`/`signDirect` request was therefore signed over `chain_id = "thorchain-1"`. It is self-consistent, so it fails only at the Cosmos ante-handler's chain-id check on broadcast — after a full MPC keysign ceremony has completed and the user has paid a multi-device round trip.

## Changes

- **Both dApp builders read `networkId`** — the value the same file's native deposit/wasm/send paths already use. Maya signs `mayachain-mainnet-v1`; THORChain signs whatever `THORCHAIN_NETWORK_ID` currently holds, so it is reconnected to the id `InitializeThorChainNetworkIdUseCase` refreshes at startup instead of wallet-core's baked-in constant. Nothing is visibly broken for THOR today, but a chain-id rotation would have broken THOR dApp signing with no code change — which is the whole reason that use case exists.
- **A `signDirect` request that carries its own chain id is validated.** When it is non-blank and disagrees with the resolved network id, the request is rejected before the ceremony starts rather than silently overridden. The initiator surfaces it through `KeysignFlowViewModel.setData`'s error state and a co-signer through `KeysignMessagesException`, so it reads as an error on the Join Keysign screen, not a crash. (`SignAmino` carries no chain id, so the amino builder has nothing to validate — it was simply supplying the wrong value.)

Untouched, as the issue specifies: the native paths that were already correct, how `THORCHAIN_NETWORK_ID` is fetched and cached, and `CosmosHelper` (its `coinType` is a constructor parameter and is already per-instance).

## Note on the diff size

The issue scopes this as a two-line change. The two lines are the fix, but they are not coverable as they stand: both builders take a wallet-core `PublicKey`, and `buildSignDirectSigningInput` decodes through the JNI `Base64`. `:data` JVM unit tests have no native library, so a test reaching those lines would `assumeTrue`-skip and pin nothing — the same thing `UtxoHelperTest` and `SwapKitBtcSignerTest` already do today.

So the `Cosmos.SigningInput` header both builders write is extracted into `buildDappSigningInput`, with the chain-id decision in `resolveDappChainId`. Both are JNI-free and directly testable, and the bytes each builder produces are unchanged.

## Test plan

New `ThorChainHelperDappChainIdTest` — 8 cases, verified as actually executed (`tests="8" skipped="0"`), not skipped:

- a Maya-constructed helper produces `mayachain-mainnet-v1` in Protobuf (signDirect) and JSON (signAmino) signing mode
- a THOR-constructed helper picks up a mutated `THORCHAIN_NETWORK_ID`, proving the value is no longer wallet-core's constant
- a request carrying the matching chain id is accepted; blank and whitespace-only fall back to the resolved id
- a Maya helper rejects `thorchain-1` and a THOR helper rejects `mayachain-mainnet-v1`
- the rest of the header (signing mode, public key, account number, sequence, broadcast mode) is unchanged

Verified locally:

- `./gradlew :data:testDebugUnitTest` — 3320 pass, 0 failures, 39 skipped (all pre-existing wallet-core JNI `assumeTrue` skips)
- `./gradlew assembleDebug` — successful
- `./gradlew :data:lintDebug` — successful

Manual, on device — a MayaChain `signAmino` request raised from a local Keplr-compatible page against the browser extension, then co-signed from the phone:

| dApp request (extension) | Ceremony result (Android) |
|---|---|
| maya signAmino request | signed on device |

The same request on a build without this change never reached a signature: the joining device failed at the DKLS setup-message exchange with `HTTP 404:`, surfaced as "One of your devices didn't respond in time". The setup message is stored under `md5(messageToSign)` (`DklsKeysign.kt:330`), so the phone signing over `thorchain-1` while the initiator used `mayachain-mainnet-v1` derives a different key and fetches a message that was never stored — the same failure mode `KeysignPayloadProtoMapperSignDirectTest` already pins for the LUNA/LUNC staking bug. With the fix, both devices derive the same hash and the ceremony completes.

A THORChain dApp request is unchanged.

Closes #5756

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dApp transaction signing for MayaChain and THORChain.
  * Validates requested chain IDs before signing and rejects mismatched requests.
  * Uses the configured network when resolving chain IDs.
  * Preserves other signing details while providing clearer validation errors.
  * Improved transaction memo handling for SignDirect requests.
  * Corrected MayaChain and THORChain address formatting based on the selected network.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt
@@ -473,7 +473,6 @@ class ThorChainHelper(
             )
 
         val bodyBytes = Base64.decode(signDirect.bodyBytes)
-        val txBody = Cosmos.SigningInput.parseFrom(bodyBytes)
 
         val message =
             Cosmos.Message.newBuilder()
@@ -490,12 +489,7 @@ class ThorChainHelper(
                 .build()
 
         inputBuilder.addMessages(message).setFee(buildCosmosFee(cosmosSpecific, denom = denom))
-
-        if (txBody.memo.isNotEmpty()) {
-            inputBuilder.memo = txBody.memo
-        } else {
-            memo?.let { inputBuilder.memo = it }
-        }
+        inputBuilder.memo = cosmosTxBodyMemo(bodyBytes) ?: memo.orEmpty()
 
         return inputBuilder.build().toByteArray()
     }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt
@@ -464,6 +464,14 @@ class ThorChainHelper(
         publicKey: PublicKey,
         denom: String,
     ): ByteArray {
+        val inputBuilder =
+            buildDappSigningInput(
+                signingMode = Cosmos.SigningMode.Protobuf,
+                cosmosSpecific = cosmosSpecific,
+                publicKeyData = publicKey.data(),
+                requestChainId = signDirect.chainId,
+            )
+
         val bodyBytes = Base64.decode(signDirect.bodyBytes)
         val txBody = Cosmos.SigningInput.parseFrom(bodyBytes)
 
@@ -481,16 +489,8 @@ class ThorChainHelper(
                 )
                 .build()
 
-        val inputBuilder =
-            Cosmos.SigningInput.newBuilder()
-                .setPublicKey(ByteString.copyFrom(publicKey.data()))
-                .setSigningMode(Cosmos.SigningMode.Protobuf)
-                .setChainId(coinType.chainId())
-                .setAccountNumber(cosmosSpecific.accountNumber.toLong())
-                .setSequence(cosmosSpecific.sequence.toLong())
-                .setMode(Cosmos.BroadcastMode.SYNC)
-                .addMessages(message)
-                .setFee(buildCosmosFee(cosmosSpecific, denom = denom))
+        inputBuilder.addMessages(message).setFee(buildCosmosFee(cosmosSpecific, denom = denom))
+
         if (txBody.memo.isNotEmpty()) {
             inputBuilder.memo = txBody.memo
         } else {
@@ -507,13 +507,12 @@ class ThorChainHelper(
         memo: String?,
     ): ByteArray {
         val inputBuilder =
-            Cosmos.SigningInput.newBuilder()
-                .setPublicKey(ByteString.copyFrom(publicKey.data()))
-                .setSigningMode(Cosmos.SigningMode.JSON) // Use JSON mode for Amino
-                .setChainId(coinType.chainId())
-                .setAccountNumber(cosmosSpecific.accountNumber.toLong())
-                .setSequence(cosmosSpecific.sequence.toLong())
-                .setMode(Cosmos.BroadcastMode.SYNC)
+            buildDappSigningInput(
+                signingMode = Cosmos.SigningMode.JSON, // Use JSON mode for Amino
+                cosmosSpecific = cosmosSpecific,
+                publicKeyData = publicKey.data(),
+                requestChainId = null,
+            )
 
         signAmino.msgs.forEach { cosmosMsg ->
             val message =
@@ -540,6 +539,37 @@ class ThorChainHelper(
         memo?.let { inputBuilder.memo = it }
 
         return inputBuilder.build().toByteArray()
+    }
+
+    /**
+     * Header shared by the two dApp co-signing builders. The chain id comes from [networkId] — the
+     * network this helper instance was built for — because `coinType` is a THORChain constant that
+     * MayaChain, which wallet-core models as THORChain with a `maya` hrp, can never resolve to.
+     */
+    internal fun buildDappSigningInput(
+        signingMode: Cosmos.SigningMode,
+        cosmosSpecific: BlockChainSpecific.Cosmos,
+        publicKeyData: ByteArray,
+        requestChainId: String?,
+    ): Cosmos.SigningInput.Builder =
+        Cosmos.SigningInput.newBuilder()
+            .setPublicKey(ByteString.copyFrom(publicKeyData))
+            .setSigningMode(signingMode)
+            .setChainId(resolveDappChainId(requestChainId))
+            .setAccountNumber(cosmosSpecific.accountNumber.toLong())
+            .setSequence(cosmosSpecific.sequence.toLong())
+            .setMode(Cosmos.BroadcastMode.SYNC)
+
+    /**
+     * Rejects a dApp request whose own chain id disagrees with [networkId] rather than silently
+     * signing it for the other network, which only fails after a full keysign ceremony.
+     */
+    internal fun resolveDappChainId(requestChainId: String?): String {
+        val requested = requestChainId?.trim().orEmpty()
+        require(requested.isEmpty() || requested == networkId) {
+            "dApp requested chain id \"$requested\", but this vault signs \"$networkId\""
+        }
+        return networkId
     }
 
     private fun buildCosmosFee(
--- a/data/src/test/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelperDappChainIdTest.kt
+++ b/data/src/test/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelperDappChainIdTest.kt
@@ -1,0 +1,123 @@
+package com.vultisig.wallet.data.crypto
+
+import com.google.protobuf.ByteString
+import com.vultisig.wallet.data.models.payload.BlockChainSpecific
+import java.math.BigInteger
+import kotlin.test.assertEquals
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Test
+import vultisig.keysign.v1.TransactionType
+import wallet.core.jni.proto.Cosmos
+
+/**
+ * MayaChain is signed by a [ThorChainHelper] built with `CoinType.THORCHAIN`, so the chain id can
+ * only come from the instance's own `networkId`. Taking it from the coin type produced
+ * `thorchain-1` for every Maya dApp request, and the mismatch surfaced only when the node rejected
+ * the broadcast — after a full keysign ceremony.
+ */
+class ThorChainHelperDappChainIdTest {
+
+    private val originalThorNetworkId = ThorChainHelper.THORCHAIN_NETWORK_ID
+
+    @AfterEach
+    fun tearDown() {
+        ThorChainHelper.THORCHAIN_NETWORK_ID = originalThorNetworkId
+    }
+
+    @Test
+    fun `maya signDirect header carries the maya chain id`() {
+        assertEquals(MAYA_NETWORK_ID, maya().dappHeader(Cosmos.SigningMode.Protobuf, "").chainId)
+    }
+
+    @Test
+    fun `maya signAmino header carries the maya chain id`() {
+        assertEquals(MAYA_NETWORK_ID, maya().dappHeader(Cosmos.SigningMode.JSON, null).chainId)
+    }
+
+    @Test
+    fun `thor header carries the runtime network id, not the wallet-core constant`() {
+        ThorChainHelper.THORCHAIN_NETWORK_ID = "thorchain-2"
+
+        assertEquals(
+            "thorchain-2",
+            thor().dappHeader(Cosmos.SigningMode.Protobuf, "thorchain-2").chainId,
+        )
+    }
+
+    @Test
+    fun `a request carrying the resolved chain id is accepted`() {
+        assertEquals(MAYA_NETWORK_ID, maya().resolveDappChainId(MAYA_NETWORK_ID))
+    }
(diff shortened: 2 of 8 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt
```diff
@@ -464,6 +464,14 @@ class ThorChainHelper(
         publicKey: PublicKey,
         denom: String,
     ): ByteArray {
+        val inputBuilder =
+            buildDappSigningInput(
+                signingMode = Cosmos.SigningMode.Protobuf,
+                cosmosSpecific = cosmosSpecific,
+                publicKeyData = publicKey.data(),
+                requestChainId = signDirect.chainId,
```
rkokhatskyi, line 472: Maya signDirect ceremonies now complete, but the joiner's verify card renders a `maya1` recipient as `thor1` because `ParseCosmosMessageUseCase.kt:30` pins `Bech32.encode(THOR_BECH32_HRP, it)` for `/types.MsgSend`, so the approved address matches neither the dApp nor the `dstAddress` row

data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt
```diff
@@ -464,6 +464,14 @@ class ThorChainHelper(
...
+            )
+
         val bodyBytes = Base64.decode(signDirect.bodyBytes)
         val txBody = Cosmos.SigningInput.parseFrom(bodyBytes)
```
rkokhatskyi: This parses the dApp's cosmos `TxBody` with `Cosmos.SigningInput`, where memo is field 5 while `TxBody`'s memo is field 2, so the dApp memo never reaches `txBody.memo` and the precedence branch at :494 cannot honour it. Windows decodes a real `TxBody` and applies `txBody.memo || memo`

data/src/test/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelperDappChainIdTest.kt
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.jupiter.api.AfterEach
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Test
+import vultisig.keysign.v1.TransactionType
+import wallet.core.jni.proto.Cosmos
+
+/**
+ * MayaChain is signed by a [ThorChainHelper] built with `CoinType.THORCHAIN`, so the chain id can
+ * only come from the instance's own `networkId`. Taking it from the coin type produced
+ * `thorchain-1` for every Maya dApp request, and the mismatch surfaced only when the node rejected
```
rkokhatskyi: This says the mismatch surfaced only when the node rejected the broadcast. The extension initiator signs Maya as `mayachain-mainnet-v1` and each party keys relay traffic on `message.md5()` (`KeysignViewModel.kt:828`), so the preimages diverge and the ceremony stalls before any broadcast

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt
aminsato: Confirmed, fixed in ee34299.

`ParseCosmosMessageUseCaseImpl`'s `@Inject` constructor pinned the encoder, so both `/types.MsgSend` and `/types.MsgDeposit` rendered `thor1`. MayaChain forks THORChain and reuses those type urls with the same raw 20-byte addresses, so only the request's own chain id says which hrp they belong to.

The seam is now `(String, ByteArray) -> String` (the Hilt constructor is just `Bech32::encode`) and the hrp is resolved per request from `signDirect.chainId` — `mayachain*` renders `maya1`, everything else keeps `thor`. Two tests in `ParseCosmosMessageTest` cover a Maya `MsgSend` and a Maya `MsgDeposit` signer.

data/src/main/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelper.kt
aminsato: Confirmed, fixed in ee34299.

Checked against the generated wallet-core 4.7.0 source: `Cosmos.SigningInput` has `MEMO_FIELD_NUMBER = 5` and `ACCOUNT_NUMBER_FIELD_NUMBER = 2`. `TxBody.memo` is field 2, wire type 2 (tag `0x12`), which matches no `SigningInput` case, so it went to unknown fields and `txBody.memo` was always empty — the precedence branch could never fire. Worse in the other direction: `TxBody.timeout_timestamp` (field 5, wire type 2) *does* collide with `SigningInput.memo` and would have decoded as a garbage memo string.

New `cosmosTxBodyMemo` walks the real `TxBody` for field 2 with protobuf's `CodedInputStream`, returning null (never throwing) on a body with no memo or one it cannot walk, so both call sites are `inputBuilder.memo = cosmosTxBodyMemo(bodyBytes) ?: memo.orEmpty()`. Same shape as iOS `CosmosSignDirectParser.extractMemo`.

`CosmosHelper.kt:400` carried the byte-identical defect for every other cosmos chain, so it is fixed in the same commit rather than left as a half-fix. `CosmosTxBodyMemoTest` covers the round-trip, the empty/malformed/absent cases, a memo behind a later field, and pins the actual bug: `Cosmos.SigningInput.parseFrom(bodyBytes).memo` is `""` for a body whose memo decodes fine.

data/src/test/kotlin/com/vultisig/wallet/data/crypto/ThorChainHelperDappChainIdTest.kt
aminsato: You are right — the KDoc is corrected in ee34299.

`KeysignViewModel.kt:828` sets the relay message id to `message.md5()`, so a chain id that differs from the initiator's gives a different preimage, a different message id, and the parties never exchange TSS traffic at all. It stalls mid-keysign; nothing ever reaches a node.

The KDoc now says that, and the commit body carries the same correction against the wording in be737b8.

johnnyluo (approved):
LGTM

PR #5768, opened 2026-08-30
fix(deposit): THORChain Bond fee units + consolidate Bond/Unbond flow

### LINKED ISSUE #5757: [Fix] THORChain Bond operator fee is scaled x100 on the Deposit form but sent raw on the Send form [DeFi]
# [Fix] THORChain Bond operator fee is scaled x100 on the Deposit form but sent raw on the Send form [DeFi]

## Problem

The app has two live THORChain BOND entry points and they disagree on what unit the operator-fee field is in. The Deposit form multiplies the entered value by 100 (`BondStrategy.kt:111-114`, `movePointRight(2)`), while the Send form treats the identical field as raw basis points with no scaling (`send/submit/BondStrategy.kt:84-93`). Both build the same `DepositMemo.Bond.Thor`, whose `operatorFee` is written verbatim into the broadcast memo (`DepositMemo.kt:44-51`), so the same keystroke produces two different on-chain fees.

The Deposit form is the one a user lands on by default: `DepositDataLoader.kt:161-171` lists `DepositOption.Bond` first for THORChain, and the chain-level Deposit action navigates with `depositType = null` (`ChainTokensViewModel.kt:164`), so `depositOptions.firstOrNull()` at `DepositDataLoader.kt:193` selects Bond and the operator-fee field renders immediately.

The two fields also label themselves differently. Send says `Operator fees (basis points)` (`bond_operator_fees_basis_point`) with hint `"0"`. Deposit says `Operator's Fee (optional)` (`deposit_form_operator_fee_title`) with hint `"0.0"` and states no unit anywhere on screen.

Concrete outcomes on the Deposit form:
- Entering `5` (intending 5 bps, as the Send form's label teaches) signs a real, irreversible **500 bps = 5%** operator fee, 100x the intent, with no warning.
- Entering `500` (intending 5%) signs `50000` bps = 500%, which the network rejects — after a full MPC keysign ceremony and a spent RUNE gas fee.

A range validator does exist on the Deposit path but does not prevent either case. `DepositFieldValidator.kt:89-94` rejects anything outside `1..100`, but:
- it is **never consulted before signing** — `DepositFormViewModel.deposit()` goes straight to `depositStrategies.getValue(...).build()` and never reads `operatorFeeError`, so submission proceeds with an inline error visible;
- it only runs `onLostFocus` (`DepositFormScreen.kt:393`) and is a no-op on blank input;
- it parses with `toIntOrNull()` while the strategy parses with `toBigDecimalOrNull()`, so `5.5` is flagged invalid yet still signs as `550`;
- its `1..100` range is a percent range, while the method it lives in, its KDoc, and the shared helper are all named for basis points — and that same helper also backs a genuinely-basis-points field (`DepositFieldInputCoordinator.kt:153-160`).

`BigDecimal.toInt()` at `BondStrategy.kt:114` is also a silent narrowing conversion, so an out-of-range entry wraps rather than failing.

For reference, the Send form's behaviour is the one that matches the rest of the product: iOS's single Bond flow labels the field "Operator-fees (Basis Points)" and sends the value unscaled, and the Windows Bond form embeds the raw basis-points value verbatim. The Deposit form is the only implementation that scales.

## Expected Behavior

Both THORChain BOND entry points accept the operator fee in raw basis points, label it as basis points, validate it against `0..10000`, and block submission when it is out of range — so the same input produces the same memo regardless of which screen the user reached Bond from.

## Steps to Reproduce

1. Open a vault holding RUNE, tap the THORChain chain row, and choose the Deposit / Functions action (do not use the DeFi tab's Bond button — that routes to the Send form).
2. Bond is preselected. Fill in a node address and enter `5` in the "Operator's Fee (optional)" field.
3. Continue to the Verify screen and inspect the memo: it reads `BOND:<node>::500`, i.e. 500 basis points.
4. Repeat the same entry via DeFi tab -> Bond (the Send form) and inspect the memo: it reads `BOND:<node>::5`.
5. Back on the Deposit form, enter `500`. An "Invalid amount" error appears under the field, but Continue still proceeds and the memo reads `BOND:<node>::50000`.

## Solution

Make the Deposit form's operator-fee field behave exactly like the Send form's, since that is the behaviour iOS and Windows already share and the one the memo format expects. Drop the `movePointRight(2)` scaling so the entered value is the basis-points value, relabel the field to name the unit, widen the validator to `0..10000`, and make the check blocking so an out-of-range value cannot reach a keysign ceremony. The two `BondStrategy` classes should ideally be unified afterwards so they cannot drift again, but that consolidation is not required to close this.

### Must Do
- [ ] Remove the `movePointRight(2)` scaling in `app/src/main/java/com/vultisig/wallet/ui/models/deposit/submit/BondStrategy.kt:111-114` so the operator fee is passed through as raw basis points.
- [ ] Change the Deposit operator-fee validation to a `0..10000` basis-points range, and make it block submission (have `DepositFormViewModel.deposit()` refuse to build the transaction while `operatorFeeError` is set), matching the Send form's blocking behaviour.
- [ ] Give the two operator-fee fields the same unit semantics: keep `validateBasisPoints` for the genuinely-basis-points field it already serves, and stop the percent-shaped `1..100` range from applying to the Bond operator fee.
- [ ] Rename `deposit_form_operator_fee_title` to state the unit (e.g. "Operator fees (basis points)"), and add the translation to all 9 non-English locale files alongside the existing key.
- [ ] Align the parsing so a decimal entry is rejected consistently instead of being flagged invalid and signed anyway.
- [ ] Update `app/src/test/java/com/vultisig/wallet/ui/models/deposit/submit/BondStrategyTest.kt:42-53` — it currently pins the x100 behaviour ("5" -> `BOND:thorNode::500`) and codifies it in the test name; it must assert "5" -> `BOND:thorNode::5`.
- [ ] Update the `DepositFormViewModelTest` cases that pin the `1..100` range (`"150"` errors, `"5"` clears, `"0"` errors) to the new `0..10000` range.

### Must NOT Do
- Don't change unrelated code
- Don't change the Send form's Bond behaviour — it is the reference
- Don't alter `DepositMemo.Bond.Thor.toString()`; the memo format is correct, only the value fed into it is wrong

## Acceptance Criteria
- [ ] Entering `5` on the Deposit form's Bond operator-fee field produces the memo `BOND:<node>::5`, identical to the Send form for the same input
- [ ] Entering `10001` on the Deposit form blocks submission with a visible error; no transaction is staged and no keysign ceremony starts
- [ ] Entering `0` is accepted on both forms and the memo carries no fee segment
- [ ] The Deposit form's operator-fee label names basis points, and the string exists in all 10 locale files
- [ ] `./gradlew :app:testDebugUnitTest --tests "*BondStrategyTest*" --tests "*DepositFormViewModelTest*"` passes

johnnyluo:
## Summary
- The Deposit form's Bond operator-fee field scaled the entered value x100 before writing it into the memo, while the Send form's identical field is raw basis points — the same input produced two different on-chain fees depending on which screen the user reached Bond from.
- Drops the `movePointRight(2)` scaling in `BondStrategy` so both entry points pass the raw basis-points value through unchanged.
- Widens the Deposit operator-fee validator to the `0..10000` basis-points range (matching the Send form) and makes it blocking: `DepositFormViewModel.deposit()` now validates the field and refuses to build the transaction while it errors, so an out-of-range value can't reach a keysign ceremony.
- Relabels `deposit_form_operator_fee_title` to state the unit ("Operator fees (basis points)") in all 10 locale files.
- Aligns parsing so a non-numeric/decimal entry is rejected consistently instead of being flagged invalid but signed anyway.
- **Follow-up 1**: removes THORChain Bond from the Send form entirely, routing through the Deposit form instead (mirroring MayaChain, which already did this). Deletes the now-unreachable Send-form `BondStrategy`, its operator-fee UI, and related field plumbing.
- **Follow-up 2**: removes THORChain Unbond from the Send form the same way. Deletes the Send-form `UnbondStrategy`, the shared Bond/Unbond node-address UI (`FoldableBondDestinationAddress`), and the provider-address field plumbing (`AddressBookType.PROVIDER`, `setProviderAddress`/`scanProviderAddress`).
  - **Known gap accepted by explicit request**: unlike the Send form's `AccountsLoader.publishUnbond`, the Deposit form does not yet cap the Unbond amount to the specific node's bonded balance — it shows the combined RUNE balance across every bonded node instead. A follow-up should port that per-node ceiling into the Deposit flow.
  - `SendFormScreenTestTagsTest` guards `testTag`s an external QA Maestro flow depends on (issue #4180); its `SendFormScreen.bondAddressField` selector no longer exists, and the test was updated accordingly — but the external Maestro flow itself needs separate attention from QA since it's outside this repo.
- **Follow-up 3**: audited every other DeFi action still on the Send form against the Deposit form. `STAKE_CACAO`, `UNSTAKE_CACAO`, and `REMOVE_LP` are fully implemented in Deposit, never reached via Send-form navigation, and had no dedicated Send-form logic (they already fell through to the generic `DefaultSendStrategy`) — split out of the exhaustive `when`s into their own Deposit-only bucket.
  - `ADD_LP` deliberately stays wired to Send: Deposit's `AddLiquidityStrategy` only covers the RUNE/CACAO side of a pool add, while Send form's `DefaultSendStrategy` has a real, tested branch (`isThorchainRouterDeposit`) for the non-native EVM asset side, which Deposit doesn't cover.
  - RUJI/TCY/YBRUNE/sTCY staking and rewards withdrawal (`STAKE_RUJI`, `STAKE_SRUJI`, `STAKE_TCY`, `STAKE_YBRUNE`, `UNSTAKE_RUJI`, `UNSTAKE_SRUJI`, `UNSTAKE_TCY`, `UNSTAKE_STCY`, `UNSTAKE_YBRUNE`, `WITHDRAW_RUJI`) have no Deposit-form equivalent (Deposit's `Stake`/`Unstake` options are TON-only) — left untouched.
  - "Leave" was investigated but was never present on the Send form (it only exists as `DepositOption.Leave` in the Deposit form already), so there was nothing to remove there.

## Test plan

Fixes #5757

https://claude.ai/code/session_01NGkNoiaWwrgmojrCbHDKnH

## Summary by CodeRabbit

* **New Features**
  * Bonding and unbonding now use the Deposit flow.
  * Operator fees accept basis-point values from 0 to 10,000.
  * Bond submissions pass fees through as raw basis-point values.

* **Bug Fixes**
  * Blank operator-fee fields no longer retain stale validation errors.
  * Invalid fees now show a range-specific error.
  * Unsupported DeFi actions no longer submit as regular transfers.

* **UI & Localization**
  * Updated fee labels and input hints to clarify basis-point values.
  * Removed obsolete bond and unbond form labels, titles, and provider fields.


app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1524,66 +1524,39 @@ constructor(
...
-                        type = DeFiNavActions.BOND.type,
-                        chainId = Chain.ThorChain.id,
-                        tokenId = runeCoin.id,
-                        address = nodeAddress,
-                    )
+            navigator.route(
+                Route.Deposit(
+                    vaultId = vaultId,
+                    chainId = Chain.ThorChain.id,
+                    depositType = DeFiNavActions.BOND.type,
```
rkokhatskyi, line 1526: Bond on the deposit form never bounds the amount because `deposit/submit/BondStrategy.kt:117` converts it with no balance check, while `VerifyDepositViewModel.kt:241` checks only the fee and trusts the form to have sized it. The deleted send strategy blocked this with `getAvailableTokenBalance` and iOS and Windows both cap on the wallet balance

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1524,66 +1524,39 @@ constructor(
...
-                        chainId = Chain.ThorChain.id,
-                        tokenId = runeCoin.id,
-                        address = nodeAddress,
-                        bondedAmount = bondedAmount?.toString(),
-                    )
+            navigator.route(
+                Route.Deposit(
+                    vaultId = vaultId,
+                    chainId = Chain.ThorChain.id,
+                    depositType = DeFiNavActions.UNBOND.type,
```
rkokhatskyi, line 1539: Unbond now labels the amount with `state.balance`, the liquid wallet RUNE from `AccountsRepository.kt:304`, and enforces no ceiling. The deleted `publishUnbond` capped to this node's bond, as do iOS and Windows, so the figure shown is unrelated to what is actually unbondable

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
```diff
@@ -537,6 +537,15 @@ constructor(
         viewModelScope.launch {
             try {
                 val vaultId = vaultId ?: return@launch
+
+                if (
+                    state.value.depositOption == DepositOption.Bond &&
+                        operatorFeeFieldState.text.isNotEmpty()
+                ) {
+                    fieldInputCoordinator.validateOperatorFee()
+                    if (state.value.operatorFeeError != null) return@launch
```
rkokhatskyi, line 547: `setProvider` at `DepositFieldInputCoordinator.kt:196` skips the validation its sibling `setNodeAddress` runs, and no gate reads `providerError`, so a pasted provider reaches the memo via `deposit/submit/BondStrategy.kt:124` with no error shown. The deleted send strategy blocked it through `resolveName`

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1524,66 +1524,39 @@ constructor(
...
-            if (runeCoin != null) {
-                navigator.route(
-                    Route.Send(
-                        vaultId = vaultId,
-                        type = DeFiNavActions.BOND.type,
-                        chainId = Chain.ThorChain.id,
-                        tokenId = runeCoin.id,
-                        address = nodeAddress,
-                    )
+            navigator.route(
```
aminsato, line 1522: Routing Bond from the DeFi tab to the Deposit form drops the insufficient-balance guard the Send form applied. `send/submit/BondStrategy` rejected `tokenAmountInt > getAvailableTokenBalance(account, gasFee)` before building the transaction; `deposit/submit/BondStrategy.build()` has no balance check at all, `DepositAmountHelper.requireTokenAmount` is not called on this path, the Continue button only gates on node/dst address errors, and `VerifyDepositViewModel.checkFeeAffordability` only checks the gas fee, not the bond amount. Scenario: a vault holding 10 RUNE taps Bond on a node, enters 100, and gets all the way through a full keysign ceremony before THORChain rejects the transfer. Same change also removes the `RUNE coin not found in vault` precondition (its test `a vault without RUNE routes nowhere on bond` was deleted), so a vault without THORChain enabled now navigates into a Deposit form that can only fail with a generic "no address" error.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1524,66 +1524,39 @@ constructor(
...
-                navigator.route(
-                    Route.Send(
-                        vaultId = vaultId,
-                        type = DeFiNavActions.UNBOND.type,
-                        chainId = Chain.ThorChain.id,
-                        tokenId = runeCoin.id,
-                        address = nodeAddress,
-                        bondedAmount = bondedAmount?.toString(),
-                    )
+            navigator.route(
```
aminsato, line 1535: Unbond loses the per-node ceiling that `AccountsLoader.publishUnbond` provided (deleted in this PR) and gains nothing in its place: `deposit/submit/UnbondStrategy.build()` accepts any positive amount, and the Deposit form displays `balanceDecimal` from the vault's RUNE account. Scenario: a vault bonded to node A (500 RUNE) and node B (2000 RUNE) taps Unbond on node A — the form shows the *combined* bond as the balance and happily builds `UNBOND:thorA:250000000000` for 2500 RUNE, which the node refuses. iOS gates this with `AmountBalanceValidator` against the node's own bond. The PR body flags this as accepted, but it is a user-visible regression on the exact flow being re-routed here, so it should at least land behind a follow-up issue referenced from the code.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -652,7 +652,7 @@ constructor(
 
         // Cancel any in-flight collector before starting a new one. getActiveNodes never
         // completes, so without this each refresh would stack another collector that writes
-        // `activeBondedNodes`/state out of order — leaving onClickUnBond to resolve a stale bond.
+        // `activeBondedNodes`/state out of order.
```
aminsato: With `onClickUnBond` no longer reading it, `activeBondedNodes` (line 189) is now written and never read — the `getActiveNodes` collector at line 718-720 keeps the field alive purely as dead state, and this comment still explains a cancel that no longer protects any reader. Either drop the field and its write, or note what still depends on it.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/load/DepositFieldInputCoordinator.kt
```diff
@@ -138,7 +138,7 @@ constructor(
     fun validateOperatorFee() {
         val text = fields.operatorFeeFieldState.text.toString()
         if (text.isNotEmpty()) {
-            val errorText = fieldValidator.validateBasisPoints(text.toIntOrNull())
+            val errorText = fieldValidator.validateOperatorFee(text.toIntOrNull())
```
aminsato: The `text.isNotEmpty()` guard means a previously set `operatorFeeError` is never cleared when the field is emptied, and this now matters more than before because the error is rendered under the field (`DepositFormScreen.kt:387-395`). Scenario: on THORChain Bond the user types `50000`, blurs (inline error appears), then clears the field — the field is now blank and valid (`deposit()` lets it through, and the new test asserts exactly that), yet "Operator fee must be a valid number between 0 and 10000 basis points" stays on screen with nothing to fix. `resetTextFields()` in `DepositOptionCoordinator` also clears the field without clearing `operatorFeeError`, so switching Bond → Unbond → Bond leaves the same orphaned error. Clearing the error on empty input in the `else` branch fixes both.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormUiModel.kt
```diff
@@ -224,8 +210,7 @@ internal enum class SendFocusField {
 }
 
 enum class AddressBookType {
-    OUTPUT,
-    PROVIDER,
+    OUTPUT
```
aminsato: `AddressBookType` is now a single-constant enum with no remaining references anywhere in the app — `openAddressBook()` no longer takes a type and `PROVIDER` was the only reason the type existed. Dead type; delete it (and the now-unused `editable` parameter of `AddressInputSection` in `SendFormAddressSection.kt`, whose only `false` caller was the deleted `FoldableBondDestinationAddress`).

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
johnnyluo: Fixed in 0aaddc0: BondStrategy now calls `requireTokenAmount` (the same helper TransferIbcStrategy already uses) for THORChain bonds, so an amount above the wallet's RUNE balance is rejected before keysign.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
johnnyluo: Acknowledged as a known gap (also called out in the PR description). Added a code comment on `UnbondStrategy` in 0aaddc0 documenting the missing per-node ceiling so it doesn't rot silently; porting the actual per-node cap is left for a follow-up.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
johnnyluo: Fixed in 0aaddc0: `setProvider` now revalidates on input, and `deposit()` gates on `providerError` for Bond/Unbond the same way it already does for the operator-fee field.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
johnnyluo: Fixed in 0aaddc0: BondStrategy now routes the THORChain amount through `requireTokenAmount`, which checks the wallet's RUNE balance (plus gas) and throws before the transaction is built, closing this path.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
johnnyluo: Agreed this needs to be referenced from the code, not just the PR body — added a doc comment on `UnbondStrategy` in 0aaddc0 pointing at the missing per-node ceiling and the follow-up needed. The actual per-node cap is left for that follow-up rather than this PR.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
johnnyluo: Fixed in 0aaddc0: dropped `activeBondedNodes` (field, write, and the stale comment referencing it).

app/src/main/java/com/vultisig/wallet/ui/models/deposit/load/DepositFieldInputCoordinator.kt
johnnyluo: Fixed in 0aaddc0: `validateOperatorFee()` now clears the error on blank input instead of no-op'ing, and `resetTextFields()` also clears `operatorFeeError`/`providerError` so switching Bond → Unbond → Bond doesn't leave an orphaned error either.

app/src/main/java/com/vultisig/wallet/ui/models/send/SendFormUiModel.kt
johnnyluo: Fixed in 0aaddc0: deleted `AddressBookType` and the always-true `editable` parameter of `AddressInputSection` (its only `false` caller was the deleted `FoldableBondDestinationAddress`).

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
rkokhatskyi: Verified in 0aaddc0. `requireTokenAmount` now runs on the ThorChain branch and throws when `nativeTokenValue < tokenAmountInt + gas.value`, the same bound the deleted send strategy enforced. Closing this one.

app/src/main/java/com/vultisig/wallet/ui/models/deposit/DepositFormViewModel.kt
rkokhatskyi: Verified in 0aaddc0. `deposit()` calls `validateProvider()` itself before reading `providerError`, so the gate covers every writer of the field including the paste path, not only the ones that revalidate. Closing this one.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
rkokhatskyi: Fair enough, your call. Correcting the record so the follow-up is scoped right: the form shows the liquid bank balance from `/cosmos/bank/v1beta1/balances`, not the combined bond, and there is no cap of any kind since `DepositStrategyFactory` never passes `requireTokenAmount` to `UnbondStrategy`. Funds are safe regardless, thornode `refundBond` clamps to the provider's own bond and refunds to their own address, so the worst case is an unintended full unbond recoverable for the gas fee.

rkokhatskyi (approved):
LGTM

PR #5769, opened 2026-08-31
Fix THORChain empty Midgard status fallback

### LINKED ISSUE #5758: [Fix] THORChain/Maya status trusts the native node on the first empty Midgard response, so a to-be-refunded tx can latch as Confirmed [Transaction Status]
# [Fix] THORChain/Maya status trusts the native node on the first empty Midgard response, so a to-be-refunded tx can latch as Confirmed [Transaction Status]

## Problem

`ThorMayaChainStatusProvider.checkStatus()` falls back to the native node the instant the *first* Midgard poll returns an empty actions array:

```
69:  val action = response.actions.firstOrNull() ?: return checkNativeStatus(txHash, chain)
```

That single elvis is the whole mechanism. The provider is stateless, so there is no attempt counter, no elapsed-time gate and no requirement that the array be empty more than once. `checkNativeStatus()` then maps a raw `tx_response.code == 0` straight to `Confirmed` (lines 89-90).

The problem is that "Midgard has no action" is ambiguous, and the fallback resolves the ambiguity the wrong way. It can mean either:
- **no action will ever exist** — a plain native transfer, or a deposit the node rejected during execution. This is the case the fallback was built for, and for it `code == 0` is a correct answer; or
- **no action exists *yet*** — Midgard has to observe the block and then index and derive the action, which structurally lags the raw transaction landing.

For a swap or an LP add, the second case is exactly what a poll sees in the window right after inclusion. During that window the raw transaction has `code == 0` — it was accepted and committed — even when THORChain is about to refund it (paused pool, slip limit exceeded). So the fallback reports `Confirmed` for a transaction whose real outcome is a refund.

That verdict is unrecoverable. `Confirmed` is terminal at every layer:
- `PollingTxStatusUseCase.kt:50-53` does `return@flow` on `Confirmed`, exiting the flow builder outright.
- `KeysignTxStatusPoller.kt:132` collects with `first { it.isTerminal() }`, and `isTerminal()` includes `Confirmed` (`:216-221`), so collection is cancelled.
- `TransactionStatusService.kt:101-115` stops polling and tears down the foreground service.
- The background refresh cannot correct it either: `TransactionDao.kt:69-71` selects only rows whose status is in `('BROADCASTED', 'PENDING', 'NotFound')`, so a `CONFIRMED` row is excluded by construction. Every write is separately guarded with `AND status NOT IN ('CONFIRMED','FAILED','REFUNDED')` (`:97`, `:107`, `:117`, `:131`, `:151`), and the one query that could bypass that guard, `demoteForReorg` (`:136-145`), has no production caller.

So once a refunded swap is written as `Confirmed` it stays `Confirmed` forever, and the user is told a transaction succeeded when the funds are being returned. This is the same failure class that #4468 previously closed, reopened through a different route. The mis-classification is also most likely under congestion — precisely the conditions that make refunds more common.

The fallback itself was added deliberately in #5002 to unstick plain transfers and node-rejected deposits that Midgard never indexes; that PR's own description notes Midgard only indexes recognised actions (swap, addLiquidity, withdraw, savers, loan, bond) and states the native query is "a fallback only used when no action exists". The gap is that the code cannot distinguish "no action exists" from "no action exists yet", and applies the fallback unconditionally to memo types Midgard *does* eventually index.

**Adjacent defect on the same path, to fix in the same change:** `PollingTxStatusUseCase.kt:51-52` lists only `Confirmed` and `Failed` in its terminal arm and omits `Refunded`, even though both consumers treat `Refunded` as terminal (`TransactionStatusService.kt:103`, `KeysignTxStatusPoller.kt:220`). The use case therefore keeps polling after a refund and stops only because its collectors cancel it. THORChain/Maya is the chain that actually emits `Refunded` (`ThorMayaChainStatusProvider.kt:99`, `:106`), so this asymmetry sits directly on the path being changed.

## Expected Behavior

A swap or LP transaction that Midgard will eventually index is not declared `Confirmed` on the strength of a single empty actions response. The native-node fallback is trusted immediately only for memo types Midgard structurally never tracks; for everything else it is trusted only once the actions array has come back empty repeatedly, so ordinary indexing lag resolves through Midgard and a refund is reported as a refund.

## Steps to Reproduce

1. Broadcast a THORChain swap into a pool that will refund it (a paused pool, or a slip limit the trade exceeds).
2. Let the status poll run immediately after the transaction is included in a block, while Midgard has not yet indexed the resulting action — `GET /v2/actions?txid=...` returns `{"actions": [], "count": "0"}`.
3. The provider falls back to the native node, which reports the committed transaction with `code == 0`, and the status becomes `Confirmed`.
4. Polling stops and the history row is written as `CONFIRMED`.
5. Once Midgard indexes the action it shows a refund, but the app never re-checks the row and continues to display the transaction as successful.

## Solution

Distinguish the two meanings of an empty actions array instead of collapsing them. Classify the outgoing transaction by its memo: for memo types Midgard is known to index (swap, add/withdraw liquidity, savers, loan, bond) require the actions array to come back empty across at least two consecutive polls before trusting the native fallback, and for memo types Midgard structurally never tracks (plain transfers, and deposits the node rejects) keep trusting it on the first response as #5002 intended. This preserves the stuck-in-progress fix while closing the window in which indexing lag is mistaken for a final answer.

### Must Do
- [ ] Make the empty-actions fallback at `ThorMayaChainStatusProvider.kt:69` conditional rather than unconditional: trust the native node immediately only for memo types Midgard does not index, and otherwise require consecutive empty Midgard responses first.
- [ ] Carry whatever state that requires (a consecutive-empty count or equivalent) explicitly, since the provider is currently stateless and each `checkStatus` call is independent.
- [ ] Keep the behaviour #5002 added for the cases it targeted: a plain native transfer and a node-rejected deposit must still resolve without waiting for Midgard.
- [ ] Add `Refunded` to the terminal arm of `PollingTxStatusUseCase.kt:51-52` so the use case stops on its own rather than relying on collector cancellation.
- [ ] Extend `ThorMayaChainStatusProviderTest` with a case that pins the new behaviour: an indexable memo type plus one empty Midgard response plus native `code == 0` must NOT yield `Confirmed`, and the same input on a non-indexable memo type must still yield `Confirmed`.

### Must NOT Do
- Don't change unrelated code
- Don't remove the native-node fallback — it fixes a real stuck-status bug
- Don't change the Midgard action mapping (refund / failed / success / pending) — it is correct
- Don't loosen the terminal-status write guards in `TransactionDao`

## Acceptance Criteria
- [ ] A swap whose Midgard actions array is empty on the first poll and whose native result is `code == 0` is not reported `Confirmed`; polling continues and the eventual Midgard refund action is reported as `Refunded`
- [ ] A plain native transfer with no Midgard action and native `code == 0` still resolves to `Confirmed` without waiting for Midgard
- [ ] A deposit the node rejects still resolves to `Failed` with the node's `raw_log`, unchanged
- [ ] A native `404` still maps to `Pending` and keeps polling, unchanged
- [ ] `PollingTxStatusUseCase` stops emitting after `Refunded` without depending on the collector to cancel it
- [ ] `./gradlew :data:testDebugUnitTest --tests "*ThorMayaChainStatusProviderTest*"` passes, including the existing 18 cases

johnnyluo:
## Summary
- Gate native-node success fallback for THORChain/Maya indexable memos until Midgard returns repeated empty action responses
- Decode native tx body memo so empty-action fallback can distinguish indexed memo types from plain/non-indexed transfers
- Treat Refunded as terminal in PollingTxStatusUseCase

## Tests
- ./gradlew :data:testDebugUnitTest --tests "*ThorMayaChainStatusProviderTest*"
- ./gradlew :data:testDebugUnitTest --tests "*PollingTxStatusUseCaseTest*"
- ./gradlew assembleDebug

Fixes #5758

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction status handling for confirmed, failed, pending, and refunded transactions across THORChain and Maya.
  - Preserved transaction memos during status checks.
  - Prevented premature confirmation while transaction details are still being indexed.
  - Polling now stops correctly when a transaction is refunded.
  - Improved recognition of deposit-related transaction memos and delayed status responses.

- **Tests**
  - Added coverage for delayed confirmations, native success handling, memo processing, and refunded transaction polling.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
@@ -1,6 +1,7 @@
 package com.vultisig.wallet.data.api.txstatus
 
 import com.vultisig.wallet.data.api.models.cosmos.CosmosEnvelopedTxResponse
+import com.vultisig.wallet.data.api.models.cosmos.CosmosTxBody
 import com.vultisig.wallet.data.models.Chain
 import com.vultisig.wallet.data.usecases.txstatus.TransactionResult
 import com.vultisig.wallet.data.usecases.txstatus.TransactionStatusProvider
@@ -9,6 +10,7 @@ import io.ktor.client.HttpClient
 import io.ktor.client.request.get
 import io.ktor.client.request.parameter
 import io.ktor.http.HttpStatusCode
+import java.util.concurrent.ConcurrentHashMap
 import javax.inject.Inject
 import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.serialization.SerialName
@@ -41,13 +43,20 @@ import timber.log.Timber
  * plain native transfer (e.g. sending a secured asset) or a native deposit the node rejected during
  * execution never becomes an action, so `/v2/actions` returns an empty array for it. Treating that
  * as [TransactionResult.Pending] left such txs stuck "in progress" forever while the app kept
- * polling Midgard. When no action is found we therefore fall back to the native node's
+ * polling Midgard. When no action is found we therefore consult the native node's
  * `cosmos/tx/v1beta1/txs/{hash}` endpoint, which reports the committed tx's result code directly:
- * code 0 → [TransactionResult.Confirmed], non-zero → [TransactionResult.Failed] with the node's
- * `raw_log`, and 404/not-yet-committed → [TransactionResult.Pending] so polling continues.
+ * non-zero → [TransactionResult.Failed] with the node's `raw_log`, and 404/not-yet-committed →
+ * [TransactionResult.Pending] so polling continues. A successful native result is trusted
+ * immediately for memos Midgard does not index; for Midgard-indexed memos it is trusted only after
+ * repeated empty action responses, giving Midgard's action index time to surface refunds.
  */
-class ThorMayaChainStatusProvider @Inject constructor(private val httpClient: HttpClient) :
-    TransactionStatusProvider {
+class ThorMayaChainStatusProvider
+internal constructor(
+    private val httpClient: HttpClient,
+    private val nowMillis: () -> Long,
+) : TransactionStatusProvider {
+
+    @Inject constructor(httpClient: HttpClient) : this(httpClient, System::currentTimeMillis)
 
     private val midgardUrls =
         mapOf(
@@ -61,12 +70,17 @@ class ThorMayaChainStatusProvider @Inject constructor(private val httpClient: Ht
             Chain.MayaChain to MAYACHAIN_NATIVE_TX_URL,
         )
 
+    private val emptyActionStreaks = ConcurrentHashMap<EmptyActionKey, EmptyActionStreak>()
+
     override suspend fun checkStatus(txHash: String, chain: Chain): TransactionResult {
         val baseUrl = midgardUrls[chain] ?: return TransactionResult.Failed("Unknown chain")
         return try {
             val response: MidgardActionsResponse =
                 httpClient.get(baseUrl) { parameter(MIDGARD_TXID_PARAM, txHash) }.bodyOrThrow()
-            val action = response.actions.firstOrNull() ?: return checkNativeStatus(txHash, chain)
+            val action =
+                response.actions.firstOrNull()
+                    ?: return checkNativeStatusAfterEmptyAction(txHash, chain)
+            emptyActionStreaks.remove(EmptyActionKey(chain, txHash))
             mapAction(action)
         } catch (e: CancellationException) {
             throw e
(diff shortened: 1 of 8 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
```diff
@@ -77,20 +85,56 @@ class ThorMayaChainStatusProvider @Inject constructor(private val httpClient: Ht
...
+            return TransactionResult.Confirmed
+        }
+
+        val streak = (emptyActionStreaks[key] ?: 0) + 1
+        emptyActionStreaks[key] = streak
+        return if (streak >= MIN_INDEXABLE_EMPTY_ACTION_POLLS) {
+            emptyActionStreaks.remove(key)
+            TransactionResult.Confirmed
+        } else {
+            TransactionResult.Pending
```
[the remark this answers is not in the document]
johnnyluo: Addressed in 4acada466: empty Midgard responses are now counted before native status lookup, so a native 404 contributes to the streak before a later indexed native success is gated. Added a regression test for the 404-then-success case.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
```diff
@@ -77,20 +85,60 @@ class ThorMayaChainStatusProvider @Inject constructor(private val httpClient: Ht
...
+        val envelope = response.bodyOrThrow<CosmosEnvelopedTxResponse>()
+        val txResponse = envelope.txResponse
+        val result =
+            when (txResponse.code) {
+                0 -> TransactionResult.Confirmed
+                null -> TransactionResult.Pending
+                else ->
+                    TransactionResult.Failed(nonBlankOr(txResponse.rawLog, DEFAULT_FAILED_REASON))
+            }
+        return NativeStatusResult(result, envelope.tx?.body?.memo)
```
rkokhatskyi: THORChain swaps never engage this gate because `ThorchainSwapHelper.kt:86` never sets `SigningInput.memo`, so the swap memo lives only in MsgDeposit while this line reads `tx.body.memo`. Live Vultisig swap `2F5B908F…35E8` carries an empty body memo, so a refundable swap still confirms on poll one

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
```diff
@@ -77,20 +85,60 @@ class ThorMayaChainStatusProvider @Inject constructor(private val httpClient: Ht
...
+     * stays pending, and native success is gated for memo types Midgard is expected to index.
      */
-    private suspend fun checkNativeStatus(txHash: String, chain: Chain): TransactionResult {
-        val nativeTxUrl = nativeTxUrls[chain] ?: return TransactionResult.Pending
+    private suspend fun checkNativeStatusAfterEmptyAction(
+        txHash: String,
+        chain: Chain,
+    ): TransactionResult {
+        val key = EmptyActionKey(chain, txHash)
+        val streak = (emptyActionStreaks[key] ?: 0) + 1
```
rkokhatskyi: The streak increments before `checkNativeStatus` and the 404 branch returns Pending without clearing the key, so pre-commit polls after a SYNC broadcast consume both counts. The first poll that ever sees `code == 0` already sits at streak 2 and confirms, which `ThorMayaChainStatusProviderTest.kt:382` pins

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
```diff
@@ -77,20 +85,60 @@ class ThorMayaChainStatusProvider @Inject constructor(private val httpClient: Ht
...
+            }
+            return nativeStatus.result
+        }
+
+        if (!nativeStatus.memo.isMidgardIndexedMemo()) {
+            emptyActionStreaks.remove(key)
+            return TransactionResult.Confirmed
+        }
+
+        return if (streak >= MIN_INDEXABLE_EMPTY_ACTION_POLLS) {
```
rkokhatskyi: `streak` counts calls rather than elapsed time on a singleton map with no dedupe, so the 15s in-flight sweep and the backoff-exempt `refreshOne` at `RefreshPendingTransactionsUseCase.kt:61` can take counts 1 and 2 milliseconds apart, giving Midgard no indexing time before Confirmed latches

rkokhatskyi (changes requested):

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
johnnyluo: Addressed in 38e4b619e: gated native success now requires both repeated committed empty-action observations and at least 15 seconds elapsed from the first committed observation, preventing refreshOne plus another poll from confirming milliseconds apart.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
```diff
@@ -77,20 +85,60 @@ class ThorMayaChainStatusProvider @Inject constructor(private val httpClient: Ht
...
+
+        val nativeStatus = checkNativeStatus(txHash, chain)
+        if (nativeStatus.result != TransactionResult.Confirmed) {
+            if (nativeStatus.result is TransactionResult.Failed) {
+                emptyActionStreaks.remove(key)
+            }
+            return nativeStatus.result
+        }
+
+        if (!nativeStatus.memo.isMidgardIndexedMemo()) {
```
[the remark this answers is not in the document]
johnnyluo: Addressed in 38e4b619e: a missing native memo is no longer treated as non-indexed. Native success with no available memo now goes through the guarded empty-action path instead of confirming on the first poll.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
johnnyluo: Addressed in 38e4b619e: native 404 no longer increments the gated-success streak. The streak starts only once the native tx has committed with code 0, so pre-commit SYNC polls cannot consume both counts.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
johnnyluo: Addressed in 38e4b619e: the native response model now decodes tx.body.messages and the provider prefers /types.MsgDeposit.memo over tx.body.memo. This covers THORChain swaps whose body memo is empty but whose actionable memo lives in MsgDeposit.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
```diff
@@ -61,12 +70,17 @@ class ThorMayaChainStatusProvider @Inject constructor(private val httpClient: Ht
             Chain.MayaChain to MAYACHAIN_NATIVE_TX_URL,
         )
 
+    private val emptyActionStreaks = ConcurrentHashMap<EmptyActionKey, EmptyActionStreak>()
```
aminsato, line 71: `emptyActionStreaks` lives in a `@Singleton` provider (`TxStatusModule.bindThorChainStatusProvider`) and entries are only removed on a terminal outcome: a Midgard action appears, native returns `Failed`, or the gate fires. Any tx that stops being polled while still gated keeps its entry for the process lifetime — the 5-minute `maxWaitSeconds` timeout in `PollingTxStatusUseCase`, the flow being cancelled when the user leaves the screen, or the one-shot checks from `RefreshPendingTransactionsUseCase`.

That has two effects. The map grows without bound, and a stale entry defeats the gate it was added for: on the next polling session for the same hash, `count` is already >= 2 and `lastObservedAtMillis - firstObservedAtMillis` spans the whole gap since the last session, so the *first* empty-action poll returns `Confirmed` immediately instead of giving Midgard the intended 15s window.

Aging entries out inside `compute` — discard `current` when `now - current.lastObservedAtMillis` exceeds some staleness bound and start a fresh streak — fixes both, since the map then self-prunes on the next touch of a key and never credits a streak across an interruption.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
```diff
@@ -77,20 +91,98 @@ class ThorMayaChainStatusProvider @Inject constructor(private val httpClient: Ht
...
+        val streak =
+            emptyActionStreaks.compute(key) { _, current ->
+                val now = nowMillis()
+                current?.copy(count = current.count + 1, lastObservedAtMillis = now)
+                    ?: EmptyActionStreak(
+                        count = 1,
+                        firstObservedAtMillis = now,
+                        lastObservedAtMillis = now,
+                    )
+            } ?: EmptyActionStreak(
```
aminsato: `ConcurrentHashMap.compute` returns null only when the remapping function returns null, and this one always returns a value, so the `?: EmptyActionStreak(count = 0, ...)` fallback is unreachable. It also reads as if `count = 0` were a real state the gate has to handle, which it isn't. Worth dropping in favour of a non-null read of the compute result.

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
johnnyluo: Fixed in 6ede59037: `compute()` now discards a streak entry once it's older than `EMPTY_ACTION_STREAK_TTL_MS` (5 min) before incrementing, so a resumed polling session starts a fresh streak instead of inheriting a stale one. Added a regression test (`stale empty action streak does not carry over to a later polling session`).

data/src/main/kotlin/com/vultisig/wallet/data/api/txstatus/ThorMayaChainStatusProvider.kt
johnnyluo: Fixed in 6ede59037: dropped the unreachable `?: EmptyActionStreak(count = 0, ...)` fallback and used `!!` on the compute result instead, since the remapping lambda always returns a non-null value.

PR #5777, opened 2026-08-31
feat(passcode): offer biometric unlock from the lock screen

### LINKED ISSUE #5762: [Update] App lock: add "Use biometrics" to the passcode lock screen, never auto-prompt [Passcode]
**Design:** https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=79835-127499

Lock screen = 6-dot passcode field, keypad, and a "Use biometrics" link that launches `BiometricPrompt`. No automatic prompt.

**Today:** `PasscodeGuard.kt:71` routes `Locked` to `PasscodeLockScreen`, which has no biometric affordance — a passcode user cannot use their fingerprint at all. `BiometryAuthScreen` (`BiometryAuthPrompt.kt:49-61`) auto-launches, but only on the passcode-disabled path.

**Change:** add the link to `PasscodeLockScreen` when the DEK has a biometric wrap (`PasscodeCipher.kt:22`); tapping it runs `launchBiometricPrompt` with `BIOMETRIC_STRONG` (#5755) and unlocks via the wrapped DEK. Never launch on composition.

**Why:** parity with vultisig/vultisig-ios#5291; passcode users currently lose biometrics entirely.

Screenshot of the design attached below.

aminsato:
Closes #5762

## Summary

A user who sets a passcode currently loses biometrics entirely. `PasscodeGuard.kt:71` routes `Locked` to `PasscodeLockScreen`, which has no biometric affordance, and `BiometryAuthScreen` only ever runs on the passcode-*disabled* path.

**The issue understates the work.** It points at `PasscodeCipher.kt:22` for the biometric wrap, but that line is a doc sentence saying the DEK *can* be wrapped that way — nothing implemented it. `PasscodeCredentials` holds only `salt` + `wrappedDataKey`. So this PR builds the path rather than adding a link to something that isn't there.

- **`BiometricUnlockStore`** (new, `:data`) — a second copy of the data key, encrypted under an AndroidKeyStore AES-GCM key the hardware releases only on a strong biometric match. `setInvalidatedByBiometricEnrollment(true)`, per-use auth (`setUserAuthenticationParameters(0, AUTH_BIOMETRIC_STRONG)` on R+, `validityDuration = -1` below). Everything fails closed to the passcode — no hardware, nothing enrolled, changed enrolment, lockout and a missing blob all resolve to null.
- **`PasscodeRepository`** — `isBiometricUnlockEnabled`, enable/disable, and `unlockWithBiometrics`. `disablePasscode` clears the copy *first*; `setPasscode` clears a stray copy before minting a new key.
- **Lock screen** — a "Use biometrics" link, shown only when a copy exists and the device can still use it.
- **Settings** — an "Unlock with biometrics" switch. Without one no wrap could ever exist and the link would have nothing to offer.
- 8 new strings across all 10 locales.

## Two deliberate divergences

**No binding blob.** iOS binds its copy to `SHA256(wrappedDataKey)` because Keychain items survive uninstall. Android is `allowBackup="false"` and both halves — the preference and the keystore key — go with the app, so no copy can outlive its data key. The invariant is enforced structurally instead: the copy is cleared wherever the key is retired. This is what lets `changePasscode` keep working with **no second prompt**, where iOS has to rebind.

**`BIOMETRIC_STRONG` only, no `DEVICE_CREDENTIAL`.** The existing `BiometryAuthPrompt` allows the device PIN. Accepting it here would hand back exactly what the app passcode exists to prevent, so the prompt and the keystore key both refuse it.

## Before / After

The link is gated on a stored copy, so the "before" render is the same composable with the shortcut off — i.e. what every user sees today, and what a user who leaves the switch off keeps seeing.

| Before — no biometric affordance | After — "Use biometrics" |
|--------|-------|
| before | after |

New settings row:

settings

Real Android renders from the emulator via `PreviewActivity`, not Figma exports. Matched against the Figma frame ([node 79835:127499](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=79835-127499)): 20dp icon + label at 8dp gap, `#5CA7FF` (`Theme.v2.colors.alerts.info`), Brockmann Medium 14/18, 36dp under the card.

**The prompt itself is not pictured.** Pressing the link shows the system `BiometricPrompt` sheet, titled "Unlock Vultisig" with a "Use passcode" button — that window is `FLAG_SECURE`, so `adb screencap` renders it black. It is drawn by Android, not by us.

## What happens on each outcome

| Outcome | Behaviour |
|---|---|
| Match | The data key is recovered, the resume sweep runs, the gate comes down. A standing wrong-passcode penalty is cleared — a match is not a guess. |
| Cancel / dismiss / hardware lockout | Silent. The user caused it and the passcode field is already in front of them. |
| Copy gone (new finger enrolled) | The copy is dropped, the link disappears, and the screen says so rather than failing quietly. |

## Test plan

- `:data` 3327 tests, `:app` 2263 tests — 0 failures. 21 new tests; XML reports checked so JUnit 5 is not silently skipping any.
- Repository: enable/disable, unlock parity with the passcode path, lockout clearing, the interrupted-sweep resume, refusal while locked, a retired copy, and an enable the user abandons.
- View models: the link only appears with a copy, **biometrics never fire on their own**, the authorised cipher (not the offered one) is what unlocks, a cancel says nothing, a second tap starts nothing.
- `./gradlew :app:assembleDebug` and `:app:lintDebug` pass. ktfmt applied.

### Not covered by tests

The keystore path is exercised against a fake — a key only the hardware releases cannot exist in a JVM test. **Worth one manual pass on a physical device with a fingerprint enrolled** before merge: turn the switch on, lock, unlock with a match; then enrol a new finger and confirm the shortcut turns itself off and the passcode still works.

## Summary by CodeRabbit

* **New Features**
  * Added biometric unlock for passcode-protected wallets.
  * Added biometric enable/disable controls in passcode settings.
  * Added availability checks, cancellation handling, and passcode fallback messaging.
  * Preserved biometric access when changing the passcode and cleared it when appropriate.
  * Added localized biometric messaging across supported languages.

* **Bug Fixes**
  * Added safeguards for unavailable, unenrolled, changed, or invalid biometric credentials.

* **Tests**
  * Added coverage for biometric unlocking, configuration, failures, cancellation, and persistence scenarios.


app/src/main/java/com/vultisig/wallet/ui/screens/passcode/PasscodeSettingsScreen.kt
```diff
@@ -25,28 +30,36 @@ import com.vultisig.wallet.ui.models.settings.SettingsItemUiModel
...
 import com.vultisig.wallet.ui.utils.UiText
+import com.vultisig.wallet.ui.utils.asString
 
 @Composable
 internal fun PasscodeSettingsScreen(navController: NavHostController) {
     val viewModel = hiltViewModel<PasscodeSettingsViewModel>()
     val state by viewModel.state.collectAsStateWithLifecycle()
+    val context = LocalContext.current
+    val launcher = rememberBiometricUnlockLauncher()
+    val biometricAvailability = remember(context) { context.biometricUnlockAvailability() }
```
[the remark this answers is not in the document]
aminsato: Fixed in b47033a58. Real — the note tells the user to enrol in device settings, and `remember(context)` meant they came back to the same dead row.

Used `LifecycleEventEffect(Lifecycle.Event.ON_RESUME)` rather than collecting `eventFlow` by hand; it is the idiom already in `ThorchainDefiPositionsScreen` and `CosmosStakingPositionsScreen`, and it replays ON_RESUME on a lifecycle that is already resumed, so the initial `mutableStateOf` value only covers the first frame before the effect runs.

`PasscodeLockScreen` asked the same question the same way, so it got the same fix. The case that bites there is a sensor lockout expiring, or biometrics being switched back on, while the lock screen is still the one on top.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
```diff
@@ -0,0 +1,257 @@
...
+            // leaves behind. Clearing keeps the shortcut from advertising itself as available.
+            Timber.w(e, "The biometric unlock key is unusable; removing it")
+            clear()
+            null
+        } catch (e: GeneralSecurityException) {
+            Timber.w(e, "Could not read the biometric unlock key")
+            null
+        }
+
+    private fun keyStore(): KeyStore = KeyStore.getInstance(ANDROID_KEY_STORE).apply { load(null) }
```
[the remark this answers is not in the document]
aminsato: Fixed in b47033a58. Real, and the consequence is the bad part: `IOException` is not a `GeneralSecurityException` and Kotlin does not check it, so it escaped `clear()` and `isEnabled()` into `setPasscode` and `disablePasscode`, where `withContext(dispatcher) { biometrics.clear() }` catches nothing. Setting a passcode at all could fail over a shortcut that is optional — the opposite of the fail-closed contract at the top of the file.

Handled it in a single `keyStoreOrNull()` instead of adding a catch at each call site. Both callers already deal in absence (`?.deleteEntry`, `?.getKey`), and opening the keystore fails for reasons neither of them can tell apart, so one place to refuse reads better than two. `KeyStoreException` from `getInstance` moves in there too, which changes only the log line.

app/src/main/java/com/vultisig/wallet/ui/components/BiometricUnlockPrompt.kt
```diff
@@ -0,0 +1,143 @@
...
+                        }
+
+                        // onAuthenticationFailed is a finger that did not match. The prompt stays
+                        // up for another try, so there is nothing to resume with yet.
+                    },
+                )
+
+            continuation.invokeOnCancellation { prompt.cancelAuthentication() }
+
+            prompt.authenticate(
```
rkokhatskyi, line 145: `prompt.authenticate` runs only after a keystore round trip, so an onStop in that window makes `isStateSaved()` true and androidx biometric 1.1.0 returns with no callback, latching the `isVerifying` set at `PasscodeLockViewModel.kt:150` and leaving the passcode field dead until the process is killed

app/src/main/java/com/vultisig/wallet/ui/screens/passcode/PasscodeSettingsScreen.kt
```diff
@@ -100,12 +125,44 @@ internal fun PasscodeSettingsScreen(
...
+                                        ),
+                                    leadingIcon = R.drawable.ic_biometrics,
+                                    trailingSwitch = state.isBiometricUnlockEnabled,
+                                ),
+                            // The row stays visible on a device that cannot do this, with the note
+                            // below saying why — a capability that silently is not there reads as
+                            // one the app forgot to build.
+                            onClick = {
+                                if (
+                                    biometricAvailability == BiometricUnlockAvailability.Available
```
rkokhatskyi: The availability gate wraps both directions, yet turning the shortcut off needs no hardware at `PasscodeSettingsViewModel.kt:93`, so once every enrolled biometric is deleted the row keeps reading ON and every tap is swallowed, leaving the stored data key copy unremovable

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
```diff
@@ -0,0 +1,276 @@
...
+            null
+        }
+    }
+
+    override fun store(dataKey: ByteArray, cipher: Cipher): Boolean =
+        try {
+            // The IV is the cipher's own, generated when it was initialised for encryption, and it
+            // is not secret — it is stored beside the ciphertext exactly as PasscodeCipher does.
+            val blob = cipher.iv + cipher.doFinal(dataKey)
+            prefs.edit(commit = true) { putString(KEY_WRAPPED_KEY, encode(blob)) }
```
rkokhatskyi: `store` returns true unconditionally because androidx core `edit(commit = true)` discards the Boolean `commit()` returns, so a refused write reports success and the switch shows ON with nothing persisted, while `PasscodeStore.kt:102` checks that result for keys in the same file

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
```diff
@@ -0,0 +1,276 @@
...
+                    KEY_ALIAS,
+                    KeyProperties.PURPOSE_ENCRYPT or KeyProperties.PURPOSE_DECRYPT,
+                )
+                .setBlockModes(KeyProperties.BLOCK_MODE_GCM)
+                .setEncryptionPaddings(KeyProperties.ENCRYPTION_PADDING_NONE)
+                .setKeySize(PasscodeCipher.DATA_KEY_LENGTH * Byte.SIZE_BITS)
+                // Refuses a caller-supplied IV, so the same key can never encrypt twice under one.
+                .setRandomizedEncryptionRequired(true)
+                .setUserAuthenticationRequired(true)
+                // Destroys the key when a biometric is added or removed, so a newly enrolled
```
rkokhatskyi: This comment should say what the platform contract says, because `setInvalidatedByBiometricEnrollment` invalidates only on a new enrolment or once all biometrics are deleted, so added or removed promises a revocation Android never performs when one of two enrolled fingers is deleted

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
```diff
@@ -0,0 +1,276 @@
...
+                // fingerprint cannot open a wallet it was never shown to.
+                .setInvalidatedByBiometricEnrollment(true)
+                .apply {
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
+                        // Per use, and only a strong biometric: a timeout of zero means every
+                        // operation carries its own authentication, and the device credential is
+                        // deliberately not accepted — someone who knows the device PIN must not
+                        // reach a wallet the app passcode exists to protect.
+                        setUserAuthenticationParameters(0, KeyProperties.AUTH_BIOMETRIC_STRONG)
```
rkokhatskyi: The below R branch is unreachable because `gradle/libs.versions.toml:6` sets `minSdk = 30`, so `setUserAuthenticationParameters` always runs and the deprecated arm plus its comment describe an authentication policy no installable device can build

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/components/BiometricUnlockPrompt.kt
aminsato: Fixed in cf9eb058b. Real, and confirmed against the 1.1.0 source: `authenticateInternal` logs `Unable to start authentication. Called after onSaveInstanceState().` and returns at `BiometricPrompt.java:986` without touching the callback. Nothing cancels the coroutine either — `reset()` only fires on a transition *to* `Locked`, and the state is already `Locked` — so `isVerifying` stays set and `isInputEnabled` stays false until the process dies.

Guarded on `supportFragmentManager.isStateSaved` immediately before `authenticate` rather than reaching for a timeout: the check and the call both run on the main thread, which is also the thread that saves the state, so nothing can slip between them. A refusal now returns null, which the view model already treats as a dismissal, and the `finally` clears the flag.

app/src/main/java/com/vultisig/wallet/ui/screens/passcode/PasscodeSettingsScreen.kt
aminsato: Fixed in cf9eb058b. Real. Deleting the last enrolled biometric invalidates the keystore key rather than removing it — `getKey` still hands it back, invalidation only surfaces at `Cipher.init` — so `isEnabled()` reads true while availability reads `NotEnrolled`.

The part that makes it unrecoverable rather than merely stuck: the lock screen's self-heal path (`decryptCipherOrNull` → `KeyPermanentlyInvalidatedException` → `clear()`) is gated on the same availability at `PasscodeLockScreen.kt:158`, so the link is not offered either and nothing anywhere could drop the copy.

The gate now covers only the enable direction. Four instrumented cases in `PasscodeSettingsBiometricRowTest` pin both directions across all three availabilities.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
aminsato: Fixed in cf9eb058b. Real — androidx's `edit` calls `editor.commit()` and returns `Unit`, so the answer went on the floor.

The consequence is the bad half: `commit()` updates the in-memory map before it writes the file and leaves it updated when that write fails, so `isEnabled()` kept agreeing and the switch showed ON until the process restarted and the shortcut turned out never to have existed.

Now `prefs.edit().putString(…).commit()` with the result returned, and a refusal calls `clear()` so the map goes back as well — without that, `refreshBiometricUnlockEnabled()` would immediately publish the stale true it had just written. Covered by a `RefusingPreferences` test in the same shape as the ones `SharedPreferencesPasscodeStore` already had.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
aminsato: Fixed in cf9eb058b. Correct — the platform revokes on a new enrolment or once all biometrics are deleted, so "added or removed" promised a revocation that never happens when one of two enrolled fingers goes.

The class KDoc had the same gap one paragraph up: it called this the Android counterpart of iOS's `.biometryCurrentSet`, which does invalidate on any change to the set. Both now say what Android actually guarantees, and that the set which turned the shortcut on is not necessarily the set that can still use it.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
aminsato: Fixed in cf9eb058b. `gradle/libs.versions.toml:6` sets `minSdk = "30"`, consumed by both `app/` and `data/build.gradle.kts:21`, and `VERSION_CODES.R` is 30 — so `SDK_INT >= R` was always true and the branch was unreachable.

Removed it along with the `android.os.Build` import it was the only reader of, and `setUserAuthenticationParameters` now sits directly in the builder chain rather than behind an `apply`. `:app:lintDebug` and `:data:lintDebug` both still pass.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
```diff
@@ -0,0 +1,282 @@
...
+@Inject
+constructor(private val prefs: SharedPreferences) : BiometricUnlockStore {
+
+    override fun isEnabled(): Boolean = readBlob() != null && loadKeyOrNull() != null
+
+    override fun encryptCipherOrNull(): Cipher? =
+        try {
+            // Replaces whatever is there: generating over an existing alias overwrites it, so a
+            // second enable cannot leave the first copy readable.
+            val key = generateKey()
```
rkokhatskyi: `generateKey` throws `java.security.ProviderException` for any keystore failure other than a missing enrolment, and that is a RuntimeException neither catch covers, so it escapes the plain launch at `PasscodeSettingsViewModel.kt:91` and crashes the app instead of failing closed as this file's header promises

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
```diff
@@ -0,0 +1,282 @@
...
+            // And drops the copy that key used to open, in the same breath. The blob is already
+            // unreadable at this point, and a prompt the user then cancels would otherwise leave
+            // an unreadable copy behind that isEnabled() — which asks whether the alias exists —
+            // would go on reporting as a working shortcut.
+            prefs.edit(commit = true) { remove(KEY_WRAPPED_KEY) }
```
rkokhatskyi: `prefs.edit(commit = true)` here discards the Boolean that `store` now acts on at `BiometricUnlockStore.kt:144`, so a refused removal strands the old blob under the fresh alias, and the decrypt arm at :178 returns null without the clear its sibling performs, so it never heals

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
```diff
@@ -401,6 +478,13 @@ internal class PasscodeRepositoryImpl(
     override suspend fun disablePasscode(passcode: String): PasscodeUnlockResult =
         mutex.withLock {
             verifyLocked(passcode) { key ->
+                // The biometric copy goes first, before anything else moves. It holds this data
+                // key, and the moment the credentials are gone it is the only thing that does — so
+                // a crash anywhere after this point leaves no copy of a retired key behind, which
+                // is the invariant BiometricUnlockStore relies on instead of iOS's binding blob.
+                withContext(dispatcher) { biometrics.clear() }
```
rkokhatskyi: This clears the biometric copy and deletes its keystore key before `unprotectAll` can fail, so the Failed returns at :499 and :523 leave the shortcut off while the KDoc at `PasscodeRepository.kt:81` still promises nothing was changed

rkokhatskyi (dismissed):

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
aminsato: Fixed in e0a2c933d. Real. The enrolment case is the one that was already covered — `KeyGenerator.init(spec)` raises `InvalidAlgorithmParameterException` for it, a `GeneralSecurityException` — and every other keystore failure comes back as `ProviderException`, which Kotlin does not check and neither arm caught. From the bare `viewModelScope.launch` at `PasscodeSettingsViewModel.kt:91` that is the process, over a shortcut that is optional.

Refused it at every keystore call in the file rather than only at generation: `Cipher.init` and `doFinal` on a keystore key raise the same thing, and `loadKeyOrNull` and `clear` run inside `setPasscode` and `disablePasscode`, which catch nothing — the same path `keyStoreOrNull` was added for last round.

No test for this one, and I would rather say so than pretend: `generateKey` cannot run on the JVM at all (the provider is not there, so the call fails with the checked exception first), and a device cannot be made to produce a `ProviderException` on demand either — an emulator with nothing enrolled gives the checked one that was already handled.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/BiometricUnlockStore.kt
aminsato: Fixed in e0a2c933d. Both halves real.

Fixed the first by ordering rather than by adding a second repair path: the copy is dropped *before* the alias is replaced, and the removal's own `commit()` decides whether the key is minted at all. A refused removal now touches nothing — the old copy and the key it was made under both still stand — so a shortcut that works goes on working and only this enable refuses. Nothing can be stranded under a fresh alias, so there is no recovery to get right.

The heal is worth having regardless of how a mismatch arose, so `AEADBadTagException` now clears the copy. Narrow on purpose: an authentication that has expired arrives as `IllegalBlockSizeException`, and the copy behind that one is still good, so clearing on `GeneralSecurityException` wholesale would delete working shortcuts over a timeout.

`clear()` still ignores its own answer, and that one is deliberate: the alias goes immediately after and `isEnabled()` wants both, so a removal the disk refuses still leaves the shortcut reading off. There is a comment there now saying why it differs from the two writes that do act on it.

Covered by two tests in `KeyStoreBiometricUnlockStoreTest` — the mismatched copy being dropped, and a round trip pinning that an ordinary read does not drop it.

data/src/main/kotlin/com/vultisig/wallet/data/passcode/PasscodeRepository.kt
aminsato: Fixed in e0a2c933d. Real — `Failed` promised nothing was changed, and a user who hit one unreadable keyshare lost the shortcut with it and had to set it up again behind a fresh prompt.

Moved the clear to immediately before `store.clearCredentials()` inside the `NonCancellable` block rather than to after `unprotectAll` returns. That keeps what the original ordering was for: the copy holds this data key, and the moment the credentials are gone it is the only thing that does, so it must not outlive them. :499 now leaves the shortcut exactly as it found it.

:523 cannot be given the same treatment — remaking the copy needs another biometric prompt, so a `clearCredentials` that throws does leave the shortcut off. Rather than quietly diverge from the doc you quoted, `Failed` now promises what it can keep: the passcode is unchanged, still in force if it was set. The arm says why the shortcut is not.

`a disable that fails leaves the biometric shortcut where it found it` pins :499, and fails on the old ordering.

johnnyluo (approved):
LGTM

PR #5787, opened 2026-08-31
fix(token detail): fit five actions instead of clipping the fifth

### LINKED ISSUE #5778: [Fix] Token detail: 5th action (Receive) is truncated — 5 buttons must share width [Token Detail]
**Screen:** token detail sheet (screenshot is USDT).

**Bug:** with all 5 actions visible (Swap, Send, Buy, Function, Receive) the last button is clipped at the sheet edge. "Receive" wraps to three lines (`Rec` / `eiv` / `e`).

**Wanted:** if the row needs 5 buttons, they all shrink together to the same size so every label stays on one line. Do not keep the first four at full size and clip the fifth.

**Today:**
- `AssetActionButton` icon box is a fixed `52.dp` (`AssetActionButton.kt:44`).
- Token-detail row uses `Arrangement.spacedBy(20.dp)` and no `weight` (`TokenDetailScreen.kt:184-229`).
- 5 × 52 + 4 × 20 = 340.dp for the icons alone, plus 24.dp horizontal padding each side (`ContentPadding`) → 388.dp minimum. Overflows a typical 360-width phone.
- Same 5-button set on the chain screen (`ChainTokensScreen.kt:204-245`, 12.dp gaps — less tight, same pattern). Home `TxButtons` is 4 max, not this bug.

**Repro:** open token detail on a chain that shows Swap + Send + Buy + Function + Receive (ETH USDT in the screenshot).

**Fix direction:** give every `AssetActionButton` in the row equal width (`Modifier.weight(1f)`) and size the icon box from that width so all five stay identical. Shrink the gap if needed so labels stay single-line. Same treatment on `ChainTokensScreen` if 5 buttons can appear there.

token detail fifth button truncated

aminsato:
## Problem

With all five actions visible the token detail row asked for `5 × 52 + 4 × 20 = 340dp` of icons alone, plus 24dp of content padding each side. On a 360dp phone that leaves 312dp, so **Receive** was pushed off the sheet edge and its label broke across three lines.

## Fix

The gap closes before anything else does. `AssetActionRow` measures each button's natural width — the icon box, or the label when the label is wider — and spends the leftover space on the gaps, from the caller's maximum down to an 8dp floor. Only once that floor is spent do the buttons themselves give way, and then **all of them equally**, with the icon box sized from the shared width, so the fifth is never the only one squeezed.

That is iOS's design (`CoinActionsView.swift` computes the same clamped spacing around a fixed 52pt icon), not the `weight(1f)` the issue suggests: weighted children fill the row, which would spread the four- and three-button rows that are correct today. **Up to four buttons the row resolves to natural widths at the full gap, so those cases are unchanged.**

`ChainTokensScreen` uses the same row, holding its gap at its current 12dp, since the same five actions can appear there. Home `TxButtons` is untouched — it is capped at four.

## Locales are the binding constraint, not English

Labels measured out of `brockmann_bold.ttf` at 12sp, against 312dp of content width:

| Locale | Natural sum | Outcome |
|---|---|---|
| en, pt | 260dp | 13dp gap, natural widths, nothing truncated |
| de, nl | 273–278dp | 8.5–9.8dp gap, natural widths |
| es, ru | 292dp / 395dp | cannot fit at any gap → 56dp per button, icons still 52dp, labels ellipsize |

Russian `Отправить` alone is 101.5dp, so its five labels can never all be shown in full. Ellipsis at a row that fits is the degradation; today it overflows the screen instead. A 320dp phone puts English into that branch too, at 48dp buttons.

## Before / After

Captured on the Small_Phone AVD (720×1280 @320dpi = exactly 360dp wide). Medium_Phone is 411dp and does not reproduce the bug.

| Before | After |
|--------|-------|
| before | after |

Three-action sheet on the same build, to show the unchanged case:

## Test plan

- `AssetActionRowLayoutTest` — 10 tests over the pure layout function: the four-button row keeps the full gap, the five-button row closes the gap instead of shrinking, the gap never drops below the floor, a row too narrow for its labels shrinks every button by the same amount and still fits, and a degenerate row never asks for a negative width.
- `:app:testDebugUnitTest` — 2277 tests, 0 failures.
- `:app:lintDebug`, `:app:compileDebugAndroidTestKotlin`, `ktfmtFormat`.
- On device, at 360dp: token detail with Swap / Send / Buy / Function / Receive, and the same five on the chain screen.

Closes #5778

## Summary by CodeRabbit

* **New Features**
  * Added a responsive action row for token and chain asset screens.
  * Action buttons now adapt their widths and spacing to available space.
  * Added support for displaying all available token actions in previews.

* **Improvements**
  * Long action labels are truncated cleanly to remain on one line.
  * Available actions remain dynamically tailored, including send, receive, swap, buy, and deposit.

* **Tests**
  * Added coverage for action-row sizing, spacing, narrow layouts, and empty or single-action states.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -374,6 +374,7 @@ class PreviewActivity : ComponentActivity() {
                     "asset_action_button" -> AssetActionButtonPreview()
                     "token_detail_sheet_loading" -> TokenDetailSheetLoadingPreview()
                     "token_detail_sheet_full" -> TokenDetailSheetFullPreview()
+                    "token_detail_sheet_actions" -> TokenDetailSheetFullPreview(allActions = true)
                     "camera_button" -> CameraButton(onClick = {})
                     "banner" -> BannerPreview()
                     "send_tx_done" -> SendTxDonePreview()
@@ -3666,7 +3667,7 @@ private const val CHART_POINT_COUNT = 48
  * that actually exercises the partial rest position, the drag up to full, and the faded cut edge.
  */
 @Composable
-private fun TokenDetailSheetFullPreview() {
+private fun TokenDetailSheetFullPreview(allActions: Boolean = false) {
     val points =
         List(CHART_POINT_COUNT) { index ->
             val price = 73.5 + sin(index / 5.0) * 6.0 + index * 0.12
@@ -3695,6 +3696,8 @@ private fun TokenDetailSheetFullPreview() {
                     ),
                 canSwap = true,
                 canBuy = true,
+                canDeposit = allActions,
+                chainAddress = if (allActions) "9ceRgz57Jj1kmDBQtBWJyeSRhSpxvzWLPQb1LzGxKPWa" else "",
                 chart =
                     ChartUiModel(points = points, isPositive = true, changePercentText = "+4.21%"),
                 marketStats =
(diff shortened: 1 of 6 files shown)


johnnyluo (approved):
LGTM

PR #5791, opened 2026-09-01
fix(tokens): add TON and Tron custom-token search

### LINKED ISSUE #5771: [Fix] TON and Tron custom-token add always fails — SearchTokenUseCase has no branch for either chain [Tokens]
# [Fix] TON and Tron custom-token add always fails — SearchTokenUseCase has no branch for either chain [Tokens]

## Problem

Both chains show a fully working-looking "+ Custom" tile that fails 100% of the time, for every address, with a misleading "token not found".

`SearchTokenUseCase.kt:44-51` dispatches by chain and has no arm for either:

```kotlin
private suspend fun dispatch(chain: Chain, chainId: String, address: String): CoinAndPrice? =
    when {
        chain.standard == EVM -> searchEvmToken(chainId, address)
        chain.standard == SOL -> searchSolToken(address)
        chain == Chain.Terra || chain == Chain.TerraClassic -> searchTerraToken(chain, address)
        chain == Chain.Sui -> searchSuiToken(address)
        else -> null
    }
```

`Chain.Ton` and `Chain.Tron` fall to `else -> null`. Address validation is not the blocker — `isValidAddressOn` falls through to `chainAccountAddressRepository.isValid` → `chain.coinType.validate(address)`, which accepts an `EQ…` jetton master and a `T…` TRC20 contract, so a perfectly valid address reaches `else -> null` and the ViewModel reports not-found.

The affordance is genuinely shown, and is not chain-gated:
- `TokenSelectionScreen.kt:38,120` — `hasCustomToken = true`, hardcoded; the tile renders via `TokenSelectionList.kt:162`.
- `ChainTokensScreen.kt:273` and `ChainTokensTabMenuAndSearchBar.kt:82` render the "Edit tokens" pill unconditionally.
- `Chain.kt:170-190` does define `canSelectTokens` (true for both Tron and TON), but it is written in `ChainTokensViewModel.kt:73,369` and read by no composable — dead in the shipped v2 UI, so it gates nothing either way.

Because neither chain has auto-discovery either (`TokenRepository` returns an empty list and `GetChainTokensUseCase` has no arm), any TON jetton outside the 9 curated entries and any TRC20 outside the 5 curated entries is simply unreachable: the balance is invisible and the token cannot be sent. A Vultisig vault is MPC with no exportable seed, so the user cannot fall back to another wallet — they have to use the iOS or desktop client.

**This exact defect shape has been filed and fixed twice already, per chain.** Sui was #5507, closed by merged PR #5514 (`486b82af1`, "fix(sui): add a Sui branch to custom token search"), whose commit message notes the "+ Custom" tile is "offered on every chain". Terra was #5164. TON and Tron are what remain.

The data needed already exists on Android: TON jetton masters expose their metadata via toncenter `/v3/jetton/masters`, already wired as `TonApi.getJettonMetadata`; TRC20 contracts expose `symbol()`/`decimals()` via `triggerconstantcontract`, which `TronApi` already issues for fee estimation.

## Expected Behavior

Pasting a valid jetton-master or TRC20 contract address into "+ Custom" resolves its ticker and decimals and offers an "Add token" button, as it already does on EVM, Solana, Terra and Sui — and as iOS and Windows do for these two chains.

## Steps to Reproduce

Default state, shipped UI, no flags:

1. Open a vault with TON (or Tron) enabled.
2. Home → tap the TON (or Tron) chain row.
3. Tap the "Edit tokens" pill.
4. Tap the "+ Custom" tile.
5. Paste a valid jetton master address (any TON jetton outside the 9 curated ones) or a valid TRC20 contract (JST, WIN, SUN — anything outside the 5 curated ones) and search.
6. Observed: spinner, then "token not found" with a Retry that can never succeed — for every address, regardless of validity.

Expected: ticker and decimals resolve and an "Add token" button appears.

## Solution

Mirror the merged Sui fix twice, entirely inside the token-search layer. Add a TON resolver using the existing `tonApi.getJettonMetadata(masterAddress)`, and a Tron resolver issuing `symbol()` and `decimals()` via `triggerconstantcontract` — the same shape `TronApi` already uses — failing closed when either is unresolvable so a token is never added with guessed decimals. Then add the two `when` arms and bind both in `DataUsecasesModule`. The existing `hasSaneMetadata()` ticker/decimal guard already screens bad responses.

Prefer the curated `Coins.Ton` catalog entry for logo and `priceProviderID` when the master address matches, the way `SearchSuiTokenUseCase` does.

Worth doing in the same change, since it is the reason the failure is confusing rather than merely absent: make `TokenSelectionScreen`'s `hasCustomToken` respect `Chain.canSelectTokens` instead of the hardcoded `true`, and have a composable actually read the already-computed-but-dead `ChainTokensViewModel.canSelectTokens` — so chains with no custom-token concept stop offering a control that cannot work. Windows already gates its prompt this way.

### Must Do
- [ ] Add a TON custom-token resolver built on the existing `TonApi.getJettonMetadata`, producing a `Coin(chain = Ton, ticker, decimal, contractAddress, isNativeToken = false)`.
- [ ] Add a Tron custom-token resolver reading `symbol()` and `decimals()` via `triggerconstantcontract`, failing closed if either is unresolvable.
- [ ] Add `chain == Chain.Ton ->` and `chain == Chain.Tron ->` arms to `SearchTokenUseCaseImpl.dispatch`.
- [ ] Bind both in `DataUsecasesModule`.
- [ ] Prefer the curated catalog entry for logo/`priceProviderID` when the address matches a known token.
- [ ] Gate the "+ Custom" affordance on `Chain.canSelectTokens` instead of the hardcoded `true`, and make the dead `ChainTokensViewModel.canSelectTokens` actually drive the UI.
- [ ] Extend `SearchTokenUseCaseImplTest` with resolve / invalid-address / null-response cases for both chains, matching the existing EVM, Solana and Sui arms.

### Must NOT Do
- Don't change unrelated code
- Don't add a Ripple branch — XRP tokens are covered by open #5211 and #5212 and already have their own discovery
- Don't fold in the Cosmos-family chains — their token model is bank/IBC denoms rather than a contract address, which turns a bounded parity fix into an open-ended one
- Don't touch Cardano native tokens — separate, and open #5713 already owns part of it
- Don't touch Tron fee or energy estimation — open #5479 owns that
- Don't address this by adding more curated jettons; the point is user-supplied tokens
- Don't change any signing or broadcast code — the fix stops at producing a `Coin`. No keysign, `TonHelper` or `TronHelper` changes.

## Acceptance Criteria
- [ ] Pasting a valid jetton master address into "+ Custom" on TON resolves ticker and decimals and can be added
- [ ] Pasting a valid TRC20 contract address on Tron does the same
- [ ] An invalid or non-existent address on either chain still reports not-found cleanly, and a contract whose `decimals()` cannot be read is refused rather than added with a guessed value
- [ ] Added tokens show their balance on the chain screen
- [ ] The "+ Custom" tile is no longer offered on chains where custom tokens cannot work
- [ ] EVM, Solana, Terra and Sui custom-token add are unchanged

johnnyluo:
## Summary
- `SearchTokenUseCase` had no dispatch arm for TON or Tron, so a valid jetton master / TRC20 contract always reported "not found" via the "+ Custom" tile.
- Add `SearchTonTokenUseCase`, resolving ticker/decimals via the existing `TonApi.getJettonMetadata`, preferring the curated `Coins.Ton` catalog entry when the master matches.
- Add `SearchTronTokenUseCase`, resolving `symbol()`/`decimals()` via a new `TronApi.readContractConstant` (`triggerconstantcontract`), failing closed when either read is unresolvable so no token is added with guessed decimals.
- Gate the "+ Custom" tile (`TokenSelectionScreen`) and the "Edit tokens" pill (`ChainTokensTabMenuAndSearchBar`) on `Chain.canSelectTokens` instead of a hardcoded `true`.

## Issue
Closes #5771

## Test Plan
- [x] `./gradlew :data:testDebugUnitTest` passes, including new `SearchTokenUseCaseImplTest` cases for TON/Tron and a new `SearchTronTokenUseCaseImplTest` for ABI decode/fail-closed behavior
- [x] `./gradlew lint` passes
- [ ] Manual: paste a valid jetton master / TRC20 contract into "+ Custom" on TON/Tron and confirm ticker+decimals resolve

## Summary by CodeRabbit

* **New Features**
  * Added token search support for TON and TRON networks, including curated and newly discovered custom tokens.
  * Added TRON contract constant reading for token metadata.
  * Custom-token options and editing controls now appear only when supported by the selected chain.

* **Bug Fixes**
  * Improved token identification to prevent ticker collisions for custom TON and TRON tokens.
  * Updated existing token records to use contract-qualified identifiers.

* **Tests**
  * Added coverage for token searches, metadata decoding, address validation, migrations, and failure scenarios.


data/src/main/kotlin/com/vultisig/wallet/data/models/Coin.kt
```diff
@@ -26,12 +26,13 @@ data class Coin(
...
-     * their own `USD` trust line. Every other coin type keeps the plain `ticker-chainId` form
-     * unchanged.
+     * their own `USD` trust line. TON jettons and TRON TRC-20 tokens are also contract-qualified:
+     * their symbols are user-controlled metadata, so two unrelated contracts can both report the
+     * same ticker. Every other coin type keeps the plain `ticker-chainId` form unchanged.
      */
     val id: TokenId
         get() =
-            if (isSecuredAsset() || isRippleIssuedToken) {
+            if (isSecuredAsset() || isRippleIssuedToken || isContractQualifiedCustomToken()) {
```
aminsato, line 35: `Coin.id` now contract-qualifies TON/TRON tokens, but the persisted mirror of it in `TokenValueEntity.tokenId` was not updated — its doc comment explicitly says "Both coin types Coin.id contract-qualifies must be qualified here too, or a cached balance row collapses onto the plain `ticker-chain` id and no longer matches its coin".

A cached TON jetton row still yields `USDT-Ton` while the coin is now `USDT-Ton-EQCxE6...`, so in `BalanceRepository.getCachedTokenBalances` both `prices.find { it.first == tokenEntity.tokenId }` and `coins.find { it.id == tokenEntity.tokenId }?.decimal ?: 0` miss, and `AccountsRepository.fetchAccountFromDb`'s `it.coinId == acc.token.id` never matches.

Concrete effect: on every cold start / offline open, every TON jetton and every TRC-20 token (including the curated Tron USDT) shows no cached balance and no cached fiat value until the live fetch lands; where the value does render it is scaled with `decimals = 0`. `isContractQualifiedCustomToken()` needs an equivalent `isContractQualifiedCustomTokenLike()` in `TokenValueEntity`.

app/src/main/java/com/vultisig/wallet/data/db/migrations/MIGRATIONS.kt
```diff
@@ -1075,3 +1075,96 @@ internal val MIGRATION_41_42 =
...
+                AND contractAddress != ''
+                AND id = ticker || '-' || chain
+            )
+            """
+                    .trimIndent()
+            )
+
+            db.execSQL(
+                """
+            UPDATE disabledCoin
```
aminsato, line 1120: This statement can never update a row. `VaultDao.disableTokenFromVault` deletes the coin row and *then* inserts the `disabledCoin` row (`deleteTokenFromVault(vaultId, tokenId)` runs first, and `VaultRepository` even comments "delete token from coins table and add it to disabled coins table"). So for any genuinely disabled token there is no `coin` row left, and both the `SET` subquery and the `WHERE ... coinId IN (...)` subquery require `coin.id = disabledCoin.coinId` to match one.

Result: legacy `disabledCoin.coinId` values for TON/TRON custom tokens stay at the un-qualified `ticker-chain` form after migration, so they no longer match the new `Coin.id` and the "disabled" record is silently orphaned. It happens not to re-add tokens today only because `TokenRepository.getTokensWithBalance` returns `emptyList()` for Ton/Tron — that is an accident of the current discovery code, not a guarantee. Either derive the contract from the curated `Coins` registry (the only ids that can be reconstructed) or drop the statement and say why in the comment; as written it reads as coverage that does not exist, and `Migration42To43Test` only asserts on the SQL text so it cannot catch this.

app/src/main/java/com/vultisig/wallet/ui/screens/v2/chaintokens/components/ChainTokensTabMenuAndSearchBar.kt
```diff
@@ -76,9 +77,11 @@ fun ChainTokensTabMenuAndSearchBar(
                     )
                 }
 
-                UiSpacer(size = 8.dp)
+                if (canSelectTokens) {
```
aminsato, line 80: `Chain.canSelectTokens` is not the same predicate as "this chain supports custom-token search". It returns `false` for `Chain.CronosChain` and `Chain.ZkSync`, but both are `standard == EVM`, so `SearchTokenUseCase.dispatch` routes them to `searchEvmToken` and custom ERC-20 lookup works there today.

This pill is the only entry point to that flow: `onEditClick` -> `ChainTokensViewModel.selectTokens()` -> `Route.SelectTokens` -> `TokenSelectionScreen` -> `+ Custom` -> `Route.CustomToken`. Hiding it means a user on Cronos or zkSync (each has exactly one registry coin, the native one) can no longer add any token at all — a regression this PR's description doesn't mention. Gating on something like `canSelectTokens || chain.standard == EVM`, or excluding those two, keeps the intended TON/TRON fix without dropping EVM custom tokens.

data/src/main/kotlin/com/vultisig/wallet/data/usecases/SearchTronTokenUseCase.kt
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /** Standard ABI dynamic-`string` return: 32B offset, 32B length, then the UTF-8 bytes. */
+    private fun decodeAbiString(hex: String): String? {
+        val clean = hex.stripHexPrefix()
+        if (clean.length < 128) return null
+        val length = BigInteger(clean.substring(64, 128), 16).toInt()
+        if (length <= 0 || clean.length < 128 + length * 2) return null
+        return clean
+            .substring(128, 128 + length * 2)
```
aminsato, line 58: `length * 2` is Int arithmetic on an attacker-chosen value and can overflow, defeating the bounds check. A TRC-20 contract whose `symbol()` returns a length word of `0x40000000` gives `length = 1073741824`, `length * 2` overflows to `Int.MIN_VALUE`, so `clean.length < 128 + (-2147483648)` is `false` and the next line calls `clean.substring(128, -2147483520)` -> `StringIndexOutOfBoundsException`.

`BigInteger(clean.substring(64, 128), 16)` (and `BigInteger(it, 16)` in `decodeUint`) also throws `NumberFormatException` when `constant_result[0]` is not pure hex. None of these are caught: `CustomTokenViewModel.searchCustomToken` runs on a plain `viewModelScope.launch` with no try/catch, so the throw escapes the coroutine and crashes the app instead of reporting "token not found" -- exactly what the rest of this use case fails closed to avoid.

Doing the width check in `Long` before narrowing keeps it fail-closed:

```suggestion
        val length =
            runCatching { BigInteger(clean.substring(64, 128), 16) }.getOrNull()?.toLong()
                ?: return null
        if (length <= 0L || clean.length < 128L + length * 2L) return null
        return clean
            .substring(128, 128 + (length * 2L).toInt())
```

data/src/main/kotlin/com/vultisig/wallet/data/usecases/SearchTonTokenUseCase.kt
```diff
@@ -0,0 +1,76 @@
...
+        val coin =
+            Coin(
+                chain = Chain.Ton,
+                ticker = metadata.ticker,
+                logo = metadata.logo.orEmpty(),
+                address = "",
+                decimal = metadata.decimals,
+                hexPublicKey = "",
+                priceProviderID = "",
+                contractAddress = masterAddress,
```
aminsato, line 43: The master address is stored exactly as the user typed it, while `curatedCoin` right below canonicalizes both sides through `tonUserFriendlyAddress` before comparing. Since this PR makes `contractAddress` part of `Coin.id`, that inconsistency now produces duplicate identities: pasting a non-curated jetton master as `EQ...` and later as the same master's `UQ...` or raw `0:hex` form yields two `Coin`s with different ids (`FOO-Ton-EQ...` vs `FOO-Ton-UQ...`) for the same asset, so the vault shows it twice and each gets its own balance/price cache row.

Canonicalizing here would make the id stable regardless of the form pasted:

```suggestion
                contractAddress = tonUserFriendlyAddress(masterAddress) ?: masterAddress,
```

Issue #5219, opened 2026-07-07
Create Referral screen: migrate to V3Scaffold for consistent padding


aminsato:
## Problem

The **Create Referral** screen (`ReferralCreateScreen.kt`) uses a bespoke `Scaffold` + `VsTopAppBar` with an ad-hoc `.padding(16.dp)` instead of the standard `V3Scaffold` used across the rest of the V3 screens.

As a result the screen's horizontal content inset (16dp) is tighter than the app-wide V3 standard (24dp), and it does not inherit the shared V3 topbar, background, and inset handling. This makes the page feel cramped and visually inconsistent with the other screens.

## Expected

- Use `V3Scaffold` so the screen picks up the standard content padding (`V3Scaffold.PADDING_HORIZONTAL` = 24dp / `V3Scaffold.PADDING_VERTICAL` = 12dp), the V3 topbar, and V3 background.
- The bottom CTA padding should reference the `V3Scaffold` companion constants rather than hardcoded values.

## Screenshots

| Before (16dp `Scaffold`) | After (`V3Scaffold`, 24dp) |
|--------------------------|----------------------------|
| ![before](https://i.imgur.com/A1sr1Lm.png) | ![after](https://i.imgur.com/GnRUcSW.png) |

## Scope

- `app/src/main/java/com/vultisig/wallet/ui/screens/referral/ReferralCreateScreen.kt`

PR #5794, opened 2026-09-01
feat(done): drive the transaction hero from the decoder

aminsato:
Mirrors [vultisig-ios#5229](https://github.com/vultisig/vultisig-ios/pull/5229). Follows on from #5706 (the decoder foundation, iOS [#5217](https://github.com/vultisig/vultisig-ios/pull/5217)) under the parent ticket #5657.

## Summary

The normal-send Done card already matched the design; what was missing was the data behind it. It now reads the final `KeysignPayload` on both the initiating and the co-signing device, so both describe the same signed transaction instead of always defaulting to **Sent**.

- Past-tense Done vocabulary for all 28 `DecodedOperation` cases, localized across every shipping locale. Plain transfers, generic contract calls, and unreadable transactions keep the existing **Sent** default.
- Amounts, assets, and fiat come from the signed content and the vault's own trusted coins. A zero, a carrier amount, or a quantity that rounds away at display precision degrades to the bare verb rather than being presented as the operation amount.
- Projections resolve behind a 5s deadline. A failed or slow position read never blocks the screen and never fabricates a number.
- Richer providers keep precedence: an XRPL TrustSet, a limit-order cancellation, and a Blockaid simulation keep their layouts and figures, and only borrow the Done verb.
- Both swap Done paths and the existing card's layout are unchanged.

## Please read before reviewing: this is wiring, not yet visible behaviour

`SignedTransactionDecoder`'s registry is still empty on Android — the per-chain decoders (iOS #5219–#5224) and the Verify integration (iOS #5218) are not mirrored yet. Until a chain decoder registers, every transaction decodes to `Unknown`, lands in the fallback set, and the Done card renders exactly as it does today.

So this PR is intentionally a visual no-op. It is the layer the chain decoders plug into, and the acceptance criteria about verbs replacing **Sent** can only be demonstrated once those land.

## Three deliberate divergences from the iOS diff

Android's architecture differs, so a literal port would have been worse:

1. **No `TransactionHeroResolver` surface/provider machinery.** Android already encodes that precedence as branch order in `SendTxOverviewScreen`'s `tokenContent` (trustSet → signRipple → default). `operationHero` reaches only the default branch, so the stronger providers keep precedence without a new indirection layer.
2. **Resolution lives in `KeysignViewModel`, not the composable.** Both devices construct that ViewModel from the same payload, which is what gives initiator/co-signer parity. iOS uses a per-screen `.task(id:)`.
3. **No `refreshedFiat` / rate-change re-render path.** Fiat is priced once in the resolving coroutine through `ConvertTokenValueToFiatUseCase`, which already refreshes on a zero price. Amounts are formatted with `TokenValueToDecimalUiStringMapper` so a decoder-driven card reads identically to the one it replaces — same grouping, precision, and large-value suffixes.

Also skipped on purpose: `HeroContent.Receive` (belongs to iOS #5218 and is unreachable here) and `scopeDelegatedAmountAfterRent` (`DecodedAmount.AccountFunding` has no Android equivalent).

## Localization

28 keys across all 10 locale files. German, Spanish, Croatian, Italian, Korean, Portuguese, and Chinese come from the iOS strings.

**Dutch and Russian have no iOS source** — those two were written against the terminology already established in each locale file (`Gestaked` / `Gebonden` / `Limietorder`, `Застейкано` / `Привязано` / `Лимитный ордер`). A native-speaker glance at `values-nl` and `values-ru` would be welcome.

## Test plan

- [x] `:app:compileDebugKotlin`, `:app:lintDebug` — clean
- [x] `:app:testDebugUnitTest` + `:data:testDebugUnitTest` — pass
- [x] Existing `KeysignViewModel` tests updated for the new constructor dependency
- [x] Done screen verified unchanged for send, deposit, XRPL TrustSet, dApp XRPL, and limit-order-cancel paths — with an empty decoder registry every one of them takes the existing fallback

https://claude.ai/code/session_013SErms3rPgwTB5VNa6f1U2

## Summary by CodeRabbit

- **New Features**
  - Improved completed transaction summaries with operation-specific labels, assets, amounts, fiat values, and chain details.
  - Added projected estimates and execution scopes for staking, unstaking, rewards, and supported operations.
  - Added graceful handling for unavailable, unsupported, empty, or approximate transaction amounts.
  - Enhanced token displays with optional chain logos, fiat values, scope text, and flexible formatting.

- **Localization**
  - Added translated completion labels, limit-order cancellation statuses, and staking-related messages across multiple languages.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -380,6 +380,8 @@ class PreviewActivity : ComponentActivity() {
                     "send_tx_done" -> SendTxDonePreview()
                     "send_tx_done_decoded" -> SendTxDoneDecodedPreview()
                     "send_tx_done_projected" -> SendTxDoneProjectedPreview()
+                    "send_tx_done_unresolved" -> SendTxDoneUnresolvedPreview()
+                    "send_tx_done_title" -> SendTxDoneTitleOnlyPreview()
                     "deposit_mint_done" -> DepositMintDonePreview()
                     "transaction_history_empty" -> TransactionHistoryEmptyState()
                     "limit_orders_tab" -> LimitOrdersTabPreview()
@@ -1612,6 +1614,30 @@ private fun SendTxDoneProjectedPreview() {
                 ),
             scope = stringResource(R.string.withdrawing_share_of_staked_position, "50%"),
         )
+    )
+}
+
+/**
+ * The guard that matters: a position read that failed or timed out. The verb and the asset are
+ * still true, so both are shown — the amount and the fiat line are omitted rather than filled with
+ * the payload's carrier value, which for a fractional unstake is dust.
+ */
+@Composable
+private fun SendTxDoneUnresolvedPreview() {
+    SendTxDoneWithOperationHero(
+        HeroContent.Projected(
+            title = stringResource(R.string.done_verb_unstaked),
+            estimate = null,
+            scope = stringResource(R.string.withdrawing_share_of_staked_position, "50%"),
+        )
+    )
+}
+
+/** A reading that names the operation but states no quantity at all. */
+@Composable
+private fun SendTxDoneTitleOnlyPreview() {
+    SendTxDoneWithOperationHero(
+        HeroContent.Title(stringResource(R.string.done_verb_claimed_rewards))
     )
 }
 
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -1592,7 +1592,7 @@ private fun SendTxDoneDecodedPreview() {
                 HeroCoinAmount(
                     amount = "1,250.5",
                     ticker = "TCY",
-                    logo = Coins.ThorChain.RUNE.logo,
+                    logo = Coins.ThorChain.TCY.logo,
                     fiatValue = "$412.66",
                 ),
         )
@@ -1609,7 +1609,7 @@ private fun SendTxDoneProjectedPreview() {
                 HeroCoinAmount(
                     amount = "625.25",
                     ticker = "TCY",
-                    logo = Coins.ThorChain.RUNE.logo,
+                    logo = Coins.ThorChain.TCY.logo,
                     fiatValue = "$206.33",
                 ),
             scope = stringResource(R.string.withdrawing_share_of_staked_position, "50%"),
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -378,6 +378,8 @@ class PreviewActivity : ComponentActivity() {
                     "camera_button" -> CameraButton(onClick = {})
                     "banner" -> BannerPreview()
                     "send_tx_done" -> SendTxDonePreview()
+                    "send_tx_done_decoded" -> SendTxDoneDecodedPreview()
+                    "send_tx_done_projected" -> SendTxDoneProjectedPreview()
                     "deposit_mint_done" -> DepositMintDonePreview()
                     "transaction_history_empty" -> TransactionHistoryEmptyState()
                     "limit_orders_tab" -> LimitOrdersTabPreview()
@@ -1569,6 +1571,76 @@ private fun tonJettonSendState(decoded: Boolean): VerifyTransactionUiModel {
         consentAddress = false,
         consentAmount = false,
         hasFastSign = false,
+    )
+}
+
+/**
+ * The done card as the decoder drives it: the past-tense verb, and the asset and amount read from
+ * the signed content rather than from the payload's carrier coin.
+ *
+ * Stands in for a registered chain decoder, which the app does not have yet — every real payload
+ * currently decodes to `Unknown` and takes the [SendTxDonePreview] fallback instead.
+ */
+@Composable
+private fun SendTxDoneDecodedPreview() {
+    SendTxDoneWithOperationHero(
+        HeroContent.Send(
+            title = stringResource(R.string.done_verb_staked),
+            coin =
+                HeroCoinAmount(
+                    amount = "1,250.5",
+                    ticker = "TCY",
+                    logo = Coins.ThorChain.RUNE.logo,
+                    fiatValue = "$412.66",
+                ),
+        )
+    )
+}
+
+/** The same card for an execution-set amount, where the projected quantity stays approximate. */
+@Composable
+private fun SendTxDoneProjectedPreview() {
+    SendTxDoneWithOperationHero(
+        HeroContent.Projected(
+            title = stringResource(R.string.done_verb_unstaked),
+            estimate =
+                HeroCoinAmount(
+                    amount = "625.25",
+                    ticker = "TCY",
+                    logo = Coins.ThorChain.RUNE.logo,
+                    fiatValue = "$206.33",
+                ),
+            scope = stringResource(R.string.withdrawing_share_of_staked_position, "50%"),
+        )
+    )
+}
+
+@Composable
+private fun SendTxDoneWithOperationHero(operationHero: HeroContent) {
+    val runeCoin = Coins.ThorChain.RUNE
+
+    SendTxOverviewScreen(
+        showToolbar = true,
+        showSaveToAddressBook = false,
+        transactionHash = "9F2C1B...7A4E",
+        transactionLink = "https://runescan.io/tx/9F2C1B",
+        transactionStatus = TransactionStatus.Confirmed,
+        onComplete = {},
+        onBack = {},
+        onAddToAddressBook = {},
+        tx =
+            UiTransactionInfo(
+                type = UiTransactionInfoType.Deposit,
+                token = ValuedToken(token = runeCoin, value = "0.00000001", fiatValue = "$0.00"),
+                from = "thor1abcdefghijklmnopqrstuvwxyz0123456789ab",
+                to = "",
+                memo = "tcy-:5000",
+                networkFeeTokenValue = "0.02 RUNE",
+                networkFeeFiatValue = "$0.03",
+                operationHero = operationHero,
+            ),
+        isTransactionDetailVisible = false,
+        onTransactionDetailVisibleChange = {},
     )
 }
 
@@ -3697,7 +3769,8 @@ private fun TokenDetailSheetFullPreview(allActions: Boolean = false) {
                 canSwap = true,
                 canBuy = true,
                 canDeposit = allActions,
-                chainAddress = if (allActions) "9ceRgz57Jj1kmDBQtBWJyeSRhSpxvzWLPQb1LzGxKPWa" else "",
+                chainAddress =
+                    if (allActions) "9ceRgz57Jj1kmDBQtBWJyeSRhSpxvzWLPQb1LzGxKPWa" else "",
                 chart =
                     ChartUiModel(points = points, isPositive = true, changePercentText = "+4.21%"),
                 marketStats =
--- a/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
@@ -64,10 +64,12 @@ import com.vultisig.wallet.data.usecases.txstatus.TransactionResult
 import com.vultisig.wallet.data.usecases.txstatus.TxStatusConfigurationProvider
 import com.vultisig.wallet.data.utils.compatibleDerivationPath
 import com.vultisig.wallet.data.utils.safeLaunch
+import com.vultisig.wallet.ui.components.hero.HeroContent
 import com.vultisig.wallet.ui.models.TransactionDetailsUiModel
 import com.vultisig.wallet.ui.models.deposit.DepositTransactionUiModel
 import com.vultisig.wallet.ui.models.sign.SignMessageTransactionUiModel
 import com.vultisig.wallet.ui.models.swap.SwapTransactionUiModel
+import com.vultisig.wallet.ui.models.transactiondecoding.DoneTransactionPresentation
 import com.vultisig.wallet.ui.navigation.Destination
 import com.vultisig.wallet.ui.navigation.NavigationOptions
 import com.vultisig.wallet.ui.navigation.Navigator
@@ -284,6 +286,11 @@ internal data class KeysignUiState(
     val approveTxLink: String = "",
     val swapProgressLink: String? = null,
     val showSaveToAddressBook: Boolean = false,
+    /**
+     * The done screen's decoder-driven hero, resolved from the final signed payload. Null until it
+     * resolves, and null for every reading the normal-send card already describes better.
+     */
+    val operationHero: HeroContent? = null,
 )
 
 /**
@@ -333,6 +340,7 @@ constructor(
     private val gasFeeToEstimatedFee: GasFeeToEstimatedFeeUseCase,
     private val inAppReviewRepository: InAppReviewRepository,
     private val pendingLimitOrderRepository: PendingLimitOrderRepository,
+    private val doneTransactionPresentation: DoneTransactionPresentation,
 ) : ViewModel() {
 
     /** Creates [KeysignViewModel] with runtime-provided assisted parameters. */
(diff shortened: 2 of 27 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -435,6 +443,51 @@ constructor(
...
+
+            _state.update { current ->
+                current.copy(
+                    operationHero = hero,
+                    // A Blockaid simulation owns figures the decoder cannot improve on, so only
+                    // its verb is replaced. Nothing happens when the reading is one the normal-send
+                    // card already describes.
+                    transactionUiModel = current.transactionUiModel?.retitledForDone(payload),
+                )
+            }
```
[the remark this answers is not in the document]
aminsato: Good catch, and it was worse than one stray decode: `_state.update` runs its lambda on the calling dispatcher *and* re-runs it on CAS contention, so `retitledForDone` could have decoded repeatedly on Main.

Fixed in 05e8f5fbc by resolving both readings in the existing `Dispatchers.IO` block and passing the finished verb into the update, which now only applies it:

```kotlin
val (hero, doneVerb) =
    withContext(Dispatchers.IO) {
        doneTransactionPresentation.resolve(payload, vault.coins) to
            doneTransactionPresentation.specificTitle(payload)
    }
```

`retitleResolvedHero` lost its only caller in the process, so it's removed rather than left as dead API.

app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
```diff
@@ -1572,6 +1574,76 @@ private fun tonJettonSendState(decoded: Boolean): VerifyTransactionUiModel {
...
+@Composable
+private fun SendTxDoneDecodedPreview() {
+    SendTxDoneWithOperationHero(
+        HeroContent.Send(
+            title = stringResource(R.string.done_verb_staked),
+            coin =
+                HeroCoinAmount(
+                    amount = "1,250.5",
+                    ticker = "TCY",
+                    logo = Coins.ThorChain.RUNE.logo,
```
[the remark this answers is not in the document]
aminsato: Correct — and it was visible on the emulator: the RUNE lightning bolt rendered directly above `1,250.5 TCY`. `Coins.ThorChain.TCY` exists, so there was no reason for the mismatch. Fixed in both previews (4308bdbdc).

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
```diff
@@ -354,6 +360,99 @@ internal fun SendTxOverviewScreen(
...
+                tokenLogo = getCoinLogo(hero.coin.logo),
+                ticker = hero.coin.ticker,
+                // The decoder resolves its asset from the signed units, which may not be the
+                // payload's coin, so the payload's chain badge would mislabel it.
+                chainLogo = null,
+                value = hero.coin.amount,
+                fiatValue = hero.coin.fiatValue,
+            )
+
+        is HeroContent.Projected -> {
```
johnnyluo, line 447: [P2] Projected Done hero drops the signed scope

`HeroContent.Projected` carries the scope text from `ProjectionCoordinator` (for example, the signed fraction/whole-stake/rewards scope), but this Done path converts it into `DoneHeroDisplay`, which has no scope field, and then renders only `VsOverviewToken`. For projected/fractional operations, especially when `estimate == null`, the user sees only a past-tense verb with a blank amount and the fallback token, losing the one signed quantity the decoder actually proved. Please render `HeroContent.Projected` with `HeroContentView`, or extend `DoneHeroDisplay`/the card to display `hero.scope`.

aminsato: You're right, and the null-estimate case was the worst of it — a past-tense verb over a blank amount and the payload's carrier token, with the one quantity the decoder actually proved thrown away.

Fixed in a6db293e9 by carrying `scope` through `DoneHeroDisplay` and rendering it under the amount, rather than switching that case to `HeroContentView` — keeping one card renderer means the projected reading can't drift from the normal-send layout.

So a failed position read now reads:

    Unstaked
    RUNE
    50% of your staked position

and a resolved one keeps the estimate above it (`≈ 625.25 TCY`, $206.33) with the exact scope underneath.

Worth flagging one thing this does not fix, since it's your call: when no estimate resolves, the ticker still falls back to the payload's coin, so a TCY unstake names RUNE. That matches iOS (`crypto = input.coin.ticker` in its `DoneHeroDisplay`) and the scope line now carries the real meaning, but if you'd rather drop the ticker entirely in that case I'll change it.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
```diff
@@ -354,6 +361,107 @@ internal fun SendTxOverviewScreen(
...
+                )
+            } else {
+                // The position read resolved nothing, so the scope is the only quantity the
+                // decoder proved. Dropping it here would leave a verb over a blank amount and
+                // discard the one figure that is certain.
+                fallback.copy(verb = verb, value = "", fiatValue = null, scope = hero.scope)
+            }
+        }
+
+        is HeroContent.Title -> fallback.copy(verb = verb, value = "", fiatValue = null)
```
johnnyluo: [P2] Title-only Done hero keeps unrelated token branding

`HeroContent.Title` is meant to be an assetless completed-action hero, but this branch copies the fallback card and only clears `value` and `fiatValue`. That leaves the payload token logo/ticker visible, so cases like a rewards claim can render as `Claimed rewards` with unrelated RUNE branding underneath. Build an assetless title-only display here, or route this case through `HeroContentView`, so the completed verb is not paired with a misleading token.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
aminsato: Both parts now done.

**Scope restored** (a6db293e9) — `scope` travels through `DoneHeroDisplay` and renders under the amount. I kept the single card renderer rather than switching this branch to `HeroContentView`, since two renderers for the same card is how the projected reading drifts from the normal-send layout later.

**Carrier asset dropped** (2266e9743) — following on from your point: with no estimate, falling back to the payload's coin named RUNE under a verb reading "Unstaked", because a TCY unstake is signed against a dust RUNE amount. That was a claim the reading never made, so the asset row goes with the amount.

An unresolved projection now reads:

    Unstaked
    50% of your staked position

and a resolved one keeps the estimate with the exact scope beneath it:

    Unstaked
    ≈ 625.25 TCY
    $206.33
    50% of your staked position

Compile, lint, and both unit suites pass. Not visually re-verified — my emulator dropped mid-session — so the two amountless previews (`send_tx_done_unresolved`, `send_tx_done_title`) are worth a look if you're running it.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
aminsato: Right, and I had a screenshot of exactly this without registering it — the `send_tx_done_title` preview rendered `Claimed rewards` over a RUNE logo and ticker.

Fixed in 6f7e5e57f. The branch was copying the fallback card and clearing only `value` and `fiatValue`, so the payload token's logo and ticker survived. It now builds an assetless display: the asset row is dropped along with the amount, leaving the verb alone.

I kept it on `VsOverviewToken` rather than routing to `HeroContentView` — 2266e9743 (your other thread) already taught the card to drop the asset row on an empty ticker for the unresolved-projection case, so this is the same mechanism rather than a second renderer that can drift.

    Claimed rewards

Compile and unit suites pass. Not visually re-verified — my emulator dropped mid-session — so the spacing with no asset row is the part worth a glance in `send_tx_done_title` and `send_tx_done_unresolved`.

app/src/main/java/com/vultisig/wallet/ui/models/transactiondecoding/ProjectionCoordinator.kt
```diff
@@ -0,0 +1,72 @@
...
+        /** Short enough that optional chain state cannot stall the done screen. */
+        val TIMEOUT = 5.seconds
+
+        /**
+         * Applies the same deadline to every optional position read. Every failure — a timeout, a
+         * network error, an empty answer — degrades to the existing scope rather than fabricating a
+         * number.
+         */
+        suspend fun estimate(read: suspend () -> HeroCoinAmount?): HeroCoinAmount? =
+            withTimeoutOrNull(TIMEOUT) { runCatching { read() }.getOrNull() }
```
johnnyluo: `runCatching` also catches `CancellationException`. Timeout is already handled by `withTimeoutOrNull`, but other cancellation, like the ViewModel being cleared, should propagate instead of being converted into a silent null estimate. Please rethrow cancellation and only swallow ordinary reader failures.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
```diff
@@ -354,6 +361,126 @@ internal fun SendTxOverviewScreen(
...
+
+    return when (hero) {
+        is HeroContent.Send ->
+            DoneHeroDisplay(
+                verb = verb,
+                tokenLogo = getCoinLogo(hero.coin.logo),
+                ticker = hero.coin.ticker,
+                // The decoder resolves its asset from the signed units, which may not be the
+                // payload's coin, so the payload's chain badge would mislabel it.
+                chainLogo = null,
```
johnnyluo: Decoded amount heroes always drop the chain badge here. The existing card shows that badge for non-native tokens and L2 natives, so once decoded units start rendering, assets like USDC on Base/Arbitrum lose chain context even though the trusted `Coin` still has it. Could this mirror the fallback badge rule for the resolved coin instead of always using `null`?

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
```diff
@@ -163,14 +168,16 @@ internal fun SendTxOverviewScreen(
                         horizontalAlignment = Alignment.CenterHorizontally,
                         modifier = Modifier.fillMaxWidth(),
                     ) {
+                        val display = doneHeroDisplay(tx)
```
johnnyluo: This new Done display is only reached from the fallback block, after `TransactionHero` has already preferred `heroContent`/`functionName`. That means decoded contract actions with a function-name fallback never show the decoder-driven Done hero; for example an unverified `approve` would still render the raw function name instead of the localized `Approved` verb/amount/scope. Consider letting `operationHero` participate before the function-name fallback, or suppressing `functionName` when a decoder Done hero exists.

app/src/main/java/com/vultisig/wallet/ui/models/transactiondecoding/ProjectionCoordinator.kt
aminsato: Agreed, fixed in a44b22a37. `runCatching` catches `Throwable`, so it was swallowing the `CancellationException` from a cleared view model — and, more subtly, the `TimeoutCancellationException` that `withTimeoutOrNull` relies on seeing thrown out of its block.

Now follows the same shape as `SignedTransactionDecoder.decode`:

```kotlin
withTimeoutOrNull(TIMEOUT) {
    try {
        read()
    } catch (cancellation: CancellationException) {
        throw cancellation
    } catch (_: Exception) {
        null
    }
}
```

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
aminsato: Good point — fixed in a44b22a37. My reasoning for `null` was that the decoder resolves its asset from the signed units rather than the payload coin, so the *payload's* badge would mislabel it. But the resolved `Coin` is right there in the presentation layer with its own chain, and I was throwing it away instead of using it.

`HeroCoinAmount` now carries `chainLogo`, populated from the resolved coin under the same rule the card already applies:

```kotlin
chainLogo = coin.chain.monoToneLogo.takeIf { !coin.isNativeToken || coin.chain.isLayer2 }
```

So a decoded USDC keeps its Base or Arbitrum badge, and a native asset still shows none. The Blockaid path passes nothing and is unchanged.

Note this diverges from iOS, which hardcodes `tokenChainLogo = nil` for these cases — worth knowing if the two are ever diffed.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
aminsato: This is the one that mattered most — thank you. `TransactionHero` resolves `heroContent` → `functionName` → fallback block, and the whole decoder display lives in that fallback, so any payload with a decoded function name never reached it. An `approve` would have rendered the bare selector rather than `Approved` with its amount, which defeats the point of the PR for exactly the case it should handle best.

Fixed in a44b22a37 by suppressing the function name when a decoder reading exists:

```kotlin
functionName = tx.functionName.takeIf { tx.operationHero == null }
```

I chose that over letting `operationHero` participate earlier because it keeps the precedence readable in one place: Blockaid simulation → decoder reading → raw function name → existing card. A simulation still outranks the decoder, since it carries figures the decoder cannot improve on.

All three verified by compile, lint, and unit suites. Not visually re-verified — my emulator dropped mid-session — so the chain badge on a decoded token is the piece I would most want eyes on.

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
```diff
@@ -104,7 +109,11 @@ internal fun SendTxOverviewScreen(
         tokenContent = {
             TransactionHero(
                 heroContent = tx.heroContent.takeUnless { it == HeroContent.Unverified },
-                functionName = tx.functionName,
+                // A decoder reading outranks the raw function name: "Approved", with the amount and
+                // the asset it was read from, says more than the selector it was decoded out of.
+                // A Blockaid simulation above still wins, since it carries figures the decoder has
+                // no way to improve on.
+                functionName = tx.functionName.takeIf { tx.operationHero == null },
                 modifier = Modifier.fillMaxWidth(),
```
rkokhatskyi, line 123: `isUnverifiedFunction` (SendTxOverviewScreen.kt:78) is derived from `tx.heroContent` alone, so once a decoder resolves an approve the card falls through to a confident "Approved 500 USDC" reading via operationHero while the Unverified function caution banner still renders beneath it for that same transaction.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
```diff
@@ -435,6 +444,52 @@ constructor(
...
+                        doneTransactionPresentation.specificTitle(payload)
+                }
+
+            _state.update { current ->
+                current.copy(
+                    operationHero = hero,
+                    // A Blockaid simulation owns figures the decoder cannot improve on, so only
+                    // its verb is replaced. Nothing happens when the reading is one the normal-send
+                    // card already describes.
+                    transactionUiModel = current.transactionUiModel?.retitledForDone(doneVerb),
```
rkokhatskyi, line 487: The decoder retitle here races the pre-existing address-book job at `KeysignViewModel.kt:436`, which rebuilds transactionUiModel from the stale constructor-captured transactionTypeUiModel instead of current, so whichever _state.update lands last wins and can silently discard the decoded verb.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/screens/transaction/SendTxOverviewScreen.kt
aminsato: Good catch — fixed in fcc2f7a98.

You're right that the two disagree, and the disagreement is in the direction that matters: the card would have grown *more* confident than the banner beneath it, on the same transaction. `isUnverifiedFunction` gates the banner off `heroContent`, while the decoder reading reached the fallback block independently.

Rather than suppress the banner, I stood the reading down — a Blockaid verdict of "could not verify this function" is a claim about this exact transaction, and that keeps the precedence the screen already follows (simulation → decoder → function name → existing card):

```kotlin
val operationHero = tx.operationHero.takeUnless { isUnverifiedFunction }
```

`doneHeroDisplay` now takes the gated value rather than reading `tx.operationHero` itself, so the two cannot drift apart again.

One thing I deliberately did **not** decide, since it's a product call: an ERC-20 `approve` moves no balance, so Blockaid returning "no balance change" for one may be a false positive rather than a real warning — in which case the decoder reading is the better information and the banner is what should yield. I've taken the conservative branch (never weaken a security caution). Say the word if you'd rather it went the other way.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/KeysignViewModel.kt
aminsato: Confirmed and fixed in fcc2f7a98 — thank you, this one was invisible from my side because the decoder registry is empty, so the retitle is currently always a no-op.

The address-book job at :436 was doing:

```kotlin
_state.update {
    it.copy(transactionUiModel = transactionTypeUiModel.withResolvedLabels(...))
}
```

ignoring `it.transactionUiModel` and rebuilding from the constructor-captured model, so it clobbered whatever the decoder job had written if it happened to land second. It now layers onto current state:

```kotlin
transactionUiModel =
    (current.transactionUiModel ?: transactionTypeUiModel).withResolvedLabels(...)
```

`withResolvedLabels` is `copy`-based, so chaining preserves the decoded verb, and the reverse ordering was already safe since my job reads `current`. Worth noting this also fixes a pre-existing latent bug: the same overwrite discarded the `withResolvedSrcVaultName` applied in the first init block, harmless today only because `withResolvedLabels` happens to set `srcVaultName` too.

johnnyluo (approved):
LGTM

PR #5796, opened 2026-09-01
fix(defi): paint the last-known positions instead of cold-starting

### LINKED ISSUE #5782: [Fix] DeFi detail screens refetch on every entry — cache last-known positions [DeFi]
**Screen:** DeFi chain detail (THORChain, Maya, Solana/Kamino, TON, Circle — all of them)

**Bug:** opening a chain detail from DeFi home always cold-starts. Positions refetch from the network, the header/tabs go through loading, and the enabled set flashes back to the defaults (`RUNE` + `TCY`) until saved selection loads. Walk in, walk out, walk in again: same wait, same re-enable.

**Wanted:**
- Keep the enabled set the user saved. Do not re-apply `defaultSelectedPositionsDialog()` on a vault that already has a selection (empty list is a real choice, not "never chosen").
- Paint last-known cards immediately on re-entry. Network refresh is background or pull-to-refresh, not a blocking reload of a blank screen.

**Today:**
- Detail ViewModels are nav-scoped (`hiltViewModel()`). Pop back to DeFi home destroys them.
- Next `setData` always refetches: THOR `loadSavedPositions` → `loadBondedNodes` / `loadStakingPositions` / `reloadLpTab` (`ThorchainDefiPositionsViewModel.kt:196-206`, `495-515`). Bonded sets `isLoading = true` (`:667`).
- Initial UI model seeds `selectedPositions` with the defaults (`ThorchainDefiPositionsUiModel.kt:113`) before the repo answers — that is the re-enable flash.
- Tron already does this right: `TronDeFiSnapshotCache` paints the snapshot first, then refreshes (`TronDeFiPositionsViewModel.kt:198-216`). Cosmos has `CosmosStakingSnapshotCache`. Copy that onto the other DeFi details.

Pull-to-refresh stays the user's way to force a network read.

aminsato:
## Summary

Opening a DeFi chain detail always cold-started: blank cards, header on a spinner, and the enabled set flashing back to `RUNE + TCY` until the store answered. Walk in, walk out, walk in again — same wait, same re-enable.

Two things were needed, not one. A snapshot on its own would have shown nothing, because **every reload destroys what is on screen before it fetches** — bonded re-arms `isLoading` over the total, staking replaces its positions with placeholder cards, LP replaces its cards with zeroed placeholders. Seeding from a snapshot and then letting the load blank it a frame later changes nothing the user can see.

## What changed

**`DeFiPositionsSnapshotCache`** — in memory, process-lifetime, keyed by vault **and** snapshot type so two view-models behind one screen cannot overwrite each other. Each screen hands its state over in `onCleared` and seeds from it on the first `setData`; the loads that follow refresh it underneath. Same trade `TronDeFiSnapshotCache` and `CosmosStakingSnapshotCache` already make: no second on-disk store, nothing to clean up when a vault is deleted, and a restart reads the chain.

Wired into **THORChain, MayaChain, TON, Kamino Earn and Circle**. Routing context these screens keep outside their state travels with the snapshot — TON's pool address, Circle's MSCA account, Kamino's total coverage and priced currency — because a restored card offering an action that silently does nothing is worse than no card.

**Reloads now replace figures in place** (THORChain / Maya). Bonded shimmers only while its total is unpriced, staking cards are matched and swapped one at a time, LP cards are reused per pool. Cold start is unchanged.

## Worth a look in review

- **Staking cards are keyed by coin, not by selection key.** RUJI and its auto-compounding sRUJI card are both toggled by the single `"RUJI"` tile, so keying on that collapses the two and renders the bonded card twice. Pinned by a test.
- **LP cards are only reused for pools the last read resolved** (`livePoolKeys`). A failed read leaves zeroed placeholders standing; carrying those forward would present a figure nothing confirmed while hiding that a read is in flight.
- **A currency switch still clears the cards.** Those strings were formatted for the currency being left, so `dropFiatPricedInPreviousCurrency` drops them and each card goes back to its shimmer until it can be re-priced.
- **Solana's native staking rows are deliberately untouched.** That view-model reads its stake accounts fresh because they gate Deactivate and Withdraw, and the amounts those build come from the same read. The Solana screen opens on Earn, which is cached, so the flash on entry is gone regardless.
- **The saved-selection half of the issue was already correct.** `DefiPositionsRepository` returns `null` for "never chose" and an empty set for "cleared on purpose", and both screens already apply the defaults only on `null`. What flashed was the initial UI model's own defaults, which the snapshot now replaces.

## Test plan

Unit tests — every new test verified to fail against the pre-change code:

| Suite | Before | After |
|---|---|---|
| `ThorchainDefiPositionsViewModelTest` | 50 | 60 |
| `MayachainDefiPositionsViewModelTest` | 41 | 46 |
| `KaminoEarnViewModelTest` | 29 | 33 |
| `TonDeFiPositionsViewModelTest` | 9 | 12 |
| `CircleDeFiPositionsViewModelTest` | 5 | 8 |
| `DeFiPositionsSnapshotCacheTest` | — | 6 |

`./gradlew :app:testDebugUnitTest` green in full.

Manual, per chain (THORChain, Maya, TON, Solana, Circle):

1. DeFi tab → open the chain → let it settle → back → open it again. Cards, totals and tabs paint immediately; values refresh underneath.
2. Manage Positions → clear the selection → Done → back → re-enter. The saved set stands; `RUNE + TCY` must not flash back on.
3. Force-stop the app and reopen — the first entry still shimmers.
4. Pull-to-refresh on a settled screen — figures stay readable, they do not blank.
5. Switch currency — every card *does* shimmer and re-price.
6. THORChain → Staked with both RUJI and Compounded RUJI, then pull-to-refresh — still two distinct cards.
7. Switch vaults and open the same chain — that vault's own figures, never the previous vault's.

Closes #5782

## Summary by CodeRabbit

* **Improvements**
  * DeFi screens now remember and restore previously loaded positions when revisiting a vault.
  * Refreshes preserve settled staking, bonded, liquidity pool, and earning cards.
  * Currency changes and failed or unresolved pricing prevent stale values and show appropriate loading states.
  * Individual liquidity pool cards now display their own loading status.
  * Restored screens retain selections and transaction routing details without unnecessarily reopening pickers.

* **Tests**
  * Added coverage for restoration, refresh behavior, failure handling, and persistence across supported DeFi screens.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
@@ -740,7 +740,16 @@ constructor(
                 onError = { e ->
                     Timber.e(e, "Failed to load Maya LP positions")
                     reportLpFiat(BigDecimal.ZERO)
-                    updateModel { it.copy(lp = it.lp.copy(isLoading = false)) }
+                    updateModel {
+                        it.copy(
+                            lp =
+                                it.lp.copy(
+                                    isLoading = false,
+                                    positions =
+                                        it.lp.positions.map { p -> p.copy(isLoading = false) },
+                                )
+                        )
+                    }
                 }
             ) {
                 // The zero has to be resolved before the placeholders are built: a failed load
@@ -758,7 +767,7 @@ constructor(
                 val placeholderPositions =
                     selectedPools.map { pool ->
                         val assetTicker = pool.ticker.substringAfter("/")
-                        loaded[pool.positionKey]
+                        loaded[pool.positionKey]?.copy(isLoading = false)
                             ?: LpPositionUiModel(
                                 titleLp = "${pool.ticker} Pool",
                                 totalPriceLp = zero,
@@ -769,11 +778,11 @@ constructor(
                                 positionKey = pool.positionKey,
                                 canRemove = false,
                                 chainLogo = pool.chainLogo as? Int,
+                                isLoading = true,
                             )
                     }
-                val isCold = selectedPools.any { loaded[it.positionKey] == null }
                 updateModel {
-                    it.copy(lp = it.lp.copy(isLoading = isCold, positions = placeholderPositions))
+                    it.copy(lp = it.lp.copy(isLoading = false, positions = placeholderPositions))
                 }
 
                 val vault = withContext(ioDispatcher) { vaultRepository.get(vaultId) }
@@ -787,7 +796,12 @@ constructor(
                     reportLpFiat(BigDecimal.ZERO)
                     updateModel {
                         it.copy(
-                            lp = it.lp.copy(isLoading = false, positions = placeholderPositions)
+                            lp =
+                                it.lp.copy(
+                                    isLoading = false,
+                                    positions =
+                                        placeholderPositions.map { p -> p.copy(isLoading = false) },
+                                )
                         )
                     }
                     return@safeLaunch
@@ -798,7 +812,12 @@ constructor(
                     reportLpFiat(BigDecimal.ZERO)
                     updateModel {
                         it.copy(
-                            lp = it.lp.copy(isLoading = false, positions = placeholderPositions)
+                            lp =
+                                it.lp.copy(
+                                    isLoading = false,
+                                    positions =
+                                        placeholderPositions.map { p -> p.copy(isLoading = false) },
+                                )
                         )
                     }
                     return@safeLaunch
--- a/app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
@@ -147,6 +147,7 @@ constructor(
     private val balanceVisibilityRepository: BalanceVisibilityRepository,
     private val defiPositionsRepository: DefiPositionsRepository,
     private val fiatValueCalculator: DefiFiatValueCalculator,
+    private val snapshotCache: DeFiPositionsSnapshotCache,
     @IoDispatcher private val ioDispatcher: CoroutineDispatcher,
 ) : ViewModel() {
 
@@ -175,6 +176,11 @@ constructor(
     private var loadStakingJob: Job? = null
     private var loadLpJob: Job? = null
 
+    // Pools whose card came from a read that answered. A read that could not resolve them leaves
+    // placeholders standing, and carrying those into the next reload would present a figure nothing
+    // has confirmed while skipping the shimmer that says a read is in flight.
+    private var livePoolKeys: Set<String> = emptySet()
+
     private val currentModel: MayachainDefiPositionsUiModel
         get() =
             (_state.value as? MayachainDefiUiState.Success)?.data ?: MayachainDefiPositionsUiModel()
@@ -190,7 +196,7 @@ constructor(
     fun setData(vaultId: VaultId) {
         this.vaultId = vaultId
         if (_state.value !is MayachainDefiUiState.Success) {
-            _state.value = MayachainDefiUiState.Success(MayachainDefiPositionsUiModel())
+            _state.value = MayachainDefiUiState.Success(restoredModel(vaultId))
         }
         loadBalanceVisibility()
         savedPositionsJob?.cancel()
@@ -201,6 +207,31 @@ constructor(
         observeTotalRawJob = observeTotalRaw()
         currencyJob?.cancel()
         currencyJob = observeCurrencyChanges()
+    }
+
+    /**
+     * The state this vault's screen was last showing, so a re-entry starts from the cards, totals
+     * and enabled set the user left behind rather than from an empty model whose defaults would
+     * flash the CACAO positions back on. Everything in it is refreshed by the loads [setData] kicks
+     * off straight after; only the position picker is dropped, because a sheet the user closed by
+     * leaving the screen must not reopen itself.
+     */
+    private fun restoredModel(vaultId: VaultId): MayachainDefiPositionsUiModel {
+        val cached =
+            snapshotCache.read(vaultId, MayachainDefiPositionsUiModel::class)
+                ?: return MayachainDefiPositionsUiModel()
+        return cached.copy(
+            showPositionSelectionDialog = false,
+            tempSelectedPositions = cached.selectedPositions,
+        )
+    }
+
+    override fun onCleared() {
+        val model = (_state.value as? MayachainDefiUiState.Success)?.data
+        if (::vaultId.isInitialized && model != null) {
+            snapshotCache.write(vaultId, model)
+        }
+        super.onCleared()
     }
 
     /**
@@ -229,6 +260,7 @@ constructor(
                 .drop(1)
                 .collect {
                     resetTotalsToPending()
+                    dropFiatPricedInPreviousCurrency()
 
                     bondedNodesRefreshTrigger.value++
                     loadBondedJob?.cancel()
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -1174,9 +1253,20 @@ constructor(
                 val zero = zeroFiat()
                 // Show placeholder cards for each selected pool first — even pools where the user
                 // has no liquidity yet should be visible so the Add button is reachable.
-                val placeholders = selectedPools.map { it.toPlaceholderUiModel(zero) }
+                // Pools the last load already priced keep their card; only a pool with nothing
+                // behind it yet falls back to the placeholder, and the shimmer is raised only for
+                // those — a refresh over cards that already carry figures leaves them readable.
+                val loaded =
+                    state.value.lp.positions
+                        .filter { it.positionKey in livePoolKeys }
```
johnnyluo: [P1] Restore live LP keys with the cached THORChain snapshot

The cached state brings back the LP cards, but `livePoolKeys` is not restored. When `reloadLpTab()` runs after the dialog data arrives, this filter treats every restored LP row as not live, replaces it with a zero placeholder, and raises loading until the network returns. Re-entry therefore still blanks real LP balances for THORChain, which is exactly the flash this PR is trying to remove. Save and restore the live pool-key set alongside the snapshot, or derive it from the restored known-good LP rows.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
```diff
@@ -683,23 +752,32 @@ constructor(
                 // freezes these exact objects into the terminal state, so a placeholder that
                 // snapshotted an unresolved zero would strand the card on the dash for good.
                 val zero = zeroFiat()
+                // Pools the last load already priced keep their card; only a pool with nothing
+                // behind it yet falls back to the placeholder, and the shimmer is raised only for
+                // those — a refresh over cards that already carry figures leaves them readable.
+                val loaded =
+                    currentModel.lp.positions
+                        .filter { it.positionKey in livePoolKeys }
```
johnnyluo: [P1] Restore live LP keys with the cached Maya snapshot

The snapshot restores the rendered LP cards, but `livePoolKeys` stays at its fresh-ViewModel default of `emptySet()`. On the first reload after re-entry, this filter rejects every restored LP card, so the screen immediately swaps the cached balances for zero placeholders until the network request completes. That breaks the purpose of the snapshot cache for Maya LP positions. Persist the resolved live pool key set with the cached snapshot, or rehydrate `livePoolKeys` from the restored LP positions that were known-good when saved.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
```diff
@@ -440,6 +477,31 @@ constructor(
...
+     */
+    private fun dropFiatPricedInPreviousCurrency() {
+        state.update { current ->
+            current.copy(
+                bonded = current.bonded.copy(totalBondedPrice = null),
+                staking =
+                    current.staking.copy(
+                        positions =
+                            current.staking.positions.map { position ->
+                                position.copy(isLoading = true)
```
johnnyluo: [P2] Clear stale staking fiat when the currency changes

This marks each staking card as loading for the new currency, but leaves the old `stakedFiatDisplay` attached. If the refresh then fails, `settleStakingPositions()` clears `isLoading` and keeps any non-null fiat string, so the card can render a value priced in the previous currency under the newly selected currency. Clear `stakedFiatDisplay` here along with setting `isLoading = true`, so a failed reprice falls back to the new zero/unavailable state instead of preserving stale fiat.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Fixed in `abe837fa1`. Confirmed exactly as you describe: `livePoolKeys` was a view-model field, so a re-entry restored `lp.positions` from the snapshot with the set at `emptySet()`, this filter matched nothing, every restored card fell to `toPlaceholderUiModel`, and `isCold` came out true.

Rather than persisting it separately I moved the set onto `LpTabUiModel` next to the cards it judges, so it travels with the snapshot by construction and cannot fall out of step with them. Deriving it from the restored rows was the other option, but a successfully-read pool holding nothing renders as the same zero placeholder a failed read leaves behind, so that would have promoted unconfirmed cards to settled ones.

Regression test: `a re-entry leaves the cached LP cards up instead of shimmering them` — loads a priced BTC pool, pops the view-model through a real `ViewModelStore`, then re-enters with the LP read held open. It fails on the old code with `isLoading == true`.

app/src/main/java/com/vultisig/wallet/ui/models/defi/MayachainDefiPositionsViewModel.kt
aminsato: Fixed in `abe837fa1`, same change as the THORChain thread: `livePoolKeys` now lives on `LpTabUiModel`, so the snapshot carries it with the cards.

One Maya-specific detail worth flagging — the reload built a fresh `LpTabUiModel(...)` for the placeholder pass and for the two abandoned loads under it (no CACAO coin, address fails validation). A fresh model would have reset the set mid-load, so a load that never lands would drop the standing cards on the *next* reload — something the private field did not do. Those three are now `it.lp.copy(...)`, which preserves the previous behaviour.

Regression test: `a re-entry leaves the cached LP cards up instead of zeroing them`, which fails on the old code.

app/src/main/java/com/vultisig/wallet/ui/models/defi/ThorchainDefiPositionsViewModel.kt
aminsato: Changed in `abe837fa1` — `stakedFiatDisplay = null` now goes in alongside `isLoading = true`, which is what the KDoc on that function already claimed happened.

To be straight about it though: I could not reach the stale render you describe on the current code. `loadStakingPositions()` runs on every currency switch, and at :872 it rebuilds the whole list from `loadDefaultStakingPositions()`. `settled` is keyed off `filterNot { it.isLoading }`, and the drop has just set `isLoading = true` on every card, so `settled` is empty and each card is rebuilt as `copy(isLoading = true, stakedFiatDisplay = zero)` with `zero` resolved in the *new* currency. Every `settleStakingPositions` call site sits after that rebuild, so the `?: zero` there is preserving the new-currency zero, not the old string. The window before the rebuild is covered too — `StakingTabComponents` renders the shimmer instead of the fiat while `isLoading` is true.

So I have taken the change as defence-in-depth and for the coherence of the state (the drop now really does drop), not as a live bug fix. Happy to be shown the path I missed if you have one.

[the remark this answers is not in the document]
aminsato: Changed in `abe837fa1` — the drop now clears `stakedFiatDisplay` alongside raising `isLoading`.

For the record, the failure path as described does not reproduce: `loadStakingPositions()` rebuilds every card at :872 with a zero resolved in the new currency before any `settleStakingPositions` call runs, so the `?: zero` there was never preserving an old-currency string. Taken as a state-coherence change rather than a bug fix. Details in the maintainer thread on the same line.

Also addressed the nitpick on `KaminoEarnViewModelTest`: `a restored total survives the next load rather than being dropped on sight` did duplicate the test above it and never ran a load. It now emits the selection the snapshot was summed over and holds `getUserPositions`, so the load actually reaches its coverage/currency drop check. Verified it fails when `totalCoverage` is not restored.

johnnyluo (approved):
LGTM

PR #5798, opened 2026-09-01
fix(defi): make the Kamino vault picker the shared selection sheet

### LINKED ISSUE #5781: [Fix] Kamino vault picker is a one-off sheet — use the shared token/chain selection [DeFi]
**Screen:** Solana DeFi → Earn → manage Kamino vaults (the pencil)

**Rule for every picker on this app:** use the shared token/chain selection sheet. No one-off sheets. Token select, chain select, THOR/Maya position select, and Kamino vault select are the same chrome.

**Today:** `KaminoVaultPickerSheet` (`KaminoEarnComponents.kt:61`) is a Material `ModalBottomSheet` with unique card rows (name + curator + risk label + Material `Checkbox`) and a `"Save Changes"` button. No search, no grid cells, not `TokenSelectionList`.

**Wanted:** the same sheet as everywhere else — `TokenSelectionList` / `PositionsSelectionDialog` (`DeFiComponents.kt:338`): title, subtitle, search, grid cells with the shared check, Done / Cancel. Feed the allow-list (`KaminoVaultRegistry.ALLOW_LIST`) as cells (vault name + underlying token logo). Persistence can stay on `KaminoPosition.isEnabled`.

iOS already does this: Kamino vaults sit in `AssetSelectionContainerSheet` with `AssetSelectionGridCell` (`DefiChainSelectPositionsScreen.swift`). Android should not invent a second picker.

Kamino one-off picker

aminsato:
Closes #5781

## Summary

Manage positions on **Solana → Earn** opened a picker of its own: a Material `ModalBottomSheet` with card rows carrying the curator and risk tier, a Material `Checkbox`, no search, and a "Save Changes" button. Every other picker in the app — token select, chain select, THOR/Maya positions — is `TokenSelectionList`: title, subtitle, search, 74dp grid cells with the shared check, Done / Cancel.

iOS has no Kamino picker either. `DefiChainSelectPositionsScreen` puts the vaults in the same `AssetSelectionContainerSheet` the coin positions use, as an `.earn` `AssetSection` **inserted at index 0** so the sheet's order matches the chain screen's tab row. So the fix is a new `earnPositions` parameter on `PositionsSelectionDialog`, rendered as the first group — not a second composable. `KaminoVaultPickerSheet` is gone.

## Before / After

| Before | After |
|--------|-------|
| before | after |

## Details

**The cell is the vault's pinned name over its underlying token's mark**, mirroring iOS `KaminoVaultSelectionGridCell` — the vault is what is picked, the token is what the user recognises. The pinned `fallbackName` rather than the live name the Earn card shows: the picker is a catalogue of what the app offers, and the catalogue is `KaminoVaultRegistry.ALLOW_LIST`. `positionKey` is the vault address, because a ticker would alias both USDC vaults onto one selection.

**Search had to widen.** iOS's `DefiSelectableAsset.searchTerms` is `[fallbackName, curator]`, while the dialog filtered on `ticker` alone — so "rockawayx" would have found nothing, since "RWA USDC" carries no trace of RockawayX in its name. `PositionUiModelDialog` gains `searchTerms`, defaulting to `listOf(ticker)`, which leaves every existing coin section matching exactly as before. The rule moves into a free `matching(query)` so it can be tested without a composition.

**The curator subtitle and risk chip deliberately do not come across.** The shared 74dp cell has no room for them and iOS's cell shows neither. They stay on the Earn card, which is untouched. That orphans `kamino_earn_title` and `kamino_earn_curated_by`, removed from all ten locale files.

**Persistence is unchanged** — the same pending set, saved by the same `KaminoVaultSelectionRepository` on Done, dismissed without saving on Cancel.

The dialog's four near-identical `remember` pairs collapse into one `rememberFilteredPositions`, so adding a fourth section shortened the function rather than growing it.

## Test plan

- [x] `./gradlew :app:testDebugUnitTest` — 2309 tests, 0 failures. 12 new across `KaminoEarnPickerPositionsTest` and `PositionUiModelDialogTest`: every allow-listed vault is offered in registry order, cells are keyed by address (not ticker, which would alias the two USDC vaults), searching a curator finds the vault it runs, and the default `searchTerms` leaves coin sections matching by ticker exactly as before.
- [x] `./gradlew :app:lintDebug`
- [x] `./gradlew :app:compileDebugAndroidTestKotlin`
- [x] `./gradlew :app:ktfmtFormat`
- [x] Rendered on device via `PreviewActivity` (`-e screen solana_defi_picker`, added in this PR) for the screenshots above.

**In-app path:** Home → Solana → DeFi → Earn tab → the pencil (Manage positions). The sheet should be the same one Bond/Stake/LP open on THORChain, with an "Earn" section. Toggle a vault, press the tick, and the Earn tab reloads with that vault's card; press the cross instead and the enabled set is untouched. Typing `rockaway` in the search should leave only RWA USDC.

## Summary by CodeRabbit

- **New Features**
  - Added a shared position-selection dialog across Earn, Bond, Stake, and Liquidity Pool views.
  - Added case-insensitive search for position tickers, names, and curator details.
  - Improved Solana Earn vault selection with curated labels, logos, ordering, and support for duplicate tokens.

- **Tests**
  - Added coverage for position searching, filtering, selection behavior, and vault display details.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -50,6 +50,7 @@ import com.vultisig.wallet.data.blockchain.cosmos.qbtc.claim.QbtcClaimError
 import com.vultisig.wallet.data.blockchain.cosmos.staking.CosmosStakePositionRow
 import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoCurator
 import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoRiskTier
+import com.vultisig.wallet.data.blockchain.solana.kamino.KaminoVaultRegistry
 import com.vultisig.wallet.data.blockchain.solana.staking.SolanaStakeState
 import com.vultisig.wallet.data.models.Account
 import com.vultisig.wallet.data.models.Address
@@ -226,8 +227,10 @@ import com.vultisig.wallet.ui.screens.transaction.toUiTransactionInfo
 import com.vultisig.wallet.ui.screens.v2.chaintokens.ChainTokensScreen
 import com.vultisig.wallet.ui.screens.v2.defi.DeFiTab
 import com.vultisig.wallet.ui.screens.v2.defi.HeaderDeFiWidget
+import com.vultisig.wallet.ui.screens.v2.defi.PositionsSelectionDialog
 import com.vultisig.wallet.ui.screens.v2.defi.maya.RemoveLpScreenContent
 import com.vultisig.wallet.ui.screens.v2.defi.model.DeFiNavActions
+import com.vultisig.wallet.ui.screens.v2.defi.solana.KAMINO_EARN_PICKER_POSITIONS
 import com.vultisig.wallet.ui.screens.v2.defi.solana.KaminoAmountContent
 import com.vultisig.wallet.ui.screens.v2.defi.solana.KaminoEarnRow
 import com.vultisig.wallet.ui.screens.v2.defi.solana.KaminoEarnUiModel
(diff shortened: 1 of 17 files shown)


johnnyluo (approved):
LGTM

PR #5799, opened 2026-09-01
fix(defi): count positions on the DeFi home chain rows

### LINKED ISSUE #5780: [Fix] DeFi home chain rows say "N Assets" — should count active positions [DeFi]
**Screen:** DeFi home chain list

**Bug:** each chain row's subtitle is the Wallet copy — `"N Assets"` (token inventory) or, when the chain has only one coin, the native amount (`3.831213 TRX`). That is not how many DeFi positions the vault holds on that chain.

**Wanted:** count **active positions** (amount > 0) and label them positions. Match iOS `DefiChainCellView`:

- count > 0 → `"%d positions"` (`defiChainPositionsCount`)
- count == 0 → `"No positions found"` (`defiChainNoPositions`)

Do **not** change Wallet rows — those keep `"N Assets"`.

**Today:**
- DeFi home reuses `AccountItem` (`VaultAccountsScreen.kt:350`, `showAddress = false`).
- Subtitle is `vault_accounts_account_assets` when `assetsSize > 1` (`AccountItem.kt:92-96`).
- `assetsSize = from.accounts.size` (`AddressToUiModelMapper.kt:37`) — wallet coins on the chain, not DeFi positions.

Screenshot: THORChain `"12 Assets"`, Solana `"16 Assets"`, Tron native amount. Those are token counts.

**Count (iOS already has this):** `DefiBalanceService.defiPositionCount` — enabled bond / stake / LP / staking / Kamino rows with amount > 0. Comment there points at Windows `positionsWithBalanceCount` (`useDefiChainPortfolios`). Port that onto the DeFi home mapper; do not reuse `accounts.size`.

DeFi home showing Assets instead of positions

FYI (not this ticket): iOS `vultisig-ios` `DefiChainCellView.swift:48-53`.

aminsato:
Closes #5780

## Problem

Every DeFi chain row was painting the Wallet subtitle: the vault's token count on that chain (`"12 Assets"`), or the native amount when the chain held a single coin (`3.831213 TRX`). Neither reports the positions the row exists for — a vault tracking twelve THORChain tokens while staking two of them read "12 Assets".

`assetsSize` is `from.accounts.size` (`AddressToUiModelMapper.kt:37`), and `AccountItem` chose between it and the native amount with no notion of a DeFi row at all.

## Fix

The count comes off the row's own accounts. A DeFi address carries one account per token the chain can hold a position in — the vault's coins plus the injected `defiOnly` / position-only receipts — and each account's token value *is* the position it backs, so the accounts above zero are the active positions. That makes the subtitle the count behind the very figure printed above it, which sums the same list, rather than a second reading that could disagree with it.

- `calculateActiveDefiPositionsCount()` on `List<Account>`, next to the existing portfolio folds.
- `AddressToUiModelMapper` sets it for DeFi rows only, and finally populates the previously-dead `isDeFiProvider` flag on `AccountUiModel`.
- `AccountItem` renders `"N positions"` / `"No positions found"` for a DeFi row. Wallet rows are untouched: still `"N Assets"`, still the native amount for single-token chains.

**Unresolved is not zero.** The fold answers `null` until something has resolved, so a cold start holds the same placeholder the native amount already used instead of claiming a funded chain has no positions and then correcting itself.

**Pluralised**, unlike iOS and Windows, which both format a flat `"%d positions"`. Tron, TON and the Cosmos staking chains are 0-or-1 by construction, so the flat form would ship "1 positions" as the common case rather than the edge one — visible as `Circle · 1 position` below.

## Strings

`no_positions_found` already existed in all ten locales and already matched iOS, so it is reused. `defi_chain_positions_count` is new in all ten, using each locale's established term for "position" — including `仓位` for zh-rCN, which is what the rest of this app's Chinese uses (iOS says `持仓`), and the full `one/few/many` sets for ru and hr.

## Two divergences from iOS, deliberate

- **The opt-in.** `DefiBalanceService.defiPositionCount` gates on the Manage Positions selection. Nothing in the Android home balance path reads that selection, and the row's fiat total does not either — gating the count alone would have made the subtitle disagree with the figure above it. Solana is the exception and already behaves: `KaminoDeFiBalanceService` gates on its own vault selection, so toggling a Kamino vault off does drop the count.
- **Bond granularity.** A vault bonded to several nodes counts as **1** here, because `ThorchainDeFiBalanceService` sums bonds into a single RUNE figure (and `MayaDeFiBalanceService` merges bond + stake into one CACAO figure) long before the row sees them. iOS counts each node. Matching it would need a parallel per-position counter reading six-plus repositories on the home screen, which would then disagree with the row's own total.

## Test plan

- `calculateActiveDefiPositionsCount`: funded-only counting, all-empty, nothing-resolved, one resolved among pending, empty address. (5 new)
- `AddressToUiModelMapper`: a DeFi row reports positions rather than its token count, reports zero, reports unknown while resolving; a wallet row carries no count. (4 new)
- Full suites green: `:data` + `:app` — 5798 tests, 0 failures.
- `:app:lintDebug` passes with no warning on the new resource (`many` included for es/it/pt, which lint asks for).
- ktfmt applied.

## Result

DeFi tab, real device — Solana and THORChain plural, Circle singular:

DeFi home rows counting positions

## Summary by CodeRabbit

* **New Features**
  * DeFi provider rows now display the number of active positions.
  * Counts support loading, zero-position, singular, and plural states.
  * Position counts are localized across supported languages.
  * Counts now reflect positions aggregated across supported DeFi balances.

* **Bug Fixes**
  * Wallet rows continue showing existing token and asset counts without DeFi position labels.

* **Tests**
  * Added coverage for active, empty, unresolved, and partially resolved DeFi positions.

### DIFF
--- a/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/VaultAccountsViewModel.kt
@@ -143,6 +143,11 @@ internal data class AccountUiModel(
     val assetsSize: Int = 0,
     val nativeTokenTicker: String = "",
     val isDeFiProvider: Boolean = false,
+    /**
+     * Active DeFi positions on the chain, or null while none of them has resolved yet. Only ever
+     * set on a DeFi row; a wallet row keeps reporting its token count instead.
+     */
+    val defiPositionsCount: Int? = null,
 )
 
 @HiltViewModel
--- a/app/src/main/java/com/vultisig/wallet/ui/models/mappers/AddressToUiModelMapper.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/models/mappers/AddressToUiModelMapper.kt
@@ -3,6 +3,7 @@ package com.vultisig.wallet.ui.models.mappers
 import com.vultisig.wallet.data.mappers.SuspendMapperFunc
 import com.vultisig.wallet.data.models.Address
 import com.vultisig.wallet.data.models.calculateAccountsPartialFiatValue
+import com.vultisig.wallet.data.models.calculateActiveDefiPositionsCount
 import com.vultisig.wallet.data.models.logo
 import com.vultisig.wallet.ui.models.AccountUiModel
 import javax.inject.Inject
@@ -36,6 +37,12 @@ constructor(
                 },
             assetsSize = from.accounts.size,
             nativeTokenTicker = nativeAccount.token.ticker,
+            isDeFiProvider = isDefiProvider,
+            // A DeFi row reports how many positions the vault holds on the chain, not how many
+            // tokens it tracks there. Counted only for DeFi rows, so a wallet row still reports
+            // its asset count.
+            defiPositionsCount =
+                if (isDefiProvider) from.accounts.calculateActiveDefiPositionsCount() else null,
         )
     }
 }
--- a/app/src/main/java/com/vultisig/wallet/ui/screens/v2/home/components/AccountItem.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/screens/v2/home/components/AccountItem.kt
@@ -16,6 +16,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.painterResource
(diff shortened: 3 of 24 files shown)


johnnyluo (approved):
LGTM

PR #5805, opened 2026-09-02
Fix sTCY DeFi portfolio balance

### LINKED ISSUE #5764: [Fix] Auto-compounded TCY (sTCY) is missing from the DeFi portfolio total and the THORChain row [DeFi]
# [Fix] Auto-compounded TCY (sTCY) is missing from the DeFi portfolio total and the THORChain row [DeFi]

## Problem

A TCY position staked with auto-compounding is fetched, priced and rendered on the THORChain DeFi screen, but is filtered out of the numbers that feed the vault home DeFi tab. The two screens therefore disagree by the whole position.

`DefaultStakingPositionService.kt:28-34` fetches it — `supportedStakingCoins` lists `sTCY`, `yRUNE`, `yTCY`, `ybRUNE`, and `getStakingDetailsFromNetwork` reads each denom's bank balance (sTCY's denom is `x/staking-tcy`, `Coins.kt:3904-3915`).

But `ThorchainDeFiBalanceService` then drops it in **both** balance paths, with the same three-way predicate that never mentions sTCY:

- remote path, `:313-318` — `it.coin.id.equals(yTCY.id) || equals(yRUNE.id) || equals(ybRUNE.id)`
- cached path, `:169-174` — identical predicate

sRUJI gets its own explicit branch at `:102-107` and `:137-150`. sTCY gets nothing.

That filtered list is what the portfolio is built from: `BalanceRepository.kt:254-266` and `:370-388` (an unmatched coin returns `TokenValue(value = BigInteger.ZERO)`), `AccountsRepository.kt:425-544` (unmatched accounts are explicitly zero-filled at `:521-535`), summed into `totalDeFiValue` at `VaultAccountsViewModel.kt:783-801`.

The exclusion is doubled: `Coins.kt:4283` defines `defiOnly = listOf(ThorChain.sRUJI, ThorChain.ybRUNE)` with sTCY absent, so `defiOnlyTokens` (`AccountsRepository.kt:437`, `:755-762`) never injects it either. Discovery cannot rescue it — `TokenRepository.kt:99-110` drops any THORChain denom outside `enabledDenoms` unless it is in `AUTO_DISCOVER_THORCHAIN_DENOMS`, which is `setOf(bRUNE.contractAddress)` only (`:297`).

Meanwhile the position **is** rendered one screen deeper: `ThorchainDefiPositionsViewModel.kt:981-1109` builds the sTCY card and folds it into that screen's own header total (`:1099-1108` -> `:299-317`), and `DeFiExtensions.kt:32-41` includes sTCY with `:59-60` selecting it by default. So no opt-in is required to see the contradiction — the user does not have to touch Manage Positions.

Nothing indicates this is deliberate. The filter predates sTCY's DeFi wiring; sTCY was added to `DefaultStakingPositionService` by the sTCY display/interaction work, which touched only the positions screen and that service. When ybRUNE was added later, merged PR #5638 deliberately added it to **both** filters. sTCY was simply left behind.

This is the same defect the team already accepted and fixed once: issue #5614 ("DeFi Portfolio reads $0.00 for an auto-compounding RUJI position") was closed by merged PR #5615, and its body generalises the cause — "any DeFi position keyed to a coin the vault does not also carry as a wallet token is dropped the same way".

The money is not at risk and the position stays redeemable from the THORChain DeFi screen. The harm is a silently understated net-worth figure and a visible self-contradiction between two screens in the same app.

## Expected Behavior

An auto-compounded TCY position counts toward the THORChain row's fiat value and the Total DeFi Value on the vault home DeFi tab, matching the figure the THORChain DeFi screen already shows for the same position.

## Steps to Reproduce

1. Open a vault holding THORChain TCY, with no bond, no plain TCY stake and no yRUNE/yTCY/ybRUNE position.
2. DeFi tab -> THORChain -> Stake TCY. Enable the auto-compounding toggle and stake, for example, 1,000 TCY. The vault receives ~1,000 sTCY as the `x/staking-tcy` bank denom.
3. Return to the vault home DeFi tab. The THORChain row and the Total DeFi Value header both read `$0.00`.
4. Tap the THORChain row. The Staked tab shows the compounded TCY card at its correct fiat value, and that screen's own header total includes it.

Two totals in the same app for the same chain disagree by the entire position.

## Solution

Add sTCY to both filters in `ThorchainDeFiBalanceService` — the cached path at `:169-174` and the remote path at `:313-318` — exactly as merged PR #5638 did for ybRUNE, so the position reaches the chain row and the portfolio total as well as its own card. This is a two-line change to an existing predicate and needs no new fetch: `DefaultStakingPositionService` already returns the balance.

One thing the implementer should check rather than assume: unlike sRUJI and ybRUNE, sTCY is currently also a wallet token (`Coins.kt:206` includes it in `ThorChain.all`). Confirm that counting it in the DeFi total does not also make it contribute to the wallet-side total for the same vault. Whether sTCY should stop being a wallet token altogether is a separate product question already raised in #3027 and is out of scope here.

### Must Do
- [ ] Add `Coins.ThorChain.sTCY` to the staking-details filter in `ThorchainDeFiBalanceService.kt:313-318` (remote path).
- [ ] Add it to the identical filter at `ThorchainDeFiBalanceService.kt:169-174` (cached path) — both must change or the value flickers between a cold and warm read.
- [ ] Verify the sTCY balance is not counted twice for a vault that shows it on both the Wallet and DeFi surfaces.
- [ ] Add a unit test asserting an sTCY staking position appears in the balances returned by both the cached and remote DeFi-balance paths.

### Must NOT Do
- Don't change unrelated code
- Don't move sTCY into `Coins.defiOnly` or remove it from `ThorChain.all` — that changes wallet-tab behaviour and is a separate product decision
- Don't change how `DefaultStakingPositionService` fetches positions; it already returns sTCY correctly
- Don't touch the yRUNE / yTCY / ybRUNE / sRUJI handling

## Acceptance Criteria
- [ ] A vault whose only THORChain DeFi position is auto-compounded TCY shows a non-zero THORChain row and a non-zero Total DeFi Value on the vault home DeFi tab
- [ ] That figure equals the value shown for the same position on the THORChain DeFi screen's own header
- [ ] The value is correct on both a cold start (remote path) and a warm start (cached path)
- [ ] Vaults with yRUNE, yTCY, ybRUNE or sRUJI positions show unchanged totals

johnnyluo:
## Summary
- Include sTCY in THORChain default staking DeFi balances for cached reads
- Include sTCY in the matching remote balance path
- Add regression coverage for fetched and cached sTCY positions

Closes #5764

## Test plan
- ./gradlew :data:testDebugUnitTest --tests com.vultisig.wallet.data.blockchain.thorchain.ThorchainDeFiBalanceServiceYbRuneTest
- ./gradlew :data:testDebugUnitTest
- ./gradlew assembleDebug

## Summary by CodeRabbit

* **New Features**
  * sTCY staking balances are now included in DeFi balance displays for both newly fetched and cached data.
  * DeFi-only sTCY positions are now included in account balances and price refreshes, even when the receipt is not held in the wallet.

* **Bug Fixes**
  * Corrected staking balance retrieval so eligible sTCY positions are reflected consistently.
  * Preserved fiat pricing for accounts containing sTCY DeFi positions.

### DIFF
--- a/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/thorchain/ThorchainDeFiBalanceService.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/blockchain/thorchain/ThorchainDeFiBalanceService.kt
@@ -168,7 +168,8 @@ class ThorchainDeFiBalanceService(
                         balances =
                             defaultDetails
                                 .filter {
-                                    it.coin.id.equals(Coins.ThorChain.yRUNE.id, true) ||
+                                    it.coin.id.equals(Coins.ThorChain.sTCY.id, true) ||
+                                        it.coin.id.equals(Coins.ThorChain.yRUNE.id, true) ||
                                         it.coin.id.equals(Coins.ThorChain.yTCY.id, true) ||
                                         it.coin.id.equals(Coins.ThorChain.ybRUNE.id, true)
                                 }
@@ -312,7 +313,8 @@ class ThorchainDeFiBalanceService(
 
                     stakingDetails
                         .filter {
-                            it.coin.id.equals(Coins.ThorChain.yTCY.id, true) ||
+                            it.coin.id.equals(Coins.ThorChain.sTCY.id, true) ||
+                                it.coin.id.equals(Coins.ThorChain.yTCY.id, true) ||
                                 it.coin.id.equals(Coins.ThorChain.yRUNE.id, true) ||
                                 it.coin.id.equals(Coins.ThorChain.ybRUNE.id, true)
                         }
--- a/data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
+++ b/data/src/main/kotlin/com/vultisig/wallet/data/repositories/AccountsRepository.kt
@@ -98,13 +98,13 @@ constructor(
         buildCacheAddresses(vaultId).flatMapLatest { (vaultCoins, addresses) ->
             channelFlow {
                 supervisorScope {
-                    // The DeFi-only receipts ride along: no vault carries one, so this refresh —
-                    // the only ungated one — is all that keeps the row they are valued from warm.
-                    // loadDeFiAddresses refreshes them too, but only on an explicit reload, and
-                    // its cached emission reads that row, so a cold start would otherwise value a
-                    // funded position at $0.00.
+                    // Position receipts ride along: no vault necessarily carries one, so this
+                    // refresh — the only ungated one — is all that keeps the row they are valued
+                    // from warm. loadDeFiAddresses refreshes them too, but only on an explicit
+                    // reload, and its cached emission reads that row, so a cold start would
+                    // otherwise value a funded position at $0.00.
                     val loadPrices = async {
-                        tokenPriceRepository.refresh(vaultCoins + defiOnlyReceiptsFor(vaultCoins))
+                        tokenPriceRepository.refresh(vaultCoins + injectedReceiptsFor(vaultCoins))
                     }
 
                     // Always emit the last-known DB snapshot first so the UI shows cached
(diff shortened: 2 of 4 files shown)


data/src/main/kotlin/com/vultisig/wallet/data/blockchain/thorchain/ThorchainDeFiBalanceService.kt
```diff
@@ -168,7 +168,8 @@ class ThorchainDeFiBalanceService(
                         balances =
                             defaultDetails
                                 .filter {
-                                    it.coin.id.equals(Coins.ThorChain.yRUNE.id, true) ||
+                                    it.coin.id.equals(Coins.ThorChain.sTCY.id, true) ||
```
aminsato, line 171: The balance is now produced, but for the vaults that most need it it still never reaches the Portfolio DeFi total.

`AccountsRepository.loadDeFiAddresses` builds the DeFi accounts from `vault.coins` + `defiOnlyTokens(chain)` + `positionOnlyTokens(chain)`. `Coins.defiOnly` is only `[sRUJI, ybRUNE]`, and `positionOnlyTokens` returns empty for anything but Solana — so an sTCY account exists only when the vault already carries sTCY as a wallet coin. Both consumption paths then drop a balance with no account: the cached path matches `balancesByTicker` into existing accounts, and the network path iterates accounts and calls `getDefiTokenBalanceAndPrice` per account token.

A vault gets sTCY into `vault.coins` only by manually enabling it: `TokenRepository.getRefreshTokens` seeds `enabledDenoms` from `vault.coins` contract addresses, and `x/staking-tcy` is not in `AUTO_DISCOVER_THORCHAIN_DENOMS`, so discovery drops it; nothing in the STAKE_STCY deposit flow calls `addTokenToVault` either.

Concrete scenario: a vault stakes TCY with auto-compound on, receives sTCY, and never enables sTCY in the token picker. `getRemoteDeFiBalance`/`getCacheDeFiBalance` now return the sTCY position, but `loadDeFiAddresses` has no sTCY account to hang it on, so the THORChain DeFi row and the DeFi header total read the same as before this change. (The dedicated positions screen still shows the card — it reads `defaultStakingPositionService` directly — which makes the portfolio/screen mismatch the visible symptom.)

Adding sTCY to `Coins.defiOnly` is not right either (it is a spendable bank denom and `spendableAccounts()` filters `isDefiOnly` out, which would break the sTCY unstake/transfer form). The plumbing likely needs sTCY injected the way `positionOnlyTokens` does it, or added to `AUTO_DISCOVER_THORCHAIN_DENOMS` so a holder discovers it.

johnnyluo: Fixed in 0509237b2. I added sTCY as a THORChain position-only injection for `loadDeFiAddresses`, so the portfolio row has an account to attach cached and refreshed sTCY DeFi balances to even when the vault has not enabled sTCY as a wallet token. I left `Coins.defiOnly` and token discovery unchanged, so an already-enabled sTCY wallet token remains spendable. Added repository tests for cached and refreshed DeFi rows without wallet sTCY, plus kept the balance-service coverage. Verified `:data:testDebugUnitTest` and `assembleDebug` locally.

PR #5808, opened 2026-09-02
feat(swap): add a custom token from the asset search empty state

### LINKED ISSUE #5786: [Feature] Swap token search empty state: + Add custom token [Swap]
**Screen:** Swap → Select asset (search with no matches)
**Figma:** https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=81853-75673

When the search returns nothing, show the empty card **and** a `+ Add custom token` action under it. Tapping it opens the existing custom-token flow for the chain currently selected in the carousel. After the token is added, it should be available in this picker.

**Figma copy:**
- Empty card: `No token found` (dashed-circle icon)
- CTA: `+ Add custom token`

**Today:** `SelectAssetScreen.kt:87-90` only renders `NoFoundContent` (`select_asset_no_result` = "No result found."). No CTA. `CustomTokenScreen` / `Route.CustomToken` already exists for manage-tokens — reuse it, don't invent a second adder.

Show the CTA only on chains that already support custom tokens.

Figma

aminsato:
Closes #5786

Figma: [Swap Screen - add custom token](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=81853-75673)

Select asset — no token found, with + Add custom token

## What

Searching the asset picker for a token the catalog doesn't carry ended at "No result found." — a dead end, even on chains where the app can resolve any contract address. Adding it meant leaving the swap, walking to the chain's manage-tokens list, adding it there, and starting the swap over.

The empty state now shows the Figma card (`No token found`, dashed-circle icon) with **+ Add custom token** under it, routing to the same `Route.CustomToken` sheet manage-tokens uses. No second adder: `CustomTokenViewModel` already responds with the resolved coin and leaves persistence to the caller, so the picker enables it exactly the way `TokenSelectionViewModel` does.

The CTA is gated on `Chain.canSelectTokens` — the same predicate manage-tokens uses — so it stays hidden on zkSync and Cronos.

## Two things the picker had to learn

**`loadAddress` is a one-shot flow.** It reads the vault once per subscription and completes, so a coin written to the vault after collection started never reaches the list — enabling the token alone would have left the empty state exactly as it was. The chain key upstream of `flatMapLatest` now carries a refresh counter that the enable bumps, restarting the read. Anything else that mutates this picker's vault coins will need the same bump.

**The query that found nothing keeps filtering.** After the token is added the search field is set to its ticker, so the user lands on the row they just created instead of the same empty card.

Dismissing the custom-token sheet leaves its `request` suspended for good, so a second tap would have parked two waiters on one rendezvous and enabled the next token twice. One cancellable job at a time; there's a test for it.

## Notes

`NoFoundContent` is shared by nine screens, so the dashed-circle icon arrives as an optional parameter rather than a global swap. Its 15sp message style is left alone even though this frame's card is Title3 — moving it would move the other eight empty states. Happy to change that globally if you'd rather.

iOS has nothing to match here: [ios#5303](https://github.com/vultisig/vultisig-ios/issues/5303) is still open and its custom-token flow lives only under `Features/Wallet/TokenSelection`, never in a swap picker. Figma is the reference.

Two new string keys in all 10 locales, reusing each locale's existing token / custom-token terminology.

## Test plan

- `:app:assembleDebug`, `:app:lintDebug` clean; ktfmt applied.
- `SelectAssetViewModelTest` — 5 tests, 0 skipped, 3 new: the CTA gate follows the carousel chain, `addCustomToken` routes to the selected chain and enables + surfaces the result, and a dismissed sheet leaves no waiter that double-enables the next token.
- Manual, on device: vault with Ethereum enabled → Swap → tap the From token pill → search `jkd` → the card and CTA appear → **+ Add custom token** → paste an uncurated ERC-20 (e.g. `0x95ad61b0a150d79219dcf64e1e6cc01f0b64c4ce`) → Search → Add → back in the picker with the ticker in the search field and the token listed. Swiping the carousel to zkSync hides the CTA.

Verified on device — the screenshot above is a real Android render on the Ethereum chain with a query that matches nothing.

## Summary by CodeRabbit

- **New Features**
  - Added support for adding custom tokens directly from the asset-selection screen when available.
  - Custom tokens are enabled, refreshed, and shown in search results after being added.
  - Improved empty search results with clearer messaging and an add-token action.
  - Added localized text for the new asset-selection states across supported languages.

- **Bug Fixes**
  - Prevented duplicate custom-token requests and preserved the search query after adding a token.

- **Style**
  - Added a dashed circular visual for the empty-state interface.

### DIFF
--- a/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
+++ b/app/src/debug/java/com/vultisig/wallet/debug/PreviewActivity.kt
@@ -2572,6 +2572,7 @@ private fun SelectAssetSecuredPreview(showSecured: Boolean) {
         searchFieldState = TextFieldState(),
         onAssetClick = {},
         onSelectChain = {},
+        onAddCustomTokenClick = {},
     )
 }
 
--- a/app/src/main/java/com/vultisig/wallet/ui/components/v2/tokenitem/NoFoundContent.kt
+++ b/app/src/main/java/com/vultisig/wallet/ui/components/v2/tokenitem/NoFoundContent.kt
@@ -1,5 +1,6 @@
 package com.vultisig.wallet.ui.components.v2.tokenitem
 
+import androidx.annotation.DrawableRes
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
@@ -20,7 +21,10 @@ import com.vultisig.wallet.ui.theme.Theme
 
 @Preview
 @Composable
-internal fun NoFoundContent(message: String = "No chains found") {
+internal fun NoFoundContent(
+    message: String = "No chains found",
+    @DrawableRes icon: Int = R.drawable.iconcrypto,
+) {
     TopShineContainer {
         Column(
             horizontalAlignment = Alignment.Companion.CenterHorizontally,
@@ -28,7 +32,7 @@ internal fun NoFoundContent(message: String = "No chains found") {
             modifier = Modifier.Companion.fillMaxWidth().padding(vertical = 24.dp),
         ) {
             Image(
-                painter = painterResource(R.drawable.iconcrypto),
+                painter = painterResource(icon),
                 contentDescription = null,
                 modifier = Modifier.Companion.size(20.dp),
             )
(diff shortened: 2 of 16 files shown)


johnnyluo (approved):
LGTM

PR #5810, opened 2026-09-02
fix(home): float the bottom navigator over the content

### LINKED ISSUE #5784: [Fix] Bottom nav paints an opaque block — overlay the pill and camera only [Home]
**Screen:** DeFi chain detail (screenshot is Solana Earn) — same chrome on home

**Bug:** the bottom navigator sits on an opaque full-width block. Content (the Kamino card) is clipped by it instead of showing through around the Wallet/DeFi pill and the camera button.

**Wanted:** no bar background. Only the navigator components (pill + camera). Content scrolls behind; add list bottom padding so the last card can clear the pill. Figma Tab Bar is already drawn that way: https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=49057-161811

**Today:** both hosts put `CryptoConnectionSelect` + `CameraButton` in Material3 `Scaffold`'s `bottomBar` slot, which paints the surface color behind the whole slot:

- `ChainDashboardScreen.kt:87-122` — DeFi detail. Also pads content with `calculateBottomPadding()`, so the card cannot draw under the bar at all.
- `ScaffoldWithExpandableTopBar.kt:47-58` — home. `containerColor = backgrounds.primary`.

Overlay the two components on the content (`Box` + `align(BottomCenter)`), keep the slot transparent, do not use Scaffold's bottomBar surface.

opaque nav block

aminsato:
Closes #5784

The bottom navigator sat on an opaque full-width block that clipped the content above it. It floats over the content now, with the pill and camera as the only painted things.

| Before (from the issue) | After (Solana DeFi, on device) |
|---|---|
|  |  |

The after shot is the Solana DeFi detail scrolled to the end: the RWA USDC card runs behind the pill and fades out, and both cards can be scrolled fully clear of it.

## What was actually wrong

Material3's `bottomBar` slot paints nothing — the block was the Scaffold container colour showing through. Both hosts padded their content out of the bar's region (`ChainDashboardScreen`'s `calculateBottomPadding()`, `VaultAccountsScreen`'s `padding(innerPadding)`), so nothing drew there and the card met a hard edge against the bare background.

## The change

A single overlay, `BottomNavigatorOverlay`, now hosts both screens:

- The pill and camera are drawn over the content (`Box` + `align(BottomCenter)`); the `bottomBar` slot is gone from `ChainDashboardScreen`, and `ScaffoldWithExpandableTopBar` loses its `bottomBarContent` parameter with it.
- `LocalBottomNavigatorPadding` publishes the 82dp the navigator covers (66dp pill + 16dp breathing room, matching iOS's 64 + 16) down the tree, so lists take it as **`contentPadding`** rather than layout padding — content draws behind, and the last row still scrolls clear. A CompositionLocal because `ChainDashboardScreen` hosts eight different screens; threading a parameter through all of them would have been pure signature churn.
- The reservation holds whether or not the navigator is showing, so opening the search bar no longer reflows the list.

Applied across all nine hosted screens: home, chain tokens, THORChain/MayaChain/Solana/Circle (scrolling `Column`s) and Tron/TON/Cosmos (`LazyColumn` `contentPadding`).

**One thing worth a look:** the existing `BottomFadeEffect` moved out of home and the token list into the overlay. At the true foot of the screen it now covers every hosted screen instead of two, which is what iOS does (`VultiTabBar.bottomGradient`, an 80pt fade in the same background colour) and what keeps content legible as it passes under an opaque pill. It is not a bar background, but it is more than "only the pill and camera" — easy to drop if you read the ask more strictly.

Out of scope: Android's pill is opaque `#0d2446` where Figma has a translucent, blurred `rgba(19,46,86,0.6)`. Left alone.

## Note on a list that wraps its content

The token list lives inside the `TopShineContainer` card and sizes itself to its content, so `contentPadding` there is measured as part of the card — a chain holding one token rendered a card with a tall band of dead space under its only row. The reservation is taken on the hosting column instead. `ChainTokensListHeightTest` pins it: on the bad version it fails with `the card runs 188.0px past its only row`.

## Test plan

- [x] `./gradlew :app:compileDebugKotlin :app:compileDebugAndroidTestKotlin`
- [x] `./gradlew :app:lintDebug`
- [x] `./gradlew :app:connectedDebugAndroidTest` — 4 new instrumented tests, `failures="0"`:
  - `BottomNavigatorOverlayTest` — content keeps the full height under the navigator, the published padding covers the navigator, hiding it leaves the reservation in place
  - `ChainTokensListHeightTest` — a single-token card hugs its row
- [x] On device: Solana DeFi detail (above), Bitcoin token list (single token, card hugs its row)

## Summary by CodeRabbit

* **New Features**
  * Added a shared bottom navigation overlay with wallet/DeFi selection and camera controls.
  * Added configurable bottom fade sizing and consistent navigation-aware spacing across dashboard, account, staking, DeFi, and token screens.
  * Added improved token-list and bottom-navigation layout handling to keep content visible above the navigator.

* **Tests**
  * Added UI coverage for navigator spacing, visibility behavior, full-height content, and single-token card layout.

### DIFF
--- a/app/src/androidTest/java/com/vultisig/wallet/ui/screens/v2/chaintokens/ChainTokensListHeightTest.kt
+++ b/app/src/androidTest/java/com/vultisig/wallet/ui/screens/v2/chaintokens/ChainTokensListHeightTest.kt
@@ -1,0 +1,84 @@
+package com.vultisig.wallet.ui.screens.v2.chaintokens
+
+import androidx.compose.ui.test.junit4.v2.createComposeRule
+import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.unit.dp
+import com.vultisig.wallet.data.models.CryptoConnectionType
+import com.vultisig.wallet.ui.models.ChainTokenUiModel
+import com.vultisig.wallet.ui.models.ChainTokensUiModel
+import com.vultisig.wallet.ui.screens.v2.home.components.BottomNavigatorOverlay
+import com.vultisig.wallet.ui.theme.OnBoardingComposeTheme
+import org.junit.Assert.assertTrue
+import org.junit.Rule
+import org.junit.Test
+
+class ChainTokensListHeightTest {
+
+    @get:Rule val compose = createComposeRule()
+
+    /**
+     * The token list wraps its content, so the room reserved for the floating navigator has to be
+     * taken on the column hosting the card and never as padding inside the list — inside, a chain
+     * holding one token renders a card with a tall band of dead space under its only row.
+     */
+    @Test
+    fun aSingleTokenCardHugsItsRow() {
+        compose.setContent {
+            OnBoardingComposeTheme {
+                BottomNavigatorOverlay(
+                    isNavigatorVisible = true,
+                    activeType = CryptoConnectionType.Wallet,
+                    onTypeClick = {},
+                    onCameraClick = {},
+                ) {
+                    ChainTokensScreen(
+                        uiModel =
+                            ChainTokensUiModel(
+                                chainName = "Bitcoin",
+                                chainAddress = "bc1ql2v0000000000000000000000000j4g60yc",
+                                totalBalance = "$15.16",
+                                tokens =
+                                    listOf(
+                                        ChainTokenUiModel(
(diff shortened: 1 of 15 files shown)


johnnyluo (approved):
LGTM

Issue #5765, opened 2026-08-30
[Fix] THORChain referral bps: the row shows the referrer's cut as the user's saving, and at the top VULT tier the code silently costs 10 bps [Swap]


rkokhatskyi:
# [Fix] THORChain referral bps: the row shows the referrer's cut as the user's saving, and at the top VULT tier the code silently costs 10 bps [Swap]

## Measured ground truth

Everything below rests on a live thornode measurement, BTC.BTC to ETH.ETH, 1 BTC, against the same gateway the app ships (`/thorchain/quote/swap`):

```
no affiliate              expected_out=3142015116   fees.affiliate=0
affiliate=va      bps=50  expected_out=3126304993   fees.affiliate=15710123
affiliate=t/va    bps=10/35  expected_out=3127876005   fees.affiliate=14139111
affiliate=va      bps=45  expected_out=3127876005   fees.affiliate=14139111   <-- identical to 10/35
affiliate=va      bps=35  expected_out=3131018030   fees.affiliate=10997086
affiliate=t/va    bps=10/0   expected_out=3138873091   fees.affiliate=3142025
affiliate=va      bps=0   expected_out=3142015116   fees.affiliate=0
```

**thornode sums the multi-affiliate legs.** `10/35` is charged exactly as flat `45` — identical `expected_amount_out` and `fees.affiliate` to the unit. It is not a carve-out of a fixed 50: if it were, `10/35` would have matched flat `50`, and it does not.

`15710123 / 3142015116` = 50.00 bps. `14139111 / 3142015116` = 45.00 bps. The delta is **exactly 5.00 bps**.

So the mechanism is: Vultisig cuts its own leg from 50 to 35 (−15) and adds the friend at 10 (+10). Friend +10, Vultisig −15, **user −5**.

## Problem 1 — the referral row overstates the saving 2x

`SwapDiscountChecker.kt:60-61` uses the referrer's share as the user's discount:

```
val referralBpsDiscount =
    THORChainSwaps.REFERRED_USER_FEE_RATE_BP.takeUnless { tierType == TierType.ULTIMATE }
```

[...]

Issue #5329, opened 2026-07-18
Swap overview: Network Fee & Swap Fee differ between initiator and joined devices


aminsato:
# [Fix] Network Fee & Swap Fee mismatch between initiator and joined devices [Swap overview]

## Problem
On the Swap overview screen (pre-keysign confirmation), the Network Fee and Swap Fee shown on the initiator device do not match the values shown on the joined (paired) device for the same swap transaction. Since both devices are co-signing one identical transaction, the fee breakdown must be identical on every device. The Swap Fee difference is the most severe ($0.06 vs $0.38, ~6×); Network Fee differs slightly (0.00040079 vs 0.0004042 ETH), and Max. Total Fee differs by $0.34 ($0.80 vs $1.14).

## Expected Behavior
For a single swap being co-signed, every device in the session must display the exact same Network Fee, Swap Fee, and Max. Total Fee. The fee breakdown shown to the user should come from one shared source of truth for the session (e.g. the initiator's quote/fee snapshot shared with joined devices) rather than being recomputed or re-fetched independently per device.

## Steps to Reproduce
1. Create/join a swap on two paired devices for the same vault.
2. Configure a swap (e.g. `0.00000799 wstETH` → min payout `0.0000098 ETH` via 1Inch on ETH).
3. Advance both devices to the **Swap overview** screen.
4. Compare the **Network Fee**, **Swap Fee**, and **Max. Total Fee** rows on the two devices.

**Result:** Network Fee and Swap Fee differ between the two devices, while swap amount, min payout, and provider match.

### Must Do
- [ ] Identify where the joined device obtains the quote/fee data rendered on the Swap overview screen, and confirm whether it recomputes/re-fetches fees locally instead of using the initiator's values.
- [ ] Make the fee breakdown (Network Fee, Swap Fee, Max. Total Fee) come from a single shared source of truth for the keysign session, so all paired devices display identical values.
- [ ] Ensure the displayed fees correspond to the actual transaction being co-signed (no device shows a fee lower than what will really be paid).

### Must NOT Do
- Don't change unrelated code
- Don't alter the actual swap/keysign transaction construction or signing logic — this is a display/data-source consistency fix
- Don't change how the initiator fetches its quote from the provider (1Inch)

## Acceptance Criteria
- [ ] `./gradlew assembleDebug` succeeds
- [ ] For the same swap session, initiator and joined devices display identical Network Fee, Swap Fee, and Max. Total Fee on the Swap overview screen
[...]

PR #5812, opened 2026-09-02
fix(swap): show the list rate in the Swap Fee row, not the net one

### LINKED ISSUE #5803: [Fix] Swap fee row shows the discounted 0.30% and the VULT cut [Swap]
**Screen:** Swap → fee breakdown (expanded)
**Figma:** https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=78798-72683

Same bug as iOS. `formatAffiliatePercent` in `SwapQuoteManager.kt` is documented as net-of-discount (`BASE_AFFILIATE_FEE_BPS - vultBPSDiscount`), so Gold shows **Swap Fee (0.30%)** and then `VultDiscountRow` shows **-20 bps** on top. Looks like two cuts.

Live Gold example (200 USDC): Vultisig Fee (0.30%) $0.59 + VULT (-20 bps) -$0.39, while Total uses only the reduced $0.59.

Figma: swap fee is the **list** rate (0.50%), discounts sit below under Applied Discounts. Total is still net.

**Do**
- Swap Fee row: **0.50%** and the **undiscounted** dollar amount (`BASE_AFFILIATE_FEE_BPS`, not net).
- VULT / Referral rows: the discount only.
- Total Fees stays net. Expanded rows must add up to it.

**Don't** change the bps sent to the provider. Form / Verify / Done should match.

aminsato:
Closes #5803

## Problem

The swap fee breakdown netted the VULT tier discount into the **Swap Fee** row — title *and* amount — and then itemized that same discount on its own row directly underneath. Read together, the panel billed one discount twice:

> Swap Fee (0.30%) **$0.59**
> VULT (Gold -20 bps) **-$0.39**
> Total Fees **$0.59** ← uses only the reduced fee

Figma ([node 78798-72683](https://www.figma.com/design/puB2fsVpPrBx3Sup7gaa3v/Vultisig-App?node-id=78798-72683)) shows the swap fee at the **list** rate, with the discounts below it.

## What changed

- `formatAffiliatePercent()` returns the list rate `0.50%` and no longer takes the tier discount. Every provider quotes against the same base, so the title is uniform.
- New `undiscountedSwapFee()` next to it computes the row's amount as **the affiliate fee the provider actually charged + the discounts the rows below itemize**.
- `SwapQuotePipeline.buildSuccess` (form) and `SwapTransactionToUiModelMapper` (verify + done) both call it, so the three screens read the same number.
- **Total Fees stays net**, and the bps sent to the provider are untouched.

For the Gold example above the row becomes `Swap Fee (0.50%) $0.99`, the VULT row keeps `-$0.39`, and Total Fees is still `$0.59`.

## On device

THORChain ETH → BTC on a Gold vault, $39.39 of source:

The rows reconcile on screen — `1.04 + 0.20 + 0.80 − 0.08 = 1.96`, the Total Fee — and Total Fee is still built from the 30 bps actually charged ($0.12), not the $0.20 displayed. Before this change the row read `Swap Fee (0.30%) $0.12` above a `-$0.08` discount.

## Why grossing up from the charged fee, rather than `50 bps × source`

- It is the only form in which the expanded rows **add up to the net Total Fee**: whatever the discount rows subtract is exactly what was added here.
- It survives the **Ultimate tier**, where the provider charges $0 and there is no non-zero fee to scale from — the row still discloses the 50 bps waived.
- When the source token has no fiat price it degrades to the provider's own figure instead of a fabricated `$0.00`.

## Notes for review

- **Cross-platform:** the extension ships the shape this issue calls a bug (`SwapProviderFeeRows.tsx` titles the row with the net bps and shows the charged amount), and iOS has it too — as the issue states. So Figma and the issue text drove this rather than a sibling implementation.
- **Coupling with #5765:** that issue's referral row claims 10 bps where the user's real saving is 5, so on a *referred* THORChain swap this row currently reads ~55 bps. The helper adds back whatever the rows claim, so it self-corrects the moment #5765 fixes that number — nothing to change here.
- **Co-signer unaffected:** `JoinSwapUiModelBuilder` sets `swapFeePercent = null` and renders no discount rows (it cannot resolve the initiator's tier), so it keeps showing the charged fee.
- **Deliberately out of scope:** Figma also has an `Applied Discounts:` header row and percentage-style labels (`VULT Gold: 20%`) where Android has `VULT (Gold -20 bps)` and no header. Not part of the issue's ask; happy to fold it in if wanted.

## Test plan

- `./gradlew :app:testDebugUnitTest` — 2372 tests, 0 failures. 9 new:
  - `SwapQuoteManagerTest` — list-rate title; the fee arithmetic across a tier discount, the Ultimate tier, a referral, and the no-discount / unpriced-source fallbacks.
  - `SwapQuotePipelineSwapFeeRowTest` (new) — the form's row is grossed up while `swapFeeFiat` (Total Fees) stays net.
  - `SwapTransactionToUiModelMapperFeeBreakdownTest` — the verify/done row is grossed up, and untouched when nothing was discounted.
- `./gradlew :app:lintDebug` — clean.
- ktfmt applied.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved swap fee displays by recalculating list-rate fees and VULT or referral discounts using current source values.
  * Preserved accurate net fees when discounts cannot be valued or reconstructed.
  * Improved handling of missing, zero, or unavailable source pricing.
  * Maintained appropriate fee visibility for supported swap types and quoted fees.
  * Applied consistent discounted fee displays during swap signing.

* **Tests**
  * Added coverage for discounted, waived, undiscounted, and provider-specific fee scenarios.
  * Added regression coverage for changing source values and unavailable pricing.


app/src/main/java/com/vultisig/wallet/ui/models/mappers/SwapTransactionToUiModelMapper.kt
```diff
@@ -178,7 +179,18 @@ constructor(
...
+                    // List-rate fee, matching the swap form: the charged affiliate fee plus the
+                    // discounts the rows below restate, so the row is not netted twice.
+                    fiatValue =
+                        fiatValueToStringMapper(
+                            undiscountedSwapFee(
+                                netFee = swapFeeFiat ?: quotesFeesFiat,
+                                srcFiat = srcFiat,
+                                vultBpsDiscount = from.vultBpsDiscount,
+                                referralBpsDiscount = from.referralBpsDiscount,
+                            ),
```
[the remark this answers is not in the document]
aminsato: Valid — fixed in 8a02f9167, though from the other side.

The mismatch is real: `srcFiat` and the affiliate fee are both re-valued here at the current price, while `vultBpsDiscountFiatValue` is the string `SwapTransactionBuilder` froze at Continue. Pairing a current-price gross with a quote-time discount row means they stop subtracting back to the net fee once the source price moves.

I didn't reconstruct from the recorded amounts, because freezing the fee row doesn't restore reconciliation — it just moves the seam. Total Fee is built from `swapFeeFiat`/`outboundFeeFiat`, which are re-valued live too, so a frozen gross would then disagree with the total by the same drift it fixed against the rows. Everything on this screen except `gasFeeFiatValue` is already a current-price figure.

So both sides now come off one snapshot instead: `bpsOfSourceFiat(srcFiat, bps)` prices each discount, and the same values feed `undiscountedSwapFee` and the discount rows the screen renders. Gross − rows = the net fee the total is built from, exactly, whatever the price does. It also declines to price a discount when the source has no price, so an unpriced token leaves the fee row on the provider's own figure rather than grossing it up by a fabricated zero. The recorded strings stay as the fallback for that case.

Covered by `revalues the discount row with the fee, so a price move cannot unbalance the panel` in `SwapTransactionToUiModelMapperFeeBreakdownTest` — source doubled since the quote, row shows the current $0.80 rather than the recorded $0.40, and $1.98 − $0.80 lands on the $1.18 in Total Fee.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -86,17 +93,56 @@ internal data class QuoteFetchResult(
...
+ * Formats the affiliate fee percentage shown in the Swap Fee row: the **list** rate
+ * ([BASE_AFFILIATE_FEE_BPS]), not the net bps sent to the provider. Every discount already has its
+ * own row directly below, so netting them into this title too billed one discount twice — a Gold
+ * vault read "Swap Fee (0.30%)" and then "VULT (Gold -20 bps)" underneath, as if both applied.
+ * Returns `"0.50%"`.
  */
-internal fun formatAffiliatePercent(vultBPSDiscount: Int?): String {
-    val netBps = (BASE_AFFILIATE_FEE_BPS - (vultBPSDiscount ?: 0)).coerceAtLeast(0)
-    return String.format(Locale.US, "%.2f%%", netBps.toBigDecimal().divide(BigDecimal(100)))
+internal fun formatAffiliatePercent(): String =
```
rkokhatskyi, line 107: `formatAffiliatePercent` now hardcodes the list rate as the title, but `bpsOfSourceFiat` returns null for an unpriced source, so `undiscountedSwapFee` falls back to the net charged fee. A Gold-tier swap of an unpriced token shows Swap Fee (0.50%) $0.60 with a VULT discount row reading -$0.00, and the same panel reaches verify and done.

app/src/main/java/com/vultisig/wallet/ui/models/mappers/SwapTransactionToUiModelMapper.kt
```diff
@@ -178,7 +188,16 @@ constructor(
                 ValuedToken(
                     token = tokenValue,
                     value = (from.swapFee ?: from.estimatedFees).value.toString(),
-                    fiatValue = fiatValueToStringMapper(swapFeeFiat ?: quotesFeesFiat, asFee = true),
+                    // List-rate fee, matching the swap form: the charged affiliate fee plus the
+                    // discounts the rows below restate, so the row is not netted twice.
+                    fiatValue =
+                        fiatValueToStringMapper(
+                            undiscountedSwapFee(
```
rkokhatskyi: `undiscountedSwapFee` grosses the initiator's Swap Fee row to the list rate, but `JoinSwapUiModelBuilder.kt:236` (unchanged) still renders the net wire-stamped fee for Kyber/LiFi/Jupiter, the exact value #5329 stamped so both co-signers would match. Gold tier on a $200 swap: $1.00 here versus $0.60 on the joiner's screen for the same transaction.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
```diff
@@ -463,10 +465,37 @@ internal class SwapQuotePipeline(
...
-        val feeText = if (isSwapKitUtxoSwap) "" else quoteResult.feeText
+        // The row is titled with the list rate, so its amount is the fee the provider charged plus
+        // every discount the rows below it itemize: together they are what the swap would have
+        // cost undiscounted, and subtracting those rows lands back on the net Total Fee.
+        //
+        // Empty text hides the row entirely (iOS `swapFeeString` returns `.empty` here): a SwapKit
+        // UTXO deposit's cost is already surfaced as the Network Fee, so there is no separate swap
+        // fee to show rather than a redundant "$0.00" row. A provider that bakes its fee into the
+        // quoted rate itemizes no amount to gross up and keeps its "included in quoted rate" copy.
+        val feeText =
```
rkokhatskyi, line 491: For a non-UTXO SwapKit route this grosses `quote.fees`, which `SwapKitQuoteSource.kt` documents as the source-chain inbound/deposit fee, not an affiliate charge. Adding the VULT/referral bps on top fabricates a tier discount on an unrelated network cost, inflating the Swap Fee row on every tiered-VULT cross-chain SwapKit swap.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -125,6 +125,7 @@ constructor(
             swapDiscountChecker = swapDiscountChecker,
             swapGasCalculator = swapGasCalculator,
             swapValidator = swapValidator,
+            fiatValueToString = fiatValueToString,
```
rkokhatskyi, line 128: `buildRouteOptions` (`SwapQuotePipelineController.kt:612`, unchanged) still reads `fetch.swapFeeFiat`, the net fee, for the Select Route picker row, while this now grosses the same route's fee to the list rate once picked. A Gold-tier Kyber swap shows $0.60 in the picker and $1.00 on the breakdown a tap later.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipeline.kt
aminsato: Valid, and the worst of the four — fixed in 66f5b6c19.

Confirmed the reading: `SwapKitQuoteSource` builds the quote's `fees` from `inboundFeeRawUnits`, the source-chain inbound/deposit cost, and SwapKit's affiliate fee is baked into `expectedDstValue` and never itemized (it goes out as `affiliateFee = request.affiliateBps` and comes back in the rate). So there was no affiliate charge here to gross, and the tier bps were being added onto a network cost.

I checked every sibling before fixing, because the same shape could have been elsewhere: Kyber (`SwapQuoteManager.kt:1076`) resolves the real `tx.swapFee`, LI.FI and Jupiter (`:1210`) carry the integrator/platform fee, and 1inch is fenced off by `swapFeeIncludedInRate`. SwapKit is the only one whose itemized amount isn't an affiliate fee.

`swapFeeRow()` now refuses to gross it, and this is where it goes slightly past what you asked, so flagging it: **the row also carries no rate for SwapKit any more.** Titling a deposit cost "0.50%" is false however the amount is valued, so keeping the label while dropping the gross-up would just move the lie. That does mean SwapKit non-UTXO diverges from #5803's "show 0.50%" for that one provider — say the word if you'd rather keep the label and I'll put it back.

Covered by `never grosses a SwapKit fee, which is an inbound cost no discount came off` (mapper) and `leaves a SwapKit inbound fee alone rather than grossing a network cost` (pipeline).

One adjacent case I deliberately left alone: Kyber's `resolveSwapFee` falls back to `gasFees` when the fee-token lookup fails, so a rare Kyber quote shows gas under the Swap Fee label — and grossing that is wrong for the same reason. It predates this PR, and gating it means threading a resolution flag through `SwapQuote.OneInch` in the `data` module. Happy to file it.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Valid — fixed in 66f5b6c19.

`buildRouteOptions` now grosses through the same `swapFeeRow()` the breakdown uses, so the picker and the panel it opens can never price one route two ways. It passes no `listRate`, since the row shows a fee and no percentage.

Two details that made this more than swapping the argument:

- The discounts have to be resolved **per candidate**, not from the winning quote's `discountInfo` — each row is a different provider, and the referral discount applies to THORChain alone. That's `QuoteCandidate.discountBps()`, and to stop it drifting from `buildSuccess` I pulled the Ultimate rule out into `referralBpsFor(tierType)` so both read the same source.
- The netFee it grosses is `fetch.swapFeeFiat`, which for THOR/Maya is affiliate + outbound. That's deliberate: gross − net is the discounts either way, so the row stays "swap + outbound, undiscounted" and keeps matching the breakdown's two rows.

Your Kyber example now reads $1.00 in both places. Covered by `grosses without a rate string for a row that shows no rate` and the two `discountBps` cases in `SwapFeeRowTest`.

app/src/main/java/com/vultisig/wallet/ui/models/mappers/SwapTransactionToUiModelMapper.kt
aminsato: Valid — fixed in 66f5b6c19, and this was the one I had accepted as a known cost when I opened the PR. You're right that it shouldn't be: #5329 exists precisely so both co-signers see the same fee on the transaction they're both signing.

The join screen now resolves the tier itself rather than reading it off the wire. `GetDiscountBpsUseCase` derives the discount from the **vault's** VULT balance and tier NFT, via the Ethereum address derived from the vault — and the joiner is a device of the same vault, so the same inputs give the same bps. No commondata change needed. Gold on a $200 Kyber swap now reads "Swap Fee (0.50%) $1.00" with a "-$0.40" VULT row on both screens, and the total stays the $0.60 net on both.

Why not the wire: adding `vult_bps_discount` to `OneInchTransaction` (the #5329 pattern) is exact and race-free, but commondata is a cross-platform contract and iOS/Windows senders wouldn't populate it — so the join screen would need this local derivation as the fallback anyway. Happy to add the field later as a precision upgrade; it isn't needed for correctness here.

The trade-offs, stated plainly:
- A VULT balance that crosses a tier boundary between the initiator's quote and the joiner's read prices the row differently. That's the same approximate parity the Thor/Maya/SwapKit fee re-fetches on this path already accept, and it's documented at the call site.
- A failed or unresolvable read yields no discount, which leaves the row exactly as it is today: the charged fee, claiming no rate.
- Only the VULT tier is re-derivable, and only it is needed — referral discounts are THORChain-only (`SwapQuotePipeline.kt:422`), which is a payload of its own.
- SwapKit-on-EVM rides `SwapPayload.EVM` tagged `provider = "SwapKit"`, not `SwapPayload.SwapKit`, so it reaches this same branch. It's excluded from grossing here for the reason in your other thread — without that check this fix would have reintroduced that bug on the co-signer.

Worth noting for the direction of the whole PR: the **THORChain** join branch already showed the gross. `JoinSwapUiModelBuilder.kt:350` re-fetches with a bare `SwapQuoteRequest`, so `bpsDiscount` defaults to 0 and the quote comes back at the full 50 bps. Grossing the initiator moves it toward that screen rather than away from it; the EVM branch was the outlier.

New tests in `JoinSwapVultDiscountTest`: the Gold case above, and a no-tier vault whose row stays at the charged fee.

Separately, and pre-existing so I left it: that Thor join branch's **total** is built from the gross fee, so it reads higher than the initiator's net total. Not touched here — tell me if you want it as a follow-up.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Valid — fixed in 66f5b6c19, including the "-$0.00" half.

Traced it exactly as you describe: `bpsOfSourceFiat` bails at `srcFiat.value <= 0`, so nothing gets added back and the amount stays the net $0.60, while the title had become an unconditional "0.50%". And the row underneath still rendered, because `vultBpsDiscountFiatValue` comes from `ConvertBpsToFiatUseCase`, which happily returns $0.00 against an unpriced source. The KDoc claimed the fallback avoided "a fabricated zero" — it avoided it in the fee row and left it in the discount row.

`swapFeeRow()` now returns the amount and the rate together, so they can't disagree: when a discount applies that the source can't price, the fee stays the provider's own figure **and** the rate is dropped. The row then reads "Swap Fee $0.60" with no rate claim, which is honest about what's known.

The rows follow the fee for the same reason — they're rendered only when it was grossed to the list rate. A row that subtracts from a fee it was never added to can't reconcile against the total, and that's true both for the unpriced source and for SwapKit. This is the second place I went past the literal comment, so flagging it: it removes the VULT row from those two cases rather than showing it at zero.

That also let me drop the mapper's fallback to the discount string the form recorded. It was priced at quote time against a fee that's re-valued at display time — the pairing from the round-1 thread — so when the current price can't value the discount, the row is now hidden rather than restated stale.

`isListRate` on `SwapFeeRow` is kept separate from `percent != null` deliberately: the route picker grosses but shows no rate, and conflating the two silently hid the discount rows in two existing tests before I split them.

Covered by `drops the rate and the discount row when the source has no price` (pipeline), `drops the rate and the discount row when the source has lost its price` (mapper), and `keeps the charged fee when a discount applies that the source cannot price` (`SwapFeeRowTest`).

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
```diff
@@ -261,6 +267,40 @@ constructor(
...
+                val vultDiscountFiat =
+                    bpsOfSourceFiat(
+                        convertTokenValueToFiat(srcToken, srcTokenValue, currency),
+                        vultBps,
+                    )
+                val feeRow =
+                    evmProvider?.let {
+                        swapFeeRow(
+                            provider = it,
+                            netFee = estimatedFee,
```
johnnyluo, line 299: **Correctness:** For LI.FI routes, the co-signer join screen double-grosses the Swap Fee by the VULT discount: `estimatedFee` already comes from `LiFiChainApi.integratorFeeAmount(dstAmount, bpsDiscount = 0)` (list rate), but it's then fed as `netFee` here into `swapFeeRow`, which adds the discount on top again.

A Gold-tier vault co-signs a LI.FI swap: `integratorFeeAmount` already returns the undiscounted 0.50% fee, then `swapFeeRow(netFee = estimatedFee, pricedDiscounts = [vultDiscountFiat])` calls `undiscountedSwapFee`, which adds `vultDiscountFiat` a second time. The co-signer's Swap Fee row shows a fee inflated by the discount amount, disagreeing with the initiator's screen and the correct Total Fee.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
```diff
@@ -604,13 +605,26 @@ constructor(
...
                             srcToken.chain.standard == TokenStandard.UTXO)
+                // Grossed to the list rate exactly as the breakdown this row opens does, so a route
+                // can't advertise one fee in the picker and a larger one a tap later. Whatever the
+                // breakdown adds back onto the charged fee is added here too, off the same source
+                // snapshot — the row carries no rate of its own, so it passes no [listRate].
+                val discountBps = candidate.candidate.discountBps()
+                val pickerFee =
+                    swapFeeRow(
+                            provider = candidate.candidate.provider,
+                            netFee = fetch.swapFeeFiat,
```
johnnyluo, line 615: **Correctness:** The Select-route picker grosses `fetch.swapFeeFiat` (affiliate + outbound combined) instead of `fetch.affiliateFeeFiat` (affiliate-only, added in this PR specifically to avoid this).

Picking a THORChain/MayaChain route with an outbound fee and a nonzero discount: the picker's fee is inflated versus the correct value shown a tap later on the breakdown screen (which uses `affiliateFeeFiat`) — the exact 'one fee in the picker, a larger one after a tap' bug the comment right above this block says it prevents.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
```diff
@@ -261,6 +267,40 @@ constructor(
                     if (isOneInchIncludedInRate) networkGasFeeFiatValue
```
johnnyluo, line 292: **Correctness:** For 1inch on this join path, `estimatedFee` (defined here) is a gas-cost placeholder, not an affiliate charge — yet the new `feeRow` block below (line ~294) grosses it by the VULT discount anyway. `swapFeeRow`'s provider guard excludes only `SwapProvider.SWAPKIT`, not 1inch / `swapFeeIncludedInRate`.

A Gold-tier vault co-signs a 1inch swap: `providerFee.fiatValue` ends up set from a grossed gas-placeholder rather than any real fee. Currently masked because current consumers (verify/done/overview) substitute 'included in quoted rate' text when `swapFeeIncludedInRate` is true, but any future consumer of `providerFee.fiatValue` would render a fabricated number. The same gap exists in `SwapTransactionToUiModelMapper.kt` (`netFee = swapFeeFiat ?: quotesFeesFiat`).

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -86,23 +94,121 @@ internal data class QuoteFetchResult(
...
+        "%.2f%%",
+        BASE_AFFILIATE_FEE_BPS.toBigDecimal().divide(BigDecimal(100)),
+    )
+
+/**
+ * Prices [bps] against the source notional — the same arithmetic ConvertBpsToFiatUseCase uses for
+ * the discount rows themselves, so a row and the fee it was taken off can be derived from one
+ * snapshot. Null when there is nothing to show: no discount, or a source with no fiat price.
+ */
+internal fun bpsOfSourceFiat(srcFiat: FiatValue, bps: Int?): FiatValue? {
```
johnnyluo, line 120: **Simplification:** `bpsOfSourceFiat` re-implements the bps-to-fiat formula that `ConvertBpsToFiatUseCase` already provides and that `SwapDiscountChecker` already calls for the same discount rows — the KDoc above even says 'the same arithmetic ConvertBpsToFiatUseCase uses.'

A future rounding/precision fix to `ConvertBpsToFiatUseCase` won't reach `bpsOfSourceFiat`, so the Swap Fee row's grossing math and its own discount sub-rows could drift apart on edge cases — the same bug class (#5803) this PR exists to fix.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -86,23 +94,121 @@ internal data class QuoteFetchResult(
...
+ * source — is what keeps the expanded breakdown reconciling to the net Total Fee: whatever the
+ * discount rows subtract is exactly what was added here, which holds only while both are read off
+ * the same [bpsOfSourceFiat] values. It also survives the Ultimate tier, where the provider charges
+ * nothing and there is no non-zero fee to scale.
+ */
+internal fun undiscountedSwapFee(netFee: FiatValue, waived: List<FiatValue?>): FiatValue {
+    val discounts = waived.filterNotNull()
+    if (discounts.isEmpty()) return netFee
+    return FiatValue(
+        value = discounts.fold(netFee.value) { total, discount -> total + discount.value },
```
johnnyluo: **Simplification:** `undiscountedSwapFee` hand-folds `FiatValue.value` with raw BigDecimal addition instead of using `FiatValue`'s existing `plus` operator, which also asserts currency equality.

Every other `FiatValue` addition in the codebase fails fast on a currency mismatch via that operator; this fold skips the guard, so a caller that ever passes a `waived` value in a different currency than `netFee` would silently produce a wrong sum instead of throwing.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -86,23 +94,121 @@ internal data class QuoteFetchResult(
...
+ * - **The itemized amount is not an affiliate charge.** SwapKit reports its source-chain inbound
+ *   (deposit) cost as the quote's fee and bakes its affiliate fee into the quoted destination
+ *   amount instead, so adding a tier discount there would invent a discount on a network cost and
+ *   inflate the row on every tiered cross-chain route — and titling that cost with an affiliate
+ *   rate would be wrong however it is valued.
+ * - **A discount applies but the source has no fiat price.** Nothing can be added back, so the
+ *   amount stays the discounted one, and "0.50%" over it would restate the very double-billing read
+ *   this row exists to fix (#5803).
+ */
+internal fun swapFeeRow(
```
johnnyluo, line 192: **Simplification:** The `discountBps`/`pricedDiscounts` pairing this helper requires is hand-assembled at four separate call sites (`JoinSwapUiModelBuilder`, `SwapTransactionToUiModelMapper`, `SwapQuotePipeline`, `SwapQuotePipelineController`) instead of behind one shared resolver, despite the PR's stated goal of routing the fee row through 'one helper.'

A future change to discount rules (a third discount type, a stacking-order change) requires updating all four call sites in lockstep; missing one — most plausibly the less-visited picker or join path — silently reintroduces the picker/breakdown or initiator/co-signer fee mismatch this PR was written to close.

johnnyluo (commented):
Automated follow-up review after the CodeRabbit thread fixes — found 6 more issues in the same fee-grossing refactor.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
```diff
@@ -86,23 +94,121 @@ internal data class QuoteFetchResult(
...
+    provider: SwapProvider,
+    netFee: FiatValue,
+    listRate: String?,
+    discountBps: List<Int?>,
+    pricedDiscounts: List<FiatValue?>,
+): SwapFeeRow {
+    if (provider == SwapProvider.SWAPKIT) {
```
rkokhatskyi: `swapFeeRow`'s guard (`SwapQuoteManager.kt:175`) excludes only `SwapProvider.SWAPKIT`. For a 1inch swap with a VULT discount, `VultDiscountRow` still renders a fabricated dollar saving under a fee correctly masked as `included in quoted rate`, since 1inch's affiliate fee is never itemized and the discount has nothing real to subtract from.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
```diff
@@ -306,12 +346,14 @@ constructor(
...
-                        // No percentage on the join path: the affiliate rate is net of the
-                        // initiator's VULT tier discount, which the signed payload doesn't carry
-                        // and can't be recovered here. Showing the base 0.50% would contradict a
-                        // discounted initiator's verify screen, so the row omits the percent
-                        // entirely — as every other provider does on this path (#5358 review).
-                        swapFeePercent = null,
+                        swapFeePercent = feeRow.percent,
+                        // Rows follow the fee: shown only when it was grossed to the list rate, so
+                        // subtracting them lands back on the net fee the total is built from.
+                        vultBpsDiscount = vultBps?.takeIf { feeRow.isListRate },
```
rkokhatskyi, line 355: `JoinSwapUiModelBuilder.kt:396` (ThorChain) and `:473` (MayaChain), both unchanged, still build `SwapQuoteRequest` with no `bpsDiscount`, so the co-signer's fee stays undiscounted while the EVM branch above now resolves the vault's real tier locally for the identical #5329 reason. Total Fee on those two paths still diverges from the initiator's for any VULT-tier vault.

rkokhatskyi (changes requested):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
aminsato: Valid — fixed in b73fc2bf1.

Confirmed exactly as described: `LiFiChainApi.integratorFeeAmount` defaults `bpsDiscount = 0`, so what came back was already the list-rate fee, and `swapFeeRow` then added the discount on top of it. The initiator doesn't have this problem because `SwapQuoteManager.kt:1251` passes `bpsDiscount = vultBPSDiscount ?: 0`; the join path passed nothing. The comment sitting directly above that line — "VULT tier discount isn't available in the join flow" — is what made it look safe, and it stopped being true in the commit that added the tier derivation.

Two things I'd add, both of which change the shape of it:

- It only fires when the payload carries no `swap_fee_chain` / `swap_fee_token_id` / `swap_fee_decimals` stamp, because `explicitSwapFee` short-circuits ahead of the LI.FI branch. So current Android senders were fine and iOS / Windows / older-Android senders were not — a cross-platform-only bug.
- `isLiFi` is a heuristic, not a provider check: `srcToken.chain != dstToken.chain && provider != SwapKit`. Any cross-chain non-SwapKit route reaches that branch, so the double-gross wasn't confined to LI.FI.

The fix charges the tier first (`bpsDiscount = vultBps ?: 0`), which makes `value` the same net fee the initiator's own quote produces — so the token amount on the row matches too, not just the fiat. That meant hoisting the tier resolution above the `value` block.

Covered by `charges the tier before grossing a LI-FI fee it derives itself` in `JoinSwapVultDiscountTest`: 200 USDC out, Gold vault — reverting the fix reads $1.40 against the correct $1.00.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuotePipelineController.kt
aminsato: Half right, and the half that's right is worth fixing — but not the way the comment proposes, so I've left the code and rewritten the doc it contradicts (b73fc2bf1).

You're correct that the picker grosses `swapFeeFiat` while the breakdown grosses `affiliateFeeFiat`, and that the KDoc I added to that field says the Swap Fee row "cannot be grossed up to the list rate from" `swapFeeFiat`. That sentence was written about the breakdown and reads as a blanket rule; the picker column is a different quantity and the doc didn't say so. That's a real defect in what I wrote and it's now explicit.

Where I don't follow is "inflated". `swapFeeFiat` is `affiliate + outbound`, and gross − net is the discounts either way, so the picker shows `affiliate_list + outbound` — the undiscounted cost of both legs, not an inflated one. The picker-vs-breakdown gap is exactly `outbound`, which is what it was before this PR too: the old code rendered `fetch.swapFeeFiat` verbatim. The bug the comment above that block warns about is a route advertising one fee and then a *different* fee for the same rows — Roman's Kyber case, where `swapFeeFiat` and `affiliateFeeFiat` are the same figure — and that stays closed.

Switching to `affiliateFeeFiat` would drop the outbound leg from a column whose whole job is comparing whole routes against each other, so a THORChain row would understate its real cost next to a Kyber row that has no outbound leg at all. If you'd rather that column mean "the breakdown's Swap Fee row, exactly" and accept the understatement, say the word and I'll change it — it's a call about what the column means, not a correctness one.

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
aminsato: Valid — fixed in b73fc2bf1, together with Roman's thread on the same guard.

Confirmed: for literal 1inch `oneInchSwapTxJson.swapFee` is empty, so `value` falls through to the `gasPrice × gas` placeholder, and `swapFeeRow`'s guard named only `SWAPKIT` — so `providerFee.fiatValue` ended up a grossed gas number. You're right that it's masked today, and right that it shouldn't be left sitting behind a string substitution.

`swapFeeRow` now takes `feeIncludedInRate` and, when a discount applies, keeps the provider's own figure, drops the rate, and itemizes nothing. The mapper you point at passes `from.swapFeeIncludedInRate` for the same reason, so both surfaces close together.

One deliberate asymmetry, flagging it since it goes past the literal comment: with **no** discount the rate is still shown. 1inch's undiscounted charge really is 0.50%, and "Swap Fee (0.50%) — included in quoted rate" is the row that ships today; dropping the label there would be a regression unrelated to this bug. Only the discounted case has nothing honest to put in the title.

Covered by `neither grosses nor itemizes a 1inch fee, which is baked into the quoted rate` (join) and `never itemizes a 1inch discount, whose fee is baked into the quoted rate` (mapper).

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Checked this one and I'm leaving it as is — reasons, in case you want to push back.

The KDoc line you quote is accurate about the arithmetic, but that arithmetic is the only shared part, and the two have deliberately different contracts:

- `ConvertBpsToFiatUseCase` is `suspend`, injected, takes `(Coin, TokenValue, bps)`, reads the currency from `AppCurrencyRepository`, and returns **ZERO** when `bps <= 0`.
- `bpsOfSourceFiat` is pure, takes an already-priced `FiatValue`, and returns **null** when there is nothing to show: no discount, or a source with no fiat price.

That null is load-bearing. It is the signal `swapFeeRow` uses to keep the charged fee and drop the rate when a discount applies that the source cannot value — the `Swap Fee (0.50%) $0.60` with a `-$0.00` row underneath that Roman filed in round 2. Routing the grossing through the use case hands it a zero there and puts that bug straight back.

The drift you describe needs a rounding change inside a `× bps / 10_000` that reaches only one of the two. If you'd still rather have one definition of that multiply, the shape I'd accept is a pure `FiatValue.timesBps(bps)` that both call, leaving each one's zero/null contract alone — happy to add it, but it's a different change from what the comment asks for.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Valid — fixed in b73fc2bf1.

`undiscountedSwapFee` is now `waived.filterNotNull().fold(netFee) { total, discount -> total + discount }`, so the currency `require` runs on every addition and the `isEmpty` early return falls out of the fold for free. No live mismatch was reachable — every caller sources both sides from one currency — but there was no reason to opt out of the guard every other `FiatValue` addition gets.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Valid — fixed in b73fc2bf1, and your framing turned out to be the right one: the LI.FI bug in your other comment *is* this drift, one call site handing the helper a fee the others don't.

Callers now pass `SwapDiscountBps(vult, referral)` and the helper prices them itself off `srcFiat`, returning the two valued discounts on `SwapFeeRow`. That removes the parallel-list pairing entirely — there is no second list left to fall out of step — and it also moves the "rows follow the fee" rule inside, so the `takeIf { feeRow.isListRate }` dance disappears from the mapper and the join builder.

What I deliberately didn't unify is the four *inputs*, because they're genuinely four different sources: the pipeline reads `discountInfo`, the picker reads per-candidate bps (the referral applies to THORChain alone), the join screen derives the tier from the vault's own balance, and the mapper reads what the transaction recorded. The value they produce is what had to be shared, not where they read it from.

app/src/main/java/com/vultisig/wallet/ui/models/swap/SwapQuoteManager.kt
aminsato: Valid — fixed in b73fc2bf1.

Confirmed: the guard named only `SWAPKIT`, so 1inch came back `isListRate = true` and sailed past the `!isListRate` gate that exists to hide rows which can't reconcile. `swapFeeRow` now also takes `feeIncludedInRate`, and when a discount applies it keeps the charged figure, drops the rate, and itemizes nothing.

Two notes on scope:

- On the initiator this row predates the PR — the gate I added in round 2 is what should have covered it and didn't. On the **join** screen it is new in this PR, since the co-signer showed no discount row at all before, so that half is a regression I introduced and you caught.
- With no discount the rate stays. 1inch's undiscounted charge is 0.50%, so "Swap Fee (0.50%) — included in quoted rate" is true and is what ships today; only the discounted case has nothing honest to put in the title.

johnnyluo's thread on `JoinSwapUiModelBuilder.kt:267` is the same guard from the other side — a grossed gas placeholder landing in `providerFee.fiatValue` — and is closed by the same change.

Covered by `drops the rate and the discount row when a discounted fee is baked into the rate` (`SwapFeeRowTest`), `never itemizes a 1inch discount, whose fee is baked into the quoted rate` (mapper), and `neither grosses nor itemizes a 1inch fee, which is baked into the quoted rate` (join).

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
aminsato: Valid — fixed in b73fc2bf1, and this one corrects something I told you in round 2.

I said there that the THORChain join branch "already showed the gross", and that grossing the initiator moved the two screens together rather than apart. That was true of the fee row, and I stopped there. It is not true of the total: the co-signer's Total Fee is built from the same undiscounted `rawFees.affiliate` the bare request returned, so a tiered vault's two devices disagreed on the number they were both about to sign. Maya is identical. Thanks for pushing on it.

Both branches now resolve the vault's tier locally and pass it as `bpsDiscount`, so the node quotes what the initiator quoted, `swapFee` / `outboundFee` come back net, and the total matches. The row itself is grossed through the same `swapFeeRow` the EVM branch uses — which I moved into `buildSwapUiModel` rather than repeating at each call site, so every native branch gets the rate and the discount row instead of each one remembering to.

The residual, stated plainly: the referral code is the initiator's local state and nothing the vault can recompute, so a referred THORChain swap still prices its referral leg undiscounted on the co-signer. Smaller than the tier gap, and closing it properly needs the wire.

`JoinSwapNativeDiscountTest` covers the Gold case on both providers (the request carries 20 bps, the row reads $1.00 with a -$0.40 beneath it), the total held at the net $0.90 rather than the grossed $1.30, and a no-tier vault left on the charged fee.

johnnyluo (approved):
LGTM

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
```diff
@@ -347,6 +390,17 @@ constructor(
...
+                // initiator had signed a discounted one, so the two devices disagreed on the
+                // affiliate fee AND on the Total Fee built from it. The tier is a function of the
+                // vault's VULT balance and tier NFT and both devices are the same vault, so
+                // resolving it here reproduces the initiator's quote — the same derivation the EVM
+                // branch uses, for the same #5329 reason.
+                //
+                // The referral code is the one input that stays out of reach: it is the
+                // initiator's local state, not anything the vault can recompute, so a referred
+                // swap still prices its referral leg undiscounted here.
+                val thorVultBps = getDiscountBps(vault.id, SwapProvider.THORCHAIN).takeIf { it > 0 }
```
rkokhatskyi, line 403: `getVultBalance` (`GetDiscountBpsUseCase.kt:98-110`) returns a cached balance row with no TTL once any row exists, so a rarely-opened co-signer device can hold a stale VULT balance indefinitely. A wrong tier here re-quotes THORChain undiscounted, so the co-signer's Swap Fee and Total can diverge from what the initiator actually signed.

rkokhatskyi (dismissed):

app/src/main/java/com/vultisig/wallet/ui/models/keysign/JoinSwapUiModelBuilder.kt
aminsato: Valid — fixed in 3bcde2cd4.

Confirmed there is nothing to age the row against: `TokenValueEntity` is keyed on chain/address/ticker/contractAddress with no written-at column, and `getCachedTokenBalances` is a bare `tokenValueDao.getTokenValues(addresses)`, so "cached" here means "written once, returned forever". The 12s `LIVE_BALANCE_TTL_MS` never applied, because `getLiveBalance` sat on the right of the elvis and the live read persists itself — `getBalanceOrNull` → `getTokenValue(...).first()`, whose `onEach` inserts into the DAO. So the very first read a device ever performed became its permanent answer, and for a vault that doesn't hold VULT in its coin list nothing refreshes it: `DiscountTiersViewModel` gates its fresh fetch on `vault.coins.find { VULT } != null`, which is exactly the case the use case's own in-memory-coin fallback exists for.

`getVultBalance` now reads live first and keeps the cached row as the fallback for a failed read. The per-vault lock and TTL that already collapse a quote fetch's provider candidates into one `eth_call` are what make that affordable: one read per vault per 12s, against a quote round-trip that is heavier. Failing back to the stale row rather than to null is deliberate — a stale tier is closer to right than a fabricated "no discount" for a holder who has one.

Two things worth adding to how you framed it:

- The failure isn't only "re-quotes undiscounted". A stale-*high* row over-discounts the co-signer just as easily — the vault that has since sold its VULT — and that direction quotes a fee the initiator never signed rather than merely a larger one. Both directions are covered by tests.
- The staleness itself predates this PR and hit the initiator identically, since the swap form quotes off the same row. This PR is only what made the join screens depend on it. Fixing it in the use case closes both sides, so the two devices now derive the tier from the same live read rather than from two independently frozen ones.

Tests: live wins over a stale row in both directions, the cached row is used when the live read fails, and the chain is read once per vault across a quote fetch's candidates.

